PR #1875, opened 2016-03-23
[MRG+1] FS cache storage misuses mtime for expiration

nanolab:
FS cache storage misuses mtime for expiration

### DIFF
--- a/scrapy/extensions/httpcache.py
+++ b/scrapy/extensions/httpcache.py
@@ -281,7 +281,7 @@ class FilesystemCacheStorage(object):
         metapath = os.path.join(rpath, 'pickled_meta')
         if not os.path.exists(metapath):
             return  # not found
-        mtime = os.stat(rpath).st_mtime
+        mtime = os.stat(metapath).st_mtime
         if 0 < self.expiration_secs < time() - mtime:
             return  # expired
         with self._open(metapath, 'rb') as f:


Digenis:
What bug does this fix?
What modifications do you need to di in the directory
that happen to update its mtime?

nanolab:
When it create cache all works fine.
Once cache expires it checks directory mtime (.scrapy\httpcache\buyincoins\7e\7e2351803c5e337e02bd649c198ec6846b4147a9\) and downloads and create all files inside this directory again. But directory mtime stay unchanged.
And it downloads and downloads....
So it have to check files mtime not directory.
I'm on Windows 10, Python 2.7.11

Digenis:
I see. Yes, it's a bug.
Can you rename the commit message to something shorter (50 chars or less)
and meaningful on the same time?
E.g. `FS cache storage misuses mtime for expiration`.

nanolab:
Renamed.

Digenis:
doesn't seem like you did

Digenis:
No, this is the title. It seems you are not familiar with git, just github.
Anyway, doesn't matter for now, let's see what the maintainers have to say.

The change seems ok.
It should also be applied to newer releases.

nyov:
+1, looks like a safe change in any case.

This should be merged into master, though (and then maybe backported). But alas, changing the target branch in an open PR isn't possible - hopefully someone does manually merge it.

@nanolab, you can change the commit title by doing a `git checkout patch-1; git commit --amend -m "FS cache storage misuses mtime for expiration"` or the equivalent in whatever GUI, then `git push --force origin patch-1`

kmike:
Thanks @nanolab for the fix and @nyov and @Digenis for the review! I think that's not a big deal to leave commit message as-is; changing it would be nice, but not required. A good point about 1.0 branch; we should cherry-pick it to 1.1 and to master after merging.

redapple:
Cherry-picked to [1.1](https://github.com/scrapy/scrapy/commit/e7a5e3717ba0c1815a0cfb234397eb7640146804) and [master](https://github.com/scrapy/scrapy/commit/a583e4d531306b7628b42f8a32c7db892ad86cf1) branches.

Issue #1569, opened 2015-10-29
Add a page which explains important Scrapy concepts in a single page


kmike:
As @plafl said: "Scrapy is very extensible but that has a cost too. There are too many concepts: spiders, items, middlewares, pipelines, exporters, extensions, signals, settings. As a newcomer I would like to know which problem they solve."

+1 :) I think we should add a page which will explain all this in a single place - what are these concepts, when and why to use them.

Granitas:
Isn't that pretty much http://doc.scrapy.org/en/latest/topics/architecture.html?highlight=scrapy%20architecture ? Which is my favorite page in the docs.

kmike:
@Granitas yeah, it is close, a good catch. 

This page is not targeted for beginners though, e.g. both for downloader and for spider middleware it says just "They provide a convenient mechanism for extending Scrapy functionality by plugging custom code." without explaining which custom code should go to a spider mw and which should go to downloader mw. You can figure it out by meditating over the architecture overview picture, but it is not an easy task if you're just starting. Also, it doesn't explain extensions at all.

darshanime:
I can write this page, I have a few questions to get started: 

1. is the page is aimed at the extensions developer
2. should it read like an article or should it contain code
3. what is the desired length of the article (how many words)
4. what concepts should be focused on

Gallaecio:
This seems to go in the lines of a question I recently tried to answer on StackOverflow: https://stackoverflow.com/q/54421455/939364

PR #1879, opened 2016-03-25
DOC improved Architecture overview

kmike:
* spiders don't have to work on specific domains;
* explain what to use Downloader middleware for
  and what to use Spider middleware for;
* Engine no longer locates spiders based on domains;
* "Spider middleware output direction" step was missing - this one was bugging me since 2013 :)

See also: GH-1569.

This overview is still incomplete because Extensions and Scraper are not explained.

### DIFF
--- a/docs/topics/architecture.rst
+++ b/docs/topics/architecture.rst
@@ -48,8 +48,7 @@ Spiders
 
 Spiders are custom classes written by Scrapy users to parse responses and
 extract items (aka scraped items) from them or additional URLs (requests) to
-follow. Each spider is able to handle a specific domain (or group of domains).
-For more information see :ref:`topics-spiders`.
+follow. For more information see :ref:`topics-spiders`.
 
 Item Pipeline
 -------------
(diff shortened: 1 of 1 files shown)


docs/topics/architecture.rst
```diff
@@ -83,32 +99,30 @@ Data flow
...
 
-8. The Engine sends scraped items (returned by the Spider) to the Item Pipeline
-   and Requests (returned by spider) to the Scheduler
+7. The Engine passes scraped items and new Requests returned by a spider
+   through Spider Middleware (output direction), and then sends processed
+   items to Item Pipelines and processed Requests to the Scheduler.
 
-9. The process repeats (from step 2) until there are no more requests from the
+8. The process repeats (from step 2) until there are no more requests from the
    Scheduler, and the Engine closes the domain.
```
redapple: I admit I never read that in detail. Now I have :) It's clear.

except this "the Engine closes the domain.". what does it mean?

redapple:
I like the "Use ___ if you want to ___ or ___ or ____" style.

In a another PR perhaps, I think we should also add simplified [sequence diagrams](https://en.wikipedia.org/wiki/Sequence_diagram) for the data flow, so as to put the numbered steps in context.

docs/topics/architecture.rst
kmike: A good catch; I think that's another artifact from the past.

kmike:
@redapple there is a picture for this data flow in docs (http://doc.scrapy.org/en/latest/topics/architecture.html).

redapple:
I know @kmike :) but I still think it could be clearer. I'll suggest something concrete

eliasdorneles:
Nice PR, LGTM!

PR #1881, opened 2016-03-26
[MRG+1] Remove duplicate code now handled by newer w3lib

nyov:
Can't say if bumping the w3lib dependency to 1.13 is a good idea yet. But maybe it is once scrapy 1.2 sees a release. So this might be for then or later.

### DIFF
--- a/scrapy/downloadermiddlewares/ajaxcrawl.py
+++ b/scrapy/downloadermiddlewares/ajaxcrawl.py
@@ -8,7 +8,6 @@ from w3lib import html
 
 from scrapy.exceptions import NotConfigured
 from scrapy.http import HtmlResponse
-from scrapy.utils.response import _noscript_re, _script_re
 
 
 logger = logging.getLogger(__name__)
@@ -89,8 +88,7 @@ def _has_ajaxcrawlable_meta(text):
     if 'content' not in text:
         return False
 
-    text = _script_re.sub(u'', text)
-    text = _noscript_re.sub(u'', text)
-    text = html.remove_comments(html.replace_entities(text))
+    text = html.remove_tags_with_content(text, ('script', 'noscript'))
+    text = html.replace_entities(text)
+    text = html.remove_comments(text)
     return _ajax_crawlable_re.search(text) is not None
-
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Twisted>=10.0.0
 lxml
 pyOpenSSL
 cssselect>=0.9
-w3lib>=1.8.0
+w3lib>=1.13.0
 queuelib
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/scrapy/utils/response.py
+++ b/scrapy/utils/response.py
@@ -3,7 +3,6 @@ This module provides some useful functions for working with
 scrapy.http.Response objects
 """
 import os
-import re
 import weakref
 import webbrowser
 import tempfile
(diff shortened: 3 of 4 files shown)


nyov:
Oops. Fixed ajaxcrawl middleware.

The CI error with the docs now can't be related though:
```pytb
Exception occurred:

  File "conf.py", line 113, in <module>

    import sphinx_rtd_theme

ImportError: No module named 'sphinx_rtd_theme'
```

redapple:
@nyov , CI error on docs should be fixed with https://github.com/scrapy/scrapy/pull/1893

scrapy/downloadermiddlewares/ajaxcrawl.py
```diff
@@ -89,8 +88,7 @@ def _has_ajaxcrawlable_meta(text):
     if 'content' not in text:
         return False
 
-    text = _script_re.sub(u'', text)
-    text = _noscript_re.sub(u'', text)
-    text = html.remove_comments(html.replace_entities(text))
+    text = html.remove_tags_with_content(text, ('script', 'noscript'))
```
kmike: could you please check that the performance is not much worse?

nyov: Yup. Done.

script:
```py
import timeit

timed = timeit.timeit("""
texts = [
    '<html><head><meta name="fragment"  content="!"/></head><body></body></html>',
    "<html><head><meta name='fragment' content='!'></head></html>",
    '<html><head><!--<meta name="fragment"  content="!"/>--></head><body></body></html>',
    '<html></html>',
]

# [1] all of ajaxcrawl func
#for text in texts:
#    _has_ajaxcrawlable_meta(text)

# [2] w3lib
#for text in texts:
#    remove_tags_with_content(text, ('script', 'noscript'))

# [3] scrapy.utils.response
for text in texts:
    _noscript_re.sub(u'', _script_re.sub(u'', text))
""", setup="""
# [1]
#from scrapy.downloadermiddlewares.ajaxcrawl import _has_ajaxcrawlable_meta
# [2]
#from w3lib.html import remove_tags_with_content
# [3]
from scrapy.utils.response import _noscript_re, _script_re
""")

print timed
```
timings for 1000000 (timeit.default_number) loops:
```
[1] on master branch (scrapy.utils.response)
42.3605451584
[1] on this PR branch (w3lib)
54.5658538342

[2] w3lib.html
33.3376588821

[3] scrapy.utils.response
13.8655879498
```
So performance looks quite a bit worse without the regex compilation.

kmike: 20% slowdown doesn't look too bad. Could you please try it with some data of size 32KB (this is what AjaxCrawlMiddleware sends as `text`)?

kmike: I worry about performance only because this function is called for each response HTML if AjaxCrawlMiddleware is enabled.

kmike: Could you please check it with 32K bodies (this is what AjaxCrawlMiddleware passes as `text` argument)? I'm worried about the performance because when AjaxCrawlMiddleware is enabled this function is called for every response; according to https://github.com/scrapy/scrapy/pull/343 existing implementation runs with 2k/sec speed which is not very fast already.

kmike: 54 vs 42 (~20% less speed) sounds fine

kmike: Yeah, you need a website with `<meta name="fragment">` to see the difference, there are shortcuts for websites without this tag, e.g. a website created with wix.com.

nyov: Hmm, those templates without content all look a lot smaller than 32kb. So I just copied the beginning of such a template with the `<meta name="fragment" content="!"/>` in it and filled the rest of it (body) with lorem ipsum again.

Well that hurt. For comparability I kept the default loop number:
```
# master branch
1163.59052396

# w3lib (PR)
945.980225086
```
Since system load might have varied over that duration, another run with only 10.000 loops:
```
# master branch
11.6963450909

# w3lib (PR)
9.49189996719
```
I understand thats just a single page test-case. I don't feel like looking for more real-world cases though. Don't happen to have any broad crawl data to look up.

kmike: Thanks @nyov for checking that! As the speed difference is only in percents it looks OK to me.

nyov: Strangely enough it times slightly faster... I don't even know why.

PR #1794, opened 2016-02-19
[MRG+1] Use best practices for TLS connections when using Twisted>=14.0

redapple:
This is another take at the various TLS, SSL, SNI issues, assuming scrapy using Twisted 14.0+

It solved #1227 for me.

It changes the default TLSv1.0-only to SSLv23 (protocol version negotiation, "version flexible" [as OpenSSL calls it](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html))

A new setting `DOWNLOADER_CLIENT_TLS_METHOD` allows setting the TLS/SSL method for the HTTP/1.1 downloader.
Setting values are strings, mapped to `OpenSSL.SSL` methods, "TLS" being the default, version flexible, method.

PR #1887, opened 2016-03-29
[MRG+2] Bump Twisted dependency to 13.1.0

nyov:
Can someone recall what was wrong with Twisted 11 on precise?
(https://github.com/scrapy/scrapy/commit/8b0c9466d6733c6523aa8c630e70c607cd845128)

Or was txlib/tx only introduced to backport features? The commit message and Readme don't quite say the same thing.

```
This source files are adapted copies from Twisted trunk to support HTTP1.1
handler under Twisted >= 11.1 and Twisted <= 13.0.0
```

It seems all the `xlib.tx` Exceptions should have been available since Tx 11.1 (except for `ResponseNeverReceived`, which doesn't seem to be explicitly used though), so
I hope partial deprecation of `xlib.tx` might be fast-tracked by pulling them in from twisted-proper instead, which only requires bumping to Tx 11.1 and not 13.0 yet.

That way only the HTTP11 download-handler code should still depend on `xlib.tx`.

ping @dangra 

(I did not remove the Exceptions so that any later diff against twisted won't throw a massive changeset.)


nyov:
Okay then maybe a bump to Twisted 11.1 would allow rolling up parts of the bundled tx code and reduce the changeset in there? (Though that might be superfluous effort.)

But I think this little bump here might make sense, since pulling in the exceptions from Twisted (which appeared in 11.1) allows 3rd party extensions not to care about imports from `xlib.tx` then to get these exceptions, so these imports can be deprecated.

Also, Twisted 10 was dependency since a9c35198 (10/2013) - scrapy 0.20 release.
(Even old debian wheezy release already shipped with Twisted 12.0, probably nobody using anything older anymore?)

dangra:
I'm OK to bump minimal twisted version but if that is the case then let's go directly to a version that allows us to drop xlib.tx completely even if it doesn't work by default in Ubuntu 12.04. By now, ubuntu users should be running at 14.04 and even planning to upgrade to 16.04. 

May be it is just me, but tying python applications support to python package versions shipped with linux distributions is so old times that I would say let's forget about supporting specific distros :)

just my 2 cents.

nyov:
I was just saying that as debian is well known for a slow release cycle, maybe the slowest. So if _they_ had Twisted 12 in the last stable release, everyone else probably had it long before that ;)

And, well, if y'all can decide to go directly for Twisted 13 or so, that probably makes things easier. But I didn't want to jump too high here :)

dangra:
Just in time [debian 8.4 (Jessie)](https://www.debian.org/News/2016/20160402) is the stable and comes with twisted [14.0](https://packages.debian.org/jessie/python-twisted).

This is convenient as debian wheezy comes with twisted [12.0](https://packages.debian.org/wheezy/python-twisted) but we need 12.1 as minimum, although it is possible to install [13.2](https://packages.debian.org/wheezy-backports/python-twisted) from wheezy-backports.

For me it is clear, set twisted minimum to 12.1 and drop xlib.tx altogether.

kmike:
I'm fine with bumping the required Twisted verision and dropping xlib.tx.

nyov:
I didn't dig too deep into xlib.tx, but was under the assumption it might have backported code even from twisted 13 inside. (After all it does use xlib up to Twisted 13.0) If Twisted 12.1 is actually good enough, that's better of course. But can anyone confirm this for sure? (Or maybe a tiny wrapper for versions up to Twisted 13 will still be necessary. Just gotta find those leftover changes...)

I couldn't find the 'common' trunk commit where xlib.tx was taken from. There always were some changes on either side when `diff`ing. So I had trouble "rolling up" the changesets.

pawelmhm:
FYI There is ticket in twisted from core contributor about actively discouraging people from using versions below 14.0: https://twistedmatrix.com/trac/ticket/8224 because 14.0 is first version with "decent" TLS support. 

So maybe it would make sense to bump version to 14?

kmike:
I'm not sure proper TLS support is that important for Scrapy.

pawelmhm:
I think I've seen couple of tickets about https bugs and there was this PR here #1794 that apparenly fixes HTTPS if version of twisted is above 14.0

nyov:
I'm currently updating this, diffing `xlib.tx`-code against several newer twisted versions, and stripping it as I go.
So as I figured out now, the `xlib.tx` seems a bit patched here and there, but the latest included code is from twisted version `13.1.0` (which makes sense, since around that time `xlib.tx` was added). Jumping there would basically allow to strip the whole thing.

Since I already diff'ed my way through several releases, I will provide intermediate commits for `11.1`, `12.1`, `12.3` and `13.0` while rolling this up, so we can see how much the `tx` code can shrink each newer twisted release, and which version might be best to fixate on.

Aside: I noticed the local code seems to be always used. The second `if` always matches as well.
I wonder if this was intentional or perhaps for debugging reasons and should have been `elif`?
```py
# scrapy/xlib/tx/__init__.py
if twisted_version > (13, 0, 0): 
    from twisted.web import client
    from twisted.internet import endpoints
    print ('tx proper')
if twisted_version >= (11, 1, 0): 
    from . import client, endpoints
    print ('overridden by xlib.tx')
else:
```
If I'm not mistaken, we are currently shadowing any newer twisted changes there (post-`13.1` release), and, worst-case, may perhaps see some new issues pop up once this is no longer true.

kmike:
@nyov I like this approach. 
And wow, it seems you've found a pretty serious issue in scrapy/xlib/tx.__init__.py

dangra:
Oh my, that is on me. https://github.com/scrapy/scrapy/commit/bb806fa0f8f77cba59506153e33cab96f91008ab#diff-e487b2562862226386f811e0659c70c0R6

dangra:
Hi guys, please take a look to https://github.com/scrapy/scrapy/pull/2403. It fix the issue described by @nyov in https://github.com/scrapy/scrapy/pull/1887#issuecomment-260047541

nyov:
Well, this looked easy in concept...
There seem to be quite a few ways to skin this cat, however.
I've been conservative I think, we could remove yet more code, which should not really be used by anyone directly but only from inside `xlib.tx` (`interfaces.py`, `iweb.py`, `_newclient.py`).

Now this may be a nightmare to review. I'm not sure after all this work whether it's actually comprehensible or useful.
So in theory there are three branches here:

```
      xlib.tx code annotated (easier diff against twisted)
      |   Code removed (easier diff against each other)
          |   Plus scrapy modifications outside xlib ("final change")
              |
              
              o
             /|
            / |
           /  |
          x   o
         /|  /|
        / | / |
       /  |/  |
 tx13 *   x   o
      |  /|  /|
      | / | / |
      |/  |/  |
 tx12 *   x   o [Import ...] commits
      |  /|  /
      | / | /
      |/  |/
 tx11 *   x [Remove ...] commits
      |  /
      | /
      |/
 tx10 * [Bump ...] commits
      |

```

Findings:
Twisted 11.1: all Exceptions can be imported from twisted.
Differences between Twisted 11.1 and 12.0.0 (debian wheezy/oldstable) for this code are  nonexisting. Bumping to 11.1 would have the same result, and be enough.
Twisted 12.1: `TCP4ClientEndpoint` can be imported from twisted.
Twisted 12.3: `Agent` and `ProxyAgent` can be imported from twisted.
Twisted 13.1: `HTTPConnectionPool` can be imported from twisted. Which is everything.

As @dangra said, wheezy-backports includes Twisted 13.2 however, so the next obvious choice would be to go for 13.1 directly I guess, and should be safe enough, if we even care about debian here.
Current debian stable ("jessie") is at Twisted 14.0 and has been around for some time. (With next stable "stretch" freezing in Feb. 17, perhaps releasing around June.)


(Aside: I had to leave this for a while, it took me 5 rebases at least through the patch stack to bring the patches in line; but every other time I edited one commit and rebased the rest back on top, git would wrongly consider some things newer and merge it into the later commits -- breaking the rebased commits. So lots of re-checking against twisted codebase and re-re-fixing. Infuriating.
Finally just had to reaquaint myself with quilt and get to know StGit, to finish this. Both seem worth knowing.

And I felt like a bomb expert while doing this, tracing all the imports that wanted to trip me up. Good thing I'm not, and have unlimited tries.)

kmike:
@nyov thanks for your work! 

Since Scrapy 1.2 the baseline is Debian Jessie, so it is fine to bump Twisted to 14.0. 
jessie-backports contain Twisted 16.2.0, but I'm not sure what's our policy regarding backports.

nyov:
Yeah I haven't looked what differences there are between 13.1 and 14.0, but from the `xlib.tx` viewpoint, anything newer than 13.1 will be sufficient to obsolete it.
Any newer version may of course improve on the situation, such as TLS support.

dangra:
The goal is to remove `xlib.tx`, setting the bar at Twisted 13.1 seems enough to me. We don't have to support this minimum twisted version, it will work in best effort manner. As a bonus existent users of systems with Ubuntu 14.04 (Twisted 13.2) remains happy.

nyov:
That's actually a very astute observation as well :)

Now I have gotten confused here and pushed this non-mergeable tree to this PR.
So apologies for another force-push.
I have renamed my experiments to [`xlibtx`](https://github.com/nyov/scrapy/tree/xlibtx) for who may care,
and pushed a mergeable solution back to here.

Let me know how much you want this squashed, and if the final commit that removes the `xlib/tx` folder is too early at this time (perhaps some code still imports something directly?)

redapple:
LGTM.
@dangra , @kmike , @eliasdorneles ?
This supercedes #2403 I believe, for master branch at least

kmike:
LGTM. I'll let someone else to merge it.

redapple:
Thanks again @nyov !

PR #1891, opened 2016-03-29
[MRG+1] Change Files/ImagesPipelines class attributes to instance attributes

### LINKED ISSUE #1850: EXPIRES in Files/ImagesPipeline should be a instance attribute and not a class attribute
Currently `EXPIRES` in `FilesPipeline` and `ImagesPipeline` is defined as a class attribute. When running more than one spider with `CrawlerProcess` it is impossible to have different expires values for different spiders. Changing the `EXPIRES` attribute to an instance attribute solves the problem.

(I would do a pull request, but I have no idea how to properly update the tests.)

djunzu:
fix #1850
Addressed issues:

1. Move default settings from `FilesPipelines.py` and `ImagesPipelines.py` to `settings/default_settings.py`
2. Change class attributes from `FilesPipelines.py` and `ImagesPipelines.py` to instance attributes. It allows that distinct spiders running together have distinct custom settings for the pipelines.

Note: I have no way to test AWS related stuff. So I haven't touched any code related to AWS.

@kmike as you have looked at the issue before, I think you should review the PR.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -213,14 +213,11 @@ class FilesPipeline(MediaPipeline):
     """
 
     MEDIA_NAME = "file"
-    EXPIRES = 90
     STORE_SCHEMES = {
         '': FSFilesStore,
         'file': FSFilesStore,
         's3': S3FilesStore,
     }
-    DEFAULT_FILES_URLS_FIELD = 'file_urls'
-    DEFAULT_FILES_RESULT_FIELD = 'files'
 
     def __init__(self, store_uri, download_func=None):
         if not store_uri:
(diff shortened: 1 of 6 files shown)


redapple:
@djunzu , would you mind rebasing against current "master" branch?
The Travis CI build failure for docs is now [fixed](https://github.com/scrapy/scrapy/pull/1893).

djunzu:
Rebased.

I realized the changes were not backward compatible, so I rewrote it. Now it should be 100% backward compatible.

Note: If I am not wrong, some parts of the both pipelines were not covered by tests before. The only change I made in those parts was update the class attributes to a instance attribute. As there were no tests for those parts before (if I am not wrong), there are still no tests for it.

scrapy/pipelines/files.py
```diff
@@ -213,19 +214,24 @@ class FilesPipeline(MediaPipeline):
...
     STORE_SCHEMES = {
         '': FSFilesStore,
         'file': FSFilesStore,
         's3': S3FilesStore,
     }
-    DEFAULT_FILES_URLS_FIELD = 'file_urls'
-    DEFAULT_FILES_RESULT_FIELD = 'files'
 
-    def __init__(self, store_uri, download_func=None):
+    def __init__(self, store_uri, settings=None, download_func=None):
```
kmike: Could you please move `settings` argument to the end to make it backwards compatible for users who user positional arguments to create FilesPipeline instance?

kmike:
@djunzu awesome, thanks for the PR! 
Could you please add docs for new options?

scrapy/pipelines/files.py
djunzu: And I thought it was backwards compatible...
I will fix it.

djunzu:
@kmike , there are no new options. They were all available before!
But indeed some of them are not documented. Is it enough document them [here](http://doc.scrapy.org/en/master/topics/media-pipeline.html)?

kmike:
@djunzu ouch, a good call! Some options are still not documented (see http://doc.scrapy.org/en/master/topics/settings.html#settings-documented-elsewhere - not all IMAGE/FILES options have links). But that's not directly related to this PR. If you can fix that it'd be great, but a PR looks good as-is :+1:

djunzu:
If we are doing it, let's do it right!

kmike:
Awesome, thanks @djunzu!

PR #1893, opened 2016-03-30
Add support for Sphinx 1.4

redapple:
See http://www.sphinx-doc.org/en/stable/changes.html#release-1-4-released-mar-28-2016

* `sphinx_rtd_theme` has become optional, needs to be added to reqs
* https://github.com/sphinx-doc/sphinx/pull/2320 changes node entries tuples to 5 values instead of 4
* `sh` syntax highlighting added very locally in selectors.rst because of this warning/error with Sphinx 1.4:

```
Warning, treated as error:
/home/paul/src/scrapy/docs/topics/selectors.rst:743:
WARNING: Could not lex literal_block as "python". Highlighting skipped.
```

### DIFF
--- a/docs/_ext/scrapydocs.py
+++ b/docs/_ext/scrapydocs.py
@@ -18,7 +18,7 @@ def is_setting_index(node):
     if node.tagname == 'index':
         # index entries for setting directives look like:
         # [(u'pair', u'SETTING_NAME; setting', u'std:setting-SETTING_NAME', '')]
-        entry_type, info, refid, _ = node['entries'][0]
+        entry_type, info, refid = node['entries'][0][:3]
         return entry_type == 'pair' and info.endswith('; setting')
     return False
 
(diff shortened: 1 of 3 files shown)


kmike:
Hi Paul,

This looks good, but I think we should specify Sphinx 1.4+ dependency somewhere, it looks like some of the changes won't work with earlier Sphinx versions.

redapple:
@kmike , which change you mean?
I tried with Sphinx 1.3.6 (forcing <1.4 in tox.ini) and it worked for me.
i'm ok with `Sphinx>=1.3.6`

kmike:
ah, yeah, sorry, I thought "changes node entries tuples to 5 values instead of 4" is backwards incompatible, but your code handles both 4 and 5.

nyov:
Thanks, @redapple !

harshasrinivas:
@redapple , since `sphinx_rtd_theme` has to be installed manually for building the docs, right now, it has only been added to the `tox.ini` file. So wouldn't it be better to add it to the `requirements.txt` as well? Or it could be added to [docs/README.rst](https://github.com/scrapy/scrapy/blob/master/docs/README.rst), something like
```
pip install 'Sphinx >= 1.3' sphinx_rtd_theme
```
What do you think?

kmike:
@harshasrinivas I think we shouldn't add sphinx to requirements.txt - it is a file with packages needed to run Scrapy; users generally don't build Scrapy docs, so sphinx is not needed for them.

+1 to adding it to `docs/README.rst`; current instructions don't look complete, a good catch.

PR #1905, opened 2016-04-06
[MRG+1] Modified read failure recovery in utils/gz.py to read only the last f.extrasize bytes of f.extrabuf[ ]

### LINKED ISSUE #1606: response.body is duplicate
Access the [text page(not mine)](http://files.qidian.com/Author4/3615059/88542882.txt) by browsers or wget and you will find  the response content is not duplicate, but scrapy's ```response.body``` is duplicate. I had tried set the scrapy's headers same as a real brower's, but it is still duplicate.

Just use the follow sample code, and you will find the issue.
```
scrapy shell "http://files.qidian.com/Author4/3615059/88542882.txt"
```

Sorry for my bad english.

rootAvish:
This fixes #1606. Building off the remarks of @jdmaeyer in that bug report, the major difference between downloading the file from this server and his pastebin link, is that this server is a Microsoft IIS instance. A look  and the http request that was being sent

```python
In [4]: print(request.headers)
{'Accept-Language': ['en'], 'Accept-Encoding': ['gzip,deflate'], 'Accept': ['text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'], 'User-Agent': ['Scrapy/1.2.0dev2 (+http://scrapy.org)']}
```
led me to `httpcompression` since we are accepted `gzip` encoded data.

I manually saved the response to a file and tried decompressing it using `gunzip`, which is also used by `scrapy.utils.gz` for the same purpose.

```bash
rootavish@rootavish:~/scrapy$ gunzip --decompress dbg.gz 

gzip: dbg.gz: unexpected end of file
```

However, the same file was handled by KDE's Ark without any issues, so I looked into the code in `scrapy.utils.gz` for problems. Changing the chunk size of the read to something smaller than the file size seemed to fix it.

```python
 chunk = read1(f, 8196) # changing chunk size to 4096 fixed the problem
 ```

This behaviour is unexpected since the read should ideally only take place till the end of the file, but due to the archive not being in pristine condition(possibly due to the compression taking place on a Windows based IIS instance and decoding on *nix), there was a wrap around read happening. The ideal solution would be to patch the `gzip` code itself, but since that's not an option we can make do with monkeypatching the same.  The idea was to replace `_read_eof` with a `no-op` for partial archives, something that won't work if GzipFile does it's checksum comparison. The problem should be fixed now.

```bash
rootavish@rootavish:~/scrapy$ scrapy shell "http://files.qidian.com/Author4/3615059/88542882.txt"
2016-04-06 08:34:51 [scrapy] INFO: Scrapy 1.2.0dev2 started (bot: scrapybot)
2016-04-06 08:34:51 [scrapy] INFO: Overridden settings: {'LOGSTATS_INTERVAL': 0, 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter'}
2016-04-06 08:34:51 [scrapy] INFO: Enabled extensions:
['scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.corestats.CoreStats']
2016-04-06 08:34:51 [scrapy] INFO: Enabled downloader middlewares:
['scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
 'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
 'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
 'scrapy.downloadermiddlewares.retry.RetryMiddleware',
 'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
 'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
 'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
 'scrapy.downloadermiddlewares.cookies.CookiesMiddleware',
 'scrapy.downloadermiddlewares.chunked.ChunkedTransferMiddleware',
 'scrapy.downloadermiddlewares.stats.DownloaderStats']
2016-04-06 08:34:51 [scrapy] INFO: Enabled spider middlewares:
2016-04-06 08:34:51 [scrapy] INFO: Enabled item pipelines:
[]
2016-04-06 08:34:51 [scrapy] DEBUG: Telnet console listening on 127.0.0.1:6023
2016-04-06 08:34:51 [scrapy] INFO: Spider opened
2016-04-06 08:34:54 [scrapy] DEBUG: Crawled (200) <GET http://files.qidian.com/Author4/3615059/88542882.txt> (referer: None)
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7f7b70ce6b50>
[s]   item       {}
[s]   request    <GET http://files.qidian.com/Author4/3615059/88542882.txt>
[s]   response   <200 http://files.qidian.com/Author4/3615059/88542882.txt>
[s]   settings   <scrapy.settings.Settings object at 0x7f7b70ce69d0>
[s]   spider     <DefaultSpider 'default' at 0x7f7b6febb090>
[s] Useful shortcuts:
[s]   fetch(req_or_url) Fetch request (or URL) and update local objects
In [1]: import requests

In [2]: resp = requests.get("http://files.qidian.com/Author4/3615059/88542882.txt")
2016-04-06 08:35:20 [urllib3.connectionpool] INFO: Starting new HTTP connection (1): files.qidian.com
2016-04-06 08:35:20 [urllib3.connectionpool] DEBUG: Setting read timeout to None
2016-04-06 08:35:20 [urllib3.connectionpool] DEBUG: "GET /Author4/3615059/88542882.txt HTTP/1.1" 200 None

In [3]: response.body_as_unicode() == resp.text
Out[3]: True

In [4]: response.body.count('document')
Out[4]: 1
```


rootAvish:
Oh.. I did not notice initially we have a test here(https://travis-ci.org/scrapy/scrapy/jobs/121056809#L319) to always discard partial archives. I guess the option then would be a fix in the `gzip` code upstream.

jdemaeyer:
Hey @rootAvish,
thanks for the PR!

> Oh.. I did not notice initially we have a test here(https://travis-ci.org/scrapy/scrapy/jobs/121056809#L319) to always discard partial archives.

The problem with that error is not the test itself but the fact that `GzipFile` doesn't have a `_read_eof()` method in Python 3. If you look at https://travis-ci.org/scrapy/scrapy/builds/121056805 you'll see that the tests passed under Python 2.

Your patch looks like you always use the content manager, maybe it's possible to only use it if we detect an incomplete archive? Also, could you add a test? :)

kmike:
A quick note: if there is some issue with GzipFile it makes sense to create a subclass instead of monkey-patching stdlib version.

rootAvish:
@kmike  I incorporated the changes you and @jdemaeyer suggested, and also added a test. Any feedback for code quality enhancements etc. is appreciated, also would you like to me to apply these changes to a fresh clone and open a different pull request with just one-two commit containing the entire patch, since merging this one would add useless commit history to the log?

tests/test_utils_gz.py
```diff
@@ -25,5 +25,10 @@ def test_gunzip_no_gzip_file_raises(self):
 
     def test_gunzip_truncated_short(self):
         with open(join(SAMPLEDIR, 'truncated-crc-error-short.gz'), 'rb') as f:
-            text = gunzip(f.read())
+            text = gunzip(f.read())
```
jdemaeyer: Seems like an extra space accidentally sneaked in here

tests/test_utils_gz.py
```diff
@@ -25,5 +25,10 @@ def test_gunzip_no_gzip_file_raises(self):
...
     def test_gunzip_truncated_short(self):
         with open(join(SAMPLEDIR, 'truncated-crc-error-short.gz'), 'rb') as f:
-            text = gunzip(f.read())
+            text = gunzip(f.read()) 
             assert text.endswith(b'</html>')
+
+    def test_gunzip_illegal_eof(self):
+        with open(join(SAMPLEDIR, 'unexpected-eof.gz'), 'rb') as f:
+            text = gunzip(f.read())
+            self.assertEqual(text.count(b'document'), 1)
```
jdemaeyer: This test might be a little vague. If you can, could you try creating a new gzipped file with illegal ending, and then compare the full expected text? I don't know how hard that is though, so if it takes you more than say 15 minutes just forget about it. ;)

jdemaeyer: Another nitpick also ;) We try to follow PEP8 in Scrapy's codebase, please end the file (meaning `test_utils_gz.py` here) in a newline (i.e. the file ends with `\n`). Note that this does not mean that there should be an empty line (which would be `\n\n`). You can probably configure your editor to do it automatically for you (iirc vim does it by default).

scrapy/utils/gz.py
```diff
@@ -9,32 +9,58 @@
...
+            # would make a partial archive read work
+            try:
+                if six.PY2:
+                    self._read_eof = lambda *args, **kwargs: None
+                    return self.read(size)
+                else:
+                    self._check_closed = lambda *args, **kwargs: None
+                    return self.read1(size)
+            # Raise any exceptions if we still get them
+            except (IOError, EOFError, struct.error):
```
jdemaeyer: This `except` clause doesn't really do anything, other than changing the traceback (you're just catching the exception and then reraising it). It should be okay to remove the inner `try`/`except`. If the read fails even though we're ignoring the illegal EOF, there's something more severe wrong with the file, and errors should bubble up to the caller.

scrapy/utils/gz.py
```diff
@@ -9,32 +9,58 @@
...
+            if six.PY2:
+                return self.read(size)
+            else:
+                return self.read1(size)
+        except(IOError):
+            # replace the _read_eof method with a no-op and try the read again,
+            # would make a partial archive read work
+            try:
+                if six.PY2:
+                    self._read_eof = lambda *args, **kwargs: None
```
jdemaeyer: @kmike I like this approach of patching the class method here, instead of overwriting it per se. It is not gonna be used again (on this instance) anyways. You think it's okay?

jdemaeyer:
Hey @rootAvish,

I've added a few more line comments. :)

> would you like to me to apply these changes to a fresh clone and open a different pull request with just one-two commit containing the entire patch, since merging this one would add useless commit history to the log?

You can prune the commit history by [squashing your commits](https://davidwalsh.name/squash-commits-git), then force-pushing, no need to create a separate PR


Cheers,
-Jakob

tests/test_utils_gz.py
rootAvish: I did create a file in the sense that I saved the gzipped response sent by the server to a file, however I do agree the test in itself isn't solid enough. Maybe I can fetch the un-compressed text using the `requests` library, save it to a file in `SAMPLEDIR` and compare the body of that to the output of gunzip?

rootAvish:
Thanks for the review, I'll do the changes and ping back once they're done sometime today.

jdemaeyer:
Hey @rootAvish,

thanks for all the changes and again for investigating this. I've spent a little time playing around with the test you wrote, and I found that the wrap-around read comes from our existing [read failure recovery mechanism](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/gz.py#L46). It looks like what happens is this:

* The first chunk is read by `read1()`, and we add it to `output`
* On the second call to `read1()`, we get an `IOError`, and go into our `except` clause
* We add all the contents of `f.extrabuf` to our `output`. *However*, for some reason `GzipFile` has not yet updated `extrabuf` and it still contains the chunk that we already read. Hence, the first 8196 are replicated in our output (the "wrap-around read").

I found that, while `f.extrabuf` was not yet updated, `f.extrasize` (a variable presumably holding the size of the extra buffer) was already updated, and the tests you created pass when I replace the line

```python
output += f.extrabuf
```

with

```python
output += f.extrabuf[-f.extrasize:]
```

(reading only the end of the extra buffer).

Maybe this could be a simpler fix?

rootAvish:
@jdemaeyer, sorry for responding so late, but yeah, this fix is much better, I wasn't aware of the existence of `f.extrasize`, I didn't comment thinking perhaps you'd like to make the fix since you're the one who solved the problem. But I've made the change here too should you want to contain both the test and the fix in the same PR.

tests/test_utils_gz.py
```diff
@@ -27,3 +27,9 @@ def test_gunzip_truncated_short(self):
         with open(join(SAMPLEDIR, 'truncated-crc-error-short.gz'), 'rb') as f:
             text = gunzip(f.read())
             assert text.endswith(b'</html>')
+
+    def test_gunzip_illegal_eof(self):
+        with open(join(SAMPLEDIR, 'unexpected-eof.gz'), 'rb') as f:
+            text = gunzip(f.read())
+            with open(join(SAMPLEDIR, 'unexpected-eof-output.txt'), 'rb') as o:
+                self.assertEqual(len(text), len(o.read().decode("utf-8")))
```
jdemaeyer: Hm, do you know why only the lengths match?

rootAvish: @jdemaeyer , the encodings of both text and the output of o.read() were different, in the new test I committed both are set to unicode, I added an assertion to make sure the content too is same.

jdemaeyer:
Sorry, must've missed your last comment here.

LGTM, could you rebase? @redapple you recently merged changes to `scrapy.utils.gz`, but it seems that they're not related to this 'properly recover from illegal EOF' issue, right? /cc @kmike

rootAvish:
@jdemaeyer I did the rebase, I think we can merge this now.

tests/test_utils_gz.py
```diff
@@ -3,6 +3,9 @@
 
 from scrapy.utils.gz import gunzip, is_gzipped
 from scrapy.http import Response, Headers
+
+from w3lib.encoding import html_to_unicode
```
kmike: Could you please move this import above scrapy imports? This is very minor, but scrapy tries to group imports into 3 groups: stdlib imports, blank line, external libraries, blank line, scrapy imports.

rootAvish: @kmike fixed, thanks for pointing that out.

redapple:
LGTM. Thanks @rootAvish !
The only thing I'm puzzled with is the difference between the short fix (which is always good)
```
-                    output += f.extrabuf
+                    output += f.extrabuf[-f.extrasize:]
```
and the description of the PR about a previous fix.
Maybe it could be updated to reflect the last implementation choice.

rootAvish:
@redapple Hi, sorry I missed that comment earlier. The fix is detailed in @jdemaeyer 's comment above, and the title of the PR has also been changed, do you want me to update the description in the commit message?

redapple:
@rootAvish , I may very well be bikesheeding here, so apologies :)
but I'd prefer a PR title that reflects the change a little better.
The current _Modification to the existing read failure recovery mechanism, fixes #1606: response.body is duplicate_ could be more explicit as to what was changed in gzip read.
Also, _fixes #1606: response.body is duplicate_ can be left in the PR description (where it is already), I believe GitHub will pick it up when this is merged.

rootAvish:
@redapple I see, is it fine now?

redapple:
Thanks @rootAvish

PR #1857, opened 2016-03-12
[MRG+1] response_status_message should not fail on non-standard HTTP codes

pawelmhm:
utility is used in retry middleware and it was failing to handle non-standard HTTP codes. Instead of raising exceptions when passing through `to_native_str` it should return "Unknown status" message.

PR #1874, opened 2016-03-22
Fix and re-enabled canonicalize_url() tests

redapple:
- Supersedes https://github.com/scrapy/scrapy/pull/1803
- + another take at https://github.com/scrapy/scrapy/pull/1664

Note: this PR contains the same `safe_url_string()` implementation as https://github.com/scrapy/w3lib/pull/45 to be able to run tests before w3lib gets updated

PR #1906, opened 2016-04-06
[Backport][1.0] response_status_message should not fail on non-standard HTTP codes

redapple:
Backporting https://github.com/scrapy/scrapy/pull/1857 into 1.0

utility is used in retry middleware and it was failing to handle non-standard HTTP codes.
Instead of raising exceptions when passing through to_native_str it should return
"Unknown status" message.

### DIFF
--- a/scrapy/utils/response.py
+++ b/scrapy/utils/response.py
@@ -55,7 +55,7 @@ def response_status_message(status):
     """
     # Implicit decode/encode is on purpose to force native strings
     # This is properly fixed in Scrapy >=1.1 at revision faf9265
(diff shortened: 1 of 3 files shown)


redapple:
Ouch, w3lib 1.14 is not compatible with 1.0 branch for some URLs. There may be real problematic cases. I'll check.

redapple:
This will probably need backporting https://github.com/scrapy/scrapy/pull/1874

redapple:
https://github.com/scrapy/w3lib/pull/50 should suffice

PR #1909, opened 2016-04-07
[MRG+1] added faq entry regarding ssl errors on osx

kneufeld:
@redapple here's that pull request you... requested.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -280,6 +280,40 @@ I'm scraping a XML document and my XPath selector doesn't return any items
 
 You may need to remove namespaces. See :ref:`removing-namespaces`.
 
+SSL Errors on OS X
+------------------
+
+At the time of this writing (April 2016), the version of OpenSSL that
+ships on OS X El Capitan (10.11.4) is quite old. The system OpenSSL is
+version ``0.9.8zg`` whereas the latest version is ``1.0.2g``.
+
+The current version of Scrapy (1.0.5) also has issues that will
+be addressed in version 1.1, so we'll need to install the release
+candidate.
+
+Assuming you're using `Homebrew <http://brew.sh>`_, install OpenSSL::
+
+    brew update
+    brew install openssl # or brew upgrade openssl
+    # the following step might not be necessary
+    # omit unless ssl errors continue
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -280,6 +280,40 @@ I'm scraping a XML document and my XPath selector doesn't return any items
...
+SSL Errors on OS X
+------------------
+
+At the time of this writing (April 2016), the version of OpenSSL that
+ships on OS X El Capitan (10.11.4) is quite old. The system OpenSSL is
+version ``0.9.8zg`` whereas the latest version is ``1.0.2g``.
+
+The current version of Scrapy (1.0.5) also has issues that will
+be addressed in version 1.1, so we'll need to install the release
+candidate.
```
kmike: I think it doesn't make much sense to document Scrapy 1.0.5 limitations in Scrapy master branch because these docs would be docs for Scrapy 1.1 (or 1.2?)

kmike: Likewise, we shouldn't suggest installing 1.1.0rc3 here because by the time these docs are default 1.1 should be already released.

kneufeld: Well I'd say we're at a bit of a catch-22 then. Since the current version is broken, the next one is not but to publish this guide you need a new version.

Merge it or not or edit it or not.

kmike: There is 1.0 branch; I think a PR which mentions 1.0 limitations should be agains this branch, not against master.

redapple: @kmike , not sure I'm following. This is about limitations because of OS X 's OpenSSL version, that are still there for Scrapy 1.1.
Am I missing something?

kmike: My point is about this note:

> The current version of Scrapy (1.0.5) also has issues that will
be addressed in version 1.1, so we'll need to install the release
candidate.

redapple: Right, this can be removed. and say `pip install Scrapy` instead now.

Gallaecio:
Closing since OS X El Capitan is not supported [since August 2018](https://en.wikipedia.org/wiki/OS_X_El_Capitan).

Issue #1807, opened 2016-02-23
Scrapy keep wrong proxy setting


DrJackilD:
Hi! I have one issue, when trying to test proxies with Scrapy. I want to check proxies with httpbin.org, and make crawler:
```python
class CheckerSpider(scrapy.Spider):
    name = "checker"
    start_urls = (
        'https://www.httpbin.org/ip'
    )
    connection = get_connection()

    def start_requests(self):

        with self.connection.cursor() as cursor:
            limit = int((datetime.now() - datetime(1970, 1, 1)).total_seconds()) - 3600
            q = """ SELECT *
                    FROM {}
                    WHERE active = 1 AND last_checked <= {} OR last_checked IS NULL;""".format(DB_TABLE, limit)
            cursor.execute(q)
            proxy_list = cursor.fetchall()

        for proxy in proxy_list[:15]:
            word = get_random_word()
            req = scrapy.Request(self.start_urls, self.check_proxy, dont_filter=True)
            req.meta['proxy'] = 'https://{}:8080'.format(proxy['ip'])
            req.meta['item'] = proxy
            user_pass = base64.encodestring('{}:{}'.format(PROXY_USER, PROXY_PASSWORD))
            req.headers['Proxy-Authorization'] = 'Basic {}'.format(user_pass)
            req.headers['User-Agent'] = get_user_agent()
            yield req

    def check_proxy(self, response):
        print response.request.meta['proxy']
        print response.meta['item']['ip']
        print response.body
```

But when I'm testing it, I've see that Scrapy connect to url only with 5 proxies and then didn't change it. Example output (just messed IP):

```python
2016-02-23 14:54:36 [scrapy] DEBUG: Crawled (200) <GET https://www.httpbin.org/ip> (referer: None)
https://192.168.100.130:8080
192.168.100.130
{
  "origin": "192.168.100.130"
}

2016-02-23 14:54:36 [scrapy] DEBUG: Crawled (200) <GET https://www.httpbin.org/ip> (referer: None)
https://192.168.100.131:8080
192.168.100.131
{
  "origin": "192.168.100.131"
}
2016-02-23 14:54:37 [scrapy] DEBUG: Crawled (200) <GET https://www.httpbin.org/ip> (referer: None)
https://192.168.100.132:8080
192.168.100.132
{
  "origin": "192.168.100.132"
}

# Here Scrapy used wrong proxy to connect to site.
[...]

PR #1912, opened 2016-04-08
[MRG+1] Fix HTTP Pool key for HTTPS proxy tunneled connections (CONNECT method)

redapple:
Should fix https://github.com/scrapy/scrapy/issues/1807

It adds the proxy host and port to the key used to cache HTTP connections, instead of just the remote host and port.

I have no concrete proposal for unittesting this though.
I only was able to debug-print HTTP pool keys and testing with https://httpbin.org/ip with a few open HTTPS proxies.
With current code, before this patch, connections do get mixed with current default Twisted `Agent` [`key = (parsedURI.scheme, parsedURI.host, parsedURI.port)`](http://twistedmatrix.com/trac/browser/tags/releases/twisted-16.0.0/twisted/web/client.py#L1597)


kmike:
The change looks good to me, but I haven't tested it on real websites.

PR #1923, opened 2016-04-11
[MRG+1] Use newer w3lib.url.safe_url_string() and re-enable HTTP request tests

redapple:
Starting from w3lib 1.14.1, `safe_url_string` implementation follows browsers' behavior better.
Let's use it from now on.

These changes were originally part of https://github.com/scrapy/scrapy/pull/1874, split here so as to seperate topics.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Twisted>=10.0.0
 lxml
 pyOpenSSL
 cssselect>=0.9
-w3lib>=1.14.1
+w3lib>=1.14.2
 queuelib
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Twisted>=10.0.0
 lxml
 pyOpenSSL
 cssselect>=0.9
-w3lib>=1.13.0
+w3lib>=1.14.1
 queuelib
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -8,7 +8,7 @@ import six
 from w3lib.url import safe_url_string
 
 from scrapy.http.headers import Headers
-from scrapy.utils.python import to_native_str, to_bytes
+from scrapy.utils.python import to_bytes
 from scrapy.utils.trackref import object_ref
 from scrapy.utils.url import escape_ajax
 from scrapy.http.common import obsolete_setter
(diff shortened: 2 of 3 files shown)


scrapy/http/request/__init__.py
```diff
@@ -50,8 +50,8 @@ def _set_url(self, url):
         if not isinstance(url, six.string_types):
             raise TypeError('Request url must be str or unicode, got %s:' % type(url).__name__)
 
-        url = to_native_str(url, self.encoding)
-        self._url = escape_ajax(safe_url_string(url))
+        s = safe_url_string(url, self.encoding)
```
kmike: Does it mean users now must use `yield scrapy.Request(url, ..., encoding=response.encoding)` in order to get correct URL escaping?

kmike: Should we write code this way in Scrapy itself, e.g. in FormRequest or in CrawlSpider?

redapple: @kmike , yes, if they want to use correct URL escaping and if it's not UTF8, then they need pass the information at some point. We should do the work for them with `FormRequest.from_response`, and making the URL safe in `Response.urljoin` (amending this PR or another).
The hard thing here is that correct encoding for query part may depend on form information within the document, document that may not have been parsed for this.
I believe for `CrawlSpider`, getting encoding right can be achieved with `encoding` information being kept with `Link` until `Request` is created from it.

redapple: [`FormRequest.from_response` uses response.encoding](https://github.com/scrapy/scrapy/blob/ebef6d7c6dd8922210db8a4a44f48fe27ee0cd16/scrapy/http/request/form.py#L39) already, that's good.

redapple: Passing the correct encoding is not new with this change though, Users had to do it also in the past. Not doing it for non-UTF8 did produce wrong URL escaping.

requirements.txt
```diff
@@ -2,7 +2,7 @@ Twisted>=10.0.0
 lxml
 pyOpenSSL
 cssselect>=0.9
-w3lib>=1.13.0
+w3lib>=1.14.1
```
kmike: let's bump it to 1.14.2

scrapy/http/request/__init__.py
kmike: Aha, so let's say in Python 2 a unicode URL is extracted from cp1251 page and the encoding is not passed. Before the change:
 
* url is encoded to utf8,
* then escaped. It means path is escaped correctly, but query arguments and fragment are not.

Now:
* path is escaped correctly, query and fragments are not.

The same, but with bytes URL. Before:

* url is passed to safe_url_string as-is; it means path is escaped incorrectly, but query and fragment are escaped correctly.

After:

* safe_url_string tries to convert URL to unicode. Sometimes it fails; w3lib escapes failing parts. For escaped parts path is encoded incorrectly, but query and fragment are encoded properly. For unescaped parts the whole URL is escaped incorrectly because unicode value is decoded from a wrong encoding.

So you're right, either way something is not correct if encoding is not passed. It is a bit worse IMHO after the change for binary URLs (less predictable), but either way it is broken, so the behavior change looks good to me.

Issue #1936, opened 2016-04-18
SSL errors crawling https sites using proxies


lhuaizhong:
I'm unable to scrape below https sites through https supported proxies, have tried version 1.0.5 and 1.1.0rc3. I'm using proxymesh, but https://www.python.org and others work well.

(env-scrapy-1.1) D:\work>scrapy version -v
Scrapy    : 1.1.0rc3
lxml      : 3.6.0.0
libxml2   : 2.9.0
Twisted   : 16.1.1
Python    : 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)]
pyOpenSSL : 16.0.0 (OpenSSL 1.0.2g  1 Mar 2016)
Platform  : Windows-10-10.0.10586

`2016-04-18 12:56:25 [scrapy] DEBUG: Gave up retrying <GET https://xxx.xxx.xxx/> (failed 3 times): [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'SSL23_GET_SERVER_HELLO', 'sslv3 alert handshake failure')]>]
2016-04-18 12:56:25 [scrapy] ERROR: Error downloading <GET https://xxx.xxx.xxx/>: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'SSL23_GET_SERVER_HELLO', 'sslv3 alert handshake failure')]>]`

=======================

D:\work>scrapy version -v
Scrapy    : 1.0.5
lxml      : 3.5.0.0
libxml2   : 2.9.3
Twisted   : 15.5.0
Python    : 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)]
pyOpenSSL : 0.15.1 (OpenSSL 1.0.2f  28 Jan 2016)
Platform  : Windows-10-10.0.10586

`2016-04-18 12:58:29 [scrapy] ERROR: Error downloading <GET https://xxx.xxx.xxx/>: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'SSL23_GET_SERVER_HELLO', 'sslv3 alert handshake failure')]>]`

The https://github.com/scrapy/scrapy/issues/1429#issuecomment-131782133 doesn't work as well.

May anyone help? thanks a lot in advance.

redapple:
[...]

PR #1938, opened 2016-04-18
Set SNI properly when using CONNECT

redapple:
(maybe) fixing https://github.com/scrapy/scrapy/issues/1936

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -122,9 +122,14 @@ class TunnelingTCP4ClientEndpoint(TCP4ClientEndpoint):
         """
         self._protocol.dataReceived = self._protocolDataReceived
         if  TunnelingTCP4ClientEndpoint._responseMatcher.match(bytes):
-            # this set proper Server Name Indication extension
-            sslOptions = self._contextFactory.creatorForNetloc(
-                self._tunneledHost, self._tunneledPort)
+            try:
+                # this sets proper Server Name Indication extension
+                # but is only available for Twisted>=14.0
+                sslOptions = self._contextFactory.creatorForNetloc(
+                    self._tunneledHost, self._tunneledPort)
+            except AttributeError:
+                # fall back to non-SNI SSL context factory
+                sslOptions = self._contextFactory
             self._protocol.transport.startTLS(sslOptions,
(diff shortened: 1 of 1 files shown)


PoulTur:
Hi Paul, you asked me to contribute to #1938 but actually I do not use ProxyMesh or Scrapy for past few months, I'm sorry I will not be able to help here.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -122,7 +122,15 @@ def processProxyResponse(self, bytes):
...
         if  TunnelingTCP4ClientEndpoint._responseMatcher.match(bytes):
-            self._protocol.transport.startTLS(self._contextFactory,
+            try:
+                # this sets proper Server Name Indication extension
+                # but is only available for Twisted>=14.0
+                sslOptions = self._contextFactory.creatorForNetloc(
+                    self._tunneledHost, self._tunneledPort)
+            except AttributeError:
+                # fall back to non-SNI SSL context factory
+                sslOptions = self._contextFactory
```
kmike: what do you think about naming the variable `connectionCreator` instead of `sslOptions`?

kmike: Other than that, the PR looks good to me.

redapple: @kmike , I don't like `connectionCreator` better tbh, sounds too generic, not related to securing the already opened transport link.
Even if [the interface says](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.ITLSTransport.startTLS.html) `IOpenSSLClientConnectionCreator` (in our case), it also says:

> For clients, use `twisted.internet.ssl.optionsForClientTLS`;
> for servers, use `twisted.internet.ssl.CertificateOptions`.

Twisted's examples [also use "options"](https://twistedmatrix.com/documents/current/core/howto/ssl.html#using-starttls) in some way.

It could be `tlsOptions` though.
Or `tlsContextFactory` to keep the "factory" idiom.

kmike: Ok :) It looks strange to my eyes that sslOptions is not some dict-like or settings-like object, but a factory, but if that's a common convention let's keep it.

Issue #1941, opened 2016-04-20
canonicalize_url in linkextractor: not what browsers do


Digenis:
By default, the link extractor calls canonicalize_url on the collected links.
The following is not what browsers do:

``` python
canonicalize_url('http://example.com/index.php?/a/=/o/')
'http://example.com/index.php?%2Fa%2F=%2Fo%2F'  # encoding forward slashes
canonicalize_url('http://example.com/index.php?a')
'http://example.com/index.php?a='  # appending = on empty arguments
```

I doubt this is a problem in canonicalize_url
because it's not meant to mimic browsers in the first place, is it?

However this is a problem for the link extractor
because it can potentially end up extracting urls
that are wrong from the server's perspective.
In this example, the server doesn't recognise the extractor's url, only the browser's:

``` python
# http://forum.laptop.bg/index.php?/discover/
LinkExtractor(restrict_xpaths=('//a[contains(@href, "/topic")]',)).extract_links(response)[0].url
# Extractor: http://forum.laptop.bg/index.php?%2Ftopic%2F57339-%D0%BB%D0%B0%D0%BF%D1%82%D0%BE%D0%BF-asus-w90vp%2F=&comment=221153&do=findComment
# Browser:   http://forum.laptop.bg/index.php?/topic/57339-%D0%BB%D0%B0%D0%BF%D1%82%D0%BE%D0%BF-asus-w90vp/&do=findComment&comment=221153
```

Was this a design decision or a bug?

kmike:
@Digenis +1 to not using canonicalize_url by default in link extractors. It causes other issues - see e.g. https://github.com/scrapy/scrapy/issues/1202. 

See also: https://github.com/scrapy/w3lib/pull/25#issuecomment-205354098.

kmike:
This change would be backwards incompatible, and link extractors beg for a rewrite. I think it should be fixed as a part of link extractors rewrite. We shouldn't alter URLs by default; IMHO canonicalize_url is for duplication filters, not for preprocessing URLs before sending them to a server.

redapple:
FWIW, regarding percent-escaping of `/`,
 for a link as

```
<a href="http://localhost:8001/query?a=/&b=?&c=@&d=:">/query?a=/&b=?&c=@&d=:</a>
```
this is what Chrome (Version 50.0.2661.75 (64-bit)) and Firefox (45.0.2) request (as received by HTTP server)
```
------------------------------------------------------------------------------------------------------------------------
Chrome
[...]

PR #1947, opened 2016-04-21
[MRG+1] Fix canonicalize_url() on Python 3 and re-enable tests

redapple:
Bits of https://github.com/scrapy/scrapy/pull/1874 related to canonicalize_url()

Does NOT handle issues raised in https://github.com/scrapy/scrapy/issues/1941

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -83,17 +83,16 @@ def canonicalize_url(url, keep_blank_values=True, keep_fragments=False,
     try:
         scheme, netloc, path, params, query, fragment = _safe_ParseResult(
             parse_url(url), encoding=encoding)
-    except UnicodeError as e:
-        if encoding != 'utf8':
-            scheme, netloc, path, params, query, fragment = _safe_ParseResult(
-                parse_url(url), encoding='utf8')
-        else:
-            raise
+    except UnicodeEncodeError as e:
+        scheme, netloc, path, params, query, fragment = _safe_ParseResult(
+            parse_url(url), encoding='utf8')
 
     # 1. decode query-string as UTF-8 (or keep raw bytes),
     #    sort values,
     #    and percent-encode them back
-    if not six.PY2:
+    if six.PY2:
+        keyvals = parse_qsl(query, keep_blank_values)
+    else:
         # Python3's urllib.parse.parse_qsl does not work as wanted
         # for percent-encoded characters that do not match passed encoding,
         # they get lost.
@@ -118,8 +117,6 @@ def canonicalize_url(url, keep_blank_values=True, keep_fragments=False,
         # IRIs (namely, to be able to include non-ASCII characters) can only be
         # used if the query part is encoded in UTF-8.
         keyvals = parse_qsl_to_bytes(query, keep_blank_values)
-    else:
-        keyvals = parse_qsl(query, keep_blank_values)
     keyvals.sort()
     query = urlencode(keyvals)
 
@@ -138,16 +135,17 @@ def _unquotepath(path):
     for reserved in ('2f', '2F', '3f', '3F'):
         path = path.replace('%' + reserved, '%25' + reserved.upper())
 
-    if six.PY3:
-        # standard lib's unquote() does not work in Python 3
-        # for non-UTF-8 percent-escaped characters, they get lost.
+    if six.PY2:
+        # in Python 2, '%a3' becomes '\xa3', which is what we want
+        return unquote(path)
+    else:
+        # in Python 3,
+        # standard lib's unquote() does not work for non-UTF-8
+        # percent-escaped characters, they get lost.
         # e.g., '%a3' becomes 'REPLACEMENT CHARACTER' (U+FFFD)
         #
         # unquote_to_bytes() returns raw bytes instead
         return unquote_to_bytes(path)
-    else:
-        # in Python 2, '%a3' becomes '\xa3', which is what we want
-        return unquote(path)
 
 
 def parse_url(url, encoding=None):
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -7,15 +7,18 @@ to the w3lib.url module. Always import those from there instead.
 """
 import posixpath
 import re
+import six
 from six.moves.urllib.parse import (ParseResult, urlunparse, urldefrag,
                                     urlparse, parse_qsl, urlencode,
-                                    unquote)
+                                    quote, unquote)
+if six.PY3:
+    from urllib.parse import unquote_to_bytes
 
 # scrapy.utils.url was moved to w3lib.url and import * ensures this
 # move doesn't break old code
 from w3lib.url import *
 from w3lib.url import _safe_chars
-from scrapy.utils.python import to_native_str
+from scrapy.utils.python import to_bytes, to_native_str, to_unicode
 
 
 def url_is_from_any_domain(url, domains):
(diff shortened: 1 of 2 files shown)


scrapy/utils/url.py
```diff
@@ -37,42 +40,114 @@ def url_has_any_extension(url, extensions):
...
+    # If supplied `encoding` is not compatible with all characters in `url`,
+    # fallback to UTF-8 as safety net.
+    # UTF-8 can handle all Unicode characters,
+    # so we should be covered regarding URL normalization,
+    # if not for proper URL expected by remote website.
+    try:
+        scheme, netloc, path, params, query, fragment = _safe_ParseResult(
+            parse_url(url), encoding=encoding)
+    except UnicodeError as e:
+        if encoding != 'utf8':
```
kmike: there are several other spellings for utf8: 

* utf_8;
* utf-8;
* U8;
* UTF;

these names are also case-insensitive. See https://docs.python.org/3.5/library/codecs.html#standard-encodings.

scrapy/utils/url.py
```diff
@@ -37,42 +40,114 @@ def url_has_any_extension(url, extensions):
...
+    # every part should be safe already
     return urlunparse((scheme, netloc.lower(), path, params, query, fragment))
 
 
 def _unquotepath(path):
     for reserved in ('2f', '2F', '3f', '3F'):
         path = path.replace('%' + reserved, '%25' + reserved.upper())
-    return unquote(path)
+
+    if six.PY3:
```
kmike: it is slightly better to use six.PY2 because six.PY3 may become False in Python 4

scrapy/utils/url.py
```diff
@@ -37,42 +40,114 @@ def url_has_any_extension(url, extensions):
...
+    # so we should be covered regarding URL normalization,
+    # if not for proper URL expected by remote website.
+    try:
+        scheme, netloc, path, params, query, fragment = _safe_ParseResult(
+            parse_url(url), encoding=encoding)
+    except UnicodeError as e:
+        if encoding != 'utf8':
+            scheme, netloc, path, params, query, fragment = _safe_ParseResult(
+                parse_url(url), encoding='utf8')
+        else:
```
kmike: When can it fail?

redapple: it fails for example for `test_canonicalize_url_unicode_query_string_wrong_encoding`
```
self = <tests.test_utils_url.CanonicalizeUrlTest testMethod=test_canonicalize_url_unicode_query_string_wrong_encoding>

    def test_canonicalize_url_unicode_query_string_wrong_encoding(self):
        # trying to encode with wrong encoding
        # fallback to UTF-8
>       self.assertEqual(canonicalize_url(u"http://www.example.com/résumé?currency=€", encoding='latin1'),
                                          "http://www.example.com/r%C3%A9sum%C3%A9?currency=%E2%82%AC")

/home/paul/src/scrapy/tests/test_utils_url.py:147: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/paul/src/scrapy/scrapy/utils/url.py:85: in canonicalize_url
    parse_url(url), encoding=encoding)
/home/paul/src/scrapy/scrapy/utils/url.py:54: in _safe_ParseResult
    quote(to_bytes(parts.query, encoding), _safe_chars),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

text = 'currency=€', encoding = 'latin1', errors = 'strict'

    def to_bytes(text, encoding=None, errors='strict'):
        """Return the binary representation of `text`. If `text`
        is already a bytes object, return it as-is."""
        if isinstance(text, bytes):
            return text
        if not isinstance(text, six.string_types):
            raise TypeError('to_bytes must receive a unicode, str or bytes '
                            'object, got %s' % type(text).__name__)
        if encoding is None:
            encoding = 'utf-8'
>       return text.encode(encoding, errors)
E       UnicodeEncodeError: 'latin-1' codec can't encode character '\u20ac' in position 9: ordinal not in range(256)

/home/paul/src/scrapy/scrapy/utils/python.py:120: UnicodeEncodeError
```

scrapy/utils/url.py
redapple: On second thought, the test is not really necessary, I can't think of an example of UTF8 not working, so testing if supplied encoding is not UTF8 does not make much sense, i.e., if it failed, then it must something else, so use UTF8 the 2nd time.

scrapy/utils/url.py
```diff
@@ -7,15 +7,18 @@
 """
 import posixpath
 import re
+import six
 from six.moves.urllib.parse import (ParseResult, urlunparse, urldefrag,
                                     urlparse, parse_qsl, urlencode,
-                                    unquote)
+                                    quote, unquote)
+if six.PY3:
```
redapple: @kmike , probably `not six.PY2` here too, right?

kmike: right

kmike:
Looks good, thanks @redapple for the hard work!

By the way, do you know if this is true? I think it can be a desirable property of canonicalize_url function.

``` 
url = canonicalize_url(url_original, encoding=X)
url2 = canonicalize_url(url, encoding=X)
# url is always the same as url2
```

scrapy/utils/url.py
```diff
@@ -37,42 +40,112 @@ def url_has_any_extension(url, extensions):
...
-    - remove fragments (unless keep_fragments is True)
+    - remove query arguments with blank values (unless `keep_blank_values` is True)
+    - remove fragments (unless `keep_fragments` is True)
 
-    The url passed can be a str or unicode, while the url returned is always a
-    str.
+    The url passed can be bytes or unicode, while the url returned is
+    always a native str (bytes in Python 2, unicode in Python 3).
 
     For examples see the tests in tests/test_utils_url.py
```
kmike: I wonder if we should add a note about what is `canonicalize_url` function for: it is not for cleaning up URLs before sending them to a server, it is for URL comparison and duplicate detection, right?

redapple: Would not hurt to remind users of this, yes.
A bit of a shame that it's used by default in link extraction

redapple:
@kmike , [`safe_url_string` is tested for this](https://github.com/scrapy/w3lib/blob/fddf599cda53adecc27da0b9b246186b3444be72/tests/test_url.py#L163) so worth a try with canonicalize_url too... and a test

kmike:
@redapple do you want to make these changes in this PR, or should we just merge it as-is?

redapple:
would be good in this one. but it'll have to be tomorrow at earliest for me.
Or someone else pushes (tries) new tests

PR #1880, opened 2016-03-25
Add encoding to Link object + re-enable HTML entities links tests

redapple:
Built on top of https://github.com/scrapy/scrapy/pull/1874

Should fix #1403.

I had to add a `encoding` attribute for the `Link` objects to build proper URLs.

[`url = urljoin(response_url, url)`](https://github.com/scrapy/scrapy/blob/master/scrapy/linkextractors/lxmlhtml.py#L61) after processing the base_URL-joined attribute looks wrong to me. I'm not even sure we should pass attributes to `process_value` after this baseURL joining.

Issue #833, opened 2014-08-01
Prevent URL encoding option


DanMcInerney:
I am having the damndest time trying to unquote the URL in some requests. Any plans to add that an as option? It seems like I have to monkeypatch to fix it as middleware won't work but monkeypatching it involves going like 3 libraries deep and I'm not sure how to do that.

Skyross:
I just used code below but prepared properly encoded `url` before CustomRequest call.

```python
class CustomRequest(Request):
    def _set_url(self, url):
        self._url = str(url)
```

moonsdarkside:
I'm having a similar issue when my sitemap spider takes URLs with special characters from a sitemap.

When I run a straightforward sitemap spider without particular handling of special characters, I get:

`ERROR: Spider error processing <GET [URL with %C3%A9 instead of é]> (referer: `[URL with %C3%A9 instead of é])
...traceback...
`TypeError: argument of type 'NoneType' is not iterable`

When I try to handle the special characters, I get:

`ERROR: Error downloading <GET [URL with é and NOT %C3%A9]>`
...traceback...
`UnicodeEncodeError: 'ascii' codec can't encode character '\xf9' in position 25: ordinal not in range(128)`

My attempts at handling the special characters include:
(1) Writing a `process_request` function in `middlewares.py` that `urllib.parse.unquote`'s each URL
(2) *I could have done this part wrong* Created a `Request` subclass which is identical to `Request` but excludes the `safe_url_string()` step

Both of the attempts above result in the same error.

(I've posted a question about this on SO here https://stackoverflow.com/questions/46840714/scrapy-dealing-with-special-characters-in-url but no luck so far.)

moonsdarkside:
My issue was solved here: https://stackoverflow.com/a/47564798/6170115

cathalgarvey:
Is there anyone here who's tried @Skyross' solution and found that it didn't work? Looking at the code, I think his solution (yielding a subclass of Request that overrides the setter for URL) should be enough to solve this.

That's not to say we couldn't make this an option on `Request`, though. It should be a pretty easy change to make, to [make these two lines](https://github.com/scrapy/scrapy/blob/45b81693b46cd4a5abd95e2f4a106ca2126540f4/scrapy/http/request/__init__.py#L59) execute by default or set URL directly if safety is disabled.

[...]

PR #1949, opened 2016-04-21
Add encoding to Link object + re-enable HTML entities links tests

### LINKED ISSUE #1403: Exception in LxmLinkExtractor.extract_links 'charmap' codec can't encode character
```
Stacktrace (most recent call last):

  File "scrapy/utils/defer.py", line 102, in iter_errback
    yield next(it)
  File "scrapy/spidermiddlewares/offsite.py", line 28, in process_spider_output
    for x in result:
  File "scrapy/spidermiddlewares/referer.py", line 22, in <genexpr>
    return (_set_referer(r) for r in result or ())
  File "scrapy/spidermiddlewares/offsite.py", line 28, in process_spider_output
    for x in result:
  File "scrapy/spidermiddlewares/urllength.py", line 37, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "scrapy/spidermiddlewares/depth.py", line 54, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "scrapy/spiders/crawl.py", line 69, in _parse_response
    for requests_or_item in iterate_spider_output(cb_res):
  File "ex_link_crawl/spiders/external_link_spider.py", line 45, in parse_obj
    for link in LxmlLinkExtractor(allow=(), deny=self.allowed_domains).extract_links(response):
  File "scrapy/linkextractors/lxmlhtml.py", line 108, in extract_links
    links = self._extract_links(doc, response.url, response.encoding, base_url)
  File "scrapy/linkextractors/__init__.py", line 103, in _extract_links
    return self.link_extractor._extract_links(*args, **kwargs)
  File "scrapy/linkextractors/lxmlhtml.py", line 57, in _extract_links
    url = url.encode(response_encoding)
  File "encodings/cp1252.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)

```

My use of extractor is following:

```
def parse_obj(self, response):
        if not isinstance(response, HtmlResponse):
            return
        for link in LxmlLinkExtractor(allow=(), deny=self.allowed_domains).extract_links(response):
            if not link.nofollow:
                yield LinkCrawlItem(domain=link.url)
```

redapple:
Add `encoding` parameter and attribute to `Link` object.
That's the only way I found to properly account for encoding information when building requests from link extractors.

Supersedes #1880 
Should fix #1403

### DIFF
--- a/scrapy/link.py
+++ b/scrapy/link.py
@@ -13,9 +13,9 @@ from scrapy.utils.python import to_bytes
 class Link(object):
     """Link objects represent an extracted link by the LinkExtractor."""
 
-    __slots__ = ['url', 'text', 'fragment', 'nofollow']
+    __slots__ = ['url', 'text', 'fragment', 'nofollow', 'encoding']
 
-    def __init__(self, url, text='', fragment='', nofollow=False):
+    def __init__(self, url, text='', fragment='', nofollow=False, encoding='utf8'):
         if not isinstance(url, str):
             if six.PY2:
                 warnings.warn("Link urls must be str objects. "
@@ -28,6 +28,7 @@ class Link(object):
         self.text = text
         self.fragment = fragment
         self.nofollow = nofollow
+        self.encoding = encoding
 
     def __eq__(self, other):
         return self.url == other.url and self.text == other.text and \
--- a/scrapy/commands/bench.py
+++ b/scrapy/commands/bench.py
@@ -56,4 +56,5 @@ class _BenchSpider(scrapy.Spider):
 
     def parse(self, response):
         for link in self.link_extractor.extract_links(response):
-            yield scrapy.Request(link.url, callback=self.parse)
+            yield scrapy.Request(link.url, encoding=link.encoding,
+                                 callback=self.parse)
--- a/scrapy/linkextractors/__init__.py
+++ b/scrapy/linkextractors/__init__.py
@@ -100,7 +100,7 @@ class FilteringLinkExtractor(object):
         links = [x for x in links if self._link_allowed(x)]
         if self.canonicalize:
             for link in links:
-                link.url = canonicalize_url(urlparse(link.url))
+                link.url = canonicalize_url(link.url, encoding=link.encoding)
         links = self.link_extractor._process_links(links)
         return links
 
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -5,6 +5,7 @@ import six
 from six.moves.urllib.parse import urlparse, urljoin
 
 import lxml.etree as etree
+from w3lib.url import safe_url_string
 
 from scrapy.link import Link
 from scrapy.utils.misc import arg_to_iter, rel_has_nofollow
@@ -56,11 +57,12 @@ class LxmlParserLinkExtractor(object):
                 url = self.process_attr(attr_val)
                 if url is None:
                     continue
-            url = to_native_str(url, encoding=response_encoding)
-            # to fix relative links after process_value
-            url = urljoin(response_url, url)
+            url = safe_url_string(url, encoding=response_encoding)
+            url = urljoin(base_url, url)
+
             link = Link(url, _collect_string_content(el) or u'',
-                        nofollow=rel_has_nofollow(el.get('rel')))
+                        nofollow=rel_has_nofollow(el.get('rel')),
+                        encoding=response_encoding)
             links.append(link)
         return self._deduplicate_if_needed(links)
 
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -5,7 +5,6 @@ import six
 from six.moves.urllib.parse import urlparse, urljoin
 
 import lxml.etree as etree
-from w3lib.url import safe_url_string
 
 from scrapy.link import Link
 from scrapy.utils.misc import arg_to_iter, rel_has_nofollow
@@ -57,7 +56,6 @@ class LxmlParserLinkExtractor(object):
                 url = self.process_attr(attr_val)
                 if url is None:
                     continue
-            url = safe_url_string(url, encoding=response_encoding)
             url = urljoin(base_url, url)
 
             link = Link(url, _collect_string_content(el) or u'',
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -59,7 +59,8 @@ class CrawlSpider(Spider):
                 links = rule.process_links(links)
             for link in links:
                 seen.add(link)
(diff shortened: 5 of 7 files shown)


redapple:
If you open this HTML in a browser

```
<html>

<head>
  <meta http-equiv="content-type" content="text/html;charset=iso8859-15" />
</head>

<body>
  <p><a href="http://www.example/&hearts;/you?c=&euro;">text</a></p>
</body>

</html>
```
you'll see `http://www.example/%E2%99%A5/you?c=%A4` being sent when clicked

scrapy/link.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, url, text='', fragment='', nofollow=False):
         self.text = text
         self.fragment = fragment
         self.nofollow = nofollow
+        self.encoding = encoding
```
kmike: In Python 2 when unicode URL is passed it is converted to bytes using utf-8 encoding, regardless of `encoding` passed. Should we use `encoding` instead? If so, the warning may become misleading. There should be a warning if encoding is not passed, but if user passes it explicitly then it may be fine to pass unicode URL in Python 2.

scrapy/link.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, url, text='', fragment='', nofollow=False):
...
 
     def __eq__(self, other):
         return self.url == other.url and self.text == other.text and \
```
kmike: I think encoding should be a part of `__eq__`, `__hash__` and `__repr__`.

scrapy/link.py
redapple: I admit I don't quite understand why Unicode URL is converted to bytes within `Link`.
Was the intent to make it safe? (in that case, we could store the safe_url version of it, using encoding).
Reading `scrapy/tests/test_link.py`, I find [the following](https://github.com/scrapy/scrapy/blob/master/tests/test_link.py#L53) odd:

```
            link = Link(u"http://www.example.com/\xa3")
            self.assertIsInstance(link.url, str)
            self.assertEqual(link.url, b'http://www.example.com/\xc2\xa3')  <-----
```

kmike: In Python 2.x all URLs were bytes, so we required link.url to be bytes. For backwards compatibility, and because of missing unicode support in url-related functions in Python 2.x, link.url is still required to be bytes in Python 2.x

kmike: no idea about what is this test for :)

redapple: so, in Python2, we could make it pass through `safe_url_string`, using encoding, and converting to bytes?

kmike: Won't it make `safe_url_string` called twice? There are 2 issues with that: the first is performance (URL parsing can be a bottleneck); the second is that it'd stay in a way of https://github.com/scrapy/scrapy/issues/833

redapple: Do you mean `safe_url_string` called twice for Unicode input in Python 2 case?
I believe with `LinkExtractor` it won't happen with [this change](https://github.com/scrapy/scrapy/blob/link-encoding/scrapy/linkextractors/lxmlhtml.py#L60) from this very PR (`safe_url_string` returns `str` in Python2)

Sorry, I don't understand the relation to #833 . That PR is about `Request`, not `Link`.
Am I missing something?

kmike: I mean that URLs extracted by LinkExtractor are then passed to Request, and it calls `safe_url_string` again. #833 is about not calling `safe_url_string` there, but if link extractor calls it anyways then it won't work for urls extracted by link extractors.

redapple: Or do you mean `safe_url_string` called twice, once in `LinkExtractor` and in `Request` init?

kmike: yep

redapple: so your comment is on https://github.com/scrapy/scrapy/blob/link-encoding/scrapy/linkextractors/lxmlhtml.py#L60, right?

redapple: Tests seem to pass fine with `safe_url_string` in `LinkExtractor`. Checking further.

kmike: My comment was about your suggestion to use `safe_url_string` in Link constructor; I haven't noticed it in LinkExtractor.

redapple:
@kmike , with `safe_url_string` removed from link extraction (https://github.com/scrapy/scrapy/pull/1949/commits/54f1c2419e96c832932a9125f719b791bdd2694c), for Python2 Unicode URL input case in `Link` init, one actually needs UTF8 for `to_bytes`, otherwise `test_restrict_xpaths_with_html_entities` even fails ("iso8859-15" can only encode part of the Unicode characters)

What happens with `to_bytes(url, encoding='utf8')` in `Link.__init__()` is illustrated below:
```
Py2 warning: u'http://example.org/\u2665/you?c=\u20ac'
--> 'http://example.org/\xe2\x99\xa5/you?c=\xe2\x82\xac'

(link extraction's) canonicalize_url() gets this:
('http://example.org/\xe2\x99\xa5/you?c=\xe2\x82\xac', 'iso8859-15')

which goes through _safe_ParseResult(),
after parse_url('http://example.org/\xe2\x99\xa5/you?c=\xe2\x82\xac', None)

(None actually saves the whole thing, decoding UTF8)

the rest is fine after that:
ParseResult(scheme=u'http', netloc=u'example.org', path=u'/\u2665/you', params='', query=u'c=\u20ac', fragment='')

_safe_ParseResult --> 'http', 'example.org', '/%E2%99%A5/you', '', 'c=%A4', ''
...
```

scrapy/link.py
redapple: @kmike , I'm not sure how to handle encoding synonyms

kmike: A good point. Maybe https://docs.python.org/2/library/codecs.html#codecs.lookup could help.

redapple: @kmike awesome :)

kmike: but I think we should benchmark it :) If performance is a problem then not normalizing encoding names doesn't look too bad.

kmike: ah, we can always normalize encoding earlier and use a cache, so performance should be solvable

kmike: I'm worried about performance because link extractors are often on a hot path: pages may have hundreds of links

redapple: hm, adding encoding to comparison makes a lot of tests fail because many HtmlResponse in tests get detected as cp1252 without explicit encoding (while I was expecting UTF-8 by default)

kmike: Hm, why are they detected as cp1252? From reading https://github.com/scrapy/scrapy/blob/afac3fd2c28ac5b16e7e3c95fca5acd1334f8c58/scrapy/http/response/text.py#L96 I got an idea it'll try ascii first, then try utf8, and only then cp1252, if ascii and utf8 fails

PR #1950, opened 2016-04-24
[MRG+1] Place brackets on own lines with JsonItemExporter

patcon:
Placing the opening and closing brackets on their own lines makes it slightly easier to sort lines after the `spider_closed` signal is fired. (sorting is useful when using scrapy to generate open data sets that should have consistent item order, despite request disorder)


kmike:
Sounds fine.

PR #1952, opened 2016-04-25
added fedora23 info

XakV:
So I raised a docs issue the other day for this, but in testing, just dnf install python-scrapy worked ok, however, without that package - in a virtualenv, eg - you would need to install some pre-reqs so that pip install worked. Hopefully this reads ok. It's my first experience with sphinx.


Gallaecio:
@XakV It’s been quite some time, and I think we are no longer providing distro-specific installation instructions. Shall we close this?

Issue #1897, opened 2016-03-31
support for  "multipart/form-data" in scrapy.FormRequest


DharmeshPandav:

aron-bordin:
Hi @redapple can I get this one ?
I'm thinking about adding a `multipart` parameter, that defaults to False, in `from_response`. If False, uses the default `application/x-www-form-urlencoded`, if True, uses the `multipart/form-data` encoding type.

redapple:
@aron-bordin , sure!
Volunteers always welcome :)
Why in `from_response` and not the constructor?

AndreCAndersen:
It seems implementation doesn't support multiple input buttons. From what I understand only the input button which is pressed should be encoded into the body of the request. This version encodes all of them, so it is as if you click all buttons at the same time.

Urahara:
Any ETA when this feature will be avaliable?

AdrianAtZyte:
Closing after the move to [form2request](https://form2request.readthedocs.io/en/latest/usage.html) as a [replacement for `FormRequest`](https://github.com/scrapy/scrapy/pull/6438).

There is already an [existing upstream issue](https://github.com/scrapy/form2request/issues/8) for this issue.

PR #1954, opened 2016-04-26
[WIP] support multipart/form-data form encoding and file upload

aron-bordin:
### PR Overview

This PR add support for multipart/form-data form encoding.

Fixes https://github.com/scrapy/scrapy/issues/1897
- [x] Handle request header
- [x] Encode formdata
- [x] support file upload
- [ ] docs
1. Form fields and form files are present in the `formdata` parameter. To add a file in the `formdata`, it's necessary to use the new `MultipartFile` object (this is not yet documented, because I'd like to confirm if this api design is good first.) From the [multipart/form-data](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4) spec, files requires a special treatment, thats why we have the `MultipartFile` object. Non-files can be used as usual in the `formdata` parameter, including binary content.
2. Adds the `class MultipartFormRequest(FormRequest)`, designed to work exclusively with multipart requests.
3. A sample Spider using  this feature:
   
   ``` python
   # coding: utf-8
   import scrapy
   from scrapy.http.request.form import MultipartFormRequest, MultipartFile
   
   
   class SampleSpider(scrapy.Spider):
   
       name = 'sample'
       start_urls = ['http://0.0.0.0:5001']
   
       def parse(self, response):
           sample_file = MultipartFile('example £.png', open('/tmp/a.png').read())
           data = {
               'name': 'Test',
               'file': sample_file,
           }
   
           return MultipartFormRequest.from_response(response, formdata=data)
   ```

**updated at Oct 13**

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -9,6 +9,8 @@ from six.moves.urllib.parse import urljoin, urlencode
 import lxml.html
 from parsel.selector import create_root_node
 import six
+import string
+import random
 from scrapy.http.request import Request
 from scrapy.utils.python import to_bytes, is_listlike
 from scrapy.utils.response import get_base_url
@@ -49,6 +51,37 @@ class FormRequest(Request):
         return cls(url=url, method=method, formdata=formdata, **kwargs)
 
 
+class MultipartFormRequest(FormRequest):
+
+    def __init__(self, *args, **kwargs):
+        formdata = kwargs.pop('formdata', None)
+
+        kwargs.setdefault('method', 'POST')
+
+        super(MultipartFormRequest, self).__init__(*args, **kwargs)
+
+        content_type = self.headers.setdefault(b'Content-Type', [b'multipart/form-data'])[0]
+        method = kwargs.get('method').upper()
+        if formdata and method == 'POST' and content_type == b'multipart/form-data':
+            items = formdata.items() if isinstance(formdata, dict) else formdata
+            self._boundary = ''
+
+            # encode the data using multipart spec
+            self._boundary = to_bytes(''.join(
+                random.choice(string.digits + string.ascii_letters) for i in range(20)), self.encoding)
+            self.headers[b'Content-Type'] = b'multipart/form-data; boundary=' + self._boundary
+            request_data = _multpart_encode(items, self._boundary, self.encoding)
+            self._set_body(request_data)
+
+
+class MultipartFile(object):
+
+    def __init__(self, name, content, mimetype='application/octet-stream'):
+        self.name = name
+        self.content = content
+        self.mimetype = mimetype
+
+
 def _get_form_url(form, url):
     if url is None:
         return urljoin(form.base_url, form.action)
@@ -60,6 +93,29 @@ def _urlencode(seq, enc):
               for k, vs in seq
               for v in (vs if is_listlike(vs) else [vs])]
     return urlencode(values, doseq=1)
+
+
+def _multpart_encode(items, boundary, enc):
+    body = []
+
+    for name, value in items:
+        body.append(b'--' + boundary)
+        if isinstance(value, MultipartFile):
+            file_name = value.name
+            content = value.content
+            content_type = value.mimetype
+
+            body.append(b'Content-Disposition: form-data; name="' + to_bytes(name, enc) + b'"; filename="' + to_bytes(file_name, enc) + b'"')
+            body.append(b'Content-Type: ' + to_bytes(content_type, enc))
+            body.append(b'')
+            body.append(to_bytes(content, enc))
+        else:
+            body.append(b'Content-Disposition: form-data; name="' + to_bytes(name, enc) + b'"')
+            body.append(b'')
+            body.append(to_bytes(value, enc))
+
+    body.append(b'--' + boundary + b'--')
+    return b'\r\n'.join(body)
 
 
 def _get_form(response, formname, formid, formnumber, formxpath):
--- a/scrapy/__init__.py
+++ b/scrapy/__init__.py
@@ -32,7 +32,7 @@ twisted_version = (_txv.major, _txv.minor, _txv.micro)
 
 # Declare top-level shortcuts
 from scrapy.spiders import Spider
-from scrapy.http import Request, FormRequest
+from scrapy.http import Request, FormRequest, MultipartFormRequest
 from scrapy.selector import Selector
 from scrapy.item import Item, Field
 
--- a/scrapy/http/__init__.py
+++ b/scrapy/http/__init__.py
@@ -9,6 +9,8 @@ from scrapy.http.headers import Headers
 
 from scrapy.http.request import Request
 from scrapy.http.request.form import FormRequest
+from scrapy.http.request.form import MultipartFormRequest
+from scrapy.http.request.form import MultipartFile
 from scrapy.http.request.rpc import XmlRpcRequest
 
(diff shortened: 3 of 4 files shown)


redapple:
@aron-bordin ,
I'm not seeing where the data is actually encoded as [`multipart/form-data`](https://tools.ietf.org/html/rfc7578). Added tests don't test it.

Current `FormRequest` parameters require `formdata` to be a dict to be URL-encoded.
How does this transpose for multipart POSTs? Does `formdata` already need to be encoded?

Multipart support could mean changes in the API (which is not bad per se, if it's backward compatible)

aron-bordin:
Hi @redapple 

Sorry, this pr was just handling the header. I read the multipart's spec [[1]](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4) [[2]](https://tools.ietf.org/html/rfc7578) and I'll try to add the data encoding.  I'm thinking about two possibilities for `formdata`:

1. Use `formdata` as usual, representing `field: value` from the form and encode it as specified. Add a new field `formfiles`, or something like that, that is a dict that contains `field: file_path`. So these files can be read and added to the request body as explained in the multipart spec.
2. Use html fields and binary data (files) in `formdata`. In this case, `formdata` would be a dict that contains `field: [text value or a binary object representing the file]`

I prefer the first solution because it provides a simpler API and it's backward compatible. Please, let me know what do you think about it.

eLRuLL:
I would prefer the second way, mostly because it is closer to a form itself, but the first way is more user friendly.

Still we could start on the way to handle files on these requests.

Another thing that isn't yet implemented is for `FormRequest.from_response` to get the `enctype` of the form to determine the `Content-Type` of the request and default `application/x-www-form-urlencoded` if not specified.

aron-bordin:
Hi @redapple and @eLRuLL 

I updated the PR as follows:

1. Form fields and form files are present in the `formdata` parameter. To add a file in the `formdata`, it's necessary to add it as follows:
	```python
	formdata = {
		...
		'my_file': {
		    'name': 'FILE NAME',
		    'mimetype': 'MIME TYPE / OPTIONAL',
		    'content': 'FILE DATA'
		}
	}
	```
	this ^ is not yet documented, because I'd like to confirm if this api design is good first.

2. Now, the `from_response` method analyzes the `enctype` field in the form. If present, uses its value as default. If not, keep the `application/x-www-form-urlencoded` as default.

3. A sample Spider using  this feature:
    ```python
    import scrapy
    from scrapy.http.request.form import FormRequest


    class SampleSpider(scrapy.Spider):

        name = 'sample'
        start_urls = ['http://0.0.0.0:5001']

        def parse(self, response):
            data = {
                'name': 'Test',
                'file': {
                    'name': 'example.jpg',
                    'content': open('/tmp/a.jpg').read()
                }
            }

            return FormRequest.from_response(response, formdata=data)
    ```


This PR is working based on the new `multipart` parameter. However, this parameter can be removed, and we can just check if the `header['Content-Type']` is multipart/form-data. I like to have this parameter because it's user friendly, however this adds some unnecessary code to handle it. Do you have some opinions about it ?

eLRuLL:
great job @aron-bordin . I have some observations:

1. Now, I think the `multipart` argument makes everything look too forced. What happens if we get a different `enctype` in the future, then we'll have to add a new argument and setup all the `if`s necessary on the entire code, just like it is happening now.

2. A dictionary seem like a good way to specify the the parameters, but also it looks to forced and not natural, I would prefer to just pass a `File` object, and set defaults on the other fields for that case, and to create a `class` that could be passed on those cases, which should take care also of all the encoding. the `File` should also be transformed to this new class for encoding.

3. Passing the `enctype` directly on the headers seems more natural to me, maybe we can discuss this some more, but an argument to determine the whole functionality of the data handling seems too much.

redapple:
Sorry for commenting this late. I'm not sure that adding more arguments to `FormRequest` is the way to go now. The API is already quite cluttered IMO.

What would you think of having a new seperate `MultipartFormRequest` class, subclass of `FormRequest` probably, that would use a new `MultipartPart` object (or `FormRequestFile` as you use -- but does it have to be files only?) to hold the data, name, headers etc. to be sent for each part?

aron-bordin:
Hi @redapple 

For me, the current implementation looks simpler to maintain and more user friendly than creating a new class `MultipartFormRequest`. I removed the new parameters in current implementation, so the encoding type is now determined by the `header[Content-Type]`. For me, this looks more intuitive than having a new object to handle it. I updated the sample above to use it.

Also, I don't see where a new `MultipartPart` object could help, because the in the multipart form we'll have the same html inputs (but now it's accepted non ascii values by default) plus binary data (files). So having an API similar to the example above looks more intuitive and it makes less modification to the current scrapy API.

Do you see some advantages in separating the `MultipartFormRequest` instead of checking the header's content-type value ?

scrapy/http/request/form.py
```diff
@@ -46,9 +59,25 @@ def from_response(cls, response, formname=None, formid=None, formnumber=0, formd
         formdata = _get_inputs(form, formdata, dont_click, clickdata, response)
         url = _get_form_url(form, kwargs.pop('url', None))
         method = kwargs.pop('method', form.method)
+
+        form_enctype = form.xpath('@enctype')
+        # update the header enctype and request multipart parameter
+        if len(form_enctype) and form_enctype[0]:
+            headers = kwargs.pop('headers', {})
```
eLRuLL: I think there is no need on using `kwargs` for getting headers, just specifying it as a method argument would do, and to not use the previous `if`, just get it with `form.xpath('@enctype').extract_first()`

aron-bordin:
Hi @redapple 
I updated this PR, and separated the multipart feature to the `MultipartFormRequest`. I updated the PR description with more details about this PR, please, let me know what do you think about it :)

Thx

ashkulz:
@redapple, @kmike: if you feel this approach is OK, I'd like to finish off the docs to get this merged.

ashkulz:
Hmm, I was able to get it working using the following snippet:

```python
from urllib3.filepost import encode_multipart_formdata

class MultipartFormRequest(Request):
        formdata = kwargs.pop('formdata', None)
        if formdata and kwargs.get('method') is None:
            kwargs['method'] = 'POST'

        super(MultipartFormRequest, self).__init__(*args, **kwargs)

        if formdata:
            body, content_type = encode_multipart_formdata(formdata)
            self.headers[b'Content-Type'] = content_type.encode('utf-8')
```
It adds a dependency on urllib3, because there's no equivalent in w3lib (see scrapy/w3lib#14). Should I submit this as a separate PR?

AndreCAndersen:
It seems like this implementation doesn't support multiple input buttons. From what I understand only the input button which is clicked should be encoded into the body of the request. This version encodes all of them, so it is as if you click all buttons at the same time.

eLRuLL:
any progress here? https://stackoverflow.com/questions/48343098/how-to-use-scrapy-post-the-multipart-form-data-to-the-api

mtianyan:
As a noob , I want to have a MultipartFormRequest class. of course, maybe a more easy method like Requests lib 's method  :
` encoder = MultipartEncoder(
        data, boundary='-----------------------------41184676334')
    headers['Content-Type'] = encoder.content_type`
` result = session.post(url, headers=headers, data=encoder.to_string(), )` 

can be more friendly to me. Thank you @eLRuLL

Urahara:
Any ETA when this feature will be avaliable?

PR #1974, opened 2016-05-09
[MRG+1] Added: Retrying scrapy.core.downloader.handlers.http11.TunnelError

starrify:
It's a small change. Please just check the diff.

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -21,6 +21,7 @@ from twisted.internet.error import TimeoutError, DNSLookupError, \
 from scrapy.exceptions import NotConfigured
 from scrapy.utils.response import response_status_message
 from scrapy.xlib.tx import ResponseFailed
+from scrapy.core.downloader.handlers.http11 import TunnelError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


redapple:
hi @starrify , what kind of error triggered `TunnelError` in your tests?
did you see retrying them succeed eventually?

starrify:
Hi @redapple , here is the error message:
```
[scrapy.core.scraper] Error downloading <GET XXXX>: Could not open CONNECT tunnel.
```
For details, please check https://staging.scrapinghub.com/p/6327/job/9/17618/#/log?filterAndHigher&filterType=error (private access only). From the job stats we know there's no retrying on such failures.

Here's another sample job after applying the same fix in this PR: https://staging.scrapinghub.com/p/6327/job/9/17622/#details (private access only) where the same error has almost disapeared. Also the job stats suggest that there are many retries.

redapple:
@starrify , could this be configured via a new setting?

starrify:
I'm afraid not, since this `EXCEPTIONS_TO_RETRY` thing is initialized here in this file and never updated later.

However I think it might be a good idea to make it configurable through settings. :)

redapple:
@starrify , I'm not sure I understand your comment.
one could still update an instance attribute with exceptions to retry, defaulting to `EXCEPTIONS_TO_RETRY`, adding `TunnelError` to the tuple if some new `RETRY_TUNNEL_ERROR` is `True` or something.
maybe I'm missing something?

starrify:
Hi, seems I'm the one who misunderstood. :( I thought you asked _could this be configured now via a downloader setting?_

I have updated this PR by introducing a new setting `RETRY_EXCEPTIONS_EXTRA` and removed `TunnelError` from the default ones.

But personally I still think `TunnelError` shall be retried by default like other errors.

redapple:
@starrify , after some though, I think your initial commit (adding `TunnelError`) makes more sense in the end. Sorry for making you update it.

starrify:
np :)

I guess making `TunnelError` default and introducing a configurable `EXCEPTIONS_TO_RETRY` are both good things to do. But one thing at a time, and I've just discarded the 2nd commit in this PR.

redapple:
Thanks @starrify

PR #1979, opened 2016-05-11
[MRG+1] Prevent empty warnings in case of certificate verification failures

### LINKED ISSUE #1976: empty WARNING message in scrapy.core.downloader.tls (1.1.0rc4/master)
Sometimes I'm getting empty warnings now, on 1.1.0rc4 and master branch.
(at least on rc3 as well)
```
2016-05-07 00:33:46 [scrapy.core.downloader.tls] WARNING: 
2016-05-07 00:33:47 [scrapy.core.downloader.tls] WARNING: 
2016-05-07 00:33:48 [scrapy.core.downloader.tls] WARNING: 
```

It happens in a broad linkcheck crawl; so I couldn't pinpoint what URLs might be responsible for that, at this time. The only other observation so far is, that it doesn't happen on a cache-replayed run (which might be obvious, as there is no TLS there).

redapple:
Fixes #1976

`VerificationError` exception can have an empty string representation ([e.g. for `service-identity<16`](https://github.com/scrapy/scrapy/issues/1976#issuecomment-218422811))

This patch adds a (more) meaningful message before the exception message just in case.

### DIFF
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -44,7 +44,9 @@ try:
                 try:
                     verifyHostname(connection, self._hostnameASCII)
                 except VerificationError as e:
-                    logger.warning(e)
+                    logger.warning(
+                        'Remote certificate is not valid for hostname "{}"; {}'.format(
+                            self._hostnameASCII, e))
 
 except ImportError:
     # ImportError should not matter for older Twisted versions


nyov:
I believe I'm okay with this, thanks for asking.
Since I don't actually know what the Exception message would say if `str(e)` returns one (perhaps then it's redundant?), I can't absolutely say.

But you probably know and tested that case, so +1;
thanks for debugging and fixing it!

redapple:
The exception messages I've seen look like this:
> VerificationError(errors=[DNSMismatch(mismatched_id=DNS_ID(hostname='www.knorr-bremsecvs.com'))])

with this change, you get something like:

```
2016-05-11 13:11:59 [scrapy] WARNING: Remote certificate is not valid for hostname "www.knorr-bremsecvs.com"; VerificationError(errors=[DNSMismatch(mismatched_id=DNS_ID(hostname='www.knorr-bremsecvs.com'))])
```

nyov:
Oh, that is an ugly string representation... so having the custom error message first is actually a nicer explanation, even if the exception message is there.

Let's merge it, I think.

PR #1984, opened 2016-05-11
[MRG+1] Update release notes and date for 1.1.0

redapple:
I'm removing "(non-Python 3 ported Twisted dependency)" as some of the limitations may already have been lifted in more recent Twisted versions, and because, in practice, we may not have tried enough to port them.


eliasdorneles:
Looks good!

PR #1986, opened 2016-05-11
Add fallback for upper case attribute names in Images Pipeline (fixes #1985)

### LINKED ISSUE #1985: Scrapy 1.1.0 ImagesPipeline backward incompatibility
#1891 is not backward compatible.
Users of Scrapy<1.1 `ImagesPipeline` could access upper-case attributes,
e.g. 

```
class CustomImagesPipeline(ImagesPipeline):
    (...)
    def item_completed(self, results, item, info):
        item = super(CustomImagesPipeline, self).item_completed(
            results, item, info)
        # Note: not all items do have an images field.
        if self.IMAGES_RESULT_FIELD not in item.fields:
            return item
```

Leading to the following exception in 1.1.0(rc4):
```
  File "/usr/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
  File "/tmp/unpacked-eggs/__main__.egg/*****/pipelines/screenshots.py", line 539, in item_completed
    results, item, info)
  File "/tmp/unpacked-eggs/__main__.egg/*****/pipelines/screenshots.py", line 242, in item_completed
    images = item.pop(self.IMAGES_RESULT_FIELD, [])
AttributeError: 'CustomImagesPipeline' object has no attribute 'IMAGES_RESULT_FIELD'
```

eliasdorneles:
This adds a fallback for upper case attribute names, with a deprecation notice to deprecate them in Scrapy 1.2 (when we can simply remove the tests and `__getattr__` methods).

Does this look good?

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -21,7 +21,6 @@ from scrapy.settings import Settings
 from scrapy.exceptions import DropItem
 #TODO: from scrapy.pipelines.media import MediaPipeline
 from scrapy.pipelines.files import FileException, FilesPipeline
-from scrapy.utils import deprecate
 
 
 class NoimagesDrop(DropItem):
@@ -38,6 +37,15 @@ class ImagesPipeline(FilesPipeline):
     """
 
     MEDIA_NAME = 'image'
+
+    _DEPRECATED_ATTRS = {
+        'IMAGES_EXPIRES': 'expires',
+        'IMAGES_URLS_FIELD': 'images_urls_field',
+        'IMAGES_RESULT_FIELD': 'images_result_field',
+        'IMAGES_MIN_WIDTH': 'min_width',
+        'IMAGES_MIN_HEIGHT': 'min_height',
+        'IMAGES_THUMBS': 'thumbs',
+    }
 
     def __init__(self, store_uri, download_func=None, settings=None):
         super(ImagesPipeline, self).__init__(store_uri, settings=settings, download_func=download_func)
@@ -60,22 +68,6 @@ class ImagesPipeline(FilesPipeline):
 
         store_uri = settings['IMAGES_STORE']
         return cls(store_uri, settings=settings)
-
-    def __getattr__(self, name):
-        lower_name = name.lower()
-
-        if name != lower_name and hasattr(self, lower_name):
-            deprecate.attribute(self, name, lower_name, version='1.2')
-            return getattr(self, lower_name)
-
-        if name.startswith('IMAGES_'):
-            alt_name = name[len('IMAGES_'):].lower()
-            if hasattr(self, alt_name):
-                deprecate.attribute(self, name, alt_name, version='1.2')
-                return getattr(self, alt_name)
-
-        msg = '{.__name__!r} object has no attribute {!r}'
-        raise AttributeError(msg.format(self.__class__, name))
 
     def file_downloaded(self, response, request, info):
         return self.image_downloaded(response, request, info)
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -21,6 +21,7 @@ from scrapy.settings import Settings
 from scrapy.exceptions import DropItem
 #TODO: from scrapy.pipelines.media import MediaPipeline
 from scrapy.pipelines.files import FileException, FilesPipeline
+from scrapy.utils import deprecate
 
 
 class NoimagesDrop(DropItem):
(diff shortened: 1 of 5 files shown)


pawelmhm:
this fixes problem I had with this, I no longer get any exceptions, so :+1:  from me

redapple:
I wonder if we shouldn't be stricter on the attribute names checked.
The warning in the following case looks a bit weird to me:

```
>>> from scrapy.pipelines.images import ImagesPipeline
>>> p = ImagesPipeline('./')
>>> p.test = 3
>>> p.TEST
__main__:1: ScrapyDeprecationWarning: ImagesPipeline.TEST attribute is deprecated and will be no longer supported in Scrapy 1.2, use ImagesPipeline.test attribute instead
3
>>> p.TESt
__main__:1: ScrapyDeprecationWarning: ImagesPipeline.TESt attribute is deprecated and will be no longer supported in Scrapy 1.2, use ImagesPipeline.test attribute instead
3
```

redapple:
Also, `FilesPipeline` had similar changes that we can address.

eliasdorneles:
Right, perhaps will be best to create a table of deprecated attributes, then.
Will update the PR soon.

eliasdorneles:
@pawelmhm can you check again please? :)

eliasdorneles:
hm, any idea why build failed in the precise build?
am I using assertRaises wrong?

pawelmhm:
hey @eliasdorneles I rerun my custom image pipelines tests with your branch and it still fails.

Old pipeline had attribute MIN_WIDTH which is now renamed to IMAGES_MIN_WIDTH. So it's still not fully backward compatible.

On my side I'm updating all this code to use new interface (get values from settings and not from uppercase class attributes) and I'm going to have code working with new pipeline on master, but if we care about not breaking things for people using undocumented interfaces we should keep MIN_WIDTH.

tests/test_pipeline_images.py
```diff
@@ -95,6 +94,25 @@ def test_convert_image(self):
         self.assertEquals(converted.mode, 'RGB')
         self.assertEquals(converted.getcolors(), [(10000, (205, 230, 255))])
 
+    def test_deprecate_uppercase_attribute_names(self):
+        # given:
+        class CustomImagesPipeline(ImagesPipeline):
+            def item_completed(self, *a, **kw):
+                print('Using result field: %s', self.IMAGES_RESULT_FIELD)
```
pawelmhm: you forgot about removing print?

pawelmhm:
Unfortunately I found another potential problem. If some user defined his own field name on pipeline subclassing image pipeline and calls super he'll likely get incorrect results now. Consider following code:


```python
from scrapy.pipelines.images import ImagesPipeline


class MyPipe(ImagesPipeline):
    IMAGES_RESULT_FIELD = "my_field"

    def item_completed(self, results, item, info):
        item = super(MyPipe, self).item_completed(results, item, info)
        print("doing some stuff with item here")
        return item

a = MyPipe("s3://foo.bar/")
results =[
    (True,
  {'checksum': 'aaa',
   'path': 'aaadsasd/adasdsads',
   'url': 'https://splash.url.com/execute?timeout=60&filters=easylist'})]
item = a.item_completed(results, {"name": "myitem"}, None)
print(item)
assert item.get("my_field") is not None
```

This code redefines IMAGES_RESULT_FIELD and calls super(). In previous Scrapy version this put images into "my_field". In new pipeline this will work differently, download results will now be placed into "images" field not into "my_field".

pawelmhm:
Sorry for multiple comments, but I stumbled across another problem.

I actually have multiple (3) pipelines inheriting from Image Pipeline (different types of images and different use cases, e.g. plain image downloads, screenshots of pages, and some other types of screensots). What will be recommended way to set image settings for multiple pipelines?

With current code I see I can define keys in project settings, but they are set per project. So there is one key "IMAGES_RESULT_FIELD" defined in settings. What if I'd like to define different fields in my 3 different pipelines for 3 different item classes? I should not use class attribute because this will be deprecated, but I can't use global project settings because I need different settings in each of pipelines. How can I customize my pipelines then?

scrapy/pipelines/images.py
```diff
@@ -38,6 +38,15 @@ class ImagesPipeline(FilesPipeline):
...
+    _DEPRECATED_ATTRS = {
+        'IMAGES_EXPIRES': 'expires',
+        'IMAGES_URLS_FIELD': 'images_urls_field',
+        'IMAGES_RESULT_FIELD': 'images_result_field',
+        'IMAGES_MIN_WIDTH': 'min_width',
+        'IMAGES_MIN_HEIGHT': 'min_height',
+        'IMAGES_THUMBS': 'thumbs',
+    }
+
     def __init__(self, store_uri, download_func=None, settings=None):
```
pawelmhm: dont you think that perhaps easier approach would be just having uppercase class attributes and defaulting to them in pipeline init method? So dont add shim for getattr but rather change things in such a way that 

a) deprecated keys are not deprecated, but they are kept as class attributes. 
b) pipeline init checks for settings values, if there are not settings it defaults to class attributes
c) pipeline uses lowercase attributes all around

This would allow us to have cake and eat it, and has following benefits:

* there are no problems with backward compatibility
* if someone is like me and has multiple pipelines he can just define different class attributes in each one of them and thus have different settings for each pipeline.

Drawback of this solution is:
* we dont deprecate class attributes (but do we really need it?)

Perhaps there are other drawbacks I'm not aware of.

pawelmhm: I created PR implementing above ideas see here #1989 1989

eliasdorneles:
Yeah, being able to set the image field per pipeline derived class is indeed a great advantage, +1. =)

I'm closing this PR.

PR #1991, opened 2016-05-15
[WIP / Discuss] Scrapy Streaming docs

aron-bordin:
## PR Overview
This is an initial work in the scrapy streaming docs.
You can read it here: http://gsoc2016.readthedocs.io/en/latest/topics/streaming.html

I'd like to open the discussion about the communication protocol. It's pretty similar with the original protocol in my proposal, with some modifications. My idea is to open the process of the API design, so I can get some feedbacks and modify this API before implementing it. 

In my proposal, I've suggested to start the implementation of this API on June 13, so it'd be helpful if we get a definitive API before this date.

Also, suggestion about new messages and new behavior are welcome :smile:

### DIFF
--- a/docs/_ext/scrapydocs.py
+++ b/docs/_ext/scrapydocs.py
@@ -99,6 +99,11 @@ def setup(app):
         rolename      = "reqmeta",
         indextemplate = "pair: %s; reqmeta",
     )
+    app.add_crossref_type(
+        directivename="message",
+        rolename="message",
+        indextemplate="pair: %s; message",
+    )
     app.add_role('source', source_role)
     app.add_role('commit', commit_role)
     app.add_role('issue', issue_role)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -69,6 +69,7 @@ Basic concepts
    topics/link-extractors
    topics/settings
    topics/exceptions
+   topics/streaming
 
 
 :doc:`topics/commands`
@@ -106,6 +107,9 @@ Basic concepts
 
 :doc:`topics/exceptions`
     See all available exceptions and their meaning.
+
+:doc:`topics/streaming`
+    Write Spiders using any Programming Language using Scrapy Streaming.
 
 
 Built-in services
--- a/docs/topics/streaming.rst
+++ b/docs/topics/streaming.rst
@@ -19,11 +19,12 @@ The external spider must communicate using the system ``stdin`` and ``stdout`` w
 You can run standalone external spiders using the :command:`streaming` command.
 
 If you want to integrate external spiders with a scrapy's project, create a file named ``external.json``
-in your project root. This file must contain an array of json objects, each object with the ``name``
-and ``command`` attributes.
+in your project root. This file must contain an array of json objects, each object with the ``name`` ,
+``command``, and ``args`` attributes.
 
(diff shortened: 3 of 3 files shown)


rolando:
I have a suggestion: focus on the response/requests part and leave for later the selector stuff.

Even though the selector parts are a nice to have, I feel like it just just adds an overhead trying to interact with the scrapy's parser capabilities from external spiders.

A few missing parts: 

* What's the result from a request being filtered out by the dupefilter? This is not an error.
* What's the result from a request that returns 404 repsonse? These non-200 responses don't reach the spider under normal circumstances (i.e. they need to set the flag to handle all statuses). But an external spider may want to know what happened to its requests.

It may be good to have a way to run this protocol against real scrapy spiders (i.e.: running scrapy spiders as external spiders). This could give us a good sense of what can be supported and what not.

kmike:
I'm late to the party, sorry! 

Starting with docs is a great way to approach this project 👍 

What do you think about creating this extension as a separate package (either in scrapy organisation or in scrapy-plugins organisation) and contributing necessary fixes to Scrapy so that it supports this kind of extensions?

aron-bordin:
Hi, thank you the feedback :) 

@rolando for both duplicate requests and non-200, I think that they should be dispatched in the ``exception`` message, because they are caught in the spider runtime and it was not expected by the spider. 

@kmike I'll talk with my mentors about it. But talking about implementation, it's not too dependent on scrapy source code, so I think that if necessary, I can implement it as an extension with a few modifications in the original proposal

eLRuLL:
@kmike I agree with starting a new repository on scrapy-plugins

aron-bordin:
Adding some comments about the implementation:

Originally, I've suggested to implement the communication channel between scrapy and the external spider using the [ProcessProtocol](http://twistedmatrix.com/documents/current/core/howto/process.html) from twisted, and as pointed in the docs, each message ends with a line break ``\n``.
I started an initial [POC](https://github.com/aron-bordin/POC-Streaming) to get an idea on how this should work.

However, this implementation could get some problems with buffering, because the messages sent by ``transport.write`` and received by ``outReceived`` can be buffered by the system.

Checking the @Preetwinder POC, he uses https://github.com/Preetwinder/ScrapyStreaming/blob/master/linereceiverprocess.py#L53 to wrap the process and avoid this buffer issues. 

Now I'm analyzing the best way to approach this possible problems with stdin/stdout buffering. 

As long as all messages must end with a line break, both implementations (streaming core and external spiders) could "buffer" the received data and process it after receiving the line break (end of the message).
Also, a different implementation could make it easier. Using the [LineReceiver](https://twistedmatrix.com/documents/9.0.0/api/twisted.protocols.basic.LineReceiver.html) in the streaming core could help while receiving data. But I'm still not sure about the best way to write in the process stdin, unfortunately ``stdbuf`` is not available in all platforms. 

As part of the communication protocol. the line break is defined as the end of the message. If the external spider developer uses this information and just analyze the received data after the line break, this should be enough.

Do you have any comments about the implementations and these possible issues ?

eLRuLL:
the https://github.com/scrapy-plugins/scrapy-streaming repository is created, let's move the conversation there I think

aron-bordin:
I'm moving this PR to the new repo https://github.com/scrapy-plugins/scrapy-streaming/pull/1

Thanks

PR #1106, opened 2015-03-25
[MRG+1] some improvements to overview page

eliasdorneles:
Hey folks!

Here is my proposal for addressing issue #609 (and replaces PR #1023, while keeping some ideas).

Since the overview is "the pitch", I tried my best to make it short and to the point.

Summary of the changes:

* Added example spider showcasing both scraping and crawling (link following)
* Wrote an explanation of what the code does, without delving much into details
* Summarized table of features in the end, and reordered them based on my gut feeling
* Cut some text
* Cut some more

Note: the example spider is also showcasing the features from PRs #1081 and #1086, that I assume will be also in Scrapy 1.0 release.

So, what do you think, does this look good?

Thank you!

PR #1994, opened 2016-05-17
[DOC][Overview] Use idiomatic .extract_first()

redapple:


kmike:
It looks like it was me who pursuaded @eliasdorneles to remove extract_first from this example (https://github.com/scrapy/scrapy/pull/1106#discussion_r27220120). I'm fine with adding it now :)

eliasdorneles:
Alrighty! :)

PR #2005, opened 2016-05-24
[MRG+1] Included new optional parameter in startproject command line

feliperuhland:
I don't know if anyone creates project like me.
I have a habit to create git repository and then I create scrapy project.

You can see something similar at [django startproject command](https://docs.djangoproject.com/en/1.9/ref/django-admin/#startproject).

Unfortunately, I had to rewrite ``copytree`` function, to make sure that the ``startproject`` won't fail if it ``project_dir`` already exists.

I am available to any feedback,

Thanks.

### DIFF
--- a/scrapy/commands/startproject.py
+++ b/scrapy/commands/startproject.py
@@ -79,10 +79,11 @@ class Command(ScrapyCommand):
 
         if len(args) == 2:
             project_dir = args[1]
-            if exists(join(project_dir, 'scrapy.cfg')):
-                self.exitcode = 1
-                print('Error: scrapy.cfg already exists in %s' % abspath(project_dir))
-                return
+
+        if exists(join(project_dir, 'scrapy.cfg')):
+            self.exitcode = 1
+            print('Error: scrapy.cfg already exists in %s' % abspath(project_dir))
+            return
 
         if not self._is_valid_name(project_name):
             self.exitcode = 1
--- a/scrapy/commands/startproject.py
+++ b/scrapy/commands/startproject.py
@@ -79,13 +79,13 @@ class Command(ScrapyCommand):
                     copy2(srcname, dstname)
             # catch the Error from the recursive copytree so that we can
             # continue with other files
-            except EnvironmentError, err:
+            except EnvironmentError as err:
                 errors.extend(err.args[0])
-            except EnvironmentError, why:
+            except EnvironmentError as why:
                 errors.append((srcname, dstname, str(why)))
         try:
             copystat(src, dst)
-        except OSError, why:
+        except OSError as why:
             if WindowsError is not None and isinstance(why, WindowsError):
                 # Copying file access times may fail on Windows
                 pass
--- a/scrapy/commands/startproject.py
+++ b/scrapy/commands/startproject.py
@@ -1,9 +1,10 @@
 from __future__ import print_function
 import re
+import os
 import string
 from importlib import import_module
 from os.path import join, exists, abspath
-from shutil import copytree, ignore_patterns, move
+from shutil import ignore_patterns, move, copy2, copystat
 
 import scrapy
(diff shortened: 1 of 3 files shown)


eliasdorneles:
Hello, @feliperuhland.

I like the feature idea, it will be nice to be able to do `scrapy startproject myproj .` inside a directory.

The build is failing for Python 3, can you have a look?

feliperuhland:
Thanks, @eliasdorneles . I am working on it.

eliasdorneles:
The template should also be updated to print a message with the correct project dir.

I just tried it with:

    $ scrapy startproject myproj customdir

And it printed:

```
    cd myproj
```

It should point me to `cd customdir`

feliperuhland:
@eliasdorneles you're right.
I will fix this.

eliasdorneles:
Looks great, thanks @feliperuhland !

scrapy/commands/startproject.py
```diff
@@ -43,41 +44,68 @@ def _module_exists(module_name):
...
-        if len(args) != 1:
+        if len(args) not in (1, 2):
             raise UsageError()
+
         project_name = args[0]
+        project_dir = args[0]
+
+        if len(args) == 2:
+            project_dir = args[1]
+            if exists(join(project_dir, 'scrapy.cfg')):
```
eliasdorneles: hey @feliperuhland, I'm looking at this again, and I'm thinking this check if `scrapy.cfg` exists is needed in both cases (` args == 1` or `args == 2`), right?
looks like it should be outside of this `if`.

scrapy/commands/startproject.py
```diff
@@ -43,41 +44,68 @@ def _module_exists(module_name):
...
                     ' only\nletters, numbers and underscores')
-        elif exists(project_name):
-            print('Error: Directory %r already exists' % project_name)
         elif _module_exists(project_name):
             print('Error: Module %r already exists' % project_name)
         else:
             return True
         return False
 
+    def _copytree(self, src, dst):
```
eliasdorneles: Also, do you think we could get rid of this function, if we use the `ignore` callable argument for `shutil.copytree` ?

scrapy/commands/startproject.py
feliperuhland: Not really, because in the case of ``len(args) == 1`` the directory will be created and can't have the ``scrapy.cfg``. But I can't see any problem in this change. What do you think? Thanks.

scrapy/commands/startproject.py
feliperuhland: Hi, @eliasdorneles. I would like to remove that function, but the original implementation always creates the destination directory and that is not useful to us. I'm not happy of rewrite some internal library behavior. Do you have any other ideias? Thanks again.

eliasdorneles: Ah yeah, you're right.
The problem is that `os.makedirs` call that is always called and requires the directory not to exist.
Can you perhaps add a docstring explaining this, then?

scrapy/commands/startproject.py
eliasdorneles: Since we're using a custom copytree that now works in existing directories, I think we shouldn't assume the directory _will_ be created.
Users can now do `mkdir myproj; scrapy startproject myproj` and scrapy will work without complaining, therefore the extra check is a good idea.

In fact, I just tried running `scrapy startproject myproj` twice and it breaks the second time.

feliperuhland:
@eliasdorneles I done what you suggested and I think it's good. Can you read the docstring to see if it's good enough?

Thanks again!

eliasdorneles:
Yup, looks great now. Thank you, @feliperuhland !

@redapple this looks ready. How do you feel about merging it for 1.2?

redapple:
Thanks @feliperuhland !

Issue #1899, opened 2016-03-31
HttpCompressionMiddleware tries to decode HEAD responses


ArturGaspar:
The middleware tries to decode the body of HEAD responses, which are not expected to have a message body.

redapple:
Fixed in https://github.com/scrapy/scrapy/pull/2008

PR #2008, opened 2016-05-24
[MRG+1] Fix issue HttpCompressionMiddleware tries to decode HEAD responses #1899

foromer4:
Not sure if this is required or not, but seems like a good way to start working on this project...

### DIFF
--- a/tests/test_downloadermiddleware_httpcompression.py
+++ b/tests/test_downloadermiddleware_httpcompression.py
@@ -153,4 +153,4 @@ class HttpCompressionTest(TestCase):
         response = response.replace(body = None)
         newresponse = self.mw.process_response(request, response, self.spider)
         self.assertIs(newresponse, response)
-        self.assertEquals(response.body, '')
+        self.assertEquals(response.body, b'')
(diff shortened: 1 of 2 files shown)


tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -144,3 +144,13 @@ def test_process_response_gzipped_contenttype(self):
...
+
+    def test_process_response_head_request_no_decode_required(self):
+        response = self._getresponse('gzip')
+        response.headers['Content-Type'] = 'application/gzip'
+        request = response.request
+        request.method = 'HEAD'
+        response = response.replace(body = None)
+        newresponse = self.mw.process_response(request, response, self.spider)
+        self.assertIs(newresponse, response)
+        self.assertEquals(response.body, '')
```
eliasdorneles: The tests are failing in Python 3 because `''` is not bytes, but a native string (which in Python 3 is the same as unicode object in Python 2).
If you update this to `b''` tests should pass.

foromer4: thank you

redapple:
Thanks @foromer4

PR #2009, opened 2016-05-24
DON'T MERGE - Travis build test

eliasdorneles:
This is just to test the Travis build, which seems to be broken


eliasdorneles:
The problem was a pkginfo release with broken wheel: https://pypi.python.org/pypi/pkginfo
I'm closing this and re-running the builds for the last couple PRs.

PR #2015, opened 2016-05-27
[MRG+1] Use "url" variable in the example

lopuhin:
Instead of hardcoded http://www.example.com, as noticed by Andrés Pérez-Albela H.: without it url variable is unused and only one request will make it past dupefilter.


redapple:
Thanks @lopuhin

PR #2026, opened 2016-06-04
[MRG+1] Encourage complete titles in pull requests

Digenis:
Linking to an issue when creating a PR is useful
but authors sometimes replace the whole subject with "Fix+tests for #000"
which makes it hard to skim through the issue tracker or its feed
especially when the description is missing from the PR itself.
This is even more important for commit messages.

With this pr I want to open the discussion;
I'm not fully convinced that this is the role of the documentation
but I wrote it there as a start.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -100,6 +100,11 @@ starting point is to send a pull request on GitHub. It can be simple enough to
 illustrate your idea, and leave documentation/tests for later, after the idea
 has been validated and proven useful. Alternatively, you can send an email to
 `scrapy-users`_ to discuss your idea first.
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -100,6 +100,11 @@ starting point is to send a pull request on GitHub. It can be simple enough to
 illustrate your idea, and leave documentation/tests for later, after the idea
 has been validated and proven useful. Alternatively, you can send an email to
 `scrapy-users`_ to discuss your idea first.
+When writting GitHub pull requests, try to keep titles short but descriptive.
```
nyov: s/writting/writing/

Issue #1915, opened 2016-04-10
Allow multiple items through pipelines?


dxue2012:
The documentation for pipeline specifies that `process_item` must either return a dict with data, `Item` object or raise a `DropItem` exception. Is there a reason why we aren't allowed to return an iterable of dictionaries with data (or `Item` objects)? It seems impossible to write a pipeline that modifies the input item and returns multiple items under the current framework.

Thank you!

kmike:
For me it makes sense to allow returning multiple items; I'm fine with merging a PR which implements this feature.

kmike:
@dxue2012 what is your use case?

dxue2012:
Cool! I'm considering writing a split/merge pipeline for my scraper that either increases or decreases the number of returned items. Here's an example of splitting an item by one field:

```
class SplitPipeline(object):
    def process_item(self, item, spider):
        """
        Example:
        item = {'a': 1, 'b': 'foo bar'}

        return an iterable with
        {'a':1, 'b': 'foo'}, {'a':1, 'b': 'bar'}
        """
        for v in item['b'].split():
            item['b'] = v
            yield item
```

pawelmhm:
It would be nice to be able to yield multiple items in pipelines. 

I think that with current state of things the only way to "split" or multiply items is either doing this in spider middleware [`process_spider_output`](http://doc.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.process_spider_output) 

```python

def process_spider_output(self, response, result, spider):
     for x in result:
          if isinstance(x, Item):
              for y in split items: 
                    yield y
                    continue
          yield x
```

or creating nested items in pipeline. 

E.g. you can have item: 

```python
{
"name":"foo",
"nested_items": {"id": 123, "value": "alfa"} # field containing other items
}
```

but nested items are not very convenient to work with and you have to think hard to get proper logical handling of them (at least this is my experience), plus you end up with more complex output.

jeremysmitherman:
Just wanna +1 this, my use case is I'm parsing magic the gathering cards from their website, but some cards are actually doublesided so it would be really useful for me to detect this in a pipeline, create a new item for the other side of the card, and then return them both to the rest of the pipeline.

eLRuLL:
something to think about would be if these new items should go again for the whole pipelines process, so if they should pass again through all pipelines, because one (the main one) already passed through some of the pipelines.

[...]

PR #2029, opened 2016-06-05
#1915: Multiple items through item pipelines

mgachhui:
My solution to #1915. Approach:

- All functions in a pipeline are put inside closures and a process chain is made out of them.
- For the n<sup>th</sup> pipeline, it's output is handled by the n+1<sup>th</sup> pipelineHandler.
- When an iterable is encountered, a parallel call is generated and a DeferredList is returned. According to the Twisted docs, if a callback returns a deferred, then the callchain is paused until that deferred completes. So the concurrency limits should be maintained in my approach.
- When the result of a DeferredList is encountered, a SplitItem exception is raised to exit the callchain.

Hiccups:

- I had to add two extra functions at the end of the pipeline just to handle corner cases. Not sure if there's another way around it.
- My plan was to show the split items when the exception is thrown, but I'm not sure how to do that.
- When an item is split, scraper._itemproc_finished should show it like a normal item, as being scraped from the original response. I don't know how to access the response without changing the function signature.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -11,7 +11,7 @@ from scrapy.utils.defer import defer_result, defer_succeed, parallel, iter_errba
 from scrapy.utils.spider import iterate_spider_output
 from scrapy.utils.misc import load_object
 from scrapy.utils.log import logformatter_adapter, failure_to_exc_info
-from scrapy.exceptions import CloseSpider, DropItem, IgnoreRequest
+from scrapy.exceptions import CloseSpider, DropItem, SplitItem, IgnoreRequest
 from scrapy import signals
 from scrapy.http import Request, Response
 from scrapy.item import BaseItem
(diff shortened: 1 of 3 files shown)


lookfwd:
It feels a bit too intrusive. Can you show a small motivating example (spider/pipeline code?)

mgachhui:
I'm not sure about what you meant by intrusive. Could you explain a bit?

An example will be as follows:
```
class Pipeline1(object):
        # Process item as required
        # process_item() can return a single item/dict 
        output = item
        # or an iterable as well
        # output = [some_item, other_item]
        return output  
```
When a pipeline component returns an iterable, each of the elements of the iterable are treated in parallel by the rest of the pipeline.
```
A -> itemA -> B -> itemB -> C
                 -> itemD -> C
```

lookfwd:
By "intrusive" I mean that if it looked like a straightforward 5-line fix, it would be much more easy to judge given my limited attention span etc. :) Continuing on #1915

wRAR:
The original issue was closed in favor of spider middlewares.

Issue #8, opened 2011-09-09
Refactor signals


dangra:
Refactor signal handling similar how [Django did for 1.0](http://code.djangoproject.com/wiki/BackwardsIncompatibleChanges#Signalrefactoring).

According to them, the new approach brings up to 90% speed improvements

jdemaeyer:
Here is a partial proof of concept for switching the backend to ``django.dispatch`` while conserving support for the old signaling API: http://jakobdemaeyer.com/x/scrapy_signaling_proof.html

In this implementation, the SignalManager is deprecated and replaced by a NewSignalManager whose sole purpose is backwards-compatibility. The major open question is how to assign the Crawler instance as default sender for a signal, as the signals are now essentially completely decoupled from anything else.

A different implementation could preserve using the SignalManager by solely rewriting the backend, using the NewSignalManager implementation without warnings. Signals would then still be emitted (and handlers registered) via the Crawler.signals attribute instead of on the Signals themselves.

jdemaeyer:
Here's a list of signaling frameworks that might be of interest: http://www.fortpedro.com/blog/2015/1/observer-pattern-event-oriented-programming-in-python

kmike:
another option: https://pypi.python.org/pypi/blinker

eLRuLL:
I would like to work on this issue :smile:

Curita:
If anybody is currently working on this issue let us know, otherwise @eLRuLL: all yours!

kmike:
I've profiled a simple spider which downloads a page and follows all links from it using  LinkExractor. There is a non-standard CrawlerProcess which listens to all signals from its Crawlers and re-sends them through its own SignalManager to provide a single place to listen for all signals - nothing complicated. I haven't checked it with a standard CrawlerProcess.

For this spider signal handling (`SignalManager.send_catch_log function`) is one of the bottlenecks - it takes ~5 times more time than HTML parsing.

redapple:
See https://github.com/scrapy/scrapy/pull/2030

tifcty:
thanks to all of them

PR #2030, opened 2016-06-05
Moving signals away from PyDispatcher

rootAvish:
Fixes https://github.com/scrapy/scrapy/issues/8

This is by no means complete, specially as of the changes committed to this point,  most of the effort to this point dealt with porting relevant parts of `django.dispatch` to Scrapy and writing parts of it without dependency on `django.utils` and introducing methods for API consistency, but opening this PR so that the changes can be monitored by the community. I'll update this as the changes come along in detail about what the individual changes concern themselves with.

To this point we have:
- Introduced the `scrapy.dispatch` module
- Changed all standard signals to be objects of the `Signal` class instead of the generic `object`.

Most of the other stuff is outside the mainline and I'll be pushing it as soon as I'm done with it, I'll preferably have a working prototype by the end of this week(Saturday), latest by the start of the next one(the Monday after that). 

/cc @jdemaeyer


jdemaeyer:
Hey @rootAvish,

thanks for the WIP. You need to replace your imports to use either the full path or a relative one (`from dispatch import Signal` -> `from scrapy.dispatch import Signal` or `from .dispatch import Signal`, and a few others in the `dispatch` subpackage, preferably the absolute path version), you can easily see what I mean when you open a Python interpreter in the repositories base directory and simply try `import scrapy`.

Did you take a look at the tests?

rootAvish:
Right, I thought I had fixed that when I pushed this but I'm really bad at tracking what I have and haven't pushed yet. I'll be pushing a complete working prototype next(by Monday morning) though, hope everything passes at that point.

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,367 @@
...
+
+            sender
+                The registered sender to disconnect
+
+            dispatch_uid
+                the unique identifier of the receiver to disconnect
+        """
+        if weak is not None:
+            warnings.warn("Passing `weak` to disconnect has no effect.",
+                          ScrapyDeprecationWarning, stacklevel=2)
```
jdemaeyer: We can probably drop `weak` from the keyword arguments, no use in introducing already deprecated code. ;)

scrapy/utils/signal.py
```diff
@@ -5,75 +5,31 @@
...
 
-def disconnect_all(signal=Any, sender=Any):
+def disconnect_all(signal, sender=None):
     """Disconnect all signal handlers. Useful for cleaning up after running
     tests
     """
-    for receiver in liveReceivers(getAllReceivers(sender, signal)):
-        disconnect(receiver, signal=signal, sender=sender)
+    for receiver in signal._live_receivers(sender=sender):
+        signal.disconnect(receiver=receiver, sender=sender)
```
jdemaeyer: This should be a method of `Signal`, with only a pass-through here (like the two functions above)

scrapy/utils/signal.py
```diff
@@ -5,75 +5,31 @@
...
-        except Exception:
-            result = Failure()
-            logger.error("Error caught on signal handler: %(receiver)s",
-                         {'receiver': receiver},
-                         exc_info=True, extra={'spider': spider})
-        else:
-            result = response
-        responses.append((receiver, result))
-    return responses
+    return signal.send_robust(sender=sender, **named)
```
jdemaeyer: We should add `DeprecationWarning`s here and in the two functions below

scrapy/signalmanager.py
```diff
@@ -1,11 +1,10 @@
 from __future__ import absolute_import
-from pydispatch import dispatcher
 from scrapy.utils import signal as _signal
```
jdemaeyer: I wonder if we should drop this, directly call the `Signal` methods, and put deprecation methods into the util functions? They are pass-throughs now anyway.

jdemaeyer:
Great work Avishkar!

I've added a few smaller notes here and there but I think this looks pretty good now. Codecov complains that not all of the newly added code is being executed during the test suite, but that will probably be remedied when we add tests for the new signalling backend.

There are two backwards compatibility issues that we should think about:

1. All signal receivers must now accept `**kwargs`. This is a big one and will definitely break a lot of user code out there (just like it already broke a lot of internal code). In the old signalling backend, we were using [`pydispatcher`'s `robustApply` method](http://pydoc.net/Python/PyDispatcher/2.0.5/pydispatch.robustapply/) to handle receivers that weren't able to accept all keyword arguments. Can we port that to work with our new backend? And then throw a warning when we connect signal receivers that don't accept `**kwargs`, but no longer completely disallow them.

2. As you pointed out earlier, user-defined signals will be broken, because they now need to be instances of `Signal` instead of any arbitrary object. This one is much smaller than the first issue, since user-defined signals should be rare. I think we should try to implement an easy solution if there is one but be ready to let this stand as backwards incompatible if necessary. Given that all signalling tasks go through the `SignalManager`, maybe it's possible to do something like this:

```python
class SignalManager(object):
    _signal_proxies = {}

    def _ensure_signal(self, obj):
        if isinstance(obj, Signal):
            return obj
        # Warn user that this is deprecated here
        self._signal_proxies.setdefault(obj, Signal())
        return self._signal_proxies[obj]

    def connect(self, receiver, signal, **kwargs):
        signal = self._ensure_signal(signal)
        # ...

    # and similar in the other methods
```

I haven't put much thought into it though, issue 1 is definitely of higher priority.

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,367 @@
...
+            sender
+                The sender of the signal. Can be any python object (normally one
+                registered with a connect if you actually want something to
+                occur).
+
+            named
+                Named arguments which will be passed to receivers. These
+                arguments must be a subset of the argument names defined in
+                providing_args.
+        """
```
jdemaeyer: Tiny nitpick: This docstring should probably still say "Returns a list of tuple pairs [(receiver, response), ... ]."

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,367 @@
...
+            sender
+                The sender of the signal. Can be anything python object (normally one
+                registered with a connect if you actually want something to
+                occur).
+
+            named
+                Named arguments which will be passed to receivers. These
+                arguments must be a subset of the argument names defined in
+                providing_args.
+        """
```
jdemaeyer: Tiny nitpick: This docstring should probably say "Returns a Deferred that fires with a list of tuple pairs [(receiver, response), ... ]." or similar

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,367 @@
...
+                    del self.receivers[index]
+                    break
+            self.sender_receivers_cache.clear()
+        return disconnected
+
+    def has_listeners(self, sender=None):
+        return bool(self._live_receivers(sender))
+
+    def send(self, sender, **named):
+        """
```
jdemaeyer: It would be good if the docstrings were formatted in the same fashion as [other ones in `scrapy`](https://github.com/scrapy/scrapy/blob/master/scrapy/settings/__init__.py#L114), but you'll probably encounter this anyways when you do docs

scrapy/signalmanager.py
```diff
@@ -1,12 +1,17 @@
...
+from scrapy.dispatch.robustapply import robustApply
+from scrapy.dispatch.utils import func_accepts_kwargs
 
 
 class SignalManager(object):
 
-    def __init__(self, sender=dispatcher.Anonymous):
+    def __init__(self, sender=None):
         self.sender = sender
+        self.signals = {}
```
jdemaeyer: Why is this dictionary called `signals` when it actually contains (patched) receivers?

scrapy/signalmanager.py
```diff
@@ -23,7 +28,16 @@ def connect(self, receiver, signal, **kwargs):
         :type signal: object
         """
         kwargs.setdefault('sender', self.sender)
-        return dispatcher.connect(receiver, signal, **kwargs)
+        if not func_accepts_kwargs(receiver):
+            warnings.warn("The use of handlers that don't accept "
+                          "**kwargs has been deprecated, plese refer "
+                          "to the Signals API documentation.",
+                          ScrapyDeprecationWarning, stacklevel=2)
+            self.signals[receiver.__repr__()] = \
```
jdemaeyer: Why are we using `receiver.__repr__()` as key and not simply `receiver`?

tests/test_pipeline_media.py
```diff
@@ -31,7 +31,7 @@ def setUp(self):
 
     def tearDown(self):
         for name, signal in vars(signals).items():
-            if not name.startswith('_'):
+            if not name.startswith('_') and name is not 'Signal':
```
jdemaeyer: Use `name != 'Signal'` here. The `is` operator does not check for equality, but for identity (i.e. "does this point to the very same position in memory?"). Using `is` when you mean `==` will sometimes work b/c CPython caches some string and number literals, but you can never be sure:

```python
In [1]: teststr = "hello"
In [2]: teststr is "hello"
Out[2]: True
In [3]: teststr += "2"
In [4]: teststr is "hello2"
Out[4]: False
In [5]: teststr == "hello2"
Out[5]: True
```

tests/test_engine.py
```diff
@@ -119,13 +118,13 @@ def run(self):
...
-        dispatcher.connect(self.stop, signals.engine_stopped)
+        signals.engine_stopped.connect(self.stop)
         return self.deferred
 
-    def stop(self):
+    def stop(self, **kw):
         self.port.stopListening()
         for name, signal in vars(signals).items():
-            if not name.startswith('_'):
+            if not name.startswith('_') and name is not 'Signal':
```
jdemaeyer: Use `name != 'Signal' here (see below)

tests/test_utils_signal.py
```diff
@@ -69,11 +69,11 @@ def _get_result(self, signal, *a, **kw):
...
-        dispatcher.connect(test_handler, test_signal)
+        test_signal = Signal()
+        test_handler = lambda **kw: defer.Deferred()
+        test_signal.connect(test_handler)
         with LogCapture() as l:
             send_catch_log(test_signal)
         self.assertEqual(len(l.records), 1)
         self.assertIn("Cannot return deferreds from signal handler", str(l))
-        dispatcher.disconnect(test_handler, test_signal)
+        test_signal.disconnect(test_handler)
```
jdemaeyer: Missing newline :)

tests/test_utils_signal.py
```diff
@@ -2,20 +2,20 @@
 from twisted.trial import unittest
 from twisted.python.failure import Failure
 from twisted.internet import defer, reactor
-from pydispatch import dispatcher
 
+from scrapy.dispatch import Signal
 from scrapy.utils.signal import send_catch_log, send_catch_log_deferred
 
 
 class SendCatchLogTest(unittest.TestCase):
```
jdemaeyer: This test should probably be moved to a new module `test_dispatcher` or something, since the real code no longer lives in `utils.signal`

scrapy/signalmanager.py
rootAvish: @jdemaeyer I read [here](http://stackoverflow.com/questions/11162201/is-it-always-faster-to-use-string-as-key-in-a-dict) that it's best to use string keys in a Python dictionary, so decided to use the string representation of the callable as key instead of a callable instance.

scrapy/dispatch/robustapply.py
```diff
@@ -17,24 +18,29 @@ def function( receiver ):
...
+    receiver, codeObject, startIndex = function(receiver)
+    acceptable = codeObject.co_varnames[startIndex+len(arguments):
+                                        codeObject.co_argcount]
     for name in codeObject.co_varnames[startIndex:startIndex+len(arguments)]:
-        if named.has_key( name ):
+        if name in named:
             raise TypeError(
-                """Argument %r specified both positionally and as a keyword for calling %r"""% (
+                """Argument %r specified both positionally and
+                as a keyword for calling %r""" % (
```
jdemaeyer: This string (as it will be printed to the user) contains the newline and all the blanks before "as a keyword":

    Argument %r specified both positionally and\n                as a keyword for calling %r.

You can use Python's magic string concatenation to avoid this:
```python
             raise TypeError(
                """Argument %r specified both positionally and """
                """as a keyword for calling %r""" % (
```

scrapy/signalmanager.py
```diff
@@ -3,15 +3,36 @@
...
+        # Ensure we got at least an old style signal
+        warnings.warn("Signals in scrapy are no longer instances of the basic "
+                      "python object but rather the Signal class defined in "
+                      "the scrapy.dispatch module. Please refer to the Signal "
+                      "API documentation.",
+                      ScrapyDeprecationWarning, stacklevel=3)
+        signal_name = signal.__repr__()
+        if signal_name not in self._signal_proxies:
+            self._signal_proxies.setdefault(signal_name, _Signal())
+        return self._signal_proxies[signal_name]
```
jdemaeyer: 1. There is a race condition here, where `signal_name` is not in `self._signal_proxies` at the time when the `if` condition is evaluated, but is already when the next line is evaluated (oh the perks of asynchronous programming). You're already doing the right thing by using `setdefault`, but then you can also skip the "check if already in dict" step.
2. Dictionary keys can be anything in Python, and more importantly, an object's `__repr__` is not necessarily unique (since it could potentially be overwritten and no longer contain the object memory address). You could avoid the `__repr__` detour and just use `signal` itself as key.

Lastly, `setdefault` has the cool quality of returning the value that already was or is now stored in the dictionary, so the four lines above (everything after the warning) could be simplified to a single one:
```python
        return self._signal_proxies.setdefault(signal, _Signal())
```

scrapy/signalmanager.py
```diff
@@ -3,15 +3,36 @@
 
 from scrapy.utils import signal as _signal
 from scrapy.exceptions import ScrapyDeprecationWarning
-from scrapy.dispatch.robustapply import robustApply
+from scrapy.dispatch.robustapply import robust_apply as _robust_apply
 from scrapy.dispatch.utils import func_accepts_kwargs
-
+from scrapy.dispatch import Signal as _Signal
```
jdemaeyer: Is there a reason for adding the leading underscores?

scrapy/signalmanager.py
```diff
@@ -3,15 +3,36 @@
...
+        standard python objects as signals. This method returns a Signal()
+        class instance and saves it to proxy the user defined signal.
+        """
+        if isinstance(signal, _Signal):
+            return signal
+        # Ensure we got at least an old style signal
+        warnings.warn("Signals in scrapy are no longer instances of the basic "
+                      "python object but rather the Signal class defined in "
+                      "the scrapy.dispatch module. Please refer to the Signal "
+                      "API documentation.",
```
jdemaeyer: It would be good if the warning were formulated as instruction instead of description, e.g. `Custom signals should be instances of scrapy.dispatch.Signal.`

scrapy/signalmanager.py
jdemaeyer: Sorry, I only read your [earlier comment](https://github.com/scrapy/scrapy/pull/2030#discussion_r71973524) just now. I would still argue strongly for using the instance itself.

* In broader terms, for the wisdom mentioned in the top answer in [the SO thread you mention](https://stackoverflow.com/questions/11162201/is-it-always-faster-to-use-string-as-key-in-a-dict) ("premature optimization is the root of all evil", a trap I fall for on a regular basis ;)).
* In this case in particular, because calculating an objects representation is a very expensive operation. Its costs vastly exceed the savings from using a string, making the `repr`-based dictionary almost 20 times slower:
  ```
In [17]: timeit.timeit('mydict[a]', 'a = object(); mydict = {a: True}')
Out[17]: 0.06446212899936654

In [18]: timeit.timeit('mydict[repr(a)]', 'a = object(); mydict = {repr(a): True}')
Out[18]: 1.168589578999672
    ```

scrapy/signalmanager.py
```diff
@@ -59,6 +82,7 @@ def send_catch_log(self, signal, **kwargs):
         through the :meth:`connect` method).
         """
         kwargs.setdefault('sender', self.sender)
+        signal = self._ensure_signal(signal)
         return _signal.send_catch_log(signal, **kwargs)
```
jdemaeyer: Is there a reason for proxying through `_signal.send_catch_log` here instead of directly using `signal.send_robust()`? Since they are all just "pass-through functions" now, I think we can keep the `utils.signal` functions only for backwards compatibility but stop using them throughout scrapy's source. Or do you think this will "bind us" to the new Signal API too much? /cc @kmike

scrapy/dispatch/utils/robustapply.py
```diff
@@ -0,0 +1,59 @@
...
+        # fc does not have a **kwds type parameter, therefore
+        # remove unacceptable arguments.
+        if six.PY2:
+            for arg in named.keys():
+                if arg not in acceptable:
+                    del named[arg]
+        else:
+            for arg in list(named):
+                if arg not in acceptable:
+                    del named[arg]
```
jdemaeyer: If I'm not mistaken this code should be suitable for Python 2 as well.

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,368 @@
...
+
+        # If DEBUG is on, check that we got a good receiver
+        settings = get_project_settings()
+        if settings.get('LOG_ENABLED') \
+                and settings.get('LOG_LEVEL') == 'DEBUG':
+            assert callable(receiver), "Signal receivers must be callable."
+            # Check for **kwargs
+            if not func_accepts_kwargs(receiver):
+               raise ValueError(
+                   "Signal receivers must accept keyword arguments(**kwargs).")
```
jdemaeyer: I think we should drop the "if debug" switch here and make this a warning instead of an exception. We have a working backwards-compatible mechanism in place and I see no reason not to use it just because we're using debug level logging.

tests/test_signalmanager_compatability.py
```diff
@@ -0,0 +1,40 @@
...
+
+def receiver(**kwargs):
+    pass
+
+
+def receiver_no_kwargs():
+    pass
+
+
+class BackwardCompatabilityTest(unittest.TestCase):
```
jdemaeyer: `Compatibility` :)

tests/test_signalmanager_compatability.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def setUp(self):
+        self.signals = SignalManager()
+
+    def tearDown(self):
+        self.assertFalse(self.signals._patched_receivers)
+        gc.collect()
```
jdemaeyer: Could you add a comment why this is necessary?

scrapy/signalmanager.py
```diff
@@ -1,12 +1,39 @@
 from __future__ import absolute_import
-from pydispatch import dispatcher
+import warnings
+
 from scrapy.utils import signal as _signal
+from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.dispatch.utils import robust_apply as _robust_apply
+from scrapy.dispatch.utils import func_accepts_kwargs
+from scrapy.dispatch import Signal as _Signal
```
jdemaeyer: (Reposting this here because I accidentally attached it to an old commit ;)) Is there a reason to add the leading underscore?

tests/test_dispatch_utils.py
```diff
@@ -1,16 +1,76 @@
...
+        def accept_kwargs(**kwargs):
+            pass
+
+        def no_kwargs():
+            pass
+        # To test the attribute error case
+        someval = 0
+        self.assertTrue(func_accepts_kwargs(accept_kwargs))
+        # TODO: The fallback here is that it is assumed that the "receiver"
+        # is callable. Correct behaviour?
```
jdemaeyer: I think we should raise a `TypeError` when the given "function" is not a callable. From the docs:

> Raised when an operation or function is applied to an object of inappropriate type.

jdemaeyer: Oops, just saw that you already did that, nice!

scrapy/dispatch/utils/robustapply.py
```diff
@@ -0,0 +1,59 @@
...
+        if hasattr(receiver.__call__, '__func__') \
+                or hasattr(receiver.__call__, '__code__'):
+            receiver = receiver.__call__
+    if hasattr(receiver, '__func__'):
+        # an instance-method...
+        return receiver, receiver.__code__, 1
+    elif not hasattr(receiver, '__code__'):
+        raise ValueError('unknown reciever type %s %s' %
+                         (receiver, type(receiver)))
+    return receiver, receiver.__code__, 0
```
jdemaeyer: I know this is from Django and I know I'm being pedantic, but could you change `0` into `False` and `1` into `True` above? ;)

scrapy/signalmanager.py
```diff
@@ -23,7 +50,18 @@ def connect(self, receiver, signal, **kwargs):
...
-        return dispatcher.connect(receiver, signal, **kwargs)
+        signal = self._ensure_signal(signal)
+        if not func_accepts_kwargs(receiver):
+            warnings.warn("The use of handlers that don't accept "
+                          "**kwargs has been deprecated, plese refer "
+                          "to the Signals API documentation.",
+                          ScrapyDeprecationWarning, stacklevel=2)
+            if receiver.__repr__() not in self._patched_receivers:
+                self._patched_receivers[receiver.__repr__()] = \
+                    lambda sender, **kw: _robust_apply(receiver, sender, **kw)
```
jdemaeyer: Hmm, I think our solution here breaks the `weak=True` keyword argument functionality (since we're always keeping a strong reference in our `_patched_receivers` dictionary). Maybe it's enough to just mention this in the docs. (Since `weak=True` is a new feature and came together with requiring `**kwargs`.)

scrapy/signalmanager.py
```diff
@@ -23,7 +50,18 @@ def connect(self, receiver, signal, **kwargs):
...
         """
         kwargs.setdefault('sender', self.sender)
-        return dispatcher.connect(receiver, signal, **kwargs)
+        signal = self._ensure_signal(signal)
+        if not func_accepts_kwargs(receiver):
+            warnings.warn("The use of handlers that don't accept "
+                          "**kwargs has been deprecated, plese refer "
+                          "to the Signals API documentation.",
+                          ScrapyDeprecationWarning, stacklevel=2)
+            if receiver.__repr__() not in self._patched_receivers:
```
jdemaeyer: Nice catch!

tests/test_signalmanager_compatability.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertFalse(self.signals.disconnect(receiver_no_kwargs,
+                                                    new_signal))
+        del self.signals._patched_receivers[receiver_no_kwargs.__repr__()]
+        self.assertFalse(self.signals.disconnect(receiver_no_kwargs,
+                                                    new_signal))
+
+    def test_disconnect_all(self):
+        new_signal = object()
+        self.signals.connect(receiver, new_signal)
+        self.signals.disconnect_all(new_signal)
```
jdemaeyer: Looks like this test makes no assertions

scrapy/dispatch/utils/robustapply.py
```diff
@@ -0,0 +1,59 @@
...
+    """Call receiver with arguments and an appropriate subset of named
+    """
+    receiver, codeObject, startIndex = function(receiver)
+    acceptable = codeObject.co_varnames[startIndex+len(arguments):
+                                        codeObject.co_argcount]
+    for name in codeObject.co_varnames[startIndex:startIndex+len(arguments)]:
+        if name in named:
+            raise TypeError(
+                """Argument %r specified both positionally and
+                as a keyword for calling %r""" % (
```
jdemaeyer: (Copying this here cause I accidentally attached it to and old commit earlier)

This string (as it will be printed to the user) contains the newline and all the blanks before "as a keyword":

    Argument %r specified both positionally and\n                as a keyword for calling %r.

You can use Python's magic string concatenation to avoid this:
```python
             raise TypeError(
                """Argument %r specified both positionally and """
                """as a keyword for calling %r""" % (
```

jdemaeyer:
Hey @rootAvish :) I've added my usual nitpicks here and there but I am quite happy with how this PR looks. Great work!

It would be cool if you could squash the commits into a couple of "meaningful" ones. Then when we have settled the remaining smaller issues (the line comments) and the benchmarks are up somewhere I think we're ready to have a second reviewer look over it and add some docs (or the other way around).

rootAvish:
Hey @jdemaeyer , thanks for the code review. In accordance with the email I sent you, the patched receiver sending was abysmally slow when compared to the current API, and in the spirit of always coming up with overly complicated solutions, this too was due to over engineering on my part. I introduced a large amount of method call overhead as well, which was taking a lot of time too. I reworked that and completely got rid of `_patched_receivers` and instead worked using the `Signal` class itself, IMO the new approach is much simpler and cleaner, and the benchmark too is fixed(in case you didn't check the email yet, here's a link to the benchmark suite: https://github.com/rootAvish/scrapysignalbench).

Would appreciate it if you had a look over that part again, do you see any problems with the new implementation?

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,386 @@
...
+                anything hashable.
+        """
+        assert callable(receiver), "Signal receivers must be callable."
+        # Check for **kwargs
+        if not func_accepts_kwargs(receiver):
+           warnings.warn("The use of handlers that don't accept "
+                          "**kwargs has been deprecated, plese refer "
+                          "to the Signals API documentation.",
+                          ScrapyDeprecationWarning, stacklevel=3)
+           self.receiver_accepts_kwargs[_make_id(receiver)] = False
```
jdemaeyer: The indentation is a space short in this block ;)

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,386 @@
...
+            for index in range(len(self.receivers)):
+                (r_key, _) = self.receivers[index]
+                if r_key == lookup_key:
+                    disconnected = True
+                    del self.receivers[index]
+                    break
+            self.sender_receivers_cache.clear()
+            if disconnected:
+                if _make_id(receiver) in self.receiver_accepts_kwargs:
+                    del self.receiver_accepts_kwargs[_make_id(receiver)]
```
jdemaeyer: There's a race condition here, you can use `self.receiver_accepts_kwargs.pop(_make_id(receiver), None)` to get rid of it

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        responses = []
+        if not self.receivers or self.sender_receivers_cache.get(sender) is NO_RECEIVERS:
+            return responses
+
+        for receiver in self._live_receivers(sender):
+            if self.receiver_accepts_kwargs[_make_id(receiver)]:
+                response = receiver(signal=self, sender=sender, **named)
+            else:
+                response = robust_apply(receiver, sender=sender, **named)
```
jdemaeyer: I wonder why we don't pass `signal=self` here?

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,386 @@
...
+        d.addCallback(lambda out: [x[1] for x in out])
+        return d
+
+    def _clear_dead_receivers(self):
+        # Note: caller is assumed to hold self.lock.
+        if self._dead_receivers:
+            self._dead_receivers = False
+            new_receivers = []
+            for r in self.receivers:
+                if isinstance(r[1], weakref.ReferenceType) and r[1]() is not None:
```
jdemaeyer: This `for` loop used to contain:
```python
                if isinstance(r[1], weakref.ReferenceType) and r[1]() is None:
                    continue
                new_receivers.append(r)
```
(i.e. everything that was not a dead weakref was added to `new_receivers`) With the changes you made it looks as if `new_receivers` will almost always be empty, except for live weakrefs. You probably meant to do this, right?
```python
                if not (isinstance(r[1], weakref.ReferenceType) and r[1]() is None):
```

scrapy/signalmanager.py
```diff
@@ -1,12 +1,32 @@
 from __future__ import absolute_import
-from pydispatch import dispatcher
+import warnings
+
 from scrapy.utils import signal as _signal
+from scrapy.dispatch import Signal as Signal
```
jdemaeyer: No need for the `as Signal` ;)

jdemaeyer:
Heyhey Avishkar, nice work! I agree, this solution is quite a bit simpler, and I like how you restored the `weak=True` functionality. As usual I've added a few small comments (at one of these the code is actually broken I think), and there's still three that are open from earlier commits. (Btw these are never commands or anything, feel free to answer with why you disagree.)

So this leaves us with the following stuff still to do, right?
  * smaller issues (line comments)
  * squash commits
  * finish docs
  * get second review

If you find time for it, maybe you can put the output for the benchmark suite somewhere (just a raw paste of the command line output should be fine)?

rootAvish:
@jdemaeyer thanks for responding on such short notice, always appreciate your insights.

>  there's still three that are open from earlier commits. (Btw these are never commands or anything, feel free to answer with why you disagree.)

Sorry for missing out on those, I don't disagree, I thought you wanted to deliberate further on deprecating `utils.signal` further so I didn't move on it at the time, I thought it would be too much of a change but then I considered that the utils pass-through methods are not directly usable after the change anyway, plus it also gets rid of the method call overhead.

> So this leaves us with the following stuff still to do, right?
> smaller issues (line comments)  

I think I took care of all of them all (I think, I'll make sure to go over all of them again to make sure I didn't miss any like before).

> squash commits  

Right, I wanted to do that once I was done with the development.

> finish docs

I'm almost done with those, I'll upload them sometime later today. I did the docstrings for `dispatcher.py` as a starting point, I tried to adhere to the formatting style used in Scrapy.

> If you find time for it, maybe you can put the output for the benchmark suite somewhere (just a raw paste of the command line output should be fine)?

Hey sure, I thought you would like to run the same on your end for verification so I didn't include my run at first, here's a sample: https://github.com/rootAvish/scrapysignalbench/blob/master/scrapysignalbench/benchmarks/README.md

As you can see, in theory connection is slower than before but most of that time is used up in checking for kwargs and raising the deprecation warning(line_profiler output that I went by in the past suggest so), in time further down the road we can always look to migrate away from non-kwargs receivers completely. Do you think slower connection times an issue?

rootAvish:
Hey @jdemaeyer , did you guys change the openSSL version or something? I don't think that break with Python 3.3 has anything to do with the code. Also, please review the docs when you get a chance. I hope the commits are as desired now.

redapple:
Hi @rootAvish , we did not change anything explicitly, but newer packages were used:
If I compare python packages from the [last PASSING Travis build](https://travis-ci.org/scrapy/scrapy/jobs/154405979)
and [this FAILING one](https://travis-ci.org/scrapy/scrapy/jobs/157872588), this is what you get:

```
diff freeze.passing.txt freeze.failing.txt 
1c1
< PASSING
---
> FAILING
3c3
< attrs==16.0.0
---
> attrs==16.1.0
5c5
< botocore==1.4.48
---
> botocore==1.4.53
7c7
< cffi==1.7.0
---
> cffi==1.8.2
9c9
< cryptography==1.4
---
> cryptography==1.5
11c11
< curtsies==0.2.6
---
> curtsies==0.2.9
35c35
< pyOpenSSL==16.0.0
---
> pyOpenSSL==16.1.0
46,48c46,48
< testfixtures==4.10.0
< traitlets==4.2.2
< Twisted==16.3.2
---
> testfixtures==4.10.1
> traitlets==4.3.0
> Twisted==16.4.0
51c51
< zope.interface==4.2.0
---
> zope.interface==4.3.2
```

redapple:
This recent build for the master branch passes and has the same packages: https://travis-ci.org/scrapy/scrapy/jobs/159358737

redapple:
hey @rootAvish , I don't know why, but a new PR of mine with _only_ your commits, rebased on current "master" branch, built fine, including on Python 3.3
https://github.com/scrapy/scrapy/issues/2235
https://travis-ci.org/scrapy/scrapy/builds/159620850

docs/topics/signals.rst
```diff
@@ -13,9 +13,20 @@ Even though signals provide several arguments, the handlers that catch them
 don't need to accept all of them - the signal dispatching mechanism will only
 deliver the arguments that the handler receives.
 
+Signals underwent a major refactoring with an eye towards speed. Although still
+backwards compatible, receivers now need to have a ``**kwargs`` argument i.e. all
+receivers should now accept a variable keyword args param. We request that you
+make sure all your receivers follow this contract.
```
kmike: It means this change is backwards incompatible, i.e. users need to change all their receivers.

kmike: Or is this optional?

scrapy/dispatch/utils/inspect.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import absolute_import
+
+import inspect
+
+import six
+
+
+def func_accepts_kwargs(func):
+        # Not all callables are inspectable with getargspec, so we'll
```
kmike: indentation is off

scrapy/dispatch/dispatcher.py
```diff
@@ -0,0 +1,390 @@
...
+            lookup_key = (_make_id(receiver), _make_id(sender))
+        accepts_kwargs_lookup = _make_id(receiver)
+        if weak:
+            ref = weakref.ref
+            receiver_object = receiver
+            # Check for bound methods
+            if hasattr(receiver, '__self__') and hasattr(receiver, '__func__'):
+                ref = WeakMethod
+                receiver_object = receiver.__self__
+            if six.PY3 and sys.version_info >= (3, 4):
```
kmike: why is six.PY3 check needed? It may return False in Python 4.x, and it looks redundant because of sys.version_info check.

scrapy/dispatch/weakref_backports.py
```diff
@@ -0,0 +1,67 @@
+"""
+weakref_backports is a partial backport of the weakref module for python
+versions below 3.4.
```
kmike: Can we use https://pypi.python.org/pypi/backports.weakref?

scrapy/utils/signal.py
```diff
@@ -1,79 +1,41 @@
...
-        d.addBoth(lambda result: (receiver, result))
-        dfds.append(d)
-    d = DeferredList(dfds)
-    d.addCallback(lambda out: [x[1] for x in out])
-    return d
+    return signal.send_catch_log_deferred(sender=sender, **named)
 
 
-def disconnect_all(signal=Any, sender=Any):
+def disconnect_all(signal, sender=None):
```
kmike: Previously signal argument was not required; also, it was posisble to pass object instead of Signal instance.

scrapy/utils/signal.py
```diff
@@ -1,79 +1,41 @@
...
-    for receiver in liveReceivers(getAllReceivers(sender, signal)):
-        d = maybeDeferred(robustApply, receiver, signal=signal, sender=sender,
-                *arguments, **named)
-        d.addErrback(logerror, receiver)
-        d.addBoth(lambda result: (receiver, result))
-        dfds.append(d)
-    d = DeferredList(dfds)
-    d.addCallback(lambda out: [x[1] for x in out])
-    return d
+    return signal.send_catch_log_deferred(sender=sender, **named)
```
kmike: signal was optional previously, and it was not required to use Signal class

docs/topics/signals.rst
rootAvish: @kmike It's optional as we did not want to break backward compatibility, but those receivers would not fully leverage the full potential of the speedup as `robust_apply` would still be invoked.

scrapy/utils/signal.py
rootAvish: @kmike The assumption here was that all signals would have been connected using the `SignalManager` class and all compatibility was implemented either there, or inside `Signal` itself, while deprecating the use of utils.signal, but I see how that can be a problem. I could always implement a layer of compatibility for `utils.signal`. I'll get on that.

rootAvish:
@redapple I'm unable to make out what the problem is at first glance, but using your build I'm sure I can figure that part out. I'll report back as soon I have something. Thanks for the pointers.

scrapy/utils/signal.py
rootAvish: @kmike , sorry I just remembered what my line of thought here was. The signals here contain just the `send_catch_log`, `send_catch_log_deferred` and `disconnect_all` methods. Assuming somebody was using `scrapy.utils.signal` directly instead of `SignalManager`, they most likely used `dispatcher.connect` from `PyDispatcher` directly to connect their signal to the receiver. If that be the case, then they are screwed(mind my language please) anyway since the entire point here is to move signals away fro `PyDispatcher`.

Assuming they were using SignalManager, that does require a `signal` as a compulsory argument, and they would use the methods from there itself, which does not use `utils.signal` anymore. I don't think implementing any kind of compatibility here would serve us a purpose. But on that note, I'm not sure what use case this module serves anymore. Your thoughts on the same?

scrapy/utils/signal.py
rootAvish: This too would depend on my what the decision is regarding my other comment here: https://github.com/scrapy/scrapy/pull/2030#discussion-diff-78588791R34.

scrapy/dispatch/weakref_backports.py
rootAvish: I glanced through the code in its repo here: https://github.com/pjdelport/backports.weakref/blob/master/src/backports/weakref.py and it looks like he has implemented the backport himself. Ours is a verbatim copy of the code in the Python 3.4 source. I can do a build with that module but it's your call in the end on which implementation of the backport you'd like to go with.

rootAvish: I'm sorry, as usual I spoke too soon. That implementation backports `weakref.finalize`. We ported `weakref.WeakMethod`. In the current state of that package we would require both this and that package.

kmike: Aha, makes sense. Maybe we should raise an issue at https://github.com/pjdelport/backports.weakref, or even contribute a PR with WeakMethod changes. Less code we have in Scrapy the better :) What do you think?

scrapy/utils/signal.py
kmike: Hm, in this case it seems we may drop the whole module, or just keep the current implementation as-is and deprecate it.

rootAvish:
Hey @redapple , sorry for not replying all this time but just wanted to give you an update. I haven't been able to solve this issue yet since the only place I was able to even reproduce the problem was on a Ubuntu 14.04 installation and on there it seems like this is an error with OpenSSL itself (scrapy-master breaks too).

```shell
[20:31:25] avishkar:Downloads $ python3.3                                                                                                         
Python 3.3.6 (default, Jan 28 2015, 17:27:09) 
[GCC 4.8.2] on linux
>>> import OpenSSL
  File "/usr/local/lib/python3.3/dist-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import rand, crypto, SSL
  File "/usr/local/lib/python3.3/dist-packages/OpenSSL/rand.py", line 12, in <module>
    from OpenSSL._util import (
  File "/usr/local/lib/python3.3/dist-packages/OpenSSL/_util.py", line 6, in <module>
    from cryptography.hazmat.bindings.openssl.binding import Binding
  File "/usr/local/lib/python3.3/dist-packages/cryptography/hazmat/bindings/openssl/binding.py", line 14, in <module>
    from cryptography.hazmat.bindings._openssl import ffi, lib
ImportError: No module named 'cryptography.hazmat.bindings._openssl'
>>> 
```

redapple:
@rootAvish , sorry for the late feedback.
I just tried again submitting a master-branch-rebased version of your commits, and it built ok: https://github.com/scrapy/scrapy/pull/2361

Not sure what issue Travis is having to be honest.

But perhaps you can try doing the same thing: open a new PR, in top of master-branch, and continue/finish-up the work there (if there's anything missing)

rootAvish:
@redapple Right, I'm anyway out of ideas on what else I can try here.

rootAvish:
@redapple, whatever issue Travis was having seems to be fixed now, waiting for @kmike 's comments on the usage of the weakrefmethod package, all the other stuff is taken care of. Sorry for the delay, was in a crunch at work.

rootAvish:
Hi,

Just wondering what happened to this, is this project still a go?

redapple:
Hey @rootAvish , sorry for the late reply.
The last time I discussed this with other maintainers, there were some doubts on whether the team/project wants to maintain the new dispatcher code vs. using a seperate package such as blinker (or gossip, that I'm discovering today).
There's a [benchmarking project for GSoC 2017](https://github.com/scrapy/scrapy-bench) that could also help compare options regarding performance gains.

redapple:
As a side note, @lopuhin forked PyDispatcher to add PyPy support: https://github.com/scrapy/pypydispatcher

rootAvish:
@redapple Has anybody done a POC using something else? I had written benchmarks for signalling back when I did this: https://github.com/rootavish/scrapysignalbench, I can try doing a performance comparison linking against blinker. Shouldn't be too hard.

redapple:
@rootAvish , I don't think there's any PoC using blinker or gossip.
This benchmarking may interest @Parth-Vader and @lopuhin

Issue #1965, opened 2016-05-02
Add option to set JSON output encoding


redapple:
Some users may prefer UTF-8 encoding of characters when using JSON output, instead of the the default `\UXXXX` espace sequences. See https://github.com/scrapy/scrapy/issues/1963#issuecomment-215943619.

The option to turn off `ensure_ascii` and/or output encoding for `JsonItemExporter` could be interesting.

mgachhui:
The options method @wi-cuckoo mentioned in #1963 seems doable. What if we extend the crawler by adding an encoding option as -e? As far as I can tell, it'll make sense for XML and CSV as well, in addition to JSON.

kmike:
Fixed by https://github.com/scrapy/scrapy/pull/2034.

PR #2034, opened 2016-06-06
[MRG+1] Added option to turn off ensure_ascii for JSON exporters

dracony:
See #1965

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -207,6 +207,7 @@ These are the settings used for configuring the feed exports:
  * :setting:`FEED_STORAGES`
  * :setting:`FEED_EXPORTERS`
  * :setting:`FEED_STORE_EMPTY`
+ * :setting:`FEED_EXPORT_ENCODING`
  * :setting:`FEED_EXPORT_FIELDS`
 
 .. currentmodule:: scrapy.extensions.feedexport
@@ -230,6 +231,20 @@ FEED_FORMAT
 
 The serialization format to be used for the feed. See
 :ref:`topics-feed-format` for possible values.
+
+.. setting:: FEED_EXPORT_ENCODING
+
+FEED_EXPORT_ENCODING
+--------------------
+
+Default: ``None``
+
+The encoding to be used for the feed.
+
+If unset or set to ``None`` (default) it uses UTF-8 for everything except JSON output,
+which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
+
+Use ``utf-8`` if you want UTF-8 for JSON too.
 
 .. setting:: FEED_EXPORT_FIELDS
 
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -33,9 +33,9 @@ class BaseItemExporter(object):
         If dont_fail is set, it won't raise an exception on unexpected options
         (useful for using with keyword arguments in subclasses constructors)
         """
+        self.encoding = options.pop('encoding', None)
         self.fields_to_export = options.pop('fields_to_export', None)
         self.export_empty_fields = options.pop('export_empty_fields', False)
-        self.encoding = options.pop('encoding', 'utf-8')
         if not dont_fail and options:
             raise TypeError("Unexpected options: %s" % ', '.join(options.keys()))
 
@@ -84,11 +84,13 @@ class JsonLinesItemExporter(BaseItemExporter):
     def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
+        kwargs.setdefault('ensure_ascii', not self.encoding)
         self.encoder = ScrapyJSONEncoder(**kwargs)
 
     def export_item(self, item):
         itemdict = dict(self._get_serialized_fields(item))
-        self.file.write(to_bytes(self.encoder.encode(itemdict) + '\n'))
+        data = self.encoder.encode(itemdict) + '\n'
+        self.file.write(to_bytes(data, self.encoding))
 
 
 class JsonItemExporter(BaseItemExporter):
@@ -96,6 +98,7 @@ class JsonItemExporter(BaseItemExporter):
     def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
+        kwargs.setdefault('ensure_ascii', not self.encoding)
         self.encoder = ScrapyJSONEncoder(**kwargs)
         self.first_item = True
 
@@ -111,7 +114,8 @@ class JsonItemExporter(BaseItemExporter):
         else:
             self.file.write(b',\n')
         itemdict = dict(self._get_serialized_fields(item))
-        self.file.write(to_bytes(self.encoder.encode(itemdict)))
+        data = self.encoder.encode(itemdict)
+        self.file.write(to_bytes(data, self.encoding))
 
 
 class XmlItemExporter(BaseItemExporter):
@@ -120,6 +124,8 @@ class XmlItemExporter(BaseItemExporter):
         self.item_element = kwargs.pop('item_element', 'item')
         self.root_element = kwargs.pop('root_element', 'items')
         self._configure(kwargs)
+        if not self.encoding:
+            self.encoding = 'utf-8'
         self.xg = XMLGenerator(file, encoding=self.encoding)
 
     def start_exporting(self):
@@ -169,9 +175,16 @@ class CsvItemExporter(BaseItemExporter):
 
     def __init__(self, file, include_headers_line=True, join_multivalued=',', **kwargs):
         self._configure(kwargs, dont_fail=True)
+        if not self.encoding:
+            self.encoding = 'utf-8'
         self.include_headers_line = include_headers_line
-        file = file if six.PY2 else io.TextIOWrapper(file, line_buffering=True)
-        self.csv_writer = csv.writer(file, **kwargs)
+        self.stream = io.TextIOWrapper(
+            file,
+            line_buffering=False,
+            write_through=True,
+            encoding=self.encoding
+        ) if six.PY3 else file
+        self.csv_writer = csv.writer(self.stream, **kwargs)
         self._headers_not_written = True
         self._join_multivalued = join_multivalued
 
@@ -200,7 +213,7 @@ class CsvItemExporter(BaseItemExporter):
     def _build_row(self, values):
         for s in values:
             try:
-                yield to_native_str(s)
+                yield to_native_str(s, self.encoding)
             except TypeError:
                 yield s
 
@@ -263,6 +276,8 @@ class PythonItemExporter(BaseItemExporter):
             warnings.warn(
                 "PythonItemExporter will drop support for binary export in the future",
                 ScrapyDeprecationWarning)
+        if not self.encoding:
+            self.encoding = 'utf-8'
 
     def serialize_field(self, field, name, value):
         serializer = field.get('serializer', self._serialize_value)
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -162,6 +162,7 @@ class FeedExporter(object):
         if not self.urifmt:
             raise NotConfigured
         self.format = settings['FEED_FORMAT'].lower()
+        self.export_encoding = settings['FEED_EXPORT_ENCODING']
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
         if not self._storage_supported(self.urifmt):
@@ -185,7 +186,8 @@ class FeedExporter(object):
         uri = self.urifmt % self._get_uri_params(spider)
         storage = self._get_storage(uri)
         file = storage.open(spider)
-        exporter = self._get_exporter(file, fields_to_export=self.export_fields)
+        exporter = self._get_exporter(file, fields_to_export=self.export_fields,
+            encoding=self.export_encoding)
         exporter.start_exporting()
         self.slot = SpiderSlot(file, exporter, storage, uri)
 
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -139,6 +139,7 @@ FEED_URI = None
 FEED_URI_PARAMS = None  # a function to extend uri arguments
 FEED_FORMAT = 'jsonlines'
 FEED_STORE_EMPTY = False
+FEED_EXPORT_ENCODING = None
 FEED_EXPORT_FIELDS = None
 FEED_STORAGES = {}
 FEED_STORAGES_BASE = {
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -371,3 +371,32 @@ class FeedExportTest(unittest.TestCase):
             ]
             yield self.assertExported(items, ['egg', 'baz'], rows,
                                       settings=settings, ordered=True)
+
+    @defer.inlineCallbacks
+    def test_export_encoding(self):
+        items = [dict({'foo': u'Test\xd6'})]
+        header = ['foo']
+        
+        formats = {
+            'json': u'[\n{"foo": "Test\\u00d6"}\n]'.encode('utf-8'),
+            'jsonlines': u'{"foo": "Test\\u00d6"}\n'.encode('utf-8'),
+            'xml': u'<?xml version="1.0" encoding="utf-8"?>\n<items><item><foo>Test\xd6</foo></item></items>'.encode('utf-8'),
+            'csv': u'foo\r\nTest\xd6\r\n'.encode('utf-8'),
+        }
+        
+        for format in formats:
+            settings = {'FEED_FORMAT': format}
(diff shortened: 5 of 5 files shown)


dracony:
For some reason I just can't get the tests to pass on Py3, so annoying...
And I don't have a local install of it atm, so I guess I'll be torturing Travis for a while)

dracony:
Fixed it finally )

dracony:
Any feedback for it? Would really love to finally have this in my scraper )

redapple:
@dracony , thanks for this.
how will people use this? use their own JSON exporter subclass?
if yes, how about introducing a setting for this? it could be simpler.
Also, [@mgachhui was suggesting](https://github.com/scrapy/scrapy/issues/1965#issuecomment-218077754) having an encoding parameter, not only default UTF-8
(and applying this to XML exports too)

tests/test_exporters.py
```diff
@@ -427,6 +427,21 @@ def test_nonstring_types_item(self):
...
         self.assertEqual(exported, item)
 
+    def test_unicode_ascii(self):
+        i1 = TestItem(name=u'Test\u2603')
+        self.assertExportResult(i1, b'{"name": "Test\\u2603"}\n')
+
+    def test_unicode_utf8(self):
+        self.ie = JsonLinesItemExporter(self.output, ensure_ascii = False)
+        i1 = TestItem(name=u'Test\u2603')
+        self.assertExportResult(i1, b'{"name": "Test\xe2\x98\x83"}\n')
```
redapple: nitpick: I find it a bit easier to read a test against `u'Test\u2603'.encode('utf8')`

scrapy/exporters.py
```diff
@@ -96,9 +100,13 @@ class JsonItemExporter(BaseItemExporter):
     def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
-        self.encoder = ScrapyJSONEncoder(**kwargs)
+        self.encoder = ScrapyJSONEncoder(ensure_ascii=self.ensure_ascii, **kwargs)
         self.first_item = True
 
+    def _configure(self, options, dont_fail=False):
+        self.ensure_ascii = options.pop('ensure_ascii', True)
```
dracony: I added check for ensure_ascii from the options, so you would not have to subclass the exporter, just pass the option to it.

Setting would probably be a better option though. Kk, I'll add that one too.

As for the encoding parameter: JSON encode accepts separate ensure_ascii and encoding parameters. Would you like them to be separate like that? Or turn of ensure_ascii automatically if the encoding is specified?

redapple: Sorry, I meant that you'd need to pass the option [when the feedexporter initializes exporters](https://github.com/scrapy/scrapy/blob/bf7f67549378269c3976afc89abcf9c2190d242f/scrapy/extensions/feedexport.py#L188).
Or maybe I'm missing some bit?
I'm still new to exporters code

dracony: Yup, I got that now. I'm using a custom feedexporter, so I could pass ensure_ascii option myself.
I'll fix it later today =)

redapple: Regarding feed encoding setting,
I believe we could to use `None` as default feed encoding,

* for JSON that would mean `ensure_ascii=True`, i.e. `\uXXXX` escape sequences
* for XML and CSV, that would mean UTF-8

if feed encoding set to UTF-8 explicitly:

* for JSON export, that would mean `ensure_ascii=False` and UTF-8 encoding
* for XML and CSV, that would change from default `None` setting

For other encodings, data is written with the supplied encoding, with warnings or exceptions for failing item encoding writes.
what do you think?

dracony: Sounds good

dracony:
And here is the FEED_EXPORT_ENCODING setting =)
Took me a while, I so hate all that encoding stuff))

dracony:
Ah damn it, The CSV tests fail on Py3 because StringIO is behaving differently =(
Any idea how to fix this?

dracony:
@redapple Yay, I made it work. 
Like it more now?

scrapy/exporters.py
```diff
@@ -33,9 +34,9 @@ def _configure(self, options, dont_fail=False):
         If dont_fail is set, it won't raise an exception on unexpected options
         (useful for using with keyword arguments in subclasses constructors)
         """
+        self.encoding=options.pop('encoding', None)
```
robsonpeixoto: Code style

dracony:
@robsonpeixoto Thanks, missed those. I amended the commit to pep8-ize it =)

dracony:
Any other feedback on it? Would really be great to have it merged (I work with quite a lot of cp-1251 encoded stuff)

scrapy/exporters.py
```diff
@@ -169,9 +180,16 @@ class CsvItemExporter(BaseItemExporter):
 
     def __init__(self, file, include_headers_line=True, join_multivalued=',', **kwargs):
         self._configure(kwargs, dont_fail=True)
+        if not self.encoding:
+            self.encoding = 'utf-8'
         self.include_headers_line = include_headers_line
-        file = file if six.PY2 else io.TextIOWrapper(file, line_buffering=True)
-        self.csv_writer = csv.writer(file, **kwargs)
+        self.stream = six.StringIO()
```
redapple: duplicated on line 189 below

scrapy/exporters.py
```diff
@@ -214,7 +232,16 @@ def _write_headers_and_set_fields_to_export(self, item):
                     # use fields declared in Item
                     self.fields_to_export = list(item.fields.keys())
             row = list(self._build_row(self.fields_to_export))
-            self.csv_writer.writerow(row)
+            self._writerow(row)
+
+    def _writerow(self, row):
+        self.csv_writer.writerow(row)
+        if six.PY2:
+            data = self.stream.getvalue()
```
redapple: why are these lines below needed?

dracony: Python 3 can make use of the TextIOWrapper to write in particular encoding. To write csv files in a specified encoding in Python 2 you need to rely on StringIO, as actually shown in the example in the docs: https://docs.python.org/2/library/csv.html

So basically in python2 the csv writer will write a utf string to StringIO buffer, and then the result will be reencoded to the the specified encoding and written to file

scrapy/exporters.py
dracony: Yup, I'll remove that

dracony:
If you know of any other way to make it work on py2 and 3 id be happy to implement ut. Although i think the current one is best

scrapy/exporters.py
dracony: Any blockers with this PR? Really want it to get merged, so I dont need to
use my own fork

scrapy/exporters.py
```diff
@@ -84,19 +85,24 @@ class JsonLinesItemExporter(BaseItemExporter):
...
-        self.encoder = ScrapyJSONEncoder(**kwargs)
+        ensure_ascii = not self.encoding
+        self.encoder = ScrapyJSONEncoder(ensure_ascii=ensure_ascii, **kwargs)
 
     def export_item(self, item):
         itemdict = dict(self._get_serialized_fields(item))
-        self.file.write(to_bytes(self.encoder.encode(itemdict) + '\n'))
+        data = self.encoder.encode(itemdict) + '\n'
+        if six.PY3:
+            data = to_unicode(data)
```
redapple: why is this needed for Python 3?
I tried with a very simple test and it seems json encoder already outputs Py3-`'str'`

```
$ python
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
>>> import json
>>> json.encoder.JSONEncoder(ensure_ascii=False).encode({u"é": 3})
'{"é": 3}'
>>> type(json.encoder.JSONEncoder(ensure_ascii=False).encode({u"é": 3}))
<class 'str'>
>>> json.encoder.JSONEncoder(ensure_ascii=True).encode({u"é": 3})
'{"\\u00e9": 3}'
>>> type(json.encoder.JSONEncoder(ensure_ascii=True).encode({u"é": 3}))
<class 'str'>
>>> from scrapy.utils.python import to_unicode
>>> type(to_unicode(json.encoder.JSONEncoder(ensure_ascii=True).encode({u"é": 3})))
<class 'str'>
>>> type(to_unicode(json.encoder.JSONEncoder(ensure_ascii=False).encode({u"é": 3})))
<class 'str'>
```

am I missing something?

dracony:
@redapple True, it seems I overcomplicated it a bit when I was trying to make the test pass. I removed those lines and the tests still pass 👍

dracony:
@redapple @robsonpeixoto Ping =) 
Can this get merged now?

scrapy/exporters.py
```diff
@@ -195,12 +207,12 @@ def export_item(self, item):
...
                                              include_empty=True)
         values = list(self._build_row(x for _, x in fields))
-        self.csv_writer.writerow(values)
+        self._writerow(values)
 
     def _build_row(self, values):
         for s in values:
             try:
-                yield to_native_str(s)
+                yield to_bytes(s) if six.PY2 else s
```
eliasdorneles: hm, why is this needed?
unless I'm missing something, `to_native_str()` is equivalent.

scrapy/exporters.py
```diff
@@ -214,9 +226,18 @@ def _write_headers_and_set_fields_to_export(self, item):
                     # use fields declared in Item
                     self.fields_to_export = list(item.fields.keys())
             row = list(self._build_row(self.fields_to_export))
-            self.csv_writer.writerow(row)
-
-
+            self._writerow(row)
+            
+    def _writerow(self, row):
```
eliasdorneles: Can you clarify why is this needed?
It looks unrelated to this PR, what am I missing?

scrapy/exporters.py
eliasdorneles: also, perhaps you want to pass `self.encoding` here too?

scrapy/exporters.py
```diff
@@ -84,19 +85,22 @@ class JsonLinesItemExporter(BaseItemExporter):
     def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
-        self.encoder = ScrapyJSONEncoder(**kwargs)
+        ensure_ascii = not self.encoding
```
eliasdorneles: It would be better to do this as:
```
kwargs.setdefault('ensure_ascii', not self.encoding)`
```
and removing the explicit `ensure_ascii` from the ScrapyJSONEncoder call.

This way it won't break subclasses that are already be passing `ensure_ascii` for this in kwargs.

scrapy/exporters.py
```diff
@@ -84,19 +85,22 @@ class JsonLinesItemExporter(BaseItemExporter):
...
 
 
 class JsonItemExporter(BaseItemExporter):
 
     def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
-        self.encoder = ScrapyJSONEncoder(**kwargs)
+        ensure_ascii = not self.encoding
+        self.encoder = ScrapyJSONEncoder(ensure_ascii = ensure_ascii, **kwargs)
```
eliasdorneles: The same comment about `ensure_ascii` above applies here.

scrapy/exporters.py
eliasdorneles: The duplication is still there.

scrapy/exporters.py
```diff
@@ -169,9 +176,14 @@ class CsvItemExporter(BaseItemExporter):
...
         self._configure(kwargs, dont_fail=True)
+        if not self.encoding:
+            self.encoding = 'utf-8'
         self.include_headers_line = include_headers_line
-        file = file if six.PY2 else io.TextIOWrapper(file, line_buffering=True)
-        self.csv_writer = csv.writer(file, **kwargs)
+        self.stream = six.StringIO()
+        self.file = file
+        self.stream = six.StringIO() if six.PY2 else io.TextIOWrapper(file,
+            line_buffering=True, encoding = self.encoding)
```
eliasdorneles: why is it enabling line_buffering here? is this really needed?

eliasdorneles:
Hey @dracony -- sorry for the delay in reviewing, I've made some comments above.

The main thing not clear to me are the changes in the CSV exporter -- is line buffering really needed?
It would be better to have it in a second PR (less decisions to be done to merge this one :) ), it might deserve its own setting, since flushing at every line can be slower.

eliasdorneles:
Also, if you could revise the style once more, there are some minor stuff there (try running `flake8 scrapy/exporters.py`).
Not a big deal, but makes it nicer for the reviewer's brain to parse it. :)
Thanks!

dracony:
@eliasdorneles 
The change https://github.com/scrapy/scrapy/pull/2034#discussion-diff-69482186L203 was indeed irrelevant and changes nothing. It was a sideeffect with me fighting with the CSV writer.  

I gave it one more try and made the CSV writer work without changing encoding line by line (although in my defence I got that approach from the csv manual page)

As for the `line_buffering` being on it does seem to be required since turning it off broke the test entirely. I would be happy to skip the TextIOWrapper altogether, but the only way to do that would be reopening the file again with the encoding parameter specified. And considering it's the file object that is being passed to the exporter, closing and reopening the file seems more like a hack

eliasdorneles:
@dracony 
Using `TextIOWrapper` is totally fine, the current PY3 code is already using it.
I still think it's weird why `line_buffering` is being required for PY3, it looks like we'll have different behavior in PY2. Which test failed?

Please see my other comment about passing `ensure_ascii` parameter -- it's best to use the `kwargs.setdefault` approach to maintain compatibility with subclasses.
Aside that little nitpicking, the PR looks good to me, nice work! :+1:

dracony:
@eliasdorneles I changed the ensure_ascii to kwargs as you suggested. Also I found a solution for why `line_buffering` `False` was failing tests. The solution was to also enable `write_through`, otherwise the buffer was not flushed to disk

eliasdorneles:
@dracony hey, thanks for digging! :+1: 
Hm, that kinda makes me suspicious of the tests, like they may be trying to read without having guarantee the file is flushed.
I don't think we should need to be forcing flush (we don't seem to need it for the PY 2 code)... I'll have a look to confirm, thank you!

dracony:
@eliasdorneles It's not the file that is nt being flushed, it's that TextIOBuffer

dracony:
Yay =)

docs/topics/feed-exports.rst
```diff
@@ -233,6 +234,13 @@ The serialization format to be used for the feed. See
 
 .. setting:: FEED_EXPORT_FIELDS
 
+FEED_EXPORT_ENCODING
+--------------------
+
+The encoding to be used for the feed. Defaults to UTF-8.
```
redapple: This is not correct for JSON(lines), the default encoding is `\uXXXX` escape sequences.
A special case indeed, but worth a mention I believe.

dracony: How would I phrase it though? Wouldn't it be confusing to say it's
"\uXXXX escape
sequences" ?

eliasdorneles: Thanks @redapple, missed that on the review.

You can say "safe numeric encoding (`\uXXXX` sequences)".

redapple: Not sure. The default value is `None`, which means is the backward-compatible default behavior (current master, before the merge), which is:
- `\uXXXX` escape sequences for JSON
- UTF-8 for XML and CSV
(if I'm not mistaken)

And users can change to UTF-8 for JSON instead of `\uXXXX` with `FEED_EXPORT_ENCODING = 'utf-8'`

If set, and different from `"utf-8"`, it changes behavior for all exporters.

I don't have good phrasing right now. Maybe @eliasdorneles has ideas?

eliasdorneles: Concretely, here is my suggestion:

```
FEED_EXPORT_ENCODING
--------------------

Default: ``None``

The encoding to be used for the feed.

If unset or set to ``None`` (default) it uses UTF-8 for everything except JSON output,
which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.

```

dracony: Let it be so =) I updated the PR with your suggestion

docs/topics/feed-exports.rst
```diff
@@ -233,6 +234,20 @@ The serialization format to be used for the feed. See
...
+Default: ``None``
+
+The encoding to be used for the feed.
+
+If unset or set to ``None`` (default) it uses UTF-8 for everything except JSON output,
+which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
+
+Use ``utf-8`` if you want UTF-8 for JSON too.
+
+.. setting:: FEED_EXPORT_ENCODING
```
redapple: Sorry, I had missed this:
it should be `.. setting:: FEED_EXPORT_FIELDS` here and `.. setting:: FEED_EXPORT_ENCODING` earlier

dracony:
Ah I see. Fixed that

redapple:
@dracony , thanks a lot!
It took a while, sorry about that.

Issue #1866, opened 2016-03-16
Move from boto2/botocore to boto3


jersub:
Hi all,

Scrapy moved to a hybrid use of both `boto2`/`botocore` since PR #1761. I feel maintaining two AWS libraries would be difficult on the long term, and testing would be even more painful. I see no point for keeping `boto2` support (apart for some needed backward compatibility) since `botocore` supports Python 2.7 as well (see https://botocore.readthedocs.org/en/latest/).

Besides, I see some benefits for using `boto3` instead of `botocore`. It supports Python 2.7 as well, see https://github.com/boto/boto3/blob/develop/tox.ini. Actually it stands at a higher level and would allow us to fix bug #960 in few lines of code (cf. multipart upload, see http://boto3.readthedocs.org/en/latest/reference/customizations/s3.html#module-boto3.s3.transfer).

Anyway issue #1790 (i.e. S3 testing) would have to be handled at the same time.

lopuhin:
+1 for using ``boto3`` for S3 file uploads. I started with signing s3 requests, and there ``botocore`` felt more convenient than ``boto3`` (and even still too high-level). But for file uploads ``boto3`` indeed looks much better.

wRAR:
As boto support will be unused after dropping Python 2, it may be a good idea to decide if we should replace it with boto3 support, remove boto support code and boto/botocore conditionals or keep that for someone who will implement boto3 support.

Gallaecio:
Since we do not have to choose one or another, I wonder if we should not just close this issue, and allow contributors to introduce `boto3` as a dependency if they wish to fix issues or introduce enbhancements.

wRAR:
I think it's time to close this.

Issue #1790, opened 2016-02-18
Run all S3 tests on travis


lopuhin:
There are some tests S3 that are currently skipped on travis (`S3FeedStorageTest`, `TestS3FilesStore`) because they need real S3. It would be nice to run them, here are some options:
- Test against real S3, using https://docs.travis-ci.com/user/environment-variables/ to hide ASW keys. But there are some problems: tests might be unstable if S3 is down, and this depends on someone providing AWS keys and paying the (tiny) bills.
- Use something like https://github.com/jubos/fake-s3 to fake S3 - if this works, than it will be much more convenient.
- Mock S3 (did not find anything satisfactory upfront, but maybe there is something).

From discussion here: https://github.com/scrapy/scrapy/pull/1761#issuecomment-185753423

darshanime:
Hey @lopuhin!
I would like to attempt to set `fake-s3` up. I installed `fake-s3` on my local machine and successfully tried a few `put`, `rb` requests. Now, for our tests, I would need to install ruby and flask-s3 in `tox` and Travis environments by editing the `tox.ini` and `.travis.yml`. In the `TestS3FilesStore` test itself, what about this:
```python
#rest of the test
        uri = "s3://test.bucket/scrapy-test" 
        #or we can set S3_TEST_FILE_URI to this in tox.ini
        store = S3FilesStore(uri)
        store.port = 4567
        store.host = 'localhost'
#rest of the test
```
(We can start the flask-s3 server at port `4567` in the `tox.ini` and `.travis.yml`.)

Am I on the correct path?

lopuhin:
Hey @darshanime , I think your approach is correct, please go ahead! Maybe it would also be nice to check that ``fake-s3`` is running at the start of the test, and skip the test if it's not.

kmike:
It may be better to start fake-s3 server as a pytest fixture, better on a random port. We have technical debt in Scrapy already with mockserver and mitmproxy - they run on fixed ports, and this makes it impossible to execute tests in parallel.

I also wonder how good is https://github.com/spulec/moto - it seems to have server mode.

darshanime:
Hey @kmike, @lopuhin!
I tried to use `moto` today to simulate S3 with mixed success.
I used pytest fixtures to setup the S3 session as suggested.

```python
from moto import mock_s3

class TestS3FilesStore(unittest.TestCase):

    @pytest.fixture(autouse=True)
    def setup_s3(self):
            mock = mock_s3()
[...]

PR #1559, opened 2015-10-28
Add support for multipart upload to the S3 feed storage

jersub:
This PR adds support for multipart upload to the S3 feed storage. It aims to allow uploading of large files (e.g. items containing HTML from a big crawl) and to fix the issue #960.

Source snippets:
- http://boto.readthedocs.org/en/latest/s3_tut.html#storing-large-data
- boto/boto#2704

PR #2036, opened 2016-06-07
[WIP] boto3 for S3 file uploads

rampage644:
This PR would address the same issue as #1559 and partly #1866.
UPDATE: #1790 is also related

TODO:
- [x] Unittest with `moto` (or `mock`) ~~as well as functional test~~
- [ ] Replace `botocore` with `boto3` for `S3FeedStorage`
- [ ] Move to multipart upload

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,5 +1,6 @@
 # Tests requirements
 mock
+moto==0.4.24
 mitmproxy==0.10.1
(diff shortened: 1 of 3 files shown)


lopuhin:
Might be also related: #1790

rampage644:
@lopuhin: yes, you're right. Will update description soon.

Issue #1507, opened 2015-09-23
custom "CsvItemExporter" doesn't support 'from_settings' ? 


jschilling1:
I used to access settings in my custom exporter using "from scrapy.conf import settings" which is now depreciated. 'from_settings' and 'from_crawler' don't work there what should i do?

```
class CustomCsvItemExporter(CsvItemExporter):
    def __init__(self, *args, **kwargs):
        super(CustomCsvItemExporter, self).__init__(*args, **kwargs)
    @classmethod
    def from_settings(cls, settings):
        print ('from_settings ran')
        return cls(settings)
```

kmike:
Fixed by https://github.com/scrapy/scrapy/pull/3348.

PR #1605, opened 2015-11-16
Add from_crawler constructor for feed exporters and storages

jdemaeyer:
Addresses #1567. This moves the "which constructor should I call" logic from the middleware manager into a util function, and re-uses it in the feed exporter extension.

Feed exporters and storages can now provide a `from_crawler()` class method and properly access settings.

PR #2044, opened 2016-06-08
Access settings from feed storages

### LINKED ISSUE #1567: Access settings from feed storages
Hi,

Feed storages should be able to access settings the new way.

For instance, `S3FeedStorage` uses both `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` from the settings. Currently they're gotten from `scrapy.conf` but this is now deprecated: this cause the build of PR #1559 to fail.

What about adding settings to the constructor of storages?

rampage644:
Intention is to acceess settings in feed storages that could require some more information to be configured propery (providing `uri` is not sufficient for them).

For example, `S3FeedStorage` requires AWS credentials that could be supplied from settings and not uri.
Given full text search over scrapy project, this storage is the only thing that actually uses `scrapy.conf.settings` to access them.

I don't want storages to implement `from_crawler` method (for anything) as `FeedExporter`, `IStorage`s and `IFeedExporters` is much simpler abstraction than `FeedExporter`, `IStorage`s and `IFeedExporters` _plus_ signals, crawlers, or whatever. We already have FeedExporter extension that could forward any signal to storages and/or exporters.

Closes #1567. #1507 Also related.
#2036 depends on this.

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -32,7 +32,7 @@ logger = logging.getLogger(__name__)
 class IFeedStorage(Interface):
     """Interface that all Feed Storages must implement"""
 
-    def __init__(uri):
+    def __init__(uri, settings=None):
         """Initialize the storage with the parameters given in the URI"""
 
     def open(spider):
@@ -63,7 +63,7 @@ class BlockingFeedStorage(object):
 @implementer(IFeedStorage)
 class StdoutFeedStorage(object):
 
-    def __init__(self, uri, _stdout=None):
+    def __init__(self, uri, settings=None, _stdout=None):
         if not _stdout:
             _stdout = sys.stdout if six.PY2 else sys.stdout.buffer
         self._stdout = _stdout
@@ -78,7 +78,7 @@ class StdoutFeedStorage(object):
 @implementer(IFeedStorage)
 class FileFeedStorage(object):
 
-    def __init__(self, uri):
+    def __init__(self, uri, settings=None):
         self.path = file_uri_to_path(uri)
 
     def open(self, spider):
@@ -93,8 +93,7 @@ class FileFeedStorage(object):
 
 class S3FeedStorage(BlockingFeedStorage):
 
-    def __init__(self, uri):
-        from scrapy.conf import settings
+    def __init__(self, uri, settings=None):
         u = urlparse(uri)
         self.bucketname = u.hostname
         self.access_key = u.username or settings['AWS_ACCESS_KEY_ID']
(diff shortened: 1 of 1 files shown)


kmike:
See also: https://github.com/scrapy/scrapy/pull/1605

rampage644:
@kmike, thanks for more information.

I personally don't like the idea implementing `from_crawler` factory for all classes but most people seem to be OK with it. 
However, I'd like to get this done. PR you mentioned wasn't merged and discussion has stopped. Is it possible to get this merged I switch to `from_crawler` methods and not regular ctors.

I'm interested in getting this done :)

kmike:
@rampage644 in #1605 the idea was to support both `from_settings` and `from_crawler`. I agree that using settings (via `from_settings`) is a simpler interface, but by supporting both (and extracting code to support both to a function) it is possible to make Scrapy more consistent - if it could have a single 'extension' interface which can be used for most components then it'd be a win.

Passing settings object directly to constructor can be a bit worse than using `from_settings` method because `from_settings` needs Settings instance, while constructor can accept just the necessary options as keyword arguments and raise an error if necessary arguments are not passed. This is definitely more ceremony than just accepting Settings or Crawler object in constructor, but it makes code less coupled.

I'm on fence about showing a warning for `from_settings` (as suggested in comments to #1605). On one hand, `from_settings` is a simpler API; on the other hand, with `from_crawler` one can just use crawler.settings, and one can use all necessary settings as constructor arguments, so `from_settings` method doesn't bring much value.

Another option is to deprecate `from_settings`, make all components use `from_crawler`, and pass settings to constructors consistently. Components which don't need crawler may avoid implementing from_crawler method (like storage classes in your PR). But this would be a much-much larger change if we want all components work like this, because all constructors will have to be changed to accept `settings` parameter.

In #1605 there is also some code for backwards compatibility with settings-less Storage constructors; it'd be nice to keep it. After changes in this PR S3FeedStorage may fail when created manually.

So +1 to get this feature in Scrapy; using `from_settings` and `from_crawler` is not perfect (it is a bit too verbose for users), but for me it looks like the best option because otherwise it is hard to make all Scrapy components consistent.

rampage644:
> while constructor can accept just the necessary options as keyword arguments and raise an error if necessary arguments are not passed. This is definitely more ceremony than just accepting Settings or Crawler object in constructor, but it makes code less coupled.

I have +1'ed this somewhere already. I believe it happened _after_ i've got that far into codebase. Putting only necessary parameters into constructor makes much more sense than using generic `Settings` object.

I'll update PR with `from_*` methods implementation.

PR #2045, opened 2016-06-09
[WIP] Change BaseSettings#__getitem__ to behave like a regular dict's one

josericardo:
The previous implementation broke the MutableMapping contract.
`setdefault`, for one, doesn't work anymore.

`__getitem__` is expected to raise KeyError for non-existing keys, unless
we're trying to emulate a defaultdict, which should be explicit in the
docs and is backwards incompatible.

### DIFF
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -91,8 +91,6 @@ class BaseSettings(MutableMapping):
         self.update(values, priority)
 
     def __getitem__(self, opt_name):
-        if opt_name not in self:
-            return None
         return self.attributes[opt_name].value
 
     def __contains__(self, name):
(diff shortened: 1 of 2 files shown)


jdemaeyer:
Ugh, good find.

For clarification, this is not "how `Settings` has always been", because this also effects dictionary-like settings. It's a regression from automatically converting dictionary-like settings to `BaseSettings`. For example,
```python
# Insert my middleware with default priority if it's not enabled already
settings['DOWNLOADER_MIDDLEWARES'].setdefault('mymodule.MyMiddleware', 100)
```
worked before Scrapy 1.1 and is now broken.

Maybe we could restore the old behaviour by moving our custom `__getitem__` (that returns `None` instead of raising `KeyError`) from `BaseSettings` to `Settings`.

josericardo:
It looks like I'll have to be a little more careful with this. Changing it to WIP

PR #2048, opened 2016-06-10
[MRG] Add FAQ entry on using BeautifulSoup in spider callbacks

redapple:
See https://twitter.com/PackOsiris/status/741114216833241089 for motivation

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -26,15 +26,17 @@ comparing `jinja2`_ to `Django`_.
 .. _jinja2: http://jinja.pocoo.org/
 .. _Django: https://www.djangoproject.com/
 
-How can I use Scrapy with BeautifulSoup?
-----------------------------------------
+Can I use Scrapy with BeautifulSoup?
+------------------------------------
 
-As mentioned :ref:`above <faq-scrapy-bs-cmp>`, BeautifulSoup can be used
+Yes, you can.
+As mentioned :ref:`above <faq-scrapy-bs-cmp>`, `BeautifulSoup`_ can be used
 for parsing HTML responses in Scrapy callbacks.
 You just have to feed the response's body into a ``BeautifulSoup`` object
 and extract whatever data you need from it.
 
-Here's an example spider using ``lxml`` parser with BeautifulSoup API::
+Here's an example spider using BeautifulSoup API, with ``lxml`` as the HTML parser
+(so you get the same parsing speed as with scrapy/parsel selectors)::
 
 
     from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -24,6 +26,36 @@ comparing `jinja2`_ to `Django`_.
...
 
+How can I use Scrapy with BeautifulSoup?
+----------------------------------------
+
+As mentioned :ref:`above <faq-scrapy-bs-cmp>`, BeautifulSoup can be used
+for parsing HTML responses in Scrapy callbacks.
+You just have to feed the response's body into a ``BeautifulSoup`` object
+and extract whatever data you need from it.
+
+Here's an example spider using ``lxml`` parser with BeautifulSoup API::
```
kmike: Why is 'lxml' needed for the example? Does it work without 'lxml'?

redapple: It does work without yes

kmike: I think it is better to keep the example minimal; it may worths adding a note about 'lxml' though - if I'm not mistaken, it is not default because lxml is hard to install, but if Scrapy is installed then lxml is already installed.

redapple: Just test, what you get without "lxml" is:

```
2016-06-10 19:17:34 [scrapy] DEBUG: Crawled (200) <GET http://www.example.com/> (referer: None)
2016-06-10 19:17:34 [py.warnings] WARNING: /home/paul/.virtualenvs/scrapybs4.py3/lib/python3.5/site-packages/bs4/__init__.py:166: UserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("lxml"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.

To get rid of this warning, change this:

 BeautifulSoup([your markup])

to this:

 BeautifulSoup([your markup], "lxml")

  markup_type=markup_type))

2016-06-10 19:17:35 [scrapy] DEBUG: Scraped from <200 http://www.example.com/>
{'url': 'http://www.example.com/', 'title': 'Example Domain'}
```
so it works but it's not the cleanest output.
but I agree we can keep the example minimal, and maybe add a in-code comment about adding "lxml"

kmike: That's not a good API by BS.. Maybe we should keep 'lxml' so, but add a comment about it, and maybe a link to https://www.crummy.com/software/BeautifulSoup/bs4/doc/#specifying-the-parser-to-use

redapple:
@kmike , I've updated the entry. Reads better?

docs/faq.rst
```diff
@@ -24,6 +26,45 @@ comparing `jinja2`_ to `Django`_.
...
+------------------------------------
+
+Yes, you can.
+As mentioned :ref:`above <faq-scrapy-bs-cmp>`, `BeautifulSoup`_ can be used
+for parsing HTML responses in Scrapy callbacks.
+You just have to feed the response's body into a ``BeautifulSoup`` object
+and extract whatever data you need from it.
+
+Here's an example spider using BeautifulSoup API, with ``lxml`` as the HTML parser
+(so you get the same parsing speed as with scrapy/parsel selectors)::
```
kmike: Are you sure about that? I recall some old benchmark where BS+lxml was much slower than just lxml.

redapple: I have no idea. I just extrapolated
> If you can, I recommend you install and use lxml for speed.
from https://www.crummy.com/software/BeautifulSoup/bs4/doc/

redapple: I'll remove that line

kmike:
Looks good, thanks @redapple!

Issue #2049, opened 2016-06-12
utils.is_gzipped returns false on application/x-gzip;charset=utf-8


Tethik:
I had a site return the following header, which caused SitemapSpider to not parse a sitemap xml that was gzipped. 
```
Content-Type: application/x-gzip;charset=utf-8
```
Looking into the code it seems that the function [utils.is_gzipped](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/gz.py#L54) does not take into account cases where the Content-Type header would include charset.

eliasdorneles:
Fixed in #2050!
@Tethik @robsonpeixoto @redapple thanks fellows!

PR #2050, opened 2016-06-12
[MRG+1] Is_gzipped for application/x-gzip;charset=utf-8

Tethik:
Fix for #2049

### DIFF
--- a/scrapy/utils/gz.py
+++ b/scrapy/utils/gz.py
@@ -55,4 +55,4 @@ _is_gzipped_re = re.compile(br'^application/(x-)?gzip\b', re.I)
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return not _is_gzipped_re.search(ctype) is None
+    return _is_gzipped_re.search(ctype) is not None
--- a/scrapy/utils/gz.py
+++ b/scrapy/utils/gz.py
@@ -7,7 +7,7 @@ except ImportError:
 from gzip import GzipFile
 
 import six
-
+import re
 
 # - Python>=3.5 GzipFile's read() has issues returning leftover
 #   uncompressed data when input is corrupted
@@ -50,8 +50,9 @@ def gunzip(data):
                 raise
     return output
 
+_is_gzipped_re = re.compile(br'^application/(x-)?gzip\b', re.I)
 
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return b'application/x-gzip' in ctype or b'application/gzip' in ctype
+    return not _is_gzipped_re.search(ctype) is None
--- a/scrapy/utils/gz.py
+++ b/scrapy/utils/gz.py
@@ -54,4 +54,4 @@ def gunzip(data):
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return ctype in (b'application/x-gzip', b'application/gzip')
+    return b'application/x-gzip' in ctype or b'application/gzip' in ctype
--- a/tests/test_utils_gz.py
+++ b/tests/test_utils_gz.py
@@ -1,7 +1,8 @@
 import unittest
 from os.path import join
 
-from scrapy.utils.gz import gunzip
+from scrapy.utils.gz import gunzip, is_gzipped
+from scrapy.http import Response, Headers
 from tests import tests_datadir
 
 SAMPLEDIR = join(tests_datadir, 'compressed')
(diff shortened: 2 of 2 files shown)


scrapy/utils/gz.py
```diff
@@ -54,4 +54,4 @@ def gunzip(data):
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return ctype in (b'application/x-gzip', b'application/gzip')
+    return b'application/x-gzip' in ctype or b'application/gzip' in ctype
```
robsonpeixoto: Just a suggestion:

```python
from six.moves import map # just if you would like a lazy solution

mimes = (b'application/x-gzip', b'application/gzip')
return any(map(lambda m: m in ctype, mimes))
```

tests/test_utils_gz.py
```diff
@@ -27,3 +28,16 @@ def test_gunzip_truncated_short(self):
         with open(join(SAMPLEDIR, 'truncated-crc-error-short.gz'), 'rb') as f:
             text = gunzip(f.read())
             assert text.endswith(b'</html>')
+
+    def test_is_gzipped(self):
+        hdrs = Headers({"Content-Type": "application/x-gzip"})
+        r1 = Response("http://www.example.com", headers=hdrs)
+        self.assertTrue(is_gzipped(r1))
```
robsonpeixoto: I do not known about the scrapy code style, but IMO it should be 4 different tests.

scrapy/utils/gz.py
Tethik: I did it as an `any`-statement first, but figured it would be overkill since there are only two comparisons. Another version:

```
any(mime in ctype for mime in (b'application/x-gzip', b'application/gzip'))
```

redapple:
According to [RFC 7231, section-3.1.1.1](https://tools.ietf.org/html/rfc7231#section-3.1.1.1), 
> The type, subtype, and parameter name tokens are case-insensitive.

so it's worth fixing that too.

Suggestion:

```
$ git diff
diff --git a/scrapy/utils/gz.py b/scrapy/utils/gz.py
index f174950..e520b42 100644
--- a/scrapy/utils/gz.py
+++ b/scrapy/utils/gz.py
@@ -1,3 +1,4 @@
+import re
 import struct
 
 try:
@@ -51,7 +52,9 @@ def gunzip(data):
     return output
 
 
+_is_gzipped_re = re.compile(br'^application/(x-)?gzip\b', re.I)
+
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return b'application/x-gzip' in ctype or b'application/gzip' in ctype
+    return _is_gzipped_re.search(ctype)
diff --git a/tests/test_utils_gz.py b/tests/test_utils_gz.py
index 3648d5c..02bf1d7 100644
--- a/tests/test_utils_gz.py
+++ b/tests/test_utils_gz.py
@@ -37,6 +37,16 @@ class GunzipTest(unittest.TestCase):
         r1 = Response("http://www.example.com", headers=hdrs)
         self.assertTrue(is_gzipped(r1))
 
+    def test_is_gzipped_case_insensitive(self):
+        hdrs = Headers({"Content-Type": "Application/X-Gzip"})
+        r1 = Response("http://www.example.com", headers=hdrs)
+        self.assertTrue(is_gzipped(r1))
+
+    def test_is_gzipped_not_quite(self):
+        hdrs = Headers({"Content-Type": "application/gzippppp"})
+        r1 = Response("http://www.example.com", headers=hdrs)
+        self.assertFalse(is_gzipped(r1))
+
     def test_is_gzipped_empty(self):
         r1 = Response("http://www.example.com")
         self.assertTrue(not is_gzipped(r1))
@@ -50,3 +60,7 @@ class GunzipTest(unittest.TestCase):
         hdrs = Headers({"Content-Type": "application/x-gzip;charset=utf-8"})
         r1 = Response("http://www.example.com", headers=hdrs)
         self.assertTrue(is_gzipped(r1))
+
+        hdrs = Headers({"Content-Type": "application/X-GZIP ; charset=utf-8"})
+        r1 = Response("http://www.example.com", headers=hdrs)
+        self.assertTrue(is_gzipped(r1))

```

Tethik:
@redapple Alright, I'll look into it. Good catch on the gzipppp

Tethik:
Done

scrapy/utils/gz.py
```diff
@@ -50,8 +50,9 @@ def gunzip(data):
                 raise
     return output
 
+_is_gzipped_re = re.compile(br'^application/(x-)?gzip\b', re.I)
 
 def is_gzipped(response):
     """Return True if the response is gzipped, or False otherwise"""
     ctype = response.headers.get('Content-Type', b'')
-    return ctype in (b'application/x-gzip', b'application/gzip')
+    return not _is_gzipped_re.search(ctype) is None
```
redapple: I usually prefer `return _is_gzipped_re.search(ctype) is not None`

Tethik: Zzzz... fine ;)

redapple: Sorry for nitpicking, but I first read (quickly) "not is_gzipped".

Tethik: Nahh it's fine. I agree `is not` looks better. I'm just lazy :)

redapple:
Looks good to me.
Thanks @Tethik!

tests/test_utils_gz.py
```diff
@@ -27,3 +28,39 @@ def test_gunzip_truncated_short(self):
         with open(join(SAMPLEDIR, 'truncated-crc-error-short.gz'), 'rb') as f:
             text = gunzip(f.read())
             assert text.endswith(b'</html>')
+
+    def test_is_gzipped_right(self):
```
robsonpeixoto: IMO should break this test in two different:
```python
def test_is_gzipped_right(self):
    hdrs = Headers({"Content-Type": "application/gzip"})
    r1 = Response("http://www.example.com", headers=hdrs)
    self.assertTrue(is_gzipped(r1))

def test_is_x_gzipped_right(self):
    hdrs = Headers({"Content-Type": "application/x-gzip"})
    r1 = Response("http://www.example.com", headers=hdrs)
    self.assertTrue(is_gzipped(r1))
```

PR #2058, opened 2016-06-17
[MRG+1] Add set serialization to ScrapyJSONEncoder

dalleng:
Tried using sets a few times as a way to get a list that avoid duplicates in field values. Only to remember that neither Python or Scrapy's encoder can serialize them.

### DIFF
--- a/scrapy/utils/serialize.py
+++ b/scrapy/utils/serialize.py
@@ -14,7 +14,9 @@ class ScrapyJSONEncoder(json.JSONEncoder):
     TIME_FORMAT = "%H:%M:%S"
 
     def default(self, o):
-        if isinstance(o, datetime.datetime):
+        if isinstance(o, set):
+            return list(o)
+        elif isinstance(o, datetime.datetime):
             return o.strftime("%s %s" % (self.DATE_FORMAT, self.TIME_FORMAT))
         elif isinstance(o, datetime.date):
             return o.strftime(self.DATE_FORMAT)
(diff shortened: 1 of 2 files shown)


kmike:
I think this change is fine.

The problem with set serialization is that you get a list back, which has a different time complexity at lookups. In your example after duplicate check will run in O(1) before the change, and in O(N) after. But for items export it looks fine.

scrapy/utils/serialize.py
```diff
@@ -14,7 +14,9 @@ class ScrapyJSONEncoder(json.JSONEncoder):
     TIME_FORMAT = "%H:%M:%S"
 
     def default(self, o):
-        if isinstance(o, datetime.datetime):
+        if isinstance(o, set):
+            return list(o)
```
kmike: Do you know if ScrapyJSONEncoder.default is applied for each element of a set? E.g. is it possible to serialize a set of datetime.datetime objects? It'd be nice to have a test for that.

redapple:
hi @dalleng , have you have time to check @kmike 's comment?

dalleng:
@redapple Yes, just haven't had much spare time recently.

scrapy/utils/serialize.py
dalleng: Will check that.

dalleng:
Updated the PR

tests/test_utils_serialize.py
```diff
@@ -23,9 +23,14 @@ def test_encode_decode(self):
         ts = "10:11:12"
         dec = Decimal("1000.12")
         decs = "1000.12"
+        s = {'foo'}
+        ss = ['foo']
+        dt_set = {dt}
+        dt_sets = ["2010-01-02 10:11:12"]
```
kmike: shouldn't it be [dts]?

dalleng: It's the same. But I'll change to reuse the `dts` variable.

kmike:
Thanks @dalleng!

Issue #2056, opened 2016-06-14
[DOC] Mention that Scrapy + Python 3 + Windows do not work together


redapple:
Even if it's announced in https://blog.scrapinghub.com/2016/05/25/data-extraction-with-scrapy-and-python-3/
it's not clear enough that Scrapy does not work on Python 3 on Windows

> ImportError: cannot import name '_win32stdio'
 
See comments at http://stackoverflow.com/a/37818668/

redapple:
See https://github.com/scrapy/scrapy/pull/2060

PR #2060, opened 2016-06-17
[MRG+1] [docs] add note about windows + python3

pawelmhm:
#2056 seems like common issue (e.g. user here: http://stackoverflow.com/questions/36857638/can-i-run-scrapy-on-windows-with-python-3/37418422) , should we also update [installation guide](http://doc.scrapy.org/en/1.1/intro/install.html) with note about Python 3? Currently it does not say anything about Python 3 but people assume 3 should be supported.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -71,7 +71,10 @@ What Python versions does Scrapy support?
 
 Scrapy is supported under Python 2.7 and Python 3.3+.
 Python 2.6 support was dropped starting at Scrapy 0.20.
-Python 3 support was added in Scrapy 1.1. Python 3 is not yet supported on Windows.
+Python 3 support was added in Scrapy 1.1.
+
+.. note::
+    Python 3 is not yet supported on Windows.
 
 Did Scrapy "steal" X from Django?
 ---------------------------------
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -71,7 +71,7 @@ What Python versions does Scrapy support?
 
 Scrapy is supported under Python 2.7 and Python 3.3+.
 Python 2.6 support was dropped starting at Scrapy 0.20.
-Python 3 support was added in Scrapy 1.1.
+Python 3 support was added in Scrapy 1.1. Python 3 is not yet supported on Windows.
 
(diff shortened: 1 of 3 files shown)


redapple:
thanks @pawelmhm ,
yes, I think we need to update installation guide and also update [the FAQ entry](http://doc.scrapy.org/en/latest/faq.html#what-python-versions-does-scrapy-support) on Python 3. we never know on which page users land from a Google search

pawelmhm:
> yes, I think we need to update installation guide and also update the FAQ entry on Python 3. we never know on which page users land from a Google search

updated

redapple:
Thanks @pawelmhm , added a few comments.

docs/faq.rst
```diff
@@ -71,7 +71,7 @@ What Python versions does Scrapy support?
 
 Scrapy is supported under Python 2.7 and Python 3.3+.
 Python 2.6 support was dropped starting at Scrapy 0.20.
-Python 3 support was added in Scrapy 1.1.
+Python 3 support was added in Scrapy 1.1. Python 3 is not yet supported on Windows.
```
pawelmhm: I changed this to "note"

docs/intro/install.rst
```diff
@@ -85,6 +85,10 @@ Windows
 
       pip install Scrapy
 
+.. note::
+     Python 3 is not supported on Windows. Installation of Scrapy on Windows
```
pawelmhm: yeah I updated this part, I see there is ticket for this in Twister repo here: https://twistedmatrix.com/trac/ticket/8018

redapple:
Thanks @pawelmhm !

Namsto:
Do we will support "Scrapy + Python 3 + Windows" in later release?

redapple:
Twisted has not announced Python 3 + Windows support yet, so Scrapy cannot support Python 3 on Windows yet.
See their [release notes for Twisted 16.4](http://labs.twistedmatrix.com/2016/08/twisted-1640-released.html)
> 35+ more modules ported to Python 3, and many many cleanups on the way to Python 3 on Windows support.

Maybe with Twisted 16.5 or 17

PR #2061, opened 2016-06-17
[MRG+1] process_spider_exception on generators

### LINKED ISSUE #220: process_spider_exception() not invoked for generators
When a spider implements _parse()_ function through generator (i.e. returning results with _yields_) and an exception raises in _parse()_, then spidermiddleware's _process_spider_exception()_ is not called.

The reason is that exception is finally invoked in one of the spidermiddlewares.

elacuesta:
This PR is a starting point to fix #220.
It could probably use some more test cases, mostly to figure out what exactly is the desired behaviour when processing the exceptions.
I can't take much credit for this: if it breaks, blame @dangra 😛


elacuesta:
Oh wait, I was so excited about the test included here passing that I forgot to run the whole test suite! `test_closespider.py`is failing, I'll look into it.

elacuesta:
All checks passing now, but I have two concerns:
* When `process_spider_exception` returns `None` the exception appears in the job stats, but it does not appear if the method returns an iterable. This was not introduced by this PR, so it should be addressed in a separate PR IMHO.
* If a Response subclass object is returned as part of the iterable, Scrapy complains with `ERROR: Spider must return Request, BaseItem, dict or None, got 'HtmlResponse'`. As stated in [this doc page](http://doc.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.process_spider_exception) responses might be returned as part of that iterable, but I'm not sure about the desired behaviour: is this a problem with Scrapy or with the docs?

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,109 @@
...
+    def parse(self, response):
+        """ should scrape 3 items """
+        yield {'value': 10}
+        raise UnicodeError
+
+
+class CatchExceptionMiddleware(object):
+    def process_spider_exception(self, response, exception, spider):
+        """ catch an exception and log it """
+        logging.warn('{} exception catched'.format(exception.__class__.__name__))
```
kmike: catched -> caught

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,109 @@
...
+        return [
+            TestItem(name='value'),
+            {'foo': 'bar'},
+        ]
+
+
+class TestSpiderMiddleware(TestCase):
+
+    def setUp(self):
+        pass
```
kmike: why is empty method needed?

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,109 @@
...
+
+class CatchExceptionMiddleware(object):
+    def process_spider_exception(self, response, exception, spider):
+        """ catch an exception and log it """
+        logging.warn('{} exception catched'.format(exception.__class__.__name__))
+        return None
+
+class HandleExceptionMiddleware(object):
+    def process_spider_exception(self, response, exception, spider):
+        """ handle and exception, return something useful """
```
kmike: and -> an

scrapy/core/spidermw.py
```diff
@@ -60,8 +60,18 @@ def process_spider_exception(_failure):
...
+                    if exception_result is None or isinstance(exception_result, Failure):
+                        raise
+                    for output in exception_result:
+                        yield output
             for method in self.methods['process_spider_output']:
-                result = method(response=response, result=result, spider=spider)
+                result = wrapper(method(response=response, result=result, spider=spider))
                 assert _isiterable(result), \
                     'Middleware %s must returns an iterable object, got %s ' % \
                     (fname(method), type(result))
```
kmike: Could you please check that this assertion is still active? I haven't tried the code, but it seems the loop in `wrapper` method will fail before `_isiterable` check.

scrapy/core/spidermw.py
```diff
@@ -60,8 +60,18 @@ def process_spider_exception(_failure):
             return _failure
 
         def process_spider_output(result):
+            def wrapper(result_iterable):
+                try:
+                    for r in result_iterable:
+                        yield r
+                except Exception as ex:
+                    exception_result = process_spider_exception(Failure(ex))
```
kmike: Docs for [process_spider_exception](https://doc.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.process_spider_exception) method say:

> This method is called when when a spider or process_spider_input() method (from other spider middleware) raises an exception.

After this change method starts to fire when `process_spider_output` of a previous middleware raises an error; I'm not sure what are the implications, but it should be documented, and it could be backwards incompatible.

elacuesta: Exceptions from `process_spider_output` are handled only when the method returns a generator. There are two cases:
- `process_spider_output` is _not_ a generator. In that case, if an exception is raised the function's return values does not pass along, and since there is no `result_iterable` the exception handler does not get called. This is exactly the same as the current behaviour and documentation.
- `process_spider_output` is a generator. In that case, the exception is not raised right away, instead it actually raises when iterating over `result_iterable`, and the exception handler is called. This is what could backwards compability, but since it's the very thing that's currently broken, I don't think anyone is relying on this functionality.

@kmike Please check the above to see if it makes any sense :-P

tests/test_spidermiddleware.py
elacuesta: It's not, I removed it :-)

tests/test_spidermiddleware.py
elacuesta: Fixed, thanks!

scrapy/core/spidermw.py
```diff
@@ -60,11 +60,22 @@ def process_spider_exception(_failure):
...
+                    if exception_result is None or isinstance(exception_result, Failure):
+                        raise
+                    for output in exception_result:
+                        yield output
             for method in self.methods['process_spider_output']:
                 result = method(response=response, result=result, spider=spider)
-                assert _isiterable(result), \
-                    'Middleware %s must returns an iterable object, got %s ' % \
-                    (fname(method), type(result))
+                if _isiterable(result):
```
elacuesta: @kmike: At first I tried to find a way of checking the result is iterable without exposing the `AssertionError` exception to be catched by some middleware's `process_spider_exception`, but then I thought, since any exception from `process_spider_input` should be passed to `process_spider_exception` (from the docs), it wouldn't be wrong if the result of `process_spider_output` passed too.

That being said, I'm getting a bit confused on what the desired/excepted behavior is/should be, please help me :-)

scrapy/core/spidermw.py
elacuesta: My answer is below, sorry. Seems like I don't fully get this new GitHub review thingy yet :-P

scrapy/core/spidermw.py
kmike: Hm, this makes sense, but it is a bit tricky, and it could be easy to break by changing the current implementation. Are there tests for that?

scrapy/core/spidermw.py
kmike: I think it doesn't make sense to catch this error in process_spider_exception - this is a programmer error, not a problem with data. So it'd be nice if `process_spider_exception` still won't be able to catch this error.

Another question: what happens if an error from `process_spider_output` is handled by `process_spider_exception`? For `process_spider_input` or for spider itself it is documented that `process_spider_output` chain is invoked, but for process_spider_exception it doesn't make sense to start process_spider_output from the beginning. 

I think executing `process_spider_output` methods which are not executed yet makes sense in this case, but this should be tested and documented.

scrapy/core/spidermw.py
elacuesta: I believe that

> This method is called when when a spider or process_spider_input() method (from other spider middleware) raises an exception

means that exceptions from previous `process_spider_output` methods should be handled by `process_spider_exception`; the way of getting a spider's result (either items, requests or exceptions) is through the `process_spider_output` chain. Right?
I'm rewriting some tests and adding some more to make this clear.

scrapy/core/spidermw.py
elacuesta: [This line](https://github.com/scrapy/scrapy/pull/2061/files#diff-7a382cc9280309e8fc166e4b479deb33R58) checks for an `AssertionError` explicitly raised by any `process_spider_input` method, and prevents it from being passed to the exception handler (the only problem would be an external exception raised with the exact same message the spider middleware manager specifies, I believe that case could be ignored).

There are [some tests here](https://github.com/scrapy/scrapy/pull/2061/files#diff-442dbe445eb9f6625bfb8186556983f2R175) to make sure that invalid return values from either `process_spider_input` or `process_spider_output` do not get handled by `process_spider_exception`.

elacuesta: @kmike [This test](https://github.com/scrapy/scrapy/pull/2061/files#diff-442dbe445eb9f6625bfb8186556983f2R269) ensures only non already executed `process_spider_output` methods are called when `process_spider_exception` returns an iterable, breaking the exception handling chain :-)

scrapy/core/spidermw.py
```diff
@@ -50,21 +55,36 @@ def process_spider_input(response):
 
         def process_spider_exception(_failure):
             exception = _failure.value
+            # ignore AssertionError from middleware's manager
+            if isinstance(exception, AssertionError) and len(exception.args) > 0 \
```
elacuesta: This line checks for an AssertionError explicitly raised by any process_spider_input method, and prevents it from being passed to the exception handler (the only problem would be an external exception raised with the exact same message the spider middleware manager specifies, I believe that case could be ignored).

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,314 @@
...
+        self.assertIn("'spider_exceptions/AssertionError'", str(log1))
+        # on middleware's output
+        crawler2 = get_crawler(InvalidReturnValueFromPreviousMiddlewareOutputSpider)
+        with LogCapture() as log2:
+            yield crawler2.crawl()
+        self.assertNotIn("AssertionError exception caught", str(log2))
+        self.assertIn("'spider_exceptions/AssertionError'", str(log2))
+
+    @defer.inlineCallbacks
+    def test_process_spider_exception_execution_chain(self):
```
elacuesta: This test ensures only non already executed `process_spider_output` methods are called when `process_spider_exception` returns an iterable, breaking the exception handling chain :-)

scrapy/core/spidermw.py
kmike: I think a better way would be to raise a special scrapy-specific exception type, not AssertionError, and ignore it instead. This code could silently ignore AssertionError raised in user code.

elacuesta: You're right, that would be way less hacky than checking the exception message. I'm on it :-)

scrapy/exceptions.py
```diff
@@ -11,6 +11,11 @@ class NotConfigured(Exception):
     """Indicates a missing configuration situation"""
     pass
 
+class InvalidValue(TypeError):
```
elacuesta: @kmike I named the exception like this, but suggestions are welcome of course :-)
Regardless of the name, I think this same exception should also be raised by the downloader middleware manager when a returned value is invalid, but that falls out of the scope of this PR, I will open a separate one.

elacuesta:
Current status of this PR is:
* Catch exceptions from previous `process_spider_output` (also documented)
* Execute only non already called `process_spider_output` methods when `process_spider_exception` returns an iterable
* Raise `InvalidValue` exception (scrapy-specific, name suggestions are welcome) when an invalid value is returned from a spider middleware's processing method

@kmike I think that addresses your latest concerns, please let me know if there's anything more I can do to get this PR moving. Thanks!

scrapy/core/spidermw.py
```diff
@@ -40,31 +43,47 @@ def process_spider_input(response):
...
+                    if result is not None:
+                        raise InvalidValue('Middleware {} must return None or raise ' \
+                            'an exception, got {}'.format(fname(method), type(result)))
                 except:
                     return scrape_func(Failure(), request, spider)
             return scrape_func(response, request, spider)
 
         def process_spider_exception(_failure):
             exception = _failure.value
+            # ignore InvalidValue exception from middleware's manager
```
kmike: I think re-raising / bubbling is not ignoring :)

scrapy/exceptions.py
kmike: I think it makes sense to make this exception private (add underscore, undocument it) because it is a Scrapy implementation detail - why would user want to use it?

InvalidValue sounds a lot like ValueError, but it is inherited from TypeError; what about _InvalidOutput or something like that?

scrapy/core/spidermw.py
```diff
@@ -3,8 +3,11 @@
 
 See documentation in docs/topics/spider-middleware.rst
 """
+import re
```
kmike: is it a leftover from a previous implementation?

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,316 @@
...
+    start_urls = ["http://example.com/"]
+    custom_settings = {
+        'SPIDER_MIDDLEWARES': {'tests.test_spidermiddleware.CatchExceptionMiddleware': 540}
+    }
+
+class NotAGeneratorSpider(BaseExceptionFromParseMethodSpider):
+    """ return value is NOT a generator """
+    name = 'not_a_generator'
+    def parse(self, response):
+        raise AssertionError
```
kmike: Not returning a generator sounds like a separate case from raising an exception

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,316 @@
...
+    value = Field()
+
+# ================================================================================
+# exceptions from a spider's parse method
+class BaseExceptionFromParseMethodSpider(Spider):
+    start_urls = ["http://example.com/"]
+    custom_settings = {
+        'SPIDER_MIDDLEWARES': {'tests.test_spidermiddleware.CatchExceptionMiddleware': 540}
+    }
+
```
kmike: a nitpick: could you please use two lines to separate classes, as per pep8?

scrapy/core/spidermw.py
elacuesta: Yes it is, I'll remove it, thanks.

scrapy/core/spidermw.py
elacuesta: You're right, I'll add a better description.

scrapy/exceptions.py
elacuesta: I like the alternative name, but I think it shouldn't be private. This exception will raise if a user-implemented middleware's `process_request` or `process_response` returns an invalid value, it's supposed to appear in the user log.

tests/test_spidermiddleware.py
elacuesta: Done :-)

tests/test_spidermiddleware.py
elacuesta: I'm afraid I don't understand what you mean :confused:

scrapy/exceptions.py
elacuesta: I renamed the exception to `InvalidOutput`, but kept it public because of the reasons above. Also, I think subclassing from [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError) (_"Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError"_) is more appropriate than [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) (_"Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value"_)

redapple:
@kmike , what do you think of this PR now?

scrapy/exceptions.py
```diff
@@ -11,6 +11,11 @@ class NotConfigured(Exception):
     """Indicates a missing configuration situation"""
     pass
 
+class InvalidOutput(TypeError):
```
kmike: +1 to inherit it from TypeError

tests/test_spidermiddleware.py
```diff
@@ -0,0 +1,340 @@
...
+
+class TestItem(Item):
+    value = Field()
+
+
+# ================================================================================
+# exceptions from a spider's parse method
+class BaseExceptionFromParseMethodSpider(Spider):
+    start_urls = ["http://example.com/"]
```
kmike: It looks like these are real spiders making requests to http://example.com; that's not good to make http requests to external services in a testing suite; it is better to use scrapy's [mockserver](https://github.com/scrapy/scrapy/blob/master/tests/mockserver.py).

elacuesta: Updated, the tests are using `tests.mockserver.MockServer` now :+1:

docs/topics/exceptions.rst
```diff
@@ -62,6 +62,17 @@ remain disabled. Those components include:
...
+InvalidOutput
+-------------
+
+.. exception:: InvalidOutput
+
+This exception can be raised by a downloader or spider middleware to
+indicate that some method returned a value not suported by the processing
+chain.
+See :ref:`topics-spider-middleware` and :ref:`topics-downloader-middleware`
+for a list of supported output values.
```
kmike: @elacuesta in https://github.com/scrapy/scrapy/pull/2061#discussion_r102538089 you argued that it is good to have this exception public & documented because it may appear in logs. But I still don't quite like making it public :) There is a lot of exceptions which can appear in logs (from Scrapy, from Twisted, other Python exceptions), and we don't document them all. To make exception in logs readable one can use a readable error message, which is already the case. 

There is a non-zero overhead of documenting an exception and making it public - user may read docs and wonder how can this exception be used. All other Scrapy documented exceptions are useful for users - user may want to either raise them or catch them; unlike all other exceptions, this exception shouldn't be raised or caught in user code. 

Docs tell that "This exception can be raised by a downloader or spider middleware", but this information can be misleading for users - we document that exceptions raised in middlewares can be caught in process_exception method, and we document that InvalidOutput can be raised in a middleware - but in fact this particular exception can't be caught in process_exception. 

For me this exception still looks like an implementation detail with no use for end users; documenting it in scrapy docs doesn't help scrapy users, only make these docs longer to read and a bit more confusing. For example, users may think they should raise this exception if they validate items yielded by a spider, and validation fails ("value not supported by the processing chain") - but they shouldn't do this, they shouldn't raise this exception themselves.

I think information you've written here is valuable for people who modify Scrapy itself; it can be great to have something like that in the exception docstring, to make it easier for people to figure out how Scrapy works internally.

elacuesta: I see your point, there is no case in which a user would raise or catch this exception, it's only something to make them check their code if they write a bad middleware. I'll remove the docs about it. Thanks for the feedback :+1:

elacuesta: I removed the exception from the docs and added an underscore to the name. I also modified the `DownloaderMiddlewareManager` class to raise `_InvalidOutput` too, not a good idea to raise a custom exception from the spider middleware and `AssertionError` from the downloader middleware IMHO.

kmike:
Sorry for a delayed review. You've addressed all my comments, but it is hard to review such pull requests because there are so many moving parts, it is hard to get a big picture. Could we have a table like this?

Exception raised in | Generator? | The same as in Scrapy 1.3 | Handling
---------------------|------------|----------------------------|-----------
`process_spider_input`| N | Y | request errback is called (?), then   process_spider_exception chain is called
request callback | N | Y | errback (?), then process_spider_exception chain
request callback | Y | N | errback (?), then process_spider_exception chain
``process_spider_output`` | N | N | `process_spider_exception` chain
... | ... | ... | ...

By the way, there is a [FIXME comment](https://github.com/scrapy/scrapy/blob/776129a9513e2b6ab6f7e8cda1dd3de66cbbff44/scrapy/core/scraper.py#L138) in scraper.py - should we remove it in this PR?

elacuesta:
I think making a table like that is a great idea, but I'm now confused about something: I thought `request.callback` and `request.errback` are at the "end" of the `process_spider_input` chain, in which case neither of them should be called if there's an error in the middle of the chain, am I correct?
Also, I don't have much experience with Twisted so please correct me if I'm wrong, but I thought one cannot jump to the errback from a callback in the same level (https://twistedmatrix.com/documents/current/core/howto/defer.html#visual-explanation), so in a spider an `errback` shouldn't be called from a `callback`.
This makes me wonder if we really need to worry about errbacks here, as I don't think the way they are handled is modified by this PR.

I hope that makes sense :-)

redapple:
FYI, I'm moving this to 1.5 milestone.

elacuesta:
After several failed attempts I think I have come up with a solution for this problem. I updated the implementation, tests and docs. The main goal here is to make the middleware chain to behave like a Twisted deferred, jumping between the process_spider_output and process_spider_exception methods from each enabled middleware as if they were callbacks/errbacks (see https://twistedmatrix.com/documents/current/core/howto/defer.html#visual-explanation).
I made a sketch of how the input/output chains will look after this change, it's just a picture of my whiteboard but if the changes are good we can probably get a designer to make a nice diagram to include in the docs. I also included a behavior table as Mikhail suggested. 

Regarding performance, I ran a test using `scrapy-bench`, modifying the spider in that tool to yield the results, there seems to be no noticeable impact (command: `scrapy-bench --n-runs=10 --book_url=http://localhost:8080/books bookworm)
`):
```
master branch (258121db, 2018-07-18), spider returns list
Mean : 49.55994650734702 Median : 50.03054505433736 Std Dev : 1.6747484769938523

master branch (258121db, 2018-07-18), spider returns generator
Mean : 51.30306411841985 Median : 51.565729573433515 Std Dev : 2.244161762734139

process_spider_exception_generator branch (20defa2e, 2018-07-19), spider returns list
Mean : 50.69768225842105 Median : 50.8686334029228 Std Dev : 0.9180605567200794

process_spider_exception_generator branch (20defa2e, 2018-07-19), spider returns generator
Mean : 50.10554590818768 Median : 49.88208980828408 Std Dev : 3.0805330389799663
```

@kmike, @dangra please take a look!



| Number | Exception raised in | Generator? | Previous behavior | Current behavior | Notes |
| --- | --- | --- | --- | --- | --- |
| 1 | process_spider_input | N/A | request errback is called | (unchanged) | Current documentation says the process_spider_exception is called (updated by this PR) |
| 2 | request callback | Yes | The exception is logged. process_spider_exception chain is not invoked | process_spider_exception chain, starting from the first middleware | Fixed by this PR |
| 3 | request callback | No | process_spider_exception chain, starting from the first middleware | (unchanged) |  |
| 4 | process_spider_output | Yes | The exception is logged. process_spider_exception chain is not invoked | process_spider_exception chain, starting from the next middleware | Fixed by this PR. Items produced before the exception are processed normally |
| 5 | process_spider_output | No | The exception is logged. process_spider_exception chain is not invoked | process_spider_exception chain, starting from the next middleware | Fixed by this PR. Items produced before the exception are lost |

spider_middleware

scrapy/core/spidermw.py
```diff
@@ -3,15 +3,40 @@
...
 from scrapy.middleware import MiddlewareManager
 from scrapy.utils.defer import mustbe_deferred
 from scrapy.utils.conf import build_component_list
 
+
 def _isiterable(possible_iterator):
     return hasattr(possible_iterator, '__iter__')
 
+
+class MutableChain:
```
kmike: would you mind moving it to scrapy.utils.python?

kmike: Also, as we support Python 2, it is clearner to inherit it from object.

scrapy/core/spidermw.py
```diff
@@ -3,15 +3,40 @@
...
+    def extend(self, *iterables):
+        self.data = chain(self.data, *iterables)
+
+    def __iter__(self):
+        return self.data.__iter__()
+
+    def __next__(self):  # py3
+        return self.data.__next__()
+
+    def next(self):  # py2
```
kmike: I wonder if you can make `def next` an alias `next = __next__`, and implement `__next__` as `next(self.data)`. `next` built-in function calls `.next()` in Python 2 and `__next__()` in Python 3 - see e.g. https://docs.python.org/2/library/functions.html#next

scrapy/core/spidermw.py
```diff
@@ -40,36 +63,71 @@ def process_spider_input(response):
...
             return scrape_func(response, request, spider)
 
-        def process_spider_exception(_failure):
+        def process_spider_exception(_failure, start_index=0):
             exception = _failure.value
-            for method in self.methods['process_spider_exception']:
+            # don't handle _InvalidOutput exception
+            if isinstance(exception, _InvalidOutput):
+                return _failure
+            for method_index, method in enumerate(self.methods['process_spider_exception'][start_index:], start=start_index):
```
kmike: could you please extract `self.methods['process_spider_exception'][start_index:]` to a variable, to make the line shorter?

docs/topics/spider-middleware.rst
```diff
@@ -112,8 +113,8 @@ following methods:
 
     .. method:: process_spider_exception(response, exception, spider)
 
-        This method is called when a spider or :meth:`process_spider_input`
-        method (from other spider middleware) raises an exception.
+        This method is called when a spider or :meth:`process_spider_output`
+        method (from a previous spider middleware) raises an exception.
```
kmike: 👍

scrapy/core/spidermw.py
```diff
@@ -24,12 +49,10 @@ def _add_middleware(self, mw):
...
         if hasattr(mw, 'process_spider_input'):
             self.methods['process_spider_input'].append(mw.process_spider_input)
-        if hasattr(mw, 'process_spider_output'):
-            self.methods['process_spider_output'].insert(0, mw.process_spider_output)
-        if hasattr(mw, 'process_spider_exception'):
-            self.methods['process_spider_exception'].insert(0, mw.process_spider_exception)
         if hasattr(mw, 'process_start_requests'):
             self.methods['process_start_requests'].insert(0, mw.process_start_requests)
+        self.methods['process_spider_output'].insert(0, getattr(mw, 'process_spider_output', None))
+        self.methods['process_spider_exception'].insert(0, getattr(mw, 'process_spider_exception', None))
```
kmike: Are you adding None to the list to make start_index (and slicing) work properly?

scrapy/core/spidermw.py
kmike: the same in process_spider_output

elacuesta:
Hi there @kmike 👋
Thanks for the feedback, I updated the PR based on your comments.

elacuesta:
This is awesome, a million thanks @kmike :raised_hands:

 Looking forward to 1.7 then, thanks again!

kmike:
Heh, poor @elacuesta - at least 4 Scrapy committers looked at it in some way; there are two +1's now, but we're not brave enough to merge it. Please don't take it personal - that's the price of tackling hard issues :) @dangra @lopuhin do you want to take another look? If no, I can merge it this week.

scrapy/core/spidermw.py
```diff
@@ -40,36 +44,73 @@ def process_spider_input(response):
...
                     result = method(response=response, spider=spider)
-                    assert result is None, \
-                            'Middleware %s must returns None or ' \
-                            'raise an exception, got %s ' \
-                            % (fname(method), type(result))
+                    if result is not None:
+                        raise _InvalidOutput('Middleware {} must return None or raise an exception, got {}' \
+                                             .format(fname(method), type(result)))
+                except _InvalidOutput:
+                    raise
```
lopuhin: Just to check that I'm understanding the code correctly, why do we re-raise ``_InvalidOutput`` here?

lopuhin: Ah never mind, I see why - to avoid it falling into the next ``except`` clause.

lopuhin:
Many thanks @elacuesta and thanks @kmike and @Gallaecio for review 👍

PR #2064, opened 2016-06-19
Add Azure Storage to files/images pipeline options

fpghost:
These changes allow the user to use Azure Storage for files or images persistence of media when using the files or images pipelines, in an analogous way to AWS S3 Storage. The user simply needs to define the following settings:

```
IMAGES_STORE = "azure://<container_name>/"
AZURE_ACCOUNT_NAME = <azure_account_name>
AZURE_SECRET_ACCESS_KEY = <azure_secret_key>
```

and instead of `boto` install `azure-storage`: `pip install azure-storage`, and all should work out of the box.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -194,6 +194,82 @@ class S3FilesStore(object):
         return extra
 
 
+class AzureFilesStore(object):
+
+    AZURE_ACCOUNT_NAME = None
+    AZURE_SECRET_ACCESS_KEY = None
+
+    HEADERS = {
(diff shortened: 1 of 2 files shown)


kmike:
Hey @fpghost,

I'm not opposed to add azure storage support to Scrapy, but it needs tests. 

Also, instead of putting it in Scrapy itself we can make Azure uploads a separate project in https://github.com/scrapy-plugins/. It requires some thought - how should Scrapy be changed to allow this kind of plugins?

Gallaecio:
Closing due to a lack of feedback from the author.

Issue #951, opened 2014-11-14
Extraction of gzipped sitemap fails in Scrapy 0.24.4


mercutio79:
retrieving a gzipped sitemap xml (tested on amazon.de) fails.

Reproduce with :
modify /utils/gz.py gunzip method to write the incoming data to a file.

gunzip the file on the command line.

the unzipped file contains garbled content

gunzip that file with garbled content a second time and get the correct content

-> I suspect that the content coming from the target server is already gzip compressed and scrapy has a bug that causes the gzip decompression to not work properly, resulting in a double compressed file arriving at the  /utils/gz.py gunzip method

dszmaj:
i've encountered the same problem with redirect downloads, without redirect everything is fine, it looks that only after redirect, content gets double gzipped

workaraund from @mercutio79 works perfectly with this

juraseg:
Compression middleware behaviour on gzipped sitemaps still seems to be wrong. 

In case of amazon it does not work anymore, because it sets header "Content-Type" to "application/octet-stream" instead of "application/x-gzip". So this for this issue - [https://github.com/scrapy/scrapy/issues/193](Issue #193) - does not work anymore. It still tries to decompress the file two times - once in HTTPCompressionMiddleware and once in Sitemap spider.

I've encountered opposite issue on other site. Sitemap file is also gzipped and has extension ".gz", but when site gets header "Accept-Encoding"="x-gzip,gzip,deflate" (which is added by compression middleware) it compresses the file once more. Response contents in this case is twice compressed. This is of course an error on site hosters side, but in my opinion Scrapy should account for this anyway - I imagine there are a lot of such sites in the web.


To get rid of all of these issues I suggest to move the whole compression behavior to HTTPCompressionMiddleware, and it should perform decompressing in presence of only "Content-Encoding" header (disregard check for "Content-Type") and do decompression in a loop until gets an error (easier to ask forgiveness than permission) - gzip module will fail earlier of first read so should not add much overhead. Another option is to check gzip file format, as suggested by @dangra in https://github.com/scrapy/scrapy/issues/193](Issue #193). This would be a more complex solution, but potentially will perform better.

Also, to account for cases when HTTPCompressionMiddleware is disabled, Sitemap can still try to decompress response content if extension is ".gz".

redapple:
Fixed in scrapy 1.1.1 (https://github.com/scrapy/scrapy/pull/2065):

Example spider for amazon.fr sitemap:

```
[...]

PR #2065, opened 2016-06-20
[MRG+2] [HttpCompressionMW] Do not decompress gzip binary/octet-stream responses

### LINKED ISSUE #2063: IOError, 'Not a gzipped file'
while trying to access sitemap from robots.txt , Scrapy fails with **IOError, 'Not a gzipped file'** error

not sure if this issue is related to following issue(s)
https://github.com/scrapy/scrapy/issues/193 -> closed issue
https://github.com/scrapy/scrapy/pull/660 -> merged pull request to address issue 193
https://github.com/scrapy/scrapy/issues/951 -> open issue

> line where code fails in gzip.py at line # 197
```python 
def _read_gzip_header(self):
        magic = self.fileobj.read(2)
        if magic != '\037\213':
            raise IOError, 'Not a gzipped file'
```

#Response Header
```
Content-Encoding: gzip
Accept-Ranges: bytes
X-Amz-Request-Id: BFFF010DDE6268DA
Vary: Accept-Encoding
Server: AmazonS3
Last-Modified: Wed, 15 Jun 2016 19:02:20 GMT
Etag: "300bb71d6897cb2a22bba0bd07978c84"
Cache-Control: no-transform
Date: Sun, 19 Jun 2016 10:54:53 GMT
Content-Type: binary/octet-stream
```

Error Log:

```log
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\utils\defer.py", line 102, in iter_errback
    yield next(it)
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\spidermiddlewares\offsite.py", line 29, in process_spider_output
    for x in result:
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\spidermiddlewares\referer.py", line 22, in <genexpr>
    return (_set_referer(r) for r in result or ())
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\spidermiddlewares\urllength.py", line 37, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\spidermiddlewares\depth.py", line 58, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "D:\projects\sitemap_spider\sitemap_spider\spiders\mainspider.py", line 31, in _parse_sitemap
    body = self._get_sitemap_body(response)
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\spiders\sitemap.py", line 67, in _get_sitemap_body
    return gunzip(response.body)
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\utils\gz.py", line 37, in gunzip
    chunk = read1(f, 8196)
  File "c:\venv\scrapy1.0\lib\site-packages\scrapy\utils\gz.py", line 21, in read1
    return gzf.read(size)
  File "c:\python27\Lib\gzip.py", line 268, in read
    self._read(readsize)
  File "c:\python27\Lib\gzip.py", line 303, in _read
    self._read_gzip_header()
  File "c:\python27\Lib\gzip.py", line 197, in _read_gzip_header
    raise IOError, 'Not a gzipped file'
```
i did download file manually and was able to extract the content so it is not like file is corrupted 

as an example sitemap url : you can follow amazon robots.txt

redapple:
This should fix #2063 and concerns from https://github.com/scrapy/scrapy/issues/951#issuecomment-180249747

As outlined in https://github.com/scrapy/scrapy/issues/2063#issuecomment-227116765 , the idea here is:
*  if `Content-Encoding` is said to be "gzip",
* to process `Content-Type: binary/octet-stream` (or `application/octet-stream`) as an opaque file that higher-level layers should handle,
* even if HTTP specs would require the engine to gunzip the body for the client application.

This is already what happens for `Content-Type: application/(x-)gzip` (see #660)

This behavior is inline with what browsers do for sitemap files for example,
e.g. Amazon being "famous" for sending back this kind of HTTP headers for https://www.amazon.fr/sitemaps.*.xml.gz files (gzipped-compressed XML sitemap files):
```
{b'Content-Encoding': [b'gzip'],
 b'Etag': [b'"212a0faf86341f24001a784bf6d8df51"'],
 b'Accept-Ranges': [b'bytes'],
 b'Server': [b'Server'],
 b'Last-Modified': [b'Fri, 08 Apr 2016 21:47:04 GMT'],
 b'Vary': [b'Accept-Encoding'],
 b'Content-Type': [b'binary/octet-stream'],
 b'Date': [b'Mon, 20 Jun 2016 14:31:26 GMT'],
 b'X-Amz-Request-Id': [b'0E43407DAC8EE8DE']}
```


redapple:
https://github.com/scrapy/scrapy/pull/2050 should also be merged into this

scrapy/utils/gz.py
```diff
@@ -51,8 +51,12 @@ def gunzip(data):
     return output
 
 _is_gzipped_re = re.compile(br'^application/(x-)?gzip\b', re.I)
+_is_octetstream_re = re.compile(br'^(application|binary)/octet-stream\b', re.I)
```
kmike: A small trick: it can be written as 

```py
_is_octetstream = re.compile(br'^(application|binary)/octet-stream\b', re.I).search
```

and then used like ` or (_is_octetstream(ctype) and cenc in ...)`

redapple: maybe @Tethik can comment too

Tethik: Nice trick. It would make the final boolean expression a tiny bit easier to read.

```python
return (_is_gzipped(ctype) is not None or
            (_is_octetstream(ctype) is not None and
             cenc in (b'gzip', b'x-gzip')))
```
:+1:

kmike: Ah right, _is_gzipped can also be written this way. I'd even drop 'is not None' checks. From https://docs.python.org/3.5/library/re.html#match-objects: 

> Match objects always have a boolean value of True

There is an example in stdlib docs:

```py
match = re.search(pattern, string)
if match:
    process(match)
```

Tethik: Doing `is not None` explicitly makes it a bool though. If we drop the `is not None` then we're going to return None or True instead of True or False. I doubt it makes a difference though. 

```python
return _is_gzipped(ctype) or
            (_is_octetstream(ctype) and
             cenc in (b'gzip', b'x-gzip')))
```

If we want to be really pedantic we could probably also drop the parantheses too.

kmike: @redapple The PR looks fine :+1:. Could you please clean up and merge it?

2016-07-09 4:25 GMT+05:00 Joakim Uddholm <notifications@github.com>:

Issue #1434, opened 2015-08-13
Issue scraping through https


teddb:
I can successfully scrape an http page through a proxy. I can also scrape an https page without using a proxy. I cannot scrape an https page through a proxy.

My spider works successfully if the 's' in 'https' of the target page is removed or I disable the proxy. I can access the https page through the proxy through my browser.

Here is what I added to the settings file:

    DOWNLOADER_MIDDLEWARES = {'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware': 110, 'test_website.middlewares.ProxyMiddleware': 100}
    PROXIES = [{'ip_port': 'us-il.proxymesh.com:31280', 'user_pass': 'username:password'}]

Here is the code for my spider:

    import scrapy
     
    class TestSpider(scrapy.Spider):
        name = "test_spider"
        allowed_domains = "ipify.org"
         start_urls = ["https://api.ipify.org"]
     
        def parse(self, response):
            with open('test.html', 'wb') as f:
                f.write(response.body)

Here is the middlewares file:

    import base64
    import random
    from settings import PROXIES
     
    class ProxyMiddleware(object):
        def process_request(self, request, spider):
            proxy = random.choice(PROXIES)
            if proxy['user_pass'] is not None:
                request.meta['proxy'] = "http://%s" % proxy['ip_port']
                encoded_user_pass = base64.encodestring(proxy['user_pass'])
                request.headers['Proxy-Authorization'] = 'Basic ' + encoded_user_pass            
            else:
                request.meta['proxy'] = "http://%s" % proxy['ip_port']

Here is the log file:

        2015-08-12 20:15:50 [scrapy] INFO: Scrapy 1.0.3 started (bot: test_website)
    2015-08-12 20:15:50 [scrapy] INFO: Optional features available: ssl, http11
    2015-08-12 20:15:50 [scrapy] INFO: Overridden settings: {'NEWSPIDER_MODULE': 'test_website.spiders', 'SPIDER_MODULES': ['test_website.spiders'], 'LOG_STDOUT': True, 'LOG_FILE': 'log.txt', 'BOT_NAME': 'test_website'}
    2015-08-12 20:15:51 [scrapy] INFO: Enabled extensions: CloseSpider, TelnetConsole, LogStats, CoreStats, SpiderState
[...]

PR #2069, opened 2016-06-21
[MRG] Add "Host" header in CONNECT requests to HTTPS proxies

redapple:
"Host" request header is a MUST according to [RFC 2616](https://tools.ietf.org/html/rfc2616#page-128)

> A client MUST include a Host header field in all HTTP/1.1 request
>    messages . If the requested URI does not include an Internet host
>    name for the service being requested, then the Host header field MUST
>    be given with an empty value. An HTTP/1.1 proxy MUST ensure that any
>    request message it forwards does contain an appropriate Host header
>    field that identifies the service being requested by the proxy. All
>    Internet-based HTTP/1.1 servers MUST respond with a 400 (Bad Request)
>    status code to any HTTP/1.1 request message which lacks a Host header
>    field.

so it must also be sent for CONNECT requests.

Fixes issue raised in https://github.com/scrapy/scrapy/issues/1434#issuecomment-227393544

Note:
HTTPS proxy tests need some love.
We could perhaps reuse https://github.com/fmoo/twisted-connect-proxy

It seems that currently the only test using CONNECT is `tests/test_downloader_handlers.py::Http11ProxyTestCase::test_download_with_proxy_https_timeout`,
which sends this over the wire (with this "Host" header fix):

```
CONNECT no-such-domain.nosuch:443 HTTP/1.1
Host: no-such-domain.nosuch:443
```
And test server returns
```
HTTP/1.1 200 OK
Server: TwistedWeb/16.2.0
Date: Tue, 21 Jun 2016 11:33:19 GMT
Content-Type: text/html
Content-Length: 25

no-such-domain.nosuch:443
```
I don't know where that `no-such-domain.nosuch:443` line at the end comes from.

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -93,7 +93,7 @@ class TunnelingTCP4ClientEndpoint(TCP4ClientEndpoint):
     for it.
     """
 
-    _responseMatcher = re.compile(b'HTTP/1\.. 200')
+    _responseMatcher = re.compile(b'HTTP/1\.. (?P<status>\d{3})(?P<reason>.{,32})')
 
     def __init__(self, reactor, host, port, proxyConf, contextFactory,
                  timeout=30, bindAddress=None):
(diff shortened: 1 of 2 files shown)


kmike:
@redapple do you know why are tests failng in Python 3.5?

redapple:
@kmike , hm, I don't know, the tests pass for me locally.
Maybe something to do with that "no-such-domain.nosuch:443" sent as response body

redapple:
Ok, so that "no-such-domain.nosuch:443" line comes from `HttpProxyTestCase` where server is set up with `UriResource()` which returns the URI by default.

redapple:
@kmike , test is passing but we should really strengthen these proxy tests

nyov:
YES!
Thanks  :heart:

scrapy/core/downloader/handlers/http11.py
```diff
@@ -151,23 +157,22 @@ def connect(self, protocolFactory):
         return self._tunnelReadyDeferred
 
 
-def tunnel_request_data(host, port, proxy_auth_header=None):
+def tunnel_request_data(host, port, proxy_auth_header=None, host_header=True):
```
kmike: why is host_header argument needed?

redapple: Not really needed. Definitely an artifact from my tests, switching this on and off.
It's a boolean because it doesn't convey a string value, but tells to add host string or not, compared to proxy_auth_header which is a string.
I can remove it.

kmike: Noticed that while reading coverage report - host_header was always True in tests. Let's drop it :) +1 to merge the PR after that.

PR #2080, opened 2016-06-29
Turn off use_ssl for botocore interface.

mengbiping:
use_ssl=True would be too expensive when calling spiders configured with ImagePipeline. In my experience, it slows down scrapy by 5X and more.

In addition, this keeps the same behaviour as we had for boto interface, as in

```
c = self.S3Connection(..., is_secure=False)
```

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -95,7 +95,7 @@ class S3FilesStore(object):
             import botocore.session
             session = botocore.session.get_session()
             self.s3_client = session.create_client(
-                's3', aws_access_key_id=self.AWS_ACCESS_KEY_ID,
+                's3', use_ssl=False, aws_access_key_id=self.AWS_ACCESS_KEY_ID,
                 aws_secret_access_key=self.AWS_SECRET_ACCESS_KEY)
         else:
             from boto.s3.connection import S3Connection


nyov:
Do not assume that people always have your preference of speed over data privacy.
Please make it a configurable setting, with the current setting as default (for backwards compatibility).

Perhaps `AWS_SECURE_CONNECTION` or something (maybe coordinate with #2085 to use the same setting-name in both places). Thanks.

mengbiping:
I agree that making it a configurable setting would be a better solution. But I think in the world of scrapy which is designed to crawl public data from the web, privacy of crawled images does not seem to be preferred by majority of the users, especially at significant expenses of CPU cycles. I think taking the way majority would benefit is most reasonable.

How do you think?

Not sure if we have a documentation/discussion of the principles here. I'd like to know the rational behind it, if I may. Thank you!

nyov:
You don't want to hardcode a bad security practice. Nope, never.

I don't see a reason for Scrapy getting 5x slower, but I have no test data. Anyone else?
An S3 SSL connection in theory shouldn't (depending on crypto algorithm) be much slower than an SSL connection of the downloader/crawler.
Perhaps it's your installation and there's some optimizing in how python/S3 talks to the underlying SSL library? Or do you run on a cheap VPS with a big pipe but no compute power.

Okay, let's talk principles.
A security policy can always be relaxed, but a security blunder can not be undone.

With scraping, there is almost always _other peoples' data_ involved (unless you use it to warm up your websites' HTTP-cache). Only because you got at the data cheap or "for free", doesn't make it worthless (or you would not take the effort to aggregate it) and you shouldn't treat it like dirt.
Whether you took it or were entrusted with it, you are the keeper now and please handle it more responsibly [than they did](https://haveibeenpwned.com/PwnedWebsites).
Today you do not care about the privacy of some strangers' data. Tomorrow a stranger will not care about yours. It's called the [Monkeysphere](http://www.cracked.com/article_14990_what-monkeysphere.html).
Every Scrapy user is a little Google. You trust them with your data, treat your collected data as you want them to treat yours. You know what they say. Do know evil. ;)

Now for this PR.

> But I think in the world of scrapy which is designed to crawl public data from the web [...]

Please carefully check your datapoints for that assumption. Because that is _your_ use-case, doesn't make it everyone else's.
See [scrapy.org](http://scrapy.org/companies/) for a number of companies who use Scrapy commercially, some of which use it to handle sensitive customer data and non-public sources. Perhaps in an intranet search-engine, or doing customer website analysis. Certainly a few use Scrapy with S3 to push around serious data.
Look at the number of "stars" and "forks" of this project, and perhaps the number of [downloads](http://pypi-ranking.info/search/scrapy/) on PyPi. The code that you contribute and that gets merged into this codebase, will at some point affect all these people using it.

> [...] privacy of crawled images does not seem to be preferred by majority of the users, especially at significant expenses of CPU cycles.

It should be. With [current image-analysis](https://how-old.net/) software, and reverse-image search capabilities, a leaked photo may be all it takes to identify and track someone all over the web. And technology doesn't do backwards. And [the](http://heresthethingblog.com/2013/12/12/google-tip-download-giant-archive/) [Internet](https://archive.org/details/utzoo-wiseman-usenet-archive) [does](https://commoncrawl.org/the-data/get-started/) [not](https://www.githubarchive.org/) [forget](http://datasift.com/platform/historics/).

Your change is to `scrapy/pipelines/files.py`. You should be aware that that affects the `FilesPipeline`. _Your_ usage may only involve the `ImagesPipeline`, but others use other parts of the codebase and it will affect them.

Consider a webservice that uses Scrapy to fetch user's digital bank statements, using S3 to push the data for text-recognition and analysis (ok, they should know better). Will they like your hidden change to speed up their transfer rates? When suddenly all their users' account numbers, balances and purchase logs are flying through the web in the clear?

wRAR:
There is now a config setting for this, `AWS_USE_SSL`.

Issue #2047, opened 2016-06-10
add "flags" kwarg to Scrapy Request


pawelmhm:
Scrapy [response.flags](http://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response.flags) are really useful and underappreciated feature. It would be useful to add a way to pass flags argument to Request at initialization, e.g.

``` python

def some_callback(self, response):
      yield Request('http://example.com', flags={"issued from": "some_callback", "going to": "somewhere")
```

and then in log output you will have:

```
[scrapy.core.engine] Crawled (200) <GET http://example.com>  ['issued_from": "some_callback', "going to": "somewhere"]
```

This is useful because it allows you to have some heplful debugging metadata directly in logs. If you want to debug some spider problem you will have some metadata directly in logs and you dont have to start debugger in some callback and investigate response meta or some other stuff.

kmike:
Fixed by #2082, though flags is a list, like in responses.

Issue #1890, opened 2016-03-29
Disk queues don't preserve Request class


kmike:
When a Request subclass (e.g. FormRequest) is sent to a disk queue a bare Request is what you get back. 

This is inconvenient for scrapy-splash: Splash requests all have Splash URL as request.url, but for logging it is nice to display the requested URL, not only Splash URL. In scrapy-splash this is implemented by changing `__repr__` in a Request subclass, but it works only when request is kept in memory.

elacuesta:
Hello @kmike! I think the problem is in [this line](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/reqser.py#L50), the `request_from_dict` function always returns a `Request` object regardless of the type of the request passed to `request_to_dict`. Please check the above PR which tries to solve the issue. Thanks!

PR #2082, opened 2016-06-30
[MRG+2] add flags to request

foromer4:
Issue #2047, by pawelmhm


kmike:
I think these flags will disappear when Requst is serialized to a disk queue - `scrapy.util.reqser` should be changed. Also, it looks like request.replace should keep flags attribute instead of discarding it.

There is a small efficiency penalty - e.g. with 1M requests in queue it is at least 64MB. Maybe it worths it to store empty flags as None internally, but I'm not sure (users should be able to do flags.append without checking for None). Opinions? Is efficiency penalty OK?

Another way to get more information out of Requests is to use custom Request subclasses which override `__repr__`; it doesn't work because of https://github.com/scrapy/scrapy/issues/1890.

scrapy/utils/reqser.py
```diff
@@ -58,7 +59,8 @@ def request_from_dict(d, spider=None):
         meta=d['meta'],
         encoding=d['_encoding'],
         priority=d['priority'],
-        dont_filter=d['dont_filter'])
+        dont_filter=d['dont_filter'],
+        flags= d['flags'])
```
kmike: I think it should be d.get('flags'), for backwards compatibility, i.e. to be able to continue a long crawl after Scrapy update. 

Maybe it makes sense not to serialize flags if flags is None, but I'm not sure about that.

kmike:
Hey @foromer4,

It looks better. I think docs are still needed: this feature should be mentioned here: https://github.com/scrapy/scrapy/blob/master/docs/topics/request-response.rst

foromer4:
Hi, I think all requested changes are in, is it possible to merge into main branch?

scrapy/utils/reqser.py
```diff
@@ -31,6 +31,7 @@ def request_to_dict(request, spider=None):
         '_encoding': request._encoding,
         'priority': request.priority,
         'dont_filter': request.dont_filter,
+        'flags' : request.flags
```
kmike: could you please remove a whitespace before `:`?

scrapy/utils/reqser.py
```diff
@@ -58,7 +59,8 @@ def request_from_dict(d, spider=None):
         meta=d['meta'],
         encoding=d['_encoding'],
         priority=d['priority'],
-        dont_filter=d['dont_filter'])
+        dont_filter=d['dont_filter'],
+        flags= d.get('flags'))
```
kmike: there is an extra space after `=`

tests/test_utils_reqser.py
```diff
@@ -25,7 +25,8 @@ def test_all_attributes(self):
             cookies={'currency': u'руб'},
             encoding='latin-1',
             priority=20,
-            meta={'a': 'b'})
+            meta={'a': 'b'},
+            flags= ['testFlag'])
```
kmike: There is an extra space after `=` - could you plese remove it? We're trying to follow PEP8 in Scrapy source code, with 80 columns limit somewhat relaxed.

foromer4:
Hopefully documentation and spacing issues are now resolved.

tests/test_utils_reqser.py
```diff
@@ -53,6 +54,7 @@ def _assert_same_request(self, r1, r2):
         self.assertEqual(r1._encoding, r2._encoding)
         self.assertEqual(r1.priority, r2.priority)
         self.assertEqual(r1.dont_filter, r2.dont_filter)
+        self.assertEquals(r1.flags, r2.flags)
```
kmike: According to https://docs.python.org/3.5/library/unittest.html#deprecated-aliases assertEquals is a deprecated alias to assertEqual - could you please use assertEqual, like in above lines?

foromer4: Right, fixed it now. thanks

foromer4: Are there any more issues I should fix?

foromer4: should I try to further advance this feature or discard it?

scrapy/http/request/__init__.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, url, callback=None, method='GET', headers=None, body=None,
         self.dont_filter = dont_filter
 
         self._meta = dict(meta) if meta else None
+        self.flags = None if flags is None else list(flags)
```
redapple: Is this to address @kmike 's https://github.com/scrapy/scrapy/pull/2082#issuecomment-231215967 ?
I haven't tested but does "_users should be able to do flags.append without checking for None_ work"?

If `None` is to be kept, what about `self.flags = list(flags) if flags else None`?

tests/test_utils_reqser.py
kmike: @foromer4 I like this feature, definitely +1 to finish it. Sorry for us being so picky!

scrapy/http/request/__init__.py
foromer4: Thanks , missed that one... now pushing a change

tests/test_utils_reqser.py
foromer4: thanks for the patience ,  hope I addressed all the issues you raised please let me know if I missed anything

kmike:
Looks good 👍 
@foromer4 could you please sqash the commits?

foromer4:
squashed commits

kmike:
@foromer4 something went wrong - there is a lot of unrelated commits in this PR. Could you please rebase it on master?

foromer4:
Please excuse my inexperience...
I tried rebasing and than squashing using rebase HEAD~[X]

But I think it caused some mess. If I understand, you can squash during the merge process of the fork:
https://github.com/blog/2141-squash-your-commits

But I don't think I have the privileges for this.

redapple:
Looks good to me. We can use GitHub's "Squash and merge" option here.

kmike:
@redapple let's try it :)

kmike:
Thanks @foromer4 for the persistence!

foromer4:
Thank you, for the patience :)

PR #2085, opened 2016-07-02
S3FilesStore enable SSL

### LINKED ISSUE #2083: S3FilesStore should not disable SSL
Per [comment in source](https://github.com/scrapy/scrapy/blame/master/scrapy/pipelines/files.py#L121), the S3 storage backend for media pipelines silently strips SSL from the connection to S3, ostensibly because of this [python core bug](http://bugs.python.org/issue5103), which oddly enough was fixed 3 years before the code in scrapy was written.

This issue has been resolved in all modern versions of Python (2.7+/3.0+) for quite some time. Can this workaround be removed?

maksimbormot:
Enable SSL because of this python bug is fixed on 2.6 and 3.x Python versions.

Fixes #2083


nyov:
+1 to default to SSL connections here, if the bug is fixed.
Have you tested with boto that there is no other regression with this switch?

But I feel that this should be a configurable setting (`AWS_SECURE_CONNECTION` or something), since it then makes sense to toggle it.
And please do not append to the comment when the comment has become obsolete - instead remove it. The reasoning for the change can (should) go into the commit message. Thanks

scrapy/pipelines/files.py
```diff
@@ -118,10 +118,10 @@ def _onsuccess(boto_key):
...
         return self._get_boto_key(path).addCallback(_onsuccess)
 
-    def _get_boto_bucket(self):
-        # disable ssl (is_secure=False) because of this python bug:
+    def _get_boto_bucket(self, AWS_SECURE_CONNECTION=False):
+        # adden ability to change secure because of this python bug is fixed on 2.6 and 3.x Python versions:
         # http://bugs.python.org/issue5103
-        c = self.S3Connection(self.AWS_ACCESS_KEY_ID, self.AWS_SECRET_ACCESS_KEY, is_secure=False)
+        c = self.S3Connection(self.AWS_ACCESS_KEY_ID, self.AWS_SECRET_ACCESS_KEY, is_secure=AWS_SECURE_CONNECTION)
         return c.get_bucket(self.bucket, validate=False)
```
nyov: Just remove the comment lines, if every change to the codebase got commented, it wouldn't be readable anymore.
Move `AWS_SECURE_CONNECTION = False` to Line 85 and change `is_secure=AWS_SECURE_CONNECTION` to `is_secure=self.AWS_SECURE_CONNECTION`.
An argument to `_get_boto_bucket()` wouldn't work.

Also I would prefer it to read `AWS_SECURE_CONNECTION = True`. Security should have priority over questionable backwards compatibility.
thanks

nyov:
I'm having second thoughts. Perhaps this shouldn't be configurable and require SSL always?
@redapple, @kmike ?

wearpants:
I'd lean towards SSL on by default but be able to turn off with a setting as proposed above. Upload performance to S3 without SSL is significantly faster, and some folks may want that option over security.

redapple:
@nyov , @wearpants, I agree with you here, i.e. enable SSL by default and have a new setting to turn it off for those wanting this (and wanting a backward compatible setup).

We've done something similar with ACL policy for 1.1, writing this in the [release notes](http://doc.scrapy.org/en/latest/news.html):

> When uploading files or images to S3 (with FilesPipeline or ImagesPipeline), the default ACL policy is now “private” instead of “public” Warning: backwards incompatible!. You can use FILES_STORE_S3_ACL to change it.

With proper warning note in the changelog, I think it's fine.

nyov:
Great.
@maksimbormot, looks okay to me now, thank you. (Though didn't you want to enable SSL in the first place?)

@wearpants, that's the second time I hear S3 w/SSL is very slow. I read up on some others confirming that info. Very unfortunate.

redapple:
@maksimbormot , can you implement @kmike 's suggestion of having `AWS_SECURE_CONNECTION` as a setting instead?

redapple:
Also, the name of the setting could be close to the other S3 related one, e.g. `FILES_S3_SECURE_CONNECTION` and `IMAGES_S3_SECURE_CONNECTION`

lopuhin:
There are two code paths in S3FilesStore code: one assumes boto2 and python2, and another is activated if botocore is installed (this is the only variant supported on python 3). Right now, I think only boto2 path is covered here.
For botocore, it's possible to pass ``use_ssl=False`` to ``create_client``: https://github.com/boto/botocore/blob/98c014428956804ecf4523b5da318a67f7d1c7c4/botocore/session.py#L700, I haven't checked how it works though.

redapple:
@maksimbormot , are you still interested in wrapping up this change?
Or would you like someone else to continue?

maksimbormot:
@redapple lets give someone a chance to do it lol

korsunowk:
@redapple I’m up for it and will make the change needed.

wzyboy:
It's 2018 now and 2019 is coming near. Major websites and browsers are pushing "secure by default." I don't `is_secure` should be configurable to users. Even if the very few users who would like to use non-secure connections, they should do it themselves at their own risk because they SHOULD know what they are doing.

Should I make a new PR to simply remove the `is_secure=False` parameter and the comment?

Issue #2071, opened 2016-06-22
In Scrapy 1.1.0 --output as xml appending to existing file when spider re-executed resulting in invalid xml file with duplicate declaration


mobcdi:
## Issue

Scrapy is smart enough not to create duplicate file store copies when spider is re-run but appends output to xml file if `--output myxmlfile.xml` specified in command creating invalid xml file with 2nd declaration statement for each spider run after the xml file is created
## Possible Solution:

If xml file exists overwrite the contents not append to existing file or add option to allow user specify overwrite or append. (While the logfile is appended when a spider is re-run its structure isn't important to valid interpretation by 3rd party tools as is the case with xml files.
## To reproduce:

Run a spider via the command prompt and include option `--output myxmlfile.xml` and log the output to a file `--logfile mylogfile` when spider completes validate the xml using http://codebeautify.org/xmlvalidate and notice the xml is valid.
If you have set-up your spider to use the files pipeline and store a copy of every page indexed you can count  the number of files created in your FILES_STORE and note for later.

Re-run the spider command and notice the xml file increases in size, re-validate the xml and you should receive a similar validation message. 

> Error on->  Line :2 Column :155535 (or how ever many characters your line is before it encounters the 2nd declaration statement)
> Message :XML declaration allowed only at the start of the document
> Error on->  Line :3 Column :1
> Message :Extra content at the end of the document 

Check the FILES_STORE and note that you don't have any increase in number of files stored on disk. Check the log file for references `File (uptodate):` and note the crawler is smart enough to know the page was already created on disk so didn't create a duplicate but when exporting the items to xml it appends the duplicates to the existing file.

redapple:
Related: https://github.com/scrapy/scrapy/issues/547

kmike:
Also related: https://github.com/scrapy/scrapy/pull/2532

wRAR:
> or add option to allow user specify overwrite or append

Done in 2.4.

Issue #547, opened 2014-01-18
Add a command-line option for overwriting exported file


kmike:
What do you think about adding an option to overwrite/recreate exported file?
Something like 

```
$ scrapy crawl myspider -o data.jl --overwrite
```

or 

```
$ scrapy crawl myspider -O data.jl
```

This is useful during development where old data is not needed. I usually run 

```
$ rm data.jl; scrapy crawl myspider -o data.jl
```

multiple times. This is not DRY: if I want to change file name for the next iteration and preserve existing file then I must be careful and update both names (of course, the command comes from shell autocompletion).

alfg:
+1

Any progress on this?

AaronTao1990:
+1 for -o option

nyov:
The current PR looks inadequate to me. Sorry.
As an intermediate step, it should just fix the current issue, which is: appending doesn't work. This would mean producing `filename.out.1 filename.out.2 filename.out.3` filenames, instead.
(This could be the easy task)

Changing the behaviour to append/overwrite, looks more daunting. Doing it right would mean modifying the Exporter's codebase to handle append or overwrite options like a File interface (IMHO); implementation depending on the actual Exporter backend (local disk, s3, datastore, ...).
Also @kmike's proposal
> `scrapy crawl myspider -o combined.jl -O last.jl` can also be useful, and users will try writing this.

would require more thought. It sounds great, but if `--append`/`--overwrite` is not just a switch to `-o`, but an argument taking a destination, Exporters would have to be able to handle that (opening multiple "filepointers" as given). That doesn't warrant an `easy` tag, I think.

jschilling1:
I've been using a custom CsvItemExporter to overwrite existing exports and use a custom CSV_DELIMITER in settings. Now i need to set the FEED_URI in the spider. The problem with that is that my custom CsvItemExporter no longer clears existing exports which it only does on init.

dannykopping:
You can simply output to stdout and redirect that output to a file:
scrapy runspider spider.py -t json --nolog -o - > out.json

lucaspottersky:
> You can simply output to stdout and redirect that output to a file:

@dannykopping this is not an option when scheduling through scrapyd.

robkorv:
In the meantime one could subclass `FileFeedStorage`. I just typed an answer over at http://stackoverflow.com/a/40823149/2546958 before I spotted this issue.

PR #2093, opened 2016-07-05
Create new files in case of feed export output file path conflict

redapple:
Should (eventually) address #547 and #2071 

A few cases illustrating chosen implementation:

* `-o test.json` and `test.json` already exists --> feed exporter writes to `test.json.1`
* `-o test.json`, and `test.json` and `test.json.1` already exist --> feed exporter writes to `test.json.2`
* `-o test.json`, and `test.json` and `test.json.5` already exist, but not `test.json.1`..`test.json.4` --> feed exporter writes to `test.json.6`, not `test.json.1` (arguable choice)
* `-o test.json.1`, and `test.json` and `test.json.1` already exist --> feed exporter writes to `test.json.1.1`, not `test.json.2` (arguable choice)

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -5,6 +5,7 @@ See documentation in docs/topics/feed-exports.rst
 """
 
 import os
+import re
 import sys
 import logging
 import posixpath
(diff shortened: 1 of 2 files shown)


kmike:
Unlike adding -O option this change is backwards incompatible. For example, someone may use `-o filename` and then read from this filename; now it will return stale data.

redapple:
Backward incompatible indeed.

Issue #2092, opened 2016-07-05
Scrapy 1.1 - exceptions.ValueError: Invalid DNS-ID.


nealhnguyen:
Hello I'm crawling websites with insecure connections. When I try to crawl, I get the stack trace,

```
2016-07-05 15:50:17 [twisted] CRITICAL: Error during info_callback
Traceback (most recent call last):
  File "c:\python27\lib\site-packages\twisted\protocols\tls.py", line 421, in dataReceived
    self._write(bytes)
  File "c:\python27\lib\site-packages\twisted\protocols\tls.py", line 569, in _write
    sent = self._tlsConnection.send(toSend)
  File "c:\python27\lib\site-packages\OpenSSL\SSL.py", line 1270, in send
    result = _lib.SSL_write(self._ssl, buf, len(buf))
  File "c:\python27\lib\site-packages\OpenSSL\SSL.py", line 933, in wrapper
    callback(Connection._reverse_mapping[ssl], where, return_code)
--- <exception caught here> ---
  File "c:\python27\lib\site-packages\twisted\internet\_sslverify.py", line 1154, in infoCallback
    return wrapped(connection, where, ret)
  File "c:\python27\lib\site-packages\scrapy\core\downloader\tls.py", line 45, in _identityVerifyingInfoCallback
    verifyHostname(connection, self._hostnameASCII)
  File "c:\python27\lib\site-packages\service_identity\pyopenssl.py", line 45, in verify_hostname
    obligatory_ids=[DNS_ID(hostname)],
  File "c:\python27\lib\site-packages\service_identity\_common.py", line 245, in __init__
    raise ValueError("Invalid DNS-ID.")
exceptions.ValueError: Invalid DNS-ID.
```

Is there any way to ignore the invalid certificate?

redapple:
@nealhnguyen , thanks for reporting.
This exception is not currently caught, only `VerificationError` is.
But we can add something to handle this case. And it would help a lot if you could tell us which website this happened with so we can test a patch (you can send me this info by email if you don't want to disclose it)

Until then, you'll have to use a custom [`DOWNLOADER_CLIENTCONTEXTFACTORY`](http://doc.scrapy.org/en/latest/topics/settings.html#downloader-clientcontextfactory) which catches `ValueError` and ignores it.

redapple:
Ok, I'm able to reproduce when using IP addresses directly, for example with https://www.python.org

```
$ nslookup www.python.org
Non-authoritative answer:
[...]

PR #2094, opened 2016-07-06
Catch and ignore certification verification exception for IP-address hosts

redapple:
Fixes GH-2092

FTR, this is what Twisted 16.3.0 gives with default `Agent` (for www.google.com at 216.58.208.132)

```
$ python
Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
[GCC 4.9.2] on linux2
>>> from twisted.internet import reactor
>>> from twisted.web.client import Agent
>>> from twisted.web.http_headers import Headers
>>> 
>>> agent = Agent(reactor)
>>> 
>>> d = agent.request(
...     'GET',
...     'https://216.58.208.132:443/',
...     Headers({'User-Agent': ['Twisted Web Client Example']}),
...     None)
>>> 
>>> 
>>> def cbResponse(ignored):
...     print('Response received')
... 
>>> 
>>> d.addCallback(cbResponse)
<Deferred at 0x7fc71fd89e18>
>>> 
>>> def cbShutdown(ignored):
...     reactor.stop()
... 
>>> 
>>> d.addBoth(cbShutdown)
<Deferred at 0x7fc71fd89e18>
>>> 
>>> reactor.run()
Error during info_callback
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/twisted/protocols/tls.py", line 423, in dataReceived
    self._write(bytes)
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/twisted/protocols/tls.py", line 571, in _write
    sent = self._tlsConnection.send(toSend)
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/OpenSSL/SSL.py", line 1270, in send
    result = _lib.SSL_write(self._ssl, buf, len(buf))
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/OpenSSL/SSL.py", line 933, in wrapper
    callback(Connection._reverse_mapping[ssl], where, return_code)
--- <exception caught here> ---
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/twisted/internet/_sslverify.py", line 1154, in infoCallback
    return wrapped(connection, where, ret)
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/twisted/internet/_sslverify.py", line 1253, in _identityVerifyingInfoCallback
    verifyHostname(connection, self._hostnameASCII)
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/service_identity/pyopenssl.py", line 45, in verify_hostname
    obligatory_ids=[DNS_ID(hostname)],
  File "/home/paul/.virtualenvs/scrapy10/local/lib/python2.7/site-packages/service_identity/_common.py", line 245, in __init__
    raise ValueError("Invalid DNS-ID.")
exceptions.ValueError: Invalid DNS-ID.

>>> 
```

And `curl`'s default (secure, i.e. without `-k`):

```
$ curl -v https://216.58.208.132:443
* Rebuilt URL to: https://216.58.208.132:443/
*   Trying 216.58.208.132...
* Connected to 216.58.208.132 (216.58.208.132) port 443 (#0)
* found 174 certificates in /etc/ssl/certs/ca-certificates.crt
* found 708 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_128_GCM_SHA256
* 	 server certificate verification OK
* 	 server certificate status verification SKIPPED
* SSL: certificate subject name (www.google.com) does not match target host name '216.58.208.132'
* Closing connection 0
curl: (51) SSL: certificate subject name (www.google.com) does not match target host name '216.58.208.132'
```

### DIFF
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -34,9 +34,15 @@ try:
                                              VerificationError)
 
     class ScrapyClientTLSOptions(ClientTLSOptions):
-        # same as Twisted's ClientTLSOptions,
-        # except that VerificationError is caught
-        # and doesn't close the connection
+        """
+        SSL Client connection creator ignoring certificate verification errors
+        (for genuinely invalid certificates or bugs in verification code).
+
+        Same as Twisted's private _sslverify.ClientTLSOptions,
+        except that VerificationError and ValueError exceptions are caught,
+        so that the connection is not closed, only logging warnings.
+        """
+
         def _identityVerifyingInfoCallback(self, connection, where, ret):
             if where & SSL_CB_HANDSHAKE_START:
                 _maybeSetHostNameIndication(connection, self._hostnameBytes)
@@ -50,8 +56,9 @@ try:
 
                 except ValueError as e:
                     logger.warning(
-                        'Ignoring remote certificate verification failure for hostname "{}"; {}'.format(
-                            self._hostnameASCII, e))
+                        'Ignoring error while verifying certificate '
+                        'from host "{}" (exception: {})'.format(
+                            self._hostnameASCII, repr(e)))
 
 except ImportError:
     # ImportError should not matter for older Twisted versions
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -50,7 +50,7 @@ try:
 
                 except ValueError as e:
                     logger.warning(
-                        'SSL/TLS verification failed for hostname "{}"; {}'.format(
+                        'Ignoring remote certificate verification failure for hostname "{}"; {}'.format(
                             self._hostnameASCII, e))
 
 except ImportError:
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -48,6 +48,11 @@ try:
                         'Remote certificate is not valid for hostname "{}"; {}'.format(
                             self._hostnameASCII, e))
 
+                except ValueError as e:
+                    logger.warning(
+                        'SSL/TLS verification failed for hostname "{}"; {}'.format(
+                            self._hostnameASCII, e))
+
 except ImportError:
     # ImportError should not matter for older Twisted versions
     # as the above is not used in the fallback ScrapyClientContextFactory
--- a/tests/test_downloader_handlers.py
+++ b/tests/test_downloader_handlers.py
@@ -335,6 +335,14 @@ class Https11WrongHostnameTestCase(Http11TestCase):
     certfile = 'keys/example-com.cert.pem'
 
 
+class Https11InvalidDNSId(Https11TestCase):
+    """Connect to HTTPS hosts with IP while certificate uses domain names IDs."""
+
+    def setUp(self):
+        super(Https11InvalidDNSId, self).setUp()
+        self.host = '127.0.0.1'
+
+
 class Http11MockServerTestCase(unittest.TestCase):
     """HTTP 1.1 test case with MockServer"""
     if twisted_version < (11, 1, 0):


scrapy/core/downloader/tls.py
```diff
@@ -48,6 +48,11 @@ def _identityVerifyingInfoCallback(self, connection, where, ret):
                         'Remote certificate is not valid for hostname "{}"; {}'.format(
                             self._hostnameASCII, e))
 
+                except ValueError as e:
```
kmike: The comment after ScrapyClientTLSOptions definition should be updated.

scrapy/core/downloader/tls.py
```diff
@@ -48,6 +48,11 @@ def _identityVerifyingInfoCallback(self, connection, where, ret):
...
+                    logger.warning(
+                        'SSL/TLS verification failed for hostname "{}"; {}'.format(
+                            self._hostnameASCII, e))
```
kmike: It looks like the problem is that ip address is not a hostname, and Twisted passes it to [verify_hostname](https://service-identity.readthedocs.io/en/stable/api.html#service_identity.pyopenssl.verify_hostname). 

So verification is failed, but it is not failed because SSL/TLS algorithm tells it should fail, but arguably because of a twisted bug (or service_identity inflexibility).

redapple: It's correct that the trouble comes from passing the IP address to verify_hostname.
[SubjectAltName apparently can include IP addresses](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sssd-ldap-domain-ip.html) but we need to check what happens when remote certificate does include it, matching what the client connected to.
The error with scrapy 1.0.6 looks identical but the agent implementation made it "work".
This PR is definitely not pretty.
Would you suggest coming up with a proper verification implementation?

kmike: I'm not sure it is Scrapy's job to implement verification - this code is better fit in Twisted or service_identity. I think we should figure out where this bug belongs, and document it as a bug, not as SSL/TLS verification failure.

redapple: Makes sense.

kmike: I didn't mean we shouldn't catch this error; maybe just a log message should be different :)

redapple: alright, what message do you suggest?

redapple: @kmike , how about `'Ignoring remote certificate verification failure for hostname "{}"; {}'.format(self._hostnameASCII, e))`, without the SSL mention?

kmike: @redapple I have 2 reservations about the error message:

1. 'certificate verification failure' can be read as 'wrong certificate' or a 'misconfigured remote website' or 'wrong configuration of local verification software'. But here it is not a bad certificate or old openssl or a problem with a remote website, it is just a bug in software; because of a bug the verification is not even attempted. I don't know how to express that well in error message though.
2. it says `'hostname {}'.format(self._hostnameASCII)`, but _hostnameASCII is not a hostname here - it is an IP, and this is precisely what causes the error. 

I think we shouldn't describe the actual bug in the error message because ValueError is too broad - it could happen other issue will cause ValueError in future, or maybe it can be triggered now by something else.

What about `"Ignoring certificate verification error for {}; {}".format(self._hostnameASCII, e)`? It doesn't fully address (1).

redapple: OK.
What about "host" or "netloc"?
And something like "Ignoring error while verifying certificate from host {}
(exception: {})" maybe,
with target host and exception
Le 13 juil. 2016 12:28 AM, "Mikhail Korobov" <notifications@github.com> a
écrit :

kmike: Sounds good 👍 
Sorry for holding it back!

Issue #2035, opened 2016-06-06
LOG_UNSERIALIZABLE_REQUESTS messages should be WARNINGs, not ERRORs


kmike:
Currently if LOG_UNSERIALIZABLE_REQUESTS is True and request can't be serialized users get

> ERROR: Unable to serialize request: ...

messages in log (see [here](https://github.com/scrapy/scrapy/blob/d3ced85ec350e8c195760a8ce975a18ca42f29a0/scrapy/core/scheduler.py#L81)). I think they are confusing because spider still works. What do you think about changing log level to WARNING?

eliasdorneles:
+1

nyov:
+1

PR #2129, opened 2016-07-17
Fixing issue #2035

mani1996:
Changed log type to warning instead of error

### DIFF
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -86,7 +86,7 @@ class Scheduler(object):
             self.dqs.push(reqd, -request.priority)
         except ValueError as e: # non serializable request
             if self.logunser:
-                logger.error("Unable to serialize request: %(request)s - reason: %(reason)s",
+                logger.warning("Unable to serialize request: %(request)s - reason: %(reason)s",
                              {'request': request, 'reason': e},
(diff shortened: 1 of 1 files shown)


mani1996:
What is meant by "0.00% of diff hit" ? Is there something I need to change in my code?

redapple:
@mani1996 , it means that there's no test covering your change.
If you want, you can try adding a test case for this, but I doubt it's crucial

kmike:
The PR looks good to me, thanks @mani1996! But it seems something is changed in scrapy master, so it can't be merged as-is - could you please rebase it to make mergeable?

scrapy/core/scheduler.py
```diff
@@ -86,7 +86,7 @@ def _dqpush(self, request):
             self.dqs.push(reqd, -request.priority)
         except ValueError as e: # non serializable request
             if self.logunser:
-                logger.error("Unable to serialize request: %(request)s - reason: %(reason)s",
+                logger.warning("Unable to serialize request: %(request)s - reason: %(reason)s",
```
Digenis: indent the next 2 lines by 2 spaces each

redapple:
This change has been applied on top of master in https://github.com/scrapy/scrapy/pull/2269
Thanks again @mani1996

PR #2032, opened 2016-06-06
Update tutorial.rst example

matt-oconnell:
"Following Links" example is currently out of date. This update works with the current version of `http://www.dmoz.org/Computers/Programming/Languages/Python/`

PR #2130, opened 2016-07-17
Some changes to the tutorial to reflect the new structure of the dmoz…

UltraBob:
….org site

The structure of the dmoz.org site has changed somewhat recently, and this updates the tutorial to reflect it.  This is related to pull request #2032 but is more complete, updating all the examples and explanation.  I also had a pull request ready for scrapy/dirbot, but [This pull request](https://github.com/scrapy/dirbot/pull/21/commits/6d0318e0d86b41d13bdd411c1be8704fadf5298f) does more to fix the python 3 compatibility issues than my patch would have.  I've done my best to explain, but keep the selectors short and simple so as not to overwhelm the reader.  Therefore, I think the selectors used in this request are more in line with the spirit of the existing tutorial.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -338,7 +338,7 @@ And from them, the site's descriptions::
 
     response.xpath('//div[@class="title-and-desc"]/div/text()').extract()
 
-Or to strip out the extra tabs, linebreaks, and whitespace around the text in the descriptions::
+Alternatively, to strip out the extra tabs, linebreaks, and whitespace around the text in the descriptions::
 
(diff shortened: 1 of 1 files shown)


kmike:
Thanks for your help @UltraBob, good work! We've decided to rewrite the tutorial to use http://toscrape.com instead of DMOZ to avoid such problems in future.

kmike:
I really appreciate the help with tutorial; it took us too much time to finally pay attention to it, but hopefully it will be more robust in future. Your PR was good, but we haven't merged it because we decided to use another solution (i.e. don't try to fix DMOZ scraping).

UltraBob:
Cheers, that seems like a better approach!

PR #2135, opened 2016-07-21
Add --editor option for "scrapy edit" command

Granitas:
Option --editor allows you to set editor everytime you want ot edit a spider. 

Personally I find using different editor (either vim or pycharm in my case) depending on how big of a change I'm looking to implement, i.e. for changing 2 lines of code I'm not gonna load whole IDE.   
With the current system you either need to manipulate the $EDITOR environment variable or set scrapy setting everytime you want to use a different editor which is tedious and require various shell hacks to do quickly.

### DIFF
--- a/scrapy/commands/edit.py
+++ b/scrapy/commands/edit.py
@@ -1,21 +1,27 @@
-import sys, os
+import sys
+import os
 
 from scrapy.commands import ScrapyCommand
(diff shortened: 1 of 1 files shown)


Granitas:
Wow, I actually missed global --set option. 
This PR can be resolved with `scrapy edit --set=EDITOR=vim spidername` ...
closing :(

PR #2138, opened 2016-07-21
[MRG+1] Apply the FILES_STORE_S3_ACL setting in ImagesPipeline

jesuslosada:
All images uploaded to S3 using the ImagesPipeline class have the default ACL policy (private), because ImagesPipeline is not respecting the FILES_STORE_S3_ACL setting.

This setting is correctly used in FilesPipeline: https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/files.py#L258

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -117,7 +117,7 @@ Supported Storage
 =================
 
 File system is currently the only officially supported storage, but there is
-also (undocumented) support for storing files in `Amazon S3`_.
+also support for storing files in `Amazon S3`_.
 
 .. _Amazon S3: https://aws.amazon.com/s3/
 
@@ -146,6 +146,23 @@ Where:
 * ``full`` is a sub-directory to separate full images from thumbnails (if
   used). For more info see :ref:`topics-images-thumbnails`.
 
+Amazon S3 storage
+-----------------
+
+.. setting:: FILES_STORE_S3_ACL
+.. setting:: IMAGES_STORE_S3_ACL
+
+:setting:`FILES_STORE` and :setting:`IMAGES_STORE` can represent an Amazon S3
+bucket. Scrapy will automatically upload the files to the bucket.
+
+For example, this is a valid :setting:`IMAGES_STORE` value::
+
+    IMAGES_STORE = 's3://bucket/images'
+
+You can modify the Access Control List (ACL) policy used for the stored files,
+which is defined by the :setting:`FILES_STORE_S3_ACL` and
+:setting:`IMAGES_STORE_S3_ACL` settings. By default, the ACL is set to
+``private``.
 
 Usage example
 =============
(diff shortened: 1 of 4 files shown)


scrapy/pipelines/images.py
```diff
@@ -88,6 +88,8 @@ def from_settings(cls, settings):
         s3store = cls.STORE_SCHEMES['s3']
         s3store.AWS_ACCESS_KEY_ID = settings['AWS_ACCESS_KEY_ID']
         s3store.AWS_SECRET_ACCESS_KEY = settings['AWS_SECRET_ACCESS_KEY']
+        s3store.POLICY = settings['FILES_STORE_S3_ACL']
```
redapple: I believe the name should be `IMAGES_STORE_S3_ACL` to be consistent with `IMAGES_STORE`. If so, it should also be documented.

jesuslosada: Good point! I thought the name was a reference to the S3FilesStore class, but the setting is only applied in FilesPipeline so it makes sense to create a different setting for ImagesPipeline.

I've just realised that FILES_STORE_S3_ACL is documented on the main settings page, but the other settings related to these pipelines (FILES_STORE, IMAGES_STORE, FILES_URLS_FIELD, etc.) are documented in the media pipeline section. Do you want to move FILES_STORE_S3_ACL (and the new IMAGES_STORE_S3_ACL setting) to the media pipeline section? Or do you prefer to add the new setting in the main settings page too? @redapple @kmike

redapple: I'd go for moving `FILES_STORE_S3_ACL` to `docs/topics/media-pipeline.rst`, and document `IMAGES_STORE_S3_ACL` there too.

jesuslosada:
I've added the IMAGES_STORE_S3_ACL setting as previously agreed. I've also updated the documentation. Feel free to suggest any improvements to the documentation.

docs/topics/media-pipeline.rst
```diff
@@ -146,6 +146,23 @@ Where:
...
+bucket. Scrapy will automatically upload the files to the bucket.
+
+For example, this is a valid :setting:`IMAGES_STORE` value::
+
+    IMAGES_STORE = 's3://bucket/images'
+
+You can modify the Access Control List (ACL) policy used for the stored files,
+which is defined by the :setting:`FILES_STORE_S3_ACL` and
+:setting:`IMAGES_STORE_S3_ACL` settings. By default, the ACL is set to
+``private``.
```
kmike: It'd be nice to explain how to make it public.

jesuslosada:
I've added documentation on how to make the files publicly available and a link to the ACLs reference on the Amazon website.

redapple:
Thanks a bunch @jesuslosada !

redapple:
Needs backporting to 1.1 branch

pawelmhm:
Hmm so if in my project I had public-read as access level and I was ok with that your PR will change this to private, right? So all users who were ok with public access will need to set new setting in their projects settings. Wouldn't it be better to keep old setting as default?

PR #2140, opened 2016-07-22
[MRG+1] Fix IMAGES_EXPIRES default value

jesuslosada:
It looks like commit 4cef1a1d0060148716ffec4ad6ce9bf709ad1ea2 introduced a little bug. According to the documentation IMAGES_EXPIRES should be 90.

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -42,7 +42,7 @@ class ImagesPipeline(FilesPipeline):
     # ImagesPipeline. They may be overridden by settings.
     MIN_WIDTH = 0
     MIN_HEIGHT = 0
-    EXPIRES = 0
+    EXPIRES = 90
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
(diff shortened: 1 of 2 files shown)


redapple:
@jesuslosada , I agree that `IMAGES_EXPIRES` should be `90`, and if it's not, the tests are missing one case.
But I admit I'm having a hard time re-reading https://github.com/scrapy/scrapy/pull/1989 and especially https://github.com/scrapy/scrapy/pull/1989/files#diff-0441c35e69fd9bb28942a9b2c47dc565R43 where not only `IMAGES_EXPIRES` but `MIN_WIDTH`, `MIN_HEIGHT` and others set to `0`
Maybe @pawelmhm can comment on this one.

pawelmhm:
> where not only IMAGES_EXPIRES but MIN_WIDTH, MIN_HEIGHT and others set to 0

MIN_WIDTH and MIN_HEIGHT should be zero. These are the values that were in old pipeline before this change: https://github.com/scrapy/scrapy/pull/1891/files

expires should be 90 and it's a bug my mistake sorry

redapple:
Thanks for the feedback @pawelmhm !

redapple:
Needs backporting to 1.1 branch.

PR #2152, opened 2016-07-29
[MRG+1] Better bumpversion config for branch 1.1

eliasdorneles:
Having the part:prerel setup as it's right now, it's forcing releases to go through a pre-release cycle dev1, rc1, before the actual release.
That is not desired for older branches, since they will have mostly minor and patch releases.

This disables that behavior, making minor and patch releases for this branch more predictable.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,20 +3,5 @@ current_version = 1.1.1
 commit = True
 tag = True
 tag_name = {new_version}
-parse = ^
-	(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
-	(?:(?P<prerel>[abc]|rc|dev)(?P<prerelversion>\d+))?
-serialize = 
-	{major}.{minor}.{patch}{prerel}{prerelversion}
-	{major}.{minor}.{patch}
 
 [bumpversion:file:scrapy/VERSION]
-
-[bumpversion:part:prerelversion]
-values = 
-	1
-	2
-	3
-	4
-	5
-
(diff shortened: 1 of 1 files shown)


eliasdorneles:
@redapple do you think this is good enough for now?
It'll ease the pain of releasing in this branch, while still leaving the enforced release cycle in `master`.
Would you prefer to have this in master too?

redapple:
@eliasdorneles , thanks, it's a good start :)
Let's see how it goes for master branch after 1.2 release.

.bumpversion.cfg
```diff
@@ -12,13 +12,6 @@ serialize =
...
 [bumpversion:file:scrapy/VERSION]
 
-[bumpversion:part:prerel]
-optional_value = gamma
-values = 
-	dev
-	rc
-	gamma
-
 [bumpversion:part:prerelversion]
```
redapple: @eliasdorneles , we could get rid of `{prerelversion}` and `[bumpversion:part:prerelversion]` section too, I suppose.

eliasdorneles: okay, perhaps let's rid of the whole prerelease configuration, then.

eliasdorneles: Updated, leaving the defaults -- should be good enough for the branches.
What do you think?

redapple: LGTM. Thanks @eliasdorneles

eliasdorneles:
I've thought a bit more and I propose doing the same for master too: https://github.com/scrapy/scrapy/pull/2159

redapple:
LGTM (this one and the master branch change)

Issue #2146, opened 2016-07-27
Response blank after being assigned to a Selector


eLRuLL:
```
$ scrapy shell "http://httpbin.org/ip"

...

In [1]: print response
<200 http://httpbin.org/ip>

In [2]: sel = Selector(text='', response=response)

In [3]: print sel.response
<200 about:blank>
```

kmike:
Why do you pass both text and response arguments? It seems we should raise an error in this case in [Selector](https://github.com/scrapy/scrapy/blob/fe68a45c82e6126b411784d57a70e7aba89f696d/scrapy/selector/unified.py#L43) constructor.

eLRuLL:
I want to use the the `text` of the selector for my xpath queries, but still be able to know which response it comes from, for example dividing a response body on different segments, where I will create a selector per segment.

eLRuLL:
what about just passing the `url` to the selector? so I can know at least which url the original response was related to

kmike:
@eLRuLL this usage of `text` argument was never intended AFAIK. You can pass url in `base_url` argument: scrapy.Selector wraps [parsel.Selector](https://github.com/scrapy/parsel/blob/2060660dd3f331e410b740bc55a3879eaf984eef/parsel/selector.py#L141) to 1) provide backwards compatibility and 2) an ability to pass response instead of text.

Digenis:
@eLRuLL, use `response.replace(body=...`.
I can only think of 3 reasons one would use separate selectors for different segments:
The document size is impacting performance and you can more efficiently parse only a fragment.
You want to modify parts of the document.
Most of the mark-up is broken but the interesting segments can be parsed.
Does yours fall into one of them?

@kmike's concern is valid. Here's a patch: https://github.com/Digenis/scrapy/commit/b57de97a22911c1c77fbefef45450f0cfadaf120
I find it a bit verbose, I typically just do an assertion,
let the traceback lead to some head scratching
and blame python for quack typing.
Wouldn't it be better?

kmike:
@Digenis your patch looks fine.

kmike:
Fixed by #2153 - thanks @Digenis!

eLRuLL:
thanks @Digenis , yeah mine was the first:

> The document size is impacting performance and you can more efficiently parse only a fragment.

I haven't tried it but I think using separate responses could do it

PR #2153, opened 2016-07-30
[MRG+1] Selector should not receive both response and text

Digenis:
Continued from #2146


eliasdorneles:
Looks good, thanks @Digenis !

PR #2159, opened 2016-08-01
[MRG+1] Remove bumpversion prerelease configuration

eliasdorneles:
I propose to remove the prerelease configuration from bumpversion, because I think its behavior is just too confusing.

The rationale for this is that making the release procedure predictable is more important than facilitating making pre-releases, which are sort of the exception in the workflow.

The current configuration makes most common cases confusing:
- bug fix releases require you have to remember to use `--serialize "{major}.{minor}.{patch}"`
- to start a pre-release cycle, you actually use `minor` or `patch`
- to do the actual minor or patch release, you use `prerel`

Also, `prerel` breaks if you run it on a branch with a final release, because it can't parse the prerelease information.

Therefore, I propose keeping the bumpversion defaults, and do the prereleases (dev1, dev2, rc1, etc) manually (with `--new-version`), which makes for a more predictable and intuitive behavior.
- `bumpversion minor` and `bumpversion patch` will work as expected
- pre-releases will be manually handled, but this seems a small overhead
  than remembering the details I mention above.

If you're happy with this, I'll also update [the wiki](https://github.com/scrapy/scrapy/wiki/Scrapy-release-procedure) with new instructions.

What do you think?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,27 +3,5 @@ current_version = 1.2.0dev2
 commit = True
 tag = True
 tag_name = {new_version}
-parse = ^
-	(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
-	(?:(?P<prerel>[abc]|rc|dev)(?P<prerelversion>\d+))?
-serialize = 
-	{major}.{minor}.{patch}{prerel}{prerelversion}
-	{major}.{minor}.{patch}
 
 [bumpversion:file:scrapy/VERSION]
-
-[bumpversion:part:prerel]
-optional_value = gamma
-values = 
-	dev
-	rc
-	gamma
-
-[bumpversion:part:prerelversion]
-values = 
-	1
-	2
-	3
-	4
-	5
-


eliasdorneles:
To make the proposal more concrete, I've put in a gist what would be the new instructions for the release procedures in the wiki: https://gist.github.com/eliasdorneles/129447468ce61a2f6e3886dde3c14a41
Let me know if that makes sense.

redapple:
I would suggest mentioning the major and minor versions release procedure first, and then talk about the RC case.
You can even use 1.2 as base (and releasing 1.3.0rc1 as example for RCs)

eliasdorneles:
Hm, won't the major and minor only make sense after explaining the branching approach taken with the RCs?

eliasdorneles:
By "major and minor", I actually meant "minor and patch".

redapple:
Right, the branching approach needs to be explained and then illustrated with examples.
Actually, on the branching approach, I'm not sure we always need branching out when we do release candidates. We could, in the general case,:
- keep working with master branch,
- force version to some RC number initially, 
- increment along the way, 
- until there's a backward compatible change to be merged (or more generally, some change we do not want in the next release that the RC is going to become).

The pain with 1.1 branch and master 1.2dev is that we branched out very (too?) early, with master and 1.1 being essentially the same until recently.

kmike:
@redapple @eliasdorneles this is all voodoo for me, feel free to merge :)

PR #2166, opened 2016-08-05
[MRG+1] Temporarily deprecate official Ubuntu packages

### LINKED ISSUE #2076: Ubuntu official repository - installation failed (not able to find python-support)
I used the [manual](http://doc.scrapy.org/en/latest/topics/ubuntu.html) to install scrapy on ubuntu 16.04, but it failed because it was not able to install python-support (>= 0.90.0). Other sources report that this package is not part of the new ubuntu xenial anymore.

Quick&Dirty-Workaround:
```
wget http://launchpadlibrarian.net/109052632/python-support_1.0.15_all.deb
sudo dpkg -i python-support_1.0.15_all.deb
sudo apt-get update && sudo apt-get install scrapy
```

 * http://askubuntu.com/questions/766169/why-no-more-python-support-in-16-04
 * https://launchpad.net/ubuntu/xenial/amd64/python-support/1.0.15

### LINKED ISSUE #2137: DOC Deprecate official Ubuntu packages (until they get updated)
Current official Ubuntu packages are not up-to-date with scrapy 1.1 and users are having trouble with them (see #2076, #2136)

The docs need to reflect that and not recommend using them.

This page http://doc.scrapy.org/en/latest/topics/ubuntu.html#topics-ubuntu is not relevant these days.

ashkulz:
They are not currently updated and fail to install on Ubuntu 16.04

fixes #2137 and closes #2076

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -89,15 +89,14 @@ Windows
      Python 3 is not supported on Windows. This is because Scrapy core requirement Twisted does not support
      Python 3 on Windows.
 
-Ubuntu 9.10 or above
---------------------
+Ubuntu 12.04 or above
+---------------------
 
 **Don't** use the ``python-scrapy`` package provided by Ubuntu, they are
-typically too old and slow to catch up with latest Scrapy.
+typically too old and slow to catch up with latest Scrapy. You will need
+to install the following system packages::
 
-Instead, use the official :ref:`Ubuntu Packages <topics-ubuntu>`, which already
-solve all dependencies for you and are continuously updated with the latest bug
-fixes.
+    sudo apt-get install python-pip python-twisted python-openssl python-lxml
 
 If you prefer to build the python dependencies locally instead of relying on
 system packages you'll need to install their required non-python dependencies
--- a/docs/topics/ubuntu.rst
+++ b/docs/topics/ubuntu.rst
@@ -11,6 +11,9 @@ those in Ubuntu, and more stable too since they're continuously built from
 `GitHub repo`_ (master & stable branches) and so they contain the latest bug
 fixes.
 
+.. caution:: These packages are currently not updated and may not work on
+   Ubuntu 16.04 and above, see :issue:`2076` and :issue:`2137`.
+
 To use the packages:
 
 1. Import the GPG key used to sign Scrapy packages into APT keyring::


docs/topics/ubuntu.rst
```diff
@@ -11,6 +11,9 @@ those in Ubuntu, and more stable too since they're continuously built from
 `GitHub repo`_ (master & stable branches) and so they contain the latest bug
 fixes.
 
+.. caution:: These packages are currently not updated and may not work on
+   `Ubuntu 16.04`_ and above, see `#2137`_.
```
redapple: I suggest you use scrapy's sphinx extension for issues
```
:issue:`2137`
```

docs/intro/install.rst
```diff
@@ -95,10 +95,6 @@ Ubuntu 9.10 or above
 **Don't** use the ``python-scrapy`` package provided by Ubuntu, they are
 typically too old and slow to catch up with latest Scrapy.
 
-Instead, use the official :ref:`Ubuntu Packages <topics-ubuntu>`, which already
-solve all dependencies for you and are continuously updated with the latest bug
-fixes.
-
 If you prefer to build the python dependencies locally instead of relying on
```
redapple: With the removal of the recommendation for Ubuntu packages, _"If you prefer to build the python dependencies locally"_ sounds a bit weird since the dependencies are now not mentioned earlier.
I believe they should be listed at a minimum (`twisted`, `cryptography`, `lxml` may be sufficient)

A link to their respective install pages for ubuntu would also help users when troubleshooting, and shed light on the list of non-python dependencies listed below.

redapple:
Thanks @ashkulz !
I made a few comments.

ashkulz:
@redapple: I've used the syntax for referring to issues and have reworded the installation instructions so that the specific system packages are mentioned explicitly. Let me know if you have any further feedback.

redapple:
LGTM. Thanks @ashkulz

ashkulz:
Anything holding it from being merged?

docs/intro/install.rst
```diff
@@ -89,15 +89,14 @@ Windows
...
 
 **Don't** use the ``python-scrapy`` package provided by Ubuntu, they are
-typically too old and slow to catch up with latest Scrapy.
+typically too old and slow to catch up with latest Scrapy. You will need
+to install the following system packages::
 
-Instead, use the official :ref:`Ubuntu Packages <topics-ubuntu>`, which already
-solve all dependencies for you and are continuously updated with the latest bug
-fixes.
+    sudo apt-get install python-pip python-twisted python-openssl python-lxml
```
kmike: Instaling these packages won't get Scrapy installed.

redapple: How do you mean @kmike ? As I understand it, this sentence is about the system dependencies needed prior to installing scrapy.
Just after, there's:
> "You can install Scrapy with ``pip`` after that"

kmike: hm, right, I thought "You can install Scrapy with pip after that" is only for the second set of packages. 

I'd drop `python-pip python-twisted python-openssl python-lxml` instructions though because even though they have a slight advantage of installing faster, package versions can be outdated, it won't work in Python 3, and it won't work in virtualenv. 

by the way, `pip install Scrapy` won't work with either set on system dependencies - users need to either create a virtualenv (a great idea), use `sudo pip install Scrapy` (a bad idea), or use `pip install --user Scrapy` (a good idea, but requires some effort to get `scrapy` command-line tool working).

redapple: +1 to recommend creating a virtualenv

kmike: I don't think we should recommend a system-wide way to install scrapy.

dangra: I agree with @kmike about not recommending system-wide installs. Only venv or --user

redapple: Are you ok with merging this and improving on top of it?

redapple:
@kmike , @dangra , @eliasdorneles ,
I've created https://github.com/scrapy/scrapy/pull/2267 with @ashkulz 's change to add installation guidelines (cf. https://github.com/scrapy/scrapy/pull/2166#r79451738).
It's using a new branch on the scrapy repo, `deprecate-ubuntu-packages`, so that you can commit any changes directly.

PR #2168, opened 2016-08-08
[MRG+1] Use w3lib.url.canonicalize_url() from w3lib 1.15.0

### LINKED ISSUE #2157: Use w3lib.url.canonicalize_url()
[w3lib v1.15](https://github.com/scrapy/w3lib/releases/tag/v1.15.0) now has a canonicalize_url() (same code as scrapy's in fact).
Let's depend on w3lib>=1.15 and use `canonicalize_url` from there.

ashkulz:
fixes #2157


kmike:
Thanks! 

It seems a few other places needs to be changed: 

* setup.py file needs a version bump for w3lib;
* it looks like some code in scrapy.utils.url was only required for canonicalize_url - I wonder if we can drop it now.

ashkulz:
Thanks, incorporated your feedback. Did a rebase, as not sure whether an additional commit would be acceptable (might be a good idea to incorporate in `docs/contributing.rst`).

kmike:
@ashkulz: it is good to have a readable commits history, thanks for taking time to do the rebase! 

Currently we are not too strict on rebase vs extra commits - sometimes we ask for a rebase when a commit history gets messy, but not always. Probably that's why it is not in contributing.rst - we haven't agreed on a policy, it is just a common sense now.

scrapy/utils/url.py
```diff
@@ -7,18 +7,12 @@
...
-                                    urlparse, parse_qsl, urlencode,
-                                    quote, unquote)
-if not six.PY2:
-    from urllib.parse import unquote_to_bytes
+from six.moves.urllib.parse import (ParseResult, urldefrag, urlparse)
 
 # scrapy.utils.url was moved to w3lib.url and import * ensures this
 # move doesn't break old code
 from w3lib.url import *
-from w3lib.url import _safe_chars
```
kmike: I believe this import is needed for backwards compatibility

scrapy/utils/url.py
```diff
@@ -40,121 +34,6 @@ def url_has_any_extension(url, extensions):
...
-    uqp = _unquotepath(path)
-    path = quote(uqp, _safe_chars) or '/'
-
-    fragment = '' if not keep_fragments else fragment
-
-    # every part should be safe already
-    return urlunparse((scheme, netloc.lower(), path, params, query, fragment))
-
-
-def _unquotepath(path):
```
kmike: hmm, this function was here for a long time; I wonder if we should import it from w3lib.url to avoid breaking user code

kmike: other removed functions were here for much less time; I think it is fine to remove them as it is unlikely removal will break anything

ashkulz:
@kmike: sorry for the delay. Is there a necessity to keep `_safe_chars` and `_unquotepath`, as those are supposed to be internal interfaces for which [backwards compatibility may not apply](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)?

redapple:
@ashkulz , we sometimes break this guideline to be on the safe side with what users import.
We may have to live with this for some time.
Another thing to do (in another PR) could be to add a warning in the release notes that a later version may not expose `_safe_chars` and `_unquotepath`.

ashkulz:
@redapple: any other feedback? I've already pushed a rebased commit which includes this change, unless you want me to include an entry in `docs/news.rst` in this PR itself.

redapple:
@ashkulz , no other feedback. Looks good to me!
Thanks again.

kmike:
Thanks @ashkulz!

Issue #1241, opened 2015-05-18
Provide a way to propagate an exit code from a Spider


nramirezuy:
I think we need a way to propagate the exit code from the Spider.

Based on: #1231

kmike:
I like this idea. 

We can't do it from Spider because spider doesn't necessarily terminate the process - it is CrawlerProcess who does that.So we may need to expose CrawlerProcess to the spider. See also: https://github.com/scrapy/scrapy/issues/1226

nramirezuy:
Well what I mean was terminate the process and be able to add an exit code with the outcome. We already have an [exception](https://github.com/scrapy/scrapy/blob/master/scrapy/exceptions.py#L23) to terminate the process from the spider.

kmike:
This exception stops the spider, not necessarily terminates the process. You can have multiple spiders in the same process running at the same time.

debosmit:
In our case, an exception is being thrown by an external API call in the `start_requests` method. The spider is being invoked using a `subprocess.check_call` that uses the `scrapy crawl spname` command. The exception is not bubbling up to the crawler process that invokes it, resulting in a 0 exit code. Do you guys have any thoughts about how this issue? Can you point me to where I can look?

Issue #1231, opened 2015-05-14
Scrapy does not use a non-zero exit code when a scrape fails


iandees:
When invoking a Scrapy spider with e.g. `scrapy crawl spidername -o output.csv` and the spider fails for some reason (in our case, timeout to the HTTP server), the exit code is zero, giving subsequent steps in a shell script no way to check if the scrape completed successfully.

See my example here:
https://gist.github.com/iandees/74f51fefeb758b57f5e7

kmike:
For me the existing behaviour makes sense. Errors with remote resources are inevitable - you can't control remote servers. Say you crawled 100 pages and 1 of them failed - should it be reported in exit code? What if you're extracting links from a webpage, and some of them are broken - is it an error with your script? Even if all requests failed it may still be the expected result. 

What logic do you propose for setting a status code?

An allowed % of errors is task-specific; I think hardcoding some value ("if 10% requests failed return a non-zero status code") is a bad idea. To check if a crawl was successful or not (according to what you consider successful) one can check spider stats.

iandees:
The usecase here is that we're scraping webpages and expect the scrape to complete without error before moving on to the next step. I can see your point, and the existing behavior probably makes sense for the general case.

Having said that, can you recommend some way to catch any connection or scraping errors and gracefully exit with a non-zero exit code? I was thinking about listening for the [`spider_error` signal](http://doc.scrapy.org/en/latest/topics/signals.html#spider-error) and calling `sys.exit(1)`, but I imagine that leaves all sorts of stuff in an "unclean" state.

nramirezuy:
@iandees you can create a [custom command](http://doc.scrapy.org/en/master/topics/commands.html#custom-project-commands). There is an extension that already closes the spider by listening that signal ([doc](http://doc.scrapy.org/en/master/topics/extensions.html#scrapy.extensions.closespider.scrapy.extensions.closespider.CloseSpider), [source](https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/closespider.py#L29)).

```diff
diff --git a/scrapy/commands/crawl.py b/scrapy/commands/crawl.py
index b7fea7b..c8089a7 100644
--- a/scrapy/commands/crawl.py
+++ b/scrapy/commands/crawl.py
@@ -55,4 +55,7 @@ class Command(ScrapyCommand):
         spname = args[0]
 
         self.crawler_process.crawl(spname, **opts.spargs)
[...]

PR #2171, opened 2016-08-10
Checking if defer gets an error callback while in start_request and killing process

debosmit:
This is to account for the case that an exception is being raised in the start_requests method. In `scrapy/crawler.py:Crawler/crawl` the values yielded are left unused. `scrapy/commands/crawl.py:Command.run` does nothing with the deferred that is yielded. This fix checks if an error is raised and calls `sys.exit(1)` if it is. The stack-trace leading to the error is also logged.

This has been tracked in issue [1231](https://github.com/scrapy/scrapy/issues/1231) and [1241](https://github.com/scrapy/scrapy/issues/1241).

### DIFF
--- a/scrapy/commands/crawl.py
+++ b/scrapy/commands/crawl.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import sys
 from scrapy.commands import ScrapyCommand
 from scrapy.utils.conf import arglist_to_dict
 from scrapy.utils.python import without_none_values
@@ -36,7 +37,8 @@ class Command(ScrapyCommand):
         try:
             opts.spargs = arglist_to_dict(opts.spargs)
         except ValueError:
-            raise UsageError("Invalid -a value, use -a NAME=VALUE", print_help=False)
+            raise UsageError(
+                "Invalid -a value, use -a NAME=VALUE", print_help=False)
         if opts.output:
             if opts.output == '-':
                 self.settings.set('FEED_URI', 'stdout:', priority='cmdline')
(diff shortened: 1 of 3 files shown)


scrapy/commands/crawl.py
```diff
@@ -54,5 +62,14 @@ def run(self, args, opts):
             raise UsageError("running 'scrapy crawl' with more than one spider is no longer supported")
         spname = args[0]
 
-        self.crawler_process.crawl(spname, **opts.spargs)
+        d = self.crawler_process.crawl(spname, **opts.spargs)
+
+        def _crawl_error(_exc):
+            global exception
```
kmike: Is it a way to have a mutable variable in a closure (i.e. something what `nonlocal` keyword does in Python 3)? If so, I think it is better to use another workaround, e.g. using a list of length 1 and changing its first item.

scrapy/commands/crawl.py
```diff
@@ -1,9 +1,17 @@
...
+import logging
+import sys
 from scrapy.commands import ScrapyCommand
 from scrapy.utils.conf import arglist_to_dict
 from scrapy.utils.python import without_none_values
 from scrapy.exceptions import UsageError
 
+log = logging.getLogger(__name__)
+log.addHandler(logging.StreamHandler())
+log.setLevel(logging.DEBUG)
```
kmike: Hm, does it respect log level set in scrapy options? Is it necessary to add logger handler and set log level to get output?

kmike:
Rasing `sys.exit(1)` makes sense; I think the approach is correct. Thanks for the PR!

PR #2179, opened 2016-08-12
Update __init__.py

mayouf:
By using nested Item loader, we should be able to do it with a selector as an input. Not only a xpath or css.


Gallaecio:
Please, consider closing this PR and opening a similar one in https://github.com/scrapy/itemloaders instead.

PR #2180, opened 2016-08-14
DMOZ Site Update

andrewpost:
The DMOZ site updated it’s format, breaking the tutorial. This restores
the tutorial, and attempts to keep the //ul/li filtering scheme.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -297,16 +297,16 @@ or ``response.css()`` which map directly to ``response.selector.xpath()`` and
 So let's try it::
 
     In [1]: response.xpath('//title')
-    Out[1]: [<Selector xpath='//title' data=u'<title>Open Directory - Computers: Progr'>]
+    Out[1]: [<Selector xpath='//title' data=u'<title>DMOZ - Computers: Programming: La'>]
  
     In [2]: response.xpath('//title').extract()
-    Out[2]: [u'<title>Open Directory - Computers: Programming: Languages: Python: Books</title>']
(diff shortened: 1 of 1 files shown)


docs/intro/tutorial.rst
```diff
@@ -323,33 +323,33 @@ use Firefox Developer Tools or some Firefox extensions like Firebug. For more
...
+    response.xpath('//div[@class="title-and-desc"]/a/@href').extract()
 
 As we've said before, each ``.xpath()`` call returns a list of selectors, so we can
 concatenate further ``.xpath()`` calls to dig deeper into a node. We are going to use
 that property here, so::
 
-    for sel in response.xpath('//ul/li'):
-        title = sel.xpath('a/text()').extract()
+    for sel in response.xpath('//div[@class="title-and-desc"]'):
+        title = sel.xpath('//div[@class="site-title"]/text()').extract()
```
stummjr: Nice work, @andrewpost!

Just an observation: here we should be using relative XPath expressions such as:
```
title = sel.xpath('.//div[@class="site-title"]/text()').extract()
desc = sel.xpath('.//div[@class="site-descr "]/text()').extract()
```

Or even simpler ones:
```
title = sel.xpath('a/div/text()').extract()
desc = sel.xpath('div/text()').extract()
```

The same observation applies to subsequent occurrences of the same code.

kmike:
Thanks @andrewpost! We've rewritten the tutorial to use http://toscrape.com instead of dmoz (https://github.com/scrapy/scrapy/pull/2236). Sorry that it took so long!

PR #2189, opened 2016-08-19
Use "tls" extra for Twisted dependency (requires Twisted>=15.1)

redapple:
service_identity started being required from Twisted 14,
but was not listed as dependency until Twisted 15.1.
Requiring Twisted>=15.1 removes the need for service_identity dependency
at scrapy level.

Using a more modern Twisted version also solves a few TLS problems.

### DIFF
--- a/requirements-py3.txt
+++ b/requirements-py3.txt
@@ -1,7 +1,6 @@
-Twisted >= 15.5.0
+Twisted[tls]>=15.5.0
 lxml>=3.2.4
 pyOpenSSL>=0.13.1
 cssselect>=0.9
 queuelib>=1.1.1
 w3lib>=1.14.2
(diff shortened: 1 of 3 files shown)


kmike:
There was a discussion about twisted version bump here: https://github.com/scrapy/scrapy/pull/1887 - it seems we agreed on at least 12.1+. Probably jumping to 15+ requires another discussion.

redapple:
Definitely.

redapple:
Twisted 15.1 is a bit to big of a step forward.
It was decided to require [what Debian Jessie has, i.e. Twisted 14](https://packages.debian.org/fr/jessie/python-twisted)
and also recommend using virtualenvs for more up-to-date packages.

PR #2190, opened 2016-08-19
[MRG+1] Fix RANDOMIZE_DOWNLOAD_DELAY description in the docs

stummjr:
The documentation currently states:

> By default, Scrapy doesn’t wait a fixed amount of time between requests, but uses a random interval between 0.5 and 1.5 * DOWNLOAD_DELAY

It might mislead readers to think that, for a DOWNLOAD_DELAY=2, the interval would be from 0.5 to 3.0, when it's actually from 1.0 to 3.0.

See: https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/__init__.py#L38

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -504,8 +504,7 @@ supported.  Example::
 
 This setting is also affected by the :setting:`RANDOMIZE_DOWNLOAD_DELAY`
 setting (which is enabled by default). By default, Scrapy doesn't wait a fixed
-amount of time between requests, but uses a random interval between 0.5 and 1.5
-* :setting:`DOWNLOAD_DELAY`.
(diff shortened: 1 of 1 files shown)


stummjr:
In addition to that, it looks like the `BOT_NAME` description is incorrect:

> The name of the bot implemented by this Scrapy project (also known as the project name). This will be used to construct the User-Agent by default, and also for logging.

From what I tested and skimmed through the source code, it looks like it's not used to build the User-Agent. Am I right?

If so, I'll fix that too and update the PR.

kmike:
> In addition to that, it looks like the BOT_NAME description is incorrect

Hm, a good catch. What happens is that when a project is generated by startproject command, settings.py file is created, and it has the following lines:

```py
BOT_NAME = '$project_name'
# Crawl responsibly by identifying yourself (and your website) on the user-agent
#USER_AGENT = '$project_name (+http://www.yourdomain.com)'
```

redapple:
@stummjr , let's make it another PR for bot name.

PR #2193, opened 2016-08-23
MediaPipeline redirect fix #2004

### LINKED ISSUE #2004: MediaPipeline (and ImagesPipeline/FilesPipeline) does not handle HTTP redirections
Basically, what's happened is that my spider is unable to download the files because the file_urls provided are actually redirected to the final download link. However, because of the following code, the redirect download middleware is effectively disabled, which makes the download fail.

```
def _check_media_to_download(self, result, request, info):
        if result is not None:
            return result
        if self.download_func:
            # this ugly code was left only to support tests. TODO: remove
            dfd = mustbe_deferred(self.download_func, request, info.spider)
            dfd.addCallbacks(
                callback=self.media_downloaded, callbackArgs=(request, info),
                errback=self.media_failed, errbackArgs=(request, info))
        else:
            request.meta['handle_httpstatus_all'] = True
            dfd = self.crawler.engine.download(request, info.spider)
            dfd.addCallbacks(
                callback=self.media_downloaded, callbackArgs=(request, info),
                errback=self.media_failed, errbackArgs=(request, info))
        return dfd
```

And here is the check in the redirect middleware that disables it:

```
if (request.meta.get('dont_redirect', False) or
                response.status in getattr(spider, 'handle_httpstatus_list', []) or
                response.status in request.meta.get('handle_httpstatus_list', []) or
                request.meta.get('handle_httpstatus_all', False)):
            return response

```

My question is: What is the point of enabling the handling of httpstatus_all, when it effectively disables all of the checks?

Granitosaurus:
Fixes https://github.com/scrapy/scrapy/issues/2004

Right now this approach offers two settings:

`MEDIA_ALLOW_REDIRECTS` - Disables handling of all 300 codes (the ones Redirect Middleware uses)
`MEDIA_HTTPSTATUS_LIST` - Only handles codes in this setting

Also added tests and a doc entry(though this might need adjusting)

### DIFF
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.python.failure import Failure
 
+from scrapy.settings import Settings
 from scrapy.utils.defer import mustbe_deferred, defer_result
 from scrapy.utils.request import request_fingerprint
 from scrapy.utils.misc import arg_to_iter
@@ -28,6 +29,8 @@ class MediaPipeline(object):
 
     def __init__(self, download_func=None, settings=None):
         self.download_func = download_func
+        if isinstance(settings, dict) or settings is None:
+            settings = Settings(settings)
         resolve = functools.partial(self._key_for_pipe,
                                     base_class_name="MediaPipeline")
         self.allow_redirects = settings.getbool(
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.python.failure import Failure
 
+from scrapy.downloadermiddlewares.redirect import RedirectMiddleware
 from scrapy.settings import Settings
 from scrapy.utils.defer import mustbe_deferred, defer_result
 from scrapy.utils.request import request_fingerprint
@@ -36,7 +37,7 @@ class MediaPipeline(object):
         self.allow_redirects = settings.getbool(
             resolve('MEDIA_ALLOW_REDIRECTS'), self.ALLOW_REDIRECTS
         )
-        self.allow_httpstatus_list = settings.getlist(
+        self.handle_httpstatus_list = settings.getlist(
             resolve('MEDIA_HTTPSTATUS_LIST'), []
         )
 
@@ -104,17 +105,15 @@ class MediaPipeline(object):
 
     def _modify_media_request(self, request):
         httpstatus_list = []
-        if self.allow_httpstatus_list:
-            httpstatus_list = self.allow_httpstatus_list
-        elif self.allow_redirects:
+        if self.handle_httpstatus_list:
+            httpstatus_list = self.handle_httpstatus_list
+        if self.allow_redirects:
             if not httpstatus_list:
-                httpstatus_list = list(range(0, 300)) + list(range(400, 1000))
+                httpstatus_list = [i for i in range(1000)
+                                   if i not in RedirectMiddleware.allowed_status]
             else:
-                for i in range(300, 400):
-                    try:
-                        httpstatus_list.remove(i)
-                    except ValueError:
-                        pass
+                httpstatus_list = [i for i in httpstatus_list
+                                   if i not in RedirectMiddleware.allowed_status]
         if httpstatus_list:
             request.meta['handle_httpstatus_list'] = httpstatus_list
         else:
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+import functools
 import logging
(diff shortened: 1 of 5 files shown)


BjBlazkowicz:
I have tested this and it does not fix the issue. My files is still not downloaded when the file_urls is a url that yield a 301 response.

Granitas:
@BjBlazkowicz interesting, how did you test it? I've tested this myself on a pretty huge spider and all redirects (from http to https) worked fine.

BjBlazkowicz:
@Granitas This were not redirects from http to https but rather from https to https where the 301 response is pointing to the same url but with some some additional query-parameters:

Try making a request to the following url:

https://jyx.jyu.fi/dspace/bitstream/123456789/42895/1/jofobesityartid862514krachleretal.pdf

ZeeD26:
For some reason it does not load the setting from the setting file for me.
I subclassed `FilesPipeline` to modify the filepaths. Now when I use
```python
settings = Settings(
        {
            'ITEM_PIPELINES': {
                '__main__.MediaFilePipeline': 1,
            },
            'MEDIA_ALLOW_REDIRECTS': True
        })

    runner = CrawlerRunner(settings)
```
And log the setting for `self.allow_redirects`, it returns `False`.
If I hardcode it to `True` it works like a charm.

RabidCicada:
I've patched my local copy of scrapy with this as it is essential for the functioning of my spider and would love to see this merged in

ZeeD26:
@RabidCicada Thank you very much! Works like a charm now. I would also like to see this merged, as it is also essential for me.

scrapy/pipelines/media.py
```diff
@@ -90,6 +103,23 @@ def _process_request(self, request, info):
         )
         return dfd.addBoth(lambda _: wad)  # it must return wad at last
 
+    def _modify_media_request(self, request):
+        httpstatus_list = []
+        if self.handle_httpstatus_list:
+            httpstatus_list = self.handle_httpstatus_list
+        if self.allow_redirects:
+            if not httpstatus_list:
+                httpstatus_list = [i for i in range(1000)
```
redapple: could this be built at init?

scrapy/pipelines/media.py
```diff
@@ -90,6 +103,23 @@ def _process_request(self, request, info):
...
+                                   if i not in RedirectMiddleware.allowed_status]
+            else:
+                httpstatus_list = [i for i in httpstatus_list
```
redapple: I would find set operation easier to read. what do you think?

scrapy/pipelines/media.py
redapple: Maybe we can use a special object to test if an integer is not in some range, something along the lines of 
```
>>> class RangeExclusion(object):
...     def __init__(self, a, b):
...         self.lower_bound = a
...         self.upper_bound = b
...     def __contains__(self, item):
...         return not (self.lower_bound <= item <= self.upper_bound)
... 
>>> 200 in RangeExclusion(300, 400)
True
>>> 301 in RangeExclusion(300, 400)
False
>>> 404 in RangeExclusion(300, 400)
True
```
or using explicit status codes (closer to your suggestion)
```
>>> class ListExclusion(object):
...     def __init__(self, reject_list):
...         self.rejected =  reject_list
...     def __contains__(self, item):
...         return item not in self.rejected
... 
>>> 200 in ListExclusion((301, 302, 303, 307))
True
>>> 301 in ListExclusion((301, 302, 303, 307))
False
>>> 307 in ListExclusion((301, 302, 303, 307))
False
>>> 401 in ListExclusion((301, 302, 303, 307))
True
>>> 
```

scrapy/pipelines/media.py
```diff
@@ -16,6 +19,7 @@
 class MediaPipeline(object):
 
     LOG_FAILED_RESULTS = True
+    ALLOW_REDIRECTS = False
```
redapple: I would prefer that no new class attribute is introduced.
Can this be implemented with a default `MEDIA_ALLOW_REDIRECTS` setting in `default_settings.py`?

RabidCicada: I'm fairly new to detailed review/edits/commits on github.  I've done full pullrequests before.  IS there a way for me to pick up and make the changes you suggest and submit them under this pullrequest as an edit?  Or must the original contributor do it?

RabidCicada: Looks like answer is No...http://stackoverflow.com/questions/13592653/multi-user-github-pull-requests

@Granitosaurus would have to accept a merge from me then it would chain through his PR

redapple: @RabidCicada , you'll probably need to clone https://github.com/Granitosaurus/scrapy/tree/mediapipeline_redirect_fix_2004 in any case.
Then you can submit a pull request to than branch in @Granitosaurus 's repo,
or submit a new PR (with @Granitosaurus work + your commit) against scrapy master

redapple:
Work continued in https://github.com/scrapy/scrapy/pull/2616

PR #1591, opened 2015-11-11
Allow updating dictionary-like settings with magic prefix

jdemaeyer:
This PR would ease updating, not replacing, dictionary settings from the command line and from `Spider.custom_settings` by introducing a magix prefix:

```python
>>> mysettings = BaseSettings({'MYDICT': {'key': 'val'}})
>>> mysettings['MYDICT']
{'key': 'val'}
>>> mysettings.set('update:MYDICT', {'new_key': 'new_val'})
>>> mysettings['MYDICT']
{'new_key': 'new_val', 'key': 'val'}
```

PR #2195, opened 2016-08-23
Allow updating dictionary-settings

nyov:
This is another proposal for updating settings, without string prefixes. (See #1591 for the other)

It fails the mock tests
`BaseSettingsTest.test_setdict_alias` and
`BaseSettingsTest.test_setmodule_alias`
and probably has other issues at this time.

The idea is that merging two dictionaries (Settings namespace + an 'UPDATE' dict, that shadows the Settings namespace) is more convenient and efficient than string-parsing each setting key for a potential "update:" prefix.
It should also be more versatile (allows `setmodule` with an update flag).

The implementation can perhaps still be improved.
Having an update flag on an update method is also confusing. But both are different "updates".

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -100,6 +100,60 @@ class.
 The global defaults are located in the ``scrapy.settings.default_settings``
 module and documented in the :ref:`topics-settings-ref` section.
 
+.. _topics-settings-update-dicts:
+
+Updating dictionary settings
+============================
+
+Scrapy has a few dictionary-valued settings, e.g. for managing middlewares
+and extensions (:setting:`SPIDER_MIDDLEWARES`,
+:setting:`DOWNLOADER_MIDDLEWARES`, :setting:`EXTENSIONS`). Sometimes, it is
+desirable to *update*, not replace, these dictionaries. For these cases, you
+can fill the special ``update`` setting dictionary. This is particularly useful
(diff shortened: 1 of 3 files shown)


Gallaecio:
I like this approach more than https://github.com/scrapy/scrapy/pull/1591, however, I’m not sure if, for the documentation example, it is better than:

```python
from myproject.settings import SPIDER_MIDDLEWARES

    name = 'myspider'

    custom_settings = {
        'SPIDER_MIDDLEWARES': {
            **SPIDER_MIDDLEWARES,
            'myproject.spidermw.SpecialSMW': 300,
        },
    }
```

Gallaecio:
For the command-line use case, where we cannot import because it is made of simple strings, I wonder if we could not implement a command-line-specific template-like syntax, with a place holder that at run time can be replaced by the corresponding values.

For example:

```
-s 'SPIDER_MIDDLEWARES={{{SPIDER_MIDDLEWARES}}, "myproject.spidermw.SpecialSMW": 300}'
```

This would allow something that neither #1591 nor #2195 do: you could allow base settings to take priority over your custom settings. Although I admit that is a rather rare use case.

I’m just thinking as a user, through, I haven’t looked into how feasible it would be to implement something like this.

AdrianAtZyte:
https://github.com/scrapy/scrapy/pull/1591#issuecomment-5292028748

AdrianAtZyte:
#6614

Issue #2196, opened 2016-08-23
Response documentation is incorrect (does not have meta parameter)


thomdixon:
http://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response

The Response does not accept a meta parameter. I'll try to give a pull request if I get time, but just wanted to make sure this is logged.

eliasdorneles:
Fixed by https://github.com/scrapy/scrapy/pull/2197

Issue #713, opened 2014-05-07
Bring docstrings back


kmike:
What do you think about moving docstrings back to classes, functions and modules?

It'll make reading and understanding Scrapy code as well as working in Scrapy shell easier. Also, it will allow adding "viewcode" Sphinx extension to docs. Of course, there should be hand-written doc pages that explain main concepts, but the API reference can be mostly auto-generated.

dufferzafar:
How could one begin to go about fixing this?

nramirezuy:
@dufferzafar Create a PR, and make a commit per file. The title of this Issue is `Bring back docstrings`, so I guess some modules should have docstrings in his history.

kmike:
I've started it here: https://github.com/scrapy/scrapy/compare/autodocs, but didn't went too far :)

dufferzafar:
Okay, then I'll just extend @kmike's work.

eliasdorneles:
Hi @dufferzafar -- are you working on this?

dufferzafar:
@eliasdorneles No, I am not.

eliasdorneles:
Okay, I'll work on this.

eliasdorneles:
What do you think of adopting the napoleon extension? The docstrings look much more clear with it: https://pypi.python.org/pypi/sphinxcontrib-napoleon

kmike:
I used numpy docstring style in a couple of projects and liked it.

kmike:
Not to forget: http://doc.scrapy.org/en/1.0/contributing.html#documentation-policies should be updated once we'll get up to the speed with this ticket.

Gallaecio:
I’ve a branch where I’ve just:
* Refactored the Core API page into per-module pages of an "API reference" section.
* Updated the "Documentation policies" to require both updating docstrings and existing regular documentation, and recommend writing topics for undocumented or new features.
 
I guess the next step if creating additional pages for additional modules. But many of those modules will be currently undocumented, and adding all the missing documentation as part of this task may prove to by quite a task. Not to mention the potential for merge conflicts.

I wonder if it wouldn’t be better to just implement these changes for the time being, and extend the API reference section further later, e.g as part of other, more specific documentation tasks.

Gallaecio:
I’ll work on it further.

PR #2197, opened 2016-08-24
[MRG+1] Correct documentation about Response parameters

thomdixon:
This fixes issue #2196


kmike:
Nice catch, thanks @thomdixon!

eliasdorneles:
This is a good fix, thanks.

An even better fix will be to move these docs to docstrings in the code and use autodocs here, but that can be done in a follow-up.
I've been meaning to do that for a while since #713 and this gives me more motivation to get back on it -- so double thanks.

PR #2165, opened 2016-08-04
[MRG+1] Updated architecture graph for organization/clarity

loreguerra:
Found the first graph hard to read when first learning about Scrapy architecture.

Open to edits!!

PR #2202, opened 2016-08-26
[MRG+1] DOC move Data Flow below the picture; add links to components

kmike:
This is a small tweak for https://github.com/scrapy/scrapy/pull/2165: I found it hard to read "Data Flow" section because the picture was far from the description. So in this PR the explanation is moved right below the picture. After that it becomes harder to understand what do components mean, so I added links to them.

By the way, I think the overview doesn't explain correctly the difference between URLs and Requests. It says that spider returns URLs, and Engine converts them to Requests, but this is confusing - spider returns Requests.

### DIFF
--- a/docs/topics/architecture.rst
+++ b/docs/topics/architecture.rst
@@ -16,20 +16,71 @@ components and an outline of the data flow that takes place inside the system
 below with links for more detailed information about them. The data flow is
 also described below.
 
+.. _data-flow:
+
+Data flow
+=========
+
 .. image:: _images/scrapy_architecture_02.png
    :width: 700
    :height: 470
    :alt: Scrapy architecture
 
+The data flow in Scrapy is controlled by the execution engine, and goes like
+this:
+
+1. The :ref:`Engine <component-engine>` gets the first URLs to crawl from the
+   :ref:`Spider <component-spiders>`.
+
+2. The :ref:`Engine <component-engine>` schedules the URLs in the
+   :ref:`Scheduler <component-scheduler>` as Requests and asks for the
+   next URLs to crawl.
+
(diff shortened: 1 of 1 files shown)


redapple:
+1. I like the move of data flow section next to the picture.

The following is maybe worth another PR or discussion, but what about explaining middleware orders here too?

For downloader middleware,
> passing through the Downloader Middleware (request direction).

also means going through `.process_request()` of each download middleware, **in increasing downloader middleware order** (100, 200, 300...).

and 
> passing through the Downloader Middleware (response direction).

means going through `.process_response()` of each download middleware, **in decreasing downloader middleware order** (..., 300, 200, 100).

In fact, this is explained in [downloader middleware page](http://doc.scrapy.org/en/latest/topics/downloader-middleware.html#activating-a-downloader-middleware) but with slightly different words (engine side vs. downloader side):

> sorted list of enabled middlewares: the first middleware is the one closer to the engine and the last is the one closer to the downloader.

So even if people may get it from "response direction" and looking at the picture for engine and downloader sides, I think we can be explicit and consistent.

For completeness,

> passing through the Spider Middleware (input direction)

means through `.process_spider_input()` of each spider middleware, **in increasing spider middleware order** (100, 200, 300, ...).

> passing through the Spider Middleware (output direction).

meaning calling `.process_spider_output` **in decreasing spider middleware order** (..., 300, 200, 100).

I admit I had to write a spider with logs to not get it wrong.

kmike:
@redapple a good idea, +1

redapple:
@kmike , I opened https://github.com/scrapy/scrapy/issues/2262 to track the improvement discussed above.

PR #2215, opened 2016-09-01
Enable PyPy tests on Travis (allowed to fail)

redapple:


lopuhin:
@redapple the build has the same number of failures as my local build. The only issue is that it took 18 minutes, compared to 8 minutes for the second longest build. We could set fast-finishing https://docs.travis-ci.com/user/customizing-the-build#Fast-Finishing but it's only temporary, until we make that build required to pass.

redapple:
can we fast-finish only for allowed failures?

lopuhin:
Looks like no - initially I thought only allowed failures will be fast-finished, but the docs say: "Now, a build will finish as soon as a job has failed, or when the only jobs left allow failures."

redapple:
That's a shame. I'd prefer the regular builds all run

lopuhin:
> is that it took 18 minutes

At least 2 minutes are ``TimeoutError: <tests.test_engine.EngineTest testMethod=test_crawler> (test_crawler) still running at 120.0 secs``, and maybe there are more such cases, so it would take less time soon.

kmike:
Are performance issues solved?

redapple:
No idea @kmike . Probably not.
I don't think we'll know unless we give it a try

kmike:
I'm fine with merging it when it is rebased, but if pypy takes half an hour to run tests, and this delays green/red status for all pull requests it may be better to revert it.

lopuhin:
The timeout issues are fixed on another branch that is based on signals refactor, so can not be merged yet. But the tests on PyPy will still take more time than on CPython for this branch.
I'm not against merging it now, but maybe it makes sense to merge this PR once we actively start working on PyPy support, and maybe also find some way to make tests run faster?

lopuhin:
Re tests speed: I checked that PyPy is much more affected by coverage than CPython, so disabling it for PyPy should make the build faster (if it's not disabled already, but I believe it's not).

redapple:
Thanks for the input @lopuhin .
I rebased the PR, and configured tox to not run code coverage for pypy.
It's now taking just under 11 minutes: https://travis-ci.org/scrapy/scrapy/builds/197310281

lopuhin:
@redapple oh, this is great, so it's even not the slowest build now! Thanks for fixing that! 👍

kmike:
Thanks @redapple and @lopuhin! Let's enable these tests :)

PR #2217, opened 2016-09-02
[MRG+1] Add Segment Analytics to Documentation

stummjr:
This PR includes a [Segment.io](http://segment.io) analytics [tracking snippet](https://segment.com/docs/sources/website/analytics.js/quickstart/#step-1-copy-the-snippet) to the Scrapy documentation (http://doc.scrapy.org).

We're already doing it on Scrapy.org, this just adds the same thing to the documentation website, which will allow to see the unified analytics data.

The analytics data is oversee by Scrapinghub staff, which include Scrapy core developers.

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,0 +1,16 @@
+{% extends "!layout.html" %}
+
+{% block footer %}
+{{ super() }}
+<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


redapple:
This would need to be backporter to 1.1 branch (as it's the current "latest" on rtd)

PR #2218, opened 2016-09-03
[MRG+1] Update exceptions.rst

mremes:
There are namely no constructors in classes in Python but an ``__init__`` method instead.

### DIFF
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -60,7 +60,7 @@ remain disabled. Those components include:
  * Downloader middlewares
  * Spider middlewares
 
-The exception must be raised in the component constructor.
+The exception must be raised in the component's ``__init__`` method
 
 NotSupported
 ------------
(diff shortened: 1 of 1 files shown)


docs/topics/exceptions.rst
```diff
@@ -60,7 +60,7 @@ remain disabled. Those components include:
  * Downloader middlewares
  * Spider middlewares
 
-The exception must be raised in the component constructor.
+The exception must be raised in the component's ``__init__`` method
```
kmike: Could you please keep the dot at the end of the sentence?

kmike:
A good catch! Even Python docs sometimes call `__init__` a constructor (see e.g. https://docs.python.org/3/library/subprocess.html#subprocess.run), but it is better to be precise.

redapple:
Backport needed to master branch.

stav:
Couldn't we call `__new__` a constructor?  But yeah, I'm +1 for this merge. (two moths after the fact)

PR #2224, opened 2016-09-08
Duplicate log bug

gabsn:
### Dubplicate Log

There is an annoying bug when using python multiprocessing lib whith scrapy :

    class CompaniesCrawler(object):

        def __init__(self):
            self.crawler = CrawlerProcess(SETTINGS)

        def _crawl(self, spider):
            self.crawler.crawl(spider)
            self.crawler.start()
            self.crawler.stop()

        def crawl(self, spider):
            p = Process(
                target=self._crawl,
                args=[spider]
            )
            p.start()
            p.join()

**Issue :** Each time we instantiate [CrawlerProcess class] (https://github.com/scrapy/scrapy/blob/ebef6d7c6dd8922210db8a4a44f48fe27ee0cd16/scrapy/crawler.py), [configure_logging function](https://github.com/scrapy/scrapy/blob/ebef6d7c6dd8922210db8a4a44f48fe27ee0cd16/scrapy/utils/log.py) is called and a new handler is created : 

    handler = _get_handler(settings)
    logging.root.addHandler(handler)

So we eventually find ourselves with a lot of duplicate log handlers and the console becomes unreadable...

scrapy-duplicate-log


gabsn:
#2228 Duplicate log

PR #2225, opened 2016-09-08
[MRG+1] Two small fixes for when using the parse command and the '-r' flag (rules).

Tethik:
1. Use default "parse" as callback when the matching rule has no callback.
2. Log error and return when no rule matches the parsed urln (only if using -r).

The first error was encountered when running something similar to this:
`scrapy parse http://www.example.com/blah/ -r`
In a project with a CrawlSpider with two rules. The matching rule did not have a callback defined, resulting in the following error when the command tries to fetch the function.
```
ERROR: Spider error processing <GET http://www.example.com/blah/ (referer: None)
  File "/usr/lib/python3.5/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
  File "/usr/lib/python3.5/site-packages/Scrapy-1.2.0.dev2-py3.5.egg/scrapy/commands/parse.py", line 173, in callback
    cb_method = getattr(spider, cb, None)
TypeError: getattr(): attribute name must be string
```


kmike:
I'm not a `scrapy parse` user, but the changes look good, thanks!

redapple:
Thanks @Tethik !
Can I suggest that we add tests for the different cases?
My take on this: https://github.com/Tethik/scrapy/pull/1
(room for factorization)

Tethik:
@redapple Sounds good, I'll take a look at the your pull with the tests when I next get the time.

Tethik:
Tests look good to me. I couldn't think of a missing case either. One minor change is that I moved them to a separate file instead (test_commands.py -> test_command_parse.py). I assume codecov doesn't detect them because the tests are generating the spider code on the fly?

Anyhow, I don't have much more input on this. Feel free to merge if you think it looks ok.

redapple:
@kmike , do you (still) agree with the change?

kmike:
@redapple yep!
@Tethik hm, I'm not sure coverage problem is caused by generated code (but maybe it is); it looks like our testing suite has problems with subprocesses again.

kmike:
Thanks @Tethik!

PR #2226, opened 2016-09-09
Use functions from w3lib.url if they are present

lopuhin:
Move `from w3lib.url import *` to the end of the file, so that w3lib.url functions are defined later. The only imports left on top are imports that are actually used in this module.

Without this change, old functions defined in `scrapy.utils.url` are used even with new w3lib (for example, `canonicalize_url`), so changes bugfixes in w3lib.url are not picked up. And we can not yet change internal scrapy imports to use functions from w3lib, because we do not require newer w3lib (we can't require it yet, right?)

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -9,10 +9,7 @@ import posixpath
 import re
 from six.moves.urllib.parse import (ParseResult, urldefrag, urlparse)
 
-# scrapy.utils.url was moved to w3lib.url and import * ensures this
-# move doesn't break old code
-from w3lib.url import *
-from w3lib.url import _safe_chars, _unquotepath
+from w3lib.url import any_to_uri, add_or_replace_parameter
 from scrapy.utils.python import to_unicode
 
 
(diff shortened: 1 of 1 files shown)


kmike:
Hm, so is the idea to allow w2lib.url override functions from scrapy.utils.url in past Scrapy versions? E.g. if we move guess_scheme to w3lib.url in 1.2 release and then change it in w3lib, Scrapy 1.1 will get an updated version as well?

lopuhin:
@kmike I was thinking mostly about current scrapy versions: say we fix a bug in canonicalize_url in w3lib - at the moment we will not get the fix without fixing it in scrapy too, because a function from scrapy will be used.
Now that I wrote that, I realized that even after this change those who still use old w3lib will not get the fix. Do you think we'll want to update it in scrapy as well anyway?

kmike:
It sounds dangerous to me to override functions defined in scrapy from another library. It could be a hell to debug - e.g. you import canonicalize_url from scrapy.utils.url, then check scrapy.utils.url and see canonicalize_url source code and think this is what you're running, but in fact it can be either this, or a version from w3lib, depending on w3lib version (whether it has canonicalize_url implementation or not).

lopuhin:
Yeah, it's indeed more confusing than the other way around - I didn't think about it.

Issue #2231, opened 2016-09-13
Make logging configuration (more) customizable


redapple:
Would it make sense to have [`DEFAULT_LOGGING`](https://github.com/scrapy/scrapy/blob/ebef6d7c6dd8922210db8a4a44f48fe27ee0cd16/scrapy/utils/log.py#L45) be read from settings before going through [`dictConfig`](https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig)

Also, reading some of the issues with logging (e.g. https://github.com/scrapy/scrapy/issues/1977), it could even be useful to make `configure_logging` fully customizable too (for the brave), or at least having a way to add handlers.

eLRuLL:
I just saw this issue, and I think it is very related to what I wanted to do in #2494.

I would say adding a LogHandler would be a good solution. Right now scrapy only supports 3 types of Log Handler, so the idea here would be to override the handler directly with a new Log Handler class right?

sreecodeslayer:
I was also searching for options lately to add a handler (using this was my plan : [Log4Mongo](https://github.com/log4mongo/log4mongo-python/)) to default scrapy logger, so that I could get it to log to MongoDB collection. I then came across this issue and then went on to this > #2242

Could anyone help me find any workaround(if any)

Thank you :slightly_smiling_face:

umrashrf:
I am trying to append extra key/value to every log message generated by Scrapy. Those key/value are provided by user input at the time of running a spider. I need to configure my own logging so I can read custom user input fields from settings in a custom log handler.

umrashrf:
I would modify https://github.com/scrapy/scrapy/blob/master/scrapy/crawler.py#L245 to look for custom LOG_CONFIG class from settings.py and calling configure_logging otherwise.

majid-vaghari:
Hello,

I'm just starting to work with Scrapy. I wanted to reduce log level on production server but as I configured LOG_LEVEL to INFO, Scrapy still prints `"DEBUG Scraped from..."`.

I tried to trace the problem and I ended up here. How can I fix this problem? Is there any workaround?

I just think it doesn't make sense to set logging level to DEBUG when I explicitly set LOG_LEVEL to INFO. Am I doing something wrong? Is there anything wrong with the Scrapy itself? Can I make a PR to fix it?

Gallaecio:
[...]

PR #2242, opened 2016-09-15
Add LOG_HANDLER to settings, so that other file handlers can be chosen

taito-zz:
Title here is self explanatory, but wanted to add possibility to use other file handlers for logging.


redapple:
I like the idea of being able to customize logging more (cf. https://github.com/scrapy/scrapy/issues/2231).
A few comments:
- the setting name could be more explicit as `LOG_FILE_HANDLER`
- it could also have the current `logging.FileHandler` as default value
- users may want to customize the file log handlers even more, with settings translating to init arguments when instantiating the handler class, e.g. `logging.handlers.TimedRotatingFileHandler` has a lot of them: `filename, when='h', interval=1, backupCount=0, encoding=None, delay=False, utc=False, atTime=None`
- I'm not sure how common it is to customize `logging.StreamHandler` but there could be a `LOG_STREAM_HANDLER`

tests/test_utils_log_handler.py
```diff
@@ -0,0 +1,29 @@
...
+        name = 'logging.handlers.TimedRotatingFileHandler'
+        self.assertEqual(log._import_hander(name), TimedRotatingFileHandler)
+
+    def test_get_handler(self):
+        """Test function: _get_handler with LOG_FILE"""
+        # First create log file path
+        import os
+        import tempfile
+        file_path = tempfile.mkstemp()[1]
+        settings = {'LOG_FILE': file_path, 'LOG_ENCODIUNG': 'utf-8', 'LOG_LEVEL': 'DEBUG'}
```
redapple: typo: `'LOG_ENCODIUNG'` --> `'LOG_ENCODING'`

fangnahz: Noob question, does this mean scrapy will support TimedRotatingFileHandler soon? If true, when is this update coming?  And any pointers for what should I do before this update? ...

AdrianAtZyte:
You can now set `LOG_INSTALL_ROOT_HANDLER` to `False` and set up whatever you like.

PR #2243, opened 2016-09-15
[MRG+1] [image & file pipeline] loading setting for user classes

### LINKED ISSUE #2198: Custom ImagesPipeline doesn't generate thumbnails
For some reason, when using a customised ImagesPipeline thumbnails are no longer generated. This seems to be the case for 1.1.1 and 1.1.2 but not for 1.0.0 so somewhere between those releases something must have changed.

This [example](https://gist.github.com/briehanlombaard/b19885f99b9ee26363070fd9883695b9) should reproduce the problem.

pawelmhm:
if user has some custom subclass of Image pipeline and no setting for
this pipeline, he should get default settings defined for Image Pipeline.

Fixes #2198

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -233,7 +233,8 @@ class FilesPipeline(MediaPipeline):
         cls_name = "FilesPipeline"
         self.store = self._get_store(store_uri)
         resolve = functools.partial(self._key_for_pipe,
-                                    base_class_name=cls_name)
+                                    base_class_name=cls_name,
+                                    settings=settings)
         self.expires = settings.getint(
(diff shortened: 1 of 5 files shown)


redapple:
@briehanlombaard , would you be able to test the fix on your end too?

briehanlombaard:
@redapple That fixes the problem. Thanks!

redapple:
Thank you for checking @briehanlombaard (and reporting before that!)

redapple:
@kmike @eliasdorneles , do you agree with the change?

eliasdorneles:
Yup, looks good to me.
It's a bit tricky to follow the code (as many things related to the media pipeline), but the new behavior is certainly better.

PR #2248, opened 2016-09-15
[MRG+1] Make scrapy available in shell without explicit import statement

redapple:
Maybe it's just me but I'm always a bit annoyed to have to do `import scrapy` in the shell.
Thoughts?

### DIFF
--- a/scrapy/shell.py
+++ b/scrapy/shell.py
@@ -139,6 +139,7 @@ class Shell(object):
     def get_help(self):
         b = []
         b.append("Available Scrapy objects:")
+        b.append("  scrapy     scrapy module (contains scrapy.Request, scrapy.Selector, etc)")
         for k, v in sorted(self.vars.items()):
             if self._is_relevant(v):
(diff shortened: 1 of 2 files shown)


eliasdorneles:
Just tested here, MUCH better! :)

scrapy/shell.py
```diff
@@ -115,6 +115,9 @@ def fetch(self, request_or_url, spider=None):
         self.populate_vars(response, request, spider)
 
     def populate_vars(self, response=None, request=None, spider=None):
+        import scrapy
+
+        self.vars['scrapy'] = scrapy
```
kmike: I think it worths mentioning in get_help, something like "scrapy is already imported"

eliasdorneles: +1.

My suggestion:

```
        b.append("Available Scrapy objects:")
        b.append('  scrapy     scrapy module (contains scrapy.Request, scrapy.Selector, etc)')
        ...
```

redapple: +1

PR #2249, opened 2016-09-15
[MRG+1] docs: update overview spider code to use toscrape.com and minor changes

eliasdorneles:
So, this will replace the spider example code from the overview that
scrapes questions from StackOverflow by a spider scraping quotes (much
like the one in the tutorial), and upates the text around it to be
consistent.

There are also minor wording changes plus a small Sphinx/reST syntax fix
on the features list at the bottom (it was creating a definition list,
causing one line to be bold).

Does this look good?
Thanks!

### DIFF
--- a/docs/intro/overview.rst
+++ b/docs/intro/overview.rst
@@ -39,7 +39,7 @@ http://quotes.toscrape.com, following the pagination::
                 }
 
             next_page = response.css('li.next a::attr("href")').extract_first()
-            if next_page:
+            if next_page is not None:
                 next_page = response.urljoin(next_page)
                 yield scrapy.Request(next_page, callback=self.parse)
 
--- a/docs/intro/overview.rst
+++ b/docs/intro/overview.rst
@@ -78,9 +78,9 @@ Spider definition inside it and ran it through its crawler engine.
 The crawl started by making requests to the URLs defined in the ``start_urls``
 attribute (in this case, only the URL for quotes in *humor* category)
 and called the default callback method ``parse``, passing the response object as
-an argument. In the ``parse`` callback we loop through the quote elements
+an argument. In the ``parse`` callback, we loop through the quote elements
 using a CSS Selector, yield a Python dict with the extracted quote text and author,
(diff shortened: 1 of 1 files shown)


docs/intro/overview.rst
```diff
@@ -19,74 +19,69 @@ Walk-through of an example spider
...
+attribute (in this case, only the URL for quotes in *humor* category)
 and called the default callback method ``parse``, passing the response object as
-an argument. In the ``parse`` callback we extract the links to the
-question pages using a CSS Selector with a custom extension that allows to get
-the value for an attribute. Then we yield a few more requests to be sent,
-registering the method ``parse_question`` as the callback to be called for each
-of them as they finish.
+an argument. In the ``parse`` callback we loop through the quote elements
+using a CSS Selector, yield a Python dict with the extracted quote text and author,
+look for a link to the next page and schedules another request using the same
```
stummjr: `and schedules another request using the same`: `schedules` should be in 1st person of plural.

eliasdorneles: fixed!

docs/intro/overview.rst
```diff
@@ -19,74 +19,69 @@ Walk-through of an example spider
...
-        def parse_question(self, response):
-            yield {
-                'title': response.css('h1 a::text').extract_first(),
-                'votes': response.css('.question .vote-count-post::text').extract_first(),
-                'body': response.css('.question .post-text').extract_first(),
-                'tags': response.css('.question .post-tag::text').extract(),
-                'link': response.url,
-            }
+            next_page = response.css('li.next a::attr("href")').extract_first()
+            if next_page:
```
kmike: I think `next_page is None` could be slightly better because an empty string is a valid relative URL

eliasdorneles: fixed!

eliasdorneles:
hey @kmike, I'll merge this to move things forward, feel free to point out if you have any other concern.
thanks for reviewing!

PR #2252, opened 2016-09-15
[MRG+2] Tutorial: rewrite tutorial seeking to improve learning path

eliasdorneles:
This changes the tutorial, removing the step of creating an item class
and also starts by presenting the start_requests method instead of
start_urls.

~~This is currently WIP because we gotta decide when to present the short version of the spider (using start_urls and relying on the default callback).~~ This is now ready for review.

Main changes:

* remove unnecessary item class definition
* present start_requests first, to empower users from the beginning
* show CSS before XPath, for less scary learning
* get the learner playing in the shell as soon as possible
* show how to parameterize spiders using arguments
* show more illustrative code examples

We were also trying to let the learners think and try to guess things (instead of spoonfeed them), so we show code examples before we explain them.

What do you think?


docs/intro/tutorial.rst
```diff
@@ -93,20 +64,23 @@ They define an initial list of URLs to download, how to follow links, and how
 to parse the contents of pages to extract :ref:`items <topics-items>`.
 
 To create a Spider, you must subclass :class:`scrapy.Spider
-<scrapy.spiders.Spider>` and define some attributes:
+<scrapy.spiders.Spider>` and define some attributes and methods:
```
stummjr: What about presenting the source code first and then explaining the methods and attrs that the needs?

docs/intro/tutorial.rst
```diff
@@ -366,7 +349,11 @@ that property here, so::
     :ref:`topics-selectors-relative-xpaths` in the :ref:`topics-selectors`
     documentation
 
-Let's add this code to our spider::
+Knowing to use selectors, extracting data from a page is just a matter of
+yield the Python dictionaries from the callback method instead of printing
+them.
+
+Let's add the necessary code to our spider::
```
stummjr: Here we could add a note about simplifying the spider via `start_urls`.

docs/intro/tutorial.rst
```diff
@@ -159,25 +130,51 @@ similar to this::
...
+returned by the ``start_requests`` method of the Spider. Upon receiving
+a response for each one, it instantiates :class:`scrapy.http.Response`
+objects and calls the ``parse`` callback method passing the response as
+argument.
+
 
-Scrapy creates :class:`scrapy.Request <scrapy.http.Request>` objects
-for each URL in the ``start_urls`` attribute of the Spider, and assigns
-them the ``parse`` method of the spider as their callback function.
+Simplifying your spider
```
eliasdorneles: I'm not sure if using start_urls can be described as simplifying, because there are actually more things entangled behind the scenes (i.e., this second version is shorter, but not necessarily simpler).

I think it's best to present as a short version, saying something like: "This pattern of starting from a few hardcoded URLs is so common that Scrapy has a shortcut for it: you can define a start_urls class attribute w/ a list of URLs, which will be used by the default implementation of Spider.start_requests to create the initial requests"

Makes sense?

stummjr:
@eliasdorneles I did an attempt to complete the section on data extraction. Would you mind having a look?

docs/intro/tutorial.rst
```diff
@@ -490,13 +454,107 @@ Another common pattern is to build an item with data from more than one page,
...
 .. note::
-    As an example spider that leverages this mechanism, check out the
-    :class:`~scrapy.spiders.CrawlSpider` class for a generic spider
-    that implements a small rules engine that you can use to write your
+    As another example spider that leverages the mechanism of following links,
+    check out the :class:`~scrapy.spiders.CrawlSpider` class for a generic
+    spider that implements a small rules engine that you can use to write your
     crawlers on top of it.
 
+Adding a spider argument
```
redapple: What do you think of using "Passing" instead of "Adding"?

eliasdorneles: Hm, I dunno, I have the feeling passing is more about giving a value to something that is already parameterized. This is trying to show how to parameterize first, and then how to "pass" (give the value).
Does this make sense?

redapple: What about simply "Using [custom] spider arguments"?

eliasdorneles: Sounds good, let's go w/ that, less ambiguity. :+1:

eliasdorneles:
I think this is ready for review, I've updated the description with the main changes and some explanation of what @stummjr and I had in mind while writing. :)

For better reviewing, comment in this PR and read the text here: https://github.com/eliasdorneles/scrapy/blob/tutorial-upgrades/docs/intro/tutorial.rst

docs/intro/tutorial.rst
```diff
@@ -55,89 +55,64 @@ This will create a ``tutorial`` directory with the following contents::
...
+================
 
-They are declared by creating a :class:`scrapy.Item <scrapy.item.Item>` class and defining
-its attributes as :class:`scrapy.Field <scrapy.item.Field>` objects, much like in an ORM
-(don't worry if you're not familiar with ORMs, you will see that this is an
-easy task).
+Spiders are classes that you define and that Scrapy uses to scrape information
+from a website (or group of websites). They must subclass
+:class:`scrapy.Spider` and define the initial requests to make, how to follow
+links in the pages, and how to parse the downloaded page content to extract
```
redapple: "how to follow links in the pages" is in theory optional

docs/intro/tutorial.rst
```diff
@@ -159,214 +134,251 @@ similar to this::
...
+    ['change', 'deep-thoughts', 'thinking', 'world']
+
+Having figured out how to extract each bit, we can now iterate over all the
+quotes elements and put them together into a Python dictionary::
+
+    >>> for quote in response.css("div.quote"):
+    ...     text = quote.css("span.text ::text").extract_first()
+    ...     author = quote.css("small.author ::text").extract_first()
+    ...     tags = quote.css("div.tags a.tag ::text").extract()
+    ...     print(dict(text=text, author=author, tags=tags))
```
redapple: While reading it, I was expecting sample output here, first dict perhaps

docs/intro/tutorial.rst
```diff
@@ -379,78 +391,67 @@ Let's add this code to our spider::
...
 
+The `JSON Lines`_ format is useful because it's stream-like, you can easily
+append new records to it. As each record is a separate line, you can also
+process big files without having to fit everything in memory, there are tools
+like `JQ`_ to help doing that at the command-line.
 
-Now crawling quotes.toscrape.com yields ``QuoteItem`` objects::
+In small projects (like the one in this tutorial), that should be enough.
+However, if you want to perform more complex things with the scraped items, you
+can write an :ref:`Item Pipeline <topics-item-pipeline>`. As with Items, a
```
redapple: "As with Items" looks odd here, as it has not been introduced yet, only seen in `items.py` in the project file tree

docs/intro/tutorial.rst
```diff
@@ -7,16 +7,16 @@ Scrapy Tutorial
...
 This tutorial will walk you through these tasks:
 
 1. Creating a new Scrapy project
-2. Defining the Items you will extract
-3. Writing a :ref:`spider <topics-spiders>` to crawl a site and extract
-   :ref:`Items <topics-items>`
-4. Exporting the scraped data using command line
+2. Writing a :ref:`spider <topics-spiders>` to crawl a site and extract data
+3. Exporting the scraped data using the command line
+4. Change spider to recursively follow links
```
stummjr: I think it would be better to use the gerund to `change` too: `4. Change spider to recursively follow links`.

docs/intro/tutorial.rst
eliasdorneles: right, we can make that explicit: `define the initial requests to make, optionally how to follow links, and how to parse the content`

docs/intro/tutorial.rst
```diff
@@ -379,78 +391,67 @@ Let's add this code to our spider::
...
 
-Now that you know how to extract data from a page, why not extract the
-pagination links in each page, follow them and then extract the data you
-want for all of them?
+Now that you know how to extract data from pages, let's see how to follow links
+from them.
 
-Here is a modification to our spider that does just that::
+Here is a modification of our spider that recursively follows the link to the next
+page, extracting data from it::
```
redapple: I think here you could have the HTML snippet for pagination elements, otherwise it may confuse readers about how `response.css('li.next a::attr("href")')` was chosen.
Also, this instroduces `::attr(href)`, another handy yet non-standard use of CSS selectors. Worth a note I think.

eliasdorneles: oh yeah, +1 to both comments! thanks, Paul

docs/intro/tutorial.rst
```diff
@@ -159,214 +134,251 @@ similar to this::
...
-  attribute ``class="mine"``. Equivalent CSS selector: ``div.mine``.
+The best way to learn how to extract data with Scrapy is trying selectors
+using the shell :ref:`Scrapy shell <topics-shell>`. Run::
 
-These are just a couple of simple examples of what you can do with XPath, but
-XPath expressions are indeed much more powerful. To learn more about XPath, we
-recommend `this tutorial to learn XPath through examples
-<http://zvon.org/comp/r/tut-XPath_1.html>`_, and `this tutorial to learn "how
-to think in XPath" <http://plasmasturm.org/log/xpath101/>`_.
+    scrapy crawl http://quotes.toscrape.com/page/1/
```
stummjr: scrapy shell http://quotes.toscrape.com/page/1/

eliasdorneles:
@redapple @stummjr I've address your comments, plus some other minor things. What do you think?

stummjr:
@eliasdorneles LGTM 👍

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-<topics-loaders>`, a mechanism with helpers to conveniently populate `Items`.
+Our first Spider
+================
 
-They are declared by creating a :class:`scrapy.Item <scrapy.item.Item>` class and defining
-its attributes as :class:`scrapy.Field <scrapy.item.Field>` objects, much like in an ORM
-(don't worry if you're not familiar with ORMs, you will see that this is an
-easy task).
+Spiders are classes that you define and that Scrapy uses to scrape information
+from a website (or group of websites). They must subclass
```
kmike: or a group

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-  Spider will begin to crawl from. The first pages downloaded will be those
-  listed here. The subsequent URLs will be generated successively from data
-  contained in the start URLs.
-
-* :meth:`~scrapy.spiders.Spider.parse`: a method of the spider, which will
-  be called with the downloaded :class:`~scrapy.http.Response` object of each
-  start URL. The response is passed to the method as the first and only
-  argument.
+* :meth:`~scrapy.spiders.Spider.start_requests`: must return a list
+  of requests where the Spider will begin to crawl from.
```
kmike: Spider doesn't return a list in start_requests example.

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-            filename = 'quotes-' + response.url.split("/")[-2] + '.html'
-            with open(filename, 'wb') as f:
-                f.write(response.body)
-
-Crawling
---------
-
-To put our spider to work, go to the project's top level directory and run::
+To put our spider to work, go to the project's top level directory (``cd
+tutorial``) and run::
```
kmike: There are two tutorial folders, which one should user go to? Previously it was said: 
> Save it in a file named ``quotes_spider.py`` under the ``tutorial/spiders`` directory in your project

I think it assumes user is already in project folder, so now user can mistakenly cd to tutorial/tutorial

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
 What just happened under the hood?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Scrapy creates :class:`scrapy.Request <scrapy.http.Request>` objects
-for each URL in the ``start_urls`` attribute of the Spider, and assigns
-them the ``parse`` method of the spider as their callback function.
+Scrapy schedules the :class:`scrapy.Request <scrapy.http.Request>` objects
+returned by the ``start_requests`` method of the Spider. Upon receiving
+a response for each one, it instantiates :class:`scrapy.http.Response`
+objects and calls the ``parse`` callback method passing the response as
```
kmike: It could read as if Scrapy always calls `parse` callback, not only in our example, because it is written in present simple ('schedules', 'instantiates'), and for `start_requests` it is indeed what always happens.

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-mechanisms see the :ref:`Selectors documentation <topics-selectors>`.
 
-.. _XPath: https://www.w3.org/TR/xpath
-.. _CSS: https://www.w3.org/TR/selectors
+    class QuotesSpider(scrapy.Spider):
+        name = "quotes"
+        start_urls = [
+          'http://quotes.toscrape.com/page/1/',
+          'http://quotes.toscrape.com/page/2/',
+        ]
```
kmike: it looks like urls are indented using 2 spaces instead of 4

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-  attribute ``class="mine"``. Equivalent CSS selector: ``div.mine``.
+The best way to learn how to extract data with Scrapy is trying selectors
+using the shell :ref:`Scrapy shell <topics-shell>`. Run::
 
-These are just a couple of simple examples of what you can do with XPath, but
-XPath expressions are indeed much more powerful. To learn more about XPath, we
-recommend `this tutorial to learn XPath through examples
-<http://zvon.org/comp/r/tut-XPath_1.html>`_, and `this tutorial to learn "how
-to think in XPath" <http://plasmasturm.org/log/xpath101/>`_.
+    scrapy shell http://quotes.toscrape.com/page/1/
```
kmike: let's write it as `scrapy shell 'http://quotes.toscrape.com/page/1/'`; "&" handling in console is an endless source of confusion.

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-response.
+    [ ... Scrapy log here ... ]
+    2016-09-19 12:09:27 [scrapy] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/1/> (referer: None)
+    [s] Available Scrapy objects:
+    [s]   crawler    <scrapy.crawler.Crawler object at 0x7fa91d888c90>
+    [s]   item       {}
+    [s]   request    <GET http://quotes.toscrape.com/page/1/>
+    [s]   response   <200 http://quotes.toscrape.com/page/1/>
+    [s]   settings   <scrapy.settings.Settings object at 0x7fa91d888c10>
+    [s]   spider     <DefaultSpider 'default' at 0x7fa91c8af990>
```
kmike: in Scrapy 1.2 "scrapy" will also be here

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-* :meth:`~scrapy.selector.Selector.extract`: returns a unicode string with the
-  selected data.
+    >>> response.css('title::text').extract()
+    [u'Quotes to Scrape']
 
-* :meth:`~scrapy.selector.Selector.re`: returns a list of unicode strings
-  extracted by applying the regular expression given as argument.
+There are two things to note here: one is that we've added ``::text`` to the
+CSS query, to mean that we want to select the text from inside the title element.
+If we don't specify ``::text``, we'd get the HTML tags::
```
kmike: If I'd be reading this I'd thought that "::text" is a way to get text from a css selector, but e.g. for `"<p>hello <b>world</b></p>"` selector `p::text` will only return `['hello ']`.

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
+:meth:`~scrapy.selector.SelectorList.extract_first` methods, you can also use
+the :meth:`~scrapy.selector.Selector.re` method to extract using a regular
+expression::
 
-    [ ... Scrapy log here ... ]
+    >>> response.css('title::text').re('Quotes.*')
+    [u'Quotes to Scrape']
+    >>> response.css('title::text').re('Q\w+')
+    [u'Quotes']
+    >>> response.css('title::text').re('(\w+) to (\w+)')
```
kmike: I think it is a good practice to use raw strings for regexes (`r'(\w+) to (\w+)'`) because it allows to avoid gotchas with some regexes, e.g. with `\b`.

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-   will not work.
+There's a lesson here: for most scraping code, you want it to be resilient to
+errors due to things not being found on a page, so that even if some parts fail
+to be scraped, you can at least get **some** data.
 
-This is what the shell looks like::
+Besides the :meth:`~scrapy.selector.Selector.extract` and
+:meth:`~scrapy.selector.SelectorList.extract_first` methods, you can also use
+the :meth:`~scrapy.selector.Selector.re` method to extract using a regular
+expression::
```
kmike: What do you think about adding a link to python stdlib docs for re module, to point to a place where users can get information about re syntax?

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
-    [s]   spider     <DefaultSpider 'default' at 0x1092808d0>
-    [s] Useful shortcuts:
-    [s]   shelp()           Shell help (print this help)
-    [s]   fetch(req_or_url) Fetch request (or URL) and update local objects
-    [s]   view(response)    View response in a browser
-    
-    >>>
+In order to find the proper CSS selectors to use, you might find useful opening
+the response page from the shell in your web browser using ``view(response)``.
+You can use your browser developer tools or extensions like Firebug. For more
```
kmike: What do you think about mentioning SelectorGadget extension?

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
+that contains the text "Next Page"**. This makes XPath very fitting to the task
+of scraping, and we encourage you to learn XPath even if you already know how to
+construct CSS selectors, it will make scraping much easier.
 
-Extracting the data
-^^^^^^^^^^^^^^^^^^^
+We won't cover much of XPath here. To learn more about XPath, we recommend
+`this tutorial to learn XPath through examples
+<http://zvon.org/comp/r/tut-XPath_1.html>`_, and `this tutorial to learn "how
+to think in XPath" <http://plasmasturm.org/log/xpath101/>`_.
```
kmike: Scrapy docs also have a section about selectors which is mostly focused on xpaths, what do you think about adding a link?

docs/intro/tutorial.rst
```diff
@@ -55,318 +55,333 @@ This will create a ``tutorial`` directory with the following contents::
...
+Having figured out how to extract each bit, we can now iterate over all the
+quotes elements and put them together into a Python dictionary::
+
+    >>> for quote in response.css("div.quote"):
+    ...     text = quote.css("span.text::text").extract_first()
+    ...     author = quote.css("small.author::text").extract_first()
+    ...     tags = quote.css("div.tags a.tag::text").extract()
+    ...     print(dict(text=text, author=author, tags=tags))
+    {'text': u'\u201cThe world as we have created it is a process of our thinking. It cannot be changed without changing our thinking.\u201d', 'tags': [u'change', u'deep-thoughts', u'thinking', u'world'], 'author': u'Albert Einstein'}
+    {'text': u'\u201cIt is our choices, Harry, that show what we truly are, far more than our abilities.\u201d', 'tags': [u'abilities', u'choices'], 'author': u'J.K. Rowling'}
```
kmike: In Python 3 users won't get those ugly `\u201cIt`. What do you think about using Python 3 in the tutorial, and mentioning that in Python 2 output can be different? I'm not sure about that.

docs/intro/tutorial.rst
```diff
@@ -379,78 +394,91 @@ Let's add this code to our spider::
...
+                }
 
-Note how we've changed to use the method ``.extract_first()``, which extracts
-the first element from a selector list returned by ``.xpath()``.
+If you run this spider, it will output the extracted data with the log::
 
-Now try crawling quotes.toscrape.com again and you'll see sites being printed
-in your output. Run::
+    2016-09-19 18:57:19 [scrapy] DEBUG: Scraped from <200 http://quotes.toscrape.com/page/1/>
+    {'tags': ['life', 'love'], 'author': 'André Gide', 'text': '“It is better to be hated for what you are than to be loved for what you are not.”'}
```
kmike: hm, it seems here Python 3 is used, while earlier output was from Python 2

docs/intro/tutorial.rst
```diff
@@ -379,78 +394,91 @@ Let's add this code to our spider::
...
 
-Now that you know how to extract data from a page, why not extract the
-pagination links in each page, follow them and then extract the data you
-want for all of them?
+This gets the anchor element, but we want the attribute ``href``. For that,
+Scrapy supports a CSS extension that let's you select the attribute contents,
+like this::
 
-Here is a modification to our spider that does just that::
+    >>> response.css('li.next a::attr("href")').extract_first()
```
kmike: are quotes around href needed?

docs/intro/tutorial.rst
```diff
@@ -490,30 +523,109 @@ Another common pattern is to build an item with data from more than one page,
...
+Using spider arguments
+======================
 
-The simplest way to store the scraped data is by using :ref:`Feed exports
-<topics-feed-exports>`, with the following command::
+You can provide command line arguments to your spiders by using the ``-a``
+option when running them::
 
-    scrapy crawl quotes -o items.json
+    scrapy crawl quotes -o items.json -a tag=humor
```
kmike: If users will follow the tutorial as-is, items.json file will contain multiple json objects because by default data is appended. I think we should mention this gotcha somwhere.

docs/intro/tutorial.rst
```diff
@@ -490,30 +523,109 @@ Another common pattern is to build an item with data from more than one page,
...
+
+    class QuotesSpider(scrapy.Spider):
+        name = "quotes"
+
+        def start_requests(self):
+            url = 'http://quotes.toscrape.com/'
+            tag = getattr(self, 'tag', None)
+            if tag is not None:
+                url = url + 'tag/' + tag
+            yield scrapy.Request(url)
```
kmike: my vote is to always pass a callback, even if it is self.parse: `yield scrapy.Request(url, self.parse)`.

docs/intro/tutorial.rst
eliasdorneles: Right! Would `must generate or return a list of requests` make sense?

docs/intro/tutorial.rst
eliasdorneles: Right, I'll just remove this `cd tutorial`.

docs/intro/tutorial.rst
eliasdorneles: oh right, lemme bring back the note about making sure to use that.

docs/intro/tutorial.rst
eliasdorneles: Right, would this be enough:

> ... we've added ``::text`` to the CSS query, meaning we want to select the text elements inside the ``<title>``.

Or do you feel we should add a note explaining about the behavior you're mentioning?

docs/intro/tutorial.rst
eliasdorneles: good idea, will do!

docs/intro/tutorial.rst
eliasdorneles: indeed, best to use Python 3 for everything.

docs/intro/tutorial.rst
eliasdorneles: AFAIK, both are accepted: https://github.com/scrapy/parsel/blob/master/parsel/csstranslator.py#L81

Quotes aren't needed, I just have the habit of putting them -- I'll remove from the examples.

docs/intro/tutorial.rst
eliasdorneles: ohh, that's kinda big deal -- yea, we'll have to write something.
For the tutorial, I'll just use different files `items-humor.json`.
didn't we have a PR adding support for `-O`?

docs/intro/tutorial.rst
eliasdorneles: Rewrote to:

```
There are two things to note here: one is that we've added ``::text`` to the
CSS query, to mean we want to select only the text elements directly inside
``<title>`` element.  If we don't specify ``::text``, we'd get the full title
element, including tags::

    >>> response.css('title').extract()
    ['<title>Quotes to Scrape</title>']
```

eliasdorneles:
@kmike I think I've addressed all your comments as well. :)

Btw, thanks for reviewing you all! :bow: :bow:

eliasdorneles:
After chatting about it w/ @kmike, I changed the recommendation on how to learn Python.

For people experienced w/ other languages, the tutorial recommends Dive into Python 3 and the official Python tutorial as alternative.

For people learning to program, it recommends Learn Python the Hard Way and alternatively the list of resources for non-programmers from Python's wiki.

docs/intro/tutorial.rst
kmike: "generate a list of requests" doesn't sound correct to me, I think it is better to write something like "must return an iterable of Requests - you can e.g. return a list of requests or write a generator function"

docs/intro/tutorial.rst
```diff
@@ -55,87 +63,63 @@ This will create a ``tutorial`` directory with the following contents::
...
 
-    class QuotesSpider(scrapy.Spider):
-        name = "quotes"
-        start_urls = [
-            'http://quotes.toscrape.com/page/1/',
-            'http://quotes.toscrape.com/page/2/',
-        ]
+* :meth:`~scrapy.spiders.Spider.parse`: a method that will be called to handle
+  the response downloaded for each of the requests made. The response parameter
+  is an instance of :class:`~scrapy.http.Response` that holds the page content and
```
kmike: We discussed it before, but response parameter can be an instance of a Response subclass, not only a Response instance. Linking to Response docs can be confusing because Response class doesn't have methods we're using later (.css, .xpath). Even after https://github.com/scrapy/scrapy/pull/2275 this issue is not fixed because Response docs still don't mention new methods, and technically it is not correct that response is an instance of Response.

docs/intro/tutorial.rst
```diff
@@ -145,228 +129,282 @@ This command runs the spider with name ``quotes`` that we've just added, that
...
+            <a class="tag" href="/tag/world/page/1/">world</a>
         </div>
     </div>
 
+Let's open up scrapy shell and play a bit to find out how to extract the data
+we want::
 
-So we can select each ``<div class="quote">`` element belonging to the site's 
-list with this code::
+    $ scrapy shell http://quotes.toscrape.com
```
kmike: url should be in quotes

docs/intro/tutorial.rst
```diff
@@ -460,19 +516,24 @@ Here is a modification to our spider that does just that::
...
-                item['text'] = quote.xpath('span[@class="text"]/text()').extract_first()
-                item['author'] = quote.xpath('span/small/text()').extract_first()
-                yield item
-            next_page = response.xpath('//li[@class="next"]/a/@href').extract_first()
-            if next_page:
+            for quote in response.css('div.quote'):
+                yield {
+                    'text': quote.css('span.text::text').extract_first(),
+                    'author': quote.css('span small::text').extract_first(),
+                    'tags': quote.css("div.tags a.tag::text").extract(),
```
kmike: A very minor nitpick: probably it is better to use single quotes here, for consistency with other selectors.

docs/intro/tutorial.rst
```diff
@@ -460,19 +516,24 @@ Here is a modification to our spider that does just that::
...
+            if next_page is not None:
                 next_page = response.urljoin(next_page)
                 yield scrapy.Request(next_page, callback=self.parse)
 
-Now after extracting an item the `parse()` method looks for the link to the next page, 
-builds a full absolute URL using the `response.urljoin` method (since the links can
-be relative) and yields a new request to the next page, registering itself as callback to handle the data extraction for the next page and to keep the crawling going through all the pages.
+
+Now, after extracting the data, the ``parse()`` method looks for the link to
+the next page, builds a full absolute URL using the ``response.urljoin`` method
```
kmike: what do you think about adding a link to response.urljoin docs?

kmike: though it is not that useful because of Response/TextResponse differences.. response.urljoin is better because it handles `<base>` html tag, but this is not explained in Response docs because it is a TextResponse feature

docs/intro/tutorial.rst
```diff
@@ -490,30 +551,109 @@ Another common pattern is to build an item with data from more than one page,
 using a :ref:`trick to pass additional data to the callbacks
 <topics-request-response-ref-request-callback-arguments>`.
 
+Another example: scraping authors
```
kmike: What do you think about moving the section above (a link to the trick to pass additional data to callbacks) below this example?

kmike:
Besides a few minor remaining comments the PR looks good, great work @eliasdorneles and @stummjr!

eliasdorneles:
Thanks for reviewing @kmike, I've addressed your comments and made other small changes in the latest commit.
@redapple @kmike Does this look good to merge?

redapple:
@eliasdorneles @stummjr , I believe the tutorial now is also compatible with 1.1 branch, correct?
Worth a backport maybe?

eliasdorneles:
It is, but it's going to be some work -- I tried cherry-picking the first commit and got a bunch of conflicts.
I'll do it w/ more patience later today. :+1:

redapple:
@eliasdorneles , could you check https://github.com/scrapy/scrapy/pull/2281 ?
I basically overwrote tutorial.rst with the content of this PR

Issue #872, opened 2014-09-02
FileFeedStorage creates empty file when no items are scraped


gbirke:
When no items are scraped, the corresponding file is created none the less, because it is created in by the `storage.open` call in `FeedExporter.open_spider`. This behavior ignores the the setting of `FEED_STORE_EMPTY` when using file export.

My proposal for this would be to add a `cleanup` method to the `IFeedStorage` interface. Then  `FeedExporter.close_spider` can call that method before returning in case `slot.itemcount` is zero and `self.store_empty` is `False`. `cleanup` could also be called internally from the `store` methods of the `IFeedStorage` implementations.

namelessGonbai:
The bug has not been resolved yet. I don't believe the resolution proposed in the PR #2258 was sufficient. Although it may fix the issue of broken JSON, the problem of FileFeedStorage creating empty files remains.
This is because FileFeedStorage.open open the actual file within it, which generates the file even if there is no writing happening.
In my opinion, one of the following is necessary:
1. Only open the file when the first item is received, similar to how start_exporting works.
2. Make FileFeedStorage use a TemporaryFile like BlockingFeedStorage does.

namelessGonbai:
So I hope this issue will be reopened.

namelessGonbai:
https://github.com/scrapy/scrapy/pull/916#issuecomment-247939197
> The problem with removing file is that it doesn't play nicely with the default behavior of appending to the result.

While removing the file may cause problems with appending, I think delaying the direct opening of the file would  be able to work well with the appending behavior and prevent the creation of an empty file.

namelessGonbai:
Since there doesn't seem to be much demand for the patch, I am thinking of making my own patch. However, I am not sure which of the two solutions suggested earlier is better. 
Creating a temporary file and writing to it seems like unnecessary overhead. On the other hand, opening the file for the first time at the moment of writing and trying to write to it would require a major change in logic and look bad.

Gallaecio:
My gut says postponing file creation until there is an item to write is the way to go, however ugly it might be. But my opinion may change depending on how ugly things can get :sweat_smile:

namelessGonbai:
Let's see how the return value of IFeedStorage.open() is used.
https://github.com/scrapy/scrapy/blob/8fbebfa943c3352f5ba49f46531a6ccdd0b52b60/scrapy/extensions/feedexport.py#L426-L450
Here is the first problem.
[...]

PR #2258, opened 2016-09-16
[MRG+1] Feed exporter: start exporting only on first item

redapple:
Fixes GH-872

This prevents the `[` files when spider outputs nothing.

* with `FEED_STORE_EMPTY` as `False` (the default), this changes produces empty files
* with `FEED_STORE_EMPTY` as `True`, for JSON output, you get an empty list, and for XML, you get a empty top `<items>` element.


kmike:
Thanks @redapple!

PR #2268, opened 2016-09-20
Make crawler.engine available in scrapy's Spider.__init__ method.

ahlinc:
Hi,
I've had situation when I needed to extend `Spider.__init__` method and access to low-level http-cache middleware instance methods to manually control scrapy's work with the cache. But in current implementation the engine will be created after spider instance and crawler property will be assigned after spider instance creation, so the next possible point where engine may be accessed is only _start_requests_ method but in my opinion it's irrational place. So in this PR I suggest to change the order of components creation and as result the engine will be available at the end of spider instance `__init__` method just by `self.crawler.engine` reference.

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -68,8 +68,8 @@ class Crawler(object):
         self.crawling = True
 
         try:
-            self.spider = self._create_spider(*args, **kwargs)
             self.engine = self._create_engine()
+            self.spider = self._create_spider(*args, **kwargs)
             start_requests = iter(self.spider.start_requests())
             yield self.engine.open_spider(self.spider, start_requests)
             yield defer.maybeDeferred(self.engine.start)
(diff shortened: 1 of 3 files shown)


redapple:
Thanks @ahlinc . Sorry for the late feedback.
Looks suspiciously harmless "à-priori" but as this touches the heart of the orchestration, I would ask @curita if she can comment, being the one who last touched (and refactored) this part.
@jdemaeyer may also want to check how this affects the Add-Ons project (if it does).

@curita , @jdemaeyer , would you have time to check this? Thanks

ahlinc:
Hey @redapple, could you give some attention for this PR, please?
Because even after 3,5 years it's still actual ;)

redapple:
@ahlinc , I'm afraid I'm not involved in scrapy for about 2 years now.
I suggest you ping @Gallaecio and @kmike

Gallaecio:
Have you considered using the [engine_started](https://docs.scrapy.org/en/latest/topics/signals.html#engine-started) or [spider_opened](https://docs.scrapy.org/en/latest/topics/signals.html#spider-opened) signals for your use case?

PR #1731, opened 2016-01-27
[MRG+2] LOG_SHORT_NAMES option to disable TopLevelFormatter

kmike:
Fixes https://github.com/scrapy/scrapy/issues/1309

This is a rebased version of https://github.com/scrapy/scrapy/pull/1583. 

Like in @akhillb's PR, default value is False, i.e. logs are switched to long names by default. Is that OK?

TODO:
- [x] tests
- [x] confirm that's OK to use long names by default
- [x] update output of examples in docs
- [x] add example of overriding log level for a specific Scrapy component

Issue #1309, opened 2015-06-15
provide an option to disable scrapy.utils.log.TopLevelFormatter


kmike:
Long names in logs (e.g. `[scrapy.core.scraper]` instead of generic `[scrapy]`) are useful because
- they show where the log message came from - this makes debugging easier;
- it has an educational value - users become aware of various Scrapy components;
- to filter out a specific kind of messages (or make them more verbose) user can set a custom log level just for one logger, but to do that user needs to know logger full name.

My preference is to even use full names by default, but an option to enable them is also fine.
Thoughts?

curita:
+1 to use full names by default, not sure if there is any value in keeping just the top level logger besides replicating the old behaviour. I'm ok with deleting TopLevelFormatter entirely.

nramirezuy:
I like `TopLevelFormatter`, long names can be too long. Besides after some time you know from where the log lines come from.

PR #2270, opened 2016-09-20
[MRG+1] Raise log level for HttpErrorMiddleware to INFO (from DEBUG)

### LINKED ISSUE #910: HttpErrorMiddleware log level should be at least INFO, not DEBUG
I want to spot immediately if one of my spiders get 403 or 404 response codes. But the current implementation of HttpErrorMiddleware, sets log as DEBUG... debug?! It really doesn't matter if a spider is not able to scrape a page? It should matter! It should be a very important log to read!

redapple:
Fixes #910

May mean log spam for some users.

### DIFF
--- a/scrapy/spidermiddlewares/httperror.py
+++ b/scrapy/spidermiddlewares/httperror.py
@@ -46,7 +46,7 @@ class HttpErrorMiddleware(object):
 
     def process_spider_exception(self, response, exception, spider):
         if isinstance(exception, HttpError):
-            logger.debug(
+            logger.info(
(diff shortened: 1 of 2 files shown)


kmike:
I think it'd be nice for this PR to go together with https://github.com/scrapy/scrapy/issues/1309 and a doc section which explains how to change verbosity of logs for a single scrapy component.

kmike:
I've made some progress on https://github.com/scrapy/scrapy/pull/1731 - do you think we can sneak it in 1.3?

redapple:
@kmike , I think so yes

Issue #2244, opened 2016-09-15
Add note on MailSender that "to" and "cc" must be list of recipients


redapple:
It can be easy to miss in the docs (see https://github.com/scrapy/scrapy/issues/2148)

PR #2271, opened 2016-09-20
[MRG+1] Add note on "to" and "cc" as lists for sending emails

redapple:
Fixes GH-2244


kmike:
What do you think about checking for strings and raising an error? Passing a string is an easy mistake to make. Or maybe we can support passing a single string, why not.

redapple:
@kmike , why not.
Can we log the issue as feature request and implement it after 1.2 though?

kmike:
@redapple yep, makes sense

PR #2275, opened 2016-09-21
[MRG+1] Add better messages for when response content isn't text (closes #2264)

### LINKED ISSUE #2264: Idea: warn users when trying to use TextResponse functionality with plain Response
Currently, if we try to use TextResponse functionality like response.text or css()/xpath() methods with a plain Response (e.g. in case of binary content), we get an AttributeError:

```
>>> response.css
AttributeError                            Traceback (most recent call last)
<ipython-input-1-7d6e256164d4> in <module>()
----> 1 response.css

AttributeError: 'Response' object has no attribute 'css'
>>> response.xpath
AttributeError                            Traceback (most recent call last)
<ipython-input-2-4f61f6e9fc6e> in <module>()
----> 1 response.xpath

AttributeError: 'Response' object has no attribute 'xpath'
>>> response.text
AttributeError                            Traceback (most recent call last)
<ipython-input-3-be6a4a00df5e> in <module>()
----> 1 response.text

AttributeError: 'Response' object has no attribute 'text'
```

Would it make sense to add a few methods/properties to explain what's going on for new users?

I was thinking instead of AttributeError, a better behavior could be a ValueError with a message giving a bit more context.

So, in plain `Response`, we could have:

```
def css(self, *args, **kw):
    raise ValueError('Response content is not text')

def xpath(self, *args, **kw):
    raise ValueError('Response content is not text')

@property
def text(self, *args, **kw):
    raise ValueError('Response content is not text')
```

This would be nice, because we'd had to explain fewer things when teaching people about responses and also about using `.css` and `.xpath` methods.

What do you think?

eliasdorneles:
Here is a implementation for the idea at #2264.

Before, for a binary response:

```
>>> response.text
AttributeError: 'Response' object has no attribute 'text'
>>> response.css('body')
AttributeError: 'Response' object has no attribute 'css'
```

After:

```
>>> response.text
  File "/home/elias/hacking/scrapinghub/scrapy/scrapy/http/response/__init__.py", line 91, in text
    raise AttributeError("Response content isn't text")
AttributeError: Response content isn't text
>>> response.css('body')
  File "/home/elias/hacking/scrapinghub/scrapy/scrapy/http/response/__init__.py", line 97, in css
    raise NotSupported("Response content isn't text")
scrapy.exceptions.NotSupported: Response content isn't text
>>> 
```

Does this look good?

### DIFF
--- a/scrapy/http/response/__init__.py
+++ b/scrapy/http/response/__init__.py
@@ -9,6 +9,8 @@ from six.moves.urllib.parse import urljoin
 from scrapy.http.headers import Headers
 from scrapy.utils.trackref import object_ref
 from scrapy.http.common import obsolete_setter
+from scrapy.exceptions import NotSupported
+
 
 class Response(object_ref):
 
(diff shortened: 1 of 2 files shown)


redapple:
Had missed that one. Thanks @eliasdorneles !

PR #2282, opened 2016-09-23
[docs] document that process_item can return Deferred

### LINKED ISSUE #2230: document that process_item can return deferred
pipeline [`process_item`](http://doc.scrapy.org/en/1.1/topics/item-pipeline.html#process_item) method can return [`Twisted.Deferred`](https://twistedmatrix.com/documents/current/core/howto/defer.html). It works fine in user code and we also do this in some default pipelines, e.g. [media pipeline](https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/media.py#L57). 

We should document that this is possible and perhaps add some examples to show users how to do it.

pawelmhm:
fixes #2230

### DIFF
--- a/docs/topics/item-pipeline.rst
+++ b/docs/topics/item-pipeline.rst
@@ -27,9 +27,10 @@ Each item pipeline component is a Python class that must implement the following
 
 .. method:: process_item(self, item, spider)
 
-   This method is called for every item pipeline component and must either return
-   a dict with data, :class:`~scrapy.item.Item` (or any descendant class) object 
-   or raise a :exc:`~scrapy.exceptions.DropItem` exception. Dropped items are no longer
+   This method is called for every item pipeline component. :meth:`process_item`
+   must either: return a dict with data, return an :class:`~scrapy.item.Item`
+   (or any descendant class) object, return a `Twisted Deferred`_ or raise
+   :exc:`~scrapy.exceptions.DropItem` exception. Dropped items are no longer
    processed by further pipeline components.
 
    :param item: the item scraped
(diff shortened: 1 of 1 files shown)


redapple:
Thanks @pawelmhm !
A handy example too

docs/topics/item-pipeline.rst
```diff
@@ -163,6 +166,52 @@ method and how to clean up the resources properly.
...
+
+    class ScreenshotPipeline(object):
+        """Pipeline that uses Splash to render screenshot of
+        every Scrapy item."""
+
+        SPLASH_URL = "http://localhost:8050/render.png?url={}"
+
+        def process_item(self, item, spider):
+            item_url = item["url"]
+            screenshot_url = self.SPLASH_URL.format(item_url)
```
kmike: this is not a good way to build an URL, GET arguments must be scaped

docs/topics/item-pipeline.rst
```diff
@@ -163,6 +166,52 @@ method and how to clean up the resources properly.
...
+            request = scrapy.Request(screenshot_url)
+            dfd = spider.crawler.engine.download(request, spider)
+            dfd.addBoth(self.return_item, item)
+            return dfd
```
pawelmhm: I'm not sure, maybe it is possible, but the goal of example is to show how to return Deferred from process_item. Example with inline callbacks would not demonstrate this clearly, it would return generator decorated with `@inlineCallbacks`, we would also have to explain what are inline callbacks, link to some docs for them.

Other comments see #2287

PR #2287, opened 2016-09-24
[MRG+1][docs/item_pipeline] process_item returning Deferred docs

pawelmhm:
* quote url
* use hash of url as filename

cc @kmike


eliasdorneles:
Thanks @pawelmhm !

Issue #2278, opened 2016-09-22
URLs vs Requests in at Architecture overview page 


kmike:
I think the [architecture overview](http://doc.scrapy.org/en/master/topics/architecture.html) in Scrapy master docs doesn't explain correctly the difference between URLs and Requests. It says that spider returns URLs, and Engine converts them to Requests, but this is confusing - spider returns Requests. There are also "URLs" on the chart, but Scrapy passes Request instances almost everywhere, not URLs. 

Old explanation (before https://github.com/scrapy/scrapy/pull/2165) didn't have this issue; as it is a regression I added '1.2' milestone.

uberux:
Here's the dataflow diagram corrected (REQUESTS instead of URLS)

<img width="700" alt="scrapy_dataflow_diagram" src="https://cloud.githubusercontent.com/assets/2480693/18926553/a1433e86-858e-11e6-8c3e-c145cd255a7c.png">

eliasdorneles:
Thank you @uberux !

PR #2293, opened 2016-09-28
docs: update data flow description and image (fixes: #2278)

eliasdorneles:
This fixes the explanation to use Requests instead of URLs,
which is what actually happens, and is also consistent with the
new tutorial, which already explains how URLs become Request objects.

I've also changed the "loop", jumping from 9 to step 2.

Does this look good?


lopuhin:
Looks good! 👍

redapple:
Thanks @eliasdorneles !

eliasdorneles:
Thanks @uberux for the updated graphics ! :+1:

PR #2296, opened 2016-09-29
[MRG+1] architecture docs: restore explanation loop to step 1 (see comment in 3ac3ac4)

eliasdorneles:
This brings back the original loop in the architecture explanation, as per comment https://github.com/scrapy/scrapy/commit/3ac3ac4d925ee7ef1eb8be8dc6519606c0fbf9b0#commitcomment-19223839

### DIFF
--- a/docs/topics/architecture.rst
+++ b/docs/topics/architecture.rst
@@ -29,7 +29,7 @@ Data flow
 The data flow in Scrapy is controlled by the execution engine, and goes like
 this:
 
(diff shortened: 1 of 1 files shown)


redapple:
This is ok for now I guess, but the use of "the first Requests to crawl from the spider" in step 1 is not clear (to me at least), it's more about getting batches from `start_requests`, or?

eliasdorneles:
As I understand, it would be still getting the initial requests from `start_requests`.

It's hard to achieve that balance between preciseness and ease to understand. :confused:

PR #2300, opened 2016-10-02
Remove duplicate colons from the feed export settings docs

mineo:


redapple:
Well spotted @mineo ! Thanks

PR #2302, opened 2016-10-03
[MRG+1] Fix JsonWriterPipeline example in docs

Granitosaurus:
The current example does not save the file properly if crawler is called from within a script since file object is never closed. 

related stackoverflow question: http://stackoverflow.com/questions/39821693/scrapy-print-pipeline-data-in-a-scripts-context

### DIFF
--- a/docs/topics/item-pipeline.rst
+++ b/docs/topics/item-pipeline.rst
@@ -106,8 +106,11 @@ format::
 
    class JsonWriterPipeline(object):
 
-       def __init__(self):
+       def open_spider(self, spider):
            self.file = open('items.jl', 'wb')
+
+       def close_spider(self, spider):
+           self.file.close()
 
        def process_item(self, item, spider):
            line = json.dumps(dict(item)) + "\n"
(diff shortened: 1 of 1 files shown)


kmike:
Hey @Granitas,

Yeah, the example is not correct; this is mentioned in the note to the next example. It is better to either to fix the example and remove the note, or move the note closer to JsonWriterPipeline example.

redapple:
@Granitas , the example definitely needs fixing, thanks for this.
I would also keep the note. I see no harm in it.

kmike:
@redapple I'm talking about this note:

> Previous example (JsonWriterPipeline) doesn't clean up resources properly.
Fixing it is left as an exercise for the reader.

redapple:
Oh, right, I didn't even see the note on GitHub.
Good to remove.

Granitas:
Got rid of the comment, it was really weirdly placed so I didn't even notice it.  
Removed the note.

kmike:
Thanks!

PR #2306, opened 2016-10-05
[MRG] Referrer policies in RefererMiddleware

### LINKED ISSUE #2142: Scrapy uses `file://` URL as referer
If I use a local file in a request (e.g. 

``` python
def start_requests(self):
    yield Request(path_to_file_uri(filename))

def parse(self, response):
    for url in json.loads(response.text):
        yield Request(url)
```

and then make a HTTP request in the callback, scrapy uses the file URL as the `referer` in the HTTP header. 

I think `file://` URL should be avoided from being used as the `referer` value (or generally any protocol other than http/s) because
1. it is not part of the normal hyperlink navigation
2. it exposes the local file structure to the remote site

redapple:
Fixes #2142

I've based the tests and implementation on https://www.w3.org/TR/referrer-policy/

By default, with this change, `RefererMiddleware` does not send a `Referer` header (referrer value):
- when the source response is `file://` or `s3://`,
- nor from `https://` response to an `http://` request.

However, it does send a referrer value for any `https://` to another `https://` (as browsers do, [`"no-referrer-when-downgrade"` being the default policy](https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade))
This needs discussion.

User can change the policy per-request, using a new `referrer_policy` meta key, with values from "no-referrer" / "no-referrer-when-downgrade" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url".

~~Still missing:~~
- [x] Try to use `urlparse_cached()`
- [x] Make default policy customizable through a `REFERER_DEFAULT_POLICY` setting
- [x] Test custom policies
- [x] Docs updates
- [x] Add tests for policy given by response headers (`Referrer-Policy`) -- is this even used in practice by web servers?
- [x] Handle referrers [during redirects](https://www.w3.org/TR/referrer-policy/#set-requests-referrer-policy-on-redirect)

To handle redirects, I added a signal handler on `request_scheduled`. I did not find a way to test both redirect middleware and referer middleware (I did not search too much though).
Suggestions to do that are welcome.

### DIFF
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -333,7 +333,7 @@ Whether to enable referer middleware.
 REFERER_POLICY
 ^^^^^^^^^^^^^^
 
-.. versionadded:: 1.3
+.. versionadded:: 1.4
 
 Default: ``'scrapy.spidermiddlewares.referer.DefaultReferrerPolicy'``
 
@@ -346,14 +346,14 @@ This setting accepts:
   (see ``scrapy.spidermiddlewares.referer``),
 - or one of the standard W3C-defined string values:
 
-  - `"no-referrer" <https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer>`_,
-  - `"no-referrer-when-downgrade" <https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade>`_,
-  - `"same-origin" <https://www.w3.org/TR/referrer-policy/#referrer-policy-same-origin>`_,
-  - `"origin" <https://www.w3.org/TR/referrer-policy/#referrer-policy-origin>`_,
-  - `"strict-origin" <https://www.w3.org/TR/referrer-policy/#referrer-policy-strict-origin>`_,
-  - `"origin-when-cross-origin" <https://www.w3.org/TR/referrer-policy/#referrer-policy-origin-when-cross-origin>`_,
-  - `"strict-origin-when-cross-origin" <https://www.w3.org/TR/referrer-policy/#referrer-policy-strict-origin-when-cross-origin>`_,
-  - or `"unsafe-url" <https://www.w3.org/TR/referrer-policy/#referrer-policy-unsafe-url>`_
+  - `"no-referrer"`_,
+  - `"no-referrer-when-downgrade"`_,
+  - `"same-origin"`_,
+  - `"origin"`_,
+  - `"strict-origin"`_,
+  - `"origin-when-cross-origin"`_,
+  - `"strict-origin-when-cross-origin"`_,
+  - or `"unsafe-url"`_
     (not recommended).
 
   (It can also be the non-standard value ``"scrapy-default"`` to use
@@ -364,14 +364,22 @@ with the addition that "Referrer" is not sent if the parent request was
 using ``file://`` or ``s3://`` scheme.
 
 .. warning::
-    By default, Scrapy's default referrer policy, just like `"no-referrer-when-downgrade"`_,
+    Scrapy's default referrer policy, just like `"no-referrer-when-downgrade"`_,
     will send a non-empty "Referer" header from any ``https://`` to any ``https://`` URL,
     even if the domain is different.
     ``same-origin`` may be a better choice if you want to remove referrer
     information for cross-domain requests.
 
 .. _Referrer Policy: https://www.w3.org/TR/referrer-policy
+.. _"no-referrer": https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer
 .. _"no-referrer-when-downgrade": https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade
+.. _"same-origin": https://www.w3.org/TR/referrer-policy/#referrer-policy-same-origin
+.. _"origin": https://www.w3.org/TR/referrer-policy/#referrer-policy-origin
+.. _"strict-origin": https://www.w3.org/TR/referrer-policy/#referrer-policy-strict-origin
+.. _"origin-when-cross-origin": https://www.w3.org/TR/referrer-policy/#referrer-policy-origin-when-cross-origin
+.. _"strict-origin-when-cross-origin": https://www.w3.org/TR/referrer-policy/#referrer-policy-strict-origin-when-cross-origin
+.. _"unsafe-url": https://www.w3.org/TR/referrer-policy/#referrer-policy-unsafe-url
+
 
 UrlLengthMiddleware
 -------------------
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -349,7 +349,8 @@ This setting accepts:
 - or one of the standard W3C-defined string values:
 
   - `"no-referrer"`_,
-  - `"no-referrer-when-downgrade"`_,
+  - `"no-referrer-when-downgrade"`_
+    (the W3C-recommended default, used by major web browsers),
   - `"same-origin"`_,
   - `"origin"`_,
   - `"strict-origin"`_,
@@ -358,18 +359,19 @@ This setting accepts:
   - or `"unsafe-url"`_
     (not recommended).
 
-  (It can also be the non-standard value ``"scrapy-default"`` to use
-  Scrapy's default referrer policy.)
+It can also be the non-standard value ``"scrapy-default"`` to use
+Scrapy's default referrer policy.
 
 Scrapy's default referrer policy is a variant of `"no-referrer-when-downgrade"`_,
 with the addition that "Referrer" is not sent if the parent request was
 using ``file://`` or ``s3://`` scheme.
 
 .. warning::
-    Scrapy's default referrer policy, just like `"no-referrer-when-downgrade"`_,
-    will send a non-empty "Referer" header from any ``https://`` to any ``https://`` URL,
+    Scrapy's default referrer policy—just like `"no-referrer-when-downgrade"`_,
+    the W3C-recommended value for browsers—will send a non-empty
+    "Referer" header from any ``http(s)://`` to any ``https://`` URL,
     even if the domain is different.
-    ``same-origin`` may be a better choice if you want to remove referrer
+    `"same-origin"`_ may be a better choice if you want to remove referrer
     information for cross-domain requests.
 
 .. note::
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -346,21 +346,21 @@ This setting accepts:
 - a path to a ``scrapy.spidermiddlewares.referer.ReferrerPolicy`` subclass,
   either a custom one or one of the built-in ones
   (see ``scrapy.spidermiddlewares.referer``),
-- or one of the standard W3C-defined string values:
+- or one of the standard W3C-defined string values
 
-  - `"no-referrer"`_,
-  - `"no-referrer-when-downgrade"`_
-    (the W3C-recommended default, used by major web browsers),
-  - `"same-origin"`_,
-  - `"origin"`_,
-  - `"strict-origin"`_,
-  - `"origin-when-cross-origin"`_,
-  - `"strict-origin-when-cross-origin"`_,
-  - or `"unsafe-url"`_
-    (not recommended).
-
-It can also be the non-standard value ``"scrapy-default"`` to use
-Scrapy's default referrer policy.
+=======================================  ========================================================================  =======================================================
+String value                             Class name
+=======================================  ========================================================================  =======================================================
+`"no-referrer"`_                         ``'scrapy.spidermiddlewares.referer.NoReferrerPolicy'``
+`"no-referrer-when-downgrade"`_          ``'scrapy.spidermiddlewares.referer.NoReferrerWhenDowngradePolicy'``      the W3C-recommended default, used by major web browsers
+`"same-origin"`_                         ``'scrapy.spidermiddlewares.referer.SameOriginPolicy'``
+`"origin"`_                              ``'scrapy.spidermiddlewares.referer.OriginPolicy'``
+`"strict-origin"`_                       ``'scrapy.spidermiddlewares.referer.StrictOriginPolicy'``
+`"origin-when-cross-origin"`_            ``'scrapy.spidermiddlewares.referer.OriginWhenCrossOriginPolicy'``
+`"strict-origin-when-cross-origin"`_     ``'scrapy.spidermiddlewares.referer.StrictOriginWhenCrossOriginPolicy'``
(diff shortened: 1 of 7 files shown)


scrapy/spidermiddlewares/referer.py
```diff
@@ -2,22 +2,282 @@
...
+        stripped = self.strip_url(req_or_resp)
+        if stripped is not None:
+            return urlunparse(stripped)
+
+    def origin_referrer(self, req_or_resp):
+        stripped = self.strip_url(req_or_resp, origin_only=True)
+        if stripped is not None:
+            return urlunparse(stripped)
+
+    def strip_url(self, req_or_resp, origin_only=False):
```
kmike: What do you think about making it an utility function?

redapple: @kmike , I moved some of it to `scrapy.utils.url`. What do you think?

scrapy/utils/url.py
```diff
@@ -103,3 +103,35 @@ def guess_scheme(url):
         return any_to_uri(url)
     else:
         return add_http_if_no_scheme(url)
+
+
+def strip_url_credentials(url, origin_only=False, keep_fragments=False):
```
kmike: Could you please add a docstring to this function?

The function is named 'strip credentials', but it also strips port, I think this is not intuitive.

redapple: Right, it strips standard HTTP and HTTPS port numbers. This is mainly for comparing origins.
Docstring is needed indeed.
Does it make sense to have this feature as switchable through an argument?

redapple:
I'm adding 2 policies that appeared in [the working draft](https://www.w3.org/TR/2016/WD-referrer-policy-20161222/):
- [`"strict-origin"`](https://www.w3.org/TR/2016/WD-referrer-policy-20161222/#referrer-policy-strict-origin)
- [`"strict-origin-when-cross-origin"`](https://www.w3.org/TR/2016/WD-referrer-policy-20161222/#referrer-policy-strict-origin-when-cross-origin)

redapple:
Alright, @kmike , @eliasdorneles , @lopuhin , @dangra , I think this is ready for review.
Thanks in advance!

redapple:
(I'm not sure why I use `ftps` in some tests. It's not supported by Scrapy)

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,51 @@ Default: ``True``
...
+
+  (It can also be the non-standard value ``"scrapy-default"`` to use
+  Scrapy's default referrer policy.)
+
+Scrapy's default referrer policy is a variant of `"no-referrer-when-downgrade"`_,
+with the addition that "Referrer" is not sent if the parent request was
+using ``file://`` or ``s3://`` scheme.
+
+.. warning::
+    By default, Scrapy's default referrer policy, just like `"no-referrer-when-downgrade"`_,
```
kmike: "By default" looks redundant

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,66 @@ Default: ``True``
...
+  Scrapy's default referrer policy.)
+
+Scrapy's default referrer policy is a variant of `"no-referrer-when-downgrade"`_,
+with the addition that "Referrer" is not sent if the parent request was
+using ``file://`` or ``s3://`` scheme.
+
+.. warning::
+    Scrapy's default referrer policy, just like `"no-referrer-when-downgrade"`_,
+    will send a non-empty "Referer" header from any ``https://`` to any ``https://`` URL,
+    even if the domain is different.
```
kmike: If that's what browsers do by default then I think it makes sense to note that, to make a warning less scary.

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,68 @@ Default: ``True``
...
+
+.. reqmeta:: referrer_policy
+
+`Referrer Policy`_ to apply when populating Request "Referer" header.
+
+This setting accepts:
+
+- a path to a ``scrapy.spidermiddlewares.referer.ReferrerPolicy`` subclass,
+  either a custom one or one of the built-in ones
+  (see ``scrapy.spidermiddlewares.referer``),
```
kmike: Is there a page in the docs we can reference to?

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,68 @@ Default: ``True``
...
+.. setting:: REFERER_POLICY
+
+REFERER_POLICY
+^^^^^^^^^^^^^^
+
+.. versionadded:: 1.4
+
+Default: ``'scrapy.spidermiddlewares.referer.DefaultReferrerPolicy'``
+
+.. reqmeta:: referrer_policy
```
kmike: It could make sense use the same spelling everywhere in Scrapy. Currently option name (REFERER_POLICY) has one 'r' while meta key name has 'rr'.

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,68 @@ Default: ``True``
...
+  - `"origin-when-cross-origin"`_,
+  - `"strict-origin-when-cross-origin"`_,
+  - or `"unsafe-url"`_
+    (not recommended).
+
+It can also be the non-standard value ``"scrapy-default"`` to use
+Scrapy's default referrer policy.
+
+Scrapy's default referrer policy is a variant of `"no-referrer-when-downgrade"`_,
+with the addition that "Referrer" is not sent if the parent request was
```
kmike: header name is Referer

docs/topics/spider-middleware.rst
```diff
@@ -328,6 +328,68 @@ Default: ``True``
...
+using ``file://`` or ``s3://`` scheme.
+
+.. warning::
+    Scrapy's default referrer policy—just like `"no-referrer-when-downgrade"`_,
```
kmike: a nitpick - there should be spaces before and after a dash :)

docs/topics/spider-middleware.rst
kmike: aha, I see, we already have REFERER_ENABLED

redapple: I'm ok with changing to `REFERRER_POLICY`. (It's a shame about that spelling mistake in the RFCs.)

docs/topics/spider-middleware.rst
redapple: What kind of reference do you have in mind? built-in classes?

We could also define an interface for policies, so people know what to implement.

redapple: @kmike , an alternative is to use a table:

```
$ git diff
diff --git a/docs/topics/spider-middleware.rst b/docs/topics/spider-middleware.rst
index 9010ae7..46a84d0 100644
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -346,21 +346,21 @@ This setting accepts:
 - a path to a ``scrapy.spidermiddlewares.referer.ReferrerPolicy`` subclass,
   either a custom one or one of the built-in ones
   (see ``scrapy.spidermiddlewares.referer``),
-- or one of the standard W3C-defined string values:
-
-  - `"no-referrer"`_,
-  - `"no-referrer-when-downgrade"`_
-    (the W3C-recommended default, used by major web browsers),
-  - `"same-origin"`_,
-  - `"origin"`_,
-  - `"strict-origin"`_,
-  - `"origin-when-cross-origin"`_,
-  - `"strict-origin-when-cross-origin"`_,
-  - or `"unsafe-url"`_
-    (not recommended).
-
-It can also be the non-standard value ``"scrapy-default"`` to use
-Scrapy's default referrer policy.
+- or one of the standard W3C-defined string values
+
+=======================================  ========================================================================  =======================================================
+String value                             Class name
+=======================================  ========================================================================  =======================================================
+`"no-referrer"`_                         ``'scrapy.spidermiddlewares.referer.NoReferrerPolicy'``
+`"no-referrer-when-downgrade"`_          ``'scrapy.spidermiddlewares.referer.NoReferrerWhenDowngradePolicy'``      the W3C-recommended default, used by major web browsers
+`"same-origin"`_                         ``'scrapy.spidermiddlewares.referer.SameOriginPolicy'``
+`"origin"`_                              ``'scrapy.spidermiddlewares.referer.OriginPolicy'``
+`"strict-origin"`_                       ``'scrapy.spidermiddlewares.referer.StrictOriginPolicy'``
+`"origin-when-cross-origin"`_            ``'scrapy.spidermiddlewares.referer.OriginWhenCrossOriginPolicy'``
+`"strict-origin-when-cross-origin"`_     ``'scrapy.spidermiddlewares.referer.StrictOriginWhenCrossOriginPolicy'``
+`"unsafe-url"`_                          ``'scrapy.spidermiddlewares.referer.UnsafeUrlPolicy'``                    NOT recommended
+``"scrapy-default"``                     ``'scrapy.spidermiddlewares.referer.DefaultReferrerPolicy'``              Scrapy's default policy (see below)
+=======================================  ========================================================================  =======================================================
 
 Scrapy's default referrer policy is a variant of `"no-referrer-when-downgrade"`_,
 with the addition that "Referer" is not sent if the parent request was
```

kmike: I like such table. 

Also, adding autodocs for these policy classes looks helpful - they already have docstrings, it is just a matter of putting several `autoclass` labels to scrapy rst docs.

kmike: as for "see ``scrapy.spidermiddlewares.referer``", the question was how would user find this ``scrapy.spidermiddlewares.referer`` - there is no link or anything.

kmike: Documenting policy interface sounds like too much for me; I'm not sure this is something users may want to extend.

redapple:
@kmike , I rebased, updated the docs with `autoclass` directives and renamed the setting to `REFERRER_POLICY`.

redapple:
Another thing could be to move the policies to scrapy/extensions, similar to HTTP cache policies (with `scrapy/scrapy/extensions/httpcache.py`). What do you think?

docs/topics/spider-middleware.rst
kmike: We're loosing either way - now option names are inconsistent :)

kmike: I guess the way to go is to rename the other option and a middleware itself as well; of course, it can be done later.

scrapy/spidermiddlewares/referer.py
```diff
@@ -2,22 +2,322 @@
...
+                # expect a string for the path to the policy class
+                try:
+                    self.default_policy = load_object(policy)
+                except ValueError:
+                    # otherwise try to interpret the string as standard
+                    # https://www.w3.org/TR/referrer-policy/#referrer-policies
+                    try:
+                        self.default_policy = _policy_classes[policy.lower()]
+                    except:
+                        raise NotConfigured("Unknown referrer policy name %r" % policy)
```
kmike: I think it is better to raise an error and stop the crawl instead of silently disabling the middleware if a policy is unknown.

scrapy/spidermiddlewares/referer.py
```diff
@@ -2,22 +2,322 @@
...
+        return mw
+
+    def policy(self, response, request):
+        # policy set in request's meta dict takes precedence over default policy
+        policy_name = request.meta.get('referrer_policy')
+        if policy_name is None:
+            policy_name = to_native_str(
+                response.headers.get('Referrer-Policy', '').decode('latin1'))
+
+        cls = _policy_classes.get(policy_name.lower(), self.default_policy)
```
kmike: What do you think about warning for unknown policies? They can be typos.

scrapy/spidermiddlewares/referer.py
redapple: Good point!

scrapy/spidermiddlewares/referer.py
redapple: Makes sense.

scrapy/spidermiddlewares/referer.py
```diff
@@ -2,22 +2,322 @@
...
+POLICY_STRICT_ORIGIN_WHEN_CROSS_ORIGIN = "strict-origin-when-cross-origin"
+POLICY_UNSAFE_URL = "unsafe-url"
+POLICY_SCRAPY_DEFAULT = "scrapy-default"
+
+
+class ReferrerPolicy(object):
+
+    NOREFERRER_SCHEMES = LOCAL_SCHEMES
+
+    def referrer(self, response, request):
```
kmike: It seems the API requires response and request (not response_url and request_url) only because of urlparse_cached - you wanted to avoid duplicate computations. It leads to some ugly code when a fake Response is created. Is this correct, or are there cases response and requestis indeed needed?

I wonder how much is the effect, given that urlparse has its own internal cache (using urls askeys, not response/requests).

kmike:
@redapple I think current code structure is fine; policies are not Scrapy extensions

scrapy/spidermiddlewares/referer.py
redapple: Right @kmike , that was the reason.
I cannot think right now of a real need for handling request/responses directly. Meta information and HTTP headers ("Referrer-Policy") are to be interpreted at middleware level, and not at policy level I believe.

Indeed the fake response is quite ugly.
Let me clean it up with URLs all along.

redapple:
@kmike , I think I have addressed all your comments. (and rebased)

kmike:
@redapple tests are failing - could you please check it?

redapple:
@kmike , done. Sorry about that.

scrapy/spidermiddlewares/referer.py
```diff
@@ -2,22 +2,360 @@
...
+        """
+        if not url:
+            return None
+        return strip_url(url,
+                         strip_credentials=True,
+                         strip_fragment=True,
+                         strip_default_port=True,
+                         origin_only=origin_only)
+
+    def origin(self, r):
```
kmike: Is there a reason some of the arguments named `r` while other are named `url`?

redapple: Not really. I kept request/response when the 2 arguments are needed. But I can use "url" in others.

kmike: Hm, what's the reason arguments are still named request/response, and not something like request_url? Maybe I got a bit too used to type hints :)

redapple: Most probably laziness and readability. I can change that too if you prefer.

kmike: Could you please change it? When looking at code I expect scrapy Request and Response objects to be passed to .referrer method, to figure out that they are urls you need to find calling code. I agree that `request_url` looks worse, but imho it can help readability here.

redapple: Done.

kmike:
Looks great, thanks @redapple!

PR #2307, opened 2016-10-05
genspider: removing www. from starturl templates

### LINKED ISSUE #2299: Genspider command prepends an inconvenient 'www' in start_urls
Almost every time I use the `genspider` command, I end up removing the `www` prefix from the value that this command generates for `start_urls`.

For example, this command:

    $ scrapy genspider quotes quotes.toscrape.com

generates this spider:

        name = "quotes"
        allowed_domains = ["quotes.toscrape.com"]
        start_urls = (
            'http://www.quotes.toscrape.com/',
        )
        ...

IMHO, there's no need to add `www` to the `start_urls`, because it might be more annoying than helpful to people using `genspider`.

Thoughts on it?

eLRuLL:
fixes #2299

### DIFF
--- a/scrapy/templates/spiders/basic.tmpl
+++ b/scrapy/templates/spiders/basic.tmpl
@@ -5,9 +5,7 @@ import scrapy
 class $classname(scrapy.Spider):
     name = "$name"
     allowed_domains = ["$domain"]
-    start_urls = (
-        'http://$domain/',
-    )
+    start_urls = ['http://$domain/']
 
     def parse(self, response):
         pass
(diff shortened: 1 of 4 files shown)


stummjr:
LGTM.

On that note, is there any specific reason why the `basic` template generates `start_urls` as a tuple while the other templates generate it as a list?

I looks a bit weird, given that `allowed_domains` is also a list:

        name = "myspider"
        allowed_domains = ["scrapy.org"]
        start_urls = (
            'http://scrapy.org/',
        )

Thoughts?

redapple:
+1 to make `start_urls` a list

PR #2313, opened 2016-10-07
fix example code

pencoa:
In the AuthorSpider, original css selector failed to get links of author pages

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -564,7 +564,7 @@ this time for scraping author information::
 
         def parse(self, response):
             # follow links to author pages
-            for href in response.css('.author a::attr(href)').extract():
+            for href in response.css('span a::attr(href)').extract():
                 yield scrapy.Request(response.urljoin(href),
                                      callback=self.parse_author)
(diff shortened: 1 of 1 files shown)


docs/intro/tutorial.rst
```diff
@@ -564,7 +564,7 @@ this time for scraping author information::
 
         def parse(self, response):
             # follow links to author pages
-            for href in response.css('.author a::attr(href)').extract():
+            for href in response.css('span a::attr(href)').extract():
```
pencoa: Hi, eliasdorneles
Of course, I agree that it is important to write code explicitly. It is my first time to contact with scrapy and web crawl. I followed the tutorial and found the original code does't work on my computer. So I slightly changed it.

Thanks for your attention!

eliasdorneles: Great, that's cool, welcome and enjoy the journey! :)

Thanks for the change, can you update it to use `.author + ` instead of `span`?
I think you can do it directly here on Github, using the edit button (pencil icon at the right of the View button at the top of the file).

PR #2315, opened 2016-10-07
Ability to respect URL fragments when calculating request fingerprint.

bmgraff:
I ran into a situation where a site was using url fragments (url.com/page/#key=value; everything after #) to affect what data appeared on the page. Nothing was passing the dupe filter, because that was the difference in the url, and the dupe filter doesn't respect fragments. This code fixes the problem, and should have limited effect elsewhere.


nyov:
I am not quite convinced this is a good thing to add, by itself.
You are obviously crawling some javascript resource and using some javascript interpreter in the workflow - or this wouldn't make sense.

Perhaps this should become an alternative `AjaxCrawlDupeFilter`, since I believe there are some caveats. Also, portia/splash should have a solution to this already.

For example url fragments can be anything, only limited by the imagination of the developer, javascript's string parsing and maybe the browser url length limitation. For that reason I would recommend against caching it at all, except for your custom use case where you know the site.
Example: `url.com/#?k1=v1&k2=v2` vs `url.com/#?k2=v2&k1=v1` would be considered different (two hashes), which may or may not be what you want.
So a url referring to itself with a link like `/self.php#refresh=<randint>` would loop forever without some safeguard (that dupefilter is supposed to be), exploding the hash cache and starving your memory.


From a technical perspective, url fragments are never transmitted to the server, they are only local anchors (ignoring the `#!`/`?_escaped_fragments` case, which the `ajaxcrawl` middleware handles).
So for dupefilter and the downloader, these requests are very much the same, and filtering them out is sensible, instead of fetching the same url/page from the server over and over on every potential ajax interaction (fragment url call).

A caching proxy will remedy this (unless perhaps with `pragma:no-cache`), but I'm not sure what the builtin httpcache would do here by default.

This is a bit of a pickle, but I may be missing an existing solution here.
The core scheduler's dupefilter should perhaps consider these resources different. But I believe the network part/downloader/httpcache should not.
After all, repeat page fetches and refreshings are the very thing "AJAX scripting" is intending to work around. A browser would consider this a single view (not leaving the page) and keep it in the `BFCache`.

I think when seeing such a url (with a fragment), a custom middleware/extension might force-cache the page for the current crawling session, independent of the httpcache module, and then return it for every other 'fragment request' to it.
On a broad crawl this could quickly become a lot of cached pages, so a round-robin cache, of the last *n* pages with an url fragment attached, may be a possible tradeoff there. In essence emulating the browser `BFCache`.

kmike:
@bmgraff what are you using for downloading? As @nyov said, fragment is not passed to remote website as a part of HTTP request. scrapy.Request also escapes ajax crawlable URLs by default (`example.com/!#foo=bar`), so fragment can be removed when Request is instantiated, you can't rely on Request(url) to preserve fragment. In [scrapy-splash](https://github.com/scrapy-plugins/scrapy-splash) there is a custom request dupefilter which takes this in account, and a workaround for passing URLs with fragments to Request; just taking fragment in account is not enough.

bmgraff:
Yes, these requests are indeed going through the Splash javascript interpreter. Due to another package that I was using, I was unable to use the dupe filter included with scrapy-splash that respects fragments.

Thank you for taking the time to explain the situation to me. It is clear that this is not the optimal place to make changes. The interaction with scrapy-splash and the other package is definitely the place to look.

Thanks again.

nyov:
Thank you for taking the time to create a PR with testcase.
Sorry it's not always the right thing for upstream, but I hope it will keep working for you.

Issue #2321, opened 2016-10-12
Decoding of "Location" header on redirects using latin-1 can be wrong


redapple:
Web servers [should use encoded URLs in their "Location" headers](http://stackoverflow.com/questions/7654207/what-charset-should-be-used-for-a-location-header-in-a-301-response), but [they don't always do](http://stackoverflow.com/questions/4400678/what-character-encoding-should-i-use-for-a-http-header/4410331#4410331).

This website for example, for this URL http://www.yjc.ir/fa/news/1815565/
redirects to www.yjc.ir/fa/news/1815565/اعزام-كوهنوردان-ايراني-به-كيليمانجارو

but the bytes received are UTF-8 encoded, and not percent-escaped:
```
'Location': ['/fa/news/1815565/\xd8\xa7\xd8\xb9\xd8\xb2\xd8\xa7\xd9\x85-\xd9\x83\xd9\x88\xd9\x87\xd9\x86\xd9\x88\xd8\xb1\xd8\xaf\xd8\xa7\xd9\x86-\xd8\xa7\xd9\x8a\xd8\xb1\xd8\xa7\xd9\x86\xd9\x8a-\xd8\xa8\xd9\x87-\xd9\x83\xd9\x8a\xd9\x84\xd9\x8a\xd9\x85\xd8\xa7\xd9\x86\xd8\xac\xd8\xa7\xd8\xb1\xd9\x88']
```

[...]

PR #2322, opened 2016-10-12
Do not interpret non-ASCII bytes in "Location" and percent-encode them

redapple:
Fixes GH-2321

The idea is to not guess the encoding of "Location" header value
and simply percent-encode non-ASCII bytes,
which should then be re-interpreted correctly by the remote website
in whatever encoding was used originally.

See https://tools.ietf.org/html/rfc3987#section-3.2

This is similar to the changes to safe_url_string in
https://github.com/scrapy/w3lib/pull/45

### DIFF
--- a/scrapy/downloadermiddlewares/redirect.py
+++ b/scrapy/downloadermiddlewares/redirect.py
@@ -5,7 +5,6 @@ from w3lib.url import safe_url_string
 
 from scrapy.http import HtmlResponse
 from scrapy.utils.response import get_meta_refresh
-from scrapy.utils.python import to_native_str
 from scrapy.exceptions import IgnoreRequest, NotConfigured
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


stav:
LGTM: it's working now using 9dcad94.

I would remove the unneeded import:

    diff --git a/scrapy/downloadermiddlewares/redirect.py b/scrapy/downloadermiddlewares/redirect.py
    index fd7a9f4..db276ee 100644
    --- a/scrapy/downloadermiddlewares/redirect.py
    +++ b/scrapy/downloadermiddlewares/redirect.py
    @@ -5,7 +5,6 @@ from w3lib.url import safe_url_string

     from scrapy.http import HtmlResponse
     from scrapy.utils.response import get_meta_refresh
    -from scrapy.utils.python import to_native_str
     from scrapy.exceptions import IgnoreRequest, NotConfigured

PR #2327, opened 2016-10-17
[MRG+1] Added documentation about accessing header values

### LINKED ISSUE #2320: Document headers methods such as getlist()
It appears that current documentation does not mention how you can get values from HTTP `Headers` object.
Things like `response.headers.getlist('Set-Cookie')` to get all cookies are not explained (see https://github.com/scrapy/scrapy/issues/2319)

bopace:
Adds documentation to resolve #2320

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -508,10 +508,10 @@ Response objects
     .. attribute:: Response.headers
 
         A dictionary-like object which contains the response headers. Values can
-        be accessed from the :class:`headers` object by using :meth:`get` to
-        return the first header value with the specified name or :meth:`getlist`
-        to return all header values with the specified name. For example, this
-        call will give you all cookies in the headers::
+        be accessed using :meth:`get` to return the first header value with the
+        specified name or :meth:`getlist` to return all header values with the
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -507,7 +507,13 @@ Response objects
 
     .. attribute:: Response.headers
 
-        A dictionary-like object which contains the response headers.
+        A dictionary-like object which contains the response headers. Values can
+        be accessed from the :class:`headers` object by using :meth:`get` to
```
redapple: I don't think _"from the :class:`headers` object by"_ is necessary

redapple:
Thanks @bopace ! I nitpicked a little.

docs/topics/request-response.rst
eliasdorneles: yup, that's not resolving to a proper link -- no need to use special markup here, just headers is good.

bopace: Thanks for the heads up. I committed a fix just now.

bopace:
Thanks! How does that change look?

eliasdorneles:
Looks good, thanks @bopace !

PR #2328, opened 2016-10-18
[MRG+1] Document `download_latency` meta key

### LINKED ISSUE #2033: Document 'download_latency' key in requests' meta dict
`request.meta['download_latency']` can be useful for analysing a spider or a website, but it's not documented (see #2021).

stummjr:
This PR aims to fix #2033

Thoughts on it?

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -299,6 +299,7 @@ Those are:
 * :reqmeta:`dont_obey_robotstxt`
 * :reqmeta:`download_timeout`
 * :reqmeta:`download_maxsize`
+* :reqmeta:`download_latency`
 * :reqmeta:`proxy`
 
 .. reqmeta:: bindaddress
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -316,6 +317,13 @@ download_timeout
 The amount of time (in secs) that the downloader will wait before timing out.
 See also: :setting:`DOWNLOAD_TIMEOUT`.
 
+.. reqmeta:: download_latency
+
+download_latency
+----------------
+
+The amount of time spent to fetch the response, since the request has been
+started.
```
kmike: I think it worths mentioning that this meta key becomes available after response is downloaded, and that it should be considered read-only. Other meta keys are used for controlling Scrapy behavior, but not this one.

stummjr: I updated, according to @kmike's observations. What about the moment when the requests are considered started?

redapple: what about
> since the request has been started, i.e. HTTP message sent over the network.

(as opposed to when the Request is yielded by a callback or queued by the scheduler)

kmike: It seems I was not precise: `redirect_urls` is also used for information only.

stummjr:
@redapple @kmike I updated the description according to @redapple's suggestion. Does this look good or should I mention that this is not the only meta key that's not used to control scrapy behavior?

kmike:
@stummjr I think it makes sense to remove "While other meta keys", or re-phrase it to something like "While most other meta keys" because currently it reads like "While all other meta keys".

stummjr:
@kmike just updated the text there.

kmike:
Thanks @stummjr!

PR #2331, opened 2016-10-18
[MRG+1] Fixes issue #2272 using arg_to_iter() to wrap single values and list() to…

### LINKED ISSUE #2272: Accept single values for "to" and "cc" arguments when sending mails
Follow up to https://github.com/scrapy/scrapy/pull/2271#issuecomment-248341246

> What about checking for strings and raising an error?
> Passing a string is an easy mistake to make.
> Or maybe we can support passing a single string.

moisesguimaraes:
(fixes #2272) using arg_to_iter() to wrap single values and list() to avoid consuming from generators.

### DIFF
--- a/docs/topics/email.rst
+++ b/docs/topics/email.rst
@@ -87,13 +87,13 @@ uses `Twisted non-blocking IO`_, like the rest of the framework.
         Send email to the given recipients.
 
         :param to: the e-mail recipients
-        :type to: list
+        :type to: str or iterable
 
         :param subject: the subject of the e-mail
         :type subject: str
 
         :param cc: the e-mails to CC
-        :type cc: list
+        :type cc: str or iterable
 
         :param body: the e-mail body
         :type body: str
--- a/docs/topics/email.rst
+++ b/docs/topics/email.rst
@@ -35,12 +35,6 @@ And here is how to use it to send an e-mail (without attachments)::
 
     mailer.send(to=["someone@example.com"], subject="Some subject", body="Some body", cc=["another@example.com"])
 
-.. note::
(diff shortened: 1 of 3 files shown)


eliasdorneles:
Looks good, thanks Moisés !

redapple:
@moisesguimaraes , can you update the docs as @kmike commented?
Other than that, looks good. Thanks.

moisesguimaraes:
Hi @redapple , I'll try to do it until Friday. I was competing in an orienteering race last Sunday and I had an accident falling over my left arm. But I'm almost 100% right now ;)

moisesguimaraes:
@kmike , I was going to change from (list) to (str or list), but since any iterable works, I changed it to (str or iterable).

If you want me to change to (str or list), please let me know.

kmike:
Hey @moisesguimaraes,

Thanks for the change! I prefer "(str or list or str)" because 

1. str is also an iterable;
2. even if any iterable is supported right now it doesn't mean we should expose that; documenting it as a list looks more future-proof.

moisesguimaraes:
Did you mean (str or list of str)?

kmike:
@moisesguimaraes yep, "or" was a typo :)

moisesguimaraes:
done.

kmike:
Thanks @moisesguimaraes!

There is also a note in docs (scroll above your changes) which should be removed:

> As shown in the example above, ``to`` and ``cc`` need to be lists
of email addresses, not single addresses, and even for one recipient,
i.e. ``to="someone@example.com"`` will not work.

redapple:
LGTM. @kmike ?

kmike:
Thanks @moisesguimaraes!

PR #2332, opened 2016-10-18
Log invalid domains

### LINKED ISSUE #2250: Add note that "allowed_domains" should be a list of domains, not URLs
(just logging the issue before I forget)

It may seem obvious by the name of the attribute that `allowed_domains` is about domain names, but it's not uncommon for scrapy users to make the mistake of doing `allowed_domains = ['http://www.example.com']`

I believe it is worth adding a note in http://doc.scrapy.org/en/latest/topics/spiders.html?#scrapy.spiders.Spider.allowed_domains

luizf2:
This warns the user when using a list of urls instead of domains.
Fixes #2250. 
Does it look good?

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -79,6 +79,9 @@ scrapy.Spider
        specified in this list (or their subdomains) won't be followed if
        :class:`~scrapy.spidermiddlewares.offsite.OffsiteMiddleware` is enabled.
 
+       Note that this need to be a list of domains, not URLs (for example, use `google.com`
+       instead of `http://google.com`.
+
    .. attribute:: start_urls
 
        A list of URLs where the spider will begin to crawl from, when no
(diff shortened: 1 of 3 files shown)


docs/topics/spiders.rst
```diff
@@ -79,6 +79,9 @@ scrapy.Spider
        specified in this list (or their subdomains) won't be followed if
        :class:`~scrapy.spidermiddlewares.offsite.OffsiteMiddleware` is enabled.
 
+       Note that this need to be a list of domains, not URLs (for example, use `google.com`
+       instead of `http://google.com`.
```
redapple: missing closing `)`

scrapy/spidermiddlewares/offsite.py
```diff
@@ -50,6 +50,12 @@ def should_follow(self, request, spider):
     def get_host_regex(self, spider):
         """Override this method to implement a different offsite policy"""
         allowed_domains = getattr(spider, 'allowed_domains', None)
+        if allowed_domains is not None:
+            invalid_domains = [d for d in allowed_domains if '/' in d]
```
redapple: this can be moved after `if not allowed_domains: return re.compile('')`

dangra:
superseded by #3011

PR #2333, opened 2016-10-18
[MRG+1] DNS cache expiration.

### LINKED ISSUE #905: Improve DNS caching
Currently, Scrapy does caching of DNS by default (setting `DNSCACHE_ENABLED` is `True` on default settings).

The current cache implementation does not implement an expiration for the entries which may result in entries that be kept _ad infinitum_. We need change the current implementation to set an expiration time by default (typical TTL is 86400 seconds -24 hours-) or maybe, customizable by a setting.

ArturGaspar:
Expires entries in the DNS cache. Fixes #905.


eliasdorneles:
Looks good to me!

scrapy/resolver.py
```diff
@@ -1,17 +1,38 @@
...
 from twisted.internet import defer
 from twisted.internet.base import ThreadedResolver
 
 from scrapy.utils.datatypes import LocalCache
 
+
 # TODO: cache misses
 
-dnscache = LocalCache(10000)
+class ExpiringCache(LocalCache):
```
redapple: This could go into `scrapy.utils.datatypes` next to `LocalCache`

redapple:
Thanks @ArturGaspar ! A long-needed improvement.
Have you considered using 3rd party libraries for this? e.g. [cachetools.TTLCache](http://pythonhosted.org/cachetools/#cachetools.TTLCache)

kmike:
@redapple the implementation is simple enough, my vote is not to add a dependency for it

redapple:
@kmike , what do you think of the PR?
If we merge it, I believe it means we start working towards a 1.3.x series

scrapy/utils/datatypes.py
```diff
@@ -304,3 +305,54 @@ def __setitem__(self, key, value):
...
+            return list(self.iteritems())
+
+        def values(self):
+            return list(self.itervalues())
+    else:
+        def iterkeys(self):
+            # For iteritems() and itervalues() above.
+            return self.keys()
+        items = iteritems
+        values = itervalues
```
ArturGaspar: `items()` and `values()` are inherited from OrderedDict. Starting with Python 3.5, they will bypass the custom `__getitem__` (because OrderedDict is implemented in C now) and expose the actual `(time, value)` tuple.

In Python up to 3.4:

    >>> list(d.items())
    [('a', 1)]

In Python 3.5:

    >>> list(d.items())
    [('a', (1481032887.784956, 1))]

sibiryakov:
Guys, each DNS record has a TTL, and I suggest to use it instead of using local setting https://support.google.com/a/answer/48090#TTL. Many DNS servers are respecting DNS record TTL when caching.

redapple:
It looks like supporting DNS-server-advertized TTLs would be quite some work as the current implementation uses `twisted.internet.base.ThreadedResolver` which will only get us one IP address, without any TTL ([`socket.gethostbyname()`](https://docs.python.org/3/library/socket.html#socket.gethostbyname) is used under the hood)

What about merging this (with a lower default expiration timeout like @dangra would prefer)
and work on a newer implementation (IPv6 compatible, TTL-respecting) as another PR?

Note that there is also `twisted.internet.endpoints.HostnameEndpoint` (since Twisted 13.2) which uses `socket.getaddrinfo`. If someone wants to have a look at that.

>  * Twisted [13.2.0] now includes a HostnameEndpoint implementation which uses
IPv4 and IPv6 in parallel, speeding up the connection by using
whichever connects first (the 'Happy Eyeballs'/RFC 6555 algorithm).
(#4859)

kmike:
I'm fine with merging a simple expiring cache first, without respecting TTLs.

Gallaecio:
Tests fail in PyPy. We are testing a rather old, beta version of PyPy, so I will look into [changing that](https://github.com/scrapy/scrapy/pull/4504), and then we can see if tests start passing.

Gallaecio:
Fails on PyPy:

```
>       setting_time, value = super().__getitem__(key)
E       TypeError: 'int' object is not iterable
```

PR #2336, opened 2016-10-18
Adding an example on the documentation of writing your own middleware (fixes #1789)

### LINKED ISSUE #1789: Document from_crawler for Scrapy middlewares
See motivation https://github.com/scrapy/scrapy/issues/1788

From @kmike 

>  is not clear that Scrapy middlewares have the same features as extensions. Accessing settings is a common use case, I think we should provide examples in middleware docs, even if they are redundant.

lfmattossch:
Sup guys?

Follows up an example in the documentation on how to access settings and other crawler attributes.

Thnaks

### DIFF
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -159,6 +159,24 @@ following methods:
         :param spider: the spider to whom the start requests belong
         :type spider: :class:`~scrapy.spiders.Spider` object
 
+Accessing settings and other crawler attributes
+-----------------------------------------------
+
+In order to access settings and other crawler attributes you have to use 
+``from_crawler`` factory method::
+
+  class CustomMiddleware(object):
+
+  def __init__(self, middleware_arg):
+      self.middleware_arg = middleware_arg
+      self.log('Middleware parameter: {}'.format(middleware_arg))
+
+  @classmethod
+  def from_crawler(cls, crawler):
+      settings = crawler.settings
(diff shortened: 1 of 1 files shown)


eLRuLL:
I don't think this is a good idea, as we'll have to include this also in downloader middlewares, and pipeline sections (there is already something like this in extensions). And also a default `from_crawler` explanation is in [settings](https://doc.scrapy.org/en/latest/topics/settings.html#how-to-access-settings)

I would say a better way would be to include some "Example Middleware" (on Downloader and Spider Middleware) that would offer a simple copy-paste example to start with, there the `from_crawler` method could be included for better understanding.

Currently there are only "Writing your own ..." sections, which explain which methods to override, but the example is missing (we already have some for extensions and pipelines).

docs/topics/spider-middleware.rst
```diff
@@ -159,6 +159,24 @@ following methods:
...
+  def __init__(self, middleware_arg):
+      self.middleware_arg = middleware_arg
+      self.log('Middleware parameter: {}'.format(middleware_arg))
+
+  @classmethod
+  def from_crawler(cls, crawler):
+      settings = crawler.settings
+      middleware_arg = settings.get('MIDDLEWARE_ARG')
+    return cls(middleware_arg)
+
```
redapple: indentation does not look correct

lfmattossch: My bad @redapple 
I will take care of it!

PR #2337, opened 2016-10-18
[MRG+1] Add middleware to template project (closes #2335)

### LINKED ISSUE #2335: Add a sample middleware to startproject's template
It will be nice to have a middleware template inside the template project to serve as an example for people that want to use it.

muriloviana:
This Pull Request add a sample spider middleware to the template project.

Looks good?


stummjr:
Looks good! I would just use the same middleware name in the `settings.py.tmpl` file, instead of the [current `MyCustomSpiderMiddleware`](https://github.com/scrapy/scrapy/blob/master/scrapy/templates/project/module/settings.py.tmpl#L49).

Any thoughts?

muriloviana:
About the last commit:

This new commit set `${ProjectName}SpiderMiddleware` from `settings.py.tmpl`.

I update the instructions in each method from sample spider middleware from `middleware.py.tmpl`, please, review and see if looks good.

One thing that I realized, it that the middleware to template project was named as `middleware.py.tmpl` instead of `middlewares.py.tmpl` (on plural). This way looks better.

What you think about this changes?

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ${ProjectName}SpiderMiddleware(object):
+
+    def process_spider_input(response, spider):
+        # Called for each response that goes through the spider
+        # middleware and into the spider.
+
+        # Should return None or raise an exception.
+        pass
```
redapple: I'm leaning towards providing working but no-op methods implementations.
So what about an explicit `return None` instead of `pass` explaining that it means that the response continues in the chain?

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -0,0 +1,39 @@
...
+
+    def process_spider_output(response, result, spider):
+        # Called with the results returned from the Spider, after
+        # it has processed the response.
+
+        # Must return an iterable of Request, dict or Item objects.
+        pass
```
redapple: Same reasoning as for `process_spider_input`,
we can have the pass-through
```
for x in result:
    yield x
```

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -0,0 +1,39 @@
...
+        # or Item objects.
+        pass
+
+    def process_start_requests(start_requests, spider):
+        # Called with the start requests of the spider, and works
+        # similarly to the process_spider_output() method, except
+        # that it doesn’t have a response associated.
+
+        # Must return only requests (not items).
+        pass
```
redapple: Same as for `process_spider_output`,
I suggest having

```
for r in start_requests:
    yield r
```

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+
+# Define here the models for your spider middleware
+#
+# See documentation in:
+# http://doc.scrapy.org/en/latest/topics/spider-middleware.html
+
+
+class ${ProjectName}SpiderMiddleware(object):
+
```
redapple: What do you think of adding `from_crawler` and maybe registering a signal on `spider_opened`, (like in [`OffsiteMiddleware`](https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/offsite.py)) to show best practices?

redapple:
Another thing is that if a method is not defined, then it's equivalent to a pass-through, no-op behavior.
It could be worth a comment in the template

redapple:
Thanks @muriloviana , it's a great addition to the project templates!

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -5,22 +5,34 @@
...
 
 class ${ProjectName}SpiderMiddleware(object):
+    # If any of methods bellow is not defined, in practice, the middleware
+    # manager only add the methods that the middleware has defined.
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        # This method is used by Scrapy to create your spiders.
+        s = cls(crawler.stats)
+        crawler.signals.connect(s.spider_opened, signal=signals.spider_opened)
```
redapple: `spider_opened` would need to be defined

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -5,22 +5,34 @@
 # See documentation in:
 # http://doc.scrapy.org/en/latest/topics/spider-middleware.html
 
+from scrapy import signals
+
 
 class ${ProjectName}SpiderMiddleware(object):
+    # If any of methods bellow is not defined, in practice, the middleware
+    # manager only add the methods that the middleware has defined.
```
redapple: The sentence read weird.
What about
> Not all methods need to be defined. If a method is not defined, scrapy acts as if the spider middleware does not modify the passed objects.

scrapy/templates/project/module/middlewares.py.tmpl
muriloviana: Sorry for this.

scrapy/templates/project/module/middlewares.py.tmpl
muriloviana: Much better!

scrapy/templates/project/module/middlewares.py.tmpl
muriloviana: If I do this, looks good?

        spider.logger.info('Spider opened: %s' % spider.name)

redapple: Yes, a good example of how to use the built-in logging helpers

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -0,0 +1,56 @@
...
+
+class ${ProjectName}SpiderMiddleware(object):
+    # Not all methods need to be defined. If a method is not defined,
+    # scrapy acts as if the spider middleware does not modify the
+    # passed objects.
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        # This method is used by Scrapy to create your spiders.
+        s = cls(crawler.stats)
```
muriloviana: You are right!
This way, the template looks more clean and functional than before, I totally agree with you.

redapple:
Thanks again @muriloviana !

eliasdorneles:
Oops, sorry for taking long -- thanks @muriloviana !

PR #2338, opened 2016-10-19
Note: raw unescaped content (fixes #1704)

### LINKED ISSUE #1704: [Docs] Mention that .re() does HTML entities decoding
Documentation on Selectors says:

> extract()
> Serialize and return the matched nodes as a list of unicode strings. Percent encoded content is unquoted.
> 
> re(regex)
> Apply the given regex and return a list of unicode strings with the matches.
> regex can be either a compiled regular expression or a string which will be compiled to a regular expression using re.compile(regex)

but it doesn't say that `.re()` (and `.re_first()`) also perform HTML-entities decoding (except `&lt;` and `&amp;`)
See https://github.com/scrapy/scrapy/issues/1700
and https://stackoverflow.com/questions/34887730/how-to-extract-raw-html-from-a-scrapy-selector#comment57542664_34897754

[parsel documentation](https://parsel.readthedocs.org/en/latest/parsel.html?highlight=re_first#parsel.selector.SelectorList.re) is even less verbose.

gustavodeandrade:


redapple:
>  Files changed: 3,675

Your PR seems incorrect, too many changed files for a documentation issue #1704.
What is this PR about @gustavodeandrade ?
Are you having trouble with git?

gustavodeandrade:
Yes, Paul. It was my first PR. I'm starting to learn...
This PR should add a note about method re(regex), in scrapy/docs/topics/selectors.rst.
Sorry, I used "git add ." instead "git add file".

redapple:
Can you open another PR with only the change in scrapy/docs/topics/selectors.rst?

gustavodeandrade:
Yeah. I will try again tomorrow, when I come back.

redapple:
Ok thanks. I'm closing this one for now, then.

PR #2345, opened 2016-10-21
[MRG+1] Fix documentation about HTML entities decoding with selector extraction

### LINKED ISSUE #1704: [Docs] Mention that .re() does HTML entities decoding

gustavodeandrade:
Fixes #1704

### DIFF
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -628,8 +628,7 @@ Built-in Selectors reference
 
     .. note::
 
-        Note that ``re`` and ``re_first`` both escape HTML entities. If you want to
-        get raw unescaped content, use extract() or extract_first()
+        Note that ``re()`` and ``re_first()`` both decode HTML entities (except ``&lt;`` and ``&amp;``).
 
   .. method:: register_namespace(prefix, uri)
 
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -626,6 +626,11 @@ Built-in Selectors reference
      ``regex`` can be either a compiled regular expression or a string which
      will be compiled to a regular expression using ``re.compile(regex)``
 
+    .. note::
+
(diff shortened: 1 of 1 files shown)


docs/topics/selectors.rst
```diff
@@ -626,6 +626,11 @@ Built-in Selectors reference
      ``regex`` can be either a compiled regular expression or a string which
      will be compiled to a regular expression using ``re.compile(regex)``
 
+    .. note::
+
+        Note that ``re`` and ``re_first`` both escape HTML entities. If you want to
+        get raw unescaped content, use extract() or extract_first()
```
Digenis: It depends on the xpath whether extract()'s result is escaped or not.
Extracted elements do have their texts, attributes etc escaped.
Extracted attributes, text nodes, function results, comments etc are unescaped.
In an xpath like `node()` which is equivalent (maybe) to `*|text()|comment()`
calling extract on the resulted SelectorList
will return mixed html/non html results.

I'm not saying that all this should be explained in this note
but I think it's better to drop the second sentence.

kmike:
Looks good, thanks!

PR #2351, opened 2016-10-21
Add artwork license, show both licenses in the docs

eliasdorneles:
This introduces a license for the artwork (adapted from http://flask.pocoo.org/docs/0.11/license/#flask-artwork-license) and adds a documentation section for both the software and artwork licenses.

Looks good?

### DIFF
--- a/artwork/LICENSE
+++ b/artwork/LICENSE
@@ -1,0 +1,11 @@
+Copyright (c) by Scrapy developers.
+Some rights reserved.
+
+This logo or a modified version may be used by anyone to refer to the Scrapy project, but does not indicate endorsement by the project.
+
+Redistribution and use in source (the Gimp file) and binary forms (rendered JPEG files etc.) of the image, with or without modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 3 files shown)


redapple:
@dangra , @pablohoffman , I think you discussed amendments to this proposal about commercial uses of the artwork.
Could you add your comments in this ticket? Thanks.

pablohoffman:
I'm OK with a permissive license.

dangra:
> I'm OK with a permissive license.

do you mean permissive to the point of letting companies use the Scrapy name, logo and other artwork for commercial use like promoting their products or even include Scrapy pas part of its name? 

I would like to handle it like other serious open source projects do, commercial use only after consent from project representatives:
- http://apache.org/foundation/marks/
- https://www.djangoproject.com/trademarks/

pablohoffman:
What you suggest makes sense to me @dangra. In this case, the lack of foundation can be replaced by Scrapinghub as the legal representative. Once Scrapy grows its own foundation (some day, some day :), the rights will be transferred to that.

PR #2353, opened 2016-10-21
Fix warning to duplicated spider. Issue 2181

### LINKED ISSUE #2181: rename spider will load the old left .pyc instead of renamed spider
Hi all, I'm not sure if this is an issue.
If got `y.py` and `y.pyc` in the `spider` directory.
Then rename it with `mv y.py x.py` and update the new `x.py`.
After updates, run `scrapy crawl spider` will use the old version of spider instead of updates in the `x.py`.
Have to remove `y.pyc` first.
Thanks!

MrMenezes:
Fixes #2181

As suggested, is notified every time there is more than one file with that name.

### DIFF
--- a/scrapy/spiderloader.py
+++ b/scrapy/spiderloader.py
@@ -21,7 +21,12 @@ class SpiderLoader(object):
             
     def _load_spiders(self, module):
         for spcls in iter_spider_classes(module):
-            self._spiders[spcls.name] = spcls
+            if spcls.name in self._spiders.keys():
+                import warnings
+                warnings.warn("There are several spiders with the same name (" + spcls.name +
+                              "), this can cause unexpected behavior", UserWarning)
+        self._spiders[spcls.name] = spcls
+    
 
     def _load_all_spiders(self):
         for name in self.spider_modules:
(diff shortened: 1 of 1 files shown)


scrapy/spiderloader.py
```diff
@@ -21,7 +21,12 @@ def __init__(self, settings):
             
     def _load_spiders(self, module):
         for spcls in iter_spider_classes(module):
-            self._spiders[spcls.name] = spcls
+            if spcls.name in self._spiders.keys():
+                import warnings
+                warnings.warn("There are several spiders with the same name (" + spcls.name +
+                              "), this can cause unexpected behavior", UserWarning)
+        self._spiders[spcls.name] = spcls
```
kmike: this should be done in a loop, not after the loop

kmike:
Thanks for the PR! Could you please add a test case for it? To check if a warning is raised you can use http://doc.pytest.org/en/latest/recwarn.html; SpiderLoader tests are here: https://github.com/scrapy/scrapy/tree/master/tests/test_spiderloader.

redapple:
hi @MrMenezes , @kmike ,
I had a go at the testcase: https://github.com/MrMenezes/scrapy/pull/1
how does it look?

redapple:
@MrMenezes , I took the liberty to submit my proposed changes on top of yours in #2612.
Hence I'm closing this PR. Thanks for bootstrapping this!

PR #2355, opened 2016-10-22
Enable LOG_CUSTOM_LEVELS in order to specify 3rd party logger log levels

iserko:
When you use 3rd party libraries inside crawler code, which log all sorts of information at various logging levels, it can become quite noisy. This code change helps out by moving the LOGGING dictionary with the list of `loggers` which we can override. Example:

```
    LOG_LEVEL = 'DEBUG'
    LOG_CUSTOM_LEVELS = {
        'boto': 'ERROR'
    }
    ```

Would silence boto while allowing you to keep your `LOG_LEVEL` at `DEBUG`.

This ties into issue #2231 as it would enable us to modify loggers. It does not touch handlers, but it does give you the ability to set them up however you want and assign loggers to specific handlers (if you know how to configure logging properly).

cc @redapple

### CODE REVIEW

**Comment by redapple** [comment]
Thanks @iserko !
Can you give a shot at updating the loggins docs?
Also, I wonder if it makes sense to redirect `DEFAULT_LOGGING` to `LOGGING`.

**Comment by iserko** [comment]
@redapple will write up some docs.

Is your concern about renaming `DEFAULT_LOGGING` because we modify it later in `configure_logging` ?

**Comment by redapple** [comment]
My concern was about people using `scrapy.utils.log.DEFAULT_LOGGING` and expecting it to still be define. Not sure it's that big of a concern.

In `default_settings.py` I prefer `LOGGING` over `DEFAULT_LOGGING`, but it also could be `LOG_CONFIG` or `LOGGING_CONFIG`.

what do @dangra , @kmike , @eliasdorneles think?

**Comment by iserko** [comment]
@redapple updated my PR. Logging tests will pass now.

I've tried to add a Deprecation Class around DEFAULT_LOGGING in scrapy.utils.log as follows:

```
class DefaultLoggingDeprecation(dict):
    def __getitem__(self, key):
        warnings.warn(
            'DEFAULT_LOGGING in scrapy.utils.log is deprecated, use LOGGING setting',
            ScrapyDeprecationWarning)
        return super(DefaultLoggingDeprecation, self).__getitem__(key)

DEFAULT_LOGGING = DefaultLoggingDeprecation({
    'version': 1,
    'disable_existing_loggers': False,
    'loggers': {
        'scrapy': {
            'level': 'DEBUG',
        },
        'twisted': {
            'level': 'ERROR',
        },
    }
})
```

But logging.config.dictConfig overrides the dictionary 😂  
See: https://github.com/python/cpython/blob/master/Lib/logging/config.py#L314
used in:
https://github.com/python/cpython/blob/master/Lib/logging/config.py#L367

**Comment by iserko** [comment]
@redapple any word on this? If there's anything you'd want me to change I will.
Have been running this on our system for the past 3 weeks now without a problem.

**Comment by redapple** [comment]
hi @iserko , looks good to me.
I'd like feedback from @dangra , @kmike and @eliasdorneles .

**Comment by eliasdorneles** [comment]
Hey fellows, sorry for taking long to reply!

So, exposing the LOGGING dict is interesting because it solves the problem of silencing external libraries, but I find the behavior very counterintuitive and error prone, specially when combined with `LOG_LEVEL` setting.

For example, if you have the following in the settings:

```
LOG_LEVEL='INFO'
LOGGING = {
    'version': 1,
    'loggers': {
        'scrapy': {'level': 'DEBUG'},
    }
}
```

... the scrapy level DEBUG will be ignored, and LOG_LEVEL will have priority over it.

Also, exposing the whole dict as a setting require the user to always put some boilerplate in the dictionary (`'version': 1`) every time she wants to customize it.

It isn't great to expect user have to copy from Scrapy's `default_settings.py`. =)

Because of these, considering the main motivation for this PR is to provide a way to customize log levels to be able to reduce noise from external libraries, I propose we introduce a setting specifically for that.

Here is my proposal:

```
CUSTOM_LOG_LEVELS = {
    'scrapy': 'DEBUG',
    'boto': 'ERROR',
    'twisted': 'ERROR',
}
```

Moreover, the configuration set in this new setting would need to have priority over `LOG_LEVEL`, since it's more specific.

This would be simpler configuration (less nesting, no boilerplate) and it's easy to predict the behavior, even when combined with `LOG_LEVEL`.

What do you think?

iserko:
@eliasdorneles sorry for taking so long with a fix ... I've changed the PR to match ... still working on tests

iserko:
Tests fixed and passing @eliasdorneles. @redapple you may want to review again as well

kmike:
I like the way it is handled in django: https://docs.djangoproject.com/en/1.10/topics/logging/#configuring-logging.

iserko:
@kmike so do I, but scrapy already has existing custom logging setup which makes the transition much harder

iserko:
I can switch it back to how I originally had it :)

iserko:
@kmike @redapple @eliasdorneles any updates here?

I'm in favor of doing it the same was as Django where if you specify a `LOGGING` dict in your settings, it uses that ... otherwise it uses the default.

iserko:
@redapple ping, any idea if this will get looked at? I've just rebased it from latest master, just in case something's changed.

redapple:
hi @iserko , I'll try and get maintainers attention on this.
I personally also like `LOGGING` and even `LOGGING_CONFIG` like in Django (I'm not very familiar with Django itself, but from what I read about those, looks sensible). `LOGGING_CONFIG` could point to `scrapy.utils.log.configure_logging` for example.

iserko:
@redapple thanks. I'm happy to rewrite this PR to do your suggestion as long as there's a decision which direction to go to.

iserko:
@redapple I don't suppose this will get any traction? if so ... I'll close the PR

Issue #1877, opened 2016-03-24
Allow to add custom metadata to Response


kmike:
Currently it is not possible to attach a custom attribute to Response in a middleware because `replace` method won't keep it. There are workarounds (e.g. putting objects to response.flags), but they are ugly.

What do you think about adding an attribute to Response where user can put anything, similar to Request.meta?

kmike:
Almost the same issue: https://github.com/scrapy/scrapy/issues/1453

kmike:
One workaround is to create custom Response subclasses and use ``response.replace(cls=MyCustomResponse)`` in a middleware.

joeharrison714:
I'd like to try to implement this, does anyone have any newer thoughts about it that could lead me in the right direction?

PR #2359, opened 2016-10-25
[WIP] Add X-Scrapy-Cache-Timestamp header in cached responses

### LINKED ISSUE #2221: HTTP cache should expose storage time
FilesystemCacheStorage, DbmCacheStorage and LeveldbCacheStorage store a timestamp to meta, but unlike all other parameters this information is not available to a spider when response is retrieved from cache. 

I think Scrapy should expose this timestamp as a meta key for cached responses - it is useful not only for expiration. For example, it is required to parse relative dates properly - what "1 hour ago" means depends on a time page was downloaded.

redapple:
Fixes #2221

### DIFF
--- a/scrapy/extensions/httpcache.py
+++ b/scrapy/extensions/httpcache.py
@@ -5,7 +5,8 @@ from six.moves import cPickle as pickle
 from importlib import import_module
 from time import time
 from weakref import WeakKeyDictionary
-from email.utils import mktime_tz, parsedate_tz
+from email.utils import formatdate, mktime_tz, parsedate_tz
+
 from w3lib.http import headers_raw_to_dict, headers_dict_to_raw
 from scrapy.http import Headers, Response
 from scrapy.responsetypes import responsetypes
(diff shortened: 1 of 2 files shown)


redapple:
@kmike , were you thinking of something like this?
(Name of header `'X-Scrapy-Cache-Timestamp'` is subject to discussion of course)

kmike:
I don't like adding it to headers because 

1. response.headers won't be the same as sent by a remote website;
2. remote website is able to spoof this header.

response.meta looks like a better place, it can be harder to implement though.

redapple:
Headers looked much easier indeed.
I'm not sure I get your concern about point 2. What would be the reason for spoofing this header? With `'cached'` in response.flags + this header, could it be wrong?
I'll leave this up for anyone wanting to give a shot at using `Response.meta`.

kmike:
A good point about `cached` in response.flags

redapple:
@kmike , so you would go for https://github.com/scrapy/scrapy/issues/1877 ?

kmike:
@redapple Hm, yeah, I think it can be a solution.

AdrianAtZyte:
#8034

PR #2372, opened 2016-10-31
S3FilesStore enable SSL ( #2085 )

korsunowk:
#2085 Hey everyone. Added two variables in config, this should be it. Looking forward to any feedback!


scrapy/pipelines/files.py
```diff
@@ -85,6 +85,8 @@ class S3FilesStore(object):
 
     POLICY = 'private'  # Overriden from settings.FILES_STORE_S3_ACL in
                         # FilesPipeline.from_settings.
+    AWS_SECURE_CONNECTION = False
```
redapple: I think you can drop the `AWS_` prefix here, as it's more about S3/boto than AWS

scrapy/pipelines/images.py
```diff
@@ -46,6 +46,7 @@ class ImagesPipeline(FilesPipeline):
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
+    IMAGES_S3_SECURE_CONNECTION = False
```
redapple: I would prefer that no new upper-case class attribute is introduced.
It does not seem needed if `IMAGES_S3_SECURE_CONNECTION` setting is now defined.

scrapy/settings/default_settings.py
```diff
@@ -161,7 +161,7 @@
 }
 
 FILES_STORE_S3_ACL = 'private'
-FILES_S3_SECURE_CONNECTION = True
+files_s3_secure_connection = True
```
redapple: Sorry if my comments were confusing.
Settings should still be upper-case.

lopuhin:
Hey @korsunowk, thanks for the PR! Right now it covers only boto2 codepath, I think boto3 should also be handled (it is used on Python 3 and can be used with Python 2 too, and is better supported), more details in this comment https://github.com/scrapy/scrapy/pull/2085#issuecomment-248315497

scrapy/pipelines/images.py
```diff
@@ -46,7 +46,7 @@ class ImagesPipeline(FilesPipeline):
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
-    IMAGES_S3_SECURE_CONNECTION = False
+    images_s3_secure_connection = False
```
redapple: `ImagesPipeline.images_s3_secure_connection` does not seem used.

korsunowk: @redapple , hmm, yes, this is true, what should be done with images_s3_secure_connection for ImagesPipeline?

wRAR:
boto2 code was removed in #4776.

PR #2383, opened 2016-11-07
Replace semicolon by full stop

rahulkant13may:
Is it really a typo or you guys intentionally terminated this line with semicolon ?


kmike:
It was me who wrote this line, and I really terminated it with semicolon. I'm not a native speaker; in Russian list elements can be parts of a sentence and follow the same punctuation rules. 

http://english.stackexchange.com/questions/75591/semicolons-at-the-end-of-list-items suggests that this style is valid in English as well.

rahulkant13may:
Yes , you are right, we can use semicolon here.

Issue #892, opened 2014-09-20
Crawl-Delay support for robots.txt


kmike:
[Crawl-Delay](http://en.wikipedia.org/wiki/Robots_exclusion_standard#Crawl-delay_directive) directive in robots.txt looks useful. If it is present the delay suggested there looks like a good way to adjust crawling rate. When crawling unknown domains it can be better than autothrottle.

A draft implementation: https://gist.github.com/kmike/76aca46cad18915b8695

Some notes:
- it may be useful to respect only Crawl-Delay but not User-Agent rules, so this middleware is not inherited from an existing middleware;
- Crawl-Delay maps nicely to Scrapy downloader slots; it is possible to implement this feature without per-request download delays (see https://github.com/scrapy/scrapy/issues/802) for which semantics are unclear;
- currently RobotsCrawlDelayMiddleware doesn't use User-Agent because UA set by Scrapy is quite arbitrary, but I'm not sure this is a right thing to do;
- instead of stdlib robotparser https://github.com/seomoz/reppy is used; robotparser can't parse robots.txt with Crawl-Delay properly.

diegovincent:
How's this issue going @kmike ? Is there a PR for it?

diegovincent:
Maybe we should reference the following:

1. https://github.com/scrapy/scrapy/pull/3796
2. https://github.com/scrapy/scrapy/issues/3969

kmike:
Thanks for the links @diegovincent.
There is no PR for this feature yet; contributions are welcome!

dannyeuu:
This will be a nice feature

diegovincent:
@dannyeuu @kmike  still nothing?

If that is the case, maybe @sergipastor, @sarafg11 or @psique want to take a look at this?

kmike:
No updates @diegovincent.

fkromer:
Scrapy uses [protego as robots.txt file parser under the hood as default](https://github.com/scrapy/scrapy/blob/82f25bc44acd2599115fa339967b436189eec9c1/scrapy/settings/default_settings.py#L259). According to the [protego readme example](https://github.com/scrapy/protego#usage) extracting the allowed crawl delay from robots.txt file is supported already. This implies that either the scrapy interface for robots.txt file parsers does not support crawl delay yet. Or the integration of protego into scrapy is buggy.

stixpjr:
[...]

PR #2385, opened 2016-11-08
[WIP] Custom robotparser

### LINKED ISSUE #2373: Wikipedia robots.txt raises exceptions
I'm scraping a page which in turn links to wikipedia.

But the wikipedia robots.txt is creating some errors/exceptions as below.

Python 2.7.12
Scrapy 1.2.1

```
2016-11-02 13:13:18 [scrapy] DEBUG: Crawled (200) <GET https://en.wikipedia.org/robots.txt> (referer: None)
2016-11-02 13:13:18 [py.warnings] WARNING: C:\Python27\lib\urllib.py:1303: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  return ''.join(map(quoter, s))

2016-11-02 13:13:18 [scrapy] ERROR: Error downloading <GET http://en.wikipedia.org/robots.txt>: u'\xd8'
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 587, in _runCallbacks
  File "C:\Python27\lib\site-packages\scrapy\downloadermiddlewares\robotstxt.py", line 97, in _parse_robots
    rp.parse(body.splitlines())
  File "C:\Python27\lib\robotparser.py", line 120, in parse
    entry.rulelines.append(RuleLine(line[1], False))
  File "C:\Python27\lib\robotparser.py", line 174, in __init__
    self.path = urllib.quote(path)
  File "C:\Python27\lib\urllib.py", line 1303, in quote
    return ''.join(map(quoter, s))
KeyError: u'\xd8'
```

redapple:
Attempt to fix #2373

At first, I tried to subclass `RobotFileParser` and ended up subclassing quite a few classes and methods.
So I opted for an adapted copy of Python 3's `robotparser.py` (namely this version https://raw.githubusercontent.com/python/cpython/4106e4237df5397efc785ce4daa257aecb5bff75/Lib/urllib/robotparser.py),
using `w3lib.url.canonicalize_url` before compare paths.

(Note that the borrowed code does parse crawl delays, but this PR does not do anything to solve #892)

I'm not sure about the license though. Does GNU GPLv2 apply here?
```
""" robotparser.py

    Copyright (C) 2000  Bastian Kleineidam

    You can choose between two licenses when using this package:
    1) GNU GPLv2
    2) PSF license for Python 2.2

    The robots.txt Exclusion Protocol is implemented as specified in
    http://www.robotstxt.org/norobots-rfc.txt
"""
```

### DIFF
--- a/scrapy/utils/robotparser.py
+++ b/scrapy/utils/robotparser.py
@@ -1,0 +1,254 @@
+# this code is adapted from Python 3.x master branch
+""" robotparser.py
+
+    Copyright (C) 2000  Bastian Kleineidam
+
+    You can choose between two licenses when using this package:
+    1) GNU GPLv2
+    2) PSF license for Python 2.2
+
+    The robots.txt Exclusion Protocol is implemented as specified in
+    http://www.robotstxt.org/norobots-rfc.txt
+
+    Changes by Scrapy developers:
+    - Adapted for dual-Python2/Python3 support
+"""
+
+import collections
+try:
+    from urllib.parse import urlparse, urlunparse, quote, unquote
+except ImportError:
+    from urlparse import urlparse, urlunparse
+    from urllib import quote, unquote
+
+from w3lib.url import safe_url_string, canonicalize_url
+
+
+__all__ = ["RobotFileParser"]
+
+class RobotFileParser:
+    """ This class provides a set of methods to read, parse and answer
+    questions about a single robots.txt file.
+
+    """
+
+    def __init__(self, url=''):
+        self.entries = []
+        self.default_entry = None
+        self.disallow_all = False
+        self.allow_all = False
+        self.set_url(url)
+        self.last_checked = 0
+
+    def mtime(self):
+        """Returns the time the robots.txt file was last fetched.
+
+        This is useful for long-running web spiders that need to
+        check for new robots.txt files periodically.
+
+        """
+        return self.last_checked
+
+    def modified(self):
+        """Sets the time the robots.txt file was last fetched to the
+        current time.
+
+        """
+        import time
+        self.last_checked = time.time()
+
+    def set_url(self, url):
+        """Sets the URL referring to a robots.txt file."""
+        self.url = url
+        self.host, self.path = urlparse(url)[1:3]
+
+    def read(self):
+        """Reads the robots.txt URL and feeds it to the parser."""
+        raw = f.read()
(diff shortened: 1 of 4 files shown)


kmike:
> I'm not sure about the license though. Does GNU GPLv2 apply here?

@redapple I think users of this file are free to choose either GPL or PSF license; choosing PSF license is fine.

The code looks small enough 👍  How hard would it be to separate robotparser and our scrapy-specific fixes? It may be hard to keep packages in sync. I can imagine creating e.g. backports.robotparser package, which has your fixes for Python 2 compatibility, but not canonicalize_url bits. Or we can make robotparser2 or something like that, which uses w3lib.

scrapy/utils/robotparser.py
```diff
@@ -0,0 +1,254 @@
...
+        self.last_checked = time.time()
+
+    def set_url(self, url):
+        """Sets the URL referring to a robots.txt file."""
+        self.url = url
+        self.host, self.path = urlparse(url)[1:3]
+
+    def read(self):
+        """Reads the robots.txt URL and feeds it to the parser."""
+        raw = f.read()
```
redapple: oh right. sorry about that. Yes, it's a leftover I haven't removed completely (it's the part that downloads robots.txt from URLs)

redapple:
@kmike , thanks for the feedback .
This parser is not scrapy-specific but `canonicalize_url` works around the root problem of Python 2's `unquote` producing unicode strings with "raw" bytes that it cannot quote back (quoting sort of acts as the normalization/canonicalization)
```
>>> from urllib import unquote, quote
>>> unquote(u'http://site.local/wiki/K%C3%A4ytt%C3%A4j%C3%A4')
u'http://site.local/wiki/K\xc3\xa4ytt\xc3\xa4j\xc3\xa4'
>>> quote(unquote(u'http://site.local/wiki/K%C3%A4ytt%C3%A4j%C3%A4'))
/usr/lib/python2.7/urllib.py:1299: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  return ''.join(map(quoter, s))
  File "/usr/lib/python2.7/urllib.py", line 1299, in quote
    return ''.join(map(quoter, s))
KeyError: u'\xc3'
>>> 
```

but I think a w3lib-free implementation is doable. I'll work on this.

redapple:
Closing in favor of #2388

PR #2386, opened 2016-11-08
[MRG+1] Ignore explicitly compiled python files.

rolando:
This avoids to include compiled files in the templates directory.

Without this change, `pyc` files are included in the build:
```
$ find build | grep pyc
build/lib/scrapy/templates/project/module/__init__.pyc
build/lib/scrapy/templates/project/module/spiders/__init__.pyc
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,3 +12,4 @@ prune docs/build
 recursive-include extras *
 recursive-include bin *
 recursive-include tests *
+global-exclude __pycache__ *.py[cod]


kmike:
@rolando could you please explain why are pyc files a problem?

rolando:
@kmike IMO, it leaves the build dirty as only templates directory have `pyc` files. But, particularly this was breaking the conda-forge build on windows: https://github.com/conda-forge/staged-recipes/pull/1646#issuecomment-249061645

The conda build scripts cleaned up pyc files, but not in the templates directory. So when attempting to recreate pyc files it failed because templates already had pyc files around.

kmike:
I am cautious because not having pyc files in install directory means Python will try to create them on first run, and it can be a problem if filesystem permissions don't allow that. Package installation could happen with write access, but user may have only read access. 

But apparently setuptools (or is it distutils? or pip?) creates them during the install. Could you please confirm that manifest changes don't prevent pyc file creation during installation?

rolando:
@kmike The Manifest.in is only for `sdist` command: https://docs.python.org/2/distutils/sourcedist.html#the-manifest-in-template

The `sdist` output already does not include any `pyc` file except those two listed above, which usually don't cause a problem because the installers  (or binary packaging tools) compile all `pyc` files for the given platform.

This only aims to have a clean source distribution without any `pyc` file (those two files listed above).

kmike:
Ok, thanks for clarifying it!

PR #2001, opened 2016-05-19
[MRG+1] Fixed choosing of response class.

matveinazaruk:
Hey scrapy team!

This fix is connected to [this](https://github.com/scrapy/scrapy/issues/2000) issue. 
Little changes performed will fix detecting of response type. Seems that passing `body` into `responsetypes.from_args(...)` was missed. 

It caused a little incomprehension when I expected text type, and it should be text, but I had a simple `Response`, not `TextResponse`. 

Thanks, 
Matvei.

PR #2393, opened 2016-11-10
[MRG] Use body to choose response type after decompression content

### LINKED ISSUE #2145: Disabling RedirectMiddleware results in HttpCompressionMiddleware errors
I wanted not to redirect `303` responses, but instead retry them.
From the docs, I thought I could achieve it through two settings:

```
REDIRECT_ENABLED = False
RETRY_HTTP_CODES = [301, 302, 307, 308, 500, 502, 503, 504, 408]
```

It ended up giving me errors on `HttpCompressionMiddleware`:

```
  File "twisted/internet/defer.py", line 1128, in _inlineCallbacks
    result = g.send(result)
  File "scrapy/core/downloader/middleware.py", line 53, in process_response
    spider=spider)
  File "scrapy/downloadermiddlewares/httpcompression.py", line 38, in process_response
  File "scrapy/http/response/text.py", line 50, in replace
    return Response.replace(self, *args, **kwargs)
  File "scrapy/http/response/__init__.py", line 77, in replace
    return cls(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'encoding'
```

redapple:
Fixes #2145

### DIFF
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -38,7 +38,7 @@ class HttpCompressionMiddleware(object):
                 encoding = content_encoding.pop()
                 decoded_body = self._decode(response.body, encoding.lower())
                 respcls = responsetypes.from_args(headers=response.headers, \
-                    url=response.url)
+                    url=response.url, body=decoded_body)
                 kwargs = dict(cls=respcls, body=decoded_body)
                 if issubclass(respcls, TextResponse):
                     # force recalculating the encoding until we make sure the
--- a/tests/test_downloadermiddleware_httpcompression.py
+++ b/tests/test_downloadermiddleware_httpcompression.py
@@ -7,6 +7,7 @@ from scrapy.spiders import Spider
 from scrapy.http import Response, Request, HtmlResponse
 from scrapy.downloadermiddlewares.httpcompression import HttpCompressionMiddleware, \
     ACCEPTED_ENCODINGS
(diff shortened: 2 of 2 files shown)


kmike:
It looks similar to https://github.com/scrapy/scrapy/pull/2001.

This is quite inconsistent in Scrapy: e.g. cache or ftp handler still won't use body to get response class.

I guess a part of the problem is that ResponseTypes.from_body is not based on any specification, so there was no strong desire to use body consistently - not using body was not seen as a bug because from_body may look like a hack. 

It seems the most up-to-date document is https://mimesniff.spec.whatwg.org, especially https://mimesniff.spec.whatwg.org/#identifying-a-resource-with-an-unknown-mime-type.

While this change looks fine, the fix is not complete, and it looks like a part of a larger problem.

redapple:
@kmike , what do you mean by the fix not being complete?
Is it because `responsetypes.from_args()` does not follow whatwg?
Or because other uses of `from_args()` do not use the body either?

The aim here was to fix an issue at http decompression where, something else than the default `Response` type if there are hints in the body.

kmike:
@redapple the patch looks good because it improves response handling in decompression middleware, so I'm fine with merging it after a rebase. 

The logic middleware uses to detect response type is still different from what browsers do, and Scrapy is inconsistent in mime sniffing it performs. I should have opened another ticket for that, but I found it while reviewing this PR, so I wrote it in a comment :)

redapple:
Maybe #2145 is so rare we don't need to care. (I have never seen it myself.)

PR #2400, opened 2016-11-18
[MRG+1] CookiesMiddleware: keep cookies from 'Cookie' request header, fix encoding

### LINKED ISSUE #1992: DEFAULT_REQUEST_HEADERS not work as expected
I am new in scrapy, and I meet some problems which I can not get answer from google, so I post it here:

1 Cookie not work even set in DEFAULT_REQUEST_HEADERS:

```
DEFAULT_REQUEST_HEADERS = {
    'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
    'accept-encoding': 'gzip, deflate, sdch',
    'cache-control': 'no-cache',
    'cookie': 'xx=yy',
    'user-agent': 'Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.94 Safari/537.36'
}
```

```
        return scrapy.http.Request(url, headers=DEFAULT_REQUEST_HEADERS)
```

I know the `make_requests_from_url` will only called once for the start_urls, and in my opinion, the first request will send the cookie I set in the `DEFAULT_REQUEST_HEADERS`, however it does not.

2 Share settings between spiders.

I have multiple spiders in the project which share most of the settings like `RandomAgentMiddleware` `RandomProxyMiddleware` `UserAgent` `DEFAULT_REQUEST_HEADERS` and etc, however they are configured inside the settings.py for each spider.

Is it possible to share these settings?

---

The 
`COOKIES_ENABLED` is set to true.

### LINKED ISSUE #3575: Double-encoded cookies
When cookies are passed as UTF8 encoded bytes to the `Request` constructor, they end up being encoded twice and escaped in the `Cookie` header.

```
$ scrapy shell
(...)
In [1]: fetch(scrapy.Request('https://httpbin.org/cookies', cookies={'a': u'á'.encode('utf8')}))

In [2]: request.headers['Cookie']
Out[2]: b"a=b'\\xc3\\xa1'"

In [3]: print(response.text)
{
  "cookies": {
    "a": "b'\\xc3\\xa1'"
  }
}
```

This seems to happen only in Python 3.
```
$ scrapy version -v
Scrapy       : 1.5.0
lxml         : 4.2.6.0
libxml2      : 2.9.8
cssselect    : 1.0.3
parsel       : 1.5.1
w3lib        : 1.19.0
Twisted      : 18.9.0
Python       : 3.6.0 (default, Sep  1 2017, 10:59:37) - [GCC 4.8.4]
pyOpenSSL    : 18.0.0 (OpenSSL 1.1.0j  20 Nov 2018)
cryptography : 2.4.2
Platform     : Linux-4.4.0-134-generic-x86_64-with-debian-jessie-sid
```

elacuesta:
Fixes #1992.

This PR modifies the `CookiesMiddleware._get_request_cookies` method to prevent the middleware from discarding cookies already set in the headers (either by the `DefaultHeadersMiddleware` or by using the `Request`'s `headers` constructor parameter).

Update: also fixes #3575

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -202,6 +202,11 @@ CookiesMiddleware
    sends them back on subsequent requests (from that spider), just like web
    browsers do.
 
+   .. caution:: When non-UTF8 encoded byte sequences are passed to a
+      :class:`~scrapy.http.Request`, the ``CookiesMiddleware`` will log
+      a warning. Refer to :ref:`topics-logging-advanced-customization`
+      to customize the logging behaviour.
+
 The following settings can be used to configure the cookie middleware:
 
 * :setting:`COOKIES_ENABLED`
--- a/scrapy/downloadermiddlewares/cookies.py
+++ b/scrapy/downloadermiddlewares/cookies.py
@@ -29,8 +29,7 @@ class CookiesMiddleware:
 
         cookiejarkey = request.meta.get("cookiejar")
         jar = self.jars[cookiejarkey]
-        cookies = self._get_request_cookies(jar, request)
-        for cookie in cookies:
+        for cookie in self._get_request_cookies(jar, request):
             jar.set_cookie_if_ok(cookie, request)
 
         # set Cookie header
@@ -68,28 +67,65 @@ class CookiesMiddleware:
                 msg = "Received cookies from: {}\n{}".format(response, cookies)
                 logger.debug(msg, extra={'spider': spider})
 
-    def _format_cookie(self, cookie):
-        # build cookie string
-        cookie_str = '%s=%s' % (cookie['name'], cookie['value'])
+    def _format_cookie(self, cookie, request):
+        """
+        Given a dict consisting of cookie components, return its string representation.
+        Decode from bytes if necessary.
+        """
+        decoded = {}
+        for key in ("name", "value", "path", "domain"):
+            if not cookie.get(key):
+                if key in ("name", "value"):
+                    msg = "Invalid cookie found in request {}: {} ('{}' is missing)"
+                    logger.warning(msg.format(request, cookie, key))
+                    return
+                continue
+            if isinstance(cookie[key], str):
+                decoded[key] = cookie[key]
+            else:
+                try:
+                    decoded[key] = cookie[key].decode("utf8")
+                except UnicodeDecodeError:
+                    logger.warning("Non UTF-8 encoded cookie found in request %s: %s",
+                                   request, cookie)
+                    decoded[key] = cookie[key].decode("latin1", errors="replace")
 
-        if cookie.get('path', None):
-            cookie_str += '; Path=%s' % cookie['path']
-        if cookie.get('domain', None):
-            cookie_str += '; Domain=%s' % cookie['domain']
-
+        cookie_str = "{}={}".format(decoded.pop("name"), decoded.pop("value"))
+        for key, value in decoded.items():  # path, domain
+            cookie_str += "; {}={}".format(key.capitalize(), value)
         return cookie_str
 
     def _get_request_cookies(self, jar, request):
-        if isinstance(request.cookies, dict):
-            cookie_list = [
-                {'name': k, 'value': v}
-                for k, v in request.cookies.items()
-            ]
-        else:
-            cookie_list = request.cookies
+        """
+        Extract cookies from a Request. Values from the `Request.cookies` attribute
+        take precedence over values from the `Cookie` request header.
+        """
+        def get_cookies_from_header(jar, request):
+            cookie_header = request.headers.get("Cookie")
+            if not cookie_header:
+                return []
+            cookie_gen_bytes = (s.strip() for s in cookie_header.split(b";"))
+            cookie_list_unicode = []
+            for cookie_bytes in cookie_gen_bytes:
+                try:
+                    cookie_unicode = cookie_bytes.decode("utf8")
+                except UnicodeDecodeError:
+                    logger.warning("Non UTF-8 encoded cookie found in request %s: %s",
+                                   request, cookie_bytes)
+                    cookie_unicode = cookie_bytes.decode("latin1", errors="replace")
+                cookie_list_unicode.append(cookie_unicode)
+            response = Response(request.url, headers={"Set-Cookie": cookie_list_unicode})
+            return jar.make_cookies(response, request)
 
-        cookies = [self._format_cookie(x) for x in cookie_list]
-        headers = {'Set-Cookie': cookies}
-        response = Response(request.url, headers=headers)
+        def get_cookies_from_attribute(jar, request):
+            if not request.cookies:
+                return []
+            elif isinstance(request.cookies, dict):
+                cookies = ({"name": k, "value": v} for k, v in request.cookies.items())
+            else:
+                cookies = request.cookies
+            formatted = filter(None, (self._format_cookie(c, request) for c in cookies))
+            response = Response(request.url, headers={"Set-Cookie": formatted})
+            return jar.make_cookies(response, request)
 
-        return jar.make_cookies(response, request)
+        return get_cookies_from_header(jar, request) + get_cookies_from_attribute(jar, request)
--- a/tests/test_downloadermiddleware_cookies.py
+++ b/tests/test_downloadermiddleware_cookies.py
@@ -1,20 +1,21 @@
-import re
 import logging
-from unittest import TestCase
 from testfixtures import LogCapture
+from unittest import TestCase
 
+from scrapy.downloadermiddlewares.cookies import CookiesMiddleware
+from scrapy.downloadermiddlewares.defaultheaders import DefaultHeadersMiddleware
+from scrapy.exceptions import NotConfigured
 from scrapy.http import Response, Request
 from scrapy.spiders import Spider
+from scrapy.utils.python import to_bytes
 from scrapy.utils.test import get_crawler
-from scrapy.exceptions import NotConfigured
-from scrapy.downloadermiddlewares.cookies import CookiesMiddleware
 
 
 class CookiesMiddlewareTest(TestCase):
 
     def assertCookieValEqual(self, first, second, msg=None):
         def split_cookies(cookies):
-            return sorted(re.split(r";\s*", cookies.decode("latin1")))
+            return sorted([s.strip() for s in to_bytes(cookies).split(b";")])
         return self.assertEqual(split_cookies(first), split_cookies(second), msg=msg)
 
     def setUp(self):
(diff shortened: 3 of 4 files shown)


scrapy/downloadermiddlewares/cookies.py
```diff
@@ -81,6 +82,14 @@ def _format_cookie(self, cookie):
         return cookie_str
 
     def _get_request_cookies(self, jar, request):
+        # from 'Cookie' request header
+        cookie_header = request.headers.get('Cookie', '')
+        cookie_list = re.split(';\s*', cookie_header.decode('latin1'))
```
kmike: Are there stdlib functions to parse cookies?

Also, cookies should be UTF-8 in most cases (but other encodings are also possible, so the code must be robust); .decode('latin1') never fails, but I wonder if non-ascii UTF-8 cookies are preserved correctly now. Could you please add a test for it?

tests/test_downloadermiddleware_cookies.py
```diff
@@ -229,3 +229,17 @@ def test_local_domain(self):
         assert self.mw.process_request(request, self.spider) is None
         self.assertIn('Cookie', request.headers)
         self.assertEqual(b'currencyCookie=USD', request.headers['Cookie'])
+
+    def test_keep_cookie_header(self):
```
kmike: could you please also add a test which checks that it works with DEFAULT_HEADERS_MIDDLEWARE?

elacuesta: Amended commit, test added 👍

scrapy/downloadermiddlewares/cookies.py
```diff
@@ -71,7 +72,10 @@ def _debug_set_cookie(self, response, spider):
 
     def _format_cookie(self, cookie):
         # build cookie string
-        cookie_str = '%s=%s' % (cookie['name'], cookie['value'])
```
elacuesta: This is the part I'm least certain about. The tests were failing for py3 because the `b` char before a `bytes` object was being appended to the string, i.e.: `'%s=%s' % ('e', b'\xc3\xaa')`returned `"e=b'\\xc3\\xaa'"`, making the test comparison fail.
@kmike What do you think about it?

elacuesta: It also works with `cookie_str = to_bytes('%s=%s' % (cookie['name'], to_native_str(cookie['value'])))` (see https://github.com/elacuesta/scrapy/commit/9a74d4257dea86fb72271edfd68ba750e9a2b995)

elacuesta:
@kmike, about
> cookies should be UTF-8 in most cases (but other encodings are also possible, so the code must be robust)

I believe the current `CookieJar` implementation (`scrapy.http.cookies.CookieJar`, based on `http.cookiejar`) does not handle encodings other than UTF8 properly. A different implementation could solve that, but IMHO that would fall out of the scope of this PR. 
Check the following snippet:

```
# -*- coding: utf-8 -*-

from __future__ import print_function
from scrapy.http import Request, Response
from scrapy.http.cookies import CookieJar

def test_cookie(encoding):
    print('# Encoding: {} #'.format(encoding))
    jar = CookieJar()
    response = Response('http://example.org', headers={'Set-Cookie': u'a=ä'.encode(encoding)})
    request = Request('http://example.org')
    cookies_from_response = jar.make_cookies(response, request)
    print('cookies:', cookies_from_response)
    for cookie in cookies_from_response:
        jar.set_cookie(cookie)
    jar.add_cookie_header(request)
    print('Cookie header:', request.headers['Cookie'])
    print()

test_cookie('utf8')
test_cookie('latin1')
test_cookie('utf16')
```
and the output:
```
$ python2 cookies_jar.py 
# Encoding: utf8 #
cookies: [Cookie(version=0, name='a', value='\xc3\xa4', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: a=ä

# Encoding: latin1 #
cookies: [Cookie(version=0, name='a', value='\xe4', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: a=

# Encoding: utf16 #
cookies: [Cookie(version=0, name='\xff\xfea\x00', value='\x00\xe4\x00', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: ��a=�

$ python3 cookies_jar.py 
# Encoding: utf8 #
cookies: [Cookie(version=0, name='a', value='ä', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: b'a=\xc3\xa4'

# Encoding: latin1 #
cookies: [Cookie(version=0, name='a', value='�', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: b'a=\xef\xbf\xbd'

# Encoding: utf16 #
cookies: [Cookie(version=0, name='��a\x00', value='\x00�\x00', port=None, port_specified=False, domain='example.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=False, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]
Cookie header: b'\xef\xbf\xbd\xef\xbf\xbda\x00=\x00\xef\xbf\xbd\x00'
```

kmike:
@elacuesta yeah, it can't handle them because the encoding is unknown. But it shouldn't raise an exception even if a header is not utf-8 - there are byte sequences which can raise UnicodeDecodeError when decoded from utf-8; the whole response shouldn't fail in this case.

elacuesta:
Right, what do you think about the following then?
1) mention in the docs that cookies should be either unicode strings or UTF8-encoded bytes
2) catch `UnicodeDecodeError` when parsing cookies in `CookiesMiddleware` and log a warning
3) discard/keep (with probably bogus value) non-UTF8 cookies

kmike:
> mention in the docs that cookies should be either unicode strings or UTF8-encoded bytes

It won't work - it is website who defines what are cookies, we can't prescribe that. Non-utf8 cookies are likely bugs though.

> catch UnicodeDecodeError when parsing cookies in CookiesMiddleware and log a warning

This sounds good to me.

Warnings can be annoying if one scrapes a website which uses such cookies consistently, so it'd be nice to document how to disable these warnings, e.g. by pointing to https://doc.scrapy.org/en/latest/topics/logging.html#advanced-customization.

> discard/keep (with probably bogus value) non-UTF8 cookies

Currently non-UTF8 cookies are kept with bogus values - `errors='replace'` is used everywhere. Though not ideal, I haven't heard anyone complaining; probably it doesn't happen often, or maybe it happens in cases where non-lossy cookie values are not important.

elacuesta:
@kmike It seems to me that `http.cookiejar` works consistently in py2/py3 only with unicode strings, I added some code to decode cookies from the `Cookie` header (which is always `bytes` thanks to the `Headers` implementation) using `utf8` and falling back to `latin1` in case of `UnicodeDecodeError`.

With this implementation the `Cookie` header would always end up encoded with `utf8` even if the cookies were already passed as bytes using a different encoding. What do you think about that?

elacuesta:
I made some changes to the tests and to the docs to point to the logging configuration page.

@kmike could you please re-check this? The current status is:

Cookies set in the `Request.cookies` attribute
1) UTF8-encoded bytes: sent as they are, #3575 is also fixed
2) Non-UTF8-encoded bytes: sent as UTF8-encoded bytes - a warning is logged
3) Unicode string: encoded with UTF8

Cookies set in the `Cookie` request header (currently being ignored)
1) UTF8-encoded bytes: sent as they are
2) Non-UTF8-encoded bytes: sent as UTF8-encoded bytes - a warning is logged
3) Unicode string: encoded with UTF8

elacuesta:
Added a check and a test case to filter out invalid cookies (dicts with no `name` or `value` in the `Request.cookies` attribute)

scrapy/downloadermiddlewares/cookies.py
```diff
@@ -69,26 +69,60 @@ def _debug_set_cookie(self, response, spider):
...
-        response = Response(request.url, headers=headers)
-
-        return jar.make_cookies(response, request)
+        """
+        Extract cookies from a Request. Values from the `Request.cookies` attribute
+        take precedence over values from the `Cookie` request header.
+        """
+        def get_cookies_from_header(jar, request):
+            cookie_header = request.headers.get('Cookie') or b''
+            cookie_list_bytes = re.split(b';\s*', cookie_header)
```
dangra: `_get_request_cookies` is on every request path, considering that we should precompile regexes and exit earlier if `Cookie` header is not set.

elacuesta: Makes a lot of sense, thanks.

elacuesta: @dangra Done, thanks for the suggestion!

docs/topics/downloader-middleware.rst
```diff
@@ -200,6 +200,11 @@ CookiesMiddleware
    send them back on subsequent requests (from that spider), just like web
    browsers do.
 
+   .. caution:: When non-UTF8 encoded byte sequences are passed to a ``Request``,
```
Gallaecio: :lipstick: ``` :class:`~scrapy.http.Request` ```

scrapy/downloadermiddlewares/cookies.py
```diff
@@ -17,6 +18,7 @@ class CookiesMiddleware(object):
     def __init__(self, debug=False):
         self.jars = defaultdict(CookieJar)
         self.debug = debug
+        self.re_cookie_header = re.compile(b';\s*')
```
Gallaecio:
```suggestion
        self.re_cookie_header = re.compile(br';\s*')
```

Otherwise, I think it will result in a warning due to `\s`.

I also wonder if it would make sense to add `\s*` before the semicolon as well.

elacuesta:
Since we're only splitting by one char (";") and ignoring whitespaces, it seems like using plain `bytes.split` and `bytes.strip` is faster than using regex.

```python
In [1]: import re
   ...:
   ...: re_split = re.compile(br'\s*;\s*')
   ...:
   ...: def cookies(n=4):
   ...:     values = [b'%i=%i' % (i, i*10) for i in range(10**n)]
   ...:     return b' ; '.join(values)
   ...:
   ...: %timeit re_split.split(cookies())
   ...: %timeit [c.strip() for c in cookies().split(b';')]
7.58 ms ± 39.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
4.5 ms ± 42.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

docs/topics/downloader-middleware.rst
elacuesta: 👌

scrapy/downloadermiddlewares/cookies.py
elacuesta: Changed the approach, not using `re` anymore, `bytes.split` and `bytes.strip` are more efficient in this case.

PR #2401, opened 2016-11-20
Fix rules not working on first url

sleighsoft:
For the first urls in `start_urls` the rule will not be applied. This is an issue that seems to be around since 2012.
The proposed changes will make the rules work on the `start_urls`.

See here for the issue from 2012: http://stackoverflow.com/questions/12736257/why-dont-my-scrapy-crawlspider-rules-work

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -42,8 +42,13 @@ class CrawlSpider(Spider):
     def parse(self, response):
         return self._parse_response(response, self.parse_start_url, cb_kwargs={}, follow=True)
 
+    def _parse_links(self, response):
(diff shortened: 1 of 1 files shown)


sleighsoft:
The test failed due to `Address already in use` that is in no way related to my changes. Is it possible to rerun it?

redapple:
hi @sleighsoft ,
I do not understand the issue.
Can you share an example `CrawlSpider` subclass and crawl logs showing the issue?

sleighsoft:
I had an issue, where the URLs extracted from start_urls did not trigger the attached callback in the rule.
My changes fixed that problem. Somehow though it does work for me now even without those changes.
I will therefore close this PR and investigate again :(

PR #2403, opened 2016-11-22
[MRG+1] Use Twisted modules for client and endpoints

dangra:
@nyov discovered in https://github.com/scrapy/scrapy/pull/1887#issuecomment-260047541 that Scrapy is using patched Twisted libs shipped under `scrapy.xlib.tx` always. I introduced this bug in https://github.com/scrapy/scrapy/commit/bb806fa0f8f77cba59506153e33cab96f91008ab#diff-e487b2562862226386f811e0659c70c0R6 so this PR is an attempt to address the issue.

### DIFF
--- a/scrapy/xlib/tx/__init__.py
+++ b/scrapy/xlib/tx/__init__.py
@@ -2,13 +2,15 @@ from scrapy import twisted_version
 if twisted_version > (13, 0, 0):
     from twisted.web import client
     from twisted.internet import endpoints
-if twisted_version >= (11, 1, 0):
+elif twisted_version >= (11, 1, 0):
     from . import client, endpoints
 else:
     from scrapy.exceptions import NotSupported
+
     class _Mocked(object):
         def __init__(self, *args, **kw):
             raise NotSupported('HTTP1.1 not supported')
+
     class _Mock(object):
         def __getattr__(self, name):
             return _Mocked


nyov:
The current behavior enables us to just strip the whole version check instead.
So after a quick think, I'm not sure if it's useful to fix it this way at this point.
`xlib.tx` is pretty much only used by the downloader, at least inside scrapy itself AFAIK.

Since the consideration is on bumping twisted dependency, it _may_ be easier to handle the current behavior (or make no difference at all).
Otherwise you probably also want to change the version number to `if twisted_version > (13, 1, 0):` since `xlib.tx` is actually at that level now (though I'm not sure if the differences in that version were important, still on it).

If you would wait a few days with merging this, I'm hoping to have some commits pushed to #1887 and we could see how best to handle this, after bumping dependencies?

I would say though that this should definitely be merged/backported into current and past releases which won't see `xlib.tx` stripped.

redapple:
Thanks @nyov . I'm ok with waiting for your commits to #1887.
A good point on backporting this to 1.0.x, 1.1.x (we haven't branched out 1.2.x yet)

nyov:
Okay I'm good with merging this (With the version at `13.1.0`)
But I'm also in favor of bumping dependency to `13.1` in #1887, right behind this and deprecating `xlib.tx`, which would make this a redundant change, except for backports.

redapple:
@nyov , with
> Okay I'm good with merging this (With the version at 13.1.0)

do you mean starting with this?:
```
if twisted_version > (13, 1, 0):
    from twisted.web import client
    ...
```

nyov:
> do you mean starting with this?

Yes, since `xlib.tx` version of `HTTPConnectionPool` code is equal to Twisted 13.1.0

redapple:
@dangra , what do you think of @nyov comment about using `if twisted_version > (13, 1, 0):` ?

dangra:
@redapple : it makes sense, I bumped the version to 13.1.0

redapple:
Right @kmike , touching Twisted intergration is never simple.
Although I'm for including this in 1.2.2, we have quite some tests around this and if there's backlash, we can revert and be better prepared for 1.3 and upcoming #1887 (btw, @dangra , I don't see any new commit)

redapple:
@kmike , another option is to 1) renamed current 1.3 milestone to 1.4, 2) release 1.2.2 without this #2403, 3) release a 1.3 only with this #2403 fix right after

kmike:
@redapple I think if we're releasing 1.3 right away we can incude #1887. 
+0 from me to release this PR as a part of 1.2.2, but I wanted to start a discussion.

redapple:
@dangra , with #1887 merged, this is not needed for "master" branch anymore, but we may want to "backport" to 1.2 and 1.1 branches. Is it worth it?

kmike:
I think we shouldn't backport it to 1.2 and 1.1 because it is possibly backwards incompatible

redapple:
Alright. So I'm closing this.

PR #2404, opened 2016-11-22
[MRG+1] Print more dependencies versions in "scrapy version" verbose output

redapple:


redapple:
In theory, scrapy requires w3lib>=1.15 so the try/except may not be needed.
(You never know what users may end up with though)

kmike:
What do you think about adding optional dependencies? boto / botocore / Pillow?

kmike:
I also think that's fine to have this feature in Scrapy 1.2.x

redapple:
Good idea about adding the optional dependencies too. Another PR maybe?

redapple:
> I also think that's fine to have this feature in Scrapy 1.2.x

Right, we can still do that. We haven't started 1.3.x

Issue #2240, opened 2016-09-14
Add FAQ entry on splitting items in spider middleware


redapple:
This is related to https://github.com/scrapy/scrapy/issues/1915 and the consensus that hacking on item pipeline to split items is not the way to go with current architecture,
and that spider middleware are a good solution for this use-case.

PR #2409, opened 2016-11-23
Docs: add FAQ entry about splitting items

elacuesta:
See #2240


elacuesta:
@redapple What do you think about this? Do you think a link in the spider middleware section would be good too?

PR #2410, opened 2016-11-24
[MRG+1] Transparently handle redirections in fetch and shell

### LINKED ISSUE #2290: Scrapy should follow redirects on scrapy shell by default
Scrapy shell doesn't follow redirects when you do `fetch('http://google.com')`, but it does if you do ` `fetch(scrapy.Request('http://google.com'))`.

I realize this is a historic behavior, but I'd argue that it breaks the expectations for most users, since the most common expectation is that Scrapy would try to do what a browser would do, so I think we should change it.

In my opinion, not following the redirect should be the exceptional behavior, and so that's what should require the user to build a `scrapy.Request` object.

I think ideally, we would be able to pass a keyword argument to `fetch`, like:

```
fetch('http://google.com', redirect=False)
```

And only for that case it would set handle_httpstatus_all to True as explained by @redapple here: https://github.com/scrapy/scrapy/issues/2177#issuecomment-239430163

redapple:
Fixes #2290

I added `--no-redirect` option to `scrapy fetch` and `scrapy shell` to revert to previous behavior.

To do:

- [x] add tests for `scrapy shell` changes
- [x] document the new options and arguments

### DIFF
--- a/docs/topics/commands.rst
+++ b/docs/topics/commands.rst
@@ -382,7 +382,9 @@ Supported options:
 
 * ``-c code``: evaluate the code in the shell, print the result and exit
 
-* ``--no-redirect``: do not follow HTTP 3xx redirects (default is to follow them)
+* ``--no-redirect``: do not follow HTTP 3xx redirects (default is to follow them);
+  this only affects the URL you may pass as argument on the command line;
+  once you are inside the shell, ``fetch(url)`` will still follow HTTP redirects by default.
 
 Usage example::
 
@@ -397,6 +399,7 @@ Usage example::
     (200, 'http://example.com/')
 
     # you can disable this with --no-redirect
+    # (only for the URL passed as command line argument)
     $ scrapy shell --no-redirect --nolog http://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F -c '(response.status, response.url)'
     (302, 'http://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F')
 
--- a/docs/topics/commands.rst
+++ b/docs/topics/commands.rst
@@ -322,6 +322,14 @@ So this command can be used to "see" how your spider would fetch a certain page.
 If used outside a project, no particular per-spider behaviour would be applied
 and it will just use the default Scrapy downloader settings.
 
+Supported options:
+
+* ``--spider=SPIDER``: bypass spider autodetection and force use of specific spider
+
+* ``--headers``: print the response's HTTP headers instead of the response's body
+
+* ``--no-redirect``: do not follow HTTP 3xx redirects (default is to follow them)
+
 Usage examples::
 
     $ scrapy fetch --nolog http://www.example.com/some/page.html
@@ -368,10 +376,30 @@ given. Also supports UNIX-style local file paths, either relative with
 ``./`` or ``../`` prefixes or absolute file paths.
 See :ref:`topics-shell` for more info.
 
+Supported options:
+
+* ``--spider=SPIDER``: bypass spider autodetection and force use of specific spider
+
+* ``-c code``: evaluate the code in the shell, print the result and exit
+
+* ``--no-redirect``: do not follow HTTP 3xx redirects (default is to follow them)
+
 Usage example::
 
     $ scrapy shell http://www.example.com/some/page.html
     [ ... scrapy shell starts ... ]
+
+    $ scrapy shell --nolog http://www.example.com/ -c '(response.status, response.url)'
+    (200, 'http://www.example.com/')
+
+    # shell follows HTTP redirects by default
+    $ scrapy shell --nolog http://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F -c '(response.status, response.url)'
+    (200, 'http://example.com/')
+
+    # you can disable this with --no-redirect
+    $ scrapy shell --no-redirect --nolog http://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F -c '(response.status, response.url)'
+    (302, 'http://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F')
+
 
 .. command:: parse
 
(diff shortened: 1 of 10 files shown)


redapple:
Another side-effect of removing `handle_httpstatus_all` by default is that 404s in shell do not generate a response object:

As-before:

```
$ scrapy shell --no-status-aware https://docs.scrapy.org/somebadpage.html
2016-11-24 14:35:21 [scrapy] INFO: Scrapy 1.2.1 started (bot: scrapybot)
(...)
2016-11-24 14:35:21 [scrapy] INFO: Spider opened
2016-11-24 14:35:22 [scrapy] DEBUG: Crawled (404) <GET https://docs.scrapy.org/somebadpage.html> (referer: None)
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7f12cfe20ad0>
[s]   item       {}
[s]   request    <GET https://docs.scrapy.org/somebadpage.html>
[s]   response   <404 https://docs.scrapy.org/somebadpage.html>
[s]   settings   <scrapy.settings.Settings object at 0x7f12cfe209d0>
[s]   spider     <DefaultSpider 'default' at 0x7f12cf9f3d90>
[s] Useful shortcuts:
[s]   fetch(req_or_url) Fetch request (or URL) and update local objects
>>> response
<404 https://docs.scrapy.org/somebadpage.html>
>>> 
```

With this patch:
```
$ scrapy shell https://docs.scrapy.org/somebadpage.html
2016-11-24 14:35:35 [scrapy] INFO: Scrapy 1.2.1 started (bot: scrapybot)
(...)
2016-11-24 14:35:35 [scrapy] INFO: Spider opened
2016-11-24 14:35:35 [scrapy] DEBUG: Crawled (404) <GET https://docs.scrapy.org/somebadpage.html> (referer: None)
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7effb04ebad0>
[s]   item       {}
[s]   request    <GET https://docs.scrapy.org/somebadpage.html>
[s]   settings   <scrapy.settings.Settings object at 0x7effb04eb9d0>
[s] Useful shortcuts:
[s]   fetch(req_or_url) Fetch request (or URL) and update local objects
>>> response
>>> 
```

redapple:
So an alternative can be to only focus on 3xx response, still solving the original issue about redirects (maybe using something like what I outlined in https://github.com/scrapy/scrapy/pull/2193#discussion_r88246581, i.e. using [`handle_httpstatus_list`](https://docs.scrapy.org/en/latest/topics/spider-middleware.html#std:reqmeta-handle_httpstatus_list) with objects allowing tests on HTTP status classes)

redapple:
Alright, so I change the implementation to only handle HTTP 3XXs by default, with a new option for `scrapy fetch` and `scrapy shell` called `--no-redirect`.
Within the shell, `fetch(url, redirect=False)` will not follow redirects.

docs/topics/commands.rst
```diff
@@ -368,11 +376,31 @@ given. Also supports UNIX-style local file paths, either relative with
 ``./`` or ``../`` prefixes or absolute file paths.
 See :ref:`topics-shell` for more info.
 
+Supported options:
+
+* ``--spider=SPIDER``: bypass spider autodetection and force use of specific spider
+
+* ``-c code``: evaluate the code in the shell, print the result and exit
+
+* ``--no-redirect``: do not follow HTTP 3xx redirects (default is to follow them)
```
kmike: Should we mention that the redirect chain won't be followed for the initial request - this option doesn't change `fetch` behavior?

scrapy/shell.py
```diff
@@ -98,14 +100,16 @@ def _open_spider(self, request, spider):
...
         if isinstance(request_or_url, Request):
             request = request_or_url
-            url = request.url
         else:
             url = any_to_uri(request_or_url)
-            request = Request(url, dont_filter=True)
-            request.meta['handle_httpstatus_all'] = True
+            request = Request(url, dont_filter=True, **kwargs)
+            if redirect:
+                request.meta['handle_httpstatus_list'] = SequenceExclude(range(300, 400))
```
kmike: This looks like an useful pattern which can be applied to user code as well 👍 
Why isn't it working with navite `range` objects?

scrapy/utils/datatypes.py
```diff
@@ -304,3 +304,13 @@ def __setitem__(self, key, value):
         while len(self) >= self.limit:
             self.popitem(last=False)
         super(LocalCache, self).__setitem__(key, value)
+
+
+class SequenceExclude(object):
```
redapple: Will do. Thanks.

docs/topics/commands.rst
redapple: I don't understand this point. What's with the initial request?

scrapy/shell.py
redapple: I believe it does work with normal range. I'll add that to the tests for https://github.com/scrapy/scrapy/pull/2410#discussion_r91807062

docs/topics/commands.rst
kmike: I mean that user may send requests from scrapy shell using `fetch` function, and `--no-redirect` doesn't change `fetch` behavior.

redapple: Ah, ok. Yes, --no-redirect only affects the redirect for "url" in `scrapy shell --no-redirect url`. Once, Inside the shell, `>>> fetch(url)` does follow redirects (with this patch applied).
Definitely worth mentioning. I haven't found good wording yet (I already struggled writing this comment)

kmike: Sorry, my original comment was plain wrong, I can see that now :)

scrapy/utils/datatypes.py
redapple: Code updated.

scrapy/shell.py
redapple: Done.

redapple:
@eliasdorneles , @elacuesta , what do you think of this patch?

redapple:
As a side node, I think we can reuse this pattern for https://github.com/scrapy/scrapy/issues/2004 + https://github.com/scrapy/scrapy/pull/2193

eliasdorneles:
Looks great, nice work! :+1: :+1:

PR #915, opened 2014-10-08
Adds parsing filename from HTTP header.

orian:
When downloading a file from URL, the HTTP server may return a name of file the content should be saved as.

PR #2415, opened 2016-11-26
Enable FilesPipeline to determine filename from 'Content-Disposition' header.

syucream:
Closes #915

`file_path()` usually doesn't return a human readable filename so I'd like to enable to change it based on `Content-Disposition` header value.

In case of a server responds the header, `media_to_download()` can't determine an actual filename. So this behavior may be changeable by settings ... ?


jtboing:
Is this still an issue?

Gallaecio:
I think this is conceptually a valid feature request, but @ahlinc’s points are spot-on and should be taken into account. I’m not closing this pull request in case someone wants to take it over at some point.

AdrianAtZyte:
#7909

PR #2417, opened 2016-11-29
[MRG+1] Add docs for some scheduler settings

### LINKED ISSUE #2414: SCHEDULER settings not documented
The FAQ mentions:
SCHEDULER_DISK_QUEUE and SCHEDULER_MEMORY_QUEUE with regards to changing between FIFO/LIFO.

Neither of these appears to be documention except for this single reference. They're not on the settings page.

Also, the link to wikipedia for "LIFO" is actually to a disambiguation page, which doesn't help if you don't know what it is in the first place. :-) (I know it's Last In First Out, but others may not).

pawelmhm:
Fixes #2414

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -317,6 +317,6 @@ I'm scraping a XML document and my XPath selector doesn't return any items
 You may need to remove namespaces. See :ref:`removing-namespaces`.
 
 .. _user agents: https://en.wikipedia.org/wiki/User_agent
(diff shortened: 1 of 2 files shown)


redapple:
Thanks @pawelmhm !

kmike:
This improves on what we have now, so +1 to merge it.
I don't like though that we're documenting available classes, not the API they should implement, but having public API for these components is a separate issue.

PR #2418, opened 2016-11-30
[MRG+1] Fix #396 re-triggered issue

### LINKED ISSUE #396: inspect_response(response) yields incorrect response in IPython shell
Example case (requires registration at example site, and even then would be hard to use as a use-case; modify to suit your needs): http://pastebin.com/GT8N893q

In the above example, the response.meta printout in after_submit callback does not match that within the inspect_response shell on the second iteration (the first is correct).  It appears that inspect_response has a stale response the second time.

ahlinc:
The InteractiveShellEmbed class is a singleton
and we need to drop the instance with its clear_instance() method
to rebuild the instance from scratch with fresh environment
for each subsequent Scrapy's shell drop in.

### DIFF
--- a/scrapy/utils/console.py
+++ b/scrapy/utils/console.py
@@ -15,6 +15,9 @@ def _embed_ipython_shell(namespace={}, banner=''):
         config = load_default_config()
         # Always use .instace() to ensure _instance propagation to all parents
         # this is needed for <TAB> completion works well for new imports
+        # and clear the instance to always have the fresh env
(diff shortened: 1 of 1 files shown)


redapple:
(you can discard the appveyor build error)

eliasdorneles:
Looks good to me, thanks @ahlinc @redapple @Tarliton !

Tarliton:
thank you all for solving this so quickly!

Issue #2430, opened 2016-12-07
`scrapy version` fails if `SPIDER_MODULES` points to non-existent module


EgoWumpus:
When executing `scrapy version` on the command line inside a scrapy project, an `ImportError: No module named 'spiders'` can occur if the `settings.py` file that is found by scrapy contains, in this case `SPIDER_MODULES = ['spiders']` **and** there is no `spiders` module in the project.

Essentially, any execution of scrapy picks up the settings file, which will report a rather generic error (googling for it will return a number of unrelated issues) if there is an issue in that file. 

I created a StackOverflow question about this that can be found here with a bit more detail:
http://stackoverflow.com/questions/41028605/importerror-no-module-named-spiders/41028628?noredirect=1#comment69264795_41028628

Sample traceback:

```
$ scrapy version
Traceback (most recent call last):
  File "/Users/nathanielford/virtualenvironments/crawler/bin/scrapy", line 11, in <module>
    sys.exit(execute())
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/cmdline.py", line 141, in execute
    cmd.crawler_process = CrawlerProcess(settings)
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/crawler.py", line 238, in __init__
    super(CrawlerProcess, self).__init__(settings)
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/crawler.py", line 129, in __init__
    self.spider_loader = _get_spider_loader(settings)
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/crawler.py", line 325, in _get_spider_loader
    return loader_cls.from_settings(settings.frozencopy())
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/spiderloader.py", line 33, in from_settings
    return cls(settings)
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/spiderloader.py", line 20, in __init__
    self._load_all_spiders()
  File "/Users/nathanielford/virtualenvironments/crawler/lib/python3.5/site-packages/scrapy/spiderloader.py", line 28, in _load_all_spiders
    for module in walk_modules(name):
[...]

PR #2433, opened 2016-12-08
[MRG+1] Warn user instead of failing for wrong SPIDER_MODULES setting

redapple:
Proposal fix for #2430

Example output:

```
$ scrapy version
/home/paul/src/scrapy/scrapy/spiderloader.py:36: RuntimeWarning: Could not load spiders from module 'version_wrong_spiders.spiderssss'; Check SPIDER_MODULES setting (exception: No module named spiderssss)
  warnings.warn(msg, RuntimeWarning)
Scrapy 1.2.2
```

### DIFF
--- a/tests/test_spiderloader/__init__.py
+++ b/tests/test_spiderloader/__init__.py
@@ -1,6 +1,7 @@
 import sys
 import os
 import shutil
+import warnings
 
 from zope.interface.verify import verifyObject
 from twisted.trial import unittest
@@ -89,3 +90,14 @@ class SpiderLoaderTest(unittest.TestCase):
         crawler = runner.create_crawler('spider1')
         self.assertTrue(issubclass(crawler.spidercls, scrapy.Spider))
         self.assertEqual(crawler.spidercls.name, 'spider1')
+
+    def test_bad_spider_modules_warning(self):
+
+        with warnings.catch_warnings(record=True) as w:
+            module = 'tests.test_spiderloader.test_spiders.doesnotexist'
+            settings = Settings({'SPIDER_MODULES': [module]})
+            spider_loader = SpiderLoader.from_settings(settings)
+            self.assertIn("Could not load spiders from module", str(w[0].message))
+
+            spiders = spider_loader.list()
+            self.assertEqual(spiders, [])
--- a/scrapy/spiderloader.py
+++ b/scrapy/spiderloader.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+import traceback
 import warnings
 
 from zope.interface import implementer
@@ -30,9 +31,9 @@ class SpiderLoader(object):
                 for module in walk_modules(name):
                     self._load_spiders(module)
             except ImportError as e:
-                msg = ("Could not load spiders from module '{}'; "
-                       "Check SPIDER_MODULES setting "
-                       "(exception: {})".format(name, str(e)))
+                msg = ("\n{tb}Could not load spiders from module '{modname}'. "
+                       "Check SPIDER_MODULES setting".format(
+                            modname=name, tb=traceback.format_exc()))
                 warnings.warn(msg, RuntimeWarning)
 
     @classmethod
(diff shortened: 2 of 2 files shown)


tests/test_spiderloader/__init__.py
```diff
@@ -89,3 +90,14 @@ def test_crawler_runner_loading(self):
...
         self.assertTrue(issubclass(crawler.spidercls, scrapy.Spider))
         self.assertEqual(crawler.spidercls.name, 'spider1')
+
+    def test_bad_spider_modules_warning(self):
+
+        with warnings.catch_warnings(record=True) as w:
+            module = 'tests.test_spiderloader.test_spiders.doesnotexist'
+            settings = Settings({'SPIDER_MODULES': [module]})
+            spider_loader = SpiderLoader.from_settings(settings)
+            self.assertIn("Could not load spiders from module", str(w[0].message))
```
kmike: I think that's not good to hide exception type and a traceback.

redapple: @kmike , so you mean that the exception type and traceback should also appear in the warning?
or something else?

kmike: Yeah, I think this information should be available somehow, because otherwise it could make it harder to debug than before.

You also said that the root cause is that scrapy loads everything before running commands like `scrapy version`; it makes sense because users can add their own commands and you need to load them all before executing a command. The approach in this PR looks good to me - if discovery fails it is fine to still allow running built-in commands.

redapple: Updated. This is the kind of warning you get now:

```
$ scrapy version
/home/paul/src/scrapy/scrapy/spiderloader.py:37: RuntimeWarning: 
  File "/home/paul/src/scrapy/scrapy/spiderloader.py", line 31, in _load_all_spiders
    for module in walk_modules(name):
  File "/home/paul/src/scrapy/scrapy/utils/misc.py", line 63, in walk_modules
    mod = import_module(path)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named spiderssss
Could not load spiders from module 'version_wrong_spiders.spiderssss'. Check SPIDER_MODULES setting
  warnings.warn(msg, RuntimeWarning)
Scrapy 1.2.2
```

kmike: Hm, I see that in #2430 the error doesn't have a full traceback pointing to the module being imported (traceback is unrolled to the importlib and then to scrapy code); here it is the same. Could you please check that there is no a good traceback in this case even in Python 3? Is it the only information available for this ImportError? Sorry for being so picky!

kmike: Thanks for explanation @redapple!

vshlapakov:
@redapple @kmike I recently found this PR and realised that now Scrapy swallows import errors not only for `scrapy version` but for other commands like `scrapy list`. The problem is that we rely on `scrapy list` results as a part of deploy process: earlier if someone forgot to add some necessary dependency - the deploy was failed at once. Now the exception is swallowed and `scrapy list` returns 0 code with some stderr, as a result platform thinks that everything is fine and there're just 0 spiders.

Of course it's possible to catch `RuntimeWarning` when running `scrapy-list`, but it's more about guessing as client code can also raise same exceptions and parsing stderr is not a reliable way. For now we started to always pass stderr logs if any, but I wonder if there's an option to add some switch between exception/warning behaviours (via Scrapy setting for example). 

Please share your thoughts, I can start a separate issue to track it.

redapple:
I think it makes sense to "swallow" the error for commands that do not need to load spiders code.
And it can be limited to these commands only indeed.
This patch here is probably too liberal and I can see how a command-specific setting could be used to only warn or fail louder.

vshlapakov:
@redapple Agree. We have a similar setting in [sh-scrapy](https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/blob/master/sh_scrapy/settings.py#L130) to decide if we should fail on import errors, or just print the error/warning, something like this should here as well. So I'd propose to limit this feature to the commands that really need to load spiders code + use a setting to change its behaviour. In other words I don't mind against current behaviour but we need to be able to configure it.

kmike:
Yeah, this all makes sense. For the record: there are users who want `scrapy list` and `scrapy crawl` to succeed even if some of the spiders are broken - see e.g. https://github.com/scrapy/scrapy/issues/2518#issuecomment-276058188

PR #2434, opened 2016-12-08
[logging] log in colors

pawelmhm:
This adds logging in colors. I still need to add tests for this but wanted to share for early feedback, let me know what you think. I used colorama (for different colors depending on log message) and pygments (for syntax highlighting of item fields). Sample output looks like this 
color_log
colors_error

Let me know what you think, I'll add tests and docs bit later.


redapple:
+1 to adding colored logs.
There was a internal discussion with @kmike and @eliasdorneles where this commit was referenced, where qutebrowser reverted the use of `colorlog`
https://github.com/The-Compiler/qutebrowser/commit/c64e5c9bd595ef310825d07bf29e1d3eb7674714

kmike:
+1 to colors in terminal, it looks like an awesome feature.

I'm not sure about highlighting items using pygments: 

1. Shouldn't it use repr, not str? Custom `__str__` of custom Item subclass may return human-readable representation of an item which is not necessarily parseable as Python. `__repr__` could do the same, but Python docs encourage to make `__repr__` unambiguous and Python-like. 
2. Won't it make scraping slower, probably much slower? How fast is highlighting for large items?
3. If (1) or (2) is an issue for user, how to turn pygments off, without turning highlighting off?

Because of these questions I'd start without pygments; it should make PR easier to merge and less controversial.

A nice feature would be to highlight different parts of log messages - e.g. dates could be more dim, as well as logger names; maybe they will look better with fixed colors (i.e. not red/green depending on log level). 🚲 ⛺️ detected.

How does highlighting look like on a white terminal background? Are colors still ok? If not, how to solve that - provide several themes, at least for white and dark backgrounds? What about other themes with fancy colors, i.e. not just red/green/cyan from 80s?

scrapy/settings/default_settings.py
```diff
@@ -187,6 +187,11 @@
 LOG_ENCODING = 'utf-8'
 LOG_FORMATTER = 'scrapy.logformatter.LogFormatter'
 LOG_FORMAT = '%(asctime)s [%(name)s] %(levelname)s: %(message)s'
+COLOR_LOG = True
+
+if COLOR_LOG:
+    LOG_FORMAT = '%(log_color)s%(asctime)s [%(name)s] %(levelname)s: %(message)s'
```
kmike: this condition is always true, even if user overrided the setting

kmike:
It'd be also nice to detect cases where colored log doesn't make any sense (e.g. when redirecting scrapy output to a file using either shell redirect `2>` or using scrapy option). Are some of these cases handled by colorlog library?

pawelmhm:
> Because of these questions I'd start without pygments; it should make PR easier to merge and less controversial.

yeah good idea. I removed pygments.

> It'd be also nice to detect cases where colored log doesn't make any sense (e.g. when redirecting scrapy output to a file using either shell redirect 2> or using scrapy option)

yes. this is already detected, when you pass `logfile` `configure_logging` in utils jumps to if filename block and never moves to if log_enabled and log_colors. Logging in color to file is not good, because ansii codes are printed to file and it looks unreadable and noisy if logfile reader doesn't support it.

> A nice feature would be to highlight different parts of log messages - e.g. dates could be more dim, as well as logger names; maybe they will look better with fixed colors (i.e. not red/green depending on log level). 

This is somewhat tricky part. There is question what exactly deserves its own colors (dates? HTTP status codes? different types of messages? item_scraped, item_dropped etc?). There is also difficult in finding color combinations that will look good. It may require some good designing skills and some form of visual sensitivity to colors that allows you to choose right color palette and it seems somewhat difficult to me. Maybe we could just provide some easy way to load different colorings via log_format string, so that users can conrtribute their own formattings?

> How does highlighting look like on a white terminal background? Are colors still ok? 

Need to check with white background, I'm not sure at the moment. 

> If not, how to solve that - provide several themes, at least for white and dark backgrounds? What about other themes with fancy colors, i.e. not just red/green/cyan from 80s?

Yeah ideally we could provide different colors, and maybe even allow users to contribute their own colors? Perhaps setting COLOR_LOG_FORMAT could control that, but i need to test that.

eLRuLL:
There is a ["related" question in SO](http://stackoverflow.com/questions/41077807/scrapy-logs-display-twice-after-install-coloredlogs) for enabling this with `coloredlogs`. It would be good if someone could help with that

pawelmhm:
@kmike @redapple I removed colorama dependency, it is apparently only needed for windows logging. I followed approach from [this commit](https://github.com/qutebrowser/qutebrowser/blob/c64e5c9bd595ef310825d07bf29e1d3eb7674714/qutebrowser/utils/log.py) and it works fine. Now only need to:


- [ ] test on windows - add some warning on windows
- [ ] add unit tests 

for tests I'll add one test to test_log_utils - testing logformatter object format() method, and probably also need some sort of integration test. If you have any suggestions that's nice

PR #2437, opened 2016-12-09
Update changelog for upcoming 1.3.0 release

redapple:


docs/news.rst
```diff
@@ -3,6 +3,40 @@
...
+-------------------------
+
+New Features
+~~~~~~~~~~~~
+
+- ``MailSender`` now accepts single strings as values for ``to`` and ``cc``
+  arguments (:issue:`2272`)
+- ``scrapy fetch url``, ``scrapy shell url`` and ``fetch(url)`` inside
+  scrapy shell now follow HTTP redirections by default (:issue:`2290`);
+  See :command:`fetch` and :command:`shell` for details.
```
kmike: I think LOG_SHORT_NAMES option can be seen as a new feature

kmike:
What do you think about mentioning the problem with scrapy.xlib.tx which lead us to make 1.3 release so soon? I.e. that previously (in 1.2) only backported code was used; now Scrapy imports from twisted.

redapple:
@kmike , good idea. Would you mind writing it up? This branch is under "scrapy" org

docs/news.rst
redapple: We can also move the whole "Logging" section points under "New features".

redapple:
@kmike , what do you think of 
```
$ git diff
diff --git a/docs/news.rst b/docs/news.rst
index 02976fe..09c47dc 100644
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -6,6 +6,14 @@ Release notes
 Scrapy 1.3.0 (2016-12-XX)
 -------------------------
 
+This release comes rather soon after 1.2.2 for one main reason:
+it was found out that releases since 0.18 up to 1.2.2 (included) use
+some backported code from Twisted, even if newer Twisted modules are available.
+Scrapy now uses ``twisted.web.client`` and ``twisted.internet.endpoints`` directly.
+
+As it is a major change, we wanted to get the bug fix out quickly
+while not breaking any projects using the 1.2 series.
+

```

kmike:
@redapple you've beaten me :) Looks perfect.

docs/news.rst
kmike: yeah, makes sense.

Issue #1964, opened 2016-05-02
twisted error stack trace doesn't output when the spider is running in scrapyd


ksimple:
The outputted log is something like the following text and no stack trace which makes diagnostic super hard.
2016-05-02 20:25:02 [twisted] CRITICAL: Unhandled error in Deferred:
2016-05-02 20:25:02 [twisted] CRITICAL:

I created a from_crawler method on spider and it runs successfully on local machine but failed on scrapyd. Finally I notice that the signature of from_crawler on scrapyd should be from_crawler(cls, crawler, **kwargs) since there is a _job parameter. This is the first hard thing to discover since the reference doesn't mention it.

But the diagnostic of this one is toooooooooooo hard since there is no stack trace in twisted log. I have to write my own log observer and use addObeserve to log it by myself and find the first issue.

So please check why the stack trace doesn't output to log. I downloaded the code but can't figure out by just reading the code, thanks.

redapple:
@ksimple , a few questions,
- what version of scrapy are you using locally?
- what version of scrapyd are you running?
- do you have a simple/minimal spider class reproducing the issue?

ksimple:
local Scrapy 1.0.5
remote Scrapy 1.2.0dev2 on docker image vimagick/scrapyd

spider
# -*- coding: utf-8 -*-
import scrapy


class ReproNologSpider(scrapy.Spider):
    name = "repro_nolog"
    allowed_domains = ["adhoc"]
    start_urls = (
        'http://www.google.com/',
    )

    @classmethod
    def from_crawler(cls, crawler):
        return cls(crawler = crawler)

    def parse(self, response):
        pas

log on scrapyd, the ktwisted is the logger I wrote to grab the lost log
2016-05-02 12:45:41 [scrapy] INFO: Scrapy 1.2.0dev2 started (bot: adhoc)
[...]

PR #2450, opened 2016-12-14
Replicate stdout and stderr to logfile

Digenis:
Context: https://github.com/scrapy/scrapy/issues/1964#issuecomment-216814406

When specifying `LOG_STDOUT = True`
the stdout shouldn't be redirected to the log
but replicated. Same for stderr.

I remember another issue
with exceptions in places such as spiders' `__init__` methods
that were missing their tracebacks in the log,
being logged only on stdout/stderr.
I recently discovered that it's a problem on twisted-15.5.0
but not 16.2.0.

The second commit can be ignored for now.
I was just testing something if the StreamLogger can be deprecated.

### DIFF
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -92,8 +92,9 @@ def configure_logging(settings=None, install_root_handler=True):
     if isinstance(settings, dict) or settings is None:
         settings = Settings(settings)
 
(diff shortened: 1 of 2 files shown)


Digenis:
Closing due to merge conflicts and loss of interest.
Patches released under scrapy's license.

PR #2456, opened 2016-12-19
[MRG+1] Feed exports: beautify JSON and XML

### LINKED ISSUE #1327: XMLGenerated xml looks ugly
Whenever i trying to get output items in xml i get something like:

``` xml
<?xml version="1.0" encoding="utf8"?>
<news><item><name>0</name><content>00000</content></item><item><name>1</name><content>11111</content></item><item><name>2</name><content>22222</content></item></news>
```

Maybe scrapy have some already provided ways to "beautify" it to something like (like xmls i found in all docs examples):

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<news>
   <item>
      <name>0</name>
      <content>00000</content>
   </item>
   <item>
      <name>1</name>
      <content>11111</content>
   </item>
   <item>
      <name>2</name>
      <content>22222</content>
   </item>
</news>
```

elacuesta:
Hello folks, these are my two cents to fix #1327.
I look forward to reading your comments 😄

### DIFF
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -140,7 +140,7 @@ output examples, which assume you're exporting these two items::
 BaseItemExporter
 ----------------
 
-.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8', indent_width=None)
+.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8', indent=None)
 
    This is the (abstract) base class for all Item Exporters. It provides
    support for common features used by all (concrete) Item Exporters, such as
@@ -149,7 +149,7 @@ BaseItemExporter
 
    These features can be configured through the constructor arguments which
    populate their respective instance attributes: :attr:`fields_to_export`,
-   :attr:`export_empty_fields`, :attr:`encoding`, :attr:`indent_width`.
+   :attr:`export_empty_fields`, :attr:`encoding`, :attr:`indent`.
 
    .. method:: export_item(item)
 
@@ -216,10 +216,14 @@ BaseItemExporter
       encoding). Other value types are passed unchanged to the specific
       serialization library.
 
-   .. attribute:: indent_width
+   .. attribute:: indent
 
-      Amount of spaces used to indent the output on each level.
-      Defaults to ``None``, which disables indentation.
+      Amount of spaces used to indent the output on each level. Defaults to ``None``,
+      which disables indentation. This argument behaves like ``indent`` in python's
+      JSON module (both for JSON and XML exporters): "If ``indent`` is a non-negative
+      integer, then array elements and object members will be pretty-printed with that
+      indent level. An indent level of 0, or negative, will only insert newlines.
+      ``None`` (the default) selects the most compact representation"
 
 .. highlight:: none
 
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -222,8 +222,8 @@ BaseItemExporter
 
       * ``indent=None`` selects the most compact representation,
         all items in the same line with no indentation
-      * ``indent<=0`` each item on it's own line, no indentation
-      * ``indent>0`` each item on it's own line, indentated with the provided numeric value
+      * ``indent<=0`` each item on its own line, no indentation
+      * ``indent>0`` each item on its own line, indented with the provided numeric value
 
 .. highlight:: none
 
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -140,7 +140,7 @@ output examples, which assume you're exporting these two items::
 BaseItemExporter
 ----------------
 
-.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8')
+.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8', indent_width=None)
 
    This is the (abstract) base class for all Item Exporters. It provides
    support for common features used by all (concrete) Item Exporters, such as
@@ -149,7 +149,7 @@ BaseItemExporter
 
    These features can be configured through the constructor arguments which
    populate their respective instance attributes: :attr:`fields_to_export`,
-   :attr:`export_empty_fields`, :attr:`encoding`.
+   :attr:`export_empty_fields`, :attr:`encoding`, :attr:`indent_width`.
 
    .. method:: export_item(item)
 
@@ -215,6 +215,11 @@ BaseItemExporter
       affects unicode values (which are always serialized to str using this
       encoding). Other value types are passed unchanged to the specific
       serialization library.
+
+   .. attribute:: indent_width
+
+      Amount of spaces used to indent the output on each level.
+      Defaults to ``None``, which disables indentation.
 
 .. highlight:: none
 
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -140,7 +140,7 @@ output examples, which assume you're exporting these two items::
 BaseItemExporter
 ----------------
 
-.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8', indent=None)
+.. class:: BaseItemExporter(fields_to_export=None, export_empty_fields=False, encoding='utf-8', indent=0)
 
    This is the (abstract) base class for all Item Exporters. It provides
    support for common features used by all (concrete) Item Exporters, such as
@@ -218,12 +218,12 @@ BaseItemExporter
 
    .. attribute:: indent
 
-      Amount of spaces used to indent the output on each level. Defaults to ``None``,
-      which disables indentation. This argument behaves like ``indent`` in python's
-      JSON module (both for JSON and XML exporters): "If ``indent`` is a non-negative
-      integer, then array elements and object members will be pretty-printed with that
-      indent level. An indent level of 0, or negative, will only insert newlines.
-      ``None`` (the default) selects the most compact representation"
+      Amount of spaces used to indent the output on each level. Defaults to ``0``.
+
+      * ``indent=None`` selects the most compact representation,
+        all items in the same line with no indentation
+      * ``indent<=0`` each item on it's own line, no indentation
+      * ``indent>0`` each item on it's own line, indentated with the provided numeric value
 
 .. highlight:: none
 
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -276,11 +276,12 @@ Default: ``0``
 
 Amount of spaces used to indent the output on each level. If ``FEED_EXPORT_INDENT``
 is a non-negative integer, then array elements and object members will be pretty-printed
-with that indent level. An indent level of ``0``, or negative, will put each item on a new line.
-``None`` selects the most compact representation
+with that indent level. An indent level of ``0`` (the default), or negative,
+will put each item on a new line. ``None`` selects the most compact representation.
 
 Currently implemented only by :class:`~scrapy.exporters.JsonItemExporter`
-and :class:`~scrapy.exporters.XmlItemExporter`
+and :class:`~scrapy.exporters.XmlItemExporter`, i.e. when you are exporting
+to ``.json`` or ``.xml``.
 
 .. setting:: FEED_STORE_EMPTY
 
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -209,7 +209,7 @@ These are the settings used for configuring the feed exports:
  * :setting:`FEED_STORE_EMPTY`
  * :setting:`FEED_EXPORT_ENCODING`
  * :setting:`FEED_EXPORT_FIELDS`
- * :setting:`FEED_EXPORT_INDENT_WIDTH`
+ * :setting:`FEED_EXPORT_INDENT`
 
 .. currentmodule:: scrapy.extensions.feedexport
 
@@ -267,15 +267,17 @@ If an exporter requires a fixed set of fields (this is the case for
 is empty or None, then Scrapy tries to infer field names from the
 exported data - currently it uses field names from the first item.
 
-.. setting:: FEED_EXPORT_INDENT_WIDTH
+.. setting:: FEED_EXPORT_INDENT
 
-FEED_EXPORT_INDENT_WIDTH
-------------------------
+FEED_EXPORT_INDENT
+------------------
 
 Default: ``None``
 
-Amount of spaces to indent on each level.
-Set to `None` to disable indentation.
+Amount of spaces used to indent the output on each level. If ``FEED_EXPORT_INDENT``
+is a non-negative integer, then array elements and object members will be pretty-printed
+with that indent level. An indent level of 0, or negative, will only insert newlines.
+``None`` (the default) selects the most compact representation
 
 Currently used by :class:`~scrapy.exporters.JsonItemExporter`
 and :class:`~scrapy.exporters.XmlItemExporter`
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -279,7 +279,7 @@ is a non-negative integer, then array elements and object members will be pretty
 with that indent level. An indent level of ``0``, or negative, will put each item on a new line.
 ``None`` selects the most compact representation
 
-Currently used by :class:`~scrapy.exporters.JsonItemExporter`
+Currently implemented only by :class:`~scrapy.exporters.JsonItemExporter`
 and :class:`~scrapy.exporters.XmlItemExporter`
 
 .. setting:: FEED_STORE_EMPTY
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -209,6 +209,7 @@ These are the settings used for configuring the feed exports:
  * :setting:`FEED_STORE_EMPTY`
  * :setting:`FEED_EXPORT_ENCODING`
  * :setting:`FEED_EXPORT_FIELDS`
+ * :setting:`FEED_EXPORT_INDENT_WIDTH`
 
 .. currentmodule:: scrapy.extensions.feedexport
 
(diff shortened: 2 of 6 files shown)


redapple:
I like the feature and implementation.
XML out definitely gets easier to read.
However, I'm not sure about having prettified-JSON as default. On the other hand, everyone uses `jq` these days right?
And it would be weird to prettify XML by default, and not JSON.
I'll have to sleep on it.

redapple:
@elacuesta , I believe it also needs tests for non-default (None, 2, etc.) widths.

elacuesta:
Hey @redapple, thanks for your comments. I added a few tests with distinct width values.
Regarding the default, I set it to `4` following @kmike's [comment](https://github.com/scrapy/scrapy/issues/1327#issuecomment-118660643) and my own desire, but I think it would be just fine to leave it optional.

elacuesta:
The build is failing, I'll check it tomorrow :-(

redapple:
Looks good to me @elacuesta

docs/topics/exporters.rst
```diff
@@ -216,6 +216,11 @@ BaseItemExporter
       encoding). Other value types are passed unchanged to the specific
       serialization library.
 
+   .. attribute:: indent_width
+
+      Amount of spaces used to indent the output on each level.
+      Defaults to ``None``, which disables indentation.
```
kmike: This argument controls newlines, not only indent spaces. I wonder if we should follow the same approach as stdlib json:

> If indent is a non-negative integer, then JSON array elements and object members will be pretty-printed with that indent level. An indent level of 0, or negative, will only insert newlines. None (the default) selects the most compact representation.

scrapy/exporters.py
```diff
@@ -128,33 +129,52 @@ def __init__(self, file, **kwargs):
             self.encoding = 'utf-8'
         self.xg = XMLGenerator(file, encoding=self.encoding)
 
+    def _beautify_newline(self):
+        if self.indent_width:
```
kmike: It seems the meaning on indent_width is different for json and xml: zero value means no newlines for XML, but for json new lines are still inserted, i.e. None and 0 are different for json.

scrapy/exporters.py
```diff
@@ -99,7 +100,7 @@ def __init__(self, file, **kwargs):
         self._configure(kwargs, dont_fail=True)
         self.file = file
         kwargs.setdefault('ensure_ascii', not self.encoding)
-        self.encoder = ScrapyJSONEncoder(**kwargs)
+        self.encoder = ScrapyJSONEncoder(indent=self.indent_width, **kwargs)
```
kmike: This looks backwards incompatible for people who passed `indent` argument for JsonItemExporter - I think it'll now raise a TypeError exception `got multiple values for keyword argument 'indent'`. For backwards compatibility it'd be great to handle this use case; `indent` in kwargs should take precedence over `indent_width`.

elacuesta: Nice! I think something like the following would work, I'll update it shortly.
```
indent = kwargs.pop('indent', self.indent_width)
self.encoder = ScrapyJSONEncoder(indent=indent, **kwargs)
```

kmike: you can also handle it the same way as ensure_ascii

elacuesta: That works too. It was so close and I didn't see it! :stuck_out_tongue:

scrapy/exporters.py
elacuesta: Actually, this current implementation is consistent for JSON and XML in the sense that `None==0`, because of this line: https://github.com/scrapy/scrapy/pull/2456/files#diff-b40c5755a5dda2b02aaac6bf4c8e2ff8R175.
However, I think you have a point and it would be even better to enforce the difference between `None` and `0`. I'm  working on it, I'll update this PR shortly 👍

elacuesta: The link is not pointing to the previous commit 😕 
Anyway, the line was `self.indent = settings.getint('FEED_EXPORT_INDENT') or None`, so `self.indent` ended up being `None` if `settings['FEED_EXPORT_INDENT']` was `0`

docs/topics/exporters.rst
elacuesta: Updated, now the code follows that behaviour for both JSON and XML 😄

tests/test_feedexport.py
```diff
@@ -443,7 +443,160 @@ def test_export_encoding(self):
...
+                'format': 'json',
+                'indent': 2,
+                'expected': b"""
+[
+{
+  "foo": [
+    "bar"
+  ]
+}
+]""",
```
kmike: hm, is this correct that here there is a newline in the beginning on the file instead of a newline at the end, and with indent=0 there are new lines both in the beginning and in the end? Wouldn't it be better not to have a new line in the beginning and always have a new line in the end (unless indent is None)?

elacuesta: That newline is there only to make the tests more pretty, it does not end up in the final expected result (note the `strip` call in https://github.com/scrapy/scrapy/pull/2456/files#diff-0f004a6e06393cc5e29012b83956eed2R632).
Up until now (before your comment), the square brackets in the JSON export had their own line (even with `indent=0`), this PR didn't modify that. But now that you mentioned it, I amended the latest commit to make it fully compliant with the difference between `indent=None` and `indent=<integer>`. The only newline I don't know how to remove when `indent=None` is the one after the XML declaration tag, but I don't think that's a problem.

redapple:
Looking at the new tests, I think this change reverts https://github.com/scrapy/scrapy/pull/1950 .

elacuesta:
You're right Paul. I personally like the square brackets having their own line even if `indent=None`, I made this latest change as per Mikhail's suggestion of following stdlib `JSONEncoder`'s behaviour.
I think diverting from the standard in some very particular cases (like this one, or the newline after the XML declaration tag) is not that big of a deal, I can make the brackets have their own line again if you guys agree. Thanks for your time 😀

kmike:
Previous way of exporting to JSON kind of makes sense; it seems it becomes unsupported after this PR: no indentation inside items, but each item is on a new line. It is not only a question about brackets. It can be seen as a backwards-incompatible change: `wc -l` will stop working as a way to count exported items in json output.

kmike:
What do you think about making `indent` option work only inside individual items in case of json, not for the whole output? That's be a bit weird though..

elacuesta:
I think we all agree that the current JSON output (one item per line) is just fine. However it's not consistent with the current XML output (all items in one line). What would you say if I modified this PR to make both outputs consistent in the following manner:
* `indent = None`: "most compact representation", i.e. all items in the same line (same as the current behaviour for XML)
* `indent <= 0`: each item on it's own line but no "internal" indent, square brackets on their own line for JSON (same as the current behaviour for JSON)
* `indent > 0`: indent with the specified number

kmike:
It sounds fine to me. Alternatively, we can support string constants for indent:

* "compact" (all items in the same line), 
* "lines" (the way scrapy currently works)
* numeric indent >= 0 (the same as in this PR)

elacuesta:
Updated according to https://github.com/scrapy/scrapy/pull/2456#issuecomment-284468472.
I also added support for string values (at the setting level, not in the `ItemExporter` classes), though I think the implementation might be a bit dirty. Looking forward to reading your comments.

kmike:
I think we should support either None/0 indent values or string values, it doesn't make sense to support both. I like string values because they look more extensible and explicit - they are easier to read, and it is easier to add another mode.

elacuesta:
I think string values should be syntactic sugar only for settings values, and the internal ItemExporter implementation should receive more consistent values. Personally, I would pick `None/Int` values over strings if we were to support only one, I feel i'ts kinda weird for a parameter to receive either an integer or a string.

kmike:
I think the problem is that there are in fact two different settings, cramped into one setting: 

1) how to format the result;
2) how many spaces to use for indentation.

Strings + integers is a way to introduce a single option instead of two options: string values activate (1), while int value activates (2). With two options there'd be 3 string values for the first option, and an int value would be active only when mode is "indent items". So that's where weirdness comes from - instead of explicit 3 possible string values + a separate option for indentation there is a single option with 2 string values which also accepts int.

For me negative and None values instead of strings feels like we're using negative and None values to choose (1) mode; it doesn't seem right - first, we won't be able to add other modes (it is not extendable), and second, why use magic number constants instead of readable string constants?

For me strings + int sounds ok, but maybe two explicit options is better.

redapple:
I've re-read the discussion this morning.
I now wonder if the compact mode for XML (current behavior, with all items on the same line) makes much sense for users. I do see the point of "lines" output though.
For XML output, I think having the "lines" mode by default (like for JSON) would be fine (but it's backward incompatible).
With that assumption, i.e. that all items on same line not being useful in practice, I would go for a setting saying "lines" (default) or "prettified/beautified" (e.g. `FEED_EXPORT_PRETTY=True/False`) and an indent int as another setting (`FEED_EXPORT_INDENT`).

kmike:
The difference between XML and JSON is that new lines are allowed inside XML tags, but in JSON keys or values, where new lines must be escaped. So you can guarantee "one item per line" for JSON, but not for XML. Another difference is that in XML whitespaces and new lines between tags are significant - you can access them in a parser, they affect parse result. In JSON indentation and new lines don't matter, parse result is the same. Just saying :) I'm not an user of XML exports myself.

redapple:
> new lines are allowed inside XML tags

Ok. This is already true with the current "all items on one line" XML (which is not really the case then).

> Another difference is that in XML whitespaces and new lines between tags are significant 

Ok. But I'm not sure it is a big problem for Scrapy's item exports and their consumers. I don't expect users/parsers of `items.xml` files to care about or choke on newlines between `<item></item>` tags. Again, this is debatable as with any format breaking change. Thoughts?

redapple:
@kmike @elacuesta ,
what do you think of what I commented earlier?
The gist being:

- change default XML output from "no newlines between `<items>` or `<item> elements`" to "one newline  after `<items>`, and one newline between consecutive `<item>`: that way, XML output is similar to current default JSON output with opening and closing brackets on their own line, and each JSON item on its own line too ; that would be the new "compact" default
- having a "set pretty items on" setting, that would be OFF by default (and maybe ON in the default project template)
- having a "indent value when prettifying"  setting to accomodate user's taste for pretty output

kmike:
Sounds good to me @redapple!

redapple:
Alright, I gave it a shot: https://github.com/elacuesta/scrapy/pull/2
Looking forward to your feedback @elacuesta and @kmike .

elacuesta:
Hello fellows, thank you for revisiting this PR, I'm sorry I didn't have the time to comment before.

I thought the whole point of separating the beautifying into two settings was to allow for multiple modes of indentation, beyond on/off. If we are not doing that anymore (it seems to me like on/off it's all we need) I'm not sure why we can't stick with the original implementation (following the standard `json` behavior).

Don't get me wrong, I really like Paul's approach. I think it's definitely the way to go in the sense that it easily allows the introduction of new modes, I just think that we might not need those modes and the code could be even simpler :-D

redapple:
@elacuesta ,
now that I check your earlier commits, and after sleeping on it a bit, it does indeed look like using `indent>0` for prettifying is sufficient. And simpler is better here: no real need for 2 settings. @kmike's [reference](https://github.com/scrapy/scrapy/pull/2456/commits/7e9153b38d7ef70f1f19a82506b669433d134b01#r102518994) to "indent" meaning for stdlib json does make a lot of sense:
> If indent is a non-negative integer, then JSON array elements and object members will be pretty-printed with that indent level. An indent level of 0, or negative, will only insert newlines. None (the default) selects the most compact representation.

But at the same time, I think this PR should still preserve current formatting of JSON with opening and trailing square brackets on their own lines and each item on their own line too.
For XML, I'll stick to my comment earlier that I would prefer to change the default XML serialization to have individual `<item>` elements starting on a new line.

I think I can summarize my updated view as:
- one setting is sufficient
- opening/closing brackets for JSON (resp. XML declaration and root `<items>` element and closing tag for XML output) should still be on their own lines: **this means a change for XML output vs. current behavior**;
- serialized items always start on a new line: **this is a change for XML output vs. current implementation;**
- this new "indent" setting affect only individual items and has the same meaning as stdlib json ;
- by default, items are serialized in the current compact form (yet, each starting on a new line, as stated above):
this makes the default serialization for JSON and XML human digestible in a text editor (you can roughly count the items, [except when there are newlines in XML data](https://github.com/scrapy/scrapy/pull/2456#issuecomment-287035426)): **this is a change for XML output vs. current implementation;**

How does that sound?
Sorry for the noise and changing of mind.
And thanks @elacuesta for being persistent!

elacuesta:
By all means, thanks to you :-)
I agree completely. I believe this PR, up until the third commit (https://github.com/scrapy/scrapy/pull/2456/commits/c7bb2fa8ce2633d92a7ec2840f84b174a5494428), covers all of that. Maybe we could just remove the fourth and last commit (https://github.com/scrapy/scrapy/pull/2456/commits/a630839d6d098fef27119de25843129f41bf0228), what do you guys think?

redapple:
I see that https://github.com/scrapy/scrapy/pull/2456/commits/c7bb2fa8ce2633d92a7ec2840f84b174a5494428#diff-0f004a6e06393cc5e29012b83956eed2R538 still has the "all on 1 line" XML

elacuesta:
Right, but that's only with `FEED_EXPORT_INDENT=None`, which is there for consistency with stdlib `json` (_"None (...) selects the most compact representation"_). The default value is `FEED_EXPORT_INDENT=0`, which causes each item to go on it's own line.

redapple:
Alright, so `FEED_EXPORT_INDENT=None` also serializes to JSON as `[{},{},{}]` without newlines?

elacuesta:
Yes, here's the test case: https://github.com/scrapy/scrapy/pull/2456/commits/c7bb2fa8ce2633d92a7ec2840f84b174a5494428#diff-0f004a6e06393cc5e29012b83956eed2R464

redapple:
Thanks @elacuesta .
I've tested https://github.com/scrapy/scrapy/compare/c7bb2fa8ce2633d92a7ec2840f84b174a5494428?expand=1 locally and it works as I would expect, for `None`, `0` or `>0`.
I think the only non-obvious thing when reviewing the change are the change in `test_export_no_items_store_empty`: you could even leave it unchanged with the default `FEED_EXPORT_INDENT=0`. You seem to cover enough for `None` with `test_export_indentation`.
Other than that, yeah, https://github.com/scrapy/scrapy/pull/2456/commits/c7bb2fa8ce2633d92a7ec2840f84b174a5494428 looks good to me.

redapple:
@kmike , what's your opinion on using only 1 setting as @elacuesta comments in https://github.com/scrapy/scrapy/pull/2456#issuecomment-294179629 ?

elacuesta:
Thanks for your commenta @redapple. I removed the 4th commit and left the branch at `c7bb2fa`.
About the changes in `test_export_no_items_store_empty`, if I remove the `FEED_EXPORT_INDENT=None` I would need to add a `\n` to the `xml` line, so there would still be a change in there. I think I prefer to explicitely set `FEED_EXPORT_INDENT=None` so we don't rely on defaults and know exactly what to expect, what do you think? Thanks again!

tests/test_feedexport.py
```diff
@@ -425,25 +425,190 @@ def test_export_encoding(self):
...
+     </item>
+</items>""",
+            },
+        ]
+
+        for row in test_cases:
+            settings = {'FEED_FORMAT': row['format'], 'FEED_EXPORT_INDENT': row['indent']}
+            data = yield self.exported_data(items, settings)
+            print(row['format'], row['indent'])
+            self.assertEqual(row['expected'].strip(), data)
```
kmike: Could you please write this test without .strip(), so that we're explicit about newlines before and after items?

elacuesta: Sure :+1:

docs/topics/exporters.rst
```diff
@@ -216,6 +216,15 @@ BaseItemExporter
       encoding). Other value types are passed unchanged to the specific
       serialization library.
 
+   .. attribute:: indent
+
+      Amount of spaces used to indent the output on each level. Defaults to ``0``.
+
+      * ``indent=None`` selects the most compact representation,
+        all items in the same line with no indentation
+      * ``indent<=0`` each item on it's own line, no indentation
```
kmike: its

docs/topics/exporters.rst
```diff
@@ -216,6 +216,15 @@ BaseItemExporter
...
+      * ``indent>0`` each item on it's own line, indentated with the provided numeric value
```
kmike: indented

docs/topics/feed-exports.rst
```diff
@@ -266,6 +267,21 @@ If an exporter requires a fixed set of fields (this is the case for
...
+
+Default: ``0``
+
+Amount of spaces used to indent the output on each level. If ``FEED_EXPORT_INDENT``
+is a non-negative integer, then array elements and object members will be pretty-printed
+with that indent level. An indent level of ``0``, or negative, will put each item on a new line.
+``None`` selects the most compact representation
+
+Currently used by :class:`~scrapy.exporters.JsonItemExporter`
+and :class:`~scrapy.exporters.XmlItemExporter`
```
kmike: I think that'd be nice to mention something like "i.e. when you're exporting to .json or .xml".

kmike:
Looks good to me, other than a few minor comments!

docs/topics/feed-exports.rst
elacuesta: That was the spirit of the following comment ("Currently used by..."), perhaps I could emphasize that _only_ those implement it for now.

kmike: I mean that for user it may be unclear what these JsonItemExporter and XmlItemExporter mean - user almost never instantiates them directly, for user these classes are implementation details of how json or xml exports are implemented. User doesn't have to know that these classes are used when `scrapy crawl foo -o result.json` is called - that's why  I suggested to mention that the option affects the format of `-o result.json` and `-o resut.xml` exports, not only some classes user may be not aware of.

redapple: @elacuesta , @kmike , I took the liberty to commit with your proposal @kmike : https://github.com/scrapy/scrapy/pull/2456/commits/3a0a86ed31df1d22fea3b5b05e853f212adc40c8

elacuesta: Thanks Paul. That's true, I'm sorry I missed it before.

redapple: Hey @elacuesta , don't be sorry! Your work is great

PR #2457, opened 2016-12-20
[MRG] Support kwargs for response.xpath()

redapple:
Allow passing named variables and `namespaces` dict arguments on `.xpath()` shortcut, now that https://github.com/scrapy/parsel/pull/45 is merged

### DIFF
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -283,6 +283,40 @@ XPath specification.
 
 .. _Location Paths: https://www.w3.org/TR/xpath#location-paths
 
+.. _topics-selectors-xpath-variables:
+
+Variables in XPath expressions
+------------------------------
+
+XPath allows you to reference variables in your XPath expressions, using
+the ``$somevariable`` syntax. This is somewhat similar to parameterized
+queries or prepared statements in the SQL world where you replace
+some arguments in your queries with placeholders like ``?``,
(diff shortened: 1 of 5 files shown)


kmike:
+1 to have it; it'd be nice to have docs for this feature right in Scrapy docs.

redapple:
@kmike , I updated the docs on Selectors with XPath variables.
Do you want to mention ad-hoc namespaces too?

kmike:
In setup.py we declare [parsel >= 0.9.5](https://github.com/scrapy/scrapy/blob/c305c46254e773be374218c95862bb497383aa25/setup.py#L52), but this feature requires parsel >= 1.1.0. Time to bump version in setup.py?

kmike:
No opinion about ad-hoc namespaces :)

redapple:
Is it strictly needed? For those not using kwargs, 0.9.5<=parsel<1.1 should work fine.
I thought setup.py was about the absolute minimum required version.
No strong opinion on that though, since parsel 1.1 is better.

kmike:
We have two options:

1) document that this feature requires a more recent parsel;
2) bump parsel version.

(2) looks like a no-brainer for me - parsel is a pure-Python package, and there are no breaking changes, while (1) requires user to think twice and check what's this 'parsel' thing and what is the installed version :)

kmike:
👍

PR #1626, opened 2015-12-04
[MRG+1] Doc improve log formating examples

Digenis:
When the formatting is done by the logger (`logger.info('%s', val)`)
and a formatting exception occurs, there are several problems:
- The exception & stacktrace is printed to stderr, not the log (#2450)
- The debugger is not invoked by twisted (#1551)
- Extensions don't see errors; Errors are not even counted in the stats.
- It's impossible to see the full stacktrace
  with frames from the caller of the logger

I demonstrated some of this in #1551.
This is why I prefer early formatting when logging.

Code updated like this is backward incompatible
because exceptions will occur early, in the caller of the log
stopping execution of any code that failed formatting.
(This is irrelevant to deprecation policy,
 I only update documentation examples)

Edit: I also updated another formatting example that could be simple.
Edit2: Build fails because of the twisted update

PR #2458, opened 2016-12-21
[MRG+1] ENH Pass arguments to logger rather than formatted message.

rolando:
This not only use the standard form but helps error aggregation
libraries (i.e.: Sentry) to avoid duplicating the message.


kmike:
I think this change is fine.

A a side not, see also: https://github.com/scrapy/scrapy/pull/1626

PR #2464, opened 2016-12-23
[MRG+2] Validate numeric values (components order)

### LINKED ISSUE #2420: String value for order of Scrapy component
If Scrapy component order is defined as a string, it leads to undefined behaviour on Python 2 and to the following errors on Python 3:
```
File "/usr/local/lib/python3.5/site-packages/scrapy/middleware.py", line 58, in from_crawler
 return cls.from_settings(crawler.settings, crawler)
File "/usr/local/lib/python3.5/site-packages/scrapy/middleware.py", line 29, in from_settings
 mwlist = cls._get_mwlist_from_settings(settings)
File "/usr/local/lib/python3.5/site-packages/scrapy/core/spidermw.py", line 21, in _get_mwlist_from_settings
 return build_component_list(settings.getwithbase('SPIDER_MIDDLEWARES'))
File "/usr/local/lib/python3.5/site-packages/scrapy/utils/conf.py", line 47, in build_component_list
 return [k for k, v in sorted(six.iteritems(compdict), key=itemgetter(1))]
builtins.TypeError: unorderable types: str() < int()
```

My guess that 1) order of a Scrapy component should be stated as of integer type (or `None`) and there should be a check somewhere, 2) or the sorting logic should be fixed.

elacuesta:
Fixes #2420

~Just as a side note, I am intentionally leaving out floats because different values would
be converted to the same `int` (for instance: `int(1.1) == int(1.2) == 1` is `True`)~

/cc @redapple, @vshlapakov

### DIFF
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import numbers
 from operator import itemgetter
 
 import six
@@ -34,6 +35,13 @@ def build_component_list(compdict, custom=None, convert=update_classpath):
             _check_components(compdict)
             return {convert(k): v for k, v in six.iteritems(compdict)}
 
+    def _validate_values(compdict):
+        """Fail if a value in the components dict is not a real number or None."""
+        for name, value in six.iteritems(compdict):
+            if value is not None and not isinstance(value, numbers.Real):
+                raise ValueError('Invalid value {} for component {}, please provide ' \
+                                 'a real number or None instead'.format(value, name))
+
     # BEGIN Backwards compatibility for old (base, custom) call signature
     if isinstance(custom, (list, tuple)):
         _check_components(custom)
@@ -43,6 +51,7 @@ def build_component_list(compdict, custom=None, convert=update_classpath):
         compdict.update(custom)
     # END Backwards compatibility
 
+    _validate_values(compdict)
     compdict = without_none_values(_map_keys(compdict))
     return [k for k, v in sorted(six.iteritems(compdict), key=itemgetter(1))]
 
--- a/tests/test_utils_conf.py
+++ b/tests/test_utils_conf.py
@@ -62,6 +62,27 @@ class BuildComponentListTest(unittest.TestCase):
         self.assertRaises(ValueError, build_component_list, duplicate_bs,
                           convert=lambda x: x.lower())
 
+    def test_valid_numbers(self):
+        # work well with None and numeric values
+        d = {'a': 10, 'b': None, 'c': 15, 'd': 5.0}
+        self.assertEqual(build_component_list(d, convert=lambda x: x),
+                         ['d', 'a', 'c'])
+        d = {'a': 33333333333333333333, 'b': 11111111111111111111, 'c': 22222222222222222222}
+        self.assertEqual(build_component_list(d, convert=lambda x: x),
+                         ['b', 'c', 'a'])
+        # raise exception for invalid values
+        d = {'one': '5'}
+        self.assertRaises(ValueError, build_component_list, {}, d, convert=lambda x: x)
+        d = {'one': '1.0'}
+        self.assertRaises(ValueError, build_component_list, {}, d, convert=lambda x: x)
+        d = {'one': [1, 2, 3]}
+        self.assertRaises(ValueError, build_component_list, {}, d, convert=lambda x: x)
+        d = {'one': {'a': 'a', 'b': 2}}
+        self.assertRaises(ValueError, build_component_list, {}, d, convert=lambda x: x)
+        d = {'one': 'lorem ipsum',}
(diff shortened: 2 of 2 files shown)


tests/test_utils_conf.py
```diff
@@ -62,6 +62,25 @@ def test_duplicate_components_in_basesettings(self):
         self.assertRaises(ValueError, build_component_list, duplicate_bs,
                           convert=lambda x: x.lower())
 
+    def test_ensure_integer(self):
+        # recognize integer values inside strings
+        d = {'one': '1', 'two': None, 'three': '\n8\n', 'four': ' 4 '}
+        self.assertEqual(build_component_list(d, convert=lambda x: x),
+                         ['one', 'four', 'three'])
+        # raise exception if there is a float
+        d = {'one': 1.5, 'two': 2}
```
kmike: what's wrong with float priorities?

kmike:
Did float priorities work before this change? They look useful.

elacuesta:
They worked, and now that I took a closer look the documentation doesn't actually say that the numbers should be integers. I'll change it to allow floats too, thanks for noticing @kmike!

elacuesta:
@redapple, @kmike The branch is still named `component_order_integer`, would you guys like me to close this PR and open a new one with a more suitable branch name?

redapple:
@elacuesta , that branch name is not an issue, don't worry

scrapy/utils/conf.py
```diff
@@ -34,6 +34,16 @@ def _map_keys(compdict):
             _check_components(compdict)
             return {convert(k): v for k, v in six.iteritems(compdict)}
 
+    def _ensure_numeric_value(compdict):
+        for name, value in six.iteritems(compdict):
+            if value is None or isinstance(value, (int, float)):
+                continue
+            try:
+                compdict[name] = float(value)
```
kmike: One more question: why is converting values to floats needed? Shouldn't we just raise an exception if the value is not int/float, what's the point of supporting non-numeric values?

scrapy/utils/conf.py
```diff
@@ -34,6 +34,16 @@ def _map_keys(compdict):
             _check_components(compdict)
             return {convert(k): v for k, v in six.iteritems(compdict)}
 
+    def _ensure_numeric_value(compdict):
+        for name, value in six.iteritems(compdict):
+            if value is None or isinstance(value, (int, float)):
```
kmike: A nitpick: this doesn't account for `long` in Python 2, and also it doesn't account for custom number types (e.g. numpy.int16). This should be a super-rare edge case, but maybe we should check `isinstance(value, numbers.Real)` instead (see https://docs.python.org/2/library/numbers.html)?

elacuesta: Sure!

scrapy/utils/conf.py
elacuesta: Allow to continue if there is a numeric value inside a string, like `" 4 "` or `\n8\n`. That would have played nice with the code referenced in #2420, but I don't really have a strong opinion about it.

kmike: While we will contibue supporting them in Python 2, order will change, as compared to scrapy 1.3. So it'd be backwards incompatible for users which for whatever reason relied on string priorities (why would anyone do that?). +1 from me to just raise an exception and require users to fix their code.

elacuesta: I'll drop support for strings then 😄
Out of curiosity, what do you mean by "order will change, as compared to scrapy 1.3"?

kmike: Hm, I mean that "10" > "2" is False, but 10 > 2 is True

elacuesta: I misunderstood your words, I thought you were talking about an upcoming change in the way of defining components order after scrapy 1.3 :-P

redapple:
Still looks good to me.
@dangra , @kmike , @vshlapakov, what do you think?

kmike:
This looks good! 

May I ask for the last change? :) `_ensure_numeric_value` name is a bit misleading; could we name it `_ensure_numeric_values` or `_raise_if_unsupported_values` or `_validate_values`, something like that? `value` -> `values` because we're working with dict values; `ensure` -> `validate / ...` because code no longer converts strings to numbers.

elacuesta:
@kmike Updated, I went with `_validate_numeric_values` :-)

vshlapakov:

kmike:
Hey @elacuesta! Sorry for being so picky. I'm not a native speaker, but for me `_validate_numeric_values` reads as "validate values if they are numeric", i.e. "check only numeric values, but not other values - e.g. ensure that they are positive", not "check that all values are numeric" :)

elacuesta:
Hahaha no problem @kmike :-)
Sure, the name wasn't really the best, I changed it to `_validate_values`

redapple:
Thank you @elacuesta !

PR #2466, opened 2016-12-24
[MRG+1] settings: fixing name of the pipeline template

eLRuLL:
`scrapy startproject myproject` creates a `pipelines.py` with an example Pipeline with already the name of the project, it only feels natural that this should also be the name on `settings.py` pipelines section.

### DIFF
--- a/scrapy/templates/project/module/settings.py.tmpl
+++ b/scrapy/templates/project/module/settings.py.tmpl
@@ -65,7 +65,7 @@ ROBOTSTXT_OBEY = True
 # Configure item pipelines
 # See http://scrapy.readthedocs.org/en/latest/topics/item-pipeline.html
 #ITEM_PIPELINES = {
-#    '$project_name.pipelines.SomePipeline': 300,
+#    '$project_name.pipelines.${ProjectName}Pipeline': 300,
 #}
 
 # Enable and configure the AutoThrottle extension (disabled by default)


kmike:
I don't have a strong opinion on that - on one hand, this adds some convenience when there is a single pipeline, but on the other hand creating single-purpose components with descriptive names is usually better.

redapple:
I think the project template needs to be consistent in naming, and the generated settings need to be coherent with the generated classes.
Currently, in `settings.py`, there's:
- `$project_name.middlewares.${ProjectName}SpiderMiddleware`, which matches the generated `class ${ProjectName}SpiderMiddleware(object)` in `middlewares.py`
- `$project_name.middlewares.MyCustomDownloaderMiddleware`, which is not in the template,
- `$project_name.pipelines.SomePipeline` **which does not match** the generated `class ${ProjectName}Pipeline(object)` in `pipelines.py`

So it's far from using descriptive component names (I agree with @kmike that well-named single-purpose components is good advice, but it may be something for the docs or via comments in the generated files?), nor is it coherent.
And there could also be a `class ${ProjectName}DownloaderMiddleware(object)` in `middlewares.py` even.

eLRuLL:
I agree with you @kmike but I think that isn't the case here. I am just changing the `SomePipeline` name that was already in the project without any real purpose. I would say it actually could confuse a beginner.

PR #2471, opened 2016-12-31
Allow image conversion/preservation in ImagePipeline

### LINKED ISSUE #1705: Image pipeline should allow to decide whether to auto-convert to jpg or not
Not 100% it's accurate, but the issue got reported here:

https://news.ycombinator.com/item?id=10941213

> Image pipeline converts to jpg automatically. Cannot be disabled without writing all the image code yourself

sambhav:
Fixes #1705

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -46,7 +46,6 @@ class ImagesPipeline(FilesPipeline):
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
-    DEFAULT_IMAGES_CONVERT_TYPE = 'JPEG'
 
     def __init__(self, store_uri, download_func=None, settings=None):
         super(ImagesPipeline, self).__init__(store_uri, settings=settings,
@@ -84,9 +83,7 @@ class ImagesPipeline(FilesPipeline):
         self.thumbs = settings.get(
             resolve('IMAGES_THUMBS'), self.THUMBS
         )
-        self.convert_type = settings.get(
-            resolve('IMAGES_CONVERT_TYPE'), self.DEFAULT_IMAGES_CONVERT_TYPE
-        )
+        self.convert_type = settings.get(resolve('IMAGES_CONVERT_TYPE'))
 
     @classmethod
     def from_settings(cls, settings):
@@ -162,6 +159,7 @@ class ImagesPipeline(FilesPipeline):
     def file_path(self, request, response=None, info=None):
         ## start of deprecation warning block (can be removed in the future)
         import warnings
+
         def _warn():
             from scrapy.exceptions import ScrapyDeprecationWarning
             warnings.warn('ImagesPipeline.image_key(url) and file_key(url) methods are deprecated, '
@@ -185,15 +183,7 @@ class ImagesPipeline(FilesPipeline):
         ## end of deprecation warning block
 
         image_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        ext = "jpeg"
-        if self.convert_type == 'PRESERVE':
-            if response:
-                ext = Image.open(BytesIO(response.body)).format.lower()
-            else:
-                warnings.warn('You are preserving image type and no response body found.'
-                              'Although the type will be preserved, the path extension will default to jpeg')
-        elif self.convert_type != 'PRESERVE':
-            ext = self.convert_type.lower()
+        ext = self._determine_extension(response)
         return 'full/%s.%s' % (image_guid, ext)
 
     def thumb_path(self, request, thumb_id, response=None, info=None):
@@ -217,18 +207,22 @@ class ImagesPipeline(FilesPipeline):
             _warn()
             return self.thumb_key(url, thumb_id)
         ## end of deprecation warning block
-
+        ext = self._determine_extension(response)
         thumb_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
+        return 'thumbs/%s/%s.%s' % (thumb_id, thumb_guid, ext)
+
+    def _determine_extension(self, response):
         ext = "jpeg"
         if self.convert_type == 'PRESERVE':
             if response:
                 ext = Image.open(BytesIO(response.body)).format.lower()
             else:
+                import warnings
                 warnings.warn('You are preserving image type and no response body found.'
                               'Although the type will be preserved, the path extension will default to jpeg')
         elif self.convert_type != 'PRESERVE':
             ext = self.convert_type.lower()
-        return 'thumbs/%s/%s.%s' % (thumb_id, thumb_guid, ext)
+        return ext
 
     # deprecated
     def file_key(self, url):
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -46,6 +46,7 @@ class ImagesPipeline(FilesPipeline):
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
+    DEFAULT_IMAGES_CONVERT_TYPE = 'JPEG'
 
     def __init__(self, store_uri, download_func=None, settings=None):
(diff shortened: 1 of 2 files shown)


samj1912:
I am a new contributor to scrapy and I wanted to start with an easy bug.
I have added a feature in the imagepipeline to allow conversion between file format.
Tried to preserve the current code as much as possible to allow backward compatibility.

Looking forward to contributing more to scrapy :)

Review please :)

scrapy/pipelines/images.py
```diff
@@ -46,6 +46,7 @@ class ImagesPipeline(FilesPipeline):
     THUMBS = {}
     DEFAULT_IMAGES_URLS_FIELD = 'image_urls'
     DEFAULT_IMAGES_RESULT_FIELD = 'images'
+    DEFAULT_IMAGES_CONVERT_TYPE = 'JPEG'
```
redapple: I would suggest to not introduce another upper-case class attribute, but add a new `IMAGES_CONVERT_TYPE` setting in `default_settings.py`

scrapy/pipelines/images.py
```diff
@@ -83,6 +84,9 @@ def __init__(self, store_uri, download_func=None, settings=None):
         self.thumbs = settings.get(
             resolve('IMAGES_THUMBS'), self.THUMBS
         )
+        self.convert_type = settings.get(
+            resolve('IMAGES_CONVERT_TYPE'), self.DEFAULT_IMAGES_CONVERT_TYPE
```
redapple: Following my comment above, this line would become:

```
self.convert_type = settings.get(resolve('IMAGES_CONVERT_TYPE'))
```

scrapy/pipelines/images.py
```diff
@@ -177,7 +185,16 @@ def _warn():
         ## end of deprecation warning block
 
         image_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        return 'full/%s.jpg' % (image_guid)
+        ext = "jpeg"
```
redapple: I believe this should be `"jpg"`

scrapy/pipelines/images.py
```diff
@@ -128,19 +138,17 @@ def get_images(self, response, request, info):
             yield thumb_path, thumb_image, thumb_buf
 
     def convert_image(self, image, size=None):
-        if image.format == 'PNG' and image.mode == 'RGBA':
```
redapple: Why is this part completely removed? Is the behavior the same (as before your patch) without it?

scrapy/pipelines/images.py
```diff
@@ -177,7 +185,16 @@ def _warn():
...
         image_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        return 'full/%s.jpg' % (image_guid)
+        ext = "jpeg"
+        if self.convert_type == 'PRESERVE':
+            if response:
+                ext = Image.open(BytesIO(response.body)).format.lower()
+            else:
+                warnings.warn('You are preserving image type and no response body found.'
+                              'Although the type will be preserved, the path extension will default to jpeg')
+        elif self.convert_type != 'PRESERVE':
```
redapple: I think this is equivalent here to an `else:`

scrapy/pipelines/images.py
```diff
@@ -177,7 +185,16 @@ def _warn():
         ## end of deprecation warning block
 
         image_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        return 'full/%s.jpg' % (image_guid)
+        ext = "jpeg"
+        if self.convert_type == 'PRESERVE':
+            if response:
+                ext = Image.open(BytesIO(response.body)).format.lower()
+            else:
+                warnings.warn('You are preserving image type and no response body found.'
```
redapple: can the extension be infered from the detected image type?

scrapy/pipelines/images.py
```diff
@@ -202,7 +219,16 @@ def _warn():
         ## end of deprecation warning block
 
         thumb_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        return 'thumbs/%s/%s.jpg' % (thumb_id, thumb_guid)
+        ext = "jpeg"
```
redapple: Maybe `file_path` and `thumb_path` can share some code.

redapple (commented):
Thanks @samj1912 ! Glad to have you as new contributor.
I've been wanting this for some time.
See my comments inline.
Also, this would need updates to the docs to be complete.

scrapy/pipelines/images.py
samj1912: Yes, the behaviour is same, it makes the transparent images have a black background.

scrapy/pipelines/images.py
samj1912: Yes, the extension can be detected from the image type, I am using the format attribute and the lower casing it to get the extension. AFAICT from the PIL documentation this should work for all formats.

scrapy/pipelines/images.py
samj1912: jpeg and jpg are interchangeable right? I kept it as jpeg to be consistent with the PIL format. ( when I do `ext = Image.format.lower()` )

samj1912:
I will have to hardcode the extension when the convert type is jpeg then.

redapple:
Perhaps, yes. Backward compatiblity can be a pain indeed :)

scrapy/pipelines/images.py
```diff
@@ -100,14 +101,20 @@ def file_downloaded(self, response, request, info):
     def image_downloaded(self, response, request, info):
         checksum = None
         for path, image, buf in self.get_images(response, request, info):
+            content_type = Image.MIME[image.format]
+            if self.convert_type != 'PRESERVE':
```
redapple: (looks like I missed to comment on that earlier)

I'm not really for using a special `'PRESERVE'` string value to mean "do not convert". I would prefer using `None`.
So `if self.convert_type != 'PRESERVE':` would become `if self.convert_type is not None:` or even `if not self.convert_type:` could work I believe.
Does it make sense? Am I missing something?

samj1912: I am not that well versed with the way the resolve method handles the value 'None', how does a typical user pass a value None? by passing a blank string?

redapple: This is an example of something to be tested (and documented).
I would say people would use `IMAGES_CONVERT_TYPE = None` (this is what the docs would recommend).
If you want to support blank values, it could also work for `if not self.convert_type`, but I would not document using blank values (you can add a test for this case though)

redapple: For imagespipeline subclasses, a test would also be needed.
The docs about custom settings is quite hard to spot. https://docs.scrapy.org/en/latest/topics/media-pipeline.html#usage-example says:

> If you have multiple image pipelines inheriting from ImagePipeline and you want to have different settings in different pipelines you can set setting keys preceded with uppercase name of your pipeline class. E.g. if your pipeline is called MyPipeline and you want to have custom IMAGES_URLS_FIELD you define setting MYPIPELINE_IMAGES_URLS_FIELD and your custom settings will be used.

So a test with `MYPIPELINE_IMAGES_CONVERT_TYPE = None`, making sure this overrides the default JPEG conversion.

barraponto:
Can we have no conversion by default? That way backwards compatibility is preserved, there is no special cases and if the user wants conversion all she needs to set is `IMAGES_CONVERT_TYPE = 'jpg'`. The default should be `None`, no setting needed.

barraponto:
That also addresses the `.jpg` / `.jpeg` issue, since the user would explicitly choose the file extension.

redapple:
@barraponto , I would not mind having no-conversion by default. But as far as I understand the code, it does [currently convert](https://github.com/scrapy/scrapy/blob/d8672689761f0bb6c0550a841f35534265e87fee/scrapy/pipelines/images.py#L122) [to JPEG](https://github.com/scrapy/scrapy/blob/d8672689761f0bb6c0550a841f35534265e87fee/scrapy/pipelines/images.py#L143) by default, so it would not be backward compatible. (am I missing something?)

barraponto:
@redapple my bad, I thought we were **adding** the feature.
I can't imagine how to keep backwards compatibility and have no conversion by default...

samj1912:
@pd60193 that is an old commit, I had already corrected it.

PR #2477, opened 2017-01-03
[MRG+2] docs: installation instructions, mention conda in the beginning (closes #2475)

### LINKED ISSUE #2475: Recommend Anaconda when installing Scrapy on Windows
Succesfully using `pip` on Windows for Scrapy (and lxml and Twisted) is far from easy.
Anaconda/Miniconda seems to work much smoother.
Let's recommend Anaconda first for Windows users in the installation page, and leave the pip-way for the brave.

eliasdorneles:
Hey fellows,

As I commented in #2475 , I've installed Windows in a VM to test Scrapy installation.

Sure enough, Conda seems the easiest way, I've had no problems at all with it.

As for PyPI, I was unable to test it: I couldn't even get Python to work without Anaconda (CPython installation kept freezing). :(

Anyway, I know this is just a data point, but given the trouble that is installing and compiling stuff on Windows, I think it makes sense to mention the conda option right in the beginning and to remove all the complication about setting up Python on Windows.

What do you think?

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -7,13 +7,24 @@ Installation guide
 Installing Scrapy
 =================
 
-Scrapy runs on Python 2.7 and Python 3.3 or above
-(except on Windows where Python 3 is not supported yet).
+Scrapy runs on Python 2.7 and Python 3.3 or above.
 
-If you’re already familiar with installation of Python packages,
+If you're using `Anaconda`_ or `Miniconda`_, you can install the package from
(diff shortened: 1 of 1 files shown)


eliasdorneles:
Btw, here is a screenshot of Scrapy working under Py3 and Win:

scrapy-py3-win-2

kmike:
I'm not anaconda user myself, but I'm fine with this fix as soon as there is someone around in the issue tracker who can help in case of problems.

Issue #2461, opened 2016-12-22
TypeError: 'float' object is not iterable (on Twisted dev + Scrapy dev)


rolando:
This happens on Twisted trunk and with latest Scrapy master.

```
$ scrapy shell http://localhost:8081/
2016-12-22 12:52:01 [scrapy.utils.log] INFO: Scrapy 1.2.2 started (bot: scrapybot)
2016-12-22 12:52:01 [scrapy.utils.log] INFO: Overridden settings: {'LOGSTATS_INTERVAL': 0, 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter'}
2016-12-22 12:52:01 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.corestats.CoreStats']
2016-12-22 12:52:01 [scrapy.middleware] INFO: Enabled downloader middlewares:
['scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
 'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
 'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
 'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
 'scrapy.downloadermiddlewares.retry.RetryMiddleware',
 'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
 'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
 'scrapy.downloadermiddlewares.cookies.CookiesMiddleware',
 'scrapy.downloadermiddlewares.stats.DownloaderStats']
2016-12-22 12:52:01 [scrapy.middleware] INFO: Enabled spider middlewares:
['scrapy.spidermiddlewares.httperror.HttpErrorMiddleware',
 'scrapy.spidermiddlewares.offsite.OffsiteMiddleware',
 'scrapy.spidermiddlewares.referer.RefererMiddleware',
 'scrapy.spidermiddlewares.urllength.UrlLengthMiddleware',
 'scrapy.spidermiddlewares.depth.DepthMiddleware']
2016-12-22 12:52:01 [scrapy.middleware] INFO: Enabled item pipelines:
[]
[...]

PR #2480, opened 2017-01-04
[WIP] Test against Twisted 16.7.0rc1

redapple:
Issues found:

- #2461 (patched with https://github.com/redapple/scrapy/commit/bccf38d5cac0898dc2e63407c9ea6e646c3cf4ed)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
  #- TOXENV=jessie
  #- TOXENV=py33
  - TOXENV=py35
- - TOXENV=docs
+ #- TOXENV=docs
 install:
  - pip install -U tox twine wheel codecov
 script: tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ branches:
     - /^\d\.\d+\.\d+(rc\d+|dev\d+)?$/
 env:
  - TOXENV=py27
- - TOXENV=jessie
- - TOXENV=py33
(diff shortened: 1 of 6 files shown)


redapple:
Alright, so the trouble seems to _only_ be around name resolution.

When skipping the install of Scrapy's custom caching name resolver (https://github.com/scrapy/scrapy/pull/2480/commits/dcea9b78b93dec7783a937177b3e83dab50d0040), and therefore using Twisted's [new `HostnameEndpoint`](https://github.com/twisted/twisted/pull/624)-based name resolution, all  tests pass, except for a DNS retry test (`CrawlTestCase.test_retry_dns_error`, which mocks `socket.gethostbyname`, not used in the new default Twisted name resolution)

glyph:
I'm assuming that scrapy has a caching resolver for performance reasons?  I suspect that Twisted should add something similar to the core.

Twisted trunk should now have a thing which un-breaks scrapy if it installs its own resolver, though; fixed here: https://github.com/twisted/twisted/commit/88a71947b41af21065f7fdfa56a3ffe22a7f5c57

redapple:
Twisted 16.7 RC1 was abandoned

glyph:
Is there a new PR for 17.1 (which is now available in prerelease)?

redapple:
@glyph , yes there is: https://github.com/scrapy/scrapy/pull/2521

PR #2489, opened 2017-01-10
Update architecture.rst

aPaulius:
In the data flow image arrows are red.


redapple:
Indeed! Good catch, @aPaulius , thanks!

PR #2494, opened 2017-01-12
[WIP] adding LOG_FILEMODE settings for enabling mode in logging FileHandler

eLRuLL:
I have been seeing a lot of questions on Stackoverflow related to controlling more the scrapy logs. One in particular is the ability to rewrite the logs in the same file, which would be easily handled with this improvement by adding `LOG_FILEMODE='w'` in settings.

This is still WIP, so if you agree with this improvement I will continue working on it (add documentation, tests, etc.)

Another possibility I was thinking is to add the ability to override an entire LogHandler (just like `DUPEFILTER_CLASS`), that way there would be more control over it, as we currently only support `FileHandler`, `StreamHandler` and `NullHandler`.

Please let me know what do you think.


eLRuLL:
@redapple any idea here? would that be ok to create a `LOGHANDLER_CLASS` configurable settings variable?

AdrianAtZyte:
`LOG_FILE_APPEND` shipped in 2.6.0

PR #2495, opened 2017-01-13
[MRG] Buffer CONNECT response bytes from proxy until all HTTP headers are received

### LINKED ISSUE #2491: TLS connection fails through HTTPS proxy after CONNECT tunnel is established
I set proxy by this code:
```
class HttpProxyMiddleware(object):
        request.meta['proxy'] = 'https://127.0.0.1:8787'

```

It's error is:
> scrapy.core.downloader.handlers.http11.TunnelError: Could not open CONNECT tunnel with proxy 127.0.0.1:8787

Then I test the proxy by requests:
`resp = requests.get('https://......', proxies={'https': 'https://127.0.0.1:8787'})`
It' work!

So, what is it happen?

redapple:
Fixes #2491 

See https://github.com/scrapy/scrapy/issues/2491#issuecomment-272229679 for rationale.
A simple fix, but ideally, we'd want to use a proper HTTP parser.
Something for another PR I believe.

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -105,7 +105,7 @@ class TunnelingTCP4ClientEndpoint(TCP4ClientEndpoint):
         self._tunneledHost = host
         self._tunneledPort = port
         self._contextFactory = contextFactory
-        self._connectBuffer = b''
+        self._connectBuffer = bytearray()
 
     def requestTunnel(self, protocol):
         """Asks the proxy to open a tunnel."""
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -105,6 +105,7 @@ class TunnelingTCP4ClientEndpoint(TCP4ClientEndpoint):
         self._tunneledHost = host
         self._tunneledPort = port
         self._contextFactory = contextFactory
+        self._connectBuffer = b''
 
     def requestTunnel(self, protocol):
         """Asks the proxy to open a tunnel."""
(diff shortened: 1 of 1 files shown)


scrapy/core/downloader/handlers/http11.py
```diff
@@ -121,8 +122,16 @@ def processProxyResponse(self, rcvd_bytes):
         created, notifies the client that we are ready to send requests. If not
         raises a TunnelError.
         """
+        self._connectBuffer += rcvd_bytes
```
kmike: This could have a different algorithmic complexity in pypy; it seems using bytearray could help. I'm not sure if this is a real problem though.

kmike:
LGTM. Relevant document: http://www.freeproxy.ru/ru/free_proxy/faq/draft-luotonen-web-proxy-tunneling-01.txt:

```
An example of a tunneling request/response in an interleaved
   multicolumn format:

     CLIENT -> SERVER                        SERVER -> CLIENT
     --------------------------------------  -----------------------------------
     CONNECT home.netscape.com:443 HTTP/1.0
     User-agent: Mozilla/4.0
     <<< empty line >>>
                                             HTTP/1.0 200 Connection established
                                             Proxy-agent: Netscape-Proxy/1.1
                                             <<< empty line >>>
                  <<< data tunneling to both directions begins >>>
```

so this PR does as specified - waits for an empty line.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -121,8 +122,16 @@ def processProxyResponse(self, rcvd_bytes):
...
+        # make sure that enough (all) bytes are consumed
+        # and that we've got all HTTP headers (ending with a blank line)
+        # from the proxy so that we don't send those bytes to the TLS layer
+        #
+        # see https://github.com/scrapy/scrapy/issues/2491
+        if b'\r\n\r\n' not in self._connectBuffer:
```
kmike: This also can be O(N^2) if connectBuffer is extended byte-by-byte, but I'm not sure how to fix it, and likely it shouldn't be a problem as the response should be small.

PR #2497, opened 2017-01-15
[MRG+1] Adding formid to FormRequest documentation

eLRuLL:
The `formid` argument on `FormRequest.from_response` was added, but never documented.


kmike:
Thanks!

Issue #2501, opened 2017-01-19
scrapy view <url> raise exc in v1.3.0


wingyiu:
````
(py35) wingyiu@mbp101:~$scrapy view http://www.scrapy.org
2017-01-19 22:13:54 [scrapy.utils.log] INFO: Scrapy 1.3.0 started (bot: scrapybot)
2017-01-19 22:13:54 [scrapy.utils.log] INFO: Overridden settings: {}
Traceback (most recent call last):
  File "/Users/user/venv/py35/bin/scrapy", line 11, in <module>
    sys.exit(execute())
  File "/Users/user/venv/py35/lib/python3.5/site-packages/scrapy/cmdline.py", line 142, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/Users/user/venv/py35/lib/python3.5/site-packages/scrapy/cmdline.py", line 88, in _run_print_help
    func(*a, **kw)
  File "/Users/user/venv/py35/lib/python3.5/site-packages/scrapy/cmdline.py", line 149, in _run_command
    cmd.run(args, opts)
  File "/Users/user/venv/py35/lib/python3.5/site-packages/scrapy/commands/fetch.py", line 58, in run
    if not opts.no_redirect:
AttributeError: 'Values' object has no attribute 'no_redirect'
````

redapple:
Ouch! Thanks for reporting @wingyiu .
Looks like a regression in 1.3.0 because of https://github.com/scrapy/scrapy/pull/2410

PR #2502, opened 2017-01-19
remove add_options in the view command class

wingyiu:


redapple:
hi @wingyiu , this is related to #2501 , right?
Your patch removes the `--spider` option for the [`view` command](https://docs.scrapy.org/en/latest/topics/commands.html#std:command-view). `--spider` looks undocumented. Honestly, I don't know if this option adds anything.
If not, lets remove it indeed.

kmike:
Thanks @wingyiu for the pull request! I merged a more conservative fix (https://github.com/scrapy/scrapy/pull/2503), but your fix also looked fine.

PR #2507, opened 2017-01-23
[MRG+1] Changing README to README.rst

eLRuLL:
README files are not readable (well-formatted) inside github because of the missing extension, we can  also exclude them from sphinx build with `:orphan:`


kmike:
Thanks @eLRuLL!

PR #2509, opened 2017-01-24
[MRG+1] BUG: Fix __classcell__ propagation required in Python 3.6.

rolando:
Fixes https://github.com/scrapy/scrapy/issues/2511

Python 3.6 added simpler customization of class creation but this
requires to propagate correctly the __classcell__ attribute in custom
__new__ methods.

See https://docs.python.org/3.6/whatsnew/3.6.html#pep-487-simpler-customization-of-class-creation

Related:
https://github.com/django/django/pull/7653
https://bugs.python.org/issue23722

### DIFF
--- a/tests/test_item.py
+++ b/tests/test_item.py
@@ -1,7 +1,13 @@
+import sys
 import unittest
 
-from scrapy.item import Item, Field
 import six
+
+from scrapy.item import ABCMeta, Item, ItemMeta, Field
+from tests import mock
+
+
+PY36_PLUS = (sys.version_info.major >= 3) and (sys.version_info.minor >= 6)
 
 
 class ItemTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


rolando:
This aims to fix https://github.com/scrapy-plugins/scrapy-djangoitem/issues/18

dangra:
+1

was the regression catch by Scrapy test suite? If not, I guess we need to add a test case before merging.

redapple:
No, it wasn't caught by Scrapy's test suite.
When Python 3.6 build was added to Travis, [everything was green](https://github.com/scrapy/scrapy/pull/2485).
So +1 to a test on scrapy.Item custimization

rolando:
OK!

tests/test_item.py
```diff
@@ -244,5 +250,37 @@ class TestItem(Item):
...
+                    return __class__
+
+            MyItem()
+
+        (first_call, second_call) = new_mock.call_args_list[-2:]
+
+        mcs, class_name, bases, attrs = first_call[0]
+        assert '__classcell__' not in attrs
+        mcs, class_name, bases, attrs = second_call[0]
+        assert '__classcell__' in attrs
```
kmike: Hm, is it possible to write a more integration-style test which shows the actual error we've got?

rolando: I added ``ItemMetaClassCellRegression`` which reproduces the bug with simplest case I could find.

kmike:
Thanks @rolando!

redapple:
Thanks @rolando !
I tested with scrapy-djangoitem as described in https://github.com/scrapy-plugins/scrapy-djangoitem/issues/18#issue-202632911 and it worked fine.

PR #2510, opened 2017-01-24
[MRG+1] Preserve request class when converting to/from dicts (utils.reqser)

### LINKED ISSUE #1890: Disk queues don't preserve Request class

elacuesta:
Attempting to fix #1890

### DIFF
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -5,6 +5,7 @@ import six
 
 from scrapy.http import Request
 from scrapy.utils.python import to_unicode, to_native_str
+from scrapy.utils.misc import load_object
 
 
 def request_to_dict(request, spider=None):
@@ -32,6 +33,8 @@ def request_to_dict(request, spider=None):
         'priority': request.priority,
         'dont_filter': request.dont_filter,
     }
+    if type(request) is not Request:
+        d['_class'] = request.__module__ + '.' + request.__class__.__name__
     return d
 
 
@@ -47,7 +50,8 @@ def request_from_dict(d, spider=None):
     eb = d['errback']
     if eb and spider:
         eb = _get_method(spider, eb)
-    return Request(
+    request_cls = load_object(d['_class']) if '_class' in d else Request
+    return request_cls(
         url=to_native_str(d['url']),
         callback=cb,
         errback=eb,
--- a/tests/test_utils_reqser.py
+++ b/tests/test_utils_reqser.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import unittest
 
-from scrapy.http import Request
+from scrapy.http import Request, FormRequest
 from scrapy.spiders import Spider
 from scrapy.utils.reqser import request_to_dict, request_from_dict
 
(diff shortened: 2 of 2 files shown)


scrapy/utils/reqser.py
```diff
@@ -20,6 +19,7 @@ def request_to_dict(request, spider=None):
     if callable(eb):
         eb = _find_method(spider, eb)
     d = {
+        '__class__': request.__class__,
```
elacuesta: This doesn't work well with `scrapy.squeues.MarshalFifoDiskQueue` and `scrapy.squeues.MarshalLifoDiskQueue`, the resulting dicts cannot be serialized. I'm working on it.

kmike: I think it make sense not to store default Request class - store request class only if it is not default: 

1) it would allow to save some memory and CPU;
2) by supporting queues without `__class__` key we're making this PR compatible with disk queues created by previous Scrapy versions.

elacuesta: That was fast! I made a few modifications to this PR and now it stores a string pointing to the class, this allows the dict to be serialized using `Marshal`-based queues. It also works well with requests without `__class__` key, falling back to the default `Request`, but it is true that we would save some resources saving the class only when it's not `Request`; let me patch that.

elacuesta: @kmike Done, please check again

scrapy/utils/reqser.py
```diff
@@ -47,7 +50,11 @@ def request_from_dict(d, spider=None):
     eb = d['errback']
     if eb and spider:
         eb = _get_method(spider, eb)
-    return Request(
+    try:
```
voith: What has been the thought process here to use a try except block? Why not:
```python
 request_cls = load_object(d['_class']) if '_class' in d else Request
```
This makes more sense to me since `d['_class']` is only set when `type(request) is not Request` and since `d['_class']` stores import path of request, `load_object (d['_class'])` should not error. Or is there a case when it will throw an error which I'm unaware of?

elacuesta: The `load_object` call could fail as well. Imagine you have a disk queue containing [`scrapy_splash.request.SplashRequest`](https://github.com/scrapy-plugins/scrapy-splash/blob/master/scrapy_splash/request.py#L14) and you try to read the queue from an enviroment which doesn't have `scrapy_splash` installed.
I'm not saying how often that would happen, just that it _could_ happen :-)

kmike: I think in this case it is better to fail loudly with an exception.

voith: @elacuesta If I understood you right, what you're trying to say is that requests were stored to a disk queue by a separate project and were read by another. I don't know if that is sensible use case. My opinion here goes with Mikhail i.e fail in such a case.

elacuesta: I can make it fail, I was just following (what I thought was) the spirit of your [comment](https://github.com/scrapy/scrapy/pull/2510#discussion_r97566649) @kmike.
It would also be a bit backward incompatible, currently it wouldn't fail but yield a regular `Request` object.

elacuesta: Done, thanks for the suggestion @voith

elacuesta:
Hello there @kmike, looking forward to reading your comments after the latest changes.

kmike:
Looks good! Thanks @elacuesta and @voith.

redapple:
Thanks @elacuesta

PR #2521, opened 2017-01-31
[WIP] Test against Twisted 17.1.0rc1

redapple:


tests/test_crawl.py
```diff
@@ -91,12 +91,11 @@ def test_retry_conn_failed(self):
...
-        with mock.patch('socket.gethostbyname',
-                        side_effect=socket.gaierror(-5, 'No address associated with hostname')):
-            crawler = self.runner.create_crawler(SimpleSpider)
-            with LogCapture() as l:
-                yield crawler.crawl("http://example.com/")
-            self._assert_retried(l)
+        crawler = self.runner.create_crawler(SimpleSpider)
+        with LogCapture() as l:
+            # try to fetch the homepage of a non-existent domain
+            yield crawler.crawl("http://example.commmmm/")
```
dangra: looks tentative but we already did this in the past and it doesn't work when there is a search domain defined in `/etc/resolv.conf` like:

```
$ cat /etc/resolv.conf
nameserver 172.16.0.23
search ec2.internal scrapinghub.com

$ host example.commmm
example.commmm.scrapinghub.com has address 136.243.118.219
```

Although, adding a dot to make it explicit seems to work fine: 

```
$ host example.commmm.
Host example.commmm. not found: 3(NXDOMAIN)

$ curl http://example.commmmm/ -v
* About to connect() to example.commmmm port 80 (#0)
*   Trying 136.243.118.219... connected
> GET / HTTP/1.1
> User-Agent: curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3
> Host: example.commmmm
> Accept: */*
>
* Empty reply from server
* Connection #0 to host example.commmmm left intact
curl: (52) Empty reply from server
* Closing connection #0
$ curl http://example.commmmm./ -v
* getaddrinfo(3) failed for example.commmmm.:80
* Couldn't resolve host 'example.commmmm.'
* Closing connection #0
curl: (6) Couldn't resolve host 'example.commmmm.'
```

glyph:
As per https://www.rfc-editor.org/rfc/rfc2606.txt, you should use `.invalid` for names that should explicitly not be resolvable.

redapple:
Nice! thanks @glyph

redapple:
Patches for Twisted 17 have been merged.

PR #2528, opened 2017-02-02
Make DNS retry test compatible with Twisted 17+

redapple:
Follow up on #2462 and #2521

### DIFF
--- a/tests/test_crawl.py
+++ b/tests/test_crawl.py
@@ -92,7 +92,7 @@ class CrawlTestCase(TestCase):
         crawler = self.runner.create_crawler(SimpleSpider)
         with LogCapture() as l:
             # try to fetch the homepage of a non-existent domain
-            yield crawler.crawl("http://dns.resolution.invalid/")
+            yield crawler.crawl("http://dns.resolution.invalid./")
         self._assert_retried(l)
 
     @defer.inlineCallbacks
--- a/tests/test_crawl.py
+++ b/tests/test_crawl.py
@@ -91,12 +91,11 @@ class CrawlTestCase(TestCase):
 
     @defer.inlineCallbacks
     def test_retry_dns_error(self):
-        with mock.patch('socket.gethostbyname',
-                        side_effect=socket.gaierror(-5, 'No address associated with hostname')):
-            crawler = self.runner.create_crawler(SimpleSpider)
-            with LogCapture() as l:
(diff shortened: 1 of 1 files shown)


tests/test_crawl.py
```diff
@@ -91,12 +91,11 @@ def test_retry_conn_failed(self):
...
-        with mock.patch('socket.gethostbyname',
-                        side_effect=socket.gaierror(-5, 'No address associated with hostname')):
-            crawler = self.runner.create_crawler(SimpleSpider)
-            with LogCapture() as l:
-                yield crawler.crawl("http://example.com/")
-            self._assert_retried(l)
+        crawler = self.runner.create_crawler(SimpleSpider)
+        with LogCapture() as l:
+            # try to fetch the homepage of a non-existent domain
+            yield crawler.crawl("http://dns.resolution.invalid/")
```
kmike: In the original PR (https://github.com/scrapy/scrapy/commit/877e345d79c6a14b6ab24a3f15653f6373918cfb) @dangra mentioned that `localhost666` can be resolvable under certain circumstances. Does it apply for .invalid domains as well?

redapple: According to @glyph in https://github.com/scrapy/scrapy/pull/2521#issuecomment-276429894 and the referenced RFC2606 , `.invalid` is a reserved TLD, so I hope it's fine to use here.

dangra: it is fine as long as you add a final `.`. `.invalid` only ensures that any domainname under that tld won't ever exists.

dangra: to be more precise,  the final url should be `http://dns.resolution.invalid./`.

glyph: A trailing `.` in a domain name - or the authority section of a URL, for that matter - is perfectly valid; it's used to indicate top-level termination rather than using an implicit search domain.  If the crawler isn't fetching those it's a bug somewhere in the HTTP stack.

glyph: Oops, I think I misunderstood @dangra's comment.  I can see why you'd want to use the trailing dot there :).

redapple: alright @dangra , @kmike , @glyph , done. Thanks the input.

kmike:
Thanks @redapple, @dangra and @glyph!

PR #2530, opened 2017-02-03
[MRG+1] Use credentials from request.meta['proxy']

### LINKED ISSUE #2526: Scrapy ignores proxy credentials when using "proxy" meta key
Code `yield Request(link, meta={'proxy': 'http://user:password@ip:port’})` ignores user:password. 
Problem is solved by using header "Proxy-Authorization" with base64, but it is better to implement it inside Scrapy.

elacuesta:
Fixes #2526

At first I implemented this using [`w3lib.http.basic_auth_header`](http://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.http.basic_auth_header) but then I realized that would ignore the [`HTTPPROXY_AUTH_ENCODING`](https://scrapy.readthedocs.io/en/latest/topics/downloader-middleware.html#httpproxy-auth-encoding) setting. Would it make sense to add an `encoding` parameter to `w3lib.http.basic_auth_header`?

Also, I'm not sure how to add a test for this, I think I would need to publish credentials, but it works with something like the following (from the scrapy shell):

```
req = Request('http://example.org', meta={'proxy': 'https://APIKEY:@proxy.crawlera.com:8010'})
fetch(req)
```

/cc: @vezunch1k, @redapple

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -681,10 +681,9 @@ HttpProxyMiddleware
    * ``no_proxy``
 
    You can also set the meta key ``proxy`` per-request, to a value like
-   ``http://username:password@some_proxy_server:port``. Keep in mind
-   this value will take precedence over ``http_proxy``/``https_proxy``
-   environment variables, and it will also ignore ``no_proxy`` environment
-   variable.
+   ``http://some_proxy_server:port`` or ``http://username:password@some_proxy_server:port``.
+   Keep in mind this value will take precedence over ``http_proxy``/``https_proxy``
+   environment variables, and it will also ignore ``no_proxy`` environment variable.
 
 .. _urllib: https://docs.python.org/2/library/urllib.html
 .. _urllib2: https://docs.python.org/2/library/urllib2.html
@@ -952,7 +951,15 @@ enable it for :ref:`broad crawls <topics-broad-crawls>`.
 HttpProxyMiddleware settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. setting:: HTTPPROXY_ENABLED
 .. setting:: HTTPPROXY_AUTH_ENCODING
+
+HTTPPROXY_ENABLED
+^^^^^^^^^^^^^^^^^
+
+Default: ``True``
+
+Whether or not to enable the :class:`HttpProxyMiddleware`.
 
 HTTPPROXY_AUTH_ENCODING
 ^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -681,7 +681,10 @@ HttpProxyMiddleware
    * ``no_proxy``
 
    You can also set the meta key ``proxy`` per-request, to a value like
-   ``http://some_proxy_server:port``.
+   ``http://username:password@some_proxy_server:port``. Keep in mind
+   this value will take precedence over ``http_proxy``/``https_proxy``
+   environment variables, and it will also ignore ``no_proxy`` environment
+   variable.
 
 .. _urllib: https://docs.python.org/2/library/urllib.html
 .. _urllib2: https://docs.python.org/2/library/urllib2.html
--- a/scrapy/downloadermiddlewares/httpproxy.py
+++ b/scrapy/downloadermiddlewares/httpproxy.py
@@ -8,7 +8,6 @@ except ImportError:
 from six.moves.urllib.parse import urlunparse
 
 from scrapy.utils.httpobj import urlparse_cached
-from scrapy.exceptions import NotConfigured
 from scrapy.utils.python import to_bytes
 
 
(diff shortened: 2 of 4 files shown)


scrapy/downloadermiddlewares/httpproxy.py
```diff
@@ -45,6 +48,10 @@ def _get_proxy(self, url, orig_type):
     def process_request(self, request, spider):
         # ignore if proxy is already set
         if 'proxy' in request.meta:
+            # extract auth info if present
```
redapple: I would also test to not overwrite an already set `"Proxy-Authentication"` header

elacuesta: Sure, makes sense

elacuesta: Updated!

elacuesta:
I fixed something that was making the build fail, please check it again @redapple

redapple:
Regarding tests, I believe you can adapt https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware_httpproxy.py#L50 and the subsequent ones (i.e. instead of setting proxy as an envvar, set `"proxy"` meta on the request and check Proxy-Auth header value)

elacuesta:
Oh right, I was thinking about actually performing the requests, but I see now that checking the headers should be enough. Thanks Paul, I'll add tests similar to those 👍

elacuesta:
Hi @redapple, I included some tests. I still have one doubt though. Should scrapy skip proxies based on `no_proxy` env var even for proxies set on `request.meta`? I think users would expect that explicitly setting a proxy in `meta` would override `no_proxy`, but I would like your opinion on this. Thank you!

redapple:
I would say that a valid `"proxy"` in `Request.meta` takes precedence over env vars.
Documentation probably needs updating/clarifying regarding this.

elacuesta:
Thanks, I added a few lines to the docs explaining that, please check. Another question, if you don't mind: [this code](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpproxy.py#L23-L24) raises `NotConfigured` if there are no proxies set on `http_proxy`/`https_proxy` env vars, effectively disabling the middleware even for requests with `meta['proxy']`. Should we change that?

redapple:
Oh, nice catch. Indeed, checking "proxy" meta would require to have the middleware enabled all the time. I don't mind as long as the "no-pxoy set -> skip" case is fast enough.

redapple:
Another option could be to have an extra setting like `HTTPPROXY_CHECK_REQUEST_META` of some sort, but I am not sure I like it.

docs/topics/downloader-middleware.rst
```diff
@@ -681,7 +681,9 @@ HttpProxyMiddleware
    * ``no_proxy``
 
    You can also set the meta key ``proxy`` per-request, to a value like
-   ``http://some_proxy_server:port``.
+   ``http://some_proxy_server:port``. Keep in mind this value will take
+   precedence over ``http_proxy``/``https_proxy`` enviroment variables, and
```
redapple: Typo here: `environment`.

Also, might as well mention that the proxy value can have crendentials in it.

elacuesta:
```
elif not self.proxies:
    return
``` 
after `if 'proxy' in request.meta` in the middleware's `process_request` would skip the processing after only two simple checks, I don't think that would have much performance penalty, what do you think?

redapple:
`elif not self.proxies` would leave early enough, indeed.

elacuesta:
What about a setting (like `HTTPPROXY_ENABLED`), allowing to completely disable the middleware? Not sure what should be the default value, though.

scrapy/downloadermiddlewares/httpproxy.py
```diff
@@ -14,29 +14,32 @@
...
         self.proxies = {}
         for type, url in getproxies().items():
             self.proxies[type] = self._get_proxy(url, type)
 
-        if not self.proxies:
-            raise NotConfigured
-
     @classmethod
     def from_crawler(cls, crawler):
+        enabled = crawler.settings.getbool('HTTPPROXY_ENABLED')
```
redapple: It's probably simpler to just `raise NotConfigured` here directly, similarly to [how `CookiesMiddleware` does](https://github.com/scrapy/scrapy/blob/129421c7e31b89b9b0f9c5f7d8ae59e47df36091/scrapy/downloadermiddlewares/cookies.py#L23).

elacuesta: True! Although it does make the test a little bit more complicated :-P

docs/topics/downloader-middleware.rst
```diff
@@ -681,7 +681,10 @@ HttpProxyMiddleware
    * ``no_proxy``
 
    You can also set the meta key ``proxy`` per-request, to a value like
-   ``http://some_proxy_server:port``.
+   ``http://username:password@some_proxy_server:port``. Keep in mind
```
kmike: What do you think about keeping both examples?

> ... to a value like  ``http://some_proxy_server:port`` or ``http://username:password@some_proxy_server:port``

scrapy/downloadermiddlewares/httpproxy.py
```diff
@@ -20,23 +20,25 @@ def __init__(self, auth_encoding='latin-1'):
...
+        if not crawler.settings.getbool('HTTPPROXY_ENABLED'):
+            raise NotConfigured
         auth_encoding = crawler.settings.get('HTTPPROXY_AUTH_ENCODING')
         return cls(auth_encoding)
 
+    def _basic_auth_header(self, username, password):
+        user_pass = to_bytes(
+            '%s:%s' % (unquote(username), unquote(password)),
+            encoding=self.auth_encoding)
+        return base64.b64encode(user_pass).strip()
```
kmike: Do you know why is it different from w3lib version? b64encode vs urlsafe_b64encode, unqoute vs no unquote?

kmike:
Looks good 👍

scrapy/downloadermiddlewares/httpproxy.py
elacuesta: I don't really know, what I did was move that code to a separate function to be able to reuse it. I think the `w3lib` version should work just fine here, the only thing missing would be the ability to define encoding.

docs/topics/downloader-middleware.rst
elacuesta: Sure, makes sense :-)

kmike:
Thanks @elacuesta!

PR #2532, opened 2017-02-03
CsvItemExporter: don't write headers to files when appending

funkyfuture:
this handles the case when an export feed uri points to an existing csv file, which would otherwise contain repeated header lines. and hardly any of them on top.


scrapy/exporters.py
```diff
@@ -175,17 +175,22 @@ class CsvItemExporter(BaseItemExporter):
...
         self.stream = io.TextIOWrapper(
             file,
             line_buffering=False,
             write_through=True,
             encoding=self.encoding
         ) if six.PY3 else file
+
+        # don't write a header to a file with contents
+        self.stream.seek(0, 2)
+        if self.stream.tell() > 0:
```
kmike: Previously exporters didn't require .seek or .tell methods to be implemented. I'm not sure it works e.g. with StdoutFeedStorage.

I think this is a great feature, but autodetection can be tricky. 
What do you think about adding an option to disable CSV headers instead? It won't be as nice and easy to use, but user will have a full control on it.

funkyfuture:
good point. you mean an option in the settings module (`FEED_OPTIONS`)? that certainly would be a solution, but that should imply to implement a generic way to supply exporter specific, and more or less arbritray, options. which is imo an overkill to solve that particular issue. thus i'd prefer to just catch the exception from the `io` module and proceed.

kmike:
@funkyfuture I was thinking about a single `FEED_EXPORT_WRITE_HEADER` option just to enable/disable csv header, nothing generic.

PR #2533, opened 2017-02-05
[MRG+1] Use yield with nested parsing of responses

djrobust:
In my project I found that `return` made the spider terminate early, while `yield` worked as intended. `yield` is also used throughout other examples of the `parse` function in the documentation, so even if functionality is similar, using `yield` in this example improves consistency.


kmike:
I think this change is fine.

PR #2535, opened 2017-02-06
[MRG+1] [httpcompression] add support for br - brotli content encoding

pawelmhm:
brotli encoding is used by multiple websites, e.g. amazon started to use it recently. Support in Scrapy is added via python library brotlipy https://github.com/python-hyper/brotlipy/


redapple:
Does the library require anything special/additional to get installed properly?
Maybe importing brotlipy can be tested at inittime/runtime and support for `br` only advertized in headers if succesful?

pawelmhm:
> Does the library require anything special/additional to get installed properly?

I didnt have to install anything, but it seems like some environments are failing to install, so most likely some systems are missing required libraries. Testing for support at runtime may be much safer yes, I'll add that @redapple

redapple:
@pawelmhm , I do not see brotlipy being installed for Python 3.x tests on Travis.

pawelmhm:
@redapple sorry missed that updated in af802bad14f833178bc4e

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -1,23 +1,33 @@
 import zlib
 
+
```
kmike: pep8 says 1 line between import groups :)

tests/requirements.txt
```diff
@@ -6,6 +6,7 @@ pytest==2.9.2
 pytest-twisted
 pytest-cov==2.2.1
 jmespath
+brotlipy==0.6
```
kmike: I'd prefer not to pin brotlipy version here: users will be installing it with 'pip install brotlipy', and we won't be able to detect if a brotlipy upgrade broke scrapy if we pin a version in tests.

kmike:
The PR looks good, besides a couple of minor points 👍 A nice feature.
Have you compared gzip and brotli for amazon? How's the download size and CPU usage?

scrapy/downloadermiddlewares/httpcompression.py
pawelmhm: good point, we could add flake8 style checks for PR. Flake8 could run on diff and point out all things like this.

tests/requirements.txt
pawelmhm: updated fb4ef21a1d

pawelmhm:
@kmike re performance there are good benchmarks and charts here: https://quixdb.github.io/squash-benchmark/#ratio-vs-compression

kmike:
@pawelmhm I was more worried about Python wrappers speed

Issue #1982, opened 2016-05-11
'=' added at end of urls extracted from LinkExtractor


gmargari:
```
$ scrapy shell "https://uat.payleap.com/transactservices.svc"
> from scrapy.linkextractors import LinkExtractor
> [ l.url for l in LinkExtractor(allow=()).extract_links(response) ]
['https://uat.payleap.com/TransactServices.svc?wsdl=']
```
Scrapy shell info:
```
2016-05-11 16:37:21 [scrapy] INFO: Scrapy 1.0.5 started (bot: scrapybot)
2016-05-11 16:37:21 [scrapy] INFO: Optional features available: ssl, http11
2016-05-11 16:37:21 [scrapy] INFO: Overridden settings: {'LOGSTATS_INTERVAL': 0, 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter'}
2016-05-11 16:37:22 [scrapy] INFO: Enabled extensions: CloseSpider, TelnetConsole, CoreStats, SpiderState
2016-05-11 16:37:22 [scrapy] INFO: Enabled downloader middlewares: HttpAuthMiddleware, DownloadTimeoutMiddleware, UserAgentMiddleware, RetryMiddleware, DefaultHeadersMiddleware, MetaRefreshMiddleware, HttpCompressionMiddleware, RedirectMiddleware, CookiesMiddleware, ChunkedTransferMiddleware, DownloaderStats
2016-05-11 16:37:22 [scrapy] INFO: Enabled spider middlewares: HttpErrorMiddleware, OffsiteMiddleware, RefererMiddleware, UrlLengthMiddleware, DepthMiddleware
2016-05-11 16:37:22 [scrapy] INFO: Enabled item pipelines: 
2016-05-11 16:37:22 [scrapy] DEBUG: Telnet console listening on 127.0.0.1:6024
2016-05-11 16:37:22 [scrapy] INFO: Spider opened
2016-05-11 16:37:23 [scrapy] DEBUG: Crawled (200) <GET https://uat.payleap.com/transactservices.svc> (referer: None)
```

redapple:
@gmargari , this is the (unfortunate) behavior of `canonicalize_url` (enabled by default in `LinkExtractor`).

You can disable it though:

```
In [1]: from scrapy.linkextractors import LinkExtractor

In [2]: [ l.url for l in LinkExtractor(allow=()).extract_links(response) ]
[...]

PR #2537, opened 2017-02-06
[MRG+1] Set canonicalize=False for LinkExtractor

kmike:
A fix for #1941 and #1982.

TODO: 

* [x] fix deprecated link extractors

### DIFF
--- a/scrapy/linkextractors/__init__.py
+++ b/scrapy/linkextractors/__init__.py
@@ -101,7 +101,7 @@ class FilteringLinkExtractor(object):
         links = [x for x in links if self._link_allowed(x)]
         if self.canonicalize:
             for link in links:
-                link.url = canonicalize_url(urlparse(link.url))
+                link.url = canonicalize_url(link.url)
         links = self.link_extractor._process_links(links)
         return links
 
--- a/docs/topics/link-extractors.rst
+++ b/docs/topics/link-extractors.rst
@@ -51,7 +51,7 @@ LxmlLinkExtractor
    :synopsis: lxml's HTMLParser-based link extractors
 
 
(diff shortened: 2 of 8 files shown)


scrapy/linkextractors/__init__.py
```diff
@@ -101,7 +101,7 @@ def _process_links(self, links):
         links = [x for x in links if self._link_allowed(x)]
         if self.canonicalize:
             for link in links:
-                link.url = canonicalize_url(urlparse(link.url))
+                link.url = canonicalize_url(link.url)
```
kmike: @redapple do you recall why did you use urlparse here? Is it just a left-over from refactoring, or is it something else?

redapple: I'm not 100% sure but I believe it was to have the same behavior as for [`SgmlLinkExtractor`](https://github.com/scrapy/scrapy/blob/155ea08ea1c50f347c21cb4635b36dd47141f4cc/scrapy/contrib/linkextractors/sgml.py#L150) prior to refactoring with `FilteringLinkExtractor`.

kmike: I'm also not 100% sure, but for me it looks like SgmlLinkExtractor did that as a performance optimization, in order to avoid parsing URL multiple times. But now there is no performance benefit, and documented canonicalize_url API is to accept URLs as strings.

redapple:
What release would you expect this in?

kmike:
@redapple I don't know.. This is a a bug fix, but it is also a big change (e.g. fragments are preserved by default now). Waiting until 1.4 sounds fine to me.

dangra:
needs rebase since https://github.com/scrapy/scrapy/pull/2547 is merged

redapple:
Should this be marked as backwards-incompatible?

kmike:
@redapple right, e.g. fragments are no longer stripped.

kmike:
Rebased.

Issue #1940, opened 2016-04-20
Better API for creating requests from responses


kmike:
Sometime Request needs information about the response to be sent correctly. There are at least 2 use cases:
1. URL encoding depends on response encoding (see https://github.com/scrapy/scrapy/pull/1923);
2. relative URLs should be resolved based on response.url or its base url (see https://github.com/scrapy/scrapy/issues/548).

I think the current API is not good enough. The most obvious code is not correct:

``` py
for url in response.xpath("//a/@href").extract():
    yield scrapy.Request(url, self.parse)
```

To do that correctly user has to write the following:

``` py
for url in response.xpath("//a/@href").extract():
    yield scrapy.Request(response.urljoin(url), self.parse, encoding=response.encoding)
```

Or this:

``` py
for link in LinkExtractor().extract_links(response):
    yield scrapy.Request(link.url, self.parse, encoding=response.encoding)
```

LinkeExtractor solution has gotchas, e.g. canonicalize_url is called by default and fragments are removed. It means that e.g. Ajax crawlable URLs are not handled (no `escaped_fragment` even if a website supports it); it also makes it harder to use Scrapy with scrapy-splash which can handle fragments.

This all is too easy to get wrong; I think just documenting these gotchas is not good enough for a framework - it should make the easiest way to write something the correct way. IMHO in the API shouldn't require user to instantiate weird objects or pass response encoding:

``` py
for url in response.xpath("//a/@href").extract():
    something.send_request(url, self.parse)
```

This can be implemented if we provide a method on Response to send new requests.

A related use case is `async def` functions or methods (https://github.com/scrapy/scrapy/issues/1144#issuecomment-141843616): it is not possible to yield Requests in `async def` functions, so adding a request should be either a method of `self` or a method of `response` if we want to support `async def` callbacks.

`FormRequest.from_response(response, ...)` can also be written as something like `response.submit(...)`.

Digenis:
More LinkExtractor gotchas: #1941

I think providing a method in the response to prepare requests and urls
can be implemented fast and reliably without breaking compatibility.

The rest, sending the request to the scheduler, sounds like something that requires a separate issue.

kmike:
Another way to fix response.encoding problem is to create a middleware which sets request.encoding to response.encoding if it is not set explicitly. But it would be backwards incompatible, and @dangra didn't like middleware approach for relative URLs and ruled them out (which was a good decision).

dangra:
[...]

PR #2540, opened 2017-02-07
response.follow

kmike:
A proposed fix for https://github.com/scrapy/scrapy/issues/1940.

TODO:

* [x] tests
* [x] docs
* [x] Response.follow
* [x] default encoding is response encoding, not utf8 - is it OK for auto-detected encodings ~~(which is cp1252 by default for responses without headers/meta tags)~~?

### DIFF
--- a/scrapy/http/response/text.py
+++ b/scrapy/http/response/text.py
@@ -11,6 +11,7 @@ from six.moves.urllib.parse import urljoin
 import parsel
 from w3lib.encoding import html_to_unicode, resolve_encoding, \
     html_body_declared_encoding, http_content_type_encoding
+from w3lib.html import strip_html5_whitespace
 
 from scrapy.link import Link
 from scrapy.http.request import Request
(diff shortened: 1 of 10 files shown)


scrapy/http/response/text.py
```diff
@@ -116,3 +120,57 @@ def xpath(self, query, **kwargs):
...
+        * attribute Selector (not SelectorList) - e.g.
+          ``response.css('a::attr(href)')[0]`` or
+          ``response.xpath('//img/@src')[0]``.
+        * a Selector for ``<a>`` element, e.g.
+          ``response.css('a.my_link')[0]``.
+        """
+        if isinstance(url, Link):
+            url = url.url
+        elif isinstance(url, parsel.Selector):
+            url = _url_from_selector(url).strip()
```
kmike: this could benefit from `trim_href_attribute` function from https://github.com/scrapy/scrapy/pull/2547

scrapy/http/response/text.py
```diff
@@ -116,3 +120,57 @@ def xpath(self, query, **kwargs):
 
     def css(self, query):
         return self.selector.css(query)
+
+    def follow(self, url, callback=None, method='GET', headers=None, body=None,
```
dangra: why is `TextResponse` only method?

kmike: Right, I was going to add it to Response as well :)

Issue #2173, opened 2016-08-11
Dump stats to log periodically, not only at the end of the crawl


kmike:
It is useful to check Scrapy stats as spider runs, but there is no a built-in way to do that. What do you think about adding `DUMP_STATS_INTERVAL` option and outputting current stats to logs each `DUMP_STATS_INTERVAL` seconds?

Another related proposal (sorry for putting them all into this ticket) is to add more logging to Downloader and log periodically a number of pages Downloader is currently trying to fetch (`MONITOR_DOWNLOADS_INTERVAL` option). Checking that helps to understand what is crawler doing - is it busy downloading data or not.

Implementation draft:

``` python
import logging
import pprint

from twisted.internet.task import LoopingCall
from scrapy import signals

logger = logging.getLogger(__name__)


class _LoopingExtension:
    def setup_looping_task(self, task, crawler, interval):
        self._interval = interval
        self._task = LoopingCall(task)
        crawler.signals.connect(self.spider_opened, signal=signals.spider_opened)
        crawler.signals.connect(self.spider_closed, signal=signals.spider_closed)

    def spider_opened(self):
        self._task.start(self._interval, now=False)

    def spider_closed(self):
        if self._task.running:
            self._task.stop()


class MonitorDownloadsExtension(_LoopingExtension):
    """
    Enable this extension to periodically log a number of active downloads.
    """
    def __init__(self, crawler, interval):
        self.crawler = crawler
        self.setup_looping_task(self.monitor, crawler, interval)

    @classmethod
    def from_crawler(cls, crawler):
        # fixme: 0 should mean NotConfigured
        interval = crawler.settings.getfloat("MONITOR_DOWNLOADS_INTERVAL", 10.0)
        return cls(crawler, interval)

    def monitor(self):
        active_downloads = len(self.crawler.engine.downloader.active)
        logger.info("Active downloads: {}".format(active_downloads))


class DumpStatsExtension(_LoopingExtension):
    """
    Enable this extension to log Scrapy stats periodically, not only
    at the end of the crawl.
    """
    def __init__(self, crawler, interval):
        self.stats = crawler.stats
        self.setup_looping_task(self.print_stats, crawler, interval)

    def print_stats(self):
        stats = self.stats.get_stats()
        logger.info("Scrapy stats:\n" + pprint.pformat(stats))

    @classmethod
    def from_crawler(cls, crawler):
        interval = crawler.settings.getfloat("DUMP_STATS_INTERVAL", 60.0)
[...]

PR #2543, opened 2017-02-07
[MRG+1] Retry stats

kmike:
What do you think about providing retry counts in scrapy stats?

~~This PR is not backwards compatible for anyone who used `RetryMiddleware.__init__` directly.~~

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -36,18 +36,16 @@ class RetryMiddleware(object):
                            ConnectionLost, TCPTimedOutError, ResponseFailed,
                            IOError, TunnelError)
 
-    def __init__(self, crawler):
-        settings = crawler.settings
+    def __init__(self, settings):
         if not settings.getbool('RETRY_ENABLED'):
             raise NotConfigured
         self.max_retry_times = settings.getint('RETRY_TIMES')
         self.retry_http_codes = set(int(x) for x in settings.getlist('RETRY_HTTP_CODES'))
         self.priority_adjust = settings.getint('RETRY_PRIORITY_ADJUST')
-        self.stats = crawler.stats
 
     @classmethod
     def from_crawler(cls, crawler):
-        return cls(crawler)
+        return cls(crawler.settings)
 
     def process_response(self, request, response, spider):
         if request.meta.get('dont_retry', False):
@@ -65,6 +63,7 @@ class RetryMiddleware(object):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
+        stats = spider.crawler.stats
         if retries <= self.max_retry_times:
             logger.debug("Retrying %(request)s (failed %(retries)d times): %(reason)s",
                          {'request': request, 'retries': retries, 'reason': reason},
@@ -77,11 +76,11 @@ class RetryMiddleware(object):
             if isinstance(reason, Exception):
                 reason = global_object_name(reason.__class__)
 
-            self.stats.inc_value('retry/count')
-            self.stats.inc_value('retry/reason_count/%s' % reason)
+            stats.inc_value('retry/count')
+            stats.inc_value('retry/reason_count/%s' % reason)
             return retryreq
         else:
-            self.stats.inc_value('retry/max_reached')
+            stats.inc_value('retry/max_reached')
             logger.debug("Gave up retrying %(request)s (failed %(retries)d times): %(reason)s",
                          {'request': request, 'retries': retries, 'reason': reason},
                          extra={'spider': spider})
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -22,6 +22,7 @@ from twisted.web.client import ResponseFailed
 from scrapy.exceptions import NotConfigured
 from scrapy.utils.response import response_status_message
 from scrapy.core.downloader.handlers.http11 import TunnelError
+from scrapy.utils.python import global_object_name
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


scrapy/downloadermiddlewares/retry.py
```diff
@@ -70,8 +72,10 @@ def _retry(self, request, reason, spider):
             retryreq.meta['retry_times'] = retries
             retryreq.dont_filter = True
             retryreq.priority = request.priority + self.priority_adjust
+            self.stats.inc_value('retry/count')
```
redapple: what about including the retry reason in the stats? a bit like DownloaderStats

kmike: I thought that it could be redundant - you'll get the same exception names as in downloader, e.g. `'downloader/exception_type_count/twisted.internet.error.ConnectionRefusedError': 155` and `'retry/twisted.internet.error.ConnectionRefusedError': 155`; `downloader/response_status_count/` will be also duplicated - it could be 20x lines in stats with very little new information.

kmike: I can see though how it makes it more obvious for the user what caused these retries

scrapy/downloadermiddlewares/retry.py
```diff
@@ -70,8 +72,10 @@ def _retry(self, request, reason, spider):
...
             return retryreq
         else:
+            self.stats.inc_value('retry/stopped')
```
redapple: I have doubts on using "stopped" here, but I can't find a satisfying alternative (`retry/givingup`, `retry/max_times_reached`...)

kmike: Yeah, a better name is welcome.

scrapy/downloadermiddlewares/retry.py
redapple: It would tell you explicitly that these exceptions are retried, while `downloader/exception` "only" tells you they happen. Then you may want to disable retries or retry less.

kmike: RIght, but all these exceptions are listed above, in EXCEPTIONS_TO_RETRY list. We're not documenting this list though, and names could be different for subclasses.

redapple: Sorry, I don't get that.

redapple: You may spot a retried status code that you didn't intend to (wrong setting perhaps), or a missing retried exception etc.

kmike: Yeah, this is true. My point is that you can spot it without adding this information to stats: status codes and exceptions to retry are static, they don't depend on website - you can check settings and EXCEPTIONS_TO_RETRY list. You check `retry/count`, notice that the value is high (=> there is some issue with crawling this website), and then investigate it further: check response status codes and downloader exceptions. So this PR does this minimal change which allows to spot and debug the problem.

I'm not opposed to adding more information to stats, it is just a bit more controversial IMHO (https://github.com/scrapy/scrapy/issues/2173 becomes more verbose). But more stats could make it easier to debug retry issues, that's true.

redapple: I don't mind having verbose stats if it's to debug problems. Having to look at the code for `EXCEPTIONS_TO_RETRY` and default `RETRY_HTTP_CODES`  (and settings to double check if it's not the default) feels not-straightforward.
I'm not sure I would spot `retry/count` more than high `downloader/exception_type_count` values.

kmike: Do you want per-reason counts for retry/stopped as well?

redapple: > Do you want per-reason counts for retry/stopped as well?

I don't think so, no

kmike: @redapple done!

kmike:
@redapple I tried running scrapy with this patch on a crawl, and having per-error stats look useful indeed.

scrapy/downloadermiddlewares/retry.py
redapple: what about `retry/max_reached`?

redapple:
Cool. Do you have sample logs to share showing how it looks?

kmike:
In my case the log is a bit unusual because most times exceptions are handled by another middleware before they get to standard retry middleware, so there is not a lot of duplication:

```
 'downloader/exception_count': 49828,
 'downloader/exception_type_count/builtins.UnicodeEncodeError': 31,
 'downloader/exception_type_count/scrapy.core.downloader.handlers.http11.TunnelError': 2,
 'downloader/exception_type_count/twisted.internet.error.ConnectionRefusedError': 3370,
 'downloader/exception_type_count/twisted.internet.error.NoRouteError': 248,
 'downloader/exception_type_count/twisted.internet.error.TCPTimedOutError': 578,
 'downloader/exception_type_count/twisted.internet.error.TimeoutError': 35726,
 'downloader/exception_type_count/twisted.web._newclient.ResponseFailed': 296,
 'downloader/exception_type_count/twisted.web._newclient.ResponseNeverReceived': 9577,
 'downloader/request_bytes': 215175880,
 'downloader/request_count': 530706,
 'downloader/request_method_count/GET': 530706,
 'downloader/response_bytes': 2823160968,
 'downloader/response_count': 480778,
 'downloader/response_status_count/200': 462638,
 'downloader/response_status_count/301': 12610,
 'downloader/response_status_count/302': 25,
 'downloader/response_status_count/307': 146,
 'downloader/response_status_count/401': 6,
 'downloader/response_status_count/403': 1462,
 'downloader/response_status_count/404': 513,
 'downloader/response_status_count/500': 835,
 'downloader/response_status_count/502': 339,
 'downloader/response_status_count/503': 1842,
 'downloader/response_status_count/504': 324,
 'downloader/response_status_count/520': 30,
 'downloader/response_status_count/525': 8,
 'dupefilter/filtered': 174522,
 'dupefilter2/seen': 1150356,
 'item_scraped_count': 462637,
 'links/ad': 48946,
 'links/page': 429,
 'log_count/INFO': 9176,
 'pages/ad': 462059,
 'pages/category': 578,
 'request_depth_max': 9,
 'response_received_count': 462637,
 'retry/count': 5,
 'retry/reason_count/twisted.internet.error.ConnectionRefusedError': 1,
 'retry/reason_count/twisted.internet.error.TimeoutError': 2,
 'retry/reason_count/twisted.web._newclient.ResponseNeverReceived': 2,
 'scheduler/dequeued': 530706,
 'scheduler/dequeued/disk': 530706,
 'scheduler/enqueued': 112089,
 'scheduler/enqueued/disk': 112089,
```

tests/test_proxy_connect.py
```diff
@@ -101,7 +101,9 @@ def test_https_noconnect_auth_error(self):
         self._assert_got_response_code(407, l)
 
     def _assert_got_response_code(self, code, log):
+        print(log)
```
redapple: Is this leftover from debugging?

kmike: Not exactly a leftover; stdout is hidden by default, but pytest shows it if tests fail. So by printing log it becomes easier to debug the issue in case of failures.

kmike:
I've searched our internal codebase and there are several examples of RetryMiddleware subclasses with `__init__` method overridden.

kmike:
Merging it, given that https://github.com/scrapy/scrapy/pull/2566 implementation is the same and I had @redapple's +1 for the previous backwards incompatible version.

PR #2544, opened 2017-02-07
[MRG+1] Remove code required to support ancient twisted versions.

kmike:
A follow-up to #1887.


redapple:
Great @kmike !

stav:
Wow, it's really nice to see someone going thru and making changes like this that don't provide "incredible new feature!" to the release notes but make the code tighter, lighter and cleaner.

Issue #1614, opened 2015-11-24
`canonicalize_url` should not accept whitespace in the begin of an URI


abmxer:
Accepting whitespace in the begin of URI introduces a crawling issue. Unique URIs are generated on each depth and crawler gets in a loop. 

When a URI is extracted and begins with whitespace it is being considered a relative URI and its concatenated at the end of current URI, which leads to unique URI on each crawler depth. 

But, accordingly to RFC2396 [ Page 38], the scheme parte of a URI should begin with an ALPHA character. 

> "The syntax for URI scheme has been changed to require that all schemes begin with an alpha character." 

After further investigation, the core python `urlparse` function doesn't follow RFC strictly, they consider some rules. Like on Python 3.5 they say: 

> "Changed in version 3.3: The fragment is now parsed for all URL schemes (unless allow_fragment is false), in accordance with RFC 3986. Previously, a whitelist of schemes that support fragments existed." 

However a better implementation should be in accordance with RFC 3986 (which is an update of RFC2396 mentioned before), the `urlparse` should not accept whitespace at the begin of URI. 

When compared to the with Ruby implementation of RFC 2396,  they don't allow whitespace at the begin of absolute or relative URIs and throws an exception "bad URI(is not URI?)".

In order to fix that on Scrapy level, I suggest the following. 

In `scrapy/utils/url.py` add at the begin of `canonicalize_url`:

```
if re.match('[a-zA-Z]', url[:1]) == None:
    raise ValueError('Bad URI (is a valid URI?)')
```
An a test in `tests/test_utils_url.py`:
```
def test_rfc2396_scheme_should_begin_with_alfa(self):
    self.assertRaises(ValueError, canonicalize_url," http://www.example.com")
```
But raising an exception breaks a lot of tests and can introduce unexpected behaviors. You could simple `.strip()` url, but won't make it RFC compliant.

kmike:
Hi @abmxer,

I think in practice the question is where users get URLs with whitespaces from - they usually get them from a href attributes. [HTML5](https://www.w3.org/TR/html5/links.html#links-created-by-a-and-area-elements) explicitly allows whitespaces before and after href value, and requires browsers to strip them.

There are several places we may fix it - `canonicalize_url`, link extractors (it won't help with links returned by `resp.xpath('//a/@href').extract()` though), `safe_download_url`, maybe somewhere else. I'm not sure about the best place to fix it.

[...]

PR #1603, opened 2015-11-16
clean the crawled urls

simonkuang:
clean urls which have whitespace/tab/cr/lf chars.

PR #1021, opened 2015-01-22
Stripping hrefs extracted by linkextractor of preceding and following spaces

Granitosaurus:
stripping all extracted `href` and `src` attributes of preceding and following spaces to reduce infite looping as described in issue https://github.com/scrapy/scrapy/issues/838.

PR #2547, opened 2017-02-08
[MRG+1] LinkExtractors: strip whitespaces

### LINKED ISSUE #838: Links with a preceding space are not parsed correctly.
For example:

Imagine on this page http://www.example.co.uk/?post=123 we have this markup:

```
<a href=" http://test.example.co.uk/">Test</a>
```

It's subtle but there's a preceding space before the URL in the href.

Instead of crawling to the URL http://test.example.co.uk/ it crawls this:

http://www.example.co.uk/?post=123%20http://test.example.co.uk/

Further more, if this page exists, then it may likely contain the same link, which means there's a link like this:

http://www.example.co.uk/?post=123%20http://test.example.co.uk/%20http://test.example.co.uk/

This causes an infinite loop.

Essentially, the solution is to trim URLs before crawling them.

kmike:
I thought this fixes #838, but after re-reading the issue it was not only about link extractors :) But LinkExtractors should be fixed anyways, because html5 rules are clear.

See also: https://github.com/scrapy/scrapy/issues/1614, https://github.com/scrapy/scrapy/pull/1021, https://github.com/scrapy/scrapy/pull/1603.

This PR conflicts with https://github.com/scrapy/scrapy/pull/2537 - when one of them is merged the other have to be fixed.

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -103,3 +103,16 @@ def guess_scheme(url):
         return any_to_uri(url)
     else:
         return add_http_if_no_scheme(url)
+
+
+def trim_href_attribute(href):
+    """
+    Process href attribute of ``a`` or ``area`` elements according to HTML5
+    standards (strip all leading and trailing whitespaces). References:
+
+    * https://www.w3.org/TR/html5/links.html#links-created-by-a-and-area-elements
+    * https://www.w3.org/TR/html5/infrastructure.html#valid-url-potentially-surrounded-by-spaces
+    * https://www.w3.org/TR/html5/infrastructure.html#strip-leading-and-trailing-whitespace
+    * https://www.w3.org/TR/html5/infrastructure.html#space-character
+    """
+    return href.strip(' \t\n\r\x0c')
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -103,16 +103,3 @@ def guess_scheme(url):
         return any_to_uri(url)
     else:
         return add_http_if_no_scheme(url)
-
-
-def trim_href_attribute(href):
-    """
-    Process href attribute of ``a`` or ``area`` elements according to HTML5
-    standards (strip all leading and trailing whitespaces). References:
-
-    * https://www.w3.org/TR/html5/links.html#links-created-by-a-and-area-elements
-    * https://www.w3.org/TR/html5/infrastructure.html#valid-url-potentially-surrounded-by-spaces
-    * https://www.w3.org/TR/html5/infrastructure.html#strip-leading-and-trailing-whitespace
-    * https://www.w3.org/TR/html5/infrastructure.html#space-character
-    """
-    return href.strip(' \t\n\r\x0c')
--- a/docs/topics/link-extractors.rst
+++ b/docs/topics/link-extractors.rst
@@ -51,7 +51,7 @@ LxmlLinkExtractor
    :synopsis: lxml's HTMLParser-based link extractors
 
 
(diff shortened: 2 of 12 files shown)


scrapy/utils/url.py
```diff
@@ -103,3 +103,16 @@ def guess_scheme(url):
         return any_to_uri(url)
     else:
         return add_http_if_no_scheme(url)
+
+
+def trim_href_attribute(href):
```
redapple: what do you think of using a more generic name like `strip_whitespace`?

kmike: `strip_html_whitespace` or something like that sounds fine to me, a good idea. In this case we should probably move it somewhere else. 

This function uses whitespace characters defined in html5 standard; just 'strip whitespaces' is `my_string.strip()`.

kmike: Though code is easier to read with `trim_html_attribute` name - I think it makes assumptions made by link extractors more clear.

scrapy/utils/url.py
```diff
@@ -103,3 +103,16 @@ def guess_scheme(url):
...
+    """
+    Process href attribute of ``a`` or ``area`` elements according to HTML5
+    standards (strip all leading and trailing whitespaces). References:
+
+    * https://www.w3.org/TR/html5/links.html#links-created-by-a-and-area-elements
+    * https://www.w3.org/TR/html5/infrastructure.html#valid-url-potentially-surrounded-by-spaces
+    * https://www.w3.org/TR/html5/infrastructure.html#strip-leading-and-trailing-whitespace
+    * https://www.w3.org/TR/html5/infrastructure.html#space-character
+    """
+    return href.strip(' \t\n\r\x0c')
```
redapple: Could the Unicode character (`\N{FF}` or `\u000c`) be used here?
or is it dealing with bytes?

scrapy/utils/url.py
redapple: I'd go for `strip_w3c_whitespace` then. With "html", one could understand stripping whitespace from HTML source code or something. (just my 2 cents)

scrapy/utils/url.py
kmike: It deals with text if I'm not mistaken. Constants are from https://www.w3.org/TR/html5/infrastructure.html#space-character. `'\u000c' == '\x0c'` in Python 3, and Python 2 doesn't understand `'\u000c'`.

scrapy/utils/url.py
redapple: As for code readability, I think the part of link extractors where it's used is already self-explanatory, that it's stripping the attribute value. (And it could be used by some users for non-href, although I don't know how that's defined in the specs, for `src` for example)

scrapy/utils/url.py
redapple: Why not `u' \t\n\r\u000c'`?

scrapy/utils/url.py
kmike: Whitespaces should be stripped form `` values as well - https://www.w3.org/TR/html5/embedded-content-0.html#attr-img-src. So `trim_href_...` is not a good name indeed.

kmike: The same applies for form action, iframe src, base href, etc.; this seems to be consistent.

scrapy/utils/url.py
kmike: The difference is that `u' \t\n\r\u000c'` promotes result to unicode in Python 2, while `' \t\n\r\x0c'` keeps its type.

kmike: It seems Python prefers `\x0c` way to write such constants:

```
>>> u' \t\n\r\u000c'
u' \t\n\r\x0c'
```

scrapy/utils/url.py
kmike: I moved it to https://github.com/scrapy/w3lib/pull/86 - what do you think @redapple?

redapple: Sounds good @kmike

PR #2548, opened 2017-02-08
[MRG+1] FormRequest: handle whitespaces in action attribute properly

kmike:
~~It requires w3lib 0.17.0+, but I haven't bumped the requirement in this PR.~~

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -5,10 +5,13 @@ This module implements the FormRequest class which is a more convenient class
 See documentation in docs/topics/request-response.rst
 """
 
+import six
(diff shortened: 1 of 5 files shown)


scrapy/http/request/form.py
```diff
@@ -51,7 +54,10 @@ def from_response(cls, response, formname=None, formid=None, formnumber=0, formd
 
 def _get_form_url(form, url):
     if url is None:
-        return urljoin(form.base_url, form.action)
```
kmike: this urljoin was pointless because form.action already calls `urljoin(form.base_url, action)`

kmike:
Rebased on top of #2559.

Issue #2555, opened 2017-02-13
Scrapy incompatible with twisted 17.1.0 on TLS/SSL enabled websites.


Unode:
Referencing #2479

Running `pip install scrapy` on a new virtualenv got me Twisted 17.1.0.
Quite a lot of HTTPS websites started to fail (e.g. google) with a twisted error `SSL23_GET_SERVER_HELLO - tlsv1 alert internal error`.

The main cause of the SSL23 error is that code introduced after #1794 on [core/downloader/tls.py](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/tls.py#L49) imports `_maybeSetHostNameIndication` which has been removed from [_sslverify.py](https://github.com/twisted/twisted/blob/twisted-17.1.0/src/twisted/internet/_sslverify.py).
This code is also wrapped on a large `try: ... except ImportError: pass` which silences the actual cause of the problem. 

Manually downgrading twisted to 16.0.0 solved the problem.
Also to avoid these incompatibilities shouldn't scrapy on pypi be limited to known working versions of dependencies?

```
Scrapy    : 1.3.1
lxml      : 3.7.2.0
libxml2   : 2.9.3
cssselect : 1.0.1
parsel    : 1.1.0
w3lib     : 1.17.0
Twisted   : 17.1.0
Python    : 3.4.5 (default, Jan 24 2017, 17:55:08) - [GCC 4.9.3]
pyOpenSSL : 16.2.0 (OpenSSL 1.0.2j  26 Sep 2016)
Platform  : Linux-4.6.0-sabayon-x86_64-Intel-R-_Core-TM-_i7-4710MQ_CPU_@_2.50GHz-with-gentoo-2.2
```

Rafalonso:
Yup, just noticed on a fresh install.

```
[scrapy.downloadermiddlewares.retry] DEBUG: Retrying <GET https://renegadeline.com/robots.txt> (failed 1 times): [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'SSL23_GET_SERVER_HELLO', 'tlsv1 alert internal error')]>]
```

This solution is not working https://github.com/scrapy/scrapy/issues/1429#issuecomment-131782133

pawelmhm:
[...]

PR #2556, opened 2017-02-13
requirements.txt remove Twisted 17 from supported versions

pawelmhm:
quick fix until we know what's going on and why https is failing for some websites. #2555


kmike:
Superseded by https://github.com/scrapy/scrapy/pull/2558. Thanks @pawelmhm for the heads up!

PR #2558, opened 2017-02-13
Fixed compatibility with twisted 17+

kmike:
I've replaced try/except with explicit version checks and ported code to be compatible with Twisted 17+. This should fix https://github.com/scrapy/scrapy/issues/2555.

For the record, issue was caused by https://twistedmatrix.com/trac/ticket/8916.

Also, we no longer support Ubuntu 12.04, so its tox environment is replaced with Ubuntu 14.04 environment (we still don't run it on Travis though).

### DIFF
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -1,6 +1,8 @@
 import logging
 from OpenSSL import SSL
 
+from scrapy import twisted_version
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


redapple:
Looks good to me. Thanks @kmike!

Le 13 févr. 2017 4:08 PM, "Mikhail Korobov" <notifications@github.com> a
écrit :

pawelmhm:
should we add more tests for https so that potential problems will be detected easier? Seems like our test suite passed ok with Twisted 17 so apparently our tests are not covering situation that occurred. I understand unit testing tls bugs is probably not easy, but maybe we could find some way.

PR #2561, opened 2017-02-13
Install Scrapy on Windows with Python 3

krysros:
Since there is no Twisted wheel file for Python 3.6 on PyPI, we need to install free Visual C++ 2015 Build Tools. Probably this step can be omitted in the future.

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -7,8 +7,7 @@ Installation guide
 Installing Scrapy
 =================
 
-Scrapy runs on Python 2.7 and Python 3.3 or above
-(except on Windows where Python 3 is not supported yet).
+Scrapy runs on Python 2.7 and Python 3.6 or above.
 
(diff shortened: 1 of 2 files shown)


docs/intro/install.rst
```diff
@@ -7,8 +7,7 @@ Installation guide
 Installing Scrapy
 =================
 
-Scrapy runs on Python 2.7 and Python 3.3 or above
-(except on Windows where Python 3 is not supported yet).
+Scrapy runs on Python 2.7 and Python 3.6.
```
kmike: I think we shouldn't drop Python < 3.6 support yet.

redapple:
Thanks @krysros .
The thing is that we are now recommending Anaconda for Windows users.
conda-forge has scrapy for Python 2.7, Python 3.5 and Python 3.6.
Hence I'm closing this.

Issue #1495, opened 2015-09-13
deprecate Spider.make_requests_from_url


kmike:
Hey,

`Spider.make_requests_from_url` is a shortcut for a single use case: sending initial requests from a spider with start_urls attribute. It saves a single line of code (`for url in start_urls`) in one specific use case, but yet we have to document it, users have to read about it and understand what it is for. 

The name suggests it is more general - but we can't yield Link objects and get requests created by `make_requests_from_url`, it does nothing for CrawlSpider (I'd expect URLs to be passed through this method), and it is ignored for SitemapSpider. This is inconsistent.

What about deprecating `make_requests_from_url` and removing it from docs? IMHO it will make API simpler, this hook does nothing useful now, and if we ever want a general 'process URL and get a Request' method we'd have to use another name for backwards compatibility anyways.

nramirezuy:
I agree; this method isn't that useful. Also now you even have a shorter import for `Request` if you wanna be lazy (:

chekunkov:
+1

PR #2562, opened 2017-02-14
[MRG+1] Separate building request from _requests_to_follow in CrawlSpider

terut:
I use scrapy-splash for javascript and sometimes want to change the request to another request  like splash one because it should be convenient using same logic for html after runing javascript.

scrapy-splash's request is following:
https://github.com/scrapy-plugins/scrapy-splash#examples

What do you think about that? Should I implement new spider?

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -48,6 +48,11 @@ class CrawlSpider(Spider):
     def process_results(self, response, results):
         return results
 
+    def _build_request(self, rule, link):
+        r = Request(url=link.url, callback=self._response_downloaded)
+        r.meta.update(rule=rule, link_text=link.text)
+        return r
+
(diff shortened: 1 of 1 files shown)


kmike:
This change makes sense to me. I guess the main question is whether we want this method to be public & documented & tested or not.

See also: https://github.com/scrapy/scrapy/issues/1495.

terut:
Thank you for your reply. I saw https://github.com/scrapy/scrapy/issues/1495.

I agree with you and I think this method should be private for now because someone who wants to change request object can override this method and `start_requests` method. That is reasonable.

Maybe it is better to have a public method to build many kinds of request object but that method would be too complicated because of many arguments.

kmike:
Makes sense. I think this change is fine if we look at it as a simple refactoring / code cleanup.

PR #2564, opened 2017-02-14
[MRG+1] Doc: binary mode is required for exporters

### LINKED ISSUE #2553: CsvItemExporter fails on py3
```
from scrapy.exporters import CsvItemExporter
with open('temp.csv', 'w') as f:
    exporter = CsvItemExporter(f)
    exporter.start_exporting()
    exporter.export_item({'a': 'b'})
    exporter.finish_exporting()
```

The previous snippet works fine on python 2, however the following error appears when running on python 3:

```
  File "exporter.py", line 5, in <module>
    exporter.export_item({'a': 'b'})
  File "/Users/eugenio/.../venv/lib/python3.5/site-packages/scrapy/exporters.py", line 206, in export_item
    self._write_headers_and_set_fields_to_export(item)
  File "/Users/eugenio/.../venv/lib/python3.5/site-packages/scrapy/exporters.py", line 230, in _write_headers_and_set_fields_to_export
    self.csv_writer.writerow(row)
TypeError: write() argument must be str, not bytes
```

Am I missing something? Thanks!

elacuesta:
Closes #2553

~I left `XmlItemExporter` out of the modifications, it's the only class which does not seem to have problems writing to either `bytes` or `str` streams (`io.BytesIO` or `io.StringIO`, for instance).~

```
from io import BytesIO, StringIO

with BytesIO(b'') as f:
    exporter = XmlItemExporter(f)
    exporter.start_exporting()
    exporter.export_item({'some': 'value'})
    exporter.finish_exporting()
    print(f.getvalue())

with StringIO(u'') as f:
    exporter = XmlItemExporter(f)
    exporter.start_exporting()
    exporter.export_item({'some': 'value'})
    exporter.finish_exporting()
    print(f.getvalue())
```
outputs:
```
b'<?xml version="1.0" encoding="utf-8"?>\n<items><item><some>value</some></item></items>'
<?xml version="1.0" encoding="utf-8"?>
<items><item><some>value</some></item></items>
```
/cc @kmike

### DIFF
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -225,7 +225,8 @@ XmlItemExporter
 
    Exports Items in XML format to the specified file object.
 
-   :param file: the file-like object to use for exporting the data.
+   :param file: the file-like object to use for exporting the data. Its ``write`` method should
+                accept ``bytes`` (a disk file opened in binary mode, a ``io.BytesIO`` object, etc)
 
    :param root_element: The name of root element in the exported XML.
    :type root_element: str
@@ -281,7 +282,8 @@ CsvItemExporter
    CSV columns and their order. The :attr:`export_empty_fields` attribute has
    no effect on this exporter.
 
-   :param file: the file-like object to use for exporting the data.
+   :param file: the file-like object to use for exporting the data. Its ``write`` method should
+                accept ``bytes`` (a disk file opened in binary mode, a ``io.BytesIO`` object, etc)
 
    :param include_headers_line: If enabled, makes the exporter output a header
       line with the field names taken from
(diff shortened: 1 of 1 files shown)


kmike:
While XmlItemExporter works with text files, I'm not sure it is intentional. For example, 'encoding' argument doesn't work properly with XmlItemExporter + text files.

docs/topics/exporters.rst
```diff
@@ -281,7 +281,8 @@ CsvItemExporter
    CSV columns and their order. The :attr:`export_empty_fields` attribute has
    no effect on this exporter.
 
-   :param file: the file-like object to use for exporting the data.
+   :param file: the file-like object to use for exporting the data. It's ``write`` method should
```
kmike: its

elacuesta: oops

elacuesta:
Right, the following:
```
from io import StringIO

with StringIO(u'') as f:
    exporter = XmlItemExporter(f, encoding='utf8')
    exporter.start_exporting()
    exporter.export_item({'some': u'valûe'})
    exporter.finish_exporting()
    print(f.getvalue())
```
Fails in python 2, it seems like the binary mode is more consistent. I'll add the note to the `XmlItemExporter` as well. Thanks!

kmike:

elacuesta:
Thanks to you and @TDominiqueWilliams :smile:

PR #2565, opened 2017-02-15
using custom request types for sitemap spider

codinguncut:
I am deriving from SitemapSpider for exploratory crawls and only want to see sitemap links, without necessarily following them.
For this reason it's important to be able to differentiate whether a Request is from a SitemapIndex or from a UrlSet.
I have derived (named) classes from Request to make this easily discoverable in derived classes and/or pipeline.

### DIFF
--- a/scrapy/spiders/sitemap.py
+++ b/scrapy/spiders/sitemap.py
@@ -9,6 +9,10 @@ from scrapy.utils.gz import gunzip, is_gzipped
 
 logger = logging.getLogger(__name__)
 
+# differentiating requests for use in child classes
+class SitemapIndexRequest(Request): pass
+class UrlSetRequest(Request): pass
+class RobotsRequest(Request): pass
 
 class SitemapSpider(Spider):
 
@@ -33,7 +37,7 @@ class SitemapSpider(Spider):
     def _parse_sitemap(self, response):
         if response.url.endswith('/robots.txt'):
             for url in sitemap_urls_from_robots(response.text, base_url=response.url):
-                yield Request(url, callback=self._parse_sitemap)
+                yield RobotsRequest(url, callback=self._parse_sitemap)
         else:
             body = self._get_sitemap_body(response)
             if body is None:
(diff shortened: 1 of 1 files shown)


kmike:
The feature sounds reasonable!

Previously the way to do it would be to add something to request.meta. But since https://github.com/scrapy/scrapy/pull/2510 (included in scrapy 1.3.2) custom request subclasses should also work. This PR can be a good chance to dogfood these custom request classes. So the implementation also looks fine to me.

Could you please add docs and tests for this feature? Existing docs: https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst and tests: https://github.com/scrapy/scrapy/blob/afac3fd2c28ac5b16e7e3c95fca5acd1334f8c58/tests/test_spider.py#L293

scrapy/spiders/sitemap.py
```diff
@@ -9,6 +9,10 @@
 
 logger = logging.getLogger(__name__)
 
+# differentiating requests for use in child classes
+class SitemapIndexRequest(Request): pass
+class UrlSetRequest(Request): pass
+class RobotsRequest(Request): pass
```
kmike: Could you please use boring pep8 style (`pass` on a new line, two lines between classes)? 

If you leave it as-is then over the next year or two someone will inevitably submit a pull request to fix code style; that'd be just a time waste for nothing :)

kmike:
yet another way to implement this: https://github.com/scrapy/scrapy/pull/2082

scrapy/spiders/sitemap.py
codinguncut: sorry, lost track of this PR. will do docs, test and pep8 compliance

codinguncut:
Hi,

I have rebased my branch, made code changes PEP8 compliant and added minimal documentation.

I think I messed up the branch/pull request because of the rebase from upstream scrapy. Maybe it would make sense to reopen this with a fresh branch/commit...

I was a bit unsure how to test this, and also since all I'm doing is to replace a class with it's child class I'm not sure how much testing is necessary at least for backward compatibility/regression.

I'd be grateful if you could merge this...

Johannes

PR #2566, opened 2017-02-15
[MRG+1] HttpErrorMiddleware stats

kmike:
Similar to https://github.com/scrapy/scrapy/pull/2543.

### DIFF
--- a/scrapy/spidermiddlewares/httperror.py
+++ b/scrapy/spidermiddlewares/httperror.py
@@ -46,6 +46,10 @@ class HttpErrorMiddleware(object):
 
     def process_spider_exception(self, response, exception, spider):
         if isinstance(exception, HttpError):
+            spider.crawler.stats.inc_value('httperror/response_ignored_count')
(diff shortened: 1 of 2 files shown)


scrapy/spidermiddlewares/httperror.py
```diff
@@ -46,6 +46,10 @@ def process_spider_input(self, response, spider):
 
     def process_spider_exception(self, response, exception, spider):
         if isinstance(exception, HttpError):
+            spider.crawler.stats.inc_value('httperror/response_dropped_count')
+            spider.crawler.stats.inc_value(
```
redapple: Maybe it should use the same "ignored" term as the subsequent INFO message ("Ignoring" is also in the `HttpError` exception message) instead of "dropped".
Besides that small nitpick, looks good to me.

kmike: @redapple updated, thanks for the suggestion!

PR #2572, opened 2017-02-17
[MRG+1] document issue with FormRequest.from_response due to bug in lxml

ashkulz:
This can make the spider fail due to incorrect values being posted server-side, which is extremely hard to debug because it is easy to miss leading/trailing whitespace, even with a logging proxy.

The fix was merged for lxml 3.8 in lxml/lxml#228 so document that as well.


kmike:
A good catch, thanks @ashkulz!

PR #2577, opened 2017-02-20
Set context factory implementation based on Twisted version

### LINKED ISSUE #2560: BrowserLikeContextFactory not available in some conditions
While tracing the error that lead to #2555 I tried the workaround [mentioned in the documentation](https://doc.scrapy.org/en/latest/topics/settings.html#downloader-clientcontextfactory) without success.

This code lives [incontextfactory.py](https://github.com/scrapy/scrapy/blob/c3411373e8a8ee2786588bdad7be469c69a25e2a/scrapy/core/downloader/contextfactory.py#L63) but was not reachable as the import was failing on my system due to #2555.

This file is a large `try/except` block with many potential points of failure and it's likely to trip other users in the future.
That said, could this be refactored to provide a fallback for `BrowserLikeContextFactory` or otherwise reduce the scope of the `try/except` to avoid breaking the API?

redapple:
Fixes #2560

### DIFF
--- a/scrapy/core/downloader/contextfactory.py
+++ b/scrapy/core/downloader/contextfactory.py
@@ -1,15 +1,15 @@
 from OpenSSL import SSL
 from twisted.internet.ssl import ClientContextFactory
 
-try:
+from scrapy import twisted_version
+
+if twisted_version >= (14, 0, 0):
 
     from zope.interface.declarations import implementer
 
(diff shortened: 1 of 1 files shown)


kmike:
Is it a fix for https://github.com/scrapy/scrapy/issues/2560?

redapple:
Indeed.

Issue #1612, opened 2015-11-23
scrapy LOG_LEVEL setting in Spider.custom_settings does not work


YAmikep:
I set the `LOG_LEVEL` setting to `INFO` in the Spider class via the `custom_settings` attribute but I still see the `DEBUG` messages in the console.

When I set it on the `settings.py` file or via the command line option `--loglevel`, it works.

I thought any settings could be set via the `custom_settings` attribute. Is that a bug? (Scrapy 1.0.3 and python 2.7.10)

```
class TestSpider(scrapy.Spider):
    name = "Test"
    ...
    custom_settings = {
        'LOG_LEVEL': 'INFO',
    }
```

jdemaeyer:
Looks like a bug, although support for setting `LOG_LEVEL` on the spider *might* be dropped in the future (#1580)

@curita Probably because we're using `settings` instead of `self.settings` [in this line](https://github.com/scrapy/scrapy/blob/master/scrapy/crawler.py#L37)?

nramirezuy:
I think changing `LOG_LEVEL` from the spider doesn't make too much sense anyways.

YAmikep:
@nramirezuy Could you elaborate on your answer? I'm new to scrapy. Why it does not make sense to set the logging level on a Spider? 
Where do you set it? On a project? What if I have different Spider?

Thanks

nramirezuy:
@YAmikep It doesn't make sense because there are 2 clear stages, development and production. During development you will probably want to use `DEBUG`, but for production you will want `INFO` or `WARNING`. This 2 should be 2 completely different environments, making the setting to have no sense.

What if you need to enable `DEBUG` on production for some reason; well then you can use command line to set the `LOG_LEVEL`.

YAmikep:
Thanks @nramirezuy . I see what you meant now.
I agree and actually, only the command line option makes sense. Changing it at the project setting does not make sense either.

nramirezuy:
@YAmikep I guess you are right. But you can cheat it using [`local_settings.py`](https://groups.google.com/forum/#!topic/scrapy-users/pO6BGH5vlII).

jackie2013:
@nramirezuy I set LOG_FILE setting in custom_settings for every Spider to get a different log, but it not work, what should i do? Thanks.

djunzu:
@nramirezuy , I agree with you: development (`DEBUG`) is different from production (`INFO`). But I am facing a particular case...

[...]

PR #2581, opened 2017-02-21
[MRG+1] Respect custom log level: fixes GH-1612

lopuhin:
See #1612: ~~this uses ``self.settings`` instead of ``settings``, as elsewhere in the crawler init.~~
Main change is that now a new root logger is installed when a crawler is created if one was already installed before. This allows to respect custom settings related to logging, such as LOG_LEVEL, LOG_FILE, etc.

As the result, in case of several spiders, logging settings from the last spider take precedence. Without this change, all logging settings in ``custom_settings`` are ignored. Both behaviours are not great, but it seems that in most cases there is just one spider running in the process, so the new behaviour seems less bad, but it's a tough call.

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -16,7 +16,9 @@ from scrapy.signalmanager import SignalManager
 from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.utils.ossignal import install_shutdown_handlers, signal_names
 from scrapy.utils.misc import load_object
-from scrapy.utils.log import LogCounterHandler, configure_logging, log_scrapy_info
+from scrapy.utils.log import (
+    LogCounterHandler, configure_logging, log_scrapy_info,
+    get_scrapy_root_handler, install_scrapy_root_handler)
 from scrapy import signals
 
 logger = logging.getLogger(__name__)
@@ -35,8 +37,11 @@ class Crawler(object):
         self.signals = SignalManager(self)
         self.stats = load_object(self.settings['STATS_CLASS'])(self)
 
-        handler = LogCounterHandler(self, level=settings.get('LOG_LEVEL'))
+        handler = LogCounterHandler(self, level=self.settings.get('LOG_LEVEL'))
         logging.root.addHandler(handler)
+        if get_scrapy_root_handler() is not None:
+            # scrapy root handler alread installed: update it with new settings
+            install_scrapy_root_handler(self.settings)
         # lambda is assigned to Crawler attribute because this way it is not
         # garbage collected after leaving __init__ scope
         self.__remove_handler = lambda: logging.root.removeHandler(handler)
--- a/tests/test_crawl.py
+++ b/tests/test_crawl.py
@@ -97,8 +97,8 @@ class CrawlTestCase(TestCase):
 
     @defer.inlineCallbacks
     def test_start_requests_bug_before_yield(self):
+        crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
         with LogCapture('scrapy', level=logging.ERROR) as l:
-            crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
             yield crawler.crawl(fail_before_yield=1)
 
         self.assertEqual(len(l.records), 1)
@@ -108,8 +108,8 @@ class CrawlTestCase(TestCase):
 
     @defer.inlineCallbacks
     def test_start_requests_bug_yielding(self):
+        crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
         with LogCapture('scrapy', level=logging.ERROR) as l:
-            crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
             yield crawler.crawl(fail_yielding=1)
 
         self.assertEqual(len(l.records), 1)
--- a/tests/test_crawl.py
+++ b/tests/test_crawl.py
@@ -97,8 +97,8 @@ class CrawlTestCase(TestCase):
 
     @defer.inlineCallbacks
     def test_start_requests_bug_before_yield(self):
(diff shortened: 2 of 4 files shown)


kmike:
The fix makes sense, but I wonder what happens if there are two spiders in the same process (executed via CrawlerRunner), and one of them changes LOG_LEVEL - does it affect the other spider?

lopuhin:
Yes, I don't think you can have two spiders with different log levels in one process, and I don't immediately see a backwards-compatible fix for that.

djunzu:
I guess the last spider to start will override any previous log level.
In this case I think a warning note should be somewhere in the docs saying it.

kmike:
@djunzu I'd expect it to still follow log level for this spider, at least if ``spider`` is in context. This is not always possible though because some messages are global (without ``spider`` in context). But if we document current behavior then maybe we'll documet a bug; I'm not sure. Anyways, it looks like a different issue.

djunzu:
@kmike I decided to test it just to know what happens. I tried
```python
# settings.py:
LOG_LEVEL = 'WARNING'

#testspider.py:
    name = "Test"
    custom_settings = {
        'LOG_LEVEL': 'DEBUG',
    }

# output just with warning messages
```
and
```python
# settings.py:
LOG_LEVEL = 'DEBUG'

#testspider.py:
    name = "Test"
    custom_settings = {
        'LOG_LEVEL': 'WARNING',
    }

# output with a lot of debug messages
```
and
```python
# settings.py:
#LOG_LEVEL = 'DEBUG'

#testspider.py:
    name = "Test"
    custom_settings = {
        'LOG_LEVEL': 'WARNING',
    }

# output with a lot of debug messages (default LOG_LEVEL)
```
None of them works. It does not solve the problem and it is still not possible to set log level using `custom_settings` attribute. (This or I did something very wrong!)

tested with scrapy 1.3.2

@lopuhin Did you do any other test than the one you wrote?

lopuhin:
@djunzu yes, I did check that this pull request solves the issue for me (specifically, I was using ``scrapy runspider`` command), and I checked that the test I wrote fails without this PR. Let me also check the ``scrapy crawl`` command.

lopuhin:
@djunzu you are right, this is my mistake. Honestly, I'm not sure how did I test it and get what I expected. Thanks for checking! I'll see if I can fix it.
The test also needs updating: it just check log stats (which do respect log level in custom_settings), but not the actual log output.

lopuhin:
The tests are green finally, yay! @kmike @djunzu do you mind having another look at this? I updated the PR description with summary of changes and also checked that ``scrapy crawl`` and ``scrapy runspider`` commands both respect custom log settings.

scrapy/crawler.py
```diff
@@ -35,8 +35,9 @@ def __init__(self, spidercls, settings=None):
         self.signals = SignalManager(self)
         self.stats = load_object(self.settings['STATS_CLASS'])(self)
 
-        handler = LogCounterHandler(self, level=settings.get('LOG_LEVEL'))
+        handler = LogCounterHandler(self, level=self.settings.get('LOG_LEVEL'))
         logging.root.addHandler(handler)
+        configure_logging(self.settings)
```
kmike: Hm, previously this function was only called by Scrapy commands, not by Crawler. I think we should check carefully that calling it in Crawler doesn't affect usage examples in https://doc.scrapy.org/en/latest/topics/logging.html#scrapy.utils.log.configure_logging.

lopuhin: Right, thanks for the catch! Probably it's wrong to call it unconditionally when starting any crawler - when you are using scrapy as a library you'd like to be in control of the logging.  Initially I added this call to support other log settings as well, but now I see that this is problematic. Maybe it's better to change scrapy root handler only if it's already installed.

lopuhin:
Now a new root logger is installed when a crawler is created only if one was already installed before.

expansion:
[https://mailguru.s3.amazonaws.com/images/spark/qr.v2/935cf1c585724824a8e35714de5e0627.png]

谢谢！



Spark by Readdle<https://itunes.apple.com/app/id997102246?mt=8&uo=4&at=10l6UJ&ct=Spark_QR_viral>

        Spark by Readdle<https://itunes.apple.com/app/id997102246?mt=8&uo=4&at=10l6UJ&ct=Spark_QR_viral>

tests/test_crawl.py
```diff
@@ -97,8 +97,8 @@ def test_retry_dns_error(self):
 
     @defer.inlineCallbacks
     def test_start_requests_bug_before_yield(self):
+        crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
         with LogCapture('scrapy', level=logging.ERROR) as l:
-            crawler = self.runner.create_crawler(BrokenStartRequestsSpider)
```
kmike: what is the reason for this change?

lopuhin: I made that changes because a new root log handler was installed when crawler was created, and so LogCapture did not work (and many other tests also have crawler creation outside of LogCapture). But it looks like it is no longer required, at least when the tests are run in isolation. I revered this changes in 6f55ca4, let's see if the full test suite passes.

lopuhin: Yep, the tests passed, probably it's better to leave it as it was in master. Thanks for the catch!

kmike:
Looks good to me, thanks @lopuhin for the fix and @djunzu for the review.

djunzu:
:+1: 


Back to my initial comment:
I think that docs should have this note: "in case of several spiders, logging settings from the last spider take precedence".

@kmike, regarding
> But if we document current behavior then maybe we'll documet a bug

I think documented bugs are good: everyone knows what to expect. Undocumented bugs are bad: when you are not expecting it, something blows up in front of you!

Nonetheless, +1 for merge.

Issue #2586, opened 2017-02-23
DataLoss error when content-length doesn't match actual body size


rolando:
The example website below returns a `Content-Length` header larger than the actual response size. The webserver itself is broken (or misconfigured) because it returns the `Content-Length` from the uncompressed body while it sends the compressed body.

```
$ scrapy shell "http://www.ebk-gruppe.com/"  --set RETRY_ENABLED=0
2017-02-23 00:26:18 [scrapy.utils.log] INFO: Scrapy 1.3.2 started (bot: scrapybot)
2017-02-23 00:26:18 [scrapy.utils.log] INFO: Overridden settings: {'LOGSTATS_INTERVAL': 0, 'RETRY_ENABLED': '0', 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter'}
2017-02-23 00:26:18 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.corestats.CoreStats',
 'scrapy.extensions.memusage.MemoryUsage']
2017-02-23 00:26:19 [scrapy.middleware] INFO: Enabled downloader middlewares:
['scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
 'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
 'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
 'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
 'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
 'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
 'scrapy.downloadermiddlewares.cookies.CookiesMiddleware',
 'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware',
 'scrapy.downloadermiddlewares.stats.DownloaderStats']
2017-02-23 00:26:19 [scrapy.middleware] INFO: Enabled spider middlewares:
['scrapy.spidermiddlewares.httperror.HttpErrorMiddleware',
 'scrapy.spidermiddlewares.offsite.OffsiteMiddleware',
 'scrapy.spidermiddlewares.referer.RefererMiddleware',
[...]

Issue #2513, opened 2017-01-25
twisted error in log


whwq2012:
> 2017-01-19 14:41:21 [twisted] CRITICAL: Unhandled Error
> Traceback (most recent call last):
> Failure: twisted.internet.error.ConnectionDone: Connection was closed cleanly.

This is error, I check my code, and I think the error that from scrapy

scrapy version:1.30
os:centos7.0
python version: 2.7.11

redapple:
@whwq2012 , thanks for reporting.
In fact, this error is raised at Twisted HTTP state machine level.
I've seen this happen when the webservers do not send as many bytes as they should ([this is a recent example](https://github.com/scrapy/scrapy/pull/2493#issuecomment-272294062))
Does it happen with the same website(s) all the time?
If yes, would you mind sharing some of the URL for which you see these `twisted.internet.error.ConnectionDone` messages?
A priori, it's a web server issue, but there might be something else.

whwq2012:
@redapple 
1、There is the only bug in log, and it repeat 268 times.
2、The error happen  when my spider have ran 2 days.
3、if you need I can proive my spider code.

whwq2012:
@redapple This is my error log. [https://drive.google.com/open?id=0B7iXisqaWzlENHFjT2ZHQzlOY0k](url).
The log contain two type of error, one is the twisted error, another is redis auth error.

redapple:
@whwq2012 , you probably ran your crawl with LOG_LEVEL=ERROR
Unfortunately, this does not give enough information on what website you are accessing.
If you can run with LOG_LEVEL=DEBUG , it may give more information.
Your spider code may help. I will probably not run it as-is though.
I do not know if running with your custom stats collector using Redis has any influence.

whwq2012:
@redapple OK, I try to use LOG_LEVEL=DEBUG. But please give me some time. Because my server have no transfer until next month. I will run spider in 2.1. About 2.3~2.6. I will you send you log

redapple:
@whwq2012 , I don't know what you mean by 2.1, 2.3 and 2.6

whwq2012:
@redapple it means February 1st, February 3rd, February 6th. Sorry My English is poor

redapple:
Understood.

whwq2012:
@redapple here is log. https://drive.google.com/file/d/0B7iXisqaWzlEZEd6dTlSSlNqS1U/view?usp=sharing

redapple:
[...]

PR #2493, opened 2017-01-12
Fix issue where redirects with cleanly closed connections would not be followed

dan-blanchard:
Some servers, like `m.huffpost.com` send a 30X redirect response but close the connection after they've assumed you've read the headers. This causes scrapy to raise a Twisted `ResponseFailed` exception. We catch these exceptions in the `RedirectMiddleware`, take the headers out of the response (since that much is there), create a new scrapy `Response` object out of it, and return that for processing as usual.

At Parse.ly we routinely crawl about 100 pages per second 24 hours per day, and we found that we were hitting this issue literally thousands of times per day.

Also, I think it should be added to the documentation for the `RedirectMiddleware` that you have to set `HTTPERROR_ALLOWED_CODES = [301, 302, 303, 307]` in your settings to allow it to work. Otherwise you just raise `HttpError` exceptions and no redirects occur.

Issue #2413, opened 2016-11-25
Chunked decoder in 'CHUNK_LENGTH' state, still expecting more data to get to 'FINISHED' state


duangy:
Traceback (most recent call last):
  File "/home/env/lib/python3.4/site-packages/twisted/internet/defer.py", line 1183, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/env/lib/python3.4/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/env/lib/python3.4/site-packages/scrapy/core/downloader/middleware.py", line 43, in process_request
    defer.returnValue((yield download_func(request=request,spider=spider)))
twisted.web._newclient.ResponseFailed: [<twisted.python.failure.Failure twisted.internet.error.ConnectionDone: Connection was closed cleanly.>, <twisted.python.failure.Failure twisted.web.http._DataLoss: Chunked decoder in 'CHUNK_LENGTH' state, still expecting more data to get to 'FINISHED' state.>]

Did Anyone meet the same problem， and have any solution?

redapple:
Hi @duangy , I don't think this has been reported before.
Is it systematic? can you provide some sample code reproducing the issue?

redapple:
Hi @duangy , do you have any updates on this issue?

rolando:
This is related to https://github.com/scrapy/scrapy/issues/2586

kmike:
@rolando it should be fixed by https://github.com/scrapy/scrapy/pull/2590, right?

rolando:
@kmike yes, this is the case of the broken chunked response.

hcoura:
I don't know if it is related, but I am commenting here due to the error been the same.
The example is on the repo below, it only happens when I have a USER_AGENT set.

https://github.com/hcoura/forbes_scrapy_error

Although I manage to get my production code working through overriding the request header when crawling forbes it felt kinda odd, so I decided to share

kmike:
[...]

PR #2590, opened 2017-02-23
[MRG+2] Handle data loss gracefully.

rolando:
Websites that return a wrong ``Content-Length`` header may cause a data
loss error. Also when a chunked response is not finished properly.

This change adds a new setting ``DOWNLOAD_FAIL_ON_DATALOSS`` (default: ``True``) and request.meta key ``download_fail_on_dataloss``.

---

Related issues:

* https://github.com/scrapy/scrapy/issues/2513
* https://github.com/scrapy/scrapy/issues/2586
* https://github.com/scrapy/scrapy/pull/2493
* https://github.com/scrapy/scrapy/issues/2413

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -604,6 +604,32 @@ If you want to disable it set to 0.
 
     This feature needs Twisted >= 11.1.
 
+.. setting:: DOWNLOAD_FAIL_ON_DATALOSS
+
+DOWNLOAD_FAIL_ON_DATALOSS
+-------------------------
+
+Default: ``True``
+
+Whether or not to fail on broken responses, that is, declared
+``Content-Length`` does not match content sent by the server or chunked
+response was not properly finish. If ``True``, these responses raise a
+``ResponseFailed([_DataLoss])`` error. If ``False``, these responses
+are passed through and the flag ``dataloss`` is added to the response, i.e.:
+``'dataloss' in response.flags`` is ``True``.
+
+Optionally, this can be set per-request basis by using the
+:reqmeta:`download_fail_on_dataloss` Request.meta key to ``False``.
+
+.. note::
+
+  A broken response, or data loss error, may happen under several
+  circumstances, from server misconfiguration to network errors to data
+  corruption. It is up to the user to decide if it makes sense to process
+  broken responses considering they may contain partial or incomplete content.
+  If setting:`RETRY_ENABLED` is ``True`` and this setting is set to ``True``,
+  the ``ResponseFailed([_DataLoss])`` failure will be retried as usual.
+
 .. setting:: DUPEFILTER_CLASS
 
 DUPEFILTER_CLASS
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -12,9 +12,9 @@ from twisted.internet import defer, reactor, protocol
 from twisted.web.http_headers import Headers as TxHeaders
 from twisted.web.iweb import IBodyProducer, UNKNOWN_LENGTH
 from twisted.internet.error import TimeoutError
-from twisted.web.http import PotentialDataLoss
+from twisted.web.http import _DataLoss, PotentialDataLoss
 from twisted.web.client import Agent, ProxyAgent, ResponseDone, \
-    HTTPConnectionPool
+    HTTPConnectionPool, ResponseFailed
 from twisted.internet.endpoints import TCP4ClientEndpoint
 
 from scrapy.http import Headers
@@ -51,13 +51,15 @@ class HTTP11DownloadHandler(object):
             warnings.warn(msg)
         self._default_maxsize = settings.getint('DOWNLOAD_MAXSIZE')
         self._default_warnsize = settings.getint('DOWNLOAD_WARNSIZE')
+        self._fail_on_dataloss = settings.getbool('DOWNLOAD_FAIL_ON_DATALOSS')
         self._disconnect_timeout = 1
 
     def download_request(self, request, spider):
         """Return a deferred for the HTTP download"""
         agent = ScrapyAgent(contextFactory=self._contextFactory, pool=self._pool,
             maxsize=getattr(spider, 'download_maxsize', self._default_maxsize),
-            warnsize=getattr(spider, 'download_warnsize', self._default_warnsize))
+            warnsize=getattr(spider, 'download_warnsize', self._default_warnsize),
+            fail_on_dataloss=self._fail_on_dataloss)
         return agent.download_request(request)
 
     def close(self):
@@ -233,13 +235,14 @@ class ScrapyAgent(object):
     _TunnelingAgent = TunnelingAgent
 
     def __init__(self, contextFactory=None, connectTimeout=10, bindAddress=None, pool=None,
-                 maxsize=0, warnsize=0):
+                 maxsize=0, warnsize=0, fail_on_dataloss=True):
         self._contextFactory = contextFactory
         self._connectTimeout = connectTimeout
         self._bindAddress = bindAddress
         self._pool = pool
         self._maxsize = maxsize
         self._warnsize = warnsize
+        self._fail_on_dataloss = fail_on_dataloss
         self._txresponse = None
 
     def _get_agent(self, request, timeout):
@@ -326,6 +329,7 @@ class ScrapyAgent(object):
         maxsize = request.meta.get('download_maxsize', self._maxsize)
         warnsize = request.meta.get('download_warnsize', self._warnsize)
         expected_size = txresponse.length if txresponse.length != UNKNOWN_LENGTH else -1
+        fail_on_dataloss = request.meta.get('download_fail_on_dataloss', self._fail_on_dataloss)
 
         if maxsize and expected_size > maxsize:
             error_msg = ("Cancelling download of %(url)s: expected response "
@@ -345,7 +349,8 @@ class ScrapyAgent(object):
             txresponse._transport._producer.loseConnection()
 
         d = defer.Deferred(_cancel)
-        txresponse.deliverBody(_ResponseReader(d, txresponse, request, maxsize, warnsize))
+        txresponse.deliverBody(_ResponseReader(
+            d, txresponse, request, maxsize, warnsize, fail_on_dataloss))
 
         # save response for timeouts
         self._txresponse = txresponse
@@ -380,13 +385,16 @@ class _RequestBodyProducer(object):
 
 class _ResponseReader(protocol.Protocol):
 
-    def __init__(self, finished, txresponse, request, maxsize, warnsize):
+    def __init__(self, finished, txresponse, request, maxsize, warnsize,
+                 fail_on_dataloss):
         self._finished = finished
         self._txresponse = txresponse
         self._request = request
         self._bodybuf = BytesIO()
         self._maxsize  = maxsize
         self._warnsize  = warnsize
+        self._fail_on_dataloss = fail_on_dataloss
+        self._fail_on_dataloss_warned = False
         self._reached_warnsize = False
         self._bytes_received = 0
 
@@ -415,7 +423,22 @@ class _ResponseReader(protocol.Protocol):
         body = self._bodybuf.getvalue()
         if reason.check(ResponseDone):
             self._finished.callback((self._txresponse, body, None))
-        elif reason.check(PotentialDataLoss):
+            return
+
+        if reason.check(PotentialDataLoss):
             self._finished.callback((self._txresponse, body, ['partial']))
-        else:
-            self._finished.errback(reason)
+            return
+
+        if reason.check(ResponseFailed) and any(r.check(_DataLoss) for r in reason.value.reasons):
+            if not self._fail_on_dataloss:
+                self._finished.callback((self._txresponse, body, ['dataloss']))
+                return
+
+            elif not self._fail_on_dataloss_warned:
+                logger.warn("Got data loss in %s. If you want to process broken "
+                            "responses set the setting DOWNLOAD_FAIL_ON_DATALOSS = False"
+                            " -- This message won't be shown in further requests",
+                            self._txresponse.request.absoluteURI.decode())
+                self._fail_on_dataloss_warned = True
+
+        self._finished.errback(reason)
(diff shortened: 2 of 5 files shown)


rolando:
This fixes #2586

tests/test_downloader_handlers.py
```diff
@@ -341,6 +373,34 @@ def test_download_with_large_maxsize_per_spider(self):
...
+                if any(r.check(_DataLoss) for r in failure.value.reasons):
+                    return None
+            return failure
+
+        d.addCallback(lambda _: self.fail("No DataLoss exception"))
+        d.addErrback(checkDataLoss)
+        return d
+
+    def test_download_broken_content_allow_data_loss(self):
+        request = Request(self.getURL('broken'), meta={'download_allow_dataloss': True})
```
redapple: Can you add a similar test when using the DOWNLOAD_ALLOW_DATALOSS setting?

redapple:
Looks good @rolando , thanks.
The new setting needs to be documented.

tests/test_downloader_handlers.py
rolando: Added test via setting here https://github.com/scrapy/scrapy/pull/2590/files#diff-d6a30363eaf8467303835b3adce0af7dR404

redapple:
Looks good. Thanks @rolando !

docs/topics/settings.rst
```diff
@@ -604,6 +604,26 @@ If you want to disable it set to 0.
...
+does not match content sent by the server. If ``False``, these responses
+trigger a ``ResponseFailed`` error. If ``True``, these responses are allowed
+and the flag ``partial`` is added.
+
+Optionally, this can be set per-request basis by using the
+:reqmeta:`download_allow_dataloss` Request.meta key.
+
+.. note::
+
+  This feature needs Twisted >= 11.1 and the use of the default HTTP/1.1
```
kmike: Scrapy minimum supported Twisted version is 13.1

rolando:
@paul forgot to mention that I added a warning to hint the use of the setting, which is shown before the error:

```
scrapy shell "http://www.ebk-gruppe.com/"  --set RETRY_ENABLED=0  --loglevel WARN
2017-02-28 14:08:29 [scrapy.core.downloader.handlers.http11] WARNING: Found data loss in http://www.ebk-gruppe.com/. If you want to process broken responses set the setting DOWNLOAD_ALLOW_DATALOSS = True
  File "/Users/rolando/miniconda3/envs/dev/bin/scrapy", line 11, in <module>
    load_entry_point('Scrapy', 'console_scripts', 'scrapy')()
  File "/Users/rolando/Projects/sh/scrapy/scrapy/cmdline.py", line 142, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/Users/rolando/Projects/sh/scrapy/scrapy/cmdline.py", line 88, in _run_print_help
    func(*a, **kw)
  File "/Users/rolando/Projects/sh/scrapy/scrapy/cmdline.py", line 149, in _run_command
    cmd.run(args, opts)
  File "/Users/rolando/Projects/sh/scrapy/scrapy/commands/shell.py", line 73, in run
    shell.start(url=url, redirect=not opts.no_redirect)
  File "/Users/rolando/Projects/sh/scrapy/scrapy/shell.py", line 48, in start
    self.fetch(url, spider, redirect=redirect)
  File "/Users/rolando/Projects/sh/scrapy/scrapy/shell.py", line 115, in fetch
    reactor, self._schedule, request, spider)
  File "/Users/rolando/miniconda3/envs/dev/lib/python3.5/site-packages/twisted/internet/threads.py", line 122, in blockingCallFromThread
    result.raiseException()
  File "/Users/rolando/miniconda3/envs/dev/lib/python3.5/site-packages/twisted/python/failure.py", line 372, in raiseException
    raise self.value.with_traceback(self.tb)
twisted.web._newclient.ResponseFailed: [<twisted.python.failure.Failure twisted.internet.error.ConnectionDone: Connection was closed cleanly.>, <twisted.python.failure.Failure twisted.web.http._DataLoss: >]
```

docs/topics/settings.rst
rolando: @kmike Do you think can we remove the note there? I copied that from one setting above.

kmike: Yes, it looks safe to remove (here and in the setting above).

docs/topics/settings.rst
```diff
@@ -604,6 +604,26 @@ If you want to disable it set to 0.
...
+
+DOWNLOAD_ALLOW_DATALOSS
+-----------------------
+
+Default: ``False``
+
+Whether or not to allow broken responses, where declared ``Content-Length``
+does not match content sent by the server. If ``False``, these responses
+trigger a ``ResponseFailed`` error. If ``True``, these responses are allowed
+and the flag ``partial`` is added.
```
kmike: There are now request flags and response flags; I think it'd be nice to have a concrete example on how to check it - something like

> these responses are allowed and the flag ``partial`` is added to the response, i.e. `'partial' in response.flags` is True.

docs/topics/settings.rst
rolando: Removed note.

docs/topics/settings.rst
```diff
@@ -604,6 +604,26 @@ If you want to disable it set to 0.
...
+.. setting:: DOWNLOAD_ALLOW_DATALOSS
+
+DOWNLOAD_ALLOW_DATALOSS
+-----------------------
+
+Default: ``False``
+
+Whether or not to allow broken responses, where declared ``Content-Length``
+does not match content sent by the server. If ``False``, these responses
+trigger a ``ResponseFailed`` error. If ``True``, these responses are allowed
```
kmike: Are ResponseFailed resposes retried by default? It'd be helpful to mention that - i.e. what happens if there is data loss detected?

docs/topics/settings.rst
rolando: The flag is not exclusive to this condition though. Legit ``PotentialDataLoss`` responses which were not dropped have the ``partial`` flag. I was wondering if this condition ``_DataLoss`` should have its own flag, like ``data-loss``.

docs/topics/settings.rst
kmike: Would it be useful to have an option to retry these responses, but keep the last one if retry attempts failed? Based on your experience, can retries fix this data loss problem, or is it usually just a server configuration issue?

docs/topics/settings.rst
```diff
@@ -604,6 +604,21 @@ If you want to disable it set to 0.
 
     This feature needs Twisted >= 11.1.
 
+.. setting:: DOWNLOAD_ALLOW_DATALOSS
+
+DOWNLOAD_ALLOW_DATALOSS
```
kmike: The name is tricky :) Arguably we allow data loss when we don't pass partial content to the callback and drop the whole response instead. I don't have a better naming idea though.

kmike: Data loss at network level happens regardless of Scrapy options, this isn't something we can control

docs/topics/settings.rst
rolando: Yes, ``ResponseFailed`` is handled by the retry middleware. And I think there are several reasons why a ``ResponseFailed`` can be raised, hence the ``_DataLoss`` explicit check. I don't know if it can be a legit scenario where a response declares ``Content-Length: X`` and suddenly the connection is closed after less than ``X`` bytes transmitted (i.e.: server crashed?). In that case a retrying the request looks like a correct behavior.

docs/topics/settings.rst
rolando: > I don't have a better naming idea though.

Neither did I :)

docs/topics/settings.rst
kmike: Was there a 'partial' flag for such responses before this PR?

kmike: I doubt these flags were widely used, but renaming 'partial' to 'data-loss' for this case is backwards incompatible (if the answer to previous question is "yes"); I think that's fine to do this change, but it must be mentioned in release notes.

rolando: > Was there a 'partial' flag for such responses before this PR?
 
Yes, when a ``PotentialDataLoss`` error was raised by twisted, scrapy added the flag ``partial`` and let the response pass through.

I have updated the doc with the ``'partial'`` check example.

docs/topics/settings.rst
rolando: > Would it be useful to have an option to retry these responses, but keep the last one if retry attempts failed? 

Seems like this feature needs to be discussed on its own issue.

> can retries fix this data loss problem, or is it usually just a server configuration issue?

There are several scenarios where this can happe. For example, if the server returns a static file, sets the ``Content-Length`` correctly but fails to read a part of the file due to data corruption or disk failure. The common misconfiguration case seems to be when the application uses a content compression filter which is enabled after the ``Content-Length`` is set.

We can't tell how often a retry fixed this issue unless we mine past logs.

kmike: I'm asking because the best default value for the option is not clear for me. False makes sense because it mimics existing behavior, True makes sense if the problem is usually caused by server misconfiguration, retry+True sounds like a compromise which could work better as a default in practice because it handles both cases, but I'm not really sure.

kmike: But I agree that it can be a separate issue :)

docs/topics/settings.rst
kmike: Some ideas on how to avoid ambiguous "data loss" name:

* `DOWNLOAD_VALIDATE_CONTENT_LENGTH`
* `DOWNLOAD_DROP_PARTIAL`

scrapy/core/downloader/handlers/http11.py
```diff
@@ -415,7 +423,21 @@ def connectionLost(self, reason):
...
+            return
+
+        if reason.check(ResponseFailed) and any(r.check(_DataLoss) for r in reason.value.reasons):
+            if self._allow_dataloss:
+                self._finished.callback((self._txresponse, body, ['partial']))
+                return
+
+            elif not self._allow_dataloss_hinted:
+                logger.warn("Found data loss in %s. If you want to process broken "
+                            "responses set the setting DOWNLOAD_ALLOW_DATALOSS = True",
```
kmike: I think it worths adding a note that the message won't be shown for further requests, like we do in dupefilters.

kmike:
The PR looks good overall 👍 Todo:
* I still don't quite like 'data loss' term here because it is ambiguous; I think we should find a better option name.
* It'd be nice to explain in docs how this feature interacts with retries.
* A nitpick about warn message
* decide on 'partial' vs 'data-loss'; I'm fine with both.

docs/topics/settings.rst
rolando: Maybe ``DOWNLOAD_FAIL_ON_DATALOSS``, which describes the current behavior: throw an error when ``ResponseFailed(reasons=[_DataLoss])`` happens (or any ``ResponseFailed`` error for that matter).

rolando:
> I still don't quite like 'data loss' term here because it is ambiguous; I think we should find a better option name.

A new candidate name: ``DOWNLOAD_FAIL_ON_DATALOSS``. This describes the current behavior.

> It'd be nice to explain in docs how this feature interacts with retries.

Please check current setting documentation.

> A nitpick about warn message

Added a mention that no more messages will be shown.

> decide on 'partial' vs 'data-loss' flag; I'm fine with both.

I arbitrarily changed it to ``dataloss`` (as most flags are single-word).

docs/topics/settings.rst
kmike: This `_DataLoss` reason which Twisted returns doesn't necessarily means data loss in practice, right? If it can mean something else then there is no reason to follow this naming convention and call it 'data loss', we can do better. That's why I tried to avoid it in these name ideas. Other than that `DOWNLOAD_FAIL_ON_DATALOSS` sounds fine, it is not ambiguous.

rolando: Yes, a ``_DataLoss`` reason in twisted means we did not get the amount of data we expected. And that's up to the transfer decoder in twisted.

The [chunked decoder](https://github.com/twisted/twisted/blob/e8e8e8a58634e058a9f7887c9e4621e70438be8e/src/twisted/web/http.py#L1725) raises a ``_DataLoss`` exception when there is no terminating chunk.

The [identity decoder](https://github.com/twisted/twisted/blob/e8e8e8a58634e058a9f7887c9e4621e70438be8e/src/twisted/web/http.py#L1591) (non-chunked) raises a ``PotentialDataLoss`` error when there is no ``content-length`` declared and a ``_DataLoss`` error when the received content length does not match the declared ``content-length``.

kmike: Aha, so DOWNLOAD_VALIDATE_CONTENT_LENGTH is not a good name because for chunked decoder it means a different thing.

rolando:
Updated the setting name to ``DOWNLOAD_FAIL_ON_DATALOSS`` (default: ``True``).

kmike:
LGTM. Thanks @rolando!

rolando:
:+1:

tests/test_downloader_handlers.py
```diff
@@ -118,6 +120,35 @@ def render(self, request):
         return request.requestHeaders.getRawHeaders(b"content-length")[0]
 
 
+class ChunkedResource(resource.Resource):
```
redapple: I understand this acts as a valid server response using chunked encoding. Is that correct?
Would it be possible to add a case for a wrong chunked-encoded response?

rolando: That's a bit tricky but I added a broken chunked resource with accompanying tests: https://github.com/scrapy/scrapy/pull/2590/files#diff-d6a30363eaf8467303835b3adce0af7dR134

It relies on a internal `request.chunked` flag which, if its purpose changes in further twisted releases, it will break the tests. I verified the behavior by using tcpdump:

* This is a normal chunked response

```
127.000.000.001.60134-127.000.000.001.60217: HTTP/1.1 200 OK
Transfer-Encoding: chunked
Content-Type: text/html
Server: TwistedWeb/16.6.0
Date: Wed, 01 Mar 2017 14:25:57 GMT

8
chunked
8
content

0

```

* This is a broken chunked response:
```
127.000.000.001.60720-127.000.000.001.60754: HTTP/1.1 200 OK
Transfer-Encoding: chunked
Content-Type: text/html
Date: Wed, 01 Mar 2017 14:30:42 GMT
Server: TwistedWeb/16.6.0

8
chunked
8
content


```

redapple: Looks good @rolando . Thanks!

Issue #2589, opened 2017-02-23
Headers.to_unicode_dict() fails with badly encoded values


rolando:
Here is an example:

```
n [21]: response.headers
{b'Accept-Ranges': b'bytes',
 b'Cache-Control': b'max-age=604800',
 b'Content-Type': b'text/html',
 b'Date': b'Wed, 22 Feb 2017 04:03:39 GMT',
 b'Etag': b'"801764d9a9acf1:0"',
 b'Last-Modified': b'Mon, 06 Jan 2014 06:38:03 GMT',
 b'Public-Key-Pins': b'pin-sha256=\x94SPKI_digest#1"; pin-sha256="SPKI_digest#2"; max-age=31536000',
 b'Server': b'Microsoft-IIS/7.5',
 b'Strict-Transport-Security': b'max-age=31536000; includeSubDomains',
 b'Vary': b'Accept-Encoding,Accept-Encoding',
 b'X-Powered-By': b'ASP.NET',
 b'X-Powered-By-Plesk': b'PleskWin'}

In [22]: response.headers.to_unicode_dict()
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-22-259fb34b3193> in <module>()
----> 1 response.headers.to_unicode_dict()

<env>/lib/python3.5/site-packages/scrapy/http/headers.py in to_unicode_dict(self)
     87             (to_unicode(key, encoding=self.encoding),
     88              to_unicode(b','.join(value), encoding=self.encoding))
---> 89             for key, value in self.items())
     90
     91     def __copy__(self):

<env>/lib/python3.5/site-packages/scrapy/utils/datatypes.py in __init__(self, seq)
    191         super(CaselessDict, self).__init__()
    192         if seq:
--> 193             self.update(seq)
    194
    195     def __getitem__(self, key):

<env>/lib/python3.5/site-packages/scrapy/utils/datatypes.py in update(self, seq)
    227         seq = seq.items() if isinstance(seq, dict) else seq
    228         iseq = ((self.normkey(k), self.normvalue(v)) for k, v in seq)
--> 229         super(CaselessDict, self).update(iseq)
    230
    231     @classmethod

<env>/lib/python3.5/site-packages/scrapy/utils/datatypes.py in <genexpr>(.0)
    226     def update(self, seq):
    227         seq = seq.items() if isinstance(seq, dict) else seq
[...]

Issue #2592, opened 2017-02-23
Improving headers values decoding (utf-8 vs latin1)


rmax:
This raises from https://github.com/scrapy/scrapy/issues/2589 where a server returns a non-UTF-8 header value. According to [this RFC](https://tools.ietf.org/html/rfc7230#section-3.2.4):

>    Historically, HTTP has allowed field content with text in the
   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
   through use of [RFC2047] encoding.  In practice, most HTTP header
   field values use only a subset of the US-ASCII charset [USASCII].
   Newly defined header fields SHOULD limit their field values to
   US-ASCII octets.  A recipient SHOULD treat other octets in field
   content (obs-text) as opaque data.

So the `latin1` encoding may be preferred over `utf-8` which is the Scrapy's default.  At the other hand, `requests` in some places uses `utf-8` with a fallback to `latin1`: https://github.com/kennethreitz/requests/blob/d6f4818c0b40bc6c00433c013b7daaea83b2cd51/requests/models.py#L908

Following @kmike's [suggestion](https://github.com/scrapy/scrapy/pull/2588#issuecomment-282112022), it seems that decoding headers with `utf-8` having `iso-8859-1` as fallback is the best option.

rolando:
See PR https://github.com/scrapy/scrapy/pull/2593 as a proposed implementation. The nice thing about this is that, when dealing with non-UTF8 strings, it doesn't require user intervention as with #2588.

wRAR:
#5917 was related, it seems (though it only touched conbtent-type specific code paths)

PR #2593, opened 2017-02-23
[MRG+1] Decode header values with UTF-8 and ISO-8859-1 as fallback.

rolando:
In the example below, all headers except `Public-Key-Pins` are UTF-8 compatible:

```
In [3]: response.headers

{b'Accept-Ranges': b'bytes',
 b'Cache-Control': b'max-age=604800',
 b'Content-Type': b'text/html',
 b'Date': b'Thu, 23 Feb 2017 22:21:35 GMT',
 b'Etag': b'"3acb6d9a9acf1:0"',
 b'Last-Modified': b'Mon, 06 Jan 2014 06:38:03 GMT',
 b'Public-Key-Pins': b'pin-sha256=\x94SPKI_digest#1"; pin-sha256="SPKI_digest#2"; max-age=31536000',
 b'Server': b'Microsoft-IIS/7.5',
 b'Strict-Transport-Security': b'max-age=31536000; includeSubDomains',
 b'Vary': b'Accept-Encoding',
 b'X-Powered-By': b'ASP.NET',
 b'X-Powered-By-Plesk': b'PleskWin'}

In [4]: response.headers.to_unicode_dict()

{'accept-ranges': 'bytes',
 'cache-control': 'max-age=604800',
 'content-type': 'text/html',
 'date': 'Thu, 23 Feb 2017 22:21:35 GMT',
 'etag': '"3acb6d9a9acf1:0"',
 'last-modified': 'Mon, 06 Jan 2014 06:38:03 GMT',
 'public-key-pins': 'pin-sha256=\x94SPKI_digest#1"; pin-sha256="SPKI_digest#2"; max-age=31536000',
 'server': 'Microsoft-IIS/7.5',
 'strict-transport-security': 'max-age=31536000; includeSubDomains',
 'vary': 'Accept-Encoding',
 'x-powered-by': 'ASP.NET',
 'x-powered-by-plesk': 'PleskWin'}
```

### DIFF
--- a/scrapy/http/headers.py
+++ b/scrapy/http/headers.py
@@ -7,7 +7,19 @@ from scrapy.utils.python import to_unicode
 class Headers(CaselessDict):
     """Case insensitive http headers dictionary"""
 
-    def __init__(self, seq=None, encoding='utf-8'):
+    def __init__(self, seq=None, encoding=None):
+        """Initialize headers object.
+
+        Parameters
+        ----------
+        seq : items, optional
+            Headers items.
+        encoding : str, optional
+            Encoding used for encoding/decoding header values. If not given,
+            default for encoding is UTF-8 and for decoding UTF-8 with
+            ISO-8859-1 fallback.
+
+        """
         self.encoding = encoding
         super(Headers, self).__init__(seq)
 
@@ -26,13 +38,14 @@ class Headers(CaselessDict):
 
         return [self._tobytes(x) for x in value]
 
-    def _tobytes(self, x):
+    def _tobytes(self, x, default_encoding='utf-8'):
+        encoding = self.encoding or default_encoding
         if isinstance(x, bytes):
             return x
         elif isinstance(x, six.text_type):
-            return x.encode(self.encoding)
+            return x.encode(encoding)
         elif isinstance(x, int):
-            return six.text_type(x).encode(self.encoding)
+            return six.text_type(x).encode(encoding)
         else:
             raise TypeError('Unsupported value type: {}'.format(type(x)))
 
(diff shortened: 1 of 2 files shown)


rolando:
This address https://github.com/scrapy/scrapy/issues/2592 an fixes https://github.com/scrapy/scrapy/issues/2589

rolando:
ping @redapple @kmike

redapple:
Looks good to me.

As a side note, and probably for another PR, I had a need for getting the Unicode value of a header for Referrer-Policy in https://github.com/scrapy/scrapy/pull/2306/files#diff-30f505d1ef277d884613e1042c761aa7R326 and I am not particularly happy with specifying the encoding. `Headers`  could benefit from a `.get_unicode(header_key)` method of some sort.

kmike:
I'm not fond of fallbacks because one can silently get data corrupt - e.g. if a header value can be decoded using utf-8, while in fact it is a non-ascii latin1. Fallback is also less deterministic - one can end up with unicode headers where the same header value is decoded using several encodings for different pages of a website. Because of that I initially suggested using latin1 by default, and using utf-8 for a small set of known headers like cookie headers. But I wonder what are browsers doing, is there a de-facto standard?

rolando:
@kmike found this related changes in webkit's changelog:

> New in 2.7.3
>  - Use latin1 instead of UTF-8 for HTTP header values.
> ...
> New in 1.11.1
> - Fix a crash in network backend with non-UTF8 HTTP header names.

I'm in favor of using `latin1` as default, but that may be a backwards incompatible change.

rolando:
@kmike the fallback idea came from requests: https://github.com/scrapy/scrapy/issues/2592#issue-209880064

kmike:
The linked requests code uses utf8 -> latin1 fallback to decode non-ascii message parts of status codes; it sounds fine to do this because one hardly ever cares about these non-ascii messages, and because they are website bugs, not something standard (standard is to use a fixed set of ascii-only names).

As for backwards compatibility, I'm not sure we should care that much - this method is not mentioned in scrapy docs, and it is only used for some of the s3 uploads. If s3 uploads need headers decoded from utf8 (e.g. because s3 encodes them back to utf8) then it should use another method, it can be s3-specific.

rolando:
Closing in favor of https://github.com/scrapy/scrapy/pull/2637

PR #2595, opened 2017-02-25
Add omitted "self" arguments

Lodour:
“self” argument is omitted in some methods.


kmike:
A good catch, thanks!

PR #2596, opened 2017-02-25
[MRG+1] Use single quotes uniformly

arvindch:
The default spider template mixes single and double quotes. This fixes that.


lopuhin:
Thanks @arvindch !

PR #2599, opened 2017-02-27
Fix FTP downloader and re-enable FTP tests on Python 3

redapple:
FTP support on Python 3 is available again with Twisted 17.1+

### DIFF
--- a/tests/test_downloader_handlers.py
+++ b/tests/test_downloader_handlers.py
@@ -687,9 +687,6 @@ class BaseFTPTestCase(unittest.TestCase):
     password = "passwd"
     req_meta = {"ftp_user": username, "ftp_password": password}
 
-    if six.PY3:
-        skip = "Twisted missing ftp support for PY3"
-
     def setUp(self):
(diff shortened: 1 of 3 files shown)


tests/test_downloader_handlers.py
```diff
@@ -687,9 +687,6 @@ class BaseFTPTestCase(unittest.TestCase):
     password = "passwd"
     req_meta = {"ftp_user": username, "ftp_password": password}
 
-    if six.PY3:
-        skip = "Twisted missing ftp support for PY3"
```
kmike: Do you think it still should be skipped for unsupported Twisted versions in Python 3? Or do we support only latest Twisted in Python 3?

redapple: The current tox/Travis use the latest Twisted version when Python 3 is used.
I don't think we need to skip the FTP tests, but we can do it if you think it's cleaner.

kmike: Ok, sounds fine! Thanks for the fix.

Issue #2604, opened 2017-02-28
log path to http cache


kmike:
What do you think about logging a full path to HTTP cache when HttpCacheMiddleware is instantiated? It may help to debug issues like https://github.com/scrapy/scrapy/issues/2601.

redapple:
Sounds useful indeed.

jorenham:
Hi, we're a team of four Computer Science Master students. For a course we have to contribute to an open-source project, and Scrapy looks really awesome to all of us.
So we would like to work is this one, is that ok?
@Eauwzeauw
@Threore
@jojona

kmike:
@jorenham sounds great!

redapple:
Closed via https://github.com/scrapy/scrapy/issues/2611

PR #2611, opened 2017-03-02
[MRG+1] [logging] HttpCacheMiddleware: log cache directory at instantiation

jorenham:
#2604

### DIFF
--- a/scrapy/downloadermiddlewares/httpcache.py
+++ b/scrapy/downloadermiddlewares/httpcache.py
@@ -1,3 +1,5 @@
+import logging
+
 from email.utils import formatdate
 from twisted.internet import defer
 from twisted.internet.error import TimeoutError, DNSLookupError, \
@@ -7,6 +9,9 @@ from twisted.web.client import ResponseFailed
 from scrapy import signals
 from scrapy.exceptions import NotConfigured, IgnoreRequest
 from scrapy.utils.misc import load_object
+
+
+logger = logging.getLogger(__name__)
 
 
 class HttpCacheMiddleware(object):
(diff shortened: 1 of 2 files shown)


scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -24,6 +29,8 @@ def __init__(self, settings, stats):
         self.ignore_missing = settings.getbool('HTTPCACHE_IGNORE_MISSING')
         self.stats = stats
 
+        logger.debug("Using cache directory %(cachedir)s" % {'cachedir': self.storage.cachedir})
```
kmike: All built-in cache storages have 'cachedir' attribute, but it is not guaranteed for third-party cache storages. I think it could be better to either check that this attribute is present, or (better) move logging to storages. The latter will allow to log not only directory, but also a file name - for some storages cache is stored in a single file.

scrapy/extensions/httpcache.py
```diff
@@ -216,6 +220,8 @@ def __init__(self, settings):
         self.dbmodule = import_module(settings['HTTPCACHE_DBM_MODULE'])
         self.db = None
 
+        logger.debug("Using DBM cache storage in %(cachedir)s" % {'cachedir': self.cachedir})
```
kmike: What do you think about logging exact filename of the database used for cache, not only a folder containing these databases? See open_spider method.

jorenham:
@kmike I applied your suggestions. But did not manage to also log the temp file FilesystemCacheStorage. This is because `_get_request_path` takes `spider` and `request` as arguments, which aren't accessible in `__init__`.

kmike:
Unlike other storages FilesystemCacheStorage uses multiple files, so it makes sense to log a folder for it. But for cache storages which use a single file I think it is better to log a path to this file.

jorenham:
@kmike So if I understand you correctly I should move the log in `DbmCacheStorage` and `LeveldbCacheStorage` from `__init__` to `open_spider` and include the `dbpath` in the message. Is that correct?

kmike:
@jorenham yes, I think it makes sense. For consistency logging can be moved to open_spider method for FilesystemCacheStorage as well, so that they all log a message at the same time point.

jorenham:
@kmike I've moved the logging and it now logs the entire paths when available

kmike:
Looks good to me, thanks!

Issue #2449, opened 2016-12-14
Retry downloading in media pipeline


guohengkai:
How to implement retrying in media pipeline? Thanks.

redapple:
Hi @guohengkai , indeed there are issues with handling redirections (at least).
See https://github.com/scrapy/scrapy/issues/2004 and a patch proposal https://github.com/scrapy/scrapy/pull/2193

redapple:
@guohengkai , my bad, I confused this issue with the redirect handling.

peterlupu:
for the retry, maybe this will help someone

## The Pipeline:

```py
class MyImagesPipeline(FilesPipeline):

    def get_media_requests(self, item, info):
        return [scrapy.Request(x, meta={'is_media': True}) for x in item.get(self.files_urls_field, [])]
```

notice I set meta on the request is_media = True

## Retry Middleware

```py
class RetryMediaMiddleware:
    """
    retry failed media requests (status>=400)
    maximum retries = 10
    """

    def process_response(self, request, response, spider):
        is_media = request.meta.get('is_media', False)
        if not is_media:
            return response

        if response.status >= 400:
            retries = request.meta.get('retry_times', 0) + 1
            if retries > 10:
                return response
            retryreq = request.copy()
            retryreq.meta['retry_times'] = retries
            retryreq.dont_filter = True
            retryreq.priority = request.priority + 2
            return retryreq
        return response
```

there're a few hardcoded values, but it's good enough for the concept

AdrianAtZyte:
> Seem that the DownloaderMiddleware does not work for pipeline.

Media pipeline requests are downloaded through the standard downloader, so the downloader middlewares and the download slots do apply to them, and they did back when you asked as well. That means `RETRY_ENABLED`, `RETRY_TIMES`, `DOWNLOAD_DELAY`, the concurrency settings and AutoThrottle all work with no extra code; only the spider middlewares are skipped. It is [in the docs](https://docs.scrapy.org/en/latest/topics/media-pipeline.html), on step 3 of the typical workflow.

A quick check with `RETRY_TIMES = 2` and a `file_urls` entry pointing at a URL that always returns a 503:

```
"downloader/request_count": 4,
"downloader/response_status_count/200": 1,
"downloader/response_status_count/503": 3,
"retry/count": 2,
"retry/reason_count/503 Service Unavailable": 2,
"retry/max_reached": 1
```

[...]

PR #2616, opened 2017-03-03
[MRG] Allow redirections in media files downloads

### LINKED ISSUE #2004: MediaPipeline (and ImagesPipeline/FilesPipeline) does not handle HTTP redirections

redapple:
Continuation of #2193
Fixes #2004 and #2449

Compared to #2193, I chose to remove `MEDIA_HTTPSTATUS_LIST`: I don't think this level of detail on status codes is needed for redirections. Debatable.


tests/test_pipeline_media.py
```diff
@@ -249,3 +257,61 @@ def test_use_media_to_download_result(self):
         self.assertEqual(new_item['results'], [(True, 'ITSME')])
         self.assertEqual(self.pipe._mockcalled, \
                 ['get_media_requests', 'media_to_download', 'item_completed'])
+
+
+class MediaPipelineAllowRedirectSettingsTestCase(unittest.TestCase):
```
kmike: How hard would it be to have an integration-style test for this?

redapple: I'll have a go at it.

redapple: @kmike , I added some resources to MockServer to list files and redirect, and used similar tests to `tests/test_crawl.py`

kmike:
Looks good, thanks @redapple!

kmike:
@redapple how does it fix #2449?

redapple:
@kmike , looks like I misread #2449. It has nothing to do with redirects.

lopuhin:
@kmike @redapple what do you think of adding ``MEDIA_ALLOW_REDIRECTS = True`` to the new project template, so that it's on by default for new projects?

redapple:
@lopuhin , sounds fine to me.

lopuhin:
@redapple thanks, I'll make a PR!

lopuhin:
Ah, although I see that media pipeline is not enabled in the sample project settings, so it might seem strange to have a setting that does nothing... And it's already documented, so someone who will be enabling this pipeline will see this setting too.

PR #2617, opened 2017-03-03
Remove redundant slot.add_request() call in ExecutionEngine

redapple:
The `slot.add_request(request)` in `download()` is also done in `_download()` which is called right after.


kmike:
👍

PR #2622, opened 2017-03-05
[MRG+1] Abort connection earlier and avoid to buffer data when max size limit is reached

### LINKED ISSUE #1616: Big download is not being cancelled properly
I am running scrapy 1.0.1 in my pyhon web crawler and setting the max download size like this: 

`DOWNLOAD_MAXSIZE = 41943040 #40MB`.

When I run into a bigger file, scrappy keeps logging the error: `ERROR: Received (51316256) bytes larger than download max size (41943040)` but never stops. I have looked into the code and saw a call to cancel the processe [here](https://github.com/scrapy/scrapy/blob/1.0.1/scrapy/core/downloader/handlers/http11.py#L306), but in my case is not working, the download keeps going on.

Any thoughts on why this would happen ? It seems to be something wrong on twisted not cancelling the download.

Detailed version info:

```
scrapy version -v
2015-11-25 10:44:54 [scrapy] INFO: Scrapy 1.0.1 started (bot:)
2015-11-25 10:44:54 [scrapy] INFO: Optional features available: ssl, http11
2015-11-25 10:44:54 [scrapy] INFO: Overridden settings: {'COOKIES_DEBUG': True, 'DOWNLOAD_TIMEOUT': 600, 'SPIDER_MODULES': ['spiders'], 'CONCURRENT_REQUESTS': 10, 'RANDOMIZE_DOWNLOAD_DELAY': False, 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter', 'RETRY_TIMES': 10, 'BOT_NAME': 'bot', 'DOWNLOAD_MAXSIZE': 41943040, 'USER_AGENT': 'Mozilla/5.0 (X11; Linux i686; rv:23.0) Gecko/20100101 Firefox/23.0', 'NEWSPIDER_MODULE': 'spiders'}
Scrapy  : 1.0.1
lxml    : 3.5.0.0
libxml2 : 2.9.2
Twisted : 15.4.0
Python  : 2.7.10 (default, Aug 13 2015, 12:27:27) - [GCC 4.9.2]
Platform: Linux-3.16.0-38-generic-x86_64-with
```

rolando:
Fixes #1616 

A symptom of this issue was having the log message "Received (X) bytes
larger than download max size (Y)" several times printed, with increased
X values.

```
2017-03-04 09:55:58	ERROR	[scrapy.core.downloader.handlers.http11] Received (3538944) bytes larger than download max size (524288).
2017-03-04 09:55:58	ERROR	[scrapy.core.downloader.handlers.http11] Received (3542211) bytes larger than download max size (524288).
2017-03-04 09:55:58	ERROR	[scrapy.core.downloader.handlers.http11] Received (3542212) bytes larger than download max size (524288).
2017-03-04 09:55:58	ERROR	[scrapy.core.downloader.handlers.http11] Received (3547136) bytes larger than download max size (524288).
2017-03-04 09:55:58	ERROR	[scrapy.core.downloader.handlers.http11] Received (3550280) bytes larger than download max size (524288).
```

This may lead to high memory consumption in under high concurrency settings.

### DIFF
--- a/tests/test_downloader_handlers.py
+++ b/tests/test_downloader_handlers.py
@@ -177,6 +177,16 @@ class EmptyContentTypeHeaderResource(resource.Resource):
         return request.content.read()
 
 
+class LargeChunkedFileResource(resource.Resource):
+    def render(self, request):
+        def response():
+            for i in range(1024):
+                request.write(b"x" * 1024)
+            request.finish()
+        reactor.callLater(0, response)
+        return server.NOT_DONE_YET
+
+
 class HttpTestCase(unittest.TestCase):
 
     scheme = 'http'
@@ -202,6 +212,7 @@ class HttpTestCase(unittest.TestCase):
         r.putChild(b"broken-chunked", BrokenChunkedResource())
         r.putChild(b"contentlength", ContentLengthHeaderResource())
         r.putChild(b"nocontenttype", EmptyContentTypeHeaderResource())
+        r.putChild(b"largechunkedfile", LargeChunkedFileResource())
         r.putChild(b"echo", Echo())
         self.site = server.Site(r, timeout=None)
         self.wrapper = WrappingFactory(self.site)
@@ -383,6 +394,25 @@ class Http11TestCase(HttpTestCase):
 
         d = self.download_request(request, Spider('foo', download_maxsize=9))
         yield self.assertFailure(d, defer.CancelledError, error.ConnectionAborted)
+
+    @defer.inlineCallbacks
+    def test_download_with_maxsize_very_large_file(self):
+        with mock.patch('scrapy.core.downloader.handlers.http11.logger') as logger:
+            request = Request(self.getURL('largechunkedfile'))
+
+            def check(logger):
+                logger.error.assert_called_once_with(mock.ANY, mock.ANY)
+
+            d = self.download_request(request, Spider('foo', download_maxsize=1500))
+            yield self.assertFailure(d, defer.CancelledError, error.ConnectionAborted)
+
+            # As the error message is logged in the dataReceived callback, we
+            # have to give a bit of time to the reactor to process the queue
+            # after closing the connection.
+            d = defer.Deferred()
+            d.addCallback(check)
+            reactor.callLater(.1, d.callback, logger)
+            yield d
 
     @defer.inlineCallbacks
     def test_download_with_maxsize_per_req(self):
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -348,7 +348,8 @@ class ScrapyAgent(object):
                            {'size': expected_size, 'warnsize': warnsize})
 
         def _cancel(_):
-            txresponse._transport._producer.loseConnection()
+            # Abort connection inmediately.
+            txresponse._transport._producer.abortConnection()
 
         d = defer.Deferred(_cancel)
         txresponse.deliverBody(_ResponseReader(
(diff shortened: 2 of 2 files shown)


tests/test_downloader_handlers.py
```diff
@@ -372,6 +380,13 @@ def test_download_with_maxsize(self):
         yield self.assertFailure(d, defer.CancelledError, error.ConnectionAborted)
 
     @defer.inlineCallbacks
+    def test_download_with_maxsize_very_large_file(self):
+        request = Request(self.getURL('largechunkedfile'))
+
+        d = self.download_request(request, Spider('foo', download_maxsize=1500))
+        yield self.assertFailure(d, defer.CancelledError, error.ConnectionAborted)
```
kmike: to clarify: is this test failing before the changes, but pass after them?

rolando: This test is incomplete at this time. I'm thinking in checking the calls to the logger or there received bytes in the response reader. It shouldn't pass certain threshold which depends on the chunks size.

I think there still a little room for improvement. Twisted keeps a buffer of the data received until the time it can be passed to the client protocol, see:

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/_newclient.py#L1094
https://github.com/twisted/twisted/blob/trunk/src/twisted/web/_newclient.py#L1130

So, if the reactor is busy enough, it can hold several chunks of data which then are passed at once to our response reader causing the repeated log messages.

rolando:
@redapple @kmike This should be ready for review.

When the fix is not applied, the test fails as follows:

```
trial tests.test_downloader_handlers.Http11TestCase.test_download_with_maxsize_very_large_file
tests.test_downloader_handlers
  Http11TestCase
    test_download_with_maxsize_very_large_file ...                      [ERROR]

===============================================================================
[ERROR]
  File "/Users/rolando/Projects/gh/twisted/src/twisted/internet/defer.py", line 1299, in _inlineCallbacks
  File "/Users/rolando/Projects/gh/twisted/src/twisted/python/failure.py", line 393, in throwExceptionIntoGenerator
  File "/Users/rolando/Projects/sh/scrapy/tests/test_downloader_handlers.py", line 399, in test_download_with_maxsize_very_large_file
    yield d
  File "/Users/rolando/Projects/gh/twisted/src/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/Users/rolando/Projects/sh/scrapy/tests/test_downloader_handlers.py", line 388, in check
    logger.error.assert_called_once_with(mock.ANY, mock.ANY)
  File "/Users/rolando/miniconda3/envs/dev/lib/python3.5/unittest/mock.py", line 802, in assert_called_once_with
    raise AssertionError(msg)
builtins.AssertionError: Expected 'error' to be called once. Called 63 times.
```

redapple:
Looks good to me @rolando . Thanks.

tests/test_downloader_handlers.py
```diff
@@ -177,6 +177,13 @@ def render(self, request):
         return request.content.read()
 
 
+class LargeChunkedFileResource(resource.Resource):
+    def render(self, request):
+        for i in range(1024):
+            request.write(b"\x00" * 1024)
+        return b""
```
dangra: I think it should be `request.finish()` instead of `return b""` even if it doesn't make a difference for the test case.

rolando: IIRC, I tried several variants of this render method and using request.finish didn't behave same as returning an empty string.

redapple: @dangra , are you ok with keeping `render` as @rolando implemented?

dangra: yes and no, I would like to understand in what case it makes a difference. But it is not relevant to the test case so I am fine if you want to move forward and merge it.

rolando: Tweaked the render method to use an approach similar to the other chunked resource.

rolando:
PR rebased.

redapple:
@dangra , what do you think of the PR now?

PR #2627, opened 2017-03-06
Fix referrer policy from response headers and support explicit empty string

### LINKED ISSUE #2626: lots of warnings `RuntimeWarning: Could not load referrer policy ''` when running tests
Check `tox -e py36 -- tests/test_crawl.py -s` output. @redapple is it a testing issue, or a problem with the warning? I've first noticed it in https://travis-ci.org/scrapy/scrapy/jobs/208185821 output.

redapple:
Fixes #2626

### DIFF
--- a/scrapy/spidermiddlewares/referer.py
+++ b/scrapy/spidermiddlewares/referer.py
@@ -261,6 +261,9 @@ _policy_classes = {p.name: p for p in (
     DefaultReferrerPolicy,
 )}
 
+# Reference: https://www.w3.org/TR/referrer-policy/#referrer-policy-empty-string
+_policy_classes[''] = NoReferrerWhenDowngradePolicy
+
 
 def _load_policy_class(policy, warning_only=False):
     """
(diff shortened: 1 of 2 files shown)


kmike:
Looks good! Just to be sure - is there a test which checks that a default policy is used if no policy is set, and not e.g this new policy for ''?

redapple:
I'm not sure I understand the question. But:
- [`TestRefererMiddlewareDefault`](https://github.com/scrapy/scrapy/blob/7e8453cf1ec992e5df5cebfeda08552c58e7c9bc/tests/test_spidermiddleware_referer.py#L290) tests that Scrapy's default policy is used when there's no overriden setting, and no request meta for referrer and no response referrer policy in headers
- if there's a `Referrer-Policy` header in the response, and no request meta, the HTTP header value from response, even if it's the empty string, takes precedence ; this is tested with [this new test here in this PR](https://github.com/scrapy/scrapy/pull/2627/files#diff-b2477fd1894a79e7bd77d1fea15b1105R529)

Let me know if you meant another use-case.

kmike:
Thanks Paul for the explanation!

PR #2630, opened 2017-03-06
Release notes for 1.4.0

redapple:


kmike:
It looks good, but I'd love to see a bit more tutorial-like content in these release notes, highlighting a couple of new features. For example, it could explain motivation behind response.follow and maybe give an example of a correct code before and after, and give a high-level overview of bug fixes, before diving in details. I can write response.follow section if you want.

redapple:
Sounds good. PRs to the release notes branch welcome.

Le 23 avr. 2017 12:27 PM, "Mikhail Korobov" <notifications@github.com> a
écrit :

redapple:
@kmike , what features would you like to highlight apart from `response.follow`?

redapple:
I would suggest highlighting these:
- FTP on Python 3 (with latest Twisted) and anonymous FTP support
- default link extractor now stripping leading and trailing attribute whitespace
- pretty-printing JSON and XML item exports 
- response.follow

@kmike , what do you think?

kmike:
Sounds good!

> default link extractor now stripping leading and trailing attribute whitespace

All linkextractors do this, not only the default, and FormRequest also has a similar fix. I was thinking about whitespace and canonicalize fixes as of a single batch of changes to make sure links are extracted properly from HTML attributes. `response.follow` is also a step in this direction, though it could deserve a separate mention.

redapple:
@kmike , @stummjr , I added an intro before the list of fixed issues and merged pull requests.
Feel free to edit! (the branch is not in my fork)

docs/news.rst
```diff
@@ -3,6 +3,148 @@
...
+
+There's a new ``response.follow()`` shortcut for creating URLs directly
+from the response instance in callbacks.
+For example, instead of::
+
+    scrapy.Request(response.urljoin(somehrefvalue))
+
+you can now use the simpler::
+
+    response.follow(somehrefvalue)
```
stummjr: Does it return a `Request` object? The description above may lead people to think that it just creates the URL from the `Response` instance (which is what `response.urljoin` does).

What about:
> There's a new ``response.follow()`` shortcut for creating requests directly from a response instance and a relative URL.

redapple: Updated @stummjr . Looks good/better?

stummjr: Yup! :)

docs/news.rst
```diff
@@ -3,6 +3,155 @@
...
+do. Especially, leading and trailing whitespace are removed from attributes
+(think ``href="   http://example.com"``) when building ``Link`` objects.
+This whitespace-stripping also happens for ``action`` attributes with
+``FormRequest``.
+**Please also note that link extractors do not canonicalize URLs by default
+anymore.** This was puzzling users every now and then, and it's not what
+browsers do in fact, so we removed that extra transformation on extractred
+links.
+
+There's a new ``response.follow()`` shortcut for creating requests directly
```
kmike: We could link to https://doc.scrapy.org/en/master/intro/tutorial.html#response-follow-example or https://doc.scrapy.org/en/master/topics/request-response.html#scrapy.http.TextResponse.follow

docs/news.rst
```diff
@@ -3,6 +3,155 @@
...
+from a response instance and a relative URL.
+For example, instead of::
+
+    scrapy.Request(response.urljoin(somehrefvalue))
```
kmike: ``response.follow`` a bit more advanced than that:

1. scrapy.Request(response.urljoin(href)) is not correct for non-ascii hrefs if page encoding is not utf-8; response.follow handles that case. 
2. response.follow supports not only relative URLs, but absolute URLs as well; it also supports some of selectors, so in a lot of cases there is no need to call `.extract()`. For `<a>` elements it extracts their href values, so no need to write `:attr(href)` as well.

So instead of this:

```py
for href in response.css('li.next a::attr(href)').extract():
    url = response.urljoin(href)
    yield response.follow(url, callback=self.parse, encoding=response.encoding)
```

One can write this:

```py
for a in response.css('li.next a'):
    yield response.follow(a, callback=self.parse)
```

The original motivation was to make it easy to write correct link extraction code; obvious `scrapy.Request(href)` is not correct (href can be relative, encoding is set incorrectly), while `response.follow(href)` is just as short, but handles these cases, and as an extra bonus it works with some of selectors directly, not only with hrefs as strings. I see it as a general-purpose scrapy.Request(href) replacement in spider code which allows to avoid bugs and save some typing.

redapple:
Thanks for the feedback @kmike.
Can I ask you to push your suggestions to the release notes branch? It's
probably easier than back and forth with edits by me.
Thank you.

docs/news.rst
redapple: @kmike , maybe the example above could go into https://doc.scrapy.org/en/master/topics/request-response.html#scrapy.http.Response.follow (which does not currently say much on the support for passing selectors),
and we keep the release notes shorter with a link to the requests/responses doc page?

kmike: Response.follow doesn't allow passing Selectors; Response.follow docs mention that TextResponse.follow alllows that, and https://doc.scrapy.org/en/master/topics/request-response.html#scrapy.http.TextResponse.follow mentions selectors and other stuff.

I think a verbose notice in the changelog is worth adding, as it is a huge change in recommended Scrapy API - starting with Scrapy 1.4, instead of scrapy.Request in most cases users should use response.follow, as it is more correct.

redapple: Alright then

dangra: the example pasted by @kmike in first comment on this thread is perfect to show how `request.follow` can improve spider code a lot.

kmike:
@redapple I've updated release notes with a response.follow promotion. How does it look now?

PR #2637, opened 2017-03-09
Set default encoding to latin1

### LINKED ISSUE #2592: Improving headers values decoding (utf-8 vs latin1)

rmax:
Fixes #2592.
See also #2589, #2593.

### DIFF
--- a/scrapy/http/headers.py
+++ b/scrapy/http/headers.py
@@ -7,7 +7,7 @@ from scrapy.utils.python import to_unicode
 class Headers(CaselessDict):
     """Case insensitive http headers dictionary"""
 
-    def __init__(self, seq=None, encoding='utf-8'):
+    def __init__(self, seq=None, encoding='latin1'):
         self.encoding = encoding
         super(Headers, self).__init__(seq)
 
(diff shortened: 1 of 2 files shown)


rolando:
This potentially is a backward incompatible change.

Issue #2642, opened 2017-03-10
allow to set RETRY_TIMES per-request


kmike:
I think it'd be nice to allow set RETRY_TIMES per-request using a `request.meta` key ("max_retry_times"?) - some requests could be more important than others.

atultherajput:
I am a newbie to open source. I know Python programming language. Please guide me to fix this issue.

saurabhgoyaliitr:
is this still open  to try?

redapple:
hi @saurabhgoyaliitr , do you mean you want to work on fixing it?
https://github.com/scrapy/scrapy/issues/2643 by @harshasrinivas has already done most of the work I believe.

redapple:
Fixed in https://github.com/scrapy/scrapy/issues/2643

PR #2643, opened 2017-03-11
[MRG+1] Add feature to set RETRY_TIMES per request (#2642)

harshasrinivas:
Allows the user to set `RETRY_TIMES` per request using `request.meta` key `max_retry_times`.
Recommended in #2642

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -852,6 +852,11 @@ Default: ``2``
 
 Maximum number of times to retry, in addition to the first download.
 
+.. reqmeta:: max_retry_times
+
+If :attr:`Request.meta <scrapy.http.Request.meta>` has ``max_retry_times`` key
+set to some value, this setting will be ignored by this middleware for the corresponding request.
+
 .. setting:: RETRY_HTTP_CODES
 
 RETRY_HTTP_CODES
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -852,10 +852,10 @@ Default: ``2``
 
 Maximum number of times to retry, in addition to the first download.
 
-.. reqmeta:: max_retry_times
-
-If :attr:`Request.meta <scrapy.http.Request.meta>` has ``max_retry_times`` key
-set to some value, this setting will be ignored by this middleware for the corresponding request.
+Maximum number of retries can also be specified per-request using
+:reqmeta:`max_retry_times` attribute of :attr:`Request.meta <scrapy.http.Request.meta>`.
+When initialized, the :reqmeta:`max_retry_times` meta key takes higher
+precedence over the :setting:`RETRY_TIMES` setting.
 
 .. setting:: RETRY_HTTP_CODES
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -308,6 +308,7 @@ Those are:
 * ``ftp_user`` (See :setting:`FTP_USER` for more info)
 * ``ftp_password`` (See :setting:`FTP_PASSWORD` for more info)
 * :reqmeta:`referrer_policy`
+* :reqmeta:`max_retry_times`
 
 .. reqmeta:: bindaddress
 
@@ -341,6 +342,13 @@ download_fail_on_dataloss
 
 Whether or not to fail on broken responses. See:
 :setting:`DOWNLOAD_FAIL_ON_DATALOSS`.
+
+.. reqmeta:: max_retry_times
+
+max_retry_times
+---------------
+
+The meta key is used set retry times per request. When initialized, the :setting:`RETRY_TIMES` setting will be ignored by the downloader middleware.
 
 .. _topics-request-response-ref-request-subclasses:
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -348,7 +348,9 @@ Whether or not to fail on broken responses. See:
 max_retry_times
 ---------------
 
-The meta key is used set retry times per request. When initialized, the :setting:`RETRY_TIMES` setting will be ignored by the downloader middleware.
+The meta key is used set retry times per request. When initialized, the
+:reqmeta:`max_retry_times` meta key takes higher precedence over the
+:setting:`RETRY_TIMES` setting.
 
 .. _topics-request-response-ref-request-subclasses:
 
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -63,6 +63,9 @@ class RetryMiddleware(object):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
+        if 'max_retry_times' in request.meta:
+            self.max_retry_times = request.meta['max_retry_times']
+
         stats = spider.crawler.stats
         if retries <= self.max_retry_times:
             logger.debug("Retrying %(request)s (failed %(retries)d times): %(reason)s",
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -63,7 +63,10 @@ class RetryMiddleware(object):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
-        retry_times = request.meta.get('max_retry_times') or self.max_retry_times
+        retry_times = self.max_retry_times
+
+        if 'max_retry_times' in request.meta:
+            retry_times = request.meta['max_retry_times']
 
         stats = spider.crawler.stats
         if retries <= retry_times:
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -63,10 +63,7 @@ class RetryMiddleware(object):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
-        retry_times = self.max_retry_times
-
-        if 'max_retry_times' in request.meta:
-            retry_times = request.meta['max_retry_times']
+        retry_times = request.meta.get('max_retry_times') or self.max_retry_times
 
         stats = spider.crawler.stats
         if retries <= retry_times:
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -63,11 +63,13 @@ class RetryMiddleware(object):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
+        retry_times = self.max_retry_times
+
         if 'max_retry_times' in request.meta:
-            self.max_retry_times = request.meta['max_retry_times']
+            retry_times = request.meta['max_retry_times']
 
         stats = spider.crawler.stats
-        if retries <= self.max_retry_times:
+        if retries <= retry_times:
             logger.debug("Retrying %(request)s (failed %(retries)d times): %(reason)s",
                          {'request': request, 'retries': retries, 'reason': reason},
                          extra={'spider': spider})
--- a/tests/test_downloadermiddleware_retry.py
+++ b/tests/test_downloadermiddleware_retry.py
@@ -103,6 +103,34 @@ class RetryTest(unittest.TestCase):
         req = self.mw.process_exception(req, exception, self.spider)
         self.assertEqual(req, None)
 
+    def test_different_retry(self):
+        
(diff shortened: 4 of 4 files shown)


scrapy/downloadermiddlewares/retry.py
```diff
@@ -63,6 +63,9 @@ def process_exception(self, request, exception, spider):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
+        if 'max_retry_times' in request.meta:
+            self.max_retry_times = request.meta['max_retry_times']
```
redapple: This would override the middleware `max_retry_times` attribute for all subsequent retried requests, while `request.meta['max_retry_times']` should only affect this particular request.

harshasrinivas:
@redapple  If this sounds good, I'll start working on the unit test and docs too.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -63,8 +63,13 @@ def process_exception(self, request, exception, spider):
     def _retry(self, request, reason, spider):
         retries = request.meta.get('retry_times', 0) + 1
 
+        retry_times = self.max_retry_times
+
+        if 'max_retry_times' in request.meta:
```
redapple: I think you could even have
```
retry_times = request.meta.get('max_retry_times') or self.max_retry_times
```

redapple:
The patch can be simplified a little (see my comment).
Yes, please submit tests and doc update. Thank you.

harshasrinivas:
@redapple I've added the test functions. Let me know whether it could be simplified as well. Thanks!

tests/test_downloadermiddleware_retry.py
```diff
@@ -103,6 +103,34 @@ def _test_retry_exception(self, req, exception):
...
+        req = self.mw.process_exception(req, exception, self.spider)
+        assert isinstance(req, Request)
+
+        retry_times = req.meta.get('max_retry_times') or self.mw.max_retry_times
+
+        while req.meta['retry_times'] != retry_times:
+            req = self.mw.process_exception(req, exception, self.spider)
+            assert isinstance(req, Request)
+        
+        self.assertEqual(req.meta['retry_times'], retry_times)
```
redapple: Can I suggest that `_test_retry` accepts an argument to actually test an expected retry times number?
so that `retry_times` variable in the test does not mimic what the implementation in the middleware does,
i.e. you're testing a behavior:

> if I set a max retry number in request meta, is the request retried X many times?,

instead of
> I know how retry time is computed, so I'm testing that the same code gives the same result

Also, I would find interesting to test with different `RETRY_TIMES` settings: `RETRY_TIMES=0`, `RETRY_TIMES` bigger, than `meta['max_retry_times']`, lower than etc.

harshasrinivas: Just realized that I was just mimicking the middleware's working. :)
I've made the update now.

tests/test_downloadermiddleware_retry.py
```diff
@@ -103,6 +103,41 @@ def _test_retry_exception(self, req, exception):
...
 
+    def test_different_retry(self):
+        
+        req = Request('http://www.scrapytest.org/invalid_url', meta={'max_retry_times': 3})
+
+        # SETINGS: meta(max_retry_times) = 3, RETRY_TIMES = 2
+        self._test_retry(req, DNSLookupError('foo'), 3)
+
+        req2 = Request('http://www.scrapytest.org/invalid_url')
+
```
redapple: I find the tests with `req2` difficult to follow with inline comments.
They seem to test the case for changing `RETRY_TIMES` setting only.
So the new meta key and the setting combinations are not tested much apart from `meta(max_retry_times) = 3` and  `RETRY_TIMES = 2`

I suggest that you create a different testcase class than `RetryTest(unittest.TestCase)`,
with each individual test method using a different couple of `RETRY_TIMES` and `meta['max_retry_times']`, initializing a new middleware instance, and checking that `meta['max_retry_times']` takes precedence in all cases where it is set (this is what we want I think). Testing the nominal "without `meta['max_retry_times']`" case is also important of course.
What do you think?

Using seperate test methods within a TestCase class helps when identifying regressions: you can spot 1 test that fails more easily (when running tox with `-v`) than checking which line of code the test broke in a longer test method.
This is not consistent in scrapy tests I admit (and I myself have written long test methods that do not follow what I write here).

docs/topics/downloader-middleware.rst
```diff
@@ -852,6 +852,11 @@ Default: ``2``
 
 Maximum number of times to retry, in addition to the first download.
 
+.. reqmeta:: max_retry_times
```
redapple: This looks redundant with what's in `docs/topics/request-response.rst`. I suggest that you only reference the meta key here, something like:
```
Maximum number of retries can also be specified per-request using :reqmeta:`max_retry_times` blablabla
```
(You can rephrase this of course.)
There are other uses of `:reqmeta:` in this downloader-middleware.rst

docs/topics/request-response.rst
```diff
@@ -342,6 +343,13 @@ download_fail_on_dataloss
 Whether or not to fail on broken responses. See:
 :setting:`DOWNLOAD_FAIL_ON_DATALOSS`.
 
+.. reqmeta:: max_retry_times
+
+max_retry_times
+---------------
+
+The meta key is used set retry times per request. When initialized, the :setting:`RETRY_TIMES` setting will be ignored by the downloader middleware.
```
redapple: I find the phrasing saying that the setting is being ignored a bit confusing.
I would suggest saying the meta key takes precedence over the setting.
Also, try using shorter lines (around 80 chars), they will be formatted nicely by Sphinx.

tests/test_downloadermiddleware_retry.py
harshasrinivas: Great! Creating a new testcase class seems like a very good option. Working on it!

docs/topics/request-response.rst
harshasrinivas: Sure, will update it soon.

harshasrinivas:
@redapple , Just realized that the statement `retry_times = request.meta.get('max_retry_times') or self.max_retry_times` would cause trouble when `meta('max_retry_times')` is set to zero. In that case, Python considers it to be [`False`](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) and sets `retry_times = self.max_retry_times` instead.

tests/test_downloadermiddleware_retry.py
```diff
@@ -104,5 +104,86 @@ def _test_retry_exception(self, req, exception):
...
+        self.crawler = get_crawler(Spider)
+        self.spider = self.crawler._create_spider('foo')
+        self.mw = RetryMiddleware.from_crawler(self.crawler)
+        self.mw.max_retry_times = 2
+
+    def test_without_metakey(self):
+        
+        req = Request('http://www.scrapytest.org/invalid_url')
+
+        # SETTINGS: RETRY_TIMES is NON-ZERO
```
redapple: I believe you can split this into 3 different tests.
You can set `self.mw.max_retry_times` to a new value at the start of each test method. (or use a local `mw` each time).

Also, changing the middleware `max_retry_times` attribute while working on the same `req` object makes it hard (for me at least) to follow the expected behavior.
I think it's easier to test: 1) I have this middleware with these settings, + 1 request object with some meta (or not), 2) it goes through the middleware handling an exception for that request 3) check how many times it is being retried 4) end test

tests/test_downloadermiddleware_retry.py
```diff
@@ -104,5 +104,86 @@ def _test_retry_exception(self, req, exception):
...
+        self.mw.max_retry_times = 5
+        self._test_retry(req, DNSLookupError('foo'), 5)
+
+        # SETTINGS: RETRY_TIMES = 0
+        self.mw.max_retry_times = 0
+        self._test_retry(req, DNSLookupError('foo'), 0)
+
+        # RESET RETRY_TIMES SETTINGS
+        self.mw.max_retry_times = 2
```
redapple: Resetting values can be done in a [`TestCase.tearDown()`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.tearDown)

tests/test_downloadermiddleware_retry.py
```diff
@@ -104,5 +104,86 @@ def _test_retry_exception(self, req, exception):
...
+
+    def test_with_metakey_preceding(self):
+        # request with meta(max_retry_times) is called first
```
redapple: I don't understand the comment _"request with meta(max_retry_times) is called first"_ what do  you mean by that?
Also, you can use docstrings to describe what the test is about.

tests/test_downloadermiddleware_retry.py
```diff
@@ -104,5 +104,86 @@ def _test_retry_exception(self, req, exception):
...
+        # SETINGS: RETRY_TIMES > meta(max_retry_times)
+        self.mw.max_retry_times = 5
+        self._test_retry(req3, DNSLookupError('foo'), 4)
+        self._test_retry(req2, DNSLookupError('foo'), 5)
+
+        # RESET RETRY_TIMES SETTINGS
+        self.mw.max_retry_times = 2
+
+    def test_with_metakey_succeeding(self):
+        # request with meta(max_retry_times) is called second
```
redapple: Same as above, I do not understand what is meant by _"request with meta(max_retry_times) is called second"_

tests/test_downloadermiddleware_retry.py
```diff
@@ -104,5 +104,86 @@ def _test_retry_exception(self, req, exception):
...
+
+    def test_with_metakey_zero(self):
+        
+        req = Request('http://www.scrapytest.org/invalid_url', meta={'max_retry_times': 0})
+        self._test_retry(req, DNSLookupError('foo'), 0)
+
+
+    def _test_retry(self, req, exception, max_retry_times):
+        
+        while max_retry_times > 0:
```
redapple: Could this loop forever in case of a regression where `req.meta['retry_times']` is not updated?
What do you think of looping on a range?

tests/test_downloadermiddleware_retry.py
harshasrinivas: @redapple , thanks! I just read through the [TestCase.tearDown()](https://docs.python.org/2/library/unittest.html#unittest.TestCase.tearDown) docs.

In that case, `tearDown()` would just reset ` self.mw.max_retry_times = 2` - which is also set in `setUp()` as well. So don't you think `tearDown()` is not neccesary for this particular situation?

tests/test_downloadermiddleware_retry.py
harshasrinivas: I meant the order in which the GET requests are made.

In this case, there are two requests made one after the other.
The first GET - `req1` (has meta key set)
The second GET - `req2` (has no meta key set)

I have implemented this function to check this: 
Even after making a GET request [with `meta(max_retry_times)`],
the subsequent request [without `meta(max_retry_times)`] must be retried for the correct number of times.

tests/test_downloadermiddleware_retry.py
harshasrinivas: The same as above, but the order in which GET requests are made, is different.

tests/test_downloadermiddleware_retry.py
redapple: So I guess that this reset at the end of each test is not required.

tests/test_downloadermiddleware_retry.py
redapple: Different requests will not influence each other within the middleware. Each request has its own meta dict. I don't think we need to be testing this. One test for each `(setting, meta[max_retry])` combination should suffice.

redapple:
Another thing to test is the bahevior mixed with `dont_retry` meta key. I believe it should take precedence over `max_retry_times`. And it's probably that way already with this PR. A check for this is needed in the test suite.

tests/test_downloadermiddleware_retry.py
harshasrinivas: @redapple , thanks! I've made the necessary modifications.

tests/test_downloadermiddleware_retry.py
harshasrinivas: Sure! I've removed the extra test function as well.

tests/test_downloadermiddleware_retry.py
harshasrinivas: I've changed the `_test_retry()` function to loop on a range now.

redapple:
hi @harshasrinivas , sorry for the late feedback.
Your latest changes look good to me, thanks again!

kmike:
Thanks @harshasrinivas for the fix and @redapple for the review!

PR #2644, opened 2017-03-12
[MRG+1] Fixed the FIXME in FSFilesStore

jorenham:
More specific exception catching: [os.path.getmtime](https://docs.python.org/2/library/os.path.html#os.path.getmtime) can only raise [os.error](https://docs.python.org/2/library/os.html#os.error).


redapple:
Looks good to me. Thanks @jorenham

PR #2646, opened 2017-03-13
[MRG+1] Allowed passing objects of Mapping class or its subclass to the CaselessDict initializer

woxcab:
CaselessDict can process sequence of general Mapping/MutableMapping class or subclass of Mapping.

Documentations both for Python 2 and Python 3 declare that all instances of `Mapping` class have `items()` method, so it's safely to generalize check from `isinstance(seq, dict)`. 

Real usage example: `requests` package returns responses which headers are instances of `requests.structures.CaseInsensitiveDict` that is subclass of `MutableMapping` and this patch allows to pass these headers directly to `scrapy.Response` initializer.

### DIFF
--- a/tests/test_utils_datatypes.py
+++ b/tests/test_utils_datatypes.py
@@ -1,5 +1,6 @@
 import copy
 import unittest
+from collections import Mapping, MutableMapping
 
 from scrapy.utils.datatypes import CaselessDict, SequenceExclude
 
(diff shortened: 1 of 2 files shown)


kmike:
The feature make sense. 

I wonder if we should go further and just check for 
`hasattr(seq, 'items') and callable(seq.items)`. Thoughts?

woxcab:
> I wonder if we should go further and just check for
>  hasattr(seq, 'items') and callable(seq.items). Thoughts?
Generic `items()` method can have another semantic and can return anything. But `Mutable.items()` behaviour is limited and specified by documentation and it's currently the most generic form of required behaviour, so I think the current version of pull request is more appropriate.

kmike:
ok, makes sense!

tests/test_utils_datatypes.py
```diff
@@ -18,6 +19,48 @@ def test_init(self):
...
+            def __iter__(self):
+                return iter(self._d)
+
+            def __len__(self):
+                return len(self._d)
+
+        seq = MyMutableMapping(red=1, black=3)
+        d = CaselessDict(seq)
+        self.assertEqual(d['red'], 1)
+        self.assertEqual(d['black'], 3)
```
kmike: Could you please split it into several separate methods? Original `test_init` implementation was not a good example, it should have been two methods in a first place.

woxcab: @kmike, into which semantic methods to split this method?

kmike: by input to the CaselessDict constructor: dict, sequence, Mapping, MutableMapping

woxcab: Completed.

kmike: Thanks @woxcab, looks good!

Issue #2654, opened 2017-03-15
Support for ptpython REPL


Granitosaurus:
https://github.com/jonathanslenders/ptpython

ptpython is the new fancy repl shell (like ipython, bpython) and it would be nice to have support for it.

redapple:
Closed via #2655

PR #2655, opened 2017-03-15
[MRG+1] add support for embeded ptpython shell

Granitosaurus:
Added support for ptpython embed shell for `scrapy shell` and such.

Related issue: https://github.com/scrapy/scrapy/issues/2654

### DIFF
--- a/scrapy/utils/console.py
+++ b/scrapy/utils/console.py
@@ -31,6 +31,15 @@ def _embed_bpython_shell(namespace={}, banner=''):
         bpython.embed(locals_=namespace, banner=banner)
     return wrapper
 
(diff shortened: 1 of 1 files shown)


Granitosaurus:
ptpython doesn't support direct banner argument for the embeded shell itself but I think print statement before initiating the embed should be good enough? I've tested with `scrapy shell` and `inspect_response` commands and didn't notice any stdout issues.

kmike:
The implementation works for me, +1 to merge.

dangra:
no pushbacks, just tried ptpython and seems quicker than bpython. I think it needs a similar test case than other consoles https://github.com/scrapy/scrapy/blob/master/tests/test_utils_console.py

PR #2657, opened 2017-03-16
Base class for downloader middleware

jorenham:
As part of #2633

### DIFF
--- a/scrapy/downloadermiddlewares/__init__.py
+++ b/scrapy/downloadermiddlewares/__init__.py
@@ -1,0 +1,15 @@
+class DownloaderMiddleware(object):
+    """Base class for downloader middleware.
+    """
+    def process_request(self, request, spider):
+        pass
+
+    def process_response(self, request, response, spider):
+        return response
+
+    def process_exception(self, request, exception, spider):
+        pass
+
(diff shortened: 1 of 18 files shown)


eLRuLL:
what about making it optional for middlewares to be included in `DOWNLOADER_MIDDLEWARES`, as we could tell scrapy to look for all `DownloaderMiddleware` based classes. Priority could be a class parameter and a default could be specified too (so every middleware would have same priority).

Same idea for pipeline, extensions, etc.

jorenham:
@eLRuLL I really like your idea. I'll get started on it.

scrapy/downloadermiddlewares/__init__.py
```diff
@@ -0,0 +1,15 @@
+class DownloaderMiddleware(object):
+    """Base class for downloader middleware.
+    """
+    def process_request(self, request, spider):
```
djunzu: `BaseDownloaderMiddleware`?

redapple: I wouldn't mind `DownloaderMiddleware`. `BaseSpider` for example was deprecated some time ago.

PR #2659, opened 2017-03-17
[MRG+1] Update item-pipeline.rst

clarksun:
insert is deprecated, use insert_one instead.


redapple:
That's indeed what https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.insert says.
Looks good to me.
Thank you @clarksun

PR #2663, opened 2017-03-18
Bug fix -- LinkExtractor.extract_links() raises AttributeError

karlind:
If response.body contains just a comment(such as `<-- hello world -->`), `LinkExtractor.extract_links()` will raise `AttributeError`

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -53,6 +53,9 @@ class LxmlParserLinkExtractor(object):
                 yield (el, attrib, attribs[attrib])
 
     def _extract_links(self, selector, response_url, response_encoding, base_url):
+        if selector.root is None:
+            return []
+
         links = []
         # hacky way to get the underlying lxml parsed document
         for el, attr, attr_val in self._iter_links(selector.root):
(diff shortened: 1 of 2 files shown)


kmike:
It looks like a bug which should be fixed either in https://github.com/scrapy/parsel or in lxml. I've seen you've sent pull requests to parsel already (https://github.com/scrapy/parsel/pull/77), but they are closed now; could you please open an issue in parsel bug tracker?

karlind:
@kmike One of `parsel`'s member said:
>I don't see a clear generic solution that would fit Parsel, I think right now it's best to handle these cases in the caller code.

So I launched a pull request for `scrapy` project

kmike:
Preprocessing of a body is not a good option indeed, but we can detect None result and return e.g. an empty document instead. I think such fix is appropriate for parsel library.

karlind:
@kmike I think so and that's exactly what I did before. But question was I couldn't pass through exist unit tests.

redapple:
@karlind , I've submitted https://github.com/scrapy/parsel/pull/82 continuing on your https://github.com/scrapy/parsel/pull/77.
I think the tests on that PR failed because there was no `else:` on the `None` result check.

PR #2666, opened 2017-03-19
Base class for extensions

jorenham:
As part of #2633

### DIFF
--- a/docs/topics/extensions.rst
+++ b/docs/topics/extensions.rst
@@ -104,10 +104,11 @@ Here is the code of such extension::
     import logging
     from scrapy import signals
     from scrapy.exceptions import NotConfigured
+    from scrapy.extensions import Extension
 
     logger = logging.getLogger(__name__)
 
-    class SpiderOpenCloseLogging(object):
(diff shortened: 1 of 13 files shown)


jorenham:
The methods in extensions are callbacks for certain signals. In each extension this is done manually. So I think it would be nice to have this automated in the `BaseExtension` class. I figured that there are two ways of implementing this, and I'm not sure which one to choose.

1. Modify [`Crawler.__init__`] to automatically hook the crawlers signals to the appropriate extension methods. For 'old-style' extensions there is the risk that something goes wrong with signals (like double callback execution) since the method names aren't always the same (e.g. [`throttle.py`](https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/throttle.py#L18)).
2. Create an extra method in the `Extension` base class, e.g. `def connect_signals(self, crawler)`, that hooks all the methods to the crawler signals. This is the most safe option and doesn't break 'old-style' extensions, but requires the most code change in the built-in extensions.

Would this be appreciated, and if so, which implementation should I use and I create a separate PR / issue for this?

PR #2668, opened 2017-03-20
[MRG+1] Add sphinx_rtd_theme to docs setup readme

harshasrinivas:
Concerns Sphinx 1.4 release #1893 . `sphinx_rtd_theme` has to be installed manually.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -15,7 +15,7 @@ and all its dependencies run
 
 ::
 
-    pip install 'Sphinx >= 1.3'
+    pip install 'Sphinx >= 1.3' sphinx_rtd_theme
 
 
 Compile the documentation
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -15,7 +15,7 @@ and all its dependencies run
 
 ::
 
-    pip install 'Sphinx >= 1.3' sphinx_rtd_theme
+    pip install -r requirements.txt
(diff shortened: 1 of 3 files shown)


redapple:
What about adding a `requirements.txt` to `scrapy/docs/` and referencing if in `tox.ini's [docs]`?

harshasrinivas:
@redapple Yes, I too thought of the same. :) 
But since there are just 2 requirements for building the docs, do you think adding a requirements.txt would be beneficial?

redapple:
I think it would still be easier to `pip install -r docs/requirements.txt`

harshasrinivas:
Sure @redapple , adding right away.

docs/README.rst
```diff
@@ -15,7 +15,7 @@ and all its dependencies run
 
 ::
 
-    pip install 'Sphinx >= 1.3' sphinx_rtd_theme
+    pip install -r requirements.txt
```
redapple: I'm not sure that it's clear that you would run this inside `scrapy/docs/`

harshasrinivas: Just like the `Compile the documentation` part of this README, to maintain uniformity shall I mention "run the following command from this dir"?

redapple: That would not hurt indeed.

PR #949, opened 2014-11-13
Use Reppy instead of robotsparser

mattfullerton:

Issue #754, opened 2014-06-21
RobotsTxtMiddleware doesn't support wildcards in Disallow rules


mattfullerton:
This is because, as I understand, Python's robotparser module doesn't support them either. There is an alternative, drop-in module, Robotexclusionrulesparser:

http://nikitathespider.com/python/rerp/

This line would need to be changed:
https://github.com/scrapy/scrapy/blob/master/scrapy/contrib/downloadermiddleware/robotstxt.py#L7

redapple:
I believe this is the one: https://github.com/scrapy/scrapy/pull/949
@mattfullerton , are you still working on it?

mattfullerton:
No :blush:

Parth-Vader:
Hi, I'm trying to make the PR for this issue. 
I have made the required changes, and after running the local test `tests/test_downloadermiddleware_robotstxt.py`, one test is failed.

The error is : 

```
twisted.internet.defer.FirstError: FirstError[#1, [Failure instance: Traceback: <class 'twisted.trial.unittest.FailTest'>: did not catch an error, instead got None
/home/parth/Codes/scrapy/.tox/py35/lib/python3.5/site-packages/twisted/internet/defer.py:653:_runCallbacks
/home/parth/Codes/scrapy/scrapy/downloadermiddlewares/robotstxt.py:77:cb
/home/parth/Codes/scrapy/.tox/py35/lib/python3.5/site-packages/twisted/internet/defer.py:459:callback
/home/parth/Codes/scrapy/.tox/py35/lib/python3.5/site-packages/twisted/internet/defer.py:567:_startRunCallbacks
--- <exception caught here> ---
/home/parth/Codes/scrapy/.tox/py35/lib/python3.5/site-packages/twisted/internet/defer.py:653:_runCallbacks
/home/parth/Codes/scrapy/.tox/py35/lib/python3.5/site-packages/twisted/trial/_asynctest.py:67:_cb
]]
```

redapple:
@Parth-Vader , which tests fails?
One can usually see the failing test when running `tox -- -v`:

```
$ tox -- -v tests/test_downloadermiddleware_robotstxt.py
GLOB sdist-make: /home/paul/src/scrapy/setup.py
py27 inst-nodeps: /home/paul/src/scrapy/.tox/dist/Scrapy-1.3.2.zip
[...]

PR #2669, opened 2017-03-20
Use Reppy instead of robotsparser

Parth-Vader:
Made the changes using https://github.com/scrapy/scrapy/pull/949 for https://github.com/scrapy/scrapy/issues/754.


scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -83,21 +88,11 @@ def _logerror(self, failure, request, spider):
...
-                # If we found garbage, disregard it:,
-                # but keep the lookup cached (in self._parsers)
-                # Running rp.parse() will set rp state from
-                # 'disallow all' to 'allow any'.
-                pass
-        # stdlib's robotparser expects native 'str' ;
-        # with unicode input, non-ASCII encoded bytes decoding fails in Python2
+        policy = HeaderWithDefaultPolicy(default=1800, minimum=600)
+
+        rp=Robots.fetch(response.url, ttl_policy=policy)
```
redapple: I understand that this performs an HTTP request to get the robots.txt file. Can you check?
We do not want to re-fetch the file, nor use a synchronous networking call (reppy uses python-requests as far as I understand)

requirements.txt
```diff
@@ -8,3 +8,4 @@ six>=1.5.2
 PyDispatcher>=2.0.5
 service_identity
 parsel>=1.1
+reppy>=0.3.0
```
redapple: Is 0.3.0 the first Python 3 compatible release?

redapple:
Thank you @Parth-Vader for stepping in!

requirements.txt
Parth-Vader: Yes, I ran the tests using `py35`.

redapple:
It looks like [either Cython or a C++ compiler](https://github.com/seomoz/reppy/blob/v0.4.6/setup.py#L39) is needed [to build reppy](https://github.com/seomoz/reppy/blob/v0.4.6/dev-requirements.txt) now.
Can you try updating Travis config to add Cython?
I'm not sure we want to add another non-python dependency to Scrapy by default. Do they have a non-C++/less-speedy/Python-only fallback?

kmike:
@redapple relevant discussion: https://github.com/seomoz/reppy/pull/33

Parth-Vader:
Version 0.3.3 seems to be Cython free : 
https://github.com/seomoz/reppy/blob/v0.3.3/dev-requirements.txt

kmike:
I think depending on old and unsupported version of a library is not OK, and depending on a library which is hard to install is also not OK.

Parth-Vader:
Agreed.
So, is Reppy the right option for the purpose?

redapple:
It looks like reppy is not (anymore) the right option for Scrapy's needs.

Gallaecio:
#3796 made it possible to configure which robots.txt parser to use, including a built-in adapter fopr Reppy.

PR #2670, opened 2017-03-21
[MRG+1] Update spiders.rst

qhuang872:
Added a note to allowed_domains attribute, reminding users not to add urls into the list.


kmike:
I like putting such examples to docs 👍

Issue #2375, opened 2016-11-02
ImportError: No module named 'twisted.mail.smtp'


duangy:
I use scrapy in python3.5, when I try to send a mail by MailSender, it's crashed.
```python
  File "/home/guangyao/Documents/Code/financer_crawl/env/lib/python3.5/site-packages/scrapy/mail.py", line 85, in send
    dfd = self._sendmail(rcpts, msg.as_string())
  File "/home/guangyao/Documents/Code/financer_crawl/env/lib/python3.5/site-packages/scrapy/mail.py", line 108, in _sendmail
    from twisted.mail.smtp import ESMTPSenderFactory
ImportError: No module named 'twisted.mail.smtp'
```

redapple:
@duangy , thanks for reporting.
Unfortunately, Scrapy does not support sending mail with Python 3.
This is actually a limitation of Twisted, which has not ported its `twisted.mail` module yet:
see https://twistedmatrix.com/trac/ticket/8770 and https://github.com/twisted/twisted/pull/509#issuecomment-242239177

MrLokans:
I confirm having the same issue on the latest scrapy on python 3.5. So, before twisted will fix that issue, what is the proper way of sending email reports from scrapy on python 3+ installations?

kmike:
@MrLokans I think you can use smtplib and email packages from standard library; see https://docs.python.org/3/library/email-examples.html. They are blocking though - it means that all processing will stop while email is waiting to be sent. So make sure sending is fast: run a local daemon (exim, postfix) and use it to send emails, so that when you call "send email" the message is put to a local queue, and then daemon handles the rest. Using e.g. gmail smtp interface directly is worse because it can be much slower.

MrLokans:
@kmike Thanks a lot for the reply.

TilakMaddy:
hmm .. learnt something today

rodrigc:
@redapple I've gone ahead and merged https://github.com/twisted/twisted/pull/509 .
It is not in any released version of Twisted yet.
If you are willing to try out Twisted directly from git, can you try it and and let me know how it goes?

Hades1996:
Someone knows if this issue is fixed? i'm using python 3.5, twisted 17.1.0, scrapy 1.3.0 and still can't use email service.

rodrigc:
I submitted https://github.com/scrapy/scrapy/pull/2671
which needs to be merged.

redapple:
[...]

PR #2671, opened 2017-03-21
[MRG+1] ESMTPSenderFactory compatibility for Python 3

rodrigc:
twisted.mail.smtp was ported to Python 3 in this patch: https://github.com/twisted/twisted/pull/509
It is not in Twisted 17.1.0, but should be in the next released version of Twisted.

Fixes: https://github.com/scrapy/scrapy/issues/2375

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -30,14 +30,20 @@ from scrapy.utils.python import to_bytes
 logger = logging.getLogger(__name__)
 
 
+def _to_bytes_or_none(text):
+    if text is None:
+        return None
+    return to_bytes(text)
+
+
 class MailSender(object):
 
     def __init__(self, smtphost='localhost', mailfrom='scrapy@localhost',
             smtpuser=None, smtppass=None, smtpport=25, smtptls=False, smtpssl=False, debug=False):
         self.smtphost = smtphost
         self.smtpport = smtpport
-        self.smtpuser = to_bytes(smtpuser) if smtpuser is not None else None
-        self.smtppass = to_bytes(smtppass) if smtppass is not None else None
+        self.smtpuser = _to_bytes_or_none(smtpuser)
+        self.smtppass = _to_bytes_or_none(smtppass)
         self.smtptls = smtptls
         self.smtpssl = smtpssl
         self.mailfrom = mailfrom
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -5,10 +5,13 @@ See documentation in docs/topics/email.rst
 """
 import logging
 
+try:
+    from cStringIO import StringIO as BytesIO
+except ImportError:
+    from io import BytesIO
 import six
 
 from email.utils import COMMASPACE, formatdate
-from io import BytesIO
 from six.moves.email_mime_multipart import MIMEMultipart
 from six.moves.email_mime_text import MIMEText
 from six.moves.email_mime_base import MIMEBase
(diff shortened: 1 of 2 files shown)


redapple:
Hey @rodrigc , thank you for this!
When is the next Twisted release planned for?

rodrigc:
I'm not sure @hawkowl is the release engineer and decides.

redapple:
@rodrigc , @hawkowl , there's a couple of issues with current Twisted trunk when using `ESMTPSenderFactory` with Python 3
See https://github.com/scrapy/scrapy/issues/2375#issuecomment-299226510

redapple:
@rodrigc , I pushed a few additional changes

rodrigc:
@redapple these changes look fine.

scrapy/mail.py
```diff
@@ -5,7 +5,10 @@
 """
 import logging
 
-from six.moves import cStringIO as StringIO
+try:
+    from cStringIO import StringIO as BytesIO
```
rodrigc: @redapple you should just remove cStringIO import and just import BytesIO everywhere, to get consistent behavior on Python 2.7 and 3.x

scrapy/mail.py
```diff
@@ -32,8 +36,8 @@ def __init__(self, smtphost='localhost', mailfrom='scrapy@localhost',
             smtpuser=None, smtppass=None, smtpport=25, smtptls=False, smtpssl=False, debug=False):
         self.smtphost = smtphost
         self.smtpport = smtpport
-        self.smtpuser = smtpuser
-        self.smtppass = smtppass
+        self.smtpuser = to_bytes(smtpuser) if smtpuser is not None else None
```
rodrigc: @redapple can you modify to_bytes() so that it accepts None, and returns None?
If you need to do "is not None else None" every place you use to_bytes, that looks ugly.

scrapy/mail.py
redapple: what do you mean by "everywhere"?
The other places were ported to Python 3 in a few iterations with a few maintainers. I wouldn't want to change this in elsewhere.
I'll let other maintainers comment and review this PR.

scrapy/mail.py
redapple: Other uses of to_bytes() do no require this as far as I understand.
True, it may look ugly here, but I can live with that.
Let's wait for other reviews.

scrapy/mail.py
rodrigc: io.BytesIO behaves the same on Python 2.7 and Python 3.x.
I see no reason to import cStringIO as BytesIO on Python 2.7.
By "everywhere" I mean Python 2.7 and Python 3.x.

kmike: @rodrigc cStringIO doesn't copy passed data, while io.BytesIO does a copy, even if access is read-only. This is fixed in Python 3.5, but prior to 3.5 (it seems this includes Python 2.x) io.BytesIO can be 100x slower or more than cStringIO if you just want to provide a file-like interface for existing data. Probably it is not that important for mail sender, but still, using cStringIO when available is a right thing to do.

scrapy/mail.py
redapple: Alright @rodrigc , I added a `_to_bytes_or_none` helper. Looks nicer indeed :)

scrapy/mail.py
rodrigc: @kmike OK

rodrigc:
@redapple Twisted 17.9.0 has been released so it would be good to merge this.

redapple:
Hey @rodrigc , indeed, I just tested with Twisted 17.9.0 and sending emails is working in Python 3 (for me at least)

kmike:
Looks good to me. @redapple feel free to merge :)

rodrigc:
@redapple @kmike can you merge https://github.com/scrapy/scrapy/pull/2942 also?  There were some ESMTP fixes in Twisted in 17.9.0

PR #2672, opened 2017-03-22
Update exporters.py

kooy:
In windows, the exported csv will produce extra blank lines. Add **lineterminator = '\ n'**, to solve this problem.


kmike:
I think this issue is solved in https://github.com/scrapy/scrapy/pull/3039.

Issue #2452, opened 2016-12-15
Image Background converting to green.


mkaya93:
Hi, 

Problem is I'm downloading images with crawler but they are transparency. So image is looking like this.

![a2062c7f64b9a136c16f1a3d8491b70902986fc4](https://cloud.githubusercontent.com/assets/6665723/21229545/97a44c8c-c2ea-11e6-9f96-5e60106bd340.jpg)

But it should look like this.
 ##
![wd-bvbz0120jch-12tb-my-cloud-ex2-ultra-gigabit-ethernet-kisisel-bulut-depolama](https://cloud.githubusercontent.com/assets/6665723/21229505/7bb3b8d2-c2ea-11e6-94fc-921adb2f21ee.png)

kmike:
I'm not sure why it becomes green (and not other color), but the reason is that currently scrapy converts images to jpeg, and jpeg doesn't support transparency. See also: https://github.com/scrapy/scrapy/issues/1705, https://github.com/scrapy/scrapy/issues/1037.

samj1912:
I would like to work on this and implement it as suggested in https://github.com/scrapy/scrapy/issues/1705#issuecomment-234697561

simongartz:
I had the exact same problem, for us it was a PNG in palette mode. You have to convert it to a RGBA, make the background white and then convert that image to RGB.
For a quick fix you can inherit ImagesPipeline and override the convert_image method with

```    
class CustomImagesPipeline(ImagesPipeline):
    def convert_image(self, image, size=None):
            if image.format == 'PNG' and image.mode == 'RGBA':
                background = Image.new('RGBA', image.size, (255, 255, 255))
                background.paste(image, image)
                image = background.convert('RGB')
            elif image.mode == 'P': #  Here follows the changes from the original Scrapy 1.3.2 code
                image = image.convert("RGBA")
                background = Image.new('RGBA', image.size, (255, 255, 255))
                background.paste(image, image)
                image = background.convert('RGB')
            elif image.mode != 'RGB':
                image = image.convert('RGB')
    
            if size:
                image = image.copy()
                image.thumbnail(size, Image.ANTIALIAS)
    
            buf = BytesIO()
            image.save(buf, 'JPEG')
            return image, buf

```

Tested with Scarpy 1.3.2

Gallaecio:
Fixed by #2675

PR #2675, opened 2017-03-22
[MRG+1] Fixes conversion of transparent PNG with palette images to jpg #2452

simongartz:
Images with transparency using palette mode (pillow mode = "P") fails when converting to jpg. The resulting image will have a green background instead of white as expected.
By converting mode P images to RGBA before pasting into the white background and then converting the resulting image to RGB will fix the problem.
See #2452


kmike:
Looks good, thanks @simongartz! There is some code duplication, but it doesn't bother me much.

Issue #2677, opened 2017-03-23
Duplicate Content-Length header for POST requests with empty body


redapple:
Originally reported [on StackOverflow](https://stackoverflow.com/questions/42966405/scrapy-post-header-has-two-content-length-fields).

HTTP requests with POST method and no request body are sent with 2 `Content-Length: 0` headers.

Reproducible with Scrapy 1.3.3 and Twisted 17.1:

```
$ scrapy version -v
Scrapy    : 1.3.3
lxml      : 3.7.3.0
libxml2   : 2.9.3
cssselect : 1.0.1
parsel    : 1.1.0
w3lib     : 1.17.0
Twisted   : 17.1.0
Python    : 2.7.12 (default, Nov 19 2016, 06:48:10) - [GCC 5.4.0 20160609]
pyOpenSSL : 16.2.0 (OpenSSL 1.0.2g  1 Mar 2016)
Platform  : Linux-4.8.0-41-generic-x86_64-with-Ubuntu-16.10-yakkety
$ scrapy shell
>>> fetch(scrapy.Request('http://httpbin.org/post', method='POST'))
2017-03-23 10:58:34 [scrapy.core.engine] DEBUG: Crawled (200) <POST http://httpbin.org/post> (referer: None)
```
(note that httpbin.org/post output does not show duplicate headers, but the Wireshark capture does)

Wireshark sniffing:

```
POST /post HTTP/1.1
Content-Length: 0
Content-Length: 0
Accept-Language: en
Accept-Encoding: gzip,deflate
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
User-Agent: Scrapy/1.3.3 (+http://scrapy.org)
Host: httpbin.org

HTTP/1.1 200 OK
Connection: keep-alive
Server: gunicorn/19.7.1
Date: Thu, 23 Mar 2017 09:58:34 GMT
Content-Type: application/json
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Content-Length: 458
Via: 1.1 vegur

{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8", 
    "Accept-Encoding": "gzip,deflate", 
    "Accept-Language": "en", 
    "Connection": "close", 
    "Content-Length": "0", 
    "Host": "httpbin.org", 
[...]

PR #2678, opened 2017-03-23
Set bodyproducer with empty content for POST

redapple:
Fixes https://github.com/scrapy/scrapy/issues/2677

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -279,8 +279,10 @@ class ScrapyAgent(object):
             headers.removeHeader(b'Proxy-Authorization')
         if request.body:
             bodyproducer = _RequestBodyProducer(request.body)
+        elif method == b'POST':
+            bodyproducer = _RequestBodyProducer(b'')
         else:
-            bodyproducer = _RequestBodyProducer(b'') if method == b'POST' else None
+            bodyproducer = None
         start_time = time()
         d = agent.request(
             method, to_bytes(url, encoding='ascii'), headers, bodyproducer)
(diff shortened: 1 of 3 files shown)


scrapy/core/downloader/handlers/http11.py
```diff
@@ -280,19 +280,7 @@ def download_request(self, request):
...
-            #
-            # RFC 7230#section-3.3.2:
-            # "a Content-Length header field is normally sent in a POST
-            # request even when the value is 0 (indicating an empty payload body)."
-            #
-            # Twisted Agent will not add "Content-Length: 0" by itself
-            if method == b'POST':
-                headers.addRawHeader(b'Content-Length', b'0')
-
+            bodyproducer = _RequestBodyProducer(b'') if method == b'POST' else None
```
dangra: I would go for:
```python
if request.body:
  bodyproducer = _RequestBodyProducer(request.body)
elif method == b'POST':
  bodyproducer = _RequestBodyProducer(b'')
else:
  bodyproducer = None
```

dangra: btw, what is the expected behavior in terms of Content-Length header for PUT and PATCH requests without body content?

redapple: Twisted's fix considers both PUT and POST: https://github.com/twisted/twisted/pull/670/

dangra:
looks good except for a missing comment explaining why it needs a special case for POST requests. A comment saying we can drop once minimum twisted version is bumped above 14.0.0

Also, a test case checking content-length is set only once as safe check for past and future compatibility with twisted.

redapple:
@dangra , PR updated.

PR #2685, opened 2017-03-25
runspider - warn in case of multiple spiders

### LINKED ISSUE #2608: scrapy runspider in case of several spiders in a file
[Docs](https://doc.scrapy.org/en/latest/topics/commands.html?highlight=runspider#std:command-runspider) say: "run a spider self-contained in a Python file, without having to create a project."

In case there are multiple spiders in a file, one of them will be selected and run, but spider selection is not deterministic.

Although it's technically not backwards compatible, maybe the best course of action would be to raise an error in this case and do not run any spiders?

harshasrinivas:
Fixes #2608


harshasrinivas:
Some of the builds seem to fail because sometimes the first spider is getting executed. But, in the [`runspider.py`](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/runspider.py#L86), the code seems to pick and run the last-defined spider only.

Will need some help with this.

AdrianAtZyte (approved):

PR #2690, opened 2017-03-29
FAQ Rewrite note on Python 3 support on Windows

redapple:


kmike:
Looks good, thanks Paul!

PR #2695, opened 2017-04-02
[MRG+1] command-list-fix

LMKight:
Remove "commands" from the command list (the one that gets printed when you type e.g. "scrapy" in the command line), since that is the baseclass of all commands and cannot be executed.


kmike:
A good catch! 

I'm not sure I like the fix though - it is non-obvious what this code does. The goal is to exclude base ScrapyCommand class from the list, right? It seems a more clear fix is to exclude it in `_iter_command_classes`: `issubclass(ScrapyCommand, ScrapyCommand)` is True, so an extra check is needed.

kmike:
Looks good to me, thanks @LMKight!

PR #2368, opened 2016-10-28
PEP8 the code generated by startproject and genspider 

stummjr:
The code generated by the `startproject` and `genspider` (csvfeed, xmlfeed and crawl templates) is not PEP8 compliant (mostly due to [error E265](http://pep8.readthedocs.io/en/latest/intro.html#error-codes)).

This PR makes the templates code PEP8 compliant.

PR #2700, opened 2017-04-05
PEP8 warning in project template

wddwycc:
If PEP8 linter installed, `settings.py` looks like this:


redapple:
Hi @wddwycc ,
I believe it's the same as https://github.com/scrapy/scrapy/pull/2368
The thing is that [we want to keep commented code](https://github.com/scrapy/scrapy/pull/2368/files#r85601018) without a space after the #

wddwycc:
@redapple Ok, I understand, close the PR then.

PR #2707, opened 2017-04-11
allow specifying max_active_size for scraper.slot in settings

### LINKED ISSUE #1410: allow specifying max_active_size for scraper.slot
It's seems to me that such behaviour is undocumented - scraper will be in back out state once there is over 5M requests, as in https://github.com/scrapy/scrapy/blob/342cb622f1ea93268477da557099010bbd72529a/scrapy/core/scraper.py#L29

where the default value is always used.
https://github.com/scrapy/scrapy/blob/342cb622f1ea93268477da557099010bbd72529a/scrapy/core/scraper.py#L79

I am reaching a deadlock scenario that I have over 5M requests in the slot, then since enqueue request will check for backout, new request can never be completed.

I can submit a PR if it make sense to extract to settings, say SCRAPER_SLOT_MAX_ACTIVE_SIZE

RajatGoyal:
fixes #1410

For some cases when I have many big (~20-30mb)  files to scrape, the spider just waits until everything of one file is scraped and enqueued and then takes on the new file, which makes my scraping almost synchronous. If I increase the `max_active_size` to a big value (x100) since we have machines with large ram, it works fine. 

This parameter needs to be configurable for the project/spider.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -26,7 +26,7 @@ class Slot(object):
 
     MIN_RESPONSE_SIZE = 1024
 
-    def __init__(self, max_active_size=5000000):
+    def __init__(self, max_active_size):
         self.max_active_size = max_active_size
         self.queue = deque()
         self.active = set()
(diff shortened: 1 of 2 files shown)


scrapy/core/scraper.py
```diff
@@ -26,7 +26,7 @@ class Slot(object):
 
     MIN_RESPONSE_SIZE = 1024
 
-    def __init__(self, max_active_size=5000000):
+    def __init__(self, max_active_size):
```
dangra: Removing the default will break existent code, we should keep it for backwards compatibility.

dangra:
superseded by #3551

PR #2720, opened 2017-04-27
Change "localhost" test server certificate

redapple:
I had this strange issue this afternoon that I could not understand (failing HTTPS tests even with master branch or latest tag code)
I was seeing tons of 
```
ResponseNeverReceived: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'ssl3_read_bytes', 'tlsv1 alert internal error')]>]
```
TLS connections were being closed by the test server in `tests/test_downloader_handlers.py`; The server was returning ServerHello but followed by an Alert, in the same payload:

```
Secure Sockets Layer
    TLSv1.2 Record Layer: Handshake Protocol: Server Hello
    TLSv1.2 Record Layer: Handshake Protocol: Certificate
    TLSv1.2 Record Layer: Alert (Level: Fatal, Description: Internal Error)
```

I eventually logged what was happening within Twisted and got this in :
```
_checkHandshakeStatus(): Error = Error([('rsa routines', 'RSA_sign', 'digest too big for rsa key'), ('SSL routines', 'ssl3_send_server_key_exchange', 'EVP lib')],)
```

So I tried regenerating the certificate and key for the "localhost" server and it worked.

I'm not sure what changed on my system to trigger this. But I think it doesn't hurt to use certificate and key that we can (re)generate (hence the accompanying README)


tests/test_downloader_handlers.py
```diff
@@ -183,8 +183,8 @@ class HttpTestCase(unittest.TestCase):
     download_handler_cls = HTTPDownloadHandler
 
     # only used for HTTPS tests
-    keyfile = 'keys/cert.pem'
-    certfile = 'keys/cert.pem'
```
kmike: are old files still needed?

redapple: Nope. I just removed it and the other reference to it.

Issue #667, opened 2014-03-25
Support dropping default input values in from_response() 


dangra:
It should be possible to ignore fields when submitting forms on `FormRequest.from_response()`

A proposal to consider is to drop any field whose value is `None` as passed in `formdata`.

source: https://github.com/scrapy/scrapy/pull/412#issuecomment-38553046

cathalgarvey:
@dangra - It looks like this might have been resolved? May I close this issue?

dangra:
you're right @cathalgarvey, thanks.

PR #2721, opened 2017-04-30
[MRG+1] Allow dropping field in from_response formdata

HarrisonGregg:
Implements #667.  `FormRequest.from_response` is very helpful, but sometimes it includes a field that you would like to remove.  With this change, you can include a field with value `None` in `formdata` when calling `from_response`, and the field will be omitted from the body of the returned request.

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -135,7 +135,7 @@ def _get_inputs(form, formdata, dont_click, clickdata, response):
         if clickable and clickable[0] not in formdata and not clickable[0] is None:
             values.append(clickable)
 
-    values.extend(formdata.items())
+    values.extend((k, v) for k, v in formdata.items() if v is not None)
     return values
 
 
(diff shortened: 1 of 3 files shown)


scrapy/http/request/form.py
```diff
@@ -135,7 +135,7 @@ def _get_inputs(form, formdata, dont_click, clickdata, response):
         if clickable and clickable[0] not in formdata and not clickable[0] is None:
             values.append(clickable)
 
-    values.extend(formdata.items())
+    values.extend([(k, v) for k, v in formdata.items() if v is not None])
```
dangra: no need to build an intermediate list. Just use:

    values.extend((k, v) for k, v in formdata.items() if v is not None)

dangra:
Nice to have feature, simple implementation and good test case. thanks @HarrisonGregg

scrapy/http/request/form.py
HarrisonGregg: I didn't realize you could do that. Thanks!  Would you prefer if I squashed this into the other commit on this file?

dangra:
Yes, please.

HarrisonGregg:
@dangra Squashed.

dangra:
@redapple I think this is ready to go. Is it still in time for 1.4.0?

redapple:
@dangra , alright, lets include it.
Thank you @HarrisonGregg

PR #2729, opened 2017-05-08
Fix doc: open file with "wb" mode will get an error in python 3

yandongxu:
python: 3.6.0rc2
scrapy: 1.3.0

output:
```
TypeError: a bytes-like object is required, not 'str'
```


kmike:
A good catch, thanks!

PR #2732, opened 2017-05-09
[WIP] add FILES_STORE_S3_STORAGE_CLASS

shaform:
It would be great if we could change storage class to save cost. Especially if we are using external machines such as the ones on scrapinghub.

p.s. I haven't tested this yet, thus WIP.


redapple:
Is this PR superseded by https://github.com/scrapy/scrapy/pull/2738 ?

shaform:
close for now as we've found other workarounds

PR #2738, opened 2017-05-12
Allow custom files storage class

shaform:
Recently I found that it is too restrictive to only have S3 storage when using scrapy on scrapinghub.
Hopefully we could use any custom storage classes.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -217,6 +217,11 @@ class FilesPipeline(MediaPipeline):
 
     MEDIA_NAME = "file"
     EXPIRES = 90
+    STORE_SCHEMES = {
+        '': FSFilesStore,
+        'file': FSFilesStore,
+        's3': S3FilesStore,
+    }
     DEFAULT_FILES_URLS_FIELD = 'file_urls'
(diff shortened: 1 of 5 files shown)


redapple:
Makes sense. What do you think about naming the settings with `FILES_STORE_SCHEMES` instead of `FILES_STORES`, to be consistent with the class attribute?
Did you have a specific new store in mind?

shaform:
Thanks for the review. Surely, I'll modify it.

I was actually trying to add a B2 store https://github.com/shaform/ptt-beauty/blob/master/ptt_beauty/pipelines.py#L27.
But as you could see in the above file, this would require some ways to read `B2_ACCOUNT_ID` and `B2_APPLICATION_KEY` from settings.py. I haven't actually come up with a general way to achieve this for any custom stores.

shaform:
One possibility might be adding a `settings` argument in the `__init__` function for all file stores including the s3store. In that case the following lines could also be moved into the `__init__` https://github.com/scrapy/scrapy/blob/d8672689761f0bb6c0550a841f35534265e87fee/scrapy/pipelines/images.py#L89. What do you think?

PR #2740, opened 2017-05-16
[MRG+1] Add support for executing scrapy using -m option of python

luzfcb:
This pull-request enable execute scrapy from python module.

```
python -m scrapy
```
On windows, this probably works, too. Although I did not test because I do not have Windows

```
py -3 -m scrapy
```

### DIFF
--- a/scrapy/__main__.py
+++ b/scrapy/__main__.py
@@ -1,0 +1,4 @@
+from scrapy.cmdline import execute
+
+if __name__ == '__main__':
+    execute()


kmike:
+1 to merge, I also wanted to add this feature.

luzfcb:
For reference, I've learned to do this based on this pull-request, done by one of Python's core commiters

https://github.com/audreyr/cookiecutter/pull/788

PR #2746, opened 2017-05-17
Add __str__ method on stats collector class

redapple:
I was playing with formatting messages with Python templates including crawler stats and thought it might be useful to have a standard formatting for `StatsCollector' instances.


scrapy/statscollectors.py
```diff
@@ -43,13 +43,16 @@ def open_spider(self, spider):
...
         if self._dump:
-            logger.info("Dumping Scrapy stats:\n" + pprint.pformat(self._stats),
+            logger.info("Dumping Scrapy stats:\n{}".format(self),
                         extra={'spider': spider})
         self._persist_stats(self._stats, spider)
 
     def _persist_stats(self, stats, spider):
         pass
 
+    def __format__(self, _):
```
kmike: I'm not sure ignoring format type is a right thing to do - for example, in the current implementation invalid format specifier for this data type (e.g. `"{:0.3f}".format(statscol)`) doesn't cause an exception - specifier just has no effect, pformat is always used.

redapple: hm right, but I wouldn't know what to do with a format specifier for a statscollector. Any suggestions?

kmike: I think the way to go is to check the format, use pformat if a string is requested, use repr if repr is requested, raise for numeric formats. But it looks like a lot of code without a clear purpose; maybe using `__str__` hook instead of `__format__` is easier.

redapple: @kmike , I've changed to a `__str__`

redapple:
This is what you can see in the telnet console:

```
>>> est()
Execution engine status

time()-engine.start_time                        : 5.67940211296
engine.has_capacity()                           : False
len(engine.downloader.active)                   : 0
engine.scraper.is_idle()                        : False
engine.spider.name                              : timewaste
engine.spider_is_idle(engine.spider)            : False
engine.slot.closing                             : False
len(engine.slot.inprogress)                     : 1
len(engine.slot.scheduler.dqs or [])            : 0
len(engine.slot.scheduler.mqs)                  : 0
len(engine.scraper.slot.queue)                  : 0
len(engine.scraper.slot.active)                 : 1
engine.scraper.slot.active_size                 : 1270
engine.scraper.slot.itemproc_size               : 0
engine.scraper.slot.needs_backout()             : False


>>> stats
<scrapy.statscollectors.MemoryStatsCollector object at 0x7fe949b0cfd0>
>>> p(stats)
<scrapy.statscollectors.MemoryStatsCollector object at 0x7fe949b0cfd0>
>>> print(stats)
{'downloader/request_bytes': 301,
 'downloader/request_count': 1,
 'downloader/request_method_count/GET': 1,
 'downloader/response_bytes': 913,
 'downloader/response_count': 1,
 'downloader/response_status_count/200': 1,
 'log_count/DEBUG': 3,
 'log_count/INFO': 6,
 'memusage/max': 47063040,
 'memusage/startup': 47063040,
 'response_received_count': 1,
 'scheduler/dequeued': 1,
 'scheduler/dequeued/memory': 1,
 'scheduler/enqueued': 1,
 'scheduler/enqueued/memory': 1,
 'start_time': datetime.datetime(2017, 6, 7, 15, 8, 18, 856955)}
>>> 
```

kmike:
While I like the way stats are displayed in `print`, this is also a bit misleading, as stats object doesn't support getting or setting value by `stats[key]`. What do you think about changing it to something which doesn't look like a dict (e.g. adding MemoryStatsCollector name in front of dict)? Or maybe we should finally support getitem/setitem for stats?

redapple:
> Or maybe we should finally support getitem/setitem for stats?

+1 to that!

AdrianAtZyte (approved):

wRAR (approved):

PR #2765, opened 2017-05-30
DOC not supporting python 3 yet

OlivierKnell:
Specified that it doesn't support python 3 yet


kmike:
Scrapy doesn't support Python 2 now, so closing :)

Issue #2766, opened 2017-05-30
Cryptic traceback for non-callable callback


redapple:
Originally from https://stackoverflow.com/questions/44259172/scrapy-twisted-internet-defer-defgen-return-exception

When a `scrapy.Request` is created with a callback that is a string (and not a callable),
> callback (callable) – the function that will be called with the response of this request (once its downloaded) as its first parameter. 

Twisted chokes with a confusing `twisted.internet.defer._DefGen_Return` exception traceback.

The error of using a string for a callback comes from allowing a string in [`CrawlSpider` rules](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.Rule).

> callback is a callable or a string (in which case a method from the spider object with that name will be used) to be called for each link extracted with the specified link_extractor.

### Suggestion

* either allow callback to be string and matched with a spider method, like in `CrawlSpider` rules
* or fail earlier in `scrapy.Request.__init__()` if a non-`None` callback is not callable

### How to reproduce

```
$ scrapy version -v
Scrapy    : 1.4.0
lxml      : 3.7.3.0
libxml2   : 2.9.3
cssselect : 1.0.1
parsel    : 1.2.0
w3lib     : 1.17.0
Twisted   : 17.1.0
Python    : 3.6.0+ (default, Feb 24 2017, 17:40:01) - [GCC 6.2.0 20161005]
pyOpenSSL : 17.0.0 (OpenSSL 1.0.2g  1 Mar 2016)
Platform  : Linux-4.8.0-53-generic-x86_64-with-debian-stretch-sid


$ cat noncallable/spiders/example.py 
# -*- coding: utf-8 -*-
import scrapy


class ExampleSpider(scrapy.Spider):
    name = 'example'
    start_urls = ['http://example.com/']

    def parse(self, response):
        yield scrapy.Request('http://httpbin.org/get?q=1', callback='parse_item')

    def parse_item(self, response):
        pass


$ scrapy crawl example
2017-05-30 16:04:47 [scrapy.utils.log] INFO: Scrapy 1.4.0 started (bot: noncallable)
(...)
2017-05-30 16:04:48 [scrapy.core.engine] INFO: Spider opened
2017-05-30 16:04:48 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
[...]

PR #2769, opened 2017-05-31
[MRG+1] Add verification to check if Request callback is callable

stummjr:
This PR intends to fix https://github.com/scrapy/scrapy/issues/2766.

When users pass a string to `Request.__init__`'s `callback` parameter, they end up getting a Twisted exception, which doesn't help to uncover this, which is a rather simple issue.

This PR changes that, so that users get a proper message when passing something that's not callable to the `callback` and `errback` parameters.

### DIFF
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -142,7 +142,8 @@ class Command(ScrapyCommand):
                 logger.error('Unable to find spider for: %(url)s',
                              {'url': url})
 
-        request = Request(url, opts.callback)
+        # Request requires callback argument as callable or None, not string
+        request = Request(url, None)
         _start_requests = lambda s: [self.prepare_request(s, request, opts)]
         self.spidercls.start_requests = _start_requests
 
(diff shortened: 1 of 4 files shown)


stummjr:
It seems that it broke other tests. I'll have a look on it. (I ran only the `test_http_request` tests locally)

stummjr:
@redapple I just updated the PR. I had to update the `parse` command implementation and also the request serialization tests.

redapple:
Looks good to me @stummjr

redapple:
@kmike @eliasdorneles @dangra @lopuhin , what do you think?

lopuhin:
The changes look good to me, 👍 
My only comment is maybe instead of saying ``callback must be a function, got ...``, we could say ``callback must be callable, got ...``, but I'm not sure here - the latter is more precise, but "function" is probably more familiar.

dangra:
I second Konstantin but would like to vote for _callable_ as it is technically more correct.

stummjr:
@redapple , @dangra, @lopuhin: I just fixed the error message. Thanks for your comments. 😃

Issue #2759, opened 2017-05-26
Document CloseSpider extension better


redapple:
See https://github.com/scrapy/scrapy/issues/2748#issuecomment-302536094

@kmike 's details:

> When CLOSERSPIDER_ITEMCOUNT limit is reached, Scrapy stops the spider gracefully; it means requests which are currently in the downloader queue (up to CONCURRENT_REQUEST requests) are still processed. This is the expected behavior.

redapple:
Closed via #2777

OfirD1:
> See [#2748 (comment)](https://github.com/scrapy/scrapy/issues/2748#issuecomment-302536094)
> 
> @kmike 's details:
> 
> > When CLOSERSPIDER_ITEMCOUNT limit is reached, Scrapy stops the spider gracefully; it means requests which are currently in the downloader queue (up to CONCURRENT_REQUEST requests) are still processed. This is the expected behavior.

I think it's worth adding the same documentation for `CLOSESPIDER_PAGECOUNT` and `CLOSESPIDER_ERRORCOUNT`. If it were there, I wouldn't have to test it myself to verify :)

eLRuLL:
@OfirD1 yeah I think we should put this part:

>    Requests which  are currently in the downloader queue (up to :setting: `CONCURRENT_REQUESTS` requests) are still processed. 

in a more generic part, as all closing behaviors still continue processing pending requests.

Gallaecio:
@OfirD1 Could you please open a new ticket about it?

OfirD1:
@Gallaecio, you mean open a new issue? or just edit it [here](https://github.com/scrapy/scrapy/edit/2.3/docs/topics/extensions.rst) and hit "propose changes"? I'm not sure what "open a new ticket" is, but I want to learn!

Gallaecio:
Sorry, I meant a new issue.

OfirD1:
Done.

#4836

PR #2777, opened 2017-06-05
Update extensions.rst

chuanjin:
#2759


kmike:
Thanks @chuanjin and @redapple!

PR #2781, opened 2017-06-09
use suggest method instead of DEPRECATED one

crasker:
see http://api.mongodb.com/python/current/api/pymongo/collection.html?highlight=insert#pymongo.collection.Collection.insert


kmike:
Thanks @crasker!

PR #2784, opened 2017-06-12
CaselessDict, Headers: Preserve capitalization

### LINKED ISSUE #2711: Scrapy capitalizes headers for request
I'm setting the headers following way
```python
headers = {
    'accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
    'cache-control': 'no-cache',
    ...
}
```
And calling request like that:
```
yield scrapy.Request(url='https:/myurl.com/', callback=self.parse, headers=headers, cookies=cookies, meta={'proxy': 'http://localhost:8888'})
```
And it makes that scrapy capitalizes all these headers and it looks like that (I'm using Charles proxy for debugging):
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Cache-Control: no-cache
```
And this is not working correctly for my case.

If I'm using curl and set headers lowercase
```
accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
cache-control: no-cache
```
everything works like a charm.

Is there any way how I can disable this capitalizing behavior in Scrapy?
Thanks for any help!

elacuesta:
Fixes #2711

I see two ways of implementing this feature:
- Store a `lower key: actual key` mapping within the `CaselessDict` class, and provide options to store the actual keys as they are provided (capitalize by default for backwards compatibility). This increases space complexity, and it's how this PR does it.
- Normalize and compare any provided key with the stored keys on all operations that require it (lookup or deletion, for instance). This would increase time complexity.

An example showing the changes of this PR:
```
>>> from scrapy.http import Request, Headers
>>> # capitalize keys (current behaviour)
>>> req1 = Request('http://example.org', headers={'will-be-capitalized': 'value'})
>>> req1.headers
{'Will-Be-Capitalized': ['value']}
>>> # preserve keys
>>> h = Headers({'will-be-preserved': 'value'}, capitalize=False)
>>> req2 = Request('http://example.org', headers=h)
>>> req2.headers
{'will-be-preserved': ['value']}
```

Looking forward to reading your comments!


mkky:
The actual headers that will be send controlled by Twisted. And not-capitalized keys will be capitalized. For reference please see
http://twistedmatrix.com/documents/8.2.0/api/twisted.web.http_headers.Headers.html#getAllRawHeaders

WigzyDev:
Is this still a pending issue that needs to be worked on? I'd be willing to draft up a PR if it is.

PR #2788, opened 2017-06-13
Check if attribute has a value before moving forward with the link extraction

rocioar:
I run into this bug using a link extractor to get all the images in a web page. The link extractor was returning the base_url for some reason. Investigated, and found out that the logic is not checking if the attribute is empty before url joining it with the base url.

As attr_val was empty this line:

`attr_val = urljoin(base_url, attr_val) `

was returning just the base_url.

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -60,6 +60,8 @@ class LxmlParserLinkExtractor(object):
             try:
                 if self.strip:
                     attr_val = strip_html5_whitespace(attr_val)
+                if not attr_val:
+                    continue
                 attr_val = urljoin(base_url, attr_val)
             except ValueError:
                 continue  # skipping bogus links


kmike:
As discussed with @rocioar: from puristic point of view it is correct to return links to the page itself, but in practice extracting a link to the same page is usually not helpful - you've already downloaded this page, so likely link is not needed.

I think it can be hard to make this a default now, as it is backwards incompatible and a bit opinionated. It can be an option ('drop_empty=True' or something like that), or maybe we can just provide an example in docs which uses LinkExtractor's `process` argument to drop empty links.

rocioar:
@kmike the problem with using the process method is that it's run after doing the urljoin, so the value won't be empty

kmike:
Yep, but if I'm reading the code correctly, it runs after urljoin with baseurl (which is usually empty), not with response url. 

I'm not sure why is URL joined with base url before running process function, it looks like a bad API; maybe this is what we shuld fix.

rocioar:
@kmike agree, we need to run the url joined after the process is run.. the problem is that this could break people's code, right?

kmike:
@rocioar right, the change would be backwards incompatible and we shouldn't take it lightly; it requires a discussion, and likely a separate ticket. I wonder why was it joining attributes with base url in a first place.

redapple:
@rocioar , @kmike ,
I believe this `urljoin()` with base URL dates back to the adding of `LxmlLinkExtractor`.
It looks like this was to do (roughly) the same as `lxml.html.HtmlElement.make_links_absolute()`
(see https://github.com/scrapy/scrapy/pull/559/files) (by me).
In hindsight, it surely does not look very natural to process the attribute _after_ urljoin-ing.
I cannot recall if this was because of failing tests or if this was missed.
We may want to check what `make_links_absolute()` does in practice where processing the attribute makes sense (extracting from JavaScript code for example, like [in the docs](https://docs.scrapy.org/en/latest/topics/link-extractors.html))

Issue #2785, opened 2017-06-13
[idea] Supporting also `link` tags in `Response.follow`


starrify:
Currently the `Response.follow` method supports `Selector` objects whose root tag are `<a>`. It'll be nice to support also `link` tags.

Related code: [here](https://github.com/scrapy/scrapy/blob/fad6b70d/scrapy/http/response/text.py#L168)

Sample code:

```python
# Now
for href in response.css('link[rel=next]::attr(href)').extract():
    yield response.follow(href)

# When the proposed feature is available
for link_sel in response.css('link[rel=next]'):
    yield response.follow(link_sel)
```

kmike:
I'm fine with such addition.
Do you have a real use case for it though?

dangra:
implemented in #2789

PR #2789, opened 2017-06-14
[MRG+1] Added: Now supporting <link> tags in Response.follow

starrify:
Implements #2785


kmike:
Looks good to me, thanks!

PR #2791, opened 2017-06-15
[MRG+1] Added doc for `scrapy.exceptions.DontCloseSpider`. Also fixes inaccurate doc for `scrapy.signals.spider_idle`.

### LINKED ISSUE #2709: A spider might still get closed even if it does have scheduled some requests in a `spider_idle` handle
Regarding the `spider_idle` signal, [the doc says](https://doc.scrapy.org/en/latest/topics/signals.html#spider-idle):

> If the idle state persists after all handlers of this signal have finished, the engine starts closing the spider. After the spider has finished closing, the spider_closed signal is sent.
> 
> You can, for example, schedule some requests in your `spider_idle` handler to prevent the spider from being closed.

However, the spider may still close even when it has scheduled some requests within a `spider_idle` handler, if all such requests are filtered due to duplication.  
In this case, the engine would still have `self.spider_is_idle(spider)` evaluated to `True` after finishing all `spider_idle` handlers.
[Related code in `scrapy/core/engine.py`](https://github.com/scrapy/scrapy/blob/05ce1296c6a60f23e81af7ec38ac1855e78be79f/scrapy/core/engine.py#L284)

Below is an example illustrating this issue. In this example, one may expect the spider to send three requests, but with the current version of Scrapy (`05ce1296`) there'll be only one request.

```python
# coding: utf8

import scrapy
import scrapy.signals


    name = 'test'
    urls = None

    def start_requests(self):
        self.urls = [
            'http://httpbin.org/get?foo=bar',
            'http://httpbin.org/get?bar=baz',
            'http://httpbin.org/get?baz=foo',
            'http://httpbin.org/get?baz=foo',
        ]
        self.crawler.signals.connect(self.on_idle, scrapy.signals.spider_idle)
        return []

    def on_idle(self, spider):
        if self.urls:
            url = self.urls.pop()
            self.crawler.engine.crawl(scrapy.Request(url), self)

    def parse(self, response):
        pass
```
(Save it to `a.py` and use `scrapy runspider a.py` to test.)

My suggestion is to have `ExecutionEngine._spider_idle` updated to make sure the spider does not get closed, as long as there's any request scheduled within a `spider_idle` handler.

starrify:
Resolves #2709

### DIFF
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -39,6 +39,14 @@ For example::
         if 'Bandwidth exceeded' in response.body:
             raise CloseSpider('bandwidth_exceeded')
 
+DontCloseSpider
+---------------
+
+.. exception:: DontCloseSpider
+
+This exception can be raised in a :signal:`spider_idle` signal handler to
+prevent the spider from being closed.
+
 IgnoreRequest
 -------------
 
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -189,13 +189,19 @@ spider_idle
     the engine starts closing the spider. After the spider has finished
     closing, the :signal:`spider_closed` signal is sent.
 
-    You can, for example, schedule some requests in your :signal:`spider_idle`
-    handler to prevent the spider from being closed.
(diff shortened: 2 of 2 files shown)


kmike:
Looks good to me, thanks!

PR #2799, opened 2017-06-21
Async HTTP cache using Deferreds

shirk3y:
Leverages Twisted `Deferred`s to make HTTP cache storage optionally asynchronous. Useful when using non-local services to store/retrieve responses.

No docs/tests yet, just piece of code proven to work in several projects.


redapple:
Looks useful to have built-in, indeed. Thanks @shirk3y !
As you mention already, tests will be required to have this merged.

PR #2807, opened 2017-06-30
python-attrs support to Scrappy Items

### LINKED ISSUE #2749: Feature Request: Item class field access and item creation using this
Items provide a simple way to structure data. However, as items are currently implemented, you can only populate an item by referencing a field using a string that matches the field's name e.g `item['field_name'] = value`. This becomes a problem when you want to change the name of a field after creation. It also requires that you remember the name of the fields defined for an item.

I think it would be nice to be able to populate items using syntax like the following:

```
item = Item()
item[Item.field] = value
```

In this way, you can use code completion in an IDE to be sure that you are referencing a field the Item class has defined and can leverage refactoring to change the names of fields across your entire project wherever Item.field is referenced.

Perhaps there is a good reason why this option was discarded during development but it isn't obvious to me why.

anaisabel7:
Fixes #2749 

Providing new ways to populate Scrapy items and adding the possibility of having @attr.s decorated Scrappy items.

This PR tries to solve the suggestions presented in issue #2749 

Certain arguments need to be passed to the decorator in order to make the decorated class (a subclass of Scrapy Item) compatible with the functioning of Scrapy Items. This is covered in the form of tests in the PR.

At the moment, the passing of default values to the attributes of the class (equivalent to Fields in normal Scrapy Items) is covered, but the passing of other metadata, such as validators, has not been considered.

This PR also covers the dot notation access to the Fields in Scrapy Items.

### DIFF
--- a/scrapy/item.py
+++ b/scrapy/item.py
@@ -70,13 +70,12 @@ class DictItem(MutableMapping, BaseItem):
 
     def __getattr__(self, name):
         if name in self.fields:
-            raise AttributeError("Use item[%r] to get field value" % name)
+            return self.__getitem__(name)
         raise AttributeError(name)
 
     def __setattr__(self, name, value):
         if not name.startswith('_'):
-            raise AttributeError("Use item[%r] = %r to set field value" %
-                (name, value))
+            self.__setitem__(name, value)
         super(DictItem, self).__setattr__(name, value)
 
     def __len__(self):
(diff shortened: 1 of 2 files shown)


kmike:
Hm, interesting. Your approach makes attribute access work. I was thinking about something a bit different - supporting any class decorated with `attr.s`, not Item subclasses decorated with `attr.s`.

gregbrowndev:
Would definitely be interested in support for using `attr`-based classes. Did you get any further along with this?

PR #2809, opened 2017-07-01
extract file extension properly

katz:
Remove query string part from the url first so that the logic can file extension properly.

The original code cannot extract the file extension properly if the media url contains query string like below:
https://example.net/some.jpg?ab=1e9b29b5d3d0aa13d601505c6a67ee3e&dc=59C43AAF


scrapy/pipelines/files.py
```diff
@@ -417,7 +417,14 @@ def _warn():
...
 
         media_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        media_ext = os.path.splitext(url)[1]  # change to request.url after deprecation
+        
+        media_ext = os.path.splitext(url)[1] # change to request.url after deprecation
+        if not media_ext[1:-1].isalnum():
+            # the file extension contains non alnum characters. e.g. "?"
+            # this usually happens if url is somethin like below
+            #   https://example.net/some.jpg?ab=1e9b29b5d3d0aa13d601505c6a67ee3e&dc=59C43AAF
+            media_ext = os.path.splitext(url.split('?',1)[0])[1]
```
dangra: why not removing the `media_ext` instead of resplitting `url` ?

Another approach that looks more natural is to use urlparse to remove the query string from url.

```python
>>> url = 'https://example.net/some.jpg?ab=1e9b29b5d3d0aa13d601505c6a67ee3e&dc=59C43AAF'
>>> os.path.splitext(urlparse(url).path)
('/some', '.jpg')
```

katz: > why not removing the query part from `media_ext `instead of resplitting `url `?

Good question.
If you simply remove the query part, the existing test code below will fail.
https://github.com/scrapy/scrapy/blob/cc06b6b1f68d33c1a5e5429d6dfb9b39577624fe/tests/test_pipeline_files.py#L49-L50
```
self.assertEqual(file_path(Request("http://www.dfsonline.co.uk/get_prod_image.php?img=status_0907_mdm.jpg")),
                         'full/4507be485f38b0da8a0be9eb2e1dfab8a19223f2.jpg')
```

scrapy/pipelines/files.py
```diff
@@ -417,7 +417,14 @@ def _warn():
         ## end of deprecation warning block
 
         media_guid = hashlib.sha1(to_bytes(url)).hexdigest()  # change to request.url after deprecation
-        media_ext = os.path.splitext(url)[1]  # change to request.url after deprecation
+        
+        media_ext = os.path.splitext(url)[1] # change to request.url after deprecation
+        if not media_ext[1:-1].isalnum():
```
redapple: What's the reason for testing on `[1:-1]` only?

maximium: Hey! Do you merge this if done right? Those filenames with query string is really annoying, need to add custom pipeline every time..

wRAR:
Closed by #7414

Issue #2727, opened 2017-05-06
Bring back docstring comments in CrawlSpider


sseyboth:
Can we please bring back the docstring comments in CrawlSpider that were removed in https://github.com/scrapy/scrapy/commit/e2290a5359ee80c75d8bcf8de8b6894e61fad83a? The code has zero comments now. Just spent two hours pulling my hair out trying to pass extra params down the crawler. Until I found the removed docstrings in this commit that where hugely helpful :) Thanks!

kmike:
Yep, +1 to fix it. See https://github.com/scrapy/scrapy/issues/713. So far nobody had found time to do it though, so pull requests are welcome.

kanav-raina:
hello  ,
Should i only have to add  this as a heading in crawling rules in scrapy/docs/topics/spiders.rst?

class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None)

and define the process request and process_links .

kanav-raina:
When I visit   scrapy/docs/topics/spiders.rst , there is no .. class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None)
but when i try to edit it ,it become visible  in line 377. Will you please check this?

kanav-raina:
please look at this @cathalgarvey

cathalgarvey:
Huh, you're right; I have no idea why that's happening! :)  I'm not very experienced with RST, it would be interesting to figure out why that's failing to display!

kanav-raina:
should i open an issue on this?

kmike:
@kanav-raina github doesn't process .rst docs fully, it doesn't support all Sphinx instructions, autodocs extensions, etc. This is not a bug. Please build Scrapy docs locally.

kanav-raina:
ok @kmike

kanav-raina:
@kmike  will you please brief  me a bit what I actually  have to do in this issue?

kmike:
@kanav-raina sure! So, currently CrawlSpider and related classes are documented in Scrapy docs. This in not convenient when checking CrawlSpider source code - methods should have docstrings. At the same time, duplicating the same documentation in two places (.rst files and method docstrings) is bad - it is very easy to modify one without changing the other. To solve it, Sphinx (software we use to build HTML docs from .rst files) provides [autodoc](http://www.sphinx-doc.org/en/stable/ext/autodoc.html) extension. 

[...]

PR #2814, opened 2017-07-04
Add documentation to CrawlSpider

yongzx:
#2727 
This brings back the docstring comments in CrawlSpider that were removed in [e2290a5](https://github.com/scrapy/scrapy/commit/e2290a5359ee80c75d8bcf8de8b6894e61fad83a).

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -32,20 +32,47 @@ class Rule(object):
 
 
 class CrawlSpider(Spider):
+    """
+    Class for spiders that crawl over web pages and extract/parse their links
+    given some crawling rules.
 
+    These crawling rules are established by setting the 'rules' class attribute,
+    which is a tuple of Rule objects.
+    When the spider is running, it iterates over these rules with each response
+    and do what it has to (extract links if follow=True, and return items/requests if
+    there's a parsing method defined in the rule).
+    """
     rules = ()
 
     def __init__(self, *a, **kw):
+        """Constructor takes care of compiling rules"""
         super(CrawlSpider, self).__init__(*a, **kw)
         self._compile_rules()
 
     def parse(self, response):
+        """
+        This function is called by the framework core for all the
+        start_urls. Do not override this function, override parse_start_url
+        instead.
+        """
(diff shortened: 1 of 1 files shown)


kmike:
Hey @yongzx,

Thanks for the pull request! 

It'd be great to avoid duplicating docstrings in https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst and CrawlSpider source code. I see you've used wording from https://github.com/scrapy/scrapy/commit/e2290a5359ee80c75d8bcf8de8b6894e61fad83a for docstrings, which is different from wording in the current docs. But still, it'd be better to use Sphinx autodocs extension, so that parts of documentation for CrawlSpider are pulled from CrawlSpider source code (parse_start_url method, main CrawlSpider docstring, maybe something else?)

yongzx:
Thank you for your reply! (I am still a beginner in contributing to the OS)

I will improve the documentation by removing the duplicated documentation and using the Sphinx autodocs extension to create the documentation for Crawl Spider.

yongzx:
Thank you for your reply! (I am still a beginner in contributing to the OS)

Let me see if I understand your reply correctly. I would need to replace the duplicated docstrings by using the wording from [https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst](https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst) and a better way to do this is to use Sphinx autodocs extension to pull the documentation for CrawlSpider from the spiders.rst. 
Do I understand correctly?

kmike:
@yongzx yeah, you're almost correct.

> I would need to replace the duplicated docstrings by using the wording from https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst 

Yes, this makes sense, as wording in spider.rst should be more recent.

> and a better way to do this is to use Sphinx autodocs extension to pull the documentation for CrawlSpider from the spiders.rst.

It is actually the other way around - spiders.rst should be populated from docstrings. So you move docstrings to CrawlSpider source code, and then pull them *to* spiders.rst from CrawlSpider docstirngs using Sphinx autodocs extension.

PR #2819, opened 2017-07-07
Documentation of CrawlSpider methods in Spiders.rst 

yongzx:
#2727 
Added the documentation from CrawlSpider source code to the [spiders.rst](https://github.com/scrapy/scrapy/blob/master/docs/topics/spiders.rst)

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -362,7 +362,7 @@ CrawlSpider
        described below. If multiple rules match the same link, the first one
        will be used, according to the order they're defined in this attribute.
 
-   This spider also exposes an overrideable method:
+   This spider also exposes a few overrideable methods:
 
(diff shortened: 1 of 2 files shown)


docs/topics/spiders.rst
```diff
@@ -371,6 +371,28 @@ CrawlSpider
...
+      the response which originated that results. It must return a list
+      of results (Items or Requests).
+
+   .. method:: _requests_to_follow(response)
+
+      This method iterates over each of the spider's rules, extracts the links
+      matching each case, filters them (if needed), and returns a list of unique
+      requests per response.
+
+   .. method:: _response_downloaded(response)
```
kmike: It seems there is still duplication. Do you think it is possible to make http://www.sphinx-doc.org/en/stable/ext/autodoc.html#directive-automethod work for this?

PR #2823, opened 2017-07-13
[WIP] An option to limit the cache size for the media pipeline

### LINKED ISSUE #939: Media pipeline problem
I think contrib.pipline.media.MediaPipeline should be add config to disable downloaded cache, when I use ImagePipeline to scrapy full site image, alway make memory leak. The media download should be make unique by user not scrapy framwork.

lopuhin:
Fixes #939  (the memory issue with media pipeline)

Only docs so far - if this looks good, I'll proceed with implementation. An option to limit the cache size to some fixed number complicated implementation a little bit, but seems it's worth it in case the crawler is downloading all images, so the same image can be referenced from multiple pages.

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -113,6 +113,18 @@ For the Images Pipeline, set the :setting:`IMAGES_STORE` setting::
 
    IMAGES_STORE = '/path/to/valid/dir'
 
+
+It is also recommended to review several options that are explained
+in more details below:
+
+- ``MEDIA_ALLOW_REDIRECTS`` to allow redirects when downloading media items:
+  by default redirects are not followed.
+  See :ref:`topics-media-pipeline-redirects`.
+- ``MEDIA_CACHE_SIZE`` to limit the cache size: by default meta information
+  (url, path and checksum) for all downloaded files and images is cached.
+  See :ref:`topics-media-pipeline-cache`.
+
+
 Supported Storage
 =================
 
@@ -320,6 +332,8 @@ all be dropped because at least one dimension is shorter than the constraint.
 
 By default, there are no size constraints, so all images are processed.
 
+.. _topics-media-pipeline-redirects:
+
 Allowing redirections
 ---------------------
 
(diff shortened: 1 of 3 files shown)


docs/topics/media-pipeline.rst
```diff
@@ -332,6 +346,23 @@ To handle media redirections, set this setting to ``True``::
...
+
+By default media pipeline caches meta information
+(url, path and checksum) for all downloaded files and images.
+This might consume noticeable amount of memory for large-scale crawls
+with millions of items downloaded. Set it to the number of most recently
+used items you wish to keep::
+
+    MEDIA_CACHE_SIZE = 10000
+
+Zero is also an allowed value, this will disable the cache completely.
```
kmike: We should mention that zero is default. It'd be also nice to mention it in https://doc.scrapy.org/en/latest/topics/leaks.html and https://doc.scrapy.org/en/latest/topics/broad-crawls.html.

docs/topics/media-pipeline.rst
```diff
@@ -113,6 +113,18 @@ For the Images Pipeline, set the :setting:`IMAGES_STORE` setting::
...
 
+
+It is also recommended to review several options that are explained
+in more details below:
+
+- ``MEDIA_ALLOW_REDIRECTS`` to allow redirects when downloading media items:
+  by default redirects are not followed.
+  See :ref:`topics-media-pipeline-redirects`.
+- ``MEDIA_CACHE_SIZE`` to limit the cache size: by default meta information
+  (url, path and checksum) for all downloaded files and images is cached.
```
kmike: we may note that this is done to avoid uploading the same file several times

docs/topics/media-pipeline.rst
lopuhin: Right, thanks for the pointers! Actually, in this case zero is not the default, default is an empty value that would be treated differently from zero. Zero is no cache, empty is unlimited cache. Does this sound confusing?

kmike: Ah yeah, a good point, it make sense that zero disables the cache. But then we should explain how to set an unlimited cache size.

docs/topics/media-pipeline.rst
```diff
@@ -332,6 +346,23 @@ To handle media redirections, set this setting to ``True``::
...
+Limiting cache size
+-------------------
+
+.. setting:: MEDIA_CACHE_SIZE
+
+By default media pipeline caches meta information
+(url, path and checksum) for all downloaded files and images.
+This might consume noticeable amount of memory for large-scale crawls
+with millions of items downloaded. Set it to the number of most recently
+used items you wish to keep::
```
kmike: Does cache size depend on a number of items, or on a number of files uploaded/downloaded?

lopuhin: Hm, this sentence is very confusing: what I wanted to say it that the cache will keep N recently used items - probably makes sense to be most recently accessed, even by cache hits. So the cache size is fixed by the user.

redapple:
As a side note, it would be awesome to have an object for the cache that would also work for https://github.com/scrapy/scrapy/issues/905 (see https://github.com/scrapy/scrapy/pull/2333 ), if possible.

lopuhin:
> As a side note, it would be awesome to have an object for the cache that would also work for #905 (see #2333 ), if possible.

I was also thinking that adding an extra dependency would be an overkill, thanks for the link!

isaacimholt:
Just wanted to chime in since this pr would address an issue we encountered as well. I don't see it explicitly mentioned anywhere but I would suggest to make this an LRU cache such that if necessary the spider can still benefit from commonly-found images but not hold on to unique or infrequent images.

PR #2826, opened 2017-07-14
[MRG+1] Tweak the CSVFeedSpider documentation

dguo:
I was confused by the `headers` description, so this is my attempt to make it clearer.


kmike:
It sounds better to me, thanks!

PR #2827, opened 2017-07-14
Added a feature to allow setting of custom log config file

umrashrf:
Added a feature to allow setting of custom log config file from project's settings file.

Need ideas for:
- Tests
- Whether or not it should be YAML or other supporting config file types


umrashrf:
Related https://github.com/scrapy/scrapy/issues/2231

redapple:
Initial thoughts: I would drop YAML support. I don't think we'd want to depend on PyYAML just for logging config reasons.

scrapy/utils/log.py
```diff
@@ -87,17 +88,29 @@ def configure_logging(settings=None, install_root_handler=True):
...
+        dictConfig(DEFAULT_LOGGING)
+
     if settings.getbool('LOG_STDOUT'):
         sys.stdout = StreamLogger(logging.getLogger('stdout'))
 
     if install_root_handler:
         install_scrapy_root_handler(settings)
 
+    if not LOG_CONFIG_FILE_LOADED:
+        logger.error("LOG_CONFIG_FILE from settings could not be loaded. Make sure it's a valid YAML file.")
```
umrashrf: Only log when LOG_CONFIG_FILE is provided.

umrashrf:
@redapple alright, another way would be to support YAML but log an error when YAML dep is not installed. Leaving the installation to the user.

```python
try:
    import yaml
except ImportError:
    logger.error("settings.LOG_CONFIG_FILE defined a yaml file but yaml package is not installed")
```

PR #2835, opened 2017-07-18
Add option to expand date and time in logs as UTC

### LINKED ISSUE #2802: Support ISO 8601 timestamps in logging
I haven't been able to configure logging via `LOG_DATEFORMAT` to output ISO 8601 timestamps.
Given that the logging formatter is passed local `time`s apparently (??), I suppose that I need to use the `%z` qualifier.
But `%z` always returns `+0000` in my configuration (Ubuntu 16.04, Linux 2.7) — whereas in my case `%Z` correctly produces `CEST`.
My understanding is that Python's `logging.Formatter()` doesn't support timezone offset: https://stackoverflow.com/questions/27858539/python-logging-module-emits-wrong-timezone-information##answer-27865750
Am I missing something obvious to be able to produce ISO 8601 timestamps in my Scrapy logs?

redapple:
Fixes #2802

This adds a `LOG_DATETIME_UTC` setting, defaulting to `False`, to expand datetimes as UTC if set to `True`.
This new setting would be most useful with `LOG_DATEFORMAT` with a `Z` or `+0000` suffix.

### DIFF
--- a/docs/topics/logging.rst
+++ b/docs/topics/logging.rst
@@ -149,6 +149,7 @@ These settings can be used to configure the logging:
 * :setting:`LOG_LEVEL`
 * :setting:`LOG_FORMAT`
 * :setting:`LOG_DATEFORMAT`
+* :setting:`LOG_DATETIME_UTC`
 * :setting:`LOG_STDOUT`
 * :setting:`LOG_SHORT_NAMES`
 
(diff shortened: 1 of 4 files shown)


AdrianAtZyte:
https://github.com/scrapy/scrapy/issues/2802#issuecomment-5761765283

PR #2841, opened 2017-07-19
[WIP] Support V4 signature with region name in S3 operations (botocore only)

### LINKED ISSUE #2448: S3FilesStore: add option for AWS Signature v4 support
Newer AWS locations will only authenticate clients that use AWS Signature Version 4.
For example, when using `Images` addon with a custom S3 location you can encounter with:
```
boto.exception.S3ResponseError: S3ResponseError: 400 Bad Request
The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.
```
The problem is pretty well known: https://github.com/boto/boto/issues/2741.

Boto library [supports](https://github.com/boto/boto/blob/2.39.0/boto/auth.py#L1025) v4 if you pass a host explicitly and set `S3_USE_SIGV4` envvar. It's a bit [different](https://github.com/boto/botocore/issues/424#issuecomment-68394401) for `botocore`, but should work fine when providing a region short name.

My proposal is to implement it in a similar way to this [approach](https://github.com/danilop/yas3fs/issues/101): `S3FilesStore` could handle new `AWS_HOST` and `AWS_S3_USE_SIGV4` settings, use it to set a corresponding envvar and pass the host to `boto.S3Connection`(or extract the region from host and pass it to botocore client).

redapple:
Fixes #2448

Still work in progress.

For the core S3 download handler, this patch makes use of Signature version 4 in all non-anonymous cases.
[SigV4 is supported by all regions](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region). And is also [the signature one should use](https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html#signature-versions).

> Signature Versions
> 
> AWS supports two signature versions: Signature Version 4 and Signature Version 2. You should use Signature Version 4. All AWS services support Signature Version 4, except Amazon SimpleDB which requires Signature Version 2. For AWS services that support both versions, we recommend that you use Signature Version 4.
> 
> All AWS regions support Signature Version 4.

**Note: this does NOT add SigV4 support for non-botocore users.**

Missing:

- [x] Support SigV4 in media pipelines
- [ ] Fix AWS signature tests


redapple:
@lopuhin , @vshlapakov , what do you think of only introducing a region name setting, and supporting V4 signature only for botocore?
I see that recent botocore even handles wrong regions by correcting the signature after AWS reports the error (this introduces additional roundtrip though.)

scrapy/pipelines/files.py
```diff
@@ -261,13 +270,19 @@ def from_settings(cls, settings):
...
             scheme = 'file'
         else:
             scheme = urlparse(uri).scheme
         store_cls = self.STORE_SCHEMES[scheme]
-        return store_cls(uri)
+        store = store_cls(uri)
+        # this is a bit ugly but since S3FilesStore does not accept
+        # settings when instantiated, and to keep it backward-ompatible,
+        # the store object is patched with the region name
+        if scheme == 's3':
```
vshlapakov: Other way would be to implement `from_settings` class method in `S3FilesStore`, in the same way like Scrapy [works](https://github.com/scrapy/scrapy/blob/c22a4e3bb84448f778613717c3749f226df7880f/scrapy/middleware.py#L35-L40) with middlewares, for me it looks a bit weird that `*Store` instances don't have direct access to Scrapy settings, and I'm afraid that amount of such hacks will grow over time. I wonder if it adding such method breaks backward compatibility too though..

kmike: @vshlapakov related: https://github.com/scrapy/scrapy/pull/2044 and https://github.com/scrapy/scrapy/pull/1605

vshlapakov:
> only introducing a region name setting, and supporting V4 signature only for botocore
Sounds good to me, Paul, I guess it's fine for now

scrapy/pipelines/files.py
vshlapakov: @kmike I felt like it's not my own idea, thanks for the links :smile:

lopuhin:
> only introducing a region name setting, and supporting V4 signature only for botocore

yes, sounds good to me too, we could even deprecate boto2 at some point

scrapy/pipelines/files.py
redapple: - #1605 touches feed exporters and storages,
- this PR touches feed storage but only for S3, + media pipelines (which are pretty scrary regarding settings-based config), + S3 downloader handler

A revived #1605 looks like a pre-requisite for proper/harmonized S3 setting passing to the various components.

wRAR:
Does anyone know if this is still needed, especially considering https://github.com/scrapy/scrapy/issues/2448#issuecomment-316965047?

Gallaecio:
I think so. I’m closing this and the source issue. They can always be reopened if I am wrong.

jackblk:
Coming from Google, I have solved this issue by using `AWS_DEFAULT_REGION` in my env var with latest botocore and default FEEDS. Hope it can help others.

PR #2845, opened 2017-07-24
from six.moves import xrange for Python 3

cclauss:


extras/qpsclient.py
```diff
@@ -41,7 +41,7 @@ def start_requests(self):
 
         slots = int(self.slots)
         if slots > 1:
-            urls = [url.replace('localhost', '127.0.0.%d' % (x + 1)) for x in xrange(slots)]
+            urls = [url.replace('localhost', '127.0.0.%d' % (x + 1)) for x in range(slots)]
```
voith: or may be `from six.moves import xrange` to get the generator version in both versions.

cclauss: I thought about that but slots should never be too large.

cclauss:
from six.moves import xrange

extras/qpsclient.py
redapple: I think it's fine here to use `range()` for both Python 2 and Python 3

cclauss:
Either this PR or #2849

dangra:
closed on favour of #2849

PR #2847, opened 2017-07-24
Handle HTTP 308 Permanent Redirect

### LINKED ISSUE #2844: Redirect 308 missing
I did a check on the RedirectMiddleware and noticed that code 308 is missing. Is there a reason for that?
Some websites don't update their sitemap and have a long list of 308 from http to https.

(side note: is there a way to add "s" before a link is scraped?)

redapple:
Fixes #2844

### DIFF
--- a/tests/test_downloadermiddleware_redirect.py
+++ b/tests/test_downloadermiddleware_redirect.py
@@ -42,6 +42,10 @@ class RedirectMiddlewareTest(unittest.TestCase):
         _test('POST')
         _test('HEAD')
 
+        _test('GET', status=307)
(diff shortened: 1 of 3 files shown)


tests/test_downloadermiddleware_redirect.py
```diff
@@ -42,6 +42,10 @@ def _test(method):
         _test('POST')
         _test('HEAD')
 
+        _test('GET', status=308)
+        _test('POST', status=308)
+        _test('HEAD', status=308)
```
dangra: tests are fine but looks like a good opportunity to add tests for 307 status too

Issue #2545, opened 2017-02-08
enable test_proxy_connect in Python 3


kmike:
mitmproxy is ported to Python 3 (in fact, it is now Python3-only), so it'd be nice to enable test_proxy_connect.py in Python 3.

ShivramIITG:
Hey, i am new to open source, this is the first issue i would work on (hopefully). Is this a beginner level issue? If so, i would like to start working on it. Kindly guide me.

kai3n:
Me too, I am new and want to contribute this open source.. Could you recommend some beginner level issue?

kmike:
Hey @ShivramIITG and @kai3n! Yes, I think this is a good isses for a beginner. It is not a super-easy issue, but it is clear what the end result should be.

I think to work on it one have to 

1. figure out how to run Scrapy tests (the whole testing suite and individual tests);
2. check that test_proxy_connect.py tests pass in Python 2;
3. enable them in Python 3 (remove a line from py3-ignores.txt) and run;
4. figure out why are they failing (likely because mitmproxy is not installed or is not available in Python 3);
5. add new mitmproxy (or some of its related libraries, I'm not sure - they refactored the code structure quite a lot) to tox.ini Python 3 environments, keeping old mitmproxy for Python 2 environments;
6. figure out what changed in mitmproxy and how to port tests to the new version, so that they work both in Python 2 (with an old version) and Python 3 (with the new version).

Maybe there is a way to use only a new mitmproxy version, I haven't checked that - it'd be ideal.

Parth-Vader:
@kmike I ran the whole test for both python2 and python3.
For `python2`, only `test_proxy_connect.py` ran (among the proxy ones) while no proxy tests for `python3`.

But I, currently, am using a proxy network.Does it affect the test?

Parth-Vader:
Doing `sudo tox -e py35 -- tests/test_proxy_connect.py`
gave the error : 
`ImportError: No module named 'libmproxy'`

And seeing this [Move to Python 3](https://github.com/mitmproxy/mitmproxy/issues/326), I believe it hasn't been moved to python 3.

Parth-Vader:
ping @kmike 
What has to be done next?

kmike:
Hey @Parth-Vader,

The ticket you linked to says that mitmproxy is moved to Python 3 (but Python 2 support is dropped). Sowe need a new version of mitmproxy in Python 3 environments and old version of mitmproxy in Python 2 environments,

[...]

PR #2864, opened 2017-07-29
Use getfullargspec under the scenes for py3 to stop DeprecationWarning

### LINKED ISSUE #2862: DeprecationWarning: inspect.getargspec()
Hi,

I've tried searching for a similar problem but couldn't find a discussion in scrapy about it, so starting one.

```bash
build  │   func_args, _, _, _ = inspect.getargspec(func)
build  │ /usr/local/lib/python3.6/site-packages/scrapy/utils/python.py:201: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```

We've stumbled into this log a LOT, actually getting spammed by it currently when we execute some `twisted.trial.unittest` tests where each unittest spins up a `CrawlerRunner` for our live tests that do exactly what unittests do but with live data fetching.

Could we leave the current behavior for py2 but port the `getargspec` to `getfullargspec` as is suggested in the warning? I see latter for py3 includes additional return values, but should be a quick port. Would you be ok with a patch? Thanks

andreip:
fixes #2862

I also ran `tox` for the tests, but there are slight differences for the packages used inside the library, so in `tests/requirements.txt`:
- `netlib` can be easily updated to `0.17` to support both py2 and py3
- `mitmproxy` on the other hand can't w/o some modifications; its version `0.18+` that supports both py3 and py2 doesn't contain `libmproxy` that's used in `tests/test_proxy_connect.py`. So yea, related to https://github.com/scrapy/scrapy/issues/2545 actually, so nevermind.

Other than that issue, everything worked fine, no errors for py2 nor py3.

Also I tested the branch in our project and the `DeprecationWarning` has disappeared now :).

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -197,8 +197,8 @@ def binary_is_text(data):
     return all(c not in _BINARYCHARS for c in data)
 
 
-def getargspec_py23(func):
-    """getargspec_py23(function) -> named tuple ArgSpec(args, varargs, keywords,
+def _getargspec_py23(func):
+    """_getargspec_py23(function) -> named tuple ArgSpec(args, varargs, keywords,
                                                         defaults)
 
     Identical to inspect.getargspec() in python2, but uses
@@ -208,19 +208,19 @@ def getargspec_py23(func):
     >>> def f(a, b=2, *ar, **kw):
     ...     pass
 
-    >>> getargspec_py23(f)
+    >>> _getargspec_py23(f)
     ArgSpec(args=['a', 'b'], varargs='ar', keywords='kw', defaults=(2,))
     """
-    if six.PY3:
-        full_argspec = list(inspect.getfullargspec(func))
-        return inspect.ArgSpec(*full_argspec[:4])
-    return inspect.getargspec(func)
+    if six.PY2:
+        return inspect.getargspec(func)
+
+    return inspect.ArgSpec(*inspect.getfullargspec(func)[:4])
 
 
 def get_func_args(func, stripself=False):
     """Return the argument name list of a callable"""
     if inspect.isfunction(func):
-        func_args, _, _, _ = getargspec_py23(func)
+        func_args, _, _, _ = _getargspec_py23(func)
     elif inspect.isclass(func):
         return get_func_args(func.__init__, True)
     elif inspect.ismethod(func):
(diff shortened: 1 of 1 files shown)


scrapy/utils/python.py
```diff
@@ -10,11 +10,19 @@
...
 from scrapy.utils.decorators import deprecated
 
 
+# In py3, getargspec is deprecated, so use getfullargspec with same
+# behavior to silence the warning.
+if six.PY3:
+    def getargspec(func):
+        full_argspec = list(inspect.getfullargspec(func))
+        return inspect.ArgSpec(*full_argspec[:4])
+    inspect.getargspec = getargspec
```
kmike: What do you think about creating an utility function and using it, without patching inspect module?

andreip: any name suggestion? `getargspec_py23`?

scrapy/utils/python.py
```diff
@@ -198,10 +197,30 @@ def binary_is_text(data):
...
+    inspect.getfullargspec() for python3 behind the scenes to avoid
+    DeprecationWarning.
+
+    >>> def f(a, b=2, *ar, **kw):
+    ...     pass
+
+    >>> getargspec_py23(f)
+    ArgSpec(args=['a', 'b'], varargs='ar', keywords='kw', defaults=(2,))
+    """
+    if six.PY3:
```
redapple: If I recall, @kmike prefers `if not six.PY2`

dangra: I would go the other way around:

```python
if six.PY2:
    return inspect.getargspec(func)

return inspect.ArgSpec(*inspect.getfullargspec(func)[:4])
```

dangra: And I would make this function private, so we reserve the right to drop it if not needed by other utils in this module.

andreip: ok, makes sense to do that as long as usage of `inspect.getargspec` stays within the `utils.python` module, as is now. If at some point `inspect.getargspec` gets used outside, it probably makes sense to do `from scrapy.utils.python import getargspec` instead of `from inspect import getargspec` to be able to use that for both py2 and py3 without triggering deprecation warning.

dangra:
thanks!

PR #2866, opened 2017-07-30
Small fix in description of startproject arguments

simik-ru:


redapple:
Thank you @simik-ru !

PR #2875, opened 2017-08-07
add BASE_DIR to settings

darshanime:
add a `BASE_DIR` setting to `settings.py`. This is useful when you want to read csvs (in spiders for eg)
```python
        filepath = os.path.join(
            settings.BASE_DIR,
            '/path/to/file.csv'
        )

        with open(filepath, 'r', errors='ignore') as csvfile:
            # code
```

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -180,6 +180,23 @@ such as the :ref:`S3 feed storage backend <topics-feed-storage-s3>`.
 
 .. setting:: BOT_NAME
 
+BASE_DIR
+--------
+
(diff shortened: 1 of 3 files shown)


darshanime:
@kmike, @redapple review please.

redapple:
hi @darshanime , I do not think that it needs to be a setting.
You can have it as a helper method, e.g. `scrapy.utils.project.get_base_dir()`.
By the way, Scrapy already has `scrapy.utils.project.project_data_dir`. Have you looked at it? If it doesn't fit your use-case, what is missing?

PR #2882, opened 2017-08-14
add "self" parameter for middlewares.py.tmpl

BruceDone:
fix the startproject command generate the wrong middleware for project


kmike:
Hey @BruceDone,

this issue is fixed in the current 1.4 release by https://github.com/scrapy/scrapy/pull/2595; any particular reason you want it in 1.3? Is upgrading an option?

BruceDone:
Hi @kmike , thanks for your quick response , since i use the pip to install this version of package  , it seems we should always supply the right package for user to install , since the current version has already fixed it , feel free to close this pr  :)

kmike:
Thanks for the pull request @BruceDone, even though it is not merged!

PR #2887, opened 2017-08-19
Add debug print of config file path

### LINKED ISSUE #2825: Add debug print of config file path (if found and used)
Motivation: https://stackoverflow.com/questions/45076871/scrapy-is-trying-to-find-a-module-that-doesnt-exist

We could (DEBUG) log which `scrapy.cfg` is being used to configure settings.

gtseres:
Fixes #2825

tests/test_utils_conf.py and tests/test_utils_log.py passed. Would need some help if I need to write any unit tests for this change, and how I can get the scrapy.log file in the testcase.

### DIFF
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import sys
 import numbers
@@ -10,6 +11,7 @@ from scrapy.settings import BaseSettings
 from scrapy.utils.deprecate import update_classpath
 from scrapy.utils.python import without_none_values
 
+logger = logging.getLogger(__name__)
 
 def build_component_list(compdict, custom=None, convert=update_classpath):
     """Compose a component list from a { class: order } dictionary."""
@@ -72,6 +74,7 @@ def closest_scrapy_cfg(path='.', prevpath=None):
     path = os.path.abspath(path)
     cfgfile = os.path.join(path, 'scrapy.cfg')
     if os.path.exists(cfgfile):
+        logger.debug('Scrapy config file used: {}'.format(cfgfile))
         return cfgfile
     return closest_scrapy_cfg(os.path.dirname(path), path)
 
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -74,7 +74,6 @@ def closest_scrapy_cfg(path='.', prevpath=None):
     path = os.path.abspath(path)
     cfgfile = os.path.join(path, 'scrapy.cfg')
     if os.path.exists(cfgfile):
-        logger.debug('Scrapy config file used: {}'.format(cfgfile))
         return cfgfile
     return closest_scrapy_cfg(os.path.dirname(path), path)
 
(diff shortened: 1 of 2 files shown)


scrapy/utils/conf.py
```diff
@@ -72,6 +74,7 @@ def closest_scrapy_cfg(path='.', prevpath=None):
     path = os.path.abspath(path)
     cfgfile = os.path.join(path, 'scrapy.cfg')
     if os.path.exists(cfgfile):
+        logger.debug('Scrapy config file used: {}'.format(cfgfile))
         return cfgfile
     return closest_scrapy_cfg(os.path.dirname(path), path)
```
kmike: it doesn't seem the log message is dislayed in all cases

gtseres: Given that the function is recursive, isn't it the base case that will eventually be hit?

redapple: @gtseres , there's also 
```
if path == prevpath:
         return ''
```

redapple:
@gtseres , I tested the patch locally and it looks like the `logger.debug()` call does not output the config file path, probably because logging has not been configured yet (I used `scrapy crawl` within a basic default `scrapy startproject`-generated project).
I wonder if we can put that debug info in `scrapy.utils.log.log_scrapy_info`, simply calling `closest_scrapy_cfg()` there.

scrapy/utils/log.py
```diff
@@ -151,6 +151,11 @@ def log_scrapy_info(settings):
     d = dict(overridden_settings(settings))
     logger.info("Overridden settings: %(settings)r", {'settings': d})
 
+    # Let's check for the scrapy config file and write the path to the log
+    closest = closest_scrapy_cfg()
```
redapple: Missing import I believe

gtseres: Thanks, that was the issue

gtseres:
Is there anything else pending from my end so that we can get this merged?

redapple:
@gtseres , I've tested another setup that may not be so common, but which isn't handled as I would expect.
Suppose you have a `[settings]` section in your `~/.scrapy.cfg` that defines the default settings module, and you run a `scrapy runspider` outside of a Scrapy project directory, what appears in the logs is
```
2017-08-28 15:43:51 [scrapy.utils.log] DEBUG: Scrapy config file used: ""
```
while in fact Scrapy will use what's in `~/.scrapy.cfg` (the relevant bit is in [`get_sources()`](https://github.com/scrapy/scrapy/blob/a8893190a10798b61029d141a52fbb8deeb86d3b/scrapy/utils/conf.py#L102).)

So it looks like my suggestion of using `closest_scrapy_cfg()` directly is not sufficient. Sorry about that.
However, I don't have a good proposal for how to do this for all cases.

The original issue was that if the settings module is populated by a config file, Scrapy could log the path to that file. (The user was using `get_project_settings()`.)
When a script is used to run Scrapy, `get_project_settings()` can be used, but it's not mandatory and settings can be set manually in the script.

cathalgarvey:
Just to revive this one; is there any reason not to add the logging directly to the conf.py file instead, so that the final result of fetching config file path is what gets logged?

PR #2892, opened 2017-08-21
Define TelnetConsole.port = None in __init__()

### LINKED ISSUE #2702: error at telnet shutdown in Python 3
Scrapy version:

```
Scrapy    : 1.3.3
lxml      : 3.7.3.0
libxml2   : 2.9.3
cssselect : 1.0.1
parsel    : 1.1.0
w3lib     : 1.17.0
Twisted   : 17.1.0
Python    : 3.5.2 (default, Nov 17 2016, 17:05:23) - [GCC 5.4.0 20160609]
pyOpenSSL : 0.15.1 (OpenSSL 1.0.2g  1 Mar 2016)
Platform  : Linux-4.4.0-59-generic-x86_64-with-Ubuntu-16.04-xenial
```

I've got this exception on shutdown:

```
2017-04-05 18:18:10 [scrapy.core.engine] INFO: Spider closed (finished)
2017-04-05 18:18:10 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method TelnetConsole.stop_listening of <scrapy.extensions.telnet.TelnetConsole object at 0x7fbb993bfeb8>>
  File "/home/ubuntu/.local/lib/python3.5/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pydispatch/robustapply.py", line 55, in robustApply
  File "/home/ubuntu/.local/lib/python3.5/site-packages/scrapy/extensions/telnet.py", line 63, in stop_listening
    self.port.stopListening()
AttributeError: 'TelnetConsole' object has no attribute 'port'
```

Not sure if it is a real problem, but it'd be nice to shut down without error messages. No idea how to reproduce it, I've only seen it once.

cclauss:
Fixes  #2702 by only calling self.port.stopListening() if self.start_listening() has already been called.

### DIFF
--- a/scrapy/extensions/telnet.py
+++ b/scrapy/extensions/telnet.py
@@ -43,6 +43,7 @@ class TelnetConsole(protocol.ServerFactory):
             raise NotConfigured
         self.crawler = crawler
         self.noisy = False
+        self.port = None
         self.portrange = [int(x) for x in crawler.settings.getlist('TELNETCONSOLE_PORT')]
         self.host = crawler.settings['TELNETCONSOLE_HOST']
         self.crawler.signals.connect(self.start_listening, signals.engine_started)
@@ -60,7 +61,9 @@ class TelnetConsole(protocol.ServerFactory):
                      extra={'crawler': self.crawler})
 
     def stop_listening(self):
-        self.port.stopListening()
+        if self.port:
+            self.port.stopListening()
+            self.port = None
 
     def protocol(self):
         telnet_vars = self._get_telnet_vars()


kmike:
Thanks @cclauss! I think this fixes the shutdown error, but I don't understand why the error happens in a first place - why can `start_listening` call be missed in Python 3, but not in Python 2? Won't we silence a symptom without fixing a root issue?

cclauss:
What this fix is saying is: “_Let’s be certain about whether the port has been opened or not.  If the port has been opened then make sure to close it when done._”  This is just the intended behavior and does not mask anything.  As far as timing differences between Python 2 and 3, a whole book could probably be written on that topic.

Issue #1343, opened 2015-07-07
custom_settings isn't reflected in the "Overridden settings:" log message


nautilus28:
Hi,

I would like to notice that custom_settings per spider does not work properly. Global settings updated nothing when I added custom_settings in spiders, and unittests for that are missing too.

Regards,
Canh

kmike:
Do you have an exampe? Which settings don't work?

barraponto:
This one works just fine:
https://github.com/barraponto/fosshackers/blob/master/fosshackers/spiders/djangopatcherscrawl.py

nautilus28:
in my project settings.py I placed DOWNLOAD_TIMEOUT = 60 and in ./spiders/simple_spider.py


class SimpleSpider(scrapy.Spider):
 
    name = 'simple'
    custom_settings = {
        'DOWNLOAD_TIMEOUT': 30,
        'CONCURRENT_REQUESTS_PER_DOMAIN': 4
    }

    def start_requests(self):
        yield Request('http://www.google.com/', callback='parse')


and Scrapy logs say INFO: Overridden settings: {'DOWNLOAD_TIMEOUT': 60} when I ran the simple spider. It looks like that DOWNLOAD_TIMEOUT was not updated.

curita:
Hi @nautilus28 ! That log message gets printed before settings are updated with `custom_settings` :/ You can verify it by adding this `print` to the parse method in your spider:

```python
def parse(self, response):
    ...
    print 'DOWNLOAD_TIMEOUT: {}'.format(self.settings.getint('DOWNLOAD_TIMEOUT'))
```

curita:
If that's not it, make sure you're using Scrapy 1.0.

barraponto:
Shouldn't scrapy print that message later then?

curita:
Yes, I agree. I'll update the issue title to reflect the actual problem.

nramirezuy:
@curita  I solved this by just using an [`Extension`](https://gist.github.com/nramirezuy/5aba864e0c36986f548f), maybe it helps :smile:

nautilus28:
I place CONCURRENT_REQUESTS_PER_DOMAIN = 2 in custom_settings, but this does not shown in the log message. Then how can I make sure that custom_settings do really work?

ngoanhtan:
@nautilus28 : just see in debug log. It should alway have 1-2 request in same time.Or you can override open_spider/close_spider method for count it.

[...]

PR #2894, opened 2017-08-23
Move logging of overriden settings to Crawler init

redapple:
Fixes https://github.com/scrapy/scrapy/issues/1343

Proposal for https://github.com/scrapy/scrapy/issues/1343.
The idea here is to keep `log_scrapy_info()` only with Scrapy-framework related stuff,
and move settings and custom settings logging closer to where the framework is instantiated to perform a crawl (Crawler, Engine etc)

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -11,7 +11,7 @@ from scrapy.core.engine import ExecutionEngine
 from scrapy.resolver import CachingThreadedResolver
 from scrapy.interfaces import ISpiderLoader
 from scrapy.extension import ExtensionManager
-from scrapy.settings import Settings
+from scrapy.settings import overridden_settings, Settings
 from scrapy.signalmanager import SignalManager
 from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.utils.ossignal import install_shutdown_handlers, signal_names
@@ -33,6 +33,9 @@ class Crawler(object):
         self.spidercls = spidercls
         self.settings = settings.copy()
         self.spidercls.update_settings(self.settings)
+
+        d = dict(overridden_settings(self.settings))
+        logger.info("Overridden settings: %(settings)r", {'settings': d})
 
         self.signals = SignalManager(self)
         self.stats = load_object(self.settings['STATS_CLASS'])(self)
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -9,7 +9,7 @@ from twisted.python.failure import Failure
 from twisted.python import log as twisted_log
 
 import scrapy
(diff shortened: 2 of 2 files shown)


kmike:
Looks good to me! +1 to merge.

This is very slightly backwards incompatible, because logger which logs a message is not the same, so custom user code which does anything with that message (e.g. disable it) won't do the same. I don't think this is a problem.

redapple:
Right @kmike . The logger changes. I also think this backward-incompatibility is acceptable given the benefot of properly logging overridden settings. We can be explicit about this in the release notes.

kmike:
Thanks @redapple! 

I realized that custom settings are not covered by `overridden_settings`, but it looks like a separate issue. We may want custom components (e.g. a third-party middleware) to say "this is my option, and this its default value", so that overrides can be shown for these options as well.

redapple:
@kmike , you mean setting keys that are not in default_settings? e.g. `MYPLUGIN_URL`

kmike:
@redapple yes.

redapple:
Right. A seperate issue for me too.

MalikRumi:
@redapple Would you please be so kind as to clarify for me exactly what the Overridden settings info in the log when a spider starts is telling me? If these are the settings that are being overridden, what setting are they being overridden by/with? And exactly which settings module are being overridden? I ask because I find this ambiguous. I have not set any settings on the command line, nor have I set custom ones on the spider, so to me that means I am using the default project settings. If that's true, then it seems like this warning is telling me that the project settings are being overridden by the project settings, which is both redundant and unnecessary. Thank you in advance for the clarification.

redapple:
"Overriden settings" is relative to `scrapy.settings.default_settings` (see https://github.com/scrapy/scrapy/blob/d4a8f43816adc695dd0b6f165ca3bc3b16bf469e/scrapy/settings/__init__.py#L512)
What happens with command line commands is that some commands do override a few default settings, and also a `scrapy startproject` generated project does create a `settings.py` file with some non-default settings (like[ `ROBOTSTXT_OBEY=True`](https://github.com/scrapy/scrapy/blob/master/scrapy/templates/project/module/settings.py.tmpl))

MalikRumi:
@redapple Thank you. I never understood that even something as basic as 'startproject' was overriding _anything_.  (-:

I realize this is not a general purpose help forum, and in fact I just took a break from plowing thru your github repo to get a better understanding (which I haven't found yet) but can you point me to something that explains crawler.settings, and how they are different from spider and project settings? I've been looking since last night and have not been able to find such explanation. I may be about to come across it in the repo, but a specific pointer to a specific reference would still be helpful. Thanks.

Issue #802, opened 2014-07-14
Per request delay


chekunkov:
Sometimes I feel like scrapy is missing per request delays. Any reasons why they weren't implemented?

Where can per request delays be used:
- to add exponential backoff for the retry request
- to add some huge delay when site maintenance is detected from request callback

I think there could be another situations where they can be useful.

Seems like this question was also raised in #254, but I don't like the way it was implemented there and could spend some time on another implementation and provide PR

Curita:
@chekunkov What do you think of adding this feature as an idea for this year's GSoC? Don't know if you've already started implementing it, but seems like a really fun and challenging project for a student, and provided of good mentoring (I'd love to have you as a mentor if you're interested) seems feasible on GSoC's timeframe.

chekunkov:
@Curita Yes, seems like a good idea, I haven't started working on it so it's okay to give this to someone as a GSoC project. Concerning your proposal to make me a mentor for this project - I need some time to think on this, not sure if I'll have enough time and energy to help students :)

rocioar:
hey @chekunkov @Curita I have this implemented for epubdirect project, will generalize it and make a PR, if that sounds good.

chekunkov:
@rocioar yay, sounds great I think

kmike:
Global DOWNLOAD_DELAY means "time between two requests for the same downloader slot (==website)". It is also possible to set the download delay per Downloader slot (website), though this API is neither public nor documented.

Delay is a time between two events, right? What does "per-request" delay mean? Which two events are separated by this delay?

umrashrf:
I think they meant the delay between the scheduler and downloader. So you schedule a request with delay and then it is downloaded after that delay.

nramirezuy:
@chekunkov Something like this makes sense to you?

```python
# DownloaderMiddleware
class PerRequestDelayMware(object):
    ...
    # signal
    def request_scheduled(self, request, spider):
        request.meta.setdefault('_scheduled_time', time.time())

    def process_request(self, request, spider):
        scheduled_time = request.meta.get('_scheduled_time')
        per_request_delay = request.meta.get('per_request_delay')
        if not scheduled_time or not per_request_delay:
            return
        if scheduled_time + per_request_delay < time.time():
            request.priority += some_negative_value
            return request
```

rocioar:
Well, my idea of it was:

[...]

PR #2897, opened 2017-08-24
[WIP][PoC] Allow per-slot concurrency and delay settings

redapple:
Motivation: https://stackoverflow.com/questions/45861515/how-to-set-up-different-download-delay-from-different-domains-scrapy-python

I've also seen use cases of the same spider contacting 2 different endpoints from the same main domain, e.g. www.example.com and api.example.com, but which have different throttling requirements.

This quick hack changes `Downloader` to check a new (optional) DOWNLOADER_SLOTS setting to create a new downloader Slot, 
so as to configure DOWNLOAD_DELAY and CONCURRENT_REQUEST_PER_DOMAIN
per-hostname instead of globally.

Obviously, this needs a bit more work to cover various cases:
- per-IP concurrency, 
- wildcard settings keys perhaps e.g. `*.example.com`
- check how this plays with AutoThrottle.
- check how this plays with `request.meta['download_slot']`, for example to support non-hostname keys
...

What do others think?

I played with this with this spider:
```
# -*- coding: utf-8 -*-
import scrapy


    name = 'example'
    allowed_domains = ['example.com',
                       'httpbin.org']
    start_urls = ['http://example.com/']

    def parse(self, response):
        for u in (['http://example.com/?q=%d' % i for i in range(11,21)] +
                  ['http://httpbin.org/get?q=%d' % i for i in range(21,51)]):
            yield scrapy.Request(u, callback=self.sink)

    def sink(self, response):
        pass
```

and these settings:
```
CONCURRENT_REQUESTS = 32
DOWNLOAD_DELAY = 2
DOWNLOADER_SLOTS = {
    'httpbin.org': {
        'DOWNLOAD_DELAY': 0.5,
        'CONCURRENT_REQUESTS_PER_DOMAIN': 2,
    }
}
```

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -11,6 +11,7 @@ from twisted.internet import reactor, defer, task
 from scrapy.utils.defer import mustbe_deferred
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.resolver import dnscache
+from scrapy.settings import Settings
 from scrapy import signals
 from .middleware import DownloaderMiddlewareManager
 from .handlers import DownloadHandlers
@@ -58,7 +59,7 @@ class Slot(object):
         )
 
 
-def _get_concurrency_delay(concurrency, spider, settings):
+def _get_concurrency_delay(concurrency, spider, settings, request=None):
     delay = settings.getfloat('DOWNLOAD_DELAY')
     if hasattr(spider, 'DOWNLOAD_DELAY'):
         warnings.warn("%s.DOWNLOAD_DELAY attribute is deprecated, use %s.download_delay instead" %
@@ -69,6 +70,14 @@ def _get_concurrency_delay(concurrency, spider, settings):
 
     if hasattr(spider, 'max_concurrent_requests'):
         concurrency = spider.max_concurrent_requests
+
+    if request is not None:
+        key = urlparse_cached(request).hostname or ''
+        slotkey = settings.get('DOWNLOADER_SLOTS', {}).get(key)
+        if slotkey:
+            slotsettings = Settings(slotkey)
+            delay = slotsettings.getfloat('DOWNLOAD_DELAY', delay)
+            concurrency = slotsettings.getint('CONCURRENT_REQUESTS_PER_DOMAIN', concurrency)
 
     return concurrency, delay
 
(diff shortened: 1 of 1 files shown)


kmike:
A related ticket: https://github.com/scrapy/scrapy/issues/802. In this ticket the problem is not only to change options per-slot, but also to change them as spider crawls. Think e.g. of exponential retry backoffs, or writing custom AutoThrottle using public documented Scrapy APIs. 

I can see how a static dict is useful, and it is easy to understand, but a public API for downloader slots may solve the same issue, as well as some other issues. I don't have a proposal now though :) Maybe this all is orthogonal.

kmike:
@rocioar's example from https://github.com/scrapy/scrapy/issues/802#issuecomment-79009198 looks quite similar to what this PR implements.

redapple:
@kmike , there is indeed a recurrent need for more complex requests dynamics.

From what I understand of `Downloader` and its `Slots`, it looks like `Downloader` is the one doing all the heavy job:
- pushing to slot queues (simple `deque()`), 
- popping from slot queues,
- moving requests to `slot.active` and `slot.transferring` sets
- it calls the download handlers

The one thing that `downloader.Slot` has is the delay and concurrency info, and it's somewhat static (optionally randomized but not dependent on individual requests, or patched by AutoThrottle).
The logic and orchestration is really in the downloader.

From the outside, it looks like queuing/dequeuing/call-me-later could be moved outside of Downloader and moved to some "Slot 2.0", with a new API between them (to be defined at this point).

For the time being, perhaps the API could be `Downloader.fetch(request, spider)`, `Downloader._enqueue_request(request, spider)` and `Downloader._process_queue(self, spider, slot)` even (meaning _"Downloader will call process_queue for each request"_. Users can already use their own `Downloader`, implementing their own queues. Which could be daunting if Scrapy's core doesn't provide a cleaner implementation.

In the end, reading #802 again, it seems to me that the Scheduler would be a better fit for this throttling and custom delays (I don't know is @dangra meant `core.scheduler.Scheduler` or another "request scheduler"). Right now, it _only_ does request sequencing. It already has queues, albeit without timing information.
Downloader has queues, but simpler queues (no persistence) but does some scheduling too in a sense (not fine-grained at the request level though).

wRAR:
This was implemented later.

PR #2898, opened 2017-08-24
2374 Update DepthMiddleware docs to mention request.meta['depth']

mxws:
Add description of how spider DepthMiddleware updates request.meta['depth'].
If depth is not present in request, it is added and set to 0. If present, the current value is incremented.


kmike:
Thanks @mxws! It looks good, but I've merged a slightly more elaborate https://github.com/scrapy/scrapy/pull/2963.

PR #2901, opened 2017-08-27
Fixed mail mimetype when it has attachments

sergiobellon:
An email cannot be sent in html format if there are attachments because it's set always to 'plain'

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -67,7 +67,7 @@ class MailSender(object):
             msg.set_charset(charset)
 
         if attachs:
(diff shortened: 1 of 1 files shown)


redapple:
@sergiobellon , would you have some sample code sending email with attachments that fail with current codebase?

sergiobellon:
@redapple It doesn't fail, but it doesn't send it as an html email. _Plain_ means text email, no code supported.

redapple:
Alright @sergiobellon .
Can you add a test for this in https://github.com/scrapy/scrapy/blob/master/tests/test_mail.py?

PR #2906, opened 2017-08-30
[WIP] Downloader timings in request.meta

redapple:
While working on a requests log plugin for Scrapy, I felt the need to get more fine-grained timing information on the various steps a `Request` goes through from reaching the `Downloader` to getting the associated response body.
One can work with `request.meta['download_delay']` and timing inserted from a downloader middleware but I found this additional information interesting for example for HAR output.

Note: This does not address DNS lookup time nor TCP/TLS connection time.

I would even set this to ON by default myself.

Missing:

- [ ] Documentation
- [ ] How does this play with proxies?
- [x] Handle download exceptions (DNS errors, dowload timeouts, connection errors etc.)


AdrianAtZyte:
> Everything this instruments can be obtained from signals nowadays, with no core changes and no setting: `request_reached_downloader` for the queueing time, `headers_received` and the first `bytes_received` for the response timings, and `request_left_downloader` plus `download_latency` for the end. The only one without a direct equivalent is the transfer start, and that one is the response time minus `download_latency`.

Shall we close?

PR #2915, opened 2017-09-07
Print cryptography package version

redapple:
`scrapy version -v` does not currently output `cryptography` version, which can be useful.

I've added a `try/except` when reading cryptography's version because I don't know when it became a dependency of `pyOpenSSL`. But it may be fine to remove it.
Scrapy's current baseline of Debian Jessie uses `pyOpenSSL==0.14` and `cryptography==0.6`

### DIFF
--- a/scrapy/commands/version.py
+++ b/scrapy/commands/version.py
@@ -23,8 +23,11 @@ class Command(ScrapyCommand):
 
     def run(self, args, opts):
         if opts.verbose:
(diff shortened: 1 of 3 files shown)


kmike:
+1 to add this feature, but there are some failing tests to be fixed.

redapple:
@kmike , done.

kmike:
Thanks @redapple!

Issue #482, opened 2013-12-05
S3FilesStore can use a lot of memory


kmike:
Hi,

@nramirezuy and me were debugging memory issue with one of the spiders  some time ago, and it seems to be caused by ImagesPipeline + [S3FilesStore](https://github.com/scrapy/scrapy/blob/master/scrapy/contrib/pipeline/files.py#L66). I haven't confirmed that it was the cause of memory issue, this ticket is based solely on reading the source code.

FilesPipeline reads the whole file to memory and then defers the uploading to thread (via `S3FilesStore.persist_file`, passing file contents as bytes). So there could be many files loaded to memory at the same time, and as soon as files are downloaded faster than they are are uploaded to s3, memory usage will grow. This is not unlikely IMHO because s3 is not super-fast. For ImagesPipeline it is worse because it uploads not only the image itself, but also the generated thumbnails.

I think S3FilesStore should persist files to temporary location before uploading them to S3 (at least optionally). This would allow streaming files without storing them in memory.

Parth-Vader:
Change the links to https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/files.py and https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/media.py since the ones linked are depriciated.

G-Rath:
Are folks aware this has been reported as a security issue that affects most (if not all) versions of Scrapy? https://nvd.nist.gov/vuln/detail/CVE-2017-14158 & https://github.com/pypa/advisory-database/blob/8b7a4d62a95e8f605e5dfb4e0b4f299e6403dc12/vulns/scrapy/PYSEC-2017-83.yaml

I'm currently trying to determine if the security angle has been fixed, and if not what is required for that since we're got this flagged in one of our applications; I'm happy to help with this as much as I can, but I don't work in Python a lot so suspect I'll be of limited use but I'm happy to be apart of discussions and even dive into code if someone more familiar with the codebase could spare some time to support me 🙂

fazalmajid:
This issue is also flagged as [PYSEC-2017-83](https://osv.dev/vulnerability/PYSEC-2017-83) by [pip-audit](https://github.com/trailofbits/pip-audit).

[...]

PR #2917, opened 2017-09-09
support giantfiles download ,fix CVE-2017-14158

wangtua1:
from the issue https://github.com/scrapy/scrapy/issues/482.
We can see if we download a file with filespipeline, the whole file is stored into the memory.
If the file is a giant file or an extremely giant file,(over 1G or more), the crawling thread will be crashed.
We can use this POC to prove it.
[POC.zip](https://github.com/scrapy/scrapy/files/1289514/POC.zip)

This has been asigned CVE-2017-14158.
but in this new testcase using GiantFilesPipeline in the commit,we can download it correctly.

[testdownload.zip](https://github.com/scrapy/scrapy/files/1289512/testdownload.zip)

### DIFF
--- a/scrapy/cmdline.py
+++ b/scrapy/cmdline.py
@@ -11,7 +11,6 @@ from scrapy.commands import ScrapyCommand
 from scrapy.exceptions import UsageError
 from scrapy.utils.misc import walk_modules
 from scrapy.utils.project import inside_project, get_project_settings
-from scrapy.utils.python import garbage_collect
 from scrapy.settings.deprecated import check_deprecated_settings
 
 def _iter_command_classes(module_name):
@@ -166,9 +165,4 @@ def _run_command_profiled(cmd, args, opts):
         p.dump_stats(opts.profile)
 
 if __name__ == '__main__':
-    try:
-        execute()
-    finally:
-        # Twisted prints errors in DebugInfo.__del__, but PyPy does not run gc.collect()
-        # on exit: http://doc.pypy.org/en/latest/cpython_differences.html?highlight=gc.collect#differences-related-to-garbage-collection-strategies
-        garbage_collect()
+    execute()
--- a/scrapy/contrib/pipeline/giantfiles.py
+++ b/scrapy/contrib/pipeline/giantfiles.py
@@ -1,0 +1,2 @@
+
+from scrapy.pipelines.giantfiles import *
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -2,6 +2,9 @@
 
 import re
 import logging
+import hashlib
+import os.path
+
 from io import BytesIO
 from time import time
 import warnings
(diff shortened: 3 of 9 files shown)


kmike:
I'm not sure it is a security issue: currently Scrapy keeps response data in memory, but one can use DOWNLOAD_MAXSIZE option to limit max size of downloaded files. 

Providing an option to temporarily store response content on disk instead of RAM makes a lot of sense to me though; from a high level workaround makes sense 👍 

I think that ideally this should be transparent - there could be an option which says "if response is over X bytes, use on-disk storage". It won't help for responses which content is parsed as HTML (e.g. to run selectors), but could work for file pipelines or other cases response content can be stored as-is or processed iteratively.

This feature may require API discussion: 

* how should response.body work - should it be lazy? what about middlewares which access e.g. `response.body[:4096]`? I don't think storing file path in response.body instead of real response body is a good enough API, it breaks everything.
* how should user decide if a response has content in memory or on disk? should user care?
* should we disallow css/xpath for such responses?
* what about response.text?
* how to allow reading response body iteratively? provide file-like interface for response body?
* this seems to be related to another missing Scrapy feature - one can't get partial content, scrapy always waits until full response is downloaded.
* how to specify path to store temporary content? is temp dir enough, or should this data be stored in project folder? how are these files get removed? should we limit their total size?

When implementing this, we'd need to make sure the feature plays well with Scrapy built-in middlewares and extensions (e.g. how does it work with gzip decompression middleware?); the feature needs docs and tests.

@wangtua1 we can't merge PR as-is, but thanks for making it: it raises an important issue, and provides a draft implementation, which is always useful to have. Are you interested in working on it further?

wangtua1:
So I see the reasons why it can't be merged. But I'm still interested in working on scrapy project. I also hope scrapy can be better in future.

wangtua1:
So I see the reasons why it can't be merged. But I'm still interested in working on scrapy project. I also hope scrapy can be better in future.

synhershko:
@kmike so, this is now considered by Snyk a High Severity security risk of Scrapy, hopefully this gets fixed somehow soon..

kmike:
@synhershko unfortunately it won't get fixed on its own, someone needs to work on it. I'd love to have this feature in Scrapy, but It looks like a large feature which requires effort. There is an easy workaround for users: decrease DOWNLOAD_MAXSIZE, which is 1GB by default (https://doc.scrapy.org/en/latest/topics/settings.html#download-maxsize) - memory usage is not unlimited even now. 

It is a problem for sure, but I still don't get why it a high severity security risk; it is just a DDoS attack with an easy built-in workaround. Should we e.g. assign high severity CVEs to requests library, which doesn't have a limit on download size at all, unlike Scrapy? Or for all text editors which load text files to memory in full?

grammy-jiang:
@kmike 

I was also noticed by Snyk about this issue yesterday. This could be a problem for some newbie to Scrapy who have not known the keyword of `DOWNLOAD_MAXSIZE`.

Could this be a security issue? I have some different view. There are certain keywords in `scrapy.settings.default_settings` related to the size/quota of responses/items. And also these keywords have not been well explained in the documents, or the users are so hurrying to run their spiders to ignore reading the documents/source code carefully. So I would like to say this is a documentation problem - more detail explanation should be added - rather than a security issue.

Considering other tools/modules/frameworks in the open source community, Scrapy has done a lot of work, and it is so awesome that the influence of Scrapy is spreading to many other languages. Any others' problems/issues will not make Scrapy great - the greatness of Scrapy itself makes Scrapy great. That great makes more people watching Scrapy - they may be a software engineer with more than 10 years expr., or maybe a newbie in the open source community like a 3 yrs old child without well behaviour.

So let's be peaceful and face this maybe issue, find a way to solve it.

PR #2918, opened 2017-09-09
Fix SIGINT handling when using inspect_response

HarrisonGregg:
Currently, after calling `scrapy.shell.inspect_response` and then closing the opened shell, SIGINT (Ctrl-C) no longer works to terminate the spider.  This is because `Shell.start`, called in `inspect_response`, removes the signal handler.  To fix this, save the SIGINT handler before calling `Shell.start` in `inspect_response` and add it again after the shell has closed.

It doesn't appear that there are any tests for `inspect_response`, so I haven't added any tests for this fix.

### DIFF
--- a/scrapy/shell.py
+++ b/scrapy/shell.py
@@ -164,7 +164,11 @@ class Shell(object):
 
 def inspect_response(response, spider):
     """Open a shell to inspect the given response"""
+    # Shell.start removes the SIGINT handler, so save it and re-add it after
+    # the shell has closed
+    sigint_handler = signal.getsignal(signal.SIGINT)
     Shell(spider.crawler).start(response=response, spider=spider)
+    signal.signal(signal.SIGINT, sigint_handler)
 
 
 def _request_deferred(request):


wRAR (approved):

Gallaecio:
Ideally we should include a test before merging. I wonder if subprocess with shell=True will allow to write a test for this.

PS: Closing and reopening to trigger a new CI run.

Gallaecio:
Besides having a test, I wonder if there is a different way we should approach it, either removing the code in `Shell.start` that disables Ctrl+C, having `Shell.start` re-enable it, or add a parameter to `Shell.start` to control this.

The comment and the [commit](https://github.com/scrapy/scrapy/commit/1d3b9e2ca8bd4adc73692317e17ad6c569ab0705#diff-a4e7908c2400926ebd2f76192adb9fcfe20e6a5b7254f16ac0fbae082773991b) that introduced the issue do not make it clear to me why we need that code in the first place. The comment says it is about preventing an accidental stop of the engine, but that is precisely what we want here.

Gallaecio:
After playing for a while with subprocess, I have been unable to come up with a test, so I am OK with merging as is.

Gallaecio (approved):

kmike:
Thanks @HarrisonGregg!

PR #2922, opened 2017-09-12
[Doc] Update Response.body type

stav:
Noted by `gr3yh47` in `#scrapy` IRC channel.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -525,11 +525,11 @@ Response objects
        (for single valued headers) or lists (for multi-valued headers).
     :type headers: dict
 
-    :param body: the response body. It must be str, not unicode, unless you're
-       using a encoding-aware :ref:`Response subclass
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -525,11 +525,11 @@ Response objects
        (for single valued headers) or lists (for multi-valued headers).
     :type headers: dict
 
-    :param body: the response body. It must be str, not unicode, unless you're
-       using a encoding-aware :ref:`Response subclass
+    :param body: the response body. To access the decoded text as str you can
```
kmike: `str` is not right in Python 2; what do you think about adding something like "decoded text as str (unicode in Python 2)"?

stav: Python 2?  What's that? ha 
Sure thing.

kmike:
Thanks @stav!

PR #2929, opened 2017-09-17
[MRG+1] Add m4v extension to IGNORED_EXTENSIONS in LinkExtractor.

### LINKED ISSUE #2928: LinkExtractor is not ignoring .m4v extension (video) by default
By chance I found out that LinkExtractor is not ignoring the video extension m4v in the same way it is ignoring other video formats.

https://en.wikipedia.org/wiki/M4V

djunzu:
Add m4v extension (video) to `IGNORED_EXTENSIONS` in LinkExtractor so it can be ignored by default in the same way other video formats are already ignored.

fix #2928


kmike:
Looks good to me.
Related: https://github.com/scrapy/scrapy/pull/2067.

PR #2931, opened 2017-09-18
Update install.rst

mGalarnyk:
Updated installation instructions for installing scrapy using conda.

You can now just do: 

```
conda install scrapy
```


kmike:
I'm not anaconda user myself, but won't it install Scrapy 1.3.3 by default (https://anaconda.org/anaconda/scrapy)?

PR #2950, opened 2017-10-03
WIP: loader add_* funcs: pass **kw to self.selector.xpath

immerrr:
I have found out that kwargs that one specifies for `loader.add_xpath` are not getting to the actual `selector.xpath` invocation, which looks like a bug.

I wonder if there's code out there that would be broken by this fix...


immerrr:
@kmike @dangra WDYT?

kmike:
These kwargs are eventually passed to get_value method. It seems the only tricky kwarg is 're'. But it doesn't look clean that kwargs are used for 2 completely different purposes.

immerrr:
Indeed it doesn't look too clean. What if I pass xpath-related keywords as `xpath_kw` parameter?

Gallaecio:
Given the issue about this has been moved to https://github.com/scrapy/itemloaders/issues/32, please consider closing this PR and creating a similar PR for https://github.com/scrapy/itemloaders

Issue #2927, opened 2017-09-17
Unhelpful log message from core.downloader.handlers.http11


djunzu:
Consider the two log messages:

```txt
2017-09-17 10:52:27 [scrapy.core.downloader.handlers.http11] WARNING: Expected response size (42276161) larger than download warn size (33554432).
2017-09-17 10:54:18 [scrapy.core.downloader.handlers.http11] WARNING: Received more bytes than download warn size (33554432) in request <GET https://example.com>.
```

While the second one can be useful, the first one is not helpful at all because there is no info about which request it is about. The first message should indicate the request it refers to in the same way the second message does.

NoExitTV:
Can I have the code you used to get these error messages?

djunzu:
> Can I have the code you used to get these error messages?

@NoExitTV, sorry, but I can not disclose this project's source.

I could try to give you a url that generates this message... But I don't know which url generated it in the first place, and that is the issue! But I am guessing it was generated by a .m4v file (see #2928).

NoExitTV:
@djunzu No worries. I figured it out as you've seen in my PR [#2952](https://github.com/scrapy/scrapy/pull/2952)

kmike:
Fixed by https://github.com/scrapy/scrapy/pull/2952 - thanks @NoExitTV and @djunzu!

PR #2952, opened 2017-10-05
[MRG+1] Changed unhelpful log message from core.downloader.handlers.http11

NoExitTV:
Changed the warning message to give information about the request generating it
as requested in: [#2927](https://github.com/scrapy/scrapy/issues/2927)

Tested with spider:
```
import scrapy


    name = "test_spider"

    custom_settings = {
    	'DOWNLOAD_WARNSIZE': '123',
    }

    def start_requests(self):
        urls = [
            'http://example.com',
        ]
        for url in urls:

    def parse(self, response):
        page = response.url.split("/")[-2]
        filename = 'test_spider-%s.html' % page
        with open(filename, 'wb') as f:
            f.write(response.body)
        self.log('Saved file %s' % filename)
```

Which gives the requested warning message:
```
2017-10-05 15:41:01 [scrapy.core.downloader.handlers.http11] WARNING: Expected response size (606) larger than download warn size (123) in request <GET http://example.com>.
```

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -412,9 +412,10 @@ class _ResponseReader(protocol.Protocol):
 
         if self._maxsize and self._bytes_received > self._maxsize:
             logger.error("Received (%(bytes)s) bytes larger than download "
-                         "max size (%(maxsize)s).",
+                         "max size (%(maxsize)s) in request %(request)s.",
                          {'bytes': self._bytes_received,
-                          'maxsize': self._maxsize})
+                          'maxsize': self._maxsize,
+                          'request': self._request})
             # Clear buffer earlier to avoid keeping data in memory for a long
             # time.
             self._bodybuf.truncate(0)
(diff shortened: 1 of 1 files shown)


scrapy/core/downloader/handlers/http11.py
```diff
@@ -412,9 +412,10 @@ def dataReceived(self, bodyBytes):
 
         if self._maxsize and self._bytes_received > self._maxsize:
             logger.error("Received (%(bytes)s) bytes larger than download "
-                         "max size (%(maxsize)s).",
+                         "max size (%(maxsize)s) in request %(request)s.",
                          {'bytes': self._bytes_received,
-                          'maxsize': self._maxsize})
+                          'maxsize': self._maxsize,
+                          'request': self._request})
```
djunzu: I think this is right and could be also included in this PR.
It is exactly the same way it was done in the next if ([line 425](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/handlers/http11.py#L425)).

djunzu:
Looks good to me.

Issue #2925, opened 2017-09-14
HtmlResponse check in _requests_to_follow method


Igor4uk:
Is it necessary to check for HtmlResponse here https://github.com/scrapy/scrapy/blob/master/scrapy/spiders/crawl.py#L57 ?
My site start return content-type:application/json and scrapy recognise it like a TextResponse and finish parsing after the first request to it

kmike:
If the site contains JSON then it is a right thing to do, as linkextractor works on html. If site uses a wrong content-type then it makes sense to remove the check, but I'm not sure this is common; I feel that overriding it in a subclass just for your case can be a better solution than removing a check for all users.

Igor4uk:
Maybe you are right, but by the way, I think that it is better to add more explicit debug log for such kind of error. I spend lot of time trying to deal with it

kmike:
Adding a debug message ("Skipping non-html response {}") sounds good to me. Are you up to sending a pull request?

kmike:
Hm, it can be quite verbose though if the goal of a spider is to crawl binary data..

Igor4uk:
In my case we start parsing from json objects which contains ids for links, using custom overrides linkextractor class

Igor4uk:
Maybe it will be better and faster if you create pull request? I'm not very familiar with scrapy and github workflow.

phnk:
@Igor4uk Do you still have the code which would allow me to reproduce the behaviour?

Igor4uk:
I think it will be enough. Thanks

PR #2953, opened 2017-10-06
Added debug message in spiders/crawl

phnk:
As requested in #2925.

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -56,7 +56,7 @@ class CrawlSpider(Spider):
         return r
 
     def _requests_to_follow(self, response):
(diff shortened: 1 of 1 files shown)


scrapy/spiders/crawl.py
```diff
@@ -55,6 +57,8 @@ def _build_request(self, rule, link):
 
     def _requests_to_follow(self, response):
         if not isinstance(response, HtmlResponse):
+            logger.debug("Skipping non-html response %(response)s" ,
```
kmike: I think it should use self.logger

phnk: Ok. I'll try to fix it as soon as I have time.

PR #2954, opened 2017-10-06
spiders: add OAI-PMH support WIP

kaplun:
Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

### DIFF
--- a/scrapy/core/downloader/handlers/dummy.py
+++ b/scrapy/core/downloader/handlers/dummy.py
@@ -1,0 +1,14 @@
+"""Download handlers for OAI-PMH
+"""
+
+from sickle import Sickle
+
+from scrapy.http import Response
+
+class DummyDownloadHandler(object):
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def download_request(self, request, spider):
+        url = request.url
+        return Response(url, request=request)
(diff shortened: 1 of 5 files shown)


kaplun:
This is a proposal for introducing a core spider to crawl OAI-PMH servers (https://www.openarchives.org/pmh/). This is a standard transport protocol used in the world of digital libraries to expose metadata about records. E.g. notably http://arXiv.org exposes its data via this protocol, and also http://inspirehep.net/ and many other scientific repositories.

In Python, the sickle library (https://pypi.python.org/pypi/Sickle) nicely implements this protocol, thus allowing a simple implementation in Scrapy.

What do you think? Would this be something you might be interested into including into the official code base?

david-caro:
@szymonlopaciuk

cathalgarvey:
Hi @kaplun - Sorry for the long silence!

I've found myself doing the same thing, in the past; using Scrapy to access OAI-PMH endpoints. And, I also used Sickle to do the hard work!

Unfortunately, my experience was very negative, and I don't think that Scrapy is a good way to do harvesting from endpoints of this type. In my case, I was dealing with an endpoint that sometimes failed part-way through a harvest.. and because the harvest is occurring within Sickle, Scrapy sees a single failure and kills the whole thing. This is an appropriate way for Scrapy to handle Scraping, but it's a problem for a harvesting-based workflow like Sickle's.

Thanks very much for your proposed edits! As things stand though, I'd be against merging them. Much like Scrapy isn't a good basis for a Torrent client, I also feel that Scrapy is a poor base for an OAI harvester. I'm open to arguments to the contrary, though?

Issue #1304, opened 2015-06-15
Scrapy should handle "invalid" relative URLs better


kmike:
Currently Scrapy can't extract links from http://scrapy.org/ page correctly because urls in page header are relative to a non-existing parent: `../download/`, `../doc/`, etc. Browsers resolve these links as `http://scrapy.org/download/` and `http://scrapy.org/doc/`, while response.urljoin, urlparse.urljoin and our ink extractors resolve them as `http://scrapy.org/../download/`, etc. This results in 400 Bad Request responses. 

urlparse.urljoin is not correct (or not modern) here. In the [URL Living Standard](https://url.spec.whatwg.org/#relative-path-state) for browsers it is said:

> If buffer is "..", remove url’s path’s last entry, **if any**, and then if c is neither "/" nor "\", append the empty string to url’s path.

kmike:
By the way, this also fails:
```
yield scrapy.Request("http://scrapy.org", self.parse)
```
*-* the request is filtered out because of https://github.com/scrapy/scrapy/issues/1225 and self.parse is never called. Crawling scrapy.org with Scrapy is not straightforward :)

kmike:
The most evil spider ever: looks innocent, but doesn't work for multiple reasons
```
import scrapy

class ScrapySpider(scrapy.Spider):
    name = 'scrapyspider'

    def start_requests(self):
        yield scrapy.Request("http://scrapy.org", self.parse_main)

    def parse_main(self, response):
        for href in response.xpath("//a/@href").extract():
            yield scrapy.Request(response.urljoin(href), self.parse_link)

    def parse_link(self, response):
        print(response.url)
```

jlbren:
Hello, myself and some colleagues are looking into this

kmike:
@jlbren yeah, it is still open; help is appreciated!

jlbren:
@kmike great, we will post updates as they come!

kmike:
Thanks @jlbren! Feel free to ask any questions and discuss it here. The issue doesn't look easy because it is a bug in urlparse.urljoin from standard library. Likely we'll have to add a workaround to https://github.com/scrapy/w3lib, but maybe there is another way to solve this, I don't know.

jlbren:
@kmike thanks, that is a helpful starting point. We will definitely take you up on posting questions. I am working with a team of 2 other CS grad students so our aim is to try and make significant progress as a semester project. I imagine they will post any questions they encounter as well.

[...]

PR #2955, opened 2017-10-07
Handle "invalid" relative URL issue #1304

NoExitTV:
Did some minor tweaks on how scrapy handle relative URL's as discussed in #1304 

Tested it with some basic code in the scrapy shell:
```
>>> resp = scrapy.http.response.html.HtmlResponse('http://www.example.com',
...      body='''<html>
...                  <body>
...                      <a href="../index1.html">Link1</a>
...                      <a href="../../index2.html">Link2</a>
...                      <a href="index3.html">Link3</a>
...                      <a href="other_html/../index4.html">Link4</a>
...                      <a href="other_html/folder2/../index5.html">Link5</a>
...                      <a href="other_html/index6.html">Link6</a>
...                      <a href="../other_html/index7.html">Link7</a>
...                 </body>
...          </html>''')
>>> links = scrapy.linkextractors.LinkExtractor().extract_links(resp)
>>> links
[Link(url='http://www.example.com/index1.html', text='Link1', fragment='', nofollow=False), Link(url='http://www.example.com/index2.html', text='Link2', fragment='', nofollow=False), Link(url='http://www.example.com/index3.html', text='Link3', fragment='', nofollow=False), Link(url='http://www.example.com/index4.html', text='Link4', fragment='', nofollow=False), Link(url='http://www.example.com/other_html/index5.html', text='Link5', fragment='', nofollow=False), Link(url='http://www.example.com/other_html/index6.html', text='Link6', fragment='', nofollow=False), Link(url='http://www.example.com/other_html/index7.html', text='Link7', fragment='', nofollow=False)]
```

Passed the same unit tests as the original code when running tox
```
========================================================= 1679 passed, 6 skipped, 14 xfailed in 458.89 seconds =========================================================

_______________________________________________________________________________ summary ________________________________________________________________________________

  py27: commands succeeded
  congratulations :)
```

I believe that scrapy now handle relative url's as expected in python 2.7.13
What are your thoughts?

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -58,6 +58,8 @@ class LxmlParserLinkExtractor(object):
         for el, attr, attr_val in self._iter_links(selector.root):
             # pseudo lxml.html.HtmlElement.make_links_absolute(base_url)
             try:
+                print "PRINTING attr_val"
+                print attr_val
                 if self.strip:
                     attr_val = strip_html5_whitespace(attr_val)
                 attr_val = urljoin(base_url, attr_val)
@@ -68,9 +70,6 @@ class LxmlParserLinkExtractor(object):
                 if url is None:
                     continue
             url = to_native_str(url, encoding=response_encoding)
-            print "DEBUG PRINT"
-            print url
-            print "DEBUG END"
             # to fix relative links after process_value
             url = urljoin(response_url, url)
             link = Link(url, _collect_string_content(el) or u'',
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -66,7 +66,6 @@ class LxmlParserLinkExtractor(object):
             try:
                 if self.strip:
                     attr_val = strip_html5_whitespace(attr_val)
-                attr_val = self.check_for_invalid_sublink(attr_val)
                 attr_val = urljoin(base_url, attr_val)
             except ValueError:
                 continue  # skipping bogus links
@@ -74,6 +73,10 @@ class LxmlParserLinkExtractor(object):
                 url = self.process_attr(attr_val)
                 if url is None:
                     continue
+            print "### DEBUG PRINT BEFORE FUNCTION ###"
+            url = self.check_for_invalid_sublink(url)
+            print url
+            print "### DEBUG PRINT END ###"
             url = to_native_str(url, encoding=response_encoding)
             # to fix relative links after process_value
             url = urljoin(response_url, url)
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -52,11 +52,11 @@ class LxmlParserLinkExtractor(object):
                     continue
                 yield (el, attrib, attribs[attrib])
 
-    def check_for_invalid_sublink (self, link):
-            invalid_link_substring = "../"
-            if invalid_link_substring in link:
-                link = link.replace("../", "")
-            return link
+    def check_for_invalid_sublink (self, url):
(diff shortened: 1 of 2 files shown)


djunzu:
You are just stripping out `../`. I bet there thousands of web pages with working links that would not work with this patch.

NoExitTV:
@djunzu note that I'm stripping out the `../` after the urljoin function have done it's thing. Look at the relative paths in the example I gave with the links. I think that the only time there will be a `../` to strip is when there's no folder to descend back to? Or am I wrong?

Yes, it won't work if the html file name includes `../` but is that even possible?

Feedback is highly appreciated.

djunzu:
@NoExitTV , ignore what I said. What you said makes sense.

You still need to add tests to your PR.

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -52,6 +52,12 @@ def _iter_links(self, document):
                     continue
                 yield (el, attrib, attribs[attrib])
 
+    def check_for_invalid_sublink (self, url):
+            invalid_url_substring = "../"
+            if invalid_url_substring in url:
+                url = url.replace("../", "")
+            return url
```
kmike: I'd like for it to be more conservative:

1) shouldn't we only check for '../' in path? Can it appear e.g. in query or in fragment?
2) if Python 3 works correctly, it makes sense to remove a check, to make things faster;
3) is it possible to have more than one '../' in URL? Is it possible to have them in any part of a path, not only right after the hostname? If no, it'd be good to be more conservative and only e.g. remove '../' after the hostname.

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -52,6 +52,12 @@ def _iter_links(self, document):
                     continue
                 yield (el, attrib, attribs[attrib])
 
+    def check_for_invalid_sublink (self, url):
```
kmike: A better place for it would be a function in scrapy.utils.url (or even in w3lib.url). It should have a docstring which explains the issue, and tests.

LinkExtractor is not the only place which handles relative URLs - it is also done e.g. in response.urljoin and response.follow.

kmike:
@NoExitTV thanks for the PR! This is a very risky change as it affects a lot of spiders, so we need to make sure it covers all edge cases; we shouldn't take it lightly.

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -52,6 +52,12 @@ def _iter_links(self, document):
...
+            invalid_url_substring = "../"
```
kmike: could you please use 4-space indentation, as per pep8?

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -52,6 +52,12 @@ def _iter_links(self, document):
...
+            if invalid_url_substring in url:
```
kmike: creating this `invalid_url_substring` constant doesn't make sense to me, it is not used consistently (e.g. in replace), not named as a constant (upper-case letters), and doesn't make code more readable. I think it is better to just use '../'.

scrapy/linkextractors/lxmlhtml.py
nctl144: > is it possible to have more than one '../' in URL?

@kmike I think it is possible if a static website has this kind of structure:
&nbsp;&nbsp;file1.html
&nbsp;&nbsp;folder1/folder2/file2.html

and they are trying to make a link to file1.html in file2.html. In this case, they would have to use **../../** , right?
So I think the worst case is the url in which it contains multiple **../** after the hostname. 

And what is the query and fragment that you were mentioning about? Please let me know! I am trying to make this solution available in w3lib instead!

PR #2956, opened 2017-10-09
Add from_crawler support to dupefilters

### LINKED ISSUE #2940: Give `BaseDupeFilter` access to spider-object
I am in a situation where a single item gets defined over a sequence of multiple pages, passing values between the particular callbacks using the `meta`-dict. I believe this is a common approach among scrapy-users.

However, it feels like this approach is difficult to get right. With the default implementation of `RFPDupefilter`, my callback-chain is teared apart quite easy, as fingerprints don't take the meta-dict into account. The corresponding requests are thrown away, the information in the meta-dict which made this request unique is lost.

I have currently implemented by own meta-aware DupeFilter, but I am still facing the problem that it lacks access to the specific spider in use - and only the Spider really knows the meta-attributes that make a request unique. I could now take it a step further and implement my own scheduler, but I'm afraid that all these custom extensions make my code very brittle wrt future versions of scrapy.

elacuesta:
Fixes #2940

### DIFF
--- a/tests/test_dupefilters.py
+++ b/tests/test_dupefilters.py
@@ -6,9 +6,30 @@ import shutil
 from scrapy.dupefilters import RFPDupeFilter
 from scrapy.http import Request
 from scrapy.utils.python import to_bytes
+from scrapy.utils.job import job_dir
+from scrapy.utils.test import get_crawler
 
 
 class RFPDupeFilterTest(unittest.TestCase):
+
+    def test_dupefilter_from_crawler(self):
+
+        class FromCrawlerRFPDupeFilter(RFPDupeFilter):
+
+            @classmethod
+            def from_crawler(cls, crawler):
+                debug = crawler.settings.getbool('DUPEFILTER_DEBUG')
+                df = cls(job_dir(crawler.settings), debug)
+                df.user_agent = crawler.settings.get('USER_AGENT')
+                return df
+
+        crawler = get_crawler(settings_dict={'DUPEFILTER_DEBUG': True, 'USER_AGENT': 'test ua'})
+        dupefilter = FromCrawlerRFPDupeFilter.from_crawler(crawler)
+
+        self.assertTrue(dupefilter.debug)
+        self.assertEqual(dupefilter.user_agent, 'test ua')
+
+        dupefilter.close('finished')
 
     def test_filter(self):
         dupefilter = RFPDupeFilter()
--- a/tests/test_dupefilters.py
+++ b/tests/test_dupefilters.py
@@ -5,31 +5,49 @@ import shutil
 
 from scrapy.dupefilters import RFPDupeFilter
 from scrapy.http import Request
+from scrapy.core.scheduler import Scheduler
 from scrapy.utils.python import to_bytes
 from scrapy.utils.job import job_dir
 from scrapy.utils.test import get_crawler
 
 
+class FromCrawlerRFPDupeFilter(RFPDupeFilter):
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        debug = crawler.settings.getbool('DUPEFILTER_DEBUG')
+        df = cls(job_dir(crawler.settings), debug)
+        df.method = crawler.settings.get('METHOD')
+        return df
+
+
+class FromSettingsRFPDupeFilter(RFPDupeFilter):
+
+    @classmethod
+    def from_settings(cls, settings):
+        debug = settings.getbool('DUPEFILTER_DEBUG')
+        df = cls(job_dir(settings), debug)
+        df.method = settings.get('METHOD')
+        return df
+
+
 class RFPDupeFilterTest(unittest.TestCase):
 
-    def test_dupefilter_from_crawler(self):
+    def test_from_crawler_scheduler(self):
+        settings = {'DUPEFILTER_DEBUG': True, 'METHOD': 'from_crawler',
+                    'DUPEFILTER_CLASS': __name__  + '.FromCrawlerRFPDupeFilter'}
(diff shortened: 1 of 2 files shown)


scrapy/dupefilters.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, path=None, debug=False):
             self.fingerprints.update(x.rstrip() for x in self.file)
 
     @classmethod
+    def from_crawler(cls, crawler):
+        return cls.from_settings(crawler.settings)
```
kmike: why is this method needed?

tests/test_dupefilters.py
```diff
@@ -6,10 +6,31 @@
...
+
+            @classmethod
+            def from_crawler(cls, crawler):
+                debug = crawler.settings.getbool('DUPEFILTER_DEBUG')
+                df = cls(job_dir(crawler.settings), debug)
+                df.user_agent = crawler.settings.get('USER_AGENT')
+                return df
+
+        crawler = get_crawler(settings_dict={'DUPEFILTER_DEBUG': True, 'USER_AGENT': 'test ua'})
+        dupefilter = FromCrawlerRFPDupeFilter.from_crawler(crawler)
```
kmike: This test doesn't check that Scheduler calls from_crawler method. I think this tests would pass even before all changes in this PR, so it is not really testing the new feature.

scrapy/dupefilters.py
elacuesta: Hey @kmike! Thanks for the quick response.
First I thought about leaving only `from_crawler`, but to keep this backwards compatible (and also because of your comment in the original issue: "add from_crawler support to dupefilters _in addition_ to from_settings") I kept both methods. It follows the same chain of precedence that can be found in https://github.com/scrapy/scrapy/blob/1.4/scrapy/middleware.py#L35-L40

kmike: It makes sense to follow the same logic, but if you remove from_crawler method from this class, it would do exactly the same (i.e. from_crawler doesn't seem to be needed here). 

Also, if you subclass from a dupefilter and override from_setting, overridden method won't be called after this PR.

tests/test_dupefilters.py
elacuesta: You're right, but to be fair, the current tests don't check the way the Scheduler creates the dupefilter either :-)

I can add a test for that.

scrapy/dupefilters.py
elacuesta: You're right, I see your point now, thanks! I'll remove it :+1:

elacuesta: @kmike: Related to this: is it possible that the `MiddlewareManager` is affected by this too? If I'm not mistaken, all of its subclasses (DownloaderMiddlewareManager, SpiderMiddlewareManager, ExtensionManager, ItemPipelineManager) are created using `from_crawler`, and none of them are supposed to be overridden by the user (except maybe the pipeline manager using the `ITEM_PROCESSOR` setting, but that it's not even documented). The only occurrence of the manager using `from_settings` is in the `tests/test_middleware.py` file (`test_enabled_from_settings`)
What do you think about some cleanup in https://github.com/scrapy/scrapy/blob/1.4/scrapy/middleware.py#L28-L58 to keep only the `from_crawler` method? In a different PR, of course.

johtso:
Any update on this?

tests/test_dupefilters.py
```diff
@@ -5,11 +5,50 @@
...
+        debug = settings.getbool('DUPEFILTER_DEBUG')
+        df = cls(job_dir(settings), debug)
+        df.method = settings.get('METHOD')
+        return df
 
 
 class RFPDupeFilterTest(unittest.TestCase):
 
+    def test_from_crawler_scheduler(self):
+        settings = {'DUPEFILTER_DEBUG': True, 'METHOD': 'from_crawler',
```
kmike: I think it makes sense to set method in dupefilter itself - if I'm not mistaken, currently these tests would pass if you change the dupefilter class (if a wrong method is called).

kmike:
@johtso thanks for the ping :)

@elacuesta I think this PR looks good, it needs just a small testing tweak. It'd be also good to add a test for dupefilters without from_crawler/from_settings methods. What's the reason for supporting them, by the way?

elacuesta:
Hello Mikhail! I think I addressed your latest comments (setting the string to compare in the class itself, add a test for dupefilters created without from_crawler/from_settings methods). Not sure what would be the case for the directly created dupefilters, maybe someone needs to do some initialization but doesn't need the crawler nor the settings and just a child class with a custom constructor is enough?

kmike:
Thanks @elacuesta! Test coverage is not complete because tests still don't check that dupefilter without from_crawler / from_settings methods work: dupefilter you're using inherits from a base class which has these methods.

elacuesta:
I didn't realize that before, thanks! I changed the test case, the new class doesn't implement any of the dupefilter methods but I think that's not the point of the test, it's just to ensure the right class is used.

elacuesta:
Ping @kmike :innocent:

kmike:
This looks good 👍  

However, I'd prefer to avoid copy-paste, and use the same function to create middlewares/extensions and dupefilters. Such function can be found in https://github.com/scrapy/scrapy/pull/1605, which is almost ready to merge as well; I wonder if you're up to finishing it, and using as a base for your PR :)

elacuesta:
Updated to use `scrapy.utils.misc.create_instance`. The diff shows unrelated changes but that should go away after merging #3348

dangra:
hi @elacuesta, can you rebase now that #3348 is merged? thanks

dangra:
Merge +1  sorry 🙄

PR #2962, opened 2017-10-16
Add signal `request_downloading` called right before the download handler

isra17:
It was needed for one of our project, maybe it can be useful for core as well.

The `request_downloading` signal is sent when the engine is about to download a request. If one handler raise an exception, the download is aborted. The signal supports returning deferreds from their handlers.

This is an alternative to the Downloader middlewares where there might be a significant delay between the middleware call and the download handler in case of slow queue processing.
This even handler allow some extension to tamper the request right before the download time and possibly cancel the request by raising an exception.

`signal.send_deferred` was needed for the exception handling raised by signal handler in the download manager.


IAlwaysBeCoding:
I don't really like the idea of doing errback on failed signals. I think the solution would be to introduce a some sort of signal middleware that can catch signal exceptions. Adding too many callbacks and errbacks to handle signal handling is going to get Scrapy into callback hell territory soon.

Also, why did you add yourself to the contributor list? You know that list hasn't been edited in about 4 years. I don't think you can add yourself there without getting the main approval from @kmike, @pablohoffman  or @redapple . However, I do believe that list should be updated to reflect everyone who has contributed.

isra17:
>I don't really like the idea of doing errback on failed signals. I think the solution would be to introduce a some sort of signal middleware that can catch signal exceptions. Adding too many callbacks and errbacks to handle signal handling is going to get Scrapy into callback hell territory soon.

Using signals here is just much simpler than using full blown middleware chain. Especially since it fits only a few specific use case and most of the time DownloaderMiddleware should be used. I don't see where you see too many callback, from the user point of view it's just another signal. From scrapy, the handling is literally one line: `dfd = self.signals.send_deferred(...)`. I don't see the hell you are talking about.

>Plus, you should use the send_catch_log_deferred() method instead of implementing send_deferred().

Those are not the same... `send_catch_log_deferred` swallows errors and the deferred always callback. `send_deferred` errbacks when an handler fail just like `pydispatcher.dispatcher.send` raise when one of the handlers do.

>Also, why did you add yourself to the contributor list? You know that list hasn't been edited in about 4 years. I don't think you can add yourself there without getting the main approval from @kmike, @pablohoffman or @redapple . However, I do believe that list should be updated to reflect everyone who has contributed.

Because I'm contributing? I just read and followed the contributing guidelines as defined [here](http://scrapy.readthedocs.io/en/latest/contributing.html)... Project owners are free to ask otherwise.

tests/test_utils_signal.py
```diff
@@ -77,3 +78,42 @@ def test_error_logged_if_deferred_not_supported(self):
...
+        handlers_called = set()
+
+        dispatcher.connect(self.error_handler, signal=test_error_signal)
+        dispatcher.connect(self.ok_handler, signal=test_ok_signal)
+        result = yield defer.maybeDeferred(
+            self._get_result, test_ok_signal, arg='test',
+            handlers_called=handlers_called
+        )
+
+        with pytest.raises(defer.FirstError):
```
isra17: I tried first with `self.assertRaises` but somehow the `py27:jessies` tests was failing with `TypeError: assertRaises() takes at least 3 arguments (2 given)` which is really weird since python2.7 should support assertRaises with `contextmanager` and is used in this form in other scrapy tests.

Maybe the `yield` in the context does trigger some fixed bugs with `assertRaises`.

IAlwaysBeCoding:
Your implementation is the one swallowing errors, not the current Scrapy implementation.

First of all, you aren't catching at all the `request_downloading` signal, you left that out completely. 

Second of all, the current implementation adds the failure and logs it, then continues with the next set of signal. That is the correct way to do it. Your way basically consumes the errors and fire backs on the first error encountered. You could have just added an extra keyword to fireback on first error encountered, and it would have been better. 

Heck that could have been an extra Scrapy Setting added as a flag. You don't need to create `send_deferred` , so I don't know why the hell is in there.

Anyways I like everything else, and I think the signal `request_downloading` is highly needed and you did a great job at implementing it on this pull request.

isra17:
My implementation work work this way:
```python
dfd = self.signals.send_deferred(...)
dfd.addCallback(_download_request)
dfd.addCallback(_downloaded)
return dfd.addBoth(finish_transferring)
```

My goal was to make it so that if one of the signal handler raise an error, `_download_request` and `_downloaded` are skipped and we go right to `finish_transferring`. By using `send_catch_log_deferred`, all handler are chained to an `addErrback` that doesn't reraise, which is what I mean by error being "swallowed". Therefor deferred object *always* callback, resulting with `_download_request` being followed, even though one of the signal handler might have raised.

My implementation of `send_deferred` make it so that if one of the signal handler raise, the returned deferred will not callback, but errback. In the chain above, this mean that we skip right to `finish_transferring` and avoid downloading the request.

I could have add a flag in `send_catch_log_deferred`, but this would have been at the price of mixing a bunch of `if` in it and making just as much as a callback hell than having a simple and readable function.

Users that want to ensure their handler are called every time should use middleware (Such as logging), the `request_downloading` as implemented here is really a last chance to cancel a request or mutate some property, but once one handler cancel the request, there's no point to wait for the other handlers.

The costly feature here is the one cancelling requests, otherwise a simple `dfd = self.signals.send_catch_log_deferred(...)` would have worked just fine.

Hope this makes it clearer!

IAlwaysBeCoding:
Hey @kmike after thinking about it hard enough. I think this would be a nice addition to Scrapy, but I want to know what you think about the function `send_deferred`. However, we do need a `response_downloading` built-in signal in my honest opinion.

PR #2972, opened 2017-10-21
Add a note to allowed_domains

aitoehigie:


cathalgarvey:
Hi @aitoehigie - I think the docs as written are already clear on this point. If you feel they are unclear, is there a change in wording to the _existing_ text, that would make it clearer?

wRAR (changes requested):
As written above, this shouldn't just reiterate what is already mentioned.

Gallaecio:
I’m closing given the lack of action from @aitoehigie since @cathalgarvey’s feedback.

PR #2982, opened 2017-10-28
Fix broken links and use https links wherever possible

codeaditya:
I haven't touched scrapy shell commands and other places which have the potential to break something else if not careful. Besides the changes made, I would like to point out two more issues:

1. https://github.com/scrapy/scrapy/blob/master/scrapy/http/cookies.py#L115 - ``WrappedRequest`` class mentions and links to ``urllib2.Request`` (the link is broken). Since in Python 3 it is ``urllib.request.Request``, I am not sure about the best way to fix this.

2. Shall we switch the default user agent to https as well?

### DIFF
--- a/scrapy/extensions/httpcache.py
+++ b/scrapy/extensions/httpcache.py
@@ -70,8 +70,8 @@ class RFC2616Policy(object):
         return True
 
     def should_cache_response(self, response, request):
-        # What is cacheable - http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec14.9.1
-        # Response cacheability - http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.4
+        # What is cacheable - https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec14.9.1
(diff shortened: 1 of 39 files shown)


scrapy/extensions/httpcache.py
```diff
@@ -163,7 +163,7 @@ def _get_max_age(self, cc):
 
     def _compute_freshness_lifetime(self, response, request, now):
         # Reference nsHttpResponseHead::ComputeFreshnessLifetime
-        # http://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/http/nsHttpResponseHead.cpp#410
+        # https://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/http/nsHttpResponseHead.cpp#706
```
kmike: 👍

kmike:
Thanks @codeaditya for the careful pull request!
Yes, I think that'd be good to change User-Agent as well.

codeaditya:
Awesome. Thanks @kmike.. It's been a very pleasurable experience.. Would send a PR for User-Agent too..

PR #2983, opened 2017-10-28
[MRG+1] Use https link in default user agent

codeaditya:


kmike:
Looks good to me, thanks @codeaditya! 
This change is technically backwards incompatible, we must be sure to mention it in the changelog.

codeaditya:
Yeah @kmike.. [docs/news.rst](https://github.com/scrapy/scrapy/blob/master/docs/news.rst) seems to only document changes for released versions. Where do we document this change in development version?

Issue #2701, opened 2017-04-05
Make Exceptions Handled by Retry Middleware Configurable


tad3j:
I was having troubles finding which exceptions are handled by Retry Middleware, so I asked on freenode scrapy irc channel (should obviously check the source code). Paul from scrapinhub helped me there and pointed me to the source code. 

Beside that he also asked me to open an issue here.
He gave a great suggestion to make exceptions handled by RetryMiddleware configurable, so if we don't like defaults we can configure our own.

Beside that he said that the handled exceptions in question are not documented well; I'm just mentioning it here, didn't want to open another issue (I guess by adding them to settings will also require some documentation).

sanjucta:
Hi, 
Since this is tagged as an as 'help-wanted' and 'enhancement', I think the help wanted here is to make the exceptions handled configurable. I'd like to work on this. 
I think the way the exceptions should be made configurable is to allow the user to exclude certain exceptions rather than add exceptions, since  we only want to re-try in the case of exceptions in which the issue could have been resolved when we retry . Does that sound reasonable?

doprdele:
^^ would also like to see this. how exactly should we retry on exceptions now?

PR #2984, opened 2017-10-30
Added: Making the list of exceptions to retry configurable via settings

starrify:
Implements #2701

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -13,28 +13,16 @@ there is no more failed pages to retry this middleware sends a signal
 """
 import logging
 
-from twisted.internet import defer
-from twisted.internet.error import TimeoutError, DNSLookupError, \
-        ConnectionRefusedError, ConnectionDone, ConnectError, \
-        ConnectionLost, TCPTimedOutError
-from twisted.web.client import ResponseFailed
-
+import six
 from scrapy.exceptions import NotConfigured
 from scrapy.utils.response import response_status_message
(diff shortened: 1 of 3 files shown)


scrapy/downloadermiddlewares/retry.py
```diff
@@ -13,35 +13,27 @@
...
+from scrapy.utils.misc import load_object
 
 logger = logging.getLogger(__name__)
 
 
 class RetryMiddleware(object):
 
-    # IOError is raised by the HttpCompression middleware when trying to
-    # decompress an empty response
-    EXCEPTIONS_TO_RETRY = (defer.TimeoutError, TimeoutError, DNSLookupError,
```
kmike: Previously the way to add a new exception to retry was to subclass RetryMiddleware and add something to EXCEPTIONS_TO_RETRY. It means that the change is backwards incompatible. We should make sure that code like this still works, but shows a warning:

```py
from scrapy.downloadermiddlewares.retry import RetryMiddleware
from scrapy.core.downloader.handlers.http11 import TunnelError

class MyRetryMiddleware(RetryMiddleware):
    EXCEPTIONS_TO_RETRY = RetryMiddleware.EXCEPTIONS_TO_RETRY + (TunnelError,)
```

 I think we can make it backwards compatible this way:

1. make EXCEPTIONS_TO_RETRY a property; raise a warning if it is accessed. This property should return the same exception list. You can load it from default_settings if no settings are available.

2. If EXCEPTIONS_TO_RETRY attribute is overridden, middleware should use this attribute instead of exceptions specified in settings.py.

PR #2985, opened 2017-10-30
utils.curl: add parse_curl_cmd func

immerrr:
Given that the major browsers are able to export the requests in cURL format, it's logical that we have a utility in scrapy to make it a request.

This PR works towards adding a function that creates kwargs that could be passed to a `Request` constructor.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,4 +3,5 @@ usefixtures = chdir
 python_files=test_*.py __init__.py
 python_classes=
 addopts = --doctest-modules --assert=plain
+testpaths = scrapy tests
 twisted = 1
--- a/scrapy/utils/curl.py
+++ b/scrapy/utils/curl.py
@@ -1,0 +1,51 @@
+import argparse
+from shlex import split
+from six.moves.http_cookies import SimpleCookie
+from six import string_types
+
+
+class CurlParser(argparse.ArgumentParser):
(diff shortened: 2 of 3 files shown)


scrapy/utils/misc.py
```diff
@@ -117,3 +117,38 @@ def md5sum(file):
 def rel_has_nofollow(rel):
     """Return True if link rel attribute has nofollow type"""
     return True if rel is not None and 'nofollow' in rel.split() else False
+
+
+
+import argparse
```
kmike: I think you can create scrapy.utils.curl

kmike:
I like this idea.

kmike:
PyPy test failures should be resolved in master, but Python 3 test failures look like problems with the code.

immerrr:
@kmike fixed

kmike:
It needs some docs before a merge.

Also, what do you think about providing `Request.from_curl` alternative constructor, to make this feature discoverable?

With `parse_curl_cmd` name it is not immediately obvious the result can be used as Request kwargs, so if there is no `Request.from_curl` constructor, it could make sense to rename this function to `curl_to_request_kwargs` or something similar; if there is a public Request.from_curl constructor then existing name looks ok (though a docstring can still help).

Gallaecio:
Closing in favor of #3862

Issue #1878, opened 2016-03-24
Expose cookiejars


kmike:
Scrapy cookiejar API is limited:

* meta key is called `cookiejar`, but you can't put CookieJar object there, in fact it means `cookiejar_id` or `session_id`, not `cookiejar`; this is confusing. It should have been called `session_id` IMHO.
* there is no way to get or set current cookies; it is a popular issue we don't have a solution for (see  http://stackoverflow.com/questions/8708346/access-session-cookie-in-scrapy-spiders and https://github.com/scrapy/scrapy/issues/1448).

I think we should provide a better API for 'sessions'. It should allow to 

1. access current session cookies;
2. 'fork' a session - start separate sessions from the current session.

Currently I'm using an ugly hack to access cookies:

```py
class ExposeCookiesMiddleware(CookiesMiddleware):
    """
    This middleware appends CookieJar with current cookies to response flags.

    To use it, disable default CookiesMiddleware and enable
    this middleware instead::

        DOWNLOADER_MIDDLEWARES = {
            'scrapy.downloadermiddlewares.cookies.CookiesMiddleware': None,
            'autologin.middleware.ExposeCookiesMiddleware': 700,
        }

    """
    def process_response(self, request, response, spider):
        response = super(ExposeCookiesMiddleware, self).process_response(
            request, response, spider)
        cookiejarkey = request.meta.get("cookiejar")
        response.flags.append(self.jars[cookiejarkey])
        return response


def get_cookiejar(response):
    for obj in response.flags:
        if isinstance(obj, CookieJar):
            return obj
```

I don't have a concrete API proposal, but likely it should use word 'session' :)

pawelmhm:
getting and setting cookies in Scrapy is really huge pain so big :+1: from me.

In project I'm working on now we use following solution that sets "jars" from Cookie Middleware on spider, and then allows you to use it.

```python

class CustomCookiesMiddleware(cookies.CookiesMiddleware):
    @classmethod
    def from_crawler(cls, crawler):
        o = super(CustomCookiesMiddleware, cls).from_crawler(crawler)
        crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
        return o

    def spider_opened(self, spider):
        self.enabled = getattr(spider, 'cookies_enabled', self.enabled)
        spider._cookiejars = self.jars
```

```python
def BaseSpider(Spider):
    def get_cookie(self, name, cookiejar=None):
           if cookiejar not in self._cookiejars:
[...]

PR #2986, opened 2017-10-30
CookiesMiddleware: add "reset_cookies" meta to clear the jar

immerrr:
This PR adds a `reset_cookies` meta to clean the active cookiejar.

When I try working with sessions I often find myself in a situation when I'd like to restart the session "from scratch," including cookies. It's doable by just setting `cookiejar` meta to an arbitrary value, but then we'd be accumulating the cookiejars over time.

I'm not entirely sure about the meta name, as I'd probably like it namespaced, e.g. `cookies_reset`, but there's a precedent already with `dont_merge_cookies`, so I chose to follow that pattern.

@kmike am I missing something here?


kmike:
I'd love to see cookiejar exposed to user directly instead of adding these "commands" in request.meta (thought on https://github.com/scrapy/scrapy/issues/1878 are welcome!). What's the use case? Note that with reset_cookies you're restarting from scratch at arbitrary point of time, when Request is started to download, not when you send a Request. Is it what you really want, or do you really want to clear cookiejar when sending a request?

If we're going down this request.meta route, the implementation looks fine. As for the name, what about cookiejar_reset, to use some namespace? We have many names prefixed with `dont_`, but other prefixes ar not common.

immerrr:
> What's the use case?

I think I've mentioned it in the OP, I'm doing sessions on top of Crawlera and want to clean up "banned" cookiejars instead of generating new ones and accumulating the old ones over time.

> Request is started to download, not when you send a Request. Is it what you really want, or do you really want to clear cookiejar when sending a request?

I don't care much about it, as long as the outgoing request is clean it's fine.

immerrr:
@kmike anything else to fix here? (aside from adding tests)

Issue #900, opened 2014-09-23
Centralized Request fingerprints


kmike:
It is very easy to have a subtle bug when using a custom duplicates filter that changes how request fingerprint is calculated.
- Duplicate filter checks request fingerprint and makes Scheduler drop the request if it is a duplicate.
- Cache storage checks request fingerprint and fetches response from cache if it is a duplicate.
- If fingerprint algorithms differ we're in trouble.

The problem is that there is no way to override request fingerprint globally; to make Scrapy always take something extra in account (an http header, a meta option) user must override duplicates filter and all cache storages that are in use.

Ideas about how to fix it:
1. Use duplicates filter `request_fingerprint` method in cache storage if this method is available;
2. create a special Request.meta key that `request_fingerprint` function will take into account;
3. create a special Request.meta key that will allow to provide a pre-calculated fingerprint;
4. add a settings.py option to override request fingerprint function globally.

michael-yin:
I think we should make the fingerprint calculation in a isolated download middleware which focus on fingerprint, after the fingerprint has been calculated by the middleware, save the value in response.meta. If other middleware need to know the fingerprint of the url, just find it in response.meta. 

I think this process have many advantages:

* Fingerprint of the url will be global, if dev want to change the way it calculate, they can write custom middleware and activate it in settings.py
* Fingerprint will be calculated only once, which is efficient. 
* In some cases some middleware may change the url, and the fingerprint might need to be calculated before or after the change,  the priority number in the DOWNLOADER_MIDDLEWARES can give developer more control of the fingerprint calculation.

We can do it step by step:

1. We need to write a built-in middleware (we can call it Fingerprintmiddleware) which use the default 'request_fingerprint' to calculate the fingerprint and save the value in response.meta, and the built-in duplicate filter and cache storage use the value in response.meta. 

2. I think DUPEFILTER_CLASS might not be useful because now there is middleware take care of the fingerprint of url. Actually I think duplicate filter now should concert about if the url has been crawled, not the fingerprint calculation.

We can add FINGERPRINT_CLASS in settings.py, it might look like this.

```py
FINGERPRINT_CLASS = {
   'Custome_class_fingerprint' : 200
} 
```

If FINGERPRINT_CLASS is in settings.py, we will disable the built-in Fingerprintmiddleware, the number in the config is the priority number which control the order.

hecris:
[...]

PR #2988, opened 2017-10-31
Added: Customizing request fingerprint calculation via request meta

starrify:
Partially implements #900


kmike:
Thanks @starrify for getting the ball rolling! I'm -1 for merging a partial fix though; this one is too partial :) For example, it won't work with redirects: meta is copied, and so fingerprint is copied, and so redirected request has the same fingerprint as original request => it is always filtered out.

Issue #2932, opened 2017-09-18
Structure of XmlExportPipeline example class in docs is confusing


colinmorris:
The [Exporters section of the docs](https://doc.scrapy.org/en/latest/topics/exporters.html#using-item-exporters) give an example pipeline class called `XmlExportPipeline`. It defines methods `spider_opened` and `spider_closed`.

I found this confusing because the [pipeline docs](https://doc.scrapy.org/en/latest/topics/item-pipeline.html#writing-your-own-item-pipeline) describe methods called `open_spider` and `close_spider`, with seemingly the same semantics. At first I thought the example methods were typo'd, until I saw that they were explicitly connected to the corresponding signals in `from_crawler`.

I'm guessing this is the "old way" of doing it, and the example predates the existence of the `open_spider` and `close_spider` methods?

Unless there's a reason to do it this way, I think the example would be clearer (and shorter) if it used the built-in names.

kmike:
A good catch. There is no need to connect to signals manually now. `self.files` is also not required, as there is a single spider per crawler now. A pull request is welcome :)

kmike:
Fixed by https://github.com/scrapy/scrapy/pull/2989. Thanks @colinmorris!

PR #2989, opened 2017-10-31
[MRG+1] Revise/modernize item exporter example in docs

colinmorris:
Addresses #2932

@kmike pointed out that the concept of a file-per-spider no longer made sense, so I changed the use case to one inspired by my own experience. It's maybe a little more complicated than the previous example, but it still comes out to fewer lines of code after removing the overhead of manually connecting the signals.

### DIFF
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -49,14 +49,14 @@ value of one of their fields::
             self.year_to_exporter = {}
 
         def close_spider(self, spider):
-            for exporter in self.year_to_exporter.itervalues():
+            for exporter in self.year_to_exporter.values():
                 exporter.finish_exporting()
                 exporter.file.close()
 
         def _exporter_for_item(self, item):
             year = item['year']
             if year not in self.year_to_exporter:
-                f = open('{}.xml'.format(year), 'w+b')
+                f = open('{}.xml'.format(year), 'wb')
                 exporter = XmlItemExporter(f)
                 exporter.start_exporting()
                 self.year_to_exporter[year] = exporter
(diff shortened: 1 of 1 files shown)


docs/topics/exporters.rst
```diff
@@ -36,38 +36,36 @@ to export
...
+    from scrapy.exporters import XmlItemExporter
+
+    class PerYearXmlExportPipeline(object):
+        """Distribute items across multiple XML files according to their 'year' field"""
+
+        def open_spider(self, spider):
+            self.year_to_exporter = {}
+
+        def close_spider(self, spider):
+            for exporter in self.year_to_exporter.itervalues():
```
kmike: It should be .values() in order to be compatible with Python 3. Creating a temporary list in Python 2 is not a problem as a list would be small.

docs/topics/exporters.rst
```diff
@@ -36,38 +36,36 @@ to export
...
+                exporter.finish_exporting()
+                exporter.file.close()
+
+        def _exporter_for_item(self, item):
+            year = item['year']
+            if year not in self.year_to_exporter:
+                f = open('{}.xml'.format(year), 'w+b')
```
kmike: why is 'w+b' needed, not just 'wb'?

kmike:
I like the revised example, exporting to several locations is a common problem users ask for 👍 
The PR looks good to me, besides a couple of minor comments.

docs/topics/exporters.rst
colinmorris: Good call. I just copied it from the previous example, but as you say, it doesn't seem to be necessary.

kmike:
Looks good to me, thanks @colinmorris!

PR #2991, opened 2017-11-01
Fix PyPy build

### LINKED ISSUE #2990: pypy tests fail on CI
It seems pyenv changed something again, so pypy installation fails; all recent PRs are red. //cc @lopuhin :)

lopuhin:
Fixes #2990

We'll see if it fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,16 +24,10 @@ matrix:
 install:
   - |
       if [ "$TOXENV" = "pypy" ]; then
-        export PYENV_ROOT="$HOME/.pyenv"
-        if [ -f "$PYENV_ROOT/bin/pyenv" ]; then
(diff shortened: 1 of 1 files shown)


lopuhin:
This is ready - now instead of using pyenv and relying on regexp we download portable pypy directly from https://github.com/squeaky-pl/portable-pypy - this repo is a known and a long-time stable source of portable pypy builds. The reason why we need portable is that we need openssl to work correctly, but this is not achieved on Travis with default pypy build. Portable builds bundle the most recent stable OpenSSL version https://github.com/squeaky-pl/portable-pypy#a-word-about-openssl so I think we should be fine.

The only drawback is that we won't be automatically using the latest PyPy, but at least the build should become more stable.

kmike:
Thanks @lopuhin!

PR #2995, opened 2017-11-08
Allow passing Failure object to middlewares

isra17:
Trying to handle errors coming from middleware, it happens that Twisted strip traceback from an exception returned from a completed deferred (Needed to avoid GC issues). This mean that trying to use `exception.__traceback__` always yield `None`.

This PR adds a decorator that can be used on middleware `process_exception` or `process_spider_exception` to get a Failure object instead of an Exception. The Failure object provides a `getTracebackObject` to get a `Traceback`-like object which come really handy when trying to pinpoint an issue.

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -148,6 +148,11 @@ more of the following methods:
       :meth:`process_exception` methods of the middleware the same as returning a
       response would.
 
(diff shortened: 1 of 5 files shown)


wRAR:
"twisted.internet.defer.Deferred no longer removes the traceback object from Failures."

PR #2996, opened 2017-11-08
Add signals to handle error from downloader or pipeline

leameow:
As of right now, there's is no simple way to handle exception coming from items pipeline or a the `process_response` of a downloader middleware.

This PR adds two signals for those use case. Note that the downloader signals also catch any exception from the downloader engine, including `process_request` and `process_exception`.


AdrianAtZyte:
`item_error` has been created already, but I have my doubts about implementing `download_error`, so I think it may be best to close.

PR #3011, opened 2017-11-18
[MRG+1] Issues a warning when user puts a URL into allowed_domains (#2250)

Jane222:


Jesse-Bakker:
Please also write a test case for this

Jane222:
@Jesse-Bakker thanks for your quick feedback, I fixed the looping over elements issue.

How exactly should a test case for this look like? All I did was insert a warning (because URL-formatted items in `allowed_domains` were being ignored, already), so should I just add a test class in `tests/test_spidermiddleware_offsite.py` that puts a URL in the spider's `allowed_domains` and then checks whether the warning was triggered by doing something like

`self.assertTrue(logger.warning.called)`?

Also, any pointers on how to get hold of the correct `logger` while inside `tests/test_spidermiddleware_offsite.py`? Would this be correct?

```
import logging
```

Thanks for your help!

Jesse-Bakker:
I recommend emitting a python warning and testing it as follows:
https://docs.python.org/3/library/warnings.html#testing-warnings

Jane222:
@Jesse-Bakker Thanks for your suggestion, but again, I am correct in assuming that a test class for this should be added in `tests/test_spidermiddleware_offsite.py` and this test class should put a URL in the spider's allowed_domains and then check whether the warning was triggered or not (using catch_warnings)?

Jane222:
@Jesse-Bakker figured it out, I changed `logger.warn()` to `warnings.warn()` in `offsite.py` and added a test class in `test_spidermiddleware_offsite.py` that checks whether that warning was triggered when there was a URL present in `allowed_domains.`

Jesse-Bakker:
@Jane222 looks pretty good. I do suggest subclassing the default Warning class and testing for the subclass instead of the message. That way is more robust. After that, I'm +1 on merging.

Jane222:
@Jesse-Bakker done, there is now a URLWarning class that I check for.

Jesse-Bakker:
Great! Now you just have to get it merged by someone with commit rights!

Jane222:
I see. Are there any rules about whom to contact? For example, this issue was opened by @redapple and seems like he's made plenty of commits - should I e-mail him privately (seems intrusive)  or wait for him to respond to this pull request on his own or what is the protocol here?

redapple:
Hey @Jane222 , I won't be available much for scrapy and reviews in the near future, so your best bet is to ping @kmike or @dangra . But do know that there's no guarentee as their time is really scarce.
And yes, please keep all communication in the open (here on GitHub in this PR) and refrain from personal emails to mainteners.

Jane222:
@kmike ,  @dangra could one of you take a look at this pull request and merge? Sorry to bother you, I know you're busy!

scrapy/spidermiddlewares/offsite.py
```diff
@@ -52,9 +53,18 @@ def get_host_regex(self, spider):
         allowed_domains = getattr(spider, 'allowed_domains', None)
         if not allowed_domains:
             return re.compile('') # allow all by default
+        for domain in allowed_domains:
+            url_pattern = re.compile("^https?://.*$")
```
dangra: No need to compile the regex on each iteration, I'd compile it as class attribute, or in `__init__` or at least before line 56.

scrapy/spidermiddlewares/offsite.py
```diff
@@ -52,9 +53,18 @@ def get_host_regex(self, spider):
...
+            if url_pattern.match(domain):
+                warnings.warn("allowed_domains accepts only domains, not URLs. Ignoring URL entry %s in allowed_domains." % domain, URLWarning)
```
dangra: What about going the other way around, checking domain is a valid hostname string so it discards (and warns) on anything not matching.

scrapy/spidermiddlewares/offsite.py
Jane222: @dangra done, regex is now being compiled before the loop

scrapy/spidermiddlewares/offsite.py
Jane222: @dangra In this pull request, I wanted to keep it as simple as possible...

There was some discussion in the issue thread about whether removing invalid domain names should be done in addition to issuing a warning but the consensus seemed to have been that a new issue should be opened for that.

The issue I was trying to solve here was specifically about URLs vs. domain names, which was easy to do... But to implement your request to check for a valid domain name, we'd first need to have a discussion about how that regex should look like, how strict we should be...

I considered removing elements that matched the `url_pattern` regex from allowed_domains` but then it turned out that they are not being crawled anyway... and again, for actually removing instead of issuing a warning, there was the suggestion to open a separate issue.

dangra: Fair point. thanks

lopuhin:
Thanks @Jane222 !

PR #3012, opened 2017-11-19
Fix Deprecated SafeConfigParser -> ConfigParser

Jesse-Bakker:
When running with a low Log Level on python >= 3.2, you get a DeprecationWarning, saying SafeConfigParser was renamed to ConfigParser in python 3.2. Since Scrapy only supports python 2.7 and python 3.3 and above, this is a simple fix preventing an annoying warning.

### DIFF
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -4,7 +4,7 @@ import numbers
 from operator import itemgetter
 
 import six
-from six.moves.configparser import SafeConfigParser
+from six.moves.configparser import ConfigParser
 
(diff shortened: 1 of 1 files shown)


kmike:
Won't it make Python 2.7 use ConfigParser instead of SafeConfigParser? They are different, as per https://docs.python.org/2/library/configparser.html#ConfigParser.ConfigParser.

Jesse-Bakker:
I just checked and it will indeed. Maybe it would be better to fix this in six to let SafeConfigParser point to ConfigParser in python 3

PR #3016, opened 2017-11-22
Added Scrapy logo

Letharrick:
I replaced the "Scrapy" text with the logo shown on the Scrapy website

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,5 @@
-======
-Scrapy
-======
+.. image:: http://www.preserveware.com/wp-content/uploads/2017/06/scrapylogo1.png
+    :target: https://scrapy.org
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1,6 +1,5 @@
-======
-Scrapy
-======
+.. image:: http://www.preserveware.com/wp-content/uploads/2017/06/scrapylogo1.png
+    :target: https://scrapy.org
```
kmike: The logo in README looks fine to me, but I'm not comfortable serving an image from a third-party location we don't have control over.

Letharrick: Good point.
I've changed it  so that it will use the [logo that is in the repo](https://github.com/scrapy/scrapy/blob/master/artwork/scrapy-logo.jpg)

PR #3023, opened 2017-11-29
Add settings to the request dont_filter = True

zhaojiedi1992:


IAlwaysBeCoding:
I fail to see the focal point of this commit. How does adding don_filter=True to the requests inside the tutorial correlate to what you are describing on the commit message? Also,  why?

PR #3038, opened 2017-12-12
DOC update "Contributing" docs

kmike:
* suggest Stack Overflow for Scrapy usage questions;
* encourage users to submit test-only pull requests with reproducable examples;
* encourage users to pick up stalled pull requests;
* we don't use AUTHORS file as a main acknowledgement source;
* suggest using Sphinx autodocs extension

What do you think @dangra @eliasdorneles @cathalgarvey @redapple?

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -127,8 +127,15 @@ original pull request author haven't had time to address them.
 In this case consider picking up this pull request: open
 a new pull request with all commits from the original pull request, as well as
 additional changes to address the raised issues. Doing so helps a lot; it is
-not considered rude as soon as original the author is acknowledged by keeping
+not considered rude as soon as the original author is acknowledged by keeping
 his/her commits.
+
+You can pull an existing pull request to a local branch
+by running ``git fetch upstream pull/$PR_NUMBER/head:$BRANCH_NAME_TO_CREATE``
+(replace 'upstream' with a remote name for scrapy repository,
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -100,11 +119,22 @@ starting point is to send a pull request on GitHub. It can be simple enough to
...
+
+Sometimes there is an existing pull request for the problem you'd like to
+solve, which is stalled for some reason. Often the pull request is in a
+right direction, but changes are requested by Scrapy maintainers, and the
+original pull request author haven't had time to address them.
+In this case consider picking up this pull request: open
+a new pull request with all commits from the original pull request, as well as
+additional changes to address the raised issues. Doing so helps a lot; it is
+not considered rude as soon as original the author is acknowledged by keeping
+his/her commits.
```
eliasdorneles: +1, it is great to point that out explicitly!

It could be nice to add also that one can checkout a PR into a local branch by running:

    git fetch upstream pull/$PR_NUMBER/head:$BRANCH_NAME_TO_CREATE

kmike: Awesome, I didn't know about this option!

eliasdorneles: Yeah, it's pretty convenient, though a bit annoying to type all that each time. :)

I'm lazy so I setup [some aliases](https://github.com/eliasdorneles/dotfiles/blob/master/gitconfig#L81-L91) in order to just do `git pr $PR_NUMBER`

PR #1674, opened 2016-01-13
[MRG+1] Add Python 3.5 tox env + Python 3.5 tests in Travis

redapple:

PR #3045, opened 2017-12-19
[MRG+1] Drop support for EOL Python 3.3

### LINKED ISSUE #3035: Python3.3 support and requirements without it
Scrapy still supports py3.3 (at least according to its trove classifiers in setup.py and the CI conf)
but some of its dependencies dropped support some time ago.
https://github.com/pyca/service_identity/blob/master/CHANGELOG.rst#backward-incompatible-changes-1
https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#20---2017-07-17

This caused some problems when testing scrapy daemon with py3.3,
which was resolved by installing the [enum-compat virtual package](https://pypi.python.org/pypi/enum-compat/0.0.2)

There are several options here.
scrapy1.5 can drop support for python3.3,
scrapy1.4 can restrict the max versions for those dependencies
and enum-compat can become a requirement,
although there may be more things broken.

I didn't figure out why the python3.3 build for scrapy doesn't fail
but here is a failed scrapyd build https://travis-ci.org/scrapy/scrapyd/jobs/299029712

hugovk:
Fixes #3035.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ matrix:
       env: TOXENV=jessie
     - python: 2.7
       env: TOXENV=pypy
+    - python: 3.4
+      env: TOXENV=py34
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ matrix:
       env: TOXENV=jessie
     - python: 2.7
       env: TOXENV=pypy
-    - python: 3.3
-      env: TOXENV=py33
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -164,7 +164,7 @@ you can install Scrapy with ``pip`` after that::
 
 .. note::
     The same non-Python dependencies can be used to install Scrapy in Debian
-    Jessue (8.0) and above.
+    Jessie (8.0) and above.
 
 
 .. _intro-install-macos:
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -7,7 +7,7 @@ Installation guide
 Installing Scrapy
 =================
 
-Scrapy runs on Python 2.7 and Python 3.3 or above.
+Scrapy runs on Python 2.7 and Python 3.4 or above.
 
 If you're using `Anaconda`_ or `Miniconda`_, you can install the package from
 the `conda-forge`_ channel, which has up-to-date packages for Linux, Windows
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -13,8 +13,6 @@ matrix:
       env: TOXENV=jessie
     - python: 2.7
       env: TOXENV=pypy
-    - python: 3.3
-      env: TOXENV=py33
```
dangra: not tested under python3.4?

hugovk: I've added it in a new commit.

In https://github.com/scrapy/scrapy/pull/1674#issuecomment-171288714 they decided not to test on 3.4, let me know if you prefer it left out.

dangra: by that time 3.5 was the latest release on 3.x branch and 3.3 was the smaller version we supported by then.

Looks like the unwritten policy was to test for the minimum version and the latest, what was 3.3. and 3.5 then now it is 3.4 and 3.6. 

Anyway, I'd left 3.5 until we realize it is taking too long to run travis-ci builds, or simply until 3.7 get released.

dangra:
LGTM.

docs/intro/install.rst
```diff
@@ -163,8 +163,8 @@ you can install Scrapy with ``pip`` after that::
     pip install scrapy
 
 .. note::
-    The same non-python dependencies can be used to install Scrapy in Debian
-    Wheezy (7.0) and above.
+    The same non-Python dependencies can be used to install Scrapy in Debian
+    Jessue (8.0) and above.
```
kmike: Jessia

hugovk: Oops, fixed!

setup.py
```diff
@@ -53,14 +53,14 @@ def has_environment_marker_platform_impl_support():
...
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Topic :: Internet :: WWW/HTTP',
         'Topic :: Software Development :: Libraries :: Application Frameworks',
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
```
kmike: Sorry, I've never used this option - is it silently ignored when old pip/setuptools version is used to install a package, without raising an error about unknown metadata or something like that?

hugovk: Yes, old pip < 9 ignores it when installing and old setuptools <  24.2.0 ignores it when creating distributions. 

See https://stackoverflow.com/a/42792413/724176

kmike: Thanks for clarification @hugovk!

Issue #3046, opened 2017-12-21
Item loader missing values from base item


stummjr:
ItemLoaders behave oddly when they get a pre-populated item as an argument and `get_output_value()` gets called for one of the pre-populated fields before calling `load_item()`.

Check this out:

```python
>>> from scrapy.loader import ItemLoader
>>> item = {'url': 'http://example.com', 'summary': 'foo bar'}
>>> loader = ItemLoader(item)
>>> loader.load_item()
{'summary': 'foo bar', 'url': 'http://example.com'}

# so far, so good... what about now?
>>> item = {'url': 'http://example.com', 'summary': 'foo bar'}
>>> loader = ItemLoader(item)
>>> loader.get_output_value('url')
[]
>>> loader.load_item()
{'summary': 'foo bar', 'url': []}
```

There are **2** unexpected behaviors in this snippet (at least from my point of view):

**1)** `loader.get_output_value()` doesn't return the pre-populated values, even though they end up in the final item.

It seems to be like this on purpose, though. The `get_output_value()` method only queries the `_local_values` defaultdict ([here](https://github.com/scrapy/scrapy/blob/master/scrapy/loader/__init__.py#L125)).

**2)** once we call `loader.get_output_value('url')`, that field is not included in the `load_item()` result anymore.

This one doesn't look right, IMHO.

It happens because when we call `loader.get_output_value('url')` for the first time, such value is not available on `_local_values`, and so a new entry in the `_local_values` defaultdict will be created with an empty list on it ([here](https://github.com/scrapy/scrapy/blob/master/scrapy/loader/__init__.py#L121)). Then, when `loader.load_item()` gets called, [these lines](https://github.com/scrapy/scrapy/blob/master/scrapy/loader/__init__.py#L116-L117) overwrite the current value from the internal item because the value returned by `get_output_value()` is `[]` and not `None`.

Any thoughts on this?

cathalgarvey:
The first issue is a UX issue, but I agree it should be fixed. The second one, I think is more like a bug, as a side-effect of a getter method is wiping valid data in the output.

yashrsharma44:
Here even if we run the code like this, we get the result - 
```
>>> from scrapy.loader import ItemLoader
>>> item = {'url':'http://example.com','summary':'foo bar'}
>>> loader = ItemLoader(item)
>>> loader.load_item()
{'summary': 'foo bar', 'url': 'http://example.com'}
>>> loader.get_output_value('url')
[]
[...]

PR #3047, opened 2017-12-21
Avoid missing base item fields in item loaders

stummjr:
This is an attempt to fix the behavior described in #3046.

Instead of just checking if the value inside the loader is not None in order to decide if a field from the initial item should be overwritten or not, `load_item()` should also make sure that the value returned by `get_output_value()` is not an empty list.

That is because `self._local_values` , which stores the new values included via `add_*` or `replace_*` methods, is a[`defaultdict(list)`](https://github.com/scrapy/scrapy/blob/master/scrapy/loader/__init__.py#L37). Then, when we call `get_output_value()` for a field only available in the initial item, an empty list will be set for that field in `self._local_values` (because of [this](https://github.com/scrapy/scrapy/blob/master/scrapy/loader/__init__.py#L125)).

This way, we make sure we don't miss fields from the initial item, in case `get_output_value()` gets called for one of the pre-populated fields before `load_item()`, as described on #3046.

### DIFF
--- a/scrapy/loader/__init__.py
+++ b/scrapy/loader/__init__.py
@@ -113,7 +113,7 @@ class ItemLoader(object):
         item = self.item
         for field_name in tuple(self._values):
             value = self.get_output_value(field_name)
-            if value is not None:
+            if value is not None and value != []:
                 item[field_name] = value
 
         return item
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -99,6 +99,15 @@ class BasicItemLoaderTest(unittest.TestCase):
         il.replace_value('sku', [valid_fragment], re=sku_re)
         self.assertEqual(il.load_item()['sku'], u'1234')
 
+    def test_get_output_value_before_load_item(self):
+        item = TestItem(url='http://example.com', summary='foo bar')
+        il = ItemLoader(item)
+        il.get_output_value('url')
+        self.assertEqual(il.load_item(), {
+            'url': 'http://example.com',
+            'summary': 'foo bar',
+        })
+
     def test_self_referencing_loader(self):
         class MyLoader(ItemLoader):
             url_out = TakeFirst()


dangra:
@stummjr it makes sense and I don't consider the change of behavior a backward incompatibility, more a gotcha removal. thanks

@kmike any comment before merging and including in 1.5.0 release?

kmike:
Argh, item loaders! As I understand the code, this change affects not only get_output_value, but output processors as well: if an output processor returns an empty list, after the change load_item will be returning a default value instead of this empty list. 

No idea how large is the issue. Example use case, rather theoretical: MapCompose output processor which drops some values from the result; when all results are dropped, after this change a default value is returned in .load_item() instead of an empty list.

That said, the way lists play with ItemLoaders is weird anyways. For example:

```py
ld = ItemLoader({'colors': ['white', 'black']})
ld.replace_value('colors', ['red', 'yellow'])
ld.load_item()  
# {'colors': ['red', 'yellow']}

ld = ItemLoader({'colors': ['white', 'black']})
ld.replace_value('colors', [])
ld.load_item() 
# {'colors': ['white', 'black']}

ld = ItemLoader({'colors': ['white', 'black']})
ld.replace_value('colors', 'blue')
ld.load_item()  
# {'colors': ['blue']}
```

I can't find in ItemLoader docs that `loader.replace_value(name, None)` or `loader.replace_value(name, [])` or tuple() or set() doesn't replace the value, but sets it to default, unlike any other value, including empty dicts, empty strings, False, 0, etc.

So I'm not against merging this PR, as it fixes a real-world issue @stummjr had, and there is undocumented item loader behavior anyways. But at the same time, this PR seems to add more undocumented behavior to ItemLoaders.

dangra:
What if `load_items` only attempt to set field's value if a call to `_add_value` was made for that field. It means loader has to keep track of "modified" fields and it can't rely only on `_local_values` dict.

```python
>>> item = {'colors': ['white', 'black'], 'foo': 'bar'}
>>> ld = ItemLoader(item)
>>> ld.get_output_value('colors')
[]
>>> ld.load_item()
{'colors': ['white', 'black'], 'foo': 'bar'}
>>> ld.replace_value('colors', [])
>>> ld.get_output_value('colors')
[]
>>> ld.load_item()
{'colors': [], 'foo': 'bar'}
```

scrapy/loader/__init__.py
```diff
@@ -113,7 +113,7 @@ def load_item(self):
         item = self.item
         for field_name in tuple(self._values):
             value = self.get_output_value(field_name)
-            if value is not None:
+            if value is not None and value != []:
```
cathalgarvey: I can imagine cases where someone _expects_ a failed load to still populate an empty list, and this change might break things for them. Perhaps instead the code should check whether adding an empty list would clobber an existing field? Because, per your issue in #3046 I think that's the more bug-like behaviour?

Gallaecio:
Closing given https://github.com/scrapy/scrapy/issues/3046 has been fixed.

Issue #2213, opened 2016-09-01
PyPy support


lopuhin:
It would be nice to enable pypy tests on travis, mention pypy support in the docs and benchmark a couple of spiders.

What is currently broken in tests:
- [x] #2030 - there are failing signals tests under PyPy (old signals also didn't work under PyPy) - probably it's better to fix them after the merge, since the PR is already big.
- [x] There are some failures with encoding (different number of <?> inserted), filed https://bitbucket.org/pypy/pypy/issues/2389/different-behavior-of-bytesdecode-utf8
- [x] Failure in `ImagesPipelineTestCase.test_convert_image` looks like a Pillow bug under PyPy?
- [x] There are some failures in tests that rely on implementations differences or bugs that are absent on PyPy - here just the tests must be updated.

Other tasks:
- [x] Test and document installation under PyPy (there can be more issues than for CPython due to missing or outdated binary wheels, e.g. https://github.com/pyca/cryptography/issues/2138)

Status: 1 failure in https://github.com/rootAvish/scrapy/compare/signal-rewrite...lopuhin:pypy?expand=1

redapple:
@lopuhin , what do you think of adding PyPy build to Travis and adding it to allowed failures?

lopuhin:
@redapple it's possible to do it now, but I'd rather do it after merge of #2030, unless we want to fix pypy issues in it before the merge. I think turning it on after #2030 is less work overall, because although that PR brings some failing tests, it still reduces overall number of pypy failures, including many crawler and spider tests.

lopuhin:
@redapple I think I misunderstood you - "adding it to allowed failures" means we'll know the status of pypy tests, but the build will not fail if there are pypy failures? If that is correct then yes, I'm +1!

redapple:
@lopuhin , correct, this will allow running the tests, monitor progress with PyPy, while still having the greenlight from Travis (assuming other Py2.7..Py3.5 tests pass)

redapple:
Tentative: https://github.com/scrapy/scrapy/pull/2215

lopuhin:
I checked speed on one CPU bound crawler, it was about 2x (or a little more) faster.

There were also a couple of small issues not covered with tests:
- if one of the middlewares fails to import, the exception is not printed with ``scrapy crawl`` command
- links with non-ascii domains fail:
```
    for link in self.le.extract_links(response):
[...]

PR #3048, opened 2017-12-25
[MRG+1] Mention PyPy support, add PyPy to install docs

lopuhin:
As discussed in #2213, mention that we have PyPy support, and describe installation issues. I put PyPy section at the end, along with platform issues, although it's not strictly a platform.

~~I discovered one issue though that I'll try to resolve in another PR: right now pypy-specific dependency PyPyDispatcher is never picked up. A way to check and a workaround is also in the docs, but not everyone reads them so I'd like to try fixing this before release.~~

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -69,9 +69,11 @@ Here's an example spider using BeautifulSoup API, with ``lxml`` as the HTML pars
 What Python versions does Scrapy support?
 -----------------------------------------
 
-Scrapy is supported under Python 2.7 and Python 3.4+.
+Scrapy is supported under Python 2.7 and Python 3.4+
+under CPython (default Python implementation) and PyPy (only for Python 2.7).
 Python 2.6 support was dropped starting at Scrapy 0.20.
 Python 3 support was added in Scrapy 1.1.
+PyPy support was added in Scrapy 1.4, PyPy version tested is PyPy2-v5.9.0.
 
 .. note::
     For Python 3 support on Windows, it is recommended to use
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -7,7 +7,8 @@ Installation guide
 Installing Scrapy
 =================
 
-Scrapy runs on Python 2.7 and Python 3.4 or above.
+Scrapy runs on Python 2.7 and Python 3.4 or above
+under CPython (default Python implementation) and PyPy (only for Python 2.7).
 
 If you're using `Anaconda`_ or `Miniconda`_, you can install the package from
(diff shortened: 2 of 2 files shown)


docs/faq.rst
```diff
@@ -69,9 +69,11 @@ Here's an example spider using BeautifulSoup API, with ``lxml`` as the HTML pars
 What Python versions does Scrapy support?
 -----------------------------------------
 
-Scrapy is supported under Python 2.7 and Python 3.4+.
+Scrapy is supported under Python 2.7 and Python 3.4+
+under CPython (default Python implementation) and PyPy (only for Python 2.7).
```
kmike: Do you know what's the reason pypy3 is not supported? Is it just us not running tests for it, or are there things which work in pypy2, but not in pypy3?

lopuhin: Last time I checked, lxml did not install on PyPy 3. I filed an issue about it and it was quickly fixed, but then it stopped working again. I didn't check in a while though, let me check again.

lopuhin:
Turned out that scrapy 1.4.0 didn't have that pypy-specific dependency yet, and if I build a wheel from master and install it locally, then PyPyDispatcher is picked up correctly - hopefully this will be the same for the wheel for 1.5.0 (I checked that py27 build installs setuptools-38.2.5 which is more than enough for this feature to work).

docs/intro/install.rst
```diff
@@ -223,6 +224,28 @@ After any of these workarounds you should be able to install Scrapy::
...
+that is to ``brew install openssl`` and then export the flags that this command
+recommends (only needed when installing scrapy). Installing on Linux has no special
+issues besides installing build dependencies.
+Installing scrapy with PyPy on Windows is not tested.
+
+You can check that scrapy is installed correctly by running ``scrapy bench``.
+If this command gives errors such as
+``TypeError: ... got 2 unexpected keyword arguments``, this means
+that setuptools was unable to pick up one PyPy-specific dependency.
+To fix this issue, run ``pip install 'PyPyDispatcher>=2.1.0'``.
```
kmike: @lopuhin do you prefer to keep this note just in case, or to remove it?

lopuhin: I'd rather keep it just in case. In theory pypy comes with a recent enough setuptools, but I remember that someone still had this issue somehow (although that could be with 1.4.0 which I thought already had this feature). If we don't get any reports about this, we can remove it in the next release.

kmike:
Looks good! Let's tackle pypy3 separately.

PR #3049, opened 2017-12-25
[MRG+1] setup.py: mention that we support PyPy. See GH-2213.

kmike:
Mention in setup.py that we support CPython and PyPy.


lopuhin:
nice catch, looks good to me! 👍

PR #3050, opened 2017-12-25
Add PyPy3 support

lopuhin:
Just checking how they run on Travis. Turns out that lxml on PyPy3 is no longer a problem (``PyPy 5.9.0``, ``Cython==0.27.3`` and ``lxml==4.1.1``).

Related to #2213

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ matrix:
       env: TOXENV=jessie
     - python: 2.7
       env: TOXENV=pypy
+    - python: 2.7
+      env: TOXENV=pypy3
     - python: 3.4
       env: TOXENV=py34
     - python: 3.5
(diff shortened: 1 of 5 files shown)


lopuhin:
Totally unscientific speed comparison (on a slow Xeon, running ``scrapy bench -s CLOSESPIDER_TIMEOUT=30``):


  | rps
-- | --
CPython 2.7 | 65
PyPy2-5.9.0 | 112
CPython 3.6 | 29
PyPy3-5.9.0 | 47

Would be good to deploy https://github.com/scrapy/scrapy-bench, benchmarks there are more representative I think.

lopuhin:
This is almost ready, only waiting for the second pypy3 build to finish to check if it still takes 22 minutes 😱 

Right now only Linux binaries are available for PyPy3, so I tested only Linux installation.

lopuhin:
pypy3 took 23 minutes again, although the second longest py35 took 15 minutes. Locally, pypy3 took 16 minutes, and I don't have any ideas for quick speedups of the tests.

PR #3058, opened 2018-01-01
Fix link in news.rst

jesuslosada:
Fix link pointing to issue 2884.


kmike:
👍 Thanks @jesuslosada!

PR #3061, opened 2018-01-03
Make autothrottle slow down on HTTP 429 response

Matthijsy:
When a response has status 429 (Too Many Requests) it is ignored by the AutoThrottle (because the latency is low, so otherwise the spider will speed up). I think that the wanted behaviour is that the spider will slow down when a 429 is received, but currently the spider will stay at a constant scraping speed. 

This PR adds a new setting `AUTOTHROTTLE_429_DELAY` when it is set the download delay will be increased with this value every time a 429 is received. It still respects the 'MAX_DOWNLOAD_DELAY` value.

### DIFF
--- a/docs/topics/autothrottle.rst
+++ b/docs/topics/autothrottle.rst
@@ -59,8 +59,10 @@ AutoThrottle algorithm adjusts download delays based on the following rules:
    and ``N`` is :setting:`AUTOTHROTTLE_TARGET_CONCURRENCY`.
 3. download delay for next requests is set to the average of previous
    download delay and the target download delay;
-4. latencies of non-200 responses are not allowed to decrease the delay;
-5. download delay can't become less than :setting:`DOWNLOAD_DELAY` or greater
+4. When the ::setting::`AUTOTHROTTLE_429_DELAY` is set and a 429 is received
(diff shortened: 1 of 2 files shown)


kmike:
I like the idea of handling 429 status code in AutoThrottle 👍 
Implementation looks fine to me. Behavior won't be ideal, but better than now. According to my understanding, this is what would happen with the current implementation:

* spider receives 429 and slows down until there are no 429 errors;
* then spider starts to gradually speed up;
* after some time spider starts getting 429 again, and the process repeats.

So this approach would allow to get some pages if there are 429 errors, but spider won't settle on a speed suggested by a website ever, it will be getting these 429 errors periodically, likely every N requests. I wonder if there is a better logic, which would allow approaching and maintaining the rate limit. Time to do some maths? :)

Matthijsy:
As far as I know is most of the logic behind a 429 based on the amount of request within a specified time frame. I think it is quite hard to calculate those two variables. If someone knows a better implementation that would be great but I have no idea at the moment

Matthijsy:
I currently do not have enough time to fix this PR, and do not have an idea how to do a good implementation, so I will close this PR to keep the list clean

PR #3062, opened 2018-01-03
Use as a library: Reactor decorator code + MWE

BrandonSmithJ:
Started a discussion which has additional information here: https://www.reddit.com/r/scrapy/comments/7nwedz/contribution_use_as_a_library/

I'm not entirely sure where to put this or if it's even within the scope of what you want to incorporate into scrapy, but I thought I'd offer it as a potential solution to one of the problems I ran into while using scrapy.

### DIFF
--- a/extras/reactor_decorator.py
+++ b/extras/reactor_decorator.py
@@ -1,0 +1,123 @@
+from functools import wraps
+from twisted.internet import reactor, error
+import traceback
+import logging 
+
+
+''' 
+The reactor_process decorator is used to isolate the decorated 
+function in a separate process, so as to allow the twisted reactor
+to be started and stopped multiple times. 
+This normally isn't possible (and is known problem with using scrapy
+in a library capacity: https://stackoverflow.com/questions/35289054/scrapy-crawl-multiple-times-in-long-running-process )
+but by using the multiprocessing library and some hacky runtime code
+swapping, we can abstract the whole thing into a single decorator.
+
+--- The one remaining issue is that the decorated function needs to import 
+--- all necessary modules _within_ that function. It's an unfortunate side 
+--- effect of having _helper_process defined here, as it won't contain the 
+--- same context as the actual function being called. 
+--- There should be a way to fix this; hopefully there's a way to do it with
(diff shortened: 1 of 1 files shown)


kmike:
The intended way to solve this is to start reactor once, and then use CrawlerRunner - see https://doc.scrapy.org/en/latest/topics/practices.html?#running-multiple-spiders-in-the-same-process. 

Example from docs:

```py
from twisted.internet import reactor, defer
from scrapy.crawler import CrawlerRunner

class MySpider1(scrapy.Spider):
    # Your first spider definition
    ...

class MySpider2(scrapy.Spider):
    # Your second spider definition
    ...

configure_logging()
runner = CrawlerRunner()

@defer.inlineCallbacks
def crawl():
    yield runner.crawl(MySpider1)
    yield runner.crawl(MySpider2)
    reactor.stop()

crawl()
reactor.run() # the script will block here until the last crawl call is finished
```

^^ here you don't have to stop the reactor in `crawl()` method, you can wait for new crawl tasks as well. https://github.com/scrapinghub/scrapyrt and https://github.com/TeamHG-Memex/arachnado work this way. 

If it is hard to structure code of your application around Twisted reactor, then multiprocessing can indeed be a workaround. But it has its own issues. I wonder if we can just provide a good example in README which starts a CrawlerProcess in a separate process, in order to fix an issue with non-restarting reactor. There is also a library which allows to work around this Twisted inconvenience: https://github.com/itamarst/crochet, and a library to run Scrapy in a blocking way, based on that library: https://github.com/rmax/scrapydo.

BrandonSmithJ:
Indeed, that method of blocking with the reactor isn't feasible in a dynamic web setting. Perhaps it would be possible to connect to an isolated external process to manage crawlers like that, but it would require more set up than simply offering an interface to a restart-able reactor. 

I wasn't aware of crochet or scrapydo - thanks for the links. They seem to follow the same idea as the multiprocessing approach, but by monitoring external threads. In any case, I just wanted to offer a concise solution to the problem; feel free to use it if it's helpful, or discard if not.

PR #3096, opened 2018-01-27
[MRG+1] Fix OS signal names

jesuslosada:
Using Ctrl-C to stop Scrapy when running it on Python 2 leads to this message:
`Received SIGINT, shutting down gracefully. Send again to force`

However, on some Python 3 versions (at least Python 3.5 and 3.6) Scrapy does not correctly detect the SIGINT signal:
`Received SIG_SETMASK, shutting down gracefully. Send again to force` 

Some constants defined in the signal module need to be skipped in order to filter only the signals.
For reference:
https://github.com/python/cpython/blob/master/Lib/signal.py

### DIFF
--- a/scrapy/utils/ossignal.py
+++ b/scrapy/utils/ossignal.py
@@ -1,16 +1,17 @@
 
 from __future__ import absolute_import
+import signal
 
 from twisted.internet import reactor
 
-import signal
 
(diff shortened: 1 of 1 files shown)


kmike:
A good catch!

PR #3101, opened 2018-01-31
Add JsonStatsCollector class to log scraping run stats as json

TeodorIvanov:
I have added a JsonStatsCollector class which inherits MemoryStatsCollector and overrides the close_spider method to log stats as json.
This would make further parsing of the logs easier, i.e. when using the ELK stack and Filebeat.


kmike:
Logging stats as json is a good feature, but I'm not sure putting it to a stats collector is the best way to implement it. Stats collectors are responsible for keeping stats somewhere; I think that it is better to create Scrapy extension which does the logging.

PR #3105, opened 2018-02-01
for consistency purpose: change badge square shape to round

yuzhoujr:


README.rst
```diff
@@ -22,7 +22,7 @@ Scrapy
    :target: https://codecov.io/github/scrapy/scrapy?branch=master
    :alt: Coverage report
 
-.. image:: https://anaconda.org/conda-forge/scrapy/badges/version.svg
+.. image:: https://img.shields.io/badge/Anaconda%20Cloud-1.5.0-blue.svg
```
kmike: Hm, this would require updating Anaconda version manually, right, so there is a risk we display a wrong version?

Issue #3090, opened 2018-01-25
JsonItemExporter puts lone comma in the output if encoder fails


tlinhart:
If `JsonItemExporter` is unable to encode the item, it still writes a delimiter (comma) to the output file. Here is a sample spider:

```
# -*- coding: utf-8 -*-
import datetime
import scrapy

class DummySpider(scrapy.Spider):
    name = 'dummy'
    start_urls = ['http://example.org/']

    def parse(self, response):
        yield {'date': datetime.date(2018, 1, 1)}
        yield {'date': datetime.date(1234, 1, 1)}
        yield {'date': datetime.date(2019, 1, 1)})
```

Encoding the second items fails:

```
2018-01-25 09:05:57 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method ?.item_scraped of <scrapy.extensions.feedexport.FeedExporter object at 0x7fcbfbd81250>>
Traceback (most recent call last):
  File "/home/pasmen/SW/Python/data-sci-env/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/pasmen/SW/Python/data-sci-env/local/lib/python2.7/site-packages/pydispatch/robustapply.py", line 55, in robustApply
    return receiver(*arguments, **named)
  File "/home/pasmen/SW/Python/data-sci-env/local/lib/python2.7/site-packages/scrapy/extensions/feedexport.py", line 224, in item_scraped
    slot.exporter.export_item(item)
  File "/home/pasmen/SW/Python/data-sci-env/local/lib/python2.7/site-packages/scrapy/exporters.py", line 130, in export_item
    data = self.encoder.encode(itemdict)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/home/pasmen/SW/Python/data-sci-env/local/lib/python2.7/site-packages/scrapy/utils/serialize.py", line 22, in default
    return o.strftime(self.DATE_FORMAT)
ValueError: year=1234 is before 1900; the datetime strftime() methods require year >= 1900
```

The output looks like this:

```
[
{"date": "2018-01-01"},
,
{"date": "2019-01-01"}
]
```

[...]

PR #3111, opened 2018-02-07
Partial fix for #3090 - only addresses JSON feeds.

cathalgarvey:
Quick fix for #3090, but the general problem of failure-during-multistep-serialisation also probably affects our XML exporter. A more broad fix would be to buffer changes and then "commit" them once everything has succeeded, but that's a bigger change than I felt comfortable writing without some discussion. :)

### DIFF
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -129,9 +129,9 @@ class JsonItemExporter(BaseItemExporter):
 
     def export_item(self, item):
         itemdict = dict(self._get_serialized_fields(item))
-        data = self.encoder.encode(itemdict)
+        data = to_bytes(self.encoder.encode(itemdict), self.encoding)
         self._add_comma_after_first()
-        self.file.write(to_bytes(data, self.encoding))
+        self.file.write(data)
 
 
 class XmlItemExporter(BaseItemExporter):
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -112,6 +112,13 @@ class JsonItemExporter(BaseItemExporter):
         if self.indent is not None:
             self.file.write(b'\n')
 
+    def _add_comma_after_first(self):
+        if self.first_item:
+            self.first_item = False
+        else:
+            self.file.write(b',')
+            self._beautify_newline()
+
     def start_exporting(self):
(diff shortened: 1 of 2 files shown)


cathalgarvey:
Odd; Travis failed because the invalid datetime didn't consistently fail to JSON encode. I've changed to a complex number, let's see if that works better. :)

scrapy/exporters.py
```diff
@@ -121,13 +128,9 @@ def finish_exporting(self):
...
     def export_item(self, item):
-        if self.first_item:
-            self.first_item = False
-        else:
-            self.file.write(b',')
-            self._beautify_newline()
         itemdict = dict(self._get_serialized_fields(item))
         data = self.encoder.encode(itemdict)
+        self._add_comma_after_first()
         self.file.write(to_bytes(data, self.encoding))
```
kmike: If we want to be super-strict on it, to_bytes can also fail.

cathalgarvey: It was harder to write a test case for encoding that worked across versions. :) In the end the sanest thing was a new test class, sorry for the bloat.

cathalgarvey:
The travis failure appears to be sphinx-related, and I haven't edited the docs. 🤷‍♂️

tests/test_exporters.py
```diff
@@ -485,6 +497,25 @@ def test_nonstring_types_item(self):
         self.assertEqual(exported, [item])
 
 
+class JsonItemExporterToBytesTest(BaseItemExporterTest):
```
kmike: Here you're running 5 more tests from the base class - is it intended? If you don't want to do this, you can add another test case to JsonItemExporterTest and use `ie = self._get_exporter(encoding='latin-1')` instead of self.ie

kmike:
I've opened https://github.com/scrapy/scrapy/pull/3123 for the docs CI fix.

tests/test_exporters.py
cathalgarvey: I didn't mean to add the extra 5 tests, no; I'll revisit trying to use `ie` directly but the methods to handle it as an attribute and get the output were very helpful in keeping the new test-case small and tidy. :/

PR #3113, opened 2018-02-07
[MRG+1] Update robotstxt.py

WenbinZhang:
Add message to IgnoreRequest exception so that it can be detectedin the errbak method of a spider


kmike:
Thanks @WenbinZhang, I think that's a good change.

PR #3123, opened 2018-02-13
fix docs building on CI

kmike:
Don't use deprecated sphinx options and imports, to fix the build with sphinx 1.7


lopuhin:
Looks good, thanks @kmike !

PR #3124, opened 2018-02-14
BugFix: Corrected Infinite Recursion bug 

### LINKED ISSUE #3095: Cannot retry requests in `scrapy parse`
This spider:

```python
import scrapy


class HttpBinSpider(scrapy.Spider):

    name = "httpbin.org"

    start_urls = ['https://httpbin.org/'] 

    def parse(self, response):
        if response.meta.get('retried'):
            self.logger.info("done")
            return
        response.meta['retried'] = True
        yield response.request.replace(dont_filter=True)
```

works fine with `scrapy crawl httpbin.org`, but `scrapy parse "https://httpbin.org/" -d 2` fails with the following traceback:

```
  File "/home/jakob/.local/lib/python3.6/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 195, in callback
    items, requests = self.run_callback(response, cb)
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 117, in run_callback
    for x in iterate_spider_output(cb(response)):
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 195, in callback
    items, requests = self.run_callback(response, cb)
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 117, in run_callback
    for x in iterate_spider_output(cb(response)):

  [... the four lines above over and over again ...]

  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 195, in callback
    items, requests = self.run_callback(response, cb)
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 117, in run_callback
    for x in iterate_spider_output(cb(response)):
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 169, in callback
    cb = response.meta['_callback']
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/http/response/__init__.py", line 30, in meta
    return self.request.meta
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Explicitly also replacing the request callback (`yield response.request.replace(dont_filter=True, callback=self.callback)` fixes the issue.

This is possibly caused by not properly cleaning up [this](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/parse.py#L216) and [this](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/parse.py#L207) line.

Shout-out to @VMRuiz for finding this. ;)

ViralMehtaSWE:
Fixes [#3095](https://github.com/scrapy/scrapy/issues/3095).

### DIFF
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -191,6 +191,7 @@ class Command(ScrapyCommand):
 
             # parse items and requests
             depth = response.meta['_depth']
+            response.request.callback = cb
 
             items, requests = self.run_callback(response, cb)
             if opts.pipelines:
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -191,7 +191,7 @@ class Command(ScrapyCommand):
 
             # parse items and requests
             depth = response.meta['_depth']
-            response.request.callback = cb
+            response.request.callback = response.meta['_callback']
 
             items, requests = self.run_callback(response, cb)
             if opts.pipelines:
--- a/tests/test_command_parse.py
+++ b/tests/test_command_parse.py
@@ -64,11 +64,20 @@ class MyRetryRequestsSpider(CrawlSpider):
     start_urls = ['https://httpbin.org/'] 
 
     def parse(self, response):
+        self.logger.info('In parse()')
+        
+        return scrapy.Request(
+            'https://httpbin.org/ip', callback=self.second_parse)
+
+    def second_parse(self, response):
+        self.logger.info('In second_parse()')
         if response.meta.get('retried'):
             self.logger.info("done")
-            return
-        response.meta['retried'] = True
-        yield response.request.replace(dont_filter=True)
+            self.logger.info("All Good!")
+            yield scrapy.Item()
+        else:
+            response.meta['retried'] = True
+            yield response.request.replace(dont_filter=True)
 
 class MyBadCrawlSpider(CrawlSpider):
     '''Spider which doesn't define a parse_item callback while using it in a rule.'''
@@ -209,4 +218,4 @@ ITEM_PIPELINES = {'%s.pipelines.MyPipeline': 1}
         status, out, stderr = yield self.execute(
             ['--spider', 'retryRequests'+self.spider_name, '-d', '2', self.url('/html')]
         )
-        self.assertIn("""INFO: Closing spider (finished)""", to_native_str(stderr))
+        self.assertIn("""All Good!""", to_native_str(stderr))
--- a/tests/test_command_parse.py
+++ b/tests/test_command_parse.py
@@ -216,6 +216,6 @@ ITEM_PIPELINES = {'%s.pipelines.MyPipeline': 1}
     @defer.inlineCallbacks
     def test_retry_requests_in_callbacks(self):
         status, out, stderr = yield self.execute(
-            ['--spider', 'retryRequests'+self.spider_name, '-d', '2', self.url('/html')]
+            ['--spider', 'retryRequests'+self.spider_name, '-d', '3', self.url('/html')]
         )
         self.assertIn("""All Good!""", to_native_str(stderr))
--- a/tests/test_command_parse.py
+++ b/tests/test_command_parse.py
@@ -59,6 +59,16 @@ class MyGoodCrawlSpider(CrawlSpider):
     def parse(self, response):
         return [scrapy.Item(), dict(nomatch='default')]
 
+class MyRetryRequestsSpider(CrawlSpider):
+    name = 'retryRequests{0}'
+    start_urls = ['https://httpbin.org/'] 
+
+    def parse(self, response):
+        if response.meta.get('retried'):
+            self.logger.info("done")
+            return
+        response.meta['retried'] = True
+        yield response.request.replace(dont_filter=True)
 
 class MyBadCrawlSpider(CrawlSpider):
     '''Spider which doesn't define a parse_item callback while using it in a rule.'''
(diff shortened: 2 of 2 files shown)


virmht:
@kmike, @cathalgarvey , please review the PR. Please tell me why did codecov fail and how can I correct it ?
Thanks.

jdemaeyer (changes requested):
Hi @virmht, thanks a lot for the PR!

I've left some comments for the fix in the code, but overall you seem to have the right idea :)

As a general note, please branch from master before creating a PR. This will allow you to easily rebase onto master if necessary, and to avoid spurious commits like the one from Mike that's also displayed as part of this PR. I've always liked [this guide](https://akrabat.com/the-beginners-guide-to-contributing-to-a-github-project/) for creating PRs to GitHub projects.

Regarding codecov, it is complaining that the code you're adding is not accompanied by a test. To add one, you will need to add a new `test_` method to the test class found in `tests/test_command_parse.py`. The test should try to run a spider similar to the one I posted inside the code, and assert that the output is as expected. Take a look at the other test methods in that class to get an idea of how to do this, and take a look at the `setUp()` method on how to add the spider code.

You can run the tests by running `tox -e py36 -- tests/test_command_parse.py` in the repo's root directory. (You could also leave out all the arguments, but that's gonna run a loot of tests ;))

virmht:
@jdemaeyer, I have made changes in the PR as per your suggestions. Please review the PR.

I have also added 1 test so as to increase coverage, however, codecov still fails. Could you please suggest how can I correct the codecov failure ?

scrapy/commands/parse.py
```diff
@@ -191,6 +191,7 @@ def callback(response):
 
             # parse items and requests
             depth = response.meta['_depth']
+            response.request.callback = cb
```
jdemaeyer: Nice, this looks like a good spot to reset the callback.

I am going to nit-pick once more here: I would rather reset the callback to its original value (the one stored in `response.meta`) and not to the function it resolved to (`cb`). This will only have an impact if the original value was `None`, but it ensures that `scrapy parse` will behave just like `scrapy crawl` should you check something like `if response.request.callback:` (for whatever reason you would do that :sweat_smile:).

tests/test_command_parse.py
```diff
@@ -59,6 +59,16 @@ def parse_item(self, response):
...
+class MyRetryRequestsSpider(CrawlSpider):
+    name = 'retryRequests{0}'
+    start_urls = ['https://httpbin.org/'] 
+
+    def parse(self, response):
+        if response.meta.get('retried'):
+            self.logger.info("done")
+            return
+        response.meta['retried'] = True
+        yield response.request.replace(dont_filter=True)
```
jdemaeyer: It would be nice if we could extend this test a little. It should include yielded items as well as a non-default callback  (i.e. a request that has `callback` set to some method).

tests/test_command_parse.py
```diff
@@ -193,3 +203,10 @@ def test_crawlspider_no_matching_rule(self):
         )
         self.assertRegexpMatches(to_native_str(out), """# Scraped Items  -+\n\[\]""")
         self.assertIn("""Cannot find a rule that matches""", to_native_str(stderr))
+    
+    @defer.inlineCallbacks
+    def test_retry_requests_in_callbacks(self):
+        status, out, stderr = yield self.execute(
+            ['--spider', 'retryRequests'+self.spider_name, '-d', '2', self.url('/html')]
+        )
+        self.assertIn("""INFO: Closing spider (finished)""", to_native_str(stderr))
```
jdemaeyer: This will actually be printed even if the bug wasn't fixed:

```
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/commands/parse.py", line 169, in callback
    cb = response.meta['_callback']
  File "/home/jakob/.local/lib/python3.6/site-packages/scrapy/http/response/__init__.py", line 30, in meta
    return self.request.meta
RecursionError: maximum recursion depth exceeded while calling a Python object
2018-02-26 10:39:30 [scrapy.core.engine] INFO: Closing spider (finished)
```

Maybe you can check the output against something that is only printed when the spider run was successful / the bug is fixed?

jdemaeyer (commented):
Hi @virmht, thanks for the update! I've left some comments in the code :)

Don't worry about codecov for now. I assume that it doesn't realize the code is actually tested because it's called in a separate process, but as long as we can manually tell we're testing the code we should be good.

jdemaeyer:
> TDD requires us to write tests first before writing the code; so how to go about it?

@yashrsharma44 Well, you just write the tests before you write the code. ;)

Say you plan to implement a function `rounded_square()` that takes a number as argument and returns the nearest integer to its square. Start by thinking about how you want your function to _behave_, and *transform it into tests*:

```python
import unittest

from my_math_module import rounded_square
 

class RoundedSquareTestCase(unittest.TestCase):
 
    def test_integer_square(self):
        self.assertEqual(rounded_square(4), 16)
        self.assertEqual(rounded_square(0), 0)
        self.assertEqual(rounded_square(-9), 81)

    def test_float_square(self):
        self.assertEqual(rounded_square(4.01), 16)
        self.assertEqual(rounded_square(4.1), 17)
        self.assertEqual(rounded_square(0.1), 0)
        self.assertEqual(rounded_square(-4.1), 17)

    def test_invalid_type_square(self):
        with self.assertRaises(TypeError):
            rounded_square('some_string')
        with self.assertRaises(TypeError):
            rounded_square(None)
```

Only after you've completely defined the desired function behavior do you start to actually implement the function. During development you can test your progress, and your implementations correctness, by running the tests.

There's a number of benefits to this, the biggest one being (in my eyes) that it forces you to think about your specifications rather than implementation details, and consequentially makes you write tests _against these specs_ rather than against the code you have just written. [Wikipedia](https://en.wikipedia.org/wiki/Test-driven_development) has a broader overview of advantages:

> It helps ensure that the application is written for testability, as the developers must consider how to test the application from the outset rather than adding it later. It also ensures that tests for every feature get written. Additionally, writing the tests first leads to a deeper and earlier understanding of the product requirements, ensures the effectiveness of the test code, and maintains a continual focus on software quality. When writing feature-first code, there is a tendency by developers and organisations to push the developer on to the next feature, even neglecting testing entirely. The first TDD test might not even compile at first, because the classes and methods it requires may not yet exist. Nevertheless, that first test functions as the beginning of an executable specification."

virmht:
@jdemaeyer, thank you for your valuable feedback. As suggested, I have added a test which includes both: yielded items as well as a custom callback. Feel free to let me know of any further changes required to be made :)

PR #3129, opened 2018-02-17
restore request's callback after calling custom callback

malloxpb:
Restore the request callback after calling the custom callback. Bug #3095

### DIFF
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -192,7 +192,9 @@ class Command(ScrapyCommand):
             # parse items and requests
             depth = response.meta['_depth']
 
-            request.callback = cb
+            if response.request.callback:
+                response.request.callback = response.meta['_callback']
+
             items, requests = self.run_callback(response, cb)
             if opts.pipelines:
                 itemproc = self.pcrawler.engine.scraper.itemproc
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -205,12 +205,14 @@ class Command(ScrapyCommand):
                     req.meta['_depth'] = depth + 1
                     req.meta['_callback'] = req.callback
                     req.callback = callback
+                    req.callback = original_callback
                 return requests
 
         #update request meta if any extra meta was passed through the --meta/-m opts.
         if opts.meta:
             request.meta.update(opts.meta)
 
+        original_callback = request.callback
         request.meta['_depth'] = 1
         request.meta['_callback'] = request.callback
         request.callback = callback
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -192,6 +192,7 @@ class Command(ScrapyCommand):
             # parse items and requests
             depth = response.meta['_depth']
 
+            request.callback = cb
             items, requests = self.run_callback(response, cb)
             if opts.pipelines:
                 itemproc = self.pcrawler.engine.scraper.itemproc
@@ -205,14 +206,12 @@ class Command(ScrapyCommand):
                     req.meta['_depth'] = depth + 1
                     req.meta['_callback'] = req.callback
                     req.callback = callback
-                    req.callback = original_callback
                 return requests
 
         #update request meta if any extra meta was passed through the --meta/-m opts.
         if opts.meta:
             request.meta.update(opts.meta)
 
-        original_callback = request.callback
         request.meta['_depth'] = 1
         request.meta['_callback'] = request.callback
         request.callback = callback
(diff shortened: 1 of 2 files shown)


scrapy/commands/parse.py
```diff
@@ -205,12 +205,14 @@ def callback(response):
                     req.meta['_depth'] = depth + 1
                     req.meta['_callback'] = req.callback
                     req.callback = callback
+                    req.callback = original_callback
```
jdemaeyer: Hi @nctl144, thanks for the PR! :)

At this point, the custom callback method was not called yet. You are simply resetting the requests callback to its original _immediately_ after it had been set to the custom callback. Thereby, you disable the whole functionality of `scrapy parse` for all subsequent requests, and instead of printing the requests and items of the _last_ request, scrapy will print those of the first request.

You can see this if you yield an item before printing "done", like this:
```python
import scrapy


class HttpBinSpider(scrapy.Spider):

    name = "httpbin.org"

    start_urls = ['https://httpbin.org/']

    def parse(self, response):
        if response.meta.get('retried'):
            self.logger.info("done")
            yield {'some': 'item'}
            return
        response.meta['retried'] = True
        yield response.request.replace(dont_filter=True)
```

With this PR, the end of the crawl log with `scrapy parse "https://httpbin.org/" -d 2` looks like this:
```
>>> STATUS DEPTH LEVEL 1 <<<
# Scraped Items  ------------------------------------------------------------
[]
# Requests  -----------------------------------------------------------------
[<GET https://httpbin.org/>]
```

whereas with a correct fix it should look like this:
```
>>> STATUS DEPTH LEVEL 2 <<<
# Scraped Items  ------------------------------------------------------------
[{'some': 'item'}]
# Requests  -----------------------------------------------------------------
[]
```

The real call to the original callback is done in [line 195](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/parse.py#L195), so you should try to restore the original `request.callback` attribute somewhere around there.

scrapy/commands/parse.py
```diff
@@ -205,12 +205,14 @@ def callback(response):
...
                 return requests
 
         #update request meta if any extra meta was passed through the --meta/-m opts.
         if opts.meta:
             request.meta.update(opts.meta)
 
+        original_callback = request.callback
```
jdemaeyer: This is already saved in `request.meta['_callback']` (see two lines below)

nctl144:
Hey @jdemaeyer , I have fixed the code just the way you suggested. Can you please review it for me? Thank you!

nctl144:
@lopuhin @cathalgarvey hey guys, when you have time can you please review this PR for me? Let me know if there is anything that I can improve!

tests/test_command_parse.py
```diff
@@ -70,6 +70,19 @@ class MyBadCrawlSpider(CrawlSpider):
...
+
+    name = 'infinitecb{0}'
+
+    def parse(self, response):
+        if response.meta.get('retried'):
+            yield dict(foo='bar')
+            return
+
+        response.meta['retried'] = True
+        yield response.request.replace(dont_filter=True)
```
jdemaeyer: May you can add another test case that tests non-default callbacks, like the example [in this comment](https://github.com/scrapy/scrapy/pull/3124#discussion_r169702883)?

scrapy/commands/parse.py
```diff
@@ -192,6 +192,7 @@ def callback(response):
             # parse items and requests
             depth = response.meta['_depth']
 
+            request.callback = cb
```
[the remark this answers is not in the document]
nctl144: Aight I got it. I just fixed it to ensure the request callback is restored only when it is not **None** and I also reset the callback to its value in **response.meta**!

nctl144:
You are right @jdemaeyer , maybe I was just a little bit excited for the first PR to get accepted 😂.
I just added a custom callback test for this PR and fixed the code in **parse.py**. It might not be perfect yet but please let me know if there is anything that I can further improve!
And yes I actually learned a lot through this PR and some that I made in other scrapy repos. I am enjoying this so far! And thanks for your help too, I really appreciate it :)

scrapy/commands/parse.py
```diff
@@ -192,6 +192,9 @@ def callback(response):
             # parse items and requests
             depth = response.meta['_depth']
 
+            if response.request.callback:
```
gekco: Hi,
I am new to the project and to open source development as well. but would you explain why 
if response.request.callback: check is neccessary. also If i am not wrong , you can just use cb and not request.meta['_callback']

nctl144: Hey @gekco , I believe there's a comment in the review changes that has more detail about this!

Issue #3128, opened 2018-02-16
[suggest ] `response.follow` should raise a exception when called on None or an empty string, instead of crawling the current page again


NewUserHa:
`response.follow` will raise a exception when url='' or none in stead of crawl the (base) page itself again.

none will use follow to crawl the source(base) page again right? all parsers will be passed without warning if that way.

thanks

cathalgarvey:
Hi @NewUserHa - I re-worded the title, to more clearly match what I _think_ you're suggesting. Can you confirm that I understood you correctly?

This does look like an odd behaviour, and I can think of plenty of ways that a `None` could creep in there in normal circumstances, sadly. I don't think this should be the expected behaviour; maybe for special pseudo-URLs like "#" or ".", but not IMO for the empty string or None?

kmike:
If url is '', Scrapy should follow the same page, this is an intended behavior. Doing otherwise would be non-standard, and can break a lot of things.

When is url None? Could you please give an example?

cathalgarvey:
Hi @kmike, take this contrived example:

```python
In [1]: fetch("http://httpbin.org/html")
2018-02-20 13:18:40 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://httpbin.org/html> (referer: None)

In [2]: response.css("a[href]::attr(href)").extract_first() is None
Out[2]: True

In [3]: fetch(response.follow(response.css("a[href]::attr(href)").extract_first()))
2018-02-20 13:18:52 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://httpbin.org/html> (referer: None)
```

Dupefilter should catch these cases, probably. But if Dupefilter is disabled for some reason and custom logic enabled, then it's possible for this to cause unexpected behaviour.

Good to know about empty-string being expected behaviour.

kmike:
Right, I would expect response.follow to raise an exception if None is passed, +1 to fix that. From docs:

```
        Return a :class:`~.Request` instance to follow a link ``url``.
        It accepts the same arguments as ``Request.__init__`` method,
        but ``url`` can be not only an absolute URL, but also
        
        * a relative URL;
        * a scrapy.link.Link object (e.g. a link extractor result);
        * an attribute Selector (not SelectorList) - e.g.
          ``response.css('a::attr(href)')[0]`` or
          ``response.xpath('//img/@src')[0]``.
        * a Selector for ``<a>`` or ``<link>`` element, e.g.
[...]

PR #3131, opened 2018-02-21
[MRG+1] Fix part of issue #3128 - None should not be a valid type for 'url' in Response.follow

NewUserHa:
the issue reason should be: 'urljoin from urllib return baseurl if targeturl is None'

the empty url issue haven't been touched because @kmike says: 'this is an intended behavior.'


kmike:
Thanks @NewUserHa!

PR #3137, opened 2018-02-24
[RFC] scrapy-httpcache plugin

nyov:
I'd like to propose the move of the HTTPCACHE extension to scrapy-plugins.
This 'PR' shows the code to be moved, and I prepared a repository with
the httpcache code here: https://github.com/nyov/scrapy-httpcache

Pro:

- The code of the httpcache extension is quite stable and doesn't see many updates, code quality (when living elsewhere) is unlikely to degrade
- httpcache has been a `contrib` module in pre-1.0 scrapy
- httpcache is disabled by default and removal from scrapy core should not break behavior of a default installation of scrapy
- httpcache uses several 3rd party libraries (leveldb, dbm); and would include even more if more backends would be merged
- PRs with feature additions to the httpcache extension have not been merged for this reason (scmongo, deltaleveldb)
- The extension has an orthogonal 'HTTPCACHE_BACKEND' "extension" system
- A scrapy-plugin version could define the backend-requirements as optional extras, to make
  the dependency list more lightweight; such as `pip install scrapy-httpcache[leveldb]`?
- As a standalone plugin, it would be easier to fork and for the maintainer to
  keep their fork up-to-date with API changes (no need for tracking all of scrapy)
- Small speed improvement to scrapy unittests from dropping a few testcases

Contra:

- httpcache may be considered a "core" functionality (i.e. often used), so perhaps some people prefer it bundled by default

Comments?

My motive for this is to perhaps be able to merge some more features into this later
if it can be a plugin (such as the mentioned scmongo backend and the delta storage),
maybe also have a place to upstream my personal radoscache backend.

Perhaps there are even other projects that would profit from a httpcache system and this would
evolve into a more generic python module, but that's probably a bit farfetched.


nyov:
No opinions or comments? sadface

nyov:
So I believe I will be alone in this, I'll just go ahead and call it a fork for now and add my personal changes to it.
Still. I was wondering, if I maintain this as a alternative to the built-in HTTP-cache MW, would there be a place for it at [`scrapy-plugins`](https://github.com/scrapy-plugins)? Or is that org reserved for scrapinghub-internal projects?
I wouldn't mind having it there, for the chance of some co-maintainers with access to it; to try and keep it on the code quality level of scrapy. (Because I myself can only do this in my spare time.)

Who do I ask about that? ping @kmike, @dangra

grammy-jiang:
Hi, @nyov 

I am happy to see this, and I have another middleware highly developed just like this one.

I am looking forward to your comment's feedback.

kmike:
Hi @nyov! Thanks for the detailed proposal. From my part, there was no comment because on one hand, I think we absolutely want to move more things outside Scrapy, and have a richer non-core plugin ecosystem. On the other hand, I think http cache is an essential Scrapy feature, and there should be a built-in way to enable cache, with a default policy and a backend which are good enough as defaults. 

I haven't checked https://github.com/nyov/scrapy-httpcache in detail yet, and haven't compared it with Scrapy's implementation. But according to my existing knowledge, I'd prefer to handle it this way:

1. Cache middleware stays in Scrapy. If improvements are needed for this middleware, we do them in Scrapy.
2. There are still default backends provided by Scrapy. I think that Filesystem backend should be kept, while DBM backend should be moved to plugins or deprecated. I'm on fence regarding LevelDB - on one hand, it requires a third-party dependency, but on the other hand, it is the recommended backend at the moment, as it is most robust. 
3. There are still default policies provided by Scrapy. There should be a "dummy" policy for sure. I'm not sure about browser-like cache policy (RFC2616), but leaning towards keeping it in Scrapy as well.
4. DBM backend is moved to scrapy-plugins/scrapy-httpcache-extra or something like that. Other backends like MongoDB go there as well. I'm not 100% sure about delta leveldb backend, probably it could go to plugins as well.
5. Additional policies go to this plugins as well - though I'm not sure what they are. 

So, tl;dr - +1 from me to create scrapy-plugins/scrapy-cache-storages repo, move dbm backend there and add more community-provided backends. `scrapy-plugins` is certainly not for ScrapingHub-only repositories, though that's true there is no a clear process of how to create a repo there, one have to resort to asking scrapy maintainers to do that on a case-by-case basis, without an "official" communication channel.

nyov:
Thanks for the detailed response; I guess from my viewpoint scrapy is already depending on external packages to function, i.e. `parsel` and `w3lib`, so I don't see why the same dependency couldn't be true for the httpcache.

Oh, https://github.com/nyov/scrapy-httpcache is (was) just an export of the whole HttpCache middleware + extension (policies and storages) -- everything I removed in this PR --, so it is the same as Scrapy's implementation at present.

I didn't think about splitting the middleware from the policies or storage backends, for me the interface between scrapy and plugin was so obviously on 'plugging' the functionality through the `DOWNLOADER_MIDDLEWARES` setting.
But after thinking about it some, I'm not certain I would like your splitting approach.

I've just seen [this MongoDB cache](https://github.com/xiaowangwindow/scrapy-httpcache/blob/master/scrapy_httpcache/downloadermiddlewares/httpcache.py) which does mod the middleware itself, so I guess some storages will want to make changes to this part.
Yes, we could (ask to) make that MW change in scrapy; but then we'll have a tight version coupling `scrapy`<->`httpcache` instead of a loose one: the plugin would have to track scrapy version and be blocked from doing a release that depends on such change until scrapy does a new (MW) release. Subclassing would work but lose the point of not having to change `DOWNLOADER_MIDDLEWARES['s.dmw.httpcache.HttpCacheMiddleware']`.

So breaking up the functionality of "http caching" this way - middleware in one package, cache storages in another, and policies in both - IMO just makes this utterly confusing, where scrapy's feature-set is already confusing enough.

I must admit, I would then rather prefer keeping it as it is in scrapy (can still drop the DBM storage), and providing an alternative `scrapy-httpcache` as the whole package (middleware, policies, storages), where the MW may in future diverge in implementation from scrapy's but is cohesive in itself and always compatible with it's backends (which is actually the current situation). Does that make sense?

kmike:
> Thanks for the detailed response; I guess from my viewpoint scrapy is already depending on external packages to function, i.e. parsel and w3lib, so I don't see why the same dependency couldn't be true for the httpcache.

A big difference is that these packages can be used on their own, without Scrapy - dependency is uni-directional. I don't see a point in a splitting which results in a bi-directional dependency, i.e. scrapy depends on a package, and a package depends on scrapy. 

From my point of view, it should be either "package is independent, Scrapy depends on it" - like parsel, or "package depends on Scrapy, but Scrapy doesn't depend on it" - like scrapy-djangoitem or any packaged scrapy middleware/extension.

That's the reason we were able to extract https://github.com/scrapy-plugins/scrapy-djangoitem, but not https://github.com/scrapy/scrapy-itemloader: ItemLoaders are still suggested in docs, which means Scrapy would have to depend on an external package, which depends on Scrapy itself - so we haven't moved forward with scrapy-itemloader; in case of scrapy-djangoitem we just moved everything out of core and provided a backwards compatibility shim, which was subsequently removed.

For me the question is whether `scrapy-httpcache` follows `scrapy-djangoitem` or `scrapy-itemloader` pattern, and my feeling is that it is more akin to item loaders, not to django items - we want to have an option to cache right after you've installed scrapy, so scrapy should depend on scrapy-httpcache. 

At the same time, individual cache storages can be similar to scrapy-djangoitem: they are great to have in specific cases, but not essential; a good enough default storage + an extension point (and maybe pointers to "blessed" extensions) is what a framework should aim for.

> I've just seen this MongoDB cache which does mod the middleware itself, so I guess some storages will want to make changes to this part.

and this is exactly why it'd be great to have a centralized middleware in scrapy itself. As I can see, MongoDB cache added support for async cache storages, which would be a great feature for a framework, which can benefit other cache storages. From this point of view, it is better to have a single middleware with all features like this.

> Yes, we could (ask to) make that MW change in scrapy; but then we'll have a tight version coupling scrapy<->httpcache instead of a loose one: the plugin would have to track scrapy version and be blocked from doing a release that depends on such change until scrapy does a new (MW) release.

Right; this is the only advantage I can see in moving such components to separate repositories/packages: separate release cycles, and probably separate contribution process. 

It is an additional overhead though: e.g. 

* likely such components should trigger Scrapy CI when they change, so that scrapy master is checked with their master commits;
* testing setup is not shared, so e.g. fixes for pypy CI have to be applied everywhere;
* it requires more effort to track dependencies on user side, to understand which versions of packages are compatible with which versions of dependencies; 
* it can break old scrapy releases if scrapy depends on such package without specifying a max version, and a package is updated to break compatibility; 
* compatibility is harder to enforce for maintainers (e.g. Python 3.7 support can be added to Scrapy, but not to scrapy-httpcache at the same time), etc.

> I must admit, I would then rather prefer keeping it as it is in scrapy (can still drop the DBM storage), and providing an alternative scrapy-httpcache as the whole package (middleware, policies, storages), where the MW may in future diverge in implementation from scrapy's but is cohesive in itself and always compatible with it's backends (which is actually the current situation). Does that make sense?

We can't prevent anyone from creating competing implementations of Scrapy built-in components, that's great it is possible to create such components without forking Scrapy itself. But having two diverging implementations which are largely the same looks like a worst case for me - support efforts are duplicated, versions may get different bugs and different features, it is confusing for users what to use, where to report issues, where to contribute. 

Such forking makes sense for non-code-related reasons, mainly if the current contribution/maintenance process is not satisfactory, and there is no hope to bring it in order (or estimated effort is too large). If that's the main reason to move http cache out of scrapy, it'd be great to try addressing this root cause first - maybe there is a solution.

PR #3147, opened 2018-02-27
Coverage fliljeda

fliljeda:
Three coverages added and the only one I am able to make tests for is in utils/python.py. The other two are never tested and I do not understand how to call them in tests. In pyhton.py there is one branch that is never taken and is easily tested with a new test


fliljeda:
Mistake pr. Sorry

PR #3149, opened 2018-03-02
Bug Fix for (#3046) ItemLoader missing values from base item

yashrsharma44:
Bug Fix for #3046 . I haven't added test cases, so tests are not supposed to pass. Any comments regarding the flow are most welcome. I will add test case in short time.

### DIFF
--- a/scrapy/loader/__init__.py
+++ b/scrapy/loader/__init__.py
@@ -35,6 +35,8 @@ class ItemLoader(object):
         self.parent = parent
         self._local_item = context['item'] = item
         self._local_values = defaultdict(list)
+        if item:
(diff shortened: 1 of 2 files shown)


yashrsharma44:
I think that the build is not passing due to the Assertion that item['field_name'] != ['value']. As Scrapy considers each item as iterable, so the tests need to be modified a little bit. What are your thoughts ? @cathalgarvey

yashrsharma44:
I have added some changes in the test case : tests/test_loader.

kmike:
The original issue has been fixed in another PR. Thanks @yashrsharma44!

PR #3150, opened 2018-03-02
[WIP] Run tests for Python 3.7

patiences:


lopuhin:
The build does not show in GH UI for some reason, but here it is: https://travis-ci.org/scrapy/scrapy/builds/348250229

patiences:
Looks like 3.7 errored out (in 19sec)! It seems that the error log is unavailable, any advice on how to debug this? :-)

lopuhin:
Indeed, there is no output. I restarted the build, it might help to get some output. Also travis seems to not support 3.7 yet, but only 3.7-dev: https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions, so ``python: 3.7-dev`` should be more correct, while ``TOXENV`` should probably still be ``py37``

patiences:
@lopuhin Done! Thank you so much for your help.

lopuhin:
Great, now the build worked: https://travis-ci.org/scrapy/scrapy/jobs/348294263

Unfortunately, the twisted syntax error issue is blocking many tests, and it can be a while before the fix is it's merged and new twisted version is released. In the meantime, it's possible to install twisted from a branch.

nctl144:
I believe this is for this issue #3143 right :)

patiences:
@lopuhin yep, yikes! Do you mean that I should use the [PR](https://github.com/twisted/twisted/pull/966) version of Twisted that fixes the syntax error issue temporarily so that we can run the tests? How can I do that? :-) 

 @nctl144 yes it is!

lopuhin:
> Do you mean that I should use the PR version of Twisted that fixes the syntax error issue temporarily so that we can run the tests? How can I do that? :-)

@patiences yes - the easiest way seems to change ``requirements.txt`` and install Twisted from a branch on github instead of from PyPI - this way it will be picked up both locally (with/without tox) and on travis.

patiences:
@lopuhin 3.7 tests pass in the [most recent build](https://travis-ci.org/scrapy/scrapy/builds/349230674), yay! :-) 

But I'm not sure what's up with the jessie test failures -- they should not have been affected by these changes. Is this test suite known to be unstable?

lopuhin:
> 3.7 tests pass in the most recent build, yay! :-)

Awesome, great news! I expected some failures, thanks for doing this!

> But I'm not sure what's up with the jessie test failures -- they should not have been affected by these changes. Is this test suite known to be unstable?

I agree, there should be no difference for jessie, it might unstable indeed, let me try restarting this job.

lopuhin:
Now all builds have passed, I'll raise an issue if I see it failing again.

Current status: Python 3.7 tests added to tox and travis, but they use not yet merged PR to Twisted. Waiting for PR to get merged.

patiences:
Great, thank you so much @lopuhin for all your help on this! :-)

.travis.yml
```diff
@@ -23,6 +23,8 @@ matrix:
       env: TOXENV=py36
     - python: 3.6
       env: TOXENV=docs
+    - python: 3.7-dev
```
kmike: I think now, as 3.7 is released, it may be better to use Python 3.7 relase (though it may be not straightforward); according to https://github.com/travis-ci/travis-ci/issues/9815 3.7-dev is an old beta.

kmike:
Resumed in https://github.com/scrapy/scrapy/pull/3326.

Issue #1381, opened 2015-07-24
link extractor joining base href to 'tel:' directive


Itamar-Gero:
The end result I'm getting on the process_links hook is something like:

http://www.domain.com/somepage.htmltel:123456
http://www.domain.com/blog/posttel:123456

When there's an <a href="tel:123456">our phone: 123456</a> Tag

I would eliminate it at the process_links stage but it gets confusing when the 'tel:' is appended at the end of a url.

Digenis:
Do you mean URIs such as `tel:+359-123-456-789`?
(By the way, "tel" is a scheme in this case, not a directive.)

Itamar-Gero:
Thanks for the correction, yes it's the tel: scheme

redapple:
Looks weird indeed:

```
In [1]: from scrapy.http import HtmlResponse

In [7]: hrsp = HtmlResponse(body=b"""<html>

<head>
        <title>title</title>
         
</head>

<body>
<a href="tel:123456">some phone number</a>
</body>

</html>
""", url="http://www.example.com/index.html")

In [8]: LinkExtractor().extract_links(hrsp)
Out[8]: [Link(url='http://www.example.com/tel:123456', text='some phone number', fragment='', nofollow=False)]
```

tos-kamiya:
Hi,

Sorry for (possible off-the-topic?) interruption.
I've noticed that ftp links are not recognized either. 

E.g., the following code prints no links (an empty list `[]`).

```python
from scrapy.http import HtmlResponse
from scrapy.linkextractors import LinkExtractor
hrsp = HtmlResponse(body=b"""<html>
<body>
<a href="ftp://www.external.com/">another site</a>
</body>
</html>
""", url="http://www.example.com/index.html")
print(LinkExtractor().extract_links(hrsp))
```

The root cause seems a line of module `linkextractor`s (`/scrapy/linkextractors/__init__.py:44`):
```
_is_valid_url = lambda url: url.split('://', 1)[0] in {'http', 'https', 'file'}
```

I hope this will help.

kmike:
@tos-kamiya a good catch!

nctl144:
Do you think this problem still needs a fix @kmike @redapple ? If yes, is it a wise choice to create a function in w3lib to handle these cases? 😄😄

nctl144:
[...]

PR #2505, opened 2017-01-20
FTP file listing

Digenis:
I'm trying to revive #941.
I made the changes easier for review by rebasing it on master (dropped the merges),
moving some hunks from the last commit to its parent to make the diffs cleaner
and creating a separate commit from the hunk that adds the doc. 

Cc: @bernardotorres (the author) and @kmike (the reviewer)
Fixes both #1478 and #941

PR #3152, opened 2018-03-03
[MRG+1] add ftp to the scheme list

nctl144:
This might be the solution to the issue mentioned in #1381. I thought it might be a good idea to solve it :). I added ftp to link extractor so it recognizes ftp links. Please let me know if there's anything that I can do to improve this PR 😄


kmike:
Thanks @nctl144! I think it is a right thing to do. We have some issues with processing ftp links (see https://github.com/scrapy/scrapy/pull/2505), but I think it still makes sense to extract them in link extractors.

nctl144:
Thank you @kmike 😄

Issue #3139, opened 2018-02-26
FormRequest.from_response() clickdata ignores input[type=image]


Nadya24:
No way to click image inputs now, see [here](https://github.com/scrapy/scrapy/blob/73668ce4076b87d2d2493f2c9b445c643da9055a/scrapy/http/request/form.py#L165).

cathalgarvey:
Hi @Nadya24, thanks for the report!

When I first read this, I thought `input[type=image]` was for uploading images, so it's a good thing I searched it to [make sure](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/image). So, it looks like this element is generally used for clickable submit buttons, so it should be supported here.

Is this something you're interested in [contributing a PR](https://scrapy.readthedocs.io/en/latest/contributing.html) for?

virmht:
@cathalgarvey , I want to write a fix for this bug. May I submit a PR for the same ?

gekco:
has this been assigned yet? I want to work on this issue.

cathalgarvey:
@virmht - It doesn't look like you've made any commits towards this issue yet, right? Glancing at your fork, by the way, here's a tip: Make a branch for any edits you make. That allows you to push the branch to your repository, and make a pull-request from the branch, instead of your master. That way, you can make more than one PR at a time, without getting them confused.

Also, PRs are always welcome. Propose fixes here, and work on them if there are no objections. If ye feel like working together, see [this documentation on how](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/). The short version is, one of you starts the fork/branch/PR, and give the other push access.

I'd suggest, though, that @virmht has precedence here, as he suggested a fix first. What do you think?

virmht:
@cathalgarvey, I have made some commits, but haven't pushed them on GitHub. Just like you said, I have made a different branch which enabled me to raise 2 PR's simultaneously. Its

lopuhin:
Fixed in #3153, thanks @virmht !

PR #3153, opened 2018-03-03
[MRG+1] Fixed bug FormRequest.from_response() clickdata ignores input[type=image]

virmht:
BugFix for issue #3139

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -170,8 +170,8 @@ def _get_clickable(clickdata, form):
     """
     clickables = [
         el for el in form.xpath(
-            'descendant::input[re.test(@type, "^(submit|image)$", "i")]'
-            '|descendant::button[not(@type) or re.test(@type, "^submit$", "i")]',
+            'descendant::input[re:test(@type, "^(submit|image)$", "i")]'
+            '|descendant::button[not(@type) or re:test(@type, "^submit$", "i")]',
             namespaces={"re": "http://exslt.org/regular-expressions"})
         ]
     if not clickables:
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -170,11 +170,8 @@ def _get_clickable(clickdata, form):
     """
     clickables = [
         el for el in form.xpath(
-            'descendant::*[(self::input or self::button)'
-            ' and re:test(@type, "^submit$", "i")]'
-            '|descendant::*[(self::input or self::button)'
-            ' and re:test(@type, "^image$", "i")]'
-            '|descendant::button[not(@type)]',
+            'descendant::input[re.test(@type, "^(submit|image)$", "i")]'
+            '|descendant::button[not(@type) or re.test(@type, "^submit$", "i")]',
             namespaces={"re": "http://exslt.org/regular-expressions"})
         ]
     if not clickables:
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -168,15 +168,17 @@ def _get_clickable(clickdata, form):
     if the latter is given. If not, it returns the first
     clickable element found
     """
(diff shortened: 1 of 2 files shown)


virmht:
@cathalgarvey , please review the PR. Thanks :)

cathalgarvey:
Thanks @virmht I'll give it a look :)

cathalgarvey:
Looking at this, I think it's correct. And it includes a test-case.

I'm not sure though whether the XPath should be reformatted to use a single regex pattern, matching `^(submit|image)$` for example, or whether we should accept that the list of OR predicates may grow, and reformat the Xpath expression to keep one OR option per line?

E.g., the difference between a DRYer single-pattern match for images and submits:

```python
form.xpath(
    'descendant::*[(self::input or self::button)'
    ' and re:test(@type, "^(submit|image)$", "i")]'
   '|descendant::button[not(@type)]',
    namespaces={"re": "http://exslt.org/regular-expressions"})
```
..and with one-option-per-line, keeping options clearly separate to assist debug..
```python
form.xpath(
    'descendant::*[(self::input or self::button) and re:test(@type, "^submit$", "i")]'
    '|descendant::*[(self::input or self::button) and re:test(@type, "^image$", "i")]'
    '|descendant::button[not(@type)]',
    namespaces={"re": "http://exslt.org/regular-expressions"})
```

@kmike - I know you care about DRY, but Xpath expressions will be missed by linters and other automated tools, so maybe clarity in each case is a useful way to keep it clean? What do you think?

@virmht - What's your opinion? :)

cathalgarvey:
(With or without changes, though, consider this MRG+1 from me - I'll change the title to reflect)

virmht:
@cathalgarvey, I get your point. Which pattern to follow depends on the expected length of the list of OR predicates. If the list size doesn't increase by more than 3-4 then, maybe, we will be just fine with using a single regex pattern, otherwise, using the one-predicate-per-line pattern will be a better choice. So, how long, do you think, can the list get ?

scrapy/http/request/form.py
```diff
@@ -170,10 +170,12 @@ def _get_clickable(clickdata, form):
     """
     clickables = [
         el for el in form.xpath(
-            'descendant::*[(self::input or self::button)'
-            ' and re:test(@type, "^submit$", "i")]'
-            '|descendant::button[not(@type)]',
-            namespaces={"re": "http://exslt.org/regular-expressions"})
```
dangra: is it me or these lines are changed from 4 spaces indentation to 5 spaces?

dangra: TL;DR it would be much easier to review if you keep the indentation at 4 spaces.

virmht: Sure, I will make the necessary changes.

virmht: @dangra @cathalgarvey Resolved this review. Please see the new changes.

scrapy/http/request/form.py
```diff
@@ -172,6 +172,8 @@ def _get_clickable(clickdata, form):
         el for el in form.xpath(
             'descendant::*[(self::input or self::button)'
             ' and re:test(@type, "^submit$", "i")]'
+            '|descendant::*[(self::input or self::button)'
+            ' and re:test(@type, "^image$", "i")]'
```
dangra: Considering `button[type=image]` isn't valid, I would go for:

```
descendant::input[re.test(@type, "^(submit|image)$", "i")]
|descendant::button[not(@type) or re.test(@type, "^submit$", "i")]
```

virmht: Done! @dangra, please see the new changes.

virmht:
@dangra, @cathalgarvey, - suggestions are welcome :)

dangra:
thanks @virmht !

Issue #3125, opened 2018-02-15
Speed regression in ItemLoader since 1.0


juraseg:
Hi 

TLDR: ItemLoader does not reuse response.selector when you are passing response to it as argument. And looks like it was reusing it up to Scrapy==1.0

Recently we were trying to upgrade our setup to use newer version of Scrapy (we are on 1.0.5).
We noticed huge slowdown in a lot of our spiders. 

I dag down a bit and found that the bottleneck was ItemLoader, in particular when you create many ItemLoaders (one for each Item) passing response to every one of them and the response size is relatively big.

The speed drop down goes back to version 1.1. So on Scrapy==1.1 and above the performance degradation is present.

Test spider code (`testfile` is just a random file of 1MB size):

```python
# -*- coding: utf-8 -*-
import os.path

import scrapy
from scrapy.loader import ItemLoader
from scrapy.item import Item, Field

HERE = os.path.abspath(os.path.dirname(__file__))


class Product(Item):
    url = Field()
    name = Field()


class TestSpeedSpider(scrapy.Spider):
    name = 'test_speed'

    def start_requests(self):
        file_path = HERE + '/testfile'

        file_url = 'file://' + file_path
        yield scrapy.Request(file_url)

    def parse(self, response):
        for i in xrange(0, 300):
            loader = ItemLoader(item=Product(), response=response)
            loader.add_value('name', 'item {}'.format(i))
            loader.add_value('url', 'http://site.com/item{}'.format(i))

            product = loader.load_item()

            yield product
```

cathalgarvey:
Hi @juraseg - thanks for raising this! I'm looking back over the history to get familiar with how these objects were instantiated before, and now. Lots of ground to cover. But, it looks like you could deliberately provide a selector instance to the constructor for ItemLoader, and sidestep the issue entirely. Is this something you've explored?

cathalgarvey:
As an experiment, I added a line to the ItemLoader constructor that would check `response._cached_selector is not None` and would use it if present. On Py2.7 and Py3.5 locally, it seems to pass tests well enough. So, this ought to be fixable, but we should probably discuss where is the most appropriate place to re-use selectors in this way. If implemented incautiously, then it might clobber custom selector subclasses silently and unexpectedly.

kmike:
[...]

PR #3157, opened 2018-03-08
[WIP] use response selector cache

malloxpb:
Use response selector cache instead of creating new one every time an ItemLoader object is created. As mentioned in issue #3125. I created this as a sample code to continue the discussion :)

### DIFF
--- a/scrapy/loader/__init__.py
+++ b/scrapy/loader/__init__.py
@@ -26,7 +26,10 @@ class ItemLoader(object):
 
     def __init__(self, item=None, selector=None, response=None, parent=None, **context):
         if selector is None and response is not None:
-            selector = self.default_selector_class(response)
+            if response.selector.__class__ == Selector:
+                selector = response.selector
+            else:
+                selector = self.default_selector_class(response)
         self.selector = selector
         context.update(selector=selector, response=response)
         if item is None:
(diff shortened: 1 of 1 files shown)


scrapy/loader/__init__.py
```diff
@@ -26,7 +26,10 @@ class ItemLoader(object):
 
     def __init__(self, item=None, selector=None, response=None, parent=None, **context):
         if selector is None and response is not None:
-            selector = self.default_selector_class(response)
+            if response.selector.__class__ == Selector:
```
kmike: Hi @nctl144,

There are two issues with this line:

1. If a subclass overrides `default_selector_class` attribute, it won't have an effect after this change. 
2. As I mentioned in https://github.com/scrapy/scrapy/issues/3125#issuecomment-366316774, this approach has an issue: when someone accesses `response.selector` first time, Scrapy parses response body; when this condition is False, we'd be parsing response body twice (here + later, at `self.default_selector_class(response)` line). This looks unnecesary.

nctl144:
Hey @kmike , I got what you mean. So I modified the code so that we use the response.selector even when the response._cached_selector is None. Since **response.selector** returns a new object of Selector class when the selector has no cache, I think the code is still the same right?

Issue #3140, opened 2018-02-26
Round Robin Domain Crawling second scheduler to improve performance


tianhuil:
Propose merging a DomainScheduler implemented in https://github.com/tianhuil/domain_scheduler into scrapy. It scrapes in a domain-smart way: by round-robin cycling through the domains. This has two benefits:

1. Spreading out load on the target servers instead of hitting the server with many requests at once
2. Reducing delays caused by server-throttling or scrapy's own `CONCURRENT_REQUESTS_PER_IP` restrictions.  Empirical testing has shown this to be quite effective.

It implements the solution proposed in https://github.com/scrapy/scrapy/issues/1802#issue-135260562 which found similar performance improvements.  **Original proposal** was first posted https://github.com/scrapy/scrapy/issues/1802 and https://github.com/scrapy/scrapy/issues/2474.  

**Note:** It requires more than just using `SCHEDULER_PRIORITY_QUEUE` as it needs an API change to the queue (passing in a non-integer key, i.e. the domain, to a new round robin queue).  Therefore, it is dependent on first merging scrapy/queuelib#21. 

I'm happy to setup a PR to refactor `Scheduler` to allow for both `DomainScheduler` and `Scheduler` to exist in `scrapy` as there is significant code overlap if there's interest.

cathalgarvey:
This is a really interesting idea, thanks @tianhuil!

For some (most?) use-cases, this makes perfect sense, but for others (especially free-crawling over the open web) I fear it might create problems by creating a bias towards breadth-first behaviour. So my opinion is that if we merge this in as a new feature, it should start out in an experimental or opt-in status.

I haven't done enough crawls at that scale though to give a proper opinion, so I'll wait for others here to weigh in. :)

tianhuil:
@cathalgarvey: thanks for your thoughts.  My proposal was to keep both the original Scheduler (perhaps renamed as `DefaultScheduler`) and a `DomainScheduler` and have them configurable in `settings.py`'s `SCHEDULER` variable.  Obviously, we would keep `DefaultScheduler` the default choice.

tianhuil:
Implemented in PR https://github.com/scrapy/scrapy/pull/3160

kmike:
Even though RoundRobin is not implemented (it turns out to be unnecessary), the issue should be solved by #3520.

Rashid1152:
@tianhuil how we can use it with redis_scrapy?

Issue #2474, opened 2017-01-02
Optimised dequeing


mohmad-null:
A feature request.
I've got my settings set up so that Scrapy is being nice and polite to any given site. For example:

```
# Very high
CONCURRENT_REQUESTS = 128

CONCURRENT_REQUESTS_PER_IP = 3
DOWNLOAD_DELAY = 2

AUTOTHROTTLE_ENABLED = True
AUTOTHROTTLE_TARGET_CONCURRENCY = 3
AUTOTHROTTLE_START_DELAY = 1
AUTOTHROTTLE_MAX_DELAY = 15

#because I want to try and focus on breadth - I'm mostly making few a requests to many many domains
#With a few domains getting many requests
DEPTH_PRIORITY = 1
SCHEDULER_DISK_QUEUE = 'scrapy.squeues.PickleFifoDiskQueue'
SCHEDULER_MEMORY_QUEUE = 'scrapy.squeues.FifoMemoryQueue'
```

As you can see, for any given IP, scrapy is going to be slow, but in theory scrapy can handle many different IP's at once - that's why CONCURRENT_REQUESTS is very high. The problem is, it doesn't seem to.

Consider the following hypothetical queue (assume all domains are on different IPs):

> example.com/1
> example.com/2
> example.com/3
> example.com/4
> example.com/5
> example.com/6
> example.com/7
> example.com/8
> google.example/1
> google.example/2
> google.example/3
> google.example/4
> google.example/5
> google.example/6
> google.example/7
> blue.example
> yellow.example
> red.example
> green.example
> purple.example
> pink.example
> blank.example

Assuming scrapy is parsing the above top-down (and ignoring whether we're doing FIFO or reverse), what seems to happen is that scrapy will scrape all of the example.com pages first, then the google.example ones, in both instances hitting the per-ip/autothrottle limits and waiting for them to clear down.

However, the optimal way to do the above list is to note all the different domains, and then try and do them all concurrently (within the limit of CONCURRENT_REQUESTS). This doesn't appear to happen at present.

(At least, I'm guessing that's the reason I'm seeing very slow crawl rates, a low number of tcp/ip threads, as well as negligible CPU usage (<20% of a single core), and almost non-existent network usage (about 1MB per min - 130k bits/s!!!))

kmike:
[...]

Issue #1802, opened 2016-02-21
Modify scrapy scheduler to allow for using queue class other than PriorityQueue


KrisB88:
I'm currently working on a project for a client that is a rather broad crawl, something around 200-300 urls. I've been tweaking this project to maximize the scrapy performance because my item pipelines really don't add much overhead if any to the performance.  However, I ran into a problem with scrapy's built-in queue strategy.  
    The problem has been that the url queue becomes loaded with really long sequences of urls from a single domain.  Because I'm trying to maintain some kind of politeness to these sites with a reasonable download rate, the performance suffers greatly because every url has some kind of download delay which essentially stalls the scraper.  To fix this I implemented a simple queue that uses a heap which holds a PriorityQueue from queuelib to further filter urls by their domain.  Domains are managed separately in this class with heapq which allows for choosing the next url based on the domain which has been called least recently.  By doing so, I was able to increase the crawler's performance by over 500 percent using the exact same settings and only running a single instance of the spider.  This is just a first implementation that can be improved with some custom extensions, but as a first run has shown fantastic improvements in both download rates and cpu utilization.  
    There was one issue I ran into when making this change to scrapy which is that the scheduler initializes its memory queue to a PriorityQueue.  Therefore, I had to create a custom scheduler which only deviated from the built-in scheduler in its from_crawler.  **init**, and open methods.  Seems like a decent candidate for changes to the scheduler class and something I would be happy to do.  Anyways, just my thoughts and I'd love to hear what everyone else thinks.  I can share my code as well for anyone that wants to see the queue I'm using for this broad crawl.

nyov:
First, since the Scheduler is a configurable setting, that's what it's for.
That said, if all you needed to change was to use a custom priority queue, then it's probably a good candidate for making into a setting. I've wondered myself in the past why it's hard-coded there, but hadn't any need to change it.

Is #1822 doing what you had in mind there?

And if you came up with a good alternative for the PQ that may make sense as a default, that'd be nice to have a look at!

KrisB88:
Your modifications to the scheduler are exactly what I had done as well. I'll see if I can come up with a general and flexible enough default queue that could work.  I saw such an increase in performance for a generic broad crawl without any added complexity from making it distributed that I definitely think it could be worth it.  If you want to check out the queue I'm currently using and maybe offer some advice that would be great! You can find it at [https://github.com/chamilto/tnscraper/blob/master/tnscraper/queues/queue.py](url)

[...]

PR #3160, opened 2018-03-09
Adding domain

tianhuil:
Merges DomainScheduler implemented in https://github.com/tianhuil/domain_scheduler into scrapy. It scrapes in a domain-smart way: by round-robin cycling through the domains. This has two benefits:

1. Spreading out load on the target servers instead of hitting the server with many requests at once
2. Reducing delays caused by server-throttling or scrapy's own `CONCURRENT_REQUESTS_PER_IP` restrictions.  Empirical testing has shown this to be quite effective.

It implements the solution proposed in https://github.com/scrapy/scrapy/issues/1802#issue-135260562 which found similar performance improvements.  **Original proposal** was first posted https://github.com/scrapy/scrapy/issues/1802 (code no longer available), https://github.com/scrapy/scrapy/issues/2474, and https://github.com/scrapy/scrapy/issues/3140.  

**Note:** It requires more than just using `SCHEDULER_PRIORITY_QUEUE` as it needs an API change to the queue (passing in a non-integer key, i.e. the domain, to a new round robin queue).

- Implements https://github.com/scrapy/scrapy/issues/3140
- Depends on merged queuelib PR https://github.com/scrapy/queuelib/pull/21

Would love to get feedback @pablohoffman, @dangra, and @kmike!

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1095,11 +1095,11 @@ If enabled, Scrapy will respect robots.txt policies. For more information see
 SCHEDULER
 ---------
 
-Default: ``'scrapy.core.scheduler.DefaultScheduler'``
+Default: ``'scrapy.core.scheduler.Scheduler'``
 
-The scheduler to use for crawling.  Can also be set to a domain-aware that round
-robin cycles through the domains as it outputs requests to reduce the load on
-servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
+The scheduler to use for crawling.  Can also be set to a domain-aware scheduler
+that round-robin cycles through the domains as it outputs requests to reduce
+the load on servers and accelerate scraping: ``'scrapy.core.scheduler.RoundRobinScheduler'``.
 
 .. setting:: SCHEDULER_DEBUG
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1100,6 +1100,7 @@ Default: ``'scrapy.core.scheduler.Scheduler'``
 The scheduler to use for crawling.  Can also be set to a domain-aware scheduler
 that round-robin cycles through the domains as it outputs requests to reduce
 the load on servers and accelerate scraping: ``'scrapy.core.scheduler.RoundRobinScheduler'``.
+In this latter is used, priorities are ignored (see class for more details).
 
 .. setting:: SCHEDULER_DEBUG
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1097,11 +1097,12 @@ SCHEDULER
 
 Default: ``'scrapy.core.scheduler.DefaultScheduler'``
 
-The scheduler to use for crawling.
+The scheduler to use for crawling.  Can also be set to a domain-aware that round
+robin cycles through the domains as it outputs requests to reduce the load on
+servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
 
 .. setting:: SCHEDULER_DEBUG
 
-Can also be set to a domain-aware that round robin cycles through the domains as it outputs requests to reduce the load on servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
 
 SCHEDULER_DEBUG
 ---------------
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1095,11 +1095,13 @@ If enabled, Scrapy will respect robots.txt policies. For more information see
 SCHEDULER
 ---------
 
-Default: ``'scrapy.core.scheduler.Scheduler'``
+Default: ``'scrapy.core.scheduler.DefaultScheduler'``
 
 The scheduler to use for crawling.
 
 .. setting:: SCHEDULER_DEBUG
+
+Can also be set to a domain-aware that round robin cycles through the domains as it outputs requests to reduce the load on servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
 
 SCHEDULER_DEBUG
 ---------------
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -1,28 +1,31 @@
-import os
+from abc import ABC
+
+import hashlib
 import json
 import logging
+import os
 from os.path import join, exists
+
+from queuelib import RoundRobinQueue
 
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.job import job_dir
 from scrapy.utils.misc import load_object, create_instance
 from scrapy.utils.reqser import request_to_dict, request_from_dict
-from queuelib import RoundRobinQueue
 
 logger = logging.getLogger(__name__)
 
+
 def _make_file_safe(string):
     """
-    Make key file safe.
-
-    Warning: not a general function.  Works for all numbers and the vast majority of domains.
-    Collisions in file names will only degrade (midly) performance,
-    not lead to incorrect results.
+    Make string file safe but readable.
     """
-    return "".join([c if c.isalnum() or c in {'-._'} else '_' for c in string])
+    clean_string = "".join([c if c.isalnum() or c in '-._' else '_' for c in string])
+    hash_string = hashlib.md5(string).hexdigest()
+    return "{}-{}".format(clean_string[:40], hash_string[:10])
 
 
-class BaseScheduler(object):
+class BaseScheduler(ABC):
 
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
                  logunser=False, stats=None, pqclass=None):
@@ -82,7 +85,6 @@ class BaseScheduler(object):
             self._mqpush(request)
             self.stats.inc_value('scheduler/enqueued/memory', spider=self.spider)
         self.stats.inc_value('scheduler/enqueued', spider=self.spider)
-        self.stats.inc_value('scheduler/enqueued/key/{}'.format(self.request_key(request)))
         return True
 
     def next_request(self):
@@ -157,7 +159,7 @@ class BaseScheduler(object):
             return dqdir
 
 
-class DefaultScheduler(BaseScheduler):
+class Scheduler(BaseScheduler):
     """
     Key is the priority of the request (an integer)
     """
@@ -166,16 +168,22 @@ class DefaultScheduler(BaseScheduler):
         return -request.priority
 
 
-class DomainScheduler(BaseScheduler):
+class RoundRobinScheduler(BaseScheduler):
     """
-    Key is the domain and we round robin.  The pqclass parameter is ignored.
+    Key is the domain and we round robin, choosing the new request to be from
+    the domain that was requested last.  The default Scheduler sends multiple
+    consecutive requests to a domain if multiple links were discovered and
+    added consecutively.  This scheduler spreads those out that workload.
+
+    Uses `RoundRobinQueue`.  The pqclass parameter and SCHEDULER_PRIORITY_QUEUE
+    are ignored.
     """
 
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
                  logunser=False, stats=None, pqclass=None):
-        super(DomainScheduler, self).__init__(dupefilter, jobdir=jobdir, dqclass=dqclass,
-                                              mqclass=mqclass, logunser=logunser,
-                                              stats=stats, pqclass=RoundRobinQueue)
+        super(RoundRobinScheduler, self).__init__(dupefilter, jobdir=jobdir, dqclass=dqclass,
+                                                  mqclass=mqclass, logunser=logunser,
+                                                  stats=stats, pqclass=RoundRobinQueue)
 
     def _request_key(cls, request):
         return urlparse_cached(request).hostname or ''
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -1,3 +1,4 @@
+import abc
 import hashlib
 import json
 import logging
@@ -24,6 +25,7 @@ def _make_file_safe(string):
 
 
 class BaseScheduler(object):
+    __metaclass__ = abc.ABCMeta
 
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
                  logunser=False, stats=None, pqclass=None):
@@ -53,6 +55,7 @@ class BaseScheduler(object):
 
         return self._request_key(request)
 
+    @abc.abstractmethod
     def _request_key(self, request):
         raise NotImplementedError
 
@@ -132,6 +135,7 @@ class BaseScheduler(object):
     def _newmq(self, key):
         return self.mqclass()
 
+    @abc.abstractmethod
     def _newdq(self, key):
         raise NotImplementedError
 
@@ -161,11 +165,11 @@ class Scheduler(BaseScheduler):
     Key is the priority of the request (an integer)
     """
 
-    def _request_key(cls, request):
+    def _request_key(self, request):
         return -request.priority
 
     def _newdq(self, key):
-        return self.dqclass(join(self.dqdir, 'p%s' % priority))
+        return self.dqclass(join(self.dqdir, 'p%s' % key))
 
 
 class RoundRobinScheduler(BaseScheduler):
@@ -185,7 +189,7 @@ class RoundRobinScheduler(BaseScheduler):
                                                   mqclass=mqclass, logunser=logunser,
                                                   stats=stats, pqclass=RoundRobinQueue)
 
-    def _request_key(cls, request):
+    def _request_key(self, request):
         return urlparse_cached(request).hostname or ''
 
     def _newdq(self, key):
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -1,5 +1,4 @@
 from abc import ABC
-
 import hashlib
 import json
 import logging
@@ -21,7 +20,7 @@ def _make_file_safe(string):
     Make string file safe but readable.
     """
     clean_string = "".join([c if c.isalnum() or c in '-._' else '_' for c in string])
-    hash_string = hashlib.md5(string).hexdigest()
+    hash_string = hashlib.md5(string.encode('utf8')).hexdigest()
     return "{}-{}".format(clean_string[:40], hash_string[:10])
 
 
@@ -97,7 +96,6 @@ class BaseScheduler(ABC):
                 self.stats.inc_value('scheduler/dequeued/disk', spider=self.spider)
         if request:
             self.stats.inc_value('scheduler/dequeued', spider=self.spider)
-            self.stats.inc_value('scheduler/dequeued/key/{}'.format(self.request_key(request)))
         return request
 
     def __len__(self):
@@ -136,7 +134,7 @@ class BaseScheduler(ABC):
         return self.mqclass()
 
     def _newdq(self, key):
-        return self.dqclass(join(self.dqdir, _make_file_safe('k%s' % key)))
+        raise NotImplementedError
 
     def _dq(self):
         statef = join(self.dqdir, 'active.json')
@@ -167,6 +165,9 @@ class Scheduler(BaseScheduler):
     def _request_key(cls, request):
         return -request.priority
 
+    def _newdq(self, key):
+        return self.dqclass(join(self.dqdir, 'p%s' % priority))
+
 
 class RoundRobinScheduler(BaseScheduler):
     """
@@ -180,10 +181,13 @@ class RoundRobinScheduler(BaseScheduler):
     """
 
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
-                 logunser=False, stats=None, pqclass=None):
+                 logunser=False, stats=None):
         super(RoundRobinScheduler, self).__init__(dupefilter, jobdir=jobdir, dqclass=dqclass,
                                                   mqclass=mqclass, logunser=logunser,
                                                   stats=stats, pqclass=RoundRobinQueue)
 
     def _request_key(cls, request):
         return urlparse_cached(request).hostname or ''
+
+    def _newdq(self, key):
+        return self.dqclass(join(self.dqdir, _make_file_safe('k%s' % key)))
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -1,4 +1,3 @@
-from abc import ABC
 import hashlib
 import json
 import logging
@@ -24,7 +23,7 @@ def _make_file_safe(string):
     return "{}-{}".format(clean_string[:40], hash_string[:10])
 
 
-class BaseScheduler(ABC):
+class BaseScheduler(object):
 
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
                  logunser=False, stats=None, pqclass=None):
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -175,8 +175,8 @@ class RoundRobinScheduler(BaseScheduler):
     consecutive requests to a domain if multiple links were discovered and
     added consecutively.  This scheduler spreads those out that workload.
 
-    Uses `RoundRobinQueue`.  The pqclass parameter and SCHEDULER_PRIORITY_QUEUE
-    are ignored.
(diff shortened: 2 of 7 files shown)


tianhuil:
Thanks for deploying `queuelib` to pypi @dangra.  It's now passing tests.

kmike:
Thanks @tianhuil for pushing it forward!

But I'd like to give it more thought before merging; round-robin is an improvement over what we have now, but it is not the best solution (see e.g. https://github.com/scrapy/scrapy/issues/2474#issuecomment-288222104). Round-robin assumes requests from all domains are handled with roughly the same speed, which is not the case in a real world, esp. if you're not doing broad crawl, but crawling a specific list of domains.

scrapy/core/scheduler.py
```diff
@@ -132,3 +141,27 @@ def _dqdir(self, jobdir):
...
+    """
+
+    def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
+                 logunser=False, stats=None, pqclass=None):
+        super(DomainScheduler, self).__init__(dupefilter, jobdir=jobdir, dqclass=dqclass,
+                                              mqclass=mqclass, logunser=logunser,
+                                              stats=stats, pqclass=RoundRobinQueue)
+
+    def request_key(cls, request):
+        return urlparse(request.url).netloc
```
kmike: If we go with round robin, it should be probably based on `request.meta['download_slot']`, or use the same algorithm. 

I also think that's good to have it overridable via request.meta, e.g. in https://github.com/TeamHG-Memex/linkdepth/blob/master/queues.py#L21 I used `request.meta['scheduler_slot']`. This allows to tweak these slots from spiders, and also allows to enable round-robin queue by default: scheduler_slot is empty, so all requests go to the same queue, keeping existing behavior; users who need round-robin can set a scheduler slot per-request.

scrapy/core/scheduler.py
```diff
@@ -6,11 +6,15 @@
 from scrapy.utils.reqser import request_to_dict, request_from_dict
 from scrapy.utils.misc import load_object
 from scrapy.utils.job import job_dir
+from queuelib import RoundRobinQueue
 
 logger = logging.getLogger(__name__)
 
+def _make_file_safe(string):
+    return "".join([c if c.isalnum() or c in {'-._'} else '_' for c in string])
```
kmike: it seems this function is lossy - won't it make some of the requests take the same slot why the shouldn't?

whalebot-helmsman:
@tianhuil Do you have will to work on proposed changes?

tianhuil:
I'm happy to make the smaller changes proposed by @kmike but I thought this was held up by considerations of https://github.com/scrapy/scrapy/pull/3160#issuecomment-372788831

Let me know if I should proceed with everything except for https://github.com/scrapy/scrapy/pull/3160#issuecomment-372788831

whalebot-helmsman:
Considerations in https://github.com/scrapy/scrapy/pull/3160#issuecomment-372788831are reasonable. At the same time. this PR is a prerequisite to address them. After this PR landed in trunk we can continue work on smarter way of scheduling requests.

tianhuil:
OK -- I"ll respond to everything except for https://github.com/scrapy/scrapy/pull/3160#issuecomment-372788831 to help move things forward.  Happy to do it.  Give me a little bit.

scrapy/core/scheduler.py
tianhuil: Agree.  But `c.isalnum() or c in {'-._'}` handles the vast majority of domain characters and any collisions will only degrade performance.  I added a comment.

scrapy/core/scheduler.py
tianhuil: Thanks!  Just implemented.

tianhuil:
Sorry for the delay.  Reponded to both of @kmike's helpful comments.

akirmse:
Is there any news on this?  I've been eagerly awaiting this feature because I'm also seeing scrapy get stuck on a single domain in a targeted crawl.

scrapy/core/scheduler.py
```diff
@@ -61,6 +82,7 @@ def enqueue_request(self, request):
             self._mqpush(request)
             self.stats.inc_value('scheduler/enqueued/memory', spider=self.spider)
         self.stats.inc_value('scheduler/enqueued', spider=self.spider)
+        self.stats.inc_value('scheduler/enqueued/key/{}'.format(self.request_key(request)))
```
kmike: Hm, I like an extra debugging options in stats, but we may have a problem if a spider is to download e.g. pages from 100K domains, one per domain - stats would become huge and unreadable.

scrapy/core/scheduler.py
```diff
@@ -45,9 +66,9 @@ def open(self, spider):
 
     def close(self, reason):
         if self.dqs:
-            prios = self.dqs.close()
+            state = self.dqs.close()
             with open(join(self.dqdir, 'active.json'), 'w') as f:
-                json.dump(prios, f)
+                json.dump(state, f)
```
kmike: 👍 to rename it to state. I wonder if we should rename the file as well (though for backwards compatibility an old name should be still supported, when reading it).

scrapy/core/scheduler.py
```diff
@@ -3,14 +3,26 @@
...
 
 logger = logging.getLogger(__name__)
 
+def _make_file_safe(string):
+    """
+    Make key file safe.
+
+    Warning: not a general function.  Works for all numbers and the vast majority of domains.
+    Collisions in file names will only degrade (midly) performance,
+    not lead to incorrect results.
```
kmike: I'm still not convinced we should use a function with known limitations, while there is an easy workaround - e.g. add md5 or sha1 hash (or its part?) of the original content to the end. I'm not sure, but replacing a list comprehension with a simpler regex + an md5 hash may even make this code faster. 

We also probably want to limit string length, as file systems can have problems with that.

If we're to be strict, `isalnum` check can also be problematic, as it doesn't guarantee ascii.

Anyways, I agree that the current code works in most cases, and probably it is not a big deal to leave it as-is, but it doesn't seem too hard to make it work in all cases: be more aggressive with the human-readable part, add a hash.

scrapy/core/scheduler.py
```diff
@@ -34,6 +46,15 @@ def from_crawler(cls, crawler):
         return cls(dupefilter, jobdir=job_dir(settings), logunser=logunser,
                    stats=crawler.stats, pqclass=pqclass, dqclass=dqclass, mqclass=mqclass)
 
+    def request_key(self, request):
+        if 'scheduler_slot' in request.meta:
+            return request.meta['scheduler_slot']
```
kmike: Sorry for being misleading, but now I'm not sure we need `scheduler_slot`, probably reusing `download_slot` is better. Though I don't recall why exactly :) //cc @dangra do you recall our discussion?

scrapy/core/scheduler.py
```diff
@@ -34,6 +46,15 @@ def from_crawler(cls, crawler):
...
+
+        return self._request_key(request)
+
+    def _request_key(self, request):
+        raise NotImplementedError
```
kmike: If this method is required, it is better to use abstract classes (see https://docs.python.org/3/library/abc.html). 

We use twisted interfaces in a few other places, which solve a similar issue, but I'd prefer to move to abc everywhere (though I don't know what others think :)

scrapy/core/scheduler.py
kmike: Also, a nitpick: one more blank line is needed before this function according to pep8.

scrapy/core/scheduler.py
Desprado666: 感觉你们好牛逼呀

scrapy/core/scheduler.py
```diff
@@ -132,3 +155,27 @@ def _dqdir(self, jobdir):
             if not exists(dqdir):
                 os.makedirs(dqdir)
             return dqdir
+
+
+class DefaultScheduler(BaseScheduler):
```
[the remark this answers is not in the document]
tianhuil: Done

docs/topics/settings.rst
```diff
@@ -1106,12 +1106,14 @@ If enabled, Scrapy will respect robots.txt policies. For more information see
...
 ---------
 
-Default: ``'scrapy.core.scheduler.Scheduler'``
+Default: ``'scrapy.core.scheduler.DefaultScheduler'``
 
 The scheduler to use for crawling.
 
 .. setting:: SCHEDULER_DEBUG
 
+Can also be set to a domain-aware that round robin cycles through the domains as it outputs requests to reduce the load on servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
```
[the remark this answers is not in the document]
tianhuil: Done

scrapy/core/scheduler.py
```diff
@@ -132,3 +155,27 @@ def _dqdir(self, jobdir):
...
+    """
+    Key is the priority of the request (an integer)
+    """
+
+    def _request_key(cls, request):
+        return -request.priority
+
+
+class DomainScheduler(BaseScheduler):
```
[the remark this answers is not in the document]
tianhuil: Done

scrapy/core/scheduler.py
```diff
@@ -132,3 +155,27 @@ def _dqdir(self, jobdir):
...
+    """
+    Key is the domain and we round robin.  The pqclass parameter is ignored.
```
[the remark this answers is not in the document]
tianhuil: Done, except I decided not to ask the user to specify `SCHEDULER_PRIORITY_QUEUE`.

scrapy/core/scheduler.py
```diff
@@ -3,14 +3,26 @@
 import logging
 from os.path import join, exists
 
-from scrapy.utils.reqser import request_to_dict, request_from_dict
-from scrapy.utils.misc import load_object
+from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.job import job_dir
+from scrapy.utils.misc import load_object
+from scrapy.utils.reqser import request_to_dict, request_from_dict
+from queuelib import RoundRobinQueue
```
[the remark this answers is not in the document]
tianhuil: Done.

scrapy/core/scheduler.py
tianhuil: Thanks.

scrapy/core/scheduler.py
tianhuil: Removed.

scrapy/core/scheduler.py
```diff
@@ -132,3 +155,27 @@ def _dqdir(self, jobdir):
...
+    """
+
+    def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
+                 logunser=False, stats=None, pqclass=None):
+        super(DomainScheduler, self).__init__(dupefilter, jobdir=jobdir, dqclass=dqclass,
+                                              mqclass=mqclass, logunser=logunser,
+                                              stats=stats, pqclass=RoundRobinQueue)
+
+    def _request_key(cls, request):
+        return urlparse_cached(request).hostname or ''
```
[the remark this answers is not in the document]
tianhuil: I don't really know what to do here, perhaps because I don't know what `download_slot` is meant to do.  Perhaps you could contribute the commit here?

tianhuil:
@kmike: let me know what your thoughts are here.

tianhuil:
@kmike: let me know what your thoughts are: would love to get this merged.

docs/topics/settings.rst
```diff
@@ -1106,12 +1106,15 @@ If enabled, Scrapy will respect robots.txt policies. For more information see
 SCHEDULER
 ---------
 
-Default: ``'scrapy.core.scheduler.Scheduler'``
+Default: ``'scrapy.core.scheduler.DefaultScheduler'``
```
kmike: Shouldn't it be Scheduler?

docs/topics/settings.rst
```diff
@@ -1106,12 +1106,15 @@ If enabled, Scrapy will respect robots.txt policies. For more information see
...
 
-The scheduler to use for crawling.
+The scheduler to use for crawling.  Can also be set to a domain-aware that round
+robin cycles through the domains as it outputs requests to reduce the load on
+servers and accelerate scraping: ``'scrapy.core.scheduler.DomainScheduler'``.
```
kmike: DomainScheduler is renamed

scrapy/core/scheduler.py
```diff
@@ -1,16 +1,31 @@
-import os
+from abc import ABC
```
kmike: I don't think it works in Python 2

scrapy/core/scheduler.py
```diff
@@ -1,16 +1,31 @@
...
+
```
kmike: there shouldn't be a blank line between stdlib imports

scrapy/core/scheduler.py
```diff
@@ -1,16 +1,31 @@
...
 logger = logging.getLogger(__name__)
 
 
-class Scheduler(object):
+def _make_file_safe(string):
+    """
+    Make string file safe but readable.
+    """
+    clean_string = "".join([c if c.isalnum() or c in '-._' else '_' for c in string])
+    hash_string = hashlib.md5(string).hexdigest()
```
kmike: this doesn't work if a string is unicode, while the code above assumes it is unicode (in Python 3)

scrapy/core/scheduler.py
kmike: The idea was to mark this method as abstractmethod, so that it fails earlier

scrapy/core/scheduler.py
```diff
@@ -73,6 +97,7 @@ def next_request(self):
                 self.stats.inc_value('scheduler/dequeued/disk', spider=self.spider)
         if request:
             self.stats.inc_value('scheduler/dequeued', spider=self.spider)
+            self.stats.inc_value('scheduler/dequeued/key/{}'.format(self.request_key(request)))
```
kmike: if `enqueued` stat is removed, this one should be removed as well

scrapy/core/scheduler.py
```diff
@@ -99,28 +124,28 @@ def _dqpush(self, request):
...
                 return request_from_dict(d, self.spider)
 
-    def _newmq(self, priority):
+    def _newmq(self, key):
         return self.mqclass()
 
-    def _newdq(self, priority):
-        return self.dqclass(join(self.dqdir, 'p%s' % priority))
+    def _newdq(self, key):
+        return self.dqclass(join(self.dqdir, _make_file_safe('k%s' % key)))
```
kmike: hm, actually this is backwards incompatible - after this PR scrapy won't be able to resume crawls from earlier Scrapy versions

scrapy/core/scheduler.py
```diff
@@ -132,3 +157,33 @@ def _dqdir(self, jobdir):
...
+    the domain that was requested last.  The default Scheduler sends multiple
+    consecutive requests to a domain if multiple links were discovered and
+    added consecutively.  This scheduler spreads those out that workload.
+
+    Uses `RoundRobinQueue`.  The pqclass parameter and SCHEDULER_PRIORITY_QUEUE
+    are ignored.
+    """
+
+    def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
+                 logunser=False, stats=None, pqclass=None):
```
kmike: I think it should be possible to override `pqclass` argument somehow, now it is silently discarded

scrapy/settings/default_settings.py
```diff
@@ -241,7 +241,7 @@
 
 ROBOTSTXT_OBEY = False
 
-SCHEDULER = 'scrapy.core.scheduler.Scheduler'
+SCHEDULER = 'scrapy.core.scheduler.DefaultScheduler'
```
kmike: there is no DefaultScheduler class

kmike:
hey @tianhuil, I've left some more comments. Besides them, there are the following issues:

1. Merge conflicts should be resolved.
2. Tests should pass.
3. It is still not documented that request priorities don't work. I'd prefer them working, but if they don't, it should be clearly documented, both in scheduler docs and in Request docs.

scrapy/core/scheduler.py
tianhuil: fixed

scrapy/core/scheduler.py
tianhuil: Good point.

scrapy/core/scheduler.py
tianhuil: I put this into the base class.

scrapy/core/scheduler.py
tianhuil: That's a good point.  I just removed it from the signature of the child class.

scrapy/settings/default_settings.py
tianhuil: Fixed

scrapy/core/scheduler.py
tianhuil: Don't know enough about `ABC`.  Perhaps we can just leave out this part and have it inherit from `object`?

tianhuil:
@kmike: it's I've documented that priorities have not been used.  Looks like it's passing.

scrapy/core/scheduler.py
tianhuil: Fixed.

tianhuil:
@kmike: let me know if you want to merge this.  Would like to do it before conflicts re-emerge.

tianhuil:
@kmike: let me know if you have any other questions or comments. Would like to do it before conflicts re-emerge.

tianhuil:
@kmike: let me know if you have any other questions or comments. Would like to do it before conflicts re-emerge.

scrapy/core/scheduler.py
```diff
@@ -1,16 +1,31 @@
...
+    """
+    Make string file safe but readable.
+    """
+    clean_string = "".join([c if c.isalnum() or c in '-._' else '_' for c in string])
+    hash_string = hashlib.md5(string.encode('utf8')).hexdigest()
+    return "{}-{}".format(clean_string[:40], hash_string[:10])
+
+
+class BaseScheduler(object):
+    __metaclass__ = abc.ABCMeta
```
kmike: this doesn't work in Python 3. Please use [six.with_metaclass](https://pythonhosted.org/six/#six.with_metaclass)

kmike:
Hey @tianhuil! Sorry for going back and forth. 

My main concern is a lack of support for request priorities. There are implementations which support request priorities - in fact, all other implementations I'm aware of (mentioned in previous tickets) support request priorities - they are incomplete in different ways though. 

It is not only a user-facing feature, to allow users to set custom priorities for different Requests (`scrapy.Request(url, self.parse, priority=5`). Scrapy uses request priorities itself, internally - they are used for retries and redirects; they are also a base of DepthMiddleware. 

Not having priority support may cause strange issues: for example, redirects won't necessarily be on top of the queue, so any redirect can potentially go to the end of the crawl, and, depending on a spider, may be never crawled - it seems to be the case if user uses FIFO queues, as recommended in FAQ [section](https://doc.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order) on how to configure BFO crawl. Actually, BFO won't work at all, as DEPTH_PRIORITY won't work.

Documentation should be updated in several more places, maybe even regardless of the implementation:

* Request [docs](https://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request), 
* [Broad Crawls](https://doc.scrapy.org/en/latest/topics/broad-crawls.html), 
* this FAQ [entry](https://doc.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order).

tianhuil:
@kmike: can you provide a clear set of specifications you want this PR to meet?  I'm happy to work on this but I don't have time to chase a moving target.

kmike:
@tianhuil I'm fine with almost everything which makes things better and keeps the existing features working. I haven't realized at first that not supporting priorities is not just a problem with `priority` argument, it is also a problem for retries, redirects, BFO, scrapy-splash, etc - sorry about that. So I think RoundRobin with support for priorities is good. There are several ways to support priorities in Round-Robin:

1. Priorities are respected for each domain, but domain is chosen in a round-robin fashion.
2. Only requests of the same priority are chosen in a round-robin fashion; priorities are respected globally.

(1) makes more sense to me as a default. We just need to make sure it is clearly documented.

---

Eventually we should be aiming for an even more efficient scheduler. From my point of view, we're fixing a bug here: Scrapy may under-utilize the downloader by filling it with requests it can't download because of its own concurrency limits. 

Round-robin is a solution, but a partial one: it assumes all domains can be downloaded with the same speed. It is not always the case. For example, let's say we have domains A and B, both with many requests in a queue, and request can be finished twice as fast for domain A. Global concurrency is set to 4, and per-domain concurrency is limited to 2. With Round-Robin downloader queue will be filled like this:

1. `____ -> ABAB`: requests are added in a round-robin fashion, total size is 4.
2. `_BAB -> ABAB`: A is downloaded, next request is for A, we add it to the queue
3. `AB_B -> ABBB`: A is downloaded (it is twice as fast), B is added according to round-robin rules
4. `A_BB -> AABB`: B is downloaded, we add A
5. `_ABB -> BABB`: A is downloaded, we add B
6. `B_BB -> BABB`: A is downloaded, we add A
7. `BAB_ -> BABB`: B is downloaded, we add B
...

Note how we almost always have less A requests in the downloader queue - one instead of optimal two; it means in this case there are no parallel requests to domain A, while user requested 2 parallel requests. At the same time, there are more B requests than can be downloaded in parallel (as the limit is 2), they're just sitting in the queue, taking slots which could have been used for other domains.

In broad crawls it is rather common to get some domains which can timeout on all requests. Maybe some DNS issues. Default Scrapy timeout is 180s; with 2 retries it is 180*3=540s to process a request in a worst case. If for the "average" domain speed is 1s, and we download 100 domains in parallel, we may easily get downloader queue filled 4/5 (or so, I haven't estimated it properly) with requests from this slow domain, leaving less opportunity for parallelization of other domains, and so crawling slower than possible for no good reason.

To solve this issue scheduler should avoid adding request to the downloader if there is no free slot for it. This is how it works in frontera. Round Robin can be modified to do the same - and that's the reason I'm fine with merging round-robin implementation - it can be modified to support this feature in future.  For example, it can keep track of the current downloader state using `request_reached_downloader` (see https://github.com/scrapy/scrapy/pull/3393) and other signals, and skip adding a request from a domain if adding it would make Downloader go over concurrency limit for this domain. Instead of signals we may also query Downloader status directly.

akirmse:
Hello, is there any progress on this?  We have an application where we have a targeted crawl over a small number of domains.  Performance is currently very poor, as the queue quickly gets filled with a single slow domain.

We don't need priorities or anything fancy.  A limited, 90% solution today would be infinitely preferable to a perfect 100% solution in a year (this review has already taken 7 months).

kmike:
hey @akirmse! There are third-party solutions which you can use without changing Scrapy. For example, you can use RoundRobinPriorityQueue from https://github.com/TeamHG-Memex/linkdepth/blob/master/queues.py:

```
SCHEDULER_PRIORITY_QUEUE = 'queues.RoundRobinPriorityQueue'
SCHEDULER_DISK_QUEUE = 'queues.DiskQueue'
```

It actually supports priorities (so all Scrapy features works properly), but has other limitations - crawl resume is not supported. Priorities is not something fancy, they are needed for e.g. for retries, redirects, BFO crawling, scrapy-splash to work properly.

There are also other workarounds, e.g. using frontera. It is not like this is impossible to do with Scrapy, and it is not even a lot of code to get a partial solution which doesn't need any changes in Scrapy core. That's why I'd prefer to have something proper, integrated well with all Scrapy features, if we're going to make a change in Scrapy itself.

amonroe:
@kmike Can this be merged in? @tianhuil makes scrapy usable for more than one domain.  The current version is unusable due to the scheduler.  As this isn't the default scheduler, declaring it beta and getting usage would solve a lot of issues that have existed for years.

kmike:
@amonroe https://github.com/scrapy/scrapy/pull/3520 is much closer to get merged at the moment; it solves the same problem, but covers more edge cases, has more tests, and has better behavior - looking for downloader activity instead of using round-robin can speed up crawling significantly in many practical cases.

amonroe:
Thanks @kmike.  I'll try that PR as well.  So far #3160 has started to show some (not much) balancing across domains after having been blackholed on one for a day.  Prior backlog is probably impacting results.   No issues as of yet.

kmike:
Closing it, as https://github.com/scrapy/scrapy/pull/3520 is merged. Thanks @tianhuil for getting the ball rolling!

Issue #3097, opened 2018-01-29
`item_urls` is not defined in example


tobiasfielitz:
[here](https://github.com/scrapy/scrapy/blob/1.5/docs/topics/debug.rst) in `def parse`

kmike:
Hey @tobiasfielitz,

It seems the comment in `parse` method is not 100% clear: the idea is that a spider is not a complete example, but there is some code which should collect `item_urls`. The same happens in other `parse_..` methods. Do you have suggestions how to improve it, to make it more clear? Pull requests are welcome :)

FannyVieira:
Are item_urls those defined  in the url of the response?

cathalgarvey:
Hi @FannyVieira - I think the comment on [line 21](https://github.com/scrapy/scrapy/blame/1.5/docs/topics/debug.rst#L21), and other comments like it, are supposed to be placeholders for code-that's-not-included-in-this-demo. So `item_urls` is simply assumed to be parsed somehow from the response, but we don't show how that happened.

This isn't clear to everyone, so either we should:

* Include a minimal-but-complete example where all variables are extracted and defined fully
* Change the way we present "stuff happened, but we didn't show you" in the examples. Maybe show the variables being defined by a method call, but the method-body is commented as `# < processing code not shown >` to be super clear.

By the way, are you interested in [GSoC 2018](http://gsoc2018.scrapinghub.com/participate/)? :)

malikrohit16:
I think may be item_urls refers to start_urls.

bhaveshpoddar94:
First Time Contributor. Can I take a look into it?

FannyVieira:
Hi @cathalgarvey so by what I understand url items is the same as self.start_urls? If it's just that, I can open a pr to make the necessary changes..
Yes, I really want to participate in GSOC, I study computer science but i have seventeen years old, Can I participate anyway?

malikrohit16:
Hello @FannyVieira i am afraid that you won't be able to participate this year but you still can contribute to the community if you like. This will give you upper hand for the next year.

cathalgarvey:
Hi @FannyVieira - Well, if you turn 18 before 27th March, I think you can still apply under the PSF and related orgs like Scrapinghub. Otherwise, please keep us in mind for next year! And if you're keen to contribute, go right ahead and don't be afraid to ask for advice. :)

ayushmankoul:
[...]

PR #3181, opened 2018-03-20
Debug spiders doc

nctl144:
I updated the doc to make the comments in those functions clearer to the reader. This issue is mentioned in #3097. This is what suggested by @cathalgarvey :) Let me know if there is anything that I can do to improve this PR

### DIFF
--- a/docs/topics/debug.rst
+++ b/docs/topics/debug.rst
@@ -19,21 +19,21 @@ Consider the following scrapy spider below::
 
         def parse(self, response):
             # < your code goes here >
-            # which should collect `item_urls`
+            # which collects `item_urls`
             for item_url in item_urls:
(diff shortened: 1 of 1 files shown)


docs/topics/debug.rst
```diff
@@ -18,19 +18,22 @@ Consider the following scrapy spider below::
             )
 
         def parse(self, response):
-            # collect `item_urls`
+            # < your code goes here >
+            # which collects `item_urls`, an array of urls
```
djunzu: I think it would be better something like

```python
item_urls = # collects some urls
```

or 

```python
# collect some urls and populate `item_urls` with it
```

wRAR:
The original issue was fixed in other PRs. If someone wants to propose improved warning for what is currently in the repo, feel free to open a new PR.

PR #3187, opened 2018-03-26
Update files.py

Lam17:
Fix the problem "file name too long"


scrapy/pipelines/files.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, basedir):
         self.created_directories = defaultdict(set)
 
     def persist_file(self, path, buf, info, meta=None, headers=None):
-        absolute_path = self._get_filesystem_path(path)
+        absolute_path = self._get_filesystem_path(path)[:255]
```
lopuhin: Does this work correctly if ``absolute_path`` is longer than 255? It seems that ``path`` and ``absolute_path`` won't match any more.

PR #3201, opened 2018-04-05
[MRG+1] fix a mistake in topic spider-middleware.rst

grammy-jiang:
The return iterable of `process_spider_exception` should contain `scrapy.http.Request`, not `scrapy.http.Response`.

### DIFF
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -116,7 +116,7 @@ following methods:
         method (from other spider middleware) raises an exception.
 
         :meth:`process_spider_exception` should return either ``None`` or an
-        iterable of :class:`~scrapy.http.Response`, dict or
+        iterable of :class:`~scrapy.http.Request`, dict or
         :class:`~scrapy.item.Item` objects.
 
         If it returns ``None``, Scrapy will continue processing this exception,


lopuhin:
This fix looks consistent with the docs, ``process_spider_exception`` docs say

> If it returns an iterable the :meth:`process_spider_output` pipeline kicks in, and no other :meth:`process_spider_exception` will be called.

and ``process_spider_output`` docs say

> :meth:`process_spider_output` must return an iterable of  :class:`~scrapy.http.Request`, dict or :class:`~scrapy.item.Item` objects.
> :param result: the result returned by the spider :type result: an iterable of :class:`~scrapy.http.Request`, dict or :class:`~scrapy.item.Item` objects

Although this is not checked in the [code](https://github.com/scrapy/scrapy/blob/129421c7e31b89b9b0f9c5f7d8ae59e47df36091/scrapy/core/spidermw.py#L51-L68)

grammy-jiang:
Hi, @lopuhin 

I have written several middlewares just about this method, `process_spider_exception`. They all work well but I have not checked the source code. The documents could be consistent between the classes of the spider and the spider middleware.

lopuhin:
@grammy-jiang I agree - I think your fix is good and should be merged, I was just wondering what will happen if a response is returned.

grammy-jiang:
@lopuhin Emmm... good point.

In my view, the method `process_spider_output` will be called when an iterable is returned from `process_spider_exception`, which means the thing will happen just like a response yielded from a spider.  

So your question becomes what happens when a response yielded from a spider? I have no idea, no relative experience before...

lopuhin:
Right, returning a response from a spider looks very strange indeed!

grammy-jiang:
@lopuhin I have checked the class `spidermw.py` and I have not seen any result type assert here. The response type will go through spider middleware without any exception.

So will there any type asset in engine or scheduler? I am not good enough at twisted, would you help to check? Thanks.

PR #3210, opened 2018-04-10
Automatic port selection for servicies in unit tests

whalebot-helmsman:
On a busy host usual ports are already occupied by different services. Tests just fails if some base server can not listen on port.
This pull-request add ability searching of free port for `mockserver` and `proxyserver`

### DIFF
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -211,11 +211,20 @@ def ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.c
 
 
 if __name__ == "__main__":
+    http_port = 8998
+    https_port = 8999
+
+    if len(sys.argv) > 1:
+        http_port = int(sys.argv[1])
+
+    if len(sys.argv) > 2:
+        https_port = int(sys.argv[2])
+
     root = Root()
     factory = Site(root)
-    httpPort = reactor.listenTCP(8998, factory)
+    httpPort = reactor.listenTCP(http_port, factory)
     contextFactory = ssl_context_factory()
-    httpsPort = reactor.listenSSL(8999, factory, contextFactory)
+    httpsPort = reactor.listenSSL(https_port, factory, contextFactory)
 
     def print_listening():
         httpHost = httpPort.getHost()
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -201,7 +201,7 @@ class MockServer():
     HTTPS = 'mock_https'
 
     @classmethod
-    def from_mock(cls, path, is_secure=False):
+    def url(cls, path, is_secure=False):
         key = cls.HTTP
         if is_secure:
             key = cls.HTTPS
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -194,8 +194,11 @@ class MockServer():
     HTTPS = 'mock_https'
 
     @classmethod
-    def from_mock(cls, path):
-        return os.environ[cls.HTTP] + path
+    def from_mock(cls, path, is_secure=False):
+        key = cls.HTTP
+        if is_secure:
+            key = cls.HTTPS
+        return os.environ[key] + path
 
     def __enter__(self):
         from scrapy.utils.test import get_testenv
@@ -214,7 +217,7 @@ class MockServer():
 
             self._oldenv = os.environ.copy()
             os.environ[self.HTTP] = 'http://localhost:%s' % (http_port, )
-            os.environ[self.HTTPS] = 'http://localhost:%s' % (https_port, )
+            os.environ[self.HTTPS] = 'https://localhost:%s' % (https_port, )
 
             return self
 
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -190,6 +190,9 @@ class Root(Resource):
 
 class MockServer():
 
+    HTTP = 'mock_http'
+    HTTPS = 'mock_https'
+
     def __enter__(self):
         from scrapy.utils.test import get_testenv
 
@@ -203,14 +206,19 @@ class MockServer():
 
             if self.proc.poll() is not None:
                 http_port = https_port
+                continue
 
             self._oldenv = os.environ.copy()
+            os.environ[self.HTTP] = 'http://localhost:%s' % (http_port, )
+            os.environ[self.HTTPS] = 'http://localhost:%s' % (https_port, )
+
             return self
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.proc.kill()
         self.proc.wait()
         time.sleep(0.2)
+        os.environ = self._oldenv
 
 
 def ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.crt'):
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -198,7 +198,6 @@ class MockServer():
         http_address = self.proc.stdout.readline().strip().decode('ascii')
         https_address = self.proc.stdout.readline().strip().decode('ascii')
 
-        self._oldenv = os.environ.copy()
         self.http_address = http_address
         self.https_address = https_address
 
@@ -208,7 +207,6 @@ class MockServer():
         self.proc.kill()
         self.proc.wait()
         time.sleep(0.2)
-        os.environ = self._oldenv
 
     def address(self):
         return MockServerAddress(self.http_address, self.https_address)
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -190,16 +190,6 @@ class Root(Resource):
 
 class MockServer():
 
-    HTTP = 'mock_http'
-    HTTPS = 'mock_https'
-
-    @classmethod
-    def url(cls, path, is_secure=False):
-        key = cls.HTTP
-        if is_secure:
-            key = cls.HTTPS
-        return os.environ[key] + path
-
     def __enter__(self):
         from scrapy.utils.test import get_testenv
 
@@ -209,8 +199,8 @@ class MockServer():
         https_address = self.proc.stdout.readline().strip().decode('ascii')
 
         self._oldenv = os.environ.copy()
-        os.environ[self.HTTP] = http_address
-        os.environ[self.HTTPS] = https_address
+        self.http_address = http_address
+        self.https_address = https_address
 
         return self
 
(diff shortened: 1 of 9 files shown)


whalebot-helmsman:
Problem with ephemeral port - there is no reliable way to return it from mockserver to process running unit tests. I didn't count piping as a reliable way.
And about splash tests they do not always work. In my case it obtained port for external IP of host. For security reasons almost all ports are closed for external IP. I get no responses to splash in docker container.

tests/mockserver.py
```diff
@@ -17,6 +18,12 @@
 from scrapy.utils.python import to_bytes, to_unicode
 
 
+def get_ephemeral_port():
+    s = socket.socket()
+    s.bind(("", 0))
```
kmike: Hm, so you said there was an issue because Splash tests are binding port for 0.0.0.0 (the same as ""?), and it was problematic? Should we try finding a port by binding to localhost only?

tests/mockserver.py
```diff
@@ -190,17 +197,37 @@ def render(self, request):
...
+        http_port = str(get_ephemeral_port())
+        https_port = str(get_ephemeral_port())
+
+        self.proc = Popen([sys.executable, '-u', '-m', 'tests.mockserver', http_port, https_port],
                           stdout=PIPE, env=get_testenv())
         self.proc.stdout.readline()
+
+        self._oldenv = os.environ.copy()
+        os.environ[self.HTTP] = 'http://localhost:%s' % (http_port, )
+        os.environ[self.HTTPS] = 'https://localhost:%s' % (https_port, )
```
kmike: I think one of the goals should be to allow running Scrapy tests in parallel, including parallel in the same process. For this to work ports shouldn't be global - they should be attributes of the MockServer instance. I know it can be a larger change :(

tests/mockserver.py
```diff
@@ -190,17 +197,37 @@ def render(self, request):
 
 class MockServer():
 
+    HTTP = 'mock_http'
+    HTTPS = 'mock_https'
+
+    @classmethod
+    def from_mock(cls, path, is_secure=False):
```
kmike: What do you think about renaming it to `get_url`?

tests/mockserver.py
kmike: Hm, but using a single MockServer can be fine for Scrapy purposes, so maybe having host/port global is fine.

tests/mockserver.py
whalebot-helmsman: Not a problem for me

tests/mockserver.py
```diff
@@ -190,17 +197,37 @@ def render(self, request):
...
+        os.environ[self.HTTP] = 'http://localhost:%s' % (http_port, )
+        os.environ[self.HTTPS] = 'https://localhost:%s' % (https_port, )
+
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.proc.kill()
         self.proc.wait()
         time.sleep(0.2)
+        os.environ = self._oldenv
```
kmike: does it work, i.e. does it replace environment of a proces, or does it just replace attribute of the os module? If that's the former, then why do we need to use env variables, instead of using global variables (class attributes?)

whalebot-helmsman:
Considering all pypy tests fail. `get_ephemeral_port` just binds one port and releases at function end

tests/mockserver.py
whalebot-helmsman: copied this solution from ProxyConnectTestCase.tearDown

whalebot-helmsman: It is a variable, not a module https://stackoverflow.com/a/43487762/8640733

tests/mockserver.py
whalebot-helmsman: To make port as variable of object you need to pass it to spider classes. And there is no clean way to do this. `scrapy.crawler.CrawlerRunner.create_crawler` does not accept arguments. There are lot of such places.

whalebot-helmsman:
I implemented obtaining random port using kernel mechanisms for mockserver and proxyserver. Mockserver messages back addresses using pipeline. Usage of high-level mechanism for IPC(e.g `multiprocessing.Queue`) is blocked by incompatibility between `multiprocessig` and `twisted`.

Not storing host/port for mockserver in process Environment requires reimplementing all current spiders and their initiation mechanisms. Having one mockserver is not a problem, because they all are identical.

whalebot-helmsman:
Added per-test mockserver address passing to spiders. Now we have single mockserver per test case

tests/mockserver.py
```diff
@@ -192,15 +192,39 @@ class MockServer():
...
     def __enter__(self):
         from scrapy.utils.test import get_testenv
+
         self.proc = Popen([sys.executable, '-u', '-m', 'tests.mockserver'],
                           stdout=PIPE, env=get_testenv())
-        self.proc.stdout.readline()
+        http_address = self.proc.stdout.readline().strip().decode('ascii')
+        https_address = self.proc.stdout.readline().strip().decode('ascii')
+
+        self._oldenv = os.environ.copy()
```
kmike: It looks unnecessary now.

tests/test_proxy_connect.py
```diff
@@ -17,31 +18,36 @@
...
         self.mockserver.__enter__()
         self._oldenv = os.environ.copy()
-        self._proxy = HTTPSProxy(8888)
+
+        self._proxy = HTTPSProxy()
         self._proxy.start()
+
         # Wait for the proxy to start.
         time.sleep(1.0)
-        os.environ['http_proxy'] = 'http://scrapy:scrapy@localhost:8888'
```
kmike: what's the reason http_proxy env variable is deleted?

tests/test_proxy_connect.py
```diff
@@ -94,10 +103,14 @@ def test_https_tunnel_without_leak_proxy_authorization_header(self):
 
     @defer.inlineCallbacks
     def test_https_noconnect_auth_error(self):
-        os.environ['https_proxy'] = 'http://wrong:wronger@localhost:8888?noconnect'
+        proxy = os.environ['https_proxy']
+        bad_auth_proxy = list(urlsplit(proxy))
+        bad_auth_proxy[1] = bad_auth_proxy[1].replace('scrapy:scrapy@', 'wrong:wronger@')
```
kmike: I think it makes sense to create a small utility function, to avoid copy-paste

tests/test_proxy_connect.py
whalebot-helmsman: No reason, just a mistake

whalebot-helmsman:
@kmike after merging https://github.com/scrapy/scrapy/pull/3263 test are good

tests/test_proxy_connect.py
```diff
@@ -17,31 +18,42 @@
...
-        self._proxy = HTTPSProxy(8888)
+
+        self._proxy = HTTPSProxy()
         self._proxy.start()
+
         # Wait for the proxy to start.
         time.sleep(1.0)
-        os.environ['http_proxy'] = 'http://scrapy:scrapy@localhost:8888'
-        os.environ['https_proxy'] = 'http://scrapy:scrapy@localhost:8888'
+        os.environ['https_proxy'] = self._proxy.http_address()
```
lopuhin: should it be ``self._proxy.https_address()`` instead?

tests/test_proxy_connect.py
```diff
@@ -52,17 +64,17 @@ def tearDown(self):
     def test_https_connect_tunnel(self):
         crawler = get_crawler(SimpleSpider)
         with LogCapture() as l:
-            yield crawler.crawl("https://localhost:8999/status?n=200")
+            yield crawler.crawl(self.mockserver.url("/status?n=200"))
```
lopuhin: missing ``is_secure=True``?

tests/test_proxy_connect.py
```diff
@@ -52,17 +64,17 @@ def tearDown(self):
...
 
     @defer.inlineCallbacks
     def test_https_noconnect(self):
-        os.environ['https_proxy'] = 'http://scrapy:scrapy@localhost:8888?noconnect'
+        proxy = os.environ['https_proxy']
+        os.environ['https_proxy'] = proxy + '?noconnect'
         crawler = get_crawler(SimpleSpider)
         with LogCapture() as l:
-            yield crawler.crawl("https://localhost:8999/status?n=200")
+            yield crawler.crawl(self.mockserver.url("/status?n=200"))
```
lopuhin: missing ``is_secure=True``?

tests/test_proxy_connect.py
```diff
@@ -73,18 +85,17 @@ def test_https_connect_tunnel_error(self):
...
+            yield crawler.crawl(self.mockserver.url("/status?n=200", is_secure=True))
         # The proxy returns a 407 error code but it does not reach the client;
         # he just sees a TunnelError.
         self._assert_got_tunnel_error(l)
-        os.environ['https_proxy'] = 'http://scrapy:scrapy@localhost:8888'
 
     @defer.inlineCallbacks
     def test_https_tunnel_without_leak_proxy_authorization_header(self):
-        request = Request("https://localhost:8999/echo")
+        request = Request(self.mockserver.url("/echo"))
```
lopuhin: missing ``is_secure=True``?

tests/test_proxy_connect.py
whalebot-helmsman: We have two type of protocols - protocol we talk with a crawled server and protocol we talk with a proxy. For different protocols for crawled server(http and https) we use same proxy protocol(http).

lopuhin:
@whalebot-helmsman many thanks for fixing this and thanks @kmike for review! 👍

lopuhin:
For reference, master build failed on pypy3 for a seemingly unrelated reason, I restarted it and it worked. 
I think the errors were at teardown of ``FeedExportTest.test_export_indentation`` and ``FeedExportTest.test_export_feed_export_fields``: ``twisted.internet.error.AlreadyCancelled: Tried to cancel an already-cancelled event.``. Overall stats: ``== 2 failed, 1660 passed, 11 skipped, 13 xfailed, 2 error in 1343.60 seconds ===`` (I downloaded original log too). I don't see it mentioned in the issues, if this repeats I'll create a separate issue.

whalebot-helmsman:
@lopuhin there is incorrect interaction between scrapy, twisted(or may be https://github.com/scrapy/scrapy/pull/3238

PR #3220, opened 2018-04-16
Update tutorial.rst

radostyle:
Request should have dont_filter=True for spider example.


wRAR:
Why do you think this is needed? True, it follows the default start_urls handler but I don't think it's needed by itself.

Issue #2735, opened 2017-05-10
Functions don't work as item processors in Python 3


wRAR:
The documentation says "processors are just callable objects, which are called with the data to be parsed, and return a parsed value. So you can use any function as input or output processor" and this works in Python 2:
```python
def proc(i):
    return i

class FooItemLoader(ItemLoader):
    foo_in = proc
```
But `FooItemLoader.foo_in` is a bound method in Python 3 so it receives a `self` argument and not just the field value.

redapple:
@wRAR , the following test fails for me locally both in Python 2.7 and Python 3.6:

```
$ git diff origin/master tests/test_loader.py
diff --git a/tests/test_loader.py b/tests/test_loader.py
index 2693a18..05b9f77 100644
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -282,6 +282,18 @@ class BasicItemLoaderTest(unittest.TestCase):
         il.add_value('name', [u'mar', u'ta'])
         self.assertEqual(il.get_output_value('name'), u'Mar Ta')
 
+    def test_output_processor_using_function2(self):
+
+        def proc(i):
+            return i
+
+        class TakeFirstItemLoader(TestItemLoader):
+            name_out = proc
+
+        il = TakeFirstItemLoader()
+        il.add_value('name', [u'mar', u'ta'])
+        self.assertEqual(il.get_output_value('name'), u'Mar Ta')
+
```
```
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.12+, pytest-2.9.2, py-1.4.33, pluggy-0.3.1
rootdir: /home/paul/src/scrapy, inifile: pytest.ini
plugins: twisted-1.5, cov-2.2.1
collected 57 items 

tests/test_loader.py ..........................F..............................

================================================================= FAILURES ==================================================================
_________________________________________ BasicItemLoaderTest.test_output_processor_using_function2 _________________________________________

self = <tests.test_loader.BasicItemLoaderTest testMethod=test_output_processor_using_function2>

    def test_output_processor_using_function2(self):
    
        def proc(i):
            return i
    
        class TakeFirstItemLoader(TestItemLoader):
            name_out = proc
    
        il = TakeFirstItemLoader()
        il.add_value('name', [u'mar', u'ta'])
>       self.assertEqual(il.get_output_value('name'), u'Mar Ta')

/home/paul/src/scrapy/tests/test_loader.py:295: 
[...]

PR #3224, opened 2018-04-19
[MRG+1] Improve document about functions as processors

lucywang000:
I came across this earlier today, and after quite some (result-less) debugging, google led me to #2735 . Hopefully this improvement can help others avoid this pitfall.

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -136,6 +136,20 @@ accept one (and only one) positional argument, which will be an iterator.
    containing the collected values (for that field). The result of the output
    processors is the value that will be finally assigned to the item.
 
+If you want to use a plain function as a processor, make sure it receives
+``self`` as the first argument::
+
+    def lowercase_processor(self, values):
+        for v in values:
+            yield v.lower()
+
+    class MyItemLoader(ItemLoader):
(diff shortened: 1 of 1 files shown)


kmike:
Looks good, thanks @lucywang000!

I think in a long term we should support plain functions somehow. There was also a discussion about moving item loaders to a separate Python package.

lopuhin:
Thanks @lucywang000 !

lucywang000:
Thanks for the review @kmike @lopuhin !

IAlwaysBeCoding:
@kmike I think we should seriously have a third-party library that would have several processors that are highly needed. I've came up with several that I use on my own Scrapy projects. 

Some processors that I have coded myself are:

* Processor that will output a unique value for true and/or false if an element is found using an xpath/css. * Processor that will choose an n value from the extracted list, it's basically like a TakeFirst() but you can decide what n's index you want.
* A TakeLast() processor as well.
* Processor that will replace values
* Processor that will split the output into values from a given delimeter.

and many more...

You can't really fully utlize the `ItemLoader` if you don't have a library of Processors, the default Processors that come with `Scrapy` left me no choice than to cook my own.

Issue #3211, opened 2018-04-11
Incorrect return code for spiders with exception raising in __init__ or from_crawler


whalebot-helmsman:
Sometimes you raise an exception from spider `__init__` or `from_crawler` methods. Exit code of `scrapy runspider discovery.py` process is 0 in such case. This behaviour prevents correct management of spider runs by any supervisor.  

Should scrapy report such problems using non-zero exit code?

E.g. for spider
```
import scrapy

class DiscoverySpider(scrapy.Spider):
    name = 'discovery'

    @classmethod
    def from_crawler(cls, crawler, *args, **kwargs):
        raise ValueError('Exception in from_crawler method')
```

output is 

```
[]$  scrapy runspider discovery.py 
2018-04-11 12:51:22 [scrapy.utils.log] INFO: Scrapy 1.5.0 started (bot: scrapybot)
2018-04-11 12:51:22 [scrapy.utils.log] INFO: Versions: lxml 4.0.0.0, libxml2 2.9.5, cssselect 1.0.1, parsel 1.2.0, w3lib 1.18.0, Twisted 17.9.0, Python 3.4.3 (default, Nov 17 2016, 01:08:31) - [GCC 4.8.4], pyOpenSSL 17.3.0 (OpenSSL 1.1.0f  25 May 2017), cryptography 2.0.3, Platform Linux-4.4.0-96-generic-x86_64-with-Ubuntu-14.04-trusty
2018-04-11 12:51:22 [scrapy.crawler] INFO: Overridden settings: {'SPIDER_LOADER_WARN_ONLY': True}
2018-04-11 12:51:22 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.logstats.LogStats',
 'scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.corestats.CoreStats',
 'scrapy.extensions.memusage.MemoryUsage']
Unhandled error in Deferred:
2018-04-11 12:51:22 [twisted] CRITICAL: Unhandled error in Deferred:

2018-04-11 12:51:22 [twisted] CRITICAL: 
Traceback (most recent call last):
  File "python3.4/site-packages/twisted/internet/defer.py", line 1386, in _inlineCallbacks
    result = g.send(result)
  File "python3.4/site-packages/scrapy/crawler.py", line 79, in crawl
    self.spider = self._create_spider(*args, **kwargs)
  File "python3.4/site-packages/scrapy/crawler.py", line 102, in _create_spider
[...]

PR #3226, opened 2018-04-20
[MRG+1] Addressing issue #3211

whalebot-helmsman:
Return non-zero exit code from commands in case if we have error in spiders constructor


scrapy/crawler.py
```diff
@@ -221,6 +227,11 @@ def join(self):
         while self._active:
             yield defer.DeferredList(self._active)
 
+    def is_crawlers_has_spider(self):
+        return reduce(lambda x,y: x and y,
+                      self.crawlers_has_spiders,
+                      True)
```
kmike: `any(self.crawlers_has_spiders)`?

kmike: as @VMRuiz noted, it is actually `all(...)`, not `any(...)`, which is an additional argument to change it, as the code is not obvious :P

whalebot-helmsman:
Test failed due to new twisted release, fix in twisted trunk https://github.com/twisted/twisted/pull/1008 but not released to pypy

tests/test_commands.py
```diff
@@ -1,3 +1,5 @@
+import pdb
+import inspect
```
kmike: these imports are unneeded

tests/test_crawler.py
```diff
@@ -11,7 +11,10 @@
 from scrapy.utils.log import configure_logging, get_scrapy_root_handler
 from scrapy.utils.spider import DefaultSpider
 from scrapy.utils.misc import load_object
+from scrapy.utils.test import get_crawler
 from scrapy.extensions.throttle import AutoThrottle
+from twisted.internet import defer
+import twisted.trial.unittest
```
kmike: as per pep8, twisted imports should go in a separate group, between stdlib imports and scrapy imports

tests/test_commands.py
```diff
@@ -220,6 +222,16 @@ def test_runspider(self):
         self.assertIn("INFO: Closing spider (finished)", log)
         self.assertIn("INFO: Spider closed (finished)", log)
 
+    def test_run_fail_spider(self):
+        proc = self.runspider("import scrapy\n" + inspect.getsource(ExceptionSpider))
```
kmike: nice trick!

kmike:
Looks good to me, thanks @whalebot-helmsman! I'd remove `_is` from method names, but it is just a style preference. +1 to merge.

scrapy/crawler.py
```diff
@@ -221,6 +226,9 @@ def join(self):
         while self._active:
             yield defer.DeferredList(self._active)
 
+    def _is_spider_created_for_every_crawler(self):
+        return all(self.crawlers_has_spiders)
```
dangra: I'd like to propose the removal of all this methods and use a single boolean flag that summarizes the outcome of bootstraping the spider. 

The public attribute name would be `bootstrap_failed` and we set it at `_crawl / _done()` like:
```python
    self.bootstrap_failed |= not getattr(crawler, 'spider', None)
```
this public attribute replaces `_is_spider_created_for_every_crawler()` method and `crawlers_has_spiders` atttribute, but also remove the need for private `_is_spider_created()` method.   

then the check in `runspider.py` and `crawl.py` looks like:

```python
if not self.crawler_process.bootstrap_failed:
    self.exitcode = 1
```

kmike:
Thanks @whalebot-helmsman and @dangra!

dangra:
thanks @whalebot-helmsman and @kmike

dangra:
I restarted the failed build to be sure it was a temporal failure.

PR #3230, opened 2018-04-22
Add __setitem__, __getitem__ and __iter__ to class StatsCollector

leonardfrank:
In the issue #2746 , new methods are required to add to class `StatsCollector`. And this requirement is also supported by @kmike and @redapple .

As answering the call from them, now I have written the methods of `__setitem__`, `__getitem` and `__iter__` with a fully unittest, and also the related narrative in the document Stats Collection.

With these new methods, class `StatsCollector` now can be used more easily to set, get values, and also iterate.

### DIFF
--- a/scrapy/statscollectors.py
+++ b/scrapy/statscollectors.py
@@ -1,8 +1,8 @@
 """
 Scrapy extension for collecting scraping stats
 """
-import pprint
 import logging
+import pprint
 
 logger = logging.getLogger(__name__)
 
@@ -50,6 +50,15 @@ class StatsCollector(object):
     def _persist_stats(self, stats, spider):
         pass
 
+    def __getitem__(self, key):
+        return self._stats.get(key)
+
+    def __setitem__(self, key, value):
+        self._stats[key] = value
+
+    def __iter__(self):
+        return iter(self._stats)
+
 
 class MemoryStatsCollector(StatsCollector):
 
@@ -80,5 +89,3 @@ class DummyStatsCollector(StatsCollector):
 
     def min_value(self, key, value, spider=None):
         pass
-
-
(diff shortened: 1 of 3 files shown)


scrapy/statscollectors.py
```diff
@@ -50,6 +50,15 @@ def close_spider(self, spider, reason):
     def _persist_stats(self, stats, spider):
         pass
 
+    def __getitem__(self, key):
+        return self._stats.get(key)
```
kmike: I think it makes more sense to raise an exception if a key is not found, like dicts do.

leonardfrank: Hi, @kmike 

Thanks for your suggestion, I will add that feature to `__getitem__`.

kmike: > Now, what should be expected when a key is missing? I think to return 0 and create the missing key at the same time could be more consistent with previous behaviors (a debug level logging message could be added here).

I agree that returning 0 and creating a key is convenient in many cases, but if we're implementing a dict-like interface, we better follow dict semantics. Users still have all existing methods available, and they can use them if these methods are a better fit. Implementing `setdefault` method could be better than changing `__getitem__` semantics, as one can write `value = stats.setdefault('my/stats', 0)` to return 0 and create a new key if it is not present, just like with regular Python dicts. It also allows to use non-numbers as default values (not all stats values are numbers).

> Also in this class StatsCollector, when I am going through the source code, I find the behavior of this class is like a dict - why not just inherit the class UserDict in the module collections, and add the necessary methods?

UserDict is removed in Python 3, but inheriting from MutableMapping is a good idea.

AdrianAtZyte:
Shall we close this?

wRAR:
No strong opinion, this may be useful but doesn't make much difference to the API.

PR #3231, opened 2018-04-23
[MRG+2] Added: Allowing optional arguments for `scrapy.http.cookies.CookieJar.clear`

starrify:
So it matches the signature of the underlying [`http.cookiejar.CookieJar.clear`](https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.CookieJar.clear) object.

### DIFF
--- a/scrapy/http/cookies.py
+++ b/scrapy/http/cookies.py
@@ -58,8 +58,8 @@ class CookieJar(object):
     def clear_session_cookies(self, *args, **kwargs):
         return self.jar.clear_session_cookies(*args, **kwargs)
 
-    def clear(self):
-        return self.jar.clear()
+    def clear(self, domain=None, path=None, name=None):
+        return self.jar.clear(domain, path, name)
 
     def __iter__(self):
         return iter(self.jar)


scrapy/http/cookies.py
```diff
@@ -58,8 +58,8 @@ def _cookies(self):
     def clear_session_cookies(self, *args, **kwargs):
         return self.jar.clear_session_cookies(*args, **kwargs)
 
-    def clear(self):
-        return self.jar.clear()
+    def clear(self, domain=None, path=None, name=None):
```
kmike: What do you think about using args/kwargs, like in `clear_session_cookies`? This way we won't have to change this method if Python decide to change signature for some reason.

kmike:
A good idea!

scrapy/http/cookies.py
starrify: Currently there're two types of function signatures here:
- Using args/kwargs, like `clear_session_cookies`.
- Using the exact signature of the underlying CookieJar object, like `set_cookie` and `set_policy`.

For now I don't think any type has a significant advantage to the other. Thus I'm okay to use either type :) Please let me know if you want me to make the change.

kmike: Fair enough, I'm fine with merging it as-is.

PR #1081, opened 2015-03-18
Allow spiders to return dicts. 

kmike:
A PR to fix GH-1064.

Docs are missing.

Issue #616, opened 2014-02-26
ItemLoader should work with copy of item passed as an argument


chekunkov:
Sometimes it is needed to parse several variants of item from single page, where only a couple of fields differ. Obvious way to do this is to create base item with all common fields collected and then create new items (for example in a loop) using new ItemLoader instance with base item passed to it.

The problem I'm facing now is that ItemLoader.load_item() modifies base item - which is counterintuitive and can result in weird behavior (for example if variant items can have different fields - after those field were added to base item - they would appear in all loaded items).

```
In [5]: item = Item()

In [7]: item['url'] = 'foo'

In [8]: item
Out[8]: {'url': 'foo'}

In [9]: l = ItemLoader(item)

In [12]: l.add_value('category', 'bar')

In [13]: item
Out[13]: {'url': 'foo'}

In [14]: item_copy = l.load_item()

In [15]: item_copy
Out[15]: {'category': 'bar', 'url': 'foo'}

In [16]: item
Out[16]: {'category': 'bar', 'url': 'foo'}

In [17]: id(item)
Out[17]: 49468304

In [18]: id(item_copy)
Out[18]: 49468304
```

Now I'm using workaround like this to suppress such behavior:

```
loader = ItemLoader(selector=sel)
...
item = loader.load_item()
for variant in variants:
    item_copy = item.copy()
    loader = ItemLoader(item_copy)
    ...
    yield loader.load_item()
```

What do you think about using item copy inside ItemLoader by default?

redapple:
@chekunkov , I'm closing this issue as you seem(ed) ok with it.
@kmike , @chekunkov , feel free to object and re-open.

LanetheGreat:
Sorry to comment on a long closed issue, just noting a possible solution for others to use that worked for me if they haven't come up with their own solution or if this could be added to the next version of Scrapy (this was tested in v1.7.3). But my project had the same issue while scraping a website that had variations on single page responses.

I have a custom `ProductLoader` loader that would pull 1 or more items from a page with products as custom `ProductItem` items, the description of a product was in 1 part of the page and applied to each variation so it needed duplicated while fields like color, UPC, etc. where listed differently for each. It'd parse each line in the description and the output formatter would then join them together as a single HTML string formatted as paragraphs (also converted \<br\> to \<p\> to make them look more consistent):
```html
<p>Paragraph 1</p>
<p>Paragraph 2</p>
<p>Paragraph 3</p>
```
[...]

PR #3244, opened 2018-05-03
Allow spiders to also return/yield ItemLoader.

vincent-ferotin:
Behind, Scrapy will convert ItemLoader to Item by calling
:meth:`ItemLoader.load_item`.


whalebot-helmsman:
@vincent-ferotin Problem with unit-tests was fixed in master. Can you merge master to this PR?

vincent-ferotin:
@whalebot-helmsman Here it is, I hope so? (not enough familiar/confident with git/GitHub to be sure, please let tell me if it is not ok for you ^^;)

whalebot-helmsman:
Thanks, @vincent-ferotin mentioned problems with twisted is gone. I don't know the reason why other tests are in failed state.

kmike:
Hi @vincent-ferotin! I can see how this feature can save some typing, and API improvements are good to have. However, currently I'm -0 to add such feature because: 

* there was an idea to move ItemLoaders out of Scrapy;
* it complicates API of a spider; custom middlewares would need to be modified in order to support a new possible output type (?).

The implementation also needs work: 

* if I'm not mistaken, currently it can lead to subtle bugs (see https://github.com/scrapy/scrapy/issues/616)
* there are no tests;
* documentation is not updated (you can probably check https://github.com/scrapy/scrapy/pull/1081 for suggestions of doc locations to be updated).

Gallaecio:
Closing as, even if implemented, https://github.com/scrapy/itemadapter would be the right place now.

PR #3251, opened 2018-05-09
Fix OffsiteMiddleware tests

rpkilby:
The tests for `OffsiteMiddleware` are not actually functioning as expected. `TestOffsiteMiddleware3` and `TestOffsiteMiddleware4` do not correctly provided the spider args, so default to the spider created in `TestOffsiteMiddleware1`, effectively testing nothing.

I'm submitting the test changes first to demonstrate the failures.

I've refactored the offsite tests to help ensure these failures don't happen in the future. Instead of unintentionally succeeding, the `setUp` would fail. Additionally, factored away the testcase inheritance, which is a little confusing to follow. This can be reverted if preferred.

### DIFF
--- a/scrapy/spidermiddlewares/offsite.py
+++ b/scrapy/spidermiddlewares/offsite.py
@@ -52,12 +52,16 @@ class OffsiteMiddleware(object):
         """Override this method to implement a different offsite policy"""
         allowed_domains = getattr(spider, 'allowed_domains', None)
         if not allowed_domains:
-            return re.compile('') # allow all by default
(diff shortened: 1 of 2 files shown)


whalebot-helmsman:
@rpkilby tests fails are related to new version of twisted https://github.com/scrapy/scrapy/pull/3226#issuecomment-387428744

whalebot-helmsman:
@rpkilby Problem with unit-tests was fixed in master. Can you merge master to this PR?

rpkilby:
Updated.

PR #3253, opened 2018-05-09
[MRG+1] Update depth middleware stats (fixes #3245)

### LINKED ISSUE #3245: DEPTH_STATS has no effect
It looks like this flag stopped functioning in #99. `crawler.stats` are always passed to the middleware, and are always collected.

Additionally, it looks like this middleware would error if stats weren't provided, as the initial depth value would otherwise never be set. 

https://github.com/scrapy/scrapy/blob/c4f096d3a55def4c986174d50ff8b84c8f00503d/scrapy/spidermiddlewares/depth.py#L53-L56

rpkilby:
- Update the docs & default settings to reference `DEPTH_STATS_VERBOSE` instead of `DEPTH_STATS`, which is not currently used. 
- Remove unneeded branches for `self.stats` checks, as stats are always provided (no way to *not* provide a stats object).

(fixes #3245)


whalebot-helmsman:
@rpkilby Problem with unit-tests was fixed in master. Can you merge master to this PR?

rpkilby:
Updated

kmike:
Thanks @rpkilby, it looks good! This fixes https://github.com/scrapy/scrapy/issues/3245, right?

PR #3254, opened 2018-05-10
Fixes #3246:OffsiteMiddleware silently filtering (invalid) URLs

### LINKED ISSUE #3246: OffsiteMiddleware silently filtering (invalid) URLs
We've recently run into a weird situation where the requests were created reading the URL values from a sitemap and as a result they contained leading and trailing newline characters (`\nhttp://example.com\n`).

Consider this base example:
```
class ToscrapeSpider(scrapy.Spider):
    name = 'toscrape'
    allowed_domains = ['toscrape.com']
    start_urls = ['http://books.toscrape.com/']

    def parse(self, response):
        yield scrapy.Request(
            'http://quotes.toscrape.com/',
            callback=self.parse2,
        )

    def parse2(self, response):
        pass
```
which sends 2 requests as expected:
```
2018-05-04 09:17:49 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://books.toscrape.com/> (referer: None)
2018-05-04 09:17:49 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/> (referer: http://books.toscrape.com/)
2018-05-04 09:17:49 [scrapy.core.engine] INFO: Closing spider (finished)
``` 

And now check what happens when you replace `http://quotes.toscrape.com/` with `\nhttp://quotes.toscrape.com/`:
``` 
class ToscrapeSpider(scrapy.Spider):
    name = 'toscrape'
    allowed_domains = ['toscrape.com']
    start_urls = ['http://books.toscrape.com/']

    def parse(self, response):
        yield scrapy.Request(
            '\nhttp://quotes.toscrape.com/',
            callback=self.parse2,
        )

    def parse2(self, response):
        pass
``` 
Results:
``` 
2018-05-04 09:19:49 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://books.toscrape.com/> (referer: None)
2018-05-04 09:19:49 [scrapy.core.engine] INFO: Closing spider (finished)
``` 
So Scrapy (the OffsiteMiddleware to be more specific) is silently filtering the second request and the only way to know this is by checking the stats: `'offsite/filtered': 1`. In this situation I would expect Scrapy to show some kind of warning (or even an error) when creating the Request or when processing the request in the OffsiteMiddleware.

For reference: https://github.com/scrapy/scrapy/blob/c4f096d3a55def4c986174d50ff8b84c8f00503d/scrapy/spidermiddlewares/offsite.py#L36
`domain` is None in this case and the OffsiteMiddleware doesn't add any messages to the log. 
I think the OffsiteMiddleware should log the filtered requests even if it can't properly parse the domain.

IAlwaysBeCoding:
Fixes #3246 

`urlparse_cached()` strips out any \n, \t from the domain's url before passing it to `urlparse()`. This will make sure we don't get empty hostnames  from a domain like this : `\nhttp://quotes.toscrape.com/`.

### DIFF
--- a/scrapy/utils/httpobj.py
+++ b/scrapy/utils/httpobj.py
@@ -10,5 +10,7 @@ def urlparse_cached(request_or_response):
     Request or Response object
     """
     if request_or_response not in _urlparse_cache:
-        _urlparse_cache[request_or_response] = urlparse(request_or_response.url)
+        #Make sure we clean out any domains that might have \n, \t in them.
+        url = request_or_response.url.strip()
+        _urlparse_cache[request_or_response] = urlparse(url)
     return _urlparse_cache[request_or_response]
--- a/tests/test_utils_httpobj.py
+++ b/tests/test_utils_httpobj.py
@@ -21,6 +21,29 @@ class HttpobjUtilsTest(unittest.TestCase):
         assert req1a is not req2
         assert req1a is not req2
 
+    def test_urlparse_cached_strip():
+
+        #Test out that urlparse_cached strips out any \n
+        # or \t from its url
+
+        normal_url = "http://quotes.toscrape.com/
+        dirty_url = "\nhttp://quotes.toscrape.com/\n"
+
+        request1 = Request(normal_url)
+        request2 = Request(dirty_url)
+        req1a = urlparse_cached(request1)
+        req1b = urlparse_cached(request1)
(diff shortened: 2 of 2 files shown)


wRAR:
The issue was fixed in w3lib.

PR #3255, opened 2018-05-10
Allow configuring url escaping on Request objects.

### LINKED ISSUE #833: Prevent URL encoding option

dyeray:
Provide a boolean parameter "escape_url" on the Request constructor that indicates whether the url will be escaped or left as it is. This parameter is saved on the request object so a new request object can be built from the previous one using the replace method.

Fixes #833

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -126,6 +126,9 @@ Request objects
     :param flags:  Flags sent to the request, can be used for logging or similar purposes.
     :type flags: list
 
+    :param escape_url: Whether the parameter ``url`` should be escaped or left as-is.
+    :type escape_url: bool
+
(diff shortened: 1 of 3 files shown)


dyeray:
Tested on py27 and py36 (twisted 17.9.0).

whalebot-helmsman:
@dyeray Problem with unit-tests was fixed in master. Can you merge master to this PR?

dyeray:
Done.

dyeray:
Implements #833

PR #3256, opened 2018-05-11
[MRG+1] add item_error to be catchable

chainly:
add item_error to be catchable (ie. by `extension` with `signal`).

### DIFF
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -135,6 +135,29 @@ item_dropped
         to be dropped
     :type exception: :exc:`~scrapy.exceptions.DropItem` exception
 
+item_error
+------------
+
+.. signal:: item_error
+.. function:: item_error(item, response, spider, failure)
+
+    Sent when a :ref:`topics-item-pipeline` generates an error (ie. raises
+    an exception), except :exc:`~scrapy.exceptions.DropItem` exception.
+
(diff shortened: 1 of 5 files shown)


lucywang000:
@chainly can you explain the motivation for this, like a typical use case to justify this new signal? What's more I think a new test case is required.

chainly:
I want to count how many `item` dropped by unhandled exception occurred in `pipeline` easily ,  but only printed in log.

kmike:
This feature makes sense to me, and implementation looks good. The only question I have is why Failure is returned, not an exception, like in item_dropped?

chainly:
1. There are mistake in test case(`self.run.itemerror` ==> `self.run.itemresp`), now fix. Thanks to the question : )
2. For the question: if someone need `traceback`, `Failure` a good choise, see the doc [failure.Failure](https://github.com/twisted/twisted/blob/twisted-18.4.0/src/twisted/python/failure.py#L167-L174), and [failure.Failure.getTracebackObject](https://github.com/twisted/twisted/blob/twisted-18.4.0/src/twisted/python/failure.py#L540). Why lost, not known.

whalebot-helmsman:
@chainly Problem with unit-tests was fixed in master. Can you merge master to this PR?

chainly:
OK, fast-forwarded and push again.

PR #3260, opened 2018-05-15
[MRG+1] simple grammatical/formatting changes (minor)

herinckc:


whalebot-helmsman:
@herinckc Problem with unit-tests was fixed in master. Can you merge master to this PR?

herinckc:
@whalebot-helmsman I believe I've successfully merged this pull request with master. Also removed a previous change that was causing the build to fail. Let me know if I need to change anything

kmike:
Hey @herinckc! We haven't written these texts ourselves; license is BSD 3-clause, and Code of Conduct is from https://www.contributor-covenant.org/version/1/4/code-of-conduct. I'd prefer to keep them as-is; if you think there is an error in text, it is better to fix that in canonical texts. I'm not sure where to report issues with BSD license text (probably it is fixed forever); Code of Conduct spelling issues can be reported and fixed here: https://github.com/ContributorCovenant/contributor_covenant

cathalgarvey:
I'd agree these are changes to upstream to the source of the policy. However, unless we actively track the policy as it evolves and changes (and, I don't think we do?), I see no harm in merging these changes as-is, too.

kmike:
Ok, fair enough. +1 to merge after squashing then.

LICENSE
```diff
@@ -5,10 +5,10 @@ Redistribution and use in source and binary forms, with or without modification,
 are permitted provided that the following conditions are met:
 
     1. Redistributions of source code must retain the above copyright notice,
-       this list of conditions and the following disclaimer.
+       this list of conditions, and the following disclaimer.
 
     2. Redistributions in binary form must reproduce the above copyright
-       notice, this list of conditions and the following disclaimer in the
+       notice, this list of conditions, and the following disclaimer in the
```
dangra: I was reluctant to accept this change until I looked for more info and found this 😂 :
image

source: http://themoosebody.tumblr.com/post/10317138331/unlikelywords-why-the-oxford-comma-is

Issue #2658, opened 2017-03-16
large response causes scrapy to hang


rjbks:
I have tried stackoverflow but haven't received any answers for this as of yet.

[scrapy hangs on large response](http://stackoverflow.com/questions/42822010/scrapy-hanging-on-api-call-which-returns-large-json-object)

Summary:

Using scrapy to make thousands of API calls, all of them work fine except for 1 which returns a response of around 2GB (takes roughly 4 mins to complete using requests library). I have increased or disabled the related settings which may limit download size or timeout (DOWNLOAD_MAXSIZE DOWNLOAD_TIMEOUT). The logs show a 200 response then the warning for download size (which I left enabled just to indicate some form of progress) I then get 3 consecutive INFO logs from the stat tracker (1 minute apart), then it just hangs. I have left it running for up to 30 mins then I had to force quit (control-c does not work). Based on that, it would seem to hang after roughly 3 minutes after initiating that API call without any more logging whether I leave the download_timeout enable (for either 6.5 or 7 minutes when several attempts with requests package yields results in about 4 minutes) or disabled. Any ideas what could be causing this?

redapple:
hi @rjbks ,
I was able to download a large file with Scrapy [from MusicBrainz](http://ftp.musicbrainz.org/pub/musicbrainz/data/fullexport/20170315-001454/mbdump.tar.bz2) with `-s DOWNLOAD_TIMEOUT=100000 -s DOWNLOAD_MAXSIZE=0`
https://gist.github.com/redapple/42e800de71eaf663e3fcbfb02fa3a606

I have patched Scrapy to get more feedback during the download:
```
$ git diff
diff --git a/scrapy/core/downloader/handlers/http11.py b/scrapy/core/downloader/handlers/http11.py
index 6a0c6e4..44ef274 100644
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -427,9 +427,10 @@ class _ResponseReader(protocol.Protocol):
                             'request': self._request})
         if self._warnsize:
             if self._bytes_received > ((self._warnsize / 10) * (self._warnsize_times + 1)):
+                from humanize import naturalsize
                 self._warnsize_times = self._bytes_received // (self._warnsize / 10)
-                logger.debug("Received (%(bytes)s) bytes",
-                             {'bytes': self._bytes_received})
+                logger.debug("Received %(bytes)10s",
[...]

PR #3270, opened 2018-05-22
Fix large compressed responses causes scrapy to hang (#2658)

fbergen:
This is a fix for issue #2658 

**Problem**
some websites, e.g. http://www.safetradeintl.com sends a small gzipped file with >99% compression ratio, this causes 
1) The gunzip method in scrapy to hang, the way gunzip is uncompressing files is very slow for large files
2) Virtually unbounded sized responses, making the scraping less robust (note that download_maxsize setting doesn't work here)

An example site below:
$ wget -v http://www.safetradeintl.com
[...]
HTTP request sent, awaiting response... 200 OK
Length: 10420385 (9.9M) [text/html]
Saving to: 'index.html'
[...]

$ gunzip -l index.html.gz 
  compressed uncompressed  ratio uncompressed_name
    10420385   2147483648  99.5% index.html

**Solution**
1) Improve the gunzipping to not copy the entire output string in each iteration, effectively causing an O(n^2) performance
2) introduce a limit for uncompressed body size in httpcompression middleware

The PR in the current form is intended as a Draft and comments are welcomed!
Thanks!

### DIFF
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -3,8 +3,9 @@ import zlib
 from scrapy.utils.gz import gunzip
 from scrapy.http import Response, TextResponse
 from scrapy.responsetypes import responsetypes
-from scrapy.exceptions import NotConfigured
+from scrapy.exceptions import NotConfigured, IgnoreRequest
 
+logger = logging.getLogger(__name__)
 
 ACCEPTED_ENCODINGS = [b'gzip', b'deflate']
 
@@ -18,11 +19,15 @@ except ImportError:
 class HttpCompressionMiddleware(object):
     """This middleware allows compressed (gzip, deflate) traffic to be
     sent/received from web sites"""
+
+    def __init__(self, settings):
+        self._max_uncompressed_size = settings.get('MAX_UNCOMPRESSED_SIZE'], 0)
+
     @classmethod
     def from_crawler(cls, crawler):
         if not crawler.settings.getbool('COMPRESSION_ENABLED'):
             raise NotConfigured
-        return cls()
+        return cls(crawler.settings)
 
     def process_request(self, request, spider):
         request.headers.setdefault('Accept-Encoding',
@@ -66,4 +71,14 @@ class HttpCompressionMiddleware(object):
                 body = zlib.decompress(body, -15)
         if encoding == b'br' and b'br' in ACCEPTED_ENCODINGS:
             body = brotli.decompress(body)
+
+        if len(body) > self._max_uncompressed_size:
+            error_msg = ("Cancelling processing of %(url)s: Uncompressed "
+                         "response size %(size)s larger than max allowed size "
+                         "(%(maxsize)s).")
+            error_args = {'url': '', 'size': len(body),
+                          'maxsize': self._max_uncompressed_size}
+            logger.error(error_msg, error_args)
+            raise IgnoreRequest(error_msg % error_args)
+
         return body
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -22,7 +22,7 @@ class HttpCompressionMiddleware(object):
     sent/received from web sites"""
 
     def __init__(self, settings):
-        self._max_uncompressed_size = settings.get('MAX_UNCOMPRESSED_SIZE'], 0)
+        self._max_size = settings.get('MAX_UNCOMPRESSED_SIZE', 0)
 
     @classmethod
     def from_crawler(cls, crawler):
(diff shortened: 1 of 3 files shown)


kmike:
I think that's a great idea; +1 to add such option.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -37,6 +43,16 @@ def process_response(self, request, response, spider):
             if content_encoding:
                 encoding = content_encoding.pop()
                 decoded_body = self._decode(response.body, encoding.lower())
+                if self._max_size and len(decoded_body) > self._max_size:
```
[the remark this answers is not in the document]
kmike: It'd be also nice to figure out if we can support this option for deflate and brotli

scrapy/utils/gz.py
```diff
@@ -30,25 +30,25 @@ def gunzip(data):
...
             # complete only if there is some data, otherwise re-raise
             # see issue 87 about catching struct.error
-            # some pages are quite small so output is '' and f.extrabuf
+            # some pages are quite small so output_list is empty and f.extrabuf
             # contains the whole page content
-            if output or getattr(f, 'extrabuf', None):
+            if output_list or getattr(f, 'extrabuf', None):
                 try:
-                    output += f.extrabuf[-f.extrasize:]
+                    output_list.append(f.extrabuf[-f.extrasize:])
```
kmike: Have you seen a speed improvement after this change? AFAIK algoritm wasn't O(N^2) in CPython because, but O(N^2) in pypy; are you using pypy?

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -37,6 +43,16 @@ def process_response(self, request, response, spider):
...
+                    error_msg = ("Cancelling processing of %(url)s: "
+                                 "Uncompressed response size %(size)s larger "
+                                 "than max allowed size (%(maxsize)s).")
+                    error_args = {'url': response.url,
+                                  'size': len(decoded_body),
+                                  'maxsize': self._max_size}
+                    logger.error(error_msg, error_args)
+                    raise IgnoreRequest(error_msg % error_args)
```
fbergen: Not sure IgnoreRequest is the right thing to raise here

scrapy/downloadermiddlewares/httpcompression.py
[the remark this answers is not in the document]
fbergen: Indeed, I was toying around with pushing this down to the gunzip method, but decided against it because I anyway want to keep a highlevel check to make sure this covers all decompressing algorithms brotly/deflate 

Happy to complement by adding an optional argument to gunzip(data, max_size=0) and abort sooner,

scrapy/utils/gz.py
fbergen: Thanks!

I'm using cpython, and see very big improvement, for big files this is a matter of completing within few minutes, or stalling forever.

If you refer to cpython's realloc instruction I believe for big files realloc ends up needing to copy the string frequently,  please let me know if there is another reason why this should be O(n) in Cpython)

scrapy/downloadermiddlewares/httpcompression.py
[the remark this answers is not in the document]
fbergen: +1 to also support brotli/deflate, I couldn't find a trivial way to accomplish that, other than rolling our own brotli decompression function (similar to what we do for gunzip [which tbh I don't fully understand the rationale behind]), 

Afraid this PR becomes too big if we want to implement this for brotli/deflate, but happy to follow your advice here

Thanks!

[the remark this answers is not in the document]
kmike: Right, it makes sense to support it both at gunzip level and on a lavel higher - at least it'd allow this option to work for brotli/deflate, even though performance won't be there yet.

scrapy/downloadermiddlewares/httpcompression.py
kmike: A good question. I think IgnoreRequest is a right thing to raise, but it is better to create a subclass, so that users can catch this exception specifically, and so that it shows up under a different name in logs and stats.

scrapy/utils/gz.py
kmike: Yes, I was talking about optimization which CPython has. You're right it is not O(N) because of realloc, but it is still pretty fast; I've tried a couple of benchmarks, and there was no significant difference, it was not clear which one is faster:

```
In [1]: %%time
   ...: out = ""
   ...: for x in range(10000):
   ...:     out += "." * 8000
   ...:
CPU times: user 41.6 ms, sys: 31.4 ms, total: 73 ms
Wall time: 71.8 ms
```

```
In [3]: %%time
   ...: res = []
   ...: for x in range(10000):
   ...:     res.append("." * 8000)
   ...: out = "".join(res)
   ...:
CPU times: user 37.9 ms, sys: 30.2 ms, total: 68.1 ms
Wall time: 66.2 ms
```

time vary quite a lot, but overall it seems roughly the same.

BUT it seems this optimization is not implemented for bytes in Python 3! And we missed it while porting Scrapy to Python 3. In Python 2 bytes was the same as str, and had this optimization, so the code worked fine; in Python 3 only `str` has this optimization (aka `unicode` in Python 2). So the following takes forever, unlike a version with `b"".join`:

```
%%time
out = b""
for x in range(10000):
    out += b"." * 8000
``` 

I think this a bug separate from `MAX_UNCOMPRESSED_SIZE`. What do you think about opening a separate pull request with a fix? A benefit is that it could be merged faster, as it is a clear porting bug which doesn't require much discussion.

fbergen: Nice find! 

Opened #3281

Gallaecio:
https://github.com/scrapy/scrapy/issues/2658#issuecomment-1384372360

PR #3279, opened 2018-05-30
Update spiders.rst

lewoudar:
I changed URLs to :class:`~scrapy.http.Request` in start_urls explanation of the default spider


kmike:
A good catch, thanks @lewoudar!

lewoudar:
you are welcome :)

PR #3281, opened 2018-06-01
[MRG+2] Improve gunzip performance for big files on Python 3

fbergen:
Copied from discussion on #3270 , partial fix for #2658 

**Problem**
gunzip method runs forever for big files in python 3

bytes in Python 3 does not include string concatenation improvements that Python 2 has, effectively making the following take forever:

```
%%time
out = b""
for x in range(10000):
    out += b"." * 8000
```

**Solution** 
Improve the gunzip method to not copy the entire output string in each iteration, but instead append to list + join, significantly improving performance from  O(n^2)

Thanks @kmike for your help debugging!

### DIFF
--- a/scrapy/utils/gz.py
+++ b/scrapy/utils/gz.py
@@ -30,25 +30,25 @@ def gunzip(data):
     This is resilient to CRC checksum errors.
     """
     f = GzipFile(fileobj=BytesIO(data))
-    output = b''
+    output_list = []
     chunk = b'.'
     while chunk:
         try:
             chunk = read1(f, 8196)
-            output += chunk
+            output_list.append(chunk)
         except (IOError, EOFError, struct.error):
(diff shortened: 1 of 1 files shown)


kmike:
//cc @nctl144 and @lopuhin, as you're working on Scrapy performance improvements and becnhamarks. This change fixes an issue explained here: https://github.com/scrapy/scrapy/issues/2658#issuecomment-342880551 - do you think large and gzip-compressed responses should be added to https://github.com/scrapy/scrapy-bench? Maybe add an option to reproduce not only HTML content of real-world websites from Alexa top, but raw binary responses and HTTP headers as well?

kmike:
Looks great, thanks @fbergen! +1 to merge after tests pass.

cathalgarvey:
Looks good to me, thanks @fbergen!

PR #3283, opened 2018-06-08
Fix #3247: Allow scrapy.FormRequest.from_response method to handle duplicate keys

### LINKED ISSUE #3247: `scrapy.FormRequest.from_response` observed to eliminate duplicate keys in `formdata`
This looks good:
```
In [2]: scrapy.FormRequest('http://example.com', method='GET', formdata=(('foo', 'bar'), ('foo', 'baz')))
Out[2]: <GET http://example.com?foo=bar&foo=baz>
```

While here is the issue:
```
In [3]: response = scrapy.http.TextResponse(url='http://example.com', body='<form></form>', encoding='utf8')

In [4]: scrapy.FormRequest.from_response(response, method='GET', formdata=(('foo', 'bar'), ('foo', 'baz')))
Out[4]: <GET http://example.com?foo=baz>
```

(Tested with `Scrapy 1.5.0` and `Python 3.6.5`)

CCInCharge:
Fix #3247 

Dicts do not allow duplicate keys. When form requests have duplicate keys, one of the values of the duplicate keys is dropped when created via the FormRequest's from_response method.

`formdata` is either a dict or iterable of tuples - this patch fixes the above issue by not converting `formdata` to a dict and keeping it as an iterable of tuples if possible.

Also, a unit test is added to verify correct functionality. This is my first PR, definitely let me know if any changes are needed!

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -114,10 +114,12 @@ def _get_form(response, formname, formid, formnumber, formxpath):
 
 def _get_inputs(form, formdata, dont_click, clickdata, response):
     try:
-        formdata = dict(formdata or ())
+        formdata_keys = dict(formdata or ()).keys()
     except (ValueError, TypeError):
         raise ValueError('formdata should be a dict or iterable of tuples')
 
+    if not formdata:
+        formdata = ()
     inputs = form.xpath('descendant::textarea'
                         '|descendant::select'
                         '|descendant::input[not(@type) or @type['
@@ -128,14 +130,17 @@ def _get_inputs(form, formdata, dont_click, clickdata, response):
                             "re": "http://exslt.org/regular-expressions"})
     values = [(k, u'' if v is None else v)
               for k, v in (_value(e) for e in inputs)
-              if k and k not in formdata]
+              if k and k not in formdata_keys]
 
     if not dont_click:
         clickable = _get_clickable(clickdata, form)
         if clickable and clickable[0] not in formdata and not clickable[0] is None:
             values.append(clickable)
 
-    values.extend((k, v) for k, v in formdata.items() if v is not None)
+    if isinstance(formdata, dict):
+        formdata = formdata.items()
+
+    values.extend((k, v) for k, v in formdata if v is not None)
     return values
 
 
--- a/tests/test_http_request.py
+++ b/tests/test_http_request.py
@@ -401,6 +401,29 @@ class FormRequestTest(RequestTest):
         self.assertEqual(fs[u'test2'], [u'xxx µ'])
         self.assertEqual(fs[u'six'], [u'seven'])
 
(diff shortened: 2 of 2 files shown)


scrapy/http/request/form.py
```diff
@@ -114,10 +114,12 @@ def _get_form(response, formname, formid, formnumber, formxpath):
 
 def _get_inputs(form, formdata, dont_click, clickdata, response):
     try:
-        formdata = dict(formdata or ())
+        formdata_keys = dict(formdata or ()).keys()
     except (ValueError, TypeError):
         raise ValueError('formdata should be a dict or iterable of tuples')
 
+    if not formdata:
+        formdata = dict(formdata or ())
```
HassanQamar07: This can be simplified to `formdata = ()`

kmike:
Thanks @CCInCharge! Would you mind adding a test which checks how it interacts with fields of the same name - similar to https://github.com/CCInCharge/scrapy/blob/d4b4118bad04768e68b43ec8ac3deb5348a6a95a/tests/test_http_request.py#L447, but with several values?

CCInCharge:
Ok @kmike and @HassanQamar07 , just edited the commit to incorporate your comments.

tests/test_http_request.py
```diff
@@ -401,6 +401,29 @@ def test_from_response_post_nonascii_unicode(self):
...
+                method='GET',
+                formdata=(('foo', 'bar'), ('foo', 'baz')))
+        self.assertEqual(urlparse(req.url).hostname, 'www.example.com')
+        self.assertEqual(urlparse(req.url).query, 'foo=bar&foo=baz')
+    
+    def test_from_response_override_duplicate_form_key(self):
+        response = _buildresponse(
+            """<form action="get.php" method="POST">
+            <input type="hidden" name="one" value="1">
+            <input type="hidden" name="two" value="3">
```
kmike: I think having value=2 would make the test case even better - now it is not clear if an existing value is overridden or not

CCInCharge:
Thanks! Test case has been changed per @kmike request.

dangra:
> Thanks! Test case has been changed per @kmike request.

was it? I don't see any commit addressing this comment https://github.com/scrapy/scrapy/pull/3283#discussion_r205036094

CCInCharge:
782f866 looks like it should have the changes (Lines 420-425)

maramsumanth:
Hey I have an issue in this, what happens if there is fragment in url?

```response = scrapy.http.TextResponse(url='http://example.com/#frag', body='<form></form>',encoding='utf8')```

```scrapy.FormRequest.from_response(response, method='GET', formdata=(('foo', 'bar'), ('foo', 'baz')))```

should have output as
```<GET http://example.com/?foo=bar&foo=baz#frag>```
but it is showing
```<GET http://example.com/#frag?foo=bar&foo=baz>```

PR #3307, opened 2018-06-25
Add python-scrapy and python3-scrapy links

yaamas:
There is no official way described to download as ubuntu package in the docs.
Also, they are up to date currently.

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -141,9 +141,18 @@ twisted and pyOpenSSL, and is compatible with recent Ubuntu distributions.
 But it should support older versions of Ubuntu too, like Ubuntu 14.04,
 albeit with potential issues with TLS connections.
 
-**Don't** use the ``python-scrapy`` package provided by Ubuntu, they are
-typically too old and slow to catch up with latest Scrapy.
+**Try** not to use the ``python3-scrapy`` / ``python-scrapy`` package provided by Ubuntu, because they are **sometimes** typically too old and slow to catch up with latest Scrapy. To check whether ``python3-scrapy`` or ``python-scrapy`` is up to date with the latest available version, check the ubuntu updates official website below:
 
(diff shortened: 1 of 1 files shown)


grammy-jiang:
The failure in Travis-ci is very wired. I have seen these no sense errors these days. 

Does anyone know what happens?

lopuhin:
I think these are failures of 1.5 branch, the base was changes but probably the build was using the old base and wasn't restarted, I'll try restarting it manually.

grammy-jiang:
hi, @lopuhin,

How things is going on? I have the same problem in my PR, and I am looking forward to passing my testings.

kmike:
The problem is that these packages will become outdated soon - we're going to release scrapy 1.6, and then they will stay outdated for another 5 years (18.04 is a recent LTS release), right?

nyov:
Whether they are outdated or not, I don't think there needs to be a tutorial on how to install packages in Ubuntu, here. Then the next guy wants to add a line on how to install them with aptitude or synaptic instead :p My 2 cents.
Also: `**sometimes** typically` ?

Kogam22:
@nyov : 
>I don't think there needs to be a tutorial on how to install packages in Ubuntu, here. Then the next guy wants to add a line on how to install them with aptitude or synaptic instead :p My 2 cents.

Yup that is sensible. But it should at least be available in tavailable in the Docshe Docs right ?

nyov:
I don't have a strong opinion either way. Only that I think people writing python scrapers are knowledgeable enough to know how to use their system package manager or build toolchain.
The dependencies are the actual package names, right? And the location is the standard system repository (`apt search scrapy` is enough to find it?), not a 3rd-party source that would require a linking, so I don't see the point really.
But I also haven't been a new user in a long time, so YMMV.

I do think the emphasis on "don't use [system packages]" could be relaxed. It's usually only the devs who think everyone should use their latest and greatest version. A version packaged and frozen 5 years ago (with dependencies) still works today as it did then. Some people might prefer that stability (until they hit on an old bug).

PR #3308, opened 2018-06-26
TST exclude lxml==4.2.2 from tests

kmike:
This PR is an example of Travis-driven development, as I can't reproduce an issue locally on OS X.

1. forcing image to loading before temp file is gone - not a fix.
2. reverting back to BytesIO, as a Pillow issue seems to be solved (see https://github.com/scrapy/scrapy/pull/803#issuecomment-49028433 for motivation for moving to TemporaryFile) - not a fix, though probably it makes the code cleaner.

TODO: fix it. It seems the issue is related to PNG codec, as JPEG images work.


grammy-jiang:
I have met this problem these days, and I have several PRs and their testings to wait to pass.

But sorry I am not familiar with the image related libraries, could not provide any help.

Hope this issue could be fixed soon.

kmike:
@grammy-jiang this is actually caused by lxml (see https://github.com/python-pillow/Pillow/issues/3208); lxml reverted the offending commit in a new release (https://github.com/lxml/lxml/pull/265). So the fix would be to exclude bad lxml version, I'll do it in this PR.

PR #3309, opened 2018-06-26
[MRG+1] TST test agains latest pypy

kmike:
I think it is better to test against latest pypy, as it is still a moving target, and we don't absolutely need to support older pypy versions.


kmike:
The problem is that tests segfault with pypy3 6.0 :)

PR #3311, opened 2018-06-29
[MRG+1] TST make it clear which requirements are Python 2-only

kmike:
I was always forgetting if requirements-py3.txt is applied on top of requirements.txt (false), or if they are totally separate requirements files (true). To avoid confusion, requirements.txt is renamed to requirements-py2.txt in this PR.

* rename requirements.txt to requirements-py2.txt, to make it clear they are Python 2-only
* make requirements-py3.txt consistent with requirements-py2.txt


grammy-jiang:
BTW, since Python 2.7 will not be supported after 2020 officially, is there any schedule for Scrapy to give up Python 2.7 support?

kmike:
@grammy-jiang no schedule yet. I don't think we'll be too aggressive.

grammy-jiang:
@kmike 

Ok, now I have to import a third party package in order to compatible with Python 2.7. Is there any specific guideline about importing the new package in this project? I have read the page of contribution, and find nothing related.

kmike:
Hey @grammy-jiang! functools32 is now a [parsel](https://github.com/scrapy/parsel) dependency (since yesterday), and parsel is a Scrapy dependency, so you can count on it being available. But it is better to add this dependency explicitly to Scrapy as well if you're using it there, so that Scrapy doesn't rely on transitive dependencies; check how it is done in parsel's [setup.py](https://github.com/scrapy/parsel/blob/master/setup.py).

grammy-jiang:
Hi, @kmike 

Thanks for your reply.

I see, in parsel, six.PY2 is used to choose where the method lru_cache is imported from. That's impressive.

But why not use try and except to do this, just like what we did [here](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpproxy.py#L4-L7)?

kmike:
> But why not use try and except to do this, just like what we did here?

I don't think it matters much, either way is fine.

grammy-jiang:
That's great, thanks!

Issue #3195, opened 2018-03-31
Downloader middleware httpproxy enhancement


grammy-jiang:
Hi,

The downloader middleware httpproxy is a simple and clear middleware of scrapy. Until now it works well - all functions provided by it is following the relative RFC:

- [RFC 7235 - Hypertext Transfer Protocol (HTTP/1.1): Authentication](https://tools.ietf.org/html/rfc7235#section-4.4)
- [RFC 7617 - The 'Basic' HTTP Authentication Scheme](https://tools.ietf.org/html/rfc7617)

And also:

- [Proxy-Authorization - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization)

But there are still some places can be enhanced:

- [ ] The proxy bypass now is called each time when a request comes into downloader; this could be ignored by lru_cache (decorator or traditional way), now working on PR #3316
- [ ] The data (proxies) and the usage of proxies are coupled in the one middleware; it should be like another middleware - httpcache - decouple the usage and data (backend), and it even has another object to control the caching strategy (HTTPCACHE_POLICY, e.g. DummyPolicy)
- [ ] The data (proxies, backend or storage) could be many, just following the same protocol (interface)

With these features, this middleware can provide more powerful ways to use proxies:

- use more proxies from different backends/storages
- manage the proxies, e.g. put invalidated proxies into the blacklist and ignore it for a certain period or forever

I have done all of above in my repo: [grammy-jiang/scrapy-proxy-management: A proxy management middleware and extension for scrapy](https://github.com/grammy-jiang/scrapy-proxy-management), so you can easily understand what I would like to contribute by reading my repo.

I will split the code into pieces to pull back to scrapy for easy review.

Any suggestions or discussions are welcome!

AdrianAtZyte:
I am not sure such complexity should be added to core. We could make changes if needed to make it easier for a proxy rotation extension to be implemented, but as far as I know there are many such extensions and they work fairly well.

Issue #2568, opened 2017-02-15
finish relocations


kmike:
What do you think about removing shims for moved/renamed modules from https://github.com/scrapy/scrapy/issues/1063? Scrapy 1.0 release was 20 months ago, and upgrade path for users is clear.

Parth-Vader:
If the modules have been renamed, I think it should be removed for the newer versions of Scrapy.

kmike:
My main motivation is that I'm getting imports wrong every time - IDE autocomplete shows both deprecated and new paths (e.g. scrapy.dupefilter and scrapy.dupefilters), and it is hard to remember which one to use. I'm checking their source code every time to avoid deprecation warning :)

Parth-Vader:
@kmike I'm new to this, but in my opinion , a user would get confused if autocomplete shows both options. Then he'd have to look up in the docs.
Hence, my main purpose is that the user should have a smooth experience.

kmike:
@Parth-Vader the downside or removing deprecations is that obviously it could break a lot of existing code.

Parth-Vader:
@kmike Yeah, that's true. It's always hard to remove things once the project becomes so complex. :stuck_out_tongue:

djunzu:
Wouldn't be the best course of action establish a deprecation policy, announce it on the next release, and start to execute it in the following release. So, if the policy says "deprecated methods will be dropped after three minor releases", then everyone knows when it will be removed.

kmike:
@djunzu it makes sense, though we tend to keep deprecations for a long time if they don't hurt. From the developer point of view some of the deprecations affect productivity, and some don't. 

I think currently we're following this approach: it is OK to deprecate something in major release X and remove it in major release X+1 (given that they separated by at least half year or a year), but if deprecated path doesn't hurt it can be kept for much longer. "Deprecated methods will be dropped after three minor releases" can be more aggressive than the current policy, though more predicatable. I agree that it is good to have policy written down explicitly.

kmike:
@dangra @eliasdorneles @lopuhin what's your opinion on doing this in 1.6 release? Old locations were deprecated more than 3 years ago (Scrapy 1.0.0 is released 2015-06-19).

dangra:
3 years is a lot already, +1 from me to cleanup.

PR #3316, opened 2018-07-04
Make the structure of the HttpProxyMiddleware clearly and add lru_cache to proxy_bypass

grammy-jiang:
This PR is mainly to:

* clear the structure of the class of HttpProxyMiddleware
* add lru_cache to **proxy_bypass**

based on the issue #3195, to prepare for the next step development.

The things changed in this PR include:

* Add the third party package functools32 in order to be compatible with Python 2.7
* Move the methods **_basic_auth_header**, **_get_proxy** out of the class HttpProxyMiddleware, to make the structure of this class clearly (keep fake methods inside the class to maintain backward compatibility and pop the deprecated warning, noted by @kmike)
* Use lru_cache to cache the result of the method of **proxy_bypass**
* Because of the import of lru_cache, the testcase of this middleware need to add a procedure to clear the cache before each assert

Please refer to issue #3195 for more details.


scrapy/downloadermiddlewares/httpproxy.py
```diff
@@ -11,13 +15,35 @@
 from scrapy.utils.python import to_bytes
 
 
+def basic_auth_header(username, password, auth_encoding):
```
kmike: +1 to extract it, though I think it is better to refactor code to use w3lib's [basic_auth_header](https://github.com/scrapy/w3lib/blob/c1a030582ec30423c40215fcd159bc951c851ed7/w3lib/http.py#L81) function instead.

scrapy/downloadermiddlewares/httpproxy.py
```diff
@@ -26,30 +52,13 @@ def from_crawler(cls, crawler):
...
-        proxy_type, user, password, hostport = _parse_proxy(url)
-        proxy_url = urlunparse((proxy_type or orig_type, hostport, '', '', '', ''))
-
-        if user:
-            creds = self._basic_auth_header(user, password)
-        else:
-            creds = None
-
-        return creds, proxy_url
-
```
kmike: Removing these methods is backwards-incompatible, because HttpProxyMiddleware subclasses which override or call these methods will stop working properly. I know they are private and undocumented, but they are called (though not overridden) in real-world HttpProxyMiddleware subclasses - I've found examples in a large codebase. What do you think about keeping them, but maybe just calling functions from there, and maybe after issuing a deprecation warning?

scrapy/downloadermiddlewares/httpproxy.py
grammy-jiang: Sure, that's ok.

grammy-jiang: Hi, @kmike 

Now I use the one provided in w3lib, the code becomes less and better. 

But [the test case **test_proxy_auth_encoding**](https://github.com/scrapy/scrapy/blob/1.5/tests/test_downloadermiddleware_httpproxy.py#L110-L113) failed. The failure log:

```
Failure
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/internet/defer.py", line 310, in addCallbacks
    self._runCallbacks()
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/trial/_asynctest.py", line 141, in deferTestMethod
    d = self._run(self._testMethodName, result)
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/trial/_asynctest.py", line 112, in _run
    utils.runWithWarningsSuppressed, self._getSuppress(), method)
--- <exception caught here> ---
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/internet/utils.py", line 201, in runWithWarningsSuppressed
    reraise(exc_info[1], exc_info[2])
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/python/compat.py", line 463, in reraise
    raise exception.with_traceback(traceback)
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/internet/utils.py", line 197, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/home/grammy-jiang/PycharmProjects/scrapy/tests/test_downloadermiddleware_httpproxy.py", line 124, in test_proxy_auth_encoding
    self.assertEqual(req.headers.get('Proxy-Authorization'), b'Basic /HNlcjpwYXNz')
  File "/home/grammy-jiang/.virtualenvs/webcrawler/lib64/python3.6/site-packages/twisted/trial/_synctest.py", line 432, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/usr/lib64/python3.6/unittest/case.py", line 846, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/python3.6/unittest/case.py", line 839, in _baseAssertEqual
    raise self.failureException(msg)
builtins.AssertionError: b'Basic _HNlcjpwYXNz' != b'Basic /HNlcjpwYXNz'
```
The log shows the base64 encoding different at the first character. Then I checked the source code, and found base64 encode method is different:

* the original method is used [base64.b64encode](https://github.com/scrapy/scrapy/blob/1.5/scrapy/downloadermiddlewares/httpproxy.py#L34), which is purely encoding without any escape special character
* the one in w3lib is used [base64.urlsafe_b64encode](https://github.com/scrapy/w3lib/blob/v1.19.0/w3lib/http.py#L99), which [docstring](https://github.com/python/cpython/blob/3.6/Lib/base64.py#L114-L116) said:
**The result is returned as a bytes object.  The alphabet uses '-' instead of '+' and '_' instead of '/'.**

That character escape makes the test case failed. And also I go through [the test suit of w3lib.http](https://github.com/scrapy/w3lib/blob/master/tests/test_http.py) - there are some cases for encoding, but it seems that all of them bypass this failure case perfectly.

Is this a issue about the **w3lib.http.basic_auth_header**?

scrapy/downloadermiddlewares/httpproxy.py
grammy-jiang: Reasonable, I have kept these two methods now, with a deprecated warning.

The message said the method will be removed in the future release, but which version will be the future release? Is there a schedule for deprecation in scrapy?

kmike: We don't have scheduled removal of deprecated features; we keep them for +1 release for sure, and then we can keep them for much longer, depending on whether they cause maintenance issues. There was a bit of discussion about deprecation policy here: https://github.com/scrapy/scrapy/issues/2568.

scrapy/downloadermiddlewares/httpproxy.py
kmike: @grammy-jiang I'm not sure, I guess it needs some detective-style work with git (and github) to figure out why a different encoding is used for Authorization and Proxy-Authorization headers, maybe one of them is incorrect.

grammy-jiang: Hi, @kmike 

I have gone through the following documents to make the proxy authentication clear:

* [HTTP authentication - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication)
This page introduces nearly all information about authentication in HTTP protocol, and it states that the headers **Authorization** and **Proxy-Authorization** have several schemes (encode or encrypt ways), the **basic** scheme is one of them ([here](http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) has a list of authentication schemes, maybe in the future all of them will get support by scrapy). Then I keep gathering more information about the **basic** scheme.

  * [HTTP authentication - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Basic_authentication_scheme)
Yes, there is a section in the same page to talk about **basic** scheme, and there is one sentence:
    > The "Basic" HTTP authentication scheme is defined in RFC 7617, which transmits credentials as user ID/password pairs, encoded using base64.

  * [RFC 7617 - The 'Basic' HTTP Authentication Scheme](https://tools.ietf.org/html/rfc7617)
This RFC should be the guideline for the **basic** scheme, and actually, it has only one sentence:
    > This document defines the "Basic" Hypertext Transfer Protocol (HTTP) authentication scheme, which transmits credentials as user-id/password pairs, encoded using Base64.

  None of them mention the base64 encode should be escaped for url safe. I think it is because the credentials only exist in headers, not in url.

And also I try to find some third parties' tools to test again:

* [Basic Authentication Header Generator](https://www.blitter.se/utils/basic-authentication-header-generator/)

The result from this online tool proves that the **basic** scheme only use base64 encoding, without special character escaping.

Conclusion: the function basic_auth_header in w3lib should use **base64.b64encode**, not **base64.urlsafe_b64encode**.

grammy-jiang: Hi, @kmike 

I change this basic_auth_header back to the original one. I will use the one in **w3lib.http** when it can pass the original test case in Scrapy.

PR #3326, opened 2018-07-09
Python 3.7 support

### LINKED ISSUE #3143: Python 3.7 support
The goal is to add python 3.7 to travis and pass all tests, the first beta was already released at the end of January.

lopuhin:
Continue @patiences work on python 3.7 compatibility from #3150, fixes #3143. Thanks a lot @patiences!

It's likely that we'll release scrapy 1.6 before new twisted version with https://github.com/twisted/twisted/pull/966 is released (sorry for making the PR stall), so in this PR I work around this issue, so that scrapy at least works, although without telnet console support yet.

Also pytest is updated, so that ``tox -e py37`` can collect the tests (2.9 was released more than 2 years ago).

~~WIP as I want to check how tests work on travis early (there are still a few failures due to PEP 479 left which didn't show up in #3150).~~

### DIFF
--- a/scrapy/extensions/telnet.py
+++ b/scrapy/extensions/telnet.py
@@ -6,6 +6,7 @@ See documentation in docs/topics/telnetconsole.rst
 
 import pprint
 import logging
+import traceback
 
 from twisted.internet import protocol
 try:
@@ -13,6 +14,7 @@ try:
     from twisted.conch.insults import insults
     TWISTED_CONCH_AVAILABLE = True
 except (ImportError, SyntaxError):
+    _TWISTED_CONCH_TRACEBACK = traceback.format_exc()
     TWISTED_CONCH_AVAILABLE = False
 
 from scrapy.exceptions import NotConfigured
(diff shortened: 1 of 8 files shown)


lopuhin:
Travis build is green now, this is ready for review.

scrapy/extensions/telnet.py
```diff
@@ -40,7 +40,8 @@ def __init__(self, crawler):
         if not crawler.settings.getbool('TELNETCONSOLE_ENABLED'):
             raise NotConfigured
         if not TWISTED_CONCH_AVAILABLE:
-            raise NotConfigured
+            raise NotConfigured('TelnetConsole not enabled: failed to import '
+                                'required twisted modules.')
```
kmike: I wonder if we should log the traceback here, and maybe explain that to remove an error TELNETCONSOLE_ENABLED can be set to False explicitly, if telnet is not needed.

lopuhin: Thanks for suggestion @kmike , I think it gives more visibility to the issue but does not harm anyone. This is how it looks for me after 9428a4a3:
```
2018-07-09 21:06:20 [scrapy.middleware] WARNING: Disabled TelnetConsole: TELNETCONSOLE_ENABLED setting is True but required twisted modules failed to import:
  File "/Users/kostia/shub/scrapy/scrapy/extensions/telnet.py", line 13, in <module>
    from twisted.conch import manhole, telnet
  File "/Users/kostia/shub/scrapy/py37/lib/python3.7/site-packages/twisted/conch/manhole.py", line 154
    def write(self, data, async=False):
                              ^
SyntaxError: invalid syntax
```

lopuhin:
Btw, I did a quick check of speed via ``scrapy bench`` and Python 3.7 is a bit faster than Python 3.6

kmike:
Thanks @lopuhin and @patiences!

PR #3327, opened 2018-07-09
[MRG+1] Remove deprecated CrawlerSettings class and Settings attributes

nyov:
On the heels of removing those relocation shims, I believe the time has come for `CrawlerSettings`, `Settings.overrides` and `Settings.defaults` as well.
(`...is deprecated and won't be supported in Scrapy 0.26` ... yah, right. Lies!)

As a note to anyone napping the last four years; use `Settings.set(name, value, priority='cmdline')` (overrides) and `Settings.set(name, value, priority='default')` (defaults) instead.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,15 @@
 Release notes
 =============
 
+Scrapy 1.6.0 (unreleased)
+-------------------------
+
+Cleanups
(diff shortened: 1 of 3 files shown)


kmike:
Looks good to me, thanks for the cleanup @nyov!

PR #3335, opened 2018-07-13
update default RETRY_HTTP_CODES on docs

stummjr:
The docs on `RETRY_HTTP_CODES` are outdated in regards to the list of HTTP status codes that are retried by default by the `RetryMiddleware`.

Here's the updated list: https://github.com/scrapy/scrapy/blob/master/scrapy/settings/default_settings.py#L239


kmike:
A great catch! FTR: we forgot to update them here: https://github.com/scrapy/scrapy/pull/2852.

PR #3342, opened 2018-07-19
[MRG+1] Include flags when copying requests

elacuesta:
I noticed flags were missing after a request was retried by the retry downloader middleware, traced the issue and realized that they were not being copied.
Small test case included.


kmike:
A good catch, thanks @elacuesta!

PR #3346, opened 2018-07-19
Create class to allow send emails through Amazon SES

### LINKED ISSUE #3345: Allow Scrapy to send e-mails using Amazon SES
I need to create an e-mail report and send it using Amazon SES. Scrapy only has `scrapy.email.MailSender` class that only allows to send e-mails using a SMTP server.

Considering that Scrapy already has integration for some Amazon services (like S3 storage), I would like to suggest the creation of `scrapy.email.SESMailSender` with the same interface of `MailSender` (with the expected differences, like providing AWS keys instead of SMTP host).

rennerocha:
This pull-request is related to #3345 . It still needs some code, tests and docs, but I would like to send it to review, so more people can review it.

Fixes #3345


scrapy/mail.py
```diff
@@ -5,6 +5,8 @@
 """
 import logging
 
+import boto
```
kmike: this import should be near twisted import; pep8 [suggests](https://www.python.org/dev/peps/pep-0008/?#imports) the following groups in import blocks:

1) Standard library imports (we consider `six` as stdlib imports in Scrapy).
2) Related third party imports.
3) Local application/library specific imports.

scrapy/mail.py
```diff
@@ -36,28 +38,18 @@ def _to_bytes_or_none(text):
...
-            settings.getbool('MAIL_TLS'), settings.getbool('MAIL_SSL'))
+        raise NotImplementedError
 
     def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', charset=None, _callback=None):
+        raise NotImplementedError
+
+    def _get_message(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', charset=None):
         if attachs:
-            msg = MIMEMultipart()
+            msg = MIMEMultipart('alternative')
```
kmike: Why is this needed? Is it related to SESMailSender?

scrapy/mail.py
```diff
@@ -68,10 +60,6 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
         msg['To'] = COMMASPACE.join(to)
         msg['Date'] = formatdate(localtime=True)
         msg['Subject'] = subject
-        rcpts = to[:]
-        if cc:
-            rcpts.extend(cc)
-            msg['Cc'] = COMMASPACE.join(cc)
```
kmike: after moving it out of here `cc` argument seems to be unused

scrapy/mail.py
```diff
@@ -82,12 +70,37 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
...
+        self.smtptls = smtptls
+        self.smtpssl = smtpssl
+        self.mailfrom = mailfrom
+        self.debug = debug
+
+    @classmethod
+    def from_settings(cls, settings):
+        return cls(settings['MAIL_HOST'], settings['MAIL_FROM'], settings['MAIL_USER'],
+            settings['MAIL_PASS'], settings.getint('MAIL_PORT'),
+            settings.getbool('MAIL_TLS'), settings.getbool('MAIL_SSL'))
```
kmike: I know it was like this before, but with such long lists of arguments it usually makes more sense to use kwargs - much less chance to pass a value to a wrong argument.

scrapy/mail.py
```diff
@@ -98,6 +111,10 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
                           'mailattachs': len(attachs)})
             return
 
+        rcpts = to[:]
+        if cc:
+            rcpts.extend(cc)
```
kmike: cc is no longer passed through `list(arg_to_iter(cc))`, and `msg['Cc']` is no longer updated, this looks backwards incompatible

scrapy/mail.py
```diff
@@ -135,3 +152,37 @@ def _sendmail(self, to_addrs, msg):
...
+        self.aws_secret_key = aws_secret_key
+        self.mailfrom = mailfrom
+        self.debug = debug
+
+    @classmethod
+    def from_settings(cls, settings):
+        return cls(
+            settings['MAIL_SES_AWS_ACCESS_KEY'],
+            settings['MAIL_SES_AWS_SECRET_KEY'],
+            settings['MAIL_FROM']
```
kmike: I'd prefer to use kwargs here as well, for robustness

scrapy/mail.py
```diff
@@ -135,3 +152,37 @@ def _sendmail(self, to_addrs, msg):
...
+        )
+
+    def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', charset=None, _callback=None):
+        msg = self._get_message(to, subject, body, cc, attachs, mimetype, charset)
+
+        if _callback:
+            _callback(to=to, subject=subject, body=body, cc=cc, attach=attachs, msg=msg)
+
+        if self.debug:
+            logger.debug('Debug mail sent OK: To=%(mailto)s Cc=%(mailcc)s '
```
kmike: Cc is unused, right?

kmike:
What do you think about adding an option for default MailSender? Currently StatsMailer and MemoryUsage have MailSender class hardcoded.

scrapy/mail.py
rennerocha: `boto` is only used with `SESMailSender`, so the user is not obligated to have it installed. I am moving this import inside the `send` method of the mentioned class (this is done the same way in `S3FeedStorage` for example).

kmike: Right, a good catch.

scrapy/mail.py
rennerocha: Fixed. I noticed that we don't have a `bcc` field as well. Maybe it would be interesting to include this new argument, what do you think?

scrapy/mail.py
```diff
@@ -66,13 +56,13 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
 
         msg['From'] = self.mailfrom
         msg['To'] = COMMASPACE.join(to)
-        msg['Date'] = formatdate(localtime=True)
-        msg['Subject'] = subject
-        rcpts = to[:]
+
         if cc:
-            rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
```
rennerocha: `cc` is back!

scrapy/mail.py
rennerocha: Not needed. Only if we want to have an alternative text version of the e-mail content together with the HTML version. I don't think it is needed here.

scrapy/mail.py
rennerocha: `cc` is being used again in `create_email_message` function.

scrapy/mail.py
rennerocha: Not sure if I understood the way you are thinking to use kwargs in this situation. Are you talking in something like this?

```
    def __init__(self, debug=False, **kwargs):
        self.aws_access_key = kwargs.get('MAIL_SES_AWS_ACCESS_KEY')
        self.aws_secret_key = kwargs.get('MAIL_SES_AWS_SECRET_KEY')
        self.mailfrom = kwargs.get('MAIL_FROM', 'scrapy@localhost')
        self.debug = debug

    @classmethod
        return cls(**settings)
```

kmike: I meant 

```py
return cls(
    aws_access_key=settings['MAIL_SES_AWS_ACCESS_KEY'],
    aws_secret_key=settings['MAIL_SES_AWS_SECRET_KEY'],
    mailfrom=settings['MAIL_FROM']
)
```

rennerocha:
@dangra , yes, this is the idea. If no other issues are pointed out in the current implementation, I will improve testing coverage and write the docs.

wRAR:
Closing together with #6612

PR #3347, opened 2018-07-20
Doc: update copyright notice

korigod:
The years have been updated. The hyphen was replaced with an en dash.


kmike:
Thanks @korigod!

PR #3348, opened 2018-07-20
[MRG+1] Add from_crawler constructor for feed exporters and storages

elacuesta:
Picking up #1605 to add some changes requested by @kmike


tests/test_feedexport.py
```diff
@@ -130,13 +133,49 @@ def test_invalid_folder(self):
...
     @defer.inlineCallbacks
     def test_store(self):
         assert_aws_environ()
         uri = os.environ.get('S3_TEST_FILE_URI')
         if not uri:
             raise unittest.SkipTest("No S3 URI available for testing")
-        storage = S3FeedStorage(uri)
+        access_key = os.environ.get('AWS_ACCESS_KEY_ID')
+        secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
+        storage = S3FeedStorage(uri, access_key, secret_key)
```
elacuesta: @kmike I believe this test is skipped, that's why passing `access_key=Settings()` didn't break. Is `S3_TEST_FILE_URI` available here?

scrapy/extensions/feedexport.py
```diff
@@ -93,12 +93,29 @@ def store(self, file):
...
+                import warnings
+                from scrapy.exceptions import ScrapyDeprecationWarning
+                warnings.warn(
+                    "Initialising `scrapy.extensions.feedexport.S3FeedStorage` "
+                    "without AWS keys is deprecated. Please supply credentials or "
+                    "use the `from_crawler()` constructor.",
+                    category=ScrapyDeprecationWarning,
+                    stacklevel=2
+                )
+                access_key = settings['AWS_ACCESS_KEY_ID']
```
elacuesta: Just checking: given that the operator here is an `or`, a `KeyError` might raise if one of the keys is missing, I guess that's expected? User code will break with a clear message I think.

kmike: Settings object returns None if a key is missing (see https://github.com/scrapy/scrapy/blob/258121db044c212493560dc5529b4355b148dae8/scrapy/settings/__init__.py#L93)

tests/test_feedexport.py
kmike: It is supposed to be set in the environment which runs tox - see https://github.com/scrapy/scrapy/blob/258121db044c212493560dc5529b4355b148dae8/tox.ini#L20. 

AFAIK we're not running s3 tests on travis if they require real s3 objects, i.e. this env variable is not set on Travis.

kmike:
Looks good to me, thanks for picking it up @elacuesta, and thanks @jdemaeyer for the initial PR!

scrapy/extensions/feedexport.py
elacuesta: In that case, I _think_ this might still break in the `boto` side if one of `access_key`/`secret_key` is not set either in the settings or in `uri`, the error message there will probably be clear enough.

PR #1455, opened 2015-08-24
[WIP] POC for asyncio+aiohttp

dangra:
*\* DO NOT MERGE - Strongly toxic and cool drug here **

Integration of a http downloader handler based on [aiohttp](https://github.com/KeepSafe/aiohttp) and asyncio on python3.

~Behind the scenes it hooks asyncio loop as a twisted reactor (provided by [txtulip](https://github.com/itamarst/txtulip)).~
**Update:** After #4294 it's possible to enable the asyncio-based reactor with the `TWISTED_REACTOR` setting.

Tests are not supposed to pass, this is just a working proof of concept!

Clone [testspiders's scrapy project](https://github.com/scrapinghub/testspiders) and run `scrapy crawl justfollow -s TWISTED_REACTOR=twisted.internet.asyncioreactor.AsyncioSelectorReactor`, see how smoothly it hits websites :)

```
$ scrapy crawl justfollow -s TWISTED_REACTOR=twisted.internet.asyncioreactor.AsyncioSelectorReactor
2020-01-20 09:34:22 [scrapy.utils.log] INFO: Scrapy 1.8.0 started (bot: testspiders)
2020-01-20 09:34:22 [scrapy.utils.log] INFO: Versions: lxml 4.2.6.0, libxml2 2.9.9, cssselect 1.0.3, parsel 1.5.1, w3lib 1.19.0, Twisted 18.9.0, Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 26 2018, 19:50:54) - [GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)], pyOpenSSL 18.0.0 (OpenSSL 1.1.0i  14 Aug 2018), cryptography 2.3.1, Platform Darwin-18.7.0-x86_64-i386-64bit
2020-01-20 09:34:22 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.asyncioreactor.AsyncioSelectorReactor
2020-01-20 09:34:22 [scrapy.crawler] INFO: Overridden settings:
{'TWISTED_REACTOR': 'twisted.internet.asyncioreactor.AsyncioSelectorReactor',
 'BOT_NAME': 'testspiders',
 'CLOSESPIDER_PAGECOUNT': 1000,
 'CLOSESPIDER_TIMEOUT': 3600,
 'COOKIES_ENABLED': False,
 'EDITOR': 'nano',
 'NEWSPIDER_MODULE': 'testspiders.spiders',
[...]

PR #3362, opened 2018-08-02
Asyncio support for Scrapy | Gsoc 2018

yashrsharma44:
I have made the initial pull request, I am working on the requirements as we discussed, and I will update this PR, as things are completed @dangra .


yashrsharma44:
Hey @dangra, I have made the following changes - 
1. Cleaning of the code base, removing any unnecessary comments
2. Allowing Fetch to accept a request parameter.
3. Enabled the request to be passed through spider middlewares( though Fetch is an awaitable, so it cannot be passed through spidermiddlewares)
4. Sharing the crawler and spider object implicitly.

scrapy/utils/ossignal.py
```diff
@@ -2,6 +2,14 @@
...
 import signal
 
+import asyncio
+'''
+from twisted.internet import asyncioreactor
+try:
+    asyncioreactor.install(asyncio.get_event_loop())
+except Exception:
+    pass
+'''
```
dangra: what is the point of importing `asyncio` here?

scrapy/Fetch/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from scrapy.core import engine
+
+from scrapy.utils.log import logformatter_adapter, failure_to_exc_info
+from scrapy import signals
+from scrapy.core.downloader import Downloader 
+from twisted.internet.defer import Deferred
+import asyncio
+import logging
+import scrapy
+import warnings
```
dangra: See https://www.python.org/dev/peps/pep-0008/#imports for import standard rules.

Specially ordering:

    1. Standard library imports.
    2. Related third party imports.
    3. Local application/library specific imports.

scrapy/core/downloader/__init__.py
```diff
@@ -6,7 +6,16 @@
...
 import six
-from twisted.internet import reactor, defer, task
+import asyncio
+'''
+from twisted.internet import asyncioreactor
+try:
+    asyncioreactor.install(asyncio.get_event_loop())
+except Exception:
+    pass
+'''
```
dangra: why are you importing `asyncio` in many different places? If it needs to be imported before anything else please put it on `scrapy/__init__.py` only.

yashrsharma44:
Hey @kmike, I was looking to get this code merged, could you provide updates, if the code is suitable/unsuitable to be merged, leaving aside tests and documentation? If it is suitable, I would gladly complete the tests and documentation.

scrapy/Fetch/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+from scrapy import signals
+from scrapy.core.downloader import Downloader 
+
+
+
+logger = logging.getLogger(__name__)
+
+
+
+class Fetch():
```
kmike: I think we should make Request class to be awaitable, not create a separate Fetch class.

scrapy/Fetch/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, Request):
+        self.Request = Request
+        self.crawler = engine.concraw
+        self.spider = engine.conspid
```
kmike: There can be several Crawlers running at the same time, and several spiders; we can't use global variables. I think to access Crawler contextvars should be used.

scrapy/__main__.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
```
kmike: why is this import needed?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -8,6 +8,7 @@
 from six.moves.urllib.parse import urldefrag
 
 from zope.interface import implementer
+from twisted.internet import reactor
```
kmike: why is this import needed?

scrapy/core/engine.py
```diff
@@ -4,29 +4,39 @@
...
 from scrapy.http import Response, Request
 from scrapy.utils.misc import load_object
+from scrapy.utils.defer import mustbe_deferred
 from scrapy.utils.reactor import CallLaterOnce
 from scrapy.utils.log import logformatter_adapter, failure_to_exc_info
 
 logger = logging.getLogger(__name__)
 
+conspid
+concraw
```
kmike: stray lines?

scrapy/spidermiddlewares/depth.py
```diff
@@ -54,5 +56,11 @@ def _filter(request):
             response.meta['depth'] = 0
             if self.verbose_stats:
                 self.stats.inc_value('request_depth_count/0', spider=spider)
-
-        return (r for r in result or () if _filter(r))
+        result = list_to_simplevalue(result)
+        if isinstance(result, GeneratorType):
```
kmike: It'd be much better if the feature doesn't require modifications in existing middlewares; there must be a way to support it in a framework. Otherwise users will have to modify all their custom middlewares in a similar way.

kmike:
Hey @yashrsharma44! Unfortunately, I think the current code is very far from being merged :( There can be some useful parts - maybe some of the modifications to the `inline_requests` library, or helpers like `asyncfut_parallel` - but it is very hard for me to assess, because other code needs a rewrite, and it is not clear if these helpers will be useful after a rewrite or not.

scrapy/Fetch/__init__.py
yashrsharma44: Regarding awaiting Request class, I was having circular import issues.

kneufeld:
I'm repeating a comment I left on #1455 in case it's more relevant here:

My friend @meejah wrote https://github.com/crossbario/txaio which is `a helper library for writing code that runs unmodified on both Twisted and asyncio`

This may help bridge the twisted/asyncio/python2/python3 gaps?

nctl144:
Hi @kmike, do you think I should make some more progress on this PR?

yashrsharma44:
My 2cents. I was stuck with using contextvars, so that can be looked into.

scrapy/crawler.py
```diff
@@ -2,11 +2,25 @@
 import signal
 import logging
 import warnings
+try:
+    import asyncio
+except ImportError:
+    warnings.warn("Please install asyncio to support asyncio frameworks. Without asyncio, frameworks might not be supported!")
```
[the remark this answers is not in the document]
nctl144: @kmike I suppose we should change it to "this project is using Python < 3.4" right?

wRAR:
This is no longer needed but was useful, thanks.

Issue #3145, opened 2018-02-27
Make public attribute formdata for FormRequest


kalombos:
I use **FormRequest.from_response** method with custom formdata parameter but it is not enough flexible for me. Because some websites(.aspx) require special order of formdata parameters. Here is my workaround:

```
import scrapy 

class CustomFormRequest(scrapy.FormRequest):
    def __init__(self, *args, **kwargs):
        if 'formdata' in kwargs:
            self.formdata = kwargs['formdata']
        super(FormRequest, self).__init__(*args, **kwargs)


def some_request(self, response):
    form_request = CustomFormRequest.from_response(response)
    new_formdata = make_new_formdata_with right_order(form_request.formdata)
    yield CustomFormRequest(url=form_request.url, formdata=new_formdata)
```
So if formdata would be a public attribute i would not have to create CustomFormRequest class. I can make a PR if my proposal would be approved.

kmike:
How does `make_new_formdata_with_right_order` work, and what order do you need? I wonder if we can have fields ordered in a correct way by default.

kalombos:
Here is part of a code from my case:

``
    from collections import OrderedDict
   
    form_request = CustomFormRequest.from_response(response)
   
    # here is make_new_formdata_with_right_order part
    prefix_data = [('ctl00$ContentPlaceHolder1$sm1', 'SOME_LONG_VALUE')]
    new_formdata = OrderedDict(prefix_data + form_request.formdata)
    new_formdata['__EVENTTARGET'] = 'ctl00$ContentPlaceHolder1$gvSearchResults$ctl03$ctl00'
    del new_formdata['ctl00$ContentPlaceHolder1$btnReset']
    del new_formdata['ctl00$ContentPlaceHolder1$btnSearch']

    yield CustomFormRequest(url=form_request.url, formdata=new_formdata)
``

As you can see i need an opportunity to insert some data at the begin of request(but in the common case it can be any part of request)  and i need an opportunity to remove some data from the request.

kmike:
I think a solution could be to provide a function which extracts formdata without creating FormRequest. You then can use this function, post-process the result and pass it to FormRequest. FormRequest.from_response should call this function as well. Maybe there should be several functions, for different tasks. 

[...]

PR #3363, opened 2018-08-03
Make public attribute formdata for FormRequest

kalombos:
#3145

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -16,6 +16,9 @@ from scrapy.http.common import obsolete_setter
 
 class Request(object_ref):
 
+    replaced_attributes = ['url', 'method', 'headers', 'body', 'cookies', 'meta', 'flags',
+                           'encoding', 'priority', 'dont_filter', 'callback', 'errback']
+
     def __init__(self, url, callback=None, method='GET', headers=None, body=None,
                  cookies=None, meta=None, encoding='utf-8', priority=0,
                  dont_filter=False, errback=None, flags=None):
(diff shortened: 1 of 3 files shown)


scrapy/http/request/__init__.py
```diff
@@ -16,6 +16,9 @@
 
 class Request(object_ref):
 
+    replaced_attributes = ['url', 'method', 'headers', 'body', 'cookies', 'meta', 'flags',
```
dangra: Considering the context I'd go for `replaceable_attributes` name and a comment explaining this are the attribute list that get propagated on `.replace()` call.

scrapy/http/request/form.py
```diff
@@ -35,6 +44,8 @@ def __init__(self, *args, **kwargs):
             else:
                 self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
 
+    formdata = property(_get_formdata, obsolete_setter(_set_formdata, 'formdata'))
```
kmike: why should we use obsolete_setter in new code?

scrapy/http/request/__init__.py
kmike: Also, what do you think about making this list either private (`_replacable_attributes`), or upper-cased, or both?

scrapy/http/request/form.py
kmike: I think there is a problem with API: obvious `request.formdata['foo'] = 'bar'` won't work as expected. And if we want to make it work, a naive solution can be bad for the performance (body can be rebuilt after each change to formdata).

PR #3387, opened 2018-08-15
Use Request(url, dont_filter=True)  instead of make_requests_from_url

Anderson-Liu:
use Request(url, dont_filter=True)  instead of make_requests_from_url

### DIFF
--- a/scrapy/spiders/__init__.py
+++ b/scrapy/spiders/__init__.py
@@ -76,11 +76,8 @@ class Spider(object_ref):
                     cls.__module__, cls.__name__
                 ),
             )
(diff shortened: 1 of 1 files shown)


kmike:
Hey @Anderson-Liu! 

`Request(url, dont_filter=True)` is used if an user writes spider in a modern way, `make_requests_from_url` is only called when an user has overridden this method; doing so ensures that old code works. I think it is not a right time yet to drop this backwards compatibility code.

Anderson-Liu:
Ok, @kmike  you are right, thanks!

PR #3394, opened 2018-08-17
Make lazy loading Download Handlers optional

hcoura:
Download Handlers load only when first request for a scheme is needed, this was put in place because s3 download handler used to be very slow to create.

This makes it optional for Download Handlers to be lazy loaded which is relevant when writing custom handlers that have required settings and wants to raise `NotConfiguredException` during instantiation.

### DIFF
--- a/scrapy/core/downloader/handlers/__init__.py
+++ b/scrapy/core/downloader/handlers/__init__.py
@@ -24,6 +24,13 @@ class DownloadHandlers(object):
             crawler.settings.getwithbase('DOWNLOAD_HANDLERS'))
         for scheme, clspath in six.iteritems(handlers):
             self._schemes[scheme] = clspath
+        for scheme in self._schemes:
+            path = self._schemes[scheme]
+            dhcls = load_object(path)
+            lazy = getattr(dhcls, 'lazy', False)
+            if lazy:
+                continue
+            self._load_handler(scheme, dhcls)
 
         crawler.signals.connect(self._close, signals.engine_stopped)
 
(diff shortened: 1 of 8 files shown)


scrapy/core/downloader/handlers/__init__.py
```diff
@@ -40,8 +47,12 @@ def _get_handler(self, scheme):
             return None
 
         path = self._schemes[scheme]
+        dhcls = load_object(path)
+        self._load_handler(scheme, dhcls)
+        return self._handlers[scheme]
+
+    def _load_handler(self, scheme, dhcls):
```
kmike: could you please check the logging.error call below? `_load_handler` method doesn't have `path` as an argument, but uses it in a message. Probably the idea was to catch load_object errors as well?

scrapy/core/downloader/handlers/__init__.py
```diff
@@ -24,6 +24,13 @@ def __init__(self, crawler):
             crawler.settings.getwithbase('DOWNLOAD_HANDLERS'))
         for scheme, clspath in six.iteritems(handlers):
             self._schemes[scheme] = clspath
+        for scheme in self._schemes:
```
kmike: I think code can be simplified if you do this all in a loop above

scrapy/core/downloader/handlers/__init__.py
```diff
@@ -24,6 +24,13 @@ def __init__(self, crawler):
...
+            path = self._schemes[scheme]
+            dhcls = load_object(path)
+            lazy = getattr(dhcls, 'lazy', False)
```
kmike: Technically, lazy=False by default is backwards incompatible for custom user download handlers. What do you think about keeping it as True by default, but adding lazy=False for default handlers other than s3?

hcoura: Yeah, I agree

scrapy/core/downloader/handlers/__init__.py
hcoura: Indeed, trying to rush things at the end of the day on Friday is never a good thing.

PR #3409, opened 2018-08-29
FIX rewrite invalid user-agent middleware method

KaLZeK:
because headers does not actually has the setdefault method, here if we directly give value to the [‘user-agent’], things work out fine

### DIFF
--- a/scrapy/downloadermiddlewares/useragent.py
+++ b/scrapy/downloadermiddlewares/useragent.py
@@ -20,4 +20,4 @@ class UserAgentMiddleware(object):
 
     def process_request(self, request, spider):
         if self.user_agent:
-            request.headers.setdefault(b'User-Agent', self.user_agent)
+            request.headers[b'User-Agent'] = self.user_agent


KaLZeK:
That is exactly why I posted this PR. I made a project myself and my middleware is like below(I need my project to switch user agent from time to time) :
···
class RandomUserAgentMiddleware(object):
    '''
    随机选择user-agent的中间件
    '''
    def __init__(self):
        from project.conf.user_agent import USER_AGENT_LIST
        self.user_agent_list = USER_AGENT_LIST 
        self.customized_user_agent = None

    @classmethod
        obj = cls()
        crawler.signals.connect(obj.spider_opened, signal=signals.spider_opened)
        return obj

        # this is user customized user_agent, assign it in Spider.__init__
        self.customized_user_agent = getattr(spider, 'user_agent', self.customized_user_agent)

        if self.customized_user_agent:
            request.headers[b'User-Agent'] = self.customized_user_agent
        else: 
            request.headers[b'User-Agent'] = random.choice(self.user_agent_list)
        print("request.headers:", request.headers)
···
And this works perfectly fine. But when I rewrite 
···
request.headers[b'User-Agent'] = random.choice(self.user_agent_list)
···
to
···
request.headers.setdefault([b'User-Agent'], random.choice(self.user_agent_list))
···

It simply won't work.

PR #3415, opened 2018-09-05
[MRG+1] Telnet console authentication

dangra:
This PR adds username/password authentication to Telnet console extension.

ToDo:
- [x] Implement username/password auth
- [x] Implement random password by default (?)
- [x] Add test case
- [x] Update docs


nyov:
I don't see the point. :trollface:

Okay I apologize for raining on this parade, don't take it as an argument, I just couldn't walk away :|
I'm _still_ ambivalent about scrapy launching a telnet server by default. So I should be glad about this, but I'm afraid of what the availability of this feature could suggest to users.

If you add authentication to telnet, it stands to reason you do this to expose it to an insecure network (or 'suggesting' to people that it can be used that way).
If you do "telnet-like" over an insecure network in this decade (of deep packet inspection), you'll want transport encryption to actually safeguard your authentication data or the password "protection" is somewhat academic really. (Perhaps a molly-guard?)
This means what you really want is an SSH console, because if you were using that telnet over a VPN, you would have authentication via the VPN already and not need it in your telnet console.
(I suppose a TLS wrapper for the tcp connection is also possible protection, if less userfriendly if you want users to fiddle with e.g. openssl s_client + netcat.)

In my worthless opinion, trying to shoehorn security measures into the console here is out of scope for scrapy. It takes time maintaining, likely falls short at security, and will only lead to more "improvement" work down the road[1]: Does this code log any warnings on bad login tries? Will it lock out any? Would it throttle connection attempts? Or does the use of Twisted here only mean it would happily let me run a dictionary attack on 20 connections in parallel?

Dedicated software for doing this exists (ssh, vpn, et al), it's only less comfortable to make people use it than pointing at a simple built-in config option.

[1] Such as handling [stupid CVE's](https://security-tracker.debian.org/tracker/CVE-2017-14158) in optional dependencies that are unlikely to be actually relevant for most users, but might prevent maintainers from packaging the software... Only goes to show that some features are better served as optional extensions.
(I mean, S3! Really? A child should understand that uploading data to a slow pipe, that you download at much higher bandwidth and concurrency, can lead to clogged pipe issues. How is that even a remote attack vector or high severity, if the scrapy service isn't compromised. I'm ranting again, aren't I?)

nyov:
So, after this stellar contribution :stuck_out_tongue_winking_eye: :sunglasses:, I shall at least go to the trouble and explain how I can handle such stuff with SSH as a cheap tunnel.

On my client, I create a new ssh key-pair (identity), e.g. `~/.ssh/id_rsa.scrapy[.pub]` using `ssh-keygen`. (Which is _not_ my normal login identity for the machine.)
Then I setup `~/.ssh/config` with my scrapy server alias, to use this identity:
```
Host scrapy
	HostName 127.0.0.1
	Port 1111
	User scrapy
	IdentityFile ~/.ssh/id_rsa.scrapy
```
and also make sure the SSH client doesn't offer all it's known keys to any server I connect to, but only the `IdentityFile` I set up, otherwise it would login with the wrong identity in one case (depending on the order in which ssh would offer the keys) _and anyone should do this anyway, since that SSH habit is a nasty privacy leak when the server secretly collects such info_:
```
Host *
	IdentitiesOnly yes
```

Finally I add the new identity's public key to my remote scrapy server into `~/.ssh/authorized_keys`,
beside my regular login identity (which uses a different key):
```
ssh-rsa AAAA[...]= regular login key
command="/bin/echo \"est()\" | /bin/netcat -N 127.0.0.1 6023",no-agent-forwarding,no-port-forwarding,no-user-rc,no-X11-forwarding ssh-rsa AAAA[...]= scrapy shell login
```
And here's the magic happening. Depending on the identity I authenticate with, the `command` line relating to that identity is executed.
For example this status query to a running scrapy telnet console. Or a telnet client which does no longer need an auth now, as I stick it to localhost only.

For more awesome with `authorized_keys` see [the man page](https://man.openbsd.org/sshd#AUTHORIZED_KEYS_FILE_FORMAT).

Usually of course I put that stuff into a script which I then reference in `command`.
Don't forget to use absolute paths since we have no shell with a `PATH` info here.

So, to run a scrapy shell session I could do...
```
command="/usr/bin/toilet -w100 -F border:gay -f pagga 'Scrape the Planet!'; /usr/bin/scrapy shell",no-agent-forwarding,no-port-forwarding,no-user-rc,no-X11-forwarding ssh-rsa AAAA[...]= scrapy shell login
```
and that would look like this:

scrapyplanet

Perhaps it can be useful info to someone.

dangra:
hi @nyov, we (@kmike and me) think your point is valid and your explanation on how to wrap the telnet console with ssh is wonderful to the point it deserves a section in our docs.

Considering the following comment: 

> If you add authentication to telnet, it stands to reason you do this to expose it to an insecure network (or 'suggesting' to people that it can be used that way). 

We are going to add a bold warning in docs stating the console is not meant to be used under public networks and can't be considered secure. That being said, I'd like to remember that by default Scrapy's telnet console binds to 127.0.0.1 only.

We have our reasons to add a login to telnet console, it happens that under certain situations browsers can be tricked to send requests to localhost, and the console as it stands it is a real [manhole](https://twistedmatrix.com/documents/current/api/twisted.conch.manhole.html)  for your toilet banner.

kmike:
Hey @dangra! This PR looks good, but I think we should make a docs update as a part of this PR.

Issue #3392, opened 2018-08-17
dh key too small


zyx648951430:
url:    “https://portal.hyit.edu.cn/zfca/login”
error:  ERROR: Error downloading <GET https://portal.hyit.edu.cn/zfca/login>: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'tls_process_ske_dhe', 'dh key too small')

wRAR:
According to the [SSL Labs test](https://www.ssllabs.com/ssltest/analyze.html?d=portal.hyit.edu.cn) the website is indeed using an insecure HTTPS configuration, and OpenSSL refuses to talk to it.

I couldn't find a way to disable this particular check in OpenSSL, but disabling DH ciphers in OpenSSL helps with at least this website. To do that in Scrapy you need to subclass `scrapy.core.downloader.contextfactory.ScrapyClientContextFactory`, replacing `DEFAULT_CIPHERS` with `AcceptableCiphers.fromOpenSSLCipherString('DEFAULT:!DH')` and then set the [`DOWNLOADER_CLIENTCONTEXTFACTORY`](https://doc.scrapy.org/en/latest/topics/settings.html#downloader-clientcontextfactory) setting to this new class. It's probably worth making the cipher string a Scrapy setting if that works, but for now you'll need to do this.

zyx648951430:
Thank you very much !!!  According to your suggestion, I already have the solution.

Gallaecio:
This is now configurable through a setting thanks to @wRAR’s #3442.

PR #3442, opened 2018-10-02
Support for overriding OpenSSL ciphers

wRAR:
This adds the `DOWNLOADER_CLIENT_TLS_CIPHERS` setting which is then used in `ScrapyClientContextFactory`. There is a real-world example of when it may be needed in #3392 (the server offers a cipher that is rejected by the client so we need to not advertise it), there are other theoretical  problems e.g. a server supports only ciphers that are supported by the client but disabled. I've constructed the second case in the tests because constructing the first case seems to be impossible with a modern OpenSSL.
There are two tests, the one in `tests/test_webclient.py` uses `ScrapyClientContextFactory` directly (its base class also just tests the default context factory with the default SSL server) and the one in `tests/test_downloader_handlers.py` uses the setting and the download handler. I don't know if the first one is worth keeping. I also don't like the amount of duplicated code in the added test cases, maybe some reshuffling of base HTTP(S) test classes will be useful.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -445,6 +445,24 @@ accepts a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
 :setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``tls_verbose_logging``
 parameter (``bool``).
 
+.. setting:: DOWNLOADER_CLIENT_TLS_CIPHERS
+
+DOWNLOADER_CLIENT_TLS_CIPHERS
+-----------------------------
+
+Default: ``'DEFAULT'``
+
+Use  this setting to customize the TLS/SSL ciphers used by the default
+HTTP/1.1 downloader.
+
+The setting should contain a string in the `OpenSSL cipher list format`_,
+these ciphers will be used as client ciphers. Changing this setting may be
+necessary to access certain HTTPS websites: for example, you may need to use
+``'DEFAULT:!DH'`` for a website with weak DH parameters or enable a
+specific cipher that is not included in ``DEFAULT`` if a website requires it.
+
+.. _OpenSSL cipher list format: https://www.openssl.org/docs/manmaster/man1/ciphers.html#CIPHER-LIST-FORMAT
+
 .. setting:: DOWNLOADER_CLIENT_TLS_METHOD
 
 DOWNLOADER_CLIENT_TLS_METHOD
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -442,8 +442,9 @@ or even enable client-side authentication (and various other things).
 
 If you do use a custom ContextFactory, make sure its ``__init__`` method
 accepts a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
-:setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``tls_verbose_logging``
-parameter (``bool``).
+:setting:`DOWNLOADER_CLIENT_TLS_METHOD`), a ``tls_verbose_logging``
+parameter (``bool``) and a ``tls_ciphers`` parameter (see
+:setting:`DOWNLOADER_CLIENT_TLS_CIPHERS`).
 
 .. setting:: DOWNLOADER_CLIENT_TLS_CIPHERS
 
--- a/scrapy/core/downloader/contextfactory.py
+++ b/scrapy/core/downloader/contextfactory.py
@@ -1,5 +1,5 @@
 from OpenSSL import SSL
-from twisted.internet.ssl import optionsForClientTLS, CertificateOptions, platformTrust
+from twisted.internet.ssl import optionsForClientTLS, CertificateOptions, platformTrust, AcceptableCiphers
 from twisted.web.client import BrowserLikePolicyForHTTPS
 from twisted.web.iweb import IPolicyForHTTPS
 from zope.interface.declarations import implementer
@@ -19,15 +19,20 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
      understand the SSLv3, TLSv1, TLSv1.1 and TLSv1.2 protocols.'
     """
 
-    def __init__(self, method=SSL.SSLv23_METHOD, tls_verbose_logging=False, *args, **kwargs):
+    def __init__(self, method=SSL.SSLv23_METHOD, tls_verbose_logging=False, tls_ciphers=None, *args, **kwargs):
         super(ScrapyClientContextFactory, self).__init__(*args, **kwargs)
         self._ssl_method = method
         self.tls_verbose_logging = tls_verbose_logging
+        if tls_ciphers:
+            self.tls_ciphers = AcceptableCiphers.fromOpenSSLCipherString(tls_ciphers)
+        else:
+            self.tls_ciphers = DEFAULT_CIPHERS
 
     @classmethod
     def from_settings(cls, settings, method=SSL.SSLv23_METHOD, *args, **kwargs):
         tls_verbose_logging = settings.getbool('DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING')
-        return cls(method=method, tls_verbose_logging=tls_verbose_logging, *args, **kwargs)
+        tls_ciphers = settings['DOWNLOADER_CLIENT_TLS_CIPHERS']
+        return cls(method=method, tls_verbose_logging=tls_verbose_logging, tls_ciphers=tls_ciphers, *args, **kwargs)
 
     def getCertificateOptions(self):
         # setting verify=True will require you to provide CAs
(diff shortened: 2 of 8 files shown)


scrapy/core/downloader/contextfactory.py
```diff
@@ -28,9 +29,13 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
...
         """
 
-        def __init__(self, method=SSL.SSLv23_METHOD, *args, **kwargs):
+        def __init__(self, method=SSL.SSLv23_METHOD, ciphers_string=None, *args, **kwargs):
             super(ScrapyClientContextFactory, self).__init__(*args, **kwargs)
             self._ssl_method = method
+            if ciphers_string:
+                self._ssl_ciphers = AcceptableCiphers.fromOpenSSLCipherString(ciphers_string)
+            else:
+                self._ssl_ciphers = DEFAULT_CIPHERS
```
wRAR: `DEFAULT_CIPHERS` is a singleton so I kept it as it maybe improves the performance.

tests/mockserver.py
```diff
@@ -222,6 +224,14 @@ def ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.c
          )
 
 
+def broken_ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.crt', cipher_string='DEFAULT'):
+    factory = ssl_context_factory(keyfile, certfile)
+    ctx = factory.getContext()
+    ctx.set_options(SSL.OP_CIPHER_SERVER_PREFERENCE | SSL.OP_NO_TLSv1_2)
```
wRAR: TLS 1.2 has it's own always enabled ciphers so I disable it.

Twisted recommends using `CertificateOptions` instead of `DefaultOpenSSLContextFactory`, and its API is cleaner, but it requires file-like objects for certs and also I don't know if all supported Twisted versions support this API.

tests/mockserver.py
```diff
@@ -222,6 +224,14 @@ def ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.c
          )
 
 
+def broken_ssl_context_factory(keyfile='keys/localhost.key', certfile='keys/localhost.crt', cipher_string='DEFAULT'):
```
wRAR: Maybe it should be renamed.

tests/test_webclient.py
```diff
@@ -363,3 +365,47 @@ def _check_Encoding(self, response, original_body):
...
+        yield self.port.stopListening()
+        shutil.rmtree(self.tmpname)
+
+    def testPayload(self):
+        s = "0123456789" * 10
+        return getPage(self.getURL("payload"), body=s).addCallback(
+            self.assertEqual, to_bytes(s))
+
+
+class WebClientBrokenSSLTestCase(WebClientSSLTestCase):
```
wRAR: Maybe it should be renamed.

docs/topics/settings.rst
```diff
@@ -431,6 +431,24 @@ parameter at init (this is the ``OpenSSL.SSL`` method mapping
 
 .. setting:: DOWNLOADER_CLIENT_TLS_METHOD
 
+DOWNLOADER_CLIENT_TLS_CIPHERS
+-----------------------------
+
+Default: ``'DEFAULT'``
```
kmike: I wonder if we should use a list of ciphers which browsers use as a default, instead of using OpenSSL defaults, and provide another default value ("OPENSSL_DEFAULT"). This is a larger task, and I have no idea where to get these defaults, but it could make Scrapy work as expected in more cases. AFAIK in OpenSSL defaults are really for the server side, these defaults are more restrictive than browsers'. 

Also, if I recall correctly, some of the ciphers which browsers accept are disabled at compile time in OpenSSL, which is an additional challenge.

kmike:
Awesome, this is a much-needed feature which may solve quite a lot of issues in Scrapy bug tracker 👍

tests/mockserver.py
```diff
@@ -3,6 +3,8 @@
 from six.moves.urllib.parse import urlencode
 from subprocess import Popen, PIPE
 
+from OpenSSL import SSL
+
```
kmike: a nitpick: this import should be with twisted imports according to pep8 (3 groups: stdlib imports, third-party imports, internal imports)

tests/mockserver.py
kmike: `CertificateOptions` was there at least in twisted 9.0.0 (https://twistedmatrix.com/documents/9.0.0/api/twisted.internet.ssl.CertificateOptions.html), so it should be fine to use. But I have no idea what I'm talking about, or what is better in this particular case :)

tests/mockserver.py
kmike: Could you please explain what this code does? Does it 1) add cipher_string support, and 2) disables default ciphers which come from TLSv1_2? And the purpose is to create servers which use outdated ciphers not supported by current http clients by default?

If so, what do you think about 

1) removing this function,
2) adding cipher_string argument to ssl_context_factory, None by default,
3) when cipher_string is not None, ssl_context_factory becomes broken_ssl_context_factory,
4) add a comment which explains why `ctx.set_options(SSL.OP_CIPHER_SERVER_PREFERENCE | SSL.OP_NO_TLSv1_2)` line exists.

tests/test_downloader_handlers.py
```diff
@@ -520,6 +520,47 @@ def setUp(self):
...
+        r = static.File(self.tmpname)
+        self.site = server.Site(r, timeout=None)
+        self.wrapper = WrappingFactory(self.site)
+        self.host = 'localhost'
+        self.port = reactor.listenSSL(
+            0, self.wrapper, broken_ssl_context_factory(self.keyfile, self.certfile, cipher_string='CAMELLIA256-SHA'),
+            interface=self.host)
+        self.portno = self.port.getHost().port
+        self.download_handler = self.download_handler_cls(
+            Settings({'DOWNLOADER_CLIENT_TLS_CIPHERS': 'CAMELLIA256-SHA'}))
```
kmike: could you please also add a test which shows that downloading doesn't work with default DOWNLOADER_CLIENT_TLS_CIPHERS, to make sure the other test is valid?

docs/topics/settings.rst
kmike: Aha, DEFAULT actually came from OpenSSL, so let's keep it as-is. Browser-like defaults is a separate feature, let's first get DOWNLOADER_CLIENT_TLS_CIPHERS merged. 

My worry was that we may have to change "DEFAULT" to something else if we change the actual DOWNLOADER_CLIENT_TLS_CIPHERS default value to a browser-like, and so we should think about it now, but as "DEFAULT" comes from OpenSSL, it is better to keep it as-is.

scrapy/core/downloader/contextfactory.py
kmike: I'm not sure it helps, as download handler always pass ciphers_string, which is "DEFAULT" by default.

tests/test_webclient.py
kmike: BrokenSSL -> OldCiphers / InsecureCIphers?

wRAR:
The tests fail on the "jessie" env, I tried different pyOpenSSL and cryptography versions and found that the tests only work on cryptography 2.0+ (IIRC).

kmike:
@wRAR could you please check what cryptography versions are available in Debians and Ubuntus we care about?

wRAR:
@kmike 
Ubuntu Trusty 14.04 - not packaged, `python-openssl` (0.13) doesn't depend on it.
Ubuntu Xenial 16.04 - 1.2.3
Ubuntu Bionic 18.04 - 2.1.4
Debian 8 Jessie - 0.6.1
Debian 9 Stretch - 1.7.1
Debian 10 Buster - 2.6.1

We currently don't set the version constraints for pyOpenSSL or cryptography on envs other than jessie (and trusty, which we don't test anymore).

Note that cryptography 2.0 only works with pyOpenSSL 16.2+.

The failing test expects a connection error with a non-default server cipher and the default client setting but the connection works on these older libraries. I didn't see anything in the cryptography 2.0 changelog and didn't try to investigate further.

scrapy/core/downloader/contextfactory.py
```diff
@@ -54,7 +60,7 @@ def getCertificateOptions(self):
                         method=getattr(self, 'method',
                                        getattr(self, '_ssl_method', None)),
                         fixBrokenPeers=True,
-                        acceptableCiphers=DEFAULT_CIPHERS)
+                        acceptableCiphers=self.tls_ciphers)
```
Gallaecio: We should remove or deprecate the `DEFAULT_CIPHERS` import in this file, and probably also its definition in `scrapy/core/downloader/tls.py`.

wRAR: > We should remove or deprecate the `DEFAULT_CIPHERS` import in this file, and probably also its definition in `scrapy/core/downloader/tls.py`.

And move the default to `__init__`?

Gallaecio: I mean that `DEFAULT_CIPHERS`, after your change, is no longer used in this file, nor anywhere else in the Scrapy code base.

Gallaecio: And I am completely wrong about that :|

kmike:
Thanks @wRAR!

Issue #2726, opened 2017-05-06
Expose certificate information on HTTPS responses


alex:
For scraper I'm working on, I'd like to have access to the certificate supplied by the server for HTTPS requests. As far as I can tell, this is not currently exposed on either requests or responses.

redapple:
Hey @alex , sorry for the late feedback.
There's a related issue about logging TLS options, cipher suites (https://github.com/scrapy/scrapy/issues/2111)

I haven't really looked into how we could expose that, but it'd probably be somewhere in the [TLS options object](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/tls.py) that we pass to Twisted via [the context factory](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/contextfactory.py#L20).

PR #3450, opened 2018-10-05
Log cipher, certificate and temp key info on establishing an SSL connection

### LINKED ISSUE #2111: Add option to Log actual TLS connection options/Cipher Suite on 1st connection to a domain
See https://github.com/scrapy/scrapy/issues/1435#issuecomment-182393180 by @nyov for motivation

wRAR:
Fix #2111. Also related to #2726, though that ticket most likely asks for a programmatical access to all of this (I don't know if it's easily doable).

Output example:

> 2018-10-05 20:08:36 [scrapy.core.downloader.tls] DEBUG: SSL connection to market.yandex.ru using protocol TLSv1.2, cipher ECDHE-RSA-AES128-GCM-SHA256
> 2018-10-05 20:08:36 [scrapy.core.downloader.tls] DEBUG: SSL connection certificate: issuer "/C=RU/O=Yandex LLC/OU=Yandex Certification Authority/CN=Yandex CA", subject "/C=RU/O=Yandex LLC/OU=ITO/L=Moscow/ST=Russian Federation/CN=market.yandex.ru"
> 2018-10-05 20:08:36 [scrapy.core.downloader.tls] DEBUG: SSL temp key: ECDH, P-256, 256 bits

I'm not sure this should be enabled by default as a lot of websites are now HTTPS and this will be printed on all requests including redirects, but I don't know how to access settings in this class.

Note that there is a lot of direct FFI code to access the temporary key params, I hope it works correctly and doesn't cause memleaks. It also requires OpenSSL 1.0.2, but I couldn't test the check with an older OpenSSL, as I couldn't run tests on an actual jessie system.

There is other info that can be added, look at the `Connection` methods (easy) and at the `openssl s_client` output (may be harder, like with the temp key info).

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -438,9 +438,10 @@ or even enable client-side authentication (and various other things).
     which uses the platform's certificates to validate remote endpoints.
     **This is only available if you use Twisted>=14.0.**
 
-If you do use a custom ContextFactory, make sure it accepts a ``method``
-parameter at init (this is the ``OpenSSL.SSL`` method mapping
-:setting:`DOWNLOADER_CLIENT_TLS_METHOD`).
+If you do use a custom ContextFactory, make sure its ``__init__` method accepts
+a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
+:setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``settings`` parameter (this is
+the Scrapy settings object).
 
 .. setting:: DOWNLOADER_CLIENT_TLS_METHOD
 
@@ -467,6 +468,20 @@ This setting must be one of these string values:
 
     We recommend that you use PyOpenSSL>=0.13 and Twisted>=0.13
     or above (Twisted>=14.0 if you can).
+
+.. setting:: DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING
+
+DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING
+-------------------------------------
+
+Default: ``False``
+
+Setting this to ``True`` will enable DEBUG level messages about TLS connection
+parameters after establishing HTTPS connections. The kind of information logged
+depends on the versions of OpenSSL and pyOpenSSL.
+
+This setting is only used for the default
+:setting:`DOWNLOADER_CLIENTCONTEXTFACTORY`.
 
 .. setting:: DOWNLOADER_MIDDLEWARES
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -440,8 +440,8 @@ or even enable client-side authentication (and various other things).
 
 If you do use a custom ContextFactory, make sure its ``__init__`` method
 accepts a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
-:setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``settings`` parameter (this is
-the Scrapy :class:`~scrapy.settings.Settings` object).
+:setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``tls_verbose_logging``
+parameter (``bool``).
 
 .. setting:: DOWNLOADER_CLIENT_TLS_METHOD
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -438,10 +438,10 @@ or even enable client-side authentication (and various other things).
     which uses the platform's certificates to validate remote endpoints.
     **This is only available if you use Twisted>=14.0.**
 
-If you do use a custom ContextFactory, make sure its ``__init__` method accepts
-a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
+If you do use a custom ContextFactory, make sure its ``__init__`` method
+accepts a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
 :setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``settings`` parameter (this is
-the Scrapy settings object).
+the Scrapy :class:`~scrapy.settings.Settings` object).
 
 .. setting:: DOWNLOADER_CLIENT_TLS_METHOD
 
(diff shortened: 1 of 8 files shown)


scrapy/core/downloader/tls.py
```diff
@@ -65,13 +81,51 @@ class ScrapyClientTLSOptions(ClientTLSOptions):
...
+                             connection.get_cipher_name(),
+                             )
+                server_cert = connection.get_peer_certificate()
+                logger.debug('SSL connection certificate: issuer "%s", subject "%s"',
+                             x509name_to_string(server_cert.get_issuer()),
+                             x509name_to_string(server_cert.get_subject()),
+                             )
+
+                if hasattr(pyOpenSSLutil.lib, 'SSL_get_server_tmp_key'):  # requires OpenSSL 1.0.2
+                    # adapted from OpenSSL apps/s_cb.c::ssl_print_tmp_key()
```
kmike: hey! Wopuld you mind moving it to a function? Not logging itself, just collecting all the information.

kmike: I think we can create scrapy.util.ssl or scrapy.util.openssl with such helpers (x509name_to_string, etc.)

kmike:
> I'm not sure this should be enabled by default as a lot of websites are now HTTPS and this will be printed on all requests including redirects, but I don't know how to access settings in this class.

It seems you can access settings from HTTP11DownloadHandler and HTTP10DownloadHandler, and pass arguments to DOWNLOADER_CLIENTCONTEXTFACTORY, which can pass them to ScrapyClientTLSOptions. It can be also a good chance to introduce from_settings/from_crawler support for more Scrapy components, instead of passing just specific option values - basically, create objects using scrapy.utils.misc.create_instance instead of creating them directly.

docs/topics/settings.rst
```diff
@@ -438,9 +438,10 @@ or even enable client-side authentication (and various other things).
     which uses the platform's certificates to validate remote endpoints.
     **This is only available if you use Twisted>=14.0.**
 
-If you do use a custom ContextFactory, make sure it accepts a ``method``
-parameter at init (this is the ``OpenSSL.SSL`` method mapping
-:setting:`DOWNLOADER_CLIENT_TLS_METHOD`).
+If you do use a custom ContextFactory, make sure its ``__init__` method accepts
```
kmike:
```suggestion
If you do use a custom ContextFactory, make sure its ``__init__`` method accepts
```

docs/topics/settings.rst
```diff
@@ -438,9 +438,10 @@ or even enable client-side authentication (and various other things).
...
+a ``method`` parameter (this is the ``OpenSSL.SSL`` method mapping
+:setting:`DOWNLOADER_CLIENT_TLS_METHOD`) and a ``settings`` parameter (this is
+the Scrapy settings object).
```
Gallaecio: :lipstick: 

```suggestion
the Scrapy :class:`~scrapy.settings.Settings` object).
```

kmike:
@wRAR tests are failing for pypy/py27/jessie, and it looks like a genuine failure - could you please take a look?

kmike:
@wRAR can you please also add a test for the new option? I'm not sure it should be super-detailed and cover all code paths for all OpenSSL versions, but checking that after enabling this option crawling still works, and there are some messages which make sense, could be nice.

scrapy/core/downloader/contextfactory.py
```diff
@@ -28,9 +29,17 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
          understand the SSLv3, TLSv1, TLSv1.1 and TLSv1.2 protocols.'
         """
 
-        def __init__(self, method=SSL.SSLv23_METHOD, *args, **kwargs):
+        def __init__(self, method=SSL.SSLv23_METHOD, settings=None, *args, **kwargs):
             super(ScrapyClientContextFactory, self).__init__(*args, **kwargs)
             self._ssl_method = method
+            if settings:
+                self.tls_verbose_logging = settings['DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING']
```
kmike:
```suggestion
                self.tls_verbose_logging = settings.getbool('DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING')
```

Without it, `-s DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING=0` in command-line may be evaluated as True, because self.tls_verbose_logging will be '0'.

scrapy/core/downloader/contextfactory.py
```diff
@@ -28,9 +29,17 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
          understand the SSLv3, TLSv1, TLSv1.1 and TLSv1.2 protocols.'
         """
 
-        def __init__(self, method=SSL.SSLv23_METHOD, *args, **kwargs):
+        def __init__(self, method=SSL.SSLv23_METHOD, settings=None, *args, **kwargs):
```
kmike: I think a common practice is to pass tls_verbose_logging, not Settings instance, and extract option value in from_settings / from_crawler.

wRAR: I'm fine with changing this, though I don't like that we need to list all optional args in the documentation and the error message, as I was going to pass yet another setting here in a different PR, maybe we can rephrase the messages.

scrapy/core/downloader/contextfactory.py
```diff
@@ -29,17 +29,18 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
...
-            if settings:
-                self.tls_verbose_logging = settings['DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING']
-            else:
-                self.tls_verbose_logging = False
+            self.tls_verbose_logging = tls_verbose_logging
 
         @classmethod
         def from_settings(cls, settings, method=SSL.SSLv23_METHOD, *args, **kwargs):
-            return cls(method=method, settings=settings, *args, **kwargs)
+            if settings:
```
kmike: I think settings should always be passed here

kmike:
Thanks @wRAR for the fix and @Gallaecio for the review!

PR #3476, opened 2018-10-29
[MRG+1] Use collections.deque instead of list to store MiddlewareManager's methods

elacuesta:
According to https://wiki.python.org/moin/TimeComplexity `list.insert` has an average time complexity of O(n), while `deque.appendleft` is O(1).

From that page (emphasis mine): 

> Internally, a list is represented as an array; the largest costs come from growing beyond the current allocation size (because everything must move), or from **inserting or deleting somewhere near the beginning** (because everything after that must move). **If you need to add/remove at both ends, consider using a collections.deque instead.**

It seems like for deques "_looking at the middle is slow, and adding to or removing from the middle is slower still_", which should not be a problem given that these method containers are iterated sequentially and not accessed randomly.

I first saw this while working on #2061, I'll adapt that one if this PR gets merged.

### DIFF
--- a/scrapy/core/downloader/middleware.py
+++ b/scrapy/core/downloader/middleware.py
@@ -26,9 +26,9 @@ class DownloaderMiddlewareManager(MiddlewareManager):
         if hasattr(mw, 'process_request'):
             self.methods['process_request'].append(mw.process_request)
         if hasattr(mw, 'process_response'):
-            self.methods['process_response'].insert(0, mw.process_response)
+            self.methods['process_response'].appendleft(mw.process_response)
         if hasattr(mw, 'process_exception'):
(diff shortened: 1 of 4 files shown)


kmike:
I've checked on a small benchmark that iteration is indeed of about the same speed for list and deque:

```py
In [1]: from collections import deque
In [2]: lst = [x for x in range(100000)]
In [3]: dq = deque(lst)
In [8]: %timeit [x for x in dq]
2.51 ms ± 33.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [9]: %timeit [x for x in lst]
2.5 ms ± 42.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

PR #3480, opened 2018-10-31
 make FormRequests support the formdata parameter can be a string type

zssing:
When I use FormRequests the parameter formdata can not be urlencoded to the result I want.It could be a character encoding problem.So I make FormRequests support the formdata parameter can be a string type.Just put formdata as a string after urlencoded,Like "tabid=261&wmguid=20aae8dc-4a0c-4af5-aedf-cc153eb6efdf&p=87f11024-55ab-4faf-a0af-46371e33ae66"

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -25,10 +25,13 @@ class FormRequest(Request):
             kwargs['method'] = 'POST'
 
         super(FormRequest, self).__init__(*args, **kwargs)
-
         if formdata:
             items = formdata.items() if isinstance(formdata, dict) else formdata
-            querystr = _urlencode(items, self.encoding)
+            
+            if isinstance(formdata, str):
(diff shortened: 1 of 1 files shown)


scrapy/http/request/form.py
```diff
@@ -175,8 +175,9 @@ def _get_clickable(clickdata, form):
     """
     clickables = [
         el for el in form.xpath(
-            'descendant::input[re:test(@type, "^(submit|image)$", "i")]'
-            '|descendant::button[not(@type) or re:test(@type, "^submit$", "i")]',
```
wRAR, line 179: These are unrelated changes.

wRAR (changes requested):
Not sure how useful is this, as if you can create a string directly but not a dict that can be passed to FormRequest it sounds like a bug in your code. In any case the change should be documented and tests added.

PR #3485, opened 2018-11-08
[WIP] async def parse support

lopuhin:
This is extremely early, just to check the test failures. Supported syntax examples are in https://github.com/lopuhin/scrapy/blob/async-def-parse/tests/aio_spiders.py


wRAR:
This is no longer needed but was useful, thank you.

PR #3494, opened 2018-11-15
Add missing '/' when the URL path is empty

### LINKED ISSUE #1133: Scrapy doesn't normalize URLs returned in redirects which leads to Bad Requests (400)
When some site wants to make a redirect and returns a URL like:
http://www.example.com?test=1
Scrapy follows the redirect as is, which results in HTTP error 400 (Bad Request).
This is because such URL translates to the following line in the HTTP request:
GET ?test=1 HTTP/1.1

Obviously it's wrong because it doesn't have the path component so most HTTP servers return error 400.

The correct way to handle such an address would be to normalize it to the following form:
http://www.example.com/?test=1
And then the request line would look like this:
GET /?test=1 HTTP/1.1

In 'normal' browsers it works fine - they normalize such incorrect URLs.

influentcoder:
Fixes #1133

When performing redirects, scrapy creates a new request based on the URL from the response location header.

In some cases, the redirect URL that we receive in the response is not properly normalized.
When performing a HTTP GET request with this URL, we may receive a HTTP error.

However, when using an actual web browser, it works fine, because the browser normalizes the location URL before performing the next request.

For example, when querying https://www.watsons.com.sg, the response location header contains:

`https://queue.watsons.com.sg?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=55&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F`

We can see that this part of the URL is not normalized: `queue.watsons.com.sg?c=aswatson`, it's missing a slash.
If you send a HTTP GET on this URL, you'll get some bad response code back (HTTP 400).

Firefox actually does the normalization to: `queue.watsons.com.sg/?c=aswatson` (notice the additional slash).

I've added a unit test to show the behavior.

More details in my answer in Stackoverflow: https://stackoverflow.com/a/53323565/869764

### DIFF
--- a/scrapy/downloadermiddlewares/redirect.py
+++ b/scrapy/downloadermiddlewares/redirect.py
@@ -1,7 +1,7 @@
 import logging
 from six.moves.urllib.parse import urljoin
 
-from w3lib.url import safe_url_string
+from w3lib.url import canonicalize_url, safe_url_string
 
 from scrapy.http import HtmlResponse
 from scrapy.utils.response import get_meta_refresh
@@ -68,7 +68,7 @@ class RedirectMiddleware(BaseRedirectMiddleware):
         if 'Location' not in response.headers or response.status not in allowed_status:
             return response
 
-        location = safe_url_string(response.headers['location'])
+        location = canonicalize_url(safe_url_string(response.headers['location']))
 
         redirected_url = urljoin(request.url, location)
 
--- a/tests/test_downloadermiddleware_redirect.py
+++ b/tests/test_downloadermiddleware_redirect.py
@@ -133,11 +133,15 @@ class RedirectMiddlewareTest(unittest.TestCase):
         req2 = self.mw.process_response(req1, rsp1, self.spider)
         rsp2 = Response('http://scrapytest.org/redirected', headers={'Location': '/redirected2'}, status=302)
         req3 = self.mw.process_response(req2, rsp2, self.spider)
+        rsp3 = Response('http://scrapytest.org/redirected', headers={'Location': 'http://scrapytest.org?k=v'}, status=302)
+        req4 = self.mw.process_response(req3, rsp3, self.spider)
 
         self.assertEqual(req2.url, 'http://scrapytest.org/redirected')
         self.assertEqual(req2.meta['redirect_urls'], ['http://scrapytest.org/first'])
         self.assertEqual(req3.url, 'http://scrapytest.org/redirected2')
         self.assertEqual(req3.meta['redirect_urls'], ['http://scrapytest.org/first', 'http://scrapytest.org/redirected'])
+        self.assertEqual(req4.url, 'http://scrapytest.org/?k=v')
+        self.assertEqual(req4.meta['redirect_urls'], ['http://scrapytest.org/first', 'http://scrapytest.org/redirected', 'http://scrapytest.org/redirected2'])
 
     def test_spider_handling(self):
         smartspider = self.crawler._create_spider('smarty')


guillaume-humbert:
@kmike I'm not sure how this is exactly handled in the different browsers to be honest. I've had a look at how `curl` handles it.

```
$ curl -v -L https://www.watsons.com.sg/
...
...
* Issue another request to this URL: 'https://queue.watsons.com.sg?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=55&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F'
* Rebuilt URL to: https://queue.watsons.com.sg/?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=55&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F
...
...
```

By glimpsing into the `curl` code base, I've found this: https://github.com/curl/curl/blob/curl-7_58_0/lib/url.c#L2280

```c
  /* If the URL is malformatted (missing a '/' after hostname before path) we
   * insert a slash here. The only letters except '/' that can start a path is
   * '?' and '#' - as controlled by the two sscanf() patterns above.
   */
  if(path[0] != '/') {
    /* We need this function to deal with overlapping memory areas. We know
       that the memory area 'path' points to is 'urllen' bytes big and that
       is bigger than the path. Use +1 to move the zero byte too. */
    memmove(&path[1], path, strlen(path) + 1);
    path[0] = '/';
    rebuild_url = TRUE;
}
```

They simply add a `/` before the URL path if it's missing.

I could create a function that does that, something like this:

```python
>>> from six.moves.urllib.parse import urlsplit, urlunsplit
>>> url = 'https://queue.watsons.com.sg?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=55&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F'
>>> scheme, netloc, path, query, fragment = urlsplit(url)
>>> path = path if path.startswith('/') else ('/' + path)
>>> urlunsplit((scheme, netloc, path, query, fragment))
'https://queue.watsons.com.sg/?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=55&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F'
>>>
```

What would be the best place to put this function? I'm thinking of calling it `rebuild_url`.

Do you think it should be defined in the redirect middleware script or is it general enough to be part of `w3clib` in `url.py`?

kmike:
hey @guillaume-humbert, thanks for checking it! However, I think we should make sure we're doing a right thing; curl is not a gold standard - generally it is good, but I recall some incorrect handling of edge cases in past. "Do what a browser is doing" is a gold standard here, as web developers are testing their websites against browsers; "do what's in RFC" may also work, though if this conflicts with browser behavior, browser wins.

So, questions:

* does this normalization happen only for redirects (Location header), or for other requests as well? What's the best stage we should normalize URLs at?
* does it happen only when query is present, or always?
* etc.

The best resource I found to answer these questions is https://url.spec.whatwg.org/#url-representation. It says:

> A URL’s path is a list of zero or more ASCII strings holding data, usually identifying a location in hierarchical form. It is initially empty.
Note: A special URL always has a non-empty path. 

"special URL" is defined here https://url.spec.whatwg.org/#is-special as having a special scheme (https://url.spec.whatwg.org/#special-scheme), which is one of "ftp", "file", "gopher", "http", "https", "ws" or "wss".

So this suggests we should always add `/` to an URL if a path is empty (regardless of redirect/etc) and a scheme is one of those, but not otherwise.

I think the best place to put such code is w3lib.url. And it looks like this is already implemented here: https://github.com/scrapy/w3lib/blob/4e6865e0a8eae05c64ff766b4869332d6942e9ab/w3lib/url.py#L87, in `safe_download_url` function (though not 100% correctly, as this is done regardless of scheme). But for some reason this function is not used in Scrapy! I have no idea why.

pc2000sg:
I am trying this and it looks like the middleware not executed. It goes as following. Any suggestions? I tried to put a print statement in "process_response" under class "FixLocationHeaderMiddleWare" but no output also.  Thanks.

1.) **Scrapy log when running the script**
......
2018-11-27 22:30:49 [scrapy.middleware] INFO: Enabled spider middlewares:
 **'diffmarts.middlewares.FixLocationHeaderMiddleWare',**
2018-11-27 22:30:49 [scrapy.middleware] INFO: Enabled item pipelines:
['diffmarts.pipelines.DiffmartsPipeline']
2018-11-27 22:30:49 [scrapy.core.engine] INFO: Spider opened                                                            2018-11-27 22:30:49 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)                                                                                                                       2018-11-27 22:30:50 [scrapy.spidermiddlewares.httperror] INFO: **Ignoring response <400 https://queue.watsons.com.sg?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=62&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F>: HTTP status code is not handled or not allowed**
2018-11-27 22:30:50 [scrapy.core.engine] INFO: Closing spider (finished)
2018-11-27 22:30:52 [scrapy.core.engine] ERROR: Scraper close failure

2.) **My main scrapy.py file**

class WatsonsSpider(scrapy.Spider):
    name = 'watsons'
    start_urls = ['http://www.watsons.com.sg/']
    print("debug1")
    **custom_settings = {
        **'diffmarts.middlewares.FixLocationHeaderMiddleWare':300**
    }**
    def parse(self, response):
        #print(response.url)
        for category_selector in response.xpath("//div[contains(@class, 'navSubContainer-box clearfix')]//div[contains(@class, 'navCol')]//dt/a"):
            cat_name = ' '.join(category_selector.xpath(".//text()").extract()).strip(' \n\r\t')

3.) **My middlewares.py file:**
...
        spider.logger.info('Spider opened: %s' % spider.name)

class FixLocationHeaderMiddleWare:

        if 'location' in response.headers:
            response.headers['location'] = canonicalize_url(response.headers['location'])
        return response

guillaume-humbert:
@pc2000sg If you look closely at my [stackoverflow answer](https://stackoverflow.com/a/53323565/869764), you will see that your settings are wrong, instead of:

```python
custom_settings = {
'diffmarts.middlewares.FixLocationHeaderMiddleWare':300
}
```

It should be:
```python
custom_settings = {
    'DOWNLOADER_MIDDLEWARES': {
        'diffmarts.middlewares.FixLocationHeaderMiddleWare': 650
    }
}
```

Also, make sure that the order number is greater than 600, as it needs to be applied after the `RedirectMiddleware` which has a order number of 600.

pc2000sg:
OK got it. Let me give it a try. Thanks for your advice!

pc2000sg:
@guillaume-humbert
I finally figured out the way to code the middleware changes. However after calling safe_download_url and adding "/" to the url it seems that the 400 error still persisted. Attached the scrapy log as reference below. Wonder if there is other possibility that caused the 400 error? Thanks.

 _'diffmarts.middlewares.FixLocationHeaderMiddleWare',
2019-01-01 21:17:14 [scrapy.middleware] INFO: Enabled item pipelines:
['diffmarts.pipelines.DiffmartsPipeline']
2019-01-01 21:17:14 [scrapy.core.engine] INFO: Spider opened
2019-01-01 21:17:14 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2019-01-01 21:17:15 [scrapy.spidermiddlewares.httperror] INFO: Ignoring response <400 https://queue.watsons.com.sg/?c=aswatson&e=watsonprdsg&ver=v3-java-3.5.2&cver=62&cid=zh-CN&l=PoC+Layout+SG&t=https%3A%2F%2Fwww.watsons.com.sg%2F>: HTTP status code is not handled or not allowed
2019-01-01 21:17:15 [scrapy.core.engine] INFO: Closing spider (finished)
2019-01-01 21:17:17 [scrapy.core.engine] ERROR: Scraper close failure
Traceback (most recent call last):_

PR #3495, opened 2018-11-16
[MRG+1] Include additional files in sdists

toddrme2178:
In particular this includes files needed for running the tests, as well as the changelog.


lopuhin:
@toddrme2178 thanks for taking care of this, since we already include tests, makes sense to include all supporting files as well.

PR #3503, opened 2018-11-22
Versions printing in separate lines

maramsumanth:
Printing the scrapy component versions in separate lines in the form of tuples for viewing it easily to the user.

### DIFF
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-
+from __future__ import print_function
 import sys
 import logging
 import warnings


scrapy/utils/log.py
```diff
@@ -144,10 +144,7 @@ def _get_handler(settings):
 def log_scrapy_info(settings):
     logger.info("Scrapy %(version)s started (bot: %(bot)s)",
                 {'version': scrapy.__version__, 'bot': settings['BOT_NAME']})
-    logger.info("Versions: %(versions)s",
-                {'versions': ", ".join("%s %s" % (name, version)
-                    for name, version in scrapy_components_versions()
-                    if name != "Scrapy")})
+    print(*scrapy_components_versions(), sep = "\n")
```
42B: Wouldn't this be Python 3 syntax only? I don't see a `from __future__ import print_function` at the top of this file, is it somewhere else?

maramsumanth: Yes , I forgot to take care about Python 2. I will commit the changes. Thanks :)!

42B: This might be a good time to delete the spaces around the equals sign too.

```python
print(*scrapy_components_versions(), sep="\n")
```

maramsumanth: But why this checks were not successful? @42B

rennerocha:
@maramsumanth ,
Using a `print` statement instead of the `logging` module is not a good practice as it removes some convenient functions that are available in this module. See https://docs.python.org/3/howto/logging.html#when-to-use-logging for more details. I also don't see any gain in changing the way that the versions are presented in the logs. It will just make the logs longer.

Regarding your questions about why the checks failed, when it happens, check the `Details` links that will show you what failed. In this case, your change broke the `VersionTest` (located in tests/test_command_version.py). Every time you do a change in any project (even if it looks like a tiny and harmless change), run the tests to check if you didn't break anything.

Issue #3504, opened 2018-11-24
JSON requests support


kasun:
This is a minor enhancement when dealing with JSON requests.

Right now to do a json request we have to do something like this,

    headers = {
        'Content-Type': 'application/json'
    }
    yield Request(url, method='POST', body=json.loads(data), headers=headers)

I'm proposing a new Request subclass that can save a few keystrokes.

    yield JSONRequest(url, data=data)

Will be sending a PR for this.

kmike:
Fixed by #3505 .

PR #3505, opened 2018-11-24
[MRG+2] Request subclass for json requests #3504

kasun:
PR for #3504 

This is a more convenient subclass of Request when dealing with json requests. This would allow something like this.

    from scrapy.http import JSONRequest

    data = {
        'name': 'value',
    }
    yield JSONRequest(url, data=data)

Unit tests are included. Will work on documentation depending on the feedback.


lopuhin:
Thanks for PR @kasun ! I didn't check the code really carefully, but this sounds like a useful feature to me, since we already have FormRequest and XmlRpcRequest, and given that scraping often involves calling APIs directly which often accept JSON payload.
Any other opinions?

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,28 @@
...
+
+from scrapy.http.request import Request
+
+
+class JSONRequest(Request):
+    def __init__(self, *args, **kwargs):
+        if 'method' not in kwargs:
+            kwargs['method'] = 'POST'
+
+        data = kwargs.pop('data', {})
```
lopuhin: I had a look into ``XmlRpcRequest ``, which seems quite close in purpose:
https://github.com/scrapy/scrapy/blob/886513c3751b92e42dcc8cb180d4c15a5a11ccaf/scrapy/http/request/rpc.py#L20

It's approach makes more sense to me:
- it allows passing raw ``body``
- it allows passing ``params`` (I'd rather also have ``params`` here for consistency) if no ``body`` is given
- if nothing is passed, it uses usual scrapy default for ``body``, and not an empty dict which seems arbitrary to me

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,28 @@
...
+
+import json
+
+from scrapy.http.request import Request
+
+
+class JSONRequest(Request):
+    def __init__(self, *args, **kwargs):
+        if 'method' not in kwargs:
+            kwargs['method'] = 'POST'
```
lopuhin: Maybe this is a slightly controversial thing to default to POST method, I think it makes sense and ``XmlRpcRequest`` does this too, but would like someone else's opinion.

tests/test_http_request.py
```diff
@@ -1147,5 +1148,57 @@ def test_latin1(self):
...
+
+        r4 = self.request_class(url="http://www.example.com/", body='body1', data=data)
+        self.assertEqual(r3.body, to_bytes(json.dumps(data)))
+
+    def test_replace(self):
+        """Test Request.replace() method"""
+        r1 = self.request_class("http://www.example.com")
+        hdrs = Headers(r1.headers)
+        hdrs[b'key'] = b'value'
+        r2 = r1.replace(body="New body", headers=hdrs)
```
lopuhin: For ``XmlRpcRequest``, this would replace the body, and for me it would look surprising that the body value would be ignored. What is the advantage of such behaviour? Here is how ``XmlRpcRequest`` handles this in ``__init__``: https://github.com/scrapy/scrapy/blob/886513c3751b92e42dcc8cb180d4c15a5a11ccaf/scrapy/http/request/rpc.py#L20
With such implementation it seems a custom ``replace`` method is not needed, and behaviour is less surprising.

scrapy/http/request/json_request.py
lopuhin: Regarding the naming: maybe ``data`` is still better, because e.g. requests lib uses ``data`` for this purpose, while ``params`` means a different thing.

lopuhin:
> Will work on documentation depending on the feedback.

I think feedback is positive, so if you could also add the docs that would be awesome 👍

scrapy/http/request/json_request.py
asadurski: I inspected some (~20) existing spiders/websites and, quite logically, POST was always used when any data was sent in the request and otherwise (when using only query string, no data sent) it's GET. So maybe it should depend on `data`?

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,28 @@
...
+
+        data = kwargs.pop('data', {})
+        kwargs['body'] = json.dumps(data)
+        super(JSONRequest, self).__init__(*args, **kwargs)
+        self.headers.setdefault(b'Content-Type', b'application/json')
```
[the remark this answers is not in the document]
raphapassini: +1

tests/test_http_request.py
```diff
@@ -1147,5 +1148,57 @@ def test_latin1(self):
...
+        r2 = self.request_class(url="http://www.example.com/", body=b"")
+        self.assertEqual(r2.body, b'{}')
+
+        data = {
+            'name': 'value',
+        }
+        r3 = self.request_class(url="http://www.example.com/", data=data)
+        self.assertEqual(r3.body, to_bytes(json.dumps(data)))
+
+        r4 = self.request_class(url="http://www.example.com/", body='body1', data=data)
```
lucywang000: I think an exception (or at least a warning) should be printed when both `body` and `data` is used in the constructor.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def replace(self, *args, **kwargs):
+        """ Create a new Request with the same attributes except for those
+            given new values. """
+
+        kwargs.pop('body', None)
```
lucywang000: why do we need this?

scrapy/http/request/json_request.py
kasun: Thanks @lopuhin for the quick review. Now,
body is allowed to be passed
data is ignored if body is passed closely resembling the XmlRpcRequest

lopuhin:
Thanks @kasun going to have a look now.

Jessie test failure here https://travis-ci.org/scrapy/scrapy/builds/465618944 (not linking directly as it produces a huge log) looks completely unrelated to the PR, has anyone seen it before?

docs/topics/request-response.rst
```diff
@@ -508,6 +508,38 @@ method for this job. Here's an example spider which uses it::
...
+The JSONRequest class extends the base :class:`Request` class with functionality for
+dealing with JSON requests.
+
+.. class:: JSONRequest(url, [data, ...])
+
+   The :class:`JSONRequest` class adds a new argument to the constructor called data. The
+   remaining arguments are the same as for the :class:`Request` class and are
+   not documented here.
+
+   Using the :class:`JSONRequest` will set the `Content-Type` header to `application/json`
```
lopuhin: Single ticks \` will be rendered in italic, while double \`` will be rendered as ``code``, and I think in this line and below it's better to use double quotes.
```suggestion
   Using the :class:`JSONRequest` will set the ``Content-Type`` header to ``application/json``
```

docs/topics/request-response.rst
```diff
@@ -508,6 +508,38 @@ method for this job. Here's an example spider which uses it::
...
+   and `Accept` header to `application/json, text/javascript, */*; q=0.01`
```
lopuhin:
```suggestion
   and ``Accept`` header to ``application/json, text/javascript, */*; q=0.01`` 
```

docs/topics/request-response.rst
```diff
@@ -508,6 +508,38 @@ method for this job. Here's an example spider which uses it::
...
+   remaining arguments are the same as for the :class:`Request` class and are
+   not documented here.
+
+   Using the :class:`JSONRequest` will set the `Content-Type` header to `application/json`
+   and `Accept` header to `application/json, text/javascript, */*; q=0.01` 
+
+   :param data: is any JSON serializable object that needs to be JSON encoded and assigned to body.
+      if :attr:`Request.body` argument is provided this parameter will be ignored.
+      if :attr:`Request.body` argument is not provided and data argument is provided :attr:`Request.method` will be 
+      set to POST automatically.
```
lopuhin: For consistency with the rest of the doc:
```suggestion
      set to ``'POST'`` automatically.
```

docs/topics/request-response.rst
```diff
@@ -508,6 +508,38 @@ method for this job. Here's an example spider which uses it::
...
+JSONRequest usage example
+-------------------------
+
+Sending a JSON POST request with a JSON payload::
+
+   data = {
+       'name1': 'value1',
+       'name2': 'value2',
+   }
+   yield JSONRequest(url='http://www.example.com/post/action', data=data)
```
lopuhin: 👍

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class JSONRequest(Request):
+    def __init__(self, *args, **kwargs):
+        body_passed = 'body' in kwargs
+        data = kwargs.pop('data', None)
+        if body_passed and data:
+            warnings.warn('Both body and data passed. data will be ignored')
+
+        elif not body_passed and data:
```
lopuhin: I think that if ``data`` is an empty list (which would be false), we still want to send it. I suggest checking both ``body`` and ``data`` for ``None`` - e.g. ``JSONRequest(body=None, data=[])`` should still be valid, I don't think we should distinguish between passing ``None`` and not passing anything (usually it's not done for convenience).

docs/topics/request-response.rst
kasun: Changed.

docs/topics/request-response.rst
kasun: changed

scrapy/http/request/json_request.py
kasun: Changed

kasun:
@lopuhin Suggested changes are done. Please check.

The only remaining thing is the jessie test failure.

lopuhin:
Thanks @kasun , 

> The only remaining thing is the jessie test failure.

Looks like it's failing on all builds, so please ignore it here.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, *args, **kwargs):
+        body_passed = kwargs.get('body', None) is not None
+        data = kwargs.pop('data', None)
+        data_passed = data is not None
+
+        if body_passed and data_passed:
+            warnings.warn('Both body and data passed. data will be ignored')
+
+        elif not body_passed and data_passed:
+            kwargs['body'] = json.dumps(data)
```
lopuhin: I'm not 100% sure here, asking for feedback, based on discussion with @kmike 
Shall we use ``sort_keys=True`` for ``json.dumps``? The reason to use them is to improve duplicate filter behaviour, without ``sort_keys``, it can happen that two requests for which ``data`` are equal, will get encoded to different bodies, and will not be filtered out.

lopuhin:
I don't have any other comments besides ``sort_keys=True`` question. @kmike do you have some extra comments or questions?

scrapy/http/request/json_request.py
kasun: @lopuhin Agree. Changed this.

docs/topics/request-response.rst
```diff
@@ -508,6 +508,38 @@ method for this job. Here's an example spider which uses it::
 
             # continue scraping with authenticated session...
 
+JSONRequest
+-----------
+
+The JSONRequest class extends the base :class:`Request` class with functionality for
+dealing with JSON requests.
+
+.. class:: JSONRequest(url, [data, ...])
```
kmike: Hey! I wonder how can we make argument docs more clear here - currently users may assume `data` is a second argument, while a second argument is callback.

lopuhin: 👍 
> is callback.

@kmike you probably meant "a keyword argument" here.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, *args, **kwargs):
+        body_passed = kwargs.get('body', None) is not None
+        data = kwargs.pop('data', None)
+        data_passed = data is not None
+
+        if body_passed and data_passed:
+            warnings.warn('Both body and data passed. data will be ignored')
+
+        elif not body_passed and data_passed:
+            kwargs['body'] = json.dumps(data, sort_keys=True)
```
kmike: There are other kwargs which can be passed to dumps function, most importantly - ensure_ascii=False; without it some request bodies can be 2x larger than needed. What do you think about adding `dumps_kwargs` parameter, which allows to override keyword arguments passed to json.dumps (including ensure_ascii and sort_keys)?

I think it can also be a good idea to move serialization to a method; it would make it easier to e.g. make a subclass which uses ujson instead of stdlib json library.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class JSONRequest(Request):
+    def __init__(self, *args, **kwargs):
+        body_passed = kwargs.get('body', None) is not None
+        data = kwargs.pop('data', None)
+        data_passed = data is not None
+
+        if body_passed and data_passed:
+            warnings.warn('Both body and data passed. data will be ignored')
```
kmike: Why isn't it an exception?

tests/test_http_request.py
```diff
@@ -1147,5 +1149,65 @@ def test_latin1(self):
...
+class JSONRequestTest(RequestTest):
+    request_class = JSONRequest
+    default_method = 'GET'
+    default_headers = {b'Content-Type': [b'application/json'], b'Accept': [b'application/json, text/javascript, */*; q=0.01']}
+
+    def setUp(self):
+        warnings.simplefilter("always")
+        super(JSONRequestTest, self).setUp()
+
+    def test_data(self):
```
kmike: would you mind split this test method into several methods, for individual cases?

kmike: Also, could you please add a test for .replace method?

docs/topics/request-response.rst
kmike: I meant that one should be able to write `JsonRequest(url, self.parse_result, data=...)`

lopuhin: Ah right this makes sense, sorry for the noise.

kasun: Thanks @kmike for the review.

This is slightly changed now.

scrapy/http/request/json_request.py
kasun: Agree. Changed.

scrapy/http/request/json_request.py
kasun: Was trying to closely resemble the XmlRpcRequest class

tests/test_http_request.py
kasun: Test for replacing data file done. Testing for replacing body is already in the base RequestTest class which JSONRequestTest class inherits from.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,50 @@
...
+
+        elif not body_passed and data_passed:
+            kwargs['body'] = self.dump(data, **self._dumps_kwargs)
+
+        return super(JSONRequest, self).replace(*args, **kwargs)
+
+    def dump(self, data, **kwargs):
+        """Convert to JSON """
+        return json.dumps(data, sort_keys=True, **kwargs)
```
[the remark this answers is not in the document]
kasun: @kmike Agree. Btw do you think users should be allowed to disable sort_keys? Do you see a use case for that?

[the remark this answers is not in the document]
kasun: Now sort_key=True is set in self._dumps_kwargs. Let me know if you think we should allow to override it. I think if we allow it, although very unlikely users have the chance to theoretically mess up the duplicate filter if they turn off sort_keys.

scrapy/http/request/json_request.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if body_passed and data_passed:
+            warnings.warn('Both body and data passed. data will be ignored')
+
+        elif not body_passed and data_passed:
+            kwargs['body'] = self.dump(data, **self._dumps_kwargs)
+
+        return super(JSONRequest, self).replace(*args, **kwargs)
+
+    def dump(self, data, **kwargs):
```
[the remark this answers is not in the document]
kasun: This is made private while also removing **kwargs from the method.

scrapy/http/request/json_request.py
```diff
@@ -13,7 +14,10 @@
 
 class JSONRequest(Request):
     def __init__(self, *args, **kwargs):
-        dumps_kwargs = kwargs.pop('dumps_kwargs', {})
+        dumps_kwargs = copy.deepcopy(kwargs.pop('dumps_kwargs', {}))
+        dumps_kwargs['sort_keys'] = True
```
kmike: Hey, I think it is better to have .setdefault here, i.e. allow user to override it. I don't think it would harm, and there can be weird cases where disabled sort_keys is needed (e.g. user wants a specific field order, and relies on Python 3.7 dict key order to preserve it).

kasun: Hey @kmike it is changed

kmike: 👍

kmike:
Thanks @kasun, I think that's fine to merge it. @lopuhin do you want to re-check it, since there were new commits after your approval?

lopuhin:
Looks great, thank you @kasun and thanks everyone for reviews 👍

kasun:
Thanks everyone for the reviews.

PR #3512, opened 2018-11-29
[MRG+1] Add sitemap_filter function to SitemapSpider class

victor-torres:
it makes it possible to filter sitemap urls by any available attribute

for example, you can filter urls with lastmod greater than a given datetime

it can be helpful when the url loc itself does not aggregate that information

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -680,7 +680,7 @@ SitemapSpider
 
         Default is ``sitemap_alternate_links`` disabled.
 
-    .. method:: sitemap_filter(urls)
+    .. method:: sitemap_filter(entries)
 
         Specifies a function to filter sitemap entries and their attributes.
 
@@ -691,19 +691,19 @@ SitemapSpider
                 <lastmod>2005-01-01</lastmod>
             </url>
 
-        We can define a ``sitemap_filter`` function to filter ``urls`` by date::
+        We can define a ``sitemap_filter`` function to filter ``entries`` by date::
 
-            def sitemap_filter(urls):
+            def sitemap_filter(entries):
                 from datetime import datetime
-                for url in urls:
-                    date_time = datetime.strptime(url['lastmod'], '%Y-%m-%d')
+                for entry in entries:
+                    date_time = datetime.strptime(entry['lastmod'], '%Y-%m-%d')
                     if date_time.year >= 2005:
-                        yield url
+                        yield entry
 
-        This would retrieve only ``urls`` modified on 2005 and the following
+        This would retrieve only ``entries`` modified on 2005 and the following
         years.
 
-        If you omit this method, all urls found in sitemaps will be
+        If you omit this method, all entries found in sitemaps will be
         processed, observing other attributes and their settings.
 
 
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -694,7 +694,8 @@ SitemapSpider
 
         We can define a ``sitemap_filter`` function to filter ``entries`` by date::
 
-            from scrapy.spiders.sitemap import SitemapSpider
+            from datetime import datetime
+            from scrapy.spiders import SitemapSpider
 
             class FilteredSitemapSpider(SitemapSpider):
                 name = 'filtered_sitemap_spider'
@@ -702,7 +703,6 @@ SitemapSpider
                 sitemap_urls = ['http://example.com/sitemap.xml']
 
                 def sitemap_filter(self, entries):
-                    from datetime import datetime
                     for entry in entries:
                         date_time = datetime.strptime(entry['lastmod'], '%Y-%m-%d')
                         if date_time.year >= 2005:
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -680,7 +680,7 @@ SitemapSpider
 
         Default is ``sitemap_alternate_links`` disabled.
 
-    .. attribute:: sitemap_filter
+    .. method:: sitemap_filter(urls)
 
         Specifies a function to filter sitemap entries and their attributes.
 
@@ -703,7 +703,7 @@ SitemapSpider
         This would retrieve only ``urls`` modified on 2005 and the following
         years.
 
-        If you omit this attribute, all urls found in sitemaps will be
+        If you omit this method, all urls found in sitemaps will be
         processed, observing other attributes and their settings.
 
 
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -682,7 +682,8 @@ SitemapSpider
 
     .. method:: sitemap_filter(entries)
 
-        Specifies a function to filter sitemap entries and their attributes.
+        This is a filter funtion that could be overridden to select sitemap entries
+        based on their attributes.
 
         For example::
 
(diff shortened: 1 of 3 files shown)


raphapassini:
@victor-torres seems like a nice feature. But we'll need some tests

victor-torres:
Done @raphapassini

scrapy/spiders/sitemap.py
```diff
@@ -31,6 +31,14 @@ def start_requests(self):
         for url in self.sitemap_urls:
             yield Request(url, self._parse_sitemap)
 
+    def sitemap_filter(self, urls):
```
kmike: I think `urls` name is confusing, as one would expect it is a list of strings, not a list of dicts with information. Any ideas about a better name for this argument? We may also need to document it better, at least to explain which keys are available.

kmike:
Hey @victor-torres! I think this feature makes sense, and implementation is fine, but the current API won't be clear for users.

scrapy/spiders/sitemap.py
victor-torres: Sure. I had the same thought but weren't able to think about a better name. Indeed they are <url> tags being iterated... As in the .xml file.

victor-torres:
Should we discuss it with someone else, @kmike?

scrapy/spiders/sitemap.py
dangra: what about `links` or `snippets` ?

victor-torres: `links` will lead to the same confusion, someone would expect a list of strings... 

I like `entries` tho.

victor-torres:
I'll rename the `url` param to `entries` as suggested by @dangra. 

Still need to fix the coverage problem stated by dodecov exten.

scrapy/spiders/sitemap.py
kmike: +1 to entries, for links I'd expect a list of scrapy.link.Link instances.

dangra:
please rebase to get travis-ci fix  https://github.com/scrapy/scrapy/pull/3538

dangra:
@victor-torres do you plan to add a test case to cover the lines that are missing coverage?

victor-torres:
> @victor-torres do you plan to add a test case to cover the lines that are missing coverage?

I'm not sure about what additional tests I should write to increase the coverage on this case.

victor-torres:
All checks have passed :)

kmike:
Hey @victor-torres! The implementation and argument name looks fine, but I still think we should document what's inside a single `entry` - what keys/values are available, what they mean.

victor-torres:
@kmike, thanks for your contribution here.

I think this is already documented in the `docs/topics/spiders.rst` file when we show this example:

```

.. method:: sitemap_filter(entries)

    Specifies a function to filter sitemap entries and their attributes.

    For example::

        <url>
            <loc>http://example.com/</loc>
            <lastmod>2005-01-01</lastmod>
        </url>

    We can define a ``sitemap_filter`` function to filter ``entries`` by date::

        def sitemap_filter(entries):
            from datetime import datetime
            for entry in entries:
                date_time = datetime.strptime(entry['lastmod'], '%Y-%m-%d')
                if date_time.year >= 2005:
                    yield entry

    This would retrieve only ``entries`` modified on 2005 and the following
    years.

    If you omit this method, all entries found in sitemaps will be
    processed, observing other attributes and their settings.
```

Given the context, it's clear to me that entries come from the sitemap document and the keys/values will depend upon the document structure itself. Explaining the sitemap XML format here doesn't seem to be in the scope. 

Of course my opinion may be biased and I'd love to hear some suggestions from you guys.

kmike:
@victor-torres I agree that documenting sitemap XML doesn't make sense, though entries we return don't match attributes exactly - see https://github.com/scrapy/scrapy/blob/4306886ac6030c9312547bdbcff3d88eb89c11d0/scrapy/utils/sitemap.py#L22

For example, namespaces seem to be removed, there is an extra "alternate" key, and rows without "loc" are dropped - this is what I think is nice to have documented.

docs/topics/spiders.rst
```diff
@@ -703,6 +703,16 @@ SitemapSpider
...
 
+        Entries are dict objects extracted from the sitemap document.
+        Usually, the key is the tag name and the value is the text inside it.
+
+        It's important to notice that:
+
+        - as the loc attribute is required, entries without this tag are discarded
+        - alternate links are stored in a list with the key ``alternate``
+          (see ``sitemap_alternate_links``)
+        - namespaces are removed, so lxml tags named as ``{foo}bar`` become only ``bar``
```
dangra: Is the syntax for namespaces correct? the docs you propose used curly braces syntax `{foo}bar` while the namespaces as being tested in `test_spider.py` uses `<namespace:tagname>` syntax.

dangra: btw, I think we lack a testcase for namespaces in test for sitemap filter.

victor-torres: @dangra, when we're parsing the tags with lxml it puts the tag's namespace inside curly braces. 

`<namespace:tagname>` would be named as `{namespace}tagname`.

That's why we're doing this: https://github.com/scrapy/scrapy/blob/4306886ac6030c9312547bdbcff3d88eb89c11d0/scrapy/utils/sitemap.py#L27

victor-torres: I have just added a test case to cover sitemap filter usage with alternate links. If you're satisfied with the test case, could you resolve this conversation? Otherwise, I'm open to additional comments and directions.

victor-torres: I have also slightly changed docs: `{foo}bar` is now `{namespace}tagname` as you have just used. Think it will make things clearer.

victor-torres:
Force pushed twice because I rebased from my fork instead of this repo. Hope the rebase fixes the tests.

docs/topics/spiders.rst
```diff
@@ -680,6 +680,42 @@ SitemapSpider
...
+        For example::
+
+            <url>
+                <loc>http://example.com/</loc>
+                <lastmod>2005-01-01</lastmod>
+            </url>
+
+        We can define a ``sitemap_filter`` function to filter ``entries`` by date::
+
+            def sitemap_filter(entries):
```
kmike: Shouldn't it have self as a first argument? What do you think about showing a complete example, with subclassing?

victor-torres: Good observation! Fixed.

Also took the chance to improve the docs as suggested by you.

docs/topics/spiders.rst
```diff
@@ -680,6 +680,42 @@ SitemapSpider
 
         Default is ``sitemap_alternate_links`` disabled.
 
+    .. method:: sitemap_filter(entries)
+
+        Specifies a function to filter sitemap entries and their attributes.
```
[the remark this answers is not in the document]
victor-torres: Addressed in the last commit.

victor-torres:
Done @kmike

docs/topics/spiders.rst
```diff
@@ -693,12 +694,17 @@ SitemapSpider
 
         We can define a ``sitemap_filter`` function to filter ``entries`` by date::
 
-            def sitemap_filter(entries):
-                from datetime import datetime
-                for entry in entries:
-                    date_time = datetime.strptime(entry['lastmod'], '%Y-%m-%d')
-                    if date_time.year >= 2005:
-                        yield entry
+            class FilteredSitemapSpider(scrapy.SitemapSpider):
```
kmike: SitemapSpider is not exposed at top level, I think an import from scrapy.spiders is needed (see SitemapSpider examples section).

victor-torres: Updated

docs/topics/spiders.rst
```diff
@@ -694,7 +694,9 @@ SitemapSpider
 
         We can define a ``sitemap_filter`` function to filter ``entries`` by date::
 
-            class FilteredSitemapSpider(scrapy.SitemapSpider):
+            from scrapy.spiders.sitemap import SitemapSpider
```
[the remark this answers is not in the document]
victor-torres: No problem, @kmike. That's what reviews are for and you're showing up as a really helpful reviewer :) 

I have just pushed those changes.

kmike:
Thanks @victor-torres!

Issue #2473, opened 2017-01-02
Issue with running scrapy spider from script.


tituskex:
Hi, I'm trying to run scrapy from a script like this:

    import scrapy
    from scrapy.crawler import CrawlerProcess

    class MySpider(scrapy.Spider):
        name = "basic"
        allowed_domains = ["web"]
        start_urls = ['http://www.example.com']

        def parse(self, response):
            l = ItemLoader(item=PropertiesItem(), response = response)
            l.add_xpath('title', '//h1[1]/text()')

            return l.load_item()
    process = CrawlerProcess({
        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)'
    })

    process.crawl(MySpider)
    process.start()

However, when I run this script I get the following error:

    File "/Library/Python/2.7/site-packages/Twisted-16.7.0rc1-py2.7-macosx-10.11-
    intel.egg/twisted/internet/_sslverify.py", line 38, in <module>
    TLSVersion.TLSv1_1: SSL.OP_NO_TLSv1_1,
    AttributeError: 'module' object has no attribute 'OP_NO_TLSv1_1'

Does anyone know how to fix this? Thanks in advance.

IAlwaysBeCoding:
I would try to downgrade your twisted version from `Twisted==16.7.0rc1` to `Twisted==16.4.1`. I got some weird errors too on the downloader part when I ran my Scrapy spiders with the same version you are running.


```python

2017-01-02 14:25:00 [scrapy] ERROR: Error downloading <GET http://www.citysearch.com/profile/645344264/jackson_ms/wright_patrick_b_md_patrick_b_wright_md.html>
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/usr/local/lib/python2.7/dist-packages/twisted/python/failure.py", line 393, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/usr/local/lib/python2.7/dist-packages/scrapy/core/downloader/middleware.py", line 43, in process_request
    defer.returnValue((yield download_func(request=request,spider=spider)))
  File "/usr/local/lib/python2.7/dist-packages/scrapy/utils/defer.py", line 45, in mustbe_deferred
    result = f(*args, **kw)
[...]

PR #3517, opened 2018-12-03
[MRG+1] Add a troubleshooting section to the installation instructions

Gallaecio:
Its initial content covers the workaround for #2473.


kmike:
Thanks @Gallaecio!

PR #3393, opened 2018-08-17
[MRG+1] New signal for reqeuests reached downloader

whalebot-helmsman:
New signal added with documentation to be emitted for requests reached downloader. It is done to provide better integration between scheduler and downloader without adding much complexity, based on previous discussions: https://github.com/scrapy/scrapy/issues/2435 https://github.com/scrapy/scrapy/issues/2474

Signal name is discussable. Documentation can be improved.

PR #3520, opened 2018-12-05
[MRG+1] Downloader-aware Priority Queue for Scrapy

whalebot-helmsman:
This is long long-awaited by community(https://github.com/scrapy/scrapy/issues/1802, https://github.com/scrapy/scrapy/issues/2474 and https://github.com/scrapy/scrapy/issues/3140) priority queue implementation for multi host case. Contrary to past proposals it is based on ~Downloader signals~ direct interaction with downloader, not on round-robin principle. Round-robin is much slower according to benchmarks. ~It became possible after new signal was added in https://github.com/scrapy/scrapy/pull/3393~.
Intention was to provide solution for this pain in shortest term as possible. So two problems left out of scope:
- I didn't directly use https://github.com/scrapy/scrapy/pull/3160 but take a lot of inspiration from it.  Thanks @tianhuil. https://github.com/scrapy/scrapy/pull/3160 replaces priorities by hostnames as a slot for scheduling. As a result priorities of requests does not affect scheduling. I tried to merge priority and hostname to single structure. This structure is consumable by old downstream priority queue. As a result priorities affect scheduling. Both ways are not perfect. Perfect solution is to refactor current Scheduler structure.
- This new priority queue is not set as default priority queue, because it does not support `CONCURRENCY_PER_IP` option. To enable such support we should resolve hostnames on adding them to queue. This is deferred operation and code calling to `Scheduler.enqueue_request` should be ready for this situation. This is not the case now.

I run benchmarks using https://github.com/scrapy/scrapy-bench. In single host case(bookworm spider) new priority queue does not affect performance. In multi host case(broadworm spider) new priority queue improves it 10 times. Exact numbers available https://docs.google.com/spreadsheets/d/1f8iSzfmLuiG9O-mDgG4iDb4fGvAlnkC4sH_zr9Qlj-s/edit#gid=0, full logs are available https://gist.github.com/whalebot-helmsman/83ed45a1aa1b4221ab6be2f36226c5e3. Round robin based queue also does not affect single host case, but improves multi host one only 2-3x. That's why I remove round robin implementation.

Settings used for bookworm https://github.com/scrapy/scrapy-bench/blob/f03011d63f4149eeaad6dd7908ffa6fcbbb3304e/books/books/settings.py
Settings used for broadworm https://github.com/scrapy/scrapy-bench/blob/f03011d63f4149eeaad6dd7908ffa6fcbbb3304e/broad/broad/settings.py

List of changes:
- added tests for current priority queue
- create priority queue in scheduler using `create_instance`
- added `DOWNLOADER_SLOT` constant to Downloader
- added wrapper for `queuelib.PriorityQueue` to handle merged priorities
- added DownloaderAwarePriorityQueue implementation
- added tests for DownloaderAwarePriorityQueue 
- ~added `request_left_downloader` signal as `response_downloaded` sent only in case of successful download~


scrapy/pqueues.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+        as we replace some letters we can get collision for different slots
+        add we add unique part
+    """
+    unique_slot = hashlib.md5(x.encode('utf8')).hexdigest()
+
+    return '-'.join([pathable_slot, unique_slot])
+
+
+class PrioritySlot:
```
kmike: This looks like a tuple with a custom `__str__` method; do you think some code can be simplified if we make it just a tuple? For example, would it allow to remove PriorityAsTupleQueue?

scrapy/pqueues.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def __eq__(self, other):
+        return (self.priority, self.slot) == (other.priority, other.slot)
+
+    def __lt__(self, other):
+        return (self.priority, self.slot) < (other.priority, other.slot)
+
+    def __str__(self):
+        return '_'.join([text_type(self.priority),
+                         _pathable(text_type(self.slot))])
```
kmike: Could you please explain, what's the main use case for this method? Is it for debuggability, or for something else?

scrapy/pqueues.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class PriorityAsTupleQueue(PriorityQueue):
+    """
+        Python structures is not directly (de)serialized (to)from json.
+        We need this modified queue to transform custom structure (from)to
+        json serializable structures
+    """
```
kmike: We usually use a different indentation, i.e. text inside a docstring is not indented further than quotes.

scrapy/pqueues.py
```diff
@@ -0,0 +1,218 @@
...
+    return slot
+
+
+def _pathable(x):
+    pathable_slot = "".join([c if c.isalnum() or c in '-._' else '_' for c in x])
+
+    """
+        as we replace some letters we can get collision for different slots
+        add we add unique part
+    """
```
kmike: style nitpick: we use regular comments (`#`) in such cases

scrapy/pqueues.py
whalebot-helmsman: This method is used for creating names for queue files on disk

scrapy/pqueues.py
whalebot-helmsman: I will be glad if we can drop `PriorityAsTupleQueue`. Tuples encoded to json as lists, lists decoded from json as lists. Lists are unhashable. Only hashable types can be used inside common `PriorityQueue`

whalebot-helmsman:
It is not possible to check correct scheduling for `DownloaderAwarePriorityQueue` in integration test. A lot of things take place simultaneously - signals, site-response and scheduling moments. I just remove integration test and keep unit test.

tianhuil:
I really hope we can get this feature.  It's an important feature and I just haven't had time to work on my PR.

kmike:
I've started to look at the code; proposed (incomplete) cleanup so far: https://github.com/whalebot-helmsman/scrapy/pull/2

scrapy/pqueues.py
```diff
@@ -0,0 +1,217 @@
...
+
+        slot = min(slots)[1]
+        request = self._slot_pqueues.pop_slot(slot)
+        self.mark(request)
+        return request
+
+    def push(self, request, priority):
+        slot = _set_scheduler_slot(request)
+        priority_slot = PrioritySlot(priority=priority, slot=slot)
+        self._slot_pqueues.push_slot(slot, request, priority_slot)
```
lucywang000: can you explain why you need this  `PrioritySlot` structure, if all requests in `self._slot_pqueues.push_slot[slot]` would always have the same `priority_slot.slot`?

tests/test_scheduler.py
```diff
@@ -0,0 +1,320 @@
...
+        for url, slot in _SLOTS:
+            request = Request(url)
+            _scheduler_slot_write(request, slot)
+            self.scheduler.enqueue_request(request)
+
+        self.close_scheduler()
+        self.create_scheduler()
+
+        slots = list()
+        requests = list()
```
lucywang000: why `list()` not `[]`?

scrapy/pqueues.py
```diff
@@ -0,0 +1,217 @@
...
+                             for c in text])
+    # as we replace some letters we can get collision for different slots
+    # add we add unique part
+    unique_slot = hashlib.md5(text.encode('utf8')).hexdigest()
+    return '-'.join([pathable_slot, unique_slot])
+
+
+class PrioritySlot(namedtuple("PrioritySlot", ["priority", "slot"])):
+    """ ``(priority, slot)`` tuple which uses a path-safe slot name
+    when converting to str """
```
lucywang000: could you explain why it need to be path-safe?

scrapy/pqueues.py
```diff
@@ -0,0 +1,217 @@
...
+        crawler.signals.connect(self.on_response_download,
+                                signal=response_downloaded)
+        crawler.signals.connect(self.on_request_reached_downloader,
+                                signal=request_reached_downloader)
+
+    def mark(self, request):
+        meta = _get_request_meta(request)
+        if not isinstance(meta, dict):
+            raise ValueError('No meta attribute in %s' % (request, ))
+        meta[self._DOWNLOADER_AWARE_PQ_ID] = id(self)
```
lucywang000: please also remove this meta key properly, for instance in "on_response_download"

tests/test_scheduler.py
```diff
@@ -0,0 +1,320 @@
...
+class TestSchedulerOnDisk(BaseSchedulerOnDiskTester, unittest.TestCase):
+    priority_queue_cls = 'queuelib.PriorityQueue'
+
+
+_SLOTS = [("http://foo.com/a", 'a'),
+          ("http://foo.com/b", 'a'),
+          ("http://foo.com/c", 'b'),
+          ("http://foo.com/d", 'b'),
+          ("http://foo.com/e", 'c'),
+          ("http://foo.com/f", 'c')]
```
lucywang000: Please also add test for scheduling urls with different domains situation in order to verify new queue's capability of multi-domain aware scheduling

whalebot-helmsman: Added tests for `_set_scheduler_slots` function in https://github.com/scrapy/scrapy/pull/3520/commits/5d3a2e9f3a8cd067f04172b7935398f2da1b0db7

scrapy/pqueues.py
```diff
@@ -0,0 +1,217 @@
...
+            raise ValueError("DownloaderAwarePriorityQueue accepts "
+                             "``startprios`` as a dict; %r instance is passed."
+                             " Only a crawl started with the same priority "
+                             "queue class can be resumed." % startprios.__class__)
+        self._slot_pqueues = SlotPriorityQueues(pqfactory,
+                                                slot_startprios=startprios)
+
+        self._active_downloads = {slot: 0 for slot in self._slot_pqueues.pqueues}
+        crawler.signals.connect(self.on_response_download,
+                                signal=response_downloaded)
```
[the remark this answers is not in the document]
whalebot-helmsman: Yes @lucywang000 you are right. Has to add `request_left_downloader` signal, as `response_downloaded` sent only in case of successful download

whalebot-helmsman:
I removed commits related to `request_left_downloader` signal, but htey preserved in https://github.com/whalebot-helmsman/scrapy/tree/signal-backup branch

docs/topics/settings.rst
```diff
@@ -1142,9 +1142,14 @@ Type of in-memory queue used by scheduler. Other available type is:
...
 SCHEDULER_PRIORITY_QUEUE
 ------------------------
-Default: ``'queuelib.PriorityQueue'``
-
-Type of priority queue used by scheduler.
+Default: ``'scrapy.pqueues.ScrapyPriorityQueue'``
+
+Type of priority queue used by scheduler. Another available type is
+``scrapy.pqueues.DownloaderAwarePriorityQueue``.
+``scrapy.pqueues.DownloaderAwarePriorityQueue`` is works better than
```
kmike:
```suggestion
``scrapy.pqueues.DownloaderAwarePriorityQueue`` works better than
```

docs/topics/settings.rst
```diff
@@ -1142,9 +1142,14 @@ Type of in-memory queue used by scheduler. Other available type is:
...
+``scrapy.pqueues.ScrapyPriorityQueue`` when you crawl many different
+domains in parallel. But ``scrapy.pqueues.DownloaderAwarePriorityQueue``
```
kmike:
```suggestion
domains in parallel. But currently ``scrapy.pqueues.DownloaderAwarePriorityQueue``
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.crawler.stop()
+
+    @defer.inlineCallbacks
+    def test_integration_downloader_aware_priority_queue(self):
+        with MockServer() as mockserver:
+
+            url = mockserver.url("/status?n=200", is_secure=False)
+            slots = [url] * 6
```
kmike: What do you think about renaming it to start_urls? This list is used as spider's start_urls, and if I'm not mistaken, all of them end up in a single Downloader slot and in a single scheduler slot.

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+class StartUrlsSpider(Spider):
+
+    def __init__(self, start_urls):
+        self.start_urls = start_urls
+
+    def parse(self, response):
+        pass
+
+
+class TestIntegrationWithDownloaderAwareOnDisk(TestCase):
```
kmike: If I'm not mistaken, disk queues are not used in this test, because JOBDIR is not set.

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+        self.assertEqual(len(slots), len(_SLOTS))
+
+        for request in requests:
+            slot = downloader._get_slot_key(request, None)
+            self.mock_crawler.engine.downloader.decrement(slot)
+
+        unique_slots = len(set(s for _, s in _SLOTS))
+        for i in range(0, len(_SLOTS), unique_slots):
+            part = slots[i:i + unique_slots]
+            self.assertEqual(len(part), len(set(part)))
```
kmike: This code looks very similar to _is_slots_unique function

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+def _is_slots_unique(base_slots, result_slots):
```
kmike: Hm, I see what this function is checking; this check makes sense. But it was not easy to understand, and the check has caveats - e.g. it will fail for any algorithm (including good algorithms) for some input slots, e.g. if you have 1 request for slot A and 5 requests for slot B; it only works if there is roughly the same number of requests (+-1?) for each slot in the input data.

So I think it'd be good to 

a) rename it - e.g. `_assert_fair`, `_assert_round_robin` (though this one is not precise)?, and
b) add either a comment or an assert about kind of input data this check works on properly.

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+    unique_slots = len(set(s for _, s in base_slots))
```
kmike: I think that's confusing that 3 variables which are named in a very similar way contain completely different data:

* unique_slots is an integer number,
* base_slots is a list of (url, slot) tuples
* result_slots is a list of slots.

Could you please clean it up - e.g. rename unique_slots to n_unique_slots, and make this function accept a list of slots in `base_slots` argument, instead of (url, slot) tuples?

tests/test_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+class TestSchedulerWithDownloaderAwareOnDisk(BaseSchedulerOnDiskTester,
+                                             unittest.TestCase):
+    priority_queue_cls = 'scrapy.pqueues.DownloaderAwarePriorityQueue'
+
+    def test_logic(self):
+        downloader = self.mock_crawler.engine.downloader
+
+        for url, slot in _SLOTS:
+            request = Request(url)
+            downloader._set_slot_key(slot, request, None)
```
kmike: As _set_slot_key is not a part of Downloader API, I think it could be better to remove it from MockDownloader, and just create `request = Request(url, meta={'download_slot': slot})` here. `download_slot` is a public API, and it looks like MockDownloader supports it.

kmike:
Hey @whalebot-helmsman! 

I've left some comments about tests, but besides these comments the implementation looks good to me. 

Could you please also clean up the commit history?

I think after fixing these minor issues with tests and cleaning up the commit history the PR is ready to merge 🎉

lucywang000:
@whalebot-helmsman @kmike 

I made some improvements to the tests in https://github.com/whalebot-helmsman/scrapy/pull/6

docs/topics/broad-crawls.rst
```diff
@@ -39,6 +39,17 @@ you need to keep in mind when using Scrapy for doing broad crawls, along with
 concrete suggestions of Scrapy settings to tune in order to achieve an
 efficient broad crawl.
 
+Use proper :setting:`SCHEDULER_PRIORITY_QUEUE`
+==============================================
+
+Default scrapy's scheduler priority queue is ``'queuelib.PriorityQueue'``.
```
kmike:
```suggestion
Default scrapy's scheduler priority queue is ``'scrapy.pqueues.ScrapyPriorityQueue'``.
```

kmike:
I think that's good to merge, but commit history is not super-clean still:

* there are commits not related to this PR, such as https://github.com/scrapy/scrapy/pull/3520/commits/afdb69ea6daac8bd4f580d6c20bf9e93b741957b or https://github.com/scrapy/scrapy/pull/3520/commits/9c314800e4b195df41e5c0aba0d9ffe4bcffec8e - if I'm not mistaken, they'd be duplicated in Scrapy history if we merge it; 
* some of the commit names can be improved, e.g. https://github.com/scrapy/scrapy/pull/3520/commits/90934959d07db881aec5933fd7b77bcd2dccfa4f shouldn't be named `"actually apply __slots__ suggestion"`, it is a refactoring

I'm ok with merging it as-is, but if that's not too hard, it'd be good to clean it up.

scrapy/core/scheduler.py
```diff
@@ -1,53 +1,96 @@
...
+        self.crawler = crawler
 
     @classmethod
     def from_crawler(cls, crawler):
         settings = crawler.settings
         dupefilter_cls = load_object(settings['DUPEFILTER_CLASS'])
         dupefilter = create_instance(dupefilter_cls, settings, crawler)
         pqclass = load_object(settings['SCHEDULER_PRIORITY_QUEUE'])
+        if pqclass is PriorityQueue:
+            # backwards compatibility
```
Gallaecio: I think this inline comment is not necessary.

scrapy/core/scheduler.py
```diff
@@ -1,53 +1,96 @@
...
     def open(self, spider):
         self.spider = spider
-        self.mqs = self.pqclass(self._newmq)
+
+        # in-memory PriorityQueue instance
+        self.mqs = self._mq()
+
+        # on-disk PriorityQueue instance
         self.dqs = self._dq() if self.dqdir else None
+
```
Gallaecio: I think the empty lines and inline comments here are not necessary.

scrapy/core/scheduler.py
```diff
@@ -103,32 +145,54 @@ def _mqpush(self, request):
 
     def _dqpop(self):
         if self.dqs:
-            d = self.dqs.pop()
-            if d:
-                return request_from_dict(d, self.spider)
+            return self.dqs.pop()
+
+    def _mqpop(self):
+        return self.mqs.pop()
```
Gallaecio: What would you think about removing this method? It has a single, short line, and it is used only once.

You might even consider removing `_dqpop`, now that you’ve simplified it significantly.

whalebot-helmsman: I would prefer to preserve `_dqpop`, as it reduces complexity of already `if`-reach code

kmike:
Thanks @whalebot-helmsman and everyone who provided feedback!

I think we don't need a third opinion @Gallaecio (asked @dangra a while ago about this).

whalebot-helmsman:
Thanks for review @kmike, @lucywang000 and @Gallaecio for review

elacuesta:
Hi there @whalebot-helmsman, I think the `request_left_downloader` could still be useful, do you have any plans to submit a new PR with it?

PR #3526, opened 2018-12-10
remove "sudo: false" now that travis no longer supports it

lucywang000:
https://changelog.travis-ci.com/deprecation-container-based-linux-build-environment-82037


lucywang000:
The build failed on jessie but is ok on all other environments.

PR #3527, opened 2018-12-11
unify the quote style

hsiaoyi0504:
Just a minor modification of quote (use single quote for string).


kmike:
Thanks @hsiaoyi0504!

PR #3536, opened 2018-12-15
Update __init__.py

soubenz:
The current condition of checking ':' in the url doesn't provide a real url validation, I implemented a regex check that cover more cases.

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -4,6 +4,7 @@ requests in Scrapy.
 
 See documentation in docs/topics/request-response.rst
 """
+import re
 import six
 from w3lib.url import safe_url_string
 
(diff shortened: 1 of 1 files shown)


scrapy/http/request/__init__.py
```diff
@@ -57,8 +58,10 @@ def _set_url(self, url):
 
         s = safe_url_string(url, self.encoding)
         self._url = escape_ajax(s)
-
-        if ':' not in self._url:
+        # add better url scheme verification @stephenhay
+        pattern = "https://[^\s/$.?#].[^\s]*|^http://[^\s/$.?#].[^\s]*"
```
lopuhin: I think ``http`` and ``https`` are not the only possible schemes, in fact any scheme is possible with a custom downloader that supports it. I think this check is simple on purpose, it's main goal is to provide an error message for a common error, so it's fine if it's not exact - stuff will error out later in any case.
@soubenz are you trying to solve a specific use-case when an existing check was not enough?

soubenz: I think Scrapy should at least show a warning log message when the url doesn't match the most common schemes for webpages http & https. That is of course in case of using the default scrapy downloader.

PR #3543, opened 2018-12-22
Modular sitemap spider

stranac:
Currently,  the structure of `SitemapSpider` makes extraction of additional data from a sitemap unnecessarily complicated.
It would require copy-pasting the 22-line `_parse_sitemap` method, and then processing the created sitemap in a for loop.

This PR splits `_parse_sitemap` functionality into 3 additional methods, making this kind of processing simpler.
It also moves the link extraction inside the spider.

I also added tests for some of the previously untested `SitemapSpider` functionality.


wRAR:
While I like the idea, I wonder if the new methods should be public and, consequently, documented.

Also `iterloc` should be deprecated first, not just removed.

PR #3544, opened 2018-12-23
Fix item-pipeline example

joaquingx:
`if item['price']:` and `if item['price_excludes_vat']:` could return `KeyError` exception, this patch aims to solve it.

### DIFF
--- a/docs/topics/item-pipeline.rst
+++ b/docs/topics/item-pipeline.rst
@@ -87,8 +87,8 @@ contain a price::
         vat_factor = 1.15
 
         def process_item(self, item, spider):
-            if item['price']:
-                if item['price_excludes_vat']:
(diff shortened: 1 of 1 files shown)


docs/topics/item-pipeline.rst
```diff
@@ -87,8 +87,8 @@ contain a price::
         vat_factor = 1.15
 
         def process_item(self, item, spider):
-            if item['price']:
-                if item['price_excludes_vat']:
+            if 'price' in item and item['price']:
```
nasirhjafri: Why not `item.get('price')` ?

joaquingx: You're right, I'll change it.

dangra:
thanks

PR #3549, opened 2018-12-26
1.6 release notes

kmike:
This is work in progress. TODO:

* [x] port parts of parsel release notes;
* [x] more highlights
* [x] organize sections better
* [x] recent PRs
* [x] telnet changes

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -6,41 +6,83 @@ Release notes
 Scrapy 1.6.0 (unreleased)
 -------------------------
 
-Highlights for this release:
+Highlights:
 
-* better Windows compatibility;
+* better Windows support;
 * Python 3.7 compatibility;
 * big documentation improvements, including a switch
-  from ``.extract() / .extract_first()`` API to ``.get() / .getall()`` API;
-* Feed exports, FilePipeline and MediaPipeline improvements;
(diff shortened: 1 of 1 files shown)


docs/news.rst
```diff
@@ -6,11 +6,185 @@ Release notes
...
 -------------------------
 
-Cleanups
-~~~~~~~~
+Highlights:
+
+* better Windows support;
+* Python 3.7 compatibility;
+* big documentation improvements, including a switch
+  from ``.extract()`` + ``.extract_first()`` API to ``.get()`` + ``.getall()``
```
Gallaecio: It is a really minor thing, but both here and in references below I would suggest to switch the order of one of the pairs so that there is parallelism between the method references. That is, either:
```suggestion
  from ``.extract()`` + ``.extract_first()`` API to ``.getall()`` + ``.get()``
```
or
```suggestion
  from ``.extract_first()`` + ``.extract()`` API to ``.get()`` + ``.getall()``
```

kmike: Thanks @Gallaecio, applied! `.extract()` should be replaced with `.get()` for Selector, so it is not truly parallel, but anyways, your suggested order is better.

PR #3550, opened 2018-12-26
Backport security fix #3415

dangra:
Backport #3415 to 1.5 branch.


kmike:
Thanks @dangra! I guess we need to backport jessie+boto fix as well?

PR #3551, opened 2018-12-26
[MRG+1] Add ability to change max_active_size by setting

jpbalarini:
This PR fixes https://github.com/scrapy/scrapy/issues/1410

In our setup we had urls that were bigger than 5Mb (the default value) and we were reaching a deadlock scenario.
These changes allow for configuration of the `max_active_size` by setting the `SCRAPER_SLOT_MAX_ACTIVE_SIZE` configuration.

Thanks!

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1266,7 +1266,7 @@ does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
 
 SCRAPER_SLOT_MAX_ACTIVE_SIZE
 ----------------------------
-Default: ``5000000``
+Default: ``5_000_000``
 
 Soft limit (in bytes) for response data being processed.
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1262,6 +1262,17 @@ Type of priority queue used by the scheduler. Another available type is
 domains in parallel. But currently ``scrapy.pqueues.DownloaderAwarePriorityQueue``
 does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
 
+.. setting:: SCRAPER_SLOT_MAX_ACTIVE_SIZE
+
+SCRAPER_SLOT_MAX_ACTIVE_SIZE
+----------------------------
+Default: ``5000000``
+
+Soft limit (in bytes) for response data being processed.
+
+While the sum of the sizes of all responses being processed is above this value,
+Scrapy does not process new requests.
+
 .. setting:: SPIDER_CONTRACTS
 
 SPIDER_CONTRACTS
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -78,7 +78,7 @@ class Scraper(object):
     @defer.inlineCallbacks
     def open_spider(self, spider):
         """Open the given spider for scraping and allocate resources for it"""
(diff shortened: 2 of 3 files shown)


jpbalarini:
@kmike what do you think of these changes?
Thanks!

docs/topics/settings.rst
```diff
@@ -1146,6 +1146,15 @@ Default: ``'queuelib.PriorityQueue'``
...
 Type of priority queue used by scheduler.
 
+.. setting:: SCRAPER_SLOT_MAX_ACTIVE_SIZE
+
+SCRAPER_SLOT_MAX_ACTIVE_SIZE
+----------------------------
+Default: ``5000000``
+
+Maximum size (in bytes) which the scraper can handle at once. Change this setting
+if your requests will have a body larger than this amount.
```
dangra: The explanation about the setting is not accurate, even if the value is 1 byte the scraper will process bigger responses. 

It is a soft limit, in the sense that it process multiple responses until the sum of all responses being concurrently processed by scraper component (spider code mainly) goes above the limit. When it goes over, Scrapy core pauses the downloader (active downloads continue as normal, but new ones are not started). This mechanism prevents responses bodies from blowing up process memory by keeping only a sensible number of responses in memory.

dangra:
I think the change is simple and makes sense to make that limit configurable by settings, but I wonder what real case motivates it. 

IMO increasing the limit will only mitigate and delay the real reason for the deadlock mentioned in #1410.

jpbalarini:
@dangra what do you think of the response I made to your comment above? If you are ok with that I can update the Readme with that.
Thanks!

docs/topics/settings.rst
Gallaecio: What about this?

> Soft limit (in bytes) for response data being processed.
>
> While the sum of the sizes of all responses being processed is above this value, Scrapy does not process new requests.

jpbalarini:
@Gallaecio changed documentation to what you proposed. Let me know what do you think.
Thanks!

docs/topics/settings.rst
```diff
@@ -1174,6 +1174,17 @@ Type of priority queue used by the scheduler. Another available type is
 domains in parallel. But currently ``scrapy.pqueues.DownloaderAwarePriorityQueue``
 does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
 
+.. setting:: SCRAPER_SLOT_MAX_ACTIVE_SIZE
+
+SCRAPER_SLOT_MAX_ACTIVE_SIZE
+----------------------------
+Default: ``5000000``
```
Gallaecio: :lipstick: I recently found about [PEP 515](https://www.python.org/dev/peps/pep-0515/), maybe we should use it here? Even if it only works in Python 3.6+, I think it makes sense to use it for the documentation.

Gallaecio:
I left a minor, aesthetic comment, but the changes look good to me already as they are.

jpbalarini:
> I left a minor, aesthetic comment, but the changes look good to me already as they are.

Corrected what you said for the documentation 👍

jpbalarini:
> I left a minor, aesthetic comment, but the changes look good to me already as they are.

@Gallaecio Corrected what you said for the documentation

kmike:
Thanks @jpbalarini!

PR #3559, opened 2018-12-31
Scheduler: minimal interface, API docs

### LINKED ISSUE #3537: Document Scheduler API
The Scheduler component can be overridden with the [`SCHEDULER`](https://doc.scrapy.org/en/latest/topics/settings.html#scheduler) setting, but its interface is not documented.

elacuesta:
Fixes #3537


kmike:
Hey @elacuesta! Thanks, I think we should really document this API, though I'd

1) wait for https://github.com/scrapy/scrapy/pull/3520 to settle, and
2) move most of this documentation to Scheduler docstrings, using autodocs to pull from them.

elacuesta:
That's good @kmike, thanks! I can wait on this, but I wonder how far away is that from being merged? The same use case that motivated me to understand the Scheduler would also benefit greatly by the addition of the `request_left_downloader` signal. Perhaps the signal could be added independently if the PR is not yet ready as a whole.

kmike:
Finishing that PR is kind-of priority :)

> The same use case that motivated me to understand the Scheduler would also benefit greatly by the addition of the request_left_downloader signal. 

what's your use case?

Gallaecio:
Ping! #3520 has been merged :slightly_smiling_face:

docs/topics/scheduler.rst
```diff
@@ -0,0 +1,59 @@
...
+
+=========
+Scheduler
+=========
+
+The Scheduler receives requests from the :ref:`Engine <component-engine>`
+and enqueues them into disk and/or disk queues
+(:meth:`scrapy.core.scheduler.Scheduler.enqueue_request`).
+It also gets those requests and feeds them back when the engine requests
+them (:meth:`scrapy.core.scheduler.Scheduler.next_request`).
```
Gallaecio, line 11: I think it may be better to make some changes to this introduction:

- I think we should make it clear what the ultimate sources of requests are, possibly in an additional paragraph after this introduction: `start_requests` or `start_urls`, callbacks, methods A and B of spider middlwares, methods C and D of downloader middlewares.
- I think how the scheduler stores requests should be described in a later paragraph. And we need to be careful, because many of it seems specific to the default scheduler, a custom scheduler could handle request storage in a completely different way. Maybe here we should simply say that it is responsible for storing those request, without going into the specifics of how the default scheduler stores them. And we could then add a section about the default scheduler, and cover how it handles storage there.
- I think we should emphasize that the scheduler determines the order in which those requests are then fed into the engine.

docs/topics/scheduler.rst
```diff
@@ -0,0 +1,59 @@
...
+
+
+Scheduler interface
+===================
+
+This is the basic Scheduler API, i.e. the methods that are defined in the default Scheduler.
+
+.. module:: scrapy.core.scheduler
+
+.. autoclass:: Scheduler
```
Gallaecio, line 27: I think we need to make a clear distinction between the API that a scheduler class must provide, and the API of the default scheduler class.

elacuesta: Done in f27946a8d96f7b6958f203a3c45b710b316a10cd

elacuesta:
> I think we need to make a clear distinction between the API that a scheduler class must provide, and the API of the default scheduler class.

I think the latest changes make it clear that all the queue management stuff that the default scheduler does is not technically _essential_ to perform the scheduler functions.

<details>
  <summary>Now this extremely useful example is possible!</summary>

```python
from scrapy import Spider


class FriendlyScheduler:
    def __init__(self):
        self.requests = dict()

    def has_pending_requests(self):
        return bool(self.requests)

    def open(self, spider):
        print(f"Hello {spider.__class__.__name__}, thanks for using this scheduler")
        return None

    def close(self, reason):
        print("Farewell my friend")
        return None

    def enqueue_request(self, request):
        if request.url in self.requests:
            return False
        print("By all means, I will store this request for you")
        self.requests[request.url] = request
        return True

    def next_request(self):
        if self.has_pending_requests():
            _, request = self.requests.popitem()
            print(f"Enjoy your next request: {request.url}")
            return request
        return None


class QuotesSpider(Spider):
    name = "quotes"
    start_urls = [
        "http://quotes.toscrape.com/tag/friends/",
        "http://quotes.toscrape.com/tag/life/",
        "http://quotes.toscrape.com/tag/humor/",
    ]
    custom_settings = {
        "SCHEDULER": FriendlyScheduler,
        "LOG_LEVEL": "INFO",
    }

    def parse(self, response):
            yield {
                "author": quote.xpath("span/small/text()").get(),
                "text": quote.css("span.text::text").get(),
            }
```

```
$ scrapy runspider test-spiders/friendly_scheduler.py
2021-03-24 19:44:50 [scrapy.utils.log] INFO: Scrapy 2.4.1 started (bot: scrapybot)
2021-03-24 19:44:50 [scrapy.utils.log] INFO: Versions: lxml 4.6.2.0, libxml2 2.9.10, cssselect 1.1.0, parsel 1.6.0, w3lib 1.22.0, Twisted 21.2.0, Python 3.8.6 (v3.8.6:db455296be, Sep 23 2020, 13:31:39) - [Clang 6.0 (clang-600.0.57)], pyOpenSSL 20.0.1 (OpenSSL 1.1.1j  16 Feb 2021), cryptography 3.4.6, Platform macOS-10.15.7-x86_64-i386-64bit
2021-03-24 19:44:50 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.selectreactor.SelectReactor
2021-03-24 19:44:50 [scrapy.crawler] INFO: Overridden settings:
{'EDITOR': 'nano',
 'LOG_LEVEL': 'INFO',
 'SCHEDULER': <class 'friendly_scheduler.FriendlyScheduler'>,
 'SPIDER_LOADER_WARN_ONLY': True}
2021-03-24 19:44:50 [scrapy.extensions.telnet] INFO: Telnet Password: f376c0e3bc669451
2021-03-24 19:44:50 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.corestats.CoreStats',
 'scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.memusage.MemoryUsage',
 'scrapy.extensions.logstats.LogStats']
2021-03-24 19:44:50 [scrapy.middleware] INFO: Enabled downloader middlewares:
['scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
 'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
 'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
 'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
 'scrapy.downloadermiddlewares.retry.RetryMiddleware',
 'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
 'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
 'scrapy.downloadermiddlewares.cookies.CookiesMiddleware',
 'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware',
 'scrapy.downloadermiddlewares.stats.DownloaderStats']
2021-03-24 19:44:50 [scrapy.middleware] INFO: Enabled spider middlewares:
2021-03-24 19:44:50 [scrapy.middleware] INFO: Enabled item pipelines:
[]
2021-03-24 19:44:50 [scrapy.core.engine] INFO: Spider opened
Hello QuotesSpider, thanks for using this scheduler
2021-03-24 19:44:50 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2021-03-24 19:44:50 [scrapy.extensions.telnet] INFO: Telnet console listening on 127.0.0.1:6023
By all means, I will store this request for you
Enjoy your next request: http://quotes.toscrape.com/tag/friends/
By all means, I will store this request for you
Enjoy your next request: http://quotes.toscrape.com/tag/life/
By all means, I will store this request for you
Enjoy your next request: http://quotes.toscrape.com/tag/humor/
2021-03-24 19:44:51 [scrapy.core.engine] INFO: Closing spider (finished)
Farewell my friend
2021-03-24 19:44:51 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'downloader/request_bytes': 688,
 'downloader/request_count': 3,
 'downloader/request_method_count/GET': 3,
 'downloader/response_bytes': 6900,
 'downloader/response_count': 3,
 'downloader/response_status_count/200': 3,
 'elapsed_time_seconds': 0.509905,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2021, 3, 24, 22, 44, 51, 325836),
 'item_scraped_count': 24,
 'log_count/INFO': 10,
 'memusage/max': 51703808,
 'memusage/startup': 51703808,
 'response_received_count': 3,
 'start_time': datetime.datetime(2021, 3, 24, 22, 44, 50, 815931)}
2021-03-24 19:44:51 [scrapy.core.engine] INFO: Spider closed (finished)
```
</details>

scrapy/core/scheduler.py
```diff
@@ -16,17 +17,75 @@
 logger = logging.getLogger(__name__)
 
 
-class Scheduler:
+class BaseScheduler:
```
Gallaecio, line 20: I think we should either raise `NotImplementedError` or mark as @abstractmethod those methods that do not simply work as implemented here (all except `open` and `close`).

scrapy/core/scheduler.py
```diff
@@ -48,33 +153,56 @@ def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
...
-    def close(self, reason):
-        if self.dqs:
+    def close(self, reason: str) -> Optional[Deferred]:
+        """
+        (1) dump pending requests to disk if there is a disk queue
+        (2) return the result of the dupefilter's ``close`` method
+        """
+        if self.dqs is not None:
             state = self.dqs.close()
+            assert isinstance(self.dqdir, str)
```
elacuesta, line 192: This is only to avoid a typing error in the next line (`_write_dqs_state` expects a `str` but it gets `Optional[str]`). At this point, if `self.dqs is not None` it's only because `self.dqdir` is a `str`.

docs/topics/scheduler.rst
```diff
@@ -0,0 +1,57 @@
...
+
+   .. automethod:: open
+
+   .. automethod:: close
+
+   .. automethod:: has_pending_requests
+
+   .. automethod:: enqueue_request
+
+   .. automethod:: next_request
```
Gallaecio, line 37:
```suggestion
.. autoclass:: BaseScheduler
   :members:
```

scrapy/core/engine.py
```diff
@@ -73,12 +74,22 @@ def __init__(self, crawler, spider_closed_callback: Callable) -> None:
...
         self._spider_closed_callback = spider_closed_callback
 
+    def _get_scheduler_class(self, settings: BaseSettings) -> type:
+        from scrapy.core.scheduler import BaseScheduler
+        scheduler_cls = load_object(settings["SCHEDULER"])
+        if not issubclass(scheduler_cls, BaseScheduler):
+            raise TypeError(
+                f"The provided scheduler class ({settings['SCHEDULER']})"
+                " does not fully implement the scheduler interface"
+            )
```
Gallaecio, line 90: This will break any existing project with a custom scheduler class :sweat_smile:

scrapy/core/scheduler.py
```diff
@@ -1,42 +1,169 @@
...
+        ``True`` if the scheduler has enqueued requests, ``False`` otherwise
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def enqueue_request(self, request: Request) -> bool:
+        """
+        Process a request received by the engine.
+
+        Return ``True`` if the request is to be considered scheduled, ``False`` otherwise.
```
Gallaecio, line 90: I think it would be best to extend this to make it more clear why you would ever return `False` (e.g. your scheduler stores requests in an external service that is temporarily unreachable?), and what’s the consequence of doing so (is the request discarded, force-feeded into the engine, re-enqueued later?).

scrapy/core/scheduler.py
```diff
@@ -1,42 +1,169 @@
...
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def next_request(self) -> Optional[Request]:
+        """
+        Return the next :class:`~scrapy.http.Request` to be processed, or ``None``
+        to indicate that there are no requests to be considered ready at the moment.
```
Gallaecio, line 98: It would be great to be more explicit about what happens when you return `None`, and how the engine plays with this method and with `has_pending_requests`. Maybe provide an example of a scenario where it makes sense to return `None`.

Gallaecio (commented):
:heart:

scrapy/core/engine.py
elacuesta: It shouldn't, the subclass check is done via a [metaclass](https://github.com/scrapy/scrapy/pull/3559/files#diff-e957998ce47b270d8ae072df61f91368d94a20bb7bff60c337a0900eb192bf6dR27), it makes sure the necessary methods exist.

docs/topics/scheduler.rst
elacuesta: Nice, I didn't know about this

scrapy/core/engine.py
Gallaecio: Cool!

docs/topics/scheduler.rst
```diff
@@ -0,0 +1,46 @@
...
+
+   .. automethod:: __len__
```
Gallaecio, line 46:
```suggestion
.. autoclass:: Scheduler
   :members:
```

I haven’t checked the result, though. I wonder if `__len__` gets covered automatically as part of the set, or if we need to keep its `automethod` line to force it to be covered.

Gallaecio (approved):

docs/topics/scheduler.rst
elacuesta: Indeed, it's not included and adding `.. automethod:: __len__` after `:members:` does not help either, I'll revert to the previous verbose version.

docs/topics/scheduler.rst
```diff
@@ -0,0 +1,35 @@
...
+   :members:
+
+
+Default Scrapy scheduler
+========================
+
+.. autoclass:: Scheduler
+   :members:
+
+   .. automethod:: __len__
```
Gallaecio, line 35: Is not much more succinct, but it seems like [this](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-special-members) is “the way”:

```suggestion
   :members:
   :special-members: __len__
```

PR #3563, opened 2019-01-03
[MRG+1] Callback kwargs

### LINKED ISSUE #1138: Alternative way to pass arguments to callback
Currently to pass data between callbacks users need to use request.meta. Example from docs (simplified):

``` py
def parse_page1(self, response):
    item = MyItem(main_url=response.url)
    request = scrapy.Request("http://www.example.com/some_page.html",
                             callback=self.parse_page2)
    request.meta['item'] = item
    return request

def parse_page2(self, response):
    item = response.meta['item']
    item['other_url'] = response.url
    return item
```

While it works fine, there are some issues:
1. it seems understanding request.meta is a common struggle for beginners;
2. we're mixing parameters for Scrapy components with user data.

What about providing an alternative way?

``` py
def parse_page1(self, response):
    item = MyItem(main_url=response.url)
    return scrapy.Request("http://www.example.com/some_page.html",
                             callback=self.parse_page2,
                             kwargs={'item': item})

def parse_page2(self, response, item):
    item['other_url'] = response.url
    return item
```

Advantages:
1. If you're writing some extraction code without Scrapy (e.g. requests+lxml), then likely parsing functions have arguments. So this change makes code more natural/straightforward. 
2. Optional arguments or arguments with default values are easier to handle - just provide a default value using Python syntax.
3. User state is separated from Scrapy internals better.
4. Less code.
5. One can see which data callback needs just by looking at callback definition.
6. This way it is easier to add extra data to meta without a risk of breaking Scrapy extensions. There should be fewer bugs with missing `meta.copy()`.
7. In case of missing argument callback will fail earlier.

The implementation could add `__kwargs` field to request.meta and pass `**meta.get('__kwargs', {})` to the callback. Alternatively, we could put keyword arguments in another dict similar to meta. It will allow to separate them better. Also, rules for passing kwargs may be different from rules for passing meta (e.g. maybe meta should be preserved/copied in some cases, but not kwargs, I'm not sure).

elacuesta:
Fixes #1138

~This is just a first approach. It's currently lacking docs and tests, I'll add those if the implementation is good.~ **Update**: added tests and docs

I see (at least) two points for discussion:
1. Should we also pass the same arguments to the errbacks? Or maybe add a different parameter? `errback_kwargs` or something like that. On the other hand, the request object is available in the failure received by the errback, `failure.request.cb_kwargs` gives access to the arguments, so I think it shouldn't be necessary.
2. ~I'm not a fan of the `kwargs` name, I think it could be easily confused with Python's own "kwargs" naming convention, i.e., people could understand that any remaining keyword argument passed to the Request constructor will be passed to the callbacks. Is "callback_kwargs" too verbose? Maybe it's not compatible with the previous point.~ **Update** Renamed to `cb_kwargs`

Sample spider:
```python
import scrapy

class TestCallbackKwargsSpider(scrapy.Spider):
    name = 'callback_kwargs'

    def start_requests(self):
        data = {'a': 123, 'b': 456}
        yield scrapy.Request('https://example.org', cb_kwargs=data)

    def parse(self, response, a, b):
        yield {'url': response.url, 'a': a, 'b': b}
        yield response.follow(
            response.css('a::attr(href)').get(),
            self.parse_other,
            cb_kwargs={'source': response.url})

    def parse_other(self, response, source):
        yield {'url': response.url, 'source': source}
        yield response.follow(
            response.css('a::attr(href)').get(),
            self.parse_regular)

    def parse_regular(self, response):
        yield {'url': response.url}
```

Output:
```
(...)
2019-01-03 17:40:38 [scrapy.core.engine] INFO: Spider opened
2019-01-03 17:40:38 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2019-01-03 17:40:38 [scrapy.extensions.telnet] INFO: Telnet console listening on 127.0.0.1:6023
2019-01-03 17:40:38 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://example.org> (referer: None)
2019-01-03 17:40:39 [scrapy.core.scraper] DEBUG: Scraped from <200 https://example.org>
{'url': 'https://example.org', 'a': 123, 'b': 456}
2019-01-03 17:40:39 [scrapy.downloadermiddlewares.redirect] DEBUG: Redirecting (302) to <GET https://www.iana.org/domains/reserved> from <GET http://www.iana.org/domains/example>
2019-01-03 17:40:40 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://www.iana.org/domains/reserved> (referer: None)
2019-01-03 17:40:40 [scrapy.core.scraper] DEBUG: Scraped from <200 https://www.iana.org/domains/reserved>
{'url': 'https://www.iana.org/domains/reserved', 'source': 'https://example.org'}
2019-01-03 17:40:40 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://www.iana.org/> (referer: https://www.iana.org/domains/reserved)
2019-01-03 17:40:40 [scrapy.core.scraper] DEBUG: Scraped from <200 https://www.iana.org/>
{'url': 'https://www.iana.org/'}
2019-01-03 17:40:40 [scrapy.core.engine] INFO: Closing spider (finished)
(...)
```

**Tasks**
- [x] `Request` changes
- [x] `Response.follow` changes
- [x] `Scraper` changes
- [x] Request serialization
- [x] Tests
- [x] Docs

/cc @kmike @dangra


tests/test_crawl.py
```diff
@@ -23,6 +23,12 @@ def setUp(self):
     def tearDown(self):
         self.mockserver.__exit__(None, None, None)
 
+    @defer.inlineCallbacks
+    def test_callback_kwargs(self):
+        crawler = self.runner.create_crawler(KeywordArgumentsSpider)
+        yield crawler.crawl(mockserver=self.mockserver)
+        self.assertEqual(crawler.spider.checks, set([True]))
```
elacuesta: Not sure if this is the best way to test this, please advice if it's not.

elacuesta:
@dangra @kmike I'm sorry to bother you guys, but do you have any comments on this?

scrapy/http/request/__init__.py
```diff
@@ -40,8 +40,15 @@ def __init__(self, url, callback=None, method='GET', headers=None, body=None,
         self.dont_filter = dont_filter
 
         self._meta = dict(meta) if meta else None
+        self._kwargs = dict(kwargs) if kwargs else None
         self.flags = [] if flags is None else list(flags)
 
+    @property
+    def kwargs(self):
+        if self._kwargs is None:
```
ejulio: This check could go in `__init__`, right?
That way we can avoid this `if`

tests/spiders.py
```diff
@@ -28,6 +28,40 @@ def closed(self, reason):
...
+
+    def parse_general(self, response, **kwargs):
+        if response.url.endswith('/general_with'):
+            self.checks.add(kwargs['key'] == 'value')
+            self.checks.add(kwargs['number'] == 123)
+        elif response.url.endswith('/general_without'):
+            self.checks.add(kwargs == {})
+
+    def parse_no_kwargs(self, response):
+        pass
```
ejulio: Not sure if required, we are not testing anything :(
Maybe check the url?

tests/test_crawl.py
ejulio: They look fine for me, just not sure if testing a set with only one true value.
What do you think of a list? That way we can ensure that there `n` true elements to make sure all assertion were checked.

tests/spiders.py
elacuesta: It's required to avoid getting `AttributeError: 'KeywordArgumentsSpider' object has no attribute 'parse_no_kwargs'`, but it's true that it also could be doing something useful. Thanks!

scrapy/http/request/__init__.py
elacuesta: I have no strong argument in favor of this pattern, just that it's the same way the `meta` attribute is handled.

tests/test_crawl.py
elacuesta: That makes sense, it would be useful to verify the amount of checks. Thanks for the suggestion :-)

elacuesta:
Yeah I like `callback_kwargs` more than just `kwargs`, I'll better change it now to save time before we get close to 1.7 :-)
There were some comments about `partial` in the original issue, but it didn't seem to work. I think using the Deferred's API for this is the most natural choice, passing keyword arguments to a callback is precisely what we need.

elacuesta:
I decided to go with `cb_kwargs` instead of `callback_kwargs`. It's less verbose, and it's also already in use within Scrapy (https://docs.scrapy.org/en/latest/topics/spiders.html?highlight=cb_kwargs#scrapy.spiders.Rule)

docs/topics/request-response.rst
```diff
@@ -200,11 +214,9 @@ Example::
...
 
 In some cases you may be interested in passing arguments to those callback
-functions so you can receive the arguments later, in the second callback. You
-can use the :attr:`Request.meta` attribute for that.
-
-Here's an example of how to pass an item using this mechanism, to populate
-different fields from different pages::
+functions so you can receive the arguments later, in the second callback.
+The following two examples show how to achieve this by using the 
+:attr:`Request.meta` and :attr:`Request.cb_kwargs` attributes respectively::
```
Gallaecio: I think the documentation should cover only the usage of `Request.cb_kwargs`, and do not mention `meta` but for those cases where it is the only possible way (e.g. parameters for middleware).

docs/topics/request-response.rst
```diff
@@ -219,6 +231,22 @@ different fields from different pages::
...
         yield item
 
+::
+
+    def parse_page1(self, response):
+        item = MyItem()
+        item['main_url'] = response.url
+        request = scrapy.Request("http://www.example.com/some_page.html",
+                                 callback=self.parse_page2)
+        request.cb_kwargs['item'] = item
```
Gallaecio: I see you followed the same pattern as the existing example about `meta`, however I wonder if the example should use the constructor parameter instead of using the object property. Or at least both should be covered (one field set on each).

docs/topics/request-response.rst
```diff
@@ -219,6 +231,22 @@ different fields from different pages::
...
+        request.cb_kwargs['foo'] = 'bar'
+        yield request
+
+    def parse_page2(self, response, item, foo):
+        item['other_url'] = response.url
+        item['foo'] = foo
```
Gallaecio: I see a slight problem with this example: It would have made more sense to define `item['foo'] = foo` on `parse_page1`, since you are already instantiating an item.

I’m thinking we could switch to something like:

```
    def parse_page1(self, response):
        # You can pass a cb_kwargs dictionary to a Request constructor:
        request = scrapy.Request("http://www.example.com/some_page.html",
                                 callback=self.parse_page2,
                                 cb_kwargs={'main_url': response.url})
        # You can alternatively modify the cb_kwargs property of an existing Request object:
        request.cb_kwargs['title'] = response.css('title::text').get()
        yield request

    def parse_page2(self, response, title, main_url):
        yield {'title': title, 'main_url': main_url, 'other_url': response.url}
```

docs/topics/request-response.rst
elacuesta: The original issue talks about "an _alternative_ way to pass arguments to callbacks". Do we want to encourage users to absolutely stop using `Request.meta` to pass data around?

docs/topics/request-response.rst
elacuesta: Updated, thanks.

docs/topics/request-response.rst
ejulio: I think so.
`cb_kwargs` should be the option for callback data.
`Request.meta` should be the option for other components like _middlewares_.

kmike: What do you think about mention meta, but saying that in previous Scrapy versions this style was common, but now we encourge you to use callback kwargs for user data, and meta for components like middlewares? People will be reading a lot of code which uses meta, this may help them.

tests/spiders.py
```diff
@@ -28,6 +28,45 @@ def closed(self, reason):
         self.meta['close_reason'] = reason
 
 
+class KeywordArgumentsSpider(MockServerSpider):
```
kmike: Could you please add a test which checks how Scrapy behaves if there is a mismatch between parameters `parse` accept and parameters passed via callback kwargs, e.g. some required argument is missing, or an extra argument is passed? Another case which could be nice to check explicitly is how default values are handled.

kmike:
By searching our docs for "meta" it is possible to find a few other places where docs may need an update:

* examples at https://docs.scrapy.org/en/latest/topics/debug.html?highlight=meta
* it'd be nice to mention this new feature in addition to meta in https://docs.scrapy.org/en/latest/topics/leaks.html?highlight=meta; example is better to be updated as well;
* example at https://docs.scrapy.org/en/latest/topics/jobs.html?highlight=meta
* `scrapy parse` command supports `--meta` - do we need to support arguments as well, to make transition smooth? (https://docs.scrapy.org/en/latest/topics/commands.html?highlight=meta)
* Docs for Request callback argument may mention this feature as well (technically it is still correct though)
* Request.replace docs only say that meta is copied, while the cb_kwargs behavior is the same

I could miss something - it'd be great to check all cases "meta" is mentioned somewhere.

kmike:
All other things equal, I'd prefer either `kwargs` or `callback_kwargs` over `cb_kwargs`, but the fact we use this name already in Rule is a good argument to go with `cb_kwargs` name.

tests/spiders.py
elacuesta: Added tests for defaults and argument mismatch.
Errback is not called with this current implementation, it does if we add the callback and the errback to the Deferred in two separate steps, i.e.
```python
dfd.addCallback(request.callback or spider.parse, **request.cb_kwargs)
dfd.addErrback(request.errback)
```
instead of
```python
dfd.addCallbacks(
    callback=request.callback or spider.parse,
    errback=request.errback,
    callbackKeywords=request.cb_kwargs)
```
In any case, the `Request` object is not bounded to the `Failure` received by the errback. Personally, I don't think calling the errback would be appropriate here, since it's not an error with the request/response itself, but with the code that handles it. The logged error is very descriptive, and similar to the one that currently appears when, for instance, a callback does not take a second positional argument (`TypeError: parse() takes 1 positional argument but 2 were given`).

docs/topics/request-response.rst
elacuesta: Updated, please take another look everyone :eyes:

elacuesta:
Updated/replaced several occurences of `Request.meta` accros the docs.
I'm not sure why the Codecov check is failing, I added a test for the `--cb_kwargs` option in the `parse` command :thinking:

docs/topics/commands.rst
```diff
@@ -461,6 +461,9 @@ Supported options:
 * ``--meta`` or ``-m``: additional request meta that will be passed to the callback 
   request. This must be a valid json string. Example: --meta='{"foo" : "bar"}'
 
+* ``--cb_kwargs``: additional keyword arguments that will be passed to the callback.
```
Gallaecio: I wonder if we should apply a CLI-specific naming convention to the parameter there. I’m used to hyphens (``--cb-kwargs``), although skipping separators altogether seems to be the current convention (``--cbkwargs``).

I also wonder if we should even deviate from the Python parameter name here, and use the whole word (‘callback’).

Just thinking out loud, though. I’m don’t oppose merging with `--cb_kwargs`.

docs/topics/request-response.rst
```diff
@@ -165,19 +168,30 @@ Request objects
...
 
-    .. method:: Request.replace([url, method, headers, body, cookies, meta, encoding, dont_filter, callback, errback])
+    .. method:: Request.replace([url, method, headers, body, cookies, meta, flags, encoding, priority, dont_filter, callback, errback, cb_kwargs])
 
        Return a Request object with the same members, except for those members
        given new values by whichever keyword arguments are specified. The
-       attribute :attr:`Request.meta` is copied by default (unless a new value
-       is given in the ``meta`` argument). See also
+       :attr:`Request.cb_kwargs` and :attr:`Request.meta` attributes are copied by default
+       (unless new values are given as arguments). See also
```
Gallaecio: I know this is an issue of the original string, but since you are modifying the string, maybe it should also mention that the copy is shallow, as described above.

docs/topics/request-response.rst
```diff
@@ -200,12 +214,37 @@ Example::
...
 In some cases you may be interested in passing arguments to those callback
-functions so you can receive the arguments later, in the second callback. You
-can use the :attr:`Request.meta` attribute for that.
+functions so you can receive the arguments later, in the second callback.
+The following example shows how to achieve this by using the
+:attr:`Request.cb_kwargs` attribute:
+
+::
+
+    # pass information to the next callback using the Request.cb_kwargs attribute
```
Gallaecio: This comment seems a bit unnecessary to me.

docs/topics/request-response.rst
```diff
@@ -200,12 +214,37 @@ Example::
...
+    def parse(self, response):
+        request = scrapy.Request('http://www.example.com/index.html',
+                                 callback=self.parse_page2,
+                                 cb_kwargs=dict(main_url=response.url))
```
Gallaecio: Why `dict()` instead of `{}`? Same below.

docs/topics/request-response.rst
```diff
@@ -200,12 +214,37 @@ Example::
...
+            foo=foo,
+        )
 
-Here's an example of how to pass an item using this mechanism, to populate
-different fields from different pages::
+.. caution:: :attr:`Request.cb_kwargs` was introduced in version ``1.7``.
+   Prior to that, :attr:`Request.meta` was the recommended option for passing
+   information around callbacks. However, after ``1.7``, using :attr:`Request.cb_kwargs`
+   became the preferred way of passing user information, leaving :attr:`Request.meta`
+   to be populated by internal components like spider or downloader middlewares.
```
Gallaecio: > leaving :attr:`Request.meta` to be populated by internal components like spider or downloader middlewares

I think this part is a bit misleading. It sounds like users will never use meta from a spider. It should sound like they are going to write and read meta fields only to exchange information with middlewares and extensions. And I would not call those components internal, since users are expected to write them themselves.

Of course, this does not matter if we remove the outdated documentation altogether.

docs/topics/request-response.rst
```diff
@@ -200,12 +214,37 @@ Example::
...
+    def parse_page2(self, response, main_url, foo):
+        yield dict(
+            main_url=main_url,
+            other_url=response.url,
+            foo=foo,
+        )
 
-Here's an example of how to pass an item using this mechanism, to populate
-different fields from different pages::
+.. caution:: :attr:`Request.cb_kwargs` was introduced in version ``1.7``.
```
Gallaecio: I’m thinking that we could just leave the warning about the version where cb_kwargs was introduced. If users realize that they are not using Scrapy 1.7, they should switch to the right version of the documentation anyway.

docs/topics/commands.rst
elacuesta: Makes sense, `--cbkwargs` would be coherent with the current `--nolinks`, `--noitems`, `--nocolour` options.

docs/topics/request-response.rst
elacuesta: :ok_hand:

docs/topics/request-response.rst
elacuesta: True, I forgot about the possibility of navigating older versions, thanks for the suggestion.

docs/topics/request-response.rst
elacuesta: :ok_hand:

docs/topics/request-response.rst
elacuesta: Why not?

docs/topics/request-response.rst
elacuesta: :ok_hand:

kmike:
Thanks for the work @elacuesta, and thanks for a careful review @Gallaecio and @ejulio! The PR looks great. I think we can merge it without errback kwargs support, and discuss this feature separately. 

May I ask for one additional test though? It would be awesome if a middleware can change request.cb_kwargs, and changes have an effect; it is not clear if this is supported.

This feature would enable some cool use cases, mostly related to dependency injection, similar to how pytest fixtures work. For example, it'd be possible to implement this:

```py
    # ...
    def parse(self, response):
        # ... business as usual

    def parse_page(self, response, cookiejar: scrapy.CookieJar):
        # we ask for a current cookiejar object;
        # cookie middleware inspects `callback.__annotations__`, figures out 
        # user wants to read / write current cookies, and provides 
        # (injects) cookiejar object.
```

elacuesta:
@kmike Added tests for downloader/spider middlewares.
That CookieJar injection idea is awesome, and not too hard to implement (https://gist.github.com/elacuesta/edfb297fdb0eaa0e5e415835c148c564), but it does require to override the default Cookies middleware because it is the only one that knows about cookies. Would you consider a PR to add the contents of the above gist to Scrapy itself? There should be no version-specific syntax problems if we use `getattr` to get the annotations.

kmike:
That's awesome it works @elacuesta, thanks for checking it and adding more tests! 

Regarding actually implementing cookiejar feature, it may need a bit of thought. For example, @pawelmhm raised a good point in https://github.com/scrapy/scrapy/issues/1878#issuecomment-204300765: cookiejar API is not adequate, it is hard to use. So probably if we make this feature built-in, we may need to have some wrapper to make working with cookies straightforward. We may also think if that's possible to fix some other issues using the same API: e.g. how to "fork" a cookiejar (copy current cookies, but update cookiejars separately afterwards). 

It seems this all needs a proposal and a separate discussion. Some starter implementation can aid discussion, but overall it looks more complex than making a PR out of gist.

It'd probably be the easiest to release this middleware as a gist snippet, or a small Python package in the meantime, so that people can start using it, while waiting for a solution in Scrapy itself.

elacuesta:
Thanks! I thought about making a small package with that gist, I'll do that :rocket:

PR #3568, opened 2019-01-07
Add text sitemap support

arimbr:
I have added some changes to support text sitemaps that are part of the standard based on https://www.sitemaps.org/protocol.html

When generating myself a sitemap for some sites I am crawling I find it easier to pass to a spider a `.txt` file that contains one URL per line than generating a `.xml` file.

For example in https://www.xml-sitemaps.com/urllist.txt:
```
https://www.xml-sitemaps.com/
https://www.xml-sitemaps.com/standalone-google-sitemap-generator.html
https://www.xml-sitemaps.com/about-sitemaps.html
...
```

I could then do:

```
class MySitemapSpider(SitemapSpider):
     name = 'mysitemapspider'
     sitemap_urls = ('https://www.xml-sitemaps.com/urllist.txt',)
```

Is that something that you would be interested to add support for?

### DIFF
--- a/scrapy/spiders/sitemap.py
+++ b/scrapy/spiders/sitemap.py
@@ -49,6 +49,13 @@ class SitemapSpider(Spider):
                 logger.warning("Ignoring invalid sitemap: %(response)s",
                                {'response': response}, extra={'spider': self})
                 return
+            if response.url.endswith('.txt'):
+                for loc in iterloc_text_sitemap(response.text):
+                    for r, c in self._cbs:
+                        if r.search(loc):
+                            yield Request(loc, callback=c)
+                            break
+                return
 
             s = Sitemap(body)
             it = self.sitemap_filter(s)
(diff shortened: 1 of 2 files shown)


scrapy/spiders/sitemap.py
```diff
@@ -49,6 +49,13 @@ def _parse_sitemap(self, response):
                 logger.warning("Ignoring invalid sitemap: %(response)s",
                                {'response': response}, extra={'spider': self})
                 return
+            if response.url.endswith('.txt'):
+                for loc in iterloc_text_sitemap(response.text):
```
Gallaecio: :lipstick:  This does not seem worth a new function to me:

```suggestion
                for loc in response.text.splitlines():
```

tests/test_spider.py
```diff
@@ -348,6 +353,25 @@ def test_get_sitemap_urls_from_robotstxt(self):
                           'http://example.com/sitemap-uppercase.xml',
                           'http://www.example.com/sitemap-relative-url.xml'])
 
+    def test_iterloc_text_sitemap(self):
+        text = """http://example.com/1
+http://example.com/2
+"""
+        self.assertEqual(list(iterloc_text_sitemap(text)),
```
Gallaecio:
```suggestion
        self.assertEqual(text.splitlines(),
```

scrapy/spiders/sitemap.py
```diff
@@ -49,6 +49,13 @@ def _parse_sitemap(self, response):
...
+                    for r, c in self._cbs:
```
Gallaecio: :lipstick: This part might, though, as it’s repeated below.

```
for r, c in self._cbs:
    if r.search(loc):
        yield Request(loc, callback=c)
        break
```

Could be extracted as:

```
    def _url_request(self, url):
        for regex, callback in self._cbs:
            if regex.search(url):
                return Request(url, callback=callback)

    # …

                    request = self._url_request(loc)
                    if request:
                        yield request
```

Gallaecio (commented):
Nice improvement!

Gallaecio (approved):

PR #3578, opened 2019-01-12
Deprecation removals for Scrapy 1.7

nyov:
Removing deprecations of 2015 and prior (pre-1.1).


nyov:
Damn hub client didn't take my text :( So here goes:

Here is a list of current deprecations that have been around since before 2016, in some form, according to git blame (Scrapy 1.1 and before); some even longer, dating back to 2010.

I am not sure how long it makes sense to keep them around, certainly legacy code benefits from the ability to continue to run on newer Scrapy versions unchanged,
but I also feel that it might prevent the codebase from further evolving after a certain point, as it becomes a hassle to track the deprecated APIs.
(Can't see further possible improvements with all the half-dead code around ;)

Let me know what you think about these proposals, I was too lazy to do feature-(un)commits for this, but it's certainly not all-or-nothing.

scrapy/selector/unified.py
```diff
@@ -104,16 +87,6 @@ def __init__(self, response=None, text=None, type=None, root=None, _root=None, *
...
     # Deprecated api
-    @property
-    def _root(self):
-        warnings.warn("Attribute `_root` is deprecated, use `root` instead",
-                      ScrapyDeprecationWarning, stacklevel=2)
-        return self.root
-
     @deprecated(use_instead='.xpath()')
-    def select(self, xpath):
+    def x(self, xpath):
```
kmike: what's .x?

kmike:
+1 to deprecate this all for 1.7 release.

nyov:
`.x` is a shorthand on `SelectorList`, which I removed and then decided not to. I must have mistakenly re-added it to `Selector` as well. Would you say that I should remove it, or is it a convenient one-letter shorthand to keep?

kmike:
@nyov I think we should remove .x, it was deprecated ~10 years ago: https://github.com/scrapy/scrapy/commit/48b40bd620319c80cd1196058d7e241a1b6f9307#diff-15415c3ac7385abcfb0fb020668d7ec5

nyov:
@kmike, done.

I have one further undecided change, regarding `S3FeedStorage`:

For the intended 1.6 target, I changed the `from scrapy.conf import settings` line into  
`from scrapy.utils.project import get_project_settings`, though perhaps for 1.7 it is possible to remove the whole codeblock?
https://github.com/scrapy/scrapy/blob/55debc93aa94118d374cf43b4a6e0c4536ba6613/scrapy/extensions/feedexport.py#L100-L114
That would be useful to me since I don't know how to change the test to `mock.patch` `get_project_settings` in `tests/test_feedexport.py`:
```
-    @mock.patch('scrapy.conf.settings', new={'AWS_ACCESS_KEY_ID': 'conf_key',
-                'AWS_SECRET_ACCESS_KEY': 'conf_secret'}, create=True)
[...]
-        # Backwards compatibility for initialising without settings
-        with warnings.catch_warnings(record=True) as w:
-            storage = S3FeedStorage('s3://mybucket/export.csv')
-            self.assertEqual(storage.access_key, 'conf_key')
-            self.assertEqual(storage.secret_key, 'conf_secret')
-            self.assertTrue('without AWS keys' in str(w[-1].message))
```
Then I could just drop the test. :)

kmike:
Thanks @nyov!

I'm not sure we should drop S3FeedStorage backwards compatibility code yet, because it was added in 1.6 (see https://github.com/scrapy/scrapy/pull/3348).

nyov:
Ah, okay. I was only checking the blame dates, my bad.
But I don't know how to fix the testcase, sorry. (Strangely it only fails on Py2.7/jessie tox env, after removing the `mock.patch` line, perhaps it's broken anyhow?)

Gallaecio:
@nyov Could you please fix the conflicts introduced by other commits?

nyov:
Ok, done.
Build looks ok,
Those conflicts were only changes to code which is removed by this commit anyway.

kmike:
Thanks @nyov!

nyov:
And, as always, thank you guys for reviewing.

PR #3269, opened 2018-05-22
 Fix FormRequest.formdata with GET method duplicates same key in querystring (#2919)

iamriel:
This is the fix for issue #2919 This is my first time contributing to an open source project.  This PR is based on gekco's PR [here](https://github.com/scrapy/scrapy/pull/3151).  I followed the instruction in [Submitting Patches](https://doc.scrapy.org/en/master/contributing.html#submitting-patches) for updating an existing PR.  Hopefully I have done it correctly.

Issue #2919, opened 2017-09-10
FormRequest.formdata with GET method duplicates same key in query string


talhashraf:
```
>>> from scrapy import FormRequest
>>> FormRequest('http://example.com/?id=9', method='GET', formdata={'id': '8'}).url
'http://example.com/?id=9&id=8'
```
Notice how `id` is being duplicated in query string. It must replace the value if same key is found. I believe this can be handled in [form.py#L36](https://github.com/scrapy/scrapy/blob/master/scrapy/http/request/form.py#L36).

kmike:
Yes, this looks like a bug.

jslay88:
Hopefully that works for you. First time making a PR, so please excuse any errors.

keremgocen:
anyone working on this?

Gallaecio:
I’m not sure this should be considered a bug. Specially after giving some though to https://github.com/scrapy/scrapy/pull/4730.

Gallaecio:
Fixed by #3579

PR #3579, opened 2019-01-13
[MRG+1] Issue #2919: Fix FormRequest.formdata with GET method duplicates same key in query string

maramsumanth:
Solved https://github.com/scrapy/scrapy/issues/2919.

Combining ```querystr``` and ```formdata``` into dictionary, and thereby eliminating the duplicates. @kmike , @lopuhin , please review. 
I have also corrected the code, by creating a list where we want to pass ```duplicate keys```. as described in comment https://github.com/scrapy/scrapy/pull/3269#issuecomment-412260614

I am currently preparing for GSoC 2019, please guide me if there are any mistakes/suggestions. I would be glad if you give me suggestions for getting selected in scrapy.
Thanks in advance :) 👍


maramsumanth:
@lopuhin , can you please review this ? :)

maramsumanth:
@lopuhin , what can be the reason here for failure of codecov/patch ?

maramsumanth:
@elacuesta , could you review ?

maramsumanth:
@kmike , can you review this PR? :)

elacuesta:
I couldn't take a deep look into the algorithm itself, these are just a few Python suggestions to improve readibility:

1. There are some duplicated occurences of `urlsplit(self.url)`, you could assign that to a variable instead. `urlsplit`

2. It's better to iterate sequences directly instead of by index, i.e.
```python
In [1]: keys = ['a', 'b', 'c']                                                                                                                                                                                                                                                  

In [2]: for k in keys:  # good 
   ...:     print(k)                                                                                                                                                                                                                                                            
a
b
c

In [3]: for i in range(len(keys)):  # not so good 
   ...:     print(keys[i])                                                                                                                                                                                                                                                      
a
b
c
```

3. If you need to iterate at the same time over more than one iterator you could use [`zip`](https://docs.python.org/3.6/library/functions.html#zip)
```python
In [4]: values = [1, 2, 3]                                                                                                                                                                                                                                                      

In [5]: for k, v in zip(keys, values): 
   ...:     print('{}={}'.format(k, v))                                                                                                                                                                                                                                         
a=1
b=2
c=3
```

Finally, if I'm not mistaken the Codecov failure is because some of the new lines are not hit by any of the existing tests, that's why it says "will decrease coverage by 0.04%"

maramsumanth:
@elacuesta , thanks for the suggestion :) 👍. I have modified code to improve readability.

scrapy/http/request/form.py
```diff
@@ -33,7 +33,25 @@ def __init__(self, *args, **kwargs):
...
-                self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
+                url_split = urlsplit(self.url)
+                formdata_key_list = []
+                for k in querystr.split('&'):
+                    formdata_key_list.append(k.split('=')[0])
+
+                if url_split.query:
+                    queries = (url_split.query).split('&')
+                    query_dict = {}
+                    for x in queries:
```
amarynets: Why are you not using `parse_qs` or `parse_qsl` for parsing and create `query_dict ` dict?

scrapy/http/request/form.py
```diff
@@ -33,7 +33,25 @@ def __init__(self, *args, **kwargs):
...
+                        k = x.split('=')[0]
+                        v = x.split('=')[1]
+                        if formdata_key_list.count(k)==0:
+                            query_dict[k] = v
+                        query_str = ''
+                        for k, v in query_dict.items():
+                            query_str += (k + '=' + v + '&')
```
amarynets: Also you can use `urlencode` for creating qs from dict

scrapy/http/request/form.py
```diff
@@ -33,7 +33,25 @@ def __init__(self, *args, **kwargs):
...
+                    self._set_url(self.url[:self.url.index('?')+1] + query_str + querystr + ('#' + url_split.fragment if url_split.fragment else ''))
```
amarynets: I think would be better to use `urllib.parse.urlunparse()` for create new URL instead concatenation.

maramsumanth:
Done @amarynets. Thanks for your suggestions! :) 👍

maramsumanth:
@kmike @lopuhin , can you please look into this?

maramsumanth:
@lopuhin @elacuesta , can you look into this?

maramsumanth:
@dangra , can you review? Thanks.

maramsumanth:
I have made the necessary changes, @kmike can you review it? :)

maramsumanth:
@lopuhin , can you please look into this?

scrapy/http/request/form.py
```diff
@@ -33,7 +33,14 @@ def __init__(self, *args, **kwargs):
                 self.headers.setdefault(b'Content-Type', b'application/x-www-form-urlencoded')
                 self._set_body(querystr)
             else:
-                self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
+                url_split = urlsplit(self.url)
+                formdata_key_list = []
+                for k in querystr.split('&'):
+                    formdata_key_list.append(k.split('=')[0])
+                items = []
+                items += [(k, v) for k, v in parse_qsl(url_split.query) if k not in formdata_key_list]
```
lopuhin: Why create an empty ``items`` here and then add to this list? Can this be reduced to one operation?

scrapy/http/request/form.py
```diff
@@ -33,7 +33,14 @@ def __init__(self, *args, **kwargs):
                 self.headers.setdefault(b'Content-Type', b'application/x-www-form-urlencoded')
                 self._set_body(querystr)
             else:
-                self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
+                url_split = urlsplit(self.url)
+                formdata_key_list = []
+                for k in querystr.split('&'):
```
lopuhin: It's better to use functions for URL parsing from standard library and w3lib instead of doing this manually here, unless they are not doing what you need here (in that case it would make sense to leave a comment)

scrapy/http/request/form.py
```diff
@@ -33,7 +33,14 @@ def __init__(self, *args, **kwargs):
...
+                    formdata_key_list.append(k.split('=')[0])
+                items = []
+                items += [(k, v) for k, v in parse_qsl(url_split.query) if k not in formdata_key_list]
+                query_str = _urlencode(items, self.encoding)
+                self._set_url(urljoin(self.url,'?'+ (query_str + '&' if query_str else '') + querystr + ('#'+ url_split.fragment if url_split.fragment else '')))
```
lopuhin: Same here, I think it's better to use stdlib functions for URL construction from parts (see e.g. `urlunsplit` / `urlunparse`)

lopuhin:
```
http://www.example.com/?a=0&a=2&b=1&c=3#fragment
data = {'a': 1}
```
> Looks like it's wrong request. I tested it by Postman and got 400 status code
> As far as I know all data from GET form should be sent by query string and duplicates of keys can exist because on the server it will be transformed to list. More detail info on SO

@amarynets thanks for review! Could you please clarify what you mean here? I agree that duplicate GET arguments are possible and we should preserve them, but not add new duplicates. Do you mean it's not handled correctly in this PR? If yes, could you please clarify which case is not handled correctly?

@maramsumanth I see that some duplicate cases are checked in tests, but could you please add an explicit test for the case then there is a duplicate GET argument and that it's preserved when we add another unlreated argument? (it seems this is not tested directly).

scrapy/http/request/form.py
maramsumanth: Here ```items``` is a local variable, since this is the process of appending to a empty list. If we write simply ```items+=[........]``` it will be error as ```local variable referenced before assignment```

maramsumanth: :+1:

scrapy/http/request/form.py
maramsumanth: I am not be able to do that, because I have ```query_str``` and ```querystr``` to be appended. We can also have the case where ```query_str can be empty```. So, I explicitly defined in the urljoin. Do you have any suggestion @lopuhin ?

maramsumanth:
@lopuhin Thanks for your review. I have made the changes :) :+1:

scrapy/http/request/form.py
```diff
@@ -33,7 +33,11 @@ def __init__(self, *args, **kwargs):
                 self.headers.setdefault(b'Content-Type', b'application/x-www-form-urlencoded')
                 self._set_body(querystr)
             else:
-                self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
+                url_split = urlsplit(self.url)
+                formdata_key_list = list(dict(parse_qsl(querystr)).keys())
```
lopuhin: Since this is used only for checking filtering in the next line, it would be more efficient with a set instead of a list:
```
formdata_keys = set(dict(parse_qsl(querystr)).keys())
```

scrapy/http/request/form.py
lopuhin: > I am not be able to do that, because I have both query_str and querystr to be appended. We can also have the case where query_str can be empty.

@maramsumanth I see, in that case you could still fist concatenate query strings and pass them to urlunsplit, it already has the logic to handle empty parts:

```
In [1]: from urllib.parse import urlunsplit

In [2]: urlunsplit(('a', 'b', 'c', 'd', 'e'))
Out[2]: 'a://b/c?d#e'

In [3]: urlunsplit(('a', 'b', 'c', 'd', ''))
Out[3]: 'a://b/c?d'

In [4]: urlunsplit(('a', 'b', 'c', 'd', ''))
Out[4]: 'a://b/c?d'

In [5]: urlunsplit(('a', 'b', '', '', 'foo'))
Out[5]: 'a://b#foo'
```

maramsumanth:
@lopuhin , changes made :+1: , but it is becoming much lengthy, is it ok?

scrapy/http/request/form.py
maramsumanth: I thought of using urlunsplit, but both urljoin and urlunsplit yield the same thing.

maramsumanth:
@whalebot-helmsman @lopuhin. I have made the necessary changes, can you review?

whalebot-helmsman:
I think code will be much easier to read if you rename `querystr` to `form_query_str` and `query_str` to `url_query_str`

maramsumanth:
Done @whalebot-helmsman :+1:

whalebot-helmsman:
@amarynets @lopuhin @maramsumanth I think `FormRequest` behaviour should mirror browser behaviour. To test browser behaviour I create simple HTML file https://gist.githubusercontent.com/whalebot-helmsman/a6c3d2eb78ae57a072af11f8b726ff83/raw/3f281be8ce0f08ea5d81deb443c816b6fbe846f3/forms.html. You can download local copy and test. In case of GET form request both Chrome and Firefox drop `url_query_str` from URL first and add `form_query_str`.

whalebot-helmsman:
@maramsumanth Based on browser behaviour I proposed to change implementation in current PR. You up voted this proposal and didn't apply proposed changes. Is there something what prevent you from applying  these changes?

maramsumanth:
Thanks for reminding me @whalebot-helmsman .I was working on the Redirect Middleware part of scrapy.
Coming to this PR ,I was little bit confused. In the following example:-
```FormRequest('http://example.com/?id=9', method='GET', formdata={'id': '8'}).url``` should give
```'http://example.com/?id=8'``` i.e it should replace the value if same key is found. I tested in chrome 
   and it gave me the correct output. Can you give me detailed explanation of what is the problem. :)
Thanks

**Comment by maramsumanth** [comment]
@whalebot-helmsman , can you please look into this so that I can solve it ASAP. :+1:

**Comment by whalebot-helmsman** [comment]
I referenced HTML file as example. It has two forms. First one uses GET request, has `a` and `b` as keys  and targets http://example.com/f?a=1&d=2 URL. After you click `Submit` button you get to http://example.com/f?a=3&b=2. Browser not only replace value for `a` key, but drop value for `d` key. Reformulating in terms of python structures we got
```
FormRequest('http://example.com/?id=9&smth=a', method='GET', formdata={'id': '8'}).url == 'http://example.com/?id=8'
```

**Comment by maramsumanth** [comment]
I have Ubuntu OS in which I am using chromium browser.
But for ```FormRequest('http://example.com/?id=9&smth=a', method='GET', formdata={'id': '8'}).url```
my output is ```'http://example.com/?smth=a&id=8'``` which is != ```'http://example.com/?id=8'```

**Comment by maramsumanth** [comment]
@whalebot-helmsman , when I tested the HTML file you provided,  I can see that in case of GET form request both Chrome and Firefox drop ```url_query_str``` from URL and add ```form_query_str```. But running through the command line as described in the above comment, I cannot see any problem.

**Comment by whalebot-helmsman** [comment]
>I can see that in case of GET form request both Chrome and Firefox drop `url_query_str` from URL and add `form_query_str`. 

Scrapy and related libraries should mirror browser behaviour. 
At this moment in master and in your branch
```
>>> from scrapy.http.request.form import FormRequest;FormRequest('http://example.com/?id=9&smth=a', method='GET', formdata={'id': '8'}).url == 'http://example.com/?id=8'
False
```
This is incorrect representation of browser behaviour. I understand, such behaviour was implemented before your PR. As we started to work on this part we should also fix this.

**Comment by maramsumanth** [comment]
Is there any python code to get the default browser the user is having @whalebot-helmsman ?

**Comment by whalebot-helmsman** [comment]
> @whalebot-helmsman , when I tested the HTML file you provided, I can see that in case of GET form request both Chrome and Firefox drop `url_query_str` from URL and add `form_query_str`. 

It is a browser behaviour. We agreed on that. 

Scrapy implementation is separate from installed browser. Scrapy behaviour doesn't depend on installed browser. In master branch we preserve both `url_query_str` and `form_query_str`, In branch for this PR we replace values for common keys in `url_query_str` by values from `form_query_str`.
Scrapy's behaviour in both branches doesn't mirror browser behaviour. It is a problem, we should fix it.

I propose to fix it in this PR. 

Current behaviour
```
>>> from scrapy.http.request.form import FormRequest;FormRequest('http://example.com/?id=9&smth=a', method='GET', formdata={'id': '8'}).url
'http://example.com/?smth=a&id=8'
```
Behaviour after fix
```
>>> from scrapy.http.request.form import FormRequest;FormRequest('http://example.com/?id=9&smth=a', method='GET', formdata={'id': '8'}).url
'http://example.com/?id=8'
```

**Comment by maramsumanth** [comment]
@whalebot-helmsman Thanks for guiding me.
Till now I was thinking that, we have to somewhere account for the browser name and proceed accordingly. Now I have updated the PR and changed tests as you suggested :)

**Comment by maramsumanth** [comment]
> Scrapy behaviour doesn't depend on installed browser

@whalebot-helmsman Thanks for guiding me.
Till now I was thinking that, we have to somewhere account for the browser name and proceed accordingly. Now I have updated the PR and changed tests as you suggested :)

**Comment by maramsumanth** [comment]
@whalebot-helmsman , should I change anything or is it safe to merge. :+1: 
Thanks :)

**Comment by whalebot-helmsman** [comment]
It is safe to merge now. But merging takes more time and not required by [GSoC rules](http://python-gsoc.org/index.html#apply). We can keep it as it is for now.

File: tests/test_http_request.py [hunk 8]
```diff
@@ -273,6 +273,18 @@ def assertQueryEqual(self, first, second, msg=None):
     def test_empty_formdata(self):
         r1 = self.request_class("http://www.example.com", formdata={})
         self.assertEqual(r1.body, b'')
+        
+    def test_formdata_overrides_querystring_duplicates(self):
+        data = (('a', 'one'), ('a', 'two'), ('b', '2'))
+        url = self.request_class('http://www.example.com/?a=0&b=1&c=3#fragment', method='GET', formdata=data).url.split('#')[0]
+        fs = _qs(self.request_class(url, method='GET', formdata=data))
+        self.assertEqual(set(fs[b'a']), {b'one', b'two'})
+        self.assertEqual(fs[b'b'], [b'2'])
```
Gallaecio: Maybe also assert that `c` is not in `fs`?

maramsumanth: Also changed test method name :)

maramsumanth:
Now its fine :) :+1:

wRAR (approved):

PR #3588, opened 2019-01-18
[MRG+1] Adding requests referer to RFPDupeFilter log messages

ejulio:
I just changed the log message when DUPEFILTER_DEBUG is active because it helps knowing which request is generating the duplicates (I'd this problem recently).

Also, I added some tests for the log method to check the proper messages and stats.

### DIFF
--- a/scrapy/dupefilters.py
+++ b/scrapy/dupefilters.py
@@ -3,8 +3,7 @@ import os
 import logging
 
 from scrapy.utils.job import job_dir
-from scrapy.utils.request import request_fingerprint
-
+from scrapy.utils.request import referer_str, request_fingerprint
 
 class BaseDupeFilter(object):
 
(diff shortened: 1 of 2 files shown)


dangra:
@kmike if you don't see any objections, I'll merge and ship it in 1.6.0

dangra:
thanks Julio.

scrapy/dupefilters.py
```diff
@@ -61,8 +60,9 @@ def close(self, reason):
 
     def log(self, request, spider):
         if self.debug:
-            msg = "Filtered duplicate request: %(request)s"
-            self.logger.debug(msg, {'request': request}, extra={'spider': spider})
+            msg = "Filtered duplicate request: %(request)s (referer: %(referer)s)"
+            args = {'request': request, 'referer': referer_str(request) }
```
kmike: Hey! Sorry for delay. No objections to me, a nice feature. I was just wondering if Referer should be added to the next message as well, when debug is False.

ejulio: Fair point.
I just added to keep the same behavior, but it makes sense to log referer only in debug mode.

Issue #3590, opened 2019-01-19
Ability to retry a request from inside a spider callback


kasun:
There are situations where websites return 200 responses but the content is not available due to bans or temporal issues which can be fixed by retrying requests.

There should be an easier way to retry requests inside spider callbacks, which should ideally reuse the code in Retry downloader middleware.

I see two approaches for this.

1) Introduce new exception called RetryRequest which can be raised inside a spider callback to indicate a retry. I personally prefer this but the implementation of this is a little untidy due to this bug #220

    ```
    from scrapy.exceptions import RetryRequest

    def parse(self, response):
        if response.xpath('//title[text()="Content not found"]'):
            raise RetryRequest('Missing content')
    ```
2) Introduce a new class RetryRequest which wraps a request that needs to be retried. A RetryRequest can be yielded from a spider callback to indicate a retry

    ```
    from scrapy.http import RetryRequest

    def parse(self, response):
        if response.xpath('//title[text()="Content not found"]'):
            yield RetryRequest(response.request, reason='Missing content')
    ```

Will be sending two PRs for the two approaches. Happy to hear about any other alternatives too.

Gallaecio:
What aspects of the [current approach](https://stackoverflow.com/a/28640118/939364) don’t you like? It can be done with a one-liner (`yield response.request.replace(dont_filter=True)`).

kasun:
@Gallaecio I miss everything the retry middleware currently does.

You can't infinitely do this. you need to count the retries and give up after a certain retry attempts
You need to have stats and log messages

IMO there should be a easier way to do above things than re-implement the same code in every project/spider.

ejulio:
I think the idea is nice, though I think you should focus on only one approach.
IMHO, number 1 is better because it makes more sense to retry when some specific exception happened.
I don't think a special kind of Request would be the best approach here.

Also, I think that we should focus on only one of them because _There should be one-- and preferably only one --obvious way to do it._

apalala:
We frequently, comfortably, and successfully use the `raise RetryRequest("a reason")` approach in BV. It counts and limits retries, logs the _reason_, and, in general, does what it should without much work from the developer. As the OP mentioned, corrupted content over 200 responses is frequent enough.

stav:
What about attaching the middleware routine to the response:

    class RetryMiddleware(scrapy.downloadermiddlewares.retry.RetryMiddleware):

        def process_response(self, request, response, spider):
            """Attach a .retry() method to every response"""
[...]

PR #3591, opened 2019-01-19
Retry exception #3590

kasun:
PR for first approach discussed in #3590

A way to retry requests inside a spider callback by raising an exception

Example
```

def parse(self, response):
```

The implementation is a little untidy due to this bug #220 
Will work on docs and tests if there is agreement on this approach


kasun:
Closing this in favor of https://github.com/scrapy/scrapy/pull/3685

PR #3592, opened 2019-01-19
Retry a request using a request wrapper #3590

kasun:
Second approach discussed in #3590

Ability to retry a request inside a spider callback by yielding a RetryRequest instance which wraps a request to be retried.

example
```

def parse(self, response):
```

Will work on docs and tests if there is agreement on this approach


kasun:
closing this in favor of other approaches.

PR #3607, opened 2019-01-29
[MRG+1] add FEED_STORAGE_S3_ACL setting

victor-torres:
This is similar to the IMAGES_STORE_S3_ACL setting.

Just opened the pull request to validate the idea. 

Still, need to:

- update tests
- update docs

Example usage on `settings.py`:
```
FEED_STORAGE_S3_ACL = 'bucket-owner-full-control'
```


scrapy/extensions/feedexport.py
```diff
@@ -118,6 +118,7 @@ def __init__(self, uri, access_key=None, secret_key=None):
         self.secret_key = u.password or secret_key
         self.is_botocore = is_botocore()
         self.keyname = u.path[1:]  # remove first "/"
+        self.policy = settings.get('FEED_STORAGE_S3_ACL', 'private')
```
lucywang000: I think the `settings` variable is only defined in the `if no_defaults` block above

lucywang000: and I think it's better to use a default value of `None` instead of private since the default value for `polify` kw arg in boto `set_contents_from_file` is also None

```
set_contents_from_file(fp, headers=None, replace=True, cb=None, num_cb=10, policy=None, md5=None, reduced_redundancy=False, query_args=None, encrypt_key=False, size=None, rewind=False)
```

http://boto.cloudhackers.com/en/latest/ref/s3.html#boto.s3.key.Key.set_contents_from_file

dangra: > I think the settings variable is only defined in the if no_defaults block above

that's true, but it also means the policy should be passed from `from_crawler` method as constructor argument defaulting to `None`.

victor-torres:
@dangra I believe it's ready. Could you take another look?

scrapy/extensions/feedexport.py
```diff
@@ -131,18 +132,28 @@ def __init__(self, uri, access_key=None, secret_key=None):
     @classmethod
     def from_crawler(cls, crawler, uri):
         return cls(uri, crawler.settings['AWS_ACCESS_KEY_ID'],
-                   crawler.settings['AWS_SECRET_ACCESS_KEY'])
+                   crawler.settings['AWS_SECRET_ACCESS_KEY'],
+                   crawler.settings.get('FEED_STORAGE_S3_ACL'))
```
ejulio: I don't think you need `settings.get()` here because there is a default value of in `default_settings.py`

scrapy/extensions/feedexport.py
```diff
@@ -131,18 +132,28 @@ def __init__(self, uri, access_key=None, secret_key=None):
...
         file.seek(0)
         if self.is_botocore:
+            kwargs = dict()
+            if self.acl:
+                kwargs.update(dict(ACL=self.acl))
+
             self.s3_client.put_object(
-                Bucket=self.bucketname, Key=self.keyname, Body=file)
+                Bucket=self.bucketname, Key=self.keyname, Body=file,
+                **kwargs)
```
ejulio: Any reasons to use `**kwargs` instead of simple method call with a named argument?

tests/test_feedexport.py
```diff
@@ -186,6 +187,138 @@ def test_store(self):
...
+        crawler = get_crawler(settings_dict=settings)
+        storage = S3FeedStorage.from_crawler(
+            crawler,
+            's3://mybucket/export.csv'
+        )
+        self.assertEqual(storage.access_key, 'access_key')
+        self.assertEqual(storage.secret_key, 'secret_key')
+        self.assertEqual(storage.acl, 'custom-acl')
+
+    def test_store_in_thread_botocore_without_acl(self):
```
ejulio: I don't know if you should be calling the private method in a test.
The whole idea is to test the API, so calling `store` would make more sense and easier for future maintenance.
If there is a problem with threading, you can decorate the test with `@defer.inlineCallbacks`.

scrapy/extensions/feedexport.py
victor-torres: I'm just trying to avoid changing the previous behavior. I'm not sure about the side-effects of explicitly passing `ACL` or `policy` arguments even if they're defined as `None`. That's because the `ACL` argument, for example, is not a named argument, but loaded via `kwargs` in the `put_object` method.

scrapy/extensions/feedexport.py
victor-torres: Makes sense now that I introduced default values in the `default_settings.py` file.

tests/test_feedexport.py
victor-torres: It's true, I shouldn't be testing private methods.

victor-torres:
@ejulio I have just updated the source code after your suggestions. Could you take another look, please?

scrapy/extensions/feedexport.py
ejulio: Ok. Got it.
@dangra can you share some thought here as well?

tests/test_feedexport.py
```diff
@@ -186,6 +187,158 @@ def test_store(self):
...
+        )
+        self.assertEqual(storage.access_key, 'access_key')
+        self.assertEqual(storage.secret_key, 'secret_key')
+        self.assertEqual(storage.acl, None)
+
+        def _defer(f, *args, **kwargs):
+            return f(*args, **kwargs)
+
+        with mock.patch('botocore.client.BaseClient._make_api_call') as m:
+            with mock.patch('twisted.internet.threads.deferToThread',
```
ejulio: I don't think you need to mock `twisted`, but you can `yield storage.store()` and decorate the test with `@defer.inlineCallbacks`

tests/test_feedexport.py
```diff
@@ -186,6 +187,158 @@ def test_store(self):
...
+            'access_key',
+            'secret_key',
+            'custom-acl'
+        )
+        self.assertEqual(storage.access_key, 'access_key')
+        self.assertEqual(storage.secret_key, 'secret_key')
+        self.assertEqual(storage.acl, 'custom-acl')
+
+        storage.is_botocore = False
+        storage.connect_s3 = mock.MagicMock()
```
ejulio: Maybe `mock.create_autospec()` is better than `MagicMock` because it could help catching errors if the API changes or if there is a typo in the test/implementation.

tests/test_feedexport.py
victor-torres: Done. I've just pushed these changes.

tests/test_feedexport.py
victor-torres: I don't know. I decided to go with this simple method because I just would like to guarantee we're calling the `put_object` and `new_key` methods with `ACL` and `policy` arguments respectively.

If we change the method names, the code would break because the mocks wouldn't have been called from the test.

Let me know if I am missing anything.

victor-torres:
@ejulio I have just updated the source code. Thanks again for your valuable suggestions 😄

tests/test_feedexport.py
ejulio: It is fine for me, the main concern was _if the API (method names) change, the unit test won't catch these errors and we'll notice them in production or so. But, if we `create_autospec` it will create a mock with the same methods available in the API. So, if the API changes and we call a method that does not exist anymore, we'll get an error._
Again, that is just a thought and I was collecting some ideas as well :)

scrapy/extensions/feedexport.py
```diff
@@ -131,18 +132,28 @@ def __init__(self, uri, access_key=None, secret_key=None):
...
-                   crawler.settings['AWS_SECRET_ACCESS_KEY'])
+                   crawler.settings['AWS_SECRET_ACCESS_KEY'],
+                   crawler.settings['FEED_STORAGE_S3_ACL'])
 
     def _store_in_thread(self, file):
         file.seek(0)
         if self.is_botocore:
+            kwargs = dict()
+            if self.acl:
+                kwargs.update(dict(ACL=self.acl))
```
kmike: dict() is not necessary here, as .update supports keyword arguments. Also, as it is a single value, it can just an assignement: `kwargs['ACL'] = self.acl`, or `kwargs = dict(ALC=self.acl)`. If you go with kwargs=.. assignement, it can be made shorter further, as `kwargs = {'ACL': self.acl} if self.acl else {}`. 

Whether to use if/else assignement is up to stylistic preferences, but I think that'd be good to at least replace `kwargs.update(dict(ACL=self.acl))` with `kwargs['ACL'] = self.acl`.

scrapy/extensions/feedexport.py
```diff
@@ -131,18 +132,28 @@ def __init__(self, uri, access_key=None, secret_key=None):
     @classmethod
     def from_crawler(cls, crawler, uri):
         return cls(uri, crawler.settings['AWS_ACCESS_KEY_ID'],
-                   crawler.settings['AWS_SECRET_ACCESS_KEY'])
+                   crawler.settings['AWS_SECRET_ACCESS_KEY'],
+                   crawler.settings['FEED_STORAGE_S3_ACL'])
```
kmike: would you mind using keyword arguments here, for all optional parameters?

kmike: (long lists of positional arguments are easy to break)

scrapy/settings/default_settings.py
```diff
@@ -158,6 +158,8 @@
 }
 FEED_EXPORT_INDENT = 0
 
+FEED_STORAGE_S3_ACL = None
```
kmike: I'm not 100% sure it is the way to go, but I've suggested using empty strings instead of None in such cases (see https://github.com/scrapy/scrapy/pull/3199#discussion_r181327270), and there is code which does that.

scrapy/extensions/feedexport.py
victor-torres: Done.

scrapy/settings/default_settings.py
victor-torres: I've just read the comment you linked here. I understand that `''` or `None` would produce the same outcome because both values will eval to `False`.

Anyway, I have just replaced its default value to `''` (empty string) to make it consistent with other similar settings.

victor-torres:
Thank you for your review, @kmike. 

I have just implemented your suggestions here, could you take another look, please?

tox.ini
```diff
@@ -47,6 +47,7 @@ deps =
     lxml==3.4.0
     Twisted==14.0.2
     boto==2.34.0
+    botocore==1.12.89
```
kmike: I think we should be dropping jessie support, but if we're keeping it, botocore should be 0.62 (see https://packages.debian.org/en/jessie/python-botocore)

tests/test_feedexport.py
```diff
@@ -186,6 +187,144 @@ def test_store(self):
...
+        storage = S3FeedStorage(
+            's3://mybucket/export.csv',
+            'access_key',
+            'secret_key',
+        )
+        self.assertEqual(storage.access_key, 'access_key')
+        self.assertEqual(storage.secret_key, 'secret_key')
+        self.assertEqual(storage.acl, None)
+
+        with mock.patch('botocore.client.BaseClient._make_api_call') as m:
```
kmike: The heavy use of mocks worries me, especially as we're mocking a private method; tbh I'd prefer not to use mocks at all, and write tests in some other way (including having less of them). 

On the other hand, checking for exact operation names like `PutObject` looks nice here. So I'm ok with merging these tests as-is if others are ok with it, but ideas on how to avoid mocking a private method are welcome.

kmike:
Thanks @victor-torres for a careful implementation! 

I have some concerns about use of mocks in tests, but they are more of personal preferences, so if others are fine with the way tests are written, I'm +1 to merge it, as soon as tests pass. 

Also, thanks @lucywang000 and @ejulio for the reviews, this is really appreciated.

tests/test_feedexport.py
victor-torres: Now that you pointed it's also bothering me. I have just refactored the tests not to use private methods. Thanks.

tox.ini
victor-torres: Updated tox.ini. Waiting for tests to pass. Thanks.

victor-torres: botocore version supported by debian jessie does not contain required methods... I'll just skip this test when in jessie environments.

tests/test_feedexport.py
```diff
@@ -186,6 +186,147 @@ def test_store(self):
...
+            crawler,
+            's3://mybucket/export.csv'
+        )
+        self.assertEqual(storage.access_key, 'access_key')
+        self.assertEqual(storage.secret_key, 'secret_key')
+        self.assertEqual(storage.acl, 'custom-acl')
+
+    @defer.inlineCallbacks
+    def test_store_botocore_without_acl(self):
+        if os.getenv('TOXENV') == 'jessie':
```
kmike: What do you think about skipping a test if botocore is not installed?

victor-torres: If someone accidentally removes botocore from other test environment requirements we could end up skipping this test and therefore hiding some errors.

Leaving it as it is now is also a good point for when someone is removing the jessie support. A quick grep on the source code would highlight this conditional statement and the exception raise line.

victor-torres:
Thank you for your review, @kmike. Tests are passing now.

tests/test_feedexport.py
kmike: I think that it is better to skip the test if botocore is not present because of two reasons:

1. botocore is an optional Scrapy requirement; we should be running tests without it, to make sure Scrapy works without botocore. So it is fine not to have botocore in an environment.
2. People are running Scrapy tests without tox, e.g. on their environments; tests currently are not aware of tox and don't require it - checking for TOXENV variable changes that.

Which methods are missing in botocore? Should we specify a minimal botocore version somewhere (`is_botocore` function? docs?)

victor-torres: Ok, I'll just check the `is_botocore` method then. 

About the other question, I did some research and this is what I found out:

- the botocore version specified in Debian Jessie page is 0.62 (https://github.com/scrapy/scrapy/pull/3607#discussion_r256587948)
- the `Session.create_client` method appeared in version 0.64 (https://github.com/boto/botocore/commit/dd5aad504fd6bbe13e327460f10fbac96cc632b9)
- the `aws_access_key_id` and `aws_secret_access_key` args were introduced in 0.66 (https://github.com/boto/botocore/commit/59a1c336881a618eb321e317ea70beae81017836)
- there was a bug with `_send_request` receive an extra argument solved in 0.69 (https://github.com/boto/botocore/commit/a84a06df9e1df64f1ac96b0b54a1319f6c89881b)

Those versions are pretty old so I believe we shouldn't have problems.

victor-torres:
@kmike I believe someone has to rebuild this last commit. Does not seem to be related to the latest changes.

victor-torres:
> @kmike I believe someone has to rebuild this last commit. Does not seem to be related to the latest changes.

Nervermind, made a blank commit.

tests/test_feedexport.py
ejulio: Just writing a thought here.
I do agree that we should not test private methods, but the interaction with the API we're given.
Otherwise we are getting tests coupled to internals, and what we want is that _given some input, this method will give me some output, no matter how_.

Regarding _mocks_, no problems, just think we should be careful when using them.

:smile:

victor-torres:
@kmike done, tests are passing.

kmike:
👍

docs/topics/feed-exports.rst
```diff
@@ -185,6 +185,10 @@ passed through the following settings:
  * :setting:`AWS_ACCESS_KEY_ID`
  * :setting:`AWS_SECRET_ACCESS_KEY`
 
+You can also define a custom ACL for exported feeds using this setting:
+
+ * :setting:`FEED_STORAGE_S3_ACL`
```
Gallaecio: I think we should consider changing the name to `FEED_ACL`, and follow an approach similar to other feed storage settings here such as `FEED_URI`: The backend-specific documentation is provided on the backend-specific documentation.

Gallaecio: Oh, I see this approach is already followed in other places. I guess any change to that is then a different beast to treat in a different changeset. Lets merge.

kmike:
Thanks @victor-torres for the implementation and addressing all our comments, and thanks @ejulio, @lucywang000, @dangra and @Gallaecio for reviews !

victor-torres:
You're all welcome. Thanks.

PR #3608, opened 2019-02-01
[MRG+1] Fix for #685 Add Google Cloud Storage Feed Export

### LINKED ISSUE #685: Google Cloud Storage Support (Storage backends)
Does it make sense to support [Google Cloud Storage](https://cloud.google.com/products/cloud-storage/) as storage backend? Boto already supports Cloud storage: http://boto.readthedocs.org/en/latest/ref/gs.html

ejulio:
Hey.
This is just a draft to add GCS feed export.
I followed the same behavior as the FilesPipeline, so I'm just using the `GCS_PROJECT_ID` setting and it requires the developer to provide the credentials through environment variables (refer to https://googleapis.github.io/google-cloud-python/latest/core/config.html#authentication for more information).

About the credentials above, I think we could also add a new setting with a path to the credentials file and initialize it using https://googleapis.github.io/google-cloud-python/latest/core/auth.html#service-accounts. I think this might be easier then setting environment variables when deploying.

The list of ToDos, besides your thoughts are:
* [x] Add documentation
* [x] Add `gcs` schema
* [x] Add bucket policy options

Please note that I worked with unit tests here, but I'm open to idea of integration tests. Do you have any thoughts?

Thanks for your time to review this PR :smile: 

(edit) Closes #685


scrapy/extensions/feedexport.py
```diff
@@ -146,6 +146,32 @@ def _store_in_thread(self, file):
...
+
+class GCSFeedStorage(BlockingFeedStorage):
+
+    project_id = None
+    bucket_name = None
+    blob_name = None
+
+    def __init__(self, uri, project_id):
+        self.project_id = project_id
+        u = urlparse(uri)
```
ejulio: Note that these three lines are similar to `S3FeedStorage`.
I thought about extracting them in a method, but it does seem to hurt right now

scrapy/extensions/feedexport.py
```diff
@@ -146,6 +146,32 @@ def _store_in_thread(self, file):
             key.close()
 
 
+
+class GCSFeedStorage(BlockingFeedStorage):
+
+    project_id = None
```
victor-torres: It doesn't seem right to store this information as class attributes. 
You should be defining those values in the init method, don't you?

scrapy/extensions/feedexport.py
```diff
@@ -146,6 +146,32 @@ def _store_in_thread(self, file):
...
+    def from_crawler(cls, crawler, uri):
+        return cls(uri, crawler.settings['GCS_PROJECT_ID'])
+
+    def _store_in_thread(self, file):
+        file.seek(0)
+        from google.cloud.storage import Client
+        client = Client(project=self.project_id)
+        bucket = client.get_bucket(self.bucket_name)
+        blob = bucket.blob(self.blob_name)
+        blob.upload_from_file(file)
```
victor-torres: Could you check for optional Policy options for the bucket? I believe we have something similar on Media Pipelines.

scrapy/extensions/feedexport.py
ejulio: You mean to make it private?
I don't see any difference in:
```
class A:
   attribute = None
   def __init__(self, a):
        self.attribute = a

class A:
   def __init__(self, a):
        self.attribute = a
```

Actually, I prefer the first one because it makes things explicit (there is an attribute called `attribute`).

scrapy/extensions/feedexport.py
ejulio: :+1: Added to the todo list

scrapy/extensions/feedexport.py
victor-torres: Never mind. Mixed some concepts here.

victor-torres: No, I mean to declare them as instance attributes instead of class attributes.
This is how we are doing in `S3FeedStorage`, for example.

Class attributes are really useful when storing constants or helper values, things that are shared across multiple instances of the same class.

In this case, I don't think you should reuse the storage settings.

In worst case scenarios, it could also lead to indesirable mutability when trying to change an instance attribute value causes the class attribute value and therefore all other instance attribute values to be changed too.

In a nutshell, you **need** those values in order to initialize the instance, so there's no reason to keep an unused default value as a class attribute.

scrapy/extensions/feedexport.py
stav: > but it does seem to hurt right now
It DOES hurt? as in... performance?

ejulio: Sorry @stav, I meant _it doesn't_

ejulio:
@victor-torres and @stav .
Just updated the PR with and some of your thoughts.

tests/test_feedexport.py
```diff
@@ -187,6 +188,48 @@ def test_store(self):
...
+        settings = {'GCS_PROJECT_ID': '123', 'FEED_STORAGE_GCS_ACL': 'publicRead' }
+        crawler = get_crawler(settings_dict=settings)
+        storage = GCSFeedStorage.from_crawler(crawler, 'gcs://mybucket/export.csv')
+        assert storage.project_id == '123'
+        assert storage.acl == 'publicRead'
+        assert storage.bucket_name == 'mybucket'
+        assert storage.blob_name == 'export.csv'
+
+    @defer.inlineCallbacks
+    def test_store(self):
```
victor-torres: Would you like to add tests to cover the upload without an ACL/Policy defined (`None`)?

ejulio: Not sure.
The test will be a _copy + paste_ from `test_store` with the only difference in the assignment `acl = None`.
Probably, an option could be parameterized tests.

victor-torres: In this case, you would be testing the default value for the ACL setting.

ejulio:
@dangra, @kmike and @lopuhin can I have a word from you regarding this PR?
Specially the integration/unit test part :D

victor-torres:
@ejulio, I believe we need to cover the `GOOGLE_APPLICATION_CREDENTIALS` environment variable and the `Service Account` credentials in JSON format provided by Google Cloud.

I believe you have tested your scripts locally, but we'll need to specify the credentials somehow when deploying to remote platforms such as Scrapy Cloud. 

We still need to define how to pass these credentials using spider settings. One suitable way would be to encode them using Base64 or any other hash algorithm and defining the value in Scrapy Cloud. We would need to write the decoded Base64 value into a file and define the environment variable with its location during runtime. This would avoid having to version the credentials file.

What do you think?

ejulio:
@victor-torres , I thought about it.
But, since GCS for IMAGES_STORE, currently, only works by reading the credentials from a JSON file, I kept the same behavior.
I think this is something we can work on, but not sure if in this PR, since it would make sense to make this authentication available for IMAGES_STORE as well.

docs/topics/feed-exports.rst
```diff
@@ -185,6 +185,27 @@ passed through the following settings:
...
+
+Google Cloud Storage (GCS)
+--------------------------
+
+The feeds are stored on `Google Cloud Storage`_.
+
+ * URI scheme: ``gcs``
+ * Example URIs:
+
+   * ``gcs://mybucket/path/to/export.csv``
```
victor-torres: gsutil uses `gs://` as an URI prefix. See here: https://cloud.google.com/storage/docs/gsutil.

victor-torres: gsutil uses `gs://` as an URI prefix. See here: https://cloud.google.com/storage/docs/gsutil.

scrapy/extensions/feedexport.py
```diff
@@ -146,6 +146,32 @@ def _store_in_thread(self, file):
...
+        return cls(
+            uri,
+            crawler.settings['GCS_PROJECT_ID'],
+            crawler.settings['FEED_STORAGE_GCS_ACL']
+        )
+
+    def _store_in_thread(self, file):
+        file.seek(0)
+        from google.cloud.storage import Client
+        client = Client(project=self.project_id)
```
victor-torres: You can also create the client using:

```python
json_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../google-cloud-service-account.json')
client = Client.from_service_account_json(json_path)
```

ejulio: Yup, this was in my original approach.
But, I kept compatibility with IMAGES_STORE for now.

docs/topics/feed-exports.rst
ejulio: That's a good one.
For media pipelines, it is also `gs` https://docs.scrapy.org/en/latest/topics/media-pipeline.html#id1
I'll change it :smile:

ejulio: @victor-torres , I updated the scheme!

victor-torres:
I'll take a better look tomorrow. Thanks for updating.

ejulio:
@dangra can you review this PR as well?

docs/topics/feed-exports.rst
```diff
@@ -185,6 +185,27 @@ passed through the following settings:
...
+
+   * ``gs://mybucket/path/to/export.csv``
+
+ * Required external libraries: `google-cloud-storage <https://cloud.google.com/storage/docs/reference/libraries#client-libraries-install-python>`_.
+
+For more information about authentication, please refer to `Google Cloud documentation <https://cloud.google.com/docs/authentication/production>`_.
+
+You can set a *Project ID* and *Access Control List (ACL)* through the following settings:
+
+ * :setting:`FEED_STORAGE_GCS_ACL`
```
Gallaecio: Since this setting is specific to Google Cloud storage, maybe the name should be simply `GCS_ACL`?

docs/topics/settings.rst
```diff
@@ -749,6 +749,14 @@ The Feed Temp dir allows you to set a custom folder to save crawler
 temporary files before uploading with :ref:`FTP feed storage <topics-feed-storage-ftp>` and
 :ref:`Amazon S3 <topics-feed-storage-s3>`.
 
+.. setting:: FEED_STORAGE_GCS_ACL
+
+FEED_STORAGE_GCS_ACL
+--------------------
+
+The Access Control List (ACL) used when storing items to :ref:`Google Cloud Storage <topics-feed-storage-gcs>`.
+For more information on how to set this value, please refer to `Google Cloud documentation <https://cloud.google.com/storage/docs/access-control/lists>`_.
```
Gallaecio: I had a quick look at the linked documentation, and when using [predefined ACLs](https://cloud.google.com/storage/docs/access-control/lists#predefined-acl), I would not be sure whether the value defined here would have to be the one from the `JSON API` column, the one from the `XML API/gsutil`, or either. What would you think about clarifying it here?

docs/topics/feed-exports.rst
Gallaecio: Oh, I see this goes in line with #3607

ejulio: Yup.
Also, this aligns to other ACLs like FILE_STORAGE_ACL and so on.
My opinion is that we could refactor them, but to keep this PR shorter I prefer a different setting.

docs/topics/settings.rst
ejulio: Just updated to mention the JSON API

docs/topics/feed-exports.rst
Gallaecio: I’ve [left a comment about this](https://github.com/scrapy/scrapy/pull/3607#discussion_r268109665) in #3607.

I would wait for that discussion to be over before merging these changes. I believe this pull request is ready to merge otherwise, though.

Gallaecio: Nevermind.

scrapy/settings/default_settings.py
```diff
@@ -158,6 +159,7 @@
 }
 FEED_EXPORT_INDENT = 0
 
+FEED_STORAGE_GCS_ACL = None
```
kmike: I still wonder if that's ok to use None values as defaults, instead of empty strings, because there is no way to set an option to None using `-s` from command line (or am I wrong here?). So, once overridden, you can't get it back to the default value, when starting a spider. That's the reason most other default values are empty strings instead of None - we should probably treat empty string as None in most cases. By having it `''` in default_settings, and checking that it works, we make sure that `-s FEED_STORAGE_GCS_ACL=` from command line also works.

alepuccetti:
Any update on this feature?

PabloReyes:
Great work on this PR. Is there any update on it?

ejulio:
@kmike @Gallaecio , should I fix the conflicts to merge it?

Gallaecio:
Solved. I don’t think it was blocking review, though. There’s just lots of stuff to review :sweat_smile:

PabloReyes:
Thanks @Gallaecio and @ejulio ! There seem to be some issues now with the `tests.mock` now. I've seen in other places you're using `unittest.mock` and that seems to work OK. Don't know if they are two completely different libraries with different purposes though.

Gallaecio:
It looks like since https://github.com/scrapy/scrapy/commit/92ffd2f249ec276436277385e525397a2fb8b0a0#diff-498cf53d35427897613cdfc4b76fc6ea mock should be imported from `unittest` directly on each test file, instead of importing it from `tests/__init__.py`.

ejulio:
I can work on the fixes, but only if it moves forward to be merged.
As it's been flagged for merging for 6+ months...

Gallaecio:
That’s why I did not want to bother you, there’s no guarantee it will get merged any sooner :slightly_frowning_face:

elacuesta:
Hi @ejulio, this is awesome work :rocket:

I was trying this myself but I was getting the following exception:
```
  File "/.../scrapy/venv-scrapy/bin/scrapy", line 11, in <module>
    load_entry_point('Scrapy', 'console_scripts', 'scrapy')()
  File "/.../scrapy/scrapy/cmdline.py", line 145, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/.../scrapy/scrapy/cmdline.py", line 99, in _run_print_help
    func(*a, **kw)
  File "/.../scrapy/scrapy/cmdline.py", line 153, in _run_command
    cmd.run(args, opts)
  File "/.../scrapy/scrapy/commands/runspider.py", line 87, in run
    self.crawler_process.crawl(spidercls, **opts.spargs)
  File "/.../scrapy/scrapy/crawler.py", line 176, in crawl
    crawler = self.create_crawler(crawler_or_spidercls)
  File "/.../scrapy/scrapy/crawler.py", line 209, in create_crawler
    return self._create_crawler(crawler_or_spidercls)
  File "/.../scrapy/scrapy/crawler.py", line 214, in _create_crawler
    return Crawler(spidercls, self.settings)
  File "/.../scrapy/scrapy/crawler.py", line 64, in __init__
    self.extensions = ExtensionManager.from_crawler(self)
  File "/.../scrapy/scrapy/middleware.py", line 53, in from_crawler
    return cls.from_settings(crawler.settings, crawler)
  File "/.../scrapy/scrapy/middleware.py", line 35, in from_settings
    mw = create_instance(mwcls, settings, crawler)
  File "/.../scrapy/scrapy/utils/misc.py", line 146, in create_instance
    return objcls.from_crawler(crawler, *args, **kwargs)
  File "/.../scrapy/scrapy/extensions/feedexport.py", line 242, in from_crawler
    o = cls(crawler.settings)
  File "/.../scrapy/scrapy/extensions/feedexport.py", line 227, in __init__
    if not self._storage_supported(self.urifmt):
  File "/.../scrapy/scrapy/extensions/feedexport.py", line 309, in _storage_supported
    self._get_storage(uri)
  File "/.../scrapy/scrapy/extensions/feedexport.py", line 328, in _get_storage
    return self._get_instance(self.storages[urlparse(uri).scheme], uri)
  File "/.../scrapy/scrapy/extensions/feedexport.py", line 322, in _get_instance
    *args, **kwargs)
  File "/.../scrapy/scrapy/utils/misc.py", line 150, in create_instance
    return objcls(*args, **kwargs)
TypeError: __init__() missing 2 required positional arguments: 'project_id' and 'acl'
```

I took the liberty of merging `master` into this branch to solve the conflict in `tests/test_feedexport.py`, and now I'm able to export files :ok_hand:.

Besides that, the new tests are not being executed in Travis, since `google-cloud-storage` is not installed. Could you update that? Thanks!

wRAR:
@ejulio I'm ready to merge this after you add `google-cloud-storage` to `tests/requirements-py3.txt` (I tried to push to your branch but it looks like github doesn't allow me)

elacuesta:
I added `google-cloud-storage` to the "pinned" tox environment, and [all tests pass](https://travis-ci.org/github/scrapy/scrapy/builds/700256089). However, the docs build [fails](https://travis-ci.org/github/scrapy/scrapy/jobs/700256093#L352) because the `GCS_PROJECT_ID` setting is already defined in `docs/topics/media-pipeline.rst`.

wRAR:
Should we merge this? I think everything is fine now.

PR #3609, opened 2019-02-01
Document FilesPipeline.file_path and ImagesPipeline.file_path

### LINKED ISSUE #2253: Document how to customize path for written files with media pipelines
Motivation: https://github.com/scrapy/scrapy/issues/923

Gallaecio:
Fixes #2253

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -392,6 +392,36 @@ See here the methods that you can override in your custom Files Pipeline:
 
 .. class:: FilesPipeline
 
+   .. method:: file_path(request, response, info)
+
+      This method is called once per downloaded item. It returns the
+      download path of the file originating from the specified
+      :class:`response <scrapy.http.Response>`.
+
+      In addition to ``response``, this method receives the original
+      :class:`request <scrapy.Request>` and
+      :class:`info <scrapy.pipelines.media.MediaPipeline.SpiderInfo>`.
+
+      You can override this method to customize the download path of each file.
+
+      For example, if file URLs end like regular paths (e.g.
+      ``https://example.com/a/b/c/foo.png``), you can use the following
+      approach to download all files into the ``files`` folder with their
+      original filenames (e.g. ``files/foo.png``)::
+
+        import os
+        from urllib.parse import urlparse
+
+        from scrapy.pipelines.files import FilesPipeline
+
+        class MyFilesPipeline(FilesPipeline):
+
+            def file_path(self, request, response, info):
+                return 'files/' + os.path.basename(urlparse(request.url).path)
+
+      By default the :meth:`file_path` method returns
+      ``full/<request URL hash>.<extension>``.
+
    .. method:: FilesPipeline.get_media_requests(item, info)
 
       As seen on the workflow, the pipeline will get the URLs of the images to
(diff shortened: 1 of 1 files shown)


docs/topics/media-pipeline.rst
```diff
@@ -392,6 +392,30 @@ See here the methods that you can override in your custom Files Pipeline:
...
+      :class:`response <scrapy.http.Response>`.
+
+      In addition to ``response``, this method receives the original
+      :class:`request <scrapy.Request>` and
+      :class:`info <scrapy.pipelines.media.MediaPipeline.SpiderInfo>`.
+
+      You can override this method to customize the download path of each file.
+
+      For example, to download all files into the ``files`` folder with their
+      original filenames::
```
kmike: I think we should be a bit more clear on that it is just an example, which works for a certain kind of URLs - `http://example.com/foo.png` would be stored properly, but e.g. all files like `http://example.com/file?name=foo.png` would be stored as "file", overwriting each other. Possible solutions:

* say that it is an example, which works on websites where file URLs follow certain rules
* make it more robust (but less readable) by adding full URL hash to the path, in addition to guessed file name.

A full implementation should probably also check Content-Disposition header, to get a file name.

docs/topics/media-pipeline.rst
```diff
@@ -475,6 +499,30 @@ See here the methods that you can override in your custom Images Pipeline:
...
+
+        import os
+        from urllib.parse import urlparse
+
+         def file_path(self, request, response, info):
```
schjoq: Hi. I found this definition wouldn't work, without assigning default values to the keyword arguments ``response`` and ``info``:
```suggestion
         def file_path(self, request, response=None, info=None):
```
Just like in:
https://github.com/scrapy/scrapy/blob/b364bfb68b481a696141aa6b7b3312679670c38e/scrapy/pipelines/files.py#L460

Gallaecio: I’ll look into it in detail later, but if I recall right those two fields are only needed when using some deprecated methods.

Gallaecio: Verified, that is only needed to support subclasses of these pipelines that still use the deprecated `file_key` method.

docs/topics/media-pipeline.rst
Gallaecio: I’ve gone with option 1, and also tried to make it obvious that the example implementation does not take subdirectories into account either.

docs/topics/media-pipeline.rst
```diff
@@ -392,6 +392,32 @@ See here the methods that you can override in your custom Files Pipeline:
...
+      For example, if file URLs end like regular paths (e.g.
+      ``https://example.com/a/b/c/foo.png``), you can use the following
+      approach to download all files into the ``files`` folder with their
+      original filenames (e.g. ``files/foo.png``)::
+
+        import os
+        from urllib.parse import urlparse
+
+         def file_path(self, request, response, info):
+             return 'files/' + os.path.basename(urlparse(request.url).path)
```
kmike: to make it a bit more clear, could you please add some boilerplate, e.g. 

```py
# ...
class MyFilesPipeline(FilesPipeline):
    # ...
    def file_path(self, request, response, info):
```

Gallaecio: I’ve gone with the whole implementation, as it only required a few more lines.

kmike:

PR #3610, opened 2019-02-01
[MRG+1] Check that spidercls arguments in scrapy.crawler classes are not spid…

### LINKED ISSUE #2283: Add note that CrawlerProcess.crawl() needs a crawler or spider class, not a spider instance
Even if [it's in the docs](http://doc.scrapy.org/en/latest/topics/api.html#scrapy.crawler.CrawlerProcess.crawl), 

> crawler_or_spidercls (Crawler instance, Spider subclass or string) – already created crawler, or a spider class or spider’s name inside the project to create it

users still miss sometimes that `.crawl()` expects a Spider class (or a crawler) but not a Spider instance.

e.g. http://stackoverflow.com/q/39639568

> I overlooked the fact that the class itself has to be passed as an argument, and not an instance of that class.
> This is quite subtle on the eye but makes all the difference

This ticket is to make it a bit more prominent.

Gallaecio:
…er objects

I think this is a better approach to fix #2283. Instead of trying to make the documentation more obvious, have the code fail gracefully with a useful error message.

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -7,6 +7,7 @@ import sys
 from twisted.internet import reactor, defer
 from zope.interface.verify import verifyClass, DoesNotImplement
 
+from scrapy import Spider
 from scrapy.core.engine import ExecutionEngine
 from scrapy.resolver import CachingThreadedResolver
 from scrapy.interfaces import ISpiderLoader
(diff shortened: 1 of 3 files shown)


Gallaecio:
I’ll review the failing text next week, it probably needs updated expectations.

Gallaecio:
Oh, right, I should cover the checks with automated tests.

kmike:
Perfect, thanks @Gallaecio!

PR #3532, opened 2018-12-13
[WIP] Create an API reference section generated from the source code

Gallaecio:
- [x] Split Core API into separate pages.
- [x] Provide separate pages for the parts of the API that were not covered in the Core API topic.
- [x] Use the simplest import path for members that have aliases.
- [x] Move documentation from rst files to py files.
  Includes searching for `\.\. (?:auto)?(?:module|function|data|exception|class|attribute|(?:static|class)?method|decorator(?:method)?)::` outside the API documentation.
- [x] Search for `…` and `TODO` in the documentacion (used to mark places where additional or rewritten documentation is expected).
- [x] Add links to the sources.
- [x] Normalize the indentation of `:members:` and similar.
- ~~Run `pylint` or similar to correct line length issues caused by moving documentation to the sources.~~ (skipped since #2144 is open)
- [x] Use ~ for API references where it makes sense
- [x] Include all API member paths (e.g. both `scrapy.Request` and `scrapy.http.Request`) heading the corresponding documentation.
- [x] Fix broken API references.
- [x] Make use of the `currentmodule` directive in topics to shorten API references (first add currentmodule where it makes sense, then review all API references un pages or sections using the directive, and finally make sure there are no broken API references).
- [x] Review the documentation of modules included with `automodule`.
- [x] Switch references to the Python documentation by URL to actual references.
- [x] Switch references to the Parsel documentation by URL to actual references. (depends on https://github.com/scrapy/parsel/pull/135)
- [x] Fix doctests with Python 3, and ensure they are not executed with Python 2. (depends on #3601)
- [x] Switch references to the Twisted documentation by URL to actual references.
- [ ] Switch references to documentations hosted on readthedocs.io (which means they use Sphinx) by URL to actual references.
- [ ] Include URL references in the same file where they are used (source file instead of documentation page that uses autodoc)
- [ ] Move the documentation of settings that belong to a specific Scrapy component (e.g. extension, downloader middleware, spider middleware)
- [ ] Review the occurrences of “alias of” in the generated documentation, try to avoid them.
- [ ] Try to achieve an effects similar to that of multiple signatures (https://docs.python.org/3/library/stdtypes.html#dict) when describing multiple import paths of the same class.
- [ ] Manually review the whole non-API documentation to find writing issues caused by the API documentation (notes about there not being API documentation, "below" references on topics).
[...]

PR #3612, opened 2019-02-01
[WIP] Use docstrings where possible

### LINKED ISSUE #713: Bring docstrings back

Gallaecio:
Fixes #713 (and #2727). This is the first part of #3532.

I’ve gone through all matches of the following regular expression:

    \.\.\s+(?:function|data|exception|class|attribute|(?:static|class)?method|decorator(?:method)?)::

It should match all scenarios where `autodoc` can be used but it is not currently used.

I kept `module` outside of the regular expression on purpose, because none of the current usages of `.. module::` is designed in a way that it makes sense to extract the documentation from a docstring. There seem to be there just so that following a module link from the index leads somewhere.

After these changes, only the following 16 matches for the regular expression above remain while searching with a simple `*.rst` filename filter. Of those:

-   There is a match, `Item.fields`, where a [Sphinx bug](https://github.com/sphinx-doc/sphinx/issues/741) prevents removing the documentation from the reStructuredText file.

-   All other matches are due to fictional API members, members that are not really part of the Scrapy API but that users may implement in subclasses.

Regarding the latter, I believe the best approach to move that documentation to docstrings would be to implement the corresponding (zope) interfaces (à la [`scrapy.interfaces.ISpiderLoader`](https://github.com/scrapy/scrapy/blob/master/scrapy/interfaces.py)) for the following affected fictional members: [`scrapy.downloadermiddlewares.DownloaderMiddleware`](https://doc.scrapy.org/en/latest/topics/downloader-middleware.html#scrapy.downloadermiddlewares.DownloaderMiddleware), [pipeline](https://doc.scrapy.org/en/latest/topics/item-pipeline.html#writing-your-own-item-pipeline), [`scrapy.spidermiddlewares.SpiderMiddleware`](https://doc.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware). I see no solution at the moment for the remaining fictional members, two methods of [`scrapy.contracts.Contract`](https://doc.scrapy.org/en/latest/topics/contracts.html#scrapy.contracts.Contract).

If you agree with the interface definition approach for those remaining API members, please let me know whether I should create a separate pull requests or include those changes in this pull requests (squashed or as a separate commit).


Gallaecio:
Some of the failed checks are due to doctests that used to be in `.rst` files on the `docs` folder, hence ignored, but are now in `.py` files on the `scrapy` folder, hence processed.

Of those, I don’t know how I should fix the ones that fail due to the output being different on Python 2 and Python 3. For example:

    https://travis-ci.org/scrapy/scrapy/jobs/487505259#L2191

How should I proceed to fix those scenarios? Should the output be for Python 2 or for Python 3? Is it OK to skip those doctests in the other versions of Python?

Gallaecio:
I went with using Python 3 in the documentation and skipping those tests in environments with Python 2.

Gallaecio:
Tests passed (by skipping doctests in Python < 3). I’m ignoring the coverage report, as the changes only affect documentation files, comments and configuration files.

kmike:
Hey @Gallaecio! I know it is a huge amount of work you've done here; my worry is that it'd be hard to merge such big pull request - there is a lot of changes to review, and master is changing as well, adding merge conflicts. 

* Getting stuck in discussing even a single change may delay the merge of the whole PR, and increase work on resolving merge conflicts. 
* It is also hard to get back to a large pull request and figure out for which parts the review is finished, and for which there is still discussion ongoing.
* Reviewing big pull request requires a large time investement once; it is easier to find time here and there to review smaller pull requests.

Do you think it'd be possible to split it into several pull requests, e.g. a PR with some groundwork (adding extensions, CI changes), and then separate pull requests for different parts of Scrapy? This way it should be much easier to eventually merge all changes, one by one.

Gallaecio:
Yes, I think that will be the only sane approach. I’ll get to it next week.

PR #3617, opened 2019-02-07
[WIP] adding the process_spider_failure method to spider middlewares

eLRuLL:
I think it is a good idea to return the `failure` directly instead of the `failure.value` to the `process_spider_failure` method of a spider middleware (deprecating the `process_spider_exception` if possible in newer releases).

The main idea behind this is to give the spider middleware the ability to perform actions on the `failure` object instead of just the `exception`, as we lose information because of just passing the `exception`, mostly "traceback" information that could be relevant for some cases.

Basically it could give the ability to spider middlewares to do something like this:

    def process_spider_failure(self, response, failure, spider):
        spider.logger.error(failure.getTraceback())

which is not possible right now as `process_spider_exception` only gets the exception instance.

The idea behind this PR is to give `process_spider_failure` precedence over `process_spider_exception` and will only check one of them if both specified.


eLRuLL:
please let me know if you guys think it could be a good idea, so I can include documentation. Thanks

PR #3620, opened 2019-02-12
Improve the documentation of dont_merge_cookies

### LINKED ISSUE #2998: dont_merge_cookies docs are incomplete
dont_merge_cookies [docs](https://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request) are incomplete: they say that 

> When some site returns cookies (in a response) those are stored in the cookies for that domain and will be sent again in future requests. That’s the typical behaviour of any regular web browser. However, if, for some reason, you want to avoid merging with existing cookies you can instruct Scrapy to do so by setting the dont_merge_cookies key to True in the Request.meta.

But this flag not only prevents merging of cookies, but also prevents sending of them: https://github.com/scrapy/scrapy/blob/b8870ee8a10360aaa74298324d97c823b88ec5c6/scrapy/downloadermiddlewares/cookies.py#L27

Maybe a separate issue, but it'd be nice to have separate flags for sending and merging of cookies.

//cc @whalebot-helmsman

Gallaecio:
Clarify that it supresses cookie handling both ways, on the request and on the response.

Fixes #2998

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -83,13 +83,21 @@ Request objects
         .. reqmeta:: dont_merge_cookies
 
         When some site returns cookies (in a response) those are stored in the
-        cookies for that domain and will be sent again in future requests. That's
-        the typical behaviour of any regular web browser. However, if, for some
-        reason, you want to avoid merging with existing cookies you can instruct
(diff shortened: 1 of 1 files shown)


Gallaecio:
Oh, my. I created [the same pull request](#4028) twice :man_facepalming:

PR #3621, opened 2019-02-12
Document that the crawl order is BFO for small numbers of start requests

### LINKED ISSUE #1739: DFO vs BFO in Scrapy FAQ
@vincent-ferotin sent a message to scrapy-users a while ago (https://groups.google.com/forum/#!searchin/scrapy-users/Vincent$20F%C3%A9rotin/scrapy-users/n56O2sCAbp0/IbN8XGusAgAJ) and created a repo (https://github.com/vincent-ferotin/scraping-github) to demonstrate an issue: http://doc.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order says scrapy crawls DFO by default, but in practice this is not observed. 

It looks like the tricky part is Downloader. By default Scrapy processes 16 requests in parallel. It means that Downloader asks Scheduler for 16 requests; after Downloader get them they are no longer handled by Scheduler. Downloader processes these requests in no particular order if there is enough concurrency per domain (8 by defualt), or uses a FIFO queue if concurrency is not enough. FIFO means BFO crawl, so for short request queues order is BFO regardless of Scheduler settings.

I think we should at least clarify that in docs.

See also: https://github.com/scrapy/scrapy/pull/1727, https://github.com/scrapy/scrapy/pull/1440, https://github.com/scrapy/scrapy/issues/1371.

Gallaecio:
Fixes #1739

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -129,12 +129,22 @@ Does Scrapy crawl in breadth-first or depth-first order?
 
 By default, Scrapy uses a `LIFO`_ queue for storing pending requests, which
 basically means that it crawls in `DFO order`_. This order is more convenient
-in most cases. If you do want to crawl in true `BFO order`_, you can do it by
+in most cases.
+
+If you do want to crawl in true `BFO order`_, you can do it by
 setting the following settings::
 
     DEPTH_PRIORITY = 1
     SCHEDULER_DISK_QUEUE = 'scrapy.squeues.PickleFifoDiskQueue'
     SCHEDULER_MEMORY_QUEUE = 'scrapy.squeues.FifoMemoryQueue'
+
+While pending requests are below the configured values of
+:setting:`CONCURRENT_REQUESTS`, :setting:`CONCURRENT_REQUESTS_PER_DOMAIN` or
+:setting:`CONCURRENT_REQUESTS_PER_DOMAIN`, those requests are sent
+concurrently. As a result, the first few requests of a crawl rarely follow the
+desired order. Lowering those settings to ``1`` enforces the desired order, but
+it significantly slows down the crawl as a whole.
+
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -136,6 +136,9 @@ setting the following settings::
     SCHEDULER_DISK_QUEUE = 'scrapy.squeues.PickleFifoDiskQueue'
     SCHEDULER_MEMORY_QUEUE = 'scrapy.squeues.FifoMemoryQueue'
```
ejulio: What do you think of
```
By default, Scrapy uses a LIFO queue for storing pending requests, which basically means that it crawls in DFO order. This order is more convenient in most cases. However, when the number of start request is small it always crawls in BFO order <https://github.com/scrapy/scrapy/issues/1739>`_. [...]
```
?

Gallaecio: Much better, thanks for the feedback!

docs/faq.rst
```diff
@@ -129,7 +129,12 @@ Does Scrapy crawl in breadth-first or depth-first order?
 
 By default, Scrapy uses a `LIFO`_ queue for storing pending requests, which
 basically means that it crawls in `DFO order`_. This order is more convenient
-in most cases. If you do want to crawl in true `BFO order`_, you can do it by
+in most cases.
+
+However, when the number of start request is small it always `crawls in BFO
```
kmike: I don't think it is correct: if there is a single request in start_requests, but then spider schedules many requests (e.g. after extracting links), behavior would be as usual, maybe besides first few requests.

Gallaecio:
I’ve switched to a completely new wording that I hope is more accurate and clear.

kmike:
Thanks @Gallaecio, I like the clarity of the explanation!

PR #3625, opened 2019-02-13
[MRG+1] A different S3 Endpoint URL is now possible when uploading images

Ordepsousa:
AWS_ENDPOINT_URL in Settings was read by files.py but not by images.py. I don't know if it was intentionally but i couldn't connect to my local minio S3 Server when uploading images with ImagesPipeline


kmike:
A great catch, thanks @Ordepsousa! 

I wonder if you can add other missing AWS options here as well, as you're on it - it looks like AWS_REGION_NAME, AWS_USE_SSL and AWS_VERIFY are also missing.

By the way, the existing code doesn't look good at all: it mutates a global object, so one can't start two spiders via CrawlerRunner and have different AWS options for them. Also, the code is duplicated between files and images pipelines. But that're separate issues.

Ordepsousa:
No problem @kmike :)

Added the missing AWS settings as you asked for
Yeah, it doesn't look so good, probably will need some cleaning in the future

PR #3626, opened 2019-02-13
[MRG+1] Document that the main entry point of downloader and spider middlewar…

### LINKED ISSUE #1789: Document from_crawler for Scrapy middlewares

Gallaecio:
…es is from_crawler()

Fixes #1789


kmike:
Looks great, +1 to merge. Thanks @Gallaecio.

PR #3627, opened 2019-02-14
[MRG+1] Fix #2933: Add URI params to FILES_STORE

### LINKED ISSUE #2933: Overwrite settings in image pipeline?
Hello,
from the documentation it is not clear to me how to change IMAGES_STORE path inside a pipeline.
I am scraping two different images which shall be stored in different paths.
So my idea was to write for each image an extra pipeline and overwrite the images_store path.
Is that even possible? Or is there another method to change the path depending on which image i am downloading?

ejulio:
Fix #2933 
The idea here is to make `FILES_STORE` URI similar to `FEED_URI` by enabling `%()s` format with spider attributes.

Why?
We worked on a project that required a custom `FEED_URI` for each spider because it was a SFTP connection and each spider had it's own credentials to access this storage.
With this feature, we could write something like

```
# settings.py
FILES_STORE = 'sftp://%(username)s:%(password)s:80/%(name)s/'

# spider.py

class MySpider(Spider):
    name = 'my_spyder'
    username = 'my_user'
    password = 'my_secure_password'
```

I don't know if we should rename `FILES_STORE` to `FILES_URI` to make it similar to `FEED_URI`. Thoughts here?

TODOs:
* [x] Add documentation


scrapy/pipelines/files.py
```diff
@@ -308,7 +303,8 @@ def __init__(self, store_uri, download_func=None, settings=None):
 
     @classmethod
     def from_settings(cls, settings):
-        s3store = cls.STORE_SCHEMES['s3']
+        cls._load_storages(settings)
+        s3store = cls.STORAGES['s3']
```
ejulio: I thought about refactoring this piece as well.
Currently we get class and change it's attributes.
I think that settings the proper attributes would a responsibility of the given `S3`, `GCS`, `FS` class.
So, we could rewrite to instantiate the class giving the URI and the settings.

kmike: +1, the current code doesn't look good, I also noticed it recently (https://github.com/scrapy/scrapy/pull/3625#issuecomment-463333661). It'd be good this in a separate PR, if possible.

ejulio: thanks @kmike, I can work on that after this PR to avoid merge conflicts.

scrapy/settings/default_settings.py
```diff
@@ -160,6 +160,13 @@
 
 FILES_STORE_S3_ACL = 'private'
 FILES_STORE_GCS_ACL = ''
+FILES_STORAGES = {}
```
ejulio: I don't know if `FILES_STORAGES` and `FILES_STORAGES_BASE` are good names.

Gallaecio: They seem coherent with similar settings, so :+1: from me.

Gallaecio (approved):
Looks good to me.

scrapy/utils/misc.py
```diff
@@ -142,3 +142,7 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
         return objcls.from_settings(settings, *args, **kwargs)
     else:
         return objcls(*args, **kwargs)
+
+
+def get_object_attributes_as_dict(obj):
+    return { k: getattr(obj, k) for k in dir(obj) }
```
kmike:
```suggestion
    return {k: getattr(obj, k) for k in dir(obj)}
```

scrapy/utils/misc.py
```diff
@@ -142,3 +142,7 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
         return objcls.from_settings(settings, *args, **kwargs)
     else:
         return objcls(*args, **kwargs)
+
+
+def get_object_attributes_as_dict(obj):
```
kmike: could you please add a small docstring?

scrapy/pipelines/files.py
```diff
@@ -269,24 +269,19 @@ class FilesPipeline(MediaPipeline):
...
-        if not store_uri:
+    def __init__(self, urifmt, download_func=None, settings=None):
+        if not urifmt:
             raise NotConfigured
 
         if isinstance(settings, dict) or settings is None:
             settings = Settings(settings)
 
         cls_name = "FilesPipeline"
-        self.store = self._get_store(store_uri)
```
kmike: This is almost backwards compatible, because self.store is set in open_spider method. 
Do you need to use open_spider because crawler.spider is not set when FilesPipeline.from_crawler is called? (haven't checked it)

scrapy/pipelines/files.py
```diff
@@ -269,24 +269,19 @@ class FilesPipeline(MediaPipeline):
 
     MEDIA_NAME = "file"
     EXPIRES = 90
-    STORE_SCHEMES = {
```
kmike: I wonder if that's possible to have backwards compatibility layer for this; currently the change breaks any pipeline subclasses which override STORE_SCHEMES.

ejulio:
@kmike , added compatibility for `store` and `STORE_SCHEMES`.
Not sure if I should add a specific test for the warning.

Gallaecio:
@ejulio There are some conflicts, mainly due to https://github.com/scrapy/scrapy/pull/3961

PR #3629, opened 2019-02-16
[MRG+1] rel_has_nofollow: remove redundant if statement

johndela1:
no need for if because expression already returns bool


kmike:
Thanks for the cleanup @johndela1, and thanks @ejulio for checking it.

PR #3630, opened 2019-02-17
Drop few words about getall selector being called

mouchh:
To avoid people opening scrapy source code to make sure getall is called over get

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -31,7 +31,8 @@ using the Item class specified in the :attr:`ItemLoader.default_item_class`
 attribute.
 
 Then, you start collecting values into the Item Loader, typically using
-:ref:`Selectors <topics-selectors>`. You can add more than one value to
+:ref:`Selectors <topics-selectors>`. To note, the Item Loader will extract data
+by using the selector ``.getall()`` method. You can add more than one value to
 the same item field; the Item Loader will know how to "join" those values later
 using a proper processing function.
 
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -65,7 +65,11 @@ data that will be assigned to the ``name`` field later.
 Afterwards, similar calls are used for ``price`` and ``stock`` fields
 (the latter using a CSS selector with the :meth:`~ItemLoader.add_css` method),
 and finally the ``last_update`` field is populated directly with a literal value
-(``today``) using a different method: :meth:`~ItemLoader.add_value`.
+(``today``) using a different method: :meth:`~ItemLoader.add_value`. To note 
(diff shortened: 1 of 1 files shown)


docs/topics/loaders.rst
```diff
@@ -31,7 +31,8 @@ using the Item class specified in the :attr:`ItemLoader.default_item_class`
 attribute.
 
 Then, you start collecting values into the Item Loader, typically using
-:ref:`Selectors <topics-selectors>`. You can add more than one value to
+:ref:`Selectors <topics-selectors>`. To note, the Item Loader will extract data
+by using the selector ``.getall()`` method. You can add more than one value to
```
kmike: While this is technically correct, I'm not sure it is a right place to document this, because it doesn't look related to the following example. 

Could you please elaborate, what was you motivation to add this note?

mouchh: Hi kmike, I was actually writting my first Item Loader and tried to follow the given example but it was not precise enough to let me know in what format my data will be sent to my input processor (string? list?). As this can have a huge impact depending on the potential input processor used (thinking about Compose or MapCompose), I felt like this information was missing in the doc.

That being said; you're right in a way that I did not choose the best place to include this note. Maybe we can include it like this 

> Afterwards, similar calls are used for price and stock fields (the latter using a CSS selector with the add_css() method), and finally the last_update field is populated directly with a literal value (today) using a different method: add_value(). To note add_css() and add_xpath() methods will always return a list as they are actually calling selector getall() method over get() which may return a string value. Thus, input processors should be implemented by waiting for a list instead of a single string. 

instead of just

> Afterwards, similar calls are used for price and stock fields (the latter using a CSS selector with the add_css() method), and finally the last_update field is populated directly with a literal value (today) using a different method: add_value().

What do you think?

docs/topics/loaders.rst
```diff
@@ -64,7 +65,11 @@ data that will be assigned to the ``name`` field later.
 Afterwards, similar calls are used for ``price`` and ``stock`` fields
 (the latter using a CSS selector with the :meth:`~ItemLoader.add_css` method),
 and finally the ``last_update`` field is populated directly with a literal value
-(``today``) using a different method: :meth:`~ItemLoader.add_value`.
+(``today``) using a different method: :meth:`~ItemLoader.add_value`. To note 
+:meth:`~ItemLoader.add_css` and :meth:`~ItemLoader.add_xpath` methods will always
+return a list as they are actually calling selector ``getall()`` method over
```
kmike: add_css and add_xpath methods don't return anything - I see what you mean, but it seems a different wording is needed.

malberts: How about: 

> ... methods will always store a list in the resulting item's field as they are actually calling selector ``getall()`` method over ...

AdrianAtZyte:
#4036

PR #3636, opened 2019-02-22
[ON HOLD] Provide a single page that covers all backward-incompatible changes and deprecations

### LINKED ISSUE #1385: defect of documentions
Hi,

Today I was trying to upgrade from 0.24.5 to 1.0.1. But I didn't find any document about how to upgrade from an older version to new one. So I upgraded Scrapy by running the codes.
I think it is not a good idea to upgrade like this. Maybe I missed some parts on testing and upgrading.
So I think it is a good practice to add a document about "How to upgrade from X to XX?"

Gallaecio:
Fixes #1385

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -3,9 +3,9 @@
 Deprecations
 ============
 
-This document outlines the Scrapy deprecation policy, and lists when various
-pieces of Scrapy have been removed or altered in a backward incompatible way,
-following their deprecation.
+This document outlines the Scrapy deprecation policy, how to handle deprecation
+warnings, and lists when various pieces of Scrapy have been removed or altered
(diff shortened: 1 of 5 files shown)


Gallaecio:
I think this already got to a point where it may be reviewed.

It covers only down to 1.0.0, because for lower versions there does not seem to be a great coverage of backward-incompatible changes in `news.rst`. Should I cover all the way down to 0.7 anyway?

I have a lot of mixed feelings with the result, so please do not hesitate to suggest changes. Should related commits or pull requests be referenced? Should the structure of the page change? Should changes be mentioned in the timeline both when they are deprecated and when the deprecation is enforced?

Gallaecio:
I’m delaying updates until there’s feedback, as updating this can be time consuming.

PR #3638, opened 2019-02-24
[MRG+1] Show elapsed time in statscollector

maramsumanth:
Showing elapsed time in seconds in stats, also changing from utc to local time for better user experience.
(In my opinion time&date should be displayed based on the local time)

### DIFF
--- a/scrapy/extensions/corestats.py
+++ b/scrapy/extensions/corestats.py
@@ -27,8 +27,8 @@ class CoreStats(object):
         finish_time = datetime.datetime.utcnow()
         elapsed_time = finish_time - self.stats.get_value('start_time')
         elapsed_time_seconds = elapsed_time.seconds + elapsed_time.microseconds/1000000
-        self.stats.set_value('finish_time', finish_time, spider=spider)
         self.stats.set_value('elapsed_time_seconds', elapsed_time_seconds, spider=spider)
+        self.stats.set_value('finish_time', finish_time, spider=spider)
         self.stats.set_value('finish_reason', reason, spider=spider)
 
     def item_scraped(self, item, spider):
--- a/scrapy/extensions/corestats.py
+++ b/scrapy/extensions/corestats.py
@@ -6,7 +6,8 @@ import datetime
 from scrapy import signals
 
 class CoreStats(object):
-
+    start_time = 0
+    finish_time = 0
     def __init__(self, stats):
         self.stats = stats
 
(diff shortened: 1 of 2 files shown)


maramsumanth:
@lopuhin , can you look into this?

scrapy/extensions/corestats.py
```diff
@@ -21,10 +22,16 @@ def from_crawler(cls, crawler):
...
     def spider_opened(self, spider):
-        self.stats.set_value('start_time', datetime.datetime.utcnow(), spider=spider)
+        start_time = datetime.datetime.now()
+        self.start_time = start_time
+        self.stats.set_value('start_time', start_time, spider=spider)
 
     def spider_closed(self, spider, reason):
-        self.stats.set_value('finish_time', datetime.datetime.utcnow(), spider=spider)
+        finish_time = datetime.datetime.now()
+        elapsed_time = finish_time - self.start_time
```
lopuhin: Maybe fetching ``start_time`` from stats would be more robust? That would allow to remove start_time and finish_time variables.

scrapy/extensions/corestats.py
```diff
@@ -21,10 +22,16 @@ def from_crawler(cls, crawler):
         return o
 
     def spider_opened(self, spider):
-        self.stats.set_value('start_time', datetime.datetime.utcnow(), spider=spider)
```
lopuhin: I don't think moving from UTC time to a local time is a good idea, due to the following:
- this is a backwards incompatible change, we can't just change this
- time zone on the server might be different from the user time zone anyway, so using UTC seems less confusing to me - this is just my opinion though

maramsumanth: @lopuhin Yeah I agree with you and I have made the necessary changes! :)

lopuhin: @maramsumanth code looks nice, thanks! 👍

maramsumanth: @lopuhin , Is it necessary to add tests?

maramsumanth:
@kmike , can you please look at this small feature. Thanks :)

maramsumanth:
Hey @kmike @whalebot-helmsman , what do you think of adding this feature?

scrapy/extensions/corestats.py
```diff
@@ -24,7 +24,11 @@ def spider_opened(self, spider):
         self.stats.set_value('start_time', datetime.datetime.utcnow(), spider=spider)
 
     def spider_closed(self, spider, reason):
-        self.stats.set_value('finish_time', datetime.datetime.utcnow(), spider=spider)
+        finish_time = datetime.datetime.utcnow()
+        elapsed_time = finish_time - self.stats.get_value('start_time')
+        elapsed_time_seconds = elapsed_time.seconds + elapsed_time.microseconds/1000000
```
kmike: In Python 2 `elapsed_time.microseconds/1000000` is always 0

kmike:
Hey @maramsumanth! I think having `elapsed_time_seconds` in stats is a good feature. 

It'd be good to add a test for this feature (e.g. modify test_closespider_timeout - which is incorrect, by the way, as it adds seconds and microseconds)

scrapy/extensions/corestats.py
maramsumanth: I think this is failing for python 2.7, what should I modify then @kmike ?

maramsumanth:
@kmike, can you please review this. Should I change close_on , since it is failing for version 2.7

scrapy/extensions/corestats.py
Gallaecio: What about using [`timedelta.total_seconds()`](https://docs.python.org/2/library/datetime.html#datetime.timedelta.total_seconds)?

```suggestion
        elapsed_time_seconds = elapsed_time.total_seconds()
```

maramsumanth:
Thanks @Gallaecio :)

kmike:
Thanks @maramsumanth and @Gallaecio!

PR #3647, opened 2019-03-01
removing parameters from url in _get_sitemap body if the sitemap url has parameters

hiddenWarrior:
sometimes the the url have parameters so it's better to check for the path only in the spider to avoid this problem in the sitemap spider
example:
https://jewel-hub.com/sitemap_products_1.xml?from=9832560205&to=1980793258038


hiddenWarrior:
@Gallaecio what do you think

scrapy/spiders/sitemap.py
```diff
@@ -2,17 +2,17 @@
...
 import six
 
+from six.moves.urllib.parse import urlparse
+
 from scrapy.spiders import Spider
 from scrapy.http import Request, XmlResponse
 from scrapy.utils.sitemap import Sitemap, sitemap_urls_from_robots
 from scrapy.utils.gz import gunzip, gzip_magic_number
 
-
```
Gallaecio: Please, remove unnecessary changes like this line removal and the other one at line 15.

scrapy/spiders/sitemap.py
```diff
@@ -81,7 +81,9 @@ def _get_sitemap_body(self, response):
         # without actually being a .xml.gz file in the first place,
         # merely XML gzip-compressed on the fly,
         # in other word, here, we have plain XML
-        elif response.url.endswith('.xml') or response.url.endswith('.xml.gz'):
+        file_path = urlparse(response.url).path
+        # to avoid parameters in the url
```
Gallaecio: I think this comment is unnecessary, your code is quite clean and obvious.

Gallaecio (changes requested):
I’ve suggested a few changes. 

However, I do wonder if you should rethink your whole patch. I am trying to think about backward compatibility, and I am thinking that, at least in theory, some scenarios may be broken by your patch.

For example:

    https://example.com/static.php?file=sitemap.xml.gz

Now, I’ve never encountered such an URL, just as I’ve never encountered the type of sitemap URL you are supporting with your patch.

However, I am thinking that a simpler approach, `if re.search(r'\.xml(\.gz)?(\?|$)', response.url)`, would support the proposed scenario without removing support for any other scenario that is currently supported.

hiddenWarrior:
@Gallaecio sorry for being late but you have a point and i made your suggestions

scrapy/spiders/sitemap.py
```diff
@@ -81,7 +81,7 @@ def _get_sitemap_body(self, response):
         # without actually being a .xml.gz file in the first place,
         # merely XML gzip-compressed on the fly,
         # in other word, here, we have plain XML
-        elif response.url.endswith('.xml') or response.url.endswith('.xml.gz'):
+        if re.search(r'\.xml(\.gz)?(\?|$)', response.url):
```
Gallaecio: Turning the `elif` into an `if` makes sense in this case, but then the previous `elif` should be changed as well, I think, to keep the method consistent.

Gallaecio (approved):

PR #3648, opened 2019-03-01
backwards → backward (adj.)

Gallaecio:
Based on https://english.stackexchange.com/a/2449/87187

Applied to documentation, source comments and a test name.


kmike:
TIL. Thanks @Gallaecio.

PR #3649, opened 2019-03-01
[MRG+1] Switch from ` to `` where inline code formatting is desired

Gallaecio:


kmike:
Nice cleanup, thanks @Gallaecio!

PR #3654, opened 2019-03-05
Allow exit code propagation in case of CLOSESPIDER_ERRORCOUNT extension (#3653)

### LINKED ISSUE #3653: exit code propagation in case of CLOSESPIDER_ERRORCOUNT extension?
When I run scrapy with `CLOSESPIDER_ERRORCOUNT=1` and a spider raises unhandled exception, the command exits with successful code 0. 
However, in case of the errors, it would make sense to propagate the non-zero exit code to indicate that there were enough error occurred thus the job finished prematurely. 
I think this will help prevent the next batch job scheduled to not run on the wrong data.

```
scrapy crawl $SPIDER_NAME
...some stuff...
$ERROR_NAME: $ERROR_DETAIL
...Traceback...
[scrapy.core.engine] INFO: Closing spider (closespider_errorcount)
[scrapy.utils.signal] ERROR: Error caught on signal handler: <...>
[scrapy.statscollectors] INFO: Dumping Scrapy stats:
{ ...
'spider_exceptions/$ERROR_NAME': 1
....}
$ echo $?
0
```

HyunTruth:
Fixes #3653.
Changing exit code to non-zero for exit by CLOSESPIDER_ERRORCOUNT extension,
It will be helpful if the error in spider is critical in the following executions (bash, &&, etc.) to better reflect the existence of error.


scrapy/commands/crawl.py
```diff
@@ -55,7 +55,10 @@ def run(self, args, opts):
         spname = args[0]
 
         self.crawler_process.crawl(spname, **opts.spargs)
+        crawler = list(self.crawler_process.crawlers)[0]
```
Gallaecio: What happens if an error occurs on the second crawler? Or is there never a second crawler?

HyunTruth: Aha. From the source I saw a comment that multiple spider is deprecated, so thought there's only one instantiation. Thanks for letting me know. So there are indeed multiple crawlers?

HyunTruth:
```suggestion
        crawler = list(self.crawler_process.crawlers)
        self.crawler_process.start()
        closed_by_error_count = any([x.stats.get_value('finish_reason') == 'closespider_errorcount' for x in crawler])
```
Then we might try this instead?

Gallaecio: It looks better to me.

Although I believe `any` accepts an iterator, so you could use `any(…)` instead of `any([…])`.

HyunTruth: Well, I've used list comprehension to convert each crawler `x` to `x.stats.get_value('finish_reason') == 'closespider_errorcount'`, a boolean variable. I guess I can use map and lambda to do the same thing, like below.
```suggestion
        self.crawler_process.crawl(spname, **opts.spargs)
        self.crawler_process.start()

        closed_by_error_count = any(map(lambda x: x.stats.get_value('finish_reason') == 'closespider_errorcount',
                                        self.crawler_process.crawlers))
```

PR #3670, opened 2019-03-08
Clarify the documentation of DEPTH_PRIORITY further

### LINKED ISSUE #2289: Priority value, Higher or Lower?
There is a contradiction in the documentation.

http://doc.scrapy.org/en/latest/topics/settings.html?highlight=priority#depth-priority

vs

http://doc.scrapy.org/en/latest/topics/request-response.html?highlight=priority#scrapy.http.Request

The first says a positive number will mean lower priority. The latter says a higher number means a higher priority. So what is it?

Gallaecio:
Fixes #2289

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -331,16 +331,16 @@ Default: ``0``
 
 Scope: ``scrapy.spidermiddlewares.depth.DepthMiddleware``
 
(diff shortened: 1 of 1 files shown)


kmike:
This reads better, thanks @Gallaecio!

PR #3671, opened 2019-03-08
Implement Item.deepcopy()

### LINKED ISSUE #1493: Copying items not work correctly with list values
> > item = Item()
> > item['some_field'] = []
> > 
> > c_item = Item(item)
> > c_item['some_field'].append(1)
> > 
> > print item['some_field'], c_item['some_field']
> > [1] [1]

Gallaecio:
Fixes #1493

### DIFF
--- a/docs/topics/items.rst
+++ b/docs/topics/items.rst
@@ -40,6 +40,7 @@ objects. Here is an example::
         name = scrapy.Field()
         price = scrapy.Field()
         stock = scrapy.Field()
+        tags = scrapy.Field()
         last_updated = scrapy.Field(serializer=str)
 
 .. note:: Those familiar with `Django`_ will notice that Scrapy Items are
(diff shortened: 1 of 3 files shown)


docs/topics/items.rst
```diff
@@ -183,6 +174,33 @@ Creating items from dicts::
...
+
+    >>> product2 = Product(product)
+    >>> print(product2)
+    Product(name='Desktop PC', price=1000, tags=['tag1'])
+
+    >>> product3 = product.copy()
+    >>> print(product3)
+    Product(name='Desktop PC', price=1000, tags=['tag1'])
+
+    >>> product4 = product.deepcopy()
```
kmike: Hey, what do you think about providing some text describing it, not only an example? It'd be nice to mention caveats of using `Product(product)` or `product.copy()`, and tell where user should prefer which way of copying.

Gallaecio:
I’ve created a separate documentation section just for copying items. There I introduce the concepts of shallow copy and deep copy with an example specific to Scrapy items, and link to the Python documentation for further information. Then I introduce the syntax variantes to copy Scrapy items with inline code examples.

scrapy/item.py
```diff
@@ -96,6 +97,13 @@ def __repr__(self):
     def copy(self):
         return self.__class__(self)
 
+    def deepcopy(self):
+        """Returns a `deep copy`_ of this item.
```
kmike: Super-minor nitpick: according to [PEP-257](https://www.python.org/dev/peps/pep-0257/), it should be "Return ...", not "Returns ...".

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

kmike:
Updated docs look awesome, thanks @Gallaecio!

kmike:
Thanks @Gallaecio! AppVeyor failure is not related to this PR, so I'm merging it.

PR #3672, opened 2019-03-08
Add a FAQ entry about splitting items in item pipelines

### LINKED ISSUE #2240: Add FAQ entry on splitting items in spider middleware

Gallaecio:
Fixes #2240

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -319,6 +319,29 @@ I'm scraping a XML document and my XPath selector doesn't return any items
 
 You may need to remove namespaces. See :ref:`removing-namespaces`.
 
+How to split an item into multiple items in an item pipeline?
+-------------------------------------------------------------
+
+:ref:`Item pipelines <topics-item-pipeline>` cannot yield multiple items per
+input item. :ref:`Create a spider middleware <custom-spider-middleware>`
+instead, and use its
+:meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_spider_output`
+method for this puspose. For example::
+
+    from copy import deepcopy
+
+    from scrapy.item import BaseItem
+
+
+    class MultiplyItemsMiddleware:
+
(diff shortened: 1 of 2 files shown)


docs/faq.rst
```diff
@@ -319,6 +319,24 @@ I'm scraping a XML document and my XPath selector doesn't return any items
...
+input item. :ref:`Create a spider middleware <custom-spider-middleware>`
+instead, and use its
+:meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_spider_output`
+method for this puspose. For example::
+
+    class MultiplyItemsMiddleware:
+
+        def process_spider_output(self, response, result, spider):
+            for item in result:
+                if isinstance(item, scrapy.Item):
```
kmike: I think this is slightly more precise:
```suggestion
                if isinstance(item, (scrapy.item.BaseItem, dict)):
```
(it'd be good to add necessary imports as well)

docs/faq.rst
```diff
@@ -319,6 +319,24 @@ I'm scraping a XML document and my XPath selector doesn't return any items
...
+                    for _ in range(item['multiply_by']):
+                        yield item.copy()
```
kmike: What do you think about using `deepcopy.copy(item)` here, to avoid surprises? 
We can't use `item.deepcopy()` because item can be a dict.

Gallaecio:
Done.

I’ve also included the import statements, just in case, and made the question slightly shorted (“how can I” → “how to”).

Gallaecio:
Done.

I’ve also included the import statements, just in case, and made the question slightly shorted (“how can I” → “how to”).

kmike:

PR #3677, opened 2019-03-09
Require incremental>=16.10.1 to use Twisted>=17.9.0 under python 3 env

### LINKED ISSUE #2808: Mac os Python pip3 install Scrapy or Twisted gives“Command ”python setup.py egg_info“ failed with error code 1 in /private/tmp/pip-build-51x6vr4n/Twisted/”
sudo pip3 install Scrapy -i https://pypi.douban.com/simple 
Command "python setup.py egg_info" failed with error code 1 in /private/tmp/pip-build-s3504tu6/Twisted/

so I use sudo pip3 install Scrapy -i https://pypi.douban.com/simple but failed again. 
Please give me some advices, Thank you so much

silviaclaire:
Fixes #2808

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ setup(
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     install_requires=[
         'Twisted>=13.1.0',
+        'incremental',
         'w3lib>=1.17.0',
         'queuelib',
         'lxml',
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -1,4 +1,5 @@
 Twisted>=13.1.0
+incremental
 lxml
 pyOpenSSL
 cssselect>=0.9
--- a/requirements-py3.txt
+++ b/requirements-py3.txt
@@ -1,4 +1,5 @@
 Twisted>=17.9.0
+incremental>=16.10.1
 lxml>=3.2.4
 pyOpenSSL>=0.13.1
 cssselect>=0.9


setup.py
```diff
@@ -66,6 +66,7 @@ def has_environment_marker_platform_impl_support():
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     install_requires=[
         'Twisted>=13.1.0',
+        'incremental',
```
kmike: What do you think about adding a comment here, to explain why incremental is in install_requires?

kmike:
The fix makes sense to me, I don't see obvious downsides.

silviaclaire:
A downside case I could think of (mentioned  [here](https://github.com/scrapy/scrapy/issues/2808#issuecomment-471172778) but I didn’t  make it clear): 

If Twisted updates with a newer version of incremental **without fixing their issue**, then Scrapy will have to :
- look into what version of incremental they are using
- update our requirements for incremental to make the latest Twisted work

However, adding a simple statement in instruction (pip 10.0+) could get rid of this tracking and fix the problem as well.

Gallaecio:
@silviaclaire I believe requirement files are there for automated tests only, I don’t think they affect regular Scrapy installations.

Because of that, based on the [Twisted issue report](https://twistedmatrix.com/trac/ticket/9457) I am worried that adding `incremental` to `setup.py` might not solve the issue [described at StackOverflow](https://stackoverflow.com/a/43895561/939364). If `incremental` is already installed on the system, as described in StackOverflow, I believe this change will have no effect; if a Twisted version requiring a newer version of `incremental` is installed along with Scrapy, the issue might remain.

If I’m right, the only choice here might be to add an entry to the troubleshooting section of the installation guide to cover this.

PR #3682, opened 2019-03-13
[MRG+1] Rule.process_request: access Response object

elacuesta:
Allow for the `process_request` function passed to `Rule` objects to take the `Response` from which the `Request` was extracted as second positional argument.

This change grants the ability to access previously collected information in the `Response.meta` dict, for instance, which is not currently possible AFAICT (at least not without overriding private methods 🤷‍♂️ ). It is implemented in a backward compatible manner, i.e. request processing functions that take only one argument still work.

Updated docs, added tests for the existing and new functionality.

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -402,10 +402,12 @@ Crawling rules
    of links extracted from each response using the specified ``link_extractor``.
    This is mainly used for filtering purposes.
 
-   ``process_request`` is a callable, or a string (in which case a method from
-   the spider object with that name will be used) which will be called with
-   every request extracted by this rule, and must return a request or None (to
-   filter out the request).
+   ``process_request`` is a callable (or a string, in which case a method from
+   the spider object with that name will be used) which will be called for
+   every request extracted by this rule. This callable should take a Request object
+   as first positional argument and, optionally, the Response object from which the
+   Request originated as second positional argument. It must return a request or None
+   (to filter out the request).
 
 CrawlSpider example
 ~~~~~~~~~~~~~~~~~~~
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -403,11 +403,11 @@ Crawling rules
    This is mainly used for filtering purposes.
 
    ``process_request`` is a callable (or a string, in which case a method from
-   the spider object with that name will be used) which will be called for
-   every request extracted by this rule. This callable should take a Request object
-   as first positional argument and, optionally, the Response object from which the
-   Request originated as second positional argument. It must return a request or None
-   (to filter out the request).
+   the spider object with that name will be used) which will be called for every
+   :class:`~scrapy.http.Request` extracted by this rule. This callable should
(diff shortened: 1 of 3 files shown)


scrapy/spiders/crawl.py
```diff
@@ -19,17 +19,28 @@ def identity(x):
...
             self.follow = False if callback else True
         else:
             self.follow = follow
 
+    def _process_request(self, request, response):
+        """
+        Wrapper around the request processing function to maintain backward compatibility
+        with functions that do not take a Response object as parameter.
+        """
+        argcount = self.process_request.__code__.co_argcount
```
raphapassini: Very elegant solution :)

raphapassini:
LGTM

docs/topics/spiders.rst
```diff
@@ -402,10 +402,12 @@ Crawling rules
...
-   ``process_request`` is a callable, or a string (in which case a method from
-   the spider object with that name will be used) which will be called with
-   every request extracted by this rule, and must return a request or None (to
-   filter out the request).
+   ``process_request`` is a callable (or a string, in which case a method from
+   the spider object with that name will be used) which will be called for
+   every request extracted by this rule. This callable should take a Request object
+   as first positional argument and, optionally, the Response object from which the
+   Request originated as second positional argument. It must return a request or None
+   (to filter out the request).
```
Gallaecio: Do we want to keep supporting a single parameter forever? Otherwise, I would remove the ‘optionally’ part.

scrapy/spiders/crawl.py
Gallaecio: What about a deprecation warning when the specified function only expects one argument?

elacuesta: That's a good idea, I'll get to it, thanks!

kmike: There are also some related utilities in `scrapy.utils.python`: `get_func_args` and `get_spec`. I wonder if we can use an existing function. If not, I think this code should be extracted to a new helper in scrapy.utils.python, with dedicated tests for it.

elacuesta: Awesome, `get_func_args` did the trick.

elacuesta:
Updated, now a `ScrapyDeprecationWarning` is raised if the `process_request` function takes only one argument.
Currently, the warning is logged only when the function is "compiled" (had to move the compiling code, the alternative was to check on every function call and that would be a lot of warnings). I think it would be better if it were executed when the `Rule` is instantiated, the problem is they need to be compiled from the `CrawlSpider` object to which they belong because `process_response` could be a string with the name of a method in the spider (in which case there is no way of analyzing the function from the `Rule`'s constructor).
However, as far as I understand `Rule` objects are not really meant for standalone use and are not supposed to be ready until they are compiled, so the above is probably not an issue in practice.
Looking forward to reading your comments.

docs/topics/spiders.rst
elacuesta: Updated docs, rephrased a bit and removed the "optionally" part

kmike:

PR #3685, opened 2019-03-15
Retry request from spider callback

### LINKED ISSUE #3590: Ability to retry a request from inside a spider callback

kasun:
New approach for #3590

This,
1) provides a method retry_request() in response class that will return a request instance that is marked as to be retried
2) A spidermiddleware (RetryMiddleware) will check whether this request can be retried, do the necessary modifications and retry the request

Resolves #3590

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -96,3 +96,31 @@ class Request(object_ref):
             kwargs.setdefault(x, getattr(self, x))
         cls = kwargs.pop('cls', self.__class__)
         return cls(*args, **kwargs)
+
+    def mark_for_retry(self, reason=None, **kwargs):
+        """Return a new instance that is marked as to be retried.
+        New request is same as this request except attributes provided as keyword arguments.
+
+        :param reason: Reason for retrying. Optional, defaults to None.
+        :type reason: str"""
+        new_req = self.replace(**kwargs)
+        new_req.meta['_retry_requested'] = True
+        new_req.meta['_retry_request_reason'] = reason
+        return new_req
+
+    def unmark_as_retry(self):
+        """Unmark this request as to be retried"""
+        self.meta.pop('_retry_requested', None)
+        self.meta.pop('_retry_request_reason', None)
+
+    def is_marked_for_retry(self):
+        """Return whether this request is marked as to be retried."""
+        return self.meta.get('_retry_requested') is True
+
+    def get_retry_reason(self):
+        """Return reason for retry if this request is marked for retry"""
+        return self.is_marked_for_retry() and self.meta.get('_retry_request_reason') or None
+
+    def is_retrying_enabled(self):
+        """Return whether retrying is enabled on this request"""
+        return not bool(self.meta.get('dont_retry', False))
--- a/scrapy/http/response/__init__.py
+++ b/scrapy/http/response/__init__.py
@@ -113,8 +113,8 @@ class Response(object_ref):
         It accepts the same arguments as ``Request.__init__`` method,
         but ``url`` can be a relative URL or a ``scrapy.link.Link`` object,
         not only an absolute URL.
-        
-        :class:`~.TextResponse` provides a :meth:`~.TextResponse.follow` 
+
+        :class:`~.TextResponse` provides a :meth:`~.TextResponse.follow`
         method which supports selectors in addition to absolute/relative URLs
         and Link objects.
         """
@@ -133,3 +133,23 @@ class Response(object_ref):
                        priority=priority,
                        dont_filter=dont_filter,
                        errback=errback)
+
+    def retry_request(self, reason=None, **kwargs):
+        """Retry the request tied to this response.
+
+        This returns a new :class:`~.Request` instance that is marked to be retried
+        which is as same as the request tied to this response except attributes
+        provided as keyword arguments. Scrapy will schedule this new request if request retrying
+        is `enabled <https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-enabled>`_
+        and maximum number of
+        `allowed retries <https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-times>`_
+        are not reached.
+
+        :param reason: Reason for retrying. Optional, defaults to None.
+        :type reason: str
+        """
+        request = self.request
+        if not request:
+            raise NotSupported('This response is not tied to any request')
+
+        return request.mark_for_retry(reason, **kwargs)
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -258,6 +258,7 @@ SPIDER_MIDDLEWARES_BASE = {
     'scrapy.spidermiddlewares.referer.RefererMiddleware': 700,
     'scrapy.spidermiddlewares.urllength.UrlLengthMiddleware': 800,
     'scrapy.spidermiddlewares.depth.DepthMiddleware': 900,
+    'scrapy.spidermiddlewares.retry.RetryMiddleware': 950,
     # Spider side
 }
 
--- a/scrapy/spidermiddlewares/retry.py
+++ b/scrapy/spidermiddlewares/retry.py
@@ -1,0 +1,50 @@
+"""
+RetryMiddleware is used to retry temporary issues or bans where the response has to be
+processed by spider before been retried.
+
+You can issue a retry request by calling :meth:`scrapy.http.response.Response.retry_request` method on a
+:class:`scrapy.http.response.Response` instance.
+"""
+import logging
+
+from scrapy import Request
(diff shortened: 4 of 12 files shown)


scrapy/spidermiddlewares/retry.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.settings)
+
+    def process_spider_output(self, response, result, spider):
+        for x in result:
+            if isinstance(x, Request):
+                if x.is_marked_for_retry():
+                    if not self.enabled:
```
ejulio: This could be the first if statement of the method.
Maybe, it would be even better to `raise NotConfigured` in `__init__` if `RETRY_ENABLED` is _False_

scrapy/utils/retry.py
```diff
@@ -0,0 +1,54 @@
...
+class RetryHandler(object):
+    """utilities to handle retries"""
+
+    def __init__(self, spider, request):
+        self.spider = spider
+        self.request = request
+        self.max_retry_times = spider.settings.getint('RETRY_TIMES')
+        self.priority_adjust = spider.settings.getint('RETRY_PRIORITY_ADJUST')
+
+    def is_exhausted(self):
```
ejulio: Maybe a property.

scrapy/http/request/__init__.py
```diff
@@ -96,3 +96,31 @@ def replace(self, *args, **kwargs):
...
+        new_req.meta['_retry_requested'] = True
+        new_req.meta['_retry_request_reason'] = reason
+        return new_req
+
+    def unmark_as_retry(self):
+        """Unmark this request as to be retried"""
+        self.meta.pop('_retry_requested', None)
+        self.meta.pop('_retry_request_reason', None)
+
+    def is_marked_for_retry(self):
```
ejulio: Maybe a property here and in the following ones.

scrapy/http/request/__init__.py
```diff
@@ -96,3 +96,31 @@ def replace(self, *args, **kwargs):
             kwargs.setdefault(x, getattr(self, x))
         cls = kwargs.pop('cls', self.__class__)
         return cls(*args, **kwargs)
+
+    def mark_for_retry(self, reason=None, **kwargs):
```
ejulio: I think that `return req.retry()` reads better than `req.mark_for_retry()`.
Besides that, _mark_, for me, reads as something that is changing the instance and not creating a new one.

scrapy/http/response/__init__.py
```diff
@@ -133,3 +133,23 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        provided as keyword arguments. Scrapy will schedule this new request if request retrying
+        is `enabled <https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-enabled>`_
+        and maximum number of
+        `allowed retries <https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-times>`_
+        are not reached.
+
+        :param reason: Reason for retrying. Optional, defaults to None.
+        :type reason: str
+        """
+        request = self.request
```
ejulio: Not sure if the variable is required.

scrapy/settings/default_settings.py
```diff
@@ -258,6 +258,7 @@
     'scrapy.spidermiddlewares.referer.RefererMiddleware': 700,
     'scrapy.spidermiddlewares.urllength.UrlLengthMiddleware': 800,
     'scrapy.spidermiddlewares.depth.DepthMiddleware': 900,
+    'scrapy.spidermiddlewares.retry.RetryMiddleware': 950,
```
ejulio: This can cause some confusion because now there are two `RetryMiddlewares`.
One is a spider middleware and the other one is a downloader middleware.

scrapy/spidermiddlewares/retry.py
```diff
@@ -0,0 +1,50 @@
...
+        self.enabled = settings.getbool('RETRY_ENABLED')
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.settings)
+
+    def process_spider_output(self, response, result, spider):
+        for x in result:
+            if isinstance(x, Request):
+                if x.is_marked_for_retry():
```
ejulio: Maybe `if x.marked_for_retry:`

scrapy/spidermiddlewares/retry.py
```diff
@@ -0,0 +1,50 @@
...
+                        )
+                        continue
+
+                    reason = x.get_retry_reason() or 'Retry Requested'
+                    retry_handler = RetryHandler(spider, x)
+                    if retry_handler.is_exhausted():
+                        retry_handler.record_retry_failure(reason)
+                    else:
+                        new_req = retry_handler.make_retry_request()
+                        retry_handler.record_retry(new_req, reason)
```
ejulio: I think this should be called inside `make_retry_request`.
Otherwise, I could create a retry without recording it.

scrapy/http/request/__init__.py
ejulio: Besides that, I don't know if this behavior should be here.
Should a request know about the configurations other modules need to retry it?
Or should the retrying module set this configurations in the request?

scrapy/spidermiddlewares/retry.py
```diff
@@ -0,0 +1,50 @@
...
+                        logger.debug(
+                            "Dropping retry request because request retrying is disabled: %(request)s",
+                            {'request': x},
+                            extra={'spider': spider}
+                        )
+                        continue
+
+                    reason = x.get_retry_reason() or 'Retry Requested'
+                    retry_handler = RetryHandler(spider, x)
+                    if retry_handler.is_exhausted():
```
ejulio: Maybe `make_retry_request` should return `None` if it is exhausted.
Then you can remove this if statement from here.

scrapy/utils/retry.py
```diff
@@ -0,0 +1,54 @@
+from scrapy.utils.python import global_object_name
+
+
+class RetryHandler(object):
+    """utilities to handle retries"""
+
+    def __init__(self, spider, request):
```
ejulio: For me, it seems that the most important dependency is the request.
So, I would change the precedence to put the request first.

scrapy/utils/retry.py
```diff
@@ -0,0 +1,54 @@
...
+        self.spider = spider
+        self.request = request
+        self.max_retry_times = spider.settings.getint('RETRY_TIMES')
```
ejulio: Maybe setting default values, or raising an specific exception if these settings aren't available.

scrapy/utils/retry.py
```diff
@@ -0,0 +1,54 @@
+from scrapy.utils.python import global_object_name
+
+
+class RetryHandler(object):
```
ejulio: This handler does not required any state, so it could be a set of functions.
At least, it could be instantiated only once to store the settings and then the requests could be arguments to methods.

scrapy/http/request/__init__.py
apalala: I agree with `req.retry()`, which is similar to `req.replace()`.

scrapy/http/response/__init__.py
```diff
@@ -133,3 +133,23 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
                        priority=priority,
                        dont_filter=dont_filter,
                        errback=errback)
+
+    def retry_request(self, reason=None, **kwargs):
```
apalala: Here too, there's `Response.follow()` which returns a `Request`. Why not `Response.retry()`?

scrapy/utils/retry.py
apalala: It should be possible to override `max_retry_times` per request.

scrapy/spidermiddlewares/retry.py
kasun: This was mostly done for the debug message. Just in case if a user disabled request retrying globally and then tries to retry a request.

ShanikaEdiriweera:
Any update?

fellipegpbotelho:
Any updates?

PR #3688, opened 2019-03-15
Create a separate page about getting help

Gallaecio:


docs/index.rst
```diff
@@ -201,6 +185,9 @@ Solving specific problems
 :doc:`topics/jobs`
     Learn how to pause and resume crawls for large spiders.
 
+:doc:`getting-help`
+    Find support beyond the Scrapy documentation.
```
kmike: What do you think about making this link more prominent somehow? Currently it is in a middle of index, there are 30 topics before it and 10 topics after.

Gallaecio: Part of my goal putting it there was to try and prevent people from going straight to StackOverflow before reading the relevant documentation. However, given the new page covers that part, I would not mind moving the topic back up.

Maybe at the end of the First steps section?

Gallaecio: Moved to the end of *First steps*.

Gallaecio:
@kmike Would you like me to move it to the first part of First Steps, or change this pull request otherwise?

Gallaecio:
Moved the new “Getting help” to the beginning of the “First steps” section, so it continues to be the first topic of the whole documentation.

PR #3689, opened 2019-03-15
[MRG+1] Do not degrade JPEG files.

### LINKED ISSUE #3055: [bug] pillow will always recode images in imagepipieline
https://github.com/scrapy/scrapy/blob/aa83e159c97b441167d0510064204681bbc93f21/scrapy/pipelines/images.py#L151

this line will always recode images silently and damage the image quality.

please add an option to avoid this.

anubhavp28:
Fixes #3055 by not converting a JPEG file again.


anubhavp28:
Hey @Gallaecio, could you help me please? The tests are failing only for Python 3.7. I am not sure if it is related to my changes.

anubhavp28:
Hey @kmike, could you help me with this please? The tests are failing only for Python 3.7. I am not sure if it is related to my changes.

Gallaecio:
The error is not related to your pull request. That test failure occurs randomly, we’ll have to look into it eventually. In the meantime, I’ve closed and reopened your pull request, which triggers the test runs again. Hopefully there will be no false positives this time.

anubhavp28:
@Gallaecio Can you review this again? I have made the changes suggested by you.

tests/test_pipeline_images.py
```diff
@@ -79,28 +79,29 @@ def test_convert_image(self):
         SIZE = (100, 100)
         # straigh forward case: RGB and JPEG
         COLOUR = (0, 127, 255)
-        im = _create_image('JPEG', 'RGB', SIZE, COLOUR)
-        converted, _ = self.pipeline.convert_image(im)
+        im, buf = _create_image('JPEG', 'RGB', SIZE, COLOUR)
+        converted, converted_buf = self.pipeline.convert_image(im, buf)
         self.assertEqual(converted.mode, 'RGB')
         self.assertEqual(converted.getcolors(), [(10000, COLOUR)])
+        self.assertEqual(converted_buf.read(), buf.read())
```
Gallaecio: This test is insufficient, because if you comment out the two lines that you added to fix the issue, it passes nonetheless. We need a test that fails before the fix and succeeds afterwards.

You might need to use an actual binary image, as small as possible, instead of creating an image on the fly with Pillow.

anubhavp28:
@Gallaecio Can you review this again please? If I am not wrong, my changes have decreased the code coverage, hence codecov is complaining. How should I increase it?

scrapy/pipelines/images.py
```diff
@@ -152,6 +152,10 @@ def convert_image(self, image, size=None):
             image = image.copy()
             image.thumbnail(size, Image.ANTIALIAS)
 
+        if not size and image.format == 'JPEG':
+            buf = BytesIO(response_body.read())
```
Gallaecio: Is this line necessary?

tests/test_pipeline_images.py
```diff
@@ -79,28 +79,70 @@ def test_convert_image(self):
         SIZE = (100, 100)
         # straigh forward case: RGB and JPEG
         COLOUR = (0, 127, 255)
-        im = _create_image('JPEG', 'RGB', SIZE, COLOUR)
-        converted, _ = self.pipeline.convert_image(im)
+        im, buf = _create_image('JPEG', 'RGB', SIZE, COLOUR)
+        converted, converted_buf = self.pipeline.convert_image(im, buf)
```
Gallaecio: Purely :lipstick: aesthetic, so feel free to ignore it: since the `converted_buf` is not used, it may make sense not to name it (also applies to other changes):

```suggestion
        converted, _ = self.pipeline.convert_image(im, buf)
```

tests/test_pipeline_images.py
```diff
@@ -79,28 +79,70 @@ def test_convert_image(self):
...
+        b'\xe3\xe4\xe5\xe6\xe7\xe8\xe9\xea\xf2\xf3\xf4\xf5\xf6\xf7\xf8\xf9'
+        b'\xfa\xff\xda\x00\x0c\x03\x01\x00\x02\x11\x03\x11\x00?\x00\xbb'
+        b'\xe2\x1b\xcb\x88\xe4\x90\t\x0e1\xd2\xb9\xab\x1dF\xe6\xda\xe0\xb4l\xd9'
+        b'5>\xaf\xac\xc7y}$k\x92\x03u\xaaL\xeb\x18VL\x9fZ+\xb9\xb9EBV]Ow\x0f\x8a\xc0'
+        b'\xac\x14\xbd\xa2NG_\xa5x\x8esg\xfb\xd9\x8e\xed\xc79\xa2\xa1\xd0\xf4\x8bk'
+        b'\x9d5e\xdeN\xe6=\xa8\xae\xa9V\xc2\xb6|\xdf2\xeep:w\xfc\x84d\x04g\x9e\xf5\xd3'
+        b'L\xa9\x0c\x1f*/>\xa2\x8a+\xca\x93z\x1c\x15]\xa9\xc6\xc4Q\xea\x97V\xca'
+        b'c\x85\xc2\xaes\x80(\xa2\x8a\xe3\x92W<\xdb\x9f\xff\xd9'))
+        im = Image.open(buf)
+        buf.seek(0)
```
Gallaecio: Is this line necessary?

tests/test_pipeline_images.py
```diff
@@ -79,28 +79,70 @@ def test_convert_image(self):
...
+        converted, converted_buf = self.pipeline.convert_image(im, buf)
+        converted_buf.seek(0)
+        buf.seek(0)
+        self.assertEqual(im.format, "JPEG")
+        self.assertEqual(converted_buf.read(), buf.read())
```
Gallaecio: I think that, if you remove the line I mentioned from `scrapy/pipelines/images.py`, here you could directly compare the input parameters and the output parameters, asserting that they are the same, so you do not need to use `seek()` or `read()` at all:

```
converted, converted_buf = self.pipeline.convert_image(im, buf)
self.assertEqual(im, converted)
self.assertEqual(buf, converted_buf)
```

Gallaecio:
I’ve left a few minor comments. But do not worry about coverage, its integration with GitHub feels a bit clunky at times, if you look at the actual details I think you are increasing coverage.

tests/test_pipeline_images.py
anubhavp28: Got it! I was confused. For some reason, I thought the function should always return a copy of `image` and `buffer`.

Gallaecio:
I think this is ready. Thank you!

scrapy/pipelines/images.py
```diff
@@ -127,15 +127,15 @@ def get_images(self, response, request, info):
...
         yield path, image, buf
 
         for thumb_id, size in six.iteritems(self.thumbs):
             thumb_path = self.thumb_path(request, thumb_id, response=response, info=info)
-            thumb_image, thumb_buf = self.convert_image(image, size)
+            thumb_image, thumb_buf = self.convert_image(image, buf, size)
             yield thumb_path, thumb_image, thumb_buf
 
-    def convert_image(self, image, size=None):
+    def convert_image(self, image, response_body, size=None):
```
kmike: Unfortunately this is backwards incompatible; there are image pipeline subclasses in a wild which override this method - they'll stop working because of signature change. Pipeline may also override a method which calls this method. Could you please make it backwards compatible? It seems something like this is needed:

* make response_body optional (None by default), make it a last argument
* return buf if response_body is not passed (i.e. disable the feature in this PR if a middleware uses deprecated signature).
* because user may override convert_image method, pipeline shouldn't always pass response_body - it should inspect convert_image method signature, and pass response_body only if this argument is present.
* issue a warning when response_body is not passed (ScrapyDeprecationWarning) - it seems two warnings are needed, one when convert_image is overridden in incompatible way, and another one is when convert_image is called from an overridden method in incompatible way.

anubhavp28: @kmike I have made changes, could you review this again?

Gallaecio: @anubhavp28 One thing you could do, since the goal is to support the previous API, is to, instead of modifying the tests to use the new API, keep the old tests and add new tests that test the new API.

You could also extend the old tests to ensure that a deprecation warning is used.

anubhavp28: @Gallaecio Not sure how to write tests to ensure that deprecation warnings are received. Currently, I am just counting the number of warnings. Would that be enough?
https://github.com/scrapy/scrapy/blob/881bade2c1b2d5842f7d0bdf2648455996f814b7/tests/test_pipeline_images.py#L111

scrapy/pipelines/images.py
```diff
@@ -126,16 +126,49 @@ def get_images(self, response, request, info):
...
+            image, buf = self.convert_image(orig_image)
+        else:
+            image, buf = self.convert_image(orig_image, response_body=BytesIO(response.body))
         yield path, image, buf
 
         for thumb_id, size in six.iteritems(self.thumbs):
             thumb_path = self.thumb_path(request, thumb_id, response=response, info=info)
-            thumb_image, thumb_buf = self.convert_image(image, size)
+            if convert_image_overriden:
+                _warn()
```
Gallaecio: I don’t think you need to log a warning each time you check the value of `convert_image_overriden`, logging the warning once should be enough.

scrapy/pipelines/images.py
Gallaecio: @anubhavp28 Please, check other usages of `catch_warnings` in the tests and you’ll see how to check the warning message as well. You could ensure that the warnings contain certain text to ensure that they are the expected warnings, and not some other warnings.

anubhavp28:
@Gallaecio I have made few changes. Though the AppVeyor build is failing, it doesn't seem to be related to my changes? Could you review this again?

scrapy/pipelines/images.py
```diff
@@ -126,16 +126,48 @@ def get_images(self, response, request, info):
...
+            if 'response_body' not in convert_image_signature.args:
+                return True
+            return False
 
-        image, buf = self.convert_image(orig_image)
+        def _warn():
+            from scrapy.exceptions import ScrapyDeprecationWarning
+            import warnings
+            warnings.warn('ImagesPipeline.convert_image() method overriden in a incompatible way, '
+                          'overriden method does not accept response_body argument.',
```
Gallaecio: You are actually making quite an effort to make the old way work, so I would say “deprecated” instead of “incompatible“ here :slightly_smiling_face: 

Also applied to the other warning below.

anubhavp28:
@Gallaecio I have made the changes suggested by you. Could you review this again?

scrapy/pipelines/images.py
```diff
@@ -126,16 +126,37 @@ def get_images(self, response, request, info):
...
+        
+        def _warn():
+            from scrapy.exceptions import ScrapyDeprecationWarning
+            import warnings
+            warnings.warn('ImagesPipeline.convert_image() method overriden in a deprecated way, '
+                          'overriden method does not accept response_body argument.',
+                          category=ScrapyDeprecationWarning, stacklevel=1)
 
-        image, buf = self.convert_image(orig_image)
+        if 'response_body' not in get_func_args(self.convert_image):
```
Gallaecio: I wonder if we should call `get_func_args` only once and cache its response, instead of calling it every time we need to check its result. `self.convert_image` is not likely to change at run time.

scrapy/pipelines/images.py
```diff
@@ -126,16 +126,38 @@ def get_images(self, response, request, info):
...
+        
+        def _warn():
+            from scrapy.exceptions import ScrapyDeprecationWarning
+            import warnings
+            warnings.warn('ImagesPipeline.convert_image() method overriden in a deprecated way, '
+                          'overriden method does not accept response_body argument.',
+                          category=ScrapyDeprecationWarning, stacklevel=1)
 
-        image, buf = self.convert_image(orig_image)
+        convert_image_overriden = 'response_body' not in get_func_args(self.convert_image)
```
Gallaecio: I was actually thinking of caching it even further:

```python
def __init__(self, …):
    # …
    self._deprecated_convert_image = None

def get_images(self, …):
    # …

    if self._deprecated_convert_image is None:
        self._deprecated_convert_image = 'response_body' not in get_func_args(self.convert_image)
        if self._deprecated_convert_image:
            import warnings
            warnings.warn('ImagesPipeline.convert_image() method overriden in a deprecated way, '
                          'overriden method does not accept response_body argument.',

    if self._deprecated_convert_image:
        image, buf = self.convert_image(orig_image)
    else:
        image, buf = self.convert_image(orig_image, response_body=BytesIO(response.body))

    # …
```

scrapy/pipelines/images.py
```diff
@@ -127,15 +129,37 @@ def get_images(self, response, request, info):
             raise ImageException("Image too small (%dx%d < %dx%d)" %
                                  (width, height, self.min_width, self.min_height))
 
-        image, buf = self.convert_image(orig_image)
+        if self._deprecated_convert_image is None:
+            self._deprecated_convert_image = 'response_body' not in get_func_args(self.convert_image)
+            if self._deprecated_convert_image:
+                from scrapy.exceptions import ScrapyDeprecationWarning
+                import warnings
```
kmike: Could you please move these import to the top level, as well as imports in convert_image?

scrapy/pipelines/images.py
```diff
@@ -127,15 +129,37 @@ def get_images(self, response, request, info):
...
-            thumb_image, thumb_buf = self.convert_image(image, size)
+            if self._deprecated_convert_image:
+                thumb_image, thumb_buf = self.convert_image(image, size)
+            else:
+                thumb_image, thumb_buf = self.convert_image(image, size, buf)
             yield thumb_path, thumb_image, thumb_buf
 
-    def convert_image(self, image, size=None):
+    def convert_image(self, image, size=None, response_body=None):
+        if not response_body:
```
kmike:
```suggestion
        if response_body is None:
```
I think response_body can be really empty in some cases.

scrapy/pipelines/images.py
```diff
@@ -151,7 +175,9 @@ def convert_image(self, image, size=None):
         if size:
             image = image.copy()
             image.thumbnail(size, Image.ANTIALIAS)
-
+        elif response_body and image.format == 'JPEG':
```
kmike:
```suggestion
        elif response_body is not None and image.format == 'JPEG':
```

tests/test_pipeline_images.py
```diff
@@ -76,35 +76,71 @@ def test_thumbnail_name(self):
                          'thumbs/50/850233df65a5b83361798f532f1fc549cd13cbe9.jpg')
 
     def test_convert_image(self):
+        # tests for old API
+        with warnings.catch_warnings(record=True) as w:
```
kmike: could you please move "tests for old API" to a separate test method (or multiple test methods)?

Issue #3690, opened 2019-03-15
Separate Attribute to allow offsite requests


kasun:
Right now Request.dont_filter attribute is used for two intentions. To allow duplicated requests and to allow offsite requests. This behavior is not desirable since, you could want to allow a duplicate request but still want to filter out offsite requests and vice versa.

VIDMORE:
Hello, is this Issue still open?

PR #3691, opened 2019-03-15
Attribute to control offsite filtering

kasun:
This is a fix for #3690

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -24,7 +24,7 @@ below in :ref:`topics-request-response-ref-request-subclasses` and
 Request objects
 ===============
 
-.. class:: Request(url[, callback, method='GET', headers, body, cookies, meta, encoding='utf-8', priority=0, dont_filter=False, allow_offsite_requests=False, errback, flags])
+.. class:: Request(url[, callback, method='GET', headers, body, cookies, meta, encoding='utf-8', priority=0, dont_filter=False, errback, flags])
 
     A :class:`Request` object represents an HTTP request, which is usually
     generated in the Spider and executed by the Downloader, and thus generating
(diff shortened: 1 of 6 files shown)


raphapassini:
LGTM

docs/topics/request-response.rst
```diff
@@ -311,6 +311,7 @@ Those are:
 * ``ftp_password`` (See :setting:`FTP_PASSWORD` for more info)
 * :reqmeta:`referrer_policy`
 * :reqmeta:`max_retry_times`
+* :reqmeta:`allow_offsite_requests`
```
Gallaecio: Since this is a value for `Request.meta`, maybe it could be named simply `allow_offsite`?

docs/topics/spider-middleware.rst
```diff
@@ -318,10 +320,10 @@ OffsiteMiddleware
    :attr:`~scrapy.spiders.Spider.allowed_domains` attribute, or the
    attribute is empty, the offsite middleware will allow all requests.
 
-   If the request has the :attr:`~scrapy.http.Request.dont_filter` attribute
-   set, the offsite middleware will allow the request even if its domain is not
-   listed in allowed domains.
 
+   If the request has `allow_offsite_requests` set to True in
```
Gallaecio: :lipstick: ``` ``True`` ```

scrapy/spidermiddlewares/offsite.py
```diff
@@ -28,7 +28,7 @@ def from_crawler(cls, crawler):
     def process_spider_output(self, response, result, spider):
         for x in result:
             if isinstance(x, Request):
-                if x.dont_filter or self.should_follow(x, spider):
+                if x.meta.get('allow_offsite_requests', False) or self.should_follow(x, spider):
```
Gallaecio: I am thinking this would keep this change backward-compatible:
```suggestion
                if x.meta.get('allow_offsite_requests', x.dont_filter) or self.should_follow(x, spider):
```

If you agree, then I suggest you also change your test changes, so that you keep existing tests unchanged and simply add additional tests scenarios.

docs/topics/request-response.rst
kasun: Agreed.

scrapy/spidermiddlewares/offsite.py
kasun: I'm not too sure about this even with the advantage of backward compatibility because when we want to duplicate a request but not skip the offsite filter ( which I assume is the case for majority of instances where we set dont_filter=True) we have to add meta={'allow_offsite': False} which is extra work. Plus there will be two ways to skip offsite filtering.

Hence I'm proposing to do this in a backward incompatible way if possible.

PR #3696, opened 2019-03-22
Allow customizing export column names

### LINKED ISSUE #1008: Non-ascii field names
In Python 2 variable names are limited by ascii character set.
CSV Scrapy exporter uses Item's field names as an csv title row.
I want to use unicode characters for csv title row.
Can I do that without writing my own exporter?

### LINKED ISSUE #3266: Allow defining field names that don't follow the restrictions of python variables
Sometimes we may require defining field names that are not valid Python variables (for example containing spaces). I believe the correct way of doing this right now is that the developer needs to add a Pipeline that maps the internal name to the name we wan't to export. Scrapy could ship with a Pipeline (maybe enabled by default) that does this mapping for fields that define a field parameter. For example:

```
class MyItem(Item):
   date_from = Field(export_name='Initial date')
```

This is a very simple pipeline. However, since this is a very common use case, I think it is something that could be shipped by default. I'm happy to provide a PR for this.

Gallaecio:
Fixes #1008, fixes #3266


docs/topics/feed-exports.rst
```diff
@@ -259,18 +259,36 @@ FEED_EXPORT_FIELDS
...
+
+    .. _OrderedDict: https://docs.python.org/library/collections.html#collections.OrderedDict
+
+    For example::
+
+        FEED_EXPORT_FIELDS = OrderedDict((
+            ("foo", "Foo"),
+            ("bar", "Bar"),
+            ("baz", "Baz"),
+        ))
```
kmike: What do you think about making it possible/official to use dicts as well? 

* it can be easier if people don't care about the order
* dict order is actually guaranteed in Python 3.7 (and it is preserved in practice since Python 3.6)

Gallaecio: Done.

I’ve also refactored the documentation a bit more, trying to keep things simpler (less text to say the same) and avoid repetition within reason (the setting documentation links to the attribute documentation.

Please, let me know what you think.

scrapy/settings/__init__.py
```diff
@@ -197,6 +198,39 @@ def getdict(self, name, default=None):
...
+    def getdictorlist(self, name, default=None):
+        """Get a setting value as either an ``OrderedDict`` or a list.
+
+        If the setting is already a dict or a list, a copy of it will be
+        returned.
+
+        If it is a string it will be evaluated as JSON, or as a comma-separated
+        list of strings as a fallback.
+
+        For example, settings populated through environment variables will
```
kmike: Do we support populating settings through environment variables? I recall there was something about it, but not sure it is documented.

scrapy/settings/__init__.py
```diff
@@ -197,6 +198,39 @@ def getdict(self, name, default=None):
...
+
+        :param name: the setting name
+        :type name: string
+
+        :param default: the value to return if no setting is found
+        :type default: any
+        """
+        value = self.get(name, default)
+        if value is None:
+            return {}
```
kmike: This is not OrderedDict, as documented.

tests/test_feedexport.py
```diff
@@ -606,6 +607,62 @@ class MyItem2(scrapy.Item):
...
+        ]
+        yield self.assertExported(items, header, rows,
+                                  settings=settings, ordered=True)
+
+        # custom output field names can be specified
+        header = OrderedDict((
+            ("foo", "Foo"),
+            ("baz", "Baz"),
+            ("hello", "Hello"),
+        ))
```
kmike: Could you please add a test for a regular dict, and ensure that order is preserved in Python 3.7+?

Gallaecio:
Should we consider this a potential candidate for 2.7?

wRAR:
I'm fine with that.

scrapy/settings/__init__.py
```diff
@@ -197,6 +198,38 @@ def getdict(self, name, default=None):
             value = json.loads(value)
         return dict(value)
 
+    def getdictorlist(self, name, default=None):
+        """Get a setting value as either an ``OrderedDict`` or a list.
```
kmike, line 202: I wonder if OrderedDict support is needed, given that Scrapy requires Python 3.6+

tests/test_feedexport.py
```diff
@@ -959,22 +961,71 @@ def test_export_multiple_item_classes(self):
...
+    def test_export_items_field_names(self):
+        items = [{'foo': 'bar'}]
+        header = {'foo': 'Foo'}
+        rows = [{'Foo': 'bar'}]
+        settings = {'FEED_EXPORT_FIELDS': header}
+        yield self.assertExported(items, list(header.values()), rows,
+                                  settings=settings)
+
+    @pytest.mark.skipif(sys.version_info < (3, 7),
+                        reason='Only official in Python 3.7+')
```
kmike, line 1008: it seems this can be removed now

tests/test_feedexport.py
```diff
@@ -1118,7 +1169,7 @@ def test_export_feed_export_fields(self):
                 {'egg': 'spam2', 'foo': 'bar2', 'baz': 'quux2'}
             ]
             yield self.assertExported(items, ['foo', 'baz', 'egg'], rows,
-                                      settings=settings, ordered=True)
+                                      settings=settings)
```
kmike, line 1172: Do we still need "ordered" parameter in assertExported / assertExportedCsv, or can we remove code for handling ordered=False?

Gallaecio: It seems so, https://github.com/scrapy/scrapy/pull/3696/commits/24f382fa459434cccfa4c0a8884a48d09d75e243 caused at least 3 tests to fail due to a different output order.

Gallaecio: Hmm… Looking in more details, maybe the issues are solvable without allowing a different order…

kmike:
Thanks @Gallaecio!

PR #3701, opened 2019-03-25
Update the documentation policies: Ask to use docstrings to document …

Gallaecio:
…API members

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -165,18 +165,18 @@ Scrapy:
 Documentation policies
 ======================
 
-* **Don't** use docstrings for documenting classes, or methods which are
-  already documented in the official (sphinx) documentation. Alternatively,
-  **do** provide a docstring, but make sure sphinx documentation uses
-  autodoc_ extension to pull the docstring. For example, the
-  :meth:`ItemLoader.add_value` method should be either
-  documented only in the sphinx documentation (not as a docstring), or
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -165,18 +165,9 @@ Scrapy:
...
-  documented only in the sphinx documentation (not as a docstring), or
-  it should have a docstring which is pulled to sphinx documentation using
-  autodoc_ extension.
-
-* **Do** use docstrings for documenting functions not present in the official
-  (sphinx) documentation, such as functions from ``scrapy.utils`` package and
-  its sub-modules.
+Use docstrings for documenting API members (classes, methods, etc.) and make
+sure that the Sphinx documentation uses the autodoc_ extension to pull the
+docstring.
```
kmike: 👍 

What do you think about making it a bit more nuanced? I think that's fine to have more lenghty tutorial-style docs in actual docs, even if they relate to a particular class. In this case only part of the docs may be pulled from docstrings. There shouldn't be duplication, and we should prefer docstrings, but we may use a mix of docs + docstrings when it makes sense. Are you ok with something like that? If yes, do you think we should mention this point explicitly?

Gallaecio: I agree, but I’m unsure how to word it. I’ve just made an attempt, let me know what you think.

kmike: Looks perfect to me, thanks @Gallaecio!

PR #3703, opened 2019-03-27
Add a topic about reaching data that selectors cannot reach

Gallaecio:
Write a new topic that addresses [Can scrapy be used to scrape dynamic content from websites that are using AJAX?](https://stackoverflow.com/q/8550114/939364), one of the StackOverflow questions about Scrapy with the most votes.

Note to self: https://docs.scrapy.org/en/latest/topics/developer-tools.html

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -158,6 +158,7 @@ Solving specific problems
    topics/practices
    topics/broad-crawls
    topics/developer-tools
+   topics/dynamic-content
    topics/leaks
    topics/media-pipeline
    topics/deploy
@@ -182,6 +183,9 @@ Solving specific problems
 
 :doc:`topics/developer-tools`
     Learn how to scrape with your browser's developer tools.
+
+:doc:`topics/dynamic-content`
+    Read webpage data that is loaded dynamically.
 
 :doc:`topics/leaks`
     Learn how to find and get rid of memory leaks in your crawler.
--- a/docs/topics/dynamic-content.rst
+++ b/docs/topics/dynamic-content.rst
@@ -1,0 +1,246 @@
+.. _topics-dynamic-content:
+
+====================================
+Selecting dynamically-loaded content
+====================================
+
+Some webpages show the desired data when you load them in a web browser.
+However, when you download them using Scrapy, you cannot reach the desired data
+using :ref:`selectors <topics-selectors>`.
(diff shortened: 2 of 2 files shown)


dangra:
hi @Gallaecio, this doc topic looks good and ready to merge, isn't it?

Gallaecio:
It is ready by me. But maybe someone else (@kmike?) wants to review it first.

docs/topics/beyond-selectors.rst
```diff
@@ -0,0 +1,239 @@
...
+request with Scrapy.
+
+It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
+method and URL. However, you may also need to reproduce the body, headers and
+form parameters (see :class:`~scrapy.http.FormRequest`) of that request.
+
+Once you get the expected response, you can :ref:`extract the desired data from
+it <topics-handling-response-formats>`.
+
+If you cannot manage to get the expected response,
```
Gallaecio: Actually, it seems I did not finish writing here :flushed:

Gallaecio:
Completed.

kmike:
Hi @Gallaecio! I think it is a much-needed section of the documentation. I've checked it, and it looks good to me, not comments or suggestions :) 

However, I wonder if that's possible to find a better name for this section. "Data which selectors can't reach", "beyond selectors" doesn't sound precise - the problem is not with selectors (and in fact we're suggesting to use them later), the problem is that data is not available in the downloaded HTML, or that data can't be easily extracted by parsing HTML itself, and instead requires parsing some other format embedded in HTML (such as CSS or JS).

Gallaecio:
I must say I’m not happy with the name either. Any suggestions?

Gallaecio:
If we extend the topic to cover the easiest scenario as well, that where selectors work as expected, we could go for a simple, generic title in the lines of “Extracting data” or “Locating and extracting the desired data”.

raphapassini:
I would vote for something in the lines "Selecting dynamically loaded content"

Gallaecio:
Hmm… "Dynamically-loaded" definitely fits everything covered one way or another. I like it.

kmike:
👍 👍

pablohoffman:
great addition to the doc! 💪

PR #3706, opened 2019-03-29
Move documentation from topics/api to docstrings

Gallaecio:


Gallaecio:
There seems to be an issue with zope documentation at the moment: https://bitbucket.org/birkenfeld/sphinx-contrib/issues/208/zopeext-typeerror-object-object-is-not

PR #3707, opened 2019-03-29
Use pytest-xdist

Gallaecio:
In summary:
- Add `pytest-xdist` to `tests/requirements*.txt` and remove versions from `pytest*` packages.
- In the tox configuration, add `-n auto` for all environments running `pytest`.
- Because Windows does not seem to support `pytest-xdist` cleanly, create a new tox environment specifically for Windows that does not use `pytest-xdist` (no `-n auto`).

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -205,6 +205,29 @@ To run a specific test (say ``tests/test_loader.py``) use:
 
     ``tox -- tests/test_loader.py``
 
+To run the tests on a specific tox_ environment, use ``-e <name>`` with an
+environment name from ``tox.ini``. For example, to run the tests with Python
+3.6 use::
+
+    tox -e py36
+
+You can also specify a comma-separated list of environmets, and use `tox’s
+parallel mode`_ to run the tests on multiple environments in parallel::
+
+    tox -e py27,py36 -p auto
+
+To pass command-line options to pytest_, add them after ``--`` in your call to
+tox_. Using ``--`` overrides the default positional arguments defined in
+``tox.ini``, so you must include those default positional arguments
+(``scrapy tests``) after ``--`` as well::
+
+    tox -- scrapy tests -x  # stop after first failure
+
+You can also use the `pytest-xdist`_ plugin. For example, to run all tests on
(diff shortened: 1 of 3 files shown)


kmike:
Hey! I'm surprised tests pass, I was thinking they won't work out of box 👍 

What do you think about adding pytest-xdist to tox, documenting how to use it, but not having `-n auto` by default, or at least allowing to override it? In some cases using all CPUs may be not desirable, as it may use too much RAM, and make tests slower, not faster; also, with pytest-xdist sometimes it is harder to stop tests cleanly. It could also make things slower if you're just running a few tests, selected by `-k`, as setup of these processes may take longer than just running everything in a single process.

Gallaecio:
Then, what about simply installing pytest-xdist and documenting its usage?

docs/contributing.rst
```diff
@@ -205,6 +205,24 @@ To run a specific test (say ``tests/test_loader.py``) use:
...
+tox_. Using ``--`` overrides the default positional arguments defined in
+``tox.ini``, so you must include those default positional arguments
+(``scrapy tests``) after ``--`` as well::
+
+    tox -- scrapy tests -x  # stop after first failure
+
+If you have multiple CPU cores, you can use ``-n <number>`` to use a specific
+number of CPU cores to run the tests, or ``-n auto`` to use all of them::
+
+    tox -- scrapy tests -n auto
```
kmike: I think in this example tests are executed for all environments; there is a different way to parallelize it (https://tox.readthedocs.io/en/latest/example/basic.html#parallel-mode). I'm not sure what is better, to run individual environments with several processes, or to run environments in parallel in this case. What do you think about having e.g. `tox -e py36 -- scrapy tests -n auto` as an example?

kmike:
@Gallaecio Thanks, the PR looks great!

docs/contributing.rst
Gallaecio: You can actually use both in combination, as it turns out, but I guess it’s better to cover them separately, and let developers do what works best for them.

kmike:
Thanks @Gallaecio!

PR #3709, opened 2019-03-31
Enable Auto Completion for pip in Zsh

sathve:
Written in INSTALL for auto . completion in pip


dangra:
This change is not related to Scrapy, it doesn't add Scrapy completion to ZSH just standard PIP completion. Better to suggest such change to PIP project instead. thanks

PR #3712, opened 2019-04-01
[MRG+1] [Docs] CrawlSpider: update Rule docs

### LINKED ISSUE #3711: way to get text inside anchor tag in Crawlspider
I have a crawlspider which crawls given site upto certain dept and download the pdfs on that site. Everything works fine but along with link of pdf, i also need text inside anchor tag. 

for eg:

    <a href='../some/pdf/url/pdfname.pdf'>Project Report</a>

consider this anchor tag, in callback i get response object and along with this object i need text inside that tag for eg 'Project Report'.
Is there any way to get this information along with the response object. i have gone through https://docs.scrapy.org/en/latest/topics/selectors.html link but it not something that i am looking for.

elacuesta:
Fixes #3711

More information at https://stackoverflow.com/q/55450472/939364


docs/topics/spiders.rst
```diff
@@ -377,7 +377,10 @@ Crawling rules
 .. class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None)
 
    ``link_extractor`` is a :ref:`Link Extractor <topics-link-extractors>` object which
-   defines how links will be extracted from each crawled page.
+   defines how links will be extracted from each crawled page. Each produced link will
+   be used to generate a :class:`~scrapy.http.Request` object, which will contain the
+   link's text in its ``meta`` attribute.
+   The link text can be accessed from the callback method though ``response.meta['link_text']``
```
Gallaecio: I wonder if it may make more sense to add this documentation to the description of the ``callback`` parameter below, instead of ``link_extractor``.

elacuesta: On second thought, I removed that particular line, it seemed a bit repetitive.

docs/topics/spiders.rst
```diff
@@ -377,13 +377,16 @@ Crawling rules
...
+   link's text in its ``meta`` attribute.
 
    ``callback`` is a callable or a string (in which case a method from the spider
    object with that name will be used) to be called for each link extracted with
-   the specified link_extractor. This callback receives a response as its first
-   argument and must return a list containing :class:`~scrapy.item.Item` and/or
-   :class:`~scrapy.http.Request` objects (or any subclass of them).
+   the specified link extractor. This callback receives a :class:`~scrapy.http.Response`
+   as its first argument and must return either a single instance or an iterable of
+   :class:`~scrapy.item.Item`, ``dict`` and/or :class:`~scrapy.http.Request` objects
```
kmike: 👍

docs/topics/spiders.rst
```diff
@@ -377,13 +377,16 @@ Crawling rules
 .. class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None)
 
    ``link_extractor`` is a :ref:`Link Extractor <topics-link-extractors>` object which
-   defines how links will be extracted from each crawled page.
+   defines how links will be extracted from each crawled page. Each produced link will
+   be used to generate a :class:`~scrapy.http.Request` object, which will contain the
+   link's text in its ``meta`` attribute.
```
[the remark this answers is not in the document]
suraj-deshmukh: i think it would be great if we mention key name. that would help user to understand.

docs/topics/spiders.rst
```diff
@@ -377,13 +377,16 @@ Crawling rules
...
 
    ``callback`` is a callable or a string (in which case a method from the spider
    object with that name will be used) to be called for each link extracted with
-   the specified link_extractor. This callback receives a response as its first
-   argument and must return a list containing :class:`~scrapy.item.Item` and/or
-   :class:`~scrapy.http.Request` objects (or any subclass of them).
+   the specified link extractor. This callback receives a :class:`~scrapy.http.Response`
+   as its first argument and must return either a single instance or an iterable of
+   :class:`~scrapy.item.Item`, ``dict`` and/or :class:`~scrapy.http.Request` objects
+   (or any subclass of them).
```
Gallaecio: I still wonder if we should mention `link_text` in the documentation about the response, since users will most often need to access that meta field from their callback.

kmike:
I think these changes are improvement to what we have in master, so +1 to merge them as-is.

docs/topics/spiders.rst
elacuesta: I agree, but I was concerned about duplicating the same content. I ended up basically copying the same text from the previous part, but adding "As mentioned above" to make it clear that it's not an accidental duplication but an intended remark.

PR #3714, opened 2019-04-02
Fixed wrong name of attachment when use non-ascii filename

runbing:
If parameter `attachs` of the  function `send` use non-ascii filename like below, the attachment filename can not display correctly.

```attachs=[(u'更新条目.json', 'text/json', open('/tmp/update.json'))]```

This was solved by updated code as I put below:

```part.add_header('Content-Disposition', 'attachment', filename=attach_name)```

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -82,8 +82,7 @@ class MailSender(object):
                 part = MIMEBase(*mimetype.split('/'))
                 part.set_payload(f.read())
                 Encoders.encode_base64(part)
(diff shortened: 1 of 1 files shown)


runbing:
> Could you add a test, to ensure that we do not break this again in the future?
> 
> There is already a test for sending an attachment, I think you would simply need to copy it with a different name and change the filename to something in Unicode.

Thanks for your remind, I create a new new pull request #3722.

Gallaecio:
As you wish.

Gallaecio:
As you wish.

PR #3718, opened 2019-04-02
Deprecate the scrapy.utils.gz.is_gzipped function

Gallaecio:
The function seems to be no longer used.

I suspect its usage has been replaced by the inspecting the actual response bodies, which is more reliable than trusting headers.


kmike:
A great catch, thanks @Gallaecio!

PR #3722, opened 2019-04-04
Fixed UnicodeEncodeError and wrong attachment filename when use non-ascii in MailSender

### LINKED ISSUE #3710: Got "UnicodeEncodeError: 'ascii' codec can't encode characters ..." using MailSender
I want to send a message with MailSender when the spider closed. Here is the code snip:

```
# -*- coding:utf-8 -*-

import scrapy
from scrapy.mail import MailSender

class UpdateSpider(scrapy.Spider):
    name = 'update'

    # ...

    def close(self):
        mailer = MailSender.from_settings(self.settings)
        body = u'测试内容。'
        return mailer.send(to=["runbing@example.com"], subject=u"测试标题",
               body=body, mimetype="text/html")
```

but if the parameter **body**  is non-ascii string, I will get a error like blow.

```
2019-03-31 22:03:59 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method ?.close of <UpdateSpider 'update' at 0x7f4932496650>>
  File "/home/runbing/Code/Scrapy/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/runbing/Code/Scrapy/venv/lib/python2.7/site-packages/pydispatch/robustapply.py", line 55, in robustApply
  File "/home/runbing/Code/Scrapy/amazon/amazon/spiders/update.py", line 81, in close
    mimetype="text/html")
  File "/home/runbing/Code/Scrapy/venv/lib/python2.7/site-packages/scrapy/mail.py", line 101, in send
    dfd = self._sendmail(rcpts, msg.as_string().encode(charset or 'utf-8'))
  File "/usr/lib64/python2.7/email/message.py", line 137, in as_string
    g.flatten(self, unixfrom=unixfrom)
  File "/usr/lib64/python2.7/email/generator.py", line 83, in flatten
    self._write(msg)
  File "/usr/lib64/python2.7/email/generator.py", line 108, in _write
    self._dispatch(msg)
  File "/usr/lib64/python2.7/email/generator.py", line 134, in _dispatch
    meth(msg)
  File "/usr/lib64/python2.7/email/generator.py", line 180, in _handle_text
    self._fp.write(payload)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

I spend all my day to solve this problem, but get no luck, I just find a way like below to hack the problem, but I know this is not recommended. 

```
import sys
reload(sys)
sys.setdefaultencoding('utf-8')
```
Is there any help for this problem?

runbing:
Fixes #3710 

This updated codes solve two problems on using non-ascii in MailSender:

1. Fixed UnicodeEncodeError when the parameter `body` is non-ascii string in python 2.7.X.
2. Fixed displaying wrong attachment filename when the parameter `filename`  is non-ascii string.

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -27,6 +27,8 @@ from twisted.internet import defer, reactor, ssl
 from scrapy.utils.misc import arg_to_iter
 from scrapy.utils.python import to_bytes
 
+from email.header import Header
+
 logger = logging.getLogger(__name__)
 
 
@@ -73,20 +75,19 @@ class MailSender(object):
             rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
 
-        if charset:
-            msg.set_charset(charset)
-
         if attachs:
+            if charset:
+                msg.set_charset(charset)
             msg.attach(MIMEText(body, 'plain', charset or 'us-ascii'))
             for attach_name, mimetype, f in attachs:
                 part = MIMEBase(*mimetype.split('/'))
                 part.set_payload(f.read())
                 Encoders.encode_base64(part)
                 part.add_header('Content-Disposition', 'attachment; filename="%s"' \
-                    % attach_name)
+                    % Header(attach_name).encode())
                 msg.attach(part)
         else:
-            msg.set_payload(body)
+            msg.set_payload(body, charset)
 
         if _callback:
             _callback(to=to, subject=subject, body=body, cc=cc, attach=attachs, msg=msg)
--- a/tests/test_mail.py
+++ b/tests/test_mail.py
@@ -3,6 +3,7 @@
 import unittest
 from io import BytesIO
 from email.charset import Charset
+from email.header import Header
 
 from scrapy.mail import MailSender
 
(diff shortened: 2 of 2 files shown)


scrapy/mail.py
```diff
@@ -27,6 +27,8 @@
 from scrapy.utils.misc import arg_to_iter
 from scrapy.utils.python import to_bytes
 
+from email.header import Header
```
Gallaecio: Please, move this import to the group of imports from the standard library. See [Imports, PEP 8](https://www.python.org/dev/peps/pep-0008/#imports).

scrapy/mail.py
```diff
@@ -73,20 +75,19 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
             rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
 
-        if charset:
-            msg.set_charset(charset)
```
Gallaecio: Why are you not calling `msg.set_charset` anymore when there are no attachments?

PR #3725, opened 2019-04-05
Docs - Item Pipeline - Add filenames to code snippets

float13:
I'm new to Scrapy, and the tutorial is great, but beyond that I got lost very fast.

I think it would be really helpful for all code blocks in all docs to mention the filename that they belong in. 

For example, in the Item Pipeline ```docs/topics/item-pipeline.rst```, I figured out that the very last code snippet goes in the ```settings.py``` file. (I eventually realized there's a link to the doc for settings, but writing it explicitly in addition would be helpful). 

Similarly, for Item Loaders ```docs/topics/loaders.rst``` it's not immediately clear that the Item Loader declaration goes in the ```items.py``` file while the parsing code stays in the spider files.

I'm also new to Sphinx/.rst format, so stylistically, what is a good way to do this? Is it possible to add a caption to the bottom of a code block, with the file name?

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -511,7 +511,7 @@ We can try extracting it in the shell::
     '<a href="/page/2/">Next <span aria-hidden="true">→</span></a>'
 
 This gets the anchor element, but we want the attribute ``href``. For that,
-Scrapy supports a CSS extension that let's you select the attribute contents,
+Scrapy supports a CSS extension that lets you select the attribute contents,
 like this::
 
     >>> response.css('li.next a::attr(href)').get()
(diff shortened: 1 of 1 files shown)


float13:
**NOTE** I think I included commits from my other pull request by accident in this branch/PR - I meant to make this PR empty to start the discussion re: style - how do I remove those commits from this PR?

Gallaecio:
@float13 You did that because you created your new branch from the existing branch that you used for the other pull request.

To be honest, I think you will have to create another pull request.

There are lots of different ways to remove those commits from your new pull request with Git, however the best approach is to create a new branch again, and this time make sure you branch out of *master*.

On the command line, you can create a new branch ensuring it branches from master using:

    git checkout master -b <name>

PR #3726, opened 2019-04-05
[MRG+1] fix: do not catch system exceptions like KeyboardInterrupt

ankostis:
I noticed that occasionally my **Ctrl+C** is ignored.  In general this happens when using plain `except:`. In this case, you probably do not want to catch system-exceptions like `KeyboardInterrupt`.


Gallaecio:
Thanks!

I’m thinking we should probably use `pytest-pylint` to detect this kind of issues. I’ll add that to my personal to-do.

kmike:
Thanks @ankostis!

PR #3727, opened 2019-04-05
Use pylint

Gallaecio:
Has pytest run pylint on Python files.

`pylintrc` is currently skipping all tests that affect the current codebase. Later changes may solve one or more of the corresponding issues and remove them from the list of ignored issues.


dangra:
> `pylintrc` is currently skipping all tests that affect the current codebase. Later changes may solve one or more of the corresponding issues and remove them from the list of ignored issues.

I like the approach although I'd prefer to be able to disable checks only for the files that are not passing them. It was part of the approach we took with tests to port the codebase to Python3.

Gallaecio:
I’ll work on it.

Gallaecio:
Done.

The only exception is a rule that is specific to Python 3 (not subclassing `object`) and should be only applied once Python 2 support is dropped.

Gallaecio:
I’ve configured pytest-pylint to run on Python 3 only. This is because Pylint dropped support for Python 2 a while ago, and most of the issues found on Python 2 were due to Pylint bugs that have been already fixed but will never be backported for Python 2.

Gallaecio:
Updated to resolve conflicts and run in its own continuous integration worked, instead of using pytest-pylint to run in all workers, in line with https://github.com/scrapy/scrapy/pull/3945

Gallaecio:
Updated to skip checks globally again. Unlike [flake8](https://github.com/scrapy/scrapy/pull/3945), pylint does not support ignoring a list of files specifying the checks to ignore per file.

And because most changes will be to existing files, and most files do not pass one of the checks, excluding files instead of checks would not cover many files.

So, the plan is to ignore specific checks globally, and fix checks one by one in later pull requests.

kmike:
thanks @Gallaecio!

PR #3728, opened 2019-04-05
[MRG+1] fix typo on Spider Middleware template (Response -> Request)

victor-torres:
Check [Spider Middleware docs](https://github.com/scrapy/scrapy/blob/b5c552d17ff9e9629434712c3d0595c02853bcfc/docs/topics/spider-middleware.rst) for more information

### DIFF
--- a/scrapy/templates/project/module/middlewares.py.tmpl
+++ b/scrapy/templates/project/module/middlewares.py.tmpl
@@ -39,7 +39,7 @@ class ${ProjectName}SpiderMiddleware(object):
         # Called when a spider or process_spider_input() method
         # (from other spider middleware) raises an exception.
 
-        # Should return either None or an iterable of Response, dict
+        # Should return either None or an iterable of Request, dict
         # or Item objects.
         pass
 


scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -39,7 +39,7 @@ class ${ProjectName}SpiderMiddleware(object):
         # Called when a spider or process_spider_input() method
         # (from other spider middleware) raises an exception.
 
-        # Should return either None or an iterable of Response, dict
+        # Should return either None or an iterable of Request, dict
```
elacuesta: Good catch. Since you're at it, could you maybe add `Item` as well? And perhaps mention that any subclass would work too.

Gallaecio: Item is on the next line (that got me too). And I think the subclass part is implicit. Otherwise, there are lots of places in code and documentation that need updating :)

elacuesta: Ouch, the next line! 🤦‍♂️ 
Nevermind then, sorry for the noise.

Issue #3704, opened 2019-03-27
Set environment variable when running scrapy check


Jesse-Bakker:
Sometimes it is nice to be able to enable/disable functionality, e.g. calculating things in settings.py when just checking spider contracts instead of running a crawl. I therefor propose setting an environment variable like `SCRAPY_CHECK` when using the check command.

PR #3729, opened 2019-04-06
added an --no-check option for check, allowing user only check syntax…

Gravellent:
added an --no-check option for check, allowing user only check syntax


Gallaecio:
@Gravellent Could you share some use cases for this option?

Gravellent:
I was trying to fix the issue for 
https://github.com/scrapy/scrapy/issues/3704
After checking the other PR that tries to solve the issue, maybe I misunderstood the requirement from the ticket.

PR #3730, opened 2019-04-06
[MRG+1] doc: document LOGSTATS_INTERVAL setting

ankostis:
The setting `LOGSTATS_INTERVAL` is mentioned in various samples and issues but is not documented in the main topic.


kmike:
Good catch!

PR #3735, opened 2019-04-09
[MRG+1] Changes a parameter name in class `DummyPolicy`

### LINKED ISSUE #3734: Name parameters uniformly (Cache Policy API)
https://github.com/scrapy/scrapy/blob/92801850546121f2faf95fd21c923904fde35c28/scrapy/extensions/httpcache.py#L34

It will look better if we name the second parameter as `cachedresponse` (in place of just `response`).

In fact, implementation of policy for RFC2616 compliant HTTP cache names it correctly.
https://github.com/scrapy/scrapy/blob/92801850546121f2faf95fd21c923904fde35c28/scrapy/extensions/httpcache.py#L101

I am sending a pull request for the same.

anubhavp28:
Fixes #3734

### DIFF
--- a/scrapy/extensions/httpcache.py
+++ b/scrapy/extensions/httpcache.py
@@ -31,7 +31,7 @@ class DummyPolicy(object):
     def should_cache_response(self, response, request):
         return response.status not in self.ignore_http_codes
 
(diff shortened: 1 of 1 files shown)


kmike:
Thanks @anubhavp28!

PR #3738, opened 2019-04-09
Docs for writing custom cache policy

### LINKED ISSUE #3683: Document how to implement custom HTTP cache storage backed & cache policy.
https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache does let readers know they could implement custom HTTP cache storage backend and cache policy, but provides no information about the API.  I would like to work on documenting it. Should I go ahead?

anubhavp28:
Fixes #3683

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -357,7 +357,7 @@ HttpCacheMiddleware
         * :ref:`httpcache-policy-dummy`
 
     You can change the HTTP cache policy with the :setting:`HTTPCACHE_POLICY`
-    setting. Or you can also implement your own policy.
+    setting. Or you can also :ref:`implement your own policy. <httpcache-policy-custom>`
 
     .. reqmeta:: dont_cache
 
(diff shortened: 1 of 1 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -357,7 +357,7 @@ HttpCacheMiddleware
         * :ref:`httpcache-policy-dummy`
 
     You can change the HTTP cache policy with the :setting:`HTTPCACHE_POLICY`
-    setting. Or you can also implement your own policy.
+    setting. Or you can also :ref:`implement your own policy. <httpcache-policy-custom>`
```

```suggestion
    setting. Or you can also :ref:`implement your own policy <httpcache-policy-custom>`.
```

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,64 @@ In order to use this policy, set:
 * :setting:`HTTPCACHE_POLICY` to ``scrapy.extensions.httpcache.RFC2616Policy``
 
 
+.. _httpcache-policy-custom:
+
+Writing your own policy
+~~~~~~~~~~~~~~~~~~~~~~~
+
+You can implement a cache policy by creating a Python class that
```
Gallaecio: :lipstick:  I think we can omit ‘Python’ here :slightly_smiling_face:

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,64 @@ In order to use this policy, set:
...
+defines the methods described below.
+
+.. module:: scrapy.extensions.httpcache
+
+.. class:: CachePolicy
+
+    .. method:: should_cache_request(request)
+
+      Returns whether the request is cacheable or not
```
Gallaecio: Instead of ‘whether’, I think using ``True`` and ``False`` in the description may be more effective.

Also, see https://github.com/scrapy/scrapy/pull/3671#discussion_r268213806

And I personally feel like in many cases documenting parameters separately is overkill. For example, if the word ‘request’ here links to the API documentation of the `Request` class, I think we could remove the parameter-specific documentation lines below. @kmike What do you think?

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,64 @@ In order to use this policy, set:
...
+      :param response: the response from the server to use for validation
+      :type response: :class:`~scrapy.http.Response` object
+
+      :param request: the request for which the cached response is intended
+      :type request: :class:`~scrapy.http.Request` object
+
+
+In order to use this policy, set:
+
+* :setting:`HTTPCACHE_POLICY` to the Python import path of your custom policy class.
```
Gallaecio: Don’t use single-item lists :slightly_smiling_face:

anubhavp28: I was trying to follow the style used to describe other policies. Single-item lists are also used in description of `Dummy policy` and `RFC2616 policy`, should I modify those too?

Gallaecio: Sorry, I hadn’t notice.

Feel free to leave this as is, consistency is more important. We can deal with this in a separate, future change.

anubhavp28:
@Gallaecio I have made the changes suggested by you. Could you review my changes?

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,40 @@ In order to use this policy, set:
...
+You can implement a cache policy by creating a class that
+defines the methods described below.
+
+.. module:: scrapy.extensions.httpcache
+
+.. class:: CachePolicy
+
+    .. method:: should_cache_request(request)
+
+      Return True if the :class:`~scrapy.http.Request` is cacheable otherwise False
```
Gallaecio: A few :lipstick: aesthetic suggestions for this and the following lines:

```suggestion
      Return ``True`` if the :class:`~scrapy.http.Request` is cacheable or ``False`` otherwise.
```

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,40 @@ In order to use this policy, set:
...
+
+    .. method:: should_cache_response(response, request)
+
+      Return True if :class:`~scrapy.http.Response` should be cached otherwise False
+
+    .. method:: is_cached_response_fresh(cachedresponse, request)
+
+      Return True if cached :class:`~scrapy.http.Response` has expired otherwise False
```
Gallaecio: This seems contradictory with the name of the method. Could you please check that `True` is not returned when the specified cached response has *not* expired?

docs/topics/downloader-middleware.rst
```diff
@@ -429,6 +429,40 @@ In order to use this policy, set:
...
+
+    .. method:: is_cached_response_valid(cachedresponse, response, request)
+
+      Return True if cached :class:`~scrapy.http.Response` is valid otherwise False
```
Gallaecio: It is my understanding that `is_cached_response_valid` receives a new response and the previously-cached version of that same response, and you must return `True` if the cached response should be used instead of the fresh response, or `False` if the new response should be used and added to the cache, overriding the previously-cached response.

I think the description should make this clear.

Gallaecio (changes requested):

AdrianAtZyte:
(already covered nowadays)

PR #3739, opened 2019-04-10
[MRG+1] Add SCRAPY_CHECK environment variable

### LINKED ISSUE #3704: Set environment variable when running scrapy check

Matthijsy:
This PR implements the environment variable when running `scrapy check` as asked in #3704 . This way it is possible to have different behaviour in the scraper when running the check (like requiring less settings to be set)

fixes #3704

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -156,7 +156,7 @@ def set_environ(**kwargs):
     try:
         yield
     finally:
-        for k, v in original_env:
+        for k, v in original_env.items():
             if v is None:
                 del os.environ[k]
             else:
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -1,6 +1,8 @@
 """Helper functions which don't fit anywhere else"""
+import os
 import re
 import hashlib
+from contextlib import contextmanager
 from importlib import import_module
 from pkgutil import iter_modules
 
@@ -9,7 +11,6 @@ from w3lib.html import replace_entities
 
 from scrapy.utils.python import flatten, to_unicode
 from scrapy.item import BaseItem
-
 
 _ITERABLE_SINGLE_VALUES = dict, BaseItem, six.text_type, bytes
 
@@ -40,7 +41,7 @@ def load_object(path):
     except ValueError:
         raise ValueError("Error loading object '%s': not a full path" % path)
 
-    module, name = path[:dot], path[dot+1:]
+    module, name = path[:dot], path[dot + 1:]
     mod = import_module(module)
 
     try:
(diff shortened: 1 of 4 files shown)


Gallaecio:
Could you please add an automated test to cover this change?

Matthijsy:
I've added a test for the `set_environ` util. I couldn't find the test of scrapy check so don't know how I should write a test for that part. Is this untested or can you tell me where I can find it?

Gallaecio:
`scrapy/tests/test_cmdline/__init__.py` tests some commands, but there are no tests for `scrapy check` currently, no, so I’m OK with your test coverage.

It would be great if you could cover the new environment variable in the documentation about `scrapy check`. Other than that, I think this is ready to merge.

Matthijsy:
@Gallaecio Thanks for your feedback! I've added the documentation, hope this is the right way to document it and otherwise I would love to update the documentation.

scrapy/utils/misc.py
```diff
@@ -1,6 +1,8 @@
 """Helper functions which don't fit anywhere else"""
+import os
 import re
 import hashlib
+from contextlib import contextmanager
```
Gallaecio: :lipstick: Please, keep the alphabetical order.

scrapy/utils/misc.py
```diff
@@ -10,7 +12,6 @@
 from scrapy.utils.python import flatten, to_unicode
 from scrapy.item import BaseItem
 
-
```
Gallaecio: :lipstick: Please, keep those 2 lines between imports and the rest of the code.

tests/test_utils_misc/__init__.py
```diff
@@ -3,7 +3,7 @@
 import unittest
 
 from scrapy.item import Item, Field
-from scrapy.utils.misc import arg_to_iter, create_instance, load_object, walk_modules
+from scrapy.utils.misc import arg_to_iter, create_instance, load_object, walk_modules, set_environ
```
Gallaecio: :lipstick: Alphabetical order, please :slightly_smiling_face:

scrapy/utils/misc.py
Matthijsy: Is this the wanted order? I don't know which alphabetic order should be aplied since re and hashlib were also not in alphabetic order. I know made it start with the normal import in alphabetic order followed by the from imports in alphabetic order

Gallaecio: I see, I did not realize it was common to have `from … import` lines after regular `import` lines. Ignore my comment :slightly_smiling_face:

Matthijsy:
Thanks for the comment on how to build the documentation! I builded it and think it looks fine now

Gallaecio:
I’ve refactored your documentation a bit. Please, review my changes.

Matthijsy:
Great! Looks fine to me

PR #3741, opened 2019-04-10
add TextResponse.re() and .re_first()

apalala:
The `.re()` and `.re_first()` methods were left out of `TextResponse` when `.xpath()` and `.css()`  were added in #690. The reasons for leaving them out at the last minute (they were in the original pull request) are undocumented.

This pull requests adds the two methods.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -5,6 +5,11 @@ Release notes
 
 .. _release-1.6.0:
 
+Scrapy 1.6.1 (2019-MM-DD)
+-------------------------
+
+*   `TextResponse.re()` and `TextResponse.re_first()` added
+
 Scrapy 1.6.0 (2019-01-30)
 -------------------------
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -530,7 +530,7 @@ dealing with JSON requests.
 
    :param data: is any JSON serializable object that needs to be JSON encoded and assigned to body.
       if :attr:`Request.body` argument is provided this parameter will be ignored.
-      if :attr:`Request.body` argument is not provided and data argument is provided :attr:`Request.method` will be 
+      if :attr:`Request.body` argument is not provided and data argument is provided :attr:`Request.method` will be
       set to ``'POST'`` automatically.
    :type data: JSON serializable object
 
@@ -761,6 +761,18 @@ TextResponse objects
         A shortcut to ``TextResponse.selector.css(query)``::
 
             response.css('p')
+
+    .. method:: TextResponse.re(self, regex, replace_entities=True)
+
+        A shortcut to ``TextResponse.selector.re(regex, replace_entities=True)``::
+
+            response.re(r'-(\w+)')
+
+    .. method:: TextResponse.re_first(self, regex, default=None, replace_entities=True)
+
+        A shortcut to ``TextResponse.selector.re_first(regex, default=None, replace_entities=True)``::
+
+            response.re_first(r'-(\w+)', default='')
 
     .. automethod:: TextResponse.follow
 
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -69,9 +69,10 @@ constructed by passing either :class:`~scrapy.http.TextResponse` object or
 markup as an unicode string (in ``text`` argument).
 Usually there is no need to construct Scrapy selectors manually:
 ``response`` object is available in Spider callbacks, so in most cases
-it is more convenient to use ``response.css()`` and ``response.xpath()``
-shortcuts. By using ``response.selector`` or one of these shortcuts
-you can also ensure the response body is parsed only once.
+it is more convenient to use ``response.css()``, ``response.xpath()``,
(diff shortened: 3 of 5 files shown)


docs/news.rst
```diff
@@ -5,6 +5,11 @@ Release notes
 
 .. _release-1.6.0:
```
stav: This should be:

    .. _release-1.6.1:

docs/news.rst
```diff
@@ -5,6 +5,11 @@ Release notes
...
 
+Scrapy 1.6.1 (2019-MM-DD)
+-------------------------
+
+*   `TextResponse.re()` and `TextResponse.re_first()` added
+
 Scrapy 1.6.0 (2019-01-30)
```
stav: This should be preceded by:

    .. _release-1.6.0:

docs/news.rst
```diff
@@ -5,6 +5,11 @@ Release notes
 
 .. _release-1.6.0:
 
+Scrapy 1.6.1 (2019-MM-DD)
```
kmike: This is unlikely to be in 1.6.1: first, it is a new feature (which may be not suited for a point release), and second, we don't know if we'll have 1.6.1, or go straight into 1.7. 

Could you please remove a specific version here? You can either replace it with "Unreleased (TBA)", or just remove changelog changes from this pull request.

kmike:
Let me try recalling why we deferred this shortcut :) I think one of the reasons could be inefficiency. This is what happens when you call `TextResponse.selector.re(..)`: 

1. body is parsed to lxml tree if it is not parsed yet (.selector is lazy)
2. `etree.tostring` is called on the resulting tree. In some cases you may get "1" string constant or "0" string constant (see https://github.com/scrapy/parsel/blob/e01093cf6342c90445028de28034b3cc3d2ead8b/parsel/selector.py#L297).
3. [extract_regex](https://github.com/scrapy/parsel/blob/e01093cf6342c90445028de28034b3cc3d2ead8b/parsel/utils.py#L59) is called on the resulting string. It includes (optional, ON by default) replacing entities like `&amp;` in the text string, using [replace_entities](https://github.com/scrapy/w3lib/blob/0108fab2f627b2bfdf3d6942a44ed10a78ff443f/w3lib/html.py#L41) function.

Because of (1) and (2), the tree can be changed. Lxml can insert e.g. missing closing tags, or reorder some tags. I'm not sure, but order of html attributes can change as well. Regex can't be written properly by looking at the page source, because it works on serialized => deserialized version.

I'd expect response.re work in a diffrent way - call `re` functions directly, maybe call `extract_regex` directly. HTML parsing / converting back to a string looks unnecessary, slow and prone to surprises.

docs/news.rst
apalala: Restored.

apalala:
@kmike Are you suggesting that is best to leave the asymmetry in the signature and have users do `re.findall(pattern, response.text)`?

Without the added methods users will be inclined to do things like`response.xpath('.//text()').re(r'IncidentId:(\w+)')`, which has all the problems you mention, just because it gets the job done.

And it seems that the problems you mention occur with any use of `.re()` if the previous selector is not carefully crafted.

An option is to document the above use case, or, better, change the implementation of the methods to the likes of:

```
''.join(r.xpath('.//text()').getall()).re(pattern, ...)
```

apalala:
Also, the usage pattern is already available, just asymmetrically not exposed. Users can currently do:

```
response.selector.re_first('Thank You (\w+)')
```

And keep wondering why `response.re()` is not available.

kmike:
I think having .re shortcut is fine, but we need to get both semantics and implementation right.

Top-level response.re which calls response.selector.re is inefficient and incomplete - it does unneeded work, has gotchas and doesn't support scrapy.Response (which doesn't have .selector). Selector.re has no other choice, but for the main response `response.text` is available. 

We shouldn't promote arguable a bad practice of running `response.selector.re(...)` - there may be use cases for it (not sure - something about escaping?), but if user wants to find something in HTML as-is, it should be `re.findall(pattern, response.text)` - faster, less gotchas.

There is a way to implement `.re()` to provide the same features as selector.re, but without calling etree.tostring - use [extract_regex](https://github.com/scrapy/parsel/blob/e01093cf6342c90445028de28034b3cc3d2ead8b/parsel/utils.py#L59) directly on response.text. Maybe this is how the shortcut should be implemented. Though it is basically re.findall, with some tweaks; I'm not sure I understand the replace_entities part properly.

Only allowing to run regex on extracted text looks rather limited. For example, one may want to find particular attribute values using .re, or e.g. extract everything url-like, to follow it, even if these links are inside JS code or comments. I think Response.re shouldn't run only on text, because this way it would be incompatible with response.selector.re, which can be more confusing than not having this method at all.

> ''.join(r.xpath('.//text()').getall()).re(pattern, ...)

I kind-of agree that in many cases users may want to run .re just on the extracted text. But text extraction using `' '.join` of `//text()` or `//string()` has many gotchas; there is a pull request to parsel for cleaner text extraction: https://github.com/scrapy/parsel/pull/127.

PR #3750, opened 2019-04-18
Check command verbose

Gravellent:
Scrapy check has been a great tool for me to check how my spider works, however, this unittest tool doesn't provide too much useful information when it's running. Specifically, even if the --verbose option is on, it only tells you if each individual test works, rather than the actual result of the test. I think it would be better if the check command can print out more useful info for the users. 

In this PR, I added some information when -v is on. It prints out the spider and function to be checked, and tells the user when a test passed, how the test works (how many requests/ what kind of fields have been found). I think this can help users to debug a spider.

### DIFF
--- a/scrapy/commands/check.py
+++ b/scrapy/commands/check.py
@@ -59,7 +59,7 @@ class Command(ScrapyCommand):
     def run(self, args, opts):
         # load contracts
         contracts = build_component_list(self.settings.getwithbase('SPIDER_CONTRACTS'))
-        conman = ContractsManager(load_object(c) for c in contracts)
+        conman = ContractsManager([load_object(c) for c in contracts], opts.verbose)
         runner = TextTestRunner(verbosity=2 if opts.verbose else 1)
         result = TextTestResult(runner.stream, runner.descriptions, runner.verbosity)
 
--- a/scrapy/commands/check.py
+++ b/scrapy/commands/check.py
@@ -73,6 +73,7 @@ class Command(ScrapyCommand):
             spidercls.start_requests = lambda s: conman.from_spider(s, result)
 
             tested_methods = conman.tested_methods_from_spidercls(spidercls)
+            print('Checking spider %s...' % spidername)
             if opts.list:
                 for method in tested_methods:
                     contract_reqs[spidercls.name].append(method)
--- a/scrapy/contracts/__init__.py
+++ b/scrapy/contracts/__init__.py
@@ -40,9 +40,11 @@ class ContractsManager(object):
 
     def from_spider(self, spider, results):
         requests = []
+        print('Testing Methods:')
         for method in self.tested_methods_from_spidercls(type(spider)):
             bound_method = spider.__getattribute__(method)
             try:
+                print('-- %s' % bound_method.__name__)
                 requests.append(self.from_method(bound_method, results))
             except Exception:
                 case = _create_testcase(bound_method, 'contract')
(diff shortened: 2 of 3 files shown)


scrapy/commands/check.py
```diff
@@ -59,7 +59,7 @@ def add_options(self, parser):
     def run(self, args, opts):
         # load contracts
         contracts = build_component_list(self.settings.getwithbase('SPIDER_CONTRACTS'))
-        conman = ContractsManager(load_object(c) for c in contracts)
+        conman = ContractsManager([load_object(c) for c in contracts], opts.verbose)
```
Gallaecio: Any reason to turn the generator into a list?

scrapy/commands/check.py
```diff
@@ -73,6 +73,7 @@ def run(self, args, opts):
             spidercls.start_requests = lambda s: conman.from_spider(s, result)
 
             tested_methods = conman.tested_methods_from_spidercls(spidercls)
+            print('Checking spider %s...' % spidername)
```
Gallaecio: How is this different from the `print(spider)` at line 87/88, where not only spiders are printed but also their methods (`print('  * %s' % method)`)?

Gallaecio (commented):
I see that, with the proposed verbose mode, extra information is printed about passing tests.

I’m used to verbose modes in test frameworks (e.g. pytest) providing additional details about failed tests, not about passed tests.

I feel printing additional information about passing tests could be a bit counterproductive: the more information you get about working tests, the harder it is to read information about failing tests. I would expect a verbose mode in `scrapy check` to work closer to how pytest works.

@kmike What’s your take?

scrapy/commands/check.py
Gravellent: I was trying to keep it as a generator, but then it would have a syntax error once I add another argument(opts.verbose) to it. This is the easiest solution I have. Do you know if there is a better way to do this?

scrapy/commands/check.py
Gravellent: > I see that, with the proposed verbose mode, extra information is printed about passing tests.
> 
> I’m used to verbose modes in test frameworks (e.g. pytest) providing additional details about failed tests, not about passed tests.
> 
> I feel printing additional information about passing tests could be a bit counterproductive: the more information you get about working tests, the harder it is to read information about failing tests. I would expect a verbose mode in `scrapy check` to work closer to how pytest works.
> 
> @kmike What’s your take?

I guess I'm looking at this from more of a beginner's point of view... I don't have much experience on pytest, but my idea was that if a user turns "verbose" on, he/she should be able to get more comprehensive information on how the crawler functioned. 

I can think of a couple of use cases of this proposed change:
1. When I try to write the check, I need to understand define a range, say 5 to 30 requests. If the website is somewhat dynamic, it might return a different number of requests each time. In this case, I think a verbose option should allow the user to know exactly how many requests they got. 
2. Check function should be a helpful tool for another developer to understand the crawler. If check can return more information, i.e., the fields that the crawler actually finds, it may help them to read the code faster. Of course they can read the source code as well, but scrapy check --verbose can act as a "minimal code" for them to see how things work.

Gravellent: > How is this different from the `print(spider)` at line 87/88, where not only spiders are printed but also their methods (`print(' * %s' % method)`)?

 I guess this is not necessary, but if the crawler runs really slow for some reason, if you do the scrapy check command, then nothing shows up on the screen for the first 20 seconds... It might be nice for a user to see that the check command actually started.

PR #3753, opened 2019-04-22
Added pathlib.Path support for FEED_URI

### LINKED ISSUE #3731: Support pathlib.Path in FEED_URI
Make things work the same when the value assigned to the `FEED_URI` setting is a string containing a path or an instance of `pathlib.Path`.

shreyashah1903:
Fixes #3731 

[ x ] - Check if uri is instance of pathlib.Path
[ x ] - Resolve the path.

### DIFF
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -3,7 +3,7 @@ lxml
 pyOpenSSL
 cssselect>=0.9
 queuelib
-pathlib
+pathlib2
 w3lib>=1.17.0
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -3,6 +3,7 @@ lxml
 pyOpenSSL
 cssselect>=0.9
 queuelib
+pathlib
 w3lib>=1.17.0
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -3,7 +3,6 @@ lxml
 pyOpenSSL
 cssselect>=0.9
 queuelib
-pathlib2
 w3lib>=1.17.0
 six>=1.5.2
 PyDispatcher>=2.0.5
--- a/requirements-py3.txt
+++ b/requirements-py3.txt
@@ -3,7 +3,6 @@ lxml>=3.2.4
 pyOpenSSL>=0.13.1
 cssselect>=0.9
 queuelib>=1.1.1
-pathlib>=1.0.1
 w3lib>=1.17.0
 six>=1.5.2
 PyDispatcher>=2.0.5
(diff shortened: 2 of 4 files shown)


abhinavsagar:
Don't think resolve should be used here. Rather just change it to `pathlib.Paths` which will also make the build pass. Let me know what you think.

requirements-py3.txt
```diff
@@ -3,6 +3,7 @@ lxml>=3.2.4
 pyOpenSSL>=0.13.1
 cssselect>=0.9
 queuelib>=1.1.1
+pathlib>=1.0.1
```
Gallaecio: In Python 3, pathlib is part of the standard library: https://docs.python.org/3/library/pathlib.html

setup.py
```diff
@@ -68,6 +68,7 @@ def has_environment_marker_platform_impl_support():
         'Twisted>=13.1.0',
         'w3lib>=1.17.0',
         'queuelib',
+        'pathlib',
```
Gallaecio: No reason to modify Scrapy’s dependencies for this.

scrapy/extensions/feedexport.py
```diff
@@ -79,7 +80,10 @@ def store(self, file):
 class FileFeedStorage(object):
 
     def __init__(self, uri):
-        self.path = file_uri_to_path(uri)
+        if isinstance(uri, Path):
+            self.path = uri.resolve()
+        else:
+            self.path = file_uri_to_path(uri)
```
Gallaecio: I think this would be easier to fix on line 191, where `settings['FEED_URI']` is read. I suspect casting settings['FEED_URI'] to a string if it is not `None` should be enough.

scrapy/extensions/feedexport.py
```diff
@@ -191,6 +191,8 @@ def __init__(self, settings):
         self.urifmt = settings['FEED_URI']
         if not self.urifmt:
             raise NotConfigured
+        else:
```
Gallaecio: You can remove this `else` line.

scrapy/extensions/feedexport.py
```diff
@@ -191,6 +191,8 @@ def __init__(self, settings):
         self.urifmt = settings['FEED_URI']
         if not self.urifmt:
```
Gallaecio: I would avoid repeating both `settings['FEED_URI']` and `self.urifmt` in these lines.

You could probably switch to:

```python
if not settings['FEED_URI']:
    raise NotConfigured
self.urifmt = str(settings['FEED_URI'])
```

requirements-py2.txt
```diff
@@ -3,6 +3,7 @@ lxml
 pyOpenSSL
 cssselect>=0.9
 queuelib
+pathlib2
```
Gallaecio: This line should not be necessary anymore.

Gallaecio:
Superseded by #4074

PR #3760, opened 2019-04-27
Fix the link extactor

lhenslee:
I made a function called LinkExtractor in the linkextractors directory that takes a url and returns a list of links that begin with https.


kmike:
Hey @lhenslee! It seems you've submitted wrong changes, so I'm closing this PR.

PR #3763, opened 2019-05-03
Fix itemloader cannot process zero

houbaron:
When we use extend custom input processors, if we input a zero value, it will output only a `[]`, like this:
```Python
from scrapy.loader.processors import TakeFirst, Compose


class TestItem(Item):
    name = Field()


class TestItemLoader(ItemLoader):
    default_item_class = TestItem
    name_in = Compose(TakeFirst(), int)


il = TestItemLoader()
il.add_value('name', u'0')
assert il.get_output_value('name') == []
assert il.get_output_value('name') == [0]
```
And we will get:
```
  File "xxx.py", line 18, in <module>
    assert il.get_output_value('name') == [0]
AssertionError
```

### DIFF
--- a/scrapy/loader/__init__.py
+++ b/scrapy/loader/__init__.py
@@ -89,7 +89,7 @@ class ItemLoader(object):
     def _add_value(self, field_name, value):
         value = arg_to_iter(value)
         processed_value = self._process_input_value(field_name, value)
-        if processed_value:
+        if processed_value or processed_value == 0:
             self._values[field_name] += arg_to_iter(processed_value)
 
     def _replace_value(self, field_name, value):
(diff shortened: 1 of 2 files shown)


Gallaecio:
This change is backward incompatible and has the potential to break user expectations.

Is there a reason why you can’t use `name_out` for converting the value into an integer? For example:

```python
from scrapy.loader.processors import Compose, MapCompose, TakeFirst


class TestItem(Item):
    name = Field()


class TestItemLoader(ItemLoader):
    default_item_class = TestItem
    name_in = Compose(TakeFirst())
    name_out = MapCompose(int)


il = TestItemLoader()
il.add_value('name', u'0')
assert il.get_output_value('name') == [0]
```

PR #3768, opened 2019-05-08
Implement the METAREFRESH_IGNORE_TAGS setting

### LINKED ISSUE #1422: Html http-equiv Tag with status code 200
Dear Scrapy developer,

I crawl following url : 

"http://www.baidu.com/link?url=trNkDflaMpWNZbbB5OphH_bi_8jMpse4zyjTFrexyYw-XAg9M9VmS6YfvsEdUaJEbHKMKfdoPu1LUVMynFNXv1DVVRE6vnmYvCZ1s4jkJae&wd=&eqid=ac8e85b700005ad90000000455c83eee"

and response with status code 200 and content as follow:

```
"<script>window.location.replace("http://testrepository.readthedocs.org/en/latest/MANUAL.html")</script>\n<noscript><META http-equiv="refresh" content="0;URL=\'http://testrepository.readthedocs.org/en/latest/MANUAL.html\'"></noscript>\n"

```

It's won't auto redirect to correct url, why? Maybe it's not scrapy bug, but is there any solution to solve this?

Very thanksful.

Best wishes.

Gallaecio:
Implements a new setting that can be used to fix #1422.

I actually believe that the default value should be an empty list, so that we follow an approach more similar to that of web browsers, and #1422 gets fixed automatically without requiring changing a setting. But these changes aim for backward compatibility.


kmike:
Thanks @Gallaecio! 

Should we keep #1422 open, as a place of discussion for a default, or should we create a separate ticket?

Gallaecio:
> Should we keep #1422 open, as a place of discussion for a default, or should we create a separate ticket?

I’ve just created #3844 for it.

kmike:
Thanks @Gallaecio!

PR #3770, opened 2019-05-10
add sticky meta spider middleware

### LINKED ISSUE #3645: Sticky request/response meta
## Problem 

There should be a default behaviour for enabling sticky meta for all spider requests.  

Sticky meta means transfer Response.meta to Request.meta automatically without having to define that explicitly in every Request object.

## Problem Example

Example for meta carry-over:

    class MySpider(Spider):
        order = [1,2,3]
        def start_request(self):
            for order in self.orders:
                yield Request('url',
                              self.parse_order1,
                              meta={'order': order}
                              )

        def parse_order1(self, response):
            yield Request(
                'url',
                self.parse_order2,
                meta={'order': response.meta['order']}
            #     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            )

        def parse_order2(self, response):
            yield {'item': 'foo', 'order': response.meta['order']}

The intermediate `parse_order1` callback has to explicitly state that meta attribute `order` should be carried over. This gets very awkward if the request chain is several callbacks long or even worse - several callbacks and several keys long.  

## Solution 

One solution would be enable sticky carry over spider middleware that manages that and automatically carries over any predefined keys:

    class StickyMeta:
            sticky_meta = getattr(spider, 'sticky_meta', [])
            if not sticky_meta:
                yield from result
                return
            for value in result:
                if not issubclass(type(value), Request):
                    yield value
                    continue
                for k, v in response.meta.items():
                    if k in sticky_meta and k not in value.meta:
                        value.meta[k] = v
                yield value

    class MySpider(Spider):
        order = [1, 2, 3]
        sticky_meta = ['order']

edit: :point_up: more detailed implementation in referenced PR :point_down:

Granitosaurus:
This change adds sticky meta spider middleware for enabling sticky meta behaviour as described in issue https://github.com/scrapy/scrapy/issues/3645

Fixes #3645

```
StickyMeta middleware transfers specified meta keys to further
requests in spider request chain.

Sticky meta keys can be configured either by:
* Setting STICKY_META_KEYS eg:
    STICKY_META_KEYS = ['key1', 'key2']
* Spider attribute sticky_meta eg:
    spider.sticky_meta = ['key1','key2']
* Meta attribute "sticky" eg:
    Request(meta={'sticky': ['foo'], 'foo': 'bar'})
resolve priority is meta > spider attribute > setting
```

### DIFF
--- a/scrapy/spidermiddlewares/stickymeta.py
+++ b/scrapy/spidermiddlewares/stickymeta.py
@@ -1,0 +1,45 @@
+"""
+StickyMeta middleware transfers specified meta keys to further
+requests in spider request chain.
+
+Sticky meta keys can be configured either by:
+* Setting STICKY_META_KEYS eg:
+    STICKY_META_KEYS = ['key1', 'key2']
+* Spider attribute sticky_meta eg:
+    spider.sticky_meta = ['key1','key2']
+* Meta attribute "sticky" eg:
+    Request(meta={'sticky': ['foo'], 'foo': 'bar'})
+resolve priority is meta > spider attribute > setting
+"""
+from scrapy import Request
+
+
+class StickyMeta:
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.settings)
+
+    def __init__(self, settings):
+        self.settings_sticky_meta = settings.getlist('STICKY_META')
+
+    def process_spider_output(self, response, result, spider):
+        # without attached request response will not have meta to pass on
+        if not response.request:
+            for value in result:
+                yield value
+            return
+
+        spider_sticky_meta = getattr(spider, 'sticky_meta', None)
+        for request in result:
+            if not isinstance(request, Request):
+                yield request
+                continue
+            # priority: meta > spider attribute > setting
+            sticky = response.meta.get(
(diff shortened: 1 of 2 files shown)


scrapy/spidermiddlewares/stickymeta.py
```diff
@@ -0,0 +1,45 @@
...
+                yield value
+            return
+
+        spider_sticky_meta = getattr(spider, 'sticky_meta', None)
+        for request in result:
+            if not isinstance(request, Request):
+                yield request
+                continue
+            # priority: meta > spider attribute > setting
+            sticky = response.meta.get(
```
Gallaecio: Shouldn’t this be out of the loop, just as `spider_sticky_meta = …`?

Gallaecio (commented):
Nice!

ejulio:
Here is an implementation https://github.com/heylouiz/scrapy-sticky-meta-params

felipeboffnunes:
New PR for the topic #5852 based on @heylouiz [repo](https://github.com/heylouiz/scrapy-sticky-meta-params)

PR #3772, opened 2019-05-10
Fix #3771: Add fallback css and xpath selectors

### LINKED ISSUE #3771: Add fallback selectors to ItemLoader
In some cases it is common to have fallback selectors for certain fields.
This way, we end up writing a piece of code like
```
loader = MyLoader(response=response)
loader.add_css('my_field', 'selector1')
loader.add_css('my_field', 'selector2') # fallback 1
loader.add_css('my_field', 'selector3') # fallback 2
```

However, a, maybe, better way would be
```
loader = MyLoader(response=response)
loader.add_css('my_field', [
    'selector1',
    'selector2', # fallback 1
    'selector3', # fallback 2
])
```

The API above would be the equivalent of the first example.
However, @cathalgarvey also shared a nice idea to stop in the first matching selector.
```
loader = MyLoader(response=response)
loader.add_css('my_field', [
    'selector1',
    'selector2', # fallback 1
    'selector3', # fallback 2
], selectors_as_preferences=True)
```
Then, if `selector1` yields a result, the other ones are attempted, otherwise we fallback to `selector2` and so on.

The same API should be applied to `loader.add_xpath`.

ejulio:
Fixes #3771 

So, scrapy already handled `loader.add_css('my_field', ['selector1', 'selector2'])`.
Though, I added this behavior to the docs.

I added a new _kwarg_ called `stop_on_first_match` to both `add_css` and `add_xpath`.
This way we can control if the field is a composite of various selectors or if it is a set of fallbacks.

I'm not sure about the name `stop_on_first_match`.
`processor.Compose` has a `stop_on_none` argument (https://docs.scrapy.org/en/latest/topics/loaders.html#scrapy.loader.processors.Compose).
Maybe we can use another that is similar, or even `stop_on_none` it is checking explicitly for `None` right now.

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -391,6 +391,8 @@ ItemLoader objects
             loader.add_xpath('name', '//p[@class="product-name"]')
             # HTML snippet: <p id="price">the price is $1200</p>
             loader.add_xpath('price', '//p[@id="price"]', re='the price is (.*)')
+            # HTML snippet: <p class="product-name1">Color</p><p class="product-name2">TV</p>
+            loader.add_xpath('name', ['//p[@class="product-name1"]', '//p[@class="product-name2"]'])
 
     .. method:: replace_xpath(field_name, xpath, \*processors, \**kwargs)
 
@@ -434,6 +436,8 @@ ItemLoader objects
             loader.add_css('name', 'p.product-name')
             # HTML snippet: <p id="price">the price is $1200</p>
             loader.add_css('price', 'p#price', re='the price is (.*)')
+            # HTML snippet: <p class="product-name1">Color</p><p class="product-name2">TV</p>
+            loader.add_css('name', ['p.product-name1', 'p.product-name2'])
 
     .. method:: replace_css(field_name, css, \*processors, \**kwargs)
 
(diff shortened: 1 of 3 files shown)


docs/topics/loaders.rst
```diff
@@ -385,12 +385,26 @@ ItemLoader objects
         :param xpath: the XPath to extract data from
         :type xpath: str
 
+        :param stop_on_first_match: when ``xpath`` is a list of selectors, should it
```
Gallaecio: I wonder whether or not this parameter should be added, since I believe the same effect can be achieved already with `TakeFirst`.

Or are you adding it for performance reasons?

ejulio: Truth be told, I forgot that we could get the same behavior with `TakeFirst`.
Not sure if performance is a problem at all.

My concern was related to semantics.

`loader.add_css('field', ['sel1', 'sel2'])` gives the idea of a composition, `field` is a composition of both `'sel1'` and `'sel2'`.

`loader.add_css('field', ['sel1', 'sel2'], stop_on_first_match=True)` would mean they are fallbacks because of different page layouts.

Issue #3752, opened 2019-04-20
pip install scrapy


gbell27:
You need first "pip install incremental" due to a misspelling in the lines for the installation of dependencies.

noahshpak:
I don't have this problem in python3.6

gbell27:
I first tried on 2.7 but went the same on 3.6 using the cached version of the package which wasn’t correctly installed. I saw that it was a common error but I solved it differently from the solutions written by someone else on stackoverflow and others. The error was python setup.py egg_info failed with error 1 described in github.com/palantir/python-language-server/issues/370

Gallaecio:
@gbell27 May this be a duplicate of #2808?

Gallaecio:
Closing due to a lack of feedback.

Issue #2810, opened 2017-07-02
Issue with Scrapy Install


tsgoud69:
Hi,
I am new to python and trying to install scrapy in my virtual env. But its fails with following error. Can any one susggest what is the issue?

python --version
Python 2.7.13

 virtualenv --version
15.1.0

 pip install scrapy
Collecting scrapy
  Using cached Scrapy-1.4.0-py2.py3-none-any.whl
Collecting service-identity (from scrapy)
  Using cached service_identity-17.0.0-py2.py3-none-any.whl
Collecting parsel>=1.1 (from scrapy)
  Downloading parsel-1.2.0-py2.py3-none-any.whl
Requirement already satisfied: six>=1.5.2 in /home/swaraj/E2/VenvP27E1/lib/python2.7/site-packages (from scrapy)
Collecting w3lib>=1.17.0 (from scrapy)
  Downloading w3lib-1.17.0-py2.py3-none-any.whl
Collecting lxml (from scrapy)
  Downloading lxml-3.8.0-cp27-cp27m-manylinux1_x86_64.whl (6.8MB)
    100% |████████████████████████████████| 6.8MB 172kB/s 
Collecting Twisted>=13.1.0 (from scrapy)
  Could not find a version that satisfies the requirement Twisted>=13.1.0 (from scrapy) (from versions: )
No matching distribution found for Twisted>=13.1.0 (from scrapy)

kmike:
It looks like a manifestation of http://twistedmatrix.com/trac/ticket/9046. 

You can e.g. download twisted.tar.bz2 file from https://pypi.python.org/pypi/Twisted, unpack it and install from a local folder.

redapple:
@tsgoud69 , did you manage to install scapy?

iahmed16:
@kmike I installed it manually as you said, but I have requirements.txt with klein==17.10.0 ... so I get the same error as klein depends on Twisted.

and at the same time I downloaded Twisted 18.7.0 and installed successfully.

how to get rid of this error

PR #3789, opened 2019-05-23
Potentially fix issues like #2810, #3752

claudred:
An install script that tries to install items for users at least with osx; tries to install homebrew, python, a virtual env, then install python 3 in the virtual env, then install scrapy in the virtual env.


Gallaecio:
I’m quite conflicted about this.

I really think proper documentation is the way to go here. I think some of the steps that the script automates should actually be done manually by users, because they are steps that need to be followed consciously.

Beyond that, the script in its current form seems rather raw to me:
1. The script runs all its statements regardless of whether previous statements succeed or fail.
2. It changes the working directory to the home folder without going back to the original working directory.
3. It evaluates a Bash script downloaded remotely, and I’m not sure how it behaves if the download fails for some reason.
…

This could also probe to be a maintenance nightmare.

PR #3790, opened 2019-05-24
[MRG+1] Account for mangling when serializing requests with private callbacks

andrewbaxter:
Sorry for making a PR without an issue~ just hit an issue where mixins with similarly named methods used private names to prevent method conflicts.  The spider worked properly in memory but had errors when requests were serialized to a disk queue.

The Python documentation specifically documents the mangling convention so I think it's reasonable to assume it works in other implementations, but I haven't tested except with CPython.

### DIFF
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -2,14 +2,10 @@
 Helper functions for serializing (and deserializing) requests.
 """
 import six
-import re
 
 from scrapy.http import Request
 from scrapy.utils.python import to_unicode, to_native_str
 from scrapy.utils.misc import load_object
-
-
-private_name_regex = re.compile('^__.*[^_]_?$')
 
 
 def request_to_dict(request, spider=None):
@@ -71,6 +67,10 @@ def request_from_dict(d, spider=None):
         flags=d.get('flags'))
 
 
+def _is_private_method(name):
+    return name.startswith('__') and not name.endswith('__')
+
+
 def _find_method(obj, func):
     if obj:
         try:
@@ -80,7 +80,7 @@ def _find_method(obj, func):
         else:
             if func_self is obj:
                 name = six.get_method_function(func).__name__
-                if private_name_regex.search(name):
+                if _is_private_method(name):
                     classname = obj.__class__.__name__.lstrip('_')
                     name = '_%s%s' % (classname, name)
                 return name
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -2,10 +2,14 @@
 Helper functions for serializing (and deserializing) requests.
 """
 import six
+import re
 
 from scrapy.http import Request
 from scrapy.utils.python import to_unicode, to_native_str
 from scrapy.utils.misc import load_object
+
+
+private_name_regex = re.compile('^__[^_](.*[^_])?_?$')
 
 
 def request_to_dict(request, spider=None):
@@ -76,7 +80,7 @@ def _find_method(obj, func):
         else:
             if func_self is obj:
                 name = six.get_method_function(func).__name__
-                if name.startswith('__'):
+                if private_name_regex.search(name):
                     classname = obj.__class__.__name__.lstrip('_')
                     name = '_%s%s' % (classname, name)
                 return name
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -9,7 +9,7 @@ from scrapy.utils.python import to_unicode, to_native_str
 from scrapy.utils.misc import load_object
 
 
-private_name_regex = re.compile('^__[^_](.*[^_])?_?$')
+private_name_regex = re.compile('^__.*[^_]_?$')
 
 
 def request_to_dict(request, spider=None):
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -71,6 +71,16 @@ def _is_private_method(name):
     return name.startswith('__') and not name.endswith('__')
 
 
+def _mangle_private_name(obj, func, name):
+    qualname = getattr(func, '__qualname__', None)
+    if qualname is None:
+        classname = obj.__class__.__name__.lstrip('_')
+        return '_%s%s' % (classname, name)
+    else:
+        splits = qualname.split('.')
+        return '_%s%s' % (splits[-2], splits[-1])
+
+
 def _find_method(obj, func):
     if obj:
         try:
@@ -81,13 +91,7 @@ def _find_method(obj, func):
             if func_self is obj:
                 name = six.get_method_function(func).__name__
                 if _is_private_method(name):
-                    qualname = getattr(func, '__qualname__', None)
-                    if qualname is None:
-                        classname = obj.__class__.__name__.lstrip('_')
-                        name = '_%s%s' % (classname, name)
-                    else:
-                        splits = qualname.split('.')
-                        name = '_%s%s' % (splits[-2], splits[-1])
+                    return _mangle_private_name(obj, func, name)
                 return name
     raise ValueError("Function %s is not a method of: %s" % (func, obj))
 
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -75,7 +75,11 @@ def _find_method(obj, func):
             pass
         else:
             if func_self is obj:
-                return six.get_method_function(func).__name__
+                name = six.get_method_function(func).__name__
+                if name.startswith('__'):
+                    classname = obj.__class__.__name__.lstrip('_')
+                    name = '_%s%s' % (classname, name)
+                return name
     raise ValueError("Function %s is not a method of: %s" % (func, obj))
 
 
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -81,8 +81,13 @@ def _find_method(obj, func):
             if func_self is obj:
                 name = six.get_method_function(func).__name__
                 if _is_private_method(name):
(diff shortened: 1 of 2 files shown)


andrewbaxter:
AppVeyor build failure looks unrelated.

elacuesta:
This looks good I think. Just one comment: according to https://docs.python.org/3.5/tutorial/classes.html#private-variables, _"Any identifier of the form __spam (at least two leading underscores, at most one trailing underscore) is textually replaced with _classname__spam, where classname is the current class name with leading underscore(s) stripped"_. I wonder if we should check the " at most one trailing underscore" part as well.

scrapy/utils/reqser.py
```diff
@@ -2,12 +2,16 @@
...
 """
 import six
+import re
 
 from scrapy.http import Request
 from scrapy.utils.python import to_unicode, to_native_str
 from scrapy.utils.misc import load_object
 
 
+private_name_regex = re.compile('^__[^_](.*[^_])?_?$')
```
elacuesta: I'm not sure I understand the use of `^` in the middle of the pattern. IIRC `^` means the start of the string (https://docs.python.org/3/library/re.html#regular-expression-syntax). Also, `.*` without `?` is greedy so it would match everything, including trailing underscores.
What about `^___*.*?_?$`?

elacuesta: This is not working for mw with the following:
```python
re.match('^__[^_](.*[^_])?_?$', '___private_stuff_')
```
I think this is difficult to do just with a regex, because you need to allow an arbitrary amount of underscores in the middle of the identifier. What about something like `'^(_*).*?(_*)$'`, which captures underscores at the beginning and end, and then you count the amount?
I'm also concerned about the performance impact of applying the regex, this would be done for each request that goes through the scheduler.

tests/test_utils_reqser.py
```diff
@@ -87,6 +93,9 @@ def parse_item(self, response):
     def handle_error(self, failure):
         pass
 
+    def __parse_item_private(self, response):
```
elacuesta: Could you maybe add a test that accounts for methods that do not match the requirements? More than one trailing underscore, for instance.

scrapy/utils/reqser.py
Gallaecio: > I'm not sure I understand the use of `^` in the middle of the pattern. IIRC `^` means the start of the string (https://docs.python.org/3/library/re.html#regular-expression-syntax). Also, `.*` without `?` is greedy so it would match everything, including trailing underscores.

`^` at the beginning of `[…]` means that it matches any character _not_ in `[…]`.

Gallaecio: The only issue with `'^__[^_](.*[^_])?_?$'` is that it does not allow more than 2 underscores at the beginning.

What about `'^__+.*[^_]_?$'`?

andrewbaxter: Thanks!  I updated the regex and did something basically like that, and added a unit test to test the regex directly since it was easier doing that than creating a bunch of classes/callbacks to test various name underscore combinations.

I'll try the bench - if this is too slow maybe it could be memoized?  Function names are typically constant through the job.

tests/test_utils_reqser.py
andrewbaxter: Ah, commented above but adding a test for the regex directly to make sure it matches/doesn't match variously underscored names.

scrapy/utils/reqser.py
elacuesta: True, thanks @Gallaecio! It's explained in the same section of the docs I linked :man_facepalming:
That also explains the _greedy_ situation I mentioned.
Anyway, now I wonder if maybe using regex is too expensive? What about just `name.startswith('__') and not name.endswith('__')`?

```python
In [1]: import re 
   ...: private_name_regex = re.compile('^__.*[^_]_?$') 
   ...: names = ['__foo', 'foo_bar', '__foo_bar_', '___foobar__']                                                                                                                                                                             

In [2]: %timeit [private_name_regex.search(n) for n in names]                                                                                                                                                                                 
1.45 µs ± 40.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit [n.startswith('__') and not n.endswith('__') for n in names]                                                                                                                                                                  
806 ns ± 1.38 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)ndswith('__') for n in names]
```

elacuesta:
> AppVeyor build failure looks unrelated.

It seems to be related to the change, I believe it has to do with the temporary spider's default callback in the `parse` command. What I'm not sure is why it fails only in AppVeyor and not in Travis :thinking:

scrapy/utils/reqser.py
elacuesta: > if this is too slow maybe it could be memoized? Function names are typically constant through the job.

I agree, creating a separate function and using `functools.lru_cache` looks like a good option here.

andrewbaxter:
I had to look at the logs for like 10 minutes, but it seems like there's a ton of tests failing and they're all failing because an include path is wrong or something:
```
overage.py warning: --include is ignored because --source is set (include-ignored)\r\nc:\\projects\\scrapy\\.tox\\py36\\scripts\\python.exe: Error while finding module specification for 'scrapy.cmdline' (ModuleNotFoundError: No module named 'scrapy')
```
So the parse tests are failing but it's because `scrapy.cmdline` isn't there... AFAICT.

scrapy/utils/reqser.py
andrewbaxter: Sorry, didn't get to the benchmarking but I switched to the `startswith`/`endswith` solution because it's easier to read and faster :) and all the tests pass.

Still important to memoize?

elacuesta: I think it would be good to memoize, you're right that several requests will have the same callback so it's probably not necessary to check them every time.

andrewbaxter: Not what I was originally going to bench, but benched with/without lru_cache:

**With lru_cache**
```
Test = 'Book Spider' Iterations = '10'


Mean : 65.64734686683296 Median : 65.72720576665657 Std Dev : 0.4875654825761466
```

**Without lru_cache**
```
Test = 'Book Spider' Iterations = '10'


Mean : 65.23675882544786 Median : 65.02926278845428 Std Dev : 0.6786585837183575
```

Both were using the code on this branch otherwise.

So my computer isn't exactly a clean bench environment and the samples are fairly close -- according to [this :D](http://courses.umass.edu/pubp608/lectures/l6.pdf), I think any optimization at this level is probably lost in the noise.  Also, I lru_cache'd `_find_method` which might be beyond the scope of this change - maybe optimizing function lookup could be a separate PR?

andrewbaxter:
Okay, now both CIs are failing :sweat_smile: 

I'm claiming innocence on the Travis py36 build as well - python 3.6 tests work fine on my computer and I can't for the life of me imagine how this change would affect deprecation notice warnings.

scrapy/utils/reqser.py
elacuesta: Yeah that sounds good. Also, sorry for bringing `functools.lru_cache` into the conversation without checking its availability first, but it was added in py3.2 and we're not dropping support for py2 just yet :-/

Gallaecio:
(closing and reopening for tests to re-run)

Gallaecio:
I would simply ignore the Windows-specific and coverage CI builds here.

andrewbaxter:
I found an issue - this breaks if the method's declaring class is different than the `obj` class (ex: the method was inherited).

AFAICT there's no way to reliably get the declaring class in multiple python versions without a lot of extra code -- Python 3.3+ has `__qualname__` which can be parsed to generate the correct name, there's a `qualname` package which tries to do it by source inspection, and otherwise the solution is to walk the class hierarchy tree and try to locate the first appearance of the method.  There's a `six` issue referencing this but they defer to the`qualname` package.

Since this is essentially a new feature I expect only new projects will require this and it may be safe to additionally assume they're using a recent Python version, so I think a solution that relies on `__qualname__` may be the cleanest solution here.  Updating.

andrewbaxter:
Build failure again looks unrelated, this time something about the docs build.

Gallaecio:
Yes, I’ve just [taken care](https://github.com/scrapy/scrapy/pull/3806) of that documentation issue.

scrapy/utils/reqser.py
```diff
@@ -75,7 +79,16 @@ def _find_method(obj, func):
...
-                return six.get_method_function(func).__name__
+                name = six.get_method_function(func).__name__
+                if _is_private_method(name):
+                    qualname = getattr(func, '__qualname__', None)
+                    if qualname is None:
+                        classname = obj.__class__.__name__.lstrip('_')
+                        name = '_%s%s' % (classname, name)
+                    else:
+                        splits = qualname.split('.')
+                        name = '_%s%s' % (splits[-2], splits[-1])
```
kmike: do you think it is possible to extract this logic to a separate function, e.g. `_mangled_name` (?), to make the intention of the code more clear, and also to be able to add tests for it? You can check that names produced are the same as Python makes.

andrewbaxter: I added tests for the two cases (own method, inherited method)... I feel like further testing would be going into "writing tests to prove Python behavior" rather than "writing tests to prove Scrapy behavior" but please let me know if you think a test with a nested class or something could be helpful.

kmike: 👍 thanks @andrewbaxter!

andrewbaxter:
Tests fixed (uh, relatively) and ready for re-review.

PR #3791, opened 2019-05-24
[MRG+1] Fix documentation for spiderloader

barraponto:
Docs mentioned the wrong module.


Gallaecio:
Good catch, thanks!

kmike:
👍

PR #3792, opened 2019-05-24
[MRG+1] Add 429 to RETRY_HTTP_CODES

barraponto:
I believe 429 should be added to RETRY_HTTP_CODES. It literally means to be retried after a certain while.


barraponto:
Second thoughts: `429` requests (and `503` and `301`) may have `Retry-After` headers specifying dates (or delay times) to wait before re-trying.

Gallaecio:
That is correct, but that is already the case with `503`, which is also in the list. I still think adding `429` to the list make sense. Handling delays on retries is a different topic.

PR #3794, opened 2019-05-25
[MRG+1] Fix form methods in FormRequest.from_response (#3777)

### LINKED ISSUE #3777: Whitelist form methods in FormRequest.from_response
Hi team, according to this [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method), there are 3 methods that are accepted in form's `method` attribute. I can't remember about other methods, but I'd agree to consider the rest of [REST verbs](https://restfulapi.net/http-methods/) if that's common. Anyway, I think it should be a good idea to whitelist the accepted methods to avoid scenarios like this [vulnerability exploitation](https://medium.com/alertot/web-scraping-considered-dangerous-exploiting-the-telnet-service-in-scrapy-1-5-2-ad5260fea0db) that took advantage of form's method.

This issue affects `FormRequest.from_response` and the affected line is: https://github.com/scrapy/scrapy/blob/a3d38041e230f886d55569f861a11a96ab8a1913/scrapy/http/request/form.py#L51

I want to know if there some reason behind this behavior, otherwise I could send a pull request.

csalazar:
Fixes #3777

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -18,7 +18,7 @@ from scrapy.utils.response import get_base_url
 
 
 class FormRequest(Request):
-    valid_form_methods = ['GET', 'POST', 'DIALOG']
+    valid_form_methods = ['GET', 'POST']
 
     def __init__(self, *args, **kwargs):
         formdata = kwargs.pop('formdata', None)
@@ -50,9 +50,11 @@ class FormRequest(Request):
         formdata = _get_inputs(form, formdata, dont_click, clickdata, response)
         url = _get_form_url(form, kwargs.pop('url', None))
 
-        method = kwargs.pop('method', form.method).upper()
-        if method not in cls.valid_form_methods:
-            method = 'GET'
+        method = kwargs.pop('method', form.method)
+        if method is not None:
+            method = method.upper()
+            if method not in cls.valid_form_methods:
+                method = 'GET'
 
         return cls(url=url, method=method, formdata=formdata, **kwargs)
 
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -18,6 +18,7 @@ from scrapy.utils.response import get_base_url
 
 
 class FormRequest(Request):
+    valid_form_methods = ['GET', 'POST', 'DIALOG']
 
     def __init__(self, *args, **kwargs):
         formdata = kwargs.pop('formdata', None)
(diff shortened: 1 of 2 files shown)


scrapy/http/request/form.py
```diff
@@ -48,7 +49,11 @@ def from_response(cls, response, formname=None, formid=None, formnumber=0, formd
         form = _get_form(response, formname, formid, formnumber, formxpath)
         formdata = _get_inputs(form, formdata, dont_click, clickdata, response)
         url = _get_form_url(form, kwargs.pop('url', None))
-        method = kwargs.pop('method', form.method)
+
+        method = kwargs.pop('method', form.method).upper()
+        if method not in cls.valid_form_methods:
+            raise ValueError('Invalid form method in chosen form')
```
kmike: I wonder if we should follow standard more closely, and consider unknown and invalid methods as GET (see https://www.w3.org/TR/html5/sec-forms.html#element-attrdef-form-method):

> The invalid value default for these attributes is the GET state. The missing value default for the method attribute is also the GET state. (There is no missing value default for the formmethod attribute.)

csalazar: Yep, I agree with that, I'm going to modify the PR.

csalazar:
Updated with Mikhail's feedback

scrapy/http/request/form.py
```diff
@@ -48,7 +49,11 @@ def from_response(cls, response, formname=None, formid=None, formnumber=0, formd
         form = _get_form(response, formname, formid, formnumber, formxpath)
         formdata = _get_inputs(form, formdata, dont_click, clickdata, response)
         url = _get_form_url(form, kwargs.pop('url', None))
-        method = kwargs.pop('method', form.method)
+
+        method = kwargs.pop('method', form.method).upper()
```
kmike: I think previously passing method=None explicitly was supported, but now it is not - could you please fix this small backwards compatibility issue?

csalazar: and `method=None` should be `GET` or `None`?

Gallaecio: `GET`, I believe.

kmike: That's an interesting question, because current behavior is POST. 

1) method is not passed => we use form.method fallback
2) `<form method>` not defined in HTML => I believe lxml returns "GET" in this case (need to check it), so method becomes GET
3) method=None is passed explicitly => currently method=None is passed through to `__init__`, which means method becomes POST. 

(3) is what is changed - a case where an user pass method=None explicitly. Not sure about the use case, just noticed that it is backwards incompatible. Passing through None should be enough to fix it.

kmike:
Thanks @csalazar!

PR #3796, opened 2019-05-27
[MRG+1] [GSoC 2019] Interface for robots.txt parsers

anubhavp28:
This pull request is not ready for merging. Looking for reviews and suggestions. Excited for the awesome summer ahead. :)

@Gallaecio @whalebot-helmsman 

Will there be any benefit from using python's abc library for creating `BaseRobotsTxtParser` class here?

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -952,31 +952,31 @@ interface described below.
 
     .. method:: __init__(content)
 
-      Parse content of robots.txt_ file as bytes.
+      Parse the content of a robots.txt_ file as bytes.
 
-      :param content: Content of robots.txt_ file.
+      :param content: Content of a robots.txt_ file.
       :type content: bytes
 
-    .. method:: allowed(url, useragent)
+    .. method:: allowed(url, user_agent)
 
-      Return ``True`` if ``url`` is allowed for crawling by the given ``useragent``, otherwise return ``False``.
+      Return ``True`` if  ``user_agent`` is allowed to crawl ``url``, otherwise return ``False``.
 
-      :param url: Absolute URL to find crawling permission for.
+      :param url: Absolute URL
       :type url: string
 
-      :param useragent: User agent to find crawling permission for.
-      :type useragent: string
+      :param user_agent: User agent
+      :type user_agent: string
 
     .. method:: sitemaps()
 
       Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return an empty generator.
 
-    .. method:: crawl_delay(useragent)
+    .. method:: crawl_delay(user_agent)
 
       Return time (in seconds) specified with ``Crawl-delay`` directive as a float. If nothing is specified, return ``None``.
 
-      :param useragent: User agent to find ``Crawl-delay`` for.
-      :type useragent: string
+      :param user_agent: User agent to find ``Crawl-delay`` for.
+      :type user_agent: string
 
     .. method:: preferred_host()
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -997,47 +997,15 @@ In order to use this parser:
 Implementing support for a new parser
 -------------------------------------
 
-You can implement support for a new robots.txt_ parser by implementing the
-interface described below.
+You can implement support for a new robots.txt_ parser by subclassing
+the abstract base class :class:`~scrapy.robotstxt.RobotParser` and
+implementing the methods described below.
 
-.. class:: scrapy.robotstxt.RobotsParser
+.. module:: scrapy.robotstxt
+   :synopsis: robots.txt parser interface and implementations
 
-    .. method:: from_crawler(crawler, robotstxt_body)
-
-      Parse the content of a robots.txt_ file as bytes. This must be a class method.
-      It must return a new instance of the parser backend.
-
-      :param crawler: crawler which made the request
-      :type crawler: :class:`~scrapy.crawler.Crawler` instance
-
-      :param robotstxt_body: content of a robots.txt_ file.
-      :type robotstxt_body: bytes
-
-    .. method:: allowed(url, user_agent)
-
-      Return ``True`` if  ``user_agent`` is allowed to crawl ``url``, otherwise return ``False``.
-
-      :param url: Absolute URL
-      :type url: string
-
-      :param user_agent: User agent
-      :type user_agent: string
-
-    .. method:: sitemaps()
-
-      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return an empty generator.
-
-    .. method:: crawl_delay(user_agent)
-
-      Return time (in seconds) specified with ``Crawl-delay`` directive as a float. If nothing is specified, return ``None``.
-
-      :param user_agent: User agent to find ``Crawl-delay`` for.
-      :type user_agent: string
-
-    .. method:: preferred_host()
-
-       Return preferred domain specified with ``Host`` directive. If nothing is specified, return ``None``.
-
+.. autoclass:: RobotParser
+   :members:
 
 .. _robots.txt: http://www.robotstxt.org/
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -954,35 +954,31 @@ interface described below.
 
     .. method:: __init__(content)
 
-      Parse content of `robots.txt <http://www.robotstxt.org/>`_ file.
+      Parse content of `robots.txt <http://www.robotstxt.org/>`_ file as bytes.
 
-      :param content: Content of `robots.txt <http://www.robotstxt.org/>`_ file. For Python 2.7, if ``content``
-                    contains non-ASCII characters, it should be ``UTF-8`` encoded. 
-      :type content: ``str``
+      :param content: Content of `robots.txt <http://www.robotstxt.org/>`_ file.
+      :type content: bytes
 
     .. method:: allowed(url, useragent)
 
       Return ``True`` if ``url`` is allowed for crawling by the given ``useragent``, otherwise return ``False``.
 
-      :param url: URL to find crawling permission for. For Python 2.7, if ``url`` contains non-ASCII characters, 
-                  it should be ``UTF-8`` encoded.
-      :type url: ``str``
-    
-      :param useragent: User agent to find crawling permission for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
-                        it should be ``UTF-8`` encoded.
-      :type useragent: ``str`` 
+      :param url: Absolute URL to find crawling permission for.
+      :type url: string
+
+      :param useragent: User agent to find crawling permission for.
+      :type useragent: string
 
     .. method:: sitemaps()
 
-      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return ``None``.
+      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return an empty generator.
 
     .. method:: crawl_delay(useragent)
 
-      Return time (in seconds) specified with ``Crawl-delay`` directive. If nothing is specified, return ``None``.
+      Return time (in seconds) specified with ``Crawl-delay`` directive as a float. If nothing is specified, return ``None``.
 
-      :param useragent: User agent to find ``Crawl-delay`` for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
-                        it should be ``UTF-8`` encoded.
-      :type useragent: ``str`` 
+      :param useragent: User agent to find ``Crawl-delay`` for.
+      :type useragent: string
 
     .. method:: preferred_host()
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -926,9 +926,9 @@ RobotsTxtMiddleware
 
     Scrapy ships with support for the following robots.txt_ parsers:
 
-    * `RobotFileParser`_ (default)
-    * `Reppy`_
-    * `Robotexclusionrulesparser`_
+    * :ref:`RobotFileParser <python-robotfileparser>` (default)
+    * :ref:`Reppy <reppy-parser>`
+    * :ref:`Robotexclusionrulesparser <rerp-parser>`
 
     You can change the robots.txt_ parser with the :setting:`ROBOTSTXT_PARSER`
     setting. Or you can also :ref:`implement support for a new parser <support-for-new-robots-parser>`.
@@ -940,6 +940,52 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
 the request will be ignored by this middleware even if
 :setting:`ROBOTSTXT_OBEY` is enabled.
 
+.. _python-robotfileparser:
+
+RobotFileParser
+~~~~~~~~~~~~~~~
+
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
+Python's inbuilt ``robots.txt`` parser. The parser is fully compliant with `Martijn Koster's 
+1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_, with support for
+``Crawl-delay`` and ``Request-rate`` directives also available from Python 3.6. It lacks
+support for wildcard matching. Scrapy uses this parser by default.
+
+In order to use this parser, set:
+
+* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.PythonRobotParser``
+
+.. _rerp-parser:
+
+Robotexclusionrulesparser
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_ is fully compliant
+with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
+with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
+
+In order to use this parser, set:
+
+* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.RerpRobotParser``
+
+.. _reppy-parser:
+
+Reppy parser
+~~~~~~~~~~~~
+
+`Reppy <https://github.com/seomoz/reppy/>`_ is a Python wrapper around `Robots Exclusion
+Protocol Parser for C++ <https://github.com/seomoz/rep-cpp>`_. The parser is fully compliant
+with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
+with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
+Unlike `RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ and 
+`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_, it uses the length based 
+rule, in particular for ``Allow`` and ``Disallow`` directives, where the most specific rule based 
+on the length of the path trumps the less specific (shorter) rule.
+
+In order to use this parser, set:
+
+* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.ReppyRobotParser``
+
 .. _support-for-new-robots-parser:
 
 Implementing support for a new parser
@@ -950,12 +996,15 @@ interface described below.
 
 .. class:: scrapy.extensions.robotstxtparser.RobotsParser
 
-    .. method:: __init__(content)
+    .. method:: __init__(content, crawler)
 
       Parse the content of a robots.txt_ file as bytes.
 
       :param content: Content of a robots.txt_ file.
       :type content: bytes
+
+      :param crawler: crawler which made the request
+      :type crawler: :class:`~scrapy.crawler.Crawler` instance
 
     .. method:: allowed(url, user_agent)
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -922,12 +922,71 @@ RobotsTxtMiddleware
     To make sure Scrapy respects robots.txt make sure the middleware is enabled
     and the :setting:`ROBOTSTXT_OBEY` setting is enabled.
 
+    This middleware has to be combined with a `robots.txt <http://www.robotstxt.org/>`_ parser.
+
+    Scrapy ships with support for three `robots.txt <http://www.robotstxt.org/>`_ parsers:
+
+    * `RobotFileParser`_
+    * `Reppy`_
+    * `Robotexclusionrulesparser`_
+
+    You can change the `robots.txt <http://www.robotstxt.org/>`_ parser with the :setting:`ROBOTSTXT_PARSER`
+    setting. Or you can also :ref:`implement support for a new parser <support-for-new-robots-parser>`.
+
 .. reqmeta:: dont_obey_robotstxt
 
 If :attr:`Request.meta <scrapy.http.Request.meta>` has
 ``dont_obey_robotstxt`` key set to True
 the request will be ignored by this middleware even if
 :setting:`ROBOTSTXT_OBEY` is enabled.
+
+.. _support-for-new-robots-parser:
+
+Implementing support for a new parser
+-------------------------------------
+
+You can implement support for a new `robots.txt <http://www.robotstxt.org/>`_ parser by implementing the
+interface described below.
+
+.. module:: scrapy.extensions.robotstxtparser
+
+.. class:: RobotsParser
+
+    .. method:: __init__(content)
+
+      Parse content of `robots.txt <http://www.robotstxt.org/>`_ file.
+
+      :param content: Content of `robots.txt <http://www.robotstxt.org/>`_ file. For Python 2.7, if ``content``
+                    contains non-ASCII characters, it should be ``UTF-8`` encoded. 
+      :type content: ``str``
+
+    .. method:: allowed(url, useragent)
+
+      Return ``True`` if ``url`` is allowed for crawling by the given ``useragent``, otherwise return ``False``.
+
+      :param url: URL to find crawling permission for. For Python 2.7, if ``url`` contains non-ASCII characters, 
+                  it should be ``UTF-8`` encoded.
+      :type url: ``str``
+    
+      :param useragent: User agent to find crawling permission for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
+                        it should be ``UTF-8`` encoded.
+      :type useragent: ``str`` 
+
+    .. method:: sitemaps()
+
+      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return ``None``.
+
+    .. method:: crawl_delay(useragent)
+
+      Return time (in seconds) specified with ``Crawl-delay`` directive. If nothing is specified, return ``None``.
+
+      :param useragent: User agent to find ``Crawl-delay`` for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
+                        it should be ``UTF-8`` encoded.
+      :type useragent: ``str`` 
+
+    .. method:: preferred_host()
+
+      Return preferred domain specified with ``Host`` directive. If nothing is specified, return ``None``.
 
 
 DownloaderStats
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -947,13 +947,12 @@ RobotFileParser
 
 `RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
 Python's inbuilt ``robots.txt`` parser. The parser is fully compliant with `Martijn Koster's 
-1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_, with support for
-``Crawl-delay`` and ``Request-rate`` directives also available from Python 3.6. It lacks
+1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_. It lacks
 support for wildcard matching. Scrapy uses this parser by default.
 
 In order to use this parser, set:
 
-* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.PythonRobotParser``
+* :setting:`ROBOTSTXT_PARSER` to ``scrapy.robotstxt.PythonRobotParser``
 
 .. _rerp-parser:
 
@@ -962,11 +961,15 @@ Robotexclusionrulesparser
 
 `Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_ is fully compliant
 with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
-with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
+with support for wildcard matching.
 
-In order to use this parser, set:
+In order to use this parser:
 
-* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.RerpRobotParser``
+* Install `Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_ by running
+  ``pip install robotexclusionrulesparser``
+
+* Set :setting:`ROBOTSTXT_PARSER` setting to
+  ``scrapy.robotstxt.RerpRobotParser``
 
 .. _reppy-parser:
 
@@ -976,15 +979,18 @@ Reppy parser
 `Reppy <https://github.com/seomoz/reppy/>`_ is a Python wrapper around `Robots Exclusion
 Protocol Parser for C++ <https://github.com/seomoz/rep-cpp>`_. The parser is fully compliant
 with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
-with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
-Unlike `RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ and 
-`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_, it uses the length based 
-rule, in particular for ``Allow`` and ``Disallow`` directives, where the most specific rule based 
-on the length of the path trumps the less specific (shorter) rule.
+with support for wildcard matching. Unlike
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ and
+`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_, it uses the length based
+rule, in particular for ``Allow`` and ``Disallow`` directives, where the most specific
+rule based on the length of the path trumps the less specific (shorter) rule.
 
-In order to use this parser, set:
+In order to use this parser:
 
-* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.ReppyRobotParser``
+* Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``
+
+* Set :setting:`ROBOTSTXT_PARSER` setting to
+  ``scrapy.robotstxt.ReppyRobotParser``
 
 .. _support-for-new-robots-parser:
 
@@ -994,18 +1000,18 @@ Implementing support for a new parser
 You can implement support for a new robots.txt_ parser by implementing the
 interface described below.
 
-.. class:: scrapy.extensions.robotstxtparser.RobotsParser
+.. class:: scrapy.robotstxt.RobotsParser
 
-    .. method:: from_crawler(crawler, content)
+    .. method:: from_crawler(crawler, robotstxt_body)
 
       Parse the content of a robots.txt_ file as bytes. This must be a class method.
       It must return a new instance of the parser backend.
 
       :param crawler: crawler which made the request
       :type crawler: :class:`~scrapy.crawler.Crawler` instance
-      
-      :param content: Content of a robots.txt_ file.
-      :type content: bytes
+
+      :param robotstxt_body: content of a robots.txt_ file.
+      :type robotstxt_body: bytes
 
     .. method:: allowed(url, user_agent)
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -996,15 +996,16 @@ interface described below.
 
 .. class:: scrapy.extensions.robotstxtparser.RobotsParser
 
-    .. method:: __init__(content, crawler)
+    .. method:: from_crawler(crawler, content)
 
-      Parse the content of a robots.txt_ file as bytes.
-
-      :param content: Content of a robots.txt_ file.
-      :type content: bytes
+      Parse the content of a robots.txt_ file as bytes. This must be a class method.
+      It must return a new instance of the parser backend.
 
       :param crawler: crawler which made the request
       :type crawler: :class:`~scrapy.crawler.Crawler` instance
+      
+      :param content: Content of a robots.txt_ file.
+      :type content: bytes
 
     .. method:: allowed(url, user_agent)
 
@@ -1029,7 +1030,7 @@ interface described below.
 
     .. method:: preferred_host()
 
-      Return preferred domain specified with ``Host`` directive. If nothing is specified, return ``None``.
+       Return preferred domain specified with ``Host`` directive. If nothing is specified, return ``None``.
 
 
 .. _robots.txt: http://www.robotstxt.org/
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -922,15 +922,15 @@ RobotsTxtMiddleware
     To make sure Scrapy respects robots.txt make sure the middleware is enabled
     and the :setting:`ROBOTSTXT_OBEY` setting is enabled.
 
-    This middleware has to be combined with a `robots.txt <http://www.robotstxt.org/>`_ parser.
+    This middleware has to be combined with a robots.txt_ parser.
 
-    Scrapy ships with support for three `robots.txt <http://www.robotstxt.org/>`_ parsers:
+    Scrapy ships with support for the following robots.txt_ parsers:
 
-    * `RobotFileParser`_
+    * `RobotFileParser`_ (default)
     * `Reppy`_
     * `Robotexclusionrulesparser`_
 
-    You can change the `robots.txt <http://www.robotstxt.org/>`_ parser with the :setting:`ROBOTSTXT_PARSER`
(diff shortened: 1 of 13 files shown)


Gallaecio:
> Will there be any benefit from using python's abc library for creating `BaseRobotsTxtParser` class here?

https://stackoverflow.com/a/19328146/939364 looks like a good read on the subject. I think we should abstain from using them unless we find a good reason to do it.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
+from six.moves.urllib_robotparser import RobotFileParser
+
+from scrapy.utils.python import to_native_str
+
+class BaseRobotsTxtParser():
```
Gallaecio: Maybe this class should not exist in code, but only in the documentation. I think this is the usual approach in Scrapy.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, robotstxt_url):
```
Gallaecio: Do parsers need to know the URL of the `robots.txt` file? Does it affect rules?

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError
+
+    def parse(self, robotstxt_body):
```
Gallaecio: I assume that the download of the `robots.txt` file will be done by Scrapy, as I assume is currently the case, and not the parser. Based on that assumption (is it wrong?), I wonder if it may make sense to merge the `parse` method into the `__init__` constructor.

Also, since classes implementing this API are expected to be `robot.txt` parsers, I wonder if we should rename `robotstxt_body` to simply `body`, `content` or similar.

Finally, documentation-wise, I think that it should be clear what the data type of the input parameter is. It could be a string, bytes, a file-like object…

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
...
+        """Parse content of robots.txt file."""
+        raise NotImplementedError
+
+    def allowed(self, url, useragent):
+        """Return True if url is allowed for crawling, otherwise return False."""
+        raise NotImplementedError
+
+    def sitemaps(self):
```
Gallaecio: How will Scrapy use this method?

If it will use it, I wonder if it would make sense to make `sitemaps` return an iterator.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
...
+        """Return a list of links of sitemaps on the website. If there is no sitemap specified in robots.txt, return None."""
+        raise NotImplementedError
+
+    def crawl_delay(self, useragent):
+        """Return time specified with Crawl-delay directive. If nothing is specified, return None."""
```
Gallaecio: The documentation should probably clarify the unit of the result (e.g. seconds).

scrapy/extensions/robotstxtparser.py
anubhavp28: No. We don't need to feed the URL of the `robots.txt` file since we are not dependent on them to fetch the file. The URL of `robots.txt` does not affect rules. I am changing this.

whalebot-helmsman: I think, `robotstxt_url` is used as an argument in `RobotFileParser`'s constructor to properly catch domain name. We need to check `RobotFileParser` behaviour in case of domain mismatch between `robotstxt_url` and tested `url`.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,52 @@
...
+        """Return preferred domain specified with Host directive. If nothing is specified, return None.""" 
+        raise NotImplementedError
+
+class PythonRobotParser(BaseRobotsTxtParser):
+    def __init__(self, robotstxt_url):
+        self.robotstxt_url = robotstxt_url
+        self.rp = RobotFileParser(robotstxt_url) 
+
+    def parse(self, robotstxt_body):
+        self.rp.parse(to_native_str(robotstxt_body).splitlines())
```
whalebot-helmsman: Convert to proper type of passed data unsing `to_native_str` should be done once on a side of `RobotsTxtMiddleware`. Otherwise every implementation have to do same thing

scrapy/extensions/robotstxtparser.py
anubhavp28: This doesn't seem to be the case. I looked at `RobotFileParser` source code. It uses `urllib.parse.urlparse` to separate domain name and path. It seems to be the case with both Python 2.7 and Python 3.X .The URL is only used to fetch `robots.txt` file.

```Python
>>> rp = RobotFileParser()
>>> rp.set_url('https://www.google.com/robots.txt')
>>> content = ("User-agent: abc \n"
... "Disallow: /disallowed \n"
... "Alloe: /allow \n")
>>> rp.parse(content.splitlines())
>>> rp.can_fetch('abc', 'https://www.google.com/disallowed')
False
>>> rp.can_fetch('abc', 'https://www.google.com/allowed')
True
>>> rp.can_fetch('abc', 'https://www.facebook.com/disallowed')
False
>>> rp.can_fetch('abc', 'https://www.facebook.com/allowed')
True
```

scrapy/extensions/robotstxtparser.py
anubhavp28: Fetching the `robots.txt` file will be done by Scrapy. I am combining the `parse` and `__init__` method.

scrapy/extensions/robotstxtparser.py
anubhavp28: Currently, Scrapy does not use `sitemaps`. Sitemap is a part of robot inclusion protocol and provides a list of URLs available for crawling on the website. I thought, Scrapy can make use of this list as a starting point for crawling the website. This could be useful if say, some pages are disconnected (no other page links to them but that is very unlikely to happen). But, it seems to me that sitemaps is rarely used. I am unable to make an opinion on whether to include it in the interface or not.

@whalebot-helmsman what's your opinion on this?

Gallaecio: Well, we do have a SitemapSpider class. I’ve just looked at the [SitemapSpider documentation](https://docs.scrapy.org/en/latest/topics/spiders.html#sitemapspider), and it looks like it already supports reading sitemap URLs from a `robots.txt` file. Maybe, as a stretch goal, we could look into getting SitemapSpider to use the configured `robots.txt` parse.

So yes, +1 to `sitemaps()`, but I do think an iterator may be more flexible. So, instead of returning a list, the method could yield the sitemap URLs.

anubhavp28:
I am thinking that we won't like to have `reppy` and `Robotexclusionrulesparser` as Scrapy's dependencies. We won't have any guarantee that these modules are available. Therefore, inside the interface class we need to perform some kind of test to check the availability of these modules. How should we handle the situation such as when the user sets `ROBOTTXT_PARSER` setting to the `reppy` interface class but does not have `reppy` installed?

scrapy/extensions/robotstxtparser.py
whalebot-helmsman: In that case we should drop argument from constructor signature

whalebot-helmsman:
> How should we handle the situation such as when the user sets ROBOTTXT_PARSER setting to the reppy interface class but does not have reppy installed?

I think, it is common in scrapy to raise import exception in case some class is missing(in middlewares or queues). Nothing special required here.

scrapy/extensions/robotstxtparser.py
Gallaecio: Great research!

Gallaecio:
To ellaborate a bit more, you could use a `try … except` to import the modules that require additional dependencies, and as @whalebot-helmsman suggest, raise a specific exception from the constructor if those imports did not work, similar to how it’s done with middlewares when they are not enabled in the configuration.

The `robots.txt` middleware, which takes care of creating the parser instance, should catch this exception, log an error and possibly fallback to the default parser class.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,72 @@ RobotsTxtMiddleware
...
+.. module:: scrapy.extensions.robotstxtparser
+
+.. class:: RobotsParser
+
+    .. method:: __init__(content)
+
+      Parse content of `robots.txt <http://www.robotstxt.org/>`_ file.
+
+      :param content: Content of `robots.txt <http://www.robotstxt.org/>`_ file. For Python 2.7, if ``content``
+                    contains non-ASCII characters, it should be ``UTF-8`` encoded.
```
Gallaecio: I think the content type should be bytes (encoded string), since some of the parsers may work with bytes and for those that don’t parser classes can simply decode those bytes first.

Also, unlike with the documentation of an API that developers _use_, when you document an interface that developers _implement_, parameters are documented how they _are_ and return values how they _must be_. That is, you should not say ‘it should be’, you should say ‘it is’.

In this specific case, however, I would actually refactor the sentence a bit further. If we are passing the file contents as bytes (`response.body` from Scrapy), we should do it regardless of Python version. So this may be enough:

    Content of `robots.txt <http://www.robotstxt.org/>`_ file as bytes.

You don’t need to mention the encoding. Since Scrapy is simply going to pass the received content _as is_, we can not actually enforce a specific encoding, so the encoding really depends on the website serving the file. You can add a note that explains that robots.txt files are expected to be encoded in UTF-8, but you should not suggest that Scrapy ensures this in any way.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,72 @@ RobotsTxtMiddleware
...
+
+      :param content: Content of `robots.txt <http://www.robotstxt.org/>`_ file. For Python 2.7, if ``content``
+                    contains non-ASCII characters, it should be ``UTF-8`` encoded. 
+      :type content: ``str``
+
+    .. method:: allowed(url, useragent)
+
+      Return ``True`` if ``url`` is allowed for crawling by the given ``useragent``, otherwise return ``False``.
+
+      :param url: URL to find crawling permission for. For Python 2.7, if ``url`` contains non-ASCII characters,
```
Gallaecio: The parameter and return types should be ‘the same’ across Python versions (as in string vs bytes, Python 2 and Python 3 string types are not really the same underneath).

On a different note: What will the URL parameter that comes from Scrapy contain? Will it be the absolute, normalized URL? Just an absolute URL? Maybe not be an URL at all, and instead be an absolute path relative to the underlying domain (e.g. `/index.html` for `https://example.com/index.html`)?

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,72 @@ RobotsTxtMiddleware
...
+                  it should be ``UTF-8`` encoded.
+      :type url: ``str``
+    
+      :param useragent: User agent to find crawling permission for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
+                        it should be ``UTF-8`` encoded.
+      :type useragent: ``str`` 
+
+    .. method:: sitemaps()
+
+      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return ``None``.
```
Gallaecio: Actually, since your are using a generator, you can simply return an empty generator instead of `None` if there are no sitemap links.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,72 @@ RobotsTxtMiddleware
...
+
+    .. method:: crawl_delay(useragent)
+
+      Return time (in seconds) specified with ``Crawl-delay`` directive. If nothing is specified, return ``None``.
+
+      :param useragent: User agent to find ``Crawl-delay`` for. For Python 2.7, if ``useragent`` contains non-ASCII characters, 
+                        it should be ``UTF-8`` encoded.
+      :type useragent: ``str``
```
Gallaecio: I wonder if an `int` your make more sense.

docs/topics/downloader-middleware.rst
anubhavp28: It will be an absolute URL but not normalised.

docs/topics/downloader-middleware.rst
anubhavp28: `robots.txt` specification doesn't limit crawl delay time to integers, hence `float` should be more appropriate as the return type of `crawl_delay` function. I have made the change.

docs/topics/downloader-middleware.rst
anubhavp28: > The parameter and return types should be ‘the same’ across Python versions (as in string vs bytes, Python 2 and Python 3 string types are not really the same underneath).

Could you elaborate more on this? Not sure, If I understood you completely here.

Gallaecio: Yes, my phrasing was very unfortunate.

The indicated type is `str`. However, you mention that in Python 2.7, if non-ASCII characters are used, UTF-8 encoding must be used.

The thing is, `str` values do not need an encoding at all, they are the decoded version of byte strings.

So, if the type is `str`, the sentence about encoding should be removed, it does not really apply. If the type were `bytes`, then the sentence should stay, but it should not be about Python 2.7 specifically, it would also apply to Python 3.

scrapy/extensions/robotstxtparser.py
Gallaecio: @anubhavp28 has actually done it on the `RobotFileParser` implementation, not in the middleware, but I believe that is the right approach here.

I was not familiar with `to_native_str`, but it turns out it returns bytes in Python 2 and a string in Python 3 to deal with those Python 2 implementations that did not support Python 2’s `unicode` type, such as `RobotFileParser`.

Other parsers might need to use bytes on both versions of Python or a string on both versions of Python, so it makes sense that `to_native_str` is used by the implementations that need it and not forced by the middleware.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,68 @@ RobotsTxtMiddleware
     To make sure Scrapy respects robots.txt make sure the middleware is enabled
     and the :setting:`ROBOTSTXT_OBEY` setting is enabled.
 
+    This middleware has to be combined with a `robots.txt <http://www.robotstxt.org/>`_ parser.
```
Gallaecio: You have lots of references from `robots.txt` to http://www.robotstxt.org/ all over this page. You could benefit from the first form of external linking here: http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

That way, you only need to specify the target link once at the end of the page.

I’m not sure, but you might need to use:

    `robots.txt`_

instead of just:

    robots.txt_

if the latter does not work, and it might not work because of the period.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,68 @@ RobotsTxtMiddleware
...
+
+Implementing support for a new parser
+-------------------------------------
+
+You can implement support for a new `robots.txt <http://www.robotstxt.org/>`_ parser by implementing the
+interface described below.
+
+.. module:: scrapy.extensions.robotstxtparser
+
+.. class:: RobotsParser
```
Gallaecio: :lipstick: You should be able to indicate the complete class path, thus making the `.. module` line above unnecessary.

```suggestion
.. class:: scrapy.extensions.robotstxtparser.RobotsParser
```

tox.ini
```diff
@@ -16,6 +16,8 @@ deps =
     Pillow != 3.0.0
     leveldb
     -rtests/requirements-py2.txt
+    reppy
+    robotexclusionrulesparser
```
Gallaecio: Instead of adding them here and below, add them to the corresponding test requirement files (`tests/requirements-py2.txt`, `tests/requirements-py3.txt`), as they are necessary to run the tests.

tests/test_extension_robotstxtparser.py
```diff
@@ -0,0 +1,143 @@
+from twisted.trial import unittest
+from scrapy.utils.python import to_native_str
+from scrapy.extensions.robotstxtparser import PythonRobotParser, ReppyRobotParser, RerpRobotParser
+
+class PythonRobotParserTest(unittest.TestCase):
+    def test_allowed(self):
+        robotstxt_content = ("User-agent: * \n"
```
whalebot-helmsman: Code for most of the tests methods copied across test classes:
- `test_allowed` is a full copy
- `test_preferred_host` is a full copy
- `test_crawl_delay` is almost a full copy
Can you use some mechanism to write common code only once?

tests/test_extension_robotstxtparser.py
```diff
@@ -0,0 +1,143 @@
...
+                            "Allow: /allowed \n"
+                            "Crawl-delay: 10".encode('utf-8'))
+        rp = PythonRobotParser(robotstxt_content)
+        from six.moves.urllib_robotparser import RobotFileParser
+        if hasattr(RobotFileParser, "crawl_delay"):
+            self.assertTrue(rp.crawl_delay("*") == 10.0)
+        else:
+            self.assertTrue(rp.crawl_delay("*") is None)
+
+        robotstxt_content = ("User-agent: * \n"
```
whalebot-helmsman: I think, these should be two separate test cases `test_crawl_delay` and `test_no_crawl_delay`

tests/test_extension_robotstxtparser.py
```diff
@@ -0,0 +1,143 @@
...
+                            "Disallow: /disallowed \n"
+                            "Allow: /allowed \n"
+                            "Sitemap: https://example.com/sitemap.xml".encode('utf-8'))
+        rp = ReppyRobotParser(robotstxt_content)
+        sitemaps = list(rp.sitemaps())
+        self.assertTrue(len(sitemaps) == 1)
+        self.assertTrue("https://example.com/sitemap.xml" in sitemaps)
+
+        robotstxt_content = ("User-agent: * \n"
```
whalebot-helmsman: I think, these should be two separate test cases `test_sitemap` and `test_no_sitemap`

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,64 @@
+from six.moves.urllib_robotparser import RobotFileParser
+
+from scrapy.utils.python import to_native_str
```
whalebot-helmsman: There are more test cases for parser in https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware_robotstxt.py, can you copy them here
https://github.com/scrapy/scrapy/blob/f755507d682ba4e093c5ea85c6a4761d94c71436/tests/test_downloadermiddleware_robotstxt.py#L54
https://github.com/scrapy/scrapy/blob/f755507d682ba4e093c5ea85c6a4761d94c71436/tests/test_downloadermiddleware_robotstxt.py#L90
https://github.com/scrapy/scrapy/blob/f755507d682ba4e093c5ea85c6a4761d94c71436/tests/test_downloadermiddleware_robotstxt.py#L112

tests/test_extension_robotstxtparser.py
anubhavp28: I have moved all the test cases to a base class. Then I have used a separate class for each parser, to call relevant test cases. Would this help?

anubhavp28:
Is there any particular reason why we are using `unittest` module from `twisted.trial` rather using the Python's inbuilt `unittest` module? `twisted.trial.unitttest` seems to be lacking functionality such as `expectedFailure` (https://twistedmatrix.com/trac/ticket/6336) which we can use to reduce duplication of code.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,97 @@
...
+            content = ''
+        self.rp = RobotFileParser()
+        self.rp.parse(content.splitlines()) 
+
+    def allowed(self, url, useragent):
+        try:
+            useragent = to_native_str(useragent)
+            url = to_native_str(url)
+        except UnicodeDecodeError:
+            return False
```
Gallaecio: Maybe we should log a warning here.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,97 @@
...
+        return self.rp.can_fetch(useragent, url)
+
+    def sitemaps(self):
+        """RobotFileParser does not support Sitemaps directive."""
+        def _get_empty_generator():
```
Gallaecio: You can remove this line, the `return _get_empty_generator()` one, and unindent the remaining 2 lines, and it should work the same way :)

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,97 @@
...
+            return
+            yield
+        return _get_empty_generator()
+
+    def crawl_delay(self, useragent):
+        """RobotFileParser does not support Crawl-delay directive for Python version < 3.6 ."""
+        try:
+            useragent = to_native_str(useragent)
+        except UnicodeDecodeError:
```
Gallaecio: May be worth logging a warning.

Gallaecio:
> Is there any particular reason why we are using `unittest` module from `twisted.trial` rather using the Python's inbuilt `unittest` module? `twisted.trial.unitttest` seems to be lacking functionality such as `expectedFailure` (https://twistedmatrix.com/trac/ticket/6336) which we can use to reduce duplication of code.

I’m guessing that is has something to do with Twisted’s asynchronous nature, that some tests may not work otherwise. But I don’t know for sure.

scrapy/extensions/robotstxtparser.py
anubhavp28: I think we should modify the `__init__` function to take `spider` as an argument, we can use this spider reference while logging any message. Otherwise, the user might face difficulty trying to figure out which spider a log message corresponds to.

Gallaecio: Maybe we should pass the ``crawler``, it gives access to the spider, the settings, and some other data that developers may want to have available in their custom parsers.

scrapy/extensions/robotstxtparser.py
anubhavp28: @Gallaecio I think, we should remove the `try/catch` block here. Since if we go by our spec : `user_agent` and `url` have to be strings. Hence `UnicodeDecodeError` can never occur. What do you think?

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -85,26 +88,25 @@ def _logerror(self, failure, request, spider):
...
-        rp.parse(to_native_str(body).splitlines())
+        
+        try:
+            rp = load_object(self._parserimpl)(response.body)
+        except ImportError as e:
+            # For Python 2 compatibility
+            errmsg = e.msg if hasattr(e, 'msg') else e.message
+
+            missingmodule = re.match('No module named (.+)', errmsg).group(1).strip("'")
+            logger.error('Unable to use %(robotparser)s . Do you have %(module)s installed?'
```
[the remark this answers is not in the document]
anubhavp28: @Gallaecio Do we need tests for making sure that a log message is created, when a parser module is not found?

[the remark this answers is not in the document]
Gallaecio: That would be great, at least to check that a warning is indeed issues, as opposed to no warning. You should be able to find examples of log message capture in other tests.

scrapy/extensions/robotstxtparser.py
Gallaecio: I think you are right.

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -85,26 +88,25 @@ def _logerror(self, failure, request, spider):
...
+            errmsg = e.msg if hasattr(e, 'msg') else e.message
+
+            missingmodule = re.match('No module named (.+)', errmsg).group(1).strip("'")
+            logger.warning('Unable to use %(robotparser)s . Do you have %(module)s installed?'
+                        'Falling back to the default robots.txt parser.',
+                        {'robotparser': self._parserimpl, 'module': missingmodule}, 
+                        exc_info=sys.exc_info(),
+                        extra={'spider': spider})
+            
+            rp = PythonRobotParser(response.body, self.crawler)
```
kmike: Why do we need a fallback here? I think it user asks for a specific robots.txt parsing backend, and backend is not available, we should be raising an error, not a warning.

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,130 @@
+import sys
+import logging
+from six.moves.urllib_robotparser import RobotFileParser
+
+from scrapy.utils.python import to_native_str, to_unicode
+
+logger = logging.getLogger(__name__)
+
+class PythonRobotParser():
+    def __init__(self, content, crawler):
```
kmike: I think we can make it a Scrapy component like others, with from_crawler method, and pass just `spider` to `__init__` method (it is needed only for logging, right?).

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,130 @@
...
+        self.crawler = crawler
```
kmike: is this variable used?

anubhavp28: No, I have removed that variable in the follow-up commit.

scrapy/extensions/robotstxtparser.py
anubhavp28: Yes, it is used only for logging. I have made the change.

tox.ini
```diff
@@ -110,3 +110,28 @@ changedir = {[docs]changedir}
...
+basepython = python3.7
+deps = 
+    {[testenv:py34]deps}
+    robotexclusionrulesparser
+
+[testenv:rerp-py27]
+basepython = python2.7
+deps = 
+    {[testenv]deps}
+    robotexclusionrulesparser
```
kmike: 4 new environments look a bit too much to me - what do you think about having two, and installing (or not installing) extra dependencies there, both reppy and robotexclusionrulesparser? It may be not specific to robots.txt, we may have a couple of environments with various optional dependencies installed (or not installed, depending on what we prefer to test more).

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,118 @@ RobotsTxtMiddleware
...
 
+.. _python-robotfileparser:
+
+RobotFileParser
+~~~~~~~~~~~~~~~
+
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
+Python's inbuilt ``robots.txt`` parser. The parser is fully compliant with `Martijn Koster's 
+1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_, with support for
+``Crawl-delay`` and ``Request-rate`` directives also available from Python 3.6. It lacks
```
kmike: While technically correct, these (and other) docs suggest that Scrapy supports Crawl-delay directive, while it does nothing. I think this should be made clear it is not supported, or we may restraing from mentioning them until support is implemented.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,118 @@ RobotsTxtMiddleware
...
+.. _rerp-parser:
+
+Robotexclusionrulesparser
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_ is fully compliant
+with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
+with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
+
+In order to use this parser, set:
```
kmike: It'd be nice to provide some install instructions as well (`pip install repr`? if not, maybe link to a page with install instructions?). This applies to reppy as well.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,118 @@ RobotsTxtMiddleware
...
+
+      :param url: Absolute URL
+      :type url: string
+
+      :param user_agent: User agent
+      :type user_agent: string
+
+    .. method:: sitemaps()
+
+      Return a generator yielding URL to sitemaps on the website. If there is no sitemap specified, return an empty generator.
```
kmike: we're not using this method anywhere, right?

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,118 @@ RobotsTxtMiddleware
...
+
+.. _support-for-new-robots-parser:
+
+Implementing support for a new parser
+-------------------------------------
+
+You can implement support for a new robots.txt_ parser by implementing the
+interface described below.
+
+.. class:: scrapy.extensions.robotstxtparser.RobotsParser
```
kmike: do you think it is possible to use autodocs here? See https://docs.scrapy.org/en/master/contributing.html#documentation-policies

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -85,27 +87,24 @@ def _logerror(self, failure, request, spider):
...
-
+        
+        try:
+            rp = load_object(self._parserimpl).from_crawler(self.crawler, response.body)
+        except ImportError as e:
+            # For Python 2 compatibility
+            errmsg = e.msg if hasattr(e, 'msg') else e.message
+
+            missingmodule = re.match('No module named (.+)', errmsg).group(1).strip("'")
+            logger.error('Unable to use \'%(robotparser)s\'. Do you have \'%(module)s\' module installed?',
```
kmike: why is it logger.error, and not a actual exception?

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, crawler):
         self.crawler = crawler
         self._useragent = crawler.settings.get('USER_AGENT')
```
kmike: I know this is how it worked before, and this is a separate issue, but using `crawler.settings.get('USER_AGENT')` doesn't feel right - user agent middleware can be disabled, and users may also override user-agent per request.

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -85,27 +87,24 @@ def _logerror(self, failure, request, spider):
...
-                # Running rp.parse() will set rp state from
-                # 'disallow all' to 'allow any'.
-                self.crawler.stats.inc_value('robotstxt/unicode_error_count')
-        # stdlib's robotparser expects native 'str' ;
-        # with unicode input, non-ASCII encoded bytes decoding fails in Python2
-        rp.parse(to_native_str(body).splitlines())
-
+        
+        try:
+            rp = load_object(self._parserimpl).from_crawler(self.crawler, response.body)
```
kmike: why is `_parserimpl` loaded for each request, why not load it in  `__init__`?

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,137 @@
+import sys
+import logging
+
+from scrapy.utils.python import to_native_str, to_unicode
+
+logger = logging.getLogger(__name__)
+
+class PythonRobotParser():
+    def __init__(self, content, spider):
```
kmike: What doyou think about renaming `content` variable name to `robotstxt_body`, or something like that, for clarity?

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,137 @@
+import sys
+import logging
+
+from scrapy.utils.python import to_native_str, to_unicode
+
+logger = logging.getLogger(__name__)
+
+class PythonRobotParser():
```
kmike: it should be either `PythonRobotParser:` (Python 3 style), or `PythonRobotParser(object)` (careful Python 2 style)

scrapy/extensions/robotstxtparser.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def preferred_host(self):
+        """RobotFileParser does not support Host directive."""
+        logger.warning("Failure while retrieving preferred host using %(parser)s." 
+            "%(parser)s does not support Host directive.",
+            {'parser': "RobotFileParser"},
+            exc_info=sys.exc_info(),
+            extra={'spider': self.spider})
+        return None
+
```
kmike: nitpick: according to pep8, there should be two blank lines between class definitions

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,118 @@ RobotsTxtMiddleware
...
+with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
+with support for wildcard matching, ``Crawl-delay`` directive, and ``Sitemap`` directive.
+Unlike `RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ and 
+`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_, it uses the length based 
+rule, in particular for ``Allow`` and ``Disallow`` directives, where the most specific rule based 
+on the length of the path trumps the less specific (shorter) rule.
+
+In order to use this parser, set:
+
+* :setting:`ROBOTSTXT_PARSER` to ``scrapy.extensions.robotstxtparser.ReppyRobotParser``
```
kmike: I'm not sure scrapy.extensions is a right place for parsers, because extensions have a specific meaning - they are put into EXTENSIONS setting, created by Scrapy when spider starts, etc.; they are also expected to subscribe to signals. Here these backends are created by the middleware, similar to storages for FilePipeline, or exporters for feed exports. 

Not sure about the best place for this. `scrapy.robotstxt`?  @dangra @Gallaecio what do you think?

scrapy/downloadermiddlewares/robotstxt.py
kmike: to clarify: I think that if user put `ROBOTSTXT_PARSER="..."` which can't be loaded (e.g. because a dependency  is not installed), we should fail early, don't even start the crawl.

docs/topics/downloader-middleware.rst
anubhavp28: Currently No, but we do have plans to to make `scrapy.spiders.SitemapSpider` use this interface.

docs/topics/downloader-middleware.rst
anubhavp28: We don't actually have a `scrapy.extensions.robotstxtparser.RobotsParser` class in code. We tried to keep it in line with how currently interfaces are described in Scrapy. Like for instance, there is no `scrapy.downloadermiddlewares.DownloaderMiddleware`  class, and the interface is directly described in the documentations.

docs/topics/downloader-middleware.rst
anubhavp28: I am removing the mention of `Crawl-delay` and other directives which Scrapy does not use right now.

scrapy/downloadermiddlewares/robotstxt.py
anubhavp28: It now raises an exception.

scrapy/downloadermiddlewares/robotstxt.py
anubhavp28: `_parserimpl` is now loaded only once in `__init__`, and at the same place we are checking if any parser dependencies are unmet, if so we raise an exception.

docs/topics/downloader-middleware.rst
anubhavp28: I moved all the parser interface classes to `scrapy.robotstxt`.

tox.ini
anubhavp28: I have changed the number of environments to 2 (one for Python 3.X and one for Python 2.7).

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -25,11 +25,21 @@ class RobotsTxtMiddleware(object):
...
+        # check if parser dependencies are met
+        try:
+            self._parserimpl.from_crawler(self.crawler, b'')
+        except ImportError as e:
+            # For Python 2 compatibility
+            errmsg = e.msg if hasattr(e, 'msg') else e.message
+
+            missingmodule = re.match('No module named (.+)', errmsg).group(1).strip("'")
+            raise ImportError('Unable to use \'%s\'. Do you have \'%s\' module installed?' %
+                              (self._parserimpl.__name__, missingmodule))
```
kmike: I'd just leave it as is, to have the full original traceback :) ImportError is self-explanatory.

docs/topics/downloader-middleware.rst
Gallaecio: @anubhavp28 Maybe we can remove `sitemaps()` from this patch and, once this is merged, work on a second patch that implements this and the integration with `SitemapSpider`.

docs/topics/downloader-middleware.rst
```diff
@@ -922,13 +922,124 @@ RobotsTxtMiddleware
...
+
+.. _support-for-new-robots-parser:
+
+Implementing support for a new parser
+-------------------------------------
+
+You can implement support for a new robots.txt_ parser by implementing the
+interface described below.
+
+.. class:: scrapy.robotstxt.RobotsParser
```
Gallaecio: In regards to a comment from @kmike, maybe we should replace `scrapy.robotstxt.RobotsParser` by something like `MyRobotsParser`, to make it obvious that it is not an actual class.

docs/topics/downloader-middleware.rst
anubhavp28: Yes, I do think that will be better for both `sitemaps` and `crawl-delay` since we are not using it anywhere currently. Also, keeping these two functions in interface reference could give the impression that only parsers with support for `Crawl-delay` and `Sitemaps` can work with Scrapy.

anubhavp28: > @anubhavp28 Maybe we can remove `sitemaps()` from this patch and, once this is merged, work on a second patch that implements this and the integration with `SitemapSpider`.

Are you suggesting to remove `sitemaps` mentions from the documentation or to remove `sitemaps` function from code? I think we could just remove `sitemaps` mentions in documentation while keeping the code same.

Gallaecio: I was actually suggesting both, to keep the changes as clean as possible, i.e. do not add any useless code.

Keep the changes somewhere you can recover them easily, though, e.g. a local branch or text file somewhere. As soon as these changes are merged, it would be great for you to start working on a patch for sitemap support.

docs/topics/downloader-middleware.rst
kmike: Alternatively, make it a real abstract base class - see https://docs.python.org/2.7/library/abc.html

anubhavp28: I created an abstract base class, and moved API reference to the docstring. Also, I have removed `sitemap`, `crawl-delay` and `preferred_host` functions that Scrapy doesn't use currently.

scrapy/downloadermiddlewares/robotstxt.py
Gallaecio: @anubhavp28 has made it so that per-request user agent strings are respected.

As for the user agent middleware, I think it makes sense for the robotstxt middleware to rely on the same setting and the request header regardless of whether or not the middleware is enabled. I think this offers the most intuitive behavior when the robotstxt middleware is enabled, regardless of whether or not the user agent middleware is enabled as well.

@anubhavp28: we do need to update the documentation of the `USER_AGENT` setting to make it clear that both the user-agent middleware and this robotstxt middleware use it, and how (i.e. how it can be overiden in each case).

There is one use case I can think of that we are not covering with the current approach: a user wants to send one user-agent header value but have a different user agent string evaluated to decide which pages to crawl.

To support this, I think we could implement a `ROBOTSTXT_USER_AGENT` setting with a higher priority than both the `USER_AGENT` setting and the header of a specific request, and a `robotstxt_user_agent` meta key that allows overriding the `ROBOTSTXT_USER_AGENT` setting for specific requests.

Gallaecio:
I think the only remaining point from the feedback so far is [how to handle user agent strings](https://github.com/scrapy/scrapy/pull/3796#discussion_r300036852). Lets wait for additional input regarding that, as well as any additional feedback there might be, but the patch is looking really good!

scrapy/downloadermiddlewares/robotstxt.py
whalebot-helmsman: We already have `crawler.settings.get('USER_AGENT')` and `request.headers['User-Agent']` we should support them and add corresponding documentation.

> To support this, I think we could implement a ROBOTSTXT_USER_AGENT setting with a higher priority than both the USER_AGENT setting and the header of a specific request, and a robotstxt_user_agent meta key that allows overriding the ROBOTSTXT_USER_AGENT setting for specific requests.

I think `request.meta['robotstxt_user_agent']` should be enough here, no need in `ROBOTSTXT_USER_AGENT`

In next order:
- request.meta['robotstxt_user_agent']
-  request.headers['User-Agent']
- crawler.settings.get('USER_AGENT')

scrapy/robotstxt.py
```diff
@@ -0,0 +1,113 @@
...
+import logging
+from abc import ABCMeta, abstractmethod
+
+from scrapy.utils.python import to_native_str, to_unicode
+
+logger = logging.getLogger(__name__)
+
+
+class RobotParser:
+    __metaclass__ = ABCMeta
```
kmike: This syntax only works in Python 2, six.with_metaclass should be used if you want code to be compatible with both Python 2 and 3.

anubhavp28: I made the change.

scrapy/downloadermiddlewares/robotstxt.py
anubhavp28: @kmike What's your thoughts on this? Would `request.meta['robotstxt_user_agent']` be enough?

Gallaecio: I’m now wondering whether or not it makes sense to allow per-request user agents to affect the robots.txt parser.

The only situation I can think of is that where two sites ask crawlers to specify their user agent string in a different format, and a user wants to use the same spider for both sites. But that sounds a bit bizarre.

Gallaecio: @kmike I don’t have strong feelings against @whalebot-helmsman’s proposal. It it looks OK to you, let’s go that route.

kmike: hey! I shouldn't have asked to fix this issue in this PR in a first place, as it is unrelated to introducing robots.txt parser interface.

Previously Scrapy was only taking USER_AGENT option in account. Now support for User-Agent header is added as well, which in most cases will be the same as USER_AGENT anyways. This can be backwards incompatible in some cases, e.g. it may affect crawlers when people use rotating user-agent middlewares.

At the same time, I think the current change is good. It is polite to follow robots.txt with user agent you're sending, and I think it should be a default; using a separate user agent for robots.txt is a more aggressive option, probably similar to disabling robots.txt middleware. 

I'm fine with adding both `request.meta['robotstxt_user_agent']` and `ROBOTSTXT_USER_AGENT` options, or adding only one of them, but see it as a separate discussion, which shouldn't block merging of the current changes. User now can easily override RobotParser.allowed method and implement any logic there, we have a public documented interface for this, so this use case looks covered as well - not in a super-convenient way, but it can be fine, as it is an advanced use case.

kmike:
@Gallaecio for me it looks like an issue which can be solved separately (https://github.com/scrapy/scrapy/pull/3796#discussion_r309334667); I'm still fine with merging this PR as-is.

scrapy/downloadermiddlewares/robotstxt.py
anubhavp28: @Gallaecio Based on suggestion from @kmike, I am thinking of creating new issue and pull requests for discussing and resolving this issue. Merging this pull request will allow other work such as integrating `SitemapCrawler` with this interface to commence.

PR #3805, opened 2019-06-03
fix retry's with another proxy_authorization

ljm9104:
If one request is failed and the request uses ip proxy, in the fllowing times retry, if this times uses another ip  authorization, the request has already made request.headers.['Proxy-Authorization'] and this time we still use the original Proxy-Authorization, so I think we shouldn't judge the request.headers.get('Proxy-Authorization')  isexist or not, we should overwrite request.headers.['Proxy-Authorization'] every times,  or we should judge every time if it is isexsists and request.headers.['Proxy-Authorization'] is or not  the same with original. At last, I prefer the first  modify method.

`
# the downloadermiddlewares httpproxy
    # ignore if proxy is already set
    if 'proxy' in request.meta:
        if request.meta['proxy'] is None:
            return
        # extract credentials if present
        creds, proxy_url = self._get_proxy(request.meta['proxy'], '')
        request.meta['proxy'] = proxy_url
        if creds and not request.headers.get('Proxy-Authorization'):
            request.headers['Proxy-Authorization'] = b'Basic ' + creds
        return
    elif not self.proxies:
        return
`

### DIFF
--- a/scrapy/downloadermiddlewares/httpproxy.py
+++ b/scrapy/downloadermiddlewares/httpproxy.py
@@ -51,7 +51,7 @@ class HttpProxyMiddleware(object):
             # extract credentials if present
             creds, proxy_url = self._get_proxy(request.meta['proxy'], '')
             request.meta['proxy'] = proxy_url
-            if creds and not request.headers.get('Proxy-Authorization'):
+            if creds:
                 request.headers['Proxy-Authorization'] = b'Basic ' + creds
             return
         elif not self.proxies:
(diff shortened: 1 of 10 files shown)


Gallaecio:
Could you include a test for your change?

ljm9104:
OK, I just add one demo project for test, I  add one proxy middleware in the demo project and some logging for recording the bug log, in order to test the bug, I have set the download_timeout to 0.1s, so it usually can't recieve one response,  just as you can see, in the log file, it retries for 10 times, but every time it uses the same `request.headers[b'Proxy-Authorization'] `.

Gallaecio:
Sorry, that is not what I meant.

I mean for you to extend the automated tests of Scrapy to include one that verifies that your changes work (even though both you and I know they do), so that if someone accidentally breaks your changes in the future, your automated test detects it.

Please, see https://docs.scrapy.org/en/master/contributing.html#tests

Gallaecio:
https://github.com/scrapy/scrapy/pull/5477#issuecomment-1194028505

PR #3806, opened 2019-06-03
Fix a double indexing of the scrapy.statscollectors module in the doc…

Gallaecio:
…umentation

The issue is being reported by the CI system.


kmike:
Thanks @Gallaecio!

PR #3813, opened 2019-06-05
Fix a memory leak on the Media Pipeline (Files and Images)

victor-torres:
We're storing exceptions captured by Twisted on the media pipeline
cache, but we're also using the defer.returnValue method with our
own methods decorated with @defer.inlineCallbacks.

The defer.returnValue method passes returned values forward by
throwing a defer._DefGen_Return exception, which in its turn
extends the BaseException class and is captured by Twisted.

This way, the latest exception stored in the Failure's object may
also have an HtmlResponse object in its `__context__` attribute. As
the Response object also keeps track of the Request object that
has originated it, you could figure it out how many RAM we're
wasting here.

This could easily lead to a Memory Leak problem when running
spiders with Media Pipeline enabled and a particular Request set
that tends to raise a significant number of exceptions.

Example triggers:
- media requests with 404 status responses
- userland exceptions coming from custom middlewares
- etc.

### DIFF
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -140,7 +140,22 @@ class MediaPipeline(object):
             result.frames = []
             result.stack = None
 
-            # See twisted.internet.defer.returnValue docstring
+            # This code fixes a memory leak by avoiding to keep references to
+            # the Request and Response objects on the Media Pipeline cache.
+            #
+            # Twisted inline callbacks pass return values using the function
+            # twisted.internet.defer.returnValue, which encapsulates the return
+            # value inside a _DefGen_Return base exception.
+            #
+            # What happens when the media_downloaded callback raises another
+            # exception, for example a FileException('download-error') when
+            # the Response status code is not 200 OK, is that it stores the
+            # _DefGen_Return exception on the FileException context.
+            #
+            # To avoid keeping references to the Response and therefore Request
+            # objects on the Media Pipeline cache, we should wipe the context of
+            # the exception encapsulated by the Twisted Failure when its a
+            # _DefGen_Return instance.
             if isinstance(result.value.__context__, _DefGen_Return):
                 result.value.__context__ = None
 
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -156,7 +156,10 @@ class MediaPipeline(object):
             # objects on the Media Pipeline cache, we should wipe the context of
             # the exception encapsulated by the Twisted Failure when its a
             # _DefGen_Return instance.
-            if isinstance(result.value.__context__, _DefGen_Return):
+            #
+            # This problem does not occur in Python 2.7 since we don't have
+            # Exception Chaining (https://www.python.org/dev/peps/pep-3134/).
+            if isinstance(getattr(result.value, '__context__'), _DefGen_Return):
                 result.value.__context__ = None
 
         info.downloading.remove(fp)
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 import functools
 import logging
 from collections import defaultdict
-from twisted.internet.defer import Deferred, DeferredList
+from twisted.internet.defer import Deferred, DeferredList, _DefGen_Return
 from twisted.python.failure import Failure
 
 from scrapy.settings import Settings
@@ -139,6 +139,11 @@ class MediaPipeline(object):
             result.cleanFailure()
             result.frames = []
             result.stack = None
+
+            # See twisted.internet.defer.returnValue docstring
+            if isinstance(result.value.__context__, _DefGen_Return):
+                result.value.__context__ = None
+
         info.downloading.remove(fp)
         info.downloaded[fp] = result  # cache result
         for wad in info.waiting.pop(fp):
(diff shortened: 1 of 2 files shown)


victor-torres:
**Before this patch**

You can see lots of `HtmlResponse` and `Request` objects being held for no reason. They continue in the memory until the spider ends its execution or we're stopped by a `memusage_exceeded` reason. The oldest object of those types continues increasing their elapsed seconds as the spider goes. It's counting one thousand `Requests` here just after 5 minutes running.

```
>>> prefs()            
                                   
Live References                                           
                                                          
HtmlResponse                      371   oldest: 328s ago
MySpider                            1   oldest: 333s ago
Request                          1000   oldest: 329s ago
Selector                           54   oldest: 65s ago 
TextResponse                      579   oldest: 304s ago
TradesyItem                        12   oldest: 14s ago
```

**After this patch**

With 50 concurrent requests, we're holding only 62 `Request` objects and 12 `HtmlResponse` objects in memory. Also, the oldest ones were created within the last minute, even the spider being running for 30 minutes.

```
>>> prefs()
                                                                                                                                                                         
Live References                                                                                                                                                                     
                                                                                                                                                                                    
HtmlResponse                       12   oldest: 22s ago                                                                                                                             
MySpider                            1   oldest: 1822s ago                                                                                                                           
Request                            62   oldest: 62s ago                                                                                                                             
Selector                           12   oldest: 22s ago                                                                                                                             
```

scrapy/pipelines/media.py
```diff
@@ -139,6 +139,11 @@ def _cache_result_and_execute_waiters(self, result, fp, info):
             result.cleanFailure()
             result.frames = []
             result.stack = None
+
+            # See twisted.internet.defer.returnValue docstring
```
ejulio: Since there are no tests, at least a comment mentioning it fixes a memory leak

victor-torres: The commit message has the same description as this pull request. 
I guess we're good checking the commit message for more details.

victor-torres:
@ejulio, I've tried to write a simple test case just to cover the regression of the issue. 

The ideal scenario would be to count references to the Request and Response objects after executing the cache function, but that would involve some difficulty. I've tried to do that but ended up giving up after dealing with some magic properties from Pypy (example `_rawexcinfo`) and Twisted's version of unittest.

scrapy/pipelines/media.py
dangra: `_DefGen_Return` is private and very inner detail of inlineCallbacks implementation, I think it deserves a long comment explaining what it tries to prevents.

tests/test_pipeline_media.py
```diff
@@ -90,6 +91,30 @@ def test_modify_media_request(self):
         self.pipe._modify_media_request(request)
         assert request.meta == {'handle_httpstatus_all': True}
 
+    def test_cache_result(self):
+        request = Request('http://url')
+        response = Response('http://url', body=b'', request=request)
+
```
dangra: This test is also very synthetic. We should add a comment here explaining the lines below are intended to prepare a specific Failure object that holds references to Response/Request objects in its context.

The test docstring should make a reference to the memory leak it is trying to prevent.

victor-torres:
@dangra, thanks for your feedback. I've just updated the source code renaming the test case and adding a proper docstring to it. I've included additional comments between the source code lines and improved the comment on the cache function as well. Let me know if you have an additional suggestion.

dangra:
Although the code is not Python 2.7 compatible.

tests/test_pipeline_media.py
```diff
@@ -129,7 +129,7 @@ def test_should_remove_req_res_references_before_caching_the_results(self):
             try:
                 # Simulate the media_downloaded callback raising a FileException
                 # This usually happens when the status code is not 200 OK
-                raise FileException('download-error') from def_gen_return_exc
+                raise FileException('download-error')
```
dangra: I think you can use [six.raise_from](https://six.readthedocs.io/index.html?highlight=raise_from#six.raise_from) instead

victor-torres: Thanks, but I believe it's not needed. According to [PEP 3134](https://www.python.org/dev/peps/pep-3134/), the `__context__` attribute is automatically set when a chained exception is raised within a try/except block.

I've also made additional changes on Python 2.7 failures that had been suppressed by the huge amount of failing tests.

victor-torres:
@dangra, thank you for the feedback regarding Python 2.7 tests. We should be good to go now.

tests/test_pipeline_media.py
dangra: fair point 👍

scrapy/pipelines/media.py
```diff
@@ -139,6 +139,30 @@ def _cache_result_and_execute_waiters(self, result, fp, info):
...
+            # To avoid keeping references to the Response and therefore Request
+            # objects on the Media Pipeline cache, we should wipe the context of
+            # the exception encapsulated by the Twisted Failure when its a
+            # _DefGen_Return instance.
+            #
+            # This problem does not occur in Python 2.7 since we don't have
+            # Exception Chaining (https://www.python.org/dev/peps/pep-3134/).
+            context = getattr(result.value, '__context__', None)
+            if isinstance(context, _DefGen_Return):
+                setattr(result.value, '__context__', None)
```
dangra: curious why did you use `setattr()` instead of  `result.value.__context__ = None`

victor-torres: No particular reason. Just copied from the line above and replaced `get` with `set`. Would you like me to change it?

Gallaecio:
Great work!

Issue #2997, opened 2017-11-08
filepipeline handle urls with "?" wrong


NewUserHa:
like http://www.hebpr.cn/WebbuilderMIS/attach/attachdown.jspx?attachGuid=c986d80b-4d9d-4a50-950f-fea61b828dda&dataSourceName=EpointHbztbCMS

and how can I only replace file_path method of FilesPipeline to make my custom pipeline work?
I googled many times but no answer

thanks

NewUserHa:
and is there an easy way in settings.py to let every spider within one project write log to disk with the file name like '%spidername%.log' instead of adding a custom setting for every spider?

iAnanich:
@NewUserHa you can write a custom subclass of the `Spider` and override `from_crawler` method, to set specific setting using your spider's name.

NewUserHa:
could you show me some examples please?

NewUserHa:
@iAnanich

raphapassini:
Hey @NewUserHa you can simple use the ``custom_settings`` inside your spider

```
class PythonOrgSpider(scrapy.Spider):
    name = 'python_org'
    allowed_domains = ['python.org']
    start_urls = ['http://python.org']
    custom_settings = {
        'LOG_FILE': name + '.txt'
    }
```

The code above will generate a file called ``python_org.txt`` when you run the crawler :)

NewUserHa:
hi, thanks.
what I want to know is that is there any way put a setting into settings to reach this?

raphapassini:
No, there isn't a _setting_ that you can put in ``settings.py`` to do so.

But I think you can use a custom ``SpiderLoader`` to achieve this , please, take a look at: https://doc.scrapy.org/en/latest/topics/api.html#module-scrapy.loader

NewUserHa:
thanks! that seems what I want since it's annoying to put the same code in every spider

iAnanich:
@NewUserHa it's normal to copy-paste some code. But if it is too annoying for you, you can use class factory (metaclass).

raphapassini:
Hey @kmike I think the title of this issue should be changed to reflect the real problem of the user and to help others to find the same solution easily :)

kmike:
@raphapassini what title do you suggest? I can change it.

As for the ticket itself, I'm closing it; we're trying to use Scrapy issue tracker for bugs and concrete enhancement proposals; for general Scrapy help there are better resources like https://stackoverflow.com/ (use tag "scrapy").

[...]

Issue #1199, opened 2015-04-30
FilesPipeline: Optionally parse media extension from response headers if provided.


ldgarcia:
I needed to download some autogenerated datasheets (in PDF format), and I decided to go with the FilesPipeline provided with Scrapy.

However, the URL structure didn't follow the assumption of the `file_path` method in the `FilesPipeline` class (a well structured filename, i.e. ````example.pdf````):

spider.py
````python
item['file_urls'] = []
item['file_urls'].append("https://www.domain.name/path/path/DatasheetService?control=%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22%3F%3E%3Cpdf_generator_control%3E%3Cmode%3EPDF%3C%2Fmode%3E%3Cpdmsystem%3EPMD%3C%2Fpdmsystem%3E%3Ctemplate_selection+mlfb%3D%22{}%22+system%3D%22PRODIS%22%2F%3E%3Clanguage%3Ees%3C%2Flanguage%3E%3Ccaller%3EMall%3C%2Fcaller%3E%3C%2Fpdf_generator_control%3E".format(item['code']))
````
scrapy.contrib.pipeline.files
````python
media_ext = os.path.splitext(url)[1]  # change to request.url after deprecation
````

The value of ```media_ext``` for this URL is
````txt
.0%22+encoding%3D%22UTF-8%22%3F%3E%3Cpdf_generator_control%3E%3Cmode%3EPDF%3C%2Fmode%3E%3Cpdmsystem%3EPMD%3C%2Fpdmsystem%3E%3Ctemplate_selection+mlfb%3D%22CODE%22+system%3D%22PRODIS%22%2F%3E%3Clanguage%3Ees%3C%2Flanguage%3E%3Ccaller%3EMall%3C%2Fcaller%3E%3C%2Fpdf_generator_control%3E
````

Which is not only an incorrect extension type, but also results (in this particular case) in a IOException (filename too long).

One quick way to fix this is appending something like ```&type=.pdf``` to the URL while constructing the ```file_urls``` list:

````python
item['file_urls'] = []
[...]

PR #3817, opened 2019-06-07
Fixing FilePipeline file extensions

phrfpeixoto:
This is a proposed fix for #2997 and #1199, since #2809 has not been merged or updated in a some time now.

I've also seen @idgarcia's [approach](https://github.com/ldgarcia/scrapy/commit/a01f24900b3a454f4b09f594a6d529da14394367), but in using mimetype, but in my tests that produced some uncommon extensions, such as `.jpe` for `image/jpeg` content type. 

I do can implement a control flag if you deem necessary.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -24,7 +24,7 @@ from twisted.internet import defer, threads
 from scrapy.pipelines.media import MediaPipeline
 from scrapy.settings import Settings
 from scrapy.exceptions import NotConfigured, IgnoreRequest
-from scrapy.http import Request
+from scrapy.http import Request, Response
 from scrapy.utils.misc import md5sum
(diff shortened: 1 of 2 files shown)


scrapy/pipelines/files.py
```diff
@@ -480,9 +480,38 @@ def _warn():
...
+                * http://www.example.com/media.do?filename=image.jpg yields '.do'
+                * http://www.example.com/image yields an empty string
+        """
+
+        filename = urlparse(url).path[1:]
+        content_type = response.headers.get('Content-Type', b'')
+        content_disposition = response.headers.get('Content-Disposition', b'')
+
+        if content_type.startswith(b'application/octet-stream '):
+            return content_type[25:]
```
Gallaecio: What will happen if `Content-Type` is `'application/octet-stream /../../../../../asdf'`?

PR #3820, opened 2019-06-07
Added pathlib.Path support for FEED_URI - Fix for #3753

somaran:
- [ ] Added Support for pathlib.Path 
- [ ] Added test to cover the usecase

### DIFF
--- a/requirements-py2.txt
+++ b/requirements-py2.txt
@@ -8,3 +8,4 @@ six>=1.5.2
 PyDispatcher>=2.0.5
 parsel>=1.5
 service_identity
+pathlib2
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -11,6 +11,11 @@ import posixpath
 from tempfile import NamedTemporaryFile
 from datetime import datetime
 import six
+
+if six.PY2:
+    from pathlib2 import Path
+else:
+    from pathlib import Path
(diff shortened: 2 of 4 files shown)


Gallaecio:
Hmm… Was there something wrong with the approach followed at #3753? (just casting the setting value with `str`)

somaran:
well, yes. when you type cast the value, it doesn't resolve back to file uri format. 
`file://` is the expected format for pathlib i believe.

Gallaecio:
I see. I thought local files did not need it, but according to the documentation that’s only when paths are absolute, so indeed a relative `Path` object would fail with the `str`-casting approach.

The suggested approach, however, will make Scrapy stop working if users do not install `pathlib2`. We could add `pathlib2` to `setup.py`, however, I’m not sure whether or not it makes sense to add a Python 2-specific dependency to Scrapy at this point, so close to 2020 (Python 2’s end of life), just for this feature.

@kmike, @dangra Any thoughts?

somaran:
@Gallaecio May be we should let the support be in till Jan 2020 ? @kmike  @dangra Any suggestions ?

kmike:
I think we're going to drop Python 2 support in master after 1.7 release (which is hopefully around the corner), so no need to worry about pathlib2.

wRAR:
If this is still useful it needs a py2 cleanup

wRAR:
This was actually implemented in #4074

PR #3822, opened 2019-06-09
Fixes Python 3.4 support incompatibility with twisted, Fixes #3821

### LINKED ISSUE #3821: Issue with Twisted and Python 3.4

somaran:
Twisted doesn't support Python 3.4 at the moment and the test suite fails because of the imports. 
- Fixes Py34 test suite failure


somaran:
This pull request fixes

PR #3824, opened 2019-06-11
Run Python 3.4 tests with Twisted 19.2.0

### LINKED ISSUE #3821: Issue with Twisted and Python 3.4

Gallaecio:
Twisted 19.2.0 is the last Twisted version with Python 3.4 support.

Fixes #3821


Gallaecio:
#3825 seems to do the job.

PR #3825, opened 2019-06-11
Require Twisted<=19.2.0 for Python 3.4

### LINKED ISSUE #3821: Issue with Twisted and Python 3.4

Gallaecio:
[Twisted 19.2.0 is the final release with Python 3.4 support](https://github.com/twisted/twisted/blob/twisted-19.2.0/NEWS.rst).

These changes are based on the [setuptools documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies) and [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers).

Fixes #3821

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,8 @@ setup(
     ],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     install_requires=[
-        'Twisted>=13.1.0',
+        'Twisted>=13.1.0;python_version!="3.4"',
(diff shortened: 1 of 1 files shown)


victor-torres:
I'd say to add some test cases for py3.4. Go for it if you have some time. 
Anyway, seems like we should be dropping Python 3.4 support soon.

dangra:
> Anyway, seems like we should be dropping Python 3.4 support soon.

Let's drop Python 3.4 in next release.

PR #3829, opened 2019-06-14
[MRG+1] Add a setting to use active mode in FTPFeedStorage

heylouiz:
FTP servers can be configured in active and passive mode, by default
Python and Scrapy use passive mode and there was no way to use active.

This commit addes a setting FEED_STORAGE_FTP_ACTIVE to allow a feed
to be exported to a FTP server configured in active mode.


scrapy/extensions/feedexport.py
```diff
@@ -156,19 +156,29 @@ def _store_in_thread(self, file):
...
         self.username = u.username
         self.password = u.password
         self.path = u.path
+        self.active = active
+
+    @classmethod
+    def from_crawler(cls, crawler, uri):
+        return cls(
+            uri=uri,
+            active=crawler.settings['FEED_STORAGE_FTP_ACTIVE'] or False
```
dangra: Please, use `.getbool()` method as documented at [scrapy.settings.BaseSettings.getbool](https://docs.scrapy.org/en/latest/topics/api.html#scrapy.settings.BaseSettings.getbool)

scrapy/extensions/feedexport.py
```diff
@@ -156,19 +156,29 @@ def _store_in_thread(self, file):
 
 class FTPFeedStorage(BlockingFeedStorage):
 
-    def __init__(self, uri):
+    def __init__(self, uri, active=False):
```
dangra: I think `use_active_mode` is less likely to be miss-interpreted than `active` alone.

scrapy/extensions/feedexport.py
```diff
@@ -156,19 +156,29 @@ def _store_in_thread(self, file):
...
+            active=crawler.settings['FEED_STORAGE_FTP_ACTIVE'] or False
+        )
 
     def _store_in_thread(self, file):
         file.seek(0)
         ftp = FTP()
         ftp.connect(self.host, self.port)
         ftp.login(self.username, self.password)
+        if self.active:
+            ftp.set_pasv(False)
```
dangra: Any chances of adding a test case to cover this line?

scrapy/extensions/feedexport.py
heylouiz: Ok, I was going to use it but since this is the way settings are accessed in this file I decided to follow the pattern.
Will change.

scrapy/extensions/feedexport.py
heylouiz: Agreed, will change.

scrapy/extensions/feedexport.py
heylouiz: Sure @dangra but I am not sure what should I use in [FEEDTEST_FTP_URI and FEEDTEST_FTP_PATH](https://github.com/scrapy/scrapy/blob/4196d4869b8c5c14c9db19290d875fdd9958265d/tests/test_feedexport.py#L74). Should I add the unit test, start two FTP servers on my machine and test both passive and active method? Or Scrapy has a server to test it?

heylouiz:
Hey @dangra, I made the changes you requested, can you please review again?

I had to instantiate the test spider using from_crawler because the test was failing in https://github.com/scrapy/scrapy/pull/3829/files#diff-b40c5755a5dda2b02aaac6bf4c8e2ff8R50
```AttributeError: 'Spider' object has no attribute 'crawler'```

File: docs/topics/feed-exports.rst [hunk 4]
```diff
@@ -164,6 +164,10 @@ The feeds are stored in a FTP server.
  * Example URI: ``ftp://user:pass@ftp.example.com/path/to/export.csv``
  * Required external libraries: none
 
+By default transfers are made in passive mode, you use active mode using this setting:
```
Gallaecio: I wonder if we should extend this for people only familiar with the basics of FTP that will learn about the active and passive FTP connection modes from our documentation.

For example:

```
FTP supports two different connection modes: [active or passive](
https://stackoverflow.com/a/1699163). Scrapy uses the passive connection mode
by default. To use the active connection mode instead, set the
:setting:`FEED_STORAGE_FTP_ACTIVE` setting to ``True``.
```

We should probably also include [that Stackoverflow link](https://stackoverflow.com/a/1699163) (or other link that better explains FTP connection modes) in the documentation of the setting itself.

heylouiz: @Gallaecio  I agree with you, I didn't know about these modes, my project started to fail after the FTP server restarted and I had to try everything till I found and solved locally.
I will update the docs.

docs/topics/feed-exports.rst
```diff
@@ -164,7 +164,10 @@ The feeds are stored in a FTP server.
  * Example URI: ``ftp://user:pass@ftp.example.com/path/to/export.csv``
  * Required external libraries: none
 
-By default transfers are made in passive mode, you use active mode using this setting:
+FTP supports two different connection modes: [active or passive](
+https://stackoverflow.com/a/1699163). Scrapy uses the passive connection mode
+by default. To use the active connection mode instead, set the
+:setting:`FEED_STORAGE_FTP_ACTIVE` setting to ``True``.
 
  * :setting:`FEED_STORAGE_FTP_ACTIVE`
```
Gallaecio: I think this line becomes unnecessary given the paragraph above.

docs/topics/feed-exports.rst
```diff
@@ -307,6 +315,15 @@ Default: ``{}``
...
 The keys are URI schemes and the values are paths to storage classes.
 
+.. setting:: FEED_STORAGE_FTP_ACTIVE
+
+FEED_STORAGE_FTP_ACTIVE
+-----------------------
+
+Default: ``False``
+
+Whether to use active mode when exporting feeds to a FTP server.
```
Gallaecio: Maybe "active mode" here should also link to https://stackoverflow.com/a/1699163 ?

docs/topics/feed-exports.rst
heylouiz: Whoops..

PR #3835, opened 2019-06-17
Per request delay implementation using request per_request_delay meta key #802

GeorgeA92:
Changes in this PR provide possibility to set per request delays (as it explained in this https://github.com/scrapy/scrapy/issues/802#issuecomment-500245345)

     ....
        next =  response.css("li.next a ::attr(href)").extract_first()
        if next:
            yield scrapy.Request(url=response.urljoin(next),
                                 callback=self.parse, meta={"per_request_delay": 3})

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -34,6 +34,9 @@ class Slot(object):
         return self.concurrency - len(self.transferring)
 
     def download_delay(self):
+        if self.queue:
+            if "per_request_delay" in self.queue[0][0].meta.keys():
+                return self.queue[0][0].meta["per_request_delay"]
         if self.randomize_delay:
             return random.uniform(0.5 * self.delay, 1.5 * self.delay)
         return self.delay


scrapy/core/downloader/__init__.py
```diff
@@ -34,6 +34,9 @@ def free_transfer_slots(self):
         return self.concurrency - len(self.transferring)
 
     def download_delay(self):
+        if self.queue:
```
realslimshanky, line 37: This adds delay to the entire slot and not just one request.
Taking an example of this spider
```
import scrapy


    name = "quotes"

    def start_requests(self):
        urls = [
            'http://quotes.toscrape.com/page/1/',
            'http://quotes.toscrape.com/page/2/',
            'http://quotes.toscrape.com/page/3/',
            'http://quotes.toscrape.com/page/4/',
            'http://quotes.toscrape.com/page/5/',
            'http://quotes.toscrape.com/page/6/',
            'http://quotes.toscrape.com/page/7/',
            'http://quotes.toscrape.com/page/8/',
            'http://quotes.toscrape.com/page/9/',
            'http://quotes.toscrape.com/page/10/',
        ]
        for index, url in enumerate(urls, 1):
            if index == 1 :
                yield scrapy.Request(url=url, callback=self.parse, meta={'per_request_delay': 5})
            if index == 2 :
                yield scrapy.Request(url=url, callback=self.parse, meta={'per_request_delay': 8})
            else:

    def parse(self, response):
        print(f'Parsed {response.url}')
```
if you run `$ scrapy crawl quotes`, you'll see the order of download as 
```
2021-02-08 16:56:41 [scrapy.core.engine] DEBUG: Crawled (404) <GET http://quotes.toscrape.com/robots.txt> (referer: None)
2021-02-08 16:56:46 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/1/> (referer: None)
Parsed http://quotes.toscrape.com/page/1/
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/2/> (referer: None)
Parsed http://quotes.toscrape.com/page/2/
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/7/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/4/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/5/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/3/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/9/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/6/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/8/> (referer: None)
2021-02-08 16:56:54 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/10/> (referer: None)
```
As you can see the delay was added to the entire slot. But we are trying to delay individual requests right?

GeorgeA92: One of main reasons for per request delay feature request - is to provide way add additional delay to handle various cases like temp ban,  exponential backoff for the retry request, huge delay when site maintenance is detected from request callback https://github.com/scrapy/scrapy/issues/802#issue-37790671.

If for example website gave temp 1 hour ban - it is not expected scrapy app will send requests during this period.
So in fact it is required to implement delay on `downloader.slot` basis.

realslimshanky: There are two ways to handle this, either delay the entire slot or delay just the individual request. I was looking for a way to apply delay on individual requests. and with your explanation, I understand that this should be called `per_slot_delay` or `slot_delay` instead. What do you think about this?

GeorgeA92: Cases like:
to add some huge delay when site maintenance is detected from request callback
to add delay in case if temp ban detected from request callback (this PR mostly aimed to solve these cases).
can't be solved by individual delay (only slot delay can be appliable for these cases).

For Cases like exponential/other backoff for the retry request (if you want to retry some specific request in another delay without breaking remaining downloader slot queue sequence).
I would create another downloader slot for that specific request with is's `lastseen` variable set to `now() + delay.

I think that any of these things are achievable in current versions of scrapy by modifying existing downloader slots.

And I don't think that quality/quantity of current solutions are enough to decide what exactly we need to add to scrapy

realslimshanky: @GeorgeA92 Totally agree. After testing this PR more delaying slot makes more sense. I wonder if delaying individual requests can be used in a real world scenario, well only time will tell. ;)

PR #3858, opened 2019-07-07
[MRG+1] Support for exporting to multiple feeds in a single crawl

### LINKED ISSUE #1336: Support multiple exporters when crawling
I'm scraping a website to export the data into a semantic format (n3). However, I also want to perform some data analysis on that data, so having it in a csv format is more convenient. To get the data in both formats I can do.

```
scrapy spider -t n3 -o data.n3
scrapy spider -t csv -o data.csv
```

However, this scrapes the data twice and I cannot afford it with big amounts of data.

A solution that avoids scraping the data twice consists on implementing Pipeline that exports the data (see [alecxe suggestion](http://stackoverflow.com/questions/31032340/export-scraping-data-in-multiple-formats-using-scrapy/31033769#31033769) for details). However, as the documentation explains, this is not the preferred way to export data.

Thus, I consider it would be interesting scrappy's support for multiple exporters.

```
scrapy crawl <url> -t n3 -o data.n3 -t csv -o data.csv
```

elacuesta:
Fixes #1336

Introduced a new setting (`FEEDS`) and deprecated `FEED_URI` and `FEED_FORMAT`. `FEEDS` is a dictionary of feed definitions, which enables the user to override certain parameters independently for each feed. An example of this can be seen at https://github.com/scrapy/scrapy/pull/3858#issuecomment-511924714
In the CLI, format is deduced according to https://github.com/scrapy/scrapy/issues/1336#issuecomment-125750712, falling back to the feed extension.

A simple snippet to illustrate:

```python
import scrapy

    name = 'example'
    start_urls = ['https://example.org']

    def parse(self, response):
        return {
            'url': response.url,
            'link': response.xpath('//a/@href').get(),
        }
```

```
$ scrapy runspider example.py -o items.dat:jsonlines -o jitems.json -o -:csv
(...)
2019-07-07 02:09:45 [scrapy.core.scraper] DEBUG: Scraped from <200 https://example.org>
{'url': 'https://example.org', 'link': 'http://www.iana.org/domains/example'}
2019-07-07 02:09:45 [scrapy.core.engine] INFO: Closing spider (finished)
2019-07-07 02:09:45 [scrapy.extensions.feedexport] INFO: Stored jsonlines feed (1 items) in: items.dat
2019-07-07 02:09:45 [scrapy.extensions.feedexport] INFO: Stored json feed (1 items) in: jitems.json
2019-07-07 02:09:45 [scrapy.extensions.feedexport] INFO: Stored csv feed (1 items) in: stdout:
2019-07-07 02:09:45 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
(...)
2019-07-07 02:09:45 [scrapy.core.engine] INFO: Spider closed (finished)
url,link
https://example.org,http://www.iana.org/domains/example
```

**Tasks:**
- [x] Feed exporter changes
- [x] Crawl and Runspider commands
- [x] Tests
- [x] Docs

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -12,7 +12,7 @@ generating an "export file" with the scraped data (commonly called "export
 feed") to be consumed by other systems.
 
 Scrapy provides this functionality out of the box with the Feed Exports, which
-allows you to generate a feed with the scraped items, using multiple
+allows you to generate feeds with the scraped items, using multiple
 serialization formats and storage backends.
 
 .. _topics-feed-format:
@@ -36,7 +36,7 @@ But you can also extend the supported format through the
 JSON
 ----
 
- * :setting:`FEED_FORMAT`: ``json``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``json``
  * Exporter used: :class:`~scrapy.exporters.JsonItemExporter`
  * See :ref:`this warning <json-with-large-data>` if you're using JSON with
    large feeds.
@@ -46,7 +46,7 @@ JSON
 JSON lines
 ----------
 
- * :setting:`FEED_FORMAT`: ``jsonlines``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``jsonlines``
  * Exporter used: :class:`~scrapy.exporters.JsonLinesItemExporter`
 
 .. _topics-feed-format-csv:
@@ -54,7 +54,7 @@ JSON lines
 CSV
 ---
 
- * :setting:`FEED_FORMAT`: ``csv``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``csv``
  * Exporter used: :class:`~scrapy.exporters.CsvItemExporter`
  * To specify columns to export and their order use
    :setting:`FEED_EXPORT_FIELDS`. Other feed exporters can also use this
@@ -66,7 +66,7 @@ CSV
 XML
 ---
 
- * :setting:`FEED_FORMAT`: ``xml``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``xml``
  * Exporter used: :class:`~scrapy.exporters.XmlItemExporter`
 
 .. _topics-feed-format-pickle:
@@ -74,7 +74,7 @@ XML
 Pickle
 ------
 
- * :setting:`FEED_FORMAT`: ``pickle``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``pickle``
  * Exporter used: :class:`~scrapy.exporters.PickleItemExporter`
 
 .. _topics-feed-format-marshal:
@@ -82,7 +82,7 @@ Pickle
 Marshal
 -------
 
- * :setting:`FEED_FORMAT`: ``marshal``
+ * Value for the ``format`` key in the :setting:`FEEDS` setting: ``marshal``
  * Exporter used: :class:`~scrapy.exporters.MarshalItemExporter`
 
 
@@ -91,8 +91,8 @@ Marshal
 Storages
 ========
 
-When using the feed exports you define where to store the feed using a URI_
-(through the :setting:`FEED_URI` setting). The feed exports supports multiple
+When using the feed exports you define where to store the feed using one or multiple URIs_
+(through the :setting:`FEEDS` setting). The feed exports supports multiple
 storage backend types which are defined by the URI scheme.
 
 The storages backends supported out of the box are:
@@ -211,41 +211,66 @@ Settings
 
 These are the settings used for configuring the feed exports:
 
- * :setting:`FEED_URI` (mandatory)
- * :setting:`FEED_FORMAT`
+ * :setting:`FEEDS` (mandatory)
+ * :setting:`FEED_EXPORT_ENCODING`
+ * :setting:`FEED_STORE_EMPTY`
+ * :setting:`FEED_EXPORT_FIELDS`
+ * :setting:`FEED_EXPORT_INDENT`
  * :setting:`FEED_STORAGES`
  * :setting:`FEED_STORAGE_FTP_ACTIVE`
  * :setting:`FEED_STORAGE_S3_ACL`
  * :setting:`FEED_EXPORTERS`
- * :setting:`FEED_STORE_EMPTY`
- * :setting:`FEED_EXPORT_ENCODING`
- * :setting:`FEED_EXPORT_FIELDS`
- * :setting:`FEED_EXPORT_INDENT`
 
 .. currentmodule:: scrapy.extensions.feedexport
 
-.. setting:: FEED_URI
+.. setting:: FEEDS
 
-FEED_URI
---------
+FEEDS
+-----
 
-Default: ``None``
+.. versionadded:: 2.1
 
-The URI of the export feed. See :ref:`topics-feed-storage-backends` for
-supported URI schemes.
+Default: ``{}``
 
-This setting is required for enabling the feed exports.
+A dictionary in which every key is a feed URI (or a :class:`pathlib.Path`
+object) and each value is a nested dictionary containing configuration
+parameters for the specific feed.
+This setting is required for enabling the feed export feature.
 
-.. versionchanged:: 2.0
-   Added :class:`pathlib.Path` support.
+See :ref:`topics-feed-storage-backends` for supported URI schemes.
 
-.. setting:: FEED_FORMAT
+For instance::
 
-FEED_FORMAT
------------
+    {
+        'items.json': {
+            'format': 'json',
+            'encoding': 'utf8',
+            'store_empty': False,
+            'fields': None,
+            'indent': 4,
+        }, 
+        'items.xml': {
+            'format': 'xml',
+            'fields': ['name', 'price'],
+            'encoding': 'latin1',
+            'indent': 8,
+        },
+        pathlib.Path('items.csv'): {
+            'format': 'csv',
+            'fields': ['price', 'name'],
+        },
+    }
 
-The serialization format to be used for the feed. See
-:ref:`topics-feed-format` for possible values.
+The following is a list of the accepted keys and the setting that is used
+as a fallback value if that key is not provided for a specific feed definition.
+
+* ``format``: the serialization format to be used for the feed.
+  See :ref:`topics-feed-format` for possible values. 
+  Mandatory, no fallback setting
+* ``encoding``: falls back to :setting:`FEED_EXPORT_ENCODING`
+* ``fields``: falls back to :setting:`FEED_EXPORT_FIELDS`
+* ``indent``: falls back to :setting:`FEED_EXPORT_INDENT`
+* ``store_empty``: falls back to :setting:`FEED_STORE_EMPTY`
 
 .. setting:: FEED_EXPORT_ENCODING
 
@@ -400,7 +425,7 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
         'csv': None,
     }
 
-.. _URI: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
+.. _URIs: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
 .. _Amazon S3: https://aws.amazon.com/s3/
 .. _botocore: https://github.com/boto/botocore
 .. _Canned ACL: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -249,7 +249,7 @@ For instance::
             'fields': None,
             'indent': 4,
         }, 
-        'items.xml': {
+        '/home/user/documents/items.xml': {
             'format': 'xml',
             'fields': ['name', 'price'],
             'encoding': 'latin1',
--- a/scrapy/commands/crawl.py
+++ b/scrapy/commands/crawl.py
@@ -1,7 +1,5 @@
-import os
 from scrapy.commands import ScrapyCommand
-from scrapy.utils.conf import arglist_to_dict
-from scrapy.utils.python import without_none_values
+from scrapy.utils.conf import arglist_to_dict, feed_process_params_from_cli
 from scrapy.exceptions import UsageError
 
 
@@ -19,7 +17,7 @@ class Command(ScrapyCommand):
         ScrapyCommand.add_options(self, parser)
         parser.add_option("-a", dest="spargs", action="append", default=[], metavar="NAME=VALUE",
                           help="set spider argument (may be repeated)")
-        parser.add_option("-o", "--output", metavar="FILE",
+        parser.add_option("-o", "--output", metavar="FILE", action="append",
                           help="dump scraped items into FILE (use - for stdout)")
         parser.add_option("-t", "--output-format", metavar="FORMAT",
                           help="format to use for dumping items with -o")
@@ -31,21 +29,8 @@ class Command(ScrapyCommand):
         except ValueError:
             raise UsageError("Invalid -a value, use -a NAME=VALUE", print_help=False)
         if opts.output:
-            if opts.output == '-':
-                self.settings.set('FEED_URI', 'stdout:', priority='cmdline')
-            else:
-                self.settings.set('FEED_URI', opts.output, priority='cmdline')
-            feed_exporters = without_none_values(
-                self.settings.getwithbase('FEED_EXPORTERS'))
-            valid_output_formats = feed_exporters.keys()
-            if not opts.output_format:
-                opts.output_format = os.path.splitext(opts.output)[1].replace(".", "")
-            if opts.output_format not in valid_output_formats:
-                raise UsageError("Unrecognized output format '%s', set one"
-                                 " using the '-t' switch or as a file extension"
-                                 " from the supported list %s" % (opts.output_format,
-                                                                  tuple(valid_output_formats)))
-            self.settings.set('FEED_FORMAT', opts.output_format, priority='cmdline')
+            feeds = feed_process_params_from_cli(self.settings, opts.output, opts.output_format)
+            self.settings.set('FEEDS', feeds, priority='cmdline')
 
     def run(self, args, opts):
         if len(args) < 1:
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -4,24 +4,27 @@ Feed Exports extension
 See documentation in docs/topics/feed-exports.rst
 """
 
+import logging
 import os
 import sys
-import logging
-from tempfile import NamedTemporaryFile
+import warnings
 from datetime import datetime
-from urllib.parse import urlparse, unquote
+from tempfile import NamedTemporaryFile
+from urllib.parse import unquote, urlparse
 
-from zope.interface import Interface, implementer
 from twisted.internet import defer, threads
 from w3lib.url import file_uri_to_path
+from zope.interface import implementer, Interface
 
 from scrapy import signals
-from scrapy.utils.ftp import ftp_store_file
-from scrapy.exceptions import NotConfigured
-from scrapy.utils.misc import create_instance, load_object
-from scrapy.utils.log import failure_to_exc_info
-from scrapy.utils.python import without_none_values
+from scrapy.exceptions import NotConfigured, ScrapyDeprecationWarning
 from scrapy.utils.boto import is_botocore
+from scrapy.utils.conf import feed_complete_default_values_from_settings
+from scrapy.utils.ftp import ftp_store_file
+from scrapy.utils.log import failure_to_exc_info
+from scrapy.utils.misc import create_instance, load_object
+from scrapy.utils.python import without_none_values
+
 
 logger = logging.getLogger(__name__)
 
@@ -98,8 +101,6 @@ class S3FeedStorage(BlockingFeedStorage):
             from scrapy.utils.project import get_project_settings
             settings = get_project_settings()
             if 'AWS_ACCESS_KEY_ID' in settings or 'AWS_SECRET_ACCESS_KEY' in settings:
-                import warnings
-                from scrapy.exceptions import ScrapyDeprecationWarning
                 warnings.warn(
                     "Initialising `scrapy.extensions.feedexport.S3FeedStorage` "
                     "without AWS keys is deprecated. Please supply credentials or "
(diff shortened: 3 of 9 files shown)


scrapy/commands/crawl.py
```diff
@@ -19,10 +17,8 @@ def add_options(self, parser):
         ScrapyCommand.add_options(self, parser)
         parser.add_option("-a", dest="spargs", action="append", default=[], metavar="NAME=VALUE",
                           help="set spider argument (may be repeated)")
-        parser.add_option("-o", "--output", metavar="FILE",
+        parser.add_option("-o", "--output", metavar="FILE", action="append",
```
Gallaecio: We should extend the documentation of this option to cover the usage of `:` to specify the output format.

It should be specially clear that, in order to use an output file name that contains one or more colons, specifying the output format is mandatory. For example, to write to `data:countries.csv` in CSV format, you must write `data:countries.csv:csv`.

We could support `data:countries.csv:` instead (if the format is an empty string, the file extension is used), but I’m not sure saving a few characters is worth making the implementation more complex.

scrapy/utils/conf.py
```diff
@@ -108,3 +109,31 @@ def get_sources(use_closest=True):
...
+            feed_uri, feed_format = element.rsplit(':', 1)
+        except ValueError:
+            feed_uri = element
+            feed_format = os.path.splitext(element)[1].replace(".", "")
+        else:
+            if feed_uri == '-':
+                feed_uri = 'stdout:'
+        if feed_format not in valid_output_formats:
+            raise UsageError("Unrecognized output format '%s', set one after a"
+                             " semicolon (i.e. -o <URI>:<FORMAT>) or as a file"
```
Gallaecio:
```suggestion
                             " colon (i.e. -o <URI>:<FORMAT>) or as a file"
```

scrapy/commands/crawl.py
```diff
@@ -19,10 +17,8 @@ def add_options(self, parser):
...
                           help="dump scraped items into FILE (use - for stdout)")
-        parser.add_option("-t", "--output-format", metavar="FORMAT",
```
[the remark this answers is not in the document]
elacuesta: Sounds good!

scrapy/extensions/feedexport.py
```diff
@@ -176,28 +176,39 @@ def _store_in_thread(self, file):
...
+        if not settings['FEED_URI']:
             raise NotConfigured
-        self.format = settings['FEED_FORMAT'].lower()
+
+        # Begin: Backward compatibility for string values in the FEED_URI setting
+        if isinstance(settings['FEED_URI'], dict):
+            self.urifmt = {uri.strip(): feed_format.strip().lower()
+                           for uri, feed_format in settings['FEED_URI'].items()}
+        else:
+            self.urifmt = {settings['FEED_URI']: settings['FEED_FORMAT'].lower()}
```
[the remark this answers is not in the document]
elacuesta: Yup, that was one of my next moves.

scrapy/extensions/feedexport.py
```diff
@@ -176,28 +176,39 @@ def _store_in_thread(self, file):
...
+
+        self.slots = []
         self.export_encoding = settings['FEED_EXPORT_ENCODING']
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
-        if not self._storage_supported(self.urifmt):
+        if not all(map(self._storage_supported, self.urifmt.keys())):
             raise NotConfigured
-        if not self._exporter_supported(self.format):
+        if not all(map(self._storage_supported, self.urifmt.values())):
```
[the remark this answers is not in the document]
elacuesta: Oops, fixing.

scrapy/utils/conf.py
elacuesta: True, I alwasys get colons and semicolons confused by the name.

scrapy/commands/crawl.py
elacuesta: NIce case, I hadn't thought of that, thanks!

scrapy/extensions/feedexport.py
```diff
@@ -187,28 +187,45 @@ def _store_in_thread(self, file):
...
-        self.urifmt = settings['FEED_URI']
-        if not self.urifmt:
+
+        if not settings['FEED_URI']:
             raise NotConfigured
-        self.format = settings['FEED_FORMAT'].lower()
+
+        # Begin: Backward compatibility for string values in the FEED_URI setting
+        if isinstance(settings['FEED_URI'], dict):
+            self.urifmts = {uri.strip(): feed_format.strip().lower()
```
Gallaecio: I think we should do something if `FEED_FORMAT` is used in combination with a `dict` value for `FEED_URI`.

Maybe we could log the warning below as long as `FEED_FORMAT` format is used, regardless or the value of `FEED_URI`?

elacuesta: I see your point, but `FEED_FORMAT` has a default value of `jsonlines` in `scrapy/settings/default_settings.py`. I'd be ok with setting that to `None`, but it would technically be a backward imcompatible change. What do you think?

Gallaecio: I think we should remove `FEED_FORMAT` from that file (in line with how [`SPIDER_MANAGER_CLASS` was renamed there in the past](https://github.com/scrapy/scrapy/commit/403e7c7c7059486c5a4c59f900a65378de2ba3d2#diff-ff579a5763caae1855c840beac29a49b), but this time simply removing), and as long as you replace `settings['FEED_FORMAT'].lower()` with `settings.get('FEED_FORMAT', 'jsonlines').lower()` below, it should be backward compatible.

scrapy/commands/crawl.py
elacuesta: I was thinking about _where_ to put this note, since the documentation about the commands does not go that deep as to mention `-o/-t`, but then I realized that the warning in this case is pretty self-explanatory IMHO:
```
$ scrapy runspider example.py -o data:countries.csv
Usage
=====
  scrapy runspider [options] <spider_file>

runspider: error: Unrecognized output format 'countries.csv', set one after a colon using the -s option (i.e. -o <URI>:<FORMAT>) or as a file extension, from the supported list ('json', 'jsonlines', 'jl', 'csv', 'xml', 'marshal', 'pickle')
```

What do you think @Gallaecio ?

docs/topics/feed-exports.rst
```diff
@@ -15,6 +15,15 @@ Scrapy provides this functionality out of the box with the Feed Exports, which
 allows you to generate a feed with the scraped items, using multiple
 serialization formats and storage backends.
 
+.. note::
+    Since version 1.X, Scrapy supports exporting to multiple feeds in a single crawl.
```
elacuesta: Which version? Is it too late to include it in 1.7?

scrapy/commands/crawl.py
Gallaecio: Yeah, I guess you are right.

docs/topics/feed-exports.rst
Gallaecio: Put 1.7 just in case. We can always change it later if it does not make the cut for 1.7.

scrapy/extensions/feedexport.py
elacuesta: Updatedm althought in a slightly different wasy as you suggested. Please check again, thanks.

Gallaecio:
I’m not sure about documenting deprecated behaviors, other than in the release notes. I would prefer that the general documentation describes only the non-default behavior, as if it had always been like this, and let users relying on deprecated behaviors get warnings that explain the deprecations to them and get them to read the release notes.

elacuesta:
hm, interesting point. Should we remove `FEED_FORMAT` completely from the docs then?

Gallaecio:
I would. It does mean that we cannot :setting:`FEED_FORMAT` in the release notes, but that is already the case with other removed settings.

elacuesta:
Makes sense, updated.

Gallaecio:
I was going over the [feed export documentation](https://docs.scrapy.org/en/latest/topics/feed-exports.html#std:setting-FEED_EXPORTERS), and I’m starting to think that there are other settings that may also make sense to be able to override in a per-file basis: `FEED_EXPORT_ENCODING`, `FEED_EXPORT_FIELDS`, `FEED_EXPORT_INDENT`, `FEED_STORE_EMPTY`.

I’m OK with only allowing per-file `FEED_FORMAT` as part of this pull request, in fact I think it may even be better, easier to review.

However, I do wonder if, in preparation for a future where we introduce additional per-file settings, we shouldn’t use something like `{'<path>': {'format': '<format>`}}` instead of simply `{'<path>': '<format>`}`.

As for the command line, I’m not sure what the best approach would be to eventually support additional per-file options. Although maybe that is less likely to happen, since the CLI is intended for simpler usage. Yet again, from that point of view, we could let the CLI interface as is, and not allow specifying multiple output files from the CLI, only through spider settings.

elacuesta:
I thought a bit about that, but figured that it would be ok to share all the other options among the different feeds. But since you pointed it out, I think it's something that would be useful indeed. I'd need to do some additional checks, let me think a bit on the best way to do it.
Regarding the CLI, I think it's better to offer multiple feed support (CLI support was in fact the main request in the original issue) but with minimal configuration, i.e. only file format.

elacuesta:
Actually, if we're adding that kind of functionality I think the name `FEED_URI` is just not good anymore for the task. I think we could use a new setting and use the "general" values as fallback if they're not specified for a specific feed, something like
```python
FEEDS = [{
    'uri': 'items.json',
    'format': 'json',
    'store_empty': True,
    'encoding': 'utf8',
    'fields': None,
    'indent': 4,
}]
```

elacuesta:
I refactored this a bit. Now something like 
```python
from scrapy import Spider

class ExampleSpider(Spider):
    name = 'example'
    start_urls = ['https://example.org']
    custom_settings = {
        'FEEDS': [{
            'uri': 'items.json',
            'format': 'json',
            'store_empty': True,
            'encoding': 'utf8',
            'fields': None,
            'indent': 3,
        }, {
            'uri': 'items.xml',
            'format': 'xml',
            'store_empty': True,
            'fields': ['name', 'price'],
            'encoding': 'latin1',
            'indent': 10,
        }, {
            'uri': 'items.csv',
            'format': 'csv',
            'fields': ['price', 'name'],
        }]
    }

    def parse(self, response):
        return dict(name='Cháïr', price=120.5, stock=10)
```

produces stuff like the following:

```xml
<?xml version="1.0" encoding="latin1"?>
<items>
          <item>
                    <name>Ch��r</name>
                    <price>120.5</price>
          </item>
</items>
```

```json
[
{
   "name": "Cháïr",
   "price": 120.5,
   "stock": 10
}
]
```

```csv
price,name
120.5,Cháïr
```

Now I need to rewrite some tests and docs :smile:

docs/topics/feed-exports.rst
```diff
@@ -211,38 +215,61 @@ Settings
...
-FEED_FORMAT
------------
 
-The serialization format to be used for the feed. See
-:ref:`topics-feed-format` for possible values.
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam posuere iaculis mollis. Duis cursus rutrum congue. Duis ut dolor sagittis, efficitur mi in, facilisis augue. Vivamus gravida pharetra velit id bibendum. Interdum et malesuada fames ac ante ipsum primis in faucibus.
+
+.. note::
+    
+    The ``FEED_URI`` and ``FEED_FORMAT`` settings have been deprecated in favor of :setting:`FEEDS`.
```
elacuesta: @Gallaecio I know we talked about removing and non documenting deprecated stuff, but as a user I'd expect at least one line, just saying the settings I've been relying in for so long are gone.

Gallaecio: I personally prefer to keep that line in the release notes. With that, and as long as the old, now-deprecated settings continue to work for a while, and issue a warning about their deprecation, I think we are giving users a good upgrade path.

@kmike?

elacuesta: Fair point. If we were to keep these in the main docs we'd need to monitor it and remove after a while, something that's not needed with the release notes. I'll remove it.

elacuesta:
Hi there @Gallaecio and @kmike, I believe this PR is in good shape to face a final and comprehensive review. I've already changed a couple things based on Adrián's comments, but I look forward to read any further feedback you may have. This is a feature I would very much like to see in Scrapy :rocket:  

PS:
I thought about adding documentation for the `FEED_URI_PARAMS` setting, which I think has been always undocumented, but since it relies in `load_object` I think it would be best to wait for #3873. Nobody has complained for the lack of docs there, I think it can wait a bit more :wink:

docs/topics/feed-exports.rst
```diff
@@ -211,38 +215,63 @@ Settings
...
+        },
+    }
 
-.. setting:: FEED_FORMAT
+The following is a list of the accepted keys and the Scrapy setting that is used
+as a fallback value if that key is not provided for a specific feed definition.
 
-FEED_FORMAT
------------
+* ``FORMAT`` (mandatory, no fallback setting)
```
Gallaecio: I think that, at least for `FORMAT`, we should keep here the documentation that used to be there for `FEED_FORMAT`:

> The serialization format to be used for the feed. See :ref:`topics-feed-format` for possible values.

docs/topics/feed-exports.rst
```diff
@@ -211,38 +215,63 @@ Settings
...
+* ``ENCODING`` (falls back to :setting:`FEED_EXPORT_ENCODING`)
+* ``STORE_EMPTY`` (falls back to :setting:`FEED_STORE_EMPTY`)
+* ``FIELDS`` (falls back to :setting:`FEED_EXPORT_FIELDS`)
+* ``INDENT`` (falls back to :setting:`FEED_EXPORT_INDENT`)
 
-The serialization format to be used for the feed. See
-:ref:`topics-feed-format` for possible values.
+See :ref:`topics-feed-storage-backends` for supported URI schemes.
```
Gallaecio: :lipstick: Maybe this should me covered before the list of accepted keys.

scrapy/extensions/feedexport.py
```diff
@@ -186,86 +187,113 @@ def _store_in_thread(self, file):
...
+    def __init__(self, crawler):
+        self.crawler = crawler
+        self.settings = crawler.settings
+        self.feeds = {}
+        self.slots = []
+
+        if not self.settings['FEEDS'] and not self.settings['FEED_URI']:
+            raise NotConfigured
+
+        FEEDS = self.settings['FEEDS'] or {}
```
Gallaecio: :lipstick: 

```suggestion
        feeds = self.settings['FEEDS'] or {}
```

scrapy/extensions/feedexport.py
```diff
@@ -186,86 +187,113 @@ def _store_in_thread(self, file):
...
+        for uri, feed in FEEDS.items():
+            self.feeds[uri] = feed_complete_default_values_from_settings(
+                {key.strip().upper(): value for key, value in feed.items()},
+                self.settings
+            )
+
+        # Begin: Backward compatibility for FEED_URI and FEED_FORMAT settings
+        if not self.feeds and self.settings['FEED_URI']:
```
Gallaecio: I think a warning about `self.settings['FEED_URI']` should be issued regardless of whether or not `self.settings['FEEDS']` is also used.

docs/topics/feed-exports.rst
```diff
@@ -15,6 +15,10 @@ Scrapy provides this functionality out of the box with the Feed Exports, which
 allows you to generate a feed with the scraped items, using multiple
 serialization formats and storage backends.
 
+.. note::
+    Since version 1.7, Scrapy supports exporting to multiple feeds in a single crawl.
```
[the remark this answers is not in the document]
elacuesta: That's right, the paragraph is probably not necessary, the `versionadded` is already in the setting, I just need to update the number. Also, I'll change the number in the above paragraph (feed -> feeds)

elacuesta:
Hey @Gallaecio, many thanks for your feedback! I think I addressed all of your points, let me know if there is anything else.

docs/topics/feed-exports.rst
```diff
@@ -211,38 +211,65 @@ Settings
...
-FEED_FORMAT
------------
+The following is a list of the accepted keys and the Scrapy setting that is used
+as a fallback value if that key is not provided for a specific feed definition.
 
-The serialization format to be used for the feed. See
-:ref:`topics-feed-format` for possible values.
+* ``FORMAT``: the serialization format to be used for the feed.
+  See :ref:`topics-feed-format` for possible values. 
+  Mandatory, no fallback setting)
```
Gallaecio: These bullet points should not end in `)` anymore :slightly_smiling_face:

elacuesta: :man_facepalming:

Gallaecio:
@elacuesta Please, make sure that these changes work when the setting is defined from the command line (as a string). This is necessary not only for the command line use, but also for other scenarios where settings are defined as strings, such as Scrapy Cloud.

elacuesta:
Sure, I'll add some tests regarding that, stay tuned! :radio:

Gallaecio:
It’s not just about the command line, setting values as strings can come from environment variables, etc.

What you need to do is use one of the method of the `Settings` class when you read the setting values. Good news is that I was wrong, `getdict` should do the job, for some reason I though it the root structure was a list.

It is probably enough to change https://github.com/scrapy/scrapy/pull/3858/files#diff-b40c5755a5dda2b02aaac6bf4c8e2ff8R236 to call `self.settings.getdict('FEEDS')` instead of `self.settings['FEEDS']`.

elacuesta:
Alright, makes sense, but out of genuine curiosity, is populating settings through env variables currently working? I couldn't find references in [the docs](https://docs.scrapy.org/en/1.7/topics/settings.html#populating-the-settings) nor in [the code itself](https://github.com/scrapy/scrapy/blob/1.7/scrapy/settings/__init__.py#L17-L23), although I did find some docstrings about it (https://github.com/scrapy/scrapy/blob/1.7/scrapy/settings/__init__.py#L122, https://github.com/scrapy/scrapy/blob/1.7/scrapy/settings/__init__.py#L172).
Am I missing something?

Gallaecio:
I’m not 100% sure whether or not defining settings through environment variables works.

In any case, this is also needed to support setting a feed configuration with the `-s` command-line switch of `scrapy crawl`.

elacuesta:
It's actually not needed, that's precisely what https://github.com/scrapy/scrapy/pull/3858/commits/d69573b18df43cb2435213e4bc3b0ae7103e4a46#diff-59262c12118a93348565a41ab7f6ce42R40 proves.
I traced this to:
* `scrapy.commands.ScrapyCommand.process_options` calls `scrapy.settings.BaseSettings.setdict` (https://github.com/scrapy/scrapy/blob/1.7.3/scrapy/commands/__init__.py#L78), which in turn calls
* `scrapy.settings.BaseSettings.update`, which decodes from JSON if the value is a string (https://github.com/scrapy/scrapy/blob/1.7.3/scrapy/settings/__init__.py#L299-L323)

Anyway, I don't really have any strong reason or preference for `settings['FEEDS']` over `settings.getdict('FEEDS')`, so let's just do it. Thanks!

elacuesta:
@Gallaecio Updated, it does look cleaner :-)

Also added a line in the function that processes the options from the CLI to make the `FEEDS` setting take precedence over the `-o` and `-t` options, which is the reasonable thing to do IMHO.

scrapy/extensions/feedexport.py
```diff
@@ -186,86 +187,112 @@ def _store_in_thread(self, file):
...
+        for uri, feed in self.feeds.items():
+            uri = uri % self._get_uri_params(spider, feed['URI_PARAMS'])
+            storage = self._get_storage(uri)
+            file = storage.open(spider)
+            exporter = self._get_exporter(
+                file=file,
+                feed_format=feed['FORMAT'],
+                fields_to_export=feed['FIELDS'],
+                encoding=feed['ENCODING'],
+                indent=feed['INDENT'],
```
Gallaecio: I was writing an answer for a [StackOverflow question](https://stackoverflow.com/q/57996812/939364) about customizing `XmlItemExporter` parameters, and I think maybe we should pave the way to make this code easier to extend in the future.

For example, if we use lowercase settings in `FEEDS`, and we deprecate `feed_format` and `fields_to_export` in item exporters and replace them with `format` and `fields`, we can use `exporter = self._get_exporter(file=file, **feed)` here.

That in turn allows users to take advantage of additional item exporter `__init__` parameters that are currently not exposed as settings, such as XmlItemExporter’s `item_element` and `root_element`, which would make it possible to solve the use case described in the [StackOverflow question](https://stackoverflow.com/q/57996812/939364) using only the `FEEDS` setting.

Then we would just need to document those additional `__init__` parameters that each item exporter supports.

elacuesta: Great idea, thanks :smile:

PR updated, just a few comments:
* The `feed_format` name was introduced earlier in this PR, to avoid overriding the [`format` built-in function](https://docs.python.org/3/library/functions.html#format). It was only used internally in the `extensions/feedexport.py` file. I reverted that change, `format` shouldn't be a problem since the binding applies only to "private" (i.e. underscored) classes and methods.
* I'm hesitant about which name to use to configure the fields in the `FEEDS` setting. On the one hand, `fields` seems prettier and simpler than `fields_to_export`, on the other hand using `fields_to_export` would be consistent with the existing exporter init parameter and would allow us to make the follow-up change in a backward compatible manner. Thoughts?

Gallaecio: If we want to eventually rename the internal parameter to `fields`, I think it’s best we name the user-facing parameter `fields` already.

That way, once we change the internal field, users don’t need to change their `FEEDS` setting. Only people using the private `_get_exporter` method will need to update their code.

elacuesta: True, but `fields_to_export` is [already documented](https://docs.scrapy.org/en/1.7/topics/exporters.html?highlight=fields_to_export#baseitemexporter).
Since the `FEEDS` setting is new, we could adapt to the current name and use `FEEDS.fields_to_export` instead of `FEEDS.fields` and prevent a backward-compatible change in the future.

Gallaecio: Oh, so stick to `fields_to_export`, with no plans of eventually changing it to just `fields`.

Works for me :+1:

Gallaecio:
This will need some conflict resolution, specially after #4074.

docs/topics/feed-exports.rst
```diff
@@ -211,38 +211,65 @@ Settings
...
-.. setting:: FEED_URI
+.. setting:: FEEDS
 
-FEED_URI
---------
+FEEDS
+-----
 
-Default: ``None``
+.. versionadded:: 1.8
```
kmike: 1.8 is released

vs4vijay:
Hey, 

Is this feature merged?

elacuesta:
Ping @kmike 😇 
Do you think is there anything else I need to address here?

docs/topics/feed-exports.rst
```diff
@@ -211,41 +211,66 @@ Settings
...
-.. setting:: FEED_FORMAT
+    {
+        'items.json': {
+            'format': 'json',
+            'encoding': 'utf8',
+            'store_empty': False,
+            'fields': None,
+            'indent': 4,
+        }, 
+        'items.xml': {
```
wRAR: Maybe add a full URI example here?

scrapy/utils/conf.py
```diff
@@ -106,3 +109,66 @@ def get_sources(use_closest=True):
     if use_closest:
         sources.append(closest_scrapy_cfg())
     return sources
+
+
+def feed_complete_default_values_from_settings(feed, settings):
+    out = feed.copy()
+    if 'encoding' not in out:
```
wRAR: Can this be replaced with `.setdefault()`?

wRAR:
Thanks @elacuesta !

PR #3859, opened 2019-07-08
More deprecations

kmike:
I noticed that there are 2 import-only modules in scrapy.utils, and one which is almost import-only. It looks like we should start sending deprecation warnings for them all, and this should have been done a while ago.

### DIFF
--- a/scrapy/utils/http.py
+++ b/scrapy/utils/http.py
@@ -4,8 +4,19 @@ Transitional module for moving to the w3lib library.
 For new code, always import from w3lib.http instead of this module
 """
 
+import warnings
+
+from scrapy.exceptions import ScrapyDeprecationWarning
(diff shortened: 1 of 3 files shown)


nyov:
@kmike, would you or someone perhaps also find time to correctly deprecate this (or just rip it out)?: https://github.com/scrapy/scrapy/blob/master/scrapy/utils/project.py#L70-L79
Or is it still needed.

Gallaecio:
> @kmike, would you or someone perhaps also find time to correctly deprecate this (or just rip it out)?: https://github.com/scrapy/scrapy/blob/master/scrapy/utils/project.py#L70-L79
> Or is it still needed.

I’ve just created #3871 so that we do not loose track of this.

PR #3861, opened 2019-07-09
Increase the documentation coverage of exceptions

Gallaecio:
This removes 2 out of 3 API members of `scrapy.exceptions` from the documentation coverage report. The remaining class is to be covered by #3636.


kmike:
Thanks @Gallaecio!

PR #3862, opened 2019-07-09
[MRG+1] Create Request from curl command 

noviluni:
This is a WIP, but I open this PR to discuss how to implement this feature. 

Based on: https://github.com/scrapy/scrapy/pull/2985 (abandoned)

### DIFF
--- a/docs/topics/developer-tools.rst
+++ b/docs/topics/developer-tools.rst
@@ -276,7 +276,7 @@ request::
 
 Alternatively, if you want to know the arguments needed to recreate that
 request you can use the :func:`scrapy.utils.curl.curl_to_request_kwargs`
-function and you will get a dictionary with the equivalent arguments.
+function to get a dictionary with the equivalent arguments.
 
 As you can see, with a few inspections in the `Network`-tool we
 were able to easily replicate the dynamic requests of the scrolling 
--- a/docs/topics/developer-tools.rst
+++ b/docs/topics/developer-tools.rst
@@ -254,7 +254,32 @@ page-number greater than 10), we increment the ``page`` attribute
 and ``yield`` a new request, inserting the incremented page-number 
 into our ``url``. 
 
-You can see that with a few inspections in the `Network`-tool we 
+In more complex websites, it could be difficult to easily reproduce the
+requests, as we could need to add ``headers`` or ``cookies`` to make it work. In those
+cases you can export the requests in `cURL <https://curl.haxx.se/>`_ format, by
+right-clicking on it in the network tool and using the
+:meth:`~scrapy.http.Request.from_curl()` method to generate an equivalent
+request. To do so, just import the :class:`~scrapy.http.Request` class and use this
+method to get the desired request::
+
+    from scrapy import Request
+
+    request = Request.from_curl(
+        "curl 'http://quotes.toscrape.com/api/quotes?page=1' -H 'User-Agent: Mozil"
+        "la/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0' -H 'Acce"
+        "pt: */*' -H 'Accept-Language: ca,en-US;q=0.7,en;q=0.3' --compressed -H 'X"
+        "-Requested-With: XMLHttpRequest' -H 'Proxy-Authorization: Basic QFRLLTAzM"
+        "zEwZTAxLTk5MWUtNDFiNC1iZWRmLTJjNGI4M2ZiNDBmNDpAVEstMDMzMTBlMDEtOTkxZS00MW"
+        "I0LWJlZGYtMmM0YjgzZmI0MGY0' -H 'Connection: keep-alive' -H 'Referer: http"
+        "://quotes.toscrape.com/scroll' -H 'Cache-Control: max-age=0'")
+
+
+Alternatively, if you want to know the arguments needed to recreate that
+request you can use the :func:`scrapy.utils.curl.curl_to_request_kwargs`
+function and you will get a dictionary with the equivalent arguments.
+
+
+As you can see, with a few inspections in the `Network`-tool we
 were able to easily replicate the dynamic requests of the scrolling 
 functionality of the page. Crawling dynamic pages can be quite
 daunting and pages can be very complex, but it (mostly) boils down
@@ -262,7 +287,7 @@ to identifying the correct request and replicating it in your spider.
 
 .. _Developer Tools: https://en.wikipedia.org/wiki/Web_development_tools
 .. _quotes.toscrape.com: http://quotes.toscrape.com
-.. _quotes.toscrape.com/scroll: quotes.toscrape.com/scroll/
+.. _quotes.toscrape.com/scroll: http://quotes.toscrape.com/scroll
 .. _quotes.toscrape.com/api/quotes?page=10: http://quotes.toscrape.com/api/quotes?page=10
 .. _has-class-extension: https://parsel.readthedocs.io/en/latest/usage.html#other-xpath-extensions
 
--- a/docs/topics/developer-tools.rst
+++ b/docs/topics/developer-tools.rst
@@ -252,15 +252,16 @@ If the handy ``has_next`` element is ``true`` (try loading
 `quotes.toscrape.com/api/quotes?page=10`_ in your browser or a
 page-number greater than 10), we increment the ``page`` attribute 
 and ``yield`` a new request, inserting the incremented page-number 
-into our ``url``. 
+into our ``url``.
+
+.. _requests-from-curl:
 
 In more complex websites, it could be difficult to easily reproduce the
-requests, as we could need to add ``headers`` or ``cookies`` to make it work. In those
-cases you can export the requests in `cURL <https://curl.haxx.se/>`_ format, by
-right-clicking on it in the network tool and using the
+requests, as we could need to add ``headers`` or ``cookies`` to make it work.
+In those cases you can export the requests in `cURL <https://curl.haxx.se/>`_
+format, by right-clicking on each of them in the network tool and using the
 :meth:`~scrapy.http.Request.from_curl()` method to generate an equivalent
-request. To do so, just import the :class:`~scrapy.http.Request` class and use this
-method to get the desired request::
+request::
 
     from scrapy import Request
 
@@ -273,11 +274,9 @@ method to get the desired request::
         "I0LWJlZGYtMmM0YjgzZmI0MGY0' -H 'Connection: keep-alive' -H 'Referer: http"
         "://quotes.toscrape.com/scroll' -H 'Cache-Control: max-age=0'")
 
-
 Alternatively, if you want to know the arguments needed to recreate that
 request you can use the :func:`scrapy.utils.curl.curl_to_request_kwargs`
 function and you will get a dictionary with the equivalent arguments.
-
 
 As you can see, with a few inspections in the `Network`-tool we
 were able to easily replicate the dynamic requests of the scrolling 
--- a/docs/topics/dynamic-content.rst
+++ b/docs/topics/dynamic-content.rst
@@ -85,11 +85,12 @@ It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
 method and URL. However, you may also need to reproduce the body, headers and
 form parameters (see :class:`~scrapy.http.FormRequest`) of that request.
 
-As all the major browsers allow to export the requests in `cURL
-<https://curl.haxx.se/>`_ format, scrapy incorporates the method
+As all major browsers allow to export the requests in `cURL
+<https://curl.haxx.se/>`_ format, Scrapy incorporates the method
 :meth:`~scrapy.http.Request.from_curl()` to generate an equivalent
 :class:`~scrapy.http.Request` from a cURL command. To get more information
-visit :ref:`the network tool section <topics-network-tool>`.
+visit :ref:`request from curl <requests-from-curl>` inside the network
+tool section.
 
 Once you get the expected response, you can :ref:`extract the desired data from
 it <topics-handling-response-formats>`.
--- a/docs/topics/dynamic-content.rst
+++ b/docs/topics/dynamic-content.rst
@@ -85,6 +85,12 @@ It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
 method and URL. However, you may also need to reproduce the body, headers and
 form parameters (see :class:`~scrapy.http.FormRequest`) of that request.
 
+As all the major browsers allow to export the requests in `cURL
+<https://curl.haxx.se/>`_ format, scrapy incorporates the method
+:meth:`~scrapy.http.Request.from_curl()` to generate an equivalent
+:class:`~scrapy.http.Request` from a cURL command. To get more information
+visit :ref:`the network tool section <topics-network-tool>`.
+
 Once you get the expected response, you can :ref:`extract the desired data from
 it <topics-handling-response-formats>`.
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -194,6 +194,11 @@ Request objects
        copied by default (unless new values are given as arguments). See also
        :ref:`topics-request-response-ref-request-callback-arguments`.
 
+    .. classmethod:: Request.from_curl(curl_command)
+
+       Create a Request object from a string containing a curl command. It populates
+       the HTTP method, the URL, the headers, the cookies and the body.
+
 .. _topics-request-response-ref-request-callback-arguments:
 
 Passing additional data to callback functions
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -194,10 +194,7 @@ Request objects
        copied by default (unless new values are given as arguments). See also
        :ref:`topics-request-response-ref-request-callback-arguments`.
 
-    .. classmethod:: Request.from_curl(curl_command)
-
-       Create a Request object from a string containing a curl command. It populates
-       the HTTP method, the URL, the headers, the cookies and the body.
+    .. automethod:: from_curl
 
 .. _topics-request-response-ref-request-callback-arguments:
 
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -118,13 +118,16 @@ class Request(object_ref):
         finding unknown options call this method by passing
         ``ignore_unknown_options=False``.
 
-        .. caution:: Although the request is built based on the content of the
-                     cURL command, some headers could be added when requesting
-                     attending to the enabled middlewares (E.g.:
+        .. caution:: Enabled
+                     :ref:`downloader middlewares <topics-downloader-middleware>`
+                     and
+                     :ref:`spider middlewares <topics-spider-middleware>`,
+                     such as
                      :class:`~scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware`,
                      :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`,
+                     or
                      :class:`~scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware`,
-                     etc.).
+                     may modify the :class:`~scrapy.http.Request` object.
 
        """
         request_kwargs = curl_to_request_kwargs(curl_command, ignore_unknown_options)
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -12,6 +12,7 @@ from scrapy.utils.python import to_bytes
 from scrapy.utils.trackref import object_ref
 from scrapy.utils.url import escape_ajax
 from scrapy.http.common import obsolete_setter
+from scrapy.utils.curl import curl_to_request_kwargs
 
 
 class Request(object_ref):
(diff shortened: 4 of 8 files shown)


noviluni:
Based on comments from #2985 (@kmike), I have renamed the function, added docs, added some tests and implemented `Request.from_curl`.

I have also improved the function by adding `basic auth` support and fixing when scheme is missing.

My idea is to add more tests with special cases (got some ideas from: https://github.com/NickCarneiro/curlconverter/tree/master/fixtures/curl_commands) and of course implement all missing features.

Any ideas or feedback? :)

docs/topics/request-response.rst
```diff
@@ -194,6 +194,11 @@ Request objects
        copied by default (unless new values are given as arguments). See also
        :ref:`topics-request-response-ref-request-callback-arguments`.
 
+    .. classmethod:: Request.from_curl(curl_command)
```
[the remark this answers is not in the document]
noviluni: Done!

docs/topics/request-response.rst
```diff
@@ -194,6 +194,11 @@ Request objects
...
+
+       Create a Request object from a string containing a curl command. It populates
```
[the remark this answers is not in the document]
noviluni: Good idea :)

scrapy/http/request/__init__.py
```diff
@@ -103,3 +104,9 @@ def replace(self, *args, **kwargs):
             kwargs.setdefault(x, getattr(self, x))
         cls = kwargs.pop('cls', self.__class__)
         return cls(*args, **kwargs)
+
+    @classmethod
+    def from_curl(cls, curl_command):
```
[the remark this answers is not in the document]
noviluni: Implemented. I changed it to accept the same arguments as `Request`. Also documented that those arguments take preference over those contained in the "curl command". I was doubting if it's necessary to override them or if we could just extend (in case of `meta`, `headers` and `cookies` for example), but I decided just to override because if we allow to extend those fields and the programmer want to override them it could be counterintuitive. If someone want just to change one header it could be possible to create the Request and then change that specific header.

Tell me if you think I'm wrong or if you think it could be better to extend those arguments.

scrapy/http/request/__init__.py
```diff
@@ -103,3 +104,14 @@ def replace(self, *args, **kwargs):
             kwargs.setdefault(x, getattr(self, x))
         cls = kwargs.pop('cls', self.__class__)
         return cls(*args, **kwargs)
+
+    @classmethod
+    def from_curl(cls, curl_command, **kwargs):
+        """Create a Request object from a string containing a `curl <https://curl.haxx.se/>`_ command. It populates the
+        HTTP method, the URL, the headers, the cookies and the body. It accepts the same arguments as the
+        :class:`Request` class. Note that all the arguments provided take preference and will override the values of
+        the same arguments contained in the curl command.
```
kmike: Could you please limit line length? We're not super-strict about it, but unless there is a reason, we try to follow e.g. 79 columns. For example, reviewing this diff on github requires horizontal scrolling :)

scrapy/utils/curl.py
```diff
@@ -0,0 +1,80 @@
...
+    if isinstance(curl_args, string_types):
+        curl_args = split(curl_args)
+
+    if curl_args[0] != 'curl':
+        raise ValueError('A curl command must start with "curl"')
+
+    parsed_args, argv = curl_parser.parse_known_args(curl_args[1:])
+    if argv:
+        msg = 'Unrecognized arguments: %s'
+        raise ValueError(msg % (argv,))
```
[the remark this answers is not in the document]
noviluni: I'm not sure about this question. I think the way is implemented (accepting some arguments and ignore them but raising those arguments that are not recognized) is the best as if someone uses this method assuming it will generate exactly the same request than the curl command and we just ignore some arguments without raising that could fail without realizing it. However I understand that this could make this code more difficult to mantain.

What do you think? @kmike @Gallaecio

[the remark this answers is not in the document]
noviluni: (Of course, if we decide to continue like this I should add a lot of parameters to this code).

kmike:
@noviluni just heads up: we want to make Scrapy release next week; if you finish this PR this week, we'll get it in :)

noviluni:
> Changes look great, thanks for picking this up.
> 
> I think there are a few pages of the documentation that would greatly benefit from covering this feature, if you are up to it:
> 
>     * https://docs.scrapy.org/en/master/topics/dynamic-content.html#reproducing-requests
> 
>     * https://docs.scrapy.org/en/master/topics/developer-tools.html#the-network-tool

@Gallaecio Could you please review the changes I did in my last commit? Feel free to suggest any changes you think should be added.

noviluni:
> @noviluni just heads up: we want to make Scrapy release next week; if you finish this PR this week, we'll get it in :)

@kmike That would be great, but I need feedback about what to do with unrecognized parameters (comment above).

Apart of the above, we could explain in the docs that there is a limited support and that certain parameters are not allowed.

docs/topics/developer-tools.rst
```diff
@@ -254,15 +254,40 @@ page-number greater than 10), we increment the ``page`` attribute
 and ``yield`` a new request, inserting the incremented page-number 
 into our ``url``. 
 
-You can see that with a few inspections in the `Network`-tool we 
+In more complex websites, it could be difficult to easily reproduce the
+requests, as we could need to add ``headers`` or ``cookies`` to make it work. In those
+cases you can export the requests in `cURL <https://curl.haxx.se/>`_ format, by
+right-clicking on it in the network tool and using the
```
Gallaecio:
```suggestion
right-clicking on each of them in the network tool and using the
```

docs/topics/developer-tools.rst
```diff
@@ -254,15 +254,40 @@ page-number greater than 10), we increment the ``page`` attribute
...
+:meth:`~scrapy.http.Request.from_curl()` method to generate an equivalent
+request. To do so, just import the :class:`~scrapy.http.Request` class and use this
```
Gallaecio: Since you are already linking to the reference documentation of `from_curl`, where users can check how to import `Request`, and moreover the example already covers how to import `Request` as well, I think the last sentence of this paragraph can be omitted (i.e. end with `to generate an equivalent request::`).

docs/topics/developer-tools.rst
```diff
@@ -254,15 +254,40 @@ page-number greater than 10), we increment the ``page`` attribute
...
+    request = Request.from_curl(
+        "curl 'http://quotes.toscrape.com/api/quotes?page=1' -H 'User-Agent: Mozil"
+        "la/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0' -H 'Acce"
+        "pt: */*' -H 'Accept-Language: ca,en-US;q=0.7,en;q=0.3' --compressed -H 'X"
+        "-Requested-With: XMLHttpRequest' -H 'Proxy-Authorization: Basic QFRLLTAzM"
+        "zEwZTAxLTk5MWUtNDFiNC1iZWRmLTJjNGI4M2ZiNDBmNDpAVEstMDMzMTBlMDEtOTkxZS00MW"
+        "I0LWJlZGYtMmM0YjgzZmI0MGY0' -H 'Connection: keep-alive' -H 'Referer: http"
+        "://quotes.toscrape.com/scroll' -H 'Cache-Control: max-age=0'")
+
+
```
Gallaecio: 1 line break is enough between paragraphs. Also applies _after_ the following paragraph.

docs/topics/dynamic-content.rst
```diff
@@ -85,6 +85,12 @@ It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
 method and URL. However, you may also need to reproduce the body, headers and
 form parameters (see :class:`~scrapy.http.FormRequest`) of that request.
 
+As all the major browsers allow to export the requests in `cURL
```
Gallaecio:
```suggestion
As all major browsers allow to export the requests in `cURL
```

docs/topics/dynamic-content.rst
```diff
@@ -85,6 +85,12 @@ It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
...
+<https://curl.haxx.se/>`_ format, scrapy incorporates the method
```
Gallaecio:
```suggestion
<https://curl.haxx.se/>`_ format, Scrapy incorporates the method
```

docs/topics/dynamic-content.rst
```diff
@@ -85,6 +85,12 @@ It might be enough to yield a :class:`~scrapy.http.Request` with the same HTTP
...
+:meth:`~scrapy.http.Request.from_curl()` to generate an equivalent
+:class:`~scrapy.http.Request` from a cURL command. To get more information
+visit :ref:`the network tool section <topics-network-tool>`.
```
Gallaecio: Instead of using `topics-network-tool` here, I think it would be better if you added a line with something like `.. _requests-from-curl::` right above the new documentation you added to the network tool topic, and replaced `topics-network-tool` with `requests-from-curl` here. That should make this link to the specific paragraphs that cover copying requests in cURL format.

docs/topics/developer-tools.rst
```diff
@@ -254,15 +254,40 @@ page-number greater than 10), we increment the ``page`` attribute
 and ``yield`` a new request, inserting the incremented page-number 
 into our ``url``. 
 
-You can see that with a few inspections in the `Network`-tool we 
+In more complex websites, it could be difficult to easily reproduce the
```
Gallaecio: This is what I meant in my comment about the `:ref:` link on the dynamic content documentation:

```suggestion
.. _requests-from-curl::

In more complex websites, it could be difficult to easily reproduce the
```

scrapy/utils/curl.py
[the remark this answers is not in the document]
noviluni: @Gallaecio  What do you thing about adding something like "Note that it has limited support as not all the cURL arguments have sense in Scrapy context" in docstrings?

Gallaecio:
> […] we could explain in the docs that there is a limited support and that certain parameters are not allowed.

Yes. Maybe we should even list all the specific parameters that are supported. And, if we end up raising an error on unrecognized parameters, we should also list supported parameters that have no impact on the resulting `Request`.

scrapy/utils/curl.py
```diff
@@ -0,0 +1,82 @@
...
+    result = {
+        'method': parsed_args.method.upper(),
+        'url': url,
+    }
+
+    headers = []
+    cookies = {}
+    for header in parsed_args.headers or ():
+        name, val = header.split(':', 1)
+        name = name.strip().title()
```
kmike: why is .title() needed?

immerrr: My memory is foggy, since it was a while ago, so take it with a pinch of salt, IIRC it provided some case-sensitivity consistency. 

Headers are case-insensitive according to spec, and with JS code being all over the place as it is in the wild, I have seen a lot of cases with curl commands having all-lowercase headers. For those, when you would try to check or override them you would use the more canonical `Title-Cased` version and it would return nothing or end up having both titlecase and lowercase versions at the same time.

I suppose this is a non-issue now that the preferred public API goes through creating a `Request` object that wraps headers in a `CaselessDict`, so we can probably drop this.

scrapy/utils/curl.py
```diff
@@ -0,0 +1,82 @@
...
+        'url': url,
+    }
+
+    headers = []
+    cookies = {}
+    for header in parsed_args.headers or ():
+        name, val = header.split(':', 1)
+        name = name.strip()
+        val = val.strip()
+        if name == 'Cookie':
```
kmike: comparison should be case-insensitive, e.g. `if name.title() == 'Cookie'`

scrapy/utils/curl.py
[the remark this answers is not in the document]
kmike: I think there are 2 use cases:

1) make Request similar to a curl command, on a best-effort basis, hoping that it would be enough;
2) make Request exactly as close as possible to curl.

Our main use case is (1), where user tries to reproduce the response from a working curl command, copy-pasting it e.g. from a browser. It means that we shouldn't raise an exception by default if not all options are supported - but having a warning would be nice, as well as a message in a docstring.

The problem with a warning is that we detect that something is not supported, but don't return this information to an end user, so users can't raise an error themselves if they want - so (2) is not supported if we only have a warning. I think this can be solved in several ways, e.g.:

1) have options parsing in a separate function, and return parsed options + leftovers, so that users can check if there are any leftovers, and act on it. 
2) have a flag (ignore_unknown_options=True), similar to request's raise_for_status, which allows user to turn a warning to an exception.

scrapy/utils/curl.py
```diff
@@ -39,9 +53,13 @@ def curl_to_request_kwargs(curl_args):
         raise ValueError('A curl command must start with "curl"')
 
     parsed_args, argv = curl_parser.parse_known_args(curl_args[1:])
+
     if argv:
-        msg = 'Unrecognized arguments: %s'
-        raise ValueError(msg % (argv,))
+        msg = 'Unrecognized options: {}'.format(', '.join(argv))
+        if ignore_unknown_options:
+            logger.warning(msg)
```
kmike: Why not warnings.warn?

scrapy/utils/curl.py
```diff
@@ -21,14 +25,24 @@ def error(self, message):
...
 
+safe_to_ignore_arguments = [
+    ['--compressed'],
+    ['-s', '--silent'],
+    ['-v', '--verbose'],
+    ['-#', '--progress-bar']
+]
+
+for argument in safe_to_ignore_arguments:
+    curl_parser.add_argument(*argument, action='store_true')
```
noviluni: Not sure about this. IMO those arguments we know that can be ignored shouldn't raise an error even when passing `ignore_unknown_options=False` as we know that they don't affect to the Scrapy request and could be really common (i.e.: --compress, which used to be in the cURL commands coming from the network tool in browsers).

@kmike I can delete this or maintain it here. Up to you.

kmike: I like an idea of having such list.

It seems however that `--compressed` is not safe to ignore, as it affects request headers.

scrapy/utils/curl.py
noviluni: Was my fault, I fix it.

scrapy/utils/curl.py
noviluni: I based that on the original work and I also saw that this tool: https://curl.trillworks.com/ ignore it. 

However, I've seen that using `--compressed` cURL adds the `Accept-Encoding: deflate, gzip` header. 

Do you think I should implement that?

kmike: Scrapy adds this header by default (see https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpcompression.py), so I'm not sure what to do.

Without this option we shouldn't set this header, but it can be tricky to implement here - having an empty header is not the same, if I'm not mistaken.

So a strict way would be probably to show a warning (a different one) if `--compressed` is not passed? And then implement a way to disable setting Accept header per-request in HttpCompressionMiddleware, and use it here.

kmike: we may have a similar issue with UserAgentMiddleware, DefaultHeadersMiddleware and other middlewares which set headers. Not sure how big is an issue though, maybe a warning in docs is fine.

noviluni: Ok, so I added a "caution" message in docs.
captura_13

Maybe that should be enough to start with (?).

scrapy/http/request/__init__.py
```diff
@@ -103,3 +104,29 @@ def replace(self, *args, **kwargs):
...
+        URL, the headers, the cookies and the body. It accepts the same
+        arguments as the :class:`Request` class, taking preference and
+        overriding the values of the same arguments contained in the cURL
+        command.
+
+        Unrecognized options are ignored by default. To raise an error when
+        finding unknown options call this method by passing
+        ``ignore_unknown_options=False``.
+
+        .. caution:: Although the request is built based on the content of the
```
Gallaecio: I would explicitly mention downloader middlewares and spider middlewares. I would also simply state that they can change the request without mentioning headers, since technically other aspects of requests could also be modified (including dropping requests, duplicating them, etc.).

```
        .. caution: Enabled
                     :ref:`downloader middlewares <topics-downloader-middleware>`
                     and
                     :ref:`spider middlewares <topics-spider-middleware>`,
                     such as
                     :class:`~scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware`,
                     :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`,
                     or
                     :class:`~scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware`,
                     may modify the resulting :class:`~scrapy.http.Request`
                     object.
```

scrapy/utils/curl.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def curl_to_request_kwargs(curl_args, ignore_unknown_options=True):
+    """Convert a cURL command syntax to Request kwargs.
+
+    :param str curl_args: string containing the curl command
```
Gallaecio: :lipstick: 

```suggestion
    :param str curl_command: string containing the curl command
```

scrapy/utils/curl.py
```diff
@@ -0,0 +1,97 @@
...
+    :param bool ignore_unknown_options: If true, only a warning is emitted when
+    cURL options are unknown. Otherwise raises an error. (default: True)
+    :return: dictionary of Request kwargs
+    """
+
+    if isinstance(curl_args, string_types):
```
Gallaecio: Maybe we can omit the type check? According to the documentation a string if the only allowed value.

scrapy/utils/curl.py
kmike: Could you please add a comment to `--compressed` item in the ignored list, which explains that it is not safe to ignore, but ignored because this is Scrapy default?

scrapy/utils/curl.py
noviluni: Hi @Gallaecio, thank you for your feedback!
The `isinstance` is there since the first/original implementation was made and was there to support passing list arguments instead of a string. I’m not sure why that support was added, I think that maybe was added to support to use this function with already parsed arguments (python script to execute directly from the terminal?) but as we don’t need now I removed that part. Shouldn’t be difficult to implement if we need it in the future.

scrapy/utils/curl.py
noviluni: Of course! I did it :+1:

scrapy/http/request/__init__.py
```diff
@@ -103,3 +104,32 @@ def replace(self, *args, **kwargs):
...
+                     :class:`~scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware`,
+                     :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`,
+                     or
+                     :class:`~scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware`,
+                     may modify the :class:`~scrapy.http.Request` object.
+
+       """
+        request_kwargs = curl_to_request_kwargs(curl_command, ignore_unknown_options)
+        request_kwargs.update(kwargs)
+        return Request(**request_kwargs)
```
kmike: I wonder if it should use cls instead of Request, not sure about this. Opinions?

kmike:
Thanks @noviluni! This looks really good. I got a late comment about a class which from_curl constructor should create, but no strong opinion on this.

scrapy/http/request/__init__.py
noviluni: Oh, that's right... It’s not as trivial as it seems. Suppose someone uses from_curl from JSONRequest:

`JSONRequest.from_curl('curl example.org')`.

Using `Request` the result would be:

```
{'_encoding': 'utf-8',
 'method': 'GET',
 '_url': 'http://example.org',
 '_body': b'',
 'priority': 0,
 'callback': None,
 'errback': None,
 'cookies': {},
 'headers': {},
 'dont_filter': False,
 '_meta': None,
 '_cb_kwargs': None,
 'flags': []}
```

Using `cls` the result would be:

```
{'_dumps_kwargs': {'sort_keys': True},
 '_encoding': 'utf-8',
 'method': 'GET',
 '_url': 'http://example.org',
 '_body': b'',
 'priority': 0,
 'callback': None,
 'errback': None,
 'cookies': {},
 'headers': {b'Content-Type': b'application/json',
  b'Accept': b'application/json, text/javascript, */*; q=0.01'},
 'dont_filter': False,
 '_meta': None,
 '_cb_kwargs': None,
 'flags': []}
```

(The same happens with `XmlRpcRequest`)
I think that the first approach could be better, because it respects the curl command, but I understand that for some users could be counter-intuitive.. seems a UX dilema. In any case, we should probably add something to the docstring. What do you think? @Gallaecio any thoughs?

Gallaecio: If we are not using `cls`, maybe we should reimplement the new class method in our built-in `Request` subclasses to raise a `NotImplementedError` exception.

immerrr: I think it's the same situation as with middlewares: scrapy should try to create the request with the inputs specified, and then if they are overridden by the class itself, or by the middlewares, you just need to be prepared to that it might happen and not expect 100% conformance.

noviluni: It seems a valid argument. We could use `cls` and extend the "caution" message regarding the middlewares adding that if it's used from `Request` subclasses they could add some modifications to the request.

noviluni: I add a commit with that idea. @kmike @Gallaecio tell me what you think. :)

Gallaecio:
The Python 3.7 test failures look like Scrapy’s typical random test failures.

PR #3869, opened 2019-07-12
[MRG+1][Py3] Check for 'return' with arguments inside generators

### LINKED ISSUE #3484: On better handling use cases of PEP-380 inside request callbacks
Recently I observed people writing something like this (when using Python 3.3+):

```
# coding: utf8

import scrapy


    name = 'test'
    start_urls = ['http://httpbin.org/get']

    def parse(self, response):
        yield scrapy.Request('http://httpbin.org/get?foo=bar')
        return {'url': response.url}
```

That won't work as expected as the `return <some value>` statement here, as defined in [PEP-380](https://www.python.org/dev/peps/pep-0380/), is just equivalent to `raise StopIteration(<some value>)`.
Sample job logs: [link](https://gist.github.com/starrify/98ab313c64ce2e2926e94d8d178b9f70)

Currently (as of dc65e75) Scrapy simply ignores the "returned" value, which might be causing confusions to users. Sample job log of the code above: [link](https://gist.github.com/starrify/5dabc9433c3d68552d64b841c885e0b2)

A proposal is to generate an error (or at least warning) when such cases happen (`StopIteration` observed to have an value).

elacuesta:
Fixes #3484

I also considered a version using `ast.NodeVisitor`, I think the `walk`-based solution _should_ be more efficient since it can stop the iteration if the conditions are satisfied, while the `NodeVisitor` class would visit _all_ `Return` nodes.

```python
class ReturnNodeVisitor(ast.NodeVisitor):

        super(ReturnNodeVisitor, self).__init__(*args, **kwargs)
        self.return_value = False

    def visit_Return(self, node):
        """
        Sets the 'return_value' instance variable to True if the value returned
        by the node is not None (either explicitly or implicitly)
        """
        if not is_safe_return_node(node):
            self.return_value = True


def check_gen_with_no_none_return_visitor(callable):
    """
    """
    if inspect.isgeneratorfunction(callable):
        tree = ast.parse(inspect.getsource(callable))
        node_visitor = ReturnNodeVisitor()
        node_visitor.visit(tree)
        if node_visitor.return_value:
            print('The generator has a return statement with a value different than None')
```

I plan on doing some benchmarking to determine if this has a noticeable performance impact, but I look forward to reading your feedback.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -9,7 +9,7 @@ from twisted.internet import defer
 
 from scrapy.utils.defer import defer_result, defer_succeed, parallel, iter_errback
 from scrapy.utils.spider import iterate_spider_output
-from scrapy.utils.misc import load_object
+from scrapy.utils.misc import load_object, warn_on_generator_with_return_value
 from scrapy.utils.log import logformatter_adapter, failure_to_exc_info
 from scrapy.exceptions import CloseSpider, DropItem, IgnoreRequest
 from scrapy import signals
@@ -17,6 +17,7 @@ from scrapy.http import Request, Response
 from scrapy.item import BaseItem
 from scrapy.core.spidermw import SpiderMiddlewareManager
 from scrapy.utils.request import referer_str
+
 
 logger = logging.getLogger(__name__)
 
@@ -99,11 +100,13 @@ class Scraper(object):
     def enqueue_scrape(self, response, request, spider):
         slot = self.slot
         dfd = slot.add_response_request(response, request)
+
         def finish_scraping(_):
             slot.finish_response(response, request)
             self._check_if_closing(spider, slot)
             self._scrape_next(spider, slot)
             return _
+
         dfd.addBoth(finish_scraping)
         dfd.addErrback(
             lambda f: logger.error('Scraper bug processing %(request)s',
@@ -123,7 +126,7 @@ class Scraper(object):
         callback/errback"""
         assert isinstance(response, (Response, Failure))
 
-        dfd = self._scrape2(response, request, spider)  # returns spiders processed output
+        dfd = self._scrape2(response, request, spider)  # returns spider's processed output
         dfd.addErrback(self.handle_spider_error, request, response, spider)
         dfd.addCallback(self.handle_spider_output, request, response, spider)
         return dfd
@@ -142,7 +145,10 @@ class Scraper(object):
     def call_spider(self, result, request, spider):
         result.request = request
         dfd = defer_result(result)
-        dfd.addCallbacks(callback=request.callback or spider.parse,
+        callback = request.callback or spider.parse
+        warn_on_generator_with_return_value(spider, callback)
+        warn_on_generator_with_return_value(spider, request.errback)
+        dfd.addCallbacks(callback=callback,
                          errback=request.errback,
                          callbackKeywords=request.cb_kwargs)
         return dfd.addCallback(iterate_spider_output)
@@ -172,8 +178,8 @@ class Scraper(object):
         if not result:
             return defer_succeed(None)
         it = iter_errback(result, self.handle_spider_error, request, response, spider)
-        dfd = parallel(it, self.concurrent_items,
-            self._process_spidermw_output, request, response, spider)
+        dfd = parallel(it, self.concurrent_items, self._process_spidermw_output,
+                       request, response, spider)
         return dfd
 
     def _process_spidermw_output(self, output, request, response, spider):
@@ -200,8 +206,7 @@ class Scraper(object):
         """Log and silence errors that come from the engine (typically download
         errors that got propagated thru here)
         """
-        if (isinstance(download_failure, Failure) and
-                not download_failure.check(IgnoreRequest)):
+        if isinstance(download_failure, Failure) and not download_failure.check(IgnoreRequest):
             if download_failure.frames:
                 logger.error('Error downloading %(request)s',
                              {'request': request},
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -8,6 +8,7 @@ This module must not depend on any module outside the Standard Library.
 import collections
 import copy
 import warnings
+import weakref
 from collections.abc import Mapping
 
 from scrapy.exceptions import ScrapyDeprecationWarning
@@ -240,7 +241,6 @@ class LocalCache(collections.OrderedDict):
     """Dictionary with a finite number of keys.
 
     Older items expires first.
-
     """
 
     def __init__(self, limit=None):
@@ -252,6 +252,35 @@ class LocalCache(collections.OrderedDict):
             while len(self) >= self.limit:
                 self.popitem(last=False)
         super(LocalCache, self).__setitem__(key, value)
+
+
+class LocalWeakReferencedCache(weakref.WeakKeyDictionary):
+    """
+    A weakref.WeakKeyDictionary implementation that uses LocalCache as its
+    underlying data structure, making it ordered and capable of being size-limited.
+
+    Useful for memoization, while avoiding keeping received
+    arguments in memory only because of the cached references.
+
+    Note: like LocalCache and unlike weakref.WeakKeyDictionary,
+    it cannot be instantiated with an initial dictionary.
+    """
+
+    def __init__(self, limit=None):
+        super(LocalWeakReferencedCache, self).__init__()
+        self.data = LocalCache(limit=limit)
+
+    def __setitem__(self, key, value):
+        try:
+            super(LocalWeakReferencedCache, self).__setitem__(key, value)
+        except TypeError:
+            pass  # key is not weak-referenceable, skip caching
+
+    def __getitem__(self, key):
+        try:
+            return super(LocalWeakReferencedCache, self).__getitem__(key)
+        except TypeError:
+            return None  # key is not weak-referenceable, it's not cached
 
 
 class SequenceExclude(object):
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -1,13 +1,18 @@
 """Helper functions which don't fit anywhere else"""
+import ast
+import inspect
 import os
 import re
 import hashlib
+import warnings
 from contextlib import contextmanager
 from importlib import import_module
 from pkgutil import iter_modules
+from textwrap import dedent
 
 from w3lib.html import replace_entities
 
+from scrapy.utils.datatypes import LocalWeakReferencedCache
 from scrapy.utils.python import flatten, to_unicode
 from scrapy.item import BaseItem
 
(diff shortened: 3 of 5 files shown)


elacuesta:
Right, before Python 3.3 `'return' with argument inside generator` was a `SyntaxError`, I need to skip the test in py2. Good news is the py3 tests are green \o/

scrapy/utils/misc.py
```diff
@@ -162,3 +164,49 @@ def set_environ(**kwargs):
...
+_trailing_whitespaces_re = re.compile(r'^(\s*)')
+
+
+def scan_generator_for_return_statement(callable):
+    """
+    Returns True if a callable is a generator function and includes a
+    'return' statement with a value different than None, False otherwise
+    """
+
+    def reduce_indentation_level(callable_lines):
```
elacuesta: I'm not really fond of this, suggestions are welcome.

elacuesta:
Travis error is unrelated

elacuesta:
Closing and reopening to force the Travis build to run again.

tests/test_utils_misc/test_return_with_argument_inside_generator.py
```diff
@@ -0,0 +1,39 @@
...
+
+        def j():
+            yield 1
+
+        assert is_generator_with_return_value(f)
+        assert is_generator_with_return_value(g)
+        assert not is_generator_with_return_value(h)
+        assert not is_generator_with_return_value(i)
+        assert not is_generator_with_return_value(j)
+
```
Gallaecio: I can think of a test case (expected failure?) that we might want to cover here: A test that verifies that, if something like `yield from` is used and this issue affects the nested generator, we will still not detect it because we do not inspect callbacks recursively (which I assume would make things much more complicated and/or have a potentially-significant negative impact on performance).

elacuesta: You're absolutely right. The nested generator could even be defined in a different module, it could get messy to get the source code to build the AST.

scrapy/utils/misc.py
elacuesta: Exactly what I needed, thanks! Too bad it's not available on py2, but hopefully that won't be an issue for too long :crossed_fingers:

elacuesta:
I should also write a note about this on the docs.

scrapy/utils/misc.py
elacuesta: Well, actually there's no need to worry about that, precisely because of https://github.com/scrapy/scrapy/pull/3869#issuecomment-510744562

docs/topics/spiders.rst
```diff
@@ -42,6 +42,19 @@ Even though this cycle applies (more or less) to any kind of spider, there are
 different kinds of default spiders bundled into Scrapy for different purposes.
 We will talk about those types here.
 
+.. note:: Spider callbacks can be implemented as ``generators``. In Python 3.3+
```
Gallaecio: I don’t think we need to document any of this, at least not in this page.

We could add an entry to the FAQ if we though some people may need additional information on top of the warning. However, in this case, I think the warning message is quite detailed, and already points to Python documentation about the topic.

scrapy/core/scraper.py
```diff
@@ -18,6 +19,11 @@
 from scrapy.core.spidermw import SpiderMiddlewareManager
 from scrapy.utils.request import referer_str
 
+
+if six.PY3:
```
Gallaecio: As @kmike mentioned on other pull requests, ``if not six.PY2:`` would be more future-proof. Same a few lines below.

docs/topics/spiders.rst
elacuesta: Fair enough. Although it took me a while to write this and be happy about the wording :cry:

scrapy/utils/misc.py
```diff
@@ -162,3 +164,38 @@ def set_environ(**kwargs):
...
+                    return True
+        return False
+
+    def warn_on_generator_with_return_value(spider, callable, logger):
+        """
+        Logs a warning if a callable is a generator function and includes
+        a 'return' statement with a value different than None
+        """
+        if is_generator_with_return_value(callable):
+            logger.warning(
```
kmike: What do you think about using warnings.warn instead? Otherwise this message will be logged at each request. By using warnings you'll also avoid passing logger argument.

scrapy/core/scraper.py
```diff
@@ -142,7 +150,11 @@ def _scrape2(self, request_result, request, spider):
     def call_spider(self, result, request, spider):
         result.request = request
         dfd = defer_result(result)
-        dfd.addCallbacks(callback=request.callback or spider.parse,
+        callback = request.callback or spider.parse
+        if not six.PY2:
+            warn_on_generator_with_return_value(spider, callback, logger)
+            warn_on_generator_with_return_value(spider, request.errback, logger)
```
kmike: Do you know if there is any pefrormance hit of doing code inspection on each request? I'm mostly concerned about pypy.

scrapy/utils/misc.py
elacuesta: Good point. I was passing the logger so the warning would come from `scrapy.core.scraper`, but I guess that's really not relevant. About the multiple warnings, I hadn't realized about that, thanks. Will `warning.warn` be issued only once?

scrapy/core/scraper.py
elacuesta: Yup, I need to run some benchmarks. Stay tuned!

scrapy/utils/misc.py
kmike: yes, warning.warn is only issued once

elacuesta: Updated, thanks for the suggestion. Seems like the default class is [`UserWarning`](https://docs.python.org/3/library/exceptions.html#UserWarning), which seems reasonable enough to me according to its description "Base class for warnings generated by user code."

scrapy/core/scraper.py
elacuesta: Great idea. I often forget about `lru_cache`. The default maxsize is 128, which seems perhaps a bit too much (I don't think I have seen spiders with that many callbacks). I don't have a strong opinion on it, so feel free to suggest other values.

Gallaecio: Mind https://stackoverflow.com/questions/11815873/memoization-library-for-python-2-7

Gallaecio: Waiting until Python 3 makes more sense now :)

elacuesta: Actually the whole PR works only in PY3, returns with values in generators are syntax errors in PY2

elacuesta: That's why I'm skipping the tests in py2 and importing/defining the functions in py3 only.

elacuesta: But since v1.7 was just released I can just remove all of those checks now :-)

Gallaecio: Right, I forgot about the `six.PY2` ifs!

> But since v1.7 was just released I can just remove all of those checks now :-)

Actually, there might be a 1.8, so don’t start removing `six` references just yet :slightly_smiling_face:

elacuesta: Oh I just did hahaha. Nevermind, I can just revert the commit :-)

elacuesta:
@kmike:

`scrapy-bench` @ f03011d, on my admittedly old i3-4160

In some tests I modified [this line](https://github.com/scrapy/scrapy-bench/blob/f03011d63f4149eeaad6dd7908ffa6fcbbb3304e/books/books/spiders/followall.py#L55) from `return r` to `yield from r`, to trigger the code inspection.

Command: `scrapy-bench --n-runs=20 --book_url http://domain1:8880/index.html bookworm`

* Scrapy `master` branch @ ae4eab98
    `py36`, `return r`
    Mean : 59.91042690444443 Median : 61.46755669534238 Std Dev : 4.994336655654268

    `pypy`, `return r`
    Mean : 45.292893988704805 Median : 45.20211211580239 Std Dev : 1.0998095175238132

    `py36`, `yield from r`
    Mean : 61.68034161779178 Median : 61.85817134315842 Std Dev : 2.102064300135808
    
    `pypy`, `yield from r`
    Mean : 43.366948759982265 Median : 43.56445793776966 Std Dev : 2.036978866845991


* Scrapy `detect_return_in_generator_callbacks` branch @ 8ed9e806
    `py36`, `return r`
    Mean : 54.98886208722822 Median : 57.48276553924386 Std Dev : 7.392744405205249
    
    `pypy`, `return r`
    Mean : 43.1211460939031 Median : 43.05007627473238 Std Dev : 0.7440402187424218
    
    `py36`, `yield from r`
    Mean : 55.381843120702726 Median : 60.00684786392303 Std Dev : 8.969017328279934
    
    `pypy`, `yield from r`
    Mean : 42.526517136254654 Median : 43.12169432034433 Std Dev : 2.2184298213790634

There doesn't seem to be a noticeable impact on performance (in fact, there is a few seconds improvement, but that's probably just by chance :stuck_out_tongue:). If I'm not mistaken there is only one callback in this spider, and the function call is cached now, so the results might be partial.

elacuesta:
Travis failure is due to the fact that I removed the py2 compatibility, I can revert that commit at any time if there is in fact going to be a 1.8 release.

Gallaecio:
Taking https://stackoverflow.com/questions/11815873/memoization-library-for-python-2-7 into account, and since adding a new Python 2-specific dependency for this at this point may not be wise, it may still make sense to make your changes only work on Python 3, using a `six.P2` check as you used to.

scrapy/utils/misc.py
```diff
@@ -162,3 +165,40 @@ def set_environ(**kwargs):
                 del os.environ[k]
             else:
                 os.environ[k] = v
+
+
+if not six.PY2:
+    from functools import lru_cache
+    from textwrap import dedent
+
+    @lru_cache(maxsize=128)
```
kmike: My only concern is this cache function: I wonder how it works with e.g. partial functions passed as callbacks - sometimes people use them to pass additional data. Won't it keep bound arguments of these functions for a long time? E.g. imagine something like `Request(url, partial(self.parse2, prev_response=response))` - would ~128 responses be kept alive in is_generator_with_return_value cache after requests are processed? I haven't check it, but I'm not sure lru_key uses weakrefs for keys by default.

Writing such code in spiders is not a good style, but here the tradeoff is a nice-to-have warning vs increased memory usage (and probably a failure) of some spiders.

elacuesta: Indeed, `lru_cache` does keep references to the cached objects (https://bugs.python.org/issue19859).
I'm working on a weak referenced cache implementation (based on `scrapy.utils.datatypes.LocalCache`) that does not add much overhead IMHO, will update shortly.

elacuesta: Updated, please see [`scrapy.utils.datatypes.LocalWeakReferencedCache`](https://github.com/scrapy/scrapy/pull/3869/files#diff-2f75773d25e485f98def0ba43b0ce78eR324)

scrapy/utils/datatypes.py
```diff
@@ -307,19 +308,48 @@ class LocalCache(collections.OrderedDict):
...
     """
 
     def __init__(self, limit=None):
         super(LocalCache, self).__init__()
         self.limit = limit
 
     def __setitem__(self, key, value):
-        while len(self) >= self.limit:
-            self.popitem(last=False)
+        if self.limit:
```
elacuesta: This is a bugfix: without the `if` statement, it can fail with `TypeError: '>=' not supported between instances of 'int' and 'NoneType'`

elacuesta: See #4123

kmike:
Thanks @elacuesta! As usual, thanks @Gallaecio for the review :)

Digenis:
This triggers an [issue](https://github.com/hylang/hy/issues/1696) when writing spiders in hy-lang

Was [pep 484#annotating-generator-functions-and-coroutines](https://peps.python.org/pep-0484/#annotating-generator-functions-and-coroutines) considered when solving this?

I know my hy-lang scenario is extreme
but I don't think anyone would disagree that such things usually belong to static analysis.
Is it postponed for #4041 or is it more complicated than it seems?

Gallaecio:
I don’t think we considered annotations as a solution here, but I am not sure they are a good fit because the point was to log something at run time without requiring users to run a static analysis tool.

Issue #3870, opened 2019-07-12
Allow to pass objects in Settings?


nyov:
See code example; why can I not reference plain objects into `Settings()`, but need to let Scrapy handle the import magic?
Would it make sense to have this? it seems "unclean" to do this in the usual `settings.py` environment, but in a single-script setup it looks less convoluted than to refer scrapy to import from current module?

```python3
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
#
import logging
import scrapy

logger = logging.getLogger(__name__)


class Spider(scrapy.Spider):

    name = 'Spidy'

    def start_requests(self):
        yield scrapy.Request('https://scrapy.org/')

    def parse(self, response):
        logger.info('Here I fetched %s for you. [%s]' % (response.url, response.status))
        return {
            'status': response.status,
            'url': response.url,
            'test': 'item',
        }


class LogPipeline(object):

    def process_item(self, item, spider):
        logger.warning('HIT ME PLEASE')
        logger.info('Got hit by:\n %r' % item)
        return item


if __name__ == "__main__":
    from scrapy.settings import Settings
    from scrapy.crawler import CrawlerProcess

    settings = Settings(values={
        'TELNETCONSOLE_ENABLED': False, # necessary evil :(
        'EXTENSIONS': {
            'scrapy.extensions.telnet.TelnetConsole': None,
        },
        'ITEM_PIPELINES': {
            #'myproject.pipelines.LogPipeline': 800, # << as resolved by `scrapy.utils.project.get_project_settings()`
            #'__main__.LogPipeline': 800, # << works, but still resolved and imported by scrapy magic

            LogPipeline: 800, # << WHY CAN'T I DO THIS?
        },
    })

    spider = Spider()

    process = CrawlerProcess(settings=settings)
    process.crawl(spider)
    process.start()
```

kmike:
Thanks for bringing it up, I don't think we have a dedicated ticket for this. +1 to add this feature.
 
It has been discussed in https://github.com/scrapy/scrapy/pull/1215, https://github.com/scrapy/scrapy/pull/1032; I recall https://github.com/scrapy/scrapy/pull/1272/files also has an implementation for this.

nyov:
Oh, now that you bring those links, I faintly remember... 
In fact I have been using the same or similar code as in #1032 myself, to work around this issue back in Scrapy 0.24 already. haha:
```python
# haxx: sad monkeypatch, might break
from importlib import import_module
def load_object(path):
[...]

PR #1215, opened 2015-05-07
allow additon of objects instead of paths to settings

youzhonghui:
Make this available:

```python
class CustomHttpProxyMiddleware(object):
    ...

def main():
    settings = Settings({
        DOWNLOADER_MIDDLEWARES{
            CustomHttpProxyMiddleware: 400    #! Here
        }
    })
    crawler = CrawlerProcess(settings)
    crawler.crawl(spider)
    crawler.start()
```

PR #1032, opened 2015-02-01
allow for direct classes

h4ck3rm1k3:
This is more of a hack but would like top open a discussion.

The patch will allow you to specify a class directly instead of a path name for loading. 

Usage here  https://github.com/flosokaks/scrapy_ks/blob/master/scrapy_openkansas.py

PR #3873, opened 2019-07-12
Allow passing classes directly in Settings

nyov:
Implements #3870 (See MWE code over there):
- Pass class objects directly in Settings(), instead of needing to go through Scrapy magic.

As @kmike kindly pointed out, this functionality was previously PR'd in some form in #1215, #1032, and as part of Addons.
However it never got done. Let's get this done?


kmike:
The implementation looks good to me.

I think to make it in it'd be awesome to have tests and docs for the end-user feature - class objects in Settings.

nyov:
Hm, I could write some docs. The Settings-related tests... are a boatload.
Tbh, I don't even know where I should add a case there and where not, as I don't grasp the whole `Settings`-functionalities since they got the `Attributes` and priorities layers.

I will have a think on that, and need to find some time to parse that code.

kmike:
you may have a test similar to your example - create a custom middleware class, pass it in settings, check that it works

nyov:
I tried. Sorry to disappoint, it seems that's not something I can manage in a reasonable amount of time.
Oh well, I found this: https://github.com/jdemaeyer/scrapy/commit/7fca7b59b87f9700a2ccb1cd7a2ade60497605f3
Would these work?

Gallaecio:
It may also be worth adding a test that checks that non-callables non-strings raise a `TypeError`.

kmike:
> But do we need those MW-changes over there, also?

@nyov I think these changes were to support object instances, not object classes; I'd prever to leave them out for now.

nyov:
Thanks for the hints!
I'm not sure if those tests now are suitable or good enough, probably not. But maybe you can give some more hints now that I did some work? :p 
The paragraph in the docs also doesn't sound really awesome, but I didn't manage to make it more concise without sacrificing legibility for (likely non-native) readers. Improvements welcome.

scrapy/utils/misc.py
```diff
@@ -33,10 +33,19 @@ def arg_to_iter(arg):
...
 
-    object can be a class, function, variable or an instance.
+    If ``path`` is not a string, but an object, then return it as is.
+
+    The object can be a class, function, variable or an instance.
     path ie: 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware'
     """
 
+    if not isinstance(path, six.string_types):
+        if callable(path):
```
Gallaecio: What if it’s a non-callable variable or object?

nyov: Sorry this has languished so.

But I'm unsure how to treat the logic then.
`callable()` will return True even for some objects that aren't (like an empty class).
But for variables, it's not as simple, if they're strings they could be meant as a classpath to load, or a variable to pass as-is. The distinction would be hard to make.

So should the logic for `load_object` just return anything that doesn't happen to be a string which has a dot (the check for valid classpath), or fails to import?

Gallaecio: I’m OK with checking that is it not a string and returning it otherwise; I assume a setting that is loaded using `load_object` will not allow a string as value, and we really not want to silence import errors due to typos.

Gallaecio:
I took the liberty to refactor the documentation:

- Move it to a section of its own, so that the scope of `versionadded` is clearer
- Refactor the first two paragraph so that they do not speak in terms of before and after (more fitting for the release notes), but rather describe the current situation.
- Removed the inline examples, as the example below show suffice.
- Removed the note about deprecation handling; it was not 100% accurate (deprecation checks should still work, and happen when those objects are used at run time; fixing the paths of _removed_ imports is was will not be there, but I don’t think that needs to be described explicitly)

Gallaecio (approved):
While I still think we could support anything but a string, I’m OK with only supporting callables for the time being.

scrapy/utils/misc.py
Gallaecio: https://github.com/scrapy/scrapy/pull/3873#pullrequestreview-471665059

tests/test_settings/__init__.py
```diff
@@ -385,6 +385,38 @@ def test_getdict_autodegrade_basesettings(self):
...
+
+        mypipeline, priority = settings.getdict('ITEM_PIPELINES').popitem()
+        self.assertEqual(priority, 800)
+        self.assertEqual(mypipeline, TestPipeline)
+        self.assertIsInstance(mypipeline(), TestPipeline)
+        self.assertEqual(mypipeline().process_item('item', None), 'item')
+
+        myhandler = settings.getdict('DOWNLOAD_HANDLERS').pop('ftp')
+        self.assertEqual(myhandler, FileDownloadHandler)
+        myhandler_instance = create_instance(myhandler, None, get_crawler())
```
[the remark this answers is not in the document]
Gallaecio: I looked for existing tests that used import paths in settings for objects that were defined in the test files themselves, and switched them to use regular object references.

Should we consider this covered with that, or should we write a specific test for this as well?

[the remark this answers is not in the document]
kmike: Nice idea to repurpose existing tests for that 👍 I'd consider this covered.

scrapy/utils/misc.py
```diff
@@ -39,10 +39,19 @@ def arg_to_iter(arg):
 def load_object(path):
     """Load an object given its absolute object path, and return it.
 
-    object can be the import path of a class, function, variable or an
-    instance, e.g. 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware'
+    The object can be the import path of a class, function, variable or an
+    instance, e.g. 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware'.
+
+    If ``path`` is not a string, but a callable() object, then return it as is.
```
kmike, line 45: What do you think about mentioning that classes are callable? I think this is a rather arcane knowledge, but it is the main use case here.

kmike (approved):
Thanks @nyov and @Gallaecio, looks great! +1 to merge.

I've left a minor comment about docs; feel free to merge with or without addressing iit @Gallaecio.

PR #3876, opened 2019-07-13
Prevent DeprecationWarning(s)

elacuesta:
`assertRaisesRegexp` was renamed to `assertRaisesRegex` in 3.2 and the old name was deprecated: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex. Not sure when the change happened in Twisted.

This should ease the transition into Scrapy 2.0 (dropping the support for py2)

### DIFF
--- a/scrapy/_monkeypatches.py
+++ b/scrapy/_monkeypatches.py
@@ -14,6 +14,11 @@ if sys.version_info[0] == 2:
         from urlparse import uses_query
         uses_query.append('s3')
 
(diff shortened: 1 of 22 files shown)


elacuesta:
Inspired by (stolen from? :stuck_out_tongue:) https://github.com/buildbot/buildbot/pull/2785/files#diff-f0a01b227c22bed15ccc305b05bc12c4

elacuesta:
Well, this is no longer only about `TestCase`, but it's still safe IMHO :-D

PR #3846, opened 2019-06-27
Write the 1.7 release notes

Gallaecio:
Current coverage: https://github.com/scrapy/scrapy/commit/44eb21aa51cc0212e68dc0709aaa5c10bfe64e7e

Includes a note about dropping Python 2 support in Scrapy 2.0.

Fixes #3872.

PR #3877, opened 2019-07-14
Prevent deprecation warnings

elacuesta:
I messed up #3876, had to open a new PR :man_facepalming: 

I left out some Twisted warnings, I believe they are more delicate and deserve their own treatment.

### DIFF
--- a/scrapy/_monkeypatches.py
+++ b/scrapy/_monkeypatches.py
@@ -14,6 +14,11 @@ if sys.version_info[0] == 2:
         from urlparse import uses_query
         uses_query.append('s3')
 
+    # Prevent the DeprecationWarning about SafeConfigParser -> ConfigParser
+    import configparser
+    if not getattr(configparser, 'ConfigParser', None):
+        configparser.ConfigParser = configparser.SafeConfigParser
+
 
 # Undo what Twisted's perspective broker adds to pickle register
 # to prevent bugs like Twisted#7989 while serializing requests
--- a/scrapy/_monkeypatches.py
+++ b/scrapy/_monkeypatches.py
@@ -1,7 +1,8 @@
-import sys
+import six
 from six.moves import copyreg
 
-if sys.version_info[0] == 2:
+
+if six.PY2:
     from urlparse import urlparse
 
     # workaround for https://bugs.python.org/issue7904 - Python < 2.7
@@ -13,11 +14,6 @@ if sys.version_info[0] == 2:
     if urlparse('s3://bucket/key?key=value').query != 'key=value':
         from urlparse import uses_query
         uses_query.append('s3')
-
-    # Prevent the DeprecationWarning about SafeConfigParser -> ConfigParser
-    import configparser
-    if not getattr(configparser, 'ConfigParser', None):
-        configparser.ConfigParser = configparser.SafeConfigParser
 
 
 # Undo what Twisted's perspective broker adds to pickle register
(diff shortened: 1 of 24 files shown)


scrapy/_monkeypatches.py
```diff
@@ -14,6 +14,11 @@
         from urlparse import uses_query
         uses_query.append('s3')
 
+    # Prevent the DeprecationWarning about SafeConfigParser -> ConfigParser
+    import configparser
+    if not getattr(configparser, 'ConfigParser', None):
+        configparser.ConfigParser = configparser.SafeConfigParser
```
kmike: I think a deprecation warning in tests doesn't worth monkey-patching stdlib

scrapy/item.py
```diff
@@ -4,16 +4,22 @@
...
-from abc import ABCMeta
 import six
 
 from scrapy.utils.trackref import object_ref
 
 
+if six.PY3:
+    MutableMapping = collections.abc.MutableMapping
+else:
+    MutableMapping = collections.MutableMapping
```
kmike: too bad https://github.com/benjaminp/six/issues/155 is still open

tests/__init__.py
```diff
@@ -35,3 +35,18 @@ def get_testdata(*paths):
     path = os.path.join(tests_datadir, *paths)
     with open(path, 'rb') as f:
         return f.read()
+
+
+# FIXME: delete after dropping py2 support
+# Monkey patch the unittest module to prevent the
+# DeprecationWarning about assertRaisesRegexp -> assertRaisesRegex
```
kmike: can we solve it without a monkey patch?

tests/__init__.py
```diff
@@ -35,3 +35,18 @@ def get_testdata(*paths):
...
+import sys
+if sys.version_info[0] == 2:
```
kmike: six.PY2

scrapy/item.py
```diff
@@ -4,16 +4,22 @@
...
 from copy import deepcopy
+import collections
 
-from abc import ABCMeta
 import six
 
 from scrapy.utils.trackref import object_ref
 
 
+if six.PY3:
```
kmike: It is better to use six.PY2 instead of six.PY3, because the condition may stop working (or stop making sense) when Python 4 is released

tests/test_utils_datatypes.py
```diff
@@ -1,11 +1,18 @@
 import copy
 import unittest
-from collections import Mapping, MutableMapping
+
+import six
+if six.PY3:
```
kmike: it is better to use six.PY2

tests/__init__.py
elacuesta: I see this as a very disposable piece of code that should go away once we drop Py2 support. Any particular reason for using `six`? Should we change https://github.com/scrapy/scrapy/pull/3877/files#diff-44aa4a970c57972e05d57dba01077acaL4 too?

scrapy/_monkeypatches.py
elacuesta: Makes sense, but the class used in `scrapy.utils.conf.get_config`, I think we should still do something about it. I'll change monkey-patching for a `try... except ImportError` approach.

tests/__init__.py
elacuesta: Right, according to #3846 it's "planned for Q4 2019 or Q1 2020" (:tada:) given that this is only used in the tests it should be fine IMHO

kmike:
Thanks @elacuesta! I still feel uncomfortable with merging a monkey-patch to stdlib module, but as it only happens when Scrapy tests are executed (not when Scrapy itself is imported), this should be fine.

PR #3878, opened 2019-07-14
Deprecate scrapy.utils.datatypes.MergeDict in favor of collections.ChainMap

elacuesta:
Use `collections.ChainMap` when available, fall back to `scrapy.utils.datatypes.MergeDict` (currently in use only in `scrapy.loader.processors.MapCompose`)

### DIFF
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -245,6 +245,13 @@ class MergeDict(object):
     first occurrence will be used.
     """
     def __init__(self, *dicts):
+        if six.PY3:
+            warnings.warn(
+                "scrapy.utils.datatypes.MergeDict is deprecated in favor "
+                "of collections.ChainMap (introduced in Python 3.3)",
+                category=ScrapyDeprecationWarning,
+                stacklevel=2,
+            )
         self.dicts = dicts
 
     def __getitem__(self, key):
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -245,7 +245,7 @@ class MergeDict(object):
     first occurrence will be used.
     """
     def __init__(self, *dicts):
-        if six.PY3:
+        if not six.PY2:
             warnings.warn(
                 "scrapy.utils.datatypes.MergeDict is deprecated in favor "
                 "of collections.ChainMap (introduced in Python 3.3)",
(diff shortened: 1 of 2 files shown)


scrapy/utils/datatypes.py
```diff
@@ -245,6 +245,13 @@ class MergeDict(object):
     first occurrence will be used.
     """
     def __init__(self, *dicts):
+        if six.PY3:
```
kmike: it is better to use six.PY2

elacuesta: Do you mean issue the warning in py2 or use `six.PY2` but negated? My reasoning for this was that it doesn't make much sense to say "use collections.ChainMap" in py2 if it's not available. And since it was introduced in py3.3 and Scrapy supports py3.4+ we are safe in that regard.

kmike: the same reason as https://github.com/scrapy/scrapy/pull/3877/files#discussion_r303310439 :)

elacuesta: I'm sorry but I still don't get it :confused:
If we raised this warning in Py2, users who import `ItemLoader` would see it and will have no idea what's it about, because it's used internally.
I put this warning here in case someone imports MergeDict for standalone use, which would be reasonable in Py2 since ChainMap is not available, but I think suggesting to use ChainMap instead doesn't make much sense.

Gallaecio: Using a negated `six.PY2` the code will work on Python 4 as well.

elacuesta: Ok that explains it. I'll change it. but to be fair, I was thinking about removing this once we dropped support for py2. I think a big six cleanup in in order for Scrapy 2.0.
And for what I could found, there is no timeline for py4. Guido once linked an article from a CPython core dev (https://twitter.com/gvanrossum/status/501172370690699265) that says that for him py4 would be "the release that comes after Python 3.9" (personally I hope it's not the case, it would be a gross semver violation) but an update to that article adds that "discussions on the core python-dev mailing list led to the conclusion that the release after 3.9 will probably just be 3.10".

kmike:
A good catch, thanks @elacuesta!

elacuesta:
By all means, thanks to you both @kmike and @Gallaecio :muscle:

PR #3881, opened 2019-07-15
Support for dataclass and attrs items

### LINKED ISSUE #3761: Items as Dataclasses
Is there any plan to support dataclasses as `yield`-able results from crawlers (and possibly `meta` in `Request`/`Result`)?

It would be a very welcome addition for us, mostly for typing reasons, IDE completion, additional logic.
And it should also be simple to implement (just call `dataclasses.asdict()`)?

elacuesta:
Fixes #3761, closes #2807

It seems to me like there would be a limitation in the way we could use dataclassess with ItemLoaders. Given the fact that all fields are supposed to be provided (except for those with default values, or unless a custom constuctor is provided) and that loaders try to _append_ values, I think only dataclasses defined as the following would work:
```python
from dataclasses import dataclass, field

@dataclass
class InventoryItem:
    name: list = field(default_factory=list)
    price: list = field(default_factory=list)
    stock: list = field(default_factory=list)

lo = ItemLoader(item=InventoryItem())
lo.add_value('name', 'foo bar')
print(lo.load_item())  # prints "InventoryItem(name=['foo bar'], price=[], stock=[])"
```

Thanks in advance for your feedback!

---

**Update**

Checklist:

- [x] Scraper class changes
- [x] Exporters support
- [x] Item loader support
- [x] Parse command
- [x] Shell command
- [x] Contracts
- [x] Docs
- [x] Tests

This patch works with the built-in [`dataclassess` module](https://docs.python.org/3.7/library/dataclasses.html) in Python 3.7+, and with the [`dataclassess` backport](https://pypi.org/project/dataclasses/) in Python 3.6.

**Update 2**
Checklist for the new implementation based on the `ItemAdapter` class
- [x] ItemAdapter class
- [x] Refactor existing components to use ItemAdapter class
- [x] ItemAdapter tests
- [x] docs (WIP)

**Update 3:**
Supporting [`python-attrs`](https://www.attrs.org) now too. Implementation is based on https://github.com/scrapy/itemadapter.


elacuesta:
Minimal spider, with feed export support:

```python
from dataclasses import dataclass

from scrapy import Spider


@dataclass
class InventoryItem:
    name: str
    price: float
    stock: int


class DataClassSpider(Spider):
    name = 'example'
    start_urls = ['https://example.org']

    def parse(self, response):
        return InventoryItem('Chair', 120.5, 10)
```

```
$ scrapy runspider example.py -o items.json
(...)
2019-07-15 13:16:06 [scrapy.core.engine] INFO: Spider opened
2019-07-15 13:16:06 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2019-07-15 13:16:06 [scrapy.extensions.telnet] INFO: Telnet console listening on 127.0.0.1:6023
2019-07-15 13:16:06 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://example.org> (referer: None)
2019-07-15 13:16:06 [scrapy.core.scraper] DEBUG: Scraped from <200 https://example.org>
InventoryItem(name='Chair', price=120.5, stock=10)
2019-07-15 13:16:06 [scrapy.core.engine] INFO: Closing spider (finished)
2019-07-15 13:16:06 [scrapy.extensions.feedexport] INFO: Stored json feed (1 items) in: items.json
(...)
```

```
$ cat items.json
[
{"name": "Chair", "price": 120.5, "stock": 10}
]
```

scrapy/core/scraper.py
```diff
@@ -182,7 +184,7 @@ def _process_spidermw_output(self, output, request, response, spider):
         """
         if isinstance(output, Request):
             self.crawler.engine.crawl(request=output, spider=spider)
-        elif isinstance(output, (BaseItem, dict)):
+        elif isinstance(output, (BaseItem, dict)) or is_dataclass_instance(output):
```
Gallaecio: Maybe we should implement a high-level helper function for use cases like this?
```suggestion
        elif is_item(output):
```

This would make our current FAQ entry about splitting an item into multiple items much more readable.

scrapy/shell.py
```diff
@@ -159,7 +156,7 @@ def get_help(self):
         return "\n".join("[s] %s" % l for l in b)
 
     def _is_relevant(self, value):
-        return isinstance(value, self.relevant_classes)
+        return isinstance(value, self.relevant_classes) or is_item_like(value)
```
Gallaecio: Looking at line 122 (old 125), I think this file does not need any change, given that `_is_relevant` is only used with the contents of `self.vars`.

scrapy/spiders/feed.py
```diff
@@ -31,7 +31,7 @@ def process_results(self, response, results):
         processing required before returning the results to the framework core,
         for example setting the item GUIDs. It receives a list of results and
         the response which originated that results. It must return a list of
-        results (Items or Requests).
+        results (Requests, Items, dicts, dataclasses).
```
Gallaecio: :lipstick:  We could alternatively modify the string not to refer to classes, but concepts.
```suggestion
        results (items or requests).
```

scrapy/shell.py
elacuesta: Actually it seems to be necessary for the item to be displayed in the list of variables. It's just an aesthetic thing, since the variable is populated nonetheless, just not displayed.

```python
# inventory.py
from dataclasses import dataclass, field

@dataclass
class InventoryItem:
    name: list = field(default_factory=list)
    price: list = field(default_factory=list)
    stock: list = field(default_factory=list)
```

```
$ scrapy shell https://example.org -s DEFAULT_ITEM_CLASS=inventory.InventoryItem
2019-11-29 13:37:40 [scrapy.utils.log] INFO: Scrapy 1.8.0 started (bot: scrapybot)
(...)
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7fcd90b1cbe0>
[s]   item       InventoryItem(name=[], price=[], stock=[])
[s]   request    <GET https://example.org>
[s]   response   <200 https://example.org>
[s]   settings   <scrapy.settings.Settings object at 0x7fcd8c94d7f0>
[s]   spider     <DefaultSpider 'default' at 0x7fcd8b668f60>
[s] Useful shortcuts:
In [1]: item
Out[1]: InventoryItem(name=[], price=[], stock=[])
```

For reference, here is where the `item` variable gets populated: https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/shell.py#L125

kmike:
Hey! I think it requires more documentation, because it affects spider middlewares: process_spider_output method can now get dataclass instances. The same applies to item_scraped, item_dropped, item_error signal handlers, and to item pipelines.

If I'm not mistaken, most item pipelines will need to be changed. For example, does FilesPipeline work with dataclasses?

It is a bit trickier than with dict vs Item, because dicts and Items both support `item["field"]` access, while dataclasses don't, so some code which worked with items and dicts without any explicit checks will stop working for dataclasses.

I think we need to 

1. Document how to make your components ready for dataclasses.
2. After documenting (looking at all steps required), try to simplify it, if possible.

elacuesta:
Thanks for the review @kmike! I added a FAQ entry about this and some notes in the signals, pipelines and spider middleware pages, let me know what you think about this approach.
I had some problems with the code sample in the docs, [`sybil.parsers.skip.skip`](https://sybil.readthedocs.io/en/latest/parsers.html#skip) does not seem to be working correctly at 5a97a38 (I also tried with `.. skip: start` and `.. skip: end`, to no avail). I ended up ignoring the `docs/faq.rst`, but I'm not a fan of this, please let me know if you know a way to solve it (cc @Gallaecio, the original author of the Sybil integration).

Gallaecio:
@elacuesta This seems to do the job:
https://github.com/Gallaecio/scrapy/commit/df5a6d2d96d50a997900ce5a9df7f54ab375c709
https://github.com/Gallaecio/scrapy/commit/ea397d34191e6de6212bf53089416d2dc8da0e50

kmike:
Thanks @elacuesta for the update! It is much more complete now. Remaining concerns: 

* users can't really use dataclasses as items, because of the existing middlewares, item pipelines, etc. What we really support is `@subscriptable_dataclass`. I think that's a good idea to have something like that, to make items more compatible with existing pipelines, but it may be not enough. 
*  ^^ applies even for built-in components - e.g. FilesPipeline doesn't support dataclasses.
* We don't explain how to write middlewares which support dataclasses directly, and don't provide any helpers for that.

elacuesta:
Hey @kmike, thanks for the feedback :smile:

One thing I'd like to discuss about the `subscriptable_dataclass` thing: I'd prefer to take the burden of making dataclass items Scrapy-ready away from the user as much as possible, i.e. not _requiring_ them to decorate items in order for built-in components to work properly. With that in mind, the media pipelines "natively" handle dataclass items now (tests updated). This is done by converting `ìtem`s to `dict`s if necessary in their item processing methods, in order to populate such items with the URLs to be downloaded (without doing something [like this](https://stackoverflow.com/a/52536584), kinda hackish IMHO). The implication is that the output of those methods is not of the same type as the input.

Regarding spider middlewares, I added a note to the `process_spider_output` method docs pointing to the FAQ entry about dataclass items, which explains the implementation and mentions the `subscriptable_dataclass` decorator, I believe that should be enough.

I checked built-in components such as extensions, pipelines and middlewares, I think nothing else needs updating at this point but another pair of eyes checking this fact are always welcome. 

Looking forward to hearing your comments, thanks again!

mredaelli:
Sorry to barge in.

> The implication is that the output of those methods is not of the same type as the input.

Not sure if I understand this correctly: do you mean that the dataclass item is replaces by a dict in all the subsequent pipelines etc?

If so, my personal preference and suggestion would be *not* to do that.

Part of the reason I asked for dataclass support was exactly so that I could write pipelines and middlewares that could receive a dataclass and all its methods, validation, typechecking with mypy, etc.
We have several of those, all currently dealing with loose dicts, and we would love to move them to dataclasses.

If the issue is some missing fields, I wonder if that could be worked around by either asking the user to inherit the dataclass from a base class (as vanilla as possible -- even more than the current `Item`), or by a different annotation replacing `@dataclass` which also injects the fields?

elacuesta:
> do you mean that the dataclass item is replaces by a dict in all the subsequent pipelines etc?

Indeed, that's what I meant. I don't love the idea either, but I thought it was a reasonable compromise in order to make dataclass items work out of the box. I guess it's not.

This PR introduces a `subscriptable_dataclass` decorator, I wanted to avoid _requiring_ it but I think we can probably require in a first version, and eventually relax this requirement. That said, it currently doesn't support _adding_ new fields to the items, I'll need to change that.

mredaelli:
I wonder if using a base class would be easier instead -- but I haven't even checked the current implementation.

I say that because:
- there's less magic going around, which is always good (and if it is a matter of adding a field accessor and some required fields, it might even be simpler)
- it would be more broadly compatible. For instance, in my perfect world I would love to use not vanilla dataclasses, but [pydantic](https://github.com/samuelcolvin/pydantic/)'s version of the annotation, so that I can offload even validation to the item definition. Having a conflicting decorator in `scrapy` would make that probably much harder.

What do you think?

kmike:
Another magical option, for discussion: make a "proxy" class which provides dict-like access to objects, but modifies the original object. Make a decorator which wraps input items to this proxy objects, and unwraps them in the output. Users can then use this decorator in their middlewares to add support for dataclasses.

elacuesta:
After checking the [current implementation](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/pipelines/files.py#L471), there seems to be no need to do things differently: unless the item is a `dict`, both pipelines only populate the item with the result of the download if it already contains a field for it, we can just do exactly that with dataclass items.

docs/faq.rst
```diff
@@ -353,8 +353,52 @@ method for this purpose. For example::
...
+Support for :class:`dataclasses.dataclass` objects as items was added in version 2.0.
+This works natively in Python 3.7+, or using the `dataclasses backport`_ in Python 3.6.
+
+Most of the examples in this documentation assume you are using either :class:`dict`
+or :class:`~scrapy.item.Item` objects, and access their values in a dictionary-like
+manner. However, ``dataclass`` objects expose their values trough attributes instead,
+which means you might need to update your existing components
+(such as :ref:`item pipelines <topics-item-pipeline>`,
+:ref:`signal handlers <topics-signals>` handlers or
+:ref:`spider middlewares <topics-spider-middleware>`) to make them work correctly.
```
kmike: Could you please document helper functions here?

docs/faq.rst
```diff
@@ -353,8 +353,52 @@ method for this purpose. For example::
...
+
+Alternatively, you can use the ``scrapy.utils.decorators.subscriptable_dataclass``
+decorator, which adds the appropriate methods in order to make ``dataclass`` objects
+capable of being accessed like dictionaries:
```
kmike: I think we should suggest this decorator to make items work with third-party middlewares/signal handlers which are not updated yet to support dataclasses.

scrapy/utils/datatypes.py
```diff
@@ -291,3 +292,29 @@ def __init__(self, seq):
...
+
+
+def set_item_field(item, field_name, value):
+    """
+    Set item fields on both "regular" and dataclass-based items
+    """
+    if is_dataclass_instance(item):
+        if not hasattr(item, "_field_names"):
+            from dataclasses import fields
+            item._field_names = [f.name for f in fields(item)]
```
kmike: Is it a performance optimization, to cache available field names? I'm not sure that's good to add attributes to user's objects.

kmike: The same applies to subscriptable_dataclass.

scrapy/utils/misc.py
```diff
@@ -34,6 +34,10 @@ def arg_to_iter(arg):
         return [arg]
 
 
+def is_item_like(obj):
```
kmike: Could you please add a docstring and tests for this function?

scrapy/utils/python.py
```diff
@@ -1,19 +1,32 @@
...
+try:
+    from dataclasses import is_dataclass, asdict as dataclass_asdict
+except ImportError:
+    def is_dataclass_instance(_):
+        return False
+
+    def dataclass_asdict(_):
+        raise ImportError("no module named 'dataclasses'")
+else:
+    def is_dataclass_instance(obj):
```
kmike: Could you please add a bit more tests to this function?

tests/test_loader.py
```diff
@@ -528,7 +538,7 @@ def test_get_output_value_list(self):
         self.assertEqual(il.get_output_value('name'), ['foo', 'bar'])
         loaded_item = il.load_item()
         self.assertIsInstance(loaded_item, self.item_class)
-        self.assertEqual(loaded_item, dict({'name': ['foo', 'bar']}))
+        self.assertEqual(self.to_dict(loaded_item), dict({'name': ['foo', 'bar']}))
```
kmike: Is self.to_dict needed because item loaders can return dataclass instances now? If so, could you please add tests for this behavior?

kmike: Also, could you please check Item Loaders docs, and update it accordingly? I think it still says that Item instances are expected, or dict-like objects, which is no longer true.

tests/test_pipeline_images.py
```diff
@@ -117,43 +126,74 @@ def thumb_key(self, url, thumb_id):
...
-            url = 'http://www.example.com/images/1.jpg'
-            item = cls({'name': 'item1', 'image_urls': [url]})
-            pipeline = ImagesPipeline.from_settings(Settings({'IMAGES_STORE': 's3://example/images/'}))
-            requests = list(pipeline.get_media_requests(item, None))
-            self.assertEqual(requests[0].url, url)
-            results = [(True, {'url': url})]
-            pipeline.item_completed(results, item, None)
-            self.assertEqual(item['images'], [results[0][1]])
+        url = 'http://www.example.com/images/1.jpg'
+        item = self.item_class(name='item1', image_urls=[url])
```
kmike: Could you please add a dataclass example to media-pipeline.rst as well (see "Usage example" - it shows how to add a field for Item)?

elacuesta:
Hey @kmike, thanks for the thorough review. Indeed, seems like I'm missing a lot of things, I'll get to it ASAP. Thanks again!

scrapy/utils/python.py
elacuesta: This is an upstream recipe, taken verbatim from the official docs: https://docs.python.org/3/library/dataclasses.html#dataclasses.is_dataclass

tests/test_loader.py
elacuesta: Most testing methods The `InitializationTestMixin` have an assertion to check if the produced item matches the expected type: `self.assertIsInstance(loaded_item, self.item_class)`. Would that be enough?

kmike: Aha, I missed that. Yes, I think this is enough.

docs/faq.rst
```diff
@@ -360,6 +360,62 @@ Yes, by setting :setting:`DNS_RESOLVER` to ``scrapy.resolver.CachingHostnameReso
...
+>>> d["name"]
+'foobar'
+>>> d["price"] = 5
+>>> d
+InventoryItem(name='foobar', price=5)
+
+.. skip: end
+
+In addition, two helper functions are available to interact with items without
+having to check their base class:
```
kmike: Looking at how built-in components were updated, is_item_like and is_dataclass_instance functions were also useful.

scrapy/utils/decorators.py
```diff
@@ -43,3 +43,31 @@ def inthread(func):
...
+except ImportError:
+    pass
+else:
+    def subscriptable_dataclass(cls):
+        """
+        Allow dictionary-like access on dataclass instances
+        """
+
+        def __getitem__(self, key):
+            field_names = [f.name for f in dataclass_fields(self)]
```
kmike: `if key in [f.name for f in dataclass_fields(self)]` is repeated at least 5 times in the Scrapy codebase, I wonder if we should make a small helper for that.

scrapy/utils/python.py
kmike: Even if so, I think it is better to add tests, because the example in docs is only guaranteed to work with the Python these docs are for, while we support many Pythons, and a backport as well.

docs/faq.rst
```diff
@@ -360,6 +360,62 @@ Yes, by setting :setting:`DNS_RESOLVER` to ``scrapy.resolver.CachingHostnameReso
...
+
+Can I use dataclasses as items?
+-------------------------------
+
+Support for :class:`dataclasses.dataclass` objects as items was added in version 2.1.
+This works natively in Python 3.7+, or using the `dataclasses backport`_ in Python 3.6.
+
+Most of the examples in this documentation assume you are using either :class:`dict`
+or :class:`~scrapy.item.Item` objects, and access their values in a dictionary-like
+manner. However, ``dataclass`` objects expose their values trough attributes instead,
```
kmike:
```suggestion
manner. However, ``dataclass`` objects expose their values through attributes instead,
```

docs/faq.rst
```diff
@@ -360,6 +360,62 @@ Yes, by setting :setting:`DNS_RESOLVER` to ``scrapy.resolver.CachingHostnameReso
...
+which means you might need to update your existing components
+(such as :ref:`item pipelines <topics-item-pipeline>`,
+:ref:`signal handlers <topics-signals>` handlers or
+:ref:`spider middlewares <topics-spider-middleware>`) to make them work correctly.
+
+Alternatively, you can use the
+:func:`~scrapy.utils.decorators.subscriptable_dataclass` decorator:
```
kmike: I think we should think a bit more about the transition path, and update the docs accordingly. 

There are two possible ways people can adopt dataclasses.

Path 1: start using dataclasses. Realize that some middlewares in a project don't work. Use `@subscriptable_dataclass@dataclass` instead of `@dataclass`.

Path 2: start using dataclasses. Realize that some middlewares in a project don't work. Update the middlewares.

My feeling is that we're pushing users strongly to (1), because 

* There are no examples of how to write a middleware which is compatible with dataclasses, or how to update an existing middleware to work with dataclasses. 
* We don't have middleware examples in docs updated, so "by default" people are learning how to write non-compatible middlewares: "Most of the examples in this documentation assume you are using either :class:`dict`
or :class:`~scrapy.item.Item` objects";
* We don't mention that all built-in Scrapy components are updated to work with dataclasses (I think reader could get an impression they are not)
* We say "you might need to update your components ... [not sure how]. Alternatively, use a decorator [this is how]: ... This decorator is *also* useful when working with third-party components". 

I would prefer to push users to (2) more, i.e. to updating the middlewares, not using the decorator. The end result of (1) is that nobody can use dataclasses without a decorator, which means that Scrapy doesn't really work with third-party dataclasses. 

It will still be easier to throw a decorator to make it work, but we can reduce the gap significantly by providing helper functions and examples, promoting it as a good practice of writing middlewares. I think decorator should be presented as a "quick hack to make it work", and as a way to use a third-party component which is not updated yet, not as a default solution.

Thoughts? :)

Gallaecio:
While working on improving the documentation, I’m starting to think that we need to rethink the API we are going to provide for handling items (or “item-like objects”).

I don’t think we should have components use type-specific functions like `is_dataclass_instance` or `dataclass_asdict`. I’m hoping we can come up with something more generic, so that if we support yet a new type of item later, components only need to be upgraded to support new features (but not, for example, for reading or writing key-value pairs). 

Would it make sense to create an item adapter class that components can use to interact with input items without needing to care about the underlying type?

elacuesta:
@Gallaecio After our recent conversation I went to update the media pipelines and realized there were a lot more checks in other components as well (item loader and exporters, for instance) so I'm sold on the item adapter suggestion.
I just pushed one commit with said change (91a9a1721c7860603c37797b74bb075fa03b7ca8). Tests are broken because I haven't adapted them yet but I think the implementation looks much more clean now. There are no dataclass-related checks other than in the `scrapy.item` and `scrapy.utils.decorators` modules.
Please take a look and let me know what you think, I'll be updating the tests shortly.

scrapy/utils/item.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    try:
+        from dataclasses import is_dataclass
+    except ImportError:
+        return False
+    else:
+        return is_dataclass(obj) and not isinstance(obj, type)
+
+
+class ItemAdapter:
```
kmike, line 28: A good idea to have such class!

Do you think we can make it nicer - e.g. use a more dict-like API for it, or even MutableMapping? E.g. why .get_item_field, and not just .get? Why .set_item_field, and not `["key"] = value`? This could allow to reduce usages of as_dict

scrapy/utils/item.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            self.item[field] = value
+
+    def as_dict(self):
+        """
+        Return a class:`dict` instance with the same data as the stored item.
+        Returns the stored item unaltered if is already a dict object.
+        """
+        if isinstance(self.item, dict):
+            return self.item
```
kmike, line 64: Should it be a shallow copy, like in other cases? I think it is important to ensure that the behavior is consistent; it seems currently modifications in as_dict result apply to the original item if it was a dict, but not in other cases.

elacuesta: Very good point, thanks

scrapy/utils/item.py
kmike: Thanks for the update @elacuesta! I still wonder if we should implement MutableMapping interface. It requires to implement 5 methods: `__getitem__`, `__setitem__`, `__delitem__`, `__iter__`, `__len__`; after that you get `__contains__`, `keys`, `items`, `values`, `get`, `__eq__`, `__ne__`, `pop`, `popitem`, `clear`, `update`, and `setdefault` automatically.

scrapy/utils/item.py
```diff
@@ -0,0 +1,104 @@
...
+            from dataclasses import asdict
+            return asdict(self.item)
+        elif isinstance(self.item, dict):
+            return self.item.copy()
+        elif isinstance(self.item, BaseItem):
+            return dict(self.item)
+
+    def field_names(self):
+        """
+        Returns a generator with the names of the item's fields
```
kmike, line 95: I'd prefer keys method, as key view is more useful than a generator. For example, for dict.keys() `__repr__` shows actual keys. List is also ok. I can't imagine a case where having all item keys in memory (as opposed to making them on-by-one in  a generator) would cause memory issues.

scrapy/utils/item.py
kmike: And one more thing: do you see any issues with allowing to yield ItemAdaptor instances as items? Do you see any use case for that?

scrapy/utils/item.py
elacuesta: Makes sense, I'll make it return a list to keep the return type consistent (IIRC there is no way to create `dict_keys` objects in user code).
Regarding the name, I think it'd be better to keep it as it is now, to make it clear that it returns *all* field names. This method is only used in the `scrapy.exporters.BaseItemExporter` class, where sometimes there is the need to know all the defined fields for a given item. In `BaseItem`-based items, `Ìtem.keys()` and `Item.fields.keys()` can be different if a field is defined but unset:

```python
   ...: class TestItem(Item):
   ...:     name = Field()
   ...:     value = Field()
   ...: i = TestItem(name="asdf")
   ...: print(i.keys())
   ...: print(i.fields.keys())
   ...:
dict_keys(['name'])
dict_keys(['name', 'value'])
```

scrapy/utils/item.py
elacuesta: Implemented the `MutableMapping` interface, that simplified a few things (no need for `asdict` and `__contains__` anymore, for instance) 🥳 
Just as a note, I see no other way to implement `__delitem__` but to use [`delattr`](https://docs.python.org/3/library/functions.html#delattr), which seems to leave `dataclass` objects in an unstable state, but this also happens if we define our own `__init__` method:

```python
>>> from dataclasses import dataclass, fields, asdict
>>> @dataclass(init=False)
... class TestItem:
...     name: str
...     value: int
...     def __init__(self, name):
...         self.name = name
...
>>> i = TestItem(name="asdf")
>>>
>>> [f.name for f in fields(i)]  # all good here
['name', 'value']
>>> i.name
'asdf'
>>> i.value  # expected to raise AttributeError
AttributeError: 'TestItem' object has no attribute 'value'
>>>
>>> i  # __repr__ breaks
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/dataclasses.py", line 357, in wrapper
    result = user_function(self)
  File "<string>", line 2, in __repr__
AttributeError: 'TestItem' object has no attribute 'value'
>>>
>>> asdict(i)  # this breaks too
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/dataclasses.py", line 1044, in asdict
    return _asdict_inner(obj, dict_factory)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/dataclasses.py", line 1051, in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
AttributeError: 'TestItem' object has no attribute 'value'
>>>
```

elacuesta: Regarding allowing `ItemAdapter` objects as items, I _think_ there should be no problems, it should be only a matter of allowing it in `is_item_like`.
If we want to avoid the extra wrapping layer that it would create, we could add some `classmethod` to the class in order to return the object itself if it's already an adapter (I did some experiments with `__new__` but that didn't work). Or perhaps we don't really care about that. What do you think?

Gallaecio: > And one more thing: do you see any issues with allowing to yield ItemAdaptor instances as items? Do you see any use case for that?

I think we should consider preventing this, either by automatically _unwrapping_ items before moving them forward, or by raising an exception.

At the moment, if a user yields an item of a certain type from a spider callback, that user can trust that the item will be the same object the user receives in any custom middleware or pipelines. We lose this if third-party middlewares are allowed to yield these adapter objects and not just use them.

I am not completely against it, but if we want to go this way, it may be better to just convert any item into an Item instance.

scrapy/utils/item.py
kmike: I _think_ you should be getting `.keys()` method for free if you implement MutableMapping, but I haven't actually checked it.

elacuesta: Nice, indeed:
```python
In [1]: from scrapy.utils.item import ItemAdapter                                                                                                                                                                                             

In [2]: ItemAdapter({}).keys()                                                                                                                                                                                                                
Out[2]: KeysView(ItemAdapter for type dict: {})
```
I think the `field_names` function might still be necessary though, since the current exporter implementation needs all fields when exporting empty fields, and `keys` only returns those which are set:
```python
In [10]: from scrapy.item import Item, Field 
    ...: from scrapy.utils.item import ItemAdapter 
    ...: class TestItem(Item): 
    ...:     name = Field() 
    ...:     value = Field() 
    ...: i = TestItem(name="asdf") 
    ...: a = ItemAdapter(i) 
    ...: print(i.keys()) 
    ...: print(i.fields.keys()) 
    ...: print(list(a.keys()))                                                                                                                                                                                                                
dict_keys(['name'])
dict_keys(['name', 'value'])
['name']
```

docs/faq.rst
```diff
@@ -360,6 +360,59 @@ Yes, by setting :setting:`DNS_RESOLVER` to ``scrapy.resolver.CachingHostnameReso
...
+>>> adapter = ItemAdapter(item)
+>>> adapter.item is item
+True
+>>> adapter["name"]
+'foo'
+>>> 'foo'
+'foo'
+>>> adapter["name"] = "bar"
+>>> adapter["price"] = 5
+>>> adapter.item
```
kmike, line 411:
```suggestion
>>> item
```
Just to make it more clear that adapter changes an item in-place.

scrapy/exporters.py
```diff
@@ -56,13 +57,16 @@ def _get_serialized_fields(self, item, default_value=None, include_empty=None):
...
             include_empty = self.export_empty_fields
+
         if self.fields_to_export is None:
-            if include_empty and not isinstance(item, dict):
-                field_iter = item.fields.keys()
+            if include_empty:
+                field_iter = item.field_names()
             else:
-                field_iter = item.keys()
+                field_iter = (x for x in item.field_names() if x in item)
```
kmike, line 69: would item.keys() work here now?

scrapy/item.py
```diff
@@ -15,15 +15,18 @@
...
+    - it is a dataclass object (see the *dataclasses* module in Python 3.6+)
+    For example, when the output of a spider callback is evaluated, only item
+    instances are passed to :ref:`item pipelines <topics-item-pipeline>`.
 
     If you need instances of a custom class to be considered items by Scrapy,
-    you must inherit from either :class:`BaseItem` or :class:`dict`.
+    you must either make said class inherit from :class:`BaseItem` or
+    :class:`dict`, or decorate it as a dataclass object.
 
     Unlike instances of :class:`dict`, instances of :class:`BaseItem` may be
```
kmike, line 31: unlike dicts and dataclasses?

scrapy/utils/item.py
```diff
@@ -0,0 +1,104 @@
+from collections.abc import MutableMapping
+
+from scrapy.item import BaseItem
+
+
+def _is_dataclass_instance(obj):
+    """
+    Returns True if *obj* is a dataclass object, False otherwise.
```
kmike, line 8: a nitpick: as per [pep 257](https://www.python.org/dev/peps/pep-0257/), it should be "Return ...", not "Returns ...":

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

tests/test_utils_item.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertIsInstance(adapter.get_field("name"), Field)
+        self.assertIsInstance(adapter.get_field("value"), Field)
+        self.assertIs(adapter.get_field("name")["serializer"], str)
+        self.assertIs(adapter.get_field("value")["serializer"], int)
+
+    def test_asdict(self):
+        for cls in filter(None, [TestItem, dict, DataClassItem]):
+            item = cls(name="asdf", value=1234)
+            adapter = ItemAdapter(item)
+            self.assertEqual(dict(name="asdf", value=1234), dict(adapter.items()))
```
kmike, line 175: Does dict(adapter) work?

kmike:
@elacuesta one more idea: add ItemAdapter imports to the Scrapy project template, for middlewares / pipelines.

scrapy/utils/item.py
kmike: As for delitem, I think that's fine to raise NonImplementedError - but it seems you already implemented it :)

elacuesta: Alright, now I think we should not allow yielding adapters. It reminds me of the question of allowing to yield item loaders, the implementation should not be super hard but it doesn't add much IMHO and it's probably not necessary.
Regarding `__delitem__`, I didn't think about `NonImplementedError` :thinking: 
Sounds as reasonable as using `delattr` to me, I don't have a strong opinion on the subject.

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -7,6 +7,9 @@
 
 from scrapy import signals
 
+# useful for handling different item types with a single interface
+from scrapy.utils.item import ItemAdapter
+
```
elacuesta, line 12: @kmike Added these lines to both templates, let me know if you had something else in mind

kmike: I was thinking we can have it in the example code below as well- something like

```py
adapter =ItemAdapter(item)
# ... processing code should work with adapter, not with an item;
# item is going to be modified in-place
return item 
```

But unsure now, as it starts to feel heavy.

Gallaecio:
Given the way I’m modifying the documentation, it may make sense to rename `is_item_like` to something like `is_item` or `is_item_instance`. The concept of “item-like” would disappear, in favor of taking about “item types”.

elacuesta:
That works for me :+1: I think I like `is_item` better, short and simple. Should I rename it here or do you plan to do it in the docs PR?

Gallaecio:
I can do it in the documentation PR.

scrapy/utils/item.py
```diff
@@ -0,0 +1,104 @@
...
+    Return True if *obj* is considered a Scrapy *item*, False otherwise.
+
+    An object is considered an *item* if it is:
+    - a scrapy.item.BaseItem or dict instance (or any subclass)
+    - a dataclass object
+    """
+    return isinstance(obj, (BaseItem, dict)) or _is_dataclass_instance(obj)
+
+
+class ItemAdapter(MutableMapping):
```
kmike, line 30: I don't want to derail the PR, but I'm wondering now if we should create a separate package with ItemAdapter. A package which provides  an uniform dict-like access to various sources of data: scrapy Items, dicts, dataclasses, probably attr.s.

Where this idea comes from: scrapy.loader is being separated to a different package, and after this PR scrapy.loader depends on ItemAdapter.

kmike: /cc @ejulio

elacuesta: @kmike That makes sense. I can create a repo under my account and start working on it, we can transfer it later. Does that sound good?

kmike: @elacuesta sure, that'd be perfect!

elacuesta: Working on this at https://github.com/elacuesta/scrapy-itemadapter

elacuesta: @Gallaecio @kmike `itemadapter 0.0.1` is on PyPI, I can start updating this PR. One thing I think we might want to discuss is whether or not we want to import and use directly from the package, or maybe do something similar to what we do with Parsel. An argument in favour of this would be restricting the usage to `Item` objects in Scrapy, since I don't think we should do that in `itemadapter` (because, for the purposes of the library, any `MutableMapping` does work). What do you think?

Gallaecio: My main concern is the handling of `{'fields': '…'}` in non-`Item` objects. I’m not sure it is safe to assume that if the attribute exists we can treat the object as a Scrapy `Item`. Best case scenario: a puzzling `TypeError` exception is raised. Worst case scenario: thinks work in unintended ways.

kmike: @elacuesta I agree with @Gallaecio; I think scrapy.Item should be handled not in the same way as MutableMapping in itemadapter package, and it is better to do a conditional import of scrapy.Item in itemadapter.

So in a way, it is a circular dependency (not necessarily circular import). itemadapter can work without Scrapy, but supports Scrapy Items, just like it supports dataclasses or attr.s objects.

For me it looks like a separate wrapper is not required here; we can use it directly.

elacuesta: I think we would lose generality in `itemadapter` if we do an explicit check for Scrapy items. The way I see it is:
* We check for Scrapy items, and decide to remove support for meta in general for mutable mappings. We're disallowing something that we can support and could be useful for users
* We check for Scrapy items, and decide to keep support for meta in general for mutable mappings. The code for getting the fields is the same for items as it is for mutable mappings, it ends up being the same as not checking in the first place

I think one alternative is to restrict things in the Scrapy side, i.e.:

```python
# scrapy/utils/item.py

from scrapy.item import Item

from itemadapter.adapter import ItemAdapter as BaseItemAdapter
from itemadapter.utils import is_attrs_instance, is_dataclass_instance


def is_item(obj):
    return isinstance(obj, Item) or is_dataclass_instance(obj) or is_attrs_instance(obj)


class ItemAdapter(BaseItemAdapter):
    def __init__(self, item):
        if not is_item(item):
            raise TypeError("Expected a valid item, got %r instead: %s" % (type(item), item))
        super().__init__(item)
```

Gallaecio: By “support for meta” do you mean the 2 methods to get field metadata? Wouldn’t it make sense _not_ to support that in `MutableMapping` given `MutableMapping` does not support defining metadata?

Otherwise we are saying that, in mappings used with this library, the `fields` key is reserved for field metadata. If field metadata support is desired, I think something other than a simple `dict`, `MutableMapping` or `Mapping` should be used.

elacuesta: Not the key but the _attribute_, i.e. `obj.fields` not `obj["fields"]`. My reasoning was "if we support arbitrary dicts, why not support arbitrary dicts which also happen to define a `fields` attribute?"
Anyway, this probably will not be an issue in practice, as you say, I imagine users using other types if they need metadata, I just thought it would be nice to allow and not difficult to maintain. I can remove that in itemadapter, it's already released but it's not like many people are using it.

Gallaecio: > Not the key but the attribute, i.e. obj.fields not obj["fields"]

I did not realize this :facepalm: . In that case I think it is OK to support it. If we do, though, the documentation should definitely cover this, and explain how to implement such an attribute.

kmike: Hey! Currently in Scrapy it works like that:

* dict is an item, meta is not supported
* Item is an item, meta is supported

After this PR, before itemadapter package:
* dataclass instance is an item, meta is not supported (for now; we could support https://docs.python.org/3.9/library/typing.html#typing.Annotated in future)

In the itemadapter, non-controversial:
* attr.s instance is an item, meta is supported (though is the support implemented?)

The thing we're discussing, a controversial change in itemadapter:
* arbitrary MutableMapping is an item
* for arbitrary MutableMapping meta is supported if it has "fields" attribute.

I think the last part may require a separate discussion - do we allow any arbitrary MutableMapping as an item, or not. If we allow arbitrary MutableMapping, how to annotate fields. This change is much larger than adding dataclass support. I recall initially we decided to be very explicit, and so Scrapy is checking just for `dict` now, not for any MutableMapping - it'd be easy to add support for arbitrary MutableMapping in future, but it'd be hard to remove it, if we find some issue with that. It may have a merit, and use cases, and could be non-problematic, but I think it deserves a separate discussion.

What do you think about focusing on non-controversial, low risk changes first, to get everything merged? I.e. rollback / move to PR controversial stiff. And then we can discuss MutableMapping thing separately.

kmike: The difference is that for dicts, scrapy.Item, dataclasses and attr.s instances we know what we're working with; for MutableMapping we don't, as all kinds af object in Python ecosystem may be implementing it. That's why I see it as a higher risk change.

elacuesta: Sounds fine to me, I removed meta for MutableMappings already, I'll remove MutableMappings altogether now, it'd be easy to add it back if we decide to.

elacuesta: Done in [v0.0.5](https://github.com/elacuesta/itemadapter/releases/tag/v0.0.5).
One thing I wonder is: will there be any issues when running `itemadapter`'s tests (which install Scrapy to check for `scrapy.item.Item`) if Scrapy setup requires `ìtemadapter`?

docs/topics/items.rst
```diff
@@ -8,29 +8,116 @@ Items
...
+
+.. _custom-items:
+
+Custom items
+------------
+
+Subclass :class:`BaseItem` to define additional item types not based on any of
+the above:
+
+.. autoclass:: scrapy.item.BaseItem
```
elacuesta, line 108: @Gallaecio @kmike I wonder if we should discourage inheriting from `BaseItem`, since it only inherits from `object_ref` and it doesn't provide any additional functionality?

kmike: @elacuesta as we discussed today: this makes sense, but it'd be best if we tackle  it separately, and keep BaseItem checks for itemadapter.

elacuesta: I deprecated `BaseItem` in https://github.com/scrapy/scrapy/pull/3881/commits/23c1739965119df3d91f5e288520c8c21e86312e, do you think it would be too much out of the scope of this PR? Opening a new PR would be fine too, I just think it'd be good to have it deprecated in 2.2 instead of waiting a new full version bump.

elacuesta: Moved to a separate PR: #4534

elacuesta:
@Gallaecio @kmike @wRAR With #4534 merged and [`itemadapter`](https://github.com/scrapy/itemadapter) available on [PyPI](https://pypi.python.org/pypi/itemadapter), this PR should be ready to be reviewed and merged. Let me know if you have any comments.

docs/faq.rst
```diff
@@ -360,6 +360,15 @@ Yes, by setting :setting:`DNS_RESOLVER` to ``scrapy.resolver.CachingHostnameReso
...
 
+.. _faq-dataclass-items:
+
+Can I use dataclasses as items?
+-------------------------------
+
+Support for :class:`dataclasses.dataclass` objects as items was added in version 2.1.
+This works natively in Python 3.7+, or using the `dataclasses backport`_ in Python 3.6.
+See :ref:`topics-items` for more information.
+
```
Gallaecio, line 371: Do we need this entry, now that the documentation about items offers a section about it?

docs/topics/item-pipeline.rst
```diff
@@ -79,16 +83,18 @@ Let's take a look at the following hypothetical pipeline that adjusts the
 (``price_excludes_vat`` attribute), and drops those items which don't
 contain a price::
 
+    from itemadapter import ItemAdapter
     from scrapy.exceptions import DropItem
-
+
```
Gallaecio, line 88:
```suggestion

```

docs/topics/leaks.rst
```diff
@@ -61,8 +61,8 @@ Debugging memory leaks with ``trackref``
 ========================================
 
 :mod:`trackref` is a module provided by Scrapy to debug the most common cases of
-memory leaks. It basically tracks the references to all live Requests,
-Responses, Item and Selector objects.
+memory leaks. It basically tracks the references to all live Request,
+Response, BaseItem, Spider and Selector objects.
```
Gallaecio, line 65: Given #4534 I guess we need to review the new usages of `BaseItem`, and in cases like these revert them to `Item`.

docs/topics/media-pipeline.rst
```diff
@@ -243,26 +243,21 @@ Usage example
...
-        image_urls = scrapy.Field()
-        images = scrapy.Field()
+When using :ref:`item types <item-types>` where fields are defined beforehand,
+you must define both the URLs field and the results field. For example, when
+using the images pipeline, items must define both the ``image_urls`` and the
+``images`` field.
 
 If you want to use another field name for the URLs key or for the results key,
-it is also possible to override it.
+it is also possible to override it:
```
Gallaecio, line 260: :lipstick: No strong opinion (and there’s a high chance I’m the author of this line change), but I think it may make sense, given this colon, to convert the two items below into unordered list items.

scrapy/exporters.py
```diff
@@ -4,16 +4,18 @@
...
-import warnings
 import pickle
+import pprint
+import warnings
 from xml.sax.saxutils import XMLGenerator
 
-from scrapy.utils.serialize import ScrapyJSONEncoder
-from scrapy.utils.python import to_bytes, to_unicode, is_listlike
-from scrapy.item import _BaseItem
+from itemadapter import ItemAdapter, is_item
```
Gallaecio, line 13: :lipstick: 

```suggestion
from itemadapter import is_item, ItemAdapter
```

scrapy/exporters.py
```diff
@@ -314,22 +323,22 @@ def serialize_field(self, field, name, value):
...
         encode_func = to_bytes if self.binary else to_unicode
         if isinstance(value, (str, bytes)):
             return encode_func(value, encoding=self.encoding)
         return value
 
-    def _serialize_dict(self, value):
+    def _serialize_item(self, value):
         for key, val in value.items():
             key = to_bytes(key) if self.binary else key
             yield key, self._serialize_value(val)
```
Gallaecio, line 338: I think we need to use `ItemAdapter` here.

scrapy/loader/__init__.py
```diff
@@ -127,13 +129,12 @@ def get_value(self, value, *processors, **kw):
...
-        item = self.item
+        adapter = ItemAdapter(self.item)
         for field_name in tuple(self._values):
             value = self.get_output_value(field_name)
             if value is not None:
-                item[field_name] = value
-
-        return item
+                adapter[field_name] = value
+        return adapter.item
```
Gallaecio, line 137:
```suggestion
        return self.item
```

scrapy/templates/project/module/middlewares.py.tmpl
```diff
@@ -5,6 +5,9 @@
 
 from scrapy import signals
 
+# useful for handling different item types with a single interface
+from itemadapter import ItemAdapter
```
Gallaecio, line 9: For spider middlewares, `is_adapter` can be handy as well. Also, we need to update the comments below to replace `dict or Item objects` with `` :ref:`item objects <topics-items>` ``.

kmike:
@elacuesta itemadapter code could be worth reviewing as well (e.g. there are some things in README which we may need to change). How should we do it, what's your opinion?

scrapy/loader/__init__.py
elacuesta: It's the same object, and I'd argue that `return adapter.item` is a little bit more clear in the sense that in order to understand `return self.item` you need to know that `ItemAdapter` modifies the item in-place, which you might not know a priori.

Gallaecio: I think that it’s precisely because it’s important for people to understand that `ItemAdapter` edits the item in place that we should avoid `adapter.item` unless necessary.

That said, this is internal code and not documentation, so I guess that aspect is not that relevant.

elacuesta:
@kmike Of course, please do review `itemadapter` as well. If necessary, I think it'd be ok to create issues there and link them here. This PR adds `itemadapter>=0.0.6` to `install_requires`, if you find any issues with the README alone I think it's fine to just update it, if code changes are necessary too I can release a new version and update `install_requires` accordingly.

docs/topics/media-pipeline.rst
elacuesta: In that case, let's revert it. The PR is long enough as it is, and I don't think this specific change is necessary for the overall PR to be understood.

docs/topics/item-pipeline.rst
```diff
@@ -112,7 +119,7 @@ format::
            self.file.close()
 
        def process_item(self, item, spider):
-           line = json.dumps(dict(item)) + "\n"
+           line = json.dumps(dict(ItemAdapter(item).items())) + "\n"
```
kmike, line 122: is .items() required here?

docs/topics/item-pipeline.rst
```diff
@@ -155,7 +163,7 @@ method and how to clean up the resources properly.::
             self.client.close()
 
         def process_item(self, item, spider):
-            self.db[self.collection_name].insert_one(dict(item))
+            self.db[self.collection_name].insert_one(dict(ItemAdapter(item).items()))
```
kmike, line 166: is .items() required here?

docs/topics/items.rst
```diff
@@ -8,29 +8,124 @@ Items
...
+.. note:: Field names and types are not enforced at run time.
+
+They work natively in Python 3.7 or later, or using the `dataclasses
+backport`_ in Python 3.6.
+
+.. _dataclasses backport: https://pypi.org/project/dataclasses/
+
+.. _attrs-items:
+
+Attr.s objects
```
kmike, line 101:
```suggestion
attr.s objects
```

docs/topics/item-pipeline.rst
elacuesta: [It is not](https://github.com/scrapy/itemadapter/blob/v0.0.6/tests/test_adapter.py#L149-L153), thanks for noticing.

docs/topics/items.rst
```diff
@@ -8,29 +8,124 @@ Items
...
+
+Scrapy supports :ref:`multiple types of items <item-types>`. When you create an
+item, you may use whichever type of item you want. When you write code that
+receives an item, your code should :ref:`work for any item type
+<supporting-item-types>`.
+
+.. _item-types:
+
+Item Types
+==========
```
kmike, line 22: I think it'd be nice to show how to define each item type. Currently there is only an example for Item, and it is in a separate section.

docs/topics/items.rst
```diff
@@ -8,29 +8,124 @@ Items
...
+Attr.s objects
+--------------
+
+.. versionadded:: 2.2
+
+:func:`attr.s` allows defining item classes with field names,
+so that :ref:`item exporters <topics-exporters>` can export all fields by
+default even if the first scraped object does not have values for all of them.
+
+`attr.s` also allow defining the type and default value of each defined field.
```
kmike, line 110: Is field metadata supported for attr.s and dataclasses?

docs/topics/items.rst
```diff
@@ -8,29 +8,124 @@ Items
...
+
+.. note:: Field names and types are not enforced at run time.
```
kmike, line 112: I _think_ you can have it by defining a class with `@attr.s(slots=True)`

docs/topics/items.rst
elacuesta: Yes it is! https://github.com/scrapy/itemadapter#metadata-support

docs/topics/items.rst
elacuesta: I don't really know, but I can test it and see what happens. Let's summon @Gallaecio, who authored this particular change, in case he can provide some insight.

docs/topics/media-pipeline.rst
```diff
@@ -243,23 +243,18 @@ Usage example
...
+respectively), the pipeline will put the results under respective field
+(``files`` or ``images``).
 
-        # ... other item fields ...
-        image_urls = scrapy.Field()
-        images = scrapy.Field()
+When using :ref:`item types <item-types>` where fields are defined beforehand,
+you must define both the URLs field and the results field. For example, when
+using the images pipeline, items must define both the ``image_urls`` and the
+``images`` field.
```
kmike, line 257: the example deleted above could work here

scrapy/commands/parse.py
```diff
@@ -117,7 +117,7 @@ def run_callback(self, response, callback, cb_kwargs=None):
         items, requests = [], []
 
         for x in iterate_spider_output(callback(response, **cb_kwargs)):
-            if isinstance(x, (_BaseItem, dict)):
+            if is_item(x):
```
kmike, line 120: should print_items be modified as well?

scrapy/contracts/default.py
```diff
@@ -103,8 +103,8 @@ class ScrapesContract(Contract):
 
     def post_process(self, output):
         for x in output:
-            if isinstance(x, (_BaseItem, dict)):
+            if is_item(x):
                 missing = [arg for arg in self.args if arg not in x]
```
kmike, line 107: should you use ItemAdapter here?

scrapy/exporters.py
```diff
@@ -71,7 +76,10 @@ def _get_serialized_fields(self, item, default_value=None, include_empty=None):
 
         for field_name in field_iter:
             if field_name in item:
-                field = {} if isinstance(item, dict) else item.fields[field_name]
+                try:
+                    field = item.get_field_meta(field_name)
+                except TypeError:
+                    field = {}
```
kmike, line 82: Are there cases when `item.get_field_meta(field_name)` for an item type which doesn't support meta should return anything other than `{}`?

scrapy/utils/serialize.py
```diff
@@ -26,8 +26,8 @@ def default(self, o):
             return str(o)
         elif isinstance(o, defer.Deferred):
             return str(o)
-        elif isinstance(o, _BaseItem):
-            return dict(o)
+        elif is_item(o):
+            return dict(ItemAdapter(o).items())
```
kmike, line 30: is .items() needed?

tests/test_loader.py
```diff
@@ -466,15 +475,15 @@ def test_keep_single_value(self):
         il = ItemLoader(item=input_item)
         loaded_item = il.load_item()
         self.assertIsInstance(loaded_item, self.item_class)
-        self.assertEqual(dict(loaded_item), {'name': ['foo']})
+        self.assertEqual(dict(ItemAdapter(loaded_item).items()), {'name': ['foo']})
```
kmike, line 478: is .items() needed? There are a few more cases like that.

tests/test_pipeline_files.py
```diff
@@ -119,43 +137,73 @@ def test_file_expired(self):
...
+    # overridden fields
+    custom_file_urls = Field()
+    custom_files = Field()
+
+
+class FilesPipelineTestCaseFieldsItem(FilesPipelineTestCaseFieldsMixin, unittest.TestCase):
+    item_class = FilesPipelineTestItem
+
+
+@skipIf(not make_dataclass, "dataclasses module is not available")
```
kmike, line 189: as we advertise attr.s support, should we check that it works, i.e. have tests for attr.s as well?

docs/topics/items.rst
kmike: I think it counts as a feature of a particular item type; it'd be nice to mention it.

scrapy/exporters.py
elacuesta: I'm not sure I understand the question, but I see how using `TypeError` could be bad. Adding a `default` parameter sounds coherent with `dict.get`. I'll make the changes in itemadapter.

kmike:
@elacuesta itemloader issues created; I've checked the package itself, checked README briefly, and a few other files. Haven't finished reviewing tests.

scrapy/exporters.py
elacuesta: See https://github.com/scrapy/itemadapter/pull/19

docs/topics/items.rst
elacuesta: For the record, it doesn't seem like `slots=True` enforces types. Also, I couldn't find relevant information at https://www.attrs.org/en/stable/examples.html#types.

```python
In [1]: import attr 
   ...: @attr.s(slots=True) 
   ...: class AttrsItem: 
   ...:     name = attr.ib(type=str, metadata={"serializer": str}) 
   ...:     value = attr.ib(type=int, metadata={"serializer": int}) 
   ...:

In [2]: AttrsItem(name=999, value="not-an-int")
Out[2]: AttrsItem(name=999, value='not-an-int')
```

docs/topics/items.rst
elacuesta: Done in 2b2791a7343b9d849c3266be9af2311acade491c

docs/topics/items.rst
elacuesta: Done in 2b2791a7343b9d849c3266be9af2311acade491c

docs/topics/items.rst
```diff
@@ -8,29 +8,161 @@ Items
...
+
+* define custom `field metadata`_, which can be used to
+  :ref:`customize serialization <topics-exporters-field-serialization>`.
+
+.. note:: Field names and types are not enforced at run time.
+
+They work natively in Python 3.7 or later, or using the `dataclasses
+backport`_ in Python 3.6.
+
+.. _field metadata: https://docs.python.org/3/library/dataclasses.html#dataclasses.field
```
Gallaecio, line 109: What about this to avoid non-InterSphinx links to the Python documentation?

```suggestion
* define custom field metadata through :func:`dataclasses.field`, which can be used to
  :ref:`customize serialization <topics-exporters-field-serialization>`.

.. note:: Field names and types are not enforced at run time.

They work natively in Python 3.7 or later, or using the `dataclasses
backport`_ in Python 3.6.

```

docs/topics/items.rst
```diff
@@ -8,29 +8,161 @@ Items
...
+.. _dataclasses backport: https://pypi.org/project/dataclasses/
+
+Example::
+
+    from dataclasses import dataclass
+
+    @dataclass
+    class CustomItem:
+        one_field: str
+        another_field: int
```
Gallaecio, line 119: Maybe we could include a `field` usage example here?

docs/topics/items.rst
Gallaecio: Well, at least they enforce the names, as `Item` does.

What about removing the note from there and refactoring the first list item above as follows?

```
* Define the type and default value of each defined field.

  Use :term:`slotted classes` so that an :exc:`AttributeError` exception
  is raised when using undefined field names (i.e. prevent typos 
  going unnoticed).

  .. note:: Field types are not enforced at run time.

* Define custom field `metadata`_, which can be used to
  :ref:`customize serialization <topics-exporters-field-serialization>`.
```

Notes:
- This requires adding `attr.s` to the InterSphinx mapping.
- I’m not sure if `` :term:`slotted classes` `` will work right away. It may require to prefix the term with the `attr.s` InterSphinx alias, e.g. `` :term:`attrs:slotted classes` ``
- Since we are adding `attr.s` to the InterSphinx mapping, we should use it for the metadata link as well, maybe also the main link (point to https://www.attrs.org/en/stable/index.html, `` :doc:`attrs:index` ``).

tests/test_pipeline_files.py
Gallaecio: I’m +0 here.

As long as the new library verifies that both data types work as expected when using `ItemAdapter`, and we make sure we only access items through `ItemAdapter`, I think we should be OK.

On the other hand, I cannot think of a way to make sure that we are using `ItemAdapter` other than making sure that things work with dataclasses, `attr.s`, etc.

docs/topics/items.rst
elacuesta: I think we should only include a minimal example. If we included a `field` example, we should include more complex examples for `attrs` as well, and it's out of the scope of Scrapy to explain how to use `dataclasses` and `attrs`. Also, these dataclass fields are only used within Scrapy for metadata, and the docs about metadata point to the `dataclasses.field` function.

docs/topics/items.rst
elacuesta: I changed the note to "Field types are not enforced at run time". I actually think this note might not be necessary in general, because it's not a Scrapy characteristic. I left it for `dataclasses` but removed the `attrs` one, which doesn't even mention types.
I don't think we need to mention slotted classes, that's not specific to Scrapy but to `attrs`. The links to the upstream docs should be enough IMHO.

scrapy/exporters.py
kmike: `item.get_field_meta` currently raises an exception if an item doesn't provide a way to set meta. Now I think that there should be a way for it not to raise an exception. What's the use case for this exception?

If there is no use case, we should just return `{}`. There is no use case for an exception in Scrapy itself, as in all Scrapy usages exception handling won't be needed if a `{}` default is returned when meta is absent.

If there is a possibility that there is a use case, but it is not clear which use case is that, it'd be good to keep an option to have this exception, but make the lookup return `{}` by default, to match the common usage better. 

If there is a use case, then we should consider making exception a default behavior.

For me returning empty dict as a field meta for dicts (or other future possibile items which don't provide meta) feels ok. Meta is optional in dataclasses and attrs, so it is normal behavior that the meta is not populated. As it is optional, that's not a problem that for dict meta is empty - it could be empty for attrs or dataclasses as well. The difference is that dict doesn't allow to set meta, but that's user choice. If they realize they need to set meta, they switch to scrapy Items or attrs or dataclasses. It is actually similar for attrs and dataclasses: you can't set meta if you use type annotation-based definitions; if user realizes they need meta, they switch to e.g. to an explicit attr.ib(...).

elacuesta: Updated in a3672d5e81aeec76b6c5718dd9f9f61f43edf1f3, https://github.com/scrapy/itemadapter/commit/3594f8f52767e499c86182e25509dce893cbf6ae

tests/test_pipeline_files.py
elacuesta: Tests for `attrs` items: b4e535721a37035db596ca8b56b3cbd4779d4568

docs/topics/media-pipeline.rst
elacuesta: Updated in 529a15502e87a75feedc6429cc5ee00ef9ef9745

scrapy/loader/__init__.py
```diff
@@ -174,11 +175,12 @@ def _process_input_value(self, field_name, value):
                                     value, type(e).__name__, str(e)))
 
     def _get_item_field_attr(self, field_name, key, default=None):
-        if isinstance(self.item, Item):
-            value = self.item.fields[field_name].get(key, default)
+        try:
+            field_meta = ItemAdapter(self.item).get_field_meta(field_name)
+        except TypeError:
```
kmike, line 180: this exception can't be reaised now

elacuesta: Oh nice catch, I only remembered the one in `exporters.py`

docs/topics/exporters.rst
```diff
@@ -167,9 +169,10 @@ BaseItemExporter
       value unchanged except for ``unicode`` values which are encoded to
       ``str`` using the encoding declared in the :attr:`encoding` attribute.
 
-      :param field: the field being serialized. If a raw dict is being
-          exported (not :class:`~.Item`) *field* value is an empty dict.
-      :type field: :class:`~scrapy.item.Field` object or an empty dict
+      :param field: the field being serialized. If the source :ref:`item object
+          <item-types>` does not define field metadata, *field* is an empty
+          :class:`dict`.
+      :type field: :class:`~scrapy.item.Field` object or an empty :class:`dict`
```
kmike, line 175:
```suggestion
      :type field: :class:`~scrapy.item.Field` object or a :class:`dict` instance
```

kmike: the dict doesn't have to be empty, as we support meta for attr.s and dataclasses

docs/topics/item-pipeline.rst
```diff
@@ -112,7 +119,7 @@ format::
            self.file.close()
 
        def process_item(self, item, spider):
-           line = json.dumps(dict(item)) + "\n"
+           line = json.dumps(dict(ItemAdapter(item))) + "\n"
```
kmike, line 122: I know I pushed for this pattern, but it may be not future proof - see https://github.com/scrapy/itemadapter/issues/27.

kmike:
That's monumental work @elacuesta. 

Last comments:

* +0 to add meta examples to docs with field definitions. It would also be good to add examples of slotted classes for attr.s, and elaborate a bit more on pro/cons of various item types (e.g. dataclasses is stdlib, attr.s provide validation and slots, etc.). But we can do this all later; it can be better to discuss it separately, and not only because it delays this PR. I think the current docs are good, and these proposed doc changes are not a blocker for this PR.
* https://github.com/scrapy/itemadapter/issues/27 looks important; I think it would be good to address it before merging, as it changes ItemAdapter interface.
* the PR looks in awesome state otherwise.

So, +1 to merge after we get a plan for adapter => dict conversion with nested dataclass/attr.s items.

Gallaecio:
There’s also https://github.com/scrapy/scrapy/pull/3881#discussion_r428598241

I see there have been changes in those lines, but I see no mention of “slotted” in the changes.

elacuesta:
Merging then, thanks to everyone involved :heart: 

To be clear, we discussed the slotted thing recently, and agreed that there is no need to mention it. Field names are enforced, in the sense that (un-slotted) `attrs` classes allow to define random attributes (as any Python class) but they are not considered fields:
```python
In [1]: import attr

In [2]: @attr.s 
   ...: class AttrsItem: 
   ...:     foo = attr.ib() 
   ...:

In [3]: i = AttrsItem(foo="bar")

In [4]: i.some_attribute = 1

In [5]: attr.asdict(i)
Out[5]: {'foo': 'bar'}
```

kmike:
> To be clear, we discussed the slots thing recently, and agreed that there is no need to mention it. 

@elacuesta I still think slots need to me mentioned, because that's almost the point of having scrapy.Item - to prevent mistakes. With a non-slotted attrs class, a typo in a field name could be even worse than a typo in a dictionary - the data will be silently dropped. scrapy.Item and attr.s with slots prevent this kind of errors; we should be promoting it as a best practice. 

But this can be done in another PR, that's not a blocker here.

PR #3882, opened 2019-07-16
add instructions about how to define output file when running scrapy …

MagdalenaDeschner:
…from script instead of cmd

The following stackoverflow questions show, that it is unclear how to define the feed format and name within a script. Equivalent to "-o" flag when starting crawler process within cmd.
https://stackoverflow.com/questions/43468275/running-scrapy-from-a-script-with-file-output/43468852
https://stackoverflow.com/questions/23574636/scrapy-from-script-output-in-json

### DIFF
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -35,11 +35,16 @@ Here's an example showing how to run a single spider with it.
         ...
 
     process = CrawlerProcess({
-        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)'
+        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)',
+        'FEED_FORMAT':'json',
+        'FEED_URI':'items.json'
     })
 
     process.crawl(MySpider)
     process.start() # the script will block here until the crawling is finished
+
+Define settings within dictionary in
+CrawlerProcess. FEED_FORMAT and FEED_URI are the equivalent to "-o items.json" when using the scrapy crawl shell command.
 
 Make sure to check :class:`~scrapy.crawler.CrawlerProcess` documentation to get
 acquainted with its usage details.
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -34,8 +34,7 @@ Here's an example showing how to run a single spider with it.
         # Your spider definition
         ...
 
-    process = CrawlerProcess({
-        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)',
+    process = CrawlerProcess(settings={
         'FEED_FORMAT':'json',
         'FEED_URI':'items.json'
     })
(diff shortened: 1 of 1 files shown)


docs/topics/practices.rst
```diff
@@ -35,12 +35,17 @@ Here's an example showing how to run a single spider with it.
         ...
 
     process = CrawlerProcess({
-        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)'
+        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)',
```
Gallaecio: What about removing `USER_AGENT` as part of these changes? I believe the point of `USER_AGENT` was make it obvious that the first parameter of `CrawlerProcess` is a dictionary of settings, which the new settings already accomplish.

docs/topics/practices.rst
```diff
@@ -35,12 +35,17 @@ Here's an example showing how to run a single spider with it.
...
+        'FEED_FORMAT':'json',
+        'FEED_URI':'items.json'
     })
 
     process.crawl(MySpider)
     process.start() # the script will block here until the crawling is finished
 
+Define settings within dictionary in
+CrawlerProcess. FEED_FORMAT and FEED_URI are the equivalent to "-o items.json" when using the scrapy crawl shell command.
```
Gallaecio: > Define settings within dictionary in CrawlerProcess.

I think the example makes this clear, and the paragraph below does encourage reading the documentation of `CrawlerProcess`.

> FEED_FORMAT and FEED_URI are the equivalent to "-o items.json" when using the scrapy crawl shell command.

This information here feels out of the blue to me. If we really think that users will continue to have doubts despite the example featuring `FEED_FORMAT` and `FEED_URI`, then maybe we should add an entry to the FAQ instead, and really address the specific issue of “How to I use -o with CrawlProcess?”.

MagdalenaDeschner:
@Gallaecio I totally agree with you, telling that you can define any setting within a dictionary as a parameter of CrawlerProcess should be enough. There shouldn't be an extra emphasis on any specific setting. Shall I make a new pull request or can you mark it as "needs work"?

docs/topics/practices.rst
MagdalenaDeschner: maybe we can add CrawlerProcess(settings={}) so that it's totally clear.

docs/topics/practices.rst
MagdalenaDeschner: I totally agree with you, telling that you can define any setting within a dictionary as a parameter of CrawlerProcess should be enough. There shouldn't be an extra emphasis on any specific setting. Shall I make a new pull request or can you mark it as "needs work"?

Gallaecio:
> Shall I make a new pull request or can you mark it as "needs work"?

Please use the same pull request.

You can edit the title and prefix it with `[WIP]` to indicate that is currently _work in progress_, so that people don’t bother reviewing it until you remote the prefix from the title again.

docs/topics/practices.rst
```diff
@@ -34,15 +34,16 @@ Here's an example showing how to run a single spider with it.
         # Your spider definition
         ...
 
-    process = CrawlerProcess({
-        'USER_AGENT': 'Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)'
+    process = CrawlerProcess(settings={
+        'FEED_FORMAT':'json',
+        'FEED_URI':'items.json'
```
kmike: a nitpick: could you please add whitespace after `:`, as per pep8?

MagdalenaDeschner: sure

kmike:
Thanks @MagdalenaDeschner, and thanks @Gallaecio for a review!

PR #3885, opened 2019-07-18
Test Python 2.7-3.6 also on xenial.

wRAR:


kmike:
What's the reason for running tests there, what kind of issues do you expect to catch? We're trying not to over-do environments on CI, e.g. this increases CI run time from ~15 minutes to ~25 minutes.

kmike:
Related: https://github.com/scrapy/scrapy/pull/3892

wRAR:
@kmike I was thinking about different system OpenSSL libs, but I already saw that it increases the test time, and if we drop trusty this becomes non-issue.

PR #3886, opened 2019-07-18
[MRG+1] Change DOWNLOAD_MAXSIZE logger level from Error to Warning

### LINKED ISSUE #3874: DOWNLOAD_MAXSIZE logger level shouldn't be Error
When triggered, the `DOWNLOAD_MAXSIZE` setting raises an ERROR level logging event. I don't think it should be at this level. Error is "something went wrong"; I've explicitly told Scrapy I don't want files of this size, so it's not an error to not download them. WARN or even just INFO makes more sense.

I'm logging Error levels and above to stderr so I can keep track of them easily and it's being flooded by these. In fact, apart from a couple of genuine errors, of the ~100,000 requests my spider has made so far, these are the only Errors that are showing up.

WinterComes:
This PR based on #3874

Raising error on reaching download maxsize is removed. 
Logger level is replaced to warning.

Closes #3874

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -379,13 +379,14 @@ class ScrapyAgent(object):
         fail_on_dataloss = request.meta.get('download_fail_on_dataloss', self._fail_on_dataloss)
 
         if maxsize and expected_size > maxsize:
-            error_msg = ("Cancelling download of %(url)s: expected response "
-                         "size (%(size)s) larger than download max size (%(maxsize)s).")
-            error_args = {'url': request.url, 'size': expected_size, 'maxsize': maxsize}
+            warning_msg = ("Expected response size (%(size)s) larger than "
+                           "download max size (%(maxsize)s) in request %(request)s.")
+            warning_args = {'request': request, 'size': expected_size, 'maxsize': maxsize}
 
-            logger.error(error_msg, error_args)
+            logger.warning(warning_msg, warning_args)
+
             txresponse._transport._producer.loseConnection()
-            raise defer.CancelledError(error_msg % error_args)
+            raise defer.CancelledError(warning_msg % warning_args)
 
         if warnsize and expected_size > warnsize:
             logger.warning("Expected response size (%(size)s) larger than "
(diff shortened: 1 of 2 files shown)


scrapy/core/downloader/handlers/http11.py
```diff
@@ -366,13 +366,13 @@ def _cb_bodyready(self, txresponse, request):
...
-            error_args = {'url': request.url, 'size': expected_size, 'maxsize': maxsize}
+            warning_msg = ("Expected response size (%(size)s) larger than "
+                           "download max size (%(maxsize)s) in request %(request)s.")
+            warning_args = {'request': request, 'size': expected_size, 'maxsize': maxsize}
+
+            logger.warning(warning_msg, warning_args)
 
-            logger.error(error_msg, error_args)
             txresponse._transport._producer.loseConnection()
-            raise defer.CancelledError(error_msg % error_args)
```
Gallaecio: I’m not sure removing this line was part of the #3874, where the reported only complaints about the logging level, not that the request is sent to the errback, which is probably the expected behavior.

WinterComes: Yeah, you are right, I returned it back, thanks.

tests/test_downloader_handlers.py
```diff
@@ -548,8 +548,8 @@ def test_download_with_content_length(self):
         # http://localhost:8998/partial set Content-Length to 1024, use download_maxsize= 1000 to avoid
         # download it
         yield crawler.crawl(seed=Request(url=self.mockserver.url('/partial'), meta={'download_maxsize': 1000}))
-        failure = crawler.spider.meta['failure']
-        self.assertIsInstance(failure.value, defer.CancelledError)
+        reason = crawler.spider.meta['close_reason']
+        self.assertTrue(reason, 'finished')
```
Gallaecio: Shouldn’t tests be failing now that you’ve added `raise defer.CancelledError` back?

WinterComes: Oh, It's ```finished``` in both ways :) I'll return check on ```CancelledError```
thanks

sprzedwojski:
May I suggest 2 things before this is merged?
1) Adding "Closes #3874" or similar to automatically close the related issue (see [here](https://help.github.com/en/articles/closing-issues-using-keywords))
2) Squashing all commits to a single commit

WinterComes:
thanks @sprzedwojski

scrapy/core/downloader/handlers/http11.py
```diff
@@ -379,13 +379,14 @@ def _cb_bodyready(self, txresponse, request):
         fail_on_dataloss = request.meta.get('download_fail_on_dataloss', self._fail_on_dataloss)
 
         if maxsize and expected_size > maxsize:
-            error_msg = ("Cancelling download of %(url)s: expected response "
-                         "size (%(size)s) larger than download max size (%(maxsize)s).")
-            error_args = {'url': request.url, 'size': expected_size, 'maxsize': maxsize}
+            warning_msg = ("Expected response size (%(size)s) larger than "
```
kmike: I think previous error message made more sense, because now we're getting two identical messages for warnsize and maxsize - there is no indication that response is dropped here.

PR #3887, opened 2019-07-19
Fix configparser import for python2

marsam:
The latest release breaks scrapy on python2
Closes https://github.com/scrapy/scrapy/issues/3889

### DIFF
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -1,7 +1,7 @@
 import os
 import sys
 import numbers
-import configparser
+from six.moves import configparser
 from operator import itemgetter
 
 import six


kmike:
wow, I wonder why aren't tests detecting this - Python 2.7 tests pass on CI

elacuesta:
Well, this is on me, sorry :hand: #3877
In my defense, Travis tests use Python 2.7.14, which definitely has the lowercase version available.

Locally I have 2.7.15+ and they are different modules AFAICT
```python
Python 2.7.15+ (default, Nov 27 2018, 23:36:35) 
[GCC 7.3.0] on linux2
>>> import configparser
>>> import ConfigParser
>>> configparser
<module 'configparser' from '/home/eugenio/.local/lib/python2.7/site-packages/configparser.pyc'>
>>> ConfigParser
<module 'ConfigParser' from '/usr/lib/python2.7/ConfigParser.pyc'>
>>>
```

I couldn't find anything in the [2.7.14 changelog](https://raw.githubusercontent.com/python/cpython/84471935ed2f62b8c5758fd544c7d37076fe0fa5/Misc/NEWS) (which includes everything until 2.7.14) that indicates an early (before 3.0) renaming. However, I did found https://pypi.org/project/configparser/, which is indeed what my local 2.7.15 runtime is using and I suspect is also what Travis uses.

```
$ head /home/eugenio/.local/lib/python2.7/site-packages/configparser.py
#!/usr/bin/env python
# -*- coding: utf-8 -*-

"""Convenience module importing everything from backports.configparser."""
```

kmike:
In https://github.com/scrapy/scrapy/pull/3877/files#diff-ab96eed25d6502f751749c9b001f7393 SafeConfigParser call was changed to ConfigParser, which seems to be "slightly backwards incompatible"; this PR fixes the problem with Python 2, but has the same issue. I wonder if we should revert the change in this file instead.

Initially I was thinking it may be also related to case sensitivity of file system, but lower-case configparser backport makes sense. Maybe some other code brings this library as a dependency. We shouldn't rely on it.

elacuesta:
In py3 (3.2 to be precise, but that's not supported here anymore) `SafeConfigParser` was renamed to `ConfigParser` ([source](https://github.com/python/cpython/blob/cd90f6a3692e0f7ef0a13aae651e19a08d1f9b31/Lib/configparser.py#L1224-L1234)) so there is no change there.
However, I must admit that it might have been bad move in my behalf to make the change in py2, although I suspect it shouldn't make an actual difference.

@kmike I'm not sure exactly what's the object that the `six.moves` module ends up using (and I'm too lazy to look it up :stuck_out_tongue: ), what do you think about the following? It's more explicit IMHO.

```python
if six.PY2:
    from ConfigParser import SafeConfigParser as ConfigParser
else:
    from configparser import ConfigParser
```

kmike:
@elacuesta your proposed fix looks good to me.

elacuesta:
Ok, I'll open a quick PR just now, just in case we plan to release a bugfix version.

elacuesta:
@kmike Done, please see #3896

marsam:
addressed by https://github.com/scrapy/scrapy/pull/3896

Issue #3854, opened 2019-07-04
strip trailing backslashes from domain when redirecting to avoid invalid domains


ddebernardy:
To reproduce:
```$ scrapy shell 'http://mlaib.com'
[...]
2019-07-04 12:04:30 [scrapy.downloadermiddlewares.redirect] DEBUG: Redirecting (301) to <GET http://www.belgoal.com\> from <GET http://mlaib.com>
Traceback (most recent call last):
  File "/usr/local/bin/scrapy", line 10, in <module>
    sys.exit(execute())
  File "/usr/local/lib/python3.7/site-packages/scrapy/cmdline.py", line 150, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/usr/local/lib/python3.7/site-packages/scrapy/cmdline.py", line 90, in _run_print_help
    func(*a, **kw)
  File "/usr/local/lib/python3.7/site-packages/scrapy/cmdline.py", line 157, in _run_command
    cmd.run(args, opts)
  File "/usr/local/lib/python3.7/site-packages/scrapy/commands/shell.py", line 74, in run
    shell.start(url=url, redirect=not opts.no_redirect)
  File "/usr/local/lib/python3.7/site-packages/scrapy/shell.py", line 48, in start
    self.fetch(url, spider, redirect=redirect)
  File "/usr/local/lib/python3.7/site-packages/scrapy/shell.py", line 115, in fetch
    reactor, self._schedule, request, spider)
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/threads.py", line 122, in blockingCallFromThread
    result.raiseException()
  File "/usr/local/lib/python3.7/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
ValueError: invalid hostname: www.belgoal.com\

$ curl -I http://mlaib.com
HTTP/1.1 301 Moved Permanently
[...]
Location: http://www.belgoal.com\
[...]
```

I presume I can fix it in a downloader middleware, but it should probably get handled by Scrapy.

Gallaecio:
We should probably try to imitate web browsers here.

OmarFarrag:
@Gallaecio @ddebernardy I can work on this, would check `redirected_url` in `process_response()` in `RedirectMiddleware` class.. If you suggest imitating the web browsers no problem working on that but would need to design an approach and confirm it with you first.

ddebernardy:
[...]

Issue #3855, opened 2019-07-05
Scrapy finding lots of invalid links - could we have functionality to filter them out?


ddebernardy:
Using this selector:

```
links = response.xpath('//a[@href]')
```

I picked up a lot of spaces, as well as random schemes like `mailto:`, `tel:`, and `javascript:`, which makes me presume there might be a better function. If Link extractors merely are wrappers around the above though, it might be worth adding some kind of option to sanitize link urls. (Perhaps that functionality is buried somewhere in the source code or in a section of the docs I skimmed over?)

Anyway, more importantly than the above, I found the following nuggets in my data's href fields, and methinks Scrapy should definitely provide some kind of built-in filtering mechanism for those:

```
Improperly closed tag (href is missing its closing bracket; likely an lxml bug):

https://so.dajie.com/job/search?positionFunction=120610&positionName=视频算法>
                                        视频算法
                                    </a>
                                </li>
                                <li>
                                    <a id=

---
Same issue with a pretty quote:

https://asg.to/contentsPage.html?mcd=dGQRAGplZCkLCfbw”>サンプル動画はコチラ</a><br />
</dd>
	<dt>2013.7.16</dt>
	<dd><h3>プレミアム会員限定オリジナル動画新作配信！</h3>
	アゲサゲのプレミアムメンバーに向けて完全オリジナル動画が更新されました！<br />今回はフリーターのりおちゃん！スレンダーな感じやすい体がそそります！
	<a href=

---
PHP error:

<br />
<b>Warning</b>:  Use of undefined constant url - assumed 'url' (this will throw an Error in a future version of PHP) in <b>/home/forge/jaidefinichon.com/public/wp-content/themes/jaidefinichon/header.php</b> on line <b>116</b><br />
https://jaidefinichon.com
```

Also, it've been finding urls with carriage returns, line feeds, tabs, and spaces at random location. Sometimes they'd be valid had they been url encoded. Other times, it's basically a miracle that browsers accept the links as is. I've seen the latter characters before and after each of `.`, `?`, `=`, `&`, and `#`.

It might be an upstream problem in `urlparse()`. For instance:

```
>>> url = '''https://www.booking.com/country
... .html?label=gen173nr-1FCAEoggI46AdIM1gEaGeIAQGYATG4AQfIAQ3YAQHoAQH4AQKIAgGoAgO4Av2s9OgFwAIB;sid=514ad6c12110c2103c6a2618a429af6d'''
>>> url
[...]

PR #3891, opened 2019-07-20
Fix url backslashes (invalid domains) like that in #3854 

OmarFarrag:
When trying to send a request, or even redirection, with url containing backslashes, it fails. You can see examples of that and how to reproduce the error in #3854. 

To fix that, I have added a downloader middleware called `URLMiddleware` that for now converts all backslashes in the url to forwardslashes, as this is the behavior of many browsers.

If confirmed, will immediately add tests and documentation

### DIFF
--- a/scrapy/downloadermiddlewares/url.py
+++ b/scrapy/downloadermiddlewares/url.py
@@ -1,0 +1,19 @@
+""" This module implements the URLMiddleware which does any needed operations on the url 
+    like changing backslashes to forwardslashes
+"""
+
+from scrapy import signals
+
+
+class URLMiddleware(object):
+    """ This middleware does any needed operations on the url 
+    like changing backslashes to forwardslashes
+    """
+
+    def process_request(self, request, spider):
+        new_url = request.url.replace('\\','/')
+        if new_url == request.url:
+            return
+        return request.replace(url = new_url)
+        
+
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -98,6 +98,7 @@ DOWNLOADER_MIDDLEWARES_BASE = {
     'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware': 400,
     'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware': 500,
     'scrapy.downloadermiddlewares.retry.RetryMiddleware': 550,
+    'scrapy.downloadermiddlewares.url.URLMiddleware': 555,
     'scrapy.downloadermiddlewares.ajaxcrawl.AjaxCrawlMiddleware': 560,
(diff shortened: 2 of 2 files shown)


elacuesta:
I humbly believe this should be addressed somewhere else, one possibility being [`w3lib.url.safe_url_string`](https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.safe_url_string)

OmarFarrag:
@elacuesta I thought that too, but w3lib won't support that behavior as it's not a standard (Some browsers do that, some don't) , so I think it should be handled by scrapy out of the box.. what do you think ?

elacuesta:
`w3lib.url.safe_url_string` mentions RFC-3986, and the only reference to the backslash I could find in RFC-3986 was in section 7.3 (emphasis mine):

>Special care should be taken when the URI path interpretation process
involves the use of a back-end file system or related system
functions.  **File systems typically assign an operational meaning to
special characters, such as the "/", "\\", ":", "[", and "]"
characters**, and to special device names like ".", "..", "...", "aux",
"lpt", etc.  In some cases, merely testing for the existence of such
a name will cause the operating system to pause or invoke unrelated
system calls, leading to significant security concerns regarding
denial of service and unintended data transfer.  It would be
impossible for this specification to list all such significant
characters and device names.  Implementers should research the
reserved names and characters for the types of storage device that
may be attached to their applications and restrict the use of data
obtained from URI components accordingly.

Still, I'm not even sure I can interpret that as the backslash being an "unsafe character" :man_shrugging: 

That being said, there is already a PR to add additional functionality to `safe_url_string` which is not strictly related to RFC-3986 (https://github.com/scrapy/w3lib/pull/133).

TL;DR

I believe adding a middleware for this is a bit of overkill. Even if `safe_url_string` is not the right place to fix it, this could be done in the `Request` constructor, for instance. I know that different browsers seem to behave different here, but IMHO it doesn't make any sense to comply with a browser that chokes on an (admittedly invalid) URL.

elacuesta:
Also, my understanding (I can't stress enough that this is my own personal opinion so I could be completely mistaken as to the philosophy of the library) is that `w3lib` provides helpers to deal with problems that the built-in `urllib` module doesn't solve, in _reasonable_ ways, even if that means sometimes go against some browsers. Like in this particular case, choking on an invalid URL is not reasonable to me if you can make it work by just replacing or removing a character (that shouldn't be there in the first place).
See also https://github.com/scrapy/scrapy/issues/3855#issuecomment-510559112

OmarFarrag:
@elacuesta I believe you are right. I thought we couldn't modify w3lib code that's why I didn't think of going there, but now what you say seems more reasonable to me. 

Gonna close this PR..

elacuesta:
@OmarFarrag I believe all projects under the Scrapy organization are open to contributions, so PRs are welcome. Thanks for taking an interest in this!

PR #3892, opened 2019-07-21
[MRG+1] Drop py34 support - Update CI envs

elacuesta:
Based on https://github.com/scrapy/scrapy/pull/3825#issuecomment-501447037 by @dangra 

I personally would like to see this because type hints were introduced in py35, and I'm currently working on #3881 (although it seems like _variable annotations_ were introduced in 3.6 :crying_cat_face: - [PEP 526](https://www.python.org/dev/peps/pep-0526/))


setup.py
```diff
@@ -63,10 +62,10 @@ def has_environment_marker_platform_impl_support():
         'Topic :: Software Development :: Libraries :: Application Frameworks',
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     install_requires=[
-        'Twisted>=13.1.0;python_version!="3.4"',
-        'Twisted>=13.1.0,<=19.2.0;python_version=="3.4"',
+        'Twisted>=13.1.0;python_version=="2.7"',
+        'Twisted>=15.5;python_version>="3.5"',
```
elacuesta: Bumped Twisted to `>=15.5` on py35+, according to https://github.com/scrapy/scrapy/blob/master/docs/news.rst#scrapy-110-2016-05-11 (emphasis mine)
> Scrapy 1.1 has beta Python 3 support (**requires Twisted >= 15.5**)

elacuesta:
I'm failing to understand why the py2 tests are raising `SyntaxError`:
```python
running check
  File "setup.py", line 79, in <module>
    extras_require=extras_require,
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/opt/python/2.7.14/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/opt/python/2.7.14/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/opt/python/2.7.14/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/setuptools/command/sdist.py", line 52, in run
    self.run_command(cmd_name)
  File "/opt/python/2.7.14/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/opt/python/2.7.14/lib/python2.7/distutils/dist.py", line 970, in run_command
    cmd_obj = self.get_command_obj(command)
  File "/opt/python/2.7.14/lib/python2.7/distutils/dist.py", line 845, in get_command_obj
    klass = self.get_command_class(command)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/setuptools/dist.py", line 838, in get_command_class
    self.cmdclass[command] = cmdclass = ep.load()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2434, in load
    return self.resolve()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2440, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/readme_renderer/integration/distutils.py", line 21, in <module>
    from distutils.command.check import check as _check
  File "/opt/python/2.7.14/lib/python2.7/distutils/command/check.py", line 13, in <module>
    from docutils.utils import Reporter
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/docutils/utils/__init__.py", line 21, in <module>
    import docutils.io
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/docutils/io.py", line 348
    (self.destination.mode, mode)), file=self._stderr)
                                        ^
SyntaxError: invalid syntax
```

 I know `file` is a built-in type in py2, but I don't think this PR changes those tests, and they work for me locally with the same module that triggers the error (`file=self._stderr`,  in line 348 of `docutils/io.py`):

```
$ python --version
Python 2.7.15+
$ grep --color "file=self._stderr" -C 5 -n /.../venv-py2/local/lib/python2.7/site-packages/docutils/io.py 
343-        elif (# destination is file-type object -> check mode:
344-              mode and hasattr(self.destination, 'mode')
345-              and mode != self.destination.mode):
346-                print(('Warning: Destination mode "%s" '
347-                               'differs from specified mode "%s"' %
348:                               (self.destination.mode, mode)), file=self._stderr)
349-        if not destination_path:
350-            try:
351-                self.destination_path = self.destination.name
352-            except AttributeError:
353-                pass
```

Does anyone have any clue?

elacuesta:
Well I'm having the same issue with the py2 tests on #3881, and that one definitely does not update any packages :confused:

elacuesta:
It seems to be a problem in a recent `docutils` upgrade: https://sourceforge.net/p/docutils/bugs/365/

kmike:
hey @elacuesta! I think it makes sense to do this as a part of dropping support for old OS, setting a new baseline Ubuntu/Debian. We need to decide what's the baseline, and then bump minimum package requirements and minimum Python version based on that.

kmike:
My proposal:

* Ubuntu baseline: 16.04 xenial. It is LTS release; Python 3.5.1, Twisted 16.0 (https://launchpad.net/ubuntu/xenial/+package/python3-twisted), lxml 3.5. 
* Debian baseline: stretch. It has Python 3.5.3, Twisted 16.6, lxml 3.7.1.

Ubuntu 14.04 trusty in only supported for paid customers (if I read announcements correctly), and it uses Python 3.4, twisted 13.2.0 (no Python 3 support), lxml 3.3.3.

Debian jessie has LTS support until mid-2020. But it is also getting old (Python 3.4.2, Twisted 14.0.2 without Python 3 support, lxml 3.4).

Is it too aggressive? //cc @dangra

elacuesta:
Xenial sounds good to me :+1: 
According to the [Travis docs](https://docs.travis-ci.com/user/reference/overview/#linux) Trusty is the default distribution, and it's the one we're currently using for all builds except for the Python 3.7 one.
Excuse my ignorance, but I'm failing to see how the Debian tests are executed. There is no Debian option in the docs I mentioned, and in fact the builds with `TOXENV=jessie` have `Build dist: trusty` under the `Build system information` section at the beginning of the logs.

wRAR:
@elacuesta jessie is emulated by installing the specific versions of modules into a trusty system.

wRAR:
There is a CI on salsa.debian.org, I planned to learn it and maybe set up something automatical one day.

elacuesta:
> @elacuesta jessie is emulated by installing the specific versions of modules into a trusty system.

Oh nice, that's what I suspected but I wasn't sure. Thanks!

kmike:
@elacuesta: we agreed to use Ubuntu 16.04 and Debian stretch as a new baseline for the 1.8 release. Feel free to update tox environments, requirements, Python versions, etc. to match it :)

wRAR:
So the tests fail on stretch because `twisted.protocols.ftp` wasn't ported to Py3 in that version (only since 17.1.0: https://twistedmatrix.com/trac/ticket/8865).

It seems we should only support the version already in requirements-py3.txt, which is 17.9, and for better asyncio support I would even like to bump it to 18.4. People on older systems will need a virtualenv to run scrapy anyway.

elacuesta:
That makes a lot of sense (although to be honest I could swear I tried it with 16.6.0 in a virtualenv in my machine and it worked - perhaps the virtualenv was not actually enabled or something 🤷‍♂ ).
Right, I'll bump Twisted to 17.9, which is mentioned in the release notes for 1.5.
Thanks for the debugging @wRAR 🤘

wRAR:
These problems are because of my recently merged #3450

Gallaecio:
@wRAR You mean the “Fatal Python error: Segmentation fault” error?

wRAR:
@Gallaecio yes, I'm currently looking into it, it has different behavior on different pyOpenSSL/cryptography versions

.travis.yml
```diff
@@ -6,26 +6,30 @@ branches:
...
+    - env: TOXENV=py27
       dist: xenial
-      sudo: true
-    - python: 3.6
-      env: TOXENV=docs
+      python: 2.7
+    - env: TOXENV=pypy
+      dist: xenial
+      python: 2.7
+    - env: TOXENV=stretch
```
dangra: Sorry to pop this topic "again" 🙏 . 

I don't see the point of pinning specific python packages for *Debian* (Jessie in the past, now Stretch) but testing it under an Ubuntu distribution.

IMO we should aim to define the baseline of system libraries (i.e: libssl, libxml2(?)...) which are required to install Scrapy and its dependencies. Furthermore, considering Travis-CI is our CI service, its build environment is Ubuntu based and that Ubuntu LTS releases had proven to be reliable; I vote to choose from Trusty, Xenial or Bionic as system baseline. I vogue to drop the idea of Debian as baseline simply because we can't test on it (at least for now). 

Then we intersect what the baseline system can make platform dependent packages like _pyOpenSSL_ or _cryptography_ works with, with what major libraries Scrapy depends-on (i.e. Twisted). The minimum system libraries available by the Linux distribution of choice in addition to the minimum python package versions required by Scrapy to work are what define the baseline. 

Nowadays the only things that can't be overridden in a Python virtual environment are system libraries, while Python packages can be easily isolated from the host system. IMHO Python packages packaged by distributions can't be considered constrains for Scrapy baseline. For example, if we choose Xenial as baseline system, it doesn't matter what Twisted version is provided by `python-twisted` deb package as long as Twisted can be installed without affecting the host system.

In that sense, and if all maintainers agree, I'd remove this entry from the matrix and all related entries in `tox.ini`

Gallaecio: > Python packages packaged by distributions can't be considered constrains for Scrapy baseline

I agree with this.

Gallaecio: Since [Travis CI supports Docker](https://docs.travis-ci.com/user/docker/), I wouldn’t discard Debian-based tests just because Travis CI Linux images are all Ubuntu.

That said, I don’t have an opinion on Debian vs Ubuntu (vs any other Linux distribution).

wRAR: Agreed, supporting working on system Python modules is too hard and doesn't seem useful to me.

setup.py
```diff
@@ -63,10 +62,10 @@ def has_environment_marker_platform_impl_support():
...
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     install_requires=[
-        'Twisted>=13.1.0;python_version!="3.4"',
-        'Twisted>=13.1.0,<=19.2.0;python_version=="3.4"',
+        'Twisted>=13.1.0;python_version=="2.7"',
+        'Twisted>=17.9.0;python_version>="3.5"',
         'w3lib>=1.17.0',
         'queuelib',
         'lxml;python_version!="3.4"',
```
kmike: workarounds for 3.4 can be removed

tox.ini
```diff
@@ -25,36 +25,6 @@ passenv =
...
-
-[testenv:jessie]
-# https://packages.debian.org/en/jessie/python/
-# https://packages.debian.org/en/jessie/zope/
-basepython = python2.7
-deps =
-    cryptography==0.6.1
-    pyOpenSSL==0.14
-    lxml==3.4.0
-    Twisted==14.0.2
```
kmike: Without version pinning we'll be only testing latest Twisted, etc. version. Is it intended? If yes, probably setup.py should be updated, as we're allowing early Twisted versions there, which we no longer test for. If no, it could make sense to have an environment with pinned package versions, the ones which we support - it doesn't have to be tied to Ubuntu or Debian distributions, but it can look very similar.

.travis.yml
kmike: 1) I think that's fine to not use Debian as a baseline in any way, it'd be simpler for us.
2) There is still value in allowing Scrapy to run with not the latest lxml / Twisted / cryptography(?) / Pillow, as there can be other software running in the same environment as Scrapy, which is not upgraded to the latest version yet; also, there can be regressions. We may decide to support only recent enough versions not tied to any distribution, but it should be clear what these versions are, and we should test on these versions explicitly, to make sure Scrapy works.

wRAR: So we should test on the minimal supported versions (which we should choose and put in the docs), on "current" versions and maybe on some versions in between.

kmike: Yes; having versions "in between" can be an overkill though, unless we want to support a really wide range of versions for some package.

dangra: > ... We may decide to support only recent enough versions not tied to any distribution, but it should be clear what these versions are, and we should test on these versions explicitly, to make sure Scrapy works.

It makes sense, let's add a "baseline" tox env for 1.8 that is an upgrade of current Jessie env but with versions that works on Xenial, and once we start working on 2.0 we bump it again taking Bionic as system base.

elacuesta:
Hello everyone. I updated this PR to add an environment with pinned versions that are known to work well. I'm only running it on py36, I think that's probably the most reasonable compromise to avoid adding many more envs to the Travis matrix, specially now that we have two more for the Robots parsing. I updated the `requirements-py3.txt` file with a few comments explaining the reasons for the picked versions.
To be honest I didn't pay much attention to the py2 tests, I'm not sure it's worth at this point.
Let me know what you think, and as usual, thanks for your comments :rocket:

kmike:
Thanks @elacuesta!

> I'm only running it on py36

Why not 3.5? Oldest Python, oldest dependencies?

> To be honest I didn't pay much attention to the py2 tests, I'm not sure it's worth at this point.

I think it worths it, as we may end up supporting 1.8 a bit longer than other versions. There is also a different Twisted minimum version requirement in setup.py; if we're not bumping it to match Python 3, it would be good to have a py27-pinned-versions env as well.

setup.py
```diff
@@ -63,20 +62,19 @@ def has_environment_marker_platform_impl_support():
...
-        'Twisted>=13.1.0,<=19.2.0;python_version=="3.4"',
-        'w3lib>=1.17.0',
-        'queuelib',
-        'lxml;python_version!="3.4"',
-        'lxml<=4.3.5;python_version=="3.4"',
-        'pyOpenSSL',
-        'cssselect>=0.9',
-        'six>=1.5.2',
-        'parsel>=1.5',
+        'Twisted>=13.1.0;python_version=="2.7"',
```
kmike: Do we need to support Twisted 13.1.0? It is already 16.0.0 on xenial.

elacuesta:
Hey @kmike, I made some updates based on your feedback. Thanks!

tox.ini
```diff
@@ -25,72 +25,68 @@ passenv =
...
-    boto==2.34.0
-    Pillow==2.6.1
+    -ctests/constraints.txt
     cssselect==0.9.1
-    zope.interface==4.1.1
```
Gallaecio: I see some previously-pinned libraries are no longer pinned, namely `cryptography`, `Pillow` and `zope.interface`. Maybe we should pin them as well?

As for boto/botocore, if we are dropping support for boto maybe we should update the documentation accordingly (see https://github.com/scrapy/scrapy/pull/3294/files).

tox.ini
```diff
@@ -25,72 +25,68 @@ passenv =
...
+    PyDispatcher==2.0.5
+    pyOpenSSL==16.2.0
+    queuelib==1.4.2
+    service_identity==16.0.0
+    six==1.10.0
+    Twisted==17.9.0
+    w3lib==1.17.0
+    -rtests/requirements-py3.txt
+    # Extras
+    Pillow
```
Gallaecio: Comparing to the Python 2 pinned environment, I notice botocore missing here. Is this on purpose?

tox.ini
elacuesta: True, those dependencies were only pinned in the Debian env, so they were deleted along with it. I can add them to the requirements files and the pinned deps lists. Thanks for noticing.

elacuesta:
Updated, added minimum and pinned versions of `cryptography` and `zope.interface`. `botocore` and `Pillow` are only installed for the tests, as they are only required by certain parts (S3 feed storage and images pipeline IIRC) and are not _required_ for Scrapy itself to run.

elacuesta:
Random Travis error about not being able to install `tox` in one of the `py27` envs. All other tests are green, closing and reopening to trigger another build.

kmike:
Thanks @elacuesta! Farewell Python 3.4!

PR #3899, opened 2019-07-24
[Py3] Item loaders: allow single argument functions as processors

elacuesta:
Python 3 only: `__qualname__` was added in Python 3.3 ([PEP 3155](https://www.python.org/dev/peps/pep-3155/)). I thought about a py2-compatible version using `six.get_method_function` (which accesses `__func__` in py3 and `im_func` in py2) and doing some regex checks on the obtained function's name, but that wasn't pretty. Since we're dropping py2 support soon and this is technically backward incompatible I figured it would be ok to leave this for Scrapy 2.0

Initially reported in #2735

### DIFF
--- a/scrapy/loader/__init__.py
+++ b/scrapy/loader/__init__.py
@@ -4,8 +4,7 @@ Item Loader
 See documentation in docs/topics/loaders.rst
 """
 from collections import defaultdict
-
-import six
+from contextlib import suppress
 
 from scrapy.item import Item
 from scrapy.loader.common import wrap_loader_context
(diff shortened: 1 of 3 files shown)


scrapy/loader/__init__.py
```diff
@@ -3,16 +3,26 @@
 See documentation in docs/topics/loaders.rst
 
 """
+from contextlib import suppress
```
Gallaecio: :lipstick: a-z sorting, both here and in the Scrapy imports below.

elacuesta: Done, thanks!

PR #3907, opened 2019-07-27
[MRG+1] Update _monkeypatches.py

sbs2001:
The workarounds are not required assuming the bugs regarding urlparse are absent in  Python versions >2.7. We already exit the program if Python  version<2.7 in the __init__.py(line 17).The monkeypatches are deployed after this check at line 27  in  the __init__.py .

### DIFF
--- a/scrapy/_monkeypatches.py
+++ b/scrapy/_monkeypatches.py
@@ -1,21 +1,4 @@
-import six
 from six.moves import copyreg
-
-
-if six.PY2:
-    from urlparse import urlparse
-
-    # workaround for https://bugs.python.org/issue7904 - Python < 2.7
-    if urlparse('s3://bucket/key').netloc != 'bucket':
-        from urlparse import uses_netloc
-        uses_netloc.append('s3')
-
-    # workaround for https://bugs.python.org/issue9374 - Python < 2.7.4
-    if urlparse('s3://bucket/key?key=value').query != 'key=value':
-        from urlparse import uses_query
-        uses_query.append('s3')
-
-
 # Undo what Twisted's perspective broker adds to pickle register
 # to prevent bugs like Twisted#7989 while serializing requests
 import twisted.persisted.styles  # NOQA
--- a/scrapy/_monkeypatches.py
+++ b/scrapy/_monkeypatches.py
@@ -1,20 +1,20 @@
-from six.moves import copyreg
 import six
+from six.moves import copyreg
 
-if six.PY2:    
+
+if six.PY2:
     from urlparse import urlparse
+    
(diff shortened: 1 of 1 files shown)


scrapy/_monkeypatches.py
```diff
@@ -1,21 +1,4 @@
-import six
 from six.moves import copyreg
-
-
-if six.PY2:
-    from urlparse import urlparse
-
-    # workaround for https://bugs.python.org/issue7904 - Python < 2.7
```
kmike: 👍

scrapy/_monkeypatches.py
```diff
@@ -1,21 +1,4 @@
...
-    if urlparse('s3://bucket/key').netloc != 'bucket':
-        from urlparse import uses_netloc
-        uses_netloc.append('s3')
-
-    # workaround for https://bugs.python.org/issue9374 - Python < 2.7.4
```
kmike: I think we shouldn't remove the second workaround, as it may be needed in 2.7 < Python < 2.7.4

sbs2001: Yep my bad, python 2.7.4 is still permitted . I am newbie to the whole project so this may sound stupid but why is the Travis CI build is failing? I noticed it with other pull requests too.

sbs2001: Added the second workaround in  the new commit.

Gallaecio: > why is the Travis CI build is failing? I noticed it with other pull requests too.

At the moment #3913 and #3917. If you stick around, though, I’m sure you’ll have the opportunity to see it start failing for other reasons; it happens now and then :slightly_smiling_face:

sbs2001:
> Thanks for your patch, but looking at the changes I see a few unrelated format changes, some of them for the worse. Could you please make your changes only remove the unnecessary workaround?

Sorry for the formatting. Anyway I have updated the code accordingly.Please review.

scrapy/_monkeypatches.py
```diff
@@ -4,18 +4,13 @@
...
-        uses_netloc.append('s3')
-
+    
     # workaround for https://bugs.python.org/issue9374 - Python < 2.7.4
     if urlparse('s3://bucket/key?key=value').query != 'key=value':
         from urlparse import uses_query
         uses_query.append('s3')
 
-
+
```
Gallaecio: You missed a spot :slightly_smiling_face:

sbs2001: Fixed I hope.

kmike:
Thanks @sbs2001, and thanks @Gallaecio for the review!

Issue #1668, opened 2016-01-11
Set support for robots.txt to be on by default


mvl22:
We are seeing sites receiving crawls with this bot's signature which disobey robots.txt, falling into our honeytrap.

Please set support for robots.txt to be on by default. Currently the default is do not obey:

http://doc.scrapy.org/en/latest/topics/settings.html#robotstxt-obey

I would argue that crawlers have a responsibility to ensure that they act as a well-behaved web citizen by default. This applies even more to open-source code crawlers, which are likely to be used by people the author has no control over.

So please kindly ensure this behaviour is on by default. If someone is installing this crawler, and they are using it on their own website, they will have full knowledge of why robots.txt rules are in place, and so can decide to disable the feature. If they are using it on another site, then the robots.txt should be respected, because that is what the site owner has specifically defined as being reasonable usage.

kmike:
+1 to make this change. 

It is backwards-incompatible though, so ideally we should do a deprecation cycle, which can get a bit tricky.

redapple:
+1 for me too.
I would also like to have autothrottle on by default

mvl22:
> It is backwards-incompatible though, so ideally we should do a deprecation cycle

Sure; that's understandable. As long as things are set in place to get the change made, so that scrapy can be a good net citizen by default.

PR #3908, opened 2019-07-28
Update default_settings.py ,ROBOTSTXT_OBEY should be by default true

sbs2001:
I guess by default scrapy is supposed to obey robots.txt, hence the edit.


Gallaecio:
@sbs2001 This has been suggested in the past. To maintain backward compatibility, the approach back then was to keep the default value as `False` but make the `startproject` subcommand generate a `settings.py` file where it’s set to `True`.

@kmike I wonder if we should consider changing this for 2.0. Maybe even enabling autothrottle by default, as @redapple [suggested in the past](https://github.com/scrapy/scrapy/issues/1668#issuecomment-170545983).

PR #3909, opened 2019-07-28
Refactor `open_in_browser`

OmarFarrag:
Refactor `open_in_browser` function by adding `file_extension` property to response classes, so to get the extension only use `response.file_extension`

### DIFF
--- a/scrapy/http/response/__init__.py
+++ b/scrapy/http/response/__init__.py
@@ -134,3 +134,8 @@ class Response(object_ref):
                        dont_filter=dont_filter,
                        errback=errback,
                        cb_kwargs=cb_kwargs)
+    
+    @property
+    def file_extension(self):
+        raise  TypeError("Unsupported response type: %s" %
+                        self.__class__.__name__)
--- a/scrapy/http/response/html.py
+++ b/scrapy/http/response/html.py
@@ -8,4 +8,6 @@ See documentation in docs/topics/request-response.rst
 from scrapy.http.response.text import TextResponse
 
 class HtmlResponse(TextResponse):
-    pass
+    @property
+    def file_extension(self):
+        return '.html'
--- a/scrapy/http/response/text.py
+++ b/scrapy/http/response/text.py
@@ -158,6 +158,9 @@ class TextResponse(Response):
             cb_kwargs=cb_kwargs,
         )
 
+    @property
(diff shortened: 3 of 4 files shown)


Gallaecio:
I’m not convinced by this approach of adding a `file_extension` property to HTTP response objects.

We could implement something in the response that returns the MIME type (I’m not sure if we should rely on the Content-Type header when present, or directly on the body), and have the open_in_browser logic determine a fitting file extension based on the MIME type. What do you think?

OmarFarrag:
@Gallaecio  I agree with that, I did think that the response class shouldn't know about file extensions, but I couldn't think of another approach. So, I think we may add mime type property to the base response class, and check for Content-Type if exists, and if not extract it from the body, like done with encoding. Then, in `open_in_browser` define a dict with mime types as keys and corresponding extensions as values.
If you are okay with that approach, I can start working right away.

Gallaecio:
I think that can work.

It’s only indirectly related, but you might want to check out `scrapy.responsetypes`. It seems to handle the choosing of a Response subclass based on the underlying response data MIME type, Content-Type header, etc.

Ideally, where `scrapy.responsetypes` is used, the MIME type of the corresponding response should be set to the one used to determine the Response subclass to use. However, such a change would be tedious, specially in a backward-compatible way that includes deprecation warnings where needed. So feel free _not_ to go for that optimization as part of these changes. However, once these changes are merged, opening a enhancement proposal ticket for that would be great.

I think the best thing in `scrapy.responsetypes` for you is the use of `MimeTypes`, which has a `guess_extension` method.

kmike:
The standard we should follow when dealing with MIME types is https://mimesniff.spec.whatwg.org/; Scrapy doesn't handle it yet.

OmarFarrag:
Okay so let me get that straight, What we want to do for now is to add `mime_type` property to the base `Response` class that returns MIME type extracting it from `Content-Type` header, and in `open_in_browser` use `MimeTypes.guess_extxension`
**But**, only after that I start working on implementing the standard mentioned by @kmike in `responsetypes` and use it to extract the MIME type for any response and set the response's MIME type to that extracted.
That's right ?

OmarFarrag:
Okay cool. Will close this, work on the new PR, and mention you when done.

Issue #3871, opened 2019-07-12
Deprecate hacky code from get_project_settings()


Gallaecio:
[Reported](https://github.com/scrapy/scrapy/pull/3859#issuecomment-510838622) by @nyov:

> @kmike, would you or someone perhaps also find time to correctly deprecate this (or just rip it > out)?: https://github.com/scrapy/scrapy/blob/9c90d9515a50ede29415b8b5d6ba11229f333b49/scrapy/utils/project.py#L70-L79
> Or is it still needed.

elacuesta:
Deprecated by #3910 and #4300.

PR #3910, opened 2019-07-28
[MRG+1] Update project.py removed one 'hack', seems irrelevant.

sbs2001:
As mentioned by @Gallaecio in issue #3871, the 'hack'  is cleared. I also  double checked whether the environment variable "SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE" was ever set in our codebase and it turns out we didn't set it or used it anywhere else.So I guess the 'hack' was not used in the current version.

### DIFF
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -73,7 +73,7 @@ def get_project_settings():
     if pickled_settings:
         warnings.warn("Use of environment variable "
                       "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' "
-                      " is deprecated.",ScrapyDeprecationWarning)
+                      "is deprecated.", ScrapyDeprecationWarning)
         settings.setdict(pickle.loads(pickled_settings), priority='project')
 
     # XXX: deprecate and remove this functionality
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -66,7 +66,13 @@ def get_project_settings():
     settings_module_path = os.environ.get(ENVVAR)
     if settings_module_path:
         settings.setmodule(settings_module_path, priority='project')
-    
+
+    # XXX: remove this hack
+    pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
+    if pickled_settings:
+        warnings.warn("Use of environmental variable 'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' is deprecated",DeprecationWarning)
+        settings.setdict(pickle.loads(pickled_settings), priority='project')
+
     # XXX: deprecate and remove this functionality
     env_overrides = {k[7:]: v for k, v in os.environ.items() if
                      k.startswith('SCRAPY_')}
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -66,8 +66,6 @@ def get_project_settings():
     settings_module_path = os.environ.get(ENVVAR)
     if settings_module_path:
         settings.setmodule(settings_module_path, priority='project')
-
-    # XXX: remove this hack
     
     # XXX: deprecate and remove this functionality
     env_overrides = {k[7:]: v for k, v in os.environ.items() if
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -68,10 +68,7 @@ def get_project_settings():
         settings.setmodule(settings_module_path, priority='project')
 
     # XXX: remove this hack
-    pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
-    if pickled_settings:
-        settings.setdict(pickle.loads(pickled_settings), priority='project')
-
+    
     # XXX: deprecate and remove this functionality
     env_overrides = {k[7:]: v for k, v in os.environ.items() if
                      k.startswith('SCRAPY_')}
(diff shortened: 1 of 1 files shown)


scrapy/utils/project.py
```diff
@@ -68,10 +68,7 @@ def get_project_settings():
         settings.setmodule(settings_module_path, priority='project')
 
     # XXX: remove this hack
```
noviluni: I think that `# XXX: remove this hack` should be removed as well.

sbs2001: The code actually has 2 hacks,so thought it won't matter. Anyway I am doing what you suggested. Thank you.

Gallaecio:
I think we should to first deprecate the variable, logging a deprecation warning when it’s defined, and actually drop it at a latter version.

sbs2001:
I would like to handle that. Please can you give me an idea about how the warning should look like and  also the type of warning.

scrapy/utils/project.py
```diff
@@ -70,6 +70,7 @@ def get_project_settings():
...
     pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
     if pickled_settings:
+        warnings.warn("Use of environmental variable 'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' is deprecated",DeprecationWarning)
```
Gallaecio: A few notes:
* We use `ScrapyDeprecationWarning` (from `scrapy.exceptions`) so that deprecation warnings are not silenced by default in Python 2.
* environmental variable → environment variable
* Please, split the warning about multiple lines. See [Coding Style](https://docs.scrapy.org/en/latest/contributing.html#coding-style).

sbs2001: @Gallaecio Is it ok now? Please review.

Gallaecio: Not exactly. Breaking lines like that creates extra space between the words before and after line breaks. Please, take https://github.com/scrapy/scrapy/blob/master/scrapy/utils/log.py#L196-L198 as an example.

scrapy/utils/project.py
```diff
@@ -70,6 +71,9 @@ def get_project_settings():
     # XXX: remove this hack
     pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
     if pickled_settings:
+        warnings.warn("Use of environment variable "
+                      "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE'"
```
Gallaecio:
```suggestion
                      "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' "
```

scrapy/utils/project.py
```diff
@@ -70,6 +71,9 @@ def get_project_settings():
...
+                      " is deprecated.",ScrapyDeprecationWarning)
```
Gallaecio:
```suggestion
                      "is deprecated.", ScrapyDeprecationWarning)
```

noviluni: To follow PEP8 it's necessary to add a whitespace after the comma (that means between: `" is deprecated.",` and `ScrapyDeprecationWarning`)

```suggestion
                      " is deprecated.", ScrapyDeprecationWarning)
```

noviluni: Already suggested by Gallaecio, resolving :)

scrapy/utils/project.py
sbs2001: @Gallaecio  I checked  whether breaking lines like that creates extra spaces, it didn't actually.Anyways I have made the changes according to the example.

```
>>> warnings.warn("Use of environment variable \
... 'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' \
... is deprecated",DeprecationWarning)
__main__:3: DeprecationWarning: Use of environment variable 'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' is deprecated

>>> warnings.warn("Use of environment variable "
... "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE'"
... " is deprecated",DeprecationWarning)
__main__:3: DeprecationWarning: Use of environment variable 'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' is deprecated
```
Is that problem maybe specific to ScrapyDeprecationWarning?

Gallaecio: I would say it’s a problem specific to Python :)

```
[adrian@pondal ~]$ python3
Python 3.6.7 (default, Nov 27 2018, 21:03:46) 
[GCC 8.2.0] on linux
>>> print("adsf \
...        asdf")
adsf        asdf
>>> 
```

In your test you are not using indentation, as in our code. The problem with using a backslash to escape a line break is that all the indentation spaces of the next line remain.

scrapy/utils/project.py
```diff
@@ -1,6 +1,7 @@
 import os
 from six.moves import cPickle as pickle
 import warnings
+from scrapy.exceptions import ScrapyDeprecationWarning
```
kmike: one more nitpick: could you please move this import to another scrapy.exceptions import below? According to pep8, imports should be grouped in 3 groups: stdlib imports, third-party imports, project imports; this one should be in  the last group.

sbs2001: @kmike  missed that one thanks for the suggestion. I have made the changes accordingly.

kmike:
Thanks @sbs2001 for a fix, and everyone else for feedback!

PR #3920, opened 2019-07-30
Fix memory handling and error handling in utils.ssl.get_temp_key_info.

wRAR:
Found in #3892. Tested on xenial with various pyOpenSSL and cryptography combinations.


kmike:
I have no idea how it works, but as it fixes a segfault, you tested it, and all this code is disabled by default, +1 to merge it :)

PR #3927, opened 2019-08-01
Update mail.py,gz.py,decompression.py

sbs2001:
This seems better,we are preventing the whole error this way instead of getting along it.

### DIFF
--- a/scrapy/downloadermiddlewares/decompression.py
+++ b/scrapy/downloadermiddlewares/decompression.py
@@ -8,8 +8,11 @@ import zipfile
 import tarfile
 import logging
 import six
-from six import BytesIO
 from tempfile import mktemp
+if six.PY2:
+    from cStringIO import StringIO as BytesIO
+else:
+    from io import BytesIO
 
 from scrapy.responsetypes import responsetypes
 
--- a/scrapy/downloadermiddlewares/decompression.py
+++ b/scrapy/downloadermiddlewares/decompression.py
@@ -7,8 +7,8 @@ import gzip
 import zipfile
 import tarfile
 import logging
-import six
 from tempfile import mktemp
+import six
 if six.PY2:
     from cStringIO import StringIO as BytesIO
 else:
--- a/scrapy/downloadermiddlewares/decompression.py
+++ b/scrapy/downloadermiddlewares/decompression.py
@@ -7,14 +7,9 @@ import gzip
 import zipfile
 import tarfile
 import logging
-from tempfile import mktemp
-
 import six
-
-try:
-    from cStringIO import StringIO as BytesIO
-except ImportError:
-    from io import BytesIO
+from six import BytesIO
+from tempfile import mktemp
 
 from scrapy.responsetypes import responsetypes
 
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -12,9 +12,11 @@ from six.moves.email_mime_text import MIMEText
 from six.moves.email_mime_base import MIMEBase
 from six import BytesIO
 if six.PY2:
+    from cStringIO import StringIO as BytesIO
     from email.MIMENonMultipart import MIMENonMultipart
     from email import Encoders
 else:
+    from io import BytesIO
     from email.mime.nonmultipart import MIMENonMultipart
     from email import encoders as Encoders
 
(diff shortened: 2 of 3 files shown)


scrapy/mail.py
```diff
@@ -4,11 +4,6 @@
 See documentation in docs/topics/email.rst
 """
 import logging
-
-try:
-    from cStringIO import StringIO as BytesIO
-except ImportError:
-    from io import BytesIO
 import six
```
dangra: Considering we are using `six`, I'd go for https://six.readthedocs.io/#six.BytesIO

sbs2001: Your suggestions seem better, I have made the changes accordingly.Thank you.

sbs2001:
I discovered that we are using the same 'try and except' in several other files I have fixed them.

scrapy/downloadermiddlewares/decompression.py
```diff
@@ -7,14 +7,9 @@
...
 import logging
-from tempfile import mktemp
-
 import six
-
-try:
-    from cStringIO import StringIO as BytesIO
-except ImportError:
-    from io import BytesIO
+from six import BytesIO
```
kmike: This is not the same as before, because six defines BytesIO as StringIO.StringIO, not cStringIO.StringIO, which may cause a big performance hit.

kmike:
I think try/except in Python is fine, or at least debatable; it seems to be a recommended approach in Python docs - see https://docs.python.org/3/glossary.html#term-eafp

scrapy/downloadermiddlewares/decompression.py
sbs2001: @kmike can you check my first commit. We can replace try/except with if/else.

sbs2001:
@kmike do you think it will be better for us if we replace try/except with if/else.

kmike:
@sbs2001 I think in this particular case your original change looks good. As a general case, Python convention is EAFP, not [LBYL](https://docs.python.org/3/glossary.html#term-lbyl).

sbs2001:
> @sbs2001 I think in this particular case your original change looks good. However, as a general rule, Python convention is EAFP, not [LBYL](https://docs.python.org/3/glossary.html#term-lbyl).

Should I replace 'try/except' with  'if/else'  then?

scrapy/mail.py
```diff
@@ -4,17 +4,13 @@
...
 
 from email.utils import COMMASPACE, formatdate
 from six.moves.email_mime_multipart import MIMEMultipart
 from six.moves.email_mime_text import MIMEText
 from six.moves.email_mime_base import MIMEBase
+from six import BytesIO
```
kmike: If I'm not missing something, we need to preserve cStringIO here as well

sbs2001: Sorry, missed that, thanks for correcting me.

scrapy/utils/gz.py
```diff
@@ -20,9 +14,11 @@
 # - scrapy does not support Python 3.2
 # - Python 2.7 GzipFile works fine with standard read() + extrabuf
 if six.PY2:
+    from cStringIO import StringIO as BytesIO
```
kmike: I think now it reads like BytesIO is imported because of GzipFile.read problem described above, while it doesn't.

scrapy/downloadermiddlewares/decompression.py
```diff
@@ -7,13 +7,11 @@
 import zipfile
 import tarfile
 import logging
-from tempfile import mktemp
-
 import six
-
-try:
+from tempfile import mktemp
```
kmike: a nitpick - it was good to have tempfile import before six, because tempfile is in stdlib. We kind-of consider six to be stdlib as well, because you import stdlib modules using six, but still, in most cases it is below. Sorry for nitpicks! But the PR is about style changes anyways, right? :)

scrapy/utils/gz.py
sbs2001: I didn't realise that, thank you for bringing it to my notice.

scrapy/downloadermiddlewares/decompression.py
sbs2001: Please don't feel sorry for suggesting nitpicks. I am myself kind of suggesting nitpicks too. In fact  I'd be thankful if you keep suggesting them even if they seem trivial to you.

sbs2001:
@Gallaecio  why has the code coverage for  gz.py has decreased by such an amount.

Gallaecio:
I wouldn’t worry too much about coverage in this case.

scrapy/utils/gz.py
kmike: hey @sbs2001! I don't think it is solved: read1 functions work aroung gzip problems, and comment should relate to them, not to io imports

wRAR:
is this still needed after dropping py2 and six?

Issue #3929, opened 2019-08-02
JSONRequest naming issue


noviluni:
A few months ago `JSONRequest` was added to Scrapy (https://github.com/scrapy/scrapy/pull/3505) and released with the new Scrapy version 1.7.

However, I realized that the name is not in accordance with the rest of the classes: `HtmlResponse`, `XmlResponse`, `XmlRpcRequest`.

It should be named:
`JsonRequest`.

I know it isn't backward-compatible, but as it's a really recent addition and Scrapy 2.0 is on the way it could be fixed.

Gallaecio:
_Maybe_ we should consider renaming the other 3 instead to [comply with PEP8](https://stackoverflow.com/a/2853609).

noviluni:
@Gallaecio Totally agree with you. We should decide what to do: 
* to change it or not

If we decide to change it:
* to follow PEP8 or not

I have just opened an issue regarding PEP8: https://github.com/scrapy/scrapy/issues/3944. Maybe it should be addressed before deciding the second question.

Gallaecio:
> Hey! As JSONRequest is what pep8 recommends, we (me, @dangra, @Gallaecio and @wRAR) agreed to rename other classes instead. PRs are welcome :)

@kmike at https://github.com/scrapy/scrapy/pull/3930#issuecomment-519518983

noviluni:
Looking over I have seen that this issue affects at least the following classes:

```
HtmlResponse
HtmlParserLinkExtractor

HttpProxyMiddleware
HttpAuthMiddleware
HttpCacheMiddleware
HttpCompressionMiddleware
HttpErrorMiddleware
HttpError

JsonWriterPipeline
JsonLinesItemExporter
JsonItemExporter

XmlRpcRequest
XmlItemExporter
XmlResponse
```

kmike:
👍 Thanks for checking it @noviluni! I was thinking it is less things to fix..

kmike:
Tbh, I'd leave everything as-is, probably including JSONRequest; it is not super consistent, but my feeling is that changing it doesn't worth the development efforts required to upgrade (change in settings, imports, etc.)

noviluni:
I'm agree with you, there are too much classes to deprecate... If you other are agree feel free to close this issue,

kmike:
We may still change JSONRequest though, to at least match other Request classes - that's what you propsed in #3930 .

noviluni:
[...]

PR #3930, opened 2019-08-02
fix JSONRequest naming issue

noviluni:
fixes https://github.com/scrapy/scrapy/issues/3929

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -64,7 +64,7 @@ New features
     provides a cleaner way to pass keyword arguments to callback methods
     (:issue:`1138`, :issue:`3563`)
 
-*   A new :class:`~scrapy.http.JSONRequest` class offers a more convenient way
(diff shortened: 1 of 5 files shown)


dangra:
Backwards compatibility is important in case we move forward with this change, which I am +0.5 if you ask, therefore take a look at `create_deprecated_class()` util:

```python
from scrapy.utils.deprecate import create_deprecated_class
# ...
JSONRequest = create_deprecated_class("JSONRequest", JsonRequest)
```

 https://github.com/scrapy/scrapy/blob/8e813953bd4b25b60d30981a948e48fee43f8dad/scrapy/utils/deprecate.py#L15-L48

kmike:
Hey! As JSONRequest is what pep8 recommends, we (me, @dangra, @Gallaecio and @wRAR) agreed to rename other classes instead. PRs are welcome :)

Issue #3903, opened 2019-07-25
Can I get remote server's ip address via response?


imfht:
Can I get remote server's ip address via response?
For some reason. I'll need get remote site's ip address when parsing response. I looked the document but found nothing.
Any one know that?
Thanks!

anubhavp28:
You could use the `url` property of the [Response](https://docs.scrapy.org/en/latest/topics/request-response.html#response-objects) object, pass it to [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) fuction to obtain the domain name. Then you can pass the domain to [socket.gethostbyname](https://docs.python.org/3/library/socket.html#socket.gethostbyname) to get the IP address. I haven't tested it, but it should work.

Gallaecio:
I don’t think this is possible at the moment, but it sounds like a valid feature request.

imfht:
Thanks @anubhavp28 .
I don't think it is the right way to get remote ip in scrapy
get_host_by_name cause another new dns resolve. Actually we've got remote ip in scrapy.
class `CrawlerProcess` has an [built-in dnscache](https://github.com/scrapy/scrapy/blob/8e813953bd4b25b60d30981a948e48fee43f8dad/scrapy/resolver.py#L10) object but we can not get it in pipeline.

PR #3933, opened 2019-08-03
Add `host_ip` property to response

OmarFarrag:
This PR adds support to get the host IP address from a response as proposed in #3903 .
Will write tests once approved

### DIFF
--- a/scrapy/http/response/__init__.py
+++ b/scrapy/http/response/__init__.py
@@ -4,7 +4,8 @@ responses in Scrapy.
 
 See documentation in docs/topics/request-response.rst
 """
-from six.moves.urllib.parse import urljoin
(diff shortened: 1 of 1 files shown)


elacuesta:
Hi there @OmarFarrag, thanks for working on this. Unfortunately, I believe this might not be the right approach. AFAICT sockets in Python work in blocking mode by default. There is some documentation about [blocking/non-blocking mode](https://docs.python.org/3/library/socket.html#socket.socket.setblocking), but I'm not even sure that plays well with the underlying Twisted infrastructure.
That said, I must admit I don't have a lot of experience working directly with network sockets, so I'm open to be proven wrong.

imfht:
I don't think it is the right way to get remote ip.
get_host_by_name cause another new dns resolve. Actually we've got the address in scrapy.

PR #3937, opened 2019-08-05
Update reactor.py, updated 'if' sequencing .

sbs2001:
This should be the proper ordering.This is the explanation.
  If `'not portrange'` is True ,it is guaranteed that `not hasattr(portrange, '__iter__')`  is also True  the converse of this is not always true.(for example, consider portrange=None, for such case we were executing the logic for `not hasattr(portrange, '__iter__')` . ).Such case is eliminated by this PR.


Webson35 (commented):
D154077E-81CF-4A8A-82A8-E230B64AA2A0

PR #3939, opened 2019-08-05
Report all missing fields when a scrapes contract fails

Gallaecio:
Hijacking of #766 addressing remaining feedback.


kmike:
Looks good, thanks @Gallaecio and @tpeng!

Issue #1031, opened 2015-01-31
IPv6 support? Problem running home page example from an IPv6 network


ronnix:
I'm running into problems while trying to run the example on the scrapy.org home page from the FOSDEM IPv6-only Wi-Fi network. (The scraper works fine from an IPv4 network.)

If both IPv4 and IPv6 are enabled on my computer (OS X Yosemite), and the IPv4 is configured with DHCP, and thus gets a self-assigned address (169.254.x.x), then I get timeout errors:

```
$ scrapy runspider myspider.py
:0: UserWarning: You do not have a working installation of the service_identity module: 'No module named service_identity'.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module and a recent enough pyOpenSSL to support it, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.
2015-01-31 12:13:32+0100 [scrapy] INFO: Scrapy 0.24.4 started (bot: scrapybot)
2015-01-31 12:13:32+0100 [scrapy] INFO: Optional features available: ssl, http11
2015-01-31 12:13:32+0100 [scrapy] INFO: Overridden settings: {}
2015-01-31 12:13:32+0100 [scrapy] INFO: Enabled extensions: LogStats, TelnetConsole, CloseSpider, WebService, CoreStats, SpiderState
2015-01-31 12:13:32+0100 [scrapy] INFO: Enabled downloader middlewares: HttpAuthMiddleware, DownloadTimeoutMiddleware, UserAgentMiddleware, RetryMiddleware, DefaultHeadersMiddleware, MetaRefreshMiddleware, HttpCompressionMiddleware, RedirectMiddleware, CookiesMiddleware, ChunkedTransferMiddleware, DownloaderStats
2015-01-31 12:13:32+0100 [scrapy] INFO: Enabled spider middlewares: HttpErrorMiddleware, OffsiteMiddleware, RefererMiddleware, UrlLengthMiddleware, DepthMiddleware
2015-01-31 12:13:32+0100 [scrapy] INFO: Enabled item pipelines:
2015-01-31 12:13:32+0100 [blogspider] INFO: Spider opened
2015-01-31 12:13:32+0100 [blogspider] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
[...]

PR #3940, opened 2019-08-05
Get server IP address for HTTP/1.1 Responses

### LINKED ISSUE #3903: Can I get remote server's ip address via response?

elacuesta:
Fixes #3903

Extract from a `scrapy shell` session:
```
(...)
2019-08-05 16:41:17 [scrapy.core.engine] INFO: Spider opened
2019-08-05 16:41:18 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://example.org> (referer: None)
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7fc3e2cb2f98>
[s]   item       {}
[s]   request    <GET https://example.org>
[s]   response   <200 https://example.org>
[s]   settings   <scrapy.settings.Settings object at 0x7fc3deadd748>
[s]   spider     <DefaultSpider 'default' at 0x7fc3de0d6860>
[s] Useful shortcuts:
In [1]: response.ip_address
Out[1]: IPv4Address('93.184.216.34')

In [2]: fetch('https://docs.scrapy.org')
2019-08-05 16:41:48 [scrapy.downloadermiddlewares.redirect] DEBUG: Redirecting (302) to <GET https://docs.scrapy.org/en/latest/> from <GET https://docs.scrapy.org>
2019-08-05 16:41:49 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://docs.scrapy.org/en/latest/> (referer: None)

In [3]: response.ip_address
Out[3]: IPv4Address('104.17.32.82')
```

For reference, it seems like there is no _public_ API for getting this directly from Twisted: https://twistedmatrix.com/trac/ticket/9030

**Tasks**:

- [x] Implementation
- [x] Tests
- [x] Docs

**Update**:
Py3-only: Implemented using the [`ipaddress` module](https://docs.python.org/3/library/ipaddress.html), only available in Python 3.3+.


OmarFarrag:
I am having in my mind a few scenarios where `dnscache.get(hostname)` might not work:
1- If the cached ip was popped out of the cache due to limit constraints ( `scrapy.utils.datatypes.LocalCache.__setitem__`)
2- If the `DNSCACHE_ENABLED` is false, the ip is not cached (`scrapy.resolver.getHostByName`)

OmarFarrag:
I believe `reactor.resolve` would be the right way, provided that the reactor uses the pre installed resolver which is the `CachingThreadedResolver`, but I can't seem to figure from twisted documentation if that is the behavior or not. Hope this helps you :)

Gallaecio:
The cache will not always work, as @OmarFarrag points out. And resolving after the response has been received is not the answer either, the IP address may not match the server that actually sent the response.

It seems to me like the right way would be to modify the low-level code that builds the response using the Twisted API, and use something like [transport.getHost()](https://twistedmatrix.com/documents/8.0.0/api/twisted.internet.interfaces.ITransport.html#getHost) to get the actual IP address that sent the response. The code would probably be much more complex, though.

Also, if we go this route, we might want to consider making the server IP address part of the `Response` class, instead of a `Response.meta` key.

elacuesta:
That's great feedback, thanks both! 
I'm on mobile now, so without rechecking the code I suspect the `DNSCACHE_ENABLED` setting might indeed be the cause of the failed test.
I will try to build something based upon your comments,  thanks again @OmarFarrag and @Gallaecio :metal:

elacuesta:
Updated to use [`getPeer()`](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.ITransport.getPeer.html) as per [this StackOverflow answer](https://stackoverflow.com/a/42034252). It gets the data by accessing a private attribute, but it's actually not the first private Twisted attribute in use in that file :sweat_smile: 
Also, as Adrián suggested I added a `server_ip_address` attribute (other naming suggestions are welcome of course), because indeed the server IP belongs to the responses, not to the requests. I just need to update the docs to reflect this.
I hope this makes more sense now, looking forward to reading your feedback!

tests/test_crawl.py
```diff
@@ -277,3 +278,30 @@ def test_crawl_multiple(self):
...
         self._assert_retried(log)
         self.assertIn("Got response 200", str(log))
+
+    @defer.inlineCallbacks
+    def test_dns_server_ip_address(self):
+        from socket import gethostbyname
+
+        crawler = self.runner.create_crawler(SingleRequestSpider)
+        url = 'https://example.org'
+        yield crawler.crawl(seed=url)
```
Gallaecio: I’m not sure about making a request to https://example.org as part of the test suite. @kmike Thoughts?

Gallaecio:
Looks great. Let’s now get some reviewers more familiar with Scrapy’s internals.

tests/test_crawl.py
elacuesta: I don't like it either, and I've been told about this before, precisely by Mikhail 😄 (https://github.com/scrapy/scrapy/pull/2061#pullrequestreview-25525781)
But I don't know how else to check for an actual DNS resolution, the MockServer URLs uses only IP addresses.

Gallaecio:
I do wonder if we should consider a shorter name, though. (I know, I know, I proposed the current one). Would `Response.address` work?

elacuesta:
> Would `Response.address` work?

I think it does, let's do it.

docs/topics/request-response.rst
```diff
@@ -607,6 +607,9 @@ Response objects
         This represents the :class:`Request` that generated this response.
     :type request: :class:`Request` object
 
+    :param address: The IP address of the server from which the Response originated.
+    :type address: |twisted.internet.address.IPv4Address|_ / |twisted.internet.address.IPv6Address|_ object
```
Gallaecio: I don’t think Scrapy supports IPv6 yet (#1031), so this might be misleading.

We will have to remember to mention `twisted.internet.address.IPv6Address` once IPv6 support is implemented, though.

docs/topics/request-response.rst
```diff
@@ -814,3 +823,7 @@ XmlResponse objects
 
 .. _Twisted Failure: https://twistedmatrix.com/documents/current/api/twisted.python.failure.Failure.html
 .. _bug in lxml: https://bugs.launchpad.net/lxml/+bug/1665241
+.. |twisted.internet.address.IPv4Address| replace:: ``twisted.internet.address.IPv4Address``
```
Gallaecio: This is awesome, I had no idea this could be done to use both monospace font and a link.

However, since we got [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) now, you could add an entry to `intersphinx_mapping` (in `docs/conf.py`) for the Twisted API (probably `'twisted': ('https://twistedmatrix.com/documents/current/api', None),`) and use:

```
:class:`twisted.internet.address.IPv4Address`
```

docs/topics/request-response.rst
elacuesta: Interesting, I wasn't aware of that. I think I can remove the v6 reference and leave a FIXME comment. The links between the issues should also help not to forget about it.

docs/topics/request-response.rst
elacuesta: Nice, will do.

elacuesta: I didn't know it was possible either, I found this approach in a StackOverflow answer :-)
It seems a little bit of a hack to me tough, since it involves two steps to create the link and replace the text. The `intersphinx` approach looks much cleaner, thanks!

docs/topics/media-pipeline.rst
```diff
@@ -442,7 +442,7 @@ See here the methods that you can override in your custom Files Pipeline:
         successfully or ``False`` if it failed for some reason
 
       * ``file_info_or_error`` is a dict containing the following keys (if success
-        is ``True``) or a `Twisted Failure`_ if there was a problem.
+        is ``True``) or a :class:`twisted.python.failure.Failure` object if there was a problem.
```
Gallaecio: :lipstick: 
```suggestion
        is ``True``) or a :class:`~twisted.python.failure.Failure` object if there was a problem.
```

(_if_ you want it to show up as just `Failure`)

elacuesta:
> Although that test… I can’t help but wonder if having no test at all would be better smile. Let’s see what @kmike thinks.

hahaha fair enough :smile:

docs/topics/media-pipeline.rst
elacuesta: If you don't mind, I'd prefer to keep the full class path. It's good to know that syntax trick tough :wink:

docs/topics/request-response.rst
```diff
@@ -607,6 +606,12 @@ Response objects
         This represents the :class:`Request` that generated this response.
     :type request: :class:`Request` object
 
+    :param address: The IP address of the server from which the Response originated.
+    :type address: :class:`twisted.internet.address.IPv4Address` object
```
kmike: Hey @elacuesta and @Gallaecio! Why do you prefer address to ip_address, or server_ip_address? :) I liked previous name more, because there can be different meaning of response.address: URL address, company address string extracted from the response content, some address in memory, etc.

docs/topics/request-response.rst
```diff
@@ -675,6 +680,14 @@ Response objects
         they're shown on the string representation of the Response (`__str__`
         method) which is used by the engine for logging.
 
+    .. attribute:: Response.address
+
+        A Twisted object containing information about the server IP address.
+        See :class:`twisted.internet.address.IPv4Address` for additional information.
```
kmike: It could be nice to avoid exposing twisted-specific structures, in case we move away from Twisted

docs/topics/request-response.rst
Gallaecio: `ip_address` works for me.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -445,6 +447,17 @@ def __init__(self, finished, txresponse, request, maxsize, warnsize, fail_on_dat
...
         self._reached_warnsize = False
         self._bytes_received = 0
+        self._ip_address = None
+
+    def connectionMade(self):
+        if self._ip_address is None:
+            peer = self.transport._producer.getPeer()
+            if isinstance(peer, TxIPv4Address):
+                self._ip_address = IPv4Address(type=peer.type, host=peer.host, port=peer.port)
+            # FIXME: uncomment once IPv6 is supported
```
elacuesta: This could either be kept like this, or be removed and added when IPv6 support is added. The same with the `scrapy.utils.datatypes.IPv6Address` definition.

scrapy/utils/datatypes.py
```diff
@@ -20,6 +20,13 @@
     Mapping = collections.abc.Mapping
 
 
+# See https://twistedmatrix.com/documents/current/api/twisted.internet.address.html
+IPv4Address = collections.namedtuple('IPv4Address', ['type', 'host', 'port'])
```
kmike: one more option is to use https://docs.python.org/3/library/ipaddress.html#ip-addresses; it is new in 3.3, but there can be backport available

kmike: I think that's fine not to return type and port, as they should be available from url (?). In stdlib ip address data structure doesn't include port and type - and it makes sense to me, as they are not a part of ip address

elacuesta: Here I am re-inventing the wheel :-)

I have a few conflicting ideas about this. On the one hand, why not expose information that we have available, like type and port? The overhead of the additional classes is minimal, since they are `namedtuples`. On the other hand, I understand that if the idea is to be Twisted-independent, an eventual future implementation might not expose that, and we'd need to do backward incompatible changes. But then again, how far are we from moving away from Twisted? I wonder if that is perhaps too much early optimization :stuck_out_tongue: 

In addition, I think adding a 3rd party library backport for py2 is in fact a bit of an overkill. What about returning an `ipaddress.IPv4Address` in py3 and a plain string in py2?

kmike: one more option: wait for 1.8 (we aim for August), drop Python 2 support in master, merge afterwards

elacuesta: Yes, that sounds reasonable. Let me change the title and add this to the 2.0 milestone. Thanks for the feedback :-)

elacuesta: Done. py2 tests are failing, which is expected.

tests/test_crawl.py
wRAR: I don’t like this too, I’ll think what to do about it

kmike: +1 not to use example.com, even if it means we need to figure out how to start a dummy DNS server in tests. It looks doable, as e.g. twisted has DNS server implementation (see https://twistedmatrix.com/documents/current/names/howto/custom-server.html)

elacuesta: Indeed, I have a DNS forwarder running on a Raspberry Pi which is based on (should I say "copied from"?)  that page. I'll try to come up with a mock server. Thanks for the pointer.

kmike:
Could you please check how https://github.com/scrapy/scrapy/pull/4227 affects this PR?

elacuesta:
I tried with a mock DNS server but that didn't work well (the client actually, not the server, see https://stackoverflow.com/q/59902042/6946615).

> Could you please check how #4227 affects this PR?

I was able to use the `DNS_RESOLVER` setting to enable a mock resolver that would resolve all names to `127.0.0.1`, so no actual DNS request is performed. Since this resolver affects the reactor, I thought it'd be best to test it with a CrawlerProcess script.

Looking forward to hearing your comments.

tests/test_crawl.py
elacuesta: Upstream request removed 🙌

elacuesta:
@kmike After the latest developments in the above SO question, I was able to create a mock DNS server, which resolves everything to `127.0.0.1`.
`tests.mockserver.MockDNSServer` is similar to `tests.mockserver.MockServer`, it can be used as a context manager as well.

docs/topics/request-response.rst
```diff
@@ -679,6 +682,10 @@ Response objects
         they're shown on the string representation of the Response (`__str__`
         method) which is used by the engine for logging.
 
+    .. attribute:: Response.ip_address
+
+        The IP address of the server from which the Response originated.
```
kmike: I'm not sure we follow it well for other attributes, but it'd be good to say that it can be None as well. Maybe also mention when this may happen ("not all download handlers may support this attribute" or something like that, maybe in a more user-friendly way, as nobody knows what's a download handler).

elacuesta: Added a note about this. I intentionally didn't mention the S3 handler, which uses the HTTP handler internally, because those responses are technically also http. Also didn't mention responses with no body (https://github.com/scrapy/scrapy/pull/3940/files#diff-18150b1d259c93bf10bf1d4e5028d753R384-R386), I think that's probably a very specific edge case.

tests/CrawlerRunner/ip_address.py
```diff
@@ -0,0 +1,38 @@
...
+        url = "http://not.a.real.domain:{port}/echo".format(port=port)
+
+        servers = [(mock_dns_server.host, mock_dns_server.port)]
+        reactor.installResolver(createResolver(servers=servers))
+
+        configure_logging()
+        runner = CrawlerRunner()
+        d = runner.crawl(LocalhostSpider, url=url)
+        d.addBoth(lambda _: reactor.stop())
+        reactor.run()
```
elacuesta, line 38: Updated this block @kmike, thanks for pointing it out :+1:

kmike:
Great job, thanks @elacuesta and everyone involved!

PR #3943, opened 2019-08-06
Update file_path() example in media-pipeline.rst

jachang820:
`def file_path(self, request, response, info):` throws error `Failure instance: Traceback (failure with no frames): <class 'TypeError'>: file_path() missing 1 required positional argument: 'response'` under item_completed() results. Setting default arguments fixes the issue.

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -529,7 +529,7 @@ See here the methods that you can override in your custom Images Pipeline:
 
         class MyImagesPipeline(ImagesPipeline):
 
-            def file_path(self, request, response, info):
+            def file_path(self, request, response = None, info = None):
                 return 'files/' + os.path.basename(urlparse(request.url).path)
 
       By default the :meth:`file_path` method returns


kmike:
@jachang820 which Scrapy version are you using?

jachang820:
> 
> 
> @jachang820 which Scrapy version are you using?

1.7.3

Gallaecio:
@jachang820 Could you check if turning the `path = self.file_path(request, info=info)` line into `path = self.file_path(request, response=None, info=info)` in `scrapy/pipelines/files.py` fixes the issue for you?

You may already know this, but you can clone the Scrapy code locally, modify it and install it with `pip install -e <path to the cloned folder>` into any virtual environment, or even globally.

Gallaecio:
It looks like the signature has been `file_path(self, request, response=None, info=None)` for years, so it looks indeed like a documentation bug to me.

@jachang820 Could you please apply the fix also to line 419 of the same file, and keep the spacing as in the original code? (`def file_path(self, request, response=None, info=None):`)

wRAR:
Fixed in #4290

Issue #3944, opened 2019-08-07
Addressing PEP8 by including pytest-flake8


noviluni:
As mentioned in the Scrapy **Coding Style** section in the docs (https://docs.scrapy.org/en/master/contributing.html#coding-style) :

> “Unless otherwise specified, follow PEP 8.”.

However, it’s known that the current codebase has a lot of parts that don’t respect PEP8 guidelines (https://github.com/scrapy/scrapy/issues/2144).

I know that it’s difficult to address that, as it involves a lot of code to be reviewed but, however, the more time we skip that task the harder it becomes.

Two obvious things that we could do are:
* Fixing the current code
* Adding a pipeline to check pep8

Fixing the current code can’t be done in just one commit, but instead of fixing pep8 in specific project parts (example: https://github.com/scrapy/scrapy/pull/2207, https://github.com/scrapy/scrapy/pull/2429), it could better and easier to review if we fix rules one by one (for example: "Unused imports", "Missing blank lines after class or function definition."...). 

My proposal to avoid to write a hard to maintain pipeline is to include the `pytest-flake8`  package that executes `flake8` when executing `pytest` (by adding the `--flake8` flag) and raises an error when the check fails. 

If we **start excluding all the rules** then we can just **fix one by one** from more obvious rules to more discussed rules just **excluding that rule from the ignored rules list**. In that way we ensure that new code respects the previous agreed flake8/pep8 rules and on the other hand **we can discuss every rule** if we want to follow it or not (example: https://github.com/scrapy/scrapy/pull/3697).

I will add a PR to illustrate this idea.

noviluni:
What do you think?? I’ve been evaluating other possible solutions and I’ve seen that `scikit-learn` uses a pipeline to check flake8 against the new/diff lines with a bash script https://github.com/scikit-learn/scikit-learn/blob/master/build_tools/circle/flake8_diff.sh but it seems really hard to mantain.

Gallaecio:
Maybe we could have a separate tox environment for these analysis? (#3727 could also be run there)

noviluni:
Hi @Gallaecio . Sounds good, I will try to update my PR when I have time.

noviluni:
Hi @Gallaecio I've updated the PR (https://github.com/scrapy/scrapy/pull/3945) with a new tox environment and a its travis pipeline and seems that worked. Feel free to review it and give your feedback :)

noviluni:
[...]

PR #3945, opened 2019-08-07
Add pytest-flake8

noviluni:
proposal for: https://github.com/scrapy/scrapy/issues/3944

~Please, take in mind that it’s only a PoC. In this way the testing time increases and maybe it’s not necessary to check flake8 against all environments.~


noviluni:
Seems that pipeline worked: https://travis-ci.org/scrapy/scrapy/jobs/569955789

Gallaecio:
@noviluni Looks great. However be ready for something like https://github.com/scrapy/scrapy/pull/3727#issuecomment-501902135 from @dangra :)

noviluni:
Hi @Gallaecio thank you for your feedback. I've been thinking that, and I'm not totally agree (in fact, I can't fully understand the reasons)... The idea here is to avoid big PRs to be reviewed and adding `noqa` marks to every line/function/file implies a lot of changes :/

As my original proposal was to avoid that, for me it has no sense. I can't see the benefits of adding all those lines and then delete them when fixing the code. It involves more work (twice) and probably more "risk" (looking at your PR: 170 files to review is kind of risky).

@dangra what do you think? Is it necessary to mark all the cases one by one and then start fixing them or could we start this way and when fixing every flake8 rule add the necessary `noqa` to those "exceptions" we want to keep?

Thanks in advance

noviluni:
Updated to use a list files (with the excluded rules) instead of ignoring all rules directly. In that way is possible to disable checks only for the files that are not passing them.

.travis.yml
```diff
@@ -7,6 +7,8 @@ branches:
     - /^\d\.\d+\.\d+(rc\d+|\.dev\d+)?$/
 matrix:
   include:
+    - env: TOXENV=flake8
+      python: 3.5
```
Gallaecio: Shouldn’t we use 3.7? (also relevant for `tox.ini`)

pytest.ini
```diff
@@ -4,3 +4,256 @@ python_files=test_*.py __init__.py
...
+    scrapy/pqueues.py E501
+    scrapy/resolver.py E302
+    scrapy/responsetypes.py E128 E501 E305
+    scrapy/robotstxt.py E302 E501
+    scrapy/shell.py E501
+    scrapy/signalmanager.py E501
+    scrapy/spiderloader.py E225 F841 E501 E126
+    scrapy/squeues.py E128
+    scrapy/statscollectors.py E501 W391
+    # tests
```
Gallaecio: Shouldn’t tests be excluded from these checks?

.travis.yml
noviluni: It's ok for me, in fact I don't remember why I put 3.5.... Already pushed a commit with this change, thank you :+1:

pytest.ini
noviluni: hmmm I'm don't see any reason to do that, could you explain me why? :thinking: 

From my point of view, fixing the PEP8/style in the tests will contribute to their maintainability.

noviluni:
> I’ve left a couple of comments, but I’m OK with merging this _as is_.

thank you for your feedback

PR #3946, opened 2019-08-07
[MRG+1] Simplify version reporting

elacuesta:
After the recent updates of #3892 the minimum package versions for `w3lib` and `pyOpenSSL` were upgraded beyond the cases considered in the existing checks, I think those checks are not needed anymore.

Also, I'm sure there is a historical reason for catching `ImportError` on `cryptography`, but it is also required in `setup.py` so it should always be available for import.

### DIFF
--- a/scrapy/utils/versions.py
+++ b/scrapy/utils/versions.py
@@ -4,12 +4,12 @@ import sys
 import cryptography
 import cssselect
 import lxml.etree
-import OpenSSL
 import parsel
 import twisted
 import w3lib
 
 import scrapy
+from scrapy.utils.ssl import get_openssl_version
 
 
 def scrapy_components_versions():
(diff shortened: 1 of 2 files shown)


scrapy/utils/versions.py
```diff
@@ -13,38 +15,24 @@
...
-        import OpenSSL
-        openssl = OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION)\
-            .decode('ascii', errors='replace')
-    # pyOpenSSL 0.12 does not expose openssl version
-    except AttributeError:
-        openssl = 'Unknown OpenSSL version'
-
+    openssl = OpenSSL.SSL.SSLeay_version(
+        OpenSSL.SSL.SSLEAY_VERSION
+    ).decode('ascii', errors='replace')
```
kmike: What do you think about moving this to https://github.com/scrapy/scrapy/blob/master/scrapy/utils/ssl.py?

elacuesta: Sounds reasonable, I can do it later today. Thanks!

PR #3950, opened 2019-08-09
Remove obsolete version checks

elacuesta:
Removing more version checks after bumping minimum package versions in #3892 (Twisted, pyOpenSSL)

I must admit, in the last commit I might have got too excited about indentation and PEP8 (but not the 79-char limit), so feel free to disregard that one (hopefully you won't :crossed_fingers: :smile:)

### DIFF
--- a/scrapy/core/downloader/contextfactory.py
+++ b/scrapy/core/downloader/contextfactory.py
@@ -1,112 +1,85 @@
 from OpenSSL import SSL
-from twisted.internet.ssl import ClientContextFactory
+from twisted.internet.ssl import optionsForClientTLS, CertificateOptions, platformTrust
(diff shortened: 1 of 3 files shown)


elacuesta:
I know, I know, sorry about that :sweat_smile: 
In my defense, I suspected there should be a way to view only a subset of the changes, and indeed it seems like Githhub does allow to select a specific range of commits to review: the first 8 commits can be seen at https://github.com/scrapy/scrapy/pull/3950/files/d5dcc5eaef80ef383dff90f19349c0e06f1836a6. The version-specific changes are more obvious that way in the downloader handler file.

kmike:
Thanks @elacuesta!

Issue #3957, opened 2019-08-14
responsetypes.py can have better backward compatibility


sbs2001:
In line 8 we are directly importing StringIO from  io  which  as far as  I know is not supported in  Python 2.x.x

I think we should first check  the Python version and then do the import.

Gallaecio:
https://docs.python.org/2.7/library/io.html#io.StringIO

PR #3958, opened 2019-08-14
Update responsetypes.py solves the issue #3957

sbs2001:
I have first checked for the python version and then  imported StringIO.

### DIFF
--- a/scrapy/responsetypes.py
+++ b/scrapy/responsetypes.py
@@ -5,8 +5,12 @@ based on different criteria.
 from __future__ import absolute_import
 from mimetypes import MimeTypes
 from pkgutil import get_data
-from io import StringIO
 import six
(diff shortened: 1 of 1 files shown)


anubhavp28:
Scrapy only supports Python 2.7 and it does have `io.StringIO` - [link](https://docs.python.org/2/library/io.html#io.StringIO).

Gallaecio:
This is based on #3927.

While the logic sounds reasonable (the `cStringIO` should be more efficient in Python 2), [cStringIO.StringIO does not support Unicode, while io.StringIO does](https://stackoverflow.com/a/37463095/939364).

So, while a similar change for BytesIO was fine, I don’t think this change should be done.

Gallaecio:
@sbs2001 Shall we close this?

PR #3961, opened 2019-08-18
Add FTPFileStore to FilesPipeline

### LINKED ISSUE #3928: Save images on remote server
I want to save images on a remote server via (FTP?). Is this possible with scrapy? 
(I know about S3 or Google but i mean in a custom FTP server).

OmarFarrag:
In this PR, a `FTPFileStore` class is added to `scrapy.pipelines.files` to add support for saving files on FTP servers. 

I added `close_spider` function to the `FilesPipeline` and from it call `close_connection` function of the store object, if exists. This allows opening the FTP connection only once at initialization and closing it when the spider is closed (This is different from `scrapy.extensions.feedexports` where the connection is opened and closed every time an item is exported).

There are two things I need feedback on them. First, if the username and password should be supplied in the `file_store` setting or have separated settings fields for them. Second thing if a more exception handling is needed when opening the connection.

After resolving these things, will write tests and documentation :) 

CLOSES #3928


webtekindo:
def __init__(self, uri):
        assert uri.startswith('ftp://')
        u = urlparse(uri)
        self.ftp = FTP()
        self.ftp.connect(u.hostname, u.port or '21')
        self.ftp.login(u.username, u.password)
        self.basedir = u.path + '/'
        ftp_makedirs_cwd(self.ftp, self.basedir)

    def persist_file(self, path, buf, info, meta=None, headers=None):
        buf.seek(0)
        filename = path.split('/')[-1]
        ftp_makedirs_cwd(self.ftp, self.basedir + '/'.join(path.split('/')[:-1]))
        return threads.deferToThread(
            self.ftp.storbinary,
            'STOR %s' % filename,
            buf
        )


I'm not sure if this is the best solution but something like that will allow to use custom file_path.

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,42 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
 
 
+class FTPFilesStore(object):
+    
+    def __init__(self, uri):
+        assert uri.startswith('ftp://')
+        u = urlparse(uri)
+        self.ftp = FTP()
+        self.ftp.connect(u.hostname, u.port or '21')
+        self.ftp.login(u.username, u.password)
```
Gallaecio: Maybe we could use the existing `FTP_USERNAME` and `FTP_PASSWORD` settings as fallbacks if the URL does not define a username or password?

See how it’s done for `AWS_ACCESS_KEY_ID` and similar in `FilesPipeline`.

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,42 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
+        self.basedir = u.path + '/'
+        ftp_makedirs_cwd(self.ftp, self.basedir+'full')
```
Gallaecio: Why are you modifying the received path, adding a `/` to it?

Moreover, in this line you are creating a hardcoded `/full` directory; while the default file pipelines do use such a directory, this code should support alternative file pipelines that do not.

You should probably extract the destination path from filenames on `persist_file`, and create the target folder if you have not created it before.

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,42 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
+                last_modified = float(self.ftp.voidcmd("MDTM " + self.basedir + '/' + path)[4:].strip())
+                m = hashlib.md5()
+                self.ftp.retrbinary('RETR %s' % self.basedir + path, m.update)
+                return {'last_modified': last_modified, 'checksum': m.hexdigest()}
+            # The file doesn't exist
+            except Exception as e :
+                return {} 
+        return threads.deferToThread(_stat_file, path)
+    
+    def close_connection(self):
```
Gallaecio: Is this really necessary? Isn’t it better to just let the connection remain open until the `self.ftp` object gets destroyed naturally?

scrapy/pipelines/files.py
OmarFarrag: I did this to avoid checking the existence of the directory every time in `persist_file`, but like you said and also mentioned by @webtekindo it should support custom paths

scrapy/pipelines/files.py
OmarFarrag: Let me check that :)

scrapy/pipelines/files.py
OmarFarrag: Cool, will add that :)

scrapy/pipelines/files.py
OmarFarrag: The `FTP` class does the cleanup in its `__exit__` method, so yeah it is not necessary like you said :D

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,50 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
+        
+    def persist_file(self, path, buf, info, meta=None, headers=None):
+        buf.seek(0)
+        # If the path is like 'x/y/z.ext' the 'x/y' is rel_path and 
+        # 'z.ext' is file name
+        # If path is only the file name 'z.ext', then rel_path is
+        # the empty string and filename is 'z.ext'
+        x = path.rsplit('/',1)
+        rel_path, filename = ('/' + x[0], x[1]) if len(x) > 1 else ('', x[0])
+        abs_path = self.basedir + rel_path
```
Gallaecio: What if basedir ends with `/`?

OmarFarrag:
@Gallaecio I have considered your comments and modified the code .. 
I have a note, the `ImagesPipeline` inherits from the `FilesPipeline` and it implements `from_settings` method just the same except for getting the `store_uri`, I could add the ftp username and password extraction to that method as well, but I suggest creating one method in `FilesPipeline` that does this implementation and takes store identification parameter and this method would be called by the two `from_settings` methods of `FilesPipeline` and `ImagesPipeline`

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,50 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
+        ftp_makedirs_cwd(self.ftp, abs_path)
```
Gallaecio: FTP does not seem thread-safe, given it has a `cwd` parameter. Maybe we do need to create separate FTP instance to upload each file.

scrapy/pipelines/files.py
OmarFarrag: 'Never trust the user' is what I should have done , hahaha. Will fix that

scrapy/pipelines/files.py
OmarFarrag: I believe you are right, and now I see why `FTPFeedStorage` had this specific implementation. Kinda same, right ?

OmarFarrag:
@Gallaecio I do believe this implementation is nearly the same as `scrapy.extensions.feedexport.FTPFeedStorage` and may be refactored

scrapy/pipelines/files.py
```diff
@@ -248,6 +250,53 @@ def persist_file(self, path, buf, info, meta=None, headers=None):
...
+
+    FTP_USERNAME = None
+    FTP_PASSWORD = None
+    
+    def __init__(self, uri):
+        assert uri.startswith('ftp://')
+        u = urlparse(uri)  
+        self.port = u.port
+        self.host = u.hostname
+        self.port = int(u.port or '21')
```
Gallaecio: :lipstick: 

```suggestion
        self.port = int(u.port or 21)
```

scrapy/pipelines/files.py
```diff
@@ -284,7 +330,6 @@ def __init__(self, store_uri, download_func=None, settings=None):
 
         if isinstance(settings, dict) or settings is None:
             settings = Settings(settings)
-
```
Gallaecio: :lipstick:  Please, undo this and the next line removals.

scrapy/pipelines/files.py
```diff
@@ -321,6 +365,11 @@ def from_settings(cls, settings):
         gcs_store.GCS_PROJECT_ID = settings['GCS_PROJECT_ID']
         gcs_store.POLICY = settings['FILES_STORE_GCS_ACL'] or None
 
+        ftp_store = cls.STORE_SCHEMES['ftp']
+        ftp_store.FTP_USERNAME = settings['FTP_USER']           # Default is 'anonymous'
```
Gallaecio: :lipstick: I think it’s better to remove the comments about their default values; in case we update the default values in the future, we are will likely not remember to update them here as well.

scrapy/utils/ftp.py
```diff
@@ -13,3 +15,20 @@ def ftp_makedirs_cwd(ftp, path, first_call=True):
...
+    ftp = FTP()
+    ftp.connect(host, port)
+    ftp.login(username, password)
+    if use_active_mode:
+        ftp.set_pasv(False)
+    file.seek(0)
+    dirname, filename = posixpath.split(path)
+    ftp_makedirs_cwd(ftp, dirname)
+    ftp.storbinary('STOR %s' % filename, file)
+    ftp.quit()
```
Gallaecio: :lipstick: It’s customary to leave a line at the end of the file, and recommended in PEP8, I believe.

scrapy/pipelines/files.py
OmarFarrag: Will do

OmarFarrag:
@Gallaecio I added settings extraction in `ImagesPipeline`, just the same as in `FilesPipeline`, and I do also believe this could be refactored. In both cases, I will start with the tests and docs now.

scrapy/utils/ftp.py
OmarFarrag: Will leave one :)

scrapy/pipelines/files.py
OmarFarrag: When did I do that 😆

webtekindo:
We very often get an error:

error_perm: 530 Sorry, the maximum number of clients (10) for this user are already connected.

Can we somehow keep only one FTP connection instead of connecting each time? That will also improve performance.

Gallaecio:
We could make the number of FTP clients configurable. Whether to do that as part of these changes or in a later pull request, I’m not sure.

OmarFarrag:
@Gallaecio @webtekindo Okay, once this PR is merged, I will open another PR and add this option.

@Gallaecio I am working on the test now but there's something I wanna understand, the line `uri = os.environ.get('ENVVAR')`, how to know the value of the env var. If the test is to be written like for S3 and GCS, it would be `FTP_TEST_FILE_URI` , but in `test_feedexport.FTPFeedStorageTest` , `FEEDTEST_FTP_URI` is used.

Gallaecio:
I think `FTP_TEST_FILE_URI` is OK for this test.

docs/topics/media-pipeline.rst
```diff
@@ -148,6 +148,27 @@ Where:
...
+------------------
+
+.. setting:: FTP_USER
+.. setting:: FTP_PASSWORD
+
+:setting:`FILES_STORE` and :setting:`IMAGES_STORE` can represent a FTP server.
+Scrapy will automatically upload the files to the server.
+
+:setting:`FILES_STORE` value: should be written in the form 
+`ftp://username:password@address:port/path` or `ftp://address:port/path`. In
```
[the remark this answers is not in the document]
OmarFarrag: Moved them to a formatted separate paragraph, better ?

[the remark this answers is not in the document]
OmarFarrag: Will remove the note. I think we shouldn't care about empty `path`s from the beginning.

OmarFarrag:
> Left a few comments on the documentation. Sorry for being very picky 😅

Totally okay of course 😄 Actually it's first time working with .rst files so I kinda had no idea what I was doing 😆

docs/topics/media-pipeline.rst
```diff
@@ -148,6 +148,25 @@ Where:
...
+:setting:`FILES_STORE` and :setting:`IMAGES_STORE` can point to an FTP server.
+Scrapy will automatically upload the files to the server.
+
+:setting:`FILES_STORE` and :setting:`IMAGES_STORE` should be written in one of the
+following forms::
+
+    ftp://username:password@address:port/path
+    ftp://address:port/path
+    
+If ``username`` and ``password`` are not provided, they are taken from :setting:`FTP_USER` and
```
Gallaecio:
```suggestion
If ``username`` and ``password`` are not provided, they are taken from the :setting:`FTP_USER` and
```

scrapy/utils/ftp.py
```diff
@@ -13,3 +15,20 @@ def ftp_makedirs_cwd(ftp, path, first_call=True):
         ftp.mkd(path)
         if first_call:
             ftp.cwd(path)
+
+def ftp_store_file( 
+        path, file, host ,port,
+        username, password, use_active_mode=False):
```
kmike: what do you think about making all argument kwargs-only, to discourage positional argument usage here?

scrapy/utils/ftp.py
```diff
@@ -13,3 +15,20 @@ def ftp_makedirs_cwd(ftp, path, first_call=True):
         ftp.mkd(path)
         if first_call:
             ftp.cwd(path)
+
+def ftp_store_file( 
+        path, file, host ,port,
```
kmike:
```suggestion
        path, file, host, port,
```

scrapy/utils/ftp.py
```diff
@@ -13,3 +15,20 @@ def ftp_makedirs_cwd(ftp, path, first_call=True):
...
+        username, password, use_active_mode=False):
+    """Opens a FTP connection with passed credentials,sets current directory
+    to the directory extracted from given path, then uploads the file to server
+    """
+    ftp = FTP()
```
kmike: I think it can be better to use context manager here (see https://docs.python.org/3/library/ftplib.html#ftplib.FTP); this may allow to remove some code, and handle errors in a more correct way.

sauraCactus:
+1

sauraCactus:
Is there any option to get this feature before official merge?

Thanks

Gallaecio:
@sauraCactus You could [install](https://stackoverflow.com/a/20101940/939364) the corresponding branch of @OmarFarrag’s repository with the feature.

scrapy/utils/ftp.py
OmarFarrag: For error handling, for now exceptions are just thrown, do you suggest creating new custom exceptions and throw them instead ? If this is not the case, I don't think it will make a difference. I can change it though.

scrapy/utils/ftp.py
OmarFarrag: I think it would be better yes, I just saw most of the code base use positional arguments so didn't try to be odd 😄  go for it ?

scrapy/utils/ftp.py
kmike: I mean use 

```py
with FTP(host, user=username, passwd=password) as ftp:
    if use_active_mode:
        ftp.set_pasv(False)
    # ...   
```
This would ensure ftp.quit is called even if some step in the middle fails.

But now, as I'm looking at it, it is not clear how to pass `port` argument this way. No idea why isn't it a part of api.

scrapy/utils/ftp.py
kmike: go for it :) It is a new function, so no backwards compatibility issues.

kmike: and thanks for trying to follow the style of the existing code - it makes sense, and makes reviewing easier.

OmarFarrag: Done :)

scrapy/utils/ftp.py
OmarFarrag: Yes, the `FTP` constructor has no `port` parameter, it uses the default port, but the `connect` function has `port` parameter, so might use a work around like
```
with FTP() as ftp:
    ftp.connect(host, port)
    ftp.login(username, password)
    # ...
```
Not much code will be removed then, but port could be changed and `quit` would always be called. what do you think?

kmike: sounds good @OmarFarrag! Could you please also check merge conflicts?

OmarFarrag: Okay on it

kmike:
I haven't checked how it works, but the code looks good to me - merging, relying on @OmarFarrag to get it right :) Thanks @OmarFarrag, @Gallaecio and @wRAR!

PR #3966, opened 2019-08-19
Adds ROBOTSTXT_USER_AGENT setting

### LINKED ISSUE #3931: Using separate user agent for robots.txt
The current implementation of `RobotsTxtMiddleware` ([link](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/robotstxt.py)) uses `USER_AGENT` setting and `User-Agent` header of the request to determine the user agent to use for matching in `robots.txt`. There is no easy way for a user to send one user-agent header value but have a different user agent string evaluated to decide which pages to crawl. This use case seems to be more common that we thought previously for example - while [Googlebot](https://support.google.com/webmasters/answer/182072?hl=en) uses `"Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"`  as `User-Agent` header of the request, it actually uses user agent `"Googlebot"` in `robots.txt` for deciding which pages to crawl. [Here is the an article describing this.](https://support.google.com/webmasters/answer/1061943?hl=en) The same seems to be the case with Bing - [link](https://www.bing.com/webmaster/help/which-crawlers-does-bing-use-8c184ec0). Infact, Google made this distinction very clear in their proposed [robots.txt specification](https://github.com/google/robotstxt/blob/master/protocol-draft/draft-rep-wg-topic-00.txt) where they described the user agent used in `robots.txt` as the "product token".

The proposed solutions are implementing a new `ROBOTSTXT_USER_AGENT` setting, implementing a `robotstxt_user_agent` meta key for overriding user agent per request or both. Here is a thread that has previous discussion on the topic - https://github.com/scrapy/scrapy/pull/3796#discussion_r300036852.

I'm creating this issue as a place for discussion regarding whether to add this functionality into Scrapy? If yes, which route we should take?

anubhavp28:
Fixes #3931

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -989,6 +989,12 @@ RobotsTxtMiddleware
     To make sure Scrapy respects robots.txt make sure the middleware is enabled
     and the :setting:`ROBOTSTXT_OBEY` setting is enabled.
 
+    The :setting:`ROBOTSTXT_USER_AGENT` setting can be used to specify the
+    user agent string to use for matching in the robots.txt_ file. If it
+    is ``None``, the User-Agent header you are sending with the request or the
+    :setting:`USER_AGENT` setting (in that order) will be used for determining
+    the user agent to use in the robots.txt_ file.
+
     This middleware has to be combined with a robots.txt_ parser.
 
     Scrapy ships with support for the following robots.txt_ parsers:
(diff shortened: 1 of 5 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -1074,6 +1074,21 @@ implementing the methods described below.
 .. autoclass:: RobotParser
    :members:
 
+RobotsTxtMiddleware Settings
```
Gallaecio: The setting should only be described in one page, probably the settings page. In this page you can instead mention that the user agent that the middleware uses may be overridden with this setting, providing a link to the settings page entry. See how `ROBOTSTXT_OBEY` is referenced in this page but its documentation is in the settings page only.

docs/topics/settings.rst
```diff
@@ -1409,7 +1421,9 @@ USER_AGENT
 
 Default: ``"Scrapy/VERSION (+https://scrapy.org)"``
 
-The default User-Agent to use when crawling, unless overridden.
+The default User-Agent to use when crawling, unless overridden. This user agent is
+also used in robots.txt if :setting:`ROBOTSTXT_USER_AGENT` setting is ``None`` and
```
Gallaecio: Instead of `in robots.txt`, I would say `by RobotsTxtMiddleware`, with a link to the middleware documentation.

kmike:
Thanks @anubhavp28!

PR #3967, opened 2019-08-19
Allow users to use SitemapSpider with the new robots.txt interface.

anubhavp28:


Gallaecio:
I’m unsure about this feature.

What I would like to see, and I think would be an improvement over the current implementation, is to be able to use both RobotsTxtMiddleware and SitemapSpider in combination, and that a request for robots.txt is issued only once.

Maybe a way to issue requests with some `meta` key so that the RobotsTxtMiddleware, after parsing the robots.txt file from the original request, does not issue the original request, but instead yields requests for the sitemap URLs found in the robots.txt file.

Also, if a robots.txt parser does not support sitemap iteration, I would expect the current Scrapy implementation of those to be used instead automatically.

I’m just thinking out loud here, though. This might require some prior discussion.

PR #3446, opened 2018-10-03
WIP: Asyncio support for Scrapy

nctl144:
This PR is supposed to continue the work from PR #3362 .

Issue #3148, opened 2018-03-01
GSoC 2018: Intro & questions about adding async/await support 


patiences:
Hi there! 

I've got a couple of questions regarding [this project idea](http://gsoc2018.scrapinghub.com/ideas/#async-await-spiders), but I'll introduce myself a bit first :-)

I'm a 4th year computer science student at the University of British Columbia in Vancouver, Canada (though I'm currently studying abroad in Uppsala, Sweden!). My development experience includes working on integration of software development tools at [Tasktop](https://www.tasktop.com/) (Java/Javascript) and most recently, customisation of data processors at Microsoft (C#/SQL). I've been programming in python for about 2 years. (More information about me [here](http://patiences.github.io)!) 

I've used scrapy both for work and in my personal projects, which is why I'd love to have an opportunity to contribute meaningfully this summer! 

There's already been some discussion around this project on the subreddit as well as in [this issue](https://github.com/scrapy/scrapy/issues/3106) which I have read. Still curious about a few things: 

1. I gathered that the project is not about migrating to asyncio. Is the goal to allow users to leverage async/await syntax in their spiders or custom middleware? What might that look like? Could you, for example, extract all the images from multiple urls concurrently? 

2. Could the project also be about refactoring parts of scrapy itself to use async/await or other asyncio-related frameworks as in [this POC](https://github.com/scrapy/scrapy/pull/1455/files)?

I'll get started on picking up an issue and familiarizing myself with the codebase. If there is anything else that might be useful to know while I'm researching this project please let me know! 

Thank you!

patiences:
Sorry — just realized I never mentioned that what I’m interested in is being a GSoC student!

cathalgarvey:
Hi @patiences - great to meet you here!

You're right to look first at #3106 - particularly @dangra's comment. It sounds like you already have a firm grasp on what the project's all about. :) He may pop in here with some thoughts on your specific queries above, don't be afraid to ask more questions as they come up.

kmike:
Hi! My 2c, some examples of where PEP 492-compatible code can be useful:

1.  Support `async def parse(self, response)`. This can provide a nice way to make several requests in a single callback, in order to aggregate results (e.g. if an item needs to be built from several pages - or, as you said, aggregate images). See https://github.com/scrapy/scrapy/issues/1144 for a discussion.

2. Support `async def` methods in middlewares and extensions. Many of these methods already support Deferreds.

[...]

PR #3972, opened 2019-08-22
[WIP] Base support for asyncio

wRAR:
This is the first PR that is required for any other asyncio-related changes. It assumes no Python 2 support exists so it shouldn't be merged before the 1.8 release and removing Python 2 tests and other support.

Related issues and PRs: #3485, #3148, #3362, #3446, and probably others.

This PR adds installing the asyncio reactor in `scrapy.cmdline` and in tests, a test env that uses the default reactor, and some related tests. It also adds some helper functions which are not currently useful but will be useful in the future changes.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -29,3 +29,8 @@ if six.PY3:
 def chdir(tmpdir):
     """Change to pytest-provided temporary directory"""
     tmpdir.chdir()
+
+
+@pytest.fixture()
+def reactor_pytest(request):
+    request.cls.reactor_pytest = request.config.getoption("--reactor")
--- a/scrapy/cmdline.py
+++ b/scrapy/cmdline.py
@@ -7,9 +7,9 @@ import inspect
 import pkg_resources
 
 import scrapy
-from scrapy.crawler import CrawlerProcess
 from scrapy.commands import ScrapyCommand
 from scrapy.exceptions import UsageError
+from scrapy.utils.asyncio import install_asyncio_reactor, is_asyncio_supported
 from scrapy.utils.misc import walk_modules
 from scrapy.utils.project import inside_project, get_project_settings
 from scrapy.utils.python import garbage_collect
@@ -121,6 +121,10 @@ def execute(argv=None, settings=None):
             settings['EDITOR'] = editor
     check_deprecated_settings(settings)
 
+    # needs to be before _get_commands_dict() as that imports the command modules
+    # which may import twisted.internet.reactor
+    install_asyncio_reactor()
+
     inproject = inside_project()
     cmds = _get_commands_dict(settings, inproject)
     cmdname = _pop_command_name(argv)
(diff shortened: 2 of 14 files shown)


scrapy/cmdline.py
```diff
@@ -121,6 +121,10 @@ def execute(argv=None, settings=None):
             settings['EDITOR'] = editor
     check_deprecated_settings(settings)
 
+    # needs to be before _get_commands_dict() as that imports the command modules
+    # which may import twisted.internet.reactor
+    install_asyncio_reactor()
```
wRAR: Some settings object is available in the lines above.

scrapy/utils/asyncio.py
```diff
@@ -0,0 +1,26 @@
+#coding: utf-8
+
+
+def install_asyncio_reactor():
+    """ Tries to install AsyncioSelectorReactor
```
wRAR: This currently just tries to install the reactor, it can also return if the reactor is the asyncio one and/or set some var. See also is_asyncio_supported below. 

Note that checking the reactor directly, like is_asyncio_supported does, it better than checking a var set or returned from install_asyncio_reactor, as the reactor could be installed outside it.

scrapy/utils/defer.py
```diff
@@ -104,3 +109,21 @@ def iter_errback(iterable, errback, *a, **kw):
...
+    if hasattr(asyncio, 'isfuture'):
+        return asyncio.isfuture(o)
+    return isinstance(o, asyncio.futures.Future)
+
+
+def deferred_from_coro(o):
+    """Converts a coroutine into a Deferred, or returns the object as is if it isn't a coroutine"""
+    if isinstance(o, defer.Deferred):
+        return o
+    if asyncio.iscoroutine(o) or isfuture(o) or inspect.isawaitable(o):
```
wRAR: This means any `async def` function will be converted to an asyncio Task (which then is wrapped into an asyncio Future and a Twisted Deferred) and thus will use an asyncio loop. All `async def` function pass `asyncio.iscoroutine` and I don't know yet if it's possible to distinguish simpler functions from ones really using an asyncio loop and requiring the asyncio reactor. For the simpler functions returning `defer.ensureDeferred(o)` here should be enough.

tests/test_crawler.py
```diff
@@ -245,3 +256,10 @@ def test_crawler_runner_bootstrap_failed_for_several(self):
         yield runner.crawl(NoRequestsSpider)
 
         self.assertEqual(runner.bootstrap_failed, True)
+
+    @defer.inlineCallbacks
+    def test_asyncio_supported(self):
```
wRAR: This really just checks the current reactor via `is_asyncio_supported()`, as `CrawlerRunner` doesn't do anything special here and doesn't run any new processes and the reactor is global, so maybe this test can be dropped.

tests/test_downloadermiddleware.py
```diff
@@ -176,3 +177,31 @@ def process_exception(self, request, exception, spider):
         dfd.addBoth(results.append)
         self.assertIsInstance(results[0], Failure)
         self.assertIsInstance(results[0].value, _InvalidOutput)
+
+
+class MiddlewareUsingDeferreds(ManagerTestCase):
+    """Middlewares using Deferreds should work"""
```
wRAR: This is a strictly unrelated change. It just shows that deferred middleware methods work and that they won;t break when adding coroutine method support.

conftest.py
```diff
@@ -29,3 +29,8 @@ def _py_files(folder):
 def chdir(tmpdir):
     """Change to pytest-provided temporary directory"""
     tmpdir.chdir()
+
+
+@pytest.fixture()
```
wRAR: We'll also need code that will allow us to skip tests when running with the default reactor, looks like there is no simple way to do that.

elacuesta:
#1455 might be relevant as well.

PR #3973, opened 2019-08-22
Provide complete API documentation coverage of scrapy.exporters

Gallaecio:


docs/topics/exporters.rst
```diff
@@ -410,3 +416,8 @@ JsonLinesItemExporter
    this exporter is well suited for serializing large amounts of data.
 
 .. _JSONEncoder: https://docs.python.org/2/library/json.html#json.JSONEncoder
+
+MarshalItemExporter
+-------------------
```
kmike: I wonder if we should support MarshalItemExporter or deprecate it. Any idea, what is it for?

Gallaecio: It’s use case must be the same as `PickleItemExporter`’s. It should perform much better than pickle as long as its drawback are acceptable for your use case.

As to whether or not to deprecate it, I don’t mind either way.

Note that although the class was not documented, the feed export that uses it is: https://docs.scrapy.org/en/latest/topics/feed-exports.html#marshal

kmike:
Thanks for the explanation @Gallaecio!

Issue #2960, opened 2017-10-11
Class Request without the method __eq__ can't be compared directly in the unittest


grammy-jiang:
I met a interesting failure when I did a unittest about the method `process_spider_exception` of the `spider middleware`:

In my project, this method returns a iterable (list) of request objects, which will be sent to the method `process_spider_output`. Now I want to assert this iterable (list) of request objects with `assertEqual` or `assertSequence`, but the test is always failed. Then I debug my code, also with the test case, then I realized that: Class `Request` has no method `__eq__`, so python will compare the id instead, and in my middleware the request is modified with `replace` method, definitely not the previous one, the test will never return `True`.

My solution is to compare `__dict__` of each pair requests in a loop, as mentioned in the following code:

**spider middleware**

```python
class BlockInspector(object):
    def __init__(self, crawler):
        self.crawler = crawler
        self.settings = crawler.settings
        self.signal_manager = crawler.signals
        self.stats = crawler.stats

        self.inspect_block = load_object(
            self.settings.get('BLOCK_INSPECTOR_SPIDERMIDDLEWARE'))
        if self.settings.get('RECYCLE_BLOCK_REQUEST_SPIDERMIDDLEWARE'):
            self.recycle_block_request = load_object(
                self.settings.get('RECYCLE_BLOCK_REQUEST_SPIDERMIDDLEWARE'))
        else:
            self.recycle_block_request = lambda x: x
        self.block_signals = list(map(
            lambda x: load_object(x),
            self.settings.get('BLOCK_SIGNALS_SPIDERMIDDLEWARE')))

    @classmethod
    def from_crawler(cls, crawler):
        obj = cls(crawler)
        return obj

    def process_spider_input(self, response, spider):
        if self.inspect_block(response):
            self.stats.inc_value('block_inspector/block', spider=spider)
            raise BlockException(response)

    def process_spider_exception(self, response, exception, spider):
        if isinstance(exception, BlockException):
            _ = list(map(
                lambda x: self.signal_manager.send_catch_log(
                    x, spider=spider, response=response, exception=exception),
                self.block_signals))

            r = response.request.replace(dont_filter=True)

            return [self.recycle_block_request(r)]
```

**test case**

```python
class TestBlockInspector(TestCase):
    def setUp(self):
        crawler = get_crawler(Spider, settings_dict={
            'BLOCK_INSPECTOR_SPIDERMIDDLEWARE': 'tests.test_spidermiddleware_block_inspector._block_inspector',
[...]

PR #3975, opened 2019-08-26
Implement centralized, highly-configurable request fingerprinting

### LINKED ISSUE #900: Centralized Request fingerprints

### LINKED ISSUE #3597: may be 'accessible'?
in the function [request_fingerprint](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/request.py) ，‘accesible’ may be ‘accessible’ in comments. OCD  XD..

Gallaecio:
Fixes #900, fixes #3420, fixes #3597. Closes #4113 (alternative implementation).

Things to look for during review:

- The proposed approach requires calling `request_fingerprint(request, settings)` every time you want to read the fingerprint of a request. It works, but it’s odd, so I’m more than open to architectural change requests.

-   It is OK for `REQUEST_FINGERPRINT_PROCESSORS` to be a list? Should it be
    something else, such as a dictionary like that of
    ``DOWNLOADER_MIDDLEWARES``?

-   Should we add PyYAML as a dependency to Scrapy and use a PyYAML-based
    serialization function, so that we support binary data (`bytes`, e.g.
    `request.body`) in the fingerprint data dictionary? (otherwise, since JSON
    does not support serializing bytes, `bytes.hex()` or similar must be used
    in processors)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -270,4 +270,5 @@ coverage_ignore_pyobjects = [
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
+    'w3lib': ('https://w3lib.readthedocs.io/en/latest', None),
 }
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -165,6 +165,7 @@ Solving specific problems
    topics/autothrottle
    topics/benchmarking
    topics/jobs
+   topics/request-fingerprinting
 
 :doc:`faq`
     Get answers to most frequently asked questions.
@@ -204,6 +205,9 @@ Solving specific problems
 
 :doc:`topics/jobs`
     Learn how to pause and resume crawls for large spiders.
+
+:doc:`topics/request-fingerprinting`
+    Change how your requests are fingerprinted.
 
 .. _extending-scrapy:
 
--- a/docs/topics/request-fingerprinting.rst
+++ b/docs/topics/request-fingerprinting.rst
@@ -21,6 +21,8 @@ canonical version (:func:`w3lib.url.canonicalize_url`) of
 Continue reading to learn how to change request fingerprinting, for example to
 take into account some headers or compare URLs case-insensitively.
 
+.. _configuring_request_fingerprinting:
+
 Configuring request fingerprinting
 ==================================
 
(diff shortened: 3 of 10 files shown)


elacuesta:
Awesome work. I promise to take a deep look, hopefully sooner than later, but in the meantime, what do you think about adding a fingerprint-based `__eq__` implementation to requests? I think the addition of the fingerprint attribute is a great chance to address #2960.

Regarding `__hash__` (mentioned in the issue) I'm not so sure after reading the following section from the [Data Model page](https://docs.python.org/3/reference/datamodel.html#object.__hash__) (emphasis mine):
> **If a class does not define an __eq__() method it should not define a __hash__() operation either**; if it defines __eq__() but not __hash__(), its instances will not be usable as items in hashable collections. **If a class defines mutable objects and implements an __eq__() method, it should not implement __hash__()**, since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

Requests are supposed to be handled by using the `copy` and `replace` methods, but the `meta` attribute often changes during its lifetime, so I don't know :man_shrugging:

Gallaecio:
Well, I guess if the attribute has been filled in both objects already it can be used for request comparison. However, I’ll get for this to get a proper review before implementing anything else, as we might want to go in a different direction.

The current proposal requires calling `request_fingerprint(request, settings)` every time that you want to read the fingerprint of a request, and requests cannot determine their own fingerprint on its own, which feels a bit off. However, I could not figure out a better approach.

PR #3983, opened 2019-08-28
[WIP] Remove Python 2 support

wRAR:
This removes Python 2 from documentation and tests and removes most of Python 2-only code, including all six.PY2/PY3 conditionals and all or most of `try: ... except ImportError: ...` code. 

This also deprecates `scrapy.utils.gz.read1` and `scrapy.utils.python.to_native_str`.


tests/test_downloader_handlers.py
```diff
@@ -633,18 +629,16 @@ def test_download_gzip_response(self):
...
-            # download_maxsize = 50 is enough for the gzipped response
-            failure = crawler.spider.meta.get('failure')
-            self.assertTrue(failure == None)
-            reason = crawler.spider.meta['close_reason']
-            self.assertTrue(reason, 'finished')
-        else:
-            # See issue https://twistedmatrix.com/trac/ticket/8175
-            raise unittest.SkipTest("xpayload only enabled for PY2")
+        # See issue https://twistedmatrix.com/trac/ticket/8175
+        raise unittest.SkipTest("xpayload only enabled for PY2")
```
kmike: is the code below reachable?

tests/test_downloadermiddleware_robotstxt.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
```
kmike: such imports can be deleted

tests/test_proxy_connect.py
```diff
@@ -1,120 +0,0 @@
...
-    def http_address(self):
-        return 'http://scrapy:scrapy@%s:%d' % self.server.socket.getsockname()
-
-
-def _wrong_credentials(proxy_url):
-    bad_auth_proxy = list(urlsplit(proxy_url))
-    bad_auth_proxy[1] = bad_auth_proxy[1].replace('scrapy:scrapy@', 'wrong:wronger@')
-    return urlunsplit(bad_auth_proxy)
-
-class ProxyConnectTestCase(TestCase):
```
kmike: Are these tests removed because mitmproxy version we're using is Python 2-only? If so, it seems we need to upgrade mitmproxy, but keep tests.

tests/test_downloader_handlers.py
wRAR: I've kept it because https://twistedmatrix.com/trac/ticket/8175 is still open, though it's open for 4 years without activity.

The SkipTest message needs changing though.

tests/test_proxy_connect.py
wRAR: Great question.

We currently have `mitmproxy==0.10.1` and `netlib==0.10.1` in the test requirements, the commit message says "pin mitmproxy 0.10.1 as >0.11 does not work with tests". I also found #2545. I'll look into updating the test for the current mitmproxy.

kmike:
@wRAR could you please check tox/travis config - is seems no tests are executed on CI for this PR?

wRAR:
@kmike there was a run, https://travis-ci.org/scrapy/scrapy/builds/577807912 and I can see it here

kmike:
@wRAR yes, but if you check individual jobs, no tests are executed

Gallaecio:
I wonder if it wouldn’t be better to split this pull request as much as possible across different pull requests, to make reviews easier and conflicts less likely.

For example, CI removals could be on a pull requests of their own, probably the same for the documentation.

wRAR:
@Gallaecio no problem

wRAR:
Obsolete, split into several separate branches.

PR #3987, opened 2019-08-29
LogFormatter: Add the ability to skip log messages

### LINKED ISSUE #3984: Add flag to disable Scraped debug messages
## Summary
Add a flag to settings (e.g. `LOG_SCRAPED_ENABLED`) to be able to disable the logging of each scraped Item on `DEBUG` level.

## Motivation
Crawling with Scrapy on log level `DEBUG` provides a lot of useful context during development. The only problem is the "scraped" message (constructed in `scrapy.logformatter.LogFormatter.scraped`) - it obfuscates the log entirely in the command line and can be really frustrating in general. Adding a flag to disable this log message (with a default value so that the current behaviour doesn't change) would be really useful.

Right now you either put up with it or use `INFO` level instead which makes you lose a lot of useful log entries available in `DEBUG` level.

watsta:
Resolves #3984


elacuesta:
Thanks for the PR, LGTM :+1: 

The change is easy to understand, however could you add some tests? A few [testfixtures.LogCapture.check_present](https://testfixtures.readthedocs.io/en/latest/logging.html#the-check-methods) calls should be enough.

Gallaecio:
Although the source issue explicitly asks for a setting, I think this should be done by extending the [LogFormatter](https://docs.scrapy.org/en/master/topics/logging.html?highlight=Logformatter#custom-log-formats) class, which users can subclass to customize core messages, to allow skipping messages.

watsta:
Hi @Gallaecio,
Thanks for the input. How would you skip a message by subclassing `LogFormatter`? As of now it's only responsible for the format of the message and setting the level, I don't see how it could be used to entirely skip it. Would you rather have it do it the actual logging as well?

Gallaecio:
It’s currently not possible, and that’s what I mean.

I think this pull request should make it possible, instead of adding a setting.

watsta:
> It’s currently not possible, and that’s what I mean.

Thanks, I understood what you meant. My question was rather how would you go about it?
The `LogFormatter` class could log the message instead of the calling `scrapy.core.scraper.Scraper` object, but then whoever defined a `LOG_FORMATTER` entry in their settings file with a different class than `scrapy.logformatter.LogFormatter` would have unexpected behaviour. Also the name `LogFormatter` wouldn't fit anymore.
We could wrap the `LogFormatter` in another class (e.g. `LogProvider`) which actually does the logging and define it the same way as the `LogFormatter`, make it configurable via the settings file. I guess this is viable, though it might be a big change for such a small issue.

What do you think?

elacuesta:
This is interesting. I must admit I wasn't aware of the `LogFormatter` class and `LOG_FORMATTER` setting, since they're not in the _latest_ docs version yet.

Seems like while you cannot currently skip the log entry, you could return a shorter message. Something like:
```python
from scrapy.logformatter import LogFormatter

class IgnoreScrapedFormatter(LogFormatter):
    def scraped(self, item, response, spider):
        return {
            'level': logging.INFO,
            'msg': 'Scraped',
            'args': {},
        }
```

And then set `LOG_FORMATTER` to your custom formatter.

watsta:
I second the proposed solutions, I could go with both but I like the latter more.
Wouldn't it be an idea to implement the `IgnoreScrapedFormatter` as well in `scrapy/logformatter.py`?

Gallaecio:
> How would you skip a message by subclassing `LogFormatter`? As of now it's only responsible for the format of the message and setting the level, I don't see how it could be used to entirely skip it.

I’m not saying it’s the cleanest approach, but what I had in mind is something in the lines of allowing `LogFormatter` methods to return something that indicates that the message is not meant to be logged (e.g. return `None`, or a dictionary containing something like `"dont_log": True`), and modify the code where that return value is used to not log anything in those scenarios.

About providing a ready-to-use implementation of `IgnoreScrapedFormatter`, I would not do that. Maybe a `SilentLogFormatter` that avoids logging for all core messages, but I would rather don’t do that either.

watsta:
@elacuesta can you please advise in which test file I could cover this PR with tests? Couldn't find one which would already invoke the methods with the recent changes.

elacuesta:
I think the most appropriate place would be https://github.com/scrapy/scrapy/blob/1.7.3/tests/test_logformatter.py

elacuesta:
Updated the PR name, since it's no longer about adding a setting.

watsta:
I'm not sure, these tests are only related to the formatting of the log message, not for the actual logging. The changes I made have no effect on the default `LogFormatter` implementation.

elacuesta:
That's right, the new tests might not be like the existing ones that check the output format, however I think `tests/test_logformatter.py` is the right file to put them. I would use `testfixtures.LogCapture` as I mentioned in https://github.com/scrapy/scrapy/pull/3987#issuecomment-526120282.

scrapy/logformatter.py
```diff
@@ -29,6 +29,9 @@ class LogFormatter(object):
     *   ``args`` should be a tuple or dict with the formatting placeholders for ``msg``.
         The final log message is computed as ``msg % args``.
 
+    Users can define their own ``LogFormatter`` class if they want to customise how
+    each action is logged or if they want to omit it entirely. In order to omit
+    logging an action the method must return ``None``.
```
Gallaecio: :lipstick: What about adding an empty line at the end, to make the new text a paragraph of its own instead of part of the existing paragraph?

watsta: Makes sense, updated in e5a5ecfd

watsta:
Thanks @elacuesta for the advice. I looked into it but unfortunately I am not familiar with testing methods which return deferred results and I am also unsure about on which level I should write the tests. The functions to test are called in `ExecutionEngine` and in `Scraper` classes, but setting these classes up for test would be really tedious especially since I am unfamiliar with the codebase as well. Any help would be appreciated, otherwise I don't think I'll be able to provide these tests.

elacuesta:
@watsta That's fine, don't worry. I can write some tests, I'll see if I can do it later today.

elacuesta:
Many thanks @watsta for the contribution and @Gallaecio for the review and merge 🚀

PR #3988, opened 2019-08-29
CallbackKeywordArgumentsContract

### LINKED ISSUE #3985: Contracts can't be written for callbacks that receive cb_kwargs
### Description

Contracts can't be written for callbacks that receive `cb_kwargs`

### Steps to Reproduce
1. Run `scrapy check` with the following spider:
```
class CbKwargsContract(scrapy.Spider):
    name = 'cb_kwargs_contract'

    def start_requests(self):
        yield scrapy.Request("https://httpbin.org/get", cb_kwargs={"arg1": "foo"})

    def parse(self, response, arg1):
        """
        @url https://httpbin.org/get
        """
        self.logger.info(arg1)
```

**Expected behavior:**
Ideally would be possible to add args to the contract (e.g., `@url https://httpbin.org/get foo`)

**Actual behavior:** Fails with a TypeError

### Versions
Scrapy 1.7.3

elacuesta:
Fixes #3985

~This patch works _partially_, in the sense that it prevents the reported `TypeError` (I checked, for instance by adding a simple `print(request.cb_kwargs)` to https://github.com/scrapy/scrapy/blob/1.7.3/scrapy/core/scheduler.py#L90, and the generated request does include a valid `cb_kwargs` attribute). However, adding the `@cb_kwargs` contract causes the whole callback to be excluded from the tests, and I can't for the life of me figure out why.~

Update: Working now, thanks to @victor-torres :rocket: 
 
Remaining tasks:

- [x] Docs
- [x] Tests
- [x] Figure out why it doesn't work

### DIFF
--- a/docs/topics/contracts.rst
+++ b/docs/topics/contracts.rst
@@ -35,11 +35,19 @@ This callback is tested using three built-in contracts:
 
 .. class:: UrlContract
 
-    This contract (``@url``) sets the sample url used when checking other
+    This contract (``@url``) sets the sample URL used when checking other
     contract conditions for this spider. This contract is mandatory. All
     callbacks lacking this contract are ignored when running the checks::
 
     @url url
+
+.. class:: CallbackKeywordArgumentsContract
+
+    This contract (``@cb_kwargs``) sets the :attr:`cb_kwargs <scrapy.http.Request.cb_kwargs>`
+    attribute for the sample request. It must be a valid JSON dictionary.
+    ::
+
+    @cb_kwargs {"arg1": "value1", "arg2": "value2", ...}
 
 .. class:: ReturnsContract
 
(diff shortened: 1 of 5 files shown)


victor-torres:
Despite of that, you might have some problems when trying to pass a Python object through the `cb_kwargs` because it won't be possible to serialize and deserialize data.

elacuesta:
That's true, but it's a current limitation of `Request.meta` as well, and it's solved by enqueuing to memory instead of disk. I don't think we need to worry about that in the context of Scrapy contracts though, since the requests are built from docstings and the `cb_kwargs` attribute in particular is decoded as JSON.

kmike:
Thanks @elacuesta, and thanks @victor-torres and @Gallaecio for reviews!

Issue #374, opened 2013-08-21
LogFormatter should handle formatting of errors


shaneaevans:
In particular, being able to customize downloader and spider errors would be very useful.

PR #3989, opened 2019-08-29
LogFormatter improvements

### LINKED ISSUE #3986: Logformatter for errors
## Summary

Allow users to specify the logging behaviour in case of errors in the item pipeline

## Motivation

Looking at https://github.com/scrapy/scrapy/blob/3adf09bc4f1468f54a703481d6d7ebfe25b6f232/scrapy/core/scraper.py, it is possible to override the logging behaviour for the `DropItem` exception, but there is no equivalent
`self.logformatter.error`.

In our case, sometimes the item can be very large, in which case it being printed completely to the log is problematic -- specifically, on ScrapyingHub the log entry is refused and the exception is not visible.

Modify that file adding a generic error logformatter function.

elacuesta:
Fixes #3986

~I'll add tests shortly~ Done

### DIFF
--- a/tests/test_logformatter.py
+++ b/tests/test_logformatter.py
@@ -23,13 +23,13 @@ class CustomItem(Item):
         return "name: %s" % self['name']
 
 
-class LoggingFormatterTest(unittest.TestCase):
+class LogFormatterTestCase(unittest.TestCase):
 
     def setUp(self):
         self.formatter = LogFormatter()
         self.spider = Spider('default')
 
-    def test_crawled(self):
+    def test_crawled_with_referer(self):
         req = Request("http://www.example.com")
         res = Response("http://www.example.com")
         logkws = self.formatter.crawled(req, res, self.spider)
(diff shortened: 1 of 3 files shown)


elacuesta:
Hold on, I will rename the method to `item_error`, and make the necessary changes so we can close #374 as well.

tests/test_logformatter.py
```diff
@@ -75,26 +85,30 @@ def test_scraped(self):
...
         return {
             'level': kwargs['level'],
             'msg': CRAWLEDMSG,
-            'args': kwargs['args']
+            'args': log_args,
         }
 
 
-class LogformatterSubclassTest(LoggingContribTest):
+class LogformatterSubclassTest(unittest.TestCase):
```
kmike: hey - I think previously this test  was checking that tests pass for LogFormatter subclasses as well; now it only checks test_flags_in_request - is this changed on purpose?

elacuesta: Not on purpose, just a mistake :slightly_smiling_face: 
Updated, overriding three tests which expect different values due to the different LogFormatter. Thanks!

PR #3990, opened 2019-08-30
Add a requestfromcurl command

noviluni:


noviluni:
Tests are failing but as it is just a PoC I will fix them if we decide to proceed (https://github.com/scrapy/scrapy/issues/3991).

PR #3994, opened 2019-08-30
Initial Python 2 removal

wRAR:
This pull request, split out from #3983, removes Python 2 support from documentation and basic stuff like CI.  I've also fixed running tox tests.


.travis.yml
```diff
@@ -7,14 +7,6 @@ branches:
...
 matrix:
   include:
-    - env: TOXENV=py27
-      python: 2.7
-    - env: TOXENV=py27-pinned
-      python: 2.7
-    - env: TOXENV=py27-extra-deps
-      python: 2.7
-    - env: TOXENV=pypy
-      python: 2.7
```
Gallaecio: In this file you also need to update the deployment code below, which is currently tied to the `py27`.

wRAR: @Gallaecio indeed, I was going to ask what to change there

Gallaecio: @kmike You probably have though about this a few times.

One thing I can think of is renaming `py37` to `py`, keep `py` always for the latest version (i.e. when Python 3.8 is release, `py` becomes 3.8 and `py37` is created), and use `py` in the deployment code.

docs/intro/install.rst
```diff
@@ -102,10 +102,8 @@ just like any other Python package.
 (See :ref:`platform-specific guides <intro-install-platform-notes>`
 below for non-Python dependencies that you may need to install beforehand).
 
-Python virtualenvs can be created to use Python 2 by default, or Python 3 by default.
-
-* If you want to install scrapy with Python 3, install scrapy within a Python 3 virtualenv.
-* And if you want to install scrapy with Python 2, install scrapy within a Python 2 virtualenv.
+Python virtualenvs can be created to use Python 2 by default, or Python 3 by default. As Scrapy
+only supports Python 3, make sure you created a Python 3 virtualenv.
```
Gallaecio: I was going to suggest updating the documentation links to point to https://docs.python.org/3/library/venv.html. However, I would like to propose a complete rewrite of this section once these changes are merged (#4001), so no need to worry too much about this part in this pull request.

tox.ini
```diff
@@ -4,18 +4,15 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27
```
Gallaecio: Instead of removing this entry, I think we need to update it. I believe this is what is executed when you run `tox` without manually specifying `-e <list of environments>`.

wRAR:
Remade as #4091

Issue #3976, opened 2019-08-26
`ItemLoader` fields initialized from `item` are reprocessed


alexander-matsievsky:
### Description

#3804 introduced a bug where `ItemLoader` fields are reprocessed.
Related #3897.

### Steps to Reproduce

```python
from pprint import pprint

from scrapy import Field, Item
from scrapy.loader import ItemLoader
from scrapy.loader.processors import TakeFirst


class X(Item):
    x = Field(output_processor=TakeFirst())


loader = ItemLoader(X())
loader.add_value("x", ["value1", "value2"])
x = loader.load_item()
pprint(x)
# {'x': 'value1'}

pprint(ItemLoader(x).load_item())
# {'x': 'v'}
```

**Expected behavior:** `ItemLoader` initialized from the `x` item does not reprocess its fields and  loads `{'x': 'value1'}`.

**Actual behavior:** `ItemLoader` initialized from the `x` item reprocesses its fields and  loads `{'x': 'v'}`.

### Versions

```
Scrapy       : 1.7.3
lxml         : 4.4.1.0
libxml2      : 2.9.9
cssselect    : 1.1.0
parsel       : 1.5.2
w3lib        : 1.21.0
Twisted      : 19.7.0
Python       : 3.6.5 (default, May  3 2018, 10:08:28) - [GCC 5.4.0 20160609]
pyOpenSSL    : 19.0.0 (OpenSSL 1.1.1c  28 May 2019)
cryptography : 2.7
Platform     : Linux-4.4.0-127-generic-x86_64-with-LinuxMint-18.1-serena
```

### Additional context

Here's the behavior of the previous version:

```
Scrapy       : 1.6.0
lxml         : 4.4.0.0
libxml2      : 2.9.9
cssselect    : 1.0.3
parsel       : 1.5.1
w3lib        : 1.20.0
Twisted      : 19.7.0
Python       : 3.6.5 (default, May  3 2018, 10:08:28) - [GCC 5.4.0 20160609]
pyOpenSSL    : 19.0.0 (OpenSSL 1.1.1c  28 May 2019)
cryptography : 2.7
Platform     : Linux-4.4.0-127-generic-x86_64-with-LinuxMint-18.1-serena
```

```python
# {'x': 'value1'}
# {'x': 'value1'}
```

alexander-matsievsky:
Related #3897.

ava7:
[...]

Issue #3897, opened 2019-07-22
ItemLoaders can break if instantiated with pre-populated items


elacuesta:
Before I start, I know item loaders have been a big source of discussion for a long time; I'm only opening this issue because the latest release breaks some of our spiders.

In one of our projects, our [Autounit](https://github.com/scrapinghub/scrapy-autounit) tests fail under 1.7.1 due to some item loaders which are created from partially populated items. I suspect the relevant change is #3819 (which BTW I think inadvertently closes #3046).
Personally I think a better approach here would be something _closer_ to the solution proposed in #3149, although not exactly the same.

Consider the following:

```python
In [1]: import scrapy

In [2]: scrapy.__version__
Out[2]: '1.6.0'

In [3]: from scrapy.loader import ItemLoader
   ...: lo = ItemLoader(item={'key': 'value'})
   ...: lo.add_value('key', 'other value')
   ...: print(lo.load_item())
{'key': ['other value']}
```

```python
In [1]: import scrapy

In [2]: scrapy.__version__
Out[2]: '1.7.1'

In [3]: from scrapy.loader import ItemLoader
   ...: lo = ItemLoader(item={'key': 'value'})
   ...: lo.add_value('key', 'other value')
   ...: print(lo.load_item())
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-6aa64a41edb1> in <module>
      1 from scrapy.loader import ItemLoader
      2 lo = ItemLoader(item={'key': 'value'})
----> 3 lo.add_value('key', 'other value')
      4 print(lo.load_item())

~/venv-temporal/lib/python3.6/site-packages/scrapy/loader/__init__.py in add_value(self, field_name, value, *processors, **kw)
     77                 self._add_value(k, v)
     78         else:
---> 79             self._add_value(field_name, value)
     80
     81     def replace_value(self, field_name, value, *processors, **kw):

~/venv-temporal/lib/python3.6/site-packages/scrapy/loader/__init__.py in _add_value(self, field_name, value)
     93         processed_value = self._process_input_value(field_name, value)
     94         if processed_value:
---> 95             self._values[field_name] += arg_to_iter(processed_value)
     96
     97     def _replace_value(self, field_name, value):

TypeError: must be str, not list
```
I'm not directly opening a PR because I think this needs discussion. What if we changed
```python
for field_name, value in item.items():
    self._values[field_name] = self._process_input_value(field_name, value)
```

to

```python
[...]

Issue #3804, opened 2019-06-02
Items missing


azharf:
Hi,
 when working with the itemLoader()  to populate items from a created list in python.  Here I have attached a working of the suspected bug. 

This appears to be present when the loader.get_output_value method has been executed. Once that executes then, the populated list initially created would have one of the items missing. This appears to be a malfunction bug within the software presented.  If we only have loader.load_item, the values are presented, however loder.get_output_value() fails to display the value/s in the data set as tested. 
<img width="1440" alt="scrapybug" src="https://user-images.githubusercontent.com/35243679/58755752-0d574380-849f-11e9-8ae7-7fd1f0c6dbd4.png">

 loader.get_output_value causes the items value to be missing from School:

BurnzZ:
Confirmed this bug, but only when `ItemLoader` is being instantiated with an `item`. Here's an example:

```
import scrapy

from scrapy.http import TextResponse
from scrapy.loader import ItemLoader

class TestItem(scrapy.Item):
	title = scrapy.Field()

class TestItemLoader(ItemLoader):
	default_item_class = TestItem

body = '<html><title>This is a title</title></html>'
response = TextResponse('https://test.com', body=body)

loader_from_response = TestItemLoader(response=response)
loader_from_response.add_css('title', 'title::text')
loader_from_response.load_item()  # {'title': [u'This is a title']}
loader_from_response.get_output_value('title')  # [u'This is a title']
loader_from_response.load_item()  # {'title': [u'This is a title']}

# The loading above is the most common approach in parsing the website
# contents. The bug occurs below when `ItemLoader` is being instantiated
# with an 'item'. 

input_item = {'title': 'Title from dict-like item'}
loader_from_item = ItemLoader(item=input_item)
loader_from_item.load_item()  # {'title': 'Title from dict-like item'}
loader_from_item.get_output_value('title')  # []
loader_from_item.load_item()  # {'title': []}
```

sortafreel:
Working on a pull request to fix it :)

kmike:
Fixed by #3819.

ava7:
Hey, fellows, I am afraid that this bugfix introduces an another problem...
Let me try to explain. If we take @BurnzZ 's  example https://github.com/scrapy/scrapy/issues/3804#issuecomment-498683757 and modify it just a little bit to pass a dict via `item` in TestItemLoader's constructor:
```TestItemLoader(response=response, item=TestItem(loader.load_item()))```
we end up with the following error:

[...]

PR #3998, opened 2019-09-01
Add reprocessing tests

sortafreel:
I tried to somehow understand when the item is processed the second-third-n time, so ignore the processor, but there's no clear way to find out how field processors were populated - initiated in a Class object, or taken from the ready item (so need to reprocess). 

To be sure, I added a lot of new tests simulating the issue (https://github.com/scrapy/scrapy/issues/3976) logic and tested all of them through scrapy 1.6.0. to get the right results. And it looks like, the working solution is to check if the item is initiated from dict. Because, if item initiated from actual Item class object - it has own "meta", and no need to reprocess it, while dict == raw data and it needs to be processed through default processors to be sure it's valid.

If I'm too naive and missing something - let me know what tests to add to simulate an error, because all current test variations are passing.


scrapy/loader/__init__.py
```diff
@@ -35,8 +35,9 @@ def __init__(self, item=None, selector=None, response=None, parent=None, **conte
         self._local_values = defaultdict(list)
         # Preprocess values if item built from dict
         # Values need to be added to item._values if added them from dict (not with add_values)
-        for field_name, value in item.items():
-            self._values[field_name] = self._process_input_value(field_name, value)
+        if type(item) == dict:
```
elacuesta: [`isinstance`](https://docs.python.org/3/library/functions.html#isinstance) would be a better option here, since it also works with subclasses.

elacuesta:
```suggestion
        if isinstance(item, dict):
```
Just trying the _suggestion_ feature 😄

sortafreel:
@elacuesta @Gallaecio If an item is initiated from a dict - all single values are converted to lists with a single element. Added more tests to cover both https://github.com/scrapy/scrapy/issues/3897 and https://github.com/scrapy/scrapy/issues/3976.

elacuesta:
I think dicts are well covered here, however I believe values from `Item` subclasses should also be added to the loader. Maybe modifying `_add_value` to take an optional `preprocess` flag, or call `arg_to_iter` directly?

scrapy/loader/__init__.py
```diff
@@ -35,8 +35,10 @@ def __init__(self, item=None, selector=None, response=None, parent=None, **conte
         self._local_values = defaultdict(list)
         # Preprocess values if item built from dict
         # Values need to be added to item._values if added them from dict (not with add_values)
-        for field_name, value in item.items():
-            self._values[field_name] = self._process_input_value(field_name, value)
+        if isinstance(item, dict):
+            for field_name, value in item.items():
+                # Convert all single values to lists because of following output processors
+                self._add_value(field_name, value)
```
Gallaecio: I’m not familiar enough with the internals of `ItemLoader`, so this may be a stupid question: Why do we want `_process_input_value` applied to dictionaries? Shouldn’t we be assigning the values *as is* (but as lists), as it *seems* we do with item classes?

sortafreel: Not sure I understand the question correctly. If the question is "why to preprocess dicts" - it was the solution to save missing items (https://github.com/scrapy/scrapy/issues/3804), because without preprocessing scrapy dropped values. If I didn't understand the question - please, add more details :slightly_smiling_face:

sortafreel:
> I think dicts are well covered here, however I believe values from `Item` subclasses should also be added to the loader. Maybe modifying `_add_value` to take an optional `preprocess` flag, or call `arg_to_iter` directly from `__init__`?

I assume we had no problems with processing Item objects, only problems were with initiating Items from dicts. Can you add more details, please?

elacuesta:
> I assume we had no problems with processing Item objects, only problems were with initiating Items from dicts. Can you add more details, please?

With the current check in the `ItemLoader` initializer (`if isinstance(item, dict):`), values from the `item` variable will be added to the loader only if `item` is a dictionary, not if it's a `scrapy.item.Item` object.

sortafreel:
> > I assume we had no problems with processing Item objects, only problems were with initiating Items from dicts. Can you add more details, please?
> 
> With the current check in the `ItemLoader` initializer (`if isinstance(item, dict):`), values from the `item` variable will be added to the loader only if `item` is a dictionary, not if it's a `scrapy.item.Item` object.

Can you provide a code example, please? I mean, whole `if isinstance(item, dict):` block wasn't there in scrapy 1.6, I added it only to handle items initiated from dicts. If items were processed correctly in  scrapy 1.6 they must be processed correctly now too.

elacuesta:
Sure, here's a code sample showing that values from the initial item were not preserved correctly in 1.6.0:

```python
In [1]: import scrapy

In [2]: scrapy.__version__
Out[2]: '1.6.0'

In [3]: from scrapy.item import Item, Field 
   ...: from scrapy.loader import ItemLoader 
   ...:  
   ...: class MyItem(Item): 
   ...:     name = Field() 
   ...:  
   ...: loader_item = ItemLoader(item=MyItem(name="foo")) 
   ...: print("From Item, before add_value:", loader_item.load_item()) 
   ...: loader_item.add_value("name", "bar") 
   ...: print("From Item, after add_value:", loader_item.load_item()) 
   ...:  
   ...: loader_dict = ItemLoader(item=dict(name="foo")) 
   ...: print("From dict, before add_value:", loader_dict.load_item()) 
   ...: loader_dict.add_value("name", "bar") 
   ...: print("From dict, after add_value:", loader_dict.load_item())

From Item, before add_value: {'name': 'foo'}
From Item, after add_value: {'name': ['bar']}
From dict, before add_value: {'name': 'foo'}
From dict, after add_value: {'name': ['bar']}
```

This case (adding values after creating the loader with an initial item) should be tested.

scrapy/loader/__init__.py
Gallaecio: I mean that, with the current implementation, input processors are applied to input dictionary values, but not to item objects. `_add_value` calls `_process_input_value`, which passes dictionary values through input processors.

I wonder if it wouldn’t be better, to get values from dictionaries into `_values`, to use something similar to `_add_values` that does _not_ apply input processing.

I may be getting things wrong, though, and even if this is an issue it would be an issue that already exists before this patch, so maybe it’s better to leave this topic for later, for a different issue or pull request.

Gallaecio:
Superseded by #4036

PR #3999, opened 2019-09-03
Provide complete API documentation coverage of scrapy.item

Gallaecio:
As part of these changes I was going to document `scrapy.item.DictItem`. However, trying to understand its purpose, I came to the conclusion that it is a leftover that no longer has a purpose, so in this set of changes I’m deprecating it.

I believe it was introduced in decbdc534644384c18a603dc35650b044d0b5a0a so that `DjangoItem` could have everything that `Item` has but using a different metaclass. However, [`Item` seems to be currently used in scrapy-djangoitem instead](https://github.com/scrapy-plugins/scrapy-djangoitem/blob/bfe9622c2e618bcd117ce248fb3a2d463f24bce1/scrapy_djangoitem/__init__.py).

I am not 100% sure about `scrapy.item.DictItem` no longer having a purpose, though. So please, look into it for yourselves before approving these changes.

### DIFF
--- a/scrapy/item.py
+++ b/scrapy/item.py
@@ -4,13 +4,15 @@ Scrapy Item
 See documentation in docs/topics/item.rst
 """
 
-from abc import ABCMeta
-from pprint import pformat
-from copy import deepcopy
 import collections
+from abc import ABCMeta
+from copy import deepcopy
+from pprint import pformat
+from warnings import warn
 
 import six
 
+from scrapy.utils.deprecate import ScrapyDeprecationWarning
 from scrapy.utils.trackref import object_ref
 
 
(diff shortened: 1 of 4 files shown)


elacuesta:
Sounds good to me, for the sake of simplicity and having less classes that do basically the same. However I think the [`Item` class definition](https://github.com/scrapy/scrapy/blob/1.7.3/scrapy/item.py#L115) should be modified not to inherit from `DictItem` (Github won't let me comment in the appropriate line because it's not modified in this PR).
Perhaps the names could be swapped, so the `Item` class doesn't lose all the methods it currently inherits from `DictItem`.

Gallaecio:
> I think the [`Item` class definition](https://github.com/scrapy/scrapy/blob/1.7.3/scrapy/item.py#L115) should be modified not to inherit from `DictItem`

I’m not sure we need to change the inheritance tree at this point. Although unlikely, it’s possible that changing the inheritance breaks code silently (e.g. `isinstance(this, DictItem)` would stop evaluating to `True` for `Item`).

If instead we keep the inheritance tree as is, and simply remove `DictItem` in a future version (moving all its code to `Item` then), such code will fail as well, but loudly with an error upon importing `DictItem`.

elacuesta:
Forget what I said, I got confused. `cls is DictItem` is `True` only when the object being created is in fact a `DictItem`, not when creating from a subclass like `Item`.

scrapy/item.py
```diff
@@ -56,6 +74,12 @@ class DictItem(MutableMapping, BaseItem):
 
     fields = {}
 
+    def __new__(cls, *args, **kwargs):
+        if cls is DictItem:
+            warn('scrapy.item.DictItem is deprecated, please use '
+                 'scrapy.item.Item instead', ScrapyDeprecationWarning)
```
kmike: how does a warning message look like, does it point to user code? I wonder if a non-default stacklevel can make the message more clear.

Gallaecio:
Should I write a test to cover the code changes here?

kmike:
@Gallaecio yes, could you please add a test for it?

Gallaecio:
Tests added, and implementation improved to also warn about inheriting from `DictItem`.

kmike:

Issue #781, opened 2014-07-02
CrawlSpider improvements


dangra:
A ticket to kickoff the discussion on CrawlSpider enhancements (if any)

ideas:
- Simplify rule definitions by using an implicit linkextractor instanciated from LxmlLinkExtractor.
- ... what else?

redapple:
I would also add something related to https://github.com/scrapy/scrapy/issues/929, that is passing the response to `process_request` in `_requests_to_follow`, so that one can tweak the generated requests with some context. (If there's another way currently than overriding `_request_to_follow`, it'd be happy for it to be documented :)

Oh, and also adding `errback` to rules could be handy at times: see http://stackoverflow.com/a/35870000/2572383

guillaumedsde:
> I would also add something related to #929, that is passing the response to `process_request` in `_requests_to_follow`, so that one can tweak the generated requests with some context. (If there's another way currently than overriding `_request_to_follow`, it'd be happy for it to be documented :)
> 
> Oh, and also adding `errback` to rules could be handy at times: see http://stackoverflow.com/a/35870000/2572383

is this a planned feature?

elacuesta:
> is this a planned feature?

@guillaumedsde Please see #3682 (included in version `1.7.1`) and #4000 (in progress).

elacuesta:
Seems to me like everything has been addressed :ok_hand:

PR #4000, opened 2019-09-04
Add errback parameter to scrapy.spiders.crawl.Rule

elacuesta:
As suggested in https://github.com/scrapy/scrapy/issues/781#issuecomment-247020413.


scrapy/spiders/crawl.py
```diff
@@ -74,48 +81,61 @@ def parse_start_url(self, response):
...
+            for request_or_item in iterate_spider_output(results):
+                yield request_or_item
+
     def _compile_rules(self):
-        self._rules = [copy.copy(r) for r in self.rules]
-        for rule in self._rules:
-            rule._compile(self)
+        self._rules = []
+        for rule in self.rules:
+            self._rules.append(copy.copy(rule))
```
elacuesta: I left the copying as it was, but I'm wondering if it's really necessary :thinking:

Gallaecio: I’m not sure either.

elacuesta: Seems like the relevant commit is 6d5a130b96411f97a21e0810866f1e8f675510b3. Since it's not a `deepcopy` I'm going to remove it to simplify things.

docs/topics/spiders.rst
```diff
@@ -374,7 +374,7 @@ CrawlSpider
 Crawling rules
 ~~~~~~~~~~~~~~
 
-.. class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None)
+.. class:: Rule(link_extractor, callback=None, cb_kwargs=None, follow=None, process_links=None, process_request=None, errback=None)
```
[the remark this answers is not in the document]
elacuesta: Nice, thanks.

scrapy/spiders/crawl.py
elacuesta: Or not :smile: 
I guess this is important when compiling rules that are shared between spiders (which they shouldn't IMHO, but it's probably better not to take the chances of breaking existing spiders).

scrapy/spiders/crawl.py
```diff
@@ -74,48 +81,61 @@ def parse_start_url(self, response):
...
 
         if follow and self._follow_links:
             for request_or_item in self._requests_to_follow(response):
                 yield request_or_item
 
+    def _handle_failure(self, failure, errback):
+        if errback:
+            results = errback(failure) or ()
+            if hasattr(failure.value, 'response'):
+                results = self.process_results(failure.value.response, results)
```
elacuesta: For the record, I'm thinking a good change here would be to call `process_results` with the failure object, and document it as `process_results(response_or_failure, results)`, but since the method also exists on other specific spider I think it might be better keep the scope of this PR as is and make a new one for those changes.

scrapy/spiders/crawl.py
```diff
@@ -74,48 +81,61 @@ def parse_start_url(self, response):
...
-        for n, rule in enumerate(self._rules):
-            links = [lnk for lnk in rule.link_extractor.extract_links(response)
-                     if lnk not in seen]
-            if links and rule.process_links:
-                links = rule.process_links(links)
-            for link in links:
+        for rule_index, rule in enumerate(self._rules):
+            links = (lnk for lnk in rule.link_extractor.extract_links(response)
+                     if lnk not in seen)
+            for link in rule.process_links(links):
```
kmike: I think it is better to keep links as lists, because it was documented  that Rule.process_links input is a list (see  process_links docs).

elacuesta: Makes sense. I'll revert this change, and will open a follow-up PR including this, [this comment](https://github.com/scrapy/scrapy/pull/4000#discussion_r321303103) and some other documentations improvements. Thanks!

elacuesta: @kmike Updated, please check again. Thanks!

PR #4003, opened 2019-09-05
[py3] Remove six.moves and __future__ imports

elacuesta:
Related to #3994, #3983, #4091

Another chapter in the _Remove Python 2 support_ series. In this case, I'm removing `six.moves` and `__future__` imports.

~This is definitely going to have conflicts with the above PRs, I will fix them when the time comes.~

~Python 2 tests are expected to fail.~

@wRAR what do you think about pushing our branches to the main repo instead of our forks, to make merging and rebasing easier?


Gallaecio:
> @wRAR what do you think about pushing our branches to the main repo instead of our forks, to make merging and rebasing easier?

Is there much difference? I usually do simply `git fetch upstream && git rebase upstream/master` from my branch to get it updated from the `master` branch of the main repository (configured as the `upstream` remote).

elacuesta:
Right, but if I want to base my branch on to of Andrey's it's not that simple, I'd need to either set up his fork as a remote or do something like the procedure explained [here](https://docs.scrapy.org/en/latest/contributing.html#submitting-patches).
Perhaps I'm being too lazy, and this would only save me a couple of commands :-)

Gallaecio:
You do have a point.

If we don’t release 1.8, we might want to consider creating a 2.x branch, even. Although that should be a last resort.

wRAR:
@elacuesta I've pushed `remove-py2-base` (from #3994) to this repo. I was going to split the other commits into separate PRs but don't want to do that before this one is merged. I also have commits removing all six.moves and removing all __future__ imports locally but you published them before me :)

elacuesta:
Awesome, many thanks @wRAR and @Gallaecio. I've rebased on top of the `remove-py2-base` branch and there were no conflicts :raised_hands:
Some unrelated changes appeared in the diff but that's expected, that will be fixed once the other PRs get merged.

scrapy/_monkeypatches.py
```diff
@@ -1,5 +1,6 @@
+import copyreg
+
 import six
```
Gallaecio: I believe we should be able to get rid of `six` imports all over, and of the dependency itself (at least as a direct dependency).

scrapy/cmdline.py
```diff
@@ -1,19 +1,18 @@
...
-from scrapy.crawler import CrawlerProcess
 from scrapy.commands import ScrapyCommand
+from scrapy.crawler import CrawlerProcess
 from scrapy.exceptions import UsageError
+from scrapy.settings.deprecated import check_deprecated_settings
 from scrapy.utils.misc import walk_modules
 from scrapy.utils.project import inside_project, get_project_settings
 from scrapy.utils.python import garbage_collect
-from scrapy.settings.deprecated import check_deprecated_settings
+import scrapy
```
Gallaecio: I believe convention is for `from … import …` lines to always go after `import …` lines. It affects a few other modified files as well.

elacuesta: You're right, it doesn't look so good. It was the result of a blind alphabetical order. Thanks for noticing.

scrapy/_monkeypatches.py
elacuesta: Yup, I thought about that but came to the same conclusion, there are still some `six.string_types` and `six.text_type`, for instance.

Gallaecio:
The error looks legit. I guess we need to remove Protego from the list of extra requirements, since it is now a dependency of Scrapy itself, being the default robots.txt parser.

elacuesta:
Closing, see #4121

Issue #3969, opened 2019-08-19
Make Protego the default robots.txt parser


anubhavp28:
[Protego](https://github.com/scrapy/protego) is more compliant robots.txt parser than the current default [RobotFileParser](https://docs.python.org/3/library/urllib.robotparser.html). It supports wildcard matching, length based directive ordering, sitemaps, crawl-delay and request-rate directive. It is also compatible with [Google's robots.txt parser](https://github.com/google/robotstxt). AFAIK, It has bigger test suite than any other open source robots.txt parser. The latest version of Protego is on PyPI and can be installed with `pip install protego`.

It also faster than other Python based candidates. I performed a benchmark using popular robots.txt parsers -

I crawled multiple domains (~1,100 domains) and downloaded links to pages as the crawler encounter them. I downloaded 111, 824 links in total. Next I made each robots.txt parser - parse and answer query (I made parsers answer each query 5 times) in an order similar to how they would need to in a broad crawl. Here are the results :

* [Protego](https://github.com/scrapy/protego) (Python based) :
25th percentile : 0.002419 seconds
50th percentile : 0.006798 seconds
75th percentile : 0.014307 seconds
100th percentile : 2.546978 seconds
Total Time : 19.545984 seconds

* [RobotFileParser](https://docs.python.org/3/library/urllib.robotparser.html) (default in Scrapy) :
25th percentile : 0.002188 seconds
50th percentile : 0.005350 secondsstyle
75th percentile : 0.010492 seconds
100th percentile : 1.805923 seconds
Total Time : 13.799954 seconds

* [Robotexclusionrulesparser](http://nikitathespider.com/python/rerp/) (Python based) :
25th percentile : 0.001288 seconds
50th percentile : 0.005222 seconds
75th percentile : 0.014640 seconds
100th percentile : 52.706880 seconds
Total Time : 76.460496 seconds

* [Reppy Parser](https://github.com/seomoz/reppy) (C++ based with Python interface) :
25th percentile : 0.000210 seconds
50th percentile : 0.000492 seconds
75th percentile : 0.000997 seconds
100th percentile : 0.129440 seconds
Total Time: 1.405558  seconds
 
The code and data used for benchmarking is [here](https://github.com/anubhavp28/robotstxt-parser-benchmark).

Gallaecio:
cc @kmike @dangra @whalebot-helmsman

kmike:
Fixed by https://github.com/scrapy/scrapy/pull/4006.

[...]

PR #4006, opened 2019-09-09
Use protego as a default robots.txt parser

whalebot-helmsman:
Implementation of proposal https://github.com/scrapy/scrapy/issues/3969

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1089,7 +1089,7 @@ Based on `Reppy <https://github.com/seomoz/reppy/>`_:
 
 * uses the length based rule
 
-Native implementation provides better speed than Protego.
+Native implementation, provides better speed than Protego.
 
 In order to use this parser:
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -513,7 +513,7 @@ defines the methods described below.
 
     .. method:: open_spider(spider)
 
-      This method gets called after a spider has been opened for crawling. It handles 
+      This method gets called after a spider has been opened for crawling. It handles
       the :signal:`open_spider <spider_opened>` signal.
 
       :param spider: the spider which has been opened
@@ -521,8 +521,8 @@ defines the methods described below.
 
     .. method:: close_spider(spider)
 
-      This method gets called after a spider has been closed. It handles 
-      the :signal:`close_spider <spider_closed>` signal. 
+      This method gets called after a spider has been closed. It handles
+      the :signal:`close_spider <spider_closed>` signal.
 
       :param spider: the spider which has been closed
       :type spider: :class:`~scrapy.spiders.Spider` object
@@ -1020,10 +1020,10 @@ the request will be ignored by this middleware even if
 RobotFileParser
 ~~~~~~~~~~~~~~~
 
-`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
-Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's 
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is
+Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_. It lacks
-support for wildcard matching. Scrapy uses this parser by default.
+support for wildcard matching.
 
 In order to use this parser, set:
 
@@ -1074,12 +1074,11 @@ Protego parser
 
 `Protego <https://github.com/scrapy/protego>`_ is a pure-Python robots.txt_ parser.
 The parser is fully compliant with `Google's Robots.txt Specification
-<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard 
+<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard
 matching, and uses the length based rule similar to `Reppy <https://github.com/seomoz/reppy/>`_.
+Scrapy uses this parser by default.
 
 In order to use this parser:
-
-* Install `Protego <https://github.com/scrapy/protego>`_ by running ``pip install protego``
 
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ProtegoRobotParser``
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1021,7 +1021,7 @@ RobotFileParser
 ~~~~~~~~~~~~~~~
 
 `RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is
-Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
+Python's built-in robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_. It lacks
 support for wildcard matching.
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1000,10 +1000,10 @@ RobotsTxtMiddleware
 
     Scrapy ships with support for the following robots.txt_ parsers:
 
-    * :ref:`RobotFileParser <python-robotfileparser>` (default)
+    * :ref:`Protego <protego-parser>` (default)
+    * :ref:`RobotFileParser <python-robotfileparser>`
     * :ref:`Reppy <reppy-parser>`
     * :ref:`Robotexclusionrulesparser <rerp-parser>`
-    * :ref:`Protego <protego-parser>`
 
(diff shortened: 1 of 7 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -1020,10 +1020,10 @@ the request will be ignored by this middleware even if
 RobotFileParser
 ~~~~~~~~~~~~~~~
 
-`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
-Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's 
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is
+Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
```
elacuesta:
```suggestion
Python's built-in robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
```
I know this line was like this before, but since it's being changed we might as well fix it.

requirements-py2.txt
```diff
@@ -15,3 +15,4 @@ service_identity>=16.0.0
 six>=1.10.0
 Twisted>=16.0.0
 zope.interface>=4.1.3
+protego
```
elacuesta:
```suggestion
protego>=0.1.15
```
Nitpick: Could you also please move `protego` to the section at the top of the file?)

scrapy/settings/default_settings.py
```diff
@@ -246,7 +246,7 @@
 RETRY_PRIORITY_ADJUST = -1
 
 ROBOTSTXT_OBEY = False
-ROBOTSTXT_PARSER = 'scrapy.robotstxt.PythonRobotParser'
+ROBOTSTXT_PARSER = 'scrapy.robotstxt.ProtegoRobotParser'
```
kmike: I think docs needs to be changed as well, see https://github.com/scrapy/scrapy/blob/master/docs/topics/settings.rst#robotstxt_parser.

docs/topics/downloader-middleware.rst
```diff
@@ -1074,13 +1074,12 @@ Protego parser
 
 `Protego <https://github.com/scrapy/protego>`_ is a pure-Python robots.txt_ parser.
 The parser is fully compliant with `Google's Robots.txt Specification
-<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard 
+<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard
 matching, and uses the length based rule similar to `Reppy <https://github.com/seomoz/reppy/>`_.
+Scrapy uses this parser by default.
```
kmike: It could make sense to change the order of the introduction of different parsers, so that the default one goes first.

kmike: If we move it to the top, description of this parser shouldn't mention reppy, it would be confusing otherwise.

docs/topics/downloader-middleware.rst
```diff
@@ -1020,10 +1020,10 @@ the request will be ignored by this middleware even if
 RobotFileParser
 ~~~~~~~~~~~~~~~
 
-`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is 
-Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's 
+`RobotFileParser <https://docs.python.org/3.7/library/urllib.robotparser.html>`_ is
+Python's built-in robots.txt_ parser. The parser is fully compliant with `Martijn Koster's
 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_. It lacks
-support for wildcard matching. Scrapy uses this parser by default.
+support for wildcard matching.
```
kmike: Probably makes sense to add a note on why this parser is useful:
* faster than protego
* compatibility with previous Scrapy versions

docs/topics/downloader-middleware.rst
```diff
@@ -1074,13 +1074,12 @@ Protego parser
...
 `Protego <https://github.com/scrapy/protego>`_ is a pure-Python robots.txt_ parser.
 The parser is fully compliant with `Google's Robots.txt Specification
-<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard 
+<https://developers.google.com/search/reference/robots_txt>`_ hence supports wildcard
 matching, and uses the length based rule similar to `Reppy <https://github.com/seomoz/reppy/>`_.
+Scrapy uses this parser by default.
 
 In order to use this parser:
 
-* Install `Protego <https://github.com/scrapy/protego>`_ by running ``pip install protego``
```
kmike: currently you still need to install protego manually; this issue is hidden on CI because protego is included in all requirements.txt files. To fix it, protego needs to be added to setup.py install_requires.

docs/topics/downloader-middleware.rst
kmike: RobotFileParser is still mentioned as default in docs above, where all parsers are listed. Could you please re-read the whole docs about robots.txt, and check that they make sense, assuming protego is a default parser?

docs/topics/downloader-middleware.rst
```diff
@@ -1015,50 +1015,78 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
...
-Python's inbuilt robots.txt_ parser. The parser is fully compliant with `Martijn Koster's 
-1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_. It lacks
-support for wildcard matching. Scrapy uses this parser by default.
-
-In order to use this parser, set:
+Based on `RobotFileParser
+<https://docs.python.org/3.7/library/urllib.robotparser.html>`_:
 
-* :setting:`ROBOTSTXT_PARSER` to ``scrapy.robotstxt.PythonRobotParser``
+* is Python's built-in robots.txt_ parser.
```
Gallaecio: :lipstick: List entries of parsers are inconsistent in whether or not they end in a period. It would be great if they all ended the same way. Since they are not capitalized, I would remove the period from the ones that currently have one.

docs/topics/downloader-middleware.rst
```diff
@@ -1015,50 +1015,78 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
...
 
+Parsers vary in several aspects:
+
+* Language of implementation
+
+* Supported specification
+
+* Support for wildcard matching
+
+* Usage of length based rule: in particular for ``Allow`` and
```
Gallaecio: :lipstick: I wonder if we should link to some external resource here for users who want to know more about this, e.g. https://developers.google.com/search/reference/robots_txt#order-of-precedence-for-group-member-lines

docs/topics/downloader-middleware.rst
```diff
@@ -1015,50 +1015,78 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
...
+  <https://github.com/seomoz/rep-cpp>`_.
+
+* is compliant with `Martijn Koster's 1996 draft specification
+  <http://www.robotstxt.org/norobots-rfc.txt>`_.
+
+* supports wildcard matching
+
+* uses the length based rule
+
+Native implementation provides better speed than Protego.
```
Gallaecio: Should we indicate, in average, how the performance of different parsers compares to the default parser?

docs/topics/downloader-middleware.rst
```diff
@@ -1015,50 +1015,81 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
...
-~~~~~~~~~~~~~~~~~~~~~~~~~
+* lacks support for wildcard matching
 
-`Robotexclusionrulesparser <http://nikitathespider.com/python/rerp/>`_ is fully compliant
-with `Martijn Koster's 1996 draft specification <http://www.robotstxt.org/norobots-rfc.txt>`_,
-with support for wildcard matching.
+* doesn't use the length based rule
 
-In order to use this parser:
+It is faster than Protego and backward-compatible with versions of Scrapy before 1.8.0 .
```

```suggestion
It is faster than Protego and backward-compatible with versions of Scrapy before 1.8.0.
```

docs/topics/downloader-middleware.rst
```diff
@@ -1015,50 +1015,81 @@ If :attr:`Request.meta <scrapy.http.Request.meta>` has
...
+  <https://github.com/seomoz/rep-cpp>`_
+
+* is compliant with `Martijn Koster's 1996 draft specification
+  <http://www.robotstxt.org/norobots-rfc.txt>`_
+
+* supports wildcard matching
+
+* uses the length based rule
+
+Native implementation provides better speed than Protego.
```

```suggestion
Native implementation, provides better speed than Protego.
```

kmike:
Hey! The PR looks fine to me. 

I'm on fence on it, but maybe it can be nice to review protego API a bit before making it default, to avoid churn in future - e.g. in Python's default parser (in upcoming 3.8) there is .site_maps() method, while in protego there is .sitemaps property. On the other hand, it doesn't affect Scrapy directly, so maybe it is not needed.

whalebot-helmsman:
>  review protego API a bit before making it default, to avoid churn in future - e.g. in Python's default parser (in upcoming 3.8) there is .site_maps() method, while in protego there is .sitemaps property. 

I see no problem here. We don't directly use parsers' API in scrapy. Every parser should provide proper adapter. We tied to adapter's API, not parser's API.

kmike:
Ok, let's do it!

PR #4008, opened 2019-09-11
Request: remove restriction about errback without callback

### LINKED ISSUE #3586: Request doesn't set callback to self.parse automatically when errback is set
Hi, I'm facing an `AssertionError` when set `errback`  and not `callback` in `Request`, is that expected? :

```
...
    def start_requests(self):
        yield Request(
            self.start_url,
            # callback=self.parse,
            errback=self.parse_error,
            meta={
                'dont_redirect': True,
            },
        )
...
```
trigger this : 
```
2019-01-16 15:47:22 [scrapy.core.engine] ERROR: Error while obtaining start requests
  File "/home/joaquin/Repos/example/env/lib/python3.6/site-packages/scrapy/core/engine.py", line 127, in _next_request
  File "/home/joaquin/Repos/example/example/spiders/br/rj/example.py", line 128, in start_requests
    'dont_redirect': True,
  File "/home/joaquin/Repos/example/env/lib/python3.6/site-packages/scrapy/http/request/__init__.py", line 34, in __init__
    assert callback or not errback, "Cannot use errback without a callback"
AssertionError: Cannot use errback without a callback
2019-01-16 15:47:22 [scrapy.core.engine] INFO: Closing spider (finished)
2019-01-16 15:47:22 [scrapy.core.engine] INFO: Spider closed (finished)
```

elacuesta:
Fixes #3586

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -32,7 +32,6 @@ class Request(object_ref):
             raise TypeError('callback must be a callable, got %s' % type(callback).__name__)
         if errback is not None and not callable(errback):
             raise TypeError('errback must be a callable, got %s' % type(errback).__name__)
-        assert callback or not errback, "Cannot use errback without a callback"
         self.callback = callback
         self.errback = errback
 
--- a/tests/test_http_request.py
+++ b/tests/test_http_request.py
@@ -246,25 +246,41 @@ class RequestTest(unittest.TestCase):
         self.assertRaises(AttributeError, setattr, r, 'url', 'http://example2.com')
         self.assertRaises(AttributeError, setattr, r, 'body', 'xxx')
 
-    def test_callback_is_callable(self):
+    def test_callback_and_errback(self):
         def a_function():
             pass
-        r = self.request_class('http://example.com')
-        self.assertIsNone(r.callback)
-        r = self.request_class('http://example.com', a_function)
-        self.assertIs(r.callback, a_function)
-        with self.assertRaises(TypeError):
-            self.request_class('http://example.com', 'a_function')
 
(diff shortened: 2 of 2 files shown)


docs/topics/request-response.rst
```diff
@@ -119,8 +118,9 @@ Request objects
        raised while processing the request. This includes pages that failed
        with 404 HTTP errors and such. It receives a `Twisted Failure`_ instance
        as first parameter.
-       For more information,
-       see :ref:`topics-request-response-ref-errbacks` below.
+       Please note that setting an ``errback`` without an explicit ``callback``
+       will result in an error.
```
kmike: Why is that? It looks more like a bug to me, which we should fix, not document :)

elacuesta: I assumed there was a reason for it, since it's a deliberate `assert` and not an unhandled exception, but it is a rather [old line](https://github.com/scrapy/scrapy/commit/73305b1eb333d6cf2c42b0cb45e4ed75e9c523a6#diff-adcd0debfae0b9c2b1f34a01d2aafc53R30) and perhaps there is no such reason anymore.  Let me think about alternative ways to approach this.

elacuesta:
Tests are green without the restriction (the one failure is because of #4014, it can be ignored IMHO).
@pablohoffman since you authored the commit that introduced the assertion, do you have any objections to remove it?

pablohoffman:
no objections @elacuesta!

Gallaecio:
> Tests are green without the restriction

Unless we already have it, though, this might be a great opportunity to add a test for this, that specifically verifies that a callback can be omitted even if an errback is provided.

kmike:
Thanks @elacuesta!

PR #4010, opened 2019-09-12
Base support for asyncio

wRAR:
This is https://github.com/scrapy/scrapy/pull/3972, rebased against the non-py2 branch. I can copy my comments from there if needed.


wRAR:
I've added the `ASYNCIO_SUPPORT` option as discussed recently. I've wrote https://github.com/scrapy/scrapy/wiki/Asyncio-use-cases that helped me to understand possible real-world use cases and what should we do with them.

scrapy/utils/asyncio.py
```diff
@@ -0,0 +1,23 @@
+def install_asyncio_reactor():
+    """ Tries to install AsyncioSelectorReactor
+    """
+    try:
```
elacuesta: Could you refactor these `except: pass` ocurrences to use [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress)?

scrapy/utils/asyncio.py
```diff
@@ -0,0 +1,23 @@
...
+        import asyncio
+        from twisted.internet import asyncioreactor
+    except ImportError:
+        pass
```
Gallaecio: If asyncio is explicitly requested but not available, shouldn’t we raise an Exception (maybe let the `ImportError` raise) or at the very least log an error?

Or is this better done from the calling code instead?

scrapy/utils/asyncio.py
```diff
@@ -0,0 +1,23 @@
...
+    else:
+        from twisted.internet.error import ReactorAlreadyInstalledError
+        try:
+            asyncioreactor.install(asyncio.get_event_loop())
+        except ReactorAlreadyInstalledError:
+            pass
```
Gallaecio: Similar here. Shoudn’t we call `is_asyncio_reactor_installed` here and raise or log something if the installed reactor does not provide asyncio support? (assuming we can check for this, I’m guessing an alternative asyncio reactor that does not inherit from the one we use may be a problem)

scrapy/utils/defer.py
```diff
@@ -1,11 +1,15 @@
 """
 Helper functions for dealing with Twisted deferreds
 """
+import asyncio
+import asyncio.futures
```
Gallaecio: :lipstick: I don’t think this line is needed.

tox.ini
```diff
@@ -106,3 +106,15 @@ deps =
...
+basepython = python3.5
+deps = {[testenv]deps}
+commands =
+    py.test --cov=scrapy --cov-report= --reactor=asyncio {posargs:scrapy tests}
+
+[testenv:py38-asyncio]
+basepython = python3.8
+deps = {[testenv]deps}
+commands =
+    py.test --cov=scrapy --cov-report= --reactor=asyncio {posargs:scrapy tests}
```
Gallaecio: :lipstick: I _think_ you can omit the `deps` line in these two environments.

You could also define `[asyncio]` with this `commands` and then use `commands = {[asyncio]commands}` in these two environments. Although I’m not sure it’s worth it for only 2 entries, so up to you.

scrapy/utils/asyncio.py
wRAR: @elacuesta one of these cases uses `else` specifically to handle the situation when there is no `ImportError`, can that still be refactored?

docs/topics/settings.rst
```diff
@@ -160,6 +160,31 @@ to any particular component. In that case the module of that component will be
 shown, typically an extension, middleware or pipeline. It also means that the
 component must be enabled in order for the setting to have any effect.
 
+.. setting:: ASYNCIO_SUPPORT
+
+ASYNCIO_SUPPORT
```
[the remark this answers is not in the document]
wRAR: no problem

wRAR:
I've found that I once again forgot to write the ASYNCIO_ENABLED handling for `CrawlerProcess`. So making a note here.

Looks like we'll need a new test that runs a new process with `CrawlerProcess` inside.

scrapy/utils/asyncio.py
Gallaecio: I don’t think so, although using a `return` instead of `pass` in this specific case may be _slightly_ more readable.

scrapy/utils/asyncio.py
wRAR: It can be discussed.

`install_asyncio_reactor` is called in various "early" places if ASYNCIO_ENABLED is on. If it fails, we can abort the spider or we can log the warning. Currently we log it, with ERROR, in `log_scrapy_info()` when the spider starts, which is not the place where it's installed but later.

wRAR: We also compare `is_asyncio_reactor_installed()` and `ASYNCIO_ENABLED` when coroutines are used, and raise an exception (which I think will break the spider).

scrapy/utils/asyncio.py
elacuesta: Right, I missed that. I agree, `return` instead of `pass` in the first occurrence does sound better to me.

scrapy/utils/defer.py
```diff
@@ -113,3 +117,27 @@ def iter_errback(iterable, errback, *a, **kw):
             break
         except Exception:
             errback(failure.Failure(), *a, **kw)
+
+
+def isfuture(o):
```
elacuesta: I hope this isn't too much bikeshedding, but I wonder if this function shouldn't be underscored, if it's only needed for py35

wRAR:
@Gallaecio can you please explain how to investigate the coverage?

Gallaecio:
@wRAR I’m not sure at the moment, but I’ve created #4230 to deal with that separately, I don’t think coverage data should block these changes.

Gallaecio:
@wRAR The first thing I noticed is that `deferred_from_coro` is defined but never used, which would explain why it is not included in the coverage data.

Gallaecio:
I bet it’s far from trivial, but would it be possible to write tests for [those import errors](https://codecov.io/gh/scrapy/scrapy/pull/4010/src/scrapy/utils/asyncio.py)?

wRAR:
@Gallaecio hmmm. 

The requirements have changed since that code was first written, `asyncio` is now available in all supported Python versions and `twisted.internet.asyncioreactor.AsyncioSelectorReactor` is available in our minimal required Twisted version (17.9.0). I think these checks can be removed.

I also wonder if we should test asyncio in the "pinned" env, maybe use py35-asyncio one for that. I remember some features being only available in newer Twisted.

tox.ini
```diff
@@ -96,3 +96,17 @@ changedir = {[docs]changedir}
 deps = {[docs]deps}
 commands =
     sphinx-build -W -b linkcheck . {envtmpdir}/linkcheck
+
+[asyncio]
+commands =
+    py.test --cov=scrapy --cov-report= --reactor=asyncio {posargs:scrapy tests}
```
kmike: is there a reson posargs in this command are not the same as in the main testenv?

kmike:
Looks good. Thanks @wRAR for the feature, and @Gallaecio and @elacuesta for the reviews!

CarterPape:
@wRAR This feature has been on my mind since I started using this package. I am so happy that you and the Scrapinghub team have initiated the support for `asyncio`.

PR #4016, opened 2019-09-13
Crawling rules: make link extractors optional

elacuesta:
As proposed in #781


kmike:
Thanks @elacuesta, makes sense.

Issue #1988, opened 2016-05-12
"Content-Encoding" header gets stripped from response headers


mborho:
SEE https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpcompression.py#L36

IMHO the "Content-Encoding" header should get preserved, since the spider probably wants to see all the original response headers.

kmike:
It makes sense to me, +1. Likely the header is deleted to signal that response body is decoded; we should provide another way to do that (add a flag to response.flags?) if we keep the header.

The change is backwards incompatible, so I think we should make it optional, disable it by default, but enable in settings.py generated by `scrapy startproject`.

foromer4:
I started to implement this as suggested, but I'm unsure about the settings:
in the httpcompression class, How can I access the golbal settings? and how can I change the value  there for test purposes?

Trim21:
"Content-Encoding" header is still removed from `response.header`.
Is there any progress about this issue up to now?

VorBoto:
Hello, I am looking to try and help and this seemed the least intimidating task.
I want to see if I'm getting my bearing on what is looking to be done correct.

There is a desire to have a bool that can be adjusted directly in default_settings.py, or in a different file, to prevent HttpCompresionMiddleware.process_responce() from stripping away the content-encoding header value. 

Could this be done by passing process_response an instance of settings then use settings.getbool() like is done in the HttpErrorMiddleware.__init__(self, settings)? Also like what some  downloadermiddleware like in ajaxcrawl and redirect that have an __int__  with setting passed to them? 
Or would a class variable like redirect.BaseRedirectMiddleware has 'enable_setting' be perfered?

Also where are downloadermiddleares called or imported I have been looking but not finding yet.

Gallaecio:
> Could this be done by passing process_response an instance of settings then use settings.getbool() like is done in the HttpErrorMiddleware.**init**(self, settings)? Also like what some downloadermiddleware like in ajaxcrawl and redirect that have an **int** with setting passed to them?
> Or would a class variable like redirect.BaseRedirectMiddleware has 'enable_setting' be perfered?

I believe the way to go is to store `crawler.settings` into `self.settings` in `__init__`, and then use `self.settings.getbool()` from `process_response`.

[...]

PR #4017, opened 2019-09-14
Add HEADERS_KEEP to settings and to HttpCompressionMiddleware (#1988)

VorBoto:
Added setting and variable to HttpCompressionMiddleware.
Not 100% if thats good to go but figured I can give it a shot and learn from any mess-ups.


VorBoto:
Not working. Can ignore.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -18,6 +18,9 @@
 class HttpCompressionMiddleware(object):
     """This middleware allows compressed (gzip, deflate) traffic to be
     sent/received from web sites"""
+
+    HEADERS_KEEP = settings.getbool('HEADERS_KEEP')
```
elacuesta: The reason this is failing is because the `settings` variable is not defined here, hence the `NameError` exception.

What would be the motivation behind this change? Perhaps there is an alternative solution.

VorBoto: Thank you for taking the time to respond on here. Yeah I had a feeling it wouldn't work but was just throwing it out there to kinda break the ice for me. But I was trying to work on issue [#1988](https://github.com/scrapy/scrapy/issues/1988) where they thought about keeping the encoding-header instead of chopping it off right at the end of the process_repsonse function. 

@Gallaecio commented on the few questions I possed on that thread and think I'm going to try and model it after the ajaxcrawl.py in the downloadermiddlewares that as an __init__ that's passed setting and set a variable something like self.keep_headers.

I was hesitant to try that cause I thought that would create a bigger problem on top of the namespace error. Since I thought it would have to change the arguments that are passed to HttpCompressionMiddleware where ever else it is invoked but they said it was created depending on the user's settings but also that that might be thinking too far ahead is how I took it. In addition to having trouble seeing where crawler and other object were coming from to be used as arguments.

It has been some time since I've worked on classes and projects as spread out as this.

elacuesta:
Makes sense. I see you opened #4021, let's continue the discussion there.

PR #4018, opened 2019-09-14
Fixed a typo

jp26jp:


Gallaecio:
Did you notice you are sending a patch against a very old branch? DjangoItems were removed from recent versions of Scrapy into their own project: https://github.com/scrapy-plugins/scrapy-djangoitem

Issue #187, opened 2012-11-07
request.replace not working in process_response, return response never gets called


adaschevici:
Hello. I have searched for this but i haven't found any satisfactory answer:

This is my code. I found some replies that say this should work.
But when i run it as part of my spider i get an infinite loop that apparently get stuck on the line that prints rt.url .
I suspect that there may be a problem with the priority of middlewares but i am not sure.

``` python
from auctionzip.settings import USER_AGENT_LIST
import random
from scrapy import log
class RandomUserAgentMiddleware(object):
   def process_request(self, request, spider):
        ua  = random.choice(USER_AGENT_LIST)
        tmp = "http://www.freelancer.com"
        rt = request.replace(url = tmp)
        #if ua:
        #    request.headers.setdefault('User-Agent', ua)
        #    request.headers.setdefault('Referrer', 'http://www.auctionzip.com')
        print rt.url
        return rt
        #log.msg('>>>> UA %s'%request.headers)

    def process_response(self, request, response, spider):
        print "in response handler"
        return response
```

This is the order of my middlewares:

``` python
DOWNLOADER_MIDDLEWARES = {
    'auctionzip.middlewares.random_user_agent.RandomUserAgentMiddleware': 100,
    #'wines_crawler.middlewares.tor_anonymizer.TorMiddleWare' : 401,
    #'wines_crawler.middlewares.random_proxy.ProxyMiddleware' : 401,
    #'scrapy.contrib.downloadermiddleware.useragent.UserAgentMiddleware': None,
    'scrapy.contrib.downloadermiddleware.cookies.CookiesMiddleware': 402,
    'scrapy.contrib.downloadermiddleware.downloadtimeout.DownloadTimeoutMiddleware' : 600,
    #'wines_crawler.middlewares.retry_change.RetryTORChangeProxyMiddleWare' : 600,
    #'wines_crawler.middlewares.retry_change.RetryChangeProxyMiddleware' : 600,
}
```

I tried this with scrapy 0.16 and 0.17.

PR #4021, opened 2019-09-16
Adding keep-encoding option to HttpCompressionMiddleware  issue #1988

VorBoto:
Model this after the ajaxcrawl.py middle ware to pass setting to an __init__ and store the setting in an instance variable.


scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -18,6 +18,12 @@
 class HttpCompressionMiddleware(object):
     """This middleware allows compressed (gzip, deflate) traffic to be
     sent/received from web sites"""
+
+    def __int__(self, crawler):
+        #setting to decide to keep or discard encoding header
```
elacuesta: I think these comments are unnecessary, the setting should be documented instead.

VorBoto:
@elacuesta @Gallaecio 
Looking at the tests from the travis-ci and the tests of both httpcompression and ajaxcrawl AjaxCrawlMiddleware calls a class function from_crawler to instantiate its instance. I'm assuming that's what the decorator `@classmethod` does with `return cls(crawler.settings)` . But then for httpcompression its just `HttpCompressionMiddleware()`, in the tests, would I need to go through and add the settings or a from_crawler class method to HttpCompressionMiddleware and load it in like Ajax is in the test. Of would simply adding going back to pass `HttpCompressionMiddleware(self, settings)` and adding the settings passed to it in each test?

Am I allowed to touch tests?

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -45,8 +51,9 @@ def process_response(self, request, response, spider):
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
                 response = response.replace(**kwargs)
-                if not content_encoding:
-                    del response.headers['Content-Encoding']
+                if not self.keep_encoding_header:
```
elacuesta: The problem I see with this solution is that responses could end up with a "Content-Encoding" header that does not fully match the actual encoding.
I think the ideal solution for this would be #187, but while that is not available, `Response.flags` is probably the less ugly way, as mentioned in https://github.com/scrapy/scrapy/issues/1988#issuecomment-219405456

elacuesta:
@VorBoto Sorry, I started my review before seeing your latest comment.
Indeed, you are not only allowed to touch tests, I'd say you are encouraged to do so. In this particular case, new tests are probably needed as well.

scrapy/downloadermiddlewares/httpcompression.py
VorBoto: No worries I put it in just as you were commenting. I need to slow down a little.

So first with the from_crawler classmethod I overlooked, have it be `return cls(crawler.settings)` similar to Ajax. Then if I want to continue that path instead of passing HttpCompressionMiddleware anything it should get called with the from_crawler method. Which should be accessible since HttpComp test imports what ajax does, besides `from scrapy.utils.test import get_crawler`. 

Then to fix the possible mix match of the header value. Should I double check that the bytes passed are the same encoding as the value specified? That would need to be before its decoded. Or are you thinking to have the header value that is passed with the returned response object has the header encoding value be that of what it was decoded to?

Then response flags. Looks like they are passes when getting a Response object or through the replace method that takes keyword pairs. So then reworking the logic at the tail end to simply replace the encoding header value when it's being kept cause I don't see a way to tack on an additional header value.

VorBoto:
So @elacuesta this is what I've done now
I removed the comments. 
Placed `return cls(crawler.settings)` in the from_crawler method.
Placed `self.mw = HttpCompressionMiddleware.from_crawler(crawler)` in its test.

I'm just not 100 on how to handle the Response.flags. I see the response gets stripped down in the processing. But the headers do get placed in the object that will ultimately be used to replace the kwargs. Would that be the place to adjust the 'Content-Encoding' value, once I know what to adjust it to?

VorBoto:
There is a problem with cls not being an object that takes parameters.
I will need to read up on `@classmethod` decorators and cls before I attempt this again I think.

VorBoto:
I'm sorry everyone for cloging this up with a bunch of trash. I am going to do things more properly and use tox before I just toss things in the system.

PR #4023, opened 2019-09-17
Fix windows file descriptor leak

### LINKED ISSUE #3391: File descriptor leak in FeedExporter
Test suite under windows fails to remove a file at https://github.com/scrapy/scrapy/pull/3315/files#diff-0f004a6e06393cc5e29012b83956eed2L245 because FeedExporter instances keeps an open file descriptor created for the test case.

We need to find the leak and close the file descriptor properly on close_spider signal.

Gallaecio:
Fixes #3391

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -242,7 +242,7 @@ class FeedExporter(object):
     def close_spider(self, spider):
         slot = self.slot
         if not slot.itemcount and not self.store_empty:
-            return
+            return defer.maybeDeferred(slot.storage.store, slot.file)
         if self._exporting:
             slot.exporter.finish_exporting()
             self._exporting = False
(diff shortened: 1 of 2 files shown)


scrapy/extensions/feedexport.py
```diff
@@ -242,7 +242,7 @@ def open_spider(self, spider):
     def close_spider(self, spider):
         slot = self.slot
         if not slot.itemcount and not self.store_empty:
-            return
+            return defer.maybeDeferred(slot.storage.store, slot.file)
```
kmike: Could you please explain how this fix works?

Gallaecio: In `FeedExporter.open_spider` we are opening the output file, regardless of the value of `self.store_empty`. `slot.storage.store` is responsible for closing that file. See `scrapy.extensions.feedexport.FileFeedStorage.store`.
The code here, however, was skipping that step when there was no item to store and storing an empty file was disabled.

kmike: Aha, this makes sense. Could you please add a small comment here though? Now I see that this is a copy-paste of the code below (`d = defer.maybeDeferred(slot.storage.store, slot.file)`). But it is not obvious, from these few lines themselves, that slot.storage.store closes a file.

PR #4025, opened 2019-09-18
Place Content-encoding header into response's flags is desired #1988

VorBoto:
Threw up some trash not completely knowing what I was doing but got help from some other contributors on this project and hope this will be acceptable.

I added the ability to keep the 'Content-Encoding' header value in a response as a member of the 'flag' attribute of a response.


scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -18,11 +18,15 @@
 class HttpCompressionMiddleware(object):
     """This middleware allows compressed (gzip, deflate) traffic to be
     sent/received from web sites"""
+
+    def __init__(self, settings):
+        self.keep_encoding_header = settings.getbool('HTTPCOMPRESSION_HEADERS_KEEP')
```
Gallaecio: What about renaming `HTTPCOMPRESSION_HEADERS_KEEP` to `COMPRESSION_KEEP_ENCODING_HEADER`?

Although `HTTPCOMPRESSION` seems in line with other middlewares, this middleware already uses `COMPRESSION` as prefix (there’s a `COMPRESSION_ENABLED` setting to enable the middleware).

And since this setting only affects the `Accept-Encoding` header, I think it makes sense to make the setting name explicit about that.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,11 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                if self.keep_encoding_header:
+                    kwargs['flags'] = [encoding]
```
Gallaecio: What does this do?

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,11 @@ def process_response(self, request, response, spider):
...
                 response = response.replace(**kwargs)
                 if not content_encoding:
                     del response.headers['Content-Encoding']
```
Gallaecio: Shouldn’t this *not* be done if `self.keep_encoding_header` is `True`?

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -27,8 +27,11 @@
 class HttpCompressionTest(TestCase):
 
     def setUp(self):
-        self.spider = Spider('foo')
-        self.mw = HttpCompressionMiddleware()
+        #need to find way to access settings here
+        crawler = get_crawler(Spider,{'HTTPCOMPRESSION_HEADERS_KEEP':True,
+                                      'COMPRESSION_ENABLED':True})
+        self.spider = crawler._create_spider('foo')
+        self.mw = HttpCompressionMiddleware.from_crawler(crawler)
```
Gallaecio: Maybe this should be moved out of the setUp, extracted into a separate function that returns a spider and a middleware, and received an optional parameter that determines whether `HTTPCOMPRESSION_HEADERS_KEEP` is enabled or not. This function can then be called at the beginning of every test with the right parameters for each test.

scrapy/downloadermiddlewares/httpcompression.py
VorBoto: Sounds good. I wasn't sure what to name it so tried to keep it close to where it would be making an effect.

scrapy/downloadermiddlewares/httpcompression.py
VorBoto: This adds an entry to the kwargs dict for 'flags' and places the encoding that the response was in as a list as its value.

scrapy/downloadermiddlewares/httpcompression.py
VorBoto: Ar first I had it just skip that but there are a few tests that assert 'Content-Encoding' is None and also there was what @kmike asked in the [issue comments](https://github.com/scrapy/scrapy/issues/1988#issuecomment-219405456) about placing it in the response.flags. So with those two things decided to allow the header to be removed but have the encoding placed in the response.flags to signal it was compressed before processing. ( I believe that's what flags are used for if I'm misunderstanding their use as signals about the what has been done or what is wanted with a response?)

tests/test_downloadermiddleware_httpcompression.py
VorBoto: I can understand that. I just modeled this off of the ajax downloadermiddleware test did in its setup function.

VorBoto:
@Gallaecio I did some of the changes you recommended but wasn't sure which tests should get what parameters so this will fail the tests but wanted to see if I pulled the right parts out for a separate function.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -18,11 +18,16 @@
 class HttpCompressionMiddleware(object):
     """This middleware allows compressed (gzip, deflate) traffic to be
     sent/received from web sites"""
+
+    def __init__(self, settings):
+        self.keep_encoding_header = settings.getbool(
+            'COMPRESSION_KEEP_ENCODING_HEADERS')
```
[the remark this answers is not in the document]
Gallaecio: Well, I guess it can be understood as referring to the header in all requests.

scrapy/downloadermiddlewares/httpcompression.py
VorBoto: OOh okay I see what you mean. That make sense, I wasn't aware enough of how the headers where used later on and the particulars of what flags where for. Especially them being checked before hand, can you think of a good example of that in the code, I'm having trouble?

VorBoto:
I have the if statement for putting in `decoded`, should I put it in the `if ssubclass(respcls,TextResponse):` body or on the same level, does it matter much?

Gallaecio:
> I have the if statement for putting in `decoded`, should I put it in the `if ssubclass(respcls,TextResponse):` body or on the same level, does it matter much?

I think it’s on the same level, but I am not 100%, I’m not familiar with this part of the code base yet. In any case, tests (existing and new) should tell whether or not you made the right choice.

VorBoto:
Not sure where to remove or change things to test with different values testing with or without headers removed.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,17 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                if self.keep_encoding_header:
+                    if isinstance(kwargs['flags'], list):
```
Gallaecio: `kwargs` is declared in line 46 above. `kwargs['flags']` will never be defined at this point, so this should be throwing a `KeyError` exception.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,17 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                if self.keep_encoding_header:
+                    if isinstance(kwargs.get('flags'), list):
```
Gallaecio: When I [said impossible](https://github.com/scrapy/scrapy/pull/4025#discussion_r328028801), I meant impossible. I think you may be mixing up `kwargs` and `response` here. Please, make sure you understand what is going on in this function.

VorBoto: Its meant to process a response, ie decode and encoded input. If the request used to get the response was just for the 'HEAD' it finishes and the response that is the head is just returned. (Not 100% on what a 'HEAD' request is I'm assuming it like a header and contains general info on the actual site that a 'GET' could be sent to.

Then if it is a Response decode the response body into human readable(?) format. then create a another response, `respcls`, from the passed in response. Then create the kwargs dict with the fileds cls, class right?, as respcls and body taht is the decoded human readable format. Then if the created response, `respcls`, is a response that has a body of test add an additional entry for encoding that is None.

Then I added the if keep encoding header if where it checks if there is a kwargs entry for flags, an attribute that a typical Response object has correct or at least accumulates as it goes through the various filters and processing? Depending on, from the looks there wont be, if there are `flags` values add the flag 'decoded' 

Then to finalize it the original response has the values in kwargs replaced.......I'm not pulling the flags from the original response and then simple over riding them by creating the entry of flags with a single value erasing the accumulated flags from before aren't i?

After that clean up the headers but removing 'Content-Encoding' if that is desired. And ultimately return the processed and readable response object.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,18 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                kwargs['flags'] = response.flags
+                if self.keep_encoding_header:
+                    if isinstance(kwargs.get('flags'), list):
```
Gallaecio: Now that you set `kwargs['flags'] = response.flags` beforehand, this type check is not necessary, it will always be a list. See https://github.com/scrapy/scrapy/blob/master/scrapy/http/response/__init__.py#L25

VorBoto: Dang it. Thank you for helping guide me through this by the way too!

Gallaecio:
> Not sure where to remove or change things to test with different values testing with or without headers removed.

I think the way to go now is to remove the `setUp` function, and have each test call your new text function at the beginning:

```
spider, mw = self.create_spider_mw()
```

Then you can add new tests covering your changes that also call that function, but use parameter values different from the defaults when needed for your test cases.

VorBoto:
Okay I'll have to work on that a little bit at a time through out the week.
Have an opportunity at work I have to do some additional research and work that at the moment has a little higher priority.
Just wanted to get that out there so if I'm not active for a day or three its not like I just ghosted this just getting something else done before coming back.

VorBoto:
Isn't setup function an inherited function from unittest that needs to be filled out for it to work?
I'm not to familiar with unittest I'm just looking it up and seeing what is going on.
Is there possibly someone who is more versed in testing phase of the  project we could ask for more insight?(Not saying you're not capable or don't know. I am just guessing there might be someone who while it may not be their primary responsibility knows the testing functionality and base better than the rest. Or am I assuming to much?)

Gallaecio:
The setup method only needs to be defined to to set up stuff that is common to all tests in the class.

This was the right way to define the spider and middleware, since the same spider and middleware was used in every test.

Now that we want to add tests that use a different spider and middleware, you can remove that function and instead create a spider and a middleware in each test method.

Alternatively, and to minimize changes, you could leave existing tests as they are and leave the setup method as is, and simply use a different spider and middleware in your new tests that require one. It will actually make for a cleaner patch, so it is probably a good idea.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,13 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                if self.keep_encoding_header:
+                    kwargs['flags'] = response.flags
+                    kwargs['flags'].append(b'decoded')
```
Gallaecio: :lipstick: In a single line: ``kwargs['flags'] = response.flags + [b'decoded']``

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,13 @@ def process_response(self, request, response, spider):
...
                 response = response.replace(**kwargs)
-                if not content_encoding:
-                    del response.headers['Content-Encoding']
-
+                if not self.keep_encoding_header:
```
Gallaecio: :lipstick:  I’m not sure if it makes the code more readable or less readable, but you could do ``if not self.keep_encoding_header and not content_encoding:``

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -26,10 +26,23 @@
...
+                              'COMPRESSION_KEEP_ENCODING_HEADERS':
+                               compression_header})
+        spider = crawler._create_spider('foo')
+        mw = HttpCompressionMiddleware.from_crawler(crawler)
+        return spider, mw
+                            
     def setUp(self):
-        self.spider = Spider('foo')
-        self.mw = HttpCompressionMiddleware()
+        #need to find way to access settings here
```
Gallaecio: This comment should go.

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -26,10 +26,23 @@
 
 class HttpCompressionTest(TestCase):
 
+    def create_spider_mw(self
+                         ,compression_enabled=True
```
Gallaecio: :lipstick: Having the commas at the beginning feels off to me :slightly_frowning_face:

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -26,10 +26,23 @@
...
+                         ,compression_header=False):
+        crawler = get_crawler(Spider,
+                              {'COMPRESSION_ENABLED':
+                               compression_enabled,
+                              'COMPRESSION_KEEP_ENCODING_HEADERS':
```
Gallaecio:
```suggestion
                               'COMPRESSION_KEEP_ENCODING_HEADERS':
```

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -249,3 +262,34 @@ def test_process_response_head_request_no_decode_required(self):
         newresponse = self.mw.process_response(request, response, self.spider)
         self.assertIs(newresponse, response)
         self.assertEqual(response.body, b'')
+
+    def test_process_response_gzip_keep_headers(self):
+        self.spider, self.mw = self.create_spider_mw(
```
Gallaecio:
```suggestion
        spider, mw = self.create_spider_mw(
```

Don’t overwrite the existing class variables. Simply declare new ones local to the test function, and use those instead of the `self.*` equivalent in the rest of the test method. Same for the other new test method.

VorBoto: Would it better or indifferent if I assigned them names such as `test_spider` and `test_mw` to not have them get confused with the self variables?

tests/test_downloadermiddleware_httpcompression.py
VorBoto: I was trying to hard to stick to the 80 width limit here.

tests/test_downloadermiddleware_httpcompression.py
VorBoto: Gotcha, wasn't sure what was preffered and I like to have operators at the beginning of a new line instead of trailing on the previous lines so I carried that over to the comas.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -44,10 +48,11 @@ def process_response(self, request, response, spider):
                     # force recalculating the encoding until we make sure the
                     # responsetypes guessing is reliable
                     kwargs['encoding'] = None
+                if self.keep_encoding_header:
+                    kwargs['flags'] = response.flags +[b'decoded']
```
Gallaecio: :lipstick: 

```suggestion
                    kwargs['flags'] = response.flags + [b'decoded']
```

Gallaecio:
It has been a while, and I need to review my review :sweat_smile: 

This needs also a documentation change for the new setting, as well as taking into account backward compatibility for users subclassing the middleware.

PR #4030, opened 2019-09-20
Add option to send logs directly to a Graylog server using graypy

viniciusd-avila:
Also fixes an issue of duplicate logging.

### DIFF
--- a/docs/topics/logging.rst
+++ b/docs/topics/logging.rst
@@ -241,6 +241,22 @@ e.g. in the spider's ``__init__`` method::
 If you run this spider again then INFO messages from
 ``scrapy.spidermiddlewares.httperror`` logger will be gone.
 
+Graylog settings
+----------------
+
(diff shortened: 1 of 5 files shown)


elacuesta:
Hey @viniciusd-avila, thanks for your contribution.

I know this is currently a "draft" PR, but I'm -1 on this subject.

I don't think Scrapy (or any other framework, for that matter) should include specific integrations like this one, it should instead be extensible enough to allow hooking custom functionality to it. In that spirit and if I'm not mistaken, the desired results could be accomplished by adding the Graylog handler to the root logger, perhaps in a custom [extension](https://docs.scrapy.org/en/1.7/topics/extensions.html), which means this code could be distributed as a Python package.

This is of course my personal opinion and I'm always willing to be proven wrong, let's keep the discussion open.

viniciusd-avila:
@elacuesta, thanks very much for you feedback. I agree that altering the source code this way is not ideal, just what I came up with out of necessity. Your suggestion looks doable, I'll try to implement an extension instead.

PR #4031, opened 2019-09-22
remove .keys() to avoid creating a tmp list/keyview obj

s-sanjay:
This is a small straight forward refactor to avoid creating a list or keyview object ( list in 2.x and key in 3.x ). 
There are other folders where this change can be made but starting with commands folder.

### DIFF
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -60,10 +60,12 @@ class Command(ScrapyCommand):
 
     @property
     def max_level(self):
-        levels = list(self.items.keys()) + list(self.requests.keys())
-        if not levels:
-            return 0
-        return max(levels)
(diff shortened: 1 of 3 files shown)


s-sanjay:
the test seems to only fail in py35-pinned env... weird..

s-sanjay:
> The 2nd and 3rd changes seem straightforward to me, but I wonder what is the impact of the first change (`max_level`) in performance.

so performance wise we are avoiding creating a dynamic list of size of both the list. 
we loop through both the dict, and take the max and take max of that max. 
earlier we loop through both the dict and create combined new list of keys and then find max in that list.

s-sanjay:
@Gallaecio can this be merged ?

PR #3819, opened 2019-06-07
[MRG+1] Fix missing values

sortafreel:
Issue: https://github.com/scrapy/scrapy/issues/3804

PR #4036, opened 2019-09-25
ItemLoader: improve handling of initial item

### LINKED ISSUE #3897: ItemLoaders can break if instantiated with pre-populated items

### LINKED ISSUE #3976: `ItemLoader` fields initialized from `item` are reprocessed

elacuesta:
Third time's the charm 🤞 

Fixes #3897, fixes #3976, supersedes #3998.

Continuing the work of @sortafreel, updating the implementation and adding more granular tests. Please do let me know if you think I'm missing some test case.


tests/test_loader.py
```diff
@@ -61,7 +67,7 @@ def test_load_item_using_default_loader(self):
         il.add_value('name', u'marta')
         item = il.load_item()
         assert item is i
-        self.assertEqual(item['summary'], u'lala')
+        self.assertEqual(item['summary'], [u'lala'])
```
elacuesta: I think this line technically makes the PR backward incompatible. I believe this is a reasonable price to pay, considering that the current code breaks if one wants to add more values to the same field, i.e. adding `il.add_value('summary', u'foobar')` after creating this loader results in
```
ERROR: test_load_item_using_default_loader (__main__.BasicItemLoaderTest)
----------------------------------------------------------------------
  File "tests/test_loader.py", line 62, in test_load_item_using_default_loader
    il.add_value('summary', u'foobar')
  File "/.../scrapy/loader/__init__.py", line 79, in add_value
    self._add_value(field_name, value)
  File "/.../scrapy/loader/__init__.py", line 95, in _add_value
    self._values[field_name] += arg_to_iter(processed_value)
TypeError: can only concatenate str (not "list") to str

**Comment by Gallaecio** [comment]
Do you think you could increase coverage of the changes?

If I’m looking at it right, what’s missing is a case where `if processed_value` evaluates to `False`, and a case where `elif value` evaluates to `False`.

**Comment by elacuesta** [comment]
Thanks for looking into this @Gallaecio.
I simplified the implementation, now test coverage is complete :tada:

**Comment by elacuesta** [comment]
Code samples from #3976 at 7c2ec902:

```python
In [1]: from pprint import pprint 
   ...:  
   ...: from scrapy import Field, Item 
   ...: from scrapy.loader import ItemLoader 
   ...: from scrapy.loader.processors import TakeFirst 
   ...:  
   ...: class X(Item): 
   ...:     x = Field(output_processor=TakeFirst()) 
   ...:  
   ...: loader = ItemLoader(X()) 
   ...: loader.add_value("x", ["value1", "value2"]) 
   ...: x = loader.load_item() 
   ...:  
   ...: pprint(x) 
   ...: pprint(ItemLoader(x).load_item())                                                                                                                                                                                                     
{'x': 'value1'}
{'x': 'value1'}
```

```python
In [2]: from scrapy.item import Item, Field 
   ...: from scrapy.loader import ItemLoader 
   ...: from scrapy.loader.processors import TakeFirst, Compose, Identity 
   ...:  
   ...: class Temp(Item): 
   ...:     temp = Field() 
   ...:     def __init__(self, *args, **kwargs): 
   ...:         super().__init__(self, *args, **kwargs) 
   ...:         self.setdefault('temp', 0.3) 
   ...:  
   ...: class Loader(ItemLoader): 
   ...:     default_item_class = Temp 
   ...:     default_input_processor = Identity() 
   ...:     default_output_processor = Compose(TakeFirst()) 
   ...:  
   ...: loader = Loader() 
   ...: loader.load_item()                                                                                                                                                                                                                    
Out[2]: {'temp': 0.3}
```

kmike:
Hey! I haven't checked the problem in detail, but I wonder if we can actually document the expected behavior, to avoid further confusion.

kmike:
//cc @andrewbaxter @ejulio @alexander-matsievsky - could you please check if the PR solves the problem in your projects?

andrewbaxter:
@kmike I'm on @elacuesta 's team so some bias here, but IIRC this will fix our problem and I think re-running output processors and not input processors is a reasonable policy.

If you're able though, a word on what level of backwards incompatibility is acceptable in any of the issues would be appreciated.  I imagine there's many projects out there that may need to upgrade and will be bitten unless we revert the original change and make it a separate factory method (for example).  If that's an expected cost I'm on board with this solution.

tests/test_loader.py
```diff
@@ -493,6 +464,213 @@ class TestItemLoader(ItemLoader):
...
+
+class InitializationFromDictTest(InitializationTestMixin, unittest.TestCase):
+    item_class = dict
+
+
+class InitializationFromItemTest(InitializationTestMixin, unittest.TestCase):
+    item_class = NameItem
+
+
+class NoInputReprocessingFromDictTest(unittest.TestCase):
```
ejulio: I'd say that `NoInputReprocessingFromDictTest` and `NoInputReprocessingFromItemTest` should use the same loader definition. To make the behavior clear.

elacuesta: Updated. The loaders are still defined separately because they need different `default_item_class` attributes, but now they both inherit from `BaseNoInputReprocessingLoader`. Thanks!

docs/topics/loaders.rst
```diff
@@ -35,6 +35,12 @@ Then, you start collecting values into the Item Loader, typically using
 the same item field; the Item Loader will know how to "join" those values later
 using a proper processing function.
 
+.. note:: Collected data is internally stored as lists,
+   allowing to add several values to the same field.
+   If an ``item`` argument is passed when creating a loader,
+   each of the item's values will be stored as-is if it's already
```
Gallaecio: stored as-is → converted into a list?

elacuesta: Not necessarily, since this piece of code uses [`arg_to_iter`](https://github.com/scrapy/scrapy/blob/1.7.4/scrapy/utils/misc.py#L19-L30), non-list iterables are not converted.

Gallaecio: Then “stored as lists” is not correct, and they are stored as iterables? Won’t there be issues if a generator is used, when trying to add more values to it?

elacuesta: True, although that's what all of the `add_*` methods do, because that's what `arg_to_iter` does. _Generally_, this isn't a problem because items and dicts are wrapped in lists, and the Selector query methods return SelectorLists.
Not sure I like it either, but I think making such change is probably out of the scope of this PR. Perhaps this grants a separate discussion/patch.

scrapy/loader/__init__.py
```diff
@@ -33,10 +33,9 @@ def __init__(self, item=None, selector=None, response=None, parent=None, **conte
         self.parent = parent
         self._local_item = context['item'] = item
         self._local_values = defaultdict(list)
-        # Preprocess values if item built from dict
-        # Values need to be added to item._values if added them from dict (not with add_values)
+        # values from initial item
         for field_name, value in item.items():
-            self._values[field_name] = self._process_input_value(field_name, value)
+            self._values[field_name] += arg_to_iter(value)
```
pawelmhm, line 38: I know this is old commit, but I'm now in the process of understanding this whole issue, and chain of events started with this https://github.com/scrapy/scrapy/pull/3819 and this is still a breaking change for many projects, people were certainly getting different data in many cases after all these changes. See below code as example, I have several spiders relying on old behavior and now looking for workaround.

```python

from scrapy import Item, Field


class AItem(Item):
    name = Field()


class ALoader(ItemLoader):
    default_item_class = AItem


# Breaking change between scrapy 1.7 and 2.4.
# Above is behavior that several spiders rely on, this was behavior up until 17.
# And it is not how loader works by default in Scrapy after 1.7 version.
def test_loader_breaking_change_scrapy_v17_v24():
    # To see this run this test in different environment, one with Scrapy 1.7
    # and old loader library, other with Scrapy 2.4.1 and itemloader library.
    loader = ALoader()
    loader.add_value('name', 'hello world')
    l = loader.load_item()
    new_loader = ALoader(l)
    new_loader.add_value("name", "lovely")
    l = new_loader.load_item()
    # after Scrapy 1.7.4 it gives: {'name': ['hello world', 'lovely']} 
    # until Scrapy 1.7.4 it was {'name': ['lovely']}
    assert l['name'] == ['lovely']
```

Gallaecio: Would using a “TakeLast” output processor (e.g. `lambda v: v[-1]`) work for your scenario?

pawelmhm: I have TakeFirst processor defined, and different spiders doing different things, some of them are fine after this change, but some are not. Changing output processors will change several things. I wonder if we could add some option for loader to keep old behavior? It seems dirty, but would make things easier. I could also refactor spiders, but there is many of them and even finding which one is affected is not that easy

Gallaecio: What kind of change do you have in mind? Something like a parameter to `ItemLoader` that sets the old behavior? Our would it work for you if we simply extended the `ItemLoader` API with something like `set_value` to replace any value added so far?

pawelmhm: > Something like a parameter to ItemLoader that sets the old behavior?

I was thinking about parameter or attribute of Loader. I could set this when defining my subclass of ItemLoader. But I see potential problem from library point of view - it can be hard to maintain in the future, there will be 2 different logics, not sure how many people will even think about this attribute or parameter. If I will be only project using it - it is not much sense, I can just subclass ItemLoader my side and change some things. I could also stick to old ItemLoader library but there is no easy way, because itemloaders were moved to separate repo, so I cannot simply use some old version of itemLoader with Scrapy 2.4.1. First release of itemloaders is after this change and scrapy imports itemloaders.

> Our would it work for you if we simply extended the ItemLoader API with something like set_value to replace any value added so far?

There is already replace_value, which is doing what I would need here. I guess the problem is because in the past in many cases loader.add_value was actually replacing value when people intended to add_value, now code relies on this and this creates mess when situation is cleared up.

PR #2354, opened 2016-10-21
[MRG+1] doc: wording

stav:
Am I being pedantic about wanting to change the word "constructor" to "initializer"?

PR #4045, opened 2019-09-30
Provide complete API documentation coverage of scrapy.linkextractors

Gallaecio:
I made the assumption that `FilteringLinkExtractor` was originally created for code sharing between the implementations based on `lxml` and `sgmllib`.

Here I’m marking `FilteringLinkExtractor` as deprecated, assuming that upon removal its code can go into `LxmlLinkExtractor`.

I’m also updating the documentation about link extractors to talk about `LxmlLinkExtractor` only, without even mentioning the possibility of alternative implementations.


elacuesta:
Looks good to me, except for the following:

> I’m also updating the documentation about link extractors to talk about LxmlLinkExtractor only, without even mentioning the possibility of alternative implementations.

Personally I don't remember seeing a custom Link extractor implementation, but I'm not sure about removing that section. It doesn't hurt, it doesn't need any particular code to be handled and it's technically still supported.
Just a comment, as I said I haven't seen or written any custom extractor so my opinion on the subject is not that strong.

Gallaecio:
If we leave it, I would like to at least mention a few use cases, and show a code example. But I cannot think of any use case myself.

Gallaecio:
@kmike What do you think, should custom link extractors remain in the documentation or go?

docs/topics/link-extractors.rst
```diff
@@ -4,46 +4,33 @@
...
-of :class:`scrapy.link.Link` objects. Link extractors are meant to be
-instantiated once and their ``extract_links`` method called several times
-with different responses to extract links to follow.
-
-Link extractors are used in the :class:`~scrapy.spiders.CrawlSpider`
-class (available in Scrapy), through a set of rules, but you can also use it in
-your spiders, even if you don't subclass from
-:class:`~scrapy.spiders.CrawlSpider`, as its purpose is very simple: to
-extract links.
+The constructor of :class:`~scrapy.linkextractors.lxmlhtml.LxmlLinkExtractor`
```
kmike: See https://github.com/scrapy/scrapy/pull/2354 :)

kmike:
I like the simplification of the docs, both the wording and content. +1 to deprecate / un-expose FilteringLinkExtractor. As I recall, it was needed when we had several link extractors, but now this code is all way to complex for what it is doing.

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -117,6 +117,14 @@ def __init__(self, allow=(), deny=(), allow_domains=(), deny_domains=(), restric
                                                 restrict_text=restrict_text)
 
     def extract_links(self, response):
+        """Returns a list of :class:`~scrapy.link.Link` objects from the
+        specified :class:`response <scrapy.http.Response>`.
+
+        Only links that match the settings passed to the link extractor
+        constructor are returned.
```
Gallaecio: constructor → __init__ method here as well

tests/test_linkextractors.py
```diff
@@ -506,3 +509,32 @@ def test_link_restrict_text(self):
...
         super(LxmlLinkExtractorTestCase, self).test_restrict_xpaths_with_html_entities()
+
+    def test_filteringlinkextractor_deprecation_warning(self):
+        """Make sure the FilteringLinkExtractor deprecation warning is not
+        issued for LxmlLinkExtractor"""
+        with catch_warnings(record=True) as warnings:
+            LxmlLinkExtractor()
+            self.assertEqual(len(warnings), 0)
+
+            class SubclassedItem(LxmlLinkExtractor):
```
kmike: What do you think about just "Subclassed", or "SubclassedLxmlLinkExtractor", or "SubclassedLinkExtractor"?

kmike:
Thanks @Gallaecio!

PR #4052, opened 2019-10-01
Fix TypeError when using DummyStatsCollector

### LINKED ISSUE #4007: Exception when using DummyStatsCollector
### Description

Using the DummyStatsCollector results in an exception:

```
2019-09-09 13:51:23 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method CoreStats.spider_closed of <scrapy.extensions.corestats.CoreStats object at 0x7f86269cac18>>
  File ".../lib/python3.6/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File ".../lib/python3.6/site-packages/pydispatch/robustapply.py", line 55, in robustApply
  File ".../lib/python3.6/site-packages/scrapy/extensions/corestats.py", line 28, in spider_closed
    elapsed_time = finish_time - self.stats.get_value('start_time')
TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType'
```

This problem has been introduced in aa46e1995cd5cb1099aba17535372b538bd656b3.

### Steps to Reproduce

Set `STATS_CLASS = "scrapy.statscollectors.DummyStatsCollector"` in the settings module as described in the documentation (https://docs.scrapy.org/en/latest/topics/stats.html#dummystatscollector).

**Expected behavior:** no exception
**Actual behavior:** exception thrown
**Reproduces how often:** always

### Versions

At least master as of 534de7395da3a53b5a2c89960db9ec5d8fdab60c

### Fix

A possible fix is to use the elapsed time as a default argument so that `get_value()` does not return None. I can prepare a PR if needed.

```diff
--- a/scrapy/extensions/corestats.py
+++ b/scrapy/extensions/corestats.py
@@ -25,7 +25,7 @@ class CoreStats(object):
 
     def spider_closed(self, spider, reason):
         finish_time = datetime.datetime.utcnow()
-        elapsed_time = finish_time - self.stats.get_value('start_time')
+        elapsed_time = finish_time - self.stats.get_value('start_time', finish_time)
         elapsed_time_seconds = elapsed_time.total_seconds()
         self.stats.set_value('elapsed_time_seconds', elapsed_time_seconds, spider=spider)
         self.stats.set_value('finish_time', finish_time, spider=spider)
```

elacuesta:
Fixes #4007

### DIFF
--- a/tests/test_stats.py
+++ b/tests/test_stats.py
@@ -1,8 +1,57 @@
+from datetime import datetime
 import unittest
 
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from scrapy.extensions.corestats import CoreStats
 from scrapy.spiders import Spider
 from scrapy.statscollectors import StatsCollector, DummyStatsCollector
 from scrapy.utils.test import get_crawler
+
+
+class CoreStatsExtensionTest(unittest.TestCase):
+
+    def setUp(self):
+        self.crawler = get_crawler(Spider)
+        self.spider = self.crawler._create_spider('foo')
+
+    @mock.patch('scrapy.extensions.corestats.datetime')
+    def test_core_stats_default_stats_collector(self, mock_datetime):
+        fixed_datetime = datetime(2019, 12, 1, 11, 38)
+        mock_datetime.utcnow = mock.Mock(return_value=fixed_datetime)
+        self.crawler.stats = StatsCollector(self.crawler)
+        ext = CoreStats.from_crawler(self.crawler)
+        ext.spider_opened(self.spider)
+        ext.item_scraped({}, self.spider)
+        ext.response_received(self.spider)
+        ext.item_dropped({}, self.spider, ZeroDivisionError())
+        ext.spider_closed(self.spider, 'finished')
+        self.assertEqual(
+            ext.stats._stats,
+            {
+                'start_time': fixed_datetime,
+                'finish_time': fixed_datetime,
+                'item_scraped_count': 1,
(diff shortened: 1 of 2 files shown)


tests/test_stats.py
```diff
@@ -51,3 +100,7 @@ def test_dummy_collector(self):
         stats.set_value('test', 'value', spider=self.spider)
         self.assertEqual(stats.get_stats(), {})
         self.assertEqual(stats.get_stats('a'), {})
+
+
+if __name__ == "__main__":
+    unittest_main()
```
kmike: this looks unnecessary, tests should be selected using py.test options. I think we should be removing this code from most files.

elacuesta: Sorry, tox was taking too long to run a single test :sweat_smile: 
I'll remove it, thanks!

elacuesta:
Rebased and squashed

kmike:
Thanks @elacuesta!

tests/test_stats.py
Gallaecio: You can use pytest parameters with tox, e.g. `tox -e py36 -- tests/test_stats.py`

elacuesta: > `tox -e py36 -- tests/test_stats.py`

I use that, it takes forever anyway. I might need to read tox docs for ways to speed things up.

Gallaecio: Oh, right, test collection does take a bit :sweat:

PR #4053, opened 2019-10-01
CsvItemExporter: Warning message on dropped fields

### LINKED ISSUE #4002: Warning messages to add in the CSV exporter when there's any field dropped
## Summary

A warning message shall be generated when there're some fields dropped within `CsvItemExporter` (or also in other exporters?) during exporting.
We may have this kind of warning messages emitted only for the first occurrence to avoid generating too much noise, similar to the warning messages when there's possible data loss ([related code](https://github.com/scrapy/scrapy/blob/d4b8bf18b2c7f563198d4334f5f5622f1006041d/scrapy/core/downloader/handlers/http11.py#L492)).

## Motivation

Due to nature of the CSV format, the `CsvItemExporter` needs to determine the list of fields to write before writing the 1st record ([related code](https://github.com/scrapy/scrapy/blob/d4b8bf18b2c7f563198d4334f5f5622f1006041d/scrapy/exporters.py#L253)), which would in some occasions cause unintentional loss of data.

Unfortunately I've yet observed this kind of mistakenly introduced data loss issue from different developers, where the common cause may be:
- Returning a `dict` as the 1st item that doesn't have all fields included.
- Returning more than one type of items that have different fields.

(both when not having `FEED_EXPORT_FIELDS` properly configured)

Properly configuring the `FEED_EXPORT_FIELDS` setting entry would surely have the issue resolved. However that's not always guaranteed (a developer may just forget to, etc.).

amardeep24:
Fix for [Warning messages to add in the CSV exporter when there's any field dropped #4002](https://github.com/scrapy/scrapy/issues/4002).

We check if the FEED_EXPORT_FIELDS is set or not, if not a warning `Possible chance of data loss detected -- This message won't be shown in further requests` is logged. This warning is also logged when the first row length and subsequent row length differs signalling that there is a possibility of data loss.

Fixes #4002

### DIFF
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -8,6 +8,7 @@ import sys
 import pprint
 import marshal
 import six
+import logging
 from six.moves import cPickle as pickle
 from xml.sax.saxutils import XMLGenerator
 
@@ -21,6 +22,9 @@ import warnings
 __all__ = ['BaseItemExporter', 'PprintItemExporter', 'PickleItemExporter',
            'CsvItemExporter', 'XmlItemExporter', 'JsonLinesItemExporter',
            'JsonItemExporter', 'MarshalItemExporter']
+
+
+logger = logging.getLogger(__name__)
 
 
 class BaseItemExporter(object):
@@ -220,6 +224,8 @@ class CsvItemExporter(BaseItemExporter):
         self.csv_writer = csv.writer(self.stream, **kwargs)
         self._headers_not_written = True
         self._join_multivalued = join_multivalued
+        self._is_data_loss_warning_displayed = False
+        self._fail_on_dataloss_warned = False
 
     def serialize_field(self, field, name, value):
         serializer = field.get('serializer', self._join_if_needed)
@@ -253,12 +259,23 @@ class CsvItemExporter(BaseItemExporter):
     def _write_headers_and_set_fields_to_export(self, item):
         if self.include_headers_line:
             if not self.fields_to_export:
+                self._is_data_loss_warning_displayed = True
                 if isinstance(item, dict):
                     # for dicts try using fields of the first item
                     self.fields_to_export = list(item.keys())
                 else:
                     # use fields declared in Item
                     self.fields_to_export = list(item.fields.keys())
+            elif not self._fail_on_dataloss_warned:
+                if isinstance(item, dict):
+                    if len(item.keys()) > self.fields_to_export:
+                        self._is_data_loss_warning_displayed = True
+                else:
+                    if len(item.fields.keys()) > self.fields_to_export:
+                        self._is_data_loss_warning_displayed = True
+            if self._is_data_loss_warning_displayed and not self._fail_on_dataloss_warned:
+                logger.warning("Possible chance of data loss detected -- This message won't be shown in further requests")
+                self._fail_on_dataloss_warned = True
             row = list(self._build_row(self.fields_to_export))
             self.csv_writer.writerow(row)
 
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -249,8 +249,10 @@ class CsvItemExporter(BaseItemExporter):
         if not self._data_loss_already_warned:
             if self.fields_to_export is None:
                 logger.warning("Data loss detected when exporting items -- This message won't be shown for further items")
+                self._data_loss_already_warned = True
             elif set(fields) != set(self.fields_to_export):
                 logger.info("Data loss detected when exporting items -- This message won't be shown for further items")
+                self._data_loss_already_warned = True
         values = list(self._build_row(x for _, x in serialized_fields))
         self.csv_writer.writerow(values)
 
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -268,7 +268,7 @@ class CsvItemExporter(BaseItemExporter):
                 self._is_data_loss_warning_displayed = True
                 self.fields_to_export = fields
             elif not self._fail_on_dataloss_warned:
-                if len(fields) > len(self.fields_to_export):
+                if fields != self.fields_to_export:
                     self._is_data_loss_warning_displayed = True
             if self._is_data_loss_warning_displayed and not self._fail_on_dataloss_warned:
                 logger.warning("Possible chance of data loss detected -- This message won't be shown in further requests")
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -258,21 +258,18 @@ class CsvItemExporter(BaseItemExporter):
 
     def _write_headers_and_set_fields_to_export(self, item):
         if self.include_headers_line:
+            if isinstance(item, dict):
+                # for dicts try using fields of the first item
+                fields = list(item.keys())
+            else:
+                # use fields declared in Item
+                fields = list(item.fields.keys())
             if not self.fields_to_export:
                 self._is_data_loss_warning_displayed = True
-                if isinstance(item, dict):
-                    # for dicts try using fields of the first item
-                    self.fields_to_export = list(item.keys())
-                else:
-                    # use fields declared in Item
-                    self.fields_to_export = list(item.fields.keys())
+                self.fields_to_export = fields
             elif not self._fail_on_dataloss_warned:
-                if isinstance(item, dict):
-                    if len(item.keys()) > self.fields_to_export:
-                        self._is_data_loss_warning_displayed = True
-                else:
-                    if len(item.fields.keys()) > self.fields_to_export:
-                        self._is_data_loss_warning_displayed = True
+                if len(fields) > len(self.fields_to_export):
+                    self._is_data_loss_warning_displayed = True
             if self._is_data_loss_warning_displayed and not self._fail_on_dataloss_warned:
                 logger.warning("Possible chance of data loss detected -- This message won't be shown in further requests")
                 self._fail_on_dataloss_warned = True
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -244,11 +244,13 @@ class CsvItemExporter(BaseItemExporter):
             self._headers_not_written = False
             self._write_headers_and_set_fields_to_export(item)
 
-        serialized_fields = self._get_serialized_fields(item, default_value='',
-                                             include_empty=True)
+        serialized_fields = self._get_serialized_fields(item, default_value='', include_empty=True)
         fields = self._get_fields_from_item(item)
-        if (self.fields_to_export == None or set(fields) != set(self.fields_to_export)) and not self._data_loss_already_warned:
-            logger.warning("Data loss detected when exporting items -- This message won't be shown in further items")
+        if not self._data_loss_already_warned:
+            if self.fields_to_export is None:
+                logger.warning("Data loss detected when exporting items -- This message won't be shown for further items")
+            elif set(fields) != set(self.fields_to_export):
+                logger.info("Data loss detected when exporting items -- This message won't be shown for further items")
         values = list(self._build_row(x for _, x in serialized_fields))
         self.csv_writer.writerow(values)
 
@@ -269,7 +271,7 @@ class CsvItemExporter(BaseItemExporter):
                 if set(fields) != set(self.fields_to_export):
                     self._show_data_loss_warning = True
             if self._show_data_loss_warning and not self._data_loss_already_warned:
-                logger.warning("Data loss detected when exporting items -- This message won't be shown in further items")
+                logger.warning("Data loss detected when exporting items -- This message won't be shown for further items")
                 self._data_loss_already_warned = True
             row = list(self._build_row(self.fields_to_export))
             self.csv_writer.writerow(row)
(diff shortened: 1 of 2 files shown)


scrapy/exporters.py
```diff
@@ -253,12 +259,23 @@ def _build_row(self, values):
...
             if not self.fields_to_export:
+                self._is_data_loss_warning_displayed = True
                 if isinstance(item, dict):
                     # for dicts try using fields of the first item
                     self.fields_to_export = list(item.keys())
                 else:
                     # use fields declared in Item
                     self.fields_to_export = list(item.fields.keys())
+            elif not self._fail_on_dataloss_warned:
+                if isinstance(item, dict):
```
elacuesta: Can we reduce code duplication here by doing something like `fields = item.keys() if isinstance(item, dict) else item.fields.keys()` and do the following checks only once?

scrapy/exporters.py
```diff
@@ -253,12 +259,23 @@ def _build_row(self, values):
...
+                    if len(item.keys()) > self.fields_to_export:
```
elacuesta: Related to https://github.com/scrapy/scrapy/issues/4002#issuecomment-536662554:
I think these checks should compare the actual fields, not the amount of fields: `len({"a", "b"}) == len({"b", "c"})` but `{"a", "b"} != {"b", "c"}`.

scrapy/exporters.py
amardeep24: Yeah, sorry I will remove the redundant if statements

scrapy/exporters.py
amardeep24: As per my understanding, if the `fields_to_export` is set and if some of the rows have more number of columns than the columns in `fields_to_export` then we should be showing the warning. That is the reason why I check for the length, if on any row the length of the row is greater than the length of the `fields_to_export`, I was showing the warning.

You are right, I will add the `len(fields_to_export)` to compare instead of comparing directly with `fields_to_export` which is not a number.

Gallaecio: The number is not enough, because as with the example that @elacuesta provided (`{"a", "b"} != {"b", "c"}`), two items may have the same number of columns but different field names.

scrapy/exporters.py
```diff
@@ -252,13 +258,21 @@ def _build_row(self, values):
...
-                else:
-                    # use fields declared in Item
-                    self.fields_to_export = list(item.fields.keys())
+                self._is_data_loss_warning_displayed = True
+                self.fields_to_export = fields
+            elif not self._fail_on_dataloss_warned:
+                if fields != self.fields_to_export:
+                    self._is_data_loss_warning_displayed = True
+            if self._is_data_loss_warning_displayed and not self._fail_on_dataloss_warned:
+                logger.warning("Possible chance of data loss detected -- This message won't be shown in further requests")
```
elacuesta: I think this warning should:
* Appear only when there is actual data loss, i.e. when the current fields are definitely different from the stored in the exporter instance. Hence "Possible chance of" should go away IMHO (https://github.com/scrapy/scrapy/issues/4002#issuecomment-536662554)
* Give a little bit more information about the data loss, maybe "Data loss detected **when exporting items**" (although this might be covered by the logger name). Perhaps mentioning that some fields are not being included in the output.
* Talk about items instead of requests ("This message won't be shown **for further items**")

scrapy/exporters.py
```diff
@@ -252,13 +258,21 @@ def _build_row(self, values):
...
+                # use fields declared in Item
+                fields = list(item.fields.keys())
             if not self.fields_to_export:
-                if isinstance(item, dict):
-                    # for dicts try using fields of the first item
-                    self.fields_to_export = list(item.keys())
-                else:
-                    # use fields declared in Item
-                    self.fields_to_export = list(item.fields.keys())
+                self._is_data_loss_warning_displayed = True
```
elacuesta: I'm failing to see the need for two flags to handle this feature. At the very least, I don't understand the names, i.e. how `self.fields_to_export` being unset implies setting `self._is_data_loss_warning_displayed = True`.
Could this be simplified to use only one flag, or maybe do some changes to have more self-explanatory flag names?

tests/test_feedexport.py
```diff
@@ -622,6 +622,45 @@ def test_export_dicts(self):
         yield self.assertExportedCsv(items, ['egg', 'foo'], rows_csv, ordered=False)
         yield self.assertExportedJsonLines(items, rows_jl)
 
+    @defer.inlineCallbacks
+    def test_export_dicts_warning(self):
```
elacuesta: This looks great, but could you also add tests using the `Item` class? Thanks!

scrapy/exporters.py
amardeep24: Cool I will do those fixes, but what if `fields_to_export` is not set but there is no actual data loss.In other words the row size and fields remain the same for all fields but the `fields_to_export` is not set. Do I need to show the same message?

scrapy/exporters.py
amardeep24: There are three cases when we should show the warning:
1. When the `fields_to_export ` is not set and we get the fields from the first row.
2. When the `fields_to_export ` is set but the fields in the rows differ
3. When the `fields_to_export ` is not set and we get the fields from the first row but the subsequent rows differ in content from the first row.

So simply checking the `fields_to_export ` is set or not won't let me understand whether to show the warning or not so i need a flag to check the conditions 2 & 3.

Also I need another flag to check if I have already shown the warning, if it is shown already then I should not show it again.

I renamed the flags to `_show_data_loss_warning` & ` _data_loss_already_warned`.

tests/test_feedexport.py
amardeep24: Added a new test case! Thanks!

scrapy/exporters.py
amardeep24: @elacuesta did I get it right? what do you think? :smiley:

elacuesta: I believe the following piece of code should be equivalent:
```python
if self.fields_to_export:
    if set(fields) != set(self.fields_to_export) and not self._data_loss_already_warned:
        logger.warning("Data loss detected when exporting items -- This message won't be shown in further items")
        self._data_loss_warned = True
else:
    self.fields_to_export = fields
```
Tests fail because this ignores (1), which I think should not be accounted for (imagine `fields_to_export` is unset but all fields have the same fields).

However, while checking this I realized the fix only runs on `_write_headers_and_set_fields_to_export`, which is called only once (if the header row was not written already). If I understand the issue correctly, we would want this check to be executed for every item.

scrapy/exporters.py
elacuesta: I think there should be no warning in that case (https://github.com/scrapy/scrapy/pull/4053#discussion_r331736731)

scrapy/exporters.py
amardeep24: You mean apart from adding the fix in `_write_headers_and_set_fields_to_export` we need to add it in `export_item` method also?

amardeep24: Then we can define a new attribute called `_fields_size` and set to an initial value, now if it varies because of the changing row size we would trigger the warning?

amardeep24: @elacuesta I added a new commit to check if data loss is happening for every item, the check is added in the `export_item ` method.

scrapy/exporters.py
```diff
@@ -238,9 +244,12 @@ def export_item(self, item):
             self._headers_not_written = False
             self._write_headers_and_set_fields_to_export(item)
 
-        fields = self._get_serialized_fields(item, default_value='',
+        serialized_fields = self._get_serialized_fields(item, default_value='',
                                              include_empty=True)
-        values = list(self._build_row(x for _, x in fields))
+        fields = self._get_fields_from_item(item)
+        if (self.fields_to_export == None or set(fields) != set(self.fields_to_export)) and not self._data_loss_already_warned:
```
Gallaecio: If `self.fields_to_export` is defined, as per @kmike’s comment [here](https://github.com/scrapy/scrapy/issues/4002#issuecomment-536655863), and INFO statement should be logged instead of a WARNING, because the use has manually chosen a subset of all possible files.

elacuesta: Please take a look at this @amardeep24

elacuesta: > Comparisons to singletons like None should always be done with is or is not, never the equality operators

Source: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

scrapy/exporters.py
```diff
@@ -238,9 +244,12 @@ def export_item(self, item):
...
+            logger.warning("Data loss detected when exporting items -- This message won't be shown in further items")
```
elacuesta:
```suggestion
            logger.warning("Data loss detected when exporting items -- This message won't be shown for further items") 
```
I believe this sounds better, although I'm no native English speaker.

scrapy/exporters.py
```diff
@@ -238,9 +244,12 @@ def export_item(self, item):
             self._headers_not_written = False
             self._write_headers_and_set_fields_to_export(item)
 
-        fields = self._get_serialized_fields(item, default_value='',
+        serialized_fields = self._get_serialized_fields(item, default_value='',
                                              include_empty=True)
```
elacuesta: This line is not aligned with the one above anymore

scrapy/exporters.py
```diff
@@ -252,16 +261,28 @@ def _build_row(self, values):
 
     def _write_headers_and_set_fields_to_export(self, item):
         if self.include_headers_line:
+            fields = self._get_fields_from_item(item)
             if not self.fields_to_export:
```
elacuesta: I'm not sure this is correct. Unless I'm missing something, this will show a warning if `FEED_EXPORT_FIELDS` was not set even if all items have the same fields, and hence all fields are exported. Could you add a test case for this scenario?

elacuesta (changes requested):

scrapy/exporters.py
amardeep24: So I am adding an INFO level logger if the user explicitly sets the exported items as a subset. This INFO logger will be issued only once.

```
 if not self._data_loss_already_warned:
            if self.fields_to_export is None:
                logger.warning("Data loss detected when exporting items -- This message won't be shown in further items")
            elif set(fields) != set(self.fields_to_export):
                logger.info("Data loss detected when exporting items -- This message won't be shown in further items")
```

scrapy/exporters.py
amardeep24: Yes, it does sound better!

scrapy/exporters.py
amardeep24: Yes, this was originally the intention that if the `FEED_EXPORT_FIELDS `is not set, then we would show that there could be a chance of a _**possible**_ data loss. But subsequently it was decided we would not tell _possible_  but will tell that an actual data loss is detected when the fields length differ.

scrapy/exporters.py
```diff
@@ -238,9 +244,14 @@ def export_item(self, item):
...
-        fields = self._get_serialized_fields(item, default_value='',
-                                             include_empty=True)
-        values = list(self._build_row(x for _, x in fields))
+        serialized_fields = self._get_serialized_fields(item, default_value='', include_empty=True)
+        fields = self._get_fields_from_item(item)
+        if not self._data_loss_already_warned:
+            if self.fields_to_export is None:
+                logger.warning("Data loss detected when exporting items -- This message won't be shown for further items")
+            elif set(fields) != set(self.fields_to_export):
+                logger.info("Data loss detected when exporting items -- This message won't be shown for further items")
```
Gallaecio: Why is `self._data_loss_already_warned` not set to `True` here? Won’t that cause this message to happen on every exported item?

amardeep24: Yes, I don't know how I missed that. I apologize.

scrapy/exporters.py
amardeep24: @elacuesta should I keep this logic? or change the warning message to `"Possible data loss detected when exporting items -- This message won't be shown for further items"` if the `FEED_EXPORT_FIELDS` was not set even if all the items have the same fields.

Gallaecio (commented):
I think we need to improve the messages. Users that see this message need to know what is causing the potential data loss, and how to address that.

ejulio:
Just adding an idea here.
What if we had a flag `FEED_EXPORT_WARN_DATA_LOSS` (bool).
Then, if there was any data loss (different fields then the header) it would log a message.
IMHO, this warning should be raised even if `FEED_EXPORT_FIELDS` is set, as it can happen that we update the item definition, but not `FEED_EXPORT_FIELDS`.
By default, the flag should be `True` and the user should disable it on purpose.

Also, be aware that a simple comparison `set(item_fields) != set(fields_to_export)` is not going to work.
Because, if the `item` has less fields than `fields_to_export`, it shouldn't be a problem.
It's only problematic when it has more fields. So, a better idea is to use `<=` (

```
>>> item1_fields = {'a', 'b'}  # no data loss
>>> item2_fields = {'a', 'b', 'c'}  # data loss
>>> item3_fields = {'a'}  # no data loss
>>> fields_to_export = {'a', 'b'}
>>> item1_fields <= fields_to_export
True
>>> item2_fields <= fields_to_export
False
>>> item3_fields <= fields_to_export
True
>>> {'d', 'e'} <= fields_to_export  # data loss
False
```

PR #4054, opened 2019-10-01
Expose certificate for HTTPS responses

elacuesta:
Fixes #2726

The approach I took here is very similar to the one in #3940.

~~No tests at the moment (not sure how to test this without requesting external pages since the mockserver uses HTTP, any help is appreciated :heart:),~~ posting a `scrapy shell` session instead:
```
$ scrapy shell https://example.org
2019-10-01 14:23:58 [scrapy.utils.log] INFO: Scrapy 1.7.0 started (bot: scrapybot)
(...)
Available Scrapy objects:
[s]   scrapy     scrapy module (contains scrapy.Request, scrapy.Selector, etc)
[s]   crawler    <scrapy.crawler.Crawler object at 0x7fdfc53b0208>
[s]   item       {}
[s]   request    <GET https://example.org>
[s]   response   <200 https://example.org>
[s]   settings   <scrapy.settings.Settings object at 0x7fdfc1147898>
[s]   spider     <DefaultSpider 'default' at 0x7fdfc073af98>
[s] Useful shortcuts:
[s]   fetch(url[, redirect=True]) Fetch URL and update local objects (by default, redirects are followed)
[s]   fetch(req)                  Fetch a scrapy.Request and update local objects 
[s]   shelp()           Shell help (print this help)
[s]   view(response)    View response in a browser
In [1]: response.certificate                                                                                                                                                                                                                  
Out[1]: <Certificate Subject=b'www.example.org' Issuer=b'DigiCert SHA2 Secure Server CA'>

In [2]: response.certificate.getIssuer()                                                                                                                                                                                                      
Out[2]: <DN 'commonName': b'DigiCert SHA2 Secure Server CA', 'organizationName': b'DigiCert Inc', 'countryName': b'US'>

In [3]: print(response.certificate.getIssuer().inspect())                                                                                                                                                                                     
               Common Name: DigiCert SHA2 Secure Server CA
              Country Name: US
         Organization Name: DigiCert Inc
```

PR #4055, opened 2019-10-01
[Doc] Use autoclass in topics/request-response.rst

elacuesta:
As suggested in https://github.com/scrapy/scrapy/pull/4054#discussion_r330207869, plus other wording fixes.


kmike:
The change looks good, but I haven't checked how it looks after the rendering. @Gallaecio if you checked it, please merge :)

PR #4056, opened 2019-10-02
Fix internal links in the tutorial and release notes

Gallaecio:


docs/news.rst
```diff
@@ -2448,7 +2455,7 @@ Scrapy changes:
 - added options ``-o`` and ``-t`` to the :command:`runspider` command
 - documented :doc:`topics/autothrottle` and added to extensions installed by default. You still need to enable it with :setting:`AUTOTHROTTLE_ENABLED`
 - major Stats Collection refactoring: removed separation of global/per-spider stats, removed stats-related signals (``stats_spider_opened``, etc). Stats are much simpler now, backward compatibility is kept on the Stats Collector API and signals.
-- added :meth:`~scrapy.contrib.spidermiddleware.SpiderMiddleware.process_start_requests` method to spider middlewares
+- added :meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_start_requests` method to spider middlewares
```
kmike: I'm not sure we should be doing this, because by the time of the release scrapy.contrib.spidermiddleware.SpiderMiddleware was changed, not scrapy.spidermiddlewares.SpiderMiddleware. Links in changelog are going to be broken anyways after changes, e.g. when we remove some class.

On the other hand, it doesn't hurt to have a link, and it stays the same in docs for previous versions, so I'm fine with this change - but probably it doesn't make sense to be very strict maintaining links in changelog.

docs/topics/items.rst
```diff
@@ -240,6 +240,10 @@ Item objects
     Items replicate the standard `dict API`_, including its constructor. The
     only additional attribute provided by Items is:
 
+    .. automethod:: copy
```
kmike: dict method is present in Python dict. Previous text doesn't match the following docs now - "The only additional attribute provided by Items is", and then instead of this single additional attribute two methods are described.

kmike:
Good changes, thanks @Gallaecio!

docs/news.rst
Gallaecio: It was a error on my side. The goal for news.rst references is to fix links (or remove them when impossible) but always keep the original display text. On it.

Gallaecio: Nevermind, it think both you and me missed the `~` at the beginning, which causes the display text to be just `process_start_requests`, so only the link is changing.

kmike:
Thank you @Gallaecio!

PR #4057, opened 2019-10-02
Response.follow_all

### LINKED ISSUE #2582: response.follow_all or SelectorList.follow_all shortcut
What do you think about adding response.follow_all shortcut, which returns a list of requests? This is inspired by this note in docs:

> `response.follow(response.css('li.next a'))` is not valid because response.css returns a list-like object with selectors for all results, not a single selector. A for loop like in the example above, or response.follow(response.css('li.next a')[0]) is fine.

So instead of 

```py
for href in response.css('li.next a::attr(href)'):
    yield response.follow(href, callback=self.parse)
```

users would be able to write (in Python 3)

```py
yield from response.follow_all(response.css('li.next a::attr(href)'), self.parse)
```

We can also add 'css' and 'xpath' support to it, as keyword arguments; it would shorten the code to this:

```py
yield from response.follow_all(css='li.next a::attr(href)', callback=self.parse)
```

(this is a follow-up to https://github.com/scrapy/scrapy/issues/1940 and https://github.com/scrapy/scrapy/issues/2540)

elacuesta:
Fixes #2582


kmike:
Thanks @elacuesta! 

I think there is one tricky case with follow_all API design: a case where some of the links are invalid.

For example, `response.follow_all(response.css("a"))`, or `response.follow_all(css="a")` may match some `<a>` elements without "href" attribute. This is not an uncommon case; I recall noticing it early, when updating a tutorial for response.follow. Check https://docs.scrapy.org/en/latest/intro/tutorial.html#more-examples-and-patterns - code like this is used everywhere:

```py
    # follow pagination links
    for href in response.css('li.next a::attr(href)'):
        yield response.follow(href, self.parse)
```

instead of 

```py
    # follow pagination links
    for a in response.css('li.next a'):
        yield response.follow(a, self.parse)
```

because the latter often fails in practice. This is common e.g. for pagination links - often a link to the current page is still an `<a>` element, but without href.

It'd be nice to be able to write

```py
yield from response.follow_all(".pagination a")
```

but with the current implementation `response.follow` would fail for these elements (raise exceptions), which would make some of the links on page followed, and some not - only links before the first exceptions are processed. 

I wonder if we should solve it somehow in the API of response.follow_all, and skip some of the "bad" cases. 

It requires a discussion though, I'm not sure what's the best approach.

elacuesta:
Good point :thinking: 

If we leave it as it currently is, I think the error message is probably clear enough (`<a> element has no href attribute`) and the same solution used for `follow` also applies to `follow_all`:
```python
yield from response.follow_all(css=".pagination a::attr(href)")
```
On the other hand, I agree that it would be cleaner to write `.pagination a`, but I wouldn't like to silently skip invalid links. An alternative could be to add a `skip_invalid` boolean parameter to `follow_all`, to be more explicit about that behaviour. What do you think @kmike ?

elacuesta:
On second thought, who would pass `skip_invalid=True`? Skipping invalid links should be fine as long as it's properly documented.

elacuesta:
@kmike updated, please check again :slightly_smiling_face:

scrapy/http/response/text.py
```diff
@@ -158,6 +159,66 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        In addition, ``css`` and ``xpath`` arguments are accepted to perform the link extraction
+        within the ``follow_all`` method (only one of ``urls``, ``css`` and ``xpath`` are accepted).
+
+        Note that when using the ``css`` or ``xpath`` parameters, this method will not produce
+        requests for selectors from which links cannot be obtained (for instance, anchor tags
+        without ``href`` attribute)
+        """
+        arg_count = len(list(filter(None, (urls, css, xpath))))
+        if arg_count != 1:
+            raise ValueError('Please supply exactly one of the following arguments: {urls, css, xpath}')
```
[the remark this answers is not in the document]
elacuesta: I just felt it would look more like a set :man_shrugging:

scrapy/http/response/text.py
```diff
@@ -158,6 +159,66 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+                expression = css
+            elif xpath:
+                selector_method = getattr(self, 'xpath')
+                expression = xpath
+            for selector in selector_method(expression):
+                try:
+                    urls.append(_url_from_selector(selector))
+                except ValueError:
+                    pass
+        return (
```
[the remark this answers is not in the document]
elacuesta: My first implementation was like that, the problem is that the checks that raise exceptions are not executed until the result is first iterated. I'm all about generators and not putting everything in memory if it's not strictly necessary, but since the `css` and `xpath` methods return selector lists, there is not really a lot to be done in that regard.

scrapy/http/response/text.py
```diff
@@ -158,6 +159,66 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        """
+        arg_count = len(list(filter(None, (urls, css, xpath))))
+        if arg_count != 1:
+            raise ValueError('Please supply exactly one of the following arguments: {urls, css, xpath}')
+        if not urls:
+            urls = []
+            if css:
+                selector_method = getattr(self, 'css')
+                expression = css
+            elif xpath:
```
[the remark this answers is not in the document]
elacuesta: If I'm not mistaken, that shouldn't be necessary, it's covered by the following:
https://github.com/scrapy/scrapy/blob/c7c54f5453eaa16f2b6d6441be68b0a272606ec9/scrapy/http/response/text.py#L188-L190

I'm not sure why the `elif xpath` line is yellow in Codecov, it should be reached by:
https://github.com/scrapy/scrapy/blob/c7c54f5453eaa16f2b6d6441be68b0a272606ec9/tests/test_http_response.py#L586-L603

scrapy/http/response/text.py
```diff
@@ -158,6 +159,66 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        initializer, except that each ``urls`` element does not need to be an absolute
+        URL, it can be any of the following:
+
+        * a relative URL;
+        * a scrapy.link.Link object (e.g. a link extractor result);
+        * an attribute Selector (not SelectorList) - e.g.
+          ``response.css('a::attr(href)')[0]`` or
+          ``response.xpath('//img/@src')[0]``.
+        * a Selector for ``<a>`` or ``<link>`` element, e.g.
+          ``response.css('a.my_link')[0]``.
```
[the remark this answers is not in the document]
elacuesta: Updated, also the `follow` docstring, which is the one I used as base. Thanks!

scrapy/http/response/text.py
```diff
@@ -158,6 +159,66 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
 
+    def follow_all(self, urls=None, callback=None, method='GET', headers=None, body=None,
+                   cookies=None, meta=None, encoding=None, priority=0,
+                   dont_filter=False, errback=None, cb_kwargs=None,
+                   css=None, xpath=None):
+        # type: (...) -> Generator[Request, None, None]
+        """
+        A generator that produces :class:`~.Request` instances to follow all
+        links in ``urls``. It accepts the same arguments as the :class:`~.Request`
+        initializer, except that each ``urls`` element does not need to be an absolute
```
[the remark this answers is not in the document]
elacuesta: My understanding is that `__new__` are constructors, and `__init__` are initializers, although this doesn't seem to matter much in practice in most of the docs I have read. Should we sacrifice correctness in favour of clarity?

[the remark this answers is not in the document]
Gallaecio: I’ve not been able to find a reference to ‘initializer’ in https://docs.python.org/3/reference/datamodel.html, but looking at https://docs.python.org/3/reference/datamodel.html#object.__new__ ‘constructor’ is not appropriate either, as it seems to be used to refer to the class itself when used to create an instance.

Maybe we should explicitly say ‘`__init__` method’.

scrapy/http/response/text.py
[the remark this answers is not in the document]
Gallaecio: It may just be that the coverage reporter cannot be smart enough to determine that an `else` here is not possible at the moment. I still we should go with `else & assert`, though, even if it’s just to silence the coverage checker.

scrapy/http/response/text.py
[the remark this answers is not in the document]
elacuesta: I suspect the official docs might intentionally skip the "constructor" and "initializer" names, to avoid this kind of confusion (and perhaps because even the core devs do not agree on them 😅)
You are correct, mentioning `__init__` sounds good, thanks!

scrapy/http/response/text.py
```diff
@@ -129,23 +128,82 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        if not urls:
+            if css:
+                selector_list = self.css(css)
+            if xpath:
+                selector_list = self.xpath(xpath)
+            urls = []
+            for selector in selector_list:
+                try:
+                    urls.append(_url_from_selector(selector))
+                except ValueError:
```
kmike: I think it'd be better to modify _url_from_selector to raise a more specific exception (ValueError subclass), just in case we have some unexpected ValueError

tests/sample_data/link_extractor/sgml_linkextractor_no_href.html
```diff
@@ -0,0 +1,25 @@
+<html>
+  <head>
+    <base href='http://example.com' />
+    <title>Sample page with anchor tags containing no href attribute, to test the TextResponse.follow_all method</title>
```
kmike: why is this file called sgml_linkextractor_no_href? sgml link extractor is deprecated.

kmike:
Thanks @elacuesta for working on it, and @Gallaecio + @wRAR for reviews!

I think that's very close to be ready, but it'd be good to update non-reference docs as well (tutorial?). We an do it after the merge as well, if @Gallaecio can take it.

tests/sample_data/link_extractor/sgml_linkextractor_no_href.html
elacuesta: True, I didn't realize that. I copied the test from [`scrapy.tests.test_http_response.BaseResponseTest._links_response`](https://github.com/scrapy/scrapy/blob/1.8.0/tests/test_http_response.py#L176-L179), but that doesn't seem to be specific to the `SgmlLinkExtractor`. I'll update the names, thanks for noticing.

Gallaecio:
> it'd be good to update non-reference docs as well (tutorial?).

@elacuesta https://github.com/elacuesta/scrapy/pull/3

Only the tutorial had `follow` usages that were better suited for `follow_all`.

docs/intro/tutorial.rst
```diff
@@ -625,12 +625,12 @@ attribute automatically. So the code can be shortened further::
     for a in response.css('li.next a'):
         yield response.follow(a, callback=self.parse)
 
-.. note::
+To create multiple requests from an iterable, you can use
```
kmike: This explanation is a bit confusing, because `li.next a` should match a single link in most cases. 

A more natural example would be something like `response.css('.author ~ a')` - follow all links to authors - though the selector could be too complex for a tutorial (or maybe not?).

Gallaecio: Yeah, I looked at the look and did not noticed the actual CSS expression. May I remove the look as I revert back to `follow`?

kmike: Thanks @Gallaecio! 

There is a use case for a loop though (and for follow_all on a single link), as it works on a last page, where there is no "next" link; new code (with `[0]`) fails with exception. 

I'm on fence on whether this pattern is good or not (use follow_all for cases where 0 or 1 result is expected). I can see myself using this pattern in spider code, but it can make the code less readable.

scrapy/http/response/text.py
```diff
@@ -155,19 +160,79 @@ def follow(self, url, callback=None, method='GET', headers=None, body=None,
...
+        * a :class:`~scrapy.selector.Selector` object for a ``<link>`` or ``<a>`` element, e.g.
+          ``response.css('a.my_link')[0]``
+        * an attribute :class:`~scrapy.selector.Selector` (not SelectorList), e.g.
+          ``response.css('a::attr(href)')[0]`` or
+          ``response.xpath('//img/@src')[0]``
+
+        In addition, ``css`` and ``xpath`` arguments are accepted to perform the link extraction
+        within the ``follow_all`` method (only one of ``urls``, ``css`` and ``xpath`` is accepted).
+
+        Note that when using the ``css`` or ``xpath`` parameters, this method will not produce
```
kmike: So, 

```py
yield from response.follow_all(css='a')
```

is not a shortcut for 

```py
sel = response.css('a')
yield from response.follow_all(sel)
```

? The difference is in what happens when some of the links don't have href attribute. With `css='a'` such links are skipped; with `sel` links before the first "a without href" are scheduled, then exception is raised?

elacuesta: Nice catch, thanks for noticing. Updated the implementation, docs and tests accordingly.

docs/intro/tutorial.rst
elacuesta: To be honest, I don't remember if we agreed on something specific here, but the following is a somewhat verbose `IndexError`-safe alternative using `response.follow`:
```python
a = response.css('li.next a')
if a:
    yield response.follow(a[0], callback=self.parse)
```

elacuesta: Replaced `li.next` occurrences with `ul.pager`, which is still consistent with the source page. Hopefully it's clearer now, let me know what you think.

Screenshot at 2020-01-22 10-41-14

PR #4058, opened 2019-10-02
Added error message in stats if exception is thrown in start_requests of spider

### LINKED ISSUE #3463: raise CloseSpider("Error Message") is not saving error message

amardeep24:
Fixes #3463

This enhancement aims to capture the error message from the `engine._next_request` method and add it in the stats property `finish_reason`. The previous behavior was to always populate `finished` in the `finish_reason` property when the Exception was raised in the `start_requests ` method of the spider.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -132,7 +132,7 @@ class ExecutionEngine(object):
                 slot.start_requests = None
                 logger.error('Error while obtaining start requests',
                              exc_info=True, extra={'spider': spider})
-                reason = str(exception)
+                reason = repr(exception)
             else:
                 self.crawl(request, spider)
 
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -112,6 +112,7 @@ class ExecutionEngine(object):
 
     def _next_request(self, spider):
         slot = self.slot
+        reason = "finished"
         if not slot:
             return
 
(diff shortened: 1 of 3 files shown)


amardeep24:
@Gallaecio please review! :smiley:

scrapy/core/engine.py
```diff
@@ -127,15 +128,16 @@ def _next_request(self, spider):
                 request = next(slot.start_requests)
             except StopIteration:
                 slot.start_requests = None
-            except Exception:
+            except Exception as exception:
                 slot.start_requests = None
                 logger.error('Error while obtaining start requests',
                              exc_info=True, extra={'spider': spider})
+                reason = str(exception)
```
Gallaecio:
```suggestion
                reason = repr(exception)
```

https://youtu.be/V2fGAv2R5j8?t=9m2s

Gallaecio (commented):
Could you add a test for this?

scrapy/core/engine.py
amardeep24: @Gallaecio sure, working on it

amardeep24:
@Gallaecio Please review, I added a test case.

tests/test_engine.py
```diff
@@ -296,7 +303,16 @@ def test_close_engine_spiders_downloader(self):
...
         self.assertEqual(len(e.open_spiders), 0)
-
+    
+    @defer.inlineCallbacks
+    def test_crawler_with_error_in_start_requests(self):
+         self.run = CrawlerRun(ErrorInStartRequestsSpider)
+         yield self.run.run()
+         print(type(self.run.crawler.stats.get_stats()['finish_reason']))
+         # Check both as the latter one is produced for Python < 3.7
+         # https://bugs.python.org/issue30399
```
Gallaecio: You could check the run-time Python version and use one check or the other depending on that.

Gallaecio (approved):
I’ve left a minor comment, but overall it looks good to me. Thanks!

Gallaecio:
I’m now thinking that maybe we should only add the exception class name to the stats. I’m guessing the stat breadcrumbs are not meant to have `/` characters, and we are opening up that possibility here.

PR #4061, opened 2019-10-05
Fix case of GitHub.

jbampton:


kmike:
Thanks @jbampton!

jbampton:
Nice one @kmike :)

Issue #3553, opened 2018-12-28
Genspider prepend `http://` without checking it in the `domain`


ThunderMind2019:
genspider prepend http:// But when i enter address like https://example.com it becomes http://https://example.com that, when run scrapy crawl throws an error.
What it should do, it should first check the receiving domain than take decision according to the passing domain whether it needs a http:// or nothing.

ThunderMind2019:
i have corrected that, grant me permissions to upload my branch and to create a PR.

ThunderMind2019:
#3555 @kmike

ambarmendez:
Hello everybody! I would like to contribute to Scrapy, specifically with this issue and I already fork the repo and made the changes but before issuing a pull request I would like to discuss it and reach to an agreement.

I made the changes over 5 files:
- **scrapy/commands/genspider.py**. Where `urllib.parse.urlparse` is being used on the domain parameter and prepend `http` scheme in the case is not there. But, this is a very tricky situation because of the command syntax, which asks for a domain and not an URL, so we're just putting a very big bet that it should be HTTP but what about if we ask for a URL instead of a domain, then the contradiction is solved right away since we can extract the domain without making any assumption. Don't you think fellas?
- **Each template file**. In order to receive domain and URL values.
I checked the test just for tests/test_commands.py, I didn't add test because I don't know how to make a test for this particular scenario where two attributes are set based on user input on a class created through a template. If someone has an idea how it could be made please I'm all ears

Lastly and maybe more important, I don't know how to try the code it's my first time and I don't have any idea how to execute it. So any help or guide would be really appreciated.

Thank you in advance and happy coding!!!! =D

Gallaecio:
I would suggest you open a pull request already, discussions over actual code are usually easier.

Also, make sure you check out previous pull requests (#3554, #3558). Feedback there, specially @kmike’s, could answer some of your points.

ambarmendez:
Thank you @Gallaecio! Let's see how it goes =D

Gallaecio:
Now URLs are supported as input: https://github.com/scrapy/scrapy/pull/4439

However, we still simply extract the domain from them. Which means that the input protocol is not respected for `start_urls`, and instead it is always forced to be HTTP (as opposed to, for example, HTTPS).

Gallaecio:
[...]

PR #4064, opened 2019-10-06
Fix[#3553] Add http by checking scheme in domain in genspider

ambarmendez:
In **genspider**, `urllib.parse.urlparse` is being used on the domain parameter and prepend 'http' scheme in the case is not there. But, this is a very tricky situation because of the command syntax, which asks for a domain and not an URL, so we're just putting a very big bet that it should be HTTP but what about if we ask for a URL instead of a domain, then the contradiction is solved right away since we can extract the domain without making any assumption. Don't you think fellas?
Lastly, **each template file**, in order to receive domain and URL values.
I checked the test just for tests/test_commands.py, I didn't add test because I don't know how to make a test for this particular scenario where two attributes are set based on user input on a class created through a template. If someone has an idea how it could be made please I'm all ears

### DIFF
--- a/scrapy/commands/genspider.py
+++ b/scrapy/commands/genspider.py
@@ -28,7 +28,7 @@ def sanitize_module_name(module_name):
 class Command(ScrapyCommand):
 
     requires_project = False
-    default_settings = {'LOG_ENABLED': True}
+    default_settings = {'LOG_ENABLED': False}
 
     def syntax(self):
         return "[options] <name> <domain|url>"
--- a/scrapy/commands/genspider.py
+++ b/scrapy/commands/genspider.py
@@ -5,7 +5,7 @@ import string
 
 from importlib import import_module
 from os.path import join, dirname, abspath, exists, splitext
-from urllib.parse import urlparse
+from six.moves.urllib.parse import urlparse
 
 import scrapy
 from scrapy.commands import ScrapyCommand
@@ -65,9 +65,8 @@ class Command(ScrapyCommand):
 
         parsed_url = urlparse(domain)
 
-        # Automatically it is prepended 'http' if the scheme is missing
         if not parsed_url.scheme:
-            parsed_url = urlparse('http://' + domain)
+            parsed_url._replace(scheme='http')
 
         module = sanitize_module_name(name)
 
--- a/scrapy/commands/genspider.py
+++ b/scrapy/commands/genspider.py
@@ -5,6 +5,7 @@ import string
 
 from importlib import import_module
 from os.path import join, dirname, abspath, exists, splitext
+from urllib.parse import urlparse
 
 import scrapy
 from scrapy.commands import ScrapyCommand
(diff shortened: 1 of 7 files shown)


scrapy/commands/genspider.py
```diff
@@ -5,6 +5,7 @@
 
 from importlib import import_module
 from os.path import join, dirname, abspath, exists, splitext
+from urllib.parse import urlparse
```
elacuesta:
```suggestion
from six.moves.urllib.parse import urlparse
```
Tests are failing in py2 because `urllib.parse` is py3-only.

scrapy/commands/genspider.py
```diff
@@ -61,6 +62,13 @@ def run(self, args, opts):
             raise UsageError()
 
         name, domain = args[0:2]
+
+        parsed_url = urlparse(domain)
+
+        # Automatically it is prepended 'http' if the scheme is missing
```
elacuesta: I think this comment is unnecessary, this particular action is done in one line and it should be self-explanatory.

scrapy/commands/genspider.py
```diff
@@ -61,6 +62,13 @@ def run(self, args, opts):
...
+        if not parsed_url.scheme:
+            parsed_url = urlparse('http://' + domain)
```
elacuesta: Since you already have a named tuple with the parsed URL, you can use the [`_replace` method](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace) to update the schema (see also https://docs.python.org/3/library/urllib.parse.html?highlight=_replace#urllib.parse.urlparse)

ambarmendez: For what I understand, I accept your changes but still, I need to add 2 more: to remove the unnecessary comment and to use the `_replace` method to update the schema. Sorry If I am being repetitive but it's a little bit confusing what do I have to do after your reviewing.

ambarmendez:
Sorry! :worried: Just a little bit embarrassed over here.
As you, @elacuesta can see.  I am still confused about what to do next. But still committed to helping and eager to learn how to do it right. 
Again thank you for your feedback! :grin:

scrapy/commands/genspider.py
```diff
@@ -61,6 +62,12 @@ def run(self, args, opts):
             raise UsageError()
 
         name, domain = args[0:2]
+
+        parsed_url = urlparse(domain)
+
+        if not parsed_url.scheme:
+            parsed_url._replace(scheme='http')
```
elacuesta:
```suggestion
            parsed_url = parsed_url._replace(scheme='http')
```
This is because `_replace` does not modify the object in-place, it returns a new `ParseResult` instead.

scrapy/commands/genspider.py
```diff
@@ -61,6 +62,12 @@ def run(self, args, opts):
             raise UsageError()
 
         name, domain = args[0:2]
+
+        parsed_url = urlparse(domain)
```
elacuesta:
```suggestion
        parsed_url = urlparse(domain.rstrip('/'))
```
Perhaps not strictly necessary, but if I'm not mistaken an URL that ends with a slash would result in two slashes when it's used later in the template (`$url/feed.xml` -> `https://example.org//feed.xml` if `$url == https://example.org/`)

Gallaecio:
@kmike @elacuesta I wonder if, instead of simply allowing the parameter to be either a domain or a URL, we should deprecate domains (do not mention them in the documentation or command-line help, and if used work as up-until-know but logging/printing a deprecation warning). I think allowing a domain may have been a small design flaw in the first place.

wRAR:
Fixed in #5808

PR #4065, opened 2019-10-07
reorder default console shell

NewUserHa:
according to https://docs.scrapy.org/en/latest/topics/shell.html#configuring-the-shell, make `highly recommend` ipython as the first choice!

### DIFF
--- a/scrapy/utils/console.py
+++ b/scrapy/utils/console.py
@@ -56,8 +56,8 @@ def _embed_standard_shell(namespace={}, banner=''):
     return wrapper
 
 DEFAULT_PYTHON_SHELLS = OrderedDict([
-    ('ptpython', _embed_ptpython_shell),
     ('ipython', _embed_ipython_shell),
+    ('ptpython', _embed_ptpython_shell),
     ('bpython', _embed_bpython_shell),
     ('python', _embed_standard_shell),
 ])


kmike:
A good catch!

kmike (approved):

NewUserHa:
https://github.com/prompt-toolkit/ptpython/issues/241

Granitosaurus:
> @Granitosaurus Given [1a452c0](https://github.com/scrapy/scrapy/commit/1a452c038cc0547924051a7cd0786215ec7c2104), I would love to have your feedback here.

Hey, since ptpython can wrap around ipython I don't see why it shouldn't have higher priority, see https://github.com/prompt-toolkit/ptpython#ipython-support

I'd vote for keeping ptpython at the top for this reason.

NewUserHa:
1. documents recommend `ipython`
2. ptpython has an issue with `scrapy shell` whose comments were posted above.
3. ptipython has flash/lagging after input.
4. ptpython as the first ≠ ptipython as the first ≠ ipython as the first

Gallaecio:
OK, so I’ve just tested all 4 prompts.

prompt-toolkit/ptpython#241 is indeed a problem for users. [It also affects `ipython`](https://stackoverflow.com/questions/49442523/prevent-showing-debugging-log-info-inside-ipython-shell), however, so putting `ipython` first is not the best choice. It does not affect `bpython`.

I believe the issue is not of the shells, though, but of Scrapy, to some extent. They expect the log level to be higher than DEBUG, and Scrapy sets it to DEBUG. And it has an easy workaround: adding ``-L INFO`` to the ``scrapy shell`` command.

A quick solution in the line of the initial proposed solution would be to make `bpython` the first choice. But honestly, I don’t think the order matters that much, most people will only install the one they want to use.

A good solution would go in the line of changing the logging level of the logger that Scrapy initializes. But I’m not sure if we should change it only for shells that we know are affected, for ``scrapy shell`` as a whole, or what.

Suggestions?

Granitosaurus:
I agree I think the logging issue should be addressed separetly and is really unrelated to default embed shell order. 



> A good solution would go in the line of changing the logging level of the logger that Scrapy initializes. But I’m not sure if we should change it only for shells that we know are affected, for scrapy shell as a whole, or what. Suggestions?

I think changing default log level to INFO/WARNING for `shell` command would be a good start since the only logging complaint we have here is that `parso` throws up a bunch of logging on ever key press if scrapy sets global log level below INFO.   
Right now easy solution is to just disable it on command (e.g. `scrapy shell --nolog`), so this really isn't a critical issue IMHO. It's often good idea to do that either way as some scrapy's pipelines/middlewares are idiomatic to output status messages every N seconds (e.g. `crawled 100 items at 10items/minute`) - these tend to just be annoying in `scrapy shell`.

PR #4068, opened 2019-10-10
Update exceptions.rst

AlekseiMarinichenko:
response.body is a bytes value, but "Bandwidth exceeded" is str. 
You can't mix the types because "TypeError: a bytes-like object is required, not 'str'" will be raised.

Use bytes literal:
if b'Bandwidth exceeded' in response.body:

### DIFF
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -36,7 +36,7 @@ CloseSpider
 For example::
 
     def parse_page(self, response):
-        if 'Bandwidth exceeded' in response.body:
+        if b'Bandwidth exceeded' in response.body:
             raise CloseSpider('bandwidth_exceeded')
 
 DontCloseSpider
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -36,7 +36,7 @@ CloseSpider
 For example::
 
     def parse_page(self, response):
-        if b'Bandwidth exceeded' in response.body:
+        if 'Bandwidth exceeded' in re  sponse.text:
             raise CloseSpider('bandwidth_exceeded')
 
 DontCloseSpider
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -36,7 +36,7 @@ CloseSpider
 For example::
 
     def parse_page(self, response):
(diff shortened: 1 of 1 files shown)


Gallaecio:
I think we can ignore the tests here, given the change only affects documentation. This can only be applied once we drop Python 2 support, though.

kmike:
@Gallaecio why won't it work in Python 2?

FTR, this is a PR against 1.7 branch, not for master.

docs/topics/exceptions.rst
```diff
@@ -36,7 +36,7 @@ CloseSpider
 For example::
 
     def parse_page(self, response):
-        if 'Bandwidth exceeded' in response.body:
+        if b'Bandwidth exceeded' in response.body:
```
VMRuiz: Working with bytes could be difficult, I think it's better if we just use `str` :

```suggestion
        if 'Bandwidth exceeded' in response.text:
```

Gallaecio: > @Gallaecio why won't it work in Python 2?

I had no idea that the `b''` syntax was supported in Python 2. It indeed works.

Gallaecio:
> @Gallaecio why won't it work in Python 2?

I had no idea that the `b''` syntax was supported in Python 2. It indeed works.

docs/topics/exceptions.rst
Gallaecio: I’m not against this suggestion, but I don’t see why working with bytes would be a problem here.

VMRuiz: It becomes an issue when you are working with non-English websites that uses different encoding.

I believe docs should set good example and use strings when looking for text string without trusting the underlying binary representation.

Gallaecio:
@AlekseiMarinichenko What do you think about going with @VMRuiz’s [suggestion](https://github.com/scrapy/scrapy/pull/4068#discussion_r334785039)?

AlekseiMarinichenko:
I think that @VMRuiz suggestion has to be approved.

Gallaecio:
@elacuesta If you are also OK with the change, I think we can squash and merge.

elacuesta:
Yes, `response.text` looks better to me, thanks for the update.

PR #4069, opened 2019-10-10
update zsh completion

seregaxvm:
Features:
- full option completion
- automatic project detection
- spider name completion


Gallaecio:
@Dineshs91 I know it’s been a long time since you contributed the original zsh completion, but is there any chance that you can have a look here?

PR #4074, opened 2019-10-13
Added Pathlib.Path support: Issue #3731

### LINKED ISSUE #3731: Support pathlib.Path in FEED_URI

purvaudai:
Fixes #3731


Gallaecio:
I believe the proposed fix #3753 is slightly cleaner.

Also, both patches are missing an accompanying test that covers the change. Could you add one?

purvaudai:
I am confused on how and where to add a test. Could you please guide a bit?

Gallaecio:
Have a look at the tests in https://github.com/scrapy/scrapy/blob/master/tests/test_feedexport.py, see if you can get some ideas. Otherwise, I’ll see if I can take a closer look myself and give you some better feedback.

purvaudai:
I did look into that file. However, I could not even find the class which I had made changes to (FeedExporter) being importted into the test file.

Gallaecio:
Forget about the class, navigating the internals of Scrapy can be daunting at first. Instead, focus on the way users would take advantage of your changes: the `FEED_URI` setting. You just need to test that setting `FEED_URI` to a `Path` works the same as using a string.

[Existing tests get a temporary folder automatically created for them and assigned to `FEED_URI`](https://github.com/scrapy/scrapy/blob/master/tests/test_feedexport.py#L406). You can create a test that uses a custom `FEED_URI` value instead.

For example, based on [`test_export_no_items_store_empty`](https://github.com/scrapy/scrapy/blob/master/tests/test_feedexport.py#L552) and the temporary folder creation I’ve just mentioned, you could write something in the lines of:

```python
@defer.inlineCallbacks
def test_pathlib_uri(self):
    tmpdir = tempfile.mkdtemp()
    feed_uri = Path(tmpdir) / 'res'
    settings = {
        'FEED_FORMAT': 'csv',
        'FEED_URI': feed_uri,
        'FEED_STORE_EMPTY': True,
    }
    data = yield self.exported_no_data(settings)
    self.assertEqual(data, b'')
    shutil.rmtree(tmpdir, ignore_errors=True)
```

Gallaecio:
I think it’s OK not to support Python 2, as we will drop support soon, right after releasing the upcoming 1.8 version of Scrapy.

purvaudai:
I do not understand why the Travis CI build is not triggering for the last 2 commits. Am I doing anything wrong?

Gallaecio:
Your changes are currently in conflict with the `master` branch. Until [you resolve them](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-on-github), the tests cannot run.

purvaudai:
@Gallaecio 
Oh, I see. Now that the Travis-CI build has passed, what else is required now?

Gallaecio:
Now people will review your changes. We usually merge a pull request once two core developers have approved it.

requirements-py3.txt
```diff
@@ -16,3 +16,4 @@ lxml>=3.5.0
 service_identity>=16.0.0
 six>=1.10.0
 zope.interface>=4.1.3
+pathlib2>=2.0
```
Gallaecio: This should not be necessary

tests/test_feedexport.py
```diff
@@ -29,6 +29,8 @@
 from scrapy.utils.python import to_native_str
 from scrapy.utils.project import get_project_settings
 
+from pathlib2 import Path
```
Gallaecio: You should use `pathlib.Path` from the standard library, not from this third-party package.

tests/test_feedexport.py
```diff
@@ -843,3 +846,17 @@ def test_init_exporters_storages_with_crawler(self):
...
         self.assertTrue(FromCrawlerFileFeedStorage.init_with_crawler)
+
+    @defer.inlineCallbacks
+    def test_pathlib_uri(self):
+        tmpdir = tempfile.mkdtemp()
+        feed_uri = Path(tmpdir) / 'res'
+        settings = {
+            'FEED_FORMAT': 'csv',
+            'FEED_STORE_EMPTY': True,
+            'FEED_URI': feed_uri,
```
Gallaecio: Shouldn’t this be a file path instead of a folder path?

purvaudai: According to previous tests, this is the manner in which they were done. I am unable to understand what you mean.
Thefeed_uri, from my understanding, is the file path, which the spider writes to.

Gallaecio: Nevermind, my bad.

PR #4077, opened 2019-10-14
[WIP] Add S3MultipartFeedStorage and S3Writer

victor-torres:
This Feed Storage writes items to the S3 Bucket during crawler execution.
It doesn't wait for the spider to be completed to start exporting items.
Therefore, we can save disk space writing data directly to Amazon S3 using
a small amount of RAM to buffer some data (50 MiB default).

I'd like to have some feedback about this feature and maybe additional suggestions.

TODO:

- [ ] add documentation
- [ ] add test cases

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -4,6 +4,7 @@ Feed Exports extension
 See documentation in docs/topics/feed-exports.rst
 """
 
+import io
 import os
 import sys
 import logging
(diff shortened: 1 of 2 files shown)


victor-torres:
Relates to #1559

scrapy/extensions/feedexport.py
```diff
@@ -155,6 +156,39 @@ def _store_in_thread(self, file):
...
 
+class S3MultipartFeedStorage(S3FeedStorage):
+    """S3 Feed Storage that uses the Multipart Upload feature to export items.
+
+    This Feed Storage writes items to the S3 Bucket during crawler execution.
+    It doesn't wait for the spider to be completed to start exporting items.
+    Therefore, we can save disk space writing data directly to Amazon S3 using
+    a small amount of RAM to buffer some data (50 MiB default).
+    """
+    DEFAULT_BUFFER_SIZE = 52428800  # 50 MiB
```
Gallaecio: I think this should be configurable, à la `FEED_STORAGE_S3_ACL`.

PR #4088, opened 2019-10-21
docs: use __init__ method instead of constructor

### LINKED ISSUE #4086: constructor, initializer → __init__ method
There are several places in the documentation and comments where we use the words ‘constructor’ or ‘initializer’ to refer to the `__init__` method of a class. We should change the wording to explicitly say ‘`__init__` method’ instead.

Related to #2354.

ammarnajjar:
Resolves #4086

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -494,7 +494,7 @@ ItemLoader objects
     .. attribute:: default_item_class
 
         An Item class (or factory), used to instantiate items when not given in
-        the `__init__` method
+        the ``__init__`` method
 
     .. attribute:: default_input_processor
 
@@ -509,15 +509,15 @@ ItemLoader objects
     .. attribute:: default_selector_class
 
         The class used to construct the :attr:`selector` of this
-        :class:`ItemLoader`, if only a response is given in the `__init__` method
-        If a selector is given in the `__init__` method this attribute is ignored.
+        :class:`ItemLoader`, if only a response is given in the ``__init__`` method
+        If a selector is given in the ``__init__`` method this attribute is ignored.
         This attribute is sometimes overridden in subclasses.
 
     .. attribute:: selector
 
         The :class:`~scrapy.selector.Selector` object to extract data from.
-        It's either the selector given in the `__init__` methodor one created from
-        the response given in the `__init__` methodusing the
+        It's either the selector given in the ``__init__`` methodor one created from
+        the response given in the ``__init__`` methodusing the
         :attr:`default_selector_class`. This attribute is meant to be
         read-only.
 
@@ -642,7 +642,7 @@ Here is a list of all built-in processors:
 .. class:: Identity
 
     The simplest processor, which doesn't do anything. It returns the original
-    values unchanged. It doesn't receive any `__init__` method arguments, nor does it
+    values unchanged. It doesn't receive any ``__init__`` method arguments, nor does it
     accept Loader contexts.
 
     Example::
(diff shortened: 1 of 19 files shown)


Gallaecio:
You have successfully used double ` in many documentation changes, but some changes use a single ` instead, which is incorrect (it applies italics).

ammarnajjar:
thanks for the review, I hope I got it right this time 😅

docs/topics/loaders.rst
```diff
@@ -265,7 +265,7 @@ There are several ways to modify Item Loader context values:
       loader.context['unit'] = 'cm'
 
 2. On Item Loader instantiation (the keyword arguments of Item Loader
-   constructor are stored in the Item Loader context)::
+   ``__init__`` methodare stored in the Item Loader context)::
```
Gallaecio:
```suggestion
   ``__init__`` method are stored in the Item Loader context)::
```

I’ve seen this issue in a few more replacements below, where the character right after ‘method’ (a space, a dot, a comma) was accidentally removed. Could you please take a look?

ammarnajjar: 🤦‍♂ of course

ammarnajjar: @Gallaecio thanks for your patience, change pushed.

tests/test_utils_misc/__init__.py
```diff
@@ -109,11 +109,11 @@ def _test_with_crawler(mock, settings, crawler):
...
-        # Check usage of correct constructor using four mocks:
-        #   1. with no alternative constructors
-        #   2. with from_settings() constructor
-        #   3. with from_crawler() constructor
-        #   4. with from_settings() and from_crawler() constructor
+        # Check usage of correct __init__ method using four mocks:
+        #   1. with no alternative __init__ methods
+        #   2. with from_settings() __init__ method
+        #   3. with from_crawler() __init__ method
+        #   4. with from_settings() and from_crawler() __init__ method
```
kmike: I think here "constructor" is replaced with `__init__` incorrectly; "constructor" here is a single word for from_crawler/from_settings/etc., not as a synonym to `__init__`.

ammarnajjar: so should one leave it as is "constructor" or use “factory method” instead as @Gallaecio  mentioned in https://github.com/scrapy/scrapy/pull/4088#pullrequestreview-305189732?

kmike: Yeah, I think so; no strong opinion about constructor/factory methods, I was always thinking about them as of "alternative constructors", so comment before the change as-is looks fine to me.

ammarnajjar: @kmike thanks for the review, 7a84a4b is pushed.

scrapy/extensions/feedexport.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, uri, access_key=None, secret_key=None, acl=None):
                 warnings.warn(
                     "Initialising `scrapy.extensions.feedexport.S3FeedStorage` "
                     "without AWS keys is deprecated. Please supply credentials or "
-                    "use the `from_crawler()` constructor.",
+                    "use the `from_crawler()` ``__init__`` method.",
```
kmike: it should be kept as "constructor" here

sep/sep-009.rst
```diff
@@ -55,7 +55,7 @@ singletons members of that object, as explained below:
      ``STATS_CLASS`` setting)
    - **crawler.log**: Logger class with methods replacing the current
      ``scrapy.log`` functions. Logging would be started (if enabled) on
-     ``Crawler`` constructor, so no log starting functions are required.
+     ``Crawler`` ``__init__`` method, so no log starting functions are required.
```
kmike:
```suggestion
     ``Crawler`` instantiation, so no log starting functions are required.
```

PR #4090, opened 2019-10-21
Fix references to Python types in parameter type fields

Gallaecio:
The values of these fields are automatically interpreted as Python types, and Sphinx tries to link them to the Python documentation, so they need to be valid types.

Additional details (e.g. that a list is a list of strings) must be provided in the field description instead.


kmike:
> Additional details (e.g. that a list is a list of strings) must be provided in the field description instead.

Would you mind filling them in the same PR?

Gallaecio:
Done for the two parameters where information was lost in the process. Also switched to `object` as a replacement for `any`, instead of removing the parameter type.

docs/topics/request-response.rst
```diff
@@ -54,7 +54,7 @@ Request objects
       ``body`` is not given, an empty string is stored. Regardless of the
       type of this argument, the final value stored will be a ``str`` (never
       ``unicode`` or ``None``).
-    :type body: str or unicode
+    :type body: str
```
kmike: technically, this is no longer correct in Python 2, but let's not worry about it

scrapy/settings/__init__.py
```diff
@@ -245,14 +245,14 @@ def set(self, name, value, priority='project'):
         otherwise they won't have any effect.
 
         :param name: the setting name
-        :type name: string
+        :type name: str
 
         :param value: the value to associate with the setting
-        :type value: any
+        :type default: object
```
kmike, line 251: shouldn't it be `type value: object`?

Gallaecio: :facepalm:

docs/topics/request-response.rst
```diff
@@ -56,7 +56,7 @@ Request objects
...
```
kmike, line 59: I think this is incorrect; docs above are also incorect. There is no `unicode` in Python 3, and request.body is bytes, not str.

docs/topics/request-response.rst
```diff
@@ -107,7 +107,7 @@ Request objects
     :param encoding: the encoding of this request (defaults to ``'utf-8'``).
        This encoding will be used to percent-encode the URL and to convert the
        body to ``str`` (if given as ``unicode``).
```
kmike, line 109: request body is converted to bytes, not to str

docs/topics/request-response.rst
```diff
@@ -772,7 +772,7 @@ TextResponse objects
        body, it will be encoded using this encoding (remember the body attribute
        is always a string). If ``encoding`` is ``None`` (default value), the
```
kmike, line 773: this is not correct: body attribute is always bytes, not string

kmike:
Hey @Gallaecio! The changes look good, but it seems quite a lot of docs in Request & Response are incorrect, as far as str vs bytes is concerned; this makes the result of some of the changes incorrect as well. 

Do you think this can be cleaned up as a part of this PR? This is a completely different issue, but if we won't fix it, we'll be merging incorrect docs - e.g. you're changing "str or unicode" to "str", while it is "bytes", but writing "bytes" doesn't make sense as-is, because docs above contradict that.

docs/topics/request-response.rst
```diff
@@ -42,21 +42,19 @@ Request objects
...
 
-    :param body: the request body. If a ``unicode`` is passed, then it's encoded to
-      ``str`` using the ``encoding`` passed (which defaults to ``utf-8``). If
-      ``body`` is not given, an empty string is stored. Regardless of the
-      type of this argument, the final value stored will be a ``str`` (never
-      ``unicode`` or ``None``).
-    :type body: str or unicode
+    :param body: the request body. If a string is passed, it is converted to
+        bytes using *encoding*, which defaults to ``utf-8``. If not passed or
+        ``None`` is passed, an empty bytes array is stored.
```
kmike, line 56:
```suggestion
        ``None`` is passed, an empty bytes object is stored.
```
I think "bytes array" could be confussed with built-in bytearray.

docs/topics/request-response.rst
```diff
@@ -768,27 +766,27 @@ TextResponse objects
...
        The same as ``response.body.decode(response.encoding)``, but the
        result is cached after the first call, so you can access
        ``response.text`` multiple times without extra overhead.
 
        .. note::
 
-            ``unicode(response.body)`` is not a correct way to convert response
-            body to unicode: you would be using the system default encoding
+            ``str(response.body)`` is not a correct way to convert the response
+            body into a string: you would be using the system default encoding
```
kmike, line 789: it is incorect for a diifferent reason now:

```py
In [1]: str(b'foo')
Out[1]: "b'foo'"
```

Gallaecio: :open_mouth:

kmike:
Thanks @Gallaecio!

PR #4091, opened 2019-10-21
Initial Python 2 removal

wRAR:
This pull request, split out from #3983, removes Python 2 support from documentation and basic stuff like CI.

This is #3994 but made as a local branch instead of a fork. It's also rebased to the current master.


.travis.yml
```diff
@@ -7,14 +7,6 @@ branches:
...
 matrix:
   include:
-    - env: TOXENV=py27
-      python: 2.7
-    - env: TOXENV=py27-pinned
-      python: 2.7
-    - env: TOXENV=py27-extra-deps
-      python: 2.7
-    - env: TOXENV=pypy
-      python: 2.7
```
wRAR: The deploy line still needs feedback @Gallaecio @kmike

Gallaecio: If you mean line 63 below, I think we just need to change `$TOXENV == py27` to `$TOXENV == py37`.

wRAR: done

Gallaecio:
This needs a `deps` line removal because of Python 3.8 support being added to `master`.

docs/topics/feed-exports.rst
```diff
@@ -99,12 +99,12 @@ The storages backends supported out of the box are:
 
  * :ref:`topics-feed-storage-fs`
  * :ref:`topics-feed-storage-ftp`
- * :ref:`topics-feed-storage-s3` (requires botocore_ or boto_)
+ * :ref:`topics-feed-storage-s3` (requires botocore_)
  * :ref:`topics-feed-storage-stdout`
 
 Some storage backends may be unavailable if the required external libraries are
 not available. For example, the S3 backend is only available if the botocore_
-or boto_ library is installed (Scrapy supports boto_ only on Python 2).
```
kmike: should we remove boto support from the code as well (sorry, probably this is already handled elswhere)?

.travis.yml
```diff
@@ -7,14 +7,6 @@ branches:
     - /^\d\.\d+\.\d+(rc\d+|\.dev\d+)?$/
 matrix:
   include:
-    - env: TOXENV=py27
-      python: 2.7
-    - env: TOXENV=py27-pinned
-      python: 2.7
-    - env: TOXENV=py27-extra-deps
-      python: 2.7
-    - env: TOXENV=pypy
```
kmike: in `install` section there is code related to pypy

docs/topics/feed-exports.rst
wRAR: There is #1866, we need to decide on the solution

wRAR: (see my comment)

.travis.yml
wRAR: @kmike indeed, what should we do about that section? Just remove the `"$TOXENV" = "pypy"` part?

Gallaecio: Remove the whole `if … fi` section.

Gallaecio: You could also remove the `pypy3` part, and use `python: pypy3` in the matrix (see https://github.com/scrapy/w3lib/pull/140)

wRAR: #4116

PR #4092, opened 2019-10-21
Add Python 3.8 official support

### LINKED ISSUE #4085: Make Python 3.8 support official
This involves, among other things:
- Enabling a Python 3.8 environment in Tox
- Enabling a Python 3.8 job in the Travis CI
- Updating the `setup.py` file

further-reading:
Fixes #4085

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,9 @@ matrix:
       python: 3.6
     - env: TOXENV=py37
       python: 3.7
-    - env: TOXENV=py37-extra-deps
-      python: 3.7
     - env: TOXENV=py38
       python: 3.8
-    - env: TOXENV=py37-extra-deps
+    - env: TOXENV=py38-extra-deps
       python: 3.8
     - env: TOXENV=docs
       python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ matrix:
       python: 3.7
     - env: TOXENV=py37-extra-deps
       python: 3.7
+    - env: TOXENV=py38
+      python: 3.8
+    - env: TOXENV=py37-extra-deps
+      python: 3.8
     - env: TOXENV=docs
       python: 3.6
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       python: 3.7
     - env: TOXENV=py38
       python: 3.8
-    - env: TOXENV=py37-extra-deps
+    - env: TOXENV=py38-extra-deps
       python: 3.8
     - env: TOXENV=docs
       python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       python: 3.7
     - env: TOXENV=py38
       python: 3.8
-    - env: TOXENV=py38-extra-deps
+    - env: TOXENV=py37-extra-deps
       python: 3.8
     - env: TOXENV=docs
       python: 3.6
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
-        # 'Programming Language :: Python :: 3.8',  not supported yet
+        'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
         'Topic :: Internet :: WWW/HTTP',
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
-        'Programming Language :: Python :: 3.8',
+        # 'Programming Language :: Python :: 3.8',  not supported yet
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
         'Topic :: Internet :: WWW/HTTP',
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
         'Topic :: Internet :: WWW/HTTP',
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -154,6 +154,7 @@ class FilesystemStorageGzipTest(FilesystemStorageTest):
         new_settings.setdefault('HTTPCACHE_GZIP', True)
         return super(FilesystemStorageTest, self)._get_settings(**new_settings)
 
+@pytest.mark.xfail(reason='leveldb not supported in python 3.8')
 class LeveldbStorageTest(DefaultStorageTest):
 
     pytest.importorskip('leveldb')
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -6,6 +6,7 @@ import unittest
 import email.utils
 from contextlib import contextmanager
 import pytest
+import sys
 
 from scrapy.http import Response, HtmlResponse, Request
 from scrapy.spiders import Spider
@@ -160,7 +161,7 @@ class LeveldbStorageTest(DefaultStorageTest):
         pytest.importorskip('leveldb')
     except SystemError:
         # Happens in python 3.8
-        pytest.skip("'SystemError: bad call flags' occurs on Python 3.8")
+        pytestmark = pytest.skip("'SystemError: bad call flags' occurs on Python 3.8")
     storage_class = 'scrapy.extensions.httpcache.LeveldbCacheStorage'
 
 
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -6,6 +6,7 @@ import unittest
 import email.utils
 from contextlib import contextmanager
 import pytest
+import sys
 
 from scrapy.http import Response, HtmlResponse, Request
 from scrapy.spiders import Spider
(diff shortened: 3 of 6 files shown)


.travis.yml
```diff
@@ -27,6 +27,10 @@ matrix:
       python: 3.7
     - env: TOXENV=py37-extra-deps
       python: 3.7
+    - env: TOXENV=py38
+      python: 3.8
+    - env: TOXENV=py37-extra-deps
```
kmike:
```suggestion
    - env: TOXENV=py38-extra-deps
```

kmike: Actually, I think that's fine to have a single Python 3.x -extra-deps environment, to save CI time; 3.7 or 3.8-based sound fine to me.

wRAR:
```
==================================== ERRORS ====================================
________ ERROR collecting tests/test_downloadermiddleware_httpcache.py _________
tests/test_downloadermiddleware_httpcache.py:157: in <module>
    class LeveldbStorageTest(DefaultStorageTest):
tests/test_downloadermiddleware_httpcache.py:159: in LeveldbStorageTest
    pytest.importorskip('leveldb')
E   SystemError: bad call flags
________ ERROR collecting tests/test_downloadermiddleware_httpcache.py _________
tests/test_downloadermiddleware_httpcache.py:157: in <module>
    class LeveldbStorageTest(DefaultStorageTest):
tests/test_downloadermiddleware_httpcache.py:159: in LeveldbStorageTest
    pytest.importorskip('leveldb')
E   SystemError: bad call flags
```

Hmm?

wRAR:
So it looks like the latest leveldb (released in 2016) doesn't support Python 3.8, here is a patch in Fedora (which we obviously cannot use): https://src.fedoraproject.org/rpms/python-leveldb/c/57adbb30b4d4c1ee58976eaa8ca5b3cea9c516ab?branch=master

Scrapy uses leveldb for `scrapy.extensions.httpcache.LeveldbCacheStorage`. The module itself is only in the test requirements, the docs tell the user to install the module manually to use `LeveldbCacheStorage`. 

Maybe we should skip the test for 3.8, at least for now, and document the problem at least in the test sources.

kmike:
Hey! I think it'd be nice to fix it before declaring Python 3.8 support. Switch to another leveldb wrapper, e.g. https://github.com/wbolster/plyvel? The advantage of the old package is that it bundles leveldb (so it is easier to install), but it seems plyvel bundles leveldb as well (though not always) since v1.0.1:

> Provide binary packages (manylinux1 wheels) for Linux.
These wheel packages have the LevelDB library embedded.

Another option would be to clone a repo to scrapy organization, fix an issue there, then release - we've done this for https://github.com/scrapy/pypydispatcher to get pypy support for a less maintained package. Though I'd just switch.

In this PR, to make it bounded, we can 

* mark test as xfail (run=False), 
* keep Python 3.8 on CI
* remove one of the -extra-deps environments
* remove Python 3.8 from setup.py for now?

further-reading:
Thanks a lot for the help! I've come upon another issue where trying to run tox.ini is getting an error trying to load Twisted.

```
ERROR: Could not find a version that satisfies the requirement Twisted!=18.4.0 (from -c tests/constraints.txt (line 1)) (from versions: none)
ERROR: No matching distribution found for Twisted!=18.4.0 (from -c tests/constraints.txt (line 1))
```

I'll have a deeper look into that and your other feedback as soon as I can!

wRAR:
The Debian leveldb maintainer just replied to the Py3.8 bug that they don't want leveldb in Debian anymore and prefer Plyvel instead.

further-reading:
Hey! I've added xfails now to handle the issue with leveldb in python 3.8.

To give some additional info - the issue is likely similar to this one https://bugs.python.org/msg348804 which is caused by a python 3.8 change to the C-API that generates `SystemError: bad call flags` when some modules are imported. To incorporate skipping properly I added a try/except to the `class LeveldbStorageTest` object when it tries to import leveldb and then approrpiate xfails to the tests it would run on.

From what I understand the `pytest.importorskip('leveldb')` attribute in the class should be handling this - its purpose seems to be for skipping tests in a class if the module fails to import? Looking at the code in pytest it triggers only on ImportError, so I'll look into adding a feature there to have it trigger on SystemErrors too.

.travis.yml
```diff
@@ -27,6 +27,10 @@ matrix:
       python: 3.7
     - env: TOXENV=py37-extra-deps
       python: 3.7
+    - env: TOXENV=py38
+      python: 3.8
+    - env: TOXENV=py38-extra-deps
```
wRAR: We agreed to keep one -extra-deps env, though we didn't decide which one, it doesn't matter that much.

further-reading: Oh, kmike told me to change that initially. I see now that he put a comment to the resolved conversation disregarding the suggestion after I committed it. I'll revert it back to using the 3.7 extra dependency environment.

.travis.yml
```diff
@@ -27,6 +27,10 @@ matrix:
       python: 3.7
     - env: TOXENV=py37-extra-deps
       python: 3.7
+    - env: TOXENV=py38
+      python: 3.8
+    - env: TOXENV=py37-extra-deps
+      python: 3.8
```
Gallaecio: I think we should keep `extra-deps` in the latest supported Python version, just as `pinned` is in the oldest. So I think this requires the following changes:
- Removing lines 28-29 (`py37-extra-deps` entry)
- Changing `py37-extra-deps` back to to `py38-extra-deps` in new line 32.

setup.py
```diff
@@ -56,6 +56,7 @@ def has_environment_marker_platform_impl_support():
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
+        # 'Programming Language :: Python :: 3.8',  not supported yet
```
Gallaecio:
```suggestion
        'Programming Language :: Python :: 3.8',
```

tests/test_downloadermiddleware_httpcache.py
```diff
@@ -156,7 +160,12 @@ def _get_settings(self, **new_settings):
 
 class LeveldbStorageTest(DefaultStorageTest):
 
-    pytest.importorskip('leveldb')
+    try:
+        pytest.importorskip('leveldb')
+    except SystemError:
+        # Happens in python 3.8
+        # This will cause xfail in DefaultStorageTest to trigger
+        pass
```
Gallaecio: I think you can use `pytest.skip("Cannot import leveldb")` here (raises an exception), and not have to use `@pytest.mark.xfail` above.

setup.py
further-reading: Kmike said above that ye agreed on removing 3.8 from setup for now?

.travis.yml
further-reading: My understanding was that the leveldb issue means that 3.8 is not officially supported yet, so shouldn't we leave it that extra dep check in 3.7 for now?

tests/test_downloadermiddleware_httpcache.py
further-reading: Yeah, that should work. When I'm back at my computer later I'll remove those xfails and use the pytest.skip here instead.

Gallaecio:
We also need this: https://github.com/further-reading/scrapy/pull/1

I think that’s all. The remaining documentation changes (mentioning 3.8 support and LevelDB storage backend deprecation) belong to the news.rst documentation file, which is filled in a separate pull request soon before a release.

setup.py
Gallaecio: @kmike, @wRAR and me had a discussion after that comment from @kmike, and we decided to deprecate the LevelDB storage backend in 1.8 and remove it in Scrapy 2.0, after which delaying the `setup.py` update is no longer necessary.

further-reading: Thanks for clarifying, will have these changes pushed shortly!

tox.ini
```diff
@@ -128,6 +132,13 @@ deps =
     reppy
     robotexclusionrulesparser
```
Gallaecio: I guess we should also remove this.

tests/test_downloadermiddleware_httpcache.py
```diff
@@ -156,7 +157,11 @@ def _get_settings(self, **new_settings):
 
 class LeveldbStorageTest(DefaultStorageTest):
 
-    pytest.importorskip('leveldb')
+    try:
+        pytest.importorskip('leveldb')
+    except SystemError:
+        # Happens in python 3.8
+        pytestmark = pytest.skip("'SystemError: bad call flags' occurs on Python 3.8")
```
Gallaecio: Is the assignment needed?

```suggestion
        pytest.skip("'SystemError: bad call flags' occurs on Python 3.8")
```

further-reading: Yes - travis CI called it out specifically in the last build https://travis-ci.org/scrapy/scrapy/jobs/603880139?utm_medium=notification&utm_source=github_status

```
Using pytest.skip outside of a test is not allowed. To decorate a test function, use the @pytest.mark.skip or @pytest.mark.skipif decorators instead, and to skip a module use `pytestmark = pytest.mark.{skip,skipif}.
```

Gallaecio: This is getting complex.

I think you may be able to move the `try-except` to something like https://docs.pytest.org/en/latest/xunit_setup.html#method-and-function-level-setup-teardown, but feel free to go back to your initial approach of decorators based on the running Python version.

further-reading: I'm not sure how using a setup method is gonna make it less complex? You mean making something like or is there another approach I'm missing?

```
class LeveldbStorageTest(DefaultStorageTest):
    storage_class = 'scrapy.extensions.httpcache.LeveldbCacheStorage'

    @classmethod
    def setup_class(cls, *args, **kwargs):
        test_class = cls(*args, **kwargs)
        try:
            test_class.leveldb = pytest.importorskip('leveldb')
        except SystemError:
            test_class.pytestmark = pytest.mark.skip("'SystemError: bad call flags' occurs on Python >= 3.8")

        return test_class
```

Right now I have it refactored to a single check localized to the specific method where the problem arises, with clear commenting/logging when it triggers. Reverting back to the multiple decorators or incorporating a setup method seems like it'd be adding complexity as I am adding adding checks to multiple parts of the code or I am having to incorporate additional syntax to mimic the current behavior.

further-reading:
For the latest travis test - The failure is happening in the python 3.6 enviornment and seems unrelated to the changes I made. https://travis-ci.org/scrapy/scrapy/jobs/603892627?utm_medium=notification&utm_source=github_status I've seen this test fail randomly a few times only to work fine on an additional run.

In this latest commit the py3.8 tests are working on travis.

tests/test_downloadermiddleware_httpcache.py
```diff
@@ -6,6 +6,7 @@
 import email.utils
 from contextlib import contextmanager
 import pytest
+import sys
```
Gallaecio: :lipstick: This does not seem to be used below.

Issue #2552, opened 2017-02-09
scrapy.Request no init error on invalid url


pawelmhm:
I stumbled on some weird issue, spider got some invalid url, but instead of crashing loudly when trying to create scrapy.Request() with invalid url it just silently ignored this error. Sample to reproduce

```python
from scrapy.spiders import Spider
from scrapy import Request


class DmozSpider(Spider):
    name = "dmoz"
    allowed_domains = ["dmoz.org"]
    start_urls = [
        "http://www.dmoz.org/Computers/Programming/Languages/Python/Books/",
    ]

    def parse(self, response):
        invalid_url = "/container.productlist.productslist.productthumbnail.articledetaillink.layerlink:open-layer/0/CLASSIC/-1/WEB$007cARBO$007c13263065/null$007cDisplay$0020Product$002f111499$002fAil$0020blanc$007c?t:ac=13263065"
        yield Request(invalid_url)
```

this generates following output:

```
2017-02-09 12:21:04 [scrapy.core.engine] INFO: Spider opened
2017-02-09 12:21:04 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2017-02-09 12:21:04 [scrapy.extensions.telnet] DEBUG: Telnet console listening on 127.0.0.1:6024
2017-02-09 12:21:04 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://www.dmoz.org/Computers/Programming/Languages/Python/Books/> (referer: None)
2017-02-09 12:21:04 [scrapy.core.engine] INFO: Closing spider (finished)
```

there is no information about trying to generate this Request with invalid_url, no stacktrace, no error info from middleware. Why?

redapple:
Disturning indeed.
If you have an invalid URL in `start_urls`, scrapy will complain:

```
2017-02-09 12:30:40 [scrapy.core.engine] ERROR: Error while obtaining start requests
Traceback (most recent call last):
  File "/home/paul/.virtualenvs/scrapy13/local/lib/python2.7/site-packages/scrapy/core/engine.py", line 127, in _next_request
    request = next(slot.start_requests)
  File "/home/paul/.virtualenvs/scrapy13/local/lib/python2.7/site-packages/scrapy/spiders/__init__.py", line 70, in start_requests
    yield self.make_requests_from_url(url)
[...]

PR #4094, opened 2019-10-22
Improve URL schema validation on scrapy.Request initialization

victor-torres:
This PR should fix issue #2552 by improving the request schema validation on its initialization. 
The implementation is based on feedback posted by @pawelmhm and @hsumerf

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -66,7 +66,7 @@ class Request(object_ref):
         s = safe_url_string(url, self.encoding)
         self._url = escape_ajax(s)
 
-        if ':' not in self._url:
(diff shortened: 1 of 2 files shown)


victor-torres:
@Gallaecio, indeed. In the original thread, @elacuesta proposed a more robust approach. There's a discussion about performance implications going with it, but the issue isn't updated since Mar 2018. So I'm opening a PR with the simple suggestions already proposed.

Thank you for your quick review.

lopuhin:
This makes ``about:blank`` URLs invalid (and some others https://ru.wikipedia.org/wiki/About:_URI) which can be useful when working in a headless browser context. I wonder what would be the best way to handle them?

Gallaecio:
I think we could simply add support for it, just as we support `data`.

PR #4048, opened 2019-10-01
DO NOT MERGE try to see what causes flakiness

s-sanjay:

PR #4095, opened 2019-10-22
Use communicate() instead of wait() after killing the mock server

### LINKED ISSUE #4014: Random(?) test_crawler.py failure on Travis
Every now and then the Travis build fails on [`tests.test_crawler.CrawlerRunnerTestCase.test_deprecated_attribute_spiders`](https://github.com/scrapy/scrapy/blob/1.7.3/tests/test_crawler.py#L173), I'm not sure why.
Some examples:
* https://travis-ci.org/scrapy/scrapy/jobs/582184413#L407
* https://travis-ci.org/scrapy/scrapy/jobs/580207692#L419
* https://travis-ci.org/scrapy/scrapy/jobs/584701732#L415
* https://travis-ci.org/scrapy/scrapy/jobs/584762868#L418

PS: Sorry, I feel bad for ignoring the bug report template, but I felt like it didn't apply in this case.

Gallaecio:
Fixes #4014

When using pipes, and we are using a pipe for `stdout` in `MockServer`, you are meant to call `process.communicate()` instead of `process.wait()`.

This was causing unexpected warnings in `test_crawler.py` tests, when the garbage collection happened to warn about unclosed resources while one of the affected tests was capturing warnings.

I’ve run the tests a few rounds with this fix, and it seems to have worked.

### DIFF
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -205,9 +205,8 @@ class MockServer():
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.proc.communicate()  # Close the stdout PIPE
         self.proc.kill()
-        self.proc.wait()
+        self.proc.communicate()
         time.sleep(0.2)
 
     def url(self, path, is_secure=False):
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -205,6 +205,7 @@ class MockServer():
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
+        self.proc.communicate()  # Close the stdout PIPE
         self.proc.kill()
         self.proc.wait()
         time.sleep(0.2)
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -207,7 +207,6 @@ class MockServer():
     def __exit__(self, exc_type, exc_value, traceback):
         self.proc.kill()
         self.proc.communicate()
-        time.sleep(0.2)
 
     def url(self, path, is_secure=False):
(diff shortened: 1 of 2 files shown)


tests/mockserver.py
```diff
@@ -206,7 +206,7 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.proc.kill()
-        self.proc.wait()
+        self.proc.communicate()
         time.sleep(0.2)
```
elacuesta: I commented out this line locally and it worked, perhaps it was an attempt to prevent this kind of issues and it's not necessary anymore if we use `communicate`?

s-sanjay:
@Gallaecio how did you root cause the issue to this test ?

Gallaecio:
> @Gallaecio how did you root cause the issue to this test ?

It took more time that I would like to admit to figure things out.

In the end, search engine results for similar issues started to suggest, based on the naming (positive integers) of the unclosed file descriptors (which we first knew thanks to your #4048, by the way!), that the files were created by subprocess pipes.

Once I reached that conclusion, locating the affected code was not hard. It was unlikely that we were using subprocess in Scrapy itself, so it had to be in the tests. And the MockServer was using subprocess with a pipe, and was used strongly in the test file executed right after the file with random failing tests.

Then it was a matter of finding out what was the right way to get pipe files closed, see how we were actually doing it, and hence this patch. After 4-5 runs of the patched build being unable to reproduce the issue, I decided it was very likely that this patch had fixed the issue (which hopefully it has).

When I checked the documentation about

PR #4096, opened 2019-10-22
Retrying Requests when an IOError/OSError exception is raised when decompressing Response

### LINKED ISSUE #1900: RetryMiddleware doesn't actually retry HttpCompressionMiddleware errors
A comment in the code of RetryMiddleware says

```
    # IOError is raised by the HttpCompression middleware when trying to
    # decompress an empty response
```

when including IOError in the exceptions to retry.

But IOError would be raised in HttpCompressionMiddleware's `process_response()`, errors from which are not handled by `process_exception()` in downloader middlewares.

victor-torres:
fix #1900 by handling IOError and OSError exceptions raised while trying to decompress responses

- IOError/OSError exceptions would be raised in HttpCompressionMiddleware.process_response
- exceptions raised from process_response are not handled by downloader middlewares

This might not be the best approach for solving this bug, but it was the best I could think that would not involve a huge refactoring, including some kind of a `RetrySpiderMiddleware`.

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -730,6 +730,13 @@ HttpCompressionMiddleware
    This middleware also supports decoding `brotli-compressed`_ responses,
    provided `brotlipy`_ is installed.
 
+   If an exception is raised while trying to decompress the response body,
+   we encapsulate the exception in ``response.meta['_http_compression_exc']``
+   and return the original ``scrapy.Response``. The ``RetryMiddleware`` will
+   check for this exception and try to reprocess the ``scrapy.Request`` if
+   the captured exception is specified in the
+   ``RetryMiddleware.EXCEPTIONS_TO_RETRY`` tuple.
+
 .. _brotli-compressed: https://www.ietf.org/rfc/rfc7932.txt
 .. _brotlipy: https://pypi.python.org/pypi/brotlipy
 
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -36,7 +36,11 @@ class HttpCompressionMiddleware(object):
             content_encoding = response.headers.getlist('Content-Encoding')
             if content_encoding:
                 encoding = content_encoding.pop()
-                decoded_body = self._decode(response.body, encoding.lower())
+                try:
+                    decoded_body = self._decode(response.body, encoding.lower())
+                except Exception as exc:
+                    response.meta['_http_compression_exc'] = exc
+                    return response
                 respcls = responsetypes.from_args(headers=response.headers, \
                     url=response.url, body=decoded_body)
                 kwargs = dict(cls=respcls, body=decoded_body)
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -32,7 +32,7 @@ class RetryMiddleware(object):
     EXCEPTIONS_TO_RETRY = (defer.TimeoutError, TimeoutError, DNSLookupError,
                            ConnectionRefusedError, ConnectionDone, ConnectError,
                            ConnectionLost, TCPTimedOutError, ResponseFailed,
-                           IOError, TunnelError)
+                           IOError, OSError, TunnelError)
 
     def __init__(self, settings):
         if not settings.getbool('RETRY_ENABLED'):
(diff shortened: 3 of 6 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -737,6 +737,7 @@ HttpCompressionMiddleware Settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. setting:: COMPRESSION_ENABLED
+.. setting:: RETRY_HTTP_COMPRESSION_ERRORS
```
Gallaecio: This line should be right above the `RETRY_HTTP_COMPRESSION_ERRORS` section instead :slightly_smiling_face:

docs/topics/downloader-middleware.rst
```diff
@@ -745,6 +746,21 @@ Default: ``True``
 
 Whether the Compression middleware will be enabled.
 
+RETRY_HTTP_COMPRESSION_ERRORS
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Default: ``False``
+
+Whether we should retry requests when an exception is raised while trying
```
Gallaecio: :lipstick: Style suggestions:
* I think in English “whether” only makes sense if there is an “or” somewhere. You can alternatively use “if” instead.
* I would avoid “we” here, and instead switch to passive.
* Since you use “its” in the next line, using singular “request” makes more sense; you could also change “its” to “theirs” instead, but I find that using singular in this context is more common in English.

```suggestion
Whether or not a request should be retried when an exception is raised while trying
```

docs/topics/downloader-middleware.rst
```diff
@@ -745,6 +746,21 @@ Default: ``True``
...
+to decompress its response body.
+
+If an exception is raised while trying to decompress the response body,
```
```suggestion
If an exception is raised while trying to decompress a response body,
```

docs/topics/downloader-middleware.rst
```diff
@@ -745,6 +746,21 @@ Default: ``True``
...
+we store it in the request meta under the ``_http_compression_exc`` key and
```
Gallaecio:
```suggestion
that exception is stored in the request :attr:`~scrapy.http.Request.meta` under the ``_http_compression_exc`` key and
```

docs/topics/downloader-middleware.rst
```diff
@@ -745,6 +746,21 @@ Default: ``True``
...
+return the original ``scrapy.Response`` object. If the ``RetryMiddleware`` is
+enabled, it will check for this key and try to reprocess the ``scrapy.Request``
```
Gallaecio:
```suggestion
enabled, it will check for this key and try to reprocess the :class:`~scrapy.http.Request`
```

docs/topics/downloader-middleware.rst
```diff
@@ -745,6 +746,21 @@ Default: ``True``
...
+if the captured exception is specified in the
+``RetryMiddleware.EXCEPTIONS_TO_RETRY`` tuple.
```
Gallaecio: I think we should either remove this “if …” statement from the documentation or expose `EXCEPTIONS_TO_RETRY` through a setting.

docs/topics/downloader-middleware.rst
victor-torres: I'm used to native speakers omitting the _or not_ part of the _whether_ expression as it's most of the time redundant.

Here you can find some references:
- https://dictionary.cambridge.org/dictionary/english/whether
- https://english.stackexchange.com/questions/3382/whether-or-not-vs-whether

It's also being used this way by the `COMPRESSION_ENABLED` setting definition above.

In any case, I do find your suggested change better and I'm willing to commit it. Thank you.

docs/topics/downloader-middleware.rst
victor-torres: I do agree with you. Although I prefer to expose this class attribute through a setting, this is out of the scope of this ticket. So I'm just removing the _if_.

docs/topics/downloader-middleware.rst
victor-torres: My bad 🤦‍♂

docs/topics/downloader-middleware.rst
Gallaecio: Thanks for the links! Good to know :smiley:

Gallaecio:
Since #1900 is a bug, I wonder if maybe we should not expose a setting for this, and instead apply the change transparently and require users to implement their own retry middleware if they need the old behavior, just as they currently have to implement their own retry middleware if they want a different set of exceptions to be retried.

If we implement a setting, maybe it should be exposing the exceptions that the retry middleware catches.

@kmike what’s your take on this? Should we keep the proposed settings to allow for backward compatibility?

PR #4098, opened 2019-10-22
Changed to proper `if-elif` sequencing in spidermw.py.

sbs2001:
Just a nitpick. We will  avoid the '_isiterable' method call in case `result` is `None` and directly continue.


Gallaecio:
Unless `result` being `None` is significantly more common than `result` being an iterable, this will have a slightly negative impact on performance.

elacuesta:
This patch changes behavior in a way that is not correct IMHO. The current code checks `result` _after_ applying the current exception handler, while the new code performs the check _before_ the handler, which means it checks the result from the previous iteration. This is the reason the tests are failing.

sbs2001:
@elacuesta  I guess I missed that, thank you for pointing that out!

PR #4099, opened 2019-10-23
update docs of scrapy.loader.ItemLoader.item

### LINKED ISSUE #3574: ItemLoader adds wrong value instead of an empty array
Please take a look at this code:

```python
import scrapy


class MyItem(scrapy.Item):
    a = scrapy.Field()


class MyLoader(ItemLoader):
    item = MyItem()


loader = MyLoader()
loader.add_value("a", "b")
item = loader.load_item()
print(item.get("a"))

loader = MyLoader()
loader.add_value("a", [])
item = loader.load_item()
print(item.get("a"))
```

The output should be something like
```
['b']
None
```
But it actually is 
```
['b']
['b']
```

BurnzZ:
Attempts to clear up the doc misunderstanding in #3574 that results in wrong expected behavior due to incorrect usage.

Fixes #3574

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -485,6 +485,8 @@ ItemLoader objects
     .. attribute:: item
 
         The :class:`~scrapy.item.Item` object being parsed by this Item Loader.
+        This is mostly used as a property so when attempting to override this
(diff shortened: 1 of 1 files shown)


kmike:
A good idea, thanks @BurnzZ!

PR #4100, opened 2019-10-23
Replaced constructor with __init__ method

paritoshM9:


wRAR:
Isn't this better fixed in https://github.com/scrapy/scrapy/pull/4088 ?

Gallaecio:
@paritoshM9 Thank you for your work, but indeed there is an existing proposal (#4088) which is close to completion and covers much more ground.

Issue #4067, opened 2019-10-09
Question: Handling fragment identifiers


vella-nicholas:
I am scraping a web page which has fragment identifiers (on the home page) for same page navigation. To make it clear I have this anchor tag `<a href="#idx1">1. Histoire</a>` and when you click on it you are taken to the 'Histoire' header on the same page.  `<h2 id="idx1">1. Histoire</h2>`.

Scrapy is following this fragment as `https://www.mysite.com/#idx1` (or `https://www.mysite.com#idx1`) and scraping the whole page again. Since this is the home page, it is yielding the same results as if it scraped the home page.

This behavior is not repeated for the rest of the fragment identifiers (I have many more). So for this anchor tag (on the same page) the above scenario is not manifested. `<a href="#idx2">2. Idées reçues</a>` -> Do not repeat itself for this.

Is this intentional or am I missing something?

```
import scrapy
from datetime import datetime
from netlinking.items import NetlinkItem
from urllib.parse import urlparse, urlunparse


class NetlinkingSpider(scrapy.Spider):
    name = "netlinking"
    referent_netloc = ''

    def __init__(self, url=None, *args, **kwargs):
        super(NetlinkingSpider, self).__init__(*args, **kwargs)
        parsed_uri = urlparse(url)
        self.referent_netloc = parsed_uri.netloc
        self.referent_netloc = self.referent_netloc.replace('www.', '')
        self.allowed_domains = [('%s' % self.referent_netloc)]
        self.start_urls = [('%s' % url)]

    def parse(self, response):
        for anchor in response.css('a'):
            netlink = NetlinkItem()
            netlink['anchor_text'] = anchor.css('a::text').get()
            netlink['referent_website'] = response.request.url
            netlink['referent_netloc'] = self.referent_netloc
            netlink['target_url'] = anchor.css('a::attr(href)').get()

            parsed_target_uri = urlparse(netlink['target_url'])
            if parsed_target_uri.netloc.strip() \
                    and (self.referent_netloc !=
                         parsed_target_uri.netloc.lower().replace('www.', '')):

                netlink['target_website'] = \
                    '{uri.scheme}://{uri.netloc}/'.format(
                    uri=parsed_target_uri)

                netlink['target_netloc'] = '{uri.netloc}'.format(
                    uri=parsed_target_uri).replace('www.', '')

                netlink['timestamp'] = datetime.now().strftime(
                    "%Y-%m-%d %H:%M:%S")

                yield netlink

[...]

PR #4104, opened 2019-10-25
Allow not to dedupe urls with different fragments

boogheta:
w3lib's canonicalize_url has an option to keep fragments but it is not accessible from scrapy's request_fingerprint method, so it's impossible without changing scrapy's code to make a custom dedupe filter which would keep urls identical except for fragments.

We need this in our webcrawler Hyphe for our future headless crawling feature using chromedriver (currently developed in https://github.com/medialab/hyphe/pull/288 ) to crawl modern websites with routing included in the fragment.

For what I understand the solution discussed in #4067 using `dont_filter=False` on each query would completely disable deduping for the spider which is not an option for a crawler that adds new links from each crawled page as it would then run indefinitely in loop

cc @arnaudmolo

### DIFF
--- a/scrapy/utils/request.py
+++ b/scrapy/utils/request.py
@@ -42,6 +42,11 @@ def request_fingerprint(request, include_headers=None, keep_fragments=False):
     the fingeprint. If you want to include specific headers use the
     include_headers argument, which is a list of Request headers to include.
 
+    Also, servers usually ignore fragments in urls when handling requests,
+    so they are also ignored by default when calculating the fingerprint.
+    If you want to include them, set the keep_fragments argument to True
+    (for instance when handling requests with a headless browser).
+
     """
     if include_headers:
         include_headers = tuple(to_bytes(h.lower())
--- a/scrapy/utils/request.py
+++ b/scrapy/utils/request.py
@@ -16,7 +16,7 @@ from scrapy.utils.httpobj import urlparse_cached
 
 
 _fingerprint_cache = weakref.WeakKeyDictionary()
-def request_fingerprint(request, include_headers=None):
+def request_fingerprint(request, include_headers=None, keep_fragments=False):
     """
     Return the request fingerprint.
 
@@ -47,10 +47,11 @@ def request_fingerprint(request, include_headers=None):
         include_headers = tuple(to_bytes(h.lower())
                                  for h in sorted(include_headers))
     cache = _fingerprint_cache.setdefault(request, {})
-    if include_headers not in cache:
+    cache_key = (include_headers, keep_fragments)
+    if cache_key not in cache:
         fp = hashlib.sha1()
         fp.update(to_bytes(request.method))
-        fp.update(to_bytes(canonicalize_url(request.url)))
+        fp.update(to_bytes(canonicalize_url(request.url, keep_fragments=keep_fragments)))
         fp.update(request.body or b'')
         if include_headers:
             for hdr in include_headers:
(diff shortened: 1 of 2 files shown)


elacuesta:
The feature makes sense and the implementation looks good IMHO. Current test failure is caused by #4014 so it's fine. Could you add a test to https://github.com/scrapy/scrapy/blob/1.7.4/tests/test_utils_request.py?

wRAR:
This looks useful, can you please add a test for this to test_utils_request.py?

boogheta:
Hi @elacuesta and thanks for the quick feedback and the test addition recommendation: it made me realise my naive approach was not taking account of the fingerprint cache, I repushed with a slight modification to do so, with according tests

Gallaecio:
> it's impossible without changing scrapy's code to make a custom dedupe filter which would keep urls identical except for fragments.

Isn’t this true even after these changes? Don’t you still need to create your own duplicate filter that uses `request_fingerprint` with `keep_fragments=False` and assign it to [`DUPEFILTER_CLASS`](https://docs.scrapy.org/en/latest/topics/settings.html#dupefilter-class)?

I’m hesitant about this because it may complicate further the solution to #900.

boogheta:
Hi @Gallaecio, we've ran tests on our codebase with the above change and it works with it in our case.

It is true we need anyway to create our own duplicate filter that uses request_fingerprint with keep_fragments=False, the problem the PR solves it that doing so was not possible otherwise, since the keep_fragments argument could not be passed to canonicalize_url throuh request_fingerprint.

I didn't see the discussion about #900 before and I understand indeed that this change should be integrated within it when it is implemented but I don't think it would too much of a hassle to do, and I'll be happy to participate then ! :)

Gallaecio:
Could you update the docstring of the function? There’s a mention to the `include_headers` parameter towards the end, it would be good for it to also mention the new parameter in a similar way.

boogheta:
Ha yes indeed, sorry I forgot, will do right away!

boogheta:
Done!

boogheta:
Thank you!
May I ask if we can expect it within a release soon?

Gallaecio:
Well, that depends on your definition of soon :slightly_smiling_face:. The current plans are for the next release to be 2.0, sometime in 2020 Q1.

boogheta:
Ah that's far indeed ;)
And I must say a bit frightening consideringaccording to semantic versioning, 2.0 usually means breaking changes by then.
I saw you released a 1.8.0 just before merging, wouldn't it make sense to see a few other minor releases with bugfixes and slight improvements such as a 1.8.1 by then?

Gallaecio:
In patch versions we usually include only important bugfixes, so I’m sorry to say that, even if there is a 1.8.1 version, it will probably not include this change :slightly_frowning_face:

boogheta:
I don't understand, isn't it the whole point of making minor releases to include small changes along with bugfixes? What would be the problem to include such changes that only add functionality without breaking anything? It would feel very frustrating to contribute only to see changes completely overwritten and broken by the time they finally end up in a release :-(

Gallaecio:
As I said, we tend to stick to important bugfixes, to keep patch releases small and simple. But I guess we can evaluate including this change in 1.8.1 _if_ we decide to have a 1.8.1 release. I’ll create a milestone so we do not forget.

boogheta:
Thanks for taking it into consideration!

elacuesta:
@boogheta In the meantime, is it possible for you to install using the commit hash, i.e. `pip install git+https://github.com/scrapy/scrapy.git@6d6da78eda3cc0bba1bfdf70194fdf655fac8aeb`?

boogheta:
Yes, this is probably what we will end up doing in our repo by then, although this is not a very good practice and I know some colleagues will shame us about doing so ;)

PR #4108, opened 2019-10-28
CrawlSpider: Prevent stop following if callback raises exception

pauloromeira:
In my opinion, exceptions when parsing a response shouldn't be a blocker for the crawler to keep following links.

This might also be a setting (stop following if callback raises exception).

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -102,15 +102,16 @@ class CrawlSpider(Spider):
         return self._parse_response(response, rule.callback, rule.cb_kwargs, rule.follow)
 
     def _parse_response(self, response, callback, cb_kwargs, follow=True):
-        if callback:
-            cb_res = callback(response, **cb_kwargs) or ()
-            cb_res = self.process_results(response, cb_res)
-            for requests_or_item in iterate_spider_output(cb_res):
-                yield requests_or_item
-
-        if follow and self._follow_links:
-            for request_or_item in self._requests_to_follow(response):
-                yield request_or_item
+        try:
+            if callback:
+                cb_res = callback(response, **cb_kwargs) or ()
+                cb_res = self.process_results(response, cb_res)
+                for requests_or_item in iterate_spider_output(cb_res):
+                    yield requests_or_item
+        finally:
+            if follow and self._follow_links:
+                for request_or_item in self._requests_to_follow(response):
+                    yield request_or_item
 
     def _compile_rules(self):
         self._rules = [copy.copy(r) for r in self.rules]


elacuesta:
Hello @pauloromeira, thanks for the PR

Could you elaborate a bit more? In a regular Spider, exceptions from callbacks are not treated specially, I'm not sure I see a reason to do so with the CrawlSpider.
Also, the current implementation is not ideal in the sense that exceptions are silently ignored, we definitely do not want that.

pauloromeira:
Hi @elacuesta, of course!

To exemplify the issue, please consider the set of rules:
```python
    rules = (
        Rule(link_extractor_1, callback='parse_1', follow=True),
        Rule(link_extractor_2, callback='parse_2')
    )
```

In the current implementation, even with the argument `follow=True`, the crawler will not run the second rule if the `parse_1` callback raises an exception. In my opinion this exception shouldn't avoid next rules from running. Or maybe there should be a setting or an argument on `Rule` to choose its behaviour.

About the exceptions being silently ignored, that's not happening. The implementation is not catching any of them. It's a `try-finally` block, so it's just guaranteeing that the `finally` block will run no matter what.

elacuesta:
> About the exceptions being silently ignored, that's not happening. The implementation is not catching any of them. It's a `try-finally` block, so it's just guaranteeing that the `finally` block will run no matter what.

Interesting, I wasn't aware of this behaviour. Thanks for the explanation! :smile: 

Please correct me if I'm wrong, but if I understand correctly, what this does is allow the spider to extract requests from responses even if their user-provided callback fails. I believe this is not exactly the same as running the next rules in the list/tuple, since each extracted link is only processed by the first matching rule.

Assuming I got it right, I am still unsure about this, I think it might be too unintuitive compared to the regular `Spider` class, which doesn't do anything by default when exceptions are raised. Could you add tests and docs? Perhaps that would clarify some things and help us review.

pauloromeira:
> Please correct me if I'm wrong, but if I understand correctly, **what this does is allow the spider to extract requests from responses even if their user-provided callback fails**. I believe this is not exactly the same as running the next rules in the list/tuple, since each extracted link is only processed by the first matching rule.

Yes, you've explained it perfectly! The response won't be processed by the next rules, it will only allow the crawler to extract links even if the user-provided callback fails. I'm sorry I didn't explain it correctly before.

To be honest I'm not too sure about the most intuitive behaviour, but if you think about the async nature of Scrapy, if you yield a bunch of requests the crawler won't stop if the callback of the first request fails. So _maybe_ in a CrawlSpider crawling should be decoupled from parsing.

I'll try to add tests and docs when I have time. Thanks for reviewing!

AdrianAtZyte:
Should we consider this handled by #4000?

PR #4111, opened 2019-10-30
refactor LogStats extension to log IPM and RPM to stats on spider_close

### LINKED ISSUE #4110: LogStats Extension should log the IPM and RPM to the stats on spider_closed signal
## Summary

By default, the `scrapy.extensions.logstats.LogStats` log the details below to the standard logger every 60 seconds:

- IPM (_Items per Minute_)
- Items scraped
- RPM (_Requests per Minute_)
- Requests produced

## Motivation

We often find the need to retrieve the **RPM** and **IPM** after a job is finished. We can calculate this manually or we can check the logs for the last update of *LogStats*. However, this is tedious and inefficient as `LogStats` already has this data but isn't putting them somewhere that is readily available after the job is finished.

We can pull out the last update of `LogStats` from the logs but pulling it from the API and filtering it out is expensive.

## Additional context

None

BurnzZ:
Fixes #4110 

**NOTE:** I noticed that the `LogStats` extension doesn't have any tests at all. I will write the tests later when this refactoring approach is approved.


scrapy/extensions/logstats.py
```diff
@@ -35,18 +38,25 @@ def spider_opened(self, spider):
...
+                    'items': self.items, 'itemrate': self.irate}
         logger.info(msg, log_args, extra={'spider': spider})
 
     def spider_closed(self, spider, reason):
         if self.task and self.task.running:
             self.task.stop()
+
+        self.calculate_stats()
+        self.stats.set_value('IPM', self.irate)
+        self.stats.set_value('RPM', self.prate)
```
Gallaecio: :lipstick:  I think `items_per_minute` and `requests_per_minute` would be a better choice here.

BurnzZ: Thanks @Gallaecio! I'll update this one along with the tests. Please don't merge this one yet as I'd like to have tests on this module. Marking it as `WIP:` at the moment.

kmike:
The feature looks good to me, but it needs an update to the current master branch.

BurnzZ:
Thanks for unearthing this @kmike as I totally forgot about it 😅 . I've added tests and rebased the PR to master.

kmike (approved):

PR #4114, opened 2019-10-31
Remove py2 tests

### LINKED ISSUE #2545: enable test_proxy_connect in Python 3

wRAR:
This removes Python2-only tests and updates test_proxy_connect.py to new mitmproxy.

Fixes #2545.


tests/test_proxy_connect.py
```diff
@@ -1,63 +1,78 @@
 import json
 import os
+import re
+from subprocess import Popen, PIPE
+import sys
```
Gallaecio: :lipstick: I think the `subprocess` import should be below `sys`, being in the `from … import …` form. See https://stackoverflow.com/a/20763446/939364

tests/test_proxy_connect.py
```diff
@@ -1,63 +1,78 @@
...
+                           stdout=PIPE, env=get_testenv())
+        line = self.proc.stdout.readline().decode('utf-8')
+        host_port = re.search(r'listening at http://([^:]+:\d+)', line).group(1)
+        address = 'http://%s:%s@%s' % (self.auth_user, self.auth_pass, host_port)
+        return address
+
+    def stop(self):
+        self.proc.kill()
+        self.proc.wait()
+        time.sleep(0.2)
```
Gallaecio: From recent experience, I think we need to change these last two lines to `self.proc.communicate()` :slightly_smiling_face:

tests/test_proxy_connect.py
```diff
@@ -100,9 +117,10 @@ def test_https_tunnel_without_leak_proxy_authorization_header(self):
         with LogCapture() as l:
             yield crawler.crawl(seed=request)
         self._assert_got_response_code(200, l)
-        echo = json.loads(crawler.spider.meta['responses'][0].body)
+        echo = json.loads(crawler.spider.meta['responses'][0].body.decode('utf-8'))
```
Gallaecio: Is this a Scrapy `Response` object? Because then:

```suggestion
        echo = json.loads(crawler.spider.meta['responses'][0].text)
```

conftest.py
```diff
@@ -8,11 +7,10 @@
 ]
 
 
-if six.PY3:
-    for line in open('tests/py3-ignores.txt'):
-        file_path = line.strip()
-        if file_path and file_path[0] != '#':
-            collect_ignore.append(file_path)
+for line in open('tests/py3-ignores.txt'):
```
Gallaecio: Should we take the chance to rename that file to simply `ignores.txt`?

tests/test_proxy_connect.py
```diff
@@ -67,6 +81,7 @@ def test_https_connect_tunnel(self):
             yield crawler.crawl(self.mockserver.url("/status?n=200", is_secure=True))
         self._assert_got_response_code(200, l)
 
+    @pytest.mark.xfail(reason='mitmproxy gives an error for noconnect requests')
```
kmike: What error is that? Is it some feature missing in mitmproxy, which was present before? If so, it'd be good to open a ticket in their issue tracker, and add a link  here.

wRAR: "Invalid request scheme: https" like in https://github.com/mitmproxy/mitmproxy/issues/848.

See https://github.com/scrapy/scrapy/issues/2545#issuecomment-532153176

Note that the mitmproxy issue above is open but I'm not 100% sure it's the same use case. Actually, I'm not sure *what* is the use case for the Scrapy code to skip CONNECT.

wRAR: https://github.com/scrapy-plugins/scrapy-crawlera/pull/44

wRAR: @kmike I've added a comment about this as requested

kmike:
Thanks @wRAR! I'm merging it; could you please open an issue to add noconnect tests back, to keep track of it (even if we decided not to add them back)?

PR #4115, opened 2019-10-31
Remove py2 code

wRAR:
This removes most of Py2-only code, except `six` wrappers and stuff covered in #4003 and #4114.

This also deprecates `scrapy.utils.gz.read1` and `scrapy.utils.python.to_native_str`.

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -474,7 +474,7 @@ DBM storage backend
 
     A DBM_ storage backend is also available for the HTTP cache middleware.
 
-    By default, it uses the dbm_ module, but you can change it with the
+    By default, it uses the `dbm module`_, but you can change it with the
     :setting:`HTTPCACHE_DBM_MODULE` setting.
 
 .. _httpcache-storage-custom:
@@ -1202,4 +1202,4 @@ The default encoding for proxy authentication on :class:`HttpProxyMiddleware`.
 
 
 .. _DBM: https://en.wikipedia.org/wiki/Dbm
-.. _dbm: https://docs.python.org/3/library/dbm.html
+.. _dbm module: https://docs.python.org/3/library/dbm.html
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -474,7 +474,7 @@ DBM storage backend
 
     A DBM_ storage backend is also available for the HTTP cache middleware.
 
-    By default, it uses the anydbm_ module, but you can change it with the
+    By default, it uses the dbm_ module, but you can change it with the
     :setting:`HTTPCACHE_DBM_MODULE` setting.
 
 .. _httpcache-storage-custom:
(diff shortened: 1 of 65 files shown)


scrapy/downloadermiddlewares/decompression.py
```diff
@@ -4,18 +4,14 @@
 
 import bz2
 import gzip
+from io import BytesIO
```
elacuesta: Nitpick: could you group the `^import ...` and `^from ...` lines together? There are a couple other occurrences of this in PR.

wRAR: All `^import` together and then all `^from` together? I've noticed only recently that @Gallaecio is promoting this style, I didn't use it before.

wRAR:
@elacuesta no, I didn't start working on other `six` stuff.

docs/topics/downloader-middleware.rst
```diff
@@ -474,7 +474,7 @@ DBM storage backend
 
     A DBM_ storage backend is also available for the HTTP cache middleware.
 
-    By default, it uses the anydbm_ module, but you can change it with the
+    By default, it uses the `dbm module`_, but you can change it with the
```
Gallaecio: Since we have InterSphinx configured for Python documentation, you should be able to use:

```suggestion
    By default, it uses the :mod:`dbm`, but you can change it with the
```

And not need the link reference from line 1205 anymore.

scrapy/downloadermiddlewares/decompression.py
Gallaecio: I must say I didn’t use it either until recently. I think I found about it in some pull request, and https://stackoverflow.com/a/20763446/939364 seems to confirm that this is the convention.

elacuesta:
> @elacuesta no, I didn't start working on other `six` stuff.

Great, I opened #4121, I'd appreciate your feedback there.

tests/test_link.py
```diff
@@ -46,12 +44,5 @@ def test_repr(self):
         self._assert_same_links(l1, l2)
 
     def test_non_str_url_py2(self):
```
kmike: could you please rename the test?

tests/test_utils_python.py
```diff
@@ -231,12 +231,11 @@ def __call__(self, a, b, c):
...
         else:
-            stripself = not six.PY2  # PyPy3 exposes them as methods
             self.assertEqual(
-                get_func_args(six.text_type.split, stripself), ['sep', 'maxsplit'])
-            self.assertEqual(get_func_args(" ".join, stripself), ['list'])
+                get_func_args(six.text_type.split, True), ['sep', 'maxsplit'])
+            self.assertEqual(get_func_args(" ".join, True), ['list'])
             self.assertEqual(
-                get_func_args(operator.itemgetter(2), stripself), ['obj'])
+                get_func_args(operator.itemgetter(2), True), ['obj'])
```
kmike: could you please use kwargs here?

kmike:
Thanks @wRAR, and thanks @elacuesta and @Gallaecio for the feedback! Let's merge it.

PR #4122, opened 2019-11-03
fix zsh completion file extension

seregaxvm:
Look for `.html` files instead of `.http` when completing `scrapy shell` command.


Gallaecio:
Thank you!

Issue #3793, opened 2019-05-24
Progress bar for large downloads


nemec:
What are your thoughts on adding a progress bar to the scrapy HTTP handler? I recently wrote a crawler that would scrape a site and throw any files into a `FilesPipeline` for download. Some of these files were 100+ MB in size which made the terminal seem to "freeze" while they downloaded in the background. I know scrapy isn't really designed to be an efficient file downloader like `aria2` or `jdownloader`, but it's a handy tool and I was already using it to scrape the file list.

I wrote a proof of concept using the Python library [tqdm](https://github.com/tqdm/tqdm) and it went even better than expected - tqdm automatically handles multiple progress bars at a time (scrapy queue) so I got a clean section a the bottom of the console showing individual progress for each pending file over 5MB in size.

Since I leaned so heavily on tqdm, the change to the scrapy source was only ~15 lines of code to fully implement (the POC patch is at the bottom of this post). If this feature is worth including, I'd expect other changes too since I'm sure you don't want scrapy to take a hard dependency on tqdm and the progress bar should have some configuration options, too.

![Screenshot from 2019-05-24 15-39-33](https://user-images.githubusercontent.com/136378/58356950-98e92880-7e3e-11e9-97a7-b196114ace1b.png)

## Considerations

* Disable the progress bar in noninteractive mode (does scrapy have this? how does scrapinghub behave?)
* Optional dependency on tqdm (or code the feature from scratch within scrapy? - this may be a lot of work)
* Configurable minimum size threshold for triggering the progress bar.
  * If tqdm is allowed as an optional dependency, the http11 handler should set a warning if devs try to set the minimum threshold but do not have tqdm installed
*  What to do when `txresponse.length` is `UNKNOWN_LENGTH`? This can happen if the server does not return a `Content-Length` header. Should it be disabled entirely? Or monitor `_bytes_received` and lazily create a progress bar if it crosses the threshold?

## Patch

I am using scrapy 1.5.0 in my POC but it looks like the source for http11 in master is unchanged except for the [addition of one line](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/handlers/http11.py#L36) disabling `lazy`, so the patch line numbers are mostly off by one.

```
--- ~/scrapy-1.5.0/scrapy/core/downloader/handlers/http11.py
+++ ~/.local/lib/python3.6/site-packages/scrapy/core/downloader/handlers/http11.py
@@ -28,6 +28,9 @@
[...]

PR #4126, opened 2019-11-05
Download handlers: from_crawler factory method, take crawler in __init__

elacuesta:
The inspiration for this came because of the proposed "bytes received" signal (#3793). With this change, the handlers could access the [signal manager stored in the crawler](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/crawler.py#L42), allowing them to trigger signals. 

I guess this could lead to compatibility issues with custom handlers that inherit from the built-in ones, I figured that shouldn't be a problem if we're changing the main version number to 2.


scrapy/core/downloader/handlers/datauri.py
```diff
@@ -8,8 +8,8 @@
 class DataURIDownloadHandler(object):
     lazy = False
 
-    def __init__(self, settings):
-        super(DataURIDownloadHandler, self).__init__()
+    def __init__(self, crawler=None):
+        pass
```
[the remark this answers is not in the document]
elacuesta: You are correct. Thanks for pointing it out!

scrapy/core/downloader/handlers/__init__.py
```diff
@@ -49,7 +49,11 @@ def _load_handler(self, scheme, skip_lazy=False):
             dhcls = load_object(path)
             if skip_lazy and getattr(dhcls, 'lazy', True):
                 return None
-            dh = dhcls(self._crawler.settings)
+            dh = create_instance(
+                dhcls,
+                self._crawler.settings,
+                self._crawler,
+            )
```
kmike: 👍

scrapy/core/downloader/handlers/s3.py
```diff
@@ -67,7 +66,11 @@ def __init__(self, settings, aws_access_key_id=None, aws_secret_access_key=None,
             except Exception as ex:
                 raise NotConfigured(str(ex))
 
-        self._download_http = httpdownloadhandler(settings).download_request
+        self._download_http = httpdownloadhandler(crawler).download_request
```
kmike: shouldn't it use create_instance as well?

tests/test_downloader_handlers.py
```diff
@@ -45,20 +44,32 @@ class DummyDH(object):
...
 
 class OffDH(object):
     lazy = False
 
     def __init__(self, crawler):
         raise NotConfigured
 
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
```
kmike: Are these methods used? Can you remove from_crawler and `__init__` methods from all DHs, and heep `__init__` with an exception only here?

tests/test_downloader_handlers.py
```diff
@@ -239,7 +250,7 @@ def setUp(self):
         else:
             self.port = reactor.listenTCP(0, self.wrapper, interface=self.host)
         self.portno = self.port.getHost().port
-        self.download_handler = self.download_handler_cls(Settings())
+        self.download_handler = self.download_handler_cls.from_crawler(get_crawler())
```
kmike: What do you think about using create_instance here, to simulate real usage better?

scrapy/core/downloader/handlers/ftp.py
```diff
@@ -70,10 +71,14 @@ class FTPDownloadHandler(object):
         "default": 503,
     }
 
-    def __init__(self, settings):
-        self.default_user = settings['FTP_USER']
-        self.default_password = settings['FTP_PASSWORD']
-        self.passive_mode = settings['FTP_PASSIVE_MODE']
+    def __init__(self, crawler):
```
kmike: I think if you keep settings argument, instead of switching to crawler, then FTPDownloader subclasses which override `__init__` method would continue to work. The same applies to HTTP10DownloadHandler.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -31,7 +31,9 @@
 class HTTP11DownloadHandler(object):
     lazy = False
 
-    def __init__(self, settings):
+    def __init__(self, crawler):
```
kmike: It looks easy enough to make this "almost" backwards compatible: keep settings argument, add crawler=None argument.

kmike:
Hey! `create_instance` for download handlers is a great new feature, +1 to add it. 

I think we should try more to preserve backwards compatibility; 2.0 is not a reason to make backwards incompatible changes without deprecation period. We're dropping Python 2 support in Scrapy 2.0, but Scrapy already worked with Python 3 for years.

scrapy/core/downloader/handlers/s3.py
```diff
@@ -30,11 +31,11 @@ def _mexe(self, http_request, *args, **kwargs):
     return _S3Connection
 
 
-class S3DownloadHandler(object):
+class S3DownloadHandler:
 
-    def __init__(self, settings, aws_access_key_id=None, aws_secret_access_key=None,
+    def __init__(self, settings, crawler=None,
```
elacuesta: Adding a new parameter in between existing ones is strange, I know. But I think adding it between `httpdownloadhandler` and `**kw` might be a bit strange too :man_shrugging: 
Let me know what you think.

elacuesta:
Hi @kmike! Good point about the deprecation period, thanks :+1:
Updated the PR, left a note about the S3 handler, would love you hear your thoughts about it.

scrapy/core/downloader/handlers/__init__.py
```diff
@@ -4,17 +4,17 @@
...
 from scrapy.utils.misc import create_instance, load_object
 from scrapy.utils.python import without_none_values
-from scrapy import signals
 
 
 logger = logging.getLogger(__name__)
 
 
-class DownloadHandlers(object):
+class DownloadHandlers:
```
kmike: 👍 It could make sense to do this cleanup globally (though not in this PR, of course)

scrapy/core/downloader/handlers/ftp.py
```diff
@@ -63,22 +63,22 @@ def close(self):
...
     CODE_MAPPING = {
         "550": 404,
         "default": 503,
     }
 
-    def __init__(self, crawler):
-        self.default_user = crawler.settings['FTP_USER']
-        self.default_password = crawler.settings['FTP_PASSWORD']
-        self.passive_mode = crawler.settings['FTP_PASSIVE_MODE']
+    def __init__(self, settings, crawler=None):
```
kmike: crawler is not needed here

kmike: It is not a goal for `__init__` methods of all downloader handlers to have the same signature, because downloader handlers are created using `create_instance` - it gives us flexibility to use any parameters which make sense in `__init__`, while still being able to create them in a generic way.

scrapy/core/downloader/handlers/s3.py
kmike: I think we actually can make a backwards incompatible change here: make all arguments other than settings  keyword-only. In this case order won't matter.

tests/test_downloader_handlers.py
```diff
@@ -809,7 +821,13 @@ def _mocked_date(self, date):
 
     def test_extra_kw(self):
         try:
-            S3DownloadHandler.from_crawler(get_crawler(), extra_kw=True)
+            crawler = get_crawler()
+            create_instance(
+                objcls=S3DownloadHandler,
+                settings=crawler.settings,
```
kmike: this argument can be skipped, because when crawler is passed, settings are default to crawler.settings in create_instance

scrapy/core/downloader/handlers/http10.py
```diff
@@ -1,17 +1,23 @@
...
-class HTTP10DownloadHandler(object):
+class HTTP10DownloadHandler:
     lazy = False
 
-    def __init__(self, settings):
-        self.HTTPClientFactory = load_object(settings['DOWNLOADER_HTTPCLIENTFACTORY'])
-        self.ClientContextFactory = load_object(settings['DOWNLOADER_CLIENTCONTEXTFACTORY'])
+    def __init__(self, settings, crawler=None):
+        self.HTTPClientFactory = load_object(crawler.settings['DOWNLOADER_HTTPCLIENTFACTORY'])
+        self.ClientContextFactory = load_object(crawler.settings['DOWNLOADER_CLIENTCONTEXTFACTORY'])
```
kmike: I'd prefer to use settings here, because crawler.settings doesn't exist when crawler=None; if we ever enable mypy, it'll bark at it. 

Maybe this None default doesn't work at all already - though I hope it should work in some cases, maybe with some features missing.

scrapy/core/downloader/handlers/s3.py
kmike: Making arguments kwarg-only is the same as this change, but allows to fail earlier.

To make it backwards compatible we can pick crawler from `kw` argument instead.

elacuesta: > make all arguments other than settings keyword-only

Nice, I like that. I suppose you're talking about [PEP 3102](https://www.python.org/dev/peps/pep-3102/)? I've seen the syntax but I haven't used it before.

scrapy/core/downloader/handlers/http10.py
elacuesta: I had it like that, but lost it between stashes. Thanks for pointing it out :ok_hand:

kmike:
Thanks for all the updates @elacuesta!

Issue #2174, opened 2016-08-11
gzip-compressed item exports


kmike:
I think compressing exported data can be useful in a lot of cases. It'd be good to have a built-in way for compressed exports in Scrapy. I have this implementation, but probably it makes more sense to handle .jl.gz / csv.gz / ... extensions just like .jl / .csv / ... instead of creating a storage, I'm not sure:

``` py
# -*- coding: utf-8 -*-
import os
import gzip

from zope.interface import Interface, implementer
from w3lib.url import file_uri_to_path
from scrapy.extensions.feedexport import IFeedStorage


@implementer(IFeedStorage)
class GzipFileFeedStorage(object):
    """
    Storage which exports data to a gzipped file.
    To use it, add

    ::

        FEED_STORAGES = {
            'gzip': 'myproject.exports.GzipFileFeedStorage',
        }

    to settings.py and then run scrapy crawl like this::

        scrapy crawl foo -o gzip:/path/to/items.jl

    The command above will create ``/path/to/items.jl.gz`` file
    (.gz extension is added automatically).

    Other export formats are also supported, but it is recommended to use .jl.
    If a spider is killed then gz archive may be partially broken.
    In this case it user should read the broken archive line-by-line and stop
    on gzip decoding errors, discarding the tail. It works OK with .jl exports.
    """
    COMPRESS_LEVEL = 4

    def __init__(self, uri):
        self.path = file_uri_to_path(uri) + ".gz"

    def open(self, spider):
        dirname = os.path.dirname(self.path)
        if dirname and not os.path.exists(dirname):
            os.makedirs(dirname, exist_ok=True)
        return gzip.open(self.path, 'ab', compresslevel=self.COMPRESS_LEVEL)

    def store(self, file):
        file.close()
```

What do you think?

redapple:
what about something along the lines of:
```
$ git diff
diff --git a/scrapy/extensions/feedexport.py b/scrapy/extensions/feedexport.py
index c3fc66d..27e0ad3 100644
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -163,6 +163,7 @@ class FeedExporter(object):
             raise NotConfigured
         self.format = settings['FEED_FORMAT'].lower()
         self.export_encoding = settings['FEED_EXPORT_ENCODING']
+        self.export_compression = settings.get('FEED_EXPORT_COMPRESSION', False)
         self.storages = self._load_components('FEED_STORAGES')
[...]

PR #4131, opened 2019-11-05
Add Compression with Associated Documentation

tianhuil:
Implementing https://github.com/scrapy/scrapy/issues/2174

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -262,7 +262,7 @@ Use ``utf-8`` if you want UTF-8 for JSON too.
 .. setting:: FEED_EXPORT_COMPRESSION
 
 FEED_EXPORT_COMPRESSION
---------------------
+-----------------------
 
 Default: ``False``
 
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -219,6 +219,7 @@ These are the settings used for configuring the feed exports:
  * :setting:`FEED_EXPORTERS`
  * :setting:`FEED_STORE_EMPTY`
  * :setting:`FEED_EXPORT_ENCODING`
+ * :setting:`FEED_EXPORT_COMPRESSION`
  * :setting:`FEED_EXPORT_FIELDS`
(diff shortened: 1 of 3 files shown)


docs/topics/feed-exports.rst
```diff
@@ -258,6 +259,15 @@ which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
...
 Use ``utf-8`` if you want UTF-8 for JSON too.
 
+.. setting:: FEED_EXPORT_COMPRESSION
+
+FEED_EXPORT_COMPRESSION
+-----------------------
+
+Default: ``False``
+
+Set to ``True`` to enable GZip compression.
```
Gallaecio: :lipstick: Maybe we should call it ‘gzip’ and link to https://en.wikipedia.org/wiki/Gzip for reference.

scrapy/extensions/feedexport.py
```diff
@@ -204,6 +204,7 @@ def __init__(self, settings):
             raise NotConfigured
         self.format = settings['FEED_FORMAT'].lower()
         self.export_encoding = settings['FEED_EXPORT_ENCODING']
+        self.export_compression = settings.get('FEED_EXPORT_COMPRESSION', False)
```
Gallaecio: I _think_ adding the default value to `default_settings.py` is enough, no need to add it here as well. And `Settings.getbool` should be used to ensure that changing the setting from environment variables, the command line and such works the same.

```suggestion
        self.export_compression = settings.getbool('FEED_EXPORT_COMPRESSION')
```

elacuesta:
Should we merge #3858 first and rebase this on top of it?

wRAR:
@tianhuil the multi-feed PR is now merged, feel free to continue working on this.

Koen87:
This method doesn't work combined with the S3 Feed Storage. I've been trying to make it work myself with the code changes in this PR but haven't had any success yet.

Gallaecio:
Implemented in https://github.com/scrapy/scrapy/pull/5190

PR #4140, opened 2019-11-07
Use the latest Python version to build the documentation

Gallaecio:


kmike:
I think we need to use the same Python version as we configure on readthedocs here. Is the plan to update readthedocs to use Python 3.8 then, is it possible?

Gallaecio:
Indeed, we should use the same. [It seems possible](https://docs.readthedocs.io/en/latest/config-file/v2.html#python-version).

.readthedocs.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+sphinx:
+  configuration: docs/conf.py
+python:
+  version: 3.8
+  install:
+    - requirements: docs/requirements.txt
```
Gallaecio: Only merge after #4169

PR #4148, opened 2019-11-12
Allow opening the source code from the API documentation

Gallaecio:


kmike:
Awesome, I was thinking about it just yesterday - thanks @Gallaecio for your telepathy skills!

PR #4150, opened 2019-11-12
Feature: XmlItemExporter can use custom tags for iterables.

lorlouis:
At the moment scrapy exports element in a list using the <value> tag and it is impractical to change this behavior.
My patch aims to make exporting an xml according to a specified formal easier.

- the function is 100% retro compatible
- you can define the default tag to be anything else using the "default_key" argument
- you can specify custom names for elements in a given iterable field using the "serialized_keys" dictionary 
e.g. serialized_keys={'array1':'CustomNameForArray1Elements'}
- you can choose to enable derived names so that fields with a plural name have elements with the same name but singular.
e.g. fruits[]'s elements will have the fruit tag

### DIFF
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -136,10 +136,7 @@ class XmlItemExporter(BaseItemExporter):
     def __init__(self, file, **kwargs):
         self.item_element = kwargs.pop('item_element', 'item')
         self.root_element = kwargs.pop('root_element', 'items')
-        self.serialized_keys = kwargs.pop('serialized_keys', dict())
-        self.default_key = kwargs.pop('default_key', 'value')
-        self.enable_derive = kwargs.pop('enable_derive', False)
-        self._configure(kwargs, dont_fail=True)
+        self._configure(kwargs)
         if not self.encoding:
             self.encoding = 'utf-8'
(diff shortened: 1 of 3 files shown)


Gallaecio:
I have several doubts about this.

I think an easier and more flexible approach would be to move list serialization into a separate method, so that subclassing the default XML exporter to cover these use cases is much easier.

lorlouis:
Would you prefer to get rid of theses features or keep them in the new method?

Gallaecio:
I would personally just extract that into a method, so that it’s easy to implement what you want in a subclass. I feel like the flexibility that your changes bring is not worth 3 Scrapy settings.

But maybe we should wait for other people to give their opinion, there is no need to rush a decision.

lorlouis:
We have been using scrapy at jobillico and we use xml to encode lots of data between internal and external systems so I thought packaging our solution more neatly into the repo would be a good idea, but it is true that more control over xml output isn't that useful for smaller projects.

lorlouis:
I removed the features their tests and moved the serialization of list like items into it's own method.

Gallaecio:
What I had in mind was something more in the lines of #4161. What do you think?

PR #4151, opened 2019-11-13
Deprecate BaseDupeFilter.log() and improve dupefilter docs

Gallaecio:
It is never called because request_seen() always returns False


kmike:
Won't it make it more confusing to understand what's the interface one have to implement to build a custom dupefilter?

Gallaecio:
Shall I make up for it with documentation?

nyov:
Or define an interface `IDupeFilter`, that is implemented by `BaseDupeFilter`?

Gallaecio:
We only seem to have 1 interface in Scrapy at the moment. I’m not sure which direction we should go, have more interfaces or have none.

nyov:
I agree, we should have a general directive for that. I don't know if `zope.interfaces` is still good to use, or out of style, or depending on context?
We have `ISpiderLoader` in interfaces.py, `IFeedStorage` in feedexport.py.
`pqueues` seem to have a `DownloaderInterface` class which is not a zope interface.
HTTP CacheStorages and pipelines probably _could_ be interfaces, but aren't.
So things are slightly confusing. I don't know if we should use it more, or less, either.

kmike:
I think this change might be backwards incompatible, because subclasses of BaseDupeFilter may implement .log() method which calls `super().log()`.

docs/topics/settings.rst
```diff
@@ -955,15 +955,65 @@ Default: ``'scrapy.dupefilters.RFPDupeFilter'``
...
-You can disable filtering of duplicate requests by setting
-:setting:`DUPEFILTER_CLASS` to ``'scrapy.dupefilters.BaseDupeFilter'``.
-Be very careful about this however, because you can get into crawling loops.
-It's usually a better idea to set the ``dont_filter`` parameter to
-``True`` on the specific :class:`~scrapy.Request` that should not be
-filtered.
+To change how duplicates are checked you can subclass
+:class:`~scrapy.dupefilters.RFPDupeFilter` and override its
+:meth:`~scrapy.dupefilters.RFPDupeFilter.request_fingerprint` method using the
+:func:`scrapy.utils.request.request_fingerprint` helper function.
```
wRAR, line 964: This now needs an update.

wRAR (approved):

docs/topics/settings.rst
```diff
@@ -955,15 +955,85 @@ Default: ``'scrapy.dupefilters.RFPDupeFilter'``
...
+
+        def __init__(
+            self,
+            path: str | None = None,
+            debug: bool = False,
+            *,
+            fingerprinter: RequestFingerprinterProtocol | None = None,
+        ) -> None:
+            super().__init__(
+                path=pat, debug=debug, fingerprinter=CustomRequestFingerprinter()
```
wRAR, line 987:
```suggestion
                path=path, debug=debug, fingerprinter=CustomRequestFingerprinter()
```

wRAR (approved):

PR #4152, opened 2019-11-13
Install the sphinx-notfound-page Sphinx extension

Gallaecio:


kmike:
I don't mind adding it, though do we have a problem which this extension solves? I have never seen a 404 page with broken styles in scrapy docs, when can this happen?

Gallaecio:
https://scrapy.readthedocs.io/en/latest/asdf

It happens on any page that no longer exists (e.g. if we ever move or rename a page). As things are at the moment, you get a page that does not follow the style of the rest of the pages and does not even allow you to go to an existing page.

With the extension, you get a similar style to existing pages, an the sidebar with the search box and the table of contents so that you can try and find the content you were looking for when you reached the non-existing page.

kmike:
I see, thanks for the explanation!

PR #4153, opened 2019-11-13
MutableChain: return self from __iter__

Gallaecio:
https://lgtm.com/rules/4850080/

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -388,9 +388,7 @@ class MutableChain(object):
         self.data = chain(self.data, *iterables)
 
     def __iter__(self):
-        return self.data.__iter__()
+        return self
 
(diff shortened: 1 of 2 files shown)


scrapy/utils/python.py
```diff
@@ -388,9 +388,7 @@ def extend(self, *iterables):
         self.data = chain(self.data, *iterables)
 
     def __iter__(self):
-        return self.data.__iter__()
+        return self
 
     def __next__(self):
         return next(self.data)
-
-    next = __next__
```
kmike: this one is technically backwards incompatible

Gallaecio: It is. I guess I can deprecate it instead.

kmike:
👍

PR #4155, opened 2019-11-13
Add a code complexity check to CI

Gallaecio:


kmike:
@Gallaecio could you please give an example of code where this thing would fail?

Gallaecio:
[This](https://github.com/scrapy/scrapy/files/3848351/complexity.txt) is what you get when you set all three parameters to A.

kmike:
Hm, for me it looks a bit non-transparent. I went throught some examples which had worse scores (C), and it was not immediately obvious what should actions be. 

What do you think about making it non-failing? I'm currently -0 for failing a CI build if this check fails, but just the output can be interesting to review. We can make it a bit more verbose in this case as well. If it can't fail a build, +0 to merge.

Gallaecio:
Yeah, it will be confusing for contributors when it fails. And since it is fast and easy to run locally, I think it might be better just not to add it to CI. I’ll use it locally when looking for room for improvement.

PR #4159, opened 2019-11-13
Add settings hook to allow overriding of Scraper class

### LINKED ISSUE #4156: Importance of Requests
## Summary

It would be nice to have the concept of request importance to affect the logging level of download failures, i.e. a way to yield a request but say, if this request has a download failure then only log a warning. A feature to keep track of how many requests of low importance have failed so far would also be useful, as there could be a threshold on the number of failures where error logging starts to kick back in.

## Motivation

We use scrapy extensively and we track log files to help us monitor scraping processes. 
Imagine a scrape that yields one request (R1) to start, then in the first callback it yields 10000 requests (R2s). From a monitoring perspective if R1 fails that is a huge problem, but if one of the R2s fails I don't really care. If lots of the R2s fail though that is a big deal.

If I see an error in the log files for code I did not write, it's not immediately clear to me if the failure is an R1 or R2 request, I have to go and read the code to find out.

## Additional context

It looks like all we need to do is override the Scraper class slightly, and maybe write new Request types. It's kind of hard to plug in a custom Scraper class currently, this line could be changed to pull a class from settings like the settings above it though https://github.com/scrapy/scrapy/blob/c911e802097ecd3309bb826d48b7b08ce108f4ce/scrapy/core/engine.py#L70
I thought I would reach out to see if this behavior is something that might be useful to others as part of the default Scraper, we're happy to contribute a PR if so.

FHTMitchell:
Put a hook in the settings to allow developers to implement their own scrapers (via subclassing the Scraper class). 

This allows overriding of the error handling when spiders fail; fixes #4156.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1241,6 +1241,14 @@ Type of priority queue used by the scheduler. Another available type is
 domains in parallel. But currently ``scrapy.pqueues.DownloaderAwarePriorityQueue``
 does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
 
+.. setting:: SCRAPER
+
+SCRAPER
+-------
+Default: ``'scrapy.core.scraper.Scraper'``
+
+The scraper to use for crawling.
+
 .. setting:: SPIDER_CONTRACTS
 
 SPIDER_CONTRACTS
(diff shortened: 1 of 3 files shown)


Gallaecio:
A setting exposing such a core part of Scrapy should include automated tests and extensive documentation: how to use it, which scenarios may require using it, etc.

PR #4161, opened 2019-11-14
Ease element name customization through XmlItemExporter subclassing

Gallaecio:
Note: Most likely breaks existing `XmlItemExporter` subclasses that override private methods.


lorlouis:
It is pretty good, but the use of 'breadcrumbs' to hold most of the data makes the whole thing more confusing than it should be in my opinion. And from my experience it is pretty rare that you have to change the tag of a serialized list to a single value, I would argue that replacing 'list_element' to a dictionary containing associations between specific container names and the value we wish to replace the tag with would be way more useful.

scrapy/exporters.py
```diff
@@ -144,47 +145,54 @@ def _beautify_newline(self, new_item=False):
...
         self.xg.endElement(self.root_element)
         self.xg.endDocument()
 
-    def _export_xml_field(self, name, serialized_value, depth):
-        self._beautify_indent(depth=depth)
-        self.xg.startElement(name, {})
+    def element(self, breadcrumbs, value):
+        if isinstance(breadcrumbs[-1], int):
+            return {'name': self.list_element}
+        assert isinstance(breadcrumbs[-1], str)
```
elacuesta, line 174: No `assert`s in production code :wink:

Gallaecio: Just in case: this is a safe use of `assert` in production code: the condition should always be `True`, it’s a way of telling readers (including coverage checkers) that conditions outside the current scope ensure that this is always `True`.

docs/topics/exporters.rst
```diff
@@ -236,7 +236,7 @@ PythonItemExporter
 XmlItemExporter
 ---------------
 
-.. class:: XmlItemExporter(file, item_element='item', root_element='items', \**kwargs)
+.. class:: XmlItemExporter(file, *, item_element='item', root_element='items', list_item_element='value', **kwargs)
```
Gallaecio, line 239: The new `*` is a documentation fix, because the class uses `**kwargs` for the following keyword arguments, which means they cannot be specified positionally.

I’m not sure why `\**` was there, but it is showing as such in the rendered documentation, and removing it works.

These both as issues that affect the code base as a whole, so I’ll create separate issues to deal with them globally.

Gallaecio:
Renamed `list_element` to `list_item_element` for accuracy.

PR #4164, opened 2019-11-14
Improve the performance of the DOWNLOAD_DELAY test

Gallaecio:
Take the ~30s `DOWNLOAD_DELAY` test (see #4163) and split it in 2 ~2s tests.


kmike:
Thanks @Gallaecio!

Issue #3288, opened 2018-06-11
Deprecation Warning of Twisted


Aqua-Dream:
May you do a little modification to avoid these two warnings?

> [py.warnings] WARNING: /.../scrapy/core/downloader/webclient.py:4: DeprecationWarning: twisted.web.client.HTTPClientFactory was deprecated in Twisted 16.7.0: please use https://pypi.org/project/treq/ or twisted.web.client.Agent instead
  from twisted.web.client import HTTPClientFactory

> [py.warnings] WARNING: /.../scrapy/core/downloader/contextfactory.py:51: DeprecationWarning: Passing method to twisted.internet.ssl.CertificateOptions was deprecated in Twisted 17.1.0. Please use a combination of insecurelyLowerMinimumTo, raiseMinimumTo, and lowerMaximumSecurityTo instead, as Twisted will correctly configure the method.
  acceptableCiphers=DEFAULT_CIPHERS)

Unknowny:
I've recently also started getting these and a few other warnings but it turned out to be one of the spiders in the project was importing a certain library which tinkered with the `warnings.simplefilter` and such. Was a bit of a headache to track it down. I'd check all foreign imports first if I were you @Aqua-Dream

Aqua-Dream:
Thanks @Unknowny 
I also found out that python will not display DeprecationWarning by default.

Gallaecio:
@kmike I can reproduce the first warning (made visible by pytest):

```
venv/lib/python2.7/site-packages/scrapy/core/downloader/webclient.py:4
  /home/adrian/scrapinghub/veritec-solutions-crawler/venv/lib/python2.7/site-packages/scrapy/core/downloader/webclient.py:4: DeprecationWarning: twisted.web.client.HTTPClientFactory was deprecated in Twisted 16.7.0: please use https://pypi.org/project/treq/ or twisted.web.client.Agent instead
    from twisted.web.client import HTTPClientFactory
```

We are indeed using `HTTPClientFactory`: https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/webclient.py#L4

elacuesta:
I find this somewhat frustrating. I could not find any release notes for v16.7.0 talking about this deprecation, in fact there is no mention of "16.7.0" in https://github.com/twisted/twisted/blob/trunk/NEWS.rst and https://twistedmatrix.com/documents/current/api/twisted.web.client.HTTPClientFactory.html says nothing about it being deprecated.
[...]

PR #4165, opened 2019-11-14
Merge Twisted’s deprecated HTTPClientFactory into ScrapyHTTPClientFactory

Gallaecio:
Fixes one of the warnings reported at #3288


wRAR:
Should we mention in the source that this is a copy of some Twisted code?

PR #4170, opened 2019-11-18
Raise error when start_url found instead of start_urls.

### LINKED ISSUE #4133: Handle it gracefully when start_url is used instead of start_urls
Over the last year I’ve seen a few cases ([recent example](https://stackoverflow.com/q/58664004/939364)) of this, people missing the `s` at the end of the `start_urls`.

It may be nice to find a way to gracefully let the developer know where the issue is, why there is no crawling happening.

mabelvj:
Fixes #4133

Raise AttributeError error when empty 'start_urls' and 'start_url' found. Added test.


scrapy/spiders/__init__.py
```diff
@@ -68,6 +68,10 @@ def _set_crawler(self, crawler):
 
     def start_requests(self):
         cls = self.__class__
+        if not any(self.start_urls) and hasattr(self, 'start_url'):
+            raise AttributeError(
+                "Crawling could not start: 'start_urls' list not found (found"
+                "'start_url={%s}' instead)." % getattr(self, 'start_url'))
```
Gallaecio:
```suggestion
                "a 'start_url' attribute instead, did you miss an 's'?).")
```

scrapy/spiders/__init__.py
```diff
@@ -68,6 +68,10 @@ def _set_crawler(self, crawler):
 
     def start_requests(self):
         cls = self.__class__
+        if not any(self.start_urls) and hasattr(self, 'start_url'):
+            raise AttributeError(
+                "Crawling could not start: 'start_urls' list not found (found"
```
Gallaecio:
```suggestion
                "Crawling could not start: 'start_urls' list not found (found "
```

scrapy/spiders/__init__.py
```diff
@@ -68,6 +68,11 @@ def _set_crawler(self, crawler):
 
     def start_requests(self):
         cls = self.__class__
+        if not self.start_urls and hasattr(self, 'start_url'):
+            raise AttributeError(
+                "Crawling could not start: 'start_urls' list not found "
```
elacuesta: I know [the docs](https://docs.scrapy.org/en/1.8/topics/spiders.html#scrapy.spiders.Spider.start_urls) do mention a "list", but since it could technically be [any iterable](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/spiders/__init__.py#L69-L70) I would remove the "list" part. Maybe "the 'start_urls' attribute" is good enough.

PS: looking at this, I figured it would be good to remove the `make_requests_from_url` method, it's been deprecated for a long time now. I'll open a PR for that.

mabelvj:
I am having trouble with the tests since it seems we should start a crawling for getting the result from `start_requests` -> https://github.com/scrapy/scrapy/pull/4170/commits/96f7e976fe4a9302b05ab5cec8def480bf58a862.

Maybe just checking when trying to set the variable could be a good solution.

scrapy/spiders/__init__.py
```diff
@@ -23,6 +23,13 @@ class Spider(object_ref):
     name = None
     custom_settings = None
 
+    def __setattr__(self, attr, value):
+        if attr == 'start_url' and not getattr(self, 'start_urls', []):
```
Gallaecio: Won’t this behave differently depending on the order in which `start_url` and `start_urls` appear in the `Spider` subclass?

kmike:
Hey! There are real-world spiders which use start_url option, as a way to customize things (found them using our internal code search). Is there a way to distinguish between intended and unintended usage? It is not good to show a warning for a perfectly valid use case.

I wonder if a better fix would be to have a warning in the default start_requests implementation: if start_urls is None, but start_url is present, then show a warning. There won't be a warning if start_requests function is using `start_url` attribute in this case, as it'd be overridden. 

It could make sense to check real spiders which use start_request attribute, to see how any change we make affect them.

mabelvj:
You are right @kmike, that was one of my other options. Setting this warning in `start_requests` will allow overriding it when using `start_url`.

scrapy/spiders/__init__.py
mabelvj: You are right. I updated it in any case to follow the previous approach, do the check in start_requests.

kmike:
Thanks @mabelvj!

PR #4172, opened 2019-11-18
Use InterSphinx to link to the Twisted documentation

Gallaecio:


kmike:
Nice cleanup, thanks @Gallaecio.

PR #4179, opened 2019-11-20
Make tox configuration more user friendly

Gallaecio:
- Change the default environments to `security`, `flake` and `py3`.
- Do not enforce a specific Python 3 interpreter in most tox environments, let them pick whichever is available.
- Enforce the minimum tox version (`1.7.0`) from the tox configuration, not the documentation.
- Sync the sorting of tox environment configurations with Travis CI.


tox.ini
```diff
@@ -48,34 +66,11 @@ deps =
...
     Pillow==3.4.2
 
-[testenv:py36]
-basepython = python3.6
-
-[testenv:py37]
-basepython = python3.7
-
-[testenv:py38]
-basepython = python3.8
```
kmike: Is there a way now to run tests for multiple interpreters locally?

Gallaecio: Yes, `py35`, `py36` and so on still work as expected nonetheless (notice I left them being used in Travis CI). `tox` seem to handle them “magically”.

kmike:
Thanks @Gallaecio for the cleanup!

PR #4186, opened 2019-11-21
Remove unused imports

Gallaecio:
https://lgtm.com/rules/6770079/

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -12,7 +12,6 @@ from six.moves.urllib.parse import (ParseResult, urldefrag, urlparse, urlunparse
 # scrapy.utils.url was moved to w3lib.url and import * ensures this
 # move doesn't break old code
 from w3lib.url import *
-from w3lib.url import _safe_chars, _unquotepath
 from scrapy.utils.python import to_unicode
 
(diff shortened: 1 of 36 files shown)


scrapy/utils/url.py
```diff
@@ -12,7 +12,6 @@
 # scrapy.utils.url was moved to w3lib.url and import * ensures this
 # move doesn't break old code
 from w3lib.url import *
-from w3lib.url import _safe_chars, _unquotepath
```
kmike: I think this one was intentional

tox.ini
```diff
@@ -73,6 +73,7 @@ commands =
 basepython = python3.8
 deps =
     {[testenv]deps}
+    -r docs/requirements.txt
```
kmike: is this change related to removal of unneeded imports?

Gallaecio: When cleaning up `pytest.ini`, I accidentally removed the `--ignore=docs/conf.py` line, which made `flake8` tests require doc dependencies. I reverted that line removal, but forgot to revert this. :man_facepalming:

kmike:
Thanks @Gallaecio!

PR #4192, opened 2019-11-25
Make ‘[source]’ link to Github

Gallaecio:
This is probably more useful than `viewcode`, as GitHub’s code browser is quite advanced (e.g. you can hover objects and get extra information), and users can for example create quick pull requests from the GitHub page itself to fix typos and the likes.

I tested these changes locally and in a [personal ReadTheDocs account](https://scrapy-gallaecio.readthedocs.io/en/master/topics/request-response.html), as I was not confident enough about the link-resolution code without testing it first. In that ReadTheDocs accounts you can see how `master` links to the `master` branch while `latest` links to the `1.8.0` tag.


docs/conf.py
```diff
@@ -285,7 +287,67 @@
...
 
 hoverxref_auto_ref = True
+
+
+# Options for the linkcode extension
+# ----------------------------------
+
+# Based on numpy’s:
+# https://github.com/numpy/numpy/blob/dedc4178fc334329de9872ab42df870d2ac7a270/doc/source/conf.py#L313
+def linkcode_resolve(domain, info):
```
kmike: Hey! My only concern is that we get a large untested function in a config file. Is it possible to move it somewhere else, and add some tests for it (at least doctests)?

Gallaecio: Moved into a separate extension, implementation simplified for our use case, and tests added.

Gallaecio:
https://github.com/scrapy/scrapy/pull/7058

PR #4193, opened 2019-11-25
Use super().__init__ in BaseItemExporter subclasses

Gallaecio:
https://lgtm.com/rules/5990072/

This change is not necessary from a functional perspective, but instead of marking the code to be ignored, I though it may be best to switch to an implementation that allows calling `super().__init__`.


kmike:
Nice cleanup, thanks @Gallaecio!

PR #4195, opened 2019-11-26
Keep 2 spaces between code and inline comments

Gallaecio:
Please, pay special attention to the change to `tests/test_engine.py`. I’m hoping tests will pass, but even if they do, I wonder what consequences this change may have.

I also removed a class from a test file, as it did not seem used anymore.

### DIFF
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -84,8 +84,7 @@ DOWNLOADER = 'scrapy.core.downloader.Downloader'
 DOWNLOADER_HTTPCLIENTFACTORY = 'scrapy.core.downloader.webclient.ScrapyHTTPClientFactory'
 DOWNLOADER_CLIENTCONTEXTFACTORY = 'scrapy.core.downloader.contextfactory.ScrapyClientContextFactory'
 DOWNLOADER_CLIENT_TLS_CIPHERS = 'DEFAULT'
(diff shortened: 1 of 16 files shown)


scrapy/settings/default_settings.py
```diff
@@ -84,8 +84,9 @@
 DOWNLOADER_HTTPCLIENTFACTORY = 'scrapy.core.downloader.webclient.ScrapyHTTPClientFactory'
 DOWNLOADER_CLIENTCONTEXTFACTORY = 'scrapy.core.downloader.contextfactory.ScrapyClientContextFactory'
 DOWNLOADER_CLIENT_TLS_CIPHERS = 'DEFAULT'
-DOWNLOADER_CLIENT_TLS_METHOD = 'TLS' # Use highest TLS/SSL protocol version supported by the platform,
-                                     # also allowing negotiation
+# Use highest TLS/SSL protocol version supported by the platform, also allowing
+# negotiation:
```
elacuesta: Nitpick: Could this be included in the line above? Even if we stick to the 79-char limit (which I believe we shouldn't, at least we should allow for 99 chars IMHO), I don't think this file could ever comply to that without some seriously ugly backslashing, given for instance that the line `DOWNLOADER_CLIENTCONTEXTFACTORY = 'scrapy.core.downloader.contextfactory.ScrapyClientContextFactory'` is 100 chars long.

PR #4197, opened 2019-11-27
[Docs] Fix Twisted links

elacuesta:
The current link for the Twisted API is not working for me when I generate the docs. There are no public links to show, since this change has not been deployed to the main docs yet.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -279,7 +279,7 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'sphinx': ('https://www.sphinx-doc.org/en/master', None),
     'tox': ('https://tox.readthedocs.io/en/latest', None),
-    'twisted': ('https://twistedmatrix.com/documents/current', None),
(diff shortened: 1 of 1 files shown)


Gallaecio:
This breaks links to topics. We may need two different mappings, one for the API and one for the topics.

docs/conf.py
```diff
@@ -279,7 +279,7 @@
     'python': ('https://docs.python.org/3', None),
     'sphinx': ('https://www.sphinx-doc.org/en/master', None),
     'tox': ('https://tox.readthedocs.io/en/latest', None),
-    'twisted': ('https://twistedmatrix.com/documents/current', None),
+    'twisted': ('https://twistedmatrix.com/documents/current/api', None),
```
Gallaecio: Try simply adding a `'twistedapi': ('https://twistedmatrix.com/documents/current/api', None),` line.

elacuesta: Great, that seems to work. Thanks!

kmike:

Issue #4187, opened 2019-11-21
Add new tests for the proxy noconnect mode


wRAR:
This is related to #2545 and #4114.

Currently we cannot test the proxy noconnect mode because we use mitmproxy as a test proxy and it doesn't seem to support this mode in the current version. Because of that, the tests for this mode are currently disabled.

As described in a comment near the old test functions, now marked as expected failures:

> The noconnect mode isn't supported by the current mitmproxy, it returns
> "Invalid request scheme: https" as it doesn't seem to support full URLs in GET at all,
> and it's not clear what behavior is intended by Scrapy and by mitmproxy here.
> https://github.com/mitmproxy/mitmproxy/issues/848 may be related.
> The Scrapy noconnect mode was required, at least in the past, to work with Crawlera,
> and https://github.com/scrapy-plugins/scrapy-crawlera/pull/44 seems to be related.

PR #4198, opened 2019-11-28
Deprecate the HTTPS proxy noconnect mode.

wRAR:
This deprecates the noconnect mode for HTTPS proxies and removes related tests. As the mode was made for old Crawlera, it doesn't look like anything else needs it.

scrapy_crawlera version is important because before 1.3.0 it added '?noconnect' to all Crawlera URLs automatically.

Closes: #4187 (as not needed anymore).

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -16,6 +16,7 @@ from twisted.web.http import _DataLoss, PotentialDataLoss
 from twisted.web.client import Agent, ResponseDone, HTTPConnectionPool, ResponseFailed, URI
 from twisted.internet.endpoints import TCP4ClientEndpoint
 
+from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.http import Headers
 from scrapy.responsetypes import responsetypes
 from scrapy.core.downloader.webclient import _parse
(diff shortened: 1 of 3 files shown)


scrapy/core/downloader/handlers/http11.py
```diff
@@ -285,6 +286,12 @@ def _get_agent(self, request, timeout):
             scheme = _parse(request.url)[0]
             proxyHost = to_unicode(proxyHost)
             omitConnectTunnel = b'noconnect' in proxyParams
+            if omitConnectTunnel:
+                warnings.warn("Using HTTPS proxies in the noconnect mode is deprecated. "
+                              "If you use Crawlera, it doesn't require this mode anymore, "
+                              "so you should update scrapy-crawlera to 1.3.0+ "
+                              "and remove '?noconnect' from the Crawlera URL.",
+                              ScrapyDeprecationWarning)
```
Gallaecio: I’m guessing we need to use a higher stack level to point to the user code where noconnect is used.

wRAR: I'm afraid it's not possible as there is a lot of unrelated frames there, also the user code is usually in `settings.py`, in `CRAWLERA_URL`.

tests/test_downloader_handlers.py
```diff
@@ -687,16 +687,6 @@ def _test(response):
         request = Request('http://example.com', meta={'proxy': http_proxy})
         return self.download_request(request, Spider('foo')).addCallback(_test)
 
-    def test_download_with_proxy_https_noconnect(self):
```
Gallaecio: Can we keep this test and modify it just to verify that the warning is logged?

kmike:
👍

Issue #1363, opened 2015-07-16
Deprecate `download_delay` spider attribute


barraponto:
Since we now have custom settings per spider, I don't see any reason to support the `download_delay` attribute anymore (other than backwards compatibility). Should we deprecate it?
 
http://doc.scrapy.org/en/latest/topics/settings.html#download-delay

PR #4201, opened 2019-11-30
Add deprecate warning for Spider.download_delay attribute

GeniusLearner:
Closes: #1363

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -59,6 +59,8 @@ class Slot(object):
 def _get_concurrency_delay(concurrency, spider, settings):
     delay = settings.getfloat('DOWNLOAD_DELAY')
     if hasattr(spider, 'download_delay'):
+        warnings.warn("download_delay is deprecated"
+                      category=ScrapyDeprecationWarning, stacklevel=2)
         delay = spider.download_delay
 
     if hasattr(spider, 'max_concurrent_requests'):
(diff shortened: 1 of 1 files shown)


GeniusLearner:
This is for issue #1363

wRAR:
Hi @GeniusLearner , please test your modifications before submitting them. You've forgot a comma.

scrapy/core/downloader/__init__.py
```diff
@@ -59,6 +59,8 @@ def __str__(self):
 def _get_concurrency_delay(concurrency, spider, settings):
     delay = settings.getfloat('DOWNLOAD_DELAY')
     if hasattr(spider, 'download_delay'):
+        warnings.warn("download_delay is deprecated"
```
elacuesta:
```suggestion
        warnings.warn("The Spider.download_delay attribute is deprecated",
```
In addition to @wRAR's comment, I think the current message doesn't make it clear that what's deprecated is the _spider_ attribute.

scrapy/core/downloader/__init__.py
```diff
@@ -59,6 +59,8 @@ def __str__(self):
 def _get_concurrency_delay(concurrency, spider, settings):
     delay = settings.getfloat('DOWNLOAD_DELAY')
     if hasattr(spider, 'download_delay'):
+        warnings.warn("The Spider.download_delay attribute is deprecated",
+                      category=ScrapyDeprecationWarning, stacklevel=2)
```
Gallaecio: `warnings` and `ScrapyDeprecationWarning` are not imported in this file, according to the flake8 test. Also, please add a test that verifies that this warning is indeed logged; there should be a few existing, similar tests to serve as inspiration.

noviluni: Hi @GeniusLearner

To do it, it should be as easier as doing:

```python
with warnings.catch_warnings(record=True) as w:
    <whatever you need to do>

    self.assertEqual(len(w), 1)
    self.assertIn("Ignoring invalid sitemap", w[0].message)
```

You can also use `self.assertWarns` and `self.assertWarnsRegex`.

Issue #3880, opened 2019-07-14
Does scrapy.Requests has streaming method?


lycanthropes:
I want download many PDF files, and I want to use scrapy to complete this task. But I find scrapy couldn't download them completely, there are always 1KB files in the result. So I think it is necessary to use a method such as requests.streaming in requests module.

kmike:
Yes, that'd be a great feature to add.

lycanthropes:
That is good. Thank you for listening to my advice ,and I am looking forward to it.

Gallaecio:
A potential use case would be to only fetch the response body until the desired data is found, reducing network traffic: https://stackoverflow.com/q/59541755/939364

chetan1029:
Yes most of the time we just need to scrap the title of the page so if is it possible to stream the data in chunks and skip the rest of the chunks.

PR #4205, opened 2019-12-02
Add bytes_received signal

elacuesta:
As discussed in https://github.com/scrapy/scrapy/issues/3793#issuecomment-496307176


scrapy/core/downloader/handlers/http11.py
```diff
@@ -454,6 +473,12 @@ def dataReceived(self, bodyBytes):
         self._bodybuf.write(bodyBytes)
         self._bytes_received += len(bodyBytes)
 
+        self._crawler.signals.send_catch_log(
+            signal=signals.bytes_received,
+            data=bodyBytes,
+            request=self._request,
+        )
```
Gallaecio: When the signal is not assigned to any handler, does this have any noticeable impact on performance?

elacuesta: Using https://github.com/scrapy/scrapy-bench:

`scrapy-bench --n-runs=10 --book_url=http://localhost:8000 bookworm`

scrapy @ 454de31d20d1d7676c8db0efaf46da4c671f81e6 (`bytes_received_signal` branch)
```
The results of the benchmark are (all speeds in items/sec) : 
Test = 'Book Spider' Iterations = '10'
Mean : 62.16695929434989 Median : 62.43078124454085 Std Dev : 1.1148966043160284
```

scrapy @ 07b8cd28aa84fb322a072467b49e2557d4aa3881 (`master` branch)
```
The results of the benchmark are (all speeds in items/sec) : 
Test = 'Book Spider' Iterations = '10'
Mean : 60.77521979854436 Median : 60.963007266556204 Std Dev : 1.6349215293441601
```

docs/topics/signals.rst
```diff
@@ -158,6 +158,23 @@ item_error
     :param failure: the exception raised
     :type failure: twisted.python.failure.Failure
 
+bytes_received
+--------------
+
+.. signal:: bytes_received
+.. function:: bytes_received(data, request)
```
kmike: Could you please add spider argument to this signal as well, like for other signals?

docs/topics/signals.rst
```diff
@@ -158,6 +158,23 @@ item_error
...
+
+    Sent by the HTTP 1.1 download handler when a group of bytes is
+    received for a specific request.
```
kmike: I think it makes sense to call out that this signal can be sent multiple times for a single request, with partial data each time.

tests/test_engine.py
```diff
@@ -263,19 +271,54 @@ def _assert_scraped_items(self):
                 self.assertEqual('Item 2 name', item['name'])
                 self.assertEqual('200', item['price'])
 
-    def _assert_signals_catched(self):
-        assert signals.engine_started in self.run.signals_catched
-        assert signals.engine_stopped in self.run.signals_catched
-        assert signals.spider_opened in self.run.signals_catched
-        assert signals.spider_idle in self.run.signals_catched
-        assert signals.spider_closed in self.run.signals_catched
+    def _assert_bytes_received(self):
```
kmike: Could you please add a test which shows the signal behavior when the data is larger than a buffer, i.e. when a signal is sent several times by Scrapy?

docs/topics/signals.rst
```diff
@@ -158,6 +158,23 @@ item_error
     :param failure: the exception raised
     :type failure: twisted.python.failure.Failure
 
+bytes_received
+--------------
+
+.. signal:: bytes_received
+.. function:: bytes_received(data, request)
+
+    Sent by the HTTP 1.1 download handler when a group of bytes is
```
kmike: I wonder if a signal should be renamed, if it is just for HTTP 1.1 download handler. 

Is it sent e.g. for S3 responses, or for FTP responses? It could be, because S3DownloadHandler uses http download handler,though I haven't checked it. If it is sent, we should document that. If it is not sent for all handlers, imagine in future we decide to add support for them (e.g. S3 handler starts to send this signal). It'd be a backwards incompatible change, as user code may rely on bytes_received to be only issues by HTTP 1.1 download handler, not by S3 or FTP. Having different signal names can avoid this confusion. Alternatively, we can decide to support everything from the beginning.

tests/test_engine.py
```diff
@@ -155,6 +159,9 @@ def item_error(self, item, response, spider, failure):
     def item_scraped(self, item, spider, response):
         self.itemresp.append((item, response))
 
+    def bytes_received(self, data, request):
+        self.bytes[request] += data
```
kmike: I know this is only tests, but += is O(N^2) for bytes in Python 3. It is O(N) in CPython for str (because of some bytecode tricks), and it used to be O(N) in Python 2 for bytes, but this optimization was lost in transition.

tests/test_engine.py
elacuesta: Added in 7753e6bed0aabc73b336dcdcf314f6b4e7be9855.
Not sure about the `2**14` thing though, it doesn't seem to slow the test down but there might be a better way to force a smaller buffer size.

Running only `tests/test_engine.py` with the large payload:
```
real	0m4.765s
user	0m2.445s
sys	0m0.330s
```

without it:
```
real	0m4.751s
user	0m2.367s
sys	0m0.338s
```

docs/topics/signals.rst
elacuesta: Just pushed 8beb77c3f33eeb16b42719fa3a263c840ddaaf49, which adds a `source` parameter to the signal (current values could be "http11" or "s3"). I'm not sure how to test the occurrence of the "s3" value without actually requesting stuff from S3.
Let me know what you think of this approach @kmike.

tests/test_engine.py
```diff
@@ -263,19 +279,58 @@ def _assert_scraped_items(self):
...
+                    b"\n<html>\n"
+                    b"  <head><title>404 - No Such Resource</title></head>\n"
+                    b"  <body>\n"
+                    b"    <h1>No Such Resource</h1>\n"
+                    b"    <p>File not found.</p>\n"
+                    b"  </body>\n"
+                    b"</html>\n"
+                )
+            elif self.run.getpath(request.url) == "/random":
+                self.assertTrue(len(data) > 1)  # signal was fired multiple times
```
kmike: Is it tested what's the content of the data chunks received, i.e. that b''.join(data) gets the full data, and that it is not that the signal receives e.g. ABC first, then ABCDEF, then ABCDEFGHJ, etc?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -409,7 +416,16 @@ def _cancel(_):
...
-            _ResponseReader(d, txresponse, request, maxsize, warnsize, fail_on_dataloss)
+            _ResponseReader(
+                d,
+                txresponse,
+                request,
+                maxsize,
+                warnsize,
+                fail_on_dataloss,
+                self._crawler,
+                self._source,
```
kmike: Could you please switch to keyword arguments here?

Gallaecio:
I wonder if we should consider https://github.com/scrapy/scrapy/issues/3880#issuecomment-572043373 and allow the signal to somehow stop the response download and send the response to the corresponding callback as is.

Although that could wait for a future pull request, and I’m not sure if a signal is the right place to expose such a functionality.

elacuesta:
That's an interesting use case, though I agree signal handlers might not be the best place to do it.

docs/topics/signals.rst
kmike: Let's check pro/cons of these aproaches, from user point of view.

**Use case 1**: user wants to check traffic received from websites. Two signals:

```py
def on_bytes_received_http(data):
    # ...

crawler.signals.connect(on_bytes_received_http, signals.bytes_received_http)
```

One signal, "source" field:

```py
def on_bytes_received(data, source):
    if source != 'http11':
        return
    # ...

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

**Use case 2**: user wants to check traffic received from websites and from downloading s3 files (e.g. with seed urls used in start_requests?), but wants to handle them differently:

```py
def on_bytes_received_http(data):
    # ...

def on_bytes_received_s3(data):
    # ...

crawler.signals.connect(on_bytes_received_http, signals.bytes_received_http)
crawler.signals.connect(on_bytes_received_s3, signals.bytes_received_s3)
```

One signal, "source" field:

```py
def on_bytes_received(data, source):
    if source == 'http11':
        # ...
    elif source == 's3':
        # ...

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

**Use case 3**: user wants overall downloaded data stats, source doesn't matter.

Two signals:

```py
def on_bytes_received(data):
    # ...

crawler.signals.connect(on_bytes_received, signals.bytes_received_http)
crawler.signals.connect(on_bytes_received, signals.bytes_received_s3)
```

One signal, "source" field:

```py
def on_bytes_received(data, source):
    # ...

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

To my taste, "two signals" approach 

* wins in Use case 1 (shorter, less error prone)
* wins in Use case 2 (less error prone, but more code)
* loses in Use case 3 (more code, need to find all signals related to downloading)

In first two use cases "less error prone" means that if you make a typo in a signal name, an error will be raised (there is no such a signal), while if you check against non-existing source, it will silently work wrong. Because of this "less error prone" I think having 2 separate signals is better than relying on "source" field.

For the stats use case a single signal could make sense, which is fired if any other bytes_received signal are fired. It will also allow to add more signal sources without users having to update their code. It can actually be the current bytes_received signal. But I wonder what's the performance overhead of that, having to send 2x signals on every chunk, as compared to benefit (it may still worth it).

---

One more thing, I'm not sure http11 is a good name here, because it may be set for http1.0 traffic as well - though I may be wrong :) I'm assuming that HTTP11DownloadHandler handles communication with all kind of of http servers, but maybe I'm wrong.

Gallaecio: I was in favor of multiple signals, but re-reading I’m starting to think that a single signal is much cleaner and should work, as long as we clearly document that additional download handlers may support this signal in the future, and recommend to use the `request` object to filter out non-HTTP traffic if desired.

This does not allow to know whether the traffic comes from HTTP 1.0, 1.1 or 2.0. Do we want to allow that? I don’t think we need that at the moment, and in the future we could implement a separate, new signal, e.g. `protocol_negotiated` or similar, that provides this information along with the affected request.

elacuesta: That's absolutely true, thanks for pointing it out. The request object should be enough to determine whether or not the source is S3.
Regarding the version, we do not support HTTP/2 as far as I remember, and HTTP/1.0 needs to be explicitly enabled by changing the download handler. +1 on the `protocol_negotiated` signal once we add support for HTTP/2.
I'll go ahead and simplify the implementation, thanks both ❤️

Gallaecio: @kmike These would be the 3 use cases with a single signal and relying on `request.url`:

**Use case 1**: user wants to check traffic received from websites:

```py
def on_bytes_received(data, request, spider):
    if urlparse(request.url).protocol not in {'http', 'https'}:
        return
    # …

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

**Use case 2**: user wants to check traffic received from websites and from downloading s3 files (e.g. with seed urls used in start_requests?), but wants to handle them differently:

```py
def on_bytes_received(data, request, spider):
    protocol = urlparse(request.url).protocol
    if protocol in {'http', 'https'}:
        # …
    elif protocol == 's3':
        # …

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

**Use case 3**: user wants overall downloaded data stats, source doesn't matter:

```py
def on_bytes_received(data, request, spider):
    # …

crawler.signals.connect(on_bytes_received, signals.bytes_received)
```

docs/topics/signals.rst
```diff
@@ -370,6 +370,27 @@ request_left_downloader
...
 
+bytes_received
+~~~~~~~~~~~~~~
+
+.. signal:: bytes_received
+.. function:: bytes_received(data, request, spider)
+
+    Sent by the HTTP 1.1 and S3 download handlers when a group of bytes is
+    received for a specific request. This signal might be fired multiple
+    times for the same request, with partial data each time.
```
kmike, line 381: There are 2 ways to return partial data: 

1) return data which is received so far (in this case "partial" means that the response is not fully received yet)
2) return a next received chunk

Would it be possible to clarify in the docs, say that (2) is correct?

docs/topics/signals.rst
kmike: Ok, fair enough. Let's proceed with a single signal, without source :)

kmike: Re examples, nitpick (in case we decide to put them somewhere, or if someone gets to this page via search) - they should use `scheme = urlparse_cached(request).scheme`, like scrapy's DownloadHandlers.

docs/topics/signals.rst
elacuesta: Updated in e1948b492317eb5b11550d119d91c61b74b3a37f

kmike:
Thanks @elacuesta!

PR #4207, opened 2019-12-04
Use a non-zero exit code when a pipeline's open_spider method throws an exception

### LINKED ISSUE #4175: Scrapy does not use a non-zero exit code when pipeline's open_spider throws the exception
### Description
In our case, we execute command `scrapy crawl` in airflow task and the exit code would be used to judge this task success or failure. I agree that `scrapy crawl` ignores spider exceptions because it's unpredictable in the crawling process. 

Back to our case, we export data to file or database in the pipeline and we create the directory or database connection in `open_spider(self, spider)`. I think if there is an exception happens during this function, it's reasonable to propagate a non-zero exit code. it because we normally do some initialization in this function.

### Steps to Reproduce

- scrapy startproject test_spider
- cd test_spider
- scrapy genspider example example.com
- modify spiders/example.py to  
```
# -*- coding: utf-8 -*-
import scrapy


    name = 'example'
    allowed_domains = ['example.com']
    start_urls = ['http://example.com/']

    custom_settings = {
        'ITEM_PIPELINES': {
            'test_spider.pipelines.TestSpiderPipeline': 300
        }
    }

    def parse(self, response):
        pass
```
- modify pipelines.py to 
```
# -*- coding: utf-8 -*-

# Define your item pipelines here
#
# Don't forget to add your pipeline to the ITEM_PIPELINES setting
# See: https://docs.scrapy.org/en/latest/topics/item-pipeline.html


class TestSpiderPipeline(object):

        raise Exception('error')

        return item
```
- scrapy crawl example
- echo $?  

return non-zero exit code

return zero exit code

100%

### Versions
Scrapy       : 1.8.0
lxml         : 4.3.3.0
libxml2      : 2.9.9
cssselect    : 1.0.3
parsel       : 1.5.1
w3lib        : 1.20.0
Twisted      : 19.2.0
Python       : 3.7.3 (default, Mar 27 2019, 09:23:39) - [Clang 10.0.0 (clang-1000.11.45.5)]
pyOpenSSL    : 19.0.0 (OpenSSL 1.1.1b  26 Feb 2019)
cryptography : 2.6.1
Platform     : Darwin-18.5.0-x86_64-i386-64bit

### Additional context

I could get the expected behavior if I change `def run(self, args, opts)` in scrapy/commands/crawl.py  to 
```
    def run(self, args, opts):
        if len(args) < 1:
            raise UsageError()
        elif len(args) > 1:
            raise UsageError("running 'scrapy crawl' with more than one spider is no longer supported")
        spname = args[0]

        res = self.crawler_process.crawl(spname, **opts.spargs)

        if hasattr(res, 'result') and res.result is not None and issubclass(res.result.type, Exception):
            self.exitcode = 1
        else:
            self.crawler_process.start()

            if self.crawler_process.bootstrap_failed:
                self.exitcode = 1
```
original `def run(self, args, opts)`
```
    def run(self, args, opts):
        if len(args) < 1:
            raise UsageError()
        elif len(args) > 1:
            raise UsageError("running 'scrapy crawl' with more than one spider is no longer supported")
        spname = args[0]

        self.crawler_process.crawl(spname, **opts.spargs)
        self.crawler_process.start()

        if self.crawler_process.bootstrap_failed:
            self.exitcode = 1
```

Is it the proper way to modify the code for achieving this purpose? if it is, could I create a PR request for this issue?

gunblues:
Fixes #4175


scrapy/commands/crawl.py
```diff
@@ -54,8 +54,16 @@ def run(self, args, opts):
...
+        if hasattr(res, 'result') and res.result is not None and issubclass(res.result.type, Exception):
             self.exitcode = 1
+        else:
+            self.crawler_process.start()
+
+            if self.crawler_process.bootstrap_failed or \
+                    (hasattr(self.crawler_process, 'has_exception') and self.crawler_process.has_exception):
+                self.exitcode = 1
+
+
```
wRAR: flake8 failed because of these extra blank lines

scrapy/commands/crawl.py
```diff
@@ -54,8 +54,16 @@ def run(self, args, opts):
...
 
-        self.crawler_process.crawl(spname, **opts.spargs)
+        res = self.crawler_process.crawl(spname, **opts.spargs)
         self.crawler_process.start()
 
-        if self.crawler_process.bootstrap_failed:
+        if hasattr(res, 'result') and res.result is not None and issubclass(res.result.type, Exception):
             self.exitcode = 1
+        else:
+            self.crawler_process.start()
```
wRAR: So this will run twice?

scrapy/commands/crawl.py
gunblues: already removed the extra blank line, thanks

scrapy/commands/crawl.py
gunblues: sorry, I didn't notice it. I've removed the redundant code. it should start if no exception after executing self.crawler_process.crawl()

scrapy/commands/crawl.py
```diff
@@ -54,8 +54,14 @@ def run(self, args, opts):
             raise UsageError("running 'scrapy crawl' with more than one spider is no longer supported")
         spname = args[0]
 
-        self.crawler_process.crawl(spname, **opts.spargs)
-        self.crawler_process.start()
-
-        if self.crawler_process.bootstrap_failed:
+        res = self.crawler_process.crawl(spname, **opts.spargs)
```
Gallaecio: :lipstick: Can we rename `res` to something that cannot be confused with `resource`, or mean `result` leading `res.result` to mean ‘result of result’? 

I’m thinking something in the lines of `process_defer`, `crawl_defer`… I’m not really sure, but `res` feels too ambiguous.

scrapy/commands/crawl.py
```diff
@@ -54,8 +54,14 @@ def run(self, args, opts):
...
+        
+        if hasattr(res, 'result') and res.result is not None and issubclass(res.result.type, Exception):
```
Gallaecio: :lipstick: I think we can make this shorter:

```suggestion
        if getattr(res, 'result') is not None and issubclass(res.result.type, Exception):
```

scrapy/commands/crawl.py
Gallaecio: Still failing. I recommend you run flake8 locally: `tox -e flake8`.

scrapy/commands/crawl.py
gunblues: good suggestion, I'll modify res to crawl_defer

scrapy/commands/crawl.py
gunblues: cool, I'll use this shorter code

scrapy/commands/crawl.py
gunblues: I've fixed it, thanks

gunblues:
> A change like this should have a test. Could you please add one?
> 
> Also, after your latest change to remove redundant code, I don’t think it will work. `res` is a deferred that will not be processed until `start` is called, and you are now checking it for errors before `start` is called.

We can get res.result.type and it belongs Exception when the exception is raised at open_spider in the pipeline. 

Otherwise, I'll try to think about how to add a test for this, thanks

scrapy/commands/crawl.py
gunblues: I found this shorter code, it will cause `AttributeError: ‘Deferred’ object has no attribute ‘result’` if no exception is raised so I'll change it back

Gallaecio: My bad, you need to add a third parameter with the default value to avoid that: `getattr(res, 'result', None)`

gunblues: no problem, I've modified it, thanks

gunblues:
@Gallaecio @wRAR I've added the test, could you help to review it, thanks

gunblues:
> Great work with the test code, I don’t think there was much existing test code to help figuring out how to implement it.
> 
> However, I do think we can omit some of the new files created as part of a project. See #4218

@Gallaecio thanks for your reply. I think I need to run scrapy/commands/crawl.py for testing but I don't know how to run it by the way you provided #4218

gunblues:
> Well, the tests seem to work, so I think we can go ahead with this implementation.
> 
> But I think there is still some 💄 cleanup we can do before merging:
> 
> * [Sort](https://stackoverflow.com/a/42701457/939364) those imports at ` tests/test_cmdline_crawl_with_pipeline/__init__.py`
> * Remove the unused `setUp` method.
> * Remove comments generated by Scrapy’s project generation tool.
> * Remove the `[deploy]` section from the `scrapy.cfg` file (I don’t think it’s needed here)
> * Remove `BOT_NAME` and `NEWSPIDER_MODULE` from `settings.py` (I think there are not needed either, although I’m less sure about `NEWSPIDER_MODULE`)

@Gallaecio I've modified the code via your suggestion but I have to reserve BOT_NAME in settings to let crawling function work normally, thanks for your review

gunblues:
hello, may I know the approximate time of merging this pr?

gunblues:
@Gallaecio @wRAR may I know when will merge this pull request back?

Gallaecio:
Could you fix the issue reported by Travis CI?

`/home/travis/build/scrapy/scrapy/tests/test_cmdline_crawl_with_pipeline/test_spider/spiders/__init__.py:1:1: W391 blank line at end of file`

gunblues:
> Could you fix the issue reported by Travis CI?
> 
> `/home/travis/build/scrapy/scrapy/tests/test_cmdline_crawl_with_pipeline/test_spider/spiders/__init__.py:1:1: W391 blank line at end of file`

Hi @Gallaecio , thanks for pointing out this issue, I've fixed it
May I know when will merge the PR If Travis CI passes?

Gallaecio:
There is no fixed time. We usually merge once two maintainers have had a chance to review the changes and approve them.

gunblues:
> There is no fixed time. We usually merge once two maintainers have had a chance to review the changes and approve them.

got it, thanks a lot. so I still need one more maintainer to approve it. @wRAR could you help to review?  thanks a lot

gunblues:
@kmike, @wRAR could you help to review my pull request? it needs two maintainers to approve, thank you

PR #4215, opened 2019-12-05
Move staticmethods out of CookiesMiddleware class

grammy-jiang:
This commit moves the static methods out of the CookiesMiddleware class:
* format_cookie
* get_request_cookies

This commit tries to make the code more clear and clean. The changes pass the unit test cases, but the two new functions do not have their own unit test cases.


Gallaecio:
If the goal is not to have an unused `self`, I think the best way would be to add the `@classmethod` decorator to them.

If the methods are tightly related to the class, even if `self` is not used, I think the code is cleaner having them defined as part of the class. 

Moreover, your changes are making them public methods (removing the `_` prefix). Having those methods be private is important, we are telling people not to rely on them, because we will change them or remove them without prior warning (deprecation). Unless there is a need to make them public, they should remain private.

PR #4218, opened 2019-12-06
Add a test for a CrawlerProcess script.

wRAR:


conftest.py
```diff
@@ -17,13 +24,3 @@
 def chdir(tmpdir):
     """Change to pytest-provided temporary directory"""
     tmpdir.chdir()
-
-
-def pytest_collection_modifyitems(session, config, items):
```
Gallaecio: Why remove this?

conftest.py
```diff
@@ -1,11 +1,18 @@
+import glob
+
 import pytest
 
 
+def _py_files(folder):
+    return glob.glob(folder + "/*.py") + glob.glob(folder + "/*/*.py")
```
Gallaecio: https://stackoverflow.com/a/2186565/939364 ?

conftest.py
```diff
@@ -1,11 +1,18 @@
...
+
+
 collect_ignore = [
     # not a test, but looks like a test
     "scrapy/utils/testsite.py",
 ]
 
+collect_ignore += _py_files("tests/CrawlerProcess")
```
Gallaecio: You can probably add this as `*_py_files("tests/CrawlerProcess")` to the list above instead, and maybe include a comment in the line of the existing one, with a brief explanation of why these files must be ignored.

PR #4226, opened 2019-12-10
Fix for Issue # 3345

prajwalchoudhry:
Added AWS SES Functionality for Scrapy.

### DIFF
--- a/scrapy/aws.py
+++ b/scrapy/aws.py
@@ -1,0 +1,101 @@
+import logging
+import smtplib
+from email import encoders as Encoders
+from email.mime.base import MIMEBase
+from email.mime.multipart import MIMEMultipart
+from email.mime.nonmultipart import MIMENonMultipart
+from email.mime.text import MIMEText
+from email.utils import COMMASPACE, formatdate
+from io import BytesIO
+
+from scrapy.utils.misc import arg_to_iter
+
+class SESender(object): 
+
+    def __init__(self, AWS_HOST_NAME, AWS_USER_NAME, AWS_PASSWORD, AWS_PORT_NUMBER, FROM_ADDRESS):
+        self.host_name = AWS_HOST_NAME
+        self.user_name = AWS_USER_NAME
+        self.password = AWS_PASSWORD
+        self.port_number = AWS_PORT_NUMBER
+        self.from_address = FROM_ADDRESS
+
+    def construct_message(self, to, subject, body, cc, attachments, mimetype, charset):
+
+        # Check to see whether attachments exist
+        if attachments: 
+            message = MIMEMultipart()
+        else: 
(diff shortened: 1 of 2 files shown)


scrapy/aws.py
```diff
@@ -0,0 +1,101 @@
...
+from email.mime.base import MIMEBase
+from email.mime.multipart import MIMEMultipart
+from email.mime.nonmultipart import MIMENonMultipart
+from email.mime.text import MIMEText
+from email.utils import COMMASPACE, formatdate
+from io import BytesIO
+
+from scrapy.utils.misc import arg_to_iter
+
+class SESender(object):
```
elacuesta:
```suggestion
class SESSender: 
```

Perhaps an even better name could be found for this class. Also, we're in py3 territory now, no need to inherit from `object`.

scrapy/aws.py
```diff
@@ -0,0 +1,101 @@
...
+from email.mime.nonmultipart import MIMENonMultipart
+from email.mime.text import MIMEText
+from email.utils import COMMASPACE, formatdate
+from io import BytesIO
+
+from scrapy.utils.misc import arg_to_iter
+
+class SESender(object): 
+
+    def __init__(self, AWS_HOST_NAME, AWS_USER_NAME, AWS_PASSWORD, AWS_PORT_NUMBER, FROM_ADDRESS):
```
elacuesta: Any particular reason why these parameters are uppercase?

scrapy/aws.py
```diff
@@ -0,0 +1,101 @@
...
+        self.host_name = AWS_HOST_NAME
+        self.user_name = AWS_USER_NAME
+        self.password = AWS_PASSWORD
+        self.port_number = AWS_PORT_NUMBER
+        self.from_address = FROM_ADDRESS
+
+    def construct_message(self, to, subject, body, cc, attachments, mimetype, charset):
```
elacuesta: Could some of these be assigned default values? `attachments=None` sounds like a reasonable one, for instance.

scrapy/aws.py
```diff
@@ -0,0 +1,101 @@
...
+        # Construct the message data-structure
+        message = MIMEMultipart()
+
+        # Add content to the body
+        message['From'] = self.from_address
+        message['To'] = COMMASPACE.join(to)
+        message['Date'] = formatdate(localtime = True)
+        message['Subject'] = subject
+        message['Body'] = body
+        recipients = to[:]
```
elacuesta: `to.copy()` looks more readable to me, but I guess that's just a matter of personal preference.

elacuesta:
Also, there is #3346 about this same issue.

wRAR:
This is mostly a copy of code from `scrapy.mail` without, it seems, any code specific to Amazon.

PR #4231, opened 2019-12-14
Docs fix

noviluni:
Remove reference to an old (Python 2.7) environment and update the year of the Copytight.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,7 @@
 # serve to show the default.
 
 import sys
+from datetime import datetime
 from os import path
 
 # If your extensions are in another directory, add it here. If the directory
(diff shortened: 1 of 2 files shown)


Gallaecio:
Should we wait 18 days to merge? :)

docs/conf.py
```diff
@@ -50,7 +50,7 @@
 
 # General information about the project.
 project = u'Scrapy'
-copyright = u'2008–2018, Scrapy developers'
+copyright = u'2008–2020, Scrapy developers'
```
Gallaecio: Maybe we could put the current year through code?

noviluni: It is definitely better. Done!

noviluni:
> Should we wait 18 days to merge? :)
It shouldn't be necessary now :stuck_out_tongue_winking_eye:

kmike:
Thanks @noviluni!

Issue #4234, opened 2019-12-15
The line length is not consistent through the whole project


noviluni:
Before adding the flake8 checker, we were allowing any line length, recommending people to follow pep8 (79 characters) but allowing any length “if it improves readability”. 

When adding the flake8 checker we needed to add the E501 rule exception to most of the files (https://github.com/scrapy/scrapy/blob/master/pytest.ini), so technically, **now we are allowing any line length in those files and a line length of 79 characters in the other files**.

As this is bad for the code readability and maintenance (and **self-contradictory**), I think we should decide a clearer position of the line length for the entire project.

### Some possibilities are:
1. Allow **any length** in all the project (if that’s the case we can just ignore the E501 rule for the whole project).
2. Restrict to **79 characters** as PEP8 stands for.
3. Restrict to **another agreed length**.

In the 2nd and 3rd case it could be possible to allow longer lines individually (and not per file) by adding the `# noqa: E501` to the line.

### Looking at the project
To get a big picture of the issue I decided to check the current status of the project and I graphed this (please, obviate the y axis scale), that represents the line length frequency (for those lines with length > 79).

![graph](https://user-images.githubusercontent.com/22377678/70866912-a0555800-1f6f-11ea-90ed-6d5404b65b86.png)

When looking at the graph, it is curious to see that when arriving to the 110 characters it decreases a lot and that there is a “natural” end in about the 140 characters.

You can find here the full file list: https://gist.githubusercontent.com/noviluni/0993ac9d2cbdad12e84b8f5ba2a18d17/raw/d84c298645bd5d6b10545726c2292a417432888c/line_too_long_ordered.txt

##### 1. Allow any length
I think it’s obvious that this (and the current) option is the worst, as it allows some lines to be excessively long. Take this example: https://github.com/scrapy/scrapy/blob/c841a1f3e7c0d44b277dd42300948ba2f6109e67/tests/test_utils_sitemap.py#L26 that is 237 characters long without improving the readability (in fact, it gets worse).

#####  2. Restrict to 79 characters (PEP8)
Some maintainers have declared in the past that this is the prefered option, but that if the code readability improves we should allow longer lines. If we decide to follow this, **we should first decide or at least try to describe what does “improves the readability” mean** and allow only those cases to get a longer length by adding a `# noqa: E501` to those lines, avoiding the current status, where we allow all the file to have any line length.

##### 3. Restrict to another agreed length.
There are **several projects where the maintainers have decided to follow another line length**. 
[...]

PR #4237, opened 2019-12-16
Set flake8-max-line-length to 119

elacuesta:
Related to #4234.

~There are still a lot of files to update, I stopped at the `scrapy/http` directory because even if these changes are not difficult, they are a bit tedious, and I didn't want to continue if this doesn't end up being the accepted decision.~

I must say I like the argument mentioned by @noviluni in #4234, about Django using 119 because it is the width of GitHub's code review :thinking: 

**Update:** Line length set to 119


noviluni:
Hi, @elacuesta ! That seems a hard job, thanks for this "kind of PoC"! While doing this, did you find any of those lines that reducing the length would affect to the readability? just to use it as an example and to discuss about it.

pytest.ini
```diff
@@ -19,74 +19,51 @@ addopts =
...
+    scrapy/core/downloader/webclient.py E731 E502 E128 E126 E226
+    scrapy/core/engine.py E128 E127 E306 E502
+    scrapy/core/scraper.py E306 E128 W504
+    scrapy/core/spidermw.py E731 E502 E126 E226
     # scrapy/downloadermiddlewares
-    scrapy/downloadermiddlewares/ajaxcrawl.py E501 E226
-    scrapy/downloadermiddlewares/decompression.py E501
-    scrapy/downloadermiddlewares/defaultheaders.py E501
-    scrapy/downloadermiddlewares/httpcache.py E501 E126
+    scrapy/downloadermiddlewares/ajaxcrawl.py E226 E501
```
noviluni: is this `E501` in the `scrapy/downloadermiddlewares/ajaxcrawl.py` for some reason?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -199,9 +208,13 @@ class TunnelingAgent(Agent):
...
-    def __init__(self, reactor, proxyConf, contextFactory=None,
-                 connectTimeout=None, bindAddress=None, pool=None):
-        super(TunnelingAgent, self).__init__(reactor, contextFactory, connectTimeout, bindAddress, pool)
+    def __init__(
+        self, reactor, proxyConf, contextFactory=None,
+        connectTimeout=None, bindAddress=None, pool=None
+    ):
+        super(TunnelingAgent, self).__init__(
+            reactor, contextFactory, connectTimeout, bindAddress, pool
+        )
```
nyov: I don't agree that changes like this one make the code more legible. So if a convention hinders rather than helps code comprehension, is it a good convention to obey?

noviluni: Oh, for me it's quite better. Look the next image, in the second version you can read the full code without needing to scroll horizontally.
Screenshot_20200228-180238~2

I would prefer more width (119 or 120), but for me the readibility, at least in the `super` line, has improved.

nyov: Okay, but for me, that's _visibility_ rather than readability. With that argument, I could counter that the change didn't help me any, because I'm shortsighted, have a bigger font, and still need to scroll. Or that my device has less screen-width than yours in that screenshot.

Or that since there are now 4 more lines, I can't see the whole function codeblock in my terminal window, so I need to scroll down/up instead of sideways.

nyov:
> I must say I like the argument mentioned by @noviluni in #4234, about Django using 120 because it is the width of GitHub's code review

I would also favor the 119/120 limit over the 99 characters, for a hard limit, if we must have one.

pytest.ini
elacuesta: No reason, it must have slipped out. Updated, thanks for noticing!

elacuesta:
Updated to 119 after our recent discussion.

scrapy/spiders/crawl.py
```diff
@@ -54,8 +54,12 @@ def _compile(self, spider):
         self.process_request = _get_method(self.process_request, spider)
         self.process_request_argcount = len(get_func_args(self.process_request))
         if self.process_request_argcount == 1:
-            msg = 'Rule.process_request should accept two arguments (request, response), accepting only one is deprecated'
-            warnings.warn(msg, category=ScrapyDeprecationWarning, stacklevel=2)
+            warnings.warn(
+                "Rule.process_request should accept two arguments"
+                " (request, response), accepting only one is deprecated",
```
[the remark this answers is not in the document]
elacuesta: Thanks, I must admit I don't have a strong preference with this particular scenario. "Making it more obvious that the line continues below" does make sense, I think I'm going to stick with that.

kmike:
Hey! I like most changes, and I'm fine with merging 119 as a *hard limit*. 

What are we doing with a *soft limit*, recommended line length?

My personal preference on soft limit: something like 79 (pep8) or 88 (black) sounds better; codebase which is consistently formatted with old-school 79 line limit, with occasional 119 lengths (where it improves readability or makes things easier) is preferrable to a codebase consistently formatted to 119 (for me). 

In the contributing.rst we currently [write](https://docs.scrapy.org/en/master/contributing.html#coding-style):

> Unless otherwise specified, follow PEP 8. It’s OK to use lines longer than 80 chars if it improves the code readability.

This line is not changed in this PR, so is it right that we're still aiming for 79 columns for most code?

tests/test_downloader_handlers.py
```diff
@@ -493,7 +493,10 @@ def test_download_broken_chunked_content_allow_data_loss_via_setting(self):
 class Https11TestCase(Http11TestCase):
     scheme = 'https'
 
-    tls_log_message = 'SSL connection certificate: issuer "/C=IE/O=Scrapy/CN=localhost", subject "/C=IE/O=Scrapy/CN=localhost"'
+    tls_log_message = (
+        'SSL connection certificate: issuer "/C=IE/O=Scrapy/CN=localhost", '
+        'subject "/C=IE/O=Scrapy/CN=localhost"'
```
kmike, line 498: Almost all changes in this PR are for good, but splitting constants like that is why I don't like strict line length restrictions. Here the wrapping makes it easier to make an error (+- a whitespace), and harder to understand what's in the constant. Nothing critical, but I wonder if for such lines we can ignore the warning, instead of changing the code.

tests/test_utils_url.py
```diff
@@ -28,7 +28,10 @@ def test_url_is_from_any_domain(self):
         self.assertTrue(url_is_from_any_domain(url, ['192.169.0.15:8080']))
         self.assertFalse(url_is_from_any_domain(url, ['192.169.0.15']))
 
-        url = 'javascript:%20document.orderform_2581_1190810811.mode.value=%27add%27;%20javascript:%20document.orderform_2581_1190810811.submit%28%29'
+        url = (
+            'javascript:%20document.orderform_2581_1190810811.mode.value=%27add%27;%20'
+            'javascript:%20document.orderform_2581_1190810811.submit%28%29'
+        )
```
kmike, line 34: This is another example where a strict limit arguably makes an example worse. URL is a single line; splitting them in several lines make it harder (at least for me) to understand the code, and adds a surface for bugs (whitespace errors).

tests/test_downloader_handlers.py
elacuesta: Makes sense, yes. I'm gonna go with `# noqa` in the the specific lines, I feel like keeping the files in `pytest.ini` is too much of a carte blanche.

elacuesta:
> What are we doing with a _soft limit_, recommended line length?

Actually, I feel like if we allow exceptions like in https://github.com/scrapy/scrapy/pull/4237#discussion_r418997228 and https://github.com/scrapy/scrapy/pull/4237#discussion_r418997339, we're not imposing a hard limit at all. Not saying it's a bad thing, just that they are different levels of soft limits.

Personally, I like 99 as a recommended length (I feel like 79 is too restrictive and not really necessary with the current displays), but I see this as a matter of personal preference and thus I think it's hard to reach an agreement.

Gallaecio:
I think recommending PEP8 (79) but allowing up to 119 when it improves readability (and further on very special cases) works.

I don’t expect most contributors to read what our limits are, to be honest, but I hope that most will aim for 79 already because that is what PEP8 says, and for fearsome people without limits, the Flake8 tests should make sure they do not go overboard.

kmike:
Thanks everyone!

PR #4242, opened 2019-12-18
Remove requirements-py3.txt

whalebot-helmsman:
Intention is to remove `requirements-py3.txt`. It is direct copy of `install_requires` at this moment. Supporting both takes double effort.


noviluni:
Good job! :tada:

Issue #4041, opened 2019-09-27
Adding Type Hints to Scrapy and its Modules


royahsan:
## Summary

We should add Variable Annotations/  Type hints as supported in [PEP 526](https://www.python.org/dev/peps/pep-0526/) , Python 3.6 to Scrapy to help out existing and new contributors and developers in understanding scrapy code. 

## Motivation

Intellisense  enabled IDES like PyCharm need Type hints to provide better experience.
For new contributors to understand Scrapy comprehensively, type hints are vital. 

Consider someone not that familiar with scrapy,  stumbling upon scheduler's constructor.

```
    def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
                 logunser=False, stats=None, pqclass=None, crawler=None):
        self.df = dupefilter
        self.dqdir = self._dqdir(jobdir)
        self.pqclass = pqclass
        self.dqclass = dqclass
        self.mqclass = mqclass
        self.logunser = logunser
        self.stats = stats
        self.crawler = crawler
```

Gallaecio:
Related to #3618

elacuesta:
Agreed. This sounds great, we should just keep in mind that Scrapy won't drop support for Python 2.7 until [2.0](https://github.com/scrapy/scrapy/milestone/10)

Also (not exactly what this issue is about, but related), _variable_ annotations were introduced in Python 3.6 ([PEP 526](https://www.python.org/dev/peps/pep-0526/))

royahsan:
@elacuesta great!, actually you are right , type annotation were introduced in [PEP 526](https://www.python.org/dev/peps/pep-0526/) , only function annotations were introduced in [PEP 484](https://www.python.org/dev/peps/pep-0484/) , i was in in little hurry finding the PEP version lol.

laurentS:
> Agreed. This sounds great, we should just keep in mind that Scrapy won't drop support for Python 2.7 until [2.0](https://github.com/scrapy/scrapy/milestone/10)

Based on [mypy docs](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages) it seems that it's possible to provide the type hints as separate stub files, which shouldn't interfere with older versions of python that don't support them. Depending on the timeline for scrapy 2.0, this might be a way to fix this issue earlier...

kmike:
[...]

PR #4243, opened 2019-12-18
Add typing hint to httpcache downloadermiddlewares

grammy-jiang:
This commit adds typing hint to httpcache downloadermiddlewares as a
test for #4041

This commit is a trial for adding typing hint to this project


scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -11,41 +13,54 @@
     TimeoutError,
 )
 from twisted.web.client import ResponseFailed
+from typing import Optional, Tuple, Type, Union
```
wRAR: `typing` is a standard module so it should be in the first import block :)

scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -11,41 +13,54 @@
...
 
-    def __init__(self, settings, stats):
+    def __init__(self, settings: Settings, stats: Type[StatsCollector]) -> None:
         if not settings.getbool('HTTPCACHE_ENABLED'):
             raise NotConfigured
-        self.policy = load_object(settings['HTTPCACHE_POLICY'])(settings)
-        self.storage = load_object(settings['HTTPCACHE_STORAGE'])(settings)
-        self.ignore_missing = settings.getbool('HTTPCACHE_IGNORE_MISSING')
+        self.policy: Union[DummyPolicy, RFC2616Policy] = load_object(settings['HTTPCACHE_POLICY'])(settings)
+        self.storage: Union[DbmCacheStorage, FilesystemCacheStorage] = load_object(settings['HTTPCACHE_STORAGE'])(settings)
```
wRAR: These can be any other (user-defined) classes, right?

wRAR:
Tests failed because of "SyntaxError: future feature annotations is not defined" on 3.6 and "SyntaxError: invalid syntax" at `DOWNLOAD_EXCEPTIONS: Tuple = (defer.TimeoutError, TimeoutError, DNSLookupError` on 3.5.

scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -73,7 +88,7 @@ def process_request(self, request, spider):
         # process_response hook
         request.meta['cached_response'] = cachedresponse
 
-    def process_response(self, request, response, spider):
+    def process_response(self, request: Type[Request], response: Type[Response], spider: Type[Spider]) -> Type[Response]:
```
kmike: Type is not needed here, and in other places as well. Type means that a class instance is received, not an object of this type, which is not the case.

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: Yes, and now we have only these classes (`DummpyPolicy`, `RFC2616Policy`, `DbmCacheStorage`, `FilesystemCacheStorage`) in this project.

grammy-jiang:
> Tests failed because of "SyntaxError: future feature annotations is not defined" on 3.6 and "SyntaxError: invalid syntax" at `DOWNLOAD_EXCEPTIONS: Tuple = (defer.TimeoutError, TimeoutError, DNSLookupError` on 3.5.

Yes. For the classmethod `from_crawler` the typing hint of the return is the class itself. This kind of typing hint itself work is only allowed with `__future__.annotations`.

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: Right, I find the right way is to use TypeVar.

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: Agreed. Change is made.

scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -13,39 +16,58 @@
...
+from scrapy.statscollectors import StatsCollector
 from scrapy.utils.misc import load_object
 
+TCrawler = TypeVar("TCrawler", bound=Crawler)
+TException = TypeVar("TException", bound=Exception)
+TRequest = TypeVar("TRequest", bound=Request)
+TResponse = TypeVar("TResponse", bound=Response)
+TSpider = TypeVar("TSpider", bound=Spider)
+TStatsCollector = TypeVar("TStatsCollector", bound=StatsCollector)
+
```
grammy-jiang: Defining the types used in every current module is quite boring. There should be a better way to do this.

elacuesta: I'm by no means a typing expert, but can't you just use the classes themselves? Like you are doing with the `settings` argument in `__init__`.

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: TypeVar is removed. Now the native class is adopted for typing hint.

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: The comment is adopted. Now the native class is used for typing hint.

grammy-jiang:
> If I'm not mistaken, you should be able to work around the classmethod issue by supplying the _name_ of the class, i.e.: `def from_crawler(cls, crawler: Crawler) -> "HttpCacheMiddleware"`
> 
> Also, I don't think we are (yet) in a position to support variable annotations, we still support Python 3.5 ([#4041 (comment)](https://github.com/scrapy/scrapy/issues/4041#issuecomment-536039359))

Yes, in Python 3.5 the variable annotation is not supported.

Adding typing hints is not a simple work as I expected before. From this PR we have some valuable discussion and we can see there are still some certain misunderstand for the typing module in Python 3.5+. Practicing more will be much helpful.

grammy-jiang:
Since some of the members of this project have approved this commit, I think maybe we can write a guidelines about how to add the typing hint for Scrapy?

Gallaecio:
I guess we can extend the contributing page with such guidelines. What do you have in mind?

elacuesta:
Sounds good. We should also think about adding a typing check step to the tox/travis configurations.

scrapy/downloadermiddlewares/httpcache.py
```diff
@@ -33,19 +40,19 @@ def __init__(self, settings, stats):
         self.stats = stats
 
     @classmethod
-    def from_crawler(cls, crawler):
+    def from_crawler(cls, crawler: Crawler) -> "HttpCacheMiddleware":
```
kmike: this method doesn't return HttpCacheMiddleware in a subclass, see https://stackoverflow.com/questions/44640479/mypy-annotation-for-classmethod-returning-instance for a discussion

kmike:
+1 to add automatic type checks on CI; I think this should be done even before merging any type annotations.

grammy-jiang:
> I guess we can extend the contributing page with such guidelines. What do you have in mind?

From the discussion in this pull request, I can see the following points should be  included in this document:

- The place to add typing hint at the first stage:
  To be compatible with python 3.5, only add typing hint for input and output, not for variables; for `classmethod` should we add typing hint for `return`, and in which way, a string like `HttpcacheMiddleware`? or leave it until in python 3.7 with the power of `__future__.annotations` to annotate itself?
- The class of typing hint to mention:
  Should the typing hint include the subclass (we already have two comments about this topic in this pull request and I can imagine there will be more discussion in the community in the future)? Should the typing hint mention all possible classes (with typing.Union)?
- Some special cases, e.g. the typing hint for a decorator is a nightmare because what kind of function will be decorated in the user's development is very hard to predict.

There are many things uncertain about typing hint. But at least we can start from some easy and clear places now, and keep discussing at the same time until we find a good solution.

> Part of code is typing hinted is better than nothing. - Alan Turing

scrapy/downloadermiddlewares/httpcache.py
grammy-jiang: Em, I can see this will be a hot topic in this typing hint topic.

First, should all kinds of possible typing hint be covered? Using `typing.Type` to cover all subclasses is not a problem. The problem is should we use `typing.Type` for all classes in Scrapy, just like the previous `Request` and `Response` typing hint in this pull request?

Second, annotating the class itself is supported until python 3.7 with the help of `__future__.annotations`. Since Scrapy is still compatible with python 3.5, what should we do for `classmethod` return at this moment? A string typing hint is not a good solution in my opinion.

grammy-jiang:
> +1 to add automatic type checks on CI; I think this should be done even before merging any type annotations.

Yes, this will be much helpful for the contributors to test their code before creating the pull request.

> Fixing the bug before merging is much easier than after. - John von Neumann

elacuesta:
[Typing build](https://travis-ci.org/github/scrapy/scrapy/jobs/670670051) added @kmike 🚀

mypy.ini
```diff
@@ -0,0 +1,63 @@
+[mypy]
```
kmike, line 1: what do you think about using setup.cfg, to not introduce a new config file?

elacuesta: That works for me :ok_hand:

elacuesta: Updated in 9eaaa1ea1dc2481577dca7dce15da139c79d0699

tox.ini
```diff
@@ -22,6 +22,13 @@ passenv =
     GCS_PROJECT_ID
 commands =
     py.test --cov=scrapy --cov-report= {posargs:--durations=10 docs scrapy tests}
+    
+[testenv:typing]
+basepython = python3
+deps =
+    mypy==0.770
+commands =
+    mypy --config-file=setup.cfg {posargs:scrapy}
```
kmike, line 31: What do you think about running it on tests as well?

kmike: I was also thinking mypy is going to pick up setup.cfg automatically

elacuesta: Done in f28da1b0d648a9ef518302061d48fea48404c464. Initially I was under the impression that `setup.cfg` would be picked up, the docs say so. However, for some reason it wasn't working for me, so I decided to go the explicit way. Now it does, so perhaps I was doing something wrong :shrug:

scrapy/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 # Scrapy version
 import pkgutil
-__version__ = pkgutil.get_data(__package__, 'VERSION').decode('ascii').strip()
+__version__ = (pkgutil.get_data(__package__, 'VERSION') or b'').decode('ascii').strip()
```
kmike, line 10: why is this change needed? Is it related to typing?

elacuesta: Indeed:
```
scrapy/__init__.py:10: error: Item "None" of "Optional[bytes]" has no attribute "decode"
```

I just reverted it and added the file to the ignore list

elacuesta:
Tests are failing because [`typing.Type`](https://docs.python.org/3/library/typing.html#typing.Type) was introduced in Python 3.5.2, and we're testing with 3.5.1 since #4518. According to the [release schedule page](https://www.python.org/dev/peps/pep-0478/#release-schedule), 3.5.2 was released on June 26th 2016, six months after 3.5.1. It should be safe to bump the requirement IMHO.

Gallaecio:
Thank you!

PR #4249, opened 2019-12-19
Use Python 3.7 to build the documentation

Gallaecio:
I broke the build changing it to 3.8 in #4140, it turns out Read The Docs does not support 3.8 yet.


kmike:
👍

PR #732, opened 2014-05-24
change Scraper API to call internal `_parse` method (fix #781)

nyov:
As a fix to the CrawlSpider-can't-have-parse useability issue (#712 / now #781) ,
I propose having another internal `_parse` method called by the `Scraper` instead of `parse`, and using that in spiders who want internal "pre-processing", then exposing `parse` as the public, documented, no-baggage method to implement/override from a spider.

`Spider` simply calls the public `parse` method from `_parse`, while `CrawlSpider` and cohorts can do their thing in `_parse` and  give the cold shoulder to users trying to break them by re-defining `parse` without calling `super()`.

Since all spiders presently should subclass from `Spider`, this also doesnt't break any existing spider classes who keep using `parse` exclusively.

PR #4254, opened 2019-12-23
Change Scraper API to call internal `_parse` method

elacuesta:
This is #732, cherry-picked on top of the current `master` branch and updated to add keyword argument support on callbacks.

Related to #781.


elacuesta:
Removed the note about the `CrawlSpider.parse` method, added a note about explicitly setting the callback for new requests in CrawlSpider/XMLFeedSpider. This isn't a new requirement, just something that is not currently mentioned in the docs: the current implementation send requests without an explicit callback to the spider's `parse` method, which is supposed not to be overridden; the user doesn't necessarily know exactly what it does.

elacuesta:
@scrapy/maintainers any thoughts about this?

elacuesta:
Thanks @Gallaecio and @wRAR for the reviews, and many thanks @nyov for the initial implementation :smile:

PR #4257, opened 2019-12-27
Filtering examples in Scrappy logging documentation

### LINKED ISSUE #4216: Cover arbitrary filtering in the Scrapy logging documentation
It should be clear, from reading the documentation, how to filter out a specific log message that we wish to ignore.

This is specially important for warnings that depend on input, like the one introduced in #4214. Since you seldom have the power to fix the issue that triggers the warning message, caused by the content or behavior of the website you are scraping, you may need to simply ignore those warning messages.

Exposing a setting or a LogFormatter method for each of those warnings does not seem scalable to me, specially when such warnings can come from third-party Scrapy extensions.

akamanzi:
Added tutorial example to filter substring in log message using logging filters. 
Example given doesn't cover the use of regular expressions, there is an external link to redirect users to Python's standard library `re` module.

Fixes #4216

### DIFF
--- a/docs/topics/logging.rst
+++ b/docs/topics/logging.rst
@@ -242,6 +242,23 @@ e.g. in the spider's ``__init__`` method::
 If you run this spider again then INFO messages from
 ``scrapy.spidermiddlewares.httperror`` logger will be gone.
 
+You could also filter logs based on a substring. This can be achieved using `Filter` class. For example, if you want to filter log message that contains a string 'filter', you can do it like this::
+
+    import logging
+
+    class ContentFilter(logging.Filter):
+        def filter(self, record):
+            return 'filter' in record.msg
+
+    logger = logging.getLogger('mycustomlogger')
+    logger.addFilter(ContentFilter())
+    logger.info("This is a filter based on a substring")
+
+adding the code above will return log messages that contain substring 'filter'.
+
+Additionally, you can also filter message using regular expression, for this you can check more on `regex <https://docs.python.org/3/library/re.html>`_
+
+
 scrapy.utils.log module
 =======================
 


docs/topics/logging.rst
```diff
@@ -242,6 +242,23 @@ e.g. in the spider's ``__init__`` method::
...
 
+You could also filter logs based on a substring. This can be achieved using `Filter` class. For example, if you want to filter log message that contains a string 'filter', you can do it like this::
+
+    import logging
+
+    class ContentFilter(logging.Filter):
+        def filter(self, record):
+            return 'filter' in record.msg
+
+    logger = logging.getLogger('mycustomlogger')
```
Gallaecio: I think we should make it clear how to grab the root logger that Scrapy installs.

akamanzi: @Gallaecio, thank you for the feedback, I will look into this next week!

Thank you

akamanzi: @Gallaecio, sorry for this late reply, i had exams at the moment. so by root logger do you mean calling `logging.getLogger()` instead of `logging.getLogger('mycustomlogger')` as in the above example? could you please elaborate on this?

Gallaecio: I’m not sure if `logging.getLogger()` would work, or if you would need to access something else that is not a logger. The point is that we should document an approach that can be used for:
- The whole project. If a matching message is logger by a middleware, a spider, or something else, it should be filtered out. I’m not sure if the same steps using `getLogger()` would work, or something else is needed. Try.
- A specific logger (i.e. from a specific middleware or from a specific spider). If this is as simple as using `getLogger` with the right parameter, I think the documentation just needs to show an example of logging and indicate what is the logger name there (for Scrapy users not familiar yet with the basics of Python logging).

akamanzi: @Gallaecio , i now get your point and it is valid. am not sure how either to get the root logger for the whole project. trying to look into it.

PR #4258, opened 2019-12-27
Edited the link provided to homepage of lxml's website

atul-g:
The link "https://lxml.de" is redirecting to a completely different and unintended website. I changed the link to the index page of lxml's official website. I thought of changing it to the PyPi page of lxml, but even they are providing the same "https://lxml.de" link which doesn't seem to be working now.

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -278,7 +278,7 @@ For details, see `Issue #2473 <https://github.com/scrapy/scrapy/issues/2473>`_.
 
 .. _Python: https://www.python.org/
 .. _pip: https://pip.pypa.io/en/latest/installing/
-.. _lxml: http://lxml.de/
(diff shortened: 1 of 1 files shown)


elacuesta:
For reference: https://bugs.launchpad.net/lxml/+bug/1820328

atul-g:
Cool, thanks for the reference.

atul-g:
Cool, thanks for the reference.

elacuesta:
I updated the target branch to `master`. The build is failing because it's based on an old branch that doesn't include the #4095 fix, I don't think we need to worry about it.

atul-g:
Thank you for that! I'm quite new to this process of PR and still learning here. Had no idea why the build was failing when all I changed was a link.

kmike:
Thanks @atul-g and @elacuesta!

PR #4259, opened 2019-12-30
Asyncio support in downloader middlewares

wRAR:


scrapy/utils/defer.py
```diff
@@ -114,3 +118,25 @@ def iter_errback(iterable, errback, *a, **kw):
             break
         except Exception:
             errback(failure.Failure(), *a, **kw)
+
+
+def _isfuture(o):
+    # workaround for Python before 3.5.3 not having asyncio.isfuture
+    if hasattr(asyncio, 'isfuture'):
+        return asyncio.isfuture(o)
+    return isinstance(o, asyncio.Future)
```
Gallaecio: If the coverage data is correct, `_isfuture` could be replaced by `asyncio.isfuture` and work the same. I wonder what the best approach here would be. Maybe we should just use `asyncio.isfuture`.

If we need to exclude future-like content that are not `Future` instances, such as `Task` instances, I think we may need a more complex check.

wRAR: We don't test on Python before 3.5.3 :)

tests/test_downloadermiddleware.py
```diff
@@ -206,3 +209,46 @@ def process_request(self, request, spider):
...
+        self.assertIs(results[0], resp)
+        self.assertFalse(download_func.called)
+
+    @mark.only_asyncio()
+    def test_asyncdef_asyncio(self):
+        resp = Response('http://example.com/index.html')
+
+        class CoroMiddleware:
+            async def process_request(self, request, spider):
+                await asyncio.sleep(0.1)
```
kmike: Could you please also add a test where some result is awaited?

wRAR: Such as?
Looks like asyncio without 3rd-party modules supports async networking, queues and subprocesses, not sure if anything else. Or you can create a Task and manage it.
We would also want similar code in all other asyncio testcases?

kmike: I'm not sure which library to use and what action to perform exactly, but we need to test that results work - getting something from DB or making a HTTP request can be a standard use case for async def methods. If it is not easy to set something up using stdlib, that's fine to have additional requirements (e.g. aiohttp). That said, it should be possible to implement something using subprocess or queue.

scrapy/utils/defer.py
Gallaecio: :man_facepalming: 

:thinking: I think we should either change that or not bother supporting 3.5.3-.

wRAR: Yup.
My only concern is that the current Python3 version in Ubuntu 16.04 LTS is 3.5.2. Though I couldn't find how to make travis actually use the system Python3.

wRAR: Let's see if it works fine on travis 3.5.2.

tests/test_downloadermiddleware.py
wRAR: I've added a Queue with sync put and async get, it's not very interesting though, let me know if we should use something heavier.

scrapy/utils/defer.py
Gallaecio: The proxy_connect tests will fail. In https://github.com/scrapy/scrapy/pull/4261 I’ve installed 3.5.0, and it works, but I’ve had to skip mitmproxy-based tests, as for Python 3.5.3 and earlier you need to use mitmproxy 2, and even after updating the command-line parameters accordingly I got timeouts.

wRAR: Yet https://travis-ci.org/scrapy/scrapy/jobs/631286391 worked with 3.5.2

Gallaecio: Hmm… It worked…

.travis.yml
```diff
@@ -18,6 +18,8 @@ matrix:
       python: 3.5
     - env: TOXENV=py35-asyncio
       python: 3.5
+    - env: TOXENV=py35-asyncio
+      python: 3.5.2
```
kmike: What do you think about removing python35-asyncio env? They're almost the same; here we're testing lowest and highest supported Python.

wRAR: Yeah

tests/test_downloadermiddleware.py
kmike: Looks good!

PR #4267, opened 2020-01-04
change link for brotli package to a supported one

KristobalJunta:
I have noticed that the offcial documentation suggests using [brotlipy](https://pypi.org/project/brotlipy/). While this works, the project seems to be stale or abandoned. Most recent release was in 2017 and most recent commit - in 2018.

I suggest replacing that with a more recent and seemingly more actively supported [Brotli](https://pypi.org/project/Brotli) by google. It's most recent release was in 2018 and most recent commit - this month. This library has same API and judging by my experience, has proved to work seamlessly.

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -704,13 +704,14 @@ HttpCompressionMiddleware
    sent/received from web sites.
 
    This middleware also supports decoding `brotli-compressed`_ as well as
-   `zstd-compressed`_ responses, provided that `brotlipy`_ or `zstandard`_ is
+   `zstd-compressed`_ responses, provided that `brotli`_ or `zstandard`_ is
    installed, respectively.
 
 .. _brotli-compressed: https://www.ietf.org/rfc/rfc7932.txt
-.. _brotlipy: https://pypi.org/project/brotlipy/
+.. _brotli: https://pypi.org/project/Brotli/
 .. _zstd-compressed: https://www.ietf.org/rfc/rfc8478.txt
 .. _zstandard: https://pypi.org/project/zstandard/
+
 
 HttpCompressionMiddleware Settings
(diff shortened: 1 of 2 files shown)


kmike:
No strong opinion from me, but brotlipy doesn't look abandoned - there are recent discussions in pull requests. Google's brotli Python wrapper is also not changing much: if you check https://github.com/google/brotli/commits/master/python, a 2019 change is a comment added, and in 2018 there are 2 changes, one of which is test-only - this is not that different from brotlipy. 

On the other hand, Google's version has a more recent brotli bundled.

If we suggest Google's library, we should change tests to run against it.

KristobalJunta:
@kmike yes, that's quite reasonable. Thank for the feedback!
I'll try and provide the tests for Google's implementation.

KristobalJunta:
@kmike I have updated requirements for tests and run them. Everything seems to pass.

kmike:
@pawelmhm any opinions on brotli package, are you ok with changing it (as you implemented brotli support originally)?

pawelmhm:
I'm fine with that change @kmike let's merge this! Sorry for delay I completely missed this notification. It would be good to clean up brotli situation now, I think it should be ok to install it by default now, without making it optional dependency.

Gallaecio:
@KristobalJunta Do you think you will have time to resolve the current conflicts?

KristobalJunta:
@Gallaecio let me look into that and come back with update

Gallaecio:
Thank you! I think the failing jobs are unrelated to your changes. I’ll look into addressing https://github.com/scrapy/scrapy/issues/5425 and re-run the CI jobs here afterwards.

KristobalJunta:
@Gallaecio thank you! Looked into jobs myself and it seems as you said indeed.

Gallaecio:
Thanks!

PR #4269, opened 2020-01-09
Support for async def callbacks.

wRAR:
This is not very useful yet, as it doesn't allow `yield` in callbacks, but it's a first step.

### DIFF
--- a/tests/spiders.py
+++ b/tests/spiders.py
@@ -106,6 +106,17 @@ class AsyncDefAsyncioSpider(SimpleSpider):
         self.logger.info("Got response %d" % status)
 
 
+class AsyncDefAsyncioReturnSpider(SimpleSpider):
+
+    name = 'asyncdef_asyncio_return'
+
+    async def parse(self, response):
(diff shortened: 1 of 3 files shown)


kmike:
Does it allow to return a list of Requests or items? It yes, it'd be nice to test it.

tests/spiders.py
```diff
@@ -83,6 +87,36 @@ def parse(self, response):
...
+
+class AsyncDefAsyncioReturnSpider(SimpleSpider):
+
+    name = 'asyncdef_asyncio_return'
+
+    async def parse(self, response):
+        await asyncio.sleep(0.2)
+        status = await get_from_asyncio_queue(response.status)
+        self.logger.info("Got response %d" % status)
+        return [{'id': 1}, {'id': 2}]
```
kmike: Nice, so lists of items are supported. What about requests?

kmike:
I'm merging it; test for behavior with Requests can be added separately. Thanks @wRAR!

PR #4270, opened 2020-01-09
async def support in pipelines

wRAR:


scrapy/utils/defer.py
```diff
@@ -2,6 +2,7 @@
 Helper functions for dealing with Twisted deferreds
 """
 import asyncio
+from functools import wraps
```
Gallaecio: :lipstick: Below `import inspect`?

kmike:
Thanks @wRAR!

scrapy/utils/defer.py
kmike: I like @Gallaecio's style proposal a bit more than what's in the code. Merging PR anyways, this is not a blocker.

PR #4271, opened 2020-01-09
async def support for signal handlers that already supported Deferreds

wRAR:


scrapy/utils/defer.py
```diff
@@ -140,3 +140,20 @@ def deferred_from_coro(o):
             # wrapping the coroutine into a Future and then into a Deferred, this requires AsyncioSelectorReactor
             return defer.Deferred.fromFuture(asyncio.ensure_future(o))
     return o
+
+
+def maybeDeferred_coro(f, *args, **kw):
+    """ Copy of defer.maybeDeferred that also converts coroutines to Deferreds. """
+    try:
+        result = f(*args, **kw)
+    except:  # noqa: E722
```
Gallaecio: Shouldn’t this at least be `except Exception`, to avoid things like catching a `KeyboardInterrupt`?

wRAR: It's a copy of `defer.maybeDeferred` ¯\\\_(ツ)\_/¯

Gallaecio: I guess `defer.fail` will capture whatever exception to be properly processed where it must be processed.

tests/test_utils_signal.py
```diff
@@ -63,8 +67,24 @@ def ok_handler(self, arg, handlers_called):
...
+
+
+@mark.only_asyncio()
+class SendCatchLogDeferredAsyncioTest(SendCatchLogDeferredTest):
+
+    async def ok_handler(self, arg, handlers_called):
+        handlers_called.add(self.ok_handler)
+        assert arg == 'test'
+        await asyncio.sleep(0.2)
+        return await get_from_asyncio_queue("OK")
```
kmike: Could you please add integration-style test for this feature, maybe not to this module? Take some real signal which Scrapy is sending, implement async def callback for it, check that it works.

wRAR: I looked at the async-enabled signals and how are they used. There seem to be two groups.

`item_scraped`, `item_dropped`, `item_error`: these are related to `Scraper._itemproc_finished` but as far as I can see the result of that function is ignored and nothing waits for those signal handlers to finish, but I may be wrong.

`engine_started`, `engine_stopped`, `spider_opened`, `spider_closed`: these are included in the spider/engine startup/shutdown processes and the process waits for the handler to finish.

kmike: That's actually a great information to document.

tests/test_signals.py
```diff
@@ -0,0 +1,39 @@
...
+    def setUp(self):
+        self.mockserver = MockServer()
+        self.mockserver.__enter__()
+        self.items = []
+
+    def tearDown(self):
+        self.mockserver.__exit__(None, None, None)
+
+    async def _on_item_scraped(self, item):
+        self.items.append(item)
```
kmike: How hard would it be to have some await here?

kmike:
Thanks @wRAR!

PR #4272, opened 2020-01-10
Spider middleware: catch spider callback exceptions early

### LINKED ISSUE #4260: First Spider Middleware does not process exception for generator callback
### Description

`process_spider_exception` method of a spider middleware is ignored when spider middleware is first and callback is a generator.

### Steps to Reproduce

1. Create spider middleware with defined `process_spider_exception`
2. Put it into `SPIDER_MIDDLEWARES` with number more than `900` (to make it first)
3. Raise an exception in a spider callback
4. UPD. Yield an item from the callback

**Expected behavior:** `process_spider_exception` is called for this exception

**Actual behavior:** `process_spider_exception` is not called

### Versions

```
Scrapy       : 1.7.1
lxml         : 4.2.5.0
libxml2      : 2.9.8
cssselect    : 1.0.3
parsel       : 1.5.1
w3lib        : 1.19.0
Twisted      : 18.9.0
Python       : 3.6.8 (default, May  8 2019, 05:35:00) - [GCC 6.3.0 20170516]
pyOpenSSL    : 18.0.0 (OpenSSL 1.1.0j  20 Nov 2018)
cryptography : 2.4.2
Platform     : Linux-4.9.184-linuxkit-x86_64-with-debian-9.9
```

elacuesta:
Fixes #4260.

Evaluates the output iterable right after the spider callback, as it's currently being done in the `process_spider_output` chain.

(Plus some minor styling adjustments)

### DIFF
--- a/scrapy/core/spidermw.py
+++ b/scrapy/core/spidermw.py
@@ -117,14 +117,12 @@ class SpiderMiddlewareManager(MiddlewareManager):
             return MutableChain(result, recovered)
 
         def process_callback_output(result):
-            if isinstance(result, Failure):
-                return process_spider_exception(result)
             recovered = MutableChain()
             result = _evaluate_iterable(result, 0, recovered)
             return MutableChain(process_spider_output(result), recovered)
 
         dfd = mustbe_deferred(process_spider_input, response)
-        dfd.addCallbacks(callback=process_callback_output, errback=process_callback_output)
+        dfd.addCallbacks(callback=process_callback_output, errback=process_spider_exception)
         return dfd
 
     def process_start_requests(self, start_requests, spider):
--- a/scrapy/core/spidermw.py
+++ b/scrapy/core/spidermw.py
@@ -3,18 +3,26 @@ Spider Middleware manager
 
 See documentation in docs/topics/spider-middleware.rst
 """
-from itertools import chain, islice
+from itertools import islice
 
 from twisted.python.failure import Failure
+
 from scrapy.exceptions import _InvalidOutput
 from scrapy.middleware import MiddlewareManager
-from scrapy.utils.defer import mustbe_deferred
 from scrapy.utils.conf import build_component_list
+from scrapy.utils.defer import mustbe_deferred
 from scrapy.utils.python import MutableChain
 
 
 def _isiterable(possible_iterator):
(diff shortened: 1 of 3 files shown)


elacuesta:
Don't know what the Codecov comment is about, there are no changes to `scrapy/robotstxt.py` here 🤷‍♂

scrapy/core/spidermw.py
```diff
@@ -69,47 +86,43 @@ def process_spider_exception(_failure, start_index=0):
...
+                    # might fail directly if the output value is not a generator
                     result = method(response=response, result=result, spider=spider)
                 except Exception as ex:
                     exception_result = process_spider_exception(Failure(ex), method_index+1)
                     if isinstance(exception_result, Failure):
                         raise
                     return exception_result
-                if _isiterable(result):
-                    result = evaluate_iterable(result, method_index)
                 else:
```
kmike: 💄 is else needed here?

elacuesta: Not really, because of the above `return exception_result`. I just thought it would make the separation clearer.

elacuesta: Removed in ad4477d335bee8b10bc3bbca969defddd9b316f8.
The `else` is still there and the comment does not appear as "outdated" because it's now part of the `if _isiterable(result)` block, not the `try...except` one.

scrapy/core/spidermw.py
```diff
@@ -69,47 +86,42 @@ def process_spider_exception(_failure, start_index=0):
...
+                    # might fail directly if the output value is not a generator
                     result = method(response=response, result=result, spider=spider)
                 except Exception as ex:
                     exception_result = process_spider_exception(Failure(ex), method_index+1)
                     if isinstance(exception_result, Failure):
                         raise
                     return exception_result
                 if _isiterable(result):
-                    result = evaluate_iterable(result, method_index)
+                    result = _evaluate_iterable(result, method_index+1, recovered)
```
kmike: I wonder if there is a better name possible for method_index variable, because here method_index+1 is passed. Something like process_exception_index?

elacuesta: Renamed to `exception_processor_index`, thanks!

kmike:
Thanks @elacuesta!

PR #4275, opened 2020-01-15
Fixed artwork/README formatting

abhishekh2001:


artwork/README.rst
```diff
@@ -1,5 +1,4 @@
-:orphan:
```
kmike: I think it was intentional to have this marker.

Gallaecio: It [looks](https://github.com/scrapy/scrapy/pull/2507#issue-102643392) like it was added to avoid Sphinx picking it up, but it also [looks](https://travis-ci.org/scrapy/scrapy/jobs/637250957) like it is not being picked up after the change, so I’m OK with the change.

kmike:
@abhishekh2001 could you please clarify: issue title says "Fixed artwork/README formatting"; where does the result look broken?

Gallaecio:
@kmike I bet GitHub: [master](https://github.com/scrapy/scrapy/blob/master/artwork/README.rst) → [this change](https://github.com/scrapy/scrapy/blob/03241aa4a66f9b0e8be4dc104807011f546840e8/artwork/README.rst)

kmike:
Ok, let's merge it. Though many of the documents are going to be broken if looked from github, and it'll get worse once we put more API docs to docstrings.

Issue #4276, opened 2020-01-17
Fields not registered in Items.fields if set as static variable


scratchmex:
### Description

The static variables set in `scrapy.Items` class are not registered in the internal `_values` dict and therefore I cannot access them via `self['field_name']`, i.e., the `__getattr__` function.

### Steps to Reproduce
In my items_error.py file I have this. (the actual code is larger but I isolate the code causing the error)
```
from scrapy import Item, Field

class BaseItem(Item):
    url=Field()
    id=Field()
    type=Field()

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if not self['type']: #<--- error here. Raises KeyError
            raise NotImplementedError('type of ad must be set as static class variable')
        #use url hash as id to avoid duplicates
        self['id']=hash(self.get('url'))

class ChildItem(BaseItem):
    type='estate'
    other_fields=Field()
```

When I ran it this raises:
```
C:\Users\ivangonzalez\Documents\Code\scrady>python
Python 3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 21:48:41) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from scrady.items_error import *
>>> ChildItem()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\ivangonzalez\Documents\Code\scrady\scrady\items_error.py", line 10, in __init__
    if not self['type']: #<--- error here. Raises KeyError
  File "C:\Users\ivangonzalez\Anaconda3\lib\site-packages\scrapy\item.py", line 91, in __getitem__
    return self._values[key]
KeyError: 'type'
>>> BaseItem()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\ivangonzalez\Documents\Code\scrady\scrady\items_error.py", line 10, in __init__
    if not self['type']: #<--- error here. Raises KeyError
  File "C:\Users\ivangonzalez\.virtualenvs\scrady-d4yjsu4b\lib\site-packages\scrapy\item.py", line 91, in __getitem__       return self._values[key]
KeyError: 'type'
>>>
```

**Expected behavior:** `self['field_name']` should return the `Field()` attribute.

**Actual behavior:** static fields are not registered by `scrapy.Items`  in `scrapy.Items._values[key]`.

**Reproduces how often:** Ran in pipenv, so 100%.

### Versions

[...]

PR #4278, opened 2020-01-17
Fixed: Fields not registered in Items.fields if set as static variable

scratchmex:
This solves #4276. All tests passed in my computer. Any questions or modifications let me know. Thanks.

### DIFF
--- a/scrapy/item.py
+++ b/scrapy/item.py
@@ -46,11 +46,16 @@ class ItemMeta(ABCMeta):
         new_bases = tuple(base._class for base in bases if hasattr(base, '_class'))
         _class = super(ItemMeta, mcs).__new__(mcs, 'x_' + class_name, new_bases, attrs)
 
-        fields = getattr(_class, 'fields', {})
+        fields = {}
+        for base in bases:
+            fields.update(getattr(base, 'fields', {}))
+        fields.update(getattr(_class, 'fields', {}))
         new_attrs = {}
         for n in dir(_class):
             v = getattr(_class, n)
             if isinstance(v, Field):
+                fields[n] = v
+            elif n in fields:
                 fields[n] = v
             elif n in attrs:
                 new_attrs[n] = attrs[n]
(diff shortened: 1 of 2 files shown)


scrapy/item.py
```diff
@@ -75,7 +81,10 @@ def __new__(cls, *args, **kwargs):
 
     def __init__(self, *args, **kwargs):
         self._values = {}
-        if args or kwargs:  # avoid creating dict for most common case
+        for k, v in self.fields.items():
```
elacuesta: I don't think this is correct. It could end up assigning the _field itself_ as value:
```python
   ...:
   ...: class CustomItem(Item):
   ...:     name = Field(serializer=str)
   ...:
   ...: i = CustomItem()

In [2]: type(i["name"])
Out[2]: scrapy.item.Field

In [3]: i["name"]
Out[3]: {'serializer': str}
```

tests/test_item.py
```diff
@@ -171,6 +174,34 @@ class E(C, B): pass
...
+        class ChildItem(TestItem):
+            name = 'John'
+            age = Field()
+
+        item = ChildItem()
+        self.assertEqual(item.get('name'), 'John')
+        self.assertRaises(KeyError, item.__getitem__, 'age')
+
+        class GrandChildItem(ChildItem):
+            age = 20
```
elacuesta: I think there might be a misunderstanding here. Of course items can have arbitrary attributes, properties and methods just like any other class, but for an attribute to be treated like a _field_ it should be declared as a `scrapy.item.Field` object (https://docs.scrapy.org/en/1.8/topics/items.html#declaring-items). Otherwise, unexpected behaviour can occur (for instance, some components like [item exporters](https://github.com/scrapy/scrapy/blob/1.8/scrapy/exporters.py#L64) check the item's `fields` attribute).

scratchmex:
What about migrating to use Pydantic library (https://pydantic-docs.helpmanual.io/) for data validation?

Gallaecio:
Data validation is usually performed at a different level. https://spidermon.readthedocs.io/ is a common approach nowadays.

PR #4279, opened 2020-01-19
Updating flags in Response.follow and TextResponse.follow

### LINKED ISSUE #4277: Response.follow() method not consistent with Request.__init__(). Missing `flags`.
### Versions Affected:

All versions after PR #2082 when flags were added to `Request`. (>1.4.0)

### Response.follow() method not consistent with Request.__init__(). Missing `flags`.

No sure how this got missed since PR #2082 was merged, but looking at the parameters for `Request.__init__` and `Response.follow` it appears `flags` wasn't added to `Response.follow` to keep it in line with how creating new `Request` instances works. I'm not sure if this was just overlooked or was intentional, when `flags` support was added to requests. It's also missing for the subclasses of `Response` as well (`TextResponse` will need updated, the rest currently use inheritance).

I was looking to be able to add custom `flags` to certain `Response.follow()` calls but realized looking at the source they wouldn't be carried over to the new responses, but this can easily be worked around by manually added the flag after Request creation. This seems like an easy fix to include in 1.8.1 or later but doesn't seem too high priority and I just thought you guys should know or that it should at least be noted somewhere on this repo since the [docs](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response.follow) still say, "It accepts the same arguments as `Request.__init__` method...".

Prime-5:
In reference to #4277, added "flags" parameter to ```Response.follow``` and ```TextResponse.follow```

(edit) Fixes #4277


elacuesta:
Hello, and many thanks for the contribution. I have to requests:
1. Could you add the `flags` parameter to the end? I know it would make `Request.__init__` and `Response.follow` have their parameters sorted differently, but it's possible that some users are already passing them as positional argument and this would break backward compatibility.
2. Could you add a test case for this change? I'm aware that the current tests only check the URLs for the generated requests, but I believe it shouldn't be too hard to do.

I think the first comment should be addressed before approval. I would be ok with leaving the second one to a separate PR, since it's not a matter of just adding one line or test function to an existing test case.

elacuesta:
Thanks for the update. One additional comment I forgot to make the first time: flags also need to be passed when [creating the Request](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/http/response/__init__.py#L126-L136) and when [invoking the method from the parent class](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/http/response/text.py#L148-L159).

Prime-5:
Yes I was about to ask the same, on it

Prime-5:
I'll add the test case in a separate PR. 
For adding the test case, I will have to add a function that tests the flags in ```test_http_response```, right?

elacuesta:
Thanks. Indeed, `test_http_response.py` would be the appropriate place for it IMHO.

kmike:
@Prime-5 we usually do tests together with changes/features; if you intend to send a PR with tests anyways, would you mind adding commits to this PR instead?

Prime-5:
Hello @kmike, I've added a test case but I'm not sure why there's a conflict.

Prime-5:
The updated __init__.py which has a new function ```follow_all``` calls the ```follow``` function without the flags parameter. Could this be the reason behind the conflict in the first file?

elacuesta:
Yes, that is almost certainly what's happening, it should be solved by adding the flags to the `follow_all` methods as well. When I started #4057 we weren't aware of this issue.

Prime-5:
I've ran into a little trouble. I'm sorry for asking this here because it's not related to scrapy but to the workflow of github. 
I updated my fork by committing the changes in the original codebase to my ```master``` branch. I can view all the new changes only in my ```master```, and my commits are only there in the ```update_flags``` branch. I tried merging them in my fork but that's not working either.

elacuesta:
My personal setup involves having two git remotes, `origin` pointing to my fork and `upstream` pointing to the source repo. When I need to get commits from the upstream repo I usually do `git fetch upstream` followed by `git merge upstream/master`.
Looking at the commit history from your fork, I'd recommend you reset your history, something like:
```
git remote add upstream git@github.com:scrapy/scrapy.git
git fetch upstream
git checkout master
git reset --hard upstream/master
```

Then you can merge the changes from `master` into your feature branch:
```
git checkout update_flags
git merge master
```
At this point, you will have some conflicts, which you can solve by following [this guide](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-using-the-command-line#competing-line-change-merge-conflicts).

Hope this helps, let us know if you still have problems. Cheers!

Prime-5:
@elacuesta thanks a lot for the patient reply. I've added the flags parameter to the ```follow_on()``` function as well.

Prime-5:
The error is in the test function. For calling the ```Response.follow()``` should I be using ```self.response_class.follow()```?

elacuesta:
`follow` and `follow_all` are instance methods, they will fail if you call them directly from the class.
You need to create a response object (you can use `self.response_class` for this) object and then call `follow/follow_all` on it.

Prime-5:
Should I add tests for testing the ```flags``` parameter in ```Response.follow_all()``` and ```TextResponse.follow_all()```?

Prime-5:
Hello, sorry for the delay, I was caught up in my university exams. I have now added the parameter in ```TextResponse.follow_all()```

elacuesta:
Thanks for the update. Since we're at it, could you add a test to make sure `follow_all` methods work as expected?

Prime-5:
I've added tests to check ```flags``` in ```follow_all()``` over a list of three urls. Do we need to add more tests?

Prime-5:
@elacuesta should there be a feature to allow returning different flag values over succeeding urls when called from ```follow_all()```?
I can't point out any specific scenario right away where it might be useful but I thought it might be worth suggesting.

elacuesta:
I don't think that is necessary, all other parameters (cookies, headers, etc) are passed to all generated requests.

PR #4283, opened 2020-01-23
Updated RFPDupeFilter line separator for correct universal newlines mode usage.

LanetheGreat:
Hey guys, found another small non-breaking glitch that I'm not sure if anybody else noticed that existed since RFPDupfilter was implement back on commit 549725215e79fbd3c7b3590c2a47fc9c0ad30a1b that's specific to Windows and the RFPDupeFilter used in Scrapy. It looks like when RFPDupeFilter writes its request hashes to the `requests.seen` file for a job, it seems to be using [Python's universal newlines](https://docs.python.org/3.8/library/functions.html#open-newline-parameter) feature incorrectly and ends up writing '\r\r\n' as each line ending instead of '\r\n', though it works fine in Linux and writes '\n' as expected.

Here's an example file demonstrating what's happening, opened in Notepad++ and a Hex editor with one of the line endings highlighted:
Newlines

I've tested this PR (using a subclass in my actual project overriding the base behavior) and it works fine on both Linux and Windows writing '\n' and '\r\n' accordingly. This bug's not that serious but would be a nice data storage optimization for those running Scrapy on Windows using job storage, instead of writing thousands of extra carriage returns (\r) to the file.

### DIFF
--- a/tests/test_dupefilters.py
+++ b/tests/test_dupefilters.py
@@ -2,6 +2,7 @@ import hashlib
 import tempfile
 import unittest
 import shutil
+import os
 from testfixtures import LogCapture
 
 from scrapy.dupefilters import RFPDupeFilter
@@ -128,6 +129,26 @@ class RFPDupeFilterTest(unittest.TestCase):
         assert case_insensitive_dupefilter.request_seen(r2)
 
         case_insensitive_dupefilter.close('finished')
+
+    def test_seenreq_newlines(self):
+        """ Checks against adding duplicate \r to
+        line endings on Windows platforms. """
+
+        r1 = Request('http://scrapytest.org/1')
+
+        path = tempfile.mkdtemp()
+        try:
+            df = RFPDupeFilter(path)
+            df.open()
+            df.request_seen(r1)
+            df.close('finished')
+
+            with open(os.path.join(path, 'requests.seen'), 'rb') as seen_file:
+                line = next(seen_file).decode()
+                assert not line.endswith('\r\r\n')
+
+        finally:
+            shutil.rmtree(path)
 
     def test_log(self):
         with LogCapture() as l:
--- a/tests/test_dupefilters.py
+++ b/tests/test_dupefilters.py
@@ -85,17 +85,21 @@ class RFPDupeFilterTest(unittest.TestCase):
         path = tempfile.mkdtemp()
         try:
             df = RFPDupeFilter(path)
-            df.open()
-            assert not df.request_seen(r1)
-            assert df.request_seen(r1)
-            df.close('finished')
+            try:
+                df.open()
+                assert not df.request_seen(r1)
+                assert df.request_seen(r1)
+            finally:
+                df.close('finished')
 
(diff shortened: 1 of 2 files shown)


elacuesta:
Would it be possible to add some tests for this?

LanetheGreat:
Just updated my pull request to include a new test for the incorrect line endings. Also while writing this I noticed a slight bug with the test `test_dupefilter_path` when I accidentally triggered it testing my test. It appears to leave `requests.seen` open and locked when it asserts (because `df/df2`'s `.close()` method doesn't get called) causing `shutil.rmtree` to fail with PermissionError. I added a commit that fixes that in my PR but I'm not sure if that should be intended behavior or not because by **not** deleting it that allowed my to still see the temporary `requests.seen` file and realize my error sooner...

tests/test_dupefilters.py
```diff
@@ -129,6 +134,26 @@ def request_fingerprint(self, request):
...
+        path = tempfile.mkdtemp()
+        try:
+            df = RFPDupeFilter(path)
+            df.open()
+            df.request_seen(r1)
+            df.close('finished')
+
+            with open(os.path.join(path, 'requests.seen'), 'rb') as seen_file:
+                line = next(seen_file).decode()
+                assert not line.endswith('\r\r\n')
```
elacuesta: Could you also check if the line ends with `\n` here?

elacuesta:
I believe the test changes are ok.
Regarding the underlying newline change, I'm not 100% sure because I'm not a Windows user, but it does make sense to me, specially considering the following quote from https://docs.python.org/3/library/os.html#os.linesep (emphasis mine):

> os.linesep
The string used to separate (or, rather, terminate) lines on the current platform. This may be a single character, such as '\n' for POSIX, or multiple characters, for example, '\r\n' for Windows. **Do not use os.linesep as a line terminator when writing files opened in text mode (the default); use a single '\n' instead, on all platforms.**

tests/test_dupefilters.py
LanetheGreat: Just updated the PR, how's that? It should check that universal line endings mode is selected and working as expected now on each platform (Linux/Mac and Windows).

kmike:
@LanetheGreat could you please check if updated RFPDupeFilter can load a file created by a previous version of RFPDupeFilter? It could be fine if it doesn't load, but in this case we would need to mark this change as backwards incompatible.

LanetheGreat:
> @LanetheGreat could you please check if updated RFPDupeFilter can load a file created by a previous version of RFPDupeFilter? It could be fine if it doesn't load, but in this case we would need to mark this change as backwards incompatible.

Just created a virtual environment (pipenv with Python 3.6.8/Windows 10) and installed my PR version of scrapy into it and tested loading the same old `requests.seen` file I used in my first comment above and it looks like it loaded all the hashes fine with all line endings properly stripped. So looks like it doesn't conflict trying to read old files. (Cut it down to the first 20 to not overfill the console screen here)

image

PR #4286, opened 2020-01-25
[Docs] Fix variable name `author_page_links`

petervandenabeele:
I did not test this code, but the change from `href` to this
`author_page_links` seems to have a typo ?


kmike:
Thanks @petervandenabeele!

PR #4288, opened 2020-01-26
[Docs] 2 typos + 1 clarification in docs

petervandenabeele:
Fixing 2 small typos and adding 1 word as clarification in the downloader-middlewares.

Also, I was confused with the entries like `ref:Reppy <reppy-parser>` and similar entries.
Are these supposed to be links to other parts of the doc, or is this the intended way of showing these references ?


elacuesta:
I'm no Sphinx expert, but if I understand correctly `:ref:` directives create links to other sections.
```
:ref:`section`
```
sets the link text using the default name defined in the section itself, while
```
:ref:`Custom name <section>`
```
allows to override that text.

kmike:
Thanks @petervandenabeele!

PR #4290, opened 2020-01-26
FilesPipeline.file_path has optional arguments

dekimsey:
Documented signature doesn't match the actual interface in [files.py](https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/files.py#L520).

Specifically, it looks like it may be [called](https://github.com/scrapy/scrapy/blob/master/scrapy/pipelines/files.py#L422) without a response value.

I found this when I was implementing the pipeline with the signature `file_path(self, request, response, info)` and the following error was being return in my results :

    [(False, <twisted.python.failure.Failure builtins.TypeError: file_path() missing 1 required positional argument: 'response'>)]

Scrapy==1.8.0

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -410,7 +410,7 @@ See here the methods that you can override in your custom Files Pipeline:
 
 .. class:: FilesPipeline
 
-   .. method:: file_path(request, response, info)
+   .. method:: file_path(self, request, response=None, info=None)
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
Could you update the `ImagesPipeline` one as well?

dekimsey:
Will do!

> Could you update the `ImagesPipeline` one as well?

Will do

kmike:
Thanks @dekimsey, nice catch!

PR #4291, opened 2020-01-27
add zsh -h autocomplete option

seregaxvm:
This PR adds autocompletion for `-h` option for zsh. 
This option was not added to #4069 due to the problem described in [this so question](https://stackoverflow.com/q/58312199/5745120).


kmike:
I'm not a zsh user myself, but hope this works :P

PR #4293, opened 2020-01-29
[HttpCompressionMiddleware] fix delimiter for Accept-Encoding header

edorofeev:
Accept-Encoding header should have `, `  (with space) as a delimiter between encodings instead of `,` (non space) (https://tools.ietf.org/html/rfc2616#section-14.3)


dangra:
+1 LGTM.

PR #4300, opened 2020-02-04
Deprecate overriding settings with SCRAPY-prefixed environment variables

drewseibert:
Related to issue: https://github.com/scrapy/scrapy/issues/3871

### DIFF
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -68,7 +68,6 @@ def get_project_settings():
     if settings_module_path:
         settings.setmodule(settings_module_path, priority='project')
 
-    # XXX: remove this hack
     pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
     if pickled_settings:
         warnings.warn("Use of environment variable "
@@ -76,10 +75,9 @@ def get_project_settings():
                       "is deprecated.", ScrapyDeprecationWarning)
         settings.setdict(pickle.loads(pickled_settings), priority='project')
 
-    # XXX: deprecate and remove this functionality
     env_overrides = {k[7:]: v for k, v in os.environ.items() if
                      k.startswith('SCRAPY_')}
     if env_overrides:
+        warnings.warn("Use of 'SCRAPY_'-prefixed environment variables to override settings is deprecated.", ScrapyDeprecationWarning)
         settings.setdict(env_overrides, priority='project')
-
     return settings


elacuesta:
Hey @drewseibert, thanks for the PR.

I _think_ it's ok to remove the first block, but the second one (overriding setting using env variables) needs to be deprecated first. Tests are failing now because this it's missing.

drewseibert:
@elacuesta Hello! Thanks for the feedback! What exactly do you mean when you say we have to deprecated the second block? I noticed the Travis CI build failed. 

Also, when I run pytest against my changes I am seeing this error below?..

../../conftest.py:17: in <module>
    for line in open('tests/ignores.txt'):
E   FileNotFoundError: [Errno 2] No such file or directory: 'tests/ignores.txt'

drewseibert:
@elacuesta I see that the flake8 test is failing with the error in the screenshot below..
Screen Shot 2020-02-04 at 3 37 23 PM

elacuesta:
Indeed, the screenshot you posted shows the final summary of the errors. If you look [up in the build log](https://travis-ci.org/scrapy/scrapy/jobs/646150022#L572-L578) you can see a detailed list of problems.
Regarding the deprecation of the second block, what I mean is that we cannot just remove that without a deprecation period, i.e. adding the same check and warning that the `SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE` section currently has. After that, we could remove it in a future version.

drewseibert:
@elacuesta Thank you for the helpful feedback. I have made the appropriate changes to get the Travis CI build to pass. I added the warning in, similar to the one you referred to in a previous PR. What do you think about how it looks now? Thanks :)

scrapy/utils/project.py
```diff
@@ -68,18 +66,10 @@ def get_project_settings():
...
-    # XXX: remove this hack
-    pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
-    if pickled_settings:
-        warnings.warn("Use of environment variable "
-                      "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' "
-                      "is deprecated.", ScrapyDeprecationWarning)
-        settings.setdict(pickle.loads(pickled_settings), priority='project')
-
-    # XXX: deprecate and remove this functionality
+    # XXX: deprecating the block below
```
elacuesta: I don't think this comment is necessary

scrapy/utils/project.py
```diff
@@ -68,18 +66,10 @@ def get_project_settings():
...
     env_overrides = {k[7:]: v for k, v in os.environ.items() if
                      k.startswith('SCRAPY_')}
     if env_overrides:
+        warnings.warn("Use of environment variable 'SCRAPY_' is deprecated.", ScrapyDeprecationWarning)
```
elacuesta: I'd rephrase this a bit to clearly indicate the purpose of the variables. Something like `Use of 'SCRAPY_'-prefixed environment variables to override settings is deprecated`

elacuesta: Would `usage` be better here? I'm not a native speaker, I read a few comments about use/usage but I'm still not sure.

drewseibert:
@elacuesta Thanks for the feedback! I restored the spaces you mentioned for readability. I also modified the warning message. I am not sure if usage is really any better for the second block. Also, should we put the first block back in then? If so, we can deprecate it much like the second block looks. Thanks!

elacuesta:
Indeed, I'd appreciate if you could restore the first block, the prevailing opinion among maintainers is to have a longer deprecation period.

drewseibert:
@elacuesta Hi there! Sorry I have been away from this PR for a few days. I put back the first deprecated code block. However the Travis build is failing. Here is the link to the last build attempt:
https://travis-ci.org/scrapy/scrapy/builds/648526984?utm_source=github_status&utm_medium=notification

scrapy/utils/project.py
```diff
@@ -68,18 +68,16 @@ def get_project_settings():
...
         settings.setmodule(settings_module_path, priority='project')
 
-    # XXX: remove this hack
     pickled_settings = os.environ.get("SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE")
     if pickled_settings:
         warnings.warn("Use of environment variable "
                       "'SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE' "
                       "is deprecated.", ScrapyDeprecationWarning)
-        settings.setdict(pickle.loads(pickled_settings), priority='project')
+    settings.setdict(pickle.loads(pickled_settings), priority='project')
```
Gallaecio: I’m guessing this indentation change was accidental :)

elacuesta: Indeed, that's what making the build fail.

elacuesta:
```suggestion
        settings.setdict(pickle.loads(pickled_settings), priority='project')
```

drewseibert:
@elacuesta @Gallaecio Thanks for the eyes on that! How does it look now? Okay for merge? Thanks!

drewseibert:
@elacuesta Sounds good, I have squashed all commits. 👍

nyov:
I disagree with removing the `# XXX: remove this hack` lines without removing the hacks.
I fail to see how this actually cleans up the code. :confused:

elacuesta:
That's a good point @nyov. To be honest, I didn't pay too much attention to those removals. For instance, personally I would have kept the empty line at https://github.com/scrapy/scrapy/pull/4300/files#diff-d5d764f0ce0d7be9d2e1208d2f12a932L84, but I felt like that would have been too bike-sheddy of me.

I think the `ScrapyDeprecationWarning` occurrences are enough reminders to remove stuff, once a few versions have passed. Of course we could probably improve the process, maybe agreeing on a deprecation period and/or keeping a list of deprecations somewhere to prevent them from living too long.

Gallaecio:
> Hm, okay, I guess you meant to turn the hack into a proper code deprecation.
> Isn't that kind of validating the hack? Oh well, let's have it for another few years then.

Rather than validating the hack, I believe the point is to deprecate it properly, giving whoever depends on it an the moment proper warning before removing the feature they are using.

But yes, that may mean keeping it for another few years :sweat_smile:. Well, we are actually yet to decide on a deprecation policy, so it could be any amount of time: 6 months, 1 year, 2 years, 5 years…

azenkwed:
> @elacuesta Hello! Thanks for the feedback! What exactly do you mean when you say we have to deprecated the second block? I noticed the Travis CI build failed.
> 
> Also, when I run pytest against my changes I am seeing this error below?..
> 
> ../../conftest.py:17: in 
> for line in open('tests/ignores.txt'):
> E FileNotFoundError: [Errno 2] No such file or directory: 'tests/ignores.txt'

How did you solve this? I am getting the same error.

Gallaecio:
> > Also, when I run pytest against my changes I am seeing this error below?..
> > ../../conftest.py:17: in
> > for line in open('tests/ignores.txt'):
> > E FileNotFoundError: [Errno 2] No such file or directory: 'tests/ignores.txt'
> 
> How did you solve this? I am getting the same error.

Sounds like `pytest` is not being executed from the folder containing the `tests` folder. If `ls` (or `dir` in Windows) does not list a `tests` folder, you are running the tests from the wrong folder.

PR #4303, opened 2020-02-05
request_left_downloader signal

whalebot-helmsman:
As requested in https://github.com/scrapy/scrapy/pull/3520#issuecomment-580100847 (cc @elacuesta )


docs/topics/signals.rst
```diff
@@ -295,6 +295,23 @@ request_reached_downloader
     :param spider: the spider that yielded the request
     :type spider: :class:`~scrapy.spiders.Spider` object
 
+request_left_downloader
+---------------------------
+
+.. signal:: request_left_downloader
+.. function:: request_left_downloader(request, spider)
+
+    Sent when a :class:`~scrapy.http.Request` left downloader even in case of
```
elacuesta:
```suggestion
    Sent when a :class:`~scrapy.http.Request` leaves the downloader even in case of
```

docs/topics/signals.rst
```diff
@@ -295,6 +295,23 @@ request_reached_downloader
...
+    failure.
+
+    The signal does not support returning deferreds from their handlers.
```
elacuesta:
```suggestion
    This signal does not support returning deferreds from its handlers.
```
I'm hopping to get the other occurrences of "their" updated in #4205

tests/test_request_left.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class SignalCatcherSpider(Spider):
+    name = 'signal_catcher'
+
+    def __init__(self, crawler, url, *args, **kwargs):
+        super(SignalCatcherSpider, self).__init__(*args, **kwargs)
+        crawler.signals.connect(self.on_response_download,
+                                signal=request_left_downloader)
+        self.catched_times = 0
```
elacuesta:
```suggestion
        self.caught_times = 0
```
The subsequent occurrences should be updated too.

elacuesta:
Thanks @whalebot-helmsman! Code looks great, just I left a few stylistic comments.

tests/test_request_left.py
```diff
@@ -0,0 +1,60 @@
...
+        self.start_urls = [url]
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = cls(crawler, *args, **kwargs)
+        return spider
+
+    def on_response_download(self, request, spider):
```
elacuesta: Should this be called differently, if it will called even if the request fails and the response is not actually downloaded?

docs/topics/signals.rst
whalebot-helmsman: It is done in accord with description of other signals, e.g https://github.com/scrapy/scrapy/blame/master/docs/topics/signals.rst#L257

tests/test_request_left.py
whalebot-helmsman: Renamed in https://github.com/scrapy/scrapy/pull/4303/commits/4be19e443e9c101a248c21509ae8000ce500d51a

docs/topics/signals.rst
elacuesta: I'm no native English speaker, but I believe "this" sounds better than "the". That's probably subjective though.
However I'm pretty sure "their" is incorrect, since we're talking about one signal, not many.

docs/topics/signals.rst
```diff
@@ -295,6 +295,23 @@ request_reached_downloader
     :param spider: the spider that yielded the request
     :type spider: :class:`~scrapy.spiders.Spider` object
 
+request_left_downloader
+---------------------------
```
Gallaecio:
```suggestion
-----------------------
```

I’m surprised the documentation build does not fail :slightly_smiling_face:

docs/topics/signals.rst
```diff
@@ -295,6 +295,23 @@ request_reached_downloader
...
+
+.. signal:: request_left_downloader
+.. function:: request_left_downloader(request, spider)
+
+    Sent when a :class:`~scrapy.http.Request` leaves the downloader even in case of
```
Gallaecio: :lipstick: 

```suggestion
    Sent when a :class:`~scrapy.http.Request` leaves the downloader, even in case of
```

docs/topics/signals.rst
```diff
@@ -295,6 +295,23 @@ request_reached_downloader
...
+    failure.
+
+    The signal does not support returning deferreds from their handlers.
+
+    :param request: the request that reached downloader
```
Gallaecio:
```suggestion
    :param request: the request that reached the downloader
```

kmike:
Thanks @whalebot-helmsman!

PR #4101, opened 2019-10-24
Removed envirovment documentation about python 2

MrKovalski:
Removed envirovment documentation about python 2, because it is no longer suported. As requested in issue #4001 . If that is what you meant. If not sorry, first time contributing.

PR #4309, opened 2020-02-06
Update installation instructions regarding Python 3 and virtual environments

### LINKED ISSUE #4001: Rewrite the virtual environment documentation section for Python-3-only Scrapy
I think https://docs.scrapy.org/en/latest/intro/install.html#using-a-virtual-environment-recommended can be made simpler once Scrapy only supports Python 3.

Gallaecio:
Simplify by not repeating information that can be found in upstream, linked documentation.

Fixes #4001, closes #4101

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -81,34 +81,17 @@ Python packages can be installed either globally (a.k.a system wide),
 or in user-space. We do not recommend installing Scrapy system wide.
 
 Instead, we recommend that you install Scrapy within a so-called
-"virtual environment" (`virtualenv`_).
-Virtualenvs allow you to not conflict with already-installed Python
+"virtual environment" (:mod:`venv`).
+Virtual environments allow you to not conflict with already-installed Python
 system packages (which could break some of your system tools and scripts),
 and still install packages normally with ``pip`` (without ``sudo`` and the likes).
 
-To get started with virtual environments, see `virtualenv installation instructions`_.
-To install it globally (having it globally installed actually helps here),
-it should be a matter of running::
+See :ref:`tut-venv` on how to create your virtual environment.
 
-    $ [sudo] pip install virtualenv
-
-Check this `user guide`_ on how to create your virtualenv.
-
-.. note::
-    If you use Linux or OS X, `virtualenvwrapper`_ is a handy tool to create virtualenvs.
-
-Once you have created a virtualenv, you can install Scrapy inside it with ``pip``,
+Once you have created a virtual environment, you can install Scrapy inside it with ``pip``,
 just like any other Python package.
 (See :ref:`platform-specific guides <intro-install-platform-notes>`
 below for non-Python dependencies that you may need to install beforehand).
-
-Python virtualenvs can be created to use Python 2 by default, or Python 3 by default. As Scrapy
-only supports Python 3, make sure you created a Python 3 virtualenv.
-
-.. _virtualenv: https://virtualenv.pypa.io
-.. _virtualenv installation instructions: https://virtualenv.pypa.io/en/stable/installation/
-.. _virtualenvwrapper: https://virtualenvwrapper.readthedocs.io/en/latest/install.html
-.. _user guide: https://virtualenv.pypa.io/en/stable/userguide/
 
 
 .. _intro-install-platform-notes:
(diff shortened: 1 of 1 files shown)


docs/intro/install.rst
```diff
@@ -1,245 +1,56 @@
...
-you can install Scrapy and its dependencies from PyPI with::
+You can install Scrapy from the `Python Package Index`_::
 
     pip install Scrapy
 
-Note that sometimes this may require solving compilation issues for some Scrapy
-dependencies depending on your operating system, so be sure to check the
-:ref:`intro-install-platform-notes`.
-
-We strongly recommend that you install Scrapy in :ref:`a dedicated virtualenv <intro-using-virtualenv>`,
```
kmike: I think that'd be nice to preserve it, as it is not said like this in https://packaging.python.org/tutorials/installing-packages/.

docs/intro/install.rst
```diff
@@ -1,245 +1,56 @@
...
-albeit with potential issues with TLS connections.
-
-**Don't** use the ``python-scrapy`` package provided by Ubuntu, they are
-typically too old and slow to catch up with latest Scrapy.
-
-
-To install Scrapy on Ubuntu (or Ubuntu-based) systems, you need to install
-these dependencies::
-
-    sudo apt-get install python3 python3-dev python3-pip libxml2-dev libxslt1-dev zlib1g-dev libffi-dev libssl-dev
```
kmike: In past I found it helpful to have a list of packages needed to get `pip install scrapy` working. It can be too much to try to find this information in install docs of lxml, cryptography, searching on stackoverflow why packages which require a compiler don't work (python3-dev), etc. This list can become outdated, but in practice it doesn't happen that often.

docs/intro/install.rst
```diff
@@ -1,245 +1,56 @@
...
-
-.. note::
-    The same non-Python dependencies can be used to install Scrapy in Debian
-    Jessie (8.0) and above.
-
-
-.. _intro-install-macos:
-
-Mac OS X
---------
```
kmike: The problem with removing this section is that it is not covered by any links here. If I recall correctly, they were added to remedy a constant stream of tickets related to installation - for many people making Scrapy project is the first time they're using Python.

+1 to simplify it, maybe mention `xcode-select --install` and ask to use virtualenv.

docs/intro/install.rst
```diff
@@ -1,245 +1,56 @@
...
-that is to ``brew install openssl`` and then export the flags that this command
-recommends (only needed when installing Scrapy). Installing on Linux has no special
-issues besides installing build dependencies.
-Installing Scrapy with PyPy on Windows is not tested.
-
-You can check that Scrapy is installed correctly by running ``scrapy bench``.
-If this command gives errors such as
-``TypeError: ... got 2 unexpected keyword arguments``, this means
-that setuptools was unable to pick up one PyPy-specific dependency.
-To fix this issue, run ``pip install 'PyPyDispatcher>=2.1.0'``.
```
kmike: Unfortunately this kind of problems is not covered anywhere as well. Is there a better place for such notes?

docs/intro/install.rst
```diff
@@ -1,245 +1,56 @@
...
-We strongly recommend that you install Scrapy in :ref:`a dedicated virtualenv <intro-using-virtualenv>`,
-to avoid conflicting with your system packages.
-
-For more detailed and platform specifics instructions, as well as
-troubleshooting information, read on.
-
-
-Things that are good to know
-----------------------------
+For more information, see :ref:`packaging:installing-packages`.
```
kmike: I wonder if there is a more beginner-friendly document available. This one is good because it is official, but it can be scary. Consider how it starts:

> This section covers the basics of how to install Python packages.

And then, the next paragraph:

> It’s important to note that the term “package” in this context is being used as a synonym for a distribution (i.e. a bundle of software to be installed), not to refer to the kind of package that you import in your Python source code (i.e. a container of modules). It is common in the Python community to refer to a distribution using the term “package”. Using the term “distribution” is often not preferred, because it can easily be confused with a Linux distribution, or another larger software distribution like Python itself.

Not exactly the information beginners are looking for.

Gallaecio:
I better rewrite the patch focusing on the virtual environment changes only.

docs/intro/install.rst
```diff
@@ -81,35 +81,18 @@ Python packages can be installed either globally (a.k.a system wide),
...
-Virtualenvs allow you to not conflict with already-installed Python
+"virtual environment" (:mod:`venv`).
+Virtual environments allow you to not conflict with already-installed Python
 system packages (which could break some of your system tools and scripts),
 and still install packages normally with ``pip`` (without ``sudo`` and the likes).
 
-To get started with virtual environments, see `virtualenv installation instructions`_.
-To install it globally (having it globally installed actually helps here),
-it should be a matter of running::
+See :ref:`tut-venv` on how to create your virtual environment.
```
Gallaecio: That reference points to https://docs.python.org/3/tutorial/venv.html

kmike:
Thanks @Gallaecio!

PR #4310, opened 2020-02-06
Cover version directive usage in the documentation policy

### LINKED ISSUE #3971: Establish a policy on the use of Sphinx directives for deprecations and version changes
I think we should make the following changes to make a better use of Sphinx’s `deprecated`, `versionadded`, and `versionchanged` directives:

* Changes should include these tags when appropriate.
  We need to update the contribution documentation accordingly, and remember to enforce this during code reviews.
* These tags should be removed in major version changes.
  Because of this, Scrapy 2.0 is probably the best time to implement this policy.

Gallaecio:
Fixes #3971


docs/topics/spider-middleware.rst
```diff
@@ -357,8 +353,6 @@ Whether to enable referer middleware.
 REFERRER_POLICY
 ^^^^^^^^^^^^^^^
 
-.. versionadded:: 1.4
```
kmike: What's the policy of keeping old versionadded directives?

Gallaecio: I was going with major versions, but it should probably be the same as for deprecation removals.

docs/contributing.rst
```diff
@@ -194,6 +194,17 @@ In any case, if something is covered in a docstring, use the
 documentation instead of duplicating the docstring in files within the
 ``docs/`` directory.
 
+Documentation updates that cover new or modified features must use Sphinx’s
+:rst:dir:`versionadded` and :rst:dir:`versionchanged` directives. Use
+``VERSION`` as version, we will replace it with the actual version right before
+the corresponding release. When we release a new major or minor version of
+Scrapy, we remove these directives if they are older than a year.
```
kmike, line 201: I think we can be less aggressive here - what's the harm in keeping versionadded directives for longer?

Gallaecio: Work for me. Do you have a figure in mind? 2 years? 3 years?

Conceptually, I think we should remove them when they refer to a Scrapy version that is too old to be relevant. The problem is defining when a Scrapy version is “too old”.

kmike: What do you think about saying `we may remove these directives if they are older than 2 years`?  (or older than a year - note "may", which make this not an obligation, but a possibility).

Gallaecio: I would rather chose a high value (e.g. 5 years), but have a clear guideline, so that, upon writing release notes for a new version, it is clear which directives to remove.

While for deprecation removals flexibility is very welcome, allowing to support common deprecated features for longer and removing old-enough deprecated features as soon as they get in the way of other code changes, I cannot think of a scenario where we would want to remove directives about the same version at different points in time.

kmike: 2-3 years is fine with me then.

Gallaecio: Applied.

Note that 3 years means removing directives about versions older than 1.5.0 at the moment, and the fresher one that had been removed so far was from 1.4.0, so I did not need to restore any of the directives removed in this patch.

kmike:
Thanks @Gallaecio!

PR #4311, opened 2020-02-06
Make METAREFRESH_IGNORE_TAGS an empty list by default

### LINKED ISSUE #3844: Consider making METAREFRESH_IGNORE_TAGS an empty list by default
As a way to allow users to fix #1422, #3768 introduced the `METAREFRESH_IGNORE_TAGS` setting.

To keep backward compatibility, the setting was introduced with `['script', 'noscript']` as default value. However, to reproduce the behavior of web browsers, it seems the right value would be `[]`.

Should we switch the default value of the `METAREFRESH_IGNORE_TAGS` setting to `[]`, even though the change breaks backward compatibility?

Gallaecio:
Fixes #3844


kmike:
Thanks @Gallaecio!

PR #4312, opened 2020-02-06
Implement the ITEM_PIPELINE_CLOSE_SPIDER_ORDER setting

### LINKED ISSUE #2645: Order of calling close_spider in pipelines
`process_item` is called based on the order of the pipeline classes mentioned in [`ITEM_PIPELINES`](https://doc.scrapy.org/en/latest/topics/settings.html#std:setting-ITEM_PIPELINES) setting. But, `close_spider` follows the exact opposite order with the `close_spider` of the last pipeline getting called first [[link](https://github.com/scrapy/scrapy/blob/129421c7e31b89b9b0f9c5f7d8ae59e47df36091/scrapy/middleware.py#L64)].

I can't think of a good reason for this. Should I submit a PR?

Gallaecio:
Fixes #2645


Gallaecio:
Should we make it `desc` by default, for backward-compatibility?

tests/test_pipelines.py
```diff
@@ -99,3 +134,63 @@ def test_asyncdef_asyncio_pipeline(self):
...
+        self.assertEqual(crawler.call_list, expected_call_list)
+
+    @defer.inlineCallbacks
+    def test_pipeline_method_order_invalid_value(self):
+        settings = {'ITEM_PIPELINE_CLOSE_SPIDER_ORDER': 'invalid'}
+        crawler = self._create_crawler(Pipeline1, Pipeline2, **settings)
+        with self.assertRaises(ValueError):
+            crawler.crawl(mockserver=self.mockserver)
+        # TODO: Fix “twisted.trial.util.DirtyReactorAggregateError: Reactor was
+        # unclean.”
```
Gallaecio: @kmike Do you know how to avoid this issue?

Issue #3276, opened 2018-05-28
[question] Why adding `dont_filter=True` for requests made from `start_urls`


starrify:
It's observed that currently (as of b364d27) in `scrapy.Spider.start_requests` the generated requests have `dont_filter=True`. (related line of code: [link](https://github.com/scrapy/scrapy/blob/b364d27247b2d9b86c164569c7e0459fa3f8391b/scrapy/spiders/__init__.py#L83))

As I've had a quick look through the history, this behavior of adding `dont_filter=True` to initial requests seems to have been introduced in the very initial commit of 83dcf8a (related line of code: [link](https://github.com/scrapy/scrapy/blob/83dcf8aff91d0178ffbf4978fb757fc7e6d5d08c/scrapy/trunk/scrapy/core/manager.py#L122)).

At least from my personal perspective of view, there seems to be no reasonable purpose for this specific behavior.

Could anyone help explaining the design and resolving my confusion? Or, if there's no such reason (or the previous reasons do not hold meaningful now), what about removing the `dont_filter=True` attribute for initial requests?

kmike:
I'm not sure about original motivation, but removing it would be backwards incompatible, as this behavior is [documented](https://doc.scrapy.org/en/latest/topics/spiders.html?#scrapy.spiders.Spider.start_requests). There can be non-obvious consequences: for example, it would expose some Scrapy issues which are currently hidden because of `dont_filter=True` - see https://github.com/scrapy/scrapy/issues/1225 and https://github.com/scrapy/scrapy/issues/1304#issuecomment-112181293.

I'm not strictly against fixing it if it causes issues, if there is enough motivation for a change. What's the reason you'd like to fix it - is it just a cleanup, or some issue you faced?

xiaowangwindow:
If you want set `dont_filter=False` for start_urls request, why not to overwrite `start_requests` method?

Gallaecio:
Should we close this, or discuss a behavior change for Scrapy 2.0?

kmike:
I think that's a good candidate for Scrapy 2.0.

Gallaecio:
In a scenario where `start_urls` is `["https://example.com/1", "https://example.com/2"]` and `https://example.com/1` redirects to `https://example.com/2`, I guess the current behavior causes `https://example.com/2` to be scraped twice.

Assuming my assumption is correct, I am in favor of changing the behavior.

MartiONE:
[...]

PR #4314, opened 2020-02-06
Do not filter duplicate requests within a redirect chain

### LINKED ISSUE #1225: dupefilter skips a request when a page is redirected to itself
Hit this when trying to run a spider against scrapinghub.com: sometimes it responds with `302 moved permanently to scrapinghub.com`. Scheduler agrees and tries to schedule another request for scrapinghub.com, but fails because dupefilter already considers it visited.

Maybe dupefilter should only add hosts when the response is not a redirect? And when it is, the scheduler should probably remember the original address too, so that all the redirection chain can be marked as visited.

Gallaecio:
Fixes #1225, closes #5793, closes #7754.

The approach is based on @immerrr’s https://github.com/scrapy/scrapy/issues/1225#issuecomment-339613236.

However, because request fingerprinting takes into account information beyond the URL, I created a new meta key just for this purpose, where redirect chain fingerprints are stored as a set.

These changes seem to fix @immerrr’s code example to reproduce the issue (https://github.com/scrapy/scrapy/issues/1225#issuecomment-265388573).

Pending:
- [x] In redirect handling, use the same request fingerprinting as the duplicate filter. This should be easy now that https://github.com/scrapy/scrapy/pull/4524 has been merged.
- [x] Write tests for the scenarios discussed in #1225.

Post-merge work:
- [ ] Re-evaluate https://github.com/scrapy/scrapy/issues/3276


wRAR:
Should this be removed from 2.7?

kmike:
@wRAR yes; removed. I acknowledge it's looking close, but not having this feature doesn't look like a blocker for 2.7 release.

scrapy/downloadermiddlewares/redirect.py
```diff
@@ -60,12 +63,14 @@ def _redirect(
...
-                "redirect_reasons", []
-            ) + [reason]
+            redirect_urls = request.meta.get("redirect_urls", []) + [request.url]
+            redirected.meta["redirect_urls"] = redirect_urls
+            redirect_reasons = request.meta.get("redirect_reasons", []) + [reason]
+            redirected.meta["redirect_reasons"] = redirect_reasons
+            fingerprints = request.meta.get("redirect_fingerprints", set())
+            assert self.crawler.request_fingerprinter is not None
+            fingerprint = self.crawler.request_fingerprinter.fingerprint(request)
+            redirected.meta["redirect_fingerprints"] = fingerprints | {fingerprint}
```
kmike, line 73: The approach looks good to me.
However, I wonder if it'd cause issues in practice, in case someone uses `request.meta.copy()`. Unlike most other meta keys, it seems this one may affect the behavior significantly.

Do you think there is a room for making it less likely to happen? E.g. try to clear redirect_fingerprints once the response without a redirect is received? Or would it cause more edge cases than it'd solve?

Gallaecio: If we clean it in scenarios where the redirect middlewares return `response`, i.e. do not handle a redirect, there is the problem that the fact that one middleware does not handle a redirect does not mean that the other does not. So I discarded that approach.

I thought about removing them from the first of the middlewares (last on response processing), but then if that middleware is disabled, the cleanup would not happen.

We could also create a middleware specifically for the purpose of this cleanup. It feels weird, but it also feels like the cleanest approach to deal with this middleware duality and optionality.

Any thoughts or additional ideas?

PR #4318, opened 2020-02-07
Support yield in async def callbacks.

wRAR:
Based on #3485. Requires Python 3.6.

I think all results of a callback are only processed when it finishes, but this seems to be the case with non-async callbacks too.


scrapy/utils/spider.py
```diff
@@ -4,12 +4,20 @@
...
+    from scrapy.utils.py36 import collect_asyncgen
+except SyntaxError:
+    collect_asyncgen = None
 
 
 logger = logging.getLogger(__name__)
 
 
 def iterate_spider_output(result):
+    if collect_asyncgen and hasattr(inspect, 'isasyncgen') and inspect.isasyncgen(result):
```
kmike: Could you please add a comment here, something like "workaround to support async generators in Python 3.6"?

kmike:
Interesting - does it mean the whole thing worked with async generators already in Python 3.7+?

tests/py36/_test_crawl.py
```diff
@@ -0,0 +1,50 @@
...
+    def start_requests(self):
+        for i in range(self.initial_reqs):
+            yield self._get_req(i)
+
+    async def parse(self, response):
+        index = response.meta['index']
+        yield {'index': index}
+        if index < 10 ** self.depth:
+            for new_index in range(10 * index, 10 * index + self.following_reqs):
+                yield self._get_req(new_index)
```
kmike: For paranoid me: could you please check that non-default callback also works, and that async def works for methods other than parse?

kmike:
@Gallaecio I think Python 3.6 milestone can be removed, because this PR doesn't break compatibility with Python < 3.6.

kmike:
> I think all results of a callback are only processed when it finishes

This is an important limitation, which we should be very upfront about. I'm fine with merging it as-is, as soon as it doesn't make things worse for non-async def callbacks.

> but this seems to be the case with non-async callbacks too.

@wRAR could you please double-check how regular callbacks behave (for requests and for items)?

wRAR:
@kmike for generator callbacks the generator itself is passed to [Scraper.handle_spider_output](https://github.com/scrapy/scrapy/blob/1.8/scrapy/core/scraper.py#L171) where it's iterated and items are passed to `Scraper._process_spidermw_output`. On the way there it's wrapped into a `MutableChain` and into several functions that catch exceptions while iterating it. I'm currently unable to say which parts of this should be changed in order to support passing async generators without turning them to lists first. On the other hand, looks like this PR doesn't support graceful handling of exceptions in async generator callbacks, I'll need to write some tests for that and see what happens.

kmike:
Thanks for the investigation @wRAR. Let's have it in, to make `async def` callbacks slightly easier to use in the meantime.

Gallaecio:
> On the other hand, looks like this PR doesn't support graceful handling of exceptions in async generator callbacks, I'll need to write some tests for that and see what happens.

@wRAR If so, I guess we should mention that in the documentation about asyncio support introduced in https://github.com/scrapy/scrapy/pull/4329

wRAR:
@Gallaecio @kmike I've just checked and if a callback yields an item and then raises an Exception, the item will be produced from the spider if the callback is normal and it won't be produced if it's an async generator.

PR #4323, opened 2020-02-08
updated with new macOS name (#4308)

### LINKED ISSUE #4308: Mac OS X, OS X → macOS
We have a few references in the documentation where we use the old name of that OS. We should update them.

MaliYudina:
changed Mac OS X, OS X for new macOS name

(edit) fixes #4308


docs/intro/install.rst
```diff
@@ -165,11 +165,11 @@ you can install Scrapy with ``pip`` after that::
 
 .. _intro-install-macos:
 
-Mac OS X
+macOS
 --------
```
elacuesta:
```suggestion
-----
```

docs/news.rst
```diff
@@ -1076,7 +1076,7 @@ Cleanups & Refactoring
 ~~~~~~~~~~~~~~~~~~~~~~
 
 - Tests: remove temp files and folders (:issue:`2570`),
-  fixed ProjectUtilsTest on OS X (:issue:`2569`),
```
elacuesta: I wonder if we should change occurrences in this file, or leave them as they are since the OS was named differently by the time those versions were released.

Gallaecio: It’s a valid point. 

I’m OK with the change, though. They all have the same [Wikipedia article](https://en.wikipedia.org/wiki/MacOS) :).

If we were mentioning a specific macOS version, then I think it would make sense to use the same branding they used for that specific version, though, unless they had rebranded it as macOS retroactively upstream.

elacuesta: Yeah sounds fine to me, I was just thinking out loud.

PR #1649, opened 2015-12-29
make spider name optional #1157

aron-bordin:
## PR Overview
This pr makes the name attribute optional(see https://github.com/scrapy/scrapy/issues/1157#issue-68467857).
### New behavior
If the class attr `name` is None, the class name will be used. 

Some scrapy components access the class name before creating an instance of the Spider. So the name attr and its default value must be visible before/after creating the obj instance.

I started using a classmethod `get_spider_name`, however I needed to edit some points to make it compatible(https://github.com/scrapy/scrapy/blob/master/scrapy/spiderloader.py#L25, https://github.com/scrapy/scrapy/blob/master/scrapy%2Fcommands%2Fcheck.py#L78, and more). And then it was failing in the tests because some methods use Spider classes and instances(https://github.com/scrapy/scrapy/blob/master/scrapy%2Futils%2Furl.py#L32), so classmethod was not compatible here. I just found this one because it failed in a test, so I think that it's  safer to implement something that is compatible with the old name attr.

So I used a new approach, a class property.

Some tests were summing that Spider0 was not valid because it had no name. These tests were updated to handle this class, and there is a test to check if the default name is being used.

PR #4327, opened 2020-02-10
Make spider names optional

### LINKED ISSUE #1157: Don't require 'name' attribute for scrapy.Spider
I think we should make Spider.name attribute optional. The name is used by SpiderManager to find spiders, but Spider can be used without a Scrapy project. It is unnecessary boilerplate for users of runspider command or for CrawlerRunner / CrawlerProcess users. 

We can also provide a default value, e.g. `self.__class__.__name__`, to help with discovery; this have an advantage of 1-to-1 mapping between spider class names and names printed to users - spiders can become easier to find.

Gallaecio:
-   Add a new ``scrapy.spiders.ignore_spider`` decorator.

    The default spider loader ignores classes with this decorator.

-   Add a new ``SPIDER_LOADER_REQUIRE_NAME`` boolean setting.

    When ``SPIDER_LOADER_REQUIRE_NAME`` is ``True``, the default spider loader
    loads all spiders that have a non-empty name (current behavior) and are not
    decorated with ``ignore_spider``.

    When ``SPIDER_LOADER_REQUIRE_NAME`` is ``False``, the default spider loader
    loads all spiders that are not decorated with ``ignore_spider``. If they
    do not have a non-empty name, the fully-qualified class name is used
    instead.

    ``SPIDER_LOADER_REQUIRE_NAME`` is ``True`` by default. New projects have it
    set to ``False``. ~Having it set to ``True`` causes a warning which can be
    silenced by setting it to ``False``.~

Fixes #1157, closes #1649


scrapy/spiderloader.py
```diff
@@ -33,12 +41,15 @@ def _check_name_duplicates(self):
             msg = ("There are several spiders with the same name:\n\n"
                    "{}\n\n  This can cause unexpected behavior.".format(
                         "\n\n".join(dupes)))
-            warnings.warn(msg, UserWarning)
+            warn(msg, UserWarning)
```
kmike, line 44: 💄 I usually prefer warning.warn, because otherwise it is less clear if that's logging.warn or warning.warn.

scrapy/spiderloader.py
```diff
@@ -18,6 +18,16 @@ class SpiderLoader:
...
+        self.require_name = settings.getbool('SPIDER_LOADER_REQUIRE_NAME')
+        if self.require_name:
+            message = (
+                'SPIDER_LOADER_REQUIRE_NAME is True. In a future version of '
+                'Scrapy, the SPIDER_LOADER_REQUIRE_NAME setting will be '
+                'removed, and Scrapy will always behave as if '
+                'SPIDER_LOADER_REQUIRE_NAME were False. To remove this '
+                'warning, set SPIDER_LOADER_REQUIRE_NAME to False.'
+            )
+            warnings.warn(message, ScrapyDeprecationWarning)
```
kmike, line 30: I don't have a strong opinion on that, but for robots.txt enable option we did a different thing:

* keep old default (robotx.txt handling diisabled)
* set new default in generated setting.py, so that new projects get a new default
* try to document carefully these default values - technically, default is False, but users should expect this option to be True
* the setting itself is not deprecated

Here the approach is different; the goal is to move to a new way of defining spiders, not to provide an alternative.
Do you think there are cases where the current behavior is preferrable? I wonder if we should try moving to the new approach like we did for robots.txt first, and then decide if we should deprecate the current approach or not. Or is the new approach clearly better than the old one?

Gallaecio: I see no reason why we cannot maintain both behaviors. I’ll make the required changes.

Gallaecio:
I wonder if instead of `@basespider` we should name the decoration something else that clearly indicates that the end result is simply that Scrapy won’t load it. I cannot think of a good name, though. `@hide`? `@skip`? `@unlist`? `@dont_run`? Their past participle equivalents (hidden, skipped…)?

kmike:
I think the implementation is fine. Is it blocked on naming @Gallaecio?

scrapy/spiders/__init__.py
```diff
@@ -14,6 +14,20 @@
...
+    """Marks a :class:`~scrapy.spiders.Spider` subclass as a :ref:`base spider
+    <base-spiders>`."""
+
+    @classmethod
+    def is_abstract(cls):
+        if cls is decorated_cls:
+            return True
+        return super(decorated_cls, cls).is_abstract()
+
+    decorated_cls.is_abstract = is_abstract
```
kmike, line 27: Naming nitpick: we talk about base spiders in docs, but there a method is called is_abstract. 
Not sure I like `is_basespider` though.

noviluni (commented):
I really like this pattern! :smile: 

Some ideas:
* I like how "`@abstractspider`" sounds, but it seems related to the common `@abstractmethod` decorator and it could seem that the spider can't be instantiated or that their methods can't be implemented.
* The idea of using `@basespider` seems accurate, as this decorator will be used to mark those "base spiders" that are not expected to be used directly. But the `is_abstract()` method should be probably renamed.
* Of course, you can go with something generic like `@skip` , `@ignore`, `@hidden` or whatever you mentioned, but it could conflict with other libraries. So in case you decide for one of these names I would add the "spider" word. Example: `@ignore_spider`.

Good job!

docs/topics/settings.rst
```diff
@@ -1351,6 +1351,30 @@ Default: ``'scrapy.spiderloader.SpiderLoader'``
...
+:class:`~scrapy.spiders.Spider` subclasses unless they are decorated with
+:func:`~scrapy.spiders.ignore_spider`. If they do not have a non-empty
+:class:`~scrapy.spiders.Spider.name`, their fully-qualified class name is used
+as a name.
+
+.. note::
+
+    While the default value is ``True`` for historical reasons, this option is
+    disabled by default in the ``settings.py`` file generated by the
+    :command:`startproject` command.
```
kmike, line 1376: What do you think about moving this note up, closer to "Default: True"?

docs/topics/spiders.rst
```diff
@@ -826,3 +832,29 @@ Combine SitemapSpider with other sources of urls::
...
+
+For example::
+
+    from scrapy.spiders import ignore_spider, Spider
+
+    @ignore_spider
+    class MyBaseSpider(Spider):
+        pass
+
+If :setting:`SPIDER_LOADER_REQUIRE_NAME` is ``True`` (default), any
```
kmike, line 858: I woder if default True would be confusing here - does it require more elaboration? For new users default is going to be False, right?

felipeboffnunes:
@Gallaecio @kmike are the only missing pieces the comments about documentation? Seems this one is almost mergeable.

Gallaecio:
Just one of many PRs I have open that are not a priority, and that I will eventually resume work on.

docs/topics/spiders.rst
Gallaecio: I think removing “(default)” altogether will make users open the setting link, and there they will be able to find the details about it.

docs/topics/spiders.rst
```diff
@@ -59,16 +59,22 @@ scrapy.Spider
...
+        :command:`runspider`.
+
+        The spider name must be unique to one spider class. If two or more
+        spiders have the same name, Scrapy commands :command:`crawl` and
+        :command:`runspider` will only be able to run one of the spiders.
+
+        If the spider scrapes a single domain, a common practice is to name the
+        spider after the domain, with or without the `TLD`_. So, for example, a
+        spider that crawls ``mywebsite.com`` would often be called
+        ``mywebsite``.
```
wRAR, line 77: I think this has an extra indent of 1?

Gallaecio: It was kind of on purpose, because the original uses 7 spaces (the first indentation uses 3, the second 4), but since the issue affects the whole class definition I’ll revert.

wRAR (approved):

scrapy/spiders/__init__.py
```diff
@@ -34,12 +52,14 @@ class Spider(object_ref):
     def __init__(self, name: Optional[str] = None, **kwargs: Any):
         if name is not None:
             self.name = name
-        elif not getattr(self, "name", None):
-            raise ValueError(f"{type(self).__name__} must have a name")
```
kmike, line 38: It looks like spider.name is going to raise an AttributeError for nameless spiders. Won't it break spider.name usages found by @GeorgeA92 in https://github.com/scrapy/scrapy/issues/1157#issuecomment-1816782204?

scrapy/spiderloader.py
```diff
@@ -53,9 +54,12 @@ def _check_name_duplicates(self) -> None:
             )
 
     def _load_spiders(self, module: ModuleType) -> None:
-        for spcls in iter_spider_classes(module):
-            self._found[spcls.name].append((module.__name__, spcls.__name__))
-            self._spiders[spcls.name] = spcls
+        classes = iter_spider_classes(module, require_name=self.require_name)
+        for spcls in classes:
+            qualname = ".".join((module.__name__, spcls.__name__))
```
kmike, line 59: should it use utils.global_object_name?

kmike: Or is the result actually different?

scrapy/spiders/__init__.py
```diff
@@ -35,11 +54,15 @@ def __init__(self, name: Optional[str] = None, **kwargs: Any):
         if name is not None:
             self.name = name
         elif not getattr(self, "name", None):
-            raise ValueError(f"{type(self).__name__} must have a name")
+            self.name = global_object_name(self.__class__)
```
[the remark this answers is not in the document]
Gallaecio: This would make spider *objects* get a name, which would solve https://github.com/scrapy/scrapy/issues/1157#issuecomment-1816782204, but spider *classes* would remain nameless.

Alternatively, we could go with the approach in https://github.com/scrapy/scrapy/pull/1649, as @GeorgeA92 [suggested](https://github.com/scrapy/scrapy/issues/1157#issuecomment-1896234239).

PR #4328, opened 2020-02-12
fix bug

wwj718:


elacuesta:
Thanks for the contribution @wwj718. This particular issue was already fixed by #4290, which was merged to the `master` branch.

Issue #4285, opened 2020-01-24
Fix "Debugging memory leaks with Guppy" section (docs)


noviluni:
After removing the Python 2.7 support, this section:

https://docs.scrapy.org/en/latest/topics/leaks.html#debugging-memory-leaks-with-guppy

 should be removed or merged with this:
https://docs.scrapy.org/en/latest/topics/leaks.html#topics-leaks-muppy

kmike:
A good catch!

sakshamb2113:
Can i work on this issue?

noviluni:
@sakshamb2113 of course, go ahead!

sakshamb2113:
![Screenshot from 2020-02-18 22-11-36](https://user-images.githubusercontent.com/44064539/74757696-5ef96580-529c-11ea-931b-e3b23c67d59c.png)
there is one more reference of guppy in the beginning of the page.Should i change this too?

krisblarq:
I see this issue is still open. Are there any other issues to be worked on?.

elacuesta:
No reason, the PR just wasn't linked, either [manually](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue) or with one of the available [keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Closing now, thanks for noticing!

PR #4343, opened 2020-02-18
Fixed "Debugging memory leaks with Guppy"( issue #4285)

sakshamb2113:
The section referring to guppy(for python 2.7) for debugging memory leaks now only refers to muppy(for python 3). Issue #4285.


sakshamb2113:
@noviluni please review

noviluni:
Hi @Gallaecio 

I've seen that after removing this we should probably delete these lines:

```
try:
    import guppy
    hpy = guppy.hpy()
except ImportError:
    hpy = None
```

from the `scrapy/extensions/telnet.py`. 

Do you agree?

scrapy/extensions/telnet.py
```diff
@@ -110,7 +105,7 @@ def _get_telnet_vars(self):
             'est': lambda: print_engine_status(self.crawler.engine),
             'p': pprint.pprint,
             'prefs': print_live_refs,
-            'hpy': hpy,
+            'hpy': None,
```
Gallaecio: I wonder what to do with this:
- Deprecate it somehow?
- Remove it?
- Use muppy?

In options 1 and 2, we might want to use muppy with a different name.

All of this is assuming muppy offers hpy or something similar, I’m not familiar with muppy or guppy at all.

sakshamb2113: Can't we use muppy along with summary modules in pympler in place of guppy.hpy() as described in the documentation ?

wRAR:
@sakshamb2113 can you please update this to resolve the conflict?

sakshamb2113:
sure

sakshamb2113:
@wRAR what should we do about the "hpy" key of telnet_vars in telnet.py ?

scrapy/extensions/telnet.py
Gallaecio: I think we should then just remove this key, instead of changing it to `None`.

PR #4363, opened 2020-02-21
Adds docs for download_slot meta key for Request

### LINKED ISSUE #869: (Feature Request) Throttling number of spiders running concurrently in the same process
It would be nice to be able to throttle number of spiders running concurrently from the script as in a docs in "Common Practices" (http://doc.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process)

I was doing spider, which started them few hundreds at once, it didn't work, logging didn't even started.  
It would be nice if we could easily set this number. 

Thank you very much, 
You're doing great job guys!

### LINKED ISSUE #3585: request.meta['download_slot'] is not documented
request.meta['download_slot'] is not documented in https://doc.scrapy.org/en/latest/topics/request-response.html#request-meta-special-keys.

asimhashmi:
Adds the missing documentation related to key `download_slot` in `Request` meta.

Resolves #3585, resolves #869.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -364,6 +364,18 @@ started, i.e. HTTP message sent over the network. This meta key only becomes
 available when the response has been downloaded. While most other meta keys are
 used to control Scrapy behavior, this one is supposed to be read-only.
 
+.. reqmeta:: download_slot
+
+download_slot
+----------------
+
+The website domain to which the request is destined. It is used to provide info
+when :setting:`AUTOTHROTTLE_ENABLED` is set to `True`. This meta key only becomes
+available when the response has been downloaded. While most other meta keys are
+used to control Scrapy behavior, this one is supposed to be read-only.
+See also: :ref:`topics-autothrottle` and its
+:setting:`AUTOTHROTTLE_ENABLED` option.
+
 .. reqmeta:: download_fail_on_dataloss
 
 download_fail_on_dataloss
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -332,6 +332,7 @@ Those are:
 * :reqmeta:`download_timeout`
 * :reqmeta:`download_maxsize`
 * :reqmeta:`download_latency`
+* :reqmeta:`download_slot`
 * :reqmeta:`download_fail_on_dataloss`
 * :reqmeta:`proxy`
 * ``ftp_user`` (See :setting:`FTP_USER` for more info)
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -364,6 +364,18 @@ started, i.e. HTTP message sent over the network. This meta key only becomes
...
+
+download_slot
+----------------
+
+The website domain to which the request is destined. It is used to provide info
+when :setting:`AUTOTHROTTLE_ENABLED` is set to `True`. This meta key only becomes
+available when the response has been downloaded. While most other meta keys are
+used to control Scrapy behavior, this one is supposed to be read-only.
+See also: :ref:`topics-autothrottle` and its
+:setting:`AUTOTHROTTLE_ENABLED` option.
```
kmike: Hey! Unfortunately this explanation is not correct; this meta key is not related to AutoThrottle directly, it is a general Downloader thing, and it is not supposed to be read-only - it allows to change a default slot for a domain, and tune concurreency options this way.

I'd suggest to either study how this slot is used, by checking Downloader source code, and maybe some examples (e.g. https://github.com/scrapy-plugins/scrapy-splash), to really understand what is this option for, or try helping Scrapy in some other way - download_slot is an advanced feature, which is not that easy to document.

asimhashmi: Hey @kmike Thanks for pointing out! I already looked at the `Downloader` source code, I understand that it uses hostname as slot when dealing with concurrency. That key has little usage in the actual project, that's it's actual purpose wasn't clear. With the example you provided I'll explore the usage and come up with a better explanation for this key.

asimhashmi: HI @kmike, can you please have a look? Any feedback is welcomed.

docs/topics/request-response.rst
```diff
@@ -364,6 +365,20 @@ started, i.e. HTTP message sent over the network. This meta key only becomes
 available when the response has been downloaded. While most other meta keys are
 used to control Scrapy behavior, this one is supposed to be read-only.
 
+.. reqmeta:: download_slot
+
+download_slot
+----------------
```
Gallaecio:
```suggestion
-------------
```

docs/topics/request-response.rst
```diff
@@ -364,6 +365,20 @@ started, i.e. HTTP message sent over the network. This meta key only becomes
...
+
+By default, it stores the hostname of the website to which request is destined.
+Scrapy uses slots as queues to process requests and `download_slot` represent the slot
+in which that request is going to process. If a slot doesn't exist then Scrapy creates
+it and uses it to process requests associated with it. A slot adds different constraints
+to the requests to tune up the concurrency and add delays e.g. :setting:`DOWNLOAD_DELAY`,
+:setting:`CONCURRENT_REQUESTS_PER_DOMAIN`, :setting:`CONCURRENT_REQUESTS_PER_IP` etc.
+By default all the requests related to a domain are processed in single slot.
+But you can set this key to whatever you like when sending the request.
```
Gallaecio: A few suggestions:
- Start by explaining what this request meta key does, not its default value.
- Use paragraphs.
- “If a slot doesn't exist then Scrapy creates it and uses it to process requests associated with it.” feels like an implementation detail, introduced before, from the reading so far, I could understood what slots are for.
- It is not clear to me, from reading, how those settings relate to slots. What happens if I give a value to each of those 3 settings, and then I use a custom download slot?
- It may be helpful to cover some of the use cases of this feature.

felipeboffnunes:
Hey @Gallaecio, do you believe adjusting the last comments will do the trick for this one? Could you give it a brief look? If it seems close to an end-product I can finish it up (starting from your latest comments).

Gallaecio:
> do you believe adjusting the last comments will do the trick for this one?

I think that would be a good starting point, but I also think additional work may be needed.

AdrianAtZyte:
https://github.com/scrapy/scrapy/pull/7933

PR #4116, opened 2019-10-31
Remove manual pypy installation from .travis.yml.

wRAR:

PR #4366, opened 2020-02-22
Remove manual pypy installation from .travis.yml.

wRAR:
Trying #4116 again


wRAR:
Looks like the problem still persists. It also looks like they still use the same pypy3 version as the last time so that was expected.

wRAR:
```
$ python --version
187Python 3.6.1 (784b254d6699, Apr 14 2019, 10:22:42)
188[PyPy 7.1.1-beta0 with GCC 6.2.0 20160901]
```

Issue #4296, opened 2020-01-31
Merge Slot back into ExecutionEngine?


wRAR:
In 409aaade `Slot` was extracted from `ExecutionEngine` and they had a many-to-one rel. In a84e5f8 that rel was changed to one-to-one, and I'm not sure what does this split achieve now.

wRAR:
This is even more confusing, actually. And it's not what I wanted to create this ticket about initially.

There is another `Slot`, `scrapy.core.scraper.Slot`, and it has a one-to-one rel with `scrapy.core.scraper.Scraper`, which was many-to-one before 58975abe.

There is also `scrapy.core.downloader.Slot`, which still has a many-to-one rel with `scrapy.core.downloader.Downloader` :)

Gallaecio:
If nothing else, we should rename those classes :smile:

nyov:
Yeah, they should definitely be called `Spot`. ;)

#1392

apurvTri123:
Hi can I work on this?

wRAR:
hi @apurvTri123 , this is a discussion ticket, we are not sure what do we want to do with this, if anything at all.

AdrianAtZyte:
https://github.com/scrapy/scrapy/pull/6931 will deprecate `downloader.Slot`, so we can probably keep that off the scope here.

AdrianAtZyte:
Should we just privatize `scraper.Slot` and consider this done?

wRAR:
I'm still not sure what does a split achieve, though if there is a clear difference in responsibilities between the main class and the slot class in both cases it makes sense to have them separate. And yeah we should make the second one private too.

Issue #1392, opened 2015-07-29
rename "anon" Slot classes in the code


nyov:
I wonder what people would say to this change?

```patch
diff --git a/scrapy/core/downloader/__init__.py b/scrapy/core/downloader/__init__.py
index 9654718..743a891 100644
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -15,7 +15,7 @@ from .middleware import DownloaderMiddlewareManager
 from .handlers import DownloadHandlers
 
 
-class Slot(object):
+class DownloadSlot(object):
     """Downloader slot"""
 
     def __init__(self, concurrency, delay, randomize_delay):
@@ -48,7 +48,7 @@ class Slot(object):
 
     def __str__(self):
         return (
-            "<downloader.Slot concurrency=%r delay=%0.2f randomize_delay=%r "
+            "<downloader.DownloadSlot concurrency=%r delay=%0.2f randomize_delay=%r "
             "len(active)=%d len(queue)=%d len(transferring)=%d lastseen=%s>" % (
                 self.concurrency, self.delay, self.randomize_delay,
                 len(self.active), len(self.queue), len(self.transferring),
@@ -105,7 +105,7 @@ class Downloader(object):
         if key not in self.slots:
             conc = self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
-            self.slots[key] = Slot(conc, delay, self.randomize_delay)
+            self.slots[key] = DownloadSlot(conc, delay, self.randomize_delay)
 
         return key, self.slots[key]
 
diff --git a/scrapy/core/engine.py b/scrapy/core/engine.py
index 992327b..6e1a47b 100644
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -21,7 +21,7 @@ from scrapy.utils.log import logformatter_adapter, failure_to_exc_info
 logger = logging.getLogger(__name__)
 
 
-class Slot(object):
+class EngineSlot(object):
 
     def __init__(self, start_requests, close_if_idle, nextcall, scheduler):
         self.closing = False
@@ -231,7 +231,7 @@ class ExecutionEngine(object):
         nextcall = CallLaterOnce(self._next_request, spider)
         scheduler = self.scheduler_cls.from_crawler(self.crawler)
         start_requests = yield self.scraper.spidermw.process_start_requests(start_requests, spider)
[...]

PR #4367, opened 2020-02-22
Rename three different Slot classes.

wRAR:
Related to #4296.


wRAR:
#1392

Gallaecio:
I wonder if we should deprecate the old names instead of just changing them.

wRAR:
@Gallaecio they don't seem to be public, and I think if we cannot just drop them then it's easier to  keep them.

Gallaecio:
I tend to consider public anything that does not start with a `_` XD. However, I guess if they are not documented we _could_ consider them implementation details. I’ll leave it for @kmike to decide.

PR #4329, opened 2020-02-12
Write release notes for Scrapy 2.0.0

Gallaecio:

PR #4370, opened 2020-02-24
Make BaseItemExporter’s dont_fail parameter keyword-only

Gallaecio:
Make the new `dont_fail` parameter of `BaseItemExporter` keyword only.

Otherwise, we are adding an non-intuitive positional parameter to a class that used to require keyword parameters only.

`dont_fail` was introduced after 1.8, so it would be great to have this change in for 2.0, otherwise we will need to introduce this change with a deprecation message for positional argument usages.

Split out from #4329


kmike:
Thanks @Gallaecio!

PR #4372, opened 2020-02-25
Changed test_fixed_delay(fixes #4268)

### LINKED ISSUE #4268: Random failures in CrawlTestCase.test_fixed_delay
On some test runs `CrawlTestCase.test_fixed_delay`  fails with
```
>       yield self._test_delay(total=3, delay=0.1)
E   twisted.trial.unittest.FailTest: twisted.trial.unittest.FailTest: True is not false : test total or delay values are too small
```

sakshamb2113:
changed delay to 0.11 in test_fixed_delay(). This will greatly reduce any chance of the test failing randomly.

Fixes #4268


wRAR:
@Gallaecio @sakshamb2113 I didn't try this change yet, but I don't have this problem on every run anyway, so I don't know how to make sure it helps :)

sakshamb2113:
@wRAR I wrote a simple bash script to execute the command
tox -e py3 tests/test_crawl.py::CrawlTestCase::test_fixed_delay
 for running tests multiple times.

wRAR:
I ran the tests with the change and I got the error once. I don't know how many runs would I do to get an approximate error rate, but I'd prefer the error to be fixed completely.

sakshamb2113:
@wRAR could you tell me what method did you use to test the function repeatedly ?

sakshamb2113:
I have changed delay to 0.2. I have not got any error until now. @wRAR could you please check it ?

wRAR:
> @wRAR could you tell me what method did you use to test the function repeatedly ?

Just `for i in {1..50}; tox -e py35 tests/test_crawl.py::CrawlTestCase::test_fixed_delay` and looking for the red lines manually.

sakshamb2113:
@wRAR I have run the tests many times with delay=0.2 without any error . I guess there is no case when this value of delay is too small.

wRAR:
Looks like it's fine with 0.2 on my machine.

PR #4373, opened 2020-02-25
Fix the ReadTheDocs build

Gallaecio:
@kmike noticed that autodoc was not working, any API documentation extracted from comments was missing in `master`.

This started happening when we removed Scrapy dependencies from `docs/requirements.txt`. The build log issued warnings, but Read The Docs was not configured to fail the build in case of warnings. Travis CI would have failed, weren’t it because Tox automatically installs Scrapy from `setup.py`, which ReadTheDocs does not.

These changes add the dependencies back to the file, and make warnings cause build failures in ReadTheDocs.

You can see it working at https://scrapy-gallaecio.readthedocs.io/en/fix-readthedocs/topics/request-response.html

I also tried just enabling to fail on warning, and indeed the ReadTheDocs build failed after that.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,18 +2,3 @@ Sphinx>=2.1
 sphinx-hoverxref
 sphinx-notfound-page
 sphinx_rtd_theme
-
-# Required for ReadTheDocs
-# Keep in sync with setup.py
-Twisted>=17.9.0
-cryptography>=2.0
-cssselect>=0.9.1
-lxml>=3.5.0
-parsel>=1.5.0
-PyDispatcher>=2.0.5
-pyOpenSSL>=16.2.0
-queuelib>=1.4.2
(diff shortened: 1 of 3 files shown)


docs/requirements.txt
```diff
@@ -2,3 +2,18 @@ Sphinx>=2.1
 sphinx-hoverxref
 sphinx-notfound-page
 sphinx_rtd_theme
+
+# Required for ReadTheDocs
```
kmike: I wonder if instead of duplicating the requirements we can ask readthedocs to install Scrapy using pip; this seems to be supported (see https://docs.readthedocs.io/en/stable/config-file/v2.html#packages)

Gallaecio: It worked, cool. TIL

kmike:
Thanks @Gallaecio!

PR #4377, opened 2020-02-25
option to output data in scrapy parse

### LINKED ISSUE #4317: Having an option to define an Items output file in the `scrapy parse` commnad line
I use `scrapy parse` command to check the Items output for given URL, and I use it a lot.

Sometimes, I need to latter explore the Items extracted by `scrapy parse`, e.g . to check that something I want to fix was fixed for that specific URL without having to touch the source code.

### Proposed solution:

Add a new parameter `-o` or `--output` in the parse command line. Allowing the client to set a desired output file for the items. 

I could also `|` the result of `scrapy parse`, filter out want I want and create a `json` file out of it, but would be much easier if I could simple write `scrapy parse {url} --callback=fixed_function -o items.json`

adityaa30:
- adds an option `-o` or `--output` in `scrapy parse` command to save
scrapped data to a file

- [x] fixes #4317
- [x] tests added/passed
- [x] add documentation


adityaa30:
@Gallaecio Okay. I have seen the implementation of `scrapy crawl -o` and will update this pr with a similar implementation for `scrapy parse`

wRAR:
It's also missing tests, though I cannot make codecov say it explicitly.

adityaa30:
So if I try to update `settings['FEED_URI']` and `settings['FEED_FORMAT']` as given in [crawl.py](https://github.com/scrapy/scrapy/blob/647cba0f106211e72a8d3e028b25c2f46859c406/scrapy/commands/crawl.py#L27), the signals defined in [FeedStorage](https://github.com/scrapy/scrapy/blob/647cba0f106211e72a8d3e028b25c2f46859c406/scrapy/extensions/feedexport.py#L131) are not be called. I cannot find the exact reason for this, but I debugged this behaviour is because of the custom `callback` defined in [parse.py](https://github.com/scrapy/scrapy/blob/4f31c3ce017db2b4f69949a81efd56fee60ee32d/scrapy/commands/parse.py#L164). @Gallaecio Can you please guide me here?

adityaa30:
@wRAR I am planning to write tests for this command after I finish the `--ouput` function. How shall I proceed to write tests?

wRAR:
@adityaa30 there are existing tests that run a command as a separate process, the parse command is tested in tests/test_command_parse.py

Gallaecio:
> So if I try to update `settings['FEED_URI']` and `settings['FEED_FORMAT']` as given in [crawl.py](https://github.com/scrapy/scrapy/blob/647cba0f106211e72a8d3e028b25c2f46859c406/scrapy/commands/crawl.py#L27), the signals defined in [FeedStorage](https://github.com/scrapy/scrapy/blob/647cba0f106211e72a8d3e028b25c2f46859c406/scrapy/extensions/feedexport.py#L131) are not be called. I cannot find the exact reason for this, but I debugged this behaviour is because of the custom `callback` defined in [parse.py](https://github.com/scrapy/scrapy/blob/4f31c3ce017db2b4f69949a81efd56fee60ee32d/scrapy/commands/parse.py#L164). @Gallaecio Can you please guide me here?

Can you push those non-working changes? The issue may be more obvious when looking at the code.

adityaa30:
@Gallaecio I have pushed my changes.

adityaa30:
I have been trying to find the exact problem why the `-o`/`--output` in 71098cfc789f361da9d5aef65613d39c5daf9ea1 do not work. By oberving the chain of callbacks in [scraper.py](https://github.com/scrapy/scrapy/blob/master/scrapy/core/scraper.py#L99-L202) 
> ...  → enqueue_scrape → (_scrape_next → _scrape → _scrape2 →  call_spider → iterate_spider_output →  handle_spider_error/handle_spider_output) x repeat → ... 

using a debugger I have observed that **[handle_spider_output](https://github.com/scrapy/scrapy/blob/2eb990a25232ba69faaf07440f3b749d6d175e4b/scrapy/core/scraper.py#L155)** in its argument `result` is only having a `scrapy.utils.python.MutableChain` of `scrapy.Request` i.e there is no `scrapy.item.BaseItem` or `dict` in it. I cannot seem to find the exact reason for this behaviour. @Gallaecio @wRaR Could you please help here?

adityaa30:
@wRaR about the tests, after we are able to fix the bug (in 71098cfc789f361da9d5aef65613d39c5daf9ea1) I think there will not be requirement for tests specifically for `parse` command as it will be using `scrapy.extensions.feedexport` (similar to `-o`/`--output` in `crawl`) which already has its tests in [test_feedexport.py](https://github.com/scrapy/scrapy/blob/master/tests/test_feedexport.py).

raphapassini:
Hey @adityaa30 you're right, is not working because of the callback and how specifically the command parse was designed.

Although a simple way to solve this is to change the print_items function https://github.com/scrapy/scrapy/blob/4f31c3ce017db2b4f69949a81efd56fee60ee32d/scrapy/commands/parse.py#L77 and call the proper exporter methods inside this function. The exporters interface is pretty simple https://github.com/scrapy/scrapy/blob/647cba0f106211e72a8d3e028b25c2f46859c406/scrapy/exporters.py you need to call start_exporting, export_item and then finish_exporting in this order and probably you're good to go.

raphapassini:
Other option is to understand exactly WHY the crawler is not connecting and listen to signal when we use the custom callback. This way we can use the same approach as the `crawl.py` command. But I think my suggestion above is simpler than doing that.

scrapy/commands/parse.py
```diff
@@ -233,6 +238,22 @@ def process_spider_arguments(self, opts):
             opts.spargs = arglist_to_dict(opts.spargs)
         except ValueError:
             raise UsageError("Invalid -a value, use -a NAME=VALUE", print_help=False)
+        if opts.output:
```
elacuesta: Please use [`scrapy.utils.conf.feed_process_params_from_cli`](https://github.com/scrapy/scrapy/blob/8845773d44329194ee73fa4985a94e768bf664e7/scrapy/utils/conf.py#L127-L172) to handle this. `FEED_URI` and `FEED_FORMAT` are deprecated in #3858 (to be released in 2.1)

adityaa30: Okay @elacuesta. Thanks a lot for mentioning 🙂

adityaa30:
> Other option is to understand exactly WHY the crawler is not connecting and listen to signal when we use the custom callback. This way we can use the same approach as the `crawl.py` command. But I think my suggestion above is simpler than doing that.

@raphapassini I debugged more and found out that the callback function was returning only a list of requests (see **[parse.py](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/parse.py#L209)**). When I change it from 
```python
return requests
```
to
```python
return items + requests
```
the `FeedExporter` works as in `crawl` command. However, we do get the logs for each scraped item. 
In 2a966e7740e8cb8fea2d7117a57011b2880ac173 my latest working changes are present, I had to pull rebase in order to get the changes done in #3858.

Gallaecio:
> I had to pull rebase

In the future, consider using merge to avoid force-pushing on a pull request:

```
git fetch upstream
git merge upstream/master
```

scrapy/commands/parse.py
```diff
@@ -206,7 +209,9 @@ def callback(response, **cb_kwargs):
                     req.meta['_depth'] = depth + 1
                     req.meta['_callback'] = req.callback
                     req.callback = callback
-                return requests
+                return items + requests
+            else:
+                return items  # dumps the remaining output to a FILE
```
Gallaecio, line 214: Maybe we should only do this when the new option is used, to minimize behavior changes?

adityaa30: Yes, that will be better. I will do that now

elacuesta:
This looks almost ready to merge, I just wonder if we should remove the `-t` option from the `parse` command, since it's deprecated in favour of specifying the format within the `-o` option.

elacuesta:
> I just wonder if we should remove the `-t` option from the `parse` command, since it's deprecated in favour of specifying the format within the `-o` option.

For the record, inheriting from the `BaseRunSpiderCommand` adds the `-t` option, which was not in the `parse` command before. I don't really have a problem with that, I think it's probably a reasonable trade-off to simplify the command's implementation. A deprecation message appears each time the option is used, so that's fine to me :ok_hand:

PR #4384, opened 2020-02-28
Drop horribly outdated deb package build files

nyov:
Debian packaging team might find this helpful.

cheers, @wRAR


nyov:
Debian has it's own `debian/` folder, which isn't based on this one. This here stems from scrapinghub's buildbot builds (and later was kept probably more for personal builds). Seeing how outdated it is, I don't think scrapinghub can still be building packages from it, though.

kmike:
+1 to remove all this. Just to be sure, let's summon @dangra to merge it :)

dangra:
Scrapinghub deb package have been dead for a long time. Long life to Debian packages

PR #4385, opened 2020-02-28
Obsolete REDIRECT_MAX_METAREFRESH_DELAY, LOG_UNSERIALIZABLE_REQUESTS

nyov:
Obsolete `REDIRECT_MAX_METAREFRESH_DELAY ` which has been deprecated since Scrapy 0.18
defc4f89b
and `LOG_UNSERIALIZABLE_REQUESTS` which has been deprecated since Scrapy 1.2.0
472a8a47


Gallaecio:
Could you please also remove the corresponding entries from `scrapy/settings/deprecated.py`?

nyov:
I was thinking of entirely removing all the entries there ;)
All of them are entirely obsolete (not just deprecated) and no longer found in the source.

But then I had the thought they they may also be meant to remind people, who still have them in legacy projects, where the setting doesn't do anything anymore (without people realising that, possibly).

Do you think we should clean them up, instead? I wouldn't mind that either.

Gallaecio:
As far as I known, the deprecation messages are only meant to last as much as the code that supports the deprecated functionality:

1. The functionality is implemented.
2. The functionality is deprecated, it continues to work but a warning is issued.
3. The functionality and the warning as removing, clean code!

So, at least for those two deprecated variables you are removing, I think we should remove them from that file as well.

As for the other ones, it really depends on how long those warnings have been there for. Based on https://github.com/scrapy/scrapy/issues/4356 I would say there is an agreement that deprecations 3+ years old are OK to remove. But it can be done in separate PRs.

nyov:
Ok, added. Err, removed.
I'll squash once review agrees that these should be removed here.

nyov:
Squashed

kmike:
Thanks @nyov!

PR #4390, opened 2020-02-28
async/deferred signal handlers

### LINKED ISSUE #4295: Document async signal handlers
Right now there are 7 signals whose handlers can return Deferreds: https://docs.scrapy.org/en/latest/topics/signals.html?highlight=deferred

With #4271 these signal handlers can also be `async def` coroutines.

AFAIK nowhere in the docs it is explained what does this mean and how can it be used. I looked at the code and the situation seems to be this:

`item_scraped`, `item_dropped`, `item_error`: I'm not 100% sure as I couldn't fully untangle the web of callbacks and Deferreds inside `Scraper` but it looks like until the handlers for items produced by a Request haven't finished, that Request isn't counted as processed, which means `spider_idle` isn't called and so the spider isn't closed, and may also mean some queue-related things. I traced the callback chain until [`Scraper.enqueue_scrape.finish_scraping`](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/core/scraper.py#L102) which also calls some `_scrape_next` method.

`engine_started`, `engine_stopped`, `spider_opened`, `spider_closed`: these are included in the spider/engine startup/shutdown processes and the process waits for the handler to finish before proceeding to the next line.

adityaa30:
Fixes #4295

### DIFF
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -71,12 +71,8 @@ Let's take an example::
         def item_scraped(self, item):
             # Send the scraped item to the server
             d = treq.post(
-                'http://localhost:3000/add',
-                json.dumps({
-                    'text': item['text'],
-                    'author': item['author'],
-                    'tags': item['tags'],
-                }).encode('ascii'),
+                'http://example.com/post',
+                json.dumps(item).encode('ascii'),
                 headers={b'Content-Type': [b'application/json']}
             )
 
@@ -91,10 +87,6 @@ Let's take an example::
                     'author': quote.css('small.author::text').get(),
                     'tags': quote.css('div.tags a.tag::text').getall(),
                 }
-            next_page = response.css('li.next a::attr(href)').get()
-            if next_page is not None:
-                next_page = response.urljoin(next_page)
-                yield scrapy.Request(next_page, callback=self.parse)
 
 See the :ref:`topics-signals-ref` below to know which signals support
 :class:`~twisted.internet.defer.Deferred`.
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -16,8 +16,7 @@ deliver the arguments that the handler receives.
 You can connect to signals (or send your own) through the
 :ref:`topics-api-signals`.
 
-Here is a simple example showing how you can catch signals and perform some action:
-::
+Here is a simple example showing how you can catch signals and perform some action::
 
     from scrapy import signals
     from scrapy import Spider
@@ -57,8 +56,7 @@ does not block Scrapy. If a signal handler returns a
 :class:`~twisted.internet.defer.Deferred`, Scrapy waits for that
 :class:`~twisted.internet.defer.Deferred` to fire.
 
-Let's take an example:
-::
+Let's take an example::
 
     class SignalSpider(scrapy.Spider):
         name = 'signals'
@@ -67,12 +65,12 @@ Let's take an example:
         @classmethod
         def from_crawler(cls, crawler, *args, **kwargs):
             spider = super(SignalSpider, cls).from_crawler(crawler, *args, **kwargs)
-            crawler.signals.connect(spider.item_scrapped, signal=signals.item_scraped)
+            crawler.signals.connect(spider.item_scraped, signal=signals.item_scraped)
             return spider
 
-        def item_scrapped(self, item):
-            # Send the scrapped item to the server
-            dfd = treq.post(
+        def item_scraped(self, item):
+            # Send the scraped item to the server
+            d = treq.post(
                 'http://localhost:3000/add',
                 json.dumps({
                     'text': item['text'],
@@ -81,14 +79,11 @@ Let's take an example:
                 }).encode('ascii'),
                 headers={b'Content-Type': [b'application/json']}
             )
-            dfd.addCallback(print_response)
+            d.addCallback(print_response)
 
-            # Next item will be scrapped only after dfd is fired
-            # As items are processed in parallel many dfd will
-            # be fired at the same time (refer settings['CONCURRENT_ITEMS'])
-            # hence, next batch of items will be processed after all
-            # dfd's are fired
-            return dfd
+            # The next item will be scraped only after
+            # deferred (d) is fired
+            return d
 
         def parse(self, response):
             for quote in response.css('div.quote'):
@@ -103,7 +98,7 @@ Let's take an example:
                 yield scrapy.Request(next_page, callback=self.parse)
 
 See the :ref:`topics-signals-ref` below to know which signals support
-:class:`~twisted.internet.defer.Deferred`
+:class:`~twisted.internet.defer.Deferred`.
 
 .. _topics-signals-ref:
 
@@ -115,8 +110,11 @@ Built-in signals reference
 
 Here's the list of Scrapy built-in signals and their meaning.
 
-engine_started
+Engine signals
 --------------
+
+engine_started
+~~~~~~~~~~~~~~
 
 .. signal:: engine_started
 .. function:: engine_started()
@@ -130,7 +128,7 @@ engine_started
     getting fired before :signal:`spider_opened`.
 
 engine_stopped
---------------
+~~~~~~~~~~~~~~
 
 .. signal:: engine_stopped
 .. function:: engine_stopped()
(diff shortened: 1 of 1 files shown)


docs/topics/signals.rst
```diff
@@ -54,6 +54,23 @@ Some signals support returning :class:`~twisted.internet.defer.Deferred`
 objects from their handlers, see the :ref:`topics-signals-ref` below to know
 which ones.
 
+.. note::
+
+    If a signal handler returns a :class:`~twisted.internet.defer.Deferred`,
+    the spider waits for that :class:`~twisted.internet.defer.Deferred` to fire.
+    The signal handler can also return a :class:`~twisted.internet.defer.Deferred`
+    obtained from some async operation which will fire as the operation is finished.
```
Gallaecio: I think this could be merged into the previous paragraph, and the reference to the list of signals be moved after the example.

For example:

> Some signals support returning :class:`~twisted.internet.defer.Deferred`
objects from their handlers, allowing you to run asynchronous code that does not block Scrapy. If a signal handler returns a :class:`~twisted.internet.defer.Deferred`, Scrapy waits for that :class:`~twisted.internet.defer.Deferred` to fire.
>
> [example]
>
> See the :ref:`topics-signals-ref` below to know which signals support Deferreds.

I’m not sure the example is very helpful:
- `getDeferredFromSomewhere()` makes me wonder where I can get a Deferred from, which is something that is not explained.
- I find the comment in the `return d` line confusing, and the other comment, above, meaningless (stating the obvious while users may not even understand how to define a callback).

An example based on a real use case, but simplified, may be better. See what @wRAR did at https://github.com/scrapy/scrapy/pull/4329/files#diff-8ff82039e3841d26573a59dfa988882eR53, for example. Maybe a similar example can be used handling one of the item signals. Although a different use case may be better. Maybe something as simple as using `treq` to send a request “out of Scrapy” to a fictional API?

In general, though, I thing these changes are OK; just mentioning “asynchronous” improves the current documentation.

In the long term, though, I think we may want to consider writing a topic of its own about running asynchronous code in Scrapy, which covers some use cases, useful libraries, different syntax (Scrapy 2.0.0 coroutines), etc. But that may be better for a different PR.

adityaa30: Thanks for the review. I will update the docs.

adityaa30: @Gallaecio I have updated the docs using a simpler example which uses `treq` to send scrapped item to the server.

docs/topics/signals.rst
```diff
@@ -52,9 +52,58 @@ Deferred signal handlers
...
 Some signals support returning :class:`~twisted.internet.defer.Deferred`
-objects from their handlers, see the :ref:`topics-signals-ref` below to know
-which ones.
+objects from their handlers, allowing you to run asynchronous code that
+does not block Scrapy. If a signal handler returns a
+:class:`~twisted.internet.defer.Deferred`, Scrapy waits for that
+:class:`~twisted.internet.defer.Deferred` to fire.
 
+Let's take an example:
+::
```
Gallaecio:
```suggestion
Let's take an example::
```

docs/topics/signals.rst
```diff
@@ -52,9 +52,58 @@ Deferred signal handlers
...
+
+        @classmethod
+        def from_crawler(cls, crawler, *args, **kwargs):
+            spider = super(SignalSpider, cls).from_crawler(crawler, *args, **kwargs)
+            crawler.signals.connect(spider.item_scrapped, signal=signals.item_scraped)
+            return spider
+
+        def item_scrapped(self, item):
+            # Send the scrapped item to the server
+            dfd = treq.post(
```
Gallaecio: I believe common practice is to call it `d` (but I see why you used `dfd`, it is used in related parts of Scrapy).

docs/topics/signals.rst
```diff
@@ -52,9 +52,58 @@ Deferred signal handlers
...
+                }).encode('ascii'),
+                headers={b'Content-Type': [b'application/json']}
+            )
+            dfd.addCallback(print_response)
+
+            # Next item will be scrapped only after dfd is fired
+            # As items are processed in parallel many dfd will
+            # be fired at the same time (refer settings['CONCURRENT_ITEMS'])
+            # hence, next batch of items will be processed after all
+            # dfd's are fired
```
Gallaecio:
```suggestion
            # The next item will be scraped only after the deferred (d)
            # is fired. Because items are processed in parallel by batches
            # (see the CONCURRENT_ITEMS setting), many deferreds
            # will be fired at the same time, but the next batch of items will
            # only start being processed after all deferreds have been fired.
```

docs/topics/signals.rst
```diff
@@ -52,9 +52,58 @@ Deferred signal handlers
...
+                    'author': quote.css('small.author::text').get(),
+                    'tags': quote.css('div.tags a.tag::text').getall(),
+                }
+            next_page = response.css('li.next a::attr(href)').get()
+            if next_page is not None:
+                next_page = response.urljoin(next_page)
+                yield scrapy.Request(next_page, callback=self.parse)
+
+See the :ref:`topics-signals-ref` below to know which signals support
+:class:`~twisted.internet.defer.Deferred`
```
Gallaecio:
```suggestion
:class:`~twisted.internet.defer.Deferred`.
```

docs/topics/signals.rst
```diff
@@ -159,6 +208,13 @@ item_error
     :param failure: the exception raised
     :type failure: twisted.python.failure.Failure
 
+.. note::
```
Gallaecio: Why is this note under `item_error`? This seems relevant for all 3 item signals.

Maybe we could add subsections under “Built-in signals reference” that group related signals together (e.g. “Engine signals”, “Item signals”…) and have the note under “Item signals”, before the signal-specific sections.

Also, I think it may be better to remove the similar note from the source code example, to keep it shorter, and only have it here. I don’t think the note is completely necessary to illustrate what the example is meant to illustrate.

docs/topics/signals.rst
adityaa30: Shall I change it to `d` or keep it `dfd` then?

adityaa30: Okay! I will change it to `d`

docs/topics/signals.rst
adityaa30: Thanks for your review. I will add the following subsections 
1. Engine signals
2. Item signals
3. Spider signals
4. Request signals
5. Response signals 

under "Built-in signals reference". 

About the note in the source code example, I feel now that it is redundant as you said and will remove it.

docs/topics/signals.rst
adityaa30: I have remove this note and updated to
```python
def item_scraped(self, item):
    # Send the scraped item to the server
    d = treq.post(
        'http://localhost:3000/add',
        json.dumps({
            'text': item['text'],
            'author': item['author'],
            'tags': item['tags'],
        }).encode('ascii'),
        headers={b'Content-Type': [b'application/json']}
    )
    d.addCallback(print_response)

    # The next item will be scraped only after
    # deferred (d) is fired
    return d
```

adityaa30:
@Gallaecio Thanks for your review. I have done the requested changes.

`test_spider_manager_verify_interface` is failing in TravisCI however when I run the tests in my laptop,  `test_spider_manager_verify_interface` passes. What could be the problem here?

Gallaecio:
> `test_spider_manager_verify_interface` is failing in TravisCI however when I run the tests in my laptop, `test_spider_manager_verify_interface` passes. What could be the problem here?

It is yet another (:man_facepalming:) unrelated issue: https://github.com/scrapy/scrapy/issues/4447

adityaa30:
Closed and reopened the pr to run the tests again

docs/topics/signals.rst
```diff
@@ -52,9 +51,54 @@ Deferred signal handlers
...
+            d = treq.post(
+                'http://localhost:3000/add',
+                json.dumps({
+                    'text': item['text'],
+                    'author': item['author'],
+                    'tags': item['tags'],
+                }).encode('ascii'),
+                headers={b'Content-Type': [b'application/json']}
+            )
+            d.addCallback(print_response)
```
elacuesta, line 82: This line might be confusing. `print_response` is not defined, but since it's not necessary for the example (as it's probably just to see the effects of the request) I'd say it'd be best to remove it.

adityaa30: @elacuesta Thanks for your review. I have removed the line in c9285431006dd03391dcadebf833db11124b61df

docs/topics/signals.rst
```diff
@@ -52,9 +51,53 @@ Deferred signal handlers
...
+            return d
+
+        def parse(self, response):
+            for quote in response.css('div.quote'):
+                yield {
+                    'text': quote.css('span.text::text').get(),
+                    'author': quote.css('small.author::text').get(),
+                    'tags': quote.css('div.tags a.tag::text').getall(),
+                }
+            next_page = response.css('li.next a::attr(href)').get()
```
elacuesta, line 94: I'd also remove this section in order to keep the example as simple as possible, navigation code is not relevant for the example.

elacuesta:
Looks good to me except for https://github.com/scrapy/scrapy/pull/4390#discussion_r403184861, sorry I didn't include that in my previous review.

docs/topics/signals.rst
adityaa30: @elacuesta Thanks for mentioning. I will remove this section

kmike:
@elacuesta are you happy with the PR?

PR #4392, opened 2020-02-28
fix flake E11X and E12X except E127, E128 and E129

noviluni:
This fixes:

E11X:

* E111: Indentation is not a multiple of four
* E114: Indentation is not a multiple of four (comment)
* E116: Unexpected indentation (comment)
* E117: Over-indented code blocks

Some E12X

* E121: Continuation line under-indented for hanging indent
* E122: Continuation line missing indentation or outdented
* E123: Closing bracket does not match indentation of opening bracket's line
* E124: Closing bracket does not match visual indentation
* E126: Continuation line over-indented for hanging indent

**Sorry for this big PR**. The E126 has several occurrences and is hard to review (and that's the reason why I decided to exclude E127, E128 and E129 from this PR).

It's a good chance to use the "viewed" checkbox when reviewing... :sweat_smile:

### DIFF
--- a/scrapy/core/downloader/contextfactory.py
+++ b/scrapy/core/downloader/contextfactory.py
@@ -86,8 +86,10 @@ class BrowserLikeContextFactory(ScrapyClientContextFactory):
         #
         # This means that a website like https://www.cacert.org will be rejected
         # by default, since CAcert.org CA certificate is seldom shipped.
-        return optionsForClientTLS(hostname.decode("ascii"),
-                                   trustRoot=platformTrust(),
-                                   extraCertificateOptions={
-                                        'method': self._ssl_method,
-                                   })
+        return optionsForClientTLS(
+            hostname.decode("ascii"),
+            trustRoot=platformTrust(),
+            extraCertificateOptions={
+                'method': self._ssl_method,
+            }
+        )
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -13,8 +13,8 @@ import logging
 
 from twisted.internet import defer
 from twisted.internet.error import TimeoutError, DNSLookupError, \
-        ConnectionRefusedError, ConnectionDone, ConnectError, \
-        ConnectionLost, TCPTimedOutError
+    ConnectionRefusedError, ConnectionDone, ConnectError, \
+    ConnectionLost, TCPTimedOutError
 from twisted.web.client import ResponseFailed
 
 from scrapy.exceptions import NotConfigured
--- a/tests/test_exporters.py
+++ b/tests/test_exporters.py
@@ -341,19 +341,20 @@ class XmlItemExporterTest(BaseItemExporterTest):
         i2 = dict(name=u'bar', age=i1)
         i3 = TestItem(name=u'buz', age=i2)
 
-        self.assertExportResult(i3,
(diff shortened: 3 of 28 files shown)


tests/test_exporters.py
```diff
@@ -341,19 +341,20 @@ def test_nested_item(self):
         i2 = dict(name=u'bar', age=i1)
         i3 = TestItem(name=u'buz', age=i2)
 
-        self.assertExportResult(i3,
+        self.assertExportResult(
+            i3,
             b'<?xml version="1.0" encoding="utf-8"?>\n'
             b'<items>'
-                b'<item>'
```
noviluni: any strong opinion here?

Not sure if you would prefer to put some of them in the same line or even exclude them from the flake8 rule.

elacuesta: I'd leave the indentation in this case, although it seems like an adjustment would be needed in the last occurrence to make it multiple of 4.

nyov: In the testcases I don't care as much about the style. I'm okay without the indentation since I won't usually have a reason to comprehend this xml structure.
But from a readability perspective, it'd make a lot more sense to me if you rewrote the longer single-line xml strings earlier on, instead of this already reasonably readable one. This again makes it look more like a change to appease a dumb script, instead of helping some developer to grasp the code better and faster.

Gallaecio: I would either leave it as is or switch to more complex code that does not compromise readability (e.g. defining the document as a single, multi-line, indented string and have it parsed and serialized without spaces, as desired here).

scrapy/core/downloader/contextfactory.py
```diff
@@ -86,8 +86,10 @@ def creatorForNetloc(self, hostname, port):
...
-                                   extraCertificateOptions={
-                                        'method': self._ssl_method,
-                                   })
+        return optionsForClientTLS(
+            hostname.decode("ascii"),
+            trustRoot=platformTrust(),
+            extraCertificateOptions={
+                'method': self._ssl_method,
+            }
+        )
```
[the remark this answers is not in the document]
Gallaecio: Flake8 must have reported the excessive indentation of  `'method': self._ssl_method,` (5 spaces). @noviluni must have taken the chance to switch to a different format. I see nothing wrong with either format; we could limit the change here to removing that extra space, but I don’t have a strong opinion either way.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -13,8 +13,8 @@
 
 from twisted.internet import defer
 from twisted.internet.error import TimeoutError, DNSLookupError, \
-        ConnectionRefusedError, ConnectionDone, ConnectError, \
-        ConnectionLost, TCPTimedOutError
+    ConnectionRefusedError, ConnectionDone, ConnectError, \
+    ConnectionLost, TCPTimedOutError
```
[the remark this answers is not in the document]
Gallaecio: No strong opinion either way either. As with the change in `scrapy/core/downloader/contextfactory.py`, it _may_ be better to stick where possible to fixing the specific issues reported by Flake8, and leave other style changes to other patches.

elacuesta:
Addressed by #4557, #4542, #4540, #4539, #4543

noviluni:
Thanks @elacuesta! 💯

PR #4395, opened 2020-03-02
Basic support for async def start_requests.

wRAR:
Again, no yield, only returning a list or iterator is supported.


wRAR:
Closed in favor of #4467

PR #4398, opened 2020-03-03
Remove deprecated Crawler.spiders property

nyov:
Two commits,
the 1st removing `Crawler.spiders`; deprecated since 4190266 (2014, Scrapy 0.25)
~the 2nd removing `CrawlerRunner.spiders`; deprecated since ad587ea (2015, Scrapy 0.25.1)~

~Adding `SPIDER_MANAGER_CLASS` to `scrapy/settings/deprecated.py` is somewhat too late, just for completeness' sake; so it's been there and can be removed again next commit?~

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -68,17 +68,6 @@ class Crawler:
         self.spider = None
         self.engine = None
 
-    @property
-    def spiders(self):
-        if not hasattr(self, '_spiders'):
-            warnings.warn("Crawler.spiders is deprecated, use "
-                          "CrawlerRunner.spider_loader or instantiate "
-                          "scrapy.spiderloader.SpiderLoader with your "
-                          "settings.",
-                          category=ScrapyDeprecationWarning, stacklevel=2)
-            self._spiders = _get_spider_loader(self.settings.frozencopy())
-        return self._spiders
-
     @defer.inlineCallbacks
     def crawl(self, *args, **kwargs):
         assert not self.crawling, "Crawling already taking place"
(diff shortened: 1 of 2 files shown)


kmike:
Hey @nyov! Are spider loader cleanups related to this issue? If no, how hard would it be to separate those into a different PR?

nyov:
Sure, no problem.
But do you mean the move of the `_get_spider_loader` function into a static method of `CrawlerRunner`?
Or the removal of the unused `SpiderLoader` class from the testcase?

nyov:
Ok I'll just put the second commit as another pull request.
Rebased.

kmike:
Hey @nyov! I meant the move of the _get_spider_loader function into a static method of CrawlerRunner. Is it needed?

Gallaecio:
It is certainly not _necessary_.

However, after the change I think the readability is slightly better: it is now obvious that `CrawlerRunner` is the only user of that function.

That said, if we expect the function to have more users in the long run, I guess it would be better to leave it as it was.

kmike:
I was making this comment more from a process perspective: merging a change which removes a deprecated "spider" property is a no-brainer, but if a PR also introducs such cleanup, it slows down the process. For example, 

* should this function be a classmethod or a staticmethod, 
* are there users of this function which we'll break after a move (and if so, should we do a deprecation); 
* some time is needed to check if it is a copy-paste, or something changed in the implementation
* some time is needed to figure out if it is actually better to have it as a function or as a method

It may turn up that in the end it is a good change to make, but hope this makes it more clear why I asked if a PR can be split.

nyov:
Sorry I wasn't fast enough with a reply. GH failed on me, and now you took it out of my hands :)

PR #4399, opened 2020-03-04
Internationalize the documentation

### LINKED ISSUE #3511: Translating the docs to Portuguese(BR)
Hi folks,

I've been talking to @lopuhin and we have decide to start a translation project for the docs to Portuguese so that we can increase the community here in Brazil.

At first he suggested me using the RTD itself (Read the docs), and he's also considering using transifex (https://www.transifex.com ).

I've read a little about both scenarios and I think we could start by creating multiple languages with ReadtheDocs once this is the first localization project.
 https://docs.readthedocs.io/en/latest/localization.html

Well, I'd like to know your opinion.

Right now I'm forking the project  to my own github account to start to understand how it works (RTD).
https://github.com/Maransatto/scrapy

I don't actually know how to merge it from my project to this one, or...
if the admins here decide to give me some permission I can create a branch for this.

Obs: I also have 3 friends here in Brazil that will help me to make the translation.

Gallaecio:
I needed the tutorial in Galician for a workshop I’m giving this weekend, so I though I’d take the chance to do this properly.

Changes:
- Short description of the internationalization and localization process.
- Addition of POT files and PO files for Galician, although only the tutorial page is translated.

Notes:
- Having the POT files under version control is common practice, even though they are generated files. More so for PO files, which we could remove from the repository if we ever move to a platform like Transifex, but common practice is to keep them in the same repository as the source code, with the same version tracking. Feel free to ignore their content when reviewing these changes, though.
- For the languages to actually show up in Read The Docs, we need to create separate projects for them. Because of this, and because it is also common practice, we should define some requirements that a translation has to meet before we will consider adding it to Read The Docs. For example, we could require 90% completion of the intro pages and 50% of the topics. My Galician translation, for example, should definitely _not_ warrant it being published in Read The Docs.
- While these changes should be enough for any experienced open source translator to translate the documentation, if we ever want to reach a wider translation audience we should consider Transifex, a kind of GitHub for translations (proprietary, yet popular among open source projects for being free, popular and effective).

Fixes  #3511


docs/conf.py
```diff
@@ -70,6 +70,8 @@
 # for a list of supported languages.
 language = 'en'
 
+gettext_compact = False
```
Gallaecio: This causes a `.pot` file to be generated per `.rst` file. Otherwise, there’s 1 `.pot` file per root documentation folder or `.rst` file (e.g. one for `intro`, one for `topics`).

I’m unsure about my choice here. I used it to make it easier for me to translate only the tutorial, but it may be better to leave the default value here and have fewer files.

Issue #4393, opened 2020-03-02
scrapy parse emits ANSI color sequences in the Windows terminal


A-hoy:
### Description
i try to use `scrapy parse` command in cmd（anaconda env），but when it logs Scraped Items and Requests, there are full of  garbled code which i show you below（Additional context）. I have no idea of it, but it seems to have nothing to do with character encoding,

### Steps to Reproduce
running command
```
>>> scrapy parse --spider=quotes3 http://quotes.toscrape.com/page/1/
```
the `quotes3` spider shows below（Additional context）

**Expected behavior:** without garbled code

**Actual behavior:** garbled code appears

**Reproduces how often:** every time

### Versions
Scrapy       : 1.8.0
lxml         : 4.4.1.0
libxml2      : 2.9.9
cssselect    : 1.1.0
parsel       : 1.5.2
w3lib        : 1.21.0
Twisted      : 19.10.0
Python       : 3.7.4 (default, Aug  9 2019, 18:22:51) [MSC v.1915 32 bit (Intel)]
pyOpenSSL    : 19.0.0 (OpenSSL 1.1.1d  10 Sep 2019)
cryptography : 2.7
Platform     : Windows-10-10.0.18362-SP0

### Additional context
spider.py
```
class QuotesSpider3(scrapy.Spider):
    name = 'quotes3'
    start_urls = ['http://quotes.toscrape.com/page/1/']

    def parse(self, response):
        for quote in response.xpath('//div[@class="quote"]'):
            yield {
                'text':
                quote.xpath('span[@class="text"]/text()').get(),
                'author':
                quote.xpath('.//small[@class="author"]/text()').get(),
                'tags':
                quote.xpath(
                    'div[@class="tags"]//a[@class="tag"]/text()').getall()
            }

        next_page = response.xpath('//li[@class="next"]//a/@href').get()
        if next_page is not None:
            next_page = response.urljoin(next_page)
            yield scrapy.Request(next_page, callback=self.parse)
```

```
(base) D:\scrapy_project>chcp
Active code page: 65001

(base) D:\scrapy_project>scrapy parse --spider=quotes3 http://quotes.toscrape.com/page/1/
2020-03-02 18:27:36 [scrapy.utils.log] INFO: Scrapy 1.8.0 started (bot: demo1)
[...]

PR #4403, opened 2020-03-04
enable ANSI color (instead of ANSI color codes) in the Windows terminal #4393

AKSHAYSHARMAJS:
modified scrapy/utils/display.py to stop ANSI color sequences in the Windows terminal (issue scrapy#4393)


AKSHAYSHARMAJS:
**Output**
`# Scraped Items  ------------------------------------------------------------
[{'author': 'Albert Einstein',
  'tags': ['change', 'deep-thoughts', 'thinking', 'world'],
  'text': '“The world as we have created it is a process of our thinking. It '
          'cannot be changed without changing our thinking.”'},
 {'author': 'J.K. Rowling',
  'tags': ['abilities', 'choices'],
  'text': '“It is our choices, Harry, that show what we truly are, far more '
          'than our abilities.”'},
 {'author': 'Albert Einstein',
  'tags': ['inspirational', 'life', 'live', 'miracle', 'miracles'],
  'text': '“There are only two ways to live your life. One is as though '
          'nothing is a miracle. The other is as though everything is a '
          'miracle.”'},
 {'author': 'Jane Austen',
  'tags': ['aliteracy', 'books', 'classic', 'humor'],
  'text': '“The person, be it gentleman or lady, who has not pleasure in a '
          'good novel, must be intolerably stupid.”'},
 {'author': 'Marilyn Monroe',
  'tags': ['be-yourself', 'inspirational'],
  'text': "“Imperfection is beauty, madness is genius and it's better to be "
          'absolutely ridiculous than absolutely boring.”'},
 {'author': 'Albert Einstein',
  'tags': ['adulthood', 'success', 'value'],
  'text': '“Try not to become a man of success. Rather become a man of '
          'value.”'},
 {'author': 'André Gide',
  'tags': ['life', 'love'],
  'text': '“It is better to be hated for what you are than to be loved for '
          'what you are not.”'},
 {'author': 'Thomas A. Edison',
  'tags': ['edison', 'failure', 'inspirational', 'paraphrased'],
  'text': "“I have not failed. I've just found 10,000 ways that won't work.”"},
 {'author': 'Eleanor Roosevelt',
  'tags': ['misattributed-eleanor-roosevelt'],
  'text': '“A woman is like a tea bag; you never know how strong it is until '
          "it's in hot water.”"},
 {'author': 'Steve Martin',
  'tags': ['humor', 'obvious', 'simile'],
  'text': '“A day without sunshine is like, you know, night.”'}]
# Requests  -----------------------------------------------------------------
[]
`

scrapy/utils/display.py
```diff
@@ -7,6 +7,8 @@
 
 
 def _colorize(text, colorize=True):
+    if sys.platform == "win32" and sys.stdout.isatty():
```
wRAR: Why is the isatty check needed here?

wRAR:
Ideally the colorization should stay enabled on the terminals that actually support it, such as the ones from WSL or MSYS, but I suspect we need to check terminfo for that which is a lot of work for no significant gain.

OTOH this PR doesn't allow enabling the colorization explicitly on Windows, maybe the check should be earlier.

AKSHAYSHARMAJS:
> Ideally the colorization should stay enabled on the terminals that actually support it, such as the ones from WSL or MSYS, but I suspect we need to check terminfo for that which is a lot of work for no significant gain.
I completely agree with this.

I just made a read for the problem and found out that Windows 10's console recently began interpreting escapes, but only with `ENABLE_VIRTUAL_TERMINAL_PROCESSING` which is off by default. So i tried enabling it and i think it works just fine.

wRAR:
I wonder if this new code works on pre-Windows 10 terminals? And again, what about non-cmd.exe terminals?

And I'd want some error handling there.

AKSHAYSHARMAJS:
> I wonder if this new code works on pre-Windows 10 terminals?

I just added a check for the windows 10 release and for specific version `10.0.14393` as the ANSI color codes problem only persist after this release of windows. So, there ll' be no problem for pre-windows 10 terminals.

AKSHAYSHARMAJS:
> what about non-cmd.exe terminals

For unix terminals in windows its working fine according to me, still you should check it once.
**WINDOWS CMD.EXE**

**CYGWIN TERMINAL (RUNNING ON WINDOWS)**

Gallaecio:
Just for reference, Django’s implementation: https://github.com/django/django/blob/master/django/core/management/color.py#L12

AKSHAYSHARMAJS:
@wRAR Is there anything that concerns you about the changes that I have made?
or what further changes can be done?

Gallaecio:
Django’s implementation seems less prone to issues to me, as it does not rely on specific Windows versions. Could you check if it also works for us?

wRAR:
@AKSHAYSHARMAJS my point about error handling is still valid.

AKSHAYSHARMAJS:
@wRAR @Gallaecio thanks for reply. I have added some error handling (used django's implementation for reference).
windows versions check is still there as windows has no real support for ANSI escape sequences.

scrapy/utils/display.py
```diff
@@ -3,24 +3,55 @@
...
+        from pygments.formatters import get_formatter_by_name
         from pygments import highlight
-        from pygments.formatters import TerminalFormatter
         from pygments.lexers import PythonLexer
-        return highlight(text, PythonLexer(), TerminalFormatter())
     except ImportError:
         return text
 
+    if sys.platform == "win32" and platform.release() == "10":
+        if platform.version() >= "10.0.14393":
```
Gallaecio: - Shouldn’t this be a single `if` statement?
- Is `platform.release() == "10"` needed, given `platform.version() >= "10.0.14393"`?
- Isn’t using `>=` for string comparison problematic here?
- I don’t think it is obvious from the code why `10.0.14393` is hardcoded there. Maybe we need a comment explaining it?

AKSHAYSHARMAJS: > * Shouldn’t this be a single `if` statement?
> 
> * Is `platform.release() == "10"` needed, given `platform.version() >= "10.0.14393"`?
yes, it should be single `if` statement and there is no need for `platform.release( ) == "10"`

>     * Isn’t using `>=` for string comparison problematic here? 
It is problematic. So, now I have used `packaging.version.parse()` for version comparison.

scrapy/utils/display.py
```diff
@@ -3,24 +3,62 @@
 """
 
 import sys
+from platform import version
+from packaging.version import parse
```
Gallaecio: :lipstick: A-Z
```suggestion
from packaging.version import parse
from platform import version
```

scrapy/utils/display.py
```diff
@@ -3,24 +3,62 @@
...
     except ImportError:
         return text
 
+    """
+    All Windows versions >= "10.0.14393" interpret ANSI escape sequences
+    using terminal processing.
+
+    Enable enivornment variable `ENABLE_VIRTUAL_TERMINAL_PROCESSING`
+    to activate terminal processing.
+    """
```
Gallaecio: :lipstick: Docstrings are not meant to be used in the middle of a function implementation.

```suggestion
    # All Windows versions >= "10.0.14393" interpret ANSI escape sequences
    # using terminal processing.
    #
    # Enable enivornment variable `ENABLE_VIRTUAL_TERMINAL_PROCESSING`
    # to activate terminal processing.
```

scrapy/utils/display.py
```diff
@@ -3,24 +3,62 @@
...
+        import ctypes
+        kernel32 = ctypes.windll.kernel32
+        kernel32.SetConsoleMode(kernel32.GetStdHandle(-11), 7)
+        # set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag
+    colors = color_support_info()
+    if colors == 256:
+        format_options = {'style': 'default'}
+    else:
+        format_options = {'bg': 'dark'}
+    format_alias = 'terminal256' if colors == 256 else 'terminal'
```
Gallaecio: :lipstick:
```suggestion
    if color_support_info() == 256:
        format_alias = 'terminal256'
        format_options = {'style': 'default'}
    else:
        format_alias = 'terminal'
        format_options = {'bg': 'dark'}
```

scrapy/utils/display.py
```diff
@@ -3,24 +3,62 @@
...
 
 def pformat(obj, *args, **kwargs):
     return _colorize(pformat_(obj), kwargs.pop('colorize', True))
 
 
 def pprint(obj, *args, **kwargs):
     print(pformat(obj, *args, **kwargs))
+
+
+def color_support_info():
```
Gallaecio:
```suggestion
def _color_support_info():
```

And maybe we should move it right before the function that uses it, above.

AKSHAYSHARMAJS:
Thanks @Gallaecio for the feedback and suggested changes. I 'll keep in mind all lipstick aesthetic
for later contributions.

scrapy/utils/display.py
```diff
@@ -3,20 +3,56 @@
 """
 
 import sys
+from packaging.version import parse
```
wRAR: packaging is not an stdlib module, is it?

scrapy/utils/display.py
```diff
@@ -3,20 +3,56 @@
...
 
+    # All Windows versions >= "10.0.14393" interpret ANSI escape sequences
+    # using terminal processing.
+    #
+    # Enable enivornment variable `ENABLE_VIRTUAL_TERMINAL_PROCESSING`
+    # to activate terminal processing.
+    if sys.platform == "win32" and parse(version()) >= parse("10.0.14393"):
+        import ctypes
+        kernel32 = ctypes.windll.kernel32
+        kernel32.SetConsoleMode(kernel32.GetStdHandle(-11), 7)
```
wRAR: Still no error handling.

scrapy/utils/display.py
```diff
@@ -3,20 +3,56 @@
...
+        # set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag
```
wRAR: Shouldn't this be before the previous line?

scrapy/utils/display.py
AKSHAYSHARMAJS: no, it is a third party utility used by setuptools.

scrapy/utils/display.py
AKSHAYSHARMAJS: yes, I have changed that.

scrapy/utils/display.py
AKSHAYSHARMAJS: @wRAR added some error handling there. Please check if its correct.

scrapy/utils/display.py
AKSHAYSHARMAJS: I think we can use `distutils.version.LooseVersion()` or `distutils.version.StrictVersion()` for it but most users must have installed setuptools.

scrapy/utils/display.py
```diff
@@ -3,20 +3,63 @@
...
+            import ctypes
+            from ctypes import FormatError
+            kernel32 = ctypes.windll.kernel32
+            handle = kernel32.GetStdHandle(-11)
+            # set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag
+            if not kernel32.SetConsoleMode(handle, 7):
+                raise ValueError
+        except ValueError:
+            sys.stderr.write(FormatError(kernel32.GetLastError()))
+            return text
```
AKSHAYSHARMAJS, line 38: I am not sure if there is need of error message `sys.stderr.write(FormatError(kernel32.GetLastError()))`, Maybe we can just return the plain text without the error message. 
@wRAR please guide me here.

scrapy/utils/display.py
AKSHAYSHARMAJS: @Gallaecio Does scrapy install ``setuptools`` package with other reqirements during installation?

Gallaecio: I don’t think so, you can check in `setup.py`. However, can’t we just [use a built-in alternative](https://stackoverflow.com/a/11887885/939364)?

AKSHAYSHARMAJS: @Gallaecio thanks for the suggestion. I have changed version comparison with inbuilt function `distutils.version.LooseVersion()`.

scrapy/utils/display.py
AKSHAYSHARMAJS: @Gallaecio Can you please tell me if we need error message 'sys.stderr.write(FormatError(kernel32.GetLastError()))' as the text "The handle is invalid" doesn't look aesthetically good(like in below screenshot)

)
I think we should just return the plain text when exception is raised.

Gallaecio: I agree, we should just return the text.

AKSHAYSHARMAJS:
@Gallaecio @wRAR any suggestions on changes so far : )

AKSHAYSHARMAJS:
@Gallaecio @wRAR any suggestions on changes so far : )

scrapy/utils/display.py
```diff
@@ -3,20 +3,61 @@
 """
 
 import sys
+from distutils.version import LooseVersion as check
```
Gallaecio, line 6: `check` seems rather ambiguous, maybe `parse_version` or similar?

AKSHAYSHARMAJS: Yes, I changed it to `parse` .
Thanks!

Gallaecio: `parse` still seems rather generic :smile:

AKSHAYSHARMAJS: @Gallaecio 😄You re' right.
but to use `parse_version`, I have to use two if statements(like below) as single if statement is failing one flake test of crossing 79 characters limit in a single line.
```
    if sys.platform == "win32":
        if parse_version(version()) >= parse_version("10.0.14393"):
```

elacuesta: We are not currently enforcing the 79-chars limit. In fact, we're working towards a higher limit (#4237). For the time being, please add this file to `pytest.ini` with the code `E501`.

AKSHAYSHARMAJS: Alright! single `if` statement will be good then.
@elacuesta thanks for the info : )

AKSHAYSHARMAJS: >  For the time being, please add this file to `pytest.ini` with the code `E501`.
Done 👍

AKSHAYSHARMAJS:
`pytest.ini` is conflicting again?

Gallaecio:
Yes, @elacuesta is making some great progress on code cleanup.

elacuesta:
Sorry about that. Implementation looks good AFAICT, but I don't also have access to a Windows machine and we're currently not testing with it (see #4458). Still, it'd be good to test that nothing breaks in the currently tested platform. From what I can see, this is only used in the `parse` command, but their tests read from stdout/stderr which unless I'm mistaken would be read as plain text, without color information. Do you think you could add some small test for the `_colorize` or `pformat` functions? Something like `assert _colorize('{"a": 1}') == '{\x1b[33m"\x1b[39;49;00m\x1b[33ma\x1b[39;49;00m\x1b[33m"\x1b[39;49;00m: \x1b[34m1\x1b[39;49;00m}\n'`

AKSHAYSHARMAJS:
Okay, I will try to add some small tests for the changes.

AKSHAYSHARMAJS:
I don't know how will I check default ansi escape sequence for different OS as I tried in linux and windows both are providing different colors in their terminal. 
Is there a way to get default escape sequence for current OS?

AKSHAYSHARMAJS:
I don't know how can I get the default escape sequence for different OS as I tried running the changes in linux and Windows both are providing different colors in terminal.
Is there a way to get default escape sequences current OS is using?

AKSHAYSHARMAJS:
okay thanks

AKSHAYSHARMAJS:
tests are failing as testing file is not running in a real terminal and that is why `sys.stdout.isatty()`condition is failing and returning only plaintext.
@Gallaecio  @elacuesta please help here?

Gallaecio:
Maybe you could use [mock](https://docs.python.org/3/library/unittest.mock.html) to trick the code into thinking it is in a terminal.

AKSHAYSHARMAJS:
`
class TestDisplay(TestCase): 

    @mock.patch('sys.stdout.isatty', autospec=True)
    def test_remove_color_no_tty(self, mock_isatty):
        mock_isatty.return_value = True
        self.assertEqual(_colorize('{"a": 1}'),'{\x1b[33m"\x1b[39;49;00m\x1b[33ma\x1b[39;49;00m\x1b[33m"\x1b[39;49;00m: \x1b[34m1\x1b[39;49;00m}\n')
`
above test is still failing as _colorize is returning only a plaintext. what wrong am i doing here?

elacuesta:
Not sure, but I suspect you might be entering in one of the `except` blocks that `return text`. I just pushed a simple test case, I made it dependent on `_color_support_info` because my regular terminal has `_color_support_info==256` but it seems like the `tox` environment has only 16 colors.

AKSHAYSHARMAJS:
Thanks @elacuesta for the changes.

> Not sure, but I suspect you might be entering in one of the `except` blocks that `return text`. I just

I figured out that `GetStdhandle()` is giving wrong handle during test I don't know why?. I tried passing actual handle to `SetConsoleMode()` then test is passing fine.

AKSHAYSHARMAJS:
image
Just for reference, maybe this [Issue](https://github.com/tox-dev/tox/issues/1441) and this [PR](https://github.com/tox-dev/tox/issues/1441) can help with above error.

Gallaecio:
The tests use the latest version of Tox, which I believe includes that PR.

AKSHAYSHARMAJS:
> I figured out that `GetStdhandle()` is giving wrong handle during test I don't know why?. I tried      > passing actual handle to `SetConsoleMode()` then test is passing fine.
Can you please tell me why this is happening? `GetStdhandle()` retrieving wrong STDOUT handle.

Gallaecio:
I see the Django implementation [I mentioned earlier](https://github.com/scrapy/scrapy/pull/4403#issuecomment-598118579) does nor rely on curses. What made you take the `curses` approach, did you see a similar approach somewhere else?

AKSHAYSHARMAJS:
yes, I saw a similar approach that uses curses for terminfo. Django implementation was not giving any colors when the platform is windows.

AKSHAYSHARMAJS:
I think maybe we should just remove check for terminfo and just check if current terminal supports color or not using `curses.has_colors()` and use normal terminal formatter for all platforms as we are
using default colors and not taking user input for specific colors.

Gallaecio:
Sounds simpler, I think it’s worth giving it a try.

AKSHAYSHARMAJS:
@Gallaecio @elacuesta Please go through the changes. I have used only normal terminal formatter for all platforms and added a separate test for windows by mocking `ctypes` in it so as to get the valid handle for stdout.

AKSHAYSHARMAJS:
> 
> 
> Thanks @elacuesta for the changes.
> 
> > Not sure, but I suspect you might be entering in one of the `except` blocks that `return text`. I just
> 
> I figured out that `GetStdhandle()` is giving wrong handle during test I don't know why?. I tried passing actual handle to `SetConsoleMode()` then test is passing fine.

Now, this test is passing fine as `GetStdHandle()` giving valid handle.

Gallaecio:
Could you work on 100% coverage of your changes? (see https://codecov.io/gh/scrapy/scrapy/compare/c86a1035dd9b8b10acaf8f9e8bdb1b5494a287e2...91bf06c48724a844886d1514e9c49b0ff3bb5a68/diff)

scrapy/utils/display.py
```diff
@@ -3,20 +3,60 @@
...
+    if sys.platform == "win32" and parse_version(version()) >= parse_version("10.0.14393"):
+        try:
+            import ctypes
+            kernel32 = ctypes.windll.kernel32
+            handle = kernel32.GetStdHandle(-11)
+            # set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag
+            if not kernel32.SetConsoleMode(handle, 7):
+                raise ValueError
+        except ValueError:
+            return text
```
AKSHAYSHARMAJS, line 35: @Gallaecio  please suggest how should I cover this part. I tried mocking windows platform but I think `parse_version` is still failing.

Gallaecio: Could you please push the test code that fails, so that I can have a look?

AKSHAYSHARMAJS: I have pushed the test code.Please have a look, windows mocking part is failing giving this error `AttributeError: <module 'ctypes' from '/usr/lib/python3.6/ctypes/__init__.py'> does not have the attribute 'windll'
`

Gallaecio:
Please, have a look at https://github.com/AKSHAYSHARMAJS/scrapy/pull/1

It’s a refactoring proposal, where separate methods are used in the color-handling implementation for easier testing and better code readability.

Note that I removed the `curses` stuff, because I still don’t understand why it is needed, and fear it may result in terminals supporting colors not displaying them just because `curses` is not installed. But I may be wrong here.

AKSHAYSHARMAJS:
@Gallaecio thanks for the changes, looks perfect and precise to me. I see removing `curses` is completely fine and also helping in testing.
I don't know if we are fine with windows versions < `10.0.14393` with no colors?

Gallaecio:
> I don't know if we are fine with windows versions < 10.0.14393 with no colors?

I thought those Windows versions would not support color.

Is that what curses was for? Do older Windows versions correctly parse ANSI color codes if curses is installed?

If so, maybe it makes sense to add back the curses code (but only for those old Windows versions?).

—

On a separate note, my refactoring broke the Flake8 checks, as well as Python 3.5 tests through the use of f-strings (not the first time I forgot Scrapy cannot use them yet :facepalm: ).

AKSHAYSHARMAJS:
`curses` was only for different formatters for different platform. but now we are only using normal formatter so there is no use of it.

> Is that what curses was for? Do older Windows versions correctly parse ANSI color codes if curses is installed?

Older versions of windows interpret ansi escape codes as terminal processing is enabled by default and in newer versions we have to enabled it by default.

AKSHAYSHARMAJS:
> On a separate note, my refactoring broke the Flake8 checks, as well as Python 3.5 tests through the use of f-strings (not the first time I forgot Scrapy cannot use them yet 🤦 ).

:smile: no problem I will fix it.

AKSHAYSHARMAJS:
> 
> 
> > On a separate note, my refactoring broke the Flake8 checks, as well as Python 3.5 tests through the use of f-strings (not the first time I forgot Scrapy cannot use them yet 🤦 ).
> 
> :smile:😄 no problem I will fix it.

AKSHAYSHARMAJS:
> 
> 
> `curses` was only for different formatters for different platform. but now we are only using normal formatter so there is no use of it.
> 
> > Is that what curses was for? Do older Windows versions correctly parse ANSI color codes if curses is installed?
> 
> Older versions of windows also interpret ansi escape codes as terminal processing is enabled by default and in newer versions we have to enabled it by default.

Gallaecio:
> Older versions of windows also interpret ansi escape codes as terminal processing is enabled by default and in newer versions we have to enable it separately.

Oh, I had no idea. I see you’ve already fixed the logic accordingly, so :+1:.

AKSHAYSHARMAJS:
Yes I have fixed it accordingly. :smiley:
Thanks @Gallaecio

AKSHAYSHARMAJS:
Please suggest, what else can be done here??

wRAR:
Thank you!

AKSHAYSHARMAJS:
I didn't get contribution credits for all the commits made in this PR. am I doing something wrong?

PR #4404, opened 2020-03-04
Remove obsolete DEPRECATED_SETTINGS (deprecated.py)

nyov:
With the removal of the latest two additions to `DEPRECATED_SETTINGS` (in 1b591ff06), everything else now remaining in `deprecated.py` is even older: From c7f821990 and b46b5a6ef back in 2012.

It seems no other deprecated settings have ever been added after 2015, and it seems they rather happen to be deprecated elsewhere in the code; making this file obsolete (which was only checked from `scrapy/cmdline.py`).
Perhaps it could become useful again in the future, but should we keep it around or reinstate it then?


nyov:
For consideration: #4407 wants to add the `SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE` setting here.

scrapy/settings/deprecated.py
```diff
@@ -1,23 +0,0 @@
...
-from scrapy.exceptions import ScrapyDeprecationWarning
-
-DEPRECATED_SETTINGS = [
-    ('TRACK_REFS', 'no longer needed (trackref is always enabled)'),
-    ('RESPONSE_CLASSES', 'no longer supported'),
-    ('DEFAULT_RESPONSE_ENCODING', 'no longer supported'),
-    ('BOT_VERSION', 'no longer used (user agent defaults to Scrapy now)'),
-    ('ENCODING_ALIASES', 'no longer needed (encoding discovery uses w3lib now)'),
-    ('STATS_ENABLED', 'no longer supported (change STATS_CLASS instead)'),
-    ('SQLITE_DB', 'no longer supported'),
```
nyov: https://github.com/scrapy/scrapy/commit/c7f821990117dc69f5c843e6bc1f2b0859c2f015#diff-183ad3e1c82fc47e1813eaf86e32293b

scrapy/settings/deprecated.py
```diff
@@ -1,23 +0,0 @@
...
-    ('AUTOTHROTTLE_MIN_DOWNLOAD_DELAY', 'use DOWNLOAD_DELAY instead'),
-    ('AUTOTHROTTLE_MAX_CONCURRENCY', 'use CONCURRENT_REQUESTS_PER_DOMAIN instead'),
```
nyov: https://github.com/scrapy/scrapy/commit/b46b5a6ef04b1f0a2043191147d544db637281a6#diff-183ad3e1c82fc47e1813eaf86e32293b

Gallaecio:
I think it makes sense to remove it. I don’t think it accounts for `Spider.custom_settings`, so when we need something similar in the future we will probably implement it differently anyway.

PR #4406, opened 2020-03-06
[fix] wrong reactor installed error

### LINKED ISSUE #4401: Easy to break installing the custom reactor
I created a spider that imports `scrapy.mail` and got "wrong reactor installed" after setting `TWISTED_REACTOR`. This is because `scrapy.mail` imports `twisted.internet.reactor` at the top level and because `scrapy crawl` imports all spider modules.

So I think we should kill all top-level `twisted.internet.reactor` imports (it was partially done when adding the feature, so that importing `CrawlerProcess` doesn't install the reactor) and add documentation that the user code shouldn't do similar top-level imports.

adityaa30:
- Error occured when TWISTED_REACTOR in settings.py was set
- remove all top-level imports for twisted.internet.reactor
- add example, reason for top-level imports in TWISTED_REACTOR in `docs/settings.rst`

Fixes #4401

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1473,7 +1473,67 @@ If a reactor is already installed,
 
 :meth:`CrawlerRunner.__init__ <scrapy.crawler.CrawlerRunner.__init__>` raises
 :exc:`Exception` if the installed reactor does not match the
-:setting:`TWISTED_REACTOR` setting.
+:setting:`TWISTED_REACTOR` setting. Top-level
+:mod:`~twisted.internet.reactor` imports in project files & imported
+3rd party libraries may also raise :exc:`Exception` as Twisted will install the
+default reactor before scrapy installs reactor specified in
+:setting:`TWISTED_REACTOR` leading to a mismatch in installed reactor.
+
+In order to use the reactor installed by scrapy::
+
+    import scrapy
+    from twisted.internet import reactor
+
+
+    class QuotesSpider(scrapy.Spider):
+        name = 'quotes'
+
+        def __init__(self, *args, **kwargs):
+            self.timeout = int(kwargs.pop('timeout', '60'))
+            super(QuotesSpider, self).__init__(*args, **kwargs)
+
+        def start_requests(self):
+            reactor.callLater(self.timeout, self.stop)
+
+            urls = ['http://quotes.toscrape.com/page/1']
+            for url in urls:
+                yield scrapy.Request(url=url, callback=self.parse)
+
+        def parse(self, response):
+            for quote in response.css('div.quote'):
+                yield {'text': quote.css('span.text::text').get()}
+
+        def stop(self):
+            self.crawler.engine.close_spider(self, 'timeout')
+
+
+which raises :exc:`Exception`, becomes::
+
+    import scrapy
+
+
+    class QuotesSpider(scrapy.Spider):
+        name = 'quotes'
+
+        def __init__(self, *args, **kwargs):
+            self.timeout = int(kwargs.pop('timeout', '60'))
+            super(QuotesSpider, self).__init__(*args, **kwargs)
+
+        def start_requests(self):
+            from twisted.internet import reactor
+            reactor.callLater(self.timeout, self.stop)
+
+            urls = ['http://quotes.toscrape.com/page/1']
+            for url in urls:
+                yield scrapy.Request(url=url, callback=self.parse)
+
+        def parse(self, response):
+            for quote in response.css('div.quote'):
+                yield {'text': quote.css('span.text::text').get()}
+
+        def stop(self):
+            self.crawler.engine.close_spider(self, 'timeout')
+
 
 The default value of the :setting:`TWISTED_REACTOR` setting is ``None``, which
 means that Scrapy will not attempt to install any specific reactor, and the
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1473,11 +1473,10 @@ If a reactor is already installed,
 
 :meth:`CrawlerRunner.__init__ <scrapy.crawler.CrawlerRunner.__init__>` raises
 :exc:`Exception` if the installed reactor does not match the
-:setting:`TWISTED_REACTOR` setting. Top-level
+:setting:`TWISTED_REACTOR` setting; therfore, having top-level
 :mod:`~twisted.internet.reactor` imports in project files & imported
-3rd party libraries may also raise :exc:`Exception` as Twisted will install the
-default reactor before scrapy installs reactor specified in
-:setting:`TWISTED_REACTOR` leading to a mismatch in installed reactor.
+3rd party libraries will make Scrapy raise :exc:`Exception` when
+it checks which reactor is installed.
 
 In order to use the reactor installed by scrapy::
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1474,11 +1474,11 @@ If a reactor is already installed,
 :meth:`CrawlerRunner.__init__ <scrapy.crawler.CrawlerRunner.__init__>` raises
 :exc:`Exception` if the installed reactor does not match the
 :setting:`TWISTED_REACTOR` setting; therfore, having top-level
-:mod:`~twisted.internet.reactor` imports in project files & imported
-3rd party libraries will make Scrapy raise :exc:`Exception` when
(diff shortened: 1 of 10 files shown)


wRAR:
I think you've missed several top-level imports, for example one in https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/__init__.py

How did you search for them?

scrapy/mail.py
```diff
@@ -112,6 +112,7 @@ def _sent_failed(self, failure, to, cc, subject, nattachs):
     def _sendmail(self, to_addrs, msg):
         # Import twisted.mail here because it is not available in python3
         from twisted.mail.smtp import ESMTPSenderFactory
+        from twisted.internet import reactor
```
wRAR: You can reorder these so that they are sorted :)

docs/topics/settings.rst
```diff
@@ -1473,7 +1473,67 @@ If a reactor is already installed,
 
 :meth:`CrawlerRunner.__init__ <scrapy.crawler.CrawlerRunner.__init__>` raises
 :exc:`Exception` if the installed reactor does not match the
-:setting:`TWISTED_REACTOR` setting.
+:setting:`TWISTED_REACTOR` setting. Top-level
+:mod:`~twisted.internet.reactor` imports in project files & imported
+3rd party libraries may also raise :exc:`Exception` as Twisted will install the
+default reactor before scrapy installs reactor specified in
+:setting:`TWISTED_REACTOR` leading to a mismatch in installed reactor.
```
wRAR: Those top-level imports won't raise an exception, Scrapy will, when it checks which reactor is installed, so this can be rephrased.

docs/topics/settings.rst
```diff
@@ -1473,7 +1473,67 @@ If a reactor is already installed,
...
+            for url in urls:
+                yield scrapy.Request(url=url, callback=self.parse)
+
+        def parse(self, response):
+            for quote in response.css('div.quote'):
+                yield {'text': quote.css('span.text::text').get()}
+
+        def stop(self):
+            self.crawler.engine.close_spider(self, 'timeout')
+
```
wRAR: I think the example is great.

scrapy/mail.py
adityaa30: Okay

adityaa30:
> I think you've missed several top-level imports, for example one in https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/__init__.py
> 
> How did you search for them?

My bad. I will fix this now. I made a mistake in my search earlier. Now I am searching for `from twisted.internet import` in all the files using the VS Code's global search feature. 

 Also, I am a bit confused about some of the import such as those in
- [benchserver.py](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/benchserver.py#L37-L45)
- [qps-bench-server.py](https://github.com/scrapy/scrapy/blob/master/extras/qps-bench-server.py#L56-L59)
- [testsite.py](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/testsite.py#L40-L43)

as these are used either at top level or inside a `__main__` scope. What to do with these imports?

Should I move top-level imports in tests ([test_downloadermiddleware_robotstxt.py](https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware_robotstxt.py#L4), [test_engine.py](https://github.com/scrapy/scrapy/blob/master/tests/test_engine.py#L18), [test_pipeline_media.py](https://github.com/scrapy/scrapy/blob/master/tests/test_pipeline_media.py#L4), [test_utils_defer.py](https://github.com/scrapy/scrapy/blob/master/tests/test_utils_defer.py#L2), [test_utils_signal.py](https://github.com/scrapy/scrapy/blob/master/tests/test_utils_signal.py#L7), [test_webclient.py](https://github.com/scrapy/scrapy/blob/master/tests/test_webclient.py#L11)) as well?

docs/topics/settings.rst
adityaa30: Okay I will change it to:
`CrawlerRunner` raises `Exception` if the installed reactor does not match the `TWISTED_REACTOR` setting; therefore, having top-level `twisted.internet.reactor` imports in project files & imported 3rd party libraries will make Scrapy raise `Exception` when it checks which reactor is installed.

docs/topics/settings.rst
adityaa30: Thanks a lot :)

adityaa30:
@wRAR Also, @nyov suggested an idea to "replace the twisted.internet.reactor function with a custom lazy-loading reactor-loading function instead". I think this is feasible and mentioned a [solution](https://github.com/scrapy/scrapy/issues/4401#issuecomment-597036185) using `importlib`. This will reduce the `twisted.internet.reactor` imports and also make it declared at top-level in scrapy codebase.

wRAR:
> Also, I am a bit confused about some of the import such as those in

`benchserver.py` is only used as a separate process. Not sure what does `extras/qps-bench-server.py` do as it's not used by the Scrapy code. OTOH `scrapy.utils.testsite` is imported by several tests so it makes sense to fix it as well, even if it's usable only in tests.

> Should I move top-level imports in tests

The reactor in tests is currently installed early, via the pytest-twisted plugin, so I think there is currently no need to do it.

> an idea to "replace the twisted.internet.reactor function with a custom lazy-loading reactor-loading function instead"

I don't currently see benefits of doing this, especially as it won't directly help with user or 3rd party code, but it's nice to have this as an option in case we need it.

adityaa30:
@wRAR Thanks for you reply. I have refactored `benchserver.py` and `testsite.py`. I feel all top-level imports are refactored now (except those in `tests/*` and `extras/qps-bench-server.py`). 

 > but it's nice to have this as an option in case we need it

Should I add the `LazyLoader` class to the codebase? If yes, where will the best place to add it?

Gallaecio:
@adityaa30 Since this is something we would like to fix rather soon, and to keep things simple, I think it’s best not to address the lazy loading in this issue.

We can address it later in a different issue if we see a use for it. A `get_reactor()` function or similar that imports and returns the reactor should be trivial to implement somewhere under `scrapy.utils.*` for anyone to use.

docs/topics/settings.rst
```diff
@@ -1473,7 +1473,66 @@ If a reactor is already installed,
 
 :meth:`CrawlerRunner.__init__ <scrapy.crawler.CrawlerRunner.__init__>` raises
 :exc:`Exception` if the installed reactor does not match the
-:setting:`TWISTED_REACTOR` setting.
+:setting:`TWISTED_REACTOR` setting; therfore, having top-level
+:mod:`~twisted.internet.reactor` imports in project files & imported
```
Gallaecio:
```suggestion
:mod:`~twisted.internet.reactor` imports in project files and imported
```

docs/topics/settings.rst
```diff
@@ -1473,7 +1473,66 @@ If a reactor is already installed,
...
+3rd party libraries will make Scrapy raise :exc:`Exception` when
```
Gallaecio:
```suggestion
third-party libraries will make Scrapy raise :exc:`Exception` when
```

docs/topics/settings.rst
```diff
@@ -1473,7 +1473,66 @@ If a reactor is already installed,
...
+it checks which reactor is installed.
+
+In order to use the reactor installed by scrapy::
```
Gallaecio:
```suggestion
In order to use the reactor installed by Scrapy::
```

adityaa30:
> @adityaa30 Since this is something we would like to fix rather soon, and to keep things simple, I think it’s best not to address the lazy loading in this issue.

Sure. 
@Gallaecio Thanks for your review. I will update this pr with requested changes.

Gallaecio:
It looks like https://github.com/scrapy/scrapy/pull/4406/commits/8bbc4453bf9a46047f1a7b9bf7906ccfabf52a85 broke the tests.

adityaa30:
@wRAR @Gallaecio I am not really sure what happened here. TravisCI passed the tests at 8bbc4453bf9a46047f1a7b9bf7906ccfabf52a85 ([Link to result](https://travis-ci.org/github/scrapy/scrapy/builds/661597131?utm_source=github_status&utm_medium=notification)) but after 762a7e3287339a6f5952d74277dfb1b67d97a6fe the tests have started to fail. What could be the possible reason?

Gallaecio:
It is not your code, it looks like it is happening in `master`.

Gallaecio:
Thanks!

PR #4407, opened 2020-03-06
[Fix] Added check for deprecated environment variables

### LINKED ISSUE #4379: Deprecate environment variables being loaded as settings
Scrapy reads the following environment variables into settings named after them without the `SCRAPY_` prefix:

- `SCRAPY_CHECK`
- `SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE`
- `SCRAPY_PROJECT`
- `SCRAPY_PYTHON_SHELL`
- `SCRAPY_SETTINGS_MODULE`

This is an unintended behavior, due to the (deprecated) feature that allows defining settings through environment variables not filtering out these environment variables that are not meant to generate settings.

I’ve seen cases where some of the resulting settings are used (`settings.get('SETTINGS_MODULE')`), so we need to deprecate this usage first, before we can eventually stop loading these environment variables into settings.

It may be tricky to do properly, because we need to avoid issuing a deprecation warning if same-name settings are defined without an environment variable. So, for example, if a project is using `settings.get('SETTINGS_MODULE')` but that project defines such a setting, we should not warn about it.

See #4375 for more context.

joybh98:
As discussed in #4379, I have added checks in `__init__.py` to check if the environment variables are same as the settings variables.

I've also added `SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE'` to `deprecated.py`

I was still not clear about what exactly I should've done, so my approach may be wrong. 

I have ran the tests using **tox**

Fixes #4379


joybhallaa:
@nyov @Gallaecio please take a look at this, would like your input

joybhallaa:
@nyov removing `deprecated.py` is not as important to my PR, I added `SCRAPY_PICKLED_SETTINGS_TO_OVERRIDE` to `deprecated.py` as I felt that it should be kept as updated as possible

joybhallaa:
@nyov you're right, I will pick up another issue till I get some more clarification, thanks for taking out the time to review it !

joybhallaa:
@Gallaecio if you're free please check this PR out.

joybhallaa:
@nyov I'll go trh

joybhallaa:
@nyov I'll go through our discussion again and make changes, I don't want to leave this issue hanging.

I closed this PR by mistake

Gallaecio:
I don’t think we need to touch `deprecated.py`. That is for notifying users of settings that they have set and are deprecated. A fix for #4379 should _not_ warn about any of the corresponding settings if it is defined by the user, we should only warn when it is not _defined_ by the user by is however _read_ by the user.

About `check_envr_var`, @nyov is right. I’m guessing you have not figured out yet an entry point where you could place the code that performs the checks. I think we could implement such code in the `Settings` class.

I haven’t put much though into it, but I’m thinking that, since in `Settings` we control the writes and reads of settings, we could do something in the lines of:

1.  Make it so that when those settings are written from environment variables by Scrapy, an additional, hidden (not intended for user usage) parameter is passed on the setter to indicate the origin of those settings. 

    If they are set later from a different point in the code (e.g. loaded from `settings.py` or `Spider.custom_settings`), not getting that hidden parameter would tell the `Settings` class that it should not warn about the corresponding setting, as it is being set from somewhere other than the environment.

    We may also need to handle the scenario where the settings have already been defined from somewhere else before they are loaded from an environment variable. _If_ that could happen, we should probably warn only if the value from the environment variable is different from the previous value.

2.  Once `Settings` knows which of those variables warrant issuing a warning based on how they were loaded, we could use the getter of `Settings` to issue a warning when one of those settings is read.

    Also, the warning should probably only be issued the first time each setting is read.

I’m hoping this makes sense.

joybhallaa:
@Gallaecio thank you, I've got a little bit of clarity. Will work on implementing these things.

joybhallaa:
> Make it so that when those settings are written from environment variables by Scrapy, an additional, hidden (not intended for user usage) parameter is passed on the setter to indicate the origin of those settings.

I was thinking of doing something like this 

Screenshot_20200314_160932

(Lines 446 to 450)
@Gallaecio thoughts?

Gallaecio:
It is not what I had in mind, I think I may have confused you more than I helped :sweat_smile:. I must say I’m not sure what that code is meant to be doing :confused:.

For example, I don’t think `isinstance()` there will ever evaluate to `True`, `Settings` is a class, not an instance of a class. Same for the inner `if` statement, I imagine `super().set()` always returns `None`, assuming you can actually call `set()` without parameters and not get an exception raised.

I think you should get familiar with the `Settings` class, and how it is used in the codebase; then get familiar with the code that currently sets those unintended settings from environment variables; and finally, try to think of a way to warn users when desired.

joybhallaa:
@Gallaecio okay, thanks for your patience ! I'm still a beginner and would like to contribute more to scrapy.

PR #4410, opened 2020-03-07
Fix handling of None in allowed_domains

Lukas0907:
`allowed_domains = [None]` leads to the following exception:

```
2020-03-07 20:40:02 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method OffsiteMiddleware.spider_opened of <scrapy.spidermiddlewares.offsite.OffsiteMiddleware object at 0x7f57c2a22d60>>
  File "/home/lukas/Projects/3rd/scrapy/scrapy/utils/defer.py", line 161, in maybeDeferred_coro
    result = f(*args, **kw)
  File "/home/lukas/.pyenv/versions/3.8.2/envs/scrapy/lib/python3.8/site-packages/pydispatch/robustapply.py", line 55, in robustApply
  File "/home/lukas/Projects/3rd/scrapy/scrapy/spidermiddlewares/offsite.py", line 67, in spider_opened
    self.host_regex = self.get_host_regex(spider)
  File "/home/lukas/Projects/3rd/scrapy/scrapy/spidermiddlewares/offsite.py", line 58, in get_host_regex
    if url_pattern.match(domain):
TypeError: expected string or bytes-like object
```

However, Nones in allowed_domains ought to be ignored and there are also tests for that scenario. This commit fixes the handling of None and also the accompanying tests which are now executed again.

### DIFF
--- a/tests/test_spidermiddleware_offsite.py
+++ b/tests/test_spidermiddleware_offsite.py
@@ -55,21 +55,21 @@ class TestOffsiteMiddleware2(TestOffsiteMiddleware):
 
 class TestOffsiteMiddleware3(TestOffsiteMiddleware2):
 
-    def _get_spider(self):
-        return Spider('foo')
+    def _get_spiderargs(self):
+        return dict(name='foo')
 
 
 class TestOffsiteMiddleware4(TestOffsiteMiddleware3):
 
-    def _get_spider(self):
-      bad_hostname = urlparse('http:////scrapytest.org').hostname
-      return dict(name='foo', allowed_domains=['scrapytest.org', None, bad_hostname])
+    def _get_spiderargs(self):
+        bad_hostname = urlparse('http:////scrapytest.org').hostname
+        return dict(name='foo', allowed_domains=['scrapytest.org', None, bad_hostname])
 
     def test_process_spider_output(self):
-      res = Response('http://scrapytest.org')
-      reqs = [Request('http://scrapytest.org/1')]
-      out = list(self.mw.process_spider_output(res, reqs, self.spider))
-      self.assertEqual(out, reqs)
+        res = Response('http://scrapytest.org')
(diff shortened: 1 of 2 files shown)


tests/test_spidermiddleware_offsite.py
```diff
@@ -55,21 +55,21 @@ def test_process_spider_output(self):
 
 class TestOffsiteMiddleware3(TestOffsiteMiddleware2):
 
-    def _get_spider(self):
-        return Spider('foo')
+    def _get_spiderargs(self):
+        return dict(name='foo')
```
nyov: Just an aside. Does this class even do anything? Or could we clean this up?

nyov:
Could you add a testcase for the `allowed_domains=[None]` case? Otherwise I don't see what you mean by fixing the tests. :P

tests/test_spidermiddleware_offsite.py
Lukas0907: This is already the clean up! :)

The purpose of the class was to test that not setting the "allowed_domains" attribute in a spider works. However, due to a refactoring, the test was effectively a no-op because `_get_spider()` was never executed. My change should fix the test so it actually tests that use case again.

Lukas0907:
> Could you add a testcase for the `allowed_domains=[None]` case? Otherwise I don't see what you mean by fixing the tests. :P

The test case for `allowed_domains=[None]` was basically already there, it was just not executed (that's why it did not get noticed that it was broken). I changed it slightly and now it should work again: https://github.com/scrapy/scrapy/pull/4410/files#diff-b7fe032065662610160b880a6d821becR62-R72

tests/test_spidermiddleware_offsite.py
nyov: Ah, of course. I blanked out the subclassing of `TestOffsiteMiddleware2` somehow. Looks good.

PR #4411, opened 2020-03-07
Obsolete S3FeedStorage instancing without AWS credentials

nyov:
Question: does `access_key is None and secret_key is None` make sense here?
It's been the current behaviour, but is there a use-case for having no `access_key` but a `secret_key`? Otherwise this should possibly be `or`, instead of `and`?

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -94,23 +94,10 @@ class FileFeedStorage:
 class S3FeedStorage(BlockingFeedStorage):
 
     def __init__(self, uri, access_key=None, secret_key=None, acl=None):
-        # BEGIN Backward compatibility for initialising without keys (and
-        # without using from_crawler)
-        no_defaults = access_key is None and secret_key is None
-        if no_defaults:
-            from scrapy.utils.project import get_project_settings
-            settings = get_project_settings()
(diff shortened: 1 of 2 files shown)


Gallaecio:
Actually, I wonder if we shouldn’t just remove the whole thing.

The previous code allows for the settings not to be defined at all, what it warns about is having them in settings. Your new code raises an exception in such an scenario, which is backward incompatible.

It looks like it’s possible not to use credentials at all: https://aws.amazon.com/premiumsupport/knowledge-center/s3-private-connection-no-authentication/

nyov:
By "remove the whole thing", do you mean the whole `S3FeedStorage`? I think I've been advocating that in the past; well actually that was for `S3FilesStore`, I think. Moving those boto library users into a different project (as was done for scrapy-django-item) could "fix" this [CVE-2017-14158](https://security-tracker.debian.org/tracker/CVE-2017-14158).
But I have no strong opinion on that.

Gallaecio:
I mean lines `94:112` of the first file.

nyov:
Ah okay :) That's easy enough. I can't say if it's correct or not as I don't use it. But I'll change it, so someone else can test the case without any credentials at all.

kmike:
Thanks @nyov! Closed via #4688.

PR #4412, opened 2020-03-07
Deprecate Spider.make_requests_from_url, part 2

nyov:
Follow-up for #1728, 
Blocking #4178

### DIFF
--- a/scrapy/spiders/__init__.py
+++ b/scrapy/spiders/__init__.py
@@ -78,6 +78,12 @@ class Spider(object_ref):
 
     def make_requests_from_url(self, url):
         """ This method is deprecated. """
+        warnings.warn(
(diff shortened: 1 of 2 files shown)


scrapy/spiders/__init__.py
```diff
@@ -78,6 +78,12 @@ def start_requests(self):
 
     def make_requests_from_url(self, url):
         """ This method is deprecated. """
+        warnings.warn(
+            "Spider.make_requests_from_url method is deprecated; it "
+            "won't be called by the default Spider.start_requests method "
```
kmike: In addition to not being called, this method is also going to be removed. This is important if user code calls this method.

nyov: oh indeed, thanks. I hope the updated text fits better.

PR #4414, opened 2020-03-08
Added setting to choose custom asyncio event loop

### LINKED ISSUE #4306: Add a setting to choose a custom AsyncIO loop
Just as #4294 allows to easily choose a custom Twister reactor, I think we should consider implementing a setting that allows choosing a custom AsyncIO loop (when using an AsyncIO reactor).

I’m thinking of https://github.com/MagicStack/uvloop, which I’ve used in web development, and can’t help but wonder if it could improve Scrapy performance in some scenarios.

sakshamb2113:
Added setting to choose custom asyncio loop along with some tests. Works well with uvloop.

(edit) Fixes #4306


sakshamb2113:
@Gallaecio please review

wRAR:
Please add docs about the new setting.

Your tests only check that the setting is checked against `None`, can you please also add tests that actually check that the non-default loop is installed?

I'm not sure where should the uvloop test dep be added, maybe to tests/requirements-py3.txt directly?

sakshamb2113:
thanks @wRAR for pointing out the mistakes . I have written some tests which will check the installed loop.

scrapy/utils/log.py
```diff
@@ -150,6 +150,9 @@ def log_scrapy_info(settings):
                     if name != "Scrapy")})
     from twisted.internet import reactor
     logger.debug("Using reactor: %s.%s", reactor.__module__, reactor.__class__.__name__)
+    if reactor.__module__ == "twisted.internet.asyncioreactor" and reactor.__class__.__name__ == "AsyncioSelectorReactor":
```
Gallaecio: I think `isinstance` may be best here.

scrapy/utils/log.py
```diff
@@ -150,6 +150,9 @@ def log_scrapy_info(settings):
...
+        import asyncio
+        logger.debug("Using loop: %s", asyncio.get_event_loop().__module__)
```
Gallaecio: I think we should post the whole import path:

```python
loop = asyncio.get_event_loop()
logger.debug("Using loop: %s.%s", loop.__module__, loop.__class__.__name__)
```

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,21 @@ def __call__(self):
         return self._func(*self._a, **self._kw)
 
 
-def install_reactor(reactor_path):
+def install_reactor(reactor_path, event_loop_path):
     """Installs the :mod:`~twisted.internet.reactor` with the specified
-    import path."""
+    import path.Also installs asyncio event loop as specified in the import
```
Gallaecio:
```suggestion
    import path. Also installs the asyncio event loop with the specified import
```

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,21 @@ def __call__(self):
...
+    path if asyncio reactor is enabled"""
```
Gallaecio:
```suggestion
    path if the asyncio reactor is enabled"""
```

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,21 @@ def __call__(self):
         return self._func(*self._a, **self._kw)
 
 
-def install_reactor(reactor_path):
+def install_reactor(reactor_path, event_loop_path):
```
Gallaecio:
```suggestion
def install_reactor(reactor_path, event_loop_path=None):
```

So that function calls can be cleaner for non-asyncio reactors.

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,21 @@ def __call__(self):
...
     """Installs the :mod:`~twisted.internet.reactor` with the specified
-    import path."""
+    import path.Also installs asyncio event loop as specified in the import
+    path if asyncio reactor is enabled"""
     reactor_class = load_object(reactor_path)
     if reactor_class is asyncioreactor.AsyncioSelectorReactor:
         with suppress(error.ReactorAlreadyInstalledError):
-            asyncioreactor.install(asyncio.get_event_loop())
+            if event_loop_path is not None:
+                x = __import__(event_loop_path)
```
Gallaecio: I think you should use `import_module`. And I think you should not check if it is `None`, you should use whatever is loaded, and let it fail if the value is wrong.

And maybe `x` should be renamed to `event_loop_module` or something in that line. And `loop` to `event_loop` for consistency.

tests/CrawlerProcess/asyncio_custom_loop.py
```diff
@@ -0,0 +1,22 @@
+import asyncio
+
+from twisted.internet import asyncioreactor
+asyncioreactor.install(asyncio.get_event_loop())
```
Gallaecio: Why are these imports here? Shouldn’t the settings handle this part?

sakshamb2113:
@Gallaecio I have updated some doc files regarding the new setting.

docs/topics/settings.rst
```diff
@@ -218,6 +218,25 @@ Default: ``None``
...
 The name of the region associated with the AWS client.
 
+.. setting:: ASYNCIO_LOOP
+
+ASYNCIO_LOOP
+------------
+
+Default: ``None``
+
+Import path of a given asyncio loop.
```
elacuesta: I think there is an inconsistency here. Personally, I read this setting's description as if it were the import path of the event loop _class_, but this actually expects the _module_. It is documented as such in the above "Using custom asyncio loops" section, but to be consistent with the `TWISTED_REACTOR` setting I think it should be the path of the class instead.

Also, would `ASYNCIO_EVENT_LOOP` be a better name? Not a strong suggestion, just thinking out loud.

sakshamb2113: I am changing the setting to expect the class path for better consistency. Also changing the setting name to ASYNCIO_EVENT_LOOP

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
+If the asyncio reactor is enabled this setting can be used to specify the 
+asyncio event loop to be used with it. Set the setting to the import path of the 
+desired asyncio event loop class. If the setting is set to ``None`` the default asyncio
+event loop will be used.
+
+You can also specify the event loop path with the function :func:`~scrapy.utils.reactor.install_reactor`:
+
+.. autofunction:: scrapy.utils.reactor.install_reactor
+
+Note that the event loop class must inherit from ``asyncio.AbstractEventLoop``.
```
elacuesta:
```suggestion
Note that the event loop class must inherit from :class:`asyncio.AbstractEventLoop`.
```
This would make it link to https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop

scrapy/utils/log.py
```diff
@@ -150,6 +150,11 @@ def log_scrapy_info(settings):
                     if name != "Scrapy")})
     from twisted.internet import reactor
     logger.debug("Using reactor: %s.%s", reactor.__module__, reactor.__class__.__name__)
+    from twisted.internet import asyncioreactor
+    if isinstance(reactor, asyncioreactor.AsyncioSelectorReactor):
+        import asyncio
+        loop = asyncio.get_event_loop()
+        logger.debug("Using loop: %s.%s", loop.__module__, loop.__class__.__name__)
```
elacuesta:
```suggestion
        logger.debug("Using asyncio event loop: %s.%s", loop.__module__, loop.__class__.__name__)
```

sakshamb2113: Done

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
 The name of the region associated with the AWS client.
 
+.. setting:: ASYNCIO_EVENT_LOOP
+
+ASYNCIO_EVENT_LOOP
+------------------
+
+Default: ``None``
+
+Import path of a given asyncio event loop.
```
Gallaecio:
```suggestion
Import path of a given asyncio event loop class.
```

:lipstick: It may be the most obvious interpretation, but better not to leave room for doubt.

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
+
+If the asyncio reactor is enabled this setting can be used to specify the
```
Gallaecio: :lipstick: 

```suggestion
If the asyncio reactor is enabled (see :setting:`TWISTED_REACTOR`) this setting can be used to specify the 
```

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
+Import path of a given asyncio event loop.
+
+If the asyncio reactor is enabled this setting can be used to specify the 
+asyncio event loop to be used with it. Set the setting to the import path of the 
+desired asyncio event loop class. If the setting is set to ``None`` the default asyncio
+event loop will be used.
+
+You can also specify the event loop path with the function :func:`~scrapy.utils.reactor.install_reactor`:
+
+.. autofunction:: scrapy.utils.reactor.install_reactor
```
Gallaecio: The reference documentation should only be included once in the documentation, and this is already covered in the documentation of `TWISTED_REACTOR`, so this line should be removed.

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
+Default: ``None``
+
+Import path of a given asyncio event loop.
+
+If the asyncio reactor is enabled this setting can be used to specify the 
+asyncio event loop to be used with it. Set the setting to the import path of the 
+desired asyncio event loop class. If the setting is set to ``None`` the default asyncio
+event loop will be used.
+
+You can also specify the event loop path with the function :func:`~scrapy.utils.reactor.install_reactor`:
```
Gallaecio: It sounds like, as an alternative to using the `ASYNCIO_EVENT_LOOP` setting, you can use a parameter of the `install_reactor` function.

Instead, what it should say is that, if you are installing the asyncio reactor manually, the `ASYNCIO_EVENT_LOOP` setting has no effect. If you are installing it with `install_reactor`, you can use the `event_loop_path` parameter to indicate the import path of the event loop to use.

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,20 @@ def __call__(self):
...
     reactor_class = load_object(reactor_path)
     if reactor_class is asyncioreactor.AsyncioSelectorReactor:
         with suppress(error.ReactorAlreadyInstalledError):
-            asyncioreactor.install(asyncio.get_event_loop())
+            if event_loop_path is not None:
+                event_loop_class = load_object(event_loop_path)
+                event_loop = event_loop_class()
+                asyncio.set_event_loop(event_loop)
+            else:
+                event_loop = asyncio.get_event_loop()
```
Gallaecio: Is this necessary? Can’t we just skip `asyncioreactor.install` in this scenario altogether?

tests/test_commands.py
```diff
@@ -305,6 +305,18 @@ def test_asyncio_enabled_false(self):
...
+        log = self.get_log(self.debug_log_spider, args=[
+            '-s', 'TWISTED_REACTOR=twisted.internet.asyncioreactor.AsyncioSelectorReactor', '-s', 'ASYNCIO_EVENT_LOOP=uvloop.Loop'
+        ])
+        self.assertIn("Using asyncio event loop: uvloop.Loop", log)
+
+    def test_custom_asyncio_loop_enabled_false(self):
+        log = self.get_log(self.debug_log_spider, args=[
+            '-s', 'TWISTED_REACTOR=twisted.internet.asyncioreactor.AsyncioSelectorReactor'
+        ])
+        self.assertIn("Using asyncio event loop: asyncio.unix_events._UnixSelectorEventLoop", log)
```
Gallaecio: We need to make this check cross-platform.

scrapy/utils/reactor.py
sakshamb2113: But if we remove this part and `asyncioreactor.install` then how would we deal with the case when no event loop path is specified and the user wants to install the asyncio reactor with the default event loop ?

tests/test_commands.py
sakshamb2113: I found that asyncio uses different loops for unix and windows. Maybe its better to change the function and check for the absence of uvloop (which is not provided in the settings in this function) in the logs instead.

scrapy/utils/reactor.py
Gallaecio: I assume the default event loop is already used by default, if no event loop is installed.

tests/test_commands.py
Gallaecio: I think a better approach would be to get the default loop class from asyncio programmatically, and check that the one reported one matches that one.

scrapy/utils/reactor.py
sakshamb2113: I have removed the two lines but have left `asyncioreactor.install` as without it the reactor will not be installed.

tests/test_commands.py
sakshamb2113: Done

docs/topics/settings.rst
```diff
@@ -218,6 +218,26 @@ Default: ``None``
...
+
+Import path of a given asyncio event loop class.
+
+If the asyncio reactor is enabled (see :setting:`TWISTED_REACTOR`) this setting can be used to specify the 
+asyncio event loop to be used with it. Set the setting to the import path of the 
+desired asyncio event loop class. If the setting is set to ``None`` the default asyncio
+event loop will be used.
+
+If you are installing the asyncio reactor manually using the :func:`~scrapy.utils.reactor.install_reactor`
+function, you can use the :obj:`event_loop_path` parameter to indicate the import path of the event loop
```
Gallaecio:
```suggestion
function, you can use the ``event_loop_path`` parameter to indicate the import path of the event loop 
```

:lipstick: I don’t think you can link to function parameters.

scrapy/utils/reactor.py
```diff
@@ -49,13 +49,18 @@ def __call__(self):
...
+    path if the asyncio reactor is enabled"""
     reactor_class = load_object(reactor_path)
     if reactor_class is asyncioreactor.AsyncioSelectorReactor:
         with suppress(error.ReactorAlreadyInstalledError):
-            asyncioreactor.install(asyncio.get_event_loop())
+            if event_loop_path is not None:
+                event_loop_class = load_object(event_loop_path)
+                event_loop = event_loop_class()
+                asyncio.set_event_loop(event_loop)
+            asyncioreactor.install()
```
Gallaecio: Wait, my bad. It looks like calling `asyncioreactor.install()` is necessary in all cases. Please, feel free to go to your previous implementation. Sorry :disappointed:

tests/test_commands.py
```diff
@@ -305,6 +305,21 @@ def test_asyncio_enabled_false(self):
...
+        self.assertIn("Using asyncio event loop: uvloop.Loop", log)
+
+    def test_custom_asyncio_loop_enabled_false(self):
+        log = self.get_log(self.debug_log_spider, args=[
+            '-s', 'TWISTED_REACTOR=twisted.internet.asyncioreactor.AsyncioSelectorReactor'
+        ])
+        if sys.platform == 'win32':
+            self.assertIn("Using asyncio event loop: asyncio.windows_events._WindowsSelectorEventLoop", log)
+        else:
+            self.assertIn("Using asyncio event loop: asyncio.unix_events._UnixSelectorEventLoop", log)
```
Gallaecio: This may still be a problem, e.g. given https://github.com/tornadoweb/tornado/issues/2751

What I had in mind was using something like `new_event_loop()`, assuming that it returns an instance of the default loop class regardless of the loop installed in the Twisted reactor, and making sure that the reported asyncio event loop is of the same class.

scrapy/utils/reactor.py
sakshamb2113: I think that the current implementation works fine. Once we set the custom event loop in asyncio , it automatically becomes the default loop and `asyncioreactor` installs it. We don't really need to supply any parameter to `asyncioreactor.install()`.

tests/test_commands.py
sakshamb2113: I have used `new_event_loop()` to find the default asyncio loop according to the platform. I think it  should work now.

scrapy/utils/reactor.py
Gallaecio: Interesting. However, maybe we should avoid calling `asyncio.set_event_loop()`. Changing the default loop globally may cause unexpected issues in projects that execute another asyncio loop after the Scrapy-based crawling finishes.

sakshamb2113: I have changed reactor.py and log.py. Now the event loop is only set in the `asyncioreactor`. The global asyncio event loop is no longer changed.

sakshamb2113:
I found that the test `test_custom_loop_asyncio` in `test_crawler.py` does not complete. I ran the script   
`asyncio_custom_loop.py` in PyPy and got the following error

x-special/nautilus-clipboard
copy
file:///home/saksham/Pictures/Screenshot%20from%202020-03-29%2003-05-55.png


I think its related to https://github.com/MagicStack/uvloop/issues/132

sakshamb2113:
Maybe uvloop not fully supporting PyPy3 is the cause of this failure. What should we do in this case?

Gallaecio:
In that case, I think you should disable uvloop-related tests in PyPy. We have some tests disabled in specific Python versions, it should be possible to follow a similar approach to skip PyPy.

I was going to suggest looking for a different loop, but looking at https://github.com/timofurrer/awesome-asyncio#alternative-loops I don’t have much hope for that :slightly_smiling_face:

sakshamb2113:
I have disabled all the tests which use uvloop in pypy. Now all tests are passing.

scrapy/utils/reactor.py
```diff
@@ -50,13 +50,19 @@ def __call__(self):
...
+    path if the asyncio reactor is enabled"""
     reactor_class = load_object(reactor_path)
     if reactor_class is asyncioreactor.AsyncioSelectorReactor:
         with suppress(error.ReactorAlreadyInstalledError):
-            asyncioreactor.install(asyncio.get_event_loop())
+            if event_loop_path is not None:
+                event_loop_class = load_object(event_loop_path)
+                event_loop = event_loop_class()
+            else:
+                event_loop = asyncio.new_event_loop()
```
elacuesta, line 64: Is there any reason to use `new_event_loop` here as opposed to `get_event_loop` as before?

sakshamb2113: Calling `get_event_loop` may set the asyncio event loop globally since it calls  `set_event_loop`(which sets the loop as the default event loop)  along with `new_event_loop`(which creates a new event loop object). If someone plans to use some different asyncio event loop for work other than web crawling using scrapy in the same project, it may lead to issues.

sakshamb2113: This was discussed earlier so I tried to avoid using `set_event_loop`. https://github.com/scrapy/scrapy/pull/4414#discussion_r398017430

PR #4417, opened 2020-03-10
fix #2448:add s3 region support

### LINKED ISSUE #2448: S3FilesStore: add option for AWS Signature v4 support

hanguofeng:


wRAR:
Hi @hanguofeng , looks like one part of the change uses `AWS_REGION` while another one uses `AWS_REGION_NAME`? Please also add the docs and the tests for this new setting.

PR #4418, opened 2020-03-11
[WIP] Eliminate an internal use of os.environ 

nyov:
Currently, scrapy passes the SCRAPY_SETTINGS_MODULE setting between functions by using an environment variable, instead of inside python code. This "side-channel" is unnecessary.


nyov:
I've overlooked a case that makes this a more complex rewrite of scrapy's commandline handling. Which makes the changes backwards-incompatible and thus unlikely to be accepted. I'll put this effort on hold for now.

PR #4422, opened 2020-03-13
Restrict pytest to versions prior to 5.4

Gallaecio:
Let us see if that’s what is breaking the tests.


Gallaecio:
https://github.com/pytest-dev/pytest-twisted/issues/93

PR #4426, opened 2020-03-14
[WIP] Avoid .callback() in TestProcessProtocol.processEnded()

altendky:
pytest-dev/pytest-twisted#93


altendky:
Even if tests pass I do not know at present if this is the correct fix or not.  I'll follow up as I piece together some basis one way or the other.

altendky:
This was exploratory, exploration complete, still an underlying issue but perhaps there's a Twisted change to address the symptom...  https://github.com/pytest-dev/pytest-twisted/issues/93#issuecomment-605552821

PR #2411, opened 2016-11-24
[MRG+1] Remove ChunkedTransferMiddleware from default settings

redapple:
Following up on https://github.com/scrapy/scrapy/pull/1896

PR #4431, opened 2020-03-15
Remove deprecated ChunkedTransferMiddleware

elacuesta:
Deprecated in #2411 (v1.3)

See also #4356


scrapy/downloadermiddlewares/chunked.py
```diff
@@ -1,21 +0,0 @@
-import warnings
-
-from scrapy.exceptions import ScrapyDeprecationWarning
-from scrapy.utils.http import decode_chunked_transfer
```
kmike: what about this function?

kmike: Ah, I see, it was only deprecated in 1.7. Makes sense to keep it for now.

elacuesta: Yes, I removed it in an initial version of this branch but then I realized it was deprecated very recently (#3859)

PR #4434, opened 2020-03-16
Feed exports: add batch deliveries

### LINKED ISSUE #4250: Batch deliveries for long running crawlers
## Summary

Add a new setting `FEED_STORAGE_BATCH` that will deliver a file whenever `item_scraped_count` reaches a multiple of that number.

## Motivation

For long running jobs (say we are consuming inputs from a working queue) we may want partial results instead of waiting for a long batch to finish.

Of course we can stop and restart a spider every now and then.
However, a simpler approach is to have it running as long as required, but delivering partial results.

BroodingKangaroo:
Fixes #4250


scrapy/extensions/feedexport.py
```diff
@@ -241,28 +241,19 @@ def __init__(self, crawler):
 
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
+        self.storage_batch = self.settings.getint('FEED_STORAGE_BATCH')
```
ejulio: Maybe `FEED_STORAGE_BATCH_SIZE` is a better name.
Also, we could consider time based deliveries (say every X minutes), this one is just for consideration.

scrapy/extensions/feedexport.py
```diff
@@ -285,11 +276,49 @@ def close_spider(self, spider):
...
+            fields_to_export=feed['fields'],
+            encoding=feed['encoding'],
+            indent=feed['indent']
+        )
+        slot = _FeedSlot(file, exporter, storage, uri, feed['format'], feed['store_empty'])
+        if slot.store_empty:
+            slot.start_exporting()
+        return slot
+
+    def _get_uri_of_partial(self, template_uri, feed, spider):
```
ejulio: I guess you shouldn't force the timestamp in the file name.
Just log a warning if `%(time)s` is not in the URI as it will cause errors.
Maybe, we could add a new parameter, like `%(batch_id)s` also

scrapy/extensions/feedexport.py
```diff
@@ -241,28 +241,19 @@ def __init__(self, crawler):
...
+        self.storage_batch = self.settings.getint('FEED_STORAGE_BATCH')
         for uri, feed in self.feeds.items():
             if not self._storage_supported(uri):
                 raise NotConfigured
             if not self._exporter_supported(feed['format']):
                 raise NotConfigured
 
     def open_spider(self, spider):
+        if self.storage_batch:
+            self.feeds = {self._get_uri_of_partial(uri, feed, spider): feed for uri, feed in self.feeds.items()}
```
ejulio: One approach is that, whenever we reach the expected count, just close the current file/deliver it and create a new one.

scrapy/extensions/feedexport.py
BroodingKangaroo: If I understand correctly, the user must specify %(time)s and %(batch_id)s in the path and the input should be
scrapy crawl quotes -o res/test%(time)s.json - something like that?
Is %(batch_id)s a serial number of current batch?

ejulio: Yup, the users should set the parameter as they wish.
We should warn them if that parameter is missing.

The new `batch_id` parameter can be just a sequential id starting at 1..

scrapy/extensions/feedexport.py
```diff
@@ -241,28 +243,28 @@ def __init__(self, crawler):
...
+            batch_id = 1
+            uri_params = self._get_uri_params(spider, feed['uri_params'])
+            uri_params['batch_id'] = batch_id
+            self.slots.append(self._start_new_batch(
+                previous_batch_slot=None,
+                uri=uri % uri_params,
+                feed=feed,
+                spider=spider,
+                batch_id=batch_id,
+                template_uri=uri
```
elacuesta:
```suggestion
                template_uri=uri,
```
:lipstick:

scrapy/extensions/feedexport.py
```diff
@@ -285,11 +287,56 @@ def close_spider(self, spider):
...
+        storage = self._get_storage(uri)
+        file = storage.open(spider)
+        exporter = self._get_exporter(
+            file=file,
+            format=feed['format'],
+            fields_to_export=feed['fields'],
+            encoding=feed['encoding'],
+            indent=feed['indent']
+        )
+        slot = _FeedSlot(file, exporter, storage, uri, feed['format'], feed['store_empty'], batch_id, template_uri)
```
elacuesta: These are a lot of parameters, could you pass them using keywords?

scrapy/extensions/feedexport.py
```diff
@@ -285,11 +287,56 @@ def close_spider(self, spider):
...
+            previous_batch_slot.exporter.finish_exporting()
+            previous_batch_slot.storage.store(previous_batch_slot.file)
+        storage = self._get_storage(uri)
+        file = storage.open(spider)
+        exporter = self._get_exporter(
+            file=file,
+            format=feed['format'],
+            fields_to_export=feed['fields'],
+            encoding=feed['encoding'],
+            indent=feed['indent']
```
elacuesta:
```suggestion
            indent=feed['indent'],
```
:lipstick:

scrapy/settings/default_settings.py
```diff
@@ -146,6 +146,7 @@
     's3': 'scrapy.extensions.feedexport.S3FeedStorage',
     'ftp': 'scrapy.extensions.feedexport.FTPFeedStorage',
 }
+FEED_STORAGE_BATCH_SIZE = 0
```
elacuesta: I think a better default value for this setting, meaning no batch splitting, would be `None`.

scrapy/extensions/feedexport.py
```diff
@@ -285,11 +287,56 @@ def close_spider(self, spider):
...
+                    previous_batch_slot=slot,
+                    uri=slot.template_uri % uri_params,
+                    feed=self.feeds[slot.template_uri],
+                    spider=spider,
+                    batch_id=batch_id,
+                    template_uri=slot.template_uri
+                ))
+                self.slots[idx] = None
+        self.slots = [slot for slot in self.slots if slot is not None]
+        self.slots.extend(slots)
```
elacuesta: This does nothing, because `slots` is not modified after being set as `slots = []`

scrapy/extensions/feedexport.py
```diff
@@ -306,6 +353,17 @@ def _exporter_supported(self, format):
             return True
         logger.error("Unknown feed format: %(format)s", {'format': format})
 
+    def _batch_deliveries_supported(self, uri):
+        """
+        If FEED_STORAGE_BATCH_SIZE setting is specified uri has to contain %(time)s or %(batch_id)s
+        to distinguish different files of partial output
+        """
+        if not self.storage_batch_size:
```
elacuesta: Could you combine these three conditions into one single `ìf` check? Also, please include an explicit `return False`. It's not strictly needed because there is an implicit `None` being returned, which evaluates to `False`, but [_Explicit is better than implicit_](https://www.python.org/dev/peps/pep-0020).

Also, should the log level be "warning" instead? :thinking:

scrapy/extensions/feedexport.py
BroodingKangaroo: Thank you, `slots.append(self._start_new_batch` should be instead of `self.slots.append(self._start_new_batch`

scrapy/extensions/feedexport.py
BroodingKangaroo: Done.
I thought it can be more readable if I slit it into 2 `if`.

scrapy/extensions/feedexport.py
```diff
@@ -335,7 +400,7 @@ def _get_uri_params(self, spider, uri_params):
         params = {}
         for k in dir(spider):
             params[k] = getattr(spider, k)
-        ts = datetime.utcnow().replace(microsecond=0).isoformat().replace(':', '-')
+        ts = datetime.utcnow().isoformat().replace(':', '-')
```
ejulio: Any reason for this change?

scrapy/extensions/feedexport.py
```diff
@@ -241,28 +242,28 @@ def __init__(self, crawler):
...
-                encoding=feed['encoding'],
-                indent=feed['indent'],
-            )
-            slot = _FeedSlot(file, exporter, storage, uri, feed['format'], feed['store_empty'])
-            self.slots.append(slot)
-            if slot.store_empty:
-                slot.start_exporting()
+            batch_id = 1
+            uri_params = self._get_uri_params(spider, feed['uri_params'])
+            uri_params['batch_id'] = batch_id
```
ejulio: Probably this could go in `_get_uri_params`

scrapy/extensions/feedexport.py
```diff
@@ -285,11 +286,65 @@ def close_spider(self, spider):
...
+        return slot
+
     def item_scraped(self, item, spider):
-        for slot in self.slots:
+        slots = []
+        for idx, slot in enumerate(self.slots):
             slot.start_exporting()
             slot.exporter.export_item(item)
             slot.itemcount += 1
+            if self.storage_batch_size and slot.itemcount % self.storage_batch_size == 0:
```
ejulio: I guess you should be calling a routine similar to `close_spider` here, right?
Otherwise we're just creating the local files, but never sending the S3 or some other storage that is not local.
Also, I wrote a piece of this functionality and had issues when exporting to S3.
I had to upload `file.read()` instead of just `file` as it was raising errors.

scrapy/extensions/feedexport.py
BroodingKangaroo: replace(microsecond=0) - removes microseconds. But if several files were sent within one second they will have the same names.

scrapy/extensions/feedexport.py
BroodingKangaroo: It’s true, I thought about it, but didn’t figure out how to do it. I'll try again =)

ejulio:
@BroodingKangaroo , should I review something here or are you working on some of the changes?

scrapy/extensions/feedexport.py
BroodingKangaroo: The current implementation of the `_start_new_batch()` function repeats the `close_spider()` function in some aspects.
I tried running it for S3. It seems the program worked correctly.

ejulio:
@BroodingKangaroo sure, no problems.
Once you have finished the changes, just let me know and I'll review it

BroodingKangaroo:
Hello @ejulio, sorry for such big PR. Could you review it? =)
I added some tests to be confident that this functionality works well with all formats.
I also added documentation.

ejulio:
Regarding 
> Can removing replace(microsecond=0) from params['time'] = datetime.utcnow().isoformat().replace(':', '-') break someone's pipelines?

The idea is to keep backwards compatibility as much as possible.
If it breaks something, as this is a flagged feature (only deliver batches when the user asks for it) then it shouldn't break their current workflow

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,37 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+.. setting:: FEED_EXPORT_BATCH_SIZE
+
+FEED_EXPORT_BATCH_SIZE
+----------------------
+Default: ``None``
+
+An integer number which represent number of scraped items stored in each output
+file. Whenever the number of items exceeds this setting, a new file
+creates and output redirects to it.  The name of the new file will be selected
```
ejulio, line 442: > ... a new file is created and the output ...

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,37 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+based on timestamp when the feed is being created and/or batch sequence number.
+Therefore you must specify %(time)s or %(batch_id)s or both in the file path.
```
ejulio, line 444: > or both in `FEED_URI`

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,37 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
 .. _Amazon S3: https://aws.amazon.com/s3/
 .. _botocore: https://github.com/boto/botocore
 .. _Canned ACL: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
+
+.. setting:: FEED_EXPORT_BATCH_SIZE
+
+FEED_EXPORT_BATCH_SIZE
+----------------------
+Default: ``None``
```
ejulio, line 438: Not sure if `None` or `0`.
I recall that @kmike made a point some time ago about not being able to disable stuff (setting to `None`) through the CLI

scrapy/extensions/feedexport.py
ejulio: My advice here is to extract the common behavior to a function.
As there is no logging here about a file that was created.
Also, if you look into the behavior of `close_spider`, you'll notice that file storage is called in thread, therefore not blocking the current code.
Otherwise, we'll be blocking the main thread while storing the files and, maybe, causing performance issues

tests/test_feedexport.py
```diff
@@ -464,6 +445,61 @@ def parse(self, response):
...
+        """ Run spider with specified settings; return exported data. """
+
+        FEEDS = settings.get('FEEDS') or {}
+        settings['FEEDS'] = {
+            urljoin('file:', pathname2url(str(file_path))): feed
+            for file_path, feed in FEEDS.items()
+        }
+        content = {}
+        try:
+            with MockServer() as s:
```
ejulio, line 488: Maybe `scrapy.utils.test.get_crawler`?

BroodingKangaroo: Do you mean to use `scrapy.utils.test.get_crawler` instead of `MockServer` or it is about an error which I get trying to use `S3FeedStorageTest::test_store`?

BroodingKangaroo:
Hello @ejulio, thank you for the review. 

I prepared several changes regarding your review:
- Fixed typos in docs.
- Changed requirements for %(time)s in URI to %(time_id)s.
- Extracted close slot functionality to a function. So now there is logging about file creation.
- Added test which check storing file in S3 while using FEED_STORAGE_BATCH_SIZE != None.

docs/topics/feed-exports.rst
```diff
@@ -438,12 +438,12 @@ FEED_EXPORT_BATCH_SIZE
 Default: ``None``
 
 An integer number which represent number of scraped items stored in each output
-file. Whenever the number of items exceeds this setting, a new file
-creates and output redirects to it.  The name of the new file will be selected
+file. Whenever the number of items exceeds this setting, a new file is
+created and output redirects to it.  The name of the new file will be selected
 based on timestamp when the feed is being created and/or batch sequence number.
-Therefore you must specify %(time)s or %(batch_id)s or both in the file path.
+Therefore you must specify %(time_id)s or %(batch_id)s or both in FEED_URI.
```
ejulio, line 444: If `time_id` is to avoid keep backwards incompatibility with `time`.
Maybe, `batch_time` is a better name.
As I can't read `time_id` and get a sense of what it means

tests/test_feedexport.py
ejulio: Yeah, I don't know if you really need `MockServer`.
If it's required, then go for it.
Was just suggesting `get_crawler` as an alternative, in case you weren't aware of it :smile:

BroodingKangaroo: Okay, finally got it.
MockServer was in FeedExportTest before, so believe there was any reason to do this such way. I will also remember that get_crawler is an alternative. =))

docs/topics/feed-exports.rst
BroodingKangaroo: Done.

ejulio:
Also, not sure if the docs are failing because of change here
https://travis-ci.org/github/scrapy/scrapy/builds/682137008

Gallaecio:
The documentation issue was due to InterSphinx, we download files from online documentation sources and it’s not the first time that Twisted’s fails. I’ve restarted the job, hopefully it will run fine this time. Eventually we will need to look into the InterSphinx internal, maybe implement retries for those downloads.

scrapy/extensions/feedexport.py
```diff
@@ -241,55 +242,104 @@ def __init__(self, crawler):
...
+                                            extra={'spider': spider}))
+        d.addErrback(lambda f: logger.error(logfmt % "Error storing", log_args,
+                                            exc_info=failure_to_exc_info(f),
+                                            extra={'spider': spider}))
+        return d
+
+    def _start_new_batch(self, batch_id, uri, feed, spider, template_uri):
+        """
+        Redirect the output data stream to a new file.
+        Execute multiple times if 'FEED_STORAGE_BATCH' setting is specified.
```
Gallaecio, line 293: I guess this comes from a previous name of the setting.

scrapy/extensions/feedexport.py
```diff
@@ -241,55 +242,104 @@ def __init__(self, crawler):
...
+                slots.append(self._start_new_batch(
+                    batch_id=slot.batch_id + 1,
+                    uri=slot.template_uri % uri_params,
+                    feed=self.feeds[slot.template_uri],
+                    spider=spider,
+                    template_uri=slot.template_uri,
+                ))
+                self.slots[idx] = None
+        self.slots = [slot for slot in self.slots if slot is not None]
+        self.slots.extend(slots)
```
Gallaecio, line 342: I think this can be simplified a bit:

```suggestion
                slot = self._start_new_batch(
                    batch_id=slot.batch_id + 1,
                    uri=slot.template_uri % uri_params,
                    feed=self.feeds[slot.template_uri],
                    spider=spider,
                    template_uri=slot.template_uri,
                )
            slots.append(slot)
        self.slots = slots
```

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,40 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+FEED_STORAGE_BATCH_ITEM_COUNT
+-----------------------------
+Default: ``None``
+
+An integer number that represents the number of scraped items stored in each output
+file. Whenever the number of items exceeds this setting, a new file is
+created and the output is redirected to it.  The name of the new file will be selected
+based on the timestamp when the feed is being created and/or on the batch sequence number.
+Therefore you must specify %(batch_time)s or %(batch_id)s or both in :setting:`FEED_URI`.
```
Gallaecio, line 444: Some things to consider:
- It should be more obvious, and explained as early as possible, that Scrapy only generates multiple output files if this set to a different value (so, not by default).
- This paragraph covers two significantly different topics (the setting value, and the handling of file names when a non-zero value is used), so different paragraphs make sense.

For example:
```suggestion
If assigned an integer number higher than `0`, Scrapy generates multiple output files
storing up to the specified number of items in each output file.

When generating multiple output files, you must use at least one of the following
placeholders in :setting:`FEED_URI` to indicate how the different output file names are
generated:
```

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,40 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+* ``%(batch_time)s`` - gets replaced by a timestamp when the feed is being created
+  (e.g. `2020-03-28T14-45-08.237134`)
+
+* ``%(batch_id)s`` - gets replaced by the batch  sequence number of batch
+  (e.g. `2` for the second file)
```
Gallaecio, line 450: My bad on the code quotes (and some minor issues):

```suggestion
* ``%(batch_time)s`` - gets replaced by a timestamp when the feed is being created
  (e.g. ``2020-03-28T14-45-08.237134``)

* ``%(batch_id)s`` - gets replaced by the batch sequence number
  (e.g. ``2`` for the second file)
```

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,40 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+For instance, if your settings include::
+
+    FEED_STORAGE_BATCH_ITEM_COUNT=100
+
+And your :command:`crawl` command line is::
+
+  scrapy crawl spidername -o dirname/%(batch_id)s-filename%(batch_time)s.json
+
+The resulting directory tree of above can be like::
```
Gallaecio, line 460: I find “of above” a bit awkward in that sentence. If you want to include a spatial reference to “above”, what about something more like this?

```suggestion
The command line above can generate a directory tree like::
```

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,40 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+->projectname
+-->dirname
+--->1-filename2020-03-28T14-45-08.237134.json
+--->2-filename2020-03-28T14-45-09.148903.json
+--->3-filename2020-03-28T14-45-10.046092.json
+
+Where the first and second files contain exactly 100 items. The last one contains
+<= 100 items.
```
Gallaecio, line 469: :lipstick: 

```suggestion
100 items or fewer.
```

scrapy/extensions/feedexport.py
```diff
@@ -180,14 +180,16 @@ def _store_in_thread(self, file):
 
 
 class _FeedSlot:
-    def __init__(self, file, exporter, storage, uri, format, store_empty):
+    def __init__(self, file, exporter, storage, uri, format, store_empty, batch_id, template_uri):
```
Gallaecio, line 183: I think `uri_template` would be more accurate, it’s the template of a URI, not the URI of a template.

docs/topics/feed-exports.rst
BroodingKangaroo: > If assigned an integer number higher than `0`

Should we also use double quotes here?

Gallaecio: Indeed :facepalm:

Gallaecio: Also, mind that to make a separate paragraph a new line is not enough. You need an empty line in between.

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,41 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+The command line above can generate a directory tree like::
+
+->projectname
+-->dirname
+--->1-filename2020-03-28T14-45-08.237134.json
+--->2-filename2020-03-28T14-45-09.148903.json
+--->3-filename2020-03-28T14-45-10.046092.json
+
+Where the first and second files contain exactly 100 items. The last one contains
+100 items or fever.
```
Gallaecio, line 470:
```suggestion
100 items or fewer.
```

scrapy/extensions/feedexport.py
```diff
@@ -241,55 +243,104 @@ def __init__(self, crawler):
 
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
+        self.storage_batch_size = self.settings.get('FEED_STORAGE_BATCH_ITEM_COUNT', None)
```
Gallaecio, line 246: For consistency:
```suggestion
        self.storage_batch_item_count = self.settings.get('FEED_STORAGE_BATCH_ITEM_COUNT', None)
```

scrapy/extensions/feedexport.py
```diff
@@ -241,55 +243,104 @@ def __init__(self, crawler):
...
         for uri, feed in self.feeds.items():
             if not self._storage_supported(uri):
                 raise NotConfigured
+            if not self._batch_deliveries_supported(uri):
```
Gallaecio, line 250: We are not checking if batch deliveries are supported. We are checking if the setting values affecting batch deliveries are consistent. So you can change the method name to something more in that line.

You may not even have to make it specific to batch deliveries, you could give the method a generic name in the line of `_settings_are_valid` or `_options_are_valid`. In the future we can also pass the `feed` variable to that method and use it to check the validity of several options and settings.

scrapy/extensions/feedexport.py
```diff
@@ -306,6 +357,16 @@ def _exporter_supported(self, format):
             return True
         logger.error("Unknown feed format: %(format)s", {'format': format})
 
+    def _batch_deliveries_supported(self, uri):
+        """
+        If FEED_STORAGE_BATCH_ITEM_COUNT setting is specified uri has to contain %(batch_time)s or %(batch_id)s
+        to distinguish different files of partial output
+        """
+        if self.storage_batch_size is None or '%(batch_time)s' in uri or '%(batch_id)s' in uri:
```
Gallaecio, line 365: To support `0` as well:

```suggestion
        if not self.storage_batch_size or '%(batch_time)s' in uri or '%(batch_id)s' in uri:
```

tests/test_feedexport.py
```diff
@@ -601,7 +616,7 @@ def test_export_no_items_not_store_empty(self):
                 },
             }
             data = yield self.exported_no_data(settings)
-            self.assertEqual(data[fmt], b'')
+            self.assertEqual(b'', data[fmt])
```
Gallaecio, line 619: Was there a reason for this change? :slightly_smiling_face:

tests/test_feedexport.py
```diff
@@ -969,3 +984,341 @@ def test_pathlib_uri(self):
         }
         data = yield self.exported_no_data(settings)
         self.assertEqual(data['csv'], b'')
+
+
+class PartialDeliveriesTest(FeedExportTestBase):
```
Gallaecio, line 989: Here and in a few other parts of the code it says “partial”; ”batch” is more consistent with the rest of the code, and more unambiguous (“partial” may be interpreted as not delivering all data).

```suggestion
class BatchDeliveriesTest(FeedExportTestBase):
```

tox.ini
```diff
@@ -14,6 +14,7 @@ deps =
     # Extras
     botocore>=1.3.23
     Pillow>=3.4.2
+    boto3>=1.13.0
```
Gallaecio, line 17: :lipstick: A-Z
```suggestion
    boto3>=1.13.0
    botocore>=1.3.23
    Pillow>=3.4.2
```

Gallaecio:
Most of my feedback if purely aesthetic, the code looks solid.

Mind that @ejulio’s point about the default value changing to `0` still remains. I don’t have a strong opinion about it as long as both `None` and `0` work. Looking at `default_settings.py`, `0` is indeed the common default value for integer options where `0` disables the effect altogether (`CLOSESPIDER_*`, `CONCURRENT_REQUESTS_PER_IP`, `DEPTH_LIMIT`…)

tests/test_feedexport.py
BroodingKangaroo: I saw such tox error output "expected != got" or something like this, so I supposed that first value was considered as expected, second - got. For some reason, I can't reproduce such situation now.

elacuesta:
I'll take responsibility for disabling with `None`, it was [my suggestion](https://github.com/scrapy/scrapy/pull/4434#discussion_r395814960). It looked better to me as a disabling value because it does not belong to the domain for the setting, and I guess from a mathematical point of view? (having a batch size of zero would mean having infinite batches with no elements? :exploding_head:). In any case, it looks like it's consistent to use `0` to disable stuff, so no objections from me.

scrapy/extensions/feedexport.py
```diff
@@ -306,6 +357,16 @@ def _exporter_supported(self, format):
...
         logger.error("Unknown feed format: %(format)s", {'format': format})
 
+    def _settings_are_valid(self, uri):
+        """
+        If FEED_STORAGE_BATCH_ITEM_COUNT setting is specified uri has to contain %(batch_time)s or %(batch_id)s
+        to distinguish different files of partial output
+        """
+        if not self.storage_batch_item_count or '%(batch_time)s' in uri or '%(batch_id)s' in uri:
+            return True
+        logger.warning('%(batch_time)s or %(batch_id)s must be in uri if FEED_STORAGE_BATCH_ITEM_COUNT setting is specified')
```
Gallaecio, line 367:
```suggestion
        logger.error('%(batch_time)s or %(batch_id)s must be in uri if FEED_STORAGE_BATCH_ITEM_COUNT setting is specified')
```

`_exporter_supported` and `_storage_supported` both log it as an error, and since the result if the feed export being disabled, I guess it makes sense.

scrapy/extensions/feedexport.py
```diff
@@ -331,12 +394,13 @@ def _get_exporter(self, file, format, *args, **kwargs):
...
 
-    def _get_uri_params(self, spider, uri_params):
+    def _get_uri_params(self, spider, uri_params, slot):
         params = {}
         for k in dir(spider):
             params[k] = getattr(spider, k)
-        ts = datetime.utcnow().replace(microsecond=0).isoformat().replace(':', '-')
-        params['time'] = ts
+        params['time'] = datetime.utcnow().replace(microsecond=0).isoformat().replace(':', '-')
+        params['batch_time'] = datetime.utcnow().isoformat().replace(':', '-')
```
kmike, line 402: What do you think about calling  datetime.utcnow() once? Otherwise the results may be different.

scrapy/extensions/feedexport.py
```diff
@@ -241,55 +243,104 @@ def __init__(self, crawler):
 
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
+        self.storage_batch_item_count = self.settings.get('FEED_STORAGE_BATCH_ITEM_COUNT', None)
```
kmike, line 246: Would it be possible to have batch_item_count support in FEEDS setting?

scrapy/extensions/feedexport.py
BroodingKangaroo: I guess it is very unlikely but there is no need to test our fortune =).
Thank you.

scrapy/extensions/feedexport.py
BroodingKangaroo: I guess so. Should I do it in this PR or it is already overloaded.

Gallaecio: It would be great if we could have it as part of this pull request.

docs/topics/feed-exports.rst
victor-torres: What if it's set to `1`?

scrapy/extensions/feedexport.py
victor-torres:
```suggestion
        self.storage_batch_item_count = self.settings.getint('FEED_STORAGE_BATCH_ITEM_COUNT')
```

docs/topics/feed-exports.rst
```diff
@@ -429,3 +430,41 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
 .. _Amazon S3: https://aws.amazon.com/s3/
 .. _botocore: https://github.com/boto/botocore
 .. _Canned ACL: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
+
+.. setting:: FEED_STORAGE_BATCH_ITEM_COUNT
+
+FEED_STORAGE_BATCH_ITEM_COUNT
+-----------------------------
+Default: ``None``
```
victor-torres, line 438: I agree with @ejulio:

```suggestion
Default: ``0``
```

And we could use `BaseSettings.getint` instead of `BaseSettings.get` when querying this setting.

scrapy/settings/default_settings.py
```diff
@@ -146,6 +146,7 @@
     's3': 'scrapy.extensions.feedexport.S3FeedStorage',
     'ftp': 'scrapy.extensions.feedexport.FTPFeedStorage',
 }
+FEED_STORAGE_BATCH_ITEM_COUNT = None
```
victor-torres, line 149:
```suggestion
FEED_STORAGE_BATCH_ITEM_COUNT = 0
```

tests/test_feedexport.py
```diff
@@ -978,3 +993,349 @@ def test_pathlib_uri(self):
...
+
+    @defer.inlineCallbacks
+    def assertExportedJsonLines(self, items, rows, settings=None):
+        settings = settings or {}
+        settings.update({
+            'FEEDS': {
+                os.path.join(self._random_temp_filename(), 'jl', self._file_mark): {'format': 'jl'},
+            },
+        })
+        batch_size = settings['FEED_STORAGE_BATCH_ITEM_COUNT']
```
victor-torres, line 1036: Should we use `BaseSettings.getint` here?

```suggestion
        batch_size = settings.getint('FEED_STORAGE_BATCH_ITEM_COUNT')
```

vshlapakov:
@BroodingKangaroo Good job here :+1: We're interested in moving forward with the feature, any chance you could address the last mentioned things in the short terms? If you don't have time for it, that's totally fine, we possibly could help finishing it in a separate PR on top of all the commits.

scrapy/extensions/feedexport.py
```diff
@@ -25,7 +25,6 @@
 from scrapy.utils.misc import create_instance, load_object
 from scrapy.utils.python import without_none_values
 
-
 logger = logging.getLogger(__name__)
```
rafaelcapucho, line 28: Hi, could you bring the empty line that was above the logger back? it should be 2 empty lines between imports and the code itself, It has now only one, thanks.

BroodingKangaroo: Hi. Of course. I don't know why but any time I am trying to reformat code using PyCharm it deletes that line. This time I forgot to bring it back.

BroodingKangaroo:
@vshlapakov I'm sorry it took so long to add changes from comments. I was a little busy at that time.

Does someone know what should be done to evade py3.5 test falling?

scrapy/extensions/feedexport.py
```diff
@@ -241,66 +242,115 @@ def __init__(self, crawler):
 
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
+        self.storage_batch_item_count = self.settings.getint('FEED_STORAGE_BATCH_ITEM_COUNT')
```
elacuesta, line 245: Instead of doing this here, please update [`scrapy.utils.conf.feed_complete_default_values_from_settings`](https://github.com/scrapy/scrapy/blob/2.2.0/scrapy/utils/conf.py#L114). That would simplify the `if` at line 339.

scrapy/extensions/feedexport.py
```diff
@@ -241,66 +242,115 @@ def __init__(self, crawler):
...
-        for slot in self.slots:
+        slots = []
+        for idx, slot in enumerate(self.slots):
             slot.start_exporting()
             slot.exporter.export_item(item)
             slot.itemcount += 1
+            # create new slot for each slot with itemcount == FEED_STORAGE_BATCH_ITEM_COUNT and close the old one
+            if self.feeds[slot.uri_template].get('batch_item_count', self.storage_batch_item_count) \
+                    and slot.itemcount == self.feeds[slot.uri_template].get('batch_item_count',
+                                                                            self.storage_batch_item_count):
```
elacuesta, line 341:
```suggestion
            if (
                self.feeds[slot.uri_template]['batch_item_count']
                and slot.itemcount >= self.feeds[slot.uri_template]['batch_item_count']
            ):
```

This simplification works after applying the above suggestion about updating `scrapy.utils.conf.feed_complete_default_values_from_settings`.

Also, notice the change from `==` to `>=`: if something goes wrong during the closing of the current batch, it could happen that the slot item count keeps being incremented and never reaches the batch size. I'm not sure I'm being clear, I'm also not sure that could happen in this case but I think it's safer anyway. I was bitten by an error caused by this recently and it's still fresh in my memory.

scrapy/extensions/feedexport.py
```diff
@@ -241,66 +242,115 @@ def __init__(self, crawler):
...
+            uri_template=uri_template,
+        )
+        if slot.store_empty:
+            slot.start_exporting()
+        return slot
+
     def item_scraped(self, item, spider):
-        for slot in self.slots:
+        slots = []
+        for idx, slot in enumerate(self.slots):
```
elacuesta, line 334: Seems like the `enumerate` call is a remnant from a previous implementation, could you revert it?

scrapy/extensions/feedexport.py
```diff
@@ -317,6 +367,18 @@ def _exporter_supported(self, format):
...
 
+    def _settings_are_valid(self, uri):
+        """
+        If FEED_STORAGE_BATCH_ITEM_COUNT setting is specified uri has to contain %(batch_time)s or %(batch_id)s
+        to distinguish different files of partial output
+        """
+        if not self.storage_batch_item_count or '%(batch_time)s' in uri or '%(batch_id)s' in uri:
+            return True
+        logger.error(
+            '%(batch_time)s or %(batch_id)s must be in uri if FEED_STORAGE_BATCH_ITEM_COUNT setting is specified'
```
elacuesta, line 378: This message might need an update to reflect that fact that the batch size could come from `FEEDS.batch_item_count`.

docs/topics/feed-exports.rst
```diff
@@ -429,3 +431,41 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+When generating multiple output files, you must use at least one of the following
+placeholders in :setting:`FEED_URI` to indicate how the different output file names are
+generated:
+
+* ``%(batch_time)s`` - gets replaced by a timestamp when the feed is being created
+  (e.g. ``2020-03-28T14-45-08.237134``)
+
+* ``%(batch_id)s`` - gets replaced by the batch  sequence number of batch
+  (e.g. ``2`` for the second file)
```
elacuesta, line 452: Should some kind of padding be added here? A simple numeric increment would produce unexpected results I think:
```
>>> sorted(["1", "2", "10"])
['1', '10', '2']
```

`zfill` seems like a reasonable solution, but I'm not sure about the amount of characters.

docs/topics/feed-exports.rst
```diff
@@ -429,3 +431,41 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+
+FEED_STORAGE_BATCH_ITEM_COUNT
+-----------------------------
+Default: ``0``
+
+If assigned an integer number higher than ``0``, Scrapy generates multiple output files
+storing up to the specified number of items in each output file.
+
+When generating multiple output files, you must use at least one of the following
+placeholders in :setting:`FEED_URI` to indicate how the different output file names are
```
elacuesta, line 445:
```suggestion
placeholders in the feed URI to indicate how the different output file names are
```
`FEED_URI` was deprecated with the introduction of `FEEDS`

scrapy/extensions/feedexport.py
```diff
@@ -241,66 +242,115 @@ def __init__(self, crawler):
...
-                format=feed['format'],
-                fields_to_export=feed['fields'],
-                encoding=feed['encoding'],
-                indent=feed['indent'],
-            )
-            slot = _FeedSlot(file, exporter, storage, uri, feed['format'], feed['store_empty'])
-            self.slots.append(slot)
-            if slot.store_empty:
-                slot.start_exporting()
+            uri_params = self._get_uri_params(spider, feed['uri_params'], None)
```
elacuesta, line 256: I think it would be better to set `None` as default value for `slot` in `_get_uri_params` .

elacuesta:
> Does someone know what should be done to evade py3.5 test falling?

[The build error](https://travis-ci.org/github/scrapy/scrapy/jobs/702822198#L723) is caused by the fact that dicts are not guarantied to be ordered in Python 3.5. This was changed as an implementation detail in [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) and added to the language spec in [Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html). One way to overcome this, used by other tests, is to use items which only have one field.

victor-torres:
> > Does someone know what should be done to evade py3.5 test falling?
> 
> [The build error](https://travis-ci.org/github/scrapy/scrapy/jobs/702822198#L723) is caused by the fact that dicts are not guarantied to be ordered in Python 3.5. This was changed as an implementation detail in [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) and added to the language spec in [Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html). One way to overcome this, used by other tests, is to use items which only have one field.

You're correct about the dictionary order in Python 3.5, but it seems like we're comparing two JSON strings.

AFAIK, JSON strings are not deterministic. I mean, the same data could be represented in multiple ways using JSON specification. We could have a different field order, number of whitespaces, number of new lines.

Perhaps, the best option would be to create a new method called `assertEqualJSON` or something like that, which actually parses the JSON content into a dictionary and uses Python's built-in comparison to assert that the content of the two JSON strings is equivalent.

Does it make sense?

scrapy/extensions/feedexport.py
victor-torres: Hey, @rafaelcapucho, I've just come across this comment and figured out PyCharm is doing the same thing to my code. I've decided to check the PEP 8 and this is what it says regarding [Blank Lines](https://www.python.org/dev/peps/pep-0008/#blank-lines):

> Surround top-level function and class definitions with two blank lines.
>
> Method definitions inside a class are surrounded by a single blank line.
>
> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
> 
> Use blank lines in functions, sparingly, to indicate logical sections.
>
> Python accepts the control-L (i.e. ^L) form feed character as whitespace; Many tools treat these characters as page separators, so you may use them to separate pages of related sections of your file. Note, some editors and web-based code viewers may not recognize control-L as a form feed and will show another glyph in its place.

Since this is not a top-level function or class definition, I don't think it's mandatory to surround it with two blank lines. 

Do you have another source or is it something related to Scrapy development guidelines itself?

elacuesta:
True, that is another possibility, like it's done [here](https://github.com/scrapy/scrapy/blob/2.2.0/tests/test_exporters.py#L226-L232) and [here](https://github.com/scrapy/scrapy/blob/2.2.0/tests/test_exporters.py#L317-L329).

docs/topics/feed-exports.rst
BroodingKangaroo: This is a good idea, but for use zfill we should know the maximum amount of characters in %(batch_id)s in advance. If I clearly understand what you propose.

scrapy/extensions/feedexport.py
Gallaecio: The [convention](https://stackoverflow.com/a/2953268/939364) is to keep two blank lines between imports and the rest of the code. I think it makes sense, as you already use a single blank line to separate different groups of imports.

docs/topics/feed-exports.rst
Gallaecio: I think the current approach is OK. Smart file browsers should be able to sort the generated files properly.

In the future, maybe we can extend the `%(batch_id)s` syntax to allow users to specify a number of leading zeroes.

Gallaecio:
> Since this doesn't affect the storage backends, should the setting be called `FEED_EXPORT_BATCH_ITEM_COUNT` instead, for consistency? Even more, I think `FEED_EXPORT_BATCH_SIZE` (and `FEEDS.batch_size`) would be shorter and simpler.

I think “size” should be avoided, as it can be misleading (item count? bytes?).

On a different but related topic, I think it makes sense for new `FEEDS`-related options not to have a setting of its own, and only be settable through `FEEDS`. At least that is the approach I took with https://github.com/scrapy/scrapy/pull/4512.

scrapy/utils/conf.py
```diff
@@ -115,6 +115,10 @@ def feed_complete_default_values_from_settings(feed, settings):
     out = feed.copy()
     out.setdefault("encoding", settings["FEED_EXPORT_ENCODING"])
     out.setdefault("fields", settings.getlist("FEED_EXPORT_FIELDS") or None)
+    out.setdefault(
+        "batch_item_count",
+        out.get('batch_item_count', settings.getint('FEED_STORAGE_BATCH_ITEM_COUNT'))
+    )
```
elacuesta, line 121:
```suggestion
    out.setdefault("batch_item_count", settings.getint('FEED_STORAGE_BATCH_ITEM_COUNT')))
```

No need to check for the presence of a current value, `setdefault` will act only if it's missing.

docs/topics/feed-exports.rst
elacuesta: I have traveled into the future and I've seen a lot of issues about this 😛 

The problem is that most file browsers are not that smart. The only ones that worked for me were Finder on macOS and Caja on Linux Mint, and I think that wasn't the case not so long ago.
In particular, `ls` and `aws s3 ls` do not work, and I think that's a problem.

What about going with a fixed number, like 4? It's pushing the problem into the future, but given that the current code allows only one batch, I believe allowing 10000 should be more than enough for most cases, only adding 3 bytes to the existing file names.

elacuesta:
> On a different but related topic, I think it makes sense for new FEEDS-related options not to have a setting of its own, and only be settable through FEEDS

That makes a lot of sense to keep the codebase simple, however having general settings also allow to keep project settings simple, so I don't really have a strong opinion at the moment.

On the other hand, you convinced me about the "size" name. That said, I do think that _if_ we introduce a new general setting, it should be `FEED_EXPORT_BATCH_ITEM_COUNT` instead of `FEED_STORAGE_BATCH_ITEM_COUNT`

scrapy/extensions/feedexport.py
victor-torres: Thank you for your link, @Gallaecio. 

Yet, the [top voted answer](https://stackoverflow.com/a/2953268/3072406) (marked as correct by the OP) does not state any reference:

> 1. Two blank lines between the import statements and other code.
> 2. Two blank lines between each function.

In the same line as PEP 8, [Google's Python Style Guide](https://google.github.io/styleguide/pyguide.html#35-blank-lines) also only mentions top-level function and class definitions:

> Two blank lines between top-level definitions, be they function or class definitions. One blank line between method definitions and between the class line and the first method. No blank line following a def line. Use single blank lines as you judge appropriate within functions or methods.

Let's take the `logger` variable definition, something very common across Python projects, as an example.

In Scrapy, we have some differences (both valid according to PEP 8):

- it's surrounded by two blank lines in [scrapy/downloadermiddlewares/cookies.py](https://github.com/scrapy/scrapy/blob/1d2b692666f9f5265e4845f8a30f04089515f7c5/scrapy/downloadermiddlewares/cookies.py) and in [scrapy/utils/signal.py](https://github.com/scrapy/scrapy/blob/1d2b692666f9f5265e4845f8a30f04089515f7c5/scrapy/utils/signal.py);
- but it's surrounded by only one blank line in [scrapy/spidermiddlewares/urllength.py](https://github.com/scrapy/scrapy/blob/dfbe1d95071acfbba159cea051530749b1684460/scrapy/spidermiddlewares/urllength.py) and in [scrapy/core/engine.py](https://github.com/scrapy/scrapy/blob/3d58e5e3be9e266b9b0f89b094823b76511a21b1/scrapy/core/engine.py)

Apparently, [Django is using only one blank line in this case](https://github.com/django/django/search?q=logger+%3D&unscoped_q=logger+%3D).

What I'm trying to say is that there's no reference to this convention on the [Scrapy's Coding Style](https://docs.scrapy.org/en/master/contributing.html?highlight=blank%20lines#coding-style), and IMO it's quite subjective. 

If this is important for the project to the point that we need to enforce it during code reviews, I propose we implement automatic code checks with a linter or at least document it in the Scrapy's Coding Style for future reference.

What do you think? Should we create an issue?

elacuesta: > it's surrounded by two blank lines in scrapy/downloadermiddlewares/cookies.py and in scrapy/utils/signal.py

I personally like two lines after imports, those two examples are on me 😄 

I believe there are a few unwritten conventions which are not taken directly from PEP 8. For instance, @Gallaecio has been promoting alphabetically sorted imports, putting `import` lines before `from ... import` ones. I'm a big fan of [parenthesized multi-line conditions with leading logical operators](https://github.com/scrapy/scrapy/pull/4525/files#diff-115565e38381736e9e87227cb69ac0f2L64-R70).

Perhaps the time has come to write all of these conventions down.

docs/topics/feed-exports.rst
kmike: I wonder if we can solve it by

1) documenting that batch_id is always an integer
2) using a different format specifier, e.g.  "%(batch_id)04d" or "%(batch_id)05d"
3) explainng the issue in the docs and showing how to increase leading zero count by using a different format specifier

scrapy/extensions/feedexport.py
victor-torres: Sure, @elacuesta.

I've created a [ticket](https://github.com/scrapy/scrapy/issues/4654) on Scrapy to continue this discussion outside this pull request.

@rafaelcapucho, I'm sorry if I sounded like I was nitpicking stuff. I'm just trying to turn this into an objective style guideline, ideally automatically enforced by a code linter or at least well documented, independently of the number of blank lines we end up deciding to use.

Cheers!

scrapy/extensions/feedexport.py
```diff
@@ -244,63 +245,112 @@ def __init__(self, crawler):
...
     def item_scraped(self, item, spider):
+        slots = []
         for slot in self.slots:
             slot.start_exporting()
             slot.exporter.export_item(item)
             slot.itemcount += 1
+            # create new slot for each slot with itemcount == FEED_STORAGE_BATCH_ITEM_COUNT and close the old one
+            if (
+                    self.feeds[slot.uri_template]['batch_item_count']
+                    and slot.itemcount >= self.feeds[slot.uri_template]['batch_item_count']
```
elacuesta, line 340:
```suggestion
                self.feeds[slot.uri_template]['batch_item_count']
                and slot.itemcount >= self.feeds[slot.uri_template]['batch_item_count']
```

docs/topics/feed-exports.rst
elacuesta: That's great, I knew about stuff like `"{batch_id:04d}".format(batch_id=1)` but I wasn't aware of the percentage formatting equivalent. Seems like no code changes are necessary then, documentation is the way to go. Thanks!

docs/topics/feed-exports.rst
```diff
@@ -448,16 +448,19 @@ generated:
 * ``%(batch_time)s`` - gets replaced by a timestamp when the feed is being created
   (e.g. ``2020-03-28T14-45-08.237134``)
 
-* ``%(batch_id)s`` - gets replaced by the batch  sequence number of batch
-  (e.g. ``2`` for the second file)
+* ``%(batch_id)0xd`` - gets replaced by the sequence number of the batch.
+By replacing ``x`` with an integer you set the number of leading zeroes to prevent
+inappropriate sorting like this: [``'1'``, ``'10'``, ``'2'``]. Here are some examples:
+    ``%(batch_id)01d`` for the second batch gets replaced by ``2``
+    ``%(batch_id)05d`` for the third batch gets replaced by ``00003``
```
Gallaecio, line 455: I would:
- Replace `%(batch_id)0xd` with `%(batch_id)d` (the simplest form that you can use)
- Replace “By replacing… ” by a reference to https://docs.python.org/3/library/stdtypes.html#old-string-formatting
- Remove the `01d` example (that would be the default behavior without formatting specifiers), leaving only the `05d` one. And since it is only one example, it should probably be featured inline, not as a list item.

```suggestion
* ``%(batch_id)d`` - gets replaced by the sequence number of the batch.

  Use :ref:`printf-style string formatting <python:old-string-formatting>` to
  alter the number format. For example, to make the batch ID a 5-digit
  number by introducing leading zeroes as needed, use ``%(batch_id)05d``
  (e.g. ``3`` becomes ``00003``, ``123`` becomes ``00123``).
```

scrapy/extensions/feedexport.py
```diff
@@ -367,16 +369,18 @@ def _exporter_supported(self, format):
...
-    def _settings_are_valid(self, uri):
+    def _settings_are_valid(self):
         """
-        If FEED_STORAGE_BATCH_ITEM_COUNT setting or FEEDS.batch_item_count is specified uri has to contain
+        If FEED_EXPORT_BATCH_ITEM_COUNT setting or FEEDS.batch_item_count is specified uri has to contain
         %(batch_time)s or %(batch_id)s to distinguish different files of partial output
         """
         for uri_template, values in self.feeds.items():
-            if values['batch_item_count'] and not any(s in uri_template for s in ['%(batch_time)s', '%(batch_id)s']):
+            if values['batch_item_count'] and not re.findall(r'(%\(batch_time\)s|(%\(batch_id\)0\d*d))', uri_template):
```
Gallaecio, line 378: What about just looking for `%(asdf)`, and hope that the user did not break the formatting syntax? Either we do that, or we implement something that finds all [syntax variants](https://docs.python.org/3/library/stdtypes.html#old-string-formatting).

Also, `findall` probably does not make sense here, given the preceding `not`; finding a single match should suffice in this scenario.

BroodingKangaroo: > What about just looking for %(asdf), and hope that the user did not break the formatting syntax? Either we do that, or we implement something that finds all syntax variants.

@Gallaecio, could you please explain it a little bit more because I can't clearly understand why should we allow users to break the formatting syntax if batch_id is always an integer and we only want to add leading zeros?

Gallaecio: My thinking was that we should allow as much flexibility as possible if it’s trivial to do. For example, if someone wants to use spaces instead of zeroes for padding, we would be allowing that (`%(batch_id)4d`).

Not a strong opinion about it, though.

scrapy/extensions/feedexport.py
```diff
@@ -375,9 +375,9 @@ def _settings_are_valid(self):
         %(batch_time)s or %(batch_id)s to distinguish different files of partial output
         """
         for uri_template, values in self.feeds.items():
-            if values['batch_item_count'] and not re.findall(r'(%\(batch_time\)s|(%\(batch_id\)0\d*d))', uri_template):
+            if values['batch_item_count'] and not re.search(r'%\(batch_time\)s|%\(batch_id\)', uri_template):
                 logger.error(
-                    '%(batch_time)s or %(batch_id)0xd must be in uri({}) if FEED_EXPORT_BATCH_ITEM_COUNT setting '
+                    '%(batch_time)s or %(batch_id) must be in uri({}) if FEED_EXPORT_BATCH_ITEM_COUNT setting '
```
Gallaecio, line 380: :lipstick: I would still describe it as `%(batch_id)d` in the message.

```suggestion
                    '%(batch_time)s or %(batch_id)d must be in the feed URI ({}) if FEED_EXPORT_BATCH_ITEM_COUNT setting '
```

tox.ini
```diff
@@ -12,6 +12,7 @@ deps =
     -ctests/constraints.txt
     -rtests/requirements-py3.txt
     # Extras
+    boto3>=1.13.0
```
elacuesta, line 15: Is this really needed to test the changes in this pull request?

BroodingKangaroo: Now it needs to perform `BatchDeliveriesTest::test_s3_export`. If we delete this line, the test will be skipped at every run.

Gallaecio: I think what @elacuesta means if whether or not `test_s3_export` could be reimplemented without adding a new dependency to the tests, which is a valid point. So, could it be reimplemented without boto3? And, if so, what would be the pros and cons of each test implementation?

docs/topics/feed-exports.rst
```diff
@@ -429,3 +431,45 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+  number by introducing leading zeroes as needed, use ``%(batch_id)05d``
+  (e.g. ``3`` becomes ``00003``, ``123`` becomes ``00123``).
+
+For instance, if your settings include::
+
+    FEED_EXPORT_BATCH_ITEM_COUNT=100
+
+And your :command:`crawl` command line is::
+
+    scrapy crawl spidername -o dirname/%(batch_id)d-filename%(batch_time)s.json
```
rafaelcapucho, line 464: @BroodingKangaroo Hi, I'm trying to run scrapy installed from your branch and I'm getting an error with the parentheses, following your docs example highlighted above:

```
$ scrapy crawl books -o books-%(batch_id)d-ending.json
bash: syntax error near unexpected token `('
```
It seems to accept the expression surrounded by quotes (as bellow), so I'd suggest to update the docs with it wrapped with quotes:
```
$ scrapy crawl books -o 'books-%(batch_id)d-ending.json'
```

rafaelcapucho:
@BroodingKangaroo I have added `FEED_EXPORTER_BATCH_ITEM_COUNT` to my [booksbot ](https://github.com/rafaelcapucho/booksbot) spider, as in:

```
(kangaroo-venv) kepler@virtualbox:~/kangaroo-venv/booksbot$ cat books/settings.py

BOT_NAME = 'books'

SPIDER_MODULES = ['books.spiders']
NEWSPIDER_MODULE = 'books.spiders'

ROBOTSTXT_OBEY = True
HTTPCACHE_ENABLED = True

FEED_EXPORTER_BATCH_ITEM_COUNT = 100
```

[booksbot ](https://github.com/rafaelcapucho/booksbot) always return 1000 scrapped items so I expect to end up with 10 files, right? When running the crawl as the following:

```
scrapy crawl books -o 'books-%(batch_id)d-ending.json'
```
I have the 1000 items founds, the it only generates a single `books-1-ending.json` file:

```
....
2020-07-14 02:15:10 [scrapy.core.engine] INFO: Closing spider (finished)
2020-07-14 02:15:10 [scrapy.extensions.feedexport] INFO: Stored json feed (1000 items) in: books-1-ending.json   <--- generated
2020-07-14 02:15:10 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'downloader/request_bytes': 353574,
 'downloader/request_count': 1051,
 'downloader/request_method_count/GET': 1051,
 'downloader/response_bytes': 4711804,
 'downloader/response_count': 1051,
 'downloader/response_status_count/200': 1050,
 'downloader/response_status_count/404': 1,
 'elapsed_time_seconds': 11.020807,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2020, 7, 14, 5, 15, 10, 172239),
 'httpcache/hit': 1051,
 'item_scraped_count': 1000,           <---- items found
 'log_count/DEBUG': 2052,
....
```

Am I missing something? Thx

rafaelcapucho:
@BroodingKangaroo Hi, I have `FEED_EXPORT_BATCH_ITEM_COUNT = 100` and I scrapped 1000 items:

image

I was expecting to see 10 files with 100 items per file, but I ended up with 11 files:

image

and the last one is empty:

image

rafaelcapucho:
I'm happy to see that it is also working to export S3:

image

docs/topics/feed-exports.rst
BroodingKangaroo: You are right, I will update the docs.
I should say that this also happens in the master branch with placeholders like `%(time)s`, `%(name)s` and I didn't find information about quotes in such cases in docs.

BroodingKangaroo:
> I was expecting to see 10 files with 100 items per file, but I ended up with 11 files:
Hi, @rafaelcapucho. It is similar to creating an empty file even if nothing is scrapped (current behavior of the master branch). On the other hand, in this case it is maybe less relevant to user expectations.

tox.ini
BroodingKangaroo: Using botocore (we already have this dependency) here means that we have to basically re-implement boto3 anyway, including authentication, all security protocols, exceptions. Also, we have to maintain it if S3 API changes.

docs/topics/feed-exports.rst
```diff
@@ -461,7 +461,7 @@ For instance, if your settings include::
 
 And your :command:`crawl` command line is::
 
-    scrapy crawl spidername -o dirname/%(batch_id)d-filename%(batch_time)s.json
+    scrapy crawl spidername -o 'dirname/%(batch_id)d-filename%(batch_time)s.json'
```
Gallaecio, line 464: I _think_ in Windows `cmd.exe` single quotes won’t work. See https://docs.scrapy.org/en/latest/intro/tutorial.html#extracting-data

tests/test_feedexport.py
```diff
@@ -1163,3 +1180,371 @@ def test_multiple_feeds_failing_logs_blocking_feed_storage(self):
...
+        content = defaultdict(list)
+        try:
+            with MockServer() as s:
+                runner = CrawlerRunner(Settings(settings))
+                spider_cls.start_urls = [s.url('/')]
+                yield runner.crawl(spider_cls)
+
+            for path, feed in FEEDS.items():
+                dir_name = os.path.dirname(path)
+                for file in sorted(os.listdir(dir_name)):
```
kmike, line 1207: It looks like tests fail here on Windows: https://scrapinghub.visualstudio.com/scrapy/_build/results?buildId=98&view=logs&j=76ed1102-9071-5354-720f-6f09bbc9f0c0&t=dc276db5-d83b-509a-4a8d-770723557532&l=2129

docs/topics/feed-exports.rst
```diff
@@ -446,6 +448,50 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
         'csv': None,
     }
 
+
+.. setting:: FEED_EXPORT_BATCH_ITEM_COUNT
+
+FEED_EXPORT_BATCH_ITEM_COUNT
+-----------------------------
+Default: ``0``
```
Gallaecio, line 456:
```suggestion
FEED_EXPORT_BATCH_ITEM_COUNT
----------------------------

Default: ``0``
```

docs/topics/feed-exports.rst
```diff
@@ -446,6 +448,50 @@ format in :setting:`FEED_EXPORTERS`. E.g., to disable the built-in CSV exporter
...
+* ``%(batch_id)d`` - gets replaced by the sequence number of the batch.
+
+  Use :ref:`printf-style string formatting <python:old-string-formatting>` to
+  alter the number format. For example, to make the batch ID a 5-digit
+  number by introducing leading zeroes as needed, use ``%(batch_id)05d``
+  (e.g. ``3`` becomes ``00003``, ``123`` becomes ``00123``).
+
+For instance, if your settings include::
+
+    FEED_EXPORT_BATCH_ITEM_COUNT=100
```
Gallaecio, line 477: PEP8

```suggestion
    FEED_EXPORT_BATCH_ITEM_COUNT = 100
```

kmike:
Thanks @BroodingKangaroo for all the work you put into this feature, and also for swiftly addressing all the review comments - that was impressive! An of course, thanks everyone who reviewed and tested it: @ejulio, @elacuesta, @Gallaecio, @victor-torres, @rafaelcapucho. 

I'm merging it; if there are any further minor issues, we can address them separately.

PR #4438, opened 2020-03-17
edit zsh completion

seregaxvm:
- Fix bug introduced in https://github.com/scrapy/scrapy/pull/4291
- Enforce `[command] [options] [arguments]` syntax. Do not allow options after arguments
- Exclude already used option aliases from completion list


kmike:
Thanks @seregaxvm! Merging it without checking how it works :)

PR #4439, opened 2020-03-17
extract domain from cli genspider URL

matsievskiysv:
This PR modifies behavior of cli `genspider` command to allow passing not only domain names, but also URLs. 
For example, `https://test.com/`, `http://test.com/`, `https://test.com/other` and `test.com/` arguments would all produce spider with `allowed_domains = ['test.com']` and `start_urls = ['http://test.com/']`

This is particularly useful for zsh users because autocompletion prepends schema to domain argument.


kmike:
I think that's a great idea. 

Could you please add tests?

seregaxvm:
@kmike added some tests

seregaxvm:
Sorry, didn't notice failed tests until now.

tests/test_commands.py
```diff
@@ -79,6 +80,15 @@ def kill_proc():
 
         return p, to_unicode(stdout), to_unicode(stderr)
 
+    def find_in_file(self, filename, regex):
+        '''Find first pattern occurrence in file'''
```
Gallaecio:
```suggestion
        """Find first pattern occurrence in file"""
```

:lipstick: I think this is the only area were PEP8 actually makes a choice between `'` and `"` :slightly_smiling_face:

tests/test_commands.py
```diff
@@ -5,6 +5,7 @@
 import subprocess
 import sys
 import tempfile
+import re
```
Gallaecio: :lipstick: Let’s keep the imports sorted alphabetically.

```suggestion
import re
import subprocess
import sys
import tempfile
```

wRAR (approved):

Gallaecio (approved):

Gallaecio:
Great, thanks!

wRAR:
Should we now close various other issues/PRs about passing an URL to genspider?

PR #4440, opened 2020-03-18
Remove assertions from production code

elacuesta:
Inspired by the following comment by @Gallaecio: https://github.com/scrapy/scrapy/pull/4054#discussion_r351514071

From [the docs about the `assert` statement](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement):
> The current code generator emits no code for an assert statement when optimization is requested at compile time

Please pay attention to the specific exception classes, there might be better ones in some cases.


scrapy/core/engine.py
```diff
@@ -165,7 +167,10 @@ def _next_request_from_scheduler(self, spider):
         return d
 
     def _handle_downloader_output(self, response, request, spider):
-        assert isinstance(response, (Request, Response, Failure)), response
+        if not isinstance(response, (Request, Response, Failure)):
+            raise TypeError(
+                "Incorrect type: expected Request, Response or Failure, got %s" % type(response)
```
Gallaecio: It may be good to also include `repr(response)` in the message (e.g. appending `: %r` to the message and switching `type(response)` to `(type(response), response)`), to make it easier to identify which response triggered the error. Same in similar scenarios in this patch.

elacuesta:
I wonder why there is no Travis build for the latest commit :thinking:

Gallaecio:
[There is](https://travis-ci.org/github/scrapy/scrapy/builds/668922915). No idea why it does not show up in the pull request any more, though.

kmike:
Exceptions makes sense, thanks @elacuesta! 
I added "backwards-incompatible" tag, because technically this changes the behavior.

PR #4446, opened 2020-03-20
Change defer.returnValue calls

### LINKED ISSUE #4443: Remove defer.returnValue calls
https://twisted.readthedocs.io/en/latest/core/howto/defer-intro.html#inline-callbacks-using-yield says "On Python 3, instead of writing returnValue(json.loads(responseBody)) you can instead write return json.loads(responseBody). This can be a significant readability advantage, but unfortunately if you need compatibility with Python 2, this isn’t an option.".

aditi137:
Changed all `defer.returnValue` calls in code and updated comments.
Resolves #4443

### DIFF
--- a/scrapy/core/downloader/middleware.py
+++ b/scrapy/core/downloader/middleware.py
@@ -38,14 +38,14 @@ class DownloaderMiddlewareManager(MiddlewareManager):
                     raise _InvalidOutput('Middleware %s.process_request must return None, Response or Request, got %s' % \
                                          (method.__self__.__class__.__name__, response.__class__.__name__))
                 if response:
-                    defer.returnValue(response)
-            defer.returnValue((yield download_func(request=request, spider=spider)))
+                    return response
+            yield download_func(request=request, spider=spider)
 
         @defer.inlineCallbacks
         def process_response(response):
             assert response is not None, 'Received None in process_response'
             if isinstance(response, Request):
-                defer.returnValue(response)
+                return response
 
             for method in self.methods['process_response']:
                 response = yield deferred_from_coro(method(request=request, response=response, spider=spider))
@@ -53,8 +53,8 @@ class DownloaderMiddlewareManager(MiddlewareManager):
                     raise _InvalidOutput('Middleware %s.process_response must return Response or Request, got %s' % \
                                          (method.__self__.__class__.__name__, type(response)))
                 if isinstance(response, Request):
-                    defer.returnValue(response)
-            defer.returnValue(response)
+                    return response
+            return response
 
         @defer.inlineCallbacks
         def process_exception(_failure):
@@ -65,8 +65,8 @@ class DownloaderMiddlewareManager(MiddlewareManager):
                     raise _InvalidOutput('Middleware %s.process_exception must return None, Response or Request, got %s' % \
                                          (method.__self__.__class__.__name__, type(response)))
                 if response:
-                    defer.returnValue(response)
-            defer.returnValue(_failure)
+                    return response
+            return _failure
 
         deferred = mustbe_deferred(process_request, request)
         deferred.addErrback(process_exception)
--- a/tests/test_pipeline_media.py
+++ b/tests/test_pipeline_media.py
@@ -2,7 +2,7 @@ from testfixtures import LogCapture
 from twisted.trial import unittest
 from twisted.python.failure import Failure
 from twisted.internet import reactor
-from twisted.internet.defer import Deferred, inlineCallbacks, returnValue
+from twisted.internet.defer import Deferred, inlineCallbacks
 
 from scrapy.http import Request, Response
 from scrapy.settings import Settings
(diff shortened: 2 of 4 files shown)


wRAR:
Please run the tests locally before submitting the PR. As you can see on Travis, some tests failed, so some of your changes broke something.

scrapy/core/downloader/middleware.py
```diff
@@ -38,23 +38,23 @@ def process_request(request):
                     raise _InvalidOutput('Middleware %s.process_request must return None, Response or Request, got %s' % \
                                          (method.__self__.__class__.__name__, response.__class__.__name__))
                 if response:
-                    defer.returnValue(response)
-            defer.returnValue((yield download_func(request=request, spider=spider)))
+                    return response
+            yield download_func(request=request, spider=spider)
```
wRAR: The value is not returned here anymore

tests/test_pipeline_media.py
```diff
@@ -124,8 +124,8 @@ def test_should_remove_req_res_references_before_caching_the_results(self):
         # Simulate the Media Pipeline behavior to produce a Twisted Failure
         try:
             # Simulate a Twisted inline callback returning a Response
-            # The returnValue method raises an exception encapsulating the value
-            returnValue(response)
+            # The returned Response raises an exception
```
wRAR: It doesn't. `defer.returnValue` indeed raises an exception so it should either keep the test as is or change it, this depends on the test purpose. It may be related to the comment in MediaPipeline, or maybe not.

tests/test_pipeline_media.py
```diff
@@ -140,7 +140,7 @@ def test_should_remove_req_res_references_before_caching_the_results(self):
 
         # The Failure should encapsulate a FileException ...
         self.assertEqual(failure.value, file_exc)
-        # ... and it should have the returnValue exception set as its context
+        # ... and it should have the returned exception set as its context
```
wRAR: Same here (and in the similar comment below).

tests/test_spidermiddleware_output_chain.py
```diff
@@ -292,7 +292,7 @@ def crawl_log(self, spider):
         crawler = get_crawler(spider)
         with LogCapture() as log:
             yield crawler.crawl(mockserver=self.mockserver)
-        raise defer.returnValue(log)
```
wRAR: I suspect the raise expression here was a mistake and it was ignored, so this should be rewritten as `return log`.

tests/test_pipeline_media.py
elacuesta: +1 for rewriting the test. I think it'd be good to clean [this block](https://github.com/scrapy/scrapy/blob/2.0.1/scrapy/pipelines/media.py#L141-L162) up, since the mentioned `_DefGen_Return` exception would not be in play anymore.

scrapy/core/downloader/middleware.py
joybhallaa: @wRAR should `return` be used here?
As the function is ending there, there is no need to "save" the state. which yield is useful for

tests/test_spidermiddleware_output_chain.py
joybhallaa: @wRAR  return log is leading to a lot of errors and test failures in my local machine.

scrapy/core/downloader/middleware.py
wRAR: Not sure what do you mean by "saving the state" but the function must return the result of that `yield`, just like it did before. `yield` here is used to get the result of `download_func` (because of `@defer.inlineCallbacks`).

tests/test_spidermiddleware_output_chain.py
wRAR: @joybhallaa it doesn't on mine.

scrapy/core/downloader/middleware.py
joybhallaa: @wRAR okay :+1: 

I'm still a little bit confused by  `The value is not returned here anymore` 
What exactly is it not returning ?

tests/test_spidermiddleware_output_chain.py
joybhallaa: That's cool, just thought I'd let you know.

scrapy/core/downloader/middleware.py
wRAR: It is not returning the result of `download_func`.

wRAR: (which was previously returned via `defer.returnValue` as you can see)

tests/test_spidermiddleware_output_chain.py
wRAR: @joybhallaa that probably means you changed something else.

scrapy/core/downloader/middleware.py
joybhallaa: Okay :+1:  I'm thinking of picking up this issue if the original author is not working on it.

Issue #4240, opened 2019-12-17
Wrong response type for binary responses


ejulio:
### Description

Binary (file) responses are identified as `TextResponse` instead of a plain `Response` in `spider.parse`.

### Steps to Reproduce

From this URL https://www.ecb.europa.eu/press/pr/date/2004/html/pr040702.en.html
We can get this link https://www.ecb.europa.eu/pub/redirect/pub_5874_en.html (text is `pdf 1692 kB`).
It redirects to https://www.ecb.europa.eu/pub/pdf/other/developmentstatisticsemu200406en.pdf

**Expected behavior:** [What you expect to happen]

In the spider, `isinstance(response, TextResponse)` should be `False`.

**Actual behavior:** [What actually happens]

In the spider, `isinstance(response, TextResponse)` is `True`, even though the `Content-Type` header is `application/pdf`.

### Versions

```
Scrapy       : 1.7.3
lxml         : 4.4.2.0
libxml2      : 2.9.9
cssselect    : 1.1.0
parsel       : 1.5.2
w3lib        : 1.21.0
Twisted      : 19.10.0
Python       : 3.6.8 (default, Jan 14 2019, 11:02:34) - [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1d  10 Sep 2019)
cryptography : 2.8
Platform     : Linux-4.15.0-72-generic-x86_64-with-Ubuntu-18.04-bionic
```

### Extra info

Probably we should handle known binary types in `Content-Type` here https://github.com/scrapy/scrapy/blob/master/scrapy/responsetypes.py
At least, `application/pdf` should be in the list.
It would be nice to add some mechanism to allow developers to extend the mapping on the fly, as we can get new types in a project basis and it would be easier update the desired behavior.

GeorgeA92:
@ejulio, @Gallaecio 
`ScrapyAgent._cb_bodydone` method.. chooses response class:
https://github.com/scrapy/scrapy/blob/e22a8c8c36e34ffaf12ef9e330624df654582605/scrapy/core/downloader/handlers/http11.py#L395-L400
`responsetypes.from_args` method performs several checks with following logic:

- default response type is plain `Response`,
- if scrapy identify something that require another response class instead if plain `Response` - another response class will be returned ignoring results of previous checks.

[...]

PR #4449, opened 2020-03-23
[GSOC2020] Pre-application PR

BTghub:
Hi @Gallaecio and @kmike !
I'm Brett, a computer science student at Ryerson University. I'll be entering my final year this fall and wanted to take part in GSOC as a way to improve my coding while contributing to a large scale project. 

I'm interested in working on the MIME Sniffing Library and was hoping I could receive some feedback on my application. Link: https://docs.google.com/document/d/1SfyFgSkzr-giJLze33iN75UZVvU6aFnz4T1jxV24PGY/edit?usp=sharing

I've also put together some changes for you to look at.  I took a shot at modifiying the [mimesniff](https://github.com/corona10/mimesniff) library to be able to use it within responsetypes.py.  A question I had, as I was putting this together, is wether or not this is the change you guys had in mind or were you interested in a complete overhaul of responsetypes.py.

### DIFF
--- a/scrapy/mimesniff.py
+++ b/scrapy/mimesniff.py
@@ -1,0 +1,184 @@
+# Original code created by Dong-hee Na, see https://github.com/corona10/mimesniff
+# Modified for use with scrapy
+# Date: March 22. 2020
+# By: Bretton Tan
+
+import io
+
+def from_file_content(fin):
+    with open(fin, 'rb') as f:
+        h = f.read(512)
+        res = _detect_content(h)
+        if res:
+            return res
+    return 'application/octet-stream'
+
+def from_byte_content(fin):
+    res = _detect_content(fin)
+    if res:
+        return res
+    return 'application/octet-stream'
+
+def from_stream_content(fin):
+    location = fin.tell()
+    h = fin.read(512)
+    fin.seek(location)
+    res = _detect_content(h)
+    if res:
+        return res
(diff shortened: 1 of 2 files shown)


Gallaecio:
Make sure you read https://github.com/scrapy/scrapy/issues/4240.

I think it would be good for your proposal to explain its choices. For example:
- Why copy-pasting from mimesniff into Scrapy, instead of contributing changes upstream to that project, or creating a separate project?
- Have you considered using libmagic?
- What’s the reasoning behind how you decided to group the different parts of the specs during development?

BTghub:
Hi @Gallaecio ,
I've gone ahead and made some changes to my proposal.  I made sure to go through the discussion in #4240 and incorporate the ideas being talked about there.  If you wouldn't mind taking another look, I would be happy to get some more feedback.

Gallaecio:
I like the changes.

One thing you may want to check now, as it may affect how you propose to implement some parts of the specification, is performance.

In cases that could be implemented with libmagic Python bindings, how does performance compare between that and the mimesniff library approach? A slight difference either way should not matter much, but if this check is happening on every response, performance could be an issue.

Also, if the current (faulty) Scrapy implementation is significantly faster than the new implementation, it may make sense to allow users to choose which implementation to use through a setting, similar to how we support different robots.txt parsers.

Issue #4336, opened 2020-02-17
Needs change on "Example of shell session" in Scrapy 1.8.0 docs


dedenhabibi:
### Description

I was learning how to use Scrapy shell but got error similar with this issue #3314, and got the solution in the issue as well. But, when I looked back into the Docs (1.8.0), the example still use (') instead of ("). I think it is better to change it, especially for next learners like me.

**Expected behavior:**

*Example of shell session*
...
scrapy shell "https://scrapy.org" --nolog

**Actual behavior:**

*Example of shell session*
...
scrapy shell 'https://scrapy.org' --nolog

Gallaecio:
Could you share the URL that caused you issues with `'`?

At least in Linux, I think `'` is safer than  `"`, since within `"` special characters like ` or $ could be unexpectedly interpreted.

Gallaecio:
In https://github.com/scrapy/scrapy/issues/3314 I think the issue was not using `'` instead of `"`, but missing the URL protocol.

dedenhabibi:
> 
> 
> Could you share the URL that caused you issues with `'`?
> 
> At least in Linux, I think `'` is safer than `"`, since within `"` special characters like `or`$` could be unexpectedly interpreted.

I use Windows 10 (unfortunately :D).
Any URL doesnt work, e.g. `https://www.lapan.go.id.`
It shows `ValueError: invalid hostname: 'https`.

Perhaps, the Windows? I agree with you about `"` vulnerabilities, but it just doesnt work.

Gallaecio:
OK, I think at the very least we could warn about the need to use quoting, and system gotchas related to that.

dedenhabibi:
Ahh, I see. Thank you for illuminating me 😄 🙏

gigatesseract:
@Gallaecio 
It does not work with single quotes in Windows right?
 So we can warn about the usage of ```"``` in Windows and ask them to use ``` ' ``` in Linux based systems.
It is a simple documentation text, I will work on it if that's okay.

Gallaecio:
I was translating the tutorial the other day, and I noticed we do mention this there: https://docs.scrapy.org/en/latest/intro/tutorial.html#extracting-data

We could add a similar warning to the reference documentation.

sakshamb2113:
I am working on this.

Gallaecio:
Fixed by #4450

oakla:
[...]

PR #4450, opened 2020-03-23
Added warning to use double quotes in Windows for scrapy shell in shell.rst

sakshamb2113:


sakshamb2113:
Updated shell.rst to include a warning (which is similar to the one used in tutorial.rst)  for using double quotes while using scrapy shell command in Windows. (fixes #4336)

docs/topics/shell.rst
```diff
@@ -156,6 +156,17 @@ First, we launch the shell::
 
     scrapy shell 'https://scrapy.org' --nolog
 
+.. note::
+
+   Remember to always enclose urls in quotes when running Scrapy shell from
+   command-line, otherwise urls containing arguments (i.e. ``&`` character)
+   will not work.
```
Gallaecio:
```suggestion
   Remember to always enclose URLs in quotes when running the Scrapy shell from
   the command line, otherwise URLs containing arguments (i.e. the ``&`` character)
   will not work.
```

kmike:
Makes sense, thanks @sakshamb2113 and @Gallaecio!

PR #4456, opened 2020-03-26
Project URLs in machine-readable format for showing in pypi

whalebot-helmsman:
Compare "Project Links" section from [scrapy pypi page](https://pypi.org/project/Scrapy/) with [one for doit](https://pypi.org/project/doit/)


kmike:
Makes sense, thanks @whalebot-helmsman!

PR #4458, opened 2020-03-27
Set up CI with Azure Pipelines

dangra:


kmike:
I'm not sure I know what I'm doing, but opened https://github.com/cjw296/sybil/issues/20 :)

Gallaecio:
The errors look different, I guess that’s an improvement.

Gallaecio:
https://github.com/scrapy/scrapy/pull/4458/commits/2f510fd47d8d10217f7b18b531205dd8c252eaef fixed 1 test in Python 3.7
https://github.com/scrapy/scrapy/pull/4458/commits/94ee68695a42ec8d3ccdff71bfc2fa33d5ea7049 fixed 3 tests in Python 3.7

Gallaecio:
@elacuesta [test_ipv6_alternative_name_resolver](https://github.com/scrapy/scrapy/blame/master/tests/test_crawler.py#L309) fails on Windows; it looks like it does not log any error there, and it just successfully crawls the page. Any idea?

elacuesta:
> @elacuesta [test_ipv6_alternative_name_resolver](https://github.com/scrapy/scrapy/blame/master/tests/test_crawler.py#L309) fails on Windows; it looks like it does not log any error there, and it just successfully crawls the page. Any idea?
> 
> PS: That is now the only test failing in Python 3.7.

[These are the failing assertions](https://github.com/scrapy/scrapy/blob/2.0.1/tests/test_crawler.py#L327-L334). I ran a [debug pipeline](https://scrapinghub.visualstudio.com/scrapy/_build/results?buildId=23&view=results) (I don't have access to a Windows machine) and it seems to me like it fails because of the assumption that there is no web server listening on the local host. Apparently there is something in the test environment indeed [listening on ` http://[::1]`](https://scrapinghub.visualstudio.com/scrapy/_build/results?buildId=23&view=logs&j=00b6a206-eb91-585b-3d14-cf1a4d7b1970&t=d059ca3f-6e18-5882-c1cf-8249a45638a0&l=651) so the test fails because the connection is not refused.
I think the test should not care about the presence of `ConnectionRefusedError` exceptions but about the absensce of `DNSLookupError`, I can fix that :+1:

elacuesta:
Job successful for py37: https://scrapinghub.visualstudio.com/scrapy/_build/results?buildId=25&view=logs&j=00b6a206-eb91-585b-3d14-cf1a4d7b1970

Gallaecio:
Weird, now `test_response_ip_address` fails with `[…] FailTest: 'INFO: Host: not.a.real.domain' not in […]`.

Gallaecio:
Finally! :sweat:

tests/CrawlerRunner/ip_address.py
```diff
@@ -10,6 +12,17 @@
 from tests.mockserver import MockServer, MockDNSServer
 
 
+# https://stackoverflow.com/a/32784190
+def createResolver(servers=None, resolvconf=None, hosts=None):
+    if hosts is None:
+        hosts = (b'/etc/hosts' if platform.getType() == 'posix'
+                 else r'c:\windows\hosts')
```
elacuesta, line 19:
```suggestion
        hosts = b'/etc/hosts' if platform.getType() == 'posix' else r'c:\windows\hosts'
```
Nitpick 1

tests/CrawlerRunner/ip_address.py
```diff
@@ -10,6 +12,17 @@
...
+    theResolver = Resolver(resolvconf, servers)
+    hostResolver = hostsModule.Resolver(hosts)
+    L = [hostResolver, cache.CacheResolver(), theResolver]
+    return resolve.ResolverChain(L)
```
elacuesta, line 23:
```suggestion
    chain = [hostResolver, cache.CacheResolver(), theResolver]
    return resolve.ResolverChain(chain)
```
Nitpick 2

scrapy/utils/url.py
```diff
@@ -98,7 +96,10 @@ def guess_scheme(url):
                     )?      # optional match of ".", ".." or ".blabla"
                     /       # at least one "/" for a file path,
                     .       # and something after the "/"
-                    ''', parts.path, flags=re.VERBOSE):
+                    ''', url, flags=re.VERBOSE):
+        return any_to_uri(url)
+    # Windows drive-letter path
+    elif re.match(r'''^[a-z]:\\''', url, flags=re.IGNORECASE):
```
kmike, line 102: could you please add tests for these changes? I also wonder if these two conditions can be moved to helper functions, looks_like_posix_path and looks_like_windows_path; it may allow easier testing.

tests/CrawlerRunner/ip_address.py
```diff
@@ -10,6 +12,16 @@
 from tests.mockserver import MockServer, MockDNSServer
 
 
+# https://stackoverflow.com/a/32784190
+def createResolver(servers=None, resolvconf=None, hosts=None):
+    if hosts is None:
+        hosts = b'/etc/hosts' if platform.getType() == 'posix' else r'c:\windows\hosts'
```
wRAR, line 18: *Ideally* this should detect the Windows dir path at run time

Gallaecio: Indeed. However, what we really care for here is `servers` getting picked up. The code for `hosts` is just so that we have a hosts value to pass to `hostsModule.Resolver`.

PR #4459, opened 2020-03-27
Added PYTHONOPTIMIZE

### LINKED ISSUE #4454: Add a CI job with PYTHONOPTIMIZE=2
That way we make sure we do not introduce code that relies on asserts, `__debug__` or docstrings to work, allowing people to use Scrapy when using Python in optimized mode.

joybh98:
1.Added CI job with PYTHONOPTIMIZE
2. Added a `testenv` which exports `PYTHONOPTIMIZE=2`
3. All tests passed
4. Fixed #4454 

I think my approach is wrong as I've just set an environment variable, I thought I should open a PR so people can weigh in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ matrix:
       python: 3.8
     - env: TOXENV=docs
       python: 3.7  # Keep in sync with .readthedocs.yml
+    - env: TOXENV=py-optimize
+      python: 3.5
 install:
   - |
       if [ "$TOXENV" = "pypy3" ]; then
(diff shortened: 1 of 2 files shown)


joybhallaa:
@Gallaecio only changing the `travis.yml` is not breaking the tests in any way, but I think we can use a  different approach. 
We can add the flag `-00` in each test as it is basically doing the same thing.
Thoughts ?

Gallaecio:
Let’s get an approach that works. Once we have that, and we know that it works, we can decide if we want to explore alternative approaches.

Gallaecio:
https://github.com/scrapy/scrapy/pull/4725#discussion_r477513542

PR #4460, opened 2020-03-27
Add json response

### LINKED ISSUE #2444: response.json()?
python-requests have response.json() that decodes json body and returns appropriate Python objects. Does it make sense to have something like this in Scrapy?

JochenFromm:
My first PR for Scrapy that is supposed to fix GH https://github.com/scrapy/scrapy/issues/2444

Adds a `JsonResponse` class and a `json()` function similar to the one for the "requests" library [here](https://github.com/psf/requests/blob/master/requests/models.py#L874-L898)

Test manually by 
```
# from the Scrapy package itself install Scrapy first  
python setup.py install

# then run the Scrapy shell
scrapy shell https://api.github.com/events
>> response.json
```

Fixes #2444


tests/test_http_response.py
```diff
@@ -343,6 +343,21 @@ def test_unicode_body(self):
         self.assertTrue(isinstance(r1.text, str))
         self.assertEqual(r1.text, unicode_string)
 
+    def test_json(self):
+        body = u"[{\"id\": \"1234\", \"type\": \"event\"}]"
```
elacuesta:
```suggestion
        body = """[{"id": "1234", "type": "event"}]"""
```
A little bit cleaner this way

tests/test_http_response.py
```diff
@@ -699,6 +714,30 @@ def test_html5_meta_charset(self):
         self._assert_response_values(r1, 'gb2312', body)
 
 
+class JsonResponseTest(TextResponseTest):
+
+    response_class = JsonResponse
+
+    def test_js_encoding(self):
+        body = b"[{\"id\": \"1234\", \"type\": \"event\"}]"
```
elacuesta:
```suggestion
        body = """[{"id": "1234", "type": "event"}]"""
```
A little bit cleaner this way

JochenFromm:
thanks for the good review @elacuesta , feedback has been applied

docs/topics/request-response.rst
```diff
@@ -841,6 +841,16 @@ HtmlResponse objects
 
 .. _meta http-equiv: https://www.w3schools.com/TAGS/att_meta_http_equiv.asp
 
+JsonResponse objects
+-------------------
```
Gallaecio:
```suggestion
JsonResponse objects
--------------------
```

docs/topics/request-response.rst
```diff
@@ -841,6 +841,16 @@ HtmlResponse objects
...
+
+.. class:: JsonResponse(url[, ...])
+
+    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
+    adds a function json() that decodes the Json body and returns
```
Gallaecio:
```suggestion
    adds a :func:`json` function that decodes the JSON body and returns
```

The `json()` method should be documented below as well.

scrapy/http/response/json.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+
+import json
+from scrapy.http.response.text import TextResponse
+
+
+class JsonResponse(TextResponse):
+
+    def json(self):
+        """Return body as JSON"""
```
Gallaecio:
```suggestion
        """Returns the JSON-encoded body deserialized into a Python object"""
```

docs/topics/request-response.rst
JochenFromm: This should fix the Travis error 👍

scrapy/http/response/json.py
```diff
@@ -0,0 +1,16 @@
...
+See documentation in docs/topics/request-response.rst
+"""
+
+import json
+from scrapy.http.response.text import TextResponse
+
+
+class JsonResponse(TextResponse):
+
+    def json(self):
```
elacuesta: Should we be storing the results of the `json.loads` call, to avoid performing the same operation multiple times? Something like
```python
try:
    return self._json
except AttributeError:
    self._json = json.loads(self.text)
    return self._json
```
or maybe
```python
if not hasattr(self, "_json"):
    self._json = json.loads(self.text)
return self._json
```

elacuesta: Also, I wonder if it would be appropriate to do some encoding handling. For instance, see `python-requests`'s implementation: https://github.com/psf/requests/blob/v2.23.0/requests/models.py#L874-L898

scrapy/http/response/json.py
```diff
@@ -0,0 +1,16 @@
+"""
+This module implements the JsonResponse class which adds encoding
+discovering through JSON encoding declarations to the TextResponse class.
+
+See documentation in docs/topics/request-response.rst
+"""
+
+import json
```
elacuesta: Could you please add a blank line between these imports, as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports)?

scrapy/http/response/json.py
JochenFromm: Or we can use the following check like the `text()` method in `TextResponse` :
```
if self._json is None:
    self._json = json.loads(self.text)
return self._json
```

JochenFromm: Good point. As far as I understand the code encoding is already handled in the base class `TextResponse`.

docs/topics/request-response.rst
```diff
@@ -841,6 +841,16 @@ HtmlResponse objects
...
 
+JsonResponse objects
+--------------------
+
+.. class:: JsonResponse(url[, ...])
+
+    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
+    adds a :func:`json` function that decodes the JSON body and returns
+    appropriate Python objects.
+
```
Gallaecio: You are still missing the method documentation here. Add `.. automethod:: json` indented 1 level deeper than `.. class`.

docs/topics/request-response.rst
```diff
@@ -841,6 +841,18 @@ HtmlResponse objects
 
 .. _meta http-equiv: https://www.w3schools.com/TAGS/att_meta_http_equiv.asp
 
+JsonResponse objects
+--------------------
+
+.. class:: JsonResponse(url[, ...])
+
+    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
+    adds a :func:`json` function that decodes the JSON body and returns
```
Gallaecio:
```suggestion
    adds a :func:`json` method that decodes the JSON body and returns
```

Gallaecio: Have you rendered the documentation locally? (`tox -e docs`) Are all links working as expected?

JochenFromm: Yes, I have used `make html` and `make htmlview` in the "docs" folder as the Readme says. To me it looks good.

scrapy/http/response/json.py
```diff
@@ -0,0 +1,23 @@
...
+class JsonResponse(TextResponse):
+
+    def __init__(self, *args, **kwargs):
+        self._json = None
+        super(JsonResponse, self).__init__(*args, **kwargs)
+
+    def json(self):
+        """Returns the JSON-encoded body deserialized into a Python object"""
+        if self._json is None:
+            self._json = json.loads(self.text)
```
Gallaecio: :lipstick:

Because `null` is a valid JSON document that would be translated into Python as `None`, such documents would always be reloaded.

It is not that much of a issue, but it may be better to use a custom `object()` instance (e.g. define `_NONE = object()` or similar outside the class, and use `_NONE` instead of `None` as default value for `_json` and for the `if` statement). This approach is a common pattern in Python and it has a name, but I cannot remember it at the moment :man_facepalming:

scrapy/http/response/json.py
elacuesta: Right, but AFAICT that handles the body conversion from bytes to unicode, I wonder if there are some considerations to be made when decoding the unicode body to JSON. Perhaps there is actually nothing to be done in this regard, in any case we can always fix any issues later.

elacuesta: Nicely done, I wasn't aware of that. Can I request a last small adjustment, to keep things consistent? Could you rename the attribute to something like `_cached_decoded_json`?

scrapy/responsetypes.py
```diff
@@ -21,8 +21,8 @@ class ResponseTypes:
         'application/xhtml+xml': 'scrapy.http.HtmlResponse',
         'application/vnd.wap.xhtml+xml': 'scrapy.http.HtmlResponse',
         'application/xml': 'scrapy.http.XmlResponse',
-        'application/json': 'scrapy.http.TextResponse',
-        'application/x-json': 'scrapy.http.TextResponse',
+        'application/json': 'scrapy.http.JsonResponse',
+        'application/x-json': 'scrapy.http.JsonResponse',
```
kmike, line 25: Actually I wonder if it is better to add .json() method to TextResponse, which is contrary to what's been discussed here before. A problem with using HTTP headers is that in a real world not all APIs would return a proper Content-Type header, and .json() method wouldn't be available for them. What's the reason to only providing .json() for responses we're sure are json?

PR #4461, opened 2020-03-28
Require sybil ≥ 1.3.0 for tests

Gallaecio:
Next step after https://github.com/cjw296/sybil/issues/20 to pass the first issues found in https://github.com/scrapy/scrapy/pull/4458


kmike:
I wonder if it is better to merge it to azure-pipelines branch first, to ensure it fixes the problem. What do you think @Gallaecio?

Gallaecio:
Let’s keep things at #4458.

PR #4462, opened 2020-03-29
Feature/autothrottle 429

### LINKED ISSUE #4424: Adjust throttling for 429 response codes
HTTP 429 response code is returned when we reach a rate limit for an API at a given time.
Usually, it is a matter of waiting a bit to start sending new requests.
The "problem" is that, if concurrency settings are greater than the allowed number of requests from the API, we'll always get 429s.

A solution would be to tune throttling so it delays requests based on 429s.
It can be a extension/middleware, as `AutoThrottle` seems quite specific for throttle control over latency.

Also it could be worth considering that some APIs return the waiting time https://github.com/scrapy/scrapy/issues/3849

Here is a previous PR for this https://github.com/scrapy/scrapy/pull/3061

vortexkd:
I've got a draft PR for #4424 .

The main idea is to try to estimate both the interval the API blocks requests at and the number of requests that trigger the limit.

There are 4 common intervals used for api rate limits that are considered here. Per second, Per Minute, Per Hour, Per Day (per month limits are unlikely to be fixed at the throttler level).

In implementing these there are 2 possible strategies that can be used to measure the time.

Fixed interval measurement. -> N requests per minute is measured for each minute, so theoretically you could have up to 2N requests within a 60 second interval without bad responses, if you started requests at 30s past the minute mark.
Measurement from first request time. -> I don't think this is very common, but the idea is that the clock starts on the first request, and from that time a time interval T is measured within which a maximum of N requests can be made.

This is still a work in progress, but do give me any feedback.

I'm also making the uncomfortable assumption that a rate limiting entity is bound to a domain
which is not actually true, (example.com/names and example.com/places may have different rate limits), 
Also as mentioned by @Gallaecio this is not going to work optimally when multiple slots are used on the same api / domain because i'm setting the delay at the slot level.

Fixes #4424.

### DIFF
--- a/scrapy/core/downloader/webclient.py
+++ b/scrapy/core/downloader/webclient.py
@@ -141,6 +141,7 @@ class ScrapyHTTPClientFactory(HTTPClientFactory):
 
     def _build_response(self, body, request):
         request.meta['download_latency'] = self.headers_time - self.start_time
+        request.meta['sent_at'] = self.start_time
         status = int(self.status)
         headers = Headers(self.response_headers)
         respcls = responsetypes.from_args(headers=headers, url=self._url)
--- a/scrapy/extensions/throttle.py
+++ b/scrapy/extensions/throttle.py
@@ -139,7 +139,7 @@ class TimeIntervalManager:
     def get_time_part(self, time_stamp: datetime, degree: TimeInterval):
         return time_stamp.__getattribute__(self.time_interval_attr[degree])
 
-    def get_seconds_until_end_of_interval(self, time_stamp: datetime, interval: TimeInterval) -> int:
+    def get_seconds_until_end(self, time_stamp: datetime, interval: TimeInterval) -> int:
         # if interval is seconds, return 1.
         # if interval is minutes, return:  get_seconds(1 minute) - current second
         # if interval is hour, return:  get_seconds(1 hour) - (get_seconds(1 minute) * (60 minutes - current_minutes)
@@ -211,7 +211,6 @@ class SlidingWindow:
     def get_key(self, _datetime):
         return self._tim.get_timestamp_key(_datetime, self.interval)
 
-
     def __repr__(self):
         return "SlidingWindow: {}\n Current Window: {}\n Current Count{}\n Previous Count{}" \
             .format(self.interval, self._tim.get_timestamp_key(self.current, self.interval),
@@ -243,8 +242,8 @@ class AutoThrottle429Handler:
         self._set_special_delay(domain, sent_at, time_slot)
         # reduce failover lives
         self.failover_lives[domain] = self.failover_lives.get(domain, self.__initial_failover_lives__) - 1
-        logger.log(logging.INFO, "Decreasing request rate due to rate limit.\n" +
-                   "New rate: {}, {}".format(self.delay.get(domain), time_slot.name))
+        logger.log(logging.INFO, "Decreasing request rate due to rate limit.\n"
+                   + "New rate: {}, {}".format(self.delay.get(domain), time_slot.name))
 
     def get_min_delay(self, response):
         return self.delay.get(self._get_domain(response), 0)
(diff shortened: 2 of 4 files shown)


scrapy/extensions/throttle.py
```diff
@@ -2,6 +2,9 @@
 
 from scrapy.exceptions import NotConfigured
 from scrapy import signals
+from enum import Enum
+from datetime import datetime, timedelta
+from typing import List
```
Gallaecio, line 7: :lipstick: These should be under `import logging`, and sorted alphabetically:

```suggestion
from datetime import datetime, timedelta
from enum import Enum
from typing import List
```

scrapy/extensions/throttle.py
```diff
@@ -15,6 +18,12 @@ def __init__(self, crawler):
 
         self.debug = crawler.settings.getbool("AUTOTHROTTLE_DEBUG")
         self.target_concurrency = crawler.settings.getfloat("AUTOTHROTTLE_TARGET_CONCURRENCY")
+        self.enable_429 = crawler.settings.getbool("AUTOTHROTTLE_HANDLE_API_RATE_LIMIT")
+        self.rate_limit_handler = None
+        self.rate_limit_status = crawler.settings.getint("AUTOTHROTTLE_RATE_LIMIT_STATUS")
+        if self.enable_429:
+            self.rate_limit_handler = AutoThrottle429Handler(
+                self.target_concurrency)
```
Gallaecio, line 26: :lipstick: Given the length of lines above and below, wouldn’t this be better?
```suggestion
            self.rate_limit_handler = AutoThrottle429Handler(self.target_concurrency)
```

scrapy/extensions/throttle.py
```diff
@@ -15,6 +18,12 @@ def __init__(self, crawler):
 
         self.debug = crawler.settings.getbool("AUTOTHROTTLE_DEBUG")
         self.target_concurrency = crawler.settings.getfloat("AUTOTHROTTLE_TARGET_CONCURRENCY")
+        self.enable_429 = crawler.settings.getbool("AUTOTHROTTLE_HANDLE_API_RATE_LIMIT")
```
Gallaecio, line 21: Based on similar settings, something in the line of `AUTOTHROTTLE_HANDLE_RATE_LIMIT_ENABLED` may be better. In any case, the internal variable (`enable_429`) should definitely not have 429 in it if we are going to allow a different status code to be considered a too-many-requests response.

Gallaecio (commented):
I’ve left a few minor comments, but I see a few red flags:
- Making the assumption that slots are domain-based seems wrong, so if we need to rely on slots then I think we need to at least find a way to emit a warning for users when they use non-domain slots
- We are assuming that rate-limiting is domain-based. This seems like a big assumption to me, since it can be based on all sorts of other things. I think:
  1. assuming domain-based rate limiting is wrong, 
  2. determining the actual rate-limiting scope seems impossible to me, and 
  3. allowing users to configure the right rate-limiting scope, while the best approach I can think of, seems pointless; if users know the rate limiting rules  somehow would be ideal, but then there’s not much point in AutoThrottle handing this: if users know the rate limiting scope, they probably know the rate limits they need to apply, no need for AutoThrottle to figure them out.

scrapy/extensions/throttle.py
```diff
@@ -91,3 +112,193 @@ def _adjust_delay(self, slot, latency, response):
...
+        self.previous_count = new_previous_count
+
+    def is_next_window(self, label: str):
+        return self.get_key(self.current + self._tim.get_delta(self.interval)) == label
+
+    def get_key(self, _datetime):
+        return self._tim.get_timestamp_key(_datetime, self.interval)
+
+    def __repr__(self):
+        return "SlidingWindow: {}\n Current Window: {}\n Current Count{}\n Previous Count{}" \
```
elacuesta, line 215: A `__repr__` with newlines sounds strange to me, but that might be subjective.
However, we're trying to avoid backslashes, so please modify this to something like
```python
return (
    "stuff {} more stuff {} ..."
).format(
    var1,
    var2,
    ...
)
```

or maybe

```python
msg = "stuff {} more stuff {} ..."
return msg.format(
    var1,
    var2,
    ...
)
```

scrapy/core/downloader/webclient.py
```diff
@@ -141,6 +141,7 @@ def __init__(self, request, timeout=180):
 
     def _build_response(self, body, request):
         request.meta['download_latency'] = self.headers_time - self.start_time
+        request.meta['sent_at'] = self.start_time
```
elacuesta, line 144: Please note that this class is only used for HTTP/1.0 connections.

elacuesta (commented):
I left some comments, most of them aesthetic, to be honest I haven't looked at the implementation.
How does this relate to the per-request delay feature request we've had for some time? (#802)

PR #4464, opened 2020-03-30
Exposing find_method and get_method from scrapy.utils.reqser

### LINKED ISSUE #4125: Expose _find_method and _get_method from scrapy.utils.reqser
There are two functions in `scrapy.utils.reqser`, `_find_method` and `_get_method`, that are used to serialize and unserialize spider callbacks, useful e.g. when using `JOBDIR`.

It would be nice if these functions were given a non-private name, and were covered in the documentation about `JOBDIR` as functions that can be used to serialize and unserialize store spider callbacks in the request meta.

I found them handy on a spider that needs to temporarily change the callback of a request, and call the original callback from the new callback.

Related to #4124.

faizan2700:
fixes #4125

### DIFF
--- a/docs/topics/jobs.rst
+++ b/docs/topics/jobs.rst
@@ -82,3 +82,8 @@ running :class:`~scrapy.spiders.Spider` class.
 If you wish to log the requests that couldn't be serialized, you can set the
 :setting:`SCHEDULER_DEBUG` setting to ``True`` in the project's settings page.
 It is ``False`` by default.
+
+For serializing ``errback`` and ``callback`` functions, find_method is used as 
+helper function. find_method when given spider object and callable ``func`` 
+retuns function name. similarly get_method is used for unserializing.
+these helper functions are present in ``scrapy.utils.reqser.py``.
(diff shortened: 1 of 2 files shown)


docs/topics/jobs.rst
```diff
@@ -82,3 +82,8 @@ running :class:`~scrapy.spiders.Spider` class.
 If you wish to log the requests that couldn't be serialized, you can set the
 :setting:`SCHEDULER_DEBUG` setting to ``True`` in the project's settings page.
 It is ``False`` by default.
+
+For serializing ``errback`` and ``callback`` functions, find_method is used as 
+helper function. find_method when given spider object and callable ``func`` 
+retuns function name. similarly get_method is used for unserializing.
+these helper functions are present in ``scrapy.utils.reqser.py``.
```
Gallaecio: You should add docstrings to the functions, and use Sphinx’s autodoc to include that reference documentation here. You should also be able to refer to those functions with the `` :func:`function_name` `` Sphinx syntax.

wRAR:
The methods are now in `scrapy/http/request/__init__.py`.

PR #4467, opened 2020-03-31
Full async start_requests support

### LINKED ISSUE #456: Allow start_requests method running forever

### LINKED ISSUE #5627: Add async support for start_requests() method
Hi! It is often suitable to start initial requests fetching urls from some async backend, microservice etc, and not just using statically provided attributes/methods. We may use spider arguments for this, but IMO, if **start_requests** become async it should be very natural and organic in addition to async parse callbacks.

for example, if i have to request start urls from some external source (using async client) it would be nice just to do this:

```python
    async def start_requests(self):
        start_urls: Iterable[str] = await async_queue.get_urls(is_initial=True)
        for url in start_urls:
            yield Request(url=url, callback=self.on_initial_url_parser, ...)

    async def on_initial_url_parser(self, response: HtmlResponse, ...):

```

all network methods of the spider an asynchronous, match the one style, and are simple if we need to retrieve initial urls from some external async provider.

wRAR:
This should be complete but it fails one test, because I extracted processing start_requests into a separate inlineCallbacks method and the traceback now doesn't include the real exception line :-/

Closes #3237, resolves #456, resolves #5627


docs/topics/spider-middleware.rst
```diff
@@ -157,6 +160,15 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
...
         return another iterable of :class:`~scrapy.http.Request` objects.
 
+        When using Python 3.6+ `start_requests` may be an
+        :term:`python:asynchronous iterable`. To support such spiders, this
+        middleware method should be an :term:`python:asynchronous generator`,
+        which should support both synchronous and asynchronous
+        ``start_requests``. :func:`scrapy.utils.asyncgen.as_async_generator`
+        can be used to convert both kinds of ``start_requests`` to an
+        asynchronous iterable. See :ref:`more details here
+        <async-start_requests>`.
```
Gallaecio: If we use `autofunction` as I mention in an earlier comment, the “See …” sentence here could be removed, since `` :func:`scrapy.utils.asyncgen.as_async_generator` `` already links to the relevant part of the documentation.

docs/topics/coroutines.rst
```diff
@@ -46,10 +46,48 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+
+    class ProcessStartRequestsAsyncGenMiddleware:
+        async def process_start_requests(self, start_requests, spider):
+            async for r in as_async_generator(start_requests):
+                # ... do something with r
+                yield r
+
+If this method includes asynchronous code, that code will work even with
+synchronous :meth:`start_requests`.
+
```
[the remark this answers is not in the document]
wRAR: Where would it link?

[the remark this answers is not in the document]
Gallaecio: Wherever we put `.. autofunction:: scrapy.utils.asyncgen.as_async_generator`

scrapy/utils/asyncgen.py
```diff
@@ -0,0 +1,31 @@
...
+async def as_async_generator(it):
+    if isinstance(it, collections.abc.AsyncIterator):
+        async for r in it:
+            yield r
+    else:
+        for r in it:
+            yield r
+
+
+async def process_async_iterable_helper(it, in_predicate=None, out_predicate=None, processor=None):
```
wRAR, line 23: This is my idea of a generic helper for some middleware methods that process an iterable. I'm sure there can be cleaner things, so I'm open for suggestions. I've looked at our spider middlewares and several 3rd-party ones and it loks like most of them can be rewritten to use this, but the result is not always clean.

It would be much cleaner if we could just process every item of an iterable, but middlewares also often drop them (hence the predicate functions), which I think can only be done in a per-item processor function by raising a special exception which I don't like.

scrapy/utils/py36.py
```diff
@@ -1,10 +0,0 @@
-"""
-Helpers using Python 3.6+ syntax (ignore SyntaxError on import).
-"""
-
-
-async def collect_asyncgen(result):
```
wRAR, line 6: Maybe we need a deprecation for this.

I've renamed the module because it's useful to keep several things there even after we drop 3.5.

scrapy/utils/asyncgen.py
kmike: an idea: could we have a helper which converts a standart generator to asnyc generator, so that users can write `async for` always? It looks pretty awkward to limit middlewares to predicates and processors.

wRAR: Yes, `as_async_generator` is a such helper.

Also, it's not limited, you just need to write your code twice. I wrote this helper to be able to convert all our spider middlewares (their `process_spider_output` methods) to support normal and async callbacks without code duplication.

conftest.py
```diff
@@ -8,6 +8,8 @@ def _py_files(folder):
 
 
 collect_ignore = [
+    # no tests and not importable on 3.5
+    "scrapy/utils/asyncgen.py",
```
kmike, line 12: What do you think about adding it to collect_ignore only in Python 3.5?

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,71 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
 
+.. _async-start_requests:
+
+Asynchronous start_requests and spider middlewares
+==================================================
+
+.. versionadded:: 2.1
+
+The :meth:`~scrapy.spiders.Spider.start_requests` spider method can be an
+asynchronous generator. In this case all spider middlewares used with this
```
kmike, line 63: What do you think about adding an example of async generator here?

```py
    async def start_requests():
        # ...
        yield scrapy.Request(...)
        # ...
```

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,71 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+spider that have the
+:meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_start_requests`
+method must support this: if they receive an asynchronous iterable, they must
+return one as well. On the other hand, if they receive a normal iterable, they
+shouldn't break and ideally should return a normal iterable too. There can be
```
kmike, line 68: Why is it recommended to keep non-async iterables as they are? Is it because of a risk that some other middleware doesn't support async yet, or are there other reasons (e.g. performance)?

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,71 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+
+.. autofunction:: scrapy.utils.asyncgen.as_async_generator
+
+Here is an example of a universal middleware using this approach::
+
+    from scrapy.utils.asyncgen import as_async_generator
+
+    class ProcessStartRequestsAsyncGenMiddleware:
+        async def process_start_requests(self, start_requests, spider):
+            async for r in as_async_generator(start_requests):
```
kmike, line 86: Could you please change r to req or request?

docs/topics/coroutines.rst
```diff
@@ -85,7 +146,7 @@ This means you can use many useful Python libraries providing such code::
         async def parse_with_asyncio(self, response):
             async with aiohttp.ClientSession() as session:
                 async with session.get('https://additional.url') as additional_response:
-                    additional_data = await r.text()
+                    additional_data = await additional_response.text()
```
kmike, line 149: 👍

scrapy/core/engine.py
```diff
@@ -26,7 +28,10 @@ class Slot:
     def __init__(self, start_requests, close_if_idle, nextcall, scheduler):
         self.closing = False
         self.inprogress = set()  # requests in progress
-        self.start_requests = iter(start_requests)
+        if hasattr(inspect, 'isasyncgen') and inspect.isasyncgen(start_requests):
```
kmike, line 31: what do you think about adding a helper function with this code? It is used several times in the codebase, and used in docs as well.

docs/topics/coroutines.rst
kmike: Also, whatdo you think about making process_start_requests async in the gerenated templates?

scrapy/core/engine.py
wRAR: It's only needed while we support 3.5 but I agree it can be useful.

docs/topics/coroutines.rst
wRAR: @kmike do you mean https://github.com/scrapy/scrapy/blob/master/scrapy/templates/project/module/middlewares.py.tmpl ?

kmike: yes

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,71 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
     methods of
     :ref:`downloader middlewares <topics-downloader-middleware-custom>`.
 
+-   The
+    :meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_start_requests`
+    method of :ref:`spider middlewares <custom-spider-middleware>`.
+
```
Gallaecio, line 48:
```suggestion
-   The
    :meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_start_requests`
    method of :ref:`spider middlewares <custom-spider-middleware>`. See
    :ref:`async-start_requests` below for details.

```

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,71 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+        async def process_start_requests(self, start_requests, spider):
+            async for r in as_async_generator(start_requests):
+                # ... do something with r
+                yield r
+
+If this method includes asynchronous code, that code will work even with
+synchronous :meth:`~scrapy.spiders.Spider.start_requests`.
+
+Another option is to make separate methods for normal and asynchronous
+iterables and choose one at the run time::
```
Gallaecio, line 94:
```suggestion
iterables and choose one at run time::
```

docs/topics/spider-middleware.rst
```diff
@@ -157,6 +160,14 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
         It receives an iterable (in the ``start_requests`` parameter) and must
         return another iterable of :class:`~scrapy.http.Request` objects.
 
+        When using Python 3.6+ ``start_requests`` may be an
+        :term:`python:asynchronous iterable`. To support such spiders, this
+        middleware method should be an :term:`python:asynchronous generator`,
+        which should support both synchronous and asynchronous
+        ``start_requests``. :func:`scrapy.utils.asyncgen.as_async_generator`
+        can be used to convert both kinds of ``start_requests`` to an
```
Gallaecio, line 168:
```suggestion
        can be used to convert both kinds of ``start_requests`` into an
```

scrapy/utils/py36.py
Gallaecio: Yes, we should deprecate it, just in case.

Gallaecio (approved):

docs/topics/coroutines.rst
wRAR: I don't have an opinion on this

Gallaecio:
It would be great if we can include tests which cover the missing parts from https://codecov.io/gh/scrapy/scrapy/compare/febe82a9076d37ccc0b09a0ad97c8ae6ee1bd1b1...cab1207ce6add6e87a58a392b1c9411322a40dbf/diff

scrapy/utils/asyncgen.py
kmike: Hey! I'd prefer to remove process_async_iterable_helper, process_normal_iterable_helper and process_iterable_helper from the code here: they are unused in Scrapy itself, not really documented; it'd make the PR smaller.

docs/topics/coroutines.rst
kmike: What's the downside? I can see 3 downsides:

1. Generated code becomes more complicated
2. Generated code won't work with Python 3.5
3. We'll be changing the recommended way to write middleware in future, making this generated code deprecated.

I think (2) and (3) could be valid reasons not to have it as an example. Are they the case? 
As for (1), I think it worths it to have a proper implementation there, even if it means slightly more generated code.

wRAR: It's more of "do we need to change this if most people will still write only normal start_requests and normal middlewares for them". It also can be more complicated for people not familiar with async def functions, but that's probably not a big problem. 

About (2): assuming most of the use cases need an iteration, straightforward `for` loop with `yield` will indeed not work in 3.5. About (3): I'm still not sure what should we recommend for universal middlewares, but we can always update the template again.

scrapy/utils/defer.py
```diff
@@ -173,3 +173,10 @@ def maybeDeferred_coro(f, *args, **kw):
         return defer.fail(result)
     else:
         return defer.succeed(result)
+
+
+def isasyncgen(o):
```
elacuesta, line 178: What do you think about making this underscored? I understand that we need to check the existence of `inspect.isasyncgen` before using it to keep the Scrapy codebase compatible with py35, and that the same check is done in several places so it's good to make it reusable. But using this function in user code only makes sense in py36 because there are no async generators before, so the example in the docs could use `inspect.isasyncgen` directly.
There's a change users might want to keep their projects compatible with py35 too, in that case they could perform the `hasattr` check themselves. Py35 is reaching [end-of-life](https://devguide.python.org/#status-of-python-branches) relatively soon, and I think it'd be good to avoid having to maintain a function that will not be necessary after we drop py35 support.

wRAR: I agree, @kmike what do you think?

kmike: Sounds good to me. I recall this function is used in example; we can just assume Python 3.6 in the example.

wRAR:
This PR now includes changes from #3237 by @whalebot-helmsman with the following additional changes (commit 94ce9307433e3cccdb567d57b70569f704fd9faf):

- there is no `start_requests_with_control`
- the old behavior is retained for `def start_requests`  but `async def start_requests` uses the new behavior
- as the ability to emit WaitUntilQueueEmpty from the user code doesn't work with custom spider middlewares, it was decided to not add it as it is an additional feature

wRAR:
So now the PR consists of three parts:

- the `async def start_requests` implementation (the commits up to Jun 3)
- the changes from #3237 merged at b56eba7952dae323e4cd15353b6a3d73f5dddc45
- the integration between them (94ce930)

We also want the documentation for the new queue behavior, though it can be added separately before the next release.

docs/topics/coroutines.rst
```diff
@@ -42,10 +42,80 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+-   The :meth:`~scrapy.spiders.Spider.start_requests` spider method.
+
 .. _asynchronous generators were introduced in Python 3.6: https://www.python.org/dev/peps/pep-0525/
 
+.. _async-start_requests:
+
+Asynchronous start_requests and spider middlewares
+==================================================
+
+.. versionadded:: 2.2
```
kmike, line 61: It should be either 2.3 or 2.4, depending on how things go

wRAR: Right, I've updated another one and missed this one.

scrapy/crawler.py
```diff
@@ -105,6 +106,14 @@ def call_start_requests(self):
         else:
             return iter(self.spider.start_requests())
 
+    @staticmethod
+    def is_start_requests_async(start_requests_function):
+        if hasattr(inspect, 'isasyncgenfunction') and inspect.isasyncgenfunction(start_requests_function):
+            return True
+        elif inspect.iscoroutinefunction(start_requests_function):
```
Gallaecio, line 113: :lipstick: 
```suggestion
        if inspect.iscoroutinefunction(start_requests_function):
```

scrapy/spiders/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 
 class Spider(object_ref):
+
     """Base class for scrapy spiders. All spiders must inherit from this
```
Gallaecio, line 18: :lipstick: 
```suggestion
class Spider(object_ref):
    """Base class for scrapy spiders. All spiders must inherit from this
```

Gallaecio (commented):
Could you have a look at the [reported coverage changes](https://codecov.io/gh/scrapy/scrapy/pull/4467/changes) affecting `robotstxt` and `core.downloader`? Just to make sure we are not introducing some unexpected side effect.

wRAR:
@Gallaecio the robotstxt code is only used when reppy is installed (so only in the `extra-deps` env), and it's used by tests both on master and this PR, can it be because coverage runs in different envs?

Not sure about the `Slot.close()` code, I put `1/0` there and ran tests on master and it didn't hit, I suspect it depends on timing.

Gallaecio (approved):
The documentation looks good to me.

We just need to remember to remove that last paragraph from the first release dropping Python 3.5 support, which made me realize that in general we will need to look for both 3.5 and 3.6 references in documentation and code, not just 3.5 as I would have though.

PS: The coverage is 100% now, it looks like it was a temporary issue from either Travis or codecov.

Gallaecio:
Note to self: update those Sphinx version directives when preparing the release notes.

docs/topics/spiders.rst
```diff
@@ -169,6 +169,14 @@ scrapy.Spider
                    # each of them, with another callback
                    pass
 
+       .. note::
+           .. versionchanged:: 2.x
+
+           By default the requests produced by :meth:`start_requests` can be
+           processed much later than expected. For :meth:`start_requests`
+           declared as an ``async def`` function a better behavior is used. See
+           :ref:`async-start_requests-queue` for details.
```
kmike, line 178: I'm not sure "default" is a right word here. For `def start_requests` behavior is this, for `async def start_requests` behavior is that.

I guess here we're suggesting to use `def start_requests` becuase `async def start_requests` could break some third-party middlewares, is this true? Should we explain it here? Otherwise it looks a bit strange - we tell that the example is not written in the best way, but we're not explaining why.

kmike: I also wonder how to word it to make sure `versionchanged` above makes sense. start_requests behavior is not getting worse in 2.4, it is the opposite - there is now a way around it :)

docs/topics/coroutines.rst
```diff
@@ -37,8 +37,101 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+.. _async-start_requests-queue:
+
+Queue behavior with asynchronous start_requests
+===============================================
+
+.. versionadded:: 2.x
+
+By default the requests produced by
+:meth:`~scrapy.spiders.Spider.start_requests` don't follow the usual request
+priorities: they are only scheduled while the downloader queue is empty. In
```
kmike, line 121: I'm not sure this is correct - request.priority is not relevant here, and the logic depends on scheduler queue being empty, not downloader queue. I think requests are scheduled while the downloader queue is not full (because at this point scheduler queue starts to grow), not while it is empty.

docs/topics/coroutines.rst
```diff
@@ -37,8 +37,101 @@ hence use coroutine syntax (e.g. ``await``, ``async for``, ``async with``):
...
+by callbacks are scheduled. This behavior is inconvenient, confusing and
+requires workarounds (such as calling ``self.crawler.engine.schedule(request)``
+directly), but changing it would be backwards-incompatible.
+
+A better behavior, which should treat requests from
+:meth:`~scrapy.spiders.Spider.start_requests` in the same way as other
+requests, is enabled when :meth:`~scrapy.spiders.Spider.start_requests` is an
+async function (declared using ``async def``). It doesn't need to contain
+``await`` for this to work, so if you want the new queue behavior, you can just
+change ``def`` to ``async def``.
```
kmike, line 133: I know that's quite hard to act on feedback about docs :) It is also highly subjective. But I think the explanation is not 100% clear here.

This all has nothing to do with request priorities (as request.priority). It is more about if all requests from start_requests are consumed or not, or to which point they are consumed. What do you think about explaining it from this perspective? I would also start with describing the "good" behavior, and then explani the old, historical one. 

Something like this: 

Requests produced by `async def start_requests` are scheduled first. If there is any `await` inside, Scrapy may switch to processing callbacks, but eventually `async def start_requests` resumes its execution as usual (of course, depending on what you're awaiting). 

Note that for historical reasons behavior of non-async, regular `def start_requests` is different: requests are only scheduled while the scheduler queue is empty. In practive this means that only first few requests from start_requests are scheduled (up to MAX_CONCURRENT_REQUESTS), and then processing of start_request stops until all requests produced by callbacks are processed. This behavior is confusing, and may require workarounds; we might change it in future for regular start_requests as well. 

If you want the new queue behavior, you can just change ``def start_requests`` to ``async def start_requests``. Note that in this case all spider(?) middlewares with process_start_requests methods should support async generators. Built-in Scrapy middlewares should work, some third-party or internal middlewares might need to be updated, as described here (todo: add a link).

scrapy/core/engine.py
```diff
@@ -23,11 +25,15 @@
 
 class Slot:
 
-    def __init__(self, start_requests, close_if_idle, nextcall, scheduler):
+    def __init__(self, start_requests, close_if_idle, nextcall, scheduler, new_queue_behavior=False):
```
kmike, line 28: The new_queue_behavior may become not so new after some time :) What do you think about calling the new argument old_queue_behavior, or, even better, with some name which describes the behavior?

scrapy/core/engine.py
```diff
@@ -120,24 +149,23 @@ def _next_request(self, spider):
...
-                logger.error('Error while obtaining start requests',
-                             exc_info=True, extra={'spider': spider})
+        if self._waiting_for_request:
+            return
+        self._waiting_for_request = True
+        try:
+            if slot.new_queue_behavior:
+                while not self._needs_backout(spider) and slot.start_requests:
+                    yield self._schedule_next_req(spider, slot)
+                self._send_reqs_to_downloader(spider)
```
kmike, line 159: What happens in examples like that?

```py
async def start_requests(self):
    # yield many requests
    await asyncio.sleep(10)
    # yield many requests
```

I'd expect that

1) while start_request is sleeping, requests scheduled before that are being downloaded;
2) if responses are received during this 10s sleep, callbacks are executed as usual, which may result in items extracted, new requests scheduled, etc.;
3) when start_request resumes after `await`, remaining requests are put to the scheduler as usual, without waiting  for any queues to become empty

Is all of this working? Sorry if that's a stupid question, I haven't really understood the implementation yet.

wRAR: I've just tested this, and it works in a different way: the requests are scheduled, but `_send_reqs_to_downloader` is only called after `sleep` returns, so nothing is downloaded before that and these use cases cannot be tested.

scrapy/core/engine.py
wRAR: I'm fine with renaming this flag to whatever someone proposes.

docs/topics/spiders.rst
wRAR: This note is not for the example but for the `.. method:: start_requests()` block as this is the only place where it can be written except the coroutines package that it links to. I didn't intend to suggest anything, just to write here that there are two possible behaviors.

docs/topics/coroutines.rst
wRAR: Agreed.

scrapy/core/engine.py
kmike: Hm, is it possibile to fix that? I think it breaks some of the most important use cases, like this one:

```py
async def start_requests(self):
    while True:
        batch = await get_next_batch_from_redis()
        for req in batch:
            yield scrapy.Request(req['url'], ...) 
```

> and these use cases cannot be tested.

what do you mean by that?

docs/topics/coroutines.rst
wRAR: I know my explanations are not 100% clear, as I tried to document things I don't understand myself.

I'll implement your suggestions.

scrapy/core/engine.py
wRAR: > Hm, is it possibile to fix that?

Without limiting the number of requests scheduled in the while loop? No idea. I can only guess whether the initial idea was to load all available requests into the scheduler before starting to move them to the downloader.  

The only code I added here, `yield deferred_from_coro(slot.start_requests.__anext__())`, doesn't distinguish between a long wait and just getting the next value from the generator.

> what do you mean by that?

I mean this test case cannot give us answers whether your points 2 and 3 are correct (if there are use cases where they make sense).

scrapy/core/engine.py
wRAR: Maybe something about "eager" or "fair"?

scrapy/core/engine.py
kmike: I'm actually unsure what's the use case for `async def start_requests` if the requests don't start to download while something is being awaited in `async def start_requests`. Why not make e.g. a sync time.sleep, or a sync call to redis instead, won't the behavior be the same in the end? 

Do you see a use case for that? If not, I think it would be better to remove 2.4 milestone from this PR, and wait until that is fixed.

There is a use case though where we might need to ensure that all start_requests made it into the scheduler, before starting to download them. I wonder if we can have an API for that as well - maybe even a different method. It may be handled in the current implementation (aiter may be consuming requests and scheduling them without yielding to the event loop, which makes behavior the same), but I'm not sure about that.

A good point about `yield deferred_from_coro(slot.start_requests.__anext__())`. But maybe that's fine to not distinguish between those, not guarantee that downloads don't start until await.

How hard would it be to modify start_requests to ensure downloading happens in parallel?

wRAR: I agree that the current behavior is unacceptable but I don't know what should be the correct behavior. The current code works as it was written in #3237, loading all requests from `start_requests` into the scheduler. 

I'm fine with removing this PR from 2.4.

wRAR:
Not sure how to unblock this but we decided to list the use cases we want to support in this PR.

1. A very long list of start requests (doesn't require the async support, only the new queue behavior).
2. Some requests taken from some external place accessed with an async API (requires the async support but can probably also be done synchronously; queue behavior doesn't matter).
3. A proper external queue with an async API, with no guarantee on the number of requests and on the delays before getting the next one (requires the async support; requires that the spider runs fine while it waits for the next request; doesn't work if all start requests need to be scheduled before the spider starts working).

Anything else?

Gallaecio:
We should also make sure any new implementation still supports both [crawl orders](https://docs.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order).

scrapy/core/engine.py
Gallaecio: About the `sleep` issue, I wonder it we could iterate over requests from an asynchronous `start_requests` [using a timeout](https://stackoverflow.com/q/50241696/939364), so that if no requests are yielded for a certain time, other stuff is handled for some time, and iteration of `start_requests` is resumed at some later point.

I’m not sure how viable this would be, though, as I’m talking from a complete lack of familiarity with the actual code.

Gallaecio:
Today, working on speeding up a broad crawl project, I found that applying @whalebot-helmsman’s downloader-aware priority queue to `start_requests` can make a difference, because unlike requests found while crawling, start requests only reach the scheduler when it needs more requests (or something like that, my understanding of that part of the code is limited).

So, I think one use case for start_requests would be for them to be fed into the scheduler all at once, to let the scheduler sort them all, instead of just sorting a subset of them plus crawler requests.

diegocostares:
And can you think of any temporary solution to work with multiple urls without resorting to another library? 😭 😭 😭

wRAR:
Scrapy already supports "multiple URLs". If you mean e.g. #456 it alread has some workarounds suggested.

diegocostares:
I understand Scrapy supports multiple URLs, but my concern is about efficiently handling a very long list of URLs in `start_requests`. For example, if I have:

```python
    name = 'myspider'

    def start_requests(self):
        long_list_of_urls = ['http://example.com/page1', 'http://example.com/page2', ..., 'http://example.com/page10000']
        for url in long_list_of_urls:
```
With thousands of URLs in the list, it is not clear from the documentation how to optimize them so that they are processed in parallel. Could you show me an example of how it should be done?

_PS: thank you very much for your contributions and patience_

wRAR:
> Could you show me an example of how it should be done?

I believe I already pointed you to them.

Gallaecio:
@diegocostares To reiterate, Scrapy does handle long lists of start URLs efficiently and in parallel. If you feel it goes too slow, check your [concurrency settings](https://docs.scrapy.org/en/latest/topics/settings.html#concurrent-requests). This pull request is about something else.

diegocostares:
@Gallaecio I already left that setting at 16. However, after some testing, my start_requests function is the slowest part.😓Anyway thank you very much

PR #4468, opened 2020-04-01
Fix pycodestyle E2XX (whitespace)

elacuesta:
Most of the changes are to test files.

This PR deals with E2XX errors exclusively: for instance, the changed lines in `tests/test_webclient.py` are still affected by E122 ("continuation line missing indentation or outdented")

Some of the above changes conflict with #4392 (please take alook @noviluni)

### DIFF
--- a/tests/test_crawl.py
+++ b/tests/test_crawl.py
@@ -147,9 +147,10 @@ class CrawlTestCase(TestCase):
         settings = {"CONCURRENT_REQUESTS": 1}
         crawler = CrawlerRunner(settings).create_crawler(BrokenStartRequestsSpider)
         yield crawler.crawl(mockserver=self.mockserver)
-        #self.assertTrue(False, crawler.spider.seedsseen)
-        #self.assertTrue(crawler.spider.seedsseen.index(None) < crawler.spider.seedsseen.index(99),
-        #                crawler.spider.seedsseen)
+        # self.assertTrue(False, crawler.spider.seedsseen)
+        # self.assertTrue(
+        #     crawler.spider.seedsseen.index(None) < crawler.spider.seedsseen.index(99),
+        #     crawler.spider.seedsseen)
 
     @defer.inlineCallbacks
     def test_start_requests_dupes(self):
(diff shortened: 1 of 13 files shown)


tests/test_crawl.py
```diff
@@ -147,9 +147,10 @@ def test_start_requests_lazyness(self):
         settings = {"CONCURRENT_REQUESTS": 1}
         crawler = CrawlerRunner(settings).create_crawler(BrokenStartRequestsSpider)
         yield crawler.crawl(mockserver=self.mockserver)
-        #self.assertTrue(False, crawler.spider.seedsseen)
-        #self.assertTrue(crawler.spider.seedsseen.index(None) < crawler.spider.seedsseen.index(99),
-        #                crawler.spider.seedsseen)
+        # self.assertTrue(False, crawler.spider.seedsseen)
+        # self.assertTrue(
+        #     crawler.spider.seedsseen.index(None) < crawler.spider.seedsseen.index(99),
+        #     crawler.spider.seedsseen)
```
noviluni, line 153: I think it is a good moment to delete this commented code. As explained by the [eradicate](https://github.com/myint/eradicate) library...:

> With modern revision control available, there is no reason to save commented-out code to your repository.

In fact, when fixing some things (variable names, pep8 rules, or other things) we usually forget to update the comments, so I think is not a good idea to keep it. 

Of course, you can also try to uncomment it and see if it works or try to guess with the git history why it was commented.

elacuesta: You're absolutely right. Seems like these assertions never worked, they were introduced on [Oct 3 2013](https://github.com/scrapy/scrapy/commit/5eb429999e9e6a49eb9721d2dc0a47e0100345c3#diff-4c001feadd01dd8b6534e86c1aa5aecaR112-R114) and modified on [July 31 2014](https://github.com/scrapy/scrapy/commit/870438e5f4665de836f7ff423055895b305a4e7f#diff-ad82c3d0394ef19ca24ea9d022470bf7R111-R113).
I'm removing them, thanks for pointing this out :+1:

elacuesta: Well, actually `self.assertTrue(False)` is never going to work :sweat_smile:

PR #1466, opened 2015-08-29
Fix for #669

umrashrf:
Solution provided by @dangra https://github.com/scrapy/scrapy/pull/670#issuecomment-38921769

I can add tests but should I add this dependency for FTP tests? https://github.com/giampaolo/pyftpdlib 

Current FTP tests uses Twisted so I'm afraid they won't work for PY3 https://github.com/scrapy/scrapy/blob/master/tests/test_downloader_handlers.py#L523-L543

PR #4474, opened 2020-04-06
Add a downloader middleware to translate URI userinfo data into Request.meta keys

### LINKED ISSUE #669: FTP Handler should read credentials from URL
Based on recent chat in support room, it seems like currently FTP handler can't read authentication credentials from URL. If Chrome and Firefox can do this, I think Scrapy should too. Look forward to your objections and/or comments.

### LINKED ISSUE #851: HTTP Basic Auth in URL as parameter to Scrapy shell apparently triggers a "DNS lookup error"
`$ scrapy shell "http://user@pass:test.com/path/"`

Maybe this needs to be performed using the special Http Auth middleware, however it looks like the scrapy shell command doesn't enable this middleware by default. 

version: 0.24.4 

Let me know if you need anymore doc to reproduce.

Gallaecio:
This is a continuation of #1466 by @umrashrf, implementing the second option suggested by @redapple at https://github.com/scrapy/scrapy/pull/1466#issuecomment-287089498.

Fixes #669, fixes #851, closes #1466


scrapy/downloadermiddlewares/uriuserinfo.py
```diff
@@ -0,0 +1,49 @@
+from urllib.parse import unquote, urlunparse
+
+from scrapy.utils.httpobj import urlparse_cached
+
+
+class UriUserinfoMiddleware(object):
+    """Downloader middleware that replaces `URI userinfo`_ data (user credentials
+    for HTTP or FTP specified in the request URL) with the corresponding meta
+    keys for later middlewares or download handlers to use them for
```
elacuesta, line 9: I might be missing something, but AFAICT this won't be enough to authenticate HTTP requests since the current [HTTP/1.1 download handler](https://github.com/scrapy/scrapy/blob/2.0.1/scrapy/core/downloader/handlers/http11.py) does not rely on `meta` keys and the `HttpAuthMiddleware` takes the [credentials from the spider](https://github.com/scrapy/scrapy/blob/2.0.1/scrapy/downloadermiddlewares/httpauth.py#L22-L26), not the request.

Gallaecio: I updated the documentation but forgot to implement those changes :facepalm:

scrapy/downloadermiddlewares/uriuserinfo.py
```diff
@@ -0,0 +1,49 @@
+from urllib.parse import unquote, urlunparse
+
+from scrapy.utils.httpobj import urlparse_cached
+
+
+class UriUserinfoMiddleware(object):
```
elacuesta, line 6: :warning: _Nitpick alert_ :warning:
Should "info" be capitalized to "Info"? The [corresponding RFC](https://tools.ietf.org/html/rfc3986#section-3.2.1) says "userinfo", all lowercase, but if we're using camel case I think it might make sense to make it "UserInfo" like in [this example](https://docs.microsoft.com/en-us/dotnet/api/system.uri.userinfo?view=netframework-4.8)

scrapy/downloadermiddlewares/httpauth.py
```diff
@@ -26,6 +26,14 @@ def spider_opened(self, spider):
...
+        if b'Authorization' in request.headers:
+            return
+
+        usr = request.meta.get('http_user', '')
+        pwd = request.meta.get('http_pass', '')
+        if usr or pwd:
+            auth = basic_auth_header(usr, pwd)
+        else:
+            auth = getattr(self, 'auth', None)
+        if auth:
```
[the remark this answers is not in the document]
Gallaecio: I moved that part to the beginning of the method, so that if there is already such a header we save the CPU needed for the rest of the logic (e.g. `basic_auth_header`).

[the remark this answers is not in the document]
elacuesta: Oh nevermind then, didn't see that :facepalm:

PR #4480, opened 2020-04-10
Pin version for Sphinx and extensions, configure hoverxref

### LINKED ISSUE #4475: The documentation cannot be built with Sphinx 3

elacuesta:
It's green! :heavy_check_mark: 

Closes #4475

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,14 +13,13 @@
 
 import sys
 from datetime import datetime
-from os import path
 from pathlib import Path
 
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
 # absolute, like shown here.
-sys.path.append(path.join(path.dirname(__file__), "_ext"))
-sys.path.insert(0, path.dirname(path.dirname(__file__)))
+sys.path.append(str(Path(__file__).absolute().parent / "_ext"))
+sys.path.insert(0, str(Path(__file__).absolute().parent.parent))
 
 
 # General configuration
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -297,3 +297,10 @@ intersphinx_mapping = {
 hoverxref_auto_ref = True
 hoverxref_project = "scrapy"
 hoverxref_version = release
+hoverxref_role_types = {
+    "class": "tooltip",
+    "confval": "tooltip",
+    "hoverxref": "tooltip",
+    "mod": "tooltip",
+    "ref": "tooltip",
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,13 +13,13 @@
 
 import sys
 from datetime import datetime
-from pathlib import Path
+from os import path
 
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
 # absolute, like shown here.
-sys.path.append(str(Path(__file__).absolute().parent / "_ext"))
-sys.path.insert(0, str(Path(__file__).absolute().parent.parent))
+sys.path.append(path.join(path.dirname(__file__), "_ext"))
+sys.path.insert(0, path.dirname(path.dirname(__file__)))
 
 
 # General configuration
@@ -59,10 +59,10 @@ copyright = '2008–{}, Scrapy developers'.format(datetime.now().year)
 #
 # The short X.Y version.
 try:
-    version_path = Path(__file__).parent.absolute().parent.joinpath("scrapy/VERSION")
-    version = version_path.read_text().strip()
-    release = version.rsplit(".", 1)[0]
-except Exception:
+    import scrapy
+    version = '.'.join(map(str, scrapy.version_info[:2]))
+    release = scrapy.__version__
+except ImportError:
     version = ''
     release = ''
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,6 +14,7 @@
 import sys
 from datetime import datetime
 from os import path
+from pathlib import Path
 
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -59,10 +59,10 @@
 #
 # The short X.Y version.
 try:
-    import scrapy
-    version = '.'.join(map(str, scrapy.version_info[:2]))
-    release = scrapy.__version__
-except ImportError:
+    version_path = Path(__file__).parent.absolute().parent.joinpath("scrapy/VERSION")
```
elacuesta, line 62: This should be safe I think (https://github.com/scrapy/scrapy/blob/2.0.1/setup.py#L6-L7)

docs/conf.py
```diff
@@ -295,3 +295,5 @@
 # ------------------------------------
 
 hoverxref_auto_ref = True
+hoverxref_project = "scrapy"
```
elacuesta, line 298: Apparently this is necessary now, otherwise we get
```
Warning, treated as error:
hoverxref extension is not fully configured. Set hoverxref_project and hoverxref_version in your conf.py file.
```
(https://travis-ci.org/github/scrapy/scrapy/jobs/673517936#L285-L286)

docs/conf.py
```diff
@@ -295,3 +295,12 @@
...
+hoverxref_version = release
+hoverxref_role_types = {
```
elacuesta, line 300: This seems to be necessary too, otherwise we get

```
Using default style for unknown typ. Define it in hoverxref_role_types. typ=ref style=tooltip
```
Setting: 
https://sphinx-hoverxref.readthedocs.io/en/latest/configuration.html#confval-hoverxref_role_types
Failed build: https://travis-ci.org/github/scrapy/scrapy/jobs/673579524#L1209

docs/topics/settings.rst
```diff
@@ -1116,17 +1116,6 @@ multi-purpose thread pool used by various Scrapy components. Threaded
 DNS Resolver, BlockingFeedStorage, S3FilesStore just to name a few. Increase
 this value if you're experiencing problems with insufficient blocking IO.
 
-.. setting:: REDIRECT_MAX_TIMES
```
elacuesta, line 1119: Removed to avoid errors about duplicate definitions, such as
```
/home/travis/build/scrapy/scrapy/docs/topics/settings.rst:1119:duplicate setting description of REDIRECT_MAX_TIMES, other instance in topics/downloader-middleware
```
https://travis-ci.org/github/scrapy/scrapy/jobs/673570891#L348

docs/conf.py
Gallaecio: Was there something wrong with the previous implementation? `scrapy.__version__` seems cleaner to me.

docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-Sphinx>=2.1
-sphinx-hoverxref
-sphinx-notfound-page
-sphinx_rtd_theme
+Sphinx==3.0.1
+sphinx-hoverxref==0.2b1
+sphinx-notfound-page==0.4
+sphinx_rtd_theme==0.4.3
```
Gallaecio, line 4: I would not freeze versions, and instead set `>=` restrictions as before where needed. That’s how things break for every pull request when a new version breaks things, but it is also how we find out early. Since building the documentation is not something people are expected to do themselves with released versions, I think breaking early is a better tradeoff. I don’t have a strong opinion about it, though.

elacuesta: I thought pinning exact versions might be controversial :smile:
I also don't have a strong preferencem but breaking early makes sense, so let's change it.

docs/conf.py
elacuesta: Indeed, I change it because it wasn't working for me, but I just found it was because I didn't have Scrapy installed in the virtualenv for the docs. I'll revert it, thanks for noticing.

docs/conf.py
```diff
@@ -295,3 +295,12 @@
...
+    "class": "tooltip",
+    "confval": "tooltip",
+    "hoverxref": "tooltip",
+    "mod": "tooltip",
+    "ref": "tooltip",
+}
```
Gallaecio, line 306: We might want to add our custom roles as well, based on https://github.com/scrapy/scrapy/issues/4475#issuecomment-613350667

docs/topics/downloader-middleware.rst
```diff
@@ -829,6 +829,7 @@ REDIRECT_MAX_TIMES
 Default: ``20``
 
 The maximum number of redirections that will be followed for a single request.
+After this maximum the request's response is returned as is.
```
Gallaecio, line 832:
```suggestion
After this maximum, the request's response is returned as is.
```

kmike:

docs/conf.py
```diff
@@ -295,3 +295,12 @@
 # ------------------------------------
 
 hoverxref_auto_ref = True
+hoverxref_project = "scrapy"
+hoverxref_version = release
```
humitos, line 299: Hi! This breaks in production when rendered by Read the Docs.

You shouldn't override `hoverxref_version` and `hoverxref_project` since they are taken automatically from Read the Docs.

If you want to avoid your CI failing because of this, you can define the environment variables as Read the Docs does:

- READTHEDOCS_PROJECT=scrapy
- READTHEDOCS_VERSION=''

With the current configuration, _all the versions built on Read the Docs_ will point to a different _version_ on Read the Docs and this will conflict. For example, current `master` version in Read the Docs defines `hoverxref_version='2.0.0'` but that version does not exist on Read the Docs and the tooltip does not known where to get the content from.

Cheers,

docs/conf.py
Gallaecio: Created https://github.com/scrapy/scrapy/issues/4495 for this.

docs/conf.py
kmike: Thanks for the heads up @humitos!

elacuesta: Many thanks for looking into this @humitos! :rocket:

PR #4486, opened 2020-04-12
Add status (downloaded, uptodate) to files information

### LINKED ISSUE #2893: Feature suggest: add downloaded/uptodate status to information about downloaded media
Currently File/Image pipelines populate `files`/`images` fields with dicts containing information about the downloaded files (the downloaded path, the original scraped url, and the file checksum). It would be useful to have downloaded/uptodate status in this dict ([motivation](https://stackoverflow.com/questions/45823196/scrapy-how-to-get-status-of-file-download)).

It goes along with other features requests such as having width/height of images also in the dict output.

ilias-ant:
- adds file status to the downloaded files information, as emitted by media pipelines. 
- updates the documentation accordingly.

File status can be a very useful metric to many Scrapy-based applications, especially the ones that heavily rely on `FilesPipeline` or `ImagesPipeline`. This extra information available can be utilized with various (and creative) ways, enabling the end-user to have more control to his file processing algorithms, creating more synergy with other features such as file expiration policy or `file_status_count` statistics.

fixes #2893

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -50,7 +50,7 @@ this:
 4. When the files are downloaded, another field (``files``) will be populated
    with the results. This field will contain a list of dicts with information
    about the downloaded files, such as the downloaded path, the original
-   scraped url (taken from the ``file_urls`` field) , and the file checksum.
+   scraped url (taken from the ``file_urls`` field), the file checksum and the file status.
    The files in the list of the ``files`` field will retain the same order of
    the original ``file_urls`` field. If some file failed downloading, an
    error will be logged and the file won't be present in the ``files`` field.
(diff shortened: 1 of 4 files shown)


Gallaecio:
Do you think you could add tests that cover the usage of the other two statuses (or modify existing tests to include checks for them, if there are tests where they may be already covered but not checked through an assert)?

ilias-ant:
Gladly! I'll need a bit of time to familiarize myself with the respective test suite and find a proper way to set up the necessary logic to "create" `uptodate`, `cached` images - any quick general suggestion on that? :)

Gallaecio:
If it can be done with a bit of `mock` usage, I would go that way.

Worst case scenario, for complex tests that require running spiders, we have test spiders that are run with `subprocess` (see, for example, `test_crawler.py`). I guess that is also a possibility. But I would try to avoid having to setup some mock server if possible.

ilias-ant:
I have covered all 3 available status(es):

- by enriching existing tests for the cases of `downloaded`, `uptodate`.
- by creating a new test for the case of `cached`, providing a suitable mock response. I've covered this case in the most "unittest" case possible, as i couldn't find a way to produce a more "functional" test.

I'd be glad to hear any additional feedback on this, in case i missed or misinterpreted something from the earlier comments :)

kmike:
Thanks @ilias-ant!

PR #424, opened 2013-10-09
[MRG] added required option to settings module

nramirezuy:

PR #4493, opened 2020-04-15
[ON HOLD] Add a required parameter to setting getters

Gallaecio:
Fork of #424 that solves merge conflicts.

Closes #424


scrapy/settings/__init__.py
```diff
@@ -93,7 +97,12 @@ def __getitem__(self, opt_name):
     def __contains__(self, name):
         return name in self.attributes
 
-    def get(self, name, default=None):
+    def _get_default(self, value, default, required):
+        if not required and value is _NOT_SET:
+            return default
+        return value
+
+    def get(self, name, default=_NOT_SET, required=False):
```
kmike, line 105: What do you think about getting `required` a kwarg-only argument, to discourage this?
`settings.getbool("foo", True, False)`

kmike:
Thanks for picking it up @Gallaecio! I wonder if these changes should be dogfooded - do we have places in Scrapy where this feature would simplify code? Probably a search for "raise NotConfigured" could give good candidates (see e.g. logstats extension).

Gallaecio:
I’m not sure we can use it in Scrapy. I’ve had a look at all instances of `raise NotConfigured`, and all those that come after reading a setting seem to not care whether the setting is defined or not, but whether or not it evaluates to `True` as a boolean.

For example, from the logstats extension:

```py
interval = crawler.settings.getfloat('LOGSTATS_INTERVAL')
if not interval:
    raise NotConfigured
```

kmike:
@Gallaecio do you know if there are code examples in a real world which will be simplified by this feature?

Gallaecio:
I saw some interest in the [original pull request](https://github.com/scrapy/scrapy/pull/424), so I assumed there may be (or have been?), but I know of no actual example myself, and I have no personal interest in this feature — I would not oppose just closing this and https://github.com/scrapy/scrapy/pull/424 without merging.

kmike:
Let's keep it open - maybe someone will jump in and say "I want it, this is my use case!" :)

Lukas0907:
> Let's keep it open - maybe someone will jump in and say "I want it, this is my use case!" :)

@Gallaecio pointed me to this PR. I would actually need this feature in my PR (Lukas0907/scrapy#1).

The relevant snippet is as follows:

```python
        host = self.settings.get('SCHEDULER_EXTERNAL_QUEUE_REDIS_HOST')
        port = self.settings.get('SCHEDULER_EXTERNAL_QUEUE_REDIS_PORT')
        db = self.settings.get('SCHEDULER_EXTERNAL_QUEUE_REDIS_DB')
        prefix = self.settings.get('SCHEDULER_EXTERNAL_QUEUE_REDIS_PREFIX')
        if host is None or port is None or db is None or prefix is None:
            raise NotConfigured(
                "Please configure "
                "SCHEDULER_EXTERNAL_QUEUE_REDIS_HOST, "
                "SCHEDULER_EXTERNAL_QUEUE_REDIS_PORT, "
                "SCHEDULER_EXTERNAL_QUEUE_REDIS_DB, "
                "SCHEDULER_EXTERNAL_QUEUE_PREFIX "
                "in the project settings so that Scrapy can connect to Redis."
            )
```

Therefore having a `required=True` flag would save me from writing some boilerplate code. :)

scrapy/settings/__init__.py
```diff
@@ -102,10 +111,18 @@ def get(self, name, default=None):
...
+
+        :param required: if ``True`` and `default` is not specified, it will
+            raise an :class:`~scrapy.exceptions.NotConfigured` exception if the
+            specified setting has not been defined
+        :type required: boolean
         """
-        return self[name] if self[name] is not None else default
+        value = default if name not in self.attributes else self[name]
+        if required and value is _NOT_SET:
+            raise NotConfigured
```
Gallaecio, line 122: Should we indicate here the name of the setting in question?

Lukas0907: I think that could be handy, maybe:

```python
raise NotConfigured('Please configure {} in the project settings'.format(name))
```

Gallaecio:
To be noted that @Lukas0907 use-case requires a custom error message, so the approach here would not be that helpful, unless we make changes that may over-complicate it.

PR #4494, opened 2020-04-15
Add licenses for artwork and trademarks

Gallaecio:


kmike:
> The 2 bullet points relating to codes of conduct are removed, given we do not have one yet.

https://github.com/scrapy/scrapy/blob/master/CODE_OF_CONDUCT.md ?

pablohoffman:
Sorry, are we saying that Scrapinghub owns Scrapy artwork?. That doesn't sound right to me. The logo was designed by Ezequiel Rivero while working at Insophia, for example.

Just to clarify, when I mentioned in #2351:

> In this case, the lack of foundation can be replaced by Scrapinghub as the legal representative.

I was only referring to making Scrapinghub the legal contact (for any questions/disputes), not transferring copyrights from Scrapy community to Scrapinghub.

Gallaecio:
I think we should mention that it was created by Ezequiel Rivero. I did not know who the author was. For proper attribution, I think we should also include the year when the logo was created and an email address of the author.

If Ezequiel wishes the artwork to be under a given artwork license, I think we can cover that in the `LICENSE` file.

But the logo represents Scrapy, and as such I think we need to treat it as a trademark, and hence the trademark license would take precedence over the artwork license (people would have to abide by *both* licenses).

We could alternatively not consider the logo a trademark, and opt out from enforcing people not to use the logo for thing unrelated to Scrapy. In that case, we have to assume that, in the unlikely scenario that an alternative scraping framework decides to use our logo, we would have no case against it.

PR #4499, opened 2020-04-15
Cover Scrapy 2.1 in the release notes

Gallaecio:


docs/news.rst
```diff
@@ -3,6 +3,135 @@
...
+
+*   The ``MultiValueDict``, ``MultiValueDictKeyError``, and ``SiteNode``
+    classes have been removed from :mod:`scrapy.utils.datatypes`
+    (:issue:`4400`)
+
+
+Deprecations
+~~~~~~~~~~~~
+
+*   :meth:`Spider.make_requests_from_url
```
kmike, line 49: It is deprecated a long time ago; I wonder if we can use a different wording.

Gallaecio: I’ve moved it to the bug fixes section, which I guess it is technically what we did, fix the deprecation reporting.

docs/news.rst
```diff
@@ -3,6 +3,145 @@
...
+
+*   The ``REDIRECT_MAX_METAREFRESH_DELAY`` setting is no longer supported, use
+    :setting:`METAREFRESH_MAXDELAY` instead (:issue:`4385`)
+
+*   The :class:`~scrapy.downloadermiddlewares.chunked.ChunkedTransferMiddleware`
+    middleware has been removed, including the entire
+    :class:`scrapy.downloadermiddlewares.chunked` module; chunked transfers
+    work out of the box (:issue:`4431`)
+
+*   The ``spiders`` property has ben removed from
```
kmike, line 36:
```suggestion
*   The ``spiders`` property has been removed from
```

Gallaecio:
I though about listing the exceptions and the circumstances under which they arise, but since none of them were documented before (and I think some of them cannot even be caught), I think these 2 paragraphs may be enough.

docs/news.rst
```diff
@@ -3,6 +3,153 @@
...
+Backward-incompatible changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+*   :exc:`AssertionError` exceptions triggered by :ref:`assert <assert>`
+    statements have been replaced by new exception types, to support running
+    Python in optimized mode (see :option:`-O`) without changing Scrapy’s
+    behavior in any unexpected ways.
+
+    If you catch an :exc:`AssertionError` exception from Scrapy, update your
+    code to catch the corresponding new exception.
```
kmike, line 25: I think there is one exception from this - scrapy contracts still raise an AssertionError subclass. So technically, someone could be catching AssertionError from scrapy contracts. I'm not sure when it is the case, so maybe the current wording is fine.

PR #4500, opened 2020-04-15
Serialize requests with callback references as spider attribute

victor-torres:
You could define a spider attribute that references a callback method
but if this method has a different name than your spider attribute,
the request serializer is not able to find it on the spider class.

With this commit we're fixing this behavior as we're searching for
callback references in the spider object itself instead of looking
for attributes with the same function's name, that could be different.

### DIFF
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -80,6 +80,13 @@ def _find_method(obj, func):
             if func_self is obj:
                 members = inspect.getmembers(obj, predicate=inspect.ismethod)
                 for name, obj_func in members:
+                    # We need to use __func__ to access the original
+                    # function object because instance method objects
+                    # are generated each time attribute is retrieved from
+                    # instance.
+                    #
+                    # Reference: The standard type hierarchy
+                    # https://docs.python.org/3/reference/datamodel.html
                     if obj_func.__func__ is func.__func__:
                         return name
     raise ValueError("Function %s is not a method of: %s" % (func, obj))
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -70,20 +70,6 @@ def request_from_dict(d, spider=None):
     )
 
 
-def _is_private_method(name):
-    return name.startswith('__') and not name.endswith('__')
-
-
-def _mangle_private_name(obj, func, name):
-    qualname = getattr(func, '__qualname__', None)
-    if qualname is None:
-        classname = obj.__class__.__name__.lstrip('_')
-        return '_%s%s' % (classname, name)
-    else:
-        splits = qualname.split('.')
-        return '_%s%s' % (splits[-2], splits[-1])
-
-
 def _find_method(obj, func):
     if obj:
         try:
@@ -95,8 +81,6 @@ def _find_method(obj, func):
                 members = inspect.getmembers(obj, predicate=inspect.ismethod)
                 for name, obj_func in members:
                     if obj_func.__func__ is func.__func__:
-                        if _is_private_method(name):
-                            return _mangle_private_name(obj, func, name)
                         return name
     raise ValueError("Function %s is not a method of: %s" % (func, obj))
 
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -1,6 +1,8 @@
 """
 Helper functions for serializing (and deserializing) requests.
 """
+import inspect
+
 from scrapy.http import Request
 from scrapy.utils.python import to_unicode
 from scrapy.utils.misc import load_object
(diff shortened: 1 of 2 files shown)


scrapy/utils/reqser.py
```diff
@@ -90,10 +92,12 @@ def _find_method(obj, func):
             pass
         else:
             if func_self is obj:
-                name = func.__func__.__name__
-                if _is_private_method(name):
-                    return _mangle_private_name(obj, func, name)
-                return name
+                members = inspect.getmembers(obj, predicate=inspect.ismethod)
+                for name, obj_func in members:
+                    if obj_func.__func__ is func.__func__:
```
kmike, line 97: Could you please explain (add a comment) why isn't it `if obj_func is func`?

scrapy/utils/reqser.py
```diff
@@ -90,10 +92,12 @@ def _find_method(obj, func):
...
+                        if _is_private_method(name):
+                            return _mangle_private_name(obj, func, name)
```
kmike, line 99: @victor-torres could you please check coverage report (https://codecov.io/gh/scrapy/scrapy/compare/36abe9235c7e85f89fe80ae726ca2a72b8ed7971...47a992615a046d20c31bfca6f8b65c3194e9fd30/diff) - it shows that this line is never executed. I'm not sure if that's coverage report which is wrong, but it worths investigating.

scrapy/utils/reqser.py
```diff
@@ -90,10 +92,12 @@ def _find_method(obj, func):
             pass
         else:
             if func_self is obj:
-                name = func.__func__.__name__
-                if _is_private_method(name):
-                    return _mangle_private_name(obj, func, name)
-                return name
+                members = inspect.getmembers(obj, predicate=inspect.ismethod)
```
kmike, line 95: I wonder what's the performance overhead of the change - would it be possible to run a micro-benchmark, e.g. timeit for _find_method, for a typical spider from some project, before and after the change?

scrapy/utils/reqser.py
victor-torres: Very interesting. It seems like this code is not needed anymore because we're getting the already mangled name when matching func with spider attributes.

scrapy/utils/reqser.py
victor-torres:
```
We need to use __func__ to access the original
function object because instance method objects
are generated each time attribute is retrieved from
instance.

Reference: The standard type hierarchy 
https://docs.python.org/3/reference/datamodel.html
```

scrapy/utils/reqser.py
victor-torres:
```python
def test_timeit():
    import timeit

    spider = TestSpider()

    def new_find():
        _find_method(spider, spider._TestSpider__parse_item_private)

    def old_find():
        _old_find_method(spider, spider._TestSpider__parse_item_private)

    new = timeit.timeit(new_find, number=10_000)
    old = timeit.timeit(old_find, number=10_000)
    import pdb; pdb.set_trace()
```

```
(Pdb) new
0.338965222
(Pdb) old
0.010640979000000161
(Pdb) 
```

victor-torres: The result is "a float measured in seconds".

victor-torres:
I've just updated this pull request. Could you take a look at the coverage results, please? I suspect they have dropped because I've had removed some code, but I'm not sure. @kmike

scrapy/utils/reqser.py
Gallaecio: It does not look like a significant run time to me in absolute terms. Although I wonder if we could optimize for previously-supported scenarios, and only have a significant (in relative terms) increase in run time for the cases which are now supported for the first time.

kmike: The overhead looks workable, at least for CPython. For PyPy it can be worse, as I recall inspect calls break their jit

Adding cache can complicate things, e.g. when someone decides to add methods dynamically. I guess a fast path can be checking function name first - but it is problematic, because introduces the same issue as fixed by this PR (it is slightly better, but a function named `parse` can be matched to a parse callback even if it is assigned to a different attribute). I'm fine with doing nothing here.

kmike:
@victor-torres the coverage we're measuring here is a coverage of the diff, not coverage of the project - coverage of the project can drop because of code removal, and that's the reason coverage of the diff is used for the check. 

The idea of the check is to ask for the modified code (only the part which is modified) to have coverage >= avg project coverage. This encourages all new and updated code to be no worse than the whole project, in terms of coverage.

If you click on "Details" near the report, and then on "Diff", you should be able to see where the drop comes from: it is in `for name, obj_func in members` statement. I'm not sure, but it seems branch coverage is yellow because members is never empty (?), which is fine.

scrapy/utils/reqser.py
victor-torres: @Gallaecio it's possible to use something like this:

```python
def _conditional_find_method(obj, func):
    old_name = _old_find_method(obj, func)
    if _get_method(obj, old_name).__func__ is func.__func__:
        return old_name
    else:
        return  _find_method(obj, func)
```

It will only inspect and traverse object members if the old method detects a false positive (divergency in names). 

Although, I'm not sure it's the best approach because we would be duplicating logic trying to fix a performance problem that we are not sure will exist.

victor-torres: Results using the same test case:

```
(Pdb) new
0.32308619000000005
(Pdb) old
0.010666135999999993
(Pdb) cond
0.014966982000000018
(Pdb) 
```

It would be like the old method if the function name matches the object attribute and like the new method if there's a divergence on names.

Gallaecio: > We would be duplicating logic trying to fix a performance problem that we are not sure will exist.

Indeed. Let’s just use the new implementation, then.

kmike:
Thanks @victor-torres and @Gallaecio!

PR #4503, opened 2020-04-16
Fix the hoverxref configuration

### LINKED ISSUE #4496: Fix the hoverxref configuration
> You shouldn't override hoverxref_version and hoverxref_project since they are taken automatically from Read the Docs.
>
> If you want to avoid your CI failing because of this, you can define the environment variables as Read the Docs does:
> 
> READTHEDOCS_PROJECT=scrapy
> READTHEDOCS_VERSION=''
> 
> With the current configuration, all the versions built on Read the Docs will point to a different version on Read the Docs and this will conflict. For example, current master version in Read the Docs defines hoverxref_version='2.0.0' but that version does not exist on Read the Docs and the tooltip does not known where to get the content from.

@humitos at https://github.com/scrapy/scrapy/pull/4480#discussion_r409026912

Gallaecio:
Fixes #4496

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -295,8 +295,6 @@ intersphinx_mapping = {
 # ------------------------------------
 
 hoverxref_auto_ref = True
-hoverxref_project = "scrapy"
-hoverxref_version = release
 hoverxref_role_types = {
     "class": "tooltip",
     "confval": "tooltip",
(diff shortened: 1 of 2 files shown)


kmike:
Thanks @Gallaecio!

PR #4504, opened 2020-04-16
Upgrade PyPy for CI, and test both 3.5 (oldest) and 3.6 (newest)

Gallaecio:


Gallaecio:
@wRAR [testPayloadDefaultCiphers](https://github.com/scrapy/scrapy/commit/aaa5229e5db4f1c014ae69e0fb9e1a933f4952b0#diff-d0a761d246cc22421c5d7e0b784b2665R419) fails on PyPy (it unexpectedly works). Any ideas? Maybe we need to change the value of `custom_ciphers` to something other than `'CAMELLIA256-SHA'`?

wRAR:
I need to check with Wireshark again what does it actually use to establish the connection (which shouldn't have been established). I think last time it failed to disable TLS1.2+ which we do in the test but I'm not sure.

kmike:
hey! What caused pypy tests to pass?

.travis.yml
```diff
@@ -37,8 +38,8 @@ matrix:
 install:
   - |
       if [ "$TOXENV" = "pypy3" ]; then
-        export PYPY_VERSION="pypy3.5-5.9-beta-linux_x86_64-portable"
```
kmike, line 40: Do you think "Installing Scrapy" section should be updated, with the new minimal pypy version?

Gallaecio:
> hey! What caused pypy tests to pass?

They didn’t, because of some Travis CI downtime it just seemed that way.

.travis.yml
Gallaecio: Makes sense.

Once we fix the remaining issue, though, we should probably try and find the lowest version where tests pass.

kmike: I'm not sure we should care too much about supporting old pypy versions; I think it is for advanced users to run Scrapy on pypy, and they could install a recent version anyways, to get best performance & compatibility. That should be fine to pick some version which is fresh enough, and doesn't cause us problems. Of course, knowing a specific pypy version is better, but I'd not put too much effort in figuring it out - if that's easy, let's do it, otherwise - don't bother.

Gallaecio:
mitmproxy [does not officially support PyPy](https://github.com/mitmproxy/mitmproxy/issues/2014), so I guess we better skip tests that require it on PyPy.

Gallaecio:
I wonder if we should mark the 2 remaining issues as expected failure and create separate issues for them, to move this along.

.travis.yml
Gallaecio: PyPy 5.9 continues to work.

I’ve also moved the FAQ entry about supported Python versions to `install.rst`, to avoid duplication. The previous information was accurate, though, so if you have any second thoughts about the documentation changes, we can revert that commit and leave the documentation _as is_.

setup.py
```diff
@@ -21,7 +21,16 @@ def has_environment_marker_platform_impl_support():
...
+    extras_require[':platform_python_implementation == "CPython"'] = [
+        'lxml>=3.5.0',
+    ]
     extras_require[':platform_python_implementation == "PyPy"'] = [
+        # Earlier lxml versions are affected by
+        # https://bitbucket.org/pypy/pypy/issues/2498/cython-on-pypy-3-dict-object-has-no,
+        # which was fixed in Cython 0.26, released on 2017-06-19, and used to
+        # generate the C headers of lxml release tarballs published since then, the
+        # first of which was:
+        'lxml>=4.0.0',
```
Gallaecio, line 33: It is technically possible for someone to install an older version if they re-generate the C headers with Cython 0.26+. If they are willing to do that, however, they can also change this number here locally for a Scrapy version supporting that.

kmike:
Thanks @Gallaecio!

noviluni:
@Gallaecio, from what I read in the last comment here: https://github.com/scrapy/scrapy/pull/2333, as this has been merged, you should probably check that PR again

PR #4506, opened 2020-04-17
Support nested items in exporters

### LINKED ISSUE #3888: Exporters should support nested scrapy.Item objects
If you were to build a nested `scrapy.Item` structure, such as:

```python
import scrapy

class TopLevel(scrapy.Item):
    foo = scrapy.Field()
    bars = scrapy.Field()

class Bar(scrapy.Item):
    spam = scrapy.Field(serializer=str.upper)

item = TopLevel(foo=42, bars=[Bar(spam='hello'), Bar(spam='world')])
```

one might expect the field configuration for `Bar` to be honoured when exporting. Unfortunately, that's not the case:

```python
import sys
from scrapy.exporters import XmlItemExporter
exporter = XmlItemExporter(sys.stdout, indent=2)
exporter.start_exporting()
exporter.export_item(item)
exporter.finish_exporting()
```

produces

```xml
<?xml version="1.0" encoding="utf-8"?>
<items>
  <item>
    <foo>42</foo>
    <bars>
      <value>
        <spam>hello</spam>
      </value>
      <value>
        <spam>world</spam>
      </value>
    </bars>
  </item>
</items>
```

I worked around this [for this Stack Overflow question](https://stackoverflow.com/q/57110413/100297) by subclassing the exporter and passing along enough information to the serializer to recurse in the serializer itself:

```python
class RecursingXmlItemExporter(XmlItemExporter):
    def _recursive_serialized_fields(self, item):
        if isinstance(item, scrapy.Item):
            return dict(self._get_serialized_fields(item, default_value=''))
        return item

    def serialize_field(self, field, name, value):
        serializer = field.get('serializer', lambda x: x)
        try:
            return serializer(value, self._recursive_serialized_fields)
        except TypeError:
            return serializer(value)
```

The above hides the private `BaseItemExporter._get_serialized_fields()` API, but still adaptable to how the different exporter classes call this method to meet their specific needs. 

However, this does require every single nested field item to use a custom serializer. Ideally this is handled natively, ideally as fields are traversed over in `export_item()`.

michalp2213:
I believe this closes #3888 . I've also included a test case for this functionality.

Fixes #3888.


Gallaecio:
I haven’t look into it in detail, but this will need changes as soon as we merge https://github.com/scrapy/scrapy/pull/3881, which should happen shortly.

AdrianAtZyte (approved):

PR #4507, opened 2020-04-17
Docs: replace deprecated FEED_FORMAT and FEED_URI settings

elacuesta:
Update the docs after #3858


kmike:
A good catch @elacuesta!

Issue #4346, opened 2020-02-19
FilesDownloader with GCS downloading updtodate files again


lblanche:
# Description

It seems that when using Google Cloud Storage, the Files pipeline does not have the expected behavior regarding up to date files.

### Steps to Reproduce

1. Clone this repo : `git clone https://github.com/QYQ323/python.git`
2. Run the spider : `scrapy crawl examples`
3. If you run it several times, the FilesPipeline has the right behavior : it does not download uptodate files 

```
2020-02-19 14:41:36 [scrapy.pipelines.files] DEBUG: File (uptodate): Downloaded file from <GET https://matplotlib.org/examples/animation/basic_example_writer.py> referred in <None>
2020-02-19 14:41:36 [scrapy.pipelines.files] DEBUG: File (uptodate): Downloaded file from <GET https://matplotlib.org/examples/animation/basic_example.py> referred in <None>
2020-02-19 14:41:36 [scrapy.pipelines.files] DEBUG: File (uptodate): Downloaded file from <GET https://matplotlib.org/examples/animation/bayes_update.py> referred in <None>
```

4. Now change the `FILE_STORE` in `settings.py` to a gcs bucket

 `FILES_STORE = 'gs://mybucket/' `

5. If you then run the spider several times, the files are downloaded everytime : 
```
2020-02-19 14:50:44 [scrapy.pipelines.files] DEBUG: File (downloaded): Downloaded file from <GET https://matplotlib.org/examples/animation/simple_anim.py> referred in <None>
2020-02-19 14:50:44 [urllib3.connectionpool] DEBUG: https://storage.googleapis.com:443 "POST /upload/storage/v1/b/cdcscrapingresults/o?uploadType=multipart HTTP/1.1" 200 843
2020-02-19 14:50:44 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://matplotlib.org/examples/animation/double_pendulum_animated.py> (referer: None)
2020-02-19 14:50:44 [scrapy.pipelines.files] DEBUG: File (downloaded): Downloaded file from <GET https://matplotlib.org/examples/animation/double_pendulum_animated.py> referred in <None>
2020-02-19 14:50:44 [scrapy.core.scraper] DEBUG: Scraped from <200 https://matplotlib.org/examples/api/collections_demo.html>
```

**Expected behavior:** 
[...]

PR #4508, opened 2020-04-19
Add a warning/error in case of incorrect gcs permissions

michalp2213:
Added a warning in case of lack of read access to GCS bucket used with `FilePipeline` and error in case of lack of write access (as it makes it impossible to upload files). Incorrect GCS permissions could be the cause of a problem described in #4346 .

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -229,6 +229,19 @@ class GCSFilesStore:
         bucket, prefix = uri[5:].split('/', 1)
         self.bucket = client.bucket(bucket)
         self.prefix = prefix
+        permissions = self.bucket.test_iam_permissions(
+            ['storage.objects.get', 'storage.objects.create']
+        )
+        if 'storage.objects.get' not in permissions:
+            logger.warning(
+                "No 'storage.objects.get' permission for GSC bucket %(bucket)s",
+                {'bucket': bucket}
+            )
+        if 'storage.objects.create' not in permissions:
+            logger.error(
+                "No 'storage.objects.create' permission for GSC bucket %(bucket)s. Saving files will be impossible!",
+                {'bucket': bucket}
+            )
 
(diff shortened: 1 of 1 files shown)


scrapy/pipelines/files.py
```diff
@@ -229,6 +229,19 @@ def __init__(self, uri):
         bucket, prefix = uri[5:].split('/', 1)
         self.bucket = client.bucket(bucket)
         self.prefix = prefix
+        permissions = self.bucket.test_iam_permissions(
+            ['storage.objects.get', 'storage.objects.create']
+        )
+        if 'storage.objects.get' not in permissions:
+            logger.warning(
+                "No 'storage.objects.get' permission for GSC bucket %(bucket)s",
```
kmike, line 237: What do you think about adding more details about this problem - explaining that it would cause the file to be downloaded every time.

michalp2213: Just pushed a commit. I hope it's fine now.

scrapy/pipelines/files.py
```diff
@@ -229,6 +229,20 @@ def __init__(self, uri):
         bucket, prefix = uri[5:].split('/', 1)
         self.bucket = client.bucket(bucket)
         self.prefix = prefix
+        permissions = self.bucket.test_iam_permissions(
+            ['storage.objects.get', 'storage.objects.create']
+        )
+        if 'storage.objects.get' not in permissions:
+            logger.warning(
+                "No 'storage.objects.get' permission for GSC bucket %(bucket)s. \
+                Checking if files are up to date will be impossible. Files will be downloaded every time.",
```
Gallaecio, line 238:
```suggestion
                "No 'storage.objects.get' permission for GSC bucket %(bucket)s. "
                "Checking if files are up to date will be impossible. Files will be downloaded every time.",
```

Gallaecio:
Thank you!

PR #716, opened 2014-05-12
  Added --overwrite-output (-O) option

yhager:
How about this for #547? This is my first time writing code for scrapy, so please scrutinize.

Fixes #547

PR #4512, opened 2020-04-24
Allow overwriting feeds

### LINKED ISSUE #547: Add a command-line option for overwriting exported file

Gallaecio:
This is a fork of #716.

I’ve aimed for backward-compatible behavior (hence `overwrite` has a different default value for the local filesystem and FTP storages). ~Still, there is a backward-incompatible change to `IFeedStorage.__init__`; on the bright side, it will fail loudly and be easy to fix (add the new parameter to `__init__` and ignore it otherwise).~ 

On FTP, in addition to adding support for `APPE` operations, I:
- Fixed a bug when a password is not specified (hence `None`, failing when passed to `unquote`).
- Included a mock FTP server in the tests, so that they always run. I wanted to use Twisted, but its FTP server [does not support `APPE` operations](https://twistedmatrix.com/trac/ticket/4819).

To do:
- [x] Fix tests

Fixes #547, closes #716.


kmike:
@Gallaecio could you please check the tutorial - can it be updated with this feature? See https://docs.scrapy.org/en/latest/intro/tutorial.html#storing-the-scraped-data

docs/topics/feed-exports.rst
```diff
@@ -262,15 +263,41 @@ For instance::
...
+    <topics-feed-storage-backends>`:
+
+    -   :ref:`topics-feed-storage-fs`: ``False``
+
+    -   :ref:`topics-feed-storage-ftp`: ``True``
+
+        .. note:: Some FTP servers may not support appending to files (the
+                  ``APPE`` FTP command).
+
+    The remaining storage backends do not support overwritting the output file:
```
kmike, line 293: s3 supports overwriting, it doesn't support appending. Maybe show the default values in the list above, and mention that the default can't be changed?

docs/intro/tutorial.rst
```diff
@@ -464,16 +464,15 @@ Storing the scraped data
...
 
-For historic reasons, Scrapy appends to a given file instead of overwriting
-its contents. If you run this command twice without removing the file
-before the second time, you'll end up with a broken JSON file.
-
-You can also use other formats, like `JSON Lines`_::
+The ``-O`` command-line switch overwrites any existing file; use ``-o`` instead
+to append new content to any existing file. However, appending to a JSON file
+makes the file contents invalid JSON. When appending to a file, consider
+using a different serialization format, such as `JSON Lines`_::
```
kmike, line 475: Could you please also check "Using spider arguments" section? It uses -o with .json file; it is probably better to either use -o with .jl file, or use -O with .json file.

scrapy/commands/crawl.py
```diff
@@ -19,6 +19,8 @@ def add_options(self, parser):
                           help="set spider argument (may be repeated)")
         parser.add_option("-o", "--output", metavar="FILE", action="append",
                           help="dump scraped items into FILE (use - for stdout)")
+        parser.add_option("-O", "--overwrite-output", metavar="FILE", action="append",
+                          help="overwrite scraped items into FILE")
```
kmike, line 23: I wonder if it'd be more clear if help text for -o and -O is similar. E.g. (not sure about wording):

```suggestion
        parser.add_option("-o", "--output", metavar="FILE", action="append",
                          help="dump scraped items into FILE, appending them in the end (use - for stdout)")
        parser.add_option("-O", "--overwrite-output", metavar="FILE", action="append",
                          help="dump scraped items into FILE, overwrite the file")
```

scrapy/utils/conf.py
```diff
@@ -141,6 +142,13 @@ def check_valid_format(output_format):
                              " or as a file extension, from the supported list %s" %
```
kmike, line 142: should -O be mentioned in the error above as well?

scrapy/utils/conf.py
```diff
@@ -141,6 +142,13 @@ def check_valid_format(output_format):
...
+    overwrite = False
+    if overwrite_output:
+        if output:
+            raise UsageError("Please use only one of --output and --overwrite-output")
+        output = overwrite_output
+        overwrite = True
+
     if output_format:
         if len(output) == 1:
             check_valid_format(output_format)
```
kmike, line 154: the warning below only mentions -o

scrapy/utils/conf.py
```diff
@@ -141,6 +142,13 @@ def check_valid_format(output_format):
                              " or as a file extension, from the supported list %s" %
                              (output_format, tuple(valid_output_formats)))
 
+    overwrite = False
+    if overwrite_output:
+        if output:
+            raise UsageError("Please use only one of --output and --overwrite-output")
```
elacuesta, line 148:
```suggestion
            raise UsageError("Please use only one of -o/--output and -O/--overwrite-output")
```
Or something similar. I wonder if that would be more informative, perhaps not everybody knows exactly what `-o/-O` means.

docs/intro/overview.rst
```diff
@@ -46,14 +46,14 @@ http://quotes.toscrape.com, following the pagination::
...
-When this finishes you will have in the ``quotes.json`` file a list of the
+When this finishes you will have in the ``quotes.jl`` file a list of the
 quotes in JSON format, containing text and author, looking like this (reformatted
 here for better readability)::
 
-    [{
+    {
         "author": "Jane Austen",
         "text": "\u201cThe person, be it gentleman or lady, who has not pleasure in a good novel, must be intolerably stupid.\u201d"
     },
```
kmike, line 59: should it have a comma here? Also, it should be a line per record, which may be important to understand in order to be able to parse the file

Gallaecio: Updated from a local run of the spider, to be safe.

docs/intro/overview.rst
```diff
@@ -42,29 +42,18 @@ http://quotes.toscrape.com, following the pagination::
...
-
 Put this in a text file, name it to something like ``quotes_spider.py``
 and run the spider using the :command:`runspider` command::
 
     scrapy runspider quotes_spider.py -o quotes.jl
 
-
 When this finishes you will have in the ``quotes.jl`` file a list of the
 quotes in JSON format, containing text and author, looking like this (reformatted
 here for better readability)::
```
kmike, line 52: is it reformatted?

kmike:
> I’ve aimed for backward-compatible behavior (hence overwrite has a different default value for the local filesystem and FTP storages). Still, there is a backward-incompatible change to IFeedStorage.__init__; on the bright side, it will fail loudly and be easy to fix (add the new parameter to __init__ and ignore it otherwise).

I've checked our internal code search tool, and it is quite common to inherit from BlockingFeedStorage or S3FeedStorage to implement something custom, or just to call some storage class directly (maybe in some export script?). We could break a lot of projects by this change; even if a fix is easy, it could be an obstacle.

How hard would it be to get a deprecation message? It seems we've done similar things in past.

scrapy/extensions/feedexport.py
```diff
@@ -32,8 +32,9 @@
 class IFeedStorage(Interface):
     """Interface that all Feed Storages must implement"""
 
-    def __init__(uri):
-        """Initialize the storage with the parameters given in the URI"""
+    def __init__(uri, feed):
```
kmike, line 35: What do you think about calling it feed_options?

* for feed argument I'd expect some instance of Feed, whatever is it
* feed_settings is not god, because this dict is not a Settings instance

scrapy/extensions/feedexport.py
```diff
@@ -328,8 +363,40 @@ def _get_instance(self, objcls, *args, **kwargs):
...
+        def build_instance(builder, *args):
+            argument_names = get_func_args(builder)
+            if 'feed_options' in argument_names:
+                _kwargs = {'feed_options': feed_options}
+            else:
+                warnings.warn(
+                    "{} does not support the 'feed_options' keyword argument"
+                    .format(builder.__qualname__),
+                    category=ScrapyDeprecationWarning
+                )
```
kmike, line 384: This deprecation logic is repeated a few times, would it be possible to extract a helper?

tests/test_feedexport.py
```diff
@@ -969,3 +1056,232 @@ def test_pathlib_uri(self):
...
+        spider.crawler = crawler
+        with warnings.catch_warnings(record=True) as w:
+            feed_exporter.open_spider(spider)
+            messages = tuple(str(item.message) for item in w)
+            self.assertEqual(
+                messages,
+                (
+                    (
+                        "S3FeedStorageWithoutFeedOptionsWithFromCrawler.from_crawler "
+                        "does not support the 'feed_options' keyword argument"
```
kmike, line 1217: What do you think about telling people how to update their code in these messages, to make them disappear? We may also say explicitly that such custom storages are deprecated, and we'll stop supporting them in future releases. It can be easier to update all messages if they're issued in a single helper function.

scrapy/utils/python.py
```diff
@@ -194,7 +194,8 @@ def _getargspec_py23(func):
 def get_func_args(func, stripself=False):
     """Return the argument name list of a callable"""
     if inspect.isfunction(func):
-        func_args, _, _, _ = _getargspec_py23(func)
+        spec = inspect.getfullargspec(func)
+        func_args = spec.args + spec.kwonlyargs
```
kmike, line 198: Is it a cleanup or an issue fix? If that's an issue fix, could you please add a test case for get_func_args, even if without the fix a different part of scrapy fails?

docs/topics/feed-exports.rst
```diff
@@ -262,15 +263,41 @@ For instance::
...
+
+    -   :ref:`topics-feed-storage-s3`: `does not support appending to files
```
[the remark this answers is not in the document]
Gallaecio: It’s on purpose, they have different default values.

Gallaecio:
So, the tests were failing due to unclosed socket warnings randomly getting caught by the new tests. Since the issue is not easily reproducible, I’ve taken the approach to ignore those warnings in the tests that inspect warning content, and only take into account Scrapy deprecation warnings.

Issue #1777, opened 2016-02-14
Scrapy as a library


nyov:
Scrapy currently assumes in a lot of it's functionality to be used as a Framework.

Work has been done in the past, and is ongoing, to make it more usable as a library as well.
I would like to see even more change in this regard.
## 

Is it feasible to consider decoupling scrapy's django-like concept of a "project" in the filesystem from "scrapy core" -- possibly moving the behavior of creating a 'scrapy project' (and all that entails, such as file templates) into a separate project? Or the opposite of creating a `libscrapy` project for the bare essentials?
## 

A lot of scrapy's environment configuration is done through the `Settings` object.
All configurable classes/functions that are defined through settings (anything that reads like `scrapy.downloadermiddlewares.redirect.RedirectMiddleware`), get loaded through `load_object` which in turn leverages `importlib.import_module` to find a module and pull in the code.

In a scripted environment we usually just want to load an already defined or imported class into scrapy, such as an `ItemPipeline`, and I would like to see support for loading objects directly in some manner there.

**/edit**: Being discussed in #1215 (thanks @kmike)
## 

I hope this is not too "out there".
(And if someone has other `scrapy-as-a-library`-related considerations, they are welcome to add them here as well.

kmike:
Hey @nyov,

libscrapy is an interesting idea; what do you think should be in this package? 

Scrapy is different from many packages because most of its core features (e.g. redirects or cookies) are implemented as extensions/middlewares using the same extension points as users. If we bundle none of them to 'libscrapy' then libscrapy won't be useful; if we bundle all of them it'll be almost the same as scrapy. 

[...]

PR #4516, opened 2020-04-27
Fix #4005: Move ItemLoader to a new library

### LINKED ISSUE #4005: move item loaders to a separate library
## Summary

Move item loaders to a separate repository, similar to parsel.

## Motivation

There are two main reasons for doing so:

1. If item loaders are decoupled from Scrapy, they can be used without Scrapy. They would still support CSS/XPath selectors, but use parsel to populate dicts. For example, this would allow to use item loaders in PageObjects from https://github.com/scrapinghub/scrapy-po, and allow reusing them without Scrapy.
2. It would allow to develop item loaders separately, make it easier to have a different set of maintainers, allow a separate release cycle. Currently they are under-maintained  in Scrapy.

On a first sight, most of the item loaders code is not tied to Scrapy already.

## Previous attempts

We tried to move itemloaders in past, by copying the current implementation. Decision was not to go with this, as Scrapy will have to depend on scrapy-itemloaders, and scrapy-itemloaders will have to depend on Scrapy. In this proposal Scrapy depends on a new package, but package can be used without Scrapy - it depends on parsel only.

## How to do it

1. re-create https://github.com/scrapy/scrapy-itemloader repo - import current code (with git history). Or maybe use a different name for a repository - just "itemloaders"? We can rename the repo.
2. remove Scrapy-specific parts:
   *  "response" argument, response object in loader context
   * scrapy Item support. It can be re-added for attr.s or dataclass-based items, when Scrapy gets support for them.
3. implement scrapy.loaders on top of a new library - inherit, add response and item support. 
4. Provide backwards compatibility shims for scrapy.loader.processors

## Other considerations

Optional scrapy.Item support in the library itself can be fine.

ejulio:
Fix #4005

Installed the new library [`itemloaders`](https://github.com/scrapy/itemloaders) from github for now.
Remove tests that are already in `itemloaders`. Kept some that seem to be a bit specific to scrapy. Not completely on this removal. Thoughts?

Made scrapy's `ItemLoader` an extension of the base class from the library an added support for `Item`

Removed processors and imported them from `itemloaders`

Removed `common.py` with `wrap_loader_context` which moved to the new library

Before merging:

- [x] Define the new library name 
    - `itemloaders`
    - `data_loaders`: move away from scrapy's idea of items
    - `xhtml_loaders`: loads data from any XHTML text/parsel selector
- [x] Update scrapy docs mentioning the library


scrapy/loader/__init__.py
```diff
@@ -3,219 +3,35 @@
...
-class ItemLoader:
+class ItemLoader(itemloaders.ItemLoader):
 
     default_item_class = Item
-    default_input_processor = Identity()
-    default_output_processor = Identity()
     default_selector_class = Selector
 
-    def __init__(self, item=None, selector=None, response=None, parent=None, **context):
+    def __init__(self, response=None, item=None, selector=None, parent=None, **context):
```
Gallaecio, line 16: Can we keep the previous parameter order, for backward compatibility?

scrapy/loader/common.py
```diff
@@ -1,14 +0,0 @@
...
-
-
-def wrap_loader_context(function, context):
-    """Wrap functions that receive loader_context to contain the context
-    "pre-loaded" and expose a interface that receives only one argument
-    """
-    if 'loader_context' in get_func_args(function):
-        return partial(function, loader_context=context)
-    else:
-        return function
```
Gallaecio, line 14: To maximize backward compatibility, we might want to deprecate this module instead of removing it. Maybe we can import the function from the new library, though.

scrapy/loader/processors.py
```diff
@@ -1,104 +1 @@
...
-
-
-class Join:
-
-    def __init__(self, separator=u' '):
-        self.separator = separator
-
-    def __call__(self, values):
-        return self.separator.join(values)
+from itemloaders.processors import *
```
Gallaecio, line 1: We need to issue a deprecation warning, to try and get people importing from `itemloaders` directly in the long term.

victor-torres, line 1: Maybe it's better to import each processor individually. This way we keep a text reference to them instead of a dynamic "import all".

scrapy/loader/__init__.py
victor-torres, line 16: Any reason for changing arguments order here? It might break code that doesn't make use of named args.

tests/test_loader.py
```diff
@@ -630,367 +227,6 @@ def test_avoid_reprocessing_without_initial_values_list(self):
...
-    def test_init_method_with_selector_css(self):
-        sel = Selector(text=u"<html><body><div>marta</div></body></html>")
-        l = TestItemLoader(selector=sel)
-        self.assertIs(l.selector, sel)
-
-        l.add_css('name', 'div::text')
-        self.assertEqual(l.get_output_value('name'), [u'Marta'])
-
-    def test_init_method_with_response(self):
-        l = TestItemLoader(response=self.response)
```
kmike, line 758: I think we should be keeping tests like that. This removal probably needs a more detailed review: everything which is related to response, or e.g. scrapy.Item, should be tested, because these are features which Scrapy's wrapper provides, not features which itemloaders library provides.

I also wonder if many (most) other tests needs to be preserved: e.g. I can see how we could have implemented a scrapy ItemLoader subclass in a way which breaks itemloaders.ItemLoader's `selector` argument. Though we haven't done this for scrapy.Selector, so no strong opinion here.

scrapy/loader/processors.py
```diff
@@ -1,104 +1 @@
...
-                return value
-
-
-class Identity:
-
-    def __call__(self, values):
-        return values
-
-
-class SelectJmes:
```
kmike, line 78: If I'm not mistaken, this is the only place where jmespath is used in Scrapy; I wonder if its mentions should be removed (from docs, from CI?).

kmike:
As for the package name, my vote is for `itemloaders`.

scrapy/loader/processors.py
wRAR: We don't do a similar thing for `parsel`, do we?

Gallaecio: Because we have custom `Selector` subclasses in that case.

kmike: we do it for w3lib imports, e.g. https://github.com/scrapy/scrapy/blob/master/scrapy/utils/http.py (some other cases like that have been removed)

kmike: or https://github.com/scrapy/scrapy/blob/master/scrapy/utils/markup.py

scrapy/loader/common.py
ejulio: I guess keeping the import is better than adding a deprecation, thoughts?
Also, I removed it as it doesn't appear in the docs https://docs.scrapy.org/en/latest/topics/loaders.html, so it shouldn't be used outside `ItemLoader`, or am I missing something here?

scrapy/loader/processors.py
ejulio: I guess this is fine, though, for any new processor, we'll need to update this piece of code.
Unless, we import them with some deprecation warning as mentioned above.

ejulio: Here we have both options.
Consider a deprecation and help people move to `itemloaders`, or just import it as part of `scrapy`.
Which one is preferred?

tests/test_loader.py
ejulio: Makes sense to me.
The ones I removed were related to a usage of `response` in which I transformed to `selector` in `itemloaders`.
But I'm in favor of keeping the ones referencing `response`..

scrapy/loader/processors.py
ejulio: Removed the dependency.
I'll remove from the docs once I update it referencing the librayr

ejulio:
@kmike @Gallaecio @wRAR @victor-torres 
Thanks for the comments.
JUst left some thoughts on the deprecation and imports from the new library.
Once we get it sorted out, I'll update it.

scrapy/loader/processors.py
kmike: I think we should issue a deprcation warning and ask people to use itemloaders.

scrapy/loader/common.py
kmike: Yes, an import + deprecation warning. 

We're usually trying to be quite careful about deprecations, and even parts which were not publicly documented are going through deprecation process. 

Here you can also make an argument that the module is not private (not _common), and the function  is not private (not _wrap_loader_context).

scrapy/loader/processors.py
Gallaecio: Actually, since the goal is for people to import directly from `itemloaders` (which is why this module should yield a warning when imported), even if `itemloaders` adds new processors we would not want them imported here.

ejulio:
@victor-torres @kmike @Gallaecio @wRAR 
I updated the changes to log warnings whenever something that was ported is used.
Not sure if that's the best approach, so fell free to raise any concerns about it.

Also, updated the documentation removing most of the content.
The idea is to include a link to `itemloaders` library docs once it is up and running.
Regarding the docs, not sure if we should keep some sort of TOC pointing to stuff that was moved from scrapy..

docs/topics/loaders.rst
```diff
@@ -179,595 +114,3 @@ output processors are declared using the ``_out`` suffix. And you can also
...
-   them::
-
-       class ProductLoader(ItemLoader):
-           length_out = MapCompose(parse_length, unit='cm')
-
-
-ItemLoader objects
-==================
-
-.. class:: ItemLoader([item, selector, response], \**kwargs)
```
kmike, line 276: Is this signature documented anywhere now?

docs/topics/loaders.rst
```diff
@@ -179,595 +114,3 @@ output processors are declared using the ``_out`` suffix. And you can also
...
-
-    Return a new Item Loader for populating the given Item. If no item is
-    given, one is instantiated automatically using the class in
-    :attr:`default_item_class`.
-
-    When instantiated with a ``selector`` or a ``response`` parameters
```
kmike, line 282: I'm not sure this is documented anywhere, after the removal

docs/topics/loaders.rst
```diff
@@ -179,595 +114,3 @@ output processors are declared using the ``_out`` suffix. And you can also
...
-        :meth:`~ItemLoader.add_xpath`, :meth:`~ItemLoader.add_css`,
-        or :meth:`~ItemLoader.add_value`.
-    :type item: :class:`~scrapy.item.Item` object
-
-    :param selector: The selector to extract data from, when using the
-        :meth:`add_xpath` (resp. :meth:`add_css`) or :meth:`replace_xpath`
-        (resp. :meth:`replace_css`) method.
-    :type selector: :class:`~scrapy.selector.Selector` object
-
-    :param response: The response used to construct the selector using the
```
kmike, line 296: this is not documented now

docs/topics/loaders.rst
```diff
@@ -161,10 +92,14 @@ Item Loaders are declared like Items, by using a class definition syntax. Here
...
-    from scrapy.loader.processors import TakeFirst, MapCompose, Join
+    from itemloaders.processors import TakeFirst, MapCompose, Join
+
+    from myproject.items import ProductItem
 
     class ProductLoader(ItemLoader):
 
+        default_item_class = ProductItem
+
         default_output_processor = TakeFirst()
```
kmike, line 103: I think this example is not clear after processors docs are removed

scrapy/loader/processors.py
```diff
@@ -4,101 +4,61 @@
...
+    """ ATTENTION: When removing these references, also remove tests/test_loader_deprecated.py"""
+    warnings.warn(
+        f"{cls.__module__}.{cls.__name__} has moved to a new library."
+        f"Please update your reference to itemloaders.processors.{cls.__name__}",
+        ScrapyDeprecationWarning
+    )
 
 
-class MapCompose:
+class MapCompose(processors.MapCompose):
```
kmike, line 25: I wonder if you can use scrapy.utils.deprecate.create_deprecated_class here

docs/topics/loaders.rst
```diff
@@ -179,595 +114,3 @@ output processors are declared using the ``_out`` suffix. And you can also
...
-        :attr:`default_selector_class`, unless the selector argument is given,
-        in which case this argument is ignored.
-    :type response: :class:`~scrapy.http.Response` object
-
-    The item, selector, response and the remaining keyword arguments are
-    assigned to the Loader context (accessible through the :attr:`context` attribute).
-
-    :class:`ItemLoader` instances have the following methods:
-
-    .. method:: get_value(value, \*processors, \**kwargs)
```
kmike, line 306: A proposal: to avoid copy-paste, but still have docs for these methods in Scrapy itself, make them docstrings in itemloaders repo, and then use autodocs. If I'm not mistaken, this is what we've done with scrapy.Selector, when it was moved to parsel.

docs/topics/loaders.rst
ejulio: This one is a bit tricky.
I included a "link" to `itemloaders` doc after the example.
Not sure if it's the best approach

docs/topics/loaders.rst
```diff
@@ -68,8 +68,11 @@ previously extracted and collected with the :meth:`~ItemLoader.add_xpath`,
 Declaring Item Loaders
 ======================
 
-Item Loaders are declared like Items, by using a class definition syntax. Here
-is an example::
+Besides the usage of Item Laoders as instances, like the example above,
```
elacuesta, line 71:
```suggestion
Besides the usage of Item Loaders as instances, like the example above,
```

docs/topics/loaders.rst
```diff
@@ -77,91 +64,18 @@ called which actually returns the item populated with the data
...
-<topics-loaders-available-processors>` built-in for convenience.
-
-
 
 Declaring Item Loaders
 ======================
 
-Item Loaders are declared like Items, by using a class definition syntax. Here
-is an example::
+Besides the usage of Item Laoders as instances, like the example above,
```
elacuesta, line 71:
```suggestion
Besides the usage of Item Loaders as instances, like the example above,
```

docs/conf.py
```diff
@@ -288,6 +288,7 @@
     'tox': ('https://tox.readthedocs.io/en/latest', None),
     'twisted': ('https://twistedmatrix.com/documents/current', None),
     'twistedapi': ('https://twistedmatrix.com/documents/current/api', None),
+    'itemloaders': ('https://itemloaders.readthedocs.io/en/latest/', None),
```
Gallaecio, line 291: :lipstick: Let’s keep it alphabetical :nerd_face:

scrapy/loader/common.py
```diff
@@ -1,14 +1,19 @@
 """Common functions used in Item Loaders code"""
 
 from functools import partial
+import warnings
```
Gallaecio, line 4: :lipstick: 

```suggestion
import warnings
from functools import partial
```

scrapy/loader/common.py
```diff
@@ -1,14 +1,19 @@
...
+
+from itemloaders import common
+
 from scrapy.utils.python import get_func_args
+from scrapy.utils.deprecate import ScrapyDeprecationWarning
```
Gallaecio, line 9: :lipstick: 

```suggestion
from scrapy.utils.deprecate import ScrapyDeprecationWarning
from scrapy.utils.python import get_func_args
```

scrapy/loader/common.py
```diff
@@ -1,14 +1,19 @@
...
 
 
 def wrap_loader_context(function, context):
-    """Wrap functions that receive loader_context to contain the context
-    "pre-loaded" and expose a interface that receives only one argument
-    """
```
Gallaecio, line 10: It should be OK to leave the docstring until we remove the function.

tests/test_loader_deprecated.py
```diff
@@ -0,0 +1,695 @@
+"""
```
Gallaecio, line 1: Given `ItemLoader` is no longer deprecated, it may make sense to restore the original file name.

ejulio: Maybe a better name, these are the tests that were moved to the new library.
The idea is to remove them once the deprecated stuff (processors, ...) is removed.
`test_loader_moved_itemloaders.py`?

Gallaecio: Or maybe we can leave the `test_loader_deprecated` name but add a `test_loader` with the tests covering Scrapy-specific features that will remain after those deprecations.

docs/topics/loaders.rst
```diff
@@ -174,600 +88,31 @@ is an example::
...
->>> il.add_value('name', [u'Welcome to my', u'<strong>website</strong>'])
->>> il.add_value('price', [u'&euro;', u'<span>1000</span>'])
->>> il.load_item()
-{'name': u'Welcome to my website', 'price': u'1000'}
-
-The precedence order, for both input and output processors, is as follows:
-
-1. Item Loader field-specific attributes: ``field_in`` and ``field_out`` (most
-   precedence)
-2. Field metadata (``input_processor`` and ``output_processor`` key)
```
kmike, line 223: I could be wrong, but it seems this information is absent in new docs.

scrapy/loader/common.py
```diff
@@ -1,14 +1,22 @@
...
     "pre-loaded" and expose a interface that receives only one argument
     """
-    if 'loader_context' in get_func_args(function):
-        return partial(function, loader_context=context)
-    else:
-        return function
+    warnings.warn(
+        "scrapy.loader.common.wrap_loader_context has moved to a new library."
+        "Please update your reference to itemloaders.common.wrap_loader_context",
+        ScrapyDeprecationWarning
```
kmike, line 19: should it use stacklevel=2?

docs/topics/loaders.rst
```diff
@@ -174,600 +88,31 @@ is an example::
...
-.. _topics-loaders-processors-declaring:
-
-Declaring Input and Output Processors
-=====================================
-
-As seen in the previous section, input and output processors can be declared in
-the Item Loader definition, and it's very common to declare input processors
-this way. However, there is one more place where you can specify the input and
-output processors to use: in the :ref:`Item Field <topics-items-fields>`
-metadata. Here is an example::
```
kmike, line 192: Is there an example of using Item meta for declaring of input and output processors in the new docs?

docs/topics/loaders.rst
```diff
@@ -174,600 +88,31 @@ is an example::
...
-        If a selector is given in the ``__init__`` method this attribute is ignored.
-        This attribute is sometimes overridden in subclasses.
-
-    .. attribute:: selector
-
-        The :class:`~scrapy.selector.Selector` object to extract data from.
-        It's either the selector given in the ``__init__`` method or one created from
-        the response given in the ``__init__`` method using the
-        :attr:`default_selector_class`. This attribute is meant to be
-        read-only.
```
kmike, line 517: Are the docs for loader.selector attribute available? They differ in itemloaders and scrapy.loader, because the selector can be created from response in scrapy's wrapper.

scrapy/loader/common.py
ejulio: TIL :+1:

docs/topics/loaders.rst
ejulio: It was moved to `itemloaders`, but the `Item` part was missing.

tests/test_loader_deprecated.py
ejulio: @Gallaecio , but there's already a file named `test_loader.py`, which is the past one, but I removed the tests that were covered by `itemloaders` (moved them to `test_loader_deprecated`)

Gallaecio: :facepalm: 

My bad, it did not load by default in the patch viewer and because it only had deleted lines in the count of added and deleted lines I though the whole file had been deleted.

docs/topics/loaders.rst
```diff
@@ -21,26 +21,13 @@ Item Loaders are designed to provide a flexible, efficient and easy mechanism
 for extending and overriding different field parsing rules, either by spider,
 or by source format (HTML, XML, etc) without becoming a nightmare to maintain.
 
+.. note:: Item Loaders used to be a part of ``scrapy``, but moved to a self-contained library.
+    This implementation is an extension of itemloaders_ library,
+    keeping support to :ref:`Items <topics-items>` and :ref:`Response <topics-request-response>`.
+
```
Gallaecio, line 27: I think this should be written more in line with the counterpart note in `selectors.rst`, talking only about the present, not the past.

docs/topics/loaders.rst
```diff
@@ -623,151 +388,7 @@ needs.
...
-    chain.
-
-    This processor provides a convenient way to compose functions that only
-    work with single values (instead of iterables). For this reason the
-    :class:`MapCompose` processor is typically used as input processor, since
-    data is often extracted using the
-    :meth:`~scrapy.selector.Selector.extract` method of :ref:`selectors
-    <topics-selectors>`, which returns a list of unicode strings.
+.. automodule:: itemloaders.processors
+    :members:
```
Gallaecio, line 392: Since this is non-Scrapy API reference documentation, I would be OK with removing this section, and instead updating the paragraph above which mentions that `itemloaders` provides built-in processors, so that it links to the upstream documentation about `itemloaders` built-in processors.

scrapy/loader/__init__.py
```diff
@@ -3,219 +3,104 @@
...
+    .. attribute:: selector
+
+        The :class:`~scrapy.selector.Selector` object to extract data from.
+        It's either the selector given in the ``__init__`` method or one created from
+        the response given in the ``__init__`` method using the
+        :attr:`default_selector_class`. This attribute is meant to be
+        read-only.
+    """
 
     default_item_class = Item
```
Gallaecio, line 82: Is this useful? I ask because I imagine that, given `Item` has no fields defined, this always needs to be overriden with a different value.

scrapy/loader/__init__.py
```diff
@@ -3,219 +3,104 @@
...
-
-    def get_css(self, css, *processors, **kw):
-        values = self._get_cssvalues(css, **kw)
-        return self.get_value(values, *processors, **kw)
-
-    def _get_cssvalues(self, csss, **kw):
-        self._check_selector_method()
-        csss = arg_to_iter(csss)
-        return flatten(self.selector.css(css).getall() for css in csss)
+            return self.item.fields[field_name].get(key)
```
Gallaecio, line 105: I guess we need to wait for #3881 and use `ItemAdapter` here. Maybe also in the `itemloaders` library itself?

kmike: That's the question, what should we merge first :)

scrapy/loader/__init__.py
ejulio: This is here only for backwards compatibility, as this is how it was before.
I guess there should be no harm in removing it

Gallaecio: Right, backward compatibility. Keeping it makes sense.

scrapy/loader/__init__.py
ejulio: Both work for me.
Just need to keep in mind that, as we wait, new changes may come to the files that were moved, requiring to patch/merge changes in the new repo..
This can get a bit complicated, as we merged the changes there (so we may need to go back in history to merge and then apply the changes back again)

setup.py
```diff
@@ -71,6 +71,7 @@ def has_environment_marker_platform_impl_support():
         'Twisted>=17.9.0',
         'cryptography>=2.0',
         'cssselect>=0.9.1',
+        'itemloaders>=1.0.0',
```
kmike, line 74: Could you please check that itemloader's setup.py uses the same minimal package requirements as Scrapy?

scrapy/loader/__init__.py
```diff
@@ -3,219 +3,104 @@
...
+class ItemLoader(itemloaders.ItemLoader):
     """
-    Allow to use single-argument functions as input or output processors
-    (no need to define an unused first 'self' argument)
-    """
-    with suppress(AttributeError):
-        if '.' not in method.__qualname__:
-            return method.__func__
-    return method
+    Return a new Item Loader for populating the given Item. If no item is
```
kmike, line 14: It seems this docstring explains how `ItemLoader.__init__` works, not what is ItemLoader class about. Could it be moved to `__init__` method (is autodocs output reasonable?)?

Alternatively, it can be rewritten as a class docstring. This makes more sense to me, because e.g. attribute docs don't really belong to `__init__` docstring.

setup.py
victor-torres: What do you mean by "the same minimal package requirements as Scrapy"?

kmike: If itemloaders require e.g. parsel 1.6, and scrapy requires 1.5, and now scrapy requires itemloaders, than we're implicitly bumping minimal supported parsel version in scrapy.

ejulio: I'll update the requirements there..
Though, `jmespath` was optional in scrapy, but is a requirement in `itemloaders`
It shouldn't be a problem, but it changed.

ejulio: https://github.com/scrapy/itemloaders/pull/10

scrapy/loader/__init__.py
```diff
@@ -3,219 +3,103 @@
...
-
-    def get_collected_values(self, field_name):
-        return self._values[field_name]
-
-    def get_input_processor(self, field_name):
-        proc = getattr(self, '%s_in' % field_name, None)
+        context.update(response=response)
+        super().__init__(item=item, selector=selector, parent=parent, **context)
+
+    def get_default_input_processor_for_field(self, field_name):
```
kmike, line 90: hm, I suspect these extension points won't be necessary once itemloader gets itemadapter support.

ejulio: That's true

Gallaecio: Itemadapter support is in.

noviluni:
Adding this link here to generate a reference of this in that issue: https://github.com/scrapy/scrapy/issues/1777#issuecomment-184272838

Gallaecio:
Regarding the failed `typing` job, we are not enforcing type annotations at the moment, so feel free to just add the affected files to the exception list in `setup.cfg`.

ejulio:
Thanks for the heads-up @Gallaecio .
Was trying to understand the reasons for the failure.

docs/topics/loaders.rst
```diff
@@ -20,6 +20,10 @@ Item Loaders are designed to provide a flexible, efficient and easy mechanism
 for extending and overriding different field parsing rules, either by spider,
 or by source format (HTML, XML, etc) without becoming a nightmare to maintain.
 
+.. note:: Item Loaders are an extension of the itemloaders_ library that make it
+    easier to work with Scrapy by adding support for :ref:`Items <topics-items>`
```
Gallaecio, line 24: :lipstick: 

```suggestion
    easier to work with Scrapy by adding support for :ref:`items <topics-items>`
```

docs/topics/loaders.rst
```diff
@@ -173,17 +177,17 @@ The other thing you need to keep in mind is that the values returned by input
 processors are collected internally (in lists) and then passed to output
 processors to populate the fields.
 
-Last, but not least, Scrapy comes with some :ref:`commonly used processors
-<topics-loaders-available-processors>` built-in for convenience.
+Last, but not least, ``itemloaders`` comes with some :ref:`commonly used
+processors <processors>` built-in for convenience.
```
Gallaecio, line 181:
```suggestion
Last, but not least, itemloaders_ comes with some :ref:`commonly used
processors <itemloaders:built-in-processors>` built-in for convenience.
```

Note: ``` :doc:`itemloaders <itemloaders:index>` ``` could work as an alternative to `itemloaders_` that links to the home page of the `itemloaders` documentation.

docs/topics/loaders.rst
```diff
@@ -213,8 +217,8 @@ this way. However, there is one more place where you can specify the input and
 output processors to use: in the :ref:`Item Field <topics-items-fields>`
 metadata. Here is an example::
 
+    from itemloaders.processors import Join, MapCompose, TakeFirst
     import scrapy
```
Gallaecio, line 221: :lipstick: 

```suggestion
    import scrapy
    from itemloaders.processors import Join, MapCompose, TakeFirst
```

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
-    """
-    Allow to use single-argument functions as input or output processors
-    (no need to define an unused first 'self' argument)
+class ItemLoader(itemloaders.ItemLoader):
     """
-    with suppress(AttributeError):
-        if '.' not in method.__qualname__:
-            return method.__func__
-    return method
+    A user-fiendly abstraction to populate data to an :ref:`Item <topics-items>`
```
Gallaecio, line 14:
```suggestion
    A user-friendly abstraction to populate an :ref:`item <topics-items>` with data
```

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
+    by applying :ref:`field processors <topics-loaders-processors>` to scraped data.
+    When instantiated with a ``selector`` or a ``response`` it supports
+    data extraction from web pages using :ref:`selectors <topics-selectors>`.
+
+    :param item: The item instance to populate using subsequent calls to
+        :meth:`~ItemLoader.add_xpath`, :meth:`~ItemLoader.add_css`,
+        or :meth:`~ItemLoader.add_value`.
+    :type item: :class:`~scrapy.item.Item` object
```
Gallaecio, line 22: I believe this syntax works as well:

```suggestion
    :type item: scrapy.item.Item
```

No strong opinion, though.

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
+
+    :param selector: The selector to extract data from, when using the
+        :meth:`add_xpath` (resp. :meth:`add_css`) or :meth:`replace_xpath`
```
Gallaecio, line 25: What does “resp.” mean here?

:lipstick: Maybe we could just list the 4 methods, as a plain list.

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
+    :type selector: :class:`~scrapy.selector.Selector` object
+
+    :param response: The response used to construct the selector using the
+        :attr:`default_selector_class`, unless the selector argument is given,
+        in which case this argument is ignored.
+    :type response: :class:`~scrapy.http.Response` object
+
+    If no item is given, one is instantiated automatically using the class in
+    :attr:`default_item_class`.
+    The item, selector, response and the remaining keyword arguments are
```
Gallaecio, line 36:
```suggestion
    :attr:`default_item_class`.

    The item, selector, response and remaining keyword arguments are
```

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
+    assigned to the Loader context (accessible through the :attr:`context` attribute).
+
+    .. attribute:: item
+
+        The item object being parsed by this Item Loader.
+        This is mostly used as a property so when attempting to override this
```
Gallaecio, line 42:
```suggestion
        This is mostly used as a property so, when attempting to override this
```

scrapy/loader/__init__.py
```diff
@@ -3,217 +3,86 @@
...
+
+    .. attribute:: context
+
+        The currently active :ref:`Context <loaders-context>` of this Item Loader.
+        Refer to <loaders-context> for more information about the Loader Context.
 
+    .. attribute:: default_item_class
 
-class ItemLoader:
+        An Item class (or factory), used to instantiate items when not given in
```
Gallaecio, line 52: If `dict` is a supported value, even if `attr.s` and `dataclass` are experimental, maybe we should say lowercase ‘item’ with a link to the topic on items.

docs/topics/loaders.rst
```diff
@@ -20,6 +20,10 @@ Item Loaders are designed to provide a flexible, efficient and easy mechanism
 for extending and overriding different field parsing rules, either by spider,
 or by source format (HTML, XML, etc) without becoming a nightmare to maintain.
 
+.. note:: Item Loaders are an extension of the itemloaders_ library that make it
+    easier to work with Scrapy by adding support for :ref:`items <topics-items>`
```
kmike, line 24: Aren't items supported in itemloaders out of box, via itemadapter?

ejulio: Yup, I'll leave only the reference to `response`

tests/test_loader.py
```diff
@@ -7,8 +6,7 @@
 from scrapy.http import HtmlResponse
 from scrapy.item import Item, Field
 from scrapy.loader import ItemLoader
-from scrapy.loader.processors import (Compose, Identity, Join,
-                                      MapCompose, SelectJmes, TakeFirst)
+from scrapy.loader.processors import Compose, Identity, MapCompose, TakeFirst
```
Gallaecio, line 9: In this and any other file were we import processors, we should start importing them from the new library.

tests/test_loader_deprecated.py
```diff
@@ -0,0 +1,695 @@
...
+These tests are kept as references from the ones that were ported to a itemloaders library.
+Once we remove the references from scrapy, we can remove these tests.
+"""
+
+from functools import partial
+import unittest
```
Gallaecio, line 7: :lipstick: 

```suggestion
import unittest
from functools import partial
```

ejulio:
@Gallaecio , I've updated the PR with your ideas.
Just bear in mind in that, `extract_regex` was already in `parsel.utils`, so I just added a simple test in `test_loader_deprecated` so we don't need a new file only for this case.

ejulio:
I can't believe this was finally merged :tada: :stuck_out_tongue: 
Thanks for all the effort here @victor-torres @Gallaecio @kmike :clap:

PR #4261, opened 2019-12-31
Test Python 3.5.0

Gallaecio:
Changes:

- Group quick tests together (move `docs` up)
- Separate with a line break Travis job definition groups: static, minimum Python, middle Python, maximum Python, PyPy
- Use Python 3.5.0 for the pinned-dependencies job
- Test asyncio both with 3.5.0 and the latest 3.5
- Simplify the asyncio Tox configuration, allowing to run the asyncio tests on any Python version
- Use the `py` automatic Tox environment when a more specific environment name is not necessary
- Use a `PYPI_RELEASE_JOB` environment variable to mark which job is used to deploy to PyPI

Note: `tests/test_proxy_connect.py` is skipped in Python 3.5.0. I tried to get the tests to work, but the introduction of some `typing` in mitproxy 3.0.0 required using an even earlier mitproxy for these tests. mitmproxy 2 had completely different command-line options, and while I did try the following, it caused timeouts and I preferred not to spend more time on this:

```
# …
                           '-c', script,
                           '--bind-address', '127.0.0.1',
                           '--port', '0',
                           '--htpasswd', '/scrapy/htpasswd',
                           '--cert', cert_path,
                           '--insecure',
# …
```

PR #4518, opened 2020-04-28
Run tests with Python 3.5.0

Gallaecio:
Related to #4261


elacuesta:
The patch looks simple and harmless, but it seems like the `3.5.0` version is not available, at least for Ubuntu 16.04 :disappointed: 

```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/16.04/x86_64/python-3.5.0.tar.bz2
$ curl -sSf --retry 5 -o python-3.5.0.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found
Unable to download 3.5.0 archive
```

kmike:
AFAIK Ubuntu 16.04 provides 3.5.1, so probably this should be the minimal version we support, not 3.5.0.

scrapy/__init__.py
```diff
@@ -14,8 +14,8 @@
 
 # Check minimum required Python version
 import sys
-if sys.version_info < (3, 5):
-    print("Scrapy %s requires Python 3.5" % __version__)
+if sys.version_info < (3, 5, 1):
```
kmike, line 17: Hm, what do you think about allowing it to run in 3.5.0? Most likely, it'd work, so this change may break Scrapy for some users.

Gallaecio:
I’m confused. Xenial should be running 3.5.1 according to [packages.ubuntu.com](https://packages.ubuntu.com/search?suite=xenial&arch=any&searchon=names&keywords=python3), but Travis’ Xenial uses 3.5.2.

setup.py
```diff
@@ -66,7 +66,7 @@ def has_environment_marker_platform_impl_support():
         'Topic :: Software Development :: Libraries :: Application Frameworks',
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
-    python_requires='>=3.5',
+    python_requires='>=3.5.1',
```
kmike, line 69: this would prevent installing Scrapy on 3.5.0 - do we want it?

kmike:
Thanks @Gallaecio!

PR #4524, opened 2020-04-28
Minimal centralized request fingerprints

### LINKED ISSUE #900: Centralized Request fingerprints

### LINKED ISSUE #4762: request fingerprint should separate different sections when building a hash
It looks like body & headers & values are not separated where fp is being built in request_fingerprint function. 

So e.g. a request with header with "foo" and "bar" values would get the same fingerprint as a request with "fo" and "obar" values. The same applies for body, etc. URL is joined with the body, for example, so a POST request to http://example.com/foo with empty body gets the same fingerprint as a POST request to http://example.com/fo with "o" body:

```py
from scrapy import request
from scrapy.utils.request import request_fingerprint

req1 = scrapy.Request("http://example.com/foo", method="POST", body=b'')
req2 = scrapy.Request("http://example.com/fo", method="POST", body=b'o')

request_fingerprint(req1)  # 'ca64371a30748ddb06e28365b7e1177db7bfa8e9'
request_fingerprint(req2)  # 'ca64371a30748ddb06e28365b7e1177db7bfa8e9'
```

Gallaecio:
Fork of #4113, with minimal code changes, based on @nyov’s feedback.

Fixes #900, fixes #3420, closes #4113, fixes #4762.

To do _after_ merging:

- [ ] Open an issue about https://github.com/scrapy/scrapy/pull/4524#discussion_r477515464

PS: Please, do not merge without squashing.


kmike:
@Gallaecio ❤️ the design, and can see how this can be extended in Scrapy itself with something more elaborate if we decide to do so.

The only thing I'd change is making REQUEST_FINGERPRINTER a Scrapy extension instead of a callable - i.e. load it using create_instance. This would allow fingerprinter to access Scrapy settings, which would allow to implement e.g. a fingerprinter which reads SESSION_IDENTIFIERS option and removes them from the request GET arguments before taking a fingerprint.

Gallaecio:
I like the idea, but I’m not convinced about https://github.com/scrapy/scrapy/pull/4524/commits/009296232c3e0b7c3ea51e928ca335d022fd52e8.

It may be cleaner if we only support `from_settings`, but not `from_crawler`. I’m not sure if that is what you had in mind.

Also, being able to define `from_crawler` in a class while allowing subclasses to use `from_settings` seems rather inelegant. I don’t see how to avoid creating an instance of `RequestFingerprinter` without passing `crawler` to it.

scrapy/dupefilters.py
```diff
@@ -38,10 +39,32 @@ def __init__(self, path=None, debug=False):
...
+
     @classmethod
     def from_settings(cls, settings):
         debug = settings.getbool('DUPEFILTER_DEBUG')
-        return cls(job_dir(settings), debug)
+        fingerprinter = settings.getinstance(
+            'REQUEST_FINGERPRINTER',
+            singleton=True,
+        )
+        return cls(job_dir(settings), debug, fingerprinter=fingerprinter)
```
kmike, line 67: Is from_settings method required just for backwards compatibility, or is there another reason code is duplicated here?

scrapy/settings/__init__.py
```diff
@@ -233,6 +235,47 @@ def maxpriority(self):
         else:
             return get_settings_priority('default')
 
+    def getinstance(self, name, *, crawler=None, singleton=False):
```
kmike, line 238: Aha, I see the problem now: you can't just create RequestFingerprinter in a dupefilter, because other components also may need to access it.

I'm not sure though that settings is a best place to handle this singleton. 

1) All other settings return simple values at the moment - str, list, dict.
2) You need to pass crawler here, but only once - on a first call. On all other calls crawler is not necessary. But we need to ensure that the first call was a call which passed a crawler.

I haven't checked it in detail, but what do you think about making fingerprinter a more "fundamental" thing, maybe a Crawler attribute, similar to logformatter or stats? In this case logic around singletons won't be needed.

kmike: This may require an access to Crawler in some places where it was not possible before though; it could make the scope of this PR larger.

scrapy/dupefilters.py
Gallaecio: Backward compatibility (subclasses calling `super().from_settings()`?). I must say that it is not that clear to me how inheritance and these class methods should be handled.

If you are OK with it, I’ll just drop the class method. Otherwise, things get really messy.

scrapy/crawler.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, spidercls, settings=None):
         self.spidercls.update_settings(self.settings)
 
         self.signals = SignalManager(self)
+
+        self.request_fingerprinter = create_instance(
```
kmike, line 54: Any reason for putting it before creation of logging and stats component? I wonder if there is a use case for having access to them in fingerprinter's from_crawler / init methods.

Gallaecio: No, I simply wanted to make sure signals were available. I’ll move it further down.

docs/topics/request-response.rst
```diff
@@ -312,6 +312,143 @@ errors if needed::
...
+Default: ``'scrapy.utils.request.RequestFingerprinter'``
+
+The import path of a :ref:`request fingerprinter
+<custom-request-fingerprinter>`.
+
+The default request fingerprinter takes into account a canonical version
+(:func:`w3lib.url.canonicalize_url`) of
+:attr:`request.url <scrapy.http.Request.url>` and the values of
+:attr:`request.method <scrapy.http.Request.method>` and
+:attr:`request.body <scrapy.http.Request.body>`. It then generates an
```
elacuesta, line 353:
```suggestion
:attr:`request.body <scrapy.http.Request.body>`. It then generates a
```
Nitpick

docs/topics/request-response.rst
```diff
@@ -312,6 +312,143 @@ errors if needed::
...
+.. classmethod:: from_settings(cls, settings)
+
+   If present, and ``from_crawler`` is not defined, this class method is called
+   to create a request fingerprinter instance from a
+   :class:`~scrapy.settings.Settings` object. It must return a new instance of
+   the request fingerprinter.
+
+For some common use cases you can use
+:func:`~scrapy.utils.request.request_fingerprint` in your ``fingerprint``
+method implementation:
```
elacuesta, line 398: Should we mention that `scrapy.utils.request.request_fingerprint` is used in the default implementation?

docs/topics/request-response.rst
```diff
@@ -312,6 +312,143 @@ errors if needed::
...
+    class RequestFingerprinter:
+
+        def fingerprint(self, request):
+            return request_fingerprint(request, include_headers=['X-ID'])
+
+You can also write your own fingerprinting logic from scratch. However, if you
+do not use :func:`~scrapy.utils.request.request_fingerprint`, make sure you use
+:class:`~weakref.WeakKeyDictionary` to cache request fingerprints. For example,
+to take into account only the URL of a request, without any prior URL
+canonicalization::
```
elacuesta, line 420: I think it'd be good to add a line explaining the reason to use `weakref.WeakKeyDictionary`, as it might not seem obvious.

scrapy/settings/__init__.py
```diff
@@ -83,6 +83,7 @@ class BaseSettings(MutableMapping):
     def __init__(self, values=None, priority='project'):
         self.frozen = False
         self.attributes = {}
+        self._singletons = {}
```
elacuesta, line 86: Is this a leftover from a previous implementation? There seems to be no other occurrence in this diff.

docs/topics/request-response.rst
Gallaecio: Actually, I think “an” is correct here.

The use of “a” or “an” [is based on phonetics](https://writingexplained.org/a-vs-an-difference) rather than letters. Assuming that “SHA1” is pronounced by reading each letter and number separately, and because “s” as a standalone letter is pronounced [“ess”](https://en.wikipedia.org/wiki/S#cite_ref-2), I believe it should be preceded by “an”.

scrapy/settings/__init__.py
Gallaecio: Good catch!

docs/topics/request-response.rst
elacuesta: Interesting :thinking: It still looks a bit weird, but I'm no native English speaker and I have no grounds to say otherwise. Thanks for the links!

docs/topics/api.rst
```diff
@@ -34,6 +34,13 @@ how you :ref:`configure the downloader middlewares
     :class:`scrapy.spiders.Spider` subclass and a
     :class:`scrapy.settings.Settings` object.
 
+    .. attribute:: request_fingerprinter
+
+        The generator of request fingerprints of this crawler.
```
elacuesta, line 39: (Nitpick) Should we choose a different word here? I'm concerned it might be read as if we're talking about [_generators_](https://wiki.python.org/moin/Generators).

docs/topics/request-response.rst
```diff
@@ -312,6 +312,157 @@ errors if needed::
...
+
+   If present, this class method is called to create a request fingerprinter
+   instance from a :class:`~scrapy.crawler.Crawler` object. It must return a
+   new instance of the request fingerprinter.
+
+   *crawler* provides access to all Scrapy core components like settings and
+   signals; it is a way for the request fingerprinter to access them and hook
+   its functionality into Scrapy.
+
+   :param crawler: crawler that uses this pipeline
```
wRAR, line 386: "pipeline" looks like a copy paste mistake

docs/topics/request-response.rst
```diff
@@ -312,6 +312,157 @@ errors if needed::
...
+
+   Return a :class:`str` that uniquely identifies *request*.
+
+   The returned :class:`str` must be a valid filename. The hexadecimal
+   representation of a hash is recommended.
+
+   :param request: request to fingerprint
+   :type request: scrapy.http.Request
+
+Additionally, they may also implement the following methods:
```
wRAR, line 374: "it" instead of "they"?

docs/topics/request-response.rst
```diff
@@ -341,6 +342,157 @@ achieve this by using ``Failure.request.cb_kwargs``::
...
+.. _custom-request-fingerprinter:
+
+Writing your own request fingerprinter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A request fingerprinter is a class that must implement the following method:
+
+.. method:: fingerprint(self, request)
+
+   Return a :class:`str` that uniquely identifies *request*.
```
kmike, line 396: I wonder if that's our last chance to solve https://github.com/scrapy/scrapy/pull/3420 in a backwards compatible way, or at least progress towards it. Should we require fingerprinter.fingerprint return bytes?

scrapy/dupefilters.py
```diff
@@ -59,10 +58,10 @@ def request_seen(self, request):
             return True
         self.fingerprints.add(fp)
         if self.file:
-            self.file.write(fp.hex() + '\n')
+            self.file.write(fp + '\n')
 
     def request_fingerprint(self, request):
-        return self.fingerprinter.fingerprint(request)
+        return self.fingerprinter.fingerprint(request).hex()
```
kmike, line 64: That's unfortunate we won't be able to get efficiency gains from switching to bytes (less memory usage). So we're making a step in a right direction, but not actually fixing the issue. 

We would still need to make a backwards-incompatible change to RFPDupeFilter (or figure out a workaround) to actually make it. I think that's fine to handle this separately, if we're sure the current code changes allow us to do that easier. What do you think @Gallaecio?

Gallaecio: I agree with changing it separately.

Given the duplicate filter class API (which made sense before centralized fingerprings), the only backward-compatible way forward I can think of would be to deprecate the class in favor of a new one that uses binary fingerprints.

scrapy/utils/request.py
```diff
@@ -68,6 +70,67 @@ def request_fingerprint(request, include_headers=None, keep_fragments=False):
     return cache[cache_key]
 
 
+_fingerprint_cache = weakref.WeakKeyDictionary()
+
+
+def fingerprint(request, include_headers=None, keep_fragments=False) -> bytes:
```
kmike, line 76: A chance for a clean-up :)

```suggestion
def fingerprint(request, *, include_headers=None, keep_fragments=False) -> bytes:
```

scrapy/utils/request.py
```diff
@@ -68,6 +70,67 @@ def request_fingerprint(request, include_headers=None, keep_fragments=False):
...
+        fp = hashlib.sha1()
+        fp.update(to_bytes(request.method))
+        fp.update(to_bytes(canonicalize_url(request.url, keep_fragments=keep_fragments)))
+        fp.update(request.body or b'')
+        if include_headers:
+            for hdr in include_headers:
+                if hdr in request.headers:
+                    fp.update(hdr)
+                    for v in request.headers.getlist(hdr):
+                        fp.update(v)
```
kmike, line 121: Sorry, I'm getting more an more unrelated comments on this PR :) This is a nitpick, but it looks like body & headers & values are not separated where fp is being built. So e.g. a request with header with "foo" and "bar" values would get the same fingerprint as a request with "fo" and "obar" values. The same applies for body, etc. URL is joined with the body, for example, so a POST request to http://example.com with empty body may get the same fingerprint as a POST request to http://example.co with "m" body.

Maybe it was not invisible in past, because such cases are rare, and because include_headers was a bit tricky to use. Now we're documenting how to use it, so the issue may become a bit more important. 

Changing the fingerprint for a request should be considered a backwards incompatible change - e.g. disk cache will be effectively emptied after Scrapy upgrade. Here we're introducing a new function, so there can be an argument that we should make the change now, to not change the new function in future. But I think this can be done separately - a real breaking change is not the output of fingerprint function per se, but fingerprints requests get by default, so we can't avoid making this non-breaking. Opened a placeholder https://github.com/scrapy/scrapy/issues/4762 for that.

tests/test_utils_request.py
```diff
@@ -35,62 +70,51 @@ def test_request_fingerprint(self):
...
-        r1 = Request("http://www.example.com/some/page.html?arg=1")
-        self.assertEqual(request_httprepr(r1), b'GET /some/page.html?arg=1 HTTP/1.1\r\nHost: www.example.com\r\n\r\n')
 
-        r1 = Request("http://www.example.com", method='POST',
-                     headers={"Content-type": b"text/html"}, body=b"Some body")
-        self.assertEqual(
-            request_httprepr(r1),
-            b'POST / HTTP/1.1\r\nHost: www.example.com\r\nContent-Type: text/html\r\n\r\nSome body'
-        )
+class RequestFingerprintTest(FingerprintTest):
```
kmike, line 107: Would it be possibile to test that the fingerprint returned by the new function is exactly the same as a fingerprint which was returned before? Non-matching fingerprints may affect media uploads, cache, etc.; that'd be a backwards incompatible change.

tests/test_dupefilters.py
```diff
@@ -120,14 +120,14 @@ def test_request_fingerprint(self):
 
         dupefilter.close('finished')
 
-        class CaseInsensitiveRFPDupeFilter(RFPDupeFilter):
+        class RequestFingerprinter:
```
kmike, line 123: How hard would it be to check that from_crawler or from_settings methods are really called for RequestFingerprinters?

kmike (approved):
Great job @Gallaecio! I've left a couple of comments about tests, but this looks very good overall; feel free to merge.

scrapy/utils/request.py
Gallaecio: The main problem is that the new function is used by the (new) default request fingerprinter.

To avoid users using this function for their custom request fingerprinters, we could:
#. Create a private copy of the function, for the default fingerprinter to use to keep backward compatibility.
#. Fix the function.

That’s something that I believe would make sense to do as part of this pull request. Thoughts?

Then there’s the issue of moving users from the bugged, default fingerprinter logic, to one that is not bugged. I think that whether we make that change as part of this pull request or we handle it as part of #4762 depends on how we want to approach it:

- We could modify this pull request to introduce 2 different fingerprinter classes, one with the deprecated behavior (e.g. `DeprecatedRequestFingerprinter`, `Unsafe…`, `Bogus…`, `Old…`, `Pre2Dot4…`), and one with the fixed behavior (i.e. `RequestFingerprinter`), and use the deprecated class by default, and have it log a warning asking users to switch to the new one. For projects generated with `startproject`, we could set that automatically.
- We could do something similar, but using a setting which the default fingerprinter reads. In that case, we _can_ handle it separately as a fix for #4762, although I would not mind fixing it here.

tests/test_utils_request.py
Gallaecio: The output is the same at the moment, however if we decide to fix the fingerprint generation in the new function it will not be. Assuming we will fix the fingerprint generation in the new function, what about extending the deprecation message to indicate that the new function not only returns bytes, but also a different fingerprint?

Gallaecio:
The last commit fixes #4762 by serializing request data with `json.dumps` before hashing.

I wasnted to use `pickle` for performance reasons, but after @nyov asked about that in https://github.com/scrapy/scrapy/pull/4113#discussion_r391308583 I looked into it and it might not be valid for this purpose.

As for `json.dumps`, I’ve seen [recommendations of using `sort_keys`](https://stackoverflow.com/a/22003440/939364) when used for hashing. However, I don’t think we need them in our use case, given Python dictionary key order is deterministic.

I tested the performance change with the following code (not sure it’s the right way to test it, though):

```python
import timeit, random, warnings
from scrapy import Request
from scrapy.utils.request import request_fingerprint, fingerprint

def test_function(number, function, include_headers, keep_fragments):
    for n in range(number):
        fp = function(
            Request(f'https://www.example.com/path{n}#{n}', headers={'foo': b'bar'}),
            include_headers=include_headers,
            keep_fragments=keep_fragments,
        )

number = 200
for include_headers in (None, ['foo']):
    for keep_fragments in (False, True):
        for function in (request_fingerprint, fingerprint):
            with warnings.catch_warnings():
                warnings.simplefilter("ignore")
                time = timeit.timeit(
                    "test_function(number, function, include_headers, keep_fragments)",
                    setup="from __main__ import number, test_function, function, include_headers, keep_fragments",
                    number=number,
                )
                print(f'{function.__name__}(…, include_headers={include_headers!r}, keep_fragments={keep_fragments!r}): {time}')
```

My results were:

```
request_fingerprint(…, include_headers=None, keep_fragments=False): 2.2705133950003074
fingerprint(…, include_headers=None, keep_fragments=False): 2.3899707270011277
request_fingerprint(…, include_headers=None, keep_fragments=True): 2.3113510660004977
fingerprint(…, include_headers=None, keep_fragments=True): 2.4022065899989684
request_fingerprint(…, include_headers=['foo'], keep_fragments=False): 2.4886272530002316
fingerprint(…, include_headers=['foo'], keep_fragments=False): 2.6481095350009127
request_fingerprint(…, include_headers=['foo'], keep_fragments=True): 2.5297311330014054
fingerprint(…, include_headers=['foo'], keep_fragments=True): 2.6578162269997847
```

wRAR:
Lots of `twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.` in the tests.

Gallaecio:
The problem is my use of pytest-specific test functions instead of `unittest.Testcase`. I did that in order to use `@pytest.mark.parametrize`, which cannot be used with unittest. I think I’ve had this issue before, of trying to define tests as functions and breaking reactor-using tests as a result.

I’ve tried to move the functions as methods into a non-unittest class, but the issue remains.

I guess I’ll just have to use `for` loops instead of `parametrize`.

scrapy/dupefilters.py
```diff
@@ -7,10 +7,6 @@
 
 class BaseDupeFilter:
 
-    @classmethod
-    def from_settings(cls, settings):
-        return cls()
```
Gallaecio, line 12: Come to think of it, this is technically backward incompatible, as users may be calling `super().from_settings()` in their subclasses.

docs/topics/request-response.rst
```diff
@@ -341,6 +342,273 @@ achieve this by using ``Failure.request.cb_kwargs``::
...
+    class RequestFingerprinter:
+
+        cache = WeakKeyDictionary()
+
+        def fingerprint(self, request):
+            if request not in self.cache:
+                fp = sha1()
+                fp.update(to_bytes(request.method))
+                fp.update(to_bytes(canonicalize_url(request.url)))
+                fp.update(request.body or b'')
```
kmike, line 564: here in the docs we have the same bug as the one being fixed in the implementation

Gallaecio: On purpose, see the paragraph introducing this code example :slightly_smiling_face:

Gallaecio:
I [wrote a new benchmark](https://gist.github.com/Gallaecio/0c5ea31ea6bad218fb8c6be0d993e31c) to test alternative JSON libraries, as well as including the performance of `_request_fingerprint_as_bytes` (used by code using the new centralized fingerprints API but the 2.3 fingerprinting implementation).

[The results](https://gist.github.com/Gallaecio/99071931c5c9de313571778f3590bd62) seem to indicate that `orjson` would work best for our use case.

But the results vary greatly from one run to another in my system, so I’m not sure adding a new dependency is worth the performance boost, specially since the new system gives freedom to implement fingerprinters focused on performance only if desired.

Gallaecio:
Note: release procedure updated to make sure that last commit gets handled properly before release: https://github.com/scrapy/scrapy/wiki/Scrapy-release-procedure#release-notes

docs/topics/request-response.rst
```diff
@@ -353,6 +354,273 @@ achieve this by using ``Failure.request.cb_kwargs``::
...
+
+-   ``'PREVIOUS_VERSION'`` (default)
+
+    This implementation uses the same request fingerprinting algorithm as
+    Scrapy PREVIOUS_VERSION and earlier versions.
+
+    Even though this is the default value for backward compatibility reasons,
+    it is a deprecated value.
+
+-   ``'VERSION'``
```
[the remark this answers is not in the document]
kmike: The same applies for "PREVIOUS_VERSION"; sorry if the question is stupid, and answered in some previous comment :)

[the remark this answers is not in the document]
Gallaecio: They are both meant to be replaced, see https://github.com/scrapy/scrapy/pull/4524#issuecomment-729240112

kmike:
hey @Gallaecio! I re-checked the PR, and it looks great. Are you ok with merging it after resolving merge conflicts?

Gallaecio:
I count on tests passing now. @wRAR, @kmike, please check the [latest merge to resolve conflicts](https://github.com/scrapy/scrapy/pull/4524/commits/85022e8f4da26034f3bc6d0b95900a111660b545), as it includes some non-trivial changes.

scrapy/utils/request.py
```diff
@@ -57,7 +59,7 @@ def request_fingerprint(
     the fingerprint.
 
     For this reason, request headers are ignored by default when calculating
-    the fingerprint. If you want to include specific headers use the
+    the fingeprint. If you want to include specific headers use the
```
kmike, line 62: a typo?

scrapy/dupefilters.py
```diff
@@ -51,9 +58,26 @@ def __init__(self, path: Optional[str] = None, debug: bool = False) -> None:
...
-    def from_settings(cls: Type[RFPDupeFilterTV], settings: BaseSettings) -> RFPDupeFilterTV:
+    def from_settings(cls: Type[RFPDupeFilterTV], settings: BaseSettings, *, fingerprinter=None) -> RFPDupeFilterTV:
         debug = settings.getbool('DUPEFILTER_DEBUG')
-        return cls(job_dir(settings), debug)
+        try:
+            return cls(job_dir(settings), debug, fingerprinter=fingerprinter)
+        except TypeError:
+            result = cls(job_dir(settings), debug)
+            result.fingerprinter = fingerprinter
+            return result
```
kmike, line 68: Is the idea to support RFPDupeFilter subclasss which don't have fingerprint argument in `__init__` method? If so, would it make sense to show some kind of deprecation warning, and ask users to update the code?

scrapy/dupefilters.py
```diff
@@ -51,9 +58,26 @@ def __init__(self, path: Optional[str] = None, debug: bool = False) -> None:
...
+    def from_crawler(cls, crawler):
+        try:
+            return cls.from_settings(
+                crawler.settings,
+                fingerprinter=crawler.request_fingerprinter,
+            )
+        except TypeError:
+            result = cls.from_settings(crawler.settings)
+            result.fingerprinter = crawler.request_fingerprinter
+            return result
```
kmike, line 80: The same question - would it make sense to show a warning if "fingerprint" argument is not supported?

tests/test_dupefilters.py
```diff
@@ -64,7 +64,9 @@ def test_df_direct_scheduler(self):
         self.assertEqual(scheduler.df.method, 'n/a')
 
     def test_filter(self):
-        dupefilter = RFPDupeFilter()
+        crawler = get_crawler()
+        scheduler = Scheduler.from_crawler(crawler)
+        dupefilter = scheduler.df
```
kmike, line 69: Would it make sense to extract an utility function? Something like

```py
def _get_crawler_dupefilter(crawler):
    scheduler = Scheduler.from_crawler(crawler)
    return scheduler.df    
```

tests/test_utils_request.py
```diff
@@ -1,96 +1,679 @@
...
+            for request, _, kwargs in self.known_hashes
+        ]
+        expected = [
+            bytes.fromhex(_fingerprint)
+            for _, _fingerprint, _ in self.known_hashes
+        ]
+        self.assertEqual(actual, expected)
+
+
+_fingerprint_cache_2_3: Mapping[Request, Tuple[None, bool]] = WeakKeyDictionary()
```
kmike, line 368: should this and similar objects be renamed to `..._2_6`?

kmike (approved):
👍

PR #4525, opened 2020-04-29
Flake8: remove W504 code

elacuesta:


scrapy/spidermiddlewares/referer.py
```diff
@@ -163,9 +163,10 @@ class StrictOriginPolicy(ReferrerPolicy):
     name = POLICY_STRICT_ORIGIN
 
     def referrer(self, response_url, request_url):
-        if ((self.tls_protected(response_url) and
-             self.potentially_trustworthy(request_url))
-            or not self.tls_protected(response_url)):
+        if (
+            self.tls_protected(response_url) and self.potentially_trustworthy(request_url)
+            or not self.tls_protected(response_url)
```
kmike, line 168: Is it the same condition? Tbh I found a previous one, with (), easier to understand.

scrapy/contracts/__init__.py
```diff
@@ -17,10 +17,10 @@ def __init__(self, contracts):
             self.contracts[contract.name] = contract
 
     def tested_methods_from_spidercls(self, spidercls):
+        method_re = re.compile(r"^\s*@", re.MULTILINE)
```
kmike, line 20: a trick which works in 50% cases re.compile is used:
```suggestion
        is_method = re.compile(r"^\s*@", re.MULTILINE).search
```

scrapy/spidermiddlewares/referer.py
elacuesta: It is, because of the [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence). It seems like the `black` developers feel as you do:

```python
def referrer(self, response_url, request_url):
    if (
        self.tls_protected(response_url) and self.potentially_trustworthy(request_url)
    ) or not self.tls_protected(response_url):
        return self.origin_referrer(response_url)
```

I guess it's a matter of personal preference. I'd argue that the parenthesis are not strictly necessary, the same way they are redundant in `(2 * 3) + 4^5`. In that sense, I'd prefer to avoid them, but the truth is that I don't think they hurt legibility either.

scrapy/contracts/__init__.py
elacuesta: Nice, thanks!

scrapy/spidermiddlewares/referer.py
Gallaecio: I’m OK with the changes given the and-joined statements remain in the same line, but given the line length I guess it may indeed be more readable to keep the and-joined statements in separate lines, within parenthesis and an additional level of indentation.

elacuesta: Is line length an issue? The proposed change goes up to 90, and we have an approved PR to increase it to 119.

PR #4527, opened 2020-04-30
Small suggestion to make this a bit more clear.

plonko:
I found the previous a bit clunky, it seemed to change tense, but I think it just meant as I have suggested.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -124,9 +124,9 @@ and defines some attributes and methods:
   is an instance of :class:`~scrapy.http.TextResponse` that holds
   the page content and has further helpful methods to handle it.
 
-  The :meth:`~scrapy.spiders.Spider.parse` method usually parses the response, extracting
-  the scraped data as dicts and also finding new URLs to
-  follow and creating new requests (:class:`~scrapy.http.Request`) from them.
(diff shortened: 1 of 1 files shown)


docs/intro/tutorial.rst
```diff
@@ -124,9 +124,9 @@ and defines some attributes and methods:
   is an instance of :class:`~scrapy.http.TextResponse` that holds
   the page content and has further helpful methods to handle it.
 
-  The :meth:`~scrapy.spiders.Spider.parse` method usually parses the response, extracting
-  the scraped data as dicts and also finding new URLs to
-  follow and creating new requests (:class:`~scrapy.http.Request`) from them.
+  The :meth:`~scrapy.spiders.Spider.parse` method usually parses the response: it extracts
+  the scraped data as dicts, finds new URLs to
+  follow and creates new requests (:class:`~scrapy.http.Request`) from them.
```
Gallaecio, line 129: I see as an improvement in general, although now it reads as a 3-item list instead of a 2-item list as before, so I would suggest adding an extra ‘and’ after the comma.

```suggestion
  The :meth:`~scrapy.spiders.Spider.parse` method usually parses the response: it extracts
  the scraped data as dicts, and finds new URLs to
  follow and creates new requests (:class:`~scrapy.http.Request`) from them.
```

PR #4532, opened 2020-05-02
Ensure create_instance does not return None

### LINKED ISSUE #4528: Fail or warn if from_crawler() returns None
## Summary

Generate a warning or error if from_crawler() for a middleware/extension/etc. returns None

## Motivation

I created a custom extension and connected signals in the from_crawler() classmethod, but neglected to return the new extension instance.  Scrapy still reported the extension under "Enabled extensions", but none of the signals worked, since the instance was immediately garbage collected and its signals were silently disconnected.

This was of course an error on my part, but it would have saved me a lot of debugging if I had gotten a warning that from_crawler() was returning None, or if the extension were removed from the "Enabled extensions" list.

Would it be appropriate for utils.misc.create_instance() to raise an error or generate a warning if it's about to return None?  Or should MiddlewareManager treat create_instance() returning None the same as create_instance() raising NotConfigured?

willbeaufoy:
Currently create_instance() can return None (or any value) if an extension is
incorrectly implemented, but the extension will still show up as
enabled in the logs. This can cause confusion, as in the linked bug.

This change prevents this occurring by throwing an error if
create_instance() will not return an instance of the correct class.

(edit) fixes #4528

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -137,17 +137,26 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
     ``*args`` and ``**kwargs`` are forwarded to the constructors.
 
     Raises ``ValueError`` if both ``settings`` and ``crawler`` are ``None``.
+
+    Raises ``TypeError`` if the resulting instance is ``None`` (e.g. if an
+    extension has not been implemented correctly).
     """
     if settings is None:
         if crawler is None:
             raise ValueError("Specify at least one of settings and crawler.")
         settings = crawler.settings
     if crawler and hasattr(objcls, 'from_crawler'):
-        return objcls.from_crawler(crawler, *args, **kwargs)
+        instance = objcls.from_crawler(crawler, *args, **kwargs)
+        method_name = 'from_crawler'
     elif hasattr(objcls, 'from_settings'):
-        return objcls.from_settings(settings, *args, **kwargs)
+        instance = objcls.from_settings(settings, *args, **kwargs)
+        method_name = 'from_settings'
     else:
-        return objcls(*args, **kwargs)
+        instance = objcls(*args, **kwargs)
+        method_name = '__new__'
+    if instance is None:
+        raise TypeError("%s.%s returned None" % (objcls.__qualname__, method_name))
+    return instance
 
 
 @contextmanager
--- a/tests/test_utils_misc/__init__.py
+++ b/tests/test_utils_misc/__init__.py
@@ -114,8 +114,12 @@ class UtilsMiscTestCase(unittest.TestCase):
         #   2. with from_settings() constructor
         #   3. with from_crawler() constructor
         #   4. with from_settings() and from_crawler() constructor
-        spec_sets = ([], ['from_settings'], ['from_crawler'],
-                     ['from_settings', 'from_crawler'])
+        spec_sets = (
+            ['__qualname__'],
+            ['__qualname__', 'from_settings'],
+            ['__qualname__', 'from_crawler'],
+            ['__qualname__', 'from_settings', 'from_crawler'],
+        )
         for specs in spec_sets:
             m = mock.MagicMock(spec_set=specs)
             _test_with_settings(m, settings)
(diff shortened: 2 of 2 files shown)


scrapy/utils/misc.py
```diff
@@ -137,17 +137,24 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
...
     Raises ``ValueError`` if both ``settings`` and ``crawler`` are ``None``.
+
+    Raises ``TypeError`` if the resulting instance is ``None`` (e.g. if an
+    extension has not been implemented correctly).
     """
     if settings is None:
         if crawler is None:
             raise ValueError("Specify at least one of settings and crawler.")
         settings = crawler.settings
+    instance = None
```
elacuesta, line 148: This is not necessary, the below `if` statement covers all scenarios

scrapy/utils/misc.py
```diff
@@ -137,17 +137,24 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
...
-        return objcls.from_crawler(crawler, *args, **kwargs)
+        instance = objcls.from_crawler(crawler, *args, **kwargs)
     elif hasattr(objcls, 'from_settings'):
-        return objcls.from_settings(settings, *args, **kwargs)
+        instance = objcls.from_settings(settings, *args, **kwargs)
     else:
-        return objcls(*args, **kwargs)
+        instance = objcls(*args, **kwargs)
+    if instance is None:
+        raise TypeError("Instance is None for %s" % objcls)
```
elacuesta, line 156:
```suggestion
    if not isinstance(instance, objcls):
        raise TypeError("%r is not an instance of %r" % (instance, objcls))
```
I wonder if it wouldn't be better to check if `instance` is actually an instance of the class we're working with.

willbeaufoy: Oh yeah, thanks, will change. I'd forgotten python isn't block scoped.

scrapy/utils/misc.py
willbeaufoy: Oh yeah, thanks, will change. I'd forgotten python isn't block scoped.

scrapy/utils/misc.py
willbeaufoy: I could do this but it makes the unit tests a bit awkward, because in them we are using a MagicMock object as the objcls argument, rather than a class, so the isinstance() check fails.

I tried a few things e.g. the below (doesn't work) but haven't found a way around it yet so just thought I would check you think it's worth it first.

```python
spec_sets = (['__class__'], ['from_settings', '__class__'], ['from_crawler', '__class__'],
             ['from_settings', 'from_crawler', '__class__'])
for specs in spec_sets:
    m = mock.MagicMock(spec_set=specs)
    m.__class__ = mock.MagicMock
```

elacuesta: Hm it does seem tricky. Here are my two cents though:
* I'd prefer to adapt the tests to the implementation, not the other way around.
* I still think it'd be good to check for the class. Consider a class with a `from_crawler` class method that returns `3`. 3 is not None, but it's also probably not what Scrapy expects, I think it would be good to fail early in that case.

Sadly, I'm not a mock expert myself, so I don't have a solution at the moment. But I don't see any reason why the tests should remain how they are, we could write actual classes with the corresponding methods. I know it'd be more verbose than with mocks, but as I said, tests should adapt to the implementation IMHO.

willbeaufoy: Great, thanks for your suggestion.

willbeaufoy:
Should be ready now.

Gallaecio:
I wonder if there may be rare cases where `isinstance` may be a problem (e.g a factory class?).

willbeaufoy:
> I wonder if there may be rare cases where `isinstance` may be a problem (e.g a factory class?).
> 
> Edit: On the other hand, if they are rare enough, and given https://docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks exists (found out this week), the change may be worth it.

Yeah I thought a bit about this, but as this is my first contribution to this project I thought I didn't really know whether this rare case would actually come up. Let me know what you guys decide, if we want to allow for this rare case then I will just revert the last three commits so it just checks instance is not None again.

elacuesta:
@willbeaufoy We discussed this internally and realized that introducing the `isinstance` check would be a backward-incompatible change, which would need more extensive docs and tests. The initial check for `None` should be enough for now, I'm sorry for the trouble.

willbeaufoy:
> @willbeaufoy We discussed this internally and realized that introducing the `isinstance` check would be a backward-incompatible change, which would need more extensive docs and tests. The initial check for `None` should be enough for now, I'm sorry for the trouble.

No prob! Have reverted to my original commit so it should be ready now.

scrapy/utils/misc.py
```diff
@@ -137,17 +137,23 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
...
```
Gallaecio, line 155: I think it may be useful to indicate which class method (`from_crawler` or `from_settings`) caused the issue. You could do something like:

```py
raise TypeError("%s.%s returned None" % (objcls.__qualname__, method_name))
```

willbeaufoy: Done

PR #4533, opened 2020-05-03
Don't mention unsupported package versions in docs

kmike:
In setup.py more recent versions are required, so there is no need to warn users about require Twisted and pyopenssl versions in docs.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -421,7 +421,7 @@ connections (for ``HTTP10DownloadHandler``).
 
     HTTP/1.0 is rarely used nowadays so you can safely ignore this setting,
     unless you really want to use HTTP/1.0 and override
-    :setting:`DOWNLOAD_HANDLERS_BASE` for ``http(s)`` scheme accordingly,
+    :setting:`DOWNLOAD_HANDLERS` for ``http(s)`` scheme accordingly,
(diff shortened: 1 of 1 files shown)


docs/topics/settings.rst
```diff
@@ -420,10 +420,9 @@ connections (for ``HTTP10DownloadHandler``).
 .. note::
 
     HTTP/1.0 is rarely used nowadays so you can safely ignore this setting,
-    unless you use Twisted<11.1, or if you really want to use HTTP/1.0
-    and override :setting:`DOWNLOAD_HANDLERS_BASE` for ``http(s)`` scheme
-    accordingly, i.e. to
-    ``'scrapy.core.downloader.handlers.http.HTTP10DownloadHandler'``.
+    unless you really want to use HTTP/1.0 and override
+    :setting:`DOWNLOAD_HANDLERS_BASE` for ``http(s)`` scheme accordingly,
```
elacuesta, line 424: Since you're are it, could you improve the wording of this section? I think right now it's a bit confusing, it might be interpreted as if one needs to set `DOWNLOAD_HANDLERS_BASE` to use the http/1.0 handler, when the modified setting should be `DOWNLOAD_HANDLERS` instead.

kmike: A good idea, done in https://github.com/scrapy/scrapy/pull/4533/commits/fe6154e4faee375e7f47d61ceafabde7a3289bf3

PR #4534, opened 2020-05-04
Deprecate scrapy.item.BaseItem

elacuesta:
See https://github.com/scrapy/scrapy/pull/3881#discussion_r417317831


tests/test_item.py
```diff
@@ -321,13 +321,28 @@ class DictItemTest(unittest.TestCase):
...
+
+
+class BaseItemTest(unittest.TestCase):
+
+    def test_deprecation_warning(self):
+        with catch_warnings(record=True) as warnings:
+            BaseItem()
+            self.assertEqual(len(warnings), 1)
+            self.assertEqual(warnings[0].category, ScrapyDeprecationWarning)
+        with catch_warnings(record=True) as warnings:
```
kmike, line 342: What do you think about adding a test which checks explicitly that using Item itself, as well as its subclasses, doesn't issue a warning?

scrapy/item.py
```diff
@@ -28,7 +28,12 @@ class BaseItem(object_ref):
     Unlike instances of :class:`dict`, instances of :class:`BaseItem` may be
     :ref:`tracked <topics-leaks-trackrefs>` to debug memory leaks.
     """
-    pass
```
kmike, line 31: It would be nice to update BaseItem docstring. 

Do you know if there are other places in docs which need to be updated, after BaseItem deprecation? I guess it would require searching for BaseItem in Scrapy codebase and documentation.

One more question: any idea on how we would be able to remove it completely? :) There is code in various middlewares which check for `isinstance(item, BaseItem)`; would we be able to issue a deprecation warning for such checks in future?

Gallaecio: Regarding the latter, [it looks like it may be doable](https://stackoverflow.com/questions/9008444/how-to-warn-about-class-name-deprecation/9008509#comment91699108_26130976) (TIL).

Gallaecio:
Don’t forget [items.rst](https://github.com/scrapy/scrapy/blame/master/docs/topics/items.rst#L260).

elacuesta:
> Don’t forget items.rst.

I left it there for now because we technically  haven't removed it yet, but the autoclass directive will import the docstring which now warns about it being deprecated.

(Sorry, I meant to click "quote reply" but I ended up editing the original comment :sweat_smile: )

scrapy/item.py
elacuesta: I updated the docstring and updated a few `BaseItem` occurrences I missed before. This will probably create some conflicts with #3881, mostly regarding docs, but that's fine.

I couldn't find any occurrences of `BaseItem` in middlewares, just in other places which should work just fine by replacing `BaseItem` with `Item`:
```
$ grep -rn  "BaseItem\b" scrapy
scrapy/contracts/default.py:3:from scrapy.item import BaseItem
scrapy/contracts/default.py:54:        'item': (BaseItem, dict),
scrapy/contracts/default.py:55:        'items': (BaseItem, dict),
scrapy/contracts/default.py:106:            if isinstance(x, (BaseItem, dict)):
scrapy/utils/serialize.py:8:from scrapy.item import BaseItem
scrapy/utils/serialize.py:29:        elif isinstance(o, BaseItem):
scrapy/utils/misc.py:17:from scrapy.item import BaseItem
scrapy/utils/misc.py:20:_ITERABLE_SINGLE_VALUES = dict, BaseItem, str, bytes
scrapy/commands/parse.py:8:from scrapy.item import BaseItem
scrapy/commands/parse.py:120:            if isinstance(x, (BaseItem, dict)):
scrapy/core/scraper.py:17:from scrapy.item import BaseItem
scrapy/core/scraper.py:194:        elif isinstance(output, (BaseItem, dict)):
scrapy/core/scraper.py:203:            logger.error('Spider must return Request, BaseItem, dict or None, '
scrapy/shell.py:16:from scrapy.item import BaseItem
scrapy/shell.py:29:    relevant_classes = (Crawler, Spider, Request, Response, BaseItem,
scrapy/item.py:17:class BaseItem(object_ref):
scrapy/item.py:23:        if issubclass(cls, BaseItem) and not (issubclass(cls, Item) or issubclass(cls, DictItem)):
scrapy/item.py:24:            warn('scrapy.item.BaseItem is deprecated, please use scrapy.item.Item instead',
scrapy/item.py:26:        return super(BaseItem, cls).__new__(cls, *args, **kwargs)
scrapy/item.py:60:class DictItem(MutableMapping, BaseItem):
scrapy/item.py:105:    __hash__ = BaseItem.__hash__
scrapy/exporters.py:15:from scrapy.item import BaseItem
scrapy/exporters.py:315:        if isinstance(value, BaseItem):
```

Sorry, but TBH I'm not sure I fully understand the last question. Could you guys elaborate?

Gallaecio:
My take is that the documentation should not mention deprecated stuff outside `news.rst`. Someone reading the documentation for the first time should not need to bother reading about deprecated content.

elacuesta:
True, and to be fair it's not the first time we discuss this, my bad :hand:

scrapy/item.py
elacuesta: This should be addressed by 5256eae60d3685de51c1f3891abe157e15d14def. Introduced yet another (private) class, `_BaseItem`, as a way to avoid deprecation warnings when using `isinstance` internally. Most of the imports accross the codebase should go away after #3881, replaced by `itemadapter.is_item`.

scrapy/item.py
```diff
@@ -14,28 +14,39 @@
...
+                 ScrapyDeprecationWarning, stacklevel=2)
+        return super(BaseItem, cls).__new__(cls, *args, **kwargs)
 
 
 class Field(dict):
     """Container of field metadata"""
 
 
-class ItemMeta(ABCMeta):
+class ItemMeta(_BaseItemMeta):
```
Gallaecio, line 49: Is this necessary?

elacuesta: It is, otherwise bad things happen:
```python
In [1]: from scrapy.item import Item                                                                                                                                                                                                          
TypeError                                 Traceback (most recent call last)
<ipython-input-1-6456d4b4cb51> in <module>
----> 1 from scrapy.item import Item

~/scrapinghub/scrapy/scrapy/__init__.py in <module>
     32 
     33 # Declare top-level shortcuts
---> 34 from scrapy.spiders import Spider
     35 from scrapy.http import Request, FormRequest
     36 from scrapy.selector import Selector

~/scrapinghub/scrapy/scrapy/spiders/__init__.py in <module>
    111 
    112 # Top-level imports
--> 113 from scrapy.spiders.crawl import CrawlSpider, Rule  # noqa: F401
    114 from scrapy.spiders.feed import XMLFeedSpider, CSVFeedSpider  # noqa: F401
    115 from scrapy.spiders.sitemap import SitemapSpider  # noqa: F401

~/scrapinghub/scrapy/scrapy/spiders/crawl.py in <module>
     11 from scrapy.exceptions import ScrapyDeprecationWarning
     12 from scrapy.http import Request, HtmlResponse
---> 13 from scrapy.linkextractors import LinkExtractor
     14 from scrapy.spiders import Spider
     15 from scrapy.utils.python import get_func_args

~/scrapinghub/scrapy/scrapy/linkextractors/__init__.py in <module>
     14 
     15 from scrapy.utils.deprecate import ScrapyDeprecationWarning
---> 16 from scrapy.utils.misc import arg_to_iter
     17 from scrapy.utils.url import (
     18     url_is_from_any_domain, url_has_any_extension,

~/scrapinghub/scrapy/scrapy/utils/misc.py in <module>
     15 from scrapy.utils.datatypes import LocalWeakReferencedCache
     16 from scrapy.utils.python import flatten, to_unicode
---> 17 from scrapy.item import _BaseItem
     18 
     19 

~/scrapinghub/scrapy/scrapy/item.py in <module>
    136 
    137 
--> 138 class Item(DictItem, metaclass=ItemMeta):
    139     """
    140     Base class for scraped items.

TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

elacuesta: Metaclass magic :mage:

Gallaecio: :open_mouth:

TIL

elacuesta: Indeed, TIL too. We don't want to be invoking Cthulhu.

scrapy/item.py
```diff
@@ -127,4 +136,17 @@ def deepcopy(self):
...
+    In Scrapy, an object is considered an *item* if it is an instance of either
+    :class:`Item` or :class:`dict`. For example, when the output of a
+    spider callback is evaluated, only instances of :class:`Item` or
+    :class:`dict` are passed to :ref:`item pipelines <topics-item-pipeline>`.
+
+    If you need instances of a custom class to be considered items by Scrapy,
+    you must inherit from either :class:`Item` or :class:`dict`.
+
+    Unlike instances of :class:`dict`, instances of :class:`Item` may be
+    :ref:`tracked <topics-leaks-trackrefs>` to debug memory leaks.
```
kmike, line 151: This is not the only thing in which Item differs from dict; the most important part is that it allows to define fields. I think Item docstring should mentiion that, if we're adding it. The docstring as-is made sense for BaseItem, which didn't have these features.

kmike:
Looks good, thanks @elacuesta! I left a small comment about a docstring, but I think we're good to merge otherwise. 👍

elacuesta:
Updated, thanks @kmike

scrapy/item.py
```diff
@@ -14,28 +14,39 @@
...
+    pass
 
-    In Scrapy, an object is considered an *item* if it is an instance of either
-    :class:`BaseItem` or :class:`dict`. For example, when the output of a
-    spider callback is evaluated, only instances of :class:`BaseItem` or
-    :class:`dict` are passed to :ref:`item pipelines <topics-item-pipeline>`.
 
-    If you need instances of a custom class to be considered items by Scrapy,
-    you must inherit from either :class:`BaseItem` or :class:`dict`.
+class _BaseItemMeta(ABCMeta):
```
dangra, line 25: It looks like we are trying to reinvent `create_deprecated_class` util from https://github.com/scrapy/scrapy/blob/794ab19660d369f273abdd5b93721c209f6e4eab/scrapy/utils/deprecate.py#L18-L121

elacuesta: I looked (briefly) at that function before doing these changes, for some reason it didn't appear appropriate for this case. But I can't actually remember what that reason was, so I'll check again 👍

scrapy/item.py
```diff
@@ -127,4 +135,22 @@ def deepcopy(self):
...
+    In Scrapy, an object is considered an *item* if it is an instance of either
+    :class:`Item` or :class:`dict`. For example, when the output of a
+    spider callback is evaluated, only instances of :class:`Item` or
+    :class:`dict` are passed to :ref:`item pipelines <topics-item-pipeline>`.
+
+    If you need instances of a custom class to be considered items by Scrapy,
+    you must inherit from either :class:`Item` or :class:`dict`.
+
+    Items offer the ability to declare :class:`Field` attributes, which can be
+    used to define metadata and control the way data is processed internally.
```
kmike, line 150: I think the main use case if not storing metadata, but limiting a set of allowed keys, which allows e.g. to prevent typos in field names.

scrapy/item.py
elacuesta: Right, I tried this and remembered why I didn't use it. If I understand correctly, what you suggest is to do `BaseItem = create_deprecated_class('BaseItem', Item)`. The problem I see is that we don't want the deprecated `BaseItem` to have the functionality of `Item`, i.e. the field capabilities provided by the `ItemMeta` meta class:

```python
   ...: from scrapy.utils.deprecate import create_deprecated_class 
   ...:  
   ...: BaseItem = create_deprecated_class('BaseItem', Item, old_class_path="scrapy.item.BaseItem")  
   ...:  
   ...: class TestItem(BaseItem): 
   ...:     foo = Field() 
   ...:  
   ...: print(TestItem(foo="bar"))                                                                                                                                                                                                            
/.../scrapy/venv-scrapy/bin/ipython:6: ScrapyDeprecationWarning: __main__.TestItem inherits from deprecated class scrapy.item.BaseItem, please inherit from scrapy.item.Item. (warning only on first subclass, there may be others)
  
{'foo': 'bar'}
```

@dangra Please correct me if I'm wrong. Thanks!

elacuesta: Ping @dangra 😄

kmike:
Let's merge it :) @dangra - if you think something needs to be changed, please let us know, and we'll send a PR with a fix.

kmike:
Thanks @elacuesta!

PR #4536, opened 2020-05-06
Add support for schema-less HTTP proxy URLs (#4505)

### LINKED ISSUE #4505: Using proxy through http fails (https works)
### Description

When I scrape without proxy, both https and http urls work.
Using proxy through https works just fine. My problem is when I try http urls.
In that moment I get the `twisted.web.error.SchemeNotSupported: Unsupported scheme: b''` error

As I see, most of the people have this issue the other way around.

### Steps to Reproduce

1. Scrape a http link with proxy

**Expected behavior:** Get a 200 with the desired data.

**Actual behavior:**
```
ERROR: Error downloading <GET http://*********>
  File "/srv/scraper/venv/lib/python3.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
  File "/srv/scraper/venv/lib/python3.7/site-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/core/downloader/middleware.py", line 42, in process_request
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/utils/defer.py", line 55, in mustbe_deferred
    result = f(*args, **kw)
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/core/downloader/handlers/__init__.py", line 76, in download_request
    return handler.download_request(request, spider)
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/core/downloader/handlers/http11.py", line 82, in download_request
    return agent.download_request(request)
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/core/downloader/handlers/http11.py", line 361, in download_request
    d = agent.request(method, to_bytes(url, encoding='ascii'), headers, bodyproducer)
  File "/srv/scraper/venv/lib/python3.7/site-packages/scrapy/core/downloader/handlers/http11.py", line 262, in request
    endpoint=self._getEndpoint(self._proxyURI),
  File "/srv/scraper/venv/lib/python3.7/site-packages/twisted/web/client.py", line 1729, in _getEndpoint
    return self._endpointFactory.endpointForURI(uri)
  File "/srv/scraper/venv/lib/python3.7/site-packages/twisted/web/client.py", line 1607, in endpointForURI
    raise SchemeNotSupported("Unsupported scheme: %r" % (uri.scheme,))
twisted.web.error.SchemeNotSupported: Unsupported scheme: b''
```

**Reproduces how often:** Every time I scrape with proxy

### Versions
```
Scrapy       : 2.0.1
lxml         : 4.4.1.0
libxml2      : 2.9.9
cssselect    : 1.1.0
parsel       : 1.5.2
w3lib        : 1.21.0
Twisted      : 20.3.0
Python       : 3.7.3 (default, Apr  3 2019, 05:39:12) - [GCC 8.3.0]
cryptography : 2.7
Platform     : Linux-4.19.0-5-amd64-x86_64-with-debian-10.0
```
### Additional context

I tried to add some breakpoints at the end to see where it cracks.
I added the following lines in "twisted/web/client/py", before the cracking point:
```python
        endpoint = HostnameEndpoint(self._reactor, host, uri.port, **kwargs)
        import logging
        logger.error("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%")
        logger.error(uri)
        logger.error(uri.host)
        logger.error(uri.port)
        logger.error(uri.scheme)
        logger.error(dir(uri))
        if uri.scheme == b'http':
            return endpoint
        elif uri.scheme == b'https':
            connectionCreator = self._policyForHTTPS.creatorForNetloc(uri.host,
                                                                      uri.port)
            return wrapClientTLS(connectionCreator, endpoint)
        else:
            raise SchemeNotSupported("Unsupported scheme: %r" % (uri.scheme,))
```

Apparently in this point there is no schema. If I run the same code with a https url, this code is never reached. It seems that getting up to point there is bad and the proxy is not used

(edited to apply formatting)

HausCloud:
Constructive criticism would be much appreciated! I manually set http using setattr on self._proxyURI and that worked. I looked at Gallaecio's suggestion and implented his solution instead.

Fixes #4505

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -44,12 +44,12 @@ flake8-ignore =
     scrapy/commands/startproject.py E127 E501 E128
     scrapy/commands/version.py E501 E128
     # scrapy/contracts
-    scrapy/contracts/__init__.py E501 W504
+    scrapy/contracts/__init__.py E501
     scrapy/contracts/default.py E128
     # scrapy/core
     scrapy/core/engine.py E501 E128 E127
     scrapy/core/scheduler.py E501
-    scrapy/core/scraper.py E501 E128 W504
+    scrapy/core/scraper.py E501 E128
     scrapy/core/spidermw.py E501 E126
     scrapy/core/downloader/__init__.py E501
     scrapy/core/downloader/contextfactory.py E501 E128 E126
@@ -68,7 +68,7 @@ flake8-ignore =
     scrapy/downloadermiddlewares/httpcache.py E501 E126
     scrapy/downloadermiddlewares/httpcompression.py E501 E128
     scrapy/downloadermiddlewares/httpproxy.py E501
-    scrapy/downloadermiddlewares/redirect.py E501 W504
+    scrapy/downloadermiddlewares/redirect.py E501
     scrapy/downloadermiddlewares/retry.py E501 E126
     scrapy/downloadermiddlewares/robotstxt.py E501
     scrapy/downloadermiddlewares/stats.py E501
@@ -79,7 +79,7 @@ flake8-ignore =
     scrapy/extensions/httpcache.py E128 E501
     scrapy/extensions/memdebug.py E501
     scrapy/extensions/spiderstate.py E501
-    scrapy/extensions/telnet.py E501 W504
+    scrapy/extensions/telnet.py E501
     scrapy/extensions/throttle.py E501
     # scrapy/http
     scrapy/http/common.py E501
@@ -90,7 +90,7 @@ flake8-ignore =
     scrapy/http/response/__init__.py E501 E128
     scrapy/http/response/text.py E501 E128 E124
     # scrapy/linkextractors
-    scrapy/linkextractors/__init__.py E501 E402 W504
+    scrapy/linkextractors/__init__.py E501 E402
     scrapy/linkextractors/lxmlhtml.py E501
     # scrapy/loader
     scrapy/loader/__init__.py E501 E128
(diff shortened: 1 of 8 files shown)


HausCloud:
Hi Gallaecio, thanks for the tip. I moved it to the init method and removed the unnecessary back and forth with the parsing. I squashed it into one commit to avoid littering the git history and didn't think the if is necessary. If you need any extra changes just let me know. I'm honestly learning a lot in all of this so thanks! P.S. I may have dropped an extra line in there :/

scrapy/core/downloader/handlers/http11.py
```diff
@@ -250,14 +250,15 @@ def __init__(self, reactor, proxyURI, connectTimeout=None, bindAddress=None, poo
             bindAddress=bindAddress,
             pool=pool,
         )
-        self._proxyURI = URI.fromBytes(proxyURI)
+        self._proxyURI = URI.fromBytes(urlparse(proxyURI)._replace(scheme=bytes('http', 'utf-8')).geturl())
```
Gallaecio, line 253:
```suggestion
        self._proxyURI = URI.fromBytes(urlparse(proxyURI)._replace(scheme=b'http').geturl())
```

Gallaecio:
The next step would be a test. I’ve had a look, and it looks like there is some existing test that you could use as a basis: `HttpProxyTestCase.test_download_with_proxy` at `tests/test_downloader_handlers.py`. The same test with a `http_proxy` without protocol should do the job, I think.

It would be great if you also locate a similar test for HTTPS, and add a similar counterpart without protocol. That way we make sure we do not break the similar feature of HTTPS in the future.

HausCloud:
Thanks for hint in the right direction Gallaecio. I should have it done by the end of today.

Gallaecio:
@HausCloud Did you close the pull request accidentally?

HausCloud:
Yes I did by mistakes. I was trying to incorporate changes from scrapy into my fork. Also, I'm getting an error when I try https. I adjusted the if condition so it always handles it as https and didn't provide a schema. Below is the error I got.
Screen Shot 2020-05-06 at 4 59 55 PM

Gallaecio:
Feel free to reopen the pull request just with the HTTP test.

PR #4541, opened 2020-05-06
Pickle: use protocol 4, update tests

### LINKED ISSUE #4135: Switch pickle to protocol 4
Since we no longer support Python 2 or Python < 3.4, it looks like [protocol 4 of pickle](https://docs.python.org/3/library/pickle.html#data-stream-format) is our best choice.

elacuesta:
Fixes #4135

I started by adding a `PICKLE_PROTOCOL` setting, but that didn't work well with `scrapy.squeues._pickle_serialize` because it's defined as a top level function. There is probably a way around that, I just didn't think it was worth the trouble.

I simplified the testing of unserializable objects, since https://twistedmatrix.com/trac/ticket/7989 only affects py2 AFAICT.

A simple and non-rigorous benchmark shows a performance improvement of about 10%:

```python
from tempfile import mkdtemp
from scrapy import Request
from scrapy.utils.test import get_crawler
from scrapy.squeues import PickleFifoDiskQueue

def test(request_count):
    jobdir = mkdtemp()
    crawler = get_crawler(settings_dict={"JOBDIR": jobdir})
    queue = PickleFifoDiskQueue(crawler, jobdir)
    for i in range(request_count):
        queue.push(Request("https://example.org/{}".format(i)))
```

Protocol 2 (3878b67a3771102d4b6668ac749afbec7dc85a8f):
```python
In [2]: %timeit -n 10 test(10**5)                                                                                                                                                                                                             
7.63 s ± 75.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Protocol 4:
```python
In [2]: %timeit -n 10 test(10**5)                                                                                                                                                                                                             
6.87 s ± 11.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Additionally, I reworded the comment about `pickle.PicklingError` being raised in Python <= 3.4. We no longer support 3.4, but I think we should keep catching `pickle.PicklingError` since it's raised when trying to pickle a `lambda` function in 3.5 <= Python <= 3.8. For the record, the test actually raises `AttributeError("Can't pickle local object 'PickleFifoDiskQueueTest.test_non_pickable_object.<locals>.<lambda>'")` in CPython, but `pickle.PicklingError` is raised in `pypy`.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -166,6 +166,7 @@ flake8-ignore =
     scrapy/signalmanager.py E501
     scrapy/spiderloader.py F841 E501 E126
     scrapy/squeues.py E128
+    scrapy/squeues.py E501
     scrapy/statscollectors.py E501
     # tests
     tests/__init__.py E402 E501
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -250,7 +250,7 @@ class CsvItemExporter(BaseItemExporter):
 
 class PickleItemExporter(BaseItemExporter):
 
-    def __init__(self, file, protocol=2, **kwargs):
+    def __init__(self, file, protocol=4, **kwargs):
         super().__init__(**kwargs)
         self.file = file
         self.protocol = protocol
--- a/scrapy/extensions/httpcache.py
+++ b/scrapy/extensions/httpcache.py
@@ -250,7 +250,7 @@ class DbmCacheStorage:
             'headers': dict(response.headers),
             'body': response.body,
         }
-        self.db['%s_data' % key] = pickle.dumps(data, protocol=2)
+        self.db['%s_data' % key] = pickle.dumps(data, protocol=4)
         self.db['%s_time' % key] = str(time())
 
     def _read_data(self, spider, request):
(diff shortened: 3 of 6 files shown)


Gallaecio:
Given the backward incompatibility implications, such as cache invalidation, I think a setting is still worth exploring.

Thinking of users, I think we should only upgrade the version of pickle automatically for new projects (by adding a setting override to the project settings template). We could then cover the setting in some topic or section about performance, as one way to improve Scrapy performance in CPU-bound scenarios (and disk-bound scenarios?).

elacuesta:
I don't think we need to worry about that, since `pickle.load(s)` are able to identify the used protocol. From the [`pickle` module docs](https://docs.python.org/3/library/pickle.html#pickle.load):
> The protocol version of the pickle is detected automatically, so no protocol argument is needed

elacuesta:
Tests are failing only on pypy3. In the past, I've encountered that pypy3 handles `pickle` somewhat differently from CPython, but I don't usually test it locally before pushing. I'll take a look.

kmike:
Thanks @elacuesta!

PR #4542, opened 2020-05-06
Flake8: remove E122 and E124

elacuesta:
* Continuation line missing indentation or outdented (E122)
* Closing bracket does not match visual indentation (E124)


Gallaecio:
I would expect the XML test to fail, given the new code adds extra spaces. We should either ignore the corresponding Flake8 check or use something to remove the spaces from the string.

elacuesta:
I found that those tests look for the document structure, they failed if I altered the name of some tag.

PR #4543, opened 2020-05-06
Flake8: remove E126

elacuesta:
_Continuation line over-indented for hanging indent_

~Based on #4542 to avoid merge conflicts.~ (merged)

### DIFF
--- a/scrapy/spiderloader.py
+++ b/scrapy/spiderloader.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
-from collections import defaultdict
 import traceback
 import warnings
+from collections import defaultdict
 
 from zope.interface import implementer
 
@@ -16,6 +16,7 @@ class SpiderLoader:
     SpiderLoader is a class which locates and loads spiders
     in a Scrapy project.
     """
+
     def __init__(self, settings):
         self.spider_modules = settings.getlist('SPIDER_MODULES')
         self.warn_only = settings.getbool('SPIDER_LOADER_WARN_ONLY')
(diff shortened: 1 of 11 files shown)


scrapy/spiderloader.py
```diff
@@ -24,15 +24,17 @@ def __init__(self, settings):
...
-                           for (mod, cls) in locations)
-                 for name, locations in self._found.items()
-                 if len(locations) > 1]
+        dupes = []
+        for name, locations in self._found.items():
+            dupes.extend([
+                "  {cls} named {name!r} (in {module})".format(module=mod, cls=cls, name=name)
+                for mod, cls in locations
+            ])
+
```
elacuesta, line 33: This might be a little bit more verbose, but way more readable IMHO. Personally, I find nested `for` lops hard to follow.

kmike: It seems `if len(locations) > 1` check is removed in the new code. How does it work? Do we have tests for it?

elacuesta: Thanks for noticing this. I thought the list would be empty in that case, but it's `>1` and not `>=1`. I'll fix it.

kmike: I was thinking that's the purpose of the function - warn in case of duplicates, i.e. a single name with 2+ locations; that's bad that our tests don't catch it.

elacuesta: Makes sense. I'll open a issue to keep track of this, I don't think we should add it in this PR.

elacuesta: Done: #4549. I skipped the issue template entirely :sweat_smile:

PR #4552, opened 2020-05-07
Code sharing between crawl and runspider command

### LINKED ISSUE #4548: Increase code sharing between the crawl and runspider commands
Their `add_options` and `process_options` methods are identical. They could be moved into a shared, parent class.

jay24rajput:
Implemented a new file [common_commands](https://github.com/jay24rajput/scrapy/blob/common_commands/scrapy/commands/common_commands.py) for reusing the ```add_options``` and ```process_option``` command previously used in ```crawl.py``` and ```runspider.py```

Fixes #4548


Gallaecio:
Instead of creating a new file, what about using `commands/__init__.py`?

jay24rajput:
> Instead of creating a new file, what about using `commands/__init__.py`?

There is actually same function i.e ```add_options``` and ```process_options``` in ```commands/__init__.py``` file. And these functions initialize the parser with some common options.

In case we add the code to the __init__.py file, the options will be displayed for each file, which might not be relevant there. Thus, I created a new file so has to avoid the above redundancy.

Gallaecio:
> There is actually same function i.e `add_options` and `process_options` in `commands/__init__.py` file. And these functions initialize the parser with some common options.
> 
> In case we add the code to the` __init__.py` file, the options will be displayed for each file, which might not be relevant there. Thus, I created a new file so has to avoid the above redundancy.

I’m not sure what you mean by “the options will be displayed for each file”.

Those two objects are methods of a class, not module-level functions. If you rename your new `CommonCommands` class to something else (e.g. `BaseRunSpiderCommand`), you could just move that class to the `__init__.py` file without any conflict.

jay24rajput:
Alright! I sort of had a confusion. I will move the entire class to ```__init__.py``` and will make both the ```crawl.py``` and ```runspider.py``` call the newly created class.

jay24rajput:
@Gallaecio codecov for the patch is failing. Can you please guide me as to how I can fix it. 

Also please let me know if anything else needs to be rectified in the previous commit.

scrapy/commands/__init__.py
```diff
@@ -104,3 +106,35 @@ def run(self, args, opts):
         Entry point for running commands
         """
         raise NotImplementedError
+
+
+''''
+The BaseRunSpiderCommands class inherits the ScrapyCommand class and it Used for
+performing common functionality between crawl.py and runspider.py
+'''
```
Gallaecio, line 114: Move this right under the `class …` line below, indented.

scrapy/commands/__init__.py
```diff
@@ -79,7 +80,8 @@ def process_options(self, args, opts):
             self.settings.setdict(arglist_to_dict(opts.set),
                                   priority='cmdline')
         except ValueError:
-            raise UsageError("Invalid -s value, use -s NAME=VALUE", print_help=False)
+            raise UsageError(
+                "Invalid -s value, use -s NAME=VALUE", print_help=False)
```
Gallaecio, line 84: No need to change this line. In Scrapy, even though aiming for 79-character-long lines, style changes to existing code because of line length should be avoided as long as line length is within our hard limit: 119 characters.

It also applies to a similar line in `crawl.py`.

scrapy/commands/__init__.py
```diff
@@ -104,3 +106,35 @@ def run(self, args, opts):
...
+class BaseRunSpiderCommands(ScrapyCommand):
+
+    def add_options(self, parser):
+        ScrapyCommand.add_options(self, parser)
+        parser.add_option("-a", dest="spargs", action="append", default=[],
+                          metavar="NAME=VALUE",
+                          help="set spider argument (may be repeated)")
+        parser.add_option("-o", "--output", metavar="FILE", action="append",
+                          help="dump scraped items into FILE"
+                          + "(use - for stdout)")
```
Gallaecio, line 126: My line length comment also applies here. You should copy these two methods as is from their original location, there should be no need to change their formatting.

scrapy/commands/crawl.py
```diff
@@ -1,9 +1,8 @@
-from scrapy.commands import ScrapyCommand
-from scrapy.utils.conf import arglist_to_dict, feed_process_params_from_cli
 from scrapy.exceptions import UsageError
+from scrapy.commands import BaseRunSpiderCommands
```
Gallaecio, line 2: Let’s keep it alphabetical.

```suggestion
from scrapy.commands import BaseRunSpiderCommands
from scrapy.exceptions import UsageError
```

Gallaecio:
Feel free to ignore codecov in this case. You are modifying lines that are not covered by unit tests, so obviously your changes include lines not covered by unit tests, but given how straightforward these changes are, I don’t think we should consider test coverage part of the scope of these changes.

jay24rajput:
@Gallaecio is there anything else that needs to be done here?

scrapy/commands/__init__.py
```diff
@@ -104,3 +105,30 @@ def run(self, args, opts):
         Entry point for running commands
         """
         raise NotImplementedError
+
+
+class BaseRunSpiderCommands(ScrapyCommand):
+    ''''
+    The BaseRunSpiderCommands class inherits the ScrapyCommand class and it Used for
+    performing common functionality between crawl.py and runspider.py
+    '''
```
Gallaecio, line 114:
```suggestion
    """
    The BaseRunSpiderCommands class inherits the ScrapyCommand class and it Used for
    performing common functionality between crawl.py and runspider.py
    """
```

scrapy/commands/__init__.py
elacuesta: @jay24rajput Please address this before we can merge. Thanks!

jay24rajput: @elacuesta I have already covered it [here](https://github.com/scrapy/scrapy/pull/4552/commits/ed4f4f84082f7023734d1200d88e8f1aec57904c) IMO. Please have a look and let me know if anything is missing. Thanks!

jay24rajput: @elacuesta Done! Is there anything else that needs to be done? Thanks!

jay24rajput: @elacuesta the Travis CI build is failing. On the local system ```tox scrapy``` is building successfully and along flake8. Can you please help me find where the problem is?

elacuesta: Right, I've seen this error in other PRs as well, it seems to be directly related to `pytest-flake8`:
```
.tox/flake8/lib/python3.8/site-packages/pytest_flake8.py:191: in check_file
    app.parse_preliminary_options_and_args(args)
E   AttributeError: 'Application' object has no attribute 'parse_preliminary_options_and_args'
```
This seems safe and unrelated to that, I'll open a new issue to keep track of the build error.

jay24rajput: Alright so are we good to merge??

scrapy/commands/__init__.py
```diff
@@ -104,3 +104,28 @@ def run(self, args, opts):
         Entry point for running commands
         """
         raise NotImplementedError
+
+
+class BaseRunSpiderCommands(ScrapyCommand):
```
elacuesta, line 109:
```suggestion
class BaseRunSpiderCommand(ScrapyCommand):
```
And the corresponding updates in child classes.

scrapy/commands/__init__.py
```diff
@@ -104,3 +104,28 @@ def run(self, args, opts):
         Entry point for running commands
         """
         raise NotImplementedError
+
+
+class BaseRunSpiderCommand(ScrapyCommand):
+    """
+    The BaseRunSpiderCommands class inherits the ScrapyCommand class and it Used for
+    performing common functionality between crawl.py and runspider.py
```
elacuesta, line 112:
```suggestion
    Common class used to share functionality between the crawl and runspider commands
```
Simplify, remove a minor typo and duplicated information (no need to name the classes nor their inheritance, it's clear from the class definition)

jay24rajput:
@elacuesta @Gallaecio the checks haven't completed yet! Any solutions?

noviluni:
Hi @jay24rajput it is working, the error is not related to the code but to connection errors. However, if you want to relaunch the check you can close and open again this PR.

PR #4554, opened 2020-05-08
Remove lambda attributes in default link extractor

elacuesta:
Removes `lambda` attributes in the default link extractor, allowing instances to be serialized using `pickle`.

Motivated by https://github.com/scrapinghub/scrapy-autounit/issues/74, but this would also mean that requests which include link extractors in their `meta` attribute would be serializable too. I imagine that might be a rare use case, most link extractors I've seen are stored as spider attributes, but the possibility exists.

I intentionally left extractors other than `LxmlParserLinkExtractor` unmodified because they're deprecated. Let me know if that's an issue, I could update them as well.

/cc @fcanobrash

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -28,8 +28,9 @@ def _nons(tag):
 
 
 class LxmlParserLinkExtractor:
-    def __init__(self, tag="a", attr="href", process=None, unique=False,
-                 strip=True, canonicalized=False):
+    def __init__(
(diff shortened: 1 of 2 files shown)


kmike:
👍

ryonlife:
I opened https://github.com/scrapinghub/scrapy-autounit/issues/74, so just wanted to say thanks!

I also took a crack at a generalized fix inside of scrapy-autounit at https://github.com/scrapinghub/scrapy-autounit/pull/75.

PR #4555, opened 2020-05-08
Deprecate body as unicode

### LINKED ISSUE #4546: Deprecate Response.body_as_unicode

nsirletti:
Fixes #4546 

My first PR on this project so advice/corrections are very welcome.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -835,7 +835,6 @@ TextResponse objects
     .. automethod:: TextResponse.follow_all
 
 
-
 HtmlResponse objects
 --------------------
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -834,8 +834,6 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
-        The same as :attr:`text`, but available as a method. This method is
-        kept for backward compatibility; please prefer ``response.text``.
 
 
 HtmlResponse objects
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -834,8 +834,6 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
-    .. method:: TextResponse.body_as_unicode()
-
         The same as :attr:`text`, but available as a method. This method is
         kept for backward compatibility; please prefer ``response.text``.
 
--- a/scrapy/http/response/text.py
+++ b/scrapy/http/response/text.py
@@ -16,9 +16,9 @@ from w3lib.html import strip_html5_whitespace
 
 from scrapy.http import Request
 from scrapy.http.response import Response
+from scrapy.utils.decorators import deprecated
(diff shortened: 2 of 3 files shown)


docs/topics/request-response.rst
```diff
@@ -834,8 +834,6 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
-    .. method:: TextResponse.body_as_unicode()
-
         The same as :attr:`text`, but available as a method. This method is
         kept for backward compatibility; please prefer ``response.text``.
```
Gallaecio, line 838: These 2 lines also need to be removed.

scrapy/http/response/text.py
```diff
@@ -18,6 +18,7 @@
 from scrapy.http.response import Response
 from scrapy.utils.python import memoizemethod_noargs, to_unicode
 from scrapy.utils.response import get_base_url
+from scrapy.utils.decorators import deprecated
```
Gallaecio, line 21: We aim for alphabetical order:

```suggestion
from scrapy.utils.decorators import deprecated
from scrapy.utils.python import memoizemethod_noargs, to_unicode
from scrapy.utils.response import get_base_url
```

nsirletti:
Thanks a lot for the review ! I'm on it.

Now this might be a weird question but all this is very new to me.  From this point should git revert the test removal and create a new commit, right ?  As opposed to rebasing to get a cleaner log.

elacuesta:
Thanks for the PR @nsirletti! Don't worry about the commit history, we can always squash when merging.
Regarding tests, please remember to add a check to make sure the deprecation warning is actually logged. You could find a similar case [here](https://github.com/scrapy/scrapy/blob/2.1.0/tests/test_item.py#L263-L273).

docs/topics/request-response.rst
```diff
@@ -834,10 +834,6 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
```
Gallaecio, line 836: This is quite the nitpick, but you should remove one more empty line, there are 3 now and there should be only 2 :slightly_smiling_face:

scrapy/http/response/text.py
```diff
@@ -5,6 +5,8 @@
 See documentation in docs/topics/request-response.rst
 """
 
+import warnings
+
```
Gallaecio, line 9: We don’t leave an empty line in these cases.

scrapy/http/response/text.py
```diff
@@ -14,6 +16,8 @@
                             http_content_type_encoding, resolve_encoding)
 from w3lib.html import strip_html5_whitespace
 
+
```
Gallaecio, line 19: There should be a single empty here.

docs/topics/request-response.rst
nsirletti: Well it's minute but important.  Thanks for the remark I'll try and be more thorough next time !

Gallaecio:
Thanks!

PR #4558, opened 2020-05-08
Flake8: Remove E741

elacuesta:
_Do not use variables named 'I', 'O', or 'l'_


wRAR:
@elacuesta there are conflicts here

elacuesta:
Updated, but the build will fail until we merge #4572, let's wait for that one.

scrapy/shell.py
```diff
@@ -146,14 +146,13 @@ def get_help(self):
...
-                     "Fetch URL and update local objects "
-                     "(by default, redirects are followed)")
+                     "Fetch URL and update local objects (by default, redirects are followed)")
             b.append("  fetch(req)                  "
                      "Fetch a scrapy.Request and update local objects ")
         b.append("  shelp()           Shell help (print this help)")
         b.append("  view(response)    View response in a browser")
 
-        return "\n".join("[s] %s" % l for l in b)
+        return "\n".join("[s] %s" % line for line in b)
```
noviluni, line 155: Hi @elacuesta 

Maybe it's not necessary to do it as part of this PR, but as we are giving more meaning to a lot of variable names, it could be a good idea to rename this `b` variable to something like `banner` (I assume this is the real meaning by looking at this: `self.vars['banner'] = self.get_help()`).

elacuesta: Indeed, I almost did that, but I decided not to in order to keep the diff simpler. Perhaps we can do it in a later PR :+1:

elacuesta:
Rebased on top of #4572

elacuesta:
Updated, reverted changes to `tests/test_loader.py`. Will merge once CI passes.

noviluni:
hi @elacuesta is there something blocking to work on `tests/test_loader.py`? I can't understand if it was reverted because it was hard to fix conflicts or if there is any other reason

elacuesta:
Hey @noviluni, it was mostly to avoid adding more changes to #4516 and the need to import the changes in `itemloaders`.

noviluni:
> Hey @noviluni, it was mostly to avoid adding more changes to #4516 and the need to import the changes in `itemloaders`.

Ok, it makes sense. Thank you!

PR #4559, opened 2020-05-10
Stop response download from signal handler

elacuesta:
Allow the cancellation of a download while it's in progress.

Motivated by https://github.com/scrapy/scrapy/pull/4205#issuecomment-588155415.

Tasks:
- [x] Implementation 
- [x] Tests
- [x] Docs
    - [x] Exceptions topic
    - [x] Request/Response topic
    - [x] FAQ entry
    - [x] Note in the `bytes_received` signal docs

### DIFF
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -14,13 +14,6 @@ Built-in Exceptions reference
 
 Here's a list of all exceptions included in Scrapy and their usage.
 
-DropItem
---------
-
-.. exception:: DropItem
-
-The exception that must be raised by item pipeline stages to stop processing an
-Item. For more information see :ref:`topics-item-pipeline`.
 
 CloseSpider
 -----------
@@ -46,6 +39,14 @@ DontCloseSpider
 
 This exception can be raised in a :signal:`spider_idle` signal handler to
 prevent the spider from being closed.
+
+DropItem
+--------
+
+.. exception:: DropItem
+
+The exception that must be raised by item pipeline stages to stop processing an
+Item. For more information see :ref:`topics-item-pipeline`.
 
 IgnoreRequest
 -------------
@@ -77,3 +78,34 @@ NotSupported
 
 This exception is raised to indicate an unsupported feature.
 
+StopDownload
+-------------
+
+.. versionadded:: 2.2
+
+.. exception:: StopDownload(fail=True)
+
+Raised from a :class:`~scrapy.signals.bytes_received` signal handler to
+indicate that no further bytes should be downloaded for a response.
+
+The ``fail`` boolean parameter controls which method will handle the resulting
+response:
+
+* If ``fail=True``, the request errback is called. The response object is
+  available as the ``response`` attribute of the received
+  :class:`~twisted.python.failure.Failure` object. This is the default behaviour.
+
+* If ``fail=False``, the request callback is called instead.
+
+In both cases, the response could have its body truncated: the body contains
+all bytes received up until the exception is raised, including the bytes
+received in the signal handler that raises the exception. Also, the response
+object is marked with ``"download_stopped"`` in its :attr:`Response.flags`
+attribute.
+
+.. note:: ``fail`` is a keyword-only parameter, i.e. raising
+    ``StopDownload(False)`` or ``StopDownload(True)`` will raise
+    a :class:`TypeError`.
+
+See the documentation for the :class:`~scrapy.signals.bytes_received` signal
+and the :ref:`topics-stop-response-download` topic for additional information and examples.
--- a/docs/topics/exceptions.rst
+++ b/docs/topics/exceptions.rst
@@ -91,9 +91,12 @@ indicate that no further bytes should be downloaded for a response.
 The ``fail`` boolean parameter controls which method will handle the resulting
 response:
 
-* If ``fail=True``, the request errback is called. The response object is
-  available as the ``response`` attribute of the received
-  :class:`~twisted.python.failure.Failure` object. This is the default behaviour.
+* If ``fail=True`` (default), the request errback is called. The response object is
+  available as the ``response`` attribute of the ``StopDownload`` exception,
+  which is in turn stored as the ``value`` attribute of the received
+  :class:`~twisted.python.failure.Failure` object. This means that in an errback
+  defined as ``def errback(self, failure)``, the response can be accessed though
+  ``failure.value.response``.
 
 * If ``fail=False``, the request callback is called instead.
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -385,6 +385,51 @@ The meta key is used set retry times per request. When initialized, the
 :reqmeta:`max_retry_times` meta key takes higher precedence over the
 :setting:`RETRY_TIMES` setting.
 
+
+.. _topics-stop-response-download:
+
+Stopping the download of a Response
+===================================
+
+Raising a :exc:`~scrapy.exceptions.StopDownload` exception from a
+:class:`~scrapy.signals.bytes_received` signal handler will stop the
+download of a given response. See the following example::
+
+    import scrapy
+
+
+    class StopSpider(scrapy.Spider):
+        name = "stop"
+        start_urls = ["https://docs.scrapy.org/en/latest/"]
+
+        @classmethod
+        def from_crawler(cls, crawler):
+            spider = super().from_crawler(crawler)
+            crawler.signals.connect(spider.on_bytes_received, signal=scrapy.signals.bytes_received)
+            return spider
+
+        def parse(self, response):
+            # 'last_chars' show that the full response was not downloaded
+            yield {"len": len(response.text), "last_chars": response.text[-40:]}
+
+        def on_bytes_received(self, data, request, spider):
+            raise scrapy.exceptions.StopDownload(fail=False)
+
+which produces the following output::
+
+    2020-05-19 17:26:12 [scrapy.core.engine] INFO: Spider opened
+    2020-05-19 17:26:12 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
+    2020-05-19 17:26:13 [scrapy.core.downloader.handlers.http11] DEBUG: Download stopped for <GET https://docs.scrapy.org/en/latest/> from signal handler StopSpider.on_bytes_received
+    2020-05-19 17:26:13 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://docs.scrapy.org/en/latest/> (referer: None) ['download_stopped']
+    2020-05-19 17:26:13 [scrapy.core.scraper] DEBUG: Scraped from <200 https://docs.scrapy.org/en/latest/>
+    {'len': 279, 'last_chars': 'dth, initial-scale=1.0">\n  \n  <title>Scr'}
+    2020-05-19 17:26:13 [scrapy.core.engine] INFO: Closing spider (finished)
+
+By default, resulting responses are handled by their corresponding errbacks. To
+call their callback instead, like in this example, pass ``fail=False`` to the
+:exc:`~scrapy.exceptions.StopDownload` exception.
+
+
 .. _topics-request-response-ref-request-subclasses:
 
 Request subclasses
@@ -716,9 +761,9 @@ Response objects
         .. versionadded:: 2.1.0
 
         The IP address of the server from which the Response originated.
-        
+
         This attribute is currently only populated by the HTTP 1.1 download
-        handler, i.e. for ``http(s)`` responses. For other handlers, 
+        handler, i.e. for ``http(s)`` responses. For other handlers,
         :attr:`ip_address` is always ``None``.
 
     .. method:: Response.copy()
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -442,12 +442,8 @@ class ScrapyAgent:
             ip_address=result["ip_address"],
         )
         if result.get("failure"):
-            # This failure is not the same object that will reach the errback,
-            # so we need to temporarily store the response in the exception.
-            # It will be moved to the failure in core/scraper.py
-            failure = result["failure"]
-            failure.value.response = response
-            return failure
+            result["failure"].value.response = response
+            return result["failure"]
         return response
 
 
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -12,6 +12,7 @@ from urllib.parse import urldefrag
 from twisted.internet import defer, protocol, ssl
 from twisted.internet.endpoints import TCP4ClientEndpoint
 from twisted.internet.error import TimeoutError
+from twisted.python.failure import Failure
 from twisted.web.client import Agent, HTTPConnectionPool, ResponseDone, ResponseFailed, URI
 from twisted.web.http import _DataLoss, PotentialDataLoss
 from twisted.web.http_headers import Headers as TxHeaders
@@ -21,7 +22,7 @@ from zope.interface import implementer
 from scrapy import signals
 from scrapy.core.downloader.tls import openssl_methods
 from scrapy.core.downloader.webclient import _parse
-from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.exceptions import ScrapyDeprecationWarning, StopDownload
 from scrapy.http import Headers
 from scrapy.responsetypes import responsetypes
 from scrapy.utils.misc import create_instance, load_object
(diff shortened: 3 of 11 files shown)


scrapy/signalmanager.py
```diff
@@ -1,6 +1,7 @@
-from pydispatch import dispatcher
 from scrapy.utils import signal as _signal
 
+from pydispatch import dispatcher
+
```
Gallaecio, line 4: :lipstick: Third parties before the current library (Scrapy).

tests/test_engine.py
```diff
@@ -357,6 +369,46 @@ def test_close_engine_spiders_downloader(self):
...
+            self._assert_signals_caught()
+            self._assert_bytes_received()
+
+    def _assert_bytes_received(self):
+        self.assertEqual(9, len(self.run.bytes))
+        for request, data in self.run.bytes.items():
+            joined_data = b"".join(data)
+            self.assertTrue(len(data) == 1)  # signal was fired only once
+            if self.run.getpath(request.url) == "/numbers":
+                self.assertTrue(len(data) == 1)  # signal was fired only once
```
Gallaecio, line 404: This seems redundant, given the line two lines above.

docs/topics/request-response.rst
```diff
@@ -385,6 +385,49 @@ The meta key is used set retry times per request. When initialized, the
...
+            spider = super().from_crawler(crawler)
+            crawler.signals.connect(spider.handler, signal=scrapy.signals.bytes_received)
+            return spider
+
+        def parse(self, response):
+            # 'last_chars' show that the full response was not downloaded
+            yield {"len": len(response.text), "last_chars": response.text[-40:]}
+
+        def handler(self, data, request, spider):
+            if b"some text that I need" not in data:  # data is a 'bytes' object
```
kmike, line 416: This example is unfortunately not correct: "some text" + "that I need" can be in different chunks.

scrapy/signalmanager.py
elacuesta: I don't even know how this change got here! :shrug:

tests/test_engine.py
elacuesta: Oh right, forgot to remove this one :+1:

docs/topics/request-response.rst
elacuesta: haha fair enough. I initially had `if some_condition:` but I got concerned about someone reporting that `some_condition` raised a `NameError`.

docs/topics/exceptions.rst
```diff
@@ -14,13 +14,14 @@ Built-in Exceptions reference
...
---------
+CancelRequest
+-------------
 
-.. exception:: DropItem
+.. exception:: CancelRequest
 
-The exception that must be raised by item pipeline stages to stop processing an
-Item. For more information see :ref:`topics-item-pipeline`.
+Raised to indicate that no further bytes should be downloaded for a request.
```
kmike, line 22: I think the approach of raising an exception from a singnal handler is good, but we need to explain better the semantics, and check how it works with other parts of Scrapy.

1. Signal name. For cancelled request I'd not expect callback to be called. Here we're not cancelling a request (request is already sent); we're stopping/aborting download of the response. 
2. Scrapy can also abort download on different conditions, such as max response size. I think in this case errback is called, not callback. Should we mention this in docs? There is also DOWNLOAD_FAIL_ON_DATALOSS option which look related, though I'm not sure it should affect the behavior of the cancel exception. 
3. I'm not sure what's better, call a callback or an errback, and if there is a single best approach. Maybe there should be 2 exceptions (or a bool flag in exception), one which make download successful, and another one for errors. E.g. in the FAQ entry you're writing "detect early that its response won't be useful", but if response is not useful, it shouldn't go to a callback.
4. Are there other places where this exception can be raised? I don't have anything in mind, but the CancelRequest exception look similar to IgnoreRequest which can be raised by a downloader middlewares; this can be confusing. Are there use cases for catching CancelRequest from other signals?

docs/topics/request-response.rst
kmike: That's the great thing about DDD (docs driven development) - you try writing an usage example, it becomes complicated, and you then extract some of these complexity to helpers and add them to a library you're developing. So I think a real example could be more useful for users; they're 100% going to make the same mistake.

docs/topics/exceptions.rst
elacuesta: Those are very good points, thank you.

(1) I don't particularly love the current name, and I agree with the fact that the "request is already sent", so it's not completely accurate. What about `(Abort/Interrupt/Stop)Response`?

(2/3) I think that since this behaviour would be controlled explicitly by user code, and not because of an unexpected error, we should call the assigned callback. That said, providing a way to fire the errback instead sounds fine to me too. What about an `errback` boolean argument in the exception, which defaults to `False`?

(4) I don't think we could catch this exception anywhere else, since the signal is directly sent from the download handler. Responses in middlewares are already complete.

Gallaecio: (1) Based on the documentation, `(Abort/Cancel/Interrupt/Stop)Download`?

elacuesta: Sounds even better. At point point I thought about `(Verb)(Response/Body)Download`, since what we interrupt is technically only the body, but that was too verbose, and probably unnecessary.

elacuesta: I went for `StopDownload`. Also reworded the docs from "the response might not be useful" to "the already downloaded content might be enough". Given that (and my comment below about this being "controlled explicitly by user code") I think it'd be good to fire the callback instead of the errback. Thoughts?

Gallaecio: Adding a `DOWNLOAD_FAIL_ON_DOWNLOAD_STOP` (or `DOWNLOAD_FAIL_ON_STOP`?) setting may make sense, like `DOWNLOAD_FAIL_ON_DATALOSS`.

I have no strong on what the default behavior should be. While the error is caused by users, as @elacuesta said, I can imagine users desiring both behaviors:
- If the signal is used to stop downloads once the desired information has been downloaded, `callback` makes sense.
- If the signal is used to stop downloads as soon as they are detected as bad responses based on content, `errback` makes sense to cause a retry (as it happens for data loss with `DOWNLOAD_FAIL_ON_DATALOSS=True`).

scrapy/core/downloader/handlers/http11.py
```diff
@@ -440,6 +441,14 @@ def _cb_bodydone(self, result, request, url):
             certificate=result["certificate"],
             ip_address=result["ip_address"],
         )
+        if result.get("failure"):
+            # This failure is not the same object that will reach the errback,
+            # so we need to temporarily store the response in the exception.
+            # It will be moved to the failure in core/scraper.py
+            failure = result["failure"]
+            failure.value.response = response
```
elacuesta, line 449: I wish I could do `failure.response = response`, but the attribute gets lost somewhere and it doesn't reach the errback. I'm open to suggestions if you can think of a more elegant alternative.

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,25 @@ NotSupported
 
 This exception is raised to indicate an unsupported feature.
 
+StopDownload
+-------------
+
+.. versionadded:: 2.2
+
+.. exception:: StopDownload(errback=False)
```
kmike, line 86: Sorry, one more query, about naming :) `errback` can be confused with errback argument of Request; previously all "errback" arguments were functions, not bools. I wonder if it is possible to have a flag name without this issue. Some ideas (more ideas are welcome):

* `StopDownload(fail=True)`, `StopDownload(fail=False)`
* `StopDownload(ok=True)`, `StopDownload(ok=False)`
* `StopDownload` and `CancelDownload`
* `StopDownload` and `IgnoreRequest`

elacuesta: I like `StopDownload(fail)`. Thanks!

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,25 @@ NotSupported
...
+
+The ``fail`` boolean parameter indicates whether or not the resulting partial
+response should be handled by the request errback. In this case, the response
+will be available in the ``response`` attribute of the received
+:class:`~twisted.python.failure.Failure` object.
+``fail`` is ``False`` by default, which means the response will be handled
+by the request callback. Note that ``fail`` is a keyword-only argument.
+
+Regardless of which method handles the response, it will contain
+``"download_stopped"`` in its :attr:`Response.flags` attribute.
```
kmike, line 99:
```suggestion
The ``fail`` boolean parameter indicates whether or not the resulting partial
response should be handled by the request callback (fail=False, default) or by
the request errback (fail=True).

* If ``StopDownload(fail=False)`` is raised, request callback is called. Response
  which is passed to the callback could have its body truncated: the body
  contains only bytes received before the StopDownload exception is raised.
* When ``StopDownload(fail=True)`` exception is raised, request errback is called. 
  In this case response (possibly with a partial body) is available as the 
  ``response`` attribute of the received :class:`~twisted.python.failure.Failure` 
  object.

Note that ``fail`` is a keyword-only argument.

Regardless of which method handles the response, the response object has
``"download_stopped"`` in its :attr:`Response.flags` attribute.
```

kmike: While writing the suggestion above, I realized we don't specify which data would response.body actually contain. In particular, would it include the current data chunk received in bytes_received signal?

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,25 @@ NotSupported
 
 This exception is raised to indicate an unsupported feature.
 
+StopDownload
+-------------
+
+.. versionadded:: 2.2
+
+.. exception:: StopDownload(fail=False)
```
kmike, line 86: What do you think about either the removing the default value, or making it True by default?

Argument for removing the default: `StopDownload(fail=False)` is easier to understand, and not much longer to write.

Argument for making fail=True by default: this feature can be seen as a much more flexible DOWNLOAD_MAXSIZE, and DOWNLOAD_MAXSIZE causes errback to be called.

docs/topics/request-response.rst
```diff
@@ -385,6 +385,52 @@ The meta key is used set retry times per request. When initialized, the
...
+        def parse(self, response):
+            # 'last_chars' show that the full response was not downloaded
+            yield {"len": len(response.text), "last_chars": response.text[-40:]}
+
+        def handler(self, data, request, spider):
+            some_condition = ...
+            if some_condition:
+                raise scrapy.exceptions.StopDownload()
+
+which produces the following output::
```
kmike, line 420: It doesn't produce this output, because we have `some_condition` now, instead of a real check :)

scrapy/core/scraper.py
```diff
@@ -147,6 +147,14 @@ def _scrape2(self, request_result, request, spider):
...
     def call_spider(self, result, request, spider):
         result.request = request
+        # StopDownload exceptions: make the partial response an attribute of the failure
+        if (
+            isinstance(result, Failure)
+            and isinstance(result.value, StopDownload)
+            and hasattr(result.value, "response")
+        ):
+            result.response = result.value.response
+            delattr(result.value, "response")
```
kmike, line 157: Why not `del result.value.response`?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -440,6 +441,14 @@ def _cb_bodydone(self, result, request, url):
...
+            raise failure
```
kmike, line 450: hm, I was not aware raising a Failure instances directly is what people do.. Could you please double-check that it doesn't cause any issues, as compared to e.g. calling a method on failure to raise an exception? Or maybe it is possible to return a failure instead of raising it? Sorry, I'm clueless :)

elacuesta: I _believe_ it is the same thing, but I'm not 100% sure so don't quote me on that. Tests are not affected by this change, but in any case, let's play it safe and return instead. I hoped it would have been the key to resolve https://github.com/scrapy/scrapy/pull/4559#discussion_r425334750, but sadly it is not.

scrapy/core/scraper.py
elacuesta: No reason in particular, I just didn't think of it. Do you think `del` is more readable than `dellattr`? I don't really have a strong opinion on this.

docs/topics/request-response.rst
elacuesta: Removed that, just raising the exception should be enough as an example.

docs/topics/exceptions.rst
elacuesta: Updated, thanks!

scrapy/core/scraper.py
kmike: I think the main difference is that delattr allows arbitrary string variables, while del needs a Python expression. So for me delattr here feels like using `setattr(foo, "bar", 5)` instead of `foo.bar = 5`. 

On the other hand, with `delattr(result.value, "response")` is is more clear that we're deleting the response attribute, while with `del result.value.response` it is not immediately clear what's being deleted.

So that's fine to keep it as-is as well :)

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,31 @@ NotSupported
...
+  :class:`~twisted.python.failure.Failure` object. This is the default behaviour.
+
+* If ``fail=False``, the request callback is called instead.
+
+Note that ``fail`` is a keyword-only parameter.
+
+In both cases, the response could have its body truncated: the body contains
+all bytes received up until the exception is raised, including the bytes
+received in the signal handler that raises the exception. The response object
+willl have ``"download_stopped"`` in its :attr:`Response.flags` attribute.
```
kmike, line 105:
```suggestion
received in the signal handler that raises the exception. Also, the response 
object is marked with ``"download_stopped"`` in its :attr:`Response.flags` 
attribute.
```

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,31 @@ NotSupported
...
+
+* If ``fail=True``, the request errback is called. The response object is
+  available as the ``response`` attribute of the received
+  :class:`~twisted.python.failure.Failure` object. This is the default behaviour.
+
+* If ``fail=False``, the request callback is called instead.
+
+Note that ``fail`` is a keyword-only parameter.
+
+In both cases, the response could have its body truncated: the body contains
```
kmike, line 102: What do you think about moving a note about kwarg-only argument somewhere else, to not separate 2 cases from "In both cases, ..."?

docs/topics/request-response.rst
```diff
@@ -385,6 +385,51 @@ The meta key is used set retry times per request. When initialized, the
...
+        def from_crawler(cls, crawler):
+            spider = super().from_crawler(crawler)
+            crawler.signals.connect(spider.handler, signal=scrapy.signals.bytes_received)
+            return spider
+
+        def parse(self, response):
+            # 'last_chars' show that the full response was not downloaded
+            yield {"len": len(response.text), "last_chars": response.text[-40:]}
+
+        def handler(self, data, request, spider):
```
kmike, line 415: another💄 
```suggestion
        def on_bytes_received(self, data, request, spider):
```

kmike: (it would require changes in from_crawler as well)

docs/topics/exceptions.rst
```diff
@@ -77,3 +78,34 @@ NotSupported
...
+
+* If ``fail=True``, the request errback is called. The response object is
+  available as the ``response`` attribute of the received
+  :class:`~twisted.python.failure.Failure` object. This is the default behaviour.
+
+* If ``fail=False``, the request callback is called instead.
+
+In both cases, the response could have its body truncated: the body contains
+all bytes received up until the exception is raised, including the bytes
+received in the signal handler that raises the exception. Also, the response
```
kmike, line 102: How hard would it be to add a test for it?

elacuesta: Done in fbe915d9cbd8ab801a679b05265cff416764d558

scrapy/core/scraper.py
```diff
@@ -147,6 +147,14 @@ def _scrape2(self, request_result, request, spider):
 
     def call_spider(self, result, request, spider):
         result.request = request
+        # StopDownload exceptions: make the partial response an attribute of the failure
+        if (
+            isinstance(result, Failure)
+            and isinstance(result.value, StopDownload)
+            and hasattr(result.value, "response")
+        ):
+            result.response = result.value.response
```
kmike, line 156: Actually I wonder if we should just keep failure.value.response as-is. This seems to be consistent with https://docs.scrapy.org/en/latest/topics/request-response.html#using-errbacks-to-catch-exceptions-in-request-processing example.

scrapy/core/downloader/handlers/http11.py
kmike: This seems to be consistent with https://docs.scrapy.org/en/latest/topics/request-response.html#using-errbacks-to-catch-exceptions-in-request-processing, so I wonder if we just document that the way to get a response in a callback is failure.value.response when StopException is received.

scrapy/core/scraper.py
elacuesta: Good catch. It would make this change simpler too.

elacuesta: Updated @kmike :tada:

kmike:
Thanks @elacuesta!

PR #4564, opened 2020-05-11
Unnecessary update when value is None

jacty:
When value is None, it is not necessary to invoke update and run other methods and conditions to make the code complicated there.

### DIFF
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -83,7 +83,8 @@ class BaseSettings(MutableMapping):
     def __init__(self, values=None, priority='project'):
         self.frozen = False
         self.attributes = {}
-        self.update(values, priority)
+        if values is not None:
+            self.update(values, priority)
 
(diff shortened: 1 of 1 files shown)


scrapy/settings/__init__.py
```diff
@@ -83,7 +83,8 @@ class BaseSettings(MutableMapping):
     def __init__(self, values=None, priority='project'):
         self.frozen = False
         self.attributes = {}
-        self.update(values, priority)
+        if values is not None:
```
Gallaecio, line 86: Nice catch.

Since we are at it, maybe it would make sense to just use `if values:`. I imagine empty settings do not require `self.update` either.

jacty: > Nice catch.
> 
> Since we are at it, maybe it would make sense to just use `if values:`. I imagine empty settings do not require `self.update` either.

Agree.

wRAR:
Thanks!

PR #4568, opened 2020-05-11
Remove _monkeypatches, import-related Flake8 codes

elacuesta:
No need for the monkey patch anymore, since https://twistedmatrix.com/trac/ticket/7989 does not affect Python 3 (similar to [this code simplification related to `pickle`](https://github.com/scrapy/scrapy/pull/4541/files#diff-7949c127769229f245753fec6e35d92cL31-L44)).

Refactored `scrapy/__init__.py` to group the different sections.

Also removed the E402 Flake8 code. Most occurrences in `pytest.ini` were outdated.

/cc @dangra since he authored #1414. Please advice about this removal. Thanks in advance.


scrapy/selector/__init__.py
```diff
@@ -1,4 +1,6 @@
 """
 Selectors
 """
-from scrapy.selector.unified import *  # noqa: F401
+
+# top-level imports
+from scrapy.selector.unified import Selector, SelectorList
```
elacuesta, line 6: This is safe, these two classes are the only ones that have been exposed from [the very beginning](https://github.com/scrapy/scrapy/blame/2.1.0/scrapy/selector/unified.py#L11).

kmike:
Thanks @elacuesta!

PR #4569, opened 2020-05-12
Cleaned up the process to get the closest cfg file

jacty:


Gallaecio:
The changes are backward-incompatible. I wonder if we could take a backward-compatible route instead. Maybe through new optional, keyword-only parameters and deprecating but still supporting the old parameters?

jacty:
> The changes are backward-incompatible. I wonder if we could take a backward-compatible route instead. Maybe through new optional, keyword-only parameters and deprecating but still supporting the old parameters?

uhm, I re-checked our source repository, these three methods which I have changed wont be invoked in other places. I think it won't have any backward-incompatible issues if I don't understand you wrong.

jacty:
btw, I don't know if we should move part or all the constants into an independent file to be imported to use when in need for example ENVVAR. Cuz I am too new to do this task, just give an opinion and hope it helps. Thanks.

Gallaecio:
> I re-checked our source repository, these three methods which I have changed wont be invoked in other places.

We are changing public methods which may be used by external projects using Scrapy. That means we may be breaking user code. If it can be helped, we should try and avoid that.

jacty:
> > I re-checked our source repository, these three methods which I have changed wont be invoked in other places.
> 
> We are changing public methods which may be used by external projects using Scrapy. That means we may be breaking user code. If it can be helped, we should try and avoid that.

uhm, sorry, I am not sure if I understand you right. If we are changing,  why cannot we changing this part in our current project before it is released? My initial to make these methods simple and clear. If we have to use multiple arguments(both key word or optional) to meet this initial, I feel it goes far away from my initial. Or perhaps I should leave this part for future discussion after our changing part released.

Gallaecio:
> why cannot we changing this part in our current project before it is released?

We can change its uses in the Scrapy code base, but we should not break projects using the old code.

> My intention to make these methods simple and clear. If we have to use multiple arguments(both key word or optional) to meet this intention, I feel it goes far away from my intention.

You would be making their signature and implementation more complex (more parameters, and issuing of deprecation warnings when using the old ones), but allowing users of those functions to use them the new way while not breaking existing code.

After some time, in some future version of Scrapy, the code and parameters which you would be adding for backward compatibility would be removed, leaving just the intended, cleaner implementation.

Whether the changes are worth the trouble, I leave up to you.

jacty:
uhm, okay. I understand your answer to the 2nd question but the answer to the 1st one. I will leave this change for future after I finished reading the whole source code of Scrapy and have a better understanding of how this project is managed and maintained between different versions. First time to try to contribute to an open source project. Leave me a bit time to understand. Thanks for your time anyway.

PR #4572, opened 2020-05-13
Flake8: remove E306 and F523

noviluni:
It seems that the last released version of `pytest-flake8` fixes the issue causing the Travis pipeline to fail. However, I've seen that it's raising some new cases.

### DIFF
--- a/scrapy/core/spidermw.py
+++ b/scrapy/core/spidermw.py
@@ -19,7 +19,7 @@ def _isiterable(possible_iterator):
 
 
 def _fname(f):
-    return "%s.%s".format(
+    return "{}.{}".format(
         f.__self__.__class__.__name__,
         f.__func__.__name__
     )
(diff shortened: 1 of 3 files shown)


elacuesta:
Hey @noviluni, great work on the upstream fix, thank you very much!
I'm working on removing E741 at #4558, it seems like these didn't appear before because they are within `for` loops. Do you think you can remove E741 additions in this PR? That would leave only four ocurrences of E306/F523, do you do you think you could edit the offending files instead?

noviluni:
sure! 👍  @elacuesta

noviluni:
Hi @elacuesta, could you take a look? Is this what you requested?

scrapy/core/spidermw.py
```diff
@@ -19,7 +19,7 @@ def _isiterable(possible_iterator):
 
 
 def _fname(f):
-    return "%s.%s".format(
```
elacuesta, line 22: I'm guilty of this one, until now I hadn't even realized about its oddity.

elacuesta:
Travis is failing with 429, perhaps we are triggering too many builds? :joy:

PR #4573, opened 2020-05-13
replace set constructor with set literals

noviluni:
I replaced the uses of the set constructor with set literals as the performance is slightly better.


elacuesta:
Interesting, I had no idea. Reading [this post](https://renzo.lucioni.xyz/pythons-set-literals/), it seems like it might be possible to convert other occurrences as well, using unpacking: `{*<iterable>}`.
Are there any occurrences outside of the tests?

PR #4574, opened 2020-05-13
Returns a JSON serializable Python object.

### LINKED ISSUE #2444: response.json()?

bulatbulat48:
- added property .json 
- added tests 

Fixes #2444, closes #4460.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -834,6 +834,10 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
+    .. automethod:: Response.json()
+
+        Returns a Python object from deserialized JSON document.
+        Result is cached after the first call.
 
 HtmlResponse objects
 --------------------
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -839,6 +839,7 @@ TextResponse objects
         Returns a Python object from deserialized JSON document.
         Result is cached after the first call.
 
+
 HtmlResponse objects
 --------------------
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -834,7 +834,7 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
-    .. automethod:: Response.json()
+    .. automethod:: TextResponse.json()
 
         Returns a Python object from deserialized JSON document.
         Result is cached after the first call.
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -837,7 +837,7 @@ TextResponse objects
     .. automethod:: TextResponse.json()
 
         Returns a Python object from deserialized JSON document.
-        Result is cached after the first call.
+        The result is cached after the first call.
 
 
 HtmlResponse objects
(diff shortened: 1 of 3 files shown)


elacuesta:
Hey @bulatbulat48, thanks for the PR. For the record, this was originally requested in #2444, and there is an open PR at #4460.

scrapy/http/response/text.py
```diff
@@ -68,6 +69,10 @@ def body_as_unicode(self):
                       ScrapyDeprecationWarning)
         return self.text
 
+    @property
+    def json(self):
+        return json.loads(self.text)
```
Gallaecio, line 74: - It should include a docstring, and you should add `.. autoattribute:: TextResponse.json` to `docs/topics/request-response.rst`.
- It’s not a blocker, but I would consider https://github.com/scrapy/scrapy/pull/4460’s caching approach including https://github.com/scrapy/scrapy/pull/4460#discussion_r401033472

bulatbulat48: Many thanks for the review. Fixed in https://github.com/scrapy/scrapy/pull/4574/commits/b17a007f3a3d48957fc08e51b4cec483f7494af5

tests/test_http_response.py
```diff
@@ -669,6 +669,21 @@ def test_body_as_unicode_deprecation_warning(self):
...
 
+    def test_json_response(self):
+        json_body = b"""{"ip": "109.187.217.200"}"""
+        json_response = self.response_class("http://www.example.com", body=json_body)
+        self.assertEqual(json_response.json(), {'ip': '109.187.217.200'})
+
+        json_body = b"""null"""
+        json_response = self.response_class("http://www.example.com", body=json_body)
+        _NONE = object()
+        self.assertNotEqual(json_response.json(), _NONE)
```
Gallaecio, line 680: I think this would pass even with the old implementation (`_NONE` here is not the same as `_NONE` in the implementation).

In any case, I believe what we need to test is that `json.loads` is only called once when you call `json_response.json()` twice. We probably need to use `mock` here.

bulatbulat48: @Gallaecio great suggestion, thank you! Added mock test. Please review.

docs/topics/request-response.rst
```diff
@@ -834,6 +834,11 @@ TextResponse objects
 
     .. automethod:: TextResponse.follow_all
 
+    .. automethod:: TextResponse.json()
+
+        Returns a Python object from deserialized JSON document.
+        Result is cached after the first call.
```
Gallaecio, line 840:
```suggestion
        The result is cached after the first call.
```

bulatbulat48: Fixed.

bulatbulat48:
@elacuesta, please review. Thanks!

scrapy/http/response/text.py
```diff
@@ -68,6 +72,15 @@ def body_as_unicode(self):
                       ScrapyDeprecationWarning)
         return self.text
 
+    def json(self):
+        """
+        Deserialize a JSON document to a Python object.
+        """
+        if self._cached_decoded_json is _NONE:
+            self._cached_decoded_json = json.loads(self.text)
+
```
elacuesta, line 81: I'd remove this line

bulatbulat48: @elacuesta thank you, good catch! Fixed. If possible, can we merge it?

PR #4576, opened 2020-05-14
Add 'item_classes' key to FEEDS

### LINKED ISSUE #4575: Add 'item_classes' key to FEEDS
## Summary

Add 'item_classes' key to FEEDS that accepts a list of scrapy.Item class objects.

## Motivation

This functionality will allow us to filter different Item types to unique URIs

Currently, there isn't any way to filter by Item type in the FEED(s) settings.

## Additional context

I've already written the patch, tests, and updated documentation, I just need to create a pull request but I wanted to log it first here.

longkyle:
Fixes #4575 

I think it would be nice to be able to filter which items get exported to each URI in FEEDS.

This update allows for an 'item_classes' key (optional) to be added to FEEDS. If 'item_classes' is not provided, it defaults to `None` which includes all item classes. This behavior is meant to mimic the behavior of the 'fields' key.

```
from my_project.items import ProductItem, AnotherItem

FEEDS = {
        'product_items.json': {
            'format': 'json',
            'item_classes': ProductItem                      <-- scrapy.Item object
        },
        'other_items.json': {
            'format': 'json',
            'item_classes': (ProductItem, AnotherItem)       <-- set of scrapy.Item objects
        },
        'all_items.json': {
            'format': 'json'
        }
    }
```


scrapy/exporters.py
```diff
@@ -52,6 +53,17 @@ def start_exporting(self):
     def finish_exporting(self):
         pass
 
+    def _is_export_item(self, item):
+        """Determine whether or not an item should be exported
+        based on the 'items' key in FEEDS
+        """
+        if self.items_to_export is None:
+            return True
+        elif type(item) in self.items_to_export:
```
elacuesta, line 62:
```suggestion
        elif isinstance(item, self.items_to_export):
```

scrapy/exporters.py
```diff
@@ -88,9 +100,10 @@ def __init__(self, file, **kwargs):
         self.encoder = ScrapyJSONEncoder(**self._kwargs)
 
     def export_item(self, item):
-        itemdict = dict(self._get_serialized_fields(item))
-        data = self.encoder.encode(itemdict) + '\n'
-        self.file.write(to_bytes(data, self.encoding))
+        if self._is_export_item(item):
+            itemdict = dict(self._get_serialized_fields(item))
+            data = self.encoder.encode(itemdict) + '\n'
+            self.file.write(to_bytes(data, self.encoding))
```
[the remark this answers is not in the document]
longkyle: I like that it reduces the diff but I'm not sure I agree that it's cleaner. It adds an extra line of code and is ever so slightly less readable to me... but to each their own.

It doesn't bother me at all to make this change. I just wanted to provide my opinion, fwiw.

tests/test_feedexport.py
```diff
@@ -753,32 +753,42 @@ def test_export_encoding(self):
 
     @defer.inlineCallbacks
     def test_export_multiple_configs(self):
-        items = [dict({'foo': u'FOO', 'bar': u'BAR'})]
+        items = [
```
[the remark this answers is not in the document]
longkyle: Sure thing!

scrapy/exporters.py
longkyle: Agreed.

longkyle:
@elacuesta Thanks for taking the time to review my code!
I think you're right about naming the key, `item_classes` thanks for the suggestion! I'll update with your other suggestions as well.

Thanks again!

scrapy/exporters.py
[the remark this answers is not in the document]
elacuesta: Fair enough, in any case it's not a blocker for me. Let's see if other maintainers have anything to say, if not let's go with it :+1:

scrapy/exporters.py
```diff
@@ -52,6 +53,17 @@ def start_exporting(self):
     def finish_exporting(self):
         pass
 
+    def _should_export_item(self, item):
+        """Determine whether or not an item should be exported
+        based on the 'items' key in FEEDS
```
elacuesta, line 58:
```suggestion
        based on the 'items_classes' key in the FEEDS setting
```

elacuesta:
Many thanks for the updates. Could you take a look at the Travis failures?

longkyle:
@elacuesta I've spent quite a bit of time looking into the travis errors and am not sure what's going on, unfortunately. The tests pass for me locally but seem to fail unpredictably in travis. Might you be able to take a quick look and just see if anything stands out? Basically, the order of fields seem to get saved in different orders depending on the test/environment?

Attached is an image showing that if I correct for the travis test, it fails locally, and vice versa.
Screen Shot 2020-05-14 at 6 26 37 PM

Gallaecio:
Yes, that’s because we still support Python 3.5, where the key order of dictionaries is not kept.
You could change the tests to check that the values are both `bytes`, then load them with `json.loads`, and _then_ check that they match.

elacuesta:
Indeed. Another thing you could do would be to use items with only one field, as the `test_export_multiple_configs` and `test_export_indentation` methods do, for instance.

Gallaecio:
To avoid having to add those 2 lines to every exporter, I think we should consider applying the option through `_FeedSlot` instead. See https://github.com/scrapy/scrapy/pull/4434.

There is a `slot.exporter.export_item` line in `feedexport`. You could implement a method of `_FeedSlot` that starts with those two lines that check the item class, and then call `slot.exporter.export_item`, and you could replace the current call to `slot.exporter.export_item` with a call to that new `slot.export_item` method.

scrapy/exporters.py
```diff
@@ -52,6 +53,17 @@ def start_exporting(self):
...
+    def _should_export_item(self, item):
+        """Determine whether or not an item should be exported
+        based on the 'items_classes' key in the FEEDS setting
+        """
+        if self.item_classes_to_export is None:
+            return True
+        elif isinstance(item, self.item_classes_to_export):
+            return True
+        else:
+            return False
```
elacuesta, line 65:
```suggestion
        return (
            self.item_classes_to_export is None
            or isinstance(item, self.item_classes_to_export)
        )
```
Shorter and clearer IMHO

elacuesta:
@longkyle Seeing the message for f97b521c4035f9db2393ba7393b081e3e3ed73d2, I'd like to mention that you can trigger a new build by closing and re-opening the PR. Additionally, I'd recommend you to run the tests locally to spot any failures before pushing, it's usually much more comfortable and fast than waiting for the Travis build. See [this section](https://docs.scrapy.org/en/latest/contributing.html#tests) for some guidelines on how to run tests.

elacuesta:
For the record, the last build error is unrelated to this PR :+1:

longkyle:
@elacuesta Thanks for the tips! I had to rerun one of them because travis-ci said that everything passed but still said "pending" on github. It hung there for a few hours. If that last build error is unrelated then we should be good to go. (although I pushed another commit right before I read your tips *face palm*)

Anyway, thanks!

elacuesta:
Implementation is looking great, but I think we need to cover this more extensively in the docs. It's a useful feature and I think it's a bit hidden, only mentioned once in the allowed keys for the `FEEDS` setting and without any context. Given that it has no fallback setting, I think a small paragraph after the `FEEDS` docs could be enough, with a `versionadded` directive (possibly 2.2 if we include it in the next release, but we can always update later if that's not the case).

docs/topics/feed-exports.rst
```diff
@@ -275,6 +275,38 @@ as a fallback value if that key is not provided for a specific feed definition.
 * ``indent``: falls back to :setting:`FEED_EXPORT_INDENT`
 * ``store_empty``: falls back to :setting:`FEED_STORE_EMPTY`
 
+item_classes
+------------
```
Gallaecio, line 279: I section at the same level as `FEEDS` is not a good way to add this.

I would suggest extending the list entry with this information. See https://github.com/scrapy/scrapy/pull/4512/files#diff-b08e74952c08d71c2ff7fd960c551b10 to get an idea of how you can include details in the list itself.

I would avoid such a long code example, though. I’m not sure we need a code example at all here. A statement indicating allowed values, like the one you have above the code, should be information enough.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +270,10 @@ as a fallback value if that key is not provided for a specific feed definition.
   See :ref:`topics-feed-format` for possible values. 
   Mandatory, no fallback setting
 * ``encoding``: falls back to :setting:`FEED_EXPORT_ENCODING`
+* ``item_classes``: no fallback setting, defaults to ``None`` (which will include all item classes)
+  - .. versionadded:: 2.2
+  - Allows for filtering which item class types get exported to each URI.
+  - Acceptable values include None, a single Item class object, or a tuple of Item class objects
```
Gallaecio, line 276: What about this?

```suggestion
* ``item_classes``: filter of exported items based on their item class.

  .. versionadded:: 2.2
   
  If the value is ``None`` (default), all items are exported. If the value is 
  an item class, or a tuple of item classes, only items that are instances of 
  those item classes are exported.
```

docs/topics/feed-exports.rst
```diff
@@ -245,12 +245,14 @@ For instance::
...
             'encoding': 'utf8',
-            'store_empty': False,
+            'item_classes': (ProductItemClass, AnotherItemClass)
             'fields': None,
             'indent': 4,
+            'store_empty': False,
         }, 
         '/home/user/documents/items.xml': {
             'format': 'xml',
+            'item_classes': ProductItemClass
```
Gallaecio, line 255:
```suggestion
            'item_classes': ProductItemClass,
```

docs/topics/feed-exports.rst
```diff
@@ -245,12 +245,14 @@ For instance::
         'items.json': {
             'format': 'json',
             'encoding': 'utf8',
-            'store_empty': False,
+            'item_classes': (ProductItemClass, AnotherItemClass)
```
Gallaecio, line 248:
```suggestion
            'item_classes': (ProductItemClass, AnotherItemClass),
```

StasDeep:
Hello! Just saw this and I like the idea, but I am now thinking maybe it makes more sense to have a filter behavior in the FeedExporter rather than in ItemExporter? It just sounds a bit off to me to make ItemExporter to decide whether an item should be exported or not.

scrapy/extensions/feedexport.py
```diff
@@ -288,7 +293,7 @@ def close_spider(self, spider):
     def item_scraped(self, item, spider):
         for slot in self.slots:
             slot.start_exporting()
-            slot.exporter.export_item(item)
+            slot.export_item(item)
             slot.itemcount += 1
```
StasDeep, line 297: This increases itemcount even when item is not exported.

Gallaecio:
Sounds good to me.

I don’t think we need to implement support for complex filters in this pull request, though, unless @longkyle wishes to do it. I think here we could simply replace `item_classes` with `filter`, and make the necessary changes to move the logic out of the item exporters into the feed exporter.

That would leave room for implementing support for a more complex filter in a later pull request, where we can extend this `filter` to, in addition of Item subclasses and sequences of them, allow for a filtering component (e.g. the import path of a class supporting `from_crawler` and `from_settings` and implementing a filtering method).

scrapy/extensions/feedexport.py
longkyle: Good catch!

scrapy/extensions/feedexport.py
```diff
@@ -288,8 +294,7 @@ def close_spider(self, spider):
     def item_scraped(self, item, spider):
         for slot in self.slots:
             slot.start_exporting()
```
kmike, line 296: I think this may start export without a need, if an item would be filtered out. 

Let's say you have a single FEEDS configuration in settings.py, which defines 2 export locations, split by 2 item classes. There are 2 spiders; one spider emits items of type A, the other emits items of type B. While running a spider, empty (content-wise) file can be created for the item type which spider doesn't emit.

psychonaute:
Works great 👍  however the FEED_STORE_EMPTY settings has no more effect. Feeds without any items in them still have their file created.

scrapy/extensions/feedexport.py
Gallaecio: I don’t see how we can avoid this. I think in this scenario `FEEDS` would need to be redefined on each spider, if creating an empty file is an issue.

kmike: Inside `slot.export_item()` (called in a next line) there is a `if self.exporter._should_export_item(item):` check. My point was that it could make sense to call `slot.start_exporting()` only after this check is successful at least once. It won't fix all scenarios where an empty file can appear, but, if I'm not mistaken, this should fix the problem I described.

ejulio:
@longkyle do you intend to keep the work here?
I'd add that we could rename `item_classes` to `match` and then it can be: a class, a tuple of classes, a matching function for any complex behavior from the user perspective.
Also if the previous comment about `FEED_STORE_EMPTY` holds, this is a regression that we need to fix and leave it explicit in the docs as it can break some stuff

longkyle:
@ejulio I was initially excited about contributing this feature to the community. It was approved and then just sat here without getting folded into the new releases. Unfortunately, I don't have the time to pick this back up at this point. If anybody else would like to pickup the torch, I'd be happy to see this feature get implemented!

PR #4588, opened 2020-05-20
Incompatible with pytest 5.4 and 5.4.1

altendky:


altendky:
This is following up on https://github.com/pytest-dev/pytest-twisted/issues/93.

elacuesta:
Thanks for the work in https://github.com/pytest-dev/pytest-twisted/issues/93 :rocket:
Should consider pinning `pytest>=5.4.2` instead?

altendky:
I did this because before it was just `pytest` so I was only banning the actually failing versions instead of introducing a minimum.  If this were a new feature rather than a temporary breakage in a couple versions then `>=` would make sense to me.

altendky:
So my only concern with going ahead with this is that there still might be something funny going on with the added pytest-twisted test when running with qt5reactor.  Might be nothing, might be irrelevant, I haven't spent any time on it yet.  Anyways, just wanted to mention that.  But hey, if more versions are found to be problematic they can be listed later.

kmike:
Thanks so much @altendky for making pytest-twisted working for us!

PR #4593, opened 2020-05-22
Fix the description of the ScreenshotPipeline example item pipeline

### LINKED ISSUE #4514: Outdated item pipeline example about Deferred
The [_Take screenshot of item_](https://docs.scrapy.org/en/latest/topics/item-pipeline.html#take-screenshot-of-item) example says "This example demonstrates how to return a Deferred from the process_item() method", but that is no longer the case (async/await syntax is currently used).
I think we should remove that sentence ~and add a reminder about [enabling the `asyncio` reactor](https://docs.scrapy.org/en/latest/topics/asyncio.html#installing-the-asyncio-reactor).~

Also, I wonder if it'd be good to also restore the previous `Deferred` example and keep both of them, since not all projects might be able/willing to switch to the `asyncio` reactor.

Gallaecio:
Fixes #4514

### DIFF
--- a/docs/topics/item-pipeline.rst
+++ b/docs/topics/item-pipeline.rst
@@ -167,11 +167,13 @@ method and how to clean up the resources properly.::
 Take screenshot of item
 -----------------------
 
-This example demonstrates how to return a
-:class:`~twisted.internet.defer.Deferred` from the :meth:`process_item` method.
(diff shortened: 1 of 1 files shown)


kmike:
Thanks @Gallaecio!

PR #4599, opened 2020-05-27
Fix KeyError in LocalWeakReferencedCache

### LINKED ISSUE #4597: KeyError in is_generator_with_return_value
```Python traceback
  File "/usr/local/lib/python3.8/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/app/python/lib/python3.8/site-packages/scrapy/core/downloader/middleware.py", line 42, in process_request
  File "/usr/local/lib/python3.8/site-packages/twisted/internet/defer.py", line 1362, in returnValue
    raise _DefGen_Return(val)
twisted.internet.defer._DefGen_Return: <200 https://www.example.com>

During handling of the above exception, another exception occurred:

  File "/app/python/lib/python3.8/site-packages/scrapy/utils/defer.py", line 55, in mustbe_deferred
    result = f(*args, **kw)
  File "/app/python/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 60, in process_spider_input
  File "/app/python/lib/python3.8/site-packages/scrapy/core/scraper.py", line 148, in call_spider
  File "/app/python/lib/python3.8/site-packages/scrapy/utils/misc.py", line 202, in warn_on_generator_with_return_value
    if is_generator_with_return_value(callable):
  File "/app/python/lib/python3.8/site-packages/scrapy/utils/misc.py", line 180, in is_generator_with_return_value
    return _generator_callbacks_cache[callable]
  File "/app/python/lib/python3.8/site-packages/scrapy/utils/datatypes.py", line 281, in __getitem__
    return super(LocalWeakReferencedCache, self).__getitem__(key)
  File "/usr/local/lib/python3.8/weakref.py", line 383, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x7f06ff011720; to 'method' at 0x7f07042b5e00 (parse_foo)>
```

This is Scrapy 2.0.1. The problem happens only sometimes, but in different spiders in the same project.

elacuesta:
Fixes #4597

`LocalWeakReferencedCache` uses a `LocalCache` object, which is size-limited, to store references. However, in this case the limit is set to `None`, so I'm not sure that's the issue. Another theory is that somehow all strong references to the key object, which is a spider callback, got deleted and so Python itself deleted the key (from the [upstream implementation](https://github.com/python/cpython/blob/v3.8.3/Lib/weakref.py#L332-L341), "_Entries in the dictionary will be discarded when there is no longer a strong reference to the key_"), but that wouldn't make much sense because the key does exist when trying to access it.

In any case, the key is not available so we are better off catching the `KeyError`.

`LocalWeakReferencedCache` is only used to warn in case of generators with non-None return values (#3869), which is definitely not critical so it shouldn't be causing exceptions like this.

Without catching the `KeyError`, the added test line fails like in the original issue:
```
================================================================================================================== FAILURES ==================================================================================================================
_____________________________________________________________________________________________ LocalWeakReferencedCacheTest.test_cache_with_limit _____________________________________________________________________________________________

self = <tests.test_utils_datatypes.LocalWeakReferencedCacheTest testMethod=test_cache_with_limit>

    def test_cache_with_limit(self):
        cache = LocalWeakReferencedCache(limit=2)
        r1 = Request('https://example.org')
        r2 = Request('https://example.com')
        r3 = Request('https://example.net')
        cache[r1] = 1
        cache[r2] = 2
        cache[r3] = 3
        self.assertEqual(len(cache), 2)
        self.assertNotIn(r1, cache)
        self.assertIn(r2, cache)
        self.assertIn(r3, cache)
>       self.assertEqual(cache[r1], None)

/.../scrapy/tests/test_utils_datatypes.py:274: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/.../scrapy/scrapy/utils/datatypes.py:107: in __getitem__
    return super(LocalWeakReferencedCache, self).__getitem__(key)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <LocalWeakReferencedCache at 0x7f6710a13b90>, key = <GET https://example.org>

    def __getitem__(self, key):
>       return self.data[ref(key)]
E       KeyError: <weakref at 0x7f67109ab4d0; to 'Request' at 0x7f6710a13b10>

/usr/local/lib/python3.7/weakref.py:396: KeyError
```

### DIFF
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -105,8 +105,8 @@ class LocalWeakReferencedCache(weakref.WeakKeyDictionary):
     def __getitem__(self, key):
         try:
             return super(LocalWeakReferencedCache, self).__getitem__(key)
-        except TypeError:
-            return None  # key is not weak-referenceable, it's not cached
+        except (TypeError, KeyError):
+            return None  # key is either not weak-referenceable or not cached
 
 
 class SequenceExclude:
--- a/tests/test_utils_datatypes.py
+++ b/tests/test_utils_datatypes.py
@@ -271,6 +271,7 @@ class LocalWeakReferencedCacheTest(unittest.TestCase):
         self.assertNotIn(r1, cache)
         self.assertIn(r2, cache)
         self.assertIn(r3, cache)
+        self.assertEqual(cache[r1], None)
         self.assertEqual(cache[r2], 2)
         self.assertEqual(cache[r3], 3)
         del r2


Gallaecio:
(not merging yet since https://travis-ci.org/github/scrapy/scrapy/builds/691774704 is still running)

PR #4603, opened 2020-05-31
Change URI verification in command line tools

### LINKED ISSUE #4530: scrapy command line tools fail on custom schemes
Commands like `scrapy fetch` force their argument to be validated through `w3lib.url.is_url()`.

However, as implemented [here](https://github.com/scrapy/w3lib/blob/165f1654a1de82cbaad170a20489527733c2f015/w3lib/url.py#L116), this function only accepts `file`, `https` and `http`.

This makes these command line tools unusable for S3 or FTP protocols, or other custom protocols that may be defined in `DOWNLOAD_HANDLERS`.

I think the command line tools should only check that it's a well-formed URI, regardless of the scheme.

Thanks!

michalp2213:
This closes #4530 . I implemented a new `is_uri` function in `scrapy.utils.url` that verifies if given text is a well-formed URI as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986) and replaced calls to `w3lib.url.is_url` in `fetch` and `parse` commands with it.

### DIFF
--- a/scrapy/commands/fetch.py
+++ b/scrapy/commands/fetch.py
@@ -1,11 +1,11 @@
 import sys
-from w3lib.url import is_url
 
 from scrapy.commands import ScrapyCommand
 from scrapy.http import Request
 from scrapy.exceptions import UsageError
 from scrapy.utils.datatypes import SequenceExclude
 from scrapy.utils.spider import spidercls_for_request, DefaultSpider
+from scrapy.utils.url import is_uri
 
 
 class Command(ScrapyCommand):
@@ -13,13 +13,13 @@ class Command(ScrapyCommand):
     requires_project = False
 
     def syntax(self):
-        return "[options] <url>"
+        return "[options] <uri>"
 
     def short_desc(self):
-        return "Fetch a URL using the Scrapy downloader"
+        return "Fetch a URI using the Scrapy downloader"
 
     def long_desc(self):
-        return "Fetch a URL using the Scrapy downloader and print its content " \
+        return "Fetch a URI using the Scrapy downloader and print its content " \
             "to stdout. You may want to use --nolog to disable logging"
 
     def add_options(self, parser):
@@ -47,7 +47,7 @@ class Command(ScrapyCommand):
         sys.stdout.buffer.write(bytes_ + b'\n')
 
     def run(self, args, opts):
-        if len(args) != 1 or not is_url(args[0]):
+        if len(args) != 1 or not is_uri(args[0]):
             raise UsageError()
         request = Request(args[0], callback=self._print_response,
                           cb_kwargs={"opts": opts}, dont_filter=True)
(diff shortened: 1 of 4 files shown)


michalp2213:
I added a check in `is_uri` that uri's scheme is on the list of supported protocols (passed as an argument, defaults to `{'http', 'https', 'file'}`) and modified `fetch` and `parse` to pass a set of protocols supported by current project (from `DOWNLOAD_HANDLERS_BASE` and `DOWNLOAD_HANDLERS`). If you want to modify `is_url` in w3lib, I can put these changes as a pull request there.

Also, I think it is probably a good idea to extract the code that produces a set of supported protocols in `fetch` and `parse` to a separate function. What do you think would be a good place for it? `utils.project`?

scrapy/commands/fetch.py
```diff
@@ -47,8 +48,15 @@ def _print_bytes(self, bytes_):
         sys.stdout.buffer.write(bytes_ + b'\n')
 
     def run(self, args, opts):
-        if len(args) != 1 or not is_url(args[0]):
+        if len(args) != 1:
             raise UsageError()
+
+        settings = get_project_settings()
+        supported_protocols = settings.getdict('DOWNLOAD_HANDLERS_BASE').keys()
+        supported_protocols |= settings.getdict('DOWNLOAD_HANDLERS').keys()
```
Gallaecio, line 56: See `settings.getwithbase`. Using that, extracting this code into a function is less important, as you could simplify the line below as:

```python
        if not is_uri(args[0], settings.getwithbase('DOWNLOAD_HANDLERS')):
```

scrapy/utils/url.py
```diff
@@ -133,3 +133,15 @@ def strip_url(url, strip_credentials=True, strip_default_port=True, origin_only=
         '' if origin_only else parsed_url.query,
         '' if strip_fragment else parsed_url.fragment
     ))
+
+
+def is_uri(text, protocols={'http', 'https', 'file'}):
```
Gallaecio, line 138: :lipstick: It may be better to make `protocols` keyword only:

```suggestion
def is_uri(text, *, protocols={'http', 'https', 'file'}):
```

Also, maybe it should be `schemes` instead of `protocols`.

scrapy/utils/url.py
```diff
@@ -133,3 +133,15 @@ def strip_url(url, strip_credentials=True, strip_default_port=True, origin_only=
...
+    regex = r'^(([a-zA-Z][-+.\w]*):)(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?'
+    pattern = re.compile(regex)
+    match = pattern.fullmatch(text)
```
Gallaecio, line 141: :lipstick: This would be simpler and the performance should be the same:

```suggestion
    match = re.fullmatch(pattern, text)
```

PR #4604, opened 2020-06-01
give write access to template files after copying with startproject

MMesch:
`scrapy startproject myproject` fails with permission denied if the original template permissions are read-only. This is the case on nix-based systems that ensure that installed libraries are immutable. I am not sure what the best place is to put these permissions but here is a quick fix.

### DIFF
--- a/scrapy/commands/startproject.py
+++ b/scrapy/commands/startproject.py
@@ -79,6 +79,7 @@ class Command(ScrapyCommand):
             else:
                 copy2(srcname, dstname)
         copystat(src, dst)
+        self._set_rw_permissions(dst)
 
     def _set_rw_permissions(self, path):
         """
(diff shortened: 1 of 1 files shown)


wRAR:
Doing it in `_copytree()` after `copystat` is better IMO.

MMesch:
Shouldn't the `_copytree` and `set_rw_permissions` functions go into `utils` or something? I think they are fairly general.

wRAR:
If they are used only in one place in Scrapy and don't look useful for writing spiders, then I'd say no.

MMesch:
I think this is fine now (from my side). Handing over to @wRAR if that's OK for you.

wRAR:
My first comment still applies.

MMesch:
There you go. Coding after someone else's taste, for someone else's code is a bit frustrating if it is done without "thank you" and voluntarily in one's free time. So please just adapt, merge or don't as you please. I think that's more efficient and little effort for you.

Thank you for your feedback.

PR #4605, opened 2020-06-01
Improve support for external request queues

Lukas0907:
This PR contains a specification for how request queues are expected to behave and some adjustments to make the code conform to that specification.

The main change is the introduction of the `TransientError` exception which indicates a temporary problem (e.g. connection error) that is likely to not be permanent. A fallback to the memory queue happens in such a case. Another change is that once a priority queue object is created, a downstream queue is also created as a self-check. This allows to fail early and consistently if there is a problem with the downstream queue.

This PR is related to #4326.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -229,6 +229,7 @@ Extending Scrapy
    topics/api
    topics/signals
    topics/exporters
+   topics/queues
 
 
 :doc:`topics/architecture`
@@ -251,6 +252,9 @@ Extending Scrapy
 
 :doc:`topics/exporters`
     Quickly export your scraped items to a file (XML, CSV, etc).
+
+:doc:`topics/queues`
+    Store requests in an external queue.
 
 
 All the rest
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -229,7 +229,7 @@ Extending Scrapy
    topics/api
    topics/signals
    topics/exporters
-   topics/queues
+   topics/request-queues
 
 
 :doc:`topics/architecture`
@@ -253,8 +253,8 @@ Extending Scrapy
 :doc:`topics/exporters`
     Quickly export your scraped items to a file (XML, CSV, etc).
 
-:doc:`topics/queues`
-    Store requests in an external queue.
+:doc:`topics/request-queues`
+    Customizing the queuing of requests.
 
 
 All the rest
--- a/docs/topics/queues.rst
+++ b/docs/topics/queues.rst
@@ -1,0 +1,83 @@
+.. _topics-queues:
+
+======
+Queues
+======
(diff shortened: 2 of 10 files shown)


docs/topics/queues.rst
```diff
@@ -0,0 +1,84 @@
...
+
+======
+Queues
+======
+
+Scrapy uses queues to schedule requests. By default, a memory-based queue is
+used but if :ref:`crawls should be paused and resumed <topics-jobs>` or more
+requests than fit in memory are scheduled, using an external queue (disk
+queue) is necessary. The setting :setting:`SCHEDULER_DISK_QUEUE` determines
+the type of disk queue that will be used by the scheduler.
```
Gallaecio, line 11: Even though you are creating this page with external queues in mind, this is the page where users interested in finding general information about Scrapy request queuing system will go. So I think we need to extend this documentation to cover questions they may have, even if they are unrelated to external queues.

For example:
- Why does Scrapy use multiple queues, and not a single one?
- How does Scrapy determine that additional requests do not fit into memory? How much margin of error is there? Is there a setting to adjust this? Where does Scrapy store requests in this scenario, where `JOBDIR` has not been defined?

Also:
- `JOBDIR` could be used just to save memory, even if no pausing and resuming is desired, so it’s probably better to mention the setting being used and not the pausing and resuming
- Maybe this page should be part of the “Extending Scrapy” section, and be linked from somewhere in the “Architecture overview” page.

docs/index.rst
```diff
@@ -252,6 +253,9 @@ Extending Scrapy
 :doc:`topics/exporters`
     Quickly export your scraped items to a file (XML, CSV, etc).
 
+:doc:`topics/queues`
+    Store requests in an external queue.
```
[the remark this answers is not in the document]
Lukas0907: “Customize the queuing of requests.” sounds very good to me, thank you.

docs/topics/queues.rst
Gallaecio: > > Why does Scrapy use multiple queues, and not a single one?
> 
> Do you mean memory vs. disk queue here, or priority queues?

I meant priorities (you already covered the disk vs memory aspect).

> > Maybe this page should be part of the “Extending Scrapy” section
> 
> It's already part of the `Extending scrapy` section in `index.rst`. Or do you mean something else?

Oops, my bad :facepalm:. For some reason, after reading the reference to the topic covering JOBDIR, I just assumed it was placed after that topic. Forget that point.

Lukas0907: @Gallaecio I extended the documentation with information about priority queues and refined the part about downstream queues. Maybe you want to take a look.

Issue #4547, opened 2020-05-07
Review unicode references


Gallaecio:
The documentation still has some references to Python 2’s `unicode`. We need to review them, and update them as needed (usually changing them to `str` or “string”).

divyabhushan:
Hi @Gallaecio, Can I work on this issue, please?

Gallaecio:
All yours!

divyabhushan:
@Gallaecio 
So to get started, I could find one python code file under docs - scrapy/docs/utils/linkfix.py with python 2's unicode style.
So I would use the function to convert it to a string type which I have already tested and verified in my environment. 
Is this the expected solution? Kindly verify.
Thanks.

elacuesta:
@divyabhushan I _believe_ what @Gallaecio means is refactoring the docs themselves. For instance, take a look at the description for [`Request.body`](https://docs.scrapy.org/en/2.1/topics/request-response.html#scrapy.http.Request):

> body (`str` or `unicode`) – the request body. If a `unicode` is passed, then it’s encoded to `str` using the encoding passed (which defaults to "utf-8"). If `body` is not given, an empty string is stored. Regardless of the type of this argument, the final value stored will be a `str` (never `unicode` or `None`).

which should be something like the following, now that we only support Python 3:

> body (`bytes` or `str`) – the request body. If a `str` is passed, then it's encoded to `bytes` using the passed encoding (which defaults to "utf-8"). If `body` is not given, an empty `bytes` object is stored. Regardless of the type of this argument, the final stored value will be an instance of `bytes` (never `str` or `None`).

Gallaecio:
`Request.body` is always `bytes`, but yes, that’s what I meant.

The changes needed on the `request-response.rst` file can be copied from [#4090](https://github.com/scrapy/scrapy/pull/4090/files#diff-37334d46f8fd4663fa288ccf25fd0e3f) (in case this pull request is merged first).

elacuesta:
> `Request.body` is always `bytes`, but yes, that’s what I meant.

For the record, you can in fact instantiate a `Request` with a `str` body, which gets encoded:
```python
In [1]: from scrapy import Request

In [2]: r = Request("http://example.org", body="hello!")

In [3]: type(r.body)
Out[3]: bytes

In [4]: r.body
Out[4]: b'hello!'
```

divyabhushan:
Thanks, @Gallaecio @elacuesta I make the necessary changes.

PR #4610, opened 2020-06-02
Add H2ClientProtocol

adityaa30:


scrapy/core/http2/protocol.py
```diff
@@ -66,22 +47,45 @@ def __init__(self):
         self.is_connection_made = False
         self._pending_request_stream_pool = []
 
+    def _stream_close_cb(self, stream_id: int):
+        """Called when stream is closed completely
+        """
+        try:
+            del self.streams[stream_id]
+        except KeyError:
+            pass
```
Gallaecio, line 56: https://stackoverflow.com/a/15411146/939364

scrapy/core/http2/protocol.py
```diff
@@ -66,22 +47,45 @@ def __init__(self):
...
-        stream = Stream(self.next_stream_id, request, self)
+        stream_id = self.next_stream_id
         self.next_stream_id += 2
 
+        stream = Stream(
+            stream_id=stream_id,
+            request=request,
+            connection=self.conn,
+            write_to_transport=self._write_to_transport,
+            cb_close=lambda: self._stream_close_cb(stream_id)
```
Gallaecio, line 69: This may be problematic, when using lambdas `stream_id` may not be what you expect by the time the lambda is executed. See https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

Since the caller of `cb_close` will know the stream ID, you may set `cb_close=self._stream_close_cb` instead and have `Stream` call `self.cb_close(self.stream_id)`.

scrapy/core/http2/protocol.py
```diff
@@ -100,31 +104,34 @@ def connectionMade(self):
...
         """
-        for stream_id in self.streams.keys():
-            # TODO: Close each Stream instance in a clean manner
-            self.conn.end_stream(stream_id)
+        # Pop all streams which were pending and were not yet started
+        for stream_id in list(self.streams):
+            try:
+                self.streams[stream_id].lost_connection()
+            except KeyError:
+                pass
```
Gallaecio, line 130: It looks like you could use `self.streams.values()` instead.

I’m not sure why `KeyError` is needed, unless `list_connection()` may raise it.

scrapy/core/http2/stream.py
```diff
@@ -59,26 +93,124 @@ def initiate_request(self):
...
+
+        # Next, check what the maximum frame size is.
+        max_frame_size = self._conn.max_outbound_frame_size
+
+        # We will send no more than the window size or the remaining file size
+        # of data in this call, whichever is smaller.
+        bytes_to_send = min(window_size, self.remaining_content_length)
+
+        # We now need to send a number of data frames.
+        while bytes_to_send > 0:
```
Gallaecio, line 133: :lipstick:  `bytes_to_send_size` would be less ambiguous, more consistent with similar variables in this function, and it would allow you to make this change without making the code more confusing:

```suggestion
        while bytes_to_send_size:
```

scrapy/core/http2/stream.py
```diff
@@ -59,26 +93,124 @@ def initiate_request(self):
...
+
+            data_chunk_start = self.content_length - self.remaining_content_length
+            data_chunk = self._request_body[data_chunk_start:data_chunk_start + chunk_size]
 
-    def receive_data(self, data: bytes):
+            self._conn.send_data(self.stream_id, data_chunk, end_stream=False)
+            self._write_to_transport()
+
+            bytes_to_send = max(0, bytes_to_send - chunk_size)
+            self.remaining_content_length = max(0, self.remaining_content_length - chunk_size)
```
Gallaecio, line 143: Given the prior lines:

```
bytes_to_send = min(window_size, self.remaining_content_length)
# …
chunk_size = min(bytes_to_send, max_frame_size)
```

I _think_ it would be safe to assume that the subtraction value will be ≥0.

scrapy/core/http2/protocol.py
adityaa30: >  looks like you could use `self.streams.values()` instead.

Yup I tried using `self.stream.values()` at first but that gave me "RuntimeError: Dictionary changed size during iteration" which was because of `lost_connection()` which calls `self._stream_close_cb` which pops the element from the dictionary. I then tried [this solution](https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error) which worked. 

> I’m not sure why `KeyError` is needed, unless `list_connection()` may raise it.

Just removed it and tested -- Works 🙃

scrapy/core/http2/stream.py
adityaa30: > I think it would be safe to assume that the subtraction value will be ≥0.

It's valid only if `chunk_size` is an integer multiple of both `bytes_to_send` and `self.remaining_content_length`. I removed the `min` though and changed to [this](https://github.com/scrapy/scrapy/pull/4610/commits/01ad8b31ab7fe86fd78a70b09a6dd61a497e0ccb#diff-f9b3d5f8359fe602b12d142e3ed2ace6R143-R148) in 01ad8b31ab7fe86fd78a70b09a6dd61a497e0ccb.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,183 @@
...
+        self._handle_events(events)
+        self._write_to_transport()
+
+    def connectionLost(self, reason=connectionDone):
+        """Called by Twisted when the transport connection is lost.
+        No need to write anything to transport here.
+        """
+        # Pop all streams which were pending and were not yet started
+        for stream_id in list(self.streams):
+            self.streams[stream_id].lost_connection()
```
Gallaecio, line 124: I must say things seem a bit overcomplicated. It is weird that `protocol.streams[n].lost_connection()` pops `protocol.streams[n]` (through a callback).

Any reason not to make it simpler?

```suggestion
        for stream in self.streams.values():
            stream.stream_closed_server = True
        self.streams = {}
```

Given the protocol is the one that calls `reset`, `lost_connection` and `end_stream`, the callback approach seems unnecessary, at least given the current code.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Streams are stored in a dictionary keyed off their stream IDs
+        self.streams = {}
+
+        # Boolean to keep track the connection is made
+        # If requests are received before connection is made
+        # we keep all requests in a pool and send them as the connection
+        # is made
+        self.is_connection_made = False
+        self._pending_request_stream_pool = []
```
Gallaecio, line 48: Have you considered using a [deque](https://docs.python.org/3.8/library/collections.html#collections.deque) here?

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,183 @@
...
+        self.conn = H2Connection(config=config)
+
+        # Address of the server we are connected to
+        # these are updated when connection is successfully made
+        self.destination = None
+
+        # ID of the next request stream
+        # Following the convention made by hyper-h2 each client ID
+        # will be odd.
+        self.next_stream_id = 1
```
Gallaecio, line 38: If you use [itertools.count](https://docs.python.org/3.8/library/itertools.html#itertools.count) as a value here, you can get new stream IDs a bit simpler:

```
stream_id = next(self.stream_id_count)
```

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,183 @@
...
+        """Instantiates a new Stream object
+        """
+        stream_id = self.next_stream_id
+        self.next_stream_id += 2
+
+        stream = Stream(
+            stream_id=stream_id,
+            request=request,
+            connection=self.conn,
+            write_to_transport=self._write_to_transport,
```
Gallaecio, line 65: Any reason to use this callback instead of giving the stream direct access to `self.transport`? It looks like `self.conn` and `self.transport` is all the stream needs from the protocol to write into the transport, and you are already passing `self.conn`.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,183 @@
...
+        stream_id = event.stream_id
+        self.streams[stream_id].receive_headers(event.headers)
+
+    def stream_ended(self, event: StreamEnded):
+        stream_id = event.stream_id
+        self.streams[stream_id].end_stream()
+
+    def stream_reset(self, event: StreamReset):
+        # TODO: event.stream_id was abruptly closed
+        #  Q. What should be the response? (Failure/Partial/???)
```
Gallaecio, line 173: In HTTP 1.1 the behavior [depends on the `DOWNLOAD_FAIL_ON_DATALOSS` setting](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/handlers/http11.py#L554).

scrapy/core/http2/protocol.py
adityaa30: > Any reason not to make it simpler?

One of the reasons was I wanted to make code more readable and was planning to handle cases where in functins `lost_connection` and `reset` we have to fire the reponse deferred errback. Similarly for `end_stream` we have to fire response deferred callback with the response instance

scrapy/core/http2/protocol.py
adityaa30: Done :+1:

scrapy/core/http2/protocol.py
adityaa30: Yeah so the only of `self.transport` in `Stream` instance would be to write data into transport. So, instead of creating the same function I thought maybe passing the the `HTTP2ClientProtocol` would be better. Also, in case there are some edge cases we can directly change `HTTP2ClientProtocol._write_to_transport` to reflect changes everywhere

scrapy/core/http2/protocol.py
```diff
@@ -72,11 +73,10 @@ def _new_stream(self, request: Request):
     def _send_pending_requests(self):
         # TODO: handle MAX_CONCURRENT_STREAMS
         # Initiate all pending requests
-        for stream in self._pending_request_stream_pool:
+        while len(self._pending_request_stream_pool):
```
Gallaecio, line 76: :lipstick: 

```suggestion
        while self._pending_request_stream_pool:
```

scrapy/core/http2/stream.py
```diff
@@ -163,7 +163,7 @@ def receive_window_update(self, delta):
         Arguments:
             delta -- Window change delta
         """
-        if self.remaining_content_length > 0 and not self.stream_closed_local:
+        if self.stream_closed_local is False:
```
Gallaecio, line 166: :lipstick: 

```suggestion
        if not self.stream_closed_local:
```

Unless it can be something else (`None`), which would be a bit confusing.

scrapy/core/http2/stream.py
```diff
@@ -58,20 +68,30 @@ def __init__(
...
+
+            # Headers received after sending the request
+            'headers': Headers({})
+        }
 
         # TODO: Add canceller for the Deferred below
         self._deferred_response = Deferred()
 
+    def __str__(self):
+        return "Stream(id={})".format(self.stream_id)
```
Gallaecio, line 120: :lipstick: 

```suggestion
        return "Stream(id={})".format(repr(self.stream_id))
```

scrapy/core/http2/stream.py
```diff
@@ -178,47 +220,69 @@ def receive_data(self, data: bytes, flow_controlled_length: int):
...
+
+        event: StreamReset
+        Stream reset via RST_FRAME by the upstream hence forcefully close
+        this stream and send TODO: ?
+
+        event: None
+        No event is launched -- Hence we will simply close this stream
         """
+        # TODO: In case of abruptly stream close
+        #  Q1. Do we need to send the request again?
```
Gallaecio, line 303: Sending requests again is something we want to be able to control from Scrapy, so that should definitely not be done here.

scrapy/core/http2/stream.py
```diff
@@ -178,47 +220,69 @@ def receive_data(self, data: bytes, flow_controlled_length: int):
...
+        #  Q2. What response should we send now?
```
Gallaecio, line 304: None, I think. It probably makes sense to call the request errback at this point. See https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/handlers/http11.py#L542

setup.py
```diff
@@ -69,6 +68,7 @@ def has_environment_marker_platform_impl_support():
     python_requires='>=3.5.2',
     install_requires=[
         'Twisted>=17.9.0',
+        'Twisted[http2]>=17.9.0'
```
Gallaecio, line 71: Shouldn’t you replace the existing entry instead of just adding a new one?

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,191 @@
...
+    DataReceived, ResponseReceived, SettingsAcknowledged,
+    StreamEnded, StreamReset, WindowUpdated
+)
+from twisted.internet.protocol import connectionDone, Protocol
+from twisted.internet.ssl import Certificate
+
+from scrapy.core.http2.stream import Stream, StreamCloseReason
+from scrapy.http import Request
+
+LOGGER = logging.getLogger(__name__)
```
elacuesta, line 18: :lipstick: 
```suggestion

logger = logging.getLogger(__name__)
```

Other occurrences of the logger would need be updated too.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,191 @@
...
+        """ Write data to the underlying transport connection
+        from the HTTP2 connection instance if any
+        """
+        data = self.conn.data_to_send()
+        self.transport.write(data)
+
+        LOGGER.debug("Sent {} bytes to {} via transport".format(len(data), self._metadata['ip_address']))
+
+    def request(self, _request: Request):
+        stream = self._new_stream(_request)
```
elacuesta, line 95:
```suggestion
    def request(self, request: Request):
        stream = self._new_stream(request)
```

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,191 @@
...
+        self._handle_events(events)
+        self._write_to_transport()
+
+    def connectionLost(self, reason=connectionDone):
+        """Called by Twisted when the transport connection is lost.
+        No need to write anything to transport here.
+        """
+        # Pop all streams which were pending and were not yet started
+        for stream_id in list(self.streams):
+            self.streams[stream_id].close(StreamCloseReason.CONNECTION_LOST, reason)
```
elacuesta, line 132:
```suggestion
        for stream in self.streams.values():
            stream.close(StreamCloseReason.CONNECTION_LOST, reason)
```
Would "close" be more accurate than "pop" in the above comment?

elacuesta, line 132: The names or the arguments seem a little bit confusing to me, in the sense that `close`'s first parameter is `reason`, but here a variable called "reason" is passed as `failure`. Not a big deal, but perhaps it can be improved somehow.

scrapy/core/http2/stream.py
```diff
@@ -0,0 +1,384 @@
...
+
+        # Stream was abruptly ended here
+        elif reason is StreamCloseReason.CANCELLED:
+            # Client has cancelled the request. Remove all the data
+            # received and fire the response deferred with no flags set
+            self._response['body'].truncate(0)
+            self._response['headers'].clear()
+            self._fire_response_deferred()
+
+        elif reason in (StreamCloseReason.RESET | StreamCloseReason.CONNECTION_LOST):
```
elacuesta, line 337:
```suggestion
        elif reason in (StreamCloseReason.RESET, StreamCloseReason.CONNECTION_LOST):
```
Is there a reason to be using a bitwise or? I found it a bit confusing, I had to look into the docs to understand the subtle fact that [`enum.IntFlag`](https://docs.python.org/3/library/enum.html#enum.IntFlag) handles this (`1 in (1 | 2)` explodes as expected). I believe doing a "regular" tuple check is simpler and more intuitive. It would also allow to define `StreamCloseReason` as a subclass of `enum.Enum` instead.
Let me know if I'm missing something.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,191 @@
...
+                self.window_updated(event)
+            elif isinstance(event, SettingsAcknowledged):
+                self.settings_acknowledged(event)
+            else:
+                LOGGER.info("Received unhandled event {}".format(event))
+
+    # Event handler functions starts here
+    def data_received(self, event: DataReceived):
+        stream_id = event.stream_id
+        self.streams[stream_id].receive_data(event.data, event.flow_controlled_length)
```
elacuesta, line 166: I think it'd be simpler to do what you're doing in `stream_reset`, i.e. `self.streams[event.stream_id]`. The same in other methods for this class.

scrapy/core/http2/stream.py
adityaa30: You're right. I didn't get the idea of using a `in` tuple check earlier so went with `IntFlag`. Changed definition to
```python
class StreamCloseReason(Enum):
```

scrapy/core/http2/protocol.py
adityaa30: Earlier I was going through what `reason` actually can be used for. I realized that `reason` represents the whole connection which works in HTTP/1.x as it has one request per connection. For HTTP/2 I used it just in case when `connectionLost` is called and we have some streams left then we can pass the `reason: Failure` instance to the errback. Any suggestions on this? 

>  but perhaps it can be improved somehow

Since `reason` is instance of `twisted.python.failure.Failure` I made the argument name of `Stream.close()` as `failure` 😁. Changing the argument name of `reason` in `connectionLost` won't be appropriate here ig?

adityaa30: > Would "close" be more accurate than "pop" in the above comment?

I did not get you?

setup.py
```diff
@@ -68,7 +67,7 @@ def has_environment_marker_platform_impl_support():
     ],
     python_requires='>=3.5.2',
     install_requires=[
-        'Twisted>=17.9.0',
+        'Twisted[http2]>=17.9.0'
```
Gallaecio, line 70:
```suggestion
        'Twisted[http2]>=17.9.0',
```

scrapy/core/http2/protocol.py
Gallaecio: @elacuesta Some background: https://github.com/scrapy/scrapy/pull/4610#discussion_r440006977

elacuesta: Oh right, I remember that. Makes sense, and the comment in https://github.com/scrapy/scrapy/commit/23906b6bee953d9bc5dd8042e785711b11840797#diff-a9927075d70e58a912f1e78ff3251186R143-R145 helps, but I'm not sure I like it. I will suggest a change in that thread.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,204 @@
...
+    def stream_ended(self, event: StreamEnded):
+        self.streams[event.stream_id].close(StreamCloseReason.ENDED)
+
+    def stream_reset(self, event: StreamReset):
+        self.streams[event.stream_id].close(StreamCloseReason.RESET)
+
+    def window_updated(self, event: WindowUpdated):
+        stream_id = event.stream_id
+        if stream_id != 0:
+            self.streams[stream_id].receive_window_update()
```
elacuesta, line 200:
```suggestion
        if event.stream_id != 0:
            self.streams[event.stream_id].receive_window_update()
```

A small simplification.

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,204 @@
...
+    def connectionLost(self, reason=connectionDone):
+        """Called by Twisted when the transport connection is lost.
+        No need to write anything to transport here.
+        """
+        # Pop all streams which were pending and were not yet started
+        # NOTE: Stream.close() pops the element from the streams dictionary
+        # which raises `RuntimeError: dictionary changed size during iteration`
+        # Hence, we copy the streams into a list.
+        for stream in list(self.streams.values()):
+            stream.close(StreamCloseReason.CONNECTION_LOST, self._protocol_error)
```
elacuesta, line 147: @adityaa30 @Gallaecio I still don't like the fact that a Stream ends up being responsible from removing itself from streams dict, and that this implementation details demands the need to create a new structure here. What about this?
```suggestion
        for stream in self.streams.values():
            stream.close(StreamCloseReason.CONNECTION_LOST, self._protocol_error)
        self.streams.clear()
```
Since all streams will be popped anyway.

```

Comment (adityaa30) [reply scrapy/core/http2/protocol.py]: This looks good. I have one doubt -- the `Stream` class under differnet conditions closes itself. Eg. the client cancels the request by calling `response_defer.cancel()`. For such cases what should we do? :thinking:

File: scrapy/core/http2/types.py [hunk 25]
```diff
@@ -0,0 +1,30 @@
...
+
+from scrapy.http.headers import Headers
+
+
+class H2ConnectionMetadataDict(TypedDict):
+    """Some meta data of this connection
+    initialized when connection is successfully made
+    """
+    certificate: Union[None, Certificate]
+    ip_address: Union[None, IPv4Address, IPv6Address]
```
elacuesta, line 17:
```suggestion
    certificate: Optional[Certificate]
    ip_address: Optional[Union[IPv4Address, IPv6Address]]
```

setup.py
```diff
@@ -1,8 +1,8 @@
 from os.path import dirname, join
+
 from pkg_resources import parse_version
 from setuptools import setup, find_packages, __version__ as setuptools_version
 
-
```
elacuesta, line 5: Nitpick: these two white lines should be left unmodified

setup.py
```diff
@@ -1,8 +1,8 @@
 from os.path import dirname, join
```
elacuesta, line 1: Nitpick: the white lines in this file should be left unmodified

scrapy/core/http2/protocol.py
elacuesta: Interesting. In that case, perhaps the stream could hold a reference to its parent protocol. It already stores a `H2ConnectionMetadataDict` with metadata, it could store the protocol itself instead of it.

scrapy/core/http2/stream.py
```diff
@@ -253,14 +255,9 @@ def receive_data(self, data: bytes, flow_controlled_length: int):
...
             self._reached_warnsize = True
-            warning_msg = 'Received more ({bytes}) bytes than download ' \
-                          + 'warn size ({warnsize}) in request {request}'
-            warning_args = {
-                'bytes': self._response['flow_controlled_size'],
-                'warnsize': self._download_warnsize,
-                'request': self._request
-            }
-            logger.warning(warning_msg.format(**warning_args))
+            warning_msg = f"Received more ({self._response['flow_controlled_size']}) bytes than download " \
```
Gallaecio, line 258: In Scrapy, we tend to prefer

```python
variable = ("string over "
            "multiple lines")
```

or, specially with longer strings,

```python
variable = (
    "string over "
    "multiple lines"
)
```

over

```python
variable = "string over " \
           + "multiple lines"
```

elacuesta:
Here's a PR to refactor the closing and removal of the streams: https://github.com/adityaa30/scrapy/pull/1

scrapy/utils/log.py
```diff
@@ -52,6 +52,9 @@ def filter(self, record):
         'twisted': {
             'level': 'ERROR',
         },
+        'hpack': {
```
Gallaecio, line 55: :lipstick:  What about keeping this alphabetical? hpack, scrapy, twisted.

tox.ini
```diff
@@ -84,6 +84,7 @@ deps =
     # Extras
     botocore==1.3.23
     Pillow==3.4.2
+    h2==3.2.0
```
Gallaecio, line 87: :lipstick: 

```suggestion
    h2==3.2.0
    botocore==1.3.23
    Pillow==3.4.2
```

adityaa30: Just out of curiosity -- What's the scheme that we should follow here for ordering?

Gallaecio: Oops… That was meant to be alphabetical :facepalm:

scrapy/core/http2/stream.py
```diff
@@ -0,0 +1,422 @@
...
+            # immediately when the stream needs to be cancelled
+
+            # There maybe no :status in headers, we make
+            # HTTP Status Code: 499 - Client Closed Request
+            self._response['headers'][':status'] = '499'
+            self._fire_response_deferred()
+
+        elif reason in (StreamCloseReason.RESET, StreamCloseReason.CONNECTION_LOST):
+            self._deferred_response.errback(ResponseFailed([
+                error if error else Failure()
```
elacuesta, line 386: Is this correct? `error` is annotated as `Exception` in the function definition.

scrapy/core/http2/stream.py
```diff
@@ -0,0 +1,422 @@
...
+        body = self._response['body'].getvalue()
+        response_cls = responsetypes.from_args(
+            headers=self._response['headers'],
+            url=self._request.url,
+            body=body
+        )
+
+        response = response_cls(
+            url=self._request.url,
+            status=self._response['headers'][':status'],
```
elacuesta, line 413: Should this be cast to `int`? IIRC all headers are strings, and above there is a line that does `self._response['headers'][':status'] = '499'`

adityaa30: Right. I'll add a typecast to `int` explicitly here. I did not do that initially because the `Response` already typecast's `status` to `int`.

https://github.com/scrapy/scrapy/blob/56a6d22352071208637abd86bd7b6d81a50d5bdb/scrapy/http/response/__init__.py#L23

scrapy/core/http2/stream.py
```diff
@@ -16,6 +16,7 @@
 from scrapy.http import Request
 from scrapy.http.headers import Headers
 from scrapy.responsetypes import responsetypes
+from scrapy.utils.python import to_unicode
```
Gallaecio, line 19: `str` should work, see https://github.com/scrapy/scrapy/issues/4547

elacuesta: Makes sense to stop using `to_unicode`, but please be [mindful of the encoding](https://github.com/scrapy/scrapy/pull/4661#issuecomment-654901531). It might not be an issue in these cases, but it's something to think about IMHO.

scrapy/core/http2/stream.py
```diff
@@ -83,7 +83,9 @@ def __init__(
         self,
         stream_id: int,
         request: Request,
-        protocol: "H2ClientProtocol"
+        protocol: "H2ClientProtocol",
+        download_maxsize: int = 0,
+        download_warnsize: int = 0
```
Gallaecio, line 88: :lipstick:

```suggestion
        download_warnsize: int = 0,
```

scrapy/core/downloader/contextfactory.py
```diff
@@ -1,8 +1,12 @@
...
+import warnings
+
 from twisted.internet.ssl import optionsForClientTLS, CertificateOptions, platformTrust, AcceptableCiphers
 from twisted.web.client import BrowserLikePolicyForHTTPS
 from twisted.web.iweb import IPolicyForHTTPS
 from zope.interface.declarations import implementer
+from scrapy.core.downloader.tls import openssl_methods
+from scrapy.utils.misc import create_instance, load_object
 
 from scrapy.core.downloader.tls import ScrapyClientTLSOptions, DEFAULT_CIPHERS
```
Gallaecio, line 11:
```suggestion
import warnings

from OpenSSL import SSL
from twisted.internet.ssl import optionsForClientTLS, CertificateOptions, platformTrust, AcceptableCiphers
from twisted.web.client import BrowserLikePolicyForHTTPS
from twisted.web.iweb import IPolicyForHTTPS
from zope.interface.declarations import implementer

from scrapy.core.downloader.tls import DEFAULT_CIPHERS, openssl_methods, ScrapyClientTLSOptions
from scrapy.utils.misc import create_instance, load_object
```

scrapy/core/downloader/handlers/http2.py
```diff
@@ -0,0 +1,142 @@
...
+        self._crawler = crawler
+
+    def _get_agent(self, request: Request, timeout: Optional[float]) -> H2Agent:
+        from twisted.internet import reactor
+        bind_address = request.meta.get('bindaddress') or self._bind_address
+        proxy = request.meta.get('proxy')
+        if proxy:
+            _, _, proxy_host, proxy_port, proxy_params = _parse(proxy)
+            scheme = _parse(request.url)[0]
+            proxy_host = str(proxy_host, 'utf-8')
```
Gallaecio, line 86: :lipstick: 

```suggestion
            proxy_host = proxy_host.decode()
```

scrapy/core/downloader/handlers/http2.py
```diff
@@ -0,0 +1,142 @@
...
+                              "If you use Crawlera, it doesn't require this mode anymore, "
+                              "so you should update scrapy-crawlera to 1.3.0+ "
+                              "and remove '?noconnect' from the Crawlera URL.",
+                              ScrapyDeprecationWarning)
+
+            if scheme == b'https' and not omit_connect_timeout:
+                proxy_auth = request.headers.get(b'Proxy-Authorization', None)
+                proxy_conf = (proxy_host, proxy_port, proxy_auth)
+
+                # TODO: Return TunnelingAgent instance
```
Gallaecio, line 99: I think we should raise `NotImplementedError`, and cover this in the relevant parts of the documentation, explaining that the new downloader handler does not support this at the moment.

It should also make the `else:` below unnecessary, and allow for dedenting its `return` statement.

scrapy/core/downloader/handlers/http2.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _get_agent(self, request: Request, timeout: Optional[float]) -> H2Agent:
+        from twisted.internet import reactor
+        bind_address = request.meta.get('bindaddress') or self._bind_address
+        proxy = request.meta.get('proxy')
+        if proxy:
+            _, _, proxy_host, proxy_port, proxy_params = _parse(proxy)
+            scheme = _parse(request.url)[0]
+            proxy_host = str(proxy_host, 'utf-8')
+            omit_connect_timeout = b'noconnect' in proxy_params
```
Gallaecio, line 87: I was a bit confused by the code below. Looking at its 1.1 counterpart, I see there is a typo here: `omit_connect_timeout` → `omit_connect_tunnel`.

scrapy/core/downloader/handlers/http2.py
```diff
@@ -0,0 +1,142 @@
...
+            if omit_connect_timeout:
+                warnings.warn("Using HTTPS proxies in the noconnect mode is deprecated. "
+                              "If you use Crawlera, it doesn't require this mode anymore, "
+                              "so you should update scrapy-crawlera to 1.3.0+ "
+                              "and remove '?noconnect' from the Crawlera URL.",
+                              ScrapyDeprecationWarning)
```
Gallaecio, line 93: @kmike @elacuesta @wRAR Given this is a new downloader handler, I wonder if we should just ignore this proxy param.

And, if so, I wonder if we should still log a warning. If we do, it should probably not be a deprecation warning, but a regular warning, maybe with the same message but replacing “deprecated” with “not supported by the <class> downloader handler”.

scrapy/core/downloader/handlers/http2.py
```diff
@@ -0,0 +1,142 @@
...
+
+            if scheme == b'https' and not omit_connect_timeout:
+                proxy_auth = request.headers.get(b'Proxy-Authorization', None)
+                proxy_conf = (proxy_host, proxy_port, proxy_auth)
+
+                # TODO: Return TunnelingAgent instance
+            else:
+                return self._ProxyAgent(
+                    reactor=reactor,
+                    proxy_uri=URI.fromBytes(bytes(proxy, encoding='ascii')),
```
Gallaecio, line 103: Better use `to_bytes` instead, which also supports `bytes` as input.

I might have been the one asking you to use `bytes` in the first place, as [I only recently found out how `to_bytes` works differently](https://github.com/scrapy/scrapy/pull/4661#issuecomment-658133232), so sorry about that :sweat_smile:

scrapy/core/http2/stream.py
```diff
@@ -185,14 +185,32 @@ def _get_request_headers(self) -> List[Tuple[str, str]]:
...
+        # URIs; "http" or "https" URIs that do not contain a path component
+        # MUST include a value of '/'. The exception to this rule is an
+        # OPTIONS request for an "http" or "https" URI that does not include
+        # a path component; these MUST include a ":path" pseudo-header field
+        # with a value of '*' (refer RFC 7540 - Section 8.1.2.3)
+        if not path:
+            if self._request.method == 'OPTIONS':
+                path = path or '*'
+            else:
+                path = path or '/'
```
Gallaecio, line 198: :lipstick: (`if not path` allows simplifying inner code)

```suggestion
        if not path:
            path = '*' if self._request.method == 'OPTIONS' else '/'
```

scrapy/core/downloader/handlers/http2.py
```diff
@@ -6,7 +6,7 @@
 from twisted.internet.base import ReactorBase
 from twisted.internet.defer import Deferred
 from twisted.internet.error import TimeoutError
-from twisted.web.client import URI
+from twisted.web.client import URI, BrowserLikePolicyForHTTPS
```
Gallaecio, line 9: :lipstick: 

```suggestion
from twisted.web.client import BrowserLikePolicyForHTTPS, URI
```

tests/test_downloader_handlers.py
```diff
@@ -786,7 +786,10 @@ def test_download_with_proxy(self):
         def _test(response):
             self.assertEqual(response.status, 200)
             self.assertEqual(response.url, request.url)
-            self.assertEqual(response.body, b'http://example.com')
+            self.assertTrue(
+                response.body == b'http://example.com'  # HTTP/1.x
+                or response.body == b'/'  # HTTP/2
+            )
```
Gallaecio, line 792: What about keeping the old check, but turning `b'http://example.com'` into a class-level variable to be overridden in HTTP/2 tests? That way the check is more strict, and will fail if the protocol-version-specific expectations are not met (e.g. we suddenly start returning `/` for HTTP/1.x).

tests/test_downloader_handlers.py
```diff
@@ -796,10 +799,13 @@ def test_download_with_proxy_https_noconnect(self):
...
             self.assertEqual(response.url, request.url)
-            self.assertEqual(response.body, b'https://example.com')
+            self.assertTrue(
+                response.body == b'http://example.com'  # HTTP/1.x
+                or response.body == b'/'  # HTTP/2
+            )
 
         http_proxy = '%s?noconnect' % self.getURL('')
-        request = Request('https://example.com', meta={'proxy': http_proxy})
+        request = Request('http://example.com', meta={'proxy': http_proxy})
```
Gallaecio, line 808: Could you elaborate on the reason behind this change?

Gallaecio:
I wonder why tests are not running. I don’t see the Travis CI configuration being modified here.

tests/test_downloader_handlers.py
adityaa30: Earlier while setting up tests with HTTP/2 I was having trouble with the *https://example.com* url. I did revert the change back to `https` in c67d6dea318d6f0915ae86d46ce367b6c1e8ee51 as it works now.

adityaa30:
> being

Yes, there is a problem in Python 3.6 job with pinned versions in Travis CI -- #4669 #4681. I am not sure of the exact source of the bug yet :disappointed:

tests/test_downloader_handlers.py
```diff
@@ -315,8 +315,8 @@ def test_host_header_seted_in_request_headers(self):
         host = self.host + ':' + str(self.portno)
 
         def _test(response):
-            self.assertEqual(response.body, bytes(host, 'utf-8'))
-            self.assertEqual(request.headers.get('Host'), bytes(host, 'utf-8'))
+            self.assertEqual(response.body, to_bytes(host))
+            self.assertEqual(request.headers.get('Host'), to_bytes(host))
```
Gallaecio, line 319: `to_bytes` is meant for scenarios where the input may be either `bytes` or `str`.

Because here we control the input type, `str`, `host.encode()` would make more sense, being more explicit.

adityaa30: Okay

docs/topics/settings.rst
```diff
@@ -620,6 +620,13 @@ handler (without replacement), place this in your ``settings.py``::
         'ftp': None,
     }
 
+The default HTTPS handler uses HTTP/1.1. To use HTTP/2 update
+:setting:`DOWNLOAD_HANDLERS` as follows::
+
+    DOWNLOAD_HANDLERS = {
+        'https': 'scrapy.core.downloader.handlers.http2.H2DownloadHandler',
+    }
+
```
Gallaecio, line 629: We should probably include a note about unencrypted HTTP/2 connections not being supported at the moment. Maybe include a reference to https://http2.github.io/faq/#does-http2-require-encryption , given that Scrapy aims to emulate browser behavior.

setup.py
```diff
@@ -95,4 +97,4 @@ def has_environment_marker_platform_impl_support():
     python_requires='>=3.5.2',
     install_requires=install_requires,
     extras_require=extras_require,
-)
+)
```
Gallaecio, line 100: :lipstick: Please, revert.

tests/test_downloader_handlers.py
```diff
@@ -230,10 +232,12 @@ def setUp(self):
         r.putChild(b"echo", Echo())
         self.site = server.Site(r, timeout=None)
         self.wrapper = WrappingFactory(self.site)
-        self.host = 'localhost'
+        self.host = u'localhost'
```
Gallaecio, line 235: ?

scrapy/core/http2/stream.py
```diff
@@ -5,14 +5,14 @@
 from urllib.parse import urlparse
 
 from h2.errors import ErrorCodes
-from h2.exceptions import H2Error, StreamClosedError
+from h2.exceptions import H2Error, StreamClosedError, ProtocolError
```
Gallaecio, line 8: :lipstick: 

```suggestion
from h2.exceptions import H2Error, ProtocolError, StreamClosedError
```

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,378 @@
...
+        self._stream_id_generator = itertools.count(start=1, step=2)
+
+        # Streams are stored in a dictionary keyed off their stream IDs
+        self.streams: Dict[int, Stream] = {}
+
+        # If requests are received before connection is made we keep
+        # all requests in a pool and send them as the connection is made
+        self._pending_request_stream_pool: deque = deque()
+
+        # Counter to keep track of opened stream. This counter
```
wRAR, line 93: "of opened streams"

scrapy/core/http2/protocol.py
```diff
@@ -0,0 +1,378 @@
...
+        """Helper function to lose the connection with the error sent as a
+        reason"""
+        self._conn_lost_errors += errors
+        self.transport.loseConnection()
+
+    def handshakeCompleted(self):
+        """We close the connection with InvalidNegotiatedProtocol exception
+        when the connection was not made via h2 protocol"""
+        negotiated_protocol = self.transport.negotiatedProtocol
+        if type(negotiated_protocol) is bytes:
```
wRAR, line 221: can this be replaced with `isinstance`?

adityaa30: Yep. I'll change it

wRAR (approved):

wRAR:
Great work!

setup.py
```diff
@@ -32,6 +32,8 @@ def has_environment_marker_platform_impl_support():
     'zope.interface>=4.1.3',
     'protego>=0.1.15',
     'itemadapter>=0.1.0',
+    'h2>=3.2.0',
+    'typing-extensions>=3.7.4',
```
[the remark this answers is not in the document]
adityaa30: I have removed the `typing-extensions` dependency in 5e36f539e28631625862b84d8ca1c7c0134584b0.

Gallaecio:
Follow up: https://github.com/scrapy/scrapy/pull/4769

PR #4612, opened 2020-06-03
add --data-raw to utils.curl and fix missing method with data

noviluni:
When copying a cURL command from the browser I got the `--data-raw` parameter.

After reading this:
* https://curl.haxx.se/docs/manpage.html#--data-raw
* https://curl.haxx.se/docs/manpage.html#-d

I think we can safely add it as a synonym of `--data`.

On the other side, this also fixes when missing the method but specifying the data (body), as the method should be interpreted by default as `POST`.

### DIFF
--- a/scrapy/utils/curl.py
+++ b/scrapy/utils/curl.py
@@ -17,7 +17,7 @@ class CurlParser(argparse.ArgumentParser):
 curl_parser = CurlParser()
 curl_parser.add_argument('url')
 curl_parser.add_argument('-H', '--header', dest='headers', action='append')
-curl_parser.add_argument('-X', '--request', dest='method', default='get')
+curl_parser.add_argument('-X', '--request', dest='method')
 curl_parser.add_argument('-d', '--data', '--data-raw', dest='data')
 curl_parser.add_argument('-u', '--user', dest='auth')
 
@@ -66,7 +66,9 @@ def curl_to_request_kwargs(curl_command, ignore_unknown_options=True):
     if not parsed_url.scheme:
         url = 'http://' + url
 
-    result = {'method': parsed_args.method.upper(), 'url': url}
+    method = parsed_args.method or 'GET'
+
+    result = {'method': method.upper(), 'url': url}
 
     headers = []
     cookies = {}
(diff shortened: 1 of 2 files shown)


scrapy/utils/curl.py
```diff
@@ -90,5 +90,9 @@ def curl_to_request_kwargs(curl_command, ignore_unknown_options=True):
         result['cookies'] = cookies
     if parsed_args.data:
         result['body'] = parsed_args.data
+        if result.get('method') == 'GET':
+            # if the "data" is specified but the "method" is not specified,
+            # the default method is 'POST'
+            result['method'] = 'POST'
```
Gallaecio, line 96: While it should not be a problem in most real-life scenarios, `curl <URL> -X GET --data asdf` works. We need to distinguish when `-X` is passed with `GET` and when it is omitted.

noviluni: Hi @Gallaecio, 

I'm not sure which approach to take. From my perspective, this command tries to do the best from the cURL command, and I think that it doesn't have any sense to add the body when the method is GET.

Before deciding this I checked other tools as https://curl.trillworks.com/ , and when entering:
```bash
curl URL -X GET --data asdf
```

I got:
```python
import requests

data = 'asdf'

response = requests.post('http://URL', data=data)
```

Of course, I can change it to respect the provided GET, but I'm not sure about the best option. Any other idea / thoughts?

Gallaecio: I’m not against merging the code as is.

But a server may support a GET with a form even if it is not standard:

```
[adrian@afonsox ~]$ curl httpbin.org/anything -X GET --data asdf
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {
    "asdf": ""
  }, 
  "headers": {
    "Accept": "*/*", 
    "Content-Length": "4", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "Host": "httpbin.org", 
    "User-Agent": "curl/7.70.0", 
    "X-Amzn-Trace-Id": "Root=…"
  }, 
  "json": null, 
  "method": "GET", 
  "origin": "…", 
  "url": "http://httpbin.org/anything"
}
```

So I would either implement `GET` support here or create a separate issue to eventually add it.

noviluni: Ok, I fixed the code to respect this behavior. :+1:

PR #4613, opened 2020-06-06
Update headless browser docs

### LINKED ISSUE #4484: Scrapy with Puppeteer and/or Playwright?
Hi Team, 

I would like to aks , is there any development plan to create Scrapy for Puppeteer and/or Playwright?

tnx a lot

elacuesta:
Closes #4484


docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,34 @@ Using a headless browser
...
+            title = await page.title()
+            yield {"title": title}
+
+Keep in mind that this is just a proof of concept, since it circumvents
+most of the Scrapy components (middlewares, dupefilter, etc).
+
+There are some 3rd party projects which provide better integration:
+
+* https://github.com/elacuesta/scrapy-pyppeteer
+* https://github.com/lopuhin/scrapy-pyppeteer
```
kmike, line 277: I probably wouldn't suggest this one. @lopuhin do you intend to maintain it?

docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,34 @@ Using a headless browser
...
+    class PyppeteerSpider(scrapy.Spider):
+        name = "pyppeteer"
+        start_urls = ["data:,"]  # avoid making an actual upstream request
+
+        async def parse(self, response):
+            browser = await pyppeteer.launch()
+            page = await browser.newPage()
+            await page.goto("https:/example.org")
+            title = await page.title()
+            yield {"title": title}
```
kmike, line 269: I think this needs `await browser.close()` to be correct, probably with some sort of try/except, and avoid leaking resources.

docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,34 @@ Using a headless browser
...
+Python port of `puppeteer`_), which uses headless Chrome to download and
+render pages.
+The following is a simple snippet to illustrate its usage within Scrapy::
+
+    import pyppeteer
+    import scrapy
+
+    class PyppeteerSpider(scrapy.Spider):
+        name = "pyppeteer"
+        start_urls = ["data:,"]  # avoid making an actual upstream request
```
kmike, line 262: TIL, nice trick

docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,38 @@ Using a headless browser
 A `headless browser`_ is a special web browser that provides an API for
 automation.
 
-The easiest way to use a headless browser with Scrapy is to use Selenium_,
-along with `scrapy-selenium`_ for seamless integration.
+Since version 2.0, it is possible to integrate libraries that use the
+``async/await`` syntax. One such library is `pyppeteer`_ (an unnoficial
+Python port of `puppeteer`_), which uses headless Chrome to download and
+render pages.
```
Gallaecio, line 254: I would not mention coroutine syntax or the Scrapy version that introduced it. I would simply state here that installing the asyncio reactor (with a link to the corresponding section, which explains that support was added in Scrapy 2.0) allows you to use pyppeteer, an asyncio-based headless Chrome/Chromium automation library.

Doing this also makes the first post-example paragraph unnecessary.

docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,38 @@ Using a headless browser
...
+            await page.close()
+            yield {"title": title}
+
+For this example to work, Scrapy needs to be running on top of the
+:ref:`asyncio reactor <install-asyncio>`.
+
+Keep in mind that this is just a proof of concept, since it circumvents
+most of the Scrapy components (middlewares, dupefilter, etc).
+
+The following is a list of 3rd party projects which provide better integration:
```
Gallaecio, line 278: :lipstick: 

```suggestion
Using pypeteer_ directly circumvents most of the 
Scrapy components (middlewares, dupefilter, etc). Use
one of the following Scrapy plugins for better integration
with Scrapy:
```

docs/topics/dynamic-content.rst
```diff
@@ -248,8 +248,38 @@ Using a headless browser
...
+
+* https://github.com/elacuesta/scrapy-pyppeteer
+* https://github.com/lopuhin/scrapy-pyppeteer
+* https://github.com/clemfromspace/scrapy-puppeteer
```
Gallaecio, line 282: It would be great if we could feature some pros of using each solution. And if any solution provides no pros over the rest, it may be better not to cover it.

wRAR:
Should this be now rewritten to Playwright?

docs/topics/dynamic-content.rst
elacuesta: I removed https://github.com/lopuhin/scrapy-pyppeteer and https://github.com/clemfromspace/scrapy-puppeteer because they are unmaintained, and https://github.com/clemfromspace/scrapy-selenium because Selenium is blocking and I don't think we should encourage it's usage here.

Gallaecio (approved):
I suggest keeping the alphabetical sorting of the links at the end, but other than that :heavy_check_mark:

Issue #4561, opened 2020-05-10
`scrapy genspider` should not overwrite existing file


metaperl:
## Summary

If the file mentioned in `scrapy genspider` already exists, then genspider should refuse to generate the the file.

## Motivation

As it stands, existing code can be blown away if this command runs twice.

## Describe alternatives you've considered

Prompting the user for overwriting existing spider.

jay24rajput:
Hey, is this Issue up for grabs??

metaperl:
As far as I am concerned, yes.

jay24rajput:
@Gallaecio @elacuesta thoughts??

elacuesta:
I would not have thought this was an issue, as I'd expect spider code to be covered by source control. On the other hand, it should be only a matter of doing a file system check, so I think a clean implementation could be approved.

jay24rajput:
Alright! I will work on this one

sivoham:
Hi @metaperl @elacuesta,

I would like to understand this properly. I've tried to create a spider with the same name twice but it does not allow me to do so. 

`(p3env) bash-3.2$ scrapy genspider toi 'https://timesofindia.indiatimes.com/'
Created spider 'toi' using template 'basic' in module:
  testscrapy.spiders.toi
(p3env) bash-3.2$
(p3env) bash-3.2$
(p3env) bash-3.2$ scrapy genspider toi 'https://timesofindia.indiatimes.com/'
Spider 'toi' already exists in module:
  testscrapy.spiders.toi`

Could you please let me know what exactly is the issue so that I can try attempting to resolve or enhance?

Thanks & Regards.

faraz16iqbal:
Is this issue still up for grabs?

elacuesta:
@sivoham Indeed, there is already a check in place, but what it does is try to [load a spider](https://github.com/scrapy/scrapy/blob/2.1.0/scrapy/commands/genspider.py#L69-L78) by its [`name` attribute](https://docs.scrapy.org/en/2.1/topics/spiders.html#scrapy.spiders.Spider.name). So a file can get overridden if you have a spider which name doesn't match its file name.

@faraz16iqbal I'd recommend you not to duplicate efforts, since there are already open PRs about this. You're more than welcome to check any other open issues though. You could check the ["good first issue" tag](https://github.com/scrapy/scrapy/labels/good%20first%20issue) if you're unsure about where to start, and don't hesitate to ask for guidance if you need.

PR #4616, opened 2020-06-08
#4561 - Update genspider.py

3Dook:
Potential Issue #4561 enhancement fix. The bug/glitch occurs when genspider is called outside of a startproject. I added a small and simple check to compare the new spider names to the current files in the directory. If there is already a spider with the file name it will return and stop the function.


wRAR:
Can't you check if the file exists directly?

wRAR:
Fixed in #4623.

PR #4618, opened 2020-06-09
response.request attribute binding

### LINKED ISSUE #4529: Bind request to response only if not bound already
## Summary
Currently requests are bound to responses in [the engine](https://github.com/scrapy/scrapy/blob/2.1.0/scrapy/core/engine.py#L247) and/or in [the scraper](https://github.com/scrapy/scrapy/blob/2.1.0/scrapy/core/scraper.py#L149).
I would like to start a conversation to figure out the implications of binding them only if the `response.request` attribute is not already set, i.e. something like this:

```diff
diff --git scrapy/core/engine.py scrapy/core/engine.py
index 77d71846..72f1628a 100644
--- scrapy/core/engine.py
+++ scrapy/core/engine.py
@@ -244,7 +244,8 @@ class ExecutionEngine:
                     % (type(response), response)
                 )
             if isinstance(response, Response):
-                response.request = request  # tie request to response received
+                if response.request is None:
+                    response.request = request
                 logkws = self.logformatter.crawled(request, response, spider)
                 if logkws is not None:
                     logger.log(*logformatter_adapter(logkws), extra={'spider': spider})


diff --git scrapy/core/scraper.py scrapy/core/scraper.py
index edbb4dd6..256935da 100644
--- scrapy/core/scraper.py
+++ scrapy/core/scraper.py
@@ -146,7 +146,8 @@ class Scraper:
                 self._log_download_errors, request_result, request, spider)
 
     def call_spider(self, result, request, spider):
-        result.request = request
+        if not hasattr(result, "request") or result.request is None:
+            result.request = request
         dfd = defer_result(result)
         callback = request.callback or spider.parse
         warn_on_generator_with_return_value(spider, callback)
```

## Motivation

Recently I've been working on a downloader middleware to transparently redirect requests to a service which takes an URL and returns its response body along with some other metadata. The `process_request` method replaces the URL, and the `process_response` decodes the response body and creates a new response, making it look like it came directly from the target website, not from the download service. At this point, I'd like to be able to restore the original request as well, but any request I return in the `Response.request` attribute gets overridden.

ATM I don't see a way to modify this behaviour without altering the Scrapy core.

elacuesta:
See #4529

🚧 **Under construction** 🚧 

Implemented as suggested in the original issue. The additional attributes are temporary flags only to be checked by the tests, which should help understand the behaviour of the attribute binding.  Each test method has a docstring with a description of what happens in a particular scenario.

Should close #4529 at some point.

I'd appreciate any feedback you might have @dangra @Gallaecio @kmike @wRAR 

(edit) this PR will remain a draft, see #4632 for the actual proposed patch.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -246,11 +246,15 @@ class ExecutionEngine:
                     response._request_set_in_engine = True
                 else:
                     response._request_set_in_engine = False
-                logkws = self.logformatter.crawled(request, response, spider)
+                logkws = self.logformatter.crawled(response.request, response, spider)
                 if logkws is not None:
                     logger.log(*logformatter_adapter(logkws), extra={'spider': spider})
-                self.signals.send_catch_log(signals.response_received,
-                                            response=response, request=request, spider=spider)
+                self.signals.send_catch_log(
+                    signal=signals.response_received,
+                    response=response,
+                    request=response.request,
+                    spider=spider,
+                )
             return response
 
         def _on_complete(_):
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -241,7 +241,11 @@ class ExecutionEngine:
                     % (type(response), response)
                 )
             if isinstance(response, Response):
-                response.request = request  # tie request to response received
+                if response.request is None:
+                    response.request = request
+                    response._request_set_in_engine = True
+                else:
+                    response._request_set_in_engine = False
                 logkws = self.logformatter.crawled(request, response, spider)
                 if logkws is not None:
                     logger.log(*logformatter_adapter(logkws), extra={'spider': spider})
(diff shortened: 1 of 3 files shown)


Gallaecio:
While I have no problem with adding private members if that is needed for proper testing, in this scenario I wonder if the issue is that we are trying to test the implementation instead of testing the expectations.

I think we do not need to test whether the request is set by the scraper or the engine. Instead, we should test that `response.request`, by the time it reaches e.g. a spider callback or a downloader middleware configured before or after a middleware that changes the request, has the expected value in the different scenarios.

If we focus on the end-user expectations instead of whether the engine or the scraper make those expectations possible, the private members should be unnecessary.

elacuesta:
Right, I did this mainly so we could understand what's happening. Maybe what I can do is open a separate PR with the actual changes I intend to be merged, and keep this one as reference.

PR #4623, opened 2020-06-10
Check if file is already present on running `scrapy genspider` and terminate if so (#4561)

### LINKED ISSUE #4561: `scrapy genspider` should not overwrite existing file

ajaymittur28:
Continuing from PR #4616 to fix #4561, I've added a direct file check to see if a spider with the name provided in `scrapy genspider` is already present in the directory and not overwrite the existing spider file if present.


scrapy/commands/genspider.py
```diff
@@ -66,6 +66,10 @@ def run(self, args, opts):
             print("Cannot create a spider with the same name as your project")
             return
 
+        if exists(name + ".py"):
+            print("Spider %r already exists in module" % name)
```
elacuesta, line 70: This message is not entirely accurate, as any file with the same name would trigger it (not necessarily a spider).

tests/test_commands.py
```diff
@@ -188,6 +188,10 @@ def test_same_name_as_project(self):
         self.assertEqual(2, self.call('genspider', self.project_name))
         assert not exists(join(self.proj_mod_path, 'spiders', '%s.py' % self.project_name))
 
+    def test_same_name_as_existing_spider(self):
+        self.call('genspider', 'example', 'example.com')
+        self.assertEqual(0, self.call('genspider', 'example', 'example.com'))
```
elacuesta, line 193: This only checks that the command finishes correctly with an exit code 0. Is there a way we could also test that the file is not overridden? Checking that the printed message appears in the command's output should be enough I think.

ajaymittur28: I think we could maybe compare the md5 checksum of the file before and after if it's really necessary. Do you think we need that or should I just stick to checking the printed message?

scrapy/commands/genspider.py
ajaymittur28: Do you think a message like "File %r.py already exists in the current directory" would be more accurate?

elacuesta: I think talking about the file is indeed better, thanks for making the change. I don't think adding a `--force` option is necessary, in my opinion it'd just make the code more complex for something that could be solved easily outside of Scrapy (renaming the existing file or choosing a different name for the new file).

tests/test_commands.py
elacuesta: Just checking the message would be enough IMHO. I see you already made the change, thanks for that. Could you also check that the message does not appear the first time?

ajaymittur28: Yep sure, will add it to the test

scrapy/commands/genspider.py
ajaymittur28: I agree, so I guess no more changes here then?

tests/test_commands.py
ajaymittur28: Have added the check

elacuesta:
Many thanks for the work @ajaymittur28.

It's looking good so far, but I tested this locally and realized that since the `exists` check is done only with the passed name, it doesn't work in the context of a project:
```
$ scrapy startproject foobar
New Scrapy project 'foobar', using template directory '/.../scrapy/templates/project', created in:
    /.../foobar

    cd foobar

$ cd foobar/

$ scrapy genspider example example.com
Created spider 'example' using template 'basic' in module:
  foobar.spiders.example

# rename the spider so it's not handled by https://github.com/scrapy/scrapy/blob/2.1.0/scrapy/commands/genspider.py#L69-L78
$ sed -i -e "s/name = 'example'/name = 'renamed'/g" foobar/spiders/example.py

$ scrapy genspider example example.com
Created spider 'example' using template 'basic' in module:
  foobar.spiders.example
```

The added test confirms the patch works for standalone spiders, but it'd be good to also make it work for spiders within projects. Do you think you can make such a change?

ajaymittur28:
Yep, I can make it check for the case you just mentioned when it's run within projects as well. working on it! Thanks for the console output!

ajaymittur28:
Pushed the latest changes that handle the case mentioned above.

I made a few changes from the previous commits:
1. **Added a check to see if the command was being run as a standalone or from a project** (or rather if the `spiders` module exists), since I felt if it was being run from a project, the original issue (overwriting existing file in same directory) wouldn't be a problem since the spiders would always be generated inside the `spiders\` directory
    - This is where the case you mentioned could occur, if run within a project, and a spider with the filename passed to `genspider` exists in the `spiders\` directory but is not being handled by lines 69-78 in [scrapy/scrapy/commands/genspider.py](https://github.com/scrapy/scrapy/blob/master/scrapy/commands/genspider.py#L69-L78)

2. **Moved the code to after the attempt to load the spider was done** [here](https://github.com/scrapy/scrapy/pull/4623/files#diff-6433bda22389235c9a64af4e32a40009R72-R83) and a `KeyError` exception (spider not found) was thrown because:- 

    Output if check is done before trying to load spider
    ```
    $ scrapy genspider example example.com
    Created spider 'example' using template 'basic' in module:
      foobar.spiders.example

    $ scrapy genspider example example.com
    Spider with filename 'example.py' already exists in directory:
       <path_of_spiders_directory>
    ```

    Output if check is done after trying to load the spider
    ```
    $ scrapy genspider example example.com
    Created spider 'example' using template 'basic' in module:
      foobar.spiders.example

    $ Spider 'example' already exists in module:
      foobar.spiders.example
    ```

    Since the tests already present expect it to check for the spidername first (display second output) I moved the code to inside the `except` block which runs if the spider with the given name wasn't found; and then proceed to check if a spider with filename same as provided exists.

scrapy/commands/genspider.py
```diff
@@ -69,7 +69,18 @@ def run(self, args, opts):
...
+                if exists(name + ".py"):
+                    print("File %r already exists in the current directory" % (name + ".py"))
+                    return
+            else:
+                spiders_module = import_module(self.settings['NEWSPIDER_MODULE'])
+                spiders_dir = dirname(spiders_module.__file__)
+                spiders_dir_abs = abspath(spiders_dir)
+                if exists(join(spiders_dir_abs, name + ".py")):
+                    print("Spider with filename %r already exists in directory:" % (name + ".py"))
+                    print("  %s" % spiders_dir)
```
Gallaecio, line 82: :lipstick:  I would simplify both messages to "%r already exists", where %r is the absolute path of the destination file.

tests/test_commands.py
```diff
@@ -215,7 +238,7 @@ def start_requests(self):
         return []
 """
 
-    @contextmanager
+    @ contextmanager
```
Gallaecio, line 241: Accidental change?

scrapy/commands/genspider.py
```diff
@@ -69,7 +69,18 @@ def run(self, args, opts):
         try:
             spidercls = self.crawler_process.spider_loader.load(name)
         except KeyError:
-            pass
+            if not self.settings.get('NEWSPIDER_MODULE'):
+                if exists(name + ".py"):
+                    print("File %r already exists in the current directory" % (name + ".py"))
+                    return
+            else:
```
Gallaecio, line 76: Could you add a test to cover the case where `NEWSPIDER_MODULE` is defined?

tests/test_commands.py
ajaymittur28: Oops, yea, sorry about that will revert back

scrapy/commands/genspider.py
ajaymittur28: Seems more concise, will make the change!

scrapy/commands/genspider.py
ajaymittur28: Do you mean replacing the `else` with an `elif self.settings.get('NEWSPIDER_MODULE)`?, The `else` block should handle that case right?

ajaymittur28: I've added a test to the case @elacuesta mentioned, `test_same_filename_as_existing_spider` under `GenspiderCommandTest` where `NEWSPIDER_MODULE` will be defined when the test is run. Could you please elaborate? don't quite understand what you mean.

Gallaecio: > where NEWSPIDER_MODULE will be defined when the test is run

Where is it defined? Or do you mean that people running the tests must define it manually somewhere before running the test?

ajaymittur28: Sorry, should have phrased it better. From what I understand of the code base, when the `scrapy` command is run in the command line, the `settings` object is set in line 112 of the [cmdline](https://github.com/scrapy/scrapy/blob/master/scrapy/cmdline.py#L112) module, digging further into `get_project_settings()`, I found that the value of `NEWSPIDER_MODULE` **is set here to either an empty string or the module where the spiders are located** in case of a project. This `settings` object is then passed on to [cmd](https://github.com/scrapy/scrapy/blob/master/scrapy/cmdline.py#L137) in line 137.

At this point `cmd` **is already set to** `genspider` and the line 137 `cmd.settings = settings` assigns the previously mentioned `settings` object to `genspider.settings`. This settings dictionary is what I'm accessing through `self.settings`. So essentially the `NEWSPIDER_MODULE` property will always be defined (either to `''` or the spiders module)

Gallaecio: Makes sense.

So, given your code also handles the scenario where the `NEWSPIDER_MODULE` setting is defined with a non-empty string as value, I think we need a test for that scenario.

ajaymittur28: The value of `NEWSPIDER_MODULE` is set implicitly when the `scrapy` command is run, I think the only way we can test it for both the scenarios is if we test running `genspider` as a standalone command and then within a project scope, which I think I have handled in the tests file.

ajaymittur28: `test_same_filename_as_existing_spider` for when the command is run within a project and `test_same_name_as_existing_file` for when the command is run as a standalone

Gallaecio: OK, I finally understand [why the tests cover `NEWSPIDER_MODULE` usage](https://github.com/scrapy/scrapy/blob/master/scrapy/templates/project/module/settings.py.tmpl#L13). Sorry! :flushed:

scrapy/commands/genspider.py
```diff
@@ -69,7 +69,17 @@ def run(self, args, opts):
...
+                if exists(name + ".py"):
+                    print("%r already exists" % (abspath(name + ".py")))
+                    return
+            else:
+                spiders_module = import_module(self.settings['NEWSPIDER_MODULE'])
+                spiders_dir = dirname(spiders_module.__file__)
+                spiders_dir_abs = abspath(spiders_dir)
+                if exists(join(spiders_dir_abs, name + ".py")):
+                    print("%r already exists" % (join(spiders_dir_abs, (name + ".py"))))
+                    return
```
Gallaecio, line 82: Looking at lines 84-88, I suspect this implementation will fail even if `--force` is used.

ajaymittur28: Hmm...I had discussed the same with @elacuesta https://github.com/scrapy/scrapy/pull/4623#discussion_r439445481 and he said since it was just a matter of renaming the file which the user could do easily it wasn't required. However, I could add it if its needed.

elacuesta: Right, sorry about that. I missed the fact that the `--force` option already exists for this command. Given that, it probably makes sense to take the flag into account.

ajaymittur28: No worries, I'll add it then

scrapy/commands/genspider.py
```diff
@@ -69,7 +69,23 @@ def run(self, args, opts):
         try:
```
Gallaecio, line 69: I’m thinking you could extract the `opts.force` check out of the whole `try`-`except`, since the whole `try`-`except` is only there to check if the spider or file already exist:

```python
if not opts.force:
    try:
    #…
```

Also, since this part of the code is getting long, maybe it would make sense to take the check into its own method:

```python

if not opts.force and self._spider_exists(name):
    return

# …

def _spider_exists(self, name):
    # …
```

With this later approach, you could leave `pass` within `except KeyError`, and move the additional checks _after_ the `else` of the `try`-`except`, using early returns to minimize indentation.

ajaymittur28:
I'll refactor the code, thanks for the suggestions

scrapy/commands/genspider.py
ajaymittur28: Did you mean something like this?
```py
def _spider_exists(self, name):
        if not self.settings.get('NEWSPIDER_MODULE'):
            # if run as a standalone command and file with same filename already exists
            if exists(name + ".py"):
                print("%s already exists" % (abspath(name + ".py")))
                return True
        else:
            try:
                spidercls = self.crawler_process.spider_loader.load(name)
            except KeyError:
                pass
            else:
                # if spider with same name exists
                print("Spider %r already exists in module:" % name)
                print("  %s" % spidercls.__module__)
                return False
            # if spider same filename exists
            spiders_module = import_module(self.settings['NEWSPIDER_MODULE'])
            spiders_dir = dirname(spiders_module.__file__)
            spiders_dir_abs = abspath(spiders_dir)
            if exists(join(spiders_dir_abs, name + ".py")):
                print("%s already exists" % (join(spiders_dir_abs, (name + ".py"))))
                return True

        return False
```
Then the entire block including (L69-94) can be reduced to 
```py
if not opts.force and self._spider_exists(name): 
    return
```

scrapy/commands/genspider.py
```diff
@@ -135,6 +112,32 @@ def _list_templates(self):
...
+        else:
+            try:
+                spidercls = self.crawler_process.spider_loader.load(name)
+            except KeyError:
+                pass
+            else:
+                # if spider with same name exists
+                print("Spider %r already exists in module:" % name)
+                print("  %s" % spidercls.__module__)
+                return False
```
Gallaecio, line 130:
```suggestion
                return True
```

I don’t understand why the tests are not breaking, though. Maybe I am missing something.

If the tests should be breaking with the current code, please first make sure they do, and _then_ change the return value.

scrapy/commands/genspider.py
```diff
@@ -135,6 +112,32 @@ def _list_templates(self):
             if filename.endswith('.tmpl'):
                 print("  %s" % splitext(filename)[0])
 
+    def _spider_exists(self, name):
+        if not self.settings.get('NEWSPIDER_MODULE'):
+            # if run as a standalone command and file with same filename already exists
+            if exists(name + ".py"):
+                print("%s already exists" % (abspath(name + ".py")))
+                return True
```
Gallaecio, line 120: :lipstick: If you add a `return False` statement after this `if`, at the same indentation level as the `if`, then the `else` below of the upper `if` is not needed anymore, allowing you to remove 1 level of indentation from all the code below.

scrapy/commands/genspider.py
ajaymittur28: It looks like the [test](https://github.com/scrapy/scrapy/pull/4623/files#diff-59262c12118a93348565a41ab7f6ce42R168-R169) for this case only asserts the output, and because both "Spider %r already exists in module" and "Created spider %r using template %r " were being displayed it passed the test. On changing it to `return True` only the former will be displayed.
Apologies, can't believe I missed it.

Gallaecio: If you had not made a mistake with the return value here, we would probably have not detected the issue in the test coverage. Now you can improve the test :slightly_smiling_face:

ajaymittur28: I think we could check the modification time of the file before and after the command is run to ensure the file wasn't rewritten on top of asserting the output to make the test stronger. What do you think? The `os.path` module in python has a method `os.path.getmtime` to get the modification time, so shouldn't be too overbearing.

ajaymittur28: The tests fail as expected now

ajaymittur28: @Gallaecio could you please have a look at the code and tell me what you think when possible

ajaymittur28:
Yep, will do that, initially thought doing so will stress the memory reading the file everytime but since it's only once during the test and file doesn't seem to be that big should be feasible.

ajaymittur28:
Previous commit resolves a merge conflict in the imports in `test_command.py`

scrapy/commands/genspider.py
```diff
@@ -119,6 +112,33 @@ def _list_templates(self):
...
+        try:
+            spidercls = self.crawler_process.spider_loader.load(name)
+        except KeyError:
+            pass
+        else:
+            # if spider with same name exists
+            print("Spider %r already exists in module:" % name)
+            print("  %s" % spidercls.__module__)
+            return True
+        # if spider same filename exists
```
elacuesta, line 132:
```suggestion

        # a file with the same name exists in the target directory
```
Grammar police :oncoming_police_car:

ajaymittur28: Whoops! my bad :)

Gallaecio:
Thank you!

PR #4624, opened 2020-06-11
Updated html on quotes.html page and revised documentation per html specs

trand22:
- modified quotes.html to include more information and clarify for a landing page. CSS for styling, added new information.

- modified overview.rst in documentation to reflect information on developer salaries and opportunities to provide incentive for new users to adopt Scrapy into their skill set.


wRAR:
Also, is that html page displayed anywhere?

wRAR:
Closing as this should be two separate PRs in any case, and I agree that both changes are not desirable.

PR #4630, opened 2020-06-14
Cover Scrapy 2.2.0 in the release notes

Gallaecio:


docs/news.rst
```diff
@@ -3,6 +3,184 @@
...
+-------------------------
+
+Highlights:
+
+* Python 3.5.2+ is required now
+* :ref:`dataclass objects <dataclass-items>` and
+  :ref:`attrs objects <attrs-items>` are now valid :ref:`item types
+  <item-types>`
+* New :meth:`TextResponse.json <scrapy.http.TextResponse.json>` method
+* New :signal:`bytes_received` signal
```
kmike, line 18: What do you think about mentioning here that this signal allows to implement arbitrary logic for aborting of the downloading of the responses?

docs/news.rst
```diff
@@ -3,6 +3,194 @@
...
+
+
+New features
+~~~~~~~~~~~~
+
+*   :ref:`dataclass objects <dataclass-items>` and
+    :ref:`attrs objects <attrs-items>` are now valid :ref:`item types
+    <item-types>`, and a new itemadapter_ library makes it easy to
+    write code that :ref:`supports any item type <supporting-item-types>`
+    (:issue:`2749`, :issue:`2807`, :issue:`3761`, :issue:`3881`)
```
kmike, line 48: I'm not 100% sure, but it seems because of this change Scrapy got support for exporting of nested scrapy.Item instaces. Previously dict(item) was used in serialize function, not it is .asdict() - https://github.com/scrapy/scrapy/blob/5d541731870eaaa3dd658673b82a7d0ca056f689/scrapy/utils/serialize.py#L30

Gallaecio: I don’t think the output of `ScrapyJSONEncoder` is any different now; it changes how it reaches the final result, but not the final result itself.

The following code (based on https://github.com/scrapy/scrapy/issues/3888) outputs the same from `master` and `2.1.0`:

```python
from scrapy import Field, Item
from scrapy.utils.serialize import ScrapyJSONEncoder


class TopLevel(Item):
    foo = Field()
    bars = Field()


class Bar(Item):
    spam = Field()


item = TopLevel(foo=42, bars=[Bar(spam='hello'), Bar(spam='world')])
encoder = ScrapyJSONEncoder()
encoded = encoder.encode(item)
print(encoded)
```

The output is always:

```
{"foo": 42, "bars": [{"spam": "hello"}, {"spam": "world"}]}
```

I believe that encode method is called recursively, so what changes is that the current core would call `asdict` once and then export as usual (because it already converts all items to `dict` on the first `asdict` call), while the previous code would cast to `dict` every item in the structure before exporting those `dict` as usual.

Gallaecio: I’ve checked `asdict` usage in Scrapy, and I think the only user-visible change will be in the output of `scrapy parse`, where nested items will now be displayed as `dict` instead of as their original item class.

Not sure that’s worth mentioning, though.

kmike: Thanks for the investigation @Gallaecio!

PR #4632, opened 2020-06-15
Conditional request attribute binding for responses/failures

### LINKED ISSUE #4529: Bind request to response only if not bound already

elacuesta:
Fixes #4529, closes #4618.

Additional tests to understand the current and proposed behaviour in #4618.


scrapy/core/scraper.py
```diff
@@ -146,7 +146,8 @@ def _scrape2(self, request_result, request, spider):
                 self._log_download_errors, request_result, request, spider)
 
     def call_spider(self, result, request, spider):
-        result.request = request
+        if getattr(result, "request", None) is None:
+            result.request = request
```
Gallaecio, line 150: I reviewed the surrounding code, and it looks like we can assume that `result` here is a `Response` object. So maybe a simpler change, like the one in `engine.py`, would work the same.

I would also suggest renaming `result` to `response` if it were not a backward-incompatible change :expressionless:

elacuesta: Actually, `result` can be a `Failure` object, that's what the attribute check is needed: https://github.com/scrapy/scrapy/pull/4618/commits/3992338573ba64f8c7ebaf5cee919068fcdd2ea9#diff-d12d4e527e55eef105fcf232ef856619R80-R83

elacuesta:
> shouldn’t we modify the rest of the code of those two methods to also use `response.request` instead of the original `request` argument?

Very good point, thanks for noticing!

Gallaecio:
What about adding tests that cover https://github.com/scrapy/scrapy/pull/4632/commits/de4094c293ee3e63b703ff1a42c3b9301e7c8b3f ?

Gallaecio:
I wonder if it wouldn’t make sense to merge `test_crawled_log_message` into `test_response_received_signal`.

And I’m guessing `test_crawled_log_message` only tests 1 of the 2 changes. It would be great if we could have a second test which covers the other change.

elacuesta:
Thanks for the suggestion, I merged 3 similar tests. Regarding the missing test (about using callback/errback from the overriden request, just to make sure we're talking about the same thing), it's in my plans to add it, I'll probably do it this week.

Gallaecio:
> about using callback/errback from the overriden request, just to make sure we're talking about the same thing

Actually, I think we are talking about different things.

I meant that the code changes involve 2 changes, or better put changes in 2 places: engine and scraper. I believe we need to have separate tests that check that each of those changes work, as I _assume_ the current test would pass even if one of those changes was reverted.

amarynets:
Hi, @elacuesta thanks a lot for work that you already have done. I just checked my issue #4633  with changes that you added and it does now work correctly.
1. The id of my first request is 140571516279888(with original URL)
2. Then I make the second request with id 140571516280784
3. In the `process_response`, I have the following: `response.request = request.replace(url=request.mata['original_url'])
In callback I have `responce.request.url not from origin URL but from the last request

elacuesta:
Pushed a new test to check overriden callbacks. I tried the same with errbacks, but realized it doesn't really make sense: the method I'm mostly testing here is `process_response`, if I return a `Response` with a `request` that overrides the errback, the only way to send it to the errback is by raising an exception in a later middleware. However, later `process_exception` methods do not get the overriden response. Of course I might be missing something, please make sure this makes sense.

@Gallaecio Regarding https://github.com/scrapy/scrapy/pull/4632#issuecomment-651067859, the component that logs the "crawled" message is the engine, not the scraper, I'm not sure what else can be tested. Please let me know about any ideas  you might have.

Gallaecio:
The current tests continue to pass if we separately revert the changes in 3 lines:

```
scrapy/core/scraper.py:154: result.request → request
scrapy/core/scraper.py:156: result.request → request
scrapy/core/scraper.py:157: result.request → request
```

I think we need to figure out a way to make the tests break when any of these changes is made.

elacuesta:
Regarding the discussed lines:

https://github.com/elacuesta/scrapy/blob/c11b1696fc5a81c025146406f7527a3bb5cb8060/scrapy/core/scraper.py#L154-L157

```python
(154) warn_on_generator_with_return_value(spider, result.request.errback)
(155) dfd.addCallbacks(callback=callback,
(156)                  errback=result.request.errback,
(157)                  callbackKeywords=result.request.cb_kwargs)
```

* 154 is not really important, it's just a warning and it doesn't affect the request/response flow at all
* 156 is the one I think cannot be tested because it never changes actual behaviour. This change is about the `process_response` method, which handles successful responses, so the errback is not invoked.
* 157 is tested in e5b5d0d

Gallaecio:
@elacuesta I finally understood the reason why those 2 line changes have no effect :sweat_smile: 

I think my main issue was the way `call_spider` is written. What would everyone think about [refactoring it like this](https://github.com/elacuesta/scrapy/pull/8)?

PR #4636, opened 2020-06-18
improve _matches()

noviluni:
We don't need to check all the regex before checking if any of them worked, we can return `True` as soon as we find the first coincidence for better performance.

### DIFF
--- a/scrapy/linkextractors/__init__.py
+++ b/scrapy/linkextractors/__init__.py
@@ -48,7 +48,10 @@ _re_type = type(re.compile("", 0))
 
 
 def _matches(url, regexs):
-    return any(r.search(url) for r in regexs)
+    for r in regexs:
+        if r.search(url):
+            return True
+    return False
 
 
 def _is_valid_url(url):


wRAR:
I'd expect `any` to work in the same way, or am I missing something?

noviluni:
Oh right!

I did some tests and I got a better performance using the `for`loop, but after reading the docs (https://docs.python.org/3/library/functions.html#any) `any` does exactly the same, so now I'm not sure what to think... probably my tests are wrong... :sweat_smile: 

```python
def _matches(): 
     return any(n == 99 for n in range(1, 100))

def _matches2(): 
    for n in range(1, 100): 
        n == 99: 
            return True 
    return False                                                                                                  
```

```python
>>> timeit.timeit('_matches()', setup="from __main__ import _matches", number=1000000)              
5.645610346000467

>>> timeit.timeit('_matches2()', setup="from __main__ import _matches2", number=1000000)            
3.2327119680003307
```

```python
>>> cProfile.run("_matches()")                                                                      
         105 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <ipython-input-69-055b68aa3f72>:1(_matches)
      100    0.000    0.000    0.000    0.000 <ipython-input-69-055b68aa3f72>:2(<genexpr>)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.any}
        1    0.000    0.000    0.000    0.000 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}



>>> cProfile.run("_matches2()")                                                                     
         4 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <ipython-input-70-80f92e73d27a>:1(_matches2)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

Could you help me to understand why my tests are wrong? :slightly_frowning_face:

noviluni:
Ok, got it!

noviluni:
I did some tests and I got a better performance using the forloop, but after reading the docs (https://docs.python.org/3/library/functions.html#any) any does exactly the same, so now I'm not sure what to think... probably my tests are wrong... :sweat_smile: 

```python
def _matches(): 
     return any(n == 99 for n in range(1, 100))

def _matches2(): 
    for n in range(1, 100): 
        if n == 99: 
            return True 
    return False                                                                                                  
```

```python
>>> timeit.timeit('_matches()', setup="from __main__ import _matches", number=1000000)              
5.645610346000467

>>> timeit.timeit('_matches2()', setup="from __main__ import _matches2", number=1000000)            
3.2327119680003307
```

```python
>>> cProfile.run("_matches()")                                                                      
         105 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <ipython-input-69-055b68aa3f72>:1(_matches)
      100    0.000    0.000    0.000    0.000 <ipython-input-69-055b68aa3f72>:2(<genexpr>)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.any}
        1    0.000    0.000    0.000    0.000 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}



>>> cProfile.run("_matches2()")                                                                     
         4 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <ipython-input-70-80f92e73d27a>:1(_matches2)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

Could you help me to understand why my tests are wrong? :slightly_frowning_face:

noviluni:
I'm thinking that maybe the difference in performance is due to the `range()` :thinking:

wRAR:
If you want profiling you should profile multiple calls, as the current data is all zeroes :)

I also think it's useful to search similar comparisons on the Internet as I'm sure people already did that.

noviluni:
Anyway, the docs don't lie and this experiment allows to check that you were right:

```python
def p(x): 
    print(x) 
    return x
```

```python
>>> any(p(n) == 4 for n in range(1, 100))                                                          
1
2
3
4
True
```

I can't understand what was wrong in my firsts attempts but I will assume that I'm not inspired and that's all :smile:

PR #4637, opened 2020-06-18
Typing: Tox env, CI job

elacuesta:
As discussed recently, here's a PR that contains only the Tox/CI changes, not any specific type annotations.

See also #4243.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,13 @@ passenv =
 commands =
     py.test --cov=scrapy --cov-report= {posargs:--durations=10 docs scrapy tests}
 
+[testenv:typing]
+basepython = python3
+deps =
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -23,6 +23,13 @@ passenv =
 commands =
     py.test --cov=scrapy --cov-report= {posargs:--durations=10 docs scrapy tests}
 
+[testenv:typing]
+basepython = python3
+deps =
+    mypy==0.780
```
elacuesta, line 29: Technically, last version is 0.781, but 0.780 is the last one to have an actual [release post](https://mypy-lang.blogspot.com/2020/06/mypy-0780-released.html) so I thought it might be safer.

PR #4640, opened 2020-06-22
Authorize each and every request for basic http authentication in HttpAuthMiddleware

AKSHAYSHARMAJS:
Hi @whalebot-helmsman 
I have made some changes for the problem that we discussed. These are not perfect changes but to explain my idea. Please suggest if approach is right or wrong.
I have used a dictionary that will be declared in spider by user for accessing every username and password for corresponding urls.
I have a doubts if i should use Redirect middleware(and it's disciples) if response is not authenticated or Retry Middleware(and it's disciples) or you suggest any other way.
Thanks!
Akshay

### DIFF
--- a/scrapy/downloadermiddlewares/httpauth.py
+++ b/scrapy/downloadermiddlewares/httpauth.py
@@ -6,6 +6,7 @@ See documentation in docs/topics/downloader-middleware.rst
 
 from w3lib.http import basic_auth_header
 
+from scrapy.http.response import Response
 from scrapy import signals
 
 
@@ -15,17 +16,28 @@ class HttpAuthMiddleware:
 
     @classmethod
     def from_crawler(cls, crawler):
-        o = cls()
+        o = cls()   
         crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
         return o
 
     def spider_opened(self, spider):
-        usr = getattr(spider, 'http_user', '')
-        pwd = getattr(spider, 'http_pass', '')
-        if usr or pwd:
-            self.auth = basic_auth_header(usr, pwd)
+        pass
+
+    def _authenticated(self, response):
+        if response.status == "401":
+            return False
+
+        return True
 
     def process_request(self, request, spider):
-        auth = getattr(self, 'auth', None)
-        if auth and b'Authorization' not in request.headers:
+        if _authenticated(self, Response(request.url)):
+            if request.url in spider.url_user_pwd.keys():
+                usr = spider.url_user_pwd[request.url][0] #url_user_pwd is a dictionary of urls with their corresponding username, password
+                pwd = spider.url_user_pwd[request.url][1]
+                if usr or pwd:
+                    self.auth = basic_auth_header(usr, pwd)
+
+            auth = getattr(self, 'auth', None)
             request.headers[b'Authorization'] = auth
+
+        # here redirect or retry middleware can be used if request needs authentication


wRAR:
We discussed this some time ago and found that we want the ability to pass this auth data via command line, it's easy with the current setup but not when the data is a dict.

AKSHAYSHARMAJS:
Should I use another way of passing auth data?. Basically I m working on problem where httpauth middleware is authorizing all requests. It should support per-request switching off like other middlewares. Can you please suggest something on it.

scrapy/downloadermiddlewares/httpauth.py
```diff
@@ -15,17 +16,28 @@ class HttpAuthMiddleware:
...
 
     def spider_opened(self, spider):
-        usr = getattr(spider, 'http_user', '')
-        pwd = getattr(spider, 'http_pass', '')
-        if usr or pwd:
-            self.auth = basic_auth_header(usr, pwd)
+        pass
+
+    def _authenticated(self, response):
+        if response.status == "401":
```
elacuesta, line 27: Status codes are integers, not strings.

scrapy/downloadermiddlewares/httpauth.py
```diff
@@ -15,17 +16,28 @@ class HttpAuthMiddleware:
...
+            return False
+
+        return True
 
     def process_request(self, request, spider):
-        auth = getattr(self, 'auth', None)
-        if auth and b'Authorization' not in request.headers:
+        if _authenticated(self, Response(request.url)):
```
elacuesta, line 33: This condition will always be true, because the response is created here without a status, hence its status will be 200 by default.

scrapy/downloadermiddlewares/httpauth.py
```diff
@@ -15,17 +16,28 @@ class HttpAuthMiddleware:
...
+            if request.url in spider.url_user_pwd.keys():
```
elacuesta, line 34: Should this be checking the domain instead? With this code, slightly different URLs that point to the same resource will need duplicated entries. Also, the `keys()` call is not necessary.

scrapy/downloadermiddlewares/httpauth.py
```diff
@@ -15,17 +16,28 @@ class HttpAuthMiddleware:
...
+        o = cls()   
         crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
         return o
 
     def spider_opened(self, spider):
-        usr = getattr(spider, 'http_user', '')
-        pwd = getattr(spider, 'http_pass', '')
-        if usr or pwd:
-            self.auth = basic_auth_header(usr, pwd)
+        pass
```
elacuesta, line 24: No need for this method to be here if it does nothing.

scrapy/downloadermiddlewares/httpauth.py
AKSHAYSHARMAJS: Not sure but I can check if its working with domains.

scrapy/downloadermiddlewares/httpauth.py
AKSHAYSHARMAJS: Okay I will remove it then.

scrapy/downloadermiddlewares/httpauth.py
AKSHAYSHARMAJS: Should I check for response or only just request headers? I believe middleware should retry authentication 1 or 2 times if the authorization fails.

scrapy/downloadermiddlewares/httpauth.py
AKSHAYSHARMAJS: oh yes. I did it by mistake. 🤦

AKSHAYSHARMAJS:
> 
> 
> I'm not sure I understand the motivation for this change. The current implementation allows to authenticate all requests, by setting the `Spider.http_user` and `Spider.http_pass` attributes. Is the issue that you need different credentials for different sites within the same spider?

Yes, that's what I 'm trying to achieve. I think for current Implementation its okay for authorizing all request when username and password matches but for different urls middleware fails to authorize.

AKSHAYSHARMAJS:
I still have a doubt to use `dict` here for auth data. Should I continue with it or you suggest something else?

PR #4642, opened 2020-06-22
Docs: add note about dataclass items and loaders

elacuesta:
I only commented this as part of the description of #3881, but it should be documented in the actual docs.

### DIFF
--- a/docs/topics/loaders.rst
+++ b/docs/topics/loaders.rst
@@ -26,7 +26,7 @@ Using Item Loaders to populate items
 To use an Item Loader, you must first instantiate it. You can either
 instantiate it with an :ref:`item object <topics-items>` or without one, in which
 case an instance of :class:`~scrapy.item.Item` is automatically created in the
-Item Loader ``__init__`` method using the :class:`~scrapy.item.Item` subclass
+Item Loader ``__init__`` method using the :ref:`item <topics-items>` class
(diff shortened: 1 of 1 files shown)


docs/topics/loaders.rst
```diff
@@ -26,7 +26,7 @@ Using Item Loaders to populate items
 To use an Item Loader, you must first instantiate it. You can either
 instantiate it with an :ref:`item object <topics-items>` or without one, in which
 case an instance of :class:`~scrapy.item.Item` is automatically created in the
```
Gallaecio, line 28: ?

```suggestion
case an :ref:`item object <topics-items>` is automatically created in the
```

docs/topics/loaders.rst
```diff
@@ -76,6 +76,41 @@ called which actually returns the item populated with the data
...
 
+
+.. _topics-loaders-dataclass:
+
+Working with dataclass items
+============================
+
+By default, :ref:`dataclass items <dataclass-items>` require all fields to be
+passed when created. This could be an issue when using dataclass items with
+item loaders, since fields could be populated incrementally.
```
Gallaecio, line 87: Is “could be” the right wording? Isn’t this always a given (unless you pass a pre-initialized item object to the item loader `__init__` method?)

PR #4646, opened 2020-06-24
allowing to run .pyw files 

### LINKED ISSUE #4643: Allow run pyw scripts
scrapy/commands/runspider.py check for Python script source but it fails to allow .pyw files.
Check at row 14 is:
`    if fext != '.py':`
but it should be:
`    if fext != '.py' and fext != '.pyw':`

AKSHAYSHARMAJS:
Referenced to issue #4643. 
Also looking for other parts of code base for the mentioned issue.

Fixes #4643


scrapy/commands/runspider.py
```diff
@@ -11,7 +11,7 @@ def _import_file(filepath):
     abspath = os.path.abspath(filepath)
     dirname, file = os.path.split(abspath)
     fname, fext = os.path.splitext(file)
-    if fext != '.py':
+    if fext != '.py' and fext != '.pyw':
```
Gallaecio, line 14: :lipstick: 

```suggestion
    if fext not in ('.py', '.pyw'):
```

AKSHAYSHARMAJS:
> 
> 
> Could you include a test?

Okay, I will add some test.

AKSHAYSHARMAJS:
C:\Users\Akshay\Desktop\git\scrapy\scrapy\commands\edit.py:`
   36  
   37          sfile = sys.modules[spidercls.__module__].__file__
   38:         **sfile = sfile.replace('.pyc', '.py')**
   39          self.exitcode = os.system('%s "%s"' % (editor, sfile))
   40  

C:\Users\Akshay\Desktop\git\scrapy\scrapy\commands\genspider.py:
   99              spiders_module = None
  100              spiders_dir = "."
  101:         **spider_file = "%s.py" % join(spiders_dir, module)**
  102          shutil.copyfile(template_file, spider_file)
  103          render_templatefile(spider_file, **tvars)


C:\Users\Akshay\Desktop\git\scrapy\scrapy\commands\startproject.py:
   23: **IGNORE = ignore_patterns('*.pyc', '.svn')**


I searched code base and these are the concerned results. I don't think they require .pyw?

scrapy/commands/runspider.py
AKSHAYSHARMAJS: 👍

AKSHAYSHARMAJS:
image

I m getting this error and not able to test above changes.

wRAR:
@AKSHAYSHARMAJS try removing your .tox dir, this helps when the project requirements have changed.

AKSHAYSHARMAJS:
> 
> 
> @AKSHAYSHARMAJS try removing your .tox dir, this helps when the project requirements have change

Thanks! @wRAR

AKSHAYSHARMAJS:
Should I make test for `.pyw` files windows specific that will only run in windows. I tried mocking windows but '.pyw' doesn't support in linux. giving error (No module named 'myspider.pyw')

tests/test_commands.py
```diff
@@ -273,31 +298,52 @@ class MySpider(scrapy.Spider):
...
         log1 = self.get_log(self.debug_log_spider,
                             args=('-s', 'LOG_SHORT_NAMES=1'))
-        print(log1)
         self.assertIn("[myspider] DEBUG: It Works!", log1)
         self.assertIn("[scrapy]", log1)
         self.assertNotIn("[scrapy.core.engine]", log1)
 
         log2 = self.get_log(self.debug_log_spider,
                             args=('-s', 'LOG_SHORT_NAMES=0'))
-        print(log2)
```
AKSHAYSHARMAJS, line 292: why we need these print statements?

wRAR: I don't think we need them, they can be useful when finding why a test fails but in that case you can add them locally

AKSHAYSHARMAJS: Alright, I have already removed them. :)

Gallaecio:
I think we should refactor these tests.

I’m thinking we could have a base class that defines the tests and reads the filename (`myspider.py` or `myspider.pyw`) from a class variable, and then have 2 subclasses that simply define that variable, each with one of the values. See `HttpProxyTestCase` and its subclasses for an example of what I mean.

I would also have the tests that will also work out of Windows run in other OSes. Mostly because at the moment or CI does not run Windows tests. For the remaining tests, we can use `skipif` to skip them outside Windows.

AKSHAYSHARMAJS:
I can try this implementation.

AKSHAYSHARMAJS:
Please check, I just used RunSpiderCommandTest as base class for `.py` and subclass WindowsRunSpiderCommandTest  for '.pyw'.

tests/test_commands.py
```diff
@@ -321,74 +303,53 @@ def test_runspider_log_short_names(self):
...
-            log = self.get_log("from scrapy.spiders import Spider\n", name='myspider.pyw')
-            self.assertIn("No spider found in file", log)
-
     def test_runspider_file_not_found(self):
         _, _, log = self.proc('runspider', 'some_non_existent_file')
         self.assertIn("File not found: some_non_existent_file", log)
 
     def test_runspider_unable_to_load(self):
-        log = self.get_log('', name='myspider.txt')
+        self.spider_filename = 'myspider.txt'
```
Gallaecio, line 315: I think it would be better not to modify `self.spider_filename` as part of this test. Maybe you can define `self.get_log` so that it only uses `self.spider_filename` if the `name` is `None` (default), so that it can accept this name as a parameter. Or maybe, since it does not make sense to run this test twice (for `.py` and for `.pyw`), maybe this test should be in a separate class, on its own.

tests/test_commands.py
```diff
@@ -330,6 +334,24 @@ def test_asyncio_enabled_false(self):
         self.assertNotIn("Using reactor: twisted.internet.asyncioreactor.AsyncioSelectorReactor", log)
 
 
+class WindowsRunSpiderCommandTest(RunSpiderCommandTest):
+
+    spider_filename = 'myspider.pyw'
+
+    def setUp(self):
+        if platform != 'win32':
+            raise unittest.SkipTest(".pyw file extension only supported in Windows")
```
Gallaecio, line 343: Do all tests fail in Linux? I think we should try to run as many of them as possible also in Linux, at least until we have Windows support in the CI.

AKSHAYSHARMAJS: I checked it now 6 of them are failing maybe I should run the rest for linux too.

tests/test_commands.py
```diff
@@ -339,10 +339,27 @@ class WindowsRunSpiderCommandTest(RunSpiderCommandTest):
...
+
+        def test_runspider_log_level(self):
+            raise unittest.SkipTest("Windows required")
+
+        def test_runspider_log_short_names(self):
+            raise unittest.SkipTest("Windows required")
+
+        def test_runspider_no_spider_found(self):
+            raise unittest.SkipTest("Windows required")
+
```
Gallaecio, line 362: I think it would be better to use `@skipif` instead, in each method, and have their implementation be `super().<method_name>()`.

In any case, do all of those test fail on Linux?

AKSHAYSHARMAJS: > I think it would be better to use `@skipif` instead, in each method, and have their implementation be `super().<method_name>()`.

For that I have to import `unittest` separately as `twisted.trial.unittest` doesn't have `@skipif`

AKSHAYSHARMAJS: > In any case, do all of those test fail on Linux?

I don't think there is any case for it.

Gallaecio: If the tests are going to be Windows-only then, maybe this should wait until #4458 is merged. I’m hoping to be done with that soon™.

AKSHAYSHARMAJS: That will be nice, lets wait then.

AKSHAYSHARMAJS: Should I use `@skipIf` or current implementation is fine?

Gallaecio: I think you should use a decorator, `if` statements for method definition seem rather unidiomatic.

I wonder if you can use a skipping decorator or similar at the class level, though.

AKSHAYSHARMAJS:
Some more tests have been added that cause test failure.
Only  these two are **WindowsRunSpiderCommandTest.test_output** and **WindowsRunSpiderCommandTest.test_overwrite_output** so I should skip them for unix platforms.

kmike:
Nice, thanks @AKSHAYSHARMAJS!

AKSHAYSHARMAJS:
😃 😃

PR #4649, opened 2020-06-27
Support schema-less HTTP proxy (#4504)

### LINKED ISSUE #4505: Using proxy through http fails (https works)

ajaymittur28:
Fixes #4505 

Implemented the solution suggested by @Gallaecio [in this PR](https://github.com/scrapy/scrapy/pull/4536#discussion_r420784056).


tests/test_downloader_handlers.py
```diff
@@ -733,6 +735,16 @@ def test_download_with_proxy_https_timeout(self):
         timeout = yield self.assertFailure(d, error.TimeoutError)
         self.assertIn(domain, timeout.osError)
 
+    def test_download_with_proxy_without_http_scheme(self):
+        def _test(response):
+            self.assertEqual(response.status, 200)
+            self.assertEqual(response.url, request.url)
+            self.assertEqual(response.body, b'http://example.com')
+
+        http_proxy = self.getURL('').replace('http:', '')
```
Gallaecio, line 744: Shouldn’t we test without the `//`? Or at least test both ways?

```suggestion
        http_proxy = self.getURL('').replace('http://', '')
```

Gallaecio:
Other than that minor change, I’ve verified that the test correctly covers the underlying issue, so good work!

tests/test_downloader_handlers.py
ajaymittur28: So I did try this out as well, and the test actually doesn't give the correct result. I looked into it and found that while running the test with `http_proxy = self.getURL('').replace('http://', '')` this [line 19](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/webclient.py#L19) in `webclient` throws an exception indicating that it wasn't able to extract the hostname. It seems like the reason is `urllib.parse.urlparse` implementation

```
>>> parsed = urlparse('test-uri.com')
>>> parsed.hostname
>>> parsed.hostname == None
True
parsed = urlparse('//test-uri.com')
>>> parsed.hostname
'test-uri.com'
```

Gallaecio: Then we need to change the implementation to support `test-uri.com`. It would be great to also support `//test-uri.com`, but I would consider that optional, given that users are much more likely to use `test-uri.com`.

ajaymittur28: It actually works fine on the user-end, I created the following spider and ran it. 

```
import scrapy


    name = 'example'

    def start_requests(self):
        proxies = ["example-proxy.com:9999", "//example-proxy.com:9999", "http://example-proxy.com:9999"]
        for proxy in proxies:
            yield scrapy.Request(url='http://quotes.toscrape.com/page/1/', callback=self.parse, meta={"proxy": proxy})

    def parse(self, response):
        print('Finished Successfully')
```

The `httpproxy` middleware is able to parse the proxy url correctly because of [these 2 lines](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpproxy.py#L32-L33) (_urlunparse()_ seems to adds the '//'). 

Not sure if the request made in the unit test passes through this middleware, but I think the fix should be as simple as checking if the url contains a scheme and if not adding a '//' to the front in `webclient._parse`. What do you think?

Gallaecio: > I think the fix should be as simple as checking if the url contains a scheme and if not adding a '//' to the front in webclient._parse. What do you think?

Sounds good to me.

ajaymittur28: Looks like it's not as easy as it sounds, tried a couple of fixes:- 

1. checking `urlparse(url).scheme` and prepending  '//' if missing
2. checking if `urlparse(url).netloc` was identified and prepending '//' if missing
 
and was unsuccessful. From the [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html?highlight=urlsplit#urllib.parse.urlparse) docs:-

> Following the syntax specifications in [RFC 1808](https://tools.ietf.org/html/rfc1808.html), urlparse recognizes a netloc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component.

`urlparse` thus produces a few different results for various types of URIs (without schema) making it hard to introduce a simple 'fix'

```
>>> urlparse('127.0.0.1:11586')
ParseResult(scheme='', netloc='', path='127.0.0.1:11586', params='', query='', fragment='')
>>> urlparse('127.0.0.1:11586/')
ParseResult(scheme='127.0.0.1', netloc='', path='11586/', params='', query='', fragment='')
>>> urlparse('://127.0.0.1:11586/')
ParseResult(scheme='', netloc='', path='://127.0.0.1:11586/', params='', query='', fragment='')
>>> urlparse('://127.0.0.1:11586')
ParseResult(scheme='', netloc='', path='://127.0.0.1:11586', params='', query='', fragment='')
>>> urlparse('//127.0.0.1:11586')
ParseResult(scheme='', netloc='127.0.0.1:11586', path='', params='', query='', fragment='')
>>> urlparse('//127.0.0.1:11586/')
ParseResult(scheme='', netloc='127.0.0.1:11586', path='/', params='', query='', fragment='')
>>> urlparse('example-proxy.com')
ParseResult(scheme='', netloc='', path='example-proxy.com', params='', query='', fragment='')
>>> urlparse('example-proxy.com:3333')
ParseResult(scheme='', netloc='', path='example-proxy.com:3333', params='', query='', fragment='')
>>> urlparse('example-proxy.com:3333/')
ParseResult(scheme='example-proxy.com', netloc='', path='3333/', params='', query='', fragment='')
>>> urlparse('://example-proxy.com:3333/')
ParseResult(scheme='', netloc='', path='://example-proxy.com:3333/', params='', query='', fragment='')
>>> urlparse('://example-proxy.com:3333')
ParseResult(scheme='', netloc='', path='://example-proxy.com:3333', params='', query='', fragment='')
>>> urlparse('//example-proxy.com:3333')
ParseResult(scheme='', netloc='example-proxy.com:3333', path='', params='', query='', fragment='')
```

I personally think it would be easier for the developer to enforce the standard quoted above ([RFC 1808](https://tools.ietf.org/html/rfc1808.html)) while writing the test, otherwise (if necessary) we could open a new issue to work on it.

Gallaecio: The original issue is that users enter the domain name without a protocol, so we cannot expect them to add `//`.

What about prepending `//` if the URL does not match `^\w+://`?

ajaymittur28: This issue doesn't seem to exist when a user creates a spider and runs it, only seems to exist while writing unit tests

> It actually works fine on the user-end, I created the following spider and ran it.
> 
> ```
> import scrapy
> 
> 
> class ExampleSpider(scrapy.Spider):
>     name = 'example'
> 
>     def start_requests(self):
>         proxies = ["example-proxy.com:9999", "//example-proxy.com:9999", "http://example-proxy.com:9999"]
>         for proxy in proxies:
>             yield scrapy.Request(url='http://quotes.toscrape.com/page/1/', callback=self.parse, meta={"proxy": proxy})
> 
>     def parse(self, response):
>         print('Finished Successfully')
> ```
> 
> The `httpproxy` middleware is able to parse the proxy url correctly in [these 2 lines](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpproxy.py#L32-L33) (found while debugging that `urlunparse` prepends the '//' while unparsing the `_parse_proxy` result).

Will try out your suggestion, Thanks

Gallaecio: I’m guessing the reason it happens in the test is because it happens with `localhost` as proxy. You could argue that `localhost` being the proxy is uncommon, but still I think we should support it, specially if it’s trivial to support.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -319,9 +319,13 @@ def _get_agent(self, request, timeout):
                     pool=self._pool,
                 )
             else:
+                proxyScheme = b'http' if not proxyScheme else proxyScheme
+                proxyHost = to_bytes(proxyHost, encoding='ascii')
+                proxyPort = to_bytes(str(proxyPort), encoding='ascii')
+                proxyURI = urlunparse((proxyScheme, proxyNetloc, proxyParams, '', '', ''))
                 return self._ProxyAgent(
                     reactor=reactor,
-                    proxyURI=to_bytes(proxy, encoding='ascii'),
```
ajaymittur28, line 324: Had to construct `proxyURI` and pass it to `_ProxyAgent` since `proxy` is the schemaless url with the incorrect format (not starting with `//`)

scrapy/core/downloader/webclient.py
```diff
@@ -1,5 +1,6 @@
 from time import time
 from urllib.parse import urlparse, urlunparse, urldefrag
+from re import match
```
Gallaecio, line 3: :lipstick: `re` is usually imported as a whole. But in any case, it should still be the first import in the file, to keep things alphabetical.

scrapy/core/downloader/handlers/http11.py
Gallaecio: If it works, :100:

scrapy/core/downloader/webclient.py
ajaymittur28: Right, wasn't aware, thanks!

scrapy/core/downloader/handlers/http11.py
```diff
@@ -319,9 +319,13 @@ def _get_agent(self, request, timeout):
                     pool=self._pool,
                 )
             else:
+                proxyScheme = b'http' if not proxyScheme else proxyScheme
```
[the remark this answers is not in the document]
ajaymittur28: Updated 👍

wRAR:
Thank you!

PR #4661, opened 2020-07-06
Review unicode-related remnants of Python 2

### LINKED ISSUE #4547: Review unicode references

Gallaecio:
- Update documentation references to “unicode” as a string type (fixes #4547)
- Deprecate `to_bytes` and `to_unicode` in favor of Python’s `bytes` and `str`
- Remove the `u` prefix from string literals

### DIFF
--- a/docs/topics/dynamic-content.rst
+++ b/docs/topics/dynamic-content.rst
@@ -206,7 +206,7 @@ data from it:
     >>> import lxml.etree
     >>> from parsel import Selector
     >>> javascript = response.css('script::text').get()
-    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='utf-8')
+    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='unicode')
     >>> selector = Selector(text=xml)
     >>> selector.css('var[name="data"]').get()
     '<var name="data"><object><property name="field"><string>value</string></property></object></var>'
--- a/docs/topics/dynamic-content.rst
+++ b/docs/topics/dynamic-content.rst
@@ -206,7 +206,7 @@ data from it:
     >>> import lxml.etree
     >>> from parsel import Selector
     >>> javascript = response.css('script::text').get()
-    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='unicode')
+    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='utf-8')
     >>> selector = Selector(text=xml)
     >>> selector.css('var[name="data"]').get()
     '<var name="data"><object><property name="field"><string>value</string></property></object></var>'
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -51,12 +51,12 @@ Request objects
        given, the dict passed in this parameter will be shallow copied.
     :type meta: dict
 
-    :param body: the request body. If a ``unicode`` is passed, then it's encoded to
-      ``str`` using the ``encoding`` passed (which defaults to ``utf-8``). If
-      ``body`` is not given, an empty string is stored. Regardless of the
-      type of this argument, the final value stored will be a ``str`` (never
-      ``unicode`` or ``None``).
-    :type body: str or unicode
+    :param body: the request body. If a string is passed, then it's encoded as
+      bytes using the ``encoding`` passed (which defaults to ``utf-8``). If
+      ``body`` is not given, an empty bytes object is stored. Regardless of the
+      type of this argument, the final value stored will be a bytes object
+      (never a string or ``None``).
+    :type body: bytes
 
     :param headers: the headers of this request. The dict values can be strings
        (for single valued headers) or lists (for multi-valued headers). If
@@ -106,7 +106,7 @@ Request objects
 
     :param encoding: the encoding of this request (defaults to ``'utf-8'``).
        This encoding will be used to percent-encode the URL and to convert the
-       body to ``str`` (if given as ``unicode``).
+       body to bytes (if given as a string).
     :type encoding: string
 
     :param priority: the priority of this request (defaults to ``0``).
@@ -721,7 +721,7 @@ Response objects
     .. attribute:: Response.body
 
         The body of this Response. Keep in mind that Response.body
-        is always a bytes object. If you want the unicode version use
+        is always a bytes object. If you want the string version use
         :attr:`TextResponse.text` (only available in :class:`TextResponse`
         and subclasses).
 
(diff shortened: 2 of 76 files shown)


docs/topics/dynamic-content.rst
```diff
@@ -206,7 +206,7 @@ data from it:
     >>> import lxml.etree
     >>> from parsel import Selector
     >>> javascript = response.css('script::text').get()
-    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='unicode')
+    >>> xml = lxml.etree.tostring(js2xml.parse(javascript), encoding='utf-8')
```
kmike, line 209: if I'm not mistaken, the return type would be different; encoding='unicode' makes output unicode, while encoding='utf8' makes output utf8-encoded string (i.e. bytes).

docs/topics/request-response.rst
```diff
@@ -51,12 +51,12 @@ Request objects
...
-      ``body`` is not given, an empty string is stored. Regardless of the
-      type of this argument, the final value stored will be a ``str`` (never
-      ``unicode`` or ``None``).
-    :type body: str or unicode
+    :param body: the request body. If a string is passed, then it's encoded as
+      bytes using the ``encoding`` passed (which defaults to ``utf-8``). If
+      ``body`` is not given, an empty bytes object is stored. Regardless of the
+      type of this argument, the final value stored will be a bytes object
+      (never a string or ``None``).
+    :type body: bytes
```
elacuesta, line 59:
```suggestion
    :type body: bytes or str
```
I think this would be more accurate. This final attribute will always be a `bytes` object, but the parameter can be a string (the description mentions it already).

elacuesta:
There are a few places in which `scrapy.utils.python.to_bytes` and `scrapy.utils.python.to_to_unicode` are replaced by `bytes` and `str` without encoding, I think this might produce unexpected results because [`to_bytes`](https://github.com/scrapy/scrapy/blob/2.2.0/scrapy/utils/python.py#L106-L107) and [`to_unicode`](https://github.com/scrapy/scrapy/blob/2.2.0/scrapy/utils/python.py#L93-L94) use `utf-8` if no encoding is passed:
```
>>> from scrapy.utils.python import to_bytes, to_unicode
>>>
>>> to_bytes("á")
b'\xc3\xa1'
>>> bytes("á")
TypeError: string argument without an encoding
>>>
>>> to_unicode(b'\xc3\xa1')
'á'
>>> str(b'\xc3\xa1')
"b'\\xc3\\xa1'"
>>>
```

docs/topics/dynamic-content.rst
wRAR: Yes, this is correct.

scrapy/utils/iterators.py
```diff
@@ -44,7 +44,7 @@ def xmliter_lxml(obj, nodename, namespace=None, prefix='x'):
     iterable = etree.iterparse(reader, tag=tag, encoding=reader.encoding)
     selxpath = '//' + ('%s:%s' % (prefix, nodename) if namespace else nodename)
     for _, node in iterable:
-        nodetext = etree.tostring(node, encoding='unicode')
+        nodetext = etree.tostring(node, encoding='utf-8')
```
wRAR, line 47: `Selector` calls `.encode('utf-8')` on the text attribute in `create_root_node()`, so looks like this is not needed and may be harmful?

wRAR: (see also the comment for docs/topics/dynamic-content.rst)

tests/test_commands.py
```diff
@@ -67,7 +67,7 @@ def proc(self, *new_args, **popen_kwargs):
 
         def kill_proc():
             p.kill()
-            assert False, 'Command took too much time to complete'
+            assert False, 'Command took too mcipher_stringuch time to complete'
```
wRAR, line 74: typo

docs/topics/request-response.rst
```diff
@@ -853,16 +853,16 @@ TextResponse objects
...
        result is cached after the first call, so you can access
        ``response.text`` multiple times without extra overhead.
 
        .. note::
 
-            ``unicode(response.body)`` is not a correct way to convert response
-            body to unicode: you would be using the system default encoding
+            ``str(response.body)`` is not a correct way to convert the response
+            body into a string: you would be using the system default encoding
             (typically ``ascii``) instead of the response encoding.
```
kmike, line 866: The reason is different now:

```
>>> str(b'body')
"b'body'"
```

scrapy/core/downloader/handlers/ftp.py
```diff
@@ -106,7 +105,7 @@ def _build_response(self, result, request, protocol):
         protocol.close()
         body = protocol.filename or protocol.body.read()
         headers = {"local filename": protocol.filename or '', "size": protocol.size}
-        return respcls(url=request.url, status=200, body=to_bytes(body), headers=headers)
+        return respcls(url=request.url, status=200, body=bytes(body, 'utf-8'), headers=headers)
```
kmike, line 108: why not `body.encode('utf-8')`? Maybe it is just me, but I had to look up what the second argument of bytes does. If was thinking you're doing so to passthrough body if body is already bytes, but this is not what `bytes(body, 'utf-8')` does.

Also, was to_bytes necessary before, i.e. were there cases when body can be bytes, not unicode?

scrapy/exporters.py
```diff
@@ -324,14 +324,14 @@ def _serialize_value(self, value):
             return dict(self._serialize_item(value))
         elif is_listlike(value):
             return [self._serialize_value(v) for v in value]
-        encode_func = to_bytes if self.binary else to_unicode
+        encode_func = bytes if self.binary else str
```
kmike, line 327: I don't think `bytes` works the same way as to_bytes for binary data. I.e. if value is bytes, previously it would have been passed through, but now it raises an exception. The same applies for str.

scrapy/extensions/httpcache.py
```diff
@@ -366,7 +365,7 @@ def parse_cachecontrol(header):
 
 def rfc1123_to_epoch(date_str):
     try:
-        date_str = to_unicode(date_str, encoding='ascii')
+        date_str = str(date_str, encoding='ascii')
```
kmike, line 368: this would fail if date_str is already str, so technically this is a backwards incompatible change

docs/topics/request-response.rst
Gallaecio: I’m having a dejavù

scrapy/core/downloader/handlers/ftp.py
Gallaecio: > Also, was to_bytes necessary before, i.e. were there cases when body can be bytes, not unicode?

Yes, it looks like protocol.filename is user-defined and likely to be a string, but protocol.body is always bytes, and either can end up as the value for `body`.

Gallaecio:
I originally thought `to_bytes` and `to_unicode` were Python 2 things. But it’s becoming clear that they have a purpose valid for Python 3, which is to convert when needed and raise` TypeError` if the input is not `bytes` or `str`. So:

- What about keeping them, simply renaming `to_unicode` as `to_str` (through deprecation, of course)?
- Shall we still switch to `bytes` and `str` where the input type is known beforehand?

PR #4676, opened 2020-07-13
Fix startproject failing for existing folders

### LINKED ISSUE #4665: startproject only allows creating the project on an existing folder if the path is specified with an invalid module name
```
(venv) [adrian@afonsox temporal]$ mkdir existing_folder
(venv) [adrian@afonsox temporal]$ scrapy startproject existing_folder
Error: Module 'existing_folder' already exists
(venv) [adrian@afonsox temporal]$ cd existing_folder && scrapy startproject existing_folder .
New Scrapy project 'existing_folder', using template directory '/home/adrian/temporal/venv/lib/python3.8/site-packages/scrapy/templates/project', created in:
    /home/adrian/temporal/existing_folder

    cd .
```

Gallaecio:
Fixes #4665

The issue was that empty directories in an import path can be imported as namespace packages since Python 3.3: https://stackoverflow.com/a/49082183/939364

I’ve switched `import_module` to `find_spec`, which does something similar without actually importing, and now we are also checking whether `spec.loader` is `None` (indicating it’s a namespace package).

I suspect this has been happening forever when using Python 3.

### DIFF
--- a/scrapy/commands/startproject.py
+++ b/scrapy/commands/startproject.py
@@ -1,7 +1,7 @@
 import re
 import os
 import string
-from importlib import import_module
+from importlib.util import find_spec
 from os.path import join, exists, abspath
 from shutil import ignore_patterns, move, copy2, copystat
 from stat import S_IWUSR as OWNER_WRITE_PERMISSION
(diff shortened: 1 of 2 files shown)


wRAR:
What's the reason for prints in the tests? Pytest eats that by default I think?

Gallaecio:
Pytest eats them when tests pass, but I believe it shows them when tests fail, and they could come in handy then.

That said, I’m OK with removing them.

wRAR:
In that case it's OK to leave them I think. Thanks!

wRAR (approved):

PR #4682, opened 2020-07-16
remove python 2 reminiscence in cookies

noviluni:
The `getheaders` method was implemented to be compatible with Python 2, but as we dropped the support for Python 2 we don't need it anymore. It's enough to delete it or do you prefer to add a deprecation warning before doing it?


kmike:
A good catch @noviluni. I think we don't need a deprecation here, because it is Python stdlib who calls this method, not Scrapy users.

PR #4683, opened 2020-07-16
deprecate retry_on_eintr

noviluni:
This function is not used, is not tested and is not documented, so we should probably remove it.

It was introduced to be used in `test_commands.py` but is not used anymore (ref: https://github.com/scrapy/scrapy/commit/6217f108cce8a7db0c0ee3682848159b06c9e9d3)

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -285,6 +285,7 @@ class WeakKeyCache:
         return self._weakdict[key]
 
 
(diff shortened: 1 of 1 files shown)


GGXXLL:
Scrapyd-client use this function in scrapyd-deploy?

Gallaecio:
In that case they can copy it into their code.

PR #4686, opened 2020-07-18
Support defining file path based on item in media pipelines

### LINKED ISSUE #4628: Support choosing a file path based on item data in media pipelines
https://github.com/scrapy/scrapy/blob/8b549392f924ddad9536e55c6120638daf688dfd/scrapy/pipelines/media.py#L77-L111

### LINKED ISSUE #4677: Pass item in FilesPipeline method
## Summary

One paragraph explanation of the feature.
When called the method to define the path where the file should be stored it would be nice if the method receive the item as parameter. Then the method can dynamically define where the item should be stored based on the item properties.

## Motivation

I was working in the https://github.com/okfn-brasil/diario-oficial and we would like to store the item by its date property. But there is not way to do that without a hack. To workaround the issue we need to create a subclass of the request and add the item on it. Thus, we can get the item from the request object later in the method to define the path.

ajaymittur28:
Closes #4628, closes #4677 

The existing media pipelines (`FilesPipeline` and `ImagesPipeline`) do not allow the user to dynamically define downloaded file location based on the item returned by the spider, thus, forcing the users to use other workarounds like using the `Request.meta` to pass item data or creating their own `Request` subclass with the item passed to it.

I've updated the media pipelines to pass `item` as a parameter to the `file_path` function and have updated the same in the tests as well. This will now allow the user to define file paths based on the item.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -409,7 +409,7 @@ class FilesPipeline(MediaPipeline):
         store_cls = self.STORE_SCHEMES[scheme]
         return store_cls(uri)
 
-    def media_to_download(self, request, info, item):
+    def media_to_download(self, request, info, item=None):
         def _onsuccess(result):
             if not result:
                 return  # returning None force download
@@ -460,7 +460,7 @@ class FilesPipeline(MediaPipeline):
 
         raise FileException
 
-    def media_downloaded(self, response, request, info, item):
+    def media_downloaded(self, response, request, info, item=None):
         referer = referer_str(request)
 
         if response.status != 200:
@@ -493,7 +493,7 @@ class FilesPipeline(MediaPipeline):
 
         try:
             path = self.file_path(request, response=response, info=info, item=item)
-            checksum = self.file_downloaded(response, request, info, item)
+            checksum = self.file_downloaded(response, request, info, item=item)
         except FileException as exc:
             logger.warning(
                 'File (error): Error processing file from %(request)s '
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -409,7 +409,7 @@ class FilesPipeline(MediaPipeline):
         store_cls = self.STORE_SCHEMES[scheme]
         return store_cls(uri)
 
-    def media_to_download(self, request, info, item=None):
+    def media_to_download(self, request, info, *, item=None):
         def _onsuccess(result):
             if not result:
                 return  # returning None force download
@@ -460,7 +460,7 @@ class FilesPipeline(MediaPipeline):
 
         raise FileException
 
-    def media_downloaded(self, response, request, info, item=None):
+    def media_downloaded(self, response, request, info, *, item=None):
         referer = referer_str(request)
 
         if response.status != 200:
@@ -522,7 +522,7 @@ class FilesPipeline(MediaPipeline):
         urls = ItemAdapter(item).get(self.files_urls_field, [])
         return [Request(u) for u in urls]
 
-    def file_downloaded(self, response, request, info, item=None):
+    def file_downloaded(self, response, request, info, *, item=None):
         path = self.file_path(request, response=response, info=info, item=item)
         buf = BytesIO(response.body)
         checksum = md5sum(buf)
@@ -535,7 +535,7 @@ class FilesPipeline(MediaPipeline):
             ItemAdapter(item)[self.files_result_field] = [x for ok, x in results if ok]
         return item
 
-    def file_path(self, request, response=None, info=None, item=None):
+    def file_path(self, request, response=None, info=None, *, item=None):
         media_guid = hashlib.sha1(to_bytes(request.url)).hexdigest()
         media_ext = os.path.splitext(request.url)[1]
         # Handles empty and wild extensions by trying to guess the
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -13,8 +13,10 @@ from collections import defaultdict
 from contextlib import suppress
 from email.utils import mktime_tz, parsedate_tz
 from ftplib import FTP
+from inspect import signature
 from io import BytesIO
 from urllib.parse import urlparse
+from warnings import warn
 
 from itemadapter import ItemAdapter
 from twisted.internet import defer, threads
@@ -25,6 +27,7 @@ from scrapy.pipelines.media import MediaPipeline
 from scrapy.settings import Settings
 from scrapy.utils.boto import is_botocore
 from scrapy.utils.datatypes import CaselessDict
+from scrapy.utils.deprecate import ScrapyDeprecationWarning
 from scrapy.utils.ftp import ftp_store_file
 from scrapy.utils.log import failure_to_exc_info
 from scrapy.utils.misc import md5sum
@@ -357,6 +360,15 @@ class FilesPipeline(MediaPipeline):
         if isinstance(settings, dict) or settings is None:
             settings = Settings(settings)
 
+        # Check if file_path used by user is deprecated
+        file_path_sig = signature(self.file_path)
+        try:
+            file_path_sig.parameters['item']
+        except KeyError:
+            warn('file_path(self, request, response=None, info=None) is deprecated, '
+                 'please use file_path(self, request, response=None, info=None, item=None)',
+                 ScrapyDeprecationWarning, stacklevel=2)
+
         cls_name = "FilesPipeline"
         self.store = self._get_store(store_uri)
         resolve = functools.partial(self._key_for_pipe,
@@ -436,7 +448,7 @@ class FilesPipeline(MediaPipeline):
             checksum = result.get('checksum', None)
             return {'url': request.url, 'path': path, 'checksum': checksum, 'status': 'uptodate'}
 
-        path = self.file_path(request, info=info, item=item)
+        path = self._file_path(request, info=info, item=item)
         dfd = defer.maybeDeferred(self.store.stat_file, path, info)
         dfd.addCallbacks(_onsuccess, lambda _: None)
         dfd.addErrback(
@@ -492,7 +504,7 @@ class FilesPipeline(MediaPipeline):
         self.inc_stats(info.spider, status)
 
         try:
-            path = self.file_path(request, response=response, info=info, item=item)
+            path = self._file_path(request, response=response, info=info, item=item)
             checksum = self.file_downloaded(response, request, info, item)
         except FileException as exc:
             logger.warning(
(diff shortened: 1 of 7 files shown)


wRAR:
I would leave these tests as is and added a new test, or several if needed, to check that the item is indeed now available in the method.

ajaymittur28:
I've added a test to ensure the file path can now be generated from the item passed to `file_path`, and have also added the `item` parameter to `media_to_download` and `media_downloaded` methods of `test_pipeline_media` just to make sure the parameters are consistent with the updated interfaces in `pipelines.media`. Their implementation hasn't been changed.

Gallaecio:
@ajaymittur28 You have done a great work so far.

However, before we merge these changes, we should address backward incompatibility. If a subclass is not using the new argument in its `file_path` method signature, things should still work as before. And in that case a deprecation warning should be logged.

If you look for `inspect` usage in the code, you should be able to find some examples on how to do this. I believe even in related code we are already doing something like this.

ajaymittur28:
I can check if the user is using the deprecated version of `file_path` method or not using `inspect` and log a deprecation warning. I however couldn't think of a way to make the `file_path` method the user may be using compatible with the new item parameter, other than using `try...except` blocks wherever the call to `file_path` is made and decide whether to pass `item` or not. Do you think there is a better way to do this?

ajaymittur28:
I've added a wrapper method `_file_path` in `MediaPipeline` to pass parameters consistent with the user defined `file_path` method in the subclass and address the backwards incompatibility issue. I also added a deprecation warning for the method if the user defined method is using the previous signature. I am however unsure of the deprecation message and if I should place them in `FilesPipeline` and `ImagesPipeline` like I have currently or in their parent class `MediaPipeline`. Could you have a look? @Gallaecio

scrapy/pipelines/files.py
```diff
@@ -357,6 +360,15 @@ def __init__(self, store_uri, download_func=None, settings=None):
...
             settings = Settings(settings)
 
+        # Check if file_path used by user is deprecated
+        file_path_sig = signature(self.file_path)
+        try:
+            file_path_sig.parameters['item']
+        except KeyError:
+            warn('file_path(self, request, response=None, info=None) is deprecated, '
+                 'please use file_path(self, request, response=None, info=None, item=None)',
+                 ScrapyDeprecationWarning, stacklevel=2)
```
Gallaecio, line 370: :lipstick: `if 'item' not in file_path_sig.parameters:` would be simpler.

tests/test_pipeline_files.py
```diff
@@ -174,6 +174,18 @@ def file_path(self, request, response=None, info=None, item=None):
...
 
+    def test_file_path_backwards_compatibility(self):
+        """
+        Test file_path method without `item` parameter in its signature
+        """
+        class CustomFilesPipeline(FilesPipeline):
+            def file_path(self, request, response=None, info=None):
+                return 'full/%s' % 'path'
+
+        file_path = CustomFilesPipeline.from_settings(Settings({'FILES_STORE': self.tempdir})).file_path
```
Gallaecio, line 185: It would be great to also validate that the warning is issued.

scrapy/pipelines/images.py
```diff
@@ -51,6 +54,15 @@ def __init__(self, store_uri, download_func=None, settings=None):
...
 
+        # Check if file_path used by user is deprecated
+        file_path_sig = signature(self.file_path)
+        try:
+            file_path_sig.parameters['item']
+        except KeyError:
+            warn('file_path(self, request, response=None, info=None) is deprecated, '
+                 'please use file_path(self, request, response=None, info=None, item=None)',
+                 ScrapyDeprecationWarning, stacklevel=2)
+
```
Gallaecio, line 65: I would move it to the parent class.

I’m thinking:

- It would avoid this duplication.

- It would allow to set a private class variable (e.g. `_file_path_expects_item = True/False) which could be used by `_file_path`, to avoid 1 call when `item` is not expected, and to only perform a call without `item` when needed (currently an unrelated `TypeError` in a `file_path` implementation could trigger that call as well).

ajaymittur28: I was unsure about doing this in the parent class since it doesn't have any declaration of `file_path` or any usage other than `_file_path`, should I add an empty declaration of `file_path` under "Overridable Interfaces" or is it not needed?

scrapy/pipelines/media.py
```diff
@@ -207,3 +219,7 @@ def item_completed(self, results, item, info):
                         extra={'spider': info.spider}
                     )
         return item
+
+    def file_path(self, request, response=None, info=None, item=None):
+        """Returns the path where downloaded media should be stored"""
+        pass
```
Gallaecio, line 225: Maybe `raise NotImplementedError`, just in case someone tries to use this class directly?

ajaymittur28:
According to the [docs](https://docs.scrapy.org/en/latest/topics/media-pipeline.html#scrapy.pipelines.files.FilesPipeline) it looks like only `file_path` out of the overridable interfaces is exposed to the user to modify, so I don't think the backward compatibility issue can occur in the other methods?
I've updated the signatures of the methods you have mentioned to pass item to `file_path`, they work with both the current and new signature of `file_path`

scrapy/pipelines/media.py
ajaymittur28: We could do this, but `file_path` or `_file_path` is **not called anywhere** in the `MediaPipeline`, it is only called in its subclasses `ImagesPipeline` and `FilesPipeline` where `file_path` is implemented, so even if a user overrides or uses this class directly `file_path` will not be called. It looks like this class is not meant to be used directly by users according to the [docs](https://docs.scrapy.org/en/latest/topics/media-pipeline.html) and I'm guessing that's why its not implemented here. 

However if we raise an exception here, I think we should raise the exception in `media_to_download` and `get_media_requests` as well since they are also not implemented?

Gallaecio:
Those methods are indeed undocumented, but they are not private, and there’s a chance subclasses of the file or image classes are overriding them.

Also mind that we documented `file_path` in 1.7.0 because an issue about it was opened: https://github.com/scrapy/scrapy/issues/2253. But it has existed for a while; those other methods should probably be documented as well eventually.

ajaymittur28:
Right, makes sense, thanks for clarifying!

I'm thinking we can then generalize the `file_path` wrapper `_file_path` for the other overridable interfaces whose signatures have been modified which can be done easily using `try....except` block like how I had done it initially but this has the drawback of making 1 extra call each time, however I feel using a flag for each of the methods and comparing the flag with corresponding method may be a bit cumbersome.

How do you think we should go about it?

Gallaecio:
Maybe a flag dictionary, where keys are the names of the methods which may or may not expect `item`?

ajaymittur28:
`setattr` did the job, thanks!

I replaced the individual tests I made with a test case using a mocked `ImagesPipeline` class since it allows us to test all the methods you mentioned without having to create individual cases for each by simulating the pipeline and lets us ensure the methods were indeed called. I had to override `get_media_requests` and `inc_stats` to prevent the pipeline from failing since `crawler` is `None` while running the test.

I also realized that `item` may be passed as a positional argument and hence `kwargs.pop('item', None)` may not be enough.
To handle this I've added a check on the last parameter of `args` and determine if it should be removed or not. But, this may cause errors in the future if more parameters are added to the methods and that's why I've made (and passed) `item` as a keyworded argument. (`item` can be handled irrespective of position)

What do you think?

scrapy/pipelines/media.py
```diff
@@ -141,9 +140,11 @@ def _compatible(self, func):
         def wrapper(*args, **kwargs):
             if self._expects_item[func.__name__]:
                 return func(*args, **kwargs)
-            else:
+            elif 'item' in kwargs:
                 kwargs.pop('item', None)
-                return func(*args, **kwargs)
+            elif is_item(args[-1]):
+                args = args[:-1]
```
Gallaecio, line 146: I’m not sure about these 2 lines. `is_item` does not return `True` when the input _is_ an item, only when it _may be_ one.

What if we add `item` as a [keyword-only parameter](https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments) to those methods? Then we only need to worry about `kwargs` again.

scrapy/pipelines/media.py
Gallaecio: Fair points.

scrapy/pipelines/media.py
ajaymittur28: Yep we could do that, it would remove the uncertainty around `item` being passed as a positional/keyworded argument and make handling much easier! Thanks for the link. 
Will update the docs and warning message as well!

scrapy/pipelines/media.py
```diff
@@ -102,34 +109,74 @@ def _process_request(self, request, info):
...
+        ]
+
+        for method_name in methods:
+            method = getattr(self, method_name, None)
+            if callable(method):
+                setattr(self, method_name, self._compatible(method))
+
+    def _compatible(self, func):
+        """Wrapper for overridable methods to allow backwards compatibility"""
+        if func.__name__ not in self._expects_item:
```
Gallaecio, line 134: :lipstick: Looking at the coverage data, it looks like this `if` statement always evaluates to `True`, which seems right looking at the implementation method above. So maybe you can remove it.

scrapy/pipelines/media.py
```diff
@@ -1,12 +1,15 @@
 import functools
+from inspect import signature
 import logging
 from collections import defaultdict
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.python.failure import Failure
+from warnings import warn
```
wRAR, line 7: I think this import should be before the Twisted ones? @Gallaecio you are the expert on this :)

scrapy/pipelines/media.py
ajaymittur28: Yep you're right, I think I had added it initially as I had expected to call `_compatible` multiple times, but clearly that wasn't needed.

scrapy/pipelines/media.py
ajaymittur28: Hmm, I think it does! Maybe I could also add an empty line between standard library imports and Twisted imports as specified in the [PEP8 guidelines](https://www.python.org/dev/peps/pep-0008/#:~:text=You%20should%20put%20a%20blank%20line%20between%20each%20group%20of%20imports.&text=Standard%20library%20code%20should%20avoid,been%20removed%20in%20Python%203.)?. What say @Gallaecio?

Gallaecio: And `inspect` should be among other standard library `from` imports (`right after `from collections …`), we group those right after straight imports (without an empty line, though).

PR #4687, opened 2020-07-20
Update form.py

ANT0x1:
Ability to PUT form

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -29,7 +29,7 @@ class FormRequest(Request):
         if formdata:
             items = formdata.items() if isinstance(formdata, dict) else formdata
             querystr = _urlencode(items, self.encoding)
-            if self.method == 'POST':
+            if self.method == 'POST' or self.method == 'PUT':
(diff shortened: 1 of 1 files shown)


Gallaecio (commented):
While it [seems](https://stackoverflow.com/a/6306552/939364) like the change makes sense, it is also backward-incompatible. And I’m not 100% sure how to address that.

On the other hand, if we are improving `FormRequest` for `PUT`, I wonder if we should also update `valid_form_methods` accordingly. I was actually hoping that the fact that `PUT` is not in there would mean that adding `PUT` support is not backward-incompatible, but it seems like `valid_form_methods` is only taken into account by `from_response`.

AdrianAtZyte:
Closing after the move to [form2request](https://form2request.readthedocs.io/en/latest/usage.html) as a [replacement for `FormRequest`](https://github.com/scrapy/scrapy/pull/6438).

If you believe this issue still impacts `form2request` and should be addressed, please open an issue [upstream](https://github.com/scrapy/form2request/issues) with a link to this issue for reference.

PR #4691, opened 2020-07-22
Typing: annotate a few Spider attributes

elacuesta:
Not really a comprehensive annotation, just a few attributes to remove several files from the ignore list.

### DIFF
--- a/scrapy/spiders/__init__.py
+++ b/scrapy/spiders/__init__.py
@@ -5,6 +5,7 @@ See documentation in docs/topics/spiders.rst
 """
 import logging
 import warnings
+from typing import Optional
 
 from scrapy import signals
 from scrapy.http import Request
@@ -18,8 +19,8 @@ class Spider(object_ref):
     class.
     """
 
-    name = None
-    custom_settings = None
+    name: Optional[str] = None
+    custom_settings: Optional[dict] = None
 
     def __init__(self, name=None, **kwargs):
         if name is not None:
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -7,6 +7,7 @@ See documentation in docs/topics/spiders.rst
 
 import copy
 import warnings
+from typing import Sequence
 
 from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.http import Request, HtmlResponse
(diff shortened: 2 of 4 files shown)


tests/spiders.py
```diff
@@ -255,7 +255,7 @@ class CrawlSpiderWithParseMethod(MockServerSpider, CrawlSpider):
     A CrawlSpider which overrides the 'parse' method
     """
     name = 'crawl_spider_with_parse_method'
-    custom_settings = {
+    custom_settings: dict = {
```
elacuesta, line 258: I don't really know why this is necessary, since `custom_settings` is annotated in the base class :shrug:

scrapy/spiders/__init__.py
```diff
@@ -18,8 +19,8 @@ class Spider(object_ref):
     class.
     """
 
-    name = None
-    custom_settings = None
+    name: Optional[str] = None
```
Gallaecio, line 22: This seems to fail in Python 3.5

elacuesta: Indeed, it's not supported, that's why the PR is under the Python 3.6+ milestone.

Gallaecio: :facepalm:

Serves me right for reviewing draft pull requests.

scrapy/spiders/crawl.py
```diff
@@ -72,7 +73,7 @@ def _process_request(self, request, response):
 
 class CrawlSpider(Spider):
 
-    rules = ()
+    rules: Union[List[Rule], List, Tuple[Rule], Tuple] = ()
```
elacuesta, line 76: I don't like having `List/Tuple` here, but `Incompatible types in assignment (expression has type "Tuple[]", variable has type "Union[List[Rule], Tuple[Rule]]")` occurs otherwise. I think it would be much more reasonable for `mypy` to consider an empty tuplr as a valid object.
An alternative would be `rules: Union[List[Rule], Tuple[Rule]] = ()  # type: ignore`, but that would mean the type hint is not actually being evaluated, it'd be there just for documentation purposes. Any thoughts?

See also https://www.python.org/dev/peps/pep-0484/#type-comments

Gallaecio: From what I’m reading, maybe `List[Rule, ...]` could replace `List[Rule], List`.

Gallaecio: I wonder if we should use `Sequence[Rule, ...]` here.

elacuesta: Awesome, `rules: Sequence[Rule] = ()` works, many thanks!

kmike:

PR #4701, opened 2020-07-29
Code cleanup scrapy.utils.python.WeakKeyCache #4684

### LINKED ISSUE #4684: Should `scrapy.utils.python.WeakKeyCache` be deprecated?
## Summary
I can't find any reference to `scrapy.utils.python.WeakKeyCache` (except in tests) and I think that it should be deprecated.

## Motivation
Keeping non-used and non-documented code makes the codebase harder to maintain.

As it's tested we could keep it, but I don't think it's a good idea.

kshitijcode:
Resolves #4684

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -6,13 +6,13 @@ import gc
 import inspect
 import re
 import sys
-import weakref
 import warnings
+import weakref
 from functools import partial, wraps
 from itertools import chain
 
-from scrapy.utils.decorators import deprecated
 from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.utils.decorators import deprecated
 
 
 def flatten(x):
@@ -280,7 +280,7 @@ def equal_attributes(obj1, obj2, attributes):
 class WeakKeyCache:
 
     def __init__(self, default_factory):
-        warnings.warn("Call to deprecated Class WeakKeyCache", category=ScrapyDeprecationWarning, stacklevel=2)
+        warnings.warn("The WeakKeyCache class is deprecated", category=ScrapyDeprecationWarning, stacklevel=2)
         self.default_factory = default_factory
         self._weakdict = weakref.WeakKeyDictionary()
 
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -127,6 +127,7 @@ def re_rsearch(pattern, text, chunk_size=1024):
     In case the pattern wasn't found, None is returned, otherwise it returns a tuple containing
     the start position of the match, and the ending (regarding the entire text).
     """
+
     def _chunk_iter():
         offset = len(text)
         while True:
@@ -158,6 +159,7 @@ def memoizemethod_noargs(method):
         if self not in cache:
             cache[self] = method(self, *args, **kwargs)
         return cache[self]
+
     return new_method
 
 
@@ -271,6 +273,19 @@ def equal_attributes(obj1, obj2, attributes):
             return False
     # all attributes equal
     return True
+
+
+@deprecated
+class WeakKeyCache:
+
+    def __init__(self, default_factory):
+        self.default_factory = default_factory
+        self._weakdict = weakref.WeakKeyDictionary()
+
+    def __getitem__(self, key):
+        if key not in self._weakdict:
+            self._weakdict[key] = self.default_factory(key)
+        return self._weakdict[key]
 
 
 @deprecated
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -273,18 +273,6 @@ def equal_attributes(obj1, obj2, attributes):
     return True
 
 
-class WeakKeyCache:
-
-    def __init__(self, default_factory):
-        self.default_factory = default_factory
-        self._weakdict = weakref.WeakKeyDictionary()
-
-    def __getitem__(self, key):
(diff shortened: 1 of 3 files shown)


kshitijcode:
Cleaning up scrapy.utils.python.WeakKeyCache code and tests as per  #4684

wRAR:
Flake8 failed, can you please fix that?

kshitijcode:
@wRAR : The failures don't seem to be related to my change

can you confirm, please?

wRAR:
Indeed.

@Gallaecio are these some known intermittent failures?

kshitijcode:
@wRAR : Can I try pushing in an empty commit ?

kshitijcode:
@wRAR The indeed seem to be intermittent errors. Pushing an empty commit fixed it.

wRAR:
I've re-read the original issue, it talks about deprecating the class, not removing it, we don't remove non-deprecated classes/functions most of the time, can you please mark it as deprecated instead?

kshitijcode:
@wRAR : Would we like to bring back the test as well ?

wRAR:
No, I think it's fine to not test it as it's currently not used.

kshitijcode:
@wRAR @Gallaecio 

Made the changes as requested 👍

scrapy/utils/python.py
```diff
@@ -273,6 +275,7 @@ def equal_attributes(obj1, obj2, attributes):
     return True
 
 
+@deprecated
```
wRAR, line 278: I don't think it works on classes? Have you tried to instantiate it?

kshitijcode: @wRAR  Yep, this is working with classes as inherently we are calling the __init__ function .

I tested it out and it gave me the following warning : 

UserWarning: Call to deprecated function WeakKeyCache. Use The following API is deprecated instead.

Gallaecio: Well, the message is inaccurate, “deprecated function WeakKeyCache”. We need to either make `@deprecated` smart enough to say ‘class’ when needed, or use a different approach. We’ve been using the latter so far, I’m not sure if there are potential unexpected side effects of using `@deprecated` after changing it to say “class” for classes.

Gallaecio:
> No, I think it's fine to not test it as it's currently not used.

Actually, since the point of deprecating instead of removing is for things to remain working as expected for a while, I think it is better for the test to stay until we remove the class itself.

Also, it would be great to include a test that verifies that using the class logs the expected deprecation warning, or modify the existing test accordingly.

scrapy/utils/python.py
kshitijcode: @Gallaecio What is the latter approach which has been used so far?

kshitijcode:
@Gallaecio @wRAR : Do we follow the principle of like deprecate in release and remove in subsequent release?

Gallaecio:
> @Gallaecio @wRAR : Do we follow the principle of like deprecate in release and remove in subsequent release?

Nope. We only recently discussed internally our deprecation policy, and [we are adding it to the documentation soon](https://github.com/scrapy/scrapy/pull/4705): deprecated code continues working for at least 1 year.

scrapy/utils/python.py
Gallaecio: > @Gallaecio @wRAR What is the latter approach which has been used so far?

Logging a warning manually in the `__init__` method, with `ScrapyDeprecationWarning` as warning class, and `stacklevel=2` (so that the warning is about the call to `__init__`, and not about the implementation line where we log the warning).

> Also which is the existing test function?

I did not get this, sorry. Could you elaborate?

kshitijcode:
@wRAR @Gallaecio I went ahead with later change as there doesn't seem to be reliable way to differentiate between a class/method. 

Kindly review.

scrapy/utils/python.py
```diff
@@ -7,10 +7,12 @@
 import re
 import sys
 import weakref
+import warnings
```
Gallaecio, line 10: :lipstick: 

```suggestion
import warnings
import weakref
```

scrapy/utils/python.py
```diff
@@ -7,10 +7,12 @@
...
 from functools import partial, wraps
 from itertools import chain
 
 from scrapy.utils.decorators import deprecated
+from scrapy.exceptions import ScrapyDeprecationWarning
```
Gallaecio, line 15: :lipstick: 

```suggestion
from scrapy.exceptions import ScrapyDeprecationWarning
from scrapy.utils.decorators import deprecated
```

scrapy/utils/python.py
```diff
@@ -276,6 +280,7 @@ def equal_attributes(obj1, obj2, attributes):
 class WeakKeyCache:
 
     def __init__(self, default_factory):
+        warnings.warn("Call to deprecated Class WeakKeyCache", category=ScrapyDeprecationWarning, stacklevel=2)
```
Gallaecio, line 283: “Call” feels weird to me when talking about classes, and “class” should be lowercase.

:lipstick: What about…

```suggestion
        warnings.warn("The WeakKeyCache class is deprecated",
                      category=ScrapyDeprecationWarning, stacklevel=2)
```

kshitijcode: @Gallaecio resolved.

wRAR:
Thank you!

PR #4707, opened 2020-07-30
change super() syntax

noviluni:
As a way to modernize the code, I changed the `super()` syntax to the Python 3 `super()` syntax. I think this syntax it’s easier to read and write and looks cleaner.

### DIFF
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -134,7 +134,7 @@ class DbmStorageWithCustomDbmModuleTest(DbmStorageTest):
 
     def _get_settings(self, **new_settings):
         new_settings.setdefault('HTTPCACHE_DBM_MODULE', self.dbm_module)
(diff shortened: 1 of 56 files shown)


tests/test_downloadermiddleware_httpcache.py
```diff
@@ -151,7 +151,7 @@ class FilesystemStorageGzipTest(FilesystemStorageTest):
 
     def _get_settings(self, **new_settings):
         new_settings.setdefault('HTTPCACHE_GZIP', True)
-        return super(FilesystemStorageTest, self)._get_settings(**new_settings)
+        return super()._get_settings(**new_settings)
```
noviluni, line 154: As you can see, originally, this `super()` was referring to the `FilesystemStorageTest` superclass instead of `FilesystemStorageGzipTest`, but it's the same (`FilesystemStorageTest` doesn't implement its own `._get_settings()` method) so I decided to keep it as simply `super()`.

PR #4708, opened 2020-07-30
Cover Scrapy 2.2.1 and 2.3 in the release notes

Gallaecio:
[See rendered](https://scrapy-gallaecio.readthedocs.io/en/release-notes-2.3/news.html).

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,16 @@
 Release notes
 =============
 
+.. _release-2.2.1:
+
+Scrapy 2.2.1 (2020-07-17)
+-------------------------
+
+*   The :command:`startproject` command no longer makes unintended changes to
+    the permissions of files in the destination folder, such as removing
+    execution permissions (:issue:`4662`, :issue:`4666`)
+
+
 .. _release-2.2.0:
 
 Scrapy 2.2.0 (2020-06-24)
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,120 @@
 Release notes
 =============
 
+.. _release-2.3.0:
+
+Scrapy 2.3.0 (2020-0?-??)
+-------------------------
+
+Highlights:
+
+* :ref:`Feed exports <topics-feed-exports>` support :ref:`Google Cloud
+  Storage <topics-feed-storage-gcs>`
+* New :setting:`FEED_EXPORT_BATCH_ITEM_COUNT` setting for batch deliveries
+* The base implementation of :ref:`item loaders <topics-loaders>` has been
+  moved into :doc:`itemloaders <itemloaders:index>`
+
+
+Deprecation removals
+~~~~~~~~~~~~~~~~~~~~
+
+*   Removed the following classes and its parent modules from
+    ``scrapy.linkextractors``:
+
+    *   ``htmlparser.HtmlParserLinkExtractor``
+    *   ``regex.RegexLinkExtractor``
+    *   ``sgml.BaseSgmlLinkExtractor``
+    *   ``sgml.SgmlLinkExtractor``
+
+    Use
+    :class:`LinkExtractor <scrapy.linkextractors.lxmlhtml.LxmlLinkExtractor>`
+    instead (:issue:`4356`, :issue:`4679`)
+
+
+Deprecations
+~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


docs/news.rst
```diff
@@ -3,6 +3,130 @@
...
+  Storage <topics-feed-storage-gcs>`
+* New :setting:`FEED_EXPORT_BATCH_ITEM_COUNT` setting for batch deliveries
+* The base implementation of :ref:`item loaders <topics-loaders>` has been
+  moved into :doc:`itemloaders <itemloaders:index>`
+
+
+Deprecation removals
+~~~~~~~~~~~~~~~~~~~~
+
+*   Removed the following classes and its parent modules from
```
wRAR, line 23: Their?

Gallaecio: Great catch!

docs/news.rst
```diff
@@ -3,6 +3,132 @@
...
+.. _release-2.3.0:
+
+Scrapy 2.3.0 (2020-0?-??)
+-------------------------
+
+Highlights:
+
+* :ref:`Feed exports <topics-feed-exports>` support :ref:`Google Cloud
+  Storage <topics-feed-storage-gcs>`
+* New :setting:`FEED_EXPORT_BATCH_ITEM_COUNT` setting for batch deliveries
```
kmike, line 15: What do you think about explaining this a bit more? E.g. "previously, Scrapy's Feed Exports were exporting items to S3 (and ...?) at the end of the crawl. With FEED_EXPORT_BATCH_ITEM_COUNT option it is now possible to export items to cloud providers during the crawl, in batches of a specified size."

docs/news.rst
```diff
@@ -3,6 +3,132 @@
...
+* The base implementation of :ref:`item loaders <topics-loaders>` has been
+  moved into :doc:`itemloaders <itemloaders:index>`
```
kmike, line 17: This is again, a style preference, but I wonder if it is better to have a more detailed description of this feature in the highlights,to  highlight what is this change for. The same text as below can be used ("he base implementation of item loaders has been moved into a separate library, itemloaders, allowing usage from outside Scrapy and a separate release schedule")

docs/news.rst
```diff
@@ -3,6 +3,132 @@
...
+*   A ``parse`` callback may now be used in built-in spider subclasses, such
+    as :class:`~scrapy.spiders.CrawlSpider` (:issue:`712`, :issue:`732`,
+    :issue:`781`, :issue:`4254` )
+
+
+Bug fixes
+~~~~~~~~~
+
+*   Fixed the :ref:`CSV exporting <topics-feed-format-csv>` of
+    :ref:`dataclass items <dataclass-items>` (:issue:`4667`, :issue:`4668`)
```
kmike, line 70: it also fixes export of attr.s items, if I'm not mistaken

Gallaecio:
I’ve tried putting the long descriptions in the highlights and the short ones in the sections below. I thought about removing them from the sections below, but I think that may be confusing in some scenarios for users reading diagonally. I also though about duplicating the texts, which may be what you had in mind to begin with, and would work for me.

About the purpose of `FEED_EXPORT_BATCH_ITEM_COUNT`, I first extended the entry to describe the problem it can solve in detail, and eventually I thought that this information should be available in the regular documentation, so I created a section about this aspect of some storage backends (I called it “delayed file delivery”, alternative name suggestions are welcome), included a reference to it in the section about each of the affected storage backends, and shortened the release notes entry.

@kmike Please, have a new look when you get a chance.

docs/topics/feed-exports.rst
```diff
@@ -227,6 +241,26 @@ The feeds are written to the standard output of the Scrapy process.
...
 
+.. _delayed-file-delivery:
+
+Delayed file delivery
+---------------------
+
+As indicated above, some of the described storage backeds use delayed file
+delivery.
+
+These storage backeds do not receive items as they are scraped. Instead, Scrapy
```
kmike, line 252: This is slighhtly ambiguous: I'd consider storage backend to be a Python class which implements delivery to a particular location. These Python classes do receive items as thy are scraped.

(there is also a small typo - backeds instead of backends)

kmike:
@Gallaecio great update for the docs, thanks! Highlights section looks awesome to me.

PR #4714, opened 2020-08-05
Remove duplicated install_requires

linchiwei123:


wRAR:
Why is it extra? I only see another declaration under `if has_environment_marker_platform_impl_support():`

Gallaecio:
It _is_ extra, _my_ bad :facepalm: 

When defining it in `extras_require`, it should not be in `install_requires` as well. And, in any case, in the scenario where the `append` happens, you would end up with a duplicate.

wRAR:
Oh, I missed the `else` branch. Thanks.

PR #4716, opened 2020-08-06
IOError and other cleanup

noviluni:
From Python 3.3 the `IOError` is just an alias for OSError and is kept for compatibility reasons.

reference: https://docs.python.org/3/whatsnew/3.3.html#pep-3151-reworking-the-os-and-io-exception-hierarchy

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -33,12 +33,12 @@ logger = logging.getLogger(__name__)
 
 class RetryMiddleware:
 
-    # IOError is raised by the HttpCompression middleware when trying to
+    # OSError is raised by the HttpCompression middleware when trying to
     # decompress an empty response
     EXCEPTIONS_TO_RETRY = (defer.TimeoutError, TimeoutError, DNSLookupError,
                            ConnectionRefusedError, ConnectionDone, ConnectError,
                            ConnectionLost, TCPTimedOutError, ResponseFailed,
-                           IOError, TunnelError)
+                           OSError, TunnelError)
 
     def __init__(self, settings):
         if not settings.getbool('RETRY_ENABLED'):
--- a/scrapy/pipelines/__init__.py
+++ b/scrapy/pipelines/__init__.py
@@ -18,7 +18,7 @@ class ItemPipelineManager(MiddlewareManager):
         return build_component_list(settings.getwithbase('ITEM_PIPELINES'))
 
     def _add_middleware(self, pipe):
-        super(ItemPipelineManager, self)._add_middleware(pipe)
+        super()._add_middleware(pipe)
         if hasattr(pipe, 'process_item'):
             self.methods['process_item'].append(deferred_f_from_coro_f(pipe.process_item))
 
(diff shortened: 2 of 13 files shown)


scrapy/pipelines/__init__.py
```diff
@@ -18,7 +18,7 @@ def _get_mwlist_from_settings(cls, settings):
         return build_component_list(settings.getwithbase('ITEM_PIPELINES'))
 
     def _add_middleware(self, pipe):
-        super(ItemPipelineManager, self)._add_middleware(pipe)
+        super()._add_middleware(pipe)
```
noviluni, line 21: I missed this in this PR: https://github.com/scrapy/scrapy/pull/4707

tests/test_downloadermiddleware.py
```diff
@@ -71,7 +71,7 @@ def test_3xx_and_invalid_gzipped_body_must_redirect(self):
         In particular when some website returns a 30x response with header
         'Content-Encoding: gzip' giving as result the error below:
 
-            exceptions.IOError: Not a gzipped file
+            BadGzipFile: Not a gzipped file (...)
```
noviluni, line 74: To reproduce it, you can do:

```python
import gzip
import io

buff = io.BytesIO(b'aaaa') 
buff.seek(0) 
gz = gzip.GzipFile(fileobj=buff) 
gz.read()

File: scrapy/downloadermiddlewares/retry.py [hunk 3]
```diff
@@ -33,12 +33,12 @@
 
 class RetryMiddleware:
 
-    # IOError is raised by the HttpCompression middleware when trying to
+    # OSError is raised by the HttpCompression middleware when trying to
```
noviluni, line 36: Was this fixed? I'm not sure if the test changed here: https://github.com/scrapy/scrapy/pull/4716/files#r466436261 fixed this. 

Even if it's not fixed, do you think we should remove this comment or keep it?

Gallaecio: I think leaving the comment makes sense. I think it’s the only exception with a comment because it is the only exception for which it is not obvious why it belongs in `EXCEPTIONS_TO_RETRY`.

I do wonder if there’s any potential for backward incompatibility here, and whether we can address it or we can only document it. The best I can think of is a subclass which rewrites the tuple to remove `IOError` and leave all other exceptions, which could be implemented in a way that would fail to remove `OSError`, but I do not know how we could address that.

tests/test_downloadermiddleware.py
Gallaecio: Maybe the test should assert that `BadGzipFile` is raised, then?

It may make sense to add a comment which links this to the redirect middleware `EXCEPTIONS_TO_RETRY` attribute, and test that `BadGzipFile` is a subclass of `OSError`, just in case that changes in some future Python version, and we need to update `EXCEPTIONS_TO_RETRY` accordingly.

wRAR: Well, this test actually tests that the default middleware order prevents the httpcompression middleware from seeing this response.

Not sure if we have a test that a similar 200 response is handled by retrying via the httpcompression and retry middlewares working together.

wRAR: `test_200_and_invalid_gzipped_body_must_fail` (which is the next test here) handles the second case, and while the code raises `BadGzipFile` the tests expects `IOError` so I think we have everything we need, except maybe better comments.

wRAR (approved):

PR #4669, opened 2020-07-08
travis-ci: disable py 3.5.x, enable for http2 branch

adityaa30:

PR #4718, opened 2020-08-11
Upgrade minimum dependencies for Python 3.6 support

Gallaecio:
While working on HTTP/2 support, @adityaa30 [discovered](https://github.com/scrapy/scrapy/pull/4669) that simply upgrading the pinned Travis CI job to 3.6 would cause tests to fail.

[It turns out](https://github.com/scrapy/scrapy/pull/4681) we need to first upgrade some pinned dependencies for that to work.

This does so, in preparation for the upcoming removal of Python 3.5 support, but without actually dropping such support yet just in case.

### DIFF
--- a/tests/requirements-py3.txt
+++ b/tests/requirements-py3.txt
@@ -1,8 +1,7 @@
 # Tests requirements
 attrs
 dataclasses; python_version == '3.6'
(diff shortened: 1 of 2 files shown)


kmike:
Thanks @Gallaecio!

PR #4721, opened 2020-08-11
smarter generator check for combined return/yield statements

### LINKED ISSUE #4720: Smarter generator check for combined yield/return statements: ignore nested functions
## Summary
Currently, if spider methods are generators that yield results and contain nested function, then the following warning is issued:

```
[py.warnings] WARNING: /Library/Python/3.7/site-packages/scrapy/core/scraper.py:148: UserWarning: The "MySpider.parse" method is a generator and includes a "return" statement with a value different than None. This could lead to unexpected behaviour. Please see https://docs.python.org/3/reference/simple_stmts.html#the-return-statement for details about the semantics of the "return" statement within generators
```

The example of a simple spider that results in the warning:
```
import scrapy

    name = "MySpider"
    start_urls = ["https://scrapy.org"]
    
    def parse(self, response):
        
        def is_external(url):
            href = url.css('::attr(href)').get()
            return href.startswith('http') and 'scrapy.org' not in href
        
        links = [link for link in response.css('a') if is_external(link)]
        for link in links:
            yield {'link': link.css('::attr(href)').get(), 'text': link.css('::text').get()}
```

I know it's a bit artificial example as the nested function can be moved, but there is nothing wrong with nested function conceptually.

## Motivation

I have a midsize spider function that includes some nested helper functions that I'd like to keep close to where they are called.

Moving nested function out of the generator is an easy fix, but it constrains expressivity of the code.

## Additional context

Related function: is_generator_with_return_value

soid:
Resolves #4720

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -184,29 +184,23 @@ def set_environ(**kwargs):
                 os.environ[k] = v
 
 
-class GeneratorNodeVisitor(ast.NodeVisitor):
-    """Visits all AST nodes of a generator, but ignores functions defined inside the generator.
-    If the generator code includes a return statement then sets includes_return_statement to True.
+def walk_callable(node):
+    """Similar to ``ast.walk``, but walks only function body and skips nested
+    functions defined within the node.
     """
-    def __init__(self):
-        self.read_func_header = False
-        self.includes_return_statement = False
-
-    def visit_FunctionDef(self, node):
-        if not self.read_func_header:
-            # Visit the first FunctionDef and ignore all nested definitions
-            # because they are not part of the examined generator
-            self.read_func_header = True
-            ast.NodeVisitor.generic_visit(self, node)
-
-    def visit_Return(self, node):
-        if not GeneratorNodeVisitor._returns_none(node):
-            self.includes_return_statement = True
-
-    @staticmethod
-    def _returns_none(return_node):
-        value = return_node.value
-        return value is None or isinstance(value, ast.NameConstant) and value.value is None
+    from collections import deque
+    todo = deque([node])
+    walked_func_def = False
+    while todo:
+        node = todo.popleft()
+        if isinstance(node, ast.FunctionDef):
+            if walked_func_def:
+                # don't traverse nested functions
+                continue
+            # first function def is ok
+            walked_func_def = True
+        todo.extend(ast.iter_child_nodes(node))
+        yield node
 
 
 _generator_callbacks_cache = LocalWeakReferencedCache(limit=128)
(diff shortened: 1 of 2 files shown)


scrapy/utils/misc.py
```diff
@@ -184,6 +184,33 @@ def set_environ(**kwargs):
...
 
 
+class GeneratorNodeVisitor(ast.NodeVisitor):
+    """Visits all AST nodes of a generator, but ignores functions defined inside the generator.
+    If the generator code includes a return statement then sets includes_return_statement to True.
+    """
+    def __init__(self):
+        self.read_func_header = False
+        self.includes_return_statement = False
+        self.return_statement_lineno = None
```
wRAR, line 194: How is `return_statement_lineno` used?

elacuesta:
Thanks for this update. I considered the visitor approach in the original PR (#3869), but decided to go with `ast.walk` because of performance concerns. Is there a chance you could add a stress test to see if visiting all nodes impacts performance somehow?

soid:
@elacuesta good point, I briefly considered the importance of stopping, but mistakenly decided it would be too rare case with never executed code after return statement. In fact, return statement can be inside a condition, so it make sense to stop right away. Another source of slower performance could be the recursion used by the visitor. Previous false positives are expected to become slower – if before the function stopped after a return in a nested function, now it may run longer since it will traverse the generator function and not the nested functions.

So I rewrote the solution - essentially, I copied the ast.walk() function and made it skip nested function definition subtrees. It should be even faster since now it won't traverse nested functions.

I ran some micro-benchmarking and the performance looks comparable:
```
       new: walk_callable  new: visitor    current
count           20.000000     20.000000  20.000000
mean             5.188111      5.562837   5.215105
std              0.041617      0.064930   0.063022
min              5.130406      5.510167   5.130689
25%              5.155440      5.529315   5.181950
50%              5.177555      5.548093   5.200611
75%              5.223487      5.565969   5.234239
max              5.274709      5.805158   5.383350
```

(it is for 10K calls per sample, with caching removed, the code is here: soid/scrapy@f4f68db)
Considering that the function is cached I think it's insignificant differences.
Please let me know if you want me to run scrapy-bench as well.

scrapy/utils/misc.py
soid: Forgot to remove it, sorry. I was gonna add the line number to the error message, but then decided to limit the change.
I remove it now.

scrapy/utils/misc.py
```diff
@@ -184,6 +184,25 @@ def set_environ(**kwargs):
...
+    from collections import deque
+    todo = deque([node])
+    walked_func_def = False
+    while todo:
+        node = todo.popleft()
+        if isinstance(node, ast.FunctionDef):
+            if walked_func_def:
+                # don't traverse nested functions
+                continue
+            # first function def is ok
```
elacuesta, line 200:
```suggestion
                continue
```
These comments are unnecessary, I think their purpose is already explained the docstring.

scrapy/utils/misc.py
```diff
@@ -184,6 +184,25 @@ def set_environ(**kwargs):
                 os.environ[k] = v
 
 
+def walk_callable(node):
+    """Similar to ``ast.walk``, but walks only function body and skips nested
+    functions defined within the node.
+    """
+    from collections import deque
```
elacuesta, line 191: Could this import be moved to the top of the file?

soid: Done

PR #4723, opened 2020-08-13
test_utils_iterators.py: support Windows the right way

Gallaecio:
It should solve some issues caught by Conda Forge: https://github.com/conda-forge/scrapy-feedstock/pull/44

They were caught by Conda Forge because it uses the test files from the PyPI tarball, not from Git, so in Conda Forge the line endings of files are never changed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+tests/sample_data/** binary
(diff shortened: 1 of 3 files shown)


kmike:
TIL git does that. Thanks @Gallaecio!

PR #4725, opened 2020-08-13
Travis CI: Add a job with PYTHONOPTIMIZE=1

### LINKED ISSUE #4454: Add a CI job with PYTHONOPTIMIZE=2

Gallaecio:
To do:

-   [x] Avoid sending coverage data to codecov from the new job
-   [x] ~Add a test that should fail, and make sure it fails~ (not needed given the switch away from plain asserts)
-   [x] Test with both assert and self.assert*; we may need to replace all assert usages in tests for things to work as expected

Fixes #4454, closes #4459


wRAR:
> parsing methods must have __doc__ for pycparser to work properly

Gallaecio:
Will have to look into https://github.com/eliben/pycparser/pull/198

Gallaecio:
I guess we cannot support `PYTHONOPTIMIZE=2` then.

I’ve created https://github.com/scrapy/scrapy/issues/4737, to cover this limitation in the documentation when we write documentation about tuning memory usage (I don’t see a spot in the current documentation where it would make sense to cover this).

I’ll try with `PYTHONOPTIMIZE=1`.

Gallaecio:
That 2nd attempt at not sending data to codecov from the PYTHONOPTIMIZE job seems to work according to the logs, but the codecov report still reports changes that can only be explained by that data being sent.

docs/contributing.rst
```diff
@@ -267,6 +267,12 @@ And their unit-tests are in::
 
     tests/test_loader.py
 
+Even though ``pytest`` supports using ``assert`` statements, do not use them
+for Scrapy tests. One of our test environments runs Scrapy tests with
+``PYTHONOPTIMIZE=1``, which removes those statements, making tests that use
+those statements pointless in that test environment.
```
kmike, line 273: As a personal opinion, I'd not trade being able to use pytest asserts in tests for being able to automatically check PYTHONOPTIMIZE=1 case on every commit.

I'm not sure what are the practical reasons for the users to run scrapy under PYTHONOPTIMIZE=1. We could make fixes to ensure it works (if for some reason some user want it), and merge them to Scrapy, but a few minutes on CI on every commit + not being able to use asserts in tests - not sure this is justified.

Gallaecio: > I'm not sure what are the practical reasons for the users to run scrapy under PYTHONOPTIMIZE=1.

Scrapy may be used in an environment where `PYTHONOPTIMIZE` is also used for other package(s). But no, I don’t expect much usage of `PYTHONOPTIMIZE` with Scrapy. In Python in general, I think `PYTHONOPTIMIZE` is seldom used.

I’m OK with dropping this.

Issue #4726, opened 2020-08-14
Test failures on Debian Unstable


wRAR:
I ran the 2.3.0 test suite on the current Debian Unstable and got the following failures.

Two related to those CAMELLIA tests (`Https11CustomCiphers.test_download` and `WebClientCustomCiphersSSLTestCase.testPayload`):

```
<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'ssl3_read_bytes', 'tlsv1 alert internal error')]
```

Two related to the mitmproxy cert:

```
  File "/usr/lib/python3/dist-packages/mitmproxy/net/tls.py", line 369, in create_server_context
    context.use_certificate(cert.x509)
  File "/usr/lib/python3/dist-packages/OpenSSL/SSL.py", line 960, in use_certificate
    _raise_current_error()
  File "/usr/lib/python3/dist-packages/OpenSSL/_util.py", line 54, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.SSL.Error: [('SSL routines', 'SSL_CTX_use_certificate', 'ee key too small')]
```

PR #4727, opened 2020-08-14
Use a longer key in mitmproxy-ca.pem.

wRAR:
Fixes the second issue in #4726. Looks like the Debian OpenSSL hardening feature (`CipherString = DEFAULT@SECLEVEL=2`) is still not enabled in OpenSSL by default so the 1024 key in `tests/keys/mitmproxy-ca.pem` is usually accepted.

As mitmproxy generates its keys on a first start, I just copied the file from `~/.mitmproxy` instead of replicating the generation command.


kmike:
@wRAR could you please elaborate on this? 

> Looks like the Debian OpenSSL hardening feature (CipherString = DEFAULT@SECLEVEL=2) is still not enabled in OpenSSL by default so the 1024 key in tests/keys/mitmproxy-ca.pem is usually accepted.

Do you mean that this PR changes key length to 1024, but we'll need to change it again when `CipherString = DEFAULT@SECLEVEL=2` becomes default, or do you mean that current length is 1024, and it usually works because `CipherString = DEFAULT@SECLEVEL=2` is not a default?

wRAR:
This PR changes the key length from current 1024 to 2048, so that it works with both relaxed and strict requirements.

kmike:
Thanks @wRAR!

Issue #2607, opened 2017-03-01
add params kwarg to scrapy.Request()


pawelmhm:
python-requests have useful params kwargs, you can do

```python
params = {
    'expand': 'variations,informationBlocks,customisations',
}

response = requests.request(
    method='GET',
    url=url,
    headers=headers,
    params=params,
)
``

what do you think about adding params kwarg to scrapy,Request()? It would simplify work, there would be no need to urlencode querystring and concatenate strings for url.

ramwin:
Yes, I'm looking for this attribute either. I used to use params in `requests` lib, sometime, it is inconvenient and terrible to url encode the `<spance>, &, ?, %` yourself.

grammy-jiang:
Hi, @pawelmhm 

@leonardfrank provides a way to do this, and if you want to know more, please read [here](https://doc.scrapy.org/en/latest/topics/request-response.html#formrequest-objects).

I would like to provide another solution for this, the module coming with python: [21.8. urllib.parse — Parse URLs into components — Python 3.6.5 documentation](https://docs.python.org/3/library/urllib.parse.html). In this module there are several functions related to URL:

- urlparse
- urlunparse
- parse_qs
- parse_qsl
- urljoin

and many others. Going through this official document will help you a lot.

PR #4730, opened 2020-08-16
Add params to scrapy.Request()

bulatbulat48:
https://github.com/scrapy/scrapy/issues/2607

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -4,6 +4,8 @@ requests in Scrapy.
 
 See documentation in docs/topics/request-response.rst
 """
+from urllib.parse import urlencode
+
 from w3lib.url import safe_url_string
 
 from scrapy.http.headers import Headers
@@ -18,11 +20,12 @@ class Request(object_ref):
 
     def __init__(self, url, callback=None, method='GET', headers=None, body=None,
                  cookies=None, meta=None, encoding='utf-8', priority=0,
-                 dont_filter=False, errback=None, flags=None, cb_kwargs=None):
+                 dont_filter=False, errback=None, flags=None, cb_kwargs=None, params=None):
 
         self._encoding = encoding  # this one has to be set first
         self.method = str(method).upper()
         self._set_url(url)
+        self._set_params(params)
         self._set_body(body)
         if not isinstance(priority, int):
             raise TypeError("Request priority not an integer: %r" % priority)
@@ -54,6 +57,22 @@ class Request(object_ref):
         if self._meta is None:
             self._meta = {}
         return self._meta
+
+    def _get_params(self):
+        return self._params
+
+    def _set_params(self, params):
+        if params is None:
+            self._params = None
+            return
+        if isinstance(params, dict):
+            qs = urlencode(params, doseq=True)
+            self._set_url(self.url + ('&' if '?' in self.url else '?') + qs)
+            self._params = qs
+        else:
+            raise TypeError('Params must be dict, got {}'.format(type(params)))
+
+    params = property(_get_params, obsolete_setter(_set_params, 'params'))
 
     def _get_url(self):
         return self._url
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -33,7 +33,7 @@ class FormRequest(Request):
                 self.headers.setdefault(b'Content-Type', b'application/x-www-form-urlencoded')
                 self._set_body(querystr)
             else:
-                self._set_url(self.url + ('&' if '?' in self.url else '?') + querystr)
+                self._set_params(formdata)
 
     @classmethod
     def from_response(cls, response, formname=None, formid=None, formnumber=0, formdata=None,


Gallaecio:
A bit of food for though:

-   What would be the expected behavior in the following scenarios:

    -   `Request(url='https://example.com', params={'foo': 'bar'}).replace(url='https://example.org').url`

    -   `Request(url='https://example.com?foo=bar').params`

-   What about scenarios were keys need to be repeated? (i.e. `https://example.com?foo=bar&foo=baz`)

bulatbulat48:
> A bit of food for thought:
> 
> * What would be the expected output in the following scenarios:
>   
>   * `Request(url='https://example.com', params={'foo': 'bar'}).replace(url='https://example.org').url`
>   * `Request(url='https://example.com?foo=bar').params`
>   
>   I’m a bit hesitant about a feature like this, given that I’m not 100% sure of what the expectations should be in these scenarios. It may be better for users to use `w3lib.url.add_or_replace_parameters` or similar to build the URL, and continue to pass only the URL to `Request`.
> * What about scenarios were keys need to be repeated? (i.e. `https://example.com?foo=bar&foo=baz`)

Many thanks for the reply and detailed explonation @Gallaecio.

`Request(url='https://example.com', params={'foo': 'bar'}).replace(url='https://example.org').url`
This first case looks good for me because you had replaced the current URL.

`Request(url='https://example.com?foo=bar').params`
Same behaviour in the Requests library:
```Python
r = requests.get('https://yandex.ru?a=b')
r.params
  File "<input>", line 1, in <module>
AttributeError: 'Response' object has no attribute 'params'
```

Good point! Looks like needs to implement it like in Requests library?
```
r = requests.get('https://yandex.ru', params=[('a', 'c'), ('a','m')])
r.url
u'https://yandex.ru/?a=c&a=m'
```

bulatbulat48:
@Gallaecio added a case where the keys need to be repeated. Please review.

scrapy/http/request/__init__.py
```diff
@@ -65,12 +65,12 @@ def _set_params(self, params):
         if params is None:
             self._params = None
             return
-        if isinstance(params, dict):
+        if any([isinstance(params, dict), isinstance(params, list)]):
```
Gallaecio, line 68: You can simplify this as:

```suggestion
        if isinstance(params, (dict, list)):
```

However, since urlencode supports additional types (e.g. tuples), maybe we should just pass the argument to `urlencode` and let it complain if it is not supported.

scrapy/http/request/__init__.py
```diff
@@ -55,6 +58,22 @@ def meta(self):
...
+            self._params = None
+            return
+        if any([isinstance(params, dict), isinstance(params, list)]):
+            qs = urlencode(params, doseq=True)
+            self._set_url(self.url + ('&' if '?' in self.url else '?') + qs)
+            self._params = qs
+        else:
+            raise TypeError('Params must be dict or a list of 2-tuples, got {}'.format(type(params)))
+
+    params = property(_get_params, obsolete_setter(_set_params, 'params'))
```
Gallaecio, line 75: `obsolete_setter` seems meant for properties that use to be writable and are now read-only, which is not the case.

I think you can just omit the setter part altogether.

tests/test_http_request.py
```diff
@@ -36,12 +36,19 @@ def test_init(self):
...
         headers = {b"caca": b"coco"}
-        r = self.request_class("http://www.example.com", meta=meta, headers=headers, body="a body")
+        params = {'first': 'param1', 'second': 'param2'}
+        r = self.request_class("http://www.example.com", meta=meta, headers=headers, body="a body", params=params)
 
         assert r.meta is not meta
         self.assertEqual(r.meta, meta)
         assert r.headers is not headers
         self.assertEqual(r.headers[b"caca"], b"coco")
+        assert r.headers is not headers
```
Gallaecio, line 46:
```suggestion
        assert r.params is not params
```

Gallaecio:
> > * What would be the expected output in the following scenarios:
> >   
> >   * `Request(url='https://example.com', params={'foo': 'bar'}).replace(url='https://example.org').url`
> >   * `Request(url='https://example.com?foo=bar').params`
> >   
> >   I’m a bit hesitant about a feature like this, given that I’m not 100% sure of what the expectations should be in these scenarios. It may be better for users to use `w3lib.url.add_or_replace_parameters` or similar to build the URL, and continue to pass only the URL to `Request`.
> 
> `Request(url='https://example.com', params={'foo': 'bar'}).replace(url='https://example.org').url`
> This first case looks good for me because you had replaced the current URL.

I take it that the current implementation would return `https://example.org`. As a user, *I* am not sure what *I* would expect, if that or `https://example.org?foo=bar`.

To argue against the current implementation: why, if I did not replace `params`, it got replaced? Do I have to set params manually every time? (but I can imagine the opposite being argued as well)

This is why I’m +0 on merging this change at the moment. I won’t oppose it being merged, but I won’t approve it myself either, unless I’m convinced user expectations are likely to be either one or the other, based for example on whatever `requests` and the standard library do.

> `Request(url='https://example.com?foo=bar').params`
> Same behaviour in the Requests library:
> 
> ```python
> r = requests.get('https://yandex.ru?a=b')
> r.params
> Traceback (most recent call last):
>   File "<input>", line 1, in <module>
> AttributeError: 'Response' object has no attribute 'params'
> ```

Not the same. In your `requests` code you are trying to access the params of a response. I’m talking about the params of the request object. See https://requests.readthedocs.io/en/master/_modules/requests/models/#Request

I have not tested their code, but it *seems* to me that their behavior is different to ours: they don’t change the content of `url` based on the content of `params`, they keep their content separate. They must merge them at a later stage, outside the request.

If we go the `requests` route, which in my opinion would make expectations much more clear, I think we are talking about a significant backward-incompatible API change, as code using a `Request` object *needs* to be aware of the `params` property in order to merge `url` and `params` before working with the full URL, even if we provide a new property or method to get the merged URL.

But if we *don’t* go the `requests` route, I am not sure which API would be the most intuitive and consistent.

One other approach I can think of would be to allow `params` in `__init__` and `replace`, but not expose its content as a `params` property of `Request` or even store it, and use it simply to modify the content of the passed URL. Is not the same approach as `requests`, but it is still a quality-of-life improvement when using `__init__` and `replace`. I think I could get behind that approach, and I think expectations would be slightly more clear.

scrapy/http/request/__init__.py
bulatbulat48: +1. If the types do not match, the compiler throws their own error.

scrapy/http/request/__init__.py
bulatbulat48: Setter part removed.

wRAR:
How does this compare to the FormRequest solution presented in the original issue?

bulatbulat48:
> How does this compare to the FormRequest solution presented in the original issue?

@wRAR I think this does not belong to FormRequest solution. Mostly I just tried to reproduce this behavior https://github.com/scrapy/scrapy/issues/2607#issue-211025957

PR #4731, opened 2020-08-17
FILES_EXPIRES less than 1 day

gaodeng:
FILES_EXPIRES setting support less than 1 day expiration time


Gallaecio:
Could you add a test? I cannot find existing tests about the setting, so it may not be trivial. Let us know if you need help.

wRAR (changes requested):
This needs a test added.

AdrianAtZyte (approved):

PR #4736, opened 2020-08-18
Windows pip installation guide

### LINKED ISSUE #4715: Document Windows installation with pip
While it’s OK to recommend Anaconda, we still should provide documentation on the `pip` installation method for people that prefer to take that route, including the installation of the required compiling tools.

[A recent Reddit entry covered the required steps](https://www.reddit.com/r/scrapy/comments/hzgtwt/how_to_install_scrapy_using_pip_for_python3_in/).

AKSHAYSHARMAJS:
Added few steps. Will add all the steps soon.

Fixes #4715

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -127,16 +127,16 @@ To install Scrapy on Windows using ``pip`` :
 
 #. Download and execute `Microsoft C++ Build Tools`_ to install the Visual Studio Installer.
 
-#. Open Visual Studio Installer, install Visual Studio and modify installation.
+#. Open Visual Studio Installer, install `Visual Studio`_.
 
 #. Under workloads section, select **Desktop development with C++**.
 
-#. Below Installation details, select and install following packages :
+#. Check installation details and make sure following packages are selected as optional components:
 
   * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
   * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))
 
-Now, you should be able to :ref:`install scrapy <intro-install-scrapy>` using ``pip``.
+Now, you should be able to :ref:`install Scrapy <intro-install-scrapy>` using ``pip``.
 
 .. _intro-install-ubuntu:
 
@@ -290,5 +290,6 @@ For details, see `Issue #2473 <https://github.com/scrapy/scrapy/issues/2473>`_.
 .. _Scrapinghub: https://scrapinghub.com
 .. _Anaconda: https://docs.anaconda.com/anaconda/
 .. _Miniconda: https://docs.conda.io/projects/conda/en/latest/user-guide/install/index.html
+.. _Visual Studio: https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio
 .. _Microsoft C++ Build Tools: https://visualstudio.microsoft.com/visual-cpp-build-tools/
 .. _conda-forge: https://conda-forge.org/
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -128,7 +128,7 @@ To install Scrapy on Windows using pip :-
 
 3. Create Venv (Virtual Environment) by using::
 
-   python3 -m venv [env name]
+   python -m venv [env name]
 
 4. Download `Microsoft Visual C++ 14.0`_
 
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -287,5 +287,5 @@ For details, see `Issue #2473 <https://github.com/scrapy/scrapy/issues/2473>`_.
 .. _Scrapinghub: https://scrapinghub.com
 .. _Anaconda: https://docs.anaconda.com/anaconda/
 .. _Miniconda: https://docs.conda.io/projects/conda/en/latest/user-guide/install/index.html
-.. _Microsoft Visual C++ 14.0 : https://visualstudio.microsoft.com/visual-cpp-build-tools/
+.. _Microsoft Visual C++ 14.0: https://visualstudio.microsoft.com/visual-cpp-build-tools/
 .. _conda-forge: https://conda-forge.org/
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -123,7 +123,7 @@ To install Scrapy on Windows using ``pip`` :
 
 .. warning::
     This installation method requires “Microsoft Visual C++” for installing some 
-    scrapy dependencies which demands significantly more disk space than Anaconda.
+    Scrapy dependencies which demands significantly more disk space than Anaconda.
 
 #. Download and execute `Microsoft C++ Build Tools`_ to install the Visual Studio Installer.
 
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -119,23 +119,25 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
 
   conda install -c conda-forge scrapy
 
-To install Scrapy on Windows using ``pip`` :
+To install Scrapy on Windows using ``pip``:
 
 .. warning::
     This installation method requires “Microsoft Visual C++” for installing some 
-    Scrapy dependencies which demands significantly more disk space than Anaconda.
+    Scrapy dependencies, which demands significantly more disk space than Anaconda.
 
 #. Download and execute `Microsoft C++ Build Tools`_ to install the Visual Studio Installer.
 
-#. Run Visual Studio Installer.
+#. Run the Visual Studio Installer.
 
-#. Under workloads section, select **C++ build tools**.
+#. Under the Workloads section, select **C++ build tools**.
 
-#. Check installation details and make sure following packages are selected as optional components:
-        * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
-        * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))
+#. Check the installation details and make sure following packages are selected as optional components:
 
-#. Install Visual Studio Build Tools.
+    * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
+    
+    * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))
+
+#. Install the Visual Studio Build Tools.
 
 Now, you should be able to :ref:`install Scrapy <intro-install-scrapy>` using ``pip``.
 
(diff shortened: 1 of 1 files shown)


AKSHAYSHARMAJS:
@Gallaecio Is there need to add screenshots ?

Gallaecio:
I would not add screenshots, they are quick to become outdated and hard to update.

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
 
   conda install -c conda-forge scrapy
 
+To install Scrapy on Windows using ``pip`` :-
```
Gallaecio, line 121:
```suggestion
To install Scrapy on Windows using ``pip``:
```

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+1. Install latest version of `Python`_ (Recommended to tick [Add Python to PATH] checkbox during installation)
```
Gallaecio, line 123: I think we should skip this step: we already cover this requirement at the beginning, and the latest version of Python is not really a requirement.

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+2. Install or upgrade pip::
```
Gallaecio, line 125: This step should not be needed according to https://pip.pypa.io/en/stable/installing/#do-i-need-to-install-pip

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+    python -m pip install --upgrade pip
+
+3. Create Venv (Virtual Environment)::
```
Gallaecio, line 129: I think this step and the last one about installing Scrapy itself could be removed from these steps and, instead, we could add a paragraph after the steps that says “install Scrapy using pip” which links back up to the “Installing Scrapy” section, which already suggests creating a virtual environment.

In “Installing Scrapy”, maybe we should move the paragraph about virtual environments 1 paragraph up, so that it comes right after the `pip` command, and before the paragraph that links to these Windows-specific steps.

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+    python -m venv [env name]
+
+4. Download `Microsoft Visual C++ 14.0`_
```
Gallaecio, line 133: - I would recommend using `#. ` instead of `4. `, so that we do not need to worry about list numbers when adding, removing or moving steps in the future.

- Do we need to indicate a specific version?

- The link at the moment calls the package “Microsoft C++ Build Tools”. If that’s the name, we should use it here. If users are meant to install “Microsoft Visual C++” using the “Microsoft C++ Build Tools” installer, we should explain that (e.g. “install the Microsoft C++ Build Tools, which include the Microsoft Visual C++ required to install some Scrapy dependencies”).

docs/intro/install.rst
```diff
@@ -118,6 +118,27 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+5. Run the downloaded EXE file and let it setup the Visual Studio Installer
+
+6. Open Visual Studio Installer, select and install C++ build tools
```
Gallaecio, line 137: I’m a bit confused about these 3 steps (4, 5, 6).

I think it may be best to refactor them a bit to make clear what you are installing. For example:

1. Download and execute the **Microsoft C++ Build Tools** to install the **Visual Studio Installer**.
2. Open the **Visual Studio Installer**, and select and install the **C++ build tools**.

But I think I might be misunderstanding how this works. https://stackoverflow.com/a/55053709/939364 suggests that the exe opens an installer, and that you need to select 2 packages to be installed with it. If that is so, these steps need to be refactored.

All in all, if https://stackoverflow.com/a/55053709/939364 is correct, I think we don’t even need a list. A paragraph can explain what https://stackoverflow.com/a/55053709/939364 describes, making sure we use the component names as they appear in the installer. Then a second paragraph can link back to “Installing Scrapy” as I suggested above, which covers installing Scrapy and virtual environments.

Gallaecio: To make the documentation more future proof, I would omit versions (e.g. MSCV, Windows SDK); if it seems too ambiguous, maybe provide version-specific names as examples.

As for the size warning, does it really need to download 4 GiB if you only enable those 2 options? If so, I’m OK with including a warning in our documentation, but I’m worried about indicating a specific size in our documentation and that changing upstream with time. I’m not sure how to word it without a specific size or range, though. Maybe warn that this installation method requires significantly more disk space than Anaconda? (assuming that’s true).

AKSHAYSHARMAJS: > To make the documentation more future proof, I would omit versions (e.g. MSCV, Windows SDK); if it seems too ambiguous, maybe provide version-specific names as examples.

Alright I will omit versions and add an example to makes things less ambiguous.

docs/intro/install.rst
```diff
@@ -118,27 +119,24 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
-
-1. Install latest version of `Python`_ (Recommended to tick [Add Python to PATH] checkbox during installation)
-
-2. Install or upgrade pip::
+To install Scrapy on Windows using ``pip`` :
 
-    python -m pip install --upgrade pip
+.. warning::
+    This installation method requires “Microsoft Visual C++” for installing some 
+    scrapy dependencies which demands significantly more disk space than Anaconda.
```
Gallaecio, line 126:
```suggestion
    Scrapy dependencies which demands significantly more disk space than Anaconda.
```

docs/intro/install.rst
```diff
@@ -118,27 +119,24 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
 
-4. Download `Microsoft Visual C++ 14.0`_
+#. Under workloads section, select **Desktop development with C++**.
 
-5. Run the downloaded EXE file and let it setup the Visual Studio Installer
+#. Below Installation details, select and install following packages :
 
-6. Open Visual Studio Installer, select and install C++ build tools
+  * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
+  * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))
```
Gallaecio: Do you need to modify the installation? Can’t you install all that is needed at the same time as you install Visual Studio?

I’m thinking that we could replace these steps by indicating to the user to install Visual Studio (linking to the upstream documentation, https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio ) and indicate which workloads and individual components need to be selected during that process. I assume MSVC and Windows SDK are listed under “Individual components”.

AKSHAYSHARMAJS: > Do you need to modify the installation? Can’t you install all that is needed at the same time as you install Visual Studio?

modification window automatically pops up(including workloads and individual components) when we install visual studio.

AKSHAYSHARMAJS: > I’m thinking that we could replace these steps by indicating to the user to install Visual Studio (linking to the upstream documentation, https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio )

That's a good idea. Linking to upstream documentation will help user understanding the workflows.



>  indicate which workloads and individual components need to be selected during that process. I assume MSVC and Windows SDK are listed under “Individual components”.

under workloads when we select **Desktop development with C++** . Those 2 packages are already selected along with some other packages in option components section on the right side. Maybe we should just tell users make sure MSVC and SDK are selected.

docs/intro/install.rst
```diff
@@ -127,16 +127,16 @@ To install Scrapy on Windows using ``pip`` :
 
 #. Download and execute `Microsoft C++ Build Tools`_ to install the Visual Studio Installer.
 
-#. Open Visual Studio Installer, install Visual Studio and modify installation.
+#. Open Visual Studio Installer, install `Visual Studio`_.
```
Gallaecio, line 130: This sentence still sounds like you first install Visual Studio, and then you can access a “Workloads” section. But from reading [the upstream documentation](https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio?view=vs-2019) it sounds like you get that Workloads section as soon as you open the installer. Isn’t that so?

AKSHAYSHARMAJS: Sorry I forgot to change that and also I figured out that I was a little bit wrong [here](https://github.com/scrapy/scrapy/pull/4736#discussion_r475101391). **Desktop development with C++** comes when we install [VS IDE](https://visualstudio.microsoft.com/downloads) too along with other things but here we need only [build tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/) so [reddit post](https://www.reddit.com/r/scrapy/comments/hzgtwt/how_to_install_scrapy_using_pip_for_python3_in/) was right, this should be **C++ build tools**

docs/intro/install.rst
```diff
@@ -118,6 +119,25 @@ Once you've installed `Anaconda`_ or `Miniconda`_, install Scrapy with::
...
+
+#. Under workloads section, select **C++ build tools**.
+
+#. Check installation details and make sure following packages are selected as optional components:
+        * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
+        * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))
+
+#. Install Visual Studio Build Tools.
+
+Now, you should be able to :ref:`install Scrapy <intro-install-scrapy>` using ``pip``.
```
Gallaecio, line 140: :lipstick: 

```suggestion
To install Scrapy on Windows using ``pip``:

.. warning::
    This installation method requires “Microsoft Visual C++” for installing some 
    Scrapy dependencies, which demands significantly more disk space than Anaconda.

#. Download and execute `Microsoft C++ Build Tools`_ to install the Visual Studio Installer.

#. Run the Visual Studio Installer.

#. Under the Workloads section, select **C++ build tools**.

#. Check the installation details and make sure following packages are selected as optional components:

    * **MSVC**  (e.g MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.23) )
    
    * **Windows SDK**  (e.g Windows 10 SDK (10.0.18362.0))

#. Install the Visual Studio Build Tools.

Now, you should be able to :ref:`install Scrapy <intro-install-scrapy>` using ``pip``.
```

Gallaecio (approved):
Nice work!

Gallaecio (approved):

AKSHAYSHARMAJS:
hey @Gallaecio, I was just going through the file [install.rst](https://github.com/scrapy/scrapy/blob/master/docs/intro/install.rst). I think the link to [cryptography installation](https://cryptography.io/en/latest/installation/) is outdated.

Gallaecio:
Indeed. We should probably use InterSphinx to be safer.

Would you create a separate pull request for that, or log an issue about it so that someone else can work on it?

Issue #4211, opened 2019-12-04
SitemapSpider throws lxml.etree.XMLSyntaxError when trying to scrape a blank sitemap page


Endi1:
### Description

`SitemapSpider` throws a `lxml.etree.XMLSyntaxError` when hitting a blank sitemap page while crawling a sitemap. 

### Stack trace
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/scrapy/utils/defer.py", line 102, in iter_errback
    yield next(it)
  File "/usr/local/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 84, in evaluate_iterable
    for r in iterable:
  File "/usr/local/lib/python3.8/site-packages/sh_scrapy/middlewares.py", line 30, in process_spider_output
    for x in result:
  File "/usr/local/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 84, in evaluate_iterable
    for r in iterable:
  File "/usr/local/lib/python3.8/site-packages/scrapy/spidermiddlewares/offsite.py", line 29, in process_spider_output
    for x in result:
  File "/usr/local/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 84, in evaluate_iterable
    for r in iterable:
  File "/usr/local/lib/python3.8/site-packages/scrapy/spidermiddlewares/referer.py", line 339, in <genexpr>
    return (_set_referer(r) for r in result or ())
  File "/usr/local/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 84, in evaluate_iterable
    for r in iterable:
  File "/usr/local/lib/python3.8/site-packages/scrapy/spidermiddlewares/urllength.py", line 37, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "/usr/local/lib/python3.8/site-packages/scrapy/core/spidermw.py", line 84, in evaluate_iterable
    for r in iterable:
  File "/usr/local/lib/python3.8/site-packages/scrapy/spidermiddlewares/depth.py", line 58, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "/usr/local/lib/python3.8/site-packages/scrapy/spiders/sitemap.py", line 53, in _parse_sitemap
    s = Sitemap(body)
  File "/usr/local/lib/python3.8/site-packages/scrapy/utils/sitemap.py", line 18, in __init__
    self._root = lxml.etree.fromstring(xmltext, parser=xmlp)
[...]

PR #4214, opened 2019-12-05
Handle XMLSyntaxError on blank sitemap pages for SitemapSpider

Endi1:
This PR implements a fix for the issue mentioned on #4211. Namely, it gracefully handles the `XMLSyntaxError` that gets thrown anytime the `SitemapSpider` encounters a blank sitemap page.

Fixes #4211

Issue #3331, opened 2018-07-11
Documentation example fails with `proxy URL with no authority`


a-palchikov:
Running the [example](https://doc.scrapy.org/en/1.5/intro/overview.html#walk-through-of-an-example-spider) from the documentation yields this:
```
10:11 $ scrapy runspider quotes.py 
2018-07-11 10:12:04 [scrapy.utils.log] INFO: Scrapy 1.5.0 started (bot: scrapybot)
2018-07-11 10:12:04 [scrapy.utils.log] INFO: Versions: lxml 3.5.0.0, libxml2 2.9.3, cssselect 0.9.1, parsel 1.5.0, w3lib 1.19.0, Twisted 16.0.0, Python 2.7.12 (default, Dec  4 2017, 14:50:18) - [GCC 5.4.0 20160609], pyOpenSSL 0.15.1 (OpenSSL 1.0.2g  1 Mar 2016), cryptography 1.2.3, Platform Linux-4.4.0-130-generic-x86_64-with-Ubuntu-16.04-xenial
2018-07-11 10:12:04 [scrapy.crawler] INFO: Overridden settings: {'SPIDER_LOADER_WARN_ONLY': True}
2018-07-11 10:12:04 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.memusage.MemoryUsage',
 'scrapy.extensions.logstats.LogStats',
 'scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.corestats.CoreStats']
Unhandled error in Deferred:
2018-07-11 10:12:04 [twisted] CRITICAL: Unhandled error in Deferred:


Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/scrapy/commands/runspider.py", line 88, in run
    self.crawler_process.crawl(spidercls, **opts.spargs)
  File "/usr/local/lib/python2.7/dist-packages/scrapy/crawler.py", line 171, in crawl
    return self._crawl(crawler, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/scrapy/crawler.py", line 175, in _crawl
    d = crawler.crawl(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1274, in unwindGenerator
    return _inlineCallbacks(None, gen, Deferred())
--- <exception caught here> ---
[...]

PR #4739, opened 2020-08-19
Add test-case for issue#4211

Yogita98:
Hey @Gallaecio 
Here I am adding a test for issue #4211 
Could you please verify and let me know where I'm going wrong?

### DIFF
--- a/tests/test_spider.py
+++ b/tests/test_spider.py
@@ -412,6 +412,11 @@ class SitemapSpiderTest(SpiderTest):
     g.close()
     GZBODY = f.getvalue()
 
+    def test_parse_sitemap(self):
+        r = Response(url="http://www.example.com", body=b"")        
+        with LogCapture() as lc:            
+            spider.log('Ignoring invalid sitemap', 'WARNING')
+
     def assertSitemapBody(self, response, body):
         spider = self.spider_class("example.com")
         self.assertEqual(spider._get_sitemap_body(response), body)


Gallaecio:
I think you might not have completely understood the thread in https://github.com/scrapy/scrapy/pull/4214#discussion_r371284294

I think we made a mistake marking #4211 as “good first issue”, because even though the implementation itself is simple enough as #4214 shows, writing a test for it is far from trivial.

My recommendation is that you look for a different issue to work on. I think it was brave of you to volunteer to address this, but I think it can be a long way from this changeset to writing a test that covers the changes in #4214.

Yogita98:
No problem!
Can you suggest me some beginner friendly issues to work with?

Gallaecio:
Looking at [our list of good first issues](https://github.com/scrapy/scrapy/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22+), we don’t seem to have that many left, and only https://github.com/scrapy/scrapy/issues/3331 has no pull request associated with it.

You could give https://github.com/scrapy/scrapy/issues/3331 a try, there’s a traceback so it should be relatively easy to find the part of the code base that needs additional code.

Otherwise, you could review the existing pull requests for the other issues, and if there is any pull request where it looks like the author being unavailable to handle feedback is the reason for the pull request not to have been merged yet, you can consider resuming the work yourself as you aimed for here.

We really need to extend our list of good first issues :sweat_smile:

Issue #4732, opened 2020-08-17
Drop Python 3.5 in the CI configuration


Gallaecio:

Gallaecio:
For PyPy, I think we should support 7.2.0 as minimum version, given that’s the first version where 3.6 support went out of beta.

PR #4742, opened 2020-08-20
Update Python version references after dropping support for 3.5

### LINKED ISSUE #4733: Review references to Python 3.5 and 3.6
Now that we are dropping Python 3.5 support, such references in the code and documentation are likely to need an update.

Make sure to also look for the `version_info` equivalents (i.e. `(3, 5)`, `(3, 6)`).

Gallaecio:
Covers everything but:

- CI (#4732)
- An `if` statement that needs to check the PyPy version instead of the Python version (#4732)
- `py36.py` (#4467)

Resolves #4733


docs/intro/install.rst
```diff
@@ -9,8 +9,8 @@ Installation guide
 Supported Python versions
 =========================
 
-Scrapy requires Python 3.5.2+, either the CPython implementation (default) or
-the PyPy 5.9+ implementation (see :ref:`python:implementations`).
+Scrapy requires Python 3.6+, either the CPython implementation (default) or
+the PyPy 7.2.0+ implementation (see :ref:`python:implementations`).
```
Gallaecio, line 13: First PyPy version where 3.6 support went out of beta.

tests/test_proxy_connect.py
```diff
@@ -88,7 +86,7 @@ def test_https_connect_tunnel(self):
             yield crawler.crawl(self.mockserver.url("/status?n=200", is_secure=True))
         self._assert_got_response_code(200, log)
 
-    @pytest.mark.xfail(reason='Python 3.6+ fails this earlier', condition=sys.version_info >= (3, 6))
+    @pytest.mark.xfail(reason='fails earlier')
```
Gallaecio, line 89: I wonder if we should remove this test.

tox.ini
```diff
@@ -89,7 +89,7 @@ deps =
 basepython = python3
 deps =
     {[pinned]deps}
-    # First lxml version that includes a Windows wheel for Python 3.5, so we do
+    # First lxml version that includes a Windows wheel for Python 3.6, so we do
```
Gallaecio, line 92: Remains true for 3.6.

tests/test_proxy_connect.py
wRAR: Yeah, I think we should.

scrapy/utils/spider.py
```diff
@@ -4,15 +4,18 @@
...
-try:
-    from scrapy.utils.py36 import collect_asyncgen
-except SyntaxError:
-    collect_asyncgen = None
 
 
 logger = logging.getLogger(__name__)
 
 
+async def collect_asyncgen(result):
```
wRAR, line 12: I'm using this function in a test in the async start_requests PR, I've moved it to utils.asyncgen there.

scrapy/utils/py36.py
```diff
@@ -1,10 +0,0 @@
-"""
```
wRAR, line 1: Don't we want to deprecate this first? https://github.com/scrapy/scrapy/pull/4467/files#diff-0ec586c8bb4f8b18d1add1d954eb5448

tests/test_unsupported_python_version.py
```diff
@@ -0,0 +1,19 @@
...
+from io import StringIO
+from unittest import TestCase
+
+from pytest import mark
+
+
+class UnsupportedPythonVersionTestCase(TestCase):
+
+    @mark.skipif(sys.version_info >= (3, 6), reason='only meant for unsupported Python versions')
+    def test_unsupported_python_version(self):
```
Gallaecio, line 11: It can seem pointless, but it will allow our test coverage to eventually reach 100%, and as long as we keep the Travis CI job in Python 3.5 even after dropping 3.6 and later versions, it will prevent us from breaking this message, e.g. by switching it to an f-string.

tests/requirements-py3.txt
```diff
@@ -13,7 +12,7 @@ pytest-twisted >= 1.11
 pytest-xdist
 sybil >= 1.3.0  # https://github.com/cjw296/sybil/issues/20#issuecomment-605433422
 testfixtures
-uvloop; platform_system != "Windows"
+uvloop; platform_system != "Windows" and python_version >= '3.5'
```
kmike, line 15: why is it 3.5?

scrapy/__init__.py
```diff
@@ -28,11 +22,18 @@
 
 
 # Check minimum required Python version
-if sys.version_info < (3, 5, 2):
-    print("Scrapy %s requires Python 3.5.2" % __version__)
+if sys.version_info < (3, 6):
+    print(("Scrapy %s requires Python 3.6+" % __version__).encode())
```
kmike, line 26: why is .encode() needed here?

tests/requirements-py3.txt
Gallaecio: Because that’s [their minimum Python version](https://github.com/MagicStack/uvloop#installation). Otherwise, the new Python 2.7 jobs tries to install it, and fails.

scrapy/__init__.py
Gallaecio: I needed to do this to get the new test to pass in Python 2.7. It has to do with the fact that we are replacing `sys.stdout` to capture. I figure, being ASCII text, that this should not be a problem. But if we can test the output without having to do this, that would definitely be better.

I got issues when using StringIO, so I switched to BytesIO. I’ll try cStringIO as mentioned in https://stackoverflow.com/a/1218951/939364, maybe that will work.

tests/requirements-py3.txt
```diff
@@ -3,7 +3,6 @@ attrs
 dataclasses; python_version == '3.6'
 mitmproxy; python_version >= '3.7'
 mitmproxy >= 4, < 5; python_version >= '3.6' and python_version < '3.7'
```
kmike, line 5: Just in case, a suggestion:

```suggestion
mitmproxy >= 4.0.4, < 5; python_version >= '3.6' and python_version < '3.7'
```

There is a CVE-2018-14505 on earlier mitmproxy versions.

scrapy/__init__.py
kmike: I was concerned about this thing mostly:

```py
>>> print("foo")
foo
>>> print("foo".encode())
b'foo'
```

tests/test_unsupported_python_version.py
kmike: That's a very comprehensive approach, I'm impressed -  I've never seen a project testing against Python versions it doesn't support :) 

My main worries were about the CI time and maintenance time. 

CI time doesn't seem to be large, but that's still a minute per job. 

Regarding maintenance: installing of old Pythons may stop working, pip may stop working in older Pythons, having 2 more envs increases chances for random failure of a build (e.g. because of a network issue in one of the jobs). Also, we now use https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires in setup.py, so that the error message which we could break may become less and less relevant over time. 

Because of that, +0.5 from me for **not** testing against Python 2.7 and 3.5.

tests/requirements-py3.txt
kmike: It got reverted back after the force-push :)

Gallaecio: :facepalm:

kmike:
Thanks @Gallaecio!

PR #4743, opened 2020-08-20
Remove py35 from CI

### LINKED ISSUE #4732: Drop Python 3.5 in the CI configuration

elacuesta:
Closes #4732


Gallaecio:
Looks like we need to upgrade the pinned version of PIL for Windows to one with a wheel for 3.6

hugovk:
The lowest Pillow version supporting Python 3.6 (i.e. with wheels) is 4.0.0:

* https://pillow.readthedocs.io/en/stable/installation.html#notes
* https://pypi.org/project/Pillow/4.0.0/#files

Gallaecio:
Interesting.

```
            self.assertEqual(
                get_func_args(str.split, stripself=True), ['sep', 'maxsplit'])
            self.assertEqual(
                get_func_args(operator.itemgetter(2), stripself=True), ['obj'])
            if version_info < (3, 6):
                self.assertEqual(get_func_args(" ".join, stripself=True), ['list'])
            else:
>               self.assertEqual(get_func_args(" ".join, stripself=True), ['iterable'])
E               AssertionError: Lists differ: ['list'] != ['iterable']
E               
E               First differing element 0:
E               'list'
E               'iterable'
E               
E               - ['list']
E               + ['iterable']
```

That means my change in https://github.com/scrapy/scrapy/pull/4742/files#diff-98466f1cfa80fb4dd81e2ff23cd108d7 would be incorrect, as the different value depends not in the Python version, but in the PyPy version.

In that case, I think we may need to use `platform.python_revision()` for this `if`. Not sure which value to use, though; I would check what the value is during the failing test with the 7.2.0 PyPy version, and use that one.

PR #4746, opened 2020-08-21
Fix iternodes

Gallaecio:
This is a continuation of #861.

I’m proposing that we merge the change _as is_, given it does fix an issue in the current implementation, and then create separate tickets for:

- Applying the equivalent fix for `xmliter_lxml` (bug)
- Switching `xmliter` to `iterparse` (enhancement)

Resolves #861


wRAR (approved):

wRAR:
I'm fine with merging this but there are two instances of %-formatting, should we replace them with f-strings first?

PR #4752, opened 2020-08-24
DOWNLOAD_MAXSIZE logger level changed from Error to Warning

### LINKED ISSUE #3874: DOWNLOAD_MAXSIZE logger level shouldn't be Error

drs-11:
Fixes #3874, closes #3886.

Previously @WinterComes  tried to solve the issue but hasn't responded to one final query by @kmike.
I completed the simple query.

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -394,9 +394,9 @@ class ScrapyAgent:
         fail_on_dataloss = request.meta.get('download_fail_on_dataloss', self._fail_on_dataloss)
 
         if maxsize and expected_size > maxsize:
-            warning_msg = ("Expected response size (%(size)s) larger than "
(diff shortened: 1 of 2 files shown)


WinterComes:
@drs-11 thanks for this

Gallaecio (approved):

wRAR (approved):

wRAR:
Thanks!

PR #4753, opened 2020-08-25
Do not recode JPEG files in imagepipelines

### LINKED ISSUE #3055: [bug] pillow will always recode images in imagepipieline

drs-11:
This patch completes the last few suggestions on the pull request #3689.

Fixes #3055, closes #3689


scrapy/pipelines/images.py
```diff
@@ -5,19 +5,20 @@
...
 
-from scrapy.exceptions import DropItem
+from scrapy.exceptions import DropItem, ScrapyDeprecationWarning
 from scrapy.http import Request
 from scrapy.pipelines.files import FileException, FilesPipeline
 # TODO: from scrapy.pipelines.media import MediaPipeline
 from scrapy.settings import Settings
 from scrapy.utils.misc import md5sum
-from scrapy.utils.python import to_bytes
+from scrapy.utils.python import to_bytes, get_func_args
```
Gallaecio, line 21: :lipstick: 

```suggestion
from scrapy.utils.python import get_func_args, to_bytes
```

scrapy/pipelines/images.py
```diff
@@ -128,15 +131,33 @@ def get_images(self, response, request, info, *, item=None):
             raise ImageException("Image too small (%dx%d < %dx%d)" %
                                  (width, height, self.min_width, self.min_height))
 
-        image, buf = self.convert_image(orig_image)
+        if self._deprecated_convert_image is None:
+            self._deprecated_convert_image = 'response_body' not in get_func_args(self.convert_image)
+            if self._deprecated_convert_image:
+                warnings.warn('ImagesPipeline.convert_image() method overriden in a deprecated way, '
+                              'overriden method does not accept response_body argument.',
+                              category=ScrapyDeprecationWarning, stacklevel=1)
```
Gallaecio, line 139: :lipstick: `1` is the default stack level.

```suggestion
                              category=ScrapyDeprecationWarning)
```

scrapy/pipelines/images.py
```diff
@@ -128,15 +131,33 @@ def get_images(self, response, request, info, *, item=None):
...
+            else:
+                thumb_image, thumb_buf = self.convert_image(image, size, buf)
             yield thumb_path, thumb_image, thumb_buf
 
-    def convert_image(self, image, size=None):
+    def convert_image(self, image, size=None, response_body=None):
+        if response_body is None:
+            warnings.warn('ImagesPipeline.convert_image() method called in a deprecated way, '
+                          'method called without response_body argument.',
+                          category=ScrapyDeprecationWarning, stacklevel=1)
```
Gallaecio, line 159: :lipstick: Since the issue is in the calling code, `2` is probably a more useful stack level here. 

```suggestion
                          category=ScrapyDeprecationWarning, stacklevel=2)
```

Gallaecio (commented):
Nice take over!

The [coverage data](https://codecov.io/gh/scrapy/scrapy/pull/4753/diff) shows an issue, though: the tests only cover the change in `convert_image`, the change in `get_images`, which is the most significant, is not currently covered by tests for scenarios where the new parameter is missing.

drs-11:
Hey @Gallaecio, I think the test cases cover `get_images` now. Can you review this?

Gallaecio (approved):
Awesome!

drs-11:
Ummm @kmike you requested some changes in the previous PR. This looks good to go now?

scrapy/pipelines/images.py
```diff
@@ -129,15 +132,33 @@ def get_images(self, response, request, info, *, item=None):
                                  f"({width}x{height} < "
                                  f"{self.min_width}x{self.min_height})")
 
-        image, buf = self.convert_image(orig_image)
+        if self._deprecated_convert_image is None:
+            self._deprecated_convert_image = 'response_body' not in get_func_args(self.convert_image)
+            if self._deprecated_convert_image:
+                warnings.warn('ImagesPipeline.convert_image() method overriden in a deprecated way, '
```
kmike, line 138: What do you think about using the exact class name here instead of ImagesPipeline?

kmike: The same applies for the other warninig message

kmike:
The implementation looks good to me overall @drs-11, thanks for picking it up! I haven't really checked the tests, but rely on you all to get them right  :) +1 to merge after addressing a couple of comments about warning messages.

scrapy/pipelines/images.py
drs-11: By exact class name you mean the name of the class instance variable? I didn't quite get that.

kmike: The warning would fire if user creates a subclass of ImagesPipeline and overrides convert_image method. So it would be `MyImagesPipeline.convert_image` method which we should warn about, not `ImagesPipeline.convert_image`. Something like `self.__class__.__name__` + ".convert_images()" (haven't checked it).

wRAR (approved):

Issue #4754, opened 2020-08-25
Set exporter parameters by FEED


maranqz:
## Summary

Adding possibility set parameters by FEEDS variable.
```python
FEEDS = {
      'some.csv': {
             ...
             'errors': 'xmlcharrefreplace',
       }
}
```

## Motivation

For changing options of ExporterItem, programmer should create new class. It can be simplified if there will be possibility set parameters by FEEDS variable in direct.

## Additional context

In first I want to add possibility setting of "errors" in CsvItemExporter because sometimes I have to exporting in cyrillic files with saving unicode symbols. It is first test contribute in scrapy, then I will add the setting parameters by FEEDS.
```python
self.stream = io.TextIOWrapper(
    ...
    errors=errors,
)
```

maranqz:
Copy https://github.com/scrapy/scrapy/issues/4606

Issue #4606, opened 2020-06-01
Pass info from FEEDS to ItemExporter


StasDeep:
## Summary

The new FeedExporter class doesn't allow to pass custom info to ItemExporter through the `FEEDS` setting.

## Motivation

My use-case is a bit too long to describe, but it can be minimized to the following: for example, I have a custom ItemExporter which I want to have different options for different feeds. Say I want to pass custom `param` option to it.

```python
class CustomItemExporter(BaseItemExporter):
    
    def _configure(self, options, dont_fail=False):
        super()._configure(options, dont_fail)
        self.param = options.pop('param', None)
        
    def export_item(self, item):
        # Using self.param here
        ...
```

Ideally, I would like to do set it like this:

```python
FEEDS = {
    uri_detail: {'format': 'custom', 'item_export_kwargs': {'param' : 'Detail'}},
    uri_overview: {'format': 'custom', 'item_export_kwargs': {'param' : 'Overview'} }
}
```

## Describe alternatives you've considered

The only workaround I've found was to create a new FeedExporter class and override `open_spider` method to pass `feed['item_export_kwargs']` to `_get_exporter` method.

## Additional context

Will provide a PR this week.

StasDeep:
I will provide a PR this week.

Gallaecio:
I implemented this in #4512, as I needed it there.

StasDeep:
@Gallaecio if I understand the changes correctly, it is now possible to pass it to FeedStorage. What I want, though, is passing custom args to ItemExporters.

PR #4755, opened 2020-08-25
Add errors parameter for CsvItemExporter with tests

maranqz:
Additional context from https://github.com/scrapy/scrapy/issues/4754

### DIFF
--- a/docs/topics/exporters.rst
+++ b/docs/topics/exporters.rst
@@ -311,9 +311,10 @@ CsvItemExporter
       multi-valued fields, if found.
    :type include_headers_line: str
 
-    :param errors: The optional string that specifies how encoding and decoding errors are to be handled.
-        For more information see `documentation <https://docs.python.org/3/library/io.html#io.TextIOWrapper>`_.
-    :type errors: str
+   :param errors: The optional string that specifies how encoding and decoding
+      errors are to be handled. For more information see
+      :class:`io.TextIOWrapper`.
+   :type errors: str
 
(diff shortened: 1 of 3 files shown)


Gallaecio (commented):
Looks good to me, only the documentation update is missing: https://docs.scrapy.org/en/latest/topics/exporters.html#csvitemexporter

docs/topics/exporters.rst
```diff
@@ -311,6 +311,10 @@ CsvItemExporter
       multi-valued fields, if found.
    :type include_headers_line: str
 
+    :param errors: The optional string that specifies how encoding and decoding errors are to be handled.
+        For more information see `documentation <https://docs.python.org/3/library/io.html#io.TextIOWrapper>`_.
```
Gallaecio, line 315: Please try building the documentation locally with `tox -e docs` to verify, but I think using InterSphinx here should work:

```suggestion
        For more information see :class:`io.TextIOWrapper`.
```

Gallaecio (approved):

kmike:
Hey! Just to make sure - is my understanding is correct?

* As implemented, this is a feature which is not easily accessable for the end users
* to make it available through FEEDS option https://github.com/scrapy/scrapy/issues/4606 needs to be fixed, which was out of scope for this PR

maranqz:
@kmike Yes, you are right. I think to do that task https://github.com/scrapy/scrapy/issues/4606 in the month.

kmike:
Thanks @maranqz!

PR #4756, opened 2020-08-25
Support for delegated methods as callbacks

ivanprado:
It is useful to split the spider code around some helper classes if the spider code is too big. Using inheritance and mixings is the regular approach in this case. But using callback delegation could be a useful alternative in some cases but Scrapy does not support it. For example, this is not allowed currently:

```py
class UKHomepageParser:
  def parse(self, response):
    ...

class USHomepageParser:
  def parse(self, response):
    ...

class MySpider(Spider):
  def __init__(self, location):
    if location == "UK:
      self.parse = UKHomepageParser().parse
    else:
      self.parse = USHomepageParser().parse

  ...
```

Nothing prevents Scrapy to support this case. This PR changes the function `_find_method` in `reqser.py` removing the constraint that forces a callback `self` to be the spider. Being a method from a different class is not a problem as long as this method is a member of spider class. The required change to support this is pretty small.

### DIFF
--- a/scrapy/utils/reqser.py
+++ b/scrapy/utils/reqser.py
@@ -71,23 +71,19 @@ def request_from_dict(d, spider=None):
 
 
 def _find_method(obj, func):
-    if obj:
-        try:
-            func.__func__
-        except AttributeError:  # func is not a instance method. Not supported.
-            pass
-        else:
-            members = inspect.getmembers(obj, predicate=inspect.ismethod)
-            for name, obj_func in members:
-                # We need to use __func__ to access the original
-                # function object because instance method objects
-                # are generated each time attribute is retrieved from
-                # instance.
-                #
-                # Reference: The standard type hierarchy
-                # https://docs.python.org/3/reference/datamodel.html
-                if obj_func.__func__ is func.__func__:
-                    return name
+    # Only instance methods contain ``__func__``
+    if obj and hasattr(func, '__func__'):
+        members = inspect.getmembers(obj, predicate=inspect.ismethod)
+        for name, obj_func in members:
+            # We need to use __func__ to access the original
+            # function object because instance method objects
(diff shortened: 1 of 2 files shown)


Gallaecio (approved):
Looks good to me.

wRAR:
Can you please fix the flake8 errors?

Gallaecio:
This needs a slight update: https://docs.scrapy.org/en/latest/topics/jobs.html#request-serialization

Gallaecio (approved):

scrapy/utils/reqser.py
```diff
@@ -73,23 +73,22 @@ def request_from_dict(d, spider=None):
 def _find_method(obj, func):
```
victor-torres, line 73: Maybe you could simplify this statement to something like this:

```python
def _find_method(obj, func):
    if obj and hasattr(func, '__func__'):
            members = inspect.getmembers(obj, predicate=inspect.ismethod)
            for name, obj_func in members:
                # We need to use __func__ to access the original
                # function object because instance method objects
                # are generated each time attribute is retrieved from
                # instance.
                #
                # Reference: The standard type hierarchy
                # https://docs.python.org/3/reference/datamodel.html
                if obj_func.__func__ is func.__func__:
                    return name
    raise ValueError("Function %s is not an instance method in: %s" % (func, obj))
```

ivanprado: Good idea! Done here: https://github.com/scrapy/scrapy/pull/4756/commits/0524df866936506ae9438a5565fc4733fa5ba5b0

kmike:
Makes sense, thanks @ivanprado!

ivanprado:
Thank you @Gallaecio @victor-torres and @kmike for the review. Very quick! :smile:

PR #4765, opened 2020-08-28
Simplify tests/test_crawl

wRAR:
This adds a single function to run a spider and collect the items, logs and stats, to simplify existing and future test cases.

Includes #4764


kmike:
The diff is confusiing :) I'd expect only last changes to be present. Could you please check it @wRAR?

wRAR:
Yeah, I have no idea why does it still show the merged commit.

kmike:
Nice cleanup, thanks @wRAR!

PR #4768, opened 2020-08-30
Pass info from FEEDS to ItemExporter

maranqz:
Resolves https://github.com/scrapy/scrapy/issues/4606

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -303,6 +303,9 @@ For instance::
             'store_empty': False,
             'fields': None,
             'indent': 4,
+            'item_export_kwargs': {
+               'export_empty_fields': True,
+            },
         }, 
         '/home/user/documents/items.xml': {
             'format': 'xml',
@@ -331,6 +334,8 @@ as a fallback value if that key is not provided for a specific feed definition:
 -   ``fields``: falls back to :setting:`FEED_EXPORT_FIELDS`.
 
 -   ``indent``: falls back to :setting:`FEED_EXPORT_INDENT`.
+
+-   ``item_export_kwargs``: dict with kwargs for :ref:`Item exporters <topics-exporters>` classes.
 
 -   ``overwrite``: whether to overwrite the file if it already exists
     (``True``) or append to its content (``False``).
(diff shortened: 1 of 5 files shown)


maranqz:
The second commit was fail. In third one I do nothing and tests passed. I think it is not correct.

docs/topics/feed-exports.rst
```diff
@@ -332,6 +335,8 @@ as a fallback value if that key is not provided for a specific feed definition:
 
 -   ``indent``: falls back to :setting:`FEED_EXPORT_INDENT`.
 
+-   ``item_export_kwargs``: dict with kwargs for :ref:`Item exporters <topics-exporters>` classes.
```
Gallaecio, line 338: :lipstick: 

```suggestion
-   ``item_export_kwargs``: :class:`dict` with keyword arguments for the corresponding :ref:`item exporter class <topics-exporters>`.
```

Gallaecio (approved):
> In third one I do nothing and tests passed. I think it is not correct.

I don’t know what you mean. If I comment out the feedexport.py line, the new test fails as expected.

As far as I can tell, good work!

maranqz:
> > In third one I do nothing and tests passed. I think it is not correct.
> 
> I don’t know what you mean. If I comment out the feedexport.py line, the new test fails as expected.
> 
> As far as I can tell, good work!

I mean, there is problem with tests.
Testing of commit https://github.com/scrapy/scrapy/pull/4768/commits/fc3c66ce950e945e7ac6e19fea31e8454147ac29 was fail. Then I create https://github.com/scrapy/scrapy/pull/4768/commits/71d2c2f1a319b570eb2026707d49c4f62d59296e without valuable changes and tests passed.

Gallaecio:
> I mean, there is problem with tests.
> Testing of commit [fc3c66c](https://github.com/scrapy/scrapy/commit/fc3c66ce950e945e7ac6e19fea31e8454147ac29) was fail. Then I create [71d2c2f](https://github.com/scrapy/scrapy/commit/71d2c2f1a319b570eb2026707d49c4f62d59296e) without valuable changes and tests passed.

Oh, I see! I thought you meant there was a problem in your code, not in the CI. My bad!

Looks like it was a case of random failure. We have a few of those related to the logging of warnings and errors, sometimes they pick unrelated logging messages causing a test to fail. Nothing for you to worry about.

kmike:
Thanks @maranqz!

PR #4769, opened 2020-08-31
Implement experimental HTTP/2 support

### LINKED ISSUE #1854: HTTP 2 support
I peeked at the docs, at the issues but couldn't find any info about HTTP 2 support.

Does `scrapy` support it?

Gallaecio:
Resolves #1854

This is most of the GSoC 2020 work by @adityaa30. There is additional work which will be handled in separate pull requests.

To do:

- [x] https://github.com/scrapy/scrapy/pull/4769#issuecomment-697141375
- [x] https://github.com/scrapy/scrapy/pull/4769#issuecomment-729635590
- [x] Document the (hard-coded) maximum frame size accepted from servers
- [x] Remove the logging of server-initiated request handling
- [x] Document that server-initiated requests are ignored
- [x] Add `Request.protocol` support (#4878) (done in e494a3f73318db76d7c56e65bc632cd5875b8825)
- [x] Document that the `bytes_received` signal is not implemented
- [x] Make tests pass
- [x] Improve coverage as much as possible before merging for Scrapy 2.5

After merging:
- [ ] Create a milestone for production-ready HTTP/2 support, with the following issues:
    - [ ] Support the `bytes_received` signal
    - [ ] Successfully crawling 1 million popular domains in Zyte Scrapy Cloud (at the moment there are memory issues)
    - [ ] Write tests to cover bde96a5ad98caed74cbaaeaf1dfe4b215093c03f and 08f5ed712f4fdafec122f887c3ef06629f35ee0e
    - [ ] Handle the `dataloss` flag as in the HTTP/1.1 response handler
- [ ] Create additional issues to:
    - [ ] Implement complete code coverage for HTTP/2 code
    - [ ] Create a setting for the maximum frame size accepted from servers
    - [ ] Test handling of servers sending a non-UTF8 string as protocol


wRAR:
Is the diff here the same as in #4610 ?

Gallaecio:
> Is the diff here the same as in #4610 ?

Plus what had been already merged into the `http2` branch before that work.

Gallaecio:
It looks like there is a random error in the Windows CI. It first failed in Python 3.7, then in Python 3.6.

I don’t consider it a blocker, but it would be great if we managed to identify the cause before we merge.

These below were the failures, the source of the issue must be in one of these tests or some test that runs at some point before them.

From the first run (Python 3.7, 745dfc7f):
```
FAILED tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_POST_large_json_x10
FAILED tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_connection_timeout
ERROR tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_connection_timeout
```

From the second run (Python 3.6, 984ccceb):

```
FAILED tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_POST_large_json_x10
FAILED tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_POST_small_json_x10
ERROR tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_POST_small_json_x10
```

adityaa30:
The problem with the `tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_POST_large_json_x10` test could be the large amount of time taken (> 120). Twisted tests has a timeout of 120 sec.. However about the `tests/test_http2_client_protocol.py::Https2ClientProtocolTestCase::test_connection_timeout` test, I'm looking into the possible cause. I did try running the tests multiple times on windows but couldn't get the tests to fail.

wRAR (approved):

Gallaecio:
I was working on making `h2` optional, and only running HTTP/2 tests in a job where it is installed, while making sure that Scrapy works as usual without it installed.

However, it turns out `mitmproxy` requires `h2`.

How should we move forward?

Gallaecio:
I’ve tried to bench mark HTTP/1.1 against HTTP/2.

However, I’ve never used https://github.com/scrapy/scrapy-bench before, and I’m not sure I’m doing it right.

I’ve followed the setup instructions (to the best of my knowledge), using Docker for the server part, and installing vmprof==4.0.9 (https://github.com/vmprof/vmprof-python/issues/223#issuecomment-640130411). Then I edited `followall.py` to add:

```
    custom_settings = {
        'DOWNLOAD_HANDLERS': {
            'https': 'scrapy.core.downloader.handlers.http2.H2DownloadHandler',
        }
    }
```

Running `scrapy-bench --only_result --book_url books.toscrape.com bookworm` (using the online URL because setting up HTTP/2 with HTTPS locally seemed like a pain) yields very similar results with that new code commented out or not. In fact, one is not consistently better than the other.

Again, I’m probably doing something wrong.

Gallaecio:
Oh, the benchmark runs the spider from scratch for each URL. That could explain getting similar times.

Gallaecio:
I extended scrapy-bench with a benchmark designed specifically for this scenario: https://github.com/scrapy/scrapy-bench/pull/38

Using that, these are the results:

```
(venv) ![adrian@vento scrapy-bench]$ scrapy-bench --n-runs 10 --only_result -s 'DOWNLOAD_HANDLERS={"https": "scrapy.core.downloader.handlers.http2.H2DownloadHandler"}' httpbench 

The results of the benchmark are (all speeds in items/sec) : 


Test = 'HTTP Benchmark' Iterations = '10'


Mean : 32.26804019929387 Median : 34.452098469487915 Std Dev : 4.973753192867094

(venv) ![adrian@vento scrapy-bench]$ scrapy-bench --n-runs 10 --only_result httpbench

The results of the benchmark are (all speeds in items/sec) : 


Test = 'HTTP Benchmark' Iterations = '10'


Mean : 41.290344358415474 Median : 40.98786915590955 Std Dev : 2.3032455905131517
```

Gallaecio:
I also wrote a spider to crawl the top 300 websites (according to similarweb.com 2016 data): https://gist.github.com/Gallaecio/7036c62023b543204a702e8ff0adf0b2

My takeaways:

-   There’s a rare `KeyError` issue caught in the [log](https://gist.github.com/Gallaecio/95e43d38fef033a66d09e0621bac58a4). @adityaa30 Any idea, off the top of your head, of what could be causing it? We might want to get it fix before merging; but if we mark the feature as experimental, given it only seems to happen once (or 4 times, depending on how you count), maybe it’s OK to merge anyway and log an issue about it to fix it later.

-  The invalid procotol failures look as expected for sites that are online but do not support HTTP/2, and fail early, which is good. I also verified with https://tools.keycdn.com/http2-test a few of those.

-   60% are online and have HTTP/2 support, which is nice, and a significant improvement [compared to 2016](https://medium.com/@aysunakarsu/http2-on-top-sites-a37fc9cb377f).

adityaa30:
> Any idea, off the top of your head, of what could be causing it?

The error is raised because the stream_id is even (expected to be odd everytime as per [RFC 7540 Sec 5.1.1](https://tools.ietf.org/html/rfc7540#section-5.1.1)). 

Also,  I ran the script multiple times the error was always after sending request `<GET https://gameforge.com/unsupported.html>`. I created another Spider to simply send request to the url however, it works as expected. I am not sure about the source of the error now :thinking: 

Spider to test the url
```python
import scrapy

    name = 'test'

    start_urls = ['https://gameforge.com/unsupported.html'] * 1

    custom_settings = {
        'DOWNLOAD_HANDLERS': {
            'https': 'scrapy.core.downloader.handlers.http2.H2DownloadHandler',
        },
    }


    def parse(self, response):
        print(response.headers)
```

Gallaecio:
If I switch `start_urls` to `start_urls = ['https://gameforge.com']` in your spider, it becomes reproducible.

Gallaecio:
> I was working on making `h2` optional, and only running HTTP/2 tests in a job where it is installed, while making sure that Scrapy works as usual without it installed.
> 
> However, it turns out `mitmproxy` requires `h2`.
> 
> How should we move forward?

After internal discussion we decided that it is OK to add `h2` as a dependency to Scrapy.

Gallaecio:
Running 1M domains, I found a memory issue, and I’ve been able to reproduce the issue with a single domain: raiffeisen.ro

Stopping the spider forcefully outputs something like:

```
…, FrameTooLargeError('Received overlong frame: length 4740180, max 16384'), <twisted.python.failure.Failure twisted.internet.error.ConnectionLost: Connection to the other side was lost in a non-clean fashion: Connection lost.>]
```

Gallaecio:
The `KeyError` issue was caused by server-initiated requests. Those exists in HTTP/2, and we should probably give them some though to decide how to handle them (how does a spider define the callback of a server-initiated request?), but I think ignoring them for the initial implementation should be OK.

Gallaecio:
The memory issue on `raiffeisen.ro` was caused by the server sending a larger frame than it should according to the specification.

The specification says servers should not send more than `16384` in a frame unless the client asks for more. The current implementation does not ask for more, so the server is misbehaving by sending more.

We should consider allowing to set a different value than `16384` through a Scrapy setting, to support such servers or simply for performance tuning. But maybe we should do that in a separate, later change?

PJovy:
Hi Gallaecio, is there any progress about HTTP2 implementation?
@Gallaecio

Gallaecio:
The implementation is not 100% ready yet. I’m testing scrawling 1 million domains, and there are still some memory issues.

That said, we may consider merging the implementation in its current state as experimental, and work out those issues in later versions.

Gallaecio:
So, with the legacy resolver, beyond Flake and expected Twisted version issue, we only got `AnonymousFTPTestCase.test_ftp_download_success` failing on Mac OS X, which is probably unrelated to these changes. So I suspect most issues will be reproducible on `master`. I’ll revert the resolver change, fix the Flake issue, and then see if I can reproduce the issues with the new resolver in `master`.

elacuesta:
@Gallaecio Regarding the memory issues, I'd like to point out a similarity with the `DirtyReactorAggregateError` exception I was getting in #4897 (solved by calling `loseConnection` on the associated producer): https://github.com/scrapy/scrapy/pull/4994/checks?check_run_id=1953112362#step:4:197. It's a long shot but it might be related.

Gallaecio:
@elacuesta I ended up just re-running the tests. I know it’s a patch, but since the issue is not specific to any change here (it’s most likely reproducible in master, given it affects a documentation pull request like https://github.com/scrapy/scrapy/pull/4974), this is probably not the right pull request to fix it.

elacuesta:
Oh I didn't realize other PRs were affected as well, much less one about docs. In that case, there's no reason to hold this one from merging.

scrapy/core/http2/stream.py
```diff
@@ -431,7 +431,8 @@ def close(
             errors.insert(0, InactiveStreamClosed(self._request))
             self._deferred_response.errback(ResponseFailed(errors))
 
-        elif reason is StreamCloseReason.INVALID_HOSTNAME:
+        else:
+            assert reason is StreamCloseReason.INVALID_HOSTNAME
```
elacuesta, line 448: What about #4440? :smile:

Gallaecio: That was triggered by [this comment by me](https://github.com/scrapy/scrapy/pull/4054#discussion_r351514071) (emphasis added):

> We should not use assert statements in production code, outside tests, **where we know the assertion may fail**.

In this case, no input could ever make that `assert` raise an `AssertionError`. The code is written so that this condition will always be `True`. If that ever changes, the cause will be a bug in the code, not unexpected input.

*But*, if you prefer, we could just replace it with a comment besides the `else`. However, if we ever extend the possible values of `StreamCloseReason` and forget to update this part of the code accordingly, an `assert` could cause tests to fail.

elacuesta: Fair enough, thanks for the explanation :slightly_smiling_face:

PR #4770, opened 2020-08-31
Add Negotiation protocol & download handler

adityaa30:


wRAR:
There are some typing errors in the tests, have you checked them already?

Gallaecio:
Looks like some parts of the new code are not covered by tests: https://codecov.io/gh/scrapy/scrapy/pull/4770/diff

adityaa30:
> Looks like some parts of the new code are not covered by tests

@Gallaecio Yes, I will add more tests.

tests/test_downloader_handlers_negotiation.py
```diff
@@ -0,0 +1,206 @@
...
+        requests = []
+        for _ in range(3):
+            requests.append(self.test_secure_request())
+
+        return DeferredList(requests, fireOnOneErrback=True)
+
+
+class NegotiationHttp2TestCase(NegotiationHttp11TestCase):
+    server_acceptable_protocols = [b'h2']
+    expected_negotiated_protocol = 'h2'
```
Gallaecio, line 206: What about a third test for the scenario where the server supports both?

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -0,0 +1,284 @@
...
+from scrapy.http.request import Request
+from scrapy.settings import Settings
+from scrapy.spiders import Spider
+from scrapy.utils.datatypes import LocalCache
+from scrapy.utils.python import to_bytes
+
+SUPPORTED_PROTOCOLS = ('h2', 'http/1.1')
+
+
+class UnsupportedNegotiatedProtocol(Exception):
```
Gallaecio, line 27: If we trust the [coverage data](https://codecov.io/gh/scrapy/scrapy/pull/4770/diff), none of the code below that creates an instance of this class gets executed during tests.

adityaa30: Yes. I am not sure although how to test this, I added this in case `Twisted` adds a support for HTTP/3 Server or a request is made to the website which chooses HTTP/2 during ALPN or NPN. However, this is a very rare case we are sending `SUPPORTED_PROTOCOLS` during negotiation such that only one of HTTP/1 or HTTP/2 is chosen.

Gallaecio: For the HTTP/3 case, I think the ideal solution would be to add a mock server which only supports HTTP/3, and test that scenario.

We could alternatively [mock](https://docs.python.org/3/library/unittest.mock.html) our code in the tests, to behave as it would in the scenarios we devise.

> a request is made to the website which chooses HTTP/2 during ALPN or NPN

Could you elaborate? How does this scenario lead to `UnsupportedNegotiatedProtocol`, given `SUPPORTED_PROTOCOLS = ('h2', 'http/1.1')`?

adityaa30: > Could you elaborate? How does this scenario lead to `UnsupportedNegotiatedProtocol`, given SUPPORTED_PROTOCOLS = ('h2', 'http/1.1')?

Thinking about it. I realize that as long as we specify the `SUPPORTED_PROTOCOLS = ('h2', 'http/1.1')` the exception will never be raised. As a result either no protocol will be negotiated or one of the two. However, I'm not sure about the behaviour of the download handler when no protocol is negotiated. I will add a test for that case now. 

Should we remove the `UnsupportedNegotiatedProtocol`? I do feel now that it will never be raised as we always specify a list of protocols to be negotiated.

Gallaecio: > Should we remove the `UnsupportedNegotiatedProtocol`? I do feel now that it will never be raised as we always specify a list of protocols to be negotiated.

Yes, I think that makes sense. Thanks!

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -0,0 +1,259 @@
...
+    def __init__(self, settings: Settings, crawler=None):
+        self._settings = settings
+        self._crawler = crawler
+
+        self.acceptable_protocols = SUPPORTED_PROTOCOLS
+
+        # ToDo: Maybe load download-handlers from the settings?
+        #  Need to add default for 'h2' in DOWNLOAD_HANDLERS_BASE
+        self._http1_download_handler = HTTP11DownloadHandler(settings, crawler)
+        self._h2_download_handler = H2DownloadHandler(settings, crawler)
```
adityaa30, line 93: Should I load download handler class from the settings?

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -145,8 +125,6 @@ def get_download_handler(self, protocol):
         elif protocol == 'h2':
```
Gallaecio, line 125: Maybe turn this into an `else` and indicate in a comment that the value is expected to be `h2` here? Same in `add_connection`.

scrapy/core/downloader/handlers/negotiation.py
Gallaecio: I don’t consider it a blocker, but yes, that would be ideal.

I imagine we want a new setting that is a dictionary where keys are protocol version strings (e.g. `http1.1`, `h2`) and values are download handler classes (or their import paths). I’m not sure how to call it, though.

adityaa30: > I imagine we want a new setting that is a dictionary where keys are protocol version strings 

Should I add the setting as `HTTP_DOWNLOAD_HANDLERS_BASE`?

Gallaecio: Works for me. We can always rename later if someone can think of a better name.

adityaa30: Okay

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -37,11 +36,12 @@ def handshakeCompleted(self) -> None:
...
 
+@implementer(IProtocolNegotiationFactory)
 class NegotiateProtocolFactory(Factory):
     def __init__(
-        self,
-        acceptable_protocols: List[bytes],
-        negotiated_protocol_deferred: Deferred
+            self,
+            acceptable_protocols: List[bytes],
+            negotiated_protocol_deferred: Deferred
```
Gallaecio, line 44: :lipstick: Why the additional indentation level? I think it was fine before. And I see you’ve made similar changes below. Was it an IDE suggestion or something?

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -85,12 +87,22 @@ def __init__(self, settings: Settings, crawler=None):
         self._settings = settings
         self._crawler = crawler
 
-        self.acceptable_protocols = SUPPORTED_PROTOCOLS
+        self.acceptable_protocols = []
+
+        handlers_base = without_none_values(self._settings.get('HTTP_DOWNLOAD_HANDLERS_BASE'))
```
Gallaecio, line 92: See https://docs.scrapy.org/en/latest/topics/api.html#scrapy.settings.BaseSettings.getwithbase

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -85,12 +87,22 @@ def __init__(self, settings: Settings, crawler=None):
...
+        self.handlers: Dict = {}
+        for protocol, cls_path in handlers_base.items():
+            self.acceptable_protocols.append(protocol)
```
Gallaecio, line 95: Is `self.acceptable_protocols` needed, given there’s `self.handlers`?

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -183,29 +192,31 @@ def add_connection(self, transport, request: Request) -> None:
         negotiated_protocol = transport.negotiatedProtocol.decode()
         key = self._get_key(request)
         uri = self._parse_uri(request.url)
+        handler = self.handlers[negotiated_protocol]
 
         # Create an instance of the connection
         if negotiated_protocol == 'http/1.1':
+            assert isinstance(handler, HTTP11DownloadHandler)
```
Gallaecio, line 196: Is this a debug leftover? It seems it will fail if a user overwrites the handler for HTTP 1.1 with a custom handler that does not inherit from `HTTP11DownloadHandler`.

scrapy/core/downloader/handlers/negotiation.py
```diff
@@ -255,5 +270,5 @@ def make_request(_download_handler: Union[HTTP11DownloadHandler, H2DownloadHandl
         return d
 
     def close(self) -> Deferred:
-        self._h2_download_handler.close()
-        return self._http1_download_handler.close()
+        self.handlers['h2'].close()
```
Gallaecio, line 270: What does this return? Is it also a deferred?

Gallaecio:
@adityaa30 Can you point this to `master` now that we have removed the `http2` branch? Then I can reopen the pull request. I can also create a new pull request, but that would be slightly more messy.

PR #4775, opened 2020-09-02
Documentation of link extractor usage

### LINKED ISSUE #4751: Document link extractor usage outside CrawlSpider rules
https://docs.scrapy.org/en/latest/topics/link-extractors.html mentions that link extractors may be used outside `CrawlSpider`, but it does not go into detail on how to do that.

Also, there are broken references to `scrapy.link.Link`, we should provide reference documentation for that class.

HAKSOAT:
Added description when using link extractor outside crawlspider and created reference documentation for scrapy.link.Link class

Fixes #4751


docs/topics/link-extractors.rst
```diff
@@ -10,13 +10,21 @@ The ``__init__`` method of
...
 <scrapy.linkextractors.lxmlhtml.LxmlLinkExtractor.extract_links>` returns a
-list of matching :class:`scrapy.link.Link` objects from a
+list of matching :class:`~scrapy.link.Link` objects from a
 :class:`~scrapy.http.Response` object.
 
 Link extractors are used in :class:`~scrapy.spiders.CrawlSpider` spiders
 through a set of :class:`~scrapy.spiders.Rule` objects. You can also use link
 extractors in regular spiders.
 
+In other to use link extractors in regular spiders, you should instantiate ``scrapy.linkextractors.LinkExtractor`` into
```
elacuesta, line 20: Two comments:
* "In other" -> "In order"
* "should" is not completely accurate here, storing link extractors as class/instance variables is not mandatory

docs/topics/link.rst
```diff
@@ -0,0 +1,44 @@
...
+
+
+Link
+----
+
+.. module:: scrapy.link
+   :synopsis: Link from link extractors
+
+
+.. class:: Link(url, text='', fragment='', nofollow=False)
```
elacuesta, line 24: Could `autoclass` be used here?

docs/topics/link-extractors.rst
HAKSOAT: Thanks for catching those. Will fix.

docs/topics/link.rst
HAKSOAT: I'm pretty new to documenting with Sphinx. I'll take a look at what an **autoclass** is and get back to you.

HAKSOAT: The docstring for the Link class only contains the words """Link objects represent an extracted link by the LinkExtractor.""" 

I don't think that suffices, do you?

Gallaecio: It does not. But you _can_ move your documentation to the docstring, and _then_ use `autoclass` :slightly_smiling_face:

HAKSOAT: Done. @Gallaecio
Thanks for the suggestion @elacuesta I see it's a better option.

Gallaecio (commented):
Unless we decide to cover other use cases (and I cannot think of any at the moment), I would move the new documentation into the existing topic about link extractors, instead of creating a new page.

HAKSOAT:
> 
> 
> Unless we decide to cover other use cases (and I cannot think of any at the moment), I would move the new documentation into the existing topic about link extractors, instead of creating a new page.

Done. @Gallaecio Can you help take a look at what it looks like now? Perhaps a couple more changes to make it better?

scrapy/link.py
```diff
@@ -7,7 +7,21 @@
...
 
 class Link:
-    """Link objects represent an extracted link by the LinkExtractor."""
+    """Link objects represent an extracted link by the LinkExtractor.
+
+    Using the anchor tag sample below to illustrate the parameters::
+
+            <a href="/nofollow.html#foo" rel="nofollow">Dont follow this one</a>
+
+    :param url: the address being linked to in the anchor tag. From the sample, this is ``base_url/nofollow.html``.
```
Gallaecio, line 16: If `url` is meant to be an absolute URL, as it seems, I would make that explicit here (“address” → “absolute URL”).

As for `base_url`, I would turn it into `<base_url>` to make it a bit more obvious that it is a place holder. Alternatively, you may even consider going further, and suggest that, assuming the HTML document containing the anchor was https:/example.com/index.html, the URL would be ``https://example.com/nofollow.html``.

docs/topics/link-extractors.rst
```diff
@@ -10,13 +10,21 @@ The ``__init__`` method of
...
 extractors in regular spiders.
 
+In order to use link extractors in regular spiders, you can instantiate ``scrapy.linkextractors.LinkExtractor`` into
+a class variable in your spider. This variable can be accessed through the ``parse`` method
+and used to extract links using a format similar to the one below::
+
+    def parse(self, response):
+        for link in self.link_extractor.extract_links(response):
+            yield Request(link.url, callback=self.parse)
+
```
Gallaecio, line 27: :lipstick: What do you think about moving the last sentence above into your new paragraph, and
phrasing it as an example (since this is just _one_ possible approach):

```suggestion
through a set of :class:`~scrapy.spiders.Rule` objects.

You can also use link extractors in regular spiders. For example, you can instantiate
:class:`LinkExtractor <scrapy.linkextractors.lxmlhtml.LxmlLinkExtractor>` into a class
variable in your spider, and use it from your spider callbacks::

    def parse(self, response):
        for link in self.link_extractor.extract_links(response):
            yield Request(link.url, callback=self.parse)

```

HAKSOAT: Looks good. I just applied the suggestion. Since there are other ways to use them, I could try looking them up (I don't know them at the moment) and also add them.

Gallaecio: I think a single example is enough. I meant that users, for example, do not have to keep the link extractor object in the class, and they can use them outside callbacks. You cover the most, which I think is best when providing an example while keeping things short.

scrapy/link.py
```diff
@@ -11,9 +11,9 @@ class Link:
 
     Using the anchor tag sample below to illustrate the parameters::
 
-            <a href="/nofollow.html#foo" rel="nofollow">Dont follow this one</a>
+            <a href="https://example.com/nofollow.html#foo" rel="nofollow">Dont follow this one</a>
```
Gallaecio, line 14: I actually think that it’s good that the URL was not absolute here. We simply need to indicate, either in the preceding paragraph or in the documentation about the `url` parameter below, that the example `url` value assumes that this code comes from `https://example.com`.

Directly making this an absolute URL in the code example hides the fact that the `url` parameter in `Link` must be an absolute URL.

Although it could be argued that saying `absolute URL` instead of `address` below should be enough to clarify that. I guess keeping the absolute URL here but indicating that below would work for me as well.

HAKSOAT: True. I've modified it.

Gallaecio (approved):
Looks good to me, thanks!

kmike:
Thanks @HAKSOAT!

PR #4776, opened 2020-09-02
Remove dead boto2 code, deprecate is_botocore().

### LINKED ISSUE #4734: Remove boto import conditionals from Python 2 times

wRAR:
This:

* deprecates is_botocore() and adds an explanation how does it currently work (return True if `botocore` is available, raise an exception otherwise) and that it cannot help with boto/botocore code path choice.
* adds is_botocore_available() that should be used to check if `botocore` is available.
* for all `if is_botocore(): code1 else: code2` keeps just `code1`, adding `if not is_botocore_available(): raise NotConfigured()` (or `skip_if_no_boto()` in tests) where it makes sense.
* removes two tests which tested some dead code but were not skipped because they mock things.

Note that we currently don't run tests without `botocore` installed and because of that not all tests that need it are properly skipped and fail instead. This PR doesn't fix this.

Consider tagging `discuss` if you think it needs it.

Fixes #4734. Related to #1866.

### DIFF
--- a/scrapy/core/downloader/handlers/s3.py
+++ b/scrapy/core/downloader/handlers/s3.py
@@ -2,33 +2,9 @@ from urllib.parse import unquote
 
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
 from scrapy.exceptions import NotConfigured
-from scrapy.utils.boto import is_botocore
+from scrapy.utils.boto import is_botocore_available
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.misc import create_instance
-
-
-def _get_boto_connection():
-    from boto.s3.connection import S3Connection
-
-    class _v19_S3Connection(S3Connection):
-        """A dummy S3Connection wrapper that doesn't do any synchronous download"""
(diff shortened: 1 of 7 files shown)


wRAR:
Regrading boto3.

There were some discussions about adding boto3 support as a new optional feature or as a replacement. The code that is removed here could probably be of some use for that, but I think it was decided some time ago that it's fine to delete it because:

* boto3 depends on botocore so the logic would be different;
* for the same reason most of the code could keep using botocore without adding new checks;
* there are no current plans to add boto3 support, or any proposed design for that, so keeping the old code is not worthwhile;
* the code that actually uses boto2 would need to be ported or replaced anyway.

Gallaecio (approved):
Looks good to me.

An indeed, not having a job without the dependency prevents some code to be tested. I think that’s something worth addressing eventually, but I see no reason why we would need to do that as part of this PR. I’ll create a separate issue about that.

PS: Sorry, I missed this PR when you created it. I only saw it today by chance looking at issues not updated for the last 30 days :facepalm:

PR #4783, opened 2020-09-09
Improve support for external request queues

whalebot-helmsman:
It is continuation of work from https://github.com/scrapy/scrapy/pull/4605

### DIFF
--- a/docs/topics/request-queues.rst
+++ b/docs/topics/request-queues.rst
@@ -1,8 +1,8 @@
-.. _topics-queues:
+.. _topics-request-queues:
 
-======
-Queues
-======
+==============
+Request Queues
+==============
 
 Scrapy uses queues to schedule requests. By default, a memory-based queue is
 used but if :ref:`crawls should be paused and resumed <topics-jobs>` or more
--- a/docs/topics/request-queues.rst
+++ b/docs/topics/request-queues.rst
@@ -47,8 +47,8 @@ is ``'scrapy.squeues.LifoMemoryQueue'``.
 Disk queue
 ----------
 
-Disk queues allow :ref:`crawls to be paused and resumed <topics-jobs>` and more
-requests than fit in memory can be scheduled. The default for
+With disk queues it is possible to :ref:`pause and resume crawls <topics-jobs>`
+and schedule more requests than fit in memory. The default for
 :setting:`SCHEDULER_DISK_QUEUE` is ``'scrapy.squeues.PickleLifoDiskQueue'``.
 
 Disk queues serialize requests, e.g. using :meth:`pickle.serialize`. If
--- a/docs/topics/request-queues.rst
+++ b/docs/topics/request-queues.rst
@@ -4,11 +4,51 @@
 Request Queues
 ==============
 
-Scrapy uses queues to schedule requests. By default, a memory-based queue is
-used but if :ref:`crawls should be paused and resumed <topics-jobs>` or more
-requests than fit in memory are scheduled, using an external queue (disk
-queue) is necessary. The setting :setting:`SCHEDULER_DISK_QUEUE` determines
-the type of disk queue that will be used by the scheduler.
+Scrapy uses queues to store requests and to decide what request to schedule
+next. The scheduler does not use a specific queue implementation (called
+*downstream queue*) directly but instead the downstream queue is wrapped by a
+priority queue.
+
+Priority Queues
+===============
+
+Priority queues make sure that requests with the highest priority are scheduled
+first. For this purpose, a priority queue uses multiple downstream queues---one
+for each priority. The setting :setting:`SCHEDULER_PRIORITY_QUEUE` determines
+the type of priority queue that will be used by the scheduler.
+
+Scrapy comes with two priority queue implementations:
+
+  * ``'scrapy.pqueues.ScrapyPriorityQueue'``
+  * ``'scrapy.pqueues.DownloaderAwarePriorityQueue'``
+
+The default priority queue is ``'scrapy.pqueues.ScrapyPriorityQueue'`` and works
+best during single-domain crawls (see also
+:ref:`broad-crawls-scheduler-priority-queue`). For crawling multiple different
+domains in parallel the recommended priority queue is
+``'scrapy.pqueues.DownloaderAwarePriorityQueue'``. This priority queue takes
+downloader activity into account: Domains with the least amount of active
+downloads are dequeued first.
+
+Downstream Queues
+=================
+
+By default, the scheduler uses a memory-based queue as a downstream queue but if
+:ref:`crawls should be paused and resumed <topics-jobs>` or more requests than
+fit in memory are scheduled, using an external queue (disk queue) is necessary.
+The setting :setting:`SCHEDULER_DISK_QUEUE` determines the type of disk queue
+that will be used by the scheduler. It is
+``'scrapy.squeues.PickleLifoDiskQueue'`` by default. To enable the use of disk
+queues, the :setting:`JOBDIR` setting has to be defined (see also
+:ref:`jobs-job-directory`).
+
+.. note::
+
+    If :setting:`JOBDIR` is undefined, the scheduler uses a memory-based queue
+    regardless of how :setting:`SCHEDULER_DISK_QUEUE` is configured.
+
+Interface
+---------
 
 If you want to use your own disk queue implementation, it has to conform to
 the following interface:
--- a/docs/topics/request-queues.rst
+++ b/docs/topics/request-queues.rst
@@ -33,9 +33,10 @@ downloads are dequeued first.
 Downstream Queues
 =================
 
-Scrapy differentiates between two types of downstream queues: memory queues and
-disk queues. If the :setting:`JOBDIR` setting is defined, a disk queue is used.
-If it is not defined, a memory queue is used (this is the default).
+Scrapy differentiates between two types of downstream queues (the data
+structures that hold the actual data): memory queues and disk queues. If the
+:setting:`JOBDIR` setting is defined, a disk queue is used.  If it is not
+defined, a memory queue is used (this is the default).
 
 Memory queue
 ------------
@@ -105,8 +106,10 @@ the following interface:
 
    .. method:: pop(self)
 
-      Pops a request from the queue. In case of a temporary problem, ``None``
-      is returned.
+      Pops a request from the queue.
+
+      In case of a temporary problem, ``None`` is returned. In all other cases,
+      an exception is raised.
 
       The helper function :meth:`~scrapy.utils.reqser.request_from_dict` can
       be used to convert the deserialized dict back to a request.
@@ -117,7 +120,8 @@ the following interface:
 
       .. note::
          In case of a temporary error, the method must not raise an exception
-         but return ``None`` instead.
+         but return ``None`` instead. If an exception is raised, the crawling
+         process is halted.
 
    .. method:: close(self)
 
--- a/docs/topics/request-queues.rst
+++ b/docs/topics/request-queues.rst
@@ -33,19 +33,31 @@ downloads are dequeued first.
 Downstream Queues
 =================
 
-By default, the scheduler uses a memory-based queue as a downstream queue but if
-:ref:`crawls should be paused and resumed <topics-jobs>` or more requests than
-fit in memory are scheduled, using an external queue (disk queue) is necessary.
-The setting :setting:`SCHEDULER_DISK_QUEUE` determines the type of disk queue
-that will be used by the scheduler. It is
-``'scrapy.squeues.PickleLifoDiskQueue'`` by default. To enable the use of disk
-queues, the :setting:`JOBDIR` setting has to be defined (see also
-:ref:`jobs-job-directory`).
+Scrapy differentiates between two types of downstream queues: memory queues and
+disk queues. If the :setting:`JOBDIR` setting is defined, a disk queue is used.
+If it is not defined, a memory queue is used (this is the default).
+
+Memory queue
+------------
+
+Memory queues do not require additional configuration or additional storage and
+are therefore used by default. The default for :setting:`SCHEDULER_MEMORY_QUEUE`
+is ``'scrapy.squeues.LifoMemoryQueue'``.
+
+Disk queue
+----------
+
+Disk queues allow :ref:`crawls to be paused and resumed <topics-jobs>` and more
+requests than fit in memory can be scheduled. The default for
+:setting:`SCHEDULER_DISK_QUEUE` is ``'scrapy.squeues.PickleLifoDiskQueue'``.
+
+Disk queues serialize requests, e.g. using :meth:`pickle.serialize`. If
+serialization fails, the scheduler falls back to a memory queue.
 
 .. note::
 
-    If :setting:`JOBDIR` is undefined, the scheduler uses a memory-based queue
-    regardless of how :setting:`SCHEDULER_DISK_QUEUE` is configured.
+    :setting:`JOBDIR` has to be set so that the scheduler uses the disk queue
+    configured by :setting:`SCHEDULER_DISK_QUEUE`.
 
 Interface
 ---------
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -6,6 +6,7 @@ from os.path import join, exists
 
 from queuelib import PriorityQueue
 
+from scrapy.exceptions import TransientError
 from scrapy.utils.misc import load_object, create_instance
 from scrapy.utils.job import job_dir
 from scrapy.utils.deprecate import ScrapyDeprecationWarning
@@ -129,6 +130,10 @@ class Scheduler:
                 self.logunser = False
             self.stats.inc_value('scheduler/unserializable',
                                  spider=self.spider)
+            return
+        except TransientError as e:
+            msg = "Unable to push request to queue: %s"
+            logger.warning(msg, e, exc_info=True, extra={'spider': self.spider})
             return
         else:
             return True
--- a/scrapy/core/scheduler.py
+++ b/scrapy/core/scheduler.py
@@ -6,7 +6,7 @@ from os.path import join, exists
 
 from queuelib import PriorityQueue
 
-from scrapy.exceptions import TransientError
+from scrapy.exceptions import SerializationError, TransientError
 from scrapy.utils.misc import load_object, create_instance
 from scrapy.utils.job import job_dir
 from scrapy.utils.deprecate import ScrapyDeprecationWarning
@@ -120,7 +120,7 @@ class Scheduler:
             return
         try:
             self.dqs.push(request)
-        except ValueError as e:  # non serializable request
+        except SerializationError as e:  # non serializable request
             if self.logunser:
                 msg = ("Unable to serialize request: %(request)s - reason:"
                        " %(reason)s - no more unserializable requests will be"
--- a/scrapy/pqueues.py
+++ b/scrapy/pqueues.py
@@ -58,11 +58,7 @@ class ScrapyPriorityQueue:
         self.key = key
         self.queues = {}
         self.curprio = None
-        self.selfcheck()
         self.init_prios(startprios)
-
-    def selfcheck(self):
-        self.qfactory('selfcheck').close()
 
     def init_prios(self, startprios):
         if not startprios:
@@ -161,13 +157,10 @@ class DownloaderAwarePriorityQueue:
         self.downstream_queue_cls = downstream_queue_cls
         self.key = key
         self.crawler = crawler
-        self.selfcheck()
+
         self.pqueues = {}  # slot -> priority queue
         for slot, startprios in (slot_startprios or {}).items():
             self.pqueues[slot] = self.pqfactory(slot, startprios)
-
-    def selfcheck(self):
-        self.pqfactory('hostname.invalid').close()
 
     def pqfactory(self, slot, startprios=()):
         return ScrapyPriorityQueue(self.crawler,
--- a/scrapy/pqueues.py
+++ b/scrapy/pqueues.py
@@ -58,7 +58,11 @@ class ScrapyPriorityQueue:
         self.key = key
         self.queues = {}
         self.curprio = None
+        self.selfcheck()
         self.init_prios(startprios)
+
+    def selfcheck(self):
+        self.qfactory('selfcheck').close()
 
     def init_prios(self, startprios):
         if not startprios:
@@ -157,10 +161,13 @@ class DownloaderAwarePriorityQueue:
         self.downstream_queue_cls = downstream_queue_cls
         self.key = key
         self.crawler = crawler
-
+        self.selfcheck()
         self.pqueues = {}  # slot -> priority queue
         for slot, startprios in (slot_startprios or {}).items():
             self.pqueues[slot] = self.pqfactory(slot, startprios)
+
+    def selfcheck(self):
+        self.pqfactory('hostname.invalid').close()
 
     def pqfactory(self, slot, startprios=()):
         return ScrapyPriorityQueue(self.crawler,
--- a/scrapy/squeues.py
+++ b/scrapy/squeues.py
@@ -29,6 +29,10 @@ def _serializable_queue(queue_class, serialize, deserialize):
 
     class SerializableQueue(queue_class):
 
+        def __init__(self, path, settings=None, *args, **kwargs):
+            self.settings = settings
+            super(SerializableQueue, self).__init__(path, *args, **kwargs)
+
         def push(self, obj):
             s = serialize(obj)
             super(SerializableQueue, self).push(s)
@@ -47,7 +51,7 @@ def _scrapy_serialization_queue(queue_class):
 
         def __init__(self, crawler, key):
             self.spider = crawler.spider
-            super(ScrapyRequestQueue, self).__init__(key)
+            super(ScrapyRequestQueue, self).__init__(key, crawler.settings)
 
         @classmethod
         def from_crawler(cls, crawler, key, *args, **kwargs):
--- a/scrapy/squeues.py
+++ b/scrapy/squeues.py
@@ -8,6 +8,7 @@ import pickle
 
 from queuelib import queue
 
+from scrapy.exceptions import SerializationError
 from scrapy.utils.reqser import request_to_dict, request_from_dict
 
 
@@ -89,7 +90,7 @@ def _pickle_serialize(obj):
     # Both pickle.PicklingError and AttributeError can be raised by pickle.dump(s)
     # TypeError is raised from parsel.Selector
     except (pickle.PicklingError, AttributeError, TypeError) as e:
-        raise ValueError(str(e)) from e
+        raise SerializationError(str(e)) from e
 
 
 PickleFifoDiskQueueNonRequest = _serializable_queue(
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -229,6 +229,7 @@ Extending Scrapy
    topics/api
    topics/signals
    topics/exporters
+   topics/queues
 
 
 :doc:`topics/architecture`
@@ -251,6 +252,9 @@ Extending Scrapy
 
 :doc:`topics/exporters`
     Quickly export your scraped items to a file (XML, CSV, etc).
+
+:doc:`topics/queues`
+    Store requests in an external queue.
 
 
 All the rest
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -229,7 +229,7 @@ Extending Scrapy
    topics/api
    topics/signals
    topics/exporters
-   topics/queues
+   topics/request-queues
 
 
 :doc:`topics/architecture`
@@ -253,8 +253,8 @@ Extending Scrapy
 :doc:`topics/exporters`
     Quickly export your scraped items to a file (XML, CSV, etc).
 
-:doc:`topics/queues`
-    Store requests in an external queue.
+:doc:`topics/request-queues`
+    Customizing the queuing of requests.
 
 
 All the rest
--- a/docs/topics/jobs.rst
+++ b/docs/topics/jobs.rst
@@ -17,6 +17,8 @@ facilities:
 * an extension that keeps some spider state (key/value pairs) persistent
   between batches
 
+.. _jobs-job-directory:
+
 Job directory
 =============
 
--- a/docs/topics/queues.rst
+++ b/docs/topics/queues.rst
@@ -1,0 +1,83 @@
+.. _topics-queues:
+
+======
+Queues
+======
+
+Scrapy uses queues to schedule requests. By default, a memory-based queue is
+used but if :ref:`crawls should be paused and resumed <topics-jobs>` or more
+requests than fit in memory are scheduled, using an external queue (disk
+queue) is necessary. The setting :setting:`SCHEDULER_DISK_QUEUE` determines
+the type of disk queue that will be used by the scheduler.
+
+If you want to use your own disk queue implementation, it has to conform to
+the following interface:
+
+.. class:: MyExternalQueue
+
+   .. classmethod:: from_crawler(cls, crawler, key)
+
+      Creates a new queue object based on ``crawler`` and ``key``.
+
+      This factory method receives the ``crawler`` argument to access the
+      crawler's settings and the ``key`` argument which identifies the queue.
+      The class method creates and returns a queue object based on the
+      arguments.
+
+      The method is expected to verify the arguments and the relevant settings
+      and raise an exception in case of an error. This may involve opening
+      a connection to a remote service.
+
+      .. note::
+         In case an exception is raised, the crawling process is halted.
+
+      :raises Exception: If ``key`` or a queue-specific setting is invalid.
+
+   .. method:: push(self, request)
+
+      Pushes a request to the queue.
+
+      The helper function :meth:`~scrapy.utils.reqser.request_to_dict` can be
+      used to convert the request to a dict which can then be easily
+      serialized with, for example, :meth:`pickle.dumps`.
+
+      The scheduler will fall back to the memory queue (for this particular
+      request) in case of a :exc:`TransientError` or a :exc:`ValueError`. In
+      case of any other exception the crawling process is halted.
+
+      :raises TransientError: If pushing to the queue failed due to a
+          temporary error (e.g. the connection was dropped).
+      :raises ValueError: If pushing to the queue failed because the request
+          could not be serialized.
+
+   .. method:: pop(self)
+
+      Pops a request from the queue. In case of a temporary problem, ``None``
+      is returned.
+
+      The helper function :meth:`~scrapy.utils.reqser.request_from_dict` can
+      be used to convert the deserialized dict back to a request.
+
+      It is up to the queue implementation to decide if the most recently
+      pushed value (LIFO) or the least recently pushed value (FIFO) is
+      returned.
+
+      .. note::
+         In case of a temporary error, the method must not raise an exception
+         but return ``None`` instead.
+
+   .. method:: close(self)
+
+      Releases internal resources (e.g. closes a file or socket).
+
+   .. method:: __len__(self)
+
+      Returns the number of elements in the queue.
+
+      If the number of elements cannot be determined (e.g. because of a
+      connection problem), the method must not return 0 because this would
+      cause the queue to be closed.
+
+      .. note::
+         In case of a temporary error, the method must not raise an exception
+         but return the number of elements instead.
--- a/docs/topics/queues.rst
+++ b/docs/topics/queues.rst
@@ -42,13 +42,14 @@ the following interface:
       serialized with, for example, :meth:`pickle.dumps`.
 
       The scheduler will fall back to the memory queue (for this particular
-      request) in case of a :exc:`TransientError` or a :exc:`ValueError`. In
-      case of any other exception the crawling process is halted.
+      request) in case of a :exc:`TransientError` or a
+      :exc:`SerializationError`. In case of any other exception the crawling
+      process is halted.
 
       :raises TransientError: If pushing to the queue failed due to a
           temporary error (e.g. the connection was dropped).
-      :raises ValueError: If pushing to the queue failed because the request
-          could not be serialized.
+      :raises SerializationError: If pushing to the queue failed because the
+          request could not be serialized.
 
    .. method:: pop(self)
 
--- a/scrapy/exceptions.py
+++ b/scrapy/exceptions.py
@@ -27,6 +27,11 @@ class TransientError(Exception):
     pass
 
 
+class SerializationError(ValueError):
+    """Indicates an error with the serialization, usually of a request."""
+    pass
+
+
 # HTTP and crawling
 
 
--- a/scrapy/exceptions.py
+++ b/scrapy/exceptions.py
@@ -21,6 +21,12 @@ class _InvalidOutput(TypeError):
     pass
 
 
+class TransientError(Exception):
+    """Indicates a temporary problem (e.g. connection error) that is likely to
+    not be permanent."""
+    pass
+
+
 # HTTP and crawling
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,9 +52,6 @@ ignore_errors = True
 [mypy-scrapy.spidermiddlewares.referer]
 ignore_errors = True
 
-[mypy-scrapy.squeues]
-ignore_errors = True
-
 [mypy-scrapy.utils.httpobj]
 ignore_errors = True
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,9 @@ ignore_errors = True
 [mypy-scrapy.spidermiddlewares.referer]
 ignore_errors = True
 
+[mypy-scrapy.squeues]
+ignore_errors = True
+
 [mypy-scrapy.utils.httpobj]
 ignore_errors = True
 
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -1,7 +1,7 @@
+import collections
 import shutil
 import tempfile
 import unittest
-import collections
 
 from twisted.internet import defer
 from twisted.trial.unittest import TestCase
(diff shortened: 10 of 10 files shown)


wRAR:
Is this ready for reviewing or WIP?

whalebot-helmsman:
It is ready for review

scrapy/core/scheduler.py
```diff
@@ -119,7 +120,7 @@ def _dqpush(self, request):
             return
         try:
             self.dqs.push(request)
-        except ValueError as e:  # non serializable request
+        except SerializationError as e:  # non serializable request
```
Gallaecio, line 123: In the topic of backward compatibility, should we consider to continue catching `ValueError` but log a deprecation warning when it is caught?

scrapy/squeues.py
```diff
@@ -29,6 +30,10 @@ def _serializable_queue(queue_class, serialize, deserialize):
 
     class SerializableQueue(queue_class):
 
+        def __init__(self, path, settings=None, *args, **kwargs):
```
Gallaecio, line 33: Any chance we can make this backward compatible?

whalebot-helmsman:
Hey @Gallaecio I addressed both of your comments

scrapy/squeues.py
```diff
@@ -85,26 +86,47 @@ def _pickle_serialize(obj):
...
     except (pickle.PicklingError, AttributeError, TypeError) as e:
-        raise ValueError(str(e)) from e
+        raise SerializationError(str(e)) from e
+
+
+def _ignore_args_kwargs_passed_to_constructor(queue_class, used_kwargs):
+    class AcceptingQueue(queue_class):
+        def __init__(self, path, *_, **kwargs):
+            new_kwargs = {k: kwargs[k] for k in used_kwargs if k in kwargs}
+            super().__init__(path, **new_kwargs)
```
Gallaecio, line 96: Won’t this cause `chunksize` to now get ignored if it used to be based as a positional argument instead of a keyword argument?

whalebot-helmsman: Yes, you are right. But this very complex question. Single place where `chanksize=smth` is used in scrapy is in `tests/test_squeue.py`. It is used as named argument. Chunking behaviour tested there is not an attribute of scrapy and tested in `queuelib`. My proposition is to remove these chnunking tests. After (or in this PR) we can present something like `SCHEDULER_DISK_QUEUE_CHUNKSIZE` into scrapy's settings.

wRAR (approved):

scrapy/squeues.py
Gallaecio: My worry is backward compatibility.

Ideally, `chunksize` should continue to work as a positional argument, and if we want that to stop working in the future, we should emit the corresponding warning.

If we consider it extremely unlikely that this backward-incompatible change will affect anyone, we can simply acknowledge it in the release notes and go ahead. But we need to remember to cover it in the release notes.

whalebot-helmsman: > Ideally, chunksize should continue to work as a positional argument, and if we want that to stop working in the future

There was no places in scrapy where this argument was used. Nothing to continue heree

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+   .. classmethod:: from_crawler(cls, crawler: scrapy.crawler.Crawler, key: str, *args, **kwargs)
+
+      Return an instance of this disk downstream queue class.
+
+      *key* is the unique ID of the disk downstream queue instance. It may be
+      used, for example, to create a unique file or folder to store the queue
+      content.
+
+      If the input data is invalid, raise an exception from this class method
+      of from your ``__init__`` method to halt the crawl.
```
Gallaecio, line 98: We should describe the content and purpose of `*args` and `**kwargs` here.

scrapy/squeues.py
whalebot-helmsman: Hey @Gallaecio, implementation is provided  to take `chunksize` from `crawle.settings` (without docs etc.). Just to check validity of idea.

Gallaecio: I’m not against the idea, but if you are suggesting that as a solution for the backward-incompatibility issue, I _think_ (my memory about this patch has almost faded) it does not address it (we still had a `parameter` which is not expected anymore, so if used things will fail).

However, as [I said before](https://github.com/scrapy/scrapy/pull/4783#discussion_r511587394), I don’t mind if we do make the backward-compatibility change that is unlikely to break any existing code, provided we do mention it in the release notes just in case.

whalebot-helmsman: > Does MarshalFifoDiskQueueNonRequest.__init__ accept a chunksize parameter after the last change?

No, now they accept `crawler` argument (as other scrapy objects) and read `chunksize` from `crawler.settings['SCHEDULER_DISK_QUEUE_CHUNKSIZE']`

>  provided we do mention it in the release notes just in case.

Should I create a section for future release in `docs/news.rst`?

Gallaecio: > Should I create a section for future release in docs/news.rst?

While we usually look for the backward-incompatible label when covering issues in the release notes, it may be good to do that to make sure we do not forget. You don’t even need to work to hard on in, we just need to make sure that, when updating the release notes, we do not forget about covering this.

docs/topics/broad-crawls.rst
```diff
@@ -39,18 +39,16 @@ you need to keep in mind when using Scrapy for doing broad crawls, along with
...
-Scrapy’s default scheduler priority queue is ``'scrapy.pqueues.ScrapyPriorityQueue'``.
-It works best during single-domain crawl. It does not work well with crawling
-many different domains in parallel
-
-To apply the recommended priority queue use::
+When crawling multiple domains, setting the :setting:`SCHEDULER_PRIORITY_QUEUE`
+setting to :class:`~scrapy.pqueues.DownloaderAwarePriorityQueue` can
+significantly reduce crawl time.
 
-    SCHEDULER_PRIORITY_QUEUE = 'scrapy.pqueues.DownloaderAwarePriorityQueue'
```
kmike, line 53: I think that's nice to keep an example which users can copy-paste, even if it is redundant.

docs/topics/settings.rst
```diff
@@ -384,7 +384,7 @@ Default: ``0``
 
 Scope: ``scrapy.spidermiddlewares.depth.DepthMiddleware``
 
-An integer that is used to adjust the :attr:`~scrapy.http.Request.priority` of
+An integer that is used to adjust the priority of
```
kmike, line 387: What was the reason to change it?

scrapy/core/scheduler.py
```diff
@@ -107,7 +108,23 @@ def _dqpush(self, request):
             return
         try:
             self.dqs.push(request)
-        except ValueError as e:  # non serializable request
+        except TransientError as e:
+            msg = "Unable to push request to queue: %s"
+            logger.warning(msg, e, exc_info=True, extra={'spider': self.spider})
```
kmike, line 113: What do you think about adding a stats counter for this, similar to `scheduler/unserializable`?

scrapy/utils/reqser.py
```diff
@@ -4,15 +4,17 @@
...
 
 from scrapy.http import Request
+from scrapy.spiders import Spider
 from scrapy.utils.python import to_unicode
 from scrapy.utils.misc import load_object
 
 
-def request_to_dict(request, spider=None):
-    """Convert Request object to a dict.
+def request_to_dict(request: Request, spider: Spider = None):
```
kmike, line 12:
```suggestion
def request_to_dict(request: Request, spider: Optional[Spider] = None) -> dict:
```

this would need Optional import. See also: https://github.com/python/mypy/issues/9091

scrapy/utils/reqser.py
```diff
@@ -40,11 +42,11 @@ def request_to_dict(request, spider=None):
     return d
 
 
-def request_from_dict(d, spider=None):
-    """Create Request object from a dict.
+def request_from_dict(d: dict, spider: Spider = None) -> Request:
```
kmike, line 45: spider should be annotated as Optional. 

I also wonder about Request type annotation; subclasses may also be returned, but if someone expects a subclass to be returned by request_from_dict, mypy would be raising an error. I'm on fence here; it can be ok, as users can workaround it by using typing.cast. Thoughts?

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+
+   .. method:: push(self, request: scrapy.http.Request)
```
kmike, line 100: what's the return value? If it is None, it could make sense to annotate it as None

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+
+        -   :exc:`SerializationError`: indicates that *request* could not be
+            serialized
+
+   .. method:: pop(self)
+
+      Pop a request from the queue.
+
+      In case of a temporary problem, ``None`` is returned. In all other cases,
+      an exception is raised, causing the crawling process to halt.
```
kmike, line 124: Would it make sense to annotate output type as `Optional[Request]`?

kmike, line 124:
```suggestion
      In case of a temporary problem, ``None`` is returned.
      Exceptions raised in this method causes crawler to halt.
```
Though ^^ is still not super clear :) What does "crawling process to halt" means? Does it mean a spider gets shut down (but other spiders running in the same process can continue), that the Python process exits? If it is spider who shuts down, is it a hard shutdown or a soft one, i.e. are requests in Downloader still get processed?

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+      Release internal resources (e.g. close files or sockets).
+
+   .. method:: __len__(self)
+
+      Return the number of requests in the queue.
+
+      If the number of requests cannot be determined (e.g. because of a
+      connection problem), the method should neither return 0, which would
+      cause the queue to be closed, nor raise an exception, which would halt
+      the crawl.
```
kmike, line 141: What should it do then? Should it return None, or 1, or a large number?

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+   .. method:: pop(self)
+
+      Pop a request from the queue.
+
+      In case of a temporary problem, ``None`` is returned. In all other cases,
+      an exception is raised, causing the crawling process to halt.
+
+      The helper function :func:`~scrapy.utils.reqser.request_from_dict` can
+      be used to convert a deserialized dict back into a
+      :class:`~scrapy.http.Request` object.
```
kmike, line 128: What should `pop` method return? Is it deserialized dict (and so the caller should call request_from_dict), or is it a Request object (and so this comment is a suggestion on how to implement `pop` method)?

scrapy/core/scheduler.py
```diff
@@ -107,7 +108,23 @@ def _dqpush(self, request):
             return
         try:
             self.dqs.push(request)
-        except ValueError as e:  # non serializable request
+        except TransientError as e:
+            msg = "Unable to push request to queue: %s"
```
kmike, line 112: Hm, does `%s` work here? Which argument does it pick up to do the string formatting?

scrapy/pqueues.py
```diff
@@ -130,9 +111,16 @@ def _active_downloads(self, slot):
 
 
 class DownloaderAwarePriorityQueue:
-    """ PriorityQueue which takes Downloader activity into account:
-    domains (slots) with the least amount of active downloads are dequeued
-    first.
+    """:ref:`Priority queue <priority-queues>` that keeps a similar number of
+    on-going requests per domain.
```
kmike, line 115: I think that'd be nice to keep mentioning slots here, it is not necessarily per domain.

scrapy/pqueues.py
```diff
@@ -26,27 +26,8 @@ def _path_safe(text):
...
-    Only integer priorities should be used. Lower numbers are higher
-    priorities.
-
-    startprios is a sequence of priorities to start with. If the queue was
-    previously closed leaving some priority buckets non-empty, those priorities
-    should be passed in startprios.
-
-    """
+    """:ref:`Priority queue <priority-queues>` that sends requests based on the
+    :attr:`~scrapy.http.Request` priority value."""
```
kmike, line 30: This removes documentation about startprios, and how priorities work in general. What was the reason for this change?

scrapy/settings/default_settings.py
```diff
@@ -258,6 +258,7 @@
 SCHEDULER_DISK_QUEUE = 'scrapy.squeues.PickleLifoDiskQueue'
 SCHEDULER_MEMORY_QUEUE = 'scrapy.squeues.LifoMemoryQueue'
 SCHEDULER_PRIORITY_QUEUE = 'scrapy.pqueues.ScrapyPriorityQueue'
+SCHEDULER_DISK_QUEUE_CHUNKSIZE = 100000
```
kmike, line 261: Maybe it was discussed already, but why is it an undocumented Scrapy setting, not just a constant? Is there a reason for user to change it? If yes, we should probably document it. If no, I'd just hardcode it - make it a constant, not a setting.

There can be disk queues which don't need chunksize (e.g. SQLite ones in queuelib), and I'm worried about having undocumented settings - this happened for some options in past, and it was quite hard to figure out what are they used for, if they are used in a real world, and if we can deprecate & remove them or not.

scrapy/squeues.py
```diff
@@ -8,21 +8,23 @@
...
 from scrapy.utils.reqser import request_to_dict, request_from_dict
 
 
 def _with_mkdir(queue_class):
 
-    class DirectoriesCreated(queue_class):
+    class DirectoriesCreatedForKey(queue_class):
 
-        def __init__(self, path, *args, **kwargs):
+        def __init__(self, _, key, *args, **kwargs):
```
kmike, line 19: nitpick: for me the could would be easier to read if `crawler` name is used instead of `_`. It took me some time to figure out what is this `_` which is passed through.

scrapy/squeues.py
```diff
@@ -85,26 +87,56 @@ def _pickle_serialize(obj):
...
+        raise SerializationError(str(e)) from e
+
+
+def _chunked_file_queue(queue_class):
+    class ChunckedFileQueue(queue_class):
+
+        def __init__(self, crawler, key, *args, **kwargs):
+            chunksize = crawler.settings.getint(
+                'SCHEDULER_DISK_QUEUE_CHUNKSIZE',
+                100000,
```
kmike, line 99: We have this value defined in 3 places now: 

* here
* in default_settings
* in queuelib: https://github.com/scrapy/queuelib/blob/dc078fd3dfafafcf683dda13818664418d8e726e/queuelib/queue.py#L49

It could be good to have a single default value. What's the use case of changing queuelib's default?

scrapy/utils/reqser.py
whalebot-helmsman: It is impossible to guarantee anything except `Request` here. If user expect subclass she should explicitly check it's type.

scrapy/settings/default_settings.py
whalebot-helmsman: > Is there a reason for user to change it?
 
There is no reason for user to change it and there was no way to do it before this PR. 
It was presented as, there aree several tests in `tests/test_squeues.py` testing behavior based on different values of this  argument.

I see several ways here:
1. remove option/argument together with tests 
2. provide documentation for option

Taking into account how much attention comparing to it's importance this option collected, I would prefer first variant.

docs/topics/request-queues.rst
whalebot-helmsman: Typing is added and it's clear now

scrapy/settings/default_settings.py
kmike: Agreed, +1 to the first approach.

docs/topics/settings.rst
whalebot-helmsman: There is no such attribute in documentation https://github.com/scrapy/scrapy/blob/master/docs/topics/request-response.rst

scrapy/squeues.py
whalebot-helmsman: Resolved in https://github.com/scrapy/scrapy/pull/4783#discussion_r619700203

scrapy/core/scheduler.py
whalebot-helmsman: It picks exception message, e.g. for https://github.com/whalebot-helmsman/scrapy/blob/dbf2711e39876bad1077e49a38106d9473977864/tests/test_scheduler.py#L420 message will be something like
```
WARNING  scrapy.core.scheduler:scheduler.py:284 Unable to push request to queue: This class raises an Tr
ansientError                                                                                            
  File "/home/nikita/dev/scrapy/scrapy/core/scheduler.py", line 281, in _dqpush                         
    self.dqs.push(request)                                                                              
  File "/home/nikita/dev/scrapy/scrapy/pqueues.py", line 80, in push                                    
    q.push(request)  # this may fail (eg. serialization error)                                          
  File "/home/nikita/dev/scrapy/tests/test_scheduler.py", line 378, in push                             
    raise TransientError('This class raises an TransientError')                                         
scrapy.exceptions.TransientError: This class raises an TransientError 
```

scrapy/pqueues.py
```diff
@@ -26,17 +26,8 @@ def _path_safe(text):
...
-    """A priority queue implemented using multiple internal queues (typically,
-    FIFO queues). It uses one internal queue for each priority value. The internal
-    queue must implement the following methods:
-
-        * push(obj)
-        * pop()
-        * close()
-        * __len__()
-
-    Optionally, the queue could provide a ``peek`` method, that should return the
```
kmike, line 38: Queue classes got a new optional method recently, see https://github.com/scrapy/scrapy/pull/5112; I think it should be mentioned - not here, but in the queue interface docs.

scrapy/pqueues.py
```diff
@@ -146,9 +135,16 @@ def _active_downloads(self, slot):
...
-    domains (slots) with the least amount of active downloads are dequeued
-    first.
+    """:ref:`Priority queue <priority-queues>` that keeps a similar number of
+    on-going requests per domain (slot).
+
+    This queue can reduce the crawl time significantly when targeting multiple
+    domains.
+
+    :attr:`~scrapy.http.Request` priority is only taken into account among
+    requests targeting the same domain.
```
kmike, line 145:
```suggestion
    requests targeting the same domain (slot).
```

scrapy/pqueues.py
whalebot-helmsman: https://github.com/scrapy/scrapy/pull/5112 doesn't provide a motivation(intended usage) for method introduction. Without motivation It is hard to come up with proper protocol. Hey @elacuesta can you provide more information here? Most important part is what should we do in case of transient errors during peek?

elacuesta: My personal motivation to add this (first in https://github.com/scrapy/queuelib/pull/38) was solving a specific problem I'm having in [scrapy-playwright](https://github.com/scrapy-plugins/scrapy-playwright), which is not really relevant for scrapy itself, at least at the moment.
The method is not currently used by any internal scrapy component, and I don't expect it to be used in the near future, given how long scrapy has managed without it. In any case, I think it's a pretty standard part of any queue interface, and it's just there in case any custom code could benefit from it (even more so after documenting the scheduler at #3559). Additionally, some scrapy tests inherit from the queuelib test suite and were failing because of the absence of `peek` (fixed in the mentioned PR, #5112).
Regarding `TransientError`, if I understand correctly that's something that could be raised within `push`, I don't think we need to consider it for `peek`, which should probably mimic whatever `pop` does (return `None` as documented in this PR).

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,145 @@
...
+
+        -   :exc:`SerializationError`: indicates that *request* could not be
+            serialized
+
+   .. method:: pop(self) -> Optional[Request]
+
+      Pop a request from the queue.
+
+      In case of a temporary problem, ``None`` is returned. Exceptions raised
+      in this method woudln't be handled and may lead to halt of the crawler.
```
elacuesta, line 124:
```suggestion
      in this method wouldn't be handled and may lead to halt of the crawler.
```

scrapy/core/scheduler.py
kmike: TIL :)

tests/test_squeues.py
```diff
@@ -49,35 +49,19 @@ def test_serialize(self):
...
-
-class ChunkSize3MarshalFifoDiskQueueTest(MarshalFifoDiskQueueTest):
-    chunksize = 3
-
-
-class ChunkSize4MarshalFifoDiskQueueTest(MarshalFifoDiskQueueTest):
-    chunksize = 4
 
+class MarshalFifoDiskQueueTest(FifoDiskQueueTestMixin):
+    chunksize = 100000
```
kmike, line 56: is this attribute used now?

docs/topics/request-queues.rst
kmike: it seems this comment is still valid

kmike (approved):

docs/topics/request-queues.rst
whalebot-helmsman: Added a notion about queue internal state. Problem is storing/restoring is not supported in `ScrapyPriorityQueue`. Adding support isn't very hard, but there two reasons not implementing it right now: 
- it breaks on disk format
- implementing in backward-compatible way takes more time and postpone this PR even more
For future it is important to preserve it in protocol. In meantime a lot of good things are possible right now for queues not required storing/restoring.

What is your thoughts?

whalebot-helmsman: I thought about it little more. I don't like current protocol for state passing. Let's me think more and came up with better solution.

whalebot-helmsman: Hey @Gallaecio and @kmike I finished state passing protocol and implementation for `ScrapyPriorityQueue` with support for loading from obsolete format.

.github/workflows/checks.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         - python-version: 3.8
           env:
             TOXENV: pylint
+            TOX_PIP_VERSION: 20.3.3
```
Gallaecio, line 21: Why is this necessary? Is this something we should be addressing in a separate pull request, maybe?

conftest.py
```diff
@@ -21,10 +21,11 @@ def _py_files(folder):
...
 
-for line in open('tests/ignores.txt'):
-    file_path = line.strip()
-    if file_path and file_path[0] != '#':
-        collect_ignore.append(file_path)
+with open('tests/ignores.txt') as reader:
+    for line in reader:
+        file_path = line.strip()
+        if file_path and file_path[0] != '#':
+            collect_ignore.append(file_path)
```
Gallaecio, line 28: Is this change necessary? I like the style of your version better, but if it’s just an style change, given the extend of this pull request, maybe it would be best to minimize unrelated style changes.

scrapy/core/scheduler.py
```diff
@@ -7,6 +7,7 @@
 
 from twisted.internet.defer import Deferred
 
+from scrapy.exceptions import SerializationError, TransientError
 from scrapy.crawler import Crawler
```
Gallaecio, line 11: :lipstick: 

```suggestion
from scrapy.crawler import Crawler
from scrapy.exceptions import SerializationError, TransientError
```

scrapy/extensions/telnet.py
```diff
@@ -69,7 +69,7 @@ def start_listening(self):
                     extra={'crawler': self.crawler})
 
     def stop_listening(self):
-        self.port.stopListening()
+        return self.port.stopListening()
```
Gallaecio, line 72: This seems unrelated to this pull request.

scrapy/pqueues.py
```diff
@@ -47,9 +38,7 @@ class ScrapyPriorityQueue:
 
     startprios is a sequence of priorities to start with. If the queue was
     previously closed leaving some priority buckets non-empty, those priorities
-    should be passed in startprios.
-
-    """
+    should be passed in startprios."""
```
Gallaecio, line 41: :lipstick:  The separate """ on a new line was in line with https://www.python.org/dev/peps/pep-0257/

scrapy/pqueues.py
```diff
@@ -63,21 +52,36 @@ def __init__(self, crawler, downstream_queue_cls, key, startprios=()):
         self.curprio = None
         self.init_prios(startprios)
 
+    @classmethod
+    def bring_prios_up_to_date(_, startprios):
```
Gallaecio, line 56: Are you sure you want to make this method public? It may be better to prefix it with `_` and not need to worry about keeping it backward-compatible in the future.

scrapy/pqueues.py
```diff
@@ -63,21 +52,36 @@ def __init__(self, crawler, downstream_queue_cls, key, startprios=()):
...
+
+        return [(x, None) for x in startprios]
+
     def init_prios(self, startprios):
         if not startprios:
             return
 
-        for priority in startprios:
-            self.queues[priority] = self.qfactory(priority)
+        startprios = type(self).bring_prios_up_to_date(startprios)
```
Gallaecio, line 71: Why not `self.bring_prios_up_to_date(startprios)`?

scrapy/pqueues.py
```diff
@@ -63,21 +52,36 @@ def __init__(self, crawler, downstream_queue_cls, key, startprios=()):
...
+        def _is_valid(x):
+            return isinstance(x, (tuple, list)) and len(x) == 2
+
+        is_valid = all(map(_is_valid, startprios))
+        if is_valid:
+            return startprios
+
+        return [(x, None) for x in startprios]
+
     def init_prios(self, startprios):
```
Gallaecio, line 67: I’m not 100% sure, since quite some time has passed since I reviewed this pull request the last time and I’m not 100% sure of what has changed since then, but I think the fact that `startprios` is a different structure now may be a backward compatible change that, if unavoidable, we should at least document. I don’t think `bring_prios_up_to_date` makes up for it, since subclasses implementing their own `init_prios` are unlikely to be using `super`.

scrapy/pqueues.py
```diff
@@ -105,12 +109,6 @@ def pop(self):
         return m
 
     def peek(self):
-        """Returns the next object to be returned by :meth:`pop`,
-        but without removing it from the queue.
-
-        Raises :exc:`NotImplementedError` if the underlying queue class does
-        not implement a ``peek`` method, which is optional for queues.
-        """
```
Gallaecio, line 113: Accidental removal of documentation?

conftest.py
whalebot-helmsman: Yes, without `with` in raises `ResourceNotReleased` warning on pypy interpreter. It breaks some tests.

.github/workflows/checks.yml
whalebot-helmsman: I created a separate [PR with test fixes](https://github.com/scrapy/scrapy/pull/5160)

conftest.py
whalebot-helmsman: I created a separate [PR with test fixes](https://github.com/scrapy/scrapy/pull/5160)

scrapy/extensions/telnet.py
whalebot-helmsman: Again, without `return` raises ResourceNotReleased warning on pypy interpreter. It breaks some tests.

scrapy/pqueues.py
whalebot-helmsman: It isn't accidental. Corresponding documentation moved to `docs/topics/request-queues.rst`

.github/workflows/checks.yml
whalebot-helmsman: Without PR mentioned above tests for this(or any other) PR doesn't become green

scrapy/pqueues.py
whalebot-helmsman: Format changed only for `ScrapyPriorityQueue`, not every pq in the wild.

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+
+   .. classmethod:: __init__(cls, crawler: scrapy.crawler.Crawler, key: str, state)
+
+      Return an instance of this disk downstream queue class.
+
+      *key* is the unique ID of the disk downstream queue instance. It may be
+      used, for example, to create a unique file or folder to store the queue
+      content.
+
+      *arg* and *kwargs* are used to restore internal state.
```
wRAR, line 97: They are no longer included in the signature, should this be removed?

whalebot-helmsman: fixed

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,157 @@
...
+A :ref:`priority queue <priority-queues>` uses queues internally to store
+pending requests. These internal queues are called *downstream queues*.
+
+The organization method used by the downstream queues, usually either
+FIFO or LIFO, affects the order in which requests are sent for requests
+that the :ref:`priority queue <priority-queues>` considers to have the same
+priority.
+
+The type of downstream queue used depends on the :setting:`JOBDIR` setting (see
+:ref:`topics-jobs`):
```
Gallaecio, line 46:
```suggestion
The built-in :ref:`priority queues <priority-queues>` allow you to choose the 
type of downstream queue they use through a setting. Which setting, however, 
depends on the value of the :setting:`JOBDIR` setting (see :ref:`topics-jobs`):
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,157 @@
...
+--------------------------------------------
+
+If you want to define and use your own disk downstream queue implementation,
+it has to conform to the following interface:
+
+.. class:: MyExternalQueue
+
+   .. classmethod:: __init__(cls, crawler: scrapy.crawler.Crawler, key: str, state)
+
+      Return an instance of this disk downstream queue class.
```
Gallaecio, line 91: `__init__` is not a `@classmethod`, and it does not “return” anything.

Also, these parameters seems to be based on the custom queue class having defined a `from_crawler` method, which it does not need to define. We should cover `from_crawler` in this class documentation as an optional classmethod, similar to how it’s covered for other Scrapy components (e.g. item pipelines). But the documentation of this `__init__` method cannot assume a `from_crawler` is defined, and so it cannot assume that the `crawler` parameter is received.

whalebot-helmsman: > Also, these parameters seems to be based on the custom queue class having defined a from_crawler method

`crawler` argument is used by downstream queues to get settings etc.  

> and so it cannot assume that the crawler parameter is received.

Here I didn't understand why we cannot assume argument of function isn't passed?

whalebot-helmsman: > __init__ is not a @classmethod

Which mark should we use here for proper formatting?

Gallaecio: > `crawler` argument is used by downstream queues to get settings etc. […] Here I didn't understand why we cannot assume argument of function isn't passed?

Oh, I see now.

Usually Scrapy components do not get a crawler directly, but they are created with `create_instance` which calls the class method `from_crawler` if defined, passing the `crawler` object and all other parameters to that class method, which then calls `__init__`. I wrongly assumed that was the case here as well.

In this case, it looks like (for historic reasons?) the `crawler` object is being passed directly on the call to `create_instance`, so there is no need to define `from_crawler` to get the `crawler` object:

https://github.com/scrapy/scrapy/blob/68379197986ae3deb81a545b5fd6920ea3347094/scrapy/pqueues.py#L75-L81

So, please forget the second part of my comment completely.

> Which mark should we use here for proper formatting?

I believe something like this (classmethod → method, cls → self):

```rst
.. method:: __init__(self, crawler: scrapy.crawler.Crawler, key: str, state)
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+
+      Initializes an instance of this disk downstream queue class.
+
+      *key* is the unique ID of the disk downstream queue instance. It may be
+      used, for example, to create a unique file or folder to store the queue
+      content.
+
+      *state* is a variable containing state of queue. It should be serializeable
+      and deserializeble by json module. You need to come up with proper type
+      and structure to completely restore an internal state of your queue.
```
Gallaecio, line 100: Thanks, I think I now understand how this works from just the documentation.

What would you think about a bit of rephrasing?

```suggestion
      *state* is the data returned by the :method:`close` method of this queue 
      during a prior execution, and should be used to restore the queue to that 
      prior state. If there is no prior execution, its value is ``None``.
```

With a reference to `close` here now, I think it makes sense to keep the rest 
of the information in the documentation of the `close` method that is 
responsible for creating the state data.

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+      Raises :exc:`NotImplementedError` if the underlying queue class does
+      not implement a ``peek`` method, which is optional for queues.
+
+      In case of a temporary problem or empty queue, ``None`` is returned.
+
+   .. method:: close(self) -> Optional[Any]
+
+      Release internal resources (e.g. close files or sockets) and return
+      current internal state if any. This state will be seriliazed and passed
+      to __init__ method.
```
Gallaecio, line 147: Here we can include the detail about the data needing to be JSON-serializable, as well as a reference to the topic about pausing and resuming crawls, for better context.

```suggestion
      Release internal resources (e.g. close files or sockets) and return an 
      object serializable by :function:`json.dump` (e.g. a :class:`dict`) that 
      represents the current internal state of this queue. When :ref:`resuming 
      a crawl <topics-jobs>`, this object is passed to the :method:`__init__` 
      method of this queue class to restore the queue state.
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+      used to convert the request into a dict that can then be easily
+      serialized with, for example, :func:`pickle.dumps`.
+
+      Scrapy falls back to the memory downstream queue for *request* if one of
+      the following exceptions is raised:
+
+      -     :exc:`TransientError`: indicates a temporary failure.
+
+            For example, if storing requests on a server, raise this exception
+            if you temporarily loose access to the server.
```
Gallaecio, line 116:
```suggestion
            if you temporarily lose access to the server.
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+
+        -   :exc:`SerializationError`: indicates that *request* could not be
+            serialized
```
Gallaecio, line 119:
```suggestion
            serialized.
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+
+   .. method:: pop(self) -> Optional[Request]
+
+      Pop a request from the queue.
+
+      In case of a temporary problem or empty queue, ``None`` is returned.
+      Exceptions raised in this method woudln't be handled and may lead to
+      halt of the crawler.
```
Gallaecio, line 127:
```suggestion
      a halt of the crawler.
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,160 @@
...
+
+   .. method:: peek(self) -> Optional[Request]
+
+      Returns the next object to be returned by :meth:`pop`,
+      but without removing it from the queue.
+
+      Raises :exc:`NotImplementedError` if the underlying queue class does
+      not implement a ``peek`` method, which is optional for queues.
+
+      In case of a temporary problem or empty queue, ``None`` is returned.
```
Gallaecio, line 141: I’ve just realized that we are lying in the documentation of `ScrapyPriorityQueue.peek` currently in `master`, where we say the same thing about `NotImplementedError`. Our built-in queues just happen to raise `NotImplementedError`, but if a user does not define this method `AttributeError` will be raised instead.

I’m not sure how to best approach this, though. My current thinking is:
- Allow users not to define this method at all, and update that second paragraph above accordingly.
- Have our built-in queues without peek support either not define the method at all or raise `AttributeError` instead of `NotImplementedError`.
- Update the documentation of our public API affected by this change, such as `ScrapyPriorityQueue.peek`.
- This is technically a backward-incompatible change, but I believe the affected classes (directly and indirectly) did not have public documentation before, so maybe it could be argued that raising `NotImplementedError` was an implementation detail?

On a different note, we should use the imperative form here:

```suggestion
      Return the next object to be returned by :meth:`pop`,
      but without removing it from the queue.

      Raise :exc:`NotImplementedError` if the underlying queue class does
      not implement a ``peek`` method, which is optional for queues.

      In case of a temporary problem or empty queue, return ``None``.
```

docs/topics/request-queues.rst
whalebot-helmsman: Sorry, this is very old PR and I relied on a memory. Actually you are right. We need to pass a crawler we need `from_crawler` method here, as https://github.com/scrapy/scrapy/blob/0915be7300b19e9f2df2957734e2710a709bb086/scrapy/utils/misc.py#L171-L175

whalebot-helmsman: `__init__` is replaced by `from_crawler` in the documentation

scrapy/pqueues.py
Gallaecio: With the recent changes about `NotImplementedError` and `AttributeError`, I’ve reviewed this again and I think this docstring needs to stay.

This is not the documentation of a disk downstream queue, which is the context where `docs/topics/request-queues.rst` documents the optional `peek` method. This is a priority queue, and at the moment we don’t seem to be documenting the interface of these like we do for downstream disk queues, we only document the 2 specific implementations of priority queues that Scrapy provides, so we should keep this docstring, both here and in `DownloaderAwarePriorityQueue.peek`.

In the future we may document the interface of priority queues and memory downstream queues, which I assume is out of the scope of these changes. When we do that, we can remove these docstrings, and replace them by a note at the class-level docstring that indicates whether or not peek is supported.

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,166 @@
...
+
+      *downstream_queue_cls* class of queues lying down on queue hierarchy, e.g.
+      priority queue manages disk queues. ``None`` in case there is no downstream
+      queue
+
+      *key* is the unique ID of the disk downstream queue instance. It may be
+      used, for example, to create a unique file or folder to store the queue
+      content.
+
+      *state* is the data returned by the :method:`close` method of this queue
```
Gallaecio, line 102: My bad, it’s:

```suggestion
      *state* is the data returned by the :meth:`close` method of this queue
```

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-meth

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,166 @@
...
+            if you temporarily lose access to the server.
+
+        -   :exc:`SerializationError`: indicates that *request* could not be
+            serialized.
+
+   .. method:: pop(self) -> Optional[Request]
+
+      Pop a request from the queue.
+
+      In case of a temporary problem or empty queue, ``None`` is returned.
```
Gallaecio, line 129: :lipstick: 

```suggestion
      In case of a temporary problem or empty queue, return ``None``.
```

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,166 @@
...
+      be used to convert a deserialized dict back into a
+      :class:`~scrapy.http.Request` object.
+
+   .. method:: peek(self) -> Optional[Request]
+
+      Return the next object to be returned by :meth:`pop`,
+      but without removing it from the queue.
+
+      Raise :exc:`AttributeError` if the underlying queue class does
+      not implement a ``peek`` method, which is optional for queues.
```
Gallaecio, line 143:
```suggestion
      Implementing this method is optional. If implemented, but peek support
      is not possible, raise :exc:`AttributeError`.
```

scrapy/pqueues.py
```diff
@@ -26,17 +26,8 @@ def _path_safe(text):
...
-
-        * push(obj)
-        * pop()
-        * close()
-        * __len__()
-
-    Optionally, the queue could provide a ``peek`` method, that should return the
-    next object to be returned by ``pop``, but without removing it from the queue.
+    """:ref:`Priority queue <priority-queues>` that sends requests based on the
+    :attr:`~scrapy.http.Request` priority value.
```
Gallaecio, line 30: Here we used to define (without much detail) a common interface for downstream queues (both memory downstream queues and disk downstream queues).

Now we link to the new topic, which describes the interface of disk downstream queues, but not of memory downstream queues. So in practice we are removing the documentation of the interface of memory downstream queues.

Maybe the new interface documentation was meant to cover both memory downstream queues and disk downstream queues? I see the interface changes seem to apply to both. If so, we need to rewrite the interface documentation a bit, and make it clear that disk-specific aspects (e.g. serialization and deserialization) only apply to disk queues.

scrapy/pqueues.py
```diff
@@ -63,21 +53,36 @@ def __init__(self, crawler, downstream_queue_cls, key, startprios=()):
...
-    def qfactory(self, key):
+        self.curprio = min(p for p, s in startprios)
+
+    def qfactory(self, key, state):
         return create_instance(
             self.downstream_queue_cls,
             None,
             self.crawler,
             self.key + '/' + str(key),
+            state,
```
Gallaecio, line 85: This backward-incompatible change also needs to be covered in the release notes.

I wonder if it would not be better to deprecate the existing class and create a new one, with a new name (`PriorityQueue`?), with these changes, and let that new class become the default priority queue class.

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Writing your own disk downstream queue class
+--------------------------------------------
+
+If you want to define and use your own disk downstream queue implementation,
+it has to conform to the following interface:
+
+.. class:: MyExternalQueue
+
+   .. classmethod:: from_crawler(cls, downstream_queue_cls: Optional[T], crawler: scrapy.crawler.Crawler, key: str, state)
```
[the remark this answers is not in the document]
whalebot-helmsman: I removed `from_crawler` usage for queues and transition to contructor usage everywhere

scrapy/pqueues.py
whalebot-helmsman: I removed mentions of disk from general context and add in context of serialization. I didn't touch state context, as memory queues with state have sense for me. E.g. you store requests in memory for performance, dump it to a disk at some moment and resume crawl from dumped state later.

docs/topics/request-queues.rst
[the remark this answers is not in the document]
Gallaecio: Why? The general tendency is to try and offer `from_crawler` support in all Scrapy components. And if you are going to pass the crawler anyway, you may as well use `from_crawler` to do it. If nothing else, supporting `from_crawler` simplifies the signature for custom queues that do not need `crawler` access.

scrapy/pqueues.py
Gallaecio: > I didn't touch state context, as memory queues with state have sense for me.

The implementation never passes `state` to memory queues, though. So it would be pointless to implement state handling in a custom downstream memory queue.

docs/topics/request-queues.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Writing your own disk downstream queue class
+--------------------------------------------
+
+If you want to define and use your own downstream queue implementation,
+it has to conform to the following interface:
+
+.. class:: MyExternalQueue
+
+   .. method:: __init__(self, crawler: scrapy.crawler.Crawler, downstream_queue_cls: Optional[T], key: str, state)
```
Gallaecio, line 90: `MyExternalQueue` _is_ `downstream_queue_cls`, right?. I think `downstream_queue_cls` is only passed to priority queues, the parameter should not exist here.

PR #4799, opened 2020-09-20
httpcompression stats added

### LINKED ISSUE #4797: Usage of HttpCompressionMiddleware needs to be reflected in Scrapy stats
## Summary
Usage of `HttpCompressionMiddleware` needs to be relfected in Scrapy stats.
## Motivation
In order to estimate scrapy memory usage efficiency and prevent.. memory leaks like [this](https://stackoverflow.com/q/63936759/10884791).
I will need to know:
1. number of request/response objects that can be active (can be achieved by using [`trackref`](https://docs.scrapy.org/en/latest/topics/leaks.html#debugging-memory-leaks-with-trackref) )
2. size of memory required to store that number of request/response objects. 

A lot of websites use compression to reduce traffic. In this case I would like to calculate average size of **decompressed** responses to estimate p.2.

Decompression process means that at some point application will require to allocate memory to store both compressed and decompressed response body and I will need to know this sizes to have more complete vision of scrapy memory usage.

Also size of  decompressed body will be several times more than size of compressed response and it will affect scrapy memory usage.

The easiest one - is to change priority of `DownloaderStats` middleware and check difference in `downloader/response_bytes` stats parameter.
```
    custom_settings = {"DOWNLOAD_DELAY":1,
                       "DOWNLOADER_MIDDLEWARES":{
                           'scrapy.downloadermiddlewares.stats.DownloaderStats':50
                       }
```
Stats from quotes.toscrape.com spider (it uses `gzip` compression) with default settings:

```
{'downloader/request_bytes': 2642,
 'downloader/request_count': 10,
 'downloader/request_method_count/GET': 10,
 'downloader/response_bytes': 24534,
```
 
And with changed priority of `DownloaderStats` middleware:

```
{'downloader/request_bytes': 912, # size reduced as it didn't count size of request headers populated by downloader middlewares
 'downloader/request_count': 10,
 'downloader/request_method_count/GET': 10,
 'downloader/response_bytes': 110191,   # it counted size of decompressed data 
```

Average size of compressed response (by default) - 2453 bytes.
Average size of decompressed response  - 11019 bytes (~4.5 times more).

## Additional context
Potential solution is to add something like this:
` self.stats.inc_value('decompressed_bytes', spider=spider)`
into `process_response` method of `HttpCompressionMiddleware`

GeorgeA92:
Resolves #4797 
[related code sample](https://gist.github.com/GeorgeA92/2071807bac09b353dbfa38af13a7af44)


Gallaecio:
Some tests are failing. Also, it would be great if you could add a test that covers the change.

GeorgeA92:
@Gallaecio 
> Some tests are failing. Also, it would be great if you could add a test that covers the change.

Code updated. `stats` code and related tests implemented in the same way as it done in [`DownloaderStats`](https://github.com/scrapy/scrapy/blob/2.3.0/scrapy/downloadermiddlewares/stats.py) and in [`TestDownloaderStats`](https://github.com/scrapy/scrapy/blob/2.3.0/tests/test_downloadermiddleware_stats.py)

Gallaecio:
Looks good to me.

I thought about the possibility of making the change backward compatible for subclasses, but assuming subclasses follow best practices (calling `super().__init__(*args, **kwargs)` from `__init__` and `from_crawler` from `from_crawler`), this change seems safe to me.

What about also testing `response_bytes` in the tests, though?

GeorgeA92:
@Gallaecio 

> What about also testing response_bytes in the tests, though?

It is possible. But I am not sure which approach will be correct for this.

`downloader/request_bytes` stats parameter also is not tested in [`TestDownloaderStats`](https://github.com/scrapy/scrapy/blob/2.3.0/tests/test_downloadermiddleware_stats.py)
for [`DownloaderStats`](https://github.com/scrapy/scrapy/blob/2.3.0/scrapy/downloadermiddlewares/stats.py) middleware


For example it can be tested.. by adding lines like:
`self.assertStatsEqual('httpcompression/response_bytes', some_hardcoded_number)`
for each comression method (process_response...) test. Is this will be enough?

Gallaecio:
> For example it can be tested.. by adding lines like:
> `self.assertStatsEqual('httpcompression/response_bytes', some_hardcoded_number)`
> for each comression method (process_response...) test. Is this will be enough?

I think that should be enough.

GeorgeA92:
@Gallaecio 
> I think that should be enough.

Tests for `httpcompression/response_bytes` added.

In case if pull request #4831 will be megred. - probably additional tests required to cover usage of new stats parameters with new `zstd` compression method. As it can't be covered in this pull request.

Gallaecio:
> I thought about the possibility of making the change backward compatible for subclasses, but assuming subclasses follow best practices (calling `super().__init__(*args, **kwargs)` from `__init__` and `from_crawler` from `from_crawler`), this change seems safe to me.

After talking it over with @elacuesta and @wRAR, we think it may be better to go for maximizing backward compatibility. That is:

- Make the new `__init__` parameter optional.
- Modify `from_crawler` so that, if dealing with an `__init__` that does not take parameters, it catches the exception, and instead creates the object without parameters but then sets `object.stats` nonetheless.
- Modify the use of `self.stats` so that nothing breaks if it is `None`.

Of course, it would be great to also cover that with tests. But do let me know if you are short on time, maybe I can do that myself.

GeorgeA92:
Done

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -18,11 +18,14 @@
...
     sent/received from web sites"""
+    def __init__(self, stats=None):
+        self.stats = stats
+
     @classmethod
     def from_crawler(cls, crawler):
         if not crawler.settings.getbool('COMPRESSION_ENABLED'):
             raise NotConfigured
-        return cls()
+        return cls(crawler.stats)
```
Gallaecio, line 28: This is the other change I mentioned, in case of existing subclasses without `from_crawler` overwritten but with an `__init__` method that takes no parameters:

```suggestion
        try:
            return cls(stats=crawler.stats)
        except TypeError:
            result = cls()
            result.stats = crawler.stats
            return result
```

kmike:
Nice, thanks @GeorgeA92 and @Gallaecio!

PR #4804, opened 2020-09-21
Use mocking for S3 tests that currently need server credentials

### LINKED ISSUE #1790: Run all S3 tests on travis

Gallaecio:
Fixes #1790

I used the built-in mocking mechanism from `botocore`, the client library we use.

-   [x] `tests/test_downloader_handlers.py::S3TestCase::test_request_signing5` (removed, only meant to work on Python 2)
-   [x] `tests/test_feedexport.py::S3FeedStorageTest::test_store`
-   [x] `tests/test_feedexport.py::BatchDeliveriesTest::test_s3_export`
-   [x] `tests/test_pipeline_files.py::TestS3FilesStore::test_persist`

### DIFF
--- a/scrapy/utils/test.py
+++ b/scrapy/utils/test.py
@@ -23,18 +23,6 @@ def skip_if_no_boto():
         raise SkipTest('missing botocore library')
 
 
-def get_s3_content_and_delete(bucket, path, with_key=False):
-    """ Get content from s3 key, and delete key afterwards.
-    """
-    import botocore.session
-    session = botocore.session.get_session()
-    client = session.create_client('s3')
-    key = client.get_object(Bucket=bucket, Key=path)
-    content = key['Body'].read()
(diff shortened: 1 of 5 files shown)


wRAR (approved):
Cool!

PR #4806, opened 2020-09-24
pip install scrapy (lowercase)

john-sandall:
pip is case insensitive, specifying install instructions that appear to be case-sensitive can be confusing for new Python users.

> All comparisons of distribution names MUST be case insensitive, and MUST consider hyphens and underscores to be equivalent.

from https://www.python.org/dev/peps/pep-0426/#name

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -27,7 +27,7 @@ To install Scrapy using ``conda``, run::
 Alternatively, if you’re already familiar with installation of Python packages,
 you can install Scrapy and its dependencies from PyPI with::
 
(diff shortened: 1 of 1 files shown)


Gallaecio (commented):
I’m +0 on this.

It looks cleaner to me in lowercase.

However, [the current code is what PyPI displays](https://pypi.org/project/Scrapy/), so I think this change could cause more confusion than it may solve.

john-sandall:
Fair enough, just realised Django is in exactly the same boat, will close https://docs.djangoproject.com/en/3.1/topics/install/#installing-official-release

PR #4807, opened 2020-09-25
Custom Amazon S3 headers in FileStorage (scrapy#4788)

soumallyadev07:


Gallaecio:
Thanks!

What would you think about making this part of the media pipelines documentation page?

I think we should eventually have a separate page on storages, as they are not only used by media pipelines. But writing that would be a different beast, and out of the scope of the specific piece of information that you aim to cover in the document here.

soumallyadev07:
@Gallaecio Yeah I think that would be a better idea and since we do not have a dedicated page as of now, we should be making this part of the media pipelines documentation page.

Could you tell me a bit about one of the checks that were failed - **continuous-integration/travis-ci/pr — The Travis CI build failed**

Gallaecio:
The CI issue was caused by the new page not being included in a Sphinx table of contents; but it does not matter if you will instead add a section to an existing page.

soumallyadev07:
> The CI issue was caused by the new page not being included in a Sphinx table of contents; but it does not matter if you will instead add a section to an existing page.

Thanks @Gallaecio 
I've raised another pull request with the necessary changes and additions, do check it out.

Gallaecio:
Closing in favor of #4813

PR #4811, opened 2020-09-29
Remove todo

yifeiwu:


Gallaecio:
Removing that code also requires some major rewrites of the affected tests. I’m guessing that’s the reason it was not just removed as well from the original author of that comment.

Issue #1773, opened 2016-02-09
Unable to run scrapy on Python3.5.1


adeelehsan:
I have python3.5.1 on my system and i have also creates the virtual enviorment with python 3.5.1. I have installed scrapy 1.0.5 and when i tried to run the command "Scrapy version"  or scrapy startproject , I am getting following error:
Traceback (most recent call last):
  File "/home/adeel/Intermix_py3.5/venv/bin/scrapy", line 11, in <module>
    sys.exit(execute())
  File "/home/adeel/Intermix_py3.5/venv/lib/python3.5/site-packages/scrapy/cmdline.py", line 122, in execute
    cmds = _get_commands_dict(settings, inproject)
  File "/home/adeel/Intermix_py3.5/venv/lib/python3.5/site-packages/scrapy/cmdline.py", line 46, in _get_commands_dict
    cmds = _get_commands_from_module('scrapy.commands', inproject)
  File "/home/adeel/Intermix_py3.5/venv/lib/python3.5/site-packages/scrapy/cmdline.py", line 29, in _get_commands_from_module
    for cmd in _iter_command_classes(module):
  File "/home/adeel/Intermix_py3.5/venv/lib/python3.5/site-packages/scrapy/cmdline.py", line 21, in _iter_command_classes
    for obj in vars(module).itervalues():
AttributeError: 'dict' object has no attribute 'itervalues'

redapple:
This `.itervalues()` call for the command line tool has been [changed here](https://github.com/scrapy/scrapy/commit/3e6d6c43ac0763adf2cd92efdb4a1dc2ba165440)

@adeelehsan , what version of scrapy have you installed?
Scrapy 1.0.x is not Python 3 compatible.
Either use the current master branch or install 1.1.0rc1
`$ pip install scrapy==1.1.0rc1`

micmalti:
@redapple I've tried both of your suggestions but didn't work.

```
File "/home/micmalti/Downloads/pypy3-v5.8.0-linux64/site-packages/scrapy/http/request/form.py", line 11, in <module>
    import lxml.html
  File "/home/micmalti/Downloads/pypy3-v5.8.0-linux64/site-packages/lxml/html/__init__.py", line 54, in <module>
    from .. import etree
  File "src/lxml/xmlschema.pxi", line 22, in init lxml.etree (src/lxml/lxml.etree.c:231895)
[...]

PR #4814, opened 2020-09-30
add pip 20.2 test for tox

### LINKED ISSUE #4710: Test the upcoming pip dependency resolver in CI
From pypi-announce@python.org, about pip 20.2:

> NOTICE: This release includes the beta of the next-generation dependency resolver. It is significantly stricter and more consistent when it receives incompatible instructions, and reduces support for certain kinds of constraints files, so some workarounds and workflows may break. Please test it with the `--use-feature=2020-resolver` flag. Please see our guide on how to test and migrate, and how to report issues
<https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-2-2020>.
> 
> The new dependency resolver is *off by default* because it is *not yet
ready for everyday use*.

dswij:
Closes #4710 

Add install command for pip-20.2 to test for upcoming pip resolver

TODO:
- [x] Fix PyDispatcher dependency
- [x] Mark the test failure as expected
- [x] Minimize dependency version changes


wRAR:
"pyopenssl 18.0.0 depends on cryptography>=2.2.1"

dswij:
Seems like some tests are failing on pypy.

Gallaecio:
Weird errors. Let’s re-run the tests, see if they are reproducible.

tox.ini
```diff
@@ -62,13 +66,15 @@ commands =
...
 deps =
     -ctests/constraints.txt
-    cryptography==2.0
+    #using cryptography-2.3.1 to solve dependencies conflict on pip20.2>=
+    cryptography==2.3.1
     cssselect==0.9.1
     itemadapter==0.1.0
     parsel==1.5.0
     Protego==0.1.15
     PyDispatcher==2.0.5
```
Gallaecio, line 75: Looking at https://github.com/scrapy/scrapy/issues/1773#issuecomment-327620131 , it looks like we may need to upgrade `PyDispatcher` to `>=2.1.0` on PyPy.

So, assuming that version upgrade fixes the issue in the first of the 2 PyPy jobs, I would suggest:

1. Updating `setup.py` to require `PyDispatcher>=2.1.0` when using PyPy.
2. In `tox.ini`, creating a `[testenv:pypy-pinned]` section similar to the Windows one, and following a similar approach to install `PyDispatcher==2.1.0` only in the `pypy-pinned` environment.
3. In `travis.yml`, switch the first PyPy job from `pinned` to `pypy-pinned`.

dswij: I think [#1773 comment](https://github.com/scrapy/scrapy/issues/1773#issuecomment-327620131) mentions `PyPyDispatcher >= 2.1.0` instead of `PyDispatcher`. (which is in pypy-pinned deps and setup.py)

dswij:
So after having time to look at it further, there are some quirks on using the new pip resolver that failed some of the test:

1.  `pypy-pinned` env test failed because of an explicit `PyPyDispatcher == 2.1.0` deps
2. `pypy` env test failed because `PyDispatcher==2.0.5` is not explicitly declared
3. `asyncio-pinned` test failed because of a timeout. Installing a new version of Twisted fixed this by installing additional packages (async-timeout , ...). Not sure why the old resolver passed this test.

Although running `pip list` before `pytest` shows the exact same packages installed. Unsure on what is causing this, but I think it'll be better having two separate test for the new and old resolver for now. (since pip themselves suggest using it on a non-production environment only)

> The more feedback we can get, the more we can make sure that the final release is solid. (Only try the new resolver in a non-production environment, though - it isn’t ready for you to rely on in production!)

from [here](https://pip.pypa.io/en/stable/user_guide/#how-to-test)

Gallaecio:
I think we need to figure out why the PyPyDispatcher thing is not working.

From the PyPyDispatcher README, I see it’s meant as a standalone replacement for PyDispatcher with PyPy support, so the first thing I would try is to prevent `PyDispatcher` from being installed on PyPy; I suspect the issue may be that it gets imported instead of the PyPy replacement.

I think we should edit the setup.py file to require `PyDispatcher` outside PyPy and `PyPyDispatcher` under PyPy.

.travis.yml
```diff
@@ -48,6 +48,7 @@ install:
         tar -jxf ${PYPY_VERSION}.tar.bz2
         virtualenv --python="$PYPY_VERSION/bin/pypy3" "$HOME/virtualenvs/$PYPY_VERSION"
         source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+        $HOME/virtualenvs/$PYPY_VERSION/bin/pypy -m ensurepip --default-pip
```
Gallaecio, line 51: Is this change required?

tox.ini
```diff
@@ -62,18 +66,19 @@ commands =
...
-    pyOpenSSL==16.2.0
+    #using pyOpenSSL<18.1 to solve dependencies conflict with mitmproxy on pip20.2>=
+    pyOpenSSL==18.0.0
     queuelib==1.4.2
     service_identity==16.0.0
-    Twisted==17.9.0
+    Twisted==20.3.0
     w3lib==1.17.0
-    zope.interface==4.1.3
+    zope.interface==5.1.2
```
Gallaecio, line 81: Why is this needed?

tox.ini
```diff
@@ -83,7 +88,12 @@ deps =
 [testenv:pinned]
 deps =
     {[pinned]deps}
-    lxml==3.5.0
+    PyDispatcher==2.0.5
+
+[testenv: pypy-pinned]
+deps = 
+    {[pinned]deps}
+    PyPyDispatcher==2.1.0
```
Gallaecio, line 96: This testenv should be removed, and instead `pypy3-pinned` modified accordingly (or rather left as it was).

tox.ini
```diff
@@ -83,7 +88,12 @@ deps =
 [testenv:pinned]
 deps =
     {[pinned]deps}
-    lxml==3.5.0
```
Gallaecio, line 86: Why was this removed?

tox.ini
```diff
@@ -105,10 +115,16 @@ commands =
 
 [testenv:asyncio-pinned]
 commands = {[testenv:asyncio]commands}
-deps = {[testenv:pinned]deps}
+deps = 
+    {[testenv:pinned]deps}
+    lxml==3.5.0
```
Gallaecio, line 120: After reverting the removal of `lxml==3.5.0` from `testenv:pinned`, this change should be reverted.

Gallaecio:
As for the asyncio job getting stuck with `Twisted==17.9.0`, I’ve found that the offending test is `tests/test_downloadermiddleware.py::MiddlewareUsingCoro::test_asyncdef`. @wRAR Do you think you can have a look? Maybe we need to document a higher minimum version of Twisted when using asyncio? Or there’s some Twisted bug that we may be able to workaround?

tox.ini
dswij: To resolve dependencies conflict with Twisted. Probably should've left a comment there

tox.ini
dswij: `testenv:pinned` is only used by one job, and only add one dependencies. Wouldn't it make sense to remove it?

tox.ini
dswij: I was thinking of using the `testenv:pinned` part for replacing the `PyDispatcher` requirements for pinned environments other than pypy, (for `windows-pinned` and `asyncio-pinned`). I just remembered though that there is a test job in travis on this testenv. Let me try some other alternatives than making a mess.

wRAR:
I was able to reproduce the difference in the test behavior locally, I will look further into it.

wRAR:
Not sure if it's already known, but with the old resolver the newest Twisted gets installed even in the pinned env :) See e.g. https://travis-ci.org/github/scrapy/scrapy/jobs/734210368

wRAR:
The actual problem is fixed in 18.4.0 and I suspect it's because of http://twistedmatrix.com/trac/ticket/9390 (which is probably specific to tests?)

dswij:
> 
> 
> Not sure if it's already known, but with the old resolver the newest Twisted gets installed even in the pinned env :) See e.g. https://travis-ci.org/github/scrapy/scrapy/jobs/734210368

It appears that only Twisted has version mismatch in declared deps and installed version.  I suspect that the new resolver might install older version only because newer versions of Twisted has conflict with current zope.interface versions and resorted to the specified deps since it does not have strict range.

dswij:
> 
> 
> > Not sure if it's already known, but with the old resolver the newest Twisted gets installed even in the pinned env :) See e.g. https://travis-ci.org/github/scrapy/scrapy/jobs/734210368
> 
> It appears that only Twisted has version mismatch in declared deps and installed version. I suspect that the new resolver might install older version only because newer versions of Twisted has conflict with pinned `zope.interface` versions. The new resolver resorted to the specified version of Twisted in deps to resolve the conflict.

Turns out using an old version of Twisted and newer version of zope.interface still enforces pinned version of Twisted in new resolver. So this is not the case.

wRAR:
I've created a separate issue for the test problem.

setup.py
```diff
@@ -24,7 +24,6 @@ def has_environment_marker_platform_impl_support():
     'cssselect>=0.9.1',
     'itemloaders>=1.0.1',
     'parsel>=1.5.0',
-    'PyDispatcher>=2.0.5',
```
kmike, line 27: Does it mean that if `has_environment_marker_platform_impl_support()` rerturns False, than PyDispatcher is not installed?

Gallaecio: Oops, right, we also need to add `'PyDispatcher>=2.0.5'` in the `else` below, nice catch! :sweat_smile:

dswij: Nice catch, and great suggestion. I am currently unable to do this though.

wRAR:
It was decided that we mark the test that fails with older twisted with xfail, revert the twisted version bump and the version bumps in the pinned env that were caused by it, and then review the remaining version bumps.

Gallaecio:
For some reason the Python 3.6 jobs, with pinned dependencies, hang during dependency installation, both in Travis and Azure. However, this Dockerfile does not hang:

```Dockerfile
FROM python:3.6.1

RUN pip install tox

COPY . /app/
WORKDIR /app

CMD tox -e pinned
```

Gallaecio:
OK, the hang is https://github.com/pypa/pip/issues/8713

dswij:
> 
> 
> OK, the hang is [pypa/pip#8713](https://github.com/pypa/pip/issues/8713)

I remember the reason for the fail pinned deps is that the initial pinned deps of  `cryptography`, `PyOpenSSL`, and `mitmproxy` is conflicting (according to _20.3 version of resolver_). Although before, it will fail and exit automatically instead of hangs.

I have tried using the latest commit and changing back to non-conflicting `cryptograpy` and `PyOpenSSL` version and it works fine.

Gallaecio:
Thanks. Yes, upgrading those 2 works.

However, I think we might want to consider the opposite approach: skip tests required mitmproxy in pinned environments instead. Because the point of pinned environments is that everything works with the minimum version, I think that testing that most tests pass with all the minimum versions is more valuable than testing that all tests pass with most of the minimum versions.

wRAR:
If installing mitmproxy leads to upgrading our deps, agreed.

tox.ini
```diff
@@ -21,8 +25,12 @@ passenv =
     AWS_SECRET_ACCESS_KEY
     GCS_TEST_FILE_URI
     GCS_PROJECT_ID
+#allow tox virtualenv to upgrade pip/wheel/setuptools
+download = true 
 commands =
     py.test --cov=scrapy --cov-report= {posargs:--durations=10 docs scrapy tests}
+install_command =
+    pip install --use-feature=2020-resolver {opts} {packages}
```
kmike, line 33: Are we testing only with the new resolver now? 
Can you see any cases where new resolver works, but the old one doesn't?

Gallaecio: I don’t think we can discard that scenario. I think it’s possible that the new resolver will fix issues that the old one does not.

However, I’m not sure which jobs should get the new resolver. My original plan was to just set it on 1 job, e.g. extra-deps (or a new job also using the development version of Python, but that would probably be too unstable). 

On the other hand, I think pinned environments are the ones where a resolver may have a higher impact, so I wonder if instead we should use the new resolver in half the pinned environments. Also, given https://github.com/scrapy/scrapy/pull/4814#issuecomment-706230011, I think it would also be fair to say that we must use the new resolver in *all* pinned environments.

kmike: I'm not sure what's the best option @Gallaecio, but it looks like running new resolver only on some jobs (not on all of them) would be better. Probably it is not worth it to have a separate env just to test env a different resolver; I'm worried about CI runtime creep.

Gallaecio: I’m going to restrict it to extra-deps for now. As soon as the tests finish with that change, I think we should check the pinned-dependencies jobs for [cases where a newer Twisted is actually installed](https://github.com/scrapy/scrapy/pull/4814#issuecomment-706230011), and also switch those to the new resolver.

Gallaecio: Done.

elacuesta (approved):

PR #4817, opened 2020-10-01
Tests: use classes instead of paths in settings

elacuesta:
Simplify tests after #3873

### DIFF
--- a/tests/test_spidermiddleware_output_chain.py
+++ b/tests/test_spidermiddleware_output_chain.py
@@ -1,3 +1,5 @@
+from copy import copy
+
 from testfixtures import LogCapture
 from twisted.internet import defer
 from twisted.trial.unittest import TestCase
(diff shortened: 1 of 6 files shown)


tests/test_spidermiddleware_output_chain.py
```diff
@@ -56,9 +58,9 @@ class ProcessSpiderInputSpiderWithoutErrback(Spider):
     custom_settings = {
         'SPIDER_MIDDLEWARES': {
             # spider
-            __name__ + '.LogExceptionMiddleware': 10,
-            __name__ + '.FailProcessSpiderInputMiddleware': 8,
-            __name__ + '.LogExceptionMiddleware': 6,
```
wRAR, line 61: So it looks like it currently uses only one of these entries? :)

wRAR (approved):
I like this so much more!

tests/test_spidermiddleware_output_chain.py
elacuesta: Indeed, the behaviour makes sense. The failing middleware fails in `process_spider_input` and does not define a `process_spider_exception` method, so `LogExceptionMiddleware.process_spider_exception` (the method that logs the exception) is placed at the beginning of the chain as expected. In fact, we only need one of the occurrences for the test case, I'm removing the duplicate one.
Nice catch, thanks for noticing!

elacuesta: (For reference)
current `master` (5a386393):
```
  Enabled spider middlewares:
['tests.test_spidermiddleware_output_chain.LogExceptionMiddleware',
 'tests.test_spidermiddleware_output_chain.FailProcessSpiderInputMiddleware',
 'scrapy.spidermiddlewares.httperror.HttpErrorMiddleware',
 ```

current `test-classes-settings` (62394c2b):
```
   Enabled spider middlewares:
[<class 'tests.test_spidermiddleware_output_chain.LogExceptionMiddleware'>,
 <class 'tests.test_spidermiddleware_output_chain.FailProcessSpiderInputMiddleware'>,
 'scrapy.spidermiddlewares.httperror.HttpErrorMiddleware',

Gallaecio (approved):

PR #4818, opened 2020-10-01
Crawl rule: remove deprecated code

elacuesta:
Deprecated in #3682, which was released in [1.7](https://github.com/scrapy/scrapy/releases/tag/1.7.0) (Jul 18, 2019).

Personally I don't see a problem with removing the `Rule.process_request_argcount` attribute (which should have been underscored from the beginning, my bad :raising_hand:) without deprecating it first, because it's not documented, but I understand if any objections are made. What are your thoughts on this?

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -6,14 +6,11 @@ See documentation in docs/topics/spiders.rst
 """
 
 import copy
-import warnings
 from typing import Sequence
 
-from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.http import Request, HtmlResponse
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import Spider
-from scrapy.utils.python import get_func_args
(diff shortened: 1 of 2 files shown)


scrapy/spiders/crawl.py
```diff
@@ -37,38 +34,29 @@ def _get_method(method, spider):
...
         self.process_request = process_request or _identity_process_request
-        self.process_request_argcount = None
         self.follow = follow if follow is not None else not callback
 
     def _compile(self, spider):
         self.callback = _get_method(self.callback, spider)
         self.errback = _get_method(self.errback, spider)
         self.process_links = _get_method(self.process_links, spider)
         self.process_request = _get_method(self.process_request, spider)
-        self.process_request_argcount = len(get_func_args(self.process_request))
```
Gallaecio, line 56: I’m +0.5 on just removing it.

Gallaecio (approved):

PR #4828, opened 2020-10-04
Add extension to check settings

### LINKED ISSUE #4302: Allow checking that settings are named correctly
I would like a way to be able to detect typos on spider settings.

Some approaches that I can think of:
- Making this a core feature of Scrapy, having extensions report the settings they use somehow and having an opt-in setting to enable this kind of check.
- Writing a static analyzer for Scrapy projects that detects such things. It could be an extension for Pylint.

joaquingx:
Fixes https://github.com/scrapy/scrapy/issues/4302 

Example Script: 
```
import scrapy


    name = "quotes"
    custom_settings = {"DOWNLOAD_DELAY": 1,
                       "COOKIES_DISABLED": True,  # mistyped, should be enabled
                       "CONCURRENCY": 5,
                       "EXTENSIONS": {
                           "scrapy.extensions.checksettings.CheckSettings": 0,
                       },
                       "CHECK_SETTINGS_ENABLED": True,  # var to enable checksettings
                       "CHECK_SETTINGS_ENABLEDI": True, # mistyped
                       }

    def start_requests(self):

    def parse(self, response):
            yield {
                "quote": quote.css("span.text::text").extract(),
                "author": quote.css("small.author::text").extract(),
                "tags": quote.css("a.tag::text").extract()
            }
            break
        next = response.css("li.next a::attr(href)").extract_first()
        if next:
            yield scrapy.Request(url=response.urljoin(next), callback=self.parse)


process = CrawlerProcess()
process.crawl(QuotesToScrapeSpider)
process.start()
```
Logs added by extension will be:
```
...
2020-10-06 23:13:10 [scrapy.extensions.checksettings] WARNING: Not used settings: 
['COOKIES_DISABLED', 'CONCURRENCY', 'CHECK_SETTINGS_ENABLEDI']
2020-10-06 23:13:10 [scrapy.extensions.checksettings] INFO: Settings suggestions: 
['COOKIES_DISABLED, did you mean COOKIES_ENABLED ?',
 'CHECK_SETTINGS_ENABLEDI, did you mean CHECK_SETTINGS_ENABLED ?']
...
```


docs/topics/settings.rst
```diff
@@ -275,6 +275,27 @@ project name). This name will be used for the logging too.
...
+Default: ``True``
+
+Whether to enable Check Settings.
+
+.. setting:: CHECK_SETTINGS_MIN
+
+CHECK_SETTINGS_MIN
+------------------
+
+Default: ``70``
```
Gallaecio, line 292: I wonder if this is worth parameterizing. No strong opinion either way, though.

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
+"""
+Check settings extension
+
+See documentation in docs/topics/extensions.rst
+"""
+import logging
+import pprint
+import operator
```
Gallaecio, line 8: :lipstick: 

```suggestion
import operator
import pprint
```

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
...
+
+from fuzzywuzzy.fuzz import ratio
+from scrapy import signals
```
Gallaecio, line 11: :lipstick: 

```suggestion
from fuzzywuzzy.fuzz import ratio

from scrapy import signals
```

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
+"""
+Check settings extension
+
+See documentation in docs/topics/extensions.rst
+"""
+import logging
+import pprint
+import operator
+
+from fuzzywuzzy.fuzz import ratio
```
Gallaecio, line 10: I wonder if it would not be better to just indicate that a setting is unused, and not suggest possible matches, to avoid adding a new dependency.

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
...
+from scrapy.exceptions import NotConfigured
+
+logger = logging.getLogger(__name__)
+
+
+class CheckSettings:
+    def __init__(self, crawler):
+        self.settings = crawler.settings
+        self.check_settings_min = crawler.settings.getint("CHECK_SETTINGS_MIN")
+        self.not_used_settings = [s for s in self.settings
```
Gallaecio, line 21: Wouldn’t it be better to delay this as much as possible (i.e. on spider close)? That way we minimize the chance of false positives.

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
...
+from scrapy import signals
+from scrapy.exceptions import NotConfigured
+
+logger = logging.getLogger(__name__)
+
+
+class CheckSettings:
+    def __init__(self, crawler):
+        self.settings = crawler.settings
+        self.check_settings_min = crawler.settings.getint("CHECK_SETTINGS_MIN")
```
Gallaecio, line 20: On the topic of settings for this extensions, I wonder if it would not make sense to provide a setting do indicate a list of settings that can be ignored by this extension, so that they are considered used even if they are not.

scrapy/settings/__init__.py
```diff
@@ -37,20 +37,34 @@ class SettingsAttribute:
...
 
     def __init__(self, value, priority):
-        self.value = value
+        self._value = value
         if isinstance(self.value, BaseSettings):
-            self.priority = max(self.value.maxpriority(), priority)
+            self.priority = max(self._value.maxpriority(), priority)
         else:
             self.priority = priority
+        self.hit = False
```
Gallaecio, line 45: I wonder if there’s a better word for this, as to me `hit` could be both for setting and getting a setting value. `read` could be mistaken by a verb. `has_been_read`?

setup.py
```diff
@@ -32,6 +32,7 @@ def has_environment_marker_platform_impl_support():
     'zope.interface>=4.1.3',
     'protego>=0.1.15',
     'itemadapter>=0.1.0',
+    'fuzzywuzzy==0.18.0',
```
Gallaecio, line 35: _If_ we add this dependency, we should not hardcode a specific version. Unlike lock files, `setup.py` is meant to indicate  a range of versions known to work.

scrapy/extensions/checksettings.py
joaquingx: You're right, will change it.

joaquingx: I think that probably in cases of spider with thousands of links to follow the user will like to have the list of not used settings in the beginning. What do you think?

setup.py
joaquingx: I deleted that dependency in favor to use a builtin one(`difflib`)

docs/topics/settings.rst
joaquingx: I wasnt sure about this, will delete it.

scrapy/extensions/checksettings.py
joaquingx: Changed to difflib.

scrapy/extensions/checksettings.py
Gallaecio: You have a point. But I’m worried that settings used later will often get misreported, which could be confusing. I’m unsure; go with your gut, and once this goes out of WIP let’s see what other people think.

Gallaecio:
About using `difflib`, it’s still a dependency.

If it is an indirect dependency of Scrapy, then I think it’s definitely OK to use, but it should be added to `setup.py` nonetheless, so that if it ever stops being an indirect dependency, Scrapy does not stop working (i.e. the dependency should still be installed if the Scrapy dependency that requires it stops requiring it in a future version).

However, I think it may be available during tests simply because it’s probably a dependency of `pytest`, which would mean that the extension would not work in a regular installation of Scrapy. If this is the case, we definitely need to add it to `setup.py`, and whether or not adding the dependency is worth it will be up for debate. On the bright side, it’s definitely lighter than the previous one.

joaquingx:
> About using `difflib`, it’s still a dependency.
> 
> If it is an indirect dependency of Scrapy, then I think it’s definitely OK to use, but it should be added to `setup.py` nonetheless, so that if it ever stops being an indirect dependency, Scrapy does not stop working (i.e. the dependency should still be installed if the Scrapy dependency that requires it stops requiring it in a future version).
> 
> However, I think it may be available during tests simply because it’s probably a dependency of `pytest`, which would mean that the extension would not work in a regular installation of Scrapy. If this is the case, we definitely need to add it to `setup.py`, and whether or not adding the dependency is worth it will be up for debate. On the bright side, it’s definitely lighter than the previous one.

Hey :smile:, I didn't think about that :see_no_evil: , I agree, we should add the dependency. I'm just not sure how to do add that kind of dependency, I was searching how to do it, because seems that just adding `difflib` to https://github.com/scrapy/scrapy/blob/master/setup.py#L21 isn't right.

Gallaecio:
> seems that just adding `difflib` to https://github.com/scrapy/scrapy/blob/master/setup.py#L21 isn't right.

That’s what I would do. Maybe with `>=<version>` if we can identify the minimum version that will work, based on the code we use and the version of `difflib` where it was introduced.

Specific to Scrapy, you should also set that minimum version in pinned test requirements, in `tox.ini`. But I think that’s it.

joaquingx:
Hey @Gallaecio, when I added `difflib` to `setup.py` and tried to install it I got:
```
ERROR: Could not find a version that satisfies the requirement difflib>=2.4.0 (from Scrapy==2.3.0) (from versions: none)
```
As `difflib` is python standard library(https://docs.python.org/3/library/difflib.html), maybe is not necessary to include it?

Gallaecio:
:open_mouth: 

So, when I said:

> > About using `difflib`, it’s still a dependency.

I was *completely* wrong. I had no idea it part of the standard library. And I think I used it years ago :sweat_smile: .

In that case, I’m all for `difflib`!

docs/topics/settings.rst
```diff
@@ -275,6 +275,15 @@ project name). This name will be used for the logging too.
...
 project with the :command:`startproject` command.
 
+.. setting:: CHECK_SETTINGS_ENABLED
+
+CHECK_SETTINGS_ENABLED
+----------------------
+
+Default: ``True``
+
+Whether to enable Check Settings.
```
Gallaecio, line 285: We should either link back to the CheckSettings extension documentation here, or remove this altogether and add `.. setting:: CHECK_SETTINGS_ENABLED`` before the CheckSettings extension documentation header instead.

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
+"""
+Check settings extension
+
+See documentation in docs/topics/extensions.rst
+"""
+import logging
+import pprint
+
+from difflib import get_close_matches
```
Gallaecio, line 9: :lipstick: 

```suggestion
import logging
import pprint
from difflib import get_close_matches
```

scrapy/extensions/checksettings.py
```diff
@@ -0,0 +1,54 @@
...
+    def from_crawler(cls, crawler):
+        if not crawler.settings.getbool("CHECK_SETTINGS_ENABLED"):
+            raise NotConfigured
+        ext = cls(crawler)
+        crawler.signals.connect(ext.spider_opened, signal=signals.spider_opened)
+        return ext
+
+    def spider_opened(self):
+        self.not_used_settings = [s for s in self.settings
+                                  if not self.settings.attributes[s].has_been_read]
```
Gallaecio, line 44: Since you are going with checking earlier, I think https://github.com/scrapy/scrapy/pull/4828#discussion_r499829754 applies even more.

In fact, it might be good to review all built-in Scrapy settings that are read at `spider_opened` or later, and include those by default into the list, maybe by providing a counterpart `_BASE` setting (see `Setting.getwithbase`).

scrapy/extensions/checksettings.py
joaquingx: done

scrapy/extensions/checksettings.py
joaquingx: I added `CHECK_SETTINGS_IGNORED` setting variable. I'm not sure if a `_BASE` setting will be useful here, because if we want to check all built-in scrapy settings, the `_BASE` and the not `_BASE` setting will be the same(`[]`), please correct me if im wrong.

docs/topics/settings.rst
joaquingx: I added `Scope` to settings and a new section for `CHECK_SETTINGS_IGNORED`.

scrapy/extensions/checksettings.py
Gallaecio: Still, for the default value we should look for built-in settings that will likely be reported due to them being read after the check, if there is any.

docs/topics/extensions.rst
```diff
@@ -369,3 +369,13 @@ For more info see `Debugging in Python`_.
 This extension only works on POSIX-compliant platforms (i.e. not Windows).
 
 .. _Debugging in Python: https://pythonconquerstheuniverse.wordpress.com/2009/09/10/debugging-in-python/
+
+CheckSettings extension
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CheckSettings
```
Gallaecio, line 376: The reason why other classes above are not fully qualified is that they are preceded by a `.. module::` declaration, which indicates that anything below, until the next `.. module::` declaration, belongs to the module.

As a result, this currently references `scrapy.extensions.debug.CheckSettings`.

You can either use the fully-qualified name, or add a `.. module::` declaration above in line with similar entries.

docs/topics/settings.rst
```diff
@@ -275,6 +275,29 @@ project name). This name will be used for the logging too.
...
+.. setting:: CHECK_SETTINGS_ENABLED
+
+CHECK_SETTINGS_ENABLED
+----------------------
+
+Default: ``True``
+
+Scope: ``scrapy.extensions.checksettings.CheckSettings``
+
+Whether to enable Check Settings. Optionally a list of variables to ignore
```
Gallaecio, line 287: This is what I had in mind:

```suggestion
Whether to enable the :class:`~scrapy.extensions.checksettings.CheckSettings` extension. Optionally a list of variables to ignore
```

joaquingx: Updated :+1:

joaquingx:
Is this error from CI: https://github.com/scrapy/scrapy/pull/4828/checks?check_run_id=1573211823 or something is bad in code?

eLRuLL:
> Is this error from CI: https://github.com/scrapy/scrapy/pull/4828/checks?check_run_id=1573211823 or something is bad in code? @Gallaecio

@joaquingx from what I see several things went wrong there, I even saw some logs indicating it's using python2 instead of 3 :thinking: please reupload your code so it gets triggered again, could also be something temporary as it wasn't able to install several packages

docs/topics/settings.rst
```diff
@@ -275,6 +275,25 @@ project name). This name will be used for the logging too.
...
+extension. Optionally a list of variables to ignore.
+
+.. setting:: CHECK_SETTINGS_IGNORED
+
+CHECK_SETTINGS_IGNORED
+----------------------
+
+Default: ``[]``
+
+List of settings variables to ignore by Check Settings extension.
```
Gallaecio, line 295:
```suggestion
List of settings variables to ignore by :class:`~scrapy.extensions.checksettings.CheckSettings` extension.
```

Gallaecio:
Closing and reopening to trigger a new build…

Gallaecio:
CI errors should be eventually fixed by https://github.com/scrapy/scrapy/pull/4814

eLRuLL:
I guess to not make too much noise this should wait for #4814 to be merged, then @joaquingx you can rebase after

docs/topics/settings.rst
Gallaecio, line 295: Since people usually read settings documentation separately, as opposed to reading the whole documentation, I would also use ``` :class:`~scrapy.extensions.checksettings.CheckSettings` ``` here.

Gallaecio:
I’m fine (+0) with merging (besides minor https://github.com/scrapy/scrapy/pull/4828#discussion_r570239879), but I see a potential issue of false positives that I would like to address by either moving the settings check to the end of the spider or allow-listing all known settings by default.

joaquingx:
Ok, so the next step is to wait for more reviews right :)? @Gallaecio

AdrianAtZyte:
https://scrapy-lint.readthedocs.io/en/latest/rules/scp27.html

PR #4831, opened 2020-10-05
Adding support for zstd in HttpCompressionMiddleware

starrify:


docs/topics/downloader-middleware.rst
```diff
@@ -684,11 +684,14 @@ HttpCompressionMiddleware
    This middleware allows compressed (gzip, deflate) traffic to be
    sent/received from web sites.
 
-   This middleware also supports decoding `brotli-compressed`_ responses,
-   provided `brotlipy`_ is installed.
+   This middleware also supports decoding `brotli-compressed`_ as well as
+   `zstd-compressed`_ responses, provided that `brotlipy`_ or `zstd`_ is
```
Gallaecio, line 688: :lipstick: 

```suggestion
   `zstd-compressed`_ responses, provided that `brotlipy`_ or `zstd`_ are
```

Gallaecio (commented):
What about tests? I see there are brotli tests, so it should not be too hard to add tests for this as well.

starrify:
Hi @Gallaecio, thanks a lot for pointing that out! I didn't see a test for brotli at a quick glance, thus decided to submit the PR for review before adding tests for both.

> I didn't see a test for brotli at a quick glance

However it turns out this was actually caused by a rather stupid mistake of mine as what I did was to check only the main source directory: :facepalm: 

```
$ find scrapy/ -type f -name "*.py" | parallel grep -H brotli
```

I have now updated this PR with further enhancements and the corresponding test. :)

wRAR:
Interesting, do you know which websites use that?

starrify:
Hi @wRAR that's surely a very good question :sweat_smile: and I'm unfortunately unable to name many. (nearly unused like brotli? :thinking:)
I'm having it added majorly because I recently noticed that zstd is now included in [IANA's registry](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml).

By the way here's a list of "Zstandard is used by" tools / projects / websites / etc.: https://facebook.github.io/zstd/#:~:text=Zstandard%20is%20used%20by

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -1,4 +1,5 @@
 import zlib
+import io
```
Gallaecio, line 2: :lipstick: 

```suggestion
import io
import zlib
```

Gallaecio (approved):
Nice!

PS: The incomplete coverage is due to test always having the optional dependency installed, same as with brotli.

wRAR (approved):

wRAR:
Thanks!

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -67,4 +74,9 @@ def _decode(self, body, encoding):
                 body = zlib.decompress(body, -15)
         if encoding == b'br' and b'br' in ACCEPTED_ENCODINGS:
             body = brotli.decompress(body)
+        if encoding == b'zstd' and b'zstd' in ACCEPTED_ENCODINGS:
+            # Using its streaming API since its simple API could handle only cases
+            # where there is content size data embedded in the frame
+            reader = zstandard.ZstdDecompressor().stream_reader(io.BytesIO(body))
```
kmike, line 80: hey! I wonder if it makes sense to use a context manager, to close the reader after it is used. This is very minor, as 
it is not doing much (https://github.com/indygreg/python-zstandard/blob/53b71dc3f96961564c9c140bf88b0aa118589247/zstandard/cffi.py#L1937), but it still may clean up some references earlier.

PR #4838, opened 2020-10-08
Create README-Korean.md

dohan01:
Scrapy is a very good framework, but it is difficult for Koreans to use.
One of the main reasons is that there is no manual in Korean.
To solve this problem, this PR provides a translation.


Gallaecio:
As far as the documentation goes, there’s https://github.com/scrapy/scrapy/issues/3511

But localizing files without an internationalization system in place is not maintenable.

PR #4840, opened 2020-10-09
improve the log configuration with using RotatingFileHandler

### LINKED ISSUE #3628: Crawler logs are cut by day

PyCatFish:
Current log configuration makes only one log file that will become very big,  i think its necessary to offer more options about this.

Resolves #3628


Gallaecio:
Could you make this optional, so the current behavior continues to be the default behavior?

PR #4848, opened 2020-10-17
Add .jsonl file extension

divtiply:
.jsonl is suggested by https://jsonlines.org/ as file extension for JSON lines files.


elacuesta:
Looks good, but it's missing a mention in [the corresponding docs](https://docs.scrapy.org/en/2.4/topics/feed-exports.html#topics-feed-format-jsonlines) (`jl` is missing too). Could you please update that section as well?

scrapy/settings/default_settings.py
```diff
@@ -154,6 +154,7 @@
 FEED_EXPORTERS_BASE = {
     'json': 'scrapy.exporters.JsonItemExporter',
     'jsonlines': 'scrapy.exporters.JsonLinesItemExporter',
+    'jsonl': 'scrapy.exporters.JsonLinesItemExporter',
```
kmike, line 157: 👍

kmike:
+1 to add .jsonl as an auto-detected extension.

I'm not so sure about recommending it by default instead of .jl in examples though; it might be a more complex discussion. 
Most likely, @pablohoffman (Scrapy original author) coined "json lines" term, and it is a long tradition of naming these files .jl in Scrapy. See also: https://github.com/wardi/jsonlines/issues/14

What do you think about splitting these two changes (supporting .jsonl vs switching to .jsonl), so that they can be discussed & merged separately?

divtiply:
I personally don't see a reason to use .jl extension in examples. jsonlines.org suggests .jsonl file extension. .jl file extension is used by julia language for its scripts and is recognized as such by vs code editor default setup.

pablohoffman:
Thanks for the mention @kmike. I'm personally not married to the `jl` extension. If the world has adopted a new standard extension for these file types, Scrapy should probably switch to it.

docs/topics/feed-exports.rst
```diff
@@ -45,6 +45,7 @@ JSON lines
 ----------
 
  * Value for the ``format`` key in the :setting:`FEEDS` setting: ``jsonlines``
+   or ``jsonl``
```
Gallaecio, line 48: I have a slight preference towards leaving this as it was, even if we support `jsonl` and `jl` as aliases.

But, _if_ we add it here, we should also add `jl` for completion.

Gallaecio (approved):
:heavy_check_mark: given @pablohoffman’s stance and the Julia argument

kmike:
I'm fine with switching to .jsonl by default.

docs/topics/feed-exports.rst
kmike: +1 to keep it without alias.

divtiply:
.jl support is still available in this PR though intentionally not mentioned in the docs

docs/topics/feed-exports.rst
wRAR: What's your reasons to keep just `jsonlines`?

Gallaecio: We *must* give the freedom to users to use whatever file extension they wish.
We *should* recognize file extensions of JSON Lines files, to use JSON Lines format automatically when users use any of those file extensions.
We *can* ask users to set `"format": "jsonlines"` if they want files with arbitrary extensions to use JSON Lines format.

I believe we happen to support `"format": "jl"` and `"format": "jsonl"` because that [makes our code simpler](https://github.com/scrapy/scrapy/blob/9f81de2a50568c3bfefc731d91f326ff14bcfd76/scrapy/utils/conf.py#L132) (otherwise we would need to split `FEED_EXPORTERS`, e.g. have a setting mapping MIME types to exporters, and a separate setting mapping file extensions to MIME types), not because we want to allow shorter, less-readable setting definitions.

So I would call the fact that we support `"format": "jl"` and `"format": "jsonl"` an implementation detail, and by not documenting it, it could be argued that we are free to drop support for it without backward-compatibility concerns if that makes sense in the future.

kmike (approved):
The PR looks good to me as-is; +1 to merge after resolving merge conflicts.

tmpbook:
image

I prefer `.jl` to `.jsonl` like vscode do.

wRAR:
@tmpbook it has a custom icon because VS Code thinks it's https://code.visualstudio.com/docs/languages/julia

Which is an argument in favor of using `.jsonl`.

tmpbook:
> @tmpbook it has a custom icon because VS Code thinks it's https://code.visualstudio.com/docs/languages/julia
> 
> Which is an argument in favor of using `.jsonl`.

thanks

PR #4853, opened 2020-10-20
SqliteCacheStorage

trentbuck:
(This is my first ever browser-based PR; sorry if I do it wrong.)

I wanted to do basic statistical analysis on the HTTP cache, and also to expire old pages (so the cache doesn't grow unboundedly).  I couldn't see an easy way to do that with dbm+pickle, so I tried a few strategies before settling on simple sqlite3+json.

I have run this on 9 spiders, with cache sizes of ~500MB each, for a couple of weeks.
It seems to be working OK.
I expected sqlite3 to be slower and bigger than dbm, but all databases I tried (dbm, LMDB, KyotoCabinet, Sqlite3) were within ~2MB of the same size.

I did not bother to use sqlite3 in "noSQL" mode,
e.g. https://github.com/sqlite/sqlite/tree/master/ext/lsm1/
e.g. that actively-maintained sqlightning replacement that I forget the name of
e.g. Oracle bdb (dbm)'s sqlite front-end (apt install libdb5.3-sql)

I did not bother to performance-test rowid vs. WITHOUT ROWID.
The use of WITHOUT ROWID is a premature optimization.

Currently there is no documentation for this new storage class.

PS: I pulled this code out of a private repo; I made sure to keep the relevant history for forensic reasons, but if you just wanna squash it all down to the end result, that's cool.


trentbuck:
PS: https://github.com/LumoSQL/LumoSQL is the "new sqlightning" thing I was trying to remember.  The sales pitch is https://nlnet.nl/project/LumoSQL

Gallaecio:
While this looks nice, I wonder if it wouldn’t be better to provide it as a separate package instead of making it part of Scrapy.

Gallaecio:
@trentbuck What would you think about creating your own, separate repository just for this Scrapy component, and announce it in Reddit and get it into one of those “awesome-scrapy” lists for discoverability?

We try to keep the Scrapy core rather slim lately (see https://github.com/scrapy/itemadapter and https://github.com/scrapy/itemloaders), and while this _really_ looks nice, it probably makes sense to keep it separate.

trentbuck:
Adrián Chaves wrote:
> @trentbuck What would you think about creating your own, separate repository just for this Scrapy component, and announce it in Reddit and get it into one of those “awesome-scrapy” lists for discoverability?
> 
> We try to keep the Scrapy core rather slim lately (see https://github.com/scrapy/itemadapter and https://github.com/scrapy/itemloaders), and while this _really_ looks nice, it probably makes sense to keep it separate.

Sorry, I don't understand that at all.
Doesn't that approach end with scrapy itself being "empty", and
every scrapy user having to install dozens of trivial "left-pad" type packages?

And not only the hassle of installing them, but also not knowing they
exist and having to go discover them, instead of just looking at the
scrapy manual and having everything laid out neatly in one place.

And also audit them, because if features comes from different places, they're
probably held to different quality and security standards.

It's not like this is adding a major new feature, or any new dependencies.
Both dbm and sqlite have been essentially unchanged in CPython since 2008.


Sorry, that turned into a rant.
The bottom line is, I'm prepared to maintain this in my (private) project repo, or
to help push this upstream into scrapy.
I'm *not* prepared to maintain this as separate project half-way between those two states.

Gallaecio:
> Doesn't that approach end with scrapy itself being "empty", and every scrapy user having to install dozens of trivial "left-pad" type packages?

We aim to provide all must-have features in Scrapy, and make it easy to get access to additional features with a simple `pip install` or writing your own code.

> And not only the hassle of installing them, but also not knowing they exist and having to go discover them, instead of just looking at the scrapy manual and having everything laid out neatly in one place.

We have also been extending the documentation to include references to external projects that we feel users need to be aware of.

See for example the recent [Selecting dynamically-loaded content](https://docs.scrapy.org/en/latest/topics/dynamic-content.html) documentation page. We mention several external projects that are commonly used from Scrapy projects to deal with situations that require them.

I suggested Reddit and awesome lists because that’s usually easier to do, and can really help with discovery.

To add a new project like this into the documentation we should probably first discuss the scenarios where it’s useful, whether or not there are better approaches, and in any case write a whole topic about the handling of cached requests that serves as an excuse to mention such a project.

> It's not like this is adding a major new feature, […]

This is actually a bit of the reason why a separate project may make more sense. Most users will probably not need this. For those that may do, installing a separate package for it should not be that much of a hassle.

> I'm prepared to maintain this in my (private) project repo, or to help push this upstream into scrapy. I'm *not* prepared to maintain this as separate project half-way between those two states.

What would you think about creating it as a separate project under https://github.com/scrapy-plugins/ ?

Issue #4857, opened 2020-10-27
Cannot overwrite an output file with -O


wRAR:
Initially found at https://www.reddit.com/r/scrapy/comments/jhpc2f/o_doesnt_overwrite_on_wsl/

Looks like `_get_storage()` doesn't pass `feed_options` anywhere?

https://github.com/scrapy/scrapy/blob/75f35f558f5a9e0851c05dda85763b679d713ac1/scrapy/extensions/feedexport.py#L445

quang-nguyen-enveda:
I'll take it to fix

quang-sudo:
I'll take it to fix

Issue #4845, opened 2020-10-15
`feed_options` is not passed to build_storage()


vimagick:
https://github.com/scrapy/scrapy/blob/c340e72988fc6ec615b7b9851c3d28c16c26a839/scrapy/extensions/feedexport.py#L454-L455

I'm not sure.
Maybe we need to pass `feed_options` parameter to build_storage().

elacuesta:
Could you please elaborate?

GeorgeA92:
@elacuesta 
Without `feed_options` parameter inside `build_storage` call - Storage objects like `FileFeedStorage` ignore  overwrite `-O` option mentioned in [Scrapy 2.4 relase notes](https://docs.scrapy.org/en/latest/news.html#scrapy-2-4-0-2020-10-11) #4857

PR #4859, opened 2020-10-28
overwrite an output file with -O, (FeedExporter updated)

GeorgeA92:
related to #4857  and #4845 

`FeedExporter._get_storage()` updated - `feed_options` successfully passed to Storage objects

`CrawlCommandTest.test_overwrite_output` and
`RunSpiderCommandTest.test_overwrite_output` aimed to test `-O` arg - updated.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -680,9 +680,14 @@ class MySpider(scrapy.Spider):
         )
         return []
 """
+        with open(os.path.join(self.cwd, "example.json"), "w") as file:
+            file.write("not empty")
         args = ['-O', 'example.json']
         log = self.get_log(spider_code, args=args)
         self.assertIn('[myspider] DEBUG: FEEDS: {"example.json": {"format": "json", "overwrite": true}}', log)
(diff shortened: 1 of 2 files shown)


tests/test_commands.py
```diff
@@ -680,9 +680,14 @@ def start_requests(self):
         )
         return []
 """
+        with open(os.path.join(self.cwd, "example.json"), "w") as file:
```
kmike, line 683: `file` is a built-in in Python, should we change it to something like `fp`?

kmike (approved):
👍 Thanks for the fix and the tests @GeorgeA92!

tests/test_commands.py
GeorgeA92: Ok. I renamed it.

Gallaecio (approved):

Gallaecio:
Thank you!

wRAR:
This didn't close the linked issues, is there anything else needed for them?

Gallaecio:
I believe we simply forgot to set the pull request description to do it. Both are closed now.

Issue #4829, opened 2020-10-05
CSV item exporter does not flush on close


scientes:
### Description
The csv item exporter does not flush on close.

### Steps to Reproduce

1. Create pipeline with an CSV item exporter
2. only export few items (one or two)
3. Let spider close/finish

**Expected behavior:** All exported items are in the CSV Files

**Actual behavior:** Files with few items are empty (not even a header)

**Reproduces how often:** 100%

### Versions

Scrapy       : 2.3.0
lxml         : 4.5.2.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 20.3.0
Python       : 3.8.2 (default, Jul 16 2020, 14:00:26) - [GCC 9.3.0]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1f  31 Mar 2020)
cryptography : 2.8
Platform     : Linux-5.4.0-7642-generic-x86_64-with-glibc2.29

### Example pipline.py

```python3
from itemadapter import ItemAdapter
from scrapy.exporters import CsvItemExporter

class CricketcrawlerPipeline:
    def open_spider(self, spider):
        self.name_to_exporter = {}

    def close_spider(self, spider):
        for exporter in self.name_to_exporter.values():
            exporter.finish_exporting()

    def process_item(self, item, spider):
        exporter = self._exporter_for_item(item)
        exporter.export_item(item)
        return item

    def _exporter_for_item(self, item):
        name = item['name']
        if name not in self.name_to_exporter:
            f = open(f'{item.folder}/{name}.csv', 'wb')
            exporter = CsvItemExporter(f)
            exporter.start_exporting()
            self.name_to_exporter[name] = exporter
        return self.name_to_exporter[name]
```

Gallaecio:
They get an open file. I think it makes sense for the caller to be responsible of closing the file (just as it opened it), and not the item exporter.

scientes:
To be totally honest i didn't think about closing the file, that works as well. But for all other people who might also forget it, calling a flush on closing the CSVexporter doesn't hurt in my opinion. It just makes sure that all items are written to disc on closing the Exporter and the file-pointer is still usable after that so it shouldn't break anything.

Gallaecio:
[...]

PR #4830, opened 2020-10-05
Flush Filestream on CSVexporter

scientes:
Fixes #4829

PR #4861, opened 2020-10-30
Made CsvItemExporter flushes bytestream (solves Issue #4829)

lochem2-718:
In response to issue #4829 I added an implementationt to the method `finish_exporting()` in CsvItemExporter which flushes the bytestream after it finishes exporting.


wRAR:
There is already #4830 (marked discuss just as the original issue).

PR #4866, opened 2020-10-31
Added process_response method to robots middleware

dasdachs:
Scrapy robots middleware does not respect the robots [meta tags](https://developers.google.com/search/reference/robots_meta_tag). 

This is an attempt to add this rule to the middleware. Any help, comments, directions etc. would be greatly appreciated.

### DIFF
--- a/scrapy/downloadermiddlewares/robotstxt.py
+++ b/scrapy/downloadermiddlewares/robotstxt.py
@@ -54,6 +54,16 @@ class RobotsTxtMiddleware:
                          {'request': request}, extra={'spider': spider})
             self.crawler.stats.inc_value('robotstxt/forbidden')
             raise IgnoreRequest("Forbidden by robots.txt")
+    
+    def process_response(self, request, response, spider):
+        noindex_tags = response.xpath('//meta[@content="noindex"]')
+
+        if self._default_useragent in noindex_tags.getAll():
+            logger.debug("Forbidden by robots meta tags: %(request)s",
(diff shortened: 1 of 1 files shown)


Gallaecio:
```
  File "/home/travis/build/scrapy/scrapy/.tox/py/lib/python3.6/site-packages/scrapy/downloadermiddlewares/robotstxt.py", line 59, in process_response
    noindex_tags = response.xpath(\'//meta[@content="noindex"]/@name\').getAll()
AttributeError: \'SelectorList\' object has no attribute \'getAll\'
```

dasdachs:
@Gallaecio thanks for the feedback. 

I did not think about the point you made. But I assume it does have an affect for the links on that page, right? 

If it still seems like a good idea to you, I would like to add some tests as well.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
+}
```
Gallaecio: Please, do not add developer-specific files.

Gallaecio (commented):
> But I assume it does have an affect for the links on that page, right?

I had not though of that! Good point.

> If it still seems like a good idea to you, I would like to add some tests as well.

That would be great, thanks!

.vscode/settings.json
dasdachs: My mistake, I thought I removed it. Will double check.

PR #4873, opened 2020-11-08
Update httpcache.py

Kromitvs:
Sometimes (for instance, in sites finishing in '/', without 'html' at the end) the body is necessary to correctly identify a Response type as txt. Without passing this parameter, the body can never be used for that.

To do:

- [x] `scrapy/extensions/httpcache.py:242`
- [x] `scrapy/extensions/httpcache.py:302`
- [x] `scrapy/core/downloader/handlers/ftp.py:105`
- [x] `scrapy/core/downloader/webclient.py:113`


Kromitvs:
Update httpcache.py

This is how it's already done when httpcache is not called: https://github.com/scrapy/scrapy/blob/c292957cb19085137146bde72fe82639591c5e1c/scrapy/core/downloader/handlers/http11.py#L436

elacuesta (approved):
It'd be great if this could be tested somehow, but it seems like a fairly harmless change so I'm approving anyway.

Gallaecio:
@Kromitvs Do you think you can add a test for this?

Kromitvs:
line 242 seems to have the same issue `respcls = responsetypes.from_args(headers=headers, url=url)`

There are a few more places were the call doesn't include the body, like  `scrapy/core/downloader/webclient.py:112`. I don't know the code base that well to be sure.

I tested by scrapping the same site with and without cache. Are there other HttpCache tests I could get some inspiration from?

Gallaecio:
HTTP cache tests are at https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware_httpcache.py

If you find nothing there that helps figure out a way to test this, or if you are short on time, let me know; maybe I can have a stab at it.

Kromitvs:
Hi, sorry to say, but it seems I can't find the time for now. Maybe I can help some other way or at some other time?

kmike (approved):
The PR looks good!

PR #4874, opened 2020-11-08
Fix incorrect output on the genspider command

stummjr:
When running inside a project, the `genspider` command is outputting string placeholders rather than their actual contents.

Example:
```
$  scrapy genspider example example.com
Created spider 'example' using template 'basic' in module:
  {spiders_module.__name__}.{module}
```

This PR fixes it, by adding the missing f-string prefix to the string that represents that message.


elacuesta (approved):
Thanks Valdir! Nice to see you around 😉

kmike:
Thanks @stummjr!

PR #4878, opened 2020-11-11
Response.protocol attribute

elacuesta:
Motivated by #4769 and #4770

```
(...)
>>> response.protocol
'HTTP/1.1'
>>> 
```

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -706,7 +706,7 @@ Response objects
     .. versionadded:: 2.1.0
        The ``ip_address`` parameter.
 
-    .. versionadded:: 2.X.X
+    .. versionadded:: VERSION
        The ``protocol`` parameter.
 
     .. attribute:: Response.url
@@ -812,7 +812,7 @@ Response objects
 
     .. attribute:: Response.protocol
 
-        .. versionadded:: 2.X.X
+        .. versionadded:: VERSION
 
         A tuple containing information about the protocol that was used
         to download the response. Taken from the ``version`` attribute of the
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -693,8 +693,21 @@ Response objects
     :param ip_address: The IP address of the server from which the Response originated.
     :type ip_address: :class:`ipaddress.IPv4Address` or :class:`ipaddress.IPv6Address`
 
+    :param protocol: A tuple containing information about the protocol that was used
+        to download the response. Taken from the ``version`` attribute of the
+        corresponding :class:`twisted.web.client.Response` object, it will tipically
+        consist of the protocol and version numbers, e.g. ``(b"HTTP", 1, 1)``
+        to represent "HTTP/1.1".
+    :type protocol: :class:`tuple`
+
+    .. versionadded:: 2.X.X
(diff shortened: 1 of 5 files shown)


docs/topics/request-response.rst
```diff
@@ -693,9 +693,22 @@ Response objects
...
+        to represent "HTTP/1.1".
+    :type protocol: :class:`tuple`
+
+    .. versionadded:: 2.0.0
+       The ``certificate`` parameter.
+
     .. versionadded:: 2.1.0
        The ``ip_address`` parameter.
 
+    .. versionadded:: 2.X.X
```
kmike, line 709: I think @Gallaecio suggested using VERSION, is this right?

docs/topics/request-response.rst
```diff
@@ -693,9 +693,22 @@ Response objects
     :param ip_address: The IP address of the server from which the Response originated.
     :type ip_address: :class:`ipaddress.IPv4Address` or :class:`ipaddress.IPv6Address`
 
+    :param protocol: A tuple containing information about the protocol that was used
+        to download the response. Taken from the ``version`` attribute of the
+        corresponding :class:`twisted.web.client.Response` object, it will tipically
+        consist of the protocol and version numbers, e.g. ``(b"HTTP", 1, 1)``
+        to represent "HTTP/1.1".
+    :type protocol: :class:`tuple`
```
kmike, line 701: Would it be possible to have something more structured than a tuple, maybe a namedtuple? Is there any standard on what the tuple is - e.g. [0] is protocol name, and the rest is version?

kmike:
Nice feature 👍

docs/topics/request-response.rst
elacuesta: Do you mean as a placeholder, to have a standardized string to search for and replace before releasing?

docs/topics/request-response.rst
elacuesta: According to the docs for [`twisted.web.client.Response`](https://twistedmatrix.com/documents/20.3.0/api/twisted.web.client.Response.html), it's "HTTP version components protocol, major, minor. E.g. (b'HTTP', 1, 1) to mean b'HTTP/1.1'", but it seems to be a string for [HTTP/1.0](https://twistedmatrix.com/documents/20.3.0/api/twisted.web.http.HTTPClient.html#handleStatus) (I just realized about this, I need to update this PR).
Given that we probably will need to add it manually to the upcoming HTTP/2 implementation (somewhere [around here](https://github.com/scrapy/scrapy/blob/269fe35d6da767e56e4870ef48f00b1655c06a17/scrapy/core/http2/stream.py#L453-L461), since we don't use twisted responses), what do you think about setting a str/bytes instead?

Gallaecio: I think the safest approach may be to go with whatever the protocol actually shares. That means `HTTP/1.1` for HTTP 1.1[*](https://tools.ietf.org/html/rfc2616#section-3.1), and `h2` for HTTP/2[*](https://tools.ietf.org/html/rfc7540#section-3.1).

kmike (approved):
👍 Thanks @elacuesta! 
As @Gallaecio also checked it, I'll leave it to him to merge.

Issue #4658, opened 2020-07-05
Use black python formatter


imsahil007:
Hello Everyone! I hope you are doing okay!
I was looking at the source code and I noticed that the code indentation is poor. So, I suggest using code formatted and create tests for same. I like how "black" works. Tell me what do you think about this?
I am just a beginner to open source. Healthy criticism will be appreciated!

Gallaecio:
This is something we have discussed internally, and at the time the consensus was not to switch to Black. But we have been making style improvements to the code base for months.

Most of the code passes Flake8 checks (exceptions are configured in https://github.com/scrapy/scrapy/blob/master/pytest.ini), and we recently merged [Pylint checks](https://github.com/scrapy/scrapy/pull/3727), which we’ll review with time.

Could you indicate examples of format issues that you’ve found that are not covered by Flake8 or Pylint?

wRAR:
I've just run `black -S -l 119 scrapy tests` on the master and pushed the result to https://github.com/wRAR/scrapy/commit/1eadb1875a4dca6b209d9c1362407e7fe787ebe6 so anyone curious can preview the changes.

Gallaecio:
Looks like black does not undo string splits as I would expect: https://github.com/wRAR/scrapy/commit/1eadb1875a4dca6b209d9c1362407e7fe787ebe6#diff-bd207f7ef310a40189c53674dde833fdR141

wRAR:
I've updated the branch: https://github.com/wRAR/scrapy/commit/f842fbeaa1af80edbdb85bf8437cb4519b2a35eb

Previous run: 245 files changed, 3903 insertions(+), 4063 deletions(-)

This run: 253 files changed, 4176 insertions(+), 4115 deletions(-)

wRAR:
Compare: https://github.com/scrapy/cssselect/pull/122 has 4 times fewer changes, though most of that is in three very big files, it would be even easier to review if the changes were split between more files. So it should be possible to reformat the Scrapy repo in several batches if we decide to do that.

wRAR:
OTOH 70% of that is, curiously, in the tests.

Gallaecio:
> So it should be possible to reformat the Scrapy repo in several batches if we decide to do that.

A progressive approach works for me, just so that I don’t fall asleep during review. Shall we start with adding black but allow-listing all existing files?

wRAR:
Yeah.

avishmehta68710:
Can i work on this issue

[...]

PR #4895, opened 2020-11-19
Some type hints

elacuesta:
Just a few annotations to remove exceptions from `setup.cfg`

### DIFF
--- a/scrapy/__init__.py
+++ b/scrapy/__init__.py
@@ -22,7 +22,7 @@ __all__ = [
 
 
 # Scrapy and Twisted versions
-__version__ = pkgutil.get_data(__package__, 'VERSION').decode('ascii').strip()
+__version__ = (pkgutil.get_data(__package__, "VERSION") or b"").decode("ascii").strip()
 version_info = tuple(int(v) if v.isdigit() else v for v in __version__.split('.'))
 twisted_version = (_txv.major, _txv.minor, _txv.micro)
 
--- a/scrapy/commands/__init__.py
+++ b/scrapy/commands/__init__.py
@@ -3,6 +3,8 @@ Base class for Scrapy commands
 """
 import os
 from optparse import OptionGroup
+from typing import Any, Dict
+
 from twisted.python import failure
 
 from scrapy.utils.conf import arglist_to_dict, feed_process_params_from_cli
@@ -15,7 +17,7 @@ class ScrapyCommand:
     crawler_process = None
 
     # default settings to be used for this command instead of global defaults
-    default_settings = {}
+    default_settings: Dict[str, Any] = {}
 
     exitcode = 0
 
--- a/scrapy/http/cookies.py
+++ b/scrapy/http/cookies.py
@@ -1,8 +1,14 @@
+import re
 import time
-from http.cookiejar import CookieJar as _CookieJar, DefaultCookiePolicy, IPV4_RE
+from http.cookiejar import CookieJar as _CookieJar, DefaultCookiePolicy
 
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.python import to_unicode
+
+
+# Defined in the http.cookiejar module, but undocumented:
+# https://github.com/python/cpython/blob/v3.9.0/Lib/http/cookiejar.py#L527
+IPV4_RE = re.compile(r"\.\d+$", re.ASCII)
 
 
 class CookieJar:
--- a/scrapy/utils/httpobj.py
+++ b/scrapy/utils/httpobj.py
@@ -1,13 +1,16 @@
 """Helper functions for scrapy.http objects (Request, Response)"""
 
-import weakref
-from urllib.parse import urlparse
+from typing import Union
+from urllib.parse import urlparse, ParseResult
+from weakref import WeakKeyDictionary
+
+from scrapy.http import Request, Response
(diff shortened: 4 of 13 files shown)


scrapy/http/cookies.py
```diff
@@ -1,10 +1,14 @@
+import re
 import time
-from http.cookiejar import CookieJar as _CookieJar, DefaultCookiePolicy, IPV4_RE
+from http.cookiejar import CookieJar as _CookieJar, DefaultCookiePolicy
 
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.python import to_unicode
 
 
+IPV4_RE = re.compile(r"\.\d+$", re.ASCII)
```
elacuesta, line 9: Taken from https://github.com/python/cpython/blob/3.9/Lib/http/cookiejar.py#L527

`IPV4_RE` is neither in [`__all__`](https://github.com/python/cpython/blob/3.9/Lib/http/cookiejar.py#L28-L29) nor in the [docs](https://docs.python.org/3/library/http.cookiejar.html).

kmike: Would you mind adding something simiilar as a code comment?

scrapy/__init__.py
```diff
@@ -22,7 +22,7 @@
 
 
 # Scrapy and Twisted versions
-__version__ = pkgutil.get_data(__package__, 'VERSION').decode('ascii').strip()
+__version__ = (pkgutil.get_data(__package__, "VERSION") or b"").decode("ascii").strip()
```
Gallaecio, line 25: :lipstick: Why the change of quotes? I see the line below also uses single quotes.

Gallaecio (approved):

scrapy/__init__.py
elacuesta: It's a subtle step towards #4658 :wink:

scrapy/utils/httpobj.py
```diff
@@ -1,10 +1,10 @@
 """Helper functions for scrapy.http objects (Request, Response)"""
 
-import weakref
 from urllib.parse import urlparse
+from weakref import WeakKeyDictionary
 
 
-_urlparse_cache = weakref.WeakKeyDictionary()
+_urlparse_cache: WeakKeyDictionary = WeakKeyDictionary()
```
kmike, line 7: Would it be possible to type annotate keys/values here, and other similar places? It is a chance to make type annotations a bit more useful for reading (and probably using) the code; otherwise I suspect we'll never get to this - mypy error is silenced, all is good :)

elacuesta: I looked into [`typing`](https://docs.python.org/3/library/typing.html) and couldn't find it, so I decided to only annotate the "outer" type. 
According to [this comment](https://bugs.python.org/issue38756#msg356318), it should be possible to annotate it as a string (`from __future__ import annotations` doesn't work in py36).
I'll update this shortly.

scrapy/commands/__init__.py
```diff
@@ -15,7 +17,7 @@ class ScrapyCommand:
     crawler_process = None
 
     # default settings to be used for this command instead of global defaults
-    default_settings = {}
+    default_settings: Dict[str, Any] = {}
```
kmike, line 20: Should keys be str? We support callables as keys in settings now.

scrapy/utils/response.py
```diff
@@ -60,7 +67,7 @@ def response_httprepr(response):
     return b"".join(values)
 
 
-def open_in_browser(response, _openfunc=webbrowser.open):
+def open_in_browser(response: Response, _openfunc: Callable[[str], Any] = webbrowser.open) -> Any:
```
kmike, line 70: Nitpick: return type is not Any, but whatever _openfunc returns. TypeVar would be more precise here.

scrapy/commands/__init__.py
elacuesta: Do you mean after #3873? Those are values, keys can be other objects but only as sub-keys within other settings (`ITEM_PIPELINES`, `SPIDER_MIDDLEWARES`, etc)

scrapy/utils/response.py
elacuesta: Makes sense, but I think I need a bit of assistance. If I do:
```python
OpenInBrowserTypeVar = TypeVar("OpenInBrowserTypeVar")

def open_in_browser(
    response: Union[HtmlResponse, TextResponse],
    _openfunc: Callable[[str], OpenInBrowserTypeVar] = webbrowser.open,
) -> OpenInBrowserTypeVar:
```
I get
```
error: Incompatible default for argument "_openfunc" (default has type "Callable[[str, int, bool], bool]", argument has type "Callable[[str], OpenInBrowserTypeVar]")
```
because [`webbrowser.open`](https://docs.python.org/3/library/webbrowser.html) takes `url, new=0, autoraise=True`. However, the function is called with `_openfunc(f"file://{fname}")`, passing only a `str`. Of course I could do
```
_openfunc: Callable[[str], OpenInBrowserTypeVar] = webbrowser.open,  # type: ignore
```
but I'd like to avoid that.

I haven't found a solution so far, my next step would be to ask in SO, but maybe you know a way to solve it.

kmike:
Thanks @elacuesta!

PR #4897, opened 2020-11-19
headers_received signal

### LINKED ISSUE #1772: More possibilities to cancel downloads inside HTTP downloader handler
Hello,

Currently, a download is cancelled in the HTTP downloader if the expected size of the response is greater than the `DOWNLOAD_MAXSIZE` setting. However, there is no way to cancel a download after the headers are received and before the body is downloaded based on other conditions, such as the value of a specific header, and I see some cases where it could be useful.  

For instance, one cannot rely on LinkExtractor to filter out media links (images, videos, etc...) since a link without a media extension could still be a media. Thus, by having a way to obtain the headers of a response when they are received, one can check the value of the `Content-Type` header and trigger the cancellation of the download if necessary.  

I thought about an implementation and came up with [this](https://github.com/Djayb6/scrapy/commit/f8080748db04daf922f48650d3396ac7732d8bde) . The main idea is when the headers of the response are received, the downloader handler sends a signal `headers_received` with the `txresponse` and the `request`, and cancels the download based on the return value of the first receiver's callback. It is a quick hack but is not very intrusive. The main drawback is that when connecting to this signal in a spider, one must specifies `sender=Any` as the crawler's signal manager is not available in the downloader handler.

I'm waiting for your remarks and ideas.

Many thanks,

JB.

elacuesta:
Closes #1772

A new `headers_received` signal that fires when the headers for a response are received, before starting the download of the body. Handlers for this signal can stop the download of a response by raising a `scrapy.exceptions.StopDownload` exception, just like the handlers for the `bytes_received` signal.

Tasks:
- [x] Implementation
- [x] Tests
- [x] Docs
- [x] Fix failing tests


Gallaecio (approved):
Nice.

(not merging yet due to on-going tests)

docs/topics/exceptions.rst
```diff
@@ -85,8 +85,8 @@ StopDownload
 
 .. exception:: StopDownload(fail=True)
 
-Raised from a :class:`~scrapy.signals.bytes_received` signal handler to
-indicate that no further bytes should be downloaded for a response.
+Raised from a :class:`~scrapy.signals.bytes_received`/:class:`~scrapy.signals.headers_received`
```
Gallaecio, line 88: :lipstick: 

```suggestion
Raised from a :class:`~scrapy.signals.bytes_received` or :class:`~scrapy.signals.headers_received`
```

docs/topics/signals.rst
```diff
@@ -400,6 +400,33 @@ bytes_received
...
+    :param request: the request that generated the download
+    :type request: :class:`~scrapy.http.Request` object
+
+    :param spider: the spider associated with the response
+    :type spider: :class:`~scrapy.spiders.Spider` object
+
+.. note:: Handlers of this signal can stop the download of a response while it
+    is in progress by raising the :exc:`~scrapy.exceptions.StopDownload`
+    exception. Please refer to the :ref:`topics-stop-response-download` topic
+    for additional information and examples.
```
Gallaecio, line 428: :lipstick: Both here and in `bytes_received` above, I’m thinking that maybe this paragraph should be part of the signal description, and not a side note.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -382,6 +382,27 @@ def _cb_latency(self, result, request, start_time):
...
+            signal=signals.headers_received,
+            headers=Headers(txresponse.headers.getAllRawHeaders()),
+            request=request,
+            spider=self._crawler.spider,
+        )
+        for handler, result in headers_received_result:
+            if isinstance(result, Failure) and isinstance(result.value, StopDownload):
+                logger.debug("Download stopped for %(request)s from signal handler %(handler)s",
+                             {"request": request, "handler": handler.__qualname__})
+                with suppress(AttributeError):
```
Gallaecio, line 395: Do you know in which circumstances this happens? Maybe it would be worth indicating in a comment here.

scrapy/core/downloader/handlers/http11.py
```diff
@@ -382,6 +382,27 @@ def _cb_latency(self, result, request, start_time):
...
+                    txresponse._transport._producer.loseConnection()
+                return {
+                    "txresponse": txresponse,
+                    "body": b"",
+                    "flags": ["download_stopped"],
+                    "certificate": None,
+                    "ip_address": None,
+                    "failure": result if result.value.fail else None,
```
Gallaecio, line 403: What’s the purpose of this `if`-`else`?

Gallaecio (commented):
So glad you managed to find the issue affecting the tests! :tada: I had made some attempts myself but I had not been able to.

scrapy/core/downloader/handlers/http11.py
elacuesta: To be honest, I didn't dig too much into this part. I would much rather do `Transport.loseConnection`, which [does this check](https://github.com/twisted/twisted/blob/twisted-18.4.0/src/twisted/web/_newclient.py#L1359-L1365) and saves me one call to a private member, but it's not available in the minimum Twisted version we currently support (17.9.0)

Gallaecio: https://github.com/scrapy/scrapy/issues/4991

Gallaecio (approved):

elacuesta:
@Gallaecio please review again after https://github.com/scrapy/scrapy/pull/4897/commits/84e91b670c809aebf3cbe829b4aa2fc84f3de41d :smile:

Gallaecio:
I’m undecided on the right approach to provide the content length.

My initial reaction was: we should first fix Scrapy responses missing the Content-Length header. Then I saw https://github.com/Almad/twisted/blob/d9dccb23ffa63184e376e98bebe6a5cee32953f3/twisted/web2/channel/http.py#L202 and I though it made sense, Twisted is just following the standard. Then I thought of cases where the header and the content length do not actually match, and how in Scrapy we allow our users to still process those requests; I think exposing the Content-Length header to users would go in line with that, Scrapy users some times may need that low-level stuff.

So I’m +0 on this. I think the code looks great, I’m just not sure about the length parameter. Adding it as a parameter would require a messy deprecation later if we eventually remove it. But the only alternative I can think of would be to add the header to responses, which I think would be out of the scope of this change, and I’m unsure whether it would be the easiest thing (pick the content length like you did here and set that into the header) or require Twisted monkey patching.

Issue #4892, opened 2020-11-18
Update all links in installation guide


AKSHAYSHARMAJS:
Some of the links like [cryptography installation](https://cryptography.io/en/latest/installation/) in file `install.rst` are outdated and require an update. 
It will be nice to check for other parts of the documentation for the same issue.

mehrdadep:
I will do it, Assign it to me please.

mehrdadep:
`scarpy` only have pages in [community-staging](https://www.archlinux.org/packages/community-staging/any/scrapy/) and [community](https://www.archlinux.org/packages/community/any/scrapy/) on [archlinux](https://www.archlinux.org/), so this is also invalid `.. _AUR Scrapy package: https://aur.archlinux.org/packages/scrapy/`

Joe-Sin7h:
Can i get assigned to This issue @Gallaecio

Gallaecio:
There’s already https://github.com/scrapy/scrapy/pull/4899

PR #4899, opened 2020-11-20
Update links in installation guide

gunadhya:
Fixes https://github.com/scrapy/scrapy/issues/4892  

Insured all the other links are working.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -283,6 +283,7 @@ coverage_ignore_pyobjects = [
 intersphinx_mapping = {
     'attrs': ('https://www.attrs.org/en/stable/', None),
     'coverage': ('https://coverage.readthedocs.io/en/stable', None),
+    'cryptography' : ('https://cryptography.io/en/latest/', None),
     'cssselect': ('https://cssselect.readthedocs.io/en/latest', None),
     'itemloaders': ('https://itemloaders.readthedocs.io/en/latest/', None),
     'pytest': ('https://docs.pytest.org/en/latest', None),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -292,7 +292,6 @@ intersphinx_mapping = {
     'tox': ('https://tox.readthedocs.io/en/latest', None),
     'twisted': ('https://twistedmatrix.com/documents/current', None),
     'twistedapi': ('https://twistedmatrix.com/documents/current/api', None),
-    'cryptography' : ('https://cryptography.io/en/latest/', None),
 }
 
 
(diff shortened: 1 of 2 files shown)


wRAR (approved):
thanks!

docs/intro/install.rst
```diff
@@ -265,7 +265,7 @@ For details, see `Issue #2473 <https://github.com/scrapy/scrapy/issues/2473>`_.
 .. _cryptography: https://cryptography.io/en/latest/
 .. _pyOpenSSL: https://pypi.org/project/pyOpenSSL/
 .. _setuptools: https://pypi.python.org/pypi/setuptools
-.. _AUR Scrapy package: https://aur.archlinux.org/packages/scrapy/
+.. _ArchlinuxScrapyPackage: https://www.archlinux.org/packages/community-staging/any/scrapy/
```
elacuesta, line 268: Changing this name without changing its occurrences in the text looked strange to me, so I checked and it seems like there are in fact no occurrences at all. Should we remove it?

gunadhya: I agree it doesn't appear to be referenced anywhere. I think we can remove it. Should I make the change?

elacuesta: Yes, please do. Thanks.

elacuesta (approved):

docs/intro/install.rst
```diff
@@ -72,7 +72,7 @@ please refer to their respective installation instructions:
 * `cryptography installation`_
 
 .. _lxml installation: https://lxml.de/installation.html
-.. _cryptography installation: https://cryptography.io/en/latest/installation/
+.. _cryptography installation: https://cryptography.io/en/latest/installation.html
```
Gallaecio, line 75: Can we please switch to InterSphinx to handle this?

If we add an entry for `cryptography` to [`intersphinx_mapping`](https://github.com/scrapy/scrapy/blob/master/docs/conf.py#L283) we should be able to link with something like ```:doc:`cryptography:installation` ```.

gunadhya: Hi @Gallaecio, This is my first time working with intersphinx so please let me know I made a mistake.

docs/conf.py
```diff
@@ -291,6 +291,7 @@
     'tox': ('https://tox.readthedocs.io/en/latest', None),
     'twisted': ('https://twistedmatrix.com/documents/current', None),
     'twistedapi': ('https://twistedmatrix.com/documents/current/api', None),
+    'cryptography' : ('https://cryptography.io/en/latest/', None),
```
Gallaecio, line 294: :lipstick: Let’s keep these entries alphabetical :nerd_face:

Gallaecio (approved):
Looks good to me!

PR #4900, opened 2020-11-23
Deprecate scrapy.utils.py36 module

elacuesta:
Now that the minimum Python version is 3.6, there is no need to keep this helper in a different file and catch the `SyntaxError` when importing.

### DIFF
--- a/scrapy/utils/spider.py
+++ b/scrapy/utils/spider.py
@@ -4,7 +4,7 @@ import logging
 from scrapy.spiders import Spider
 from scrapy.utils.defer import deferred_from_coro
 from scrapy.utils.misc import arg_to_iter
-from scrapy.utils.python import collect_asyncgen
+from scrapy.utils.asyncgen import collect_asyncgen
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


scrapy/utils/spider.py
```diff
@@ -4,17 +4,14 @@
...
-except SyntaxError:
-    collect_asyncgen = None
+from scrapy.utils.python import collect_asyncgen
 
 
 logger = logging.getLogger(__name__)
 
 
 def iterate_spider_output(result):
-    if collect_asyncgen and hasattr(inspect, 'isasyncgen') and inspect.isasyncgen(result):
```
elacuesta, line 17: [`inspect.isasyncgen`](https://docs.python.org/3.6/library/inspect.html#inspect.isasyncgen) was added in Python 3.6

wRAR:
This is actually done in #4467 (not sure why I didn't think of extracting this change), though the function is moved to `scrapy.utils.asyncgen`. That module will get one more function, `as_async_generator`, but I guess it's fine to put both into `scrapy.utils.python` now that we don't need this separation because of 3.5.

wRAR (approved):

elacuesta:
Ohh :facepalm: 
Well, I guess we can still work with this PR, since it's smaller and easier to merge than #4467.
Actually, I was hesitant about putting it under `utils.python`, if there's going to be more functions I think I like a new submodule instead.

wRAR:
Well, I think it's good that this will be merged separately as it reduces the diff and works fine as a separate change.

PR #4902, opened 2020-11-23
Implement a retry request function

### LINKED ISSUE #3590: Ability to retry a request from inside a spider callback

Gallaecio:
Alternative to #3685 for implementing request retry from spider callbacks.

I prefer this approach because the implementation is simpler (basically a refactoring of existing code), while still providing a similar API, and going in line with the current implementation of retries as a downloader middleware, by not modifying `Request` or `Response` in any way.

With this implementation, there is a function which can be used to retry a request from a spider callback:

```python
# …
from scrapy.downloadermiddlewares.retry import get_retry_request

class MySpider(Spider):
    # …
    def parse(self, response):
        if not response.text:
            new_request = get_retry_request(response.request, reason='empty', spider=self)
            if new_request:
                yield new_request
            return
```

To do:

- [x] Choose the best API subset, remove the rest or make it private.
- [x] Add documentation.
- [x] Add tests.

Resolves #3590, closes #3685


scrapy/downloadermiddlewares/retry.py
```diff
@@ -31,6 +31,105 @@
 logger = logging.getLogger(__name__)
 
 
+def get_retry_request(
+    request,
+    *,
+    reason,
```
noviluni, line 37: what about adding a default reason? (i.e.: `reason='retry'` or `'default_retry'` or something similar...) in that way you don't need to give a name if you don't need to differentiate them.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -31,6 +31,105 @@
...
+
+        stats.inc_value('retry/count')
+        stats.inc_value(f'retry/reason_count/{reason}')
+        return new_request
+    else:
+        stats.inc_value('retry/max_reached')
+        logger.error("Gave up retrying %(request)s (failed %(retry_times)d times): %(reason)s",
+                        {'request': request, 'retry_times': retry_times, 'reason': reason},
+                        extra={'spider': spider})
+        return None
```
noviluni, line 75:
```suggestion
        return
```

scrapy/downloadermiddlewares/retry.py
```diff
@@ -31,6 +31,105 @@
...
+
+
+def retry_request(
+    request,
+    *,
+    reason,
```
noviluni, line 81: (same comment above)

scrapy/downloadermiddlewares/retry.py
```diff
@@ -31,6 +31,105 @@
...
+    priority_adjust=None,
+):
+    new_request = get_retry_request(
+        request,
+        reason=reason,
+        spider=spider,
+        max_retry_times=max_retry_times,
+        priority_adjust=priority_adjust,
+    )
+    if new_request:
```
noviluni, line 93: not sure if this is necessary, as we can just do this by yielding the result of `get_retry_request` (first approach in the description). I would prefer to have a good example on the docs over providing this function.

noviluni (commented):
I would prefer the first/second approach over the `RetrySpiderMixin`, the pattern it's easier if you don't need to add the mixin to the spider, however I undestand that doing `spider=self` it's not so friendly, so I can undestand/accept the third/fourth approach

scrapy/downloadermiddlewares/retry.py
Gallaecio: `unspecified`?

scrapy/downloadermiddlewares/retry.py
Gallaecio: I did it [on purpose](https://stackoverflow.com/a/15300671/939364), for readability, to make it clear that the calling code expects a return value.

scrapy/downloadermiddlewares/retry.py
noviluni: it could work :+1:

ejulio:
@Gallaecio , I was thinking about it recently and what are your thoughts on reusing `RetryMiddleware` in the following manner.

```
class RetryMiddleware: # scrapy class

        get_retry_reason = getattr(spider, 'get_retry_reason', lambda resp: None)
        self.get_retry_reason = get_retry_reason

        if request.meta.get('dont_retry', False):
            return response
        if response.status in self.retry_http_codes:
            reason = response_status_message(response.status)
        else:
            reason = self.get_retry_reason(response)
        if reason:
            return self._retry(request, reason, spider) or response
        return response


class MySpider(Spider):

    def get_retry_reason(self, response):
         return "A custom reason"

```

Basically, we'd be using the current middleware but adding an extension point to it.
This would make it easier to specify new retry conditions and it is in the proper priority chain in scrapy.
Also, the code is out from the `parse` methods, making them a bit cleaner, but allowing access to the spider.
Finally, as it doesn't end up in `parse` it might help reducing side-effects (say, processing spidermiddlwares for something that should've been retried.

Gallaecio:
@ejulio I agree that it would be best to adhere to the regular flow of Scrapy (middlewares, extensions, etc.) for this, and that it would be best to keep this in the retry middleware to avoid unnecessary processing. I had not even thought about the latter.

However, I’m not sure asking to subclass the middleware would be better. Since we are talking about enabling spider-specific conditions for retry, I think it makes sense for that code to be defined in the spider.

Your custom meta key idea, on the other hand… I think that could work. It would not require a middleware, it would be in line with the Scrapy flow, it would not let those responses go pass the retry middleware… It could even be modified by middlewares in a similar fashion to decorators. The more I think about it, the more I like it.

ejulio:
@Gallaecio , we shouldn't subclass scrapy's middleware.
The standard scrapy middleware should subscribe to `open_spider` and get the `get_retry_reason` if defined in the spider.
If not defined, it works as is, if defined, it is a hook the spider can use to handle custom retries.

I was thinking that we could go even further, but that could be a bit drastic.
We could introduce`
```
RETRY_CONDITIONS = {
    some_function: {'RETRY_TIMES': 3, other confs..}
}
```

Then, in the spider
```
custom_settings = {
    'RETRY_CONDITIONS': {
        self.custom_retry_condition: {'RETRY_TIMES': 3},
        self.other_custom_retry_condition: {'RETRY_TIMES': 3},
    }
}
```

In this scheme, the fallback to `RETRY_HTTP_CODES` would be to insert a common function if the list is defined.
But I think it's a bigger change and not sure if it's better than a single function with some conditions

Gallaecio:
I’m not convinced about the `open_spider` approach. It seems too convoluted to me. And if you want to have a different retry logic for different parts of the same spider (e.g. different callbacks), it would require you to use something like `meta` to pass information from the callback to `get_retry_reason` to condition the logic.

I agree that having custom retry times can be useful, but I think that should be done (or maybe is already possible) through `meta`.

ejulio:
Makes sense then.
Having a granular control on the request itself on amount of retries and custom condition.

Currently we don't support custom amount in meta, but should be fairly easy to add.

The custom condition, I think, is harder to do.
If we put a function in `meta` it might make the request non-serializable, disabling the disk queues.
So we need to be careful about it

Gallaecio:
Good point, I hadn’t even though of serialization. I’ll mention https://github.com/scrapy/scrapy/pull/4945 here as relatively relevant (how would someone extend request serialization to support functions in meta?).

Gallaecio:
I’ve completed the pull request as originally intended.

elacuesta (approved):

pawelmhm:
Good idea! Having option to retry in Scrapy is super useful.

```python
            new_request = get_retry_request(response.request, reason='empty', spider=self)
            if new_request:
                yield new_request
            return
```
this is still a lot of code to write, 4 lines. How about simpifying it to some exceptiopn
```python
raise RetryRequest(message)
```
?
This is much more simple. Other benefits aside from simplicity: you can keep track of retries in stats, log proper message always on retry automatically from middleware when handling it. We use something like this in our project and it works nice.

Gallaecio:
The main problem I see with raising an exception is that it prevents you from doing additional stuff in the callback, specially when the maximum number of retries is exceeded.

Imagine the request you are retrying is to fetch additional data from an item (passed in `Request.meta`), and that you want to yield the item without the additional data if the request to fetch that additional data reaches the maximum number of retries. I don’t see a clear way to do that with the `raise` approach.

pawelmhm:
> Imagine the request you are retrying is to fetch additional data from an item (passed in Request.meta), and that you want to yield the item without the additional data if the request to fetch that additional data reaches the maximum number of retries. I don’t see a clear way to do that with the raise approach

good point. You would have to probably add some errback argument to retry exception like raise RetryException(errback=self.return_item)

Gallaecio:
I’m not completely convinced yet, but I’m starting to see quite some benefits to the API you propose. I’m not completely sold on `errback` being better for something that will not happen asynchronously, but in addition to a shorter syntax I see other benefits:

- Using a spider middleware (which is how the `raise` approach can be implemented) seems more _Scrapyish_.
- We will not need to pass the `response` and the `spider` as parameters, [`process_spider_exception`](https://docs.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.process_spider_exception) already gets those.
- We can use the `errback` from the `request` instead of requiring it as a parameter.
- We can probably keep a single, optional parameter, `reason`, and make any other customization available through settings and meta keys, or making it easier to subclass the new retry spider middleware and the existing retry downloader middleware.

pawelmhm:
> We can use the errback from the request instead of requiring it as a parameter

yes, good idea!

If you'd like to I can create a PR reusing some code we have in other project and we can continue discussion there, decide which approach is better.

Gallaecio:
@pawelmhm After I add typing to this pull request, we will go forward with it, as the function can be easier for certain usage patterns (e.g. middlewares).

This does not mean a no to the exception approach. In fact, the exception spider middleware could use that function.

So it looks like Scrapy 2.5 will offer the function, but we may replace it with an exception following your work on the next Scrapy version, and leave the function around only for middleware usage and corner cases.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -23,12 +24,106 @@
...
+    follows::
+
+        def parse(self, response):
+            if not response.text:
+                new_request = get_retry_request(
+                    response.request,
+                    spider=self,
+                    reason='empty',
+                )
+                if new_request:
```
ejulio, line 63: Considering that we can `yield None`, maybe you don't need the `if`

ejulio (commented):
@Gallaecio , I get the idea and it's fine.
However, my takes are:

1. If we are already considering a new API/change (`Exception` based), maybe we shouldn't move this forward. For this I can invoke the "only explicit way to do it", and considering we usually take some time to deprecate a feature, we might need to live with two APIs for the same thing for awhile.

2. To some extent, I don't like the `get_retry_request` interface so much. It plays with `spider`, `settings` and `stats`, so lots of state going on. Not that it's a problem, but I wonder if there could be a better way. Probably, the exception approach would handle this

scrapy/downloadermiddlewares/retry.py
```diff
@@ -67,31 +162,12 @@ def process_exception(self, request, exception, spider):
...
-
-            stats.inc_value('retry/count')
-            stats.inc_value(f'retry/reason_count/{reason}')
-            return retryreq
-        else:
-            stats.inc_value('retry/max_reached')
-            logger.error("Gave up retrying %(request)s (failed %(retries)d times): %(reason)s",
-                         {'request': request, 'retries': retries, 'reason': reason},
-                         extra={'spider': spider})
+        max_retry_times = request.meta.get('max_retry_times', self.max_retry_times)
```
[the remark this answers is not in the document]
Gallaecio: The algorithm of the function is: use meta, else settings, else function parameter. So I first considered to just not pass anything from the middleware about this, and let the function handle it, as the function has access to both meta and settings, and at first sight it has the same algorithm as the middleware.

However, then I though of backward compatibility and inheritance. `self.max_retry_times` is not a private attribute of the middleware, it’s public. So there may be subclasses out there that override this attribute, for example setting a different hardcoded value or reading the value from a different source (a different setting name, multiple settings, a configuration file, etc.). In order to keep backward compatibility, I wrote this logic here and pass the value explicitly to the function.

scrapy/downloadermiddlewares/retry.py
Gallaecio: You definitely could write it without the `if`. My main concern here is readability. What about doing what you suggest and renaming the variable as `new_request_or_none`?

Gallaecio:
> 1. If we are already considering a new API/change (`Exception` based), maybe we shouldn't move this forward. […] we usually take some time to deprecate a feature.

The idea is not to eventually deprecate the function in favor of the exception. The idea is to add the exception, recommend the exception (i.e. the documentation will use the exception whenever possible), but keep the function.

So, in 2.5 we would be recommending the function to retry from a callback, in 2.6 or later we would change that recommendation to the exception approach, probably mentioning the function as an alternative where an exception is not desirable for whatever reason (mostly middlewares, but some people may prefer not to use `errback` to handle the exhaustion of retries).

I would understand if the purpose was to deprecate the function, but given it’s not, I don’t see a strong reason why we should postpone the function implementation until we also provide the exception approach.

> 2. To some extent, I don't like the `get_retry_request` interface so much. It plays with `spider`, `settings` and `stats`, so lots of state going on. Not that it's a problem, but I wonder if there could be a better way. Probably, the exception approach would handle this.

Yes, that’s the reason @pawelmhm’s approach convinced me as superior. But talking it over with @kmike, @wRAR and @elacuesta, they brought up the point that there may still be use cases for the function approach, so it would be better to provide both approaches even if we recommend the exception approach for most scenarios.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -23,12 +24,104 @@
...
+retry_logger = getLogger(__name__)
+
+
+def get_retry_request(
+    request: Request,
+    *,
+    spider: Spider,
+    reason: Union[str, Exception] = 'unspecified',
+    max_retry_times: Optional[int] = None,
+    priority_adjust: Union[int, float, None] = None,
```
kmike, line 44: can it be float?

elacuesta: `Request.priority` [can only be int](https://github.com/scrapy/scrapy/blob/master/scrapy/http/request/__init__.py#L27-L28), `new_request.priority = request.priority + priority_adjust` would make it a float if `priority_adjust` were a float. Updated at d458ccff3b2d6df94df1aa86eeb7d2505d62f2d6.

/cc @wRAR

kmike:
Thanks @Gallaecio, @elacuesta, @pawelmhm, @ejulio and @noviluni!

PR #4903, opened 2020-11-24
Update links in install.rst

Rustbolt:
Updated broken links in Installation guide with working links.


elacuesta:
Thanks for the contribution, but this particular issue is being solved at #4899 already.

PR #4910, opened 2020-11-30
PEP 8 - Changed class name to camel case

azenkwed:
PEP 8 - Changed class name to camel case.
Small first contribution.

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -20,7 +20,7 @@ from scrapy.utils.misc import md5sum
 from scrapy.utils.python import to_bytes
 
 
-class NoimagesDrop(DropItem):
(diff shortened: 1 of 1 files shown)


elacuesta:
Thanks for the contribution!
Actually, I wonder if we shouldn't just remove the exception, it's not documented and it doesn't seem to be used anywhere else:
```
$ grep -ri "noimagesdrop" scrapy tests 
scrapy/pipelines/images.py:class NoimagesDrop(DropItem):
```
@Gallaecio @kmike @wRAR  Any thoughts?

wRAR:
Yeah, looks like it was never used. Should we deprecate it first?

wRAR:
Closing in favor of #5368

PR #4913, opened 2020-12-01
Update __init__.py import

azenkwed:
Update __init__.py import to fix import error.

### DIFF
--- a/tests/keys/__init__.py
+++ b/tests/keys/__init__.py
@@ -16,8 +16,8 @@ from cryptography.x509 import (
     NameAttribute,
     random_serial_number,
     SubjectAlternativeName,
+    NameOID
 )
-from cryptography.x509.oid import NameOID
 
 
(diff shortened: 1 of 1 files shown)


wRAR:
What problem does this solve?

elacuesta:
I'm +0 on this. I haven't seen any errors from this particular import (`NameOID` is imported [here](https://github.com/pyca/cryptography/blob/3.2.1/src/cryptography/x509/__init__.py#L94)).

azenkwed:
It doesn't solve an issue. It was just cleaning the code as bit as I'm reading through it.

Gallaecio:
Thanks for the pull request, but I think it’s a change for the worse.

In the [documentation](https://cryptography.io/en/stable/x509/reference/#cryptography.x509.oid.NameOID) they document the import path that the code currently uses. If they ever decide to remove one of the import paths, I think they would remove the one that is _not_ documented.

azenkwed:
Thank you.

PR #4917, opened 2020-12-07
Add dont_encode option in scrapy.http.Request class.

### LINKED ISSUE #833: Prevent URL encoding option

Ivche1337:
Fixes #833, closes #3255, closes #3542.

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -18,9 +18,9 @@ class Request(object_ref):
 
     def __init__(self, url, callback=None, method='GET', headers=None, body=None,
                  cookies=None, meta=None, encoding='utf-8', priority=0,
-                 dont_filter=False, errback=None, flags=None, cb_kwargs=None):
+                 dont_filter=False, errback=None, flags=None, cb_kwargs=None, dont_encode=False):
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
Please, have a look at the previous attempt: https://github.com/scrapy/scrapy/pull/3542

The only thing the previous attempt is missing is raising the minimum version of w3lib that Scrapy requires, which should become `1.22.0` as part of this change. That requires editing `setup.py` accordingly, and `tox.ini`.

Then it would be great to have a test, which could be inspired in [w3lib’s](https://github.com/scrapy/w3lib/commit/0b891dc3afc044321e9839c68f44616ff46e29d7#diff-6e407480bfb30fa51ddc2501641e06ed9ec6c0b351c523c59427341df198d572R80).

kmike:
I still wonder about the API; the alternative to `dont_encode` option is to support bytes url or special URL objects, not just str, with semantics "we use these bytes as-is as an url". With Python 3 move supporting bytes URL is easier to do, though it'd require some other changes.

PR #4922, opened 2020-12-10
Update dependencies

elacuesta:
I'm having some issues at https://github.com/scrapy/scrapy/pull/4897/checks?check_run_id=1530700305, I want to make sure it's due to some actual change in the code and not because of the new dependency resolution stuff in pip 20.3.

### DIFF
--- a/tests/requirements-py3.txt
+++ b/tests/requirements-py3.txt
@@ -1,8 +1,8 @@
 # Tests requirements
 attrs
 dataclasses; python_version == '3.6'
-mitmproxy; python_version >= '3.7'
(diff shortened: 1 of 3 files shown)


tests/requirements-py3.txt
```diff
@@ -17,5 +17,5 @@ uvloop; platform_system != "Windows"
 bpython
 brotlipy  # optional for HTTP compress downloader middleware tests
 zstandard  # optional for HTTP compress downloader middleware tests
-ipython
```
elacuesta, line 20: Not specifying a version leads to [failed build](https://github.com/scrapy/scrapy/pull/4922/checks?check_run_id=1531501862), specifying a minimum version leads to the job hanging indefinitely.

PR #4924, opened 2020-12-11
Migrate CI to GH actions

elacuesta:
Based on #4814


elacuesta:
Currently I'm having the following issues:
- [ ] Jobs for pinned versions fail when installing `ipython`. I'm almost sure it has to do with the new pip dependency resolver (see #4922)
- [ ] The `pypy3` job is failing, I'm not sure why. The logs are very long, I was able to reproduce [the failure for a single test file](https://gist.github.com/elacuesta/38d54dbe84d732c8f25639a193d3f120).

tests/test_webclient.py
```diff
@@ -414,6 +417,8 @@ def testPayload(self):
         ).addCallback(self.assertEqual, to_bytes(s))
 
     def testPayloadDisabledCipher(self):
+        if sys.implementation.name == "pypy" and parse_version(cryptography.__version__) <= parse_version("2.3.1"):
+            self.skipTest("This does work in PyPy with cryptography<=2.3.1")
```
elacuesta, line 421: Well this is awkward, I don't really know why this happens :shrug:

Gallaecio: Is a “not” missing in the skip message?

As for the reason, it would be ideal to find it out and mention it. I’ve found https://github.com/pyca/cryptography/commit/ce6fe07632c17f16221f02ad19c6875327cdd10a , but it was not introduced in 2.3.1, but in 2.3. Could it be the cause? It seems like it would cause things to break starting with 2.3, not to be fixed starting from 2.3.1.

The only change specific to 2.3.1 was [the rebuild of wheels with a newer OpenSSL version](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#231---2018-08-14): 1.1.0g → 1.1.0h. From the [OpenSSL changelog](https://www.openssl.org/news/openssl-1.1.0-notes.html) I cannot tell much, but if we are sure that 2.3.1 is the point where things change, I guess we can assume that the new wheels are what fixes things, and we can mention that in the test.

elacuesta:
@Gallaecio @kmike @wRAR Uploading to the testing PyPI instance works: see https://github.com/elacuesta/scrapy/runs/1605715787 for 5521f2bd9bbce89005abdc2866e4bb5415443880 (removed other jobs to save time, hardcoded name and version - different name since [`scrapy` is taken](https://test.pypi.org/project/scrapy/))

kmike:
Nice @elacuesta! 

I meant to ask this before - do you know if wwe can have several workflow files, and e.g. separate test workflow from the publishing workflow?

elacuesta:
Indeed it's possible, I just split them into four workflows, looks like a good idea to improve organization. One missing thing though seems to be the ability to depend on jobs from other workflows (the [`needs`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) key only works within the same workflow, as seen in [this test flow](https://github.com/scrapy/scrapy/actions/runs/448211867)). Initially I configured the jobs to run the _publish_ one only if the previous ones succeeded, but perhaps we don't mind about that?

.github/workflows/tests-ubuntu.yml
```diff
@@ -0,0 +1,70 @@
+name: Tests - Ubuntu
+on: [push, pull_request]
+
+jobs:
+  tests-ubuntu:
+    name: Tests - Ubuntu
```
kmike, line 6: This is super-minor, but Github shows status as Tests - Ubuntu / Tests - Ubuntu

elacuesta: Right, seems hard to change though. If the workflow name is omitted, the full name of the file is used, while if the job name is omitted, the job key is used:
no_names

I tried setting the same name for the two test files, hoping they would get grouped, but it wasn't the case:
Same name - not grouped

Gallaecio (approved):
Looks good to me. :crossed_fingers: about `publish.yml`.

README.rst
```diff
@@ -10,8 +10,8 @@ Scrapy
    :target: https://pypi.python.org/pypi/Scrapy
    :alt: Supported Python Versions
 
-.. image:: https://img.shields.io/travis/scrapy/scrapy/master.svg
-   :target: https://travis-ci.org/scrapy/scrapy
+.. image:: https://github.com/scrapy/scrapy/workflows/Build/badge.svg
```
kmike, line 13: I'm not sure this name exists now, and it seems the names are different for Windows and Ubuntu now, sso there should be 2 badges.

.github/workflows/tests-ubuntu.yml
kmike: Again, a nitpick, but what do you think about using shorter names? This is how the popup looks like, when you click on a check mark near a commit: 

изображение

Proposal: instead of naming a job "tests-ubuntu", name it "tox". Keep workflow names as-is.

README.rst
elacuesta: Updated, thanks for letting me know. Related to this: should I add jobs for macOS?

kmike: It'd be nice, but did we have them before this PR? 
I think that's fine to do it separately, unless you're "in a flow" now.

elacuesta: Done!

.github/workflows/tests-ubuntu.yml
elacuesta: Indeed, names are still too long, thanks for insisting on this. I just renamed both the workflows and the jobs, looks much cleaner to me, what do you think?
ci-popup

Gallaecio:
Given there’s 1 badge per workflow, I wonder if maybe it would make sense to keep only 2 workflows, tests and publish. But badges are not really a strong argument, so do not take it to heart, just a though :)

kmike (approved):
👍

PR #4935, opened 2020-12-21
Avoid exceptions in is_generator_with_return_value

### LINKED ISSUE #4477: is_generator_with_return_value raises IndentationError with a flush left doc string
### Description

Code that is accepted by the python interpreter raises when fed through `textwrap.dedent`

### Steps to Reproduce

1. Create `is_generator_bug.py` with the content below (which I simplified from [the `is_generator_with_return_value` method body](https://github.com/scrapy/scrapy/blob/2.0.1/scrapy/utils/misc.py#L186-L187)
2. Run `python is_generator_bug.py`
3. Observe the kaboom

```python
import ast
import inspect
from textwrap import dedent
class Bob:
    def doit(self):
        """
this line is flush left
        """
        if True:
            yield 1234

if __name__ == '__main__':
    b = Bob()
    c = b.doit
    if inspect.isgeneratorfunction(c):
        tree = ast.parse(dedent(inspect.getsource(c)))
```

No Error

```console
$ python3.7 is_generator_bug.py
  File "is_generator_bug.py", line 16, in <module>
    tree = ast.parse(dedent(inspect.getsource(c)))
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    def doit(self):
    ^
IndentationError: unexpected indent
```

100%

### Versions

```
Scrapy       : 2.0.1
lxml         : 4.5.0.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.5.2
w3lib        : 1.21.0
Twisted      : 20.3.0
Python       : 3.7.7 (default, Mar 11 2020, 23:30:22) - [Clang 10.0.0 (clang-1000.11.45.5)]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1d  10 Sep 2019)
cryptography : 2.8
Platform     : Darwin-17.7.0-x86_64-i386-64bit
```

elacuesta:
Closes #4477

Implementing @mdaniel's suggestion from https://github.com/scrapy/scrapy/issues/4477#issuecomment-611037920. Also, the function catches `SyntaxError`s in order to prevent further errors like this one in the future.


tests/test_utils_misc/test_return_with_argument_inside_generator.py
```diff
@@ -3,6 +3,28 @@
...
+
+def top_level_function_without():
+    """
+docstring
+    """
+    url = """
+https://example.org
+"""
+    yield 1
+    return
```
elacuesta, line 25: My suggestion from https://github.com/scrapy/scrapy/issues/4477#issuecomment-611018866 was not working correctly for these cases.

scrapy/utils/misc.py
```diff
@@ -242,7 +242,11 @@ def warn_on_generator_with_return_value(spider, callable):
     Logs a warning if a callable is a generator function and includes
     a 'return' statement with a value different than None
     """
-    if is_generator_with_return_value(callable):
+    try:
+        should_warn = is_generator_with_return_value(callable)
+    except SyntaxError:  # includes IndentationError
```
Gallaecio, line 247: I guess this cannot be tested without unclean mocking, so :heavy_check_mark:

Gallaecio (approved):

elacuesta:
Full diff coverage reached :raised_hands:

scrapy/utils/misc.py
apalala, line 247: Here it would be better to capture `except SyntaxError as e:` and find a way to include `e` in the reporting.

elacuesta: I beg to differ, this function only takes care of warning if the passed function is a generator with a return value different than `None`. Other pieces of code will fail loudly in case of `SyntaxError` (and probably even fail earlier, this is likely a safety web that's not even going to be reached).

Gallaecio (approved):

Gallaecio:
Suggestion for a message to log as a warning upon finding a `SyntaxError`:

> Unable to determine whether or not `<callable>` is a generator with a return value. This will not prevent your code from working, but it prevents Scrapy from detecting potential issues in your implementation of `<callable>`. Please, report this in the Scrapy issue tracker (https://github.com/scrapy/scrapy/issues), including the code of `<callable>`.

Gallaecio (approved):

elacuesta:
5902cd2130a09e58223683491bca1542c59f8afe passes CI, but I don't understand why this seemingly harmless change breaks 🤷‍♀️ :
```diff
diff --git scrapy/utils/misc.py scrapy/utils/misc.py
index 5c986eed..8e97298b 100644
--- scrapy/utils/misc.py
+++ scrapy/utils/misc.py
@@ -242,6 +242,7 @@ def warn_on_generator_with_return_value(spider, callable):
     Logs a warning if a callable is a generator function and includes
     a 'return' statement with a value different than None
     """
+    callable_name = spider.__class__.__name__ + "." + callable.__name__
     try:
         if is_generator_with_return_value(callable):
             warnings.warn(
@@ -253,7 +254,6 @@ def warn_on_generator_with_return_value(spider, callable):
                 stacklevel=2,
             )
     except IndentationError:
-        callable_name = spider.__class__.__name__ + "." + callable.__name__
         warnings.warn(
             f'Unable to determine whether or not "{callable_name}" is a generator with a return value. '
             'This will not prevent your code from working, but it prevents Scrapy from detecting '
```

Any explanation is most welcome!

Gallaecio (approved):
I would love to understand that as well, but, if it works…

Issue #4833, opened 2020-10-06
xmliter_lxml cannot find namespaced node names


Gallaecio:
In short: make the `LxmlXmliterTestCase.test_xmliter_namespaced_nodename` test (introduced in #4746) pass.

For background, see https://github.com/scrapy/scrapy/pull/861

Gallaecio:
This was fixed in 2.11.1.

PR #4939, opened 2020-12-27
passing test for LxmlXmliterTestCase.test_xmliter_namespaced_nodename

AKSHAYSHARMAJS:
making the `LxmlXmliterTestCase.test_xmliter_namespaced_nodename` test pass.

Fixes (#4833)

### DIFF
--- a/tests/test_utils_iterators.py
+++ b/tests/test_utils_iterators.py
@@ -1,6 +1,5 @@
 import os
 
-from pytest import mark
 from twisted.trial import unittest
 
 from scrapy.utils.iterators import csviter, xmliter, _body_or_str, xmliter_lxml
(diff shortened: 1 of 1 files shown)


AKSHAYSHARMAJS:
The test was failing because of `next` function reach to an end and cannot find more nodenames rasing exception StopIteration error. I used try for that or we can just apply the normal loop instead of using `next`.

Gallaecio:
The issue here is not the test itself, but the implementation. Where `xmliter` finds the node, `xmliter_lxml` fails to find it. The goal is to improve the implementation of `xmliter_lxml` so that `next()` in that test succeeds at finding the node.

AKSHAYSHARMAJS:
Thanks, @Gallaecio now I see the problem, `lxml.etree.iterparse` is not finding the desired tag, therefore, giving no nodes.

AKSHAYSHARMAJS:
Sorry, I misunderstood the issue earlier. I have updated the test accordingly. @Gallaecio Please Check.

Gallaecio:
I did not have a deep look at your updated changes, but as long as your changes do not modify the implementation of `xmliter_lxml` and only affect test files, I can only assume the change is not what is expected from #4833, sorry.

#4833 is about a test that is failing because the code being tested does not work as expected. The code, _not_ the test, needs to be fixed. The only change needed in the test file is to remove the pytest mark. Then the implementation of `xmliter_lxml` needs to change so that the test passes in its current form.

Mind that this is probably not an easy issue to fix. If fixing `xmliter_lxml` seems hard, you may want to consider to close this pull request and look for a different issue. I say this because not only this issue may not be easy to fix, but it is also not an issue affecting many people, so fixing it may not be as gratifying as fixing other stuff.

AKSHAYSHARMAJS:
Okay @Gallaecio. Thanks, I think its better to look for another issue

PR #4940, opened 2020-12-28
Fix warning: Expected type 'bool', got 'int' instead

azenkwed:
Minor fix: **twisted.internet.defer.DeferredList**  parameter  **consumeErrors** expects _boolean_, got _integer_ instead.

### DIFF
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -86,7 +86,7 @@ class MediaPipeline:
         info = self.spiderinfo
         requests = arg_to_iter(self.get_media_requests(item, info))
         dlist = [self._process_request(r, info, item) for r in requests]
-        dfd = DeferredList(dlist, consumeErrors=1)
+        dfd = DeferredList(dlist, consumeErrors=True)
         return dfd.addCallback(self.item_completed, item, info)
 
     def _process_request(self, request, info, item):
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
     name='Scrapy',
     version=version,
     url='https://scrapy.org',
-    project_urls = {
+    project_urls={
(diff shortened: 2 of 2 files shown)


elacuesta:
Looks correct according to [the docs](https://twistedmatrix.com/documents/current/api/twisted.internet.defer.DeferredList.html#__init__). Could you please also update [this line](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/utils/defer.py#L108)?
When does this warning appear? I couldn't find any occurrences in the tests.

Gallaecio:
Looks like it’s safe to change, as the lowest version of Twisted that we support [also has the parameters as booleans in the documentation](https://twistedmatrix.com/documents/17.9.0/api/twisted.internet.defer.DeferredList.html#__init__).

The old syntax [can be seen in the documentation of Twisted 10](https://twistedmatrix.com/documents/10.0.0/api/twisted.internet.defer.DeferredList.html#__init__), for example.

azenkwed:
> Looks correct according to [the docs](https://twistedmatrix.com/documents/20.3.0/api/twisted.internet.defer.DeferredList.html#__init__). Could you please also update [this line](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/utils/defer.py#L108)?
> When does this warning appear? I couldn't find any occurrences in the tests.

I will update [this line](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/utils/defer.py#L108). 
**fireOnOneErrback** should also be updated, as it also expect a boolean, according to the doc.
If the documentation suggest using a boolean, then I advise the update since this update doesn't break the code.

Gallaecio (approved):

elacuesta (approved):

elacuesta:
Thanks @azenkwed!

PR #4943, opened 2021-01-04
Handle Check if single Image URL needs to be Downloaded

IkramKhanNiazi:
- Previously, we need need to Download a single Image URL, still, we need to make that URL a list i.e. `image_urls: [URL]`
- After this fix, we don't need to make it a list, `image_urls: URL` will work.
- A better user experience.

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -160,6 +160,7 @@ class ImagesPipeline(FilesPipeline):
 
     def get_media_requests(self, item, info):
         urls = ItemAdapter(item).get(self.images_urls_field, [])
+        urls = urls if isinstance(urls, list) else [urls]
         return [Request(u) for u in urls]
 
     def item_completed(self, results, item, info):


Gallaecio:
I’m neither in favor nor against the idea. I get that saving 2 characters to fetch a single URL is shorter, but I think the value of `image_urls` being an iterable of strings is more intuitive.

If there is support to move this change forward, I would suggest modifying the implementation, so that string iterables other than lists (e.g. tuples) continue to work (I assume that tuples work with the current implementation, but I have not checked, so I may be wrong).

elacuesta:
I must say I'm -1 on this. `image_urls` is plural, I think allowing for it not to be an iterable could lead to unnecessary confusion. [The docs](https://docs.scrapy.org/en/latest/topics/media-pipeline.html) do not explicitly state that an iterable is expected, but there is [an example](https://docs.scrapy.org/en/latest/topics/media-pipeline.html#custom-images-pipeline-example) that iterates over the field. Perhaps we could clarify it further in there.
That said, if we were to accept this change, [the files pipeline](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/pipelines/files.py#L487-L489) would need to be updated as well, and tests and docs would be good too.

Issue #4923, opened 2020-12-10
Store more data in the HTTP cache


wRAR:
Currently there is no single interface for `HTTPCACHE_STORAGE` implementations to get the data to store, and both `DbmCacheStorage` and `FilesystemCacheStorage` contain ad-hoc code which stores only basic data, which is even different in these two implementations. While things like response status, headers and body are stored, many other things are not, so a response loaded from the cache lacks these additional things.

nyov:
I've been writing some custom httpcache backends for mongodb, postgresql and ceph/rados in the past, and previously asked for the move of httpcache logic and backends into it's own domain: https://github.com/scrapy/scrapy/pull/3137
I did some minor unification work between backend implementations in there. But I've been missing the time for more major stuff, like the integration of my custom backends or the lovely LevelDB deduplication stuff.

So with that background, I'm curious to know what "many other things" you find lacking in the httpcache data being stored?
Because I'd probably be interested in having that as well.

But unless you're trying to build memento capabilities into the cache, I don't see off-hand what might be missing and can't be extracted from the requests and their responses themselves.
Except for transport layer information perhaps -- like TLS chain or handshake info, proxy IPs, target site loadbalancer IP, etc.
Do you mean that kind of stuff?

Also I've done WARC plain-text file conversions into scrapy-style Request-Response "tuples" database, before. That WARC format stores a hella bunch of (useless IMHO) metadata, and I can say I didn't think scrapy's cache was missing out on much, based on that? Most of that metadata was integrity checks against WARC plaintext corruption bugs AFAICT - record UUIDs, info UUIDs, block digests for every little thing... I didn't really see the point.

wRAR:
> So with that background, I'm curious to know what "many other things" you find lacking in the httpcache data being stored?

We looked at the relevant code when discussing a new Response attribute and checking if we should modify some code so that this attribute is saved and loaded, or it will be done automatically. We found that the data that is saved is listed explicitly in the code and, for the file storage, includes only the request URL, method, headers and body and the response status, URL, headers and body. It doesn't include meta, cb_kwargs, flags or any recently added attributes such as ip_address. Any new stuff will be also missing.

nyov:
I see the point about adding ip_address (and basically all transport-related metainfo) about a connection.

[...]

PR #4945, opened 2021-01-05
Request and response serialization

### LINKED ISSUE #1450: Response.to_dict()
I found myself wanting to have a quick overview about some requests/responses, so the first thing that came to mind was trying `dict(response)`, followed by `response.as_dict()`. It didn't work ;)

Unless I miss some other built-in way to do this, I think it'd be nice to have a method that returns a dict of all the "static data" of a response or request object; e.g. `r.status`, `r.url`, `r.encoding`, `r.flags`, `r.headers`, `r.meta`, `r.body` (, `r.request`) -- instead of needing to list them all, phew.

`r.body` could potentially something big, maybe a `.to_dict()` method should probably accept a parameter to exclude some key(s), instead of filtering them out again later in a list comp or such.

Also, the simplest way to turn a response into an item could then be

``` py
def parse(self, response):
    item = dict(response)
    # or
    item = response.to_dict()
    return item.update(some_other_stuff)
```

elacuesta:
Closes #1450, related to #4923

This PR deprecates `scrapy.utils.reqser` in favour of `Request.to_dict` and ``Request.from_dict``


scrapy/http/response/text.py
```diff
@@ -37,6 +37,15 @@ def __init__(self, *args, **kwargs):
         self._cached_selector = None
         super().__init__(*args, **kwargs)
 
+    @classmethod
+    def from_dict(cls: Type[ResponseType], d: dict) -> ResponseType:
+        """
+        Create a Response object from a dict which keys match a Response's ``__init__`` parameters.
+        """
+        response = super().from_dict(d)  # type: ignore
```
elacuesta, line 45: https://github.com/python/mypy/issues/9282

scrapy/http/request/__init__.py
```diff
@@ -43,6 +57,22 @@ def __init__(self, url, callback=None, method='GET', headers=None, body=None,
         self._cb_kwargs = dict(cb_kwargs) if cb_kwargs else None
         self.flags = [] if flags is None else list(flags)
 
+    @classmethod
+    def from_dict(cls: Type[RequestType], d: dict, spider: Optional[SpiderType] = None) -> RequestType:
+        """
+        Create a Request object from a dict which keys match a Request's ``__init__`` parameters.
```
Gallaecio, line 63:
```suggestion
        Create a Request object from a dict whose keys match a Request's ``__init__`` parameters.
```

Gallaecio (approved):
Looks good to me.

As for coverage:

-   Maybe we could have the deprecated functions tested as well through test case subclassing, but I think it would be just for the sake of making the coverage report greener, rather than giving us actual value.

- The other 2 misses/partials seem to be about lines that were not covered before either, so it should be OK to leave them uncovered.

scrapy/http/request/__init__.py
elacuesta: Nice

elacuesta:
Test coverage improved, thanks for bringing it up @Gallaecio :-)

scrapy/http/request/__init__.py
```diff
@@ -43,6 +57,22 @@ def __init__(self, url, callback=None, method='GET', headers=None, body=None,
         self._cb_kwargs = dict(cb_kwargs) if cb_kwargs else None
         self.flags = [] if flags is None else list(flags)
 
+    @classmethod
+    def from_dict(cls: Type[RequestType], d: dict, spider: Optional[SpiderType] = None) -> RequestType:
```
kmike, line 61: Hmm.. This reads (and I think mypy understands it this way) as it returns an instance of cls, but it can return arbitrary class - the one saved when serializing. So FormRequest.from_dict can return an object of arbitrary class, not just FormRequest. I wonder if that was the reason to have deserialization as a function, not as a classmethod.

elacuesta: Interesting, I hadn't thought about that. What do you think about [this changeset](https://github.com/scrapy/scrapy/compare/9ef7b99ba30ede7fc7227f75965423802e0c7dbd...bf62b608ed95b924eb17b46d91fbf97bf5342379)?

scrapy/http/request/__init__.py
```diff
@@ -141,3 +170,61 @@ def from_curl(cls, curl_command, ignore_unknown_options=True, **kwargs):
...
+            cb = _find_method(spider, cb)
+        eb = self.errback
+        if callable(eb):
+            eb = _find_method(spider, eb)
+        d = {
+            "url": to_unicode(self.url),  # urls should be safe (safe_string_url)
+            "callback": cb,
+            "errback": eb,
+            "headers": dict(self.headers),
+            "_encoding": self._encoding,
```
kmike, line 192: I'm not sure `_encoding` is handled properly in the new from_dict implementation

scrapy/http/request/__init__.py
Gallaecio: I’m not sure about forcing `FormRequest` to be the return value when using `FormRequest.from_dict` (or any other request subclass), but not forcing it for `Request.from_dict`.

I think we should either returns an instance of whatever class was serialized, raise a ValueError if the serialized class does not match the `from_dict` class (which would probably be useless in scenarios where you cannot know if that’s the case beforehand), or we go back to having a function, at least for the `from_dict` part.

scrapy/http/request/__init__.py
elacuesta: Good catch. Is working correctly because `encoding` is in `Request._attributes`, and so it gets set in the next `for` loop, but this `_encoding` assignment is unnecessary. Thanks for noticing.

scrapy/http/request/__init__.py
elacuesta: I'd prefer to keep all logic within the classes. Not forcing a particular class from `Request` makes sense to me because it's the highest class in the hierarchy. Then again, I do think it makes sense to return a specific instance if a user invokes a specific class. I believe this difference gives us good flexibility, i.e. you want your own class: call `YourRequestClass.from_dict`, you want whatever the original request was, call `Request.from_dict`. It shouldn't be too confusing for users, but I can document it properly to make it explicit.

scrapy/squeues.py
```diff
@@ -8,8 +8,12 @@
 
 from queuelib import queue
 
+<<<<<<< HEAD
+from scrapy.http.request import Request
+=======
 from scrapy.utils.deprecate import create_deprecated_class
 from scrapy.utils.reqser import request_to_dict, request_from_dict
+>>>>>>> upstream/master
```
Gallaecio, line 16: Ooops! :slightly_smiling_face:

elacuesta: What, do you think this won't work? 😅

PR #4948, opened 2021-01-08
Fix broken link for cryptography installation

manaswinidas:
Fix broken link for cryptography installation in install.rst


elacuesta:
Thanks for the PR, but this has already been solved in the master branch (#4899), and I don't think it grants a backport to the `2.4` branch.

PR #4949, opened 2021-01-08
Fix Warning - Usually first parameter of a method is named 'self'

azenkwed:
Fix Warning - Usually first parameter of a method is named 'self'

Parameter renamed from **self_** to **self**.


azenkwed:
change breaks the code.

PR #4950, opened 2021-01-08
Fix argument type (int -> bool)

azenkwed:
Fix warning: Expected type 'bool', got 'int' instead.

Changing value from **1** to **True** for parameter **doseq** as per the documenation of [urllib.parse.urlencode()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)


elacuesta:
Looks good, thanks. FYI, I merged the master branch because the diff was showing some changes that were already merged in #4940.

elacuesta (approved):

elacuesta (changes requested):
Actually, I just found a few more occurrences, could you update them as well?
```
$ grep -rni "doseq=1" .
./tests/spiders.py:48:        url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=1)}")
./tests/spiders.py:248:            url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=1)}")
./scrapy/commands/bench.py:53:        url = f'{self.baseurl}?{urlencode(qargs, doseq=1)}'
```

azenkwed:
> Looks good, thanks. FYI, I merged the master branch because the diff was showing some changes that were already merged in #4940.

Yes sorry about that, I saw that, but I am not sure why that happened.

kmike:
Hey! I'm merging this PR, as it is an improvement over what we have now, but @elacuesta's comment is valid.

PR #4952, opened 2021-01-11
Sniffpy implementation for Scrapy

woberton:


elacuesta (changes requested):
`Sniffpy` could be useful, I must say I wasn't aware of its existence.
However, `python-requests` is not an asynchronous library, it doesn't play well with Scrapy. Moreover, there's no need to make such request: according to the [documentation](https://requests.readthedocs.io/en/master/user/quickstart/#binary-response-content), `Response.content` contains the `bytes` representation of the response, which is already available in the `body` variable received by `ResponseTypes.from_body`

scrapy/responsetypes.py
```diff
@@ -76,13 +78,16 @@ def from_headers(self, headers):
             )
         if cls is Response and b'Content-Disposition' in headers:
             cls = self.from_content_disposition(headers[b'Content-Disposition'])
-        return cls
+        return cls
```
wRAR, line 81: Stray trailing space. Please review diffs before submitting (or at least after).

scrapy/responsetypes.py
```diff
@@ -76,13 +78,16 @@ def from_headers(self, headers):
             )
         if cls is Response and b'Content-Disposition' in headers:
             cls = self.from_content_disposition(headers[b'Content-Disposition'])
-        return cls
+        return cls 
 
     def from_filename(self, filename):
-        """Return the most appropriate Response class from a file name"""
-        mimetype, encoding = self.mimetypes.guess_type(filename)
+        """Return the most appropriate Response class from a file name"""
```
wRAR, line 84: Yet another stray trailing space.

scrapy/responsetypes.py
```diff
@@ -76,13 +78,16 @@ def from_headers(self, headers):
...
         if cls is Response and b'Content-Disposition' in headers:
             cls = self.from_content_disposition(headers[b'Content-Disposition'])
-        return cls
+        return cls 
 
     def from_filename(self, filename):
-        """Return the most appropriate Response class from a file name"""
-        mimetype, encoding = self.mimetypes.guess_type(filename)
+        """Return the most appropriate Response class from a file name""" 
+        r = requests.get(filename)
```
wRAR, line 85: This is a wrong place to do that in any case. This method is specifically about looking at the file name. I also suspect this doesn't work.

wRAR (changes requested):
You shouldn't add .DS_Store in your changes. Also, you should have tests for your changes.

PR #4953, opened 2021-01-11
adding variable type values support to 'scrapy.FormRequest()'

### LINKED ISSUE #4946: `formdata` of `scrapy.FormRequest()` cannot support variable type values
## Summary

`scrapy.FormRequest()` encodes the `formdata` with the following function, which implies that the values have to be either `str` or `bytes`:

https://github.com/scrapy/scrapy/blob/0a1e2fefab04984a0f3c2b470346fd9f2ffde65b/scrapy/http/request/form.py#L70-L74

However, `urllib.parse.urlencode()` doesn't restrict it:

```python
# Python 3.8.5

urllib.parse.urlencode({'offset': 0, 'limit': 5}) # 'offset=0&limit=5'
```

Neither does `requests`:

```python
# Python 3.8.5

requests.post('https://httpbin.org/post', data={'offset': 0, 'limit': 5}) # <Response [200]>
```

## Motivation

When I am crawling a restapi:

  - `formdata` needs to be converted into `str` everytime when I start yielding `scrapy.FormRequest` object
  - `formdata` needs to be converted back to original type to do arithmetic operation, like the following.

```python
# Python 3.8.5

query = {
    'offset': '0',
    'limit': '5',
    'hasMore': '1',
}

while True:
    yield scrapy.FormRequest('https://httpbin.org/post', formdata=query)

    query['offset'] = str(
        int(query['offset']) + int(query['limit'])
    )

    if not int(query['hasMore']):
        break
```

Maybe I think, we could re-implement the `_urlencode` like this:

```python
def _urlencode(seq, enc):
    return urlencode(seq, doseq=1, encoding=enc) # perserves original features
```

Shall we let the values internally be `str`?

Finally, many thanks to your art-level work and great patience!

akshaysharmajs:
adding support to `scrapy.FormRequest()`  for passing variable type values for `formdata` which is currently only supporting  `str `or `bytes` type.

Fixes #4946


AKSHAYSHARMAJS:
@Gallaecio I have removed the function `_urlencode()`. Do we need that?

Gallaecio:
`_urlencode` may not be necessary anymore.

However, I do think we need to add many tests to cover this change, trying to cover all the scenarios that the previous code handled specifically, and make sure the old and new behavior are the same for those. I’m thinking we need to test things like: bytes, strings, lists, non-utf8 bytes with encoding,

AKSHAYSHARMAJS:
Sorry for the delay, I just started working on the tests. Do I need to test different encoding for form data or the requests? maybe I am wrong here but I can see that bytes, string, lists are already tested for the form data in [test_http_request.py](https://github.com/scrapy/scrapy/blob/master/tests/test_http_request.py)

Gallaecio:
> Do I need to test different encoding for form data or the requests? I can see that bytes, string, lists are already tested for the form data in [test_http_request.py](https://github.com/scrapy/scrapy/blob/master/tests/test_http_request.py)

It seems you are right. I guess the only missing piece is adding a test to cover encoding handling, how defining a non-utf8 encoding in a request affects how `formdata` is transformed. It most likely works, but a test would be great. 

There are some tests in that file with `encoding="latin1"`, but I think none of them involving formdata. Maybe you could copy one of the formdata tests that use bytes, add some non-ASCII characters to the bytes keys and values, and test that they are transformed into the expected values.

AKSHAYSHARMAJS:
I have added tests for non-utf8 bytes and non-ASCII bytes for utf8. I think this covers all and for `encoding="latin1"` all tests were already present. Please check if any test is left, I will add that.

Gallaecio (commented):
One thing though: the purpose of this patch was to add “variable type values support” to `FormRequest`, what about adding tests for that as well?

AKSHAYSHARMAJS:
> One thing though: the purpose of this patch was to add “variable type values support” to `FormRequest`, what about adding tests for that as well?

Yes, this was supposed to be added at first but I forgot 😅 😅 
 Anyway, I have added a test for this.

AKSHAYSHARMAJS:
Why so many tests are failing?

Gallaecio:
https://github.com/scrapy/scrapy/pull/5022 should fix those test breakages.

Gallaecio (approved):

tests/test_http_request.py
```diff
@@ -406,6 +406,30 @@ def test_default_encoding_mixed_data(self):
...
+    def test_custom_nonutf8_bytes_data(self):
+        data = {b'\xbc one': b'two', b'price': b'\xbf 100'}
+        r2 = self.request_class("http://www.example.com", formdata=data)
+        self.assertEqual(r2.method, 'POST')
+        self.assertEqual(r2.encoding, 'utf-8')
+        self.assertQueryEqual(r2.body, b'price=%BF+100&%BC+one=two')
+        self.assertEqual(r2.headers[b'Content-Type'], b'application/x-www-form-urlencoded')
+
+    def test_custom_nonascii_utf8_bytes_data(self):
+        data = {b'\xc2 one': b'two', b'price': b'\xa3 100'}
```
wRAR, line 418: I'm not sure how do these tests differ, as as far as I can see `b'\xbc'`, `b'\xbf'`, `b'\xc2'` and `b'\xa3` are all non-UTF8.

Gallaecio: I think they are all UTF-8, but in any case your point remains.

wRAR: All UTF-8 single-byte characters are lower than \x80 :) `b'\xc2\xa3'` is a single UTF-8 character.

Gallaecio: Sorry, I confused UTF-8 with Unicode :facepalm:

AdrianAtZyte:
Closing after the move to [form2request](https://form2request.readthedocs.io/en/latest/usage.html) as a [replacement for `FormRequest`](https://github.com/scrapy/scrapy/pull/6438).

If you believe this issue still impacts `form2request` and should be addressed, please open an issue [upstream](https://github.com/scrapy/form2request/issues) with a link to this issue for reference.

PR #4954, opened 2021-01-11
Add content-based image filtering example

smellslikeml:
Adding an example under the [Custom Images pipeline example](https://docs.scrapy.org/en/latest/topics/media-pipeline.html#custom-images-pipeline-example) in docs. This example showcases an image classifier to filter images before storing.


Gallaecio:
I like the idea of including an example that covers how to filter images based on content. However, I would aim for a minimalistic example, as to not distract the reader’s attention with code that is not required for what the example is meant to showcase.

To that effect, I would remove the implementations of ``__init__`` and ``check_image``, and reduce ``get_images`` to around 5 lines.

smellslikeml:
@Gallaecio  Thanks for the feedback! We reduced the example to the necessary components.
We expanded some lines in the ```get_images()``` method for readability. 
We also kept the implementation of the ```__init__()``` method to instantiate a model.

wRAR (approved):

PR #4955, opened 2021-01-12
Remove redundant int/float conversions needed in Python2

eumiro:
Remove  redundant conversions to `int` or `float` that were needed in Python2 but are not needed in Python3.


Gallaecio (approved):

wRAR:
Thanks!

wRAR:
I just checked the docs, at least for `AUTOTHROTTLE_MAX_DELAY` they say the default is `60.0` so I guess this should be changed too?

eumiro:
> `AUTOTHROTTLE_MAX_DELAY`

There are not only literal numbers, but also functions that define the type of the variable, such as: https://github.com/scrapy/scrapy/blob/e42d82526a1a519ff1305b189acde064a40fbd8d/scrapy/extensions/throttle.py#L35

wRAR:
`getfloat` is fine as the user can set it to a float, I just mean the difference between the actual default and the documented one.

Gallaecio:
If we remove `.0` from the documentation, we need to make sure we make it clear that floating-point numbers are allowed in those settings. I’m OK with keeping the `.0` in the documentation only just for this purpose, though. In fact, I think it may be better to leave the `.0` in certain parts of the code as well, just for readability; at least in the file that defines the default value of settings.

eumiro:
> If we remove `.0` from the documentation, we need to make sure we make it clear that floating-point numbers are allowed in those settings. I’m OK with keeping the `.0` in the documentation only just for this purpose, though. In fact, I think it may be better to leave the `.0` in certain parts of the code as well, just for readability; at least in the file that defines the default value of settings.

Please feel free to revert those of my changes you find appropriate and then squash the commits.

Gallaecio:
I guess we could simply revert the changes in `default_settings.py` to keep those values in sync with the documentation, for consistency.

I also though about the `interval` parameter of `LogStats.__init__`, but [it’s not visible in the documentation](https://docs.scrapy.org/en/latest/topics/extensions.html#scrapy.extensions.logstats.LogStats), so I think the change is fine.

PR #4957, opened 2021-01-15
 Fix engine close error bug 

CheungChan:
Fix  when the MemoryUsage exension shutdown, it raises "RuntimeError: Engine not running"


Gallaecio:
I would like to first discard that the issue is in the extension, as engine changes like this can be risky, and break other things.

Could you provide some simple steps to reproduce the issue first?

PR #4958, opened 2021-01-16
Match the styles of empty container initializations

HsuanTingLu:
We use `set()` to initialize an empty set, and I'm wondering if turning `{}` into `dict()` looks more consistent.

The codebase does have this, in `scrapy/utils/conf.py` ([reference](https://github.com/scrapy/scrapy/blob/63becd1bc89395750b39139e2114193607f3ca61/scrapy/utils/conf.py#L124))

There are tens of those scattered in the codebase, some of which probably should keep `{}` though.
I propose we replace those that appear in initializers and class definitions, ignore those in function returns.
I can push the other commits if you think this change is a good idea.

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -73,7 +73,7 @@ class Downloader:
     def __init__(self, crawler):
         self.settings = crawler.settings
         self.signals = crawler.signals
-        self.slots = {}
+        self.slots = dict()
         self.active = set()
         self.handlers = DownloadHandlers(crawler)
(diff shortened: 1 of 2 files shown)


kmike:
I'm not sure it would be for good to make this change. It'd mean that we should be enforcing in some way the same style for new contributors, and for existing PRs, which seems unnecessary. 

By the "consistency" logic, there shouldn't be `{}` in Python at all, because there is no a special syntax for creating empty sets.

HsuanTingLu:
> By the "consistency" logic, there shouldn't be `{}` in Python at all, because there is no a special syntax for creating empty sets.

I see your point.

PR #4966, opened 2021-01-28
Fix FEED_URI_PARAMS: custom params throws KeyError

### LINKED ISSUE #4962: FEED_URI_PARAMS: custom params throws KeyError
### Description

With Scrapy version 2.4.1, any custom feed uri params (FEED_URI_PARAMS) throws KeyError. 

### Steps to Reproduce

1. on myproject/settings.py
FEED_URI_PARAMS = 'myproject.utils.uri_params'

2. on myproject/utils.py
def uri_params(params, spider):
    return {**params, 'spider_name': spider.name}

3. scrapy crawl <spider_name> -o "%(spider_name)s.jl"

**Expected behavior:** Should export the items to a jsonline file having the spider name.

**Actual behavior:**
ERROR: Error caught on signal handler: <bound method FeedExporter.open_spider of <scrapy.extensions.feedexport.FeedExporter object at 0x7fc68cb20bb0>>
  File "/PROJECT_ROOT/.venv/lib/python3.8/site-packages/scrapy/utils/defer.py", line 157, in maybeDeferred_coro
    result = f(*args, **kw)
  File "/PROJECT_ROOT/.venv/lib/python3.8/site-packages/pydispatch/robustapply.py", line 55, in robustApply
  File "/PROJECT_ROOT/.venv/lib/python3.8/site-packages/scrapy/extensions/feedexport.py", line 294, in open_spider
    uri=uri % uri_params,
KeyError: 'spider_name'

**Reproduces how often:** Everytime

### Versions
Scrapy       : 2.4.1
lxml         : 4.6.1.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 20.3.0
Python       : 3.8.5 (default, Jul 28 2020, 12:59:40) - [GCC 9.3.0]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1h  22 Sep 2020)
cryptography : 3.2.1
Platform     : Linux-4.19.104-microsoft-standard-x86_64-with-glibc2.29

### Additional context

on line 480 in /scrapy/extensions/feedexport.py,
**returns params** should be changed to **return uripar_function(params, spider)**

inspurwusixuan:
Fix when input any custom feed URI params (FEED_URI_PARAMS), it raises KeyError.

Fixes #4962


inspurwusixuan:
Hi @Gallaecio, I'm trying to figure out why some of the tests failed. Any suggestions? 

Thanks in advance!

scrapy/extensions/feedexport.py
```diff
@@ -485,4 +485,4 @@ def _get_uri_params(self, spider, uri_params, slot=None):
         params['batch_id'] = slot.batch_id + 1 if slot is not None else 1
         uripar_function = load_object(uri_params) if uri_params else lambda x, y: None
         uripar_function(params, spider)
-        return params
+        return uripar_function(params, spider)
```
Gallaecio, line 488: Maybe it is not as straightforward as I thought. Also, I think we may need to be careful to ensure backward compatibility.

I wonder what would happen if we only return the result of `uripar_function` if it is not `None`:

```python
        new_params = uripar_function(params, spider)
        return new_params if new_params is not None else params
```

So, if someone wrote a `uripar` function that returns nothing and instead simply edits the input params, this code should work as well.

Assuming this code works, then we probably just need to extend it to log a deprecation warning if `uripar_function` returns `None`, and we should update our tests to cover both use cases.

inspurwusixuan: Thanks for the reply! I will try this later and see how it goes.

inspurwusixuan: It seems like `return new_params if new_params is not None else params` passed all the checks. Thanks for the advice! Let me know if there is anything I can help. :)

Gallaecio:
So that means that, even though the documentation clearly states that the function should _return_ the resulting dictionary, some tests instead only modify the one they receive. And, chances are, some people out there do the same.

Because we do not want to break the code for those people, but we do want to eventually remove this workaround of checking if the return value is `None`, the next step would be to add code after the function call but before the final `return` statement that logs a deprecation warning if the result was `None`.

If you search the code base for `ScrapyDeprecationWarning` you should find many examples on how to log such a warning. In this case, I would avoid the `level` parameter altogether, since this code is not directly called by the user. But the warning message should mention the setting that the user may have used to define the function (there is a global setting and a `FEEDS` keyword), and explain that the function they configured is modifying the received parameters dictionary in-place, which is deprecated, and they should instead return the new dictionary.

Let me know if you have any doubt.

inspurwusixuan:
@Gallaecio Sorry, I was a bit confused about the warning message part. 

My understanding of the current situation is that if a user does not provide a `uri_params` function in the `FEED` options, then `uripar_function` will return None and the `_get_uri_params` function will return the `params`.

On the contrary, if a user does provide a `uri_params` function. Then the `_get_uri_params` function will directly return a new dictionary object `new_params` which is different from the original `params` dictionary. 

So it seems to me like returning the `params` when a user does not provide a `uri_params` function with`new_params` is None is fine. Then why we need a warning message for such a case?

Please let me know if I got anything wrong! Thanks! :)

Gallaecio:
My understanding is that at the moment a user could define a `uri_params` function that instead of returning `params` simply modified the `params` parameter in place. Even though this is not documented, it would work, and users may be relying on it.

Since we want user functions to always return `params`, if a user is currently not returning `params`, we want to warn them about it. In the future, we will drop support for functions that do not return `params`.

You bring up an important point that I had not realized: the default function, when users do not define one, returns `None`. I think we should change that, replace `lambda x, y: None` with `lambda x, y: x`.

inspurwusixuan:
That makes sense! Thanks for the explanations! @Gallaecio 

So after changing the `lambda x, y: None` to `lambda x, y: x`, the `new_params` will only return `None` when the user is modifying the dictionary in-place and hence not returning any new dictionary object. In this case, we give them a warning something like below. 

```python
uripar_function = load_object(uri_params) if uri_params else lambda x, y: x
new_params = uripar_function(params, spider)
if not new_params:
    warnings.warn(
        'Modifying the params dictionary in-place in the customized `FEED_URI_PARAMS` setting is deprecated. '
        'A new dictionary params is expected to return',
        category=ScrapyDeprecationWarning
    )
return new_params if new_params is not None else params
```

The wording of the warning message may not be very precise. Let me know if you have any thoughts.

Gallaecio:
I would use `if new_params is None:` , to avoid warning if they return and empty dictionary.

The warning looks quite good to me, I just have a few nitpicks. What about this?

```python
    warnings.warn(
        'Modifying the params dictionary in-place in the function defined in '
        'the FEED_URI_PARAMS setting or in the uri_params key of the FEEDS '
        'setting is deprecated. The function must return a new dictionary '
        'instead.',
        category=ScrapyDeprecationWarning
    )
```

inspurwusixuan:
@Gallaecio all the checks have passed! Thanks for all the help! :)

Gallaecio:
There’s 1 more things to do here before we can merge the change: cover the change with tests.

I think we need:
- A test that checks that the warning is indeed logged when a custom function does not return anything (which is the same as returning `None`)
- A test that checks that that a custom function that does return a dictionary works as expected without logging a warning. Hopefully there is already such a test, and we simply need to add a check that no warning is issued.
- A test that checks that, if no function is defined, the warning is not issued. It may also be possible to simply extend an existing test with a no-warning check, but unless a test exists specifically to test the default behavior, I would create a new test, even if it’s very similar to an existing one.

For even better coverage, when defining a custom function we could test both ways, with `FEED_URI_PARAMS` and with `FEEDS[*].uri_params`. But I would consider that an extra, 1 of those approaches should be enough to cover the scope of this change.

Writing tests can be a pain, but in this case it should be relatively easy to do based on those tests that failed with the first commit. Please let me know if I can help you in any way.

inspurwusixuan:
I did some investigation on the test cases we have in the `tests` folder. But based on my search, it seems like there is no test associated with the `FEED_URI_PARAMS` setting or the function that calls `_get_uri_params` I modified. 

And I'm not sure how to simulate the whole workflow in the test case. The way I reproduced the error right now is to simply follow the [Scrapy Tutorial](https://docs.scrapy.org/en/latest/intro/tutorial.html) to create a tutorial project and modify the `utils.py` with my `uri_params` function based on the `FEED_URI_PARAMS` documentation [here](https://docs.scrapy.org/en/latest/topics/feed-exports.html#feed-uri-params). So I'm confused about how to add the corresponding `uri_params` in the `utils.py` and initiate the process in the test case.

Could you give me an example of this? Thanks! @Gallaecio

Gallaecio:
Sure. Let me try and create the first test, then maybe you can create the remaining ones based on it.

Gallaecio:
I ended up writing them all, just to be sure I was getting them right :slightly_smiling_face:

inspurwusixuan:
> I ended up writing them all, just to be sure I was getting them right 🙂

Thank you!! It looks good to me! is there any action required from my end?

Gallaecio (approved):
Nothing else to do here other than waiting for other maintainer to review and approve the changes. Thanks!

scrapy/extensions/feedexport.py
```diff
@@ -483,6 +483,14 @@ def _get_uri_params(self, spider, uri_params, slot=None):
         params['time'] = utc_now.replace(microsecond=0).isoformat().replace(':', '-')
         params['batch_time'] = utc_now.isoformat().replace(':', '-')
         params['batch_id'] = slot.batch_id + 1 if slot is not None else 1
-        uripar_function = load_object(uri_params) if uri_params else lambda x, y: None
-        uripar_function(params, spider)
-        return params
+        uripar_function = load_object(uri_params) if uri_params else lambda x, y: x
+        new_params = uripar_function(params, spider)
```
Gallaecio, line 487: I’m just leaving a note that it is technically possible for this change to break code if the user was modifying `params` but returning a dict with a different content from that of `params`. I don’t think this should stop this change, but it may be worth mentioning in the release notes.

ned9:
Would be great to get this merged. Just spend a while tracking down why `uri_params` function wasn't working as described by the documentation.

scrapy/extensions/feedexport.py
elacuesta: What if we checked this? Making a shallow copy of the original `params` dict and making sure it didn't change after calling the `uripar_function`. Something like:

```python
original_params = params.copy()
new_params = uripar_function(params, spider)
if new_params is None or original_params != params:
   ...
```

Gallaecio:
@inspurwusixuan Do you have time for a [final change](https://github.com/scrapy/scrapy/pull/4966#discussion_r669587622)?

scrapy/extensions/feedexport.py
Gallaecio: I think it makes sense, though then we must document that as well, before people start using `.pop()` on the original to build the new one and things stop behaving as expected. Once we remove support for editing the dictionary in place, we can start allowing people to do that.

elacuesta (approved):

scrapy/extensions/feedexport.py
```diff
@@ -524,17 +524,31 @@ def build_instance(builder, *preargs):
...
+        if new_params is None or original_params != params:
+            warnings.warn(
+                'Modifying the params dictionary in-place in the function defined in '
+                'the FEED_URI_PARAMS setting or in the uri_params key of the FEEDS '
+                'setting is deprecated. The function must return a new dictionary '
+                'instead.',
+                category=ScrapyDeprecationWarning
+            )
+        return new_params if new_params is not None else params
```
elacuesta, line 551: @Gallaecio Given this `if`, should we remove the `backward-incompatible` tag?

scrapy/extensions/feedexport.py
```diff
@@ -524,17 +524,31 @@ def build_instance(builder, *preargs):
             raise TypeError(f"{feedcls.__qualname__}.{method_name} returned None")
         return instance
 
-    def _get_uri_params(self, spider, uri_params, slot=None):
+    def _get_uri_params(
+        self,
+        spider: Spider,
+        uri_params_function: Optional[Union[str, Callable[[dict, Spider], dict]]],
```
wRAR, line 530: `Union[str, Callable[[dict, Spider], dict], None]`?

wRAR (approved):

scrapy/extensions/feedexport.py
Gallaecio: I would think `Optional[Union[…]]` is more Pythonic than `Union[…, None]`, even if longer. I don’t really know, though.

scrapy/extensions/feedexport.py
Gallaecio: No. https://github.com/scrapy/scrapy/pull/4966#discussion_r573517714 still applies.

I’m not sure we can make this backward-compatible for every single case, e.g. if someone is doing something unexpected (and pointless) like returning an empty dict, or returning a dict with content A but having their code actually rely on content B from `params`.

PR #4972, opened 2021-02-02
response_httprepr memory issue fixed

### LINKED ISSUE #4964: response_httprepr in DownloaderStats middleware causing scrapy to make unnecessary memory allocation.
### Description

Usage of [`scrapy.utils.response.response_httprepr` ](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/utils/response.py#L45)inside[ `DownloaderStats`](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/downloadermiddlewares/stats.py) middleware causing application to make unnecessary memory allocation.
`response_httprepr` used only one time - to calculate response sizes for `downloader/response_bytes` stats in that middleware.

In current implementation `response_httprepr` return `bytes` (immutable type) - in order to calculate `downloader/response_bytes` application will additionally allocate nearly the same memory amount as for original response (only for calculating `len` inside middleware).
https://github.com/scrapy/scrapy/blob/26836c4e1ae9588ee173c5977fc6611364ca7cc7/scrapy/utils/response.py#L45-L60

### Steps to Reproduce
In order to demonstrate influence of this i made this spider:
<details>
  <summary>spider code</summary>

```python
import sys
from importlib import import_module
import scrapy


class MemoryHttpreprSpider(scrapy.Spider):
    name = 'memory_httprepr'
    custom_settings = {
        'DOWNLOADER_MIDDLEWARES':{
            'scrapy.downloadermiddlewares.stats.DownloaderStats': None
        }
    }
# the same as in MemoryUsage extension:
    def get_virtual_size(self):
        size = self.resource.getrusage(self.resource.RUSAGE_SELF).ru_maxrss
        if sys.platform != 'darwin':
            # on macOS ru_maxrss is in bytes, on Linux it is in KB
            size *= 1024
        return size

    def start_requests(self):
        try:
            self.resource = import_module('resource')
        except ImportError:
            pass
        self.logger.info(f"used memory on start: {str(self.get_virtual_size())}")
        yield scrapy.Request(url='https://speed.hetzner.de/100MB.bin', callback=self.parse)
        #yield scrapy.Request(url='http://quotes.toscrape.com', callback=self.parse)

        self.logger.info(f"used memory after downloading response: {str(self.get_virtual_size())}")
```

</details>
It include:

- usage of `get_virtual_size` method - directly the same as on [`MemoryUsage`](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/extensions/memusage.py) extension
- url with 100Mb binary response https://speed.hetzner.de/100MB.bin
- disabled `DownloaderStats` middleware that uses `response_httprepr` and `request_httprepr`
- usual html url http://quotes.toscrape.com (for comparison)

| log output | 100Mb default settings | 100Mb no downloader stats | quotes default |
| --- | --- | --- | --- |
| `[memory_httprepr] used memory on start:` | `61 587 456` | `61 521 920` | `61 558 784` |
| `[memory_httprepr] used memory after downloading response:` | `375 910 400` | `271 179 776` | `61 558 784` |

### Versions

`[scrapy.utils.log] Scrapy 2.4.0 started (bot: httprepr)`
`[scrapy.utils.log] Versions: lxml 4.5.2.0, libxml2 2.9.10, cssselect 1.1.0, parsel 1.6.0, w3lib 1.22.0, Twisted 20.3.0, Python 3.8.2 (default, Apr 23 2020, 14:32:57) - [GCC 8.3.0], pyOpenSSL 19.1.0 (OpenSSL 1.1.1h 22 Sep 2020), cryptography 3.1.1, Platform Linux-4.15.0-76-generic-x86_64-with-glibc2.2.5`

GeorgeA92:
This code change aimed to fix memory issue #4964

Fixes #4964


wRAR:
"/home/runner/work/scrapy/scrapy/scrapy/downloadermiddlewares/stats.py:3:1: F401 'scrapy.utils.response.response_httprepr' imported but unused"

scrapy/downloadermiddlewares/stats.py
```diff
@@ -24,7 +23,7 @@ def process_request(self, request, spider):
     def process_response(self, request, response, spider):
         self.stats.inc_value('downloader/response_count', spider=spider)
         self.stats.inc_value(f'downloader/response_status_count/{response.status}', spider=spider)
-        reslen = len(response_httprepr(response))
+        reslen = len(response.body)
```
Gallaecio, line 26: This seems to no longer take into account headers and such. I think the right solution is to create an alternative to `response_httprepr` that, without creating a new bytes object, providers the length of all parts involved.

GeorgeA92: @Gallaecio 
`headers.to_string` used inside `response_httprepr` also duplicates headers in memory just like response_httprepr duplicates response.body
https://github.com/scrapy/scrapy/blob/26836c4e1ae9588ee173c5977fc6611364ca7cc7/scrapy/utils/response.py#L57-L60

https://github.com/scrapy/scrapy/blob/26836c4e1ae9588ee173c5977fc6611364ca7cc7/scrapy/http/headers.py#L75-L76

I think that each base http structure (request, response, headers, cookies etc.) needs to have it's own `get_size` method that calculate it's own size without duplicating it in memory as  raw `str` representation.

Gallaecio: I think we basically need to replace all those functions that generate strings with functions that calculate sizes, yes.

omidhzr:
Hi @GeorgeA92!

are you still working on this issue? and btw do you need some help? we would like to contribute as a group :)

GeorgeA92:
Added function for calculating headers size.

Implementation of `get_header_size`  based on [`w3lib.http.headers_dict_to_raw`](https://github.com/scrapy/w3lib/blob/v1.22.0/w3lib/http.py#L69-L78) function used in `response_httprepr`


For url `http://quotes.toscrape.com/page/1/`
I received:
`len(response_httprepr(response))` - 2204 bytes
`len(response.body)` - 2039
`len(response.body) + get_header_size(response.headers)` - 2183

In this implementation DownloaderStats middleware instead of memory intensive `len(response_httprepr(response))` will use `len(response.body) + get_header_size(response.headers)` for calculating response size used in `downloader/response_bytes` stats parameter

Gallaecio:
So it’s still not the same size. Could you work out an implementation that returns the same as `response_httprepr`, taking into account additional data that `response_httprepr` takes into account (e.g. status code, line breaks)?

GeorgeA92:
@Gallaecio Added counting [status code](https://github.com/scrapy/scrapy/blob/e63188cbf753d560e43d8489c821bd6eb9fe54e9/scrapy/utils/response.py#L58-L64) related part and remaining linebreaks.

Gallaecio (commented):
Can we now have a test that compares the result to that of `len(response_httprepr(response))`, and verifies it matches for some corner cases? (empty/non-empty body, 0-1-2 headers)

scrapy/downloadermiddlewares/stats.py
```diff
@@ -1,7 +1,32 @@
...
+
+def get_status_size(response_status):
+    values = [
+        9,
+        len(to_bytes(str(response_status))),
+        1,
+        len(to_bytes(http.RESPONSES.get(response_status, b''))),
+        2,
+    ]
+    return sum(values)
```
elacuesta, line 29:
```suggestion
    return (
        9  # b"HTTP/1.1 "
        + len(to_bytes(str(response_status)))
        + 1  # " "
        + len(to_bytes(http.RESPONSES.get(response_status, b"")))
        + 2  # b"\r\n"
    )
```
Adding comments for future reference. Also, since we're optimizing memory usage, building a list (and assigning it to a variable) has an unnecessary overhead (even more so than a tuple, because of their mutability).
Going even further, the integers could be consolidated, but that might be counter-productive in terms of readability.

GeorgeA92: @elacuesta related codelines changed in https://github.com/GeorgeA92/scrapy/commit/d7b3cbfbaab7c040d306bf9be0a64c2e144accc2

GeorgeA92:
@Gallaecio 
> Can we now have a test that compares the result to that of len(response_httprepr(response)), and verifies it matches for some corner cases? (empty/non-empty body, 0-1-2 headers)

Ready -> https://github.com/GeorgeA92/scrapy/commit/3805babc18a2dee92a6724801f6fb760ee2346a2

scrapy/downloadermiddlewares/stats.py
```diff
@@ -1,7 +1,26 @@
...
+            for v in value:
+                size += len(b": ") + len(key) + len(v)
+    return size + len(b'\r\n') * (len(headers.keys()) - 1)
+
+
+def get_status_size(response_status):
+    return sum((
+        9,  # b"HTTP/1.1 "
+        len(to_bytes(str(response_status))),
+        1,  # " "
```
Gallaecio, line 21:
```suggestion
        13,  # b"HTTP/1.1 <100-599> "
```

scrapy/downloadermiddlewares/stats.py
```diff
@@ -24,7 +43,7 @@ def process_request(self, request, spider):
     def process_response(self, request, response, spider):
         self.stats.inc_value('downloader/response_count', spider=spider)
         self.stats.inc_value(f'downloader/response_status_count/{response.status}', spider=spider)
-        reslen = len(response_httprepr(response))
+        reslen = len(response.body) + get_header_size(response.headers) + get_status_size(response.status) + 4
```
Gallaecio, line 46: It looks like this was the only usage of `response_httprepr`. May be deprecate the function? (i.e. log a deprecation warning when it is used, lookup usage of `ScrapyDeprecation` in code).

Also, maybe worth commenting that 4 stands for the 2 `\r\n` between those 3 sections.

tests/test_downloadermiddleware_stats.py
```diff
@@ -37,6 +39,17 @@ def test_process_response(self):
         self.mw.process_response(self.req, self.res, self.spider)
         self.assertStatsEqual('downloader/response_count', 1)
 
+    def test_response_len(self):
+        body = (b'', b'not_empty')  # empty/notempty body
+        headers = ({}, {'lang': 'en'}, {'lang': 'en', 'User-Agent': 'scrapy'})  # 0 headers, 1h and 2h
+        test_responses = [
+            Response(url='scrapytest.org', status=200, body=r[0], headers=r[1])
+            for r in product(body, headers)]  # form test responses with all combinations of body/headers
```
Gallaecio, line 47: :lipstick: 
```suggestion
        test_responses = [
            Response(
                url='https://scrapy.org/',
                status=200,
                body=body,
                headers=headers,
            )
            for body, headers in product(body, headers)
        ]
```

scrapy/downloadermiddlewares/stats.py
GeorgeA92: function optimized https://github.com/scrapy/scrapy/pull/4972/commits/7a8c2541f2d4d5fc1994c6f62f3a5c2050385593

scrapy/downloadermiddlewares/stats.py
GeorgeA92: explanation (comment) added https://github.com/scrapy/scrapy/pull/4972/commits/915bd9cca933f3050250186b0479c9c71e475751
`response_httprepr` deprecated https://github.com/scrapy/scrapy/pull/4972/commits/f0a4d7857c282bf52829455d3300f3b28e366cdb

tests/test_downloadermiddleware_stats.py
GeorgeA92: @Gallaecio  fixed https://github.com/scrapy/scrapy/pull/4972/commits/fbdc6a9a25b3b457d509580349e286e5c5d5771e

Gallaecio: The used URL is still invalid (no scheme), and points to a domain we do not own which, based on recent experience, could eventually be a NSFW website.

Gallaecio (approved):
Great job!

I would still like to address https://github.com/scrapy/scrapy/pull/4972#discussion_r720636651, but since this is a test it is not that important.

tests/test_downloadermiddleware_stats.py
GeorgeA92: @Gallaecio I think that in this specific case - it is not an issue. This test creates `Response` objects from scratch without sending requests.

wRAR (approved):

scrapy/utils/response.py
```diff
@@ -50,6 +51,7 @@ def response_status_message(status: Union[bytes, float, int, str]) -> str:
     return f'{status_int} {to_unicode(message)}'
 
 
+@deprecated
 def response_httprepr(response: Response) -> bytes:
```
mario-stylesage, line 55: @wRAR FYI this function is deprecated here (and removed in 2.11) but never mentioned in the [release logs](https://docs.scrapy.org/en/2.10/news.html) (nor its removal)

Gallaecio: Great catch, I’ve logged https://github.com/scrapy/scrapy/issues/6207.

PR #4974, opened 2021-02-03
Move lists closer to their introducing paragraph

Gallaecio:
This has been bothering me for a while.

I’ve not reviewed the effect in the whole documentation, but I’ve had a look at pages from the main page to the settings page, which is about half the documentation, and it looks good to me.

Before: 
Screenshot_2021-02-03 Scrapy 2 4 documentation — Scrapy 2 4 1 documentation(1)

After:
Screenshot_2021-02-03 Scrapy 2 4 documentation — Scrapy 2 4 1 documentation

Changes to reStructuredText files are so that lists which were put within a blockquote because of the empty space that precedes their `*` are now regular lists.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,10 @@
+/* Move lists closer to their introducing paragraph */
+.rst-content .section ol p, .rst-content .section ul p {
+    margin-bottom: 0px;
+}
+.rst-content p + ol, .rst-content p + ul {
+    margin-top: -18px; /* Compensates margin-top: 24px of p  */
+}
+.rst-content dl p + ol, .rst-content dl p + ul {
+    margin-top: -6px; /* Compensates margin-top: 12px of p  */
+}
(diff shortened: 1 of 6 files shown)


kmike (approved):
Looks good to me 👍

kmike:
Hey @Gallaecio! The menu on the left doesn't look right on readthedocs: 

image

Do you know if it can be caused by this change?

Gallaecio:
I assume it was caused by this change, but I did render the latest changes locally as I was working on the release notes, so I suspect the issue is not reproducible locally.

I’ll look into it when I get a chance.

Gallaecio:
I’ve rebuild 2.4.0 and [the same happens in ReadTheDocs](https://scrapy-gallaecio.readthedocs.io/en/2.4.0/).

Gallaecio:
It turns out it was not these changes: https://github.com/readthedocs/readthedocs.org/issues/8070

PR #4872, opened 2020-11-05
Call asyncio.get_event_loop when installing the asyncio reactor

elacuesta:
Fixes #4855. See https://github.com/scrapy/scrapy/issues/4855#issuecomment-721964440 for the rationale of this change.

- [x] Code changes
- [x] Tests
- [x] Docs

In the docs, I intentionally only mentioned `set_event_loop` and omitted the call to `get_event_loop` when the default event loop is used. I've grown more convinced that this is "harmless", in the sense that `get_event_loop` will call `set_event_loop` only if it hasn't been called, and if that's the case, the user most likely does not care about the loop being used.

Issue #4841, opened 2020-10-10
tests/test_downloadermiddleware.py bug with Twisted 17.9.0 


wRAR:
As found in #4814 if you manage to install Twisted 17.9.0 (which our tests currently don't), the `tests/test_downloadermiddleware.py::MiddlewareUsingCoro::test_asyncdef` test hangs, probably because of http://twistedmatrix.com/trac/ticket/9390

Gallaecio:
Unless we can come up with a workaround that gets our tests to work with the minimum supported version of Twisted, what should we do? Should we document the minimum version of Twisted used by asyncio tests as a requirement for asyncio support, or don’t mention it in the documentatio?

wRAR:
First I need to make sure the problem is only in the tests.

wRAR:
So, the Twisted change makes the reactor use `get_event_loop()` instead of `new_event_loop()` if the loop is not passed explicitly. pytest-twisted doesn't pass it explicitly so the change affects our tests. But Scrapy now (since #4414) uses `new_event_loop()` explicitly, so I don't understand why non-test Scrapy usage is still fine with old Twisted.

elacuesta:
I am able to successfully run these tests with Twisted==17.9.0 after #4872, should we close this?

wRAR:
I think so.

PR #4978, opened 2021-02-05
Support for async callbacks

wRAR:
This PR adds support for defining `async def parse_*` and using both `yield` and `await` inside, with or without the asyncio reactor. There is one caveat: if the asyncio reactor is enabled, you cannot await on Deferreds directly and need to wrap them. I've added this to docs.

This change was tested on two real-world projects, for one of them I also replaced using `inline_requests` to get HTTP responses inside a callback with alternatively `treq` or `aiohttp`. I didn't test the performance, but as far as I can see the existing non-async callbacks shouldn't have any significant code path changes so there should be no regressions and if there are any perf problems with async callbacks we can solve that later.

This doesn't require or support async def start_requests.

My main concern with this is large and non-intuitive `scrapy.utils.defer._AsyncCooperatorAdapter`. I would like to replace it with something better if that's possible but for now it seems to work correctly. I've tried to document it as comprehensively as possible.

I've split it in several commits with different features in case this help with review for someone.


wRAR:
Not sure why some related tests fail on 3.6 asyncio-pinned, I cannot reproduce this locally at this time.

wRAR:
Now I can reproduce them on a freshly built 3.6.12 while it worked on my old 3.6.9, both from pyenv.

wRAR:
It's Twisted 17.9.0-specific again.

wRAR:
pytest-twisted installs the asyncio reactor with eventloop=None, so on Twisted 17.9.0 uses `new_event_loop()`.  We added a workaround for this in Scrapy itself, in #4841 / #4872, but it's much harder to do for pytest-twisted, unless they add some support for that (or we drop 17.9.0). It's probably possible to install the reactor in `conftest.py` though.

Gallaecio (commented):
Thanks for the separate commits, it definitely made review much more approachable.

I think it may be best to add tests for `parallel_async` that ensure complete code coverage for `_AsyncCooperatorAdapter`, specially given its complexity (I won’t claim to fully understand it).

I love the documentation entry on how to handle waiting on deferreds now! When releasing 2.5, we should probably remember to add a brief mention about this to the release notes as a backward-incompatible change (on an experimental feature, but still), linking to the section for further details.

docs/topics/spider-middleware.rst
```diff
@@ -98,20 +98,25 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
...
+
         This method is called with the results returned from the Spider, after
         it has processed the response.
 
-        :meth:`process_spider_output` must return an iterable of
-        :class:`~scrapy.http.Request` objects and :ref:`item object
-        <topics-items>`.
+        :meth:`process_spider_output` must return an iterable (normal or
+        asynchronous) of :class:`~scrapy.http.Request` objects and 
+        :ref:`item objects <topics-items>`.
```
Gallaecio, line 110: I think we may need to edit this further to make sure that we do not mislead against what the new note in `coroutines.rst` says.

Maybe a similar note?

```
        :meth:`process_spider_output` must return an iterable of
        :class:`~scrapy.http.Request` objects and 
        :ref:`item objects <topics-items>`.
        
        .. note:: When defined as a :ref:`coroutine <async>`, this method needs 
                  to be an async generator, not just return an iterable.
```

Given a clarification like this, I think we can leave the text below for `:type result:` as is.

Gallaecio:
`_AsyncCooperatorAdapter` coverage is quite good, but [it does not seem complete yet](https://app.codecov.io/gh/scrapy/scrapy/compare/4978/diff#D2L139). 2 out of of 3 seem like error scenarios that should not usually happen, but I wonder why `__iter__` is never called, maybe it is not needed?

I see there are also [a few uncovered paths](https://app.codecov.io/gh/scrapy/scrapy/compare/4978/diff#D7L36) in the new `_process_iterable_universal` decorator. Similarly, all seem like error scenarios which should not usually happen.

If testing those error scenarios is complicated, I’m OK with not covering them for now. But if it’s trivial it would be best to go for it.

wRAR:
Yeah, I checked the coverage for `_AsyncCooperatorAdapter` when writing tests, these two exceptions are just checks for things that should never happen so I don't know how to test them, initially I had much more asserts in that class but removed most of them, I kept these two because I was afraid that these things won't be noticed without an explicit check. On the other hand, `d.callback()` already checks for the same thing which makes 1.5 of these 2 checks redundant. 

As for `def __iter__`, that's just the iterable protocol, but as `Cooperator.coiterate` says it takes an iterator we can drop it.

I'll look at `_process_iterable_universal` coverage.

Gallaecio:
:tada: 

Who’s the next brave reviewer who dares have a look at this patch?

docs/topics/asyncio.rst
```diff
@@ -39,5 +39,28 @@ You can also use custom asyncio event loops with the asyncio reactor. Set the
...
+If you want to write universal code that works on any reactors,
+you should use ``maybe_deferred_to_future`` on all Deferreds::
+
+    from scrapy.utils.defer import maybe_deferred_to_future
+
+    class MySpider(Spider):
+        # ...
+        async def parse_with_deferred(self, response):
+            additional_response = await maybe_deferred_to_future(treq.get('https://additional.url'))
+            additional_data = await maybe_deferred_to_future(treq.content(additional_response))
```
pawelmhm, line 65: This looks cool, nice to have replacement for inline requests, but will I have to always make 2 deferreds and 2 await to get response? In my project when I use inline_requests I need response 99% of times is there some way to reduce API to

```python
additional_response = await maybe_deferred_to_future(treq.get('https://additional.url'))
```
instead of one await for request and one await for reading response?

scrapy/spidermiddlewares/urllength.py
```diff
@@ -37,4 +38,10 @@ def _filter(request):
...
                 return True
 
-        return (r for r in result or () if _filter(r))
+        @_process_iterable_universal
+        async def process(result):
+            async for r in result or ():
+                if _filter(r):
+                    yield r
+
+        return process(result)
```
pawelmhm, line 47: I see you add this decorator to several middlewares, what if I have my own middlewares without process_iterable_universal? Will it work fine? Or will I have to add things to my middleware?

pawelmhm:
One other problem I see is that treq is not visible in logs, this seems like a drawback compared with inline requests. Is there some way to integrate approach from this PR with Scrapy logging?

pawelmhm:
Ideally I would really love to see something like this in Scrapy someday

```python
async def parse(self, response):
    res = await Request(some_url)
   # response.body here
```
so native integration, no treq, so that everything from my spider settings like proxy settings, headers all is applied to Request, ang goes smoothly, integrated with Scrapy like a charm.

scrapy/spidermiddlewares/urllength.py
pawelmhm: UPDATE:

I tested it in one of my projects where I have some middleware added by me and it crashes. So it means users will have to update their middlewares.

wRAR:
Yeah, I'd say using this to run http requests is more of a side effect, one day we indeed want some replacement for inline-reqests (there was some old ticket about integrating it).

docs/topics/asyncio.rst
wRAR: I think that's just the treq API, you can make a wrapper for it if you really want.

scrapy/spidermiddlewares/urllength.py
wRAR: It's a tricky question and I'm going to add docs for this question and maybe do some more research when the current situation is documented. Old sync middlewares indeed cannot work with async callbacks at least with the code in this PR.

docs/topics/asyncio.rst
kmike: +1 to support awaiting Requests, but I think that's a separate task. We'd need to decide on the API, how it works with middlewares, how are exceptions handled, etc.

kmike:
@pawelmhm I think the goal of this PR is to make it possible to have async def parse methods, and enable future improvements like having a built-in way to await for scrapy Requests. In the current form it allows you to use asyncio and twisted libraries in callbacks much more easily, which is already pretty cool :)

scrapy/core/spidermw.py
```diff
@@ -3,9 +3,8 @@
 
 See documentation in docs/topics/spider-middleware.rst
 """
-import collections.abc
 from itertools import islice
-from typing import Any, Callable, Generator, Iterable, Union
+from typing import Any, Callable, Generator, Iterable, Union, AsyncIterable, AsyncGenerator
```
Gallaecio, line 7: :lipstick: 

```suggestion
from typing import Any, AsyncGenerator, AsyncIterable, Callable, Generator, Iterable, Union
```

scrapy/core/scraper.py
```diff
@@ -188,7 +186,8 @@ def handle_spider_error(self, _failure: Failure, request: Request, response: Res
     def handle_spider_output(self, result: Iterable, request: Request, response: Response, spider: Spider):
         if not result:
             return defer_succeed(None)
-        if isinstance(result, collections.abc.AsyncIterable):
+        it: Union[Generator, AsyncGenerator]
```
Gallaecio, line 189: Today I learned!

scrapy/core/spidermw.py
```diff
@@ -58,7 +60,9 @@ def _process_spider_input(self, scrape_func: ScrapeFunc, response: Response, req
                 return scrape_func(Failure(), request, spider)
         return scrape_func(response, request, spider)
 
-    def _evaluate_iterable(self, response, spider, iterable, exception_processor_index, recover_to):
+    def _evaluate_iterable(self, response: Response, spider: Spider, iterable: Union[Iterable, AsyncIterable],
+                           exception_processor_index: int, recover_to: Union[MutableChain, MutableAsyncChain]
```
Gallaecio, line 64: I wonder if it would make sense to provide in `scrapy.utils.typing` aliases for some of these sync-or-async types (e.g. `AnyIterable`, `AnyMutableChain`). Although maybe it’s more readable to specify them explicitly with `Union` as we do now, instead of having to go to a different module to find out what the type stands for.

Gallaecio (approved):

docs/topics/coroutines.rst
```diff
@@ -85,13 +87,16 @@ This means you can use many useful Python libraries providing such code::
...
-                    additional_data = await r.text()
+                    additional_data = await additional_response.text()
             # ... use response and additional_data to yield items and requests
 
 .. note:: Many libraries that use coroutines, such as `aio-libs`_, require the
           :mod:`asyncio` loop and to use them you need to
           :doc:`enable asyncio support in Scrapy<asyncio>`.
 
+.. note:: If you want to ``await`` on Deferreds, you may need to
+          :ref:`wrap them<asyncio-await-dfd>`.
```
kmike, line 98: Does it mean that the example above is incorrect? If asyncio is enabled, parse_with_deferred won't work; if asyncio is not enbaled, parse_with_asyncio won't work?

docs/topics/coroutines.rst
```diff
@@ -103,3 +108,68 @@ Common use cases for asynchronous code include:
...
+            # ... do the same with async result
+
+        def process_spider_output(self, response, result, spider):
+            if isasyncgen(result):
+                return self._async_process_spider_output(self, response, result, spider)
+            else:
+                return self._normal_process_spider_output(self, response, result, spider)
+
+If you are writing a middleware that you intend to publish or to use in many
+projects, this is likely the best way to implement it. It may be possible to
```
kmike, line 173: I wonder if we should be thinking about eventually deprecating non-async process_spider_output. Without doing it, users will always need to write code like the above, which complicates middleware implementation. They'd also need to call as_async_generator always, as `result` can be of different type.

If we disregard backwards compatibility, and all Scrapy ecosystem is updated, won't 

```py
    class ProcessSpiderOutputAsyncGenMiddleware:
        async def process_spider_output(self, response, result, spider):
            async for res in result:
                # ... do something with res
```

be a better API? If so, is there a plan to eventually get there?

Gallaecio: I’ve been thinking that Scrapy 3.0 could be about this, dropping support for synchronous definitions where async is supported. I was thinking of the expected behavioral change for `start_requests`, not this, but the point is the same. So +1 from me.

wRAR: I agree that making everything async will greatly simplify the code, but I'm not sure what should be the steps to that.

Normally there are 4-5 parts of middleware-related code that can be sync, async or both:

1. Spider methods (depends on the specific middleware method discussed).
2. Methods of middlewares in the user project.
3. Methods of 3rd-party middlewares used in the user project.
4. Methods of bundled middlewares.
5. The Scrapy code that runs middleware code on the spider method results.

4 and 5 can be rewritten when the Scrapy maintainers decide, 1 and 2 can be rewritten by the user when they decide or when it's required by other components, but I'm not sure what the maintainers of 3 should do. Initially the code should be universal but at which point can they convert it to async-only? After the Scrapy maintainers declare the sync support deprecated but before they remove it from 4 and 5?

Gallaecio: > at which point can they convert it to async-only? After the Scrapy maintainers declare the sync support deprecated but before they remove it from 4 and 5?

It’s a good point, their code will stop working with the latest Scrapy if they have not made the switch by the time Scrapy has. 

Besides increasing the Scrapy major version to make this aspect obvious, we could consider increasing the time between the original deprecation to the support removal for this specific case, e.g. 2 years instead of 1 year. So, if the last of this deprecations is introduced in Scrapy 2.n, we would not release Scrapy 3.0 until that time after the release of 2.n.

On the other hand, 1 year may already be enough, and if those middlewares do not upgrade users can always keep Scrapy in a 2.x version, as as people that still need Python 2 use Scrapy 1.8. Similarly, I imagine we can keep backporting security fixes to 2.x after 3.0 is released.

docs/topics/coroutines.rst
wRAR: @kmike we can split the example into two spiders or we can maybe replace aiohttp with something that doesn't require the asyncio reactor, what do you think?

Gallaecio: +1 to split, feature both scenarios.

docs/topics/coroutines.rst
kmike: We discussed it on a call. Thoughts:

1. It'd be nice not to require third-party and user middlewares to maintain compatibility code.
2. it seems this can be achieved by auto-promoting sync to async, and auto-downgrading async to sync. We should be warning when downgrading happens. Warning for auto-promoting is not required, because it is a safe operation. We can add it later if it proves problematic.
3. In this case the idea is to suggest always writing `async def process_spider_output`. API would change between sync and async versions: in async def version `results` is going to be async iterator, not a sync one. It means there is no reason to use as_async_generator in the user code, simplifying the API.

@wRAR is to try to implement this approach.

kmike: A comment I've been writing for some time. I think it is not valid, but don't recall why :)

---

@wRAR a good question. 

If we rewrite them as async def, it'd mean all regular middlewares would start to issue warnings, right?

If we maintain compatibility code, it'd mean warnings would be only issued when there is `async def parse` method, or when there is some async def middleware.

Let's say we choose to rewrite built-in middlewares to async def. In this case users would be getting warnings after updating to latest Scrapy version. Third-party middlewares would need to be updated. But the issue is that third-party middlewares would want to support older Scrapy versions as well, where async def won't work (even with warnings). It means they'd need to add compatibility code, which is exactly what we'd like to avoid.

So, it seems it is better to keep built-in middlewares universal.

It'd mean that there won't be warnings if all third-party and user middlewares as sync, but as soon as async def parse is used, or an async def middleware is inserted in a project which also has sync middlewares, warnings would start to appear. That's not ideal, but hopefully it'd give users some time.

When talking about compatibility, there are actually 2 concerns:

1) compatibility with different mixes of def and async def code,
2) compatibility with older Scrapy versions

Gallaecio: @kmike You do mention that we want to avoid third-party developers having to make their spider middlewares universal. However, to maintain support with old Scrapy versions, I imagine that’s the only possible route.

Moreover, universal implementations support coroutines while avoiding async-to-sync conversions. Those conversions are problematic (I do not recall why, we should cover this in the documentation), so they generate a warning. Avoiding that warning is argument enough I think for both built-in and third-party middlewares to aim for universal implementations.

So I expect third-party middlewares to go for ugly, universal implementations, until they decide to drop support for Scrapy 2.5 and earlier. While not ideal, I cannot see any other way to go about this.

I think the best we can do to improve the experience is to try an make it as easy as possible to implement these universal spider middlewares. Although I’m concerned any step in that direction would make things more messy in the long term.

We cannot provide utility code to make them easier to write, because it would not be available in old Scrapy versions.

The only thing I think we could do is allow avoiding the `isasyncgen` call by allowing to define a coroutine implementation with a different method name (e.g. async_process_spider_output) which old Scrapy versions would simply ignore. That’s the best we can do, I think. But I’m not even sure it is worth the trouble of later deprecating that method.

Gallaecio (commented):
I’m leaving some documentation suggestions, but I realize that some of the affected documentation may actually change with the automatic upgrade/downgrade approach.

Gallaecio:
> For 10, there is one problem: Scrapy cannot know before calling a method whether it's universal (takes both sync and async iterables) or not (takes async iterables only if it's `inspect.isasyncgenfunction`), and the only solution I see is adding some flag to the function object (which in practice requires a decorator) or maybe renaming the function.
> 
> […]
> 
> For 12, we need to research what benefits will making a new method name give us and whether it will cause problems in the future when dropping sync support.

What if we allow defining the method as sync (old way), as async (new way), and for universal we go with defining the method as sync but also defining an async counterpart method, with a different name? With this approach, we would be able to tell before run time if the code is universal. And we would not be switching names going forward, only universal implementation would use the alternative name, to be deprecated once we drop support for the sync way (in Scrapy 3?).

Gallaecio:
> > What if we allow defining the method as sync (old way), as async (new way), and for universal we go with defining the method as async but also defining a sync counterpart method with a different name, e.g. suffixed with _sync?
> 
> That's definitely better than adding an _async-suffixed method, I'll think about the details of this.

On second though, for old Scrapy versions, `_async` needs to be the way (old Scrapy can ignore that, but it cannot use `_sync`).

But it does not change much. `_async` would only be used for universal, new-way-only code would use `async def <regular name>`.

wRAR:
> Maybe we determine which method should be used in the case of universal middlwares

I don't think this is possible. If you mean examining the type of the previous method in the stack, then for the actually first method you need to examine the iterable you get from the _callback_.

wRAR:
I have a problem with implementing downgrading. Currently [all that code](https://github.com/scrapy/scrapy/blob/ed50a81f505e3c61226e9caeb210cf3d5c9c152c/scrapy/core/spidermw.py#L97) is synchronous and without downgrading it can be kept synchronous (we don't unroll iterables there, just wrap them in other iterables, actual unrolling including waiting on async ones will happen later). Downgrading is essentially waiting until an async generator returns all values, which needs to happen here somewhere. As the entry point for this machinery (`scrape_response()`) returns a Deferred, this should technically be possible but will require changing all of those methods in that file to be async (most likely with `@inlineCallbacks`) as all of them call each other. All of this should also be rolled back when removing downgrading support.

wRAR:
As the problem with this is `_process_spider_output` and `_process_spider_exception` calling each other (because of the feature with returning iterables from `process_spider_exception`), and other things calling `_process_spider_exception`, this seems to be easily broken by not waiting on `_process_spider_output` in `process_spider_exception` and either taking the result if it's available immediately or raising an exception. This, if I understand correctly, corresponds to something like ~~"if some `process_spider_exception` method returns an async iterable, the next middleware needs to be async-aware"~~ "async iterables returned from callbacks and `process_spider_output` can be downgraded but ones returned from `process_spider_exception` can't" which I think is fine when documented. Now I need to make sure this indeed works.

scrapy/core/spidermw.py
```diff
@@ -82,11 +110,22 @@ def _process_spider_exception(self, response: Response, spider: Spider, _failure
...
+                # _process_spider_output() returns a Deferred only because of downgrading so this can be
+                # simplified when downgrading is removed.
+                if dfd.called:
+                    # the result is available immediately if _process_spider_output didn't do downgrading
+                    return dfd.result
+                else:
+                    # we forbid waiting here because otherwise we would need to return a deferred from
+                    # _process_spider_exception too, which complicates the architecture
+                    msg = f"Async iterable returned from {method.__qualname__} cannot be downgraded"
+                    raise _InvalidOutput(msg)
```
Gallaecio, line 128: As I was working on a [proposal to refactor the new documentation](https://github.com/wRAR/scrapy/pull/1), I got stuck at trying to cover `process_spider_exception` properly.

I was trying to understand in which scenarios trigger this exception, and I came to realize the following about `process_spider_exception`:

- If defined as an asynchronous generator, it cannot return `None`. If it does not yield anything, that will count as an empty iterable. So with `async def` + `yield` we lose the ability to get Scrapy to call the next `process_spider_exception` method.
- If defined as a coroutine, an exception is raised because it is neither an iterable nor `None` (it is a coroutine), so `async def` _without_ `yield` is not an option either.

In other words, `async def process_spider_exception` can never return `None` to get Scrapy to let the next `process_spider_exception` work.

I wonder if it would not be better to limit `process_spider_exception` to being synchronous for now, until we can figure out an API for it that allows the `None` behavior as well. This feels similar to start_requests, i.e. we may need to define some special object that may be yielded from the asynchronous generator to indicate the same as `None` does in the synchronous version.

Gallaecio (changes requested):
[I am proposing some changes documentation-wise](https://github.com/wRAR/scrapy/pull/1), but otherwise this looks great to me.

auxsvr:
Shouldn't we also update `CrawlSpider._parse_response`, `CrawlSpider._requests_to_follow`, `LxmlParserLinkExtractor._extract_links` etc to support coroutines and async generators?

wRAR:
@auxsvr on one hand we haven't thought about this and doing this shouldn't be necessary to merge this feature as existing CrawlSpider-based spiders should continue to work and this support can be added separately (if you suspect this is not true please voice your suspicions), on other hand it should be possible to convert any CrawlSpider-based spider to a normal one if one wants to use async callbacks.

kmike:
A fix for the typing issues: https://github.com/wRAR/scrapy/pull/2

scrapy/spidermiddlewares/depth.py
```diff
@@ -28,31 +28,42 @@ def from_crawler(cls, crawler):
...
-                        self.stats.inc_value(f'request_depth_count/{depth}',
-                                             spider=spider)
-                    self.stats.max_value('request_depth_max', depth,
-                                         spider=spider)
-            return True
+        # base case (depth=0)
+        if 'depth' not in response.meta:
+            response.meta['depth'] = 0
+            if self.verbose_stats:
+                self.stats.inc_value('request_depth_count/0', spider=spider)
```
Gallaecio, line 35: :lipstick: Maybe worth moving to a private method, to avoid duplication?

Gallaecio (approved):
Overwhelmingly amazing!

kmike:
Tremendous work, thanks @wRAR!

PR #4987, opened 2021-02-12
Making sure we can run async def tests

wRAR:
With `twisted.trial` we are able to write asynchronous test functions by returning a Deferred from them. We also have the `deferred_f_from_coro_f` higher order function that can be used as a decorator for an `async def` test function and everything should work as expected. So I'm adding this small test case to check that we indeed can write working tests this way.

Unfortunately, a naked `async def` test function will always succeed, with a warning that a coroutine was not awaited (ideally we should get rid if warnings from tests, but this is still easy to miss).

I added these tests because I've just found for #4978 that such tests fail with old Twisted. So this PR should fail too.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,8 @@ from pathlib import Path
 
 import pytest
 
+from scrapy.utils.reactor import install_reactor
+
 from tests.keys import generate_keys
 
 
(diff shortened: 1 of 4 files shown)


wRAR:
As the only thing we use pytest-twisted for is installing the asyncio reactor (we had a dep on it before this code was merged but it was unused at that time) and we cannot easily tune the asyncio loop (which we need for Twisted 17.9.0), and we already have code to install a reactor, I've just put this code into `conftest.py` and dropped pytest-twisted.

Ironically, pytest-twisted has code for running async def tests but looks like it's not suitable for methods.

Gallaecio (approved):

PR #4993, opened 2021-02-19
Allow customization of retry middleware

elacuesta:
I've been tasked with adding [retrying functionality to an external middleware](https://github.com/scrapy-plugins/scrapy-crawlera-fetch/issues/12), and I thought about leveraging the existing retrying middleware to avoid reinventing the wheel. Unfortunately, it contains many things that are specific to the general purpose case, which makes it hard to extend:
* the logger is a module-level variable named after `__name__`, so it's always `scrapy.downloadermiddlewares.retry`
* the setting, meta and stats names are hardcoded, which does not allow custom messages (stats like `my_own_retry/count`, for instance)

Let me know if this makes any sense, I didn't want to add tests and docs yet in case this is not deemed useful.

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -41,11 +41,12 @@ class RetryMiddleware:
     SETTING_PREFIX = ""
     META_PREFIX = ""
     STATS_PREFIX = ""
+    LOGGER_NAME = __name__
 
     def __init__(self, settings):
         if not settings.getbool(f"{self.SETTING_PREFIX}RETRY_ENABLED"):
             raise NotConfigured
-        self.logger = logging.getLogger(__name__)
+        self.logger = logging.getLogger(self.LOGGER_NAME)
         self.max_retry_times = settings.getint(f"{self.SETTING_PREFIX}RETRY_TIMES")
         self.retry_http_codes = set(int(x) for x in settings.getlist(f"{self.SETTING_PREFIX}RETRY_HTTP_CODES"))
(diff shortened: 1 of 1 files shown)


scrapy/downloadermiddlewares/retry.py
```diff
@@ -40,19 +38,25 @@ class RetryMiddleware:
                            ConnectionLost, TCPTimedOutError, ResponseFailed,
                            IOError, TunnelError)
 
+    SETTINGS_PREFIX = ""
+    META_PREFIX = ""
+    STATS_PREFIX = ""
+    LOGGER_NAME = __name__
```
Gallaecio, line 44: Maybe we could allow to define a custom `logger` object, rather than a name, for more flexibility?

scrapy/downloadermiddlewares/retry.py
```diff
@@ -40,19 +38,25 @@ class RetryMiddleware:
                            ConnectionLost, TCPTimedOutError, ResponseFailed,
                            IOError, TunnelError)
 
+    SETTINGS_PREFIX = ""
+    META_PREFIX = ""
+    STATS_PREFIX = ""
```
Gallaecio, line 43: I also wonder if, instead of allowing so customize the prefixes, we should allow to customize the actual strings.

Gallaecio (commented):
Despite my “wonders”, I’m :heavy_check_mark: on this.

If it’s not too hard, it would be great to cover the change with tests, but I understand that in this case it may be more trouble than it is worth it.

elacuesta:
Closing in favor of https://github.com/Gallaecio/scrapy/pull/3

PR #4994, opened 2021-02-19
http2: acceptable protocol update, tests

elacuesta:
I felt like this change was big enough not to make it directly to the `http2` branch. With this PR, test coverage in #4769 should reach the build target (not 100% though, but I don't feel familiar enough with the original patch to be able to simulate every scenario).

### DIFF
--- a/scrapy/core/http2/protocol.py
+++ b/scrapy/core/http2/protocol.py
@@ -33,13 +33,17 @@ from scrapy.spiders import Spider
 logger = logging.getLogger(__name__)
 
 
+ACCEPTABLE_PROTOCOL = b"h2"
+
+
 class InvalidNegotiatedProtocol(H2Error):
 
-    def __init__(self, negotiated_protocol: str) -> None:
+    def __init__(self, negotiated_protocol: bytes) -> None:
         self.negotiated_protocol = negotiated_protocol
 
     def __str__(self) -> str:
-        return f'InvalidNegotiatedProtocol: Expected h2 as negotiated protocol, received {self.negotiated_protocol!r}'
+        return (f"Expected {ACCEPTABLE_PROTOCOL.decode('utf-8')} as negotiated protocol,"
+                f" received {self.negotiated_protocol.decode('utf-8')}")
 
 
 class RemoteTerminatedConnection(H2Error):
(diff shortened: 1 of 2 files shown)


scrapy/core/http2/protocol.py
```diff
@@ -33,13 +33,17 @@
...
 class InvalidNegotiatedProtocol(H2Error):
 
-    def __init__(self, negotiated_protocol: str) -> None:
+    def __init__(self, negotiated_protocol: bytes) -> None:
         self.negotiated_protocol = negotiated_protocol
 
     def __str__(self) -> str:
-        return f'InvalidNegotiatedProtocol: Expected h2 as negotiated protocol, received {self.negotiated_protocol!r}'
+        return (f"Expected {ACCEPTABLE_PROTOCOL.decode('utf-8')} as negotiated protocol,"
+                f" received {self.negotiated_protocol.decode('utf-8')}")
```
Gallaecio, line 46: :lipstick: Instead of decoding, maybe we could use `!r` and let it be shown as `b'…'`. Otherwise, this will break on non-UTF-8 stuff, which while unlikely is still possible.

scrapy/core/http2/protocol.py
```diff
@@ -33,13 +33,17 @@
 logger = logging.getLogger(__name__)
 
 
+ACCEPTABLE_PROTOCOL = b"h2"
```
Gallaecio, line 36: :lipstick: Maybe simply `PROTOCOL`? I’m definitely [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) here, so feel free to mark this as solved without any change.

Gallaecio (approved):
Changing the expected protocol to reproduce issues with unexpected server-side protocols, hadn’t though of that!

scrapy/core/http2/protocol.py
elacuesta: I don't much like `ACCEPTABLE_PROTOCOL` to be honest, but I think I'd prefer to go with `PROTOCOL_NAME`, for extra explicitness.

scrapy/core/http2/protocol.py
elacuesta: Sure, makes sense

PR #4995, opened 2021-02-20
Updated to transport.loseConnection()

### LINKED ISSUE #4991: Use transport.loseConnection()
A minor cleanup for the future.

At the moment [we use this in `http11.py`](https://github.com/scrapy/scrapy/pull/4897/files/2eabfb60e8112a85f4f1c6ce16c4a789856e3ead#r578344512):

```
with suppress(AttributeError):
    txresponse._transport._producer.loseConnection()
```

[Once we drop support for Twisted < 18.4.0](https://github.com/twisted/twisted/commit/38559237ec584e41d2c8667c3b9573865ddff19e), we should be able to switch to:

```
txresponse._transport.loseConnection()
```

gabrielztk:
Resolves #4991 comment

> Thanks! There are two additional occurrences:
> 
>     * https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/core/downloader/handlers/http11.py#L404
> 
>     * https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/core/downloader/handlers/http11.py#L523


Gallaecio:
Note: This change is not needed until we drop support for Twisted 18.4.0, which may not happen in a while.

kmike:
We now require Twisted 18.9.0+ (since https://github.com/scrapy/scrapy/pull/5514), so that's fine to proceed with this PR.

wRAR (approved):

PR #5000, opened 2021-02-23
Drop downloader requests upon stop

### LINKED ISSUE #4749: Allow to stop Crawler non-gracefully, without completing all requests in the Downloader
By default, Scrapy waits until Downloader queue is empty when doing a graceful shutdown. This behavior ensures that if request is out of Scheduler (e.g. loaded from a disk queue), it is processed; doing otherwise would make shutdown not so graceful - some requests can be lost. A downside is that the shutdown could take a significant amount of time - e.g. with the default CONCURRENT_REQUESTS=16, 3 retries and 180s timeout it can take almost 2.5 hours, in the very worst case.

Currently the alternative (double Ctrl-C) is to stop the reactor; it drops all tasks in progress, which is much faster, but still provides a bit of protection - code which is executed between yields to the event loop won't be stopped in the middle. It is much safer than just killing the process. But this kind of protection still can cause issues as well: for a request some tasks may be executed, and some missed. 

Another issue is that there is no public API for non-graceful shutdown, you need to send 2 sgnals to the process to initiate it. It may be also on a wrong level - it is not possible to stop individual Crawlers inside a CrawlerRunner / CrawlerProcess non-gracefully, this is all or nothing.

What do you think about adding another way of shutting down a Crawler? Semantics would be this:

* don't accept new requests to the Downloader
* drop all requests currently in the Downloader. Add a log message and a record in stats with an amount of dropped requests.
* finish processing of all responses
* then stop

Gallaecio:
I added a test that shows that drop request succeeds. But I wonder if we should test something more.

Resolves #4749

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -102,7 +102,10 @@ class ExecutionEngine:
             # Will also close downloader
             return self._close_all_spiders()
         else:
-            return defer.succeed(self.downloader.close())
+            for _ in self.downloader.stop():
+                pass
+            self.downloader.close()
+            return defer.succeed(None)
 
     def pause(self):
         """Pause the execution engine"""
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -308,6 +308,22 @@ class ExecutionEngine:
                     {'reason': reason},
                     extra={'spider': spider})
 
+        dropped_request_count = 0
+        for request in self.downloader.stop():
+            slot.remove_request(request)
+            dropped_request_count += 1
+        if dropped_request_count:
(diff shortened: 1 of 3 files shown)


scrapy/core/engine.py
```diff
@@ -308,6 +316,22 @@ def close_spider(self, spider, reason='cancelled'):
                     {'reason': reason},
                     extra={'spider': spider})
 
+        dropped_request_count = 0
+        for request in self.downloader.stop():
+            slot.remove_request(request)
```
Gallaecio, line 321: Maybe we should return them to the scheduler, not just drop them.

Issue #5003, opened 2021-02-23
refactoring _link_allowed to reduce cyclomatic complexity


dmariel:
## Summary

After some exploring with cyclomatic complexity tools (lizard), the function was found to have the third highest complexity.

## Motivation

Low complexity allows for higher readability, testability and maintainability.

## Solution

Refactor.

## Additional context

N/A

Gallaecio:
Closing as per https://github.com/scrapy/scrapy/pull/5004#issuecomment-784857085

Gallaecio:
For future reference, there’s no need to open an issue first if you plan to create a pull request to fix it right away, you can open the pull request directly.

PR #5004, opened 2021-02-23
Refactor _link_allowed_

djikoSal:
Co-authored-by: alkazaz alkazaz@kth.se
Co-authored-by: swill swill@kth.se
Co-authored-by: lerjevik lerjevik@kth.se
Co-authored-by: aljica aljica@kth.se

Resolves: #5003

### DIFF
--- a/scrapy/linkextractors/__init__.py
+++ b/scrapy/linkextractors/__init__.py
@@ -94,20 +94,31 @@ class FilteringLinkExtractor:
     def _link_allowed(self, link):
         if not _is_valid_url(link.url):
             return False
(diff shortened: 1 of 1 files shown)


Gallaecio (commented):
I don’t think the changes improve readability, which is the only point of decreasing cyclomatic complexity.

The only refactoring I can think of is turning this:

```
if A:
    return False
if B:
    return False
```

Into:

```
if (
    A
    or B
):
    return False
```

I definitely don’t think there’s anything in this method worth extracting into a function, beyond what already is.

djikoSal:
Thank you for the review! 

We will close it for now.

Issue #5011, opened 2021-02-25
Enable to configure the timeout for feed exporters


alepuccetti:
## Summary

Enable to configure the timeout for feed exporters using.

## Motivation

This is useful when testing spiders in local environment (or via VPN) where the internet connections is not very fast and uploading the results might result in a timeout.

## Describe alternatives you've considered

Could not find any alternative.

AdrianAtZyte:
For testing locally, why not simply output to a local file instead? I don’t quite get the goal here.

PR #5012, opened 2021-02-25
Add FEED_STORAGE_TIMEOUT setting and implement it for Google Cloud Storage

alepuccetti:
This add `FEED_STORAGE_TIMEOUT` as new settings and implement it for Google Cloud storage. I did not put `GCS` inside the name because it can be easily extended to other feed exporter like S3 or FTP.

Fixes: https://github.com/scrapy/scrapy/issues/5011


docs/topics/settings.rst
```diff
@@ -851,6 +851,13 @@ FEED_STORAGE_GCS_ACL
 The Access Control List (ACL) used when storing items to :ref:`Google Cloud Storage <topics-feed-storage-gcs>`.
 For more information on how to set this value, please refer to the column *JSON API* in `Google Cloud documentation <https://cloud.google.com/storage/docs/access-control/lists>`_.
 
+.. setting:: FEED_STORAGE_TIMEOUT
+
+FEED_STORAGE_TIMEOUT
+--------------------
+The Feed Storage Timeout you can configure the timeout of feed upload oprations.
+Currently implemented only for Google Cloud Storage.
```
Gallaecio, line 859: I think we should also mention the value unit (seconds?), as well as the default value.

scrapy/extensions/feedexport.py
```diff
@@ -160,9 +160,10 @@ def _store_in_thread(self, file):
 
 class GCSFeedStorage(BlockingFeedStorage):
 
-    def __init__(self, uri, project_id, acl):
+    def __init__(self, uri, project_id, acl, timeout):
```
Gallaecio, line 163: I think we could make this an option that you can pass to `FEEDS`, and hence read it from a `feed_options` argument here (see how `FTPFeedStorage` reads the `overwrite` option below).

tests/test_feedexport.py
```diff
@@ -421,11 +421,12 @@ def test_parse_settings(self):
         except ImportError:
             raise unittest.SkipTest("GCSFeedStorage requires google-cloud-storage")
 
-        settings = {'GCS_PROJECT_ID': '123', 'FEED_STORAGE_GCS_ACL': 'publicRead'}
+        settings = {'GCS_PROJECT_ID': '123', 'FEED_STORAGE_GCS_ACL': 'publicRead', 'timeout': 60}
```
Gallaecio, line 424: The option is not called `timeout`, so I would expect this to fail provided we put a value other than the default value.

Gallaecio:
Nice work!

Please, have a look at the [recent addition of overwrite support](https://github.com/scrapy/scrapy/pull/4512) to get a better idea of what it entails to make `FEEDS` support additional options (the command-line interface parts you can ignore).

PR #5014, opened 2021-02-26
docs: Clarify there's one extension instance per crawler

jpmckinney:
The previous documentation suggested there was a single instance of each extension. This is not the case. For example, when [running multiple spiders in the same process](https://docs.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process), the [LogStats extension](https://docs.scrapy.org/en/latest/_modules/scrapy/extensions/logstats.html#LogStats) still works correctly, and it doesn't track stats per spider.

### DIFF
--- a/docs/topics/extensions.rst
+++ b/docs/topics/extensions.rst
@@ -7,8 +7,7 @@ Extensions
 The extensions framework provides a mechanism for inserting your own
 custom functionality into Scrapy.
 
-Extensions are just regular classes that are instantiated at Scrapy startup,
-when extensions are initialized.
+Extensions are just regular classes.
 
 Extension settings
 ==================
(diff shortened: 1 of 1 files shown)


Gallaecio (commented):
Thanks, that is a good point!

However, I personally feel like this change could be reduced to appending “ per spider being run” to “instantiating a single instance of the extension class”, and things should be clear enough.

I also think that we should avoid mentioning the crawler at this stage, as the crawler object is something people reading these 2 sections may not yet know.

jpmckinney:
@Gallaecio I've made the simpler change you suggest.

Gallaecio (approved):

PR #4980, opened 2021-02-07
Fixes #802: Adding support for request delay

realslimshanky:
Problem statement: To add delay on individual requests independent of all the other request settings.

Solution: A minor update to `scrapy/core/downloader/__init__.py` allows us to delay said request using `meta` settings.
Try this example
```
import scrapy


class QuotesSpider(scrapy.Spider):
    name = "quotes"

    def start_requests(self):
        urls = [
            'http://quotes.toscrape.com/page/1/',
            'http://quotes.toscrape.com/page/2/',
        ]
        for index, url in enumerate(urls):
            if index == 0 :
                yield scrapy.Request(url=url, callback=self.parse, meta={'request_delay': 5})
            else:
                yield scrapy.Request(url=url, callback=self.parse)

    def parse(self, response):
        print(f'Parsed {response.url}')
```

With `request_delay` set to the first request, it's delayed by 5 seconds and the output would be 
```
$ scrapy crawl quotes
2021-02-08 02:28:33 [scrapy.utils.log] INFO: Scrapy 2.4.1 started (bot: tutorial)
2021-02-08 02:28:33 [scrapy.utils.log] INFO: Versions: lxml 4.6.2.0, libxml2 2.9.10, cssselect 1.1.0, parsel 1.6.0, w3lib 1.22.0, Twisted 20.3.0, Python 3.9.1 (default, Jan  8 2021, 17:17:43) - [Clang 12.0.0 (clang-1200.0.32.28)], pyOpenSSL 20.0.1 (OpenSSL 1.1.1i  8 Dec 2020), cryptography 3.3.1, Platform macOS-11.1-x86_64-i386-64bit
2021-02-08 02:28:33 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.selectreactor.SelectReactor
2021-02-08 02:28:33 [scrapy.crawler] INFO: Overridden settings:
{'BOT_NAME': 'tutorial',
 'NEWSPIDER_MODULE': 'tutorial.spiders',
 'ROBOTSTXT_OBEY': True,
 'SPIDER_MODULES': ['tutorial.spiders']}
2021-02-08 02:28:33 [scrapy.extensions.telnet] INFO: Telnet Password: ff3242e0b3c80780
2021-02-08 02:28:33 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.corestats.CoreStats',
 'scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.memusage.MemoryUsage',
[...]

PR #5015, opened 2021-02-28
Add per request delay implementation

### LINKED ISSUE #802: Per request delay

WinterComes:
Closes #802 

This changes allow to schedule request in the way of: "please process this request, but no sooner than X seconds from now". It respects rate and concurrency limits because only the scheduler `next_request` part with adding new queue is updated.
The main idea here is to have a separate queue where is possible to manage requests with delays (if this parameter is added). This allows to not update and to respects the single reponsobility principle in the current queues.

This new priority queue has the similar implementation as default `ScrapyPriorityQueue`, but it uses timestamps as priority. It allows to easily add per request delays and doesn't touch the default queue. If this priority queue doesn't contain any requests to process or the current time is less than the next time when request should be processed it returns None and request from the main priority queue will be taken or it will go as usual flow.

I've added a few tests to be sure that it works as intended. With this tests we can see so far that shceduler processes requests from the new queue at the right time, at other time it respects the order of basic priority queue.
But if this changes look promising not only for me I can add more tests to be sure that everything is ok.

I've added a test dependency `freezegun` it allows to easily play with the time. I'm not aware about the policy of adding new dependencies, so if it not acceptable please tell, I will try to find some alternative way to test this timing tests.

Briefly, this changes allow:

1. process a request no sooner than X seconds from now;
2. respect rate and concurrency limits
3. avoid of updating all queues to have delays support;
4. have a way for scheduler/queue to say "I have some requests, so don't stop the spider, but I won't give you anything now".

Also closes #3835 and closes #4980.


scrapy/core/scheduler.py
```diff
@@ -36,10 +36,12 @@ class Scheduler:
     Also, it handles dupefilters.
     """
     def __init__(self, dupefilter, jobdir=None, dqclass=None, mqclass=None,
-                 logunser=False, stats=None, pqclass=None, crawler=None):
+                 logunser=False, stats=None, pqclass=None, dpqclass=None,
+                 crawler=None):
```
Gallaecio, line 40: In its current form, this change is backward incompatible, because scheduler classes are allowed to receive parameters by position. It should be:

```suggestion
                 logunser=False, stats=None, pqclass=None, crawler=None,
                 dpqclass=None):
```

Gallaecio (commented):
Thanks! I’m in a rush today, I hope to get a chance to get a better look later this week.

Gallaecio:
Note: macOS failures occur at random at the moment, we are considering disabling them temporarily until we find a solution. Feel free to ignore them.

WinterComes:
thanks for notice @Gallaecio

scrapy/core/scheduler.py
```diff
@@ -52,12 +54,13 @@ def from_crawler(cls, crawler):
         dupefilter_cls = load_object(settings['DUPEFILTER_CLASS'])
         dupefilter = create_instance(dupefilter_cls, settings, crawler)
         pqclass = load_object(settings['SCHEDULER_PRIORITY_QUEUE'])
+        dpqclass = load_object(settings['SCHEDULER_DELAYED_REQUESTS_PRIORITY_QUEUE'])
```
Gallaecio, line 57: What about shortening it to `SCHEDULER_DELAY_QUEUE` (or `SCHEDULER_TIME_QUEUE`)?

scrapy/core/scheduler.py
```diff
@@ -52,12 +54,13 @@ def from_crawler(cls, crawler):
...
         dqclass = load_object(settings['SCHEDULER_DISK_QUEUE'])
         mqclass = load_object(settings['SCHEDULER_MEMORY_QUEUE'])
         logunser = settings.getbool('SCHEDULER_DEBUG')
         return cls(dupefilter, jobdir=job_dir(settings), logunser=logunser,
                    stats=crawler.stats, pqclass=pqclass, dqclass=dqclass,
-                   mqclass=mqclass, crawler=crawler)
+                   mqclass=mqclass, crawler=crawler, dpqclass=dpqclass)
```
Gallaecio, line 63: This can also be an issue for subclasses that do not reimplement `from_crawler`.

A safer approach would be:

```
scheduler = cls(<old parameters>)
scheduler.dpqclass = dpqclass
return scheduler
```

scrapy/pqueues.py
```diff
@@ -110,6 +111,28 @@ def __len__(self):
...
+    """This priority queue has the similar implementation as default ScrapyPriorityQueue
+    but it uses timestamps as priority. It allows to easily add per request delays and
+    doesn't touch the default queue.
+    If this priority queue doesn't contain any requests to process or the current time is less than
+    the next time when request should be processed it returns None and request from the main priority queue
+    will be taken.
+    """
+
+    def priority(self, request):
+        now = int(datetime.now().timestamp())
```
Gallaecio, line 124: :lipstick: I wonder why you did not go for `time.time()`, is there a relevant functional difference, or was it just personal preference?

Gallaecio (commented):
Looks great!

Gallaecio:
I’ll close and reopen to force tests to run with the latest changes to `master`.

scrapy/pqueues.py
WinterComes: Don't have any specific preferences. Replaced to `time.time()` in terms of shortening.

WinterComes:
thanks @Gallaecio for the review. I also added a description about the new parameter `request_delay` to docs, where the list of `Request.meta` special keys are described.

Gallaecio (commented):
I think the only thing missing is covering the new setting in `settings.rst`. You could mention that the chosen class must have the same API required for the priority queue setting, with a reference to it (```:setting:`<setting name>` ```). The description of the priority setting does not detail the required API, but that’s not your fault, and I think we can consider that out of the scope of this change.

It may also be the best place to include the API documentation of the default class with `.. autoclass:: …`, but since similar settings do not do that, maybe that’s not necessary. Up to you.

WinterComes:
I added a description of `SCHEDULER_DELAY_QUEUE`. Regarding the .. autoclass:: , I think it might be better to do this action for the all queues together. I would like to take care of this in a separate PR.

docs/topics/settings.rst
```diff
@@ -1303,6 +1303,16 @@ Type of priority queue used by the scheduler. Another available type is
...
 does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
 
+.. setting:: SCHEDULER_DELAY_QUEUE
+
+SCHEDULER_DELAY_QUEUE
+---------------------
+Default: ``'scrapy.pqueues.ScrapyDelayedRequestsPriorityQueue'``
+
+Type of priority queue used by the scheduler for delayed requests. The default class is
+a sublcass of the ``scrapy.pqueues.ScrapyPriorityQueue``, so the chosen class must have
```
Gallaecio, line 1313:
```suggestion
a sublcass of the :class:`scrapy.pqueues.ScrapyPriorityQueue`, so the chosen class must have
```

It will probably link nowhere at the moment, but if we ever provide API documentation for it, then it will.

The same could be done for the default value above, since we support specifying this type of values as classes instead of strings since 2.4.0.

Gallaecio (approved):

ejulio (commented):
The implementation looks really good and clean :clap: 
I'd just add tests for the priority as well.

My only concern i that delayed requests have priority over standard priority and I'm not sure if that should be the case.
hundreds of requests being delayed every microsecond and then a high priority request arrives in the queue.
Should it be processed right away or wait for the delayed ones?
If we respect slots and domains, I don't think it can be a problem, but it is a good scenario to have in mind, I think

WinterComes:
Yes, first the queue with delayed requests is checked and after that if there are no requests to process, requests from the standart queue are picked. Basically it allows to run the request after some approximately X seconds. It absolutely respects slots and domains, because only schduler part is changed.

I've added some tests to check order of `dequeueing` requests depends on standart and delayed priority queues [here](https://github.com/scrapy/scrapy/pull/5015/files#diff-7cab305a6a78540131fcce1fafb912ccf6b52aa50b4a0a2b794aea71004d5a95R148). What do you mean under the "tests for the priority as well"?

Gallaecio:
To address @ejulio’s concern about priority not being taken into account, maybe we could have the dequeuing not yield requests from the delay queue, but instead feed them into the priority queue and then dequeue the priority queue. The problem I see with that approach is that the delayed requests would most likely go last within their priorities.

Alternatively, we could have the dequeing peak the first request from the delayed queue and from the priority queue, compare their priorities, and dequeue and yield the one with the highest priority.

WinterComes:
@Gallaecio I think the first approach might help with the concern. The problem which you described is possible in case when we have a lot of requests with the same priority (incl. delayed request). If delayed requests has a higher priority then with high percentage it will be picked first from the standart queue.

The second approach doesn't resolve problem with requests which have the similar priority, is the delayed request has more priority in that case?

I think the second approach respects more to the standart queue, the second - less

WinterComes:
The reason why I didn't see a problem in the case when delayed queue > standart queue, depends on my use case when the delayed requests more rare event than a standart requests. But I agree and see now that it might be absolutely different use cases

WinterComes:
I've added an implementation for the first approach from [here](https://github.com/scrapy/scrapy/pull/5015#issuecomment-807162391)
Here is some tests to check behaviour: [link](https://github.com/scrapy/scrapy/pull/5015/files#diff-7cab305a6a78540131fcce1fafb912ccf6b52aa50b4a0a2b794aea71004d5a95R189)

scrapy/core/scheduler.py
```diff
@@ -88,6 +100,12 @@ def enqueue_request(self, request):
         return True
 
     def next_request(self):
+        delayed_request = self.dpqs.pop()
+        if delayed_request:
+            self.stats.inc_value('scheduler/dequeued/delayed/memory', spider=self.spider)
+            dqok = self._dqpush(delayed_request)
+            if not dqok:
+                self._mqpush(delayed_request)
```
Gallaecio, line 108: I think we may also want to reproduce the stat updates from `enqueue_request` here, as to avoid scenarios where there have been queues in disk or memory that come from delayed requests but the `enqueued` stat says 0 for disk or memory.

Maybe we could extract an `_enqueue_request` function and use it both here and in `enqueue_request`:

```python
    def _enqueue_request(self, request):
        dqok = self._dqpush(request)
        if dqok:
            self.stats.inc_value('scheduler/enqueued/disk', spider=self.spider)
        else:
            self._mqpush(request)
            self.stats.inc_value('scheduler/enqueued/memory', spider=self.spider)
```

WinterComes: yeah, it makes sense

ejulio (approved):

Gallaecio (approved):

wRAR (approved):

WinterComes:
I've updated a bit, merged master and resolved conflicts (added some typing due the new approach)

victorcatn (approved):
Looks good and clean to me

caffeinatedMike:
Hey @kmike, can you please take a look at this PR when you get a chance. This has been a long-awaited feature and I just want to make sure it doesn't fall off the radar.

WinterComes:
Hello, @Gallaecio 
Do we still have any concerns regarding this PR, or we will just wait for a review from @kmike ?

Gallaecio:
@WinterComes I personally have no concern, and we technically do not need to wait for @kmike to have a look, I would just feel much better if he gets a chance to look at it before we merge it.

I’ll put this in the next Scrapy milestone. In other words, we will not forget about it, we will deal with this before we release the next Scrapy version. I’m not sure when we will be releasing that next version, though.

kmike (commented):
Nice work @WinterComes!

tests/requirements-py3.txt
```diff
@@ -1,6 +1,7 @@
 # Tests requirements
 attrs
 dataclasses; python_version == '3.6'
+freezegun
```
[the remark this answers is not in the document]
VMRuiz: I'm not against adding this if needed, but may we could mock reactor usage with [Twisted Clocks](https://twisted.org/documents/17.9.0/api/twisted.internet.task.Clock.html) to simulate time pass.

scrapy/core/scheduler.py
```diff
@@ -256,6 +270,10 @@ def next_request(self) -> Optional[Request]:
         Increment the appropriate stats, such as: ``scheduler/dequeued``,
         ``scheduler/dequeued/disk``, ``scheduler/dequeued/memory``.
         """
+        delayed_request = self.dpqs.pop()
+        if delayed_request:
+            self.stats.inc_value('scheduler/dequeued/delayed/memory', spider=self.spider)
+            self._enqueue_request(delayed_request)
```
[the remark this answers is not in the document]
Gallaecio: To me, none of the described behaviors seem problematic. I think everything is OK as long as we make sure to explain this in detail in the documentation. And I think it is great that users can play with request priority to adjust the behavior to the liking.

Regarding the setting @VMRuiz proposes, I think we could have a setting for that effect, but I think we can go for something more flexible: implement a `DELAY_PRIORITY` setting, in line with [`DEPTH_PRIORITY`](https://docs.scrapy.org/en/latest/topics/settings.html#depth-priority), and make it 1 by default. Users who do not set priorities would get the delayed-requests-first behavior by default, even for FIFO queues. Although since LIFO is the default, maybe a default value of 0 for this setting may be OK, and less surprising.

scrapy/core/scheduler.py
```diff
@@ -321,6 +346,14 @@ def _dq(self):
                         {'queuesize': len(q)}, extra={'spider': self.spider})
         return q
 
+    def _dpq(self):
+        """ Create a new delayed requests priority queue instance, with in-memory storage """
```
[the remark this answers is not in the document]
Gallaecio: After a private discussion with @kmike, @wRAR and @VMRuiz, it seemed like the main advantage would be the flexibility. Some users may want request priority to be taken into account regardless of whether the request has been delayed or not, and scheduling into the downloader after a delay, bypassing the scheduler, removes that option.

On the other hand, the scheduler approach makes all behaviors achievable through parameters (scheduler queue settings and request priority values).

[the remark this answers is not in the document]
Gallaecio: > For example, currently float request_delay doesn't seem to be prohibited by the docs, but it may be harder to make disk queues work with float priorities, as they're not supported now.

I do not see it as a big issue, since disk queues could pick some granularity (e.g. 0.1s) and adjust the delay accordingly.

But I am OK with going for millisecond granularity on the setting and go for int.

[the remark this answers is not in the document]
kmike: > I do not see it as a big issue, since disk queues could pick some granularity (e.g. 0.1s) and adjust the delay accordingly.

With timestamp-based priorities we'll have a lot of disk queues, with 1 element each; we'd need to benchmark it, to see how it works in practice. Current queue implementation doesn't seem right for float priorities, on a first sight.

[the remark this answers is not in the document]
VMRuiz: I don't like timestamp-based priorities with the current implementation either. 
The code would need to be refactored to use a single queue.

docs/topics/settings.rst
```diff
@@ -1335,6 +1335,16 @@ Type of priority queue used by the scheduler. Another available type is
...
 
+.. setting:: SCHEDULER_DELAY_QUEUE
+
+SCHEDULER_DELAY_QUEUE
+---------------------
+Default: ``'scrapy.pqueues.ScrapyDelayedRequestsPriorityQueue'``
+
+Type of priority queue used by the scheduler for delayed requests. The default class is
+a sublcass of the :class:`scrapy.pqueues.ScrapyPriorityQueue`, so the chosen class must have
+the same API required for the priority queue setting :setting:`SCHEDULER_PRIORITY_QUEUE`.
```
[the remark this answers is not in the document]
Gallaecio: We are using the same interface for both, and I will clarify that.

What I am not sure is that we should do anything other than indicating so as part of this pull request. None of the queue settings indicate the API that is expected for custom classes, and we should probably address that in a separate pull request. They may even deserve a topic of their own.

scrapy/core/scheduler.py
Gallaecio, line 276: I am thinking that it may be best to pop `request_delay` from requests before enqueuing them back, so that if they are re-scheduled for whatever reason the same delay is not applied again.

scrapy/core/scheduler.py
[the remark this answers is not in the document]
Gallaecio: Is disk support for delayed requests something we should implement as part of this pull request, or is that something that can come later?

I do realize that if we change the behavior later to behave like with priority queues, where disk is tried first and memory is only used if serialization fails, it may significantly affect request order, lowering the priority of delayed requests below any unserializable request. But I am not sure how bad that would be, it does not seem too bad to me.

scrapy/core/scheduler.py
```diff
@@ -256,6 +333,10 @@ def next_request(self) -> Optional[Request]:
         Increment the appropriate stats, such as: ``scheduler/dequeued``,
         ``scheduler/dequeued/disk``, ``scheduler/dequeued/memory``.
         """
+        delayed_request = self.dpqs.pop()
+        if delayed_request:
+            self.stats.inc_value('scheduler/dequeued/delayed/memory', spider=self.spider)
+            self._enqueue_request(delayed_request)
```
Gallaecio, line 339: I am also thinking this should be a loop, pop and enqueue all delayed requests whose time has passed.

scrapy/core/scheduler.py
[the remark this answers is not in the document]
Gallaecio: The issues brought up here are now documented, and `DELAY_PRIORITY_ADJUST` implemented, so I am hoping we can consider this resolved.

kmike:
Hey! I acknowledge it's looking close, but not having this feature doesn't look like a blocker for 2.7 release, so I'm removing the milestone for now.

scrapy/core/scheduler.py
```diff
@@ -229,19 +287,55 @@ def close(self, reason: str) -> Optional[Deferred]:
...
         return self.df.close(reason)
 
+    def _enqueue_request(self, request: Request) -> None:
+        dqok = self._dqpush(request)
+        if dqok:
+            self.stats.inc_value("scheduler/enqueued/disk", spider=self.spider)
+        else:
+            self._mqpush(request)
+            self.stats.inc_value("scheduler/enqueued/memory", spider=self.spider)
+
```
GeorgeA92, line 297: `self._dqpush(request)` called twice per produced request: on `_enqueue_request` and on later line 335  on `enqueue_request`. So It will add 2 requests per 1 req. received from callback.
It return error on executing following scrapy as script application (w/o custom delays)
<details><summary>test1.py</summary>

```python
import scrapy; from scrapy.crawler import CrawlerProcess

    name = "quotes"; custom_settings = {"DOWNLOAD_DELAY": 1}
    def start_requests(self):

    def parse(self, response):
        yield {'w': response.url[-4:]}

if __name__ == "__main__":
    process = CrawlerProcess(); process.crawl(QuotesToScrapeSpider); process.start()
```

</details>

<details><summary>log_output(fragments)</summary>

```
2023-02-24 23:27:18 [scrapy.core.engine] INFO: Error while removing request from slot
  File "C:\Users\georg\anaconda3\envs\scrapy_dev\lib\site-packages\twisted\internet\defer.py", line 857, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "C:\Users\georg\PycharmProjects\scrapy\scrapy\core\engine.py", line 204, in <lambda>
    d.addBoth(lambda _: self.slot.remove_request(request))
  File "C:\Users\georg\PycharmProjects\scrapy\scrapy\core\engine.py", line 54, in remove_request
    self.inprogress.remove(request)
KeyError: <GET http://quotes.toscrape.com/page/9/>
2023-02-24 23:27:19 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com/page/10/> (referer: http://quotes.toscrape.com/page/9/)
2023-02-24 23:27:19 [scrapy.core.scraper] DEBUG: Scraped from <200 http://quotes.toscrape.com/page/10/>
{'w': '/10/'}
2023-02-24 23:27:19 [scrapy.core.engine] INFO: Closing spider (finished)
2023-02-24 23:27:19 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'downloader/exception_count': 9,
 'downloader/exception_type_count/builtins.KeyError': 9,
 'downloader/request_bytes': 5384,
 'downloader/request_count': 20,
 'downloader/request_method_count/GET': 20,
 'downloader/response_bytes': 21915,
 'downloader/response_count': 10,
 'downloader/response_status_count/200': 10,
 'elapsed_time_seconds': 23.200037,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2023, 2, 24, 21, 27, 19, 523219),
 'httpcompression/response_bytes': 108561,
 'httpcompression/response_count': 10,
 'item_scraped_count': 10,
 'log_count/DEBUG': 21,
 'log_count/ERROR': 9,
 'log_count/INFO': 19,
 'log_count/WARNING': 1,
 'request_depth_max': 9,
 'response_received_count': 10,
 'scheduler/dequeued': 20,
 'scheduler/dequeued/memory': 20,
 'scheduler/enqueued': 20,
 'scheduler/enqueued/memory': 20,
 'start_time': datetime.datetime(2023, 2, 24, 21, 26, 56, 323182)}
2023-02-24 23:27:19 [scrapy.core.engine] INFO: Spider closed (finished)

Process finished with exit code 0

```

</details>

It leads to inconsistency between ` 'scheduler/enqueued': 20`, and  ` 'downloader/response_count': 10,` in result stats

GeorgeA92: Recent updates resolves this issue. Now it works.. as expected

GeorgeA92 (approved):

zuozhehao:
any progress on this？

Gallaecio:
https://github.com/scrapy/scrapy/pull/5015#issuecomment-1277663951 remains relevant.

sebastian-correa:
Hi @Gallaecio @WinterComes, do any of you have any updates on an ETA for this feature? It seems the discussion has cooled a bit, but the feature was ready for prime time (or almost there). Are there plans to finish the implementation?

In our case, we want to control the total scrape time a bit better (for a spider that only crawls the start requests but hits an API that might recognize the high load and take countermeasures, so we want to enqueue them "every ~10 minutes" from when we `yield` the `Request`s). I recognize this isn't exactly what Scrapy was designed for, but we're using it for other spiders and it's just this one that needs this behavior. We can make do with per-domain (i.e., per slot, probably as @Gallaecio suggested [here](https://github.com/scrapy/scrapy/issues/802#issuecomment-680049056)) delays or even with `DOWNLOAD_DELAY`s and randomization, but it'd be better to have better fine-grained control (though in our case we don't even need to guarantee the delay is followed to a tee).

wRAR:
> #5015 (comment) remains relevant.

I wonder which comment did you actually mean, as the current link looks wrong :)

Gallaecio:
I don’t recall either.

I would also like to review the PR again, now that I have had experience with the retry policies of https://github.com/zytedata/python-zyte-api. If the implementation is the way I recall, it might not be a good solution for the intended use cases (exponential backoffs for 429 and similar responses).

PR #5019, opened 2021-03-02
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort I noticed that scrapy could not be built reproducibly

This is due to the documentation embedding the current build year in the generated files, therefore making the build vary depending on when you build it. The fix is to use [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) if it is exported to the surrounding environment.

I originally filed this in Debian as bug [#983852](https://bugs.debian.org/983852).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,6 +9,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import os
+import time
 import sys
 from datetime import datetime
 from os import path
@@ -49,7 +51,10 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'Scrapy'
-copyright = f'2008–{datetime.now().year}, Scrapy developers'
+build_date = datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
+copyright = f'2008\u2013{build_date.year}, Scrapy developers'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,7 +54,7 @@ project = 'Scrapy'
 build_date = datetime.utcfromtimestamp(
     int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
 )
-copyright = f'2008\u2013{build_date.year}, Scrapy developers'
+copyright = f'2008–{build_date.year}, Scrapy developers'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Auto-updating copyright at build time is a bad practice anyway, because building today's software next year, would produce `copyright 2008–2022, Scrapy developers` without any developer doing anything in 2022.

Gallaecio:
I agree that it’s a bad practice, only better than not updating it at all.

The best I can think of at the moment is to go for a pre-commit hook that updates the year if it is out of date. But then you need to get contributors to install pre-commit.

Alternatively, we could use https://github.com/c4urself/bump2version/issues/133 once implemented to make sure we update the year before every release.

docs/conf.py
```diff
@@ -49,7 +51,10 @@
 
 # General information about the project.
 project = 'Scrapy'
-copyright = f'2008–{datetime.now().year}, Scrapy developers'
+build_date = datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
+copyright = f'2008\u2013{build_date.year}, Scrapy developers'
```
Gallaecio, line 57:
```suggestion
copyright = f'2008–{build_date.year}, Scrapy developers'
```

docs/conf.py
```diff
@@ -9,6 +9,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import os
+import time
 import sys
```
Gallaecio, line 14:
```suggestion
import os
import sys
import time
```

Gallaecio:
Closing and reopening to re-run the tests with the latest changes from `master`…

docs/conf.py
```diff
@@ -49,7 +51,10 @@
 
 # General information about the project.
 project = 'Scrapy'
-copyright = f'2008–{datetime.now().year}, Scrapy developers'
+build_date = datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
+copyright = f'2008–{build_date.year}, Scrapy developers'
```
jayaddison, line 57:
```suggestion
copyright = f'2008–2022, Scrapy developers'
```

... plus removing the `datetime`, `os`, and `time` imports from this configuration module.

(rationale: it's nice to automate where possible -- and _authorship_ doesn't expire -- but copyright does, and figuring that out (and updating notices like this) usually benefits from human thought and discussion)

jayaddison: Although ultimately I decided that it wasn't worth keeping open as a feature request, I'll mention for completeness/connection that I filed a feature request at sphinx-contrib/sphinx-lint#53 to implement a lint checking rule to test for copyright information in Sphinx configuration files.

Something like that, I reasoned initially, would allow continuous integration to check for the presence of outdated and static copyright lines written by human authors.

Issue #4969, opened 2021-01-30
 It would be nice to get the amount of bandwidth used/min along with pages/min


maxwill-max:
It would be nice to get the amount of bandwidth used/min along with pages/min including for scrapy-splash

amarhod:
Hey! Would like to attempt implementing this feature. Is it okay @Gallaecio?

Gallaecio:
@amarhod It’s OK by me, and if the patch is easy enough, it may be better to discuss over it.

I’m assuming we are ignoring the “scrapy-splash”. @maxwill-max I imagine by that you meant to include the network input and output of the Splash instance(s) used by Scrapy. I think that should not be a part of this feature, at least not of its initial implementation. And I think that part, if implemented, should be implemented in `scrapy-splash`, only making changes to Scrapy as needed to support that downstream change.

amarhod:
@Gallaecio Right now the output is:
```
2021-03-04 10:50:38 [scrapy.extensions.logstats] INFO: Crawled 46 pages (at 2760 pages/min), scraped 0 items (at 0 items/min), used 78 KB of bandwidth (at 78 KB/s)
2021-03-04 10:50:39 [scrapy.extensions.logstats] INFO: Crawled 71 pages (at 1500 pages/min), scraped 0 items (at 0 items/min), used 127 KB of bandwidth (at 49 KB/s)
```
It was generated similarly to how pages and items (and their rates) are calculated.
Have I interpreted the feature correctly and are there other requirements needed to enable the feature in `scrapy-splash`?

amarhod:
Also, would it be meaningful to add these assertions to [BenchCommandTest](https://github.com/scrapy/scrapy/blob/91a8451c981fef25b0915fa06a4566dc880b6e16/tests/test_commands.py#L756)?
```
self.assertIn('pages/min', log)
self.assertIn('items/min', log)
self.assertIn('KB of bandwidth', log)
```

Gallaecio:
I think we should forget about scrapy-splash here.

And I don’t think those asserts help there, they seem unrelated to the purpose of that test. I don’t think we have any test for log stats at the moment. It would be great to have one that makes sure the amount of KB displayed is as expected; but if you find it overwhelming to create such a test, let me know, as I can’t think of any existing tests that you can use as a starting point.

We will also need to mention this in the documentation, and I think we should take the chance to make it clear what this number measures, because it may not be clear (I assume it does not measure pure download size, but instead total response size).

[...]

PR #5029, opened 2021-03-06
Added response bytes as a metric to LogStats extension (#4969)

amarhod:
Modified the Log Stats extension to include response byte (both total amount and rate) in the logged message by reading data from `stats`, more specifically the `downloader/response_bytes` value.
If the setting DOWNLOADER_STATS is disabled, the amount and rate (defaulted to 0) will not be included in the message that is being logged. 
Example:
When `DOWNLOADER_STATS = True`:
```
2021-03-06 12:20:20 [scrapy.extensions.logstats] INFO: Crawled 78 pages (at 4680 pages/min), scraped 0 items (at 0 items/min), received 139 KB of response bytes (at 139 KB/s)
```
When `DOWNLOADER_STATS = False`:
```
2021-03-06 12:20:53 [scrapy.extensions.logstats] INFO: Crawled 98 pages (at 5880 pages/min), scraped 0 items (at 0 items/min)
```

### DIFF
--- a/docs/topics/extensions.rst
+++ b/docs/topics/extensions.rst
@@ -162,7 +162,7 @@ Log Stats extension
 
 .. class:: LogStats
 
-Log basic stats like crawled pages and scraped items.
+Log basic stats like crawled pages, scraped items and received responde bytes.
 
 Core Stats extension
 ~~~~~~~~~~~~~~~~~~~~
--- a/scrapy/extensions/logstats.py
+++ b/scrapy/extensions/logstats.py
@@ -30,25 +30,24 @@ class LogStats:
     def spider_opened(self, spider):
         self.pagesprev = 0
         self.itemsprev = 0
-        self.bandwidthprev = 0.0
+        self.response_bytesprev = 0.0
         self.task = task.LoopingCall(self.log, spider)
         self.task.start(self.interval)
 
     def log(self, spider):
         items = self.stats.get_value('item_scraped_count', 0)
         pages = self.stats.get_value('response_received_count', 0)
-        bandwidth = self.stats.get_value('downloader/response_bytes', 0) / float(1000)
+        response_bytes = self.stats.get_value('downloader/response_bytes', 0) / float(1000)
         irate = (items - self.itemsprev) * self.multiplier
         prate = (pages - self.pagesprev) * self.multiplier
-        brate = (bandwidth - self.bandwidthprev)
-        self.pagesprev, self.itemsprev, self.bandwidthprev = pages, items, bandwidth
-
+        brate = (response_bytes - self.response_bytesprev)
+        self.pagesprev, self.itemsprev, self.response_bytesprev = pages, items, response_bytes
         msg = ("Crawled %(pages)d pages (at %(pagerate)d pages/min), "
                "scraped %(items)d items (at %(itemrate)d items/min), "
-               "used %(bandwidth)d KB of bandwidth (at %(brate)d KB/s)")
+               "received %(response_bytes)d KB of response bytes (at %(brate)d KB/s)")
         log_args = {'pages': pages, 'pagerate': prate,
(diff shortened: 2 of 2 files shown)


docs/topics/extensions.rst
```diff
@@ -162,7 +162,7 @@ Log Stats extension
 
 .. class:: LogStats
 
-Log basic stats like crawled pages and scraped items.
+Log basic stats like crawled pages, scraped items and received responde bytes.
```
Gallaecio, line 165:
```suggestion
Log basic stats like crawled pages, scraped items, and received response bytes.
```

scrapy/extensions/logstats.py
```diff
@@ -30,21 +30,25 @@ def from_crawler(cls, crawler):
     def spider_opened(self, spider):
         self.pagesprev = 0
         self.itemsprev = 0
-
+        self.response_bytesprev = 0.0
```
Gallaecio, line 33: :lipstick: I think `self.responsebytesprev` would be more in line with the existing code, even if snake case is more Pythoninc.

```suggestion
        self.response_bytesprev = 0.0
```

scrapy/extensions/logstats.py
```diff
@@ -30,21 +30,25 @@ def from_crawler(cls, crawler):
...
         self.task = task.LoopingCall(self.log, spider)
         self.task.start(self.interval)
 
     def log(self, spider):
         items = self.stats.get_value('item_scraped_count', 0)
         pages = self.stats.get_value('response_received_count', 0)
+        response_bytes = self.stats.get_value('downloader/response_bytes', 0) / float(1000)
```
Gallaecio, line 40: :lipstick: I don’t think `float` is needed here, in Python 3 divisions always return a `float`.

Also, a more common way to express literals as floats is to append `.0` to them, so if this was not a division, `1000.0` would probably have been slightly more idiomatic.

scrapy/extensions/logstats.py
```diff
@@ -30,21 +30,25 @@ def from_crawler(cls, crawler):
...
         irate = (items - self.itemsprev) * self.multiplier
         prate = (pages - self.pagesprev) * self.multiplier
-        self.pagesprev, self.itemsprev = pages, items
-
+        brate = (response_bytes - self.response_bytesprev)
+        self.pagesprev, self.itemsprev, self.response_bytesprev = pages, items, response_bytes
         msg = ("Crawled %(pages)d pages (at %(pagerate)d pages/min), "
                "scraped %(items)d items (at %(itemrate)d items/min)")
+        if self.stats.get_value('downloader/response_bytes') is not None:
+            msg = msg + ", received %(response_bytes)d KB of response bytes (at %(brate)d KB/s)"
```
Gallaecio, line 48:
```suggestion
            msg = msg + ", received %(response_bytes)d KB of response bytes (at %(brate)d KB/min)"
```

scrapy/extensions/logstats.py
```diff
@@ -30,21 +30,25 @@ def from_crawler(cls, crawler):
...
         self.itemsprev = 0
-
+        self.responsebytesprev = 0.0
         self.task = task.LoopingCall(self.log, spider)
         self.task.start(self.interval)
 
     def log(self, spider):
         items = self.stats.get_value('item_scraped_count', 0)
         pages = self.stats.get_value('response_received_count', 0)
+        response_bytes = self.stats.get_value('downloader/response_bytes', 0.0) / 1000
```
Gallaecio, line 40: I wonder if maybe we should show bytes instead of kilobytes. I mean, given the stats are in bytes, and that both pages and items are integers, it seems more consistent to me to use the byte count in the message as well.

Gallaecio (approved):
Thanks!

Any chance you can also add a test for the change? It may not be easy if there are no tests for the other 2 stats already.

PR #5033, opened 2021-03-09
Rename the macOS workflow, to bypass it’s disabling in GitHub Actions settings

### LINKED ISSUE #5032: Fix and re-enable tests on macOS
We’ve decided to disable tests in macOS because we were often getting random tests failures there from one of the jobs.

Still, we aim to support macOS, so we need to figure out how to fix those issues, and re-enable tests on macOS.

Gallaecio:
Fixes #5032 (that’s the goal anyway, it’s far from done)


Gallaecio:
Finally got a test failure after 3 runs with all tests passing. The failure affected `Https10TestCase.test_protocol`.

Based on https://blackjml.livejournal.com/23029.html, I believe the issue here is that we need to close the connection, which at least for HTTP/1.0 it looks like we are not doing.

https://github.com/scrapy/scrapy/pull/5033/commits/be2d1b8d9548a3a80a72e13c1476bf64e0fc5f46 does the call to `disconnect()`, but I did not manage to get `HTTP10DownloadHandler.close` to wait for that, so I’m not completely sure these changes will be enough.

The way I tried to wait was to create a deferred, set `factory.onConnectionLost = that_deferred`, keeping the deferred in a weak set, and returning `gatherResults(list(deferred_weak_set))` from `close`, but that would cause the test to hang.

Gallaecio:
New test failures:

```
AnonymousFTPTestCase.test_ftp_download_path_with_spaces: DirtyReactorAggregateError, delayed calls: TimeoutMixin.__timedOut()
Https2InvalidDNSPattern.test_download_with_maxsize: DirtyReactorAggregateError, selectables: TLSMemoryBIOProtocol
Https2WrongHostnameTestCase.test_custom_content_length_good: DirtyReactorAggregateError, selectables: TLSMemoryBIOProtocol
```

PR #5034, opened 2021-03-09
Support Content-Length header in responses

### LINKED ISSUE #5009: Content-Length header missing in response headers
### Description

The `Content-Length` header missing in the response headers. I stumbled upon this while working on #4897.

### Steps to Reproduce

```
(...)
>>> response.headers["Content-Length"]
  File "/.../scrapy/scrapy/http/headers.py", line 40, in __getitem__
    return super().__getitem__(key)[-1]
  File "/.../scrapy/scrapy/utils/datatypes.py", line 23, in __getitem__
    return dict.__getitem__(self, self.normkey(key))
KeyError: b'Content-Length'
```

or

```python
import scrapy

class ContentLengthSpider(scrapy.Spider):
    name = "foo"
    start_urls = ["https://example.org"]

    def parse(self, response):
        print(response.headers["Content-Length"])
```

### Versions

```
Scrapy       : 2.4.1
lxml         : 4.6.2.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 20.3.0
Python       : 3.8.2 (default, Apr 18 2020, 17:39:30) - [GCC 7.5.0]
pyOpenSSL    : 20.0.1 (OpenSSL 1.1.1i  8 Dec 2020)
cryptography : 3.4.5
Platform     : Linux-4.15.0-128-generic-x86_64-with-glibc2.27
```

### Additional context

I can see the server does return the header with `cURL`:
```
$ curl -s --http1.1 -D - https://example.org -o /dev/null | grep Content-Length
Content-Length: 1256
```
and `python-requests`:

```python
>>> import requests
>>> requests.get("https://example.org", headers={"Accept-Encoding": "identity"}).headers["Content-Length"]
'1256'
```

It seems to me like Twisted itself is dropping the header:
```python
from twisted.web.client import Agent

agent = Agent(reactor)
d = agent.request(b"GET", b"http://example.org")

def print_response(response):
    print(response.version)
    print(response.headers)

d.addCallback(print_response)
d.addCallback(lambda _: reactor.stop())

reactor.run()
```

```
(b'HTTP', 1, 1)
Headers({b'age': [b'164125'], b'cache-control': [b'max-age=604800'], b'content-type': [b'text/html; charset=UTF-8'], b'date': [b'Wed, 24 Feb 2021 14:38:05 GMT'], b'etag': [b'"3147526947+ident"'], b'expires': [b'Wed, 03 Mar 2021 14:38:05 GMT'], b'last-modified': [b'Thu, 17 Oct 2019 07:18:26 GMT'], b'server': [b'ECS (mic/9ABB)'], b'vary': [b'Accept-Encoding'], b'x-cache': [b'HIT']})
```

I wanted to ask here before opening an issue in Twisted, because this seems like a rather odd thing to do and I'm wondering if I'm missing something :thinking:

Gallaecio:
Fixes #5009

To do:
- [ ] Stop relying on private parts of the Twisted API. It will require either getting the issue fixed upstream or copying a lot of private code from Twisted into Scrapy.
- [ ] Remove the documentation note added at #5045, assuming it was merged by the time this is merged. Otherwise, close #5009. Note, however, that the test approach in #5045 is better than here (it covers HTTP/1.0 as well, where the test passes).

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -10,13 +10,28 @@ from time import time
 from urllib.parse import urldefrag
 
 from twisted.internet import defer, protocol, ssl
+from twisted.internet.defer import CancelledError, Deferred, fail, maybeDeferred
 from twisted.internet.endpoints import TCP4ClientEndpoint
 from twisted.internet.error import TimeoutError
 from twisted.python.failure import Failure
-from twisted.web.client import Agent, HTTPConnectionPool, ResponseDone, ResponseFailed, URI
+from twisted.web.client import (
+    Agent,
+    _HTTP11ClientFactory,
+    HTTPConnectionPool,
+    ResponseDone,
+    ResponseFailed,
+    URI,
+)
 from twisted.web.http import _DataLoss, PotentialDataLoss
 from twisted.web.http_headers import Headers as TxHeaders
 from twisted.web.iweb import IBodyProducer, UNKNOWN_LENGTH
+from twisted.web._newclient import (
+    HTTP11ClientProtocol,
+    HTTPClientParser,
+    RequestGenerationFailed,
+    RequestNotSent,
+    TransportProxyProducer,
+)
 from zope.interface import implementer
 
 from scrapy import signals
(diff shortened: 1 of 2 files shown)


Gallaecio:
To do this without relying on private parts of the Twisted API,  think we would need to basically copy the whole `twisted.web._newclient` module into Scrapy, and probably some more code.

Before I go that way, I’ll try the upstream approach: send a patch to Twisted to stop skipping connection headers. If they are OK with that, we would only need to mention this in the documentation (connection headers in HTTP1.1 require Twisted N.N.N, and are otherwise stripped).

Gallaecio:
Upstream issue: https://twistedmatrix.com/trac/ticket/10126

I have a patch ready locally, only need to update test expectations, but [I don’t know how to run the tests](https://twistedmatrix.com/trac/ticket/10128) :sweat_smile:

Gallaecio:
Pull request sent upstream: https://github.com/twisted/twisted/pull/1561

exarkun:
Twisted doesn't really need to change for this, nor does Scrapy need to copy the whole client implementation.

Instead, Scrapy can check the response object for the `length` attribute.  If it is not `None`, Scrapy can synthesize its own Content-Length for whoever wants it (if it is `None` that is because there was no `Content-Length` header in the response).

Gallaecio:
@exarkun Is `response.length` the `Content-Length` value, or the actual length of the response?

Gallaecio:
I’ve just created https://github.com/scrapy/scrapy/pull/5057 to handle this on the Scrapy side using the public Twisted API.

@exarkun Should we close https://github.com/twisted/twisted/pull/1561 and https://twistedmatrix.com/trac/ticket/10126 ? While I still see no good reason to strip connection headers from responses, I cannot think of real-life scenarios where the other connection headers are relevant, or where the position of the `Content-Length` header among response headers is relevant (I assume https://github.com/twisted/twisted/pull/1561 would also expose that).

exarkun:
> @exarkun Should we close twisted/twisted#1561 and https://twistedmatrix.com/trac/ticket/10126 ? While I still see no good reason to strip connection headers from responses, I cannot think of real-life scenarios where the other connection headers are relevant, or where the position of the Content-Length header among response headers is relevant (I assume twisted/twisted#1561 would also expose that).

I reckon it should be closed, yes, unless there is another use-case to consider.

PR #5036, opened 2021-03-12
UrlLengthMiddleware: don't skip silently

dmiwell:
I guess it's a good idea to not just send the skipped info to debug, since it could be critical and not so obvious to debug in case of accidental long url generating.

So it's better to log to `INFO` at least, and add count to crawler stats.

### DIFF
--- a/scrapy/spidermiddlewares/urllength.py
+++ b/scrapy/spidermiddlewares/urllength.py
@@ -14,17 +14,15 @@ logger = logging.getLogger(__name__)
 
 class UrlLengthMiddleware:
 
-    def __init__(self, maxlength, stats):
+    def __init__(self, maxlength):
         self.maxlength = maxlength
-        self.stats = stats
 
     @classmethod
-    def from_crawler(cls, crawler):
-        settings = crawler.settings
+    def from_settings(cls, settings):
         maxlength = settings.getint('URLLENGTH_LIMIT')
         if not maxlength:
             raise NotConfigured
-        return cls(maxlength, crawler.stats)
+        return cls(maxlength)
 
     def process_spider_output(self, response, result, spider):
         def _filter(request):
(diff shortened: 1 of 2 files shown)


wRAR (approved):

scrapy/spidermiddlewares/urllength.py
```diff
@@ -14,22 +14,27 @@
 
 class UrlLengthMiddleware:
 
-    def __init__(self, maxlength):
+    def __init__(self, maxlength, stats):
```
Gallaecio, line 17: Can we make the parameter optional, to maintain backward compatibility? (think of class inheritance and calls to `super().__init__(maxlength)`).

scrapy/spidermiddlewares/urllength.py
```diff
@@ -14,22 +14,27 @@
...
         self.maxlength = maxlength
+        self.stats = stats
 
     @classmethod
-    def from_settings(cls, settings):
+    def from_crawler(cls, crawler):
```
Gallaecio, line 22: I find it hard to figure out how to handle this for backward compatibility (and this is not the first time I face this issue), but I know this is backward incompatible, because [if a subclass defines `from_settings`, that `from_settings` will no longer be called due to this `from_crawler`](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/utils/misc.py#L145).

dmiwell: @Gallaecio we might not inject `stats` instance to a mw, but instead access it by `spider.crawler.stats` so no need to change signatures, and backward compatibility will be kept. 

Is that fine?

Gallaecio: Sounds great to me, it removes all the backward compatibility headaches :slightly_smiling_face:

Gallaecio (approved):

Issue #4697, opened 2020-07-27
Warn about br handling if brotlipy is not installed


Gallaecio:
Currently, the HTTP compression middleware sets the `Accept-Encoding` header to `gzip,deflate` or `gzip,deflate,br` depending on whether or not [brotlipy](https://pypi.org/project/brotlipy/) is installed.

However, if the user manually sets the `Accept-Encoding` header to a value containing `br`, and does not have `brotlipy` installed, the server can send a response compressed with `br` and Scrapy will silently fail to decompress it, so the response content will not be readable.

I think we should improve a bit the handling of this scenario.

If the user sets `br` manually on the `Accept-Encoding` header, I think we should:

- Log a warning if `brotlipy` is not installed.
- Have the middleware cause an error if the server sends the response br-encoded, and `brotlipy` is not installed.
- Log an information message the first time that the user sends a response with a manually-set `Accept-Encoding` header which is identical to the one that the HTTP compression middleware would define. It should not be a warning because the user does not need to do anything about it. It should only be displayed once because there is a chance the user is doing it on purpose (e.g. if the default headers define a different value but the user wishes to override that for some requests with the default value.

The 3rd bullet point may not be that useful, but I think the 1st and 2nd are.

GGSheridan:
Should this issue be labeled as Good First Issue?

We are doing research about how to help newcomers onboarding in an open source project and how to accelerate the issue resolving process.
We have developed a tool to automatically label the Good First Issue (GFI), i.e., the issue which is good for newcomers. 
We first extract 72 features from different dimensions of issue reports, such as readability, semantic meaning, associated code snippets, etc, and experimentally investigate to what extent these features correlate with GFI. Results turn out that a large portion of these features do have a medium to large correlation with GFI. 
We then build a machine learning model based on these features to predict GFI. The experiments on a dataset of 12k issues from Github show that 0.90 AUC can be achieved. 

Our tool shows that this issue has a very large possibility to be a GFI. 
Could you please tell us whether it should be labeled as a GFI?
And if you have any comments or suggestions with this tool, could you please tell us? 
Thank you so much.

Gallaecio:
That tool sounds very interesting!

Regarding this specific issue, it could be argued that it is indeed a good first issue. I’m unsure, though.

Dushatar:
Hello. Me and two others would like to take a look at this issue as part of a course we are taking @Gallaecio 

Hopefully we can make some progress in the, not so long, time allocated to the assignment.

[...]

PR #5037, opened 2021-03-12
Brotli import error #4697

Anupam-USP:
I tried to solve #4697 but if there is some problem please do let me know. Also this is my first pull request so if there is any formatting error, please correct me!


scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -73,7 +75,11 @@ def _decode(self, body, encoding):
                 # http://www.gzip.org/zlib/zlib_faq.html#faq38
                 body = zlib.decompress(body, -15)
         if encoding == b'br' and b'br' in ACCEPTED_ENCODINGS:
-            body = brotli.decompress(body)
+            try:
+                body = brotli.decompress(body)
+            except ImportError:
```
wRAR, line 80: This exception cannot happen here as far as I know, can it?

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -8,12 +9,13 @@
...
 
 ACCEPTED_ENCODINGS = [b'gzip', b'deflate']
+logger = logging.getLogger(__name__)
 
 try:
     import brotli
     ACCEPTED_ENCODINGS.append(b'br')
 except ImportError:
-    pass
+    logger.error("brotli not installed")
```
wRAR, line 18: We shouldn't log an error here as this is not a problem

wRAR (changes requested):

scrapy/downloadermiddlewares/httpcompression.py
Anupam-USP: Ok I thought this is an Import error, should I log a warning then?

wRAR: We still support configurations without brotli, please read the description of the needed changes in the original issue.

scrapy/downloadermiddlewares/httpcompression.py
Anupam-USP: I think here I have to make changes here and leave rest as it is, and if not installed than give a warning? Will it do the work?

wRAR: Again, please read the instructions in the issue:

> If the user sets br manually on the Accept-Encoding header, I think we should:
> 
> * Log a warning if brotlipy is not installed.
> * Have the middleware cause an error if the server sends the response br-encoded, and brotlipy is not installed.

So you need to raise an error here, it brotli is not installed, and in addition to this you need to log a warning if brotli is not installed but ACCEPTED_ENCODINGS contains 'br' (not sure which is the best place for this).

Anupam-USP: Screenshot 2021-03-12 201909

I think this much change can handle the issue?

wRAR: As we are already catching ImportError in this file, you can just set a flag there and check it here.

Also, I'm not sure whether logging an error and returning the compressed body or raising an exception instead is a better way to handle this.

Anupam-USP:
I am not able to push my change further, it is showing that `This branch is 3 commits ahead, 6 commits behind scrapy:master.` Should I close this pull request and create a new one or is there any other way? I have tried [this](https://stackoverflow.com/questions/41283955/github-keeps-saying-this-branch-is-x-commits-ahead-y-commits-behind) solution.

wRAR:
No, you shouldn't need to make new PRs if you just created additional commits in the same branch.

Anupam-USP:
Actually I tried pushing with my master and it is recognized in #4697 conversation. In this branch I got the above error and not able to push

Anupam-USP:
I am really sorry, this was my first one so I got really confused. Should I close this one and create a fresh?

PR #5038, opened 2021-03-12
br handling 

### LINKED ISSUE #4697: Warn about br handling if brotlipy is not installed

Anupam-USP:
Resolves #4697 . 

Sorry, being the new contributor after having some issues i panic and deleted my fork to start afresh. Really sorry for the trouble. Last pr was #5037 .

### DIFF
--- a/scrapy/commands/bench.py
+++ b/scrapy/commands/bench.py
@@ -50,7 +50,7 @@ class _BenchSpider(scrapy.Spider):
 
     def start_requests(self):
         qargs = {'total': self.total, 'show': self.show}
-        url = f'{self.baseurl}?{urlencode(qargs, doseq=1)}'
+        url = f'{self.baseurl}?{urlencode(qargs, doseq=True)}'
         return [scrapy.Request(url, dont_filter=True)]
 
     def parse(self, response):
--- a/tests/spiders.py
+++ b/tests/spiders.py
@@ -45,7 +45,7 @@ class FollowAllSpider(MetaSpider):
         self.urls_visited = []
         self.times = []
         qargs = {'total': total, 'show': show, 'order': order, 'maxlatency': maxlatency}
-        url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=1)}")
+        url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=True)}")
         self.start_urls = [url]
 
     def parse(self, response):
@@ -245,7 +245,7 @@ class BrokenStartRequestsSpider(FollowAllSpider):
 
         for s in range(100):
             qargs = {'total': 10, 'seed': s}
-            url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=1)}")
+            url = self.mockserver.url(f"/follow?{urlencode(qargs, doseq=True)}")
             yield Request(url, meta={'seed': s})
             if self.fail_yielding:
                 2 / 0


Anupam-USP:
Why does these checks  failed when I used simple print? Logging a warning passed all of these

Anupam-USP:
Can you please check why the test failed? I think the first time I used to log warning and this time I simply printed the statement which can be the error

Gallaecio:
The errors seem unrelated to your changes, and caused by internal issues in the GitHub infrastructure. I’ll re-run them, let’s hope the GitHub issues have been fixed.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -10,10 +10,13 @@
 ACCEPTED_ENCODINGS = [b'gzip', b'deflate']
 
 try:
+    flag=0
     import brotli
     ACCEPTED_ENCODINGS.append(b'br')
 except ImportError:
     pass
+else:
+    flag=1
```
Gallaecio, line 19: I don’t think you need a new variable here, you could just use `b'br' in ACCEPTED_ENCODINGS` to the same effect, which would be more readable.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -73,7 +76,10 @@ def _decode(self, body, encoding):
                 # http://www.gzip.org/zlib/zlib_faq.html#faq38
                 body = zlib.decompress(body, -15)
         if encoding == b'br' and b'br' in ACCEPTED_ENCODINGS:
-            body = brotli.decompress(body)
+            if (flag):
```
Gallaecio, line 79: The `b'br' in ACCEPTED_ENCODINGS` in the `if` statement above already implies that `flag` is `0` here. So `flag` is unnecessary here.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -73,7 +76,10 @@ def _decode(self, body, encoding):
...
+                body = brotli.decompress(body)
+            else:
+                print("brotli not installed")
```
ejulio, line 82: It should be a warning here, not a simple `print`.
To some extent, it could even be an error as it will fail when processing the response, as it wasn't decoded properly.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -73,7 +76,10 @@ def _decode(self, body, encoding):
...
         if encoding == b'zstd' and b'zstd' in ACCEPTED_ENCODINGS:
```
ejulio, line 83: You should apply the same condition as above here, as it falls in the same case/issue (not having the dependency)

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -10,10 +10,13 @@
 ACCEPTED_ENCODINGS = [b'gzip', b'deflate']
```
ejulio, line 10: You can change this to a `set` as we're performing many `in` (pertinence checks) and this operation is faster on `set` than on a `list`

scrapy/downloadermiddlewares/httpcompression.py
ejulio: I think the condition here needs to be
```
if encoding == b'br':
    if b'br' in ACCEPTED_ENCODINGS:
        body = brotli.decompress(body)
    else:
        # warning or error
```

ejulio: Also, I believe we'll need a test for this, to ensure we get the message when required

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -10,10 +10,13 @@
...
 
 try:
+    flag=0
     import brotli
     ACCEPTED_ENCODINGS.append(b'br')
 except ImportError:
     pass
+else:
+    flag=1
```
ejulio, line 20: Just taking the chance here.
You could change `encoding = content_encoding.pop()` to `encoding = content_encoding[0]` in `process_response`, as this removes the header when it arrives in the spider and can make our life a bit hard to detect the issue.

Not sure if it follows here or as a fix at all.
cc @Gallaecio

Gallaecio: I see what you mean, [line 48](https://github.com/scrapy/scrapy/blob/35850f12207991eb815a5d39aaa7a5bc367e7d04/scrapy/downloadermiddlewares/httpcompression.py#L48), and yes, I think you are right.

I would not require this change in this pull request, but if not done here, we should probably open a ticket for it.

PR #5042, opened 2021-03-16
Progress bar extension

### LINKED ISSUE #3793: Progress bar for large downloads

elacuesta:
Closes #3793.

Very early stage. It basically works, but it needs more work, I don't like the way it's interacting with the logs. 

I don't really have a strong opinion as to whether or not this should be included, even if disabled by default, perhaps it could be better to have it as a separate package.


johtso:
Interested in this!

Would also be good to be able to have a scrape-wide progress bar for when you know how many items you expect to get.

PR #5045, opened 2021-03-18
Response headers documentation

Gallaecio:
Related to #5009


wRAR (approved):

Gallaecio:
We should probably only merge this change if we do not merge #5057 in time for the next Scrapy release.

PR #5058, opened 2021-03-21
[CI] update pip

elacuesta:
Trying to see if #5059 is reproducible in CI


Gallaecio:
In CI we use constraints to limit backtracking. But [I’ve had to downgrade `pip` to a version before the new dependency parser to get ReadTheDocs to build](https://github.com/scrapy/scrapy/pull/5052).

PR #5062, opened 2021-03-22
[HTTP/1.1] Skip Content-Length header if its value is UNKNOWN_LENGTH

elacuesta:
I wasn't fast enough to comment this on #5057 before it got merged.

The following happens with the current `master` (ec5a7918ec2d8888ba356f9e3295dcd1ee935884):
```python
$ scrapy shell http://quotes.toscrape.com
(...)
2021-03-22 10:42:40 [scrapy.core.engine] INFO: Spider opened
resp2021-03-22 10:42:41 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://quotes.toscrape.com> (referer: None)
onse.2021-03-22 10:42:41 [asyncio] DEBUG: Using selector: EpollSelector
[s] Available Scrapy objects:
[s]   crawler    <scrapy.crawler.Crawler object at 0x7f24d38fc790>
[s]   item       {}
[s]   request    <GET http://quotes.toscrape.com>
[s]   response   <200 http://quotes.toscrape.com>
[s]   settings   <scrapy.settings.Settings object at 0x7f24d38fcee0>
[s]   spider     <DefaultSpider 'default' at 0x7f24d2b792e0>
[s] Useful shortcuts:
2021-03-22 10:42:41 [asyncio] DEBUG: Using selector: EpollSelector
In [1]: response.headers["Content-Length"]
Out[1]: b'twisted.web.iweb.UNKNOWN_LENGTH'
```

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -361,10 +361,9 @@ class ScrapyAgent:
     @staticmethod
     def _headers_from_twisted_response(response):
         headers = Headers()
-        if response.length is not None:
+        if response.length is not None and response.length != UNKNOWN_LENGTH:
             headers[b'Content-Length'] = str(response.length).encode()
-        for key, value in response.headers.getAllRawHeaders():
-            headers[key] = value
+        headers.update(response.headers.getAllRawHeaders())
         return headers
 
     def _cb_bodyready(self, txresponse, request):
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -361,7 +361,7 @@ class ScrapyAgent:
     @staticmethod
     def _headers_from_twisted_response(response):
         headers = Headers()
-        if response.length is not None and response.length != UNKNOWN_LENGTH:
+        if response.length != UNKNOWN_LENGTH:
             headers[b'Content-Length'] = str(response.length).encode()
         headers.update(response.headers.getAllRawHeaders())
         return headers


Gallaecio:
What about adding a test for it? Also, do we even need to compare to `None`?

elacuesta:
I wouldn't risk not checking for `None`, the attribute is not documented and so there's no guarantee. Worst case, we could end up with stuff like
```python
>>> str(None).encode()
b'None'
```

Let me see what I can do regarding the tests.

elacuesta:
An alternative could be 
```python
with suppress(ValueError, TypeError):
    headers[b'Content-Length'] = str(int(response.length)).encode()
```

Gallaecio:
[It _is_ documented](https://twistedmatrix.com/documents/21.2.0/api/twisted.web.iweb.IResponse.html#length). Which makes the fact that my implementation is wrong that much worse :sweat:

elacuesta:
Oh, is that new? In that case, I upvote my suggestion from https://github.com/scrapy/scrapy/pull/5062#issuecomment-804183146 :sweat_smile:

Regarding the test, I'm having a hard time getting it done. I tried [this](https://stackoverflow.com/a/39766415) but it didn't work for `Content-Length`, seems like Twisted injects the correct integer value even if I set a different one. I also tried `mock.patch`, to no avail.

Gallaecio:
It does not surprise me that Twisted sets the Content-Length itself. No test then.

elacuesta:
True, no `TypeError` if the attribute could not be `None`. Let's scratch that, then.

Gallaecio (approved):

PR #5063, opened 2021-03-22
Update UsageError message

noviluni:
This message was added 10 years ago (`Scrapy 0.25.1`), and I think that we can update it like this, as the feature of running more than one spider was deprecated a long time ago.

However, as this appears when we write more than one `arg`, it could be also a good idea to reformulate the whole message to give a hint like _"did you forget to add `-a` / `-s` , etc"_. Or something similar. 

Let me know what you think. :)

### DIFF
--- a/scrapy/commands/crawl.py
+++ b/scrapy/commands/crawl.py
@@ -16,7 +16,7 @@ class Command(BaseRunSpiderCommand):
         if len(args) < 1:
             raise UsageError()
         elif len(args) > 1:
-            raise UsageError("running 'scrapy crawl' with more than one spider is no longer supported")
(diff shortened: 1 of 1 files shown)


elacuesta (approved):

noviluni:
I see that it could be a good idea to add a test too.

Gallaecio (approved):

Gallaecio:
> I see that it could be a good idea to add a test too.

Do you want to give it a try as part of this pull request, or shall we merge and create a separate ticket for it?

kmike:
Let's handle tests separately, and get the improvement in. These tests would be unrelated to the change anyways - you can't write a test which checks that the readability improved :)

PR #5068, opened 2021-03-26
MediaPipeline: log media_to_download errors before stripping them

### LINKED ISSUE #5067: MediaPipeline exceptions passed silently
### Description

MediaPipeline exceptions passed silently both for method body or method signature errors.

### Steps to Reproduce

```
from scrapy.pipelines.files import FilesPipeline
class BuggyFilesPipeline(FilesPipeline):
    def file_path(self, request, response=None, info=None, *, item=None):
        return 1 / 0
```
**Expected behavior:** Exception logged

**Actual behavior:** Exception passed silently

### Versions

Scrapy 2.4

Gallaecio:
Fixes #5067

To do:

- [x] Make sure that existing tests pass
- [x] Add a test for the issue being fixed here


scrapy/pipelines/media.py
```diff
@@ -112,12 +112,18 @@ def _process_request(self, request, info, item):
...
-        )
         return dfd.addBoth(lambda _: wad)  # it must return wad at last
 
+    def _log_error(self, result, info):
+        logger.error(
+            result.value,
+            exc_info=failure_to_exc_info(result),
+            extra={'spider': info.spider}
+        )
+        return result
```
wRAR, line 125: Is this the only actual change?

Gallaecio: Well, the main change is actually moving the `addErrback` 1 line above. Returning `result` here is just so that the line move does not break the `addBoth` line that now goes afterwards.

wRAR: Oh, I see.

wRAR (approved):

scrapy/pipelines/media.py
Gallaecio: The reason for the line move is that `_cache_result_and_execute_waiters` removes the failure information to save memory. So we were basically removing an error and then logging it, now we should be logging it and _then_ removing it.

wRAR (approved):

PR #5070, opened 2021-03-29
Include loading settings in `Running multiple spiders in the same precess` section

ric2b:
The example in the documentation doesn't take into account how to load the project's settings.

I assume most people want to load the settings when running multiple spiders concurrently. If they don't, it's at least more straightforward to remove that functionality than to notice the settings are being ignored and then find out how to fix it.

I had to find it in a third-party example: https://botproxy.net/docs/how-to/scrapy-crawl-multiple-spiders-sharing-same-items-pipeline-and-settings-but-with-separa/

### DIFF
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -118,8 +118,8 @@ Here is an example that runs multiple spiders simultaneously:
 ::
 
     import scrapy
-    from scrapy.utils.project import get_project_settings
     from scrapy.crawler import CrawlerProcess
+    from scrapy.utils.project import get_project_settings
 
     class MySpider1(scrapy.Spider):
         # Your first spider definition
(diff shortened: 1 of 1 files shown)


Gallaecio:
Please, switch the target of this pull request to the `master` branch.

As for the examples themselves, as they seem intended to feature the alternative to `scrapy crawl` to run several spiders, I agree that it makes sense to make this change a part of the example itself. Also, we should probably reproduce the same change in the examples right below, which are meant to be based on the modified example.

ric2b:
Done :)

Btw, just noticed that how to load the settings is mentioned in the previous section. But someone that is just looking for how to run multiple spiders concurrently (like I was) will easily miss it.

Gallaecio (approved):

wRAR (approved):

wRAR:
Thank you!

PR #5072, opened 2021-04-01
[CRAWLERA-6612] Rebranding, updated GA code

zloidemon:


docs/_templates/layout.html
```diff
@@ -3,14 +3,9 @@
 {% block footer %}
 {{ super() }}
 <script type="text/javascript">
-!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="3.1.0";
```
wRAR, line 6: Is this just not needed anymore?

wRAR: "legacy Segment tracking code. I don’t think we use Segment anymore."

wRAR (approved):

docs/_templates/layout.html
zseta: Correct. Only GA code needs to be updated

zseta (approved):

Gallaecio:
@zseta @zloidemon The analytics code seems to fail now due to the removed line, because now the `analytics` JavaScript variable is not defined.

Laerte:
@Gallaecio Should we rollback the code (but keeping new domain)?

I saw this PR:
- https://github.com/scrapy/scrapy.org/issues/224

But this not belong to scrapy.org repo since the issue in here on docs (maybe move to scrapy repo then).

wRAR:
Looks like that issue is both about scrapy.org and docs.scrapy.org (maybe they are even caused by different pieces of code?)

Gallaecio:
Indeed. Let’s create a separate issue to deal with the docs.scrapy.org issue.

PR #5073, opened 2021-04-01
Fix type for urlencode's doseq argument

elacuesta:
Continuation of #4950 (https://github.com/scrapy/scrapy/pull/4950#pullrequestreview-564527188)


kmike:
Thanks for the follow-up @elacuesta!

PR #5077, opened 2021-04-03
Add typing for middleware and coroutine related code.

wRAR:


wRAR:
Typing errors should be fixed by #5076

scrapy/core/downloader/middleware.py
```diff
@@ -29,9 +33,9 @@ def _add_middleware(self, mw):
         if hasattr(mw, 'process_exception'):
             self.methods['process_exception'].appendleft(mw.process_exception)
 
-    def download(self, download_func, request, spider):
+    def download(self, download_func: Callable, request: Request, spider: Spider):
```
kmike, line 36: Do you think you can add output types as well?

scrapy/core/downloader/middleware.py
```diff
@@ -44,7 +48,7 @@ def process_request(request):
             return (yield download_func(request=request, spider=spider))
 
         @defer.inlineCallbacks
-        def process_response(response):
+        def process_response(response: Union[Response, Request]):
             if response is None:
```
kmike, line 52: if respone can be None, shouldn't it be optional in the type?

kmike: Aha, I see, probably it shouldn't :)

scrapy/core/scraper.py
```diff
@@ -3,12 +3,14 @@
 
 import logging
 from collections import deque
+from collections.abc import Iterable
```
kmike, line 6: What's your motivation for importing it from collections.abc, and not from typing?

scrapy/core/spidermw.py
```diff
@@ -62,7 +70,7 @@ def _evaluate_iterable(self, response, spider, iterable, exception_processor_ind
                 raise
             recover_to.extend(exception_result)
 
-    def _process_spider_exception(self, response, spider, _failure, start_index=0):
+    def _process_spider_exception(self, response: Response, spider: Spider, _failure: Failure, start_index=0):
```
kmike, line 73:
```suggestion
    def _process_spider_exception(self, response: Response, spider: Spider, _failure: Failure, start_index: int = 0):
```

scrapy/middleware.py
```diff
@@ -16,7 +17,7 @@ class MiddlewareManager:
 
     def __init__(self, *middlewares):
         self.middlewares = middlewares
-        self.methods = defaultdict(deque)
+        self.methods: dict[str, deque[Callable]] = defaultdict(deque)
```
kmike, line 20: I'm surprised this syntax works (`dict[str, ...]`), it is supposed to be Python 3.9+ feature, or Python 3.7+ with `from future import __annotations__` import

scrapy/core/downloader/middleware.py
wRAR: Actually, I'm not sure how much info should I add when some of it can be inferred from code.

scrapy/middleware.py
wRAR: It works on 3.6 and even catches errors :)

scrapy/core/scraper.py
wRAR: Just a habit, and it doesn't seem to make a difference unless `Iterable[foo]` is needed but I can change this to use `typing`.

scrapy/core/downloader/middleware.py
kmike: I think we should be annotating public interfaces even if the data types can be inferred, so that Scrapy users can check their code against types provided by Scrapy.

For internal code it is a judgment call if an annotation is needed or not. I'd say there are 2 reasons to annotate, and 2 reasons not to annotate: 

* (+) if it makes code more readable / easy to understand. For example, sometimes type can be inferred from the code, but you don't want to spend your time doing this, it is faster to read the definition.
* (+) if it is likely to prevent errors. For example, you may have some assumptions on what the function should be returning, and if you add e.g. output type information, the assumption would be checked by mypy, including cases when refactoring happens.
* (-) type-annotated code is more verbose, so there is an argument not to overuse it
* (-) it requires effort to add type annotations, which can be spent elsewhere; if the code piece is not complex and not public, or maybe not toched much, the effort might not be justified

scrapy/middleware.py
kmike: That's really strange!

```
>>> def foo(x: dict[str, str]): pass
...
TypeError: 'type' object is not subscriptable
```

wRAR: Nice, new PyCharm flags my code and proposes a replacement from `dict` to `typing.Dict`, is this the correct fix?

scrapy/core/spidermw.py
```diff
@@ -3,10 +3,10 @@
 
 See documentation in docs/topics/spider-middleware.rst
 """
-from collections.abc import Iterable
 from itertools import islice
-from typing import Callable, Union, Any
+from typing import Callable, Union, Any, Generator, Iterable
```
Gallaecio, line 7:
```suggestion
from typing import Any, Callable, Generator, Iterable, Union
```

scrapy/middleware.py
```diff
@@ -1,9 +1,13 @@
 import logging
 import pprint
 from collections import defaultdict, deque
-from typing import Callable
+from typing import Callable, Dict, Deque
```
Gallaecio, line 4:
```suggestion
from typing import Callable, Deque, Dict
```

scrapy/utils/defer.py
```diff
@@ -5,7 +5,7 @@
 import inspect
 from collections.abc import Coroutine
 from functools import wraps
-from typing import Callable, Iterable, Any
+from typing import Callable, Iterable, Any, Generator
```
Gallaecio, line 8:
```suggestion
from typing import Any, Callable, Generator, Iterable
```

scrapy/middleware.py
```diff
@@ -1,9 +1,13 @@
...
 
+from twisted.internet import defer
```
Gallaecio, line 6: What about `from twisted.internet.defer import Deferred` as you did in the first file? I would rather use `Deferred` than `defer.Deferred`. I would even add such an import to `scrapy.utils.defer`, which was already importing `defer`.

Gallaecio (approved):
Only :lipstick: comments.

scrapy/middleware.py
wRAR: `scrapy.utils.defer` is harder as it uses `succeed` and `fail` and I don't like aliases or importing these names as is.

Gallaecio: Makes sense to me, but still it may be best to have both imports than to use `defer.Deferred` everywhere.

wRAR: Ah yeah, I didn't think about using both.

PR #5082, opened 2021-04-05
Find bash from PATH instead of /bin/bash

### LINKED ISSUE #5080: Allow building the documentation with make in systems without /bin/bash
https://github.com/scrapy/scrapy/commit/595146e158b1efec432b35128aab59878da81295#r48800250

I’m guessing the current code exists in line with https://stackoverflow.com/a/589300/939364, in order to force Bash to be used (instead of something like `sh`).

If there is a way to set SHELL to Bash so that it finds the right path to Bash regardless of the system, that would be great. Otherwise, maybe we can check if there is any Bash-specific code in this `Makefile`. If all commands work as intended with `sh`, for example, maybe we can just remove this line as suggested.

storymode7:
Instead of hard coding executable path, we can use `/usr/bin/env` 
It searches for bash in $PATH: https://unix.stackexchange.com/a/29620/420806

This fixes #5080

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -8,7 +8,7 @@ PYTHON       = python
 SPHINXOPTS   =
 PAPER        =
 SOURCES      =
-SHELL        = /bin/bash
+SHELL        = /usr/bin/env bash
 
 ALLSPHINXOPTS = -b $(BUILDER) -d build/doctrees \
                 -D latex_elements.papersize=$(PAPER) \


Gallaecio (approved):
It works for me, and I’ve found an [online reference](https://www.cs.cmu.edu/~07131/f20/topics/makefiles/getting-started/) to this usage, so +1!

wRAR (approved):

PR #5084, opened 2021-04-07
docs: require sphinx-rtd-theme>=0.5.2 and the latest pip…

Gallaecio:
…to prevent installing breaking docutils>=0.17

We originally though this might have been caused by https://github.com/scrapy/scrapy/pull/4974, but it turned out it was an upstream issue: https://github.com/readthedocs/readthedocs.org/issues/8070

Once merged, we should backport this to the Scrapy 2.5 branch (no need for a 2.5.1 release, though).


wRAR (approved):

Gallaecio:
Fix confirmed in https://docs.scrapy.org/en/master/

Issue #542, opened 2014-01-17
Requests scheduled when idle never go through the spider middlewares


nside:
When my spider enters the idle state I schedule a request through the engine like this:

```
self.crawler.engine.schedule( req , self)
```

and I expect that request to go through the spider middlewares. But after some investigation it looks like only requests returned by Spider.start_requests and those returned from callbacks are processed by these middlewares.

Maybe I'm scheduling the request in a bad way but if so it shouldn't be public (ie I'd prefix schedule with _).

PR #5090, opened 2021-04-08
Engine: deprecations and type hints

elacuesta:
As discussed earlier @wRAR 

Type hints adapted from #3559

Deprecated methods:
* `ExecutionEngine.open_spiders`: I didn't check the history, but I suspect this probably comes from the time when multiple spiders could run in a single crawl. That hasn't been the case for a long time, so it's not necessary to have a plurally-named property with only one element.
* `ExecutionEngine.has_capacity`: same as the above, the engine only handles one spider slot, which can be checked directly. Not called from outside the engine.
* `ExecutionEngine.schedule`: Not called from outside the engine, deprecated as per https://github.com/scrapy/scrapy/issues/542#issuecomment-32621164

Also, the `spider` argument is deprecated for the `spider_is_idle`, `crawl` and `download` methods, the spider instance saved in the `ExecutionEngine.spider` after calling `open_spider` is used instead. I would also like to deprecate passing a spider argument to `close_spider`, but that would be more complicated because of the order of the positional arguments, let's leave it for a future PR.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -191,7 +191,7 @@ class ExecutionEngine:
         d = self.scraper.enqueue_scrape(result, request, self.spider)
         d.addErrback(
             lambda f: logger.error(
-                'Error while enqueuing downloader output',
+                "Error while enqueuing downloader output",
                 exc_info=failure_to_exc_info(f),
                 extra={'spider': self.spider},
             )
@@ -201,12 +201,12 @@ class ExecutionEngine:
     def spider_is_idle(self, spider: Optional[Spider] = None) -> bool:
         if spider is not None:
             warnings.warn(
-                f"Passing a 'spider' argument to {self.__class__.__name__}.spider_is_idle is deprecated",
+                "Passing a 'spider' argument to ExecutionEngine.spider_is_idle is deprecated",
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
         if self.slot is None:
-            raise RuntimeError(f"{self.__class__.__name__}.slot is not assigned")
+            raise RuntimeError("Engine slot not assigned")
         if not self.scraper.slot.is_idle():
             return False
         if self.downloader.active:  # downloader has pending requests
@@ -221,12 +221,12 @@ class ExecutionEngine:
         """Inject the request into the spider <-> downloader pipeline"""
         if spider is not None:
             warnings.warn(
-                f"Passing a 'spider' argument to {self.__class__.__name__}.crawl is deprecated",
+                "Passing a 'spider' argument to ExecutionEngine.crawl is deprecated",
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
         if self.slot is None:
-            raise RuntimeError(f"{self.__class__.__name__}.slot is not assigned")
+            raise RuntimeError("Engine slot not assigned")
         if self.spider is None:
             raise RuntimeError(f"Spider not opened when crawling: {request}")
         self._schedule_request(request)
@@ -239,10 +239,10 @@ class ExecutionEngine:
             self.signals.send_catch_log(signals.request_dropped, request=request, spider=self.spider)
 
     def download(self, request: Request, spider: Optional[Spider] = None) -> Deferred:
-        """Return a Deferred which fires with a Response as result, only download middlewares are applied"""
+        """Return a Deferred which fires with a Response as result, only downloader middlewares are applied"""
         if spider is not None:
             warnings.warn(
-                f"Passing a 'spider' argument to {self.__class__.__name__}.download is deprecated",
+                "Passing a 'spider' argument to ExecutionEngine.download is deprecated",
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
@@ -307,8 +307,8 @@ class ExecutionEngine:
     def _spider_idle(self) -> None:
         """
         Called when a spider gets idle, i.e. when there are no remaining requests to download or schedule.
-        It can be called multiple times. If some handler for the spider_idle signal raises a DontCloseSpider
-        exception the spider is not closed until the next loop and this function is guaranteed to be called
+        It can be called multiple times. If a handler for the spider_idle signal raises a DontCloseSpider
+        exception, the spider is not closed until the next loop and this function is guaranteed to be called
         (at least) once again.
         """
         assert self.spider is not None  # typing
@@ -322,7 +322,8 @@ class ExecutionEngine:
 
     def close_spider(self, spider: Spider, reason: str = "cancelled") -> Deferred:
         """Close (cancel) spider and clear all its outstanding requests"""
-        assert self.slot is not None  # typing
+        if self.slot is None:
+            raise RuntimeError("Engine slot not assigned")
 
         if self.slot.closing is not None:
             return self.slot.closing
@@ -368,24 +369,23 @@ class ExecutionEngine:
     @property
     def open_spiders(self) -> list:
         warnings.warn(
-            f"{self.__class__.__name__}.open_spiders is deprecated, "
-            f"please use {self.__class__.__name__}.spider instead",
+            "ExecutionEngine.open_spiders is deprecated, please use ExecutionEngine.spider instead",
             category=ScrapyDeprecationWarning,
             stacklevel=2,
         )
         return [self.spider] if self.spider is not None else []
 
     def has_capacity(self) -> bool:
-        warnings.warn(f"{self.__class__.__name__}.has_capacity is deprecated", ScrapyDeprecationWarning, stacklevel=2)
+        warnings.warn("ExecutionEngine.has_capacity is deprecated", ScrapyDeprecationWarning, stacklevel=2)
         return not bool(self.slot)
 
     def schedule(self, request: Request, spider: Spider) -> None:
         warnings.warn(
-            f"{self.__class__.__name__}._schedule is deprecated, "
-            f"please use {self.__class__.__name__}.crawl or {self.__class__.__name__}.download instead",
+            "ExecutionEngine.schedule is deprecated, please use "
+            "ExecutionEngine.crawl or ExecutionEngine.download instead",
             category=ScrapyDeprecationWarning,
             stacklevel=2,
         )
         if self.slot is None:
-            raise RuntimeError(f"{self.__class__.__name__}.slot is not assigned")
+            raise RuntimeError("Engine slot not assigned")
         self._schedule_request(request)
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -224,7 +224,7 @@ class ExecutionEngine:
                 stacklevel=2,
             )
             if spider is not self.spider:
-                raise RuntimeError(f"The spider '{spider.name!r}' does not match the open spider")
+                raise RuntimeError(f"The spider {spider.name!r} does not match the open spider")
         if self.spider is None:
             raise RuntimeError(f"No open spider to crawl: {request}")
         self._schedule_request(request, self.spider)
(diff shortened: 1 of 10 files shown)


scrapy/core/engine.py
```diff
@@ -6,46 +6,55 @@
...
 
 
 class Slot:
 
-    def __init__(self, start_requests, close_if_idle, nextcall, scheduler):
-        self.closing = False
-        self.inprogress = set()  # requests in progress
-        self.start_requests = iter(start_requests)
+    def __init__(
+        self, start_requests: Iterable,
```
Gallaecio, line 31:
```suggestion
        self,
        start_requests: Iterable,
```

scrapy/core/engine.py
```diff
@@ -6,46 +6,55 @@
...
+        close_if_idle: bool,
+        nextcall: CallLaterOnce,
+        scheduler,
+    ) -> None:
+        self.closing: Optional[Deferred] = None
+        self.inprogress: Set[Request] = set()  # requests in progress
```
Gallaecio, line 37: I think type hints make the commend unnecessary here.

```suggestion
        self.inprogress: Set[Request] = set()
```

scrapy/core/engine.py
```diff
@@ -102,188 +109,175 @@ def close(self):
...
+        if self.spider_is_idle(spider) and self.slot.close_if_idle:
             self._spider_idle(spider)
 
-    def _needs_backout(self, spider):
-        slot = self.slot
+    def _needs_backout(self, spider: Spider) -> bool:
         return (
             not self.running
-            or slot.closing
+            or self.slot is not None and self.slot.closing
```
Gallaecio, line 147: This gives the impression that `self.slot` may be `None`, which does not seem to be the case, given the implementation of `_next_request` and assuming that is the only caller of `_needs_backout`.

Would an `assert` before the return be enough for mypy?

Same applies to `_next_request_from_scheduler` below.

scrapy/core/engine.py
```diff
@@ -102,188 +109,175 @@ def close(self):
...
-
-        if self.slot.start_requests is not None:
-            # not all start requests are handled
-            return False
-
-        if self.slot.scheduler.has_pending_requests():
-            # scheduler has pending requests
-            return False
-
+        if self.slot is not None:
```
Gallaecio, line 193: I think it makes sense to check here, given `spider_is_idle` is a public method. But since all current code expects it not to be `None`, I wonder if we should instead raise an exception, e.g. a `RuntimeError` complaining about `spider_is_idle` being called before `open_spider`.

Same for `crawl` and `schedule` below.

scrapy/core/engine.py
```diff
@@ -102,188 +109,175 @@ def close(self):
...
-        scheduler = self.scheduler_cls.from_crawler(self.crawler)
+        scheduler = create_instance(self.scheduler_cls, settings=None, crawler=self.crawler)
         start_requests = yield self.scraper.spidermw.process_start_requests(start_requests, spider)
-        slot = Slot(start_requests, close_if_idle, nextcall, scheduler)
-        self.slot = slot
+        self.slot = Slot(start_requests, close_if_idle, nextcall, scheduler)
         self.spider = spider
-        yield scheduler.open(spider)
+        if hasattr(scheduler, "open"):
+            yield scheduler.open(spider)
```
Gallaecio, line 273: I assume this and `hasattr(self.slot.scheduler, "close")` in line 317 below are not changes about typing, but about making those methods optional. +1.

I would say we need to remember to cover that in the documentation, but this pull request is a fork of a pull request to document the scheduler, so :shrug: .

elacuesta: Oh right, I didn't mean to include this change in this particular PR, I trimmed some changes from the initial PR but forgot about this one.

elacuesta: Reverted, these changes are included in #3559

elacuesta:
Well, this PR has evolved, I couldn't help myself, sorry 😅 

Included deprecations:
*`ExecutionEngine.open_spiders`: I didn't check the history, but I suspect this probably comes from the time when multiple spiders could run in a single crawl. That hasn't been the case for a long time, so it's not necessary to have a plurally-named property with only one element.
*`ExecutionEngine.has_capacity`: same as the above, the engine only hadles one spider slot, which can be checked directly. No other Scrapy component uses this method.

scrapy/core/engine.py
```diff
@@ -90,56 +90,63 @@ def start(self) -> Deferred:
         yield self._closewait
 
     def stop(self) -> Deferred:
-        """Stop the execution engine gracefully"""
+        """
+        Gracefully stop the execution engine
+        """
```
Gallaecio, line 95: Your docstring style changes seem to go against https://www.python.org/dev/peps/pep-0257/

scrapy/core/engine.py
```diff
@@ -90,56 +90,63 @@ def start(self) -> Deferred:
...
-        """Close the execution engine gracefully.
-
+        """
+        Gracefully close the execution engine.
         If it has already been started, stop it. In all cases, close all spiders and the downloader.
         """
         if self.running:
             return self.stop()  # will also close spiders and downloader
         elif self.spider is not None:
             return self.close_spider(self.spider, reason="shutdown")  # will also close downloader
```
Gallaecio: :lipstick: 

```suggestion
        if self.spider is not None:
            return self.close_spider(self.spider, reason="shutdown")  # will also close downloader
```

scrapy/core/engine.py
```diff
@@ -148,92 +155,88 @@ def _needs_backout(self, spider: Spider) -> bool:
...
+            lambda f: logger.error(
+                'Error while enqueuing downloader output',
+                exc_info=failure_to_exc_info(f),
+                extra={'spider': self.spider},
+            )
+        )
         return d
 
     def spider_is_idle(self, spider: Spider) -> bool:
+        assert self.slot is not None  # typing
```
Gallaecio, line 204: Since this is a public method, we cannot control whether or not `self.slot` is or not `None` by the time this method is called, so I don’t think `assert` is the way here, since we should only use asserts if the code would work the same after removing them. I would raise an exception instead.

Same for `crawl` and `schedule` below.

scrapy/core/engine.py
elacuesta: Interesting, I wasn't aware of that particular point. I don't really like it, but I won't go stubborn on this. Thanks!

scrapy/core/engine.py
elacuesta: Makes sense, thanks

scrapy/core/engine.py
```diff
@@ -54,269 +64,276 @@ def _maybe_fire_closing(self):
...
+        self.running = False
+        dfd = self.close_spider(self.spider, reason="shutdown") if self.spider is not None else defer_succeed(None)
+        return dfd.addBoth(_finish_stopping_engine)
 
-        If it has already been started, stop it. In all cases, close all spiders
-        and the downloader.
+    def close(self) -> Deferred:
+        """
+        Gracefully close the execution engine.
+        If it has already been started, stop it. In all cases, close all spiders and the downloader.
```
wRAR, line 109: "All spiders" should probably be edited?

scrapy/core/engine.py
```diff
@@ -54,269 +64,276 @@ def _maybe_fire_closing(self):
...
-            return self.stop()
-        elif self.open_spiders:
-            # Will also close downloader
-            return self._close_all_spiders()
-        else:
-            return defer.succeed(self.downloader.close())
-
-    def pause(self):
-        """Pause the execution engine"""
+            return self.stop()  # will also close spiders and downloader
```
wRAR, line 112: Same

scrapy/core/engine.py
```diff
@@ -5,47 +5,57 @@
 
 """
 import logging
+import warnings
 from time import time
+from typing import Callable, Iterable, Iterator, Optional, Set, Union
 
-from twisted.internet import defer, task
+from twisted.internet.defer import Deferred, inlineCallbacks, succeed as defer_succeed
```
wRAR, line 12: Note that we have defer_succeed in utils.defer

elacuesta: `utils.defer.defer_succeed` introduces a bit of delay, this renaming is because the previous usage was `defer.succed`, and I thought just `succeed` might not be as intuitive. I'll switch back to `succeed` to avoid confusion, thanks.

wRAR: Sure, I just meant that there can be confusion and also a theoretical naming clash in the future.

scrapy/core/engine.py
```diff
@@ -54,269 +64,277 @@ def _maybe_fire_closing(self):
...
-    def open_spiders(self):
-        return [self.spider] if self.spider else []
-
-    def has_capacity(self):
-        """Does the engine have capacity to handle more spiders"""
-        return not bool(self.slot)
-
-    def crawl(self, request, spider):
-        if spider not in self.open_spiders:
-            raise RuntimeError(f"Spider {spider.name!r} not opened when crawling: {request}")
```
Gallaecio, line 217: Shouldn’t this error still be caught, if the spider argument does not match `self.spider`? (maybe with a better error message that explicitly says so, instead of the old message)

scrapy/core/engine.py
```diff
@@ -349,12 +366,26 @@ def errback(failure):
...
+    def schedule(self, request: Request, spider: Spider) -> None:
+        warnings.warn(
+            "ExecutionEngine.schedule is deprecated, please use "
+            "ExecutionEngine.crawl or ExecutionEngine.download instead",
+            category=ScrapyDeprecationWarning,
+            stacklevel=2,
+        )
+        if self.slot is None:
+            raise RuntimeError("Engine slot not assigned")
+        self._schedule_request(request)
```
Gallaecio, line 391: A potential issue I see here and in public methods where the `spider` argument is deprecated is that there is a slight behavior change that, while unlikely to affect many, if any, still happens without a deprecation period: we are just assuming that the `spider` argument, if passed, is the same as `self.spider`. And while that is what _should_ be happening, code-wise we cannot be sure of that.

So I’m thinking that, to deprecate things properly, we may need to add a check specifically for that, and have things work as before in those cases (i.e. pass the `spider` argument forward to subsequent calls, instead of letting `self.spider` be used) but log a deprecation warning about it.

scrapy/core/engine.py
```diff
@@ -54,269 +64,277 @@ def _maybe_fire_closing(self):
...
-        if self.spider_is_idle(spider):
-            self.close_spider(spider, reason='finished')
+        if self.spider_is_idle():
+            self.close_spider(self.spider, reason='finished')
 
-    def close_spider(self, spider, reason='cancelled'):
+    def close_spider(self, spider: Spider, reason: str = "cancelled") -> Deferred:
         """Close (cancel) spider and clear all its outstanding requests"""
+        if self.slot is None:
+            raise RuntimeError("Engine slot not assigned")
```
Gallaecio, line 326: This change makes a lot of sense, much better than the exception that `if slot.closing` would raise otherwise. But we should probably mention it in the release notes, even if briefly.

I wonder if we should start adding a “release notes notes” section to the description of some pull requests, to make sure we do not miss these small things.

scrapy/core/engine.py
elacuesta: Makes sense, I added the arguments back to the underscored methods that are called from public ones.
Note that this doesn't include `_spider_idle`, `_next_request`, `_next_request_from_scheduler`. `_spider_idle` and `_next_request_from_scheduler` are only called from `_next_request`, which is in turn only "called" (set to be called later, that is) from `open_spider`, the method which sets `self.spider`.
This should make things safer. Nobody should be calling `crawl/download` with other spiders though, I'm not even sure what would happen but it's probably not pretty.

scrapy/core/engine.py
elacuesta: Good call. It's updated, hopefully this is more explicit:
```python
def crawl(self, request: Request, spider: Optional[Spider] = None) -> None:
    """Inject the request into the spider <-> downloader pipeline"""
    if spider is not None:
        warnings.warn(
            "Passing a 'spider' argument to ExecutionEngine.crawl is deprecated",
            stacklevel=2,
        )
        if spider is not self.spider:
            raise RuntimeError(f"The spider '{spider.name!r}' does not match the open spider")
    if self.spider is None:
    self._schedule_request(request, self.spider)
    self.slot.nextcall.schedule()  # type: ignore[union-attr]
```

Note that I added the same verification code at `download`, which at the moment does not check if the passed spider matches the open one. This is technically backward-incompatible, but I think it's a reasonable thing to do.

scrapy/core/engine.py
elacuesta: Actually, the only risk is with `download`: `spider_is_idle` was not using the spider to begin with and `crawl` was checking if the spider was in `self.open_spiders` (now it checks if it matches `self.spider`). As I mentioned in https://github.com/scrapy/scrapy/pull/5090#discussion_r614277741, I added the same check to `download`.

scrapy/core/engine.py
Gallaecio: > Note that I added the same verification code at download, which at the moment does not check if the passed spider matches the open one. This is technically backward-incompatible, but I think it's a reasonable thing to do.

Why not just log an additional warning instead, or even an error? It should still catch the attention of anyone actually doing this (hopefully no one), while still remaining backward compatible.

Gallaecio (approved):

PR #4998, opened 2021-02-21
S3FeedStorage supports custom endpoint of object storage.

bkayranci:
Scrapy documentation says that, scrapy supports other object storages via `AWS_ENDPOINT_URL`. Unfortunately, custom object storage endpoint does not support in feed exporter.

## Related Links
- Fixes #4997
- Fixes #4871 
- https://docs.scrapy.org/en/latest/topics/settings.html?highlight=AWS_ENDPOINT_URL#aws-endpoint-url

PR #5092, opened 2021-04-08
Use AWS_REGION_NAME and AWS_ENDPOINT_URL, if available, for S3 feed export

gfrmin:
AWS_REGION_NAME and AWS_ENDPOINT_URL are already used in scrapy, but only for image pipeline. This will make them available for S3 feed export.

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -118,7 +118,7 @@ class FileFeedStorage:
 
 class S3FeedStorage(BlockingFeedStorage):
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
The endpoint part is implemented in a pending pull request, https://github.com/scrapy/scrapy/pull/4998, which you could take as a reference of a clean implementation of this (e.g. placing the `__init__` parameters at the end for backward compatibility).

wRAR:
There is also an older (and wrongly named) #4429 which seems to have the same goal.

Gallaecio:
https://github.com/scrapy/scrapy/pull/5980

PR #5094, opened 2021-04-09
Engine tests: fix item class spider, add minimal type hints

elacuesta:
Currently the `AttrsItemsSpider` is using an item class that inherits from `scrapy.item.Item`

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -97,12 +97,10 @@ except ImportError:
 else:
     TestDataClass = make_dataclass("TestDataClass", [("name", str), ("url", str), ("price", int)])
 
-    class _dataclass_spider(DictItemsSpider):
+    class DataClassItemsSpider(DictItemsSpider):  # type: ignore[no-redef]
         def parse_item(self, response):
             item = super().parse_item(response)
             return TestDataClass(**item)
-
-    DataClassItemsSpider = _dataclass_spider
 
 
 class ItemZeroDivisionErrorSpider(TestSpider):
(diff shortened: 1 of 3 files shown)


tests/test_engine.py
```diff
@@ -97,14 +97,12 @@ class AttrsItemsSpider(TestSpider):
...
         def parse_item(self, response):
             item = super().parse_item(response)
-            return TestDataClass(
-                name=item.get('name'),
-                url=item.get('url'),
-                price=item.get('price'),
-            )
+            return TestDataClass(**item)
+
+    DataClassItemsSpider = _dataclass_spider
```
elacuesta, line 105: This is to prevent mypy from complaining about DataClassItemsSpider alredy being defined

Gallaecio: I wonder if there’s a more readable way to handle this.

After reading https://github.com/python/mypy/issues/1191#issuecomment-780471195 , I wonder if it would not be better to annotate the first definition of `DataClassItemsSpider` (`DataClassItemsSpider = None`) as an optional subclass of `DictItemsSpider`, assuming that is enough to silence mypy about this.

Otherwise, I think it may be better to silence mypy with a comment (assuming that’s possible) rather than defining a class with one name and later renaming it. The last thing I would think if I find this code in the future is that this has been done for mypy, I would instead wonder how the code would work differently if the class was defined with its final name directly.

elacuesta: Right, without context it's a little hard to know why this is done this way. Since this is a test, we can probably do well with just `# type: ignore` 👍

Gallaecio (approved):

PR #5095, opened 2021-04-09
Cleanup of slot handling in Scraper.

wRAR:
This is minor cleanup of, I suppose, remains of multiple slots per Scraper, similar to parts of #5090, but I think there is nothing else to clean up/deprecate here.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -82,28 +82,26 @@ class Scraper:
 
     def close_spider(self, spider):
         """Close a spider being scraped and release its resources"""
-        slot = self.slot
-        slot.closing = defer.Deferred()
-        slot.closing.addCallback(self.itemproc.close_spider)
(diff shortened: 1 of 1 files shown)


elacuesta:
Nice 🚀 
What do you think about storing the spider received in `open_spider` as an attribute to avoid passing it to underscored methods like `_check_if_closing`, `_scrape`, `_scrape_next`, etc?

wRAR:
Good question, I guess I need to check the code that calls open/close_spider.

Gallaecio (approved):

Gallaecio:
I’m not merging already just in case you want to address https://github.com/scrapy/scrapy/pull/5095#issuecomment-816938093 as part of this pull request.

Gallaecio:
> Do we know if anything outside Scrapy could call these methods?

I would assume so, however unlikely.

PR #5097, opened 2021-04-13
Face OSError on invalid images #5079

### LINKED ISSUE #5079: ImagePipeline breaks on invalid images
@apalala reported this issue when a site sends some invalid images, and suggested a fix for the spider middleware:

```
        image_stream = self.get_images(response, request, info, item=item)
        while True:
            try:
                path, image, buf = next(image_stream)
            except StopIteration:
                break
            except Exception:
                continue
```

I’ve not checked if there’s a cleaner fix possible (i.e. closer to the call to Pillow code), but it should be trivial to fix either way. I suspect writing a test may be the hardest part here.

marlenachatzigrigoriou:
The valid images are downloaded while the invalid ones not, and the OSError is faced and not raised anymore.
I am in the process of creating a test. 
Hopefully fixes #5079.


Gallaecio:
The `security` job is failing about https://bandit.readthedocs.io/en/latest/plugins/b112_try_except_continue.html. I agree with the report that we should at least log the exception, even if the code works otherwise. That would mean adding a `logger` object to the module as other modules in Scrapy have, and using something like `logger.exception('<message>')`, which will include the original exception traceback.

Logging the error also helps with testing, as you can check for the exception in the logs to make sure it got logged. We have a few tests already that capture logs and check their messages.

scrapy/pipelines/images.py
```diff
@@ -114,25 +114,25 @@ def file_downloaded(self, response, request, info, *, item=None):
         return self.image_downloaded(response, request, info, item=item)
 
     def image_downloaded(self, response, request, info, *, item=None):
-        
+        logger = logging.getLogger()
```
Gallaecio, line 117: It’s common to define this at the module level, after imports, and passing `__name__` to `getLogger` so that the logger is named after the module.

scrapy/pipelines/images.py
```diff
@@ -114,25 +114,25 @@ def file_downloaded(self, response, request, info, *, item=None):
...
         checksum = None
         image_stream = self.get_images(response, request, info, item=item)
         while True:
             try:
                 path, image, buf = next(image_stream)
-            except OSError as e:
+            except OSError:
                 logger.exception('Could not process image')
```
Gallaecio, line 124: There should be a `continue` in this case, otherwise I imagine we would end up with duplicate processing of the image from the previous iteration.

scrapy/pipelines/images.py
```diff
@@ -17,6 +17,9 @@
 from scrapy.settings import Settings
 from scrapy.utils.misc import md5sum
 from scrapy.utils.python import to_bytes
+import logging
```
Gallaecio, line 20: Please, move this import after `import hashlib` above.

Gallaecio (commented):
The code looks good to me now. Now we need to have a test that verifies the fix.

I’ve not had a look at the existing tests, but since there are existing tests, maybe it’s possible to copy one of them and modified so that it gets an image that causes an issue (I imagine arbitrary bytes would do the job, otherwise we can probably add one of the images of the original report to the tests folder and use that).

Let me know if you have any trouble preparing a test for this, in Scrapy it’s sadly often harder to write a test than to fix the code itself.

scrapy/pipelines/images.py
marlenachatzigrigoriou: Sure!

scrapy/pipelines/images.py
```diff
@@ -115,7 +118,16 @@ def file_downloaded(self, response, request, info, *, item=None):
 
     def image_downloaded(self, response, request, info, *, item=None):
         checksum = None
-        for path, image, buf in self.get_images(response, request, info, item=item):
+        image_stream = self.get_images(response, request, info, item=item)
+        while True:
+            try:
+                path, image, buf = next(image_stream)
+            except OSError:
```
apalala, line 125: In #5079 there's an:

```python
except StopIteration:
    break
```

To break out of the `while True:` cleanly.

scrapy/pipelines/images.py
```diff
@@ -115,7 +118,16 @@ def file_downloaded(self, response, request, info, *, item=None):
...
+                logger.exception('Could not process image')
+                continue
```
apalala, line 127: This is implicit (unneeded), as the `except` will fall through.

scrapy/pipelines/images.py
```diff
@@ -115,7 +118,16 @@ def file_downloaded(self, response, request, info, *, item=None):
...
+            except Exception:
+                logger.exception('Stopped processing images')
+                break
```
apalala, line 130: The loop should not `break` because one image failed. It should continue processing the other images.

apalala (changes requested):

scrapy/pipelines/images.py
Gallaecio: It would still execute the rest of the code within the loop, though, with references to the image data from the previous iteration. I think this is needed.

scrapy/pipelines/images.py
Gallaecio: Do we expect exceptions other than `OSError` happening? If we add `except StopIteration`, maybe we don’t need this `except` here at all.

scrapy/pipelines/images.py
marlenachatzigrigoriou: Should I correct it or not?

scrapy/pipelines/images.py
marlenachatzigrigoriou: So, I should replace `break `with `continue ` or delete it at all, as long as I add `except StopIteration` ?

Gallaecio: I guess we could leave `Exception` for subclasses raising exceptions we have no control over. So `continue` here and `except StopIteration` before this `except`, I think.

marlenachatzigrigoriou:
@Gallaecio I am having trouble with the tests. I have made some but I am not sure. Should I upload it, since someone else pull-requested for this issue, too...

Gallaecio:
The tests from #5115 are not perfect. If yours go in a different line that you feel may be better, it would be great if your could upload yours as well.

marlenachatzigrigoriou:
I think the lines 126-127 including the logger (in method `image_downloaded` in [scrapy/pipelines/images.py](https://github.com/scrapy/scrapy/pull/5097/files/64187d2726680b2571e0433ed72f71617a9645a3#diff-80b0c34065951c3988ba49f86ba87467082abb1624ed9574519815ebf8f6527e)) have to be tested, too, as codecov has mentioned.
Also, as I saw in #1235, `logger.exception` is not a good practice and is not used at all in Scrapy, so it was difficult for me to find a way to test it. Maybe, it should change into `logger.error`.

Gallaecio:
Trying to figure out the right way to handle the loop `try-except`, I’ve had a deeper look at the code, and I’m not sure this is the right place to fix the issue. Despite the name, `get_images` does not really iterate the different images extracted from an item, as I thought. Instead, it seems to be executed once per image, and it yields the image and, if the `IMAGES_THUMBS` setting is used, it may yield additional images that are thumbnails of the image from the first yield. So I don’t think this will fix #5079.

I’ll look deeper into this and discuss this in #5079, as it affects any solution.

Gallaecio:
https://github.com/scrapy/scrapy/issues/5079#issuecomment-879230300

PR #5100, opened 2021-04-14
Additional typing for scraper and a small code change.

wRAR:
This was left out of the previous PR for several reasons.

This removes a check from `_log_download_errors`, I believe this check can't be False.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -3,7 +3,7 @@ extracts information from them"""
 
 import logging
 from collections import deque
-from typing import Union, Optional, Tuple, Set, Deque, Any, Iterable
+from typing import Any, Deque, Iterable, Optional, Set, Tuple, Union
 
 from itemadapter import is_item
 from twisted.internet.defer import Deferred, inlineCallbacks
@@ -19,7 +19,7 @@ from scrapy.utils.misc import load_object, warn_on_generator_with_return_value
 from scrapy.utils.spider import iterate_spider_output
 
 
-QUEUE_TUPLE = Tuple[Union[Response, Failure], Request, Deferred]
+QueueTuple = Tuple[Union[Response, Failure], Request, Deferred]
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


scrapy/core/scraper.py
```diff
@@ -3,11 +3,10 @@
 
 import logging
 from collections import deque
-from collections.abc import Iterable
-from typing import Union
+from typing import Union, Optional, Tuple, Set, Deque, Any, Iterable
```
Gallaecio, line 6:
```suggestion
from typing import Any, Deque, Iterable, Optional, Set, Tuple, Union
```

scrapy/core/scraper.py
```diff
@@ -20,6 +19,9 @@
 from scrapy.utils.spider import iterate_spider_output
 
 
+QUEUE_TUPLE = Tuple[Union[Response, Failure], Request, Deferred]
```
Gallaecio, line 22: I would use the same letter case as other classes and types, in line with https://docs.python.org/3/library/typing.html

```suggestion
QueueTuple = Tuple[Union[Response, Failure], Request, Deferred]
```

scrapy/core/scraper.py
```diff
@@ -230,10 +246,12 @@ def _log_download_errors(self, spider_failure, download_failure, request, spider
 
         if spider_failure is not download_failure:
             return spider_failure
+        return None
 
-    def _itemproc_finished(self, output, item, response, spider):
+    def _itemproc_finished(self, output: Any, item: Any, response: Response, spider: Spider) -> None:
```
Gallaecio, line 251: I don’t think it needs to be done in this pull request, but I’m thinking that it may be good to have a type alias for all `itemadapter`-supported item types, and use it in scenarios like this one instead of `Any`. It probably makes sense to implement that type alias in `itemadapter`, though.

Gallaecio (approved):

scrapy/core/scraper.py
wRAR: Yeah, assuming it's possible to express.

PR #5115, opened 2021-04-24
Fixed invalid image downloading error and added test case #5079

### LINKED ISSUE #5079: ImagePipeline breaks on invalid images

spittieUM:
PR aims to address #5079 

Fixes #5079

### DIFF
--- a/tests/test_pipeline_images.py
+++ b/tests/test_pipeline_images.py
@@ -4,6 +4,7 @@ import random
 from shutil import rmtree
 from tempfile import mkdtemp
 from unittest import skipIf
+import glob
 
 import attr
 from itemadapter import ItemAdapter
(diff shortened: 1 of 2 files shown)


spittieUM:
Linking to #5079

tests/test_pipeline_images.py
```diff
@@ -93,6 +94,14 @@ def test_thumbnail_name(self):
                                     info=object()),
                          'thumbs/50/850233df65a5b83361798f532f1fc549cd13cbe9.jpg')
 
+    def test_pipeline_invalid_images(self):
+        urls = ['https://img01.mgo-images.com/image/thumbnail/v2/content/MMV5CEBE0C5DC82A5028E7EA6B91F904DEF3.jpeg',
+                'https://img01.mgo-images.com/image/thumbnail/v2/content/MMVA6B4523491C0B5A88B392024490A44B7C.jpeg',
+                'https://scrapy.org/img/scrapylogo.png']
+        for url in urls:
+            Request(url)
+        self.assertEqual(len(glob.glob(self.pipeline.file_path + "/*")), 1)
```
Gallaecio, line 103: While this is the idea, tests should not rely on an internet connection. We should download those files into the `tests` folder somewhere, and have the tests work from them locally. I wonder if using `file://` for that will do the job, although that needs an absolute path so it will not be trivial; there may be examples of `file://` in other tests.

Gallaecio:
https://github.com/scrapy/scrapy/issues/5079#issuecomment-879230300

PR #5118, opened 2021-04-26
Fix utf-8 encoding #5096

### LINKED ISSUE #5096: Wrong charset handling in MailSender
Looks like passing `charset='utf-8'` makes a plain text message with `Content-Transfer-Encoding: base64` which then can't be read.

At https://github.com/scrapy/scrapy/blob/5a75b14a5fbbbd37c14aa7317761655ac7706b70/scrapy/mail.py#L81 `set_charset` is called but as the payload is not set yet, the underlying class just sets the headers. When later `set_payload` is called, it doesn't do any encoding, but the `Content-Transfer-Encoding` is already set. Looks like the fix should be passing the encoding to `set_payload` too, like was proposed in #3722 (and `set_charset` may be safe to be removed, not sure about this). Note that we have tests but they don't catch this.

Note also, that all of this seems to be compat code according to the Python docs.

mmitropoulou:
Hello, this possibly fixes #5096

Fixes #5096, closes #5506


wRAR:
Can you please also add tests and/or modify the existing ones so that this change is tested properly?

mmitropoulou:
Hi @wRAR. I changed the positioning of set_charset method for mails with attachments (it is now similar to how it as was at the begining) and removed it for mails without attachments (it is called anyway from the get_payload method if you pass a value for charset) because i feel like this is a better approach. I didn't create new tests or modify the tests a lot, because i feel that is not needed (everything eg. charset, body etc is already tested and I didnt add anything new that needs testing). I would like to hear your thoughts on it and get feedback. Thank you!

Gallaecio:
> I didn't create new tests or modify the tests a lot, because i feel that is not needed (everything eg. charset, body etc is already tested and I didnt add anything new that needs testing).

If you fixed https://github.com/scrapy/scrapy/issues/5096, then that’s something new that needs to be tested. You should add/modify tests as needed so that, if you reverted your fix, tests would now break.

On a separate note, I wonder if `msg.set_charset(charset)` is needed at all after your change. When there are attachments, the message body is already added with a charset parameter specified. Does removing `msg.set_charset(charset)` make any difference provided you keep passing the charset when adding the payload text without attachments?

wRAR (approved):
Thanks! I've checked the change with my original use case and it fixes it. I've also checked that the changed test doesn't pass without the code changes, not sure if we need to add an explicit check that the body is base64-encoded, though we don't do that in the other test so that's fine to leave as is.

wRAR:
> If you fixed #5096, then that’s something new that needs to be tested. You should add/modify tests as needed so that, if you reverted your fix, tests would now break.

The test change covers it, because of the added `decode=True`.

> I wonder if msg.set_charset(charset) is needed at all after your change. When there are attachments, the message body is already added with a charset parameter specified. Does removing msg.set_charset(charset) make any difference provided you keep passing the charset when adding the payload text without attachments?

Removing breaks the message, as it now lacks `Content-Type` and `Content-Transfer-Encoding` top-level headers.

PR #5122, opened 2021-04-28
Require setuptools, scrapy/cmdline.py, /setup.py and tests/test_webclient.py import pkg_resources

hrnciar:


Gallaecio (approved):
Makes sense to me. I believe we missed this because it’s very often installed by default.

wRAR (approved):

PR #5123, opened 2021-04-28
Add the explanation how to define URI parameters by yourself 

shinichi-takayanagi:
Since the way to define a new `URI variable` was not present in the document, so I have added.


docs/topics/feed-exports.rst
```diff
@@ -135,6 +135,35 @@ Here are some examples to illustrate:
...
+    process = CrawlerProcess(settings={
+        "FEEDS": {
+            "%(filename)s.json": {"format": "json"},
+        },
+    })
+
+    process.crawl(MySpider, filename="crawled_quotes")
+    process.start()
+    ...
+
```
Gallaecio, line 166: The prior documentation already states that spider attributes are replaces. What this new documentation does is expand on the ways that you can define such spider attributes. So I think this new documentation does not belong here, but in https://docs.scrapy.org/en/latest/topics/spiders.html#spider-arguments. I think we can consider the first half covered, but maybe mentioning `CrawlerProcess` makes sense, although I would not go as far as providing a complete example for that, I would indicate the `CrawlerProcess` object creation and the `crawl` call at most.

It would be great, however, to include a reference to https://docs.scrapy.org/en/latest/topics/spiders.html#spider-arguments from the existing documentation here. Maybe we could add a ``.. note::`` after the “Any other named parameter gets replaced by …” paragraph like:

```rst
.. note:: :ref:`Spider arguments <spiderargs>` become spider attributes, hence they can also be used as storage URI parameters.
```

shinichi-takayanagi: @Gallaecio 
Thanks for your comment! I've tried to reflect on your suggestion.
I would appreciate it if you could check out and merge this PR when you have a moment🚀

docs/topics/feed-exports.rst
```diff
@@ -135,34 +135,7 @@ Here are some examples to illustrate:
...
-    process = CrawlerProcess(settings={
-        "FEEDS": {
-            "%(filename)s.json": {"format": "json"},
-        },
-    })
-
-    process.crawl(MySpider, filename="crawled_quotes")
-    process.start()
-    ...
+.. note:: :ref:`Spider arguments <spiderargs>` become spider attributes, hence they can also be used as storage URI parameters.
```
Gallaecio, line 138:
```suggestion
.. note:: :ref:`Spider arguments <spiderargs>` become spider attributes, hence 
          they can also be used as storage URI parameters.
```

docs/topics/spiders.rst
```diff
@@ -294,6 +294,22 @@ The above example can also be written as follows::
...
+    ...
+    
+    process = CrawlerProcess(settings={
+        "FEEDS": {
+            "%(category)s.json": {"format": "json"},
+        },
+    })
+    process.crawl(MySpider, category="electronics")
+
+See :ref:`run-from-script` for more information about :class:`scrapy.crawler.CrawlerProcess`.
```
Gallaecio, line 311: I would avoid the `FEEDS` example here, as that’s beyond the scope of this section. And I think we can make this part slightly shorter while providing the same information:

```suggestion
If you are :ref:`running Scrapy from a script <run-from-script>`, you can 
specify spider arguments when calling 
:class:`CrawlerProcess.crawl <scrapy.crawler.CrawlerProcess.crawl>` or
:class:`CrawlerRunner.crawl <scrapy.crawler.CrawlerRunner.crawl>`::

    process = CrawlerProcess()
    process.crawl(MySpider, category="electronics")
```

shinichi-takayanagi: LGTM!

Gallaecio (approved):

Gallaecio:
@shinichi-takayanagi Please, change the target branch from `2.5` to `master`.

shinichi-takayanagi:
@Gallaecio done👍

Gallaecio (approved):

Gallaecio:
Thanks!

kmike:
Nice, thanks @shinichi-takayanagi!

PR #5124, opened 2021-04-29
[WIP] RequestList

### LINKED ISSUE #2600: Add a RequestSet class for grouping requests
This came up in #2582.

I think scrapy lacks a means of grouping requests and sharing some state between them that would become irrelevant once the last request in the group has been dealt with and should be cleaned up. Tracking groups manually with request callbacks & errorbacks is error-prone and hard to do right. A RequestSet class that allows the user to do something like

```
def parse(self, response):
    ....
    yield RequestSet(Request(...) for ...)
```

and forget about shared state management would be nice.

I think of RequestSet as something that:

- has a `DeferredList`-like API (`asyncio.gather` has the drawback of being a function rather than an object)
- ~~has its own `callback` to be run when the request set is dealt with, probably an `errback` too, for symmetry~~ has a Deferred that would fire when the RequestSet is being cleaned up
- has its own `'meta'` dictionary, much like the one shared between Request & Response objects
- knows that it contains Requests as deferreds
- since it knows that it contains Requests, it can piggyback on the first received value, which should be a response, and do `response.meta['request_set'] = self` so that the callbacks can access the shared data
- (maybe) it should silently copy fields from `request_set.meta` to `response.meta` if they are unset in `request.meta`, or maybe even make `request.meta` a ChainDict with fallback to `request_set.meta` so that middlewares don't have to trace through nested metadata dictionaries to check the fields they are interested in
- it should wrap requests coming from its respective response callbacks unless specifically asked not to do that, e.g. with `request.meta['request_set'] = None`
- (maybe) it should be possible to return other RequestSet from response callbacks 
- (maybe) returned RequestSets should be made nestable, i.e. to keep the parent RequestSet alive during their lifetime if not explicitly asked not to with `request_set.meta['request_set'] = None` (if nesting is considered, the `request_set` metadata key seems redundant and we might consider `parent_set` instead.
- not sure if it's worth it to make them nestable, i.e. if a certain response callback produces a different RequestSet, should it be owned by the parent request set?

One more thing to consider is cross-referencing RequestSets, i.e. when two requests that should belong to one RequestSet are produced by different callbacks and thus have different scopes. Maybe a simple `WeakValueDictionary` would suffice to lookup the sets and ensure the references are cleaned up as necessary. But then you'd have the usual get-or-create operation, that might be worth creating an etalon implementation for.

elacuesta:
Definitely not finished, proceed with caution 🚧 

An attempt to fix #2600, i.e. provide a way to wait on several requests before invoking a callback. This is done by wrapping the requests in a `scrapy.http.request.RequestList` object, which behaves similarly to a `DeferredList` (the feature is in fact implemented internally with a `DeferredList`). The callback will get a `scrapy.http.response.ResponseList` object, which is a wrapper around the received responses (initially I was passing the responses as `*args`, but wrapping them in a specific class helps to have better validation in the engine and the scraper classes). Currently `RequestList` objects go through the spider middlewares unchanged, I wanted to avoid disassembling them into their individual requests. 

Several things are missing and decisions need to be made, including (but possibly not limited to) the following:
- [ ] Error handling (what to do if one or more request(s) fail(s)?)
- [x] Redirects (or any new request coming into the engine from the downloader middleware)
- [ ] `RequestList.__repr__`
- [ ] Review weak references
- [x] RequestList serialization to/from dict (also update scheduler queues)
- [ ] Review method names/paramaters (most names refer to single requests)
- [ ] Review type hints
- [ ] Review scheduler flow
- [x] Spider middlewares: process_spider_input
- [x] Spider middlewares: process_spider_output
- [ ] Stats
- [ ] Tests
- [ ] Docs

In spite of that, I'm opening now to get early feedback and suggestions, all comments are welcome.

Sample spider:
```python
from scrapy import Spider, Request
from scrapy.http.request import RequestList
from scrapy.http.response import ResponseList


class RequestListSpider(Spider):
    name = "list"

    def start_requests(self):
        yield RequestList(
            Request("http://httpbin.org/robots.txt"),
            Request("http://httpbin.org/delay/2"),
            Request("http://httpbin.org/user-agent"),
            callback=self.parse_list,
        )
        yield Request("https://example.org")

    def parse(self, response):
        yield {"url": response.url, "status": response.status}

    def parse_list(self, response_list: ResponseList):
        for response in response_list:
            yield {"url": response.url, "status": response.status}
```

```
$ scrapy runspider requestlist.py
(...)
2021-04-28 20:10:43 [scrapy.core.engine] INFO: Spider opened
2021-04-28 20:10:43 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2021-04-28 20:10:43 [scrapy.extensions.telnet] INFO: Telnet console listening on 127.0.0.1:6023
2021-04-28 20:10:43 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://httpbin.org/robots.txt> (referer: None)
2021-04-28 20:10:43 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://httpbin.org/user-agent> (referer: None)
2021-04-28 20:10:44 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://example.org> (referer: None)
2021-04-28 20:10:44 [scrapy.core.scraper] DEBUG: Scraped from <200 https://example.org>
{'url': 'https://example.org', 'status': 200}
2021-04-28 20:10:45 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://httpbin.org/delay/2> (referer: None)
2021-04-28 20:10:45 [scrapy.core.scraper] DEBUG: Scraped from <scrapy.http.response.ResponseList object at 0x7f9cefe5dac0>
{'url': 'http://httpbin.org/robots.txt', 'status': 200}
2021-04-28 20:10:45 [scrapy.core.scraper] DEBUG: Scraped from <scrapy.http.response.ResponseList object at 0x7f9cefe5dac0>
{'url': 'http://httpbin.org/delay/2', 'status': 200}
2021-04-28 20:10:45 [scrapy.core.scraper] DEBUG: Scraped from <scrapy.http.response.ResponseList object at 0x7f9cefe5dac0>
{'url': 'http://httpbin.org/user-agent', 'status': 200}
2021-04-28 20:10:45 [scrapy.core.engine] INFO: Closing spider (finished)
(...)
```


Gallaecio:
My main concern is the amount of backward incompatible changes, if we consider
different parameter types backward incompatible (which I think we should, and
now I’m really glad we’ve started adding type hints that makes this type of
change so obvious).

> I wanted to avoid disassembling them into their requests, but that is not set
> in stone.

I would sesiously consider it, if that allows to make these changes backward
compatible. Did you have specific concerns regarding disassembling vs keeping
requests together?

I wonder if we can use a spider middleware that runs before any other to
disassemble RequestList, and have that spider middleware use Request.meta to
keep track of requests that belong to the same RequestList somehow.

> Error handling (what to do if a request fails?)

To keep things simple, I think we could call the errback of a RequestList if
any of the requests fails.

I’m thinking that in this case some people may wish for the errback to be
called as soon as 1 of the requests fails, but I think we should still wait for
all requests to finish. Otherwise, things could get messy if there’s pending
requests waiting in the scheduler or the downloader.

AdrianAtZyte:
With download_async() and asyncio.gather(), I wonder if we should consider #2600 done and close this PR.

PR #5127, opened 2021-05-01
Added a time limit to LocalCache

### LINKED ISSUE #905: Improve DNS caching

### LINKED ISSUE #5125: DNS Resolver should have an expire time for cache.
## Summary

DNS Cache should have expire time.

## Motivation

I am using K8S to deploy scrapy to post some data to a API concurrently. However, when I update the api, the pod may change to another host machine, therefore, the ip address behind the API's Domain is changed. But now Scrapy's resolver only uses OrderedDict to store the DNS cache with a size but without expire time. So after I update the api, Scrapy will get connection refused because it still uses the old IP.

Now, I write a downloader middleware to solve this problem. when I get connection, I will delete the cache of this domain in resolver's `dnscache` dict, and the retry. But I hope you could write a more pythonic way.

FilipePintoReis:
Closes #905 

Changed LocalCache class so it takes into account how much time has passed.

### DIFF
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -65,9 +65,9 @@ class CaselessDict(dict):
 
 
 class LocalCache(collections.OrderedDict):
-    """Dictionary with a finite number of keys.
-
-    Older items expires first.
+    """
+    Dictionary with a finite number of keys.
+    Older items expire first.
     """
 
     def __init__(self, limit=None, time_limit=None):
@@ -79,20 +79,22 @@ class LocalCache(collections.OrderedDict):
         now = datetime.now()
         value = (value, now)
 
-        if not self.limit is None:
+        if self.limit is not None:
             while len(self) >= self.limit:
                 self.popitem(last=False)
-            for _, val in self.items():
-                if not time_limit is None and int((now - val[1]).total_seconds()) > self.time_limit:
+            for val in self.values():
+                if self.time_limit is not None and (now - val[1]).total_seconds() > self.time_limit:
                     self.popitem(last=False)
                 else:
                     break
         super().__setitem__(key, value)
 
     def __getitem__(self, key):
+        if self.time_limit is None:
+            return super().__getitem__(key)[0]
         for _, val in self.items():
             now = datetime.now()
-            if not time_limit is None and int((now - val[1]).total_seconds()) > self.time_limit:
+            if (now - val[1]).total_seconds() > self.time_limit:
                 self.popitem(last=False)
             else:
                 break
--- a/scrapy/utils/datatypes.py
+++ b/scrapy/utils/datatypes.py
@@ -82,23 +82,38 @@ class LocalCache(collections.OrderedDict):
         if self.limit is not None:
             while len(self) >= self.limit:
                 self.popitem(last=False)
-            for val in self.values():
-                if self.time_limit is not None and (now - val[1]).total_seconds() > self.time_limit:
-                    self.popitem(last=False)
-                else:
-                    break
+
+        while True:
+            try:
+                val = self.values().__iter__().__next__()
+            except StopIteration as e:
+                break
+            
+            if (datetime.now() - val[1]).total_seconds() > self.time_limit:
+                self.popitem(last=False)
+            else:
+                break
+            
         super().__setitem__(key, value)
 
     def __getitem__(self, key):
         if self.time_limit is None:
             return super().__getitem__(key)[0]
-        for _, val in self.items():
-            now = datetime.now()
-            if (now - val[1]).total_seconds() > self.time_limit:
+
+        while True:
+            try:
+                val = self.values().__iter__().__next__()
+            except StopIteration as e:
+                break
+            
+            if (datetime.now() - val[1]).total_seconds() > self.time_limit:
                 self.popitem(last=False)
             else:
                 break
-        return super().__getitem__(key)[0]
+        
+        val = (super().__getitem__(key)[0],datetime.now())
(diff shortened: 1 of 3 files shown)


scrapy/utils/datatypes.py
```diff
@@ -69,16 +70,35 @@ class LocalCache(collections.OrderedDict):
...
         super().__init__()
         self.limit = limit
+        self.time_limit = time_limit
 
     def __setitem__(self, key, value):
-        if self.limit:
+        now = datetime.now()
+        value = (value, now)
+
+        if not self.limit is None:
```
qiankunxienb, line 82: if self.limit is not None. Please refer to: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

scrapy/utils/datatypes.py
```diff
@@ -69,16 +70,35 @@ class LocalCache(collections.OrderedDict):
...
             while len(self) >= self.limit:
                 self.popitem(last=False)
+            for _, val in self.items():
+                if not time_limit is None and int((now - val[1]).total_seconds()) > self.time_limit:
```
qiankunxienb, line 86: Why do you convert `(now - val[1]).total_seconds()` to int? You can just compare float with int. there is not any necessary to convert float to int first.

scrapy/utils/datatypes.py
```diff
@@ -69,16 +70,35 @@ class LocalCache(collections.OrderedDict):
...
+                    self.popitem(last=False)
+                else:
+                    break
         super().__setitem__(key, value)
 
+    def __getitem__(self, key):
+        for _, val in self.items():
+            now = datetime.now()
+            if not time_limit is None and int((now - val[1]).total_seconds()) > self.time_limit:
```
qiankunxienb, line 95: for line 95, `if not time_limit is None`, here do you mean `self.time_limit` ? If yes, then you should change this line to `if self.time_limit is not None` and move this part upper before for loop to decrease compare times to increase.

```python
def __getitem__(self, key):
        for _, val in self.items():
            if self.time_limit is None:
                return super().__getitem__(key)[0]
            now = datetime.now()
            if (now - val[1]).total_seconds() > self.time_limit:
                self.popitem(last=False)
```

qiankunxienb:
I think this pr should not be merged. Here is my reason:

1. The solution is too simple. For example, you should refresh the expire time if a cache is hitted and it still works instead of just deleting all outdated caches.
2. This code pop item from a ordereddict in a for loop of this ordereddict itself. This behavior will cause problem.
3. The author doesn't follow PEP8.

scrapy/utils/datatypes.py
FilipePintoReis: I didn't know about this, thanks!
I changed all `not ... is None` to `... is not None`

scrapy/utils/datatypes.py
FilipePintoReis: You're correct, I deleted the cast.

scrapy/utils/datatypes.py
FilipePintoReis: I'm dazzled that this worked locally before, I must have not tested it properly before; Also, good performance catch.
I applied your suggestion.

tests/test_utils_datatypes.py
```diff
@@ -246,6 +247,19 @@ def test_cache_with_limit(self):
...
+        cache['b'] = 2
+        cache['c'] = 3
+        self.assertIn('a', cache)
+        self.assertIn('b', cache)
+        self.assertIn('c', cache)
+        sleep(3)
+        self.assertNotIn('a', cache)
+        self.assertNotIn('b', cache)
+        self.assertNotIn('c', cache)
+
```
Gallaecio, line 262: Rather than sleeping, we should use something like https://github.com/spulec/freezegun to mock the pass of time.

I also don’t think having 3 entries helps much. I can understand testing more than 1 entry in case there’s ever an issue which causes only 1 entry (e.g. first or last) to be removed, but testing 2 entries should be enough to catch that kind of issue.

Finally, I think there’s additional scenarios worth testing:
- Test that entries are still there 1 second before they should be removed.
- Test that adding 1 entry, then another entry 1 second later, makes it so that after 1 second the first entry has been removed but the second entry remains, and is removed 1 second later.

FilipePintoReis:
I just found another problem with my solution;
If I renew the entry's timestamp when I access that value in the dictionary, I can no longer expect entries to be ordered by time, and thus this logic of removing expired objects until I find one that is not expired no longer works.
I believe that if I remove and add this object from the OrderedDict we can keep this logic, as this way we ensure that the dict is ordered by latest access to each given entry.

scrapy/utils/datatypes.py
```diff
@@ -64,21 +65,57 @@ def pop(self, key, *args):
...
+        now = datetime.now()
+        value = (value, now)
+
+        if self.limit is not None:
             while len(self) >= self.limit:
                 self.popitem(last=False)
+
+        while True:
+            try:
+                val = self.values().__iter__().__next__()
```
elacuesta, line 88:
```suggestion
                val = next(iter(self.values()))
```
Same in the other occurrence below.

wRAR (changes requested):
This needs at least flake8 fixes and implementing the freezegun advice.

PR #5130, opened 2021-05-04
Refactor request to/from dict

elacuesta:
An alternative approach to #4945, considering https://github.com/scrapy/scrapy/pull/4945#discussion_r553504035.

Closes #4945.

~I plan to do the same thing with the `Response` class in a later PR (`Response.to_dict` and `scrapy.utils.response.response_from_dict`)~ see #5131

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -230,7 +230,7 @@ Request objects
 
 
 Other functions related to requests
----------------------------------
+-----------------------------------
 
 .. autofunction:: scrapy.utils.request.request_from_dict
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -205,6 +205,12 @@ Request objects
         ``failure.request.cb_kwargs`` in the request's errback. For more information,
         see :ref:`errback-cb_kwargs`.
 
+    .. attribute:: Request.attributes
+
+        A tuple of :class:`str` objects containing all public attribute names (and ``__init__`` parameters)
+        for the class. Currently used by :meth:`Request.replace`, :meth:`Request.to_dict` and
+        :func:`~scrapy.utils.request.request_from_dict`.
+
     .. method:: Request.copy()
 
        Return a new Request which is a copy of this Request. See also:
@@ -221,6 +227,13 @@ Request objects
     .. automethod:: from_curl
 
     .. automethod:: to_dict
+
+
+Other functions related to requests
+---------------------------------
+
+.. autofunction:: scrapy.utils.request.request_from_dict
+
 
 .. _topics-request-response-ref-request-callback-arguments:
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -220,6 +220,8 @@ Request objects
 
     .. automethod:: from_curl
 
+    .. automethod:: to_dict
+
 .. _topics-request-response-ref-request-callback-arguments:
 
 Passing additional data to callback functions
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -649,6 +649,8 @@ dealing with JSON requests.
        data into JSON format.
    :type dumps_kwargs: dict
 
+   .. autoattribute:: JsonRequest.attributes
+
 JsonRequest usage example
 -------------------------
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -26,10 +26,6 @@ Request objects
 
 .. autoclass:: Request
 
-    A :class:`Request` object represents an HTTP request, which is usually
-    generated in the Spider and executed by the Downloader, and thus generating
-    a :class:`Response`.
-
     :param url: the URL of this request
 
         If the URL is invalid, a :exc:`ValueError` exception is raised.
@@ -205,11 +201,7 @@ Request objects
         ``failure.request.cb_kwargs`` in the request's errback. For more information,
         see :ref:`errback-cb_kwargs`.
 
-    .. attribute:: Request.attributes
-
-        A tuple of :class:`str` objects containing all public attribute names (and ``__init__`` parameters)
-        for the class. Currently used by :meth:`Request.replace`, :meth:`Request.to_dict` and
-        :func:`~scrapy.utils.request.request_from_dict`.
+    .. autoattribute:: Request.attributes
 
     .. method:: Request.copy()
 
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -13,7 +13,7 @@ import scrapy
 from scrapy.http.common import obsolete_setter
 from scrapy.http.headers import Headers
 from scrapy.utils.curl import curl_to_request_kwargs
-from scrapy.utils.python import to_bytes, to_unicode
+from scrapy.utils.python import to_bytes
 from scrapy.utils.trackref import object_ref
 from scrapy.utils.url import escape_ajax
 
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -4,17 +4,26 @@ requests in Scrapy.
 
 See documentation in docs/topics/request-response.rst
 """
+import inspect
+from typing import Optional, Tuple
+
 from w3lib.url import safe_url_string
 
+import scrapy
+from scrapy.http.common import obsolete_setter
 from scrapy.http.headers import Headers
-from scrapy.utils.python import to_bytes
+from scrapy.utils.curl import curl_to_request_kwargs
+from scrapy.utils.python import to_bytes, to_unicode
 from scrapy.utils.trackref import object_ref
 from scrapy.utils.url import escape_ajax
-from scrapy.http.common import obsolete_setter
-from scrapy.utils.curl import curl_to_request_kwargs
 
 
 class Request(object_ref):
+
+    _attributes: Tuple[str, ...] = (
+        "url", "method", "headers", "body", "cookies", "meta", "flags",
+        "encoding", "priority", "dont_filter", "callback", "errback", "cb_kwargs",
+    )
 
     def __init__(self, url, callback=None, method='GET', headers=None, body=None,
                  cookies=None, meta=None, encoding='utf-8', priority=0,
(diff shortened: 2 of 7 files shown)


scrapy/utils/reqser.py
```diff
@@ -1,95 +1,18 @@
...
 
 
-def request_to_dict(request, spider=None):
-    """Convert Request object to a dict.
+warnings.warn(
+    ("Module `scrapy.utils.reqser` is deprecated, please use `scrapy.Request.to_dict` "
+     " and/or`scrapy.utils.request.request_from_dict` instead"),
+    category=ScrapyDeprecationWarning,
+    stacklevel=2,
+)
```
Gallaecio, line 14: There’s a missing space, and I would use nothing, single quotes or double quotes instead of a [grave accent](https://en.wikipedia.org/wiki/Grave_accent).

```suggestion
warnings.warn(
    ("Module scrapy.utils.reqser is deprecated, please use scrapy.Request.to_dict "
     " and/or scrapy.utils.request.request_from_dict instead"),
    category=ScrapyDeprecationWarning,
    stacklevel=2,
)
```

scrapy/http/request/__init__.py
```diff
@@ -136,8 +142,43 @@ def from_curl(cls, curl_command, ignore_unknown_options=True, **kwargs):
...
         To translate a cURL command into a Scrapy request,
         you may use `curl2scrapy <https://michael-shub.github.io/curl2scrapy/>`_.
-
-       """
+        """
         request_kwargs = curl_to_request_kwargs(curl_command, ignore_unknown_options)
         request_kwargs.update(kwargs)
         return cls(**request_kwargs)
+
+    def to_dict(self, spider: Optional["scrapy.Spider"] = None) -> dict:
```
Gallaecio, line 150: I wonder if we should make the parameter keyword-only.

scrapy/utils/request.py
```diff
@@ -106,3 +107,26 @@ def referer_str(request: Request) -> Optional[str]:
...
+
+
+def request_from_dict(d: dict, spider: Optional[Spider] = None) -> Request:
+    """Create Request object from a dict.
+
+    If a spider is given, it will try to resolve the callbacks looking at the
+    spider for methods with the same name.
+    """
+    request_cls = load_object(d["_class"]) if "_class" in d else Request
+    kwargs = {key: value for key, value in d.items() if key in request_cls._attributes}
```
Gallaecio, line 119: Since we expect Request subclasses to extend that variable as needed, maybe it should not be private.

scrapy/utils/reqser.py
elacuesta: No objections here, thanks. Just as a note, this is done in the [`scrapy.utils.py36`](https://github.com/scrapy/scrapy/blob/2.5.0/scrapy/utils/py36.py#L8) module as well.

scrapy/utils/request.py
elacuesta: Makes sense

scrapy/http/request/__init__.py
elacuesta: Sounds good

scrapy/http/request/json_request.py
```diff
@@ -36,6 +40,12 @@ def __init__(self, *args, **kwargs):
         self.headers.setdefault('Content-Type', 'application/json')
         self.headers.setdefault('Accept', 'application/json, text/javascript, */*; q=0.01')
 
+    @property
+    def dumps_kwargs(self):
+        if self._dumps_kwargs is None:
+            self._dumps_kwargs = {}
```
Gallaecio, line 46: Looking at [the coverage results](https://app.codecov.io/gh/scrapy/scrapy/compare/5130/diff#D5L45) and then at `__init__` above, it seems these 2 lines are unnecessary.

docs/topics/request-response.rst
```diff
@@ -205,6 +205,12 @@ Request objects
         ``failure.request.cb_kwargs`` in the request's errback. For more information,
         see :ref:`errback-cb_kwargs`.
 
+    .. attribute:: Request.attributes
+
+        A tuple of :class:`str` objects containing all public attribute names (and ``__init__`` parameters)
+        for the class. Currently used by :meth:`Request.replace`, :meth:`Request.to_dict` and
+        :func:`~scrapy.utils.request.request_from_dict`.
```
Gallaecio, line 212: The current text to me sounds like it’s a superset of attribute names and `__init__` parameters (i.e. a parameter that is only available as one of those would be in `attributes`). Assuming I’m not getting the implementation wrong, what about:

```suggestion
        A tuple of :class:`str` objects containing the name of all public 
        attributes of the class that are also keyword parameters of the 
        ``__init__`` method.
        
        Currently used by :meth:`Request.replace`, :meth:`Request.to_dict` and
        :func:`~scrapy.utils.request.request_from_dict`.
```

I also wonder if it would not be better to make this a docstring and use `.. autodata` or similar here.

scrapy/utils/request.py
```diff
@@ -106,3 +107,27 @@ def referer_str(request: Request) -> Optional[str]:
     if referrer is None:
         return referrer
     return to_unicode(referrer, errors='replace')
+
+
+def request_from_dict(d: dict, spider: Optional[Spider] = None) -> Request:
```
Gallaecio, line 112:
```suggestion
def request_from_dict(d: dict, *, spider: Optional[Spider] = None) -> Request:
```

tests/test_request_dict.py
```diff
@@ -130,7 +138,22 @@ def parse(self, response):
...
+        """Callback method is not available in the spider passed to from_dict"""
+        spider = TestSpiderDelegation()
+        r = Request("http://www.example.com", callback=spider.delegated_callback)
+        d = r.to_dict(spider=spider)
+        self.assertRaises(ValueError, request_from_dict, d, spider=Spider("foo"))
+
+
+class DeprecatedMethodsRequestSerializationTest(RequestSerializationTest):
+    def _assert_serializes_ok(self, request, spider=None):
+        from scrapy.utils.reqser import request_from_dict as _request_from_dict, request_to_dict as _request_to_dict
```
Gallaecio, line 153: I’m not sure it’s worth it, but for completeness maybe we could catch the deprecation warning logged by this import, just for the sake of verifying it’s there.

elacuesta: I actually attempted to do so, but I kept having issues with the warnings not being reset and appearing only once, which made only the first check succeed. Consider the following:
```python
Python 3.8.2 (default, Apr 18 2020, 17:39:30) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import warnings

In [2]: warnings.simplefilter("always")

In [3]: from scrapy.utils.reqser import request_to_dict
<ipython-input-3-75e30ca4f2be>:1: ScrapyDeprecationWarning: Module scrapy.utils.reqser is deprecated, please use scrapy.Request.to_dict  and/or scrapy.utils.request.request_from_dict instead
  from scrapy.utils.reqser import request_to_dict

In [4]: warnings.resetwarnings()

In [5]: del request_to_dict

In [6]: from scrapy.utils.reqser import request_to_dict  # no warning here

In [7]: 
```

It seems like this is caused by the warning being raised at module level. It does work if it's deleted with `del sys.modules["scrapy.utils.reqser"]`, it should be safe within a test.

scrapy/utils/request.py
elacuesta: Done in 1277a7c1385cbec9abce330dcf0e9d4e6ffa8126

docs/topics/request-response.rst
elacuesta: Nice, I didn't know attribute docstrings were a thing, thanks.

Gallaecio (approved):

wRAR (approved):

scrapy/http/request/__init__.py
```diff
@@ -4,17 +4,36 @@
...
 
 
 class Request(object_ref):
+    """Represents an HTTP request, which is usually generated in a Spider and
+    executed by the Downloader, thus generating a :class:`Response`.
+    """
+
+    attributes: Tuple[str, ...] = (
+        "url", "method", "headers", "body", "cookies", "meta", "flags",
+        "encoding", "priority", "dont_filter", "callback", "errback", "cb_kwargs",
```
kmike, line 28: A nitpick: would it make sense to use the same order as in `__init__`, to make it easier to ensure they match?

scrapy/http/request/__init__.py
```diff
@@ -136,8 +152,43 @@ def from_curl(cls, curl_command, ignore_unknown_options=True, **kwargs):
...
         request_kwargs.update(kwargs)
         return cls(**request_kwargs)
+
+    def to_dict(self, *, spider: Optional["scrapy.Spider"] = None) -> dict:
+        """Return a dictionary containing the Request's data.
+
+        Use :func:`~scrapy.utils.request.request_from_dict` to convert back into a :class:`~scrapy.Request` object.
+
+        If a spider is given, this method will try to find out the name of the spider method used
+        as callback and include it in the output dict, raising an exception if it cannot be found.
```
kmike, line 166: this also applies to errback

scrapy/utils/reqser.py
```diff
@@ -1,95 +1,22 @@
...
-from scrapy.utils.misc import load_object
+import scrapy
+from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.utils.request import request_from_dict as _from_dict
 
 
-def request_to_dict(request, spider=None):
-    """Convert Request object to a dict.
+warnings.warn(
+    ("Module scrapy.utils.reqser is deprecated, please use scrapy.Request.to_dict"
```
kmike, line 10:
```suggestion
    ("Module scrapy.utils.reqser is deprecated, please use request.to_dict method"
```

What do you think about ^^? It is a bit more confusing, but a bit more precise as well; e.g. if you're serializing a custom request, you shouldn't be using scrapy.Request.to_dict.

kmike (approved):
Thanks @elacuesta! I've added a few minor comments, but overall the PR looks good.

PR #5131, opened 2021-05-04
Response to/from dict, HTTP cache updates

### LINKED ISSUE #1450: Response.to_dict()

### LINKED ISSUE #1877: Allow to add custom metadata to Response

### LINKED ISSUE #4923: Store more data in the HTTP cache

elacuesta:
Continuation of #5130. Closes #1450, closes #1877, closes #4923.

This PR introduces a new feature to (de)serialize responses, which is used in the HTTP cache extension.


scrapy/http/response/__init__.py
```diff
@@ -206,3 +217,21 @@ def follow_all(self, urls, callback=None, method='GET', headers=None, body=None,
...
+
+        Use :func:`~scrapy.utils.response.response_from_dict` to convert
+        back into a :class:`~scrapy.http.response.Response` object.
+        """
+        d = {
+            "url": to_unicode(self.url),  # urls are safe (safe_string_url)
+            "headers": dict(self.headers),
+            "ip_address": str(self.ip_address) if self.ip_address is not None else None,
+            "certificate": self.certificate.dumpPEM() if self.certificate is not None else None,
+        }
```
elacuesta, line 232: Should the request attribute be converted to a dict as well? 🤔

kmike:
Hey! Unlike request <=> dict conversion, which is used by Scrapy e.g. for the disk queues, here it seems the newly introduced method and function are not used by Scrapy itself. E.g. it is not used for storing of the cache data. I think that's not a great situation, because we're almost introducing an extension point, protocol for response serialization / deserialization here, but not using it in the framework itself. 

What's the use case you see for these feature as it is now? Is it just debugging, as in the original ticket?

elacuesta:
@kmike I started working on this (on #4945, rather) after #4923. This would provide a way of dealing with cached responses in an uniform fashion, taking the burden off from the cache implementations themselves.

Gallaecio:
I do wonder if maybe it makes sense to implement this along with the changes to cache handling that make use of this.

elacuesta:
This is technically backward-incompatible, but I don't think that's too much of a problem in practice (I believe invalidating an existing cache after a Scrapy version change is probably reasonable).

Gallaecio:
I’m unsure (+0) on whether or not it’s fine.

I do wonder if merging this change would affect https://github.com/scrapy/scrapy/pull/5175 in any way (as in https://github.com/scrapy/scrapy/pull/5175 could be implemented differently if we are going to break caches anyway). I’m not sure either, though.

Gallaecio:
@elacuesta What would you think of extracting a subset of these changes into a separate pull request, specifically `Response.attributes` and its use in `replace`, to address https://github.com/scrapy/scrapy/issues/1877? I suspect we’ll be able to merge that sooner.

PR #5137, opened 2021-05-08
Added unit tests and implementation for issue #5135

### LINKED ISSUE #5135: Remove UrlLengthMiddleware from default enabled middlewares
According [RFC2396](http://www.faqs.org/rfcs/rfc2396.html), section 3.2.1:
```
   The HTTP protocol does not place any a priori limit on the length of a URI.
   Servers MUST be able to handle the URI of any resource they serve, and
   SHOULD be able to handle URIs of unbounded length if they provide 
   GET-based forms that could generate such URIs. A server SHOULD 
   return 414 (Request-URI Too Long) status if a URI is longer than the server
   can handle (see section 10.4.15).
```

We have enabled by default `scrapy.spidermiddlewares.urllength.UrlLengthMiddleware` that has a default limit defined by `URLLENGTH_LIMIT` setting (that can be modified by in project settings) set to `2083`. As [mentioned here](https://github.com/scrapy/scrapy/pull/5134), the reason for this number is related to limits of Microsoft Internet Explorer to handle URIs longer than that. 

This can cause problems to spiders that will skip requests of URIs longer than that. Certainly we can change `URLLENGTH_LIMIT` on these spiders, but sometimes is not easy to set the right value and we chose to set a higher number just to make the middleware happy. This is what I am doing in a real world project, but the solution doesn't look good.

I know that we can or disable the middleware, or change the length limit, but I think it is smoother for the user not to have to worry about this artificial limit we have on Scrapy. We are not using Microsoft Internet Explorer, we don't need this limit.

Some alternatives that I considered:

- Remove `UrlLengthMiddleware` as a default enabled middlewares, so we don't need to worry about that limit unless we really need to worry about that (I don't know the exact use-case that required this limit, so keeping the middleware available may make sense);
- Change the default value to a more reasonable (difficult to find a reasonable value)

kno3comma14:
I am submitting a solution to issue #5135 . 

### Procedure
- Changed `URLLENGTH_LIMIT = -1` in settings.default_settings
- Changed urllength.py of verify if self.maxlength == -1 then just don't apply the filter
- Added new asserts to respective package

Resolves #5135.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1620,7 +1620,8 @@ Scope: ``spidermiddlewares.urllength``
 
 The maximum URL length to allow for crawled URLs. For more information about
 the default value for this setting see: https://support.microsoft.com/en-us/topic/maximum-url-length-is-2-083-characters-in-internet-explorer-174e7c8a-6666-f4e0-6fd6-908b53c12246
-If URLLENGTH_LIMIT is equals to ``0``, then the :class:`UrlLengthMiddleware` middleware is disabled.
+
+If URLLENGTH_LIMIT is ``0``, then the :class:`scrapy.spidermiddlewares.urllength.UrlLengthMiddleware` middleware is disabled.
 
 .. setting:: USER_AGENT
 
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1620,6 +1620,7 @@ Scope: ``spidermiddlewares.urllength``
 
 The maximum URL length to allow for crawled URLs. For more information about
 the default value for this setting see: https://support.microsoft.com/en-us/topic/maximum-url-length-is-2-083-characters-in-internet-explorer-174e7c8a-6666-f4e0-6fd6-908b53c12246
+If URLLENGTH_LIMIT is equals to ``0``, then the :class:`UrlLengthMiddleware` middleware is disabled.
 
 .. setting:: USER_AGENT
 
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -285,7 +285,7 @@ STATSMAILER_RCPTS = []
 
 TEMPLATES_DIR = abspath(join(dirname(__file__), '..', 'templates'))
 
-URLLENGTH_LIMIT = 0
+URLLENGTH_LIMIT = 2083
 
 USER_AGENT = f'Scrapy/{import_module("scrapy").__version__} (+https://scrapy.org)'
 
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -285,7 +285,7 @@ STATSMAILER_RCPTS = []
 
 TEMPLATES_DIR = abspath(join(dirname(__file__), '..', 'templates'))
 
-URLLENGTH_LIMIT = -1
+URLLENGTH_LIMIT = 0
 
 USER_AGENT = f'Scrapy/{import_module("scrapy").__version__} (+https://scrapy.org)'
 
--- a/scrapy/settings/default_settings.py
+++ b/scrapy/settings/default_settings.py
@@ -285,7 +285,7 @@ STATSMAILER_RCPTS = []
 
 TEMPLATES_DIR = abspath(join(dirname(__file__), '..', 'templates'))
 
-URLLENGTH_LIMIT = 2083
+URLLENGTH_LIMIT = -1
 
 USER_AGENT = f'Scrapy/{import_module("scrapy").__version__} (+https://scrapy.org)'
 
--- a/scrapy/spidermiddlewares/urllength.py
+++ b/scrapy/spidermiddlewares/urllength.py
@@ -8,6 +8,7 @@ import logging
 
 from scrapy.http import Request
 from scrapy.exceptions import NotConfigured
+from scrapy.settings import default_settings
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +21,7 @@ class UrlLengthMiddleware:
     @classmethod
     def from_settings(cls, settings):
         maxlength = settings.getint('URLLENGTH_LIMIT')
-        if not maxlength:
+        if maxlength is None:
             raise NotConfigured
         return cls(maxlength)
 
@@ -37,7 +38,7 @@ class UrlLengthMiddleware:
             else:
                 return True
 
-        if self.maxlength == -1:
+        if self.maxlength == default_settings.URLLENGTH_LIMIT:
             return result
         else:
             return (r for r in result or () if _filter(r))
--- a/scrapy/spidermiddlewares/urllength.py
+++ b/scrapy/spidermiddlewares/urllength.py
@@ -21,7 +21,7 @@ class UrlLengthMiddleware:
     def from_settings(cls, settings):
         maxlength = settings.getint('URLLENGTH_LIMIT')
         if not maxlength:
-           raise NotConfigured
+            raise NotConfigured
         return cls(maxlength)
 
     def process_spider_output(self, response, result, spider):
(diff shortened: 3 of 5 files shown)


Gallaecio:
Let’s first agree on how to handle this in https://github.com/scrapy/scrapy/issues/5135

Gallaecio:
https://github.com/scrapy/scrapy/issues/5135#issuecomment-836678049

scrapy/settings/default_settings.py
```diff
@@ -285,7 +285,7 @@
 
 TEMPLATES_DIR = abspath(join(dirname(__file__), '..', 'templates'))
 
-URLLENGTH_LIMIT = 2083
+URLLENGTH_LIMIT = 0
```
Gallaecio, line 288: The agreement is to leave this unchanged, and simply _allow_ users to set it to 0 in their settings if they wish.

scrapy/spidermiddlewares/urllength.py
```diff
@@ -20,7 +21,7 @@ def __init__(self, maxlength):
     @classmethod
     def from_settings(cls, settings):
         maxlength = settings.getint('URLLENGTH_LIMIT')
-        if not maxlength:
+        if maxlength is None:
```
Gallaecio, line 24: The expectation is for it to be 0.

```suggestion
        if not maxlength:
```

scrapy/spidermiddlewares/urllength.py
```diff
@@ -37,4 +38,7 @@ def _filter(request):
             else:
                 return True
 
-        return (r for r in result or () if _filter(r))
+        if self.maxlength == default_settings.URLLENGTH_LIMIT:
+            return result
+        else:
+            return (r for r in result or () if _filter(r))
```
Gallaecio, line 44: We do not need to change anything here. The whole middleware will be disabled if the value is 0, so this code will not even be reached in that case.

tests/test_spidermiddleware_urllength.py
```diff
@@ -20,22 +20,36 @@ def setUp(self):
...
         with LogCapture() as log:
             self.process_spider_output()
+            self.process_spider_output_infinite()
 
         ric = self.stats.get_value('urllength/request_ignored_count', spider=self.spider)
         self.assertEqual(ric, 1)
-
         self.assertIn(f'Ignoring link (url length > {self.maxlength})', str(log))
+        ric = self.stats_infinite.get_value('urllength/request_ignored_count', spider=self.spider_infinite)
+        self.assertIsNone(ric)
```
Gallaecio, line 55: I think this test method should remain unchanged, and a new one should be created.

Also, best tests usually test the limits where behavior change, and in this case I think it would be best to also test that, if the max length is set to 1, all requests are filtered out.

In a similar fashion, and this is _not_ a requirement since it’s not strictly related to this pull request, but I think the existing tests could be refactored to:
- Have a single URL instead of 2.
- Test the maximum max length that results in the request being filtered out (length of the request URL, minus 1).
- Test the minimum max length that results in the request not being filtered out (exact length of the request URL).

So, we would be testing the following limits: 0 (no filtering), 1 (filtering), URL length - 1 (filtering) URL length (no filtering).

tests/test_spidermiddleware_urllength.py
```diff
@@ -20,22 +20,36 @@ def setUp(self):
...
+
+        crawler_infinite = get_crawler(Spider)
+        self.spider_infinite = crawler_infinite._create_spider('foo_infinite')
+        self.stats_infinite = crawler_infinite.stats
+        self.mw_infinite = UrlLengthMiddleware.from_settings(settings_infinite)
         self.response = Response('http://scrapytest.org')
+
         self.short_url_req = Request('http://scrapytest.org/')
         self.long_url_req = Request('http://scrapytest.org/this_is_a_long_url')
         self.reqs = [self.short_url_req, self.long_url_req]
```
Gallaecio, line 34: This `setUp` method was written for a test case that was testing a single scenario. Now that the test is going to cover more than that, I think it makes sense to, rather than extending `setUp`, to move stuff out of `setUp`, and leave in `setUp` only whatever is common to all scenarios (if anything).

scrapy/spidermiddlewares/urllength.py
kno3comma14: A few questions here: 
- When you talk about disabling the middleware, you mean doing something like the code below right?
 ```python
settings.getdict('SPIDER_MIDDLEWARES')[spider_middleware_class] = None
```
Note: In this particular case spider_middleware_class == 'scrapy.spidermiddlewares.urllength.UrlLengthMiddleware'

- On the other hand, as a consequence of this change the expected behavior will be a `None` value when we use `from_settings(cls, settings)` method right? I am asking this for testing purposes.

File: tests/test_spidermiddleware_urllength.py (hunk 4 above)
Comment (kno3comma14) [reply tests/test_spidermiddleware_urllength.py:55]: I think I can apply the refactor for the tests but I will need the previous questions to be answered because if I disable the middleware at settings level(Using [this](https://docs.scrapy.org/en/latest/topics/spider-middleware.html) documentation as reference), I will have to deal with a `None` middleware object in the tests so it will be impossible to use process_spider_output method for URLLENGTH_LIMIT = 0 case. Probably I am confused about the 'disable middleware' term meaning, so please clarify it to me

File: scrapy/spidermiddlewares/urllength.py (hunk 3 above)
Comment (Gallaecio) [reply scrapy/spidermiddlewares/urllength.py:44]: When you raise `NotConfigured` during the component initialization (i.e. `from_settings`, `from_crawler` or `__init__`), the component is disabled. The effect is the same as the code you wrote there, setting the class to `None` on the corresponding setting, but users do not need to do that. In other words, with your change settting `MAX_URLLENGTH` to `0` becomes an alternative way to achieve the same effect as setting that middleware class to `None` in the corresponding setting.

**Review by Gallaecio — COMMENTED** [review]
I’ve just realized that the current code already disabled the middleware if the value is `0`. So I think we just need to add tests for that, and make sure it’s documented.

File: tests/test_spidermiddleware_urllength.py (hunk 4 above)
Comment (kno3comma14) [reply tests/test_spidermiddleware_urllength.py:55]: With all these information, I think we can set three states here:
1) `URLLENGTH_LIMIT = 0`: For this one we could use the disable method from the settings. I think this is more like an special configuration than an empty one, so raising a NotConfigured exception would be confusing. In this case, we have to test that any time we create a UrlLengthMiddleware with from_settings it will return None.
2) `URLLENGTH_LIMIT != 0`: Normal behavior
3) `URLLENGTH_LIMIT` doesn't exist in settings: Maybe for this case we can raise a `NotConfigured` exception. This action could be done when some users do something that make impossible to load `URLLENGTH_LIMIT`(ex. del method or a buggy update of the settings). For this case, we have to test against the raising of a `NotConfigured` exception.

Does this make sense to you?

File: docs/topics/settings.rst [hunk 6]
```diff
@@ -1620,6 +1620,7 @@ Scope: ``spidermiddlewares.urllength``
 
 The maximum URL length to allow for crawled URLs. For more information about
 the default value for this setting see: https://support.microsoft.com/en-us/topic/maximum-url-length-is-2-083-characters-in-internet-explorer-174e7c8a-6666-f4e0-6fd6-908b53c12246
+If URLLENGTH_LIMIT is equals to ``0``, then the :class:`UrlLengthMiddleware` middleware is disabled.
```
Gallaecio, line 1623: There should be an empty line above to make this new line a separate paragraph, otherwise it is appended to the end of the text above.

There’s also a minor grammatical number discordance (“equals” should be “equal”), but you can actually just remove “equals to” altogether and the sentence will have the same meaning with fewer words, which I personally tend to favor :)

Finally, you might want to render the documentation locally (`tox -e docs`) and check if that class link works. You may need to replace it with the full import path, which you can prepend with a tilde so that the display text is only the class name: ``` :class:`scrapy.spidermiddlewares.urllength.UrlLengthMiddleware` ```.

PR #5139, opened 2021-05-09
[Draft] Development Dockerfile

### LINKED ISSUE #5138: Development Dockerfile and Setup Documentation
## Summary

As a contributor to Scrapy, I'd like to have a dockerized environment to make changes to Scrapy without needing to set up what I need locally. Besides that, I'd also like to have detailed information on how to install all dependencies that Scrapy requires, in case I want to develop locally.

## Motivation

I started trying to contribute to Scrapy, but I didn't find any tutorial or explanation on how to make changes to the source code and testing them afterward. From this issue I expect a docker file and documentation page to help new wanna-be contributors to start their journey.

FilipePintoReis:
Closes #5138 
This is just a draft, I believe there's a better way to do this, so I'm open to criticism.

- Added a docker file that:
  - Installs Ubuntu image.
  - Installs Anaconda
  - Sets up Scrapy's dependencies
  - Copies repository, tutorial folder, and a file to run cmdline commands.
  - Runs mentioned file.
- A mock project.
- A file to run cmdline commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,47 @@
+# We will use Ubuntu for our image
+FROM ubuntu:latest
+
+# Updating Ubuntu packages
+RUN apt-get update && yes|apt-get upgrade
+RUN apt-get install -y emacs
+
+# Adding wget and bzip2
+RUN apt-get install -y wget bzip2
+
+# Add sudo
+RUN apt-get -y install sudo
+
+# Add user ubuntu with no password, add to sudo group
+RUN adduser --disabled-password --gecos '' ubuntu
+RUN adduser ubuntu sudo
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+USER ubuntu
+WORKDIR /home/ubuntu/
+RUN chmod a+rwx /home/ubuntu/
+
+# AnaRun conda installing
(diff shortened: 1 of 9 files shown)


wRAR:
You don't need conda, you can use pip. And I don't think adding the tutorial files to the repo for the sole purpose of building this image is a good idea.

FilipePintoReis:
> You don't need conda, you can use pip. And I don't think adding the tutorial files to the repo for the sole purpose of building this image is a good idea.

@wRAR I agree with you, I was sort of reluctant in adding the tutorial folder.
Would it seem better to you if I replace conda with pip, delete the tutorial and dockerrun files, and delete the call to dockerrun inside the docker file?

Gallaecio:
You [said](https://github.com/scrapy/scrapy/issues/5138#issuecomment-835843000):

> I argue a docker file is useful if you don't want to install Anaconda locally for whatever reason.

Since you don’t need to install Anaconda locally, I don’t see why we need a (pip-based) Dockerfile. Creating a Dockerfile for Scrapy should be trivial for whoever wants to do it, and I’m not sure there’s a strong argument to provide one.

If there’s something special about Scrapy that people using Docker need to know, I think we could cover that in the Contributing page. And there may be other improvements that we can do to that page. But I’m not sure we need more.

PR #5142, opened 2021-05-10
Document coroutines for signals.

### LINKED ISSUE #4852: Document async def interfaces for signals
It'd be nice to mention `async def` approach in https://docs.scrapy.org/en/latest/topics/signals.html#deferred-signal-handlers, and maybe even rewrite the example to it.

burakcank:
I wrote the part considering the comments of @Gallaecio. Might need more improvement, waiting for feedback.

Fixes #4852.

### DIFF
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -88,8 +88,29 @@ Let's take an example::
                     'tags': quote.css('div.tags a.tag::text').getall(),
                 }
 
+Coroutines are also supported as can be seen :ref:`here <topics-coroutines>`.
+This is a more transparent approach and should be preferred over
+:class:`~twisted.internet.defer.Deferred` whenever possible.
+
+We will use the same concept as above::
+
+    import aiohttp
+    import asyncio
+
+
+    class SignalSpider(scrapy.Spider):
+        ...
+
+        async def item_scraped(self, item):
+            async with aiohttp.ClientSession() as session:
+                async with session.post('http://example.com/post', json=item) as response:
+                    html = await response.text()
+                    print("Body: ", html[:20])
+
+        ...
+
 See the :ref:`topics-signals-ref` below to know which signals support
-:class:`~twisted.internet.defer.Deferred`.
+:class:`~twisted.internet.defer.Deferred` and :term:`awaitable objects <awaitable>`.
 
 .. _topics-signals-ref:
 
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -51,12 +51,12 @@ Deferred signal handlers
 ========================
 
 Some signals support returning :class:`~twisted.internet.defer.Deferred`
-objects from their handlers, allowing you to run asynchronous code that
-does not block Scrapy. If a signal handler returns a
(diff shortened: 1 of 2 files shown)


docs/topics/signals.rst
```diff
@@ -88,8 +88,29 @@ Let's take an example::
...
                 }
 
+Coroutines are also supported as can be seen :ref:`here <topics-coroutines>`.
+This is a more transparent approach and should be preferred over
+:class:`~twisted.internet.defer.Deferred` whenever possible.
+
+We will use the same concept as above::
+
+    import aiohttp
+    import asyncio
```
elacuesta, line 98:
```suggestion
    import asyncio

    import aiohttp
```
As per [PEP 8 (imports section)](https://www.python.org/dev/peps/pep-0008/#imports)

Gallaecio (commented):
In #4852 @kmike suggests rewriting the example with async def. Since that makes it more readable, I think we should do that.

Note that asyncio and aiohttp are not needed, the example with async def should still use treq, which is Twisted-based. To use those libraries, users would need to use the asyncio reactor, which would make the example more complex rather than simpler.

A common mistake that I and most people make at first is that coroutine implies asyncio, which turns out not to be true. That’s why we have separate documentation topics for coroutines and asyncio: you can use coroutines without using asyncio.

So, I would rewrite the example to use async def but otherwise be identical to how it is now, and I would change its introduction paragraph to something like:

```
Let's take an example using :ref:`coroutines <topics-coroutines>`::
```

That way, the example becomes simpler, but for people wishing _not_ to use coroutines they have a link to the coroutines documentation, which provides an example with and without coroutines, so that people that prefer or need to use Deferred objects instead can still learn how.

burakcank:
I realized that asyncio and aiohttp are not needed like you said, reverted back the `treq` section while changing the preceding example to use the coroutine syntax altogether. I might still be inaccurate with my terms, let me know if anything is not clear.

wRAR (approved):

docs/topics/signals.rst
```diff
@@ -68,17 +68,15 @@ Let's take an example::
...
                 'http://example.com/post',
                 json.dumps(item).encode('ascii'),
                 headers={b'Content-Type': [b'application/json']}
             )
-
-            # The next item will be scraped only after
-            # deferred (d) is fired
-            return d
+            content = await treq.content(response)
+            return content
```
Gallaecio, line 79: I don’t think these last 2 lines are correct. I believe, to reproduce the original behavior, you just need to return the result of `treq.post`, and let the caller of `item_scraped` await it. The original code does not wait for the response content, and does not return it.

@wRAR Thoughts?

wRAR: The old example: make a POST request, return a Deferred that fires when the request is completed. The result of that Deferred will be a response object.

The new example that does the same would be `return response`.

On the other hand, I don't think results of signal handlers, at least the `item_scraped` one, are used anywhere, so ending the method on just `await treq.post()` would do the same.

burakcank: @wRAR I fixed it as you suggested. Leaving the `return` statement just in case.

burakcank:
Continue with #5358.

PR #5143, opened 2021-05-11
Multiple-compressed responses support

divtiply:
HTTP response body can be compressed multiple times. In such cases `Content-Encoding` header contains a list of comma-separated encodings. Some servers instead can send multiple `Content-Encoding` headers.
This fix allows HttpCompressionMiddleware handle that.
See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding


Gallaecio:
Nice, I did not know that.

Could you please include a test?

wRAR:
Please note that the existing tests failed too.

Issue #1459, opened 2015-08-26
Headers object serializes to json differently depending on the indent value


dvdbng:
A scrapy.http.headers.Headers instance will serialize to a shallow object if indent=None and wrap the values in a list if indent is a number:

```
>>> from scrapy.http.headers import Headers
>>> h = Headers({ "X-Foo": "bar" })
>>> print json.dumps(h)
{"X-Foo": "bar"}
>>> print json.dumps(h, indent=3)
{
   "X-Foo": [
      "bar"
   ]
}
```

Tested in Scrapy 1.0.3

redapple:
Indeed. Weird.

elacuesta:
This is in fact weird, but (at least in python 2.7.10) it's not caused _only_ by passing the `indent` argument; for instance, a correct output can be obtained by passing `sort_keys=True`.

Consider the following piece of code:
```
import inspect
from json import dumps

class CustomDict(dict):
    def _inspect(self):
        current_frame = inspect.currentframe()
        outer_frames = inspect.getouterframes(current_frame)
        for frame in outer_frames:
            print(frame[1:-2])

    def __getitem__(self, key):
        self._inspect()
        return super(CustomDict, self).__getitem__(key)

    def items(self):
        self._inspect()
        return super(CustomDict, self).items()

    def iteritems(self):
        self._inspect()
        return ((k, v) for k, v in super(CustomDict, self).iteritems())
```

This allows us to get the call chain:

```
>>> d = CustomDict(key=['value'])
>>> print dumps(d)
('<stdin>', 4, '_inspect')
('<stdin>', 8, '__getitem__')
('/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py', 270, 'iterencode')
('/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py', 207, 'encode')
('/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py', 243, 'dumps')
('<stdin>', 1, '<module>')
{"key": ["value"]}
>>> print dumps(d, indent=4)
('<stdin>', 4, '_inspect')
('<stdin>', 11, 'iteritems')
('/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py', 363, '_iterencode_dict')
[...]

PR #5146, opened 2021-05-12
CaseInsensitiveDict (deprecate CaselessDict)

elacuesta:
I think the data model for `scrapy.http.headers.Headers` (and `scrapy.utils.datatypes.CaselessDict`) is a little bit inconsistent. For instance:
```python
In [1]: from scrapy.http.headers import Headers

In [2]: h = Headers({"a": "aaa", "b": "bbb"})

In [3]: h
Out[3]: {b'A': b'aaa', b'B': b'bbb'}

In [4]: dict(h)
Out[4]: {b'A': [b'aaa'], b'B': [b'bbb']}

In [5]: list(h.items())
Out[5]: [(b'A', [b'aaa']), (b'B', [b'bbb'])]

In [6]: list(h.values())
Out[6]: [b'aaa', b'bbb']
```

This has appeared before in #1459 (not reproducible anymore on py3, but the idea remains). `dict` is highly optimized at the C level and it seems like it doesn't always use overriden data access methods. For instance, `MutableMapping` implementations (like `itemadapter.ItemAdapter`) use a combination of `keys` (which I believe relies on `__iter__`) and `__getitem__` when wrapped with `dict`, while `dict` subclasses seem to bypass this:
```python
import inspect

from itemadapter import ItemAdapter


class InspectMixin:
    def _inspect(self):
        current_frame = inspect.currentframe()
        outer_frames = inspect.getouterframes(current_frame)
        for frame in outer_frames:
            print(frame[1:-2])

    def __getitem__(self, key):
        self._inspect()
        return super().__getitem__(key)

    def keys(self):
        self._inspect()
        return super().keys()


class CustomDict(InspectMixin, dict):
    pass


class CustomItemAdapter(InspectMixin, ItemAdapter):
    pass


d = CustomDict({"a": 1})
print("CustomDict:", dict(d))

print()

d = CustomItemAdapter({"a": 1})
print("CustomItemAdapter", dict(d))
```
```
$ python dictionary.py
CustomDict: {'a': 1}

('dictionary.py', 9, '_inspect')
('dictionary.py', 18, 'keys')
('dictionary.py', 36, '<module>')
('dictionary.py', 9, '_inspect')
('dictionary.py', 14, '__getitem__')
('dictionary.py', 36, '<module>')
CustomItemAdapter {'a': 1}
```

[This article](https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/) shows a few reasons why inheriting from `dict` is not a good idea.

I tried to use `CaseInsensitiveDict` as base class for `Headers`, but some tests fail (because they wrap the headers objects with `dict` before checking the contents). Personally, I believe the `Headers` data model needs a little refactoring, I'd love to hear your thought about it (should I continue this PR to include changes to the `Headers` class, for instance?).


scrapy/http/headers.py
```diff
@@ -76,13 +77,13 @@ def to_string(self):
...
         and unicode values. Multiple values are joined with ','.
         """
-        return CaselessDict(
-            (to_unicode(key, encoding=self.encoding),
-             to_unicode(b','.join(value), encoding=self.encoding))
-            for key, value in self.items())
+        return CaseInsensitiveDict(
+            (to_unicode(key, encoding=self.encoding), to_unicode(b','.join(value), encoding=self.encoding))
+            for key, value in self.items()
+        )
```
elacuesta, line 86: This change should be safe I think, `CaseInsensitiveDict`'s interface is the same as `CaselessDict`. The method is only used [here](https://github.com/scrapy/scrapy/blob/2.5.0/scrapy/core/downloader/handlers/s3.py#L67).

scrapy/utils/datatypes.py
```diff
@@ -63,6 +79,36 @@ def pop(self, key, *args):
...
+    def __getitem__(self, key: AnyStr) -> Any:
+        return super().__getitem__(self.normkey(key))
+
+    def __setitem__(self, key: AnyStr, value: Any) -> None:
+        super().__setitem__(self.normkey(key), self.normvalue(value))
+
+    def __delitem__(self, key: AnyStr) -> None:
+        super().__delitem__(self.normkey(key))
+
+    def __contains__(self, key: AnyStr) -> bool:  # type: ignore[override]
```
elacuesta, line 97: Even if this goes against [Liskov substitution principle](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides), it makes sense because being case-insensitive only applies to str or bytes, not to all hashable types.

scrapy/http/headers.py
Gallaecio: It’s still a potential issue for `isinstance` usages. I’m not saying I’m against this change, but I am unsure.

elacuesta: This should address the `isinstance` check:
```diff
diff --git scrapy/utils/datatypes.py scrapy/utils/datatypes.py
index 6eeabe1e..63c07b48 100644
--- scrapy/utils/datatypes.py
+++ scrapy/utils/datatypes.py
@@ -21,7 +21,7 @@ class CaselessDict(dict):
     def __new__(cls, *args, **kwargs):
         from scrapy.http.headers import Headers
 
-        if issubclass(cls, CaselessDict) and not issubclass(cls, Headers):
+        if issubclass(cls, CaselessDict) and not issubclass(cls, (CaseInsensitiveDict, Headers)):
             warnings.warn(
                 "scrapy.utils.datatypes.CaselessDict is deprecated,"
                 " please use scrapy.utils.datatypes.CaseInsensitiveDict instead",
@@ -79,7 +79,7 @@ class CaselessDict(dict):
         return dict.pop(self, self.normkey(key), *args)
 
 
-class CaseInsensitiveDict(collections.UserDict):
+class CaseInsensitiveDict(collections.UserDict, CaselessDict):
     """A dict-like structure that accepts strings or bytes as keys and allows case-insensitive lookups.
 
     It also allows overriding key and value normalization by defining custom `normkey` and `normvalue` methods.
```

Although as we discussed this might not be a problem, mostly because `CaselessDict` is not documented.

scrapy/utils/datatypes.py
```diff
@@ -63,6 +79,46 @@ def pop(self, key, *args):
...
+    It also allows overriding key and value normalization by defining custom `normkey` and `normvalue` methods.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        self._keys: dict = {}
+        super().__init__(*args, **kwargs)
+
+    def __getitem__(self, key: AnyStr) -> Any:
+        normalized_key = self.normkey(key)
+        return super().__getitem__(self._keys[normalized_key.lower()])
```
Gallaecio, line 94: Shouldn’t `.lower()` remain in `self.normkey()`? It feels like we are hardcoding key normalization here.

On the one hand, it suits the class name. Maybe what we should have is a parent class with most of the functionality and this class simply setting `normkey`? Or simply have the docstrings clarify that the caseinsensitive part is not customizable in subclasses.

On the other hand, since the argument not to worry too much about backward incompatibilities with the previous class was that it was not documented, maybe we should make this class private, and if so maybe we do not need to offer facilities for subclassing.

Gallaecio (approved):

scrapy/utils/datatypes.py
elacuesta: The intention is to hardcode the normalization of the access, while delegating the normalization of the keys/values themselves to `normkkey/normvalue`. Consider the following example:

```python
In [1]: from scrapy.utils.datatypes import CaseInsensitiveDict

In [2]: CaseInsensitiveDict({"FOO": "bar"})  # key remains uppercase
Out[2]: <CaseInsensitiveDict: {'FOO': 'bar'}>

In [3]: CaseInsensitiveDict({"FOO": "bar"})["foo"]  # access via lowercase key
Out[3]: 'bar'

In [4]: class TitleCaseCaseInsensitiveDict(CaseInsensitiveDict):
   ...:    def normkey(self, key):
   ...:        return key.title()
   ...: 

In [5]: TitleCaseCaseInsensitiveDict({"FOO": "bar"})  # key is converted to title case
Out[5]: <TitleCaseCaseInsensitiveDict: {'Foo': 'bar'}>

In [6]: TitleCaseCaseInsensitiveDict({"FOO": "bar"})["foo"]  # access via lowercase key
Out[6]: 'bar'
```

My plan was to override this in the new `Headers` implementation, to go from `header-name` to `Header-Name`, but that probably only needs to be an implementation detail and not part of the public interface (in fact, we might not event need or want to do that, if Twisted capitalizes headers on its own). For now I'll rename to `_normkey/_normvalue`, we might even remove them altogether later.

Gallaecio (approved):

scrapy/utils/datatypes.py
kmike: I wonder if it's possible to inherit from some type-enabled dict

wRAR (approved):

wRAR:
Thanks!

PR #5148, opened 2021-05-14
Added caution to follow_all method

### LINKED ISSUE #4796: `response.follow_all()` problem with `cb_kwargs` getting shared by all selector objects.
### Description

[Description of the issue]

I have a weird issue. I am scraping the following Arabic poetry website: https://diwany.org/. I scraped meters 'called bahr in Arabic poetry' from their links at the bottom of the main website page. The `xpath` of these links is provided in the spider code snippet below.

I used `follow_all()` method with a poems list in `cb_kwargs` dictionary to collect all poems belonging to a meter as they are distributed among many pages. The issue is that this list got shared between all the meters requests. It keeps appending poems links from the previous meters' requests while it should pass an empty list on each request. I am sure this is not the intended behavior, Is it ?!

On the other hand, this behavior does not happen when I use the `follow()` method with a `for` loop!

### Steps to Reproduce

Here is my spider code for the website. I also put comments on the important parts.

```
import scrapy

class PoemsSpider(scrapy.Spider):
    name = 'poems'
    allowed_domains = ['diwany.org']
    start_urls = ['http://diwany.org/']

    def parse(self, response):
        bahrs_links = response.xpath("//div[@class='menu-behers-ar-container']//a")
        # this is the floawed method. poems links keeps appending from the previous requests. I got that from the print statement in the callback
        # yield from response.follow_all(
        #    bahrs_links, callback=self.parse_bahr, cb_kwargs=dict(poems_links=list())
        # )
        # This code behaves as expected
        for link in bahrs_links:
            yield response.follow(link, callback=self.parse_bahr, cb_kwargs=dict(poems_links=list()))

    # bahr means meter
    def parse_bahr(self, response, poems_links):
        # a debugging statement 
        # it gives the previous poems links from the previous requests even if this is a new request when using follow_all(). but it prints an empty list when using follow()
        print('the initial number of poems links for the url:',response.url,'is',len(poems_links))
        poems_links += response.xpath("//div[@class='post hentry ivycat-post']/h2//a")
        # this is to collect poems links for the current meter as the website uses pagination
        next_page = response.css(".pip-nav-next a")
        if next_page:
            yield response.follow(
                next_page[0],
                callback=self.parse_bahr,
                cb_kwargs=dict(poems_links=poems_links),
            )
        else:
            yield from response.follow_all(
                poems_links, callback=self.parse_poem,
            )

    def parse_poem(self, response):
        '''rest of the code'''
        '''yield item'''
        return None

```

The poems list in `cb_kwargs` should be independent for each selector object!

The poems list appends links from the previous calls.

this is consistent behavior. Although I knew that if I used the duplicate filter, these links would be filtered out, but this is not the intended behavior especially when I am building some logic on this assumption! 

### Versions

Please paste here the output of executing `scrapy version --verbose` in the command line.

Scrapy       : 2.3.0
lxml         : 4.5.2.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 20.3.0
Python       : 3.8.2 (default, Jul 16 2020, 14:00:26) - [GCC 9.3.0]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1g  21 Apr 2020)
cryptography : 3.0
Platform     : Linux-5.4.0-47-generic-x86_64-with-glibc2.29

### Additional context

I went through the code but did not found anything suspicious. I am not sure if this problem is because of the generator's behaviors?

burakcank:
I added a caution paragraph addressing the issue #4796.

I would add an example but I honestly don't believe this is related with scrapy. Like @Gallaecio said this is about passing a mutable object to any function call in python. So kept it short as a heads-up.

Fixes #4796.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -851,6 +851,9 @@ Response objects
 
     .. automethod:: Response.follow_all
 
+    .. caution:: Be careful when passing a mutable object (e.g. list) to :attr:`cb_kwargs`
+        while using ``follow_all``, as this might lead to mutation of the object in the
+        subsequent request generations.
 
 .. _topics-request-response-ref-response-subclasses:
 


wRAR (approved):

Gallaecio:
I believe `meta` is also a problematic parameter here, and should be mentioned as well.

In fact, I think it’s more dangerous than `cb_kwargs` here, to the point where I’m not sure its used should be allowed when using `follow_all`. While the problem with `cb_kwargs` is usually a value withing the dictionary being mutable (e.g. `cb_kwargs['foo']`), with `meta` the problem can (will?) also be `meta` itself. I believe even built-in middlewares (e.g. the retry middleware) just edit the non-mutable `meta` keys relevant to them, and if the same `meta` object is shared by multiple requests, I think that is likely to cause issues (e.g. the retry counter is shared by all requests).

I’m not 100% sure about that, so I want to test it when I have the time. And I also wonder if it’s technically true for `cb_kwargs` as well, even if more unlikely. If the test confirms my suspicions, we should discuss what to do about `meta` (and `cb_kwargs`?). Do we disallow the argument altogether? Make a shallow copy and keep the cautionary message about mutable values being used? Make a deep copy? (probably not, could cause more issues than it would solve) Add a parameter to choose the copying method?

AdrianAtZyte:
#7962

PR #5150, opened 2021-05-16
changing dunder-str to dunder-repr

gustavo-bordin:
### **What:**
Changing all dunder-str to dunder-repr
### **Why:**
Some **\_\_repr__** were being set as **\_\_repr__** = **\_\_str__**, we actually dont need this because if the compiler does not find \_\_str__, then it will get **\_\_repr__** as an alternative. 

### **Proof**:
https://onecompiler.com/python/3wxqqch57

### DIFF
--- a/scrapy/core/http2/stream.py
+++ b/scrapy/core/http2/stream.py
@@ -150,11 +150,9 @@ class Stream:
                 self.close(StreamCloseReason.CANCELLED)
 
         self._deferred_response = Deferred(_cancel)
-
-    def __str__(self) -> str:
+    
+    def __repr__(self):
         return f'Stream(id={self.stream_id!r})'
-
-    __repr__ = __str__
 
     @property
     def _log_warnsize(self) -> bool:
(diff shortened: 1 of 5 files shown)


Gallaecio (approved):

scrapy/core/http2/stream.py
```diff
@@ -151,11 +151,9 @@ def _cancel(_) -> None:
 
         self._deferred_response = Deferred(_cancel)
 
-    def __str__(self) -> str:
+    def __repr__(self):
```
wRAR, line 154: Should the typing be added here too?

scrapy/http/request/__init__.py
```diff
@@ -121,11 +121,9 @@ def _set_body(self, body: Optional[Union[str, bytes]]) -> None:
     def encoding(self) -> str:
         return self._encoding
 
-    def __str__(self) -> str:
+    def __repr__(self):
```
wRAR, line 124: Should the typing be added here too?

wRAR (approved):

PR #5158, opened 2021-05-23
completing the urllength tests

luisotaviocap98:
This is my attempt to contribute. Analyzing the tests in the [codecov](https://codecov.io/gh/scrapy/scrapy/src/master/scrapy/spidermiddlewares/urllength.py), I decided to complete the test coverage in the [Urllength](https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/urllength.py) module, in an attempt to make the code 100% covered. For this, I created the function `test_setting`, which tests the `from_setting` function of Urllength, checking the case in which it is successful and the case of failure.

### DIFF
--- a/tests/test_spidermiddleware_urllength.py
+++ b/tests/test_spidermiddleware_urllength.py
@@ -7,6 +7,7 @@ from scrapy.http import Response, Request
 from scrapy.spiders import Spider
 from scrapy.utils.test import get_crawler
 from scrapy.settings import Settings
+from scrapy.exceptions import NotConfigured
 
 
 class TestUrlLengthMiddleware(TestCase):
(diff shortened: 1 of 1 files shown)


tests/test_spidermiddleware_urllength.py
```diff
@@ -39,3 +40,14 @@ def test_logging(self):
...
+    def test_setting(self):
+        length = 10
+        settings = Settings({'URLLENGTH_LIMIT': length})
+        test_ok = UrlLengthMiddleware.from_settings(settings)
+        settings_error = Settings({'URLLENGTH_LIMIT': 0})
+        with self.assertRaises(NotConfigured) as cm:
+            UrlLengthMiddleware.from_settings(settings_error)
+        the_exception = cm.exception
+        self.assertEquals(type(the_exception), type(NotConfigured()))
+        self.assertEquals(test_ok.maxlength, length)
```
Gallaecio, line 53: Could you split every check that makes sense on its own into a separate test method?

luisotaviocap98: ok, it's on [this commit](https://github.com/scrapy/scrapy/pull/5158/commits/a4b768d419680c2125d9b567515f214585ae050b)

tests/test_spidermiddleware_urllength.py
```diff
@@ -45,9 +45,11 @@ def test_setting(self):
...
+
+    def test_setting_error(self):
         settings_error = Settings({'URLLENGTH_LIMIT': 0})
         with self.assertRaises(NotConfigured) as cm:
             UrlLengthMiddleware.from_settings(settings_error)
         the_exception = cm.exception
-        self.assertEquals(type(the_exception), type(NotConfigured()))
-        self.assertEquals(test_ok.maxlength, length)
+        self.assertEquals(type(the_exception), type(NotConfigured()))
```
Gallaecio: `assertRaises` already checks the exception type, these lines do not seem necessary.

wRAR:
Note that the first function doesn't seem to be needed and the second function should be changed to use `from_crawler()`.

wRAR:
Done/no longer applicable.

PR #5160, opened 2021-05-24
Fix tests that expect a specific warning

whalebot-helmsman:
Two fixes a here:
- solution for https://github.com/pypa/pip/issues/9215 in our python3.8 environment. https://github.com/scrapy/scrapy/pull/5146/checks?check_run_id=2570524886 python3.8 runs are killed by 6 hours timeout now.
- don't rely on warnings order in tests

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -18,6 +18,7 @@ jobs:
         - python-version: 3.8
           env:
             TOXENV: pylint
+            TOX_PIP_VERSION: 20.3.3
         - python-version: 3.9
           env:
             TOXENV: typing
@@ -36,5 +37,8 @@ jobs:
     - name: Run check
       env: ${{ matrix.env }}
       run: |
+        if [[ ! -z "$TOX_PIP_VERSION" ]]; then
+            pip install tox-pip-version
+        fi
         pip install -U tox
         tox
--- a/.github/workflows/tests-ubuntu.yml
+++ b/.github/workflows/tests-ubuntu.yml
@@ -39,6 +39,7 @@ jobs:
         - python-version: 3.8
           env:
             TOXENV: extra-deps
+            TOX_PIP_VERSION: 20.3.3
         - python-version: 3.9
           env:
             TOXENV: asyncio
@@ -66,6 +67,9 @@ jobs:
           tar -jxf ${PYPY_VERSION}.tar.bz2
           $PYPY_VERSION/bin/pypy3 -m venv "$HOME/virtualenvs/$PYPY_VERSION"
           source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+        fi
+        if [[ ! -z "$TOX_PIP_VERSION" ]]; then
+            pip install tox-pip-version
         fi
         pip install -U tox
         tox
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -477,7 +477,8 @@ class EngineTest(unittest.TestCase):
             assert _find_in_warning_list(
                 warning_list,
                 ScrapyDeprecationWarning,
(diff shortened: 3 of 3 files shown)


Gallaecio (approved):
I wonder if the first issue could be solved by updating https://github.com/scrapy/scrapy/blob/master/tests/upper-constraints.txt with additional contraints, or updating the existing constraints. But it could break again any time in the future. So maybe it’s OK to use an old pip where the issue occurs.

About the second change, I think the current tests are correct in that they make sure that the only exceptions that happen are the expected exceptions. It should not happen that unrelated exceptions are caught. But unexpected exceptions have been leaking from unrelated tests for a while in Scrapy, so maybe this change is OK as well.

+0.5 from be for both changes. Even if they do not address the root issues, they make tests less fragile. And fragile tests are a waste of time.

whalebot-helmsman:
> I wonder if the first issue could be solved by updating https://github.com/scrapy/scrapy/blob/master/tests/upper-constraints.txt with additional contraints, or updating the existing constraints. But it could break again any time in the future. So maybe it’s OK to use an old pip where the issue occurs.

tox creates environments with latest pip available for python (3.8 in our case), so using constraints doesn't help here

whalebot-helmsman:
> About the second change, I think the current tests are correct in that they make sure that the only exceptions that happen are the expected exceptions. 

The problem is there is a race condition in https://github.com/scrapy/scrapy/blob/ee682af3b06d48815dbdaa27c1177b94aaf679e1/scrapy/extensions/telnet.py#L72 . Without return it leads to `ResourceLeakedWarning`. With return it creates closing socket after engine stop

kmike:
> The problem is there is a race condition

Sorry @whalebot-helmsman, could you please elaborate? Arre you saying that there is a bug in telnet extension, and that's the reason for extra warnings? Would it make sense to fix the bug instead of fixing the tests? I think I'm missing something obvious :)

whalebot-helmsman:
This race condition somewhere in `@inlineCallbacks`-`send_deffered_signal`-`engine.close`complex, extension is only a collateral damage. It should be fixable somehow.

pawelmhm:
@whalebot-helmsman  about fixing tests. Is it supposed to fix all possible tests failures? Because here in my PR I see some failures about mitmproxy installation and some about tox. https://github.com/scrapy/scrapy/issues/5166 is your PR fixing "no matching distribution for tox" errors? https://github.com/scrapy/scrapy/pull/5175/checks?check_run_id=2772942979

We had some talk recently with @Gallaecio about using tox in github actions. This test failure seems to indicate that tox can be a burden. I don't really understand why it fails now, do you know @whalebot-helmsman ? I test with Python 3.9.5 locally and I'm able to install tox locally, so why it says no matching version in github build?

wRAR:
This changes the checks in test_engine because of a problem apparently specific to that test suite (though I'm not sure we can't get the same race outside it) but the problem is more generic as we check warnings in other test cases too.

Should we change this for all cases to search for a specific warning in the list? Or should we decide that we actually want to check that only the specific warnings are raised (though in this case we shouldn't only check the first one in the list but also the list size) and filter unwanted but expected ones with e.g. a global warning filter?

Gallaecio:
I suggest we implement a function in the tests that is meant to (eventually) check for a perfect match, and have all warning checks use that.

However, we make the first implementation of the function ignore specific warnings that we know happen due to race conditions. Or, if we cannot reliable determine which warnings are those, make the function ignore any unexpected warning.

If we ever work on solving the issue we can modify just that 1 function to be more strict.

It has not always been my position, but at the moment I would rather have unexpected warnings in production be reported to us by users instead of found by CI, than have CI fail intermittently.

kmike:
The warnings in test_engine are addressed in https://github.com/scrapy/scrapy/pull/5561.

wRAR:
As both the pip issue and the test_engine deprecation warning issue have been fixed this is no longer needed.

PR #5162, opened 2021-05-26
Reducing amount of warnings during test run

whalebot-helmsman:


pytest.ini
```diff
@@ -20,20 +20,5 @@ addopts =
...
-    scrapy/http/__init__.py F401
-    scrapy/linkextractors/__init__.py E402 F401
-    scrapy/selector/__init__.py F401
-    scrapy/spiders/__init__.py E402 F401
-
-    # Issues pending a review:
-    scrapy/utils/url.py F403 F405
-    tests/test_loader.py E741
+filterwarnings=
+    ignore::DeprecationWarning:twisted.web.test.test_webclient
```
Gallaecio, line 24: I think we should not ignore this. We need to fix it instead.

Maybe we can create an issue to deal with this, and add a comment here with a reference to that issue, so that it’s clear that the end goal is to fix the issue, but hide the warning since we are already aware of it and tracking it.

scrapy/core/engine.py
```diff
@@ -269,7 +269,7 @@ def _downloaded(
     ) -> Union[Deferred, Response]:
         assert self.slot is not None  # typing
         self.slot.remove_request(request)
-        return self.download(result, spider) if isinstance(result, Request) else result
+        return self.download(result) if isinstance(result, Request) else result
```
Gallaecio, line 272: If I recall right, @elacuesta was working on this, and the reason that argument remains there is for backward compatibility (_downloaded is private, but callers of _downloaded that are not private pass the `spider` argumetn and that argument should be passed).

I do understand that self.download must be logging a warning when called with that parameter. Maybe we could make it so that calls that come from _downloaded do not cause the warning.

pytest.ini
whalebot-helmsman: File contains two type of entities:
- `SmthResource` which we use
- Deprecated twisted classes with deprecation tests
Importing just `SmthResources` emits warnings. So it is okay to ignore them.

Gallaecio: I now realize that this will probably only affect tests, since I don’t imagine we use `twisted.web.test` outside tests. Assuming so, I no longer things this needs to be fixed or cannot be hidden.

Gallaecio (approved):

kmike:
Nice, thanks @whalebot-helmsman!

PR #5170, opened 2021-05-31
Documentation update

halekyl:
License statement added. Remove page that was no longed used.


LICENSE
```diff
@@ -1,3 +1,4 @@
+BSD 3-clause License
```
Gallaecio, line 1: BSD 3-clause licenses are not usually titles as such. See https://opensource.org/licenses/BSD-3-Clause

NEWS
```diff
@@ -1 +0,0 @@
-See docs/news.rst
```
Gallaecio, line 1: I’m not against this change, but since the file was added [on purpose](https://github.com/scrapy/scrapy/commit/f5b87dbef8f3c657e31eeb026d8a77b0ccc77c29) and nothing has changed since then, I’m not sure why we should remove it now.

README.rst
```diff
@@ -68,6 +68,11 @@ The quick way::
 See the install section in the documentation at
 https://docs.scrapy.org/en/latest/intro/install.html for more details.
 
+License
+=============
+
+BSD 3-clause License 
+
```
Gallaecio, line 75: I think it would be simpler to mention this in the first paragraph:

> Scrapy is a fast, high-level, BSD-licensed web crawling and web scraping framework

That should be enough for readers familiar with open source licenses to tell that it’s a permissive license. Anyone interested in knowing the actual flavor of BSD license can read the license and count the number of clauses in it :smile: .

PR #5171, opened 2021-06-01
Type hints for Request and subclasses

elacuesta:


Urahara:
@elacuesta  Sorry for noobish why some times you just use `Request` and others you use this `TypeVar` that references to `Request`?

elacuesta:
That approach is based on https://stackoverflow.com/a/44644576 (https://github.com/scrapy/scrapy/pull/4243#discussion_r364441716), the idea being that `cls` argument for class methods is the same as the return type. Thanks for pointing that out, it made me re re-check a couple of mistakes I had.

wRAR (approved):

Issue #5166, opened 2021-05-28
HTTP cache should be for project but spider


internalG:
After setting HTTPCACHE_ENABLED = True I find that the file storage is separated by spider name. So the web page is still reCrawled in another spider. That makes this feature a dup of duplicate url filter. I think the cache should be valid for all spiders in project.

pawelmhm:
Good catch. Indeed looking here: https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/httpcache.py#L308 we use spider name in cache path in both DBM cache and FileSystemCache. Seems like it is like this from the beginning, when cache was added 10 years ago it was per spider. https://github.com/scrapy/scrapy/commit/3954e600ca39fd8b02e881758b1bbc0029279d1e

I wonder what is the rationale for this. It seems to be based on assumption that you have one spider per domain, which is not always the case. It would be probably better to store per domain, not per spider.

pawelmhm:
The only potential problem I see is that in large project using DBM storage will get one large file instead of several smaller files, but that should not cause significant issues. And potentially it could slow down things for people who switch from older Scrapy versions. When running with cache after this change their client won't find the old path, and it will have to redownload all things to new locations.

internalG:
I had reused FilesystemCacheStorage code to create my global file cache. Just return globalcache instead of spider.name for the folder name. It works well. So if we need the compatibility for older version we can add a new option like HTTPCACHE_GLOBAL_STORAGE = True for the new behavior and keep default to False.

The cache is very useful but underrated in scrapy and documentation.

PR #5175, opened 2021-06-08
[extensions/httpcache] dont store responses per spider

pawelmhm:
spider per netloc in filesystem cache and just in one global DBM
database in case of DBMCache.


pawelmhm:
fixes #5166

Gallaecio:
Making this backward-compatible will be tricky.

pawelmhm:
Which backward compatibility issues you mean? 

It will force refresh cache for everyone, so that's the problem, so need to inform people about this. And if they want to keep it should we add some setting for it?

PR #5176, opened 2021-06-09
Update LICENSE

HoosierTexan:
Added license name to LICENSE file.


Gallaecio:
That’s not supposed to be in the actual license text, see https://opensource.org/licenses/BSD-3-Clause

It’s just a common name to refer to it, an alias.

PR #5178, opened 2021-06-11
Feeds: Item Filters

### LINKED ISSUE #5161: Feeds Enhancement: Item Filters
## Summary

Currently there are no convenient ways to filter items before they can be exported. An ```ItemChecker``` class can be used to filter items while also providing flexibility to the user. 

## Motivation/Proposal

Scrapy currently doesn't have any convenient APIs to customize conditions for item exports. An ```ItemChecker``` class can be used by the user to define constraints for acceptable items for particular feeds.

The ```ItemChecker``` class can have 3 main public methods ```accepts```, ```accepts_class``` and ```accepts_fields```. Scrapy will mainly use ```accepts``` method to decide if an item is acceptable, ```accepts_class``` and ```accepts_fields``` will have certain default behaviors which can be overriden by the user should they want to customize them.
```python
class ItemChecker:
   """
   This will be used by FeedExporter to decide if an item should be allowed
   to be exported to a particular feed.
   :param feed_options: FEEDS dictionary passed from FeedExporter
   :type feed_options: dict
   """
   accepted_items = []    # list of Items user wants to accept

    def __init__(self, feed_options):
        # populate accepted_items with item_classes values from feed_options if present

    def accepts(self, item):
        """
        Main method to be used by FeedExporter to check if the item is acceptable according
        to defined constraints. This method uses accepts_class and accept_fields method
        to decide if the item is acceptable.
        :param item: scraped item which user wants to check if is acceptable
        :type item: scrapy supported items (dictionaries, Item objects, dataclass objects, and attrs objects)
        :return: `True` if accepted, `False` otherwise
        :rtype: bool
        """

    def accepts_class(self, item):
        """
        Method to check if the item is an instance of a class declared in accepted_items
        list. Can be overriden by user if they want allow certain item classes.
        Default behaviour: if accepted_items is empty then all items will be
        accepted else only items present in accepted_items will be accepted.
        :param item: scraped item
        :type item: scrapy supported items  (dictionaries, Item objects, dataclass objects, and attrs objects)
        :return: `True` if item in accepted_items, `False` otherwise
        :rtype: bool
        """

    def accepts_fields(self, fields):
        """
        Method to check if certain fields of the item passes the filtering
        criteria. Users can override this method to add their own custom
        filters.
        Default behaviour: accepts all fields.
        :param fields: all the fields of the scraped item
        :type fields: dict
        :return: `True` if all the fields passes the filtering criteria, else `False`
        :rtype: bool
        """
```
Such custom filters can be declared in ```settings.py```. For convenience Items can also be declared here without needing to create a custom ```ItemChecker``` class.
```python
from myproject.filterfile import MyFilter1
from myproject.items import MyItem1, MyItem2

FEEDS = {
    'items1.json': {
        'format': 'json',
        'item_filter': MyFilter1,
    },
    'items2.xml': {
        'format': 'xml',
        'item_classes': (MyItem1, MyItem2),
    },
}
```

This feature builds upon #4576. 

## Additional context

drs-11:
Scrapy currently doesn't have any convenient APIs to customize conditions for item exports. An ```ItemChecker``` class can be used by the user to define constraints for acceptable items for particular feeds.

Resolves #5161, resolves #4576


scrapy/itemchecker.py
```diff
@@ -0,0 +1,63 @@
...
+        Main method to be used by FeedExporter to check if the item is acceptable according
+        to defined constraints. This method uses accepts_class and accept_fields method
+        to decide if the item is acceptable.
+        :param item: scraped item which user wants to check if is acceptable
+        :type item: scrapy supported items (dictionaries, Item objects, dataclass objects, and attrs objects)
+        :return: `True` if accepted, `False` otherwise
+        :rtype: bool
+        """
+        adapter = ItemAdapter(item)
+        return self.accepts_item(item) and self.accepts_fields(adapter.asdict())
```
drs-11, line 34: it's supposed to be ```self.accepts_class``` instead of ```self.accepts_item```, I'll change that

scrapy/itemchecker.py
```diff
@@ -0,0 +1,63 @@
...
+    :type feed_options: dict
+    """
+
+    item_classes = ()
+
+    def __init__(self, feed_options):
+        self.feed_options = feed_options
+
+        if 'item_classes' in self.feed_options:
+            self.item_classes += self.feed_options['item_classes']
```
Gallaecio, line 21: This will fail, since `self.item_classes` is a (non-mutable) tuple.

It may make sense to make `self.item_classes` a set instead of a tuple.

I also wonder if we should support item classes to be import paths instead of actual classes, in case the FEEDS setting has been defined as a JSON string.

drs-11: That's a good idea I'll include both the cases.

drs-11: My reasoning for using ```+=```: I was planning on having item_classes declared in a tuple in settings.py so I thought I would update the class's self.item_classes tuple with the operation ```+=```.

scrapy/itemchecker.py
```diff
@@ -0,0 +1,75 @@
...
+
+    item_classes = ()
+
+    def __init__(self, feed_options):
+        self.feed_options = feed_options
+
+        if 'item_classes' in self.feed_options:
+            item_classes = self.feed_options['item_classes']
+
+            if isinstance(item_classes, (Item, str)):
```
Gallaecio, line 26: I would suggest requiring the value to be a list. It will not be much of a trouble for users, it will keep their code consistent (`'item_classes': class` seems semantically wrong, `'item_classes': [class]` seems more consistent with the option name), and it will simplify your implementation (you don’t need an `if` statement here at all).

Internally, because you are merging the user-input option with a class-defined variable, you may want to enforce the class-defined variable to be a set, and convert the user-input option into a set before you merge them. Or you could remove the class-defined variable, and only filter item classes defined by users.

scrapy/itemchecker.py
```diff
@@ -0,0 +1,75 @@
...
+        Main method to be used by FeedExporter to check if the item is acceptable according
+        to defined constraints. This method uses accepts_class and accept_fields method
+        to decide if the item is acceptable.
+        :param item: scraped item which user wants to check if is acceptable
+        :type item: scrapy supported items (dictionaries, Item objects, dataclass objects, and attrs objects)
+        :return: `True` if accepted, `False` otherwise
+        :rtype: bool
+        """
+        adapter = ItemAdapter(item)
+        return self.accepts_class(item) and self.accepts_fields(adapter.asdict())
```
Gallaecio, line 46: I think here we are introducing a potential, unnecessary performance degradation: we are converting every item into a dict even if `self.accepts_fields` returns `True` for every item. That conversion into a dict may be non-trivial, specially since Scrapy supports arbitrary objects as items by extending https://github.com/scrapy/itemadapter.

Moreover, `self.accept_fields` may want to check field metadata to determine whether or not to accept a field, for which it would need either the original item or the whole adapter.

The only way that I see to solve this performance issue without making `self.accepts_fields` pointless (same signature as `self.accepts_class`) is to actually remove the method.

The more I think about it, the more I see that it’s best to keep this default class simple, to only have the `accepts` method and let it do what `accepts_class` would do. But if you still want to keep the function for some reason, please avoid calling `asdict()` here.

Gallaecio (approved):
Looks good to me, great job!

Gallaecio:
Sorry, I went ahead of myself. We still need documentation changes (see #4576).

docs/topics/feed-exports.rst
```diff
@@ -347,6 +370,19 @@ as a fallback value if that key is not provided for a specific feed definition:
 
 -   ``fields``: falls back to :setting:`FEED_EXPORT_FIELDS`.
 
+-   ``item_classes``: list of acceptable :ref:`item classes <topics-items>` that should be exported to the feed storage.
+
+    If no items declared then there will be no filtering based on item_classes. There is no fallback value.
+
+    .. versionadded:: 2.X.X
```
drs-11, line 377: I am not sure which version to write here or if this was needed at all.

drs-11:
Hmmm not sure how code coverage dropped just by changing one documentation file.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,26 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+For instance::
+
+    class MyCustomFilter(scrapy.itemchecker.ItemChecker):
+
+        def accepts(self, item):
+            if "field1" in item and item["field1"] == "expected_data":
+                return True
+            return False
+
```
Gallaecio, line 290: I would call the section “Item filtering” instead, so that it can be about item filtering in general, and not specifically about the item filter class.

Then you could start the section by explaining that Scrapy allows to use the `item_classes` option in `FEEDS` to indicate which types of items to deliver to each feed. Then you could continue by explaining that this feature is implemented by Scrapy’s `ItemChecker` class, which is the default value of the `item_filter` option in `FEEDS`. Then  you could go on to explain how to create custom item filtering classes.

I also think that it may be best to rename `scrapy.itemchecker.ItemChecker` to `scrapy.itemfilter.ItemFilter`, for consistency. In fact, it may be best to move it to `scrapy.extensions.feedexport.ItemFilter`, for consistency with other classes of the extension.

Finally, it would be great to include here at the end the reference documentation of the `ItemChecker` class, with `.. autoclass: scrapy.itemchecker.ItemChecker`. You can then use ``` :class:`~scrapy.itemchecker.ItemChecker` ``` instead of ``` `ItemChecker` ``` when referring to the class from other parts of the documentation.

docs/topics/feed-exports.rst
Gallaecio: See https://docs.scrapy.org/en/latest/contributing.html#documentation-policies

docs/topics/feed-exports.rst
```diff
@@ -347,6 +370,19 @@ as a fallback value if that key is not provided for a specific feed definition:
...
 -   ``fields``: falls back to :setting:`FEED_EXPORT_FIELDS`.
 
+-   ``item_classes``: list of acceptable :ref:`item classes <topics-items>` that should be exported to the feed storage.
+
+    If no items declared then there will be no filtering based on item_classes. There is no fallback value.
+
+    .. versionadded:: 2.X.X
+
+-   ``item_filter``: a :ref:`filter class <item-filter>` based on ItemChecker class to filter items before they are
+    exported to the feed storage.
```
Gallaecio, line 380:
```suggestion
-   ``item_filter``: a :ref:`filter class <item-filter>` to filter items before they are
    exported to the feed storage.
```

The classes do not need to subclass `ItemChecker`, which was a possible interpretation here. This way we keep it shorter, and users can read the linked section for details on how to implement such a class.

docs/topics/feed-exports.rst
drs-11: So as this is a new feature, we should include it in 2.6.0 version release?

tests/test_feedexport.py
```diff
@@ -929,6 +929,114 @@ class MyItem2(scrapy.Item):
         yield self.assertExported(items, header, rows,
                                   settings=settings, ordered=True)
 
+    @defer.inlineCallbacks
+    def test_export_based_on_item_classes(self):
+        items = [
+            self.MyItem({'foo': 'bar1', 'egg': 'spam1'}),
+            self.MyItem2({'hello': 'world2', 'foo': 'bar2'}),
+            {'hello': 'world3', 'egg': 'spam3'},
```
adityaa30, line 937: In case you plan to use these inputs for more tests, can we declare these as constants at the top?

drs-11: I'll keep this in mind next time I add tests.

docs/topics/feed-exports.rst
Gallaecio: My point is that you can use `VERSION` as version. Whenever a release includes this change, be it 2.6 or a later version, we will replace `VERSION` with the actual version.

scrapy/extensions/feedexport.py
```diff
@@ -46,6 +46,39 @@ def build_storage(builder, uri, *args, feed_options=None, preargs=(), **kwargs):
...
+        self.item_classes = set()
+
+        if 'item_classes' in self.feed_options:
+            for item_class in self.feed_options['item_classes']:
+                self.item_classes.add(load_object(item_class))
+
+    def accepts(self, item):
+        """
+        Main method to be used by FeedExporter to check if the item is acceptable according
+        to defined constraints.
```
Gallaecio, line 69: The first paragraph of a method docstring should summarize input, operation and output, and use the imperative. What about:

```suggestion
        Return ``True`` if `item` should be exported or ``False`` otherwise.
```

Note: ``` `item` ``` is not a typo, input parameters in Python documentation are usually formatted that way in docstrings.

We should not need to specify that it’s the main method, it should be obvious from the fact that it’s the only method.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,38 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+    class MyCustomFilter:
+
+        def __init__(self, feed_options):
+            self.feed_options = feed_options
+
+        def accepts(self, item):
+            if "field1" in item and item["field1"] == "expected_data":
+                return True
+            return False
```
Gallaecio, line 295: What about this, based on @adityaa30’s feedback during our last call? (the example is untouched, just added a paragraph below)

```suggestion
For instance::

    class MyCustomFilter:

        def __init__(self, feed_options):
            self.feed_options = feed_options

        def accepts(self, item):
            if "field1" in item and item["field1"] == "expected_data":
                return True
            return False
            
You can assign your custom filtering class to the ``item_filter`` :ref:`option 
of  a feed <feed-options>`. See :setting:`FEEDS` for examples.
```

Gallaecio (approved):

adityaa30 (approved):

scrapy/extensions/feedexport.py
```diff
@@ -46,6 +46,38 @@ def build_storage(builder, uri, *args, feed_options=None, preargs=(), **kwargs):
...
+        if 'item_classes' in self.feed_options:
+            for item_class in self.feed_options['item_classes']:
+                self.item_classes.add(load_object(item_class))
+
+    def accepts(self, item):
+        """
+        Return ``True`` if `item` should be exported or ``False`` otherwise.
+
+        :param item: scraped item which user wants to check if is acceptable
+        :type item: scrapy supported items (dictionaries, Item objects, dataclass objects, and attrs objects)
```
wRAR, line 71: Should we have this list there? It's going to be outdated soon.

Gallaecio: Good point, it would be better to say “Scrapy items” and link to the items documentation.

drs-11: I'll update it accordingly.

Issue #5180, opened 2021-06-15
versionadded inside a warning in CrawlSpider docs


wRAR:
https://docs.scrapy.org/en/latest/topics/spiders.html#crawlspider has a warning block that includes the "New in version" line.

Both indenting the warning block (not sure what's the best practice for the indentation here; also, XMLFeedSpider below has the same block which should have the same indentation I think) and moving `versionadded` before it (I think I don't like it) will work but I don't know which is the correct way to position these blocks.

rudanov:
Hey! I am new here and I would like to take care of this issue.
I think the most correct way to do "New in version" block is how it's done here: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response

A block in the beginning of the class description and additional block in the description of the attribute.
I also suggest changing marking of the attributes in CrawlingRules similar to Response docs:
``link_extractor`` =>     .. attribute:: Rule.link_extractor

PR #5181, opened 2021-06-15
CrawlingRule versionadded moved out of warning block

rudanov:
#5180


wRAR:
The docs build failed with 

> `invalid signature for autoclass ('Rule\n.. versionadded:: 2.0\n  The *errback* parameter.')`

PR #5190, opened 2021-06-23
Feeds: Post-Processing

### LINKED ISSUE #2174: gzip-compressed item exports

### LINKED ISSUE #5168: Feeds Enhancement: Post-Processing
## Summary

A Feed post-processing enhancement will enable plugins such as compression, minifying, beautifier, etc. which can then be added to the Feed exporting workflow.

## Motivation/Proposal

A post-processing feature can help add more extensions dedicated to before-export-processing to scrapy. To help achieve extensibility, a ```PostProcessingManager``` can be used which will use "plugin" like scrapy components to process the data before writing it to target files.

The ```PostProcessingManager``` can act like a wrapper to the slot's storage so whenever a write event takes place, the data is run through the plugins in a pipeline-ish way to be processed and then written to the target file.

A number of plugins can be created but there will be a need to specify the order in which these plugins are used as some won't be able to process the data after it has been processed by another(eg: minifying won't work on a compression processed file). These plugins will be required to have a certain Interface so that the ```PostProcessingManager``` can use it without breaking down from unidentified components.

Few built-in plugins can be made such as for compressions: gzip, lzma, bz2.

PostProcessingManager class prototype:
```python
class PostProcessingManager:
    """
    This will manage and use declared plugins to process data in a
    pipeline.
    :param plugins: all the declared plugins for the uri
    :type plugins: list
    :param file: target file whose data will be processed before write
    :type file: file like object
    """

    def __init__(self, plugins, file):
        # 1) load the plugins here
        # 2) save file as an attribute

    def write(self, data):
        """
        Uses all the declared plugins to process data first, then writes
        the processed data to target file. 
        :param data: data passed to be written to target file
        :type data: bytes
        :return: returns number of bytes written
        :rtype: int
        """

    def close(self):
        """
        Close the target file along with all the plugins.
        """
```
PostProcessorPlugin class inteface:
```python
class PostProcessorPlugin(Interface):
    """
    Interface for plugins that will be used by PostProcessingManager. This will
    provide necessary processing method.
    """

    def __init__(self, file, feed_options):
        """
        Initialize plugin with target file to which post-processed
        data will be written and the feed-specific options.
        """

    def write(self, data):
        """
        Exposed method which will take data passed, process it and then
        write it to target file.
        :param data: data passed to be written to target file
        :type data: bytes
        :return: returns number of bytes written
        :rtype: int
        """

    def close(self):
        """
        Closes this plugin wrapper.
        """

    @staticmethod
    def process(data):
        """
        This will process the data and return it.
        :param data: input data
        :type data: bytes
        :return: processed data
        :rtype: bytes
        """
```
GzipPlugin example:
```python
@implementer(PostProcessorPlugin)
class GzipPlugin:
    COMPRESS_LEVEL = 9

    def __init__(self. file):
        # initialise various parameters for gzipping
        self.file = gzip.GzipFile(fileobj=file, mode=file.mode,
                                  compresslevel=self.COMPRESS_LEVEL)

    def write(self, data):
        return self.file.write(data)

    def close(self):
        self.file.close()

    @staticmethod
    def process(data):
        return gzip.compress(data, compresslevel=self.COMPRESS_LEVEL)
```
settings.py example:
```python
from myproject.pluginfile import MyPlugin

FEEDS = {
    'item1.json' : {
        'format': 'json',
        'post-processing': ['gzip'],
    },
    'item2.xml' : {
        'post-processing': [MyPlugin,'xz'],    # order is important
    },
}


POST_PROC_PLUGINS_BASE = {
    'gzip': 'scrapy.utils.postprocessors.GzipPlugin',
    'xz' : 'scrapy.utils.postprocessors.LZMAPlugin',
    'bz2': 'scrapy.utils.postprocessors.Bz2Plugin',
}
```

This feature idea is actually an expansion on compression support(see #2174). Item Pipelines can be used as well for compression. But implementing this feature instead can give more options to user for post-processing while making it easier for the user to activate those post-processing components for specific feeds.

## Additional context

drs-11:
A post-processing feature can help add more extensions dedicated to before-export-processing to scrapy. To help achieve extensibility, a PostProcessingManager can be used which will use "plugin" like scrapy components to process the data before writing it to target files.

Resolves #5168, resolves #2174.


scrapy/extensions/postprocessing.py
```diff
@@ -0,0 +1,76 @@
...
+    def write(self, data):
+        """
+        Uses all the declared plugins to process data first, then writes
+        the processed data to target file.
+        :param data: data passed to be written to target file
+        :type data: bytes
+        :return: returns number of bytes written
+        :rtype: int
+        """
+        return self.head_plugin.write(data)
```
Gallaecio, line 60: What do you plan to do with the returned value, the number of bytes written?

Gallaecio (commented):
Looks good.

scrapy/extensions/postprocessing.py
drs-11: I thought I should have the return value similar to file object's ```write``` because if in some other part of feed code, the code has been written in such a way to use the returned number of bytes written, that existing code doesn't break while applying a plugin wrapper.

Gallaecio:
I did not think of it when we were discussing the API, but since these plugins can be specific to a single feed, having access to Scrapy settings thgough being a Scrapy object that can declare `from_crawler` is not enough. Plugins should also be able to access feed_options, in case plugins want to expose options and let them be defined that way.

drs-11:
> I did not think of it when we were discussing the API, but since these plugins can be specific to a single feed, having access to Scrapy settings thgough being a Scrapy object that can declare `from_crawler` is not enough. Plugins should also be able to access feed_options, in case plugins want to expose options and let them be defined that way.

Oh yeah I was going to ask you about choosing between `feed_options` and `from_crawler`. That clears it up :+1:.

scrapy/extensions/postprocessing.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def write(self, data):
+        return self.lzmafile.write(data)
+
+    def close(self):
+        self.lzmafile.close()
+        self.file.close()
+
+
+class PostProcessingManager(IOBase):
```
drs-11, line 96: `CsvItemExporter` actually wrapped the storage with `io.TextIOWrapper`, and `io.TextIOWrapper` expects to have a file like object having at least the attributes and methods of `io.IOBase`. So there was some error when it only got an object with `write` and `close` methods. Subclassing `io.IOBase` fixed this in addition to declaring this object as writable with the help of `writable` method.

scrapy/extensions/feedexport.py
```diff
@@ -350,7 +351,12 @@ def _start_new_batch(self, batch_id, uri, feed_options, spider, uri_template):
         :param uri_template: template of uri which contains %(batch_time)s or %(batch_id)d to create new uri
         """
         storage = self._get_storage(uri, feed_options)
-        file = storage.open(spider)
+
+        if "postprocessing" in feed_options:
+            file = PostProcessingManager(feed_options["postprocessing"], storage.open(spider), feed_options)
+        else:
+            file = storage.open(spider)
```
Gallaecio, line 358: - `file` definition could be simplified, and made slightly more readable.
- Seeing it in action now, I see it probably makes sense to simplify the signature of `PostProcessingManager`.

```suggestion
        file = storage.open(spider)
        if "postprocessing" in feed_options:
            file = PostProcessingManager(file, feed_options)
```

scrapy/extensions/postprocessing.py
```diff
@@ -1,7 +1,7 @@
 """
 Extension for processing data before they are exported to feeds.
 """
-
+from io import IOBase
 from bz2 import BZ2File
 from gzip import GzipFile
 from lzma import LZMAFile
```
Gallaecio:
```suggestion
from bz2 import BZ2File
from gzip import GzipFile
from io import IOBase
from lzma import LZMAFile
```

scrapy/extensions/postprocessing.py
Gallaecio: Nice. I did not realize this change only needed to affect the manager and not the plugins. This is definitely the cleanest solution, great job!

I would include, however, a comment above the class definition explaining why we inherit from `IOBase` here, for future references. Indicate that it’s so that exporters can treat the result of the post-processing as a file-like writable object, and you could mention the example of CsvItemExporter using TextIOWrapper.

We should also make sure, when updating the documentation, to also review the documentation of exporters and make sure that it is clearly stated what the API of the file-like object exporters get is: that of an open, writable, non-seekable IOBase object without a `fileno` property.

Gallaecio:
Looks good to me, time for tests and documentation?

tests/test_feedexport.py
```diff
@@ -1320,6 +1320,76 @@ def test_extend_kwargs(self):
...
+                    'postprocessing': [MyPlugin1, 'scrapy.extensions.postprocessing.GzipPlugin'],
+                },
+            }
+        }
+
+        data = yield self.exported_data(items, settings)
+
+        for fmt, decompressor in format_to_decompressor.items():
+            result = decompressor.decompress(data[fmt])
+            self.assertEqual(format_to_expected[fmt], result)
```
Gallaecio, line 1391: The test looks great overall. It should serve to make sure that all built-in compression plugins work.

They do not hurt, but I don’t think the (random) extra parameters add much to the test. For example, if you set `gzip_compresslevel` to `4` in this test, but the plugin were not processing the parameter at all, the test would still pass. It may be better to remove them.

To test plugin parameters, you need tests that you break if parameters were misinterpreted. You want to make sure that the parameters have an impact on the output, that the output is not the same with different parameters. You don’t need to test the underlying library (e.g. you don’t need to test that Python’s GZipFile makes files smaller with 9 than with 0), but you do need to test somehow that the plugin is passing the parameter (e.g. from feed options to GZipFile).

You also need to mind that some compressors may not be deterministic (e.g. [GZipFile seems to not be deterministic unless you set `mtime` to a fixed value](https://discuss.python.org/t/gzip-py-allow-deterministic-compression-without-time-stamp/8221)), so it may not always be enough to test that the output is different with a different value in a parameter. In these cases, you’ll need to either make the output deterministic or check for something more. For example, with GZipFile, you can either expose the `mtime` option and use it in tests in combination with the options that you are actually testing, or you can test an option like compression level by verifying that the higher the compression the smaller the file.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+These plugins can be activated through ``postprocessing`` :ref:`option of a feed <feed-options>`.
+Parameters to plugins can be passed through the feed settings where the key will be the parameter name
+and the value will be the parameter value. These parameters will be made accessible to the plugin
+through the ``feed_options`` parameter which will be passed to their constructor method.
+
+As the data is passed from one plugin to another, users must ensure the order of plugins do not
+cause a plugin to breakdown because of some unprocessable data.
+
+Each plugin is a Python class that must implement the following methods:
```
Gallaecio, line 290:
```suggestion
Each plugin is a class that must implement the following methods:
```

I’m not saying that no one will ever try to define a plugin in a different language. I’m saying that, when they realize why their plugin is not working, they will possibly feel ashamed and they will definitely not request that we explicitly say Python here to clarify :slightly_smiling_face:

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+.. method:: __init__(self, file, feed_options)
+
+    Constructor method to initialize plugin.
```
Gallaecio, line 294: The first paragraph of a method should always be like “Do something”. Here, “Initialize the plugin” should work.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+    :param file: target file object to which post-processed data will be written
```
Gallaecio, line 296: “file object” can be too vague. Users need to know that they can and cannot expect from the API of the `file`.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+    :param feed_options: feed-specific options
+    :type feed_options: :class:`dict`
```
Gallaecio, line 299: I would include in the `param` line a link to the documentation of feed options.

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+
+.. method:: write(self, data)
+
+   Process and write `data` to plugin's target file.
```
Gallaecio, line 303:
```suggestion
   Process and write `data` into the plugin's target file.
```

You could also use this and skip the parameter table below altogether, since it wouldn’t add anything.

```suggestion
   Process and write `data` (:class:`bytes`) into the plugin's target file.
```

docs/topics/feed-exports.rst
```diff
@@ -268,6 +268,83 @@ in multiple files, with the specified maximum item count per file. That way, as
...
+GzipPlugin
+^^^^^^^^^^
+
+Compresses recieved data with with :py:mod:`gzip` module.
+
+Accpeted parameters:
+
+-   ``gzip_compresslevel``: integer from 0 to 9 controlling the level of compression.
+-   ``gzip_mtime``: numeric timestamp for the compressed file.
+-   ``gzip_filename``: filename to be included in the header of compressed file.
```
Gallaecio, line 326: Use `.. autoclass::` instead of a section. That way you can move this documentation to the code, so that it’s written in a single place but can be read from both code and documentation. This also causes the header to be the whole import path of the class, which allows users to know which value to set in `postprocessing`.

About parameters, you should list them all since we don’t support every GZipFile option arbitrarily, but you don’t need to document anything about them separately. You can instead refer users to the documentation of `GZipFile` for information about them. That way you not only need to write less, you give users access to more information, and if a future version of Python changes how one of those parameters behaves, we won’t need to update our documentation.

Gallaecio (commented):
Could you move the test method into its own class and split it into separate, granular methods that test different things?

Also, try to make the tests methods minimal. Try for each test to have the minimum code that would still make the test do its job.

If you want to avoid some repetitions once you have multiple tests testing different things, you can write non-test methods and share them between test methods. But code repetition in tests is not usually a bad thing; do not avoid it for the sake of avoiding it, only avoid duplication if it makes your life easier.

scrapy/extensions/postprocessing.py
```diff
@@ -0,0 +1,151 @@
...
+from bz2 import BZ2File
+from gzip import GzipFile
+from io import IOBase
+from lzma import LZMAFile
+from typing import Any, BinaryIO, Dict, List
+
+from scrapy.utils.misc import load_object
+
+
+class GzipPlugin:
```
adityaa30, line 13: Since all post-processing plugins have a common structure, how about declaring an interface (eg. `IPostProcessingPlugin`) with the `write` and `close` methods?

docs/topics/feed-exports.rst
```diff
@@ -275,76 +275,63 @@ Post-Processing
...
-can create their own plugins. 
+Scrapy provides an option to activate plugins to post-process feeds before they are exported
+to feed storages. In addition to using :ref:`builtin plugins <builtin-plugins>`, you
+can create your own :ref:`plugins <custom-plugins>`. 
 
-These plugins can be activated through ``postprocessing`` :ref:`option of a feed <feed-options>`.
-Parameters to plugins can be passed through the feed settings where the key will be the parameter name
-and the value will be the parameter value. These parameters will be made accessible to the plugin
-through the ``feed_options`` parameter which will be passed to their constructor method.
+These plugins can be activated through ``postprocessing`` option of a feed. The option must
```
Gallaecio, line 282:
```suggestion
These plugins can be activated through the ``postprocessing`` option of a feed. The option must
```

docs/topics/feed-exports.rst
```diff
@@ -275,76 +275,63 @@ Post-Processing
...
-    :type feed_options: :class:`dict`
+.. _builtin-plugins:
 
-.. method:: write(self, data)
+Built-in Plugins
+----------------
 
-   Process and write `data` to plugin's target file.
+GzipPlugin
+^^^^^^^^^^
```
Gallaecio, line 294: I think the section headers can be removed now that we are using `autoclass`.

docs/topics/feed-exports.rst
```diff
@@ -275,76 +275,63 @@ Post-Processing
...
--   ``gzip_mtime``: numeric timestamp for the compressed file.
--   ``gzip_filename``: filename to be included in the header of compressed file.
+.. method:: __init__(self, file, feed_options)
 
-LZMAPlugin
-^^^^^^^^^^
+    Initialize the plugin.
 
-Compresses recieved data with :py:mod:`lzma` module.
+    :param file: file like object having at least `write` and `close` method implemented
```
Gallaecio, line 319:
```suggestion
    :param file: file-like object having at least the `write` and `close` methods implemented
```

docs/topics/feed-exports.rst
```diff
@@ -275,76 +275,63 @@ Post-Processing
...
 
--   ``lzma_format``: specifies what container format should be used.
--   ``lzma_check``: specifies the type of integrity check to include in the compressed data.
--   ``lzma_preset``: an integer between 0 and 9. Higher presets produce smaller output, but make the compression process slower.
--   ``lzma_filter``: should be a filter chain specifier.
+.. method:: write(self, data)
 
-Bz2Plugin
-^^^^^^^^^
+   Process and write `data` (:class:`bytes`) into to plugin's target file.
```
Gallaecio, line 326:
```suggestion
   Process and write `data` (:class:`bytes`) into the plugin's target file.
```

docs/topics/feed-exports.rst
```diff
@@ -275,76 +275,63 @@ Post-Processing
...
-Compresses recieved data with :py:mod:`bz2` module.
+.. method:: close(self)
 
-Accepted parameters:
+    Close the target file object.
 
--   ``bz2_compresslevel``: integer from 1 to 9 controlling the level of compression.
+To pass a parameter to your plugin, simply enter the paramter name and value in
+the feed-options. You can then access those parameters through the feed specific
+`feed_options` dictionary which is passed to each declared plugin.
```
Gallaecio, line 334:
```suggestion
To pass parameters to your plugin use :ref:`feed options <feed-options>`. You 
can then access those parameters from the ``__init__`` method of your plugin.
```

docs/topics/feed-exports.rst
```diff
@@ -406,6 +393,10 @@ For instance::
             'postprocessing': [MyPlugin1, 'scrapy.extensions.postprocessing.GzipPlugin'],
             'gzip_compresslevel': 5,
         },
+        'items.jl.xz': {
+            'format': 'jsonlines',
+            'postprocessing': [LZMAPlugin],
+        },
```
Gallaecio: I think this one can be removed from the example: I don’t think it adds anything, the one above already features a class used directly in `postprocessing`.

scrapy/extensions/postprocessing.py
Gallaecio: While there is 1 case in similar Scrapy code where we use an interface, in most similar cases we tend to instead document the interface, so the current approach I think is the one most in line with the current code base.

In Python, interfaces are seldom used, and I don’t see much benefit for them being used here.

scrapy/extensions/postprocessing.py
```diff
@@ -22,14 +32,24 @@ def __init__(self, file: BinaryIO, feed_options: Dict[str, Any]) -> None:
                                  mtime=mtime, filename=filename)
 
     def write(self, data: bytes) -> int:
+        """Compress and write the `data` into the target file."""
         return self.gzipfile.write(data)
 
     def close(self) -> None:
+        """Close target file."""
         self.gzipfile.close()
         self.file.close()
```
Gallaecio: Since the purpose of the `write` and `close` methods of any plugin is described in the Custom Plugin section, there is no need to document them.

```suggestion
    def write(self, data: bytes) -> int:
        return self.gzipfile.write(data)

    def close(self) -> None:
        self.gzipfile.close()
        self.file.close()
```

If after removing their documentation they still show up when using `autoclass`, use autoclass Sphinx options to hide these 2 methods in the documentation.

scrapy/extensions/postprocessing.py
```diff
@@ -11,6 +11,16 @@
 
 
 class GzipPlugin:
+    """
+    Compresses received data with with :py:mod:`gzip` module.
```
Gallaecio, line 15: The plugins should be about compressing in a specific file format.

```suggestion
    Compresses received data using `gzip <https://en.wikipedia.org/wiki/Gzip>`_.
```

The use of a Python module from the standard library is an implementation detail, and should not be covered in the documentation. Even if we link to the Python documentation about parameters, we could technically reimplement the plugin in the future without relying on the standard library module, while supporting the same option names and values (hence it being OK to link to that documentation for option reference).

tests/test_feedexport.py
```diff
@@ -1320,96 +1323,271 @@ def test_extend_kwargs(self):
             data = yield self.exported_data(items, settings)
             self.assertEqual(row['expected'], data[feed_options['format']])
 
+
+class FeedPostProcessedExportsTest(FeedExportTestBase):
+    __test__ = True
+    items = [
+        {'foo': 'bar1', 'baz': ''},
+        {'egg': 'bar2', 'baz': 'quux'},
+    ]
```
Gallaecio, line 1332: I don’t _think_ multiple items and multiple fields can affect post-processing. So, to keep things minimal, maybe it would make sense to keep a single item with a single field.

tests/test_feedexport.py
```diff
@@ -1320,96 +1323,271 @@ def test_extend_kwargs(self):
...
+    def get_gzip_compressed(self, data, compresslevel):
+        data_stream = BytesIO()
+        gzipf = gzip.GzipFile(fileobj=data_stream, filename="", mtime=0, compresslevel=compresslevel, mode="wb")
+        gzipf.write(data)
+        gzipf.close()
+        data_stream.seek(0)
+        return data_stream.read()
+
+    @defer.inlineCallbacks
+    def test_gzip_plugin(self):
```
Gallaecio, line 1382: Instead of a single method per plugin, you could have several methods, each to test different aspects:
- Enabling the plugin results in gzip-compressed output
- A non-default value for a plugin option has the expected effect (1 test method per option)

I see no reason to test different output formats (i.e. csv, jsonlines, xml). Plugins should work the same with any, so to make things minimal you could use a single one.

What we should have is tests that verify that each of the Scrapy export plugins (for csv, jsonlines, etc.) work with the file-like interface of the plugin manager. In other words, tests that would have failed if the plugin manager was not inheriting from IOBase.

tests/test_feedexport.py
```diff
@@ -1321,6 +1324,420 @@ def test_extend_kwargs(self):
...
+
+    @defer.inlineCallbacks
+    def test_exports_compatibility_with_postproc(self):
+
+        format_to_expected = {
+            'csv': b'foo\r\nbar\r\n',
+            'json': b'[\n{"foo": "bar"}\n]',
+            'jsonlines': b'{"foo": "bar"}\n',
+            'xml': b'<?xml version="1.0" encoding="utf-8"?>\n<items>\n<item><foo>bar</foo></item>\n</items>',
+            'marshal': b'{\xda\x03foo\xda\x03bar0',
```
drs-11, line 1700: Even `marshal.dumps` is not giving consistent data. I'll look into this.

Gallaecio: If the problem is across Python versions, you could use `marshal.dumps({"foo": "bar"})` here instead of hardcoding the value.

Gallaecio (approved):
Looks good to me, great work!

adityaa30 (approved):

drs-11:
Hmm not sure why `test_deprecated_schedule` is failing. It seems to expect `ScrapyDeprecationWarning` but instead gets `ResourceWarning`. Could it be because, pypy tests are consuming a lot of resources? But they are happening because of my changes otherwise in other PRs they are not failing.

Gallaecio:
> Hmm not sure why `test_deprecated_schedule` is failing. It seems to expect `ScrapyDeprecationWarning` but instead gets `ResourceWarning`. Could it be because, pypy tests are consuming a lot of resources? But they are happening because of my changes otherwise in other PRs they are not failing.

From what I can tell, `ResourceWarnings` can easily happen during tests. In that test, it may be that both warnings are happening, but since you are only checking the first, which is `ResourceWarnings`, the test fails.

There is probably a way to avoid the resource warning altogether. You could try and log the warning message, and see if it makes any sense. Those warnings usually happens because of unclosed files, which can be OK in CPython thanks to reference counting but PyPy has no reference counting.

If you cannot figure out how to fix the root issue, and I strongly suggest you do _not_ spend too much time on it, you could instead change the expectations, so that instead of checking that the first warning is as expected, you just check that one of the warning is as expected.

tests/test_feedexport.py
```diff
@@ -1469,6 +1472,496 @@ def test_extend_kwargs(self):
...
+        for filename, compressed in filename_to_compressed.items():
+            result = lzma.decompress(data[filename])
+            self.assertEqual(compressed, data[filename])
+            self.assertEqual(self.expected, result)
+
+    @defer.inlineCallbacks
+    def test_lzma_plugin_filters(self):
+        import sys
+        if "PyPy" in sys.version:
+            raise unittest.SkipTest("lzma filters doesn't work in PyPy")
```
Gallaecio, line 1768: Please, include a code comment that points to the corresponding bug report or source of information.

tests/test_feedexport.py
```diff
@@ -1469,6 +1472,496 @@ def test_extend_kwargs(self):
...
+    expected = b'foo\r\nbar\r\n'
+
+    class MyPlugin1:
+        def __init__(self, file, feed_options):
+            self.file = file
+            self.feed_options = feed_options
+            self.char = self.feed_options.get('plugin1_char', b'')
+
+        def write(self, data):
+            return self.file.write(bytes(data) + self.char)
```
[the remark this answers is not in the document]
drs-11: Pypy's pickle implementation was writing data in `memoryview` instead of expected `bytes`. So I was getting operand error. I figured converting it to `bytes`.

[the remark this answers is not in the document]
Gallaecio: Then what about moving the casting to the Pickle plugin and adding a comment about this? (hopefully there’s an online bug report or note somewhere that we can link)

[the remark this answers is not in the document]
drs-11: The `write` function of the custom plugin is being used by the `pickle.dump` method in `PickleItemExporter`. Can't really move the casting there in the `dump` method. And I can't find anything about the `memoryview` problem either. Seems like a very rare case someone would use a wrapper in the `dump` method :sweat_smile:.

docs/topics/feed-exports.rst
```diff
@@ -336,6 +336,9 @@ Built-in Plugins
 
 .. autoclass:: scrapy.extensions.postprocessing.LZMAPlugin
 
+.. note::
+    ``lzma_filters`` cannot be used in pypy version 7.3.1 and older.
```
Gallaecio: It probably makes sense to move this to the docstring of the class instead.

tests/test_feedexport.py
```diff
@@ -1765,7 +1765,10 @@ def test_lzma_plugin_preset(self):
     def test_lzma_plugin_filters(self):
         import sys
         if "PyPy" in sys.version:
-            raise unittest.SkipTest("lzma filters doesn't work in PyPy")
+            pypy_version = sys.version.split('\n')[1].split()[1]
+            if pypy_version <= '7.3.1':
```
Gallaecio, line 1769: This is not a reliable way to test versions:

```python
>>> '10.0.0' <= '7.3.1'
True
```

See `sys.version_info` usage in the code base, something like that should work.

tests/test_feedexport.py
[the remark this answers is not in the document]
Gallaecio: Does it work if you use 2 separate write calls?

```python
written_count = self.file.write(data)
written_count += self.file.write(self.char)
return written_count
```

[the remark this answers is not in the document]
drs-11: > Does it work if you use 2 separate write calls?

Yup it works that way. It also works like this: `self.file.write(self.char + data)` weirdly.

> If we go this route, it might be good to indicate that the input data may be a memoryview, and support that officially.

By that you mean we should include in the documentation to warn the users that input could be a `memoryview` while using `PickleItemExporter`?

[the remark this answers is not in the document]
Gallaecio: I would not put the focus on `PickleItemExporter`, and instead assume that any exporter, built-in or user-provided, may do the same.

Since supporting a `memoryview` should be trivial, and we are going to support it in our post-processing plugins, it would make sense to indicate in the documentation of plugins that their `write` method may get either `memoryview` or `bytes`, and that they should be ready to handle any of those types as input.

tests/test_feedexport.py
```diff
@@ -1469,6 +1472,502 @@ def test_extend_kwargs(self):
...
+            self.assertEqual(compressed, data[filename])
+            self.assertEqual(self.expected, result)
+
+    @defer.inlineCallbacks
+    def test_lzma_plugin_filters(self):
+        import sys
+        if "PyPy" in sys.version:
+            pypy_version_str = sys.version.split('\n')[1].split()[1]
+            pypy_version = tuple(map(int, pypy_version_str.split(".")))
+            if pypy_version <= (7, 3, 1):
```
Gallaecio, line 1772: I _think_ you can simplify this part by using `sys.version_info` instead:

```suggestion
            if sys.version_info <= (7, 3, 1):
```

Gallaecio:
Hmm… It looks like the issue is also affecting newer versions of PyPy…

Gallaecio:
It seems the issue is not the one we linked. Please, have the test be skipped whatever the PyPy version. I’ll try and find a bug report, or submit one myself, and we can include that link in the test.

Gallaecio:
https://foss.heptapod.net/pypy/pypy/-/issues/3527

tests/test_feedexport.py
drs-11: image

drs-11: `sys.version_info` returns the python version info not the pypy version and I couldn't find a simpler way to get pypy's version info.

drs-11: Looks like it won't be needed now since we are skipping all pypy versions now.

Gallaecio: Just for the record:

```python
>>> tuple(sys.version_info)
(3, 9, 3, 'final', 0)
>>> sys.version_info == (3, 9, 3, 'final', 0)
True
>>> sys.version_info <= (10, 0, 0)
True
```

Gallaecio (approved):

wRAR (approved):

Gallaecio:
:tada:

PR #5191, opened 2021-06-23
CloseSpider can be raised on spider_idle signal handler to set the closing reason

ivanprado:
The execution of a spider might have been successful or not depending on many conditions. For example, you might have expected many items to be extracted but none were extracted. It might be useful to be able to evaluate at the end of the crawling if the results are successful, and if not, close the spider using the appropriate message (e.g. "too_few_results") instead of the regular one "finished".

The `spider_idle` signal handler looks like the perfect place to put such logic. At this stage, all the work has been done, so all the stats are ready to be queried to decide the final status. The problem I found is that is not possible to set the closing reason in this handler. 

In this PR I propose to reuse the exception `CloseSpider` to provide the closing reason. The required changes are not so big, and it seems to fit. 

Let me know what do you think, and if you agree on the functionality, what else would be required (tests, etc).

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -15,8 +15,11 @@ from twisted.python.failure import Failure
 
 from scrapy import signals
 from scrapy.core.scraper import Scraper
-from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning, \
-    CloseSpider
+from scrapy.exceptions import (
+    CloseSpider,
+    DontCloseSpider, 
+    ScrapyDeprecationWarning,
+)   
 from scrapy.http import Response, Request
 from scrapy.settings import BaseSettings
 from scrapy.spiders import Spider
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -15,7 +15,8 @@ from twisted.python.failure import Failure
 
 from scrapy import signals
 from scrapy.core.scraper import Scraper
-from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning
+from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning, \
+    CloseSpider
 from scrapy.http import Response, Request
 from scrapy.settings import BaseSettings
 from scrapy.spiders import Spider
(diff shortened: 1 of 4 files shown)


ivanprado:
Example of a case where this would be useful: https://stackoverflow.com/questions/65688851/scrapy-change-closing-reason-from-finished-to-myreason

kmike:
Hey! The feature makes total sense to me. 

1) Would it be possible to add a test to it?
2) Do you think it worth mentioning this use case somewhere in the docs?

ivanprado:
Hi @kmike . I've added a test case and also documented a possible use case in the signals documentation page. 

I have a problem with the test `Ubuntu / tests (3.6.12, asyncio-pinned)`. They are failing. I don't know why. Any clue?

pablohoffman:
This functionality makes a lot of sense to me too, nice contribution @ivanprado 👍

scrapy/core/engine.py
```diff
@@ -15,7 +15,8 @@
 
 from scrapy import signals
 from scrapy.core.scraper import Scraper
-from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning
+from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning, \
+    CloseSpider
```
Gallaecio, line 19: :lipstick:  We don’t really use \ at the end if it can be avoided.

```suggestion
from scrapy.exceptions import (
    CloseSpider,
    DontCloseSpider, 
    ScrapyDeprecationWarning,
)   
```

scrapy/core/engine.py
```diff
@@ -325,14 +326,20 @@ def _spider_idle(self) -> None:
...
+        (at least) once again. A handler can raise CloseSpider to provide a custom closing reason.
         """
         assert self.spider is not None  # typing
-        res = self.signals.send_catch_log(signals.spider_idle, spider=self.spider, dont_log=DontCloseSpider)
-        if any(isinstance(x, Failure) and isinstance(x.value, DontCloseSpider) for _, x in res):
+        expected_ex = (DontCloseSpider, CloseSpider)
+        res = self.signals.send_catch_log(signals.spider_idle, spider=self.spider, dont_log=expected_ex)
+        detected_ex = {ex: x.value
+                       for _, x in res for ex in expected_ex
+                       if isinstance(x, Failure) and isinstance(x.value, ex)}
```
Gallaecio, line 336: :lipstick: 

```suggestion
        detected_ex = {
            ex: x.value
            for _, x in res
            for ex in expected_ex
            if isinstance(x, Failure) and isinstance(x.value, ex)
        }
```

tests/test_engine.py
```diff
@@ -26,7 +26,7 @@
 
 from scrapy import signals
 from scrapy.core.engine import ExecutionEngine
-from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.exceptions import ScrapyDeprecationWarning, CloseSpider
```
Gallaecio, line 29: :lipstick: 

```suggestion
from scrapy.exceptions import CloseSpider, ScrapyDeprecationWarning
```

Gallaecio (approved):
Looks great to me!

Gallaecio:
> I have a problem with the test `Ubuntu / tests (3.6.12, asyncio-pinned)`. They are failing. I don't know why. Any clue?

That’s an issue in `master` unrelated to this pull request, do not worry about it.

ivanprado:
Thank you @Gallaecio for your review :beers:. I've accepted all your suggestions.

scrapy/core/engine.py
```diff
@@ -15,7 +15,11 @@
 
 from scrapy import signals
 from scrapy.core.scraper import Scraper
-from scrapy.exceptions import DontCloseSpider, ScrapyDeprecationWarning
+from scrapy.exceptions import (
+    CloseSpider,
+    DontCloseSpider, 
+    ScrapyDeprecationWarning,
+)
```
elacuesta, line 22:
```suggestion
    DontCloseSpider,
    ScrapyDeprecationWarning,
)
```
Remove the trailing whitespaces that are making the flake8 check fail.

wRAR (approved):

ivanprado:
Thank you everybody for the reviews, commend and for merging it :-)

PR #5198, opened 2021-07-02
Fix FileFeedStoragePreFeedOptionsTest fails in CI/CD pipeline

### LINKED ISSUE #5157: FileFeedStoragePreFeedOptionsTest fails if /tmp/foobar already exists with different ownership
Quick one - if, for instance, you're running the the scrapy test suite in some sort of CI setup where builds get performed by random different dedicated "build users", you'll end up in a situation where `/tmp/foobar` already exists, owned by a different user. At which point `FileFeedStoragePreFeedOptionsTest` will fail:

```
E       PermissionError: [Errno 13] Permission denied: '/tmp/foobar'
```

Fairly trivial to fix - you could put a random suffix on the filename.. better, generate a directory to contain it using `tempfile.mkdtemp`.. many options.

Full build failure log: https://nix-cache.s3.amazonaws.com/log/5fdywg4w7s6cyf224j42yn9619kw5gw1-python3.8-Scrapy-2.4.1.drv

### Versions

Scrapy 2.4.1, python 3.8, macos 10.15

Mannan2812:
FileFeedStoragePreFeedOptionsTest fails due to missing permissions to the file, as in CI/CD pipeline random users are created. I have used tempfile.NamedTemporaryFile to generate a random file on every test run.
Fixes #5157

### DIFF
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -1780,8 +1780,9 @@ class FileFeedStoragePreFeedOptionsTest(unittest.TestCase):
     maxDiff = None
 
     def test_init(self):
+        temp = tempfile.NamedTemporaryFile().name
         settings_dict = {
-            'FEED_URI': 'file:///tmp/foobar',
+            'FEED_URI': f'file:///{temp}',
             'FEED_STORAGES': {
                 'file': FileFeedStorageWithoutFeedOptions
             },
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -1782,7 +1782,7 @@ class FileFeedStoragePreFeedOptionsTest(unittest.TestCase):
     def test_init(self):
         with tempfile.NamedTemporaryFile() as temp:
             settings_dict = {
-                'FEED_URI': f'file:///{temp.name}',
+                'FEED_URI': f'file://{temp.name}',
                 'FEED_STORAGES': {
                     'file': FileFeedStorageWithoutFeedOptions
                 },
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -1780,9 +1780,9 @@ class FileFeedStoragePreFeedOptionsTest(unittest.TestCase):
     maxDiff = None
 
     def test_init(self):
-        with tempfile.NamedTemporaryFile().name as temp:
+        with tempfile.NamedTemporaryFile() as temp:
             settings_dict = {
-                'FEED_URI': f'file:///{temp}',
+                'FEED_URI': f'file:///{temp.name}',
                 'FEED_STORAGES': {
                     'file': FileFeedStorageWithoutFeedOptions
                 },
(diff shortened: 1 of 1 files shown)


tests/test_feedexport.py
```diff
@@ -1780,8 +1780,9 @@ class FileFeedStoragePreFeedOptionsTest(unittest.TestCase):
     maxDiff = None
 
     def test_init(self):
+        temp = tempfile.NamedTemporaryFile().name
```
wRAR, line 1783: Can you please use `with` here?

Mannan2812: Added

wRAR:
Please run the code before submitting. It fails with `AttributeError: __enter__`.

Mannan2812:
> Please run the code before submitting. It fails with `AttributeError: __enter__`.

Sorry about that, fixed this exception in the latest commit

wRAR (approved):

tests/test_feedexport.py
```diff
@@ -1780,14 +1780,15 @@ class FileFeedStoragePreFeedOptionsTest(unittest.TestCase):
...
-            'FEED_URI': 'file:///tmp/foobar',
-            'FEED_STORAGES': {
-                'file': FileFeedStorageWithoutFeedOptions
-            },
-        }
-        crawler = get_crawler(settings_dict=settings_dict)
-        feed_exporter = FeedExporter.from_crawler(crawler)
+        with tempfile.NamedTemporaryFile() as temp:
+            settings_dict = {
+                'FEED_URI': f'file:///{temp.name}',
```
Gallaecio, line 1785: What is the value of `temp.name` here? Aren’t there 1 too many / in the URI scheme here? And if so, how are tests working nonetheless?

Also, according to https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile:

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

Which makes me wonder how tests are passing in Windows.

Mannan2812: @Gallaecio temp.name returns the random name of the file. We create a new file with a random name to resolve the permissions issue at each test run, and the test is passing successfully.

Gallaecio: `f'file:///{temp.name}'` generates a URI like `file:////tmp/tmp_4p8ym67`, which seems to work but has 1 too many slashes at the beginning. Replacing `f'file:///{temp.name}'` with `f'file://{temp.name}'` gives me peace of mind and still works.

As for the platform issue, I see that the issue described in the documentation affects the file while open; what we are doing here is create a random file with a file-system path, remove it (when leaving the context manager), and pass it’s path to the spider which later recreates the file. It feels a bit like a hack to me, since I am not aware of any assurances that, after the file is first removed, another call to `NamedTemporaryFile` will not create a new file on the same location, but there’s probably some randomness in the file name choosing algorithm that makes that scenario very unlikely, so :+1: from me.

Gallaecio:
```suggestion
                'FEED_URI': f'file://{temp.name}',
```

Gallaecio:
:facepalm: Looks like that extra `/` is needed in Windows.

Sorry, but it looks like the way you had it is the only way that works :bow:

Gallaecio:
Merged, thanks!

PR #5200, opened 2021-07-12
[CI] fail-fast: false

elacuesta:
Allow remaining jobs to keep running even if some of them fail (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast).

As discussed with @Gallaecio


wRAR (approved):

Gallaecio:
Great!

PR #5205, opened 2021-07-15
Feeds: Batch Triggers

### LINKED ISSUE #5169: Feeds Enhancement: Batch Delivery Triggers
## Summary

Batch creation was a recently introduced feature but was limited to only item count constraints. Batch Deliver triggers will be able to make the constraints flexible and give more control to user to create batches.

## Motivation/Proposal

Batch delivery can be simplified and made extensible by creating a class ```Batch```. It will contain information about current batch so it can be used to detect when a given contraint has been exceeded and new batch needs to be created.

```BatchPerXItems```, ```BatchPerXMins```, and ```BatchPerXBytes``` can be created as builtins with each inhereting parent class ```Batch```. A parameter value `para_val` is stored in the `Batch` class which will be used to compare against the declared `constraint`. A `para_val` for ```BatchPerXItems``` can be the number of total items currently in the batch, for ```BatchPerXMins``` it can be total mins passed since it was created. So some ```update``` calls to update the ```para_val``` may not require information from ```slot_file```.

Batch class prototype:
```python
class Batch:
    """
    Batch which will store information for current batches and provides
    suitable methods to check and update batch info.
    :param slot_uri: uri for which batch is being created
    :type slot_uri: string
    :param constraint: a constraint or limit to figure out when a new batch
        must be created.
    :param para_val: a parameter value which will be updated and be compared
        against constraint to control batch creation.
    :param batch_id: id number of the current batch.
    """

    def __init__(self, slot_uri, constraint, para_val=0):
        self.uri = slot_uri
        self.constraint = constraint
        self.para_val = para_val
        self.batch_id = 1

    def update(self, slot_file):
        """
        Updates the parameter value according to stats related to paramter
        value and contraint.
        :param slot_file: slot's file which is a good source of stats
        :type slot_file: File like object
        """

    def should_trigger(self):
        """
        Checks if para_val has crossed the constraint or not.
        :return: `True` if para_val has crossed constraint, else `False`
        :rtype: bool
        """

    def reset_para_val(self):
        """
        Resets parameter value back to its initial value and increments
        self.batch_id. Will be used after closing a batch.
        """
```
Desired Batch class can then be activated in ```settings.py``` with a constraint. To help users create complex triggers they could set the constraint to any builtin type or any arbitrary object to suit their needs. If no constraint is set, it will be pointless to load the specified Batch class. Users can add their own custom Batch class by specifying their class path.

settings.py example:
```python
from myproject.customclassfile import CustomBatch

{
    'items1.json': {
        'format': 'json',
        'batch_constraint': ('item_count', 10)   # (batch_trigger, constraint)
    },
    'items2.xml': {
        'format': 'xml',
        'batch_constraint': ('byte_size', 100)
    },
    'items3.json': {
        'format': 'json',
        'batch_constraint': (CustomBatch, 10)
    },
}


# builtin triggers
FEED_BATCH_TRIGGER_BASE = {                             
    'item_count': 'scrapy.utils.feedbatch.BatchByXItems',
    'byte_size': 'scrapy.utils.feedbatch.BatchByXBytes',
    'time_interval': 'scrapy.utils.feedbatch.BatchByXMins',
}
```
To stop and create a new batch from the Spider itself a signal can be used. This will require the user to create a method ```self.trigger_batch(feed_uri)``` which will send signal ```signal.stop_batch``` with the feed's URI as argument which can then be intercepted by ```FeedExporter``` and appropriately call a method to stop and start a new batch for the specified feed.

Example custom spider code:
```python
class MySpider(Spider):
    # ...
    # user spider code
    # ...

    def trigger_batch(self, uri):
        self.crawler.send_catch_log(
            signal=signals.stop_batch,
            uri=uri
        )
```

Customizing current codebase so that user can create their own trigger will be very cumbersome and inconvenient.

## Additional context

drs-11:
Batch creation was a recently introduced feature but was limited to only item count constraints. Batch Triggers will be able to make the constraints flexible and give more control to user to create batches.

Resolves #5169.

### DIFF
--- a/scrapy/extensions/batches.py
+++ b/scrapy/extensions/batches.py
@@ -1,0 +1,73 @@
+"""
+Extensions for batch processing and support.
+"""
+
+
+class Batch:
+    """
+    Batch which will store information for current batches and provides
+    suitable methods to check and update batch info.
+    """
+
+    def __init__(self, feed_options):
+        self.feed_options = feed_options
+        self.batch_item_count = self.feed_options["batch_item_count"]
+        # get time duration constraint from feed-settings
+        # get file size constraint from feed-settings
+
+        self.current_item_count = 0
+        # add time duration parameter here
+        # add file size parameter here
+        self.batch_id = 0
+
+        self.file = None
+        self.updated_once = False
+        self.enabled = True
+        if not self.batch_item_count:
+            self.enabled = False
+
+    def update(self):
+        """
+        Update batch states parameters.
+        """
+        self.current_item_count += 1
+        # update time duration parameter
+        # update file size parameter
+
+        if not self.updated_once:
+            self.updated_once = True
+
+    def should_trigger(self):
+        """
+        Check if any batch state parameter value has crossed its
+        specified constraint.
+        :return: `True` if parameter value has crossed constraint, else `False`
+        :rtype: bool
+        """
+        if not self.enabled:
+            return False
+
+        # add file size and time duration checks
+        return self.current_item_count >= self.batch_item_count
+
+    def new_batch(self, file):
+        """
+        Resets parameter values back to its initial value and increments
+        self.batch_id.
+        """
+        self.file = file
+        self.current_item_count = 0
+        # reset other parameters as well
+        self.updated_once = False
+        self.batch_id += 1
+
+    def get_batch_state(self):
+        """
+        Get current batch state.
+        :return: A dictionary containing batch state parameters and its current value.
+        :rtype: dict
+        """
+        state = {
+            'itemcount': self.current_item_count,   # add current file size and time elapsed
+        }
+        return state
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -7,6 +7,7 @@ See documentation in docs/topics/feed-exports.rst
 import logging
 import os
 import re
+from scrapy.extensions.batches import Batch
 import sys
 import warnings
 from datetime import datetime
@@ -259,7 +260,6 @@ class _FeedSlot:
         self.uri = uri
         self.filter = filter
         # flags
-        self.itemcount = 0
         self._exporting = False
 
     def start_exporting(self):
@@ -289,6 +289,7 @@ class FeedExporter:
         self.feeds = {}
         self.slots = []
         self.filters = {}
+        self.batches = {}
 
         if not self.settings['FEEDS'] and not self.settings['FEED_URI']:
             raise NotConfigured
@@ -302,15 +303,19 @@ class FeedExporter:
             )
             uri = str(self.settings['FEED_URI'])  # handle pathlib.Path objects
             feed_options = {'format': self.settings.get('FEED_FORMAT', 'jsonlines')}
-            self.feeds[uri] = feed_complete_default_values_from_settings(feed_options, self.settings)
+            feed_options = feed_complete_default_values_from_settings(feed_options, self.settings)
+            self.feeds[uri] = feed_options
             self.filters[uri] = self._load_filter(feed_options)
+            self.batches[uri] = self._load_batch(feed_options)
         # End: Backward compatibility for FEED_URI and FEED_FORMAT settings
 
         # 'FEEDS' setting takes precedence over 'FEED_URI'
         for uri, feed_options in self.settings.getdict('FEEDS').items():
             uri = str(uri)  # handle pathlib.Path objects
-            self.feeds[uri] = feed_complete_default_values_from_settings(feed_options, self.settings)
+            feed_options = feed_complete_default_values_from_settings(feed_options, self.settings)
+            self.feeds[uri] = feed_options
             self.filters[uri] = self._load_filter(feed_options)
+            self.batches[uri] = self._load_batch(feed_options)
 
         self.storages = self._load_components('FEED_STORAGES')
         self.exporters = self._load_components('FEED_EXPORTERS')
@@ -326,7 +331,6 @@ class FeedExporter:
         for uri, feed_options in self.feeds.items():
             uri_params = self._get_uri_params(spider, feed_options['uri_params'])
             self.slots.append(self._start_new_batch(
-                batch_id=1,
                 uri=uri % uri_params,
                 feed_options=feed_options,
                 spider=spider,
@@ -341,14 +345,16 @@ class FeedExporter:
         return defer.DeferredList(deferred_list) if deferred_list else None
 
     def _close_slot(self, slot, spider):
-        if not slot.itemcount and not slot.store_empty:
+        if not self.batches[slot.uri_template].updated_once and not slot.store_empty:
             # We need to call slot.storage.store nonetheless to get the file
             # properly closed.
             return defer.maybeDeferred(slot.storage.store, slot.file)
         slot.finish_exporting()
-        logfmt = "%s %%(format)s feed (%%(itemcount)d items) in: %%(uri)s"
+        logfmt = "%s %%(format)s feed in: %%(uri)s, batch state: %%(batch_state)s"
+        batch_state = self.batches[slot.uri_template].get_batch_state()
+        batch_state_str = ", ".join("{}={}".format(k, v) for k, v in batch_state.items())
         log_args = {'format': slot.format,
-                    'itemcount': slot.itemcount,
+                    'batch_state': batch_state_str,
                     'uri': slot.uri}
         d = defer.maybeDeferred(slot.storage.store, slot.file)
 
@@ -375,7 +381,7 @@ class FeedExporter:
         )
         self.crawler.stats.inc_value(f"feedexport/success_count/{slot_type}")
 
-    def _start_new_batch(self, batch_id, uri, feed_options, spider, uri_template):
+    def _start_new_batch(self, uri, feed_options, spider, uri_template):
         """
         Redirect the output data stream to a new file.
         Execute multiple times if FEED_EXPORT_BATCH_ITEM_COUNT setting or FEEDS.batch_item_count is specified
@@ -387,6 +393,7 @@ class FeedExporter:
         """
         storage = self._get_storage(uri, feed_options)
         file = storage.open(spider)
+        self.batches[uri_template].new_batch(file)
         exporter = self._get_exporter(
             file=file,
             format=feed_options['format'],
@@ -402,7 +409,7 @@ class FeedExporter:
             uri=uri,
             format=feed_options['format'],
             store_empty=feed_options['store_empty'],
-            batch_id=batch_id,
+            batch_id=self.batches[uri_template].batch_id,
             uri_template=uri_template,
             filter=self.filters[uri_template]
         )
@@ -419,16 +426,12 @@ class FeedExporter:
 
             slot.start_exporting()
             slot.exporter.export_item(item)
-            slot.itemcount += 1
-            # create new slot for each slot with itemcount == FEED_EXPORT_BATCH_ITEM_COUNT and close the old one
-            if (
-                self.feeds[slot.uri_template]['batch_item_count']
-                and slot.itemcount >= self.feeds[slot.uri_template]['batch_item_count']
-            ):
+            self.batches[slot.uri_template].update()
+            # create new slot if slot's batch should trigger and close the old one
+            if self.batches[slot.uri_template].should_trigger():
                 uri_params = self._get_uri_params(spider, self.feeds[slot.uri_template]['uri_params'], slot)
                 self._close_slot(slot, spider)
                 slots.append(self._start_new_batch(
-                    batch_id=slot.batch_id + 1,
                     uri=slot.uri_template % uri_params,
                     feed_options=self.feeds[slot.uri_template],
                     spider=spider,
@@ -455,11 +458,11 @@ class FeedExporter:
 
     def _settings_are_valid(self):
         """
-        If FEED_EXPORT_BATCH_ITEM_COUNT setting or FEEDS.batch_item_count is specified uri has to contain
+        If batch class or related contraints are specified, uri has to contain
         %(batch_time)s or %(batch_id)d to distinguish different files of partial output
         """
-        for uri_template, values in self.feeds.items():
-            if values['batch_item_count'] and not re.search(r'%\(batch_time\)s|%\(batch_id\)', uri_template):
+        for uri_template, batch in self.batches.items():
+            if batch.enabled and not re.search(r'%\(batch_time\)s|%\(batch_id\)', uri_template):
                 logger.error(
                     '%(batch_time)s or %(batch_id)d must be in the feed URI ({}) if FEED_EXPORT_BATCH_ITEM_COUNT '
                     'setting or FEEDS.batch_item_count is specified and greater than 0. For more info see: '
@@ -532,3 +535,8 @@ class FeedExporter:
         # load the item filter if declared else load the default filter class
         item_filter_class = load_object(feed_options.get("item_filter", ItemFilter))
         return item_filter_class(feed_options)
+
+    def _load_batch(self, feed_options):
+        # load batch class if declared else load default base class
+        batch_class = load_object(feed_options.get("batch", Batch))
+        return batch_class(feed_options)


scrapy/extensions/feedexport.py
```diff
@@ -7,6 +7,7 @@
 import logging
 import os
 import re
+from scrapy.extensions.batches import Batch
```
Gallaecio, line 10: This should be along the rest of imports from `scrapy` below, and keeping the alphabetical order.

scrapy/extensions/feedexport.py
```diff
@@ -341,14 +345,16 @@ def close_spider(self, spider):
         return defer.DeferredList(deferred_list) if deferred_list else None
 
     def _close_slot(self, slot, spider):
-        if not slot.itemcount and not slot.store_empty:
+        if not self.batches[slot.uri_template].updated_once and not slot.store_empty:
```
Gallaecio, line 348: Would `current_item_count` work instead of `updated_once`? (and remove the need to define `updated_once` in the `Batch` class)

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,73 @@
...
+    suitable methods to check and update batch info.
+    """
+
+    def __init__(self, feed_options):
+        self.feed_options = feed_options
+        self.batch_item_count = self.feed_options["batch_item_count"]
+        # get time duration constraint from feed-settings
+        # get file size constraint from feed-settings
+
+        self.current_item_count = 0
```
Gallaecio, line 18: Why not just `item_count`?

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,73 @@
...
+
+class Batch:
+    """
+    Batch which will store information for current batches and provides
+    suitable methods to check and update batch info.
+    """
+
+    def __init__(self, feed_options):
+        self.feed_options = feed_options
+        self.batch_item_count = self.feed_options["batch_item_count"]
```
Gallaecio, line 14: What about `self.max_item_count`?

scrapy/extensions/feedexport.py
```diff
@@ -341,14 +345,16 @@ def close_spider(self, spider):
...
             # properly closed.
             return defer.maybeDeferred(slot.storage.store, slot.file)
         slot.finish_exporting()
-        logfmt = "%s %%(format)s feed (%%(itemcount)d items) in: %%(uri)s"
+        logfmt = "%s %%(format)s feed in: %%(uri)s, batch state: %%(batch_state)s"
+        batch_state = self.batches[slot.uri_template].get_batch_state()
+        batch_state_str = ", ".join("{}={}".format(k, v) for k, v in batch_state.items())
         log_args = {'format': slot.format,
-                    'itemcount': slot.itemcount,
+                    'batch_state': batch_state_str,
```
Gallaecio, line 357: I don’t think we need to modify the logged message. If the goal is for the logs to explain why the batch was triggered, it probably makes sense to log a message about that separately. This message is about informing users that a file with N items has been stored in a given location, regardless of the reason (time, number of items, etc.).

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,73 @@
+"""
+Extensions for batch processing and support.
+"""
+
+
+class Batch:
```
Gallaecio, line 6: I think `Batch` is a bad name, since this class does not represent a single batch. Maybe something in the line of `BatchTracker` or `BatchHandler`?

Gallaecio (commented):
I’m not completely sold on the changes so far, but mostly because I find it hard to see how they will evolve to support time-based and size-based limits. I suspect the changes will look very differently once you start supporting those limits as well, so I think it may be better to wait until you get that part working to provide feedback about the implementation.

scrapy/extensions/feedexport.py
drs-11: Since the goal is to provide more batch triggers than just item count, I had thought of this so that when the file size and time duration are introduced as well or maybe a custom batch class is being used, `updated_once` can be used to determine if a batch is empty.

scrapy/extensions/feedexport.py
drs-11: I have no idea how it got there really :laughing:

scrapy/extensions/batches.py
drs-11: I was having a bad day naming variables :D. I think `BatchHandler` would do nicely.

scrapy/extensions/feedexport.py
drs-11: I thought that using that log message to convey the batch state will serve both points: why batch was triggered and what was the last state when it was closed. Though I see your point and I will try that.

drs-11:
>  I suspect the changes will look very differently once you start supporting those limits as well

Yeah I think it's better I complete with other constraints as well, this code is confusing right now.

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if not self.enabled:
+            return False
+
+        if self.max_item_count and self.item_count >= self.max_item_count:
+            return True
+        if self.max_file_size and self.file_size >= self.max_file_size:
+            return True
+        if self.max_time_duration and self.elapsed_time >= self.max_time_duration:
+            return True
```
drs-11, line 60: These are soft limits for now. Accurate limits are planned for later.

drs-11:
For hard limit on file size, I was thinking maybe having a safety margin on currently stored file size will be better than having to get the data size of the exported item. Because if the item is already exported, it can go above the `self.max_file_size` limit but having a safety margin, we can trigger the batch before it passes the limit.

So if `self.file_size + self.safety_margin` is exceeded we can trigger the batch creation process. For choosing a `safety_margin` we can either predefine it or dynamically update it with the average of the size of the exported data so far.

Gallaecio:
While that is closer to a hard limit, I would only go that route if a proper solution is not feasible.

Could you check if it is possible to write code right before the bytes of an item are written into the actual end file, so that we can do `len(data)` on them and, if the current file length plus those would make the file go over the limit, switch batches and instead write those bytes into the next batch?

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,115 @@
...
+        h, m, s = map(int, duration.split(":"))
+        duration_in_secs = h * 60 * 60 + m * 60 + s
+        return duration_in_secs
+
+    def _in_bytes(self, size: str) -> int:
+        """
+        Convert string size in format: '<SIZE><UNIT>' to bytes in integer.
+        """
+        # https://stackoverflow.com/a/60708339/7116579
+        units = {"B": 1, "KB": 2**10, "MB": 2**20, "GB": 2**30, "TB": 2**40}
```
Gallaecio, line 104: See https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

I would make these units based on powers of 10, and support also binary prefixes for units expressed as powers of 2.

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,115 @@
...
+        size = size.upper()
+        if not re.match(r' ', size):
+            size = re.sub(r'([KMGT]?B)', r' \1', size)
+        number, unit = [string.strip() for string in size.split()]
```
Gallaecio, line 108:
```suggestion
        match = re.search(r'(?i)^\s*(\d+)\s*((?:[kMGT]i?)?B)\s*$', size)
        if not match:
            raise ValueError(f'Invalid batch size: {size!r}')
        number, unit = match[1], match[2].upper()
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,115 @@
...
+        return int(float(number) * units[unit])
+
+    def _calculate_elapsed_time(self) -> int:
+        return time.time() - self.start_time
+
+    def _calculate_batch_size(self) -> int:
+        return self.file.tell()
```
Gallaecio, line 115: These 2 functions seem superfluous, since they are one-liners used only once, where their return value is assigned to a variable whose name makes it clear what the outcome is.

scrapy/extensions/feedexport.py
```diff
@@ -420,15 +426,12 @@ def item_scraped(self, item, spider):
             slot.start_exporting()
             slot.exporter.export_item(item)
             slot.itemcount += 1
-            # create new slot for each slot with itemcount == FEED_EXPORT_BATCH_ITEM_COUNT and close the old one
-            if (
-                self.feeds[slot.uri_template]['batch_item_count']
-                and slot.itemcount >= self.feeds[slot.uri_template]['batch_item_count']
-            ):
+            self.batches[slot.uri_template].update()
+            # create new slot if slot's batch should trigger and close the old one
```
Gallaecio, line 430: This comment does not seem to add much information to what is already clear from the code, I think.

scrapy/extensions/feedexport.py
```diff
@@ -532,3 +535,8 @@ def _load_filter(self, feed_options):
         # load the item filter if declared else load the default filter class
         item_filter_class = load_object(feed_options.get("item_filter", ItemFilter))
         return item_filter_class(feed_options)
+
+    def _load_batch(self, feed_options):
+        # load batch class if declared else load default batch class
```
Gallaecio, line 540: This comment also seems superfluous.

drs-11:
Will add custom batch handler tests soon.

tests/test_feedexport.py
```diff
@@ -1587,16 +1590,60 @@ def test_export_items(self):
...
+            self.MyItem({'foo': 'bar1', 'egg': 'spam1', 'baz': 'quux1'}),
+            self.MyItem({'foo': 'bar2', 'egg': 'spam2', 'baz': 'quux2'}),
+        ]
+        fmt_to_limits = {
+            'json': ('40B', '90B'),
+            'jsonlines': ('40B', '90B'),
+            'xml': ('100B', '160B'),
+            'csv': ('30B', '40B'),
+            'marshal': ('30B', '70B'),
+            'pickle': ('50B', '110B'),
```
Gallaecio, line 1605: Looks good. However, I still think it would be best to be more precise, to use a pair of values that only differs in 1 byte.

In fact, I think you can be more precise without actually having to guess the right value to put here. Which is great, because I suspect for pickle and marshal that value may change depending on the version of Python that you use.

When you run the `1B` test, you can use the `exported_data` to tell what exact size the first file has, so using that size as maximum should still cause 2 files (`self.file_size >= self.max_file_size`), while that size plus 1 should cause a single file. So, instead of hardcoding sizes here, you could use the output of `self.exported_data(items, settings)` for `1B` for the tests based on higher sizes.

tests/test_feedexport.py
```diff
@@ -1587,16 +1590,60 @@ def test_export_items(self):
...
+            for delay, expected in {'0:0:0': 1, '0:0:0.01': 2 + 1, '0:0:0.1': 2 + 1, '0:0:1.2': 1}.items():
+                settings = {
+                    'FEEDS': {
+                        os.path.join(self._random_temp_filename(), fmt, self._file_mark): {'format': fmt},
+                    },
+                    'FEED_EXPORT_BATCH_DURATION': delay,
+                    'DOWNLOAD_DELAY': 0.5,
+                }
+                data = yield self.exported_data(items, settings, multiple_urls=2)
+                self.assertEqual(expected, len(data[fmt]))
```
Gallaecio, line 1633: This test, with a single item, seems to indirectly demonstrate that the current implementation will cause empty files to be created sometimes.

I think this is undesired behavior for the ideal batch implementation, i.e. ideally we should only create files with items. But I also think this behavior is OK for an initial implementation, and I think it is a consequence of pre-existing batch-handling code.

tests/test_feedexport.py
```diff
@@ -1587,16 +1590,60 @@ def test_export_items(self):
...
+                }
+                data = yield self.exported_data(items, settings)
+                self.assertEqual(expected, len(data[fmt]))
+
+    @defer.inlineCallbacks
+    def test_export_duration_trigger(self):
+        items = [self.MyItem({'foo': 'bar1', 'egg': 'spam1', 'baz': 'quux1'})]
+        for fmt in ('csv', 'json', 'jsonlines', 'marshal', 'pickle'):
+            #  +1 for empty files after batch has been triggered
+            for delay, expected in {'0:0:0': 1, '0:0:0.01': 2 + 1, '0:0:0.1': 2 + 1, '0:0:1.2': 1}.items():
```
Gallaecio, line 1624: Why is the last value 1 instead of 1+1? Why is the behavior with duration different than with file size above?

Gallaecio (commented):
Good work on those tests!

The coverage data in https://app.codecov.io/gh/scrapy/scrapy/compare/5205/diff is very confusing at the moment (there are changes there not in the pull request), but I can see 1 points where action is needed:

- Test an invalid batch size
- Test `feed_options` being `None` at `feedexport.py:64` (it seems like this was covered in the past)

tests/test_feedexport.py
drs-11: Yes I think in the original batch trigger PR this was pointed out. Here: [https://github.com/scrapy/scrapy/pull/4434#issuecomment-658672328](https://github.com/scrapy/scrapy/pull/4434#issuecomment-658672328).

tests/test_feedexport.py
drs-11: Ok yeah it should be trigger after getting at most 2 items. I will change that.

tests/test_feedexport.py
drs-11: Got it!

tests/test_feedexport.py
Gallaecio: This is slightly different, though. The initial implementation creates 1 extra empty file. Somehow this test, with 1 item, can create 3 different files. I’m actually not sure how this happens, but it seems to be happening because the test passes. Any idea why?

drs-11: Oh I see, I should have put a comment explaining it. So to simulate item scraping delays I am using `DOWNLOAD_DELAY` to delay url downloads. So as to get one item per download. That's why I am using `multiple_urls` argument in the `exported_data` function. You suggested that trick for testing hard limit duration :)

drs-11: So I am actually getting total of 2 items, one from each url download.

Gallaecio: :facepalm:

That makes total sense, sorry for the noise!

tests/test_feedexport.py
drs-11: image

In the last entry, I was expecting item being added after a delay of 0.5 seconds at least but it was added sooner. So the batch wasn't triggered and there was only 1 batch instead of 1 (with 2 items) + 1 empty batch. Kinda hard to get at-most-2-items batch everytime. Should I continue with expecting 1 batch with 2 items duration?

Gallaecio: Maybe you need to also limit concurrency to 1? It’s 8 by default, so multiple requests can happen in parallel.

drs-11: Tried `CONCURRENT_REQUESTS: 1` still getting some results with delays less than 0.5 secs :(

Gallaecio: Can you share code that I can use to reproduce the issue easily (i.e. what you used to print the text in the screenshot) in some branch of your Scrapy fork?

drs-11: [https://github.com/drs-11/scrapy/tree/batch-duration-debug](https://github.com/drs-11/scrapy/tree/batch-duration-debug)

I did the usual one item per url and two urls. The wrong delays does not happen everytime so you may need to run the crawl multiple times.

Gallaecio: Oh, I bet it’s https://docs.scrapy.org/en/latest/topics/settings.html#std-setting-RANDOMIZE_DOWNLOAD_DELAY

drs-11: YES! Getting consistent delays now :sweat_smile:.

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,140 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+
+.. versionadded:: VERSION
+
+Batches enable multiple sequential files to be generated for a feed instead of getting
+a single file. This is useful in getting item deliveries earlier before the crawl has
+been ended.
+
+To enable batching, you can use :ref:`criterias<batch_criterias>` such as ``batch_item_count``
+, ``batch_file_size``, ``batch_duration`` or you can create :ref:`your own batch
+handlers<custom_batch_handler>` to handle batch deliveries.
```
Gallaecio, line 288: This is written from the perspective of a user that is using the default batch handler. While this will be true for most users, the documentation should be accurate.

What about:

```suggestion
Batching is controlled by a batch handler, which is defined through the 
:ref:`batch feed option <feed-options>`.

For information about enabling and configuring the default batch handler, see
:class:`~scrapy.extensions.batches.BatchHandler`. You can alternatively 
:ref:`create your own batch handler <custom_batch_handler>`.
```

Then you could create a section called something like “Built-in Batch Handler”, and use `.. autoclass:: scrapy.extensions.batches.BatchHandler` there to provide documentation about the handler directly from its docstring. Then you can place in its docstring all the information about enabling and configuring the default batch handler, including what you currently have under the “Batch Criterias” section below (by the way, it’s “criteria” even in plural), without implying that every batch handler is supposed to be enabled and configured the same way.

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,140 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+
+When generating multiple output files, you must use at least one of the following
+placeholders in the feed URI to indicate how the different output file names are
+generated:
```
Gallaecio, line 292: You can probably replace this whole part about batch feed URIs with a paragraph that simply states that, if you enable batching, you must update your feed URI accordingly, and then link to a new section that contains all the information here.

This is to avoid the most part of the introduction about batches being something as specific as the handling of the feed URI. It also allows us to link to this information from other parts of the documentation and from the outside.

docs/topics/feed-exports.rst
```diff
@@ -389,6 +523,20 @@ as a fallback value if that key is not provided for a specific feed definition:
...
+
+-   ``batch_duration``: falls back to
+    :setting:`FEED_EXPORT_BATCH_DURATION`.
+
+    .. versionadded:: VERSION
+
+-   ``batch``: a custom batch handler to handle batch deliveries.
+
+    .. versionadded:: VERSION
+
```
Gallaecio, line 539: In the documentation about the batch settings, it should be clear that they only apply if using the default batch handler.

docs/topics/feed-exports.rst
```diff
@@ -602,39 +750,34 @@ Default: ``0``
...
-    --->2-filename2020-03-28T14-45-09.148903.json
-    --->3-filename2020-03-28T14-45-10.046092.json
+Default: ``0:0:0``
 
-Where the first and second files contain exactly 100 items. The last one contains
-100 items or fewer.
+If assigned a duration longer than ``0:0:0``, Scrapy generates multiple output files based on
+the specified duration as a soft limit.
 
+Duration must be a string in the format ``hours:months:seconds``.
```
Gallaecio, line 780: The settings documentation should also clarify the scope, the default batch handler.

You could also remove the settings from here, mention them in the docstring of the default batch handler as equivalents to the corresponding feed option keys (indicating which one has a higher priority over the other), and move the `.. setting:: …` anchors to right above the paragraph or list item that mentions each setting.

docs/topics/feed-exports.rst
```diff
@@ -659,17 +802,17 @@ The function signature should be as follows:
         Specifically:
 
         -   ``batch_id``: ID of the file batch. See
-            :setting:`FEED_EXPORT_BATCH_ITEM_COUNT`.
+            :ref:`Batches<batches>`.
 
-            If :setting:`FEED_EXPORT_BATCH_ITEM_COUNT` is ``0``, ``batch_id``
-            is always ``1``.
+            If a batch handler is not enabled or no batch trigger
+            is passed, ``batch_id`` is always ``1``.
```
Gallaecio, line 808: What about something more generic?

```suggestion
            If batching is not enabled, ``batch_id`` is always ``1``.
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,132 @@
...
+
+    - ``FEED_EXPORT_BATCH_FILE_SIZE``: Equivalent to ``batch_file_size``.
+
+        Default: ``0B``
+
+    .. setting:: FEED_EXPORT_BATCH_DURATION
+
+    - ``FEED_EXPORT_BATCH_DURATION``: Equivalent to ``batch_duration``.
+
+        Default: ``0:0:0``
```
Gallaecio, line 56: What about merging this list into the one above?

For example:

```
To activate, define one of the following :ref:`feed options <feeds-options>` or 
:ref:`settings <topics-settings>`:

.. setting:: FEED_EXPORT_BATCH_ITEM_COUNT

-   ``batch_item_count`` feed option or ``FEED_EXPORT_BATCH_ITEM_COUNT`` 
    setting (:class:`int`): the maximum number of items a batch can have.
    
…

Each feed option overrides its counterpart setting.

When using more than one type of limit, whichever limit comes first triggers a 
new batch file.
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,132 @@
...
+        self.batch_id: int = 0
+        # misc attributes
+        self.file: BinaryIO
+        self.start_time: float
+        self.enabled: bool = any([self.max_item_count, self.max_seconds, self.max_file_size])
+
+    def item_added(self) -> None:
+        """
+        Update batch state attributes.
+        """
```
Gallaecio, line 77: For a class implementing an interface, such as this one, you usually do not need to document the interface methods. The documentation of a method is meant to describe what it does, not how it does it, and that’s usually already covered in the documentation of the interface.

If your docstrings are meant to describe how a method works, and you feel this information is needed as a comment because the implementation or some aspect of it is not clear from code, you should use code comments for such information, not a docstring.

drs-11: Yeah sorry I repeated this in postprocessing as well.

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,137 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+    :type feed_options: :class:`dict`
+
+    Required attributes:
+
+    * `self.batch_id` (:class:`int`): id number of current batch, must be initialised to 0.
+    * `self.enabled` (:class:`bool`): must be `True` if a non-empty limit is passed, otherwise `False`.
+
+.. method:: item_added(self)
+
+    Update batch state info. Will be called everytime an item is exported to feed.
```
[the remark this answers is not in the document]
drs-11: Moving internal data updating code to `should_trigger` will make `item_added` method useless. I guess I should remove `item_added` if we are doing all the updating in `should_trigger`?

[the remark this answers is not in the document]
Gallaecio: I’m slightly inclined to keep it because I think it has some potential for extension in the future. I’m thinking that in the future (or maybe now?) we could add an `item` parameter to it, so that a batch handler can trigger batch switches based on item content.

I’m also thinking that we might want to eventually have a counterpart `before_item_added` method, which would be called before an item is added to a batch and get not only the `item` as a parameter, but possibly also the `data` that will be written into the file, as both things could also be used by a batch handler to determine whether or not to switch to a new batch.

Instead of removing `item_added`, maybe we could remove `should_trigger` and have `item_adapter` work the way `should_trigger` does, so instead of:

```python
self.batches[slot.uri_template].item_added()
if self.batches[slot.uri_template].should_trigger():
    ...
```

we could have:

```python
create_new_batch = self.batches[slot.uri_template].item_added()
if create_new_batch:
    ...
```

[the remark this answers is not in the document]
drs-11: Oh nice, this could also be helpful in implementing hard limit for file size.

scrapy/extensions/batches.py
```diff
@@ -99,29 +85,20 @@ def should_trigger(self) -> bool:
...
+        # Convert duration string in format: '<HOURS>:<MINUTES>:<SECONDS>' to seconds in float.
         h, m, s = map(float, duration.split(":"))
         duration_in_secs = h * 60 * 60 + m * 60 + s
         return duration_in_secs
 
     def _in_bytes(self, size: str) -> int:
-        """
-        Convert string size in format: '<SIZE><UNIT>' to bytes in integer.
-        """
+        # Convert string size in format: '<SIZE><UNIT>' to bytes in integer.
```
Gallaecio: These were docstrings of non-interface methods that described what the methods did (not _how_ they did it), so I believe docstrings were appropriate.

scrapy/extensions/feedexport.py
```diff
@@ -470,8 +472,11 @@ def _settings_are_valid(self):
         If batch class or related contraints are specified, uri has to contain
         %(batch_time)s or %(batch_id)d to distinguish different files of partial output
         """
-        for uri_template, batch in self.batches.items():
-            if batch.enabled and not re.search(r'%\(batch_time\)s|%\(batch_id\)', uri_template):
+        for uri_template, values in self.feeds.items():
+            if (
+                ('batch' in values or self.batches[uri_template].enabled)
+                and not re.search(r'%\(batch_time\)s|%\(batch_id\)', uri_template)
+            ):
```
Gallaecio, line 479: Now that batch handlers do not need to define `enabled`, this probably needs an update.

You should also add a test that would have caught this issue.

drs-11: So what I am attempting here is giving only the default batch handler, an `enabled` attribute so as to know if any in built criteria has been activated. This is not needed in custom batch handlers. This I think is an implementation detail for the default batch handler only which also has `NotConfigured` exception for incorrect values.

drs-11: So I think it's now more closer to being a Scrapy component than before :sweat_smile:

drs-11: Actually now that I think about it we can remove the automatic default batch handler assignment for non batch feeds.

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,129 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+
+    :type file: :class:`BinaryIO`
+
+An example for custom batch handler::
+
+    class CustomBatchHandler:
+
+        def __init__(self, feed_options):
+            if "batch_divisible_by" not in feed_options or feed_options["batch_divisible_by"] <= 0
+                raise NotConfigured
```
Gallaecio, line 345:
```suggestion
            if feed_options.get("batch_divisible_by", 0) <= 0:
                raise NotConfigured
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,114 @@
...
+
+      The file size format is ``<number><unit>``, where ``<unit>`` is a byte
+      unit based on powers of 2 (KiB, MiB, GiB, TiB) or powers of 10 (kB, MB,
+      GB, TB). Eg: 200MB, 100KiB.
+
+      Default: ``0B``
+
+    .. setting:: FEED_EXPORT_BATCH_DURATION
+
+    - ``batch_duration`` feed options or ``FEED_EXPORT_BATCH_DURATION``
```
Gallaecio, line 37:
```suggestion
    - ``batch_duration`` feed option or ``FEED_EXPORT_BATCH_DURATION``
```

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,129 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+.. method:: __init__(self, feed_options)
+
+    Initialize the batch handler.
+
+    :param feed_options: feed-specific :ref:`options <feed-options>`
+
+    :type feed_options: dict
+
+    .. note:: Feed options support arbitrary parameters passing which, can be
+              used by custom batch handlers to access parameters passed by users.
```
Gallaecio, line 319:
```suggestion
    .. note:: Feed options support arbitrary parameters passing, which can be
              used by custom batch handlers to access parameters passed by users.
```

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,129 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+
+.. method:: item_added(self)
+
+    Called every time after an item is exported to the feed.
+    
+    Return `True` if a new batch file should be created now, or `False` otherwise.
```
Gallaecio, line 325: My bad :sweat_smile: 

```suggestion
    Return ``True`` if a new batch file should be created now, or ``False`` otherwise.
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,114 @@
...
+      setting (:class:`str`): deliver a batch file after at least this much time has passed.
+
+      The duration format is ``hours:minutes:seconds``. Eg: 1:00:00 for 1 hour,
+      0:30:00 for a 30 minute duration.
+
+      Duration is only checked after an item is added to the batch file.
+
+      Default: ``0:0:0``
```
Gallaecio, line 45:
```suggestion
      Default: ``0:00:00``
```

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, feed_options: Dict[str, Any]) -> None:
+        if not any(k in feed_options for k in ("batch_item_count", "batch_duration", "batch_file_size")):
+            raise NotConfigured
+        # get limits from feed_settings
+        self.max_item_count: int = feed_options.get("batch_item_count", 0)
+        self.max_seconds: float = self._in_seconds(feed_options.get("batch_duration", "0:0:0"))
+        self.max_file_size: int = self._in_bytes(feed_options.get("batch_file_size", "0B"))
+        if any(limit < 0 for limit in (self.max_file_size, self.max_item_count, self.max_seconds)):
+            raise NotConfigured
```
Gallaecio, line 61:
```suggestion
        if not any(limit > 0 for limit in (self.max_file_size, self.max_item_count, self.max_seconds)):
            raise NotConfigured
```

scrapy/extensions/feedexport.py
Gallaecio: `self.enable` will always be `True`, because in scenarios where it would be `False` a `NotConfigured` exception is raised, so the batch handler object is never created or used.

scrapy/extensions/batches.py
```diff
@@ -0,0 +1,114 @@
...
+        Convert duration string in format: '<HOURS>:<MINUTES>:<SECONDS>' to seconds in float.
+        '''
+        h, m, s = map(float, duration.split(":"))
+        duration_in_secs = h * 60 * 60 + m * 60 + s
+        return duration_in_secs
+
+    def _in_bytes(self, size: str) -> int:
+        '''
+        Convert string size in format: '<SIZE><UNIT>' to bytes in integer.
+        '''
```
Gallaecio, line 106: PEP-8

```suggestion
    def _in_seconds(self, duration: str) -> float:
        """Convert duration string in format: '<HOURS>:<MINUTES>:<SECONDS>' to seconds in float."""
        h, m, s = map(float, duration.split(":"))
        duration_in_secs = h * 60 * 60 + m * 60 + s
        return duration_in_secs

    def _in_bytes(self, size: str) -> int:
        """Convert string size in format: '<SIZE><UNIT>' to bytes in integer."""
```

Gallaecio (commented):
We need tests that check how different combinations of options being defined affects whether or not the batch handler is enabled.

drs-11:
> We need tests that check how different combinations of options being defined affects whether or not the batch handler is enabled.

This is remaining, I will put them up soon.

docs/topics/feed-exports.rst
```diff
@@ -295,7 +295,7 @@ If you enable batching, you must also :ref:`set feed URIs appropriately with pla
 Default Batch Handler
 ----------------------
 
-.. autoclass:: scrapy.extensions.batches.BatchHandler
+.. autoclass:: scrapy.extensions.batches.DefaultBatchHandler
```
Gallaecio, line 298: I think it may be better to leave the previous name. The new name could be quite confusing if we switch the default value to a different one in the future.

scrapy/extensions/feedexport.py
```diff
@@ -550,6 +550,11 @@ def _load_filter(self, feed_options):
...
 
-    def _load_batch(self, feed_options):
-        batch_class = load_object(feed_options.get("batch", BatchHandler))
-        return batch_class(feed_options)
+    def _load_batch(self, feed_options, uri):
+        batch_class = load_object(feed_options.get("batch", DefaultBatchHandler))
+        try:
+            batch_obj = batch_class(feed_options)
+        except NotConfigured:
+            logger.warning("Batch Handler %s not configured for %s." % (batch_class.__name__, uri))
```
Gallaecio, line 558: This should probably be not be a warning, since batch handling not being configured is not only OK, but will also be the case most of the time. I’m not sure we need a message here in the first place, `NotConfigured` means the component controlled the issue. We can leave up to the component to log a warning if it is warranted.

drs-11:
Hmm `:noindex:` doesn't seem to fix the `sphinx-build` error. Any clues how to fix it? The error is being reported because of two `__init__` methods in the documentation.

docs/topics/feed-exports.rst
```diff
@@ -267,10 +267,129 @@ once all the file contents have been written (i.e. at the end of the crawl) is
...
+
+Custom Batch Handler
+--------------------
+
+To use a custom batch handler, write a custom batch handler class and assign 
+that class or its import string to the :ref:`batch feed option <feed-options>`.
+
+Each batch handler is a class that must implement the following methods:
+
+.. method:: __init__(self, feed_options)
```
Gallaecio, line 310: I think you need to add `:noindex:` to every `method`, `class`, and other API entries that do not describe something that is part of Scrapy, but an interface for users to implement themselves. For example:

```suggestion
.. method:: __init__(self, feed_options)
   :noindex:
```

Alternatively, you could wrap them as part of a class (e.g. `.. class:: MyBatchHandler`). I think that’s how we have done it in similar cases in the past. Although if `:noindex:` works, maybe that would be best. Up to you.

Gallaecio (approved):

Gallaecio:
I have found that a custom item exporter that overrides the input `file` will break file-size-based batch spliting:

```py
class JsonLinesGzipItemExporter(JsonLinesItemExporter):

    def __init__(self, file, **kwargs):
        gzfile = gzip.GzipFile(fileobj=file,  mode='wb')
        super(JsonLinesGzipItemExporter, self).__init__(gzfile, **kwargs)
```

Because the file in the batch is not the file where things are written.

I think we must either support this or detect cases like this and log a clear warning that provides instructions to make things work as intended (switch the logic to a post-processing plugin?).

drs-11:
But shouldn't the problem arise when the file is being replaced instead of being overridden? In the example you gave, the data is being stored to the target file after being processed by GzipFile class, so the tell method of the target file should still give correct size. Am I missing something here?

Gallaecio:
I’m guessing GZipFile buffers the data, so it is not actually written into the file object until enough data is written into the GZipFile wrapper.

drs-11:
I guess then we should proceed with a warning log to the user, as you suggested. We can compare the file objects of the exporter instance and the one of the batch's, then log a warning if they are different.

drs-11:
A simple solution of detecting file overriding that I thought of was checking if the exporter's file attribute points to the same file as the one opened in `_start_new_batch` or not. Though this may require making it mandatory for all the exporters to have a `file` attribute.

scrapy/exporters.py
```diff
@@ -144,6 +144,7 @@ def __init__(self, file, **kwargs):
         if not self.encoding:
             self.encoding = 'utf-8'
         self.xg = XMLGenerator(file, encoding=self.encoding)
+        self.file = self.xg
```
drs-11, line 147: Oh wait this is wrong, this will generate a warning everytime XML exporter is used. Will fix this.

scrapy/exporters.py
```diff
@@ -209,6 +210,7 @@ def __init__(self, file, include_headers_line=True, join_multivalued=',', errors
             errors=errors,
         )
         self.csv_writer = csv.writer(self.stream, **self._kwargs)
+        self.file = self.csv_writer
```
drs-11, line 213: Same here :sweat_smile:

Gallaecio:
Could you add a test for those last changes?

PR #5209, opened 2021-07-20
Cleanup leftover boto2 code in S3DownloaderHandler

laggardkernel:
Remove leftover attribute `S3DownloaderHandler.conn`. Seems we forgot to remove it in 5e99758.

### DIFF
--- a/scrapy/core/downloader/handlers/s3.py
+++ b/scrapy/core/downloader/handlers/s3.py
@@ -1,5 +1,3 @@
-from urllib.parse import unquote
-
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
 from scrapy.exceptions import NotConfigured
 from scrapy.utils.boto import is_botocore_available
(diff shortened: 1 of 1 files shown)


wRAR:
Please fix the flake8 failure.

laggardkernel:
Fixed the linting error. There's another `requests` connection error raised during generating document. Seems unrelated with my code.

BTW, I have problem running `tox` on my local machine. `pip` stucks at dependencies resolving.

wRAR (approved):

scrapy/core/downloader/handlers/s3.py
```diff
@@ -69,14 +67,4 @@ def download_request(self, request, spider):
             self._signer.add_auth(awsrequest)
             request = request.replace(
                 url=url, headers=awsrequest.headers.items())
-        else:
```
Gallaecio, line 72: The `elif` above can become an `else` now, since `self._signer` is never `None` if `self.anon` is `True`.

Gallaecio:
Thanks!

PR #5211, opened 2021-07-20
Update AWS credentials creation behavior to be consistent across the project

laggardkernel:
Currently we have two ways to create AWS credentials in `scrapy`

- `Session.create_client()` in pipelines, feed exporter
- `Credentials.__init__()` in `S3DownloadHandler`

And `Session.create_client()` searches credentials from:

1. Credentials parameters passed in `Session.create_client()`
2. Environment variables
3. Shared credential file (`~/.aws/credentials`)
4. AWS config file (`~/.aws/config`)
5. Assume Role provider
6. Boto2 config file (`/etc/boto.cfg` and `~/.boto`)
7. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.

Related credentials creation strategy in `Session.create_client()`.
https://github.com/boto/botocore/blob/c329fd23d3672b724a6dbd152f738de0b4d5a7ac/botocore/session.py#L825-L839

The problem is credentials creation in `S3DownloadHandler` behaves differently from credentials creation in other places. 

`Credentials.__init__()` in `S3DownloadHandler` only uses passed in credentials parameters (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`).

References

- [Configuring credentials](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials)
- #7607

### DIFF
--- a/scrapy/core/downloader/handlers/s3.py
+++ b/scrapy/core/downloader/handlers/s3.py
@@ -1,5 +1,7 @@
+import warnings
+
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
-from scrapy.exceptions import NotConfigured
+from scrapy.exceptions import NotConfigured, ScrapyDeprecationWarning
 from scrapy.utils.boto import is_botocore_available
 from scrapy.utils.httpobj import urlparse_cached
 from scrapy.utils.misc import create_instance
@@ -12,6 +14,13 @@ class S3DownloadHandler:
                  aws_access_key_id=None, aws_secret_access_key=None,
                  aws_session_token=None,
                  httpdownloadhandler=HTTPDownloadHandler, **kw):
+        if kw:
+            warnings.warn(
+                "'kw' args are deprecated. If you're trying to set s3 request anonymous."
+                " Use setting 'DOWNLOAD_HANDLERS_S3_ANONYMOUS = True' instead.",
+                category=ScrapyDeprecationWarning,
+                stacklevel=2,
+            )
         if not is_botocore_available():
             raise NotConfigured('missing botocore library')
 
@@ -23,6 +32,8 @@ class S3DownloadHandler:
             aws_session_token = settings['AWS_SESSION_TOKEN']
 
         self.anon = kw.pop('anon', None)
+        if self.anon is None:
+            self.anon = settings.getbool('DOWNLOAD_HANDLERS_S3_ANONYMOUS')
         if kw:
             raise TypeError(f'Unexpected keyword arguments: {kw}')
 
--- a/scrapy/core/downloader/handlers/s3.py
+++ b/scrapy/core/downloader/handlers/s3.py
@@ -13,11 +13,11 @@ class S3DownloadHandler:
                  crawler=None,
                  aws_access_key_id=None, aws_secret_access_key=None,
                  aws_session_token=None,
-                 httpdownloadhandler=HTTPDownloadHandler, **kw):
-        if kw:
+                 httpdownloadhandler=HTTPDownloadHandler, anon=None):
+        if anon is not None:
             warnings.warn(
-                "'kw' args are deprecated. If you're trying to set s3 request anonymous."
-                " Use setting 'DOWNLOAD_HANDLERS_S3_ANONYMOUS = True' instead.",
+                "Argument 'anon' is deprecated. Use setting"
+                " 'DOWNLOAD_HANDLERS_S3_ANONYMOUS = True' instead.",
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
@@ -31,11 +31,9 @@ class S3DownloadHandler:
         if not aws_session_token:
             aws_session_token = settings['AWS_SESSION_TOKEN']
 
-        self.anon = kw.pop('anon', None)
+        self.anon = anon
         if self.anon is None:
             self.anon = settings.getbool('DOWNLOAD_HANDLERS_S3_ANONYMOUS')
-        if kw:
-            raise TypeError(f'Unexpected keyword arguments: {kw}')
 
         self._signer = None
         import botocore.auth
--- a/scrapy/core/downloader/handlers/s3.py
+++ b/scrapy/core/downloader/handlers/s3.py
@@ -22,24 +22,37 @@ class S3DownloadHandler:
         if not aws_session_token:
             aws_session_token = settings['AWS_SESSION_TOKEN']
 
-        # If no credentials could be found anywhere,
-        # consider this an anonymous connection request by default;
-        # unless 'anon' was set explicitly (True/False).
-        anon = kw.get('anon')
-        if anon is None and not aws_access_key_id and not aws_secret_access_key:
-            kw['anon'] = True
-        self.anon = kw.get('anon')
+        self.anon = kw.pop('anon', None)
+        if kw:
+            raise TypeError(f'Unexpected keyword arguments: {kw}')
 
         self._signer = None
         import botocore.auth
-        import botocore.credentials
-        kw.pop('anon', None)
-        if kw:
-            raise TypeError(f'Unexpected keyword arguments: {kw}')
+        from botocore.credentials import Credentials
+        from botocore.session import Session
+        from botocore.exceptions import NoCredentialsError
         if not self.anon:
-            SignerCls = botocore.auth.AUTH_TYPE_MAPS['s3']
-            self._signer = SignerCls(botocore.credentials.Credentials(
-                aws_access_key_id, aws_secret_access_key, aws_session_token))
+            # Follow credential creation logic in Session.create_client(),
+            # which is used in item pipeline and feed exporter.
+            if aws_access_key_id is not None and aws_secret_access_key is not None:
+                credentials = Credentials(
+                    access_key=aws_access_key_id,
+                    secret_key=aws_secret_access_key,
+                    token=aws_session_token,
+                )
+            else:
+                credentials = Session().get_credentials()
+            if credentials is not None:
+                SignerCls = botocore.auth.AUTH_TYPE_MAPS['s3']
+                self._signer = SignerCls(credentials)
+            else:
+                # If no credentials could be found anywhere,
+                # consider this an anonymous connection request by default;
+                # unless 'anon' was set explicitly (True/False).
+                if self.anon is None:
+                    self.anon = True
+                else:
+                    raise NoCredentialsError
 
         _http_handler = create_instance(
             objcls=httpdownloadhandler,
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -730,6 +730,24 @@ The default HTTPS handler uses HTTP/1.1. To use HTTP/2:
 .. _http2 faq: https://http2.github.io/faq/#does-http2-require-encryption
 .. _server pushes: https://tools.ietf.org/html/rfc7540#section-8.2
 
+.. setting:: DOWNLOAD_HANDLERS_S3_ANONYMOUS
+
+DOWNLOAD_HANDLERS_S3_ANONYMOUS
+------------------------------
+
+Default: ``False``
+
+Disable security credential in `S3DownloadHandler` explicitly and access `s3://`
+resources anonymously.
+
+`S3DownloadHandler` not only uses credential info from settings
+:setting:`AWS_ACCESS_KEY_ID`, :setting:`AWS_SECRET_ACCESS_KEY`,
+:setting:`AWS_SESSION_TOKEN`. Even these `AWS_*` settings are set as `None`, it
+still searches for credential from other `aws credential providers`_. This setting
+is introduced to disable credential creation in `S3DownloadHandler` completely.
+
+.. _aws credential providers: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials
+
 .. setting:: DOWNLOAD_TIMEOUT
 
 DOWNLOAD_TIMEOUT
--- a/tests/test_downloader_handlers.py
+++ b/tests/test_downloader_handlers.py
@@ -788,7 +788,9 @@ class S3AnonTestCase(unittest.TestCase):
 
     def setUp(self):
         skip_if_no_boto()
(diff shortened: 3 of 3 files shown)


scrapy/core/downloader/handlers/s3.py
```diff
@@ -24,21 +24,27 @@ def __init__(self, settings, *,
...
-            raise TypeError(f'Unexpected keyword arguments: {kw}')
-        if not self.anon:
-            SignerCls = botocore.auth.AUTH_TYPE_MAPS['s3']
-            self._signer = SignerCls(botocore.credentials.Credentials(
-                aws_access_key_id, aws_secret_access_key))
+        import botocore.session
+        if self.anon is None or self.anon is False:
+            session = botocore.session.get_session()
+            # Trigger credentials creation with Session.create_client().
+            session.create_client(
```
laggardkernel, line 37: Firstly I intended to only extract [credentials related code in `Session.create_client()`](https://github.com/boto/botocore/blob/c329fd23d3672b724a6dbd152f738de0b4d5a7ac/botocore/session.py#L825-L839). 

But using `Session.create_client()` here makes the code cleaner. Besides, considering we have already been using `Session.create_client()` in pipelines and feed exporter. Sticking to the same method in `S3DownloadHandler` maximize consistence across the project.

wRAR:
Please fix the test failures (and ideally run them locally before submitting the changes).

wRAR (approved):

scrapy/core/downloader/handlers/s3.py
```diff
@@ -24,21 +24,29 @@ def __init__(self, settings, *,
...
+        from botocore.session import Session
+        if self.anon is None or self.anon is False:
+            # Cause Session.create_client() is used in pipelines and feed exporter,
+            # following the credential creation logic in Session.create_client()
+            # maximize consistence of AWS auth across the project.
+            if aws_access_key_id is not None and aws_secret_access_key is not None:
+                credentials = Credentials(access_key=aws_access_key_id,
+                                          secret_key=aws_secret_access_key)
+            else:
+                credentials = Session().get_credentials()
```
Gallaecio, line 43: If this is replicating code somewhere else, any chance we can move shared code into `utils`?

scrapy/core/downloader/handlers/s3.py
```diff
@@ -24,21 +24,29 @@ def __init__(self, settings, *,
...
+            if credentials is not None:
+                SignerCls = botocore.auth.AUTH_TYPE_MAPS['s3']
+                self._signer = SignerCls(credentials)
+                self.anon = False
+            else:
+                self.anon = True
```
Gallaecio, line 49: This change of `self.anon` was not happening before. Is it needed?

scrapy/core/downloader/handlers/s3.py
```diff
@@ -24,21 +24,29 @@ def __init__(self, settings, *,
...
-        kw.pop('anon', None)
-        if kw:
-            raise TypeError(f'Unexpected keyword arguments: {kw}')
-        if not self.anon:
-            SignerCls = botocore.auth.AUTH_TYPE_MAPS['s3']
-            self._signer = SignerCls(botocore.credentials.Credentials(
-                aws_access_key_id, aws_secret_access_key))
+        from botocore.credentials import Credentials
+        from botocore.session import Session
+        if self.anon is None or self.anon is False:
```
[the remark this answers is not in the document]
laggardkernel: Thanks. I remember I was trying to emphasize the condition is `self.anon` not set or set explicitly to `False`. Forgot to merge them.

scrapy/core/downloader/handlers/s3.py
laggardkernel: The "somewhere else" is in `botocore`'s `Session.create_client()` (only the creadential creation part), not in Scrapy. Considering the snippet is used only once, there's no need to abstract it as a `utils` function.

scrapy/core/downloader/handlers/s3.py
laggardkernel: `self.anon = True` happens before this PR, but only by checking `aws_access_key_id is not None and aws_secret_access_key is not None`.

The `self.anon = False` here may not be needed. I'm just emphasizing it's set `False` explicitly, to differentiate the `None` value when `anon` param is not passed during `S3DownloadHandler(..., anon=True/False)` init.

scrapy/core/downloader/handlers/s3.py
Gallaecio: Could we then remove the comment? It seems more like a comment for code review than a comment for future reference.

scrapy/core/downloader/handlers/s3.py
Gallaecio: > `self.anon = True` happens before this PR

Sorry, I had not realize that.

I’m trying to understand the change of behavior compared to the previous implementation, because we will want to at least cover it in the release notes:

- `self.anon` can now become `True` even if the credential parameters and settings are falsy, as long as the credentials can be automatically determined from the environment (by [botocore’s credential resolver](https://github.com/boto/botocore/blob/0bf25028b9fa66f030a72daf8499128394e45ba7/botocore/credentials.py#L56)).
  
  While I think this change of behavior is for the better, it does mean that restoring the previous behavior from settings alone is now impossible. That is, before you could force anonymous access from settings by letting them have falsy values; now that is not possible. I think maybe we should add a setting to allow users to force anonymous credentials, just in case.

- `self.anon` now becomes `False` if credentials are not provided and cannot be found.

  It is not a big change, but since it’s not needed functionality-wise and is a change in behavior that could affect user-defined subclasses, I think it would be best to avoid it and leave `self.anon` as `None` if that was its value.

laggardkernel: Thought for a while. The logic before this PR onlys set `.anon` `True` if it's `None`. In my implementation, the valued could be changed from `False` to `True. It does breaks the anonymous function. I'll try to fix.

laggardkernel: > That is, before you could force anonymous access from settings by letting them have **falsy** values; now that is not possible. I think maybe we should add a setting to allow users to force anonymous credentials, just in case.

You misunderstood the logic here. Anonymous is declared with `True`, not falsy values.

If a user declares anonymous, the s3 request will be ensured anonymous, in both previous and current implementation. The PR here doesn't break this.

---

The problem is, what if a user declares non-anonymous with `.anon = False` explcitly and can't get a credential? We haven't take this branch into consideration before this PR.

Keyword `anon` was firstly introduced in 1b4fd3a by nyov. The reason behind it is still written in the comment.

```
        # If no credentials could be found anywhere,
        # consider this an anonymous connection request by default;
        # unless 'anon' was set explicitly (True/False).
```

I broke the "unless 'anon' was set explicitly (True/False)" part. We should change the code to raise an exc if no credential could be found but `.anon == False`.

laggardkernel: > That is, before you could force anonymous access from settings by letting them have **falsy** values; now that is not possible. I think maybe we should add a setting to allow users to force anonymous credentials, just in case.

You misunderstood the logic here. Anonymous is declared with `True`, not falsy values.

If a user declares anonymous, the s3 request will be ensured anonymous, in both previous and current implementation. The PR here doesn't break this.

---

The problem is, what if a user declares non-anonymous with `.anon = False` explcitly and can't get a credential? We haven't take this branch into consideration before this PR.

Keyword `anon` was firstly introduced in 1b4fd3a by nyov. The reason behind it is still written in the comment.

```
        # If no credentials could be found anywhere,
        # consider this an anonymous connection request by default;
        # unless 'anon' was set explicitly (True/False).
```

I broke the "unless 'anon' was set explicitly (True/False)" part. We should change the code to raise an exc if no credential could be found but `.anon == False`.

scrapy/core/downloader/handlers/s3.py
laggardkernel: The comment explains why there's two branches used to create a credential. I prefer to keep it to make the code more understandable.

scrapy/core/downloader/handlers/s3.py
Gallaecio: > We should change the code to raise an exc if no credential could be found but `.anon == False`.

That makes a lot of sense, and even though raising an exception here is also a backward-incompatible change, I also believe this one is for the better as well.

> > That is, before you could force anonymous access from settings by letting them have **falsy** values; now that is not possible. I think maybe we should add a setting to allow users to force anonymous credentials, just in case.
> 
> You misunderstood the logic here. Anonymous is declared with `True`, not falsy values.

I think I actually failed to properly explain my point. I meant falsy values for the `aws_access_key_id` and `aws_secret_access_key` parameters and their corresponding settings, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.

We need to keep in mind that users usually interact with this class through settings, not through `__init__` parameters. So forget about `__init__` parameters, and think of user interactions with this class solely from settings.

Before, they could get `self.anon` to be set to `True` by simply not using any of those 2 settings:

```python
        if not aws_access_key_id:
            aws_access_key_id = settings['AWS_ACCESS_KEY_ID']
        if not aws_secret_access_key:
            aws_secret_access_key = settings['AWS_SECRET_ACCESS_KEY']
        # …
        if anon is None and not aws_access_key_id and not aws_secret_access_key:
            kw['anon'] = True
```

Now, in an environment where there _are_ credentials, there’s no way to get `self.anon` to be set to `True` from settings alone. Even if those two settings are undefined, if the environment has credentials, `self.anon` will never be `True`, environment credentials will always be used.

That’s why I think we might want to expose that as a setting now, so that users that need to prevent environment credentials to be picked can do so through settings. Maybe this could be enough:

```python
self.anon = kw.pop('anon', settings['AWS_ANONYMOUS_ACCESS'])
```

Although if the `anon` parameter is `None`, maybe it should get overridden by the setting, so maybe this would be best:

```python
self.anon = kw.pop('anon', None)
if self.anon is None:
    self.anon = settings['AWS_ANONYMOUS_ACCESS']
```

scrapy/core/downloader/handlers/s3.py
Gallaecio: I see the comment is about https://github.com/boto/botocore/blob/develop/botocore/session.py#L827-L839. Why don’t we use `Session.create_client()` here directly, as we do in the other parts references in the code comment, instead of replicating part of its implementation manually?

scrapy/core/downloader/handlers/s3.py
Gallaecio: > But `AWS_ANONYMOUS_ACCESS` sounds like we're enabling anonymous for all aws related things. Could we define it more specific, e.g. `DOWNLOAD_HANDLERS_S3_ANONYMOUSE`?

To follow a similar pattern to those used in Scrapy to solve similar situations, what we can do is, in addition to defining the `AWS_ANONYMOUS_ACCESS` setting (which may be used to get that behavior as a default for anything using S3 in Scrapy), allowing to pass S3 credential settings on a per-request basis through `Request.meta`. The download handler can then use the meta parameters to override the global parameters when processing a request.

> I propose to remove `kw` argument (deprecate it first?)

Given that the pre-existing code already raises an exception if it has anything other than `anon`, I think we can replace it with an `anon` keyword-only parameter. It will raise a different exception, but I imagine that’s fine.

As for deprecating first, I was going to say that it may not be necessary in this case, but it’s probably best to be extra careful and do that, log a deprecation error before raising `ValueError` for now, and eventually remove the `**kw` parameter for good.

laggardkernel: > which may be used to get that behavior as a default for anything using S3 in Scrapy

aws s3 is used in S3DownloadHandler for downloading, in pipeline and feed exporter for uploading. Setting up s3 without any credential for downloading means the resource is publicly accessible/readable. But for uploading, I can't imagine s3 bucket is setup publicly writable for all users. It's a security fault.

"Instance metadata service" is an alternative for this publicly writable need. It provides a credential (without any settings) but only could be got from specific Amazon EC2 instance. 


> allowing to pass S3 credential settings on a per-request basis through `Request.meta`

This is another feature, we'd better open a separate PR for it.

> Given that the pre-existing code already raises an exception if it has anything other than anon, I think we can replace it with an anon keyword-only parameter.

Thanks. It makes sense.

Gallaecio: > Still, do we have an application scenario for this feature: disable credential usage for a specific s3 request? The purpose of the intro of `anon` is to skip reading credential info those `~/.aws/credentials`, `~/.aws/config` files.

I think it’s possible for a spider to aim to download input data from a private S3 storage (e.g. source URLs), but force anonymous access to public S3 URLs that are crawled. I’m not sure if publicly readable  S3 storages fail on unexpected authentication, but we definitely do not want to leak our S3 credentials to untrusted S3 servers. Specially since S3-compatible services other than Amazon’s exist.

So, I do see a scenario where per-request customization of credentials makes sense. That said, it is not part of the scope of this pull request, to get this merged we only need to make it possible to make S3 downloads anonymous.

Since there is a clear download/upload distinction, as you mentioned, I think your idea may be best to keep the changes minimal: implement just a setting but name it so that it’s clear that it does not affect other S3 credential usages. Seeing other Scrapy settings that alter download handlers, I suggest a name like `DOWNLOADER_S3_ANONYMOUS`.

Gallaecio:
Note: Please, avoid force-pushing. Pushing only new commits instead makes it easier to review code, since we only need to review the new changes; if you force-push, we need to review the whole thing again. If you are worried about keeping the history clean, don’t; we can squash your commits before merging easily.

laggardkernel:
Sorry about the force pushing. I'll give a brief explanation about what has been changed.

The 1st force pushing is to keep the history clean. `raise NoCredentialsError` if `anon` is not set explicitly and no credential could be found.  I squashed the change into former commit. 

In 2nd force pushing, I rebased the code to the HEAD of master. After my another PR GH-5210 being merged, I changed the credential code to support `aws_session_token`, this was done by rebasing and squashing the change into the 1st commit.
Besides, I added a setting to replace `anon` argument, and obsolete `kw` arg.

PR #5218, opened 2021-07-25
Response.attributes

### LINKED ISSUE #1877: Allow to add custom metadata to Response

elacuesta:
Closes #1877.

This is a subset of the changes from #5131, as [suggested](https://github.com/scrapy/scrapy/pull/5131#issuecomment-879711114) by @Gallaecio

### DIFF
--- a/tests/test_http_response.py
+++ b/tests/test_http_response.py
@@ -848,6 +848,7 @@ class CustomResponseTest(TextResponseTest):
     def test_replace(self):
         super().test_replace()
         r1 = self.response_class(url="https://example.org", status=200, foo="foo", bar="bar", lost="lost")
+
         r2 = r1.replace(foo="new-foo", bar="new-bar", lost="new-lost")
         self.assertIsInstance(r2, self.response_class)
         self.assertEqual(r1.foo, "foo")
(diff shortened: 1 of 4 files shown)


tests/test_http_response.py
```diff
@@ -820,3 +820,42 @@ def test_selector_shortcuts_kwargs(self):
...
+        self.assertIsInstance(r2, self.response_class)
+        self.assertEqual(r1.foo, r2.foo)
+        self.assertEqual(r1.bar, r2.bar)
+        self.assertEqual(r1.lost, "lost")
+        self.assertIsNone(r2.lost)
+
+    def test_replace(self):
+        super().test_replace()
+        r1 = self.response_class(url="https://example.org", status=200, foo="foo", bar="bar", lost="lost")
+        r2 = r1.replace(foo="new-foo", bar="new-bar", lost="new-lost")
```
Gallaecio, line 851: This tests that overwriting `lost` works. Shouldn’t we also test the same as with `copy`, that if you don’t pass `lost` to `replace` the old value gets lots? And in the same test, test that _not_ passing e.g. `foo` ensures the new object has the same value as the old object.

wRAR (approved):

tests/test_http_response.py
elacuesta: Added more tests, thanks.

Gallaecio (approved):

Issue #5212, opened 2021-07-20
Test failures on 3.10


wRAR:
I ran the test suite on 3.10b4, we have three test failures. One is easily fixed by replacing `collections` with `collections.abc` in `scrapy/utils/signal.py` (`collections.Sequence` no longer exists in 3.10 and is apparently deprecated before that), other two are less trivial: https://github.com/wRAR/scrapy/runs/3111526014?check_suite_focus=true

PR #5221, opened 2021-08-03
Fixing tests for upcoming 3.10.

wRAR:
Closes: #5212.

There is an elusive problem in `SpiderLoaderTest` which I can't reproduce when running tests only for that module. It's related to #5160 because that test checks the warning list and gets an unexpected one ("_SixMetaPathImporter.find_spec() not found;").

### DIFF
--- a/.github/workflows/tests-ubuntu.yml
+++ b/.github/workflows/tests-ubuntu.yml
@@ -45,6 +45,14 @@ jobs:
           env:
             TOXENV: asyncio
 
+        # 3.10-pre
+        - python-version: "3.10.0-beta.4"
+          env:
+            TOXENV: py
+        - python-version: "3.10.0-beta.4"
+          env:
+            TOXENV: asyncio
+
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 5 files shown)


wRAR:
The remaining warning ("_SixMetaPathImporter.find_spec() not found; falling back to find_module()") looks like a valid `six` problem already fixed in its code and we don't even use it directly, so I don't think we should (or can) do anything to remove this warning. This takes us back to the generic problem of "we expect a warning but get a different one first" tackled in #5160.

Gallaecio:
xfail?

wRAR:
Actually if we go as far as adding such checks we could just edit the test to skip this six warning.

wRAR:
The most recent six version actually contains the fix for this, but six is vendored in several other libraries, and botocore contains an older version, and importing `botocore.vendored.six` before running this test makes it fail (not sure if this piece of data is useful though).

Gallaecio:
Since it is not a direct dependency, your suggestion of [ignoring warnings from six in our test runs](https://docs.pytest.org/en/6.2.x/warnings.html) makes sense to me.

wRAR:
Not sure how to express that though, do you mean warnings mentioning `_SixMetaPathImporter`? Also it's technically a warning from Python importlib.

wRAR:
category: `ImportWartning`, message: '_SixMetaPathImporter.find_spec() not found; falling back to find_module()'

Gallaecio (approved):

PR #5223, opened 2021-08-05
Fix graceful stop for Downloader.close()

laggardkernel:
Former `Downloader.stop()` contradicted the design of **graceful stop**. It force closed the `DelayedCall` of `Downloader._process_queue()`. It may cause theses delayed requests still held on `Downloader` when the process stops. And all held requests will be lost. We should wait all requests to be downloaded.

Now `Downloader.close()` returns a `Deferred` and the `Deferred` is fired only after all requests done in downloader. Similar strategies are seen in `CrawlerRunner.stop()`, `ExecutionEngine.close_spider()`, `Scraper.close_spider()`.

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -26,6 +26,7 @@ class Slot:
         self.transferring = set()
         self.lastseen = 0
         self.latercall = None
+        self.closing = None
 
     def free_transfer_slots(self):
         return self.concurrency - len(self.transferring)
@@ -36,8 +37,13 @@ class Slot:
         return self.delay
 
     def close(self):
-        if self.latercall and self.latercall.active():
-            self.latercall.cancel()
+        self.closing = defer.Deferred()
+        self._maybe_fire_closing()
+        return self.closing
+
+    def _maybe_fire_closing(self) -> None:
+        if self.closing is not None and not self.active:
+            self.closing.callback(None)
 
     def __repr__(self):
         cls_name = self.__class__.__name__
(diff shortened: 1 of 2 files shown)


Gallaecio:
Related: #5000

Gallaecio:
The code seems well thought-out.

Any ideas on how to cover the change in the tests? I am not sure there is a straightforward way to do it.

AdrianAtZyte:
<details>
<summary>Opus thinking</summary>

Read the PR and traced the current code. Short version: I'd close it — the problem it fixes is already handled a layer up, and has been since before the PR was opened.

The graceful drain lives in the engine, not the downloader. ExecutionEngine._Slot.inprogress holds every request handed to the downloader — added in _download() (scrapy/core/engine.py:492), removed in download_async()'s finally only after the download resolves (scrapy/core/engine.py:479-480). That includes requests still sitting in slot.queue behind a DOWNLOAD_DELAY, because the awaiting coroutine hasn't returned yet. And close_spider_async() awaits self._slot.close() (engine.py:620) before calling self.downloader.close() (engine.py:623).

So by the time Slot.close() cancels latercall, the queue is empty and the cancel drops nothing. The same ordering existed in 2021 (slot.close() then downloader.close() in the old close_spider), and #4749 — linked from the PR — opens by stating the drain-before-shutdown behavior as the existing default. The premise that "all held requests will be lost" doesn't hold.

The other two downloader.close() callers can't have anything queued either: the __init__ error path (engine.py:152) and close_async() when the engine never ran (engine.py:261).

</details>

PR #5224, opened 2021-08-07
Document JOBDIR option

### LINKED ISSUE #5173: JOBDIR option is not documented
On https://docs.scrapy.org/en/latest/topics/jobs.html it is explained how to use the JOBDIR setting, but the setting itself is not present on https://docs.scrapy.org/en/latest/topics/settings.html page. We should add it there.

aaron-tan:
Closes #5173

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -996,9 +996,8 @@ JOBDIR
 
 Default: ``''``
 
-A string indicating the directory for storing the required data to keep the state of a single job to enable persistence support. This directory must not be shared by different spiders or jobs/runs of the same spider.
-
-For more info on this setting, see :ref:`topics-jobs`
+A string indicating the directory for storing the state of a crawl when
+:ref:`pausing and resuming crawls <topics-jobs>`.
 
 .. setting:: LOG_ENABLED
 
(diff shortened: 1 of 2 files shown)


docs/topics/settings.rst
```diff
@@ -989,6 +989,17 @@ Default: ``{}``
...
+.. setting:: JOBDIR
+
+JOBDIR
+------
+
+Default: ``''``
+
+A string indicating the directory for storing the required data to keep the state of a single job to enable persistence support. This directory must not be shared by different spiders or jobs/runs of the same spider.
+
+For more info on this setting, see :ref:`topics-jobs`
```
Gallaecio, line 1001: Thanks!

What would you think about shortening it a bit?

```suggestion
A string indicating the directory for storing the state of a crawl when
:ref:`pausing and resuming crawls <topics-jobs>`.
```

Gallaecio (approved):

docs/topics/settings.rst
aaron-tan: I think it's much better, so it doesn't sound like we're repeating ourselves like a parrot.

wRAR (approved):

Issue #5226, opened 2021-08-11
Document Reppy Python version support


Gallaecio:
The optional dependency on reppy for one of the built-in robots.txt parsers is [preventing us from running the extra-dependencies CI job with Python 3.9+](https://github.com/seomoz/reppy/issues/122). https://github.com/seomoz/reppy has not have a commit for ~1.5 years.

So I think we should deprecate the component.

If we don’t, we should document this limitation, and schedule a deprecation for 1 year before Python 3.8 reaches end of life, ~~i.e. in 9 months~~, because once we drop Python 3.8 support we will be forced to remove this component anyway, so giving a deprecation warning 1 year before is probably in the best interest of any user of the component.

wRAR:
I would be fine with not running tests for it and documenting that it's "not really supported" but I don't think we do this? Otherwise I'm fine with deprecating and then removing it.

For the context, it was first released in 1.8, together with all other "new" robots.txt parsers as a GSoC 2019 contribution, and while it was requested in the initial GSoC issue #3656, it was attempted much earlier: #949

So I think as long as we have other supported parsers it's fine to remove this one.

wRAR:
The parser comparison in our docs: https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.robotstxt

The performance comparison (linked from where): https://anubhavp28.github.io/gsoc-weekly-checkin-12/

So the main feature of Reppy is being much faster (which isn't really important for spiders that scrape 0 or 1 robots.txt file per run but may be important e.g. for broad crawls).

Cyb-Umair:
Can I work on this issue @Gallaecio ?

umairnsr87:
Can I work on this issue @Gallaecio ?

Gallaecio:
@umairnsr87 Yes, please! We need to update the documentation about Reppy, to clearly indicate that it only work with Python 3.8 and earlier.

Issue #122, opened 2012-04-21
Unicode Error when logging dropped item


domguard:
I have python objects in my scraped item that do have accents in their unicode representation, 
Through the debugger, I can see that the text fields containing UTF-8 accents are correctly ascii-encoded but  the object representation, like the category field here keeps its accents :

```
{'category': <Category: Electro-ménager :: TV>,
 'description': u' Ecran monochrome\r\nFonction \xe9co \r\nPr\xe9sentation du num\xe9ro \r\n1 combin\xe9\r\nAutonomie : 10 h en communication / 100 h en veille\r\nSonneries : 10\r\nVerrouillage du clavier \r\nR\xe9pertoire : 20 noms\r\nFonctionnalit\xe9 main-libres\r\n',
...
```

When using a simple pipeline to drop them and raising the DropItem exception to do so, I get :

```
2012-04-21 16:43:08+0200 [-] Unhandled Error
    Traceback (most recent call last):
      File "/Users/dom/django/.virtualenvs/cheze/lib/python2.7/site-packages/scrapy/utils/defer.py", line 57, in <genexpr>
        work = (callable(elem, *args, **named) for elem in iterable)
      File "/Users/dom/django/.virtualenvs/cheze/lib/python2.7/site-packages/scrapy/core/scraper.py", line 173, in _process_spidermw_output
        dfd.addBoth(self._itemproc_finished, output, response, spider)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/twisted/internet/defer.py", line 320, in addBoth
        callbackKeywords=kw, errbackKeywords=kw)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/twisted/internet/defer.py", line 286, in addCallbacks
        self._runCallbacks()
    --- <exception caught here> ---
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/twisted/internet/defer.py", line 542, in _runCallbacks
        current.result = callback(current.result, *args, **kw)
      File "/Users/dom/django/.virtualenvs/cheze/lib/python2.7/site-packages/scrapy/core/scraper.py", line 199, in _itemproc_finished
        log.msg(log.formatter.dropped(item, ex, response, spider), \
      File "/Users/dom/django/.virtualenvs/cheze/lib/python2.7/site-packages/scrapy/logformatter.py", line 22, in dropped
[...]

PR #5231, opened 2021-08-16
updated documentation for python version for reppy

umairnsr87:
This PR fixes issue #5226

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1068,12 +1068,13 @@ Native implementation, provides better speed than Protego.
 
 In order to use this parser:
 
+.. warning:: Does not support Python 3.9+
+
 * Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``
 
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ReppyRobotParser``
 
-* Only works with python 3.8 and earlier
 
 .. _rerp-parser:
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1073,6 +1073,8 @@ In order to use this parser:
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ReppyRobotParser``
 
+* only works with python 3.8 and earlier
+
 .. _rerp-parser:
 
 Robotexclusionrulesparser
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1073,7 +1073,7 @@ In order to use this parser:
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ReppyRobotParser``
 
-* only works with python 3.8 and earlier
+* Only works with python 3.8 and earlier
 
 .. _rerp-parser:
 
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1070,9 +1070,8 @@ In order to use this parser:
 
 * Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``
 
-.. warning:: `Upstream issue #122
-  <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in
-  Python 3.9+.
+    .. warning:: `Upstream issue #122
+        <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in Python 3.9+.
 
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ReppyRobotParser``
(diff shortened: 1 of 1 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -1073,6 +1073,8 @@ In order to use this parser:
 * Set :setting:`ROBOTSTXT_PARSER` setting to
   ``scrapy.robotstxt.ReppyRobotParser``
 
+* Only works with python 3.8 and earlier
```
Gallaecio, line 1076: I suggest we put this at the beginning of the section, as a warning:

```rst
.. warning:: Does not support Python 3.9+
```

umairnsr87: @Gallaecio  did you mean just before the install?

umairnsr87: If we keep both of them then it will be more informative I guess.

docs/topics/downloader-middleware.rst
```diff
@@ -1068,11 +1068,14 @@ Native implementation, provides better speed than Protego.
 
 In order to use this parser:
 
+.. warning:: Does not support Python 3.9+
+
 * Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``
```
Gallaecio, line 1073: It’s not exactly what I had in mind, but I guess it makes sense to keep the warning close to the installation instructions.

In that case, what about nesting it within the installation step?

```suggestion
* Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``

  .. warning:: Does not support Python 3.9+
```

umairnsr87: seems good ....

umairnsr87:
seems good to me now @Gallaecio ....

Gallaecio (approved):

wRAR:
Should we link to the upstream issue and/or say that this may be false if the library is already fixed?

umairnsr87:
> Should we link to the upstream issue and/or say that this may be false if the library is already fixed?

Can you please explain @wRAR ?

wRAR:
I mean currently reppy doesn't work on 3.9 but it can be fixed in the future which will make out doc incorrect.

umairnsr87:
> I mean currently reppy doesn't work on 3.9 but it can be fixed in the future which will make our doc incorrect.

yes right... So what do you suggest @wRAR ?

umairnsr87:
> Should we link to the upstream issue and/or say that this may be false if the library is already fixed?

@Gallaecio and @wRAR should we do this in our docs ?

Gallaecio:
I agree with @wRAR. Maybe something something like this?

```rst
   .. warning:: `Upstream issue #122 
      <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in 
      Python 3.9+.
```

umairnsr87:
> I agree with @wRAR. Maybe something something like this?
> 
> ```
>    .. warning:: `Upstream issue #122 
>       <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in 
>       Python 3.9+.
> ```
Updated  @Gallaecio .Thanks ...!

docs/topics/downloader-middleware.rst
```diff
@@ -1070,9 +1070,14 @@ In order to use this parser:
 
 * Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``
 
+.. warning:: `Upstream issue #122
+  <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in
+  Python 3.9+.
```
Gallaecio, line 1075: The warning should be nested into the list item, I think:

```suggestion
* Install `Reppy <https://github.com/seomoz/reppy/>`_ by running ``pip install reppy``

  .. warning:: `Upstream issue #122
     <https://github.com/seomoz/reppy/issues/122>`_ prevents reppy usage in
     Python 3.9+.
```

umairnsr87: @Gallaecio Thanks for the review.Actually I have seen that in the whole documentation where were are defining the functionalities of the component the blocks were nested there(which will make warning block appear in the middle of the page horizontally).
for example:
-Functionality 1
    - Details of functionality 
    - some block(which will appear in the middle)

But in our scenario here we an unordered list.
* some reference
* some reference 2
* some reference 3

PS: I can make it into a nested structure if you want that or I can post the screenshot if you say and let you decide which one is better.

umairnsr87: link for without nesting of warning:
[https://ibb.co/hRq7jwL](url)

link for nesting of warning:
[https://ibb.co/khz8VX7](url)

umairnsr87: @Gallaecio  please review it and let me know. Thanks alot.

Gallaecio: I prefer it either nested to the installation step (because it’s a warning about an installation detail), or as I originally suggested: non-nested at the beginning of the whole reppy section, right under the header (since users may not want to even consider reppy knowing this issue).

umairnsr87: yes right. Great. Thanks.

umairnsr87: Done.

Gallaecio (approved):

wRAR (approved):

PR #5235, opened 2021-08-23
Fix/silence the Pylint messages added in 2.10.

wRAR:
Fix:

- use-dict-literal
- use-list-literal
- redundant-u-string-prefix
- use-sequence-for-iteration

Ignore:

- unspecified-encoding


scrapy/utils/deprecate.py
```diff
@@ -79,7 +79,7 @@ def __init__(cls, name, bases, clsdict_):
         # for implementation details
         def __instancecheck__(cls, inst):
             return any(cls.__subclasscheck__(c)
-                       for c in {type(inst), inst.__class__})
+                       for c in [type(inst), inst.__class__])
```
Gallaecio, line 82: Maybe a tuple would make more sense?

Gallaecio (approved):

PR #5239, opened 2021-09-13
Improve documentation for spider middlewares

maanijou:
Fixes ambiguity of return type for `process_spider_exception` function.

### DIFF
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -122,8 +122,8 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
         method (from a previous spider middleware) raises an exception.
 
         :meth:`process_spider_exception` should return either ``None`` or an
-        iterable of :class:`~scrapy.Request` or :ref:`item objects
-        <topics-items>`.
+        iterable of :class:`~scrapy.Request` or :ref:`item <topics-items>` 
+        objects.
 
         If it returns ``None``, Scrapy will continue processing this exception,
         executing any other :meth:`process_spider_exception` in the following
--- a/docs/topics/spider-middleware.rst
+++ b/docs/topics/spider-middleware.rst
@@ -122,7 +122,7 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
         method (from a previous spider middleware) raises an exception.
 
         :meth:`process_spider_exception` should return either ``None`` or an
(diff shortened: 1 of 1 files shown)


docs/topics/spider-middleware.rst
```diff
@@ -122,7 +122,7 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
         method (from a previous spider middleware) raises an exception.
 
         :meth:`process_spider_exception` should return either ``None`` or an
-        iterable of :class:`~scrapy.Request` objects and :ref:`item object
+        iterable of :class:`~scrapy.Request` objects or :ref:`item object
```
Gallaecio, line 125: I think this may actually make the sentence more misleading.

With “or”, you could interpret that the iterable must contain either all requests or all items, when it can actually combination of the two.

What’s exactly the potential misinterpretation from the original sentence that you are trying to clarify here? I wonder if switching “item object” to “item object**s**” would fix it.

The point in the sentence is that each object in the iterable may be a Request object or an item object.

maanijou: I think using `and` could be misleading here. Also, the documentation is not consistent with the middleware template file in [middlewares.py.tmpl](https://github.com/scrapy/scrapy/blob/8284de5e7613c47e69d40d4f6a2a1dc846b50dd6/scrapy/templates/project/module/middlewares.py.tmpl#L43): 

```Should return either None or an iterable of Request or item objects.```

I will also change “item object” to “item objects,” as you suggested.

Comment (Gallaecio) [reply docs/topics/spider-middleware.rst]: I think in the template file things are more clear thanks to the word “objects” being there only at the end. What about doing the same in the documentation?

File: docs/topics/spider-middleware.rst [hunk 2]
```diff
@@ -122,7 +122,7 @@ object gives you access, for example, to the :ref:`settings <topics-settings>`.
         method (from a previous spider middleware) raises an exception.
 
         :meth:`process_spider_exception` should return either ``None`` or an
-        iterable of :class:`~scrapy.Request` objects and :ref:`item object
+        iterable of :class:`~scrapy.Request` or :ref:`item objects
         <topics-items>`.
```
Gallaecio, line 126:
```suggestion
        iterable of :class:`~scrapy.Request` or :ref:`item <topics-items>` 
        objects.
```

Gallaecio (approved):

wRAR (approved):

PR #5243, opened 2021-09-21
added decoding of bytes for non text responses

### LINKED ISSUE #5145: ItemLoader: support non-TextResponse
At the moment, `ItemLoader(response=response)` fails if `response` is not a `TextResponse` instance.

Passing a binary response can still be useful, though. For example, to allow processors to access the response from their loader context, and hence be able to report the source URL (`response.url`) when reporting input issues.

arumugamramaswamy:
Fixes #5145 

In this pull request, I have implemented a potential fix for the issue **ItemLoader: support non-TextResponse**. I have added support for non-TextResponse by trying to decode the body of the response using `latin-1` encoding.

Please provide feedback so that I can resolve this issue

### DIFF
--- a/scrapy/selector/unified.py
+++ b/scrapy/selector/unified.py
@@ -75,7 +75,11 @@ class Selector(_ParselSelector, object_ref):
             response = _response_from_text(text, st)
 
         if response is not None:
-            text = response.text
+            try:
+                text = response.text
+            except AttributeError:
+                if isinstance(response.body, bytes):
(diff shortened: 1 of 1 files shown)


wRAR:
I don't think that's what the issue is about: it's not about converting binary responses to text ones (which should never be needed in Scrapy unless something is broken) but about creating loaders that don't try to parse the response.

By the way, assuming latin1 is a very bad idea, doubly so in web things.

Issue #5245, opened 2021-09-23
Complete conversion to f-strings


wRAR:
Recent pylint started detecting potential f-string usages, I think it's fine to keep this error and just convert the remaining usages. pylint output can be seen at https://github.com/scrapy/scrapy/pull/5244/checks?check_run_id=3671950402

GeorgeA92:
I received the same log entries from `pylint` after recent update of my PR 4792 https://github.com/scrapy/scrapy/pull/4972/checks?check_run_id=3633586376.
I can take it.

tpush:
I can fix this, can have it assigned.

wRAR:
@tpush as you can see above there is already an open PR for this.

PR #5246, opened 2021-09-23
5242 f-strings

GeorgeA92:
Related to https://github.com/scrapy/scrapy/issues/5245
Replaced string format by usage if f-strings in places mentioned in recent `pylint` checks.

PR #5247, opened 2021-09-24
fix: Update to f-string

NeelSagar39:
An update for the issue #5245


Gallaecio:
Closing, since there is already https://github.com/scrapy/scrapy/pull/5246 and it is much closer to completion.

PR #5255, opened 2021-10-03
Using Logo Scrapy in Readme.md

raihan71:
Logo scrapy used in readme.md made looks nicer

<strong>Before</strong>
image

<strong>After</strong>
image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,6 @@
+.. image:: /artwork/scrapy-logo.jpg
+   :width: 400px
+   
 ======
 Scrapy
 ======


raihan71:
my first pr wdyt? 🤝 @Gallaecio

Gallaecio (approved):

wRAR (approved):

README.rst
```diff
@@ -1,3 +1,6 @@
+.. image:: /artwork/scrapy-logo.jpg
```
kmike, line 1: hey! Readme file is used on pypi as well (https://github.com/scrapy/scrapy/blob/cbf934955223c29f0632fbba94d9586849a3b744/setup.py#L66) - do you know if relative link works there?

Gallaecio: Right :facepalm: 

We could use `https://raw.githubusercontent.com/scrapy/scrapy/master/artwork/scrapy-logo.jpg`.

We may also want to consider having `setup.py` edit the URL before passing it to `setup()` so that it points to the tag of the corresponding version instead of `master`, so that if we even change the logo old versions still use the old logo. Although that’s probably not very important.

raihan71: omg, my bad! so do I need to fix ```setup.py``` ?

Gallaecio: Not your bad, really, I should have realized this could be an issue.

Ideally yes, we should fix `setup.py`. A string replacement of `/artwork/scrapy-logo.jpg` to an URL like the one I mentioned with the version instead of `master` would be ideal, I think, e.g. `https://raw.githubusercontent.com/scrapy/scrapy/2.5.0/artwork/scrapy-logo.jpg`.

raihan71: Ok rather than using path I'll change into cdn githubcontent

raihan71: what if I change to this logo? @Gallaecio refer to #5258
image

PR #5256, opened 2021-10-03
docs: restructed phrasing for clarity

rydwhelchel:


Gallaecio:
Could you clarify which aspects of the original phrasing you found to be unclear, and how the new phrasing addresses that?

I think there is room for improvement in the paragraph, as in many others, but I don’t see much different between the original paragraph and the new one, other than the new ones seems longer.

rydwhelchel:
Sure, when I first read the paragraph that I changed, it wasn't clear to me what the "However" was referring to. I wasn't sure if it was saying that `.get()` was better than `.getall()` or that it was even related to accessing the index directly on the selector list. 

Perhaps it was just an oversight as I was reading, but I feel that the changes make it a little more clear directly in the text why you'd prefer `.get()`. I think the original was worded more succinctly, but this version is more clear.

docs/intro/tutorial.rst
```diff
@@ -277,9 +277,9 @@ As an alternative, you could've written:
 >>> response.css('title::text')[0].get()
 'Quotes to Scrape'
 
-However, using ``.get()`` directly on a :class:`~scrapy.selector.SelectorList`
-instance avoids an ``IndexError`` and returns ``None`` when it doesn't
-find any element matching the selection.
+Directly accessing an index on a :class:`~scrapy.selector.SelectorList`
+instance could potentially run into an ``IndexError``. It is recommended to use
+``.get()`` directly instead as it avoids such index errors.
```
Gallaecio, line 282: I see your point. What about this?

```suggestion
Accessing an index on a :class:`~scrapy.selector.SelectorList` instance will 
raise an :exc:`IndexError` exception if there are no results::

    >>> response.css('noelement')[0].get()
    Traceback (most recent call last):
    File "<console>", line 1, in <module>
    ...
    IndexError: list index out of range

You might want to use ``.get()`` directly on the 
:class:`~scrapy.selector.SelectorList` instead, which returns ``None`` if there 
are no results::

>>> response.css("noelement").get()
```

rydwhelchel: Looks perfect to me! Definitely clears up any issue I had with it.

PR #5251, opened 2021-10-01
Allow 'callback' key in keyword arguments for request callbacks

elacuesta:
Fixes #5237

Use `Deferred.addCallbacks` instead of `Deferred.addCallback`. Modified the test case to work with this patch but fail with the current code in master.

PR #5264, opened 2021-10-06
[CI] fix pypy test

elacuesta:
Fix a [test failure](https://github.com/scrapy/scrapy/pull/5251/checks?check_run_id=3763207639) inadvertently introduced (by me 😅) in #5251


Gallaecio:
Wow. I’ve seen it happen and just assumed it was one of those random ones :facepalm:. Good catch.

PR #5266, opened 2021-10-06
Fixed documentation example

MarvinPetzoldt:
I think here the `name` parameter should be compared with the `'price'`.


Gallaecio (approved):
Nice catch, thank you!

PR #5269, opened 2021-10-09
Fix bug #5145 - Removing Selector for Response's that are not Http or Xml

### LINKED ISSUE #5145: ItemLoader: support non-TextResponse

Ankur19:
Fixes #5145

* set **Selector** to None when the response is of type base-response **Response**

Please provide feedback


Gallaecio:
It would be great to also get a test to prove the change works.

PR #5277, opened 2021-10-14
Limit max parallel tasks to improve performance

### LINKED ISSUE #5182: Extreme performance of the waste when CONCURRENT_ITEMS is large
### Description

There is  Extreme performance of the waste When CONCURRENT_ITEMS set to a ge large number, such as 9999.

Some days ago, I wrote a spider with CONCURRENT_ITEMS=9999, and run it. I found my spider use 100% cpu and crawl speed is very very slow （40 pages/mins）, normally it would be 400+pages/min.

### Steps to Reproduce

Here is code , you can run directly. I would see cpu up to 100%. Wait 1 minute, you will see 
[scrapy.extensions.logstats] INFO: Crawled 27 pages (at 27 pages/min)

```

import scrapy


    name = 'test'

    custom_settings = {
        "LOG_LEVEL": "INFO",
        "CONCURRENT_ITEMS": 99999,
    }

    def start_requests(self):
        for _ in range(3000):
            yield scrapy.Request(
                url="http://httpbin.org/get",
                dont_filter=True,
            )

    def parse(self, response):
        pass

```

vgerak:
Closes #5182.

Limits max tasks to `cpu_count() * 30`, which should be a "reasonable" high value (in local tests, best performance is around 10*cpu).

Allows overriding this limit by use of the `"OVERRIDE_CONCURRENCY_LIMIT"` setting.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -302,7 +302,9 @@ CONCURRENT_ITEMS
 Default: ``100``
 
 Maximum number of concurrent items (per response) to process in parallel in
-:ref:`item pipelines <topics-item-pipeline>`.
+:ref:`item pipelines <topics-item-pipeline>`. Setting value is capped to
+``"# of CPU cores" * 30``, to avoid performance degradation when a too high value
+is used. Limit can be removed by the :setting:`OVERRIDE_CONCURRENCY_LIMIT` option.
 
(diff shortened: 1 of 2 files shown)


vgerak:
Closing in favor of a proposed proper benchmarking approach

PR #5285, opened 2021-10-20
Remove usage of deprecated mktemp

darakian:
Small PR to update usage following guidance from the python docs
https://docs.python.org/3/library/tempfile.html#tempfile.mktemp
The mktemp function has been marked as deprecated for some time and is considered to be insecure.
There's also talk about removing the function https://bugs.python.org/issue36309
Though it's unclear if removal will ever happen.

The only real usage of the function is in `scrapy/downloadermiddlewares/decompression.py` though I've updated the tests as well for consistency


Gallaecio:
Note: https://github.com/scrapy/scrapy/pull/5285#discussion_r733015078 still applies.

darakian:
💄 applied

Gallaecio:
Address the issue as you best see fit.

I do recommend not to worry about keeping a clean commit list in a pull request, we can squash your commits into 1 upon merging instead, and force-pushing makes review of newer changes slightly harder in GitHub.

Gallaecio (approved):

tests/test_downloader_handlers.py
```diff
@@ -107,7 +107,7 @@ def test_lazy_handlers(self):
 class FileTestCase(unittest.TestCase):
 
     def setUp(self):
-        self.tmpname = self.mktemp()
+        fd, self.tmpname = mkstemp()
```
wRAR, line 110: This actually creates the file and it's not deleted later. The old code only creates the file with the `^` suffix and deletes it in `tearDown`. As far as I can see the test just needs a temporary file with `^` as a suffix, so we can just use `suffix='^'` as proposed in a now hidden comment (not with a context manager though, as we don't want it to be deleted right away).

darakian: Feel free to edit this PR if need be 👍

scrapy/downloadermiddlewares/decompression.py
```diff
@@ -30,10 +30,11 @@ def __init__(self):
 
     def _is_tar(self, response):
         archive = BytesIO(response.body)
-        try:
-            tar_file = tarfile.open(name=mktemp(), fileobj=archive)
-        except tarfile.ReadError:
-            return
+        with NamedTemporaryFile() as temporary_file:
+            try:
+                tar_file = tarfile.open(name=temporary_file.name, fileobj=archive)
```
Gallaecio, line 35: [According to the documentation](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile), the `name` parameter is optional. [We added it 13 years ago](https://github.com/scrapy/scrapy/commit/9022f84c05a13d58a03f24cc65439ccc109ee57b) because “None is failing in some cases”. But I have no idea what those cases would be. It may be that the Python version back then had a bug or something.

I am thinking that we could remove the parameter altogether. If anyone can provide a scenario where it fails, we can try to address that scenario in a point release of Scrapy, this time documenting the issue and adding a test for it, which would be best in the long run.

tests/test_downloader_handlers.py
wRAR: #5736 

So using `suffix='^'` indeed should work.

scrapy/downloadermiddlewares/decompression.py
Gallaecio: This became irrelevant with the removal of the entire module.

tests/test_downloader_handlers.py
Gallaecio: Done as part of the [merge with the main branch](https://github.com/scrapy/scrapy/pull/5285/commits/8cf1f9fb106a0b3cdee579d9b661035531ea7195).

PR #5290, opened 2021-10-23
Place Content-encoding header into response's flags is desired

T0shik:
Fixes #1988 

Continuation of #4025 due to abandonment.

additional changes:
- `'tests/ignores.txt'` in `conftest.py` to rely on `Path`
- tests to use `unittest` assertions
- remove `create_spider_mw` and make config test specific (can change this if needed)

Resolves #4025


T0shik:
I couldn't figure out the previous build error 
https://github.com/scrapy/scrapy/runs/4026144087?check_suite_focus=true
```
docs/conftest.py:24: in <module>
    CodeBlockParser(future_imports=['print_function']),
E   TypeError: __init__() got an unexpected keyword argument 'future_imports'
```
I've amended my changes to touch less files, while in there I spotted the mutation on the original response content encoding array and decided to remove it.

I think if the following build fails in a similar fashion I'll need a hand

Gallaecio:
> I couldn't figure out the previous build error https://github.com/scrapy/scrapy/runs/4026144087?check_suite_focus=true
> 
> ```
> docs/conftest.py:24: in <module>
>     CodeBlockParser(future_imports=['print_function']),
> E   TypeError: __init__() got an unexpected keyword argument 'future_imports'
> ```

https://github.com/scrapy/scrapy/pull/5299

T0shik:
@Gallaecio thank you, will keep an eye out for when that's merged

Gallaecio:
Closing and reopening for a fresh CI run.

Gallaecio (commented):
@T0shik Thanks again for your work. I am very sorry it took me so long to get back to this.

@elacuesta 

> My concern is similar to [the one I had before](https://github.com/scrapy/scrapy/pull/4021#discussion_r324924066). In this case, the "decoded" flag does not actually say _which_ was the value of the original header. Since I wrote the above comment, #1877 was solved, so I think now an alternative would be to add a response metadata attribute to store this value (or at a minimum, store `content-encoding=<value>` as a flag).

I don’t think we need this, since after this change it is possible to access the original header.

> Furthermore, I see no reason for the flag to be `bytes`, [existing flags](https://github.com/scrapy/scrapy/blob/2.5.1/scrapy/core/downloader/handlers/http11.py#L571-L576) are all `str`.

While I agree, I think we should keep it as bytes for backward compatibility. We could add it as both, and deprecate somehow the `bytes` version (wrapping the set to detect access by key?), but I think that would be a non-trivial change and I think it may be best to keep it out of this pull request.

Issue #5298, opened 2021-10-27
Tests failing with new Sybil


wRAR:
Sybil 3.0.0 was just released, and at least the following change breaks our tests: `CodeBlockParser has been renamed to PythonCodeBlockParser`

```python-traceback
docs/conftest.py:24: in <module>
    CodeBlockParser(future_imports=['print_function']),
E   TypeError: __init__() got an unexpected keyword argument 'future_imports'
```

PR #5305, opened 2021-10-31
fix: pin most of the dependencies

azzamsa:
Fixes https://github.com/scrapy/scrapy/issues/5298

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-Sphinx>=3.0
+Sphinx==3.0.1
 sphinx-hoverxref>=0.2b1
 sphinx-notfound-page>=0.4
 sphinx-rtd-theme>=0.5.2


Gallaecio:
The pull request does not pin any dependency at the moment. Do you wish to have it merged as is, in which case you should probably update the title, or do you plan to extend the pull request with more commits, in which case you should mark it as a draft or prepend `[WIP] ` to the title?

azzamsa:
Oh, sorry. I use this PR just to show that some tests are failed. Probably due to upstream changes. Github does not allow a PR with no advanced commit.

azzamsa:
Why I don't find any macOS test being run?
As the [config)(https://github.com/scrapy/scrapy/blob/3225de725720bba246ba8c9845fe4b84bc0c82e7/.github/workflows/tests-macos.yml#L2) it should run.

cristianMuntean1 (changes requested):
@click.command()
@click.option("--count", default=1, help="Number of greetings.")
@click.option("--name", prompt="Your name", help="The person to greet.")
def hello(count, name):
    """Simple program that greets NAME for a total of COUNT times."""
    for _ in range(count):
        click.echo(f"Hello, {name}!")

if __name__ == '__main__':
    hello()

azzamsa:
I would like to close this pr temporary to avoid spamming, and reopen when it is ready for public.

Issue #4976, opened 2021-02-04
Fix and document asyncio reactor problems on Windows


wRAR:
As described in https://twistedmatrix.com/trac/ticket/9766 you cannot just enable AsyncioSelectorReactor on Windows with recent Python, you either need fixed Twisted (which is not released yet, the merged fix is https://github.com/twisted/twisted/pull/1338) or, supposedly, add some manual fix as documented [here](https://github.com/twisted/twisted/blob/09b96850c2ebcb635f448ed3f9bbf5f157be3693/src/twisted/internet/asyncioreactor.py#L35-L44). So if it's possible to add this code to Scrapy we should probably do that, at least until the next Twisted release, and even after it we should document that new enough Twisted is needed in this use case.

ivanprado:
This project https://github.com/ivanprado/scrapy-autoextract-demo can be used to reproduce the issue in Windows with Python 3.8. There is a workaround for the problem. Invoking:

```py
import asyncio
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```

fixes the problem (uncomment these lines to try it: https://github.com/ivanprado/scrapy-autoextract-demo/blob/main/scrapy_autoextract_demo/settings.py#L4-L5).

The workaround is suggested in Twisted master: https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/asyncioreactor.py#L35-L44

I don't think Twisted introduced a fix for the problem itself: even after a future Twisted release, it will be required to run the lines above to make it work. So I suggest the following options. Either:
* document the workaround in https://docs.scrapy.org/en/latest/topics/asyncio.html
* or implement some code in Scrapy that automatically invoke this lines on Windows + Python 3.8

Gallaecio:
Unless there is a good reason not to, I would try to fix this on our side.

LeMoussel:
Couldn't resolve with the fix @ivanprado mentioned.

wRAR:
@LeMoussel please provide a minimal reproducible example of this.

LeMoussel:
```
import asyncio
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

import scrapy
import playwright

class MinimalSpider(scrapy.Spider):
    name = "minimalspider"
    custom_settings = {
        'TWISTED_REACTOR': "twisted.internet.asyncioreactor.AsyncioSelectorReactor",
        'DOWNLOAD_HANDLERS': {
[...]

PR #5315, opened 2021-11-11
Set WindowsSelectorEventLoopPolicy on Windows.

wRAR:
Closes: #4976

Needs fixing tests that expect a default loop to be used, and adding docs, including manual setting for CrawlerProcess/Runner in https://docs.scrapy.org/en/latest/topics/asyncio.html#installing-the-asyncio-reactor

### DIFF
--- a/docs/topics/asyncio.rst
+++ b/docs/topics/asyncio.rst
@@ -39,15 +39,15 @@ use it instead of the default asyncio event loop.
 Windows-specific notes
 ======================
 
-The Windows implementation of `asyncio` can use two event loop implementation:
-:class:`~asyncio.SelectorEventLoop` (default before Python 3.8, required when
-using Twisted) and :class:`~asyncio.ProactorEventLoop` (default since Python
-3.8, cannot work with Twisted). So on Python 3.8+ the event loop class needs
-to be changed. Scrapy since VERSION does this automatically when you change the
-:setting:`TWISTED_REACTOR` setting or call
+The Windows implementation of :mod:`asyncio` can use two event loop
+implementations: :class:`~asyncio.SelectorEventLoop` (default before Python
+3.8, required when using Twisted) and :class:`~asyncio.ProactorEventLoop`
+(default since Python 3.8, cannot work with Twisted). So on Python 3.8+ the
+event loop class needs to be changed. Scrapy since VERSION does this
+automatically when you change the :setting:`TWISTED_REACTOR` setting or call
 :func:`~scrapy.utils.reactor.install_reactor`, but if you install the reactor
-by other means or use older Scrapy you need to call the following code before
-installing the reactor::
+by other means or use an older Scrapy version you need to call the following
+code before installing the reactor::
 
     import asyncio
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
--- a/docs/topics/asyncio.rst
+++ b/docs/topics/asyncio.rst
@@ -41,6 +41,34 @@ use it instead of the default asyncio event loop.
 
 .. _asyncio-await-dfd:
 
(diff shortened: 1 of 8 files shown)


wRAR:
Tests that hang on old Twisted because of https://twistedmatrix.com/trac/ticket/9390 also hang on Windows with these changes, I need to investigate the reason.

wRAR:
Looks like there are 100% reproducible unclean reactor problems in some tests specific to windows-3.10-asyncio.

scrapy/utils/reactor.py
```diff
@@ -57,6 +58,10 @@ def install_reactor(reactor_path, event_loop_path=None):
     reactor_class = load_object(reactor_path)
     if reactor_class is asyncioreactor.AsyncioSelectorReactor:
         with suppress(error.ReactorAlreadyInstalledError):
+            if sys.version_info >= (3, 8) and sys.platform == "win32":
+                policy = asyncio.get_event_loop_policy()
```
wRAR, line 62: Should we also load coverage data from Windows runs?

wRAR:
I've filed https://twistedmatrix.com/trac/ticket/10279 and skipped those tests.

docs/topics/asyncio.rst
```diff
@@ -41,6 +41,34 @@ use it instead of the default asyncio event loop.
...
+
+The Windows implementation of `asyncio` can use two event loop implementation:
+:class:`~asyncio.SelectorEventLoop` (default before Python 3.8, required when
+using Twisted) and :class:`~asyncio.ProactorEventLoop` (default since Python
+3.8, cannot work with Twisted). So on Python 3.8+ the event loop class needs
+to be changed. Scrapy since VERSION does this automatically when you change the
+:setting:`TWISTED_REACTOR` setting or call
+:func:`~scrapy.utils.reactor.install_reactor`, but if you install the reactor
+by other means or use older Scrapy you need to call the following code before
+installing the reactor::
```
Gallaecio, line 55: :lipstick: 

``` `asyncio` (italics) ``` → ``` ``asyncio`` (monospace) ``` (you could alternatively use ``` :mod:`asyncio` ```)
implementation → implementation**s**
use older Scrapy → use an older Scrapy version

```suggestion
The Windows implementation of ``asyncio`` can use two event loop 
implementations: :class:`~asyncio.SelectorEventLoop` (default before Python 
3.8, required when using Twisted) and :class:`~asyncio.ProactorEventLoop` 
(default since Python 3.8, cannot work with Twisted). So on Python 3.8+ the 
event loop class needs to be changed. Scrapy since VERSION does this 
automatically when you change the :setting:`TWISTED_REACTOR` setting or call
:func:`~scrapy.utils.reactor.install_reactor`, but if you install the reactor
by other means or use an older Scrapy version you need to call the following 
code before installing the reactor::
```

scrapy/utils/reactor.py
Gallaecio: Ideally, but I don’t think you need to worry about that as part of this pull request unless you want to. Maybe open a separate issue to address that eventually?

tests/test_downloader_handlers.py
```diff
@@ -287,6 +288,11 @@ def test_redirect_status_head(self):
 
     @defer.inlineCallbacks
     def test_timeout_download_from_spider_nodata_rcvd(self):
+        if self.reactor_pytest == "asyncio" and sys.platform == "win32":
+            raise unittest.SkipTest(
+                "This test produces DirtyReactorAggregateError on Windows with asyncio"
```
Gallaecio, line 293: What about including a link to the upstream issue you reported when skipping these tests, for future reference? (either as a code comment or in the skip message).

Gallaecio (approved):

Issue #5314, opened 2021-11-11
Increase the usage of augmented assignment statements


elfring:
:eyes: Some source code analysis tools can help to find opportunities for improving software components.
:thought_balloon: I propose to [increase the usage of augmented assignment statements](https://docs.python.org/3/reference/simple_stmts.html#augmented-assignment-statements "Augmented assignment statements") accordingly.

```diff
diff --git a/scrapy/core/downloader/handlers/http11.py b/scrapy/core/downloader/handlers/http11.py
index 8a91d4c5..38935667 100644
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -213,7 +213,7 @@ class TunnelingAgent(Agent):
         # proxy host and port are required for HTTP pool `key`
         # otherwise, same remote host connection request could reuse
         # a cached tunneled connection to a different proxy
-        key = key + self._proxyConf
+        key += self._proxyConf
         return super()._requestWithEndpoint(
             key=key,
             endpoint=endpoint,
diff --git a/scrapy/core/http2/stream.py b/scrapy/core/http2/stream.py
index c2a4b702..5c393c02 100644
--- a/scrapy/core/http2/stream.py
+++ b/scrapy/core/http2/stream.py
@@ -285,8 +285,8 @@ class Stream:
 
             self._protocol.conn.send_data(self.stream_id, data_chunk, end_stream=False)
 
-            bytes_to_send_size = bytes_to_send_size - chunk_size
-            self.metadata['remaining_content_length'] = self.metadata['remaining_content_length'] - chunk_size
+            bytes_to_send_size -= chunk_size
+            self.metadata['remaining_content_length'] -= chunk_size
 
         self.metadata['remaining_content_length'] = max(0, self.metadata['remaining_content_length'])
 
diff --git a/tests/test_request_left.py b/tests/test_request_left.py
index 373b2e49..4d448388 100644
--- a/tests/test_request_left.py
+++ b/tests/test_request_left.py
@@ -22,7 +22,7 @@ class SignalCatcherSpider(Spider):
         return spider
 
     def on_request_left(self, request, spider):
-        self.caught_times = self.caught_times + 1
+        self.caught_times += 1
 
 
 class TestCatching(TestCase):
```

wRAR:
Can you please make a pull request with the changes you propose?

PR #5321, opened 2021-11-15
added augmented assignment statements

IsomerX:
Found one place where augmented assignment could have been used, with regards to issue #5314


wRAR:
Doesn't that issue list 3 more cases?

wRAR:
There is also a flake8 error for this change.

Issue #3529, opened 2018-12-11
SitemapSpider memory issues 


altunyurt:
I'm using SitemapSpider on a sitemapindex consisting of 20-30 sitemaps each having  50k urls.
Even trying each sitemap alone ends up eating all the memory on a 6gb machine, let alone the millions of urls in total of all the sitemaps in the index. 

IIRC, the parser keeps the documents in memory  during the operation. So I've monkey patched the scrapy.utils.Sitemap with the following snippet. The original code is at https://stackoverflow.com/a/12161078  

```
# patching scrapy/utils/sitemap.py
import lxml.etree


class Sitemap(object):
    """Class to parse Sitemap (type=urlset) and Sitemap Index
    (type=sitemapindex) files"""

    def __init__(self, xmltext):
        self.type = "urlset" if "urlset" in xmltext[:100] else "sitemapindex"

        tag = {"urlset": "url", "sitemapindex": "sitemap"}[self.type]
        self._root = lxml.etree.iterparse(
            xmltext,
            tag=tag,
            event=("end",),
            recover=True,
            remove_comments=True,
            resolve_entities=False,
        )

    def __iter__(self):

        for event, elem in self._root:
            loc = elem.find("{*}loc")
            if loc is None:
                continue
            yield {"loc": loc.text}
            # It's safe to call clear() here because no descendants will be
            # accessed
            elem.clear()
            # Also eliminate now-empty references from the root node to elem
            for ancestor in elem.xpath("ancestor-or-self::*"):
                while ancestor.getprevious() is not None:
                    del ancestor.getparent()[0]
        del self._root

```
and using it as follows allowed me to run the task much smoothly, utilizing at most 3gb at any time.

```
worker.py 
-------------------------
import scrapy
from core.patches.sitemap import Sitemap

__all__ = ["Spider"]

scrapy.spiders.Sitemap = Sitemap


class Spider(scrapy.spiders.SitemapSpider):
    sitemap_urls = [ sitemap1, sitemap2 ...  ]

    def parse(self, response):
    ...
```

I can create a pull request if this is ok.

kmike:
See also: https://github.com/scrapy/scrapy/pull/605

joaquingx:
Hi @altunyurt, do you have any progress? I would like to work on this also :smile:

kinoute:
@altunyurt @joaquingx I will be interested in using `iterparse` too for a Sitemap Spider. I wasn't able to make it work with your snippet. 

[...]

PR #5328, opened 2021-11-22
Per slot settings

GeorgeA92:
Implementation of per slot(downloader slot) settings.

On general case scrapy create new slot for each domain using `DOWNLOAD_DELAY`, `CONCURRENT_REQUESTS_PER_DOMAIN` (or `CONCURRENT_REQUESTS_PER_IP`) and `RANDOMIZE_DOWNLOAD_DELAY` from settings.

Following changes aimed to enable configure delay, concurency, randomized delay per Downloader.Slot (per domain) to set custom delay, concurrency for individually for each domain.

Related part of downloader code updated to accept per slot settings from `PER_SLOT_SETTINGS` setting:

```python
    custom_settings = {
        "DOWNLOAD_DELAY": 3,
        "PER_SLOT_SETTINGS": {
            'quotes.toscrape.com': {
                'concurrency':1,
                'delay': 15,
                'randomize_delay': False
            },
            'books.toscrape.com': {
                'delay': 39,
            }
        }
    }
```
<details>
<summary>example test .py script</summary>

```python
import scrapy
import logging

class PerSlotTestSpider(scrapy.Spider):
    name = "per_slot_settings_set"
    custom_settings = {
        "DOWNLOAD_DELAY": 3,
        "PER_SLOT_SETTINGS": {
            'quotes.toscrape.com': {
                'concurrency':1,
                'delay': 15,
                'randomize_delay': False
            },
            'books.toscrape.com': {
                'delay': 39,
            }
        }
    }

    def start_requests(self):
        yield scrapy.Request(url="http://books.toscrape.com/", callback=self.not_parse) # per slot settings (concurrency, randomized -> from spider/project settings)
        yield scrapy.Request(url="http://quotes.toscrape.com/page/1/", callback=self.parse) # 100% from per slot settings
        yield scrapy.Request(url="http://example.com/", callback=self.not_parse) # from 100% from general settings

    def parse(self, response):
        next =  response.css("li.next a::attr(href)").extract_first()
        if next:
            yield scrapy.Request(url=response.urljoin(next), callback=self.parse)
        else:
            yield scrapy.Request(url='http://books.toscrape.com/', callback=self.not_parse, dont_filter=True)
            yield scrapy.Request(url="http://example.com/", callback=self.not_parse, dont_filter=True)

    def not_parse(self, response):
        pass
process = CrawlerProcess()
process.crawl(PerSlotTestSpider)
process.start()
```
</details>

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -108,7 +108,8 @@ class Downloader:
             randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
             new_slot = Slot(conc, delay, randomize_delay)
             self.slots[key] = new_slot
-            logger.debug(f"Downloader slot created {'from per slot settings' if key in self.per_slot_settings.keys() else ''}: {new_slot}")
+            logger.debug(
+                f"Downloader slot '{key}' - created {'(from per slot settings)' if key in self.per_slot_settings.keys() else ''}: {new_slot}")
 
         return key, self.slots[key]
 
@@ -205,4 +206,7 @@ class Downloader:
         mintime = time() - age
         for key, slot in list(self.slots.items()):
             if not slot.active and slot.lastseen + slot.delay < mintime:
-                self.slots.pop(key).close()
+                inactive_slot = self.slots.pop(key)
+                inactive_slot.close()
+                logger.debug(
+                    f"Downloader slot '{key}' - closed {'(from per slot settings)' if key in self.per_slot_settings.keys() else ''} : {inactive_slot}")
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -83,6 +83,7 @@ class Downloader:
         self.middleware = DownloaderMiddlewareManager.from_crawler(crawler)
         self._slot_gc_loop = task.LoopingCall(self._slot_gc)
         self._slot_gc_loop.start(60)
+        self.per_slot_settings = self.settings.getdict('PER_SLOT_SETTINGS', {})
 
     def fetch(self, request, spider):
         def _deactivate(response):
@@ -99,9 +100,13 @@ class Downloader:
     def _get_slot(self, request, spider):
         key = self._get_slot_key(request, spider)
         if key not in self.slots:
-            conc = self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            conc = self.per_slot_settings.get(key,{}).get('concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency)
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
-            self.slots[key] = Slot(conc, delay, self.randomize_delay)
+            delay = self.per_slot_settings.get(key,{}).get('delay', delay)
+            randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
+            new_slot = Slot(conc, delay, randomize_delay)
+            self.slots[key] = new_slot
+
 
         return key, self.slots[key]
 
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -1,3 +1,4 @@
+import logging
 import random
 from time import time
 from datetime import datetime
@@ -12,6 +13,7 @@ from scrapy import signals
 from scrapy.core.downloader.middleware import DownloaderMiddlewareManager
 from scrapy.core.downloader.handlers import DownloadHandlers
 
+logger = logging.getLogger(__name__)
 
 class Slot:
     """Downloader slot"""
@@ -106,7 +108,7 @@ class Downloader:
             randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
             new_slot = Slot(conc, delay, randomize_delay)
             self.slots[key] = new_slot
-
+            logger.debug(f"Downloader slot created {'from per slot settings' if key in self.per_slot_settings.keys() else ''}: {new_slot}")
 
         return key, self.slots[key]
 
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -15,6 +15,7 @@ from scrapy.core.downloader.handlers import DownloadHandlers
 
 logger = logging.getLogger(__name__)
 
+
 class Slot:
     """Downloader slot"""
 
@@ -102,9 +103,9 @@ class Downloader:
     def _get_slot(self, request, spider):
         key = self._get_slot_key(request, spider)
         if key not in self.slots:
-            conc = self.per_slot_settings.get(key,{}).get('concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency)
+            conc = self.per_slot_settings.get(key, {}).get('concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency)
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
-            delay = self.per_slot_settings.get(key,{}).get('delay', delay)
+            delay = self.per_slot_settings.get(key, {}).get('delay', delay)
             randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
             new_slot = Slot(conc, delay, randomize_delay)
             self.slots[key] = new_slot


scrapy/core/downloader/__init__.py
```diff
@@ -83,6 +86,7 @@ def __init__(self, crawler):
         self.middleware = DownloaderMiddlewareManager.from_crawler(crawler)
         self._slot_gc_loop = task.LoopingCall(self._slot_gc)
         self._slot_gc_loop.start(60)
+        self.per_slot_settings = self.settings.getdict('PER_SLOT_SETTINGS', {})
```
Gallaecio, line 89: I‘m thinking that we could just name the setting `DOWNLOAD_SLOTS`, in line with the `FEEDS` setting and the `download_slot` meta key. I‘m not against something like `DOWNLOAD_SLOT_SETTINGS`, and I think an argument could me made that adding `SETTINGS` makes it clear that there may be more slots that those defined there (since we are not defining slots, but slot settings), but `DOWNLOAD_SLOTS` is shorter and seems to be unambiguous enough (we probably should still clarify in the documentation that slots not in the setting may still be used).

scrapy/core/downloader/__init__.py
```diff
@@ -99,9 +103,14 @@ def needs_backout(self):
...
-            conc = self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            conc = self.per_slot_settings.get(key, {}).get('concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency)
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
-            self.slots[key] = Slot(conc, delay, self.randomize_delay)
+            delay = self.per_slot_settings.get(key, {}).get('delay', delay)
+            randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
+            new_slot = Slot(conc, delay, randomize_delay)
+            self.slots[key] = new_slot
+            logger.debug(
+                f"Downloader slot '{key}' - created {'(from per slot settings)' if key in self.per_slot_settings.keys() else ''}: {new_slot}")
```
Gallaecio, line 113: I understand this may have helped while implementing this feature, but I am not sure we should keep these debug messages around.

Same for the message below.

Gallaecio (commented):
Looks good overall, nice job! Tests and documentation should go next, before we merge.

Gallaecio:
Re-running failing jobs…

scrapy/core/downloader/__init__.py
```diff
@@ -198,4 +204,5 @@ def _slot_gc(self, age=60):
         mintime = time() - age
         for key, slot in list(self.slots.items()):
             if not slot.active and slot.lastseen + slot.delay < mintime:
-                self.slots.pop(key).close()
+                inactive_slot = self.slots.pop(key)
+                inactive_slot.close()
```
Gallaecio, line 208: According to Codecov this part is not covered by tests (it probably was not covered by tests before.

It seems functionally equivalent to the code before the change, so maybe you could just revert this change.

I am OK with the change though, it is more readable. But if you are going to change it, it would be great if you could also extend test coverage for it. Although that may not be trivial.

GeorgeA92: Yes. Sure (reverted changes). I previously made this change(assigning inactive slot to variable) for logging which is removed as result of previous feedback.

scrapy/core/downloader/__init__.py
```diff
@@ -99,9 +100,14 @@ def needs_backout(self):
...
         key = self._get_slot_key(request, spider)
         if key not in self.slots:
-            conc = self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            conc = self.per_slot_settings.get(key, {}).get(
+                'concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            )
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
-            self.slots[key] = Slot(conc, delay, self.randomize_delay)
+            delay = self.per_slot_settings.get(key, {}).get('delay', delay)
+            randomize_delay = self.per_slot_settings.get(key, {}).get('randomize_delay', self.randomize_delay)
```
kmike, line 108: A nitpick: what do you think about adding this variable:
```py
slot_settings = self.per_slot_settings.get(key, {})
```
It seems it's used a few times here.

tests/test_downloaderslotssettings.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @defer.inlineCallbacks
+    def test_delay(self):
+        crawler = CrawlerRunner().create_crawler(DownloaderSlotsSettingsTestSpider)
+        yield crawler.crawl(mockserver=self.mockserver)
+        slots = crawler.engine.downloader.slots
+        times = crawler.spider.times
+        tolerance = 0.3
+
+        delays_real = {k: v[1] - v[0] for k, v in times.items()}
```
kmike, line 72:
```suggestion
        delays_real = {k: end - start for k, (start, end) in times.items()}
```

tests/test_downloaderslotssettings.py
```diff
@@ -0,0 +1,75 @@
...
+        error_delta = {k: 1 - delays_real[k] / v.delay for k, v in slots.items()}
```
kmike, line 73: So, for one slot in the tests the delay is 2s, for another it's 1.5s, and it's 1s for the default. Let's say implementation is incorrect, and in all cases delay of 2s is used. error_delta values would be `1 - 2/1.5 = -0.33` and (1-2/1 = 01)`. The assertion below will pass: `max(-0.33, -1) = -0.33 < 0.3` is True.

If the 1.5s delay is used instead of 2s delay, then error_delta values would be `1 - 1.5/2 = 0.25`. `0.25 < 0.3` is True again. It may catch the default 1s download delay though.

Maybe there are reasons these tests may detect an issue (e.g. default DOWNLOAD_DELAY), but I think it could make sense to try improving them. Have you encountered issues with something simpler, e.g. `abs(delays_real[k] - v.delay) < tolerance`?

kmike:
Hey! The feature makes sense, and the implementation looks good 👍 I think to merge it, we need to fix a few small issues and add the docs.

scrapy/core/downloader/__init__.py
GeorgeA92: Done

tests/test_downloaderslotssettings.py
GeorgeA92: This test implemented directly in the same way as existing test for `DOWNLOAD_DELAY` setting (counting that we need to test multiple downloader slots now)
https://github.com/scrapy/scrapy/blob/80040758236ba1e35a4aa499faf33ed96fc12a92/tests/test_crawl.py#L85-L95

On my local tests `delays_real` calculated from `time.time` calls in spider by some unknown reason always a bit lower that delays from settings:
for delays [1, 1.5, 2] - I received [~0.90, ~1.43, ~ 1.90] in `delays_real`. So on this condition I don't expect issues from value that became negative.
Anyway error calculation method updated to be sure that it will not happen.

I also thought about increasing delay for test from [1, 1.5, 2] to for example [3, 5, 7] in this case with increased delays in test we can safely reduce `tolerance`.

The main question what is acceptable total time for this test?

Gallaecio: > The main question what is acceptable total time for this test?

The minimum value that still makes the test reliable, both in the sense that it does not break randomly, and that it indeed validates what it is meant to validate. If we need 30 seconds for that, so be it. But if it can be done in 5 seconds, that would be better.

GeorgeA92:
Updated pull request.
As this  https://github.com/scrapy/scrapy/issues/3585 is open and we don't have any other mention of downloader slot component in docs - at this stage it is not clear how detailed this setting needs to be documented.

docs/topics/settings.rst
```diff
@@ -771,6 +771,38 @@ The default HTTPS handler uses HTTP/1.1. To use HTTP/2:
...
+            }
+        }
+
+.. note::
+
+    For other downloader slots default settings values will be used:
+
+        -   ``DOWNLOAD_DELAY`` -> ``delay``
+        -   ``CONCURRENT_REQUESTS_PER_DOMAIN`` -> ``concurency``
+        -   ``RANDOMIZE_DOWNLOAD_DELAY`` -> ``randomize_delay``
```
Gallaecio, line 803:
```suggestion
    -   :setting:`DOWNLOAD_DELAY`: ``delay``
    -   :setting:`CONCURRENT_REQUESTS_PER_DOMAIN`: ``concurency``
    -   :setting:`RANDOMIZE_DOWNLOAD_DELAY`: ``randomize_delay``
```

Gallaecio (commented):
~~https://github.com/scrapy/scrapy/pull/5328/files#r933163347 is still not addressed, as far as I can tell. I think we need to apply per-slot settings *after* the call to `_get_concurrency_delay`.~~

docs/topics/settings.rst
GeorgeA92: Updated.

scrapy/core/downloader/__init__.py
```diff
@@ -99,9 +100,14 @@ def needs_backout(self):
     def _get_slot(self, request, spider):
         key = self._get_slot_key(request, spider)
         if key not in self.slots:
-            conc = self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            conc = self.per_slot_settings.get(key, {}).get(
+                'concurrency', self.ip_concurrency if self.ip_concurrency else self.domain_concurrency
+            )
             conc, delay = _get_concurrency_delay(conc, spider, self.settings)
```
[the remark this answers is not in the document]
GeorgeA92: @Gallaecio 
Now per-slot settings takes precedence over settings from all other places. (Assuming that on https://github.com/scrapy/scrapy/issues/5584 we will have final decision about `spider.max_concurrent_request`)

Gallaecio (approved):

wRAR:
How does this solve #3529?

GeorgeA92:
> How does this solve #3529?

This pull request doesn't solve https://github.com/scrapy/scrapy/issues/3529 itself
However applying functionality from this PR into SitemapSpider (I think) can.
By asigning requests to sitemaps to custom downloader slot ... `sitemaps` with concurrency 1 defined in `DOWNLOAD_SLOTS`

I suppose that @Gallaecio misunderstood my mention of https://github.com/scrapy/scrapy/issues/3529 when we discussed this pull request

docs/topics/settings.rst
```diff
@@ -771,6 +771,31 @@ The default HTTPS handler uses HTTP/1.1. To use HTTP/2:
...
+            "quotes.toscrape.com": {"concurrency": 1, "delay": 2, "randomize_delay": False},
+            "books.toscrape.com": {"delay": 3, "randomize_delay": False},
+        }
+
+.. note::
+
+    For other downloader slots default settings values will be used:
+
+    -   :setting:`DOWNLOAD_DELAY`: ``delay``
+    -   :setting:`CONCURRENT_REQUESTS_PER_DOMAIN`: ``concurency``
```
wRAR, line 795:
```suggestion
    -   :setting:`CONCURRENT_REQUESTS_PER_DOMAIN`: ``concurrency``
```

GeorgeA92: typo in docs fixed

wRAR (approved):

wRAR:
Thanks!

PR #5329, opened 2021-11-23
include link to NEWS file location in home directory

jcrineer:


wRAR:
I don't think this makes sense?

jcrineer:
An attempt to make navigation to the news.rst file easier.

wRAR:
> a user needs to physically navigate through several directories

This depends on the way the user looks at the files.

> Providing a direct link would make it faster.

It provides a wrong link though. Normally you would want to read the file from the version you are looking at, not from the master branch.

Issue #4485, opened 2020-04-12
TWISTED_REACTOR setting not honored from Spider.custom_settings


elacuesta:
### Description

The value of the `TWISTED_REACTOR` setting is not taken into account if the setting is specified in a spider's `custom_settings` attribute.
It works well if the setting is specified in a project's settings file, as a parameter when creating a `CrawlerProcess` object (as the tests show) or as a CLI argument with the `-s` option. 

### Steps to Reproduce

1. Create a file with the following contents:
```python
# asyncio_spider.py
import asyncio

from scrapy import Spider


class AsyncIOSpider(Spider):
    name = "asyncio"
    start_urls = ["https://example.org"]
    custom_settings = {
        "TWISTED_REACTOR": "twisted.internet.asyncioreactor.AsyncioSelectorReactor",
    }

    async def parse(self, response):
        await asyncio.sleep(1)
        yield {"foo": "bar"}
```

2. Execute the spider: `scrapy runspider asyncio_spider.py`

**Expected behavior:** 
The spider should run fine, without exceptions

**Actual behavior:**
The following exception is raised:
```
2020-04-12 00:04:23 [scrapy.core.scraper] ERROR: Spider error processing <GET https://example.org> (referer: None)
Traceback (most recent call last):
  File "/.../scrapy/venv-scrapy/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/.../scrapy/scrapy/utils/py36.py", line 8, in collect_asyncgen
    async for x in result:
  File "/.../scrapy/test-spiders/asyncio_spider.py", line 14, in parse
    await asyncio.sleep(1)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/tasks.py", line 595, in sleep
    return await future
RuntimeError: await wasn't used with future
```
This is because the `asyncio`-based reactor is not actually installed, as the third log line of the job shows:
```
2020-04-12 00:04:23 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.selectreactor.SelectReactor
```

Full logs:
```
2020-04-12 00:04:23 [scrapy.utils.log] INFO: Scrapy 2.0.1 started (bot: scrapybot)
[...]

PR #5332, opened 2021-11-26
Remove the experimental note about asyncio.

wRAR:
I also added a line about `TWISTED_REACTOR` and `custom_settings`.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1643,6 +1643,11 @@ default reactor defined by Twisted for the current platform will be used. This
 is to maintain backward compatibility and avoid possible problems caused by
 using a non-default reactor.
 
+.. note::
+    This setting has no effect in a spider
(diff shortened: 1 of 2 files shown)


docs/topics/settings.rst
```diff
@@ -1643,6 +1643,11 @@ default reactor defined by Twisted for the current platform will be used. This
 is to maintain backward compatibility and avoid possible problems caused by
 using a non-default reactor.
 
+.. note::
+    This setting has no effect in a spider
+    :attr:`~scrapy.Spider.custom_settings` attribute because it needs to be
+    read very early.
+
```
elacuesta, line 1650: I'm unsure about this note, we already have #4485 which treats this as a bug.

wRAR: Oh nice. I wonder if this is possible to fix.

Gallaecio (approved):

PR #5334, opened 2021-11-26
Fix typing of middleware methods.

wRAR:
Unfortunately, `MiddlewareManager.methods` contains methods in most cases, but methods or None for spider middleware `process_spider_output` and `process_spider_exception` methods, because those need to keep their mutual order, as they can call each other.

### DIFF
--- a/scrapy/middleware.py
+++ b/scrapy/middleware.py
@@ -9,7 +9,7 @@ from scrapy import Spider
 from scrapy.exceptions import NotConfigured
 from scrapy.settings import Settings
 from scrapy.utils.misc import create_instance, load_object
-from scrapy.utils.defer import process_parallel, process_chain, process_chain_both
+from scrapy.utils.defer import process_parallel, process_chain
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


scrapy/middleware.py
```diff
@@ -64,14 +65,17 @@ def _add_middleware(self, mw) -> None:
...
-        return process_parallel(self.methods[methodname], obj, *args)
+        methods = cast(Iterable[Callable], self.methods[methodname])
+        return process_parallel(methods, obj, *args)
 
     def _process_chain(self, methodname: str, obj, *args) -> Deferred:
-        return process_chain(self.methods[methodname], obj, *args)
+        methods = cast(Iterable[Callable], self.methods[methodname])
+        return process_chain(methods, obj, *args)
 
     def _process_chain_both(self, cb_methodname: str, eb_methodname: str, obj, *args) -> Deferred:
```
wRAR, line 75: This method is not covered because it's never used. I think we can drop it?

Gallaecio (approved):

PR #5340, opened 2021-11-29
Correcting path in `GCSFilesStore`

SeguinBe:
There is a simple bug when checking if a file already exists (so that it is not downloaded anymore) in `GCSFilesStore`.

Because of that, it was redownloading files even if it was not needed, every single time.

On another note, I always get the warning that I do not have the `storage.objects.get` IAM permission (even if I absolutely have it). But to be faire that is probably more an issue on `google-storage` side.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -223,7 +223,7 @@ class GCSFilesStore:
             else:
                 return {}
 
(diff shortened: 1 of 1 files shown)


SeguinBe:
Realised it was a duplicate of #5318. So closing...

PR #5341, opened 2021-12-04
Correct small inaccuracy

mjh-sakh:
The process repeats from step 3, the scheduler feeds request to the engine. Steps 1 and 2 are not parts of the loop as their incarnations steps 7 and 8 are parts of the loop.


wRAR:
Please change the PR to merge into master instead of 2.5.

wRAR:
Now it includes unrelated commits from non-master branches.

mjh-sakh:
> Now it includes unrelated commits from non-master branches (I think).

This I was not able to control. It bundled together automatically when I updated to `master`. I assumed that it's a preferred behavior. Plus I don't have any tools to change it. So nothing I can do unless you precisely instruct me.

wRAR:
You need to make your changes on top of the master branch, not the 2.5 one, you can try rebasing this branch on top of master and force-pushing it, if that fails you'll need to make a fresh branch and a fresh PR.

mjh-sakh:
Let me try to do it from scratch.

PR #5349, opened 2021-12-08
Updated overview for video intro

ASWU713:
Updated overview section with video introduction


Gallaecio:
I am not comfortable with linking to an external video. It can get obsolete quickly, its hosting is out of out control…

PR #5351, opened 2021-12-15
AutoThrottle: use live settings

milahu:


milahu:
nevermind, settings are constant

https://stackoverflow.com/questions/54523778/scrapy-change-settings-at-runtime-based-on-attribute-provided

PR #5352, opened 2021-12-23
Install the reactor after reading spider settings 

wRAR:
Closes: #4485

### DIFF
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -102,17 +102,6 @@ reactor after ``MySpider`` has finished running.
     d.addBoth(lambda _: reactor.stop())
     reactor.run() # the script will block here until the crawling is finished
 
-.. note::
-    .. versionchanged:: VERSION
-
-    The Twisted reactor is now installed when
-    :meth:`~scrapy.crawler.CrawlerProcess.crawl` is first called, not when a
-    :class:`scrapy.crawler.CrawlerProcess` object is created. Because of this,
-    :setting:`TWISTED_REACTOR` and :setting:`ASYNCIO_EVENT_LOOP` are now
-    honored in :attr:`~scrapy.Spider.custom_settings`. In older Scrapy versions
-    they are silently ignored when set there and you need to set these settings
-    in some other way.
-
 .. seealso:: :doc:`twisted:core/howto/reactor-basics`
 
 .. _run-multiple-spiders:
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -102,6 +102,17 @@ reactor after ``MySpider`` has finished running.
     d.addBoth(lambda _: reactor.stop())
     reactor.run() # the script will block here until the crawling is finished
 
+.. note::
+    .. versionchanged:: VERSION
+
+    The Twisted reactor is now installed when
+    :meth:`~scrapy.crawler.CrawlerProcess.crawl` is first called, not when a
+    :class:`scrapy.crawler.CrawlerProcess` object is created. Because of this,
+    :setting:`TWISTED_REACTOR` and :setting:`ASYNCIO_EVENT_LOOP` are now
+    honored in :attr:`~scrapy.Spider.custom_settings`. In older Scrapy versions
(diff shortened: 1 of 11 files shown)


scrapy/crawler.py
```diff
@@ -69,6 +70,19 @@ def __init__(self, spidercls, settings=None):
...
+            # this needs to be done after the spider settings are merged,
+            # but before something imports twisted.internet.reactor
+            if self.settings.get("TWISTED_REACTOR"):
+                install_reactor(self.settings["TWISTED_REACTOR"], self.settings["ASYNCIO_EVENT_LOOP"])
+            else:
+                from twisted.internet import default
+                default.install()
+            log_reactor_info()
+        if self.settings.get("TWISTED_REACTOR"):
+            verify_installed_reactor(self.settings["TWISTED_REACTOR"])
```
pawelmhm, line 84: works fine, tested this in my project. Minor thing, you could probably simplify this to something like this to avoid repeatedly calling self.settings.get('TWISTED_REACTOR')

```python
reactor_class = self.settings.get("TWISTED_REACTOR")
if init_reactor:
    if reactor_class:
        install_reactor(reactor_class, self.settings["ASYNCIO_EVENT_LOOP"])
... etc

```

I also wonder if default value for init_reactor should be False? Are there any legitimate cases when you don't want to init reactor if you have it in settings?

wRAR: Thanks, I'll implement this simplification. As for the flag, it was the default behavior before this change, and the use case to not init the reactor is CrawlerRunner.

docs/topics/practices.rst
```diff
@@ -102,6 +102,17 @@ reactor after ``MySpider`` has finished running.
...
+.. note::
+    .. versionchanged:: VERSION
+
+    The Twisted reactor is now installed when
+    :meth:`~scrapy.crawler.CrawlerProcess.crawl` is first called, not when a
+    :class:`scrapy.crawler.CrawlerProcess` object is created. Because of this,
+    :setting:`TWISTED_REACTOR` and :setting:`ASYNCIO_EVENT_LOOP` are now
+    honored in :attr:`~scrapy.Spider.custom_settings`. In older Scrapy versions
+    they are silently ignored when set there and you need to set these settings
+    in some other way.
```
Gallaecio, line 114: I would prefer to move this to the release notes, to have the rest of the documentation only discuss what is current.

Maybe copy this _as is_ at the beginning of `news.rst` for now? (It is very well written, and we may forget to describe it in this much detail otherwise; we can adjust the wording later when we work on the final draft of the release notes)

(the same applies to the similar note in `settings.rst`)

docs/topics/settings.rst
```diff
@@ -1638,10 +1638,18 @@ which raises :exc:`Exception`, becomes::
 
 
 The default value of the :setting:`TWISTED_REACTOR` setting is ``None``, which
-means that Scrapy will not attempt to install any specific reactor, and the
-default reactor defined by Twisted for the current platform will be used. This
-is to maintain backward compatibility and avoid possible problems caused by
-using a non-default reactor.
+means that Scrapy will install the default reactor defined by Twisted for the
+current platform will be used. This is to maintain backward compatibility and
+avoid possible problems caused by using a non-default reactor.
```
Gallaecio, line 1643:
```suggestion
means that Scrapy will install the default reactor defined by Twisted for the
current platform. This is to maintain backward compatibility and avoid possible 
problems caused by using a non-default reactor.
```

Gallaecio (approved):
Looks good to me, but I leave the conflict resolution to you :)

AlvinSartorTrityum:
When is the next version containing this fix supposed to be released? Thanks!

wRAR:
@AlvinSartorTrityum we don't currently have a date for the next release.

JSPromisel:
Any updates on when this will be released? Would be very useful to have.

elacuesta:
This was released as part of [2.6](https://docs.scrapy.org/en/latest/news.html#scrapy-2-6-0-2022-03-01)

PR #5354, opened 2021-12-23
Fixed a bug in Feed Export doc

salehdeh76:
The source code does not use return value of uri_params function and the params should be updated in place.
The document was misleading about this.

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -751,7 +751,7 @@ source spider in the feed URI:
 
         # myproject/utils.py
         def uri_params(params, spider):
-            return {**params, 'spider_name': spider.name}
+            params.update({'spider_name': spider.name})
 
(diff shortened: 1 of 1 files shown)


wRAR (approved):

wRAR (changes requested):
Please make this PR merge into master instead of 2.5.

wRAR:
This merges 2.5 into master. You need to rebase your changes on top of master instead.

salehdeh76:
> This merges 2.5 into master. You need to rebase your changes on top of master instead.

I'm Sorry. It fixed now.

wRAR (approved):

wRAR:
Thank you.

elacuesta:
Closing, this is being treated as a bug: #4962. We want to update the behavior rather than the docs.

PR #5355, opened 2021-12-24
Add error log for crawler.crawl() defered

lihanx:
When the "Custom Spider" class cannot be instantiated correctly, the running crawler will be stuck and output nothing. It makes debugging more easily to show an error log on this situation. Perhaps😛

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -23,6 +23,7 @@ from scrapy.settings import overridden_settings, Settings
 from scrapy.signalmanager import SignalManager
 from scrapy.utils.log import (
     configure_logging,
+    failure_to_exc_info,
     get_scrapy_root_handler,
     install_scrapy_root_handler,
     log_scrapy_info,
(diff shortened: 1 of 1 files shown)


wRAR:
Do you have an example that uses this?

lihanx:
> Do you have an example that uses this?

Yes. You may throw an exception in the `Spider.__init__()` or `Spider.from_crawler()` method to reproduce it. 

For Example:
```python

    name = "Example"

        super(ExampleSpider, self).__init__(*args, **kwargs)
        raise Exception("Anything")

    @classmethod
    def from_crawler(cls, crawler, *args, **kwargs):
        spider = super(ExampleSpider, cls).from_crawler(crawler, *args, **kwargs)
        return spider

    def start_requests(self):
        yield scrapy.Request(
            url="https://www.xxx.com/",
            callback=self.parse,
        )

        pass

process = CrawlerProcess()
process.crawl(ExampleSpider)

# process.start(stop_after_crawl=False)  # Output traceback and terminated normally.
process.start(stop_after_crawl=True)   # Stuck forever and output nothing unless terminated by ctrl+c
```

oh, it's only for the `stop_after_crawl` flag is `True` and the event loop won't stop automatically. I didn't realize it. Sorry.

PR #5357, opened 2021-12-30
downloader.webclient: make reactor import local

pawelmhm:
All reactor imports must be local not top level because twisted.internet.reactor import installs default reactor, and if you want to use non-default reactor like asyncio top level import will break things for you. 

In my case I had a middleware imported in spider

in spider
```python
from project.middlewares.retry import CustomRetryMiddleware
```

this was importing

```
from scrapy.downloadermiddlewares import retry
```
then retry was importing

```
from scrapy.core.downloader.handlers.http11 import TunnelError

```

and http11 was importing
```
from scrapy.core.downloader.webclient import _parse
```

and webclient was importing and installing reactor. I wonder how this was not detected earlier? It should break things for people trying to be using asyncio reactor in more complex projects.

This import of reactor may be annoying for many projects because it is easy to forget and do import reactor somewhere in your scrapy middleware or somewhere else. Then you'll get an error when using asyncio, and some people will be confused they may not know where to look for source of the problem.

### DIFF
--- a/scrapy/core/downloader/webclient.py
+++ b/scrapy/core/downloader/webclient.py
@@ -3,7 +3,7 @@ from time import time
 from urllib.parse import urlparse, urlunparse, urldefrag
 
 from twisted.web.http import HTTPClient
-from twisted.internet import defer, reactor
+from twisted.internet import defer
 from twisted.internet.protocol import ClientFactory
 
 from scrapy.http import Headers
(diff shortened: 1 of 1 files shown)


wRAR (approved):

Gallaecio (approved):

Gallaecio:
Nice catch!

PR #5358, opened 2021-12-30
Document coroutines for signals.

### LINKED ISSUE #4852: Document async def interfaces for signals

burakcank:
Fixes #4852.


wRAR (approved):

kmike:
Thanks @burakcank!

PR #5371, opened 2022-01-21
Remove scrapy.pipelines.images.NoimagesDrop

Mikeb196840:
Read the issue seems that it might be a little pointless deprecating the class vs just removing it. Please review.

Left a comment in case you might want to deprecate it later.


wRAR:
We don't remove public classes or methods without deprecating them first.

PR #5372, opened 2022-01-21
deprecate scrapy.pipelines.images.NoimagesDrop

### LINKED ISSUE #5368: Deprecate scrapy.pipelines.images.NoimagesDrop
As noticed in #4910, `scrapy.pipelines.images.NoimagesDrop` is not used in the Scrapy code so we should deprecate (and later remove) it.

GhostViz:
Deprecate scrapy.pipelines.images.NoimagesDrop 

Resolved #5368

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -60,11 +60,9 @@ Scrapy 2.5.1 (2021-10-05)
     need to upgrade scrapy-splash to a greater version for it to continue to
     work.
 
-    Deprecations
-    ~~~~~~~~~~~~
-
-    -   The :class:`scrapy.pipelines.images.NoimagesDrop` class is now
-        deprecated as it is not being used in Scrapy code. (:issue:`5368`)
+*   **Deprecations:**
+    The :class:`scrapy.pipelines.images.NoimagesDrop` class is now
+    deprecated as it is not being used in Scrapy code. (:issue:`5368`)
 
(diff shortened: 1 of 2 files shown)


docs/news.rst
```diff
@@ -60,6 +60,10 @@ Scrapy 2.5.1 (2021-10-05)
     need to upgrade scrapy-splash to a greater version for it to continue to
     work.
 
+*   **Deprecations:**
```
wRAR, line 63: This adds a note to the 2.5.1 section which is not true. And in general e don't add new entries with the changes themselves, only doing that when preparing a release. So it's fine to omit it.

scrapy/pipelines/images.py
```diff
@@ -17,12 +17,13 @@
...
 from scrapy.utils.misc import md5sum
 from scrapy.utils.python import to_bytes
+from scrapy.utils.decorators import deprecated
 
 
+@deprecated("scrapy.pipelines.images.NoimagesDrop")
 class NoimagesDrop(DropItem):
     """Product with no images exception"""
 
-
```
wRAR, line 25: PEP8 mandates two newlines between classes.

wRAR (changes requested):

wRAR:
Fixed by https://github.com/scrapy/scrapy/pull/5489.

PR #5370, opened 2022-01-21
Removed  the pywin32 docs section

NaincyKumariKnoldus:
Hey @wRAR I have removed  the pywin32 docs section from the FAQ. Please Review my PR!

Fixes #5369, resolves #5373, resolves #5373, resolves #5446

PR #5373, opened 2022-01-22
remove pywin32 docs section

### LINKED ISSUE #5369: Remove the pywin32 docs section
As noticed in #4492, the FAq question at https://docs.scrapy.org/en/latest/faq.html#scrapy-crashes-with-importerror-no-module-named-win32api is no longer useful as the Twisted bug was fixed long time ago. We should remove this question.

ThePeeps191:
Fixes #5369


calgary34:
This was already done in #5370.

PR #5374, opened 2022-01-22
Issue #5366 Move from optparse to argparse

alexpdev:
Update deprecated optparse it's equivalent in argparse  (#5366 )

(edit) fixes #5366

PR #5382, opened 2022-01-28
Updated optparse to argparse

bryangators:
Resolves use of deprecated octparse to newer argparse. Similar to [#5374](https://github.com/scrapy/scrapy/pull/5374) but this commit keeps same behavior in scrapy/commands/view.py where "--header" option is removed. New syntax needed for removing an argument.


bryangators:
Hunting down the issues now and will make new commit with all tests passing

PR #5389, opened 2022-02-04
CI: stop using tox-pip-version

### LINKED ISSUE #5388: CI fails on Python 3.8 during dependency installation
@elacuesta [has already looked into it in some detail](https://github.com/scrapy/scrapy/pull/5374#issuecomment-1028123418).

The issue is now happening in the main branch, after merging some new changes.

Gallaecio:
Aims to fix #5388.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -19,7 +19,6 @@ jobs:
         - python-version: 3.8
           env:
             TOXENV: pylint
-            TOX_PIP_VERSION: 20.3.3
         - python-version: 3.6
           env:
             TOXENV: typing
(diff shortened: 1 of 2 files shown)


Gallaecio:
I removed the only thing that the 2 failing jobs had in common that the other jobs did not have, and it has worked.

We introduced tox-pip-version in https://github.com/scrapy/scrapy/pull/5207 because otherwise CI would fail. The situation now is the opposite.

I think https://github.com/pypa/pip/pull/10481 may be the reason we can now use the latest pip without issues.

wRAR (approved):

PR #5394, opened 2022-02-05
Fix csviter call, add parse_rows test

Laerte:
Closes #5391

Issue #5391, opened 2022-02-05
Incorrect initialization of `csviter` in `scrapy.spiders.feed.CSVFeedSpider`


dchaplinsky:
### Description

According to the master branch, `scrapy.spiders.feed.CSVFeedSpider` initializes csviter like this:
```python
for row in csviter(response, self.delimiter, self.headers, self.quotechar):
```
https://github.com/scrapy/scrapy/blob/master/scrapy/spiders/feed.py#L126

But `csviter` definition says:
`def csviter(obj, delimiter=None, headers=None, encoding=None, quotechar=None):`
https://github.com/scrapy/scrapy/blob/master/scrapy/utils/iterators.py#L96

So effectively it passed quotechar instead of encoding

### Steps to Reproduce

1. Check the first link
2. Check the second link
3. [and so on...]

**Expected behavior:** CSVFeedSpider has a separate setting for the encoding or specify param names when calling csviter

**Actual behavior:** Mess

**Reproduces how often:** 100%

### Versions

```
Scrapy       : 2.5.0
lxml         : 4.6.3.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 21.2.0
Python       : 3.9.8 (main, Nov 10 2021, 09:21:22) - [Clang 13.0.0 (clang-1300.0.29.3)]
pyOpenSSL    : 20.0.1 (OpenSSL 1.1.1k  25 Mar 2021)
cryptography : 3.4.7
Platform     : macOS-11.6-x86_64-i386-64bit
```

### Additional context

Please fix it :)

PR #5397, opened 2022-02-06
Update iterators.py

ardaegrioglu:
Hello this is my first time contributing to the project, I changed the places of encoding and quotechar on the initial declaration and ran the tests, there didn't seem to be a issue.


wRAR:
(this is for #5391)

The issue already has a proposed fix, #5394. Your changes also don't fix the issue.

Issue #5404, opened 2022-02-08
tests/test_command_check.py::CheckCommandTest depends on example.com


wRAR:
Sometimes we get

`'OK' not in '/home/runner/work/scrapy/scrapy/.tox/asyncio/lib/python3.10/site-packages/coverage/inorout.py:472: CoverageWarning: --include is ignored because --source is set (include-ignored)\n  self.warn("--include is ignored because --source is set", slug="include-ignored")\nE\n======================================================================\nERROR: [check_spider] parse (errback)\n----------------------------------------------------------------------\nTraceback (most recent call last):\n  File "/home/runner/work/scrapy/scrapy/.tox/asyncio/lib/python3.10/site-packages/twisted/internet/defer.py", line 1660, in _inlineCallbacks\n    result = current_context.run(gen.send, result)\nStopIteration: <404 http://example.com>\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File "/home/runner/work/scrapy/scrapy/scrapy/core/spidermw.py", line 52, in _process_spider_input\n    result = method(response=response, spider=spider)\n  File "/home/runner/work/scrapy/scrapy/scrapy/spidermiddlewares/httperror.py", line 45, in process_spider_input\n    raise HttpError(response, \'Ignoring non-200 response\')\nscrapy.spidermiddlewares.httperror.HttpError: Ignoring non-200 response\n\n----------------------------------------------------------------------\nRan 0 contracts in 0.139s\n\nFAILED (errors=1)\n'`

This seems to be caused by getting 404 from http://example.com which I can reproduce on some requests. We shouldn't make actual network requests in any case, but in this case this is also breaking the test.

wRAR:
Should we mark these as xfail for now?

Gallaecio:
I’m OK with that as a temporary measure until we stop using http://example.com at all. Although it may be better to point them instead to e.g. https://toscrape.com/ , which I imagine will get rid of the issue most of the time (but still, as you said, the end goal should remain not to make network requests if avoidable).

PR #5405, opened 2022-02-08
Fix running tests with Twisted 22.1.0

### LINKED ISSUE #5400: Tests broken with Twisted 22.1.0
`ImportError: cannot import name 'PayloadResource' from 'twisted.web.test.test_webclient'`

`ImportError: cannot import name 'ForeverTakingResource' from 'twisted.web.test.test_webclient'`

wRAR:
Fixes #5400, see the explanation there.

I've decided to put the classes into mockserver.py because it already has a resource class and other two files already import from it. But it's possible to move them into a separate file (in which case I'd discuss if we want to keep those supplementary modules in tests/ or move them into a subdir).

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -21,5 +21,3 @@ addopts =
 markers =
     only_asyncio: marks tests as only enabled when --reactor=asyncio is passed
     only_not_asyncio: marks tests as only enabled when --reactor=asyncio is not passed
-filterwarnings=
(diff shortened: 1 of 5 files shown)


Gallaecio (approved):
Looks good to me, and I see all errors in tests so far are due to https://github.com/scrapy/scrapy/issues/5404, so feel free to merge if the same applies to pypy3-pinned (which I imagine it will).

PR #5406, opened 2022-02-09
Fix SMTP STARTTLS for Twisted >= 21.2.0 (5386)

### LINKED ISSUE #5386: Support TLS1.2 for SMPT Auth
## Summary

The [Mail settings](https://docs.scrapy.org/en/latest/topics/email.html#topics-email-settings) don't have an option to choose a TLS version. Only to enforce upgrading connections to use SSL/TLS.
Mail servers like smtp.office365.com dropped support for TLS1.0 and TLS1.1 and now require TLS1.2: https://techcommunity.microsoft.com/t5/exchange-team-blog/new-opt-in-endpoint-available-for-smtp-auth-clients-still/ba-p/2659652 

It seems that scrapy mail doesn't support TLS1.2. The error message (with `MAIL_TLS = True`):

`[scrapy.mail] Unable to send mail: To=['user@gmail.com'] Cc=[] Subject="Test" Attachs=0- 421 b'4.7.66 TLS 1.0 and 1.1 are not supported. Please upgrade/update your client to support TLS 1.2. Visit https://aka.ms/smtp_auth_tls. [AM6P194CA0047.EURP194.PROD.OUTLOOK.COM]'` 

## Motivation

Without TLS1.2 it's not possible anymore to send mails via smtp.office365.com. An option to use TLS1.2 would fix this issue

TobiMayr:
Pass hostname to the ESMTPSenderFactory when the twisted version is higher than 21.2.0 to use STARTTLS.

Fixes #5386

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -12,7 +12,7 @@ from email.mime.text import MIMEText
 from email.utils import formatdate
 from io import BytesIO
 
-from incremental import Version
+from twisted.python.versions import Version
 from twisted.internet import defer, ssl
 from twisted import version as twisted_version
 
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -12,7 +12,9 @@ from email.mime.text import MIMEText
 from email.utils import formatdate
 from io import BytesIO
 
+from incremental import Version
 from twisted.internet import defer, ssl
+from twisted import version as twisted_version
 
 from scrapy.utils.misc import arg_to_iter
 from scrapy.utils.python import to_bytes
@@ -126,31 +128,22 @@ class MailSender:
                       'mailattachs': nattachs, 'mailerr': errstr})
 
     def _sendmail(self, to_addrs, msg):
-        # Import twisted.mail here because it is not available in python3
         from twisted.internet import reactor
         from twisted.mail.smtp import ESMTPSenderFactory
-        from twisted import version as twisted_version
-        from distutils.version import Version
-
+        msg = BytesIO(msg)
         d = defer.Deferred()
 
-        factory_config = {
-            'username': self.smtpuser,
-            'password': self.smtppass,
-            'fromEmail': self.mailfrom,
-            'toEmail': to_addrs,
-            'file': BytesIO(msg),
-            'deferred': defer.Deferred(),
+        factory_keywords = {
             'heloFallback': True,
             'requireAuthentication': False,
             'requireTransportSecurity': self.smtptls
         }
 
         # Newer versions of twisted require the hostname to use STARTTLS
-        if twisted_version >= Version('21.2.0'):
-            factory_config['hostname'] = self.smtphost
+        if twisted_version >= Version('twisted', 21, 2, 0):
+            factory_keywords['hostname'] = self.smtphost
 
-        factory = ESMTPSenderFactory(**factory_config)
+        factory = ESMTPSenderFactory(self.smtpuser, self.smtppass, self.mailfrom, to_addrs, msg, d, **factory_keywords)
 
         factory.noisy = False
 
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -129,12 +129,19 @@ class MailSender:
         # Import twisted.mail here because it is not available in python3
         from twisted.internet import reactor
         from twisted.mail.smtp import ESMTPSenderFactory
+        from twisted import copyright
+        from packaging import version
         msg = BytesIO(msg)
         d = defer.Deferred()
         factory = ESMTPSenderFactory(
             self.smtpuser, self.smtppass, self.mailfrom, to_addrs, msg, d,
             heloFallback=True, requireAuthentication=False, requireTransportSecurity=self.smtptls,
         )
+
+        # Newer versions of twisted require the hostname to use STARTTLS
+        if version.parse(copyright.version) >= version.parse('21.2.0'):
+            factory.hostname = self.smtphost
+
         factory.noisy = False
 
         if self.smtpssl:
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -129,9 +129,20 @@ class MailSender:
 
     def _sendmail(self, to_addrs, msg):
         from twisted.internet import reactor
-        from twisted.mail.smtp import ESMTPSenderFactory
         msg = BytesIO(msg)
         d = defer.Deferred()
+
+        factory = self._create_sender_factory(to_addrs, msg, d)
+
+        if self.smtpssl:
+            reactor.connectSSL(self.smtphost, self.smtpport, factory, ssl.ClientContextFactory())
+        else:
+            reactor.connectTCP(self.smtphost, self.smtpport, factory)
+
+        return d
+
+    def _create_sender_factory(self, to_addrs, msg, d):
+        from twisted.mail.smtp import ESMTPSenderFactory
 
         factory_keywords = {
             'heloFallback': True,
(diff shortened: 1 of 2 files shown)


Gallaecio:
Is it OK to set the attribute after the class initialization, instead of passing it to the class __init__ method? Could you include some kind of test for the change (a test that would fail if the change was reverted)?

wRAR:
Note that usually we check the Twisted version with `from twisted import version as twisted_version; twisted_version < Version('twisted', 18, 4, 0)`

wRAR:
As for the assignment question, it should be fine BUT the attribute is `_hostname`, not `hostname` so this code is unlikely to work (and I don't like the idea of assigning to an underscored attribute of course). I think we can make a dict for all keyword args we use and pass it.

Not sure what can a test do unless we can mock a SMTP server with STARTTLS, but it may be possible to run some code with just the factory itself (e.g. `buildProtocol`) and examine the results.

TobiMayr:
Thank you for your input, I will try to fix that next week

TobiMayr:
Made the changes but I still need test it with the server and add unit tests

scrapy/mail.py
```diff
@@ -129,12 +129,29 @@ def _sendmail(self, to_addrs, msg):
...
-            heloFallback=True, requireAuthentication=False, requireTransportSecurity=self.smtptls,
-        )
+
+        factory_config = {
+            'username': self.smtpuser,
+            'password': self.smtppass,
+            'fromEmail': self.mailfrom,
+            'toEmail': to_addrs,
+            'file': BytesIO(msg),
+            'deferred': defer.Deferred(),
```
wRAR, line 143: making a separate Deferred here breaks the method

scrapy/mail.py
```diff
@@ -129,12 +129,29 @@ def _sendmail(self, to_addrs, msg):
         # Import twisted.mail here because it is not available in python3
         from twisted.internet import reactor
         from twisted.mail.smtp import ESMTPSenderFactory
-        msg = BytesIO(msg)
+        from twisted import version as twisted_version
```
wRAR, line 132: why local imports?

scrapy/mail.py
```diff
@@ -129,12 +129,29 @@ def _sendmail(self, to_addrs, msg):
...
+        from distutils.version import Version
+
         d = defer.Deferred()
-        factory = ESMTPSenderFactory(
-            self.smtpuser, self.smtppass, self.mailfrom, to_addrs, msg, d,
-            heloFallback=True, requireAuthentication=False, requireTransportSecurity=self.smtptls,
-        )
+
+        factory_config = {
```
wRAR, line 137: I would make a dict only for keyword arguments, not for all of them, but I guess that's a style question

wRAR (changes requested):

scrapy/mail.py
TobiMayr: I did it to conform to the code that was already there. You think it's better to move them all up?

scrapy/mail.py
TobiMayr: I agree, it makes it shorter as well. I will adjust it

scrapy/mail.py
TobiMayr: Didn't know that. Will remove the comment as well

TobiMayr:
I made some code changes implementing the suggestions in your comments. Thank you @wRAR 
I also tried this code with my SFTP server and it works with a newer twisted version (22.1.0). The mail was sent successfully.

What I am struggling with are the unit tests. How can I test the `factory` itself? I need to make a separate method to create the `factory`, right? In `tests/test_mail.py` I tried to execute the `_sendmail` method. But it returns a deferred and I don't know if it's possible to extract the `factory` from there. 
I tried something like:
```
        mailsender = MailSender(debug=False, smtphost='localhost', smtpport=587,
                                mailfrom='scrapy@localhost.com',
                                smtpuser='scrapy@localhost.com', smtppass='password', smtptls=True)

        msg = b'Content-Type: text/plain\nMIME-Version: 1.0\nFrom: scrapy@localhost.com\n' \
              b'To: scrapy@localhost.com\nDate: Wed, 23 Feb 2022 22:03:47 +0100\nSubject: test\n\ntest'
        d = mailsender._sendmail(to_addrs='scrapy@localhost.com', msg=msg)
```
Any ideas?

scrapy/mail.py
```diff
@@ -12,7 +12,9 @@
 from email.utils import formatdate
 from io import BytesIO
 
+from incremental import Version
```
wRAR, line 15: We usually do `from twisted.python.versions import Version` (even though it's the same class)

wRAR:
Regarding tests, I thought about running some methods that cause `ESMTPSender._getContextFactory()` to be called (as this is the method that works differently with and without `self._hostname`)

TobiMayr:
> Regarding tests, I thought about running some methods that cause `ESMTPSender._getContextFactory()` to be called (as this is the method that works differently with and without `self._hostname`)

Wouldn't I then test only the twisted library instead of the `_sendmail` method? Ideally, I should test this method and then analyse the `ESMTPSender` right?

wRAR:
Right, I understand now what did you mean.

Maybe we can extract the factory creation? Otherwise we need  a mock SMTP server, as the purpose of `_sendmail` is to actually send an email, and you can't use it for something more simple.

TobiMayr (commented):
Resolved comment but not all checks ran through. I'm not sure what the issue could be:
```
FAILED tests/test_engine_stop_download_bytes.py::EngineTest::test_deprecated_has_capacity
= 1 failed, 2697 passed, 119 skipped, 26 xfailed, 529 warnings in 1187.61s (0:19:47) =
ERROR: InvocationError for command /home/runner/work/scrapy/scrapy/.tox/pypy3-pinned/bin/pytest --durations=10 docs scrapy tests (exited with code 1)
```
Can anyone point me in the right direction?

Gallaecio:
We have a few tests that may fail randomly. Let me re-run them.

wRAR:
Have you tried to call some more high-level TLS-related methods that check `_hostname` instead of checking it directly? See https://github.com/scrapy/scrapy/pull/5406#issuecomment-1049688762

TobiMayr:
The coverage report tells me that the body of this `if`-block is not covered:
```
if twisted_version >= Version('twisted', 21, 2, 0):
    factory_keywords['hostname'] = self.smtphost
```
It seems like the coverage run is executed with a lower version of twisted.
I can't mock the `twisted_version` inside of the `mail.py`, right? Because it's an import in a different module with its own local scope. My mock wouldn't take effect. Any idea on how I can make sure this line is covered?

Gallaecio:
According to the [pull request coverage report](https://app.codecov.io/gh/scrapy/scrapy/compare/5406/diff) it gets called at least 6 times. So I think we can ignore the codecov/patch issue.

TobiMayr:
@wRAR do you think I should just write a test for `_sendmail` to cover the line of the function call (see previous comment)? I'm not sure if this would be the right way to fix the coverage. I don't see a better way though. What are your thoughts?

wRAR:
To test `_sendmail`, as I mentioned earlier, you would "need a mock SMTP server, as the purpose of _sendmail is to actually send an email, and you can't use it for something more simple." And I suspect this entire function is not currently covered anyway.

TobiMayr:
Right @wRAR. Not sure if mocking a SMTP would be beneficial here. As the only purpose would be to cover this one function call. Do you think the coverage target miss can be ignored then? Since the only other line is apparently being executed by the tests. 
Or is reaching the coverage target crucial? I seriously don't know what do do in this case :)

Gallaecio:
I wonder if we can exploit mitmproxy for a testing SMTP server, as we do for other tests that need a server.

wRAR (approved):

wRAR:
The current test implementation matches what I wanted to see initially, and as I wrote earlier, `_sendmail` was already not covered so it sounds fine if we don't add coverage for it here.

Gallaecio (approved):

PR #5429, opened 2022-03-01
Freeze and upgrade CI packages

Gallaecio:


Gallaecio:
Note to self: Do not create pull requests from the upstream repository ever again (or prevent duplicate jobs in CI somehow).

PR #5432, opened 2022-03-01
Recommend Common Crawl instead of Google Cache

### LINKED ISSUE #3582: Do not recommend Google Cache in the documentation
From the feedback at [Using Scrapy on a Google cache of a website](https://stackoverflow.com/questions/50555545/using-scrapy-on-a-google-cache-of-a-website), I think we should stop [recommending Google Cache in the Scrapy documentation](https://doc.scrapy.org/en/latest/topics/practices.html#avoiding-getting-banned).

We should review the terms of use of the Wayback Machine from archive.org, and see if we can recommend that instead.

Farsene1:
Closes #3582

The terms of use allow scraping as long as Scrapy honours the restrictions of robot.txt files and NOFOLLOW metatags
Common Crawl terms of use: https://commoncrawl.org/terms-of-use/


Gallaecio (approved):

Gallaecio:
Thanks!

wRAR (approved):

PR #5439, opened 2022-03-06
Document the Scrapy component API

### LINKED ISSUE #5110: Document the Scrapy component API
There are certain classes in Scrapy that may optionally define the `from_crawler` and `from_settings` class methods.

At the moment, we document this fact in every single one of them. We repeat ourselves a lot. But we also often fail to cover how to properly subclass a class that defines one of these methods.

I think we need to extend the documentation with a section or a page about this type of classes, and have all such classes reference this central documentation instead of providing their own documentation about this aspects of such class.

OrestisKan:
Remove duplicate definition of from_crawler and from_settings method. Make them refer to the centralized definitions in class-methods.rst where the explanation of the methods is elaborated. Closes #5110

### DIFF
--- a/docs/topics/api.rst
+++ b/docs/topics/api.rst
@@ -167,16 +167,6 @@ SpiderLoader API
     the :class:`scrapy.interfaces.ISpiderLoader` interface to guarantee an
     errorless execution.
 
-    .. method:: from_settings(settings)
-
-       This class method is used by Scrapy to create an instance of the class.
-       It's called with the current project settings, and it loads the spiders
-       found recursively in the modules of the :setting:`SPIDER_MODULES`
-       setting.
-
-       :param settings: project settings
-       :type settings: :class:`~scrapy.settings.Settings` instance
-
     .. method:: load(spider_name)
 
        Get the Spider class with the given name. It'll look into the previously
@@ -197,6 +187,13 @@ SpiderLoader API
 
        :param request: queried request
        :type request: :class:`~scrapy.Request` instance
+
+    :meth:`from_settings`:
+
+       This class method is used by Scrapy to create an instance of the class.
+       It's called with the current project settings, and it loads the spiders
+       found recursively in the modules of the :setting:`SPIDER_MODULES`
+       setting.
 
 .. _topics-api-signals:
 
--- a/docs/topics/class-methods.rst
+++ b/docs/topics/class-methods.rst
@@ -1,0 +1,29 @@
+===========================
+Class Instantiation Methods
+===========================
+
+These methods create an instance of their implementer class by 
+extracting the components needed for it from the argument that the method takes.
+
+.. py:classmethod:: from_crawler(cls, crawler)
+
+    Factory method that if present, is used to create an instance of the implementer class
+    using a :class:`~scrapy.crawler.Crawler`. It must return a new instance
+    of the implementer class. The Crawler object is needed in order to provide 
+    access to all Scrapy core components like settings and signals; It is a 
+    way for the implenter class to access them and hook its functionality into Scrapy.
+
+    :param crawler: crawler that uses this middleware
+    :type crawler: :class:`~scrapy.crawler.Crawler` object
+
+
+.. py:classmethod:: from_settings(cls, settings)
+
+    This class method is used by Scrapy to create an instance of the implementer class
+    using the settings passed as arguments.
+    In most implementations of this method, not all settings given will be used 
+    to instantiate the class, but only the ones needed in the specific scenario.
+
+
+    :param settings: project settings
+    :type settings: :class:`~scrapy.settings.Settings` instance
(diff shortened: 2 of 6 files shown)


OrestisKan:
@Gallaecio Is there anything missing or anything that I could improve?

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,29 @@
+===========================
+Class Instantiation Methods
+===========================
+
+These methods create an instance of their implementer class by 
+extracting the components needed for it from the argument that the method takes.
```
Gallaecio, line 6: I think, before we introduce the methods themselves, we need to give some background and provide a list of Scrapy objects that support this.

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,29 @@
...
+
+.. py:classmethod:: from_crawler(cls, crawler)
+
+    Factory method that if present, is used to create an instance of the implementer class
```
Gallaecio, line 10: We need to be consistent with vocabulary. We should call them either ‘instantiation methods’ or ‘factory methods’, but we should stick to 1 name. I am slightly in favor of the later, but I have no strong opinion on it.

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,29 @@
...
+.. py:classmethod:: from_settings(cls, settings)
+
+    This class method is used by Scrapy to create an instance of the implementer class
+    using the settings passed as arguments.
+    In most implementations of this method, not all settings given will be used 
+    to instantiate the class, but only the ones needed in the specific scenario.
+
+
+    :param settings: project settings
+    :type settings: :class:`~scrapy.settings.Settings` instance
```
Gallaecio, line 29: This page does not cover one of the aspects from #5110:

> But we also often fail to cover how to properly subclass a class that defines one of these methods.

I am OK with merging an initial version of the page that does not go into that much detail, and instead is limited to covering the information already in the documentation in a centralized matter, but in that case this pull request should not be marked as closing #5110.

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,29 @@
...
+    :param crawler: crawler that uses this middleware
+    :type crawler: :class:`~scrapy.crawler.Crawler` object
+
+
+.. py:classmethod:: from_settings(cls, settings)
+
+    This class method is used by Scrapy to create an instance of the implementer class
+    using the settings passed as arguments.
+    In most implementations of this method, not all settings given will be used 
+    to instantiate the class, but only the ones needed in the specific scenario.
```
Gallaecio, line 25: I don’t think this bit is needed. It is up to the user which settings they want to use, and using a lot or 1 should make no difference.

On the other hand, in this definition I miss the fact that this class method will not be called at all if `from_crawler` is also defined, which I believe it a rather important detail.

OrestisKan:
@Gallaecio the changes that you requested are done, please let me know if there is anything I can improve

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,123 @@
+===========================
+Class Factory Methods
+===========================
```
Gallaecio, line 3:
```suggestion
=====================
Class Factory Methods
=====================
```

docs/topics/api.rst
```diff
@@ -198,6 +188,13 @@ SpiderLoader API
        :param request: queried request
        :type request: :class:`~scrapy.Request` instance
 
+    :meth:`from_settings`:
+
+       This class method is used by Scrapy to create an instance of the class.
+       It's called with the current project settings, and it loads the spiders
+       found recursively in the modules of the :setting:`SPIDER_MODULES`
+       setting.
```
Gallaecio, line 196: I am thinking factory method signatures should still be kept in the API reference. It is specially important because some components may get extra arguments, I believe, and in those cases we should document them in one of the methods (e.g. `from_crawler`) and in the other method refer to the former for parameter details.

What we can do now is simplify the description, both of `from_crawler` and `from_settings`, to something like:

```
Factory method. See :ref:`class-methods`.
```

Assuming we also add `.. _class-methods:` at the beginning of the new topic.

Also, I imagine (please check) that this class also supports `from_crawler`. In that case, we should include an entry for that class method as well, with its own signature, but the same description that basically points to the new topic.

docs/topics/class-methods.rst
```diff
@@ -0,0 +1,123 @@
...
+    * ItemPipeline
+    * DownloaderMiddleware
+    * SpiderMiddleware
+    * Scheduler
+    * BaseScheduler
+    * Spider
+
+The ``from_settings`` class method is implemented in the following objects:
+    * MailSender
+    * SpiderLoader
```
Gallaecio, line 21: Some of these are classes, some are interface definitions (they define an API, but do not really exist as classes in Python code), but none of them are objects.

This also seems to suggest that the upper list supports only `from_crawler` and the lower list supports only `from_settings`. Is that so?

Gallaecio (commented):
Sorry for the late feedback. I will try to be quicker for any further feedback.

Issue #5437, opened 2022-03-03
CLOSESPIDER_TIMEOUT problem.


shahrom322:
### Description

When switching from version 2.5.1 to 2.6.1, there was a problem with the parser terminating if the shutdown condition was CLOSESPIDER_TIMEOUT.

### Steps to Reproduce

1. pip install -U scrapy
2. scrapy crawl --set 'CLOSESPIDER_TIMEOUT=1' some_crawler

**Expected behavior:** [What you expect to happen]

2022-03-03 16:09:30 [scrapy.core.engine] INFO: Spider closed (closespider_timeout)

**Actual behavior:** [What actually happens]

2022-03-03 16:09:30 [scrapy.core.engine] INFO: Spider closed (closespider_timeout)
2022-03-03 16:09:30 [scrapy.core.engine] INFO: Error while scheduling new request
Traceback (most recent call last):
  File "c:\users\onedrive\python\scrapy_test\venv\lib\site-packages\twisted\internet\task.py", line 528, in _oneWorkUnit
    result = next(self._iterator)
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\users\onedrive\python\scrapy_test\venv\lib\site-packages\twisted\internet\defer.py", line 858, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "c:\users\onedrive\python\scrapy_test\venv\lib\site-packages\scrapy\core\engine.py", line 187, in <lambda>
    d.addBoth(lambda _: self.slot.nextcall.schedule())
AttributeError: 'NoneType' object has no attribute 'nextcall'

**Reproduces how often:** [What percentage of the time does it reproduce?]

It always reproduce

### Versions

Scrapy       : 2.6.1
lxml         : 4.7.1.0
libxml2      : 2.9.12
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 21.7.0
Python       : 3.9.6 (tags/v3.9.6:db3ff76, Jun 28 2021, 15:26:21) [MSC v.1929 64 bit (AMD64)]
pyOpenSSL    : 22.0.0 (OpenSSL 1.1.1m  14 Dec 2021)
cryptography : 36.0.1
Platform     : Windows-10-10.0.19044-SP0

Laerte:
I was able to reproduce this issue using this command:
```
[...]

PR #5440, opened 2022-03-07
engine: prevent slot method call after unsetting the slot

Gallaecio:
Fixes: #5437

I must say, however, that I do not fully understand the code involved, and I suspect there must be a cleaner way to address this.

To-Do:

- [ ] ~Find out if it is possible to implement a cleaner approach~
- [x] Implement a test

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -136,7 +136,9 @@ class ExecutionEngine:
         self.paused = False
 
     def _next_request(self) -> None:
-        assert self.slot is not None  # typing
+        if self.slot is None:
+            return
+
         assert self.spider is not None  # typing
 
         if self.paused:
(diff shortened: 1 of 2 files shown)


scrapy/core/engine.py
```diff
@@ -136,7 +136,9 @@ def unpause(self) -> None:
         self.paused = False
 
     def _next_request(self) -> None:
-        assert self.slot is not None  # typing
+        if self.slot is None:
+            return
```
Gallaecio, line 140: This is necessary because `close_spider` uses `setattr` to set `self.slot` to `None`. However, I wonder if we can m,ake it so that `_next_request` in not called in the first place, in those scenarios.

scrapy/core/engine.py
```diff
@@ -184,7 +186,8 @@ def _next_request_from_scheduler(self) -> Optional[Deferred]:
         d.addErrback(lambda f: logger.info('Error while removing request from slot',
                                            exc_info=failure_to_exc_info(f),
                                            extra={'spider': self.spider}))
-        d.addBoth(lambda _: self.slot.nextcall.schedule())
+        slot = self.slot
+        d.addBoth(lambda _: slot.nextcall.schedule())
```
Gallaecio, line 190: This is also necessary for the same reason, because `close_spider` uses `setattr` to set `self.slot` to `None`. However, I wonder if we can handle this differently, or somehow prevent that `setattr` call to happen before this code is called.

Gallaecio:
I have implemented a subprocess-based test because it was taking me too long to figure out a test based on inline callbacks.

wRAR (approved):

Issue #5444, opened 2022-03-08
-o - broken in 2.6


wRAR:
`-o -:json` should be supported and means "output to stdout in the JSON format".

```
$ scrapy runspider sp.py -o -:json
Usage
=====
  scrapy runspider [options] <spider_file>
runspider: error: argument -o/--output: expected one argument
```

This works in 2.5 and was most likely broken after moving to argparse in #5374. I suspect it tries to interpret `-:json` as a separate option. Interesting that `-o -` isn't broken and correctly prints "Unrecognized output format ''"

PR #5445, opened 2022-03-08
Fix -o - broken in 2.6

Laerte:
Closes: #5444 

This happen because `argparse` module treats the `-` as prefix chars: https://github.com/python/cpython/blob/main/Lib/argparse.py#L2205:L2207

Since we cannot change the internal implementation of `_parse_optional` i created a class that inherits from `ArgumentParser` class and override the method with our use case calling the default `_parse_optional`.

### DIFF
--- a/scrapy/cmdline.py
+++ b/scrapy/cmdline.py
@@ -14,6 +14,15 @@ from scrapy.utils.project import inside_project, get_project_settings
 from scrapy.utils.python import garbage_collect
 
 
+class ScrapyArgumentParser(argparse.ArgumentParser):
+    def _parse_optional(self, arg_string):
+        # if starts with -: it means that is a parameter not a argument
+        if arg_string[:2] == '-:':
+            return None
+
+        return super()._parse_optional(arg_string)
+
+
 def _iter_command_classes(module_name):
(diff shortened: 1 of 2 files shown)


wRAR (approved):

Gallaecio:
@Laerte Could you please rebuild your branch on top of 2.6? This is a change we would like to release in 2.6.2.

There is probably a better way, but this should do the trick:

```
git remote add upstream https://github.com/scrapy/scrapy.git  # unless you have already setup this remote before
git fetch upstream
git checkout upstream/2.6 -B master  # Warning: this destroys your branch contents locally; mind that if you add further commits
git cherry-pick 7d20ce65479cdbe39fc17b7cb2dc99415c670dae  # Cherry-picks your commit here; if you had added more commits before doing this, you should note their hashes and do the same with them
git push -f  # Pushes your new master branch, which is upstream/2.6 + your commit, to your master branch, dropping the 2 upstream master commits that are not in upstream 2.6
```

Laerte:
@Gallaecio Done! Tks. 🥇

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5448, opened 2022-03-15
Add release notes for 2.6.2

Gallaecio:


pawelmhm:
Hi @Gallaecio @wRAR it is little confusing, but we have release notes here: https://docs.scrapy.org/en/latest/news.html#scrapy-2-6-2-2022-03-15 but there is no actual pypi release, most recent version is 2.6.1: https://pypi.org/project/Scrapy/#history is this a known issue or something went wrong somewhere?

Gallaecio:
There has been no release. It seems like in Read The Docs `latest` shows the content of `master` and `master` the content of `latest`. I need to look into that.

Gallaecio:
Ah, no, `latest` shows the 2.6 branch, and so does `2.6`. We should make it clear that it is not released.

czechnology:
Is there still a plan to release 2.6.2 or was that version abandoned altogether?

Gallaecio:
Still planned, but no ETA other than “before 2.7”.

PR #4066, opened 2019-10-07
Add .dmg, .iso & .apk to ignored other extensions

akhterwahab:
I run into downloading large files while scraping

Fixes #1837, fixes #2067

PR #5450, opened 2022-03-16
fixed detection of extensions like ".tar.gz" in URL

kinoute:
#4066 added more extensions to ignore by default such as `.rar` or `.zip`, which is great. 

Unfortunately, the addition of `.tar.gz` as an extension to exclude is not working because of the implementation of the function `url_has_any_extension` which can be found in `scrapy/utils/url.py`.

The actual function looks like this:

```python3
def url_has_any_extension(url, extensions):
    return posixpath.splitext(parse_url(url).path)[1].lower() in extensions
```

Here is an example that does not work:
```python3
import posixpath
from scrapy.utils.url import parse_url

parsed_url = urlparse("https://dl.google.com/go/go1.7rc3.freebsd-amd64.tar.gz")
# ParseResult(scheme='https', netloc='dl.google.com', path='/go/go1.7rc3.freebsd-amd64.tar.gz', params='', query='', fragment='')

url_has_any_extension(parsed_url, [".tar.gz"])
# False

# Breaking down "url_has_any_extension" function
posixpath.splitext(parse_url(parsed_url).path)[1].lower() in [".tar.gz"]
# False

posixpath.splitext(parse_url(parsed_url).path)[1].lower()
# '.gz'

posixpath.splitext(parse_url(parsed_url).path)
# ('/go/go1.7rc3.freebsd-amd64.tar', '.gz')
```

We can see that the function splits the path from the first dot encountered, starting from the right, which leads to a `.tar.gz` extension not detected.

I fixed the function by making use of `any` and the standard Python method `endswith`.

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -5,7 +5,6 @@ library.
 Some of the functions that used to be imported from this module have been moved
 to the w3lib.url module. Always import those from there instead.
 """
-import posixpath
 import re
 from urllib.parse import ParseResult, urldefrag, urlparse, urlunparse
 
@@ -31,8 +30,8 @@ def url_is_from_spider(url, spider):
 
 
 def url_has_any_extension(url, extensions):
-    return posixpath.splitext(parse_url(url).path)[1].lower() in extensions
-
+    """Return True if the url ends with one of the extensions provided"""
+    return any(parse_url(url).path.lower().endswith(ext) for ext in extensions)
 
 def parse_url(url, encoding=None):
     """Return urlparsed url from the given argument (which could be an already
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -31,7 +31,8 @@ def url_is_from_spider(url, spider):
 
 def url_has_any_extension(url, extensions):
     """Return True if the url ends with one of the extensions provided"""
-    return any(parse_url(url).path.lower().endswith(ext) for ext in extensions)
+    lowercase_path = parse_url(url).path.lower()
+    return any(lowercase_path.endswith(ext) for ext in extensions)
 
 def parse_url(url, encoding=None):
     """Return urlparsed url from the given argument (which could be an already
--- a/tests/test_utils_url.py
+++ b/tests/test_utils_url.py
@@ -1,6 +1,8 @@
 import unittest
 
+from scrapy.linkextractors import IGNORED_EXTENSIONS
 from scrapy.spiders import Spider
+from scrapy.utils.misc import arg_to_iter
 from scrapy.utils.url import (
     add_http_if_no_scheme,
     guess_scheme,
@@ -8,8 +10,8 @@ from scrapy.utils.url import (
     strip_url,
     url_is_from_any_domain,
     url_is_from_spider,
+    url_has_any_extension,
 )
-
 
 __doctests__ = ['scrapy.utils.url']
 
@@ -80,6 +82,15 @@ class UrlUtilsTest(unittest.TestCase):
         self.assertTrue(url_is_from_spider('http://www.example.org/some/page.html', MySpider))
         self.assertTrue(url_is_from_spider('http://www.example.net/some/page.html', MySpider))
         self.assertFalse(url_is_from_spider('http://www.example.us/some/page.html', MySpider))
+
+    def test_url_has_any_extension(self):
+        deny_extensions = {'.' + e for e in arg_to_iter(IGNORED_EXTENSIONS)}
+        self.assertTrue(url_has_any_extension("http://www.example.com/archive.tar.gz", deny_extensions))
+        self.assertTrue(url_has_any_extension("http://www.example.com/page.doc", deny_extensions))
+        self.assertTrue(url_has_any_extension("http://www.example.com/page.pdf", deny_extensions))
+        self.assertFalse(url_has_any_extension("http://www.example.com/page.htm", deny_extensions))
+        self.assertFalse(url_has_any_extension("http://www.example.com/", deny_extensions))
+        self.assertFalse(url_has_any_extension("http://www.example.com/page.doc.html", deny_extensions))
 
 
 class AddHttpIfNoScheme(unittest.TestCase):


Gallaecio:
I think we could instead rename `tar.gz` to `gz`. It will support additional scenarios that we probably want to support, while avoiding an implementation change that seems slightly more CPU-expensive.

kinoute:
> I think we could instead rename `tar.gz` to `gz`.

`tar.gz` is not the only file extension that is made out of two words. I don't think it is a good idea to only "fix" this for that extension. Example of other weird extensions:
* .tar.Z, 
* .tar.bz2,
* .tar.lz,
* .tar.xz,
* .tar.zst

Source: https://en.wikipedia.org/wiki/List_of_archive_formats#Archiving_and_compression

And on the same page we can find Pacman file extensions such as `.pkg.tar.zst`, `.pkg.tar` etc. I'm pretty sure there are plenty of file extensions like this in the wild.

> It will support additional scenarios that we probably want to support, while avoiding an implementation change that seems slightly more CPU-expensive.

The impact is minimal or even absent:

```python3
def url_has_any_extension(url, extensions):
    return posixpath.splitext(parse_url(url).path)[1].lower() in extensions

def url_has_any_extension_fixed(url, extensions):
    return any(parse_url(url).path.lower().endswith(ext) for ext in extensions)

parsed_url = urlparse("https://dl.google.com/go/go1.7rc3.freebsd-amd64.tar.gz")

%timeit url_has_any_extension(parsed_url, IGNORED_EXTENSIONS)
# 3.7 µs ± 150 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit url_has_any_extension_fixed(parsed_url, IGNORED_EXTENSIONS)
# 4.08 µs ± 123 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

**Comment by Gallaecio** [comment]
Even when we discard the performance argument, in all the extension scenarios that you mention discarding the last part of the extension should bring the desired behavior.

For it to be necessary to switch to the more robust implementation that you propose, we would need to find an extension “end leaf” that we want to _not_ exclude while _do_ excluding a multi-part extension ending on it. e.g. If we wanted to exclude `.foo.bar` but not exclude `.bar`. I cannot think of a real-life pair of extensions that would meet this criteria, when you want to ignore `.bar` you usually want to ignore `.foo.bar` as well.

That said, given the performance hit is not that significant, I am not against the change. Even if it is not useful for the extensions that we currently ignore by default, someone may find it useful if specifying custom extensions to ignore.

**Comment by kinoute** [comment]
@Gallaecio I added a few tests cases

File: scrapy/utils/url.py [hunk 1]
```diff
@@ -31,8 +30,8 @@ def url_is_from_spider(url, spider):
 
 
 def url_has_any_extension(url, extensions):
-    return posixpath.splitext(parse_url(url).path)[1].lower() in extensions
-
+    """Return True if the url ends with one of the extensions provided"""
+    return any(parse_url(url).path.lower().endswith(ext) for ext in extensions)
```
Gallaecio, line 34:
```suggestion
    lowercase_path = parse_url(url).path.lower()
    return any(lowercase_path.endswith(ext) for ext in extensions)
```

Gallaecio (commented):
I have just thought of a minor improvement, but otherwise the change looks good to me.

Gallaecio (approved):
+1 from me. We shall merge as soon as a second maintainer approves the change (mind that our time is quite limited at the moment, so it might take a while).

Note: I am aware of at least 1 behavior change introduced here because we stop using `splitext`, which is that file names starting with a period that match in their entirety an extension are not considered to match the extension. For example, a file named `.zip` is considered to have the `.zip` extension, whereas for `splitext` that is a file named `.zip` with no extension.

I do not think that is much of an issue, though; i.e. it is OK if someone wants to address that in the future, but I think we can live with that.

tests/test_utils_url.py
```diff
@@ -81,6 +83,15 @@ class MySpider(Spider):
         self.assertTrue(url_is_from_spider('http://www.example.net/some/page.html', MySpider))
         self.assertFalse(url_is_from_spider('http://www.example.us/some/page.html', MySpider))
 
+    def test_url_has_any_extension(self):
+        deny_extensions = {'.' + e for e in arg_to_iter(IGNORED_EXTENSIONS)}
```
kmike, line 87: arg_to_iter is not needed here

kmike:
Thanks @kinoute!

> Even if it is not useful for the extensions that we currently ignore by default, someone may find it useful if specifying custom extensions to ignore.

`http://www.example.com/archive.tar.gz` test fails in master, but passes here, so it seems there is a bug fix.

kmike:
Hmm, I think there might be a performance issue. First, I can't reproduce the test in https://github.com/scrapy/scrapy/pull/5450#issuecomment-1069370085: the new version is ~4x slower. But this benchmark is very optimistic - match is found very early in the list of ignored extensions. Probably a better benchmark would be for the case where extension is not matched, and for this case the new implementation is ~50x slower:

```py
In [19]: %timeit url_has_any_extension_fixed(url+"foobar", IGNORED_EXTENSIONS)
209 µs ± 386 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [21]: %timeit url_has_any_extension(url+"foobar", IGNORED_EXTENSIONS)
4.44 µs ± 9.37 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

kinoute:
@kmike I'm disappointed but I guess you're right, if there is a performance issue, it needs to be fixed. As a matter of fact, I think the performance issue could be that I pass a generator to `any` instead of a list. See this post: https://stackoverflow.com/a/44803103/6373370

Could you check again with a list?

kinoute:
Using Scrapy master, Python 3.9.13 on a MacBook Pro M1, here is what I'm getting:

```python3
import posixpath
from scrapy.utils.url import parse_url
from scrapy.linkextractors import IGNORED_EXTENSIONS

def url_has_any_extension(url, extensions):
    return posixpath.splitext(parse_url(url).path)[1].lower() in extensions

def url_has_any_extension_fixed(url, extensions):
    lowercase_path = parse_url(url).path.lower()
    return any(lowercase_path.endswith(ext) for ext in extensions)

def url_has_any_extension_fixed_new(url, extensions):
    lowercase_path = parse_url(url).path.lower()
    return any(True for ext in extensions if lowercase_path.endswith(ext))
parsed_url = urlparse("https://dl.google.com/go/go1.7rc3.freebsd-amd64.tar.gz" + "foobar")

%timeit url_has_any_extension(parsed_url, IGNORED_EXTENSIONS)
1.31 µs ± 0.711 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit url_has_any_extension_fixed(parsed_url, IGNORED_EXTENSIONS)
7.04 µs ± 33.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit url_has_any_extension_fixed_new(parsed_url, IGNORED_EXTENSIONS)
5.71 µs ± 2.14 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Still slower but a little better.

Gallaecio:
Regarding performance, I get similar results to @kinoute on my system (Python 3.10.5, AMD Ryzen 9 3900X 12-Core):

```
100000 loops, best of 5: 2.69 usec per loop
50000 loops, best of 5: 8.21 usec per loop
50000 loops, best of 5: 6.71 usec per loop
50000 loops, best of 5: 6.52 usec per loop
```

Similar results also when I switch to Python 3.7.13.

Gallaecio:
As for the behavior change regarding input extensions without a dot, while a valid concern, I think it would be best to not address it, as adding and extra `ext.startswith(".")` to the old-school implementation seems to make it almost twice as slow:

```python
# benchmark.py

from scrapy.utils.url import parse_url
from scrapy.linkextractors import IGNORED_EXTENSIONS

def url_has_any_extension_fixed_oldschool_ensure_dot(url, extensions):
    lowercase_path = parse_url(url).path.lower()
    for ext in extensions:
        if ext.startswith(".") and lowercase_path.endswith(ext):
            return True
    return False

url = "https://dl.google.com/go/go1.7rc3.freebsd-amd64.tar.gzfoobar"
parse_url(url)  # Do not slow down the first call later
deny_extensions = [f".{ext}" for ext in IGNORED_EXTENSIONS]
```

```bash
# benchmark.sh

python -m timeit -s "import benchmark" "benchmark.url_has_any_extension_fixed_oldschool_ensure_dot(benchmark.url, benchmark.deny_extensions)"
```

```
20000 loops, best of 5: 10.6 usec per loop
```

Since the bug requires bad user input, I think a proper docstring for the function, explaining that the iterable of extensions to deny needs the extensions with the dot included, should be all we need to do. If users complain, we can politely ask them to read the docs and explain what such an extra check could mean performance-wise.

In fact, it could be considered a feature, not a bug. For example, you could ignore [files ending in `~`](https://unix.stackexchange.com/a/76192/26193).

kmike:
hey @kinoute @Gallaecio I tried to reproduce it again (https://gist.github.com/kmike/1fd10869a1af9a54cddbeca38694454a), and can't. I think you're right, and we should keep it. Thanks @kinoute!

PR #5451, opened 2022-03-17
HTML Conventions

FJMonteroInformatica:
I have updated the HTML files following SonarCloud's recommendations.

### DIFF
--- a/docs/_static/selectors-sample1.html
+++ b/docs/_static/selectors-sample1.html
@@ -1,16 +1,17 @@
-<html>
- <head>
-  <base href='http://example.com/' />
-  <title>Example website</title>
- </head>
- <body>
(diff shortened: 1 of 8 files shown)


Gallaecio:
I don’t get why, but I have no argument against it either.

FJMonteroInformatica:
Hello, this is my first PR. Is the failure in the tests necessarily an error in the modification I have entered? It seems strange to me because they are very small changes.

Gallaecio:
Sadly we have a few tests in Scrapy that fail randomly. Do not worry about it.

Gallaecio (approved):

kmike:
I'm also unsure what's the advantage, but the changes look fine, let's merge them :)

PR #5452, opened 2022-03-19
removed unnecessary parameters in _DictProxy.__iter__ following SonarCloud recommendations

DrownedTales:
I've noticed that the method _iter_(self, k, v) in _DictProxy (Scrapy/Settings/_init_.py) had three arguments. This method is called by python with only the self parameter so I think this was unnecesary. Correct me if I'm wrong.

### DIFF
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -425,7 +425,7 @@ class _DictProxy(MutableMapping):
     def __delitem__(self, k):
         del self.o[k]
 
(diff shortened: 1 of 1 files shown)


elacuesta:
It's an underscored class, and it's not used anywhere. My vote goes to just remove it.

wRAR:
Its usage was removed in 4f6778aa7332aecd15f7672a0852d1f49596969b (1.6.0)

PR #5453, opened 2022-03-20
Change SSLv3 method to TLSv1_2 method

Farsene1:
SSLv3 is an older version of the protocol. Replacing SSLv3 with TLSv1.2 that is considered to be secure at the moment of making this pull request.


wRAR:
> Change SSLv3 method

> SSLv3 is an older version of the protocol.

It's not SSLv3. The variable is named `SSLv23_METHOD` and, confusingly, means version negotiation, which is written e.g. 7 lines above your first changed line.

PR #5455, opened 2022-03-21
fixed typo in httpcache docs

VMRuiz:
Fix small typo in the docs about number of storage backends available by default with Scrapy

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -366,7 +366,7 @@ HttpCacheMiddleware
     This middleware provides low-level cache to all HTTP requests and responses.
     It has to be combined with a cache storage backend as well as a cache policy.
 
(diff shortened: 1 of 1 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -366,7 +366,7 @@ HttpCacheMiddleware
     This middleware provides low-level cache to all HTTP requests and responses.
     It has to be combined with a cache storage backend as well as a cache policy.
 
-    Scrapy ships with three HTTP cache storage backends:
+    Scrapy ships with two HTTP cache storage backends:
```
Gallaecio, line 369: What about making sure we do not screw it up again?

```suggestion
    Scrapy ships with the following HTTP cache storage backends:
```

Murphy’s law: we will at some point switch to a single storage backend, and forget to change “backends” to “backend”.

Gallaecio (approved):

PR #3799, opened 2019-05-27
Fix for issue #3798 LinkExtractor with Unique = False doesn't extract fully identical Links

Ksianka:
Fixes #3798

Issue #3798, opened 2019-05-27
LinkExtractor with Unique = False doesn't extract fully identical Links


Ksianka:
Case: 
unexpected behavior identified for LinkExctractor with Unique = False if the page contains fully identical links (the same URL and text).
The current result returns one unique link instead of two or more identical ones.
For example:

1) Local http file 

```<your_path>Test_file.html
<html>
<body>
<a href='sample3.html'>sample 3 repetition</a>
<a href='sample3.html'>sample 3 repetition</a>
</body>
</html>```

2) Scrapy test code:
ignore the error about <GET file:///robots.txt>  as we use local file

`import scrapy
from scrapy.linkextractors import LinkExtractor

class QuotesSpider(scrapy.Spider):
    name = “test”
    start_urls = [
        'file:///<insert_your_local_path>/Test_file.html',
    ]

def __init__(self, *args, **kwargs):
    super(QuotesSpider, self).__init__(*args, **kwargs)
    self.le = LinkExtractor(unique=False)

def parse(self, response):
    links = self.le.extract_links(response)
    yield {'extract_link': links}`

3) run the code: scrapy crawl test, and get the result with one link:

{'extract_link': [Link(url='file:///<your_local_path>/sample3.html', text='sample 3 repetition', fragment='', nofollow=False)]}

Instead of the result with two links from HTML:

{'extract_link': [Link(url='file:///<your_local_path>/sample3.html', text='sample 3 repetition', fragment='', nofollow=False), 
Link(url='<your_local_path>/sample3.html', text='sample 3 repetition', fragment='', nofollow=False)]}

Ksianka:
Pre-solution: 
the issue is driven by the public method exctract_link of LxmlLinkExtractor(the same as LinkExctractor) which returns unique_list(all_links), see line 130 in module lxmlhtml.py, instead of all_links (similar to solution in sgml.py).

However, here we get another issue:
It looks like the original idea of this unique_list(all_links) was to avoid duplication of links if there is a repetition of tags in restrict_css and restrict_xpaths stated by user In LinkExctractor instance.  
For example: self.le =   ##LinkExtractor(unique=False, restrict_css=('a', 'a'), restrict_xpaths=('a', ‘a’)).

The duplication of  restrict_css and restrict_xpaths  is prosed to be corrected by applying function unique_list() to joined/combined list of css and x_paths in __init__.py module of class FilteringLinkExtractor(object).
Code for correction:

[...]

PR #5458, opened 2022-03-23
fix: return unique_list only when link_extractor.unique is True

reidemeister94:
In the `lxmlhtml.py` script, the function `extract_links(self, response)` always returns `unique_list(all_links)`, even when the attribute `unique` of the `link_extractor` is equal to `False`. 

The behavior is coherent with the `unique` attribute with this little change in the return line.

Fixes https://github.com/scrapy/scrapy/issues/3798, closes https://github.com/scrapy/scrapy/pull/3799, closes https://github.com/scrapy/scrapy/pull/4695


wRAR:
See also: #3798 #4695 #3799

reidemeister94:
Ok so this won't get fixed?

reidemeister94:
Any update on this?

Gallaecio:
Added to my review queue, but it may take a while :bow:

reidemeister94:
Just a ping :D

Gallaecio:
I am OK with merging as is, provided we update the docstring of `extract_links` and provide a test for the change.

reidemeister94:
I updated the docstring and fixed the html used for the tests to have tests that also cover the new case (unique = false, and the output contains duplicates). 
I hope that's enough for being merged :D

Gallaecio (approved):
Until your tests changes I did not realize the issue only affected links with matching URL *and* text, but the original issue indeed describes it like that.

Thanks!

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -195,7 +195,8 @@ def extract_links(self, response):
         Only links that match the settings passed to the ``__init__`` method of
         the link extractor are returned.
 
-        Duplicate links are omitted.
+        Duplicate links are omitted if the ``unique`` parameter is set to ``True``,
```
Gallaecio, line 198:
```suggestion
        Duplicate links are omitted if the ``unique`` attribute is set to ``True``, 
```

wRAR (approved):

Issue #5454, opened 2022-03-21
test_proxy_connect.py hangs with mitmproxy 8.0.0 (affecting 3.8 tests)


wRAR:
mitmproxy 8.0.0 was released 2 days ago, it's likely to be the cause of this.

Also it looks like we currently don't install mitmproxy at all for 3.9+.

wRAR:
(my first idea was pinning mitmproxy to < 8, but ideally we should fix the underlying problem if possible, for cases when installing an older version is not possible)

wRAR:
The part that hangs is `self.proc.stdout.readline()` at https://github.com/scrapy/scrapy/blob/8c3e5a2421f639f36e9b907d69d4d48b86447e68/tests/test_proxy_connect.py#L41

I tried this code in a debugger and the first call to `.stdout.readline()` (or to `.communicate()`) indeed hangs, but if you cancel it with Ctrl-C and call it again, it immediately returns the expected stdout line. I don't know what does this mean.

wRAR:
> Also it looks like we currently don't install mitmproxy at all for 3.9+.

FTR: it's now possible to install `mitmproxy >= 5.3.0, < 8` on 3.9+, but it requires a very old `cryptography < 3.5`, so we shouldn't do this. This would be a problem for 3.8 too, but we only run the pinned targets on 3.8 and those don't install mitmproxy. So it looks like nothing runs tests/test_proxy_connect.py, we should fix this issue (or use something else than mitmproxy) to run it again.

wRAR:
Bisected to https://github.com/mitmproxy/mitmproxy/pull/5188 which vendors `click.style`.

PR #5459, opened 2022-03-25
Pin mitmproxy to < 8 for now.

wRAR:
Related to #5454 

Should this go into 2.6.2?


Gallaecio:
> Should this go into 2.6.2?

Makes sense to me, we definitely need to run tests as part of the release process and it would be great for them not to break :)

Gallaecio (approved):

wRAR:
Rebased onto 2.6

wRAR:
Should this be merged?

Gallaecio:
I am porting this to the main branch as well.

Issue #5431, opened 2022-03-01
Improve cookie handling


Gallaecio:
There are different aspects of cookie handling in Scrapy that we should improve. This issue aims to centralize a set of improvements that could be addressed as part of a Google Summer of Code project.

-   Implement the latest standard of cookies, the one web browsers use.

    This cannot be done with the Python standard library, as its cookie implementation does not comply with the latest standards of cookies. We should look for Python libraries that do or, if none fit the bill, build our own Python library for modern cookie handling.

    As part of this implementation, we should build a comprehensive set of tests covering all aspects of cookie handling. [There is a draft pull request of initial work on this front](https://github.com/scrapy/scrapy/pull/5430).

    Related standards:

    - https://datatracker.ietf.org/doc/html/rfc6265
    - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis/
    - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site-00
    - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-alone-01

-   Let the `Cookie` header of a request be processed just the same as the `cookies` parameter of `Request`.

    Reported at https://github.com/scrapy/scrapy/issues/1992, which we [fixed](https://github.com/scrapy/scrapy/pull/2400), but we had to [revert the fix](https://github.com/scrapy/scrapy/pull/4823) due to [undesired side effects](https://github.com/scrapy/scrapy/issues/4717); there is however [a new draft pull request to address the issue for good](https://github.com/scrapy/scrapy/pull/4812).

-   Allow users to decide, for a request and its response, any combination of the following:
    
    -   Whether or not cookies defined by the user in the request itself (be it through the `cookies` parameter of `Request` or the `Cookie` header) should be included in the request cookiejar.
    -   Whether or not cookies from the cookiejar should be included into the request.
    -   Whether or not cookies received in the response should be included into the cookiejar.

    Some of the combinations are already possible, e.g. by the use of the `dont_merge_cookies` and `cookiejar` request metadata keys. We should extend support to the rest of scenarios, and make sure we document all scenarios properly.

    Related issues: #2124, #3769

-   Provide a user-friendly API to interact with cookiejars.

    Related issues: https://github.com/scrapy/scrapy/issues/1448, https://github.com/scrapy/scrapy/issues/1878, https://github.com/scrapy/scrapy/pull/2986

-   Allow cookie storage

[...]

PR #5463, opened 2022-04-04
Improve cookie handling

farsene:
Adding storage of cookies in local file that allows cross-spider access of cookies and providing interface method for the spiders to retrieve the cookies. Spiders automatically load the saved cookies when they are opened and write the new cookies to the file when they are closed such that another new opened spider can reuse the existing cookies. 

Part of #5431


docs/topics/spiders.rst
```diff
@@ -188,6 +188,32 @@ scrapy.Spider
...
+       :param return_type: return format if multiple cookies, default = list, options are list, dict
+       :type return_type: object
+
+       Get cookie by name or the cookies whose name is in names.
+       If names is used, then the return value will be in the format of the return_type
+
+   .. method:: clear_cookies()
+
+       Erase spider's cookiejar
+
```
Gallaecio, line 216: This is a big API change, which should probably be discussed in https://github.com/scrapy/scrapy/issues/1878 before working on an implementation.

PR #5469, opened 2022-04-08
added deprecated warning to NoimagesDrop exception

kuephi:
Hi! I added the warning. No test needed here, I suppose. Hope this is OK. It's my first contribution to an open source project.

Fixes  #5368

PR #5474, opened 2022-04-11
Fixed deprecetion warning happening all the time from #5469

drusca3:
Fixed the depreciation warning happening all the time when the library was loaded (see [5469](https://github.com/scrapy/scrapy/pull/5469))

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -8,7 +8,6 @@ import hashlib
 from contextlib import suppress
 from io import BytesIO
 
-
 from itemadapter import ItemAdapter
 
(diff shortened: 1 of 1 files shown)


wRAR:
This just adds a a function that is never called, so this change does nothing.

wRAR:
Have you tried to run this code?

drusca3:
Added a constructor for the deprecated class that throws the warning and any instances of a subclass.

Gallaecio:
Could you add a couple of tests, one for regular raising and another one for raising of a subclass?

I think we may also want to suggest the use of `DropItem` instead in the deprecation message, to give users a working alternative.

wRAR:
Fixed by #5489.

PR #5480, opened 2022-04-15
Added error handling for the case when Brotli is not imported

KishanPipariya:
I added the necessary error generation. Are any further changes required?

Closes: #4697

### DIFF
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -91,8 +91,11 @@ class HttpCompressionMiddleware:
                 # http://www.port80software.com/200ok/archive/2005/10/31/868.aspx
                 # http://www.gzip.org/zlib/zlib_faq.html#faq38
                 body = zlib.decompress(body, -15)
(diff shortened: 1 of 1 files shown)


wRAR:
I don't think it should be ImportError and I'm also not 100% sure it should be an exception at all (but actually running a callback here is not that useful as it certainly won't get the correctly decompressed response).

KishanPipariya:
What could be an appopriate response by the program?

KishanPipariya:
What kind of response should the user receive?

Gallaecio:
Could you also address the 1st bullet point of https://github.com/scrapy/scrapy/issues/4697?

I imagine we should create a boolean `self._can_decode_br` in `__init__`, and then in `process_request` checking if the `Accept-Encoding` header exists in the request, and `self._can_decode_br is False`, and `br` is one of the encodings in the header. If so, we log a warning.

I am also thinking that, instead of raising an exception if the response is br-encoded, we could remove `br` from `Accept-Encoding` in `process_response`, in addition to logging a warning (that would now also indicate the automatic drop of the unsupported encoding from the header).

KishanPipariya:
Will try to implement your suggestions
Thanks for the help

PR #5482, opened 2022-04-17
Fix TypeError in parse command

### LINKED ISSUE #5481: scrapy parse -h throws error
### Description
running `scrapy parse -h` from inside of a project throws an error.

### Steps to Reproduce
1.  `scrapy startproject example`
2.  `cd example`
3.  `scrapy parse -h`

**Expected behavior:**  It should show the help message output for the `parse` command
**Actual behavior:**  Throws error and displays traceback information
**Reproduces how often:**  Every time I tried it.

### Versions
v2.6.1

### Additional context

I identified the issue and will submit a PR shortly.

alexpdev:
Fixes #5481 

`scrapy parse -h`   now displays the appropriate help message.

### DIFF
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -51,7 +51,7 @@ class Command(BaseRunSpiderCommand):
         parser.add_argument("--cbkwargs", dest="cbkwargs",
                             help="inject extra callback kwargs into the Request, it must be a valid raw json string")
         parser.add_argument("-d", "--depth", dest="depth", type=int, default=1,
-                            help=f"maximum depth for parsing requests [default: {self.max_level}]")
+                            help="maximum depth for parsing requests [default: 1]")
         parser.add_argument("-v", "--verbose", dest="verbose", action="store_true",
                             help="print each depth level one by one")
 
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -51,7 +51,7 @@ class Command(BaseRunSpiderCommand):
         parser.add_argument("--cbkwargs", dest="cbkwargs",
                             help="inject extra callback kwargs into the Request, it must be a valid raw json string")
         parser.add_argument("-d", "--depth", dest="depth", type=int, default=1,
-                            help="maximum depth for parsing requests [default: %default]")
+                            help=f"maximum depth for parsing requests [default: {self.max_level}]")
         parser.add_argument("-v", "--verbose", dest="verbose", action="store_true",
                             help="print each depth level one by one")
 
--- a/scrapy/commands/parse.py
+++ b/scrapy/commands/parse.py
@@ -51,7 +51,7 @@ class Command(BaseRunSpiderCommand):
         parser.add_argument("--cbkwargs", dest="cbkwargs",
                             help="inject extra callback kwargs into the Request, it must be a valid raw json string")
         parser.add_argument("-d", "--depth", dest="depth", type=int, default=1,
-                            help="maximum depth for parsing requests [default: 1]")
+                            help="maximum depth for parsing requests [default: %(default)s]")
         parser.add_argument("-v", "--verbose", dest="verbose", action="store_true",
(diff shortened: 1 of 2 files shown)


scrapy/commands/parse.py
```diff
@@ -51,7 +51,7 @@ def add_options(self, parser):
         parser.add_argument("--cbkwargs", dest="cbkwargs",
                             help="inject extra callback kwargs into the Request, it must be a valid raw json string")
         parser.add_argument("-d", "--depth", dest="depth", type=int, default=1,
-                            help="maximum depth for parsing requests [default: %default]")
+                            help="maximum depth for parsing requests [default: 1]")
```
Laerte, line 54:
```suggestion
                            help="maximum depth for parsing requests [default: %(default)s]")
```
This way you don't need to define twice. Syntax is from documentation<sup>[1](https://docs.python.org/3/library/argparse.html)</sup>. Tested locally and it works!

> Replace strings with implicit arguments such as %default or %prog with the standard Python syntax to use dictionaries to format strings, that is, %(default)s and %(prog)s.

Laerte (changes requested):

scrapy/commands/parse.py
alexpdev: I fixed thanks for the suggestion.

wRAR (approved):

elacuesta (approved):
Looks good, thank you. Two suggestions for the future:
* There's no need to open an issue if you already have a patch in mind, you can submit a PR directly
* Use a descriptive name in the PR. This will (probably) be the message we end up using when merging, and "fixes issue #N" is not good for the history

alexpdev:
> Looks good, thank you. Two suggestions for the future:
> 
> * There's no need to open an issue if you already have a patch in mind, you can submit a PR directly
> * Use a descriptive name in the PR. This will (probably) be the message we end up using when merging, and "fixes issue #N" is not good for the history

Thanks for the advice.

Laerte (approved):

Gallaecio:
@Laerte Could you rebase this change into the 2.6 branch instead of `master`?

Laerte:
@Gallaecio You want me to cherry-pick the changes from his branch and apply above `2.6` branch? In this case we will need open another PR? (I don't have write access to his branch that's why i thinked on this approach.

alexpdev:
How do i grant write access?

Laerte:
@alexpdev You can follow these steps: https://github.com/scrapy/scrapy/pull/5445#issuecomment-1062664683

(Change the commit id of course)

alexpdev:
> @alexpdev You can follow these steps: [#5445 (comment)](https://github.com/scrapy/scrapy/pull/5445#issuecomment-1062664683)
> 
> (Change the commit id of course)

on the line that starts with `git checkout` I'm getting `fatal 'upstream/2.6' is not a commit and a branch 'master' cannot be created from it`

Laerte:
@alexpdev Just push and change the base here to 2.6.

alexpdev:
Did I do that right?

Laerte:
@alexpdev Yeah, now you just have to change this to 2.6 as well:

Screenshot_20220424-031118-346

alexpdev:
got it

Laerte:
@alexpdev Almost there... As you can see we have commits from `master` in the commits list, you need to drop them and let just yours commits. You can do this:

1.  Rebase from [636127e](https://github.com/alexpdev/scrapy/commit/636127ec1ea2b8949438015c2167ab5d009ff1bf) commit.

```
git rebase -i 636127e^
```

2. This will open a text-editor, you need to **drop** all the commits that are not yours, you can archive this replacing **pick** to **drop**.

Then you need to push again:
```
git push -f
```

alexpdev:
@Laerte Okay, got it... thanks again for all of your help

Laerte:
Nice job! 🚀

wRAR (approved):

PR #5487, opened 2022-04-30
Fix typos in three files for Flake8 check

AngelikiBoura:
Some fixes in the extensions files memusage.py and statsmailer.py and in the test file twisted_reactor_custom_settings_same.py
were needed in order for the flake8 check to be succesful.

### DIFF
--- a/scrapy/extensions/memusage.py
+++ b/scrapy/extensions/memusage.py
@@ -33,8 +33,8 @@ class MemoryUsage:
         self.crawler = crawler
         self.warned = False
         self.notify_mails = crawler.settings.getlist('MEMUSAGE_NOTIFY_MAIL')
-        self.limit = crawler.settings.getint('MEMUSAGE_LIMIT_MB')*1024*1024
-        self.warning = crawler.settings.getint('MEMUSAGE_WARNING_MB')*1024*1024
+        self.limit = crawler.settings.getint('MEMUSAGE_LIMIT_MB') * 1024 * 1024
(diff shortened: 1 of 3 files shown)


wRAR:
The changes look fine to me but the flake8 test already succeeds without these changes, does it fail for you?

AngelikiBoura:
Yeah, i run tox -e flake8 and it showed me these changes in order to pass.

wRAR:
This is strange because neither `tox -e flake8` nor installing the latest flake8 and running it directly have any errors here.

AngelikiBoura:
It is indeed. I'll try running it again then from a different computer and
let you know.

wRAR:
Actually, `flake8 scrapy tests docs` processes them but `pytest --flake8 scrapy tests docs` doesn't, because pytest skips these files via conftest.py and tests/ignores.txt. I wonder how can we improve this to not skip any files when running flake8 and other checkers.

wRAR (approved):

wRAR:
I also don't remember why those three files in tests/ignores.txt are there.

AngelikiBoura:
When i ran now the same command from my terminal the check passed successfully without the changes.
The check didn't pass when i ran the command through the terminal of my IDE (pycharm).  
I'm not sure if this info is useful but i hope it helps.

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5488, opened 2022-04-30
include True and False value for debug var in tests

andreastziortz:
Include both bool values for debug mode to increase test coverage for mail.py


wRAR:
debug=False means it actually sends an email. Test shouldn't do that.

PR #5492, opened 2022-05-06
The Request object adds the params parameter

TheWeiJun:
###  The Request object adds the params parameter

example：
```
url = "https://www.google.com.hk/search?"
params = {
     "q": "facebook",
     "tbm": "isch",
     "hl": "zh-CN",
     "safe": "strict",
     "chips": "q:facebook,online_chips:fb logo",
     "sa": "X",
     "ved": "2ahUKEwiE5_3z7sn3AhV_ATQIHV9lBZEQ4lYoBHoECAEQJQ",
     "biw": 1440,
     "bih": 708
}
url = f"{url[:-1] if url.endswith('?') else url}?{urlencode(params)}"
```
image

image

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -6,6 +6,7 @@ See documentation in docs/topics/request-response.rst
 """
 import inspect
 from typing import Callable, List, Optional, Tuple, Type, TypeVar, Union
+from urllib.parse import urlencode
 
 from w3lib.url import safe_url_string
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
Implementing such a feature is not that simple. It is not only that you are making a lot of assumptions about the URL with that `endswith(`?`)` (e.g. what about `https://example.com?a=b` or `https://example.com#id`?), but you should also make sure request serialization keeps the parameters (that is what attributes is for).

This is also something that should probably be discussed before implementing. I am not sure of exactly why this feature has not implemented in the past, but I get a feeling that there are reasons. If nothing else, that you can use something like `w3lib` to do this, so it is (arguably) not necessary to provide this feature in Scrapy itself.

wRAR:
Use `FormRequest` with method='GET' and pass the data in `formdata`.

PR #5505, opened 2022-05-21
Update README.rst

aseemsaxena:
I have added changes to promote inclusivity according to the policy and ethical standards of the project to help explain to the beginner what a web crawler is and give them an example. I believe this is vital as it could promote the beginner to get started using the project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,23 @@ Scrapy is a fast high-level web crawling and web scraping framework, used to
 crawl websites and extract structured data from their pages. It can be used for
 a wide range of purposes, from data mining to monitoring and automated testing.
 
+Welcome to Scrapy, A web crawler or spider are synonymous terms for a utility written 
+in a programming language that indexes content from all over the Internet. The goal
+of such a web crawler is to learn what the page is about and then retrieve what is
+actually needed. Scrapy is simply an instance of this application. They are called
+such as they automatically go around websites looking at metadata and obtaining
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -44,6 +44,23 @@ Scrapy is a fast high-level web crawling and web scraping framework, used to
...
+A web crawler is similar to going to a library of unorganized books and putting
+together a card catalog quickly so information can be found easily.  In order
+to do the task, the web crawler will look at the topics, title, and summary to
+figure out what goes where. This is similar to how this will function on the
+internet. You can use Scrapy in a multitude of ways. One example would be 
+extracting headlines from a news source and creating your own headline
+aggregating platform for things that you are interested in. The sky is
+the limit with Scrapy. Please see our technical documentation to get started.
+
+
```
Gallaecio, line 63: I think yours is a really good point, so first off thank you for bringing this up and working on it.

However, I don’t think this is the best way to address your point. I think, in the spirit of [Don’t Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), we should instead link the corresponding keywords from the existing paragraph to the corresponding Wikipedia pages ([web crawling](https://en.wikipedia.org/wiki/Web_crawler),  [web scraping](https://en.wikipedia.org/wiki/Web_scraping)).

If the Wikipedia topics are deemed not good enough, they can be improved upstream.

PR #5506, opened 2022-05-22
Fix charset handling in MailSender

### LINKED ISSUE #5096: Wrong charset handling in MailSender

andreastziortz:
Changes:
  
Implementation
- Set encoding utf-8 for payload inside send

Fixes #5096, closes #5118

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -77,6 +77,9 @@ class MailSender:
             rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
 
+        if charset:
+            msg.set_charset(charset)
+
         if attachs:
             msg.attach(MIMEText(body, 'plain', charset or 'us-ascii'))
             for attach_name, mimetype, f in attachs:
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -89,7 +89,6 @@ class MailSender:
                 Encoders.encode_base64(part)
                 part.add_header('Content-Disposition', 'attachment', filename=attach_name)
                 msg.attach(part)
-
         else:
             msg.set_payload(body)
             msg.set_charset(charset)
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -77,19 +77,16 @@ class MailSender:
             rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
 
-        if charset:
-            msg.set_charset(charset)
-
         if attachs:
             msg.attach(MIMEText(body, 'plain', charset or 'us-ascii'))
             for attach_name, mimetype, f in attachs:
                 part = MIMEBase(*mimetype.split('/'))
-                part.set_payload(f.read())
+                part.set_payload(f.read(), charset)
                 Encoders.encode_base64(part)
                 part.add_header('Content-Disposition', 'attachment', filename=attach_name)
                 msg.attach(part)
         else:
-            msg.set_payload(body)
+            msg.set_payload(body, charset)
 
         if _callback:
             _callback(to=to, subject=subject, body=body, cc=cc, attach=attachs, msg=msg)
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -77,11 +77,14 @@ class MailSender:
             rcpts.extend(cc)
             msg['Cc'] = COMMASPACE.join(cc)
 
+        if charset:
+            msg.set_charset(charset)
+
         if attachs:
             msg.attach(MIMEText(body, 'plain', charset or 'us-ascii'))
             for attach_name, mimetype, f in attachs:
                 part = MIMEBase(*mimetype.split('/'))
-                part.set_payload(f.read(), charset)
+                part.set_payload(f.read())
                 Encoders.encode_base64(part)
                 part.add_header('Content-Disposition', 'attachment', filename=attach_name)
                 msg.attach(part)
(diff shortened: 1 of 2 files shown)


andreastziortz:
Hello @Gallaecio @wRAR , 
I created a Pull Request for wrong charset handling in Mail Sender.
- Encode in utf-8  th payload from send inside mail.py
- Remove set encoding in utf-8 from top level in send() inside mail.py

Gallaecio (commented):
I see you are changing expectations of existing tests, and not adding a new test that would have broken before your change but works now. Could you look into it and, if your changes are as intended, explain why the modified test expectations are OK?

andreastziortz:
Changes:

Finally, I narrowed my changes to a small one.  
Implemenetation
- Set encoding utf-8 for payload inside send
 
Testing
- Did not change any of existing tests or created new ones. The reason that tests could not detect this defect until now was related with the the way set_charset works. So, set_charset after encoding the message to send it, decodes it. As a result, mail.send() returns the decoded text and thus it is not easy to test whether the body of text has been encoded or not apart from checking that the Charset type in headers has been set as expected.
Although, I 'm interested to implement any test that could prove the new behaviour, if there is a way. 
@Gallaecio @wRAR

scrapy/mail.py
```diff
@@ -89,7 +89,7 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
                 part.add_header('Content-Disposition', 'attachment', filename=attach_name)
                 msg.attach(part)
         else:
-            msg.set_payload(body)
+            msg.set_payload(body, charset)
```
Gallaecio, line 92: Looking at https://bugs.python.org/issue31820, I wonder if the solution is instead to move the 2 charset lines to _after_ the body or attachments are set, and make no change here.

As for tests, it seems `get_payload` has `decode=False` by default, so it should return the encoded body.

andreastziortz: thanks @Gallaecio for your suggestions. I tried to add set_charset after set_payload, updating tests to reflect changes but again encoded body and attaches is not visible in tests.

tests/test_mail.py
```diff
@@ -108,9 +108,8 @@ def test_send_attach_utf8(self):
 
         msg = self.catched_msg['msg']
         self.assertEqual(msg['subject'], subject)
-        self.assertEqual(msg.get_charset(), Charset('utf-8'))
-        self.assertEqual(msg.get('Content-Type'),
-                         'multipart/mixed; charset="utf-8"')
+        self.assertEqual(msg.get_charset(), None)
+        self.assertEqual(msg.get('Content-Type'), 'multipart/mixed')
```
Gallaecio, line 112: Isn’t this a refression? Shouldn’t the old test expectations continue to work? Otherwise, maybe we are doing something wrong.

andreastziortz: Thanks @Gallaecio , yes, I agree, that's why I set old tests as were before and added a new similar one. The implemented approach seems the only compatible way to avoid regressions and fix the issue. 
The main difference I can see is that if you try to revert changes on test and instead of testing:
self.assertEqual(msg.get_payload(decode=True).decode(), body) -> self.assertEqual(msg.get_payload, body.encode('utf-8')) make tests not failed, because payload attribute is overwritten at the end of set_charset() by this line:
 self._payload = charset.body_encode(payload), but maybe that's ok.
I made also a test with two attachments to test.

Gallaecio (approved):
Looks great code-wise, thanks!

scrapy/mail.py
```diff
@@ -85,11 +85,13 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
             for attach_name, mimetype, f in attachs:
                 part = MIMEBase(*mimetype.split('/'))
                 part.set_payload(f.read())
+                part.set_charset(charset)
                 Encoders.encode_base64(part)
                 part.add_header('Content-Disposition', 'attachment', filename=attach_name)
                 msg.attach(part)
         else:
             msg.set_payload(body)
+            msg.set_charset(charset)
```
wRAR, line 94: This is a no-op, at least in the bug context, as doing this doesn't recode the payload.

scrapy/mail.py
```diff
@@ -85,11 +85,13 @@ def send(self, to, subject, body, cc=None, attachs=(), mimetype='text/plain', ch
             for attach_name, mimetype, f in attachs:
                 part = MIMEBase(*mimetype.split('/'))
                 part.set_payload(f.read())
+                part.set_charset(charset)
```
wRAR, line 88: Not sure if this is needed as attachments are already handled correctly.

tests/test_mail.py
```diff
@@ -121,6 +121,43 @@ def test_send_attach_utf8(self):
         self.assertEqual(text.get_charset(), Charset('utf-8'))
         self.assertEqual(attach.get_payload(decode=True).decode('utf-8'), body)
 
+    def test_encoded_body_with_utf8(self):
```
wRAR, line 124: This test passes without the code changes.

wRAR (changes requested):

PR #5508, opened 2022-05-24
pass on item to thumb_path function as additional argument 

### LINKED ISSUE #5504: ImagesPipeline.thumb_path should allow access to item
## Summary

ImagesPipeline should pass on item to thumb_path function as additional argument.

## Motivation

The ImagesPipeline.file_path function allows access to item. So when an user overwrites this function and makes the file name
depended on the item itself, it makes sense to have the thumb path to be depended on the item as well.

This way main image file and thumbnail image file can have the same name and it is easily possible to jump between different sizes of the image.

Example of how a file and thumbnail could be build:
```python
file_path = f"full/{item['ID']}.jpg"
thumb_path = f"thumbs/small/{item['ID']}.jpg"
```

So I suggest to change the signature of ImagesPipeline.thumb_path to the following:
```python
def thumb_path(self, request, thumb_id, response=None, info=None, *, item=None):
```
which makes the functions thumb_path and file_path more consistent.

As a workaround when I overwrite `file_path` I set `request._item = item` which I can then access in `thumb_path` function.

## Additional context

In case this is of interest I am happy to work on a pull request.

mtabbasi:
Fixes #5504

### DIFF
--- a/docs/topics/media-pipeline.rst
+++ b/docs/topics/media-pipeline.rst
@@ -656,6 +656,26 @@ See here the methods that you can override in your custom Images Pipeline:
       .. versionadded:: 2.4
          The *item* parameter.
 
+   .. method:: ImagesPipeline.thumb_path(self, request, thumb_id, response=None, info=None, *, item=None)
+
+      This method is called for every item of  :setting:`IMAGES_THUMBS` per downloaded item. It returns the
+      thumbnail download path of the image originating from the specified
+      :class:`response <scrapy.http.Response>`.
+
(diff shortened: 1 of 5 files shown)


wRAR:
Can you please take a look at #4686 and update your PR accordingly?

mtabbasi:
> 

Sure

Gallaecio (approved):

kmike:
Thanks @mtabbasi, the PR looks good!

mtabbasi:
My pleasure @kmike

PR #5513, opened 2022-06-03
extract data from DBpedia page and return them in a pdf version

VirginiaKostara:
I have added code for extracting data from an open knowledge graph, DBpedia, which pulls out structured content from the information created in the Wikipedia project. The data are extracted by decoding the source code of a DBpedia page and are printed in a new PDF file. 
I have added coverage for testing the results of the functions.


elacuesta:
This doesn't seem to relate directly to Scrapy. Even if it did, we do not keep examples like this in this repository.

PR #5514, opened 2022-06-03
Drop Python 3.6 support

### LINKED ISSUE #5512: Drop Python 3.6 support
[It went end-of-life on December 2021](https://endoflife.date/python).

Laerte:
Closes #5512


elacuesta (changes requested):
Hi, thanks for the update. We still the typing, pypy and pinned envs though, we need to update their versions instead of removing them completely.

Laerte:
@elacuesta Sure! I will change to use pypy version 3.7, what you think?

elacuesta:
Indeed that works, although I don't think we have a strong requirement for pypy to be any specific major version. That's not the case for the pinned one, ideally it should be the earliest version in the 3.7 series (that's supported by CI).

.github/workflows/tests-windows.yml
```diff
@@ -8,9 +8,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python-version: 3.6
-          env:
-            TOXENV: windows-pinned
```
elacuesta, line 13: This one also needs to be updated, not removed

wRAR:
Looks like pinned lxml is too old for 3.7?

Laerte:
@wRAR Yeah, i will take a look later.

Laerte:
@wRAR @elacuesta  Had to bump some dependencies in order to make everything work with pinned 3.7. Don't know why codecov & checks (3.10, docs) are failing but is not related with the changes that i made.

harivamsi9:
was anyone able to figure out why codecov/path check is failling?

elacuesta:
The docs env is failing in master too (517cbc8d717d5f10c8e54114fa32dca6f54588a9), I wouldn't worry about it for this PR. Regarding the coverage check, I suppose it could be possible that parts of the codebase are not reached now with the new requirement versions, but I'm not completely sure that's the case.

We need to update `install_requires` too, if we know that some versions are not compatible anymore with the earliest Python version we support. For instance, I suppose that if you needed to update Twisted in the pinned env it means that `Twisted==17.9.0` is no longer valid (under py37 anyway), so it should be updated in [setup.py](https://github.com/scrapy/scrapy/blob/cf2ad16c4991f50b6fc9705c96431993689c3fee/setup.py#L22).

tox.ini
```diff
@@ -104,7 +103,7 @@ install_command =
 [testenv:pinned]
 deps =
     {[pinned]deps}
-    lxml==3.5.0
+    lxml==4.3.0
```
elacuesta, line 106: Seems like `lxml` is the same version in all envs now, perhaps it could be moved to the main pinned requirements list.

Laerte:
Some lessons learned with this PR until now:

Earliest supported release for pypy3.7 is **v7.3.5**, reason:
https://foss.heptapod.net/pypy/pypy/-/issues/3400

Bumped `Twisted` version, reason:
- https://github.com/twisted/twisted/pull/966

Bumped `cryptography` and `pyOpenSSL`, because the test below was failing: 
```tests/test_webclient.py::WebClientCustomCiphersSSLTestCase::testPayloadDisabledCipher```

**Comment by Gallaecio** [comment]
> was anyone able to figure out why codecov/path check is failling?

Looking at https://app.codecov.io/gh/scrapy/scrapy/compare/5514/diff, that is a expected non-covered path (we decided long ago not to run a CI job just to verify that Scrapy fails nicely on an unsupported Python version).

File: .github/workflows/tests-ubuntu.yml [hunk 3]
```diff
@@ -26,19 +26,19 @@ jobs:
         - python-version: pypy3
           env:
             TOXENV: pypy3
-            PYPY_VERSION: 3.6-v7.3.3
+            PYPY_VERSION: 3.7-v7.3.9
```
Gallaecio, line 29: I believe this is meant to be the latest PyPy release, [3.6 just happened to be the latest Python version available at the time](https://doc.pypy.org/en/latest/release-v7.3.1.html) when [this was first defined](https://github.com/scrapy/scrapy/commit/80db569aeae3e05480ed228c95332c17f67e81e4), and since then [we only updated it once to address a specific issue](https://github.com/scrapy/scrapy/pull/5190/commits/ba40bccc2d779d332cc74f43759fe57f71d3376e), and being focused on addressing the issue we forgot to update to Python 3.7 back then.

```suggestion
            PYPY_VERSION: 3.9-v7.3.9
```

.github/workflows/tests-windows.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python-version: 3.6
+        - python-version: 3.7
           env:
             TOXENV: windows-pinned
         - python-version: 3.7
```
Gallaecio, line 14: The non-pinned 3.7 job here can be removed now.

.github/workflows/tests-ubuntu.yml
```diff
@@ -26,19 +26,19 @@ jobs:
...
             TOXENV: pinned
-        - python-version: 3.6.12
+        - python-version: 3.7.13
           env:
             TOXENV: asyncio-pinned
         - python-version: pypy3
           env:
             TOXENV: pypy3-pinned
-            PYPY_VERSION: 3.6-v7.2.0
+            PYPY_VERSION: 3.7-v7.3.5
```
Gallaecio, line 41: Now that these run 3.7, the regular 3.7 job at the top of the list can be removed now.

README.rst
```diff
@@ -57,7 +57,7 @@ including a list of features.
 Requirements
 ============
 
-* Python 3.6+
+* Python 3.7+
```
Gallaecio, line 60: See also `docs/intro/install.rst` ([see how we changed it when dropping 3.5 support](https://github.com/scrapy/scrapy/pull/4742/files#diff-5967a18372589c0f3cdf68a1f2bddbdf069678675819eb828b1a6c9d3323238d))

Looking for “3.6” I also found references worth updating or removing in:

```
docs/contributing.rst
docs/topics/items.rst
tests/test_utils_python.py (we can probably simplify to keep only the statement within the if)
```

tests/requirements.txt
```diff
@@ -1,6 +1,5 @@
 # Tests requirements
 attrs
-dataclasses; python_version == '3.6'
 pyftpdlib
 pytest
 pytest-cov==3.0.0
```
Gallaecio, line 5: Se also the `uvloop` lines below, the first one can be removed and the second one simplified.

tox.ini
```diff
@@ -17,9 +17,8 @@ deps =
     #mitmproxy >= 5.3.0; python_version >= '3.9' and implementation_name != 'pypy'
     # The tests hang with mitmproxy 8.0.0: https://github.com/scrapy/scrapy/issues/5454
     mitmproxy >= 4.0.4, < 8; python_version >= '3.7' and python_version < '3.9' and implementation_name != 'pypy'
-    mitmproxy >= 4.0.4, < 5; python_version >= '3.6' and python_version < '3.7' and platform_system != 'Windows' and implementation_name != 'pypy'
     # newer markupsafe is incompatible with deps of old mitmproxy (which we get on Python 3.7 and lower)
-    markupsafe < 2.1.0; python_version >= '3.6' and python_version < '3.8' and implementation_name != 'pypy'
+    markupsafe < 2.1.0; python_version >= '3.7' and python_version < '3.8' and implementation_name != 'pypy'
```
Gallaecio, line 21: - `# mitmproxy does not support Windows when running Python < 3.7` can be removed now
- `mitmproxy >= 4.0.4, < 8; python_version >= '3.7' and python_version < '3.9' and implementation_name != 'pypy'` can be simplified as `mitmproxy >= 4.0.4, < 8; python_version < '3.9' and implementation_name != 'pypy'`
- `markupsafe < 2.1.0; python_version >= '3.7' and python_version < '3.8' and implementation_name != 'pypy'` can be simplified as `markupsafe < 2.1.0; python_version < '3.8' and implementation_name != 'pypy'` (it could already before your change, [an oversight from me during the last update](https://github.com/scrapy/scrapy/pull/4742/files#diff-6aa3751e2010724eb62472e6ebbb0aa229da6a9e10678a53b6e6373a40614e88))

tox.ini
```diff
@@ -17,9 +17,8 @@ deps =
...
     # Extras
     botocore>=1.4.87
 passenv =
```
Gallaecio, line 24: Remove `pip install types-dataclasses  # remove once py36 support is dropped` below.

Gallaecio (commented):
Maybe we should also remove `scrapy/utils/py36.py` here, it has been more than a year since it was first deprecated (Scrapy 2.5).

tox.ini
```diff
@@ -95,7 +95,7 @@ deps =
     # Extras
     botocore==1.4.87
     google-cloud-storage==1.29.0
-    Pillow==4.0.0
+    Pillow==7.1.0
```
Gallaecio, line 98: This also requires updating `docs/topics/media-pipeline.rst`.

docs/topics/items.rst
```diff
@@ -102,10 +102,7 @@ Additionally, ``dataclass`` items also allow to:
 * define custom field metadata through :func:`dataclasses.field`, which can be used to
   :ref:`customize serialization <topics-exporters-field-serialization>`.
 
-They work natively in Python 3.7 or later, or using the `dataclasses
-backport`_ in Python 3.6.
-
-.. _dataclasses backport: https://pypi.org/project/dataclasses/
+They work natively in Python 3.7 or later.
```
Gallaecio, line 105: I think we can remote this paragraph entirely.

Gallaecio, line 105: I think we can remote this paragraph entirely.

setup.py
```diff
@@ -19,35 +19,30 @@ def has_environment_marker_platform_impl_support():
...
 if has_environment_marker_platform_impl_support():
     extras_require[':platform_python_implementation == "CPython"'] = cpython_dependencies
     extras_require[':platform_python_implementation == "PyPy"'] = [
-        # Earlier lxml versions are affected by
-        # https://foss.heptapod.net/pypy/pypy/-/issues/2498,
-        # which was fixed in Cython 0.26, released on 2017-06-19, and used to
-        # generate the C headers of lxml release tarballs published since then, the
-        # first of which was:
-        'lxml>=4.0.0',
+        'lxml>=4.3.0',
```
Gallaecio, line 45: The 2 `'lxml>=4.3.0'` can now be removed and replaced by a single `'lxml>=4.3.0'` in `install_requires` above.

Gallaecio (approved):
Great job! I bet you did not anticipate it to be this involved :slightly_smiling_face:

Gallaecio:
(`scrapy/utils/py36.py` is still there, but I don’t think it _needs_ to be handled as part of this pull request, is more of a deprecation cleanup task)

elacuesta (approved):
Thanks!

PR #5518, opened 2022-06-08
When running the installer, I wasn't sure where to put the 'pip insta…

nguyen-bo:
…ll Scrapy' command, so I added the 'inside of the Python terminal' part to line 29.


wRAR:
It says "if you’re already familiar with installation of Python packages".

And "inside of the Python terminal" is of course wrong, it's a command, not a Python snippet.

PR #5528, opened 2022-06-14
If TWISTED_REACTOR is None, reuse any pre-installed reactor

### LINKED ISSUE #5525: Importing reactor in spider's file can cause `ReactorAlreadyInstalledError`
### Description

When running a spider with `scrapy crawl ... <spider>`, due to a change in the way the reactor is installed (possibly introduced in 60c8838), importing reactor from twisted at the top of that spider's python file will now cause an error (in the case where `TWISTED_REACTOR` is not defined):

`twisted.internet.error.ReactorAlreadyInstalledError: reactor already installed`

I was able to work around the issue by explicitly setting `TWISTED_REACTOR` which causes scrapy's `install_reactor()` helper to be used (which explicitly suppresses that error). This doesn't allow me to use twisted's default reactor per platform (without recreating the logic to do so myself).

### Steps to Reproduce

`from twisted.internet import reactor` at the top of a spider's python file
(importing the reactor module installs the default reactor)

### Versions
```
Scrapy       : 2.6.1
lxml         : 4.9.0.0
libxml2      : 2.9.12
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 22.4.0
Python       : 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 19:29:22) [MSC v.1916 32 bit (Intel)]
pyOpenSSL    : 22.0.0 (OpenSSL 3.0.3 3 May 2022)
cryptography : 37.0.2
Platform     : Windows-10-10.0.17763-SP0
```

Gallaecio:
Fixes #5525

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1638,9 +1638,10 @@ which raises :exc:`Exception`, becomes::
 
 
 The default value of the :setting:`TWISTED_REACTOR` setting is ``None``, which
-means that Scrapy will install the default reactor defined by Twisted for the
-current platform. This is to maintain backward compatibility and avoid possible
-problems caused by using a non-default reactor.
+means that Scrapy will use the existing reactor if one is already installed, or
+install the default reactor defined by Twisted for the current platform. This
+is to maintain backward compatibility and avoid possible problems caused by
+using a non-default reactor.
 
 For additional information, see :doc:`core/howto/choosing-reactor`.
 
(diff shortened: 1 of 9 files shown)


Gallaecio:
The documentation failure can be ignore, I plan to address that separately.

Gallaecio:
Hmm… I assumed Windows would be using one of the Windows-specific reactors, but I was wrong, and select seems to be the only crossplatform reactor (making it suitable for the failing test), so I will skip those tests on Windows.

PR #5529, opened 2022-06-14
Verify that the installed loop matches ASYNCIO_EVENT_LOOP

Gallaecio:
I realized this was worth doing while working on https://github.com/scrapy/scrapy/pull/5528, but I thought it best not to make it part of that pull request.

I consider it a (minor) bug that we did not check this before, so I am not completely against adding this to 2.6, but since this is not a regression, I also do not see a reason to add this before 2.7.


Gallaecio:
The documentation failure can be ignore, I plan to address that separately.

scrapy/utils/reactor.py
```diff
@@ -90,6 +90,24 @@ def verify_installed_reactor(reactor_path):
...
+        return
+    installed = (
+        f"{reactor._asyncioEventloop.__class__.__module__}"
+        f".{reactor._asyncioEventloop.__class__.__qualname__}"
+    )
+    specified = f"{loop_class.__module__}.{loop_class.__qualname__}"
+    raise Exception(
+        f"Scrapy found an asyncio Twisted reactor already "
+        f"installed, and its event loop class ({installed}) does "
+        f"not match the one specified in the ASYNCIO_EVENT_LOOP "
```
Laerte, line 106:
```suggestion
        "Scrapy found an asyncio Twisted reactor already "
        f"installed, and its event loop class ({installed}) does "
        "not match the one specified in the ASYNCIO_EVENT_LOOP "
```
Nitpicking. 🤓

Laerte: Cool! I was not aware that this produce the same byte code, basically this is cosmetic preference then. 💅🏻 I tend to use this way because is explicit that you have a string interporlation in that line.

scrapy/utils/reactor.py
```diff
@@ -90,6 +90,24 @@ def verify_installed_reactor(reactor_path):
         raise Exception(msg)
 
 
+def verify_installed_loop(loop_path):
```
kmike, line 93: What do you think about adding "asyncio" to the name? This function seems to be asyncio-specific.

kmike (approved):

scrapy/utils/reactor.py
Gallaecio: Looking at the rest of the code, `verify_installed_asyncio_event_loop` would be the most consistent naming. Would it be too verbose?

Issue #5516, opened 2022-06-07
Feed is not ovewritten when custom extension is used


labdmitriy:
### Description

I'm trying to export `scrapy crawl` results to JSON Lines format to the file with extension `.jsonl` (this is requirement of the external system in our case) and ovewrite the file for multiple executions.
As I understand, only `.jl` and `.jsonlines` extensions are supported now and `.jsonl` was discussed in #4848 but not supported yet.
So in this case I tried to use `-O` argument with `--output-format` for `scrapy crawl` command.

### Steps to Reproduce

1. `scrapy crawl -O <filename>.jsonl --output-format jl <spider_name>` 
    OR 
    `scrapy crawl -O <filename>.jsonl --output-format jsonlines <spider_name>`

**Expected behavior:** File is ovewritten with parsed content.

**Actual behavior:** Parsed content is appended to the end of existing file.

**Reproduces how often:** 100%

### Versions

Scrapy       : 2.6.1
lxml         : 4.9.0.0
libxml2      : 2.9.14
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 22.4.0
Python       : 3.10.4 (main, Jun  4 2022, 14:29:37) [GCC 9.4.0]
pyOpenSSL    : 22.0.0 (OpenSSL 3.0.3 3 May 2022)
cryptography : 37.0.2
Platform     : Linux-5.13.0-44-generic-x86_64-with-glibc2.31

### Additional context

If I use `scrapy crawl -O <filename>.jl <spider_name>` or `scrapy crawl -O <filename>.jsonlines <spider_name>`, file is overwritten successfully, but it seems that the case above is expected to have the same behaviour.

Gallaecio:
Oh, this is actually as designed, but there is room for documentation (log message) improvement.

When you use the failing syntax, you get a warning:

> ScrapyDeprecationWarning: The -t command line option is deprecated in favor of specifying the output format within the output URI. See the documentation of the -o and -O options for more information.

`-t` is `--output-format` (this is what can be improved, the message should probably say `-t/--output-format` instead).

So, since the introduction of support for multiple output feeds, you are expected to use this syntax instead:

```
scrapy crawl -O <filename>.jsonl:jsonlines
```

labdmitriy:
Thanks a lot for your feedback, 
Sorry I really didn't notice this warning because of a lot of log messages.

Your approach is working, thank you!

[...]

PR #5530, opened 2022-06-14
Update crawl output-format documentation

justinschuster:
Update documentation for crawl output format for #5516

### DIFF
--- a/docs/topics/commands.rst
+++ b/docs/topics/commands.rst
@@ -271,11 +271,17 @@ crawl
 
 Start crawling using a spider.
 
+Supported options:
+
+* ``--output-format`` or ``-o/-O``: select output format.
+
 Usage examples::
 
     $ scrapy crawl myspider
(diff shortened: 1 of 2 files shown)


docs/topics/commands.rst
```diff
@@ -271,11 +271,17 @@ crawl
 
 Start crawling using a spider.
 
+Supported options:
+
+* ``--output-format`` or ``-o/-O``: select output format.
```
Gallaecio, line 276: I don’t think we should only cover 2 options, as that is slightly worse than before. Before the documentation was incomplete, now it is misleading as users may think from reading the documentation that only these options are available. What would you think about covering all options of `scrapy crawl`?

Also mind that you are mixing 3 different options here. ``--output-format/-t` is 1 option, `-o` is another option, and `-O` is another option, and they should be covered separately.

wRAR:
Included in #5605

PR #5531, opened 2022-06-16
dictionary parameter support

lemisky:
```python
from scrapy.cmdline import execute

execute(
        ['scrapy', 'runspider', 'myspider.py', '-O', 'blogs.json'],
        {'BOT_NAME': 'blogbot'} 
    )
```

Make compatible with this writing style

### DIFF
--- a/scrapy/cmdline.py
+++ b/scrapy/cmdline.py
@@ -12,6 +12,7 @@ from scrapy.exceptions import UsageError
 from scrapy.utils.misc import walk_modules
 from scrapy.utils.project import inside_project, get_project_settings
 from scrapy.utils.python import garbage_collect
+from scrapy.settings import Settings
 
 
 def _iter_command_classes(module_name):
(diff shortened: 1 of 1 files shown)


lemisky:
Sorry, I'm not familiar with how to add a test, you can modify this PR in place, thanks 🌹🌹 .

Gallaecio:
Why did you decide to propose this change? Do you use `execute` yourself, directly or indirectly?

During our tests `settings` is never passed to this function, according to [coverage data](https://app.codecov.io/gh/scrapy/scrapy/blob/master/scrapy/cmdline.py). And not even [the commit that introduced that parameter](https://github.com/scrapy/scrapy/commit/c7f821990117dc69f5c843e6bc1f2b0859c2f015) introduced any usage of it.

Maybe we need to deprecate and eventually remove that parameter instead.

lemisky:
After accepting this PR, the code works fine and it is sometimes more convenient to set the parameters directly using the dictionary, which newcomers may not know they need the `Settings` instance parameter.

Gallaecio:
I think this may need a bit of community discussion.

I also worry that it would not be obvious for users of this method how project settings are ignored when settings are passed here, i.e. `settings={}` and `settings=None` work *very* differently, or so it seems from a quick look (did not actually test it).

Also, mind that there are alternatives:
- Use a shell script instead of a Python method for command execution. Use `-s` to define settings (without ignoring project settings)
- Define settings in the spider itself (`custom_settings`)

PR #5534, opened 2022-06-17
fine-tune the code

lemisky:
This is a better way to write

### DIFF
--- a/scrapy/commands/runspider.py
+++ b/scrapy/commands/runspider.py
@@ -14,7 +14,7 @@ def _import_file(filepath):
     if fext not in ('.py', '.pyw'):
         raise ValueError(f"Not a Python source file: {abspath}")
     if dirname:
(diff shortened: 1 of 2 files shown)


scrapy/commands/runspider.py
```diff
@@ -14,7 +14,7 @@ def _import_file(filepath):
     if fext not in ('.py', '.pyw'):
         raise ValueError(f"Not a Python source file: {abspath}")
     if dirname:
-        sys.path = [dirname] + sys.path
+        sys.path.insert(0, dirname)
```
Gallaecio, line 17: While I would probably have written it your way if writing this code from scratch, I am not sure this is worth changing. There is no functional or significant performance difference, and this is probably out of the scope of any style guide.

lemisky: Okay, I understand. Thank you for your reply.

PR #5536, opened 2022-06-17
Pin documentation requirements

Gallaecio:
https://github.com/scrapy/scrapy/pull/5535#discussion_r899994814


kmike (approved):
👍 

What do you think about applying the same fix to the master?

wRAR (approved):

PR #5541, opened 2022-06-23
Update README.md

cyai:
Added line breakers to make the documentation look cleaner.


Gallaecio (changes requested):
I don’t think there is a significant readability issue at the moment (the total content is quite short), and I don’t think this would significantly improve readability either (I think merging sections would be more effective).

PR #5542, opened 2022-06-23
MAINT: Updated f-string format

cyai:
Updated the code with the f-string method for better and cleaner understanding.


wRAR (approved):

kmike:
Thanks @cyai!

PR #5545, opened 2022-06-28
Deeper method finding for requests to/from dict

JakeCowton:
When converting a request to or from a dictionary, limited searching is done for the corresponding callback. This PR searches deeper for the callback function.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,4 +1,0 @@
-For information about installing Scrapy see:
-
-* docs/intro/install.rst (local file)
-* https://docs.scrapy.org/en/latest/intro/install.html (online version)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,4 @@
+For information about installing Scrapy see:
+
+* [Local docs](docs/intro/install.rst)
+* [Online docs](https://docs.scrapy.org/en/latest/intro/install.html)


Gallaecio:
Would the current implementation work if you did the following?

```
self.bar = foo.bar
```

JakeCowton:
Okay, I've made the example more specific to my situation. `bar`, given a URL, will find some relevant URLs on the page and do Stuff™ to them. It does this recursively for each URL it finds. It works for the first URL passed to it via `BaseSpider._foo_bar`, but subsequent URLs yielded recursively within `Utilities.bar` throws the following.

```python
ValueError: Function <bound method Utilities.bar of <importer.spiders.data.demo.Utilities object at 0x7f50ed6c9dc0>> is not an instance method in: <TheActualSpider 'demo' at 0x7f50ed6c9df0>
```

Here's the minimal implementation:

```python
import scrapy


class AnItem(scrapy.item.Item):
    related_url = scrapy.item.Field()


class Utilities(object):

    def bar(self, response, sub_urls=None, *args, **kwargs):
        if sub_urls is None:
            # Finds some other URLs in response and pass each URLs response
            # to this function
            sub_urls = [
                "http://books.toscrape.com/catalogue/soumission_998/index.html"
            ]
        else:
            return

        # Stuff

        yield response.follow(
            sub_urls.pop(0),
            callback=self.bar,
            cb_kwargs={
                "sub_urls": sub_urls,
            },
            dont_filter=True,
        )


class BaseSpider(object):

    def __init__(self, foo, *args, **kwargs):
        self.foo = foo

    @classmethod
    def from_crawler(cls, crawler, *args, **kwargs):
        foo = Utilities()
        return super().from_crawler(
            crawler, foo, *args, **kwargs
        )

    def _foo_bar(self, *args, **kwargs):
        yield from self.foo.bar(
            *args, **kwargs
        )

    def super_yield(self, response, item):
        yield response.follow(
            url=item.get("related_url"),
            callback=self._foo_bar
        )


class TheActualSpider(BaseSpider, scrapy.spiders.Spider):

    name = "demo"

    start_urls = ["http://books.toscrape.com/"]

    def parse(self, response):
        item = AnItem(
            related_url="http://books.toscrape.com/catalogue/category/books/fiction_10/index.html",
        )
        yield from self.super_yield(
            response,
            item,
        )
```

I appreciate this is very specific to my use case, but it's not a major change and solves a limitation that doesn't need to be there.

Gallaecio:
I see, the problem is that `Utilities.bar` yields requests with `callback=self.bar`.

As a workaround for the new example, you could initialize `Utilities` with a reference to the spider, and then use `self.spider.bar` within `Utilities.bar`.

A cleaner approach for this specific case would be to copy the callback from the original request, `callback=response.request.callback`. It would not work if you wanted a different method of `Utilities` as callback, though.

It may be better to just let your spider be also a subclass of `Utilities` instead (i.e. make `Utilities` a mixin for spider classes). You would be going from composition to multiple inheritance, which you may not want, but Scrapy relies on callbacks being spider methods for serialization to disk, which kind of forces your hand here.

Gallaecio:
> [The fix I am proposing is] not a major change and solves a limitation that doesn't need to be there.

We can still discuss and potentially merge your change if you provide tests covering the change. I think the code itself could use some refactoring (we probably do not need to have 2 loops), but it does not look bad. I do worry that serialization/deserialization may not be reliable in this case, but tests can put my mind at ease.

JakeCowton:
Sure, I'll get those added.

PR #5547, opened 2022-06-30
Deprecate the `scrapy.downloadermiddlewares.decompression` module

### LINKED ISSUE #5546: Deprecate the scrapy.downloadermiddlewares.decompression module
It is not used by default, not documented, and I found no relevant search results on the internet.

Laerte:
Closes #5546

### DIFF
--- a/scrapy/downloadermiddlewares/decompression.py
+++ b/scrapy/downloadermiddlewares/decompression.py
@@ -9,13 +9,13 @@ import tarfile
 import zipfile
 from io import BytesIO
 from tempfile import mktemp
-import warnings
+from warnings import warn
 
 from scrapy.exceptions import ScrapyDeprecationWarning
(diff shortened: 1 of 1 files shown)


Gallaecio:
We have a [policy](https://docs.scrapy.org/en/latest/versioning.html#deprecation-policy) not to remove features right away, and instead issue deprecation warnings for at least 1 year.

So the change at this stage should consist of doing this in the module:

```python
from warnings import warn

warn(
    'scrapy.downloadermiddlewares.decompression is deprecated',
    ScrapyDeprecationWarning,
    stacklevel=2,
)
```

The tests can remain there until the feature is removed.

Laerte:
@Gallaecio Like this?

Gallaecio (approved):
Yes, thank you!

kmike:
Looks good, thanks!

PR #5551, opened 2022-07-09
Update README.rst

cxtruong70:
Minor change to sentence for clarity. The previous version was a little awkward so clarified it with a simple list.


wRAR:
"a wide range of purposes, from data mining to monitoring and automated testing" doesn't sound awkward to me.

PR #5553, opened 2022-07-12
Fix doc: `scrapy.exporter` to `scrapy.exporters`

Rotzbua:
Fix the import in the example.


kmike:
A good catch @Rotzbua!

PR #5554, opened 2022-07-12
[doc] Remove incompatible web service project

Rotzbua:
The mentioned project has problems and should be removed because:

* Abandoned since 2017
* Not compatible with Python3


wRAR:
You should also remove it from index.rst (not sure if there are other changes needed).

Rotzbua:
@wRAR Thanks for the hint. I found references just in the `index.rst`.

wRAR (approved):

kmike:
Thanks @Rotzbua!

PR #5559, opened 2022-07-17
Fixed intersphinx references

kmike:
Build is failing at master because a few pages changed their location. It should be fixed in this PR.


kmike:
I'm trying to address the test failure from this PR in https://github.com/scrapy/scrapy/pull/5560, it's unrelated to docs.

wRAR (approved):

PR #5561, opened 2022-07-17
Less warnings in tests

kmike:
These warnings don't really affect anything, but there's a lot of them at the end of the test run, which make it harder to read the results.

It should be more reasonable after this PR. Some unique warnings are still there; total count is reduced from 569 to 207 warnings (Ubuntu + Python 3.10 env).

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -21,3 +21,6 @@ addopts =
 markers =
     only_asyncio: marks tests as only enabled when --reactor=asyncio is passed
     only_not_asyncio: marks tests as only enabled when --reactor=asyncio is not passed
+filterwarnings =
+    ignore:scrapy.downloadermiddlewares.decompression is deprecated
+    ignore:Module scrapy.utils.reqser is deprecated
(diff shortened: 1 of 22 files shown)


kmike:
This PR is ready for review; I don't plan further updates.

pytest.ini
```diff
@@ -21,3 +21,6 @@ addopts =
 markers =
     only_asyncio: marks tests as only enabled when --reactor=asyncio is passed
     only_not_asyncio: marks tests as only enabled when --reactor=asyncio is not passed
+filterwarnings =
+    ignore:scrapy.downloadermiddlewares.decompression is deprecated
+    ignore:Module scrapy.utils.reqser is deprecated
```
Gallaecio, line 26: I’m OK with this in these specific cases, but in general it may be best to instead filter out those warnings where we import those modules in tests, because if the warnings came from imports outside tests we would want them.

Gallaecio (approved):
:bow:

PR #5563, opened 2022-07-18
remove compatibility code from tests for the case dataclasses module is not available

kmike:
It was Python 3.6 compat code, and Python 3.6 support is dropped.

test_engine is not handled in this PR, to avoid conflicts with https://github.com/scrapy/scrapy/pull/5561; it's solved in https://github.com/scrapy/scrapy/pull/5561/commits/f3801f1126a42682b9a8188e64345e524295c38c.

### DIFF
--- a/tests/test_exporters.py
+++ b/tests/test_exporters.py
@@ -4,6 +4,7 @@ import marshal
 import pickle
 import tempfile
 import unittest
+import dataclasses
 from io import BytesIO
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


elacuesta (approved):
🚀

wRAR (approved):

kmike:
I knew, I knew that we'll see @elacuesta in this PR, he loves when code gets removed!

PR #4141, opened 2019-11-08
Add StickyMetaParamsMiddleware

heylouiz:
Adds a spider middleware that sticky meta parameters, forwarding
desired keys through requests and responses.

Fixes #3645, closes #3770

PR #5565, opened 2022-07-19
Extend Request.meta documentation

Gallaecio:
While reviewing https://github.com/scrapy/scrapy/pull/4141, I realized the existing `Request.meta` documentation could use some more information.


docs/topics/request-response.rst
```diff
@@ -177,18 +177,48 @@ Request objects
...
-        cloned using the ``copy()`` or ``replace()`` methods, and can also be
-        accessed, in your spider, from the ``response.meta`` attribute.
+        Request metadata can also be accessed through the
+        :attr:`~scrapy.http.Response.meta` attribute of a response.
+
+        To pass data from one spider callback to another, consider using
+        :attr:`cb_kwargs` instead.
+
+        To keep some data across 3 or more spider callbacks, however, request
+        metadata may be the right choice.
```
kmike, line 193: Hm, I think cb_kwargs also works ok in this case - you can pass something like `partial_data` dict or list as a kwarg to all callbacks

kmike (approved):
👍

docs/topics/request-response.rst
Gallaecio: The use case I am trying to describe here is where you pass some specific metadata key to almost every callback, and new future callbacks are likely to get it as well, but often most callbacks only pass it over to the next callback. So, for example, `start_url` if you want to keep track of the root URL that originated every callback.

In those cases, adding the same parameter to every callback, specially to callbacks whose only function is to pass the parameter over to the next callback, may be too verbose, and using meta instead (and eventually `STICKY_META_KEYS`, https://github.com/scrapy/scrapy/pull/4141, which is what inspired this paragraph) may be better.

I do need to come out with a better wording. Let me give it at least 1 more try.

PR #5566, opened 2022-07-19
Set an upper limit for the cache, otherwise it will cause oom

Duckweeds7:
Sorry, this is my first time to submit PR for a large project, so it's a little nonstandard, this is a very serious problem. When a user associates scrapy with Redis or Kafka, or crawls a large number of tasks, it will lead to info.downloaded becomes very large, especially when the exception object is stored, it will cause OOM and cause the crawler to be killed by the system

### DIFF
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -1,6 +1,6 @@
 import functools
 import logging
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from inspect import signature
 from warnings import warn
 
@@ -18,14 +18,13 @@ logger = logging.getLogger(__name__)
 
 
 class MediaPipeline:
-
     LOG_FAILED_RESULTS = True
 
     class SpiderInfo:
         def __init__(self, spider):
             self.spider = spider
             self.downloading = set()
-            self.downloaded = {}
+            self.downloaded = OrderedDict()
             self.waiting = defaultdict(list)
 
     def __init__(self, download_func=None, settings=None):
@@ -40,6 +39,10 @@ class MediaPipeline:
         self.allow_redirects = settings.getbool(
             resolve('MEDIA_ALLOW_REDIRECTS'), False
         )
+        self.cache_downloaded_size = settings.getint(
+            resolve('CACHE_DOWNLOADED_SIZE'), 10000
+        )
+
         self._handle_statuses(self.allow_redirects)
 
         # Check if deprecated methods are being used and make them compatible
@@ -62,9 +65,9 @@ class MediaPipeline:
         class_name = self.__class__.__name__
         formatted_key = f"{class_name.upper()}_{key}"
         if (
-            not base_class_name
-            or class_name == base_class_name
-            or settings and not settings.get(formatted_key)
+                not base_class_name
+                or class_name == base_class_name
+                or settings and not settings.get(formatted_key)
         ):
             return key
         return formatted_key
@@ -115,7 +118,7 @@ class MediaPipeline:
         dfd.addBoth(self._cache_result_and_execute_waiters, fp, info)
         dfd.addErrback(lambda f: logger.error(
             f.value, exc_info=failure_to_exc_info(f), extra={'spider': info.spider})
-        )
+                       )
         return dfd.addBoth(lambda _: wad)  # it must return wad at last
 
     def _make_compatible(self):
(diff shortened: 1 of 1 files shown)


scrapy/pipelines/media.py
```diff
@@ -1,6 +1,6 @@
 import functools
 import logging
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
```
Gallaecio, line 3: We can use `scrapy.utils.datatypes.LocalCache` instead.

scrapy/pipelines/media.py
```diff
@@ -40,6 +39,10 @@ def __init__(self, download_func=None, settings=None):
         self.allow_redirects = settings.getbool(
             resolve('MEDIA_ALLOW_REDIRECTS'), False
         )
+        self.cache_downloaded_size = settings.getint(
+            resolve('CACHE_DOWNLOADED_SIZE'), 10000
```
Gallaecio, line 43: What about `MEDIA_CACHE_SIZE`?

scrapy/pipelines/media.py
```diff
@@ -62,9 +65,9 @@ def _key_for_pipe(self, key, base_class_name=None, settings=None):
         class_name = self.__class__.__name__
         formatted_key = f"{class_name.upper()}_{key}"
         if (
-            not base_class_name
-            or class_name == base_class_name
-            or settings and not settings.get(formatted_key)
+                not base_class_name
+                or class_name == base_class_name
+                or settings and not settings.get(formatted_key)
```
Gallaecio, line 70: Here and some lines below you are breaking good style for no apparent reason.

Gallaecio (commented):
Makes sense to me.

We do need to:
- Add a test for this feature, which may not be trivial.
- Cover the new setting in the documentation, similarly to how `MEDIA_ALLOW_REDIRECTS` is covered.

It would also be great if we had a documentation topic specifically about optimizing Scrapy, where we should cover this as one of the settings to consider if memory usage is too high and a media pipeline is being used. However, there is no such topic yet, and creating one would be out of scope here.

Gallaecio:
I see you have removed the setting now, not sure why. I think it would be best to keep it.

It does complicate the implementation, though. When creating an instance of SpiderInfo, you should pass the new value to it, either the cache object itself or the limit value. I recommend for it to be a keyword-only parameter. And because SpiderInfo is a public class (not prefixed with `_`), it is possible that some users are subclassing it, in which case you need to make things work for cases where SpiderInfo does not accept the extra parameter; you could capture the exception due to the unexpected argument, log a warning, and create the object again without the extra parameter. We have a few examples in the Scrapy codebase of this approach.

GeorgeA92:
@Gallaecio @Duckweeds7  I have some concers about it
> or crawls a large number of tasks, it will lead to info.downloaded becomes very large

If total size of all downloaded and not processed responses exceed value of `SCRAPER_SLOT_MAX_ACTIVE_SIZE` (by default - 5mb) - in this case scrapy will not send new requests until application will process already downloaded responses.
It means that on **general case** with current 2.6.x version of scrapy It is not expected that total size of stored responses exceed value of `SCRAPER_SLOT_MAX_ACTIVE_SIZE` setting.
It also means that size of responses stored in `info.downloaded` is not expected to exceed value of `SCRAPER_SLOT_MAX_ACTIVE_SIZE` setting.

Duckweeds7:
> I see you have removed the setting now, not sure why. I think it would be best to keep it.
> 
> It does complicate the implementation, though. When creating an instance of SpiderInfo, you should pass the new value to it, either the cache object itself or the limit value. I recommend for it to be a keyword-only parameter. And because SpiderInfo is a public class (not prefixed with `_`), it is possible that some users are subclassing it, in which case you need to make things work for cases where SpiderInfo does not accept the extra parameter; you could capture the exception due to the unexpected argument, log a warning, and create the object again without the extra parameter. We have a few examples in the Scrapy codebase of this approach.

You're right,I need a little time to learn and revise my commit.

Duckweeds7:
> > It also means that size of responses stored in `info.downloaded` is not expected to exceed value of `SCRAPER_SLOT_MAX_ACTIVE_SIZE` setting.
> 
> It is my understanding that `info.downloaded` contains every (media pipeline, i.e. file or image) response found during the crawl, even after their corresponding response objects get destroyed.

Yes, even if the response object is cleared, it will still be saved, and the reason for this problem in our production environment is that we have rewritten the exception object, which will cause the volume of the exception object to be too large, which occurs OOM after running for a period of time.

PR #5571, opened 2022-07-22
docs: include reference to ScrapeMate

beyarkay:
Selector Gadget only works on chrome, but [ScrapeMate](https://addons.mozilla.org/en-US/firefox/addon/scrapemate/) works on firefox and does a similar job.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -312,10 +312,11 @@ the response page from the shell in your web browser using ``view(response)``.
 You can use your browser's developer tools to inspect the HTML and come up
 with a selector (see :ref:`topics-developer-tools`).
 
-`Selector Gadget`_ is also a nice tool to quickly find CSS selector for
-visually selected elements, which works in many browsers.
+`Selector Gadget`_ (for Chrome) or `Scrape Mate`_ (for FireFox) are also nice tools
+to quickly find CSS selector for visually selected elements.
 
 .. _Selector Gadget: https://selectorgadget.com/
+.. _Scrape Mate: https://addons.mozilla.org/en-US/firefox/addon/scrapemate/
 
 
 XPath: a brief intro


docs/intro/tutorial.rst
```diff
@@ -312,10 +312,11 @@ the response page from the shell in your web browser using ``view(response)``.
 You can use your browser's developer tools to inspect the HTML and come up
 with a selector (see :ref:`topics-developer-tools`).
 
-`Selector Gadget`_ is also a nice tool to quickly find CSS selector for
-visually selected elements, which works in many browsers.
+`Selector Gadget`_ (for Chrome) or `Scrape Mate`_ (for FireFox) are also nice tools
+to quickly find CSS selector for visually selected elements.
 
 .. _Selector Gadget: https://selectorgadget.com/
+.. _Scrape Mate: https://addons.mozilla.org/en-US/firefox/addon/scrapemate/
```
Gallaecio, line 319: I propose to feature “ScrapeMate” only, given it works in both browsers, and is based on Selector Gadget.

I would keep the old text, replacing the product name, updating the URL, and removing the “, which works in many browsers” which was and would still be false.

beyarkay: Fair enough, I'll make the change and will include the [Chrome link](https://chrome.google.com/webstore/detail/scrapemate-beta/daiomapeacamgnofkkmiaollhidcndld).

ChrisHC:
Seems like the Chrome version of ScrapeMate is not available anymore - the link leads to a [404 error](https://chrome.google.com/webstore/detail/scrapemate-beta/daiomapeacamgnofkkmiaollhidcndld). I hope that is only temporarily.

[SelectorGadget](https://chrome.google.com/webstore/detail/selectorgadget/mhjhnkcfbdhnjickkkdbjoemdmbfginb) for Chrome is still up and working.

PR #5574, opened 2022-07-25
Update frozen CI packages

Gallaecio:


wRAR (approved):

kmike:
Should we also upgrade mypy and related packages (if they're required)?

Gallaecio:
I was following https://github.com/scrapy/scrapy/wiki/Scrapy-release-procedure#upgrade-ci-packages, which used to list specific packages. I have updated it now to mention the files where they are defined instead, so we can find them looking for `==` instead, in future releases.

I will create later a follow-up pull request for the remaining frozen packages. Mind, however, that these updates were not supposed to happen after patch versions in the first place, only after minor and major versions :sweat_smile:

PR #5577, opened 2022-07-26
add async callbacks support to parse 

### LINKED ISSUE #5424: scrapy parse doesn't support async callbacks
In master when running `scrapy parse` for a spider with `async def parse` the page is downloaded but then the spider hangs. In #4978  it instead raises `TypeError: 'async_generator' object is not iterable`. Both problems happen because the parse command calls `iterate_spider_output` and doesn't expect a Deferred or an async iterator.

mtabbasi:
Fixes #5424

### DIFF
--- a/tests/test_command_parse.py
+++ b/tests/test_command_parse.py
@@ -29,7 +29,15 @@ class ParseCommandTest(ProcessTest, SiteTest, CommandTest):
 import scrapy
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
+from scrapy.utils.test import get_from_asyncio_queue
 
+class AsyncDefAsyncioSpider(scrapy.Spider):
+
+    name = 'asyncdef{self.spider_name}'
+
+    async def parse(self, response):
+        status = await get_from_asyncio_queue(response.status)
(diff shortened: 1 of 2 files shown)


tests/test_command_parse.py
```diff
@@ -29,7 +29,15 @@ def setUp(self):
 import scrapy
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
+from scrapy.utils.test import get_from_asyncio_queue
 
+class AsyncDefAsyncioSpider(scrapy.Spider):
+
+    name = 'asyncdef{self.spider_name}'
```
wRAR, line 36: This is not an f-string :)
But I don't think you need a dynamic value here anyway.

wRAR (approved):

Gallaecio:
Looks good to me, but I wonder why the coverage data suggest the new test is not reaching the moved code.

wRAR:
@Gallaecio the test launches a separate `scrapy parse` process, and I guess coverage for subprocesses is not collected?

Gallaecio:
Indeed. I thought coverage was able to pick that up anyway, but it seems not. The same code before the changes was not in coverage either.

Gallaecio:
Thank you!

Issue #5500, opened 2022-05-14
Postprocessing feeds do not work for S3 feed storage


Nykakin:
### Description

Example settings:
```
FEEDS = {
    "s3://base/file_%(batch_id)05d.gz": {
        "format": "csv",
        "postprocessing": [GzipPlugin],
        "gzip_compresslevel": 5,
    },
}
FEED_EXPORT_BATCH_ITEM_COUNT = 50000
```
This causes an exception:

```python
2022-05-14 01:02:12 [scrapy.extensions.feedexport] ERROR: Error storing csv feed (15 items) in: s3://base/file_00001.gz
Traceback (most recent call last):
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/threadpool.py", line 244, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/threadpool.py", line 260, in <lambda>
    inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/context.py", line 117, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/context.py", line 82, in callWithContext
    return func(*args, **kw)
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/scrapy/extensions/feedexport.py", line 194, in _store_in_thread
    file.seek(0)
io.UnsupportedOperation: seek
```
Apparently `scrapy.extensions.postprocessing.PostProcessingManager` doesn't fully implement file protocol. Adding this method to the class:
```python
    def seek(self, offset, whence=SEEK_SET):
        return self.file.seek(offset, whence)
```
Cause an exception in a different place:
```python
Traceback (most recent call last):
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/threadpool.py", line 244, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/home/mariusz/Documents/Praca/venv_wgsn/lib/python3.10/site-packages/twisted/python/threadpool.py", line 260, in <lambda>
[...]

PR #5581, opened 2022-07-28
Fix Feed Exporter issue with Post Processing

felipeboffnunes:
#5500 

It fixes the issue above. It happens because when dealing with Post Processing, _file_ is wrapped around a `PostProcessingManager`, which becomes the *arg for the following procedures that will happen in `FeedExporter`. Those procedures believe they are handling directly the file: this is not the case when Post Processing is involved, since the file becomes an attribute of the `PostProcessingManager`  in these cases.

We need to test this, but it's a very straightforward bug nevertheless.

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -350,13 +350,19 @@ class FeedExporter:
         return defer.DeferredList(deferred_list) if deferred_list else None
 
     def _close_slot(self, slot, spider):
+
+        def get_file(slot_):
+            if isinstance(slot_.file, PostProcessingManager):
+                return slot_.file.file
+            return slot_.file
+
         if not slot.itemcount and not slot.store_empty:
             # We need to call slot.storage.store nonetheless to get the file
             # properly closed.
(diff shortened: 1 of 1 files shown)


felipeboffnunes:
This is under discussion with @Gallaecio.

scrapy/extensions/feedexport.py
```diff
@@ -350,13 +350,19 @@ def close_spider(self, spider):
         return defer.DeferredList(deferred_list) if deferred_list else None
 
     def _close_slot(self, slot, spider):
+
+        def get_file(slot_):
+            if isinstance(slot_.file, PostProcessingManager):
+                return slot_.file.file
```
Gallaecio, line 356: After a quick glance at the failures, I am thinking we may need to close PostProcessingManager without closing the underlying file. Calling `slot_.close()` here before `return` may be enough.

Gallaecio:
Proposed test: https://github.com/felipeboffnunes/scrapy/pull/1

Gallaecio (approved):
_Nice_

wRAR (approved):

PR #5589, opened 2022-08-02
Remove deprecated spider parameter

### LINKED ISSUE #5588: Scrapy warns about itself
I'm using Scrapy 2.6.2. Created a new project, run a spider, and there is 

```
2022-08-02 23:50:26 [py.warnings] WARNING: [redacted]/python3.8/site-packages/scrapy/core/engine.py:279: ScrapyDeprecationWarning: Passing a 'spider' argument to ExecutionEngine.download is deprecated
  return self.download(result, spider) if isinstance(result, Request) else result
```

message in logs. We also have the same warning in our testing suite, but I haven't realized it's not a test-only issue.

Laerte:
Closes #5588


Laerte:
The error in CI is not related with the change made.

wRAR (approved):

kmike:
If I'm not mistaken, this parameter is deprecated in https://github.com/scrapy/scrapy/pull/5090; it was released with Scrapy 2.6.0 (https://docs.scrapy.org/en/latest/news.html#scrapy-2-6-0-2022-03-01), which happened 2022-03-01. According to https://docs.scrapy.org/en/latest/versioning.html#deprecation-policy, we promise at least one year before the deprecated code is removed, so unfortunately we can't remove it now.

Gallaecio:
Instead, we need to capture and silence that warning when triggered by our own code.

Laerte:
@Gallaecio I will update the code to this approach.

Gallaecio:
I would like to first understand why it is happening in the first place, though. Maybe there is a better way.

Gallaecio:
OK, so I think we need to change `download` so that:
* `spider = self.spider` never happens.
* `if spider is None` changes to `if self.spider is None`.

Hopefully that solves the issue.

Laerte:
@Gallaecio Hope i getted it the idea.

scrapy/core/engine.py
```diff
@@ -257,19 +257,17 @@ def _schedule_request(self, request: Request, spider: Spider) -> None:
...
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
             if spider is not self.spider:
                 logger.warning("The spider '%s' does not match the open spider", spider.name)
-        if spider is None:
+        if self.spider is None:
             raise RuntimeError(f"No open spider to crawl: {request}")
-        return self._download(request, spider).addBoth(self._downloaded, request, spider)
+        return self._download(request, self.spider).addBoth(self._downloaded, request, self.spider)
```
Gallaecio, line 270: I have not tested it, but I believe here we should be passing `spider`, as before, and if it is `None` the corresponding callees will already use `self.spider` in that case.

Laerte: I changed because i got a type hint error in 3.7 check, but can rollback, what you think?

scrapy/core/engine.py
```diff
@@ -257,19 +257,17 @@ def _schedule_request(self, request: Request, spider: Spider) -> None:
...
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
             if spider is not self.spider:
                 logger.warning("The spider '%s' does not match the open spider", spider.name)
-        if spider is None:
+        if self.spider is None:
             raise RuntimeError(f"No open spider to crawl: {request}")
-        return self._download(request, spider).addBoth(self._downloaded, request, spider)
+        return self._download(request, spider).addBoth(self._downloaded, request, spider)  # type: ignore
```
Gallaecio, line 270: I see…

It looks like we should change the typing of `Spider` in those 2 methods to `Optional[Spider]`, instead of ignoring the typing issue.

It also seems like `_download` should get the `if spider is None: spider = self.spider` logic.

Laerte: Yeah i think the same, done.

Gallaecio:
Looks good to me, thanks! :rocket:

How do you feel about test coverage? Do you think you can add at least a test to verify that the exception is not logged with default settings? If you do not see it clearly, I can try and have a look later.

Laerte:
> Looks good to me, thanks! 🚀
> 
> How do you feel about test coverage? Do you think you can add at least a test to verify that the exception is not logged with default settings? If you do not see it clearly, I can try and have a look later.

I'm vacations, if you can add tests i will be glad, thanks!

scrapy/core/engine.py
```diff
@@ -257,17 +257,15 @@ def _schedule_request(self, request: Request, spider: Spider) -> None:
...
+        if spider is not None:
             warnings.warn(
                 "Passing a 'spider' argument to ExecutionEngine.download is deprecated",
                 category=ScrapyDeprecationWarning,
                 stacklevel=2,
             )
             if spider is not self.spider:
                 logger.warning("The spider '%s' does not match the open spider", spider.name)
-        if spider is None:
+        if self.spider is None:
```
kmike, line 268: Hm, it looks like the behavior is changed in this PR. Previously, if `spider` is passed to `download` method, it is always used. Now a RuntimeError could be raised if `spider` is passed, and `self.spider` is None. Is it a problem?

Gallaecio: I am unsure, but I will look into that when I find some time to work on tests.

Laerte: For what i understood from PR:

- #5090 

This will be never the case it seems, since the spider instance saved `ExecutionEngine.spider` after calling `open_spider` will be always available, did i get correctly?

Laerte: @kmike @Gallaecio Lookin in all the tests we always pass the spider instance in `open_spider` (except `tests/test_pipeline_files`, `tests/test_pipeline_media.py`), the test that we can add should be the one to test `RuntimeError` exception then?

hernandezj1:
Thanks for this fix. It worked great!

kmike:
Thanks @Laerte for keeping the PR up-to-date, this helps a lot!

Issue #5579, opened 2022-07-28
Improve the overwrite feed option documentation


Gallaecio:
The [`overwrite` feed option](https://docs.scrapy.org/en/latest/topics/feed-exports.html#feed-options) has a different default value depending on the storage backend used, and we currently cover its default value for all backeds except for [Google Cloud Storage (GCS)](https://docs.scrapy.org/en/latest/topics/feed-exports.html#google-cloud-storage-gcs). We should address that.

Also, since `overwrite=True` is a potentially data-deleting operation, it may be a good idea to mention the fact that `overwrite=True` in the sections about storage backends with that default value, and not just in the section about the `overwrite` feed option.

limury:
Hi! I'd like to take on this issue. I think that GCS also does not allow appending to the storage. Also saw that the GCS initializer does not include a `feed_options` parameter so I'll add that!

PR #5593, opened 2022-08-08
added feed_options parameter to GCS feed initializer and added logging/documentation about the overwrite=True parameter

limury:
- added documentation warning that using the overwrite parameter in the IFeedStorage class can cause data deletion
- Added the feed_options parameter to the GCS Feed storage class initializer
- Added logging in the GCS feed storage class saying that overwrite=False is not available

Resolves https://github.com/scrapy/scrapy/issues/5579

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -172,7 +172,7 @@ class S3FeedStorage(BlockingFeedStorage):
             aws_secret_access_key=self.secret_key,
             aws_session_token=self.session_token,
             endpoint_url=self.endpoint_url)
-        if feed_options and feed_options.get('overwrite', True) is False:
+        if feed_options is not None and feed_options.get('overwrite', True) is False:
             logger.warning('S3 does not support appending to files. To '
                            'suppress this warning, remove the overwrite '
                            'option from your FEEDS setting or set it to True.')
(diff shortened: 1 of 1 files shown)


sisco0:
Solves https://github.com/scrapy/scrapy/issues/5579

wRAR (approved):

scrapy/extensions/feedexport.py
```diff
@@ -86,7 +86,9 @@ class IFeedStorage(Interface):
 
     def __init__(uri, *, feed_options=None):
         """Initialize the storage with the parameters given in the URI and the
-        feed-specific options (see :setting:`FEEDS`)"""
+        feed-specific options (see :setting:`FEEDS`).
+        CAREFUL: in `feed_options` the default `overwrite=True` will cause you to 
+        lose the previous version of your data"""
```
Gallaecio, line 91: This would need some format adjustments to adhere to reStructuredText syntax, but more importantly this is not accurate here (overwrite is not True for IFeedStorage by default ), and this is not where we need to cover this, since this docstring is not currently rendered into our documentation.

We need to update `docs/topics/feed-exports.rst` instead, both the section about the `FEEDS` option, to mention the default value for GCS, and the sections of the specific storage backends that use overwrite by default, to add a warning like the one you are adding here.

PR #5598, opened 2022-08-15
prevent retry_times meta key from moving to other requests

ounissi-zakaria:
# **The problem:**
    When a request fails, and it gets retries by retry middleware a `retry_times` key is added to the meta of the request to keep count of how many times we retried the request. The problem with this is when a request gets retried N times, and then it passes, and we process the response and send other requests giving them the meta of the response their retry counter will start from N (where the first request succeeded)  example in a real spider beside the test i added:

```python
import scrapy


    name = "test_retry"
    custom_settings = {
        "RETRY_TIMES": 1,
        "DOWNLOADER_MIDDLEWARES": {"playground.middlewares.FakeResponseMiddleware": 10},
    }

    def start_requests(self):
        yield scrapy.Request(
            "http://www.example.com/?page=1",
            callback=self.parse_page1,
            meta={"important_data": "important_data"},
        )

    def parse_page1(self, response):
        # Do something with response
        # and send another request
        yield scrapy.Request(
            "http://www.example.com/?page=2",
            callback=self.parse_page2,
            meta=response.meta,
        )

    def parse_page2(self, response):
        # we will never reach this
        pass

```
with the FakeResponseMiddleware to fake failing responses

```python
# fail if its first time seeing the request
class FakeResponseMiddleware:
    seen_urls = []

        if request.url in self.seen_urls:
            return Response(url=request.url, request=request, status=200)
        else:
            self.seen_urls.append(request.url)
            return Response(url=request.url, request=request, status=503)

```
When someone uses the middleware they would think that every single request will be retried the whole `RETRY_TIMES` 

# **Proposed solution** ?
At first, I thought about something like this 
```python

    # in process response of the retry middleware
        if request.meta.get('dont_retry', False):
            return response
        if response.status in self.retry_http_codes:
            reason = response_status_message(response.status)
            return self._retry(request, reason, spider) or response
        # when the response passes delete retry_times
        if 'retry_times' in request.meta:
            del request.meta['retry_times'] # or just set it to zero
        return response
```

but then after looking at some tests I realized that it will break some of them, and if it will break some tests it will definitely break someone's code (maybe they are using the key to see if the request was retried and how many times ?).

so i think a better solution would be to keep it and a `process_request` check:
```python
    # in retry middleware
        if not request.meta.get("_is_retrying", False):
            request.meta["retry_times"] = 0  # or delete it

        request.meta["_is_retrying"] = False
        if request.meta.get('dont_retry', False):
            return response
        if response.status in self.retry_http_codes:
            reason = response_status_message(response.status)
            request.meta["_is_retrying"] = True
            return self._retry(request, reason, spider) or response
        return response
```

For now, I committed the test only, and I can commit the fix when you give me the green light

### DIFF
--- a/tests/test_downloadermiddleware_retry.py
+++ b/tests/test_downloadermiddleware_retry.py
@@ -119,6 +119,27 @@ class RetryTest(unittest.TestCase):
         req = self.mw.process_exception(req, exception, self.spider)
         self.assertEqual(req, None)
 
+    def test_separate_requests_fail(self):
+        # we send a request and it doesn't pass
+        req1 = Request('http://www.example.com', meta={"important_data": "data_to_pass"})
+        rsp1 = Response('http://www.example.com', body=b'', status=503, request=req1)
+        # we retry it
+        req1 = self.mw.process_response(req1, rsp1, self.spider)
+        assert isinstance(req1, Request)
+
+        # it passes
+        rsp1 = Response('http://www.example.com', body=b'', status=200, request=req1)
+        rsp1 = self.mw.process_response(req1, rsp1, self.spider)
+        assert isinstance(rsp1, Response)
+
+        # we send a new request that carries data from first request meta
+        req2 = Request('http://www.example.com?p=2', meta=rsp1.meta)
+        rsp2 = Response('http://www.example.com?p=2', body=b'', status=503, request=req2)
+        req2 = self.mw.process_response(req2, rsp2, self.spider)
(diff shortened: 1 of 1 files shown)


wRAR:
> send other requests giving them the meta of the response

This is always a bad idea, not specific to RetryMiddleware.

Gallaecio:
See https://github.com/scrapy/scrapy/issues/3645

PR #5608, opened 2022-08-25
Changed incorrect info in documentation regarding the return type of parse method.

### LINKED ISSUE #5602: Clarify the return type of callbacks
https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.Spider.parse says "This method, as well as any other Request callback, must return an iterable of Request and/or item objects.", but it can also return a single Request/an item object. It's possible there are other places with a similar omission.

zaid-ismail031:
Fixes #5602. I could not find any other similar errors (i.e. places where it says that the return type of a `Request` callback function must only be an iterable).

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -181,9 +181,10 @@ scrapy.Spider
        scraped data and/or more URLs to follow. Other Requests callbacks have
        the same requirements as the :class:`Spider` class.
 
-       This method, as well as any other Request callback, must return an
-       iterable of :class:`~scrapy.Request` and/or :ref:`item objects
-       <topics-items>`.
+       This method, as well as any other Request callback, must return a
+       :class:`~scrapy.Request` object, an :ref:`item object <topics-items>`, an
+       iterable of :class:`~scrapy.Request` objects and/or :ref:`item objects
+       <topics-items>`, or ``None``.
 
        :param response: the response to parse
        :type response: :class:`~scrapy.http.Response`


docs/topics/spiders.rst
```diff
@@ -181,9 +181,9 @@ scrapy.Spider
        scraped data and/or more URLs to follow. Other Requests callbacks have
        the same requirements as the :class:`Spider` class.
 
-       This method, as well as any other Request callback, must return an
-       iterable of :class:`~scrapy.Request` and/or :ref:`item objects
-       <topics-items>`.
+       This method, as well as any other Request callback, must return
+       :class:`~scrapy.Request` objects, :ref:`item objects <topics-items>`, or
```
wRAR, line 185: I'd change this to "a Request object, an item object [...]" (keeping proper formatting of course), as it can only return one object if not returning an iterable.

zaid-ismail031: Okay, I've made the changes.

wRAR (approved):

docs/topics/spiders.rst
```diff
@@ -181,9 +181,9 @@ scrapy.Spider
        scraped data and/or more URLs to follow. Other Requests callbacks have
        the same requirements as the :class:`Spider` class.
 
-       This method, as well as any other Request callback, must return an
-       iterable of :class:`~scrapy.Request` and/or :ref:`item objects
-       <topics-items>`.
+       This method, as well as any other Request callback, must return a
+       :class:`~scrapy.Request` object, an :ref:`item object <topics-items>`, or
+       an iterable of these objects.
```
Gallaecio, line 186: I think with the current proposal someone could understand that the iterable is an iterable of 1 request and 1 item.

I would rather be verbose than leave room for interpretation, specially since I believe this is not the first time someone ask for clarification here:

```suggestion
       an iterable of :class:`~scrapy.Request` and/or :ref:`item 
       <topics-items>` objects.
```

zaid-ismail031: Okay. I have changed the wording to be more verbose.

elacuesta:
I think we have to explicitly mention `None` too, since it's also a valid return value.

(Based on [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), since we're using the word "must" in this phrase)

zaid-ismail031 (commented):
@elacuesta Are you okay with the wording here?

Gallaecio (approved):

kmike:
Thanks @zaid-ismail031!

PR #5611, opened 2022-08-28
BOM should take precedence over Content-Type header when detecting the encoding

### LINKED ISSUE #5601: BOM should take precedence over Content-Type header when detecting the encoding
Currently Scrapy uses headers first to detect the encoding. But browsers actually put a higher priority for BOM; this is also in WHATWG [standard](https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding). It can be checked e.g. by running this server, and opening URL in a browser - UTF-8 is used by browser, but cp1251 is used by Scrapy:

```py
import codecs
from http.server import BaseHTTPRequestHandler
from http.server import HTTPServer


class HttpGetHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-type", "text/html; charset=cp1251")
        self.end_headers()
        self.wfile.write(codecs.BOM_UTF8)
        self.wfile.write("<!DOCTYPE html>".encode('utf8'))
        self.wfile.write("Привет!".encode('utf8'))


if __name__ == '__main__':
    httpd = HTTPServer(('', 8000), HttpGetHandler)
    httpd.serve_forever()
```

When opening this page in a browser, it shows "Привет!".

Spider code to check it:

```py
import scrapy


    name = "tst"

    start_urls = ["http://0.0.0.0:8000"]

    def parse(self, response):
        return {"encoding": response.encoding, "text": response.text}


if __name__ == '__main__':
    process = CrawlerProcess()
    process.crawl(MySpider)
    process.start()
```

Spider outputs

> {'encoding': 'cp1251', 'text': 'п»ї<!DOCTYPE html>РџСЂРёРІРµС‚!'}

See also: https://github.com/scrapy/w3lib/issues/189 - it's a similar issue, but fixing it in w3lib is not enough to make it working in Scrapy.

mtabbasi:
closes #5601

### DIFF
--- a/scrapy/http/response/text.py
+++ b/scrapy/http/response/text.py
@@ -11,8 +11,13 @@ from typing import Generator, Tuple
 from urllib.parse import urljoin
 
 import parsel
-from w3lib.encoding import (html_body_declared_encoding, html_to_unicode,
-                            http_content_type_encoding, resolve_encoding)
+from w3lib.encoding import (
+    html_body_declared_encoding,
+    html_to_unicode,
+    http_content_type_encoding,
+    resolve_encoding,
+    read_bom,
+)
 from w3lib.html import strip_html5_whitespace
 
 from scrapy.http import Request
@@ -60,6 +65,7 @@ class TextResponse(Response):
     def _declared_encoding(self):
         return (
             self._encoding
+            or self._bom_encoding()
             or self._headers_encoding()
             or self._body_declared_encoding()
         )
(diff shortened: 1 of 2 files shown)


Gallaecio (approved):
Looks good to me. Could you add a test for it?

mtabbasi:
Sure

Gallaecio (approved):
Nice work!

kmike:
Thanks @mtabbasi! I like how clean is this PR.

Issue #5612, opened 2022-08-29
A test fails with new w3lib


wRAR:
`tests/test_spider.py::CrawlSpiderTest::test_process_request_instance_method` requires the host in the URL to keep uppercase which no longer happens because of https://github.com/scrapy/w3lib/issues/188 (which I don't consider to be a bug, and which could happen in any other place that uses `urlsplit`).

PR #5617, opened 2022-09-06
Fix failed tests related to w3lib

Laerte:
This fix failed tests from this issue: https://github.com/scrapy/scrapy/issues/5612

~- tox.wiki is offline that's why checks is failing (https://github.com/tox-dev/tox/issues/2486)~

### DIFF
--- a/tests/test_spider.py
+++ b/tests/test_spider.py
@@ -22,8 +22,7 @@ from scrapy.spiders import (
 from scrapy.linkextractors import LinkExtractor
 from scrapy.utils.test import get_crawler
 from tests import get_testdata
-from pkg_resources import parse_version
-from w3lib import __version__ as w3lib_version
+from w3lib.url import safe_url_string
 
 
 class SpiderTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


Gallaecio (commented):
I am thinking we should skip the tests instead, with `@pytest.mark.skipIf(…, strict=True, reason="w3lib 2.0.0 and later do not allow invalid domains.")`.

Also, while I don’t think we need to address it here, we need to keep the target issue open or create a new one to actually figure out how to test IPv6 support now.

Laerte:
@Gallaecio Done.

tests/test_spider.py
```diff
@@ -360,10 +362,12 @@ def process_request_upper(self, request, response):
...
-        self.assertEqual([r.url for r in output],
-                         ['http://EXAMPLE.ORG/SOMEPAGE/ITEM/12.HTML',
-                          'http://EXAMPLE.ORG/ABOUT.HTML',
-                          'http://EXAMPLE.ORG/NOFOLLOW.HTML'])
+        urls = ['http://EXAMPLE.ORG/SOMEPAGE/ITEM/12.HTML',
+                'http://EXAMPLE.ORG/ABOUT.HTML',
+                'http://EXAMPLE.ORG/NOFOLLOW.HTML']
+        if parse_version(w3lib_version) >= parse_version('2.0.0'):
+            urls = list(map(lambda u: u.replace("EXAMPLE.ORG", "example.org"), urls))
+        self.assertEqual([r.url for r in output], urls)
```
Gallaecio, line 370: I ask as someone completely unfamiliar with this test: would it make sense to change the test so that valid domains are used? Or is the use of invalid domains page of what the test is checking for? And, if the later, why? Should we also skip this test with w3lib 2+ and plan to think of a way to test this on w3lib 2+ later?

wRAR: I've looked at the test and thought that it just tests that the `process_request_upper()` method actually ran, by checking that the URLs were modified. Assuming this is true, it can be changed to some other modification.

Laerte: @wRAR @Gallaecio  Maybe we can use `safe_url_string` for each url, with this the output will be standarized for all `w3lib` versions. What you guys think?

Gallaecio: https://github.com/scrapy/scrapy/pull/5617/commits/582a6bf6dbcb01da40cbec6b51269add2db39cf1 is not what I thought you meant, I thought you meant using safe_url_string in the method that uppercases the URLs originally.

That said, it seems like a very valid alternative to me. Great job!

Gallaecio (approved):

tests/test_spider.py
Gallaecio: @wRAR Feel free to merge if you are also OK with the change.

PR #5624, opened 2022-09-13
fix: typo in tutorial

magsen:
fix: typo in tutorial

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -379,7 +379,7 @@ like this:
 Let's open up scrapy shell and play a bit to find out how to extract the data
 we want::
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
I believe that is done on purpose, because in the past some people have accidentally try to execute that on the interactive Python interpreter (`>>>`).

magsen:
@Gallaecio Maybe but the others bash command in the same file don't have the $ sign. It is the only one.
So we must put a dollar before all bash command or none.

Gallaecio:
Oh, you are right! Well, consistency triumphs most things in my book :)

Gallaecio (approved):

PR #5625, opened 2022-09-13
[get_retry_request] Adding ability to set log level

### LINKED ISSUE #4622: Retry logs an error when giving up
# Description

When retry gives up, scrapy is logging an error, which increases `log_count/ERROR` stats.
This key is the standard way to check for errors in scrapy, but a retry giving shouldn't be considered an error.

### Steps to Reproduce

Run the a spider against an invalid URL so it gives up retrying

**Expected behavior:**

As per my understanding, failing to retry a request shouldn't be considered an error.
If the website is down, it is not an issue in our code, but the website.
So, a validation in the project, when the job finishes, checking if any retries gave up and then throw an exception should be the standard solution in a spider (as scrapy doesn't know if it is an error for the application).

It is also dangerous, as if I handle `errback`, it's still logged as an error, even if I'm "handling" errors on the spider side.
So, maybe a better solution would be to log it as a warning and have a setting to raise a `RetryGaveUpException` that could be handled on a spider basis.
And the user can handle it in `errback`, even if no new setting is added.

Finally, my case is that I'm dealing with a broad crawler and it keeps showing errors when retry gives up, even though I don't have the power to control if the website is down or not.
So, in a per website basis, the error can be helpful, but once we don't have explicit control over the websites we're visiting, various things can happen and they may not be errors.

So, a candidate solution, that should be a mix of old/new behavior is, logging a warning or similar and rising an exception.
If the exception is not handled, it will log the error, but gives the user a simple approach to handle it properly.
Also, if you're already handling exceptions, then it'll keep the old behavior (just logging that it gave up).

### Versions

It was introduced recently
* https://github.com/scrapy/scrapy/issues/3171
* "Possible fix": https://github.com/scrapy/scrapy/issues/4216

### LINKED ISSUE #5297: get_retry_request: allow changing the log level of the gave up message
When retry requests are exceeded, get_retry_request logs an error message. I would like to be able to change the log level of that message, specifically into a warning.

further-reading:
Fixes #5297, fixes #4622

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -43,7 +43,7 @@ def get_retry_request(
     max_retry_times: Optional[int] = None,
     priority_adjust: Optional[int] = None,
     logger: Logger = retry_logger,
-    logger_give_up_level: int = 40,
+    give_up_log_level: int = 40,
     stats_base_key: str = 'retry',
 ):
     """
@@ -83,7 +83,7 @@ def get_retry_request(
 
     *logger* is the logging.Logger object to be used when logging messages
 
-    *logger_give_up_level* is the level that should be used for give up logs.
+    *give_up_log_level* is the level that should be used for give up logs.
     40 is the default level for ERROR logs.
 
     *stats_base_key* is a string to be used as the base key for the
@@ -120,7 +120,7 @@ def get_retry_request(
     else:
         stats.inc_value(f'{stats_base_key}/max_reached')
         logger.log(
-            logger_give_up_level,
+            give_up_log_level,
             "Gave up retrying %(request)s (failed %(retry_times)d times): "
             "%(reason)s",
             {'request': request, 'retry_times': retry_times, 'reason': reason},
@@ -144,7 +144,7 @@ class RetryMiddleware:
         self.max_retry_times = settings.getint('RETRY_TIMES')
         self.retry_http_codes = set(int(x) for x in settings.getlist('RETRY_HTTP_CODES'))
         self.priority_adjust = settings.getint('RETRY_PRIORITY_ADJUST')
-        self.logger_give_up_level = settings.getint('RETRY_LOG_GIVE_UP_LEVEL')
+        self.give_up_log_level = settings.getint('RETRY_GIVE_UP_LOG_LEVEL')
 
     @classmethod
     def from_crawler(cls, crawler):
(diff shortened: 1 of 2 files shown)


wRAR:
How would one use this? By subclassing the middleware?

further-reading:
> How would one use this? By subclassing the middleware?

The get_retry_request function is was added last year so one could trigger retries in a callback. https://docs.scrapy.org/en/latest/news.html#scrapy-2-5-0-2021-04-06

The purpose of this issue is to add the ability to set log level when using it in a callback. There's a seperate issue #4622 for extending the middleware to then use this feature of get_retry_request, which I'm also going to work on.

further-reading:
I've updated the middleware so this will fix #4622 as well.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -82,6 +83,9 @@ def parse(self, response):
 
     *logger* is the logging.Logger object to be used when logging messages
 
+    *logger_give_up_level* is the level that should be used for give up logs.
```
Gallaecio, line 86: What about `give_up_log_level`?

scrapy/downloadermiddlewares/retry.py
```diff
@@ -43,6 +43,7 @@ def get_retry_request(
     max_retry_times: Optional[int] = None,
     priority_adjust: Optional[int] = None,
     logger: Logger = retry_logger,
+    logger_give_up_level: int = 40,
```
Gallaecio, line 46: What about `give_up_log_level`?

Also, maybe we can make this `Union[int,str]`, to support the same API as the standard library, and use `"ERROR"` as default value, to make things more readable for people not aware of the value mappings.

Gallaecio (commented):
The code looks good to me, beyond my minor feedback.

Next we would need tests to cover the default function value, the ability to override through the function, and same for the setting. And a short documentation entry for the new setting.

scrapy/downloadermiddlewares/retry.py
further-reading: Regarding the feedback about "supporting the same API as the standard library" - the standard library does not support string arguments in the generic log function.

image

The library is written assuming that if you want to use labels as log level you should use a named method. The only way I've found to use a string as an argument is to have a mapping between the labels and their respective methods, but if a user is using a custom logger with different labels then I am not sure if we can create a mapping for them. Do you reckon it is okay to assume that everyone uses DEBUG, INFO, WARN, CRITICAL and ERROR; then leave the integer option for those who have added additional labels (and thus should be aware of their own mapping)?

Gallaecio: Oh, I did not know that. We could use [getLevelName](https://docs.python.org/3/library/logging.html#logging.getLevelName) to support custom log level strings as well. But I am also OK with forcing an int instead, given that is what the logging API supports. Whatever you prefer.

Gallaecio (approved):

wRAR:
@Gallaecio I see you previously wanted tests and docs but you've approved this now, so what should I do? :)

Gallaecio (changes requested):
I forgot about that :facepalm:

PR #5626, opened 2022-09-14
Relax Proxy-Authorization restrictions

Gallaecio:
This makes it so that the `Proxy-Authentication` header starts working again to set proxy credentials, as long as the Request metadata proxy URL has no other credentials, and for as long as that proxy URL remains the same.

After this change, many of the scenarios that 2.6.2 broke should start working again, but we will still be protecting users against the proxy-rotation scenario that made the original security change necessary.

To do:
- [x] Manually verify that old versions of scrapy-zyte-smartproxy work again.
- [x] Check if docs need changes.


wRAR (approved):

wRAR:
Should this go into 2.7?

kmike:
Thanks @Gallaecio! 

What do you think about making a 2.7.1 release, and not waiting for 2.8 to release this fix? It seems there are some other small fixes in master. The current set of merged PRs, + maybe https://github.com/scrapy/scrapy/pull/5689?

Gallaecio:
Sounds good to me!

PR #5632, opened 2022-09-25
Match pyOpenSSL and service_identity to Twisted

gliptak:
Resolves #5621

Signed-off-by: Gábor Lipták <gliptak@gmail.com>

Correcting https://github.com/scrapy/scrapy/actions/runs/3052958225/jobs/4922962575

```
    from twisted.internet import error, tcp, udp
.tox/py/lib/python3.8/site-packages/twisted/internet/tcp.py:38: in <module>
    from twisted.internet._newtls import (
.tox/py/lib/python3.8/site-packages/twisted/internet/_newtls.py:18: in <module>
    from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
.tox/py/lib/python3.8/site-packages/twisted/protocols/tls.py:45: in <module>
    from twisted.internet._sslverify import _setAcceptableProtocols
.tox/py/lib/python3.8/site-packages/twisted/internet/_sslverify.py:1828: in <module>
    defaultCiphers = OpenSSLAcceptableCiphers.fromOpenSSLCipherString(
.tox/py/lib/python3.8/site-packages/twisted/internet/_sslverify.py:1807: in fromOpenSSLCipherString
    SSL.TLS_METHOD,
E   AttributeError: module 'OpenSSL.SSL' has no attribute 'TLS_METHOD'
```

Issue #5621, opened 2022-09-08
Support for Twisted 22.8.0


wRAR:
Twisted 22.8.0 was released recently, and it says:

> Twisted now works with Cryptography versions 37 and above, and as a result, its minimum TLS protocol version has been upgraded to TLSv1.2.

Consequently, tests on some envs, including 3.8, now fail because they install older cryptography.

wRAR:
Also, I haven't yet checked but it may mean that Scrapy with this Twisted version will no longer work with some servers (it also needs to be checked if it's a hard restriction or the TLS version can be lowered  with settings)

wRAR:
We've found an interesting data point regarding the "support older TLS" part: https://chromestatus.com/feature/5759116003770368 says older TLS is no longer supported in Chromium and I can confirm it using https://badssl.com/ (https://tls-v1-1.badssl.com:1011/). While we usually strive to support as many real-world servers as possible, it should be a lower priority to support servers that are already dropped from a major browser.

Gallaecio:
Safe for Firefox. “This website might not support the TLS 1.2 protocol, which is the minimum version supported by Firefox.” Let’s burn TLS 1.1 and earlier with fire :fire: .

wRAR:
What is left to do here?

Gallaecio:
Nothing else I guess.

PR #5634, opened 2022-09-25
Cleanup METHOD_SSLv3

### LINKED ISSUE #5635: SSLv3_METHOD where did you come from???????????????
[WHAT THE F](https://github.com/scrapy/scrapy/blob/1.5.0/scrapy/core/downloader/tls.py#L17)

does not exist in pyopenssl

no help anywhere

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

docs generation is not running for PRs?

Fixes #5632, fixes #5635.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -564,7 +564,6 @@ This setting must be one of these string values:
   set this if you want the behavior of Scrapy<1.1
 - ``'TLSv1.1'``: forces TLS version 1.1
 - ``'TLSv1.2'``: forces TLS version 1.2
-- ``'SSLv3'``: forces SSL version 3 (**not recommended**)
 
 
 .. setting:: DOWNLOADER_CLIENT_TLS_VERBOSE_LOGGING
--- a/scrapy/core/downloader/contextfactory.py
+++ b/scrapy/core/downloader/contextfactory.py
@@ -21,7 +21,7 @@ class ScrapyClientContextFactory(BrowserLikePolicyForHTTPS):
     which allows TLS protocol negotiation
 
     'A TLS/SSL connection established with [this method] may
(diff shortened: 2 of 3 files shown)


Gallaecio:
We need to update https://docs.scrapy.org/en/latest/topics/settings.html#downloader-client-tls-method as well.

gliptak:
@Gallaecio updated

Gallaecio (approved):

kmike:
Hey! In install_requires we list pyOpenSSL 21.0 as a minimum requirement; should we support SSLv3 if older pyOpenSSL is installed? I.e. do getattr instead of removing the support completely?

@Gallaecio @wRAR what do you think about backporting the fix (e.g. for 2.6.3)?

Gallaecio:
Since [web browsers only support TLS 1.2+](https://github.com/scrapy/scrapy/issues/5621#issuecomment-1248231516), and [SSLv3 is worse than TLS 1.0](https://docs.twistedmatrix.com/en/twisted-20.3.0/core/howto/ssl.html#tls-protocol-options), I don’t think it is a big deal. In fact, I would be OK with dropping TLS 1.0 and 1.1 as well from there, just as the latest Twisted does, based on the argument that we aim to replicate browser behavior.

Not a strong opinion, though. @wRAR?

kmike:
Ok, let's kill SSLv3 :) Thanks for the help @gliptak!

xhMiao:
how to kill SSLv3？

kmike:
It's killed in master @xhMiao; we're also going to do 2.6.3 release (https://github.com/scrapy/scrapy/pull/5637), but it's not released yet. For now you can pin pyOpenSSL version, as described in https://github.com/scrapy/scrapy/issues/5635#issuecomment-1257415952.

PR #5639, opened 2022-09-26
fix: `batch_path_differ` test

### LINKED ISSUE #5633: Address BatchDeliveriesTest.test_batch_path_differ failures on Windows
After https://github.com/scrapy/scrapy/pull/5632, the only remaining test failure seems to be that of `BatchDeliveriesTest.test_batch_path_differ`, which seems to be failing in Windows for Python 3.7 (and pinned dependencies), Python 3.9+, but not Python 3.8, for some reason.

It is also weird that the error seems to suggest an extra item is yielded (4 != 3) in those scenarios. I cannot imagine what could cause an extra item yield in a specific platform.

Laerte:
Workaround for #5633

~~While debugging in Windows to identify the root cause i noticed that the test don't fail. So it seems that in Windows the test run too fast and this result in a invalid output, so in order to fix the test i add a delay only for Windows and this make the test suceed, since now the output is correct.~~

Check https://github.com/scrapy/scrapy/pull/5639#issuecomment-1259300827 for details.

### DIFF
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -2285,6 +2285,7 @@ class BatchDeliveriesTest(FeedExportTestBase):
             for expected_batch, got_batch in zip(expected, data[fmt]):
                 self.assertEqual(expected_batch, got_batch)
 
+    @pytest.mark.skipif(sys.platform == 'win32', reason='Odd behaviour on file creation/output')
     @defer.inlineCallbacks
     def test_batch_path_differ(self):
         """
@@ -2305,7 +2306,7 @@ class BatchDeliveriesTest(FeedExportTestBase):
             'FEED_EXPORT_BATCH_ITEM_COUNT': 1,
         }
         data = yield self.exported_data(items, settings)
-        self.assertEqual(len(items) + 1, len(data['json']))
+        self.assertEqual(len(items), len([_ for _ in data['json'] if _]))
 
     @defer.inlineCallbacks
     def test_stats_batch_file_success(self):


gliptak:
I also noticed this error being intermittent

Gallaecio:
I am OK with merging this for now as a workaround, but I would not close https://github.com/scrapy/scrapy/issues/5633 with it. I think we still need to find out why this is happening, where the extra item comes from without the delay.

Laerte:
@Gallaecio Is not a extra item that is yield on Windows.

Here is the output of `data['json']` on Linux:
```
[b'[\n{"foo": "bar1", "egg": "spam1"}\n]', b'[\n{"foo": "bar2", "egg": "spam2", "baz": "quux2"}\n]', b'[\n{"foo": "bar3", "baz": "quux3"}\n]', b'']
```
4 items, 1 is empty that the reason that test have the: 
https://github.com/scrapy/scrapy/blob/07b079defde98faa09db244b58dcb3904f2887f8/tests/test_feedexport.py#L2308
Since we only have 3 items declared:
https://github.com/scrapy/scrapy/blob/07b079defde98faa09db244b58dcb3904f2887f8/tests/test_feedexport.py#L2294-L2298

On Windows this is the output we have without the delay (That leads the test to fail):
```
[b'[\n{"foo": "bar1", "egg": "spam1"}\n]', b'[\n{"foo": "bar2", "egg": "spam2", "baz": "quux2"}\n][\n{"foo": "bar3", "baz": "quux3"}\n]']
```

As you can see the last item is kind of "appended" with the penultimate (for me this is a invalid output). Since i don't know the inner workings of Scrapy i don't know where this is happening (and only on windows). I suspect that is on [exports.py](https://github.com/scrapy/scrapy/blob/master/scrapy/exporters.py#L114)

Laerte:
@Gallaecio I refactor the test, now should work cross all environments.

kmike:
Do you know why are we getting an empty item on Linux? It's mystery to me :)
If there is a valid reason for that, probably it'd be better to ignore the empty result, instead of allowing both 3 and 4.
If there is no valid reason, we should fix it (and probably have an xfail test on Linux in the meantime).

Laerte:
@kmike The test generate 4 files, one of them is empty, this result in an empty item in list. We could add a validation but this will result in others tests to fail, for example `test_export_no_items_not_store_empty`

Here:
https://github.com/scrapy/scrapy/blob/07b079defde98faa09db244b58dcb3904f2887f8/tests/test_feedexport.py#L2033-L2038

I pushed other change basically filtering this invalid item. 😄

Laerte:
@kmike @Gallaecio  I add sleep again because even we filtering out the empty item we still have diff in output on windows [3 != 2](https://github.com/scrapy/scrapy/actions/runs/3136592335/jobs/5093700898), because in Windows instead of creating 3 files it created 2 and output the last 2 items in the last file with an invalid syntax:

image

kmike:
Nice find @Laerte! I think it could make sense to make this test XFAIL on WIndows, and open a new issue for the problem you found.

kmike:
Congrats for making the test suite green @Laerte! :)

What was the issue with xfail, how were you using it?
I think it can be good to add a link to GH issue to the reason.

Laerte:
0E867197-B01E-42FB-BDBB-37E8A3254B85

This is how i tried:
https://github.com/scrapy/scrapy/commit/23ba1ca5ebe75426d5b9459b93158a5631f72a23

kmike:
Aha, so the test is not always failing, and so when `strict` is used, you get an XPASS error. Would it make sense to remove strict=true?

Laerte:
Indeed i tried without  `strict=True’ first, odd is that in my branch that only run this test it suceed: https://github.com/Laerte/scrapy/commit/5405ce5e339f066b530d7dba12e57e973acc28b0

gliptak:
any additional updates needed before this PR ready to be merged?

kmike (approved):
Looks good, thanks @Laerte!

kmike:
@gliptak the PR looks good to me. Next step is to get a second approval from one of the Scrapy maintainers.

wRAR (approved):

PR #5645, opened 2022-10-01
Correct distutils deprecation warning

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```

### DIFF
--- a/scrapy/utils/display.py
+++ b/scrapy/utils/display.py
@@ -5,7 +5,7 @@ pprint and pformat wrappers with colorization support
 import ctypes
 import platform
 import sys
-from distutils.version import LooseVersion as parse_version
+from packaging.version import Version as parse_version
 from pprint import pformat as pformat_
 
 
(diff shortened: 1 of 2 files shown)


scrapy/utils/display.py
```diff
@@ -5,7 +5,7 @@
 import ctypes
 import platform
 import sys
-from distutils.version import LooseVersion as parse_version
+from packaging.version import Version as parse_version
```
kmike, line 8: I think it'd be nice to make `packaging` an explicit dependency in setup.py (add it to install_requires), even if it's usually available. distutils is an stdlib module.

gliptak: Updated

wRAR (approved):

kmike:
Thanks @gliptak!

Issue #5655, opened 2022-10-02
Add twine check to CI


kmike:
It'd be nice to do https://twine.readthedocs.io/en/stable/#twine-check on CI, to ensure our changes don't break pypi rendering.

armujahid:
How to create dist for twine check? I have tried `python setup.py build` and `python setup.py sdist`. 2nd one is creating dist with `Scrapy-2.6.2.tar.gz`

Running `Checking dist: ERROR    InvalidDistribution: Unknown distribution format: 'dist' ` throws this error 
```
Checking dist: ERROR    InvalidDistribution: Unknown distribution format: 'dist'                                                                                                                      
```
Looks like I am doing something wrong
Related: https://github.com/scrapy/scrapy/pull/5648

armujahid:
I think this is the correct way to run twine check. Correct me If I am wrong.
```
rm -rf dist
python setup.py sdist bdist_wheel
twine check dist/*
```

PR #5656, opened 2022-10-02
CI: add Twine check in check workflow

armujahid:
Closes: #5655

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -39,3 +39,8 @@ jobs:
       run: |
         pip install -U tox
         tox
+    - name: Twine check
+      run: |
+        pip install twine
+        python setup.py sdist
+        twine check dist/*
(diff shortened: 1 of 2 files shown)


armujahid:
Working action can be checked at https://github.com/armujahid/scrapy/actions/runs/3168893838/jobs/5160421545

.github/workflows/checks.yml
```diff
@@ -39,3 +39,8 @@ jobs:
       run: |
         pip install -U tox
         tox
+    - name: Twine check
+      run: |
+        pip install twine
+        python setup.py sdist
+        twine check dist/*
```
kmike, line 46: Could you please move this check to tox.ini, similar to other checks?

kmike: Currently this check would be run 5 times on CI, for flake8, typing, docs, etc. tox environments. We should make it a separate tox env, and run this env on CI, similar to how others are executed.

armujahid: Done

kmike (approved):
👍 Thanks @armujahid!

armujahid:
Windows/tests CI errors are unrelated

wRAR (approved):

PR #5657, opened 2022-10-02
CrawlSpider: add support for async callbacks

elacuesta:
I stumbled upon this by seeing reports of people getting errors when trying to use the CrawlSpider in combination with scrapy-playwright.

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -6,7 +6,7 @@ See documentation in docs/topics/spiders.rst
 """
 
 import copy
-from typing import Sequence
+from typing import Awaitable, Sequence
 
 from scrapy.http import Request, HtmlResponse
 from scrapy.linkextractors import LinkExtractor
@@ -109,9 +109,11 @@ class CrawlSpider(Spider):
         rule = self._rules[failure.request.meta['rule']]
         return self._handle_failure(failure, rule.errback)
 
-    def _parse_response(self, response, callback, cb_kwargs, follow=True):
+    async def _parse_response(self, response, callback, cb_kwargs, follow=True):
         if callback:
             cb_res = callback(response, **cb_kwargs) or ()
+            if isinstance(cb_res, Awaitable):
+                cb_res = await cb_res
             cb_res = self.process_results(response, cb_res)
             for request_or_item in iterate_spider_output(cb_res):
                 yield request_or_item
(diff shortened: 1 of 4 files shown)


tests/spiders.py
```diff
@@ -369,6 +369,18 @@ def parse(self, response, foo=None):
...
+class CrawlSpiderWithAsyncCallback(CrawlSpiderWithParseMethod):
+    """A CrawlSpider with an async def callback"""
+    name = 'crawl_spider_with_async_callback'
+    rules = (
+        Rule(LinkExtractor(), callback='parse_async', follow=True),
+    )
+
+    async def parse_async(self, response, foo=None):
+        self.logger.info('[parse_async] status %i (foo: %s)', response.status, foo)
+        return Request(self.mockserver.url("/status?n=202"), self.parse_async, cb_kwargs={"foo": "bar"})
```
kmike, line 381: Could you please also add a test for async generator callback?

elacuesta: Great point, I missed that. Updated, thanks for the suggestion!

elacuesta:
CI errors on Windows are unrelated, see #5633

wRAR (approved):

wRAR:
Nice!

scrapy/spiders/crawl.py
```diff
@@ -109,9 +110,13 @@ def _errback(self, failure):
         rule = self._rules[failure.request.meta['rule']]
         return self._handle_failure(failure, rule.errback)
 
-    def _parse_response(self, response, callback, cb_kwargs, follow=True):
+    async def _parse_response(self, response, callback, cb_kwargs, follow=True):
         if callback:
             cb_res = callback(response, **cb_kwargs) or ()
+            if isinstance(cb_res, AsyncIterable):
+                cb_res = await collect_asyncgen(cb_res)
```
kmike, line 117: Should we document that for CrawlSpider behavior of async callbacks is different from regular spiders? 

If I'm not mistaken, here the behavior is similar to Scrapy < 2.7, where the complete output of a callback is awaited before the processing started, while after https://github.com/scrapy/scrapy/pull/4978/ the output is processed as soon as it's sent.

elacuesta: Makes sense, thanks for pointing it out. Please note that this is coherent with `CrawlSpider.process_results`, which is documented to receive a list (although said docs only mention it in the context of the [XMLFeedSpider](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.XMLFeedSpider.process_results))

PR #5659, opened 2022-10-02
docs: update readme to use links

dschaller:
minor modifications to the readme to use relative links for the docs dir and code of conduct for easier navigation

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,8 @@ https://docs.scrapy.org/en/latest/intro/install.html for more details.
 Documentation
 =============
 
-Documentation is available online at https://docs.scrapy.org/ and in the `docs <https://github.com/scrapy/scrapy/blob/master/docs>`_ directory.
+Documentation is available online at https://docs.scrapy.org/ and in the ``docs``
+directory.
 
 Releases
 ========
--- a/README.rst
+++ b/README.rst
@@ -74,8 +74,7 @@ https://docs.scrapy.org/en/latest/intro/install.html for more details.
 Documentation
 =============
 
-Documentation is available online at https://docs.scrapy.org/ and in the ``docs``
-directory.
+Documentation is available online at https://docs.scrapy.org/ and in the `docs <./docs>`_ directory.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -95,8 +94,7 @@ See https://docs.scrapy.org/en/master/contributing.html for details.
 Code of Conduct
 ---------------
 
-Please note that this project is released with a Contributor Code of Conduct
-(see https://github.com/scrapy/scrapy/blob/master/CODE_OF_CONDUCT.md).
+Please note that this project is released with a Contributor `Code of Conduct <./CODE_OF_CONDUCT.md>`_.
```
kmike, line 97: We're using README as long_description for pypi; I'm not sure the relative link would work there.

dschaller:
@kmike updated to use full URLs of the links

README.rst
```diff
@@ -74,8 +74,7 @@ https://docs.scrapy.org/en/latest/intro/install.html for more details.
 Documentation
 =============
 
-Documentation is available online at https://docs.scrapy.org/ and in the ``docs``
-directory.
+Documentation is available online at https://docs.scrapy.org/ and in the `docs <https://github.com/scrapy/scrapy/blob/master/docs>`_ directory.
```
kmike, line 77: For me this URL redirects to https://github.com/scrapy/scrapy/tree/master/docs - would it make sense to use the final url, and avoid the redirect?

That said, I'm not sure we should be pointing users to `docs` folder; it doesn't render properly on github, because github doesn't have full sphinx support. What do you think about remoing this note instead, and pointing only to https://docs.scrapy.org/? cc @Gallaecio.

dschaller: I can revert this line and open a new PR where we can continue the discussion.

dschaller: Opene https://github.com/scrapy/scrapy/pull/5663 for further discussion

kmike (approved):
👍

Gallaecio (approved):

PR #5660, opened 2022-10-02
ci: add stalebot cron

dschaller:
At the moment there are 292 pull requests open dating back to 2015. These are likely stale and either would require large refactors to bring up to date in order to be compatible with the most recent code or closed all together. This PR adds a github workflow to run stalebot to automatically mark PRs that have been open 7+ days as stale and subsequently close them 3 days later if they are not marked as in progress.

I'm open to feedback on configuring these values (as well as the labels that bypass the stale check) to whatever fit the project.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 15,22 * * 1-5'
+jobs:
+  check_stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v6
+      with:
+        days-before-stale: 7
+        days-before-issue-stale: -1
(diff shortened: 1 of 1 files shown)


kmike:
Could you please explain, what's the motivation for using stalebot? I never understood it :)

For me closing stale PRs and issues automatically looks like an anti-pattern. It feels hostile for the contributors: they put effort, maybe it's unfinished, but it could be useful - but the PR is closed. Yes, you can apply "in progress" label, but it's not a default. It becomes harder to find things in the issue tracker if you have auto-closing: currently we're ignoring everything which is closed (because it's closed manually for a reason), but we wouldn't be able to, if PRs start to close automatically. With our workflow it doesn't seem useful for maintainers, I'm not sure what problem stalebot solves.

PR #5663, opened 2022-10-06
docs: remove docs folder reference from README

dschaller:
continuation of discussion from https://github.com/scrapy/scrapy/pull/5659#discussion_r986469558

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,8 +74,7 @@ https://docs.scrapy.org/en/latest/intro/install.html for more details.
 Documentation
 =============
 
-Documentation is available online at https://docs.scrapy.org/ and in the ``docs``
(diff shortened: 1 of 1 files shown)


kmike (approved):
👍

codePerfectPlus (approved):

Gallaecio:
tl;dr: +0 from me

I do not see the advantage of removing the reference to the ``docs`` folder for people without internet access.

However, that should be a very rare case for someone using Scrapy, and rather than navigating to ``docs`` people should probably be building the docs (`tox -e docs`) and opening them (`.tox/docs/tmp/html/index.html`) in a web browser. So I cannot argue against the change either.

PR #5664, opened 2022-10-06
Python 3.11 support

wRAR:
It looks like we can finally run CI on 3.11 and it fails. So this is the first step for adding support for 3.11.


elacuesta (approved):

kmike:
Should we declare it in setup.py?

wRAR:
@kmike right, I've made #5670

PR #5668, opened 2022-10-07
Update tool versions

wRAR:


tox.ini
```diff
@@ -58,6 +59,7 @@ deps =
     # Twisted[http2] is required to import some files
     Twisted[http2]>=17.9.0
     pytest-flake8==1.1.1
+    # newer ones don't work: https://github.com/tholo/pytest-flake8/issues/87
```
kmike, line 62: I wonder if it'd be simple enough to use flake8 directly, without pytest-flake8 plugin.

wRAR: I'll try, https://github.com/pypa/pep517/pull/147/files suggests it should be simple.

kmike:
Thanks @wRAR!

.flake8
```diff
@@ -4,6 +4,8 @@ max-line-length = 119
 ignore = W503
 
 exclude =
+    docs/conf.py
```
Gallaecio, line 7: For future reference, I would not mind _not_ excluding this file, and instead clean it up. All that commented out code and comments are not needed when you have documentation.

PR #5670, opened 2022-10-08
Add a classifier for Python 3.11.

wRAR:


kmike:
I think that's fine to add it, even if Python 3.11 is not released yet.

PR #5674, opened 2022-10-10
Updated setup-python and checkout actions workflow

Nirzak:
Updated setup-python workflow to version 4 and action checkout workflow to version 3.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -30,10 +30,10 @@ jobs:
             TOXENV: twinecheck
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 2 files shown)


wRAR (approved):

Nirzak:
Looks liek getting issue from this command pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:--durations=10 docs scrapy tests} from tox.ini in case of Python 3.10

wRAR:
The actual error is 

E           twisted.trial.unittest.FailTest: 
E           Expected: (<class 'twisted.internet.error.TimeoutError'>,)
E           Got:
E           [Failure instance: Traceback (failure with no frames): <class 'twisted.web._newclient.ResponseNeverReceived'>: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', '', 'wrong version number')]>]
E           ]

Not sure if this is something new or just a temporary problem.

Nirzak:
hmm also this particular python file is causing HTTP timeout error tests/test_downloader_handlers_http2.py but the same file fetched properly in case of python 3.11 and other version.
May be the error is temporary.

Nirzak:
Is it possible to manually re run the workflow?

wRAR:
I've restarted the failed job.

wRAR:
So it looks like an intermittent problem/

Nirzak:
> So it looks like an intermittent problem/

Yes. Looks like a temporary network blip during that particular job.

kmike:
Thanks @Nirzak!

PR #5678, opened 2022-10-11
Update PyPy version in documentation

Laerte:
The first one i bumped to match the same version of pypy3.7, after this PR was merged:
- https://github.com/scrapy/scrapy/pull/5667

I was in doubt about what version to use in second statement but both pypy3.7 and pypy3.9 use the same version, maybe we could remove this paragraph altogether since now is redundant? Version 5.9.0 is from 2017-10-05.

### DIFF
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -10,7 +10,7 @@ Supported Python versions
 =========================
 
 Scrapy requires Python 3.7+, either the CPython implementation (default) or
-the PyPy 7.3.5+ implementation (see :ref:`python:implementations`).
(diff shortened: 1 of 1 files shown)


wRAR:
I don't really know what should we do with these versions, we don't explicitly set them in the CI, so the versions in the docs can easily become outdated again.

Laerte:
@wRAR @Gallaecio Yeah i was in doubt about this too, i push new commit to remove these mentions of specific PyPy versions.

wRAR (approved):

Gallaecio:
Thanks!

Issue #4916, opened 2020-12-05
Replace `os.path` with `pathlib`?


eumiro:
Since scrapy is a Python 3.6+ project, it could use the `pathlib.Path` objects and methods to manipulate directories and files instead of the string manipulators in `os.path.*` and gain a lot of code readability. I have converted most of the code of scrapy to `pathlib.Path`, but a single PR would be too complex to review. Are you interested in this change? Would you review my PRs step by step? Any ideas?

nyov:
From what it looks like to me, much of scrapy's `pathlib` usage would just be turning strings into `Path` objects, only to `join()` them and turn them back into strings - which wouldn't cater to the benefits of Path objects at all.
But I could be wrong. Since you say you already converted code, you probably know better than me if that perception is false.

But to answer your question -  I guess I wouldn't care, as long as:
- it's not degrading codebase readability -- nice if you say it would actually enhance it by some margin
- it's not using `Path` objects in a way or place where it would impact performance or memory footprint of a running crawler negatively, compared to existing code
- it's not changing functionality; such as creating absolute paths (`Path.resolve()`), when the code didn't use or require absolute paths before

Further, I don't think a single PR should be too complex for review. It would be nice instead to see all the repeating patterns together, to find refactoring opportunities. To me, that looks like the single big benefit of doing this change: you might find repeating patterns like `makedir_if_not_exists(mydir)` to replace `os.path.exists`/`os.makedirs` combos, etc.
But of course you can split the changes over several commits to the PR branch (or "patch-queue", if you like), to make it more digestible.

Some people depending on github web reviews might be unable to review more complex patchsets, I'm aware; but we've had a few big ones for PRs in the past, and I believe scrapy's maintainers do have the skill to locally check out patches and use QoL tools like `hub`/`gh`/... for code reviews. So this shouldn't pose a limitation for you, thankfully.
If you look at PR's like #1149 or #1586, you'll see how they've grown over time (and I'm just picking some old ones so they don't get confused by the unrelated mention).

Gallaecio:
https://github.com/scrapy/scrapy/pull/5682

Issue #4497, opened 2020-04-15
Fix the “The Windows bytes API has been deprecated” warning


Gallaecio:
Seen in [Windows test runs](https://github.com/scrapy/scrapy/pull/4458).

AKSHAYSHARMAJS:
I want to work on this issue. I tried running test using tox locally on windows which doesn't show this warning but azure pipelines do capture this warning.

Gallaecio:
It looks related to `os` functions, which may return bytes for things like `getcwd`. But I cannot tell more without looking into the issue in detail myself. I do find strange that you do not get those warnings on Windows locally, though.

AKSHAYSHARMAJS:
I also find that strange. I have shared my log file of `tox` command maybe that will help.
[output.txt](https://github.com/scrapy/scrapy/files/5010152/output.txt)

AKSHAYSHARMAJS:
Please check the azure pipelines builds in #4646 don't show these warnings.

Gallaecio:
I can see them in the logs of the [latest builds](https://scrapinghub.visualstudio.com/scrapy/_build?definitionId=1&_a=summary&repositoryFilter=1&branchFilter=7%2C7%2C7%2C7%2C7%2C7).

Gallaecio:
https://github.com/scrapy/scrapy/pull/5682

PR #5682, opened 2022-10-17
use pathlib

mdonoughe:
This is a pretty tedious change. To address #4916 and hopefully #4497, I went through and used pathlib whenever possible for interacting with the filesystem.

Functions around new pathlib usage have had type annotations added. This helped somewhat with getting the correctness right. I avoided changing public interfaces to accept or expose Paths if they didn't already.

In one of the test files, Path objects were previously tested using one test and all the other tests used strings. I reversed this so only one test is using strings because I changed the shared test code to use Paths.

The tests pass, at least on my Windows machine.

This PR contains no intentional changes to functionality.


wRAR:
The tests are failing with "TypeError: 'ABCMeta' object is not subscriptable". There are also "unsupported-binary-operation" pylint errors and multiple different errors from mypy, including "X | Y syntax for unions requires Python 3.10".

mdonoughe:
"Not subscriptable" is also a Python version thing. I thought I had used some new features but the tox file said 3.8 and it passed with tox so I thought it was good. I'll figure it out tomorrow.

wRAR:
Indeed, tests on other versions fail for different reasons, e.g. "TypeError: unsupported operand type(s) for |: 'type' and 'types.GenericAlias'" on 3.9 and long/short file name comparison on Windows 3.10.

mdonoughe:
The path comparison issue seems like something weird with the agent environment, but it should work now. I was not getting the same problem even though my working directory contains a component longer than 8 characters. The type hints should all be compatible with Python 3.7.

When running the tests on 3.7, I got a weird error about mitmproxy not supporting the required version of SSL, but I don't see how I would of broken that so I'm assuming it's a problem with my environment.

wRAR:
There are still some typing issues and `tests/test_dependencies.py::ScrapyUtilsTest::test_pinned_twisted_version` failes on 3.7 (but not on later versions?)

wRAR:
And Windows tests still fail regarding short names.

wRAR:
I'll look into `test_pinned_twisted_version`

scrapy/commands/__init__.py
```diff
@@ -14,15 +15,15 @@
 class ScrapyCommand:
 
     requires_project = False
-    crawler_process = None
+    crawler_process: Any = None
```
kmike, line 18: Could it be `Optional[CrawlerProcess]` instead of Any?

scrapy/commands/genspider.py
```diff
@@ -143,17 +145,18 @@ def _spider_exists(self, name):
 
         # a file with the same name exists in the target directory
         spiders_module = import_module(self.settings['NEWSPIDER_MODULE'])
-        spiders_dir = dirname(spiders_module.__file__)
-        spiders_dir_abs = abspath(spiders_dir)
-        if exists(join(spiders_dir_abs, name + ".py")):
-            print(f"{join(spiders_dir_abs, (name + '.py'))} already exists")
+        spiders_dir = Path(cast(str, spiders_module.__file__)).parent
```
kmike, line 148: Why was mypy failing on it?

scrapy/commands/genspider.py
```diff
@@ -143,17 +145,18 @@ def _spider_exists(self, name):
...
 
         return False
 
     @property
-    def templates_dir(self):
-        return join(
-            self.settings['TEMPLATES_DIR'] or join(scrapy.__path__[0], 'templates'),
+    def templates_dir(self) -> str:
+        return str(Path(
+            self.settings['TEMPLATES_DIR'] or Path(scrapy.__path__[0], 'templates'),
```
kmike, line 160: TIL `Path.__init__` accepts multiple positional arguments, and joins them

scrapy/commands/runspider.py
```diff
@@ -1,22 +1,24 @@
...
 
 
-def _import_file(filepath):
-    abspath = os.path.abspath(filepath)
-    dirname, file = os.path.split(abspath)
-    fname, fext = os.path.splitext(file)
-    if fext not in ('.py', '.pyw'):
+def _import_file(filepath: Union[str, PathLike]) -> ModuleType:
+    abspath = Path(filepath).resolve()
+    dirname = str(abspath.parent)
```
kmike, line 15: very minor: probably it makes sense to move this line after the abspath.suffix check

scrapy/commands/startproject.py
```diff
@@ -90,36 +90,36 @@ def run(self, args, opts):
...
-        self._copytree(self.templates_dir, abspath(project_dir))
-        move(join(project_dir, 'module'), join(project_dir, project_name))
+        self._copytree(Path(self.templates_dir), project_dir.resolve())
+        move(project_dir / 'module', project_dir / project_name)
         for paths in TEMPLATES_TO_RENDER:
-            path = join(*paths)
-            tplfile = join(project_dir, string.Template(path).substitute(project_name=project_name))
-            render_templatefile(tplfile, project_name=project_name, ProjectName=string_camelcase(project_name))
+            tplfile = Path(project_dir, *(string.Template(s).substitute(project_name=project_name) for s in paths))
+            render_templatefile(str(tplfile), project_name=project_name, ProjectName=string_camelcase(project_name))
```
kmike, line 112: It seems we can avoid calling `str(tplfile)` (here and in other tests) if the first argument of render_teplatefile would be type annotated as `Union[str, PathLike]`; it seems it won't need code changes, just annotation changes.

scrapy/commands/__init__.py
kmike: (here and in the other commands as well)

scrapy/extensions/httpcache.py
```diff
@@ -303,41 +305,41 @@ def retrieve_response(self, spider, request):
...
             f.write(headers_dict_to_raw(request.headers))
-        with self._open(os.path.join(rpath, 'request_body'), 'wb') as f:
+        with self._open(rpath / 'request_body', 'wb') as f:
             f.write(request.body)
 
-    def _get_request_path(self, spider, request):
+    def _get_request_path(self, spider: Spider, request: Request) -> Path:
         key = self._fingerprinter.fingerprint(request).hex()
-        return os.path.join(self.cachedir, spider.name, key[0:2], key)
+        return Path(self.cachedir, spider.name, key[0:2], key)
```
kmike, line 335: That's a risky change; even if a method is declared as private, it's still overridden in the wild, e.g. https://github.com/scrapy-plugins/scrapy-splash/blob/f5273b3a0ef15a9f6809d305a1dbe8f3efd12c5f/scrapy_splash/cache.py#L17

scrapy/pipelines/files.py
```diff
@@ -39,41 +41,40 @@ class FileException(Exception):
...
 class FSFilesStore:
-    def __init__(self, basedir):
+    def __init__(self, basedir: str):
         if '://' in basedir:
             basedir = basedir.split('://', 1)[1]
         self.basedir = basedir
-        self._mkdir(self.basedir)
-        self.created_directories = defaultdict(set)
+        self._mkdir(Path(self.basedir))
+        self.created_directories: defaultdict[str, set[str]] = defaultdict(set)
```
kmike, line 49: If I'm not mistaken, this syntax is Python 3.9+

scrapy/spiders/__init__.py
```diff
@@ -3,21 +3,26 @@
...
+    from scrapy.crawler import Crawler
+
 
 class Spider(object_ref):
     """Base class for scrapy spiders. All spiders must inherit from this
     class.
     """
 
-    name: Optional[str] = None
+    name: str
```
kmike, line 25: This is a change in behavior. It looks fine, but I'm not sure I understand the consequences fully. Would it make sense to keep it as it was, or should we investigate further?

wRAR: This indeed looks wrong as base spider classes usually don't have the name attribute set.

scrapy/commands/genspider.py
mdonoughe: Supposedly `__file__` can be `None`.

scrapy/commands/__init__.py
mdonoughe: I originally tried it when mypy activated for this file and decided that `crawler_process` was of type `None`, but it exposes potential type errors in other parts of the code:
```
scrapy\commands\genspider.py:137: error: Item "None" of "Optional[CrawlerProcess]" has no attribute "spider_loader"  [union-attr]
```

tests/test_dependencies.py
```diff
@@ -29,11 +30,7 @@ def test_pinned_twisted_version(self):
         if not os.environ.get('_SCRAPY_PINNED', None):
             self.skipTest('Not in a pinned environment')
 
-        tox_config_file_path = os.path.join(
-            os.path.dirname(__file__),
-            '..',
-            'tox.ini',
-        )
+        tox_config_file_path = Path(__file__) / '..' / 'tox.ini'
```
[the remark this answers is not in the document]
mdonoughe: I think you are right.

scrapy/extensions/httpcache.py
mdonoughe: I guess the question is, is it private as in it's not supposed to be called by other types or is it private as in it's part of Scrapy and if you touch it you're accepting the risk of it unexpectedly changing? I can put a `str()` around it.

scrapy/pipelines/files.py
mdonoughe: You are right, but the tests somehow pass on 3.7? There are two instances of this in this file.

scrapy/spiders/__init__.py
mdonoughe: It's required not to be `None` by `httpcache.py`.
```
scrapy\extensions\httpcache.py:335: error: Argument 2 to "Path" has incompatible type "Optional[str]"; expected "Union[str, PathLike[str]]"  [arg-type]
```

scrapy/commands/runspider.py
mdonoughe: I moved it, but now I'm looking at `if dirname:` and thinking it can never fail. In both the old code and the new code, `abspath` is an absolute path to a file. Files must be contained within directories, so the absolute path will always have some parent path component. Even if you provide as input the absolute path to a path root, the parent of the root is the root itself.

The only way it could be false is if `abspath` were not an absolute path, in which case the old code would have returned an empty string and the new code would return `"."`.

```
>>> from pathlib import PurePosixPath, PureWindowsPath
>>> PurePosixPath('/').parent
PurePosixPath('/')
>>> PureWindowsPath('c:\\').parent
PureWindowsPath('c:/')
>>> import os
>>> os.path.split('/')
('/', '')
>>> os.path.split('c:\\') 
('c:\\', '')
>>> PureWindowsPath('a').parent            
PureWindowsPath('.')
>>> os.path.split('a')                     
('', 'a')
```

scrapy/spiders/__init__.py
mdonoughe: I'm not sure what to do about this sort of issue.

scrapy/pipelines/files.py
wRAR: This is not the first time I hear "this typing syntax shouldn't be supported on the oldest version but the tests pass for it". Not sure why is it so.

scrapy/spiders/__init__.py
wRAR: Well.
The base spiders I mentioned aren't supposed to be run so middlewares won't be executed :) Maybe it's enough to add a cast/assert to the middleware.
OTOH there is a WIP PR #4327 that would make the name optional even for spiders that can be run.

wRAR:
The tests passed! :)

scrapy/spiders/__init__.py
mdonoughe: Isn't it required that there is a name, even for the base `Spider`s? `InitSpider` does not have a name, but when you construct it you must provide a name as a parameter or else `Spider`'s `__init__` will raise an exception. It's these lines being removed in the other PR: https://github.com/scrapy/scrapy/pull/4327/files#diff-bfeaff1ed9c7b0a9a61d76bca0e46e4e34289df77cbc1a5143c31e4d1be6fcadL28-L29

scrapy/pipelines/files.py
mdonoughe: I thought maybe there could be missing tests, but there are tests for that code, and they pass on 3.7 even if I put the annotations back to how they were. It works, even when not running the tests.

It turns out, Python 3.7 doesn't care what's written here at all. It doesn't need to be a real type.

```
Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)] on win32
>>> a: aoeaoeaoe = None
NameError: name 'aoeaoeaoe' is not defined
>>> class A:     
...     def __init__(self):
...         self.a: aoeaoeaoe = None
...
>>> A()
<__main__.A object at 0x0000019C495C8F08>
>>>
```

scrapy/spiders/__init__.py
Gallaecio: The `Spider` class itself does not have a `name` defined, and that is not an error. And subclasses do not need a name either, as long as they are not instantiated (which is when `Spider.__init__` gets called, though a subclass could actually override its `__init__` to prevent that exception in the first place).

mdonoughe: If a subclass doesn't call the base `__init__` that seems like all promises made by the base class are off and functions provided by the base class may not work. Python doesn't enforce type hints at runtime so the code will work the same as before even if somebody skips over the code that asserts there is always a name.

mdonoughe: I'm thinking this is correct as `str`. `name` is intended not to be `None`. If the type checker is really smart, having it declared to be `str` could generate a warning if somebody does create a special subclass that doesn't call the base `__init__` to set it (I haven't checked this, but I doubt it). The type checker will definitely generate a warning if somebody writes code that explicitly sets it to `None`.

wRAR:
[There is actually `Path.absolute()` but it's only documented in 3.11](https://docs.python.org/3.11/library/pathlib.html#pathlib.Path.absolute) and it's also not equivalent to abspath: "`os.path.abspath()` normalizes the resulting path, which may change its meaning in the presence of symlinks, while `Path.absolute()` does not." ("`posixpath.abspath()` collapses "<path>/<symlink>/.." to "<path>", this is not correct")

So I guess we should keep using `resolve()`

mdonoughe:
Is this PR waiting on me to change anything? I'm sure the `str` type hint is correct because `None` is not a valid value for that attribute, but I could set it back to `Optional[str]` so it's documented that `None` is allowed but there is an exception raised if `None` is used.

wRAR:
I think it's not reviewed completely yet, so for now it's waiting for that.

mdonoughe:
Could I get a hacktoberfest-accepted?

tests/test_downloader_handlers.py
```diff
@@ -108,14 +108,13 @@ def test_lazy_handlers(self):
 class FileTestCase(unittest.TestCase):
 
     def setUp(self):
-        self.tmpname = self.mktemp()
-        with open(self.tmpname + '^', 'w') as f:
-            f.write('0123456789')
+        self.tmpname = Path(self.mktemp() + '^')
```
wRAR, line 111: Note to self: find (again) why is this ^ needed in this test suite and add a comment. Or remove it if it's not actually needed.

wRAR:
> I avoided changing public interfaces to accept or expose Paths if they didn't already.

Maybe we should create a feature request for this after this PR is merged.

tests/test_downloader_handlers.py
wRAR: #5736

wRAR (approved):

wRAR:
Thanks! I've finished reviewing this.

tests/test_utils_iterators.py
```diff
@@ -303,11 +301,6 @@ def test_xmliter_objtype_exception(self):
 
 
 class UtilsCsvTestCase(unittest.TestCase):
-    sample_feeds_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'sample_data', 'feeds')
-    sample_feed_path = os.path.join(sample_feeds_dir, 'feed-sample3.csv')
-    sample_feed2_path = os.path.join(sample_feeds_dir, 'feed-sample4.csv')
-    sample_feed3_path = os.path.join(sample_feeds_dir, 'feed-sample5.csv')
```
kmike, line 309: nice catch @mdonoughe!

tests/test_dependencies.py
```diff
@@ -29,11 +30,7 @@ def test_pinned_twisted_version(self):
         if not os.environ.get('_SCRAPY_PINNED', None):
             self.skipTest('Not in a pinned environment')
 
-        tox_config_file_path = os.path.join(
-            os.path.dirname(__file__),
-            '..',
-            'tox.ini',
-        )
+        tox_config_file_path = Path(__file__).parent / '..' / 'tox.ini'
```
kmike, line 33:
```suggestion
        tox_config_file_path = Path(__file__).parent.parent / 'tox.ini'
```

scrapy/utils/test.py
```diff
@@ -71,11 +72,11 @@ def get_crawler(spidercls=None, settings_dict=None, prevent_warnings=True):
     return runner.create_crawler(spidercls or Spider)
 
 
-def get_pythonpath():
+def get_pythonpath() -> str:
     """Return a PYTHONPATH suitable to use in processes so that they find this
     installation of Scrapy"""
     scrapy_path = import_module('scrapy').__path__[0]
-    return os.path.dirname(scrapy_path) + os.pathsep + os.environ.get('PYTHONPATH', '')
+    return str(Path(scrapy_path).parent) + os.pathsep + os.environ.get('PYTHONPATH', '')
```
kmike, line 79: Not related to this PR, but this code looks suspicious; I wonder why path is not just joined, and os.pathsep is used instead.

tests/test_downloader_handlers.py
```diff
@@ -1065,28 +1063,28 @@ def _test(r):
...
+            self.assertEqual(r.body, fname_bytes)
+            self.assertEqual(r.headers, {b'Local Filename': [fname_bytes],
                                          b'Size': [b'17']})
-            self.assertTrue(os.path.exists(local_fname))
-            with open(local_fname, "rb") as f:
-                self.assertEqual(f.read(), b"I have the power!")
-            os.remove(local_fname)
+            self.assertTrue(local_fname.exists())
+            self.assertEqual(local_fname.read_bytes(), b"I have the power!")
+            local_fname.unlink()
```
kmike, line 1081: not related to this PR: if a previous asserts fails, the file is not removed

scrapy/utils/test.py
wRAR: `os.pathsep` is a separator of paths (: or ;) as opposed to `os.sep`. It's a list of paths, not a single path.

tests/test_dependencies.py
mdonoughe: The old code was like this so I didn't change it. `'..'` and `.parent` are slightly different if there are symbolic links. It'd probably be okay to change it here.

kmike:
Thanks for the careful implementation @mdonoughe! 
There were so many chances to get it wrong - and you avoided all of them :)

PR #5683, opened 2022-10-19
Update installation process on Ubuntu Linux

nfsec:
- Note not to use too old versions of Ubuntu distribution.
- Added python3-venv package for creating virtual environments.
- Installation in venv in three steps with pip.


Gallaecio (commented):
Maybe we should reword the docs not to mention any specific Ubuntu version, and cover only the instructions for (upstream-)supported Ubuntu versions.

For virtual env instructions, rather than having the instructions on this Ubuntu-specific section, I would link to the section above that covers virtual environment installation.

nfsec:
> Maybe we should reword the docs not to mention any specific Ubuntu version, and cover only the instructions for (upstream-)supported Ubuntu versions.
> 
> For virtual env instructions, rather than having the instructions on this Ubuntu-specific section, I would link to the section above that covers virtual environment installation.

Good idea.

PR #5691, opened 2022-10-25
Replace indentation of source before parsing with ast.

### LINKED ISSUE #5323: Warning: Unable to determine whether or not callable is a generator with a return value [unable to parse decorated class methods]
### Description

If you have a decorated spider method, `scrapy.utils.misc` [throws a warning](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/misc.py#L240-L264) saying that it cannot determine if the callable is a generator with a return value.

`ast.parse()` fails [here](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/misc.py#L228-L230) when called with a decorated method.

### Steps to Reproduce

I just copied the logic from `misc.py` and used it to analyze a class with the same overall code structure:

```python
import re
import ast
import inspect

class Foo:
    @classmethod
    def func(self):
        """
        Description of func
        """  
        return

code = re.sub(r"^[\t ]+", "", inspect.getsource(Foo.func))
tree = ast.parse(code)
```

```
python3 test.py
> IndentationError: unexpected indent
```

The regex replacement isn't accounting for a possible decorator, so the code ends up looking like:

```
@classmethod
    def func(self):
        """
        Description of func
        """  
        return
```

**Expected behavior:** I'd like to be able to use decorated methods without dealing with noisy logs.

**Actual behavior:** My container logs are filled with tons of warning messages. The only workaround is to avoid the usage of decorators.

**Reproduces how often:** 100% of the time to my knowledge

### Versions

```
$ scrapy version --verbose

Scrapy       : 2.5.1
lxml         : 4.6.4.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 21.7.0
Python       : 3.8.2 (default, Dec 21 2020, 15:06:04) - [Clang 12.0.0 (clang-1200.0.32.29)]
pyOpenSSL    : 21.0.0 (OpenSSL 1.1.1l  24 Aug 2021)
cryptography : 35.0.0
Platform     : macOS-10.15.7-x86_64-i386-64bit
```

### Additional context

This is my first time filing a Scrapy issue. I'm happy to add more context if necessary, and apologies in advance if this has already been discussed elsewhere (fwiw I couldn't find anything).

Godson-Gnanaraj:
closes #5323

Replace starting indentation with pattern `(^[\t ]+)` and replace the matching indentation in other lines.

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -226,13 +226,13 @@ def is_generator_with_return_value(callable):
         return value is None or isinstance(value, ast.NameConstant) and value.value is None
 
     if inspect.isgeneratorfunction(callable):
-        pattern = r"(^[\t ]+)"
         src = inspect.getsource(callable)
-        match = re.match(pattern, src)  # Find indentation
-        code = re.sub(pattern, "", src)
+        pattern = re.compile(r"(^[\t ]+)")
+        code = pattern.sub("", src)
+
+        match = pattern.match(src)  # finds indentation
         if match:
-            # Remove indentation
-            code = re.sub(f"\n{match.group(0)}", "\n", code)
+            code = re.sub(f"\n{match.group(0)}", "\n", code)  # remove indentation
 
         tree = ast.parse(code)
         for node in walk_callable(tree):
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -226,7 +226,14 @@ def is_generator_with_return_value(callable):
         return value is None or isinstance(value, ast.NameConstant) and value.value is None
 
     if inspect.isgeneratorfunction(callable):
-        code = re.sub(r"^[\t ]+", "", inspect.getsource(callable))
+        pattern = r"(^[\t ]+)"
+        src = inspect.getsource(callable)
+        match = re.match(pattern, src)  # Find indentation
+        code = re.sub(pattern, "", src)
+        if match:
+            # Remove indentation
+            code = re.sub(f"\n{match.group(0)}", "\n", code)
+
         tree = ast.parse(code)
         for node in walk_callable(tree):
             if isinstance(node, ast.Return) and not returns_none(node):
--- a/tests/test_utils_misc/test_return_with_argument_inside_generator.py
+++ b/tests/test_utils_misc/test_return_with_argument_inside_generator.py
@@ -165,6 +165,89 @@ https://example.org
             warn_on_generator_with_return_value(None, l2)
             self.assertEqual(len(w), 0)
 
+    def test_generators_return_none_with_decorator(self):
+        def decorator(func):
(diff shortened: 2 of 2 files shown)


Gallaecio (commented):
Could you add a test for this?

scrapy/utils/misc.py
```diff
@@ -226,7 +226,14 @@ def returns_none(return_node):
         return value is None or isinstance(value, ast.NameConstant) and value.value is None
 
     if inspect.isgeneratorfunction(callable):
-        code = re.sub(r"^[\t ]+", "", inspect.getsource(callable))
+        pattern = r"(^[\t ]+)"
```
elacuesta, line 229: Can this pattern be compiled? It wasn't in the original version, but I guess it could mean a slight performance improvement.

scrapy/utils/misc.py
```diff
@@ -226,7 +226,14 @@ def returns_none(return_node):
...
+        src = inspect.getsource(callable)
+        match = re.match(pattern, src)  # Find indentation
+        code = re.sub(pattern, "", src)
+        if match:
+            # Remove indentation
+            code = re.sub(f"\n{match.group(0)}", "\n", code)
```
elacuesta, line 235: Nitpick 2:
```suggestion
            code = re.sub(f"\n{match.group(0)}", "\n", code)  # remove indentation
```
More concise IMHO, and still respects the max line length

scrapy/utils/misc.py
```diff
@@ -226,7 +226,14 @@ def returns_none(return_node):
         return value is None or isinstance(value, ast.NameConstant) and value.value is None
 
     if inspect.isgeneratorfunction(callable):
-        code = re.sub(r"^[\t ]+", "", inspect.getsource(callable))
+        pattern = r"(^[\t ]+)"
+        src = inspect.getsource(callable)
+        match = re.match(pattern, src)  # Find indentation
+        code = re.sub(pattern, "", src)
```
elacuesta, line 232: Nitpick 1:
```suggestion
        code = re.sub(pattern, "", src)
        match = re.match(pattern, src)  # find indentation
```

Just swapping the two lines. For me it looks a little bit more readable, in the sense that `match` is only used within the next `if` block. In fact, this would be even better with the walrus operator if we had Python 3.8 as minimum version.

elacuesta (commented):
Looking good, just a few suggestions. Thanks!

Gallaecio (approved):

scrapy/utils/misc.py
Godson-Gnanaraj: Fair point, Thank you! :bow:

wRAR (approved):

PR #5692, opened 2022-10-25
Handle deprecated warning with try except block

### LINKED ISSUE #5684: DeprecationWarning: ANTIALIAS is deprecated
> scrapy/pipelines/images.py:163: DeprecationWarning: ANTIALIAS is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.LANCZOS instead.
>    image.thumbnail(size, self._Image.ANTIALIAS)

This is deprecated since Pillow 9.1.0, released in April. We should check if `Resampling.LANCZOS` is already available in the earliest version we support (7.1.0 as far as I can see), and use it if it's available there. If it was added later, I think we need to decide how to proceed.

iamkaushal:
fixes #5684 
Using try except block to support backward compablility of order version of Pillow. It can we removed when the minimum requirement of Pillow is updated.

### DIFF
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -164,7 +164,7 @@ class ImagesPipeline(FilesPipeline):
                 # Image.Resampling.LANCZOS was added in Pillow 9.1.0
                 # remove this try except block,
                 # when updating the minimum requirements for Pillow.
-                resampling_filter =  self._Image.Resampling.LANCZOS
+                resampling_filter = self._Image.Resampling.LANCZOS
             except AttributeError:
                 resampling_filter = self._Image.ANTIALIAS
             image.thumbnail(size, resampling_filter)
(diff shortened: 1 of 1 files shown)


scrapy/pipelines/images.py
```diff
@@ -160,7 +160,14 @@ def convert_image(self, image, size=None):
 
         if size:
             image = image.copy()
-            image.thumbnail(size, self._Image.ANTIALIAS)
+            try:
+                # Image.Resampling.LANCZOS was added in Pillow 9.1.0
+                # remove this try except block,
+                # when updating the minimum requirements for Pillow.
+                resampling_filter =  self._Image.Resampling.LANCZOS
```
wRAR, line 167: extra space after =

wRAR, line 167: extra space after =

iamkaushal: fixed.

iamkaushal: fixed.

wRAR (approved):

Gallaecio:
Thanks!

PR #5693, opened 2022-10-26
Created README_hindi.rst

AkshitGulyan:
Added a Hindi Translation of README File !

For Hindi Reading People, it would be easy to read the README Document

### DIFF
--- a/README_hindi.rst
+++ b/README_hindi.rst
@@ -1,0 +1,113 @@
+.. image:: https://scrapy.org/img/scrapylogo.png
+   :target: https://scrapy.org/
+   
+======
+Scrapy
+======
+
+.. image:: https://img.shields.io/pypi/v/Scrapy.svg
+   :target: https://pypi.python.org/pypi/Scrapy
+   :alt: PyPI Version
+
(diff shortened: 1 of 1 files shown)


Gallaecio:
I don’t think raw translations of the README are scalable:

- We would have to keep many files on the root folder, which is messy.
- There would be no easy way for translators to keep up with changes in the main README.

I think https://github.com/scrapy/scrapy/pull/4399 is the way to go. Anything else would feel like a temporary patch to me.

AkshitGulyan:
> I don’t think raw translations of the README are scalable:
> 
> * We would have to keep many files on the root folder, which is messy.
> * There would be no easy way for translators to keep up with changes in the main README.
> 
> I think #4399 is the way to go. Anything else would feel like a temporary patch to me.

@Gallaecio Thanks for the response
I can understand .

And I saw [Hindi Page](https://hi.wikipedia.org/wiki/%E0%A4%B8%E0%A5%8D%E0%A4%95%E0%A5%8D%E0%A4%B0%E0%A5%87%E0%A4%AA%E0%A5%80) is already there on the Wikipedia.

PR #5695, opened 2022-10-27
ci: Update macos runner

jidicula:
The GitHub Actions macos-10.15 runner image is now deprecated, and GitHub Actions has begun to temporarily fail jobs referencing it during brownout periods. The image will be fully unsupported by 2022-12-01, which is just about a month away.

This change updates the macOS runner image to the latest generally-available version, to help reduce spurious CI failures during the brownout periods, and to stay abreast of the sunsetting of the macos-10.15 image.

See also: actions/runner-images#5583

### DIFF
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   tests:
-    runs-on: macos-10.15
+    runs-on: macos-11
     strategy:
       fail-fast: false
       matrix:


jidicula:
This is also a Hacktoberfest PR - if it passes muster, could you add the `hacktoberfest-accepted` label to this PR?

wRAR (approved):

kmike:
The fix makes sense, but it seems the workflow is not running on the CI. It was the case even before this PR.

wRAR:
Yes, it's disabled, see #5032

PR #5701, opened 2022-11-01
Release notes for 2.7.1.

wRAR:


docs/news.rst
```diff
@@ -3,6 +3,57 @@
...
+Scrapy 2.7.1 (Not relased yet)
+------------------------------
+
+New features
+~~~~~~~~~~~~
+
+-   Relaxed the restriction introduced in 2.6.2 so that the
+    ``Proxy-Authentication`` header can again be set explicitly, as long as the
+    proxy URL in the :reqmeta:`proxy` metadata has no other credentials, and
+    for as long as that proxy URL remains the same (:issue:`5626`).
```
kmike, line 17: What do you think about mentioning https://github.com/scrapy-plugins/scrapy-zyte-smartproxy here? Would it be helpful for our users?

wRAR: Oh yeah. Like "this restores compatibility with older versions of it"?

kmike: yep; it makes sense to call out specific version number as well

Gallaecio: > it makes sense to call out specific version number as well

I suspect this restores support for *all* previous versions of scrapy-zyte-smartproxy/scrapy-crawlera, though I did not confirm this.

docs/news.rst
```diff
@@ -3,6 +3,57 @@
...
+-   Using ``-O``/``--overwrite-output`` and ``-t``/``--output-format`` options
+    together now produces an error instead of ignoring the former option
+    (:issue:`5516`, :issue:`5605`).
+
+-   Replaced deprecated :mod:`asyncio` APIs that implicitly use the current
+    event loop with code that explicitly requests a loop from the event loop
+    policy (:issue:`5685`, :issue:`5689`).
+
+-   Fixed uses of deprecated Scrapy APIs in the Scrapy itself (:issue:`5588`,
+    :issue:`5589`).
```
Gallaecio, line 31:
```suggestion
-   Fixed uses of deprecated Scrapy APIs in Scrapy itself (:issue:`5588`,
    :issue:`5589`).
```

docs/news.rst
```diff
@@ -3,6 +3,57 @@
...
+
+Quality assurance
+~~~~~~~~~~~~~~~~~
+
+-   Re-enabled some erroneously disabled flake8 checks (:issue:`5688`).
+
+-   Ignored harmless deprecation warnings from :mod:`typing` in tests
+    (:issue:`5686`, :issue:`5697`).
+
+-   Modernized CI configuration (:issue:`5695`, :issue:`5696`).
```
Gallaecio, line 54:
```suggestion
-   Modernized our CI configuration (:issue:`5695`, :issue:`5696`).
```

Gallaecio (approved):

docs/news.rst
kmike: A good point about the lower bound; I was thinking only about mentioning a highest version number. It doesn't look like a blocker though if we don't provide the oldest version number here.

kmike (approved):
👍

docs/news.rst
Gallaecio: Oh, “scrapy-zyte-smartproxy 2.1.0 and older”, yes, that makes sense!

wRAR: Done. Should we also mention 3rd-party proxy extensions here (without specific names) or is this enough?

Gallaecio: I think this is enough.

PR #5704, opened 2022-11-02
Use Python's `yield from`

cclauss:
[`What’s New In Python 3.3`](https://docs.python.org/3/whatsnew/3.3.html)

### DIFF
--- a/tests/test_spidermiddleware.py
+++ b/tests/test_spidermiddleware.py
@@ -188,7 +188,8 @@ class ProcessSpiderExceptionAsyncIterableMiddleware:
         from twisted.internet import reactor
         reactor.callLater(0, d.callback, None)
         await maybe_deferred_to_future(d)
-        yield from ({'foo': 2}, {'foo': 3})
+        yield {'foo': 2}
+        yield {'foo': 3}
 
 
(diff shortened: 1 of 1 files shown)


tests/test_spidermiddleware.py
```diff
@@ -124,9 +124,7 @@ def _construct_mw_setting(*mw_classes, start_index: Optional[int] = None):
         return {i: c for c, i in enumerate(mw_classes, start=start_index)}
 
     def _scrape_func(self, *args, **kwargs):
-        yield {'foo': 1}
-        yield {'foo': 2}
-        yield {'foo': 3}
+        yield from ({'foo': 1}, {'foo': 2}, {'foo': 3})
```
Gallaecio, line 127: I am not sure this is better than before.

Gallaecio (commented):
The for loops do make sense to me, though. Thanks!

tests/test_spidermiddleware.py
JazzGlobal: I would say it is "better".  In the sense that moving these items from multi-line to one line does not make it any more difficult to understand while performing the same task using a new language feature.

Issue #5709, opened 2022-11-09
scrapy -h outputs a stray "commands"


wRAR:
When I run `scrapy -h` in a project on all Scrapy versions starting with 2.1, I get:

```
[...]
Available commands:
  bench         Run quick benchmark test
  check         Check spider contracts
  **commands**      
  crawl         Run a spider
  edit          Edit spider
  fetch         Fetch a URL using the Scrapy downloader
[...]
```

The stray line is not present on older versions. I haven't found its source but it should be easy I think. The list of commands is collected in `scrapy.cmdline._get_commands_dict()` and printed by `scrapy.cmdline._print_commands()`.

PR #5711, opened 2022-11-09
scrapy -h outputs a stray commands #5709

Jtwoolbright:
__init__.py was being passed as a module from scrapy/commands when walk_modules was called from function def _iter_command_classes in cmdline.py

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -79,7 +79,6 @@ def walk_modules(path):
 
     mods = []
     mod = import_module(path)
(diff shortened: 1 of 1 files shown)


wRAR:
Thanks for finding the issue! I think changing the behavior of `walk_modules` is not a correct fix here though, as I think it changes its behavior and the docstring describes the current one. Removing the incorrect module from the list in the command-related code is better, I think.

wRAR:
And it indeed breaks tests (can you please run tests locally before publishing next time?).

PR #5712, opened 2022-11-10
fix: stray `commands`

Laerte:
This behaviour was introduced on: #4552

We created a new class that inherit from `ScrapyCommand` but forget to check on `_iter_command_classes` function, this is the reason that `command` was appearing.

Fixes: #5711


wRAR (approved):

Gallaecio (approved):

Gallaecio:
Great!

PR #5713, opened 2022-11-12
Update tutorial.rst

harshhes:
1. 
```
>>> response.css('title::text').get()
'Quotes to Scrape'
```
2.
```
>>> response.css('title::text')[0].get()
'Quotes to Scrape'
```
Both 1st and 2nd line of code base returns the same response, The Alternative what i think should be:
```
>>> response.css('title::text').getall()[0]
'Quotes to Scrape'
```
Because as `getall()` returns a list and we can select the very first item from it. 
However 2. being the the alternative to 1. doesn't make any sense.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -274,7 +274,7 @@ When you know you just want the first result, as in this case, you can do:
 
 As an alternative, you could've written:
 
->>> response.css('title::text')[0].get()
+>>> response.css('title::text').getall()[0]
 'Quotes to Scrape'
 
 Accessing an index on a :class:`~scrapy.selector.SelectorList` instance will 


wRAR:
> Both 1st and 2nd line of code base returns the same response

Sorry?

> The Alternative what i think should be:

And why do you think it should?

> Because as getall() returns a list and we can select the very first item from it.

This isn't what it intends to demonstrate though.

> However 2. being the the alternative to 1. doesn't make any sense.

I don't think so, why?

harshhes:
I think that this 
> ```
> >>> response.css('title::text')[0].get()
> 'Quotes to Scrape'
> ```
should be replaced with this if we are giving the above code as a alternative:
> ```
> >>> response.css('title::text').getall()[0]
> 'Quotes to Scrape'
> ```
 It think because as `getall()` returns a list and we can select the very first item from it. 

I am just 2-3 levels up from a beginner, i thought to make a PR with correction if that could be of any use.

I might be wrong but that's how i got it. If this is of any use and is ethical you can use, otherwise close this PR

**Thank you for responding!**

wRAR:
> should be replaced with this if we are giving the above code as a alternative:

I understand this, I asked why do you think this.

> It think because as getall() returns a list and we can select the very first item from it.

I don't think this is a reason to replace the existing example. Do you think the existing one is wrong?

harshhes:
> I don't think this is a reason to replace the existing example. Do you think the existing one is wrong?

I don't think it is wrong, nor do i bragging it is a wrong approach. The point is find the one i have shared more understandable to beginners.

wRAR:
More understandable or not, it's different code. And the next few lines and snippets explain it, talking about `SelectorList`.

PR #5720, opened 2022-11-21
Remove `FilteringLinkExtractor`

Laerte:
This class was deprecated on [Scrapy 2.0.0 (2020-03-03)](https://docs.scrapy.org/en/2.0/news.html#deprecations).

I was in doubt abouve moving the private functions to `scrapy/linkextractors/lxmlhtml.py` then i discovered that is used in other parts of code too so i leaved like it was.

I followed the instructions from the [original PR](https://github.com/scrapy/scrapy/pull/4045):
> assuming that upon removal **its code can go into LxmlLinkExtractor**.

### DIFF
--- a/scrapy/linkextractors/__init__.py
+++ b/scrapy/linkextractors/__init__.py
@@ -6,18 +6,6 @@ This package contains a collection of Link Extractors.
 For more info see docs/topics/link-extractors.rst
 """
 import re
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):
:rocket:

wRAR (approved):

PR #5722, opened 2022-11-21
Adds get_virtual_size to _check_limit Logging

### LINKED ISSUE #5717: Add logging functionality to `memusage` extension
## Summary

To add logging functionality to memusage extension.

## Motivation

Scrapy jobs with `MEMUSAGE_ENABLED : True` and defined `MEMUSAGE_LIMIT_MB`  (all jobs on scrapy cloud) can be stopped early due to overuse of RAM memory and receive  `memusage_exceeded` outcome.

First thing required to debug RAM memory leaks - is to identify.. pattern of RAM memory usage.
Is RAM usage continuously increased at higher rates during runtime?
or Is RAM usage rapidly increased over limit in last several minutes after hours or even days of stable runtime performance?
Each reason require different approaches to debug RAM memory leaks.

It will be much easier to debug this if  value of `self.get_virtual_size()` will be added to log in  `_check_limit` method of `memusage` extension
https://github.com/scrapy/scrapy/blob/6ded3cf4cd134b615239babe28bb28c3ff524b05/scrapy/extensions/memusage.py#L77-L89

Applying `MEMUSAGE_WARNING_MB` setting to ~80-90% of `MEMUSAGE_LIMIT_MB` - current implementation of `memusage` extension warns only 1 time so it is not enough data for this.

Manually subclass `memusage` extension with similar changes - as well as any other option it will require to reschedule job. It may be not siutable for jobs with several days(and more) total runtime. So from this side it is preferable that it will be applied in scrapy itself and with enabled this loggin by default.
 
## Additional context

Similar functionality previously requested here https://github.com/scrapy/scrapy/issues/2173

JazzGlobal:
Fixes #5717 

Request for additional memory logging when scrapy jobs with `MEMUSAGE_ENABLED : True` and `MEMUSAGE_LIMIT_MB` defined are stopped early. This should further assist with memory debugging by providing the size of the current process.

### DIFF
--- a/scrapy/extensions/memusage.py
+++ b/scrapy/extensions/memusage.py
@@ -79,7 +79,7 @@ class MemoryUsage:
             self.crawler.stats.set_value('memusage/limit_reached', 1)
             mem = self.limit / 1024 / 1024
             logger.error("Memory usage exceeded %(memusage)dM. Shutting down Scrapy...",
-                         {'memusage': mem}, extra={'crawler': self.crawler})
+                         {'memusage': mem, 'virtualsize': self.get_virtual_size()}, extra={'crawler': self.crawler})
             if self.notify_mails:
                 subj = (
                     f"{self.crawler.settings['BOT_NAME']} terminated: "
--- a/scrapy/extensions/memusage.py
+++ b/scrapy/extensions/memusage.py
@@ -79,7 +79,7 @@ class MemoryUsage:
             self.crawler.stats.set_value('memusage/limit_reached', 1)
             mem = self.limit / 1024 / 1024
             logger.error("Memory usage exceeded %(memusage)dM. Shutting down Scrapy...",
-                         {'memusage': mem, 'virtualsize': self.get_virtual_size()}, extra={'crawler': self.crawler})
+                         {'memusage': mem}, extra={'crawler': self.crawler})
             if self.notify_mails:
                 subj = (
                     f"{self.crawler.settings['BOT_NAME']} terminated: "
@@ -92,6 +92,8 @@ class MemoryUsage:
                 self.crawler.engine.close_spider(self.crawler.engine.spider, 'memusage_exceeded')
             else:
                 self.crawler.stop()
+        else:
+            logger.info("Current memory usage is %(virtualsize)dM", {'virtualsize': self.get_virtual_size()})
 
     def _check_warning(self):
         if self.warned:  # warn only once
--- a/scrapy/extensions/memusage.py
+++ b/scrapy/extensions/memusage.py
@@ -75,7 +75,8 @@ class MemoryUsage:
         self.crawler.stats.max_value('memusage/max', self.get_virtual_size())
 
     def _check_limit(self):
-        if self.get_virtual_size() > self.limit:
+        current_mem_usage = self.get_virtual_size()
+        if current_mem_usage > self.limit:
             self.crawler.stats.set_value('memusage/limit_reached', 1)
             mem = self.limit / 1024 / 1024
             logger.error("Memory usage exceeded %(memusage)dM. Shutting down Scrapy...",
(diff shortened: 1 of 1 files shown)


Gallaecio:
I believe the point of https://github.com/scrapy/scrapy/issues/5717 is to log the memory currently in use *every time* `_check_limit` is checked, not only when it exceeds the limit. This is so that you can study the evolution of memory consumption from the logs and that way identify the pattern of memory usage, i.e. whether the memory suddenly increased towards the end or it increased steadily at all times.

JazzGlobal:
> I believe the point of #5717 is to log the memory currently in use _every time_ `_check_limit` is called, not only when it exceeds the limit. This is so that you can study the evolution of memory consumption from the logs and that way identify the pattern of memory usage, i.e. whether the memory suddenly increased towards the end or it increased steadily at all times.
> 
> So I think you need to define an `else` for the main `if` of the method, and log the memory currently in use from there.

That makes more sense, thank you. I've removed the initial addition (I wasn't even logging it anyway) and made changes as per your suggestion.

scrapy/extensions/memusage.py
```diff
@@ -92,6 +92,8 @@ def _check_limit(self):
                 self.crawler.engine.close_spider(self.crawler.engine.spider, 'memusage_exceeded')
             else:
                 self.crawler.stop()
+        else:
+            logger.info("Current memory usage is %(virtualsize)dM", {'virtualsize': self.get_virtual_size()})
```
Gallaecio, line 96: Rather than calling `self.get_virtual_size()` twice, why not declare a variable with its value at the beginning of the method, and use that value here and on the `if` statement?

Also, you are logging bytes as megabytes. You need to do `size / 1024 / 1024` as done above with the `mem` variable, but for the outcome of `self.get_virtual_size()`.

Also, for precision, maybe we should change the existing logging messages and your own message to use `MiB` instead of `M` as unit for [mebibytes](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units).

Gallaecio: > If so, this would just require the label be changed in the logging.

Yes, my point is that we are logging the value as e.g. 100M, but `M` is not a valid unit, it should be MB or, better yet, MiB.

But it is an issue from the original code, so I am OK with not addressing it here.

scrapy/extensions/memusage.py
```diff
@@ -92,6 +93,8 @@ def _check_limit(self):
                 self.crawler.engine.close_spider(self.crawler.engine.spider, 'memusage_exceeded')
             else:
                 self.crawler.stop()
+        else:
+            logger.info("Current memory usage is %(virtualsize)dM", {'virtualsize': current_mem_usage / 1024 / 1024})
```
kmike, line 97: I think this message might be misleading. We're reporting ru_maxrss, which is the max memory used by the current process over its lifetime, not the current memory usage.

scrapy/extensions/memusage.py
```diff
@@ -75,7 +75,8 @@ def update(self):
         self.crawler.stats.max_value('memusage/max', self.get_virtual_size())
 
     def _check_limit(self):
-        if self.get_virtual_size() > self.limit:
+        current_mem_usage = self.get_virtual_size()
```
kmike, line 78: similarly to the other comment, that's not the current memory usage, it's max memory usage

Gallaecio:
@JazzGlobal How do you feel about creating a new pull request to address the feedback from @kmike?

kmike:
It might also be good to solve https://github.com/scrapy/scrapy/issues/2173 eventually; it seems like an easy fix, which would help with cases like this, without going through extensions one-by-one.

JazzGlobal:
Still getting my feet wet with contributing to open source so I'd rather keep this guy separate at a minimum (and trying to adhere to the guidelines to get the PRs moving fast). I can start a new PR on a new branch to solve #2173 . 

@kmike @Gallaecio

JazzGlobal:
> Still getting my feet wet with contributing to open source so I'd rather keep this guy separate at a minimum (and trying to adhere to the guidelines to get the PRs moving fast). I can start a new PR on a new branch to solve #2173 . 
> 
> @kmike @Gallaecio 

I made this comment before seeing that this had been merged. Silly mobile apps ...

scrapy/extensions/memusage.py
JazzGlobal: Makes sense to me. Thank you for the clarification

Issue #5729, opened 2022-11-23
Remove settings._DictProxy


wRAR:
As discussed in #5452, `scrapy.settings._DictProxy` isn't needed anymore so we can remove it (looks like we decided to do that without deprecating it).

oldsiks:
scrapy don't support socks5~

Issue #5725, opened 2022-11-22
Remove scrapy.telnet from scrapy.utils.deprecate.DEPRECATION_RULES


wRAR:
`scrapy.telnet` was removed in 1.7.0 but some handling for that remains in `scrapy.utils.deprecate.DEPRECATION_RULES`.

PR #5730, opened 2022-11-23
Cleanup settings._DictProxy and scrapy.telnet

Hanzallah:
This change removes the `_DictProxy` class from `scrapy/settings/__init__.py` since it is no longer used. It also removes `scrapy.telnet` from the `DEPRECATION_RULES` as part of the cleanup. It resolves https://github.com/scrapy/scrapy/issues/5729, resolves https://github.com/scrapy/scrapy/issues/5725.

### DIFF
--- a/scrapy/utils/deprecate.py
+++ b/scrapy/utils/deprecate.py
@@ -126,9 +126,7 @@ def _clspath(cls, forced=None):
     return f'{cls.__module__}.{cls.__name__}'
 
 
-DEPRECATION_RULES = [
-    ('scrapy.telnet.', 'scrapy.extensions.telnet.'),
-]
+DEPRECATION_RULES = []
 
 
 def update_classpath(path):
--- a/scrapy/utils/deprecate.py
+++ b/scrapy/utils/deprecate.py
@@ -2,6 +2,7 @@
 
 import warnings
 import inspect
+from typing import List, Tuple
 from scrapy.exceptions import ScrapyDeprecationWarning
 
 
@@ -126,7 +127,7 @@ def _clspath(cls, forced=None):
     return f'{cls.__module__}.{cls.__name__}'
 
 
-DEPRECATION_RULES = []
+DEPRECATION_RULES: List[Tuple[str, str]] = []
 
 
 def update_classpath(path):
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -435,30 +435,6 @@ class BaseSettings(MutableMapping):
             p.text(pformat(self.copy_to_dict()))
 
 
-class _DictProxy(MutableMapping):
-
-    def __init__(self, settings, priority):
-        self.o = {}
-        self.settings = settings
-        self.priority = priority
-
-    def __len__(self):
-        return len(self.o)
-
-    def __getitem__(self, k):
-        return self.o[k]
-
-    def __setitem__(self, k, v):
-        self.settings.set(k, v, priority=self.priority)
-        self.o[k] = v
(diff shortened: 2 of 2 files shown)


scrapy/utils/deprecate.py
```diff
@@ -126,9 +126,7 @@ def _clspath(cls, forced=None):
     return f'{cls.__module__}.{cls.__name__}'
 
 
-DEPRECATION_RULES = [
-    ('scrapy.telnet.', 'scrapy.extensions.telnet.'),
-]
+DEPRECATION_RULES = []
 
 
 def update_classpath(path):
```
Laerte, line 132: I think may we should remove this function as well, since now it will just return the provided path, make sense @wRAR?

And probably pass a lambda function to `convert` parameter below (and remove from tests).
https://github.com/scrapy/scrapy/blob/45c2bd7d9c1d6adf0e051842c422c5d6af1d0a17/scrapy/utils/conf.py#L15

Hanzallah: If there is a possibility of repopulating the `DEPRECATION_RULES` list in the future we should keep the `update_classpath` method otherwise similar logic will need to be re-written.

Laerte: I think it will be not the case since `@deprecated` decorator already have this functionality:
https://github.com/scrapy/scrapy/blob/45c2bd7d9c1d6adf0e051842c422c5d6af1d0a17/scrapy/utils/decorators.py#L9

But let's see what Scrapy team has to say about.

wRAR: `update_classpath` is used in the code that parses object dicts such as `DOWNLOADER_MIDDLEWARES`: https://github.com/scrapy/scrapy/blob/45c2bd7d9c1d6adf0e051842c422c5d6af1d0a17/scrapy/utils/conf.py#L15

I think you can't use `@deprecated` for this.

Laerte: Yeah but the logic depends all on `DEPRECATION_RULES` variable right?

```python
    for prefix, replacement in DEPRECATION_RULES:
        if isinstance(path, str) and path.startswith(prefix):
            new_path = path.replace(prefix, replacement, 1)
            warnings.warn(f"`{path}` class is deprecated, use `{new_path}` instead",
                          ScrapyDeprecationWarning)
            return new_path
    return path
```
If nothing is provided it will just return what was originally send by parameter right?

wRAR: Right.

Hanzallah:
@wRAR, is this PR ready to be merged?

wRAR:
@Hanzallah we need two approvals from the core contributors to merge a PR.

wRAR (approved):

Gallaecio (approved):

scrapy/utils/deprecate.py
```diff
@@ -126,9 +126,7 @@ def _clspath(cls, forced=None):
     return f'{cls.__module__}.{cls.__name__}'
 
 
-DEPRECATION_RULES = [
-    ('scrapy.telnet.', 'scrapy.extensions.telnet.'),
-]
+DEPRECATION_RULES = []
```
Gallaecio, line 129: Based on the failing `typing` job, we need to mark this as `List[Tuple[str, str]]` now that the type cannot be inferred from the value.

Hanzallah: @Gallaecio I have now added the type annotations to the PR.

Gallaecio:
Great, thanks!

PR #5731, opened 2022-11-23
Remove `noconnect` deprecate code

Laerte:
- [x] Deprecated on [Scrapy 2.0.0 (2020-03-03)](https://docs.scrapy.org/en/latest/news.html#id47)
- [x] Deleted unused variables

I was in doubt about how to approach this one, so i basically remove all logic that depends upon `omitConnectTunnel`, maybe should we also pop the `b'noconnect` parameter?

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -3,7 +3,6 @@
 import ipaddress
 import logging
 import re
-import warnings
 from contextlib import suppress
 from io import BytesIO
 from time import time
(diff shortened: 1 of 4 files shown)


wRAR:
The logic change looks correct. Not sure about removing the param, is it removed by the current code?

Laerte:
@wRAR Is not, i just asked because i was unsure if passing the proxyParams with `?noconnect` could break something but since we not support anymore i think is not a problem.

Laerte:
The error in **Windows / tests (3.10, asyncio)** seems unrelated with the change.

Gallaecio:
> The error in **Windows / tests (3.10, asyncio)** seems unrelated with the change.

Rerunning now…

Gallaecio (approved):

wRAR (approved):

Issue #5726, opened 2022-11-22
Backward compatibility in utils.conf.build_component_list


wRAR:
There is some code from 2015 in `scrapy.utils.conf.build_component_list` marked as "Backward compatibility for old (base, custom) call signature", which was added in #1586. I couldn't understand after a quick glance why is it "backward compatibility" but if it's something deprecated we should deprecare it properly with a message, and if it's a properly supported code path we should remove the comments.

wRAR:
Actually, as #1586 "is a partial reversal of #1149", the `(base, custom, convert=update_classpath)` signature is the oldest one, #1149 replaced it with the `(compdict, convert=update_classpath)` one and #1586 restored the support for both. So I think we should properly deprecate the old signature and later remove the support for it.

wRAR:
I've tentatively added this to the next milestone so that we could deprecate this earlier and remove this earlier, but it shouldn't be a blocker.

wRAR:
Yeah, I think it's fine to merge it (but not close this issue) as the status of that code is now documented here.

PR #5732, opened 2022-11-24
Remove unnecessary backwards compatibility comments

burneyh:
Related: https://github.com/scrapy/scrapy/issues/5726. The code seems to still be in use and is being invoked therefore, we can remove the backwards compatibility comment since it is part of the feature still. I didn't come across any code that would suggest this use-case will never be executed.


Gallaecio:
https://github.com/scrapy/scrapy/issues/5726#issuecomment-1326286464

wRAR (approved):

Gallaecio:
Thanks!

Issue #5584, opened 2022-07-29
Deprecate Spider.max_concurrent_requests


Gallaecio:
This is an undocumented spider attribute that (non-intuitively, in my opinion) overrides the `CONCURRENT_REQUEST_PER_DOMAIN` setting or the `CONCURRENT_REQUEST_PER_IP` setting.

wRAR:
Done in #6994

PR #5733, opened 2022-11-24
Rename Spider.max_concurrent_requests

Hanzallah:
This resolves https://github.com/scrapy/scrapy/issues/5584. I agree with the naming suggested in the issue and this PR performs that renaming.

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -60,8 +60,8 @@ def _get_concurrency_delay(concurrency, spider, settings):
     if hasattr(spider, 'download_delay'):
         delay = spider.download_delay
 
-    if hasattr(spider, 'max_concurrent_requests'):
-        concurrency = spider.max_concurrent_requests
+    if hasattr(spider, 'slot_concurrent_requests'):
+        concurrency = spider.slot_concurrent_requests
 
     return concurrency, delay
 
--- a/extras/qpsclient.py
+++ b/extras/qpsclient.py
@@ -17,7 +17,7 @@ class QPSSpider(Spider):
     benchurl = 'http://localhost:8880/'
 
     # Max concurrency is limited by global CONCURRENT_REQUESTS setting
-    max_concurrent_requests = 8
+    slot_concurrent_requests = 8
     # Requests per second goal
     qps = None # same as: 1 / download_delay
     download_delay = None


scrapy/core/downloader/__init__.py
```diff
@@ -60,8 +60,8 @@ def _get_concurrency_delay(concurrency, spider, settings):
     if hasattr(spider, 'download_delay'):
         delay = spider.download_delay
 
-    if hasattr(spider, 'max_concurrent_requests'):
-        concurrency = spider.max_concurrent_requests
+    if hasattr(spider, 'slot_concurrent_requests'):
+        concurrency = spider.slot_concurrent_requests
```
Hanzallah, line 64: This is not a new line being added rather just a renaming happening. It might not already be covered by the test cases that is why the check does not succeed and shows,
```
Codecov / codecov/patch

scrapy/core/downloader/__init__.py#L64

Added line #L64 was not covered by tests
```

wRAR:
I'll wait for the opinion of @Gallaecio on this. Should this be impemented or discussed first, and should the old name stop being supported?

Gallaecio:
@Hanzallah When we tag an issue as `discuss`, the idea is that there is not yet a decision on how we want to address the issue, if we want to address it at all. So the expectation is that the issue first needs to be discussed in the issue thread, and only once there is a general agreement on how to proceed, we can remove the `discuss` tag and someone can go ahead and propose a fix. You have basically jumped a few steps here :sweat_smile: 

I am open to a renaming if there is any valid argument for keeping it, but my position is actually that we might want to remove the attribute instead. Spider attributes like this one often come from a time before Scrapy 1.0, which [added support for per-spider settings through the `custom_settings` class variable of spiders](https://docs.scrapy.org/en/latest/news.html#per-spider-settings-gsoc-2014). [This specific spider attribute was already there back when Scrapy moved to GitHub](https://github.com/scrapy/scrapy/commit/83dcf8aff91d0178ffbf4978fb757fc7e6d5d08c#diff-e998e10ab71f24389ab9a76146b2efc7f7969ce56cd52a8474d7ef321be1a7f5).

In any case, regardless of whether we decide to rename or remove, it must be done in line with our [deprecation policy](https://docs.scrapy.org/en/latest/versioning.html#deprecation-policy), i.e. the old name must continue to work for a time, but trigger a deprecation warning when used.

Hanzallah:
@Gallaecio thank you for the clarification! :) I will go ahead and close this PR and follow up on the discussion in the issue.

Issue #4654, opened 2020-07-01
Blank lines after import statements


victor-torres:
This issue was originated from a [discussion](https://github.com/scrapy/scrapy/pull/4434/files#r446676616) on pull request #4434.

## External resources

This is what the [Blank Lines](https://www.python.org/dev/peps/pep-0008/#blank-lines) section from the PEP 8 proposes:

> Surround top-level function and class definitions with two blank lines.
>
> Method definitions inside a class are surrounded by a single blank line.
>
> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
> 
> Use blank lines in functions, sparingly, to indicate logical sections.
>
> Python accepts the control-L (i.e. ^L) form feed character as whitespace; Many tools treat these characters as page separators, so you may use them to separate pages of related sections of your file. Note, some editors and web-based code viewers may not recognize control-L as a form feed and will show another glyph in its place.

In the same line as PEP 8, [Google's Python Style Guide](https://google.github.io/styleguide/pyguide.html#35-blank-lines) also only mentions top-level function and class definitions:

> Two blank lines between top-level definitions, be they function or class definitions. One blank line between method definitions and between the class line and the first method. No blank line following a def line. Use single blank lines as you judge appropriate within functions or methods.

## Our usage

Let's take the `logger` variable definition, something very common across Python projects, as an example.

In Scrapy, we have some differences (both valid according to PEP 8):

- it's surrounded by two blank lines in [scrapy/downloadermiddlewares/cookies.py](https://github.com/scrapy/scrapy/blob/1d2b692666f9f5265e4845f8a30f04089515f7c5/scrapy/downloadermiddlewares/cookies.py) and in [scrapy/utils/signal.py](https://github.com/scrapy/scrapy/blob/1d2b692666f9f5265e4845f8a30f04089515f7c5/scrapy/utils/signal.py);
- but it's surrounded by only one blank line in [scrapy/spidermiddlewares/urllength.py](https://github.com/scrapy/scrapy/blob/dfbe1d95071acfbba159cea051530749b1684460/scrapy/spidermiddlewares/urllength.py) and in [scrapy/core/engine.py](https://github.com/scrapy/scrapy/blob/3d58e5e3be9e266b9b0f89b094823b76511a21b1/scrapy/core/engine.py)

[...]

PR #5677, opened 2022-10-11
Pylint issues

marksmayo:
Tidied up .pylintrc file, removed unnecessary ignores and fixed a bunch of simple ones.

PR #5734, opened 2022-11-24
applying black to all the code

emarondan:
Applying black formatter to all scrapy code as part of an internal task.

Closes https://github.com/scrapy/scrapy/issues/4658 and closes https://github.com/scrapy/scrapy/issues/4654.


wRAR:
Thanks!

Several points I want to mention:

1. If you ran black with non-default options (e.g. with a non-default line length) you should add a config so that simply running `black` has the same outcome.
2. Can you please mention in the original message that this PR closes #4658 and #4654 so that they are autoclosed?
3. I see that both pylint and flake8 checks failed, we may need to silence the related warnings if they conflict with the `black` standard.
4. As mentioned in #4658, we need to add `.git-blame-ignore-revs`, but probably as a separate PR?
5. And finally, I feel like we should merge this only after both recent huge PRs (#5677 and #5682) are merged, but on the other hand this means we will need to redo and force-push the main commit of this PR later.

Gallaecio:
> probably as a separate PR?

No, we can add it here, and we probably should, we just need to make sure not to squash-and-merge (or rebase-and-merge) the changes, and instead go for a good-old, regular merge commit that does not change the hashes of these commits.

Gallaecio:
> 5\. And finally, I feel like we should merge this only after both recent huge PRs ([Pylint issues #5677](https://github.com/scrapy/scrapy/pull/5677) and [use pathlib #5682](https://github.com/scrapy/scrapy/pull/5682)) are merged, but on the other hand this means we will need to redo and force-push the main commit of this PR later.

That is a very good point :disappointed:

kmike:
Hey! Based on experience working with https://github.com/scrapy/parsel, it's a bit annoying to have black CI checks, but no easy documented way to configure black, setup the environment.

What do you think about setting up pre-commit and adding instrutions to the docs as a part of this PR? See e.g. https://github.com/scrapinghub/web-poet#developing

wRAR:
The pathlib one is merged and the pylint one is close to be merged so work on this can be restarted quite soon (and in the meantime things that are not related to actually changing the code can be worked on).

tests/test_utils_misc/test_return_with_argument_inside_generator.py
```diff
@@ -12,7 +15,7 @@ def _indentation_error(*args, **kwargs):
 
 def top_level_return_something():
     """
-docstring
+    docstring
```
wRAR, line 18: We need to disable this in some way, because these tests are for testing such code.

Gallaecio:
> the pylint one is close to be merged

Merged now.

wRAR:
The changes that remove flake8 errors will most likely be rolled back when you run `black` again. As I said previously, in this case we will need to silence them in `.flake8` (probably same for the remaining pylint ones, I haven't checked the code that was flagged).

wRAR:
Great!

The next steps are:

- adding a config for the non-default `black` options
- addinfg the CI job, I think it should be done here to prove that the changes here are correct and sufficient
- adding the pre-commit stuff (maybe as a separate PR)

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 119
```
kmike, line 2: I'd prefer to keep black default here.

Gallaecio:
We also need to add the tox-powered black check to https://github.com/scrapy/scrapy/blob/master/.github/workflows/checks.yml, to have the CI fail when black formatting is not used (e.g. if contributors do not install pre-commit).

Gallaecio:
OK, I think the only thing missing is… re-run black now that you removed back the line length setting.

Unless you are comfortable with re-applying everything from scratch as you did in the past, you could alternatively re-run black on a new commit, and add a follow-up commit that marks that second black commit to be ignored by git blame.

Gallaecio:
Hmm, the black job is still failing in CI, can you take a look?

Gallaecio (approved):

kmike:
hey! I think we should explain pre-commit in https://docs.scrapy.org/en/master/contributing.html.

docs/contributing.rst
```diff
@@ -179,6 +179,25 @@ Scrapy:
...
+The checks are defined in a configuration file called .pre-commit-config.yaml.
+
+By using pre-commit, developers can ensure that their code adheres to a consistent 
+style and passes certain tests before they make a commit. 
+This can help catch errors early in the development process and prevent them 
+from being pushed to the main branch.
+
+To use pre-commit, developers first need to install it on their local machine. 
+Once it is installed, they can specify the checks they want to run in the .pre-commit-config.yaml file. 
+After that, pre-commit will run automatically every time they make a commit.
```
Gallaecio, line 199: I don‘t think we need to focus so much on how to configure pre-commit for the repository, i.e. `.pre-commit-config.yaml` is already in the repository, and instead focus only on what contributors need to do to get it working. For example:

```suggestion
We use `pre-commit`_ to automatically address simple code issues before every 
commit.

Before you start writing a patch:

#.  `Install pre-commit <https://pre-commit.com/#installation>`_.

#.  On the root of your local clone of the Scrapy repository, run the following 
    command:

    .. code-block:: bash

       pre-commit install

Now our pre-commit hooks will run every time you create a Git commit. Upon 
finding issues, pre-commit hooks aborts your commit, and they either fix 
the corresponding issues automatically or only report them to you. If they fix 
the issues automatically, creating your commit again should succeed. Otherwise, 
you may need to address the corresponding issues manually first.
```

Also, I think it may be good to merge the first 2 bullet points from the section above, about code style, into a single one that indicates that we use black for code formatting, and that it can be executed manually with `tox` (and indicate how) or automatically with pre-commit.

Gallaecio: Mind the spacing when you apply the fix, empty lines make a different in reStructuredText.

Gallaecio:
```
/home/runner/work/scrapy/scrapy/docs/contributing.rst:193:Unexpected indentation.
```

Gallaecio: Today I Learned that copying code from GitHub suggestions breaks spacing. Sorry about that.

This should be easier to copy:

```
We use `pre-commit`_ to automatically address simple code issues before every 
commit.

.. _pre-commit: https://pre-commit.com/

Before you start writing a patch:

#.  `Install pre-commit <https://pre-commit.com/#installation>`_.

#.  On the root of your local clone of the Scrapy repository, run the following 
    command:

    .. code-block:: bash

       pre-commit install

Now pre-commit will check your changes every time you create a Git commit. Upon 
finding issues, pre-commit aborts your commit, and either fixes those issues 
automatically, or only reports them to you. If it fixes those issues 
automatically, creating your commit again should succeed. Otherwise, you may 
need to address the corresponding issues manually first.
```

The rendering is significantly different than what you have at the moment (e.g. “Before you start…” is interpreted as a definition list term in your current version).

I also included the link target for the “pre-commit” link at the beginning, that I missed in my suggestion, which I assume is why you ended up adding a reference to the header, and that is where the text now links. Mind that you need to remove the header reference line (the ```.. _pre-commit:`` line before the header) I would keep the reference that you added to the header, but this link should point to the pre-commit website instead.

Gallaecio (approved):

tox.ini
```diff
@@ -195,3 +195,10 @@ deps = {[docs]deps}
 setenv = {[docs]setenv}
 commands =
     sphinx-build -W -b linkcheck . {envtmpdir}/linkcheck
+
+[testenv:black]
+deps =
+    black==22.12.0
```
kmike, line 201: We don't need to do it in this PR, but we have black version pinned in 2 places: tox.ini and pre-commit config. It probably makes sense to just fail the CI if `pre-commit run --all-files` changes anything, instead of running these tools in tox manually.

kmike (approved):
Awesome, thanks @emarondan!

Issue #5737, opened 2022-11-27
Use isort


Gallaecio:
**WARNING:** This change should _not_ be implemented before https://github.com/scrapy/scrapy/pull/5734 is merged, because there is a high chance of conflicts.

We should use [isort](https://pycqa.github.io/isort/) to address import sorting and grouping:
- Fix it for all existing code.
- Automate it for new code.
- Check it for new code, in case someone accidentally skips the automation.

Fixing existing code should be a matter of running ``isort .`` on the root folder and committing all changes.

Automation should be handled through pre-commit. Chances are will provide pre-commit support for black, so here it would be a matter of extending that to include isort:

https://github.com/scrapy-plugins/scrapy-zyte-api/blob/016ef98092d1795b317c44c11326eefa58215f0d/.pre-commit-config.yaml#L2-L5

It also seems like we may need an `.isort.cfg` file for compatibility with black:

https://github.com/scrapy-plugins/scrapy-zyte-api/blob/016ef98092d1795b317c44c11326eefa58215f0d/.isort.cfg

Finally, we need to check on a CI job that isort has been indeed used for all new changes. First we should add an entry to [tox.ini](https://github.com/scrapy/scrapy/blob/master/tox.ini):

```ini
[testenv:isort]
deps = 
    isort==5.10.1
commands = 
    isort --check-only {posargs:.}
```

Then, we can add a job to [.github/workflows/checks.yml](https://github.com/scrapy/scrapy/blob/master/.github/workflows/checks.yml) in line with the existing ones, i.e.:

```yml
        - python-version: "3.11"
          env:
            TOXENV: isort
```

Gallaecio:
https://github.com/scrapy/scrapy/pull/5734 has not been merged yes, so it would not be a good idea to start with this already.

josh-nj-lawrence:
> #5734 has not been merged yet, so it would not be a good idea to start with this already.

No worries, I'll check out another issue to start with, perhaps 5734.

PR #5746, opened 2022-12-06
Added isort functionality and necessary config files (#5737)

cshel16:
Installed pre-commit functionality and hooked in isort. Also added necessary configurations for using isort and black.

**Will need to stay open until #5734 is merged, as per the issue description**

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -28,6 +28,9 @@ jobs:
         - python-version: "3.11"
           env:
             TOXENV: twinecheck
+        - python-version: "3.11"
+          env:
+            TOXENV: isort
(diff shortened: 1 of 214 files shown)


Gallaecio:
Thanks!

Could you have a look at those 4 failing tests, see if you can figure out why they are failing after these changes? (I could not figure it out from a quick look at the error messages, but the issue seems to be the same in all 4)

wRAR:
Note that some of those tests make real network requests so maybe something changed externally.

cshel16:
Yes, I'll do my best to look into these problems!

wRAR:
Implemenyed in #5806

Issue #5742, opened 2022-12-01
Scrapy Shell Always Raises RuntimeError but Works Fine


RealChitose:
### Description

Scrapy Shell always raises RuntimeError even though I request any URL, but results fine.

### Steps to Reproduce

Execute `scrapy shell 'https://any.url'`, return:
```2022-12-01 21:30:51 [scrapy.utils.log] INFO: Scrapy 2.7.1 started (bot: lightnovel)
2022-12-01 21:30:51 [scrapy.utils.log] INFO: Versions: lxml 4.9.1.0, libxml2 2.9.12, cssselect 1.1.0, parsel 1.7.0, w3lib 2.0.1, Twisted 22.10.0, Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)], pyOpenSSL 22.1.0 (OpenSSL 3.0.7 1 Nov 2022), cryptography 38.0.3, Platform Windows-10-10.0.19045-SP0
2022-12-01 21:30:51 [scrapy.crawler] INFO: Overridden settings:
{'AUTOTHROTTLE_ENABLED': True,
 'AUTOTHROTTLE_START_DELAY': 1,
 'AUTOTHROTTLE_TARGET_CONCURRENCY': 8.0,
 'BOT_NAME': 'lightnovel',
 'CONCURRENT_REQUESTS': 32,
 'CONCURRENT_REQUESTS_PER_IP': 16,
 'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter',
 'LOGSTATS_INTERVAL': 0,
 'LOG_LEVEL': 'INFO',
 'NEWSPIDER_MODULE': 'lightnovel.spiders',
 'REQUEST_FINGERPRINTER_IMPLEMENTATION': '2.7',
 'SPIDER_MODULES': ['lightnovel.spiders'],
 'TWISTED_REACTOR': 'twisted.internet.asyncioreactor.AsyncioSelectorReactor',
 'USER_AGENT': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
               '(KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36 '
               'Edg/107.0.1418.52 Scrapy/2.7.1 (+https://scrapy.org) '
               'LightnovelSpider/3.0'}
[...]

PR #5748, opened 2022-12-08
Fixes issue with calling fetch in scrapy shell.

### LINKED ISSUE #5740: asyncio exception in scrapy shell on Windows
We got a bug report that using `fetch(<url>)` in `scrapy shell` on Windows in a new project (so with `TWISTED_REACTOR` set to the asyncio one) raises "There is no current event loop" in `deferred_from_coro()`. The Python version is 3.9 and 3.10 and the Scrapy version is 2.7.1. It is possible that the recenty changed loop code indeed doesn't work correctly on Windows but I cannot test or fix it at this time.

alexpdev:
Reference Issue Fixes #5740 ,  #5742

You can recreate the issue with the following script.

```
import asyncio
import threading
from twisted.internet import asyncioreactor
from scrapy.utils.defer import deferred_from_coro
from scrapy.utils.reactor import get_asyncio_event_loop_policy

async def test_coro():
    pass

def test_deferred_from_coro():
    return deferred_from_coro(test_coro())

def trigger_warning_message():
    event_loop = get_asyncio_event_loop_policy().new_event_loop()
    asyncio.set_event_loop(event_loop)
    asyncioreactor.install()
    thread = threading.Thread(target=test_deferred_from_coro)
    thread.start()
    thread.join()

trigger_warning_message()
```

I was able to recreate this issue using `scrapy shell` and `fetch` on both windows and linux.  However it only occurs inside of a project with the `TWISTED_REACTOR` setting set to AsyncioSelectorReactor. 

What causes the issue is when `get_asyncio_event_loop_policy().get_event_loop()` is called, there is no event loop in the thread that the function is called from. Which raises an exception.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -279,6 +279,8 @@ class to be used.
 
 Note that the event loop class must inherit from :class:`asyncio.AbstractEventLoop`.
 
+This setting is not compatible with scrapy shell.
+
 .. caution:: Please be aware that, when using a non-default event loop
     (either defined via :setting:`ASYNCIO_EVENT_LOOP` or installed with
     :func:`~scrapy.utils.reactor.install_reactor`), Scrapy will call
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -279,7 +279,7 @@ class to be used.
 
 Note that the event loop class must inherit from :class:`asyncio.AbstractEventLoop`.
 
-This setting is not compatible with scrapy shell.
+This setting is not compatible with `scrapy shell`.
 
 .. caution:: Please be aware that, when using a non-default event loop
     (either defined via :setting:`ASYNCIO_EVENT_LOOP` or installed with
--- a/scrapy/utils/defer.py
+++ b/scrapy/utils/defer.py
@@ -267,7 +267,14 @@ def deferred_from_coro(o) -> Any:
             # that use asyncio, e.g. "await asyncio.sleep(1)"
             return ensureDeferred(o)
         # wrapping the coroutine into a Future and then into a Deferred, this requires AsyncioSelectorReactor
-        event_loop = get_asyncio_event_loop_policy().get_event_loop()
+        policy = get_asyncio_event_loop_policy()
+        try:
+            # getting event loop fails when run from separate thread like when using scrapy shell
+            event_loop = policy.get_event_loop()
+        except RuntimeError:
+            # create and set new event loop for this thread
+            event_loop = policy.new_event_loop()
+            asyncio.set_event_loop(event_loop)
         return Deferred.fromFuture(asyncio.ensure_future(o, loop=event_loop))
     return o
 
--- a/tests/test_command_shell.py
+++ b/tests/test_command_shell.py
@@ -115,3 +115,14 @@ class ShellTest(ProcessTest, SiteTest, unittest.TestCase):
         errcode, out, err = yield self.execute([url, '-c', 'item'], check_code=False)
         self.assertEqual(errcode, 1, out or err)
         self.assertIn(b'DNS lookup failed', err)
+
+    @defer.inlineCallbacks
+    def test_shell_fetch_async(self):
+        reactor_path = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
+        url = self.url('/html')
+        code = f"fetch('{url}')"
+        args = ["-c", code, "--set", f"TWISTED_REACTOR={reactor_path}"]
+        _, _, err = yield self.execute(args, check_code=True)
+        self.assertNotIn(
+            b"RuntimeError: There is no current event loop in thread", err
+        )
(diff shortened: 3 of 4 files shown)


alexpdev:
It passes all tests locally on both Linux and Windows, so I am not certain why it's failing here.

wRAR:
Can you please also add a test to test_command_shell.py that checks that there are no exceptions printed? I assume it's possible to write one that fails with the current Scrapy on Windows.

alexpdev:
> Can you please also add a test to test_command_shell.py that checks that there are no exceptions printed? I assume it's possible to write one that fails with the current Scrapy on Windows.

Sure thing

wRAR (approved):
Thanks!

scrapy/utils/defer.py
```diff
@@ -267,7 +267,14 @@ def deferred_from_coro(o) -> Any:
...
         # wrapping the coroutine into a Future and then into a Deferred, this requires AsyncioSelectorReactor
-        event_loop = get_asyncio_event_loop_policy().get_event_loop()
+        policy = get_asyncio_event_loop_policy()
+        try:
+            # getting event loop fails when run from separate thread like when using scrapy shell
+            event_loop = policy.get_event_loop()
+        except RuntimeError:
+            # create and set new event loop for this thread
+            event_loop = policy.new_event_loop()
+            asyncio.set_event_loop(event_loop)
```
Gallaecio, line 277: I don’t think this accounts for the `ASYNCIO_EVENT_LOOP`.

See the code were we set the event loop initially: https://github.com/scrapy/scrapy/blob/3a34fa839938d7640e56f3ce12ba60aed31a6ecf/scrapy/utils/reactor.py#L67-L80

I don’t see a clear way to address this, so I suggest:
- We document, in both the documentation of the setting (ASYNCIO_EVENT_LOOP) and that of the command (scrapy shell) that they are not compatible with each other.
- We create a task to make them compatible at some point in the future.
- We include a test here that demonstrates how `ASYNCIO_EVENT_LOOP` is not respected by scrapy shell. Hopefully the existing test for ASYNCIO_EVENT_LOOP helps with that. Funny enough, the test cannot run on Windows, as we use a Linux-specific alternative loop in those tests.

alexpdev: You are correct, I didn't even think about that.    I suppose the specificity of the event loop is less important in scrapy shell, since each call to `fetch` get's run in it's own thread.    

An alternative solution to this issue could be to explicitly set the event loop immediately after `Shell.fetch` creates the new thread.  Do you think that would be a more preferrable solution?

This issue is not exclusive to windows, and exists on Linux as well so the testing aspect shouldn't be an problem.

docs/topics/settings.rst
```diff
@@ -279,6 +279,8 @@ class to be used.
 
 Note that the event loop class must inherit from :class:`asyncio.AbstractEventLoop`.
 
+This setting is not compatible with scrapy shell.
```
wRAR, line 282: `scrapy shell` should be formatted (with double backticks probably?)

scrapy/utils/defer.py
Gallaecio: > Where should the test go for showing that the setting isn't respected?

I was hoping there would be a way to test this, but after looking into it, I am not sure how to do it myself :sweat: Our ASYNCIO_EVENT_LOOP tests are based on logs, but for the same reason it is not easy to use ASYNCIO_EVENT_LOOP in the first place, we do not log a message about whether or not it is being used.

I think we can skip the testing part.

Gallaecio:
I wonder if we could go a different way here: rather than create the loop if missing when needed, create it as we create the thread that causes the issue in the first place.

https://github.com/scrapy/scrapy/blob/26ebdbf4efd53bcb48da749576420c9f2fa6e28c/scrapy/commands/shell.py#L71-L80

For example, what if we modify CrawlerProcess.start, so that it takes care of initializing the loop if needed? From there we should have access to settings.

alexpdev:
> I wonder if we could go a different way here: rather than create the loop if missing when needed, create it as we create the thread that causes the issue in the first place.
> 
> https://github.com/scrapy/scrapy/blob/26ebdbf4efd53bcb48da749576420c9f2fa6e28c/scrapy/commands/shell.py#L71-L80
> 
> For example, what if we modify CrawlerProcess.start, so that it takes care of initializing the loop if needed? From there we should have access to settings, i.e. be able to respect ASYNCIO_EVENT_LOOP.

I had that same thought when you first made the observation.  But the thread call that causes the issue is the one called from the fetch command I thought, in the `shell` module.

https://github.com/scrapy/scrapy/blob/26ebdbf4efd53bcb48da749576420c9f2fa6e28c/scrapy/shell.py#L110

alexpdev:
I opened #5760 which implements the alternative approach of explicitly setting the event loop upon starting of the new thread created by `scrapy shell fetch` command.  That PR factors in the ASYNCIO_EVENT_LOOP setting, so I believe it might be the better approach.

alexpdev:
Im going to close this PR.  Thanks

PR #5754, opened 2022-12-11
Silence the `filelock` logger

### LINKED ISSUE #5753: Silence the `filelock` logger
After we started using `tldextract` we sometimes get log messages from `filelock` with the DEBUG level, it makes sense to silence them like we do for some other libraries in https://github.com/scrapy/scrapy/blob/fe60c1224e39aa3d85b20afd54566f135d9de085/scrapy/utils/log.py#L45-L59

Laerte:
resolves #5753

### DIFF
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -46,6 +46,9 @@ DEFAULT_LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
     'loggers': {
+        'filelock': {
+            'level': 'ERROR',
+        },
         'hpack': {
             'level': 'ERROR',
         },


scrapy/utils/log.py
```diff
@@ -46,6 +46,9 @@ def filter(self, record):
     'version': 1,
     'disable_existing_loggers': False,
     'loggers': {
+        'filelock': {
+            'level': 'ERROR'
```
wRAR, line 50: Can you add a trailing comma please?

wRAR (approved):

scrapy/utils/log.py
Laerte: Done.

Gallaecio (approved):

Issue #5745, opened 2022-12-05
Document Spider.update_settings


wRAR:
While overriding `Spider.update_settings` is a common practice to have dynamic custom settings, it looks like it's not documented anywhere.

It should also be mentioned in https://docs.scrapy.org/en/latest/topics/settings.html#settings-per-spider

anugcodes:
hey @wRAR ,  i would like to try this issue.But i am new here and don't know where to start.can you please help me out here.

wRAR:
@anugcodes are you asking about contributing in general or how to fix this issue?

anugcodes:
> @anugcodes are you asking about contributing in general or how to fix this issue?

I mean in general,how can I understand the codebase  and start with beginner friendly tasks.

wRAR:
@anugcodes https://docs.scrapy.org/en/latest/contributing.html

N00rAhmed:
hey @wRAR I would like to work on this issue but because I'm new to this repo I dont know where I should write the documentation for Spider.update_settings ?

wRAR:
@N00rAhmed the attributes of `scrapy.Spider` are documented in `docs/topics/spiders.rst`.

tuhinmallick:
Is this issue still open then I would like to work on it

wRAR:
@tuhinmallick there is an existing not-yet-complete PR linked above but you can try too.

PR #5757, opened 2022-12-12
Added Documentation for Spider.update_settings

N00rAhmed:
I added documentation for Spider.update_settings in relation to the issue: #5745
My contribution included a detailed explanation and a code example.


wRAR (changes requested):
I think you've misunderstood the task: it should talk about overriding this method in spider classes to modify the settings, not (just) about the existing implementation. Also it should be described in the same way other members of `Spider` are described, not as a separate section. And the "code example", which is the actual code of the base method, is not needed here. It would be useful to include an actual example of overriding it though. Also "latest version of Scrapy (beginning from 1.0)" isn't needed here, especially considering that 1.0 is 7 years old. Mentioning what do Scrapy settings do in general is not needed here either.

N00rAhmed:
Ok I will make these changes as soon as possible

N00rAhmed:
Hey I made the changes you requested. Can you check if I did it correctly?

wRAR:
This still doesn't document the method in the way other members of `Spider` are documented. Also, the example doesn't call the parent method. You also don't need to mention Scrapy in Scrapy docs or explain what does overriding a method means, it's enough to explain what does this method do and that you can override it to modify the settings. 

It's also maybe useful to mention that using it is preferable to using `custom_settings` when making spiders class hierarchies as parent `custom_settings` is shadowed but parent `update_settings()` can be called.

N00rAhmed:
hey I made the changes you requested, can you please review them.

wRAR:
It still doesn't document the method. It also has unrelated whitespace changes and I think the added parts have incorrect formatting for several reasons. They also move the `_spiderargs` label away from its section.

N00rAhmed:
where can I find the information on how to correctly document the method

docs/topics/spiders.rst
```diff
@@ -158,7 +158,7 @@ scrapy.Spider
        If you want to change the Requests used to start scraping a domain, this is
        the method to override. For example, if you need to start by logging in using
        a POST request, you could do::
-
+
```
Gallaecio, line 161: Please, make sure you are not making unintended changes. Review your own changes in GitHub before marking your pull request for review.

docs/topics/spiders.rst
```diff
@@ -259,6 +259,29 @@ to give data more structure you can use :class:`~scrapy.Item` objects::
             for href in response.xpath('//a/@href').getall():
                 yield scrapy.Request(response.urljoin(href), self.parse)
 
+.. _spiderargs:
```
Gallaecio, line 262: This line should not exist.

docs/topics/spiders.rst
```diff
@@ -259,6 +259,29 @@ to give data more structure you can use :class:`~scrapy.Item` objects::
...
+
+
+    .. method:: update_settings()
+        
+        The ``update_settings`` method allows you to modify the settings of a spider. It is a classmethod that is called to 
+        override the configuration of your spider. When called, it receives the spider's settings and passes ``custom_settings``
+        or an empty dictionary to the ``setdict`` method to register the setting with a priority of spider. It is recommended to
+        use ``update_settings`` rather than ``custom_settings`` when creating class hierarchies for spiders. This is because the parent's
+        ``custom_settings`` may be shadowed, but the parent's ``update_settings`` method can still be called.
```
Gallaecio, line 271: This is somewhat the right syntax to address this, but this documentation should go higher up in this document, before the “Let's see an example::” line above. I would put it between from_crawler and start_requests, since I assume that it when it is called; if it is called before from_crawler, then I would put it before.

Also, please consider using `.. automethod::`` instead of `.. method::`, and moving this documentation to the docstring of the actual method in code.

docs/topics/spiders.rst
```diff
@@ -259,6 +259,29 @@ to give data more structure you can use :class:`~scrapy.Item` objects::
...
+
+here is an example::
```
Gallaecio, line 273: This should have the same indentation as the paragraph above.

docs/topics/spiders.rst
```diff
@@ -259,6 +259,29 @@ to give data more structure you can use :class:`~scrapy.Item` objects::
...
+
+    class MySpider(scrapy.Spider):
+        name = 'myspider'
+
+        def update_settings(self, settings):
```
Gallaecio, line 278: This signature looks incorrect, [looking at the code](https://github.com/scrapy/scrapy/blob/4a5ef816044df868fc23a1dbe9e266bb2976db14/scrapy/spiders/__init__.py#L77-L79). Please, test your example manually and make sure it works as expected.

Gallaecio (changes requested):

docs/topics/spiders.rst
N00rAhmed: ok

docs/topics/spiders.rst
N00rAhmed: ok i've deleted it

docs/topics/spiders.rst
N00rAhmed: I've indented the code example to the same indentation as the paragraph

docs/topics/spiders.rst
```diff
@@ -145,6 +145,26 @@ scrapy.Spider
        :param kwargs: keyword arguments passed to the :meth:`__init__` method
        :type kwargs: dict
 
+   .. method:: update_settings()
```
Gallaecio, line 148: The signature seems incomplete, the parameters are missing.

docs/topics/spiders.rst
```diff
@@ -145,6 +145,26 @@ scrapy.Spider
...
+       or an empty dictionary to the ``setdict`` method to register the setting with a priority of spider. It is recommended to
+       use ``update_settings`` rather than ``custom_settings`` when creating class hierarchies for spiders. This is because the 
+       parent's ``custom_settings`` may be shadowed, but the parent's ``update_settings`` method can still be called.
+
+       here is an example::
+
+           class MySpider(scrapy.Spider):
+               name = 'myspider'
+
+               def update_settings(self, settings):
```
Gallaecio, line 161: This is not correct, you should be using `@classmethod` and `cls` instead of `self`.

https://github.com/scrapy/scrapy/blob/b337c986ca1188f4b26d30c9ae4bb7ff457ed505/scrapy/spiders/__init__.py#L80-L81

docs/topics/spiders.rst
```diff
@@ -145,6 +145,26 @@ scrapy.Spider
        :param kwargs: keyword arguments passed to the :meth:`__init__` method
        :type kwargs: dict
 
+   .. method:: update_settings()
+
+       The ``update_settings`` method allows you to modify the settings of a spider. It is a classmethod that is called to 
+       override the configuration of your spider. When called, it receives the spider's settings and passes ``custom_settings``
+       or an empty dictionary to the ``setdict`` method to register the setting with a priority of spider. It is recommended to
```
Gallaecio, line 152: Everything starting with “and passes” to the end of the sentence feels like an implementation detail, not what a user needs to know here. The user needs to know that they can edit the settings on the received `settings` object, and at what point in time this method is called in relation to other settting-defining code (e.g. are the update_settings classmethods of Scrapy components like middlewares called before or after this class method?).

Issue #5537, opened 2022-06-17
Might be nice to always call `exporter.finish_exporting`


romanchyla:
### Description

When an exception happens inside Exporter (for whatever reason) - itemcount never gets increased (that's OK). But then during closing, the code doesn't get chance to act

This is the place in the code:

https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/feedexport.py#L357

Perhaps the lines 357 and 353 should have been swapped?

### Additional context

I'm using `slot.storage` to open a database; and an exporter to start a session -- store() gets called (to close connection to the database) but session didn't have chance to commit -- yes, there is nothing to commit, but still - given the semantics (also visible from the code -- when `file` is supposed to be a file descriptor it might be worth consideration to always call `finish_exporting` to give the file descriptor chance to react before it is closed by the parent object)

necessary caveat: I might be mis-using the API and it is a minor detail, so please forgive if this looks silly (and feature not a bug)

wRAR:
Without looking at it for enough time, it looks like we should call `finish_exporting` if we called `start_exporting` before (not always, though), if we want to support various protocols and formats. The example I'm usually using when thinking about these things is an XML file that needs a closing tag if there was an opening one.

HatimZ:
Can I solve this issue. Looks like just need to add the slot.finish_exporting() before the if condition in the def _close_slot(self, slot, spider):

wRAR:
@HatimZ probably, but you need to make sure this code isn't called if `slot.start_exporting()` wasn't called (by reading the code, or maybe by making some test case). You'll also need to add some new test to test this behavior.

HatimZ:
@wRAR .Okay. Are there any examples/resources for writing test cases?

wRAR:
If you mean a test case for making sure it's ohly called when start_exporting() was called it would just be a test spider with some exporter enabled and some different edge cases like zero items, maybe an exception happening at some early point etc.

HatimZ:
I checked the code to make sure that `slot.finish_exporting()`  isn't called if `slot.start_exporting()` wasn't called  and I have come to the conclusion that indeed its not happening anywhere in the file . 
Even if `slot.finish_exporting()` was called without the start being called, it wouldn't do anything because it checks for the flag/boolean set by start. Hence even if it gets called , it wont do any harm. Am I right?
So Can I proceed to do what I suggested here ? 

[...]

PR #5758, opened 2022-12-12
Finish exporting for each start exporting

MattyMay:
Figured I'd take a crack at #5537 since it's been stale for a while.

This patch moves the call to `slot.finish_exporting()` up to before the conditional that checks if `slot.item_count > 0` so that if an exception occurs when an exporter's `export_item()` method is called, `slot.finish_exporting()` will still be called (as discussed in the issue).

I've included 4 new tests to that each check that:
1. if `slot.start_exporting()` was called that there is a call to `slot.finish_exporting()` after
2. `slot.finish_exporting()` is not called before a call to `slot.start_exporting()`. I.e. they assert that `slot.finish_exporting()` isn't called twice in a row without a call to `slot.start_exporting()` in between and that `slot.finish_exporting()` isn't called before an initial call to `slot.start_exporting`.

`test_start_finish_exporting_items_exception` fails before the patch and passes after the patch. The other 3 tests pass before and after.

Fixes https://github.com/scrapy/scrapy/issues/5537

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -350,11 +350,11 @@ class FeedExporter:
         return defer.DeferredList(deferred_list) if deferred_list else None
 
     def _close_slot(self, slot, spider):
+        slot.finish_exporting()
         if not slot.itemcount and not slot.store_empty:
             # We need to call slot.storage.store nonetheless to get the file
             # properly closed.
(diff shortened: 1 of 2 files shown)


Gallaecio:
Closing and reopening to trigger new CI jobs…

Gallaecio (approved):

wRAR (approved):

wRAR:
Thanks!

PR #5760, opened 2022-12-14
set asyncio event loop at start of new thread

### LINKED ISSUE #5740: asyncio exception in scrapy shell on Windows

alexpdev:
This is an alternative to PR https://github.com/scrapy/scrapy/pull/5748 which addresses the issues https://github.com/scrapy/scrapy/issues/5740 and https://github.com/scrapy/scrapy/issues/5742

The alternative approach used in this PR is to explicitly set the event loop in the new thread created by the scrapy shell when using the fetch command.

The main benefit to this strategy is the ability to honor the `ASYNCIO_EVENT_LOOP` setting.

Closes #5748, fixes #5740.

### DIFF
--- a/scrapy/shell.py
+++ b/scrapy/shell.py
@@ -21,6 +21,7 @@ from scrapy.utils.console import DEFAULT_PYTHON_SHELLS, start_python_console
 from scrapy.utils.datatypes import SequenceExclude
 from scrapy.utils.misc import load_object
 from scrapy.utils.response import open_in_browser
+from scrapy.utils.reactor import is_asyncio_reactor_installed, set_asyncio_event_loop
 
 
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):
Great work.

wRAR (approved):

PR #5761, opened 2022-12-15
docs: change DOWNLOAD_DELAY description format

yolile:
To be consistent with the other settings' descriptions

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -652,10 +652,9 @@ per ip address instead of per domain.
 
 .. _spider-download_delay-attribute:
 
-Note
+.. note::
 
-This delay can be set per spider using :attr:`download_delay` spider attribute.
-
+    This delay can be set per spider using :attr:`download_delay` spider attribute.
 
 .. setting:: DOWNLOAD_HANDLERS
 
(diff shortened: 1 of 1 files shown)


Gallaecio (changes requested):
I do not appreciate an improvement in the rewording.

The use of “Note” as a paragraph seems like a change for the worse. Sphinx has a directive, `note`, for that. But I am not sure a note format is right here, as this is the main place in the documentation where this spider attribute is covered.

The use of `:attr:` could be an improvement if it was pointing somewhere, but I don’t think it does.

yolile:
Ops, sorry, fixed now. My proposal is just to be consistent with the other `DOWNLOAD_` settings description format, e.g. https://docs.scrapy.org/en/latest/topics/settings.html#download-timeout

Gallaecio:
Makes sense. And I see that even though :attr: does not link anything, the formatting itself is a welcome change.

Thank you!

Gallaecio (approved):

wRAR (approved):

Gallaecio:
Note before merging: someone needs to make sure it renders OK, usually the note text starts on the same line as the directive, and I am not sure the syntax used here will work.

wRAR:
@Gallaecio web-poet uses the same syntax (even with the empty line in at least one place).

wRAR:
Actually even the same docs/topics/settings.rst includes the same syntax.

Unless I've misunderstood your concerns.

yolile:
It renders fine for me (in case this helps):
image

PR #5772, opened 2022-12-27
Added base_url arg to http.Response's urljoin method

toshi0304:
↓Actual, difficult cases that I have encountered.

page_url: "~.com/list/2(pageNumber)"
href: "123(articleId)"

in this case. current urljoion returns "-.com/list/2/123" not reaced to anywhere.
valid joined url is "~.com/123". It is reacheable to desired website.


wRAR:
> in this case. current urljoion returns "-.com/list/2/123"

Which is correct.

If `response.urljoin` is not useful, use `urllib.parse.urljoin` directly?

PR #5776, opened 2022-12-30
Deprecate direct invocation of `setup.py`

### LINKED ISSUE #5774: Deprecate direct invocation of `setup.py`
I was reading this article: [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) that explains why this is not a good practice anymore. I thought about discussing this. Should we change to new approach?

We have only a few places that we directly invoke `setup.py` so should be an easy task to replace it.

https://github.com/scrapy/scrapy/search?q=setup.py

Laerte:
Closes #5774

We don't need to install `setuptools` and `wheel` packages anymore in publish workflow because build create a virtual environment in order to build the package. We don't need to pass any argument to `build` command because the default behaviour already create a wheel/source distribution

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,8 +24,8 @@ jobs:
     - name: Publish to PyPI
       if: steps.check-release-tag.outputs.release_tag == 'true'
       run: |
-        pip install --upgrade setuptools wheel twine
-        python setup.py sdist bdist_wheel
+        pip install --upgrade build twine
+        python -m build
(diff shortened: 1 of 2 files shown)


wRAR:
I guess checking the publish action can only be done during the next release preparation :)

wRAR (approved):

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5778, opened 2022-12-31
Update tutorial.rst (grammar corrections)

dreadedhamish:
Just 2 grammar corrections.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -265,7 +265,7 @@ element, including its tags:
 >>> response.css('title').getall()
 ['<title>Quotes to Scrape</title>']
 
-The other thing is that the result of calling ``.getall()`` is a list: it is
+The second is that the result of calling ``.getall()`` is a list: it is
 possible that a selector returns more than one result, so we extract them all.
 When you know you just want the first result, as in this case, you can do:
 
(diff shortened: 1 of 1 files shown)


docs/intro/tutorial.rst
```diff
@@ -265,7 +265,7 @@ element, including its tags:
 >>> response.css('title').getall()
 ['<title>Quotes to Scrape</title>']
 
-The other thing is that the result of calling ``.getall()`` is a list: it is
+The second is that the result of calling ``.getall()`` is a list: it is
```
Gallaecio, line 268: Since there is not “the first” in earlier text, I think the proposed change is wrong, and the existing text is correct.

docs/intro/tutorial.rst
```diff
@@ -307,7 +307,7 @@ the :meth:`~scrapy.selector.SelectorList.re` method to extract using
 >>> response.css('title::text').re(r'(\w+) to (\w+)')
 ['Quotes', 'Scrape']
 
-In order to find the proper CSS selectors to use, you might find useful opening
+In order to find the proper CSS selectors to use, you might find it useful opening
```
Gallaecio, line 310: There is an additional change needed here, right?

```suggestion
In order to find the proper CSS selectors to use, you might find it useful to open
```

Issue #5765, opened 2022-12-19
AsyncioTest failures


wRAR:
There are two new failing tests (`AsyncioTest.test_install_asyncio_reactor` and `AsyncioTest.test_is_asyncio_reactor_installed`) on 3.10 and 3.11, and it looks like they started failing after updating to 3.10.9 and 3.11.1 (it may be a coincidence though). I see an asyncio-related change in those versions: "asyncio.get_event_loop() now only emits a deprecation warning when a new event loop was created implicitly. It no longer emits a deprecation warning if the current event loop was set." and it looks like it's the cause for the `test_install_asyncio_reactor` failure as that checks for a warning and maybe there is some new warning. But the second test is about installing the reactor and checking its type and the failure looks like the asyncio reactor is installed when it shouldn't.

wRAR:
Disabling `test_install_asyncio_reactor` makes `test_is_asyncio_reactor_installed` pass; I suspect that the former now installs the reactor when previously it didn't (initially this test had a comment "does nothing" but I don't remember why did this call do nothing), this needs future investigation. It may actually just depend on the test run order, maybe it actually broke some time ago. 

As for `test_install_asyncio_reactor`, the new warning is "DeprecationWarning('There is no current event loop')". Which is the same as we got in #5685 and we had a fix for it. Maybe we need a different fix for the current Python version (I wonder if we need to test older point releases, likely we don't).

PR #5782, opened 2023-01-02
test_install_asyncio_reactor: restore the original reactor when done

Gallaecio:
Fixes https://github.com/scrapy/scrapy/issues/5765

### DIFF
--- a/scrapy/utils/reactor.py
+++ b/scrapy/utils/reactor.py
@@ -81,6 +81,10 @@ def install_reactor(reactor_path, event_loop_path=None):
             installer()
 
 
+def _get_asyncio_event_loop():
+    return set_asyncio_event_loop(None)
+
+
 def set_asyncio_event_loop(event_loop_path):
     """Sets and returns the event loop with specified import path."""
     policy = get_asyncio_event_loop_policy()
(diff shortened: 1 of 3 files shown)


wRAR:
I think we shouldn't modify the reactor in the test process, either spawning subprocesses instead (like we already do with multiple CrawlerProcess tests) or doing read-only checks. There were some tests that installed a reactor and so affected later tets, I think they were removed since then. The test that is faling here was, as I mentioned in the issue, initially documented as doing nothing and basically expecting the install function to not raise an exception so we should find why is this no longer what it does and in the worst case remove/rewrite it.

scrapy/utils/reactor.py
```diff
@@ -90,11 +90,16 @@ def set_asyncio_event_loop(event_loop_path):
...
         except RuntimeError:
-            # `get_event_loop` is expected to fail when called from a new thread
+            # Raised by policy.get_running_loop() when called from a new thread
             # with no asyncio event loop yet installed. Such is the case when
-            # called from `scrapy shell`
+            # called from `scrapy shell`.
+            # In Python 3.10, but not in 3.11, policy.get_event_loop() seems to
+            # log a warning when called from the main thread, so we use
+            # asyncio.get_running_loop() to get an earlier RuntimeError as a
+            # workaround.
```
Gallaecio, line 102: I think this is happening, it feels like a Python 3.10 bug, but I have found no information about it.

If you comment out `asyncio.get_running_loop()` above, `test_install_asyncio_reactor` fails in Python 3.10 but _not_ in Python 3.11. You can try for yourself with:

```
tox -e py311,py310 -- tests/test_utils_asyncio.py::AsyncioTest tests/test_command_shell.py::ShellTest::test_shell_fetch_async
```

Gallaecio:
So, the current proposal seems to work, but… I don’t understand why. There is a scenario where, in Python 3.10, `asyncio.get_running_loop()` raises `RuntimeError` but `policy.get_event_loop()` does not, yet `policy.get_event_loop()` seems to be the right code to use, as `asyncio.get_running_loop()` does not seem to raise `RuntimeError` for the separate-thread scenario.

Gallaecio:
Also, it [seems](https://github.com/python/cpython/issues/100160#issuecomment-1345525581) they are reverting the warning soon, and re-introducing it later in Python 3.12.

scrapy/utils/reactor.py
Gallaecio: My Python 3.11 is 3.11.0, that explains the 3.10-3.11 difference I get locally.

Gallaecio:
Every attempt I was doing to fix the new errors was causing the scrapy shell test to fail until I used both get_running_loop and policy.get_event_loop together. I thought each was raising RuntimeError or not in different scenarios.

However, it turns out the problem was that in some scenarios a different part of the code, outside of this function, was causing the scrapy shell test to fail. Once I realized that, the path to a clean fix became more obvious.

Gallaecio:
Damn, get_running_loop “can only be called from a coroutine or a callback.”

wRAR (approved):

wRAR:
Great!

PR #5790, opened 2023-01-13
Enable unsafe legacy renegotiation

### LINKED ISSUE #5491: OpenSSL unsafe legacy renegotiation disabled error
### Description

I get an SSL issue on a working [site ](https://dorotheum.com) 
`twisted.web._newclient.ResponseNeverReceived: [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', '', 'unsafe legacy renegotiation disabled')]>]`

### Steps to Reproduce

1. `scrapy shell https://dorotheum.com`

**Expected behavior:** HTML page

**Actual behavior:** the error above

### Versions

```
Scrapy       : 2.6.1
lxml         : 4.8.0.0
libxml2      : 2.9.4
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 22.4.0
Python       : 3.9.12 (main, Mar 26 2022, 15:44:31) - [Clang 13.1.6 (clang-1316.0.21.2)]
pyOpenSSL    : 22.0.0 (OpenSSL 3.0.3 3 May 2022)
cryptography : 37.0.2
Platform     : macOS-12.2.1-arm64-arm-64bit
```

wRAR:
Fixes #5491.

I don't know if we can have a test for it: it would require a server configured in a certain way, I need to check if it's even possible when using OpenSSL 3 on the server side. Locally I tested it on https://dorotheum.com/ with "pyOpenSSL 21.0.0 ('OpenSSL 1.1.1j  16 Feb 2021'), cryptography 3.4.6" (no error and website is accessible with and without the change) and "pyOpenSSL 23.0.0 ('OpenSSL 3.0.7 1 Nov 2022'), cryptography 39.0.0" (website is accessible only with the change).

### DIFF
--- a/scrapy/utils/ssl.py
+++ b/scrapy/utils/ssl.py
@@ -1,12 +1,7 @@
-import OpenSSL
+import OpenSSL.SSL
 import OpenSSL._util as pyOpenSSLutil
 
 from scrapy.utils.python import to_unicode
-
-
-# The OpenSSL symbol is present since 1.1.1 but it's not currently supported in any version of pyOpenSSL.
-# Using the binding directly, as this code does, requires cryptography 2.4.
-SSL_OP_NO_TLSv1_3 = getattr(pyOpenSSLutil.lib, 'SSL_OP_NO_TLSv1_3', 0)
 
 
 def ffi_buf_to_string(buf):
@@ -22,9 +17,6 @@ def x509name_to_string(x509name):
 
 
 def get_temp_key_info(ssl_object):
-    if not hasattr(pyOpenSSLutil.lib, 'SSL_get_server_tmp_key'):  # requires OpenSSL 1.0.2
-        return None
-
     # adapted from OpenSSL apps/s_cb.c::ssl_print_tmp_key()
     temp_key_p = pyOpenSSLutil.ffi.new("EVP_PKEY **")
     if not pyOpenSSLutil.lib.SSL_get_server_tmp_key(ssl_object, temp_key_p):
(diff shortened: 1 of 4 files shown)


scrapy/utils/ssl.py
```diff
@@ -1,14 +1,9 @@
-import OpenSSL
+import OpenSSL.SSL
 import OpenSSL._util as pyOpenSSLutil
 
 from scrapy.utils.python import to_unicode
 
 
-# The OpenSSL symbol is present since 1.1.1 but it's not currently supported in any version of pyOpenSSL.
-# Using the binding directly, as this code does, requires cryptography 2.4.
-SSL_OP_NO_TLSv1_3 = getattr(pyOpenSSLutil.lib, 'SSL_OP_NO_TLSv1_3', 0)
```
wRAR, line 9: We can't deprecate this without some magic (probably the least magic way is the module-level `__getattr__`).

If we decide to deprecate it, this line can be changed into `SSL_OP_NO_TLSv1_3 = OpenSSL.SSL.OP_NO_TLSv1_3` and the deprecation will be done separately.

wRAR:
Yeah, I don't see which flags could I set on the mockserver SSL context so that it could reject normal OpenSSL 3 clients.

See also https://www.openssl.org/docs/manmaster/man3/SSL_get_secure_renegotiation_support.html#SECURE-RENEGOTIATION

Gallaecio (approved):

scrapy/utils/ssl.py
Gallaecio: No strong opinion either way from me.

PR #5800, opened 2023-01-19
Set `FEED_EXPORT_ENCODING='utf-8'` in the default template

### LINKED ISSUE #5797: Set FEED_EXPORT_ENCODING="utf-8" in the default template
The default value for FEED_EXPORT_ENCODING (`None`) means `ensure_ascii=True` for JSON exports, because that was the existing behavior before that option was added. A better default is `"utf-8"`, which creates JSON files with `ensure_ascii=False`, so we decided to add this setting to the default template like we currently do with some other settings.

The setting doc should also get a note about this like in https://docs.scrapy.org/en/latest/topics/settings.html#twisted-reactor

Laerte:
Resolves #5797

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -517,8 +517,7 @@ Use ``utf-8`` if you want UTF-8 for JSON too.
 
 .. versionchanged:: 2.8
    The :command:`startproject` command now sets this setting to
-   ``utf-8`` in the generated
-   ``settings.py`` file.
+   ``utf-8`` in the generated ``settings.py`` file.
 
 .. setting:: FEED_EXPORT_FIELDS
 
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -515,6 +515,11 @@ which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
 
 Use ``utf-8`` if you want UTF-8 for JSON too.
 
+.. versionchanged:: 2.8
+   The :command:`startproject` command now sets this setting to
+   ``utf-8`` in the generated
+   ``settings.py`` file.
+
 .. setting:: FEED_EXPORT_FIELDS
 
 FEED_EXPORT_FIELDS
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -515,7 +515,7 @@ which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
 
 Use ``utf-8`` if you want UTF-8 for JSON too.
 
-.. versionchanged:: 2.8
+.. versionchanged:: VERSION
    The :command:`startproject` command now sets this setting to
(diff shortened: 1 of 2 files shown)


tests/test_feedexport.py
```diff
@@ -1298,8 +1298,8 @@ def test_export_encoding(self):
         items = [dict({'foo': 'Test\xd6'})]
 
         formats = {
-            'json': '[{"foo": "Test\\u00d6"}]'.encode('utf-8'),
-            'jsonlines': '{"foo": "Test\\u00d6"}\n'.encode('utf-8'),
+            'json': '[{"foo": "TestÖ"}]'.encode('utf-8'),
+            'jsonlines': '{"foo": "TestÖ"}\n'.encode('utf-8'),
```
Laerte, line 1302:
```suggestion
            'json': b'[{"foo": "Test\xc3\x96"}]',
            'jsonlines': b'{"foo": "Test\xc3\x96"}\n',
```
We can use this syntax if you guys prefer. :-)

Laerte:
The only error is not related with the change, rerun should fix.

wRAR:
The idea was to add it to [the project template](https://github.com/scrapy/scrapy/blob/master/scrapy/templates/project/module/settings.py.tmpl), not change the default :) Sorry if that wasn't clear.

Laerte:
@wRAR Gotcha! sorry, check now!

docs/topics/feed-exports.rst
```diff
@@ -515,6 +515,11 @@ which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
 
 Use ``utf-8`` if you want UTF-8 for JSON too.
 
+.. versionchanged:: 2.8
```
wRAR, line 518: 2.8 here should be fine but please use `VERSION` next time to be more flexible.

docs/topics/feed-exports.rst
```diff
@@ -515,6 +515,11 @@ which uses safe numeric encoding (``\uXXXX`` sequences) for historic reasons.
...
+   The :command:`startproject` command now sets this setting to
+   ``utf-8`` in the generated
+   ``settings.py`` file.
```
wRAR, line 521: The line breaks here are weird :)

wRAR (approved):

docs/topics/feed-exports.rst
Laerte: Agreed! Fixed.

docs/topics/feed-exports.rst
Laerte:
```diff
-.. versionchanged:: 2.8
+.. versionchanged:: VERSION
```
Like this?

wRAR: Yes, it will be replaced during the release process.

Laerte: Cool, done as well.

Gallaecio (approved):

Issue #5739, opened 2022-11-29
Adding support for Path objects to APIs that take paths


wRAR:
As mentioned in #5682, we may have some public APIs that take paths and the implementations should now use `Path` objects so it should be trivial to also accept `Path` objects in these APIs. 

I think the first step is actually identifying such APIs.

kmike:
I think we don't have to solve it all at once, PRs with partial implementation are also welcome.

wRAR:
True.

harshit496:
@wRAR @kmike would like to take a first stab at this if it is not already resolved. Thanks!

wRAR:
@harshit496 it's not resolved, as you can see in the issue status.

harshit496:
@wRAR I tried searching the usages of `os.path ` in the whole code base and couldn't find any. 

```
@harshit496 ➜ /workspaces/scrapy (master) $ grep -rnw . -e 'from os'
./tests/test_feedexport.py:17:from os import PathLike
./build/lib/scrapy/utils/template.py:3:from os import PathLike
./build/lib/scrapy/commands/runspider.py:2:from os import PathLike
./build/lib/scrapy/squeues.py:7:from os import PathLike
./scrapy/utils/template.py:3:from os import PathLike
./scrapy/commands/runspider.py:2:from os import PathLike
./scrapy/squeues.py:7:from os import PathLike
@harshit496 ➜ /workspaces/scrapy (master) $ grep -rnw . -e 'from os import path'
@harshit496 ➜ /workspaces/scrapy (master) $ grep -rnw . -e 'os.path.abspath()'
@harshit496 ➜ /workspaces/scrapy (master) $ 
```

Gallaecio:
Yes, finding those APIs is probably not that easy.

One way may be to look at the settings, identify which ones are about paths, and see how they are handled.

Mind also cases where `import os` is used, and then `os.path`.

wRAR:
Not sure if searching for `os.path` usages is helpful here, as we are talking about APIs that take a string and then pass it to `pathlib.Path`. Actually we shouldn't even have os.path usages after the pathlib conversion.

alexpdev:
I just submitted PR #5801 which patches one instance of a Path object raising an exception.

ggold7046:
Hello, I'm new to opensource. Could anyone guide me in this project ?

wRAR:
@ggold7046 see https://scrapy.org/community/

[...]

PR #5801, opened 2023-01-20
Support Path Objects Issue #5739

alexpdev:
Located a couple of places where using Path object raises an exception. Relates to Issue https://github.com/scrapy/scrapy/issues/5739.

The issue can be reproduced by using a Path object in the "IMAGES_STORE" or "FILES_STORE" settings when configuring a file or image pipeline.

Code to reproduce:

```
from pathlib import Path
import scrapy

    name = 'myspider'
    custom_settings = {
        "IMAGES_URLS_FIELD": 'image_urls',
        "IMAGES_RESULT_FIELD": 'images',
        "IMAGES_STORE": Path.home() / 'Images',   # <-- here
        "ITEM_PIPELINES": {
            "scrapy.pipelines.images.ImagesPipeline": 1,
        }
    }

```

The unit tests I added will also trigger the error when tested without the changes.

And the exception that is raised:

```
2023-01-15 07:18:46 [twisted] CRITICAL:
  File "...\venv\lib\site-packages\twisted\internet\defer.py", line 1697, in _inlineCallbacks
    result = context.run(gen.send, result)

  ...
  File "...\venv\lib\site-packages\scrapy\pipelines\files.py", line 383, in _get_store
    return store_cls(uri)
  File "...\venv\lib\site-packages\scrapy\pipelines\files.py", line 43, in __init__
    if '://' in basedir:
TypeError: argument of type 'WindowsPath' is not iterable
```


scrapy/pipelines/files.py
```diff
@@ -65,8 +71,8 @@ def stat_file(self, path: str, info):
 
         return {"last_modified": last_modified, "checksum": checksum}
 
-    def _get_filesystem_path(self, path: str) -> Path:
-        path_comps = path.split("/")
+    def _get_filesystem_path(self, path: Union[str, PathLike]) -> Path:
+        path_comps = _to_string(path).split("/")
```
Gallaecio, line 75: Type-hint wise, I think this change also implies that code calling this function (e.g. the 2 methods above) now would also accept `Union[str, PathLike]`, so we should update their type hints accordingly.

scrapy/pipelines/files.py
```diff
@@ -9,12 +9,13 @@
 import mimetypes
 import os
 import time
+from os import PathLike
```
Gallaecio, line 12: Note that the latest `master` has introduced `isort` and CI will fail due to this line. See https://docs.scrapy.org/en/latest/contributing.html#pre-commit, it will fix this when you merge or rebase from the latest `master`.

Gallaecio (approved):

scrapy/pipelines/files.py
alexpdev: Fixed. thank you.

scrapy/pipelines/files.py
alexpdev: Got it. Thanks

wRAR (approved):
Thanks!

PR #5802, opened 2023-01-21
Simplify Publish workflow

Laerte:
Simplify publish workflow by using official PyPI publish action.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,6 @@ jobs:
           pip install --upgrade build twine
           python -m build
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1.6.4
+        uses: pypa/gh-action-pypi-publish@v1.6.4
         with:
           password: ${{ secrets.PYPI_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,31 +1,21 @@
 name: Publish
-on: [push]
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


Laerte:
💡 We could even simplify this workflow:

```yaml
name: Publish
on:
  push:
    tags:
      - "[0-9]+.[0-9]+.[0-9]+(rc[0-9]+|.dev[0-9]+)?$"

jobs:
  publish:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
          python-version: 3.11
      - run: | 
          pip install --upgrade build twine
          python -m build
      - name: Publish to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          password: ${{ secrets.PYPI_TOKEN }}
```
With this we can remove the **Check Tag** step and this action will be only executed when we push a new tag that matches our pattern.

Gallaecio (approved):
Thanks!

Gallaecio:
The proposed simplification also sounds good to me. In fact, we haven’t released an rc or dev version in years, and I don’t expect us to do so any time soon, so we could even remove those 2 patterns as well.

Laerte:
@Gallaecio I pushed the refactor, i also use the same workflow on my personal projet: https://github.com/Laerte/aes_pkcs5/blob/main/.github/workflows/publish.yml and works flawless!

Gallaecio (approved):
:heart:

kmike:
Thanks @Laerte!

PR #5805, opened 2023-01-25
Handle typing issues hidden by follow_imports=skip

wRAR:
This removes `follow_imports = skip` and fixes/ignores the issues that appeared.

~~There are many ignored issues regarding `twisted.internet.reactor`, it should often be possible to cast it into a suitable class/interface instead but I'm not sure how useful is that and also we still ignore interfaces (there is a [WIP mypy plugin for them](https://github.com/Shoobx/mypy-zope) which I decided not to use). I also need to try ignoring `twisted.internet.reactor` via `follow_imports = skip` but Im not sure if it will work.~~

~~I've found a typing issue in w3lib (https://github.com/scrapy/w3lib/pull/211) which we should fix and remove from here, and a possible error in the H2 code which we may need to check and rewrite.~~


.github/workflows/checks.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         - python-version: "3.11"
           env:
             TOXENV: pylint
-        - python-version: 3.7
+        - python-version: 3.8
```
wRAR, line 14: Needed because `types-lxml` is not available for 3.7.

3.7 EOL is in 4 months so I don't think this is a big problem.

scrapy/core/http2/agent.py
```diff
@@ -83,7 +88,7 @@ def _remove_connection(self, errors: List[BaseException], key: Tuple) -> None:
         pending_requests = self._pending_requests.pop(key, None)
         while pending_requests:
             d = pending_requests.popleft()
-            d.errback(errors)
+            d.errback(ResponseFailed(errors))
```
wRAR, line 91: This is indeed not covered and I guess this is why it wasn't caught by tests.

scrapy/core/scraper.py
```diff
@@ -203,10 +203,12 @@ def call_spider(
                 callback=callback, callbackKeywords=result.request.cb_kwargs
             )
         else:  # result is a Failure
-            result.request = request
+            # TODO: properly type adding this attribute to a Failure
```
Gallaecio, line 206: Is this TODO for this pull request or for a future one? (I’m OK with both)

scrapy/core/scraper.py
```diff
@@ -203,10 +203,12 @@ def call_spider(
...
+            result.request = request  # type: ignore[attr-defined]
             warn_on_generator_with_return_value(spider, request.errback)
             dfd = defer_fail(result)
-            dfd.addErrback(request.errback)
+            if request.errback:
+                dfd.addErrback(request.errback)
```
wRAR, line 211: `addErrback(None)` is allowed and does nothing, but it's not documented (there is a comment saying "Default value used to be None" so maybe it was documented before but is now kept for backwards compativility) so I changed this.

scrapy/http/request/form.py
```diff
@@ -50,7 +58,7 @@ def from_response(
         response: TextResponse,
         formname: Optional[str] = None,
         formid: Optional[str] = None,
-        formnumber: Optional[int] = 0,
```
wRAR, line 53: Passing None never worked.

Gallaecio (approved):

scrapy/core/scraper.py
wRAR: For a future one

PR #5806, opened 2023-01-25
Use isort

jxlil:
Added isort to scrapy and sorted all imports.

Closes https://github.com/scrapy/scrapy/issues/5737, fixes https://github.com/scrapy/scrapy/issues/5810.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -10,12 +10,6 @@ jobs:
         include:
         - python-version: "3.11"
           env:
-            TOXENV: security
-        - python-version: "3.11"
-          env:
-            TOXENV: flake8
-        - python-version: "3.11"
-          env:
             TOXENV: pylint
         - python-version: 3.7
           env:
@@ -26,13 +20,9 @@ jobs:
         - python-version: "3.11"
           env:
             TOXENV: twinecheck
-        - python-version: "3.11"
-          env:
-            TOXENV: black
 
     steps:
     - uses: actions/checkout@v3
-    - uses: pre-commit/action@v3.0.0
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
@@ -44,3 +34,9 @@ jobs:
       run: |
         pip install -U tox
         tox
+
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: pre-commit/action@v3.0.0
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -32,6 +32,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - uses: pre-commit/action@v3.0.0
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,3 @@
+[settings]
+profile = black
+multi_line_output = 3
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,3 +1,2 @@
 [settings]
 profile = black
-multi_line_output = 3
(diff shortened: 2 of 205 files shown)


Gallaecio:
@jxlil What do you think about making https://github.com/scrapy/scrapy/pull/5734#discussion_r1086488805 part of this pull request? Instead of adding isort to tox.ini, we could actually remove from tox.ini anything that we handle with pre-commit now, and have pre-commit run in a CI job.

jxlil:
@Gallaecio Sounds good. I will update this PR.

Gallaecio (commented):
> Sounds good. I will update this PR.

I was not just suggesting that you remove the tox entry for isort, but also add the pre-commit CI job, as described in https://github.com/scrapy/scrapy/pull/5734#discussion_r1086488805.

You don‘t have to do it, we can do it in a separate, later pull request, but then please add back the CI and tox configu you had. However we do it, after we merge this pull request we really want the CI to fail if any bad import sorting is introduced in the future.

.isort.cfg
```diff
@@ -0,0 +1,3 @@
+[settings]
+profile = black
+multi_line_output = 3
```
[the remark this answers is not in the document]
jxlil: You are right this is not necessay. This is the default settings for the black profile:

```
multi_line_output: 3
include_trailing_comma: True
force_grid_wrap: 0
use_parentheses: True
ensure_newline_before_comments: True
line_length: 88
split_on_trailing_comma: True
```

scrapy/__init__.py
```diff
@@ -8,12 +8,12 @@
...
-# Declare top-level shortcuts
-from scrapy.spiders import Spider
-from scrapy.http import Request, FormRequest
+from scrapy.http import FormRequest, Request
+from scrapy.item import Field, Item
 from scrapy.selector import Selector
-from scrapy.item import Item, Field
 
+# Declare top-level shortcuts
+from scrapy.spiders import Spider
```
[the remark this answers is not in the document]
jxlil: I did some tests with isort to avoid this in the future, but no luck. I'm going to make this change manually.

.github/workflows/checks.yml
```diff
@@ -32,6 +32,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - uses: pre-commit/action@v3.0.0
```
jxlil, line 35: I don't have much experience adding CI jobs. However I believe that this [action](https://github.com/pre-commit/action/blob/main/action.yml) should cause CI to fail if `pre-commit` makes any changes.

wRAR:
>          RuntimeError: The Poetry configuration is invalid:
>            - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'

Uhh. Is this some problem with the pre-commit action itself?

jxlil:
This problem is related to `isort`, updating to version `5.12.0` fixes the bug (https://github.com/PyCQA/isort/issues/2077). However, this version supports only Python `3.8+`, should I update the action to something like this?

```
- name: pre-commit
  uses: pre-commit/action@v3.0.0
  with:
    python-version: "3.11"
```


In the logs I see that pre-commit uses Python `3.10` by default.

Gallaecio:
The problem I see with the current CI approach is that it will run pre-commit for *every* check job. And also, we should remove the CI jobs that are not needed once we have a pre-commit CI job.

Let me make some changes to the CI file myself…

.pre-commit-config.yaml
```diff
@@ -8,12 +8,11 @@ repos:
   rev: 6.0.0
   hooks:
   - id: flake8
-- repo: https://github.com/PyCQA/pylint
-  rev: v2.15.6
-  hooks:
-  - id: pylint
-    args: [conftest.py, docs, extras, scrapy, setup.py, tests]
```
Gallaecio, line 15: Done to address https://github.com/scrapy/scrapy/issues/5810, since otherwise we would be removing pylint from checks.yml in this pull request to re-add it when we remove pylint from pre-commit.

Gallaecio (approved):
Looks good to me, thanks!

.github/workflows/checks.yml
Gallaecio: Great finding!

wRAR (approved):

PR #5814, opened 2023-01-30
apply black to templates and fixed unit tests Fixes #5809

### LINKED ISSUE #5809: Apply black to Python templates
We should look into getting black to apply to https://github.com/scrapy/scrapy/tree/master/scrapy/templates.

Mind that it would break https://github.com/scrapy/scrapy/pull/5808 at the moment, i.e. we might need to make sure template building code works with the new code style of templates.

alexpdev:
I ran black formatter on all of the template files in `scrapy/templates`.  There were very few changes that it made, most of which were changing single quotes `'` to double quotes `"`, or adding a single space in between the hash character and everything else for commented out values in the `settings.py` file.  

It did break a few unit tests for the `genspider` command used regex for string comparisons.  I fixed these by making the regex patterns ambiguous to single or double quotes. 

Fixes #5809 

This shouldn't have any effect on #5808 since I used the same quote ambiguous patterns for the regex in that PR as well.

### DIFF
--- a/scrapy/templates/project/module/settings.py.tmpl
+++ b/scrapy/templates/project/module/settings.py.tmpl
@@ -14,7 +14,7 @@ NEWSPIDER_MODULE = "$project_name.spiders"
 
 
 # Crawl responsibly by identifying yourself (and your website) on the user-agent
-#USER_AGENT = '$project_name (+http://www.yourdomain.com)'
+#USER_AGENT = "$project_name (+http://www.yourdomain.com)"
 
 # Obey robots.txt rules
 ROBOTSTXT_OBEY = True
@@ -38,32 +38,32 @@ ROBOTSTXT_OBEY = True
 
 # Override the default request headers:
 #DEFAULT_REQUEST_HEADERS = {
-#   'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
-#   'Accept-Language': 'en',
+#    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+#    "Accept-Language": "en",
 #}
 
 # Enable or disable spider middlewares
 # See https://docs.scrapy.org/en/latest/topics/spider-middleware.html
 #SPIDER_MIDDLEWARES = {
-#    '$project_name.middlewares.${ProjectName}SpiderMiddleware': 543,
+#    "$project_name.middlewares.${ProjectName}SpiderMiddleware": 543,
 #}
 
 # Enable or disable downloader middlewares
 # See https://docs.scrapy.org/en/latest/topics/downloader-middleware.html
 #DOWNLOADER_MIDDLEWARES = {
-#    '$project_name.middlewares.${ProjectName}DownloaderMiddleware': 543,
+#    "$project_name.middlewares.${ProjectName}DownloaderMiddleware": 543,
 #}
 
 # Enable or disable extensions
 # See https://docs.scrapy.org/en/latest/topics/extensions.html
 #EXTENSIONS = {
-#    'scrapy.extensions.telnet.TelnetConsole': None,
+#    "scrapy.extensions.telnet.TelnetConsole": None,
 #}
 
 # Configure item pipelines
 # See https://docs.scrapy.org/en/latest/topics/item-pipeline.html
 #ITEM_PIPELINES = {
-#    '$project_name.pipelines.${ProjectName}Pipeline': 300,
+#    "$project_name.pipelines.${ProjectName}Pipeline": 300,
 #}
 
 # Enable and configure the AutoThrottle extension (disabled by default)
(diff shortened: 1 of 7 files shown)


scrapy/templates/spiders/crawl.tmpl
```diff
@@ -4,13 +4,11 @@ from scrapy.spiders import CrawlSpider, Rule
...
-    allowed_domains = ['$domain']
-    start_urls = ['http://$domain/']
+    name = "$name"
+    allowed_domains = ["$domain"]
+    start_urls = ["http://$domain/"]
 
-    rules = (
-        Rule(LinkExtractor(allow=r'Items/'), callback='parse_item', follow=True),
-    )
+    rules = (Rule(LinkExtractor(allow=r"Items/"), callback="parse_item", follow=True),)
```
alexpdev, line 11: It did cram this into a single line which I though was unusual.

wRAR:
I think we should also run black on these files in pre-commit?

alexpdev:
> 

That I am not sure how to do.  When you attempt to run the tool on the .templ files it throws a syntax error because of all the `$variable_names` that aren't valid python variables.  The way I was able to apply it was to use the the scrapy cli to create generic versions, run black on those files and then copied over the changes to the templates.  

I suppose I could write up a script would do that whole process automatically?

Gallaecio:
I think for templates it might be OK for this to be a one-time thing, given we cannot simply rely on black. It is not only the variables, the commented-out code should be left as it was, it must be distinguishable from regular comments.

wRAR (approved):

Gallaecio:
> the commented-out code should be left as it was, it must be distinguishable from regular comments.

Could you revert the changes to commented-out code before we merge?

scrapy/templates/project/module/settings.py.tmpl
```diff
@@ -7,87 +7,87 @@
...
 
-SPIDER_MODULES = ['$project_name.spiders']
-NEWSPIDER_MODULE = '$project_name.spiders'
+SPIDER_MODULES = ["$project_name.spiders"]
+NEWSPIDER_MODULE = "$project_name.spiders"
 
 
 # Crawl responsibly by identifying yourself (and your website) on the user-agent
-#USER_AGENT = '$project_name (+http://www.yourdomain.com)'
+# USER_AGENT = '$project_name (+http://www.yourdomain.com)'
```
Gallaecio, line 17: e.g. 
```suggestion
#USER_AGENT = '$project_name (+http://www.yourdomain.com)'
```

scrapy/templates/spiders/csvfeed.tmpl
```diff
@@ -2,14 +2,14 @@ from scrapy.spiders import CSVFeedSpider
...
 class $classname(CSVFeedSpider):
-    name = '$name'
-    allowed_domains = ['$domain']
-    start_urls = ['http://$domain/feed.csv']
-    # headers = ['id', 'name', 'description', 'image_link']
-    # delimiter = '\t'
+    name = "$name"
+    allowed_domains = ["$domain"]
+    start_urls = ["http://$domain/feed.csv"]
+    #headers = ['id', 'name', 'description', 'image_link']
```
Gallaecio, line 8: Wow, things are more complicated that I foresaw… We should probably have the commented out code in these files also *blacked* somehow.

tests/test_commands.py
```diff
@@ -538,14 +538,14 @@ def test_url(self, url="test.com", domain="test.com"):
...
-                r"allowed_domains\s*=\s*\[\'(.+)\'\]",
+                r"allowed_domains\s*=\s*\[['\"](.+)['\"]\]",
             ).group(1),
         )
         self.assertEqual(
             f"http://{domain}/",
             self.find_in_file(
                 Path(self.proj_mod_path, "spiders", "test_name.py"),
-                r"start_urls\s*=\s*\[\'(.+)\'\]",
+                r"start_urls\s*=\s*\[['\"](.+)['\"]\]",
```
Gallaecio, line 548: Looks like something accidental happened with this file.

alexpdev: No that was done intentionally... The regex now will detect single and double quotes.  The test fails otherwise, because it only tests for single quotes.

scrapy/templates/spiders/csvfeed.tmpl
alexpdev: Okay  on it

tests/test_commands.py
Gallaecio: Oh, right!!! My bad.

scrapy/templates/project/module/settings.py.tmpl
```diff
@@ -38,32 +38,32 @@ ROBOTSTXT_OBEY = True
 
 # Override the default request headers:
 #DEFAULT_REQUEST_HEADERS = {
-#   'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
-#   'Accept-Language': 'en',
+#    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+#    "Accept-Language": "en",
```
Gallaecio, line 42: Nice indentation catch!

scrapy/templates/spiders/csvfeed.tmpl
```diff
@@ -2,19 +2,19 @@ from scrapy.spiders import CSVFeedSpider
...
+    name = "$name"
+    allowed_domains = ["$domain"]
+    start_urls = ["http://$domain/feed.csv"]
+    # headers = ["id", "name", "description", "image_link"]
+    # delimiter = "\t"
 
     # Do any adaptations you need here
     #def adapt_response(self, response):
-    #    return response
+    #     return response
```
Gallaecio, line 13: This indentation change seems wrong (4 spaces before, 5 now), but not a big issue, i.e. I don’t mind merging anyway.

Gallaecio (approved):
Awesome!

scrapy/templates/spiders/csvfeed.tmpl
Gallaecio: You know what, let me address it quickly, and let’s include this change into Scrapy 2.8.

PR #4844, opened 2020-10-15
improve the log configuration with using RotatingFileHandler.

Clannad1202:
Add new options about log so that users can set the size and backup count of log file.

Resolves #3628

PR #5815, opened 2023-01-30
feat(#3628): add rotating log handler to support log rotation

### LINKED ISSUE #3628: Crawler logs are cut by day

yushao2:
## Background
Opening this because PR #4844 is stale

Refer to #3628 for more context -- this PR adds the option to rotate logs based on size. Rotation can be configured using 2 settings:
- `LOG_MAX_BYTES` -- when configured, logs will automatically rotate when it approaches the set size (in bytes)
- `LOG_BACKUP_COUNT` -- number of backup / rotated logs to keep on disk. If not set and `LOG_MAX_BYTES` the logs will just be cleared when it approaches the set size.

Closes #3628

### DIFF
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -2,6 +2,7 @@ import logging
 import sys
 import warnings
 from logging.config import dictConfig
+from logging.handlers import RotatingFileHandler
 
 from twisted.python import log as twisted_log
 from twisted.python.failure import Failure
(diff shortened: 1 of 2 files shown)


scrapy/utils/log.py
```diff
@@ -130,7 +131,15 @@ def _get_handler(settings):
     if filename:
         mode = "a" if settings.getbool("LOG_FILE_APPEND") else "w"
         encoding = settings.get("LOG_ENCODING")
-        handler = logging.FileHandler(filename, mode=mode, encoding=encoding)
+        max_bytes = settings.get("LOG_MAX_BYTES", 0)
+        log_backup_count = settings.get("LOG_BACKUP_COUNT", 0)
```
wRAR, line 135: Both should use `getint` I think.

yushao2:
@wRAR I have no idea how to write unittests for the logging, so i pretty much copied from cpython's example. I hope that's fine
 https://github.com/python/cpython/blob/main/Lib/test/test_logging.py

yushao2:
seems like checks are failing on 3.7, give me some time while i set up a 3.7 env and run tests locally. it seems like manual trigger of workflows are needed given that i'm a new contributor so let me try to sort this out

wRAR:

Issue #5811, opened 2023-01-28
`BaseSettings.setdefault` does nothing


Prometheus3375:
### Description

Calling `setdefault` method of class `BaseSettings` does nothing.

### Steps to Reproduce

 ```python
 from scrapy.settings import BaseSettings
 settings = BaseSettings()
 stored = settings.setdefault('key', 'value')
 print(stored)  # prints None
 print(settings.copy_to_dict())  # prints empty dictionary
 ```

**Expected behavior:**
`settings.setdefault(key, default)` must work as described in `MutableMapping` interface: set `default` to `settings[key]` and return `default` if `key` is not present, otherwise return `settings[key]`.

**Actual behavior:**
`settings.setdefault(key, default)` does nothing regardless of holding `key` or not.

**Reproduces how often:** 100%

### Versions

Scrapy       : 2.7.1
lxml         : 4.8.0.0
libxml2      : 2.9.12
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 22.4.0
Python       : 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)]
pyOpenSSL    : 22.0.0 (OpenSSL 3.0.3 3 May 2022)
cryptography : 37.0.2
Platform     : Windows-10-10.0.19044-SP0

### Additional context

`BaseSettings` explicitly inherits from `MutableMapping` and does not redefine `setdefault` method. Thus, it uses base implementation:
```python
def setdefault(self, key, default=None):
    'D.setdefault(k[,d]) -> D.get(k,d), also set D[k]=d if k not in D'
    try:
        return self[key]
    except KeyError:
        self[key] = default
    return default
```
Base implementation refers to `self[key]` which is in fact `self.__getitem__[key]`. `BaseSettings` has own `__getitem__` implementation:
```python
def __getitem__(self, opt_name):
    if opt_name not in self:
        return None
    return self.attributes[opt_name].value
```
And here is the root of the problem: when passed `key` is not present, `__getitem__` returns `None`, and `setdefault` follows.

**Solution**
Implement own `setdefault` method. An example with matching signature:
```python
def setdefault(self, opt_name, default=None):
    if opt_name not in self:
        self.set(opt_name, default)
        return default
    return self.attributes[opt_name].value
```
`priority='project'` argument can be added although this changes signature.

[...]

PR #5821, opened 2023-02-01
Added `setdefault` to `BaseSettings`

jxlil:
Added the `setdefault` method to the `BaseSettings` class.

Now `setdefault` works as expected, if the _key_ does not exist in `BaseSettings` then this _key_ will be added and _default value_ will be assigned to it and this _value_ will be returned. For example:

```python
from scrapy.settings import BaseSettings

settings = BaseSettings()
# 'some_key' does not exist in BaseSettings, so it returns 'default_value'
print(settings.setdefault("some_key", "default_value"))

# 'some_key' now gets 'default_value'
print(settings.get("some_key"))
```

Closes https://github.com/scrapy/scrapy/issues/5811

### DIFF
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -293,6 +293,13 @@ class BaseSettings(MutableMapping):
         else:
             self.attributes[name].set(value, priority)
 
+    def setdefault(self, name, default=None, priority="project"):
+        if name not in self:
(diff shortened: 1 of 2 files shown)


wRAR (approved):

Gallaecio (approved):

Gallaecio:
Nice!

PR #5822, opened 2023-02-01
Update ip_address.py

aria1991:
- Import `os` module: To ensure that the script terminates cleanly after the reactor stops.

- Call to `os._exit(0`): To exit the script cleanly after the reactor stops, I added a call to `os._exit(0)` to exit the script with a status code of `0`, indicating a successful completion. This is added as a callback to the `d` object after the reactor stops!

### DIFF
--- a/tests/CrawlerRunner/ip_address.py
+++ b/tests/CrawlerRunner/ip_address.py
@@ -1,5 +1,5 @@
 from urllib.parse import urlparse
-
+import os
 from twisted.internet import reactor
 from twisted.names import cache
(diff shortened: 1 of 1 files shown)


Gallaecio:
Why is this necessary?

aria1991:
> Why is this necessary?

@Gallaecio  This is especially important when the script is run as a standalone script, as it ensures that the resources and processes associated with the script are properly cleaned up and terminated

wRAR:
It doesn't?

PR #5816, opened 2023-01-30
Add pre-commit hook for 'blacken-docs'

pankaj1707k:
`blacken-docs` has been added as a pre-commit hook in `.pre-commit-config.yaml`, however, it has not been run under this task.

Fixes #5813

PR #5823, opened 2023-02-02
Upgrade CI tools

Gallaecio:
Includes:
- Downgrading flake8 and isort for Python 3.7 support.
- Replacing tox environments for bandit, flake8, and black with one for pre-commit.
- Code changes due to upgrading black. Most notably, removing empty lines after `class`/`def` lines.


wRAR (approved):

Gallaecio:
Note: Whichever gets merged first, this or https://github.com/scrapy/scrapy/pull/5816/, requires the other pull request to be updated accordingly to have black versions match.

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = security,flake8,black,typing,py
+envlist = pre-commit,pylint,typing,py
```
kmike, line 7: Should we include pre-commit to github actions "checks" workflow? Any other changes needed to the workflow?

Gallaecio: We already run pre-commit in GitHub (without tox).

Since the pre-commit configuration freezes versions at the pre-commit level, and all other checks use tox, I don’t think we need anything else.

kmike:
Thanks @Gallaecio!

Issue #5831, opened 2023-02-16
Asyncio event loop handling results in hang


auxsvr:
### Description
If one attempts to use scrapy-playwright with scrapy shell, simply doing a `fetch` results in a hang.

### Steps to Reproduce

1. install scrapy-playwright,
2. launch scrapy shell with the `AsyncioSelectorReactor`,
3. run `fetch('https://www.google.com', meta={'playwright': True, 'playwright_include_page': True})`

**Expected behavior:** 
The request should be handled by scrapy-playwright and result in a response.

**Actual behavior:** 
The request is blocked soon after the second thread is created, right before the browser page is created. The traceback with Ctrl-C shows that execution is blocked in `blockingCallFromThread`. On closing the shell, the error message:

> asyncio: Task was destroyed but it is pending!
> task: <Task pending name='Task-163' coro=<ScrapyPlaywrightDownloadHandler._download_request() running at /home/petros/.local/lib/python3.10/site-packages/scrapy_playwright/handler.py:272> cb=[Deferred.fromFuture.<locals>.adapt() at /usr/lib/python3.10/site-packages/twisted/internet/defer.py:1063]>

appears.

**Reproduces how often:**
Always.

### Versions
Scrapy       : 2.8.0
lxml         : 4.9.2.0
libxml2      : 2.9.14
cssselect    : 1.2.0
parsel       : 1.7.0
w3lib        : 1.22.0
Twisted      : 22.10.0
Python       : 3.10.9 (main, Dec 08 2022, 14:49:06) [GCC]
pyOpenSSL    : 23.0.0 (OpenSSL 3.0.7 1 Nov 2022)
cryptography : 39.0.1
Platform     : Linux-6.1.8-1-default-x86_64-with-glibc2.36
### Additional context
There is a long thread about the root cause in https://discord.com/channels/851364676688543744/1073689007927590932. The summary is that `set_asyncio_event_loop` seems to have the following issues:

1. it uses `get_asyncio_event_loop_policy().get_event_loop()`, which results in the edge case described in https://github.com/python/cpython/issues/96377#issuecomment-1260006538, namely `set_asyncio_event_loop` creates an event loop, even though another already exists.
2. it uses `get_asyncio_event_loop_policy()`, which will probably be deprecated, according to https://github.com/python/cpython/issues/96377#issuecomment-1258819222.

PR #5832, opened 2023-02-16
Prevent edge case that creates extra event loop

auxsvr:
Fixes https://github.com/scrapy/scrapy/issues/5831.

`asyncio.get_event_loop_policy().get_event_loop() is asyncio.get_running_loop()` is not always true, as mentioned in https://github.com/python/cpython/issues/96377#issuecomment-1260006538. The AsyncioSelectorReactor runs in the second thread and uses the already initialised event loop in the main thread to run the crawler, so a single event loop will be running code from two threads.

According to https://github.com/python/cpython/issues/100160, we might need to replace `asyncio.get_event_loop()` with `asyncio.get_running_loop()` if semantics change. Currently, some low level aspects of asyncio are in flux.


Gallaecio:
I expect tests to break, we had to change this recently for a reason. Let’s see…

scrapy/utils/reactor.py
```diff
@@ -90,7 +90,6 @@ def _get_asyncio_event_loop():
 
 def set_asyncio_event_loop(event_loop_path):
     """Sets and returns the event loop with specified import path."""
-    policy = get_asyncio_event_loop_policy()
```
wRAR, line 93: Not calling this function will break Windows. This is actually visible in the tests.

wRAR (changes requested):

scrapy/utils/reactor.py
auxsvr: It should work with acb62ea595a5cf8f4ec5fd97c17f03150de9d331.

The main problem at the moment is that testing these changes is tricky, as I don't use Windows. The only relevant and easy unit test I can think of is to check if `set_asyncio_event_loop() is asyncio.get_running_loop()`, but this tests some semantics more than anything else.

scrapy/utils/reactor.py
```diff
@@ -54,7 +54,7 @@ def __call__(self):
         return self._func(*self._a, **self._kw)
 
 
-def get_asyncio_event_loop_policy():
+def set_asyncio_event_loop_policy():
```
Gallaecio, line 57: For backward compatibility reasons, we should keep the old function around, and log a deprecation warning when called.

Gallaecio (commented):
Any chance you can add a test that only passes after your changes? To make sure we do not break what you are fixing here later on.

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5833, opened 2023-02-17
Fix #960: S3 Feed Export throws boto error

### LINKED ISSUE #960: S3 Feed Export throws boto error: "Connection Reset By Peer"
Posted details on SO: http://stackoverflow.com/questions/27131693/scrapyd-s3-feed-export-connection-reset-by-peer

### LINKED ISSUE #5735: S3 backend can't handle uploads larger than 5GB
### Description

When feeds larger than 5GB are sent using AWS S3 backend, I'm receiving the follow error:

```bash
2022-11-24 18:45:31 [scrapy.extensions.feedexport] ERROR: Error storing csv feed (55 items) in: s3://scrapy-test/large_export.csv
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/twisted/python/threadpool.py", line 244, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/twisted/python/threadpool.py", line 260, in <lambda>
    inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/twisted/python/context.py", line 117, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/twisted/python/context.py", line 82, in callWithContext
    return func(*args, **kw)
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/scrapy/extensions/feedexport.py", line 196, in _store_in_thread
    self.s3_client.put_object(
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/ogabrielsantos/crawler-scrapy/venv/lib/python3.10/site-packages/botocore/client.py", line 960, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (EntityTooLarge) when calling the PutObject operation: Your proposed upload exceeds the maximum allowed size
```

### Steps to Reproduce

I've write a minimum exemple code to simulate this issue:

```python


class LargeExportSpider(Spider):
    name = "large_export"
    start_urls = ["http://news.ycombinator.com/"]
    custom_settings = {
        "FEEDS": {
            "s3://scrapy-test/large_export.csv": {
                "format": "csv",
            },
        },
    }

        text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque iaculis odio efficitur, ultricies"

        for _ in range(0, 55):  # creates a 5.3GB csv file
            yield {"name": "John Doe", "text": text * 1000000}

```

### Versions
`scrapy version --verbose`:
```bash
Scrapy       : 2.7.1
lxml         : 4.9.1.0
libxml2      : 2.9.13
cssselect    : 1.2.0
parsel       : 1.7.0
w3lib        : 2.0.1
Twisted      : 22.10.0
Python       : 3.10.8 (main, Oct 13 2022, 09:48:40) [Clang 14.0.0 (clang-1400.0.29.102)]
pyOpenSSL    : 22.1.0 (OpenSSL 3.0.5 5 Jul 2022)
cryptography : 38.0.1
Platform     : macOS-13.0.1-arm64-arm-64bit
```

`requirements.txt`:
```
botocore==1.29.16
Scrapy==2.7.1
```

### Additional context

Doing some investigation, I've seen that `S3FeedStorage` uses `put_object` which, as [per documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html), has a limit of 5GB per uploaded object:

https://github.com/scrapy/scrapy/blob/6ded3cf4cd134b615239babe28bb28c3ff524b05/scrapy/extensions/feedexport.py#L196

Looks like `boto3` [already have an upload method](https://boto3.amazonaws.com/v1/documentation/api/1.16.53/guide/s3-uploading-files.html) which handles multipart files, but scrapy relies on `botocore`.

jazzthief:
Fixes #960, closes #5735 by switching from `botocore` to `boto3` upload method, which supports multipart upload.

- Document the need to install boto3 for S3 support, rather than botocore.
- Implement code that uploads using boto3’s method instead of botocore to upload files. It seems the interfaces are similar.
- If boto3 is not installed, but botocore is, fall back to the current implementation, and log a deprecation warning.

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -106,7 +106,7 @@ The storages backends supported out of the box are:
 -   :ref:`topics-feed-storage-stdout`
 
 Some storage backends may be unavailable if the required external libraries are
-not available. For example, the S3 backend is only available if at least the botocore_
+not available. For example, the S3 backend is only available if the boto3_
 library is installed.
 
 
@@ -193,7 +193,7 @@ The feeds are stored on `Amazon S3`_.
 
     -   ``s3://aws_key:aws_secret@mybucket/path/to/export.csv``
 
--   Required external libraries: `boto3`_ >= 1.26.70, will fall back to botocore_ if unavailable
+-   Required external libraries: `boto3`_ >= 1.20.0
 
 The AWS credentials can be passed as user/password in the URI, or they can be
 passed through the following settings:
@@ -780,6 +780,5 @@ source spider in the feed URI:
 .. _URIs: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
 .. _Amazon S3: https://aws.amazon.com/s3/
 .. _boto3: https://github.com/boto/boto3
-.. _botocore: https://github.com/boto/botocore
 .. _Canned ACL: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
 .. _Google Cloud Storage: https://cloud.google.com/storage/


docs/topics/feed-exports.rst
```diff
@@ -101,12 +101,12 @@ The storages backends supported out of the box are:
...
 -   :ref:`topics-feed-storage-fs`
 -   :ref:`topics-feed-storage-ftp`
--   :ref:`topics-feed-storage-s3` (requires botocore_)
+-   :ref:`topics-feed-storage-s3` (requires boto3_)
 -   :ref:`topics-feed-storage-gcs` (requires `google-cloud-storage`_)
 -   :ref:`topics-feed-storage-stdout`
 
 Some storage backends may be unavailable if the required external libraries are
-not available. For example, the S3 backend is only available if the botocore_
+not available. For example, the S3 backend is only available if at least the botocore_
```
jazzthief, line 109: Not quite sure how to properly formulate the library requirements in the docs. Strictly speaking, `S3FeedStorage` won't work only there is no botocore, so unless we want to deprecate botocore altogether - we should probably mention the fallback thing in more detail.

docs/topics/feed-exports.rst
```diff
@@ -193,7 +193,7 @@ The feeds are stored on `Amazon S3`_.
 
     -   ``s3://aws_key:aws_secret@mybucket/path/to/export.csv``
 
--   Required external libraries: `botocore`_ >= 1.4.87
+-   Required external libraries: `boto3`_ >= 1.26.70, will fall back to botocore_ if unavailable
```
jazzthief, line 196: No particular reasoning behind the version, it's just what I have installed

jazzthief:
This works fine with `put_object()`. Switching to `upload_fileobj()` fails a number of tests - they probably need to be modified to support it.
Currently I'm thinking that existing tests should remain and cover only botocore's functionality, and I should add modified tests for boto3's functionality, possibly splitting them into single and multipart upload cases

jazzthief:
Fixed the call to botocore `get_session()` - now botocore functionality works as it should

jazzthief:
According to [this issue](https://github.com/boto/botocore/issues/974#issuecomment-229821575), `upload_file()` cannot be stubbed directly with botocore's `Stubber` that is currently used in the tests. I am guessing I can trace which exact methods `upload_file()` will call under the hood and stub them - or try a library like [placebo](https://github.com/garnaat/placebo) as botocore's maintainer suggested in the issue.
@Gallaecio, could you please have a look?

Gallaecio:
I will have a look, though it might take me some days :sweat:

Gallaecio (commented):
The change look good, I’ll try running the tests locally and see if I can figure out how to address the issues you mention.

jazzthief:
> The change look good, I’ll try running the tests locally and see if I can figure out how to address the issues you mention.

Got you, I will address your comments and wait for any thoughts you might have on tests

Gallaecio:
@jazzthief I see the `extra-deps-pinned` environment I added is causing issues.

If you can look into it while you work on my earlier feedback, that would be great. But I can take a look myself as well if you prefer.

jazzthief:
> If you can look into it while you work on my earlier feedback, that would be great. But I can take a look myself as well if you prefer.

I'll have a look. One question, though: I see `is_boto3_available()` used in tests - should I still switch everything to an `IS_BOTO3_AVAILABLE` variable inside `feedexport.py` or do we leave it like that?

Gallaecio:
I think it may be worth it switching to a variable, I just did not want to make it as part of my changes, and instead focus on the test part.

jazzthief:
@Gallaecio I addressed your comments.

Regarding `extra-deps-pinned`: from what I understood from boto3 changelog, the API for upload was completely implemented in 1.1.4, however it seems something was changed later, so I've put 1.20.0 into `extra-deps-pinned` config as well as the docs. This passed the tests locally.

Gallaecio (approved):
Seems to me like this is ready for review. Great job!

jazzthief:
> Seems to me like this is ready for review. Great job!

Thanks for your help! Do you think we should create an issue on reworking boto3 tests?

Gallaecio:
> reworking boto3 tests

Could you elaborate?

As far as I am concerned, test-wise we are covered here. While something in line with Stub would be great, I think mocking on our side just to make sure that we are mapping parameters correctly to what boto3 expects is a good approach, and not that different from stubbing.

jazzthief:
> I think mocking on our side just to make sure that we are mapping parameters correctly to what boto3 expects is a good approach, and not that different from stubbing.

That's what I meant. Got you, great then.

wRAR (approved):

wRAR:
Thanks!

PR #5784, opened 2023-01-06
Added Documentation for Spider.update_settings

tuhinmallick:
I added documentation for Spider.update_settings in relation to the issue: #5745 
My contribution included a detailed explanation and a code example.

PR #5846, opened 2023-03-08
Added documentation for Spider.update_settings

### LINKED ISSUE #5745: Document Spider.update_settings

jxlil:
Added documentation for the `Spider.update_settings` class method.

Fixes #5745, closes #5757, closes #5784.

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -66,8 +66,8 @@ Example::
 ----------------------
 
 Spiders (See the :ref:`topics-spiders` chapter for reference) can define their
-own settings that will take precedence and override the project ones. They can
-do so by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
+own settings that will take precedence and override the project ones. One way
+to do so is by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
 
 .. code-block:: python
 
@@ -80,6 +80,22 @@ do so by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
         custom_settings = {
             "SOME_SETTING": "some value",
         }
+
+It's often better to provide a :meth:`~scrapy.Spider.update_settings` instead,
+and settings set there should use the "spider" priority explicitly:
+
+.. code-block:: python
+
+    import scrapy
+
+
+    class MySpider(scrapy.Spider):
+        name = "myspider"
+
+        @classmethod
+        def update_settings(cls, settings):
+            settings.set("SOME_SETTING", "some value", priority="spider")
+            super().update_settings(settings)
 
 3. Project settings module
 --------------------------
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -81,7 +81,7 @@ to do so is by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
             "SOME_SETTING": "some value",
         }
 
-It's often better to provide a :meth:`~scrapy.Spider.update_settings` instead,
+It's often better to implement :meth:`~scrapy.Spider.update_settings` instead,
 and settings set there should use the "spider" priority explicitly:
 
 .. code-block:: python
@@ -94,8 +94,8 @@ and settings set there should use the "spider" priority explicitly:
 
         @classmethod
         def update_settings(cls, settings):
-            settings.set("SOME_SETTING", "some value", priority="spider")
             super().update_settings(settings)
+            settings.set("SOME_SETTING", "some value", priority="spider")
 
 3. Project settings module
 --------------------------
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -162,7 +162,7 @@ scrapy.Spider
        settings, spider attributes or other factors and use setting priorities
        other than ``'spider'``. Also, it's easy to extend ``update_settings()``
        in a subclass by overriding it, while doing the same with
-       :attr:`~scrapy.Spider.custom_settings` is hard or impossible.
+       :attr:`~scrapy.Spider.custom_settings` can be hard.
 
        For example, suppose a spider needs to modify :setting:`FEEDS`:
 
@@ -182,8 +182,8 @@ scrapy.Spider
 
                @classmethod
                def update_settings(cls, settings):
-                   settings.setdefault("FEEDS", {}).update(cls.custom_feed)
                    super().update_settings(settings)
+                   settings.setdefault("FEEDS", {}).update(cls.custom_feed)
 
    .. method:: start_requests()
 
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -179,8 +179,6 @@ scrapy.Spider
                    settings.setdefault("FEEDS", {}).update(cls.custom_feed)
                    super().update_settings(settings)
 
-
-
    .. method:: start_requests()
 
        This method must return an iterable with the first Requests to crawl for
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -151,7 +151,7 @@ scrapy.Spider
        and is called during initialization of a spider instance.
 
        It takes a :class:`~scrapy.settings.Settings` object as a parameter and
-       can add or updates the spider's configuration values. This method is a class method,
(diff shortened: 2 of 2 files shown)


Gallaecio (commented):
Besides my points below, I think it may be good to highlight the benefits of update_settings over custom_settings. I‘m thinking that the main one is that you can dynamically add, remove or change settings based on spider arguments and other settings that may be defined outside code (e.g. Scrapy Cloud settings), not sure if there are others.

docs/topics/spiders.rst
```diff
@@ -145,6 +145,41 @@ scrapy.Spider
...
+       and can be called during initialization of a spider instance.
+
+       It takes a :class:`~scrapy.settings.Settings` object as a parameter and
+       adds or updates the spider's configuration values. This method is a class method,
+       meaning that it is called on the :class:`~scrapy.Spider` class and allows all instances
+       of the spider to share the same configuration.
+
+       To create class hierarchies for spiders, it is recommended to use the :attr:`custom_settings`
+       attribute instead of ``update_settings()``, as it allows for default settings to be
+       defined and automatically inherited by subclasses.
```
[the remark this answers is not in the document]
jxlil: You're right that both methods are inherited similarly. To avoid confusion, I'll remove that paragraph and add a new one that highlights the benefits of using `update_settings()`

jxlil:
Done. The requested modifications were made and the last paragraph was replaced to highlight the benefits of `update_settings()`

docs/topics/spiders.rst
```diff
@@ -145,6 +145,41 @@ scrapy.Spider
...
+       and is called during initialization of a spider instance.
+
+       It takes a :class:`~scrapy.settings.Settings` object as a parameter and
+       can add or updates the spider's configuration values. This method is a class method,
+       meaning that it is called on the :class:`~scrapy.Spider` class and allows all instances
+       of the spider to share the same configuration.
+
+       One of the main advantages of ``update_settings()`` is that it allows
+       you to dynamically add, remove or change settings based on spider arguments
+       or other external factors.
```
Gallaecio, line 160: I just realized I got it wrong here, you cannot take spider arguments into account, because those are set in the `__init__`, and as you say here this is a class method. You can still edit settings based on other settings or external factors, though, so that part is indeed true.

docs/topics/spiders.rst
```diff
@@ -145,6 +145,41 @@ scrapy.Spider
...
+
+       For example, suppose a spider needs update :setting:`FEEDS`:
```
Gallaecio, line 162:
```suggestion
       For example, suppose a spider needs to modify :setting:`FEEDS`:
```

Besides the missing “to” typo, I think “modify” here (or “edit”, or “change”) is a slightly better word choice here, because I think the point is that, unlike `custom_settings` that only let you completely override a setting value from e.g. the command line, here you can keep parts of the command-line value while changing other parts.

docs/topics/spiders.rst
jxlil: Done. The last commit updated this part

docs/topics/spiders.rst
```diff
@@ -145,6 +145,41 @@ scrapy.Spider
        :param kwargs: keyword arguments passed to the :meth:`__init__` method
        :type kwargs: dict
 
+   .. classmethod:: update_settings(settings)
+
+       The ``update_settings()`` method is used to modify the spider's settings
+       and is called during initialization of a spider instance.
+
+       It takes a :class:`~scrapy.settings.Settings` object as a parameter and
+       can add or updates the spider's configuration values. This method is a class method,
```
Gallaecio, line 154:
```suggestion
       can add or update the spider's configuration values. This method is a class method,
```

Gallaecio (approved):

docs/topics/spiders.rst
```diff
@@ -145,6 +145,41 @@ scrapy.Spider
...
+                   "/home/user/documents/items.json": {
+                       "format": "json",
+                       "indent": 4,
+                   }
+               }
+
+               @classmethod
+               def update_settings(cls, settings):
+                   settings.setdefault("FEEDS", {}).update(cls.custom_feed)
+                   super().update_settings(settings)
```
gutsytechster, line 181: Just want to confirm that wouldn't call to the `super().update_settings` be the first thing to do and then override the required settings here.

Gallaecio: I think that’s what you would usually want, yes, but not necessarily. For example, if you define `update_settings` in a base spider class, and a child spider class defined `custom_setting`, you may want that `custom_settings` to be able to override what the base class does in `update_settings`, in which case the code here is what you want.

wRAR:
It doesn't cover everything that I want to be covered here, I'll try to suggest some additional wording.

docs/topics/settings.rst
```diff
@@ -81,6 +81,22 @@ do so by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
             "SOME_SETTING": "some value",
         }
 
+It's often better to provide a :meth:`~scrapy.Spider.update_settings` instead,
+and settings set there should use the "spider" priority explicitly:
```
Gallaecio, line 85:
```suggestion
It's often better to implement :meth:`~scrapy.Spider.update_settings` instead,
and settings set there should use the "spider" priority explicitly:
```

or

```suggestion
It's often better to provide a :meth:`~scrapy.Spider.update_settings` method 
instead, and settings set there should use the "spider" priority explicitly:
```

docs/topics/settings.rst
```diff
@@ -81,6 +81,22 @@ do so by setting their :attr:`~scrapy.Spider.custom_settings` attribute:
...
+    import scrapy
+
+
+    class MySpider(scrapy.Spider):
+        name = "myspider"
+
+        @classmethod
+        def update_settings(cls, settings):
+            settings.set("SOME_SETTING", "some value", priority="spider")
+            super().update_settings(settings)
```
Gallaecio, line 98: No strong opinion, since `priority` will handle most cases, but I do wonder if reversing these lines may be a better default suggestion, even if for something as trivial as keeping the `super()` call closer to the method definition start, to make it more clear that it is being called in cases where lots of settings are defined (i.e. lots of lines before `super` if we leave `super` for last).

```suggestion
            super().update_settings(settings)
            settings.set("SOME_SETTING", "some value", priority="spider")
```

Which also applies to the later example, where to a [similar suggestion](https://github.com/scrapy/scrapy/pull/5846/files#r1230629675) I responded “does not matter” :sweat_smile: .

docs/topics/spiders.rst
```diff
@@ -145,6 +145,46 @@ scrapy.Spider
...
+       class and allows all instances of the spider to share the same
+       configuration.
+
+       While per-spider settings can be set in
+       :attr:`~scrapy.Spider.custom_settings`, using ``update_settings()``
+       allows you to dynamically add, remove or change settings based on other
+       settings, spider attributes or other factors and use setting priorities
+       other than ``'spider'``. Also, it's easy to extend ``update_settings()``
+       in a subclass by overriding it, while doing the same with
+       :attr:`~scrapy.Spider.custom_settings` is hard or impossible.
```
Gallaecio, line 165: No strong opinion here either, but “impossible” feels like too strong a word.

```suggestion
       :attr:`~scrapy.Spider.custom_settings` can be hard.
```

Gallaecio (approved):

docs/topics/spiders.rst
```diff
@@ -145,6 +145,46 @@ scrapy.Spider
...
+                   "/home/user/documents/items.json": {
+                       "format": "json",
+                       "indent": 4,
+                   }
+               }
+
+               @classmethod
+               def update_settings(cls, settings):
+                   settings.setdefault("FEEDS", {}).update(cls.custom_feed)
+                   super().update_settings(settings)
```
wRAR, line 186:
```suggestion
                   super().update_settings(settings)
                   settings.setdefault("FEEDS", {}).update(cls.custom_feed)
```

wRAR:
Thanks!

Issue #4621, opened 2020-06-10
Storage.store is called only for the first feed when empty


StasDeep:
### Description

When having `FEED_STORE_EMPTY=False` and multiple feeds, and the output is empty, `storage.store` is called until first empty feed slot is encountered.

### Steps to Reproduce

1. Create a `test.py` file with [this spider](https://gist.github.com/StasDeep/c7f0b775eb4897b1012f2c76ce6296f5)
2. `scrapy runspider test.py -L INFO`

**Expected behavior:**

```
2020-06-10 10:50:27 [test] INFO: Storing in thread: gs://bucket/output.json
2020-06-10 10:50:27 [test] INFO: Storing in thread: gs://bucket/output.csv
```

**Actual behavior:**

```
2020-06-10 10:50:27 [test] INFO: Storing in thread: gs://bucket/output.json
```

**Reproduces how often:** always

### Versions

```
Scrapy       : 2.1.0
lxml         : 4.5.1.0
libxml2      : 2.9.10
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 20.3.0
Python       : 3.7.4 (default, Sep  4 2019, 15:20:53) - [Clang 10.0.0 (clang-1000.10.44.4)]
pyOpenSSL    : 19.1.0 (OpenSSL 1.1.1g  21 Apr 2020)
cryptography : 2.9.2
Platform     : Darwin-19.4.0-x86_64-i386-64bit
```

### Reason

In `FeedExporter.close_spider`, it **returns** a deferred from inside a loop instead of adding it to `deferred_list`

```
        deferred_list = []
        for slot in self.slots:
            if not slot.itemcount and not slot.store_empty:
                # We need to call slot.storage.store nonetheless to get the file
                # properly closed.
                return defer.maybeDeferred(slot.storage.store, slot.file)
```

StasDeep:
Will provide a fix for this.

Gallaecio:
Great catch!

PR #5847, opened 2023-03-08
Fix FeedExporter not to export empty file

### LINKED ISSUE #872: FileFeedStorage creates empty file when no items are scraped

### LINKED ISSUE #5633: Address BatchDeliveriesTest.test_batch_path_differ failures on Windows

namelessGonbai:
Fixes #872, fixes #5633 
+ Fix FeedExporter not to export empty file
+ Change default value of FEED_STORE_EMPTY

There are several problems with this patch.

1. There is no additional test.

The explanation for the FeedExporter fix is within the scope of what is already done in the existing tests, so we believe that the fixes to the existing tests are sufficient for testing.
However, I have no idea how to test for changes in defaults, even though I think the magnitude of the effect is clearly significant and needs to be tested.

2. Meaning of `test_export_no_items_multiple_feeds()`

This test from #4621.
This test seems to be testing whether a file descriptor leak has been resolved when `FEED_STORE_EMPTY=False`.
`FEED_STORE_EMPTY=False` will no longer open the file if the item is missing and this test is no longer meaningful.
This test needs to be fixed (or deleted) because I made a fix that twisted its original meaning to pass the test.


namelessGonbai:
I forgot to mention this as well. Another issue with this patch is whether changing the default values is acceptable.

namelessGonbai:
I forgot to state this explicitly.
I would like a review of the problems I listed above.

scrapy/settings/default_settings.py
```diff
@@ -141,7 +141,7 @@
 FEED_TEMPDIR = None
 FEEDS = {}
 FEED_URI_PARAMS = None  # a function to extend uri arguments
-FEED_STORE_EMPTY = False
+FEED_STORE_EMPTY = True
```
Gallaecio, line 144: We need to update [the docs](https://github.com/scrapy/scrapy/blob/d60b4edd11436e61284615ec7ce89f8ac7e46d9a/docs/topics/feed-exports.rst?plain=1#L555) accordingly.

Gallaecio (commented):
> I have no idea how to test for changes in defaults, even though I think the magnitude of the effect is clearly significant and needs to be tested.

I don’t think we need a specific test for defaults, I think your test changes already cover the default value change as far as we need.

> This test needs to be fixed (or deleted) because I made a fix that twisted its original meaning to pass the test.

I agree, and I think changing expectations is the way to go:

```diff
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -1057,13 +1057,13 @@ class FeedExportTest(FeedExportTestBase):
                 self._random_temp_filename(): {"format": "csv"},
             },
             "FEED_STORAGES": {"file": LogOnStoreFileStorage},
+            "FEED_STORE_EMPTY": False,
         }
 
         with LogCapture() as log:
             yield self.exported_no_data(settings)
 
-        print(log)
-        self.assertEqual(str(log).count("Storage.store is called"), 3)
+        self.assertEqual(str(log).count("Storage.store is called"), 0)
 
     @defer.inlineCallbacks
     def test_export_multiple_item_classes(self):
```

scrapy/settings/default_settings.py
namelessGonbai: I forgot about that.
Other than changing the default values, do I need to make any additions to the documentation with this modification? For example, mentioning that files are not exported explicitly, etc.

Gallaecio: To mention that files are not exported may be a good thing, yes.

I don’t think we need anything else, though. We will need to justify this default value change, explain the reasons behind it, in the release notes, but we usually handle release notes entries when preparing the release, so you do not need to worry about that part here.

namelessGonbai: Understood. Thank you very much.

docs/topics/feed-exports.rst
```diff
@@ -552,9 +552,10 @@ to ``.json`` or ``.xml``.
 FEED_STORE_EMPTY
 ----------------
 
-Default: ``False``
+Default: ``True``
 
 Whether to export empty feeds (i.e. feeds with no items).
+If False and there is no items, no new files are created and existing files are not modified.
```
Gallaecio, line 558:
```suggestion
If ``False``, and there are no items to export, no new files are created and 
existing files are not modified, even if the :ref:`overwrite feed option 
<feed-options>` is enabled.
```

namelessGonbai:
Findings.
The test had a potential bug.
https://github.com/scrapy/scrapy/blob/6ab49e954f25d491df7986065d270bb0068c7c89/tests/test_feedexport.py#L2500-L2508
This will overwrite the same file when the file write is fast enough and the batch_time matches.

namelessGonbai:
I can rewrite the whole test to %(batch_id)d if you need to fix it at least, and I'm sure it's the type of error that disappears when you re-run the runner in the first place, though.

Gallaecio:
I think I have seen that test fail before, I would not worry about it here, fixing it is out of the scope of this change.

Gallaecio (approved):

namelessGonbai:
No merge or additional review yet, any problems?

scrapy/extensions/feedexport.py
```diff
@@ -277,18 +277,21 @@ def _store_in_thread(self, file):
 class FeedSlot:
     def __init__(
         self,
-        file,
-        exporter,
```
wRAR, line 281: This now changes a public class, as we made FeedSlot public recently. @kmike @Gallaecio thoughts?

Gallaecio: The only backward-compatible option I can think of at the moment is to keep the current class unchanged, deprecate it, and create a new class with the new API with a new class name (`BatchSlot`?).

wRAR: We decided that as this only changes the `__init__` signature and the current usage of this class is to receive it in the `feed_slot_closed` signal handler, it's fine to just change this.

wRAR (approved):

Gallaecio (approved):

kmike:
Thanks @namelessGonbai, @wRAR and @Gallaecio!

PR #5848, opened 2023-03-12
Adding Spanish language to README

nerjimado:
I have put the same information below the README.md but in Spanish so that it can reach more users.


Gallaecio:
I think we need to handle this differently, [on the docs side](https://github.com/scrapy/scrapy/issues/3511), not on the README.

PR #5849, opened 2023-03-12
Replace deprecated BadZipfile with BadZipFile

hugovk:
`BadZipfile` (with a small `f`) has been deprecated since Python 3.2, use `BadZipFile` (big `F`) instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/issues/86437


wRAR (approved):

Gallaecio:
Thanks!

PR #5852, opened 2023-03-16
add sticky-meta-params

### LINKED ISSUE #3645: Sticky request/response meta

felipeboffnunes:
Extracting the sticky-meta-params from @heylouiz [repo](https://github.com/heylouiz/scrapy-sticky-meta-params)
Already closed #3645 as the docs already cover, but we can integrate the functionality directly inside scrapy now.

It still requires some adjustments on documentation, certainly. This is mostly to get a clear slate so we can close everything related to this.


heylouiz:
Hey @felipeboffnunes, there is also this open PR implementing this: #4141

felipeboffnunes:
Thank god. Don't know how I didnt see this one on the other ones though. Closing this one!

Gallaecio:
I forgot about it as well :facepalm:

PR #5856, opened 2023-03-24
add parse_with_rules method and log a warning when _parse_response is…

DevInTheShell-BR:
… called directly or when it is overriden. 

This is a working in progress. Although the code is working when I tested it manually, it lacks automated tests. I don't  know yet how to make the test work with async methods.

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -6,12 +6,14 @@ See documentation in docs/topics/spiders.rst
 """
 
 import copy
+import warnings
 from typing import AsyncIterable, Awaitable, Sequence
 
 from scrapy.http import HtmlResponse, Request, Response
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import Spider
 from scrapy.utils.asyncgen import collect_asyncgen
+from scrapy.utils.deprecate import method_is_overridden
 from scrapy.utils.spider import iterate_spider_output
 
 
@@ -67,7 +69,15 @@ class CrawlSpider(Spider):
         self._compile_rules()
 
     def _parse(self, response, **kwargs):
-        return self._parse_response(
+        cls = self.__class__
+        is_overriden = method_is_overridden(cls, CrawlSpider, "_parse_response")
+        if is_overriden:
+            warnings.warn(
+                "CrawlSpider._parse_response method is deprecated: "
+                "it will be removed in future Scrapy releases."
+                "Please use CrawlSpider.parse_with_rules method instead. "
+            )
+        return self.parse_with_rules(
             response=response,
             callback=self.parse_start_url,
             cb_kwargs=kwargs,
@@ -105,7 +115,7 @@ class CrawlSpider(Spider):
 
     def _callback(self, response, **cb_kwargs):
         rule = self._rules[response.meta["rule"]]
-        return self._parse_response(
+        return self.parse_with_rules(
             response, rule.callback, {**rule.cb_kwargs, **cb_kwargs}, rule.follow
         )
 
@@ -113,7 +123,22 @@ class CrawlSpider(Spider):
         rule = self._rules[failure.request.meta["rule"]]
         return self._handle_failure(failure, rule.errback)
 
-    async def _parse_response(self, response, callback, cb_kwargs, follow=True):
+    def parse_with_rules(self, response, callback, cb_kwargs, follow=True):
+        direct_call = True
+        return self._parse_response(response, callback, cb_kwargs, follow, direct_call)
+
+    async def _parse_response(
+        self, response, callback, cb_kwargs, follow=True, direct_call=False
+    ):
+        cls = self.__class__
+        is_overriden = method_is_overridden(cls, CrawlSpider, "_parse_response")
+        if not direct_call or is_overriden:
+            warnings.warn(
+                "CrawlSpider._parse_response method is deprecated: "
+                "it will be removed in future Scrapy releases."
+                "Please use CrawlSpider.parse_with_rules method instead. "
+            )
+
         if callback:
             cb_res = callback(response, **cb_kwargs) or ()
             if isinstance(cb_res, AsyncIterable):


scrapy/spiders/crawl.py
```diff
@@ -67,7 +69,15 @@ def __init__(self, *a, **kw):
...
     def _parse(self, response, **kwargs):
-        return self._parse_response(
+        cls = self.__class__
+        is_overriden = method_is_overridden(cls, CrawlSpider, "_parse_response")
+        if is_overriden:
+            warnings.warn(
+                "CrawlSpider._parse_response method is deprecated: "
+                "it will be removed in future Scrapy releases."
+                "Please use CrawlSpider.parse_with_rules method instead. "
+            )
```
Gallaecio, line 79:
```suggestion
            warnings.warn(
                "The CrawlSpider._parse_response method is deprecated: it "
                "will be removed in future Scrapy releases. Please use the "
                "CrawlSpider.parse_with_rules method instead."
            )
```

scrapy/spiders/crawl.py
```diff
@@ -105,15 +115,30 @@ def _requests_to_follow(self, response):
...
         )
 
     def _errback(self, failure):
         rule = self._rules[failure.request.meta["rule"]]
         return self._handle_failure(failure, rule.errback)
 
-    async def _parse_response(self, response, callback, cb_kwargs, follow=True):
+    def parse_with_rules(self, response, callback, cb_kwargs, follow=True):
+        direct_call = True
+        return self._parse_response(response, callback, cb_kwargs, follow, direct_call)
```
Gallaecio, line 128: This will break in cases where `_parse_response` is overridden, because the overridden method will not expect an additional parameter (`direct_call`).

DevInTheShell-BR: You are right. I did not think about that case. I fixed it in the last commit passing `direct_call` in `cb_kwargs ` without changing the original method signature.

scrapy/spiders/crawl.py
```diff
@@ -105,15 +115,27 @@ def _requests_to_follow(self, response):
...
+        return self._parse_response(response, callback, cb_kwargs, follow)
+
     async def _parse_response(self, response, callback, cb_kwargs, follow=True):
+        if cb_kwargs.get("direct_call", True):
+            warnings.warn(
+                "The CrawlSpider._parse_response method is deprecated: it "
+                "will be removed in future Scrapy releases. Please use the "
+                "CrawlSpider.parse_with_rules method instead."
+            )
+        cb_kwargs.pop("direct_call", None)
```
Gallaecio, line 137: Interesting strategy. I think it might be OK to go this route, as long as we use a more descriptive and less public name, e.g. `_direct_parse_response_call`.

However, there may be better approaches here.

For example, you could define a new private method, move the body of `_parse_response` to that method, and call that method from both `_parse_method` and `parse_with_rules`. In `_parse_method`, however, you also log a warning. Then you do not need to mess up with cb_kwargs, which is meant to end users, not for internal handlings.

It might also be good to log these warnings only once, and not once per callback call. So, you could run the method inheritance check on `__init__`, and define a private boolean attribute in `__init__` as well to keep track of whether or not `_parse_response` has been called, to only log a warning on the first of such calls.

Finally, I think the `parse_with_rules` should be async as well. If you tried that already and had issues, mind that I think you need to await your internal call to `_parse_response` (or to the new private method) for things to work as expected. That is, rather than:

```python
    async def parse_with_rules(…):
        return self._parse_response(…)
```

I think you need:

```python
    async def parse_with_rules(…):
        return await self._parse_response(…)
```

DevInTheShell-BR: I took the approach that you suggested  but I was not able to make `parse_with_rules` and `_parse_response` async. It raises this error:
```
return await self._parse_with_rules(response, callback, cb_kwargs, follow)
TypeError: object async_generator can't be used in 'await' expression
```
if I put `async/await` in the methods. I think it is because of yield on the body of the original function. It is working with only the `_parse_with_rules` being async but I don't know how much the others functions not being async would affect the operation of other parts of the code. 

I also created 3 tests to check if the warnings would be raising properly. Should I add more to test other stuff?

wRAR: `_parse_with_rules` is an async generator so you cannot await it, you need:

```python
def _parse_response(self, response, callback, cb_kwargs, follow=True):
    ...
    async for request_or_item in self._parse_with_rules(response, callback, cb_kwargs, follow):
        yield request_or_item
```

scrapy/spiders/crawl.py
```diff
@@ -65,9 +67,19 @@ class CrawlSpider(Spider):
...
         self._compile_rules()
+        cls = self.__class__
+        self._is_warned = False
+        is_overriden = method_is_overridden(cls, CrawlSpider, "_parse_response")
+        if is_overriden:
+            warnings.warn(
+                "The CrawlSpider._parse_response method is deprecated: it "
+                "will be removed in future Scrapy releases. Please use the "
+                "CrawlSpider.parse_with_rules method instead."
+            )
```
Gallaecio, line 78: What about mentioning where the problem is? Users may not realize they are subclassing CrawlSpider, or where, or they may be using a Scrapy plugin, or subclassing a CrawlSpider subclass from a third-party, etc.

```suggestion
            cls_path = f"{cls.__module__}.{cls.__qualname__}"
            warnings.warn(
                f"The CrawlSpider._parse_response method, that the "
                f"{cls_path} class reimplements, is deprecated: it will be "
                f"removed in future Scrapy releases. Please use the "
                f"CrawlSpider.parse_with_rules method instead."
            )
```

scrapy/spiders/crawl.py
```diff
@@ -65,9 +67,19 @@ class CrawlSpider(Spider):
     def __init__(self, *a, **kw):
         super().__init__(*a, **kw)
         self._compile_rules()
+        cls = self.__class__
+        self._is_warned = False
+        is_overriden = method_is_overridden(cls, CrawlSpider, "_parse_response")
+        if is_overriden:
```
Gallaecio, line 73: We should maintain backward compatibility: if they indeed override `_parse_response`, we need to call it where needed

I think something like this may work:

```suggestion
        if is_overriden:
            self.parse_with_rules = self._parse_response
```

scrapy/spiders/crawl.py
```diff
@@ -128,6 +140,19 @@ async def _parse_response(self, response, callback, cb_kwargs, follow=True):
             for request_or_item in self._requests_to_follow(response):
                 yield request_or_item
 
+    def parse_with_rules(self, response, callback, cb_kwargs, follow=True):
+        return self._parse_with_rules(response, callback, cb_kwargs, follow)
+
+    def _parse_response(self, response, callback, cb_kwargs, follow=True):
+        if not self._is_warned:
+            self._is_warned = True
```
Gallaecio, line 148: I think we need to distinguish warning from overriding and warning from use.

If you override, you get 1 warning. If you use, you get 1 warning. If you do both, you get 2 warnings.

PR #5858, opened 2023-03-24
Don't call SSL_get_server_tmp_key() if not available.

### LINKED ISSUE #5857: TLS logging broken with new cryptography
https://github.com/pyca/cryptography/pull/8391 dropped `SSL_get_server_tmp_key()` so we need to disable the code that uses it if it's not available.

wRAR:
Fixes #5857


Laerte (approved):
🚀

PR #5876, opened 2023-03-27
Add feed_slot_closed and feed_exporter_closed signals to work with FeedExporter

guillermo-bondonno:
The signals will be sent when a slot and the FeedExporter are closed respectively. 

This allows for the use of extensions related to feeds (like email feeds or slack notifications) that may require them to be used without the need for a custom FeedExporter.

### DIFF
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -322,7 +322,7 @@ feed_slot_closed
 
 
 feed_exporter_closed
-~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~
 
 .. signal:: feed_exporter_closed
 .. function:: feed_exporter_closed()
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -318,7 +318,7 @@ feed_slot_closed
     This signal supports returning deferreds from its handlers.
 
     :param slot: the slot closed
-    :type failure: scrapy.extensions.feedexport.FeedSlot
+    :type slot: scrapy.extensions.feedexport.FeedSlot
 
 
 feed_exporter_closed
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -308,26 +308,28 @@ spider_error
     :type spider: :class:`~scrapy.Spider` object
 
 feed_slot_closed
-~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~
 
 .. signal:: feed_slot_closed
 .. function:: feed_slot_closed(slot)
 
-    Sent when a slot is closed in the feed exporter.
+    Sent when a :ref:`feed exports <topics-feed-exports>` slot is closed.
 
-    This signal does not support returning deferreds from its handlers.
+    This signal supports returning deferreds from its handlers.
 
     :param slot: the slot closed
     :type failure: scrapy.extensions.feedexport._FeedSlot
 
 
 feed_exporter_closed
-~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~
 
 .. signal:: feed_exporter_closed
 .. function:: feed_exporter_closed()
 
-    Sent when the feed exporter is closed.
+    Sent when the :ref:`feed exports <topics-feed-exports>` extension is closed,
+    during the handling of the :signal:`spider_closed` signal by the extension,
+    after all feed exporting has been handled.
 
     This signal supports returning deferreds from its handlers.
 
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -307,6 +307,31 @@ spider_error
     :param spider: the spider which raised the exception
     :type spider: :class:`~scrapy.Spider` object
 
+feed_slot_closed
+~~~~~~~~~~~~
+
+.. signal:: feed_slot_closed
+.. function:: feed_slot_closed(slot)
+
+    Sent when a slot is closed in the feed exporter.
+
+    This signal does not support returning deferreds from its handlers.
+
+    :param slot: the slot closed
+    :type failure: scrapy.extensions.feedexport._FeedSlot
+
+
+feed_exporter_closed
+~~~~~~~~~~~~
+
+.. signal:: feed_exporter_closed
+.. function:: feed_exporter_closed()
+
+    Sent when the feed exporter is closed.
+
+    This signal supports returning deferreds from its handlers.
+
+
 Request signals
 ---------------
 
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -318,7 +318,7 @@ feed_slot_closed
     This signal supports returning deferreds from its handlers.
 
     :param slot: the slot closed
-    :type failure: scrapy.extensions.feedexport._FeedSlot
+    :type failure: scrapy.extensions.feedexport.FeedSlot
 
 
 feed_exporter_closed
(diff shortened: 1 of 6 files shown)


Gallaecio (commented):
Why not keep the original `deferred_list` list, store `DeferredList(deferred_list)` in a variable, and add the feed_exporter_closed signal as a callback to that `DeferredList(deferred_list)`?

VMRuiz:
> Why not keep the original `deferred_list` list, store `DeferredList(deferred_list)` in a variable, and add the feed_exporter_closed signal as a callback to that `DeferredList(deferred_list)`?

Previously, the code assumed that only the slots in `self.slots` needed to be closed when `close_spider` was called. However, with the new async approach, there may be pending processes that were started but have not yet finished closing, and thus have not raised the close feed signal yet as expected. To handle this, we register these pending processes in `pending_deferreds`, so they can be awaited if needed during the closing process.

By getting rid of the previous approach and relying only on the new pending_deferreds, we can refactor all this new logic inside the _close_slot method, as shown in the code here:


https://github.com/scrapy/scrapy/blob/a0b7ea95e854cf240aadf71299f8c3dae6fcda41/scrapy/extensions/feedexport.py#L411-L417

The current commit centralices all the new logic inside `_close_slot` and simplifies the awaiting code in the `close_spider` method. The alternative would be something like this: 

```
async def close_spider(self, spider):
    deferred_list = []
    for slot in self.slots:
        d = self._close_slot(slot, spider)
        deferred_list.append(d)
        
    # Await previous pending deferreds + current deferreds
    deferred_list += self.pending_deferreds
    if deferred_list:
        await maybe_deferred_to_future(DeferredList(deferred_list))

    # Send FEED_EXPORTER_CLOSED signal
    await maybe_deferred_to_future(
        self.crawler.signals.send_catch_log_deferred(signals.feed_exporter_closed)
    )
    
    [...]
    d = self._close_slot(slot, spider)
    self.pending_deferreds.append(d)
    d.addBoth(lambda _: self.pending_deferreds.remove(d))
    [...]
    
    [...]
    d.addCallback(
            lambda _: self.crawler.signals.send_catch_log(
                signals.feed_slot_closed, slot=slot
            )
    )
    return d
```

docs/topics/signals.rst
```diff
@@ -307,6 +307,31 @@ spider_error
...
+    :type failure: scrapy.extensions.feedexport._FeedSlot
+
+
+feed_exporter_closed
+~~~~~~~~~~~~
+
+.. signal:: feed_exporter_closed
+.. function:: feed_exporter_closed()
+
+    Sent when the feed exporter is closed.
```
Gallaecio, line 330:
```suggestion
    Sent when the ref:`feed exports <topics-feed-exports>` extension is closed,
    during the handling of the :signal:`spider_closed` signal by the extension,
    after all feed exporting has been handled.
```

scrapy/extensions/feedexport.py
```diff
@@ -310,6 +312,8 @@ def finish_exporting(self):
 
 
 class FeedExporter:
+    pending_deferreds = []
```
Gallaecio, line 315: Any reason *not* to make it private?

```suggestion
    _pending_deferreds = []
```

We usually make things private unless we have a reason to make them public. It makes it easier to implement future changes by not being bound by backward compatibility promises.

scrapy/extensions/feedexport.py
```diff
@@ -397,6 +407,15 @@ def _close_slot(self, slot, spider):
         d.addErrback(
             self._handle_store_error, logmsg, spider, type(slot.storage).__name__
         )
+
+        self.pending_deferreds.append(d)
+        d.addCallback(
+            lambda _: self.crawler.signals.send_catch_log(
+                signals.feed_slot_closed, slot=slot
+            )
```
Gallaecio, line 415: No strong opinion on it, but I wonder if it would not make sense to make this signal also awaitable, like `feed_exporter_closed`, since it might not be that much work.

Gallaecio (commented):
OK, I did not realize another method (item_scraped) was in the picture.

The approach looks good to me, then.

Next step: Please, check on the CI issues; we have some flaky tests, but at least some of the issues seem legit.

scrapy/extensions/feedexport.py
guillermo-bondonno: I addressed the changes above and this one. I can't think of any other change than using send_catch_log_deferred, I added some tests and it works. Is that what you were referring to?

docs/topics/signals.rst
```diff
@@ -308,26 +308,28 @@ spider_error
...
-    This signal does not support returning deferreds from its handlers.
+    This signal supports returning deferreds from its handlers.
 
     :param slot: the slot closed
     :type failure: scrapy.extensions.feedexport._FeedSlot
 
 
 feed_exporter_closed
-~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~
```
Gallaecio, line 325: The point here is to align with the header text length.

```suggestion
~~~~~~~~~~~~~~~~~~~~
```

Gallaecio (approved):

docs/topics/signals.rst
```diff
@@ -307,6 +307,33 @@ spider_error
...
+
+.. signal:: feed_slot_closed
+.. function:: feed_slot_closed(slot)
+
+    Sent when a :ref:`feed exports <topics-feed-exports>` slot is closed.
+
+    This signal supports returning deferreds from its handlers.
+
+    :param slot: the slot closed
+    :type failure: scrapy.extensions.feedexport._FeedSlot
```
wRAR, line 321:
```suggestion
    :type slot: scrapy.extensions.feedexport._FeedSlot
```

wRAR: It also looks wrong that the public API now exposes an underscore-prefixed class :)

Gallaecio: Hmm, that’s a good point. Although addressing that properly would not be trivial, i.e. we should probably review the _FeedSlot API and decide which parts to keep public and which parts to make private as we rename _FeedSlot to FeedSlot.

On the bright side, no need for a deprecation period for _FeedSlot, at [it does not seem overused](https://www.google.com/search?q=%22_FeedSlot%22&nfpr=1). But it would be trivial to add one, I think we have functions or decorators for that, so it might be still worth considering.

VMRuiz: It makes sense to either make `FeedSlot` class public or to create an equivalent class to inform users when a feed has been exporter somewhere, the number of items, size  and other attributes...

wRAR: `_FeedSlot` is a class that takes 9 arguments (file, exporter, storage, uri, format, store_empty, batch_id, uri_template, filter), and stores them as public attrbiutes and also stores `itemcount` (modified externally) and `_exporting` (modified internally). We probably don't need to expose all of these, though maybe we do.

wRAR: We have decided to make `_FeedSlot` public and to keep the old private name with a deprecation warning.

VMRuiz: @guillermo-bondonno  Could you please do this?

guillermo-bondonno: Sure

guillermo-bondonno:
Let me know if you had something else else in mind

Gallaecio:
The original issue reported [here](https://github.com/scrapy/scrapy/pull/5876#discussion_r1170264981), while minor, is still pending :sweat_smile:

Gallaecio (approved):

VMRuiz (approved):

Gallaecio:
Thanks!

Issue #5875, opened 2023-03-27
Fix broken JavaScript reference in the docs


Gallaecio:
As @Laerte [pointed out](https://github.com/scrapy/scrapy/pull/5072#issuecomment-1483391283), https://github.com/scrapy/scrapy.org/issues/224 also applies to docs.scrapy.org, so changes are needed in this repository as well.

PR #5877, opened 2023-03-28
Fix: `Uncaught ReferenceError: analytics is not defined`

jxlil:
@Gallaecio A rollback was performed on the code. However, the new domain was kept.

Closes https://github.com/scrapy/scrapy/issues/5875


Gallaecio:
I think that part of the code was deleted on purpose.

I think the solution here is to either unwrap those ga calls so that they are not defined on a callback of a method from the (now deleted) analytics object, or to delete this layout template override altogether, ga calls included.

jxlil:
It seems that Segment is no longer being used (https://github.com/scrapy/scrapy/pull/5072#discussion_r605544240). I think the best thing would be to delete the file.

wRAR (approved):

PR #5880, opened 2023-03-30
return failure added to _sent_failed 

### LINKED ISSUE #1611: _sent_failed cut the errback chain in MailSender
`MailSender._sent_failed` return `None`, instead of `failure`.  This cut the errback call chain, making impossible to detect in the code fail in the mails in client code.

Prathm-s:
Fixes #1611

### DIFF
--- a/scrapy/mail.py
+++ b/scrapy/mail.py
@@ -164,6 +164,7 @@ class MailSender:
                 "mailerr": errstr,
             },
         )
+        return failure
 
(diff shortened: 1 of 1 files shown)


Gallaecio (approved):

felipeboffnunes:
@wRAR if you think this is good as it is, lets merge it and close the issue. For me the PR is fair enough.

wRAR (approved):

PR #5884, opened 2023-04-02
Simplify create_instance calls

### LINKED ISSUE #5523: Simplify create_instance calls

```python
```

```python
```

```python
```

```python
```

ayushanand308:
Fixes #5523 
Separated create_instance() into two separate function.

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -140,36 +140,35 @@ def rel_has_nofollow(rel):
     return rel is not None and "nofollow" in rel.replace(",", " ").split()
 
 
-def create_instance(objcls, settings, crawler, *args, **kwargs):
-    """Construct a class instance using its ``from_crawler`` or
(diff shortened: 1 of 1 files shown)


wRAR:
Without looking at the code changes themseves I see you've removed `create_instance`. I also suspect you haven't run the tests as they would fail without that function.

I think this PR should also include switching the Scrapy code that is currently using `create_instance` to these new functions.

ayushanand308:
@wRAR Thank you for your feedback on my pull request. I apologize for not being clear in my initial message. I will make the necessary changes and update the pull request accordingly.

I understand that I need to keep the create_instance function and replace its usages in Scrapy code with the new functions build_from_crawler and build_from_settings.

wRAR:
Yup, this is correct.

Gallaecio:
Closing, since it has been months and there are other promising PRs open for this.

PR #5885, opened 2023-04-04
Replace get_func_args

### LINKED ISSUE #5872: get_func_args does not fully work in CPython
As [shown in tests](https://github.com/scrapy/scrapy/blob/ada917307844950a81226f020b596d5932187f6e/tests/test_utils_python.py#L240-L243), `get_func_args` does not work in CPython with inputs like `str.split`, `"".join` or `itemgetter(2)`.

serhii73:
Close #5872

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -3,6 +3,7 @@ This module contains essential stuff that should've come with Python itself ;)
 """
 import gc
 import inspect
+import operator
 import re
 import sys
 import weakref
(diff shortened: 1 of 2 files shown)


wRAR:
We had some discussions regarding the `operator.itemgetter` case and so far it's not clear to me how to handle it in the best way. On CPython `inspect.signature(operator.itemgetter(2))` raises `ValueError: callable operator.itemgetter(2) is not supported by signature` and my initial suggestion was to catch this exception and return `[]`. I've just checked this on PyPy and it returns `<Signature (obj)>`, so it works there. So it seems that we still need separate checks in the test for CPython and PyPy, where PyPy will assert `["obj"]` like it does now, and CPython will expect `[]`? 

Specifically omitting `operator.itemgetter` like in this PR is not a good idea because `get_func_args()` previously worked for it on PyPy and because there should be other cases that doesn't work on CPython. Not handling the exception is also worse than returning `[]`. So keeping separate test conditions seems the best approach to me.

Gallaecio (approved):

wRAR (approved):

PR #5891, opened 2023-04-11
Replace the raw checksum with b64 decoded checksum for GCS

### LINKED ISSUE #5874: Scrapy does not decode base64 MD5 checksum from GCS
### Description

Incorrect GCS Checksum processing

### Steps to Reproduce

1. Obtain the checksum for an up-to-date file.

matches the checksum of the file downloaded
NOT matches the checksum of the file downloaded
Always
### Versions
current

### Additional context

https://cloud.google.com/storage/docs/json_api/v1/objects
> MD5 hash of the data, encoded using [base64](https://datatracker.ietf.org/doc/html/rfc4648#section-4).

But, Scrapy dose not decode MD5 from GSC.

heppymxm:
Fixes #5874 

Replace the raw checksum with b64 decoded checksum for GCS

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -3,6 +3,7 @@ Files Pipeline
 
 See documentation in topics/media-pipeline.rst
 """
+import base64
 import functools
 import hashlib
 import logging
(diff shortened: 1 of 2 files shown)


Gallaecio:
Thanks!

PR #5892, opened 2023-04-11
implement request_to_curl function

guillermo-bondonno:
Implement request_to_curl util function that generates a curl command string from a scrapy.Request object

### DIFF
--- a/scrapy/utils/request.py
+++ b/scrapy/utils/request.py
@@ -346,16 +346,17 @@ def request_to_curl(request: Request) -> str:
     )
     url = request.url
     cookies = ""
-    if isinstance(request.cookies, dict) and request.cookies:
-        cookies = "--cookie '{}'".format(
-            "; ".join("{}={}".format(k, v) for k, v in request.cookies.items())
-        )
-    elif isinstance(request.cookies, list):
-        cookies = "--cookie '{}'".format(
-            "; ".join(
-                "{}={}".format(list(c.keys())[0], list(c.values())[0])
-                for c in request.cookies
+    if request.cookies:
+        if isinstance(request.cookies, dict):
+            cookies = "--cookie '{}'".format(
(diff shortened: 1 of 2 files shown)


scrapy/utils/request.py
```diff
@@ -327,3 +327,35 @@ def _get_method(obj, name):
...
+        "-H '{}: {}'".format(k.decode(), v[0].decode())
+        for k, v in request.headers.items()
+    )
+    url = request.url
+    cookies = ""
+    if isinstance(request.cookies, dict) and request.cookies:
+        cookies = "--cookie '{}'".format(
+            "; ".join("{}={}".format(k, v) for k, v in request.cookies.items())
+        )
+    elif isinstance(request.cookies, list):
```
wRAR, line 353: Maybe this also needs a check that the list is not empty?

wRAR:
Can you please also look into pylint failures?

scrapy/utils/request.py
guillermo-bondonno: Makes sense. The default is an empty dict, but I guess it is possible that an empty list ends up there.

guillermo-bondonno:
> Can you please also look into pylint failures?

@wRAR It's about using .format instead of f-strings. I think it's much cleaner to use format in this case instead of doing something like

`cookies = f"--cookie '{'; '.join(f'{list(c.keys())[0]}={list(c.values())[0]}' for c in request.cookies)}'"`

but if it can't be merged as it is, I'll find a way around it sure.
What do you think?

wRAR:
I think this warning should indeed be disabled, per-line or globally. I'll wait for other opinions though.

wRAR (approved):

Gallaecio (approved):
I think it would be ideal to cover this somewhere in the docs for discoverability, but at the same time I cannot find a good place in the existing docs for it, so +1 to merge as is.

PR #5893, opened 2023-04-12
Minor typo fix on tutorial.rst

sethherr:


Gallaecio:
Thanks!

wRAR:
This was merged into 2.8 (we really need something to check/prevent that), it's fine to have it in 2.8 but we need to cherry-pick it into master.

PR #5896, opened 2023-04-15
Typing for scrapy/core/downloader

wRAR:
It doesn't seem worth it to add typing for code that uses Twisted HTTP stuff, as that code is itself untyped and also some of our code uses zope.interface which is not really supported by mypy.

### DIFF
--- a/scrapy/core/downloader/tls.py
+++ b/scrapy/core/downloader/tls.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any, Dict
 
 from OpenSSL import SSL
 from service_identity.exceptions import CertificateError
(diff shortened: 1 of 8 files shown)


scrapy/core/downloader/tls.py
```diff
@@ -55,11 +58,12 @@ def _identityVerifyingInfoCallback(self, connection, where, ret):
                     connection.get_cipher_name(),
                 )
                 server_cert = connection.get_peer_certificate()
-                logger.debug(
-                    'SSL connection certificate: issuer "%s", subject "%s"',
-                    x509name_to_string(server_cert.get_issuer()),
-                    x509name_to_string(server_cert.get_subject()),
-                )
+                if server_cert:
```
Gallaecio, line 61: Worth it! :)

scrapy/core/downloader/webclient.py
```diff
@@ -166,8 +171,8 @@ def __init__(self, request, timeout=180):
         elif self.method == b"POST":
             self.headers["Content-Length"] = 0
 
-    def __repr__(self):
-        return f"<{self.__class__.__name__}: {self.url}>"
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}: {self._url}>"
```
Gallaecio, line 175: :tada:

Issue #5855, opened 2023-03-23
test_batch_path_differ sometimes fails


Gallaecio:
See https://github.com/scrapy/scrapy/pull/5847#issuecomment-1471778039.

tstauder:
@jmannoop and I would like to work on this. Is it possible for it to be assigned to us?

tstauder:
I think we have the fix ready to go and tested, but could use some guidance on creating a pull request. I can probably figure out how to make one but if there's anything else we need to know / tips would be greatly appreciated.

PR #5898, opened 2023-04-17
Fix for #5855

tstauder:
Changed batch_time to batch_id in test_batch_path_differ function in test_feedexport: #5855 

Fixes https://github.com/scrapy/scrapy/issues/5855

### DIFF
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -2216,7 +2216,8 @@ class FeedPostProcessedExportsTest(FeedExportTestBase):
 
 class BatchDeliveriesTest(FeedExportTestBase):
     __test__ = True
-    _file_mark = "_#%(batch_id)02d_"
+    _file_mark = "_%(batch_time)s_#%(batch_id)02d_"
+    # _file_mark = "_#%(batch_id)02d_"
 
     @defer.inlineCallbacks
     def run_and_export(self, spider_cls, settings):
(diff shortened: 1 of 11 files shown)


Gallaecio:
> https://github.com/scrapy/scrapy/issues/5855#issuecomment-1509289786

@tstauder First suggestion: review your changes as [seen in your pull request](https://github.com/scrapy/scrapy/pull/5898/files), and make sure they only include intended changes. I see some commented out code that you probably used during debugging and should not be a part of this change. I also see a bad edit of a docstring.

tstauder:
Thanks for the helpful feedback. I cleaned up those issues. I've update the pull request to include them.

tests/test_feedexport.py
```diff
@@ -2398,7 +2398,7 @@ def test_export_items(self):
         yield self.assertExported(items, header, rows, settings=settings)
 
     def test_wrong_path(self):
-        """If path is without %(batch_time)s and %(batch_id) an exception must be raised"""
+        """If path is without %(batch_id) an exception must be raised"""
```
Gallaecio, line 2401: This change does not seem necessary.

Gallaecio (approved):
Other than the minor issue I mention below, it looks good to me. Thanks!

tests/test_feedexport.py
felipeboffnunes: Click on accept suggestion for a free merge!

```suggestion
        """If path is without %(batch_time)s and %(batch_id) an exception must be raised"""
```

tstauder: Just saw this suggestion now, but I also added commit that does the same thing

felipeboffnunes (approved):

Issue #5899, opened 2023-04-18
Request.from_curl() with $-prefixed string literals


wRAR:
Chrome (and probably other things) sometimes generate curl commands with a [$-prefixed](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html) data string, probably when it's easier to represent the string in that way or when it includes non-ASCII characters, e.g. the DiscoverQueryRendererQuery XHR on https://500px.com/popular is copied as 

```
curl 'https://api.500px.com/graphql' \
<headers omitted>
  --data-raw $'{"operationName":"DiscoverQueryRendererQuery",<omitted> "query":"query DiscoverQueryRendererQuery($filters: [PhotoDiscoverSearchFilter\u0021], <the rest omitted>' \
  --compressed
```

, most likely because of `\u0021` in this payload.

`scrapy.utils.curl.curl_to_request_kwargs()` isn't smart enough to understand this kind of shell escaping, so it puts `$` into the request body which is incorrect. Ideally we should support this, though I don't know if there are existing libraries to unescape this.

PR #5901, opened 2023-04-19
Fix: Request.from_curl() with $-prefixed string literals

jxlil:
Hi @wRAR 

I added an [`action`](https://docs.python.org/3/library/argparse.html#action) for the `--data` argument that removes the _string prefix_ and decodes the string. Although `Request.from_curl()` now works correctly with requests like the one you shared in the issue, I don't know if this is the best way to approach this.

```python
curl_command = "curl 'https://www.example.org/' --data-raw $'{\"$filters\":\"Filter\u0021\"}'"
curl_to_request_kwargs(curl_command)

# output: {'method': 'POST', 'url': 'https://www.example.org/', 'body': '{"$filters":"Filter!"}'}
```

Closes https://github.com/scrapy/scrapy/issues/5899

### DIFF
--- a/scrapy/utils/curl.py
+++ b/scrapy/utils/curl.py
@@ -11,8 +11,7 @@ from w3lib.http import basic_auth_header
 class DataAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         value = str(values).encode("utf-8").decode("utf-8")
-        items = re.findall(r"\$(.+)", value)
-        value = items[0] if items else value
+        value = value[1::] if re.match(r"^\$(.+)", value) else value
         setattr(namespace, self.dest, value)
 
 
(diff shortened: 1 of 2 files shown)


scrapy/utils/curl.py
```diff
@@ -7,6 +8,13 @@
 from w3lib.http import basic_auth_header
 
 
+class DataAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        value = str(values).encode("utf-8").decode("utf-8")
+        value = value[1::] if re.match(r"^\$(.+)", value) else value
```
Gallaecio, line 14: This feels over complicated.

Wouldn’t something like this work?

```suggestion
        value = str(values)
        if value.startswith("$"):
            value = value[1:]
```

Gallaecio (approved):

scrapy/utils/curl.py
jxlil: Yes, I just checked and it works too. Thanks

wRAR:
Hmm this only handles the escapes that work in Python strings (`\uHHHH` like in the example, likely some others), but it's nice that it does this without using any additional libraries, so I'm fine with merging it.

wRAR (approved):

PR #5902, opened 2023-04-19
fix: docs/intro/tutorial.rst checks

jxlil:
Fix docs/intro/tutorial.rst checks:

command:
```
pytest docs
```

output:
```
FAILED docs/intro/tutorial.rst::line:254,column:1
FAILED docs/intro/tutorial.rst::line:350,column:1
FAILED docs/intro/tutorial.rst::line:412,column:1
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -156,7 +156,7 @@ basepython = {[testenv:pypy3]basepython}
 deps =
     {[pinned]deps}
     PyPyDispatcher==2.1.0
-commands = {[pinned]commands}
+commands = {[testenv:pypy3]commands}
 install_command = {[pinned]install_command}
 setenv =
     {[pinned]setenv}
--- a/tox.ini
+++ b/tox.ini
@@ -100,7 +100,7 @@ setenv =
 install_command =
     python -I -m pip install {opts} {packages}
 commands =
-    pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:--durations=10 docs scrapy tests}
+    pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:--durations=10 scrapy tests}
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):

Gallaecio:
I wonder what to do with pinned environments. 

Maybe we could skip doctests with pinned versions:
1. [Remove --doctest-modules from pytest.ini](https://github.com/scrapy/scrapy/blob/master/pytest.ini#L8)
2. [Add it to tox.ini’s default command](https://github.com/scrapy/scrapy/blob/master/tox.ini#L33)
3. Add custom commands to the 4 pinned environments in tox.ini that do not have that option.

jxlil:
@Gallaecio I made the changes you mention, it seems that there are no more failures related to `tutorial.rst`

tox.ini
```diff
@@ -152,7 +156,7 @@ basepython = {[testenv:pypy3]basepython}
 deps =
     {[pinned]deps}
     PyPyDispatcher==2.1.0
-commands = {[testenv:pypy3]commands}
+commands = {[pinned]commands}
```
Gallaecio, line 159: This was probably OK before, `testenv:pypy3` above does not use `--doctest-modules`. I suspect this change may be responsible for the job failing now (maybe coverage breaks pypy?)

jxlil: I undid the change, but the test failed again. So I added a custom command for `pypy3-pinned`

Gallaecio: > I remember that one was failing

I did re-run the failing test in case it was a random issue, but I did not expect it to pass :sweat_smile:

Gallaecio:
Nice, thanks!

PR #5907, opened 2023-04-21
GCP Service account can be set as string in settings.py

mo-tech55:
With this feature you do not need to store the service account json file for Google Cloud authentication. Instead you can save the string in the settings.py file


scrapy/extensions/feedexport.py
```diff
@@ -226,13 +226,19 @@ def from_crawler(cls, crawler, uri):
             uri,
             crawler.settings["GCS_PROJECT_ID"],
             crawler.settings["FEED_STORAGE_GCS_ACL"] or None,
+            crawler.settings["GCP_SERVICE_ACCOUNT_CREDENTIALS"] or None,
```
wRAR, line 229: I don't think this works as the new argument is not added to `__init__`.

PR #5908, opened 2023-04-21
Update tool versions

wRAR:
No code changes (but we don't bump flake8 until we drop 3.7).


wRAR:
Oh, `bandit` actually failed, I haven't noticed that in its output locally. It triggers https://bandit.readthedocs.io/en/latest/plugins/b113_request_without_timeout.html for code not related to `requests`, so we should just silence that.

Gallaecio (approved):

Issue #5894, opened 2023-04-13
Add support for the Parsel JMESPath feature


wRAR:
See https://github.com/scrapy/parsel/pull/181 and https://github.com/Gallaecio/scrapy/commit/d642881e25b6a7ce7d8f640b354459b9a4921790 linked there.

Note that the minimum version of Parsel that Scrapy requires should stay the same, i.e. no need to update the minimum version of Parsel in setup.py.

nitin-pandita:
I am up for it @wRAR  can you assign me this issue

wRAR:
@nitin-pandita as you can see there is already a draft PR for this issue.

nitin-pandita:
Oh yeah, I didn't notice. Thank you for the reply

Issue #5903, opened 2023-04-19
UserWarning: Selector got both text and root, root is being ignored. WARNING


gelodefaultbrain:
### Description

`UserWarning: Selector got both text and root, root is being ignored.` Only happens on Ubuntu

### Steps to Reproduce

1. Open a simple python shell
2. Paste code below
3. Run it
```
from scrapy.selector import Selector
link = "<a class='fsm fwn fcdg' target='_blank' href='/ta/id/340639-par-olaines-novada-domes-2009-gada-30-septembra-sedes-lemuma-apstiprinato-saistoso-noteikumu-nr-76-par-detalplanojuma-apstiprinasanu'>Par Olaines novada domes 2009. gada 30. septembra sdes lmum apstiprinto saistoo noteikumu Nr. 76 'Par detlplnojuma apstiprinanu nekustamajam paumam 'Vrpas' (kad. Nr. 8080 008 0095)', pielikuma Nr. 66 atcelanu</a>"
selector = Selector(text=link, type='html')
```

**Warning  output:**
```
/home/iamuser/.local/share/virtualenvs/myenv-YA2UsV5L/lib/python3.11/site-packages/scrapy/selector/unified.py:83: UserWarning: Selector got both text and root, root is being ignored.
  super().__init__(text=text, type=st, root=root, **kwargs)
```

**Expected behavior:** No warnings? It's quite strange because it doesn't appear on MacOS and Windows

**Actual behavior:** Shows warning message above

**Reproduces how often:** 100% on Ubuntu

### Versions

Please paste here the output of executing `scrapy version --verbose` in the command line.

```
Scrapy       : 2.8.0
lxml         : 4.9.2.0
libxml2      : 2.9.14
cssselect    : 1.2.0
parsel       : 1.8.1
w3lib        : 2.1.1
Twisted      : 22.10.0
Python       : 3.11.2 (main, Feb  7 2023, 13:52:42) [GCC 11.3.0]
pyOpenSSL    : 23.1.1 (OpenSSL 3.0.8 7 Feb 2023)
cryptography : 39.0.1
Platform     : Linux-5.15.0-69-generic-x86_64-with-glibc2.35
```

### Additional context

Why does this warning appear on ubuntu? I tried running it on MacOS and Windows and it doesn't appear. Is this something of a concern? I believe this should be supressed. Thanks!

gelodefaultbrain:
> 


Hi @GeorgeA92 

This is the output of my `scrapy version --verbose`

[...]

PR #5915, opened 2023-04-28
Added support for the Parsel JMESPath feature

jxlil:
Added support for `jmespath`

Closes https://github.com/scrapy/scrapy/issues/5894

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -1281,6 +1281,12 @@ TextResponse objects
     :class:`TextResponse` objects support the following methods in addition to
     the standard :class:`Response` ones:
 
+    .. method:: TextResponse.jmespath(query)
+
+        A shortcut to ``TextResponse.selector.jmespath(query)``::
+
+            response.jmespath('object.[*]')
+
     .. method:: TextResponse.xpath(query)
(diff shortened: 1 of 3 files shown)


wRAR:
The test on pinned envs failed because those use the old parsel version. We want to make sure that Scrapy and user code continue working with old parsel. Can you please check and fix those issues, probably just by marking the tests `xfail`?

wRAR:
I've checked the code and it doesn't make sure `selector.jmespath()` is available before calling it, so marking tests as xfail is wrong/not enough, you should add a check to the code and make new tests that test it. You can check https://github.com/scrapinghub/web-poet/pull/167 for both the runtime check and tests implementations.

jxlil:
@wRAR I have updated the code, do you think it is necessary to make any other changes?

Gallaecio (approved):
Looks good to me, nice work!

It would be great if we could also address https://github.com/scrapy/scrapy/issues/5903, as these 2 are the only Parsel-related changes blocking Scrapy 2.9, but we can handle that separately.

wRAR (approved):

wRAR:
Thanks!

Issue #5906, opened 2023-04-21
Add possibility to use Selector (bytes) added in parsel 1.8.0.


GeorgeA92:
## Summary

Please review possibility to apply RAM memory efficient Selector that can accept `bytes` as input 
added as result of https://github.com/scrapy/parsel/issues/210 (added in parsel 1.8.0) in scrapy (as default)

Currently UnifiedSelector (subclass of parsel's Selector used in scrapy) configured to use Response.text as input
Response.text (property) -> creates new object (Response.body -> converted to str) which is memory intensive and not needed 
https://github.com/scrapy/scrapy/blob/5a37af146f4f036168ac562918fca43adb4ac65f/scrapy/selector/unified.py#L66-L73

## Motivation

As mentioned on https://github.com/scrapy/parsel/issues/210 usage of `str` input for creating Selector object is more RAM memory intensive

## Describe alternatives you've considered

At this stage it will require to.. separately create Selector object with bytes input inside spiders callback method (and not use Response.Selector)

## Additional context
Removing other usages of Response.text will significantly reduce RAM required to processing response

wRAR:
We've discussed this briefly yesterday and found that this may not be useful if we detect the response encoding at the same time as converting body to str (I think the relevant code is at https://github.com/scrapy/scrapy/blob/5a37af146f4f036168ac562918fca43adb4ac65f/scrapy/http/response/text.py#L105 ?). We didn't do any extensive analysis of this though.

GeorgeA92:
As we can see from

https://github.com/scrapy/scrapy/blob/5a37af146f4f036168ac562918fca43adb4ac65f/scrapy/http/response/text.py#L62-L72

Response object with valid encoding received without call of `_body_inferred_encoding` (with conversion to unicode `str` ) has `_cached_ubody` as `None` and `encoding` (`not None`).

It means that on unified Selector `__init__` we can safely create `bytes` based selector for Response objects that match values mentioned above (and parsel v.1.8.0) and probably maintain current behaviour for other (w. encoding received from `_body_inferred_encoding` call) Response objects

https://github.com/scrapy/scrapy/blob/5a37af146f4f036168ac562918fca43adb4ac65f/scrapy/selector/unified.py#L66-L83

PR #5918, opened 2023-05-04
fix: Handle Parsel warning

Laerte:
resolves: #5903

Please let me know if this is the best approach to handle this.


scrapy/selector/unified.py
```diff
@@ -63,7 +68,7 @@ class Selector(_ParselSelector, object_ref):
     __slots__ = ["response"]
     selectorlist_cls = SelectorList
 
-    def __init__(self, response=None, text=None, type=None, root=None, **kwargs):
+    def __init__(self, response=None, text=None, type=None, root=_NOT_SET, **kwargs):
```
Gallaecio, line 71: I would try to avoid importing Parsel’s `_NOT_SET`.

For example, we could declare our own _NOT_SET, and handle root though the kwargs variable, i.e. if root is not _our_ _NOT_SET, we add it to kwargs:

```python
_NOT_SET = object()
...
def __init__(self, response=None, text=None, type=None, root=_NOT_SET, **kwargs):
...
if root is not _NOT_SET:
    kwargs["root"] = root
...
super().__init__(text=text, type=st, **kwargs)
```

Not a strong opinion, though. If one project can use a private member of Parsel that’s Scrapy.

Gallaecio (approved):

scrapy/selector/unified.py
GeorgeA92: On current parsel version(1.8.1) I definitely don't see option without import of `_NOT_SET` from parsel module. Any other internally created `object()` on root - will not pass `_NOT_SET` checks on parsel side

I tried similar approach
https://github.com/scrapy/scrapy/compare/master...GeorgeA92:scrapy:warning_roottext_from_parser_jmespath

Gallaecio: The key is in _not_ passing `root` at all if set to _NOT_SET on Scrapy’s end.

To elaborate on the code I shared above:

```python
_NOT_SET = object()

...

def __init__(self, response=None, text=None, type=None, root=_NOT_SET, **kwargs):
        if response is not None and text is not None:
            raise ValueError(
                f"{self.__class__.__name__}.__init__() received "
                "both response and text"
            )
        st = _st(response, type)
        if text is not None:
            response = _response_from_text(text, st)
        if response is not None:
            text = response.text
            kwargs.setdefault("base_url", response.url)
        self.response = response
        if root is not _NOT_SET:
            kwargs["root"] = root
        super().__init__(text=text, type=st, **kwargs)
```

Laerte: Very clever, thanks @Gallaecio 😮

GeorgeA92: @Gallaecio @Laerte 
Technically it will work.
Hovewer I don't like any option that make this check on each call of Selector `__init__`.

On spider like this:
```python

    def start_requests(self):

    def parse(self, response):
        links = response.css('a::attr(href)')
```

`__init__` of selector will be called ~50 times per response. On a real case it can be hundreds or even thousands calls of `Selector.__init__` per response.
On applying this - application will make additional +hudnreds/thousands checks of `_NOT_SET` per response. While on initial proposal - related code check will be executed only one time on startup (+root assignment on init). However i din't make detailed profiler tests yet.

wRAR (approved):

wRAR:
Thanks!
We can address the potential optimization improvements separately, together with already filed #5906.

PR #5919, opened 2023-05-04
Enable doc tests for selectors.rst, fix issues

### LINKED ISSUE #5912: Update Parsel output in docs/topics/selectors.rst
`docs/topics/selectors.rst` still contains `[<Selector xpath='//title/text()' data='Example website'>]` etc., it should be updated for the new Parsel version. Also I think it makes sense to modify it so that these examples are checked by tests like in docs/intro/tutorial.rst (probably an invisible code block loading docs/_static/selectors-sample1.html should be enough).

wRAR:
Fixes #5912

### DIFF
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -48,6 +48,8 @@ Constructing selectors
 
 .. highlight:: python
 
+.. skip: start
+
 Response objects expose a :class:`~scrapy.Selector` instance
 on ``.selector`` attribute:
 
@@ -65,6 +67,8 @@ more shortcuts: ``response.xpath()`` and ``response.css()``:
     'good'
     >>> response.css("span::text").get()
     'good'
+
+.. skip: end
 
 Scrapy selectors are instances of :class:`~scrapy.Selector` class
 constructed by passing either :class:`~scrapy.http.TextResponse` object or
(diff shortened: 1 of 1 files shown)


docs/topics/selectors.rst
```diff
@@ -808,8 +821,8 @@ nodes can be accessed directly by their names:
 
     >>> response.selector.remove_namespaces()
     >>> response.xpath("//link")
-    [<Selector xpath='//link' data='<link rel="alternate" type="text/html" h'>,
-        <Selector xpath='//link' data='<link rel="next" type="application/atom+'>,
+    [<Selector query='//link' data='<link rel="alternate" type="text/html" h'>,
+        <Selector query='//link' data='<link rel="next" type="application/atom+'>,
```
Gallaecio, line 825: This falls within  a skip section, so it would not have been caught. I did not check why, though, so maybe it needs to be skipped?

Gallaecio (approved):

docs/topics/selectors.rst
wRAR: This section is skipped because the code deals with an external source (we could include a sample XML file with namespaces into the doc instead, but that's out of scope here I think), but I don't see why we shouldn't change the expected output :) A user who runs this with recent Parsel will see `query=`.

PR #5925, opened 2023-05-07
Typing for scrapy/utils

wRAR:
Another batch of improved typing, mostly covering scrapy.utils.* (not everything there was processed yet but I think it's already quite large).

I have one question for @elacuesta: #3869 added `scrapy.utils.datatypes.LocalWeakReferencedCache` which returns None if "key is not weak-referenceable", and because of how `scrapy.utils.misc.is_generator_with_return_value()` uses it, that function can return None in addition to an expected bool. It even uses a pattern of putting a value into a dict and getting it back. So I wonder if None there has the same value as False (as the result of `is_generator_with_return_value()` is only checked with a simple `if`) and if it's fine to cast it to bool like I do in this PR.


wRAR:
`collections.OrderedDict` and `weakref.WeakKeyDictionary` are not generic in 3.7 and 3.8, the former can be replaced with `typing.OrderedDict` while the latter I think doesn't have that option, so I think I'll just remove specialization from it.

scrapy/utils/deprecate.py
```diff
@@ -134,7 +137,17 @@ def _clspath(cls, forced=None):
...
+def update_classpath(path: str) -> str:
+    ...
+
+
+@overload
+def update_classpath(path: Any) -> Any:
+    ...
+
+
+def update_classpath(path: Any) -> Any:
```
Gallaecio, line 150: Based on [the example on the docs](https://docs.python.org/3.7/library/typing.html#typing.overload), there should be no type hints here. Does mypy complain?

```suggestion
def update_classpath(path):
```

scrapy/utils/gz.py
```diff
@@ -18,17 +21,12 @@ def gunzip(data):
...
-                try:
-                    output_list.append(f.extrabuf[-f.extrasize :])
-                finally:
-                    break
-            else:
-                raise
+            # some pages are quite small so output_list is empty
+            if output_list:
+                break
+            raise
```
Gallaecio, line 27: Could you elaborate on this change?

tests/test_pipeline_files.py
```diff
@@ -639,31 +640,29 @@ def test_blob_path_consistency(self):
...
-        content = get_ftp_content_and_delete(
-            path,
-            store.host,
-            store.port,
-            store.username,
-            store.password,
-            store.USE_ACTIVE_MODE,
-        )
-        self.assertEqual(data.decode(), content)
+        with MockFTPServer() as ftp_server:
```
Gallaecio, line 647: Niiiice!

scrapy/utils/deprecate.py
wRAR: If there is no type hints on the actual function, its body won't be checked (at least without `--check-untyped-defs`).

scrapy/utils/gz.py
wRAR: Right, I was going to do that but forgot.

`extrabuf` and `extrasize` were removed in 3.5 when the underlying logic was rewritten, they were never documented as far as I can see. Looks like this code was intended to salvage a partially decompressed chunk when possible, and now it's probably just not possible. The main purpose of this function though is to salvage previous completely decompressed chunks and it still does that, and tests in https://github.com/scrapy/scrapy/blob/master/tests/test_utils_gz.py only test that, as far as I can see.

Gallaecio (approved):

scrapy/utils/gz.py
wRAR: For the file in that test the first read1 call returns the data we expect. I'm not going to dive further into GzipFile implementations, it's possible that it's just more robust than before 3.5, I'm fine that the tests return the expected data with the code we have.

PR #5830, opened 2023-02-09
periodic-log-dump-stats: feature added

GeorgeA92:
Aimed to solve https://github.com/scrapy/scrapy/issues/2173

The main idea of this approach is to.. get  **all**  `int` and `float` values of `crawler.stats` (and key names of course) and log.. it's changes (delta) on periodic basis.
on default log interval (60 seconds) it will log numeric stats changes for last minute.

<details><summary>script.py (books to scrape LOG_LEVEL -> INFO)</summary>

```python
import scrapy
from scrapy.crawler import CrawlerProcess

class BooksToScrapeSpider(scrapy.Spider):
    name = "books"
    custom_settings = {
        "DOWNLOAD_DELAY": 0.3,
        "LOG_LEVEL": 'INFO'
    }
    def start_requests(self):
        yield scrapy.Request(url='https://books.toscrape.com/', callback=self.parse)

    def parse(self, response):
        for book_page_link in response.css('.product_pod a::attr(href)').getall():
            yield scrapy.Request(response.urljoin(book_page_link), self.parse_book)
        if next_page := response.css("li.next a::attr(href)").get():
            yield scrapy.Request(url=response.urljoin(next_page), callback=self.parse)

    def parse_book(self, response):
        yield {'response.url': response.url}

if __name__ == "__main__":
    process = CrawlerProcess()
    process.crawl(BooksToScrapeSpider)
    process.start()
```
</details>

<details><summary>log output</summary>

```
2023-02-09 01:27:23 [scrapy.utils.log] INFO: Scrapy 2.8.0 started (bot: scrapybot)
2023-02-09 01:27:23 [scrapy.utils.log] INFO: Versions: lxml 4.9.1.0, libxml2 2.9.14, cssselect 1.1.0, parsel 1.6.0, w3lib 1.21.0, Twisted 22.2.0, Python 3.10.8 | packaged by conda-forge | (main, Nov 24 2022, 14:07:00) [MSC v.1916 64 bit (AMD64)], pyOpenSSL 22.0.0 (OpenSSL 1.1.1s  1 Nov 2022), cryptography 37.0.4, Platform Windows-10-10.0.22621-SP0
2023-02-09 01:27:23 [scrapy.crawler] INFO: Overridden settings:
{'DOWNLOAD_DELAY': 0.3, 'LOG_LEVEL': 'INFO'}
[...]

PR #5926, opened 2023-05-07
periodic_log: implemented as separate extension

GeorgeA92:
The same as current version of https://github.com/scrapy/scrapy/pull/5830 - but implemented as separate extension. So it doesn't affect `scrapy.extensions.logstats.LogStats` and no risk of breaking it's backward compatibility.


scrapy/extensions/logstats_extended.py
```diff
@@ -0,0 +1,112 @@
...
+        ext_exclude = crawler.settings.getlist("LOGSTATS_EXTENDED_EXCLUDE", [])
+        if not interval:
+            raise NotConfigured
+        o = cls(crawler.stats, interval, extended, ext_include, ext_exclude)
+        crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
+        crawler.signals.connect(o.spider_closed, signal=signals.spider_closed)
+        return o
+
+    def spider_opened(self, spider):
+        self.time_prev = datetime.utcnow()
```
Laerte, line 42: Maybe use `.now()` and pass the timezone? [Reason](https://blog.ganssle.io/articles/2019/11/utcnow.html)

scrapy/extensions/logstats_extended.py
```diff
@@ -0,0 +1,112 @@
...
+            "log_interval": self.interval,
+            "start_time": self.stats._stats["start_time"],
+            "utcnow": now,
+            "log_interval_real": (now - self.time_prev).total_seconds(),
+            "elapsed": (now - self.stats._stats["start_time"]).total_seconds(),
+        }
+        self.time_prev = now
+        return {"time": time}
+
+    def log_time(self):
```
Laerte, line 79: Can you add type hints?

scrapy/extensions/logstats_extended.py
GeorgeA92: Implemented here as `datetime.utcnow()` to correspond with  `finish_time` and `start_time` stats values which is also populated by `datetime.utcnow()` in `..extensions.corestats`
https://github.com/scrapy/scrapy/blob/92f86fab062d179fd68590f1be6a24ca13ec33b6/scrapy/extensions/corestats.py#L24C3-L29

This change making sense to apply only if we remove (replace) all other usages of `datetime.utcnow()` in this repo, which may require separate issue/discuss/approval and it is definitely not planned as part if this pull request.

Laerte: Gotcha! Thanks

GeorgeA92:
Implemented as separate extension.
Each part of this (stats, delta, timing) can be configured or enabled/disabled separately as on sample script.

<details><summary>script.py</summary>

```pythob
import scrapy

class BooksToScrapeSpider(scrapy.Spider):
    name = "books"
    custom_settings = \
        {
        "DOWNLOAD_DELAY": 0.3,
        "LOG_LEVEL": 'INFO',
        #"LOGSTATS_EXT_TIMING_ENABLED": True,
        "LOGSTATS_EXT_STATS_ENABLED": True,
        "LOGSTATS_EXT_STATS_INCLUDE": ["downloader/", "scheduler/", "log_count/", "item_scraped_count"],
        # ^ output only stat params that countain one of listed substrings, if not set - allow all
        "LOGSTATS_EXT_STATS_EXCLUDE": ["scheduler/"],
        # ^ exclude stat params that contain one of listed substrings, if not set - allow all
        # ^ It has precedence over LOGSTATS_EXT_STATS_INCLUDE
        #"LOGSTATS_EXT_DELTA_ENABLED": True,
        #"LOGSTATS_EXT_DELTA_INCLUDE": ['downloader/'],
        # ^ works the same as on LOGSTATS_EXT_STATS_ setting
        "EXTENSIONS":{
            "scrapy.extensions.logstats_extended.LogStatsExtended": 0,
        }
        }
    def start_requests(self):
        yield scrapy.Request(url='https://books.toscrape.com/', callback=self.parse)

    def parse(self, response):
        for book_page_link in response.css('.product_pod a::attr(href)').getall():
            yield scrapy.Request(response.urljoin(book_page_link), self.parse_book)

    def parse_book(self, response):
        yield {'response.url': response.url}

if __name__ == "__main__":
    process = CrawlerProcess()
    process.crawl(BooksToScrapeSpider)
    process.start()

```

</details>

scrapy/extensions/logstats_extended.py
```diff
@@ -0,0 +1,132 @@
...
+from twisted.internet import task
+
+from scrapy import signals
+from scrapy.exceptions import NotConfigured
+from scrapy.utils.serialize import ScrapyJSONEncoder
+
+logger = logging.getLogger(__name__)
+
+
+class LogStatsExtended:
```
Gallaecio, line 13: I think it may make sense to keep this new class in `log_stats.py`, rather than having a separate file for it.

I also think we may want to reconsider the class name. I would suggest making `Extended` a prefix instead (`ExtendedLogStats`), and consider an alternative word instead (e.g. `Verbose`, `Detailed`, `Flexible`, `Configurable`).

scrapy/extensions/logstats_extended.py
```diff
@@ -0,0 +1,132 @@
...
+        self.ext_delta_include = ext_delta_include
+        self.ext_delta_exclude = ext_delta_exclude
+        self.ext_timing_enabled = ext_timing_enabled
+        self.ext_timing_include = ext_timing_include
+        self.ext_timing_exclude = ext_timing_exclude
+
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        interval = crawler.settings.getfloat("LOGSTATS_INTERVAL")
```
Gallaecio, line 46: Instead of having 3 new `_ENABLED` settings, we could have a single one called `LOGSTATS_FOO`, where `FOO` would be whatever the prefix/suffix of the new class is, e.g. `LOGSTATS_EXTENDED` if we keep the current name.

```suggestion
        settings = crawler.settings
        interval = settings.getfloat("LOGSTATS_INTERVAL")
        if not interval or not settings.getfloat("LOGSTATS_EXTENDED"):
            raise NotConfigured
```

Then, to get those `_enabled` booleans, you could do:

```suggestion
        ext_stats_include = crawler.settings.getlist("LOGSTATS_STATS_INCLUDE", [])
        ext_stats_exclude = crawler.settings.getlist("LOGSTATS_STATS_EXCLUDE", [])
        ext_stats_enabled = bool(ext_stats_include) or bool(ext_stats_exclude)
```

It probably makes sense to move the `_enabled` definition later on, to `__init__`, when following this approach.

scrapy/extensions/logstats_extended.py
```diff
@@ -0,0 +1,132 @@
...
+        ext_stats_enabled = crawler.settings.getbool("LOGSTATS_EXT_STATS_ENABLED")
+        ext_stats_include = crawler.settings.getlist("LOGSTATS_EXT_STATS_INCLUDE", [])
+        ext_stats_exclude = crawler.settings.getlist("LOGSTATS_EXT_STATS_EXCLUDE", [])
+        ext_delta_enabled = crawler.settings.getbool("LOGSTATS_EXT_DELTA_ENABLED")
+        ext_delta_include = crawler.settings.getlist("LOGSTATS_EXT_DELTA_INCLUDE", [])
+        ext_delta_exclude = crawler.settings.getlist("LOGSTATS_EXT_DELTA_EXCLUDE", [])
+        ext_timing_enabled = crawler.settings.getbool("LOGSTATS_EXT_TIMING_ENABLED")
```
Gallaecio, line 53: I think we can keep `LOGSTATS_` as the prefix as long as we make sure there are no conflicts with existing settings:

```suggestion
        ext_stats_include = crawler.settings.getlist("LOGSTATS_INCLUDE", [])
        ext_stats_exclude = crawler.settings.getlist("LOGSTATS_EXCLUDE", [])
        ext_delta_include = crawler.settings.getlist("LOGSTATS_DELTA_INCLUDE", [])
        ext_delta_exclude = crawler.settings.getlist("LOGSTATS_DELTA_EXCLUDE", [])
        ext_timing_enabled = crawler.settings.getbool("LOGSTATS_EXTENDED_TIME")
```

We could also modify the existing extension to be smart about settings specific to this new extension, e.g.
- Raise `NotConfigured` if `LOGSTATS_EXTENDED` is `True`.
- Log a warning if one of these 5 settings are defined with a non-default value, indicating that they only have effect if `LOGSTATS_EXTENDED` is set to `True`.

scrapy/extensions/logstats_extended.py
GeorgeA92: I don't have strong opinion about naming anything here.

Except(maybe) to make name.. to avoid disambiguation between this one and default `logstats` extension (assuming that we keep this as separate extension).

In next iteration I will try to use `PERIODIC_LOG_` prefix like`PERIODIC_LOG_STATS` , `PERIODIC_LOG_DELTA` etc.

scrapy/extensions/logstats_extended.py
GeorgeA92: > Instead of having 3 new _ENABLED settings, we could have a single one called LOGSTATS_FOO, where FOO would be whatever the prefix/suffix of the new class is, e.g. LOGSTATS_EXTENDED if we keep the current name.

I also have thoughts in this direction to make it work with.. 3 settings only, like this:
```
        "PERIODIC_LOG_STATS" : {
            "include": ["downloader/", "scheduler/", "log_count/", "item_scraped_count"],
            "exclude": ["scheduler/"]
        },

        "PERIODIC_LOG_DELTA": {
            "include": ["downloader/"]
        },
        "PERIODIC_LOG_TIMING_ENABLED" : True,
```
Where expected to interpret delta/stats as enabled on any non `None` setting value
It will look (conceptually) closer to conf of `FEEDS` setting

scrapy/extensions/logstats_extended.py
GeorgeA92: > Log a warning if one of these 5 settings are defined with a non-default value, indicating that they only have effect if LOGSTATS_EXTENDED is set to True.

To implement this I'd prefer to use existing `overridden settings` feature in scrapy.

If we define `PERIODIC_LOG_STATS=None; PERIODIC_LOG_DELTA=None; PERIODIC_LOG_TIMING_ENABLED=None` in scrapy/settings/default_settings.py - on run with.. non default values of this settings - it will be reflected in overridden settings log entry

Gallaecio:
You might want to [install pre-commit](https://docs.scrapy.org/en/latest/contributing.html#pre-commit) to deal with the failing CI job.

Gallaecio:
From my side the implementation looks OK, pending docs and tests.

scrapy/extensions/periodic_log.py
```diff
@@ -0,0 +1,151 @@
...
+        interval = crawler.settings.getfloat("LOGSTATS_INTERVAL")
+        try:
+            ext_stats = crawler.settings.getdict("PERIODIC_LOG_STATS")
+        except ValueError:
+            ext_stats = (
+                {"enabled": True}
+                if crawler.settings.getbool("PERIODIC_LOG_STATS")
+                else None
+            )
+        except TypeError:
```
wRAR, line 48: Shouldn't these two except clauses be combined as their code is the same?

GeorgeA92:
Updated tests and extension.
Now tests cover.. the most complicated part of extension - filtering of stats fields (`include/exclude`) by name.

scrapy/extensions/periodic_log.py
GeorgeA92: Done

GeorgeA92:
Error on my test code fixed.
each call of `ext.spider_opened(spider)` in tests - start looping call and as it didn't stopped after tests - that previously started 
 looping calls affected next tests.

Gallaecio:
@GeorgeA92 Could you look into the typing issues?

docs/topics/extensions.rst
```diff
@@ -383,3 +383,104 @@ For more info see `Debugging in Python`_.
 This extension only works on POSIX-compliant platforms (i.e. not Windows).
 
 .. _Debugging in Python: https://pythonconquerstheuniverse.wordpress.com/2009/09/10/debugging-in-python/
+
+Periodic log extension
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: PeriodicLog
+
+Extension provides extended stats data periodically in addition to basic data from Log Stats and Core Stats extensions (as JSON compatible  dictionary) like: ::
```
Gallaecio, line 392:
```suggestion
This extension periodically logs rich stat data as a JSON object::
```

docs/topics/extensions.rst
```diff
@@ -383,3 +383,104 @@ For more info see `Debugging in Python`_.
...
+``"delta"`` section shows numeric difference in stats values between current and previous log entry with period of ``LOGSTATS_INTERVAL`` (60 seconds by default). Its applicable for stats with values types ``int`` and ``float``.
+Stats values displayed in this section configured by :setting:`PERIODIC_LOG_DELTA` setting.
+
+``"stats"`` section shows stats values as is at the moment of current period.
+Stats values displayed in this section configured by :setting:`PERIODIC_LOG_STATS` setting.
+
+``"time"`` This extension produce log entries on startup, periodically, and on end of crawl. As final log entry produced earlier than ``LOGSTATS_INTERVAL`` value - detailed timing data required for more precise stats.
+
+Configured by :setting:`PERIODIC_LOG_TIMING_ENABLED`
+
```
Gallaecio, line 438:
```suggestion
This extension logs the following configurable sections:

-   ``"delta"`` shows how some numeric stats have changed since the last stats 
    log message.
    
    The :setting:`PERIODIC_LOG_DELTA` setting determines the target stats. They 
    must have ``int`` or ``float`` values.

-   ``"stats"`` shows the current value of some stats.

    The :setting:`PERIODIC_LOG_STATS` setting determines the target stats.

-   ``"time"`` shows detailed timing data.

    The :setting:`PERIODIC_LOG_TIMING_ENABLED` setting determines whether or 
    not to show this section.

This extension logs data at the start, then on a fixed time interval 
configurable through the :setting:`LOGSTATS_INTERVAL` setting, and finally 
right before the crawl ends.

```

I am suggesting formatting and wording changes, but I aimed for the content to be the same. The only conscious exception is the note on why `log_interval_real` makes sense, as I don’t think we need to mention that. I think people who care will realize why it is useful, and people who don’t care do not need to be told.

docs/topics/extensions.rst
```diff
@@ -383,3 +383,104 @@ For more info see `Debugging in Python`_.
...
+.. setting:: PERIODIC_LOG_DELTA
+
+PERIODIC_LOG_DELTA
+""""""""""""""""""
+
+Default: ``None``
+
+* ``"PERIODIC_LOG_DELTA": True`` - show deltas for all ``int`` and ``float`` stats values.
+* ``"PERIODIC_LOG_DELTA": {"include": ["downloader/", "scheduler/"]}`` - include stats deltas for stats with names that have listed substrings in stats names.
+* ``"PERIODIC_LOG_DELTA": {"exclude": ["downloader/"]}`` - include all stats deltas except stats with listed substrings in stats names.
```
Gallaecio, line 465:
```suggestion
* ``"PERIODIC_LOG_DELTA": True`` - show deltas for all ``int`` and ``float`` stat values.
* ``"PERIODIC_LOG_DELTA": {"include": ["downloader/", "scheduler/"]}`` - show deltas for stats with names containing any configured substring.
* ``"PERIODIC_LOG_DELTA": {"exclude": ["downloader/"]}`` - show deltas for all stats with names not containing any configured substring.
```

docs/topics/extensions.rst
```diff
@@ -383,3 +383,104 @@ For more info see `Debugging in Python`_.
...
+.. setting:: PERIODIC_LOG_STATS
+
+PERIODIC_LOG_STATS
+""""""""""""""""""
+
+Default: ``None``
+
+* ``"PERIODIC_LOG_STATS": True`` - show all available stats keys/values
+* ``"PERIODIC_LOG_STATS": {"include": ["downloader/", "scheduler/"]}`` - include stats for keys that have listed substrings in stats names.
+* ``"PERIODIC_LOG_STATS": {"exclude": ["downloader/"]}`` - include all stats deltas except stats with listed substrings in stats names.
```
Gallaecio, line 476:
```suggestion
* ``"PERIODIC_LOG_STATS": True`` - show the current value of all stats.
* ``"PERIODIC_LOG_STATS": {"include": ["downloader/", "scheduler/"]}`` - show current values for stats with names containing any configured substring.
* ``"PERIODIC_LOG_STATS": {"exclude": ["downloader/"]}`` - show current values for all stats with names not containing any configured substring.
```

docs/topics/extensions.rst
```diff
@@ -383,3 +383,104 @@ For more info see `Debugging in Python`_.
...
+
+
+.. setting:: PERIODIC_LOG_TIMING_ENABLED
+
+PERIODIC_LOG_TIMING_ENABLED
+"""""""""""""""""""""""""""
+
+Default: ``None``
+
+``"PERIODIC_LOG_TIMING_ENABLED": True`` - enables logging of timing data
```
Gallaecio, line 486:
```suggestion
Default: ``False``

``True`` enables logging of timing data (i.e. the ``"time"`` section).
```

scrapy/extensions/periodic_log.py
```diff
@@ -0,0 +1,140 @@
...
+                {"enabled": True}
+                if crawler.settings.getbool("PERIODIC_LOG_DELTA")
+                else None
+            )
+
+        ext_timing_enabled = crawler.settings.getbool(
+            "PERIODIC_LOG_TIMING_ENABLED", False
+        )
+        if not interval:
+            raise NotConfigured
```
Gallaecio, line 61: :lipstick: Maybe we could move this earlier, to line 40?

scrapy/extensions/periodic_log.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def log_crawler_stats(self):
+        stats = {
+            k: v
+            for k, v in self.stats._stats.items()
+            if self.param_allowed(k, self.ext_stats_include, self.ext_stats_exclude)
+        }
+        return {"stats": stats}
+
+    def param_allowed(self, stat_name, include, exclude):
```
Gallaecio, line 124: I wonder if it would make sense to cache this method.

scrapy/settings/default_settings.py
```diff
@@ -240,6 +240,10 @@
 
 NEWSPIDER_MODULE = ""
 
+PERIODIC_LOG_DELTA = None
+PERIODIC_LOG_STATS = None
+PERIODIC_LOG_TIMING_ENABLED = None
```
Gallaecio, line 245:
```suggestion
PERIODIC_LOG_TIMING_ENABLED = False
```

Gallaecio (approved):
Looks good to me. I suggest applying the doc improvements and the default_settings.py one, but the rest should not block this change.

Great work!

scrapy/extensions/periodic_log.py
GeorgeA92: @Gallaecio 
I think it is required to have several hudreds of active stats values to have performance issues from this method - this is not typical scrapy usecase. 

I think that we need wider community feedback to this extension to make decision.
To implement this it is required:
1. update method to have only 1 argument `param_allowed(self, stat_name)` to make `functools.lru_cache` a bit easier to implement/maintain.
2. implement possibility to define size of lru cache (by setting).

Gallaecio:
I just merged https://github.com/scrapy/scrapy/pull/6014, so you might want to merge from the latest main branch and update those utcnow() calls.

wRAR (approved):
Thanks!

PR #5927, opened 2023-05-09
Add additional requests examples.

bulatbulat48:
Revealed a minor gap of the documentation relating to additional requests. Added async/await syntax examples for amending the obsolete `inline_requests` library.


bulatbulat48:
@wRAR Could you help me with the error in the action? 
Can you re-run the failed pipeline?

docs/topics/asyncio.rst
```diff
@@ -98,6 +98,52 @@ Futures. Scrapy provides two helpers for this:
          into your own code.
 
 
+Async additional requests
+=========================
```
Gallaecio, line 102: :lipstick: No strong opinion, but I wonder if we could find a better title, since technically yielding a request from a callback also gives you an async additional request. Maybe “Inline requests”?

docs/topics/asyncio.rst
```diff
@@ -98,6 +98,52 @@ Futures. Scrapy provides two helpers for this:
...
+
+The spider below shows a single use-case of scraping a page and gathering a price from a separate URL::
+
```
Gallaecio, line 105:
```suggestion
The spider below shows how to send a request and await its response all from 
within a spider callback:

.. code-block:: python
```

PS: Using `.. code-block:: python` will trigger a pre-commit error until you apply black formatting to the code.

Gallaecio, line 105:
```suggestion
The spider below shows how to send a request and await its response all from 
within a spider callback:

.. code-block:: python
```

PS: Using `.. code-block:: python` will trigger a pre-commit error until you apply black formatting to the code.

docs/topics/asyncio.rst
```diff
@@ -98,6 +98,52 @@ Futures. Scrapy provides two helpers for this:
...
+            additional_request = scrapy.Request('https://example.org/price')
+            deferred = self.crawler.engine.download(additional_request)
+            additional_response = await maybe_deferred_to_future(deferred)
+            yield {
+                'h1': response.css('h1').get(),
+                'price': additional_response.css('#price').get(),
+            }
+
+
+The spider gathering batch requests::
```
Gallaecio, line 121: I would avoid using the word “batch” here, as I think it can be confusing. To me, batch requests means “send a batch of requests, wait for all of them, then send another batch, wait again”, which is not exactly what we are featuring here.

```suggestion
You can also send multiple requests in parallel:

.. code-block:: python
```

docs/topics/asyncio.rst
```diff
@@ -98,6 +98,52 @@ Futures. Scrapy provides two helpers for this:
...
+                coroutines.append(maybe_deferred_to_future(deffered))
+
+            responses = await asyncio.gather(
+                *coroutines, return_exceptions=True
+            )
+            yield {
+                'h1': response.css('h1::text').get(),
+                'price': responses[0].css('.price::text').get(),
+                'color': responses[1].css('.color::text').get(),
+            }
```
Gallaecio, line 144: I don’t think that will work with the non-asyncio Twisted reactor. Could you see of you can get a working alternative with DeferredList, in line with the following (untested) code?

```suggestion
    from twisted.internet.defer import DeferredList

    class BatchRequestsSpider(scrapy.Spider):
        name = "batch"
        start_urls = ["https://example.com/product"]

        async def parse(self, response, **kwargs):
            additional_requests = [
                scrapy.Request("https://example.com/price"),
                scrapy.Request("https://example.com/color"),
            ]
            deferreds = []
            for r in additional_requests:
                deferred = self.crawler.engine.download(r)
                deferreds.append(deferred)
            responses = await maybe_deferred_to_future(DeferredList(deferreds))
            yield {
                'h1': response.css('h1::text').get(),
                'price': responses[0].css('.price::text').get(),
                'color': responses[1].css('.color::text').get(),
            }
```

docs/topics/asyncio.rst
bulatbulat48: Duplicated.

docs/topics/asyncio.rst
bulatbulat48: Fixed. Added some fixes because `response` converted to a list of tuples.

docs/topics/asyncio.rst
bulatbulat48: Added suggestion.

docs/topics/asyncio.rst
bulatbulat48: Added suggestion.

docs/topics/asyncio.rst
bulatbulat48: Moved imports from twisted to scrapy.

wRAR: > I don’t think that will work with the non-asyncio Twisted reactor. 

That's why it's in asyncio.rst, not in coroutines.rst :P

bulatbulat48: @wRAR @Gallaecio 
As a working sample, we can add `asyncio` reactor in the example:

```
class MultipleRequestsSpider(scrapy.Spider):
    ...
    @classmethod
    def update_settings(cls, settings):
        settings["TWISTED_REACTOR"] = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"

    async def parse(self, response, **kwargs):
        ...
        responses = await asyncio.gather(
            *coroutines, return_exceptions=True
        ...
```

Gallaecio: Assuming the point is to show how to use inline requests, I think moving the examples to coroutines.rst is the way to go, instead of keeping them asyncio-specific.

Gallaecio (approved):

wRAR (approved):

PR #5929, opened 2023-05-09
Configurable exception list retry middleware 2701

serhii73:
Resolves https://github.com/scrapy/scrapy/issues/2701, closes https://github.com/scrapy/scrapy/pull/2984, closes https://github.com/scrapy/scrapy/pull/3334.


scrapy/downloadermiddlewares/retry.py
```diff
@@ -9,31 +9,38 @@
...
+    @property
+    def EXCEPTIONS_TO_RETRY(self):
+        warnings.warn(
+            "Attribute RetryMiddleware.EXCEPTIONS_TO_RETRY is deprecated. "
+            "Use the RETRY_EXCEPTIONS setting instead.",
+            ScrapyDeprecationWarning,
+            stacklevel=2,
+        )
+        return tuple(
+            load_object(x) if isinstance(x, six.string_types) else x
```
wRAR, line 39:
```suggestion
            load_object(x) if isinstance(x, str) else x
```

scrapy/downloadermiddlewares/retry.py
```diff
@@ -9,31 +9,38 @@
...
-    ConnectionRefusedError,
-    DNSLookupError,
-    TCPTimedOutError,
-    TimeoutError,
-)
-from twisted.web.client import ResponseFailed
-
-from scrapy.core.downloader.handlers.http11 import TunnelError
-from scrapy.exceptions import NotConfigured
+import six
```
wRAR, line 16: We shouldn't use six as our code doesn't support Python 2.

scrapy/downloadermiddlewares/retry.py
```diff
@@ -121,23 +128,7 @@ def parse(self, response):
...
-        ConnectionDone,
-        ConnectError,
-        ConnectionLost,
-        TCPTimedOutError,
-        ResponseFailed,
-        IOError,
-        TunnelError,
-    )
-
+class RetryMiddleware(six.with_metaclass(BackwardsCompatibilityMetaclass, object)):
```
wRAR, line 131: I *think* this should be `class RetryMiddleware(metaclass=BackwardsCompatibilityMetaclass)`

scrapy/downloadermiddlewares/retry.py
```diff
@@ -147,6 +138,16 @@ def __init__(self, settings):
         )
         self.priority_adjust = settings.getint("RETRY_PRIORITY_ADJUST")
 
+        if not hasattr(
+            self, "EXCEPTIONS_TO_RETRY"
+        ):  # If EXCEPTIONS_TO_RETRY is not "overriden"
+            self.exceptions_to_retry = tuple(
+                load_object(x) if isinstance(x, six.string_types) else x
```
wRAR, line 145:
```suggestion
                load_object(x) if isinstance(x, str) else x
```

wRAR: Actually in both these lines just `load_object(x)` should work, as `load_object()` works with any classes and just returns them.

tests/test_downloadermiddleware_retry.py
```diff
@@ -110,6 +110,29 @@ def test_twistederrors(self):
...
+        # Retry
+        self.mw.exceptions_to_retry = self.mw.exceptions_to_retry + (exc,)
+        req = Request("http://www.scrapytest.org/%s" % exc.__name__)
+        self._test_retry_exception(req, exc("foo"))
+
+    def test_exception_to_retry_customMiddleware(self):
+        exc = ValueError
+
+        class MyRetryMiddleware(RetryMiddleware):
+            EXCEPTIONS_TO_RETRY = RetryMiddleware.EXCEPTIONS_TO_RETRY + (exc,)
```
wRAR, line 128: As this test uses a deprecated attribute it should also check that the warning is produced.

tests/test_downloadermiddleware_retry.py
```diff
@@ -110,6 +110,29 @@ def test_twistederrors(self):
...
+    def test_exception_to_retry_added(self):
+        # Don't retry
+        exc = ValueError
+        req = Request(f"http://www.scrapytest.org/{exc.__name__}")
+        req = self.mw.process_exception(req, exc("foo"), self.spider)
+        self.assertEqual(req, None)
+        # Retry
+        self.mw.exceptions_to_retry = self.mw.exceptions_to_retry + (exc,)
+        req = Request(f"http://www.scrapytest.org/{exc.__name__}")
+        self._test_retry_exception(req, exc("foo"))
```
Gallaecio, line 122: It would be best to test adding it the way users should add it, e.g. (untested)

```python
from scrapy.settings.default_settings import RETRY_EXCEPTIONS
settings = {
    "RETRY_EXCEPTIONS": RETRY_EXCEPTIONS + (ValueError,),
}
crawler = get_crawler(Spider, settings=settings)
mw = RetryMiddleware.from_crawler(crawler)
req = Request(f"http://www.scrapytest.org/{exc.__name__}")
self._test_retry_exception(req, exc("foo"), mw)  # Modify to accept an optional mw
```

docs/topics/downloader-middleware.rst
```diff
@@ -966,6 +967,34 @@ In some cases you may want to add 400 to :setting:`RETRY_HTTP_CODES` because
...
+        IOError,
+        'scrapy.core.downloader.handlers.http11.TunnelError',
+    ]
+
+List of exceptions to retry.
+
+It may contain either an exception or a string, in which case Scrapy would load
+the exception from the absolute object path specified by the string. 
+If an exception is not retried earlier, middlewares get to process it; 
+if none handles it, the request errback gets it.
```
Gallaecio, line 996: I would recommend something in the line of:

```suggestion
Each list entry may be an exception type or its import path as a string.

An exception will not be caught when the exception type is not in
:setting:`RETRY_EXCEPTIONS` or when the maximum number of retries for a request
has been exceeded (see :setting:`RETRY_TIMES`). To learn about uncaught
exception propagation, see
:meth:`~scrapy.downloadermiddlewares.DownloaderMiddleware.process_exception`.
```

The reference documentation of `process_exception` is not clear about `request.errback` getting the exception if unhandled by other middlewares, or how exceptions are propagated to earlier middlewares (i.e. closer to the spider), but I consider that out of the scope of these changes.

tests/test_downloadermiddleware_retry.py
```diff
@@ -110,7 +112,34 @@ def test_twistederrors(self):
...
+
+            self.assertEqual(len(warns), 1)
+
+        mw2 = MyRetryMiddleware.from_crawler(self.crawler)
+        req = Request(f"http://www.scrapytest.org/{exc.__name__}")
+        req = mw2.process_exception(req, exc("foo"), self.spider)
+        assert isinstance(req, Request)
+        self.assertEqual(req.meta["retry_times"], 1)
+
+    def _test_retry_exception(self, req, exception, mw=None):
```
Gallaecio, line 142: `mw` is unused here.

tests/test_downloadermiddleware_retry.py
```diff
@@ -110,19 +112,49 @@ def test_twistederrors(self):
             == 2
         )
 
-    def _test_retry_exception(self, req, exception):
+    def test_exception_to_retry_added(self):
+        exc = ValueError
+        self.mw.exceptions_to_retry = self.mw.exceptions_to_retry + (exc,)
```
Gallaecio, line 117: I suspect this line is pointless and can be removed completely. And maybe `ValueError` can be replaced with `exc` in the line below, to avoid repeating `ValueError`.

Gallaecio (approved):

Gallaecio:
Niiice!

PR #5931, opened 2023-05-12
Improve the overwrite feed option documentation

### LINKED ISSUE #5579: Improve the overwrite feed option documentation

belaeifert:
Fixes #5579, closes #5593

Adds to the documentation the description of the default value of the overwrite option in feeds for the Google Cloud Storage backend.

Adds alerts about the risk of losing your data for storage backends that have this default value equal to true.

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -175,6 +175,12 @@ FTP supports two different connection modes: `active or passive
 mode by default. To use the active connection mode instead, set the
 :setting:`FEED_STORAGE_FTP_ACTIVE` setting to ``True``.
 
+The default value for the ``overwrite`` key in the :setting:`FEEDS` for this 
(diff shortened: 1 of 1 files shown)


serhii73 (approved):

wRAR (approved):

wRAR:
Thanks!

Issue #5707, opened 2022-11-08
Possibly contradictory documentation about propagation of `Response.request`


henribru:
### Description

![image](https://user-images.githubusercontent.com/6639509/200525681-25629784-3146-4ae3-ad54-3b607e195ceb.png)

The docs for https://docs.scrapy.org/en/2.7/topics/request-response.html#scrapy.http.Response.request and https://docs.scrapy.org/en/2.7/topics/request-response.html#scrapy.http.Response.meta seem contradictory. The docs for `request` say that for a redirect, the original request is assigned to the final response, but then the docs for `meta` say that **unlike Response.request**, for a redirect you get the original `meta`. If the first part is true the "unlike" part seems incorrect and if the "unlike" part is correct the first part seems incorrect. Is this a mistake in the docs or am I simply misreading it? If it's a mistake it should obviously be corrected, but even if it's not it might need a clarification. If I am indeed misreading it, what is actually meant by this?

### Steps to Reproduce

N/A (since it's documentation related)

**Expected behavior:** 
N/A (since it's documentation related)

**Actual behavior:**
N/A (since it's documentation related)

**Reproduces how often:**
N/A (since it's documentation related)

### Versions
2.7

### Additional context

Gallaecio:
I think the request part may be wrong.

When a request redirects, the redirection middleware basically creates a new Request object from the original Request object, which keeps the same meta, and most of the same attributes, but is a different Python object and the URL is the Response.url (to find the original URL you actually need to inspect Request.metadata filled by the redirect middleware).

Shershon25:
Hey! I would like to explore this issue. Should I find and produce the code that would clear up the `request` docs?

Gallaecio:
@Shershon25 Please give it a go, and feel free to ask any question.

Shershon25:
As @Gallaecio mentioned, a new `request` object will be created when a redirection request is redirection response is received at the`redirect` class of the `downloadermiddlewares` module.

```
def _build_redirect_request(source_request, *, url, **kwargs):
    redirect_request = source_request.replace(
        url=url,
        **kwargs,
        cookies=None,
    )
```
The replace method does return a new Request object and hence the document is misleading in this case. Some attributes of the original request are assigned to the redirected response, not the request as a whole.

Also won't a few `meta` attributes(redirect_ttl, redirect_times) get changed in case of redirections/retries?

Shershon25:
Hey! Shouldn't we update the docs regarding this?

[...]

PR #5937, opened 2023-05-20
Change redirect text from Response.request docs

medic-code:
This is my first PR for Scrapy so any feedback would be welcome. Having looked at https://github.com/scrapy/scrapy/issues/5707, I've changed the text to reflect the comments in that issue and from my own reading of the code. Was a nice exercise in getting a little more familiar with the code base, learned a few things. 

Fixes https://github.com/scrapy/scrapy/issues/5707


Gallaecio:
Nice, thanks!

PR #5939, opened 2023-05-26
Adding support for windows absolute pathlib.Path objects in FeedExporter

alexpdev:
This PR is related to #5739  as well as some conflicting information in the docs.

On the [Feed Exports](https://docs.scrapy.org/en/latest/topics/feed-exports.html#topics-feed-storage-fs) page, in the [FEEDS](https://docs.scrapy.org/en/latest/topics/feed-exports.html#feeds) subsection it mentions that you can use a URI or a `pathlib.Path` object as the target.

https://github.com/scrapy/scrapy/blob/33b418dc846c8a0de21bef7a71c7ed259e24b269/docs/topics/feed-exports.rst?plain=1#L412-L423

Earlier on the same page however under the [Storage Backends](https://docs.scrapy.org/en/latest/topics/feed-exports.html#local-filesystem) section it mentions that you can use a standard string path as long as its an absolute path, but only on Unix systems.

https://github.com/scrapy/scrapy/blob/33b418dc846c8a0de21bef7a71c7ed259e24b269/docs/topics/feed-exports.rst?plain=1#L149-L161

Both of these statements are accurate however, if one were to miss the second statement and only read the first you might think, as I did, that using an absolute `pathlib.Path` object on windows would work fine, when in fact it doesn't.  For example using a `Path(__file__).parent / "items.json"` fails when used on windows OS.  

My proposed solution to this, which is implemented in this PR is to simply get rid of the "only on Unix" caveat and make it work for windows as well.  I achieved this by adding a conditional statement in the `FeedExporter.__init__` method that tests if the uri used in the settings is a `Path` object, and if it is then to convert it to a filesystem uri. 

I also included an accompanying unit test that would fail prior to the changes.

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -156,8 +156,7 @@ The feeds are stored in the local filesystem.
 -   Required external libraries: none
 
 Note that for the local filesystem storage (only) you can omit the scheme, and
-use an absolute or relative filesystem path (or a :class:`pathlib.Path`
-object) instead.
+use an absolute filesystem path (or a :class:`pathlib.Path` object) instead.
 
 .. _topics-feed-storage-ftp:
 
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -156,7 +156,8 @@ The feeds are stored in the local filesystem.
 -   Required external libraries: none
 
 Note that for the local filesystem storage (only) you can omit the scheme, and
-use an absolute filesystem path (or a :class:`pathlib.Path` object) instead.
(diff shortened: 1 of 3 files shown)


wRAR:
Nice, should the doc be updated though?

alexpdev:
> Nice, should the doc be updated though?

I don't think so.  I think both of those statements are still True.  Because you still cannot use a string absolute path on windows.  This only allows for using the absolute `pathlib.Path`.

alexpdev:
can we re-run the test that failed?  It passed last time.

wRAR (approved):

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5943, opened 2023-06-02
Add Content-Encoding header in response flag

hassaan-naushahi:
Fixes #1988, closes #4025, closes #5290, closes #6156, closes #8149.


wRAR (changes requested):
Two tests are failing now, one because there is a new warning which it doesn't expect and another one probably because the header is dropped by default?

This PR is also missing the default setting value, docs and tests for the new setting. Some of those already exist in #5290 so it should be possible to copy them (though I haven't checked the test changes there).

hassaan-naushahi:
Added values to default_settings.py for successful test runs

scrapy/settings/default_settings.py
```diff
@@ -37,7 +37,7 @@
 COMMANDS_MODULE = ""
 
 COMPRESSION_ENABLED = True
-
```
wRAR, line 40: You should keep the empty line before `CONCURRENT_ITEMS`

wRAR:
Can you please always run tests locally before publishing?

wRAR:
I've fixed the conflicts and removed an extra `if not stats` block in `__init__()` which looked like a mistake to me. The failing tests are on the first glance the same as ones that were failing before I merged master.

farzamfallahi:
Hi @hassaan-naushahi @AdrianAtZyte, I've been reading through the history of this issue and PR. It looks very close now — 42 of 43 checks pass, with only the Windows no-reactor test failing, plus a merge conflict that needs resolving. Would it be okay if I helped dig into either of those, or is someone already on it?

farzamfallahi:
I looked into the failing Windows no-reactor check. It's unrelated to this PR's changes — the failure is in tests/test_downloaderslotssettings.py::TestCrawl::test_delay, which asserts download delay timing stays within a 30% tolerance. It failed with a 71% timing error on the 127.0.0.1 mock server, which looks like CI runner timing variance rather than a real regression:

AssertionError: assert 0.7138123760557606 < 0.3

Re-running the job would likely resolve it, but I don't have permission to trigger that here. Happy to help with anything else needed to get this merged.

PR #5946, opened 2023-06-09
[WIP] Refactor cookie handling in CookiesMiddleware

emarondan:
- Refactoring cookie handling in CookiesMiddleware 
- Adding test case for off-domain jar storage


scrapy/downloadermiddlewares/cookies.py
```diff
@@ -40,13 +40,11 @@ def _process_cookies(self, cookies, *, jar, request):
                 cookie_domain = cookie_domain[1:]
 
             request_domain = urlparse_cached(request).hostname.lower()
-
-            if cookie_domain and _is_public_domain(cookie_domain):
+            if _is_public_domain(cookie_domain):
                 if cookie_domain != request_domain:
                     continue
-                cookie.domain = request_domain
```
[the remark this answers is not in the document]
emarondan: I just re-checked this and everything works the same only changing `jar.set_cookie_if_ok(cookie, request)` by `jar.set_cookie(cookie)` and I don't think that is the idea for this issue.

scrapy/downloadermiddlewares/cookies.py
```diff
@@ -43,10 +43,11 @@ def _process_cookies(self, cookies, *, jar, request):
 
             if cookie_domain and _is_public_domain(cookie_domain):
                 if cookie_domain != request_domain:
-                    continue
-                cookie.domain = request_domain
-
-            jar.set_cookie_if_ok(cookie, request)
+                    cookie.domain = request_domain
```
emarondan, line 46: @wRAR With this modification the cookie is always set on the jar. The cookie domain must be the same as the request one in order to be set by `jar.set_cookie_if_ok()`

scrapy/downloadermiddlewares/cookies.py
```diff
@@ -43,10 +43,11 @@ def _process_cookies(self, cookies, *, jar, request):
...
+                jar.set_cookie_if_ok(cookie, request)
+            else:
+                print(f'seting cookie {cookie.__dict__}')
+                jar.set_cookie(cookie)
```
emarondan, line 50: If the domain is not public, the cookie should be set anyway.

tests/test_downloadermiddleware_cookies.py
```diff
@@ -646,7 +646,7 @@ def test_user_set_cookie_domain_suffix_public_period(self):
             "https://foo.co.uk",
             "https://bar.co.uk",
             "co.uk",
-            cookies1=False,
+            cookies1=True,
```
emarondan, line 649: With the modification made on the handler, I think this test should change since the cookie will be present on the first request.

tests/test_downloadermiddleware_cookies.py
```diff
@@ -655,7 +655,7 @@ def test_user_set_cookie_domain_suffix_public_private(self):
             "https://foo.blogspot.com",
             "https://bar.blogspot.com",
             "blogspot.com",
-            cookies1=False,
+            cookies1=True,
```
emarondan, line 658: Same as above test.

scrapy/downloadermiddlewares/cookies.py
wRAR: I'm trying to understand why the logic is inverted by this change.

Gallaecio:
I don’t think the changes to existing tests are correct. I think we need to change the implementation instead to make sure they continue to pass as they were.

Also, the cookie processing method affects both requests and responses, so it would be good to make sure we have a test to make sure that, if we get a Set-Cookie header from a request with a cookie set for an unrelated domain, that the cookie is *not* added to the cookiejar. We trust users, not servers.

tests/test_downloadermiddleware_cookies.py
```diff
@@ -732,3 +732,30 @@ def test_server_set_cookie_domain_public_period(self):
...
+        self.assertEqual(request2.headers.get(b"Cookie"), b"foo=bar")
+
+    def test_process_response_unrelated_domain(self):
+        request = Request("http://a.example")
+        request.headers["Cookie"] = "foo=bar; domain=b.example"
+        response = Response("http://c.example")
+        response.headers["Set-Cookie"] = "asd=fgh; domain=d.example"
+        self.mw.process_response(request, response, spider=None)
+        jar = self.mw.jars[None]
+        assert not jar._cookies.get("c.example", {}).get("/", {}).get("asd")
```
Gallaecio, line 761: The problem with the new approach is, I think, that it also applies to server-side cookies.

I think this test is not right, and fixing it as follows:

```suggestion
        response.headers["Set-Cookie"] = "asd=fgh; domain=c.example"
        self.mw.process_response(request, response, spider=None)
        jar = self.mw.jars[None]
        assert not jar._cookies.get("c.example", {}).get("/", {}).get("asd")
```

Will probably make it fail with the new approach.

It would not be very clean, but I wonder if it would be possible to change the policy right before setting user-defined cookies (and only user-defined cookies, not Set-Cookie headers from servers), and restore it right after.

If not that way, we need to think of something else.

Gallaecio:
@emarondan I have made a small change to `test_process_response_unrelated_domain`, to cover the issue I fear we might introduce here. When receiving a cookie in a response for a different domain, the cookie should be ignored altogether; we should not add it to the cookie jar, not for the actual response domain, not for the cookie domain.

Please, see if you can find a way, however unclean, to get `test_off_domain_jar_storage` to pass without breaking the new `test_process_response_unrelated_domain` in the process.

PR #5949, opened 2023-06-13
Allow running tests against the installed version

### LINKED ISSUE #4914: As-installed tests support
Debian supports testing installed packages by running a test suite against them. Doing that for Scrapy is currently impossible, it contains explicit protection against this: [`scrapy.utils.test.get_pythonpath()`](https://github.com/scrapy/scrapy/blob/212163e199d5977095dfcd612181762e359ab8b4/scrapy/utils/test.py#L69) ensures `PYTHONPATH` for external scripts (including mockserver) points to the same instance of Scrapy. So:

- if you run pytest in the Scrapy repo, tests will import `scrapy.utils` from the current dir and run `python3 -m tests.mockserver` with PYTHONPATH set to the current dir, testing the local code
- if you also install Scrapy, nothing will change
- if you remove the `scrapy` dir, which is what is usually done to test the installed code, tests will import `scrapy.utils` from the system directory, set PYTHONPATH to the system directory and `python3 -m tests.mockserver` will fail because the `tests` package is not installed.

I know some Python modules also install the `tests` package but I don't think this is a good idea. Maybe `get_pythonpath()` should learn how to find the instance we want, not sure how to do that though.

wRAR:
Fixes #4914 by using get_testenv() only for `python3 -m scrapy.cmdline` and finding files under `tests` via different mechanisms.

For some reason running these as-installed tests in the Debian autopkgtest chroot always fails on `tests/test_crawl.py::CrawlTestCase::test_start_requests_laziness` with "All start requests consumed before any download happened". I've looked at the test and the spider it runs and I don't understand if it just relies on mockserver replying faster than start_requests schedules requests or there is something more stable. When I extracted this spider and ran it via `scrapy crawlspider` on a separately started mockserver it also often (but not always) fails with the same problem. It's possible that the test is just flaky and depends on performance of various components, but I don't remember if it fails on some CI runs.

### DIFF
--- a/tests/mockserver.py
+++ b/tests/mockserver.py
@@ -1,11 +1,13 @@
 import argparse
 import json
+import os
 import random
 import sys
 from pathlib import Path
 from shutil import rmtree
 from subprocess import PIPE, Popen
 from tempfile import mkdtemp
+from typing import Dict
 from urllib.parse import urlencode
 
 from OpenSSL import SSL
@@ -20,7 +22,6 @@ from twisted.web.static import File
 from twisted.web.util import redirectTo
 
 from scrapy.utils.python import to_bytes, to_unicode
-from scrapy.utils.test import get_testenv
 
 
 def getarg(request, name, default=None, type=None):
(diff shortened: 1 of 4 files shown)


Gallaecio (approved):
:heart:

wRAR:
Actually `test_start_requests_laziness` failed in build-time tests too, so whatever it is, it's unrelated to these changes and this mode of running tests, so merging this.

PR #1272, opened 2015-06-01
Add-ons

jdemaeyer:
Based on ~~#1149 and #1423~~ ~~#1586~~ master

Closes #591 and #1215

Implementation of SEP-021 ([updated version from this PR](https://github.com/jdemaeyer/scrapy/blob/enhancement/addons/sep/sep-021.rst), [original version](https://github.com/scrapy/scrapy/blob/master/sep/sep-021.rst))

[**Design decision to-do here**](https://github.com/scrapy/scrapy/pull/1272#issuecomment-152191871)

**Old to-do:**
- Design decisions
  - [x] Configuration entry point (settled on `INSTALLED_ADDONS` setting and additional support for `scrapy.cfg` so far)
  - [x] Integration into start-up process
  - [x] Add-on interactions (can add-ons configure other add-ons?)
  - [x] Find better name for `AddonManager`?
  - [x] Update SEP
- Code
  - [x] Add-on interface
  - [x] Base `Addon` class
  - [x] Load add-on configuration from `scrapy.cfg`
  - [x] Load add-on configuration from settings module
  - [x] Search add-ons by name
    - [x] Improve `get_project_path()` robustness (**done partially**, still fails when settings module does not live in project's root)
  - [x] Load & verify add-ons
  - [x] Add-on "holding" (a place where loaded add-ons live)
  - [x] Helper functions to call add-on callbacks
  - [x] Dependency clash management
    - [x] ~~Add version support~~ Use `pkg_resources`' extensive dependency management
    - [x] ~~Should `pkg_resources` be used throughout the manager and not just within the dependency checking?~~ Will go to separate PR if we decide to
  - [x] Integrate into standard start-up process
    - [x] Integrate dependency clash checks
  - [x] Allow putting objects (not just class paths) in component lists, e.g. `PIPELINES` (check #1215)
  - [x] Write add-ons for built-in components
    - [x] Fix add-ons for components which do not prepend their settings with their name
  - [x] ~~Move settings for built-in components from global namespace to add-ons (is this possible without making them harder to configure?)~~ Allow setting global namespace settings via add-on framework
  - [x] Allow spider to implement add-on callbacks (like it already does with `update_settings()`)
- Tests
  - [x] Base `Addon` class
  - [x] Loading configuration from `cfg` files
  - [x] Loading configuration from Python modules
  - [x] Search add-ons
  - [x] Load and verify add-ons
  - [x] Callback helper functions
  - [x] Dependency clash management
  - [x] Integration into start-up process
[...]

PR #5950, opened 2023-06-14
Add-ons

### LINKED ISSUE #5992: NotConfigured logging breaks when the component is added by class object
As the log message for components that raise `NotConfigured` with a message assumes `clsname` is an import path string, it raises an AttributeError when it's a class instance. https://github.com/scrapy/scrapy/blob/bddbbc522aef00dc150e479e6288041cee2e95c9/scrapy/middleware.py#L49

wRAR:
This is #1272, updated to the current code and with builtin-addons removed. There should be some unanswered questions in that PR. I've addressed some of the simplest feedback and made the CI pass, though there are still documentation references to file paths and there may be other things that need to be fixed. 

I haven't thought much about the implementation itself yet, but I already have a question "how can a user override a priority of a middleware set by an addon or replace that middleware with their own subclass of it".

Closes: #1272, fixes #5992


Gallaecio:
I’ve had a look at the changes here and read a few of the first messages in https://github.com/scrapy/scrapy/pull/1272 from @kmike and @curita (and @nyov’s [valuable language feedback](https://github.com/scrapy/scrapy/pull/1272#discussion_r34540682) :stuck_out_tongue_winking_eye:), and I keep thinking that we could go for a simpler add-on implementation:

-   Define an ``ADDONS`` setting as a list of components (``Dict[Type, int]``)  that have an ``update_settings`` method.

-   Have add-on’s ``update_settings`` be called right after the spider class ``update_settings``.

    This allows add-ons to be defined and configured on a spider class, through ``custom_settings`` and ``update_settings``.

    I also considered cases where a spider class may need to do something *after* add-on configuration, e.g. to work around add-on mess-ups, but then I realized that you can always create a custom add-on and place it last in the ``ADDONS`` setting to achieve that.

-   *Maybe* provide some utility functions to edit settings from ``update_settings`` in ways common for add-on scenarios, which could also come in handy for implementations of ``update_settings`` in a spider class.

-   *Maybe* allow defining a method or handling a signal for add-ons to react to Scrapy components raising NotConfigured or some other exception? (https://github.com/scrapy/scrapy/pull/1272#issuecomment-108393927, 2nd bullet point)

-   I *wonder* if we need a ``check_configuration`` method at all, and for which scenarios.

About anything more that than, I would like to understand how important and useful it would be, and whether or not it needs to be part of the initial implementation or can be postponed to future, backward-compatible iterations.

wRAR:
After reading the current code and later comments (starting from https://github.com/scrapy/scrapy/pull/1272#issuecomment-152191871) I think the following features are present in addition to the @Gallaecio's comment:

* addons (and actually other user code) adding/removing/enabling/disabling/configuring other addons
* 3 separate ways to specify which settings the addon modifies (a simple dict for settings.update(), a dict of addon-specific settings with prefixed names, a single pipeline/middleware/whatever to add to the relevant dict)
* version requirements that are only used to check if the enabled combination is valid, not to e.g. enable required things automatically

At least some of this can be dropped and re-added later.

wRAR:
Oh, and runtime checking via zope.interface is another feature. And I think it's not very useful when it checks only presence of `name` and `version`, especially if we also drop `version`. We could make `update_settings` mandatory but I don't know how useful is that.



So, mostly in line with the @Gallaecio's proposal above, I currently think we should drop everything related to versions, interface things, and maybe drop or simplify (some of) the helpers/attribute used in update_settings().

kmike:
I largely agree with @Gallaecio; it seems better to aim for a minimal implementation: addon is a Python class with 2 methods: update_settings and check_final_settings (check_configuration?). No modules, no updating other addons, no provides/requires/etc. I'll comment inline about the features which we can probably drop.

---

I think one of the possible benefits of addons is not realized in this PR: addons know which settings they have. So, if instead of 

```py
ADDONS = {MyAddon: 100}
MY_ADDON_PARAMETER = "foo"
```

we have something like this:

```py
ADDONS = {
    MyAddon: {
        "priority": 100,  # not sure about this
        "parameter": "foo"
    },
}
```

we can detect if incorrect option is passed to the addon, while MY_ADDON_PARAMTER is always going to be silently ignored.

On the other hand, this approach is not compatible with changing individual settings via cmdline or Scrapy Cloud, so it requires more work to become usable. So, in a spirit of simplification, we can consider it as another idea for the future :)

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+    SOMEADDON = {
+        'some_config': True,
+    }
+
+Enabling and configuring add-ons within Python code
+---------------------------------------------------
+
+The :class:`~scrapy.addons.AddonManager` will only read from Scrapy's settings
+*at the beginning* of Scrapy's start-up process.
```
kmike, line 44: This is not precise; it's not about Scrapy start-up, it's about spider/crawler startup.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+Afterwards, i.e. as soon as the :class:`~scrapy.addons.AddonManager` is
+populated, changing the ``ADDONS`` setting or any of the add-on
+configuration dictionary settings will have no effect.
```
kmike, line 47: I'm not sure why we need to document this explicitly. Isn't it how most or all other settings work?

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+If you want to enable, disable, or configure add-ons in Python code, for example
+when writing your own add-on, you will have to use the
+:class:`~scrapy.addons.AddonManager`. You can access the add-on manager through
+either ``crawler.addons`` or, if you are writing an add-on, through the
+``addons`` argument of the :meth:`update_addons` callback. The add-on manager
+provides many useful methods and attributes to facilitate interacting with the
+add-ons framework, e.g.:
```
kmike, line 55: I think we can drop this feature, and consider adding it if there is a real use case.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+Writing your own add-ons
+========================
+
+Add-ons are (any) Python *objects* that provide Scrapy's *add-on interface*.
+The interface is enforced through ``zope.interface``. This leaves the choice of
+Python object up the developer. Examples:
+
+* for a small pipeline, the add-on interface could be implemented in the same
+  class that also implements the ``open/close_spider`` and ``process_item``
+  callbacks,
```
kmike, line 83: In my experience this is an anti-pattern to have a single class act like different types of Scrapy components. The issue is that Scrapy would create multiple objects of this class; in this case, one instance would be used as an addon, and another as an extension. This can lead to confusion. If there is any work done during initialization ( even logging), it'd be done twice.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+* for larger add-ons, or for clearer structure, the interface could be provided
+  by a stand-alone module.
```
kmike, line 85: I'd say by a stand-alone class.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+The absolute minimum interface consists of two attributes:
+
+.. attribute:: name
```
kmike, line 89: Not sure we need this

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+    :class:`~scrapy.settings.Settings` object as wished, e.g. enable components
+    for this add-on or set required configuration of other extensions.
+
+    :param config: Configuration of this add-on
+    :type config: ``dict``
+
+    :param settings: The settings object storing Scrapy/component configuration
+    :type settings: :class:`~scrapy.settings.Settings`
+
+.. method:: check_configuration(config, crawler)
```
kmike, line 115: That's actually interesting; I wonder why does it accept crawler instead of settings, i.e. why does it check crawler state, not the final settings.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+    fully initialized, immediately before it starts crawling. You can perform
+    additional dependency and configuration checks here.
+
+    :param config: Configuration of this add-on
+    :type config: ``dict``
+
+    :param crawler: Fully initialized Scrapy crawler
+    :type crawler: :class:`~scrapy.crawler.Crawler`
+
+.. method:: update_addons(config, addons)
```
kmike, line 127: I think we can drop it, and add if there is a real use case.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+    extension that provides generic read/write access to a MongoDB database)
+
+The entries in the :attr:`requires` and :attr:`modifies` attributes can be add-on
+names or components from other add-ons' :attr:`provides` attribute. You can
+specify :pep:`440`-style information about required versions. Examples::
+
+    requires = ['httpcache']
+    requires = ['otheraddon >= 2.0', 'yetanotheraddon']
+
+The Python object or module that is pointed to by an add-on path (e.g. given in
```
kmike, line 166: I think we should support only Python objects, and only allow addons as classes.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+the ``ADDONS`` setting, or given to
+:meth:`~scrapy.addons.AddonManager.add`) does not necessarily have to be an
+add-on. Instead, it can provide an ``_addon`` attribute. This attribute can be
```
kmike, line 169: I'd remove support for this `_addons` attribute.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+either an add-on or another add-on path.
+
+
+Add-on base class
+=================
+
+Scrapy comes with a built-in base class for add-ons which provides some
+convenience functionality:
+
+* basic settings can be exported via :meth:`~scrapy.addons.Addon.export_basics`,
```
kmike, line 179: What does "exported" mean here?

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+  :meth:`~scrapy.addons.Addon.export_component`, configurable via
+  :attr:`~scrapy.addons.Addon.component_type`,
+  :attr:`~scrapy.addons.Addon.component_key`,
+  :attr:`~scrapy.addons.Addon.component`, and the ``order`` key in
+  :attr:`~scrapy.addons.Addon.default_config`.
+* the add-on configuration can be exposed into Scrapy's settings via
+  :meth:`~scrapy.addons.Addon.export_config`, configurable via
+  :attr:`~scrapy.addons.Addon.default_config`,
+  :attr:`~scrapy.addons.Addon.config_mapping`, and
+  :attr:`~scrapy.addons.Addon.settings_prefix`.
```
kmike, line 192: I'm not sure we need this.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+By default, the base add-on class will expose the add-on configuration into
+Scrapy's settings namespace, in caps and with the add-on name prepended. It is
```
kmike, line 195: It seems more flexible to do it without addon name prepended.

kmike: For example: you make a sublcass of an existing addon, which configures a certain extension. Suddenly the configuration is all wrong, because the extension expects old setting names.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+    from scrapy.addons import Addon
+
+    class MyAddon(Addon):
+        name = 'myaddon'
+        version = '1.0'
+        component = 'path.to.mypipeline'
+        component_type = 'ITEM_PIPELINES'
+        component_order = 200
+        basic_settings = {
+            'DNSCACHE_ENABLED': False,
```
kmike, line 221: the example doesn't work according to docs, as the setting is going to be exported as MY_ADDON_DNSCACHE_ENABLED

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+    class MyAddon(Addon):
+        name = 'myaddon'
+        version = '1.0'
+
+        def update_settings(self, config, settings):
+            try:
+                import boto
+            except ImportError:
+                raise RuntimeError("myaddon requires the boto library")
```
kmike, line 236: I wonder what's the benefit of checking it here, as opposed to checking it in the component's `__init__`. If it's checked in the component, we get the same error behavior when addons system is not used.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+Enable a component that lives relative to the add-on (see
+:ref:`topics-api-settings`)::
+
+    from scrapy.addons import Addon
+
+    class MyAddon(Addon):
+        name = 'myaddon'
+        version = '1.0'
+        component = __name__ + '.downloadermw.coolmw'
```
kmike, line 247: this is weird :) Not sure why you'd like to do it. Just import the component?

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+
+    class MyAddon(object):
+        name = 'myaddon'
+        version = '1.0'
+
+        def update_settings(self, config, settings):
+            mysqlpl = MySQLPipeline(password=config['password'])
+            settings.set(
+                'ITEM_PIPELINES',
+                {mysqlpl: 200},
```
kmike, line 263: Hm, I don't think this example works at all; ITEM_PIPELINES expects pipeline classes, not pipeline instances as keys.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+                priority='addon',
+            )
+
+Provide add-on interface along component interface::
```
kmike, line 267: Anti-pattern, as 2 instances would be created.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+            return item
+
+        def update_settings(self, config, settings):
+            settings.set(
+                'ITEM_PIPELINES',
+                {self: 200},
+                priority='addon',
+            )
+
+Enable another addon (see :ref:`topics-api-addonmanager`)::
```
kmike, line 284: Not sure we need it.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+        def update_settings(self, config, settings):
+            settings.set('DNSCACHE_ENABLED', False, priority='addon')
+
+        def check_configuration(self, config, crawler):
+            if crawler.settings.getbool('DNSCACHE_ENABLED'):
+                # The spider, some other add-on, or the user messed with the
+                # DNS cache setting
+                raise ValueError("myaddon is incompatible with DNS cache")
+
+Provide add-on interface through a module:
```
kmike, line 310: Not sure why it's needed.

docs/topics/addons.rst
```diff
@@ -0,0 +1,348 @@
...
+            {MyPipeline(): 200},
+            priority="addon",
+        )
+        settings.set(
+            "DOWNLOADER_MIDDLEWARES",
+            {MyDownloaderMiddleware(): 800},
+            priority="addon",
+        )
+
+Forward to other add-ons depending on Python version::
```
kmike, line 338: This can be done in other ways; also, the example uses six.PY3, and we don't support Python 2 for a long time. I'd drop this feature and the example as well.

scrapy/addons.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        self.export_component(config, settings)
+        self.export_basics(settings)
+        self.export_config(config, settings)
+
+
+class AddonManager(Mapping):
+    """This class facilitates loading and storing :ref:`topics-addons`.
+
+    You can treat it like a read-only dictionary in which keys correspond to
```
kmike, line 150: it's not really read-only if `__delitem__` is supported

scrapy/addons.py
```diff
@@ -0,0 +1,425 @@
...
+
+        :param settings: Crawler settings object
+        :type settings: :class:`~scrapy.settings.Settings`
+        """
+        self.export_component(config, settings)
+        self.export_basics(settings)
+        self.export_config(config, settings)
+
+
+class AddonManager(Mapping):
```
kmike, line 147: If we proceed with removal of various features, we should check which code in the AddonManager is still valuable.

scrapy/crawler.py
```diff
@@ -56,6 +57,7 @@ def __init__(
         spidercls: Type[Spider],
         settings: Union[None, dict, Settings] = None,
         init_reactor: bool = False,
+        addons: Optional[AddonManager] = None,
```
kmike, line 60: Why does Crawler need this argument? It doesn't accept e.g. a SignalManager or StatsCollector.

scrapy/crawler.py
```diff
@@ -190,13 +199,14 @@ def _get_spider_loader(settings):
             )
         return loader_cls.from_settings(settings.frozencopy())
 
-    def __init__(self, settings=None):
+    def __init__(self, settings=None, addons: Optional[AddonManager] = None):
```
kmike, line 202: Not sure we need to have `addons` argument here.

scrapy/crawler.py
```diff
@@ -321,8 +331,13 @@ class CrawlerProcess(CrawlerRunner):
     process. See :ref:`run-from-script` for an example.
     """
 
-    def __init__(self, settings=None, install_root_handler=True):
-        super().__init__(settings)
+    def __init__(
+        self,
+        settings=None,
+        install_root_handler: bool = True,
+        addons: Optional[AddonManager] = None,
```
kmike, line 338: Not sure we need to have `addons` argument here.

scrapy/interfaces.py
```diff
@@ -15,3 +15,22 @@ def list():
 
     def find_by_request(request):
         """Return the list of spiders names that can handle the given request"""
+
+
+class IAddon(Interface):
+    """Scrapy add-on"""
+
+    name = Attribute("""Add-on name""")
+    version = Attribute("""Add-on version string (PEP440)""")
```
kmike, line 24: As these attributes don't seem to be required, I'm not sure we need this interface.

scrapy/utils/conf.py
```diff
@@ -99,14 +99,20 @@ def init_env(project="default", set_syspath=True):
             sys.path.append(projdir)
 
 
-def get_config(use_closest=True):
```
kmike, line 102: Is it a general refactoring, or is it needed for the PR to work?

scrapy/utils/misc.py
```diff
@@ -72,6 +72,22 @@ def load_object(path: Union[str, Callable]) -> Any:
     return obj
 
 
+def load_module_or_object(path):
+    """Load python module or (non-module) object from given path.
```
kmike, line 76: I think we can drop python module support.

docs/topics/addons.rst
wRAR: Basically you set a dict attribute that will be merged into Scrapy settings

docs/topics/addons.rst
wRAR: The prepending is not allowed to this attribute

docs/topics/addons.rst
wRAR: So, the original PR changed load_object() to add support for types and instances. We added support for types in a separate PR but didn't add support for instances. I've removed changes to load_object() from here but this part of the doc still assumes instances are supported.

scrapy/utils/conf.py
wRAR: It's no longer needed.

docs/topics/addons.rst
wRAR: Currently this is required because of the `AddonManager._addons` dict, to be able to access, remove or disable an addon programmatically. But I think we don't want any of these features?

wRAR:
I've dropped most of the optional features. Remaining questionable ones are addon name and the idea of a per-addon dict with configuration (with an optional export of this dict as flat separate values back into settings), as we could just say that addon settings should be put as separate values as usual.

docs/topics/addons.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Set some basic configuration using the :class:`Addon` base class::
+
+    from scrapy.addons import Addon
+
+    class MyAddon(Addon):
+        name = 'myaddon'
+
+        def update_settings(self, config, settings):
+            super().update_settings(settings)
```
kmike, line 113: the signature looks different

docs/topics/addons.rst
wRAR: Another place we need the name in the current code is the autodetected name for the addon settings dict (and generating that from the class name may be problematic, especially if you want to subclass it while keeping the setting name).

docs/topics/addons.rst
Gallaecio: I can see how having a crawler object available at the point in time when this method gets called could be useful, e.g. to monkeypatch some crawler component after it has been initialized.

That said, assuming addons are Scrapy components, meaning they can define a `from_crawler` class method and keep track of it (e.g. store it in  `self.crawler`), I think that and the use of signals should be able to cover most use cases there may be for having a crawler parameter in this method, or for having this method period :slightly_smiling_face:.

docs/topics/addons.rst
```diff
@@ -0,0 +1,145 @@
...
+Scrapy's settings namespace, in upper case. It is
+easy to write your own functionality while still being able to use the
+convenience functions by overwriting
+:meth:`~scrapy.addons.Addon.update_settings`.
+
+.. module:: scrapy.addons
+   :noindex:
+
+.. autoclass:: Addon
+   :members:
```
Gallaecio, line 99: I am not strongly against it, but we could probably do without a base class at all.

docs/topics/addons.rst
```diff
@@ -0,0 +1,145 @@
...
+
+        def update_settings(self, config, settings):
+            super().update_settings(settings)
+            settings.set('DNSCACHE_ENABLED', False, priority='addon')
+
+        def check_configuration(self, config, crawler):
+            if crawler.settings.getbool('DNSCACHE_ENABLED'):
+                # The spider, some other add-on, or the user messed with the
+                # DNS cache setting
+                raise ValueError("myaddon is incompatible with DNS cache")
```
Gallaecio, line 145: I am not sure whether or not we should encourage this approach of post-setting-update checking on add-ons. But this is one example where I think we don’t need a method. For example (untested):

```python
from logging import getLogger

from scrapy.signals import spider_opened

logger = getLogger(__name__)

class MinimalisticApproachAddon:
    
    @classmethod
        return cls(crawler)        

    def __init__(self, crawler):
        crawler.signals.connect(self.spider_opened, signal=spider_opened)

    def update_settings(self, settings):
        super().update_settings(settings)
        settings.set('DNSCACHE_ENABLED', False, priority='addon')
        
        if spider.settings.getbool('DNSCACHE_ENABLED'):
            logger.error(f"DNSCACHE_ENABLED is enabled, and {self.__class__} does not work with it.")
            return spider.crawler.engine.close_spider()
```

docs/topics/addons.rst
wRAR: If we remove the per-addon automatic setting names then it's no longer useful, yeah.

wRAR: > assuming addons are Scrapy components, meaning they can define a from_crawler class method

They are currently not Scrapy components though, we need to add a `create_instance` call to AddonManager for this to work.

wRAR:
The next stage of minimization, with just an `update_settings` method. Should I add `from_crawler` support next?

wRAR:
Done.

wRAR:
I've implemented a sample addon for scrapy-zyte-api: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/96

docs/topics/addons.rst
```diff
@@ -0,0 +1,182 @@
...
+    ADDONS = {
+        'path.to.someaddon': 0,
+        path.to.someaddon2: 1,
+    }
+
+
+Writing your own add-ons
+========================
+
+Add-ons are (any) Python objects that include the following method:
```
Gallaecio, line 37: Even if this may be technically true, shouldn’t we encourage them being classes from the docs?

docs/topics/addons.rst
```diff
@@ -0,0 +1,182 @@
...
+    :type settings: :class:`~scrapy.settings.Settings`
+
+They can also have the following method:
+
+.. classmethod:: from_crawler(cls, crawler)
+   :noindex:
+
+   If present, this class method is called to create an addon instance
+   from a :class:`~scrapy.crawler.Crawler`. It must return a new instance
+   of the addon. Crawler object provides access to all Scrapy core
```
Gallaecio, line 57:
```suggestion
   of the addon. The crawler object provides access to all Scrapy core
```

docs/topics/addons.rst
```diff
@@ -0,0 +1,182 @@
...
+   components like settings and signals; it is a way for pipeline to
+   access them and hook its functionality into Scrapy.
```
Gallaecio, line 59: Is this what was meant?

```suggestion
   components like settings and signals; it is a way for the addon to
   access them and hook its functionality into Scrapy.
```

docs/topics/addons.rst
```diff
@@ -0,0 +1,182 @@
...
+
+    class MyAddon:
+        def update_settings(self, settings):
+            if settings.getbool("MYADDON_ENABLE_PIPELINE"):
+                settings["ITEM_PIPELINES"]["path.to.mypipeline"] = 200
+
+Fallbacks
+---------
+
+Some components provided by addons need to fallback to "default"
```
Gallaecio, line 85:
```suggestion
Some components provided by addons need to fall back to "default"
```

docs/topics/addons.rst
```diff
@@ -0,0 +1,182 @@
...
+   default Scrapy one (e.g.
+   ``scrapy.core.downloader.handlers.http.HTTPDownloadHandler``), but instead
+   be able to load the class of the fallback component from a special setting
+   (e.g. ``MY_FALLBACK_DOWNLOAD_HANDLER``), create an instance of it and use
+   it.
+2. The addons that include these components should read the current value of
+   the default setting (e.g. ``DOWNLOAD_HANDLERS``) in their
+   ``update_settings()`` methods, save that value into the fallback setting
+   (``MY_FALLBACK_DOWNLOAD_HANDLER`` mentioned earlier) and set the default
+   setting to the component provided byt the addon (e.g.
```
Gallaecio, line 105:
```suggestion
   setting to the component provided by the addon (e.g.
```

scrapy/addons.py
```diff
@@ -0,0 +1,65 @@
...
+
+        This will load the addon for every add-on path in the
+        ``ADDONS`` setting.
+
+        :param settings: The :class:`~scrapy.settings.Settings` object from \
+            which to read the add-on configuration
+        :type settings: :class:`~scrapy.settings.Settings`
+        """
+        paths = build_component_list(settings["ADDONS"])
+        addons = [load_object(path) for path in paths]
```
Gallaecio, line 46: I wonder if the use if `load_object` here is on purpose, given `self.add` already takes care of that.

But maybe it *is* on purpose, to normalize `addons` for the `logger.info` call later, without having to instead use `[addon.__class__ for addon in self.addons]` instead.

sep/sep-021.rst
Gallaecio, line 1: I sometimes wonder if we should remove the whole folder at some point, but that’s not in the scope of this PR.

scrapy/addons.py
```diff
@@ -0,0 +1,65 @@
...
+from scrapy.utils.conf import build_component_list
+from scrapy.utils.misc import create_instance, load_object
+
+if TYPE_CHECKING:
+    from scrapy.crawler import Crawler
+
+logger = logging.getLogger(__name__)
+
+
+class AddonManager:
```
Gallaecio, line 13: I wonder if we should make the methods private, at least `add`. No strong opinion, though.

Gallaecio (approved):
I had not realized how well the fallback setting approach could work. Great docs!

scrapy/addons.py
wRAR: Initially it was public because users were able to add addons in their code (and do other things with them that we no longer have APIs for, like removing, disabling and changing configuration).

Gallaecio:
Yes, it would be great if we could normalize it as ”add-on” outside code. No strong opinion on the code side (AddOn vs Addon).

docs/topics/addons.rst
```diff
@@ -0,0 +1,190 @@
...
+
+The ``ADDONS`` setting is a dict in which every key is an add-on class or its
+import path and the value is its priority.
+
+This is an example where two add-ons are enabled in a project's
+``settings.py``::
+
+    ADDONS = {
+        'path.to.someaddon': 0,
+        path.to.someaddon2: 1,
```
kmike, line 30:
```suggestion
        SomeAddonClass: 1,
```

docs/topics/addons.rst
```diff
@@ -0,0 +1,190 @@
...
+developers.
+
+
+Activating and configuring add-ons
+==================================
+
+Add-ons and their configuration live in Scrapy's
+:class:`~scrapy.addons.AddonManager`. During a :class:`~scrapy.crawler.Crawler`
+initialization the add-on manager will read a list of enabled add-ons from your
+``ADDONS`` setting.
```
kmike, line 20: I wonder if we should remove AddonManager here. We don't put emphasis on e.g. MiddlewareManager, as users are not goint to use this directly. 

```suggestion
During a :class:`~scrapy.crawler.Crawler`
initialization it reads a list of enabled add-ons from your
``ADDONS`` setting.
```

docs/topics/addons.rst
```diff
@@ -0,0 +1,190 @@
...
+Check dependencies:
+
+.. code-block:: python
+
+    class MyAddon:
+        def update_settings(self, settings):
+            try:
+                import boto
+            except ImportError:
+                raise RuntimeError("MyAddon requires the boto library")
```
kmike, line 135: What would happen if NotConfigured error is raised? I think this is the error which components raise when a dependency is not available. Is there a reason to prefer RuntimeError in addons?

docs/topics/addons.rst
```diff
@@ -0,0 +1,190 @@
...
+            ...
+
+Use a fallback component:
+
+.. code-block:: python
+
+    from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
+
+
+    fallback_setting = "MY_FALLBACK_DOWNLOAD_HANDLER"
```
kmike, line 161: should we upper-case a constant name?

docs/topics/addons.rst
```diff
@@ -0,0 +1,190 @@
...
+        def download_request(self, request, spider):
+            if request.meta.get("my_params"):
+                # handle the request
+                ...
+            else:
+                return self._fallback_handler.download_request(request, spider)
+
+
+    class MyAddon:
+        def update_settings(self, settings):
```
kmike, line 184: I think this addon should also enable MyHandler here.

docs/topics/api.rst
```diff
@@ -154,6 +155,17 @@ Settings API
 .. autoclass:: BaseSettings
    :members:
 
+.. _topics-api-addonmanager:
+
+AddonManager API
+================
```
kmike, line 161: I think we can remove this section, as AddonManager is an internal component. This section contains classes which users can either use directly, or override via settings with custom implementations; it doesn't seem that AddonManager fits.

tests/test_addons.py
```diff
@@ -0,0 +1,62 @@
...
+        settings.set("KEY2", "project", priority="project")
+        addon_config = {"KEY1": "addon", "KEY2": "addon", "KEY3": "addon"}
+        testaddon = GoodAddon(addon_config)
+        testaddon.update_settings(settings)
+        self.assertEqual(settings["KEY1"], "addon")
+        self.assertEqual(settings["KEY2"], "project")
+        self.assertEqual(settings["KEY3"], "addon")
+
+
+class AddonManagerTest(unittest.TestCase):
```
kmike, line 45: Could you please add a couple more tests:

1. Check that addon priority works (i.e. a right addon is used , based on the priority)
2. Check that using addon classes works in settings, not just paths

Something else?

tests/test_addons.py
```diff
@@ -0,0 +1,62 @@
+import unittest
+from typing import Any, Dict, Optional
+
+from scrapy.crawler import Crawler
+from scrapy.settings import BaseSettings
+from scrapy.utils.test import get_crawler
+
+
+class GoodAddon:
+    def __init__(self, config: Optional[Dict[str, Any]] = None) -> None:
```
kmike, line 10: I'm not sure, what's the use case? It seems you can't use such addons with a custom config in setting.py, as a class needs to be used, not an instance, right?

tests/test_addons.py
kmike: I think we should also add tests for the suggestions we make in the documentation, e.g. using "addon" priority and they way it interacts with other settings; an addon which reads the default setting and uses a fallback implementation.

docs/topics/addons.rst
Gallaecio: I agree with removing the mention, although I would suggest a slightly different wording:

```suggestion
During :class:`~scrapy.crawler.Crawler` initialization, the list of enabled 
add-ons is read from your ``ADDONS`` setting.
```

docs/topics/addons.rst
Gallaecio: I have no strong opinion on it, but I think it makes sense that, if you have enabled a plugin that requires a dependency, and that dependency is missing, your crawler should fail to initialize, forcing you to either disable the plugin or install the missing dependency. Same for missing required settings.

I see `NotConfigured` as something a Scrapy component raises when it is configured but disabled through settings.

If a plugin has an `*_ENABLED` setting, then I think it would make sense to support that plugin raising `NotConfigured`, and that causing the plugin to not be used at run time, and a log message indicating so during initialization, but things working otherwise.

So I think it may make sense to add support for `NotConfigured` on the addon manager class, but I think in some cases raising something else makes sense, including this one.

tests/test_addons.py
Gallaecio: It looked good to me on my latest pass. Though I guess the `AddonTest` below only really tests BaseSettings.update, so maybe it is not that useful.

docs/topics/addons.rst
kmike: So, the difference is that RuntimeError stops the spider, but NotConfigured doesn't? Does it work for addons the same way?

I'm not really familiar with these parts of the code, but we're using NotConfigured in similar cases:

* https://github.com/scrapy/scrapy/blob/7522aeed357d90ed95ee10d3b5a506f1b1049d1f/scrapy/pipelines/images.py#L61
* https://github.com/scrapy/scrapy/blob/7522aeed357d90ed95ee10d3b5a506f1b1049d1f/scrapy/extensions/feedexport.py#L176
* https://github.com/scrapy/scrapy/blob/7522aeed357d90ed95ee10d3b5a506f1b1049d1f/scrapy/core/downloader/handlers/s3.py#L20

tests/test_addons.py
kmike: I think that users shouldn't be defining addons with `__init__` arguments, unless the addon has a from_crawler method which passes these arguments to `__init__`.

docs/topics/addons.rst
Gallaecio: I think that’s because those components are configured by default, i.e. the `_ENABLED=True` is implicit, so doing otherwise would be a bad user experience. But for opt-in components, including add-ons, I think stopping the spider may be best.

tests/test_addons.py
Gallaecio: I don’t think it was meant as an example of a real use case, but as a shortcut for a clean test later on.

wRAR: It's just a remnant of old paradigm and it's handy to be able to use the same class with different settings, I guess I can rewrite it to a factory.

docs/topics/addons.rst
```diff
@@ -0,0 +1,189 @@
...
+            if not settings.get(FALLBACK_SETTING):
+                settings.set(
+                    FALLBACK_SETTING,
+                    settings.getwithbase("DOWNLOAD_HANDLERS")["https"],
+                    "addon",
+                )
+            settings["DOWNLOAD_HANDLERS"]["https"] = "path.to.MyHandler"
```
kmike, line 189: can you use MyHandler here directly? Or would you consider it an anti-pattern? Not sure myself :)

docs/topics/addons.rst
wRAR: OK, so nothing handles `NotConfigured` raised there.  Which is expected because it's only handled in `MiddlewareManager` (which is the base for middleware, pipeline and extension managers) and `DownloadHandlers`. We should probably add handling to AddonManager to similarly silently skip add-ons that raise it?

wRAR: If we do that I wonder if we should merge `AddonManager.update_settings()` into `AddonManager.load_settings()` or into `AddonManager._add()` so that `AddonManager.addons` never contains add-ons that raise `NotConfigured`. Also it doesn't seem useful to have a separate function that calls add-ons' `update_settings()` as we don't have any features that would benefit from it.

scrapy/addons.py
```diff
@@ -41,13 +29,29 @@ def load_settings(self, settings: Settings) -> None:
...
+                addoncls = load_object(clspath)
+                addon = create_instance(
+                    addoncls, settings=settings, crawler=self.crawler
+                )
+                addon.update_settings(settings)
+                self.addons.append(addon)
+            except NotConfigured as e:
+                if e.args:
+                    clsname = (
+                        clspath.split(".")[-1] if isinstance(clspath, str) else clspath
```
kmike, line 44: Why is this processing needed?

Gallaecio: No strong opinion. 1 sounds great, if it is not too bothersome. Otherwise, 2 makes sense. 3 is more consistent than 2, but I don’t think that matters that much here.

kmike: It looks like you went with 1 :)

kmike (approved):

scrapy/addons.py
wRAR: Yeah, I decided it's easy enough.

wRAR:
Thanks!

jdemaeyer:
:exploding_head:

PR #5952, opened 2023-06-15
Fix JsonItemExporter puts lone comma in the output if encoder fails

### LINKED ISSUE #3090: JsonItemExporter puts lone comma in the output if encoder fails

adnan-awan:
Continuation of #3111.

Resolves #3090, closes #3111.


scrapy/exporters.py
```diff
@@ -142,14 +149,15 @@ def finish_exporting(self):
...
-            self._beautify_newline()
-        itemdict = dict(self._get_serialized_fields(item))
-        data = self.encoder.encode(itemdict)
-        self.file.write(to_bytes(data, self.encoding))
+        # catch exception and don't add empty item for export
+        try:
+            itemdict = dict(self._get_serialized_fields(item))
+            data = to_bytes(self.encoder.encode(itemdict), self.encoding)
+        except TypeError as e:
+            raise e
```
wRAR, line 157: I don't think this catch+reraise does anything?

tests/test_exporters.py
```diff
@@ -599,6 +599,21 @@ def test_two_items(self):
...
+        i2 = TestItem(
+            name="Maria", age=1j
+        )  # Invalid datetimes didn't consistently fail between Python versions
+        i3 = TestItem(name="Jesus", age="44")
+        self.ie.start_exporting()
+        self.ie.export_item(i1)
+        self.assertRaises(TypeError, self.ie.export_item, i2)
+        self.ie.export_item(i3)
+        self.ie.finish_exporting()
+        print(f"values are=={self.output.getvalue()}")
```
wRAR, line 613: I guess this is a debug print?

scrapy/exporters.py
adnan-awan: @wRAR  thanks for your comment. 

Actually, [this](https://github.com/scrapy/scrapy/pull/5952/commits/0bd481de7cec4540aad8b9cfae896519fd9b2016#diff-b4b7b51f903343e1ac43a4655e42263010d0631f9d1702eb5058cbd5d5402d47R608) test expect the TypeError. Do you've something better in your mind for its replacement?.

wRAR: But as this code does nothing then removing it won't break the test.

wRAR (approved):

kmike:
Thanks @adnan-awan!

PR #5958, opened 2023-06-20
Remove `pkg_resources` usage

### LINKED ISSUE #5956: Remove pkg_resources usage
Importing pkg_resources raises a deprecation warning now so it makes sense to remove it. We can even use importlib without checking if it's available as we no longer support 3.7. I see only `pkg_resources.parse_version` and `pkg_resources.iter_entry_points` used in Scrapy, I think they can be replaced by things from `packaging` and `importlib.metadata` respectively.

Laerte:
Fix #5956 

Removed code for checking `setup_tools` version since we are checking for a version from [2015](https://pypi.org/project/setuptools/18.5/)

### DIFF
--- a/docs/topics/components.rst
+++ b/docs/topics/components.rst
@@ -70,7 +70,7 @@ If your requirement is a minimum Scrapy version, you may use
 
 .. code-block:: python
 
-    from pkg_resources import parse_version
+    from packaging.version import parse as parse_version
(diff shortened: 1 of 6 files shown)


Gallaecio (approved):

wRAR (approved):
Thanks!

Issue #5959, opened 2023-06-21
BaseSettings.pop method throws KeyError even if default value is provided


Prometheus3375:
# Description

`BaseSettings` inherit from `MutableMapping` and thus has method `pop`. This method takes a key and excludes it from the mapping returning its value. If key does not exist, then `KeyError` is raised, but if default value is provided, then it is returned instead.

# Steps to Reproduce

Run this sample:
```python
from scrapy.settings import Settings
s = Settings()
print(s.pop('meh', 'Hello world!'))
```

# Expected behavior
`Hello world!` is printed.

# Actual behavior:
```
Traceback (most recent call last):
  File "REDACTED\main.py", line 3, in <module>
    print(s.pop('meh', 'Hello world!'))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen _collections_abc>", line 918, in pop
  File "REDACTED\.venv\Lib\site-packages\scrapy\settings\__init__.py", line 362, in __delitem__
    del self.attributes[name]
        ~~~~~~~~~~~~~~~^^^^^^
KeyError: 'meh'
```

**Reproduces how often:** always

# Current workaround
Since 2.9.0:
```python
from scrapy.settings import Settings

s = Settings()
s.setdefault('meh', 'Hello world!')
print(s.pop('meh'))

```

Earlier:
```python
from scrapy.settings import Settings
s = Settings()
if 'meh' in s:
    print(s.pop('meh'))
else:
    print('Hello world!')
```

# Versions

```
Scrapy       : 2.8.0
lxml         : 4.9.2.0
libxml2      : 2.9.12
cssselect    : 1.2.0
parsel       : 1.8.1
w3lib        : 2.1.1
Twisted      : 22.10.0
Python       : 3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]
pyOpenSSL    : 23.1.1 (OpenSSL 3.1.0 14 Mar 2023)
cryptography : 40.0.2
Platform     : Windows-10-10.0.19044-SP0
```

# Additional context

Please, check that `BaseSettings` indeed implement all `MutableMapping` methods correctly. There was already [similar issue](https://github.com/scrapy/scrapy/issues/5811) with `setdefault` method. [List of methods](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) implemented inside `MutableMapping` in red rectangle:
[...]

PR #5963, opened 2023-06-27
fix: Return value instead of `SettingsAttribute` object when using `pop` method

Laerte:
As reported on https://github.com/scrapy/scrapy/issues/5959#issuecomment-1609832116, the implementation of `pop` method is returning `SettingsAttribute` object instead of the value itself which break interface. Mea culpa, sorry! 😅


caledrummond90 (approved):

Gallaecio (approved):

Gallaecio:
Thanks!

PR #5965, opened 2023-07-05
Removing hard code entries

andersoncarlosfs:
Just removing hard code statements

### DIFF
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -12,8 +12,8 @@ DATADIR_CFG_SECTION = "datadir"
 
 
 def inside_project():
-    scrapy_module = os.environ.get("SCRAPY_SETTINGS_MODULE")
-    if scrapy_module is not None:
+    scrapy_module = os.environ.get(ENVVAR)
+    if scrapy_module:
         try:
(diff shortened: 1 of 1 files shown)


scrapy/utils/project.py
```diff
@@ -12,8 +12,8 @@
 
 
 def inside_project():
-    scrapy_module = os.environ.get("SCRAPY_SETTINGS_MODULE")
-    if scrapy_module is not None:
+    scrapy_module = os.environ.get(ENVVAR)
+    if scrapy_module:
```
Gallaecio, line 16: This is a change in behavior when the `SCRAPY_SETTINGS_MODULE` is set to an empty string. That said, an empty string cannot work here, so I guess it is fine and slightly improves readability, so +1 from me.

Gallaecio (approved):

scrapy/utils/project.py
andersoncarlosfs: Yeaaaaah!!!! =]

wRAR (approved):

wRAR:
Thanks!

PR #5971, opened 2023-07-11
fix: -O/-o option does not work with absolute paths on Windows

### LINKED ISSUE #5969: crawl -o and -O options does not work with absolute paths on Windows
Absolute paths on Windows has `:` as a part of disk names which leads to unexpected behaviour.

# Steps to Reproduce

Case 1: run command `scrapy crawl spider_name -O "C:\path\to\output.json"`

Case 2: run command `scrapy crawl spider_name -O "C:\path\to\output.json:json"`

# Expected behavior
For both commands the results of the crawling must be saved to `C:\path\to\output.json` file.

# Actual behavior
Case 1: error
```
Unrecognized output format '\path\to\output.json'. Set a supported one (('json', 'jsonlines', 'jsonl', 'jl', 'csv', 'xml', 'marshal', 'pickle')) after a colon at the end of the output URI (i.e. -o/-O <URI>:<FORMAT>) or as a file extension.
```

Case 2: no errors, but file is not created. **Note**: in this example I use system disk, although file is not created if any other disk is used (i.e., permissions in not the issue).

# Reproduces how often
Always

# Versions

```
Scrapy       : 2.9.0
lxml         : 4.9.2.0
libxml2      : 2.9.12
cssselect    : 1.2.0
parsel       : 1.8.1
w3lib        : 2.1.1
Twisted      : 22.10.0
Python       : 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)]
pyOpenSSL    : 23.2.0 (OpenSSL 3.1.1 30 May 2023)
cryptography : 41.0.1
```

Laerte:
fix #5969

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -617,7 +617,7 @@ class FeedExporter:
         It supports not passing the *feed_options* parameters to classes that
         do not support it, and issuing a deprecation warning instead.
         """
-        feedcls = self.storages.get(urlparse(uri).scheme) or self.storages["file"]
+        feedcls = self.storages.get(urlparse(uri).scheme, self.storages["file"])
         crawler = getattr(self, "crawler", None)
 
         def build_instance(builder, *preargs):
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -9,7 +9,7 @@ import re
 import sys
 import warnings
 from datetime import datetime
-from pathlib import Path
+from pathlib import Path, PureWindowsPath
 from tempfile import NamedTemporaryFile
 from typing import IO, Any, Callable, List, Optional, Tuple, Union
 from urllib.parse import unquote, urlparse
@@ -591,7 +591,7 @@ class FeedExporter:
 
     def _storage_supported(self, uri, feed_options):
         scheme = urlparse(uri).scheme
-        if scheme in self.storages:
+        if scheme in self.storages or PureWindowsPath(uri).drive:
             try:
                 self._get_storage(uri, feed_options)
                 return True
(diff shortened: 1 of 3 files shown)


scrapy/extensions/feedexport.py
```diff
@@ -591,7 +591,7 @@ def _settings_are_valid(self):
 
     def _storage_supported(self, uri, feed_options):
         scheme = urlparse(uri).scheme
-        if scheme in self.storages:
+        if scheme in self.storages or PureWindowsPath(uri).drive:
```
Laerte, line 594:
```
>>> urlparse("C:\Scrapy\output.json").scheme
'c'
```
Had to add this condition because `scheme` for windows absolute paths is misleading.

scrapy/extensions/feedexport.py
```diff
@@ -617,7 +617,7 @@ def _get_storage(self, uri, feed_options):
         It supports not passing the *feed_options* parameters to classes that
         do not support it, and issuing a deprecation warning instead.
         """
-        feedcls = self.storages[urlparse(uri).scheme]
+        feedcls = self.storages.get(urlparse(uri).scheme) or self.storages["file"]
```
Laerte, line 620: Basically to avoid `KeyError` fallback to `file`

wRAR, line 620:
```suggestion
        feedcls = self.storages.get(urlparse(uri).scheme, self.storages["file"])
```

(assuming it works that way)

tests/test_commands.py
```diff
@@ -919,6 +919,63 @@ def start_requests(self):
...
+class MySpider(scrapy.Spider):
+    name = 'myspider'
+
+    start_urls = ["data:,"]
+
+    def parse(self, response):
+        yield {"hello": "world"}
+        """
+
+        args = ["-o", "/tmp/dump/output1.json:json"]
```
wRAR, line 936: Can you please use mkdtemp() here too?

Laerte: Done.

scrapy/extensions/feedexport.py
Laerte: I was wondering because i tried this before now i understand why did not work 🤡 

image

Thanks @wRAR

wRAR (approved):

Laerte:
@Gallaecio Can we merge it?

Gallaecio (approved):
Yes, thanks, and sorry for the delay.

PR #5972, opened 2023-07-12
Implement the builder pattern with the request and response classes

suarezsandicamilo:


wRAR:
Not sure if this feature is useful at all, but why do you think changing the existing Scrapy code is worth it? Some changes actually demonstrate the old form is better.

wRAR:
Is this some kind of school assignment too?

suarezsandicamilo:
Yes, we're very sorry for the confusion. This is part of a university assignment for software design. We wouldn't mind if you close the pull requests though. Thank you for understanding!

PR #5974, opened 2023-07-13
Sequence diagram of the scraping process in Scrapy added in the documentation

CristopherH11:
Description:

This contribution adds a sequence diagram that shows the crawling process in Scrapy, providing a clear visual representation of how the system works in that specific context. The diagram illustrates the key interactions between Scrapy's core components, such as the Scheduler, Spiders, and Pipelines, during the process of extracting data from a request. In addition to the sequence diagram, a detailed description has been added to accompany the diagram. This description provides a step-by-step explanation of each stage of the crawling process, highlighting the responsibilities and interactions between the components mentioned above.

Motivation:

The goal is to help Scrapy users and developers better understand the flow and internal logic of the library. Since it allows the user to see what is the order in which Scrapy's actions are executed and what are the responses generated in each action.


wRAR:
https://docs.scrapy.org/en/latest/topics/architecture.html

Issue #5548, opened 2022-07-01
The engine doesn't wait for Spider generator parse before run process_spider_output


trungtin:
### Description

According to the architect, the order of running functions from the spider middleware should be:

`process_spider_input` -> Spider's `parse` -> `process_spider_output`

but if `parse` is a generator function, the order becomes:

`process_spider_input` -> `process_spider_output` -> Spider's `parse`

and I don't see it mentioned anywhere in the doc.

### Steps to Reproduce

This simple spider reproduce the result:

```python
import scrapy
from scrapy import Request

class YcombinatorSpider(scrapy.Spider):
    name = 'ycombinator_news'
    allowed_domains = ['news.ycombinator.com']
    start_urls = ['http://news.ycombinator.com/']

    def parse(self, response):
        print('parse')
        follow = response.xpath('//a[@class="titlelink"]').getall()
        for url in follow:
            follow_url = response.urljoin(url)
            yield Request(url=follow_url, callback=self.populate_item)

    def populate_item(self, response):
        print('populate_item')
        pass

```

### Versions

scrapy==2.6.1

Gallaecio:
Could you provide a complete example, spider middleware and log output included? (you can also target toscrape.com for testing purposes)

trungtin:
> Could you provide a complete example, spider middleware and log output included? (you can also target toscrape.com for testing purposes)

you can go to this replit https://replit.com/@trungtin1/scrapy-middleware-reproduce
switch to shell tab, and run with `scrapy crawl toscrape`

<img width="186" alt="image" src="https://user-images.githubusercontent.com/8097699/176927413-6ff8dcee-f6e8-4937-9907-90447882bbc1.png">

and see that "process spider output" is logged first

Gallaecio:
I see.

Because the `result` that `process_spider_output` is that generator, `process_spider_output` can run code before iterating `result`, and until result is iterated, `parse` does nothing.

I guess it makes sense to clarify this in the documentation of `process_spider_output`, and possibly in the architecture page.

PR #5975, opened 2023-07-14
Documentation explaining anomaly regarding the order of the parse method in the middleware execution process

IgnacioRoblesM:
# Description

The current documentation suggests that the Spider Middleware's execution order is as follows: the response first goes up through the middleware chain from the engine to the spider, then, once in the spider, the response is parsed, and finally, the response goes down the spider middleware back to the engine. While this is almost correct, there is one detail that is not. The response is not parsed when the execution reaches the spider, but until the execution reaches the middleware on its way back to the engine. In more technical terms, the response is parsed by the Spider's parse method, but this method is executed until the first middleware on the way back executes it's process_spider_output method. This misunderstanding was brought to our attention by the following issue: [The engine doesn't wait for Spider generator parse before run process_spider_output](https://github.com/scrapy/scrapy/issues/5548)

Now, in order to correct this misunderstanding that fellow programmers might fall into, we decided to not only add textual documentation to the issue, but add a graphical documentation as well.

# Benefits

This way, if a future programmer like the one who wrote the issue afore mentioned runs into the anomaly that for some reason the unparsed response gets to the middleware after having passed through the spider, they might now understand why and how the flow of the execution works. This is beneficial for both users of scrapy who might want to write a middleware of their own, and for contributors who need to get under the hood of scrapy and its execution.

# Challenges

Thanks to the good file organization of the project, finding where to add the documentation was pretty easy. The one thing that wasn't very frictionless was the building of the html itself in order to test the view of the rst. I use mac and tried to install sphinx, but ran into all sorts of problems.

# Changes location

There were two main places where changes were made. Firstly, the documentation file spider-middleware.rst. In here, we added the documentation right at the end of the **Writing your own spider** section. We saw it fit to place the documentation there since it's where most of the information related to the subject is, including the process_spider_output method. And secondly, we simply added the png file of the graph to the _images directory inside the docs directory.


IgnacioRoblesM:
It is worth noting that this pull request is a part of a University project from a group of students at the University of Costa Rica for the course Software Design. We appreciate your project and it's great comfort for freshman contributers like us.

wRAR:
I'm afraid your text and especially your diagram don't actually explain the caveat found in #5548.

PR #5976, opened 2023-07-14
Refactor SpiderMiddlewares to inherit from the same class. Class is a handler to be used using Chain of Responsibility

SrDavidC:
# Description
This Pull Request introduces changes to the previous implementation of SpiderMiddlewares. Now, the spiderMiddlewares inherit from the same class instead of being independent classes. The classes for each spider's spiderMiddlewares inheritance are based on the `BaseSpiderMiddleware` class, which itself inherits from the `AbstractHandler` class. The main goal is to provide an easier way to handle responses and requests to and from the spiders.

# Benefits

- Each SpiderMiddleware now inherits from the same class, resulting in better abstraction, maintainability, and consistency.
- The flow remains the same, but now Scrapy allows for dynamic runtime changes to the sequence of SpiderMiddlewares, not just at startup.
- SpiderMiddlewares can be added or removed at runtime.
- The behavior remains consistent with the previous implementation.

# Challenges
There are some challenges involved, such as refactoring certain implementations in the `SpiderMiddlewaresManager` to work with `process_output`. Additionally, some unit tests need to be modified to accommodate these changes.

### DIFF
--- a/scrapy/core/spidermw.py
+++ b/scrapy/core/spidermw.py
@@ -28,6 +28,9 @@ from scrapy.exceptions import _InvalidOutput
 from scrapy.http import Response
 from scrapy.middleware import MiddlewareManager
 from scrapy.settings import BaseSettings
+from scrapy.spidermiddlewares.basespidermiddleware import BaseSpiderMiddleware
+from scrapy.spidermiddlewares.referer import RefererMiddleware
+from scrapy.spidermiddlewares.urllength import UrlLengthMiddleware
 from scrapy.utils.asyncgen import as_async_generator, collect_asyncgen
 from scrapy.utils.conf import build_component_list
 from scrapy.utils.defer import (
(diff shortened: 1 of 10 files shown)


SrDavidC:
This Pull Request is a contribution from a group of university students working on a project related to implementing Software Design Patterns in open source projects. We sincerely appreciate your project and we are grateful for the opportunity to be a part of it. It has been a valuable learning experience for us. Thanks for your support and collaboration

wRAR:
I don't think your code does what you claim it does, and I don't think it's a feature we want.

> This Pull Request is a contribution from a group of university students working on a project related to implementing Software Design Patterns in open source projects.

So far the proposed implementations of patterns were just spam, not something we want to merge. Please don't submit PRs that are not intended to be merged. If, on the on the other hand, you sincerely think you are submitting a useful feature, please think twice and please don't use flowery marketing language that isn't actually true when describing both changes and their benefits.

PR #5977, opened 2023-07-16
Typing for scrapy.settings.BaseSettings

wRAR:
This adds full typing for `BaseSettings`. There is one potential optimisation here: `BaseSettings` is both the Scrapy settings as a whole (with keys that can only be strings) and setting values for things like `FEEDS` (with keys of various types), if it's possible to separate these it's possible to type the former more correctly. But that seems to require changing the class hierarchy and so it's more harm than good.

Another things that is out of scope and not really useful is `Settings` vs `BaseSettings` in the annotations of the code that uses settings. The code difference between those classes is only the `__init__` method so their interfaces are currently identical.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -42,7 +42,7 @@ deps =
     types-pyOpenSSL==23.2.0.1
     types-setuptools==68.0.0.1
 commands =
-    mypy --show-error-codes {posargs: scrapy tests}
+    mypy {posargs: scrapy tests}
 
 [testenv:pre-commit]
 basepython = python3
--- a/tox.ini
+++ b/tox.ini
@@ -33,13 +33,13 @@ install_command =
 [testenv:typing]
 basepython = python3
 deps =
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -33,13 +33,14 @@ install_command =
...
-    types-Pillow==9.5.0.2
-    types-Pygments==2.15.0.0
-    types-pyOpenSSL==23.1.0.2
-    types-setuptools==67.7.0.1
+    types-Pillow==10.0.0.1
+    types-Pygments==2.15.0.1
+    types-pyOpenSSL==23.2.0.1
+    types-setuptools==68.0.0.1
 commands =
     mypy --show-error-codes {posargs: scrapy tests}
```
kmike, line 45: I think --show-error-codes is true by default in recent mypy

Gallaecio (approved):

PR #5978, opened 2023-07-17
my solution for enhancement #issue5968

SkvProgrammer:
we directly access response.body and use the json.loads() function from the json module to parse the JSON data into a Python object.

By using this approach, you avoid unnecessary memory usage and the need to store both the bytes and string representation of the response.


wRAR:
This is completely wrong.

PR #5979, opened 2023-07-18
introduce CLOSESPIDER_TIMEOUT_NO_ITEM in CloseSpider

BurnzZ:
**Motivation:** Sometimes spiders still keep running without producing any items. This could be due to a variety of reasons like poor dupe URL filtration, bad crawling strategy, etc.

**Proposal:** Auto-close the spider if it hasn't produced any items in the past number of seconds.

### DIFF
--- a/scrapy/extensions/closespider.py
+++ b/scrapy/extensions/closespider.py
@@ -4,10 +4,13 @@ conditions are met.
 See documentation in docs/topics/extensions.rst
 """
 
+import logging
 from collections import defaultdict
 
 from scrapy import signals
 from scrapy.exceptions import NotConfigured
+
+logger = logging.getLogger(__name__)
 
 
 class CloseSpider:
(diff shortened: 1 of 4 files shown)


scrapy/extensions/closespider.py
```diff
@@ -69,3 +82,31 @@ def spider_closed(self, spider):
...
+        logger.info(
+            f"Spider will stop when no items are produced after "
+            f"{self.timeout_no_item} seconds."
+        )
+
+    def item_scraped_no_item(self, item, spider):
+        self.items_in_period += 1
+
+    def _count_items_produced(self, spider):
+        if self.items_in_period >= 1:
```
BurnzZ, line 105: I initially thought of exposing this but I think it's better to keep it simple for now. Users can opt to override this method anyway if they need a customized condition of when to close the spider.

kmike (approved):
Looks good to me! If needed, we can add the "items in period" setting later.

wRAR (approved):

PR #5982, opened 2023-07-20
Feed URL has been added to the stats.

haideriqbal1:
In this PR I have added the feed URL to the stats of the spider's job. Now anybody can easily get the feed URL from the stats of a job instead of searching for it in the logs of a job.

**Testing**:
I have uploaded these changes to the internal PYPI server. You can easily install this version of Scrapy by running the below command.
`pip install scrapy (Add remaining command from internal PYPI docs)`

After installing this version of Scrapy you can run the below spider by the following command.
`scrapy crawl feed_exporter_tester -o testing/test.json`

After the completion of execution, you can check the stats of the job and hopefully, you will get a stat like this ('feedexport/feed_url/test.json': 'testing/test.json').

```
from scrapy import Spider
from scrapy import Item, Field
from itemloaders.processors import TakeFirst


class BookItem(Item):
    book_title = Field()
    book_price = Field()
    book_image_URL = Field()
    book_detail_page_URL = Field()


class BookItemLoader(ItemLoader):
    default_item_class = BookItem
    default_output_processor = TakeFirst()


class FeedExporterTesterSpider(Spider):
    name = "feed_exporter_tester"
    allowed_domains = ['books.toscrape.com']
    start_urls = ['https://books.toscrape.com/']

    def parse(self, response):
        books = response.xpath('//section/div/ol[contains(@class , "row")]//li')
        for book in books:
            book_il = BookItemLoader(selector=book)
            book_il.add_xpath('book_title', './/h3/a/@title')
            book_il.add_xpath('book_price', './/p[@class="price_color"]/text()')
            book_img_url_rel = book.xpath('.//div[@class="image_container"]/a/img/@src').extract_first()
            book_il.add_value('book_image_URL', response.urljoin(book_img_url_rel))
            detail_page_url_rel = book.xpath('.//h3/a/@href').extract_first()
            book_il.add_value('book_detail_page_URL', response.urljoin(detail_page_url_rel))
            yield book_il.load_item()

```

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -480,6 +480,8 @@ class FeedExporter:
 
     def _handle_store_success(self, f, logmsg, spider, slot_type):
         logger.info("Stored %s", logmsg, extra={"spider": spider})
+        feed_uri = logmsg.split(":")[-1].strip()
+        self.crawler.stats.set_value(f"feedexport/feed_url/{feed_uri.split('/')[-1]}", feed_uri)
         self.crawler.stats.inc_value(f"feedexport/success_count/{slot_type}")
 
     def _start_new_batch(self, batch_id, uri, feed_options, spider, uri_template):


scrapy/extensions/feedexport.py
```diff
@@ -480,6 +480,8 @@ def _handle_store_error(self, f, logmsg, spider, slot_type):
 
     def _handle_store_success(self, f, logmsg, spider, slot_type):
         logger.info("Stored %s", logmsg, extra={"spider": spider})
+        feed_uri = logmsg.split(":")[-1].strip()
+        self.crawler.stats.set_value(f"feedexport/feed_url/{feed_uri.split('/')[-1]}", feed_uri)
```
VMRuiz, line 484: Maybe this should be a Scrapy Extensions that users can opt-in

scrapy/extensions/feedexport.py
```diff
@@ -480,6 +480,8 @@ def _handle_store_error(self, f, logmsg, spider, slot_type):
 
     def _handle_store_success(self, f, logmsg, spider, slot_type):
         logger.info("Stored %s", logmsg, extra={"spider": spider})
+        feed_uri = logmsg.split(":")[-1].strip()
```
VMRuiz, line 483: Avoid this type of text transformation, is quite possible it will break on the future if the format changes. If you need `slot.uri` you can pass it as an additional argument

scrapy/extensions/feedexport.py
VMRuiz, line 484: Please, use alternatives like https://docs.python.org/3/library/pathlib.html or https://docs.python.org/3/library/urllib.parse.html to operate with URIs paths

VMRuiz (changes requested):

scrapy/extensions/feedexport.py
VMRuiz: What would you prefer? @Gallaecio @wRAR

VMRuiz:
A part from the concerns on my previous comments, you will also need to add some docs and unit tests for the PR.

Gallaecio:
To be honest, I am not sure it makes sense to add this to the stats at all.

What’s the scenario we want to address here? Not being sure that the right values are being used? Maybe logging the contents of `FEEDS` (if not done already) would be a better approach.

VMRuiz:
> To be honest, I am not sure it makes sense to add this to the stats at all.
> 
> What’s the scenario we want to address here? Not being sure that the right values are being used? Maybe logging the contents of `FEEDS` (if not done already) would be a better approach.

The idea is to facilitate Spidermon monitors to check that FEEDs have been uploaded and also to have a quick way to find the upload FEED URL rather than searching across million of logs.

wRAR:
I feel like it's not the only time when something is added to stats just so spidermon is able to take it from there, but I don't think we have better ways to pass this.

AdrianAtZyte:
> The stated need (Spidermon checking that feeds were uploaded, without grepping logs) is what feed_slot_closed — added in #5876, which landed before this PR — is for. A user-side extension of five lines can put slot.uri into stats in whatever shape that user wants, with no guesses from us about keying.

Should we close this?

PR #5996, opened 2023-08-02
Remove functions deprecated in 2.6.0

Laerte:
~The only deprecated logic that i didn't remove was:~
~- https://github.com/scrapy/scrapy/pull/5090~

~Because seems a little be tricky (should we remove the deprecated parameters?) 😄~

I will commit the removal of #5090 logic later on in this PR.


Laerte:
@wRAR Not at all! I will handle the leftover later then, i was just making sure, thanks man!

scrapy/core/engine.py
```diff
@@ -311,30 +293,20 @@ def _schedule_request(self, request: Request, spider: Spider) -> None:
...
-            if spider is not self.spider:
-                logger.warning(
-                    "The spider '%s' does not match the open spider", spider.name
-                )
         if self.spider is None:
             raise RuntimeError(f"No open spider to crawl: {request}")
-        return self._download(request, spider).addBoth(
-            self._downloaded, request, spider
+        return self._download(request, self.spider).addBoth(
+            self._downloaded, request, self.spider
```
Laerte, line 301: Maybe remove this parameter altogether from `_download` function? Since we have this logic inside the function?

https://github.com/scrapy/scrapy/blob/9f9a2292e08cb0944e1e88e64cef1f4b17486ec8/scrapy/core/engine.py#L344-L345

Laerte: What you think? @wRAR

Laerte:
@wRAR Seems something went off when we deprecated parameters from engine, can you please cancel the queued runs? I thinking in push this change in another PR so i can check why this happen. 😅

Laerte:
Thanks @wRAR I will raise another PR soon as i understand what went off.

Gallaecio (approved):

PR #6001, opened 2023-08-04
Add release notes for 2.10.0.

wRAR:


docs/news.rst
```diff
@@ -3,6 +3,176 @@
...
+
+Scrapy 2.10.0 (YYYY-MM-DD)
+--------------------------
+
+Highlights:
+
+-   Added Python 3.12 support, dropped Python 3.7 support.
+
+-   The add-ons framework that simplifies configuring 3rd-party components that
+    support it.
```
Gallaecio, line 16:
```suggestion
-   The new add-ons framework simplifies configuring 3rd-party components that
    support it.
```

docs/news.rst
```diff
@@ -3,6 +3,176 @@
...
+-   Minimum versions increased for these dependencies:
+
+    -   lxml_: 4.3.0 → 4.4.1
+
+    -   cryptography_: 3.4.6 → 36.0.0
+
+-   ``pkg_resources`` is no longer used. (:issue:`5956`, :issue:`5958`)
+
+-   boto3_ is now recommended for exporting to S3 instead of botocore_.
+    (:issue:`5833`).
```
Gallaecio, line 38:
```suggestion
-   boto3_ is now recommended instead of botocore_ for exporting to S3.
    (:issue:`5833`).
```

It may seem arbitrary, but it can minimize confusion, the current wording can be misread as “[…] instead of [exporting to] botocore”, however little sense that would make.

docs/news.rst
```diff
@@ -3,6 +3,176 @@
...
+
+Backward-incompatible changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-   The value of the :setting:`FEED_STORE_EMPTY` is now ``True`` instead of
+    ``False``. In earlier Scrapy versions empty files were created even when
+    this setting was ``False`` (which was a bug that is now fixed), so the new
+    default should keep the old behavior. (:issue:`872`, :issue:`5847`)
```
Gallaecio, line 46: the FEED_STORE_EMPTY → the FEED_STORE_EMPTY setting

```suggestion
-   The value of the :setting:`FEED_STORE_EMPTY` setting is now ``True`` 
    instead of ``False``. In earlier Scrapy versions empty files were created 
    even when this setting was ``False`` (which was a bug that is now fixed), 
    so the new default should keep the old behavior. (:issue:`872`, 
    :issue:`5847`)
```

Gallaecio (approved):
:heart:

PR #6003, opened 2023-08-06
Typing for scrapy/utils, second pass

wRAR:


scrapy/utils/iterators.py
```diff
@@ -71,30 +101,39 @@ def xmliter_lxml(obj, nodename, namespace=None, prefix="x"):
...
             self._text, self.encoding = obj.body, obj.encoding
         else:
             self._text, self.encoding = obj, "utf-8"
-        self._is_unicode = isinstance(self._text, str)
+        self._is_unicode: bool = isinstance(self._text, str)
 
-    def read(self, n=65535):
-        self.read = self._read_unicode if self._is_unicode else self._read_string
+    def read(self, n: int = 65535) -> bytes:
+        self.read: Callable[[int], bytes] = (  # type: ignore[method-assign]
```
wRAR, line 114: This code is wild: if you look closely you'll see that it assigns a different callable to itself. I have no idea if it's intended. I even think the behavior on the first and subsequent calls is different as `lstrip()` is only called on the first call.

Gallaecio: Maybe we should take the chance to fix it? e.g. assign `self._read = …` on `__init__`, and use `self._read(n).lstrip()` here?

wRAR: I think it could just use a local variable and not assign it anywhere. But, as usual, I wonder if fixing it will break anytning :)

On the other hand, it shouldn't and we have tests for it.

wRAR: Looks like it intentionally only strips on the first read, to skip the leading whitespace in the file (it's even possible that it's just for tests which have leading whitespace). `xmliter` instead uses regexes to find the document header.

In any case, we don't even advertise this function, even though it's used in some old 3rd-party code, I don't want to spend too much time on it.

scrapy/extensions/feedexport.py
```diff
@@ -282,15 +282,22 @@ def _store_in_thread(self, file):
...
+    ):
         u = urlparse(uri)
-        self.host = u.hostname
-        self.port = int(u.port or "21")
-        self.username = u.username
-        self.password = unquote(u.password or "")
-        self.path = u.path
-        self.use_active_mode = use_active_mode
-        self.overwrite = not feed_options or feed_options.get("overwrite", True)
+        assert u.hostname
```
Gallaecio, line 293: Maybe it would be better to raise a ValueError or similar here instead?

scrapy/utils/iterators.py
```diff
@@ -54,12 +75,20 @@ def xmliter(obj, nodename):
...
+    prefix: str = "x",
+) -> Generator[Selector, Any, None]:
     from lxml import etree
 
     reader = _StreamReader(obj)
     tag = f"{{{namespace}}}{nodename}" if namespace else nodename
-    iterable = etree.iterparse(reader, tag=tag, encoding=reader.encoding)
+    # technically, etree.iterparse only needs .read() AFAICS, but this is how it's typed
+    iterable = etree.iterparse(
+        cast("SupportsReadClose[bytes]", reader), tag=tag, encoding=reader.encoding
```
Gallaecio, line 90: :lipstick: Would it work to move the type hint to line 60/86 above instead of casting here?

PS: I’m not sure we need the comment.

Gallaecio (approved):
I love it when adding type hints includes bug fixes :)

scrapy/utils/iterators.py
wRAR: Casting hides `reader.encoding`.

PR #6005, opened 2023-08-09
More typing for scrapy/*.py

wRAR:


scrapy/dupefilters.py
```diff
@@ -47,10 +48,12 @@ def __init__(
         path: Optional[str] = None,
         debug: bool = False,
         *,
-        fingerprinter=None,
+        fingerprinter: Optional[RequestFingerprinter] = None,
```
Gallaecio, line 51: I’m not sure this and other uses of RequestFingerprinter for typing below are technically true.

Any object that implements the request fingerprinting interface, i.e. having a `fingerprint(request: Request) -> bytes` method, should be expected here, not only a RequestFingerprinter subclass.

wRAR: This is a bigger question to be honest.

There are many places where Scrapy provides an implementation (sometimes 2-3 implementations), allows users to provide a custom implementation and uses a setting to load the class (LOG_FORMATTER, STATS_CLASS, SPIDER_LOADER_CLASS, DOWNLOADER, SCHEDULER etc.). So far in these cases I used the default implementation for type hints, assuming that a custom implementation will have the same interface even when it's not a subclass of it. But after your comment I think we should instead explicitly define protocols and use those.

wRAR: I've added `RequestFingerprinterProtocol` and removed typing for SpiderLoader fields as we have ISpiderLoader and adding proper typing for interfaces should be done separately.

Gallaecio (approved):

PR #6014, opened 2023-08-14
Remove datetime.utcnow() usage

Laerte:
Make our datetime calls timezone-aware, `utcnow()` is deprecated and scheduled for removal, check why in this [article.](https://blog.ganssle.io/articles/2019/11/utcnow.html)


wRAR (approved):
Thanks!

Issue #5968, opened 2023-07-10
`response.json()` call makes unnecessary memory allocation


GeorgeA92:
## Summary

https://github.com/scrapy/scrapy/blob/52c072640aa61884de05214cb1bdda07c2a87bef/scrapy/http/response/text.py#L74-L82

As result of `response.text` call inside `response.json` - in addition to parsed object stored in `response._cached_decoded_json ` application will hold in RAM both `bytes` and `str` (from `response.text`) representation of response by the end of parse method (and related mw's methods).

`response.body` is already enough to receive parsed python dict object so we don't need to `response.text` (and definitely we don't need to hold response as str in RAM)
https://github.com/scrapy/scrapy/blob/52c072640aa61884de05214cb1bdda07c2a87bef/scrapy/http/response/text.py#L84-L93
Also we don't need to apply encoding identifier logic from `response.text` as according to JSON specs we expect only unicode input. 

<details><summary>reproducible code sample</summary>

```python
from json import loads, dumps
from sys import getsizeof
from scrapy.http import TextResponse

def create_record(id):
    return {"name": "quotes", "website": "quotes.toscrape.com", "type":"tutorial", "id": id}

def create_sample_json_body():
    return {'data': [create_record(r) for r in range(100)]}

# generating response body - closer to real json
body = dumps(create_sample_json_body()).encode('utf8')

a = TextResponse(url='https://quotes.toscrape.com/example_1', body=body[:], encoding='utf8', flags=['response.json()'])
b = TextResponse(url='https://quotes.toscrape.com/example_2', body=body[:], encoding='utf8', flags=['json.loads(response.text) + import'])
c = TextResponse(url='https://quotes.toscrape.com/example_3', body=body[:], encoding='utf8', flags=['json.loads(response.body) + import'])

a_parsed = a.json()
b_parsed = loads(b.text)
c_parsed = loads(c.body)

for response in [a, b, c]:
    print(f'response {response}')
    print(f'option: {response.flags[0]}')
    print(f'bytes:\t{getsizeof(response.body)}\t{response.body}'[:300])
    print(f'str:\t{getsizeof(response._cached_ubody)}\t{response._cached_ubody}'[:300])
    print(f'bytes+str size combined:\t{getsizeof(response.body) + getsizeof(response._cached_ubody)}\n')


```
</details>

<details><summary>log output</summary>

```
response <200 https://quotes.toscrape.com/example_1>
option: response.json()
[...]

PR #6016, opened 2023-08-17
Fix: response.json() call makes unnecessary memory allocation

jxlil:
Now `response.json()` uses `response.body` to retrieve the `dict` object, so the str response is no longer stored in RAM.

Closes https://github.com/scrapy/scrapy/issues/5968

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,23 @@
 Release notes
 =============
 
+.. _release-2.11.0:
+
+Scrapy 2.11.0 (to be released)
+------------------------------
+
+Backward-incompatible changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-   The :meth:`TextResponse.json <scrapy.http.TextResponse.json>` method now
+    requires the response to be in a valid JSON encoding (UTF-8, UTF-16, or
+    UTF-32).
+
+    If you need to deal with JSON documents in an invalid encoding, use
+    ``json.loads(response.text)`` instead.
+
+    (:issue:`5968`)
+
 .. _release-2.10.0:
 
 Scrapy 2.10.0 (2023-08-04)
--- a/scrapy/http/response/text.py
+++ b/scrapy/http/response/text.py
@@ -82,7 +82,7 @@ class TextResponse(Response):
         Deserialize a JSON document to a Python object.
         """
         if self._cached_decoded_json is _NONE:
-            self._cached_decoded_json = json.loads(self.text)
+            self._cached_decoded_json = json.loads(self.body)
         return self._cached_decoded_json
 
     @property
--- a/tests/test_http_response.py
+++ b/tests/test_http_response.py
@@ -844,7 +844,7 @@ class TextResponseTest(BaseResponseTest):
             with mock.patch("json.loads") as mock_json:
                 for _ in range(2):
                     json_response.json()
-                mock_json.assert_called_once_with(json_body.decode())
+                mock_json.assert_called_once_with(json_body)
 
 
 class HtmlResponseTest(TextResponseTest):


tests/test_http_response.py
```diff
@@ -844,7 +844,7 @@ def test_cache_json_response(self):
             with mock.patch("json.loads") as mock_json:
                 for _ in range(2):
                     json_response.json()
-                mock_json.assert_called_once_with(json_body.decode())
```
jxlil, line 847: Now,` json.loads()` receives bytes, so there's no need to use `decode()`

GeorgeA92 (approved):
From my side It looks OK.

However I am still not sure about.. additional tests that need (or not need) to implement here (I will prefer to wait for reviews from other contributors).

scrapy/http/response/text.py
```diff
@@ -82,7 +82,7 @@ def json(self):
         Deserialize a JSON document to a Python object.
         """
         if self._cached_decoded_json is _NONE:
-            self._cached_decoded_json = json.loads(self.text)
+            self._cached_decoded_json = json.loads(self.body)
```
kmike, line 85: The behavior is not exactly the same, if the response is not UTF-8, and properly indicates it.

From https://datatracker.ietf.org/doc/html/rfc7159#section-8.1:

> JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32. 

From https://docs.python.org/3/library/json.html#json.loads:

> s can now be of type [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) or [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray). The input encoding should be UTF-8, UTF-16 or UTF-32.

So we should be fine in most cases after this change. 

At the same time, SHALL in the RFC is not MUST (see https://datatracker.ietf.org/doc/html/rfc2119#section-3):

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

I'm fine with making the change, but we should add an entry to the changelog, which explains that the change is backwards incompatible, and that Scrapy stops support some of these weird cases to improve the efficiency for the vast majority of cases.

jxlil: Done! I've just added an entry to the **Backward-incompatible changes** section in the changelog. Please let me know if I need to correct something.

docs/news.rst
```diff
@@ -46,6 +46,13 @@ Backward-incompatible changes
     so the new default should keep the old behavior. (:issue:`872`, 
     :issue:`5847`)
 
+-   The :meth:`TextResponse.json <scrapy.http.TextResponse.json>` method now 
+    has the potential to affect results when the response body lacks UTF-8 
+    encoding. This represents a backwards-incompatible change intended to 
+    improve Scrapy's efficiency in the vast majority of scenarios. However, 
+    users still have the option of using ``json.loads(response.text)`` as an 
+    alternative to the ``response.json()`` shortcut. (:issue:`5968`)
```
Gallaecio, line 54: You are editing the release notes of the latest, already-released version of Scrapy.

We usually don’t edit release notes on PRs, and instead do it all at once during the release process. But this is actually a good idea, to make sure we do not forget about it.

So please remove this and add the following above the 2.10.0 header (I took the liberty to make some changes):

```rst
.. _release-2.11.0:

Scrapy 2.11.0 (to be released)
------------------------------

Backward-incompatible changes
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

-   The :meth:`TextResponse.json <scrapy.http.TextResponse.json>` method now
    requires the response to be in a valid JSON encoding (UTF-8, UTF-16, or
    UTF-32).

    If you need to deal with JSON documents in an invalid encoding, use
    ``json.loads(response.text)`` instead.

    (:issue:`5968`)
```

Gallaecio (approved):
Looks good to me, provided the release notes fix.

docs/news.rst
jxlil: Thanks! The release notes are already updated

jxlil:
A test failed, but I think it's not related to the changes in this PR

kmike:
Thanks @jxlil!

PR #6028, opened 2023-08-29
Fix the problem about _handleSignals 

dsppman:
Fix the problem that Twisted 23.8.0 and above remove the _handleSignals method and cause an error

### DIFF
--- a/scrapy/utils/ossignal.py
+++ b/scrapy/utils/ossignal.py
@@ -25,7 +25,10 @@ def install_shutdown_handlers(
     """
     from twisted.internet import reactor
 
-    reactor._handleSignals()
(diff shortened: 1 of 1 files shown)


scrapy/utils/ossignal.py
```diff
@@ -25,7 +25,10 @@ def install_shutdown_handlers(
     """
     from twisted.internet import reactor
 
-    reactor._handleSignals()
+    if hasattr(reactor, "_signalsFactory"):
+        reactor._signalsFactory()
```
wRAR, line 29: This is a no-op.

wRAR (changes requested):

Issue #6024, opened 2023-08-29
Problem with Twisted AttributeError: 'EPollReactor' object has no attribute '_handleSignals'


RobertoHigor:
### Description

Since the release of Twisted 23.8.0 recently, I've had problems running Scrapy. I had to fix to the previous version, Twisted==22.10.0

### Steps to Reproduce

1. Install Scrapy in a new environment
2. Run .start() from an CrawlerProcess object

**Expected behavior:** Scrapy to crawl

**Actual behavior:** 
![image](https://github.com/scrapy/scrapy/assets/14026017/4d769673-52ee-4039-887f-c810017d48e2)

**Reproduces how often:** Always in the new version

### Versions

Scrapy>=2.9.0

### Additional context

I've seen this fix on [stackoverflow](https://stackoverflow.com/questions/76995567/error-when-crawl-data-epollreactor-object-has-no-attribute-handlesignals)

mrtornado:
I'm having the same issue, but if you revert back to `pip install Twisted==22.10.0 ` it's working.

Ehsan-U:
I'm having a simiar issue as well. 
`AttributeError: 'AsyncioSelectorReactor' object has no attribute '_handleSignals'
`

GeorgeA92:
I assume that changes from twisted side that caused this issue came from this issue https://github.com/twisted/twisted/issues/11751 and this pull request https://github.com/twisted/twisted/pull/11752

wRAR:
This affects both `scrapy crawl` and direct usage of `CrawlerProcess`.

wRAR:
Looks like we are calling `reactor._handleSignals()` in `scrapy.utils.ossignal.install_shutdown_handlers()` so that we can override the Twisted handlers. Twisted itself calls `reactor._handleSignals()` in `_SignalReactorMixin._reallyStartRunning()`, I assume this happens later.

It seems that the equivalent code in new Twisted is in `twisted.internet._signals._WithSignalHandling.install()` but this is even more private and even more tied to internal logic so I hesitate to use it. It may be possible to redo our logic instead.

Unfortunately, this isn't covered by tests and deleting `reactor._handleSignals()` doesn't fail them :)

Overall, this looks like releasing a version with the Twisted version restriction is much more viable than quickly fixing this.

wRAR:
[...]

PR #6030, opened 2023-08-29
[wip] reactor handlesignals

GeorgeA92:
Fix to https://github.com/scrapy/scrapy/issues/6024

From my observations of scrapy code and related twisted pull request I conclude that call of `reactor._handleSignals()` or it's equivalent from new twisted version - are not needed inside `install_shutdown_handlers` to run scrapy.

At least on my local environment with both new/old twisted versions I didn't noticed anything unexpected.

### DIFF
--- a/scrapy/utils/ossignal.py
+++ b/scrapy/utils/ossignal.py
@@ -2,6 +2,9 @@ import signal
 from types import FrameType
 from typing import Any, Callable, Dict, Optional, Union
 
+from twisted import version as twisted_version
+from twisted.python.versions import Version
+
 # copy of _HANDLER from typeshed/stdlib/signal.pyi
 SignalHandlerT = Union[
     Callable[[int, Optional[FrameType]], Any], int, signal.Handlers, None
@@ -25,7 +28,8 @@ def install_shutdown_handlers(
     """
     from twisted.internet import reactor
 
-    reactor._handleSignals()
+    if twisted_version < Version("twisted", 23, 8, 0):
+        reactor._handleSignals()
     signal.signal(signal.SIGTERM, function)
     if signal.getsignal(signal.SIGINT) == signal.default_int_handler or override_sigint:
         signal.signal(signal.SIGINT, function)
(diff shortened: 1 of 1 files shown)


wRAR:
I think Twisted sets handlers for INT and BREAK conditionally but for TERM unconditionally, so I would expect our TERM handler to be overridden in at least some cases if this line is removed.

wRAR:
After thinking about it for some more I think it makes sense that calling or not calling this doesn't make a difference, whatever the call order. But can you please check if with old Twisted a simple `scrapy crawl` has the TERM handler last set by Scrapy or by Twisted?

GeorgeA92:
In `CrawlerProcess.start`.  `reactor.run` - called with `installSignalHandlers=False` which means that reactor will not apply it's default handlers. And as far as I understand in this case handlers from scrapy `install_shutdown_handlers` will be the only applied signal handlers
https://github.com/scrapy/scrapy/blob/a320e5f6a421ea3bae06d2f63d29bae9d327f580/scrapy/crawler.py#L389-L397

While if we will run scrapy with `CrawlerRunner` with simply `reactor.run()` as documented on https://docs.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process - reactor will run with reactor default signal handlers (and no scrapy `install_shutdown_handlers` as it called in `CrawlerProcess` only)

wRAR:
> reactor.run - called with installSignalHandlers=False

Oh!

Then yeah, it probably doesn't make sense. OTOH `twisted.internet.posixbase.PosixReactorBase._handleSignals()` also installs a SIGCHLD handler, but I don't know how useful is it for Scrapy.

Gallaecio:
> We may already have such test, I haven't checked.

I don’t think we do. It may be useful for https://github.com/scrapy/scrapy/issues/4749 as well.

wRAR:
https://stackoverflow.com/questions/74929947/twisted-application-ignoring-a-certain-unix-signal-is-it-possible suggests `reactor.addSystemEventTrigger('after', 'startup', install_your_handlers)`

PR #1580, opened 2015-11-02
Change extensions/spiders/settings initialisation order

jdemaeyer:
This is an implementation of @chekunkov's suggested changes from #1305. I will be uploading consecutive commits one-by-one in the next minutes so we can see which tests break at what point.

PR #1423, opened 2015-08-11
[MRG+1] Add ExecutionEngine.close() method

jdemaeyer:
Depending on what state the engine is in, there are different measures that need to be taken to shut it down gracefully (i.e. leaving the reactor clean): Stop the engine, close the spider, or close the downloader.

This PR adds a new method as a single entry point for shutting down the engine and integrates it into `Crawler.crawl()` for graceful error handling during the crawling process.

PR #6038, opened 2023-09-05
Change extensions/spiders/settings initialisation order, v2

### LINKED ISSUE #1305: Extensions, spider and settings initialization order
Current initialization order of spider, settings and all sorts of middlewares and extensions:

`Crawler.__init__` (see https://github.com/scrapy/scrapy/blob/master/scrapy/crawler.py#L32):
1. `self.spidercls.update_settings(self.settings)`
2. All extensions `__init__`
3. `self.settings.freeze()`

`Crawler.crawl` (see https://github.com/scrapy/scrapy/blob/master/scrapy/crawler.py#L70):
1. `Spider.__init__`
2. All downloader middlewares `__init__`
3. All spider middlewares `__init__`
4. All pipelines `__init__`

It's not clear - why extensions are initialized during `Crawler.__init__` and not in `Crawler.crawl`? Is it some legacy code untouched from times when it was possible to run several spider through the same set of extensions and middlewares?

I'm asking this because sometimes I feel like I want to change some crawl settings _after_ spider initialization and initialize middlewares only after that. For example I got request from customer to make it possible to set `CLOSESPIDER_TIMEOUT` based on passed spider argument. Due to `CloseSpider` implementation to support this I need to override it, disable default and set custom extension in settings. If initialization order was 

```
'spider init' -> 'update settings' -> 'settings.freeze' -> 'middlewares init' 
```

that task would be as easy as set CLOSESPIDER_TIMEOUT in `custom_settings`.

I don't speak about command line usage, `-s` does work in command line, but spiders often started not via command line - in Scrapy Cloud, ScrapyRT - it's not always possible to set per crawl settings in cases like that. It could also happen that spider has some logic to decide whether or not some setting should be set based on spider arguments - this is also the case when `-s` doesn't work well.

Based on above arguments I would like to propose different initialization order:

`Crawler.__init__`:
1. self.settings = settings.copy()

`Crawler.crawl`:
1. `Spider.__init__`
2. `spider.update_settings(self.settings)` - notice that in this case it isn't required for `update_settings` to be a `@classmethod`
3. `self.settings.freeze()`
4. All extensions `__init__`
5. All downloader middlewares `__init__`
6. All spider middlewares `__init__`
7. All pipelines `__init__`

What do you think about this proposal?

Discussion on this issue was originally started in https://github.com/scrapy/scrapy/pull/1276#issuecomment-110673089

### LINKED ISSUE #2392: Load settings dynamically on a per-spider basis
Feature request.

It would be good if scrapy had an easily accessible means of reading settings on a per-spider basis, and then making them accessible to the spider. From my many attempts to do this so far, in theory all of the components for this appear to be in place. Populating settings is already done:
https://doc.scrapy.org/en/latest/topics/settings.html#populating-the-settings - but then the problem is accessing them.
Ideally in a fashion that's compatible with scrapyd (so no calling `process.crawl(spider, my_settings)`).

**Ideally**: A project could have a generic project wide settings.py file with both the standard settings and any custom ones added by the developer. Then, using a command-line argument to indicate the settings file to use, the `__init__` method of the spider overrides specific settings, (much as `custom_settings` does), and these settings are then accessible throughout the spider via `self.settings` in the usual way.

**Current Problems**
*custom_settings*
Unfortunately `custom_settings` doesn't seem to be usable for this because it cannot be declared in `__init__`, but needs to be declared earlier.

*settings.py*
Currently, even if a user is willing to just use a different settings.py file entirely for each spider (thereby duplicating most of it), that's not readily possible either.

	os.environ['SCRAPY_SETTINGS_MODULE'] = 'myproject.settings'
	these_settings = get_project_settings()

The above only gets the settings into the variable `these_settings`, they're not used by the spider or accessible via `self.settings`.

**Desire for feature**
Based on StackOverFlow, this is something a lot of people want. The fact there are so many answers that are all so different shows there isn't a particular good way of doing it.

http://stackoverflow.com/questions/9814827/creating-a-generic-scrapy-spider
http://stackoverflow.com/questions/12996910/how-to-setup-and-launch-a-scrapy-spider-programmatically-urls-and-settings
http://stackoverflow.com/questions/35662146/dynamic-spider-generation-with-scrapy-subclass-init-error
http://stackoverflow.com/questions/40510526/how-to-load-different-settings-for-different-scrapy-spiders
http://stackoverflow.com/questions/2396529/using-one-scrapy-spider-for-several-websites

Being able to readily get allowed_domains and start_urls within it would also be good.

### LINKED ISSUE #3663: Make it possible to update settings in `__init__` or `from_crawler`
This issue might be related to https://github.com/scrapy/scrapy/issues/1305

I noticed that `settings` are frozen in https://github.com/scrapy/scrapy/blob/master/scrapy/crawler.py#L57
However, in a given project I had a requirement to change some settings based on some spider arguments. An alternative would be to write this spider as a base class and extend it from specific spiders setting the proper `settings`.
However, I think it would make sense to only freeze settings after the spider and other components were initialized. Or, provide some other entry point to configure settings based on arguments.
The other option is to use `-s` arguments, but in my case I was changing the `FEED_EXPORT_FIELDS` setting (https://docs.scrapy.org/en/latest/topics/feed-exports.html#std:setting-FEED_EXPORT_FIELDS).

Any thoughts here?

wRAR:
Continuation of #1580, without code changes for now.

Several tests fail, because you can no longer set `TWISTED_REACTOR` and `LOG_FILE`/`LOG_LEVEL` in `custom_settings`. Which is a problem. We can try moving the reactor installation to `crawl()`, but the logging problem was discussed in the old PR and doesn't have a good solution.

~~Also, I think we should deprecate running `Crawler.crawl()` multiple times as a separate uncontroversial PR, as proposed in the old one. #6040~~ done

Closes #1580, fixes #1305, fixes #2392, fixes #3663.


wRAR:
> We can try moving the reactor installation to crawl()

This seems to work, at least when nothing imports twisted.internet.reactor between creating a Crawler instance (or a CrawlerProcess istance)  and calling crawl() on it.

wRAR:
Direct link to the logging situation: https://github.com/scrapy/scrapy/pull/1580#issuecomment-153689154

In my experience changing logging per-spider is often done, e.g. if you have a large project where only some spiders need additional debugging. It can also be done via `-s` but that should still be supported here.

wRAR:
Maybe we just need to have both a classmethod that is called early, for backward compatibility and modifying things that need to be set early, and a new instance method that is called late.

In any case it's not clear when should the addons modify the settings.

scrapy/crawler.py
```diff
@@ -105,24 +104,8 @@ def __init__(
             crawler=self,
         )
```
Gallaecio, line 105: We might want to move the fingerprinter as well, specially if we move addons, to allow addons to change the fingerprinter.

I also wonder if there is more stuff we could move. Like logging the overriden settings, and *maybe* stats (not sure what that could break).

scrapy/crawler.py
```diff
@@ -71,7 +71,6 @@ def __init__(
 
         self.spidercls: Type[Spider] = spidercls
         self.settings: Settings = settings.copy()
-        self.spidercls.update_settings(self.settings)
 
         self.addons: AddonManager = AddonManager(self)
         self.addons.load_settings(self.settings)
```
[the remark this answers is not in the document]
Gallaecio: I am also thinking that addons could be a way to cleaning implement argument processing this way.

If we move addons after spider.update_settings, addons can use crawler.spider from from_crawler to process spider args. It would happen after spider.__init__, but other than that, it would be cleaner than needing code in a base spider, a spider mixin, or a spider itself.

scrapy/crawler.py
```diff
@@ -135,6 +118,26 @@ def crawl(self, *args: Any, **kwargs: Any) -> Generator[Deferred, Any, None]:
 
         try:
             self.spider = self._create_spider(*args, **kwargs)
+            self.spider.update_settings(self.settings)
```
Gallaecio, line 121: I wonder if update_settings running after `__init__` now could be an issue to some existing code, e.g. code reading settings from `__init__`.

Also, update_settings kind of becomes unnecessary, since spiders could now use crawler.settings from `__init__` instead for the same purpose. Still convenient, I guess.

I am thinking, although we do need setting processing to happen within `crawl`, maybe it could still happen before `spider.__init__`.

And we could expose *args and **kwargs to update_settings, with some backward-compatibility code to not pass them if unexpected, or maybe better, through a new method (of extensions?) that gets executed after setting processing but before spider init, gets access to argument and (unfrozen) settings, and can edit both.

Gallaecio:
It sounds good to me in general, although I am not 100% sure.

I imagine some existing `__init__` code might expect some components to exist beforehand, to access them from the crawler, but I am not sure how common that is. But since signals can be used to get another spider method called later on, maybe it is worth the backward-incompatible change?

Also, changing arguments based on final settings would not be possible before `__init__` with this approach, since `__init__` would be the one taking care of changing settings based on arguments. If we consider argument processing out of scope, then this is a non-issue. But what we choose here could limit what we can do when implementing argument processing support later on.

wRAR:
> I imagine some existing `__init__` code might expect some components to exist beforehand

Note that this doesn't work in the current PR code too, as `ExtensionManager` is initialized later.

> changing arguments based on final settings would not be possible before `__init__` with this approach, since `__init__` would be the one taking care of changing settings based on arguments

Not sure what is this use case?

Gallaecio:
> > changing arguments based on final settings would not be possible before `__init__` with this approach, since `__init__` would be the one taking care of changing settings based on arguments
> 
> Not sure what is this use case?

I am thinking of a future where we implement argument validation and conversion as an opt-in feature by exposing a setting for a new type of Scrapy component, in line with the fingerprinter setting, e.g. `ARG_PARSER`. The default parser could keep the current behavior, i.e. leave the arguments as they are, while a new parser could implement something like pydantic to error out in case of bad input, or convert strings to JSON or int.

My thinking is that this setting-defined component, which could have settings of its own to control its behavior (e.g. a setting to disable existing on error to instead only log a warning), should be executed before `__init__`, so that `__init__` would get those settings already processed, or not be called at all if argument processing exists earlier.

To enable such a future scenario, I am thinking we should have the following order:

1. existing settings editing
2. (future) argument parsing based on settings
3. setting editing based on (parsed) arguments
4. `__init__` getting final settings and arguments.

wRAR:
Another thing I've just found: `Spider.__init__()` does _not_ have access to the crawler: `spider.crawler` and `spider.settings` are set in `from_crawler()` after the instance is created. So the place to access them is an overridden `from_crawler()`, not an overridden `__init__()` (or some new method we can call on the instance at the end of the default `from_crawler()`.

wRAR:
There is a problem with moving the fingerprinter initialization inside `crawl()`: too many tests assume that `crawler.request_fingerprinter` exists just after creating the crawler.

I wonder if we want to split out a "prepare" method of `Crawler` that creates the spider instance etc. but doesn't start the engine, though I don't know if it would be useful outside tests.

wRAR:
I've implemented the change that keeps `update_settings()` as is and added a test that modifies settings in `from_crawler()`. As `from_crawler()` receives argument values it should cover some of the use cases we want.

wRAR:
We discussed this with @kmike and decided that we want to move more things into crawl() and fix tests to call it there.  Also that we don't want to change arguments based on settings.

wRAR:
It may be possible to extract some test changes and merge them separately if we decide to keep this approach with `crawler.crawl()` in tests, to simplify reviewing.

Gallaecio:
> Not sure yet what to do with this, as the engine is right to assume that the crawler is initialized properly and we moved most of that initialization to crawl().

Would it make sense to add an `load_settings` parameter to `Crawler`, `False` by default, and set it to `True` only for that scenario, causing the setting loading code to run in `__init__` instead of in `crawl`?

Gallaecio:
Or, to keep things simple, we keep the code that you moved in a private method of Crawler, and on the shell code we call it manually:

```
crawler = self.crawler_process._create_crawler(spidercls) 
crawler._load_settings()
```

scrapy/crawler.py
```diff
@@ -74,57 +74,19 @@ def __init__(
...
-        logging.root.addHandler(handler)
-
-        d = dict(overridden_settings(self.settings))
-        logger.info(
-            "Overridden settings:\n%(settings)s", {"settings": pprint.pformat(d)}
-        )
-
         if get_scrapy_root_handler() is not None:
             # scrapy root handler already installed: update it with new settings
             install_scrapy_root_handler(self.settings)
```
kmike, line 81: should we call it again in .crawl?

tests/test_downloadermiddleware.py
```diff
@@ -3,33 +3,33 @@
...
+        self.spider = self.crawler.spider
         self.mwman = DownloaderMiddlewareManager.from_crawler(self.crawler)
-        # some mw depends on stats collector
-        self.crawler.stats.open_spider(self.spider)
-        return self.mwman.open_spider(self.spider)
+        yield self.mwman.open_spider(self.spider)
 
     def tearDown(self):
-        self.crawler.stats.close_spider(self.spider, "")
         return self.mwman.close_spider(self.spider)
```
kmike, line 30: Should we close the crawler itself? 
A related question: instead of creating DownloaderMiddlewareManager, should we get an instance which is created by the crawler?

tests/test_downloadermiddleware_httpcache.py
```diff
@@ -2,27 +2,31 @@
...
 
+    @inlineCallbacks
     def setUp(self):
         self.yesterday = email.utils.formatdate(time.time() - 86400)
         self.today = email.utils.formatdate()
         self.tomorrow = email.utils.formatdate(time.time() + 86400)
-        self.crawler = get_crawler(Spider)
-        self.spider = self.crawler._create_spider("example.com")
+        self.crawler = get_crawler(NoRequestsSpider)
+        yield self.crawler.crawl()
```
kmike, line 28: should we stop the crawler in tearDown?

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -38,12 +39,13 @@
...
 
 
-class HttpCompressionTest(TestCase):
+class HttpCompressionTest(unittest.TestCase):
+    @inlineCallbacks
     def setUp(self):
-        self.crawler = get_crawler(Spider)
-        self.spider = self.crawler._create_spider("scrapytest.org")
+        self.crawler = get_crawler(NoRequestsSpider)
+        yield self.crawler.crawl()
```
kmike, line 46: should we stop it in tearDown?

tests/test_downloadermiddleware.py
wRAR: > Should we close the crawler itself?

The crawler should be "closed" when `yield crawl()` returns.

> instead of creating DownloaderMiddlewareManager, should we get an instance which is created by the crawler?

That instance is created by the engine, so we can do it if we decide to keep calling `crawler.crawl()` (which I was going to revert and just call `_load_settings()`).

tests/test_downloadermiddleware_httpcompression.py
```diff
@@ -374,13 +376,15 @@ def test_process_response_head_request_no_decode_required(self):
...
+class HttpCompressionSubclassTest(unittest.TestCase):
+    @inlineCallbacks
     def test_init_missing_stats(self):
         class HttpCompressionMiddlewareSubclass(HttpCompressionMiddleware):
             def __init__(self):
                 super().__init__()
 
-        crawler = get_crawler(Spider)
+        crawler = get_crawler(NoRequestsSpider)
+        yield crawler.crawl()
```
kmike, line 387: again, should the crawler be stopped? Or is it not required, and there is no clean-up needed?

kmike: (there are a few more places; I won't be adding comments about them, as they seem redundant)

tests/test_engine.py
```diff
@@ -420,12 +420,16 @@ def _assert_signals_caught(self, run: CrawlerRun):
 
     @defer.inlineCallbacks
     def test_close_downloader(self):
-        e = ExecutionEngine(get_crawler(TestSpider), lambda _: None)
+        crawler = get_crawler(TestSpider)
+        yield crawler.crawl()
+        e = ExecutionEngine(crawler, lambda _: None)
         yield e.close()
```
kmike, line 426: 1. Why is this test called test_close_downloader?
2. Why is the engine closed, but the crawler isn't? Would it make more sense to pick engine from the crawler, and the close the crawler?
3. tbh, I don't quite get what is this test testing

tests/test_downloadermiddleware.py
kmike: > The crawler should be "closed" when yield crawl() returns.

oh, right; I think most of my comments to the PR are irrelevant :)

tests/test_dupefilters.py
```diff
@@ -2,22 +2,25 @@
...
 from scrapy.dupefilters import RFPDupeFilter
 from scrapy.http import Request
 from scrapy.utils.python import to_bytes
 from scrapy.utils.test import get_crawler
-from tests.spiders import SimpleSpider
+from tests.spiders import NoRequestsSpider
 
 
+@inlineCallbacks
 def _get_dupefilter(*, crawler=None, settings=None, open=True):
```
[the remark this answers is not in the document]
wRAR: Same as with `DownloaderMiddlewareManager`, it's created via the engine, so it would make sense if we run `crawler.crawl()` and doesn't if we don't, except that here we also open and close the dupefilter explicitly, which can be problematic if it was already opened and closed during `crawler.crawl()`.

tests/test_downloadermiddleware.py
kmike: > The crawler should be "closed" when yield crawl() returns.

Hm, why would it be guaranteed? It seems crawler.crawl waits until `self.engine.start` is successful, but it doesn't wait for the crawling to be stopped.

tests/test_engine.py
wRAR: > Why is this test called test_close_downloader?

Initially there were three tests added in #1423: `test_close_downloader`, `test_close_spiders_downloader` and `test_close_engine_spiders_downloader` which called and tested different open/close things, then the last two were removed when `ExecutionEngine.open_spiders` was removed.

> tbh, I don't quite get what is this test testing

I guess it just tests that `ExecutionEngine.close()` doesn't produce exceptions.

tests/test_downloadermiddleware.py
wRAR: `ExecutionEngine.start()` returns a deferred that fires when the crawling finishes.

kmike: Overall, it seems we're doing some strange things in the tests, which is not very clean: 

* crawler is started and stopped(?)
* then some internal components are created using this stopped crawler (which we don't have a well-defined behavior for)

It seems we're not using our public APIs (or any well-defined APIs), and instead just trying to get the components in usable state, using whatever the current inernal behavior is.

I haven't understood the recent proposals about `_load_settings`, maybe it makes sense to try them.

Another alternative would be to try to use public APIs. If it's not possible with the current .crawl API, it could be an argument to improve Crawler API slightly, e.g. split .crawl() into a few methods, e.g. to init the crawler and actually start it (.crawl sould be kept as a shortcut to do both).

wRAR: > Overall, it seems we're doing some strange things in the tests, which is not very clean:

Yes, this is my main problem with the current PR code and the main reason I want a separate initialization function.

> Another alternative would be to try to use public APIs. If it's not possible with the current .crawl API, it could be an argument to improve Crawler API slightly, e.g. split .crawl() into a few methods, e.g. to init the crawler and actually start it (.crawl sould be kept as a shortcut to do both).

That's what I proposed initially in https://github.com/scrapy/scrapy/pull/6038#issuecomment-1710108413 and recently in https://github.com/scrapy/scrapy/pull/6038#issuecomment-1713601583

Though I see it as a private function that includes all that initialization we are moving from `__init__`, not something we expect the user to call. If we want public functions, I think it's better to reshuffle the code so that we have e.g. only `__init__()` and `crawl()`, not 3 functions, but I'm not sure how useful would that be.

kmike: To be more concrete, I was thinking about doing it like this:

```py
    def init_crawl(self):  # we can also make it private
        # + exception and flag handling
        self.spider = self._create_spider(*args, **kwargs)
        # all the new initialization code moved from __init__ goes here?

    @defer.inlineCallbacks
    def start_engine(self):  # we can also make it private
        # + exception and flag handling
        start_requests = iter(self.spider.start_requests())
        yield self.engine.open_spider(self.spider, start_requests)
        yield defer.maybeDeferred(self.engine.start)

    @defer.inlineCallbacks
    def crawl(self):
        self.init_crawl()
        yield self.start_engine()
```

Is the idea to just move init_crawl content to `__init__`? On a first sight, it looks more risky, as it'd start calling `spider.__init__` and `spider.from_crawler`, so we'd need to check if that's backwards compatible, and what kind of issues it can raise.

kmike: 2 separate private functions look better to me than 1 private function (_init_crawl + keep engine starting code in `crawl()`), because this way we'd 

1) keep the code in crawl() on the same level of abstraction (no low-level vs high-level code)
2) allow to do the init, then do something else, and then start the crawl (though not sure what's exactly the use case)

wRAR: The current idea is

```python
    def _load_settings(self) -> None:
        if self._settings_loaded:
            return
        self._settings_loaded = True
        # all the new initialization code moved from __init__ goes here

    def crawl(self, *args: Any, **kwargs: Any) -> Generator[Deferred, Any, None]:
        # flag handling
        try:
            self.spider = self._create_spider(*args, **kwargs)
            self._load_settings()
            # etc.
```

so the `Crawler` API is the same but `scrapy shell` and `scrapy.utils.test.get_crawler()` can call `crawler._load_settings()` (and `crawler.spider = crawler._create_spider()` if needed) to get an initialized Crawler instance. This doesn't create or start the engine, but the old API didn't either.

We could move `_create_spider` and `_create_engine` there too, to slightly simplify the code in `scrapy shell` and to allow tests to get more components from the engine, though this state may still be not suitable for some tests (e.g. `engine.scheduler` is only assigned in `engine.open_spider()`) and also making such changes to tests is out of scope.

> Is the idea to just move init_crawl content to __init__

I don't want to do this but I proposed it as a way to have no more than 2 public methods.

kmike: This sounds fine. 
Should it be called something like _apply_settings though?

Gallaecio: No opinion on the name (I proposed load, but apply sounds good too).

wRAR: @kmike do you think it should create the spider and the engine instances?

kmike: @wRAR no strong opinion on that; please follow the approach you prefer. I haven't checked all code which is relevant (e.g. scrapy shell), while you have checked it :)

wRAR:
(pypy test failures related to the reactor need to be fixed before merging)

docs/topics/settings.rst
```diff
@@ -98,6 +98,26 @@ and settings set there should use the "spider" priority explicitly:
...
+    import scrapy
+
+
+    class MySpider(scrapy.Spider):
+        name = "myspider"
+
+        @classmethod
+        def from_crawler(cls, crawler, *args, **kwargs):
+            spider = super().from_crawler(crawler, *args, **kwargs)
+            spider.settings.set("SOME_SETTING", kwargs["value"], priority="spider")
```
Gallaecio, line 118: Since it would only be 1 more line, I wonder if the example should ensure `kwargs["value"]` actually exists before using it, to encourage best practices.

Gallaecio (approved):

wRAR:
PyPy tests fail because on PyPy exceptions raised in `Crawler.crawl()` are silently discarded or handled somewhere while on CPython they are unhandled and show in the log. This happens on master as well.

docs/topics/settings.rst
```diff
@@ -98,6 +98,27 @@ and settings set there should use the "spider" priority explicitly:
...
+
+
+    class MySpider(scrapy.Spider):
+        name = "myspider"
+
+        @classmethod
+        def from_crawler(cls, crawler, *args, **kwargs):
+            spider = super().from_crawler(crawler, *args, **kwargs)
+            if "value" in kwargs:
+                spider.settings.set("SOME_SETTING", kwargs["value"], priority="spider")
```
kmike, line 119:
```suggestion
            if "some_argument" in kwargs:
                spider.settings.set("SOME_SETTING", kwargs["some_argument"], priority="spider")
```

docs/topics/spiders.rst
```diff
@@ -136,6 +136,13 @@ scrapy.Spider
        attributes in the new instance so they can be accessed later inside the
        spider's code.
 
+       .. versionchanged:: VERSION
+
+           The settings in ``crawler.settings`` can now be modified, which is
+           handy if you want to modify them based on arguments. As a
+           consequence, these settings aren't the final values as they can be
+           modified later by e.g. :ref:`add-ons <topics-addons>`.
```
kmike, line 144: Something like this maybe? I'm not sure about "when the spider starts" point, is there a more precise definition?

```suggestion
           The settings in ``crawler.settings`` can be modified in the 
           ``from_crawler`` method, which is handy if you want to modify 
           them based on spider arguments. As a consequence, the settings 
           aren't the final values as they can be modified later by 
           e.g. :ref:`add-ons <topics-addons>`. The settings
           become final when the spider starts.
```

scrapy/crawler.py
```diff
@@ -72,42 +72,55 @@ def __init__(
...
+        self.request_fingerprinter: Optional[RequestFingerprinter] = None
+        self.spider: Optional[Spider] = None
+        self.engine: Optional[ExecutionEngine] = None
 
+    def _update_root_log_handler(self) -> None:
         if get_scrapy_root_handler() is not None:
             # scrapy root handler already installed: update it with new settings
             install_scrapy_root_handler(self.settings)
+
+    def _load_settings(self) -> None:
```
kmike, line 97: a nitpick, about the method name :) _load_settings suggests that this method loads the settings (from somewhere?). It seems what this method does is to load/create the components based on the settings, not loading the settings themselves.

This seems to apply to AddonManager.load_settings name as well, btw. Not sure if it's too late to change this; it seems we could, as we haven't made AddonManager a part of the public API.

Some name suggestions:

* _apply_settings

or, if -apply_settings doesn't sound right, something like 

* _load_components
* _create_components

wRAR: _apply_settings is fine and was proposed before

scrapy/addons.py
```diff
@@ -19,8 +19,8 @@ def __init__(self, crawler: "Crawler") -> None:
         self.crawler: "Crawler" = crawler
         self.addons: List[Any] = []
 
-    def load_settings(self, settings: Settings) -> None:
-        """Load add-ons and configurations from a settings object.
+    def apply_settings(self, settings: Settings) -> None:
```
kmike, line 22: heh, should we name this method update_settings, based on what it actually does?

docs/topics/spiders.rst
Gallaecio: Rather than discussing when the settings become final, we could change the angle and instead tell users how they can get the earliest possible access to the final settings from a spider (start_requests? some signal?).

scrapy/crawler.py
```diff
@@ -72,42 +72,55 @@ def __init__(
...
+        self.spider: Optional[Spider] = None
+        self.engine: Optional[ExecutionEngine] = None
 
+    def _update_root_log_handler(self) -> None:
         if get_scrapy_root_handler() is not None:
             # scrapy root handler already installed: update it with new settings
             install_scrapy_root_handler(self.settings)
+
+    def _apply_settings(self) -> None:
+        if self._settings_loaded:
```
kmike, line 98: Can it be simplified/improved to `if self.settings.frozen`?

kmike: Also, if it's called when the settings are frozen, should it be a warning?

scrapy/addons.py
wRAR: Well, the old and the current name reflect that it reads the `ADDONS` setting, not that it runs add-ons' `update_settings` methods, so I'm not sure.

kmike: Isn't updating the settings the single useful piece of work this method does?

scrapy/crawler.py
wRAR: Or an error?

kmike: Right, and error is even better, if there is no use case for running this method twice

scrapy/addons.py
kmike: Technically, this has nothing to do with this PR, so if there is no easy agreement (cc @Gallaecio), let's revert the name change and handle it separately :)

scrapy/crawler.py
wRAR: Though when designing it I thought that we need a check so that we can safely run it twice, e.g. directly and then via crawl(). This only matters in tests as users shouldn't run _apply_settings() directly, and either a warning or an error will affect them, as we indeed have tests that call `get_crawler()` (which calls _apply_settings() so that you can get an initialized crawler) and then call `crawl()`.

docs/topics/spiders.rst
wRAR: I think `ExecutionEngine()` doesn't fire any signals so that would be `start_requests()`.

wRAR:
> PyPy tests fail because on PyPy exceptions raised in `Crawler.crawl()` are silently discarded or handled somewhere while on CPython they are unhandled and show in the log. This happens on master as well.

We rely on the "Unhandled error in Deferred" messages that are printed when such Deferreds are garbage collected (we call `CrawlerProcess.crawl()` without storing the Deferred that it returns), and on PyPy something works in some other way and they are not printed. So in tests where we expect an unhandled error we should instead add an explicit errback.

tests/CrawlerProcess/asyncio_enabled_reactor_different_loop.py
```diff
@@ -24,5 +24,6 @@ def start_requests(self):
         "ASYNCIO_EVENT_LOOP": "uvloop.Loop",
     }
 )
-process.crawl(NoRequestsSpider)
+d = process.crawl(NoRequestsSpider)
+d.addErrback(lambda failure: failure.printTraceback())
```
kmike, line 28: I wonder if we should do it by default for CrawlerRunner, maybe with `CrawlerRunner.__init__` flag to disable printing.

kmike: though let's not solve it here :)

kmike: The issue here is that we're not following https://docs.scrapy.org/en/latest/topics/practices.html#run-scrapy-from-a-script in the tests because it's problematic. So, it may be problematic for our users as well.

kmike:
The PR looks good to me, but we should make the tests pass.

wRAR:
`failure.printTraceback()` doesn't work on Twisted 18.9.0 on Python 3.8+. This is fixed in Twisted 19.7.0, and the first Twisted version that officially supports Python 3.8 is even higher, 21.2.0. We could replace the logging code in these tests or we could bump the minimal Twisted version, even though spiders work with 18.9.0 too.

scrapy/addons.py
Gallaecio: I don’t care, to me this method could be private and then the name does not matter that much :)

tests/CrawlerProcess/asyncio_enabled_reactor_different_loop.py
wRAR: I agree!

wRAR:
Though it looks like using `twisted.python.log.err()` still prints the exception we need so we can just switch to it.

kmike (approved):
🚀

PR #6040, opened 2023-09-06
Deprecate running Crawler.crawl() twice.

### LINKED ISSUE #1587: Multiple crawls with the same crawler
While we don't support using the same crawler for multiple spiders, it is currently possible to run the same crawler (with the same spider) twice, possibly with different spider arguments. I.e., this works:

``` python
crawler = Crawler(myspidercls)
crawler.crawl(my_sparg='blah')
crawler.crawl(my_sparg='other_blah')
```

This is done in some of our tests ([1](https://github.com/scrapy/scrapy/blob/master/tests/test_crawl.py#L63), [2](https://github.com/scrapy/scrapy/blob/master/tests/test_crawl.py#L132), [3](https://github.com/scrapy/scrapy/blob/master/tests/test_downloader_handlers.py#L299)). Are these tests legacy code or is support for multiple crawls part of the public API (see also [this discussion](https://github.com/scrapy/scrapy/pull/1580#issuecomment-154209093))?

wRAR:
Fixes #1587

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -124,6 +124,7 @@ class Crawler:
 
         self.settings.freeze()
         self.crawling: bool = False
+        self._started: bool = False
         self.spider: Optional[Spider] = None
         self.engine: Optional[ExecutionEngine] = None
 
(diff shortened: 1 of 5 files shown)


scrapy/crawler.py
```diff
@@ -124,14 +124,21 @@ def __init__(
...
     @inlineCallbacks
     def crawl(self, *args: Any, **kwargs: Any) -> Generator[Deferred, Any, None]:
         if self.crawling:
             raise RuntimeError("Crawling already taking place")
-        self.crawling = True
+        if self._started:
+            warnings.warn(
+                "Running Crawler.crawl() more than once is deprecated.",
+                ScrapyDeprecationWarning,
+                stacklevel=2,
```
wRAR, line 139: With 2 it's still not helpful as it points to twisted/internet/defer.py (as inlineCallbacks turns into several frames), but I'm not sure if it's possible to set a correct number here.

Gallaecio (approved):

PR #6043, opened 2023-09-11
Remove-docs-#5920

### LINKED ISSUE #5920: Remove docs for other projects from the Scrapy docs
We have some auto* statements for things from other projects, like itemadapter in https://github.com/scrapy/scrapy/blob/92f86fab062d179fd68590f1be6a24ca13ec33b6/docs/topics/items.rst?plain=1#L404 and parsel in https://github.com/scrapy/scrapy/blob/92f86fab062d179fd68590f1be6a24ca13ec33b6/docs/topics/selectors.rst#L1037.

We should find these and remove them, instead linking to the other projects docs and maybe adding some explanations locally.

munirahd:
Description:
This PR addresses issue #5920.

Changes:
Replaced direct code references with a brief explanation about itemadapter.
Added an external link directing readers to the official itemadapter documentation.
Did the same for XPath extensions, linking to the Parsel Documentation.

This is my first Open Source PR. If approved, I'm ready to proceed with replacing all the auto-linking. Feedback and guidance are appreciated!

Resolves #5920

### DIFF
--- a/docs/topics/items.rst
+++ b/docs/topics/items.rst
@@ -401,7 +401,7 @@ In code that receives an item, such as methods of :ref:`item pipelines
 :func:`~itemadapter.is_item` function to write code that works for
 any :ref:`supported item type <item-types>`:
 
-.. autoclass:: itemadapter.ItemAdapter
+The itemadapter.ItemAdapter class is a wrapper for data container objects, providing a common interface to handle objects of different types in an uniform manner, regardless of their underlying implementation. see the: `itemadapter.ItemAdapter Documentation <https://github.com/scrapy/itemadapter>`_.
 
 .. autofunction:: itemadapter.is_item
 
--- a/docs/topics/selectors.rst
+++ b/docs/topics/selectors.rst
@@ -1034,7 +1034,7 @@ described with CSS selectors.
 
 Parsel also simplifies adding your own XPath extensions.
 
-.. autofunction:: parsel.xpathfuncs.set_xpathfunc
+For more details about the XPath extensions, see the `Parsel Documentation <https://parsel.readthedocs.io/en/latest/usage.html#parsel.xpathfuncs.set_xpathfunc>`_.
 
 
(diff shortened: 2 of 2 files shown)


docs/topics/items.rst
```diff
@@ -401,7 +401,7 @@ In code that receives an item, such as methods of :ref:`item pipelines
 :func:`~itemadapter.is_item` function to write code that works for
 any :ref:`supported item type <item-types>`:
 
-.. autoclass:: itemadapter.ItemAdapter
+The itemadapter.ItemAdapter class is a wrapper for data container objects, providing a common interface to handle objects of different types in an uniform manner, regardless of their underlying implementation. see the: `itemadapter.ItemAdapter Documentation <https://github.com/scrapy/itemadapter>`_.
 
 .. autofunction:: itemadapter.is_item
```
wRAR, line 406: This needs to be removed too.

And I think it's fine to just remove both and replace the semicolon with a period in the preceding statement, as both `itemadapter.ItemAdapter` and `itemadapter.is_item` are already linked in that statement.

docs/topics/selectors.rst
```diff
@@ -1034,7 +1034,7 @@ described with CSS selectors.
 
 Parsel also simplifies adding your own XPath extensions.
 
-.. autofunction:: parsel.xpathfuncs.set_xpathfunc
+For more details about the XPath extensions, see the `Parsel Documentation <https://parsel.readthedocs.io/en/latest/usage.html#parsel.xpathfuncs.set_xpathfunc>`_.
```
wRAR, line 1037: I think the link should just be attached to "adding your own XPath extensions". And I think you should be able to use a Sphinx reference for it instead of an explicitly written link.

wRAR, line 1037: I think the link should just be attached to "adding your own XPath extensions". And I think you should be able to use a Sphinx reference for it instead of an explicitly written link.

PR #6050, opened 2023-09-14
Improve the backwards compatibility for RetryMiddleware.EXCEPTIONS_TO_RETRY

### LINKED ISSUE #6049: `downloadermiddlewares.retry.BackwardsCompatibilityMetaclass` does not provide backward compatibility for middleware instances
# Description

Previously, `EXCEPTIONS_TO_RETRY` was an attribute of `RetryMiddleware`. This allows:
- `RetryMiddleware` subclasses could access `EXCEPTIONS_TO_RETRY` via `cls.EXCEPTIONS_TO_RETRY`.
- `RetryMiddleware` instances and instances of its subclasses could access `EXCEPTIONS_TO_RETRY` via `self.EXCEPTIONS_TO_RETRY`.

In 2.10 `EXCEPTIONS_TO_RETRY` was removed and added as a property to `BackwardsCompatibilityMetaclass`. This added compatibility only for the first point.

# Steps to Reproduce

```python
class MyRetryMiddleware(RetryMiddleware):

    def process_exception(self, request, exception, spider):
        if isinstance(exception, self.EXCEPTIONS_TO_RETRY) and not request.meta.get('dont_retry', False):
            # update request
            return self._retry(request, exception, spider)
```

# Expected behavior

A warning about `EXCEPTIONS_TO_RETRY` deprecation.

# Actual behavior

AttributeError: 'MyRetryMiddleware' object has no attribute 'EXCEPTIONS_TO_RETRY'

# Versions

```
Scrapy       : 2.10.1
lxml         : 4.9.3.0
libxml2      : 2.10.3
cssselect    : 1.2.0
parsel       : 1.8.1
w3lib        : 2.1.2
Twisted      : 22.10.0
Python       : 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)]
pyOpenSSL    : 23.2.0 (OpenSSL 3.1.2 1 Aug 2023)
cryptography : 41.0.3
```

wRAR:
Fixes #6049 (with the code provided by the reporter)

### DIFF
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -23,13 +23,11 @@ from scrapy.utils.response import response_status_message
 
 retry_logger = getLogger(__name__)
 
-DEPRECATED_ATTRIBUTE = "EXCEPTIONS_TO_RETRY"
-
 
 def backwards_compatibility_getattr(self, name):
-    if name == DEPRECATED_ATTRIBUTE:
+    if name == "EXCEPTIONS_TO_RETRY":
         warnings.warn(
-            f"Attribute RetryMiddleware.{DEPRECATED_ATTRIBUTE} is deprecated. "
+            "Attribute RetryMiddleware.EXCEPTIONS_TO_RETRY is deprecated. "
             "Use the RETRY_EXCEPTIONS setting instead.",
             ScrapyDeprecationWarning,
             stacklevel=2,
@@ -146,7 +144,7 @@ class RetryMiddleware(metaclass=BackwardsCompatibilityMetaclass):
         self.priority_adjust = settings.getint("RETRY_PRIORITY_ADJUST")
 
         try:
-            self.exceptions_to_retry = self.__getattribute__(DEPRECATED_ATTRIBUTE)
+            self.exceptions_to_retry = self.__getattribute__("EXCEPTIONS_TO_RETRY")
         except AttributeError:
             # If EXCEPTIONS_TO_RETRY is not "overridden"
             self.exceptions_to_retry = tuple(
--- a/scrapy/downloadermiddlewares/retry.py
+++ b/scrapy/downloadermiddlewares/retry.py
@@ -23,12 +23,13 @@ from scrapy.utils.response import response_status_message
 
 retry_logger = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


scrapy/downloadermiddlewares/retry.py
```diff
@@ -37,6 +38,13 @@ def EXCEPTIONS_TO_RETRY(cls):
             load_object(x) if isinstance(x, str) else x
             for x in Settings().getlist("RETRY_EXCEPTIONS")
         )
+    raise AttributeError(
+        f"{self.__class__.__name__!r} object has no attribute {name!r}"
+    )
```
Gallaecio, line 43: Hmm… When is this raised?

wRAR: When you try to access an attribute that doesn't exist, like when `__getattr__()` isn't defined.

Gallaecio (approved):

scrapy/downloadermiddlewares/retry.py
```diff
@@ -23,12 +23,13 @@
...
-class BackwardsCompatibilityMetaclass(type):
-    @property
-    def EXCEPTIONS_TO_RETRY(cls):
+
+def backwards_compatibility_getattr(self, name):
+    if name == DEPRECATED_ATTRIBUTE:
         warnings.warn(
-            "Attribute RetryMiddleware.EXCEPTIONS_TO_RETRY is deprecated. "
+            f"Attribute RetryMiddleware.{DEPRECATED_ATTRIBUTE} is deprecated. "
             "Use the RETRY_EXCEPTIONS setting instead.",
```
kmike, line 33: I think we should either have both old/new attribute names as constants, or none of them.

scrapy/downloadermiddlewares/retry.py
Prometheus3375: I was thinking to define super call here, but it only makes sense if there is multiple inheritance and 2+ superclasses redefine `getattr` which is not common in middlewares.

But making super call requires function definition to be inside class (to fill `super()` with correct arguments) or to pass correct class as first argument. Since the function is shared between different classes, I decided to just raise `AtrributeError` as super call will do in regular single inheritance situation (actually super call is really weird because neither `object` nor `type` have defined `getattr` and yet super call of `getattr` somehow works).

PR #6042, opened 2023-09-08
Create SECURITY.md

benharvie:
To let the repository confirm scrapy-security@googlegroups.com as its security contact.

Resolves #5364

PR #6051, opened 2023-09-14
SECURITY.md pointing to github issues

### LINKED ISSUE #5364: Add SECURITY.md
Hey there!

I belong to an open source security research community, and a member (@ranjit-git) has found an issue, but doesn’t know the best way to disclose it.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

kmike:
Resolves #5364, resolves #6042.


SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.10.x     | :white_check_mark: |
+| < 2.10.x   | :x:                |
```
wRAR, line 8: Should we support the previous version, at least for some time after releasing the new one?

Gallaecio: Upgrading from a minor to another should not be a big deal, so it may be best to simply ask people to upgrade to the next minor to get the security fix.

Gallaecio: In any case, the current content reflects the current policy (minus maintaining support for 1.8.x, which maybe we should stop). So I think we could merge as is and discuss changing the policy separately, e.g. maybe having some LTS version or something.

wRAR (approved):

PR #6063, opened 2023-09-23
add support for multiple-compressed responses

vishesh10:
### Description
- Added support for multi-compressed responses.
- Added test case.

Resolves #5143, resolves #5964.


scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -83,6 +83,38 @@ def process_response(self, request, response, spider):
 
         return response
 
+    def _handle_encoding(self, body, content_encoding):
+        content_to_decode = body
+
+        separator = b","
+        for encodings in reversed(content_encoding):
+            if separator in encodings:
```
Gallaecio, line 91: What about handling the comma issue earlier? i.e. (untested)

```
encodings = reversed(item.strip() for value in content_encoding for item in value.split(b","))
```

That should simplify the later code.

Gallaecio: Oh, I see, there’s the editing of `Content-Encoding` to mind as well.

Gallaecio (commented):
It would be good to also test the case where multiple headers are used instead of a single header with comma-separated values. I suspect that will break the expectation that `Content-Encoding` should be empty after decoding.

I actually wonder if it wouldn’t be best to keep the header with the original value from the source website, and instead control whether or not a response has been decoded through e.g. a [response flag](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response.flags). But that would be a backward-incompatible change that might need some prior discussion, so I think it is OK if we consider this out of scope here, and only focus on keeping the existing behavior that `Content-Encoding` gets removed after decoding.

Do mind that according to the existing logic `Content-Encoding` should be kept in case of unsupported encodings, e.g. `Content-Encoding: gzip, foo, deflate` should decode only `deflate` and the resulting response should have  `Content-Encoding: gzip, foo`. A test for that scenario would be great.

vishesh10:
Hi @Gallaecio,
Thanks for reviewing the code. I have added the test case for `Content-Encoding: gzip, foo, deflate`.

Regarding the multiple `Content-Encoding` headers scenario, the `Headers` class is actually a dict so I wonder if multiple `Content-Encoding` headers can be passed.

Gallaecio:
> Regarding the multiple `Content-Encoding` headers scenario, the `Headers` class is actually a dict so I wonder if multiple `Content-Encoding` headers can be passed.

I think values can be lists, though, which translates into multiple headers. I know that’s the case in some parts of Scrapy, not sure if that’s the case here, but might be worth exploring if you can.

wRAR:
> the Headers class is actually a dict so I wonder if multiple Content-Encoding headers can be passed.

It's not a simple dict, it has `getlist()`.

vishesh10:
Added support for multiple `Content-Encoding` headers and added test cases as well.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -83,6 +83,51 @@ def process_response(self, request, response, spider):
...
         return response
 
+    def _handle_encoding(self, body, content_encoding):
+        content_to_decode = body
+
+        separator = b","
+        decoded_body = ""
+        for index, encodings in reversed(list(enumerate(content_encoding))):
+            if separator in encodings:
+                decoded_body = ""
```
Gallaecio, line 93: I think this line is unnecessary, and would break scenarios with multiple headers where some of those headers include comma-separated encodings.

Gallaecio (commented):
Besides my comment below, I get the feeling that, while you got something that works, and I am OK merging it as is (once the remaining comment is addressed), I see room for improvement on readability.

For example, I get the feeling that the code would work the same if you removed the `if separator in encodings:` statement and the whole `else` section.

I also think that it may be possible to implement a method that returns a tuple with 2 elements, the list of known encodings to decode, and the list of unknown (and known) encoding not to decode. And then the `_handle_encoding` implementation could become even simpler, something like:

```python
to_decode, to_keep = self._split_encodings(content_encoding)
for encoding in to_decode:
    body = self._decode(body, encoding)
content_encoding[:] = to_keep
return body
```

Also, maybe `_handle_encoding` should return both the new `Content-Encoding` (i.e. `return body, to_keep`) instead of modifying `content_encoding` in place, but that’s not too important.

vishesh10:
> Also, maybe `_handle_encoding` should return both the new `Content-Encoding` (i.e. `return body, to_keep`) instead of modifying `content_encoding` in place, but that’s not too important.

Since the content_encoding is getting used [here](https://github.com/scrapy/scrapy/blob/d222143847f9c591419d6f0478e6866ca505739e/scrapy/downloadermiddlewares/httpcompression.py#L81), that is why the original reference is being updated. If new `Content-Encoding` is to be returned from _handle_encoding then [this if condition](https://github.com/scrapy/scrapy/blob/d222143847f9c591419d6f0478e6866ca505739e/scrapy/downloadermiddlewares/httpcompression.py#L81) would have to be modified. 
Please suggest.

vishesh10:
```
_______________ Https2InvalidDNSId.test_download_chunked_content _______________
'NoneType' object is not iterable

During handling of the above exception, another exception occurred:
NOTE: Incompatible Exception Representation, displaying natively:

twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
Selectables:
<TLSMemoryBIOProtocol #0 on 42761>
----------------------------- Captured stderr call -----------------------------
```

Seems some error with the `Ubuntu / tests (pypy3.8, pypy3-pinned)` check itself.

wRAR:
It's one of those random test failures.

Gallaecio (approved):

wRAR (approved):
Thanks!

PR #6064, opened 2023-09-24
Refactor installing signals.

### LINKED ISSUE #6024: Problem with Twisted AttributeError: 'EPollReactor' object has no attribute '_handleSignals'

wRAR:
Fixes #6024, closes #6030.

This now needs a test described in https://github.com/scrapy/scrapy/pull/6030#issuecomment-1710403951

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -416,15 +416,17 @@ class CrawlerProcess(CrawlerRunner):
                 return
             d.addBoth(self._stop_reactor)
 
-        if install_signal_handlers:
-            install_shutdown_handlers(self._signal_shutdown)
         resolver_class = load_object(self.settings["DNS_RESOLVER"])
         resolver = create_instance(resolver_class, self.settings, self, reactor=reactor)
         resolver.install_on_reactor()
         tp = reactor.getThreadPool()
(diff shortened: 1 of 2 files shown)


wRAR:
This also needs a test for `scrapy shell` without `-c`, because it stopped working but we only see it in random test failures.

scrapy/crawler.py
```diff
@@ -404,8 +404,8 @@ def start(
         :param bool stop_after_crawl: stop or not the reactor when all
             crawlers have finished
 
-        :param bool install_signal_handlers: whether to install the shutdown
-            handlers (default: True)
+        :param bool install_signal_handlers: whether to install the OS signal
+            handlers from Twisted and Scrapy (default: True)
```
kmike, line 408: 👍

Gallaecio (approved):

PR #6072, opened 2023-09-27
added: fuzz_crawl.py and cifuzz

lbl4:
This PR is adding fuzz target from already merged PR in OSS-fuzz (https://github.com/google/oss-fuzz/pull/10991) and CIfuzz for fuzzing targets against every PR.

### DIFF
--- a/extras/fuzz_crawl.py
+++ b/extras/fuzz_crawl.py
@@ -30,11 +30,11 @@ def TestOneInput(data):
     test = fdp.ConsumeUnicodeNoSurrogates(fdp.ConsumeIntInRange(0, 4096))
 
     try:
-	process = CrawlerProcess(settings={test})
-	process.crawl(test_spider)
-	process.start()
+        process = CrawlerProcess(settings={test})
+        process.crawl(test_spider)
+        process.start()
     except:
-	pass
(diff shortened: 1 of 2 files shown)


extras/fuzz_crawl.py
```diff
@@ -0,0 +1,49 @@
...
+def TestOneInput(data):
+    fdp = atheris.FuzzedDataProvider(data)
+    test = fdp.ConsumeUnicodeNoSurrogates(fdp.ConsumeIntInRange(0, 4096))
+
+    try:
+        process = CrawlerProcess(settings={test})
+        process.crawl(test_spider)
+        process.start()
+    except CloseSpider:
+        pass
```
Gallaecio, line 40: I feel a bit out of my depth here. Not only I am not entirely sure how fuzzing works here, but I also don’t what will prevent those `start_urls` from actually being called, which is probably undesirable, or how passing the set `{test}` as the value of a parameter that expects a dict can work.

cc @wRAR

lbl4: Yes, I guess it should be a dict. Sorry about that.
I will close this PR and add better fuzz targets soon.

PR #6073, opened 2023-09-28
added: fuzz_headers.py and cifuzz

lbl4:
Adding fuzz_headers.py for fuzzing request headers and cifuzz for continuous integration with oss-fuzz.

### DIFF
--- a/extras/fuzz_headers.py
+++ b/extras/fuzz_headers.py
@@ -1,0 +1,59 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


extras/fuzz_headers.py
```diff
@@ -0,0 +1,59 @@
...
+
+    class test_spider(scrapy.Spider):
+        name = "test_spider"
+
+        url = "http://quotes.toscrape.com/page/1/"
+
+        def start_requests(self):
+            headers = {"User-Agent": test, "Accept-Language": test, "Referer": test}
+
+            scrapy.Request(self.url, headers=headers, callback=self.parse)
```
wRAR, line 39:
```suggestion
            yield scrapy.Request(self.url, headers=headers, callback=self.parse)
```

extras/fuzz_headers.py
```diff
@@ -0,0 +1,58 @@
...
+
+        def parse(self, response):
+            pass
+
+    try:
+        process = CrawlerProcess()
+        process.crawl(test_spider)
+        process.start()
+    except Exception as e:
+        print(e)
```
Gallaecio, line 48: Don’t you get a [reactor not restartable issue](https://stackoverflow.com/questions/41495052/scrapy-reactor-not-restartable) when running this code multiple times on the same process?

wRAR: They told me it indeed does.

PR #6074, opened 2023-09-28
Add more common MS Office file extensions to exclude for documents

mikkeschiren:
MS Office Word template file has dotx file ending, I think it could be a good idea to exclude it.


Gallaecio (approved):

wRAR:
Thanks!

wRAR (approved):

PR #6076, opened 2023-09-30
Changes for Improve the docs about Crawler initialization changes

### LINKED ISSUE #6057: Improve the docs about Crawler initialization changes
We identified some things we didn't document in the 2.11 backwards incompatibility release notes entry, namely that the crawler at the time of `Spider.from_crawler()` doesn't have most of its [components](https://docs.scrapy.org/en/latest/topics/api.html#scrapy.crawler.Crawler) initialized and that the `engine_started` signal handler may be a better place than `start_requests` to get the fully initialized Crawler. We should document these points in the 2.11 release notes, in the `Spider.from_crawler()` docs and in the `engine_started` signal handler docs and copy this to the 2.11 branch so that the docs are updated.

Amik-Sen-Fun:
Fixes #6057

### DIFF
--- a/docs/topics/signals.rst
+++ b/docs/topics/signals.rst
@@ -118,7 +118,8 @@ engine_started
 
 .. note:: This signal may be fired *after* the :signal:`spider_opened` signal,
     depending on how the spider was started. So **don't** rely on this signal
-    getting fired before :signal:`spider_opened`.
+    getting fired before :signal:`spider_opened`. This may be a better place 
+    than :meth:`start_requests` to get a fully initialized crawler.
 
 engine_stopped
 ~~~~~~~~~~~~~~
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -143,7 +143,11 @@ scrapy.Spider
            arguments. As a consequence, these settings aren't the final values
            as they can be modified later by e.g. :ref:`add-ons
            <topics-addons>`. The final settings are available in the
-           :meth:`start_requests` method and later.
+           :meth:`start_requests` method and later. However, crawler created at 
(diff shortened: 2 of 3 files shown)


Amik-Sen-Fun:
Commits for Improve the docs about Crawler initialization changes #6057

wRAR:
Please rebase your changes on top of master instead of 2.11.

docs/topics/signals.rst
```diff
@@ -118,7 +118,8 @@ engine_started
 
 .. note:: This signal may be fired *after* the :signal:`spider_opened` signal,
     depending on how the spider was started. So **don't** rely on this signal
-    getting fired before :signal:`spider_opened`.
+    getting fired before :signal:`spider_opened`. This may be a better place 
+    than :meth:`start_requests` to get a fully initialized crawler.
```
wRAR, line 122: This statement is unrelated to this note so it should be separate. Also you should refer to `start_requests` using a fully qualified path.

docs/topics/spiders.rst
```diff
@@ -143,7 +143,11 @@ scrapy.Spider
            arguments. As a consequence, these settings aren't the final values
            as they can be modified later by e.g. :ref:`add-ons
            <topics-addons>`. The final settings are available in the
-           :meth:`start_requests` method and later.
+           :meth:`start_requests` method and later. However, crawler created at 
+           the time of :meth:`from_crawler` doesn't have most of its components 
+           initialized, hence :signal:`engine_started` signal handler may be a 
+           better place than :meth:`start_requests` to get a fully initialized 
+           crawler.
```
wRAR, line 150: I don't think this makes sense? The choice between engine_started and start_requests is not related to the crawler state inside from_crawler.

PR #6077, opened 2023-10-01
Fixed the dont_merge_cookies example in docs

### LINKED ISSUE #5936: Docs involving Request.cookies are inaccurate
### Description

The docs for a scrapy.Request have an example that no longer work. See docs [here](https://docs.scrapy.org/en/latest/topics/request-response.html).

In the section where it says: 

> Example of a request that sends manually-defined cookies and ignores cookie storage:
> Request(
>     url="http://www.example.com",
>     cookies={"currency": "USD", "country": "UY"},
>     meta={"dont_merge_cookies": True},
> )
> 

If you submit a request with  `meta={"dont_merge_cookies": True}` then the cookies parameter of a Request object is completely ignored. The manual cookies of `cookies={"currency": "USD", "country": "UY"}` are not included in the request.

We would expect to see `'cookie': 'currency=USD; country=UY'` in the request headers.

The cookies are completely dropped.

Here is the output of `request.headers.to_unicode_dict()`:
`{'accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
 'accept-encoding': 'gzip, deflate',
 'accept-language': 'en',
 'user-agent': 'Scrapy/2.8.0 (+https://scrapy.org)'}
`

### Versions

Scrapy       : 2.8.0
lxml         : 4.9.2.0
libxml2      : 2.9.13
cssselect    : 1.2.0
parsel       : 1.7.0
w3lib        : 2.1.1
Twisted      : 22.10.0
Python       : 3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:26:40) [Clang 14.0.6 ]
pyOpenSSL    : 23.0.0 (OpenSSL 3.0.8 7 Feb 2023)
cryptography : 39.0.1
Platform     : macOS-13.2.1-arm64-arm-64bit

### Additional context

I am patching this by overriding the CookiesMiddleware.process_request and using a temporary cookiejar.

```
        if request.meta.get("dont_merge_cookies", False):
            jar = CookieJar() # Temporary cookiejar
        else:
            cookiejarkey = request.meta.get("cookiejar")
            jar = self.jars[cookiejarkey]
        cookies = self._get_request_cookies(jar, request)
        self._process_cookies(cookies, jar=jar, request=request)

        # set Cookie header
        request.headers.pop("Cookie", None)
        jar.add_cookie_header(request)
        self._debug_cookie(request, spider)


```

11-aryan:
- Removed the whole part about sending custom cookies with `dont_merge_cookies` from the example
- Added warning that the custom cookies are ignored when `dont_merge_cookies` is set to true

Resolves #5936.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -115,15 +115,6 @@ Request objects
         cookies for that domain and will be sent again in future requests.
         That's the typical behaviour of any regular web browser.
 
-        Example of a request that sends manually-defined cookies:
-
-        .. code-block:: python
-
-            Request(
-                url="http://www.example.com",
-                cookies={"currency": "USD", "country": "UY"},
-            )
-
         It is recommended not to set the ``dont_merge_cookies`` key to ``True``
         in :attr:`request.meta <scrapy.Request.meta>` as it causes custom cookies to be ignored.
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -115,8 +115,9 @@ Request objects
         cookies for that domain and will be sent again in future requests.
         That's the typical behaviour of any regular web browser.
 
-        It is recommended not to set the ``dont_merge_cookies`` key to ``True``
-        in :attr:`request.meta <scrapy.Request.meta>` as it causes custom cookies to be ignored.
+        Note that setting the :reqmeta:`dont_merge_cookies` key to ``True`` in
+        :attr:`request.meta <scrapy.Request.meta>` causes custom cookies to be
+        ignored.
 
         For more info see :ref:`cookies-mw`.
 
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -115,20 +115,17 @@ Request objects
         cookies for that domain and will be sent again in future requests.
         That's the typical behaviour of any regular web browser.
 
-        To create a request that does not send stored cookies and does not
-        store received cookies, set the ``dont_merge_cookies`` key to ``True``
-        in :attr:`request.meta <scrapy.Request.meta>`.
-
-        Example of a request that sends manually-defined cookies and ignores
-        cookie storage:
+        Example of a request that sends manually-defined cookies:
 
         .. code-block:: python
 
             Request(
                 url="http://www.example.com",
                 cookies={"currency": "USD", "country": "UY"},
-                meta={"dont_merge_cookies": True},
             )
+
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -115,21 +115,18 @@ Request objects
...
-        cookie storage:
+        Example of a request that sends manually-defined cookies:
 
         .. code-block:: python
 
             Request(
                 url="http://www.example.com",
                 cookies={"currency": "USD", "country": "UY"},
-                meta={"dont_merge_cookies": True},
             )
```
Gallaecio, line 125: The only purpose of this example here was to feature dont_merge_cookies, since there is already an example of cookie use higher up. So I think you can remote this example (and the introductory line) altogether.

11-aryan: Sure, I'll remove it.

docs/topics/request-response.rst
```diff
@@ -115,20 +115,8 @@ Request objects
...
-
-        .. code-block:: python
-
-            Request(
-                url="http://www.example.com",
-                cookies={"currency": "USD", "country": "UY"},
-                meta={"dont_merge_cookies": True},
-            )
+        It is recommended not to set the ``dont_merge_cookies`` key to ``True``
+        in :attr:`request.meta <scrapy.Request.meta>` as it causes custom cookies to be ignored.
```
Gallaecio, line 119: I would change the wording, as I think it could be understood as a general recommendation rather than something that only applies when custom cookies are set (even though it can be inferred from context), and it would be good to link `dont_merge_cookies` to its docs (even though the previous text did not).

What about this?

```suggestion
        Note that setting the :reqmeta:`dont_merge_cookies` key to ``True`` in
        :attr:`request.meta <scrapy.Request.meta>` causes custom cookies to be
        ignored.
```

11-aryan: Yeah I'll change it and also add the link

11-aryan: @Gallaecio is this the right [link](https://doc.scrapy.org/en/latest/topics/downloader-middleware.html#cookies-enabled) for the `dont_merge_cookies` docs

Gallaecio: `:reqmeta:`dont_merge_cookies` will generate a link, it’s a special syntax we have in Scrapy docs for request metadata keys.

Gallaecio (approved):

wRAR (approved):

wRAR:
Thanks!

PR #6078, opened 2023-10-01
Use concurrency groups to limit duplicate jobs

### LINKED ISSUE #5997: Use `cancel-in-progress` on workflows
This way we don't end up with a lot of workflows running in parallel for the same branch when in the end what matters is the most recent one, especially for PR's (also this will save some time as well/reduce the queues).

https://docs.github.com/en/enterprise-cloud@latest/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run

AndrewADev:
Leverage concurrency groups along with cancel-in-progress to favor running the most recent job.

Closes #5997 

Concurrency groups are on a per workflow, per branch/tag basis. So, pushing newer updates to a branch, e.g. as part of a PR, should cancel any in progress runs of workflows that have been retriggered.

Note that I've added these to all workflows, since I couldn't think of a case where it wouldn't make sense to favor the more recently added jobs. Let me know if I've missed something, or that should be done differently.

Note as well that the documentation lists a caveat about order not being guaranteed. In practice I've never had issues with the ordering, though that was for smaller repos with less chance of parallel collaboration.

See:
- https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
- https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/contexts#github-context

## Validation

Separate from the workflows running for the main Scrapy repo (which will need to be approved), I validated: 
- Syntax, via the fact that the workflows run as normal in my fork. 
- The cancellation behavior, via a test branch based off of the branch with these changes (also available on my [fork as a PR](https://github.com/AndrewADev/scrapy/pull/2)) Note that the test changes were simple comment changes in violation of the linting rules, which caused a linting failure. They were just to test the cancellation logic. They are not included here. 

Cancellation in action: 
image
[Run link](https://github.com/AndrewADev/scrapy/actions/runs/6369866027?pr=2) (will eventually expire).

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,6 +1,10 @@
 name: Checks
 on: [push, pull_request]
 
+concurrency:
+  group: ${{github.workflow}}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   checks:
     runs-on: ubuntu-latest
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,10 @@ on:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
 
+concurrency:
+  group: ${{github.workflow}}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   publish:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -1,6 +1,10 @@
 name: macOS
 on: [push, pull_request]
 
+concurrency:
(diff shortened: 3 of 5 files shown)


Gallaecio (approved):
Looks good to me. I did check [github.ref in the docs](https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/contexts#github-context), to be sure it was PR-unique (we don’t want 1 PR to cancel CI for another), and that seems to be the case.

Thanks!

wRAR (approved):

PR #6062, opened 2023-09-23
Explicitly mention PythonItemExporter output changes in 2.11 (#6060)

AlexMuresan:
resolves #6060

PR #6081, opened 2023-10-01
mention PythonItemExporter changes

### LINKED ISSUE #6060: Explicitly mention PythonItemExporter output changes in 2.11
We removed the binary mode of `PythonItemExporter` in 2.11 but it was the default one, so we should have mentioned a backwards incompatible change which is "`PythonItemExporter` used the binary output by defauilt but it no longer does" in the "Backward-incompatible changes" entry for 2.11.

kokobhara:
Fixes #6060, closes #6062


wRAR (approved):

Gallaecio (approved):
Thanks!

PR #6083, opened 2023-10-02
Adapt to the Python 3.12 final release

wRAR:


.github/workflows/tests-windows.yml
```diff
@@ -21,13 +21,13 @@ jobs:
         - python-version: "3.10"
           env:
             TOXENV: py
-        - python-version: "3.10"
```
wRAR, line 24: Not sure why we had asyncio for 3.10 and 3.11, if this was on purpose please chime in.

wRAR:
I guess we need to wait until 3.12 is on the CI servers.

.github/workflows/tests-windows.yml
Gallaecio: I had a look, and I believe it was by accident, https://github.com/scrapy/scrapy/commit/b6541830849a6edd41da6b539d272790a4661d7d#diff-ab5a8d3ff84c9430ab7fb358981634187b850ef3a5369ee94c332d59fd79dd93 followed by https://github.com/scrapy/scrapy/commit/8a0a9e6d3e5398b884e92cb9336efe3fd19677d6.

Gallaecio:
The Windows issue may be legit, related to changes in URL parsing in 3.12 that are more strict. I think even for Windows paths, `file://` URLs need to use `/` as path separator, which may be causing the issue.

wRAR:
Looks like the immediate cause is the presence of a colon in the path. Also, there is w3lib in the traceback, we should check its tests too.

wRAR:
Though we don't run w3lib tests on Windows.

tests/test_downloader_handlers.py
```diff
@@ -127,7 +127,7 @@ def _test(response):
         return self.download_request(request, Spider("foo")).addCallback(_test)
 
     def test_non_existent(self):
-        request = Request(f"file://{self.mktemp()}")
+        request = Request(path_to_file_uri(self.mktemp()))
```
wRAR, line 130: The previous test is already doing this.

Gallaecio (approved):
:rocket:

PR #6087, opened 2023-10-05
Change supported PyPy versions to 3.9 and 3.10.

wRAR:


.github/workflows/tests-ubuntu.yml
```diff
@@ -38,7 +41,7 @@ jobs:
         - python-version: 3.8.17
           env:
             TOXENV: asyncio-pinned
-        - python-version: pypy3.8
+        - python-version: pypy3.9
```
wRAR, line 44: Not sure about this: this fails because the pinned cryptography is probably too old for 3.9, but pypy3.8 is no longer supported. OTOH in the docs we say that we require Python 3.8+, either CPython or PyPy.

Gallaecio: I would keep it in pypy3.8 is the CI allows it.

Otherwise, we can upgrade the minimum crypto, or require a higher one for 3.9+, and update the docs accordingly.

Issue #3163, opened 2018-03-10
Spider logging not optimal


thernstig:
Currently the output of a spider log looks like this:

```python
>>> spider.logger.warning("test")
>>> 2018-03-10 13:42:56 [spider_name_goes_here] WARNING: test
```
The problem with this that the spider's logger has no parent. This is problematic if you have a lot of spiders. It requires you to have a root logger in order to propagate the root loggers configuration down to all spider loggers. And setting a root logger is not always recommended in all projects.

So if you have a dictConfig (e.g. via a json or yaml file) to set your config for your entire python project, you are now forced to enable the root logger OR set logging for each spider in the file.

The solution to this would be that all spider logger's name are prepended automatically with the name of the scrapy project. That way you can easily set project wide logging handling that propagates to all spiders. Maybe even that they are prepended with project_name.spiders since that is their logical structure. So the name would end up as:

```python
>>> spider.logger.warning("test")
>>> 2018-03-10 13:42:56 [project_name.spiders.spider_name_goes_here] WARNING: test
```

alexeyqu:
Hi @thernstig,
What if the output would look like this:
```python
>>> spider.logger.warning("test")
>>> 2018-03-10 13:42:56 [project_name.spiders.spider_module_name_goes_here] WARNING: test
```
There is an elegant way to reach that using ```self.__class__.__module__``` in the spider logger setup instead of ```self.name``` of the spider. 
In that way, logger will be parametrized with ```__name__``` of the module where the most-derived-class of the specific spider is stored. Which is  ```project_name/spiders/test_spider.py```.
I guess, that would solve the problem with project-wide logging.

Drawback: if the some base Scrapy class (like ```Spider``` or ```CrawlSpider```) is used "as is", without deriving from it, then the output won't contain ```project_name``` since ```__module__``` will be taken from Scrapy internals (like ```scrapy.spiders.crawl```). 
AFAIK, this is not a good way of using Scrapy, I saw it only in tests (```SpiderTest``` class). 
Thus, some tests would need to be fixed.

Disclaimer: this would be my first PR, so I'd appreciate some feedback on the idea.

thernstig:
Hi @alexeyqu and thank you for the reply!

That is definitely one way to solve it. I solved it locally now by using just `__module__` as the name for the spider.

The downside of using I suppose would be that `self.__class__.__module__` would not respect the name of the spider set with the name class variable. That name is also what is seen when doing e.g. `scrapy list`.

[...]

PR #6091, opened 2023-10-09
Set spider name in the stats logs.

roykyle8:
When the multiple crawlers are run through CrawlerProcess, logstats are not clear as which spider these stats are for. Adding a spider name in the log.

### DIFF
--- a/scrapy/extensions/logstats.py
+++ b/scrapy/extensions/logstats.py
@@ -42,10 +42,11 @@ class LogStats:
         self.pagesprev, self.itemsprev = pages, items
 
         msg = (
-            "Crawled %(pages)d pages (at %(pagerate)d pages/min), "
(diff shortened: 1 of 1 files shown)


wRAR:
I think fixing #3163 is better than adding the spider name to the message itself (and it's not specific to logstats).

wRAR:
Hmm or is this a different problem? Do you only have problems with the logstats message when you run several spiders?

roykyle8:
@wRAR I'm not sure how it will be implemented and will it work in case of stats as stats loggers separate classes.

kmike:
It seems it'd be nice to include the spider name to all log messages, optionally. Probably, it's a matter of allowing to include spider_name variable in the LOG_FORMAT option (https://docs.scrapy.org/en/latest/topics/settings.html#log-format), when the spider name is available.

PR #6093, opened 2023-10-12
feat: add line buffer try2 new repo

Bannore-Aaditya:
From previous mistake, I have installed "pre-commit" and run "pre-commit install". 
the only problem i am not able to run "tox -- tests/test_dupefilters.py" for testing as it is giving me error 
"RuntimeError: seed failed due to failing to download wheels wheel, setuptools, pip"

#6019


wRAR:
> From previous mistake, I have installed "pre-commit" and run "pre-commit install".

It doesn't look like you've made this commit after installing pre-commit though.

> "RuntimeError: seed failed due to failing to download wheels wheel, setuptools, pip"

Well, you should fix this.

PR #6112, opened 2023-10-17
Make shutdown tests more robust.

wRAR:


wRAR:
The interactive shell test still seems to fail locally from time to time with a timeout waiting for "Available Scrapy objects" (looks like the scraper initialization just stops at some point) but I wasn't able to diagnose or reliably reproduce it so we'll need to see how often it happens in the wild.

Issue #2546, opened 2017-02-08
DOC Add (short) explanation of return vs. yield in callbacks


redapple:
Follow up to https://github.com/scrapy/scrapy/pull/2533

Quite a few Scrapy users are also Python-newbies (and that's great!), and using `yield` in callbacks (and generators in general) can be something odd at first.

It would be great to explain different callback patterns:
- for single object "returns"
- for multi-item or multi-request "returns"
- for mixing items and requests

(and why/when `return` with a list is not necessarily a bad thing)

satwikkansal:
Hi @redapple I'd like to work on this issue. I'm new to scrapy so I don't know the workflow i.e. whether we need to be assigned first or we can submit the PR straightaway :)

redapple:
hi @satwikkansal , no we do not usually assign issue to people.
It's fine to comment here that you are working on the issue (and if you can add notes along the way when you think you can submit your PR, or if you get stuck, if it's taking longer than expected, so that someone can take over if needed)

neonwuchang:
hi it seems that this issue is still open? I'd like to work on this. (Please let me know if I should not do so.)

Gallaecio:
Still open, and https://github.com/scrapy/scrapy/issues/2546#issuecomment-279014919 still applies to anyone interested in working on it.

AdrianAtZyte:
Fixed in #7821.

PR #6114, opened 2023-10-17
Create yield.rst:: Addresses issue #2546 "DOC Add explanation of retu…

neonwuchang:
…rn vs. yield in callbacks"

Addresses issue #2546 "DOC Add (short) explanation of return vs. yield in callbacks" 

This commit adds a file 'yield.rst' which contains a basic explanation of 'yield' keyword in Python, how it works and is different from 'return', and when to use it, with some simple sample code, to docs/topics.


wRAR:
I'm afraid none of this talks about Scrapy callbacks.

neonwuchang:
Looks like I misunderstood that, my bad.

neonwuchang:
I'll make the changes

neonwuchang:
@wRAR could you please take a look to see if this in more in line with what's needed?

Thanks

docs/topics/yield.rst
```diff
@@ -0,0 +1,176 @@
+.. _topics-yield:
+
+==========================
+'Yield' Keyword in Python and its use for Scrapy Callbacks
+==========================
+  
+What is ``yield``?
```
wRAR, line 7: I don't think we need such a long explanation what are Python generator and how does the yield keyword in Python work, there are better existing places for that.

docs/topics/yield.rst
```diff
@@ -0,0 +1,176 @@
...
+    return 5
+  res = gen()
+  for x in res:
+    print(x)
+
+.. _topics-yield-scrapy:
+
+Where is ``yield`` used in Scrapy?
+==================================
+``Yield`` is used for callbacks in Scrapy, which help scrape multiple pages. For example, consider the following code
```
wRAR, line 121: I don't think the "which help scrape multiple pages" part makes sense here?

wRAR:
I'm also not sure if this should be a separate page, as these two paragraphs are better placed next to some related docs.

neonwuchang:
I absolutely see and agree with the point that there are better places for detailed explanations of generators and yield keyword outside the Scapy docs. (I was being extra thorough given that this was meant to help beginners). 

In this case, having a separate page is certainly not needed.

neonwuchang:
Moving forward with the assumption that the explanations are removed, would it be right to say that the content under the last two paragraphs ("Why use yield?" and "Would return work?"), is the kind of content that addresses the issue, or is that off the mark? It would be helpful to have some direction

Thanks!

neonwuchang:
I'll close this PR as I won't be able to get to this for a while because of irl tasks and I don't want to leave this hanging. And if the issue is still open, I can work on it at a later date.

PR #6116, opened 2023-10-18
removed some deprecated functions and supports

### LINKED ISSUE #6109: Remove deprecated support from scrapy.utils.log.logformatter_adapter()
This helper was added in 1.0.0 (I think), it looks like the parts of it that don't show warnings are still useful, though I haven't looked closely at them.

### LINKED ISSUE #6111: Remove deprecated scrapy.utils.response.response_httprepr
Deprecated in 2.6.0.

Chenwei-Niu:
commits aabc3db and cc2ad92 resolved #6111
resolves #6109


wRAR:
Please use pre-commit locally and run tests locally, and fix the errors produced by them.

scrapy/utils/log.py
```diff
@@ -235,15 +235,8 @@ def logformatter_adapter(logkws: dict) -> Tuple[int, str, dict]:
     if not {"level", "msg", "args"} <= set(logkws):
         warnings.warn("Missing keys in LogFormatter method", ScrapyDeprecationWarning)
```
wRAR, line 236: This also needs to be removed.

Chenwei-Niu: Thanks for guiding. I've deleted these in the latest commit.

Chenwei-Niu:
7afb4ddad370a3187da5297f977b2e7
image
I think my commits are okay to rerun the workflow. @wRAR

scrapy/utils/log.py
```diff
@@ -232,18 +230,9 @@ def logformatter_adapter(logkws: dict) -> Tuple[int, str, dict]:
...
-    if "format" in logkws:
-        warnings.warn(
-            "`format` key in LogFormatter methods has been "
-            "deprecated, use `msg` instead",
-            ScrapyDeprecationWarning,
-        )
 
     level = logkws.get("level", logging.INFO)
-    message = logkws.get("format", logkws.get("msg"))
+    message = str(logkws.get("msg"))
```
wRAR, line 235: Why str()? It wasn't needed before.

Chenwei-Niu: Hi, wRAR,
The problem here is that `logkws.get("msg")` returns `Any | None` instead of `Any`. The previous code works well since it sets a default value in get() and returns `Any`.
If we don't take measures here, the type of variable message is `Any | None`. And an incompatible return value error arises when conducting tests.
c0a2e8ee4fb9c2b8c01599fad954a21
f105114ba8921466f0a84ab5fa76967



I came up with three possible solutions:

1. ```message = logkws.get("msg", logkws.get("msg"))```
2. ```message = logkws.get("msg", "")```
3. ```message = str(logkws.get("msg"))```

Do you have any suggestions or preferences among these three solutions?

Gallaecio: The usual approach here, when you can be certain that the value is a string, is to use an assert afterwards, i.e. `assert message is not None`, which should silence mypy.

Chenwei-Niu: I‘ve already tested, the code below could handle both two scenarios you mentioned.
1. key `msg` is missing in the dictionary
2. Dictionary contains `"msg": None`
```python
    message = logkws.get("msg", "")
    assert message is not None
```
If there is no problem, I will modify and commit it.
@Gallaecio

scrapy/utils/log.py
```diff
@@ -232,18 +230,10 @@ def logformatter_adapter(logkws: dict) -> Tuple[int, str, dict]:
...
-        warnings.warn(
-            "`format` key in LogFormatter methods has been "
-            "deprecated, use `msg` instead",
-            ScrapyDeprecationWarning,
-        )
 
     level = logkws.get("level", logging.INFO)
-    message = logkws.get("format", logkws.get("msg"))
+    message = logkws.get("msg", "")
+    assert message is not None
```
Gallaecio, line 236: OK, now that I look at the warning, I see 2 things:
- The warning was about `msg` being a required key, so I think `logkws["msg"]` instead of `logkws.get("msg", "")` makes sense now that the warning is gone.
- There was no warning if `logkws["msg"]` was `None`, so we should assume that as a possibility since this dict seems to be user-provided. So we cannot `assert`, we have to fall back: `message = logkws["msg"] or ""`.

Chenwei-Niu: Hi Gallarcio,
I don't really understand why we have to fall back: `message = logkws["msg"] or ""`.

Firstly ,this part of the warnings has been removed according to   @wRAR 's comments https://github.com/scrapy/scrapy/pull/6116#discussion_r1363487495
``` python
if not {"level", "msg", "args"} <= set(logkws):
        warnings.warn("Missing keys in LogFormatter method", ScrapyDeprecationWarning)
```
----------------
Secondly:
> [Gallaecio](https://github.com/Gallaecio): The warning was about msg being a required key, so I think logkws["msg"] instead of logkws.get("msg", "") makes sense now that the warning is gone.

Although msg is a requried key, but it is only mentioned and constrained in comments in `class LogFormatter`,  there is no actual code constraints. The user can still provide a dictionary without the `msg key`. Hence, if `key msg` is missing, simply using `logkws["msg"]` would arise a KeyError.

And I agree the point we cannot assert `message is not None`. 
In fact, our current logging system could cope with the case `{"msg": None}`. So maybe only `message = logkws.get("msg", "")` should be fine.

Gallaecio: I think raising `KeyError` is fine. I think it is the natural evolution from the warning: first msg was probably not necessary, then it became necessary but it worked without it causing a warning, now it will raise KeyError if missing.

If `"msg": None` is OK, then I agree the fall back is not needed, and we can simply use `logkws["msg"]`.

Laerte: Probably the best solution for this is would be:

```python
message = logkws.get("msg") or ""
```

Laerte: Unless like @Gallaecio said we want to generate a exception so the user can fix their code, then we don’t need to ` `get` at all.

Chenwei-Niu: Hi @wRAR, could you please review and merge this pull request?
Thanks!

Laerte (approved):

Gallaecio:
Thanks everyone!

PR #6117, opened 2023-10-18
Update README.rst. Adding method to install scrapy in a virtual environment.

5onGoku:
using virtual environments can help you avoid conflicts between dependencies in different projects. A python project is always recommended to be used inside a virtual environment. For big projects like scrapy adding a guide to use virtual environments is a must.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,9 +66,51 @@ Install
 
 The quick way:
 
+System-wide installation
+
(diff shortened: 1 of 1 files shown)


wRAR:
It's already covered in https://docs.scrapy.org/en/latest/intro/install.html , linked right after your insertion.

PR #6122, opened 2023-10-26
fix: make CloseSpider exception's traceback display error message

danonya666:
make CloseSpider exception put its reason to traceback instead of ignoring it

### DIFF
--- a/scrapy/exceptions.py
+++ b/scrapy/exceptions.py
@@ -41,7 +41,7 @@ class CloseSpider(Exception):
     """Raise this from callbacks to request the spider to be closed"""
 
     def __init__(self, reason: str = "cancelled"):
-        super().__init__()
+        super().__init__(reason)
         self.reason = reason
 
 


danonya666:
fixes https://github.com/scrapy/scrapy/issues/3463

bloodforcream (approved):

wRAR:
Can you show what does this change in the output?

wRAR:
What about real use cases? You've shown an unhandled exception message while `CloseSpider` shouldn't cause one.

danonya666:
> What about real use cases? You've shown an unhandled exception message while `CloseSpider` shouldn't cause one.

It seems to me that `CloseSpider` exception is located in a public module and there is nothing in the documentation about prohibition of that type of usage. Because of that my project uses this exception in an unhandled way and it would be nice to have a real message when exception occurs, since exceptions are designed to bring the error message to the developer. It is one of the main purposes of the exceptions.

Since the implementation of this exception clearly has a bug and there is no cons in adding the message to this custom exception class from my pov, I don't see why wouldn't we fix that bug, since many other developers might fall into the trap of using this type of exception without catching it

wRAR:
> there is nothing in the documentation about prohibition of that type of usage. Because of that my project uses this exception in an unhandled way and it would be nice to have a real message when exception occurs

This sounds like you should use a different exception, because it doesn't sound like the Scrapy functionality related to `CloseSpider`, including its `reason` attribute, is being used anyway. I would even say that using `CloseSpider` in such way can be confusing.

> It is one of the main purposes of the exceptions.

The purpose of `CloseSpider` is control flow (yes, I know).

> Since the implementation of this exception clearly has a bug

I can't agree with that.

> many other developers might fall into the trap of using this type of exception without catching it

They shouldn't do that either, especially if they e.g. assume doing that will set the spider finish reason.

fabouru1985:
I'm trying to set the reason with no lucky. https://github.com/scrapy/scrapy/pull/5191

wRAR:
@fabouru1985 please use resources such as ones listed at https://scrapy.org/community/ to get help with problems in your code.

PR #6127, opened 2023-10-31
Full typing for Request and Response

wRAR:


scrapy/http/common.py
```diff
@@ -1,5 +1,8 @@
-def obsolete_setter(setter, attrname):
-    def newsetter(self, value):
+from typing import Any, Callable, NoReturn
+
+
+def obsolete_setter(setter: Callable, attrname: str) -> Callable[[Any, Any], NoReturn]:
```
wRAR, line 4: I think we should remove these btw.

scrapy/http/response/__init__.py
```diff
@@ -41,62 +63,62 @@ class Response(object_ref):
...
                 "is not tied to any request"
             )
 
     @property
-    def meta(self):
+    def meta(self) -> Dict[str, Any]:
         try:
-            return self.request.meta
+            return self.request.meta  # type: ignore[union-attr]
         except AttributeError:
```
wRAR, line 99: I have no idea why is this just not checking `self.request is not None` btw.

scrapy/http/response/__init__.py
```diff
@@ -183,6 +205,8 @@ def follow(
         .. versionadded:: 2.0
            The *flags* parameter.
         """
+        if encoding is None:
```
wRAR, line 208: Allowing None in the signature because `TextResponse` needs it allowed in its method implementations and those call the base class methods (`TextResponse` methods don't call `Response.follow()` but I'm not sure if it can be detected statically).

scrapy/http/response/text.py
```diff
@@ -39,20 +54,14 @@ class TextResponse(Response):
...
-        self._encoding = kwargs.pop("encoding", None)
+        self._encoding: Optional[str] = kwargs.pop("encoding", None)
         self._cached_benc: Optional[str] = None
         self._cached_ubody: Optional[str] = None
         self._cached_selector: Optional[Selector] = None
         super().__init__(*args, **kwargs)
 
-    def _set_url(self, url):
-        if isinstance(url, str):
-            self._url = to_unicode(url, self.encoding)
```
wRAR, line 50: This method became no-op after Python 2 code removal. Initially it encoded py2 `unicode` to py2 `str` using `self.encoding` but now both the input and the internal representiation are str-only.

tests/test_http2_client_protocol.py
```diff
@@ -600,6 +600,7 @@ def test_response_has_correct_certificate_ip_address(self):
         def assert_metadata(response: Response):
             self.assertEqual(response.request, request)
             self.assertIsInstance(response.certificate, Certificate)
+            assert response.certificate  # typing
```
wRAR, line 603: I haven't found how to make mypy look at pytest asserts :)

scrapy/http/common.py
wRAR: Though we need to make sure the docs clearly direct the user to `.replace()`.

scrapy/http/response/__init__.py
wRAR: (Tests pass with this change.)

Gallaecio (approved):

scrapy/http/request/__init__.py
```diff
@@ -174,19 +179,19 @@ def __repr__(self) -> str:
     def copy(self) -> "Request":
         return self.replace()
 
-    def replace(self, *args, **kwargs) -> "Request":
+    def replace(self, *args: Any, **kwargs: Any) -> "Request":
```
kmike, line 182: I think technically the return type is different; the same applies to .copy() - the class is preserved by default, unless "cls" attribute is passed. This could matter for classes like FormRequest or other custom request classes. Not a big deal, but I wonder if having self type as a default return type is better than hardcoding Request? We can also try to support `cls` typing directly, though it's more work.

scrapy/http/request/json_request.py
```diff
@@ -42,7 +44,7 @@ def __init__(self, *args, dumps_kwargs: Optional[dict] = None, **kwargs) -> None
     def dumps_kwargs(self) -> dict:
         return self._dumps_kwargs
 
-    def replace(self, *args, **kwargs) -> Request:
+    def replace(self, *args: Any, **kwargs: Any) -> Request:
```
kmike, line 47: isn't it returning JsonRequest, not Request?

scrapy/http/response/__init__.py
```diff
@@ -110,65 +132,65 @@ def _set_body(self, body):
 
     body = property(_get_body, obsolete_setter(_set_body, "body"))
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<{self.status} {self.url}>"
 
-    def copy(self):
+    def copy(self) -> Response:
```
kmike, line 138: Response typing seems wrong for subclasses

scrapy/http/response/__init__.py
```diff
@@ -110,65 +132,65 @@ def _set_body(self, body):
...
         """Return a copy of this Response"""
         return self.replace()
 
-    def replace(self, *args, **kwargs):
+    def replace(self, *args: Any, **kwargs: Any) -> Response:
```
kmike, line 142: I think this could cause some false positives with the type checking, when using response subclasses.

scrapy/http/request/json_request.py
wRAR: Not if you pass `cls`.

scrapy/http/response/__init__.py
wRAR: Same `cls`-related problem.

scrapy/http/response/__init__.py
wRAR: Do you mean false negatives?

scrapy/http/request/__init__.py
kmike: Hm, was thinking that Request typing is more restrictive than Self typing, for the user code. 

We're telling that the result of .replace (and other similar methods) no longer has all the new methods or attributes of the Request subclass, so mypy is going to show an error if user tries to access them.

kmike: Neither Request nor Self is 100% correct here, but I wonder if using Self is going to cause fewer false positives in the type checks for the users.

wRAR: > Without the cls edge case, Self typing looks correct

Yup.

We could add an overload to all these methods that covers the case when `cls is not None`, though that requires adding a `cls` named arg explicitly, not sure if that will break anything.

kmike: I think adding cls explicitly should be fine.

scrapy/http/response/__init__.py
kmike: False positives in the user code, i.e. mypy can raise an error where it shouldn't. For example, `text_response.replace(status=200).text` can cause a false alarm from mypy.

PR #6128, opened 2023-11-02
Update tutorial.rst: clarify how to run spider

JessA-L:
When I used the tutorial, I was confused about how to run the spider again. I later realized that I never exited the Scrapy shell. 

This edit should provide more guidance to future users of Scrapy and the tutorial.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -493,7 +493,15 @@ in the callback, as you can see below:
                     "tags": quote.css("div.tags a.tag::text").getall(),
                 }
 
-If you run this spider, it will output the extracted data with the log::
+To run this spider, exit the scrapy shell by entering::
+
+    quit()
+
+Then, run::
+
+   scrapy crawl quotes
+
+Now, it should output the extracted data with the log::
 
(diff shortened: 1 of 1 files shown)


wRAR:
I'm not convinced we should expect users to not understand the difference between console commands and Python code :-/

JessA-L:
Hi, thanks for your quick reply. 

I’m proposing this change because I had to troubleshoot it myself. Another fellow student had issues at this section as well. Further, Scrapy is used by others who aren’t super familiar with the command line, like those in the marketing field. The tutorial begins with references to basic python instruction, so it seems like it targets unexperienced users.

wRAR:
Yeah, it makes sense, but in that case I'd want something more general or something written in the beginning of the tutorial that explains these basics, as the proposed "exit the scrapy shell" line would be OK in a differently structured doc (one that expected a single terminal opened in a specific directory and showed exact commands one after another,  with "now open the scrapy shell", "now close it and run scrapy crawl" etc. steps) but looks out of place to me in this somewhat more abstract tutorial.

@Gallaecio what do you think?

Gallaecio (approved):
I think the proposed changes are great. While I agree with @wRAR’s points for general documentation, I think in a tutorial specifically hand-holding is more important than not repeating yourself.

wRAR (approved):

wRAR:
Thanks!

PR #6129, opened 2023-11-02
Full typing for scrapy/downloadermiddlewares.

wRAR:


scrapy/downloadermiddlewares/cookies.py
```diff
@@ -1,45 +1,69 @@
...
+        self, cookies: Iterable[Cookie], *, jar: CookieJar, request: Request
+    ) -> None:
         for cookie in cookies:
             cookie_domain = cookie.domain
             if cookie_domain.startswith("."):
                 cookie_domain = cookie_domain[1:]
 
-            request_domain = urlparse_cached(request).hostname.lower()
+            hostname = urlparse_cached(request).hostname
+            assert hostname is not None
```
Gallaecio, line 65: Is this a given? I was thinking some URL protocols do allow this in valid URLs, and I am not sure we prevent cookies from being used with them:

```python
>>> print(urlparse('file:///home/adrian').hostname)
None
```

scrapy/downloadermiddlewares/redirect.py
```diff
@@ -98,6 +116,7 @@ def process_response(self, request, response, spider):
         if "Location" not in response.headers or response.status not in allowed_status:
             return response
 
+        assert response.headers["Location"]
```
Gallaecio, line 119: Would mypy understand this?

```suggestion
        assert "Location" in response.headers
```

or

```suggestion
        assert isinstance(response.headers["Location"], bytes)
```

Otherwise, we are saying that a Location header cannot be an empty string, which I imagine it technically can.

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -75,24 +97,27 @@ def robot_parser(self, request, spider):
                 meta={"dont_obey_robotstxt": True},
                 callback=NO_CALLBACK,
             )
+            assert self.crawler.engine
+            assert self.crawler.stats
```
Gallaecio, line 101: I wonder how we could make this easier for users in their own code.

Would mypy understand if we moved this to `__init__` or `from_crawler`?
Is there anything we can do to enable writing a single line that asserts that the crawler has been initialized?

wRAR: Yeah, it would be good if we didn't need that. `from_crawler` is unlikely to help as checkers don't know it was called. Not sure about `__init__`, as checkers could think that the attribute became `None` since then.

scrapy/downloadermiddlewares/cookies.py
wRAR: The code will fail anyway as it calls `.lower()` on it later. But I'm not sure if it's possible to run this code in a non-HTTP context.

Gallaecio: OK. I guess the only change here is that the exception will be different, and if anyone is getting this exception they should report it anyway so that we can do something about it (maybe prevent cookie handling from working on file:// altogether).

scrapy/downloadermiddlewares/robotstxt.py
wRAR: Dping this in `__init__` doesn't help here, I'll create a separate issue for this (but I'm not sure if it's possible).

Gallaecio (approved):

PR #6133, opened 2023-11-03
scrapy.utils.data_path type hint change

cakemd:
Updating type hint for `scrapy.utils.project.data_path`. 

This is a small change which brings a part of solution of the issue [#5739](https://github.com/scrapy/scrapy/issues/5739).

The type hint change was suggested [here](https://github.com/scrapy/scrapy/issues/5739#issuecomment-1789203848).

### DIFF
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -1,9 +1,9 @@
 import os
 import warnings
 from importlib import import_module
+from os import PathLike
 from pathlib import Path
 from typing import Union
-from os import PathLike
 
 from scrapy.exceptions import NotConfigured
 from scrapy.settings import Settings
--- a/scrapy/utils/project.py
+++ b/scrapy/utils/project.py
@@ -2,6 +2,8 @@ import os
 import warnings
 from importlib import import_module
 from pathlib import Path
+from typing import Union
+from os import PathLike
 
 from scrapy.exceptions import NotConfigured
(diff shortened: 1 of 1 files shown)


scrapy/utils/project.py
```diff
@@ -44,7 +46,7 @@ def project_data_dir(project: str = "default") -> str:
     return str(d)
 
 
-def data_path(path: str, createdir: bool = False) -> str:
+def data_path(path: Union[str, PathLike, Path], createdir: bool = False) -> str:
```
wRAR, line 49: `pathlib.Path` is a subclass of `os.PathLike` so it's not needed to be mentioned separately here.

cakemd: @wRAR updated

wRAR: I meant removing `Path`, not removing `PathLike`. `Union[str, PathLike]` is the usual hint for these APIs.

cakemd: Oh, right. Updated.

wRAR:
Can you please also install and use pre-commit as described at https://docs.scrapy.org/en/latest/contributing.html#pre-commit ?

cakemd:
@wRAR sure, done. Thanks.

wRAR (approved):

Gallaecio:
Thanks!

PR #6134, opened 2023-11-03
Remove obsolete setters for body and url on Request and Response.

wRAR:


Gallaecio:
Hmm… Do we want to remove these? I think it is kind of useful to raise an `AttributeError` that explains what the user should actually do to achieve what they wanted.

wRAR:
I was on the fence with this, but then I checked other attributes and e.g. `meta` works in the same way these ones would work after the change: raising `AttributeError: property 'meta' of 'HtmlResponse' object has no setter`.

Alternatively, we can change all such properties to tell the use about `replace()` and no longer call it "obsolete" internally (and hopefully do it in a normal way, not with a magical wrapper as it's now).

Gallaecio:
I had not realized this was only done for some of the properties. I think your logic is sound.

Gallaecio (approved):

PR #6143, opened 2023-11-11
Improve type hints for copy() and replace() in Request and Response.

wRAR:
As discussed in #6127


scrapy/downloadermiddlewares/redirect.py
```diff
@@ -27,6 +27,7 @@ def _build_redirect_request(
     redirect_request = source_request.replace(
         url=url,
         **kwargs,
+        cls=None,
```
wRAR, line 30: Without this line the type of `redirect_request` is inferred to be `Any`.

kmike:
The approach looks good 👍 

It seems the typing becomes more tricky now, especially as we have py.typed in scrapy now - it affects the user code, not just the internal type consistency. What do you think about adding some basic typing tests, at least for the important public interfaces like Response?

wRAR:
What do you think should be tested, at least as a first step?

kmike:
> What do you think should be tested, at least as a first step?

I think we can do it gradually - start with the fixes, as compared to the previous implementation. E.g. that text_response.replace(...).text works properly for different scenarios (raises or doesn't raise mypy errors, as appropriate).

wRAR:
Should the typing tests run on 3.8 or 3.12? We need different syntax for them.

kmike:
> Should the typing tests run on 3.8 or 3.12? We need different syntax for them.

Not sure. Is it because the error messages differ, or is it about something else? Can we rely on mypy error codes instead of exact error messages more?

wRAR:
> Is it because the error messages differ, or is it about something else? 

Yes, mypy suggests `A | B` on newer Python but `Union[A, B]` on older Python.

> Can we rely on mypy error codes instead of exact error messages more?

We can, though we run the test on a single Python version anyway, so we know which message will be produced.

scrapy/downloadermiddlewares/redirect.py
kmike: Could you please add a test for this case? I guess it's `**kwargs` handling?

scrapy/http/request/json_request.py
```diff
@@ -44,7 +50,19 @@ def __init__(
     def dumps_kwargs(self) -> dict:
         return self._dumps_kwargs
 
-    def replace(self, *args: Any, **kwargs: Any) -> Request:
+    @overload
+    def replace(
+        self, *args: Any, cls: Type[RequestTypeVar], **kwargs: Any
+    ) -> RequestTypeVar:
```
kmike, line 56: nice!

.github/workflows/checks.yml
```diff
@@ -18,6 +18,9 @@ jobs:
         - python-version: 3.8
           env:
             TOXENV: typing
+        - python-version: 3.8
```
kmike, line 21: I wonder if we should run it in the latest Python instead.

tox.ini
```diff
@@ -46,6 +46,14 @@ deps =
 commands =
     mypy {posargs: scrapy tests}
 
+[testenv:typing-tests]
```
kmike, line 49: Should we pin the Python version here?

kmike:
@wRAR 

> We can, though we run the test on a single Python version anyway, so we know which message will be produced.

This works. Do you think thought that the more detailed matching of the error messages help us to catch more issues, as compared to just checking the mypy error code?

scrapy/downloadermiddlewares/redirect.py
wRAR: I think it's because it cannot know that `kwargs` passed to `_build_redirect_request` doesn't contain `cls`.

.github/workflows/checks.yml
wRAR: Should we also run the `typing` check on the latest Python then, as I think they check the same thing? In that case we will stop checking (and so supporting) that the type hints work as expected on older Python versions and I thought it's a nice thing to support.

wRAR: (I mean, for users that use Scrapy on older Python versions and run mypy on their code)

wRAR:
> Do you think thought that the more detailed matching of the error messages help us to catch more issues, as compared to just checking the mypy error code?

Yeah, I think error codes can be too broad to be useful.

scrapy/downloadermiddlewares/redirect.py
wRAR: Also, what test do you mean?

wRAR: I've added tests that check that the result is `Any` unless `cls` was also passed.

wRAR: Not sure if it's a useful thing to check :)

Gallaecio (approved):

PR #6145, opened 2023-11-13
httpcompression middleware: warn of unexpected and prevent manual unupported encodings

cakemd:
This change is made for solving issue [#4697](https://github.com/scrapy/scrapy/issues/4697).

PR [#4697](https://github.com/scrapy/scrapy/pull/5480) is used in terms of fast understanding where the code base must be changed. Please, inform me if this PR must be to that branch first and not directly to master of scrapy/scrapy.

Please, not that in initial version of this PR, the body becomes empty bytes in case of b"br" is used in response but not supported by executing machine. Please, inform me if it's not correct approach.

Will be happy to get review.

### DIFF
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -123,7 +123,7 @@ class HttpCompressionMiddleware:
             if b"br" in ACCEPTED_ENCODINGS:
                 body = brotli.decompress(body)
             else:
-                body = ""
+                body = bytes()
                 logger.warning(
                     "Brotli encoding received. "
                     "Cannot decompress the body as Brotli is not installed."
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import io
+import logging
 import zlib
 from typing import TYPE_CHECKING, List, Optional, Union
 
 from scrapy import Request, Spider
 from scrapy.crawler import Crawler
-from scrapy.exceptions import NotConfigured
+from scrapy.exceptions import NotConfigured, NotSupported
 from scrapy.http import Response, TextResponse
 from scrapy.responsetypes import responsetypes
 from scrapy.statscollectors import StatsCollector
@@ -17,6 +18,9 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
 ACCEPTED_ENCODINGS: List[bytes] = [b"gzip", b"deflate"]
+logger = logging.getLogger(__name__)
+
+ACCEPTED_ENCODINGS = [b"gzip", b"deflate"]
 
 try:
     import brotli
@@ -49,6 +53,8 @@ class HttpCompressionMiddleware:
     def process_request(
         self, request: Request, spider: Spider
     ) -> Union[Request, Response, None]:
+        self._raise_unsupported_compressors(request)
+
         request.headers.setdefault("Accept-Encoding", b", ".join(ACCEPTED_ENCODINGS))
         return None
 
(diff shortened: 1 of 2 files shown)


scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -17,6 +18,9 @@
     from typing_extensions import Self
 
 ACCEPTED_ENCODINGS: List[bytes] = [b"gzip", b"deflate"]
+logger = logging.getLogger(__name__)
+
+ACCEPTED_ENCODINGS = [b"gzip", b"deflate"]
```
wRAR, line 23: This looks like a merge error.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -85,6 +91,19 @@ def process_response(
 
         return response
 
+    def _raise_unsupported_compressors(self, request: Request):
+        encodings = request.headers.getlist("Accept-Encoding")
+        unsupported = [key for key in encodings if key not in ACCEPTED_ENCODINGS]
```
wRAR, line 96: You need to split the header value.

scrapy/downloadermiddlewares/httpcompression.py
cakemd: Yup. I'm working on some more changes. Many tests got broken, fixed that already. Polishing next commit now.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -99,8 +118,15 @@ def _decode(self, body: bytes, encoding: bytes) -> bytes:
                 # http://www.port80software.com/200ok/archive/2005/10/31/868.aspx
                 # http://www.gzip.org/zlib/zlib_faq.html#faq38
                 body = zlib.decompress(body, -15)
-        if encoding == b"br" and b"br" in ACCEPTED_ENCODINGS:
-            body = brotli.decompress(body)
+        if encoding == b"br":
+            if b"br" in ACCEPTED_ENCODINGS:
+                body = brotli.decompress(body)
+            else:
+                body = bytes()
```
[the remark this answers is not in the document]
cakemd: Yes, was still thinking what to return. Will change.

wRAR:
So, this raises exceptions for requests with unsupported encodings declared in Accept-Encoding. I'm not sure if this is the best course of action. In the previous issues/PRs there were also such options suggested as logging a warning and removing the unsupported encodings from the header.

In any case, the warning/error logged/raised should clearly explain what is the problem in trhe case of brotli and zstd: that the user should install a module to get it supported.

cakemd:
@wRAR we can change raising an error to logging error message, not a problem. I just followed @Gallaecio 's issue description.

I've changed PR already, so it fires exception only if it's b"br" manually added by user and not supported by machine.
(Initially, I added raising error for any not supported encoding, but I faced failed tests caused by some encodings that are not present in ACCEPTED _ENCODINGS but are in use).

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -85,6 +93,22 @@ def process_response(
 
         return response
 
+    @property
+    def _raise_unsupported(self) -> Tuple[bytes]:
+        return (b"br",)
+
+    def _raise_unsupported_compressors(self, request: Request):
+        encodings = request.headers.getlist("Accept-Encoding")
+        if encodings and len(encodings):
```
wRAR, line 102: I think the `len` check here is not needed.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -85,6 +93,22 @@ def process_response(
...
+            encodings = encodings.pop().split(ENCODINGS_DELIMETER)
```
wRAR, line 103: Not sure why is this doing `encodings.pop()` as you could use `encodings[1]`, but also this should iterate over `encodings` instead of taking the first item.

PR #6148, opened 2023-11-16
Handling CloseSpider exception if it has been raised in start_requests()

### LINKED ISSUE #3463: raise CloseSpider("Error Message") is not saving error message

cakemd:
Fixes #3463

I assume it's not the cleanest approach, so I'm opened for change requests after review.

I was trying to modify the `ExecutionEngine.open_spider` in a way that the `start_requests`'s exception would be called inside it. But it looked that it would take too much changes, so a lot of stuff can be broken.

Still, I'm opened to any ideas of how to make this change better.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -400,10 +400,11 @@ class ExecutionEngine:
             assert isinstance(ex, CloseSpider)  # typing
             self.close_spider(self.spider, reason=ex.reason)
 
-    def close_spider_on_start(self, spider: Spider, reason: str = "cancelled"):
+    def close_spider_before_start(self, spider: Spider, reason: str = "cancelled"):
         if self.slot is not None:
             raise RuntimeError("Engine slot is already assigned. Use self.close_spider")
 
+        self.start()
         nextcall_none = CallLaterOnce(lambda: None)
         scheduler = create_instance(
             self.scheduler_cls, settings=None, crawler=self.crawler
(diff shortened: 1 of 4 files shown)


cakemd:
Still working on it, got failed tests after some changes.

cakemd:
@Gallaecio @wRAR could you please take a look if have time?

scrapy/core/engine.py
```diff
@@ -400,6 +400,18 @@ def _spider_idle(self) -> None:
             assert isinstance(ex, CloseSpider)  # typing
             self.close_spider(self.spider, reason=ex.reason)
 
+    def close_spider_before_start(self, spider: Spider, reason: str = "cancelled"):
+        if self.slot is not None:
+            raise RuntimeError("Engine slot is already assigned. Use self.close_spider")
+
+        self.start()
```
wRAR, line 407: This doesn't look correct because start() returns a Deferred, but also is it OK to call it here at all if we are closing anyway?

cakemd: @wRAR 
I didn't place it at beginning. But to common `close_spider` method has several checks for engine being started. Not sure we need to have use it though, just think that the code inside it is useful. 

I've tried to avoid it, but neither of approaches I've considered allowed to get the proper result. The only other option I see so far is to simply set close reason to spider and leave it in piece.

How do you think to do it better?

wRAR: In any case if you call it you need to wait until it completes.

PR #6151, opened 2023-11-22
new `allow_offsite` parameter in OffsiteMiddleware

### LINKED ISSUE #3690: Separate Attribute to allow offsite requests

BurnzZ:
I'm proposing to have a new `allow_offsite` parameter in `OffsiteMiddleware`. Currently, it relies on `Request.dont_filter` attribute to be set to `True` to allow offsite requests. However, it seems that we cannot rely on this flag directly since setting it to True could result in multiple duplicated requests, as per its definition:

> dont_filter (bool) – indicates that this request should not be filtered by the scheduler. This is used when you want to perform an identical request multiple times, to ignore the duplicates filter. Use it with care, or you will get into crawling loops.

There are cases where we may want to visit an offsite page but also want to filter duplicate requests after.

Fixes https://github.com/scrapy/scrapy/issues/6366, fixes #3690, closes #3691.


wRAR:
I think this will show deprecation warnings for all requests with dont_filter=True, even ones that use it to skip the dupefilter?

docs/topics/spider-middleware.rst
```diff
@@ -345,9 +345,9 @@ OffsiteMiddleware
    :attr:`~scrapy.Spider.allowed_domains` attribute, or the
    attribute is empty, the offsite middleware will allow all requests.
 
-   If the request has the :attr:`~scrapy.Request.dont_filter` attribute
-   set, the offsite middleware will allow the request even if its domain is not
-   listed in allowed domains.
+   If ``allow_offsite`` is set to ``True`` in :attr:`Request.meta`, then the
+   offsite middleware will allow the request even if its domain is not listed
+   in allowed domains.
```
kmike, line 350: If allow_offsite is False, but dont_filter is True, would the middleware allow offsite requests? I think we still should document this.

BurnzZ:
> I think this will show deprecation warnings for all requests with dont_filter=True, even ones that use it to skip the dupefilter?

Yes but only once as https://docs.python.org/3/library/warnings.html states:

> Repetitions of a particular warning for the same source location are typically suppressed.

scrapy/spidermiddlewares/offsite.py
```diff
@@ -50,7 +51,15 @@ async def process_spider_output_async(
...
             return True
-        if request.dont_filter or self.should_follow(request, spider):
+        if request.dont_filter:
+            warnings.warn(
+                "The dont_filter filter flag is deprecated in OffsiteMiddleware. "
+                "Set 'allow_offsite' to True in Request.meta instead.",
+                ScrapyDeprecationWarning,
+                stacklevel=2,
+            )
+            return True
```
kmike, line 61: Hm, I wonder if it'd be better to just keep supporting dont_filter, without a deprecation warning.

First, the flag says "dont_filter", and we respect it, and don't filter out the request. An argument can be made though that "don't filter" only applies to deduplication filter, not to other types of filtering. I think that's valid, Bbut that's not the current flag behavior, and also it's not in the flag name (it's not "dont_deduplicate").

Second, the user doesn't control all the don't_filter flags, Scrapy and other components can be setting this flag. For example, the default start_requests implementation uses dont_filter=True. 

It seems it's not possibile to "deprecate don_filter flag in OffsiteMiddleware", because the user might be setting this flag not for the OffsiteMiddleware, but for other Scrapy components, but the request may still end up in OffsiteMiddleware.

BurnzZ: Makes sense. In that case, we need to clarify that the `dont_filter` value also affects some middlewares as well and not just for the Scheduler.

wRAR (approved):

wRAR:
We have tests for the offsite downloader mw but they aren't a copy of the spider mw ones so I'll add a test for this flag later.

docs/topics/downloader-middleware.rst
```diff
@@ -797,9 +797,10 @@ OffsiteMiddleware
    :attr:`~scrapy.Spider.allowed_domains` attribute, or the
    attribute is empty, the offsite middleware will allow all requests.
 
-   If the request has the :attr:`~scrapy.Request.dont_filter` attribute
-   set, the offsite middleware will allow the request even if its domain is not
-   listed in allowed domains.
+   If the request has the :attr:`~scrapy.Request.dont_filter` attribute set to
+   ``True`` or :attr:`Request.meta` has ``allow_offsite`` set to ``True``, then
```
wRAR, line 801: This probably deserves a `reqmeta` paragraph or at least that directive above this paragraph.

Gallaecio (approved):

BurnzZ:
Completely forgot about this. Cheers for the final push @wRAR @Gallaecio ! 🫡 🚀

PR #6158, opened 2023-11-27
Small improvements

pyvadev:


wRAR:
I don't think most of these are improvements?

wRAR:
And it looks like you haven't even checked that this works.

PR #6162, opened 2023-11-30
Simplify `create_instance` calls

### LINKED ISSUE #5523: Simplify create_instance calls

jxlil:
Added `build_from_crawler` and `build_from_settings` functions

Fixes #5523, closes #5884, closes #6169.

### DIFF
--- a/scrapy/utils/misc.py
+++ b/scrapy/utils/misc.py
@@ -160,6 +160,50 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
     return instance
 
 
+def build_from_crawler(objcls, crawler, /, *args, **kwargs):
+    """Construct a class instance using its ``from_crawler`` constructor, if
+    available.
+
+    The arguments ``*args`` and ``**kwargs`` are forwarded to the constructors.
+
+    Raises ``TypeError`` if the resulting instance is ``None`` (e.g. if an
+    extension has not been implemented correctly.)
+    """
+    if crawler and hasattr(objcls, "from_crawler"):
+        instance = objcls.from_crawler(crawler, *args, **kwargs)
+        method_name = "from_crawler"
+    else:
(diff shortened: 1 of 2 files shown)


scrapy/utils/misc.py
```diff
@@ -160,6 +160,50 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
...
+def build_from_crawler(objcls, crawler, /, *args, **kwargs):
+    """Construct a class instance using its ``from_crawler`` constructor, if
+    available.
+
+    The arguments ``*args`` and ``**kwargs`` are forwarded to the constructors.
+
+    Raises ``TypeError`` if the resulting instance is ``None`` (e.g. if an
+    extension has not been implemented correctly.)
+    """
+    if crawler and hasattr(objcls, "from_crawler"):
```
Gallaecio, line 172:
```suggestion
    if hasattr(objcls, "from_crawler"):
```

I think that only made sense when settings could be specified and crawler not.

scrapy/utils/misc.py
```diff
@@ -160,6 +160,50 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
...
+        instance = objcls.from_crawler(crawler, *args, **kwargs)
+        method_name = "from_crawler"
+    else:
```
Gallaecio, line 175: In this function, if `from_settings` exists, it should be used with `crawler.settings`.

Gallaecio (commented):
It is a good first step, but there is some more work to do, such as:
- Logging a deprecation warning when `create_instance` is called.
- Replacing all usages of `create_instance` in the code base by one of these new functions.

jxlil:
Hi @Gallaecio, thanks for responding, I'm going to close this PR.

PR #6164, opened 2023-11-30
Implement a close reason for robots.txt affecting all start requests

Gallaecio:


wRAR (approved):

scrapy/spidermiddlewares/robotstxt.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(self, crawler):
+        self._send_signal = crawler.signals.send_catch_log
+
+    def process_start_requests(self, start_requests, spider):
+        # Mark start requests and reports to the downloader middleware the
+        # number of them once all have been processed.
+        count = 0
+        for request in start_requests:
+            request.meta["is_start_request"] = True
```
kmike, line 17: should we make it a general Scrapy feature somehow, including the "start_requests_processed" signal? Or maybe even "on_start_request" signal in addition to that, so that we don't need a meta flag.

scrapy/core/engine.py
```diff
@@ -181,14 +190,19 @@ def _next_request(self) -> None:
...
                 self.slot.start_requests = None
+                self.signals.send_catch_log(signal=signals.start_requests_exhausted)
             except Exception:
                 self.slot.start_requests = None
                 logger.error(
                     "Error while obtaining start requests",
                     exc_info=True,
                     extra={"spider": self.spider},
                 )
+                self.signals.send_catch_log(signal=signals.start_requests_exhausted)
```
Gallaecio, line 201: I wonder if we should implement separate method for these, but it felt a bit out of scope.

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -48,23 +85,53 @@ def __init__(self, crawler: Crawler):
...
-        return cls(crawler)
+        crawler.signals.connect(
+            self._start_request_returned, signal=signals.start_request_returned
+        )
+        crawler.signals.connect(
+            self._start_requests_exhausted, signal=signals.start_requests_exhausted
+        )
+
+    def _start_request_returned(self, request):
+        self._total_start_request_count += 1
```
Gallaecio, line 99: I thought about counting them on the engine and passing the count as a parameter through the signal, but it felt weird.

docs/topics/signals.rst
```diff
@@ -466,6 +466,45 @@ headers_received
...
+    method of :ref:`spider middlewares <topics-spider-middleware>`, and before
+    that request reaches the :ref:`scheduler <topics-scheduler>`
+    (:signal:`request_scheduled` signal).
+
+    This signal does not support returning deferreds from its handlers.
+
+    :param request: Returned request.
+    :type request: scrapy.http.Request
+
+start_requests_exhausted
```
Gallaecio, line 490: I went with these names to be more in line with current signal naming, although I am not 100% sure about them. I used “returned” because that’s the wording used in the [iterator docs](https://docs.python.org/3/glossary.html#term-iterator), and with “exhausted” as a work that work both for the iterator stopping and an exception being raised.

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -48,23 +85,53 @@ def __init__(self, crawler: Crawler):
...
+        fingerprint = self._fingerprinter.fingerprint(request)
+        self._pending_start_request_fingerprints.add(fingerprint)
```
Gallaecio, line 101: This is the best I could come up with to avoid changing the request meta. It gets rather verbose in the 2 process_request methods, and weird having to remove and re-add fingerprints in the first one, but it works.

PR #6165, opened 2023-12-01
Simplifying create_instance function

### LINKED ISSUE #5523: Simplify create_instance calls

monicaq21:
fixes #5523 

Issue: the previous create_instance function has an unnecessarily verbose caller, with both "crawler" and "settings" as mandatory parameters where only one of them will be used.

Solution: added default None values for both crawler and settings to shorten the caller. We considered splitting the function into two (build_from_crawler & build_from_settings), but thought it was redundant for the purpose of simplifying the function.


monicaq21:
Hi @wRAR why was this PR closed?

monicaq21:
> @monicaq21 short answer: because it explicitly doesn't do what the issue asks to do.

Got it, thank you. We'll try to make another PR to better meet the requirements :))

PR #6170, opened 2023-12-05
Add async support for start_requests() method #5627

simoncen:
#5627 
My friend and I have added a suggested add-on to the code to provided async support for the start_requests() method.


wRAR:
https://github.com/scrapy/scrapy/issues/5627#issuecomment-1761679207

Issue #6069, opened 2023-09-27
Add JsonResponse


ma-pony:
## Summary

Adding a JsonResponse object

A one paragraph explanation of the function.

## Motivation
When I parse the response content, I want to be able to determine different processing logic based on the type of response, something like
```
if isinstance(response, JsonResponse):
    data = response.jmespath('object.[*]')
    ...
elif isinstance(response, TextResponse):
    data = response.xpath('//p')
    ...
else:
    ...

```

## Additional background
https://github.com/scrapy/scrapy/issues/2444
There has been a similar discussion before

Gallaecio:
At the moment this can be done with what came out of https://github.com/scrapy/scrapy/issues/2444 for most (all?) scenarios. Something like:

```python
if isinstance(response, TextResponse):
    try:
        json = response.json()
    except ValueError:
        data = response.xpath('//p')
    else:
        data = response.jmespath('object.[*]')
else:
    ...
```

ma-pony:
> At the moment this can be done with what came out of #2444 for most (all?) scenarios. Something like:
> 
> ```python
> if isinstance(response, TextResponse):
>     try:
>         response.json()
>     except ValueError:
>         data = response.xpath('//p')
>     else:
>         data = response.jmespath('object.[*]')
> else:
>     ...
> ```
> 
> There might be a cleaner approach calling `response.jmespath()` directly, though, and avoiding the `response.json()` call. Haven’t checked.

Personally, I find this approach more succinct and clearer than the above, can you share your thoughts on this?

```
if isinstance(response, JsonResponse):
    data = response.jmespath('object.[*]')
    ...
elif isinstance(response, TextResponse):
    data = response.xpath('//p')
    ...
else:
    ...

```
I'm willing to add MR to do this

ma-pony:
```
from scrapy.core.downloader.handlers.http11 import ScrapyAgent
from scrapy.responsetypes import ResponseTypes as ScrapyResponseTypes
from scrapy.utils.python import to_unicode


class ResponseTypes(ScrapyResponseTypes):
    CLASSES = {
        "text/html": "scrapy.http.HtmlResponse",
        "application/atom+xml": "scrapy.http.XmlResponse",
        "application/rdf+xml": "scrapy.http.XmlResponse",
        "application/rss+xml": "scrapy.http.XmlResponse",
        "application/xhtml+xml": "scrapy.http.HtmlResponse",
[...]

PR #6171, opened 2023-12-05
Added json.py file to include a JsonResponse class. Also included this JsonResponse module in _init_.py

tl07172:


wRAR:
Is this for #6069? That's not just about adding the class but about using it automatically for application/json responses.

Issue #6136, opened 2023-11-05
Investigate simplifying static checks for initialized Crawler


wRAR:
After we changed the Crawler logic to initialize components such as `Crawler.stats` later, static checkers (specifically mypy) don't know that when code like `self.crawler.stats.inc_value()` is called, `self.crawler.stats` is already initialized, so we need to add asserts in all methods that do this. We should try to find a way to do the same with fewer asserts. This also applies to 3rd-party code (middlewares etc.).

antarahebbar:
Hi! @yueric985 and I are students are we are working on a project for a class that requires us to make an open-source contribution to a library. We will be working on this issue as an example.

sa2415:
Hi! Can I work on this?

derekk10:
Question about the issue. I understand that several components of Crawler are initialized later. However, is it ensured that these components such as self.crawler.stats is always initialized before calling a method such as self.crawler.stats.inc_value()? If so, would doing something like self.stats = cast(StatsCollector, self.stats) be a valid solution? This would tell mypy to treat self.stats as an ainstance of StatsCollector for the rest of the method so we wouldn't have to use an assert statement before accessing self.stats.

wRAR:
> would doing something like self.stats = cast(StatsCollector, self.stats) be a valid solution?

I don't think this code does anything?

Also, where would you put it?

wRAR:
> initializes the actual stats object the first time it is accessed

That's the problem, it shouldn't be initialized on access, it's initialized by a separate method and the question of this ticket is how can we tell the checkers that.

kalessin:
This works for me. On `scrapy.spiders/__init__.py`:

```python
class Spider(object_ref):
   ....
   ....
   stats: StatsCollector

   def _set_crawler(self, crawler: Crawler) -> None:
      ....
      if crawler.stats is not None:
            self.stats = self.crawler.stats
```

Then along the spider, instead of `self.crawler.stats`, access via `self.stats`. In addition, here the `stats: StatsCollector` type hint is strictly correct as it will be assigned after self.crawler.stats is initialized.

mypy accepts it, and works ok on running time.

wRAR:
https://github.com/aio-libs/aiohttp/blob/4399a6c614947bcb8566201c4ecc5133e7dfa4d6/aiohttp/client_reqrep.py#L762-L764

PR #6172, opened 2023-12-05
Fix for issue 6136

derekk10:
Resolve issue #6136. 

Added a method for the crawler class called retrieve_stats() that performs the necessary assertion to check if the stat field is not none and then returns crawler.stats. This removes the need to add an assert statement every time crawler.stats needs to be accessed. Additionally, most instances of crawler.stats has been replaced with crawler.retrieve_stats() so that the assertion is performed automatically.


wRAR:
#6136 is not about stats, it's about a general solution to this that doesn't invove changing all the code. And you didn't even review your changes.

PR #6174, opened 2023-12-07
Added `JsonResponse`

jxlil:
Added a `JsonResponse` subclass to facilitate parsing of response content

```python
if isinstance(response, JsonResponse):
    data = response.jmespath('object.[*]')
    ...
elif isinstance(response, HtmlResponse):
    data = response.xpath('//p')
    ...
else:
    ...
```

Fixes https://github.com/scrapy/scrapy/issues/6069, closes https://github.com/scrapy/scrapy/pull/6171


docs/topics/request-response.rst
```diff
@@ -1328,3 +1328,12 @@ XmlResponse objects
...
 .. _bug in lxml: https://bugs.launchpad.net/lxml/+bug/1665241
+
+JsonResponse objects
+--------------------
+
+.. class:: JsonResponse(url[, ...])
+
+    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
+    adds encoding auto-discovering support by looking into the JSON declaration
+    line.  See :attr:`TextResponse.encoding`.
```
wRAR, line 1339: I don't think this makes sense?

jxlil: Oh you're right, you think this might work:

```rst
    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
    adds encoding auto-discovering support by looking for a MIME type in the
    `Content-Type` header that represent to a JSON object.  See :attr:`TextResponse.encoding`.
```

docs/topics/request-response.rst
```diff
@@ -1328,3 +1328,12 @@ XmlResponse objects
...
 .. _bug in lxml: https://bugs.launchpad.net/lxml/+bug/1665241
+
+JsonResponse objects
+--------------------
+
+.. class:: JsonResponse(url[, ...])
+
+    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` which
+    adds encoding auto-discovering support by looking for a MIME type in the
+    `Content-Type` header that represent to a JSON object.  See :attr:`TextResponse.encoding`.
```
Gallaecio, line 1339: I think @wRAR’s point is that “adds encoding auto-discovering support” seems incorrect or misleading.

Something like this is probably more accurate:

```suggestion
    The :class:`JsonResponse` class is a subclass of :class:`TextResponse` 
    that is used when the response has a `JSON MIME type 
    <https://mimesniff.spec.whatwg.org/#json-mime-type>`_ in its `Content-Type` 
    header.
```

Although I am unsure of whether or not this change, i.e. #6069, is worth implementing.

jxlil: Oh I see. Thanks

scrapy/http/response/json.py
```diff
@@ -0,0 +1,12 @@
+"""
+This module implements the JsonResponse class which adds encoding
+discovering through JSON encoding declarations to the TextResponse class.
```
wRAR, line 3: This also needs a change.

jxlil: Thanks. I just updated the module docs

wRAR (approved):
Thanks!

Gallaecio (approved):

Gallaecio:
Thank you!!!

PR #6176, opened 2023-12-10
Fix for issue 6047

### LINKED ISSUE #6047: Improve unhandled exception handling for Crawler*.crawl()
As we recently found in #6038, the following well-known snippet from docs and tests:

```python
process = CrawlerProcess()
process.crawl(MySpider)
process.start()
```

doesn't handle exceptions raised in crawl() correctly, as that method returns a Deferred and we should add an errback to it instead of relying on https://docs.twisted.org/en/latest/core/howto/defer.html#unhandled-errors, either in our code or in examples.

phyllis-feng:
Added errbacks by rewriting crawl() functoins

Still in progress

To Do:

- fix typing errors
- add more documentation

Fixes #6047


wRAR:
Is this still in progress?

PR #6178, opened 2023-12-13
resolve issue 5932

### LINKED ISSUE #5932: _store_in_thread must not receive a closed file when using postprocessing plugins
When using post-processing plugins, the FeedExporter closes the temporary file. This behavior has been introduced in commit 500aaa258fc72e943d7bbd258900aeb5af6c8b0d.

Some storage backends, such as GCSFeedStorage, expect an open file in the _store_in_thread method.

To solve the problem we should only close the e.g. GzipFile instance, not the temporary file. Removing the line self.file.close() in GzipPlugin, Bz2Plugin and LZMAPlugin should resolve the error. I think this change would be a clean solution because usually the component that created a resource should be responsible to close that resource - so closing the BinaryIO object in the plugins is a bit unexpected.

Furthermore, GCSFeedExporter should be modified to close the file in its _store_in_thread method after the upload. FTPFeedStorage and S3FeedStorage do this as well.

nluginbill:
Fixes #5932 

Removed the lines closing the temporary file in GzipPlugin, Bz2Plugin, and LMZAPlugin in postprocessing.py.

Placed file closes in GCSFeedExporter, FTPFeedStorage, and S3FeedStorage in feedexport.py.

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -269,6 +269,7 @@ class GCSFeedStorage(BlockingFeedStorage):
         bucket = client.get_bucket(self.bucket_name)
         blob = bucket.blob(self.blob_name)
         blob.upload_from_file(file, predefined_acl=self.acl)
+        self.file.close()
 
 
 class FTPFeedStorage(BlockingFeedStorage):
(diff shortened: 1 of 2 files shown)


Gallaecio:
It seems this approach breaks some feedexport scenarios.

Issue #4253, opened 2019-12-23
Settings, multiple concurrent spiders, and middlewares


mredaelli:
This is more of a question than a feature request, but I guess I can translate it to a request for an enhancement of the documentation.

This is a question I posted on [StackOverflow](https://stackoverflow.com/questions/59424281/scrapy-settings-multiple-concurrent-spiders-and-middlewares), where I didn't get any answers:

=========================

I'm used to running spiders one at a time, because we mostly work with scrapy crawl and on scrapinghub, but I know that one can run [multiple spiders concurrently](https://docs.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process), and I have seen that middlewares often have a spider parameter in their callbacks.

What I'd like to understand is:
- the relationship between Crawler and Spider. If I run one spider at a time, I'm assuming there's one of each. But if you run more spiders together, like in the example linked above, do you have one crawler for multiple spiders, or are they still 1:1?
- is there in any case only one instance of a middleware of a certain class, or do we get one per-spider or per-crawler?
- Assuming there's one, what are the crawler.settings in the middleware creation (for example, [here](https://github.com/scrapy-plugins/scrapy-deltafetch/blob/master/scrapy_deltafetch/middleware.py#L41))? In the [documentation](https://docs.scrapy.org/en/latest/topics/settings.html#topics-settings) it says that those take into account the settings overridden in the spider, but if there are multiple spiders with conflicting settings, what happens?

I'm asking because I'd like to know how to handle spider-specific settings. Take again the [DeltaFetch middleware](https://github.com/scrapy-plugins/scrapy-deltafetch/blob/master/scrapy_deltafetch/middleware.py#L41) as an example:

- enabling it seems to be a global matter, because DELTAFETCH_ENABLED is read from the crawler.settings
- however, the sqlite db is opened in spider_opened and is a unique instance variable (i.e., not depending on the spider); so if you have more than one spider and the instance is shared, when the second spider is opened, the old db is lost. And if you have only one instance of the middleware per spider, why bother passing the spider as a parameter?

Is that a correct way of handling it, or should you rather have a dict spider_dbs indexed by spider name?

elacuesta:
1) There is a 1:1 relationship between `Crawler` and `Spider`. The `CrawlerRunner.crawl` method [creates a new `Crawler`](https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/crawler.py#L183) for each spider it gets.
[...]

PR #6181, opened 2023-12-19
Add component getters to crawler

Gallaecio:
[Suggested](https://github.com/scrapinghub/scrapy-poet/pull/172/files/99ba579a3d25454774c01f44a02df15e60e1a0eb#r1385084947) by @BurnzZ.

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -179,9 +179,12 @@ class Crawler:
             yield maybeDeferred(self.engine.stop)
 
     @staticmethod
-    def _get_component(cls, components):
+    def _get_component(component_class, components):
         for component in components:
-            if type(component) is cls:
+            if (
+                type(component)  # pylint: disable=unidiomatic-typecheck
+                is component_class
+            ):
                 return component
         return None
 
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -207,10 +207,18 @@ class Crawler:
     def get_item_pipeline(self, cls):
         if not self.engine:
             raise RuntimeError(
-                "Crawler.get_item_pipeline() can only be called after "
-                "the crawl engine has been created."
+                "Crawler.get_item_pipeline() can only be called after the "
+                "crawl engine has been created."
             )
         return self._get_component(cls, self.engine.scraper.itemproc.middlewares)
+
+    def get_spider_middleware(self, cls):
+        if not self.engine:
+            raise RuntimeError(
+                "Crawler.get_spider_middleware() can only be called after the "
+                "crawl engine has been created."
+            )
+        return self._get_component(cls, self.engine.scraper.spidermw.middlewares)
 
 
 class CrawlerRunner:
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -181,10 +181,7 @@ class Crawler:
     @staticmethod
     def _get_component(component_class, components):
         for component in components:
-            if (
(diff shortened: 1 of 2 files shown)


wRAR:
Will it be helpful for #6136?

Gallaecio:
No idea, but I suspect not :disappointed:

wRAR:
Ah yeah, it covers different things, but the approach could be useful there.

wRAR (approved):

scrapy/crawler.py
```diff
@@ -178,6 +178,51 @@ def stop(self) -> Generator[Deferred, Any, None]:
             assert self.engine
             yield maybeDeferred(self.engine.stop)
 
+    @staticmethod
+    def _get_component(component_class, components):
+        for component in components:
+            if (
+                type(component)  # pylint: disable=unidiomatic-typecheck
+                is component_class
```
kmike, line 186: I think it'd be good to check what's the best behavor here, based on the use cases in the wild: should the default be to match the type exactly, or to use ininstance (i.e. allow subclasses).

Gallaecio: We could add a boolean parameter to control that, but I guess even then we should check use cases to choose a good default.

Gallaecio: Use cases I could find in the wild:
- `isinstance`: [1](https://stackoverflow.com/q/64495540/939364), [2](https://stackoverflow.com/a/69681922/939364), [3](https://stackoverflow.com/a/40385860/939364), [4](https://stackoverflow.com/a/33007741/939364), [5](https://stackoverflow.com/a/41994630/939364).
- `type` match: none found.
- `str` in: [1](https://stackoverflow.com/a/54776137/939364), [2](https://github.com/scrapy/scrapy/issues/3590#issuecomment-465796924).
- N/A: [1](https://github.com/scrapy/scrapy/issues/4253#issuecomment-568485213), [2](https://stackoverflow.com/a/60705942/939364), [3](https://stackoverflow.com/a/36623006/939364).

So, do we go with `isinstance`? Do we provide a boolean parameter to switch to a `type` match?

wRAR: I assumed checking the exact type is what people would actually want (not sure if real-world code examples show that they don't, as people could just forget about the subclasses).

kmike: The use case for isinstance: there is e.g. a middleware (Scrapy built-in, or provided by some package). Some other object (could be e.g. some extension) uses a components getter to find the middleware, and read some information from it. Then, to change the behavior of the middleware, user makes a subclass, disables the original middleware, and enables the new one. 

The question is: should the extension keep reading the information from the new middleware, or should it behave as is there is no middeware available anymore? 

It seems there are arguments for both. For me it looks like using a subclass is better, because subclass is expected to keep the features of the base class, and extend it. So, +0.5 to do isinstance.

wRAR: > It seems there are arguments for both.

Yeah.

> For me it looks like using a subclass is better, because subclass is expected to keep the features of the base class, and extend it.

Makes sense, let's do isinstance then. Shluld we still have a flag to switch it?

kmike: > Shluld we still have a flag to switch it?

I think we shouldn't do it now, without a real-world use case.

wRAR (approved):

BurnzZ (commented):
🚀

BurnzZ (approved):

kmike (approved):

PR #6198, opened 2024-01-05
refactor(): use `OSError` exception

Rotzbua:
`os.error` is an alias of the preferable `OSError`.

## Reference

https://docs.astral.sh/ruff/rules/os-error-alias/


wRAR (approved):

kmike:
👍

PR #6199, opened 2024-01-05
fix(flake8): lint errors

Rotzbua:
Fixes new errors thrown with Python 3.12. (current gh ci passes because older py version)

## Fixed problems

- E226 missing whitespace around arithmetic operator
- E201 whitespace after '{'

### DIFF
--- a/scrapy/extensions/memusage.py
+++ b/scrapy/extensions/memusage.py
@@ -128,9 +128,9 @@ class MemoryUsage:
     def _send_report(self, rcpts, subject):
         """send notification mail with some additional useful info"""
         stats = self.crawler.stats
-        s = f"Memory usage at engine startup : {stats.get_value('memusage/startup')/1024/1024}M\r\n"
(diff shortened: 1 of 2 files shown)


wRAR:
Is this caught by some newer flake8 than the one we use? Or what's the reason for the check passing without these changes?

wRAR:
If this was true it would happen on all PRs, but it doesn't. Running `pre-commit run --all-files` locally on master also doesn't report these errors. Are you sure you are using the 6.1.0 version and not e.g. some other pre-commit configuration?

wRAR:
> I use python 3.12, but that should not be the cause.

This actually should make a difference because Python 3.12 improved f-string parsing.

> pre-commit run does only check the code changed in the commit

The `Checks / pre-commit` action runs `pre-commit run --show-diff-on-failure --color=always --all-files`. I assume it uses an older Python (we use `pre-commit/action@v3.0.0` and don't specify the Python version).

PR #6201, opened 2024-01-05
Fix and re-enable newer mitmproxy usage in tests.

### LINKED ISSUE #5454: test_proxy_connect.py hangs with mitmproxy 8.0.0 (affecting 3.8 tests)

wRAR:
As usual, a two-character fix.

Fixes #5454.

### DIFF
--- a/tests/test_proxy_connect.py
+++ b/tests/test_proxy_connect.py
@@ -31,6 +31,7 @@ sys.exit(mitmdump())
         self.proc = Popen(
             [
                 sys.executable,
+                "-u",
                 "-c",
                 script,
                 "--listen-host",
@@ -46,7 +47,7 @@ sys.exit(mitmdump())
             stdout=PIPE,
         )
         line = self.proc.stdout.readline().decode("utf-8")
-        host_port = re.search(r"listening at http://([^:]+:\d+)", line).group(1)
+        host_port = re.search(r"listening at (?:http://)?([^:]+:\d+)", line).group(1)
         address = f"http://{self.auth_user}:{self.auth_pass}@{host_port}"
         return address
 
--- a/tox.ini
+++ b/tox.ini
@@ -11,11 +11,7 @@ minversion = 1.7.0
 deps =
     -rtests/requirements.txt
     # mitmproxy does not support PyPy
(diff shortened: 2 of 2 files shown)


tests/test_proxy_connect.py
```diff
@@ -46,7 +47,7 @@ def start(self):
             stdout=PIPE,
         )
         line = self.proc.stdout.readline().decode("utf-8")
-        host_port = re.search(r"listening at http://([^:]+:\d+)", line).group(1)
+        host_port = re.search(r"listening at (?:http://)?([^:]+:\d+)", line).group(1)
```
wRAR, line 50: The output format has changed in some newer version but we can support both.

tests/test_proxy_connect.py
```diff
@@ -31,6 +31,7 @@ def start(self):
         self.proc = Popen(
             [
                 sys.executable,
+                "-u",
```
wRAR, line 34: mitmproxy changed the code that prints log lines and we now need to disable stdout buffering to get it via Popen.

tox.ini
```diff
@@ -11,11 +11,7 @@ minversion = 1.7.0
 deps =
     -rtests/requirements.txt
     # mitmproxy does not support PyPy
-    # Python 3.9+ requires mitmproxy >= 5.3.0
-    # mitmproxy >= 5.3.0 requires h2 >= 4.0, Twisted 21.2 requires h2 < 4.0
-    #mitmproxy >= 5.3.0; python_version >= '3.9' and implementation_name != 'pypy'
-    # The tests hang with mitmproxy 8.0.0: https://github.com/scrapy/scrapy/issues/5454
-    mitmproxy >= 4.0.4, < 8; python_version < '3.9' and implementation_name != 'pypy'
+    mitmproxy; implementation_name != 'pypy'
```
kmike, line 14: is there a minimal mitmproxy version, like `>8`?

kmike (approved):

tox.ini
wRAR: The code should work with any mitmproxy version.

PR #6202, opened 2024-01-08
Make FTPFilesStore compatible with pathlib.Path

y26805:
Related https://github.com/scrapy/scrapy/issues/5739

Per the [docs](https://docs.scrapy.org/en/latest/topics/media-pipeline.html?highlight=FTP#ftp-server-storage), `FILES_STORE` and `IMAGES_STORE` can point to an FTP server.

However, `pathlib` removes a slash from the URI when converted to a string. This is [expected behaviour](https://bugs.python.org/issue44842).
```
from pathlib import Path
p = Path("ftp://username:password@address:21/path")
str(p)
>>> 'ftp:/username:password@address:21/path'
```

As a result, `urllib.parse.urlparse` cannot parse the string-ified path correctly. 
from [docs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse):
> Following the syntax specifications in [RFC 1808](https://datatracker.ietf.org/doc/html/rfc1808.html), urlparse recognizes a netloc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component.
```
p = "ftp://username:password@address:21/path"
urlparse(str(Path(p)))
>>> ParseResult(scheme='ftp', netloc='', path='/username:password@address:21/path', params='', query='', fragment='')

urlparse(p)
>>> ParseResult(scheme='ftp', netloc='username:password@address:21', path='/path', params='', query='', fragment='')
```

Added a test and an ugly fix for this. Am open to other suggestions on how to best fix this.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -265,10 +265,12 @@ class FTPFilesStore:
     FTP_PASSWORD = None
     USE_ACTIVE_MODE = None
 
-    def __init__(self, uri):
-        if not uri.startswith("ftp://"):
-            raise ValueError(f"Incorrect URI scheme in {uri}, expected 'ftp'")
+    def __init__(self, uri: str):
         u = urlparse(uri)
+        if u.scheme != "ftp":
+            raise ValueError(f"Incorrect URI scheme in {uri}, expected 'ftp'")
+        if not uri.startswith("ftp://"):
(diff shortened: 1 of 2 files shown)


wRAR:
It makes sense that creating a Path object from an URI string shouldn't be done, where is it currently done?

PR #6209, opened 2024-01-22
Improved Image / File URL Handling | Handle Check if single Image / File URL needs to be Downloaded 

IkramKhanNiazi:
- Previously, we need need to Download a Single Image / File URL, still, we need to make that URL a list i.e. `image_urls: [URL] or files_urls_field: [URL]`
- After this fix, we don't need to make it a list, `image_urls: URL` and `files_urls_field: URL`  will work.
- A better user experience.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -525,6 +525,7 @@ class FilesPipeline(MediaPipeline):
     # Overridable Interface
     def get_media_requests(self, item, info):
         urls = ItemAdapter(item).get(self.files_urls_field, [])
+        urls = urls if isinstance(urls, list) else [urls]
         return [Request(u, callback=NO_CALLBACK) for u in urls]
 
     def file_downloaded(self, response, request, info, *, item=None):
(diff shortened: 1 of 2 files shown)


kmike:
Hey @IkramKhanNiazi! Thanks for the effort; it's good approach to sumit simple PRs to discuss the ideas 👍 

About the proposal itself, currently I'm not convinced it's a change we should make, because of the following reasons:

1. First, it seems the cases where it helps are quite rare. I've checked some internal codebases, and haven't found compelling examples - of course, it may depend on a codebase.  
2. Even if there are such cases, `[url]` doesn't look that bad.
3. The fields are named in plural, so it makes a reader pause when they see `image_urls = url`. If I'd be reading such code, I'd want to double check if url is a string or a list, spending mental effort on this.
4. We need to document, test and maintain the feature. For example, the current implementation is not correct, and can break user's code, if some iterable is assigned to image_urls, which is not a list:  e.g. a tuple with several URLs would be incorrectly wrapped in another list. Of course, it's solvable, but it shows that maintenance doesn't come for free. 
5. There is also an additional mental burden for the Scrapy user. It's slightly more documentation, slightly more things to keep in mind, slightly more surprising code which you can write, etc. It's definitely not a lot, but such things add up.

Gallaecio:
Thank you for your PR, but I have to agree with @kmike’s points here.

PR #6213, opened 2024-01-31
Cleanup deprecated fingerprint code in scrapy.utils.request

### LINKED ISSUE #6212: Cleanup deprecated fingerprint code in scrapy.utils.request
`scrapy.utils.request.request_fingerprint()` and `REQUEST_FINGERPRINTER_IMPLEMENTATION=2.6` were deprecated in Scrapy 2.7.0 so that code should be cleaned up, though I haven't looked what exactly should be changed/removed, it may be non-trivial.

Laerte:
Closes #6212


Gallaecio:
My plan here was to deprecate `REQUEST_FINGERPRINTER_IMPLEMENTATION` itself now, and remove the setting altogether in a year.

Laerte:
> My plan here was to deprecate `REQUEST_FINGERPRINTER_IMPLEMENTATION` itself now, and remove the setting altogether in a year.

Should i add a deprecation warning to `RequestFingerprinter` constructor then?

Gallaecio:
I’m open when it comes to how to implement that, but one way to go about it would be to switch the default to some sentinel value (e.g. "SENTINEL"), and log a deprecation message if the value is something else, asking for the removal of the setting altogether.

And we should stop adding the setting to settings.py when using `startproject`.

scrapy/utils/request.py
```diff
@@ -258,32 +141,22 @@ def __init__(self, crawler: Optional["Crawler"] = None):
...
-        elif implementation == "2.7":
             self._fingerprint = fingerprint
         else:
             raise ValueError(
                 f"Got an invalid value on setting "
                 f"'REQUEST_FINGERPRINTER_IMPLEMENTATION': "
-                f"{implementation!r}. Valid values are '2.6' (deprecated) "
-                f"and '2.7'."
+                f"{implementation!r}. Valid values are '2.7' and 'SENTINEL'."
             )
```
Gallaecio, line 160: What about:

```python
if implementation != "SENTINEL":
    warn(…)
self._fingerprint = fingerprint
```

to replace the whole thing?

i.e. We no longer care if they use an invalid value or not, we ignore whatever value they set, and just warn them to stop using the setting.

scrapy/utils/test.py
```diff
@@ -76,7 +76,6 @@ class TestSpider(Spider):
 def get_crawler(
     spidercls: Optional[Type[Spider]] = None,
     settings_dict: Optional[Dict[str, Any]] = None,
-    prevent_warnings: bool = True,
```
Gallaecio, line 79: I think we might want to keep this one here, as this is part of the public API. Not only for backward compatibility, but also forward compatibility, maybe we will need to set some default settings like these in the future.

Although you will probably need to add `del prevent_warnings` in the beginning of the implementation so that linters do not complain about the unused parameter.

scrapy/utils/request.py
Laerte: Done.

scrapy/utils/test.py
Laerte: Seems that `pass` works, let's see.

Gallaecio (approved):

scrapy/utils/test.py
```diff
@@ -87,7 +87,7 @@ def get_crawler(
     # Set by default settings that prevent deprecation warnings.
     settings: Dict[str, Any] = {}
     if prevent_warnings:
-        settings["REQUEST_FINGERPRINTER_IMPLEMENTATION"] = "2.7"
+        pass
```
wRAR, line 90: Just remove these lines? :)

Laerte: Seems that work as well, local pre-commit pass, cc: @Gallaecio

wRAR: (if some linters complain and these lines are needed, we should add some comments to them to explain why they are needed)

wRAR (approved):

PR #6131, opened 2023-11-03
Remove docs #5920

nluginbill:
Description:
This PR addresses the reviews given on another open pull request that is inactive: https://github.com/scrapy/scrapy/pull/6043.

I made the changes suggested by the reviewer in that pull request. I deleted some suggested text, and changed a hyperlink to a Sphinx reference. 

Resolves https://github.com/scrapy/scrapy/issues/5920

PR #6215, opened 2024-02-02
Removed the auto statements

ArSenic04:
auto* statements for things from other projects, like item adapter have been removed


wRAR:
Please look at #6131 and the comments there.

PR #6220, opened 2024-02-13
Updated Documentation

DarrenChen2025:
- Added a table of contents
- Made an example of using scrapy to web scrape content
- Added ways to install scrapy


wRAR:
Most of this is bad and/or unnecessary, sorry.

PR #6224, opened 2024-02-16
Add TerminalConsole test_without_start_listening

cclauss:
Class variables should always be initialized in the `.__init__()` method but `TerminalConsole.port` is not.  As discussed in:
* #2702
* #2892


wRAR:
Is this a work in progress?

cclauss:
> builtins.AttributeError: 'TelnetConsole' object has no attribute 'port'

PR #6227, opened 2024-02-19
Implemented logger that prints out IDs to the terminal if branch reac…

Flug0:
…hed and added ids to _process_spider_output #1

The logger function could get write to file support but i had problems with pytest so its all through the terminal.


wRAR:
Did you intend to publish this?

Flug0:
oh, woops, no sorry

PR #6233, opened 2024-02-20
Added tests for http2 stream

elramen:
Added tests to increase coverage.

### DIFF
--- a/tests/test_http2_stream.py
+++ b/tests/test_http2_stream.py
@@ -1,0 +1,20 @@
+from unittest import TestCase
+from unittest.mock import Mock
+
+from h2.exceptions import StreamClosedError
+from scrapy.core.http2.stream import Stream, StreamCloseReason
+from scrapy.http import Request
+
+
+class Http2StreamTestCase(TestCase):
+
+    def test_close_when_stream_closed_server(self):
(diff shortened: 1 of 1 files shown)


tests/test_http2_stream.py
```diff
@@ -0,0 +1,18 @@
+from unittest import TestCase
+from unittest.mock import Mock
+
+from scrapy.core.http2.stream import Stream, StreamClosedError, StreamCloseReason
+from scrapy.http import Request
+
+
+class Http2StreamTestCase(TestCase):
+    def test_close_when_stream_closed_server(self):
+        stream = Stream(0, Request(url="http://fakeUrl.com"), Mock())
```
Gallaecio, line 10:
```suggestion
        stream = Stream(0, Request(url="http://example.com"), Mock())
```

Gallaecio (commented):
Looking at the code that this is testing, I think the right approach here, rather than creating these tests, would be to replace the code they test with asserts.

In the case of the close reason, looking at the code that calls `close()` (in `stream.py` and `protocol.py`), it seems it should always be a valid reason.

In the case of `self.metadata["stream_closed_server"]`, I am less sure if some server could trigger the issue. But even if it did, I am not sure the exception raised would help, so an assert would make sense to me as well.

PR #6234, opened 2024-02-20
Refactor check command to reduce complexity

elramen:
Current cyclomatic complexity of the run method is 13. This split reduces it to 8. 

Contributing to scrapy is part of a school project.


Gallaecio (commented):
I’m not strongly against it, but I’m not sure this is a change for the better in readability.

PR #6237, opened 2024-02-21
Refactor _get_inputs to reduce complexity

noon-io:
As part of a school assignment, the CCN of _get_inputs has been lowered from 16 to 7 by splitting up two blocks into sub-functions.


Gallaecio (commented):
I‘m not strongly against the change, but the new code seems slightly *less* readable to me.

PR #6246, opened 2024-02-26
Allow disabling the AutoThrottle extension for a given slot

Gallaecio:
Built on top of https://github.com/scrapy/scrapy/pull/6245.

### DIFF
--- a/scrapy/extensions/throttle.py
+++ b/scrapy/extensions/throttle.py
@@ -16,6 +16,11 @@ class AutoThrottle:
         self.target_concurrency = crawler.settings.getfloat(
             "AUTOTHROTTLE_TARGET_CONCURRENCY"
         )
+        if self.target_concurrency <= 0.0:
+            raise NotConfigured(
+                f"AUTOTHROTTLE_TARGET_CONCURRENCY "
+                f"({self.target_concurrency!r}) must be higher than 0."
+            )
         crawler.signals.connect(self._spider_opened, signal=signals.spider_opened)
         crawler.signals.connect(
             self._response_downloaded, signal=signals.response_downloaded
@@ -45,7 +50,7 @@ class AutoThrottle:
     def _response_downloaded(self, response, request, spider):
         key, slot = self._get_slot(request, spider)
         latency = request.meta.get("download_latency")
-        if latency is None or slot is None:
+        if latency is None or slot is None or slot.throttle is False:
             return
 
         olddelay = slot.delay
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -2,7 +2,7 @@ import random
 from collections import deque
 from datetime import datetime
 from time import time
-from typing import TYPE_CHECKING, Any, Deque, Dict, Set, Tuple, cast
+from typing import TYPE_CHECKING, Any, Deque, Dict, Optional, Set, Tuple, cast
 
 from twisted.internet import task
 from twisted.internet.defer import Deferred
@@ -24,10 +24,18 @@ if TYPE_CHECKING:
 class Slot:
     """Downloader slot"""
 
-    def __init__(self, concurrency: int, delay: float, randomize_delay: bool):
+    def __init__(
+        self,
+        concurrency: int,
+        delay: float,
+        randomize_delay: bool,
+        *,
+        throttle: Optional[bool] = None,
+    ):
         self.concurrency: int = concurrency
         self.delay: float = delay
         self.randomize_delay: bool = randomize_delay
+        self.throttle = throttle
 
         self.active: Set[Request] = set()
         self.queue: Deque[Tuple[Request, Deferred]] = deque()
@@ -52,13 +60,15 @@ class Slot:
         return (
             f"{cls_name}(concurrency={self.concurrency!r}, "
             f"delay={self.delay:.2f}, "
-            f"randomize_delay={self.randomize_delay!r})"
+            f"randomize_delay={self.randomize_delay!r}, "
+            f"throttle={self.throttle!r})"
         )
 
     def __str__(self) -> str:
         return (
             f"<downloader.Slot concurrency={self.concurrency!r} "
             f"delay={self.delay:.2f} randomize_delay={self.randomize_delay!r} "
+            f"throttle={self.throttle!r} "
             f"len(active)={len(self.active)} len(queue)={len(self.queue)} "
             f"len(transferring)={len(self.transferring)} "
             f"lastseen={datetime.fromtimestamp(self.lastseen).isoformat()}>"
--- a/tests/test_extension_throttle.py
+++ b/tests/test_extension_throttle.py
@@ -157,16 +157,17 @@ def test_startdelay_definition(min_spider, min_setting, start_setting, expected)
 
 
 @pytest.mark.parametrize(
-    ("meta", "slot"),
+    ("meta", "slot", "throttle"),
     (
-        ({}, None),
-        ({"download_latency": 1.0}, None),
-        ({"download_slot": "foo"}, None),
-        ({"download_slot": "foo"}, "foo"),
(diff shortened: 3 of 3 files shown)


wRAR (approved):

VMRuiz (approved):

GeorgeA92:
@Gallaecio 
I didn't notice this earlier.
As a developer who was responsible for preparing/testing actual implementation of `DOWNLOAD_SLOTS` I don't like approach to implement compatibility with AutoThrottle extension  from this Pull request.

During preparation of https://github.com/scrapy/scrapy/pull/5328 compatibility with Autothrottle extenstion didn't mentioned as  requirement. It tested on default settings (with disabled autothrottle extension)

Updating AutoThrottle extension to make it compatible with `DOWNLOAD_SLOT` setting values - makes sense only if all of its settings `AUTOTHROTTLE_TARGET_CONCURRENCY`, `AUTOTHROTTLE_ENABLED`, `AUTOTHROTTLE_START_DELAY`, `AUTOTHROTTLE_MAX_DELAY`  will be possible to set on per download slot basis (for example by just adding related supported keys for exiting `DOWNLOAD_SLOTS` setting) and from scrapy user perspective most likely this will be expected behaviour for `DOWNLOAD_SLOTS` with enabled AutoThrottle extention.

While disabling throttle on specific slot +- looks like applying of `AUTOHTROTTLE_ENABLED` only.

And there is no need to.. make backward incompatible changes to scrapy/core downloader slot to implement this.

Technically nothing prevents us to update `downloader.per_slot_settings` dict used for slots initially listed on `DOWNLOAD_SLOTS` setting during runtime from Autothrottle extension - as this extension affect downloader variables directly already

Gallaecio:
So, what you are suggesting is to implement an AutoThrottle-specific setting, e.g. `AUTOTHROTTLE_SKIP_SLOTS`? That sounds better to me indeed for the scenario mentioned in the docs change here.

However, my end goal here, the reason why I implemented this, is to be able to disable AutoThrottle on [slots created by scrapy-zyte-api](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/fc57a082a6920dbd900fb0d25d6c8871f2f5e91e/scrapy_zyte_api/_middlewares.py#L27-L37), so that we can [implement latency reporting without triggering AutoThrottle](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/59#discussion_r1020056364). For that, I don’t think a setting would be a good choice.

I am completely open to suggestions, though. I really do not like modifying a core component to control an extension behavior, it is simply the cleanest thing I could come up with.

GeorgeA92:
@Gallaecio 
> So, what you are suggesting is to implement an AutoThrottle-specific setting, e.g. AUTOTHROTTLE_SKIP_SLOTS? That sounds better to me indeed for the scenario mentioned in the docs change here.

I mean to something like ` "quotes.toscrape.com": {"concurrency": 1, "delay": 2, "randomize_delay": False, "AUTOTHROTTLE_START_DELAY": 5 , "AUTOTHROTTLE_MAX_DELAY":25},` etc.

> However, my end goal here, the reason why I implemented this, is to be able to disable AutoThrottle on [slots created by scrapy-zyte-api](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/fc57a082a6920dbd900fb0d25d6c8871f2f5e91e/scrapy_zyte_api/_middlewares.py#L27-L37), so that we can https://github.com/scrapy-plugins/scrapy-zyte-api/pull/59#discussion_r1020056364. For that, I don’t think a setting would be a good choice.

This will require to.. update params for all download slots with `_slot_prefix = "zyte-api@"`  which is not implemented in `DOWNLOAD_SLOTS` (will require to hardcode all expected slots with prefixed).

> I am completely open to suggestions, though. I really do not like modifying a core component to control an extension behavior, it is simply the cleanest thing I could come up with.

My suggestion is to update.. method of `AutoThrottle` extension `_adjust_delay` or `_response_downloaded` to not make changes if slot has `_slot_prefix `.
I mean changing this during runtime on active `AutoThrottle` instance from` ScrapyZyteAPIDownloaderMiddleware.open_spider` it's `spider_opened` signal handler.
It can be:
1. monkeypathing of `_adjust_delay` method.
2. diconnect `AutoThrottle._response_downloaded` signal handler and connect it's updated method (I did similar tricks with signal handlers before)

This approach doesn't require to change both downloader and autothrottle ext (only zyte api middleware code required to update)

Gallaecio:
What about exposing a `AUTOTHROTTLE_SLOT_FILTER` setting that accepts a Scrapy component for filtering, and set that from scrapy-zyte-api to a component that filters out our slots?

GeorgeA92:
@Gallaecio 
If end goal of this to affect downloader slots from zyteapi middleware/download hanlder - then (if possible - I think it is possible) only zyteapi ext. related code needs to be updated or fixed (with no additional changes to scrapy itself) - this is logic behind approach I proposed in https://github.com/scrapy/scrapy/pull/6246#issuecomment-1998697987

> What about exposing a AUTOTHROTTLE_SLOT_FILTER setting that accepts a Scrapy component for filtering, and set that from scrapy-zyte-api to a component that filters out our slots?

With approach I proposed earlier - for end user It will be possible to receive expected result only after updating version of zyteapi middleware. This approach (adding another setting on scrapy code level) as well as updates from this PR also require to update scrapy to the latest version(that include this change) that.. may not be possible for some projects.

Gallaecio:
> only zyteapi ext. related code needs to be updated

Yes, but if I understood correctly, you are suggesting to monkey-patch Scrapy code from the extension, which I would like to avoid.

> require to update scrapy to the latest version

I think this is OK, since the only thing we need this for is to support latency reporting for responses coming for scrapy-zyte-api, and most users will probably not care about that.

Gallaecio:
> As far as I know scrapy-zyte-api (handler) doesn't set `download_latency` as regular scrapy download handlers do

This was not done on purpose, though, it was something we accidentally forgot to implement, i.e. a bug. And by the time we realized it, we were already exploiting this bug to prevent AutoThrottle from affecting Zyte API traffic. But we do want to fix the bug, i.e. implement download_latency, only we need to first figure out a way to handle AutoThrottle that does not depend on download_latency not being defined.

> What if for single scrapy request - zyteapi download handler will make 4 retries and 1 200s (5 requests) with latencies like `[2.5, 6, 0.24, 1.5, 60]` ? Autothrottle extension can accept `download_latency` as single number only

I think this is fine. `download_latency` is the time the request spends in the download handler. If a request takes longer because of 429 or 521 retries, so be it, and the `download_latency` would in practice be the sum of the time every attempt took until success or exceeding retries.

> I propose to.. update scrapy-zyte-api.. to write (or to not write) something in `download_latency` request meta key as trigger to disable/enable autothrottle for that requests

I think putting a value in `download_latency` other than the time the request spent in the download handler would not be a good choice. It would break the expectations of user code using `download_latency` about what `download_latency` stands for.

scrapy/extensions/throttle.py
```diff
@@ -50,7 +50,7 @@ def _start_delay(self, spider):
     def _response_downloaded(self, response, request, spider):
         key, slot = self._get_slot(request, spider)
         latency = request.meta.get("download_latency")
-        if latency is None or slot is None:
+        if latency is None or slot is None or slot.throttle is False:
```
kmike, line 53: On a first sight, it's possible to add `dont_throttle` meta key, and avoid connecting core components with extensions.

Gallaecio: Are we talking about implementing a downloader middleware that applies that meta key based on the `DOWNLOAD_SLOTS` setting, i.e. that setting would be managed by the core expect for that specific key? Or do you have something else in mind for the user-facing API?

kmike: I was thinking that this feature can be unrelated to slots: just a per-request flag to disable autothrottle.
If this solves the scrapy-zyte-api issue, we can bypass all DOWNLOAD_SLOTS discussions for now :)

PR #6248, opened 2024-02-26
Change password on line 49 of the class test/test_extension_telnet

mery16q:


Gallaecio:
Those are fictional passwords in tests, they are not sensitive information.

PR #6251, opened 2024-02-27
Use defusedxml.xmlrpc

### LINKED ISSUE #6250: Use `defusedxml.xmlrpc`
https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html#b411-import-xmlrpclib

https://github.com/tiran/defusedxml?tab=readme-ov-file#defusedxmlxmlrpc

Laerte:
Fix #6250

### DIFF
--- a/scrapy/http/request/rpc.py
+++ b/scrapy/http/request/rpc.py
@@ -7,12 +7,12 @@ See documentation in docs/topics/request-response.rst
 import xmlrpc.client as xmlrpclib
 from typing import Any, Optional
 
-from defusedxml import xmlrpc
+import defusedxml.xmlrpc as xml_rpc
 
 from scrapy.http.request import Request
 from scrapy.utils.python import get_func_args
 
-xmlrpc.monkey_patch()
+xml_rpc.monkey_patch()
 
 DUMPS_ARGS = get_func_args(xmlrpclib.dumps)
 
--- a/scrapy/http/request/rpc.py
+++ b/scrapy/http/request/rpc.py
@@ -7,12 +7,12 @@ See documentation in docs/topics/request-response.rst
 import xmlrpc.client as xmlrpclib
 from typing import Any, Optional
 
-import defusedxml.xmlrpc as xml_rpc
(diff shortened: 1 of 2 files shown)


Gallaecio:
@wRAR @Laerte Any idea why bandit is passing even though `B411` is not in https://github.com/scrapy/scrapy/blob/master/.bandit.yml ?

Gallaecio (approved):

wRAR:
B411 is only emitted by a newer version, I haven't published a PR that bumps it yet.

scrapy/http/request/rpc.py
```diff
@@ -7,9 +7,13 @@
 import xmlrpc.client as xmlrpclib
 from typing import Any, Optional
 
+from defusedxml import xmlrpc
```
wRAR, line 10: I think it's better to import `defusedxml.xmlrpc`, as `xmlrpc` is the name of the stdlib module so this can cause confusion (and maybe name clashes if we need to import the stdlib one in the future).

Laerte: Pushed.

scrapy/http/request/rpc.py
```diff
@@ -7,9 +7,13 @@
 import xmlrpc.client as xmlrpclib
 from typing import Any, Optional
 
+import defusedxml.xmlrpc as xml_rpc
```
wRAR, line 10: I don't think this alias is much better, as I think the only alias that makes sense here is something like `defusedxml_xmlrpc` and I don't think that's better than just importing `defusedxml.xmlrpc`, unless I'm missing some reasons why it's not possible.

Laerte: Agreed indeed can cause confusion, just pushed the suggested change!

wRAR (approved):
Thanks!

wRAR:
Note that https://github.com/PyCQA/bandit/issues/1082 applies here, so we will still need to silence the error until the fix detection is added/fixed.

PR #6253, opened 2024-02-28
Update tool versions, fix some of the pylint problems

wRAR:


.flake8
```diff
@@ -1,7 +1,7 @@
 [flake8]
 
 max-line-length = 119
-ignore = W503, E203
+ignore = E203, E501, E701, E704, W503
```
wRAR, line 4: All formatting-related.

scrapy/utils/signal.py
```diff
@@ -97,7 +98,10 @@ def logerror(failure: Failure, recv: Any) -> Failure:
             robustApply, receiver, signal=signal, sender=sender, *arguments, **named
         )
         d.addErrback(logerror, receiver)
-        d.addBoth(lambda result: (receiver, result))
+        # TODO https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/cell-var-from-loop.html
```
wRAR, line 101: The article suggests that the code doesn't work the way it looks, we should investigate if this indeed applies to our code (I would be surprised that signals are sent to wrong places as we would notice that).

Gallaecio (approved):

PR #6257, opened 2024-02-28
Fix and re-enable `unnecessary-comprehension` and `use-dict-literal` pylint tags

### LINKED ISSUE #6254: Fix and re-enable `unnecessary-comprehension` and `use-dict-literal` pylint tags
Both are valid simplification hints.

Laerte:
Fixes #6254


wRAR (approved):
Thanks!

PR #6261, opened 2024-02-29
Solve incompatibility errors in _compression.py

ygalvao:
Increase compatibility across different virtual environments, particularly with Conda, when using brotli.Decompressor()

### DIFF
--- a/scrapy/utils/_compression.py
+++ b/scrapy/utils/_compression.py
@@ -61,7 +61,10 @@ def _unbrotli(data: bytes, *, max_size: int = 0) -> bytes:
     decompressed_size = 0
     while output_chunk:
         input_chunk = input_stream.read(_CHUNK_SIZE)
-        output_chunk = decompressor.process(input_chunk)
+        try:
+            output_chunk = decompressor.process(input_chunk)
+        except AttributeError:
+            output_chunk = decompressor.decompress(input_chunk)
         decompressed_size += len(output_chunk)
         if max_size and decompressed_size > max_size:
             raise _DecompressionMaxSizeExceeded(
--- a/scrapy/utils/_compression.py
+++ b/scrapy/utils/_compression.py
@@ -5,6 +5,19 @@ try:
     import brotli
 except ImportError:
     pass
+else:
+    try:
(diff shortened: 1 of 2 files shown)


wRAR:
Is this only required when `brotlipy` is installed instead of/in addition to `brotli`?

ygalvao:
Yes. I don't know why, but when setting up conda environments and installing scrapy and other packages it seems Conda / Mamba installs only brotlipy.

Gallaecio:
Maybe we should do something to make Scrapy distinguish the 2, and only enable brotli support when the right one is installed.

wRAR:
Yeah.

wRAR:
> when setting up conda environments and installing scrapy and other packages it seems Conda / Mamba only installs brotlipy.

Note that Scrapy currently does not depend on brotli (or on brotlipy), neither on PyPI nor on Conda, and if something installs brotlipy as opposed to not installing anything it's either "other packages" you mentioned or some other things you do when "setting up".

ygalvao:
I understand. However, in my current conda env, I just have a few packages installed, most of them well-known ones like Scrapy, Selenium, Pandas, Numpy, etc. and somehow, Conda or Mamba installed brotlipy, which throws an AttributeError when running Scrapy. So this simple change solves this issue, adding compatibility with brotlipy and hence with other packages installed using Conda / Mamba as well.

scrapy/utils/_compression.py
```diff
@@ -5,6 +5,19 @@
     import brotli
 except ImportError:
     pass
+else:
+    try:
+        brotli.Decompressor.process
+    except AttributeError:
+
+        def _brotli_decompress(decompressor, data):
```
wRAR, line 13: Should we add a warning here?

By the way `brotlipy` is now `brotlicffi` (with the import name adjusted accordingly), but whether we need to add support for it is a separate question.

Gallaecio: [Sounds like support may make sense for PyPy support](https://github.com/urllib3/urllib3/issues/2040), but indeed out of scope here.

wRAR (approved):

PR #6262, opened 2024-02-29
Do not use incompatible brotlipy if installed

Gallaecio:
Resolves #6261.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -226,3 +226,12 @@ setenv =
     {[pinned]setenv}
 commands =
     pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:tests -k s3}
+
+
+# Run brotlipy-specific tests.
+[testenv:brotlipy]
+deps =
+    {[testenv]deps}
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -226,3 +226,12 @@ setenv =
...
     pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:tests -k s3}
+
+
+# Run brotlipy-specific tests.
+[testenv:brotlipy]
+deps =
+    {[testenv]deps}
+    brotlipy
+commands =
+    pytest --cov=scrapy --cov-report=xml --cov-report= {posargs:tests -k brotlipy}
```
Gallaecio, line 237: Overkill?

wRAR: Yeah, it feels like overkill to me.

Gallaecio:
Let’s go with https://github.com/scrapy/scrapy/pull/6261 ’s approach to allow brotlipy as well (unofficially), but maybe move the method selection to import time.

PR #6264, opened 2024-03-01
Mark hashlib usages as not intended for security

Gallaecio:


scrapy/utils/misc.py
```diff
@@ -113,7 +113,15 @@ def md5sum(file: IO) -> str:
...
-    m = hashlib.md5()
+    warnings.warn(
+        (
+            "The scrapy.utils.misc.md5sum function is deprecated, and will be "
+            "removed in a future version of Scrapy."
+        ),
+        ScrapyDeprecationWarning,
+        stacklevel=2,
+    )
+    m = hashlib.md5(usedforsecurity=False)
```
Gallaecio, line 124: Since whether the function is intended for security or not depends on the caller, I thought it best to deprecate this function and reimplement it as private in the media pipelines, which is where it was being used internally, and where we can ensure it is not being used for security purposes.

wRAR (approved):

PR #6265, opened 2024-03-01
Bandit: allow-list lxml usages

Gallaecio:
In most cases, the actual loading of the document is done by parsel, not Scrapy.

`form.py` was 1 exception, but I have refactored it to use the response selector instead. The application of `get_base_url` in `unified.py` was a bug fix detected in the process.

Another exception is `iterparse_lxml`, which now uses `resolve_entities=False` as parsel does.

And then there was the sitemap code, which was already disabling entity resolution.

### DIFF
--- a/scrapy/http/request/form.py
+++ b/scrapy/http/request/form.py
@@ -10,21 +10,16 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Tuple, Union, cast
 from urllib.parse import urlencode, urljoin, urlsplit, urlunsplit
 
-from lxml.html import (
-    FormElement,
-    HTMLParser,
-    InputElement,
-    MultipleSelectOptions,
-    SelectElement,
-    TextareaElement,
-)
-from parsel.selector import create_root_node
+from lxml.html import FormElement  # nosec
+from lxml.html import InputElement  # nosec
+from lxml.html import MultipleSelectOptions  # nosec
+from lxml.html import SelectElement  # nosec
+from lxml.html import TextareaElement  # nosec
 from w3lib.html import strip_html5_whitespace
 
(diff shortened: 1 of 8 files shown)


scrapy/http/request/form.py
```diff
@@ -10,21 +10,16 @@
...
-from lxml.html import (
-    FormElement,
-    HTMLParser,
-    InputElement,
-    MultipleSelectOptions,
-    SelectElement,
-    TextareaElement,
-)
-from parsel.selector import create_root_node
+from lxml.html import FormElement  # nosec
```
Laerte, line 13: Question: We need to change import style because is not possible to use # no sec otherwise? Thanks

Gallaecio: isort did that automatically, not sure why, I imagine it is a style choice.

Laerte: Got it! Thanks!

scrapy/http/request/form.py
```diff
@@ -120,7 +115,7 @@ def _get_form(
     formxpath: Optional[str],
 ) -> FormElement:
     """Find the wanted form element within the given response."""
-    root = create_root_node(response.text, HTMLParser, base_url=get_base_url(response))
+    root = response.selector.root
```
wRAR, line 118: Can you please explain this? Also does this work with all supported parsel versions?

Gallaecio: Some of the code later on relies on attributes from `lxml.etree.Element` not exposed by `Selector` (e.g. `name`), so using the selector all the way was not possible.

[`Selector.root` was already available in the lowest parsel version Scrapy supports](https://github.com/scrapy/parsel/blob/v1.5.0/parsel/selector.py#L200).

wRAR (approved):

PR #6273, opened 2024-03-06
add support for custom exporter class

guillermo-bondonno:
Opening this PR to discuss the implementation of adding support for custom exporter classes. This comes from the idea of adding stream exporters that don't rely on local storage. The usage I had in mind would be something like:

```
custom_settings = {
        "FEEDS": {
            "custom:": {
                "custom_exporter_class": "project.exporters.StreamExporter",
                "format" : None,
                "item_export_kwargs" : {"myarg1": "myvalue1"}, #needs dont_fail=True
            }
        }
    }
```

```
class CustomExporter(BaseItemExporter):
    def __init__(self, *, dont_fail=True, **kwargs):
        self._configure(kwargs, dont_fail=dont_fail)
        self.this_slot_items = []

    def export_item(self, item):
        #example
        requests.post(
            "https://some-api.com", json=item
        )
        #or
        self.this_slot_items.append(item)

    def finish_exporting(self):
        #do something here
        pass
```
The batching would work the same, since a new instance would be created for every new slot and the custom exporter can handle the queuing, etc.

After this, reusable exporters for widely used services can be implemented.
Missing tests and docs.


VMRuiz:
Hello @guillermo-bondonno , 

What would be the difference between this approach and just doing: 

```PYTHON
FEED_STORAGE = {
    'custom_service': 'project.exporters.custom_exporter'
}

FEED = {
   'custom_service://some_path' :  {  
      'format': None,
      'myarg1': 'myvalue1'
   }
}

```

scrapy/settings/default_settings.py
```diff
@@ -154,6 +154,7 @@
     "gs": "scrapy.extensions.feedexport.GCSFeedStorage",
     "s3": "scrapy.extensions.feedexport.S3FeedStorage",
     "stdout": "scrapy.extensions.feedexport.StdoutFeedStorage",
+    "custom": "scrapy.extensions.feedexport.StdoutFeedStorage",
```
Gallaecio, line 157: Is this a leftover from debugging?

VMRuiz:
Hello, I like the idea of using the Feed Exporters rather than Feed Storages to do the actual exporting. You don't actually need to write any additional code to support this right now, this should be similiar to your solution: 

```python
FEED = {
    stdout:// {
       format: "StreamingAPI"
   }
}
FEED_EXPORTERS = { 
   "StreamingAPI": "StreamingAPIExporter"
}
```

The methods `start_exporting` and `finish_exporting` could be used to initialization, authentication, batching, etc...

However, there are several downsides that we would need to figure out: 

1. `export_item` is an `async` method, meaning that any blocking call like `requests.post` would block the entire Scrapy process
2. We would need to figure out how to properly manage networking / API errors and allowing for retries.

ftadao:
> 1. `export_item` is an sync method, meaning that any blocking call like `requests.post` would block the entire Scrapy process

Wouldn't this be something that the developer needs to take care of on their customized exporter code? If we're allowing the customization of `def export_item`, any developer can also adapt it to be an `async` function and remove that barrier on its terms. Of course, that might open up possible adaptations that we might do to the underlying exporter code (e.g., allow `export_item` to be an async function), but it might make the code a little more robust after we go through it.

> 2. We would need to figure out how to properly manage networking / API errors and allowing for retries.

Couldn't all this be managed by the custom exporter class itself? If we assume that there's only a limited amount of exceptions that are covered by the engine, wouldn't this be something managed by the code itself?

I think those questions can be answered better if we have an example of a customized exporter that matches those requirements. @VMRuiz what about the Airtable exporter we've previously discussed?

guillermo-bondonno:
Reopening this since it closed when I reset the branch, because the changes were not necessary. I tried @VMRuiz proposal for custom exporters and runs just fine. 
Using `defer.maybeDeferred` to export the item you can run async requests inside export_item like below. 

```
custom_settings = {
        "FEEDS" : {
            "stdout://": {
               "format": "StreamingAPI"
           }
        },
        "FEED_EXPORTERS" : { 
           "StreamingAPI": "stream_export_project.exporters.StreamExporter"
        }
    }
```
```
class StreamExporter(BaseItemExporter):
    """
    A new instance will be created for each slot/batch
    """
    def __init__(self, file, *args, dont_fail=False, **kwargs):
        self._kwargs = kwargs
        self.file = file #Or other name in case other format is used
        self._configure(kwargs, dont_fail=dont_fail)

    def start_exporting(self):
        pass

    def send_request(self, item):
        requests.post(
            "https://example.com/endpoint", json=item | self._kwargs
        )
        
    def export_item(self, item):
        dfd = threads.deferToThread(self.send_request, item)
        # maybe collect these to ensure them in finish exporting
        return dfd

    def finish_exporting(self):
        pass
```

The failing tests confuse me, because from the error they seem to be related to `ExceptionJsonItemExporter`  , but the actual tests don't use it

guillermo-bondonno:
Turns out without changing any Scrapy code I get the same behavior by calling `deferToThread` from export_item, I'll look deeper into it to see if it's safe and reliable to do that

VMRuiz:
@Gallaecio @wRAR , Do you think using deferToThread is appropriate here, or could there be performance issues if there are too many calls?

wRAR:
I suggest using `deferToThread` from time to time but I've never used it myself :) It won't create a new thread for every item as it uses a thread pool so I think the performance shouldn't be worse.

scrapy/extensions/feedexport.py
```diff
@@ -481,11 +482,11 @@ def get_file(slot_):
 
         if slot.itemcount:
             # Normal case
-            slot.finish_exporting()
+            finish_exporting_maybe_deferred = defer.maybeDeferred(slot.finish_exporting)
```
wRAR, line 485: I'd remove maybe from the name, it's an actual deferred.

wRAR:
Note that this has conflicts with master.

VMRuiz (changes requested):

Gallaecio:
If we go ahead with adding stream to feedexport, I think we should strive for a clean user API, in line with  https://github.com/scrapy/scrapy/pull/6273#discussion_r1618434374:

```python
FEEDS = {
    "postgres://user:password@server:port": {},
}
```

However, it might be worth considering implementing a separate extension (streamexport?) instead of trying to fit this into the feedexport extension. While there are features of feedexport that could be interesting to get for free for data streaming, like batching, feedexport may be too specific to files (exporters, post-processors), and there are many aspects of it that simply do not make sense for streaming data.

guillermo-bondonno:
I agree with a cleaner API, but I think a separate extension would be too similar. I still think that making users aware that custom exporters are possible is worth it (and I agree with the comments about the docs, will work on that). 

The issue with custom schemes and uris is that the user would have to point it to a custom or existing Storage, and if all we want to do is stream to a service there's no point in that.

I think a good solution is to allow a setting like `no_storage` or `stream_export` in the feed options so that:
- It doesn't check for supported storages
- It creates slots with a DummyStorage that doesn't interfere with the exporter

Also, we can pass the URI to the exporters `__init__` (handling it in `._configure()`), this won't make much difference and will make it accessible in case the exporter wants to use it (like `postgres://<user>:<password>@<host> ... etc).`

@VMRuiz @Gallaecio  what do you think? Is there an easier way to support arbitrary URIs

Gallaecio:
In order to go for the suggested clean API, wouldn’t it make more sense to make streaming work through a custom storage class? Maybe we could make it so that storage classes can enforce their own exporter.

guillermo-bondonno:
> In order to go for the suggested clean API, wouldn’t it make more sense to make streaming work through a custom storage class? Maybe we could make it so that storage classes can enforce their own exporter.

But in that case wouldn't we need to create custom storages AND exporters? I also think it's more appropriate naming-wise that the actual stream/requests are sent from the exporter and not the storage.

What I have in mind is something like 

```
"FEEDS" : {
            "somecustomscheme://": {
               "format": "StreamingAPI",
               "stream_export" : True
            }
        }
```
```
def _get_storage(
        self, uri: str, feed_options: Dict[str, Any]
    ) -> FeedStorageProtocol:
        """Fork of create_instance specific to feed storage classes

        It supports not passing the *feed_options* parameters to classes that
        do not support it, and issuing a deprecation warning instead.
        """
        use_dummy_storage = feed_options.get("stream_export", False)
        feedcls = (
            DummyStorage
            if use_dummy_storage
            else self.storages.get(urlparse(uri).scheme, self.storages["file"])
        )
        ...
```

Gallaecio:
> But in that case wouldn't we need to create custom storages AND exporters?

Not necessarily. You could enable the possibility to define a storage that does not need to use an exporter, and instead handles exporting directly.

> I also think it's more appropriate naming-wise that the actual stream/requests are sent from the exporter and not the storage.

If you have both, I guess. But you could either make it so that the storage class does not need to use an exporter, or that it can handle the same function as an exporter, or you implement a new concept, a “streamer”, and make that a component of the storage class, and storage classes that declare a “streamer” do not need an exporter.

I would not worry about the naming for now, and instead focus on what changes we need to make to the code to make it easy to provide the ideal user API for writing a streaming component for a database or similar.

PR #6274, opened 2024-03-06
Improve typing for Spider.parse().

wRAR:
Based on https://stackoverflow.com/questions/77160087/correct-typing-of-a-method-without-a-fixed-signature

Should remove most of the mypy errors for parse() in user spiders (if they use older mypy Spider.parse() will appear untyped so no override error either). Doesn't help with scrapy-poet's callbacks that narrow down `Response` to `DummyResponse`: https://github.com/scrapinghub/scrapy-poet/issues/179

This should probably have typing tests, I'll do that later.


scrapy/spiders/__init__.py
```diff
@@ -17,12 +17,17 @@
...
+
+    # typing.Concatenate requires Python 3.10
     # typing.Self requires Python 3.11
-    from typing_extensions import Self
+    from typing_extensions import Concatenate, Self
 
     from scrapy.crawler import Crawler
     from scrapy.settings import BaseSettings
 
+    CallbackT = Callable[Concatenate[Response, ...], Any]
```
wRAR, line 29: This could be useful in other places as it's a general type for a spider callback. I didn't think about that further though.

wRAR: Actually we should use it for `Request.callback`

wRAR: Looks like we can't, probably because of wrong covariance when assigning a specific callback to a request, I need to investigate this separately.

Gallaecio (approved):

PR #6279, opened 2024-03-08
Full typing for scrapy/extensions, part 2.

wRAR:


docs/topics/telnetconsole.rst
```diff
@@ -172,8 +172,8 @@ TELNETCONSOLE_PORT
 
 Default: ``[6023, 6073]``
 
-The port range to use for the telnet console. If set to ``None`` or ``0``, a
```
wRAR, line 175: `0` doesn't work with `getlist()`

scrapy/extensions/periodic_log.py
```diff
@@ -15,52 +25,59 @@ class PeriodicLog:
...
 
-        ext_timing_enabled = crawler.settings.getbool(
+        ext_timing_enabled: bool = crawler.settings.getbool(
             "PERIODIC_LOG_TIMING_ENABLED", False
         )
         if not (ext_stats or ext_delta or ext_timing_enabled):
             raise NotConfigured
+        assert crawler.stats
+        assert ext_stats is not None
+        assert ext_delta is not None
```
wRAR, line 80: The logic for these two settings is complicated because they can (per the docs) be `None` (which is equivalent to an empty dict after `getdict()` anyway), `True` or a dict and so can't be False. Setting one of them to False indeed breaks the code. So the code above that sets the vars to None if getbool() returned False is dead code that can be refactored.

scrapy/utils/reactor.py
```diff
@@ -2,26 +2,26 @@
...
+def listen_tcp(portrange: List[int], host: str, factory: ServerFactory) -> Port:  # type: ignore[return]
     """Like reactor.listenTCP but tries different ports in a range."""
     from twisted.internet import reactor
 
     if len(portrange) > 2:
         raise ValueError(f"invalid portrange: {portrange}")
     if not portrange:
         return reactor.listenTCP(0, factory, interface=host)
-    if not hasattr(portrange, "__iter__"):
-        return reactor.listenTCP(portrange, factory, interface=host)
```
wRAR, line 24: Dead code, this function went through a few rounds of fixing the logic but in any case we always pass a list to it in the only place we use it, and if somebody passed a single value to it it would explode since 2020 (or, with asserts enabled, since the beginning).

PR #6280, opened 2024-03-09
Update selectors-sample1.html

Runder55:
There was an error in an html label where we added a languaje specification at docs/static/selectorsSample1


wRAR:
I don't think this makes sense.

PR #6283, opened 2024-03-09
Update startproyect.py

Runder55:
To make it easier to change in a future


wRAR:
Please stop spamming.

Issue #6286, opened 2024-03-09
WindowsRunSpiderCommandTest isn't skipped properly in the pinned envs


wRAR:
https://github.com/scrapy/scrapy/commit/002513438204eea5062b5a1d75fb4f261880da4f was incorrect as, it turned out, top-level `unittest.skipIf` support for `twisted.trial.unittest` tests was added only in Twisted 21.2.0 and we install 18.9.0 in the pinned envs, but for some reason the test was still skipped. Today pytest 8.1.1 was released and the problem is now visible (I don't see anything related in the changelog but then I don't know why the test is skipped in 8.0.2 and earlier), so the test fails on pinned envs on Linux instead of being skipped.

We can revert the linked commit, raise the minimum supported Twisted version to 21.2.0 (it's also the first version officially supporting Python 3.8 which itself was released in Oct 2019, so a year later than Twisted 18.9.0), or add some more complicated code that skips the individual methods only for older Twisted.

PR #6290, opened 2024-03-11
Fix WindowsRunSpiderCommandTest skip outside Windows for older Twisted

Gallaecio:
Fixes https://github.com/scrapy/scrapy/issues/6286


wRAR:
Test failures are probably caused by a missing super() call?

wRAR (approved):

Issue #6292, opened 2024-03-15
Handle robots.txt files not utf-8 encoded


fkromer:
## Summary

`robots.txt` files which are not `utf-8` encoded make `scrapy` raise an `UnicodeDecodeError` atm.

```
.venv/lib/python3.11/site-packages/scrapy/robotstxt.py", line 15, in decode_robotstxt
    robotstxt_body = robotstxt_body.decode("utf-8")
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xdf in position 46: invalid continuation byte
```

## Motivation

Prevents from being able to manage all scrapers in an automated fashion, like described in "Describe alternatives you've considered". 

## Describe alternatives you've considered

Check `robots.txt` file manually for the scraper in question, disable `robots.txt` file parsing globally using `ROBOTSTXT_OBEY`. Prevents to being able to process all scrapers in an automated fashion however.

Gallaecio:
We can probably follow [Google’s approach](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt):

> Similarly, if the character encoding of the robots. txt file isn't UTF-8, Google may ignore characters that are not part of the UTF-8 range, potentially rendering robots. txt rules invalid.

So fixing this requires changing the offending like to:

```python
robotstxt_body.decode("utf-8", errors="ignore")
```

And implementing a test for it.

HenrySchwerdt:
@Gallaecio I would like to work on this.

Gallaecio:
Please, feel free to open a PR. Thanks!

HenrySchwerdt:
I'll do it tonight

PR #6298, opened 2024-03-26
Handle robots.txt files not UTF-8 encoded

lorenzoverardo:
Fixes https://github.com/scrapy/scrapy/issues/6292.

cc @Gallaecio

### DIFF
--- a/tests/test_robotstxt_interface.py
+++ b/tests/test_robotstxt_interface.py
@@ -1,5 +1,7 @@
 from twisted.trial import unittest
 
+from scrapy.robotstxt import decode_robotstxt
+
 
 def reppy_available():
     # check if reppy parser is installed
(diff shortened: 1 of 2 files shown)


tests/test_robotstxt_interface.py
```diff
@@ -141,6 +143,25 @@ def test_unicode_url_and_useragent(self):
...
+
+    def test_decode_utf8(self):
+        robotstxt_body = "User-agent: *\nDisallow: /\n".encode("utf-8")
+        decoded_content = decode_robotstxt(robotstxt_body, spider=None)
+        self.assertEqual(decoded_content, "User-agent: *\nDisallow: /\n")
+
+    def test_decode_non_utf8(self):
+        robotstxt_body = b"User-agent: *\n\xFFDisallow: /\n"
+        decoded_content = decode_robotstxt(robotstxt_body, spider=None)
+        self.assertNotEqual(decoded_content, "")
```
wRAR, line 162: Why not use `assertEqual` with the real returned value here?

lorenzoverardo: The idea behind this test case is that when a string is not in UTF-8 we can't reliably predict what the output will be as a result of adding `errors="ignore"` when decoding.
As it stands, we do obviously know what `robotstxt_body` will be and we might as well just use `assertEqual`, however I was trying to make sure that this code wouldn't fall under the [case where a `UnicodeDecodeError` is thrown](https://github.com/scrapy/scrapy/pull/6298/files#diff-0dc4cc282c7f78b4d6199a7b5de162f382efdf5d4836bb9fd567be94d15c8637R26-R36) - which just returns `""` - and I suppose we're trying to avoid that with this change?
Let me know what you think is best 🙂

Gallaecio: I also think we should use `assertEqual` with the actual value.

> was trying to make sure that this code wouldn't fall under the [case where a `UnicodeDecodeError` is thrown](https://github.com/scrapy/scrapy/pull/6298/files#diff-0dc4cc282c7f78b4d6199a7b5de162f382efdf5d4836bb9fd567be94d15c8637R26-R36)

`assertEqual` with the actual value works for this as well, as long as the actual value is not `""`, i.e. `self.assertEqual(decoded_content, "foo")` implies `self.assertNotEqual(decoded_content, "")`, but is better than the latter because it also proves we are using `ignore` and not some other error handling (e.g. replacing errors with special characters or something).

lorenzoverardo: Got it, will change it to `assertEqual` then 👍

Gallaecio (approved):

PR #6306, opened 2024-04-03
issue#6305 Replace deprecated ast.NameConstant usage and add tests

### LINKED ISSUE #6305: ast.NameConstant is deprecated and will be removed in Python 3.14; use ast.Constant instead
```
scrapy/utils/misc.py:249: DeprecationWarning: ast.NameConstant is deprecated and will be removed in Python 3.14; use ast.Constant instead
    value is None or isinstance(value, ast.NameConstant) and value.value is None
```

Also, if I understand it correctly `isinstance(value, ast.NameConstant)` is always False, as `ast.NameConstant()` returns an `ast.Constant` instance, so I wonder if this case is covered by a test.

lizdenhup:
## Problem

Within `misc.py` there is a usage of `ast.NameConstant`. `ast.NameConstant` will be deprecated beginning with Python 3.14.x so we want to remove it.

## Solution
- Replace `ast.NameConstant` with `ast.Constant` within `misc.py`
- Add tests to test situations where the generator returns an `ast.Constant` with a value, or with a `None` value

## Validation
- I added tests. The tests pass.

Fixes #6305

### DIFF
--- a/tests/test_utils_misc/test_return_with_argument_inside_generator.py
+++ b/tests/test_utils_misc/test_return_with_argument_inside_generator.py
@@ -12,6 +12,13 @@ from scrapy.utils.misc import (
 def _indentation_error(*args, **kwargs):
     raise IndentationError()
 
+def create_ast_return(value=None):
+    return (
+        unittest.mock.MagicMock(name='ast.Return', value=None)
+        if value is None
+        else unittest.mock.MagicMock(name='ast.Return',
+                                     value=unittest.mock.MagicMock(name='ast.Constant', value=value))
+    )
 
 def top_level_return_something():
     """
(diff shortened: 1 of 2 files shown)


tests/test_utils_misc/test_return_with_argument_inside_generator.py
```diff
@@ -2,6 +2,7 @@
 import warnings
 from functools import partial
 from unittest import mock
+import ast
```
wRAR, line 5: It shouldn't be used in tests directly.

lizdenhup: Hiya! Good call out -- I amended my commit to use magic mock. Let me know what you think. Cheers.

tests/test_utils_misc/test_return_with_argument_inside_generator.py
```diff
@@ -12,6 +12,13 @@
 def _indentation_error(*args, **kwargs):
     raise IndentationError()
 
+def create_ast_return(value=None):
+    return (
+        unittest.mock.MagicMock(name='ast.Return', value=None)
```
wRAR, line 17: Hmm my point is that I expect this test to test real code that produces specific ast nodes when parsed.

wRAR: (hpefully that makes sense)

lizdenhup: Yes, that makes sense. I will push a new change tomorrow. Thank you for the feedback!

PR #1806, opened 2016-02-23
[WIP] Handle 201 responses in pipelines.files 

darshanime:
Added a check for `location` field in `201` responses. The `location` field is to have the link for the resource created due to the request.
Fixes #1615

Issue #1615, opened 2015-11-24
File not downloading when response.status is 201


kurkop:
I am crawling a web that when call a image it is created on the fly, it returns 201 (create) in response.status, but this is not saved because response.status != 200.

https://github.com/scrapy/scrapy/blob/75cd056223a5a8da87a361aee42a541afcf27553/scrapy/pipelines/files.py#L243

Do you think that it is a bug?

Slepice1:
I bumped into same issue and according to this http://racksburg.com/choosing-an-http-status-code/ I think it probably should save.

redapple:
I do believe it's a bug.

darshanime:
Hey @redapple!
I would like to take a stab at the bug. 
I went thru the source and noticed that for each item, we are creating a `DeferredList` and adding callbacks to it. What I think we can do is, in [`media_downloaded`](https://github.com/scrapy/scrapy/blob/75cd056223a5a8da87a361aee42a541afcf27553/scrapy/pipelines/files.py#L243) we can check if the status code is 201. If it is, we can create a new `deferred` object there using the URL in the `Location` attribute of the 201 response's headers. We can attach the exact same callbacks as for the `DeferredList`.

This approach has one downside though, we will cache the wrong result for this request and our stats will not be accurate as we will log this particular item as `failed` when in fact, we may get the file downloaded in the redirect link.

What do you think of these ideas? I can issue a [WIP] PR once you approve this approach.

redapple:
@darshanime , thanks for looking into it!
I've never seen HTTP 201 responses myself so I'm not sure how this all works.
But if the response headers have a `Location`, one might want to mimic what `RedirectMiddleware` does for 30x's
@kurkop , @Slepice1 , @darshanime ,
do you know of any public website behaving like this so it could be tested live?

kurkop:
Yes @redapple ,

You can testing in the next url: http://www.carroya.com/web/vehiculo/nuevoficha/renault/kangoo/c.a./2016/no-disponible/1616709.do

![Firebug output](https://cloud.githubusercontent.com/assets/2367831/13178243/c1c43d6a-d6eb-11e5-9066-83520d97348e.png)

cdz620:
I think it would better the download file result status is configured with settings.py.

redapple:
Hey @kurkop , I'm resuming this one.
Do you happen to have another website doing this HTTP 201?
http://www.carroya.com seems to be returning HTTP 200 on images now.

[...]

PR #6309, opened 2024-04-07
Edits to media_downloaded in files.py to handle 201 response status (#1615 and #1806)

zoemyatt:
#1615 and #1806 discussed the importance of handling a 201 response status in `pipelines/files.py`. We have attempted to correct this error by not only checking for 200 response status but also checking for 201 response status and location headers in the newly created image. If the new image has a 201 status and a location header, the new program schedules an extra request for the file data. 

We then added a unit test to scrape a URL we knew would return a 201 status code and ran tox with the `test_pipeline_media.py` file.


tests/keys/mitmproxy-dhparam.pem
wRAR, line 1: Extra file

tests/test_pipeline_media.py
```diff
@@ -67,6 +67,11 @@ def test_default_media_downloaded(self):
         response = Response("http://url", body=b"")
         assert self.pipe.media_downloaded(response, request, self.info) is response
 
+    def test_media_downloaded_201(self):
+        request = Request("http://p2.so.qhimgs1.com/t013690341ce9bfc61e.jpg")
```
wRAR, line 71: Tests shouldn't depend on real websites.

zoemyatt:
@wRAR let me know if these changes look better. Sorry about the extra file in there!

We improved the test case to check for a 201 status code without using a real URL and deleted the extra file.

Gallaecio:
Hey, [you might want to install pre-commit](https://docs.scrapy.org/en/latest/contributing.html#pre-commit).

Issue #6308, opened 2024-04-06
Different behavior of `get_header` between urllib.Request and WrappedRequest


marinelay:
### Description

I believe `WrappedRequest` is a class for supporting methods in `urllib.Request`, but I found a behavior of `get_header` method in `WrappedRequest` is different from `urllib.Requests`.
When trying to get a header which is not present without default value, it should return `None` value (https://docs.python.org/3/library/urllib.request.html#urllib.request.Request.get_header), 
but `get_header` in `WrappedRequest` raises `TypeError: to_unicode must receive a bytes or str object, got NoneType`.

### Steps to Reproduce

```
from urllib.request import Request as _Request
from scrapy.http.request import Request
from scrapy.http.cookies import WrappedRequest

a = _Request(url="https://a.example")
print(a.get_header('xxxx'))

b = WrappedRequest(Request(url="https://a.example"))
print(b.get_header('xxxx'))
```

**Expected behavior:** 

```
None
None
```

**Actual behavior:**

```
None
TypeError: to_unicode must receive a bytes or str object, got NoneType
```

### Additional context

This issue is currently not a problem when interacting with the `CookieJar` class.
Thus, it is not an urgent matter unless one is importing and using `WrappedRequest`. 
However, I think it would enhance the reliability in this project.
Thank you!

siddharth07-ui:
Hi @marinelay and @wRAR, while looking through this good first issue, I found the line in scrapy, which seems to be the root cause of providing this error for Request with WrappedRequest as shown in the line --

=====================================================================
from urllib.request import Request as _Request
from scrapy.http.request import Request
from scrapy.http.cookies import WrappedRequest

a = _Request(url="https://a.example")
print(a.get_header('xxxx'))

**b = WrappedRequest(Request(url="https://a.example"))**    ------    This one
print(b.get_header('xxxx'))
=====================================================================
None
**TypeError: to_unicode must receive a bytes or str object, got NoneType**    ------ This one
=====================================================================

Where as per behavior, it should be returning None for both the requests.

This is line 173 in "http/response/cookies.py" --

**return to_unicode(self.request.headers.get(name, default), errors="replace")**

Because of "to_unicode" being used, which as per function definition says --
 **"""Return the unicode representation of a bytes object ``text``. If
    ``text`` is already an unicode object, return it as-is."""**

[...]

PR #6310, opened 2024-04-08
Fix WrappedRequest.get_header raising TypeError if default is None

VMRuiz:
Fixes https://github.com/scrapy/scrapy/issues/6308 

Alternatives: 

```python
    def get_header(self, name, default=None):
        value = self.request.headers.get(name, default)
        if not value:
            return value
        return to_unicode(value, errors="replace")
 ```
 
 But I don't like repeating value so much 
 
 ```python
    def get_header(self, name, default=None):
        if not (value := self.request.headers.get(name, default)):
            return value
        return to_unicode(value, errors="replace")
 ```
 The parenthesis and walrus operator make the line too convoluted
 
 ```python
     def get_header(self, name, default=None):
         try: 
             to_unicode(self.request.headers.get(name, default), errors="replace")
         except TypeError:
             return default
 ```
Using try/except this way is pythonic, but I'm always afraid I may unexpectedly capture another unrelated exception

### DIFF
--- a/scrapy/http/cookies.py
+++ b/scrapy/http/cookies.py
@@ -170,7 +170,8 @@ class WrappedRequest:
         return name in self.request.headers
 
     def get_header(self, name, default=None):
-        return to_unicode(self.request.headers.get(name, default), errors="replace")
+        value = self.request.headers.get(name, default)
+        return to_unicode(value, errors="replace") if value else value
 
     def header_items(self):
         return [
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scrapy/http/cookies.py [hunk 1]
```diff
@@ -170,7 +170,8 @@ def has_header(self, name):
         return name in self.request.headers
 
     def get_header(self, name, default=None):
-        return to_unicode(self.request.headers.get(name, default), errors="replace")
+        value = self.request.headers.get(name, default)
+        return to_unicode(value, errors="replace") if value else value
```
wRAR, line 174: I think it will return `''` where we want to return `b''` :)

VMRuiz: You are right, I fix it and add a test

PR #6324, opened 2024-04-27
issue #6323: add SpiderLoggerAdapter, change Spider.logger to return SpiderLoggerAdapter

### LINKED ISSUE #6323: Spider.logger not logging custom extra information
I noticed the implicit behavior of the Spider.logger: when logging with extra, extra ultimately do not end up in the log because they are overwritten by default `process` method of [LoggerAdapter](https://github.com/scrapy/scrapy/blob/master/scrapy/spiders/__init__.py#L47)

Current logic:
```py
>>> self.logger.info("test log", extra={"test": "very important information"})
{"message": "test log", "spider": "spider_name"}
```

Expected logic:
```py
>>> self.logger.info("test log", extra={"test": "very important information"})
{"message": "test log", "spider": "spider_name", "test": "very important information"}
```

bloodforcream:
fixes #6323

SpiderLoggerAdapter prioritizes self.extra over logged extra so it works as it worked before.
`spider` in extra will always be `self`


scrapy/utils/spider_logger_adapter.py
wRAR, line 1: This should go into `scrapy.utils.log`

wRAR:
Is it possible to add a test that checks that the Spider logging methods work as expected?

scrapy/utils/spider_logger_adapter.py
bloodforcream: I agree. That's where I tried to put it first but I get a `circular import` error.

```
..\scrapy\__init__.py:15: in <module>
..\scrapy\spiders\__init__.py:16: in <module>
    from scrapy.utils.log import SpiderLoggerAdapter
..\scrapy\utils\log.py:13: in <module>
..\scrapy\settings\__init__.py:23: in <module>
    from scrapy.settings import default_settings
..\scrapy\settings\default_settings.py:322: in <module>
    USER_AGENT = f'Scrapy/{import_module("scrapy").__version__} (+https://scrapy.org)'
E   AttributeError: partially initialized module 'scrapy' has no attribute '__version__' (most likely due to a circular import)
```

Maybe i'm wrong and it's solvable

bloodforcream:
> Is it possible to add a test that checks that the Spider logging methods work as expected?

Do you mean a test that's making sure general logging messages(not "extra") of Spider work as expected?
I'll try to code them today/tomorrow

scrapy/spiders/__init__.py
```diff
@@ -42,9 +43,9 @@ def __init__(self, name: Optional[str] = None, **kwargs: Any):
             self.start_urls: List[str] = []
 
     @property
-    def logger(self) -> logging.LoggerAdapter:
+    def logger(self) -> SpiderLoggerAdapter:
         logger = logging.getLogger(self.name)
-        return logging.LoggerAdapter(logger, {"spider": self})
+        return SpiderLoggerAdapter(logger, {"spider": self})
```
bloodforcream, line 48:
```suggestion
        return SpiderLoggerAdapter(logger, {"spider": self.name})
```
Maybe also change `self` to `self.name`? So we get actual spider name in logs instead of `<Spider 'spider_name' at 0x1b2fc089580>`

bloodforcream:
Added more tests

scrapy/spiders/__init__.py
wRAR: > Probably not the best idea since it's gonna change current behavior

Yeah

scrapy/utils/spider_logger_adapter.py
wRAR: I still don't like having a separate module for this, I see these options:

1. Move it to scrapy.spiders (probably no technical downsides as it's not useful anywhere else?).
2. Remove the SpiderLoggerAdapter import from the scrapy.spiders top-level, moving it under `if TYPE_CHECKING` and inside `def logger()`.
3. (likely a worse idea) Change some other imports in the loop.

I would go with 2 in general but probably with 1 in this case.

@Gallaecio second opinion?

bloodforcream: Done the second option

bloodforcream: To be honest importing SpiderLoggerAdapter on every single log call looks a bit inefficient, imho. 

But it's obviously up to you guys

wRAR (approved):
Thanks!

PR #6332, opened 2024-05-04
Add support for multipart/form-data to FormRequest

### LINKED ISSUE #1897: support for  "multipart/form-data" in scrapy.FormRequest

alexandresgf:
Fixes #1897, closes #1954

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -799,7 +799,7 @@ fields with form data from :class:`Response` objects.
 
 .. class:: scrapy.http.request.form.FormRequest
 .. class:: scrapy.http.FormRequest
-.. class:: scrapy.FormRequest(url, [formdata, ...])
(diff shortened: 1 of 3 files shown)


wRAR:
Doesn't look like this changes anything except Content-Type, is this correct?

alexandresgf:
> Doesn't look like this changes anything except Content-Type, is this correct?

Right, as suggested by the OP in the ticket, it should be a flag, in this case `is_multipart`, which changes the `Content-Type` to `multipart/form-data`. I also added a new parameter to `FormRequest` constructor.

wRAR:
The format of multipart/form-data requests is very different from the application/x-www-form-urlencoded one, see e.g. https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4 or the previous PR on this.

alexandresgf:
Closing this PR because I'll move the code to the library [form2request](https://github.com/scrapy/form2request)

PR #6333, opened 2024-05-04
Full typing for scrapy/*.py

wRAR:


scrapy/squeues.py
```diff
@@ -2,20 +2,28 @@
...
+if TYPE_CHECKING:
+    # typing.Self requires Python 3.11
+    from typing_extensions import Self
+
 
-def _with_mkdir(queue_class):
-    class DirectoriesCreated(queue_class):
-        def __init__(self, path: Union[str, PathLike], *args, **kwargs):
+def _with_mkdir(queue_class: Type[queue.BaseQueue]) -> Type[queue.BaseQueue]:
+    class DirectoriesCreated(queue_class):  # type: ignore[valid-type,misc]
```
wRAR, line 25: Still not sure how to do this correctly, and whether this is a deficiency in mypy, something the hints can't express or just wrong hints.

scrapy/squeues.py
```diff
@@ -122,23 +142,24 @@ def _pickle_serialize(obj):
         raise ValueError(str(e)) from e
 
 
+# queue.*Queue aren't subclasses of queue.BaseQueue
```
wRAR, line 145: I think `queuelib.queue.BaseQueue` needs to be a Protocol, not an ABC with `__subclasscheck__`.

scrapy/core/downloader/__init__.py
```diff
@@ -143,7 +143,7 @@ def _get_slot(self, request: Request, spider: Spider) -> Tuple[str, Slot]:
 
         return key, self.slots[key]
 
-    def _get_slot_key(self, request: Request, spider: Spider) -> str:
+    def _get_slot_key(self, request: Request, spider: Any) -> str:
```
Gallaecio, line 146: Not `Optional[Spider]`? It this because of tests passing junk, or something else?

Gallaecio (approved):

scrapy/core/downloader/__init__.py
wRAR: It doesn't really matter as the argument is not used. We should deprecate and remove it I think, we can't remove it right away because it's used both in private and public code (e.g. in `ScrapyZyteAPIDownloaderMiddleware`)

wRAR: https://github.com/scrapy/scrapy/issues/6340

PR #6334, opened 2024-05-04
Closes #6328. Document 'json' selector type

### LINKED ISSUE #6328: 'json' selector type not documented
https://docs.scrapy.org/en/latest/topics/selectors.html

`type defines the selector type, it can be "html", "xml" or None (default).`

But my tests are now failing because it's returning `json` type for actual json docs now.

kumar-sanchay:
added json as one of the value for type argument.
Closes #6328

### DIFF
--- a/scrapy/selector/unified.py
+++ b/scrapy/selector/unified.py
@@ -47,7 +47,7 @@ class Selector(_ParselSelector, object_ref):
     ``response`` isn't available. Using ``text`` and ``response`` together is
     undefined behavior.
 
-    ``type`` defines the selector type, it can be ``"html"``, ``"xml"``
+    ``type`` defines the selector type, it can be ``"html"``, ``"xml"``, ``"json"``
     or ``None`` (default).
 
     If ``type`` is ``None``, the selector automatically chooses the best type


Gallaecio:
Thanks!

mohmad-null:
This doesn't seem to fix the issue entirely. From the current unified.py:

```
    ``type`` defines the selector type, it can be ``"html"``, ``"xml"``, ``"json"``
    or ``None`` (default).

    If ``type`` is ``None``, the selector automatically chooses the best type
    based on ``response`` type (see below), or defaults to ``"html"`` in case it
    is used together with ``text``.

    If ``type`` is ``None`` and a ``response`` is passed, the selector type is
    inferred from the response type as follows:

    * ``"html"`` for :class:`~scrapy.http.HtmlResponse` type
    * ``"xml"`` for :class:`~scrapy.http.XmlResponse` type
    * ``"html"`` for anything else

    Otherwise, if ``type`` is set, the selector type will be forced and no
    detection will occur.
```

The second half of that also doesn't mention the `json` type, including not saying how it's inferred.

PR #6341, opened 2024-05-06
More typing for scrapy/core/downloader

wRAR:


scrapy/core/downloader/handlers/http11.py
```diff
@@ -123,83 +127,88 @@ class TunnelingTCP4ClientEndpoint(TCP4ClientEndpoint):
...
         )
         protocol.transport.write(tunnelReq)
         self._protocolDataReceived = protocol.dataReceived
-        protocol.dataReceived = self.processProxyResponse
+        protocol.dataReceived = self.processProxyResponse  # type: ignore[method-assign]
         self._protocol = protocol
         return protocol
 
-    def processProxyResponse(self, rcvd_bytes):
+    def processProxyResponse(self, data: bytes) -> None:
```
Gallaecio, line 158: Technically backward incompatible, but :see_no_evil:

scrapy/core/downloader/handlers/http11.py
```diff
@@ -230,18 +239,20 @@ class TunnelingAgent(Agent):
...
-        pool=None,
+        reactor: ReactorBase,
+        proxyConf: Tuple[str, int, Optional[bytes]],
+        contextFactory: Optional[IPolicyForHTTPS] = None,
+        connectTimeout: Optional[float] = None,
+        bindAddress: Optional[bytes] = None,
+        pool: Optional[HTTPConnectionPool] = None,
     ):
+        # TODO make this arg required instead
+        assert contextFactory is not None
```
Gallaecio, line 250: To do in this PR, or should we create an issue to track this?

Gallaecio (approved):

scrapy/core/downloader/handlers/http11.py
wRAR: Not here, I'll make an issue for both instances. Not sure how to properly do this though. We could make all args kw-only.

wRAR: https://github.com/scrapy/scrapy/issues/6343

scrapy/core/downloader/handlers/http11.py
wRAR: I assume it's only called when assigned to `protocol.dataReceived` so there shouldn't be problems even theoretically.

PR #6349, opened 2024-05-08
Closes #6343. Make certain args of ScrapyAgent and TunnelingAgent required

### LINKED ISSUE #6343: Make certain args of `ScrapyAgent` and `TunnelingAgent` required
`scrapy.core.downloader.handlers.http11.ScrapyAgent` and `scrapy.core.downloader.handlers.http11.TunnelingAgent` take some args (`contextFactory` in both and `crawler` in `ScrapyAgent`) that are currently declared optional but in fact need to be passed as the code assumes they aren't None. So ideally they should be declared as required (but that requires reordering args and/or making all args of those constructors keyword-only so there is no clear and simple fix).

kumar-sanchay:
Fixes #6343

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -239,15 +239,14 @@ class TunnelingAgent(Agent):
 
     def __init__(
         self,
+        *,
         reactor: ReactorBase,
         proxyConf: Tuple[str, int, Optional[bytes]],
-        contextFactory: Optional[IPolicyForHTTPS] = None,
+        contextFactory: IPolicyForHTTPS,
         connectTimeout: Optional[float] = None,
         bindAddress: Optional[bytes] = None,
         pool: Optional[HTTPConnectionPool] = None,
     ):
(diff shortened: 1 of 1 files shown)


wRAR:
The asserts are no longer needed.

kumar-sanchay:
> The asserts are no longer needed.

Ok, i was thinking what if we pass None to those keyword arguments??

kumar-sanchay:
> The asserts are no longer needed.

Done

wRAR:
The asserts were there for typing purposes, to show that the args can't be None normally. If None is passed there will be a normal TypeError or something like that.

kumar-sanchay:
@Gallaecio @wRAR let me know if anything got missed here :)

wRAR:
As I said in the issue this is technically backwards-incompatible but only breaks the code that calls the constructors with all these args without using their names, which is unlikely (I took a quick look at the GH code search and didn't find such code).

wRAR (approved):

Gallaecio:
Thanks!

PR #6352, opened 2024-05-09
Closes #6340. Deprecate the spider argument to Downloader._get_slot_key()

### LINKED ISSUE #6340: Deprecate the `spider` argument to `Downloader._get_slot_key()`
The `spider` argument is not used inside the method since 2012, but we can't remove it as external code calls it (either because it subclasses `Downloader` or because it wants the slot name).

Actually maybe we want to promote it to a public method for the second reason? Not sure how often it's needed but see e.g. https://github.com/scrapy-plugins/scrapy-zyte-api/blob/a2284c8cdf157ef6d36c1cc413933761c5ed792b/scrapy_zyte_api/_middlewares.py#L32

kumar-sanchay:
Fix #6340

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -4,10 +4,10 @@ Release notes
 =============
 
 
-.. _release-2.11.2:
+.. _release-VERSION:
 
-Scrapy 2.11.2
---------------------------
+Scrapy VERSION (YYYY-MM-DD)
+---------------------------
 
 Deprecations
 ~~~~~~~~~~~~
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,20 @@
 Release notes
 =============
 
+
+.. _release-2.11.2:
+
+Scrapy 2.11.2
+--------------------------
+
+Deprecations
+~~~~~~~~~~~~
+
+-   :func:`scrapy.core.downloader.Downloader._get_slot_key` is now deprecated.
+    Consider using its corresponding public method get_slot_key() instead.
+    (:issue:`6340`)
+
+
 .. _release-2.11.1:
 
 Scrapy 2.11.1 (2024-02-14)
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -1,4 +1,5 @@
 import random
+import warnings
 from collections import deque
 from datetime import datetime
 from time import time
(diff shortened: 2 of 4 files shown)


docs/news.rst
```diff
@@ -1262,6 +1262,9 @@ Deprecations
 -   :func:`scrapy.utils.response.response_httprepr` is now deprecated.
     (:issue:`4972`)
 
+-   :func:`scrapy.core.downloader.Downloader._get_slot_key` is now deprecated. Consider using its corresponding public method instead.
```
wRAR, line 1265: Can you please wrap the line to 80 chars?

scrapy/core/downloader/__init__.py
```diff
@@ -153,6 +155,14 @@ def _get_slot_key(self, request: Request, spider: Optional[Spider]) -> str:
 
         return key
 
+    def _get_slot_key(self, request: Request, spider: Optional[Spider]) -> str:
+        warnings.warn(
+            "Use of this protected method is deprecated. Consider using its corresponding public method instead.",
```
wRAR, line 160: I'd mention the new name explicitly.

wRAR (approved):
Looks good with some small changes needed.

kumar-sanchay:
@wRAR done with the changes you suggested

docs/news.rst
```diff
@@ -1262,6 +1262,10 @@ Deprecations
 -   :func:`scrapy.utils.response.response_httprepr` is now deprecated.
     (:issue:`4972`)
 
+-   :func:`scrapy.core.downloader.Downloader._get_slot_key` is now deprecated.
+    Consider using its corresponding public method get_slot_key() instead.
+    (:issue:`6340`)
+
```
Gallaecio, line 1268: This is a section for an old release. To cover this new change, you should either add it at the very beginning of the file, in a new section for an unreleased version, or do nothing, and hope we don’t forget to cover this when the build the release notes. The first approach is better, but the second approach is the usual and fine as well.

Gallaecio (commented):
It would be nice to have a test that confirms that the old method continues to work as usual but with a warning. But :+1: from me after the release notes issue is handled.

docs/news.rst
```diff
@@ -3,6 +3,20 @@
 Release notes
 =============
 
+
+.. _release-2.11.2:
+
+Scrapy 2.11.2
```
kumar-sanchay, line 9: I've not added release date here assuming this will be taken care during release

docs/news.rst
kumar-sanchay: Ahh ! I see, thank you for pointing this out

kumar-sanchay:
> It would be nice to have a test that confirms that the old method continues to work as usual but with a warning. But 👍 from me after the release notes issue is handled.

Do you believe testing would be beneficial in this case, considering that the deprecated method internally returning new public method ?

Gallaecio:
> Do you believe testing would be beneficial in this case, considering that the deprecated method internally returning new public method ?

You are right that here it would not add much, I’m OK with not adding it.

Gallaecio:
Thank you!

PR #6353, opened 2024-05-11
Use ParamSpec for callables.

wRAR:
I also changed most of `addCallbacks()` to `addCallback()`/`addErrback()`, as the official docs say the former has worse type hints (the ones I kept need to stay for various reasons).

I plan to add parameter specifications to more generics as separate PRs.

### DIFF
--- a/scrapy/utils/defer.py
+++ b/scrapy/utils/defer.py
@@ -2,21 +2,22 @@
 Helper functions for dealing with Twisted deferreds
 """
 
+from __future__ import annotations
+
 import asyncio
 import inspect
 from asyncio import Future
 from functools import wraps
 from types import CoroutineType
 from typing import (
+    TYPE_CHECKING,
     Any,
-    AsyncGenerator,
     AsyncIterable,
     AsyncIterator,
     Awaitable,
     Callable,
     Coroutine,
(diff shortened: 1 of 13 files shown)


scrapy/utils/defer.py
```diff
@@ -74,7 +83,31 @@ def defer_result(result: Any) -> Deferred:
...
+def mustbe_deferred(
+    f: Callable[_P, _T], *args: _P.args, **kw: _P.kwargs
+) -> Deferred[_T]: ...
+
+
+def mustbe_deferred(
+    f: Callable[_P, Union[Deferred[_T], Coroutine[Deferred[Any], Any, _T], _T]],
+    *args: _P.args,
+    **kw: _P.kwargs,
+) -> Deferred[_T]:
```
wRAR, line 110: This is copied from `maybeDeferred`

Gallaecio (approved):

Gallaecio:
> I also changed most of `addCallbacks()` to `addCallback()`/`addErrback()`, as the official docs say the former has worse type hints (the ones I kept need to stay for various reasons).

And in all cases readability improved, in my opinion. Nice!

PR #6367, opened 2024-05-16
Update expectations of cookies after redirects.

wRAR:


wRAR:
@Gallaecio is this correct?

PR #6368, opened 2024-05-18
Closes #6365. Fix overridable methods in MediaPipeline

### LINKED ISSUE #6365: Fix overridable methods in MediaPipeline
`MediaPipeline` defines several empty or almost empty "overridable" methods, which return things inconsistent with their overrides. I propose making all of them raise `NotImplementedError`. Alternatively `MediaPipeline` should just be made an abstract class and all those methods made abstract methods, but I have no idea if that will break anything (e.g. do all children always override all of those methods?).

Another problem is existing tests, that test specifically that e.g. `MediaPipeline.media_downloaded()` returns a response, which makes no sense to me (normally `media_downloaded()` returns a file info dict), so all those need to be changed or removed.

And another problem, indirectly related to this, is that this interface is very poorly documented, most of these functions are not mentioned in the docs at all, so it's not always clear what should they take and return (and the code uses many of them as callbacks in long callback chains so it's not clear even from the code).

kumar-sanchay:
Closes #6365


tests/test_pipeline_media.py
```diff
@@ -175,258 +149,16 @@ def test_should_remove_req_res_references_before_caching_the_results(self):
...
-        return super().media_failed(failure, request, info)
-
-    def item_completed(self, results, item, info):
-        self._mockcalled.append("item_completed")
-        item = super().item_completed(results, item, info)
-        item["results"] = results
-        return item
-
-
-class MediaPipelineTestCase(BaseMediaPipelineTestCase):
```
kumar-sanchay, line 213: These removed tests were highly dependent on override method. Let me know if you have other suggestions to handle this

scrapy/pipelines/media.py
```diff
@@ -197,35 +197,32 @@ def _cache_result_and_execute_waiters(self, result, fp, info):
...
-        if self.LOG_FAILED_RESULTS:
-            for ok, value in results:
-                if not ok:
-                    logger.error(
-                        "%(class)s found errors processing %(item)s",
-                        {"class": self.__class__.__name__, "item": item},
-                        exc_info=failure_to_exc_info(value),
-                        extra={"spider": info.spider},
-                    )
-        return item
```
wRAR, line 227: I think this (and the tests for it) should stay, in case some 3rd-party subclasses call super() for it (as bundled ones don't).

tests/test_pipeline_media.py
```diff
@@ -176,257 +143,8 @@ def test_should_remove_req_res_references_before_caching_the_results(self):
...
```
wRAR, line 213: On the first glance this, or maybe parts of this, looks like tests for the `MediaPipeline` machinery so they need to stay (unless the same tests are also in some other class/file).

kumar-sanchay: @wRAR I’ve kept required tests. Let me know if im missing something

tests/test_pipeline_media.py
```diff
@@ -35,8 +31,26 @@ def _mocked_download_func(request, info):
     return response() if callable(response) else response
 
 
+class UserDefinedPipeline(MediaPipeline):
```
wRAR, line 34: I wouldn't reuse the old MediaPipeline implementations here but followed the expected return values. So:

* media_downloaded() should return {} (technically, that's also wrong but better than returning `response`)
* file_path() should return ''

Others I think are fine.

kumar-sanchay: done

tests/test_pipeline_media.py
```diff
@@ -54,54 +68,6 @@ def tearDown(self):
...
-        request = Request("http://url")
-        response = Response("http://url", body=b"")
-        assert self.pipe.media_downloaded(response, request, self.info) is response
-
-    def test_default_media_failed(self):
-        request = Request("http://url")
-        fail = Failure(Exception())
-        assert self.pipe.media_failed(fail, request, self.info) is fail
-
-    def test_default_item_completed(self):
```
wRAR, line 75: This is also useful and should be kept.

kumar-sanchay: Ah yes you are right ! 
Still trying to understand the project :)

wRAR (approved):
Thanks!

Gallaecio (approved):

PR #6371, opened 2024-05-19
Full typing for scrapy/contracts.

wRAR:


scrapy/contracts/__init__.py
```diff
@@ -96,21 +112,26 @@ def tested_methods_from_spidercls(self, spidercls):
...
+    def extract_contracts(self, method: Callable) -> List[Contract]:
+        contracts: List[Contract] = []
+        assert method.__doc__ is not None
         for line in method.__doc__.split("\n"):
             line = line.strip()
 
             if line.startswith("@"):
-                name, args = re.match(r"@(\w+)\s*(.*)", line).groups()
+                m = re.match(r"@(\w+)\s*(.*)", line)
+                assert m is not None
```
Gallaecio, line 123: It could be `None`, though. Should we open a ticket to update it, e.g. to raise an exception about an invalid contract line being found?

Gallaecio (approved):

scrapy/contracts/__init__.py
wRAR: #6383

suvanbanerjee: @wRAR @Gallaecio Should we raise an exception or just pass the unmatched regex using `continue` ?

wRAR: As I said in #6383, "I think we should just skip lines that don't match."

PR #6372, opened 2024-05-20
Move CI from the decommissioned macos-11 to macos-latest.

wRAR:


wRAR:
https://github.com/wRAR/scrapy/actions/runs/9163227979

(3.8 was failing before too)

PR #6374, opened 2024-05-22
docs: Remove top-level reactor imports from CrawlerProces/CrawlerRunner examples

### LINKED ISSUE #6361: Remove top-level reactor imports from CrawlerProces/CrawlerRunner examples
There are several code examples on https://docs.scrapy.org/en/latest/topics/practices.html that have a top-level `from twisted.internet import reactor`, which is problematic (breaks when the settings specify a non-default reactor) and needs to be fixed.

Laerte:
fix #6361

### DIFF
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -134,6 +134,7 @@ Same example but using a non-default reactor, it's only necessary call
     install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
     runner = CrawlerRunner()
     d = runner.crawl(MySpider)
+
     from twisted.internet import reactor
 
     d.addBoth(lambda _: reactor.stop())
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -202,6 +202,7 @@ Same example using :class:`~scrapy.crawler.CrawlerRunner`:
     runner.crawl(MySpider1)
     runner.crawl(MySpider2)
     d = runner.join()
+
     from twisted.internet import reactor
 
     d.addBoth(lambda _: reactor.stop())
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -113,8 +113,7 @@ reactor after ``MySpider`` has finished running.
     reactor.run()  # the script will block here until the crawling is finished
 
 Same example but using a non-default reactor, it's only necessary call
-``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess``
- already does this automatically.
+``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess`` already does this automatically.
 
 .. code-block:: python
 
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -111,7 +111,9 @@ reactor after ``MySpider`` has finished running.
     d.addBoth(lambda _: reactor.stop())
     reactor.run()  # the script will block here until the crawling is finished
 
-Same example but using a non-default reactor, is only necessary call ``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess`` already does this automatically.
+Same example but using a non-default reactor, it's only necessary call
+``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess``
+ already does this automatically.
 
 .. code-block:: python
 
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -128,6 +128,7 @@ Same example but using a non-default reactor, it's only necessary call
 
 
     configure_logging({"LOG_FORMAT": "%(levelname)s: %(message)s"})
+
     from scrapy.utils.reactor import install_reactor
 
     install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
--- a/docs/topics/practices.rst
+++ b/docs/topics/practices.rst
@@ -92,7 +92,6 @@ reactor after ``MySpider`` has finished running.
 
 .. code-block:: python
 
-    from twisted.internet import reactor
     import scrapy
     from scrapy.crawler import CrawlerRunner
     from scrapy.utils.log import configure_logging
@@ -107,6 +106,33 @@ reactor after ``MySpider`` has finished running.
     runner = CrawlerRunner()
 
     d = runner.crawl(MySpider)
+    from twisted.internet import reactor
+
+    d.addBoth(lambda _: reactor.stop())
+    reactor.run()  # the script will block here until the crawling is finished
+
+Same example but using a non-default reactor, is only necessary call ``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess`` already does this automatically.
+
+.. code-block:: python
+
+    import scrapy
+    from scrapy.crawler import CrawlerRunner
+    from scrapy.utils.log import configure_logging
+
+
+    class MySpider(scrapy.Spider):
+        # Your spider definition
+        ...
+
+
+    configure_logging({"LOG_FORMAT": "%(levelname)s: %(message)s"})
+    from scrapy.utils.reactor import install_reactor
+
+    install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
+    runner = CrawlerRunner()
+    d = runner.crawl(MySpider)
+    from twisted.internet import reactor
+
     d.addBoth(lambda _: reactor.stop())
     reactor.run()  # the script will block here until the crawling is finished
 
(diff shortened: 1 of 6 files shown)


docs/topics/practices.rst
```diff
@@ -107,6 +106,33 @@ reactor after ``MySpider`` has finished running.
     runner = CrawlerRunner()
 
     d = runner.crawl(MySpider)
+    from twisted.internet import reactor
```
wRAR, line 109:
```suggestion

        from twisted.internet import reactor
```

wRAR: (I assume the empty line after the import is enforced by black)

docs/topics/practices.rst
```diff
@@ -107,6 +106,33 @@ reactor after ``MySpider`` has finished running.
...
+
+    d.addBoth(lambda _: reactor.stop())
+    reactor.run()  # the script will block here until the crawling is finished
+
+Same example but using a non-default reactor, is only necessary call ``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess`` already does this automatically.
```
wRAR, line 114:
```suggestion
Same example but using a non-default reactor, it's only necessary call
``install_reactor`` if you are using ``CrawlerRunner`` since ``CrawlerProcess``
 already does this automatically.
```

docs/topics/practices.rst
```diff
@@ -107,6 +106,33 @@ reactor after ``MySpider`` has finished running.
...
+
+    class MySpider(scrapy.Spider):
+        # Your spider definition
+        ...
+
+
+    configure_logging({"LOG_FORMAT": "%(levelname)s: %(message)s"})
+    from scrapy.utils.reactor import install_reactor
+
+    install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
```
wRAR, line 131: Can you please check if this example works when `AsyncioSelectorReactor` is not added to settings via settings.py/custom_settings? Not sure if the reactor is compared to the setting in this case.

docs/topics/practices.rst
```diff
@@ -209,6 +236,8 @@ Same example but running the spiders sequentially by chaining the deferreds:
         reactor.stop()
 
 
+    from twisted.internet import reactor
+
     crawl()
```
wRAR, line 241: I think install_reactor() is called here so the import is probably still too early, but I'm not 100% sure.

wRAR:
I've just recalled we have testcases for similar things, in `tests/Crawler{Process,Runner}` (the actual tests are in tests/test_crawler.py`), can you please check if there are some cases that make sense to put into docs but are not covered by those, and add them if needed?

docs/topics/practices.rst
Laerte: Is not compared.

https://github.com/scrapy/scrapy/blob/631fc65fadb874629787ae5f7fdd876b9ec96a29/scrapy/crawler.py#L117-L127

Logs:
````
2024-05-22 20:48:25 [asyncio] DEBUG: Using selector: KqueueSelector
````

File: docs/topics/practices.rst (hunk 4 above)
Comment (Laerte) [reply docs/topics/practices.rst:244]: Is not called because we are using `CrawlerRunner`, but yes if we are using different `TWISTED_REACTOR` we need to call `install_reactor` before importing `reactor` here.

**Comment by Laerte** [comment]
> I've just recalled we have testcases for similar things, in `tests/Crawler{Process,Runner}` (the actual tests are in `tests/test_crawler.py`), can you please check if there are some cases that make sense to put into docs but are not covered by those, and add them if needed?

I've taken a look and seems that https://github.com/scrapy/scrapy/blob/master/tests/CrawlerProcess/sleeping.py is the only one that would be good to add to docs (let me know if you find other interesting), e.g: scenarios where user needs to wait few seconds to something get processed by site but don't want to I/O block other requests using `time.sleep` this example is only for `asyncio` reactor, using https://stackoverflow.com/a/36985249/5314295 as reference I created this snippet:

```python
from time import time
from urllib.parse import parse_qsl

from scrapy import Spider
from scrapy.crawler import CrawlerRunner
from scrapy.utils.project import get_project_settings


class MySpider1(Spider):
    name = "my_spider"

    io_block = False  # change to simulate I/O blocking

    def start_requests(self):
        for x in range(10):
            yield Request(url=f"https://httpbin.org/anything?sleep_for={x}")

    def parse(self, response):

        d = Deferred()
        start = time()

        def yield_item():
            end = time()
            self.logger.info(f"[finished] zZzZZ {sleep_for}, took: {end - start}")
            yield response.json()

        query_params = dict(parse_qsl(response.url.split("?")[-1]))
        sleep_for = int(query_params["sleep_for"])

        if sleep_for % 2 == 0:
            self.logger.info(f"[start] zZzZ... {sleep_for}")

            if self.io_block:
                from time import sleep

                sleep(sleep_for)
                return yield_item()
            else:
                reactor.callLater(sleep_for, d.callback, yield_item())
                return d
        else:
            return response.json()


settings = get_project_settings()
configure_logging(settings)
runner = CrawlerRunner(settings)

d = runner.crawl(MySpider1)
d.addBoth(lambda _: reactor.stop())
reactor.run()
```

But it seems to verbose to me, is there a way to simplify this? So we don't need to write `yield_item` function?

wRAR:
I see I phrased that poorly, I meant that we should add (or modify) test cases so that things we have put into docs are actually tested, can you please check that if you didn't already?

As for `sleeping.py`, that file is for testing graceful/forced shutdowns so it doesn't need to be copied to docs itself. It indeed may be useful to have a sleeping example in docs (that doesn't use `asyncio.sleep()`), as a separate PR though. As for how to do it - I'm actually not sure how does your example work and it looks like it has too many functions, but I don't think it's possible to do in a clean way on pure Twisted deferreds, and probably the only sane way is making the callback `async def` and `await`ing on a deferred that gets called via `callLater`.

docs/topics/practices.rst
```diff
@@ -107,6 +106,35 @@ reactor after ``MySpider`` has finished running.
...
+    from scrapy.utils.log import configure_logging
+
+
+    class MySpider(scrapy.Spider):
+        # Your spider definition
+        ...
+
+
+    configure_logging({"LOG_FORMAT": "%(levelname)s: %(message)s"})
+    from scrapy.utils.reactor import install_reactor
```
wRAR, line 131:
```suggestion

    from scrapy.utils.reactor import install_reactor
```

docs/topics/practices.rst
```diff
@@ -107,6 +106,35 @@ reactor after ``MySpider`` has finished running.
...
+
+    install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
+    runner = CrawlerRunner()
+    d = runner.crawl(MySpider)
+    from twisted.internet import reactor
```
wRAR, line 136:
```suggestion

    from twisted.internet import reactor
```

docs/topics/practices.rst
```diff
@@ -173,6 +200,8 @@ Same example using :class:`~scrapy.crawler.CrawlerRunner`:
     runner.crawl(MySpider1)
     runner.crawl(MySpider2)
     d = runner.join()
+    from twisted.internet import reactor
```
wRAR, line 203:
```suggestion

    from twisted.internet import reactor
```

wRAR:
(I've forgot that I wanted the test changes if they are needed, please don't merge this without them)

Laerte:
> (I've forgot that I wanted the test changes if they are needed, please don't merge this without them)

Oh yeah I will check later.

Laerte:
@wRAR Things that I noticed while creating the test:

1. If you define `TWISTED_REACTOR` but don't install when using `CrawlerRunner` the script just hangs:
```
2024-05-26 20:00:17 [scrapy.addons] INFO: Enabled addons:
[]
```

To get the exception we need to define `addErrback` with call to log then the exception is printed but the reactor is not stoped, so I added another `addErrback` with `sys.exit(1)` call but don't seems right, here the snippet:

```python
from scrapy import Spider
from scrapy.crawler import CrawlerRunner


class NoRequestsSpider(Spider):
    name = "no_request"

    custom_settings = {
        "TWISTED_REACTOR": "twisted.internet.asyncioreactor.AsyncioSelectorReactor",
    }

    def start_requests(self):
        return []


configure_logging({"LOG_FORMAT": "%(levelname)s: %(message)s", "LOG_LEVEL": "DEBUG"})

INSTALL_REACTOR = False

if INSTALL_REACTOR:
    from scrapy.utils.reactor import install_reactor

    install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")

runner = CrawlerRunner()

d = runner.crawl(NoRequestsSpider)

from twisted.python import log
import sys

d.addErrback(log.err)
d.addErrback(sys.exit(1))
d.addBoth(callback=lambda _: reactor.stop())
reactor.run()
```

Let me know if you think we should add this test or is there a better way to write this script.

wRAR:
Not having the error logging is #6047, or at least directly related to it. Not sure why we need to call `sys.exit()` to properly finish the process, though (maybe it's not a very interesting question though). In any case it seems out of scope here.

Laerte:
> Not having the error logging is #6047, or at least directly related to it. Not sure why we need to call `sys.exit()` to properly finish the process, though (maybe it's not a very interesting question though). In any case it seems out of scope here.

Nice, so is good to merge then!

wRAR:
Thanks!

Issue #5817, opened 2023-01-31
Missing traceback for exceptions occurred in `open_spider` method of an item pipeline


Prometheus3375:
### Description

Traceback shown for an exception raised in a pipeline while opening a spider contains not enough information regarding the pipeline which has thrown this exception.

### Steps to Reproduce

1. Create any simple spider.
2. Create pipeline in file `pipelines.py` with next code:
    ```python
    from typing import Any
    
    from scrapy import Spider
    from scrapy.crawler import Crawler
    
    
    class MyPipeline:
        @classmethod
        def from_crawler(cls, crawler: Crawler, /):
            return cls()
    
        def open_spider(self, spider: Spider, /):
            self.value = {}['']
    
        def close_spider(self, spider: Spider, /): pass
    
        def process_item(self, item: Any, spider: Spider, /) -> Any:
            return item

    ```
5. Add pipeline to the settings:
    ```python
    ITEM_PIPELINES = {
        'pipelines.MyPipeline': 600,
    }
    ```
6. Start crawling with spider.
7. Observe logs.

**Expected behavior:**
Comprehensive traceback with exact position which caused the exception.

**Actual behavior:**

Here is an example of logs I received. Project directory was redacted.

```
2023-01-31 05:04:05 [scrapy.core.engine] INFO: Spider opened
2023-01-31 05:04:06 [scrapy.core.engine] INFO: Closing spider (shutdown)
2023-01-31 05:04:06 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method SpiderState.spider_closed of <scrapy.extensions.spiderstate.SpiderState object at 0x000001A66705EE30>>
Traceback (most recent call last):
  File "<redacted>\.venv\lib\site-packages\scrapy\crawler.py", line 118, in crawl
    yield self.engine.open_spider(self.spider, start_requests)
KeyError: ''

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<redacted>\.venv\lib\site-packages\scrapy\utils\defer.py", line 292, in maybeDeferred_coro
    result = f(*args, **kw)
  File "<redacted>\.venv\lib\site-packages\pydispatch\robustapply.py", line 55, in robustApply
    return receiver(*arguments, **named)
  File "<redacted>\.venv\lib\site-packages\scrapy\extensions\spiderstate.py", line 29, in spider_closed
    pickle.dump(spider.state, f, protocol=4)
[...]

PR #6376, opened 2024-05-26
Improve traceback for exceptions in open_spider method

mghotz:
This pull request addresses issue #5817 by enhancing the exception handling in the open_spider method of the Engine class. The current implementation does not provide sufficient traceback information when an exception occurs, making it difficult to debug issues. The proposed changes capture and log detailed traceback information, improving the debugging experience for users.

**Changes Made:**
Enhanced Exception Handling:
Wrapped the logic of the open_spider method in a try-except block.
Used the traceback module to capture and format the exception traceback.
Logged the detailed traceback in the crawler’s stats using self.crawler.stats.set_value.
Sent a spider_error signal with the detailed failure information using self.crawler.signals.send_catch_log.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -389,7 +389,7 @@ class ExecutionEngine:
             self.slot.heartbeat.start(5)
         except Exception as e:
             # Capture detailed traceback
-            tb_str = traceback.format_exception(etype=type(e), value=e, tb=e.__traceback__)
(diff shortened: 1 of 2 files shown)


wRAR:
Please use pre-commit and also run tests locally.

mghotz:
There is 58 errors before my changes. I did changes and ran tests locally. Please reviev it.

wRAR:
(tests still fail, with the same error)

Issue #5555, opened 2022-07-13
Attempting to acquire lock


Mankvis:
### Description

The latest version will continuously output the following content during use
[filelock] DEBUG: Attempting to acquire lock 4600430160 on

2022-07-13 11:15:34 [filelock] DEBUG: Attempting to acquire lock 4469884432 on ...... __a22fb8__tldextract-3.3.1/publicsuffix.org-tlds/de84b5ca2167d4c83e38fb162f2e8738.tldextract.json.lock
2022-07-13 11:15:34 [filelock] DEBUG: Lock 4469884432 acquired on  ...... __tldextract-3.3.1/publicsuffix.org-tlds/de84b5ca2167d4c83e38fb162f2e8738.tldextract.json.lock
2022-07-13 11:15:34 [filelock] DEBUG: Attempting to release lock 4469884432 on  ...... __a22fb8__tldextract-3.3.1/publicsuffix.org-tlds/de84b5ca2167d4c83e38fb162f2e8738.tldextract.json.loc

### Steps to Reproduce

After using genspider. to generate the crawler, you can execute it directly

### Versions

Scrapy       : 2.6.103
lxml         : 4.9.1.0
libxml2      : 2.9.4
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 1.22.0
Twisted      : 22.4.0
Python       : 3.10.5 (main, Jul 12 2022, 09:59:34) [Clang 13.0.0 (clang-1300.0.29.30)]
pyOpenSSL    : 22.0.0 (OpenSSL 3.0.5 5 Jul 2022)
cryptography : 37.0.4
Platform     : macOS-12.4-x86_64-i386-64bit

elacuesta:
Those are debug lines coming from the use of [tldextract](https://pypi.org/project/tldextract/) in the cookies middleware. They are expected, and from your report I don't see them actually causing a problem. If you absolutely don't want to see them, you could for instance:
* set the general log level to one higher than DEBUG via the `LOG_LEVEL` setting (`scrapy crawl spider_name -s LOG_LEVEL=INFO`)
* set the log level of that specific logger in your code
```python
import logging

logging.getLogger("filelock").setLevel(logging.INFO)
```

Yanzuzhang:
> ### Description
> The latest version will continuously output the following content during use [filelock] DEBUG: Attempting to acquire lock 4600430160 on
> 
[...]

PR #6377, opened 2024-05-26
Warn About Mutable Objects in Request.cb_kwargs When JOBDIR is Set

mghotz:
This pull request addresses issue #5555 by adding a warning for mutable objects in Request.cb_kwargs when the JOBDIR setting is enabled. Mutable objects can lead to unexpected behavior when job persistence is used, as mutations to these objects will not be persisted across spider runs.

Changes Made:
Enhanced Request Class:

Added a check in the __init__ method of the Request class to detect mutable objects in cb_kwargs.
Raised a UserWarning if mutable objects are found when JOBDIR is set.
Helper Function:

Added a _is_mutable helper function to identify mutable objects.

### DIFF
--- a/scrapy/http/request/__init__.py
+++ b/scrapy/http/request/__init__.py
@@ -8,6 +8,8 @@ See documentation in docs/topics/request-response.rst
 from __future__ import annotations
 
 import inspect
+import warnings
+from collections.abc import MutableMapping, MutableSequence
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -61,6 +63,8 @@ def NO_CALLBACK(*args: Any, **kwargs: Any) -> NoReturn:
         "called."
     )
 
+def _is_mutable(obj):
+    return isinstance(obj, (MutableMapping, MutableSequence))
 
 class Request(object_ref):
     """Represents an HTTP request, which is usually generated in a Spider and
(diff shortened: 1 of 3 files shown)


mghotz:
I open 2 pull requests. There is no any relation between this changes. Please review it.

wRAR:
Please remove unrelated commits and unrelated changes. Please add a test for your change.

scrapy/http/request/__init__.py
```diff
@@ -132,6 +136,15 @@ def __init__(
             dict(cb_kwargs) if cb_kwargs else None
         )
         self.flags: List[str] = [] if flags is None else list(flags)
+        # Check for mutable objects in cb_kwargs if JOBDIR is set
+        if settings and 'JOBDIR' in settings:
```
wRAR, line 140: `JOBDIR` is always in settings though, via `default_settings`?

mghotz: My implementation checks if JOBDIR is set to a non-None value to ensure that the warning is only issued when JOBDIR is explicitly set by the user.

mghotz:
> Please remove unrelated commits and unrelated changes. Please add a test for your change.

Okay. Thank you.

scrapy/http/request/__init__.py
wRAR: > My implementation checks if JOBDIR is set to a non-None value

Where?

PR #6381, opened 2024-05-29
Add support for multiple referer policy tokens.

0xdeb:
This PR adds support for multiple Referer policy tokens as specified in the spec. (https://www.w3.org/TR/referrer-policy/#unknown-policy-values)

I recently stumbled across this RuntimeWarning in one of the scrapers I maintain: 
```
RuntimeWarning: Could not load referrer policy 'no-referrer, strict-origin-when-cross-origin'
```

After having a look at the spec and scrapy's code, I saw that this part of the spec was not implemented and deemed it an easy fix. So here you go. 

I covered it with unit tests and all of the existing tests pass without issue. Let me know if you want me to change anything.


scrapy/spidermiddlewares/referer.py
```diff
@@ -323,15 +323,17 @@ def _load_policy_class(
...
-            return _policy_classes[policy.lower()]
-        except KeyError:
-            msg = f"Could not load referrer policy {policy!r}"
-            if not warning_only:
-                raise RuntimeError(msg)
-            else:
-                warnings.warn(msg, RuntimeWarning)
-                return None
+        tokens = [token.strip() for token in policy.lower().split(",")]
+        for token in tokens[::-1]:
```
Gallaecio, line 327: For future reference, it was not obvious to me why we are picking the last known policy instead of the first one.

```suggestion
        # https://www.w3.org/TR/referrer-policy/#parse-referrer-policy-from-header
        for token in tokens[::-1]:
```

Gallaecio:
Very nice!

0xdeb:
Cool, thank you! I also applied your suggestion with the link to the spec. 
This should then be ready to merge from my point of view.

wRAR (approved):

PR #6387, opened 2024-06-03
Full typing for scrapy/pipelines.

wRAR:


scrapy/pipelines/media.py
```diff
@@ -4,54 +4,101 @@
...
+                list
+            )
+
+    def __init__(
+        self,
+        download_func: Optional[Callable[[Request, Spider], Response]] = None,
+        settings: Union[Settings, Dict[str, Any], None] = None,
+    ):
         self.download_func = download_func
-        self._expects_item = {}
```
wRAR, line 39: Leftover from some recently removed compat code.

wRAR:
This fails on pinned Twisted because `Deferred` became a generic type only in 21.7.0 and we now want to use it at the run time (for cast()). I wonder what are our options short of bumping the dep version.

Gallaecio (approved):

PR #6388, opened 2024-06-03
added  AttributeError: 'NoneType' handling in contract parsing

### LINKED ISSUE #6383: Error handling in contract parsing
We found that there is no proper handling for unmatched regexes in `‎scrapy.contracts.ContractsManager.extract_contracts()`, so e.g. calling `from_method()` on a method with `@ foo` in the docstring produces an unhandled exception. I think we should just skip lines that don't match.

suvanbanerjee:
Added a condition to check if `m` is `None`.

The presence of `@ foo` in the docstring would result in an `AttributeError: 'NoneType' object has no attribute 'groups'`. Checking for `None` will prevent this error.

Closes #6383

### DIFF
--- a/scrapy/contracts/__init__.py
+++ b/scrapy/contracts/__init__.py
@@ -121,6 +121,8 @@ class ContractsManager:
             if line.startswith("@"):
                 m = re.match(r"@(\w+)\s*(.*)", line)
                 assert m is not None
+                if m is None:
+                    continue
                 name, args = m.groups()
                 args = re.split(r"\s+", args)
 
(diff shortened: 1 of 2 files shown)


wRAR:
The assert is no longer needed. You should also add a test for this change, I guess calling from_method() on a method with such a wrong line in the doc string should be enough. It also makes sense to have valid contract lines in the same docstring.

suvanbanerjee:
@wRAR, could you please review this?

wRAR (approved):
Thanks!

PR #6391, opened 2024-06-05
Add parameters to iterable generics, replace generators with iterables.

wRAR:


scrapy/utils/iterators.py
```diff
@@ -184,10 +181,13 @@ def csviter(
...
-
-    def row_to_unicode(row_: Iterable) -> List[str]:
-        return [to_unicode(field, encoding) for field in row_]
+    if encoding is not None:
+        warn(
+            "The encoding argument of csviter() is ignored and will be removed"
+            " in a future Scrapy version.",
+            category=ScrapyDeprecationWarning,
+            stacklevel=2,
+        )
```
Gallaecio, line 190: Shouldn’t we keep the encoding functionality until we remove the parameter?

Gallaecio (approved):

scrapy/utils/iterators.py
wRAR: The existing encoding functionality requires getting bytes in row_to_unicode(), which can't happen as all this code operates on strings (after `_body_or_str(obj, unicode=True)` converts the input with a hardcoded utf-8 encoding, anyway), so it was a no-op.

wRAR: AFAICS it was for the Py2 code path where `csv_r` produced bytes.

PR #6392, opened 2024-06-05
Add pyupgrade.

wRAR:


.pre-commit-config.yaml
```diff
@@ -22,3 +22,9 @@ repos:
   - id: blacken-docs
     additional_dependencies:
     - black==24.2.0
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.15.2
+  hooks:
+  - id: pyupgrade
+    args: [--py38-plus, --keep-runtime-typing]
+    exclude: scrapy/__init__.py
```
Gallaecio, line 30: Should we add a comment to explain the reasoning for this exclusion?

wRAR: Done.

Alternatively, we can remove that check.

Gallaecio: I’m +0.5 to remove it, given pip has supported `python_requires` in `setup.py` [for ~7.5 years](https://pip.pypa.io/en/stable/news/#v9-0-0), but I am OK merging as is.

Gallaecio (approved):

Issue #6398, opened 2024-06-10
Improve the docs of the dont_filter parameter of Request


Gallaecio:
The current docs are inaccurate, as they limit the scope of the parameter to the duplicate filter of the scheduler, while there are additional built-in middlewares that take this parameter into account, and third-party middlewares may also use it.

Maybe we could describe it as asking not to drop the request, in general, and cover a detailed list of built-in middlewares that take this into account.

PR #6401, opened 2024-06-12
Update dont_filter param within request-response.rst

bmazaoreg:
This request updates the dont_filter param documentation to increase its accuracy as suggested in #6398 
A description of what the param is asking, as well as a list of built-in middlewares that take this into account are added.

### DIFF
--- a/docs/topics/request-response.rst
+++ b/docs/topics/request-response.rst
@@ -144,10 +144,25 @@ Request objects
        Negative values are allowed in order to indicate relatively low-priority.
     :type priority: int
 
-    :param dont_filter: indicates that this request should not be filtered by
-       the scheduler. This is used when you want to perform an identical
-       request multiple times, to ignore the duplicates filter. Use it with
-       care, or you will get into crawling loops. Default to ``False``.
+    :param dont_filter: indicates that this request should not be dropped by any 
+       middleware or the scheduler. This parameter is crucial for scenarios where you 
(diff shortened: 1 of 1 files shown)


docs/topics/request-response.rst
```diff
@@ -144,10 +144,25 @@ Request objects
...
+
+       Built-in Middlewares that take 'dont_filter' into account:
+
+       -   OffSiteMiddleware: Filters out requests for URLs outside the domains covered by the spider. 
+           If the request has the `dont_filter` attribute set, the offsite middleware will allow the 
+           request even if its domain is not listed in allowed domains 
+
+       -   DepthMiddleware: Tracks the depth of each request inside the site being scraped. It sets 
+           `request.meta['depth'] = 0` whenever there is no value previously set and increments it by 1 otherwise. 
+           The `dont_filter` attribute can influence how requests are prioritized based on their depth
```
Gallaecio, line 165: I think we can keep it much shorter without losing much information. Also, looking at the code, it does not look like DepthMiddleware respects `dont_filter`.

```suggestion
    :param dont_filter: indicates whether :ref:`components <topics-components>`
        may drop this request (``False``, default) or not (``True``).

        :ref:`Built-in schedulers <topics-scheduler>` (which by default drop
        duplicate requests) and
        :class:`~scrapy.downloadermiddlewares.offsite.OffSiteMiddleware`
        respect this parameter. Other third-party Scrapy components may respect
        it as well.
```

wRAR:

PR #6406, opened 2024-06-19
Add some flake8 plugins

wRAR:


.flake8
```diff
@@ -1,8 +1,63 @@
 [flake8]
 
 max-line-length = 119
-ignore = E203, E501, E701, E704, W503
+ignore =
+    # black disagrees with flake8 about these
+    E203, E501, E701, E704, W503
 
+    # Assigning to `os.environ` doesn't clear the environment.
+    B003
```
wRAR, line 9: This looks valid, I didn't check how do we use os.environ, maybe we should pass it to subprocesses explicitly as it suggests.

.flake8
```diff
@@ -1,8 +1,63 @@
...
+    # Do not use mutable data structures for argument defaults.
+    B006
```
wRAR, line 11: I didn't check the code.

.flake8
```diff
@@ -1,8 +1,63 @@
...
+    # Loop control variable not used within the loop body.
+    B007
```
wRAR, line 13: No opinion about this (also I didn't check the code).

.flake8
```diff
@@ -1,8 +1,63 @@
...
+    # Do not perform function calls in argument defaults.
+    B008
```
wRAR, line 15: May be valid.

.flake8
```diff
@@ -1,8 +1,63 @@
...
+    # return/continue/break inside finally blocks cause exceptions to be
+    # silenced.
+    B012
+    # Star-arg unpacking after a keyword argument is strongly discouraged
+    B026
```
wRAR, line 20: I checked the code but didn't think if it can/should be changed.

Gallaecio (approved):
I’m OK with merging as is and solving the remaining issues separately later on.

PR #6413, opened 2024-06-22
Add flake8-type-checking.

wRAR:
The diff is large but the only changes are in imports (moving them under TYPE_CHECKING and a small amount of related cleanup)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,7 @@
 [flake8]
 
 max-line-length = 119
+extend-select = TC, TC1
 ignore =
     # black disagrees with flake8 about these
     E203, E501, E701, E704, W503
@@ -58,6 +59,9 @@ ignore =
     D402
     # First word of the first line should be properly capitalized
     D403
+
+    # Annotation in typing.cast() should be a string literal
+    TC006
 exclude =
     docs/conf.py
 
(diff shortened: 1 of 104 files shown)


.flake8
```diff
@@ -58,6 +59,9 @@ ignore =
     D402
     # First word of the first line should be properly capitalized
     D403
+
+    # Annotation in typing.cast() should be a string literal
+    TC006
```
wRAR, line 64: I will probably do this separately (it helps with runtime performance, though it's unclear how noticeable is the difference, and more importantly it will forbid non-string `Deferred[foo]` expressions which aren't supported on older Twisted).

wRAR:
There are some problems with Sphinx and type hints (e.g. #9813, #11225), some of them manifest here as "more than one target found for cross-reference" errors, so we'll need to keep some imports out of `TYPE_CHECKING` blocks. I've also checked the diff between docs for master and for this change and while some of the differences are for the worse (some links removed) others are unexpectedly for the better (e.g. some links _added_, or e.g. the weird unresolved Sphinx refs [here](https://docs.scrapy.org/en/2.11/topics/downloader-middleware.html#scrapy.downloadermiddlewares.retry.get_retry_request) were replaced with normal type hints).

Gallaecio (approved):

Issue #5289, opened 2021-10-22
Better error (or support?) when yielding items from start requests


Gallaecio:
We have seen more than once users trying to yield items from the `start_requests` method.

We should consider providing a better error message in this scenario, the current one is not very helpful.

We could also consider supporting this use case instead.

GeorgeA92:
On `scrapy.core.scraper.Scraper._process_spidermw_output` application call `engine.crawl` if spider output type is Request and `.itemproc.process_item` if spider output is item:
https://github.com/scrapy/scrapy/blob/8c8fb67057609d79a31e76587fc66b162d746906/scrapy/core/scraper.py#L290-L297

While on `scrapy.core.engine.ExecutionEngine._schedule_request` where it process results of `.start_requests` only Request objects is expected (we see only engine's call `self.crawl(request)`):
https://github.com/scrapy/scrapy/blob/8c8fb67057609d79a31e76587fc66b162d746906/scrapy/core/engine.py#L180-L195

On following code sample I've added.. code from `Scraper` that call item processing into.. subclass of `ExecutionEngine`

<details><summary>start_items.py</summary>

```python
import csv
from itemadapter import is_item
import logging
import scrapy
from scrapy.crawler import CrawlerProcess

logger = logging.getLogger('scrapy.core.engine')

class CsvWriterPipeline(object):

    def open_spider(self, spider):
        self.file = open('start_items.csv', 'w',encoding="utf8")
        self.csvwriter = csv.writer(self.file)
        self.csvwriter.writerow(["Author", "Quote", "Tags"])

    def process_item(self, item, spider):
        row = [item["author"], item["quote"], ",".join(item["tags"])]
        self.csvwriter.writerow(row); return item

    def close_spider(self, spider): self.file.close()

class ExecutionEngineStartItems(scrapy.core.engine.ExecutionEngine):
    def _next_request(self) -> None:
        if self.slot is None: return
        assert self.spider is not None  # typing
        if self.paused: return None
        while (not self._needs_backout() and self._next_request_from_scheduler() is not None): pass

        if self.slot.start_requests is not None and not self._needs_backout():
            try:
                request = next(self.slot.start_requests)
            except StopIteration:
                self.slot.start_requests = None
            except Exception:
                self.slot.start_requests = None
                logger.error(
                    "Error while obtaining start requests",
                    exc_info=True,
                    extra={"spider": self.spider},
                )
            else:
                if isinstance(request, scrapy.Request):
[...]

PR #6417, opened 2024-06-26
start_requests can return items

GeorgeA92:
resolve https://github.com/scrapy/scrapy/issues/5289
based on code sample from https://github.com/scrapy/scrapy/issues/5289#issuecomment-1522396870

I used this сode sample to test this locally.
<details><summary>script.py</summary>

```python
import scrapy

    def start_requests(self):
        yield {"quote": ['“It is possible to return items on start_requests. But I don\'t understand Why it\'s needed?“'], "author": ["Georgiy"], "tags": ["scrapy"]}

    def parse(self, response):
            yield {
                "quote": quote.css("span.text::text").getall(),
                "author": quote.css("small.author::text").getall(),
                "tags": quote.css("a.tag::text").getall()
            }
if __name__ == "__main__":
    p = CrawlerProcess(); p.crawl(QuotesToScrapeSpider); p.start()
```

</details>
At this moment I didn't found test where output(type) of  start_requests checked

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -211,11 +211,10 @@ class ExecutionEngine:
                 elif is_item(request_or_item):
                     self.scraper.start_itemproc(request_or_item, response=None)
                 else:
-                    type_import_path = global_object_name(type(request_or_item))
                     logger.error(
-                        f"Got an instance of {type_import_path} among start "
-                        f"requests. Only requests and items are supported. It "
-                        f"will be ignored."
+                        f"Got {request_or_item!r} among start requests. Only "
+                        f"requests and items are supported. It will be "
+                        f"ignored."
                     )
 
         if self.spider_is_idle() and self.slot.close_if_idle:
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -209,8 +209,7 @@ class ExecutionEngine:
                 if isinstance(request_or_item, Request):
                     self.crawl(request_or_item)
                 elif is_item(request_or_item):
-                    src = f"{global_object_name(self.spider.__class__)}.start_requests"
-                    self.scraper.start_itemproc(request_or_item, src)
+                    self.scraper.start_itemproc(request_or_item, response=None)
                 else:
                     type_import_path = global_object_name(type(request_or_item))
                     logger.error(
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -24,6 +24,7 @@ from typing import (
     cast,
 )
 
+from itemadapter import is_item
 from twisted.internet.defer import Deferred, inlineCallbacks, succeed
 from twisted.internet.task import LoopingCall
 from twisted.python.failure import Failure
@@ -205,7 +206,19 @@ class ExecutionEngine:
                     extra={"spider": self.spider},
                 )
             else:
-                self.crawl(request)
+                if isinstance(request, Request):
+                    self.crawl(request)
+                elif is_item(request):
+                    self.scraper.slot.itemproc_size += 1
+                    dfd = self.scraper.itemproc.process_item(
+                        request, self.crawler.spider
+                    )
+                    dfd.addBoth(
+                        self.scraper._itemproc_finished,
+                        request,
+                        f"{self.crawler.spider.__class__.__name__}.start_requests",
+                        self.crawler.spider,
+                    )
 
         if self.spider_is_idle() and self.slot.close_if_idle:
             self._spider_idle()
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -195,7 +195,7 @@ class ExecutionEngine:
 
         if self.slot.start_requests is not None and not self._needs_backout():
             try:
-                request = next(self.slot.start_requests)
+                request_or_item = next(self.slot.start_requests)
             except StopIteration:
                 self.slot.start_requests = None
             except Exception:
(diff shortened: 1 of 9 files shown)


scrapy/core/engine.py
```diff
@@ -205,7 +206,19 @@ def _next_request(self) -> None:
...
+                    self.scraper.slot.itemproc_size += 1
+                    dfd = self.scraper.itemproc.process_item(
+                        request, self.crawler.spider
+                    )
+                    dfd.addBoth(
+                        self.scraper._itemproc_finished,
+                        request,
+                        f"{self.crawler.spider.__class__.__name__}.start_requests",
+                        self.crawler.spider,
+                    )
```
Gallaecio, line 221: Looks good to me.

I wonder if maybe we should fallback to self.crawl, i.e. add else with self.crawl, to maintain backward compatibility.

I also wonder if the elif code could be wrapped into a method, if that way we can avoid any code duplication, but otherwise this looks fine to me.

We do need tests and mention this, the ability to yield items, in the reference docs of the start_requests method.

Gallaecio:
> At this moment I didn't found test where output(type) of start_requests checked.

There are a few if you search tests for `start_requests`, e.g. in [here](https://github.com/scrapy/scrapy/blob/41e15e93e7459673e93ff2591462b47b7ae01566/tests/test_crawl.py#L166) or [here](https://github.com/scrapy/scrapy/blob/41e15e93e7459673e93ff2591462b47b7ae01566/tests/test_spidermiddleware.py#L328), but it may not be straightforward to adapt.

Regardless of whether there are existing tests or not, I think at the very minimum we need a test that yields an item in a crawl with all built-in spider middlewares enabled (did not check if there are built-in ones that are not enabled by default), just to be sure that they do not fail due to expecting the output of start_request to be requests, e.g. assume that the items in the result iterable have some attribute or method they may not have.

scrapy/core/engine.py
wRAR: There is also process_start_requests in spider middlewares, at least its doc needs to be updated.

GeorgeA92:
> There are a few if you search tests for start_requests, e.g. in [here](https://github.com/scrapy/scrapy/blob/41e15e93e7459673e93ff2591462b47b7ae01566/tests/test_crawl.py#L166) or [here](https://github.com/scrapy/scrapy/blob/41e15e93e7459673e93ff2591462b47b7ae01566/tests/test_spidermiddleware.py#L328), but it may not be straightforward to adapt.

Test cases that cover items from `start_requests` added

tests/test_spidermiddleware.py
```diff
@@ -328,12 +328,13 @@ def process_start_requests(self, start_requests, spider):
 class ProcessStartRequestsSimple(BaseAsyncSpiderMiddlewareTestCase):
     """process_start_requests tests for simple start_requests"""
 
-    ITEM_TYPE = Request
+    ITEM_TYPE = Union[Request, dict]
```
wRAR, line 331: Looks like this doesn't work on 3.9 because `isinstance()` can't use it.

GeorgeA92: Changing this to tuple `ITEM_TYPE = (Request, dict)` resolve issues with `isinstance()` but it cause failing of `checks (3.8 typing-tests)`

scrapy/core/engine.py
GeorgeA92: > We do need tests and to mention this (the ability to yield items) in the reference docs of the start_requests method.

> There is also process_start_requests in spider middlewares, at least its doc needs to be updated.

Added as result of https://github.com/scrapy/scrapy/pull/6417/commits/8f24f131a64914963c8d95d81f972087cd7f239d

tests/test_spidermiddleware.py
wRAR: I doubt that, `isinstance()` supporting a tuple is unique to it, you cannot use a tuple instead of a union.

wRAR: (I haven't looked at this code yet but I suspect we need to rewrite it, e.g. by rewriting the assert)

GeorgeA92: > I doubt that, isinstance() supporting a tuple is unique to it, you cannot use a tuple instead of a union.

It is directly mentioned on python docs that it support tuples as argument
https://docs.python.org/3.8/library/functions.html?highlight=tuple#isinstance
https://docs.python.org/3.9/library/functions.html?highlight=tuple#isinstance
https://docs.python.org/3.10/library/functions.html?highlight=tuple#isinstance
https://docs.python.org/3.11/library/functions.html?highlight=tuple#isinstance

built in `issubclass` - also supports tuple for its checks as well as `isinstance` so it is not unique.
As it is directly mentioned on docs - I think it's valid to use tuple here.

wRAR: I've meant that you can use a tuple for isinstance() but not for type hints. But I've now checked the actual failure, it's "Incompatible types in assignment (expression has type "Tuple[Type[Request], Type[Dict[Any, Any]]]", base class "BaseAsyncSpiderMiddlewareTestCase" defined the type as "type")" and it should be possible to solve it by updating the type hint of `BaseAsyncSpiderMiddlewareTestCase.ITEM_TYPE` to e.g. `Union[type, Union]` or something like that.

wRAR:
Nice, what's missing here?

Gallaecio:
I think we need to deprecate `response` in item signals, change it to `src: Any`. But doing that in a backward-compatible way seems not trivial.

scrapy/core/scraper.py
```diff
@@ -333,6 +329,19 @@ def _process_spidermw_output(
             )
         return None
 
+    def start_itemproc(self, item, *, response: Optional[Response]):
```
wRAR, line 332:
```suggestion
    def start_itemproc(self, item, *, response: Optional[Response]) -> Deferred[Any]:
```

wRAR: (this most likely needs reformatting)

wRAR (approved):

scrapy/core/scraper.py
Gallaecio: Did not need it in the end (by 2 chars).

Gallaecio (approved):
@GeorgeA92 Please confirm you are also OK with the current code before we merge.

GeorgeA92:
@Gallaecio 
> @GeorgeA92 Please confirm you are also OK with the current code before we merge.

I am OK with current code

icaca:
I tried to use peewee to read some data in start_requests, and then used a for loop to call yield (no http request was made, the database was updated directly). I found that there was a 5-second delay each time. I printed logs at the entrance and end of MySQLPipeline and confirmed that the database operation was in milliseconds. I tried to set DOWNLOAD_DELAY and AUTOTHROTTLE_START_DELAY in the setting file, but it didn't work. I wonder if anyone has encountered this situation.

PR #6427, opened 2024-07-03
Update the outdated Item docstring.

wRAR:
I've just noticed this outdated text and updated it but there is a bigger related issue with https://docs.scrapy.org/en/latest/topics/items.html as this docstring (and the docstring of `Field`) isn't included there and also names of those two classes are not clickable there (not sure what's the proper fix for the last problem).


scrapy/item.py
```diff
@@ -61,16 +61,13 @@ def __new__(
...
 
-    If you need instances of a custom class to be considered items by Scrapy,
-    you must inherit from either :class:`Item` or :class:`dict`.
+    In Scrapy, an object is considered an ``item`` if it's supported by the
+    `itemadapter`_ library. For example, when the output of a spider callback
+    is evaluated, only such objects are passed to :ref:`item pipelines
+    <topics-item-pipeline>`. :class:`Item` is one of the classes supported by
+    `itemadapter`_ by default.
 
     Items must declare :class:`Field` attributes, which are processed and stored
```
Gallaecio, line 72: Maybe changing ``` :class:`Field` ``` to ``` :class:`~scrapy.Field` ``` or (more likely) ``` :class:`~scrapy.item.Field` ``` fixes the linking.

wRAR: There is also another problem, the fragments in generated links: https://docs.scrapy.org/en/latest/topics/items.html#scrapy.item.scrapy.item.Item

Probably because of `.. module:: scrapy.item` at the beginning of items.rst?

And what do you think about adding the docstring to the documentation page, or is it a separate more general task?

Gallaecio: > Probably because of `.. module:: scrapy.item` at the beginning of items.rst?

Indeed.

> And what do you think about adding the docstring to the documentation page, or is it a separate more general task?

No strong opinion, up to you.

Gallaecio (approved):

PR #6430, opened 2024-07-05
Pause request processing based on the sum of body sizes of all active responses

### LINKED ISSUE #4583: "Exploding" memory usage when using HTTP cache
_NOTE: Some parts are conjectures and I would like feedback if it is really an issue or not._

### Description

When using the HTTP cache, memory usage seems to explode as compared to un-cached processing.
To me, it seems like an issue(?) which happens because for each new request a lookup for already cached requests/responses happens and if it exists, it is immediatly loaded. For un-cached requests, the response etc. is only "populated" when the actual crawl happens, so the memory only grows slowly if any. (_Conjecture from me about the workings of the cache. Not sure, only observations._)

This is an problem for me because I run on limited memory and want to crawl directory pages that link to more than 100 sub-pages. For un-cached processing, a 700 links directory only amounts to around 300MB. But if using a cache and with over 100 cached request/response pairs, over 2-3GB of memory consumption appears. Using the built-in memory debugger and telnet, I see many live responses that are only gradually handled but with newer requests, the memory is not really freed any time. (_resource exhaustion on my device happens before that_)

---

TL;DR: My main point, I guess:
- Cached responses should only be loaded if they are being processed?!

Note: As scrapy is rather modular, I'm actually not sure when a cached request/response is being loaded. (a) immediatly when a new request is yielded or (b) when it is thrown to my callback handler. (And I did not have the setup to debug it step-by-step.) But because of caching the usual crawling delays do not happen (obviously), so I only recieve responses without being able to process them all?

### Steps to Reproduce

1. crawl a listing page with subpages, with enabled HTTP caching extension
2. abort half-way
3. restart crawl, so it can make use of cached responses

**Expected behavior:** slowly increasing memory consumption because of scraped content/Items

**Actual behavior:** "exploding" memory usage that presumably decreases with time, but is many times the amount of an un-cached crawl, my live requests in telnet console

**Reproduces how often:** 100% for large listing pages with already cached requests

### Versions

```
scrapy 2.1.0
[...]
```

I update to new versions but I'm still using the older Twisted async thingies ...
Configurations for scrapy are mostly default ones.

### Additional context

It might be an issue because of how my code is structured, but it seemed logical to me:
```
# imagine a novel (collection) website: (or something similar)
# - novel summary page with links to chapters (main -> 1., 2., 3., ...)
# - chapter pages that can be split into subpages (1., 2.1. -> 2.2., 3.1. -> 3.2. -> 3.3., 4., ...)

# my workflow:
-- summary page:
0. (optional login)
1. crawl directory/listing page with many links to sub-pages
  1.1. extract summary contents, yield as item
  1.2. for each found link on listing page, yield a new request, in a loop

-- for each page p:
1. handler for subpages extracts content and yields items
2. optionally: goes linearly to the next page for this subpage (branch) (p -> p2 -> ... pn)
```

As it is the same domain, starting crawlers for each "branch" of subpage in the listing is contradictive as I want to limit traffic and use "smart" delays and the depth of (sub-)pages is 3 (mean), at most 20, typically. And I want to use the cache if my crawl with average delay of 15-30 seconds aborts after 1-2 hours and has just another hour to go.

What I also observe is that `Selector` instances seem to accumulate depending on how I yield my items and new requests. If I yield multiple requests on a single response/selector, it seems to live a long time. Sometimes until all descendant requests are finished, and then they are all freed. This can be a typical behaviour when using async programming, not sure..?

Working solutions:
- not using a cache ...
- (still trying) setting parallel/workers/... to 1, so only a single request/response can be handled at any time (behaviour I also want for crawl delays etc.) -- http caching results not conclusive as of yet.

### LINKED ISSUE #5302: Improve diagnostics when stuck because of SCRAPER_SLOT_MAX_ACTIVE_SIZE
In #3551 a way to tune the maximum size of responses being processed was added but the user experience is bad in the use cases that caused creation of that issue and PR (the ones described as lockups). I've seen the following use case:

```python
class S(Spider):
    start_urls = [http://big/input/file.txt]

    def parse(self, response):
        for url in response.text.splitlines():
            yield Request(url, callback=self.parse_page)
```

As far as I can see, if the input file is larger than the default 5Mb this will not download any actual pages until the file finishes processing. Unfortunately, nothing suggests that increasing the setting will help here.

### LINKED ISSUE #6206: max_active_size can cause deadlock when deferring items in a pipeline
### Description

When you hit the SCRAPER_SLOT_MAX_ACTIVE_SIZE requests get dropped silently with no warning.
If you are deferring items in a pipeline then your crawler will never finish those requests if they whats needed to finish the original deferred. So it's a deadlock with no explanation which took a long time to work out why.
This is made worse by the prioritydownloader as this preferences less active sites so the bottleneck site causing the deadlock never gets the requests as they are the ones dropped.

[[Description of the issue]](https://github.com/scrapy/scrapy/pull/3551)

### Steps to Reproduce

1. Run a crawl with a pipeline that will defer items. This will mean you hit max_active_size fast.
2. Notice no warning or error
3. [and so on...]

An error when you hit max_active_size. Stats addon to show number of active requests.

Nothing shown so hard to debug

100%

### Versions

2.11.0, 2.7.0

Please paste here the output of executing `scrapy version --verbose` in the command line.

```
Scrapy       : 2.11.0
lxml         : 4.9.1.0
libxml2      : 2.9.14
cssselect    : 1.1.0
parsel       : 1.6.0
w3lib        : 2.0.1
Twisted      : 22.8.0
Python       : 3.9.18 (main, Oct  3 2023, 08:56:37) - [Clang 12.0.0 (clang-1200.0.32.2)]
pyOpenSSL    : 22.1.0 (OpenSSL 3.0.5 5 Jul 2022)
cryptography : 38.0.1
Platform     : macOS-10.16-x86_64-i386-64bit
```

### Additional context

Any additional information, configuration, data or output from commands that might be necessary to reproduce or understand the issue. Please try not to include screenshots of code or the command line, paste the contents as text instead. You can use [GitHub Flavored Markdown](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) to make the text look better.

Gallaecio:
1st part of a 2-part solution for #482. Resolves https://github.com/scrapy/scrapy/issues/4583, resolves #5302, resolves https://github.com/scrapy/scrapy/issues/6206 (the reproducible part, that is).

Implements a `RESPONSE_MAX_ACTIVE_SIZE` setting that determines the maximum sum of body sizes that is allowed; when over that limit, no new requests are processed. It can be disabled with `0`. It replaces `SCRAPER_SLOT_MAX_ACTIVE_SIZE`.

It feels a bit hackish, but in order to count response sizes for as long as responses are used, even when they are kept in user code for whatever purposes, it might be best to actually track when those responses are garbage-collected, rather than tracking their lifetime in some other way.

It is not a good-enough solution for #482 because S3 storage does not seem to keep the response object around until the `deferToThread` finishes. I think we need to change that, but I think that’s better done separately.


Gallaecio:
> I first wanted to implement the tracking out of the response class, e.g. into the engine, but given downloader middlewares can replace responses with different ones (e.g. HTTP compression middleware), I’m not sure that’s such a good idea. I think the implementation would be more complex, although it would also feel less hacky, so I’m open to try that if there’s an interest.

After sleeping on it, I think I will give an implementation on the downloader middleware handler a try, that ought to cover all cases where responses can be spawned in properly written code for Scrapy.

GeorgeA92:
> Implements a `RESPONSE_MAX_ACTIVE_SIZE` setting that determines the maximum sum of body sizes that is allowed; when over that limit, no new requests are processed. It can be disabled with `0`. It replaces `SCRAPER_SLOT_MAX_ACTIVE_SIZE`.

It is preferred to provide way to set scraper class by setting as it possible with downloader and scheduler instead of.. hardcoding this (or any other behaviour) in single pre defined Scraper class.

Something like 

```SCRAPER = "scrapy.core.scraper.Scraper"``` - in default settings 
and this: (in engine code)

```
scraper_cls: Type[Scraper] = load_object(self.settings["Scraper"])
self.scraper: Scraper  =  scraper_cls(crawler)
```

File: scrapy/core/downloader/__init__.py [hunk 1]
```diff
@@ -143,8 +165,42 @@ def _deactivate(response: _T) -> _T:
...
+                    f"crawl speed. If your code keeps non-weak references to "
+                    f"Response objects, your crawl might get stuck "
+                    f"indefinitely; you can set RESPONSE_MAX_ACTIVE_SIZE to 0 "
+                    f"to disable this limit, but then your code might run out "
+                    f"of memory. This message will only appear the first time "
+                    f"this happens. To learn how often request processing has "
+                    f"been paused during a crawl for this reason, see the "
+                    f"request_backouts/response_max_active_size stat."
+                )
+            self._count_backout("response_max_active_size")
```
GeorgeA92, line 201: As You can see on log output from code sample from https://github.com/scrapy/scrapy/issues/6206#issuecomment-2180631979 (printing scraper slot active size every time when it's checked)
This type of.. tracking (by incrementing specific stats values +1 on each time when specific code lines reached) - will not be informative as engine checks backout conditions of scrapy components irregularly.

In this case it is preferred to track... time (in seconds) for what backout condition (scraper, downloader on something else) triggered and for how many seconds sending next requests to downloader was delayed.
With something like 





    clsss backout_timer:
       def start_or_continue_timer...
           # start time
       def drop_timer..
           ##add measured seconds to related crawler.stats values 
    ...
    def _needs_backout(self) -> bool:
        assert self.slot is not None  # typing
        assert self.scraper.slot is not None  # typing
        result = False
        if not self.running:
            backout_timed.drop_timer()
            return True
        if self.downloader.needs_backout():
            backout_timer.start_or_continue_timer('downloader')
            return True
        if self.scraper.slot.needs_backout()
            backout_timer.start_or_continue_timer('scraper')
            return True
        backout_timed.drop_timer()
        return False

scrapy/core/scraper.py
```diff
@@ -96,15 +182,19 @@ def finish_response(
...
     def is_idle(self) -> bool:
         return not (self.queue or self.active)
 
     def needs_backout(self) -> bool:
-        return self.active_size > self.max_active_size
+        warn(
+            "scrapy.core.scraper.Slot.needs_backout is deprecated.",
+            ScrapyDeprecationWarning,
+        )
+        return self._active_size > self._max_active_size
```
GeorgeA92, line 197: In a context of https://github.com/scrapy/scrapy/issues/482
Other important observation from log output from https://github.com/scrapy/scrapy/issues/6206#issuecomment-2180631979 is that response is active(it's size in `scraper.slot.active_size`) until all it's items passed pipelines (including awaiting for additional req/responses from media/files/images pipeline to be finished).
And by the moment when that items reach FeedExport (S3) - it's response is not counted in `scraper.slot.active_size` (and next request can be already sent without waiting to upload remaining items to S3 - which I assume required to be delayed here).

FeedExport(S3) are completely independent from scraper slot active size limits


In context of S3. It means that `scraper.slot._needs_backout`
https://github.com/scrapy/scrapy/blob/d8ecd28c5557e27f42e00bd1223b457468ea2ea7/scrapy/core/scraper.py#L106C1-L107C55
also needs to check that remaining items are exported (send to S3) 

and to be looked like 

```
    def needs_backout(self) -> bool:
        self.crawler.stats.get(')
        s3_stats_condition = #### check numbers of crawler.stats - feedexport(uploaded items) and crawler.stats (items)
        ## if item produced bud didn't uploaded to s3 yet - this numbers must be different
        return self.active_size > self.max_active_size or s3_stats_condition
```
Where `scraper.slot` class also need to have access to `crawler.stats`

Gallaecio: [The response is available in the context that calls item pipelines](https://github.com/scrapy/scrapy/blob/d8ecd28c5557e27f42e00bd1223b457468ea2ea7/scrapy/core/scraper.py#L375), so I think we are covered here (but tests should be written to make sure that does not change in the future accidentally).

Gallaecio:
> It is preferred to provide way to set scraper class by setting.

I’m not sure this is a good idea, as it is a very internal component, not really designed for override I think.

If the reason you think so is to allow users to have a way to override the backout logic, we could instead support defining a separate component for just that.

scrapy/core/scraper.py
GeorgeA92: My assumption is that root-cause of https://github.com/scrapy/scrapy/issues/482 is that application simultaneously (trying to) do:
1. scheduling new requests and processing responses (with RAM allocations to responses)
2. sending scraped items to S3 (with related RAM allocations)
Because associated response can be not active (in terms of scraper) when it's items reach S3 exporter.

If this theory is true then application needs to make only one of this at specific moment on time (which means that scraper needs to count work progress of S3 feed.. in it's backout condition)

Gallaecio: That’s correct. That’s the reason I say this is part 1 of a 2-part fix, where the second part would be to ensure the response objects are not garbage-collected before S3 finishes uploading. I’m moreover thinking that we could make that upload use response.body instead of a copy [as it does now](https://github.com/scrapy/scrapy/blob/d8ecd28c5557e27f42e00bd1223b457468ea2ea7/scrapy/pipelines/files.py#L676), making the uploads more memory efficient.

Gallaecio: > sending scraped items to S3

This is about the media pipeline, i.e. the problem is sending *responses* to S3, not *items*. But it is true that this PR does not fix that. That’s the reason I say this is part 1 of a 2-part fix, where the second part would be to ensure the response objects are not garbage-collected before S3 finishes uploading. I’m moreover thinking that we could make that upload use response.body instead of a copy [as it does now](https://github.com/scrapy/scrapy/blob/d8ecd28c5557e27f42e00bd1223b457468ea2ea7/scrapy/pipelines/files.py#L676), making the uploads more memory efficient.

scrapy/core/downloader/__init__.py
Gallaecio: [Done](https://github.com/scrapy/scrapy/pull/6430/commits/36cf156500c0f421546fdb9df97b99db16300e35).

GeorgeA92:
> I’m not sure this is a good idea, as it is a very internal component, not really designed for override I think.
> 
> If the reason you think so is to allow users to have a way to override the backout logic, we could instead support defining a separate component for just that.

Scraper is the only scrapy component that stores accumulated response objects before it's processing which make it an obvious choice for override on some issues.

The only notable differences between Scraper and Downloader/Scheduler(which is possible to override in settings) is that
1. Scraper is less documented comparing to others.. 
2. the fact that it doesn't have specific setting to set other Scraper class (like implemented for Downloader, Scheduler)

Recently users started to often make attention to Scraper in case of issues (particularly with RAM usage) even if it is not related to Scraper because of that 2 points mentioned above.

I can Identify at least 3 issues that as I think can be resolved with changes to Scraper (too much features to hold in single Scraper class as I think):

1. RequestLists https://github.com/scrapy/scrapy/issues/2600. Obviously Scraper is the only place where received responses stored. If scraper have only specific responses (not complete responses from request set).. then It I think can be implemented with additional manipulations with `Scraper.queue` and `Scraper.active`

2. Response processing order.
   In similar feature requests or questions in community channels we often suggest users to.. count asynchronous specifics of scrapy and suggest to additionally sort data after scraping on postprocessing stage. (while it can be.. too complex in case of millions of items) - I think it can be solved by additional manipulation with Scraper.queue (make it.. to work as priority queue...)


3. "Predictive" scraper.slot.active size.
   In fact scraper max_active_size restriction works "postfactum" e.g. on cases like multiple sitemaps with 50 000 links each like on https://github.com/scrapy/scrapy/issues/3529. 
   `Scraper.active_size` can easily exceed initial 5Mb limit to 10, 20 or even 50 megabytes (because at the moment when requests send - `scraper.active_size` was 0).
   If we know that that some responses will have some.. size we can send request with something like `expected_size: 1_000_000` in meta and (for ahead) add that number to `scraper.active_size` to prevent sending too much requests with "heavy responses" (as scraper backout condition will be triggered earlier)  - when application will receive response - scraper.active_size will need to be adjusted to it's real size - that will require to update Scraper (part that responsible to scraper.active_size calculation).

Previously.. options to subclass Scraper class.. didn't appear or was skipped because of that 2 points mentioned above

scrapy/core/downloader/__init__.py
GeorgeA92: I didn't expect fast feedback for this.
It 'd be good to make it possible to.. look into this on per minute basis from for example periodic_logstats extension.
If periodic_logstats extension used - it will show all backouts data except... active (because active didn't reflect on stats as it didn't finished). 
It may looks like:.. drop(write) backout timer and make another (the same reason) timer - called every 60 seconds interval (right before - periodic_log stats extension method) to get relevant backout data on per minute basis. 
Because we often have cases when.. some backout delay started and didn't finished.

Gallaecio:
This PR deprecated features from the scraper. If you think the scraper should retain any of those (`needs_backout`?), I am OK with restoring them. But I don’t think the discussion of whether or not to expose the Scraper is in the scope of this PR.

Gallaecio:
> "Predictive" scraper.slot.active size.

This might be worth implementing here. We could also use a default, to be assumed as the size of any request before there is a response for it. Similar to the scraper slot `MIN_RESPONSE_SIZE`, but only applied until the actual response size is known, instead of overriding the response size if lower.

scrapy/utils/defer.py
```diff
@@ -531,9 +531,14 @@ def _schedule_coro(coro: Coroutine[Any, Any, Any]) -> None:
...
     it.
+
+    In both asyncio and non-asyncio modes the coroutine starts in the next
+    event-loop iteration, never in the current call stack.
     """
     if not is_asyncio_available():
-        Deferred.fromCoroutine(coro)
+        from twisted.internet import reactor
+
+        reactor.callLater(0, Deferred.fromCoroutine, coro)
```
AdrianAtZyte: The prior difference in behavior was preventing 2 of my tests from working on the default reactor.

tests/test_engine.py
```diff
@@ -446,10 +442,11 @@ def test_start_already_running_exception(self):
         e = ExecutionEngine(crawler, lambda _: None)
         crawler.engine = e
         yield deferred_from_coro(e.open_spider_async())
-        _schedule_coro(e.start_async())
+        start_deferred = deferred_from_coro(e.start_async())
         with pytest.raises(RuntimeError, match="Engine already running"):
             yield deferred_from_coro(e.start_async())
         yield deferred_from_coro(e.stop_async())
+        yield start_deferred
```
AdrianAtZyte, line 138: Needed to accommodate https://github.com/scrapy/scrapy/pull/6430/changes#r3443520656, otherwise this would hang.

PR #6434, opened 2024-07-09
Feature/closespider pagecount no item

guillermo-bondonno:
Recently I had a use case where I needed to ensure that the crawler didn't perform too many requests without getting items in return. Even though `CLOSESPIDER_ERRORCOUNT` and `CLOSESPIDER_TIMEOUT_NO_ITEM` exist, they didn't cover all possible cases.
- `CLOSESPIDER_ERRORCOUNT`: Requests that don't raise errors but don't produce items.
- `CLOSESPIDER_TIMEOUT_NO_ITEM`: Maybe requests are slower or faster than expected, especially when rendering pages or performing complex actions. 

Would it be a good idea to add this as a possible setting? 

The changes add the possibility for a `CLOSESPIDER_PAGECOUNT_NO_ITEM: int` setting. Every `response_received` a counter is incremented and every `item_scraped` it's reset to 0. When the counter reaches or passes the setting value the spider is closed, since it crawled n pages with no items scraped.


scrapy/extensions/closespider.py
```diff
@@ -60,6 +64,19 @@ def __init__(self, crawler: Crawler):
...
+                signal=signals.response_received
+            ):
+                crawler.signals.connect(
+                    self.page_count, signal=signals.response_received
+                )
+
+            if self.item_scraped not in dispatcher.getReceivers(
+                signal=signals.item_scraped
+            ):
+                crawler.signals.connect(self.item_scraped, signal=signals.item_scraped)
```
Gallaecio, line 78: What if instead of using dispatcher you extend the `if` statement above?

e.g.

```python
        if self.close_on.get("pagecount") or self.close_on.get("pagecount_no_item"):
            crawler.signals.connect(self.page_count, signal=signals.response_received)
```

scrapy/extensions/closespider.py
```diff
@@ -74,9 +91,16 @@ def error_count(self, failure: Failure, response: Response, spider: Spider) -> N
 
     def page_count(self, response: Response, request: Request, spider: Spider) -> None:
         self.counter["pagecount"] += 1
+        self.counter["pagecount_since_last_item"] += 1
         if self.counter["pagecount"] == self.close_on["pagecount"]:
             assert self.crawler.engine
             self.crawler.engine.close_spider(spider, "closespider_pagecount")
```
Gallaecio, line 97: Maybe add an early return at the end of this `if`. No need to check if the condition below has been met if this one has.

tests/test_closespider.py
```diff
@@ -34,6 +34,19 @@ def test_closespider_pagecount(self):
...
+        close_on = 5
+        crawler = get_crawler(
+            FollowAllSpider,
+            {"CLOSESPIDER_PAGECOUNT_NO_ITEM": close_on},
+        )
+        yield crawler.crawl(mockserver=self.mockserver)
+        reason = crawler.spider.meta["close_reason"]
+        self.assertEqual(reason, "closespider_pagecount_no_item")
+        pagecount = crawler.stats.get_value("response_received_count")
+        self.assertTrue(pagecount >= close_on)
```
Gallaecio, line 48: This does not test what makes this feature different from `CLOSESPIDER_PAGECOUNT`, though. Maybe you could implement an alternative to `FollowAllSpider` that allows testing that difference, i.e. build a test that would fail with `CLOSESPIDER_PAGECOUNT`.

docs/topics/extensions.rst
```diff
@@ -317,6 +317,18 @@ crawls more than that, the spider will be closed with the reason
 ``closespider_pagecount``. If zero (or non set), spiders won't be closed by
 number of crawled responses.
 
+.. setting:: CLOSESPIDER_PAGECOUNT_NO_ITEM
+
+CLOSESPIDER_PAGECOUNT_NO_ITEM
+"""""""""""""""""""""
```
Gallaecio, line 323: This needs to be the same length as the underlined text.

Gallaecio (approved):
Nice!

guillermo-bondonno:
The test is failing in some cases, I'll take a look

Gallaecio:
It seems the number of pages visited can be lower than 10. Although I do not fully understand why you are checking that.

Maybe you should make a more deterministic spider that:
1. Follows 4 pages
2. Yields 1 item
3. Follows 6 pages (verifying the spider close reason and total page count at 9 instead of 10 should cover all the logic, I think)

e.g. (untested)

```python
class CloseSpiderPageCountNoItemSpider(Spider):
    name = "closespider_pagecount_no_item"
    start_urls = ["data:,01"]

        self.items_scraped = 0

    def parse(self, response):
        page = int(response.url[-1])
        if page < 4:
            yield response.follow(f"data:,{page+1:02}")
            return
        yield {}
        if page < 10:
            yield response.follow(f"data:,{page+1:02}")
```

Gallaecio (approved):

docs/topics/extensions.rst
```diff
@@ -317,6 +317,18 @@ crawls more than that, the spider will be closed with the reason
...
+
+CLOSESPIDER_PAGECOUNT_NO_ITEM
+"""""""""""""""""""""""""""""
+
+Default: ``0``
+
+An integer which specifies the maximum number of consecutive responses to crawl without items scraped. If the spider
+crawls more consecutive responses than that and no items are scraped in the meantime, the spider will be closed with the reason
+``closespider_pagecount_no_item``. If zero (or non set), spiders won't be closed by
+number of crawled responses with no items.
```
wRAR, line 330:
```suggestion
An integer which specifies the maximum number of consecutive responses to crawl
without items scraped. If the spider crawls more consecutive responses than that
and no items are scraped in the meantime, the spider will be closed with the
reason ``closespider_pagecount_no_item``. If zero (or not set), spiders won't be
closed by number of crawled responses with no items.
```

wRAR (approved):

PR #6438, opened 2024-07-17
Deprecate FormRequest in favor of form2request

Gallaecio:
To do:
- [x] Write a section on using form2request.
- [x] Remove `FormRequest` from the docs.
- [x] Review remaining appearances of `FormRequest`.


kmike:
Hm, that's an interesting approach. Initially I was thinking about keeping the FormRequest class, but replacing the implementation with form2request calls. Is it behavior incompatibility between form2request and current FormRequest which makes this hard? Or are you more concerned about having 2 APIs which do a similar thing?

Gallaecio:
It was more about the 2 APIs, and specially my dislike of the current `FormRequest` API and some of its default parameters. But I can open a second PR for a less disruptive approach.

iandees:
It looks like `form2request` is for building a `Request` by parsing an HTML form. Over in alltheplaces, we use `FormRequest` as a shortcut to produce `Request`s that mimic a form request without necessarily parsing the form elements in the first place. What's the migration path for this usecase?

For example, our code might have:

```python
yield FormRequest(
    url="https://api.example.com/search",
    formdata={"query": "test", "page": "1"}
)
```

which now needs to be translated to

```python
from urllib.parse import urlencode

body = urlencode({"query": "test", "page": "1"}).encode('utf-8')
yield Request(
    url="https://api.example.com/search",
    method="POST",
    body=body,
    headers={'Content-Type': 'application/x-www-form-urlencoded'}
)
```

PR #6431, opened 2024-07-06
Documentation edits

mlmsmith:
Made some minor changes to the tutorial section of the documentation.

PR #6440, opened 2024-07-19
Tutorial edits

mlmsmith:
Reverted unwarranted changes from earlier PR and made additional minor grammar adjustments to the rest of the section.

### DIFF
--- a/docs/intro/tutorial.rst
+++ b/docs/intro/tutorial.rst
@@ -77,10 +77,9 @@ This will create a ``tutorial`` directory with the following contents::
 Our first Spider
 ================
 
-Spiders are classes that you define and that Scrapy uses to scrape information
-from a website (or a group of websites). They must subclass
-:class:`~scrapy.Spider` and define the initial requests to be made.
-They can also optionally define how to follow page links and how to parse the downloaded
+Spiders are classes that you define and that Scrapy uses to scrape information from a website
+(or a group of websites). They must subclass :class:`~scrapy.Spider` and define the initial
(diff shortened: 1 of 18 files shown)


docs/intro/tutorial.rst
```diff
@@ -18,11 +18,12 @@ This tutorial will walk you through these tasks:
 4. Changing spider to recursively follow links
 5. Using spider arguments
 
-Scrapy is written in Python_. If you're new to the language you might want to
-start by getting an idea of what the language is like, to get the most out of
+Scrapy is written in Python_. If you're new to the language, you might want to
+start by getting an idea of what the language is like to get the most out of
 Scrapy.
```
Gallaecio, line 23: I need an explanation for this change beyond the fact that a grammar tool suggested it. I am not opposed to the new comma, but removing the other one seems wrong to me.

mlmsmith: A comma should follow a subordinate clause; in this case, “If you’re new to the language”. When an independent clause is followed by a subordinate clause, you don’t need to use a comma, making the one between “like” and “to” incorrect.

wRAR:
Does this obsolete #6431?

docs/intro/tutorial.rst
Gallaecio: What if we replace the entire sentence with a shorter one?

> Scrapy is written in Python. The more you learn about Python, the more you can get out of Scrapy.

mlmsmith: Yes, I think that should work.

Gallaecio (approved):
Thanks!

PR #6441, opened 2024-07-20
Add @property usage of custom_settings

darwin403:
**Context**

It is often useful to modify some settings in the `custom_settings` of a spider such that they can later be accessed in the crawler like:

```python
@property
def custom_settings(self):
 return FEEDS = {
   f"output/{self.name}_output.jl" : {"overwrite": True}
  }
```
 

Currently there is no way to access a dynamic FEEDS output file path in the spider itself. Consider the following FEEDS in custom settings:

```python
FEEDS = {
 "output/%(name)s_output.jl" : {"overwrite": True}
}
```

The paramters for the crawler name are populated by the FeedExporter extension are not accesible in the spider itself. Therefore if I want to perform some clean up or post scraping operations through my Spider class, I cannot do so.

**What's next?**

Please let me know if this is acceptable idea and I'll the docs and tests whereever necessary.

### DIFF
--- a/scrapy/spiders/__init__.py
+++ b/scrapy/spiders/__init__.py
@@ -94,7 +94,12 @@ class Spider(object_ref):
 
     @classmethod
     def update_settings(cls, settings: BaseSettings) -> None:
-        settings.setdict(cls.custom_settings or {}, priority="spider")
+        if isinstance(cls.custom_settings, property):
+            custom_settings = cls.custom_settings()
+        else:
+            custom_settings = cls.custom_settings
+
+        settings.setdict(custom_settings or {}, priority="spider")
 
     @classmethod
     def handles_request(cls, request: Request) -> bool:


wRAR:
I couldn't understand both the stated problem and the proposed solution, sorry. Can you maybe provide some code that you think isn't possible without this change and possible with it?

Gallaecio:
If you want to have access to self while configuring a setting from a spider, use the `update_settings` method instead of the `custom_attributes` class variable.

darwin403:
> If you want to have access to self while configuring a setting from a spider, [use the `update_settings` method instead of the `custom_attributes` class variable](https://docs.scrapy.org/en/latest/topics/settings.html#settings-per-spider).

Yep, I'm currently using this approach like you can see here: https://github.com/desiquant/news_scraper/blob/b0e9618d187e46feb77bfa6cd6daaac9b434100e/news_scraper/spiders/base.py#L27

I thought it would be more convinient to define it as a @property, since it's also gives access to the spider instance with `self`

Gallaecio:
That would require quite some rework and break a lot of existing code. At the moment the settings are customized before the spider is created, if we change that then `Spider.__init__` would happen before settings are final, which I think it rather problematic.

So honestly I don’t see this happening.

darwin403:
> That would require quite some rework and break a lot of existing code. At the moment the settings are customized before the spider is created, if we change that then `Spider.__init__` would happen before settings are final, which I think it rather problematic.
> 
> We could go for a classproperty, but that’s not something Python supports natively, and it would complicate things. It would also be a backward-incompatible change, unless we use a new name, in which case you end up with 3 different ways to define settings at the spider level.
> 
> So honestly I don’t see this happening.

OK, thanks for the feedback!

PR #6444, opened 2024-07-26
Fix typo in title of contracts.rst

martin-martin:
It looks like the title of this documentation page should be "Spider Contracts", considering that's also the name of the relevant setting mentioned on the same page:

```
SPIDER_CONTRACTS = {
    "myproject.contracts.ResponseCheck": 10,
    "myproject.contracts.ItemValidate": 10,
}
```

Hope that makes sense. If "Spiders Contracts" is indeed the correct term, maybe you could add some info as to why. Thanks!

### DIFF
--- a/docs/topics/contracts.rst
+++ b/docs/topics/contracts.rst
@@ -1,7 +1,7 @@
 .. _topics-contracts:
 
 =================
-Spiders Contracts
+Spider Contracts
 =================
 
 Testing spiders can get particularly annoying and while nothing prevents you


docs/topics/contracts.rst
```diff
@@ -1,7 +1,7 @@
 .. _topics-contracts:
 
 =================
-Spiders Contracts
+Spider Contracts
 =================
```
Gallaecio, line 5: Please, also edit the `===…` before and after accordingly, i.e. remove 1 `=` from each.

martin-martin:
Thanks for spotting that @Gallaecio , removed the two superfluous `=`. ✅

Gallaecio (approved):

Gallaecio:
Thanks!

PR #6448, opened 2024-07-31
Do not suggest logging dropped items twice

Gallaecio:
[Found](https://github.com/zytedata/duplicate-url-discarder/pull/23#issuecomment-2259413875) by @BurnzZ.

### DIFF
--- a/docs/topics/item-pipeline.rst
+++ b/docs/topics/item-pipeline.rst
@@ -99,7 +99,7 @@ contain a price:
                     adapter["price"] = adapter["price"] * self.vat_factor
                 return item
             else:
-                raise DropItem(f"Missing price in {item}")
+                raise DropItem("Missing price")
 
 
 Write items to a JSON lines file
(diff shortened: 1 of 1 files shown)


docs/topics/item-pipeline.rst
```diff
@@ -254,7 +254,7 @@ returns multiples items with the same id:
         def process_item(self, item, spider):
             adapter = ItemAdapter(item)
             if adapter["id"] in self.ids_seen:
-                raise DropItem(f"Duplicate item found: {item!r}")
+                raise DropItem(f"Item ID already seen")
```
BurnzZ, line 257:
```suggestion
                raise DropItem(f"Item ID already seen: {adapter['id']}")
```

or simply remove the f-string: `raise DropItem("Item ID already seen")`

BurnzZ: Could we also update the similar one in **line 102**? Cheers!

BurnzZ (changes requested):
Thanks @

PR #6457, opened 2024-08-08
Option to include all tags and attrs in LinkExtractor with specified exclusions

### LINKED ISSUE #6321: Option to include all tags and attrs in LinkExtractor with specified exclusions
## Summary

## Motivation

## Additional context

jith912:
Fixes #6321 

**Requirement summary**
As of now, the tags and attributes are passed in via the parameters tags, attrs to the constructor of the class LxmlLinkExtractor. These are used only in _inclusion_ mode - meaning the extractor will return the tags and the attrs that contain these values. 
The new requirement is to add a capability to exclude certain tags and attributes.

**Approach**
The idea is to introduce the concept of modes. 

1. Inclusion mode - in this mode, the tags and attrs values that are passed in will be _included_ in the response
2. Exclusion mode - in this mode, the tags and attrs that are passed in will be _filtered out_ from the response

To achieve the concept of modes, two new variables are introduced to LxmlLinkExtractor class.

1. tag_inclusion_mode - this indicates the mode for the tags
2. attr_inclusion_mode - this indicates the mode for the attrs

These are boolean variables. True indicates inclusion , False indicates exclusion

For example, lets say the user wants to include the tag "div" and exclude the attribute "href". The user has to pass in 

1. tag_inclusion_mode = True 
2. attr_inclusion_mode = False.
4. tags = "div"
5. attrs = "href"

Please refer to the complete behaviour illustrated with the help of test cases [here](https://github.com/jith912/scrapy/wiki/Tests-for-LinkExtractor-Mode)

Note: I have added the test case numbers in the PRs just for my convenience. Will remove it at the point of merging the PR

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -64,6 +64,15 @@ def _canonicalize_link_url(link: Link) -> str:
     return canonicalize_url(link.url, keep_fragments=True)
 
 
+def _is_item_included(sourceList, inclusion, item):
+    if (
+        (not sourceList and not inclusion)
+        or (item in sourceList and inclusion)
+        or (item not in sourceList and not inclusion)
+    ):
+        return True
+
+
 class LxmlParserLinkExtractor:
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


scrapy/linkextractors/lxmlhtml.py
```diff
@@ -91,8 +107,10 @@ def __init__(
 
     def _iter_links(
         self, document: HtmlElement
-    ) -> Iterable[tuple[HtmlElement, str, str]]:
+    ) -> Iterable[Tuple[HtmlElement, str, str]]:
         for el in document.iter(etree.Element):
+
+            # if not self.scan_tag(_nons(el.tag)):
```
AdrianAtZyte, line 113: Debug leftover?

AdrianAtZyte (commented):
Looks good to me. I think with docs this would be mergeable.

PR #6460, opened 2024-08-11
Typing fixes and updates.

wRAR:


scrapy/pipelines/media.py
```diff
@@ -211,7 +211,7 @@ def _cache_result_and_execute_waiters(
             # minimize cached information for failure
             result.cleanFailure()
             result.frames = []
-            result.stack = None
+            result.stack = []
```
wRAR, line 214: This started producing a mypy error with Twisted 24.7.0 so we likely need to copy this change to the 2.11 branch.

Gallaecio (approved):

PR #6461, opened 2024-08-12
Fix cookie selection for cookies with the same name but different paths.

0xdeb:
This PR fixes the cookie selection in the `CookieJar.add_cookie_header` method of `scrapy.http.cookies`. 

In one of my scrapers I observed an cookie being sent twice in the `Cookies` header, e.g.:
```
Cookie: SOMECOOKIE=somevalue; OTHERCOOKIE=onevalue; OTHERCOOKIE=differentvalue
```

After having a look at scrapy's code I found out, that the cookie selection was not as described in RFC 6265. 

Section 5.4 (https://datatracker.ietf.org/doc/html/rfc6265#section-5.4) describes how the header should be assembled. 

Let's assume we have a cookie named `TEST`, that is present in the cookie jar under a path of `/` and `/bar`. 
Before the fix scrapy would send the cookie twice with both values for a request to `/bar/`, because the path would match both cookie's path. To follow the spec scrapy needs to select the cookie with the longest matching path and only return that one. 

I also covered the change with unit tests. Let me know if you want me to change anything!

<details>
  <summary>Newly created test failing before the fix</summary>
 
  ```
=================================================================================== FAILURES ===================================================================================
_____________________________________________________________ CookiesMiddlewareTest.test_cookie_selection_by_path ______________________________________________________________

self = <tests.test_downloadermiddleware_cookies.CookiesMiddlewareTest testMethod=test_cookie_selection_by_path>

    def test_cookie_selection_by_path(self):
        cookies = [
            {
                "name": "TESTNAME",
                "value": "value1",
                "path": "/",
                "domain": "scrapytest.org",
            },
            {
                "name": "TESTNAME",
                "value": "value2",
                "path": "/bar",
                "domain": "scrapytest.org",
            },
        ]

        req = Request("https://scrapytest.org/", cookies=cookies)
        self.mw.process_request(req, self.spider)

        # embed C1 and C3 for scrapytest.org/foo
        req = Request("https://scrapytest.org/")
        self.mw.process_request(req, self.spider)
        assert req.headers.get("Cookie") == b"TESTNAME=value1"

        # embed C2 for scrapytest.org/bar
        req = Request("https://scrapytest.org/bar")
        self.mw.process_request(req, self.spider)
>       self.assertEqual(req.headers.get("Cookie"), b"TESTNAME=value2")
E       AssertionError: b'TESTNAME=value2; TESTNAME=value1' != b'TESTNAME=value2'

/home/deb/work/qyobo/repos/scrapy/tests/test_downloadermiddleware_cookies.py:268: AssertionError

  ```
  
</details>

### DIFF
--- a/scrapy/http/cookies.py
+++ b/scrapy/http/cookies.py
@@ -72,7 +72,23 @@ class CookieJar:
             if host in self.jar._cookies:  # type: ignore[attr-defined]
                 cookies += self.jar._cookies_for_domain(host, wreq)  # type: ignore[attr-defined]
 
-        attrs = self.jar._cookie_attrs(cookies)  # type: ignore[attr-defined]
+        # cookie selection for given path in request according to RFC6265 Section 5.4:
+        # -> _cookies_for_domain returns all cookies matching domain and path
+        #     -> there can be multiple cookies with the same name but different paths
+        # -> select the one with the longest path for each name
+        # https://datatracker.ietf.org/doc/html/rfc6265#section-5.4
+        cookies_by_name: dict[str, list[Cookie]] = {}
+        for cookie in cookies:
+            if cookie.name not in cookies_by_name:
+                cookies_by_name[cookie.name] = []
+            cookies_by_name[cookie.name].append(cookie)
+
+        deduplicated_cookies = [
+            sorted(cookie_list, key=lambda c: len(c.path), reverse=True)[0]
+            for name, cookie_list in cookies_by_name.items()
+        ]
+
+        attrs = self.jar._cookie_attrs(deduplicated_cookies)  # type: ignore[attr-defined]
         if attrs:
             if not wreq.has_header("Cookie"):
                 wreq.add_unredirected_header("Cookie", "; ".join(attrs))
--- a/tests/test_downloadermiddleware_cookies.py
+++ b/tests/test_downloadermiddleware_cookies.py
@@ -238,6 +238,35 @@ class CookiesMiddlewareTest(TestCase):
         self.mw.process_request(req, self.spider)
         assert "Cookie" not in req.headers
 
+    def test_cookie_selection_by_path(self):
+        cookies = [
+            {
+                "name": "TESTNAME",
+                "value": "value1",
+                "path": "/",
+                "domain": "scrapytest.org",
+            },
+            {
+                "name": "TESTNAME",
+                "value": "value2",
+                "path": "/bar",
+                "domain": "scrapytest.org",
+            },
+        ]
+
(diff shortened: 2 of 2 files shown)


0xdeb:
The failed typing checks should be unrelated to this MR. See https://github.com/scrapy/scrapy/pull/6460

0xdeb:
@Gallaecio @wRAR 
I rebased the branch on current master. All checks pass now. Please check and consider merging.

scrapy/http/cookies.py
```diff
@@ -63,27 +72,44 @@ def add_cookie_header(self, request: Request) -> None:
             if host in self.jar._cookies:  # type: ignore[attr-defined]
                 cookies += self.jar._cookies_for_domain(host, wreq)  # type: ignore[attr-defined]
 
-        attrs = self.jar._cookie_attrs(cookies)  # type: ignore[attr-defined]
-        if attrs and not wreq.has_header("Cookie"):
-            wreq.add_unredirected_header("Cookie", "; ".join(attrs))
+        # cookie selection for given path in request according to RFC6265 Section 5.4:
+        # -> _cookies_for_domain returns all cookies matching domain and path
+        #     -> there can be multiple cookies with the same name but different paths
+        # -> select the one with the longest path for each name
```
AdrianAtZyte, line 78: As far as I can tell, the standard does says to sort by path, but it never says to drop duplicates. Am I missing something?

StantonMatt (commented):
I ran the focused test locally and it passes:

```console
tox -e py -- tests/test_downloadermiddleware_cookies.py -q -k cookie_selection_by_path
1 passed, 40 deselected
```

I think Adrian's duplicate-cookie concern is real, though. RFC 6265 section 5.4 requires matching cookies to be ordered by longer path first before the `Cookie` header is serialized; it does not say to collapse cookies that share the same name. With this branch, Scrapy now drops the shorter-path cookie when both match:

```python
cookies = [
    {"name": "TESTNAME", "value": "root", "path": "/", "domain": "scrapytest.org"},
    {"name": "TESTNAME", "value": "bar", "path": "/bar", "domain": "scrapytest.org"},
]

mw.process_request(Request("https://scrapytest.org/", cookies=cookies), spider)
mw.process_request(req := Request("https://scrapytest.org/bar"), spider)
print(req.headers.get("Cookie"))
```

On current `master`, that prints:

```text
b'TESTNAME=bar; TESTNAME=root'
```

On this branch, it prints:

```text
b'TESTNAME=bar'
```

So I think the fix should sort all matching cookies by descending path length and keep both entries, rather than deduplicating by cookie name. That would preserve the RFC ordering behavior while still fixing the path precedence issue.

PR #6464, opened 2024-08-16
Fix comment capitalization for consistency

shlaibah:
Capitalized first letter of sentence


shlaibah:
Hello! This is my first commit, so I would like to learn why the change was not merged. I do not care about it not being merged, I would just like to know what I could do better. Thank you for your guidance!

wRAR:
It's an unnecessary change.

PR #6469, opened 2024-08-27
Add support for meta in Spider Contracts

### LINKED ISSUE #6468: Add support for meta in Spider Contracts
## Summary

Today we support `cb_kwargs` but we have scenarios where the data is passed using `meta`.

## Motivation

I have some spiders that use `meta` to store information, e.g: `cookiejar`

Use callback args.

Laerte:
close #6468

### DIFF
--- a/docs/topics/contracts.rst
+++ b/docs/topics/contracts.rst
@@ -46,6 +46,14 @@ This callback is tested using three built-in contracts:
 
     @cb_kwargs {"arg1": "value1", "arg2": "value2", ...}
 
+.. class:: MetadataContract
+
+    This contract (``@meta``) sets the :attr:` meta <scrapy.Request.meta>`
+    attribute for the sample request. It must be a valid JSON dictionary.
+    ::
+
+    @meta {"arg1": "value1", "arg2": "value2", ...}
+
 .. class:: ReturnsContract
 
     This contract (``@returns``) sets lower and upper bounds for the items and
--- a/docs/topics/contracts.rst
+++ b/docs/topics/contracts.rst
@@ -20,13 +20,13 @@ following example:
         This function parses a sample response. Some contracts are mingled
         with this docstring.
 
-        @url http://www.amazon.com/s?field-keywords=selfish+gene
+        @url http://www.example.com/s?field-keywords=selfish+gene
         @returns items 1 16
         @returns requests 0 0
         @scrapes Title Author Year Price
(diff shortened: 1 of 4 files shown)


docs/topics/contracts.rst
```diff
@@ -46,6 +46,14 @@ This callback is tested using three built-in contracts:
 
     @cb_kwargs {"arg1": "value1", "arg2": "value2", ...}
 
+.. class:: MetadataContract
```
Laerte, line 49: Above in this file we have:

https://github.com/scrapy/scrapy/blob/6ce0342beb1a5b588f353e52fe03d5e0ec84d938/docs/topics/contracts.rst?plain=1#L29

I'm not sure if this statement is correct since we just just are listing the default contracts and `cb_kwargs` is not used in the snippet:

https://github.com/scrapy/scrapy/blob/6ce0342beb1a5b588f353e52fe03d5e0ec84d938/docs/topics/contracts.rst?plain=1#L23-L26

I followed the pattern and added the `MetadataContract` to the list.

docs/topics/contracts.rst
```diff
@@ -46,6 +46,14 @@ This callback is tested using three built-in contracts:
...
+
+    This contract (``@meta``) sets the :attr:` meta <scrapy.Request.meta>`
```
Laerte, line 51:
```suggestion
    This contract (``@meta``) sets the :attr:`meta <scrapy.Request.meta>`
```

docs/topics/contracts.rst
Gallaecio: I would reword that to:

> You can use the following contracts:

Also, I think we should replace amazon.com with example.com.

Gallaecio (approved):

docs/topics/contracts.rst
Laerte: Done, thanks.

scrapy/contracts/default.py
```diff
@@ -35,6 +35,20 @@ def adjust_request_args(self, args: Dict[str, Any]) -> Dict[str, Any]:
         return args
 
 
+class MetadataContract(Contract):
+    """Contract to key metadata arguments for the request.
```
Laerte, line 39:
```suggestion
    """Contract to set metadata arguments for the request.
```

wRAR (approved):

PR #6470, opened 2024-08-27
Refactor genspider slightly so template variables can be overridden

CloCkWeRX:


wRAR:
How are you going to use this?

CloCkWeRX:
So; for alltheplaces we have a large number of spiders - many of them fit a common pattern. We have ~20+ patterns, and do a lot of autodiscovery from wikidata or the name_suggestion_index to annotate the spider/data.

See https://github.com/alltheplaces/alltheplaces/pull/9849 for example.

Right now I need to basically extend and re-implement 90% of the code for generating a specific spider template.
If this PR were to land, it would mean I only need to define my own list of variables for my templates.

wRAR:
If you are doing this to still subclass the command, are you fine with this being an underscored function that can be silently removed or changed?

CloCkWeRX:
Well, ideally this would have a slightly more defined interface. But for now, I will take it.

CloCkWeRX:
(happy to refactor further; add docs; etc)

wRAR (approved):

PR #6472, opened 2024-09-01
Drop Python 3.8 Support

### LINKED ISSUE #6466: Dropping Python 3.8 support
3.8 support ends on 31 Oct, so I'm thinking about changes we can do after that. Looks like they are mostly related to typing.

* Drop Reppy which doesn't support Python 3.9+: #5226
* Adopt [PEP-585](https://peps.python.org/pep-0585/): 
  * Switch e.g. `List[T]` to `list[T]`
  * Switch `typing.Foo` imports to `collections.abc.Foo` ones
  * Maybe remove `--keep-runtime-typing` from the `pyupgrade` args, but that also changes unions to the `|` syntax in files with `from __future__ import annotations` and we would want to add those to all files with unions to keep the style consistent
* Bump the target version for `pyupgrade`

I don't see `typing_extensions` imports that were added in 3.9 (most of them are 3.10).

It's possible to do the `--keep-runtime-typing` change in advance to reduce the diff.

vsevolodbreus:
Closes #6466


scrapy/commands/startproject.py
```diff
@@ -107,8 +107,6 @@ def run(self, args: List[str], opts: argparse.Namespace) -> None:
             return
 
         self._copytree(Path(self.templates_dir), project_dir.resolve())
-        # On 3.8 shutil.move doesn't fully support Path args, but it supports our use case
-        # See https://bugs.python.org/issue32689
         move(project_dir / "module", project_dir / project_name)  # type: ignore[arg-type]
```
wRAR, line 110: You need to remove the comment too.

tox.ini
```diff
@@ -109,7 +109,7 @@ deps =
     Twisted[http2]==18.9.0
     w3lib==1.17.0
     zope.interface==5.1.0
-    lxml==4.4.1
+    lxml==4.6.0
```
wRAR, line 112: This should be synced with version restrictions in setup.py.

scrapy/commands/startproject.py
vsevolodbreus: Done.

tox.ini
vsevolodbreus: Done.

scrapy/utils/defer.py
```diff
@@ -145,7 +130,7 @@ def parallel(
     return DeferredList([coop.coiterate(work) for _ in range(count)])
 
 
-class _AsyncCooperatorAdapter(Iterator[Deferred], Generic[_T]):
+class _AsyncCooperatorAdapter(Iterator[Deferred], Generic[_T, _SelfResultT]):
```
wRAR, line 133: This causes test failures, why is this needed?

setup.py
```diff
@@ -6,7 +6,7 @@
 
 
 install_requires = [
-    "Twisted>=18.9.0",
+    "Twisted>=20.3.0",
```
wRAR, line 9: What is the reason for this bump?

scrapy/utils/trackref.py
```diff
@@ -9,27 +9,28 @@
...
+live_refs: defaultdict[type, WeakKeyDictionary] = defaultdict(WeakKeyDictionary)
 
 
 class object_ref:
     """Inherit from this class to a keep a record of live instances"""
 
     __slots__ = ()
 
-    def __new__(cls, *args: Any, **kwargs: Any) -> "Self":
+    def __new__(cls, *args: Any, **kwargs: Any) -> object_ref:
```
wRAR, line 33: I don't this is correct, `object.__new__(cls)` returns a `cls` instance.

setup.py
vsevolodbreus: Windows build is failing for 18.9.0. Trying to determine whether this is a version issue.

scrapy/utils/defer.py
vsevolodbreus: See full error stack trace here: https://github.com/vsevolodbreus/scrapy/actions/runs/10700184494/job/29663370539

scrapy/utils/trackref.py
vsevolodbreus: Reverted. But with quotes removed. I got an error message about "Self" being a string literal.

vsevolodbreus:
I am getting a lot of: "TypeError: 'type' object is not subscriptable" errors for python3.9 tests, which doesn't make sense, since type[type] (ex. list[str]) is valid syntax starting in Python 3.9
Thoughts or suggestions?

wRAR:
It's about `Deferred` which is not generic in the pinned Twisted version. It should be enough to add `from __future__ import annotations` to the affected files.

scrapy/utils/defer.py
wRAR: Looks like it was fixed in some 3.10.x (it works on 3.10.14 and doesn't work on 3.10.5, maybe it's related to "gh-98852: Fix subscription of types.GenericAlias instances containing bare generic types: for example tuple[A, T][int], where A is a generic type, and T is a type variable." from 3.10.9) and 3.11+ and wasn't fixed in 3.9. And it looks like `typing.Iterator` works here so there is a workaround.

setup.py
wRAR: I checked this and yeah, iocpsupport from Twisted < 20.3.0 fails to build on Python 3.8 so this makes sense.

wRAR:
Great work!

wRAR (approved):

PR #6474, opened 2024-09-06
Fix Twisted deprecation warning by using A twisted.web.http.HTTPClient with twisted.web.client.Agentgent instead of HTTPClient

smritisubh:


smritisubh:
twisted.web.http.HTTPClient with twisted.web.client.Agent

wRAR:
Is this from ChatGPT?

smritisubh:
> Is this from ChatGPT?

No

wRAR:
What generated it then?

PR #6480, opened 2024-09-22
Fix bench server

LucasSD:
- Resolved import conflicts to ensure the bench server works properly.
- Added unit tests for `scrapy/utils/benchserver.py` to provide test coverage.

I came across the bench server and found it wasn't functioning. It seems it might not be used often, but these changes resolve the issue and add test coverage.

If there are specific parts of the codebase that would benefit from additional unit tests, I'd appreciate any pointers or suggestions 👍


wRAR:
Nice, though I'm not sure about renaming public modules.

LucasSD:
Thanks @wRAR ! Just to confirm, is renaming `_compression.py` acceptable since it is denoted as non-public by the leading underscore?

I’ve tried some alternatives, including import aliasing, relative and absolute import paths, but haven’t yet found a viable alternative to simply renaming `signal.py`.

Other options:

1. Proceed with this PR as is, while retaining the original `signal.py` as a compatibility layer. This could help mitigate confusion caused by the name conflict with Python’s built-in `signal` module.
2. Implement lazy imports in multiple places to address the import issues with `signal.py`.

Let me know what you prefer. Happy to explore other suggestions if you have any.

Gallaecio:
We usually do something like 1, having the old module trigger a deprecation warning, and then we remove the module after 1 year.

scrapy/utils/signal.py
```diff
@@ -1,123 +1,39 @@
...
+        return disconnect_all(*args, **kwargs)
 
 
-def disconnect_all(signal: TypingAny = Any, sender: TypingAny = Any) -> None:
-    """Disconnect all signal handlers. Useful for cleaning up after running
-    tests
-    """
-    for receiver in liveReceivers(getAllReceivers(sender, signal)):
-        disconnect(receiver, signal=signal, sender=sender)
+signal = SignalDeprecated()
```
wRAR, line 39: This would be `scrapy.utils.signal.signal`

LucasSD: Is my latest change to `signal.py` acceptable for maintaining backwards compatibility?

I noticed a test failure in a specific environment. Do you have any recommendations for fixing it? I was thinking we might need to `import signal as builtin_signal` in `test_crawler.py`. Alternatively, would increasing the delay at line 909 help? Or perhaps both?

LucasSD: Hi @wRAR @Gallaecio , I noticed that the test that previously failed has now passed after being rerun. Just checking in to see if there's anything else you'd like me to address?

scrapy/utils/signal.py
```diff
@@ -1,118 +1,37 @@
...
-from twisted.internet.defer import Deferred, DeferredList
-from twisted.python.failure import Failure
-
-from scrapy.exceptions import StopDownload
-from scrapy.utils.defer import maybeDeferred_coro
-from scrapy.utils.log import failure_to_exc_info
-
-logger = logging.getLogger(__name__)
+def send_catch_log(*args, **kwargs):
+    warnings.warn(
```
wRAR, line 5: I think instead of doing this in every function you can do this at the top level like in https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/offsite.py and just re=export functions from `signal_ops` instead of calling them.

wRAR:
Can you please explain which circular import(s) is fixed by this PR?

LucasSD:
> Can you please explain which circular import(s) is fixed by this PR?

Thanks for flagging that. It was incorrect of me to mention circular imports. The issue stems from some Scrapy files sharing the same names as Python built-in modules, which leads to import conflicts. You can reproduce this by attempting to run the bench server on the master branch. I've updated the PR description

wRAR:
> The issue stems from some Scrapy files sharing the same names as Python built-in modules, which leads to import conflicts

That would explain removing `signal`, but it's no longer removed.

> You can reproduce this by attempting to run the bench server on the master branch.

How should I run it? Surely not with `python3 scrapy/utils/benchserver.py`?

LucasSD:
> How should I run it? Surely not with `python3 scrapy/utils/benchserver.py`?

Running the bench server locally with `python3 scrapy/utils/benchserver.py` highlights conflicts between Scrapy's `_compression.py` and `signal.py` files and Python's built-in `_compression` and `signal` modules. Perhaps this isn't the typical way the bench server is run, but these name conflicts could potentially affect other users or scenarios where these modules are involved, even beyond the bench server.

If this method of running the bench server isn't relevant or the name conflicts aren't considered problematic, I'm happy to adjust this PR to retain only the added tests.

Let me know what you think's best.

wRAR:
Did you run that with your changes?

wRAR:
In general, as we run that code with `python3 -m scrapy.utils.benchserver` we don't have these problems, and I'm not sure if name clashes are relevant unless you add their paths to sys.path, which you normall shouldn't.

LucasSD:
With the current changes in this branch:

- `python scrapy/utils/benchserver.py` successfully runs the bench server, though it raises a non-breaking `AttributeError` in the terminal.
- `python -m scrapy.utils.benchserver` works without any issues, just like on the master branch.

Given that you mentioned the potential name conflicts are not critical, would you prefer that I retain only the added tests (`test_utils_benchserver.py`) in this PR?

wRAR:
> python scrapy/utils/benchserver.py successfully runs the bench server, though it raises a non-breaking AttributeError in the terminal.

I didn't notice it was non-breaking (but still).

> Given that you mentioned the potential name conflicts are not critical, would you prefer that I retain only the added tests (test_utils_benchserver.py) in this PR?

I would prefer to not have code changes that only fix `python scrapy/utils/benchserver.py`. I would also prefer to not have most of these tests either, as most of them look excessive to me (are they autogenerated?). We already test that `scrapy bench` (which runs `python -m scrapy.utils.benchserver`) works in `tests/test_commands.py` and I'm not sure which parts of `benchserver.py` would benefit from unit tests.

LucasSD:
> I would prefer to not have code changes that only fix `python scrapy/utils/benchserver.py`. I would also prefer to not have most of these tests either, as most of them look excessive to me (are they autogenerated?). We already test that `scrapy bench` (which runs `python -m scrapy.utils.benchserver`) works in `tests/test_commands.py` and I'm not sure which parts of `benchserver.py` would benefit from unit tests.

The test in `tests/test_commands.py` provides a basic check that the bench command runs and ensures the presence/absence of strings in the terminal output. The unit tests I’ve added, however, offer more granular coverage, specifically targeting the internals of `benchserver.py`, such as HTML rendering and request handling. These tests increase the coverage on `benchserver.py` from 30% to 88%.

The tests are not autogenerated—I wrote them as a way to contribute to Scrapy and align with the [contribution guidelines](https://docs.scrapy.org/en/2.11/contributing.html#writing-patches), which encourage improving test coverage to get more familiar with the project. Initially, I misunderstood the bench server’s invocation and thought it was broken. In hindsight, the focus of this PR should have been solely on adding unit tests, which would have kept the review process simpler. I understand if you think the potential issue with running `python scrapy/utils/benchserver.py` is best addressed separately, possibly with a wider scope, or even just left as is.

LucasSD:
Hi @wRAR I've made a separate PR #6515  to only add the unit tests for `benchserver.py`, as I thought that would be cleaner. Please take a look when you have time

PR #6485, opened 2024-10-04
Replace `mitmproxy` with a simple proxy implementation.

wRAR:
This removes `mitmproxy` from test dependencies for good. Justification: it's too complex, accounting for ~1/3 of packages installed in the current `py` test env, and that also often leads to dependency version clashes or other problems. And the proxy tests we currently have are so basic that we don't need a complex proxy implementation. I didn't find a good lightweight pip-installable proxy so I found some old implementation that adds CONNECT support on top of `twisted.web.proxy`, fixed and modernized it and added simple `Proxy-Authorization` support on top of that. Hopefully it handles all cases we want it io handle.


kmike:
I like this idea. But we might need to add tests for this proxy implementation.

wRAR:
Yeah.

If we split mock* into another library it will be more straightforward to add tests there :)

PR #6487, opened 2024-10-07
Remove the installation dependency on setuptools.

wRAR:
It was added (in #5122) because we imported `pkg_resources`, which we no longer do since 2.10.


Gallaecio (approved):

Gallaecio:
Does this mean we can remove https://github.com/scrapy/scrapy/blob/46cddc6ecfbe9a0750676143ef789acf6c2e637d/tox.ini#L53 ?

PR #6488, opened 2024-10-08
Streaming Sitemap Parsing and Gzip Decompression - sitemap.py

AlexTodorov11:
Changes enable streaming processing for both regular and gzipped XML sitemaps, which helps handle large sitemaps more efficiently. 

By parsing the sitemap incrementally and limiting memory usage during Gzip decompression, the spider will be better suited to handle very large sitemaps without running out of memory or being too slow.

**IMPORTTANT:** Importing gzip, xml.etree.ElementTree as ET, and BytesIO from io - these are standard Python libraries that are commonly used for handling compressed files and parsing XML efficiently. It's done in efforts to boost efficiency and robustness/use.

Streaming Sitemap Parsing (_stream_sitemap):
- This function uses xml.etree.ElementTree.iterparse, which parses the XML in a streaming fashion, processing one element at a time.
- It avoids loading the entire sitemap into memory by processing each <url> tag as soon as it encounters it.
- We use elem.clear() to clear memory after processing each element, further reducing memory usage.

Gzip Decompression (_decompress_gzip_stream):
- This function decompresses the gzipped sitemap in chunks. It reads the compressed data in 1 MB chunks (this can be adjusted) and checks the total size as it decompresses.
- If the decompressed size exceeds the allowed max_size, it raises an exception and logs a warning.
- This avoids loading the entire decompressed sitemap into memory at once, which is especially useful for very large gzipped sitemaps.

### DIFF
--- a/scrapy/spiders/sitemap.py
+++ b/scrapy/spiders/sitemap.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import logging
 import re
+import gzip
+import xml.etree.ElementTree as ET
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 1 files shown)


wRAR:
Did you actually run this code?

PR #6166, opened 2023-12-01
Add Python 3.13 to CI

Gallaecio:
Current blockers:
- [x] [zstandard](https://github.com/indygreg/python-zstandard/issues/210) (directly and via mitmproxy)
- [x] [pywin32](https://github.com/mhammond/pywin32/issues/2367)
- [x] [Pillow](https://github.com/python-pillow/Pillow/issues/8076)
- [x] [typing-extensions](https://github.com/python/typing_extensions/issues/405)
- [x] [uvloop](https://github.com/MagicStack/uvloop/issues/622)
- [x] [greenlet](https://github.com/python-greenlet/greenlet/pull/396) (via bpython)
- [x] [pyftpdlib](https://github.com/giampaolo/pyftpdlib/issues/650)

PR #6492, opened 2024-10-11
add initial support for Python 3.13

rettinghaus:


wRAR:
We already have #6166

PR #6495, opened 2024-10-13
Update http11.py

yuvalsade:
Hi

Issue: Some proxy providers include a session_id within the proxy username. The session_id controls the specific proxy IP being used, meaning that different sessions correspond to different IP addresses. In the current implementation, only the proxy host is used as part of the cache key, which can lead to unintended connection reuse across different proxy sessions.

Why Caching is Harmful: When the proxy authentication credentials (which include the session_id) are not part of the cache key, requests that should use different proxies may inadvertently reuse the same cached connection. This results in multiple requests using the same IP address when they were intended to use different ones, negating the purpose of rotating proxies. Essentially, the session-specific behavior of these proxies is lost, and this can cause issues like rate-limiting, IP bans, or incorrect data retrieval due to the same proxy being reused repeatedly.

Solution: To address this, the proxy authentication credentials (including the session_id) should be included in the cache key. This ensures that whenever the user or session changes, a new connection is established with the appropriate proxy IP, honoring the intent behind the session-based proxy rotation.

Thank you,
Yuval

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -335,7 +335,7 @@ class ScrapyProxyAgent(Agent):
         proxy_key_host_and_auth = None
          
         if headers:
-            auth = headers.getRawHeaders("Proxy-Authorization1")
+            auth = headers.getRawHeaders("Proxy-Authorization")
             if auth:
                 proxy_key_host_and_auth = self._proxyURI.host.decode() + auth[0]
         
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -344,7 +344,7 @@ class ScrapyProxyAgent(Agent):
         # Cache *all* connections under the same key, since we are only
         # connecting to a single destination, the proxy:
         return self._requestWithEndpoint(
-            key=("http-proxy", self._proxyURI.host, self._proxyURI.port),
+            key=("http-proxy", proxy_key_host_and_auth, self._proxyURI.port),
             endpoint=self._getEndpoint(self._proxyURI),
             method=method,
             parsedURI=URI.fromBytes(uri),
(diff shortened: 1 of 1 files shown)


yuvalsade (commented):
Proxy-Authorization header rename

wRAR:
Why do you think this code solves your problem?

yuvalsade:
**Why this code solves the problem:**

Before the change, the proxy connection was cached only by IP and port, so the same connection was reused even when the `Proxy-Authorization` header (which can control session or country) changed. Some proxies use the username for IP rotation or country selection (e.g., `user-s{session}-country-{country}`). This update ensures that if the `Proxy-Authorization` changes, it's treated as a new proxy, preventing unwanted connection reuse and ensuring correct proxy rotation.

wRAR:
Did you run it?

yuvalsade:
Yes

בתאריך יום ב׳, 14 באוק׳ 2024, 10:30, מאת Andrey Rakhmatullin ‏<
***@***.***>:

yuvalsade:
Yes, I did, even monkey patched it in my local test.

encountered it while trying to evaluate scrapy as a proxy solution.

wRAR:
How can you explain the change in behavior when your code is a no-op?

yuvalsade:
Not really sure what you mean, but technically, if the both
Proxy-Authentication,Ip as port of the proxy does not change, the code does
not have any effect.
If It does change, then a new connection will be used for it, with the
correct  Proxy-Authentication header.

wRAR:
Please explain what effects does your code have, in details.

yuvalsade:
it makes sure that the proxy key for caching will take into account the
Proxy-Authentication header + ip + port, that's it, nothing less, nothing
more.

yuvalsade:
if i am missing something, please let me know

wRAR:
Your code is a no-op.

yuvalsade (commented):
fixed the no-op issue

yuvalsade:
yes i did, but from a container, when i copied the code to the pull request
i missed it.

It was 2am here.

sorry about that.

Gallaecio (commented):
Besides adding a test as suggested, please also consider whether or not you need to use `decode()` on the `.host`, the original code did not. Maybe instead you need to use `auth[0].encode()`.

PR #6498, opened 2024-10-17
Remove --keep-runtime-typing from pyupgrade.

wRAR:


Gallaecio:
Is that the FTP issue affecting macOS? oO

Gallaecio (approved):

wRAR:
Looks like the intermittent reactor issues affect MacOS much more often.

PR #6500, opened 2024-10-18
Deprecation removals.

wRAR:
* ``--output-format/-t`` argument: deprecated in 2.1.0
* Running `Crawler.crawl()` more than once on the same instance: deprecated in 2.11.0
* Subclassing `HttpCompressionMiddleware` without `from_crawler()` support: deprecated in 2.5.0
* `RetryMiddleware.EXCEPTIONS_TO_RETRY`: deprecated in 2.10.0
* S3 feed exports without `boto3`: deprecated in 2.10.0
* `_FeedSlot` as an alias for `FeedSlot`: deprecated in 2.10.0
* `NoimagesDrop`: deprecated in 2.8.0
* `convert_image()` without `response_body`: deprecated in 2.8.0
* The `custom` argument of `build_component_list()`: deprecated in 2.10.0
* `get_asyncio_event_loop_policy()`: deprecated in 2.9.0

I'm not 100% sure that nothing needs to be changed for the `botocore`/`botocore-pinned` test envs.

I still hesitate to remove the `FEED_URI` support even though it's deprecated for several years.

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -141,7 +141,7 @@ class Crawler:
             raise RuntimeError("Crawling already taking place")
         if self._started:
             raise RuntimeError(
-                "Running Crawler.crawl() more than once on the same instance is forbidden."
+                "Cannot run Crawler.crawl() more than once on the same instance."
             )
         self.crawling = self._started = True
 
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import logging
 import pprint
 import signal
-import warnings
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 
 from twisted.internet.defer import (
@@ -17,7 +16,6 @@ from zope.interface.verify import verifyClass
 from scrapy import Spider, signals
 from scrapy.addons import AddonManager
 from scrapy.core.engine import ExecutionEngine
-from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.extension import ExtensionManager
 from scrapy.interfaces import ISpiderLoader
 from scrapy.logformatter import LogFormatter
@@ -142,10 +140,8 @@ class Crawler:
         if self.crawling:
             raise RuntimeError("Crawling already taking place")
         if self._started:
-            warnings.warn(
-                "Running Crawler.crawl() more than once is deprecated.",
-                ScrapyDeprecationWarning,
-                stacklevel=2,
+            raise RuntimeError(
+                "Running Crawler.crawl() more than once on the same instance is forbidden."
             )
         self.crawling = self._started = True
 
--- a/scrapy/pipelines/images.py
+++ b/scrapy/pipelines/images.py
@@ -8,14 +8,13 @@ from __future__ import annotations
 
 import functools
 import hashlib
-import warnings
 from contextlib import suppress
 from io import BytesIO
 from typing import TYPE_CHECKING, Any, cast
 
 from itemadapter import ItemAdapter
 
-from scrapy.exceptions import DropItem, NotConfigured, ScrapyDeprecationWarning
+from scrapy.exceptions import NotConfigured
 from scrapy.http import Request, Response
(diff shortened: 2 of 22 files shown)


wRAR:
> I'm not 100% sure that nothing needs to be changed for the botocore/botocore-pinned test envs.

OK, this fails now. We may need an explicit pytest mark for the tests that we want to run there.

scrapy/crawler.py
```diff
@@ -142,10 +140,8 @@ def crawl(self, *args: Any, **kwargs: Any) -> Generator[Deferred[Any], Any, None
         if self.crawling:
             raise RuntimeError("Crawling already taking place")
         if self._started:
-            warnings.warn(
-                "Running Crawler.crawl() more than once is deprecated.",
-                ScrapyDeprecationWarning,
-                stacklevel=2,
+            raise RuntimeError(
+                "Running Crawler.crawl() more than once on the same instance is forbidden."
```
Gallaecio, line 144: “Forbidden” in this context sounds a bit weird to me. What about:

```suggestion
                "Cannot run Crawler.crawl() more than once on the same instance."
```

scrapy/pipelines/images.py
```diff
@@ -203,49 +188,25 @@ def get_images(
...
+            thumb_image, thumb_buf = self.convert_image(image, size, response_body=buf)
             yield thumb_path, thumb_image, thumb_buf
 
     def convert_image(
         self,
         image: Image.Image,
         size: tuple[int, int] | None = None,
-        response_body: BytesIO | None = None,
+        *,
+        response_body: BytesIO,
```
Gallaecio, line 208: I get that the goal is to make the parameter required, without a default value, but we deprecated it having `None` as value, not it being used as a positional argument, so this change is backward-incompatible.

I’m thinking we should raise `ValueError` if the value is `None` now, and log a new warning if the parameter is used positionally, so that we can eventually make this change. And maybe we should take the chance to do the same with  `size`, deprecate it being positional.

I was unsure how to handle such a deprecation without removing the typing hints. Gemini suggested using `inspect`:

```python
import inspect

def my_function(param1: int, param2: str = "default_value"):
    frame_record = inspect.stack()[1]
    args, _, _, values = inspect.getargvalues(frame_record[0])
    if param2 in args:
        print("param2 was passed positionally")
    else:
        print("param2 was passed as a keyword argument")
```

scrapy/utils/conf.py
```diff
@@ -21,7 +20,7 @@
 
 def build_component_list(
     compdict: MutableMapping[Any, Any],
-    custom: Any = None,
+    *,
```
Gallaecio, line 23: While this seems similar to the `convert_image` scenario above, I feel like here we are more justified in this change, because the Pythonic way to fix `build_component_list({}, foo, bar)` was switching to `build_component_list({}, convert=bar)`.

That said, I wonder if we should follow a similar deprecation approach, in case some users went with `build_component_list({}, None, bar)` instead, as that was also a valid way to get rid of the warning.

scrapy/utils/conf.py
```diff
@@ -159,7 +145,7 @@ def feed_complete_default_values_from_settings(
 def feed_process_params_from_cli(
     settings: BaseSettings,
     output: list[str],
-    output_format: str | None = None,
+    *,
```
Gallaecio, line 148: Same as with `build_component_list` here. I am OK with this approach, but I would see the argument for deprecating the existence of the parameter separately in a more robust way.

scrapy/pipelines/images.py
wRAR: FTR this suggestion doesn't work :) https://stackoverflow.com/questions/56468082/inspect-if-an-argument-was-passed-positionally-or-via-keyword

Gallaecio: :disappointed:

Gallaecio (approved):
Given https://github.com/scrapy/scrapy/pull/6500#discussion_r1824245893, :heavy_check_mark: from me.

PR #6501, opened 2024-10-19
Filter test-time warnings.

wRAR:


tests/test_utils_asyncio.py
```diff
@@ -1,8 +1,8 @@
 import asyncio
 import warnings
-from unittest import TestCase
 
 from pytest import mark
+from twisted.trial.unittest import TestCase
```
wRAR, line 5: Found when looking at deprecation warnings but it's actually a bug fix, the test didn't run before the change.

Gallaecio (approved):

PR #6506, opened 2024-10-24
Enhancement: Update docs to include IgnoreRequest details

### LINKED ISSUE #6405: Documentation bug - IgnoreRequest Related Documentation
The documentation around IgnoreRequest seems to be incomplete:

* DOWNLOAD_MAXSIZE - https://docs.scrapy.org/en/latest/topics/settings.html?highlight=download_maxsize#download-maxsize - doesn't say that it returns an IgnoreRequest on Failure
* REDIRECT_MAX_TIMES - https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#std-setting-REDIRECT_MAX_TIMES - explicitly says: "After this maximum, the request’s response is returned as is." - which is misleading - it returns an IgnoreRequest.

Looking through the scrapy code, I see there seem to be two other uses for IgnoreRequest in the core code:

* ROBOTSTXT_OBEY - doesn't say responses come back with IgnoreRequest. https://docs.scrapy.org/en/latest/topics/settings.html?highlight=download_maxsize#robotstxt-obey
* Something about caching. No idea if this is documented.

In fact, looking at both of those pages more broadly (settings and downloader middleware), neither explicitly says "you'll get a IgnoreRequest back for this".
As the Python ethos goes: Explicit is better than implicit. Especially for docs.

RJPalmer:
Resolves #6405

Changed the wording for DOWNLOAD_MAXSIZE and REDIRECT_MAX_TIMES to mention IgnoreRequest


docs/topics/settings.rst
```diff
@@ -889,8 +889,8 @@ DOWNLOAD_MAXSIZE
 
 Default: ``1073741824`` (1 GiB)
 
-The maximum response body size (in bytes) allowed. Bigger responses are
-aborted and ignored.
+The maximum response body size (in bytes) allowed. Bigger responses return
+:exc:`IgnoreRequest`.
```
Gallaecio, line 893: I think the previous wording was OK.

The fact that `IgnoreRequest` is raised feels to me like an internal detail, only relevant for downloader middleware writers, not for any user.

docs/topics/downloader-middleware.rst
```diff
@@ -876,7 +876,7 @@ REDIRECT_MAX_TIMES
 Default: ``20``
 
 The maximum number of redirections that will be followed for a single request.
-After this maximum, the request's response is returned as is.
+After this maximum, the request returns :exc:`IgnoreRequest`.
```
Gallaecio, line 879: Here I do think we need to fix the wording, but I think we need a wording more akin to the one from the other section (before the change), e.g.

```suggestion
After this maximum, the request is dropped.
```

docs/topics/settings.rst
RJPalmer: How does this sound?

"The maximum response body size (in bytes) allowed. Responses greater in size are aborted and ignored."

Gallaecio: Sounds OK, but I don’t see it as an improvement over the current wording, it’s a bit lengthier and I don’t think it adds clarification to reword “Bigger” as “greater in size”. I still think it may be best to leave the original wording.

docs/topics/downloader-middleware.rst
```diff
@@ -875,8 +875,8 @@ REDIRECT_MAX_TIMES
 
 Default: ``20``
 
-The maximum number of redirections that will be followed for a single request.
-After this maximum, the request's response is returned as is.
+The maximum response body size (in bytes) allowed. Bigger responses are
+aborted and ignored.
```
Gallaecio, line 879: You accidentally changes this paragraph about redirects :sweat_smile:

Gallaecio:
Sorry, but this has gotten a bit messy.

Needed changes:
- The redirect paragraph does need to change, but you had accidentally overwritten it with information that was about the big response paragraph instead. It needs to be changed, but only to indicate that once redirects are exceeded responses are “ignored and aborted” (for consistent wording with the big response paragraph).
- The big response paragraph should not change at all, the wording in the main branch is OK, we should not mention `IgnoreRequest` in this context.

RJPalmer:
- Redirect should read: _"The maximum number of redirections that will be followed for a single request. Redirections greater in size are aborted and ignored."_
- Big Response will read: _"The maximum response body size (in bytes) allowed. Bigger responses are
aborted and ignored"_

Gallaecio:
“Redirections greater in size are” → “If maximum redirections are exceeded, the request is”.

Gallaecio (approved):

wRAR (approved):

PR #6508, opened 2024-10-25
added a version checker for stack assignment

### LINKED ISSUE #6504: Fix typing for Twisted 24.10.0
This should be easy but we need this to be included in 2.12 so I've created this for visibility.

`Failure.stack` is a special deprecated property thing in Twisted 24.10.0, and assigning to it is a no-op, maybe it's better to put the assignment under a Twisted version check instead of e.g. specifically silencing the typing warning .

Rohitkr117:
**Fixed the compatibility issues with version of twisted > 24.10**

Fixes #6504 

*What's changed:*
 - added a checker as sugested in the media.py file.
 - the code won't assign anything to result.stack since that attribute is deprecated for newer versions of Twisted (24.10.0 and above)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python-version: "3.12"
+        - python-version: "3.13"
           env:
             TOXENV: pylint
         - python-version: "3.9"
@@ -24,7 +24,7 @@ jobs:
         - python-version: "3.12"  # Keep in sync with .readthedocs.yml
           env:
             TOXENV: docs
-        - python-version: "3.12"
+        - python-version: "3.13"
           env:
             TOXENV: twinecheck
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.12
+          python-version: "3.13"
       - run: | 
           pip install --upgrade build twine
           python -m build
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/tests-ubuntu.yml
+++ b/.github/workflows/tests-ubuntu.yml
@@ -24,7 +24,10 @@ jobs:
         - python-version: "3.12"
           env:
             TOXENV: py
-        - python-version: "3.12"
+        - python-version: "3.13"
+          env:
+            TOXENV: py
+        - python-version: "3.13"
           env:
             TOXENV: asyncio
         - python-version: pypy3.9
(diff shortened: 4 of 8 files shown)


scrapy/pipelines/media.py
```diff
@@ -206,7 +206,8 @@ def _cache_result_and_execute_waiters(
             # minimize cached information for failure
             result.cleanFailure()
             result.frames = []
-            result.stack = []
+            if twisted.version.major < 24 or (twisted.version.major == 24 and twisted.version.minor < 10):
```
wRAR, line 209: You shoul simplify the comparison using `Version`: https://github.com/scrapy/scrapy/blob/master/scrapy/mail.py#L223

scrapy/pipelines/media.py
```diff
@@ -206,7 +206,8 @@ def _cache_result_and_execute_waiters(
...
+                result.stack = []
```
wRAR, line 210: Please run `tox -e typing` with `Twisted>=24.10.0rc1` temporarily added to its deps, as I think mypy will still flag this line even with the condition added. if it's flagged you'll need to add an appropriate `type: ignore` here.

Rohitkr117: As per your request, I ran `tox -e typing` with `Twisted>=24.10.0rc1` temporarily added to the dependencies. During the test, I encountered several issues, which I have resolved by making appropriate changes and using `# type: ignore` where necessary.

### Issues Encountered & Fixes:

#### 1. **Error: `Module has no attribute "parse_and_bind"` in `scrapy/utils/console.py:85`**

   - **Fix:**
     This is likely a dynamically defined attribute that `mypy` doesn't recognize. To bypass the typing issue, I added:
     ```python
       readline.parse_and_bind("tab:complete") # type: ignore[attr-defined]
     ```

#### 2. **Error: `Cannot assign to a method` in `scrapy/pipelines/media.py:212`**

   - **Fix:**
   ```python
       result.stack = [] # type: ignore[method-assign]
``` 
#### 3. **Error: `Module has no attribute "SIGUSR2"` in `scrapy/extensions/debug.py:36`, Error: `Module has no attribute "SIGQUIT"; maybe "SIGINT"?` in `scrapy/extensions/debug.py:37` and `scrapy/extensions/debug.py:73`**
   - **Error:**
     ```
     scrapy/extensions/debug.py:36: error: Module has no attribute "SIGUSR2"  [attr-defined]
     scrapy\extensions\debug.py:37: error: Module has no attribute "SIGQUIT"; maybe "SIGINT"?  [attr-defined]
     scrapy/extensions/debug.py:73: error: Module has no attribute "SIGUSR2"  [attr-defined]
     
     ```
   - **Fix:**
    Fixed by adding :
     ```python
      # type: ignore[attr-defined]
     ```
All errors have been resolved, and the code now passes the `tox -e typing` test. Please review the changes and let me know if I can modify my PR with the following changes.

File: scrapy/pipelines/media.py (hunk 1 above)
Comment (Rohitkr117) [reply scrapy/pipelines/media.py:209]: ##Changed it to:
   ```python
            if twisted_version <= Version("twisted", 24, 10, 0):
                result.stack = []
```

scrapy/pipelines/media.py
wRAR: Oh, I didn't know we have special typing issues on Windows. That should be fixed separetely if we want that (and we likely want it if we want contributors to run it). Can you please only leave the `result.stack` `type:ignore` comment here and optionally create a separate PR for others?

Also the tox.ini change shouldn't be committed (I've checked that the typing test succeeds with it and with your changes though).

Please also use `pre-commit` as documented in https://docs.scrapy.org/en/latest/contributing.html#pre-commit

Rohitkr117: I'm little unsure if I did it right or not but I force pushed the required changes and this time, I commited my changes using `pre-commit`. I am not sure how this codecov report works.

wRAR: Looks likve you've made and pushed a merge commit, you should make a normal commit instead.

Rohitkr117: Ohhkkk I'll do it

PR #6511, opened 2024-10-26
Fixes for Twisted Version Check and Typing Issues

### LINKED ISSUE #6504: Fix typing for Twisted 24.10.0

Rohitkr117:
Fixes #6504 

#### This PR went through the following changes:
1. Simplified the version check using format `Version` as requested.
2. Added `type: ignore` comments to handle the typing errors flagged by `mypy`.
3. Pre-commit checks have been run.

#### Testing:
- Ran `tox -e typing` with `Twisted>=24.10.0rc1` as suggested
- Reverted the changes to the `tox.ini` file.
- Some other errors were found on windows for the file `scrapy/extensions/debug.py`

Made a new Pr as I got lost with the changes and had to reset my fork
Let me know if you need any more modifications!

### DIFF
--- a/scrapy/pipelines/media.py
+++ b/scrapy/pipelines/media.py
@@ -16,9 +16,9 @@ from typing import (
 )
 
 from twisted import version as twisted_version
-from twisted.python.versions import Version
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.python.failure import Failure
+from twisted.python.versions import Version
 
 from scrapy.http.request import NO_CALLBACK, Request
 from scrapy.settings import Settings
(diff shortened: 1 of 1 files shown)


Rohitkr117:
What should i change to pass all the checks? Is this due to the other 3 errors in the `scrapy/extensions/debug.py` file?

wRAR (approved):

Issue #6509, opened 2024-10-26
Deprecate `scrapy.twisted_version`


wRAR:
We define and even explicitly export `scrapy.twisted_version` for, as I can see, no good reason, as users should use `twisted.version` directly. We should deprecate (via, I assume, a module-level `__getattr__`) and eventually remove it.

Rohitkr117:
I'd like to work on this issue. 
I propose adding a `__getattr__` method to handle access to `scrapy.twisted_version` and emit a DeprecationWarning advising users to switch to `twisted.version` directly. This will allow us to maintain backward compatibility for the time being while preparing for its eventual removal. I'll submit a PR soon. Let me know if that sounds good!

wRAR:
Yeah, it sounds good.

Laerte:
Sorry @Rohitkr117 I only saw your message after I already open a PR. 😅

PR #6512, opened 2024-10-26
Deprecate `scrapy.twisted_version`

Laerte:
fix: #6509

### DIFF
--- a/scrapy/__init__.py
+++ b/scrapy/__init__.py
@@ -28,7 +28,7 @@ __version__ = (pkgutil.get_data(__package__, "VERSION") or b"").decode("ascii").
 version_info = tuple(int(v) if v.isdigit() else v for v in __version__.split("."))
 
 
-def __getattr__(name: str) -> None | tuple:
+def __getattr__(name):
     if name == "twisted_version":
         import warnings
 
--- a/scrapy/__init__.py
+++ b/scrapy/__init__.py
@@ -4,6 +4,7 @@ Scrapy - a web crawling and web scraping framework written for Python
 
 import pkgutil
 import sys
+import warnings
 
 # Declare top-level shortcuts
 from scrapy.http import FormRequest, Request
@@ -32,9 +33,6 @@ def __getattr__(name: str):
     if name == "twisted_version":
         import warnings
 
-        # Ignore noisy twisted deprecation warnings
-        warnings.filterwarnings("ignore", category=DeprecationWarning, module="twisted")
-
         from twisted import version as _txv
 
         from scrapy.exceptions import ScrapyDeprecationWarning
@@ -48,5 +46,9 @@ def __getattr__(name: str):
     raise AttributeError
 
 
+# Ignore noisy twisted deprecation warnings
+warnings.filterwarnings("ignore", category=DeprecationWarning, module="twisted")
+
 del pkgutil
 del sys
+del warnings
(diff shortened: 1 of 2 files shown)


scrapy/__init__.py
```diff
@@ -30,13 +26,22 @@
 # Scrapy and Twisted versions
 __version__ = (pkgutil.get_data(__package__, "VERSION") or b"").decode("ascii").strip()
 version_info = tuple(int(v) if v.isdigit() else v for v in __version__.split("."))
-twisted_version = (_txv.major, _txv.minor, _txv.micro)
 
 
-# Ignore noisy twisted deprecation warnings
-warnings.filterwarnings("ignore", category=DeprecationWarning, module="twisted")
```
Laerte, line 37: For some reason if I add this piece I get duplicated deprecation warning message, also not sure where should I place a test for this change.

wRAR: > if I add this piece I get duplicate deprecation warning message

Can you show an example?

I think this line shouldn't be removed, though it's probably not that useful, as IIRC `DeprecationWarning` warnings aren't printed anywhere by default.

wRAR:
No idea what do these pytest errors mean.

scrapy/__init__.py
Laerte: Sure, add the removed line after the warnings import and then call this snippet:

```python
print(twisted_version)
```

Laerte: Now i'm trying to figure out why tests are failing after I added ``__getattr__`` method. 😅

Laerte:
> No idea what do these pytest errors mean.

So because we defined ``__getattr__``
 the default behavior is to raise a `AttributeError` exception but It was returning None instead, that's why was failing, fixed.

Ref: https://github.com/pytest-dev/pytest/issues/4631

scrapy/__init__.py
```diff
@@ -30,13 +26,24 @@
...
+
+        from scrapy.exceptions import ScrapyDeprecationWarning
+
+        warnings.warn(
+            "The scrapy.twisted_version attribute is deprecated, use twisted.version instead",
+            ScrapyDeprecationWarning,
+        )
+        return _txv.major, _txv.minor, _txv.micro
+
+    raise AttributeError
```
Laerte, line 45: Wondering since this a module not a class if we should raise `ImportError` instead.

Laerte:
@wRAR Added tests for test coverage

scrapy/__init__.py
Laerte: Tried that, don't work. 😄

scrapy/__init__.py
Laerte: Got it, fixed.

scrapy/__init__.py
```diff
@@ -30,13 +26,27 @@
...
 
 
-# Ignore noisy twisted deprecation warnings
-warnings.filterwarnings("ignore", category=DeprecationWarning, module="twisted")
+def __getattr__(name: str):
+    if name == "twisted_version":
+        import warnings
+
+        # Ignore noisy twisted deprecation warnings
+        warnings.filterwarnings("ignore", category=DeprecationWarning, module="twisted")
```
wRAR, line 36: I think the reason for this line is to be executed before any Scrapy code, so that the user doesn't see those warnings. It was there before `from twisted import version` was even added.

Laerte: Pushed the rollback to original code.

Gallaecio:
```
scrapy/pipelines/media.py:209: error: Cannot assign to a method  [method-assign]
```

Gallaecio:
> ```
> ```

Nevermind, it seems that might have been introduced by https://github.com/scrapy/scrapy/pull/6510.

Gallaecio (approved):

wRAR (approved):

wRAR:
Thanks!

PR #6515, opened 2024-10-27
Add unit tests for scrapy/utils/benchserver.py

LucasSD:
This PR adds unit tests for `scrapy/utils/benchserver.py`, increasing its test coverage from ~30% to ~70%.

### DIFF
--- a/scrapy/utils/benchserver.py
+++ b/scrapy/utils/benchserver.py
@@ -31,10 +31,11 @@ def _getarg(request, name: bytes, default: Any = None, type=str):
     return type(request.args[name][0]) if name in request.args else default
 
 
-def main():
+if __name__ == "__main__":
     from twisted.internet import reactor
 
     root = Root()
+    factory = Site(root)
     httpPort = reactor.listenTCP(8998, Site(root))
 
     def _print_listening() -> None:
@@ -43,7 +44,3 @@ def main():
 
     reactor.callWhenRunning(_print_listening)
     reactor.run()
-
-
-if __name__ == "__main__":
-    main()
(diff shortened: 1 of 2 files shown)


LucasSD:
> This feels too _mocky_ to me. I think the right approach to test a server like this would be to perform an end-to-end test, where you try actual HTTP requests against the running server. That way, changes in the implementation, such as migrating the server to aiohttp instead of Twisted, won’t break the tests.
> 
> You could start the server similarly to [how it is done by the bench command](https://github.com/scrapy/scrapy/blob/d2bdbad8c8cc5e5b4b9d3a79c94e2411a44e94be/scrapy/commands/bench.py#L32), you can see that we do something similar already in some tests with a MockServer. And then you run a spider against the test that verifies the effects of the test parameters (or the lack of them).
> 
> Ideally the spider used to test the server should do the minimal work it can to verify the server behavior, and not actually crawl all possible URLs.

Thanks for the feedback. I think unit tests naturally involve a lot of mocking and that, ideally, a combination of unit, integration, and E2E tests provides the best overall coverage.

If you feel there’s limited value in including these unit tests as they are, I can look into implementing what you suggested. Would you prefer to include both the unit and E2E tests in this PR, or do you think the unit tests don’t add any value and should be excluded altogether?

Gallaecio:
I this case, I don’t think unit tests add much value.

LucasSD:
> I this case, I don’t think unit tests add much value.

Ok, probably best to close this PR then? If I add the tests according to your suggestion, I'll make a new PR.

Gallaecio:
Thank you, though. I’m sorry it was a bit of a wasted effort.

PR #6526, opened 2024-11-04
Improve internal refs to scrapy.Request and scrapy.Selector

wRAR:
(I've built the docs with `nitpicky=True` and there are many unresolved references, we ideally want to fix them but that needs many different things for different cases)

This change is similar to the one in #6427 and for a similar problem, though in these two cases the problem is harder. I may not be understanding it completely, but there are several components in it. 

1. We have several import paths available for some classes, and may use several of them in different places in docs. E.g. `scrapy.Selector`, `scrapy.selector.Selector`, `scrapy.selector.unified.Selector`.
2. As there is no way in Sphinx to have one `class::` directive with multiple import paths, in some cases we tried to list several `class::` directives one after another with different import paths (though even using just one entry with one import path doesn't work).
3. Point 2 doesn't work as expected because we usually have a `module::` directive in a per-module doc, and that causes e.g. `scrapy.Item` to turn into `scrapy.items.scrapy.Item` and makes anything except `scrapy.items.Item` impossible to express in a `class::` directive. Though I think this doesn't work for `autoclass::`.
4. While Sphinx doesn't have arbitrary aliases, it has `:canonical:` which allows two import paths for a class: a short one defined via `class::` and a canonical long one defined in its `:canonical:` option, which is intended to be the path to the real definition, i.e. `scrapy.selector.unified.Selector`. It's autogenerated for `autoclass::`, from the real import path. 

While this is problematic, I think the only real problem here that can't be fixed by editing our sources is lack of support for more than "short import path + real import path". This was discussed in https://github.com/sphinx-doc/sphinx/issues/10348, there is some good analysis there that applies to us, about several layers of packages exporting names from their subpackages, and the only suggestion is to decide on which API is to use as a short one. In our case it should be the shortest one, `scrapy.Foo` for several things exported at the top level and `scrapy.foo.Foo` for most of the other things, and we won't be able to refer to `scrapy.Selector` as `scrapy.selector.Selector` (and won't want to refer to it as `scrapy.selector.unified.Selector`). 

So the fixes should be like this: 

1. Change the definition to use the shortest name in `class::`/`autoclass::`.
2. If not using `autoclass::` add `:canonical:`.
3. If the interaction between `module::` and `class::` leads to an undesired import path in `class::` then fix it somehow (switching to `autoclass::` helps?).
4. Change the references to the shortest name. Looks like this also means using a full shortest name in the same file where the class is defined, as the implicit prefix from `module::` is no longer useful for it.


Gallaecio:
About 3, I believe in some cases it can be addressed with https://github.com/scrapy/scrapy/blob/d2156696c45e023479ae1bdee8623bb6212e975c/docs/topics/request-response.rst?plain=1#L480 to remove the effects of `module` for the following lines.

docs/topics/request-response.rst
```diff
@@ -7,15 +7,15 @@ Requests and Responses
 .. module:: scrapy.http
    :synopsis: Request and Response classes
 
-Scrapy uses :class:`Request` and :class:`Response` objects for crawling web
+Scrapy uses :class:`scrapy.Request` and :class:`Response` objects for crawling web
```
Gallaecio, line 10:
```suggestion
Scrapy uses :class:`~scrapy.Request` and :class:`Response` objects for crawling web
```

docs/topics/request-response.rst
```diff
@@ -69,7 +69,7 @@ Request objects
 
         .. caution:: Cookies set via the ``Cookie`` header are not considered by the
             :ref:`cookies-mw`. If you need to set cookies for a request, use the
-            :class:`Request.cookies <scrapy.Request>` parameter. This is a known
+            :class:`scrapy.Request.cookies <scrapy.Request>` parameter. This is a known
```
Gallaecio, line 72: This looks unintended, though.

wRAR:
I couldn't fix one problem (and it's actually a regression), but I decided that everything else, including fixed refs on the same page, are worth leaving it as is at least for now. Specifically, we ask Sphinx to insert docstrings for base `parsel.SelectorList` members into [the `scrapy.selector.SelectorList` doc block](https://docs.scrapy.org/en/latest/topics/selectors.html#selectorlist-objects) and those include references to `SelectorList`. These refs are now unresolved. It's quite complicated, especially as these refs refer to `parsel.SelectorList` in the project that actually contains it, but I guess they should link to `scrapy.selector.SelectorList` on the Scrapy doc page.

docs/news.rst
```diff
@@ -635,7 +635,7 @@ Bug fixes
     exception if ``default`` is ``None``.
     (:issue:`6308`, :issue:`6310`)
 
--   :class:`~scrapy.selector.Selector` now uses
+-   :class:`~scrapy.Selector` now uses
```
Gallaecio, line 638: For the record, seeing how you are using `.` as a prefix elsewhere, know you can also use it here after `~`, i.e. `~.Selector`.

Gallaecio (approved):

PR #6527, opened 2024-11-05
Replace Slot.throttle with Request.meta['dont_throttle']

Gallaecio:
https://github.com/scrapy/scrapy/pull/6246#discussion_r1827583470

cc @GeorgeA92, @VMRuiz

### DIFF
--- a/docs/topics/autothrottle.rst
+++ b/docs/topics/autothrottle.rst
@@ -57,6 +57,8 @@ request metadata key to ``True``:
 
 .. code-block:: python
 
+    from scrapy import Request
+
     Request("https://example.com", meta={"dont_throttle": True})
 
 Note, however, that AutoThrottle still determines the starting delay of every
--- a/docs/topics/autothrottle.rst
+++ b/docs/topics/autothrottle.rst
@@ -21,9 +21,14 @@ Design goals
 How it works
 ============
 
-AutoThrottle extension adjusts download delays dynamically to make spider send
-:setting:`AUTOTHROTTLE_TARGET_CONCURRENCY` concurrent requests on average
-to each remote website.
+Scrapy allows defining the concurrency and delay of different download slots,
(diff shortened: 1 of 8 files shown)


docs/topics/autothrottle.rst
```diff
@@ -47,12 +47,19 @@ effect, but there are some important differences:
...
+To disable AutoThrottle for a specific request, set the ``dont_throttle``
+request metadata key to ``True``:
+
+.. code-block:: python
+
+    from scrapy import Request
+
+    Request("https://example.com", meta={"dont_throttle": True})
 
 Note, however, that AutoThrottle still determines the starting delay of every
```
kmike, line 64: There is another caveat: dont_throttle disables the adjustment of the slot delay, but not the delay itself. So, for example, if AutoThrottle was enabled for first N requests to a website, and then there is a request with dont_throttle=True, then the last request is still going to use the delay learned previously.

It seems the `dont_throttle` name could be confusing in this case: we do throttle the request. It seems the flag works more like `autothrottle_dont_adjust_delay` (is there a shorter name?)

Gallaecio: Would `bypass_autothrottle` (or `skip_`/`ignore_`) work? Otherwise, I’m OK with `autothrottle_dont_adjust_delay`.

docs/topics/autothrottle.rst
```diff
@@ -47,12 +47,19 @@ effect, but there are some important differences:
...
-=========================================
+.. reqmeta:: autothrottle_dont_adjust_delay
 
-It is possible to disable AutoThrottle for a specific download slot at run time
-by setting its ``throttle`` attribute to ``False``, e.g. using
-:setting:`DOWNLOAD_SLOTS`.
+Disabling the throttling of a request
+=====================================
+
+To disable AutoThrottle for a specific request, set the
```
Gallaecio, line 55: Let me also update this wording.

wRAR (approved):

PR #6538, opened 2024-11-12
Add 'tests/keys/mitmproxy-dhparam.pem' to .gitignore

krysthian:
The file tests/keys/mitmproxy-dhparam.pem was generated after running tox. I've added it to the .gitignore file to prevent accidental commits of this test-generated artifact. Please let me know if this is the appropriate way to handle it or if any additional steps are needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ test-output.*
 .mypy_cache/
 /tests/keys/localhost.crt
 /tests/keys/localhost.key
+/tests/keys/mitmproxy-dhparam.pem
 
 # Windows
 Thumbs.db


wRAR:
(it's a valid problem, though I'm not sure of adding it to .gitignore rather than committing is the correct way, I need to check)

krysthian:
Thank you @wRAR, I checked it before and I should not include this file in PR. Explanation:
	1.	Auto-Generated Test Artifacts:
	•	The file is generated during the testing process and is considered a test artifact.
	•	Such files are usually not included in the version control system because they can be regenerated as needed.
	2.	Consistency and Security:
	•	Committing auto-generated files can lead to inconsistencies between different environments.
	•	Security-sensitive files like .pem files should not be committed unless absolutely necessary, as they can contain cryptographic material.
	3.	Existing Repository Content:
	•	Check if tests/keys/mitmproxy-dhparam.pem already exists in the Scrapy repository.
	•	If it exists: And your version is different, you should investigate why it’s different.
	•	If it doesn’t exist: It’s likely intended to be generated dynamically and should not be added.
	4.	.gitignore Consideration:
	•	Such files are often added to the .gitignore file to prevent accidental commits.
	5.	[Contribution Guidelines]
	•	Generally, only source code and necessary resources are committed, not generated files.

wRAR:
It's not a test artifact, it's generated by mitmproxy on the first run without it, just like `tests/keys/mitmproxy-ca.pem` IIRC, which we *do* commit.

PR #6540, opened 2024-11-12
Remove build_from_settings() and deprecate from_settings() code

### LINKED ISSUE #6534: Don't ship `build_from_settings()`
We discussed this with @kmike and decided that we want to ship `build_from_crawler()` but not `build_from_settings()` as a last-minute follow-up to #5523/#6169. This, to my knowledge, has two consequences:

1. We need to change `scrapy.middleware.MiddlewareManager` to require a `Crawler` instance for building.
2. Users that use `create_instance()` and pass `settings` but not `crawler` will need to change the logic when migrating to `build_from_crawler()`, but we think they should normally have a Crawler instance there.

`build_from_crawler()` also has a wrong docstring as it doesn't mention `from_settings()`.

wRAR:
Fixes #6534

- [x] Deprecate calling `from_settings()` in `build_from_crawler()` and `MiddlewareManager._build_from_settings()`
- [x] Redo `from_crawler()`/`from_settings()` in `MiddlewareManager`
- [x] ~~Deprecate `ISpiderLoader.from_settings()` and add `ISpiderLoader.from_crawler()` without breaking `verifyClass`~~ It looks like `SpiderLoader` cannot take a crawler and it's created without `build_from_crawler()` anyway.
- [x] Reshuffle `from_crawler()`/`from_settings()` in `MediaPipeline` and its subclasses in a backwards-compatible way

### DIFF
--- a/scrapy/middleware.py
+++ b/scrapy/middleware.py
@@ -54,14 +54,6 @@ class MiddlewareManager:
     @staticmethod
     def _build_from_settings(objcls: type[_T], settings: BaseSettings) -> _T:
         if hasattr(objcls, "from_settings"):
-            warnings.warn(
-                f"{objcls.__qualname__} has from_settings() but not from_crawler()."
-                " This is deprecated and calling from_settings() will be removed in a future"
-                " Scrapy version. You can implement a simple from_crawler() that calls"
-                " from_settings() with crawler.settings.",
-                category=ScrapyDeprecationWarning,
-                stacklevel=2,
-            )
             instance = objcls.from_settings(settings)  # type: ignore[attr-defined]
             method_name = "from_settings"
         else:
--- a/scrapy/middleware.py
+++ b/scrapy/middleware.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import logging
 import pprint
+import warnings
 from collections import defaultdict, deque
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 
-from scrapy.exceptions import NotConfigured
+from scrapy.exceptions import NotConfigured, ScrapyDeprecationWarning
 from scrapy.utils.defer import process_chain, process_parallel
-from scrapy.utils.misc import build_from_crawler, build_from_settings, load_object
+from scrapy.utils.misc import build_from_crawler, load_object
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable
@@ -20,7 +21,7 @@ if TYPE_CHECKING:
 
     from scrapy import Spider
     from scrapy.crawler import Crawler
-    from scrapy.settings import Settings
+    from scrapy.settings import BaseSettings, Settings
 
     _P = ParamSpec("_P")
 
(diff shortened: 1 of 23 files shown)


scrapy/middleware.py
```diff
@@ -50,8 +51,41 @@ def __init__(self, *middlewares: Any) -> None:
     def _get_mwlist_from_settings(cls, settings: Settings) -> list[Any]:
         raise NotImplementedError
 
+    @staticmethod
+    def _build_from_settings(objcls: type[_T], settings: BaseSettings) -> _T:
+        if hasattr(objcls, "from_settings"):
+            warnings.warn(
+                f"{objcls.__qualname__} has from_settings() but not from_crawler()."
```
Gallaecio, line 58: I suspect I am missing something, but from a quick look, it seems to be that `_build_from_settings` is called when `MiddlewareManager.from_settings` is used, and I see nothing ensuring that `objcls` does not have `from_crawler` defined.

scrapy/pipelines/files.py
```diff
@@ -443,12 +446,24 @@ def __init__(
...
+                warnings.warn(
+                    f"FilesPipeline.__init__() was called with a crawler instance and a settings instance"
+                    f" when creating {self.__class__.__qualname__}. The settings instance will be ignored"
+                    f" and crawler.settings will be used. The settings argument will be removed in a future Scrapy version.",
+                    category=ScrapyDeprecationWarning,
+                    stacklevel=2,
+                )
+            settings = crawler.settings
+        elif isinstance(settings, dict) or settings is None:
             settings = Settings(settings)
```
Gallaecio, line 467: Should we deprecate this scenario?

wRAR:
Most changes here are simple: if a class had `from_settings` but not `from_crawler` 

Summary of changes:

1. If a Scrapy class had `from_settings()` but not `from_crawler()` it now has `from_crawler()`. One exception is `ISpiderLoader` that is created before the crawler.
2. Scrapy classes' `from_settings()` now produce a deprecation warning.
3. When a class instance (Scrapy or not) is created via `build_from_crawler()`, and it has from_settings()` but not `from_crawler()` a deprecation warning will be produced.
4. `from_settings()` and `from_crawler()` methods of `MiddlewareManager` are switched so that `from_crawler()` is preferred and the crawler instance is used whenever possible, with appropriate deprecation warnings. As there is an (unlikely to be used in practice) code path that needs to create middleware instances while not having a crawler instance, a simplified copy of `build_from_settings()` was added to the class.
5. The initialization code of `MediaPipeline` and its subclasses is improved to be more in line with the usual `from_crawler()` pattern, more on this below.

scrapy/pipelines/media.py
```diff
@@ -112,13 +147,19 @@ def _key_for_pipe(
...
-        except AttributeError:
+        elif "crawler" in get_func_args(cls.__init__):
+            pipe = cls(crawler=crawler)
+        else:
             pipe = cls()
-        pipe.crawler = crawler
-        assert crawler.request_fingerprinter
-        pipe._fingerprinter = crawler.request_fingerprinter
+            warnings.warn(
+                f"{cls.__qualname__}.__init__() doesn't take a crawler argument."
```
Gallaecio, line 157: In cases like these where the stacklevel would not help to point to the offending like, I wonder if we should prefix `cls.__qualname__` with the module name, i.e. use [`globa_object_name`](https://github.com/scrapy/scrapy/blob/261c4b61dc48353346c1e0387d0783ac15ab459d/scrapy/utils/python.py#L330).

scrapy/utils/misc.py
```diff
@@ -176,7 +175,7 @@ def create_instance(objcls, settings, crawler, *args, **kwargs):
 def build_from_crawler(
     objcls: type[T], crawler: Crawler, /, *args: Any, **kwargs: Any
 ) -> T:
-    """Construct a class instance using its ``from_crawler`` constructor.
+    """Construct a class instance using its ``from_crawler`` or ``from_settings`` constructor.
```
Gallaecio, line 178: I wonder if we need this given `from_settings` is deprecated, but no strong opinion against it.

Gallaecio (approved):

PR #6532, opened 2024-11-10
Add PEP 257-Compliant Docstrings to Scrapy Command Classes

Fraanco:
Added concise, PEP 257-compliant docstrings to all public classes and methods within the Scrapy command classes. These updates ensure consistency, clarity, and support integration with Sphinx's autodoc. The changes improve code readability, maintainability, and facilitate auto-generated documentation.

This pull request is related to the issue: https://github.com/scrapy/scrapy/issues/713

PR #6541, opened 2024-11-13
Add Scrapy Documentation Policies-Compliant Docstrings to Command Classes

Fraanco:
Added concise, PEP 257-compliant docstrings to public classes and methods within check.py, edit.py, fetch.py, genspider.py, runspider.py, settings.py, shell.py, startproject.py, version.py, view.py.
The changes aim to bring more consistency to the way docstrings are added to the code, being strictly compliant to Scrapy's documentation policies (found in https://docs.scrapy.org/en/latest/contributing.html)

These updates ensure consistency, clarity, and support integration with Sphinx's autodoc. The changes improve code readability, maintainability, and facilitate auto-generated documentation.

This pull request is related to the issue: https://github.com/scrapy/scrapy/issues/713


Gallaecio:
I don’t think my meaning came across in the discussion in https://github.com/scrapy/scrapy/pull/6532. I was trying to prevent you from making a PR like this, as I think it does not make sense to add these docstrings.

Fraanco:
Since the documentation/contribution policies are not clear enough about the subjective aspects of what you might think makes sense or not, nor about which files should or should not have docstring documentation, and since my last PR was still open, I thought maybe with a trial and error approach I could find an useful documentation contribution

Gallaecio:
I don’t think any of the docstrings here meet any of the [2 scenarios I mentioned](https://github.com/scrapy/scrapy/pull/6532#issuecomment-2467878874). Is there any specific docstring here that you think meets one of those 2 scenarios?

Fraanco:
hey man. tbh besides the ones I added for consistency sake (e.g.: added to the "short_desc") I believe a lot of them help code understanding (I believe the docstrings for runspider.py, settings.py and shell.py are quite useful for people that are just starting to contribute with open source repos - such as myself)

I do understand the concerns

scrapy/commands/runspider.py
```diff
@@ -16,6 +16,20 @@
...
+
+    Returns:
+        ModuleType: The imported Python module.
+
+    Raises:
+        ValueError: If the provided file is not a valid Python source file.
+
+    Example:
+        >>> module = _import_file('my_spider.py')
+    """
```
Gallaecio, line 32: I am thankful and appreciate that you have put some work into this, but fail to see how adding docstrings like this is useful.

The intro, `Args`, `Returns` and `Example` do not indicate anything that the signature does not  already indicate, I think. If anything, I feel the intro adds some confusion by adding the word “module”, (i.e. are there other types of Python file paths that could be imported but are not a module?).

The only thing that may not be obvious from the signature is that it raises `ValueError` if the file does not have a given file extension. This could be useful in the docstring of a utility method that you may be using from a different file. When hovering the method, in addition to the method signature, an IDE could include the docstring to reveal this detail. However, in this specific case, this is a private function (`_`-prefixed to indicate so), which is only meant to be used in this file. It is only used once. It has 10 lines of code, with 1 `return` and 1 `raise`. Reading the implementation is arguably easier than reading the docstring. And more informative, it tells you the exception is raised based on the file extension, not whether the file is a valid Python file or not (i.e. a file with valid Python code but a different file extension would trigger the exception).

We have many open issues tagged with `docs`. I personally love documentation, have professionally worked as a technical writer before I became a Python programmer, and value how documentation is key for the success of any open source project.

But this feels like box checking: I see a function with a docstring, I add a docstring. While it is useful to have static analysis tools that let you know where docstrings are missing, that does not mean you should add a docstring everywhere there is no docstring.

I don’t want to be too blunt, and I fear I might be discouraging you from contributing to open source. I hope you don’t discard contributing to open source software, even if you decide to stop contributing to this project as I close this PR right after I send this comment.

But if you do decide to continue trying to contribute to Scrapy, please first have a look at the existing issues. Or, if you find nothing you like or if you really want to focus on adding docstrings, find or create an issue about adding docstrings, and let’s discuss where docstrings are needed before you start, so that you don’t waste time writing them for us to reject your work altogether.

PR #6542, opened 2024-11-14
Added failing test cases for return in finally block (#6505)

### LINKED ISSUE #6505: return in finally can swallow exception
### Description

There are two places in `scrapy/contracts/__init__.py` where a `finally:` body contains a `return` statement, which would swallow any in-flight exception. 

This means that if a `BaseException` (such as `KeyboardInterrupt`) is raised from the body, or any exception is raised from one of the `except:` clause, it will not propagate on as expected. 

The pylint warning about this was suppressed in [this commit](https://github.com/scrapy/scrapy/commit/991121fa91aee4d428ae09e75427d4e91970a41b) but it doesn't seem like there was justification for that.

These are the two locations:
https://github.com/scrapy/scrapy/blame/b4bad97eae6bcce790a626d244c63589f4134408/scrapy/contracts/__init__.py#L56

https://github.com/scrapy/scrapy/blame/b4bad97eae6bcce790a626d244c63589f4134408/scrapy/contracts/__init__.py#L86

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

nicklaustrup:
Hi all,

I added 2 test cases to the tests/test_contracts.py file that fail when a keyboard interrupt is triggered during the pre-process and post-process callbacks. I'll submit the proposed changes to scrapy/contracts/__init__.py here in a moment. 

fix  #6505


Gallaecio (approved):

wRAR:
I wonder if the whole `finally` block can be dedented in the first function instead.

Also can you please remove the `# pylint: disable=inconsistent-return-statements` comments too?

nicklaustrup:
That would make sense, if we don't want to override an inflight base exception, then the rest of the block doesn't need to be in a finally. I'll update that and push it through. 

For the `# pylint: disable=inconsistent-return-statements` comments, I don't see them in the tests or the contracts init, do you mean the line in the pylintrc?

nicklaustrup:
@wRAR Updated

wRAR:
The branch state looks very wrong to me now.

wRAR:
Please remove `# pylint: disable=inconsistent-return-statements` comments.

nicklaustrup:
I added the changes to a separate branch that has the correct branch state. Should I open a new PR or is there a way to add them to this request?

wRAR:
Why can't you do this in this PR's branch?

nicklaustrup:
To be honest, I am unsure of how to fix this branches state. However, I was able to fix it in another branch. Sorry I am still learning

wRAR:
Thanks!

wRAR (approved):

PR #6546, opened 2024-11-16
Add settings for TLS min/max version as a replacement for the TLS method

### LINKED ISSUE #3288: Deprecation Warning of Twisted

### LINKED ISSUE #4821: DOWNLOADER_CLIENT_TLS_METHOD only supports TLS 1.2 and lower
`DOWNLOADER_CLIENT_TLS_METHOD` only supports values defined in [`scrapy.core.downloader.tls.openssl_methods`](https://github.com/scrapy/scrapy/blob/5a386393595ae6505840f00d6ea4a51beda31e19/scrapy/core/downloader/tls.py#L22) and they don't include TLS 1.3. I've found that there is no similar `TLSv1_3_METHOD` in pyOpenSSL, then I've found https://stackoverflow.com/questions/62243123/python-pyopenssl-server-doesnt-negotiate-tls-1-3 but I didn't check the implications of this.

wRAR:
Fixes #3288, fixes #4821

Deprecates:
* Setting the `DOWNLOADER_CLIENT_TLS_METHOD` setting to anything other than `TLS`
* `scrapy.core.downloader.tls.METHOD_*` constants
* `scrapy.core.downloader.tls.openssl_methods` dict

It also directly removes support for `DOWNLOADER_CLIENT_TLS_METHOD` from `HttpxDownloadHandler` (especially as the docs falsely claimed that it's not supported).


wRAR:
This is finally ready. It conflicts at docs changes with #7575 but that's trivial to resolve.

AdrianAtZyte (approved):

PR #6547, opened 2024-11-16
Integrating configs into pyproject.toml

### LINKED ISSUE #6514: Migrate from setup.py to pyproject.toml
We should migrate to the modern declarative setuptools metadata approach as discussed in https://setuptools.pypa.io/en/latest/userguide/quickstart.html and https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html, but only after the 2.12 release.

swayam0322:
Fixes #6514 , closes #6520


swayam0322:
@wRAR @Gallaecio 
Some context upon whats happening in setup.cfg would be helpful, Unable to find good docs for the same

wRAR:
> Some context upon whats happening in setup.cfg would be helpful

https://mypy.readthedocs.io/en/stable/config_file.html ?

`bdist_*` shouild be dropped I think.

wRAR:
Looks like you didn't run any tests or checks before submitting.

swayam0322:
Sorry, this was supposed to be a draft PR I am still working on it

swayam0322:
@wRAR  I don't understand why the windows build is failing maybe this is the problem 
> tests/test_crawler.py 900

```
 # sending the second signal too fast often causes problems
 ```

**Comment by wRAR** [comment]
That test in particular is one of the flaky ones though I don't know why in this case it fails even after several retries. I doubt it's related to these changes.

**Comment by swayam0322** [comment]
@wRAR I think the PR is ready for review now

**Comment by wRAR** [comment]
Looks good to me, though I don't currently have time to test these changes yourself.

File: pyproject.toml [hunk 1]
```diff
@@ -0,0 +1,235 @@
...
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["scrapy", "scrapy.*",]
+
+[tool.setuptools.dynamic]
+version = {file = "./scrapy/VERSION"}
+
+[tool.distutils.bdist_rpm]
+doc_files = "docs AUTHORS INSTALL LICENSE README.rst"
```
Gallaecio, line 71: [I wonder if we should remove this](https://github.com/pypa/setuptools/issues/1988).

pyproject.toml
```diff
@@ -0,0 +1,235 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "Scrapy"
```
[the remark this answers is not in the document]
swayam0322: Commited these changes

Gallaecio (commented):
What I checked:
- Executed:
  - `pre-commit run --all`
  - `bump-my-version bump patch|minor`
  - `tox -e py -- tests/test_addons.py --cov-report=term-missing` (the `TYPE_CHECKING` line is not reported as missing, as expected)
  - `tox -e pylint`
  - `tox -e typing` (I wonder if we should take the chance to rename it to `mypy`, for consistency with all our other open source projects)
- Manually compared `setup()` params to new fields.
  - `license` seems to be missing on purpose, in line with [guidelines](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license).
  - `include_package_data` seems to be missing because [it is `True` by default in `pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data).
  - `zip_safe` seems to be missing because [it is obsolete](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html).
  - I assume the new `extract_require` mapping works based on CI passing. I could not find docs to support the change after a quick look, but it looks good.
- Same for `setup.cfg`:
  - `bdist_wheel` seems to be missing on purpose, [it’s for the old build system](https://setuptools.pypa.io/en/latest/build_meta.html), before we switched to [`python -m build`](https://github.com/scrapy/scrapy/blob/efb53aafdcaae058962c6189ddecb3dc62b02c31/.github/workflows/publish.yml#L27).
- `pytest.ini` seems to have been copied vervatim, with the right empty-string-to-empty-list conversions where needed.

Gallaecio (approved):

pyproject.toml
Gallaecio: Let’s drop it, as [wRAR suggested](https://github.com/scrapy/scrapy/pull/6547#issuecomment-2480966647).

swayam0322: Commited the changes

Issue #6550, opened 2024-11-17
Improve open_in_browser function's handling of base tags in response.py


Jonah-Gates:
The `open_in_browser` function in `response.py` uses a workaround to insert a `<base>` tag into the HTML response for local file viewing. As mentioned in the function's docstring, is "dirty" and could fail in situations where `<head>` tags are missing or improperly formed which can lead to unintended effects when rendering HTML.

Steps to Reproduce:
1. Use `open_in_browser` with an HTML response that does not contain a proper `<head>` tag.
2. Observe to see if the function does not handle this smoothly. 

Suggested Fix:
Edit the function to handle missing or properly formed `<head>` tags so that a correct `<base>` tag can be inserted.

AhmedNaeem3:
Hey, I can assist you in fixing this function and ensuring it works optimally.

vassilismylo:
Currently working on this!

PR #6564, opened 2024-11-29
Improve open_in_browser function's handling of base tags in response.py

ProgrammingCharm:
response.py for issue #6550

Updated response.py to be more versatile in handling base tags. 

This PR attempts to improve the handling of base tags in the open_in_browser function of scrapy/scrapy/utils/response.py. It addresses issue https://github.com/scrapy/scrapy/issues/6550 and issue I raised, where relative links were not being adequately handled in situations where there was inadequate base URL information. The solution in this PR is based around inserting a tag into the HTML response to handle relative links.

### DIFF
--- a/scrapy/utils/response.py
+++ b/scrapy/utils/response.py
@@ -17,6 +17,9 @@ from w3lib import html
 
 from scrapy.utils.python import to_bytes, to_unicode
 
+# Needed for open_in_browser
+from bs4 import BeautifulSoup
+
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable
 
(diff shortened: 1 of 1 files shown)


wRAR:
We won't accept a new dependency for this.

scrapy/utils/response.py
```diff
@@ -71,41 +74,30 @@ def _remove_html_comments(body: bytes) -> bytes:
...
-
-    .. code-block:: python
-
-        from scrapy.utils.response import open_in_browser
-
-
-        def parse_details(self, response):
-            if "item name" not in response.body:
-                open_in_browser(response)
+def open_in_browser(html_content, base_url):
```
wRAR, line 77: The signature is completely different.

PR #6567, opened 2024-12-02
Made path absolute to enable running pytest from any directory

emeryberger:
This minor change makes the reference to `tests/ignore.txt` absolute instead of relative, allowing tests to run from any other directory, rather than just inside the main directory.


wRAR:
Ooh. That's the reason for that annoying problem, thanks.

wRAR (approved):

Gallaecio:
Thanks!

Issue #6544, opened 2024-11-15
Default SpiderLoader doesn't take into account addons 


VMRuiz:
The default SpiderLoader class read the value of `SPIDER_MODULES` from settings before addons are configured for the project. 
This means that changes made by addons to that setting are irrelevant for commands like `scrapy list` 

I was expecting to be able to use addons to create plug-and-plays collections of spiders. The way I would do this is that addons will configure all required spider's paths into `SPIDER_MODULES` and then users would only have to import the addon in their `ADDONS` setting. 

It should be possible to create an `AddonManager` instance during `SpiderLoader` initilization and use it populate settings accordernly: 

```
    def __init__(self, settings: BaseSettings):
        # create basic crawler and populate settings
        AddonManager(get_crawler()).load_settings(settings)
        
        # keep the rest of the code as usual
        self.spider_modules: list[str] = settings.getlist("SPIDER_MODULES")
        self.warn_only: bool = settings.getbool("SPIDER_LOADER_WARN_ONLY")
        self._spiders: dict[str, type[Spider]] = {}
        self._found: defaultdict[str, list[tuple[str, str]]] = defaultdict(list)
        self._load_all_spiders()

guillermo-bondonno:
PR: https://github.com/scrapy/scrapy/pull/6568

GeorgeA92:
@VMRuiz 
The way how implemented updating settings from addons - is very similar to way how scrapy application reads `custom_settings` spider attribute. (By `update_settings` method).
As addon implementation require crawler object (almost as `custom_settings` spider attribute.. require specific spider) then I think addon settings - has the same restrictions as `custom_settings` spider attribute. Which means..  additional risks on attempt to change settings related to: logging, reactor, other settings on CrawlerRunner/CrawlerProcess level including settings related to spider loader.

As spider class can be easily changed by  `SPIDER_LOADER_CLASS` nothing prevents You to create custom Spider loader class where something like `update_settings` from addons called inside spider loader - I don't see a reason to make updates to default spiderloader class.

> I was expecting to be able to use addons to create plug-and-plays collections of spiders. The way I would do this is that addons will configure all required spider's paths into **SPIDER_MODULES** and then users would only have to import the addon in their ADDONS setting.

Not 100% sure but it looks like attempt to avoid/reduce consequences of https://github.com/scrapy/scrapy/issues/4975

PR #6568, opened 2024-12-02
Fix Default SpiderLoader doesn't take into account addons

guillermo-bondonno:
Fixes https://github.com/scrapy/scrapy/issues/6544

I had to do it at the crawler level because the settings passed to the spiderloader are frozen.
This achieves the goal of making new spider modules visible, but creates some not elegant behavior because each addon `update_settings` is ran twice. Which I still have to account for in the tests.
I couldn't find a way to make the spider visible without a dummy crawler approach, because ultimately the crawlers need a spider class to be created and we need spider loaders to know which classes are available.
I wonder if this couldn't cause issues if, for example, an Addon counts on the `from_crawler` method to properly run `update_settings`; @VMRuiz what are your thoughts?


scrapy/addons.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, crawler: Crawler) -> None:
         self.crawler: Crawler = crawler
         self.addons: list[Any] = []
 
-    def load_settings(self, settings: Settings) -> None:
+    def load_settings(self, settings: Settings | BaseSettings) -> None:
```
Gallaecio, line 25: What if we type it `BaseSettings` instead? (`Settings` is a subclass, and their API is identical, `Settings` only overrides `__init__`).

tests/test_addons.py
```diff
@@ -174,7 +174,7 @@ def update_settings(self, settings):
                 addon = LoggedAddon()
                 build_from_crawler_mock.return_value = addon
                 crawler = get_crawler(settings_dict=settings_dict)
-                logger_mock.info.assert_called_once_with(
+                logger_mock.info.assert_called_with(
```
Gallaecio, line 177: This is an issue, this message should not be logged more than once. We need to prevent the logging of this when the `SPIDER_MODULES` setting is loaded.

tests/test_crawler.py
```diff
@@ -937,15 +945,24 @@ def test_change_default_reactor(self):
 def test_log_scrapy_info(settings, items, caplog):
     with caplog.at_level("INFO"):
         CrawlerProcess(settings)
+    scrapy_version_msg = next(
+        record for record in caplog.records if record.getMessage().startswith("Scrapy ")
+    ).getMessage()
+    expected_scrapy_version_msg = (
+        f"Scrapy {scrapy.__version__} started (bot: scrapybot)"
+    )
```
Gallaecio, line 953: Same as above. We should prevent this PR to affect logs.

Gallaecio (commented):
Looking good!

@kmike @wRAR Any thoughts on the core of the approach, i.e. instantiating and using add-ons twice, first with a fake Crawler and Spider to fill the `SPIDER_MODULES` setting, then with the real parameters for the actual crawl? While hacky, I cannot think of a better approach; calling `update_settings` before loading the target spider seems like a more problematic approach.

kmike:
For example, if the crawler is not available yet (and so the addon can't be created yet), the pre-update method could be a classmethod, not a regular method.

GeorgeA92:
@guillermo-bondonno 
As previously mentioned on https://github.com/scrapy/scrapy/issues/6544#issuecomment-2512542207
attempt to change settings related to spiderloader is out of scope of addons (on it's current implementation as.. per crawler while spiderloader can be only one per process).
As it is possible to implement this functionality with custom spiderloader class set by `SPIDER_LOADER_CLASS` it is not clear why thi is needed as update for scrapy default spiderloader class.

kmike:
@GeorgeA92 we do have a use case - https://github.com/zytedata/zyte-spider-templates provides an addon which should add a few spiders to the project.

I don't think the current implementation is deliberately trying to avoid importing all the modules.

guillermo-bondonno:
> Would it make sense to have 2 methods in the addons (update_settings, and e.g. update_settings_early), and call them at different stages of crawler initialization? If we can avoid doing things twice, we should try, as it could lead to all kinds of issues.

@kmike I agree, what about requiring changes in SPIDER_MODULES to go into a specific classmethod of addons (update_spider_modules or something less specific if there's other use case), then running them from a classmethod in the AddonManager that doesn't have access/need a crawler instance?

VMRuiz:
> @guillermo-bondonno As previously mentioned on [#6544 (comment)](https://github.com/scrapy/scrapy/issues/6544#issuecomment-2512542207) attempt to change settings related to spiderloader is out of scope of addons (on it's current implementation as.. per crawler while spiderloader can be only one per process). As it is possible to implement this functionality with custom spiderloader class set by `SPIDER_LOADER_CLASS` it is not clear why thi is needed as update for scrapy default spiderloader class.

From a software design perspective, I agree with you, but requiring users to create a custom class solely to support addons creates a poor user experience. I like @kmike’s suggestion of introducing a new method that runs earlier to prevent side effects in `update_settings` methods and without requiring `crawler` to be instantiated. This approach would maintain backward compatibility and avoid breaking existing addons.

guillermo-bondonno:
@VMRuiz maybe I wasn't clear, I was proposing kinda the same as @kmike, I'll update the PR

scrapy/spiderloader.py
```diff
@@ -76,6 +78,11 @@ def _load_all_spiders(self) -> None:
                     raise
         self._check_name_duplicates()
 
+    def _load_spider_modules_from_addons(self, settings: BaseSettings) -> None:
+        settings.frozen = False
```
Gallaecio, line 82: Should we move these to `AddonManager.load_spider_modules`?

scrapy/addons.py
```diff
@@ -52,3 +52,12 @@ def load_settings(self, settings: Settings) -> None:
             },
             extra={"crawler": self.crawler},
         )
+
+    @classmethod
+    def load_spider_modules(cls, settings):
```
Gallaecio, line 57: Since it may be useful for other settings, I wonder if we could make it more generic, less specific to `SPIDER_MODULES`. e.g. name it `load_early_settings`, and call it right after https://github.com/scrapy/scrapy/blob/402500b164efc01257679247d3dd1628a5f90f5e/scrapy/crawler.py#L295

scrapy/addons.py
```diff
@@ -52,3 +52,12 @@ def load_settings(self, settings: Settings) -> None:
...
         )
+
+    @classmethod
+    def load_early_settings(cls, settings: BaseSettings):
+        for clspath in build_component_list(settings["ADDONS"]):
+            try:
+                addoncls = load_object(clspath)
+                addoncls.update_spider_modules(settings)
+            except AttributeError:
+                pass
```
Gallaecio, line 63: Why this exception?

I’m thinking we should not capture any exception here, in line with the regular method, which only captures `NotConfigured` (which we should probably not capture here, it would be weird for users to have to raise that exception from 2 places instead of simply doing nothing in the class method).

Gallaecio (commented):
It’s looking good. What about adding some docs about this new class method to the add-ons topic?

Looking at the docs, I’m also thinking that an add-on with only the new class method may make sense, so we might want to also use `hasattr` with the regular method, and only call it if it exists.

docs/topics/addons.rst
```diff
@@ -32,7 +32,7 @@ This is an example where two add-ons are enabled in a project's
 Writing your own add-ons
 ========================
 
-Add-ons are Python classes that include the following method:
+Add-ons are Python classes that include one of the following methods:
```
Gallaecio, line 35:
```suggestion
Add-ons are Python classes that include one or both of the following methods:
```

docs/topics/addons.rst
```diff
@@ -45,6 +45,17 @@ Add-ons are Python classes that include the following method:
     :param settings: The settings object storing Scrapy/component configuration
     :type settings: :class:`~scrapy.settings.Settings`
 
+.. classmethod:: update_early_settings(cls, settings)
+
+    This method is called before the initialization of the
+    :class:`~scrapy.crawler.Crawler`. Here you can do the same updates as in the
+    ``update_settings`` method, but won't have access to the crawler instance.
+    Some settings, like :setting:`SPIDER_MODULES`, need to be updated here
+    becausethey are used during the initialization of the crawler.
```
Gallaecio, line 54:
```suggestion
    because they are used during the initialization of the crawler.
```

Gallaecio (approved):

scrapy/addons.py
```diff
@@ -52,3 +53,19 @@ def load_settings(self, settings: Settings) -> None:
...
+    @classmethod
+    def load_early_settings(cls, settings: BaseSettings):
+        """Update early settings that do not require a crawler instance, such as SPIDER_MODULES.
+
+        Similar to the load_settings method, this loads each add-on configured in the
+        ``ADDONS`` setting and calls their 'update_early_settings' class method if present.
+        This method doesn't have access to the crawler instance or the addons list.
+
+        :param settings: The :class:`~scrapy.settings.BaseSettings` object from \
+            which to read the early add-on configuration
```
wRAR, line 66:
```suggestion
            which to read the early add-on configuration
        :type settings: :class:`~scrapy.settings.Settings`
```

wRAR (approved):
Well, I think it's a nice clean solution where I thought one is not possible :)

docs/topics/addons.rst
```diff
@@ -45,6 +45,17 @@ Add-ons are Python classes that include the following method:
     :param settings: The settings object storing Scrapy/component configuration
     :type settings: :class:`~scrapy.settings.Settings`
 
+.. classmethod:: update_early_settings(cls, settings)
```
kmike, line 48: I don't have a strong opinion, but "update_settings_early" sounds a bit more clear. 

The current name is saying the method will update a thing called "early settings". With "update_settings_early" it'd be "update the settings, but do it (when?) early", which makes more sense to me; we don't introduce a new concept of "early settings".

kmike (approved):

docs/topics/addons.rst
Gallaecio: I think this is a good point to clarify before merging.

My suggestion was based on the fact that there are indeed “early settings”, settings that, if updated, need to be updated before the crawler is built in order for them to be applied effectively. I think there is more than 1 such settings, I think at some point @wRAR looked into them.

Maybe we should document a list of settings that should be updated with this class method instead of the other method. I wonder if for some of those we would need to move this method even earlier.

People could indeed use this method to update other settings. And I don’t see a reason to try to prevent that. But I think documenting the settings that you are expected to update through this class method makes it OK for us to break the update of unrelated settings in the future if needed (if that could even happen).

wRAR: There is a list of early and/or singleton settings at the end of https://docs.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process (it may be incomplete and/or outdated but is a good start).

kmike: What should be the name for these settings? Is it about being "Global", not spider-specific, "singleton"?

kmike: settings on CrawlerProcess level, not on spider level?

It also seems the concept of "early" settings ("settings that, if updated, need to be updated before the crawler is built in order for them to be applied effectively") may be different from "global" settings. Are they the same settings? Or are some settings global, but not early - or early, but not global?

Gallaecio: I’ve pushed a change to rename ``early`` to ``pre_crawler``, and have the docs describe what “pre-crawler settings” and “rector settings” are, based on the existing docs. Thoughts?

wRAR: I think these two classes overlap only partially. At a quick glance, all settings listed on that page are global (that's why they are listed in the doc about multiple spiders in the same process), but only the first set of examples is needed before the `Crawler` instance is created (= `CrawlerRunner.crawl()` is called).

wRAR: Well no, I think that first set of examples (which is `SPIDER_LOADER_CLASS` and the settings used by the spider loader) is not a per-process thing, only the other two are.

wRAR: @Gallaecio yeah, looks like you also found the same categories. OTOH I'm not sure if the only things that can be per-process are things tied to the reactor. I'm even not sure if there are no such settings in the current Scrapy. Thoughts?

Gallaecio: Documentation-wise, even if there may be more settings unrelated to the reactor that are also process-specific settings, or even some other category of special settings, I don’t think we need to worry about it here.

If later on we find other built-in settings that need to be treated similar to reactor settings, we can rename “reactor settings” to something else, or group “reactor settings” and a new group of settings under a parent “process settings” or similar.

My only concern here is agreeing on the name for settings that should be changed by the new class method when implementing an add-on, since we are putting that name on the API. Other names, like “reactor settings”, have no impact on the API, so I feel like we can rename and refactor those at will at any point.

docs/topics/settings.rst
```diff
@@ -234,6 +250,61 @@ example, proper setting names for a fictional robots.txt extension would be
...
+**Reactor settings** are settings tied to the :doc:`Twisted reactor
+<twisted:core/howto/reactor-basics>`.
+
+These settings can be defined from a spider. However, because only 1 reactor
+can be used per process, these settings cannot use a different value per spider
+when :ref:`running multiple spiders in the same process
+<run-multiple-spiders>`.
+
+In general, if different spiders define different values, the first defined
+value is used. However, if two spiders request a different reactor, an
```
kmike, line 288: Hm, I didn't know that :( I was expecting the last value to be used, e.g. for dns options, or for the threadpool size.

kmike (approved):
The tests are failing, but it doesn't seem to be related to the changes in the PR. We'd need to fix them separately.

docs/topics/settings.rst
wRAR: OK, so two fun facts:

* `DNS_RESOLVER` and `REACTOR_THREADPOOL_MAXSIZE` are only honored by `CrawlerProcess`, if you use `CrawlerRunner` they will be ignored.
* `DNS_RESOLVER` and `REACTOR_THREADPOOL_MAXSIZE` are read directly from the settings passed to `CrawlerProcess`, which are *not* modified by `self.spidercls.update_settings()` in `Crawler.__init__()` so they cannot be changed via per-spider settings.

As for "the first value is used", it's true for `TWISTED_REACTOR` and `ASYNCIO_EVENT_LOOP` because the first `CrawlerProcess.crawl()` calls `install_reactor()`.

Should we make an issue to clarify the docs about this further? Should we keep the code as is?

Gallaecio: > Should we make an issue to clarify the docs about this further? 

+1, *maybe* a good first issue.

> Should we keep the code as is?

I don’t think it is urgent to change it, but it would be nice to create a separate, long-term ticket to discuss improvements to this, in case someone has a strong proposal.

PR #6576, opened 2024-12-09
Add flake8 plugin rules to ruff

wRAR:
Should we remove flake8 from pre-commit (and remove `.flake8`) or let it be duplicated? `flake8-string-format` is not ported to ruff yet, but it's likely not that useful, especially as we don't have many `str.format()` calls.


Gallaecio:
+1 to remove flake8.

Gallaecio (approved):

PR #6581, opened 2024-12-12
Ruff: enable other useful rules, turn on autofixing

wRAR:


scrapy/link.py
```diff
@@ -39,7 +37,7 @@ def __init__(
         self.fragment: str = fragment
         self.nofollow: bool = nofollow
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
```
Gallaecio, line 40: Interesting.

scrapy/utils/asyncgen.py
```diff
@@ -7,10 +7,7 @@
 
 
 async def collect_asyncgen(result: AsyncIterable[_T]) -> list[_T]:
-    results = []
-    async for x in result:
-        results.append(x)
-    return results
+    return [x async for x in result]
```
Gallaecio, line 10: TIL

tests/test_spidermiddleware.py
```diff
@@ -290,10 +289,7 @@ def process_spider_output(self, response, result, spider):
 
 class ProcessSpiderOutputCoroutineMiddleware:
     async def process_spider_output(self, response, result, spider):
-        results = []
-        for r in result:
-            results.append(r)
-        return results
+        return result
```
Gallaecio, line 292: Interesting. It is technically not the same, since result is not a list but an iterator. On the other hand, in this specific case it is most likely OK.

scrapy/link.py
wRAR: Yeah. https://docs.astral.sh/ruff/rules/any-eq-ne-annotation/ says "The Python documentation recommends the use of object to "indicate that a value could be any type in a typesafe manner". Any, on the other hand, should be seen as an "escape hatch when you need to mix dynamically and statically typed code"." which makes sense and may be useful in other places too.

Gallaecio (approved):

tests/test_spidermiddleware.py
wRAR: I rewrote this myself, the linter suggests a copy via `list(result)`.

wRAR: (Well, not a copy, as it's not necessarily a list)

PR #4203, opened 2019-12-01
Use pprint.pformat on versions log message

grammy-jiang:
This commit use pprint.pformat on overridden settings to keep consistent
with scrapy.middleware, for example:
Original:
```
2019-12-01 11:35:40 [scrapy.utils.log] INFO: Versions: lxml 4.3.4.0, libxml2 2.9.9, cssselect 1.0.3, parsel 1.5.1, w3lib 1.20.0, Twisted 19.2.1, Python 3.7.5 (default, Oct 17 2019, 12:16:48) - [GCC 9.2.1 20190827 (Red Hat 9.2.1-1)], pyOpenSSL 19.0.0 (OpenSSL 1.1.1c  28 May 2019), cryptography 2.7, Platform Linux-5.3.7-301.fc31.x86_64-x86_64-with-fedora-31-Thirty_One
```
Now:
```
2019-12-01 11:04:22 [scrapy.utils.log] INFO: Versions:
{'Platform': 'Linux-5.3.7-301.fc31.x86_64-x86_64-with-fedora-31-Thirty_One',
 'Python': '3.7.5 (default, Oct 17 2019, 12:16:48) - [GCC 9.2.1 20190827 (Red '
           'Hat 9.2.1-1)]',
 'Twisted': '19.2.1',
 'cryptography': '2.7',
 'cssselect': '1.0.3',
 'libxml2': '2.9.9',
 'lxml': '4.3.4.0',
 'parsel': '1.5.1',
 'pyOpenSSL': '19.0.0 (OpenSSL 1.1.1c  28 May 2019)',
 'w3lib': '1.20.0'}
```

PR #6582, opened 2024-12-13
Allow customizing logged software versions

Gallaecio:
Resolves: #4203

### DIFF
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -13,7 +13,7 @@ from twisted.python.failure import Failure
 
 import scrapy
 from scrapy.settings import Settings, _SettingsKeyT
-from scrapy.utils.versions import scrapy_components_versions
+from scrapy.utils.versions import get_versions
 
 if TYPE_CHECKING:
 
@@ -178,9 +178,7 @@ def log_scrapy_info(settings: Settings) -> None:
     components = settings.getlist("LOG_VERSIONS")
     if not components:
         return
-    versions = pprint.pformat(
-        dict(scrapy_components_versions(components)), sort_dicts=False
-    )
+    versions = pprint.pformat(dict(get_versions(components)), sort_dicts=False)
     logger.info(f"Versions:\n{versions}")
 
 
--- a/scrapy/utils/log.py
+++ b/scrapy/utils/log.py
@@ -175,10 +175,10 @@ def log_scrapy_info(settings: Settings) -> None:
         "Scrapy %(version)s started (bot: %(bot)s)",
         {"version": scrapy.__version__, "bot": settings["BOT_NAME"]},
     )
-    components = settings.getlist("LOG_VERSIONS")
-    if not components:
+    software = settings.getlist("LOG_VERSIONS")
+    if not software:
         return
-    versions = pprint.pformat(dict(get_versions(components)), sort_dicts=False)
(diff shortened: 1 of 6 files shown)


wRAR:
Related: #4203

scrapy/utils/versions.py
```diff
@@ -1,31 +1,48 @@
...
-        ("parsel", parsel.__version__),
-        ("w3lib", w3lib.__version__),
-        ("Twisted", twisted.version.short()),
-        ("Python", sys.version.replace("\n", "- ")),
-        ("pyOpenSSL", get_openssl_version()),
-        ("cryptography", cryptography.__version__),
-        ("Platform", platform.platform()),
-    ]
+def scrapy_components_versions(
+    components: list | None = None,
```
kmike, line 38: What do you think about not adding this argument to scrapy_components_version? In this case, implementation will be different: it'd call `get_versions(_DEFAULT_SOFTWARE)`. We still should deprecate it probably.

scrapy/utils/log.py
```diff
@@ -174,12 +175,11 @@ def log_scrapy_info(settings: Settings) -> None:
         "Scrapy %(version)s started (bot: %(bot)s)",
         {"version": scrapy.__version__, "bot": settings["BOT_NAME"]},
     )
-    versions = [
-        f"{name} {version}"
-        for name, version in scrapy_components_versions()
-        if name != "Scrapy"
-    ]
-    logger.info("Versions: %(versions)s", {"versions": ", ".join(versions)})
+    software = settings.getlist("LOG_VERSIONS")
```
Laerte, line 178: Looks good! Just wondering if this variable should be in plural.

Gallaecio: ”software” is plural, singular would be “pieces of software”, “software components”, or similar. I went with “items” to keep things short.

kmike (approved):

scrapy/utils/log.py
Laerte: TIL, thank you @Gallaecio

Laerte (approved):

scrapy/utils/versions.py
```diff
@@ -1,31 +1,46 @@
...
-        ("Twisted", twisted.version.short()),
-        ("Python", sys.version.replace("\n", "- ")),
-        ("pyOpenSSL", get_openssl_version()),
-        ("cryptography", cryptography.__version__),
-        ("Platform", platform.platform()),
-    ]
+    warn(
+        (
+            "scrapy.utils.versions.scrapy_components_versions is deprecated, "
+            "use scrapy.utils.versions.get_versions instead."
```
wRAR, line 41:
```suggestion
            "scrapy.utils.versions.scrapy_components_versions() is deprecated, "
            "use scrapy.utils.versions.get_versions() instead."
```

scrapy/utils/versions.py
```diff
@@ -1,31 +1,46 @@
...
+        ),
+        ScrapyDeprecationWarning,
+        stacklevel=2,
+    )
+    return get_versions(_DEFAULT_SOFTWARE)
```
wRAR, line 46: Should this call just `get_versions()`, or should the `get_versions()` arg become mandatory? What's your plan for after `scrapy_components_versions()` is removed?

Issue #6583, opened 2024-12-14
`scrapy.utils.url.parse_url()` duplicates `w3lib.url.parse_url()`


wRAR:
`scrapy.utils.url` imports `w3lib.url.*` but since 2016 (w3lib 1.15.0) both have `parse_url()`, and with the same implemenation. So we can drop the Scrapy one now. Related: #2186, #4577

PR #6584, opened 2024-12-14
Add pylint rules to ruff, refresh the ignore list of pylint itself

wRAR:
Only a subset of pylint checks is implemented in ruff so we should keep both.

### DIFF
--- a/scrapy/__init__.py
+++ b/scrapy/__init__.py
@@ -31,7 +31,7 @@ version_info = tuple(int(v) if v.isdigit() else v for v in __version__.split("."
 
 def __getattr__(name: str):
     if name == "twisted_version":
-        import warnings
+        import warnings  # pylint: disable=reimported
 
         from twisted import version as _txv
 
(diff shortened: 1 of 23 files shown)


scrapy/__init__.py
```diff
@@ -31,7 +31,7 @@
 
 def __getattr__(name: str):
     if name == "twisted_version":
-        import warnings
+        import warnings  # pylint: disable=reimported
```
wRAR, line 34: I assume it's some magic around `__getattr__` that doesn't execute the top level imports but I didn't check.

scrapy/utils/url.py
```diff
@@ -50,7 +50,9 @@ def url_has_any_extension(url: UrlT, extensions: Iterable[str]) -> bool:
     return any(lowercase_path.endswith(ext) for ext in extensions)
 
 
-def parse_url(url: UrlT, encoding: str | None = None) -> ParseResult:
+def parse_url(  # pylint: disable=function-redefined
```
wRAR, line 53: #6583

Gallaecio (approved):

Issue #4577, opened 2020-05-14
Deprecate w3lib objects importable from scrapy.utils.url


Gallaecio:

jay24rajput:
Does this issue include only removing the following code line from [scrapy.util.url](https://github.com/scrapy/scrapy/blob/314adf6c386848f35207c734dfe84c0d9d4e9e6a/scrapy/utils/url.py#L16)?
```from w3lib.url import _safe_chars, _unquotepath```

**Comment by Gallaecio** [comment]
Nope, it’s a bit more complex.

There’s a `from w3lib.url import *` import line in that file. This issue is about switching that to an explicit import, and somehow logging a deprecation warning when those objects are imported from this file (as opposed to being imported from `w3lib`) or used after importing them from this file.

For functions, it should not be hard. I think we already have a deprecation decorator somewhere. For module-level variables, things can be more complicated.

**Comment by Gallaecio** [comment]
Looks like we can use https://stackoverflow.com/a/48242860 now.

**Comment by Laerte** [comment]
@Gallaecio I take a look and from w3lib there are 2 functions: `add_or_replace_parameter` and`any_to_uri` that we are using from wildcard import, we also have `parse_url` that we define again but we can use from w3lib and emit a warning when someone is attempting to use directly. Can you check if my changes attend what needs to be done here before i open the PR?

https://github.com/scrapy/scrapy/compare/master...Laerte:scrapy:master?diff=split&w=

Thanks!

**Comment by Gallaecio** [comment]
I did not realize we still had a function (`parse_url`) that is defined in both, and I think the “fix” there is to warn about its use without the need to inspect the callstack; if there are internal calls to it, we can switch those to the w3lib alternative.

Besides that, I think the idea here is to replace `from w3lib.url import *` with an explicit import of [everything that would import](https://github.com/scrapy/w3lib/blob/73bfd652017d8923c08941a89a018d009f82d4e8/w3lib/url.py#L515-L528),  but making names private (e.g. `from w3lib.url import add_or_replace_parameter as _add_or_replace_parameter`), and then having some code that defined the public name in the module in a way that causes a deprecation, e.g.

```
from scrapy.utils.decorators import deprecated

@deprecated
def add_or_replace_parameter(*args, **kwargs):
    return _add_or_replace_parameter(*args, **kwargs)
```

[...]

PR #6586, opened 2024-12-16
Deprecate w3lib objects importable from scrapy.utils.url

Laerte:
Close: #4577
Close: #6583

We will emit a warning everytime someone import functions of `w3lib` from `scrapy.utils.url`

### DIFF
--- a/scrapy/utils/url.py
+++ b/scrapy/utils/url.py
@@ -1,29 +1,30 @@
 """
 This module contains general purpose URL functions not found in the standard
 library.
-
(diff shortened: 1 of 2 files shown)


Laerte:
Tests failures on macOS is not related to the changes on this PR. **Checks / checks (3.13, pylint)** failure seems due to a conflict with styles between Ruff and Pylint.

Gallaecio (approved):

wRAR (approved):

PR #6589, opened 2024-12-22
Added check and re-request for 201 response in media_downloaded

GauravR31:
Added a check for 201 response, and checking for the 'Location' header in the 201 response. If the 'Location' header exists, a new request is made for the data.

Also added a corresponding unit test.

Fixes [Issue 1615](https://github.com/scrapy/scrapy/issues/1615)
Referenced the discussions on [PR 1806](https://github.com/scrapy/scrapy/pull/1806) and [6309](https://github.com/scrapy/scrapy/pull/6309)

### DIFF
--- a/tests/test_pipeline_files.py
+++ b/tests/test_pipeline_files.py
@@ -147,6 +147,32 @@ class FilesPipelineTestCase(unittest.TestCase):
             p.stop()
 
     @defer.inlineCallbacks
+    def test_media_download_201(self):
+        item_url = "http://example.com/file.pdf"
+        item = _create_item_with_files(item_url)
+        patchers = [
+            mock.patch.object(
+                FilesPipeline,
+                "get_media_requests",
+                return_value=[
+                    Request(
+                        item_url,
+                        meta={"response": Response(item_url, status=201, body=b"data")},
+                    )
+                ],
+            )
+        ]
+
+        for p in patchers:
+            p.start()
+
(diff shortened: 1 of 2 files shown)


tests/test_pipeline_files.py
```diff
@@ -146,6 +146,32 @@ def test_file_not_expired(self):
         for p in patchers:
             p.stop()
 
+    @defer.inlineCallbacks
+    def test_media_download_201(self):
+        item_url = "http://example.com/file.pdf"
+        item = _create_item_with_files(item_url)
```
wRAR, line 152: This alread creates the expected item, so the test passes even though it doesn't call the code you added.

GauravR31: Could you let me know how I can mock a 201 response? This is my first time contributing.

wRAR: The problem is not about mocking a response, it's about checking that the pipeline actually worked. You are only checking that the `"file_urls"` field is correct, but that field is the input to the pipeline, not the output, it's supposed to always exist. You should check the `"files"` field like other tests do.

GauravR31: I modified it to check for the checksum in "files", does it look correct now?

wRAR: Your code still doesn't run, which should have been easy to check for you. Please do this next time.

wRAR:

PR #6599, opened 2025-01-01
Fix calculate_final_stats method

devfox-se:
I noticed that the values are always null and figured out there is a typo in the stats key name


wRAR:
Please run tests locally before submitting.

wRAR (approved):
Thanks!

PR #6606, opened 2025-01-04
Fix _pop_command_name and include tests

### LINKED ISSUE #6600: Investigate off-by-1 in `scrapy.cmdline._pop_command_name()`
It looks like `del argv[i]` removes the wrong item in `scrapy.cmdline._pop_command_name()` but as we don't seem to see any problems because of this it's worth investigating what exactly happens here and either fixing or refactoring the code.

protokoul:
This PR resolves #6600. I have also created a test file called test_cmdline_pop_command_name.py within tests/test_cmdline


wRAR:
I think the location for the tests should be in some of the existing `tests/test_foo.py`, as `tests/test_cmdline/` is for running tests with a settings.py, via `tests.test_cmdline.CmdlineTest._execute()`, and this test doesn't require that. `tests/test_commands.py` is about slightly different things but I'm not sure if there are better places than that.

protokoul:
I thought about writing the file `test_cmdline_pop_command_name.py` under `tests` directory but then I felt that the name is too specific for testing just one function from `scrapy.cmdline`. Is it ok if I move `test_cmdline_pop_command_name.py` from `tests/test_cmdline` to `tests` and leave the name as is?

wRAR:
I don't think tests for a single private function deserve a separate file, that's why I suggested using an existing file.

wRAR (approved):
Thanks!

PR #6608, opened 2025-01-05
 Flexible severity of logging level when items are dropped

### LINKED ISSUE #6603: Allow changing the log severity of DropItem
By default, when an item is dropped, a `WARNING` message is logged. [It would be nice to make that more flexible](https://github.com/zytedata/zyte-common-items/pull/126#discussion_r1900692613).

protokoul:
This PR is created to resolve #6603. I have added a new default setting called `DEFAULT_DROPITEM_LOG_LEVEL`, set to `WARNING` by default. I have modified the exception for `DropItem` to use the default log value unless severity is provided by the component. Modified the documentation under docs/topics/settings.rst.

This PR is currently drafted because tests are not written yet. Please suggest whether I should create a new test file or can I use any existing test file to include the tests?


Gallaecio:
Could you look into your changes to the settings docs? It seems you changed the line endings of all lines (maybe by editing them on Windows or macOS with an editor that does not respect the existing line ending characters?), which makes the PR diff hard do read.

protokoul:
My bad. I am using VS Code on Windows and checked the EOL settings. I have updated it to use LF ending instead of CRLF, and pushed the new changes so the PR diff looks better now.

I noticed that when I changed EOL settings in VS Code to use LF and modified the .rst file, then git diff only showed the lines that I changed, but when I ran git add and then git commit, the blacken-docs hook in the pre-commit modified the .rst file in a way that all lines were changed when I checked again with git diff. I am assuming that other contributors are only using linux and if no one has observed this in their machine, then this might be windows specific. I checked blacken-docs repo and saw in its [.pre-commit-config.yaml](https://github.com/adamchainz/blacken-docs/blob/main/.pre-commit-config.yaml) that there is something called end-of-file-fixer from pre-commit-hooks repo. I tried to modify the .pre-commit-config.yaml file in scrapy's repo but couldn't get it to work. Maybe `*.rst text eol=lf` can be added to .gitattribuutes to enforce that .rst files use only LF line ending. If there is a better way to do it via hooks in yaml file, I would like to know about it.

protokoul:
I realized that the commits from my previous PR 6606 are also included here and 6606 is not yet merged. If I remove those commits from PR 6608, then is it possible that if 6608 gets merged, then it may undo the changes I had included in PR 6606?

Gallaecio:
> If I remove those commits from PR 6608, then is it possible that if 6608 gets merged, then it may undo the changes I had included in PR 6606?

While it’s a valid point, if I am the one merging I will use “squash and merge”, which I would expect not to cause that issue. If for whatever reason it does happen, though, we’ll handle it, do not worry about it.

protokoul:
Sure, then I will remove those unrelated commits from this PR.

protokoul:
I have removed the unrelated commits from this PR. I messed up the first rebase which is why I think some CI jobs were triggered. The latest rebase reflects correct and relevant changes and I have also added tests in an existing test_logformatter.py file.

Gallaecio:
[Please, run `pre-commit run --all`](https://docs.scrapy.org/en/master/contributing.html#pre-commit).

docs/topics/settings.rst
```diff
@@ -418,6 +418,36 @@ This setting also affects :setting:`DOWNLOAD_DELAY` and
 :ref:`topics-autothrottle`: if :setting:`CONCURRENT_REQUESTS_PER_IP`
 is non-zero, download delay is enforced per IP, not per domain.
 
+.. setting:: DEFAULT_DROPITEM_LOG_LEVEL
```
Gallaecio, line 421: No strong opinion, but I wonder if we should drop the `DEFAULT` from the name. It’s more precise as it is, it is indeed a default value that components can override, but I wonder if such precision in the name is necessary. Thoughts?

docs/topics/settings.rst
```diff
@@ -418,6 +418,36 @@ This setting also affects :setting:`DOWNLOAD_DELAY` and
...
+
+DEFAULT_DROPITEM_LOG_LEVEL
+--------------------------
+
+Default: ``'WARNING'``
```
Gallaecio, line 426:
```suggestion
Default: ``"WARNING"``
```

(black)

docs/topics/settings.rst
```diff
@@ -418,6 +418,36 @@ This setting also affects :setting:`DOWNLOAD_DELAY` and
...
+
+This setting allows you to configure the default log severity level for dropped items in Scrapy. When an item is dropped using the ``DropItem`` exception, the severity level of the corresponding log message will be determined by this setting, unless a specific severity is explicitly provided in the exception.
```
Gallaecio, line 428: ``` ``DropItem`` ``` → ``` :exc:`scrapy.exceptions.DropItem` ```

Also, it might be useful to mention item pipelines (with a ref pointing to their docs) as an example of a Scrapy component that drops an item.

docs/topics/settings.rst
```diff
@@ -418,6 +418,36 @@ This setting also affects :setting:`DOWNLOAD_DELAY` and
...
+
+Available options are the standard logging levels: ``'DEBUG'``, ``'INFO'``, ``'WARNING'``, ``'ERROR'``, and ``'CRITICAL'``.
+
+Example:
+
+.. code-block:: python
+
+   # settings.py
+   DEFAULT_DROPITEM_LOG_LEVEL = "INFO"
```
Gallaecio, line 437:
```suggestion
.. code-block:: python
    :caption: settings.py

    DEFAULT_DROPITEM_LOG_LEVEL = "INFO"
```

scrapy/exceptions.py
```diff
@@ -58,6 +58,14 @@ def __init__(self, *, fail: bool = True):
 class DropItem(Exception):
     """Drop item from the item pipeline"""
 
+    def __init__(self, message, severity=None):
+        from scrapy.utils.project import get_project_settings
```
Gallaecio, line 62: This approach is problematic, spider-specific settings would not be taken into account.

I think we need to move the logic to read the setting to the code that captures the exception.

docs/topics/settings.rst
protokoul: While going through the settings in alphabetical order, the first setting I saw starting with D is DEFAULT_ITEM_CLASS and so I used the name DEFAULT_DROPITEM_LOG_LEVEL. It is correct that every setting will have a default value so it's not really mandatory to use DEFAULT as a prefix here. Would it also mean to remove "DEFAULT_" from DEFAULT_ITEM_CLASS and DEFAULT_REQUEST_HEADERS?

Gallaecio: I would definitely not touch those settings as part of these changes, and I think their names are good:

- I think the prefix makes sense on `DEFAULT_REQUEST_HEADERS`, as I think `REQUEST_HEADERS` could be confusing (not clear if those override Request.headers; default clarifies that they do not).
- `DEFAULT_ITEM_CLASS` maps `ItemLoader.default_item_class`.

`DEFAULT_DROPITEM_LOG_LEVEL` does feel a bit like `DEFAULT_REQUEST_HEADERS`. If it were `DROPITEM_LOG_LEVEL`, I guess it might not be clear what takes priority when you set the severity on the exception. So you have convinced me :slightly_smiling_face:.

scrapy/exceptions.py
```diff
@@ -58,6 +58,14 @@ def __init__(self, *, fail: bool = True):
 class DropItem(Exception):
     """Drop item from the item pipeline"""
 
+    def __init__(self, message, severity=None):
```
Gallaecio, line 61: Should we call it `log_level`, in line with the setting?

protokoul: Yes, that is correct. Let me rename this.

docs/topics/settings.rst
protokoul: This is how it looks after using `:caption:`. Shall I go ahead with this one?

image

scrapy/exceptions.py
protokoul: Thank you for sharing the file name. This was very helpful.

scrapy/core/scraper.py
```diff
@@ -372,6 +372,12 @@ def _itemproc_finished(
         if isinstance(output, Failure):
             ex = output.value
             if isinstance(ex, DropItem):
+                if ex.log_level is None:
+                    ex.log_level = spider.crawler.settings.get(
+                        "DEFAULT_DROPITEM_LOG_LEVEL", "WARNING"
+                    )
+                else:
+                    ex.log_level = ex.log_level.upper()
```
Gallaecio, line 380: Not sure that we need this, but if anything it should be `elif isinstance(ex.log_level, str):`, because I think we do want to support actual variables (`logging.INFO`) being used instead of strings.

protokoul: Yes, that is correct. I think I should also write a list of valid logging levels in the code and check if the log level provided by the component is valid instead of just some random string.

protokoul: If I remove `severity = getattr(logging, log_level.upper())` and set `"level": log_level`, then one of my test cases called test_log_formatter() fails.

Also, I realized that `.upper()` in `severity = getattr(logging, log_level.upper())` is unnecessary so it can be removed and the test cases still pass.

protokoul: The value returned from other functions such as `crawled()`, `scraped()`, `item_error()` inside the `LogFormatter` class in `src/logformatter.py` have a fixed value assigned to `"level"`. Same used to be the case for `dropped()` but as it is going to be more flexible, I used the `getattr()` to fetch the correct value of logging.

Gallaecio: We can make it the expectation for it to be numerical, and it makes sense, it would ensure backward compatibility.

But then I think we need to enforce through type hints that the value of the exception must also be of that type (or None), and instead of doing conversion/resolving with `getattr`, require logging.X to be used.

Gallaecio: > I have tried both logging.info and logging.INFO in the python interpreter, Python will not raise an error in either case

```
>>> logging.log(logging.INFO, "foo")
>>> logging.log(logging.info, "foo")
  File "<python-input-8>", line 1, in <module>
    logging.log(logging.info, "foo")
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/logging/__init__.py", line 2215, in log
    root.log(level, msg, *args, **kwargs)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/logging/__init__.py", line 1585, in log
    raise TypeError("level must be an integer")
TypeError: level must be an integer
```

protokoul: Oh. My bad. I was directly running `logging.info` and `logging.INFO` in the interpreter to see what does it return and was not passing it to the `logger.log()`, I didn't think that eventually the level will be passed to `logger.log()`. I understand now that we can drop `.upper()` and Python will be able to handle it when it is passed to logger.log(). Let me implement the changes.

scrapy/logformatter.py
```diff
@@ -125,7 +125,7 @@ def dropped(
         # that is used in DropItem exception, so a default value is provided
         # inside the first getattr() below
         log_level = getattr(exception, "log_level", "WARNING")
-        severity = getattr(logging, log_level.upper())
+        severity = getattr(logging, log_level)
```
Gallaecio, line 128:
```suggestion
        if isinstance(severity, str):
            severity = getattr(logging, log_level)
```

In case they pass `logging.INFO` or similar.

Gallaecio:
I’ll push some changes shortly, to see what you think.

Gallaecio:
Done, let me know what you think.

scrapy/logformatter.py
```diff
@@ -120,14 +120,12 @@ def dropped(
         spider: Spider,
     ) -> LogFormatterResult:
         """Logs a message when an item is dropped while it is passing through the item pipeline."""
-        # There is a test case called test_dropped under tests/test_logformatter.py
-        # that raises base exception which will not have the attribute log_level
-        # that is used in DropItem exception, so a default value is provided
-        # inside the first getattr() below
-        log_level = getattr(exception, "log_level", "WARNING")
-        severity = getattr(logging, log_level)
+        if (level := getattr(exception, "log_level", None)) is None:
```
protokoul, line 123: This is great. I understood that moving this logic from `scraper.py` to `logformatter.py` makes the testing easier as `dropped()` can be called directly instead of mocking a private function in the tests.

docs/topics/settings.rst
```diff
@@ -425,13 +425,20 @@ DEFAULT_DROPITEM_LOG_LEVEL
...
 
-For example, an :ref:`item pipeline <topics-item-pipeline>` is a Scrapy component that often uses the :exc:`scrapy.exceptions.DropItem` exception to discard unwanted or invalid items during processing.
+When an item is dropped by raising :exc:`scrapy.exceptions.DropItem` from the
+:func:`process_item` method of an :ref:`item pipeline <topics-item-pipeline>`,
+a message is logged, and by default its log level is the one configured in this
+setting.
 
-Available options are the standard logging levels: ``"DEBUG"``, ``"INFO"``, ``"WARNING"``, ``"ERROR"``, and ``"CRITICAL"``.
+You may specify this log level as an integer (e.g. ``20``), as a log level
+constant (e.g. ``logging.INFO``) or as a string with the name of a log level
```
protokoul, line 436: I agree. This approach makes the `log_level` attribute even more flexible. I didn't even think that I would pass a constant like `logging.INFO` directly because as a user I thought I would only pass a string and let scrapy handle the `log_level`, but this approach is a lot better.

wRAR (approved):

PR #6609, opened 2025-01-05
Switch CI to codecov/codecov-action and enable it on Windows.

### LINKED ISSUE #5360: Use coverage data from Windows runs in addition to Ubuntu ones
We have Windows-specific code and tests for it but we don't use coverage data from them so that code is reported as not covered. It would be good to fix this if possible. See e.g. https://github.com/scrapy/scrapy/pull/5315/files

wRAR:
Fixes #5360

### DIFF
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -27,4 +27,4 @@ jobs:
         tox -e py
 
     - name: Upload coverage report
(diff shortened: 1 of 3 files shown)


wRAR:
Looks like most of the coverage increase is because of `extra-deps` jobs which were ignored before for some reason, though I see at least one newly covered Windows-only line.

PR #6611, opened 2025-01-09
Modified docs to use default strings in feeds export

### LINKED ISSUE #6425: Exception in FeedExporter when Using Path Objects with Storage URI Parameters
### Description

According to the [documentation](https://docs.scrapy.org/en/latest/topics/feed-exports.html#feeds), the `FEEDS` dict accepts `Path` objects as keys:

> [...] dictionary in which every key is a feed URI (_or a [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) object_) [...]

However, when using a `Path` object with Storage URI parameters, the `FeedExporter` runs into the following exception:

```
[scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method FeedExporter.open_spider of <scrapy.extensions.feedexport.FeedExporter object at 0x00000240E9F21F00>>
  File "...\.venv\lib\site-packages\scrapy\utils\defer.py", line 348, in maybeDeferred_coro
    result = f(*args, **kw)
  File "...\.venv\lib\site-packages\pydispatch\robustapply.py", line 55, in robustApply
  File "...\.venv\lib\site-packages\scrapy\extensions\feedexport.py", line 467, in open_spider
    uri=uri % uri_params,
```

This is because the `__init__` method of the `FeedExporter` uses `as_uri()` on `Path` objects, which yields an URL-encoded URI, loosing the format characters.

### Steps to Reproduce

1. Set any key of the `FEEDS` dict to a `Path` object containing a %-formatted path:
```python
FEEDS = {
  pathlib.Path("./%(time)s.csv"): {
    "format": "csv",
    "store_empty": True,
  }
}
```
2. Run any spider `scrapy crawl <spider_name>`.

**Expected behavior:** No exception in logs and the feed file being created.

### Versions

```
Scrapy       : 2.11.2
lxml         : 5.2.2.0
libxml2      : 2.11.7
cssselect    : 1.2.0
parsel       : 1.9.1
w3lib        : 2.2.1
Twisted      : 24.3.0
Python       : 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)]
pyOpenSSL    : 24.1.0 (OpenSSL 3.2.2 4 Jun 2024)
cryptography : 42.0.8
Platform     : Windows-10-10.0.22631-SP0
```

nitintecg:
Fixes #6425 
Modified feeds-export docs to support only strings based path instead of pathlib objects.

### DIFF
--- a/docs/topics/feed-exports.rst
+++ b/docs/topics/feed-exports.rst
@@ -432,6 +432,7 @@ Default: ``{}``
 A dictionary in which every key is a feed URI (or a :class:`pathlib.Path`
 object) and each value is a nested dictionary containing configuration
 parameters for the specific feed.
+Currently, using placeholders in :class:`pathlib.Path` object as feed URI is not supported.
 
 This setting is required for enabling the feed export feature.
 


docs/topics/feed-exports.rst
```diff
@@ -429,8 +428,7 @@ FEEDS
 
 Default: ``{}``
 
-A dictionary in which every key is a feed URI (or a :class:`pathlib.Path`
-object) and each value is a nested dictionary containing configuration
+A dictionary in which every key is a feed URI and each value is a nested dictionary containing configuration
 parameters for the specific feed.
```
wRAR, line 432: This should be rewrapped.

wRAR (approved):

wRAR:
Yeah, that's another alternative. We'll need a test though, if one doesn't exist.

nitintecg:
@Gallaecio / @wRAR Should i need modify to docs and include that using Path has some limitations and its not supported ?

wRAR:
The suggestion was to mention that the placeholders aren't supported, not that using Path objects is not supported at all.

docs/topics/feed-exports.rst
```diff
@@ -458,7 +457,7 @@ For instance::
             'encoding': 'latin1',
             'indent': 8,
         },
-        pathlib.Path('items.csv.gz'): {
+        'items.csv.gz': {
```
wRAR, line 460: Why is this changed?

docs/topics/feed-exports.rst
```diff
@@ -162,7 +162,6 @@ The feeds are stored in the local filesystem.
 
 Note that for the local filesystem storage (only) you can omit the scheme if
 you specify an absolute path like ``/tmp/export.csv`` (Unix systems only).
-Alternatively you can also use a :class:`pathlib.Path` object.
```
wRAR, line 165: Why is this removed?

docs/topics/feed-exports.rst
```diff
@@ -429,9 +428,9 @@ FEEDS
 
 Default: ``{}``
 
-A dictionary in which every key is a feed URI (or a :class:`pathlib.Path`
-object) and each value is a nested dictionary containing configuration
-parameters for the specific feed.
+A dictionary in which every key is a feed URI and each value is a nested dictionary containing configuration
+parameters for the specific feed.
```
wRAR, line 432: Why is this removed?

docs/topics/feed-exports.rst
nitintecg: sorry. i forgot to revert the changes back.

PR #6612, opened 2025-01-10
Create class to allow send emails through Amazon SES

### LINKED ISSUE #3345: Allow Scrapy to send e-mails using Amazon SES

wRAR:
Fixes #3345, closes #3346

Continuation of #3346 by @rennerocha


wRAR:
Closing as it's unclear if this has demand, but we should reuse the refactoring for public mail API later.

PR #6613, opened 2025-01-12
Drop PyPy 3.9, add a pypy3-extra-deps CI job.

### LINKED ISSUE #6271: Add an extra-deps job for pypy
After https://github.com/scrapy/scrapy/pull/6269 pypy gets brotlicffi support. However, it currently runs in all pypy jobs in CI. It would be best to have a separate CI job for pypy+extra-deps, same as we have with have for other scenarios, so that we know that Scrapy works well in pypy without installing brotlicffi or any other extra.

wRAR:
Fixes #6271.

I'm somewhat on the fence with removing PyPy 3.9 though: on one hand it's not supported anymore, on the other hand if we want a `pypy3-pinned` job we need to maintain a separate slightly different list of dep versions as long as the minimum supported version of PyPy is newer than the CPython one.

Additionally, I wonder if we want a `pypy3-asyncio` job and a `pypy3-extra-deps-pinned` one.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -190,15 +190,13 @@ setenv =
 
 [testenv:pypy3]
 basepython = pypy3
-commands =
-    pytest {posargs:--durations=10 docs scrapy tests}
+;commands =
+;    pytest {posargs:--durations=10 docs scrapy tests}
 
 [testenv:pypy3-extra-deps]
 basepython = pypy3
 deps =
     {[testenv:extra-deps]deps}
-commands =
-    pytest {posargs:--durations=10 docs scrapy tests}
 
 [testenv:pypy3-pinned]
 basepython = pypy3.10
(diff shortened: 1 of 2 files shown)


wRAR:
(no coverage changes because we for some reason don't collect coverage info in pypy jobs)

wRAR:
OK, it looks like the run time of PyPy jobs with coverage enabled is pretty poor (~3x slower) so it may make sense to not enable it, as there aren't many code paths specific to it.

tox.ini
```diff
@@ -118,6 +118,7 @@ setenv =
 install_command =
     python -I -m pip install {opts} {packages}
 commands =
+    ; tests for docs fail with parsel < 1.8.0
```
Gallaecio, line 121: Isn’t this a bit out of place here?

wRAR: The context is #5902

Gallaecio: My bad, for some reason I thought this message was about freezing parsel for the tests, I did not realize it was about removing the docs from them. :facepalm:

Gallaecio (approved):

PR #6614, opened 2025-01-14
Extend BaseSettings with utils for add-ons

Gallaecio:


scrapy/settings/__init__.py
```diff
@@ -111,6 +112,31 @@ def __getitem__(self, opt_name: _SettingsKeyT) -> Any:
...
+
+        This change is applied regardless of the priority of the *name*
+        setting. The setting priority is not affected by this change either.
+        """
+        value: list[str] = self.getlist(name)
+        if item not in value:
+            self.set(name, [*value, item], self.getpriority(name) or 0)
+
+    def remove_from_list(self, name: _SettingsKeyT, item: Any) -> None:
+        """Remove *item* to the :class:`list` setting with the specified
```
wRAR, line 127:
```suggestion
        """Remove *item* from the :class:`list` setting with the specified
```

wRAR (approved):

scrapy/settings/__init__.py
```diff
@@ -299,6 +328,45 @@ def maxpriority(self) -> int:
...
+        no guarantee about which one it is.
+
+        This change is applied regardless of the priority of the *name*
+        setting. The setting priority is not affected by this change either.
+        """
+        component_list = self.getdict(name)
+        old_pos = None
+        for cls_or_path in tuple(component_list):
+            if isinstance(cls_or_path, str):
+                _cls = load_object(cls_or_path)
```
kmike, line 356: Why is this check needed? Isn't load_object handling non-strings already?

scrapy/settings/__init__.py
```diff
@@ -299,6 +325,41 @@ def maxpriority(self) -> int:
...
+            else:
+                _cls = cls_or_path
+            if _cls == old_cls:
```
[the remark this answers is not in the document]
kmike: The current approach (checking for exact classes) makes most sense to me. 

We can extend it to raise a warning on subclasses, but we haven't seen these cases in the wild yet, right?

scrapy/settings/__init__.py
```diff
@@ -344,6 +436,27 @@ def setdefault(
...
+
+        If *cls* is not already defined, it is set regardless of the priority
+        of the *name* setting. The setting priority is not affected by this
+        change either.
+        """
+        component_list = self.getdict(name)
+        for cls_or_path in tuple(component_list):
+            if not isinstance(cls_or_path, str):
+                continue
+            _cls = load_object(cls_or_path)
```
kmike, line 454: Is str check needed?

docs/topics/settings.rst
```diff
@@ -234,6 +234,55 @@ example, proper setting names for a fictional robots.txt extension would be
 ``ROBOTSTXT_ENABLED``, ``ROBOTSTXT_OBEY``, ``ROBOTSTXT_CACHEDIR``, etc.
 
 
+.. _component-lists:
+
+Component lists
+===============
+
+A **component list** is a :class:`dict` where keys are :ref:`components
```
kmike, line 242: As you probably expected, this is the piece I'm least certain about :) I found it confusing to call a priority dictionary "list", both in docs and in the programmatic API. Not sure exactly what's a better name though; it seems other names are going to be longer ("component priority dictionary"), but it could be fine.

scrapy/settings/__init__.py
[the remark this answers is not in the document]
Gallaecio: > we haven't seen these cases in the wild yet, right?

Not that I recall, no.

docs/topics/settings.rst
Gallaecio: Yes, this was the hardest part. I think I went with component list because of the existence of `build_component_list`, but what that does is take this as input and return an actual list, so if anything it seems to support the need for a different name.

scrapy/settings/__init__.py
```diff
@@ -299,6 +328,49 @@ def maxpriority(self) -> int:
...
+        of them, with no guarantee about which one it is.
+
+        This change is applied regardless of the priority of the *name*
+        setting. The setting priority is not affected by this change either.
+        """
+        component_priority_dictionary = self.getdict(name)
+        old_priority = None
+        for cls_or_path in tuple(component_priority_dictionary):
+            if load_object(cls_or_path) != old_cls:
+                continue
```
kmike, line 360: it'd be good to cover it with a test

kmike (approved):
Nice! The method names are a little long, but I'm totally comfortable with that.

Gallaecio:
I‘ll rename `dictionary` to `dict` in method names, in line with the corresponding class name. Shortening that feels OK to me, unlike component → comp or priority → prio.

PR #6618, opened 2025-01-18
Unknown encoding handling in HttpCompressionMiddleware, restore x-gzip support

### LINKED ISSUE #4697: Warn about br handling if brotlipy is not installed

wRAR:
Fixes #4697, closes #5480, closes #6145.

There are two unrelated things in this (or any other) fix for the original ticket and they should be discussed separately, even though the questions look similar :)

Handling of the request `Accept-Encoding` header when we can detect that it's "wrong":

1. Should we handle (e.g. warn about) all items in the header that we don't recognize or only ones that can be problematic (br, zstd)? 
2. If we see an encoding that we think we won't be able to decode should we log a warning, log an error, raise an exception? If log, do it once per spider or for every request?
3. If we don't raise an exception do we keep the "wrong" encoding in the header or remove it?
4. Should we (as requested in #4697) check that the user has set `Accept-Encoding` manually to the default value and log something about that?

Handing of the response `Content-Encoding` header when we cannot support some encodings in it:

1. Should we handle (e.g. warn about) all items in the header that we don't recognize or only ones that can be problematic (br, zstd)? 
2. If we see an encoding that we can't decode should we log a warning, log an error, raise an exception? If log, do it once per spider or for every request?
3. If we don't raise an exception should we keep the encoded body?


wRAR:
My opinions.

Requests:
1. Not sure if there are use cases for passing unknown encodings there (if there are, and we want to warn about all unknown ones, we should provide a way to silence that). So warning about all unknown ones looks fine to me.
2. A warning is fine. Also, either the server ignores it (e.g. if it doesn't support brotli) or there will be another message when handling the response.
3. I think we should keep the header and make the user fix the code/the environment.
4. I don't see benefit in this.

Responses:
1. Again, not sure if there are use cases for encoded responses that are decoded by a separate middleware or a callback. If there aren't, we can warn if we couldn't decode the response.
2. Depending on whether we want to support the aforementioned use cases, it should be a warning (not sure if it will be possible to silence it) or an exception.
3. Yes, clearing it doesn't make sense to me.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -78,9 +81,35 @@ def open_spider(self, spider: Spider) -> None:
         if hasattr(spider, "download_warnsize"):
             self._warn_size = spider.download_warnsize
 
+    @staticmethod
+    def _get_unsupported_suggestion(encoding: bytes) -> str:
+        if encoding == b"br":
+            return " You need to install brotli or brotlicffi to use it."
+        if encoding == b"zstd":
+            return " You need to install zstandard to use it."
+        return ""
```
Gallaecio, line 90: Seems unused.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -78,9 +81,35 @@ def open_spider(self, spider: Spider) -> None:
...
     def process_request(
         self, request: Request, spider: Spider
     ) -> Request | Response | None:
+
+        user_header = request.headers.get("Accept-Encoding")
+        if user_header:
+            for user_encoding in user_header.split(b","):
+                user_encoding = user_encoding.strip()
+                # Encoding specification can also include a weight,
+                # we won't check such encodings.
```
Gallaecio, line 101: Would this do the job?

```suggestion
                user_encoding = user_encoding.split(";")[0].strip()
```

scrapy/downloadermiddlewares/httpcompression.py
wRAR: Indeed.

TBH the code I pushed is "sample" one, as different design decisions need quite different code. That's also why there are no tests yet.

scrapy/downloadermiddlewares/httpcompression.py
wRAR: What if the weight is 0? :)

I just don't know if there is a need to complicate the code here.

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -78,9 +81,35 @@ def open_spider(self, spider: Spider) -> None:
...
+        if encoding == b"zstd":
+            return " You need to install zstandard to use it."
+        return ""
+
     def process_request(
         self, request: Request, spider: Spider
     ) -> Request | Response | None:
+
+        user_header = request.headers.get("Accept-Encoding")
+        if user_header:
```
Gallaecio, line 97:
```suggestion
        if user_header:
            user_header = user_header.lower()
```

While I found no resource to confirm that the value is meant to be interpreted as case-insensitive ([this](https://github.com/jquery/codeorigin.jquery.com/issues/43) was the closest), I suspect it should be.

Gallaecio (approved):

scrapy/downloadermiddlewares/httpcompression.py
Gallaecio: Oh, I did not realize that was a possibility! :+1:

Gallaecio:
+1 to your proposed solutions. I do think we should offer an option to silence the response warning, I wonder if a setting for that would make sense (maybe a `…_LOG_LEVEL` setting, in line with the new item drop one).

wRAR:
After a discussion with @kmike we got the following points:

0. We can merge changes to `process_request()` and `process_response()` separately if one is clear and another is blocked, as they are independent.
1. We tentatively decided to not warn about known but unsupported encodings in the request as when the header value is "incorrect" the user will get a warning about the response anyway. 
2. Unknown encodings in the request should be ignored: we decided to support cases where the user decodes the response with custom code; not all encodings are compressions, [in theory](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding); the server can harmlessly ignore the value; we don't validate other request headers as thoroughly.
3. When the middleware cannot decode the response, it should produce a warning, saying that this specific middleware cannot decode it. If it knows what module to install it should also say that. Adding a setting to disable this was deemed excessive, it should be possible to subclass the middleware though.
4. Module-level vars should be moved inside the middleware.

wRAR:
Oh, we also wanted to move `ACCEPTED_ENCODINGS` inside the middleware. I wonder how can we maintain the backwards compatibility in this case (actually, it's a question how much compatibility do we want, e.g. do we want to support modifying the module-level attribute after the middleware initialization).

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -143,9 +145,11 @@ def _handle_encoding(
             body = self._decode(body, encoding, max_size)
         return body, to_keep
 
+    @staticmethod
     def _split_encodings(
-        self, content_encoding: list[bytes]
+        content_encoding: list[bytes],
     ) -> tuple[list[bytes], list[bytes]]:
+        supported_encodings = {*ACCEPTED_ENCODINGS, b"x-gzip"}
```
Gallaecio, line 152: I wonder if we should do something to make it easier to subclass the middleware to extend support for additional encodings, and hopefully make the implementation slightly cleaner that the same time.

e.g. Define `HttpCompressionMiddleware.DECODERS` as:

```python
DECODERS = {
    "br": unbrotli,
    "deflate": inflate,
    "gzip": gunzip,
    "x-gzip": gunzip,
    "zstd": unzstd,
}

// …

body = self.DECODERS[encoding](body, max_size=max_size)

// …

if encoding not in self.DECODERS:
```

Note that I am also suggesting making decoding functions public, to enable easier mapping of new keywords, e.g. `"x-deflate": inflate`.

No strong opinion on this, though. It’s rather out of scope here. And I am not sure if making those functions public is a good idea.

Gallaecio (approved):

scrapy/downloadermiddlewares/httpcompression.py
```diff
@@ -155,19 +159,35 @@ def _split_encodings(
         to_decode: list[bytes] = []
         while to_keep:
             encoding = to_keep.pop()
-            if encoding not in ACCEPTED_ENCODINGS:
+            if encoding not in supported_encodings:
```
kmike, line 162: It seems it changes the behavior for x-gzip encoding; previously it wasn't removed from the headers, and probably wasn't decoded at all? Hm, does it mean that x-gzip handling code was not working before?

I think this change is for better, but we might want to mention it in the release notes.

scrapy/downloadermiddlewares/httpcompression.py
kmike: I recall suggesting a similar thing to @wRAR :) But it seems leaving it out of scope for this PR is better. It doesn't seem this PR adds any additional obstacles to implement it later.

scrapy/downloadermiddlewares/httpcompression.py
wRAR: `x-gzip` handling was silently broken in 2.12.0 when this code was added.

PR #6622, opened 2025-01-19
fix(docs): pillow domain is shut down permanently

Rotzbua:
### Change

Pillow website is shut down permanently. Target github repository before website is taken by evil 3rd party.

### Reference

See https://github.com/python-pillow/Pillow/issues/8585


wRAR (approved):

wRAR:
Nice, thanks. I ran linkcheck but a month earlier than the linked ticket happened.

Issue #6620, opened 2025-01-19
Chore(docs): migrate to sphinx 8


Rotzbua:
### Motivation

Currently sphinx v6 is used which is eol. Sphinx v7 & v8 have some internal changes which break the docs creation as discussed in #6177 .

### Suggested changes

- [x] Migrate to latest RTD template
- [x] Refactor config (created 17 years ago 🥳 )
- [x] Fix custom extension for sphinx v7/v8
- [x] Add basic typing to extension

wRAR:
@Rotzbua should this be closed now?

wRAR:
Thank you for the fixes!

PR #6624, opened 2025-01-19
chore(docs): bump sphinx to v8

Rotzbua:
### Changes

Refactor extension:

- Add types if possible
- Ignore types of docutils (better typed 0.22 not released yet)
- Remove and replace deprecated functions

### Reference

mentioned in #6620

### DIFF
--- a/docs/_ext/scrapydocs.py
+++ b/docs/_ext/scrapydocs.py
@@ -1,63 +1,67 @@
 # pylint: disable=import-error
+from collections.abc import Sequence
 from operator import itemgetter
+from typing import Any, TypedDict
 
 from docutils import nodes
+from docutils.nodes import Element, General, Node, document
 from docutils.parsers.rst import Directive
-from docutils.parsers.rst.roles import set_classes
+from sphinx.application import Sphinx
 from sphinx.util.nodes import make_refnode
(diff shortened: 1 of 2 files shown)


docs/_ext/scrapydocs.py
```diff
@@ -1,54 +1,62 @@
...
 
     if not hasattr(env, "scrapy_all_settings"):
-        env.scrapy_all_settings = []
+        env.scrapy_all_settings = []  # type: ignore
 
-    for node in doctree.traverse(is_setting_index):
-        targetnode = get_setting_target(node)
-        assert isinstance(targetnode, nodes.target), "Next node is not a target"
+    for node in doctree.findall(is_setting_index):
+        # This seems not needed anymore.
```
wRAR, line 50: Interesting, the page is indeed generated correctly AFAICS.

`def get_setting_target()` is also no longer needed then.

wRAR (approved):

docs/_ext/scrapydocs.py
Gallaecio: May we remove the commented out code before merging?

wRAR (approved):

PR #6625, opened 2025-01-22
Fix test_s3_export StubAssertionError by allowing ChecksumAlgorithm

turboslapper:
Description
This update modifies the S3FeedStorage test to include "ChecksumAlgorithm" unconditionally in expected_params.

Benefits
Ensures compatibility with newer boto3 versions (>=1.36.0) where ChecksumAlgorithm is included by default.
Simplifies test logic by avoiding version-based conditions.

Fixes
Resolves #6619.

Issue #6619, opened 2025-01-18
test_s3_export fails with boto3 >= 1.36.0


wRAR:
https://github.com/scrapy/scrapy/actions/runs/12845409692/job/35819443101?pr=6618

```
botocore.exceptions.StubAssertionError: Error getting response stub for operation PutObject: Expected parameters:
{'Body': <ANY>, 'Bucket': 'mybucket', 'Key': <ANY>},
but received:
{'Body': <s3transfer.utils.ReadFileChunk object at 0x7f8fc1dee750>,
 'Bucket': 'mybucket',
 'ChecksumAlgorithm': 'CRC32',
 'Key': 'export.csv/3.json'}
```

wRAR:
It doesn't look like `Stubber` supports optional keys in `expected_params` so it looks like we need to check the version (of boto3 or botocore?).

turboslapper:
Hi, I'd like to work on this issue and propose a solution. I'll be looking into handling the ChecksumAlgorithm parameter conditionally based on the boto3 version to ensure the test works with both older and newer versions. I'll submit a PR soon. Let me know if there are any specific requirements or constraints you'd like me to keep in mind. Thanks!

turboslapper:
### Additional FeedExportTest failures on Python 3.13 with Twisted 24.11.0

Hi all! I’m also encountering failures in `FeedExportTest` on my local environment, beyond the `ChecksumAlgorithm` stub mismatch. Even after allowing `"ChecksumAlgorithm": ANY`, the following tests in `tests/test_feedexport.py` fail or hang:

- `test_export_feed_export_fields`
- `test_export_indentation`
- `test_export_items`

They eventually raise these issues using this command:

twisted.internet.defer.TimeoutError: still running at 120.0 secs twisted.internet.error.AlreadyCancelled: Tried to cancel an already-cancelled event.


Here’s an excerpt from my local logs (Python 3.13.1, Twisted 24.11.0, boto3==1.36.2):
'NoneType' object is not iterable Reactor was unclean. twisted.internet.error.AlreadyCancelled: Tried to cancel an already-cancelled event. ...


I’m testing in a Tox environment (`TOXENV=extra-deps`) with dependencies pinned by `tests/upper-constraints.txt`. These errors don’t seem directly related to the `ChecksumAlgorithm` stub issue, but rather appear to be timeouts or reactor shutdown issues. 

### Explanation of Change

[...]

PR #6626, opened 2025-01-22
fix: test_s3_export fails with boto3 >= 1.36.0

Laerte:
fix: #6619

Closes https://github.com/scrapy/scrapy/pull/6625.


Laerte:
Failed test don't seems related to this change.

wRAR (approved):
Thanks!

Issue #4975, opened 2021-02-04
SpiderLoader - holding all project's spider modules in memory during scraping run.


GeorgeA92:
### Description

On `scrapy crawl spidername`
Original purpose of spiderloader is to search and load requred spider by it's `name`.
Hovewer current implementation of [`SpiderLoader`](https://github.com/scrapy/scrapy/blob/2.4.1/scrapy/spiderloader.py) literally import all project spider modules and hold it in memory up to end of process (even for cases when we need to load single spider for `crawl` command).

Probably it is already mentioned in https://github.com/scrapy/scrapy/issues/1805
### Steps to Reproduce

1. Creat Scrapy Project with at least 1 spider (valid enough to make 1 request and display `memusage/startup` stats)
2. execute following script to generate required amount of spiders

<details><summary>mass gen spider code</summary>

```python
import sys
import optparse

from scrapy.cmdline import _get_commands_dict, _pop_command_name,\
    _run_print_help, _run_command
from scrapy.crawler import CrawlerProcess
from scrapy.utils.project import inside_project, get_project_settings


def execute(argv=None, settings=get_project_settings()):
    if argv is None:
        argv = sys.argv

    inproject = inside_project()
    cmds = _get_commands_dict(settings, inproject)
    cmdname = _pop_command_name(argv)
    parser = optparse.OptionParser(formatter=optparse.TitledHelpFormatter(),
                                   conflict_handler='resolve')
    cmd = cmds[cmdname]
    parser.usage = f"scrapy {cmdname} {cmd.syntax()}"
    parser.description = cmd.long_desc()
    settings.setdict(cmd.default_settings, priority='command')
    cmd.settings = settings
    cmd.add_options(parser)
    opts, args = parser.parse_args(args=argv[1:])
    _run_print_help(parser, cmd.process_options, args, opts)

    cmd.crawler_process = CrawlerProcess(settings)
    _run_print_help(parser, _run_command, cmd, args, opts)
    # sys.exit(cmd.exitcode) <- this line prevent multiple calls of scrapy commands in script

for i in range(3200):
    name = 'spider' + str(i)
    domain = f'www{str(i)}.quotes.toscrape.com' * 512  #result should be ~26Kb long
    execute(['scrapy', 'genspider', name, domain, '-t', 'crawl'])

```

</details>

3. comparison of `memusage/startup` for different amount of spiders clearly indicates that scrapy allocates around ~(1.5x to 2x) of total spider code files memory size.

[...]

PR #6629, opened 2025-01-23
spiderloader (load spider from given location instgead of loading all project spiders) [WIP] 

GeorgeA92:
aimed to resolve https://github.com/scrapy/scrapy/issues/4975
based on suggestion from https://github.com/scrapy/scrapy/issues/4975#issuecomment-2600219359

If `SPIDER_LOADER_LOAD_FROM_LOCATION` setting set to `True`  (from project settings):
1. spiderloader - load single spider by its location like `scrapy crawl project.spiders.quotes.QuotesSpider` instead of loading all spider modules from project
2. `list` method of spiderloader (called on `scrapy list` command` - return list of spider locations like `project.spiders.quotes.QuotesSpider`

### DIFF
--- a/scrapy/spiderloader.py
+++ b/scrapy/spiderloader.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 from zope.interface import implementer
 
 from scrapy.interfaces import ISpiderLoader
-from scrapy.utils.misc import walk_modules
+from scrapy.utils.misc import load_object, walk_modules
 from scrapy.utils.spider import iter_spider_classes
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


wRAR:
I would prefer if the setting didn't call that "location", I think the usual name is "import path" but maybe we want something shorter ("class"?) .

Gallaecio:
The implementation looks good.

I agree that we might want to rethink the setting name. Although I would even consider changing its type. I wonder if a string enum like `SPIDER_ID_TYPE` (not sure if we need a `SPIDER_LOADER_` prefix, in line with `SPIDER_MODULES`) with values `"name"` (default) and `"cls"` would be more user-friendly, and open the possibility of adding new values in the future (e.g. `"all"` to support both name and import class in the future, or some other way of identifying spiders).

GeorgeA92:
Spiderloader tests added to cover `SPIDER_LOADER_LOAD_FROM_PATH:True` case.

`Spiderloader.find_by_request` disabled from if `load_from_path` active.
This functionality require to load all spider modules and check it's internals - that I am aimed to prevent

scrapy/spiderloader.py
```diff
@@ -94,6 +105,8 @@ def find_by_request(self, request: Request) -> list[str]:
         """
         Return the list of spider names that can handle the given request.
         """
+        if self.load_from_path:
+            raise NotImplementedError("Not supported")
```
wRAR, line 109: This function, via `scrapy.utils.spider.spidercls_for_request()`, is used by `scrapy fetch`, `scrapy parse` and `scrapy shell` to find which spider to use. Can you please check what happens when the new setting is enabled and how to make the user experience better, if needed? Thanks.

Issue #6649, opened 2025-02-02
Refactor tests/test_downloadermiddleware_httpcache.py


wRAR:
`HttpCacheMiddleware` has two tunables, the storage and the policy, both with currently two options. It makes sense to me that any given test code should either run for all 4 combos or for both options of one tunable while setting the other tunable to a default value (if it doesn't mater for that code), but the actual tests are more complicated, maybe unintentionally. It would be nice to untangle the class hierarchy.

wRAR:
@fbenevolo you don't need to ask to work on something.

RobertPecz:
Hello, could you explain this issue more detailed so I can jump in?

What two options have policy and storage?

What 4 combos? 

> test_dont_cache, test_storage, test_storage_never_expire, test_storage_no_content_type_header

Are those the 4 combos?

Gallaecio:
[The docs](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache) may provide some helpful context here.

Issue #6645, opened 2025-02-01
Investigating and optimizing the slowest test run time


wRAR:
I continue investigating the test run time for fun and possible optimizations and I've isolated the following subset of test files that is responsible for 70-75% of the total run time (of `tox -e py`, so without extra deps) on my test machines:

* `tests/test_command_parse.py` 
* `tests/test_command_shell.py`
*  `tests/test_commands.py`
*  `tests/test_crawler.py`
*  `tests/test_crawl.py`
*  `tests/test_downloader_handlers_http2.py`
*  `tests/test_downloader_handlers.py`
*  `tests/test_downloadermiddleware_httpcache.py`
*  `tests/test_engine_stop_download_bytes.py`
*  `tests/test_engine_stop_download_headers.py`
*  `tests/test_feedexport.py`

(`tox -e py -- tests/test_command_parse.py tests/test_command_shell.py tests/test_commands.py tests/test_crawler.py tests/test_crawl.py tests/test_downloader_handlers_http2.py tests/test_downloader_handlers.py tests/test_downloadermiddleware_httpcache.py tests/test_engine_stop_download_bytes.py tests/test_engine_stop_download_headers.py tests/test_feedexport.py`)

I've tried to classify the reasons and/or find specific issues in these files.

* `tests/test_command_parse.py`: many `scrapy` runs, unclear if it's possible to optimize the `scrapy` startup time.
* `tests/test_command_shell.py`: same.
*  `tests/test_commands.py`: same.
*  `tests/test_crawler.py`: many `tests/CrawlerProcess/*.py` runs, unclear if it's possible to optimize their startup/run time.
*  `tests/test_crawl.py`: many short tests that run a spider (in-process). Also, two tests take 5+ seconds because of `scrapy.core.engine.Slot.heartbeat`.
*  `tests/test_downloader_handlers_http2.py`: many short tests that run a spider (in-process). Needs an investigation wrt test duplication with `tests/test_downloader_handlers.py` and also why do some of them aren't skipped without `h2` installed.
*  `tests/test_downloader_handlers.py`: many short tests that run a spider (in-process). Many of them also wait for a 0.5s timeout.
*  `tests/test_downloadermiddleware_httpcache.py`: 6 sets of `time.sleep(2)` and `time.sleep(0.5)` in the tests that deal with HTTP cache expiration, should be done in some other way.
[...]

PR #6650, opened 2025-02-02
Remove a duplicate test.

wRAR:
Related to #6645 and #6649. 

This also makes some methods that previously ran with `DbmCacheStorage` to run with `FilesystemCacheStorage` (but if that breaks anything it's a bug in them). Not sure which methods, though, maybe just `tests.test_downloadermiddleware_httpcache._BaseTest.test_dont_cache()`.

### DIFF
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -14,7 +14,7 @@ from scrapy.utils.test import get_crawler
 
 
 class _BaseTest(unittest.TestCase):
-    storage_class = "scrapy.extensions.httpcache.DbmCacheStorage"
+    storage_class = "scrapy.extensions.httpcache.FilesystemCacheStorage"
     policy_class = "scrapy.extensions.httpcache.RFC2616Policy"
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


tests/test_downloadermiddleware_httpcache.py
```diff
@@ -14,7 +14,7 @@
 
 
 class _BaseTest(unittest.TestCase):
-    storage_class = "scrapy.extensions.httpcache.DbmCacheStorage"
+    storage_class = "scrapy.extensions.httpcache.FilesystemCacheStorage"
```
kmike, line 17: I wonder if it'd be better to move these constants to DefaultStorageTest, and leave some non-working values here. Currently _BaseTest seems to be a mix of helper utilities, default values, and 1 test method. It may also require moving test_dont_cache somewhere else as well.

I've noticed it because of removal of FilesystemStorageTest - it was not immediately clear where FilesystemCacheStorage is tested  - you go to the direct parent (DefaultStorageTest), and there is not storage_class set.

I haven't really checked the code though :)

wRAR: Yeah, I created #6649 after looking at this and trying to fix it, this is just a small simple fix for a part of the problem. E.g. simply moving these values from `_BaseTest` to `DefaultStorageTest` will break `DummyPolicyTest` that subclasses it directly. And `test_dont_cache` seems to be useful to run for all 4 combinations of options.

kmike (approved):

PR #6651, opened 2025-02-03
Deprecate AjaxCrawlMiddleware and stop calling escape_ajax() by default

### LINKED ISSUE #6523: Deprecate `AjaxCrawlMiddleware`
The docs it links to no longer talk about stuff it expects (`meta name=fragment`, redirects to `...?_escaped_fragment_=`), there is not a lot of info on the Internet about this at all, and I assume websites no longer do this, so I don't think this is useful anymore.

The mechanism (at least according to our docs) was allowing the Google bot to get a plain text copy of a dynamic website, but [the current docs](https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#how-googlebot-processes-javascript) say it now just uses a headless Chromium to render those.

While the middleware is in the default list, `AJAXCRAWL_ENABLED` is off by default.

wRAR:
Fixes #6523

### DIFF
--- a/docs/topics/downloader-middleware.rst
+++ b/docs/topics/downloader-middleware.rst
@@ -1249,8 +1249,7 @@ AJAXCRAWL_ENABLED
 
 Default: ``False``
 
-Whether the AjaxCrawlMiddleware will be enabled. You may want to
-enable it for :ref:`broad crawls <topics-broad-crawls>`.
+Whether the AjaxCrawlMiddleware will be enabled.
 
 HttpProxyMiddleware settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


docs/topics/downloader-middleware.rst
```diff
@@ -1249,8 +1249,7 @@ AJAXCRAWL_ENABLED
 
 Default: ``False``
 
-Whether the AjaxCrawlMiddleware will be enabled. You may want to
-enable it for :ref:`broad crawls <topics-broad-crawls>`.
+Whether the AjaxCrawlMiddleware will be enabled.
```
kmike, line 1252: It probably makes sense to add a note below "AjaxCrawlMiddleware Settings", explaining that this middleware is deprecated.

kmike (approved):

PR #6653, opened 2025-02-04
Support dark mode in scrapy's documentation

protokoul:
This PR attempts to support dark mode for scrapy's documentation by adding [sphinx-rtd-dark-mode](https://pypi.org/project/sphinx-rtd-dark-mode/) extension to sphinx

By default, this extension places the toggle button at the bottom right corner and enables dark mode. It allows to configure whether dark mode should be enabled by default. I wanted to place the toggle button at top right corner so I added custom style in custom.css. This extension adds text to the footer.

It's not an essential feature as there are browser extensions that can enable dark mode for sites, for example [dark reader browser extension](https://darkreader.org/) but I thought it will be nice to have.

I am also mentioning https://github.com/readthedocs/sphinx_rtd_theme/issues/224#issuecomment-722655021 which used dark reader browser extension to export the requried CSS and add it to the config, but this feature is not accessible in the latest update https://github.com/darkreader/darkreader/issues/13#issuecomment-2126082174 so I decided to use sphinx-rtd-dark-mode instead

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -12,8 +12,8 @@
 /*override some styles in
 sphinx-rtd-dark-mode/static/dark_mode_css/general.css*/
 .theme-switcher {
-    right: 1.2em !important;
-    bottom: 29em !important;
+    right: 0.4em !important;
+    top: 0.6em !important;
     -webkit-box-shadow: 0px 3px 14px 4px rgba(0, 0, 0, 0.30) !important;
     box-shadow: 0px 3px 14px 4px rgba(0, 0, 0, 0.30) !important;
     height: 2em !important;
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):
API object names are kept in light mode, and the plugin has had no work in 2 years. However, this does not break light mode, and dark mode looks good to me overall. So +1 from me.

protokoul:
I think I should be able to change the style for API objects. Let me see. I also noticed that the toggle button placement can be improved for mobile screens. Will push some changes for that too

wRAR (approved):
Nice!

protokoul:
I have raised a PR in the plugin's repo to improve the styles used in API object names. If it gets approved, then I can remove them from custom.css and update the plugin version.

@Gallaecio @wRAR Shall I implement this for any other documentations as well?

Gallaecio:
> Shall I implement this for any other documentations as well?

I am tempted to say “That would be awesome!”, but I wonder if it would not be better to get it fixed upstream first, so that applying the change in other repos becomes trivial. But I also suspect the upstream is dead.

protokoul:
Sure. If the plugin gets updated then I can push new changes here and likewise work on implementing this for other documentations if needed

Issue #6636, opened 2025-01-28
Reactor info logged twice


wRAR:
#6374 added a second `log_reactor_info()` call to `scrapy.crawler.Crawler._apply_settings()` so in some cases (including just `scrapy shell -s TWISTED_REACTOR="twisted.internet.asyncioreactor.AsyncioSelectorReactor"`) the info is logged twice:

```
2025-01-28 23:27:11 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.asyncioreactor.AsyncioSelectorReactor
2025-01-28 23:27:11 [scrapy.utils.log] DEBUG: Using asyncio event loop: asyncio.unix_events._UnixSelectorEventLoop
2025-01-28 23:27:11 [scrapy.utils.log] DEBUG: Using reactor: twisted.internet.asyncioreactor.AsyncioSelectorReactor
2025-01-28 23:27:11 [scrapy.utils.log] DEBUG: Using asyncio event loop: asyncio.unix_events._UnixSelectorEventLoop
```

Simply removing it is wrong though, as it breaks the test.

PR #6657, opened 2025-02-05
fix: Reactor info logged twice

Laerte:
fix: #6636

Tested manually with:
```
python tests/CrawlerRunner/change_reactor.py
```
and
```
scrapy shell -s TWISTED_REACTOR="twisted.internet.asyncioreactor.AsyncioSelectorReactor"
```

### DIFF
--- a/scrapy/crawler.py
+++ b/scrapy/crawler.py
@@ -123,7 +123,7 @@ class Crawler:
             else:
                 from twisted.internet import reactor  # noqa: F401
 
-            if logged_reactor_info is False:
+            if not logged_reactor_info:
                 log_reactor_info()
                 logged_reactor_info = True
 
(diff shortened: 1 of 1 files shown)


scrapy/crawler.py
```diff
@@ -113,20 +113,27 @@ def _apply_settings(self) -> None:
...
+            if not logged_reactor_info:
+                log_reactor_info()
+                logged_reactor_info = True
+
         if reactor_class:
             verify_installed_reactor(reactor_class)
             if is_asyncio_reactor_installed() and event_loop:
                 verify_installed_asyncio_event_loop(event_loop)
 
-            log_reactor_info()
```
Gallaecio, line 129: What if we only log it here, but remove 1 level of indentation?

wRAR: There are cases where neither of these two blocks is executed I think.

Gallaecio: Since this only logs information about the reactor being used, I wonder if it makes sense to only log in when those conditions are meant. Although maybe it does, maybe logging it always could trigger multiple logging because this function is called multiple times, but only once with the conditions for the logging.

To be safe, we could check the conditions, both, at this point:

```
if a and b:
    log()
```

Gallaecio (approved):

PR #6682, opened 2025-02-18
Changed the default to AsyncioSelectorReactor

### LINKED ISSUE #6659: Enable `AsyncioSelectorReactor` by default
Thoughts?

It's enabled in the template settings.py since Scrapy 2.7.0 (2022-10-17). It should be easy to disable it if needed, but if somebody knows reasons when it's actually needed to disable it - please tell.

tinytimcodes:
Fixes #6659

changed code to enable Asyncio reactor as default

### DIFF
--- a/scrapy/cmdline.py
+++ b/scrapy/cmdline.py
@@ -4,6 +4,7 @@ import argparse
 import cProfile
 import inspect
 import os
+os.environ.setdefault("TWISTED_REACTOR", "twisted.internet.asyncioreactor.AsyncioSelectorReactor")
 import sys
 from importlib.metadata import entry_points
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


scrapy/cmdline.py
```diff
@@ -4,6 +4,7 @@
 import cProfile
 import inspect
 import os
+os.environ.setdefault("TWISTED_REACTOR", "twisted.internet.asyncioreactor.AsyncioSelectorReactor")
```
wRAR, line 7: Why is this needed?

wRAR:
You also need to remove the TWISTED_REACTOR line from `settings.py.tmpl`, and likely change some documentation and/or tests that currently assume that `AsyncioSelectorReactor` is not the default value.

PR #6683, opened 2025-02-18
Fix #6321: Added allow_tags and deny_tags support for LxmlParserLinkExtractor

### LINKED ISSUE #6321: Option to include all tags and attrs in LinkExtractor with specified exclusions

imaanedhi:
- Implemented allow_tags and deny_tags to filter extracted links based on HTML tags.
- Created an ExtendedLink subclass to store tag information.
- Updated _extract_links() to include tag data.
- Added test cases to verify correct filtering of allowed and denied tags.
- Ensured backward compatibility with existing behavior.


wRAR:
Most of these changes are unrelated and/or make no sense, can you please explain?

PR #6684, opened 2025-02-18
curl: add support for parsing -b,--cookie

mdzw:
When parsing `curl` command line options, Scrapy currently supports grabbing cookie values from `-H "Cookie: ..."` parameters. However, curl also supports specifying cookies via `-b` and `--cookie`. This PR adds support for that.

I updated an existing test case to use the `-b` method to specify one of the cookies.

Motivation: Previously, the `Copy as cURL` option in Google Chrome's Developer Tools would generate a curl command with cookies specified via `-H "Cookie: ..."`. A recent update seems to have changed that feature to use `-b`.


wRAR (approved):
Thanks!

Issue #6590, opened 2024-12-24
Path to removal of spider middleware iterable downgrading


wRAR:
We added support for async spider middlewares in 2.7 (In October 2022), and [mixing sync and async middlewares](https://docs.scrapy.org/en/latest/topics/coroutines.html#sync-async-spider-middleware) was intended to be temporary and eventually deprecated and removed, after (most?) 3rd-party middlewares are converted. The initial discussion of this was at https://github.com/scrapy/scrapy/pull/4978#discussion_r645987466 and then at https://github.com/scrapy/scrapy/pull/4978#issuecomment-944238855 and below. I felt from the beginning that there is no definite path to the expected destination, because things can and should work until we disable them, and also because for many 3rd-party middlewares the only reason to change them would be other 3rd-party middlewares gaining async support, but I now wonder how we can make at least some next steps to the destination.

We expect one of two things from 3rd-party middleware authors: either make the middleware universal, adding an `async def process_spider_output_async` method, or make the middleware async-only, making an `async def process_spider_output` method. If neither is done, and either an async callback or an async-only middleware (called before this one in the stack) is used, there will be a warning in the spider logs. It doesn't say anything about deprecation though, just a warning that an async iterable was downgraded when passed to this middleware. And if everything else is sync or universal, there will be no warning. 

There is also another future problem, if you make your middleware universal in the more far future you will need to change it again to make it async-only. I don't think this is clear to a random maintainer and I don't know if we should make it more clear, but I think this change should be mostly trivial anyway.

So maybe as the first steps that can be done right now we should:

- make the downgrading warning say that this behavior is deprecated
- make the downgrading warning say how can it be fixed, and link to the docs
- at the load step emit a warning if a spider middleware has only the sync method, saying this is deprecated (will be deprecated? do we emit such advance warnings?)?

Should we try to check which popular middlewares are still not updated? Should we at least do this for Zyte-written ones (public and/or internal?)

Anything else we can do?

wRAR:
Another thing I've forgot to mention is helper methods that we discussed in the past that we could add to Scrapy to make writing such middlewares easier (I'm not sure if the proposals are written somewhere already).

Gallaecio:
> Should we try to check which popular middlewares are still not updated? Should we at least do this for Zyte-written ones (public and/or internal?)

Indeed. I know of at least one that has not been updated yet.

[...]

PR #6693, opened 2025-02-25
Initial BaseSpiderMiddleware.

wRAR:
Related to #6590

What do we think about this? Is this the protocol we decided to implement?

### DIFF
--- a/scrapy/spidermiddlewares/base.py
+++ b/scrapy/spidermiddlewares/base.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from scrapy import Request, Spider
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterable, Iterable
+
+    from scrapy.http import Response
+
+
+class BaseSpiderMiddleware:
+    """Optional base class for spider middlewares.
+
+    This class provides helper methods for asynchronous ``process_spider_output``
+    methods. Middlewares that don't have a ``process_spider_output`` method don't need
+    to use it.
+    """
+
+    def process_spider_output(
+        self, response: Response, result: Iterable[Any], spider: Spider
+    ) -> Iterable[Any]:
+        for o in result:
+            if isinstance(o, Request):
+                o = self._process_request(o, response, spider)
+                if o is not None:
+                    yield o
+            else:
+                o = yield self._process_item(o, response, spider)
+                if o is not None:
+                    yield o
+
+    async def process_spider_output_async(
+        self, response: Response, result: AsyncIterable[Any], spider: Spider
+    ) -> AsyncIterable[Any]:
+        async for o in result:
+            if isinstance(o, Request):
+                o = self._process_request(o, response, spider)
+                if o is not None:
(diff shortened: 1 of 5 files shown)


scrapy/spidermiddlewares/base.py
```diff
@@ -0,0 +1,55 @@
...
+            if isinstance(o, Request):
+                o = self._process_request(o, response, spider)
+                if o is not None:
+                    yield o
+            else:
+                o = yield self._process_item(o, response, spider)
+                if o is not None:
+                    yield o
+
+    def _process_request(
```
wRAR, line 47: Should the overridable methods be underscored?

Gallaecio: I don’t think so, that makes it look like they are not meant to be overridden.

scrapy/spidermiddlewares/base.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    def process_spider_output(
+        self, response: Response, result: Iterable[Any], spider: Spider
+    ) -> Iterable[Any]:
+        for o in result:
+            if isinstance(o, Request):
+                o = self._process_request(o, response, spider)
+                if o is not None:
+                    yield o
```
Gallaecio, line 28: I wonder if we should use `yield from self.process_request(o, response)`, to allow yielding multiple items/requests based on a given request (or item below), and to keep `spider` off since it feels like an obsolete parameter not needed nowadays with the 1 crawler = 1 spider constraint.

kmike: `yield from` would require users to yield the items, instead of returning them, right?

Gallaecio: Yes.

wRAR:
If we require passing a crawler instance to BaseSpiderMiddleware subclasses it requires something like this: https://github.com/wRAR/scrapy/pull/3

So this changes the `__init__()` signature for `DepthMiddleware`, `OffsiteMiddleware` and `UrlLengthMiddleware` (as it adds a new required argument, I rewrote them instead of simply adding it). In practice it should only matter for `__init__()` in subclasses of those I think, assuming nobody instantiates middlewares explicitly.

scrapy/spidermiddlewares/urllength.py
```diff
@@ -7,72 +7,49 @@
...
+class UrlLengthMiddleware(BaseSpiderMiddleware):
+    crawler: Crawler
+
     def __init__(self, maxlength: int):
         self.maxlength: int = maxlength
 
-    @classmethod
-    def from_settings(cls, settings: BaseSettings) -> Self:
-        warnings.warn(
-            f"{cls.__name__}.from_settings() is deprecated, use from_crawler() instead.",
```
wRAR, line 37: This is removed early but I don't think it's a big deal (we should try to mention it in the release notes though).

Gallaecio: Maybe edit `news.rst` already as part of this PR, so we don’t forget.

Gallaecio (approved):

PR #6696, opened 2025-03-02
Move most of the test utils inside tests.

wRAR:
This moves `scrapy.utils.{testproc,testsite}` into the newly created `tests.utils` package, moves some functions from `scrapy.utils.test` that are only used in some tests into those test modules and just deprecates ones that aren't used. The old code is kept with deprecation warnings.

Motivation: the stuff moved is very specific and unlikely to be useful outside the Scrapy test suite, and even there it's useful only for specific tests, but having it in the Scrapy itself requires us to keep it and maintain compatibility in it. `testproc` and `testsite` are used in several modules so it's better to put them into an utils package than to put them inside some test module and import from there. Other functions are only used in one test module and I'm not sure if it's possible to have more modules that will use them, but it's fine to move them into `tests.utils` too, or do that when it will be needed. We may also want to move some other code into `tests.utils` but that can be done separately when/if we identify such code.

This keeps `get_crawler()`, `get_pythonpath()`, `get_testenv()`, `get_from_asyncio_queue()` and `get_web_client_agent_req()` in the supported API because they may theoretically be useful in external code (last two likely only in some simple async tests, but still).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -13,6 +13,7 @@ def _py_files(folder):
 
 collect_ignore = [
     # not a test, but looks like a test
+    "scrapy/utils/testproc.py",
     "scrapy/utils/testsite.py",
     "tests/ftpserver.py",
     "tests/mockserver.py",
--- a/scrapy/utils/test.py
+++ b/scrapy/utils/test.py
@@ -34,11 +34,23 @@ _T = TypeVar("_T")
 
 
 def assert_gcs_environ() -> None:
+    warnings.warn(
+        "The assert_gcs_environ() function is deprecated and will be removed in a future version of Scrapy."
+        " Check GCS_PROJECT_ID directly.",
+        category=ScrapyDeprecationWarning,
+        stacklevel=2,
+    )
     if "GCS_PROJECT_ID" not in os.environ:
         raise SkipTest("GCS_PROJECT_ID not found")
 
 
 def skip_if_no_boto() -> None:
+    warnings.warn(
+        "The skip_if_no_boto() function is deprecated and will be removed in a future version of Scrapy."
+        " Check scrapy.utils.boto.is_botocore_available() directly.",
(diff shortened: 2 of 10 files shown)


scrapy/utils/testproc.py
```diff
@@ -2,18 +2,27 @@
...
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
     from twisted.python.failure import Failure
 
 
+warnings.warn(
+    "The scrapy.utils.testproc module is deprecated.",
+    ScrapyDeprecationWarning,
+)
```
Gallaecio, line 23: [Found 1 use in scrapy-poet](https://github.com/scrapinghub/scrapy-poet/blob/703290b1f7854e2837a82a02bea01a537a033795/tests/test_middleware.py#L15). Not sure if it changes anything, through.

Gallaecio: For the record, found GitHub code search quite powerful:

> [/(?-i)\btestproc\b/ AND ( org:scrapy OR org:scrapy-plugins  OR org:zytedata OR org:scrapinghub ) AND NOT repo:scrapy/scrapy](https://github.com/search?q=%2F%28%3F-i%29%5Cbtestproc%5Cb%2F+AND+%28+org%3Ascrapy+OR+org%3Ascrapy-plugins++OR+org%3Azytedata+OR+org%3Ascrapinghub+%29+AND+NOT+repo%3Ascrapy%2Fscrapy+&type=code)

Gallaecio (approved):

scrapy/utils/testproc.py
wRAR: It indeed looks useful to that scrapy-poet test but most of the `testproc.py` code is generic, only 2 or so lines are specific to Scrapy. So I don't think Scrapy is a good place for shipping it.

PR #6698, opened 2025-03-06
spider name is optional 1157 [WIP]

GeorgeA92:
Fixes https://github.com/scrapy/scrapy/issues/1157

based on approach from https://github.com/scrapy/scrapy/pull/1649  (where I had issues to resolve merge conflict as that pull request was from 2015) so technically only spider class updated to return spider class name as default value of `name` attribute  (so it doesn't require to additionally set `name` for spider)

As mentioned on https://github.com/scrapy/scrapy/issues/1157#issuecomment-1816782204 other multiple components (and unknown number of 3rd party scrapy plugins/custom middlewares) directly call `spider.name` attribute.

From https://github.com/scrapy/scrapy/pull/1649#issuecomment-178697597

>The problem is that after this change base classes (including scrapy.Spider?) will appear in `scrapy list` command; this is confusing and not backwards compatible. We may add a special attribute which can be set explicitly to mark spider as a base class (e.g. show=False or base=True), but as by default there won't be such attribute this change will require users to add it to all base spiders. So it is backwards-incompatible again. This all is obvious in hindsight, but only your PR helped me to realize that.

Yes at this stage it indeed count all classes inherited from `scrapy.Spider` including spiders that current spiderloader do not load.

Because `scrapy.utols.spideriter_spider_classes` works differently it this will be applied:
https://github.com/scrapy/scrapy/blob/35212ec5b05a3af14c9f87a6193ab24e33d62f9f/scrapy/utils/spider.py#L51-L62

### DIFF
--- a/scrapy/spiders/__init__.py
+++ b/scrapy/spiders/__init__.py
@@ -28,6 +28,14 @@ if TYPE_CHECKING:
     from scrapy.utils.log import SpiderLoggerAdapter
 
 
+class classproperty:
+    def __init__(self, fget):
+        self.fget = fget
+
+    def __get__(self, cls, owner):
+        return self.fget(owner)
+
+
 class Spider(object_ref):
     """Base class for scrapy spiders. All spiders must inherit from this
     class.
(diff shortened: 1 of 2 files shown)


Gallaecio:
I think we need a setting like in https://github.com/scrapy/scrapy/pull/4327, make the default behavior remain backward-compatible, but allow the new behavior by changing that setting. And with the new behavior, we need to make sure that built-in base classes are still not listed by `scrapy list`, and that there is a clear path for users to mark such spiders (what I did in https://github.com/scrapy/scrapy/pull/4327 through a decorator, although a decorator is not necessarily the only approach for that).

Issue #6658, opened 2025-02-06
Switch tests to full pytest style


wRAR:
We currently use pytest to run tests but write tests in the `unittest` style, also using `twisted.trial`. This has some restrictions, especially regarding pytest fixtures (see also #6478 and #6637). It seems like a good idea to switch to just using pytest, with pytest-style asserts, fixtures etc., using `pytest-twisted` if needed. Hopefully this can be done gradually. We can also try this migration on any of the smaller repos with a similar way of running tests, such as w3lib.

We may also want to rewrite async tests from `inlineCallbacks` (oir even `addCallback`) to `async def` in the process (or separately, whatever is easier).

Random related links:

* https://docs.pytest.org/en/stable/how-to/unittest.html
* https://docs.pytest.org/en/stable/how-to/xunit_setup.html
* https://github.com/pytest-dev/pytest-twisted/issues/147
* https://github.com/dannysepler/pytestify
* https://github.com/pytest-dev/unittest2pytest

wRAR:
Now that all test classes that don't use twisted.trial Deferred handling are no longer subclasses of `unittest.TestCase`, it should be straightforward to use fixtures, both pytest-provided and local ones, and also to move some functions into the module top level where it isn't useful to have a class. We should also rewrite Deferred-based test functions, where applicable, from using the callback code style, `self_wait()` and returning Deferreds explicitly to either `@inlineCallbacks()` or even right to `@deferred_f_from_coro_f` to make it easier to switch from twisted.trial to some other solution that will run sync tests for us.

PR #6702, opened 2025-03-07
Converting tests to plain asserts, part 4.

wRAR:
Partial fix for #6658


wRAR:
This broke `tests/test_squeues.py::PickleFifoDiskQueueTest::test_non_pickable_object`, not sure why as the error is in `tearDown` while the PR only changes asserts, I'll try to find the reason for that.

Issue #2002, opened 2016-05-19
Are there plans to add Web ARChive support?


robsonpeixoto:
From Wikipedia, the free encyclopedia

> The Web ARChive (WARC) archive format specifies a method for combining multiple digital resources into an aggregate archive file together with related information. The WARC format is a revision of the Internet Archive's ARC File Format[5] that has traditionally been used to store "web crawls" as sequences of content blocks harvested from the World Wide Web. The WARC format generalizes the older format to better support the harvesting, access, and exchange needs of archiving organizations. Besides the primary content currently recorded, the revision accommodates related secondary content, such as assigned metadata, abbreviated duplicate detection events, and later-date transformations.[6]
> 
> WARC is now recognised by most national library systems as the standard to follow for web archival.[7]
> 

https://en.wikipedia.org/wiki/Web_ARChive

kmike:
No plans I'm aware of. 

Does it have to be in Scrapy core? Currently in Scrapy it is user who defines what information to extract and store. It looks like WARC support is a matter of creating a function which converts scrapy Response to a WARC record; it can be done as a separate package. I'm not very familiar with WARC though; are there fields which are hard to get with Scrapy, maybe some low-level information? Alternatively, one can send all Scrapy network traffic through proxy, and this proxy may record WARC (see e.g. https://github.com/ikreymer/warcprox).

robsonpeixoto:
Thanks @kmike. I'll study more about WARC and scrapy Response.
Thanks the the tip about warcprox.

PR #6706, opened 2025-03-09
New HttpCacheMiddleware backend: WARC format

Kevin-Prichard:
Addresses #2002 (closed); relates to #4923 ("Store more data in the HTTP cache").

## Summary
- Adds 'WARCCacheStorage': an HttpCacheMiddleware backend that reads/writes each request & response pair from/to a WARC file (optionally gzipped)
- Files can be stored at a path name like the URL (without the scheme), allowing you to paste a URL in a shell to view the .warc file, for example https://scrapeme.live/shop/Ivysaur/ can be viewed in a *nix bash shell like-
  - `less scrapeme.live/shop/Ivysaur.warc` (or `zless scrapeme.live/shop/Ivysaur.warc.gz` when HTTPCACHE_GZIP is enabled), followed by the Tab key to complete the filename
- Operates transparently with Zyte API (untested on Zyte Cloud)

#### Motivation
When crawling a new-to-you website, several issues come up-
- recrawling while testing site-specific crawler code sometimes gets you blocked (solved by other cache backends, but continue reading...)
- we need to review markup for writing and fine-tuning our spider's code
- current cache backends aren't so easy to retrieve & review requests & pages (DBM & FilesystemCacheStorage)
- FilesystemCacheStorage allocates multiple inodes per URL, as request, response and metadata are spread over multiple files, decreasing inode count on the given device, limiting availability for future WARC files

#### Caveats & TODO
- Does not implement expiration (yes; would love to see a Scrapy extensio for space-and-time-constrained LRU expiry at some point)
- this pull covers a number of URL/file path formats, but likely missed some edge cases, so hit me up!
- Open to your feedback
- When Content-Encoding indicates compression, upstream Scrapy code expects a compressed body, but this patch compresses the entire file to disk, so that it may be decompressed / viewed at once for convenience.  So, to satisfy Scrapy and preserve the headers intact, we re-compress the Response body before handing it back to Scrapy.  This is obviously inefficient.  An alternative would be to modify or drop Content-Encoding to reflect an uncompressed body.

## How to use
In your Scrapy project's settings.py, add or change the following settings-
```
HTTPCACHE_ENABLED = True
HTTPCACHE_STORAGE = 'scrapy.extensions.httpcache.WARCCacheStorage'
HTTPCACHE_WARC_PATH_STYLE = 'FOLDERS_AND_FILE'
```
Optionally*, for filesystem space savings, you can add-
```
HTTPCACHE_GZIP = True
```
*In later *nix OS releases, GNU's [less](https://github.com/gwsw/less) command can uncompress gzipped files automatically (if overridden in shell .rc files by [lesspipe](https://github.com/wofr06/lesspipe).)

## Tests
- A test class was built out and offers nearly 100% coverage of the new backend: tests.test_downloadermiddleware_httpcache.WARCCacheStorageTest
- Manually tested on a fresh Python [venv](https://docs.python.org/3/library/venv.html) via `pip install git+https://github.com/Kevin-Prichard/scrapy.git@httpcache-warc`
- Untested on Windows and MacOS platforms (not a hater! just don't have those available to me at the moment, so if you do please try, let me know what breaks, and offer patches to my patch if you like).  With Windows, the main thing that won't work out of the box is pasting a URL into PS or CMD and getting it to tab complete.

## Future
- Possible future compatibility with [WASAPI](https://github.com/WASAPI-Community/data-transfer-apis/tree/master/ait-specification)
- Could be extended to store page requisites in addition to page markup, such as images and scripts, to support replay without the browser (or other HTTP client) requesting resources which may no longer be online
- Do better cache path searching, to check for pages cached at all four pathname styles, in case the spider developer changed strategy along the way and doesn't want to re-request an already-visited and cached URL


wRAR:
Should this really be shipped with Scrapy or can it be a separate package?

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "tldextract",
     "lxml>=4.6.0",
     "defusedxml>=0.7.1",
+    "warcio>=1.7.5",
```
wRAR, line 26: Even if it needs to be shipped inside Scrapy it shouldn't add new dependencies to Scrapy as it's optional.

scrapy/utils/httpobj.py
```diff
@@ -22,3 +22,73 @@ def urlparse_cached(request_or_response: Request | Response) -> ParseResult:
     if request_or_response not in _urlparse_cache:
         _urlparse_cache[request_or_response] = urlparse(request_or_response.url)
     return _urlparse_cache[request_or_response]
+
+
+# For scrapy.extensions.httpcache.WARCCacheStorage.
+# Because scrapy either doesn't expose or capture the HTTP status message,
+# so we reconstruct it when building WARC Response records for caching.
+HTTP_STATUS_MSGS = {
```
wRAR, line 30: `twisted.web.http.RESPONSES`

pyproject.toml
Kevin-Prichard: 1. Figured I would try it first as a PR, to see whether consensus / acceptability developed.  If it's not accepted in the main project, is there a registry of Scrapy plugins somewhere, to improve discoverability?
2. If it is accepted, what's the preferred way to add dependencies for new features?

Kevin-Prichard:
> Should this really be shipped with Scrapy or can it be a separate package?

No, it doesn't.  My first time contributing here, so with that feedback, I'll withdraw this PR, publish it on its own and add it to the [Awesome Scrapy repo](https://github.com/AccordBox/awesome-scrapy).

Is there a curated list of prioritized issues that I can contribute on?  I'm going to be using Scrapy for the foreseeable future and want to help out, but don't want to spin my wheels.

Cheers.

scrapy/utils/httpobj.py
Kevin-Prichard: Replaced, for completeness.. tho this PR will become its own project shortly.

pyproject.toml
Kevin-Prichard: I am unsure how that would be done, but as a third-party extension I guess the question is moot.

wRAR:
Hi! Sorry for not getting back to you right away.  To our knowledge there are two lists of 3rd-party Scrapy plugins, https://github.com/croqaz/awesome-scrapy and https://github.com/AccordBox/awesome-scrapy and both weren't updated for some years so we don't often suggest those. I think there were suggestions in the past to make a more official list but we don't yet have plans for that. So you can contact owners of either, or both, lists.

> Is there a curated list of prioritized issues that I can contribute on?

We rarely have prioritized issues, those usually get added to the [release milestones](https://github.com/scrapy/scrapy/milestones), though more often than not we either work on such issues ourselves or they are too large to tackle right away. OTOH this means that contributors can work on almost any issue from the full list.

Gallaecio:
Closing as per https://github.com/scrapy/scrapy/pull/6706#issuecomment-2715665194 (Great work, though!)

PR #5113, opened 2021-04-22
Make Twisted[http2] installation optional

Gallaecio:
Given issues like https://github.com/conda-forge/scrapy-feedstock/pull/56 and https://github.com/emmett-framework/emmett/issues/315, make h2 installation optional to allow installing Scrapy in environments where installing h2 < 4 is not possible.

PR #6713, opened 2025-03-09
Enable AsyncioSelectorReactor by default.

### LINKED ISSUE #6659: Enable `AsyncioSelectorReactor` by default

wRAR:
Fixes #6659, closes #6682

There is an unresolved problem with running `--reactor=default` tests: when an in-process crawler is created, it checks the installed reactor against the setting (because the setting is now non-empty) and raises an exception. Probably the crawler should get a `TWISTED_REACTOR = None` setting in such tests.


wRAR:
I see two options:
1. Check the `reactor_pytest` fixture in `get_crawler()`, but I think this is not possible, so it needs to be checked in the test and the result passed into `get_crawler()`. I think this requires changing all the code that calls `get_crawler()` which doesn't look like a good idea.
2. Check `is_asyncio_reactor_installed()` inside `get_crawler()`, set empty `TWISTED_REACTOR` if it isn't. I think this should work and it shouldn't break other packages that use `get_crawler()`. It could be a somewhat bad idea if we were using `get_crawler()` to test the crawler behavior regarding reactors and settings, but I think we don't as we use separate processes for that.

docs/topics/asyncio.rst
```diff
@@ -144,3 +148,14 @@ Using custom asyncio loops
...
 :setting:`ASYNCIO_EVENT_LOOP` setting to the import path of the desired event
 loop class to use it instead of the default asyncio event loop.
+
+
+.. _disable-asyncio:
+
+Switching to a non-asyncio reactor
+==================================
+
+If for some reason your code doesn't work with the asyncio reactor you can use
```
Gallaecio, line 158:
```suggestion
If for some reason your code doesn't work with the asyncio reactor you, can use
```

Gallaecio (approved):

wRAR:
`tests/test_http2_client_protocol.py::TestHttps2ClientProtocol::test_POST_large_json_x10` fails consistently on windows-pinned, with "still running at 120.0 secs" (maybe it hangs)

Gallaecio:
If a higher minimum version of Twisted solves that, I would support taking that route.

wRAR:
It looks like that test on 3.9 and 3.10 with the asyncio reactor on Windows on GitHub CI indeed takes more than 2 minutes, but not much more, something like 3 minutes. And switching out any of these requirements (so using 3.11+, the default reactor, Ubuntu or my local Windows VM) makes it pass significantly faster (about 1 minute on my VM with 3.9-3.12, faster on 3.13 but no difference between 3.9 and 3.12; 24 seconds on GitHub CI with 3.9 and the default reactor). In general it's usually one of the longest running tests on any platform when HTTP/2 is enabled, but not as significantly. As I have no desire to debug this specifically on old Python on Windows, I have 3 suggested solutions:

1. Increase the timeout for this test from 120s to ~240s - not sure if it's worth to spend this extra time running it.
2. Disable this test on some superset of affected systems (probably on Windows+asyncio like we do with a couple of other tests that are known to fail there).
3. Remove `Twisted[http2]` from `pinned`, as it's currently not a dep of Scrapy and that removal should have happened when we removed that dep in #5113, thus not running this test on - the change makes sense as it makes the test envs more consistent (I don't know if it brings any other benefit but the current set of Windows CI jobs is not ideal, e.g. the only one that runs HTTP/2 tests is 3.9/pinned, so if we do this, and even if we don't, we should try adding a Windows extra-deps job).

Gallaecio:
Any of the 3 works for me.

wRAR:
I went with the last one, adding `extra-deps` and `extra-deps-pinned` for Windows and removing HTTP/2 from `pinned`. I bumped Pillow to the first version with 3.9 wheels.

PR #6720, opened 2025-03-11
Do not run the CI on branches, only on the main branch and on PRs

Gallaecio:
Split off from https://github.com/scrapy/scrapy/pull/6715.


kmike (approved):

kmike:
Do we want to keep running CI on long-living release branches?

Issue #4463, opened 2020-03-30
start_requests bypassing rules while working with CrawlSpider


rhlr:
### Description

I have been trying to use Scrapy's CrawlSpider to crawl listings from a website. The problem is the data comes from `XMLHttpRequest`. So, I have been using `[Puppeteer As A Servivce](url)` to render javascript properly in scrapy. Now, since even the first request comes from a `XMLHttpRequest`, I wanted to use `start_requests` method instead of the regular `start_urls` way. But when I use `start_requests` method, the `CrawlSpider` doesn't follow the specified rules. And when I just use the `start_urls` it doesn't render javascript, so no links are found to be followed. Does anyone know a way around this ? Here's my code:

### Steps to Reproduce

    import scrapy
    from scrapypuppeteer import PuppeteerRequest, PuppeteerHtmlResponse, 
    PuppeteerJsonResponse
    from scrapy.spiders import CrawlSpider,Rule
    from scrapy.selector import Selector
    from scrapy.linkextractors import LinkExtractor
    from scrapy.http import Request, HtmlResponse

    class MySpider(CrawlSpider):
        name = "myspider"
        allowed_domains = []
        urls = []            

        def __init__(self, startUrl = None, config_dict = None, path = None,*a, **kw):
            self.config_dict = config_dict
            super(MySpider, self).__init__(*a, **kw)
            MySpider.rules = (Rule(LinkExtractor(allow=set([".*/dresses/.*"])), callback='parse_listing', follow=True,process_request="use_puppeteer"),)
            super(MySpider, self)._compile_rules()
            self.start_urls = ["https://www.myntra.com/red-dresses"]

        def _requests_to_follow(self, response):
           if not isinstance(
                response,
                (HtmlResponse, PuppeteerHtmlResponse, PuppeteerJsonResponse)):
            return
            seen = set()
            for n, rule in enumerate(self._rules):
                links = [lnk for lnk in rule.link_extractor.extract_links(response)
                     if lnk not in seen]
                if links and rule.process_links:
                    links = rule.process_links(links)
                for link in links:
                    seen.add(link)
                    r = self._build_request(n, link)
                    yield rule.process_request(r)

        def use_puppeteer(self, request):
            return PuppeteerRequest(request.url, callback=self.parse_listing)

        def start_requests(self):
            yield PuppeteerRequest(self.start_urls[0], callback=self.parse_listing, dont_filter=True)`

**Expected behavior:** 
`MySpider.rules` should have been followed

[...]

PR #6726, opened 2025-03-12
add parse_with_rules method

UTBucky:
Added suggested changes for [PR #5856 ](https://github.com/scrapy/scrapy/pull/5856)

This includes updates to warning to indicate where the problem is coming from in cases where CrawlSpider is subclassed from plugin, third-party, etc. [Suggestion #1](https://github.com/scrapy/scrapy/pull/5856#discussion_r1162657134)

Backward compatibility if _parse_response is overridden. [Suggestion #2](https://github.com/scrapy/scrapy/pull/5856#discussion_r1162661141)

Adjusting the warning to appear on the called line. [Suggestion 4](https://github.com/scrapy/scrapy/pull/5856#discussion_r1162668014)

There was a suggestion to distinguish warning from overriding and warning from us, although I wasn't sure how best to implement this [Suggestion 3](https://github.com/scrapy/scrapy/pull/5856#discussion_r1162666901)

Also includes addition of type hints where missing. 

Fixes [#4463](https://github.com/scrapy/scrapy/issues/4463)


wRAR:
I haven't checked the code yet but as we recently switched most of our test suite to plain asserts you should fix the test methods you added to use them.

PR #6727, opened 2025-03-12
Updated docs for reactor settings for CrawlerProcess and CrawlerRunner

### LINKED ISSUE #6627: Update the docs about global settings
We should clarify https://docs.scrapy.org/en/master/topics/settings.html#special-settings based on the analysis in https://github.com/scrapy/scrapy/pull/6568#discussion_r1925145591 and ideally also make sure all affected settings themselves also say, or link, their specifics in their docs.

ndhir98:
Resolves #6627

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -335,6 +335,23 @@ These settings are:
 installing the reactor. The rest of the settings are applied when starting
 the reactor.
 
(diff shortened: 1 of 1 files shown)


wRAR:
I don't think this does what we want, as it adds a new block of text after the existing text, partially repeating and partially contradicting it. The existing text should be fixed/improved instead.

wRAR:

PR #6728, opened 2025-03-12
Resolved #3463

### LINKED ISSUE #3463: raise CloseSpider("Error Message") is not saving error message

shinhonggwan:
Hello, I would be thankful if you could please review my fixes.

I made some changes to the def _spider_idle() and the def _spider_idle(self) in order to save the error message correctly.


wRAR:
How did you test this change?

Issue #4842, opened 2020-10-12
[Documentation add] how files works?


FrancescaFre:
I need to understand how files works during the crawling and how they are used in the crawler. Like "requests.seen", or "queue dir" or "activity.json" and so on. I had some problem with the crawler and i would try to understand how this files works. 

Should be useful to know the interaction between this kind of files with the crawler, maybe in the documentation.

michael-lazar:
Hi!

I'm still learning about scrapy and here are my personal notes for the JOBDIR. Please don't assume any of this is correct 😄 

## Crawl Directory

(using the DownloaderAwarePriorityQueue)

- {crawlname}
  - requests.queue
    - active.json
    - {hostname}-{hash}
      - {priority}
        - info.json
        - q{00000}
  - requests.seen
  - spider.state

### requests.queue

This directory contains requests that have been queued by the crawler but haven't been downloaded yet.

#### active.json

This file contains a dump of the scheduler's on-disk priority queue metadata. By default, this is provided by the [scrapy.pqueues.ScrapyPriorityQueue](https://docs.scrapy.org/en/latest/topics/settings.html?highlight=SCHEDULER_PRIORITY_QUEUE#scheduler-priority-queue) class.

Scrapy implements priority queues by keeping a list of functionally separate FIFO queues that each have a priority number assigned to them. The scheduler needs to keep track of these queue/priority mappings in order to load the queues from disk and pick up where it left off.

(Is this actually true? It seems to me like all of this information can be recovered from the directory structure itself since the priorities are used for the sub-directory names)

This file is only written to disk when the crawler is closed cleanly. If the crawler resumes but cannot access this file, it will not be able to load the requests.queue or recover any of the pending requests.

Example:

```
{"www.scrapy.org": [6, 7], "www.github.com": [7]}
```

#### {hostname}-{hash}

A sub-directory for a single slot in the crawler. The name is a filesystem-safe encoding of the hostname, along with the hostname's md5-hash to prevent rare collisions between hostnames.

(why not just use the md5-hash? It seems pointlessly complex to add the human readable hostname component.)

#### qXXXXXX

The file structure of the disk-backed queues are implemented by the [queuelib](https://github.com/scrapy/queuelib) library. Request objects that are pushed to the queue are serialized (using pickle by default) and packed into a binary file format that's chunked across multiple files. I won't get into the gritty details but the general format of the q000000, q000001, etc. files looks like this:

```
[size header][pickled request][size header][pickled request]...
```

[...]

Issue #411, opened 2013-10-03
Scrapy hangs if an exception raises in start_requests


kmike:
Symptoms are the same as in https://github.com/scrapy/scrapy/issues/83, but the reason is different, so I opened a new ticket. This issue starts to happen after this commit: https://github.com/alexcepoi/scrapy/commit/902208ca58aa99bacea57488642e0ea4129bd180. @alexcepoi any ideas?

PR #6729, opened 2025-03-12
Minimal asynchronous start requests

### LINKED ISSUE #456: Allow start_requests method running forever

### LINKED ISSUE #3477: Make requests via message queues
I'm trying to pass requests to the spider externally, via message queues, and keep it running forever.

I found some projects made by others but none of them work for the current version of scrapy, so I'm trying to fix the issues or just find a way to do it myself.

So far, I found that others got a reference to the scheduler from the `spider` in the middleware, like

```python
class MyMiddleware(object):
    # [...]
    def ensure_init(self, spider):
        self.spider = spider
        self.scheduler = spider.crawler.engine.slot.scheduler
    # [...]
        self.ensure_init(spider)
        return response
```

Then, in another, custom `"Scheduler"` class

```python
class MyScheduler(object):
    # [...]
    def open(self, spider):
        self.spider = spider
    # [...]
    def next_request(self):
        return self.spider.make_request(page)
```

And then, in the spider

```python
    # [...]
    def make_request(self, page):
        # This works, i.e. prints "Making request"
        logging.info("Making request")
        yield scrapy.Request(url=page, callback=self.parse)
    # [...]
    def parse(self, response):
        # This never gets printed
        logging.info("Got response")
```

This is the simplest example I could put together, but you get the idea, the code in reality is very messy, which makes it harder to fix.

The issue is, that in theory it should work, although I don't know when is that `next_request` method called, but it does because it calls the `make_request` method in the spider, the only problem is that it never gets to the `parse` or callback method in the spider, I don't know why.

I also tried connecting the spider to the message queue directly in the spider, which should work, but it doesn't, for example

```python
import pika

        rbmqrk = 'test'
        rmq = pika.BlockingConnection(
            pika.ConnectionParameters(host='localhost'))
        # Init channel
        rmqc = rmq.channel()

        super(TestSpider, self).__init__(*args, **kwargs)
        self.rmqc.queue_declare(queue=self.rbmqrk)

    def start_requests(self):
        self.rmqc.basic_consume(self.callback, self.rbmqrk)

    def callback(channel, method_frame, header_frame, body):
        # This gets printed! It works up to here.
        logging.info(body)
```

Up to there, everything is working fine, the body of the message received from the queue gets printed.

But, if we try to make or yield a request from the callback method in the spider, it won't work, for example

```python
       # [...] same as above...

    def start_requests(self):
        # Same as above
        self.rmqc.basic_consume(self.callback, self.rbmqrk)

    def callback(channel, method_frame, header_frame, body):
        # This DOESN'T get printed
        logging.info(body)
        yield scrapy.Request(url=body.decode(), callback=self.parse)
        # This wouldn't work either (already tried as well)
        # return scrapy.Request(url=body.decode(), callback=self.parse)

    def prase(self, response):
        # We never get here, this doesn't get printer either
        logging.info("Got response")
```

Unfortunately, yielding a generator or a request from the callback is not making a spider request.

As you can see, I've tried several things without luck, but all I need to do is to be able to make requests with the messages from the message queue, I'm not sure if there's a bug in scrapy or there's something I can fix in the code but I would love to have some input on this before I start digging deeper into the scrapy code myself.

### LINKED ISSUE #5627: Add async support for start_requests() method

Gallaecio:
Core functionality of https://github.com/scrapy/scrapy/pull/6715, without retry policies or any other extra work that may unnecessarily delay or block the merging of asynchronous seeding support.

**Main changes:**

-   Deprecate `Spider.start_requests()` in favor of a new, async `Spider.start()` that runs in parallel with request scheduling.

    Resolves #456, resolves #3477, closes #4467, resolves #5627.

    <details>

    <summary>FAQ</summary>

    -   Why deprecate `start_requests()` instead of allowing to define it as a coroutine?

        1.  It is a good chance to change a name that I think needs changing.

            I feel like `start_requests()` was never a good name for a method, because “start” here means “initial” rather than “initiate”, and method names should start with imperative verbs indicating what they “do”.

            Moreover, since item support was added, the `requests` part has also become problematic.

        2.  It allows implementing universal support in spiders, akin to what we did with spider middlewares and their `_async`-suffixed methods, but is slightly better since the async name here is final.

    -   How will the deprecation work?

        We add `start()` definitions to all built-in spiders while keeping their `start_requests()` definitions.

        The engine uses `start()`. Since its default implementation iterates `start_requests()`, old code should continue to work.

        > **Note:** Things are a bit different when spider middlewares are involved, see the details about the deprecation of `SpiderMiddleware.process_start_requests()` below.

        We inspect the MRO of the running spider class from the engine, and log a deprecation warning if the first class defining `start_requests()` does not also define `start()`. In other words, every spider class (even base classes) must be either new style (define `start()` only) or universal (define both) to silence the deprecation warning.

        In the future, once we decide that `start_requests()` has overstayed its welcome, we log a deprecation warning also for universal spider classes, and remove `start_requests()` from all built-in spiders, moving their code to the matching `start()` methods.

    -   Why `start()`?

        [Because](https://github.com/scrapy/scrapy/pull/6729#discussion_r2000763590).

    </details>

-   Deprecate `SpiderMiddleware.process_start_requests()` in favor of a new, async `SpiderMiddleware.process_start()` with a single `start: AsyncIterable[Any]` parameter.

    <details>

    <summary>FAQ</summary>

    -   Why deprecate `process_start_requests()` instead of allowing to define it as a coroutine?

        Same reasons as for `Spider.start()`.

    -   How will the deprecation work?

        The engine checks all enabled spider middlewares, and if any enabled spider middleware defines `process_start_requests()` and not `process_start()`:

        -   If the running spider is new style, or if there is any new-style-only spider middleware, close the spider with an error. In other words, no downgrade/upgrade support, to keep things simpler on our end.

        -   Else, log a deprecation warning, and use `start_requests()` instead of `start()`, but wrap the whole thing in an `AsyncIterable`  so that everything else in the code base remains the same.

        In the future, when removing `start_requests()` support, we should also start logging a deprecation warning for spider middlewares that define `process_start_requests()`.

    -   Why name the method `process_start()` and the parameter `start`?

        I feel these are nice and consistent with `Spider.start()`.

    -   Why remove the `spider` parameter?

        I feel like passing the `spider` parameter around is an anachronism from a time when a crawler could be used for multiple spiders, which [has not been the case for many years](https://github.com/scrapy/scrapy/commit/5eb429999e9e6a49eb9721d2dc0a47e0100345c3).

    </details>

-   Implement `crawler.signals.wait_for` and a `scheduler_empty` signal to allow reproducing the old behavior.

-   Document how to prioritize start requests over callback requests, and how to ensure start requests are sorted (with priorities, i.e. without changing the scheduler queue from LIFO to FIFO).

**Other changes:**

-   Switched to autodoc for parts of the API where I needed to edit the reference docs.

-   Switched from inlineCallbacks to deferred_f_from_coro_f in touched code, where possible.

-   Changed all `AsyncIterable` typing to `AsyncIterator` ([reason](https://github.com/scrapy/scrapy/pull/6729#discussion_r2016375104)).

**Post-merge to-do:**

-   [ ] Add a comment to https://github.com/scrapy/scrapy/issues/4842 to mention that there is now a `start` subfolder for start requests.


scrapy/spiders/__init__.py
```diff
@@ -72,7 +78,54 @@ def _set_crawler(self, crawler: Crawler) -> None:
         self.settings: BaseSettings = crawler.settings
         crawler.signals.connect(self.close, signals.spider_closed)
 
-    def start_requests(self) -> Iterable[Request]:
+    async def yield_seeds(self) -> AsyncIterator[Any]:
```
wRAR, line 81: Is it intentional that this is `AsyncIterator` and not just `AsyncIterable`? If we only use it with `async for` the latter is enough.

Gallaecio: I am unfamiliar with the differences between the 2, you are probably right.

wRAR: *Iterator is a superset of *Iterable, with the additions that are usually not needed. That's basically all I know without checking docs every time.

docs/intro/tutorial.rst
```diff
@@ -94,7 +94,7 @@ This is the code for our first Spider. Save it in a file named
     class QuotesSpider(scrapy.Spider):
         name = "quotes"
 
-        def start_requests(self):
+        async def yield_seeds(self):
```
kmike, line 97: I'm not sure I like the name:

1. "seed" assumes a request (something responses, requests and items "grow" from), but this method can now also return items
2. "yield" looks a bit strange in a method name in the public API, though that's subjective.

What do you think about calling it "start", i.e. dropping the "request" part from "start_requests"? It may not be perfect, but I like the connection to existing method name, and that the name is short.

```py
    async def start(self):
        for x in range(10):
            yield scrapy.Request(f"http://example.com/p/{x}", self.parse_page)
```

We might also support `def start`, not only `async def start`, because it's the more common use case, but I'm not sure.

kmike: `process_start_requests` method name becomes a bit of a challenge; `process_start`, `process_start_output`?

Gallaecio: `start` sounds good to me, and I think `process_start` works as well.

> We might also support `def start`, not only `async def start`, because it's the more common use case, but I'm not sure.

I’m not sure if it is worth supporting, since anything you can do with `def` you can also do with `async def`. But I’ll add it.

Gallaecio: > We might also support `def start`, not only `async def start`, because it's the more common use case, but I'm not sure.

Do you want to support the entire old syntax? i.e. returning an iterable, instead of defining the method as a generator? And if so, do we need to also support that when the method is `async def`? And if so, should we allow `async def` to return both an asynchronous and a synchronous iterable, or only one of those, and if so, which?

Gallaecio: Also, is it OK for `process_start` to have to be `async def`?

docs/contributing.rst
```diff
@@ -228,10 +228,9 @@ with a name of the branch you want to create locally).
 See also: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally#modifying-an-inactive-pull-request-locally.
 
 When writing GitHub pull requests, try to keep titles short but descriptive.
-E.g. For bug #411: "Scrapy hangs if an exception raises in start_requests"
-prefer "Fix hanging when exception occurs in start_requests (#411)"
-instead of "Fix for #411". Complete titles make it easy to skim through
-the issue tracker.
+E.g. For bug #411: "Scrapy hangs if an exception raises in start" prefer
+"Fix hanging when exception occurs in start (#411)" instead of "Fix for
```
kmike, line 232: I think we should either keep the issue name (https://github.com/scrapy/scrapy/issues/411), or use a different issue as an example :)

docs/intro/tutorial.rst
kmike: > Do you want to support the entire old syntax? 

I'm still not sure, if it's good to support it or not :)

Advantages of keeping non-async support:

* Slightly easier migration for start_requests users.
* The most common use case (99%?) doesn't need to be async.

Disadvantages:

* More code for us to support
* Weird typing support
* Potential confusion if user starts with a sync version, and latet tries to copy some example from docs which assume async

Another thing to consider: in https://github.com/scrapy/scrapy/pull/6730#discussion_r2000818143 I proposed to add some async primitives, instead of using seeding policies. It means that `start` method would need to be async in more cases - e.g. in order to reproduce the current Scrapy behavior; it could mean that the "most common" sync case drops from "99%" to "90%".

Given that we can add sync support later, it seems it'd be better to do it async-only for now :)
It also seems that to make decisions like this it'd be good to have #6730 changes in the same PR, as the approach we pick there may affect decisions here.

docs/news.rst
```diff
@@ -3,6 +3,98 @@
...
 =============
 
+.. _release-VERSION:
+
+Scrapy VERSION (unreleased)
+---------------------------
+
+Highlights:
+
+-   Replaced ``start_requests`` (sync) with :meth:`~scrapy.Spider.start`
```
wRAR, line 13:
```suggestion
-   Replaced ``start_requests()`` (sync) with :meth:`~scrapy.Spider.start`
```

tests/test_crawl.py
```diff
@@ -305,10 +305,10 @@ def test_referer_header(self):
         # basic asserts in case of weird communication errors
         assert "responses" in crawler.spider.meta
         assert "failures" not in crawler.spider.meta
-        # start requests doesn't set Referer header
+        # test_start doesn't set Referer header
```
wRAR, line 308: I think there are some search+replace errors in several files that produced "test_start" or e.g. "process_test_start"

docs/topics/coroutines.rst
```diff
@@ -6,8 +6,8 @@ Coroutines
 
 .. versionadded:: 2.0
 
-Scrapy has :ref:`partial support <coroutine-support>` for the
-:ref:`coroutine syntax <async>`.
+Scrapy has :ref:`partial support <coroutine-support>` for the :ref:`coroutine
```
kmike, line 9: Should drop "partial" from the description?

docs/topics/spider-middleware.rst
```diff
@@ -70,30 +70,29 @@ one or more of these methods:
...
 
-        :param start_requests: the start requests
-        :type start_requests: an iterable of :class:`~scrapy.Request`
+        You may yield the same type of objects as :meth:`~scrapy.Spider.start`.
 
-        :param spider: the spider to whom the start requests belong
-        :type spider: :class:`~scrapy.Spider` object
+        To write spider middlewares that work on Scrapy versions lower than
+        VERSION, define also a synchronous ``process_start_requests()`` method
+        that returns an iterable. For example:
```
kmike, line 90: It seems we should consider this in https://github.com/scrapy/scrapy/pull/6693

scrapy/commands/fetch.py
```diff
@@ -89,5 +89,12 @@ def run(self, args: list[str], opts: Namespace) -> None:
             spidercls = spider_loader.load(opts.spider)
         else:
             spidercls = spidercls_for_request(spider_loader, request, spidercls)
-        self.crawler_process.crawl(spidercls, start_requests=lambda: [request])
+
+        async def start(self):
+            yield self._request
+
+        spidercls._request = request  # type: ignore[assignment,attr-defined]
```
kmike, line 96:
```suggestion
        async def start(self):
            yield request
```

scrapy/crawler.py
```diff
@@ -151,8 +151,7 @@ def crawl(self, *args: Any, **kwargs: Any) -> Generator[Deferred[Any], Any, None
             self._apply_settings()
             self._update_root_log_handler()
             self.engine = self._create_engine()
-            start_requests = iter(self.spider.start_requests())
-            yield self.engine.open_spider(self.spider, start_requests)
+            yield self.engine.open_spider(self.spider)
```
kmike, line 154: There is a small difference in case start_requests is not a generator function (when it's a regular function which returns a list). But I don't think there should be any practical implications, it seems we can ignore this.

scrapy/extensions/telnet.py
```diff
@@ -104,7 +104,6 @@ def _get_telnet_vars(self) -> dict[str, Any]:
         telnet_vars: dict[str, Any] = {
             "engine": self.crawler.engine,
             "spider": self.crawler.engine.spider,
-            "slot": self.crawler.engine.slot,
```
kmike, line 107: Do you know if there was anything useful for debugging here? Something like inprogress? I'm not sure :) 

It seems we should drop this `Slot` / `_Slot`class, and just move everything to engine attributes. But it doesn't need to happen in this PR.

tests/test_crawl.py
```diff
@@ -361,27 +351,6 @@ def cb(response):
         assert s["engine.spider.name"] == crawler.spider.name
         assert s["len(engine.scraper.slot.active)"] == "1"
 
-    @defer.inlineCallbacks
-    def test_graceful_crawl_error_handling(self):
```
kmike, line 365: Is this test removed on purpose?

scrapy/core/spidermw.py
```diff
@@ -55,12 +56,56 @@ class SpiderMiddlewareManager(MiddlewareManager):
...
             self.methods["process_spider_input"].append(mw.process_spider_input)
-        if hasattr(mw, "process_start_requests"):
-            self.methods["process_start_requests"].appendleft(mw.process_start_requests)
+        if self._use_start_requests:
+            if hasattr(mw, "process_start_requests"):
+                self.methods["process_start_requests"].appendleft(
+                    mw.process_start_requests
+                )
+        elif hasattr(mw, "process_start"):
+            self.methods["process_start"].appendleft(mw.process_start)
```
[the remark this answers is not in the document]
kmike: Probably - I haven't understood the implementation yet :) 

For the universal middlewares, is it possible to prefer `process_start` over `process_start_requests` if there are non-universal middlewares in the chain?

tests/test_crawl.py
Gallaecio: It was because I first switched it to async def generator, and that scenario was covered in later changes (in #6715, `test_exception_before_yield`).

But it seems it would make sense to keep a test about this for non-generator `start_requests`.

scrapy/crawler.py
Gallaecio: https://github.com/scrapy/scrapy/pull/6729/commits/97a0fed020926af464631d31bdfb354a12aea8c3

scrapy/core/engine.py
```diff
@@ -114,22 +125,28 @@ def _get_scheduler_class(self, settings: BaseSettings) -> type[BaseScheduler]:
             )
         return scheduler_cls
 
-    @inlineCallbacks
-    def start(self) -> Generator[Deferred[Any], Any, None]:
+    @deferred_f_from_coro_f
+    async def start(self, _start_request_processing=True) -> None:
```
Gallaecio, line 129: About `_start_request_processing`:

Starting `spider.start()p  and `scheduler.next_request()` processing before `engine.start()` was problematic, because `engine.start()` sets `engine.running = True`, and before that, `engine.needs_backout()` returns `True`. This was one of the causes of the behavior differences in reactors, the asyncio one would happen to start request processing late enough for `engine.running` to be `True` already, but the default reactor would not.

So I concluded that request processing should be delayed until the engine starts.

So I moved the corresponding lines from `engine.open_spider()` to a separate, public method (should it be private?), and made `engine.start()` call it by default after setting `engine.running` to `True`. But I added a private parameter to allow disabling that, since the shell needs to delay that further.

I wonder if this change could pave the wave to make the shell open the spider before starting the engine, in line with `scrapy crawl`, but I don’t want to go down that rabbit hole now, not even as part of the follow-up PR.

scrapy/core/engine.py
```diff
@@ -362,31 +402,30 @@ def _on_complete(_: _T) -> _T:
...
+        nextcall = CallLaterOnce(self._start_scheduled_requests)
         scheduler = build_from_crawler(self.scheduler_cls, self.crawler)
-        start_requests = yield self.scraper.spidermw.process_start_requests(
-            start_requests, spider
+        self._slot = _Slot(close_if_idle, nextcall, scheduler)
+        self._start = await maybe_deferred_to_future(
+            self.scraper.spidermw.process_start(spider)
         )
-        self.slot = Slot(start_requests, close_if_idle, nextcall, scheduler)
-        self.spider = spider
```
Gallaecio, line 381: No technical reason for moving the definitions of spider and slot before the await, just felt cleaner.

tests/test_cmdline_crawl_with_pipeline/__init__.py
```diff
@@ -8,11 +8,16 @@ def _execute(self, spname):
...
+        returncode, stderr = self._execute("normal")
+        assert returncode == 0
 
     def test_exception_at_open_spider_in_pipeline(self):
-        assert self._execute("exception") == 1
+        returncode, stderr = self._execute("exception")
+        assert (
+            returncode == 0
+        )  # An unhandled exception in a pipeline should not stop the crawl
+        assert b'RuntimeError("exception")' in stderr
```
Gallaecio, line 23: This change in expectations may be worth discussing, and if it stays, I should probably add it to the list of backward incompatibilities.

Do we want *unhandled* exceptions in `SpiderMiddleware.open_spider` to prevent the spider from running normally? Should that apply to any handler of the [`spider_opened`](https://docs.scrapy.org/en/latest/topics/signals.html#std-signal-spider_opened) signal?

Gallaecio:
There is a remaining issue with a test that runs `tests/CrawlerProcess/asyncio_enabled_reactor_same_loop.py`, using uvloop, and hangs indefinitely.

scrapy/utils/reactor.py
```diff
@@ -114,8 +131,10 @@ def set_asyncio_event_loop(event_loop_path: str | None) -> AbstractEventLoop:
     """Sets and returns the event loop with specified import path."""
     if event_loop_path is not None:
         event_loop_class: type[AbstractEventLoop] = load_object(event_loop_path)
-        event_loop = event_loop_class()
-        asyncio.set_event_loop(event_loop)
+        event_loop = _get_asyncio_event_loop()
+        if not isinstance(event_loop, event_loop_class):
+            event_loop = event_loop_class()
+            asyncio.set_event_loop(event_loop)
```
Gallaecio, line 137: @wRAR I don’t fully understand why this fixed [the hang](https://github.com/scrapy/scrapy/pull/6729#issuecomment-2755790919) or, more importantly, why this PR triggered the hang in the first place.

scrapy/commands/bench.py
```diff
@@ -61,10 +59,10 @@ class _BenchSpider(scrapy.Spider):
     baseurl = "http://localhost:8998"
     link_extractor = LinkExtractor()
 
-    def start_requests(self) -> Iterable[Request]:
+    async def start(self) -> AsyncIterator[Any]:
```
Gallaecio, line 62: @wRAR I looked into the differences between [`AsyncIterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.AsyncIterable) and [`AsyncIterator`](https://docs.python.org/3/library/collections.abc.html#collections.abc.AsyncIterator), and ended up going with the later. Asynchronous generators are both, but in the end we need an iterator, and an iterable is simply an object that can provide an iterator, so asking for the iterator directly seems simpler.

scrapy/core/engine.py
```diff
@@ -88,20 +93,26 @@ def __init__(
...
+        downloader_cls: type[Downloader] = load_object(self.settings["DOWNLOADER"])
+        try:
+            self.scheduler_cls: type[BaseScheduler] = self._get_scheduler_class(
+                crawler.settings
+            )
+            self.downloader: Downloader = downloader_cls(crawler)
+            self.scraper: Scraper = Scraper(crawler)
+        except Exception:
+            self.close()
+            raise
```
Gallaecio, line 115: To exit cleanly in case of unhandled exceptions, since this is user-defined code.

scrapy/shell.py
```diff
@@ -102,25 +103,33 @@ def _schedule(self, request: Request, spider: Spider | None) -> defer.Deferred[A
...
+            assert self.crawler.engine is not None
+            self.crawler.engine.crawl(request)
+
+        d2 = self._open_spider(request, spider)
+        d2.addCallback(crawl_request)
+
         d = _request_deferred(request)
         d.addCallback(lambda x: (x, spider))
-        assert self.crawler.engine
-        self.crawler.engine.crawl(request)
```
Gallaecio, line 109: I needed this to ensure that the spider had been opened by the time `engine.crawl()` was called.

tests/test_spider.py
```diff
@@ -45,12 +44,6 @@ def test_base_spider(self):
         assert spider.name == "example.com"
         assert spider.start_urls == []  # pylint: disable=use-implicit-booleaness-not-comparison
 
-    def test_start_requests(self):
-        spider = self.spider_class("example.com")
-        start_requests = spider.start_requests()
-        assert inspect.isgenerator(start_requests)
-        assert not list(start_requests)
```
Gallaecio, line 52: Covered by `tests/test_spider_start.py`.

docs/news.rst
```diff
@@ -3,6 +3,113 @@
...
+
+-   Replaced ``start_requests()`` (sync) with :meth:`~scrapy.Spider.start`
+    (async) and changed how it is iterated.
+
+Backward-incompatible changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-   The iteration of start requests and items no longer stops once there are
+    requests in the scheduler, and instead runs continuously until all start
+    requests have been scheduled.
```
kmike, line 21: Is it a backwards incompatible change to start_requests, or is it a behavior of the new `async def start` method?

Gallaecio: At the moment, it is a backward-incompatible change of `start_requests`. And it is also a backward-incompatible change if you don’t use `start_requests` and instead use `start_urls`.

Shall I make `start_requests` behave lazily as before? Is it OK that `start_urls` has this backward-incompatible behavior by default?

scrapy/utils/python.py
```diff
@@ -19,11 +19,11 @@
 from scrapy.utils.asyncgen import as_async_generator
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Callable, Iterator
+    from collections.abc import Callable, Iterator
     from re import Pattern
 
-    # typing.Concatenate and typing.ParamSpec require Python 3.10
-    from typing_extensions import Concatenate, ParamSpec
+    # typing.Concatenate, typing.ParamSpec and typing.Self require Python 3.10
```
wRAR, line 25: `typing.Self` requires Python 3.11 :)

scrapy/core/engine.py
```diff
@@ -163,58 +179,84 @@ def pause(self) -> None:
...
             not self.running
-            or bool(self.slot.closing)
+            or bool(self._slot.closing)
             or self.downloader.needs_backout()
             or self.scraper.slot.needs_backout()
         )
 
-    def _next_request_from_scheduler(self) -> Deferred[None] | None:
-        assert self.slot is not None  # typing
+    def _start_scheduled_request(self) -> Deferred[None] | None:
```
wRAR, line 253: The returned deferred is never used, but it was the same with `_next_request_from_scheduler()`.

I wonder if it's saved somewhere deep inside the code or is just garbage collected at a random time.

wRAR: Though it may be a correct way to use deferreds?

wRAR (approved):
Great!

scrapy/core/engine.py
Gallaecio: Yes, I think you do not need to await deferreds. I think what’s an issue is when you define a deferred but never trigger or cancel it, but it is fine not to await deferreds.

docs/news.rst
kmike: Fair enough. I think that's fine to make a backwards-incompatible change here, if it's justified, and the behavior is more reasonable in most cases (which we should try to ensure). Having different behaviors could add to confusion.

docs/topics/spiders.rst
```diff
@@ -410,6 +364,96 @@ used by :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`::
...
+
+    .. code-block:: python
+
+            async def start(self):
+                priority = len(self.start_urls)
+                async for item_or_request in super().start():
+                    if isinstance(item_or_request, Request):
+                        item_or_request = item_or_request.replace(priority=priority)
+                    yield item_or_request
+                    priority -= 1
```
kmike, line 432: I'm worried about this approach, because Scrapy may not work very well with high cardinality of priorities. With disk queues, we create a folder on disk for each priority value - so e.g. if there are 20K items in `start_urls`, with this solution there are going to be 20K folders on disk. It looks like a high price to pay just to get the behavior which users may consider as expected (start_urls are processed roughly in order). It also means that we probably should't use this approach internally, if we want to ensure that start urls are processed roughly in order.

It seems for start requests LIFO doesn't make any sense, from the user point of view. It could be problematic, not just confusing - e.g. if there is a some "infinite" `async def start` which gets new requests externally, older requests may never get processed, which can be unexpected. User might see it as spider just "skipping" some requests if it can't keep up with the incoming stream.

For general crawling LIFO seems to make sense as a default. For example, you extract 10 product links from a product list page, and schedule them; they are now implicitly prioritized over downloading more product list pages. This seems to allow to keep queue sizes smaller (as less links need to be kept in memory at the same time); it also allows to get to some depth, and get to it sooner - otherwise some spiders may never reach some depth, spending all the resources discovering more links. FIFO won't be the end of the world as a default for crawling, but LIFO seems a bit better.

So, it seems we might want to mix FIFO and LIFO defaults in the same spider - use FIFO for start requests, and LIFO (or whatever user configures) for the rest. If there is a start request and a regular request of the same priority, it seems reasonable to pick start request first.

This doesn't look that easy to implement - it sounds like allowing to mix FIFO and LIFO queues in a scheduler. For example, have both FIFO and user-configured (LIFO default) queues for each priority, and put start request to FIFO part. But I haven't put much thought into this solution; there could be others, and there could be issues with this solution. 

I don't really like the solutions I'm proposing :) Any other ideas, how to make it less evil?

Gallaecio: I don’t think a separate start-request queue is a bad idea. And I don’t expect it to be that hard either.

We could:

1. Add a spider middleware, enabled by default, that adds `"is_start_request": True` to `Request.meta`.
2. Implement 2 new settings, `SCHEDULER_START_MEMORY_QUEUE` and `SCHEDULER_START_DISK_QUEUE`, `None` by default. If non-`None`, the scheduler stores such requests separately and, given 2 identical priorities, outputs start requests first.

So, things are backward-compatible, and we replace the current recommendation about forcing request order with using those settings.

For disk queues, the start requests could be in a `start` subfolder. That should be easy to implement, and allow users to enable `SCHEDULER_START_DISK_QUEUE` after pausing a job without issues.

kmike: Sounds fine overall.

> None by default. If non-None, the scheduler stores such requests separately and, given 2 identical priorities, outputs start requests first.

What is the default behavior, when these settings are None?

Gallaecio: The same as before the settings existed, all requests are stored in the same queues.

kmike: Yes, I think FIFO for start requests should be a default. I'm not sure why users would need to change it, unless they want to enable disk queues for start requests.

Gallaecio:
The lazy approach seems to consistently fail now. I need to figure out why.

Gallaecio:
Triggering CI…

Gallaecio:
Yes. And while in that commit is rare, it seems to trigger often enough in https://github.com/scrapy/scrapy/pull/6752 already.

I am not entirely sure what the right fix is, though.

On one hand, I suspect this could be easily worked around at the test level by using `data:` URLs.

But the point is to allow lazy loading in real projects. Something seems off if in the time since you get a response to the first request until you get response to the second request, the request yielded by the callback of the first response is not scheduled. I hope I can figure it out soon.

Gallaecio:
I don’t know if it is bad luck or what, but I wrote a short script to run until failure.

<details>

<summary>untilfail.sh</summary>

```bash
#!/usr/bin/env bash
count=$(( 1 ))
while "$@"; do
    echo "Succeeded on attempt $count"
    count=$((count + 1))
done
echo "Failed on attempt $count"
```

</details>

And when I run it with https://github.com/scrapy/scrapy/commit/5d16dd7a5f19335612b5f739392ae031c6e0087a, it failed on its 2nd attempt.

Then I added some logging to the code, to try and debug the issue:

<details>

<summary>git diff</summary>

```diff
diff --git a/scrapy/core/engine.py b/scrapy/core/engine.py
index 087fce234..40b639246 100644
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -185,6 +185,7 @@ class ExecutionEngine:
         If a request, it is scheduled. If an item, it is sent to item
         pipelines.
         """
+        logger.error("_process_start_next() starts")
         try:
             item_or_request = await self._start.__anext__()
         except StopAsyncIteration:
@@ -198,17 +199,20 @@ class ExecutionEngine:
             )
         else:
             if not self.spider:
+                logger.error("_process_start_next() ends")
                 return  # spider already closed
             if isinstance(item_or_request, Request):
                 self.crawl(item_or_request)
             else:
                 self.scraper.start_itemproc(item_or_request, response=None)
                 self._slot.nextcall.schedule()
+        logger.error("_process_start_next() ends")
 
     @deferred_f_from_coro_f
     async def _start_request_processing(self) -> None:
         """Starts consuming Spider.start() output and sending scheduled
         requests."""
+        logger.error("_start_request_processing() starts")
         # Starts the processing of scheduled requests, as well as a periodic
         # call to that processing method for scenarios where the scheduler
         # reports having pending requests but returns none.
@@ -222,7 +226,10 @@ class ExecutionEngine:
                 # Give room for the outcome of self._process_start_next() to be
                 # processed before continuing with the next iteration.
                 self._slot.nextcall.schedule()
+                logger.error("_start_request_processing() awaits nextcall")
                 await self._slot.nextcall.wait()
+                logger.error("_start_request_processing() awaited nextcall")
+        logger.error("_start_request_processing() ends")
 
     def _start_scheduled_requests(self) -> None:
         if self._slot is None or self._slot.closing is not None or self.paused:
diff --git a/tests/test_engine_loop.py b/tests/test_engine_loop.py
index 0c478f4e3..226acb61f 100644
--- a/tests/test_engine_loop.py
+++ b/tests/test_engine_loop.py
@@ -12,6 +12,9 @@ from scrapy.utils.test import get_crawler
 from .mockserver import MockServer
 from .test_scheduler import MemoryScheduler
 
+from logging import getLogger
+logger = getLogger(__name__)
+
 
 async def sleep(seconds: float = 0.001) -> None:
     from twisted.internet import reactor
@@ -160,8 +163,12 @@ class RequestSendOrderTestCase(TestCase):
             start = start_fn
 
             def parse(self, response):
+                logger.error(f"parse() starts ({response.request})")
                 while self.cb_requests:
-                    yield self.cb_requests.popleft()
+                    request = self.cb_requests.popleft()
+                    logger.error(f"parse() yields {request}")
+                    yield request
+                logger.error(f"parse() ends ({response.request})")
 
         actual_nums = []
 
@@ -234,8 +241,11 @@ class RequestSendOrderTestCase(TestCase):
         async def start(spider):
             for num in start_nums:
                 if spider.crawler.engine.needs_backout():
+                    logger.error(f"start() awaits scheduler_empty")
                     await spider.crawler.signals.wait_for(signals.scheduler_empty)
+                    logger.error(f"start() awaited scheduler_empty")
                 request = self._request(num, response_seconds, download_slots)
+                logger.error(f"start() yields {request}")
                 yield request
 
         await maybe_deferred_to_future(
```

</details>

And I force-stopped it at attempt 154.

I am not sure what else to do. I wonder if we should consider merging despite this known issue and open an issue to look into it. I feel further discouraged to look into it by the fact that with the [maximal PR](https://github.com/scrapy/scrapy/pull/6715) (c3a903ec8ffb743a3c1b347f23d1de7ca6e15f5e) it may not be reproducible (I got 50 successes in a row), so some change there may be indirectly fixing it already.

Gallaecio:
Nevermind. I had a look at the [maximal PR](https://github.com/scrapy/scrapy/pull/6715), to see if I had changed the `test_lazy` test itself in any way, and I noticed this:

```
response_seconds = self.seconds * 2**1  # increase if flaky
```

While I did not like the idea of switching URLs to `data:` to fix the test, because that would make the test less like a real world example, *increasing* response time seems fair since it makes the test *more* like a real scenario. And that small time response increase is enough on my PC to make https://github.com/scrapy/scrapy/pull/6774 work as well.

wRAR:
What is the actual problem that the test failure suggests? was the code expected to guarantee the order of those requests?

scrapy/core/spidermw.py
```diff
@@ -323,10 +388,91 @@ def process_spider_exception(
         dfd2.addErrback(process_spider_exception)
         return dfd2
 
-    def process_start_requests(
-        self, start_requests: Iterable[Request], spider: Spider
-    ) -> Deferred[Iterable[Request]]:
-        return self._process_chain("process_start_requests", start_requests, spider)
+    @deferred_f_from_coro_f
```
wRAR, line 391: I think we don't need the Deferred wrapper here, as this is a new function.

dangra:
Go! ❤️

PR #1443, opened 2015-08-20
Add a shortcut method `idled()` for `spider_idle` signal.

nyov:
Fixes #740

@dangra,
I require some clarification here: Is a `Crawler`<>`Spider` relationship always 1:1 nowadays? Or does a Crawler object ever handle multiple Spiders, in parallel?

`crawler.signals.connect(self.idle, signals.spider_idle)` inside the `Spider` class connects the Spider/s to the SignalManager for the Crawler instance.
So we shouldn't get any signals from any other Crawler instance, correct?
Now in the past, there has been code checking in a spider instance like this:`def idle(self): if spider != self: return`.

If I am correct, a Crawler will always only have one Spider, so `spider` will always only be `self` here now, right? So this can be a staticmethod.

PR #6730, opened 2025-03-13
Seeding policies

### LINKED ISSUE #740: Shortcut method for spider_idle signal
Implement a shortcut method handling for `spider_idle` signal similar to how it is done for `spider_closed`

A possible design of the api is an optional spider method named `spider.idled()`  called when spider_idled signal is emitted _for the spider_, the method can return an iterable of requests  to schedule.

Internally, the implementation will listen for spider_idle signal, schedule requests using `engine.crawl()` and raise `DontCloseSpider` exception.

### LINKED ISSUE #5282: Implement a decorator to feed start requests into the scheduler
@kmike described the issue best at https://github.com/scrapy/scrapy/issues/456#issuecomment-247622571:
> […] start_requests are not exhausted:
> 
> 1. Scrapy calls start_requests and gets enough requests to fill downloader
> 2. When new requests are scheduled (e.g. from responses) then scrapy pauses getting more requests from start_requests.
> 
> This is inconvenient if you e.g. have 100K websites to crawl and want to crawl their front pages (requests issued in start_requests), and follow some links on them (extracted in parse method). In this case even if you set a very high priority for requests sent in start_requests, they won't be processed. One have to use private APIs like `self.crawler.engine.schedule(request)` to make sure all 100K initial requests are scheduled.
> 
> It seems this is done in order to
> 
> 1. support start_requests with infinite generators;
> 2. send less requests to scheduler at the same time,
> 
> but I think this is implicit and confusing (and not really documented) […].

I think we could address this by providing a documentation topic about start requests, the scheduler, and other aspects related to request consumption order.

The documentation should describe how to achieve alternatives ways of request consumption, so along with the new documentation topic I would provide the following decorator, which addresses the issue in a way that that I prefer over alternatives I have seen proposed before, because instead of letting the Scrapy engine decide when to consume start requests, it gives that power to the scheduler, which I think is the ideal component to determine request order.

I already have an implementation for the decorator that I think can be used as is:

```py
def feed_scheduler(start_requests):
    """Use as a decorator for start_requests to feed its requests into the
    scheduler instead of yielding them directly."""
    def wrapper(spider):
        scheduler = spider.crawler.engine.slot.scheduler
        for request in start_requests(spider):
            scheduler.enqueue_request(request)
        request = scheduler.next_request()
        while request:
            yield request
            request = scheduler.next_request()
    return wrapper
```

I’m not creating a pull request for the decorator because I think we should only merge it along with the documentation I mentioned, which will take me some time to write properly. But I think this decorator can be useful already for people dealing with this issue, so I’m sharing it here already for easy copy-pasting into existing projects.

Gallaecio:
Allow customizing how seeds are iterated, use a greedy approach by default.

Split off from #6715, built on top of #6729.

A new `SEEDING_POLICY` setting allows choosing how Scrapy iterates values from `Spider.yield_seeds()`. It supports 4 different values:

-   `"greedy"` (default), to prioritize seeds over scheduled requests, which I believe is the new behavior proposed in #4467.

-   `"lazy"`, the current behavior, i.e. to prioritize scheduled requests over seeds.

-   `"front_load"`, to read all seeds into the scheduler before starting the crawl. Resolves #5282.

-   `"idle"`, to read a single seed whenever idle, i.e. like `"lazy"` assuming a concurrency of 1.

These policies are also exposed as values of a new `SeedingPolicy` enum, which can also be assigned to `SEEDING_POLICY` when setting its value from code.

`Spider.yield_seeds()` and `SpiderMiddleware.process_seeds()` can also yield `SeedingPolicy` values to override the configured behavior at run time at any time. For example, to enable ``"front_load"``, yield a batch of requests, and then switch to `"idle"`. Resolves #740, closes #1051, closes #1443, closes #3237.

-   Why change the behavior to greedy?
 
    > @kmike @ https://github.com/scrapy/scrapy/issues/456#issuecomment-247622571:
    >
    > My take on this: [#1051 (comment)](https://github.com/scrapy/scrapy/pull/1051#issuecomment-89024196). A related issue is that currently start_requests are not exhausted:
    > 
    > 1. Scrapy calls start_requests and gets enough requests to fill downloader
    > 2. When new requests are scheduled (e.g. from responses) then scrapy pauses getting more requests from start_requests.
    > 
    > This is inconvenient if you e.g. have 100K websites to crawl and want to crawl their front pages (requests issued in start_requests), and follow some links on them (extracted in parse method). In this case even if you set a very high priority for requests sent in start_requests, they won't be processed. One have to use private APIs like `self.crawler.engine.schedule(request)` to make sure all 100K initial requests are scheduled.
    > 
    > It seems this is done in order to
    > 
    > 1. support start_requests with infinite generators;
    > 2. send less requests to scheduler at the same time,
    > 
    > but I think this is implicit and confusing (and not really documented), and explicit `yield scrapy.WaitUntilIdle` (as in [#1051 (comment)](https://github.com/scrapy/scrapy/pull/1051#issuecomment-89024196)) is better. Currently something like WaitUntilIdle (but not quite the same) is inserted after each Request yielded from start_requests.

-   Do we need so many policies?

    I feel like `"lazy"` and `"greedy"` are a must have.

    `"idle"` should be trivial to implement, It would be quite useful as an override, e.g. for #740 scenarios. It is also a policy that we could use in scenarios where `"lazy"` and request priorities cannot lower resource usage enough.

    I have [personally](https://github.com/scrapy/scrapy/issues/5282) seen a need for `"front_load"` in some (niche) scenarios.

-   Can we have *more* policies?

    Sure, suggest away!

    I have considered only 1 more policy, but I am not sure if it is necessary or how to name it (`"drip"`?). It would be a middle ground between `"lazy"` and `"greedy"`: work like `"lazy"` but “trickle” seeds into the scheduler as well, e.g. try to read 1 seed per loop in a non-blocking way, à la [`get_nowait()`](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.get_nowait). I feel like this may be a better alternative to `"front_load"` for large amounts of seeds or async seeding, where you want to let the scheduler handle request prioritization as much as possible (e.g. to efficiently spread requests based on target domain, pending requests per domain and domain response times), but you don’t want to delay the crawl start until all requests have been scheduled.

-   Can we allow writing custom policies?

    I feel like this would be non-trivial to do right. I would rather go with built-in policies only, adding new ones as we see a need, and I am hoping that, together with the ability to override policies at run time, this will cover most (if not all) use cases.

-   Can we change the policy names?

    Yes! Suggest away!

    Some choices I considered:

    -   `"lazy"`: `"scheduled-first"`.

    -   `"greedy"`: `"seeds-first"`.

    -   `"front_load"`: `"single-scheduling"`.

    -   `"idle"`: `"serial"` (since it works similar to running spiders for each URL one by one), `"sloth"` (“lazier than lazy”), `"lazier"`, `"laziest"`.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -83,7 +83,7 @@ New features
 Bug fixes
 ~~~~~~~~~
 
--   Yielding an start item (i.e. from :meth:`~scrapy.Spider.yield_seeds` or an
+-   Yielding a start item (i.e. from :meth:`~scrapy.Spider.yield_seeds` or an
     equivalent) no longer delays the next iteration of starting requests and
     items by up to 5 seconds.
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -67,7 +67,7 @@ New features
     requests and items, e.g. reading them from a queue service or database
     using an asynchronous client, without workarounds.
 
-    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`)
+    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`, :issue:`6729`)
 
 -   The new :setting:`SEEDING_POLICY` setting allows customizing how start
     requests and items are iterated.
@@ -82,7 +82,7 @@ New features
         ability to override the active seeding policy.
 
     (:issue:`740`, :issue:`1051`, :issue:`1443`, :issue:`3237`, :issue:`4467`,
-    :issue:`5282`, :issue:`6715`)
+    :issue:`5282`, :issue:`6730`)
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -40,14 +40,16 @@ Deprecations
     use :meth:`~scrapy.Spider.yield_seeds` instead, or both to maintain support
     for lower Scrapy versions.
 
-    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`)
+    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`, :issue:`6715`,
+    :issue:`6729`)
 
 -   The ``process_start_requests()`` method of :ref:`spider middlewares
     <topics-spider-middleware>` is deprecated, use
     :meth:`~scrapy.spidermiddlewares.SpiderMiddleware.process_seeds` instead, or
     both to maintain support for lower Scrapy versions.
 
-    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`)
+    (:issue:`456`, :issue:`3477`, :issue:`4467`, :issue:`5627`, :issue:`6715`,
+    :issue:`6729`)
 
 New features
 ~~~~~~~~~~~~
@@ -65,6 +67,26 @@ New features
 
 -   The new :setting:`SEEDING_POLICY` setting allows customizing how spider
     start requests and items are consumed.
+
+    You can also override the active seeding policy from
+    :meth:`Spider.yield_seeds <scrapy.Spider.yield_seeds>` and from
+    :meth:`SpiderMiddleware.process_seeds
+    <scrapy.spidermiddlewares.SpiderMiddleware.process_seeds>`.
+
+    .. note:: Some third-party spider middlewares may need to be updated for
+        Scrapy VERSION support before you can use them in combination with the
+        ability to override the active seeding policy.
+
+    (:issue:`740`, :issue:`1051`, :issue:`1443`, :issue:`3237`, :issue:`4467`,
+    :issue:`5282`, :issue:`6715`)
+
+Bug fixes
+~~~~~~~~~
+
+-   The first :setting:`CONCURRENT_REQUESTS` start requests are no longer sent
+    in reserve order by default.
+
+    (:issue:`6715`, :issue:`6729`)
 
 
 .. _release-2.12.0:
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -11,13 +11,16 @@ Scrapy VERSION (unreleased)
 Highlights:
 
 -   Replaced ``start_requests`` (sync) with :meth:`~scrapy.Spider.yield_seeds`
-    (async)
+    (async) and changed how it is iterated by default.
 
 Backward-incompatible changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--   In ``scrapy.core.spidermw.SpiderMiddlewareManager``,
-    ``process_start_requests()`` has been replaced by ``process_seeds()``.
+-   By default, the iteration of start requests and items no longer stops once
+    there are requests in the scheduler.
+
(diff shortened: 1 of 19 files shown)


docs/news.rst
```diff
@@ -11,13 +11,16 @@ Scrapy VERSION (unreleased)
...
-    (async)
+    (async) and changed how it is iterated by default.
 
 Backward-incompatible changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--   In ``scrapy.core.spidermw.SpiderMiddlewareManager``,
-    ``process_start_requests()`` has been replaced by ``process_seeds()``.
+-   By default, the iteration of start requests and items no longer stops once
+    there are requests in the scheduler.
```
kmike, line 20: I think if we're making this change, it should go together with the new method nam.

docs/topics/spider-middleware.rst
```diff
@@ -83,9 +83,25 @@ one or more of these methods:
...
+        possible to yield a :class:`~scrapy.SeedingPolicy` enum or a matching
+        string to change the active seeding policy, for example:
+
+        .. code-block:: python
+
+            async def process_seeds(self, seeds):
+                yield "front_load"
+                async for seed in seeds:
+                    yield seed
+                yield "idle"
```
kmike, line 100: What do you think about a different approach?

1. The new method is an async generator (no change). 
2. Scrapy uses `async for` (or aiter) to always consume the async generator in full. This is not customizable via settings. 
3. For start methods which don't have `await`s inside it's similar to the "front_load" policy. It could yield the to event loop though (I'm not sure how it works, need to check it). I'm not sure it matters in practice (or maybe it is?).
4. If there is an `await` withing the generator, we don't do anything special, it just yields to the event loop. For example, if there is an await, and there are already requests in the scheduler, Scrapy should start downloading them, while waiting for something to be awaited in the "new start_requests" method.
5. To replicate the current behavior, or other required behaviors, we allow developers to await for something within their method. For example, one may call something like `await self.crawler.scheduler.queue_empty()` every time before sending a new Request.

So, in short: everything non-async is front-loaded. Things happen in parallel if there are awaits involved. The seeding behavior is controlled within the user method.

Example 1: regular start_requests, but with a new default behavior (where we put everything to the scheduler greedily):

```py
    async def start(self):   # should we support `def start`?
        for x in range(10):
            yield scrapy.Request(f"http://example.com/p/{x}", self.parse_page)
```

Example 2: reading from a queue

```py
    async def start(self):
        while True:
            next_url = await self.queue.pop()
            if next_url is None:  # some stopping condition, if needed
                break
            yield scrapy.Request(next_url, self.parse_page)
```

Example 3: limiting amount of stuff in the scheduler:

```py
    async def start(self):
        for url in self.urls:
            await self.crawler.scheduler.queue_empty()  # should it be a crawler attribute?
            yield scrapy.Request(url, self.parse_page)
```

It seems that for this to work we'd need to define a set of primitives which would allow users to implement proper seeding. They might be useful in other situations as well, not just in the new start_requests method.

Issue #6677, opened 2025-02-15
Provide a full list of asynchronous APIs


wRAR:
While we have a list of specifically coroutine APIs on https://docs.scrapy.org/en/latest/topics/coroutines.html#supported-callables it would be nice to have a list that additionally includes Deferred-based APIs, as in most cases they provide the same benefit to the user with just a small amount of glue code.

shinhonggwan:
Hello, I would like to contribute to this issue for my university project.

wRAR:
@shinhonggwan you can try but I'm not sure it's a good idea without being familiar with the codebase.

wRAR:
I've looked through the Scrapy code and hopefully found everything. Now the question is which of these we want to mention.


# Scrapy functions that user code can call

*Italic* denotes functions that are unlikely to be useful for user code.

## Returns a Deferred

- The crawler (`Crawler`):
    - `crawl()`
    - `stop()`
- `CrawlerRunner`:
    - `crawl()`
    - `stop()`
    - `join()`
- The engine (`ExecutionEngine`):
    - `start()`
    - `stop()`
    - `close()`
    - `download()`
    - *`open_spider()`*
    - *`close_spider()`*
- The scheduler (pluggable via the `SCHEDULER` setting, `Scheduler` by default, available as `crawler.engine.slot.scheduler`):
    - *`open()`*
    - *`close()`*
- The scraper (available as `engine.scraper`):
    - *`open_spider()`*
    - *`close_spider()`*
    - *`enqueue_scrape()`*
    - *`call_spider()`*
    - *`handle_spider_output()`*
    - *`start_itemproc()`*
- The downloader (pluggable via the `DOWNLOADER` setting, `Downloader` by default, available as `crawler.engine.downloader`):
    - `fetch()`
- The signal manager (`SignalManager`, available as `crawler.signals`):
    - `send_catch_log_deferred()`
- All subclasses of `MiddlewareManager` (`DownloaderMiddlewareManager`, `ExtensionManager`, `ItemPipelineManager`, `SpiderMiddlewareManager`):
    - *`open_spider()`*
    - *`close_spider()`*
- The downloader middleware manager (`DownloaderMiddlewareManager`, available as `crawler.engine.downloader.middleware`):
    - *`download()`*
- The spider middleware manager (`SpiderMiddlewareManager`, available as `crawler.engine.scraper.spidermw`):
    - *`scrape_response()`*
    - *`process_start_requests()`*
[...]

PR #6734, opened 2025-03-15
More docs about Deferred<->Future interoperability.

wRAR:
This improves the docs about wrapping deferreds, adds docs about wrapping futures and leaves a place for a list of deferred APIs for #6677


Gallaecio (approved):
:heart:

Issue #6731, opened 2025-03-14
SyntaxError occurs when dynamically modifying generator function source during runtime


TES286:
### Description

When a spider callback function is modified at runtime (e.g., during development with auto-reloading tools), Scrapy's `warn_on_generator_with_return_value` check may fail due to a syntax error caused by inconsistent code parsing. This occurs because `inspect.getsource()` retrieves the updated source code, but line/indentation calculations may reference the original version, leading to incorrect code extraction and AST parsing failures.

### Steps to Reproduce

1. Create a Scrapy spider with a generator-based callback (e.g., using `yield`).
2. Run the spider.
3. During runtime, modify the callback function in a way that changes line breaks or indentation (e.g., adding/removing parentheses across lines).
4. Trigger a new request to the modified callback.

**Expected behavior:** Scrapy should handle code changes gracefully or ignore runtime modifications.

**Actual behavior:** A `SyntaxError` (e.g., `unmatched ')'`) is raised during AST parsing due to mismatched code extraction.

**Reproduces how often:** 100% when code modifications alter line structure during runtime inspection.

### Versions

```
Scrapy       : 2.12.0
Python       : 3.13.2 
Platform     : Windows-11
```

### Additional context

**Root Cause**:  
The `is_generator_with_return_value()` function in `scrapy/utils/misc.py` uses `inspect.getsource()` to retrieve the current source code of a callback. If the function is modified during runtime (especially changes affecting line breaks/indentation), the parsed code snippet may be truncated incorrectly, causing AST failures.

**Minimal Reproduction (Non-Scrapy)**:  
Run this script and modify the `target()` function during execution:
```python
from inspect import getsource
import time

def target():
    # Edit this function during runtime
    yield 1

while True:
    try:
        src = getsource(target)
        ast.parse(src)
    except Exception as e:
        print(f"Failed: {e}")
    time.sleep(1)
```

**Affected Code**:  
The issue originates from [`scrapy/utils/misc.py` L245-L279](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/misc.py#L245-L279), where dynamic source retrieval interacts poorly with runtime code changes. While `inspect.getsource()` works as designed, Scrapy's logic doesn't account for source mutations after spider initialization.

**Suggested Fixes**:  
1. Cache the initial function source code during spider setup.
2. Add an option to disable generator return checks.
3. Implement robust error handling around AST parsing.

Gallaecio:
[...]

PR #6738, opened 2025-03-18
Syntax Error Fixed 

MehrazRumman:
Issue #6731

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -2047,6 +2047,19 @@ also used by :class:`~scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware
 if :setting:`ROBOTSTXT_USER_AGENT` setting is ``None`` and
 there is no overriding User-Agent header specified for the request.
 
+.. setting:: WARN_ON_GENERATOR_RETURN_VALUE
+WARN_ON_GENERATOR_RETURN_VALUE
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: ``True``
+
+Whether to show a warning when a generator-based callback uses a ``return <value>`` statement.
+In Scrapy, such return values are ignored, which might indicate a misunderstanding or potential bug
+in the callback logic.
+
+Set this to ``False`` to suppress these warnings.
+
+This setting helps developers catch unintended return values from generator-based callbacks early.
 
 Settings documented elsewhere:
 ------------------------------
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -2048,18 +2048,20 @@ if :setting:`ROBOTSTXT_USER_AGENT` setting is ``None`` and
 there is no overriding User-Agent header specified for the request.
 
 .. setting:: WARN_ON_GENERATOR_RETURN_VALUE
+
 WARN_ON_GENERATOR_RETURN_VALUE
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+------------------------------
 
 Default: ``True``
 
-Whether to show a warning when a generator-based callback uses a ``return <value>`` statement.
-In Scrapy, such return values are ignored, which might indicate a misunderstanding or potential bug
-in the callback logic.
+Type: ``bool``
 
-Set this to ``False`` to suppress these warnings.
+When enabled, Scrapy will warn if generator-based callback methods (like ``parse``) contain return statements with non-``None`` values. This helps detect potential mistakes in spider development.
 
-This setting helps developers catch unintended return values from generator-based callbacks early.
+Disable this setting to:
+- Prevent syntax errors when dynamically modifying generator function source code during runtime
+- Skip AST parsing of callback functions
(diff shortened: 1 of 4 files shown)


MehrazRumman:
I will add the test too !!

auvipy (changes requested):
also check why the tests are failing

Gallaecio:
I think the test failures may be due to some shortcuts taken in the current tests for the warning, i.e. the spider not being set up properly. Let me know if you need any help getting the existing tests to pass, your code looks fine.

MehrazRumman (commented):
I added the test cases !

MehrazRumman:
@Gallaecio  Can you help me to review this PR ?

scrapy/utils/misc.py
```diff
@@ -280,24 +280,32 @@ def returns_none(return_node: ast.Return) -> bool:
...
     a 'return' statement with a value different than None
     """
+    settings = getattr(spider, "settings", None)
+    getbool = getattr(settings, "getbool", None)
+    if (
+        settings
+        and callable(getbool)
+        and not getbool("WARN_ON_GENERATOR_RETURN_VALUE", True)
+    ):
+        return
```
Gallaecio, line 296: Is this logic really necessary at run time? Or are the (pre-existing) tests taking shortcuts and causing some things to not be defined during the tests, when those things are always defined when actually running a crawl?

MehrazRumman: Good point !! I actually the defensive checks because unit test uses a dummy spider object . I wanted to avoid attribute error but you are right !

MehrazRumman: I am fixing the issue !!

Gallaecio: > Let me know if you need any help getting the existing tests to pass, your code looks fine.

This is what I meant by this ↑.

Your code seemed fine to me as it was. If you have trouble making the test more real so that they pass with your changes, do let me know, it may not be trivial to fix.

MehrazRumman: I get your point now ! I just gave it a try to fix the test case . I introduce a mockspider setting attribute that has a getbool method and change all `None` to `MockSpider` .

auvipy (approved):

scrapy/utils/misc.py
```diff
@@ -286,6 +286,8 @@ def warn_on_generator_with_return_value(
     Logs a warning if a callable is a generator function and includes
     a 'return' statement with a value different than None
     """
+    if not spider.settings.getbool("WARN_ON_GENERATOR_RETURN_VALUE", True):
+        return
```
Gallaecio, line 290: Since the setting exists in `default_settings.py`, I believe it is safe to remove `True`:

```suggestion
    if not spider.settings.getbool("WARN_ON_GENERATOR_RETURN_VALUE"):
        return
```

Gallaecio (approved):

Gallaecio:
Could you add an entry about this new setting to  `settings.rst`?

Gallaecio (approved):

auvipy (approved):

docs/topics/settings.rst
```diff
@@ -2047,6 +2047,21 @@ also used by :class:`~scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware
...
+Default: ``True``
+
+Type: ``bool``
+
+When enabled, Scrapy will warn if generator-based callback methods (like ``parse``) contain return statements with non-``None`` values. This helps detect potential mistakes in spider development.
+
+Disable this setting to:
+- Prevent syntax errors when dynamically modifying generator function source code during runtime
+- Skip AST parsing of callback functions
+- Improve performance in auto-reloading development environments
```
wRAR, line 2064: Please check how is it rendered by running e.g. `tox -e docs` and opening the result (it's not rendered as a list).

docs/topics/settings.rst
```diff
@@ -2047,6 +2047,21 @@ also used by :class:`~scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware
...
+.. setting:: WARN_ON_GENERATOR_RETURN_VALUE
+
+WARN_ON_GENERATOR_RETURN_VALUE
+------------------------------
+
+Default: ``True``
+
+Type: ``bool``
+
+When enabled, Scrapy will warn if generator-based callback methods (like ``parse``) contain return statements with non-``None`` values. This helps detect potential mistakes in spider development.
```
wRAR, line 2059: We wrap lines in the docs (to about 80 chars).

MehrazRumman: Screenshot 2025-03-20 at 5 32 35 PM
this is the out put after running `tox -e docs`

MehrazRumman: @wRAR  is it ok now ? Or I need to update it further ?

wRAR:
Thanks!

wRAR (approved):

MehrazRumman:
Thanks for the guidelines @Gallaecio & @wRAR  ❤️

PR #6744, opened 2025-03-24
Remove deprecated classifier

Laerte:
https://peps.python.org/pep-0639/#deprecate-license-classifiers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,6 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: BSD License",
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


Gallaecio:
We could switch to a different builder as well, e.g.

```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

This is what I did in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/255 (as I migrated to `pyproject.toml` for an issue that turned out to be solvable by upgrading twine :sweat_smile:).

No strong opinion, though.

wRAR:
> We could switch to a different builder as well

Yeah, that's an option. I have no idea about hatchling though.

Laerte:
@wRAR I tested with `65.5.0` and `75.8.0` and worked locally, I had a [issue](https://github.com/pypa/setuptools/issues/4910) with setuptools earlier today and from that it as [find out](https://github.com/pypa/setuptools/issues/4913) that setuptools allow arbitrary fields, so appears that even that the field don't exist in a specific version it will not break the install, but let me know if you want to close this PR, thank you!

wRAR:
OK, I'll check this later

wRAR:
The current code in this PR emits a deprecation warning too, as I think this replaces one legacy option with another legacy option, and a less preferred one (https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#legacy-license-declaration). This is also why it works with older setuptools. Actually, only the warning for the PR code tells to fix it before 2026, the one for master doesn't. So I don't think we should merge this and we should either move to PEP 639 and require newer setuptools some time in 2024, but not in 2.13, or move to PEP 639 and switch to hatchling.

PR #6749, opened 2025-04-19
Add ability to include all tags and attributes to LinkExtractor.

### LINKED ISSUE #6321: Option to include all tags and attrs in LinkExtractor with specified exclusions

NickRagousis:
Feature: Implement option to include all tags and attributes in LinkExtractor

This pull request introduces new functionality to the LinkExtractor class, allowing users to consider all HTML tags and attributes for link extraction by setting the `tags` or `attrs` arguments to `True`.

Specific tags and attributes can then be excluded using the new `deny_tags` and `deny_attrs` arguments. This enables a more flexible approach where users can default to extracting all links and then selectively filter out unwanted ones, addressing the need for a more inclusive link extraction strategy.

I uncluded 13 new test cases with different combinations of tags, attrs, deny_tags, deny_attrs parmeters, to make sure the new parameters work as expected both in isolation and combined.

Note 1: With `attrs=True`, this implementation currently extracts values from attributes like `alt="something"` and `title="something"` in addition to actual link attributes like `href` and `src`. I wanted to inquire whether this is the desired behavior when the user intends to consider all attributes.

Note 2: This is my first ever open source contribution as part of a university assignment, so any feedback on this pull request would be greatly appreciated. 

Note 3: I have noticed that there is an open pull request on that issue that has been pending for several months, so I have attempted to provide an independent implementation.

Resolves #6321

### DIFF
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -186,22 +186,22 @@ class LxmlLinkExtractor:
         use_all_tags = tags is True
         use_all_attrs = attrs is True
 
-        tags_set = set() if use_all_tags else set(arg_to_iter(tags))
-        attrs_set = set() if use_all_attrs else set(arg_to_iter(attrs))
+        tags_set = set(arg_to_iter(tags)) if not use_all_tags else True
+        attrs_set = set(arg_to_iter(attrs)) if not use_all_attrs else True
         deny_tags_set = set(arg_to_iter(deny_tags))
         deny_attrs_set = set(arg_to_iter(deny_attrs))
 
-        def tag_checker(tag):
-            return tag not in deny_tags_set
+        def create_checker(allowed, denied):
+            if allowed is True:
+                return lambda x: x not in denied
 
-        if not use_all_tags:
-            tag_checker = partial(operator.contains, tags_set)
+            if denied:
+                return lambda x: x in allowed
 
-        def attr_checker(attr):
-            return attr not in deny_attrs_set
+            return partial(operator.contains, allowed)
 
-        if not use_all_attrs:
-            attr_checker = partial(operator.contains, attrs_set)
+        tag_checker = create_checker(tags_set, deny_tags_set)
+        attr_checker = create_checker(attrs_set, deny_attrs_set)
 
         self.link_extractor = LxmlParserLinkExtractor(
             tag=tag_checker,
--- a/scrapy/linkextractors/lxmlhtml.py
+++ b/scrapy/linkextractors/lxmlhtml.py
@@ -171,8 +171,10 @@ class LxmlLinkExtractor:
         allow_domains: str | Iterable[str] = (),
         deny_domains: str | Iterable[str] = (),
         restrict_xpaths: str | Iterable[str] = (),
-        tags: str | Iterable[str] = ("a", "area"),
-        attrs: str | Iterable[str] = ("href",),
+        tags: str | Iterable[str] | bool = ("a", "area"),
+        attrs: str | Iterable[str] | bool = ("href",),
+        deny_tags: str | Iterable[str] = (),
+        deny_attrs: str | Iterable[str] = (),
         canonicalize: bool = False,
         unique: bool = True,
         process_value: Callable[[Any], Any] | None = None,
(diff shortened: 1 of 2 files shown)


scrapy/linkextractors/lxmlhtml.py
```diff
@@ -181,10 +183,29 @@ def __init__(
...
-        tags, attrs = set(arg_to_iter(tags)), set(arg_to_iter(attrs))
+        use_all_tags = tags is True
+        use_all_attrs = attrs is True
+
+        tags_set = set() if use_all_tags else set(arg_to_iter(tags))
+        attrs_set = set() if use_all_attrs else set(arg_to_iter(attrs))
+        deny_tags_set = set(arg_to_iter(deny_tags))
+        deny_attrs_set = set(arg_to_iter(deny_attrs))
+
+        def tag_checker(tag):
```
rivtechprojects, line 194: Excellent changes. As a suggestion, you could consolidate the tag_checker into a single reusable function to reduce redundancy and improve maintainability.

```suggestion
     def create_checker(allowed, denied):
         if allowed is True:
             return lambda x: x not in denied
         if denied is True:
             return lambda x: x in allowed
         return partial(operator.contains, allowed)

     tag_checker = create_checker(tags_set, deny_tags_set)
     attr_checker = create_checker(attrs_set, deny_attrs_set)
```

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -181,10 +183,29 @@ def __init__(
...
+            return tag not in deny_tags_set
+
+        if not use_all_tags:
+            tag_checker = partial(operator.contains, tags_set)
+
+        def attr_checker(attr):
```
rivtechprojects, line 200: If the functionality of `attr_checker` is handled by `create_checker`, this function is no longer needed and can be safely removed. Here's the suggested change:
```suggestion

```

rivtechprojects:
@Gallaecio looks like prs for this issue have gone stale, but this contributor seems to have a reasonable fix. I've added a suggestion I think would be helpful. Any updates/feedback?

Gallaecio:
It’s in my to-do list to review it, but I am looking at a few busy weeks :sweat:

scrapy/linkextractors/lxmlhtml.py
NickRagousis: Thank you for the feedback! Your suggestion does make the code more maintainable. But one question. What do you think should happen if the same element is in both 'allowed' and 'denied' sets? Should we wait for Gallaecio's opinion?

rivtechprojects: That is a great question. I think there are two option to consider, and we should think about the intended behavior for such a conflict.

1. If denied should take precedence, then we would want to make sure those elements are excluded regardless of their presence in denied.
2. If the opposite is true and allowed should take precedence, then we should out any elements in denied before considering the allowed set.

I think it would make sense to wait for Gallaecio's opinion so we can fully understand the intended behavior, as this would impact how these conflicts are handled. My guess would be option 1 is preferred, because my instinct is denied should take precedence.

Option 1: denied takes precedence, it must be in allowed and not in denied (AND statement)
```
  def create_checker(allowed, denied):
      if allowed is True:
          return lambda x: x not in denied
      if denied is True:
          return lambda x: x in allowed
      # Denied takes precedence when both sets are provided
      return lambda x: x in allowed and x not in denied
```

Option 2: allowed takes precedence, it is still allowed if in both lists (OR statement)
```
  def create_checker(allowed, denied):
        if allowed is True:
            return lambda x: x not in denied
        if denied is True:
            return lambda x: x in allowed
      return lambda x: x in allowed or x not in denied
```

NickRagousis: Yeah it really depends on preference. We should wait for the response and discuss it, but thanks for all that interest, I appreciate it.

Gallaecio:
> few busy weeks

_5 months later…_

:eyes:

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -181,10 +183,29 @@ def __init__(
...
+        deny_attrs_set = set(arg_to_iter(deny_attrs))
+
+        def create_checker(allowed, denied):
+            if allowed is True:
+                return lambda x: x not in denied
+
+            if denied:
+                return lambda x: x in allowed
+
+            return partial(operator.contains, allowed)
```
Gallaecio, line 201: How are these different?

scrapy/linkextractors/lxmlhtml.py
```diff
@@ -171,8 +171,10 @@ def __init__(
         allow_domains: str | Iterable[str] = (),
         deny_domains: str | Iterable[str] = (),
         restrict_xpaths: str | Iterable[str] = (),
-        tags: str | Iterable[str] = ("a", "area"),
-        attrs: str | Iterable[str] = ("href",),
+        tags: str | Iterable[str] | bool = ("a", "area"),
+        attrs: str | Iterable[str] | bool = ("href",),
+        deny_tags: str | Iterable[str] = (),
+        deny_attrs: str | Iterable[str] = (),
```
Gallaecio, line 177: I think we might want to set all 4 parameters to `None` at this point, to easily detect when they are set (even to their intended default value) vs when they are not passed.

Then, later:
- If an “allow parameter” is used, its deny counterpart must be `None`, else a `ValueError` should be raised about it.
- If both are `None`, we assume the allow parameter with its default value (as was set in the function signature here).

Issue #6750, opened 2025-04-21
Deprecate passing spider instances to `StatsCollector` methods


wRAR:
Multiple methods of `StatsCollector` take a `spider` argument, even though `StatsCollector.__init__()` takes a `Crawler` instance and so it should already know the current spider.

A part of #6763

wRAR:
While doing this with keeping all backward compatibility turned out to be very hard, I wonder if there is a need for complete backward compatibility, as it seems the reason for those optional `spider` args is refactoring that happened in 2012 (81ed2d2d0b7cbf2dd5cffc0da50367c5284416a6), which is why those args aren't actually used inside `StatsCollector`.

PR #6751, opened 2025-04-22
Deprecate passing spider instances to StatsCollector methods

Laerte:
close: #6750


wRAR:
Actually, I need to think how would this impact 3rd-party `StatsCollector` subclasses / how to make this change but still keep compatibility (as we want to make this change in any case).

wRAR:
It would also be nice to remove the spider arg from `{open,close}_spider()`, this would require saving the spider or the crawler in the base `__init__()`, which is fine.

Laerte:
> It would also be nice to remove the spider arg from `{open,close}_spider()`, this would require saving the spider or the crawler in the base `__init__()`, which is fine.

I removed the spider from signals methods as well.

Laerte:
> Actually, I need to think how would this impact 3rd-party `StatsCollector` subclasses / how to make this change but still keep compatibility (as we want to make this change in any case).

I used `__getattribute__` approach so we emit the warning on subclasses as well, lmk what you think.

Laerte:
@wRAR lmk what you think or if this needs to be changed, thank you!

wRAR:
As I mentioned, making this change seems harder than I initially thought, so I haven't checked your implementation yet, but I think we need to make sure the user subclasses still work and so we need to add tests for those. Also: 

- I think the `spider` argument of `_persist_stats()` should be removed 
- I now wonder how to make these `open_spider()`/`close_spider()` changes backwards compatible (I haven't checked who calls them)
- I wonder if `DummyStatsCollector` produces deprecation warnings

Laerte:
> As I mentioned, making this change seems harder than I initially thought, so I haven't checked your implementation yet, but I think we need to make sure the user subclasses still work and so we need to add tests for those. Also:
> 
> * I think the `spider` argument of `_persist_stats()` should be removed
> * I now wonder how to make these `open_spider()`/`close_spider()` changes backwards compatible (I haven't checked who calls them)
> * I wonder if `DummyStatsCollector` produces deprecation warnings

- ~~I think the spider argument of _persist_stats() should be removed~~ - DONE
- ~~I now wonder how to make these open_spider()/close_spider() changes backwards compatible (I haven't checked who calls them)~~ - It seems that this is called on scrapy/core/engine.py.
- ~~I wonder if DummyStatsCollector produces deprecation warnings~~ - It does, I added a test for it my implementation cover this scenario to emit warning on custom stats classes.

wRAR:
> It does

Isn't that bad? :)

Laerte:
> > It does
> 
> Isn't that bad? :)

Is not the idea to notify people to remove the param?

wRAR:
Well, `DummyStatsCollector` is our class, so what I wondered was "if people set the setting to it, will they get warnings?"

Laerte:
> Well, `DummyStatsCollector` is our class, so what I wondered was "if people set the setting to it, will they get warnings?"

Only if they pass spider to one of the stats method, param:
```python
import scrapy


class Dummy(scrapy.Spider):
    start_urls = ["https://scrapy.org"]
    name = "dummy"

    custom_settings = {
        "STATS_CLASS": "scrapy.statscollectors.DummyStatsCollector",
    }

    def parse(self, response):
        self.crawler.stats.get_value("dummy_value", spider=self)
```

wRAR:
I've spent some time on doing easier changes (in #6764) and thinking how can and should we do this in pluggable components (which log formatters are) and I got some more understanding of this.

1. Changing the `close_spider()` signature seems hard even in non-pluggable ones, we even didn't change it in `ExecutionEngine` when we wanted, as the arg we want to remove is not the last positional one. It's a weird API to have the spider arg in `close_spider()` but not in `open_spider()` though. It's possible to deprecate passing anything except None as that arg to `close_spider()`, that looks less weird even though not ideal.
2. Changing the signatures of public methods is fine only if 3rd-party code that calls them doesn't break, so we cannot remove an argument right away, we need to make it optional and log a warning if the argument is passed. This of course applies to methods that can be overriden in the same way as to others. I think this part is already done in your PR.
3. Changing the calls of public methods that can't be overridden is fine but changing the calls of methods that can be overridden is AFAIK impossible (in a direct way): if a 3rd-party subclass really uses the argument, you can't stop passing it at all. I think this can be solved by a wrapper passing `self._crawler.spider` in our case, not sure if something like this was done in Scrapy before or it needs to be implemented from scratch.

Hope this is clear and I think it covers everything. And we still want to test subclasses in tests, that they work and that they produce suitable warnings.

PR #6752, opened 2025-04-22
Prioritize start requests over other requests by default, and keep them in yield order

Gallaecio:
https://github.com/scrapy/scrapy/pull/6729#discussion_r2040244861


docs/news.rst
```diff
@@ -51,6 +48,12 @@ Backward-incompatible changes
     instead of being defined as a generator, is now executed *after* the
     :ref:`scheduler <topics-scheduler>` instance has been created.
 
+-   When using :setting:`JOBDIR`, :ref:`start requests <start-requests>` are
+    now serialized into their own, ``s``-suffixed priority folders.
+
+    Set :setting:`SCHEDULER_START_DISK_QUEUE` to ``None`` or ``""`` to change
+    that.
```
kmike, line 55: It'd be good to document what happens when the option is set to None - it currently reads as it chages the behavior mentioned previously (separate folders, etc), but it doesn't tell what's the behavior after the chage.

scrapy/pqueues.py
```diff
@@ -115,29 +133,68 @@ def qfactory(self, key: int) -> QueueProtocol:
...
+            if priority not in self.queues:
+                self.queues[priority] = self.qfactory(priority)
+            q = self.queues[priority]
         q.push(request)  # this may fail (eg. serialization error)
         if self.curprio is None or priority < self.curprio:
             self.curprio = priority
 
     def pop(self) -> Request | None:
         if self.curprio is None:
             return None
```
wRAR, line 164: This can be removed now.

scrapy/spidermiddlewares/start.py
```diff
@@ -0,0 +1,38 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator, Iterable
+
+    from scrapy import Spider
+
+
+class StartSpiderMiddleware:
+    """Set ``"is_start_request"`` to ``True`` in the
+    :attr:`~scrapy.Request.meta` of :ref:`start requests <start-requests>`.
```
kmike, line 15: I think we should document the new meta key in the release notes, when mentioning that this new middleware is enabled by default.

We should also document it in https://docs.scrapy.org/en/latest/topics/request-response.html#request-meta-special-keys.

docs/topics/spiders.rst
```diff
@@ -369,70 +369,37 @@ See `Scrapyd documentation`_.
...
+-   Given the same :attr:`~scrapy.http.Request.priority`, start requests take
+    precedence.
 
--   To send start requests before other requests:
+-   Start requests are sent in the order they are yielded from
+    :meth:`~scrapy.Spider.start`, whereas other requests are sent in LIFO_
+    order.
 
-    .. code-block:: python
+    .. _LIFO: https://en.wikipedia.org/wiki/Stack_(abstract_data_type)
```
kmike, line 392: What do you think about linking to https://docs.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order? It seems it also makes sense to expand this docs section to mention start priorities. Or maybe even combine these two sections, not sure.

kmike: A nitpick: I'm asking mostly because of the next sentence, "you can change this in different ways", where it's not 100% clear what "this" is - it seems the intention was to explain how to customize start requests, but the last mention was of other requests. Anyways, it seems the explanation is useful for regular requests as well, and it partially overlaps with the FAQ entry.

scrapy/core/scheduler.py
```diff
@@ -126,56 +129,39 @@ def next_request(self) -> Request | None:
...
-    are backed by either memory or disk based queues (respectively defined by the
-    :setting:`SCHEDULER_MEMORY_QUEUE` and :setting:`SCHEDULER_DISK_QUEUE` settings).
+    are backed by either memory or disk based queues (respectively defined by
+    the :setting:`SCHEDULER_START_MEMORY_QUEUE` and
+    :setting:`SCHEDULER_START_DISK_QUEUE` settings for :ref:`start requests
+    <start-requests>` and by the :setting:`SCHEDULER_MEMORY_QUEUE` and
+    :setting:`SCHEDULER_DISK_QUEUE` settings for other requests).
 
     Request prioritization is almost entirely delegated to the priority queue. The only
     prioritization performed by this scheduler is using the disk-based queue if present
```
kmike, line 143: It also implicitly prefers start queues to regular queues.

scrapy/core/scheduler.py
```diff
@@ -195,23 +217,22 @@ def __init__(
...
-            logunser=crawler.settings.getbool("SCHEDULER_DEBUG"),
-            stats=crawler.stats,
-            pqclass=load_object(crawler.settings["SCHEDULER_PRIORITY_QUEUE"]),
-            crawler=crawler,
+        self._sdqclass: type[BaseQueue] | None = self._get_start_queue_cls(
+            crawler, "DISK"
         )
+        self._smqclass: type[BaseQueue] | None = self._get_start_queue_cls(
+            crawler, "MEMORY"
+        )
```
kmike, line 225: Is it done here (instead of passing instances to `__init__`) to simplify backwards compatibility handling?

scrapy/pqueues.py
```diff
@@ -115,29 +132,66 @@ def qfactory(self, key: int) -> QueueProtocol:
...
-            q.close()
-            prios = [p for p, q in self.queues.items() if q]
-            self.curprio = min(prios) if prios else None
-        return m
+        while self.curprio is not None:
+            if self._start_queues:
+                try:
+                    q = self._start_queues[self.curprio]
+                except KeyError:
+                    pass
```
kmike, line 167: Why is _update_curprio not called here? It seems `self._start_queues` affects the curprio.
If it should be added, then it probably make sense to have a small `_pop` helper with the popping logic, to avoid duplication.

kmike (approved):

docs/topics/spiders.rst
kmike: The new docs look good!

scrapy/core/scheduler.py
Gallaecio: Yes. I kind of default to that nowadays, specially for things that can be overridden through a setting. But it may make sense to allow subclasses to override these, maybe ignore the setting or use some other setting to decide.

@wRAR Shall I make them public?

If in doubt, we can discuss it with @kmike for a follow-up release.

scrapy/pqueues.py
Gallaecio: The logic is to call `self._update_curprio()` when *both* queues raise a `KeyError`. Here it is not called because self.curprio may still exist in `self.queues` below, and `self._update_curprio()` only needs to be changed once you know that neither queue has the current `curprio`.

scrapy/core/scheduler.py
kmike: I'm fine with either approach. My main concern is about consistency (some things are passed to `__init__`, others are created later), but it's not a big deal, if it helps to make a backwards compatible change with less code to maintain.

scrapy/pqueues.py
kmike: I see, makes sense!

scrapy/core/scheduler.py
wRAR: I'm fine with keeping the current code, it's not the first time we do newer things differently from the older ones.

scrapy/spidermiddlewares/start.py
```diff
@@ -0,0 +1,44 @@
...
+
+from scrapy.http import Request
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator, Iterable
+
+    from scrapy import Spider
+
+
+class StartSpiderMiddleware:
```
wRAR, line 13: This can now use `BaseSpiderMiddleware`

PR #6757, opened 2025-04-24
Release notes for 2.13.0

wRAR:


docs/news.rst
```diff
@@ -5,21 +5,209 @@ Release notes
...
+
+    - ``scrapy.utils.url.safe_url_string()``
+
+    - ``scrapy.utils.url.url_query_cleaner()``
+
+    - ``scrapy.utils.url.url_query_parameter()``
+
+    - ``scrapy.utils.url._unquotepath()``
+
+    - ``scrapy.utils.url._safe_chars`` attribute
```
Gallaecio, line 70: I wonder if we should remove these 2 from the notes, since they were already private.

docs/news.rst
```diff
@@ -5,21 +5,209 @@ Release notes
...
+    (:issue:`6544`, :issue:`6568`)
+
+-   Added the :setting:`DEFAULT_DROPITEM_LOG_LEVEL` setting and the
+    :attr:`scrapy.exceptions.DropItem.log_level` attribute that allow
+    customizing the log level of the message that is logged when an item is
+    dropped.
+    (:issue:`6603`, :issue:`6608`)
+
+-   Added the :setting:`LOG_VERSIONS` setting that allows customizing the
+    list of software which versions are logged when the spider starts.
```
Gallaecio, line 120:
```suggestion
    list of software whose versions are logged when the spider starts.
```

wRAR:
The test failure is interesting, there is probably a condition where the `ExecutionEngine._start_request_processing()` coroutine (which isn't awaited anywhere) didn't run or didn't finish before the spider was shut down.

Gallaecio:
Hmm… Should we try to address that now, or create a separate issue to look into it later?

Gallaecio (approved):

wRAR:
> Hmm… Should we try to address that now, or create a separate issue to look into it later?

It apparently doesn't happen often and shouldn't happen in normal (not shut down early) spiders so I'll create a separate issue.

Issue #6692, opened 2025-02-25
Slightly optimizing component fetch in Crawler class


shivanshuzyte:
## Summary

I was reading scrapy code and found that method get_extension can be slightly optimized:
https://github.com/scrapy/scrapy/blob/8c34e6d9a4994abda0059f76dbbdb64c2e8a9751/scrapy/crawler.py#L212C1-L229C1

If we can cache those lookup results so that if you call get_extension multiple times, you don’t re‑iterate over the list each time.

**Raw code:**

```
class Crawler:
    def __init__(self, spidercls, settings=None, init_reactor=False):
        # ... existing code ...
        self._component_cache = {}  # new cache

    @staticmethod
    def _get_component_from_list(component_class, components):
        for component in components:
            if isinstance(component, component_class):
                return component
        return None

    def get_extension(self, cls):
        if cls in self._component_cache:
            return self._component_cache[cls]
        if not self.extensions:
            raise RuntimeError(
                "Crawler.get_extension() can only be called after the "
                "extension manager has been created."
            )
        result = self._get_component_from_list(cls, self.extensions.middlewares)
        # Cache the result (even if None) for subsequent lookups
        self._component_cache[cls] = result
        return result
```

Gallaecio:
I’m not sure if it is worth optimizing, it should not be used very often, but I am not against it either. Although we may be able to cache it with some standard API, like lru or similar.

fbenevolo:
Hello!

I am interested in working on this issue. Can I do it?:

kmike:
What's the use case for this? Have you faced any performance issues? I'd be reluctant to accept this change unless there is a real-world benefit. It increases the amount of code we support, and it can change the behavior in some obscure reasons (e.g. when some component is changed after the list is already cached).

PR #6758, opened 2025-04-25
Add simple dictionary caching to get_extension method (#6692)

atishayk:
This PR implements a solution for issue #6692, adding caching to the `get_extension` method in the Crawler class to avoid redundant lookups.

## Implementation Details
- Added a simple dictionary cache (`self._extension_cache`) to store previously looked-up extensions
- The method now checks the cache first before performing a linear search through components
- Cache has unlimited size as extension count is typically small
- Both found and not-found results are cached for maximum efficiency

## Performance Results
Benchmarking shows that the cached implementation provides a consistent speedup over repeated calls to the `get_extension` method. While the exact performance improvement varies between runs, the cached version consistently outperforms the original implementation by eliminating redundant linear searches through the components list.

## Related Work
The existing PR #6754 implements a similar solution with a custom LRU cache implementation limited to 5 entries. This approach focuses on simplicity with an unlimited cache, as the typical number of extensions in Scrapy is small.

I'll also be submitting an alternative implementation using `functools.lru_cache` that leverages Python's built-in caching mechanism.

### DIFF
--- a/benchmark_extension_cache.py
+++ b/benchmark_extension_cache.py
@@ -1,0 +1,57 @@
+import time
+from scrapy.crawler import Crawler
+from scrapy import Spider
+from scrapy.settings import Settings
+
+# Define a simple spider class
+class MySpider(Spider):
+    name = 'myspider'
+
+# Define an extension class to look for
+class MyExtension:
+    pass
+
+def run_benchmark():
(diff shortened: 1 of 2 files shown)


wRAR:
The issue was closed as not planned, closing this too.

PR #6759, opened 2025-04-25
Extension cache lru decorator

atishayk:
# Add LRU cache to get_extension method (#6692)

This PR implements a solution for issue #6692, adding caching to the `get_extension` method in the `Crawler` class using Python's built-in `functools.lru_cache` decorator.

## Implementation Details
- Uses Python's standard library `functools.lru_cache` for caching solution
- Cache size is configurable with default maxsize=32 entries

## Related Work
The existing PR #6754 implements a similar solution with a custom LRU cache implementation. This approach differs by leveraging Python's standard library caching mechanism, which provides robust LRU functionality with less custom code.

I've also submitted an alternative implementation (https://github.com/scrapy/scrapy/pull/6758) using a simple dictionary cache which focuses on simplicity if you prefer that method.

### DIFF
--- a/benchmark_extension_cache.py
+++ b/benchmark_extension_cache.py
@@ -1,0 +1,57 @@
+import time
+from scrapy.crawler import Crawler
+from scrapy import Spider
+from scrapy.settings import Settings
+
+# Define a simple spider class
+class MySpider(Spider):
+    name = 'myspider'
+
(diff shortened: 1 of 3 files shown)


wRAR:
The issue was closed as not planned, closing this too.

PR #6760, opened 2025-04-26
Improve open_in_browser to handle missing head tags

### LINKED ISSUE #6550: Improve open_in_browser function's handling of base tags in response.py

vassilismylo:
# Fix open_in_browser to handle missing head tags

This PR improves the `open_in_browser` function to properly handle HTML responses that don't have a `<head>` tag. Previously, the function would fail to insert the `<base>` tag in such cases, causing relative links to break when viewing the page in a browser.

## Changes

- Modified the `open_in_browser` function to check for several HTML structure patterns
- Added a fallback mechanism that inserts both `<head>` and `<base>` tags when needed
- Implemented a cascading approach: first try `<head>`, then `<body>`, then `<html>`, and finally prepend to document

## Benefits

- More robust handling of various HTML structures
- Ensures relative URLs work properly in all viewed responses
- No new dependencies added

## Testing

Added new test cases that cover:
- HTML without head tags
- HTML with only body tags
- HTML with only html tags
- Minimal HTML with no structural tags

Fixes #6550

### DIFF
--- a/scrapy/utils/response.py
+++ b/scrapy/utils/response.py
@@ -98,9 +98,31 @@ def open_in_browser(
     if isinstance(response, HtmlResponse):
         if b"<base" not in body:
             _remove_html_comments(body)
-            repl = rf'\0<base href="{response.url}">'
-            body = re.sub(rb"<head(?:[^<>]*?>)", to_bytes(repl), body, count=1)
-        ext = ".html"
+            base_tag = f'<base href="{response.url}">'
+
+            # Try to insert after <head> tag first (existing behavior)
+            head_match = re.search(rb"<head(?:[^<>]*?>)", body)
+            if head_match:
+                repl = rf"\0{base_tag}"
+                body = re.sub(rb"<head(?:[^<>]*?>)", to_bytes(repl), body, count=1)
+            else:
+                # No head tag found, try to insert at beginning of body
+                body_match = re.search(rb"<body(?:[^<>]*?>)", body)
+                if body_match:
+                    repl = rf"\0<head>{base_tag}</head>"
+                    body = re.sub(rb"<body(?:[^<>]*?>)", to_bytes(repl), body, count=1)
+                else:
+                    # If no body tag, try finding html tag
+                    html_match = re.search(rb"<html(?:[^<>]*?>)", body)
+                    if html_match:
+                        repl = rf"\0<head>{base_tag}</head>"
+                        body = re.sub(
+                            rb"<html(?:[^<>]*?>)", to_bytes(repl), body, count=1
+                        )
+                    else:
+                        # As a last resort, prepend to document
+                        body = to_bytes(f"<head>{base_tag}</head>") + body
+            ext = ".html"
     elif isinstance(response, TextResponse):
         ext = ".txt"
     else:
--- a/scrapy/utils/response.py
+++ b/scrapy/utils/response.py
@@ -95,34 +95,26 @@ def open_in_browser(
 
     body = response.body
     if isinstance(response, HtmlResponse):
-        ext = ".html"
         if b"<base" not in body:
-            _remove_html_comments(body)
+            body = _remove_html_comments(body)
             base_tag = f'<base href="{response.url}">'
 
             # Try to insert after <head> tag first (existing behavior)
-            head_match = re.search(rb"<head(?:[^<>]*?>)", body)
-            if head_match:
+            if re.search(rb"<head[^<>]*?>", body):
                 repl = rf"\0{base_tag}"
-                body = re.sub(rb"<head(?:[^<>]*?>)", to_bytes(repl), body, count=1)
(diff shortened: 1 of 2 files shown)


vassilismylo:
fixing the failing check now

scrapy/utils/response.py
```diff
@@ -93,14 +93,36 @@ def parse_details(self, response):
...
+                    html_match = re.search(rb"<html(?:[^<>]*?>)", body)
+                    if html_match:
+                        repl = rf"\0<head>{base_tag}</head>"
+                        body = re.sub(
+                            rb"<html(?:[^<>]*?>)", to_bytes(repl), body, count=1
+                        )
+                    else:
+                        # As a last resort, prepend to document
+                        body = to_bytes(f"<head>{base_tag}</head>") + body
+            ext = ".html"
```
wRAR, line 125: This line is duplicated.

scrapy/utils/response.py
```diff
@@ -93,14 +93,36 @@ def parse_details(self, response):
...
         if b"<base" not in body:
             _remove_html_comments(body)
-            repl = rf'\0<base href="{response.url}">'
-            body = re.sub(rb"<head(?:[^<>]*?>)", to_bytes(repl), body, count=1)
-        ext = ".html"
+            base_tag = f'<base href="{response.url}">'
+
+            # Try to insert after <head> tag first (existing behavior)
+            head_match = re.search(rb"<head(?:[^<>]*?>)", body)
+            if head_match:
```
wRAR, line 105: I would rewrite these checks to use `:=`

tests/test_utils_response.py
```diff
@@ -204,6 +205,64 @@ def test_open_in_browser_redos_head(self):
         assert end_time - start_time < MAX_CPU_TIME
 
 
+class TestOpenInBrowserWithoutHead(unittest.TestCase):
```
wRAR, line 208: I don't think this needs to inherit from `unittest.TestCase`

vassilismylo: thank for replying, will fix these and create a new pull request

vassilismylo:
@wRAR  Thanks for the review! I've made the following improvements based on your feedback:

**Code style improvements:**
- Replaced nested if/else with clean elif chain for better readability
- Simplified regex patterns (removed unnecessary non-capturing groups)
- Removed unused walrus operator variables that were triggering linting errors - this is the commit message

- Unfortunatelly i did not write a correct commit message:
**Test improvements:**
- Test class no longer inherits from unittest.TestCase
- Using pathlib.Path instead for file operations
- Added comprehensive test coverage for all edge cases

scrapy/utils/response.py
```diff
@@ -93,13 +93,27 @@ def parse_details(self, response):
...
-            repl = rf'\0<base href="{response.url}">'
-            body = re.sub(rb"<head(?:[^<>]*?>)", to_bytes(repl), body, count=1)
+            body = _remove_html_comments(body)
+            base_tag = f'<base href="{response.url}">'
+
+            # Try to insert after <head> tag first (existing behavior)
+            if re.search(rb"<head[^<>]*?>", body):
+                repl = rf"\0{base_tag}"
+                body = re.sub(rb"<head[^<>]*?>", to_bytes(repl), body, count=1)
+            # No head tag found, try to insert at beginning of body
```
wRAR, line 106: I think this inserts `head` inside `body` but it should be inserted before it, not inside.

tests/test_utils_response.py
```diff
@@ -204,6 +204,91 @@ def test_open_in_browser_redos_head(self):
...
+        """Test with HTML that doesn't have a head tag"""
+        body = b"""<html>
+            <body><p>Hello world.</p></body>
+        </html>"""
+        response = HtmlResponse(self.url, body=body)
+
+        # Mock _openfunc to capture the written content
+        written_content = []
+
+        def mock_open(url):
```
wRAR, line 225: This function is repeated, can it be extracted?

tests/test_utils_response.py
```diff
@@ -204,6 +204,91 @@ def test_open_in_browser_redos_head(self):
         assert end_time - start_time < MAX_CPU_TIME
 
 
+class TestOpenInBrowserWithoutHead:
+    def setup_method(self):
+        self.url = "http://example.com"
```
wRAR, line 209: This can be a class attr

AdrianAtZyte:
Closing due to a lack of feedback.

Issue #6662, opened 2025-02-07
items.py fields order not respected if input_processor arg is not passed


AmmarSalmi:
### Description

In the exported ouput the order if the items' fields is not respected. The fields who have the arg input_processor passed when declared will always come first

### Steps to Reproduce
in items.py
```first_name = scrapy.Field(ouput_processor=TakeFirst())
last_name = scrapy.Field(input_processor=MapCompose(some_func), output_processor=TakeFirst())
```

**Expected behavior:** [What you expect to happen]
fields order in ouput
first_name, last_name

**Actual behavior:** [What actually happens]
 last_name , first_name
**Reproduces how often:** [What percentage of the time does it reproduce?]
100%
### Versions
Scrapy       : 2.12.0
lxml         : 5.3.0.0
libxml2      : 2.11.7
cssselect    : 1.2.0
parsel       : 1.10.0
w3lib        : 2.3.1
Twisted      : 24.11.0
Python       : 3.13.1 (tags/v3.13.1:0671451, Dec  3 2024, 19:06:28) [MSC v.1942 64 bit (AMD64)]
pyOpenSSL    : 25.0.0 (OpenSSL 3.4.0 22 Oct 2024)
cryptography : 44.0.0
Platform     : Windows-10-10.0.19045-SP0

wRAR:
I don't think this is promised anywhere but it would be nice to have.

fanisorfan:
Hi @wRAR,

I'm a university student working on an assignment, and I found this issue really interesting. Since it would enhance consistency in the exported data, I’d love to try implementing it as part of my coursework.

Would it be okay for me to work on this?

wRAR:
@fanisorfan sure

PR #6761, opened 2025-04-28
Solves #6662 - Adds support for ordered attributes in exporters

fanisorfan:
Solves https://github.com/scrapy/scrapy/issues/6662

Hey!

I've added functionality for ordered attributes by defining the `_ordered_attrs` attribute in `ItemMeta.__new__`, which tracks the order of attributes as they are defined in the `Item` class.

## Changes in `scrapy/item.py`:

```python
# scrapy/item.py
new_attrs["_ordered_attrs"] = [key for key in attrs if not key.startswith("_")]
```

The `_ordered_attrs` attribute stores the order of attributes as they are created, excluding private attributes that start with an underscore.

## Updates in `scrapy/exporters.py`:

To make use of the ordered attributes in the exporters, I've introduced a static method to fetch the ordered attributes from Item or dict objects:

```python
# scrapy/exporters.py::BaseItemExporter
@staticmethod
def _get_ordered_attrs(item: Any) -> list[str]:
    """Gets the ordered attributes list, if there are any.
    This is checking just for dict and Item types, as other types may not need this.
    """
    if isinstance(item, dict):
        return item.get("_ordered_attrs", [])

    if isinstance(item, Item):
        return item._ordered_attrs

    return []
```

## How exporters use this method:

For example, in JsonItemExporter, the export_item method has been updated to respect the order of attributes:

```python
# scrapy/exporters.py::JsonItemExporter
def export_item(self, item: Any) -> None:

    ordered_itemdict = {}
    ordered_attrs = self._get_ordered_attrs(item)
    for key in ordered_attrs:
        if key in itemdict:
            ordered_itemdict[key] = itemdict[key]

    if not ordered_attrs:
        ordered_itemdict = itemdict

    data = to_bytes(self.encoder.encode(ordered_itemdict), self.encoding)
    self._add_comma_after_first()
    self.file.write(data)
```

If the item has the `_ordered_attrs` attribute or if the dictionary keys are ordered in the same way, the JsonItemExporter will export the data in the correct order.

## How users can take advantage of this feature:

Users can now define the order of their attributes in the Item class, and the exporters will respect that order when serializing the data.

Here's how you can use it in your project:

### Use an exporter that respects the order (e.g., JsonItemExporter):

```python
# myproject/spiders/myspider.py
from scrapy.exporters import JsonItemExporter
from myproject.items import UnorderedAttributesItem
import scrapy

    name = "myspider"
    start_urls = ['http://example.com']

        self.file = open('output.json', 'wb')
        self.exporter = JsonItemExporter(self.file)
        self.exporter.start_exporting()

    def parse(self, response):
        item = UnorderedAttributesItem(
            z=1, y=2, a=3, x=4, c=5, b=6
        )
        self.exporter.export_item(item)
        yield item

    def close(self, reason):
        self.exporter.finish_exporting()
        self.file.close()
```

## What happens here:

The `UnorderedAttributesItem` class defines the order of the attributes using the `_ordered_attrs` attribute.
When exporting, the `JsonItemExporter` will respect this order and serialize the data accordingly.
For example, the output JSON file will look like this:

```json
[{"z": 1, "y": 2, "a": 3, "x": 4, "c": 5, "b": 6}]
```

The order of the fields is exactly as defined in `_ordered_attrs` or in the order the fields are added to the Item.
This work builds upon @jaymeklein's initial commit.

### DIFF
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -367,6 +367,12 @@ class PprintItemExporter(BaseItemExporter):
 
 
 class PythonItemExporter(BaseItemExporter):
+    """This is a base class for item exporters that extends
+    :class:`BaseItemExporter` with support for nested items.
+    It serializes items to built-in Python types, so that any serialization
+    library (e.g. :mod:`json` or msgpack_) can be used on top of it.
+    .. _msgpack: https://pypi.org/project/msgpack/
+    """
     def _configure(self, options: dict[str, Any], dont_fail: bool = False) -> None:
         super()._configure(options, dont_fail)
         if not self.encoding:
--- a/scrapy/exporters.py
+++ b/scrapy/exporters.py
@@ -82,48 +82,40 @@ class BaseItemExporter:
     def _get_serialized_fields(
         self, item: Any, default_value: Any = None, include_empty: bool | None = None
     ) -> Iterable[tuple[str, Any]]:
-        """Return the fields to export as an iterable of tuples
-        (name, serialized_value)
-        """
+        """Return the fields to export as an iterable of tuples (name, serialized_value)."""
         item = ItemAdapter(item)
 
         if include_empty is None:
             include_empty = self.export_empty_fields
 
+        ordered_attrs = self._get_ordered_attrs(item)
+
+        field_iter: Iterable[str] | Iterable[tuple[str, str]]
         if self.fields_to_export is None:
-            field_iter = item.field_names() if include_empty else item.keys()
-        elif isinstance(self.fields_to_export, Mapping):
-            if include_empty:
-                field_iter = self.fields_to_export.items()
+            if ordered_attrs:
+                field_iter = [f for f in ordered_attrs if f in item]
             else:
                 field_iter = (
-                    (x, y) for x, y in self.fields_to_export.items() if x in item
+                    list(str(f) for f in item.field_names())
+                    if include_empty
+                    else list(str(f) for f in item.keys())
                 )
+        elif isinstance(self.fields_to_export, Mapping):
+            if include_empty:
+                field_iter = list(self.fields_to_export.items())
+            else:
+                field_iter = [(x, y) for x, y in self.fields_to_export.items() if x in item]
         elif include_empty:
-            field_iter = self.fields_to_export
+            field_iter = list(str(f) for f in self.fields_to_export)
         else:
-            field_iter = (x for x in self.fields_to_export if x in item)
+            field_iter = [str(x) for x in self.fields_to_export if x in item]
 
         for field_name in field_iter:
             if isinstance(field_name, str):
                 item_field, output_field = field_name, field_name
             else:
                 item_field, output_field = field_name
-            if item_field in item:
-                field_meta = item.get_field_meta(item_field)
-                value = self.serialize_field(field_meta, output_field, item[item_field])
-            else:
-                value = default_value
-        
-        ordered_attrs = self._get_ordered_attrs(item)
-        if ordered_attrs:
-            field_iter = (f for f in ordered_attrs if f in item)
 
-        for field_name in field_iter:
-            if isinstance(field_name, str):
-                item_field, output_field = field_name, field_name
-            else:
-                item_field, output_field = field_name
             if item_field in item:
                 field_meta = item.get_field_meta(item_field)
                 value = self.serialize_field(field_meta, output_field, item[item_field])
@@ -131,6 +123,10 @@ class BaseItemExporter:
                 value = default_value
 
             yield output_field, value
+
+
+
+
 
 
 class JsonLinesItemExporter(BaseItemExporter):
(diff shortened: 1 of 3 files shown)


fanisorfan:
I'm uploading the new PR as requested. 

Regarding the numerous line changes, they were all generated by running:
```
pre-commit run --all-files
```

Most changes are whitespace-related and changes like:
```diff
-  </script> 
+  </script>
```

These formatting adjustments ensure code consistency according to our style guidelines. The actual functional code remains unchanged.

wRAR:
Have you seen that you committed `docs/_build/html`?

fanisorfan:
@wRAR Thank you for the feedback!
I was quite sure that the pre-commit hook modified some files automatically and I didn't realize that docs/_build/html got included unintentionally. I didn't intentionally modify the generated documentation files.
How would you suggest I proceed? Should I remove the docs/_build/html files from this PR?
Just as a note, this is one of my first pull requests, so I really appreciate your patience and any guidance!

fanisorfan:
@wRAR I have now removed `docs/_build/html` from the PR. Thank you for the review and guidance!

scrapy/exporters.py
```diff
@@ -72,33 +82,43 @@ def finish_exporting(self) -> None:
...
+                field_iter = [
                     (x, y) for x, y in self.fields_to_export.items() if x in item
-                )
+                ]
         elif include_empty:
-            field_iter = self.fields_to_export
+            field_iter = [str(f) for f in self.fields_to_export]
         else:
-            field_iter = (x for x in self.fields_to_export if x in item)
+            field_iter = [str(x) for x in self.fields_to_export if x in item]
```
wRAR, line 114: Can you explain all of these changes?

scrapy/exporters.py
```diff
@@ -54,6 +54,16 @@ def _configure(self, options: dict[str, Any], dont_fail: bool = False) -> None:
         if not dont_fail and options:
             raise TypeError(f"Unexpected options: {', '.join(options.keys())}")
 
+    @staticmethod
+    def _get_ordered_attrs(item: Any) -> list[str]:
+        """Gets the ordered attributes list, if there are any.
+        This is checking just for dict and Item types, i don't know of any
```
wRAR, line 60: The comment is no longer correct.

scrapy/exporters.py
```diff
@@ -369,5 +405,4 @@ def _serialize_item(self, item: Any) -> Iterable[tuple[str | bytes, Any]]:
             yield key, self._serialize_value(value)
 
     def export_item(self, item: Any) -> dict[str | bytes, Any]:  # type: ignore[override]
-        result: dict[str | bytes, Any] = dict(self._get_serialized_fields(item))
-        return result
+        return dict(self._get_serialized_fields(item))
```
wRAR, line 408: These are just style changes, right?

scrapy/exporters.py
fanisorfan: I replaced the generators with lists to make the code simpler and ensure predictable iteration, especially since generators can be exhausted after use.
While I realize that in some cases the number of fields might not be small, for exporters it is usually manageable, and the clarity and safety benefits seemed to outweigh the memory cost.
Of course, I'm happy to revisit this if you think a generator would be more appropriate here!

scrapy/exporters.py
fanisorfan: Working on it!

scrapy/exporters.py
fanisorfan: Yes, exactly.
This part is purely a cleanup: I removed the unreachable second return statement to match the intended style .

scrapy/exporters.py
fanisorfan: Do you approve of this ? 

        """Return the ordered attributes if available.
        This checks if the item is a dict and returns its "_ordered_attrs" key
        if it exists. Otherwise, it returns an empty list.
        """

scrapy/exporters.py
wRAR: I don't think it makes the code simpler and I'm not sure if there could be problems with iteration and exhaustion (if there are, it may be needed to change the code, sure, but you don't sound like they actually exist).

scrapy/exporters.py
wRAR: If you are talking about changes compared to the previous PR, you should actually look at the changes compared to master. This also applies to the change in `MarshalItemExporter`.

scrapy/exporters.py
wRAR: A bigger question is why this only supports dicts (unlike what you say in the first message of this PR) but I haven't actually checked the actual implementation changes yet.

scrapy/exporters.py
fanisorfan: That's alright, to avoid problems I can upload the version with generators like before.

scrapy/exporters.py
fanisorfan: Okey, just checked the master i am switching it back.

PR #6766, opened 2025-05-01
Improve test coverage of Scraper.

wRAR:
This is also important for future refactorings of this code.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -374,32 +374,24 @@ class Scraper:
         errors that got propagated thru here).
 
         spider_failure: the value passed into the errback of self.call_spider()
+        (likely raised in the request errback)
+
         download_failure: the value passed into _scrape2() from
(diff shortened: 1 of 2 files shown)


scrapy/core/scraper.py
```diff
@@ -374,32 +374,24 @@ def _log_download_errors(
...
         spider_failure: the value passed into the errback of self.call_spider()
+        (likely raised in the request errback)
+
         download_failure: the value passed into _scrape2() from
         ExecutionEngine._handle_downloader_output() as "result"
+        (likely raised in the download handler or a downloader middleware)
         """
         if not download_failure.check(IgnoreRequest):
             assert self.crawler.spider
-            if download_failure.frames:
```
wRAR, line 382: It shouldn't be possible to get a failure value without a traceback here, but also this branch works fine for such faliure values so the alternative one shouldn't be needed anyway.

Gallaecio (approved):

PR #6767, opened 2025-05-02
Refactor Scraper to use inlineCallbacks + cleanup.

wRAR:
This switches the API from callbacks to `inlineCallbacks`, merges some functions into their callers and simplifies the logic. There should be no behavior changes.

1. Changing the APIs to `inlineCallbacks` should be straightforward by itself. 
2. Some function or Deferred return values were found to never be used and were replaced with None.
3. `_scrape2()` and `_log_download_errors()` are merged into `_scrape()` and the result is refactored.
4. `_itemproc_finished()` is merged into `start_itemproc()`.
5. `call_spider()` is slightly refactored.
6. `scrapy.utils.defer._defer_sleep()` is added as a more simple replacement for `defer_succeed()` and `defer_fail()` in `inlineCallbacks` functions.

An observation: "log and send a signal" blocks are cumbersome and there are too many of them but I don't see a way to increase code readability around them.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -192,7 +192,7 @@ class Scraper:
             try:
                 # call the spider middlewares and the request callback with the response
                 output = await maybe_deferred_to_future(
-                    self.spidermw.scrape_response(  # type: ignore[return-value]
+                    self.spidermw.scrape_response(
                         self.call_spider, result, request, self.crawler.spider
                     )
                 )
--- a/scrapy/core/spidermw.py
+++ b/scrapy/core/spidermw.py
@@ -40,7 +40,8 @@ logger = logging.getLogger(__name__)
 
 _T = TypeVar("_T")
 ScrapeFunc = Callable[
-    [Union[Response, Failure], Request], Union[Iterable[_T], AsyncIterable[_T]]
+    [Union[Response, Failure], Request],
+    Deferred[Union[Iterable[_T], AsyncIterable[_T]]],
 ]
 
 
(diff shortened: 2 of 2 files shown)


scrapy/core/scraper.py
```diff
@@ -156,103 +155,101 @@ def enqueue_scrape(
...
-            iterate_spider_output
-        )
-        return dfd2
+            if not request.errback:
+                result.raiseException()
+            warn_on_generator_with_return_value(self.crawler.spider, request.errback)
+            output = request.errback(result)
+            if isinstance(output, Failure):
+                output.raiseException()
+            # else the errback returned actual output (like a callback),
```
wRAR, line 250: I don't think we document this.

But Gemini knows that this is very much possible, there is at least one test that yields a dict from an errback and there is e.g. https://stackoverflow.com/a/59217406/113586

scrapy/core/engine.py
```diff
@@ -250,11 +250,10 @@ def _remove_request(_: Any) -> None:
         )
         return d2
 
+    @inlineCallbacks
     def _handle_downloader_output(
```
Gallaecio, line 254: Did you consider `async def` with the deferred decorator? I know we cannot use it at some key places, but I wonder if this is one of them.

Gallaecio (approved):

wRAR:
I've just added async def alternatives to `call_spider()` and `handle_spider_output()`. These are only needed if we consider those public APIs, otherwise we can just replace them.

wRAR:
No, the old `call_spider()` is still needed (until we rewrite some parts of `SpiderMiddlewareManager`) but `handle_spider_output()` isn't, and it's even not covered by tests.

PR #6769, opened 2025-05-02
refactor tests/test_downloadermiddleware_httpcache.py

### LINKED ISSUE #6649: Refactor tests/test_downloadermiddleware_httpcache.py

fbenevolo:
Fixes #6649 

I did a refactoring so that the test options that are being covered are more clear. What I did was the following:
- Remove of policy_class and storage_class attributes from TestBase class
- Added StorageTestMixin and PolicyTestMixin classes that contain tests specific from storage and policy, respectively
- Added DummyPolicyTestMixin and RFC2616PolicyTestMixin that contain tests specific from those options
- Finally, added classes that contain all possible combinations between storage and policy options. Those classes are:
        TestFilesystemStorageWithDummyPolicy
        TestFilesystemStorageWithRFC2616Policy
        TestDbmStorageWithDummyPolicy
        TestDbmStorageWithRFC2616Policy
        TestDbmStorageWithCustomDbmModule (inherits from DummyPolicy version)
        TestFilesystemStorageGzipWithDummyPolicy (inherits from Filesystem version)

### DIFF
--- a/tests/test_downloadermiddleware_httpcache.py
+++ b/tests/test_downloadermiddleware_httpcache.py
@@ -15,8 +15,7 @@ from scrapy.utils.test import get_crawler
 
 
 class TestBase:
-    storage_class = "scrapy.extensions.httpcache.FilesystemCacheStorage"
-    policy_class = "scrapy.extensions.httpcache.RFC2616Policy"
+    """Base class with common setup and helper methods."""
 
     def setup_method(self):
         self.yesterday = email.utils.formatdate(time.time() - 86400)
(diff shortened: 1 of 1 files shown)


wRAR:
Looks good, thanks!

wRAR (approved):

PR #6774, opened 2025-05-05
Refactor Scraper to use inlineCallbacks + cleanup

Gallaecio:
Copy of #6767 against #6729, to look into why this change breaks lazy support. I will undo the merging of #6767 in the main branch.

cc @wRAR


Gallaecio:
@wRAR Should we try to address this for 2.13 or post-pone it?

wRAR:
It doesn't bring value to 2.13, I actually wanted to postpone it.

PR #6776, opened 2025-05-05
Add a list of Deferred-only APIs.

### LINKED ISSUE #6677: Provide a full list of asynchronous APIs

wRAR:
Fixes #6677

### DIFF
--- a/docs/topics/coroutines.rst
+++ b/docs/topics/coroutines.rst
@@ -62,17 +62,102 @@ In addition to native coroutine APIs Scrapy has some APIs that return a
 :class:`~twisted.internet.defer.Deferred` object or take a user-supplied
 function that returns a :class:`~twisted.internet.defer.Deferred` object. These
 APIs are also asynchronous but don't yet support native ``async def`` syntax.
-For example:
+In the future we plan to add support for the ``async def`` syntax to these APIs
+or replace them with other APIs where changing the existing ones is
+possible.
 
--   The :meth:`ExecutionEngine.download` method returns a
-    :class:`~twisted.internet.defer.Deferred` object.
--   A custom download handler needs to define a ``download_request()`` method that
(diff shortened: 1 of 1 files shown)


docs/topics/coroutines.rst
```diff
@@ -62,18 +62,103 @@ In addition to native coroutine APIs Scrapy has some APIs that return a
...
+-   The :meth:`ExecutionEngine.download()
+    <scrapy.core.engine.ExecutionEngine.download>` method returns a
+    :class:`~twisted.internet.defer.Deferred` object that fires with the
+    downloaded response. You can use this object directly in Deferred-based
+    code or convert it into a :class:`~asyncio.Future` object with
+    :func:`~scrapy.utils.defer.maybe_deferred_to_future`.
+-   A custom download handler needs to define a ``download_request()`` method
+    that returns a :class:`~twisted.internet.defer.Deferred` object. You can
+    write a method that works with Deferreds and returns one directly, or you
+    can write a coroutine and convert it into a functions that returns a
```
Gallaecio, line 159:
```suggestion
    can write a coroutine and convert it into a function that returns a
```

I also wonder if reStructuredText may have an issue with there not being an empty line between these list items, but not sure.

Gallaecio (approved):

docs/topics/coroutines.rst
wRAR: [master](https://docs.scrapy.org/en/master/topics/coroutines.html#using-deferred-based-apis) has no empty line either, I moved and expanded this paragraph.

PR #6777, opened 2025-05-06
Add process_start support to BaseSpiderMiddleware.

wRAR:


wRAR:
* `DepthMiddleware`: no behavior change: it would make sense to set depth=0 for start requests and that would likely even simplify the code but I'm not sure how safe it is
* `HttpErrorMiddleware`: no behavior change: no start seed or spider output processing
* `OffsiteMiddleware`: no behavior change: deprecated, explicitly doesn't process start requests as opposed to the newer downloader middleware
* `RefererMiddleware`: no behavior change: needs the previous response
* `UrlLengthMiddleware`: only this one has a behavior change: the check now applies to start requests; we could remove it but it makes sense to keep it

Gallaecio (approved):

PR #6796, opened 2025-05-14
Add AsyncCrawlerRunner.

### LINKED ISSUE #6789: Add `AsyncCrawlerRunner`
We need a coroutine-based counterpart to `CrawlerRunner`, as adding wrapped methods to `CrawlerRunner` is either clumsy or impossible (you need to know, or assume, whether the asyncio reactor *will be* installed). 

I've made the class itself, a larger part is rewriting https://docs.scrapy.org/en/latest/topics/practices.html to use it. There are also other things that can be improved there, including suggesting `twisted.internet.task.react()` and/or `CrawlerRunner.join()`. It's likely that these changes will also address #6047.

wRAR:
Fixes #6789 

Includes a partial fix for #6047. Partial because the `CrawlerProcess` examples (which don't save the result of `crawl()` and don't call `join()`, and `CrawlerProcess.start()` eats the exceptions from `join()` which it calls) are still aftected.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,6 +19,8 @@ collect_ignore = [
     "tests/mockserver.py",
     "tests/pipelines.py",
     "tests/spiders.py",
+    # contains scripts to be run by tests/test_crawler.py::AsyncCrawlerRunnerSubprocess
+    *_py_files("tests/AsyncCrawlerRunner"),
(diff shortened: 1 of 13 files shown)


Gallaecio (approved):
:tada:

Issue #6707, opened 2025-03-09
Add an aiohtp-based download handler


wRAR:
It should be possible and may be interesting to create and ship a download handler that uses aiohttp instead of Twisted to make HTTP requests. It may even help with some shortcomings of the current handler, including #4951.

thalissonvs:
It should be an optional handler?

wRAR:
If you mean whether it should replace the current default ones then it shouldn't, unless we see reasons for that.

wRAR:
#1455 :)

PaulyBearCoding:
Hi! I’d like to contribute to this issue as my first Scrapy pull request. Is it still open for work? If so, I’ll review the code and current documentation, and suggest improvements as described. Thanks!

wRAR:
@PaulyBearCoding as you can see there is already a PR in progress

Ousret:
maybe one day you'd consider [Niquests](https://github.com/jawah/niquests) for the complete coverage of sync/async http1 through http3. it perform at least twice at fast as any regular http client under concurrency pressure and is rock solid. could potentially simplify a lot of things here. open to talk about it anytime if interested.

regards,

_disclaimer: I am involved with it_

wRAR:
@Ousret thanks, we haven't yet decided which non-Twisted handler will be added to Scrapy and if there should be one or several. We'll consider niquests too.

Ousret:
Thanks for your response, and consideration,

> thanks, we haven't yet decided which non-Twisted handler will be added to Scrapy

This is what I understood so far. So this could be an opportunity.

If there any live debate or session about this, I'll be happy to participate.

Regards,

wRAR:
Available in https://github.com/scrapy-plugins/scrapy-download-handlers-incubator, we will be able to merge it into Scrapy when we decide to do it.

Issue #6783, opened 2025-05-08
`test_close_during_start_iteration()` can fail with a Task error about `_start_request_processing()`


wRAR:
> 2025-05-08 08:30:49 [scrapy.core.engine] INFO: Spider closed (shutdown)
> 2025-05-08 08:30:49 [asyncio] ERROR: Task was destroyed but it is pending!
> task: <Task pending name='Task-2848' coro=<ExecutionEngine._start_request_processing() running at /home/runner/work/scrapy/scrapy/scrapy/core/engine.py:225> wait_for=<Future pending cb=[Deferred.asFuture.<locals>.checkCancel() at /home/runner/work/scrapy/scrapy/.tox/pinned/lib/python3.9/site-packages/twisted/internet/defer.py:992, <TaskWakeupMethWrapper object at 0x7f613f2f4850>()]> cb=[Deferred.fromFuture.<locals>.adapt() at /home/runner/work/scrapy/scrapy/.tox/pinned/lib/python3.9/site-packages/twisted/internet/defer.py:1029]>

As `ExecutionEngine._start_request_processing()` is not awaited explicitly, it should indeed be possible that it's not finished (or even called?) yet before the spider finishes, as this test specifically closes the spider early.

wRAR:
It looks like it happens relatively often only on 3.9/pinned. I wonder if it's because of older Twisted.

wRAR:
Per my stress tests, following https://github.com/scrapy/scrapy/pull/6889#issuecomment-2972831004, this fails on Twisted < 24.7.0 but seems to not fail on that version. OTOH it may be easier to reap the task somewhere than to bump the Twisted version to such a recent one, I'll try to make and test some simple fixes later.

Gallaecio:
> it may be easier to reap the task somewhere

If that is so, great.

However, I personally don’t feel like a 1-year-old version of Twisted is too much to ask to be able to run the latest Scrapy.

wRAR:
I've found that `tests/test_engine.py::TestEngine::test_start_already_running_exception` reliably prints the same "Task was destroyed but it is pending!" message on all envs. I wouldn't even be surprised that it's the same message sometimes captured by LogCapture in the failing test (as I seem to remember such bleeding of messages between tests sometimes happens), though both tests deal with an early stopping of the engine and so the failures could be independent.

[...]

Issue #4951, opened 2021-01-09
Identical requests sent by Scrapy vs Requests module returning different status codes


pmdbt:
### Description

Recently a spider I made to crawl cragislist for rental listings broke. When I checked the logs, it turns out that all of my requests were hitting http 403 error codes. Of course, I assumed the issue would be from not setting proper headers and not using proxies etc, so I went ahead and added auto user-agent and header rotation as well as adding in proxy servers. None of this helped. In a last-ditch effort, I wrote a simple **GET** request using the [requests](https://pypi.org/project/requests/) module. Well, somehow this default request ended up working on the same URLs with 200 status codes even though it's the same IP address without any proxy servers or user-agents configured.

I don't understand exactly how the request is sent out by **Scrapy** vs **requests module**, but even when I configured both to share the exact same request headers, one returns 403 error while the other returns 200. It seems I'm also not the only one to experience this weird result based on [this](https://stackoverflow.com/questions/65295966/craigslist-403-blocking-urlopen-but-not-requests-get) StackOverflow post.

### Steps to Reproduce

1. Set up a default Scrapy spider with only default settings active.

2. Install the latest version of requests and make a default GET request to any site using `request.get("any site")`. Get the headers used by this default request. For me it was:
```
GET / HTTP/1.1
Host: localhost:8080
User-Agent: python-requests/2.25.1
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
```
3. Configure the headers of the Scrapy spider request call to have the exact same headers from step 2.
```python
                scrapy.Request(
                    url="any website",
                    callback=self.parse,
                    headers={
                        "User-Agent": "python-requests/2.25.1",
                        "Accept-Encoding": "gzip, deflate",
                        "Accept": "*/*",
                        "Connection": "keep-alive"
                        })
```

4. Start a Netcat server locally to make sure Scrapy and requests will send the same request object. I started mine on port 8080 with the command `nc -l 8080`. Now change the request URLs for both Scrapy and requests to **"http://localhost:8080"**. Run both and examine the results.

For me, I see the following from Netcat for the request sent with **requests module**:
```
GET / HTTP/1.1
Host: localhost:8080
User-Agent: python-requests/2.25.1
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
```
And I see the following from the **Scrapy Spider's request**:
```
GET / HTTP/1.1
User-Agent: python-requests/2.25.1
[...]

PR #6798, opened 2025-05-14
Aiohttp downloadhandler

novykovd:
should solve #6707 
this aiohttp dh passes all tox envs and the http test and should be ready to use, but admittedly this is my first contribution so if this is missing some functionality or something needs to be changed or tested im happy to adjust always


wRAR:
This is interesting! I have several comments and I think we have more tests in `tests/test_downloader_handlers.py` that can be attempted to run with this (the class structure there is very weird but at least you can skip the 1.0 parts).

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,85 @@
+import asyncio
+from typing import Self
```
wRAR, line 2: Please import this from typing_extensions like we do in other files (and under `if TYPE_CHECKING`, and with `from __future__ import annotations`)

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,85 @@
...
+    set_asyncio_event_loop,
+    verify_installed_reactor,
+)
+
+
+class AiohttpDownloadHandler:
+    def __init__(self, settings: Settings, crawler: Crawler):
+        verify_installed_reactor(
+            "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
+        )
```
wRAR, line 24: I think this should follow https://docs.scrapy.org/en/latest/topics/asyncio.html#enforcing-asyncio-as-a-requirement

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,85 @@
...
+        self._crawler = crawler
+
+    @classmethod
+    def from_crawler(cls, crawler: Crawler) -> Self:
+        return cls(crawler.settings, crawler)
+
+    def download_request(self, request: Request, spider: Spider) -> Deferred[Response]:
+        return Deferred.fromFuture(
+            asyncio.ensure_future(self._download_request(request))
+        )
```
wRAR, line 42: Please use `deferred_from_coro()` here and in other places

scrapy/settings/default_settings.py
```diff
@@ -74,6 +74,7 @@
     "data": "scrapy.core.downloader.handlers.datauri.DataURIDownloadHandler",
     "file": "scrapy.core.downloader.handlers.file.FileDownloadHandler",
     "http": "scrapy.core.downloader.handlers.http.HTTPDownloadHandler",
+    "aiohttp": "scrapy.core.downloader.handlers.aiohttp.AiohttpDownloadHandler",
```
wRAR, line 77: I don't think this does anything as the keys are URL schemes :)

tests/test_downloader_handlers.py
```diff
@@ -442,6 +443,12 @@ class TestHttps10(TestHttp10):
     scheme = "https"
 
 
+class TestAiohttp(TestHttp):
```
wRAR, line 446: As the tested code requires the asyncio reactor its tests should be decorated with `@pytest.mark.only_asyncio`

scrapy/settings/default_settings.py
novykovd: its fixed but damn this feels like an extra embarrassing mistake. in my defense i just looked at the downloader code and it looked like the scheme was an additional thing you can specify in the request as opposed to being tied to the url

novykovd:
additionally some of the tests for py 3.11-3.13 failed on the first push for unix systems because there was a test that is skipped for windows. i added code that hopefully fixes it but im not running ubuntu or macos at this time so im not sure whether its fully fixed

wRAR:
I think for the `test_timeout_download*` tests you just need to add `asyncio.TimeoutError` to the list of expected exceptions.

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+            raise ValueError(
+                "AiohttpDownloadHandler requires the asyncio Twisted "
+                "reactor. Make sure you have it configured in the "
+                "TWISTED_REACTOR setting. See the asyncio documentation "
+                "of Scrapy for more information."
+            )
+
+        verify_installed_reactor(
+            "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
+        )
```
wRAR, line 40: This is no longer needed.

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+
+    from scrapy.crawler import Crawler
+    from scrapy.http.request import Request
+    from scrapy.http.response import Response
+    from scrapy.settings import Settings
+    from scrapy.spiders import Spider
+
+
+class AiohttpDownloadHandler:
+    def __init__(self, settings: Settings, crawler: Crawler):
```
wRAR, line 29: the settings arg is not needed

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+                headers=new_headers, url=request.url
+            )
+            return respcls(
+                url=request.url, status=status, headers=new_headers, body=body
+            )
+
+    def close(self):
+        return deferred_from_coro(self.session.close())
+
+    async def _close(self):
```
wRAR, line 101: I think this is unused.

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+from scrapy.http.headers import Headers
+from scrapy.utils.defer import deferred_from_coro
+from scrapy.utils.reactor import (
+    is_asyncio_reactor_installed,
+    set_asyncio_event_loop,
+    verify_installed_reactor,
+)
+
+if TYPE_CHECKING:
+    from typing import Self
```
wRAR, line 17: This is not the correct fix

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+        """download through aiohttp interface"""
+
+        proxy = request.meta.get("proxy")
+        timeout = aiohttp.ClientTimeout(total=request.meta.get("download_timeout"))
+
+        url = request.url
+        method = request.method
+        body = request.body
+        headers = None if request.headers is None else request.headers.to_unicode_dict()
+        if isinstance(request.cookies, dict):
```
wRAR, line 68: I think this is not needed, as `request.cookies` is converted to the `Cookie` headers by `CookiesMiddleware`.

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+            cookies = request.cookies
+        else:
+            cookies = {
+                str(item["name"]): str(item["value"]) for item in request.cookies
+            }
+        body = request.body
```
wRAR, line 74: This is duplicate

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.session.cookie_jar.update_cookies(cookies)
```
wRAR, line 76: Hmm I'm not sure what should we do with the builtin aiohttp cookie handling as we want to do that ourselves.

novykovd:
im not sure how my code has an effect on the engine tests in the case that is failing atm but i will look into it

wRAR:
That's just #6783

novykovd:
i thought maybe its some versioning thing because it fails on a lower python version, but ig my code is fine right?

wRAR:
Yes, the problem is unrelated to the changes in this PR.

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,93 @@
...
+                new_headers = Headers(response.raw_headers)
+
+                respcls = responsetypes.responsetypes.from_args(
+                    headers=new_headers, url=request.url
+                )
+                return respcls(
+                    url=request.url, status=status, headers=new_headers, body=body
+                )
+        except asyncio.TimeoutError:
+            raise error.TimeoutError
```
wRAR, line 87: You shouldn't do this. I've meant changing the test expectations to also expect the asyncio exception.

novykovd:
okk so it raises and asserts the right exception i think, only thing is im not sure if there is a better/more elegant way of changing the standard http tests to account for the ansyncio exception. i was thinking maybe i can somehow call super and then add another errback but i dont think its normal to call test functions yourself and im pretty sure assert wont allow me that anyway in this case

scrapy/core/downloader/handlers/aiohttp.py
```diff
@@ -0,0 +1,92 @@
...
+                new_headers = Headers(response.raw_headers)
+
+                respcls = responsetypes.responsetypes.from_args(
+                    headers=new_headers, url=request.url
+                )
+                return respcls(
+                    url=request.url, status=status, headers=new_headers, body=body
+                )
+        except asyncio.TimeoutError as e:
+            raise e
```
wRAR, line 86: ?

wRAR: I mean, why is this except+raise needed at all?

novykovd:
this issue looks similar to that previous one i think? in both cases its something with task being pending it seems

novykovd:
do you think its ready or should i change something else?

scrapy/core/downloader/handlers/aiohttp.py
wRAR: This still applies

wRAR:
As for the tests, I would wait until #6821 is merged, after which it should be easier to subclass test base classes. I don't expect this handler to pass all of them as e.g. dataloss and maxsize handling is not implemented, but it would be better to subclass all test classes and mark the unimplemented ones with xfail.

scrapy/core/downloader/handlers/aiohttp.py
Gallaecio: In case further clarification helps:

```python
try:
    foo()
except SomeException as e:
    raise e
```

It functionally equivalent to:

```python
foo()
```

You should only be using a try-except block if you are doing some exception handling beyond re-raising the exception you get.

novykovd:
@wRAR sorry for the break, if no alternative aiohttp downloader has been merged over this time then i did some cookies testing (local docker and a small spider instance) and cookies are echoed just fine. aiohttp session also has its own jar which i turned off in a new commit so that all of the logic is on scrapy. but i am not sure if that is your concern? did you have something different in mind?

wRAR:
My concern is that we don't seem to have tests for such things, not that the code is actually incorrect :) I didn't have time yet to think about such bacis tests. On the other hand we decided to initially merge include this download handler as a test-only one so it's fine to not have these tests initially. Hopefully I'll let you know what do I think about next steps in the next week.

novykovd:
im guessing you want to make sure that some issues that aiohttp dh could resolve are resolved like #4951? or do you think there is more to restructure with the tests? or is it missing some functionality that the standard http handler has? all of the above? i could look into any of them

wRAR:
It's more simple than that: any HTTP/HTTPS download handler should do such basic things as:

* actually return the response body that the server returned
* same with headers 
* same with the response code
* etc;

and, on the other hand, to not do such things as:

* decompress the response
* follow redirects
* store cookies and insert them into a next request
* etc.

I don't think we have tests for most of these requirements in the download handler tests (the ones that we would run for all handlers). We implicitly test some of these in other tests (the ones that use the default handler). So while it should be easy to notice with a real-world spider that anything of this is broken, we still should have tests for these. The things from the first list are unlikely to be broken but things from the second list are things that are useful for someone using e.g. aiohttp directly and so likely to be enabled by default (e.g. your handler sets `allow_redirects=False`, it would be easy to miss that).

wRAR:
The solution for this, I think, is simply writing more tests in `TestHttpBase`, I don't think we want to run e.g. `test_downloadermiddleware_httpcompression.py` for all available download handlers.

wRAR:
I've created #6946 about adding more basic tests.

wRAR:
Thanks for your work but I'm closing this in favor of #6805 and/or the aiohttp handler implementation I have locally (which support most of the common download handler features, including the tests mentioned in the ticket above).

Issue #6791, opened 2025-05-12
Add `ExecutionEngine.download_async()`


wRAR:
A coroutine counterpart to `ExecutionEngine.download()`.

The implementation is trivial but it also needs tests (`download()` doesn't have them either), ~~and `RobotsTxtMiddleware` should be rewritten to use the new function~~.

fbenevolo:
Just to confirm: is it download() or _download()?

wRAR:
`download()`, the public one, suggested on https://docs.scrapy.org/en/latest/topics/item-pipeline.html#take-screenshot-of-item

PR #6802, opened 2025-05-15
Refactor RobotsTxtMiddleware to use async def process_request().

wRAR:
For #6791, but also useful by itself

### DIFF
--- a/scrapy/downloadermiddlewares/robotstxt.py
+++ b/scrapy/downloadermiddlewares/robotstxt.py
@@ -9,18 +9,16 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -107,35 +98,32 @@ def robot_parser(
...
         if isinstance(parser, Deferred):
-            d: Deferred[RobotParser | None] = Deferred()
-
-            def cb(result: RobotParser | None) -> RobotParser | None:
-                d.callback(result)
-                return result
-
-            parser.addCallback(cb)
-            return d
+            return await maybe_deferred_to_future(parser)
```
wRAR, line 114: Looks like this is not covered by the tests.

PR #6804, opened 2025-05-17
Deprecate _parse_response and implement parse_with_rules

Rodrigosnrocha:
Resolves issue #4463. I tried to stick to the solutions discussed in the issue whenever possible. The parse_with_rules method behaves like the prior _parse_response method, but is a public method instead. _parse_response now issues a deprecation warning on use, and when overridden by a subclass. Also added the appropriate tests to check if the warnings are being issued correctly, though I can certainly add more if needed

### DIFF
--- a/scrapy/spiders/crawl.py
+++ b/scrapy/spiders/crawl.py
@@ -8,6 +8,7 @@ See documentation in docs/topics/spiders.rst
 from __future__ import annotations
 
 import copy
+import warnings
 from collections.abc import AsyncIterator, Awaitable, Callable
 from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
@@ -97,7 +98,7 @@ class CrawlSpider(Spider):
         self._compile_rules()
 
     def _parse(self, response: Response, **kwargs: Any) -> Any:
-        return self._parse_response(
+        return self.parse_with_rules(
             response=response,
             callback=self.parse_start_url,
             cb_kwargs=kwargs,
@@ -137,7 +138,7 @@ class CrawlSpider(Spider):
 
     def _callback(self, response: Response, **cb_kwargs: Any) -> Any:
         rule = self._rules[cast(int, response.meta["rule"])]
-        return self._parse_response(
+        return self.parse_with_rules(
             response,
             cast("CallbackT", rule.callback),
             {**rule.cb_kwargs, **cb_kwargs},
@@ -150,7 +151,7 @@ class CrawlSpider(Spider):
             failure, cast(Callable[[Failure], Any], rule.errback)
         )
 
-    async def _parse_response(
+    async def parse_with_rules(
         self,
         response: Response,
         callback: CallbackT | None,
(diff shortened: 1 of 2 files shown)


Rodrigosnrocha:
After taking a look at the failed checks I'm really not sure what's causing them to fail? I'll take another look, and add an additional check in test_parse_response_use to get up to full test coverage (instead of one line only having partial coverage)

wRAR:
I think all failures are unrelated to your code.

scrapy/spiders/crawl.py
```diff
@@ -171,6 +179,22 @@ async def _parse_response(
             for request_or_item in self._requests_to_follow(response):
                 yield request_or_item
 
+    def _parse_response(
+        self,
+        response: Response,
+        callback: CallbackT | None,
+        cb_kwargs: dict[str, Any],
+        follow: bool = True,
+        warn: bool = True,
```
wRAR, line 188: What's the purpose of this flag?

tests/test_spider.py
```diff
@@ -476,6 +476,53 @@ class TestSpider(self.spider_class):
...
+            start_urls = "https://www.example.com"
+            _follow_links = False
+
+        with warnings.catch_warnings(record=True) as w:
+            spider = _CrawlSpider()
+            assert len(w) == 0
+            spider._parse_response(
+                TextResponse(spider.start_urls, body=b""), None, None
+            )
+            print(len(w))
```
wRAR, line 491: These prints shouldn't be here.

scrapy/spiders/crawl.py
Rodrigosnrocha: In the discussion for issue #4463 it was suggested to add a flag to the existing _parse_response method to suppress the deprecation warning, via a parameter with a default value of True, and I initially felt it best to include that suggestion in this PR.
Since that suggestion was made under the assumption that parse_with_rules would call _parse_response and not the other way around (and thus would have needed a way to avoid the warning), it may be better to remove the flag to avoid situations where users set it to False manually and forget to replace _parse_response until it is too late

wRAR: > Since that suggestion was made under the assumption that parse_with_rules would call _parse_response and not the other way around

Yes, so there is no need for the flag.

wRAR (approved):
Thanks, this looks good.

wRAR:
I've just rechecked older PRs and I see some suggestions from @Gallaecio that could be applied to your code too, I'll let him decide if they are needed.

scrapy/spiders/crawl.py
```diff
@@ -95,9 +97,15 @@ class CrawlSpider(Spider):
     def __init__(self, *a: Any, **kw: Any):
         super().__init__(*a, **kw)
         self._compile_rules()
+        if method_is_overridden(self.__class__, CrawlSpider, "_parse_response"):
+            warnings.warn(
+                "CrawlSpider._parse_response method is deprecated: "
+                "it will be removed in future Scrapy releases. "
+                "Please override CrawlSpider.parse_with_rules method instead."
```
Gallaecio, line 104:
```suggestion
                "The CrawlSpider._parse_response method is deprecated: "
                "it will be removed in future Scrapy releases. "
                "Please override the CrawlSpider.parse_with_rules method instead."
```

Gallaecio (approved):
Thanks!

PR #6806, opened 2025-05-17
Implement 201 status code handling in FilesPipeline #Fixes1615

Manan-Wadhwa:
The FilesPipeline currently doesn't handle HTTP 201 responses with Location headers correctly. When interacting with certain APIs, a 201 Created response is returned with a Location header pointing to the actual file to download. Without proper handling, this causes file downloads to fail.

Add logic to the media_downloaded method to detect HTTP 201 responses, extract the Location header, and create a new request to download the file from the location provided in the header.

Added unit tests to verify the behavior

Was tested using mock server serving 201 with location header and spider


wRAR:
Your `ruff` ignores the project configuration.

PR #6807, opened 2025-05-17
Update files.py

Manan-Wadhwa:


wRAR:
Is this still in progress?

Manan-Wadhwa:
will fix in a while finals week started forgot about it

PR #5922, opened 2023-05-05
feat: add disallowed_domains option to OffsiteMiddleware

felipecustodio:
Adds feature requested in #2376 

There is most likely a cleaner way to write the checks for the existence of the lists/arguments!

**TODO:**

- [ ] Documentation
- [ ] Add test cases to `tests/test_spidermiddleware_offsite.py`
- [ ] Code review and code style check

PR #6808, opened 2025-05-17
feat: Adding disallowed_domains to OffsiteMiddleware

### LINKED ISSUE #2376: disallow_domains
Feature request
Currently there's "allowed_domains" to create a whitelist of domains to scrape.

It would be good if there was a "disallowed_domains" or "blocked_domains" as well. I appreciate I could could probably do this in middleware, but I figure it's something quite a few people would want.

DiogoCastro:
Resolves #2376, closes #5922.

Continuing Felipe's work: https://github.com/scrapy/scrapy/pull/5922
Added the code and will be implementing the tests and documentation soon


Gallaecio:
Currently it seems that the disallowing approach may disallow unintended domains.

I think that, rather than editing get_host_regex, we might want to address disallowed domains differently, regexps are not designed for disallowing I think.

tests/test_downloadermiddleware_offsite.py
```diff
@@ -322,8 +322,7 @@ def test_process_request_invalid_disallowed_domains():
 
     for letter in ("b", "c"):
         request = Request(f"https://{letter}.example")
-        with pytest.raises(IgnoreRequest):
-            mw.process_request(request)
+        assert mw.process_request(request) is None
```
AdrianAtZyte: Why are we no longer asserting that IgnoreRequest is raised for b and c?

I am fine with not supporting invalid domain values in disallowed_domains, but if we do that, I wonder if we should have the spider not start or something as a result, and error out instead.

AdrianAtZyte:
Something else flagged by an LLM as a low-importance issue, I leave it to you do decide whether it is a valid issue or not, and how to best address it if valid:

> New disallowed_domains test coverage is incomplete for the request_scheduled signal path.
> - Why this matters: process_request and request_scheduled are expected to remain behaviorally equivalent. For disallowed_domains, parity tests were added only for process_request scenarios (dont_filter, allow_offsite, invalid entries), not request_scheduled.
> - Missing parity compared to existing allowed_domains style: test_downloadermiddleware_offsite.py, test_downloadermiddleware_offsite.py, test_downloadermiddleware_offsite.py
> - Only one request_scheduled disallowed test exists (no disallowed domains case): test_downloadermiddleware_offsite.py

tests/test_downloadermiddleware_offsite.py
DiogoCastro: Hmm, now that you mentioned, it makes sense. I guess we can throw an error in that case.
My idea here is to update the `_process_domains` to accept a "strict mode", where, once it's true, we can ensure that we throw a ValueError on any possible invalid domain, so the spider fails at startup.
I'll try to find some time today or tomorrow to implement that.

But I'm totally open to other possible solutions as well, if you think that's not quite the best, tho.

AdrianAtZyte: I would not bother with having multiple modes. If a value is not a valid domain, it’s a developer error, it is fine to require them to fix the error before the code can work, like it would happen if it was a Python syntax error or something.

Do check if raising an exception is enough to prevent the spider from moving on. In some cases, you need to actually trigger the spider to close (with an appropriate close reason).

DiogoCastro: Hey! Finally had some time to take a better look at this.
I have updated the code to raise `ValueError` on invalid entries
I noticed that simply raising in `spider_opened` was not enough to stop the spider, so I have added an except block to handle that, scheduling a spider close with reason "invalid_domain_configuration"
I have also updated the tests. Now it should be covering the checks for the error log instead of how it was before.

AdrianAtZyte:
@DiogoCastro Need to merge the latest version of the main branch into this one and resolve conflicts in the process to let CI work.

scrapy/downloadermiddlewares/offsite.py
```diff
@@ -38,7 +38,23 @@ def from_crawler(cls, crawler: Crawler) -> Self:
...
+            self.disallowed_host_regex: re.Pattern[str] | None = (
+                self._get_disallowed_host_regex(spider)
+            )
+        except ValueError as exc:
+            logger.error(
+                "Invalid domain configuration: %(error)s",
+                {"error": exc},
+                extra={"spider": spider},
+            )
+            if self.crawler.engine:
```
AdrianAtZyte: Do not use `if`, use `assert`.

tests/test_downloadermiddleware_offsite.py
```diff
@@ -115,20 +116,31 @@ def test_process_request_no_allowed_domains(value):
     assert mw.process_request(request) is None
 
 
-def test_process_request_invalid_domains():
+def test_process_request_invalid_domains(caplog):
     crawler = get_crawler(Spider)
     allowed_domains = ["a.example", None, "http:////b.example", "//c.example"]
```
AdrianAtZyte: Instead of putting the list here, use pytest’s parametrize to define 3 parameters sets, with each of the 3 invalid domains, to make sure all are handled as invalid. As it currently stands, if only 1 of the 3 was treated as an invalid domain, the test would still pass.

(I do realize this comes from the pre-existing code, but it seems worth addressing here)

scrapy/downloadermiddlewares/offsite.py
```diff
@@ -65,35 +81,67 @@ def process_request(self, request: Request, spider: Spider | None = None) -> Non
...
             if domain is None:
-                continue
+                raise ValueError(f"{domains_type} contains empty value.")
             if url_pattern.match(domain):
-                message = (
-                    "allowed_domains accepts only domains, not URLs. "
-                    f"Ignoring URL entry {domain} in allowed_domains."
+                raise ValueError(
+                    f"{domains_type} accepts only domains, not URLs. "
+                    f"Ignoring URL entry {domain} in {domains_type}."
```
AdrianAtZyte: We are no longer "ignoring" these, so the error messages may need an update.

tests/test_downloadermiddleware_offsite.py
AdrianAtZyte, line 1: It would be nice to have a test that covers combining both allowed and disallowed domains, e.g. allow example.com but disable ads.example.com, and verify that things work as expected.

tests/test_downloadermiddleware_offsite.py
```diff
@@ -115,20 +116,42 @@ def test_process_request_no_allowed_domains(value):
     assert mw.process_request(request) is None
 
 
-def test_process_request_invalid_domains():
+@pytest.mark.parametrize(
+    "allowed_domains",
+    [
+        ["a.example", None],
+        ["a.example", "http://b.example"],
+        ["a.example", "c.example:8080"],
```
AdrianAtZyte, line 126: You have changed this value. Why not keep both, the new and the old (//.c.example)?

DiogoCastro: Hmm, I guess it slipped through the current validation. This fix should be quick. I just need to add another regex to check for this pattern and we should be good. I will add it, and hopefully everything should be fine now.

scrapy/downloadermiddlewares/offsite.py
```diff
@@ -65,35 +81,68 @@ def process_request(self, request: Request, spider: Spider | None = None) -> Non
...
     def should_follow(self, request: Request, spider: Spider) -> bool:
-        regex = self.host_regex
         # hostname can be None for wrong urls (like javascript links)
         host = urlparse_cached(request).hostname or ""
-        return bool(regex.search(host))
+        # If the host matches a disallowed domain, we can reject it
+        if self.disallowed_host_regex and self.disallowed_host_regex.search(host):
+            return False
+        # Otherwise, check allowed domains
+        return bool(self.host_regex.search(host))
```
AdrianAtZyte: 💄 I feel these comments are superfluous.

AdrianAtZyte (approved):
I feel like some of the tests are a bit excessive, to be honest, but no strong opinion on it. Good work!

PR #6817, opened 2025-05-20
Add `AsyncCrawlerProcess` and `Crawler.crawl_async()`

### LINKED ISSUE #6790: Add `AsyncCrawlerProcess`?
Unlike `CrawlerRunner`, `CrawlerProcess` doesn't define any public Deferred methods, but it inherits the `CrawlerRunner` ones, and so after #6789 it's likely that we will need `AsyncCrawlerProcess`.

### LINKED ISSUE #6792: Add `crawl_async()` and `stop_async()` to `Crawler`
`Crawler.crawl()` calls `Crawl._apply_settings()` and the final value of `TWISTED_REACTOR` is known only there (because of `self.addons.load_settings(self.settings)`, and the reactor is also installed inside that function, so we cannot call `maybe_deferred_to_coro()` before that, but it may be possible to refactor the `Crawler` code so that it's possible to learn the reactor type in the place where it's needed.

Alternatively, there could be a separate `AsyncCrawler` similar to `AsyncCrawlerRunner`, that also assumes the asyncio reactor and simplifies the code.

wRAR:
Fixes #6792, fixes #6790. Still doesn't progress #6047, I need to think about it later and it's not a blocker, I just hoped we will get it for free.

It was necessary to stop supporting overriding the `ASYNCIO_EVENT_LOOP` value in add-ons and spiders when using `AsyncCrawlerProcess` because it's hard and maybe impossible to support that without some other compromise, and this solution doesn't require extensive code changes. We moved the reactor initialization from early to late in #6038, because we wanted users to still be able to set `TWISTED_REACTOR` in the per-spider settings after the `Crawler` changes in that PR, but `AsyncCrawlerProcess` implies not doing that, and `ASYNCIO_EVENT_LOOP` should be a smaller problem.

I refactored tests in `test_crawler.py` as most tests are shared between `AsyncCrawlerProcess` and `CrawlerProcess`, the latter having also additional tests related to different reactors.

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -22,6 +22,7 @@ from scrapy.exceptions import CloseSpider, DontCloseSpider, IgnoreRequest
 from scrapy.http import Request, Response
 from scrapy.utils.defer import (
     deferred_f_from_coro_f,
+    deferred_from_coro,
     maybe_deferred_to_future,
 )
 from scrapy.utils.log import failure_to_exc_info, logformatter_adapter
@@ -122,8 +123,10 @@ class ExecutionEngine:
             )
         return scheduler_cls
 
-    @deferred_f_from_coro_f
-    async def start(self, _start_request_processing=True) -> None:
+    def start(self, _start_request_processing=True) -> Deferred[None]:
+        return deferred_from_coro(self.start_async(_start_request_processing))
+
+    async def start_async(self, _start_request_processing=True) -> None:
         if self.running:
             raise RuntimeError("Engine already running")
         self.start_time = time()
@@ -392,8 +395,10 @@ class ExecutionEngine:
         finally:
             self._slot.nextcall.schedule()
 
-    @deferred_f_from_coro_f
-    async def open_spider(
+    def open_spider(self, spider: Spider, close_if_idle: bool = True) -> Deferred[None]:
+        return deferred_from_coro(self.open_spider_async(spider, close_if_idle))
+
+    async def open_spider_async(
         self,
         spider: Spider,
         close_if_idle: bool = True,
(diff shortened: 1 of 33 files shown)


scrapy/crawler.py
```diff
@@ -465,66 +490,49 @@ def crawl(
...
         executions.
         """
         while self._active:
-            await asyncio.gather(*self._active)
-
-
-class CrawlerProcess(CrawlerRunner):
-    """
-    A class to run multiple scrapy crawlers in a process simultaneously.
+            await asyncio.wait(self._active)
```
wRAR, line 532: `asyncio.gather()` has different exception handling semantics wrt `yield DeferredList()` (the former raises the first exception, the latter simply records it in the result), `asyncio.wait()` in this case does what we want, waiting until all tasks finish in some way.

scrapy/core/engine.py
```diff
@@ -392,8 +395,10 @@ def _download(
...
 
-    @deferred_f_from_coro_f
-    async def open_spider(
+    def open_spider(self, spider: Spider, close_if_idle: bool = True) -> Deferred[None]:
+        return deferred_from_coro(self.open_spider_async(spider, close_if_idle))
+
+    async def open_spider_async(
         self,
         spider: Spider,
         close_if_idle: bool = True,
```
Gallaecio, line 404: Maybe this would be a good change to apply keyword-only (*spider*) and position-only (*close_if_idle*) logic to the new async method, to make things easier for us in case we ever decide to deprecate some part of the signature.

scrapy/crawler.py
```diff
@@ -264,6 +270,28 @@ def get_spider_middleware(self, cls: type[_T]) -> _T | None:
             )
         return self._get_component(cls, self.engine.scraper.spidermw.middlewares)
 
+    async def crawl_async(self, *args: Any, **kwargs: Any) -> None:
+        if self.crawling:
+            raise RuntimeError("Crawling already taking place")
```
Gallaecio, line 275: Could this be a good chance to increase coverage? (I did not check if the coverage data is accurate or we do have tests for these but they are executed in a way that is not reflected in the coverage data)

PR #6821, opened 2025-05-22
Restructure download handler tests.

wRAR:
I need to re-read the diff after making sure all CI jobs pass.

This tries to reshuffle the test code to allow running the code with alternative download handlers. New structure:

* test_downloader_handlers.py: tests for `DownloadHandlers` and for specific non-HTTP download handlers
* test_downloader_handlers_http_base.py: only base classes for HTTP download handlers
* test_downloader_handler_twisted_http*.py: tests for HTTP10DownloadHandler, HTTP11DownloadHandler and H2DownloadHandler.

All the actually run test methods should be the same, except for 2 new but skipped methods in `test_downloader_handler_twisted_http10.TestHttp10Proxy`. The 1.1 and 2 files are now independent from each other, only depending on the base file, though some 2 classes inherit from 1.1 ones, which I don't think we can do anything about as the protocols do the same. We may need some more restructuring when we actually get alternative handlers, especially ones with 2.0, but this is a first step.


tests/test_downloader_handler_twisted_http11.py
```diff
@@ -0,0 +1,78 @@
...
+    from scrapy.core.downloader.handlers import DownloadHandlerProtocol
+
+
+DH = HTTP11DownloadHandler
+
+
+class TestHttp11(TestHttp11Base):
+    @property
+    def download_handler_cls(self) -> type[DownloadHandlerProtocol]:
+        return DH
```
Gallaecio, line 30: Looking at the amount of repetition, I wonder if a mixin class to implement the property would be a cleaner option than a global variable.

Gallaecio (approved):
:heart:

tests/test_downloader_handler_twisted_http11.py
wRAR: duh

PR #6842, opened 2025-05-29
Added download_async() to ExecutionEngine and test cases for download() and download_async()

### LINKED ISSUE #6791: Add `ExecutionEngine.download_async()`

fbenevolo:
Fixes #6791 

I added a public async_download method in ExecutionEngine class that is, the asynchronous version of download. 

I also added four test cases for each method, download and async_download:
- Test basic successful download of a request
- Test download with a redirect request
- Test download attempt when no spider is available
- Test download when the downloader raises an exception.


wRAR:
It should be `download_async()` :)

fbenevolo:
Ok, changed it

tests/test_engine.py
```diff
@@ -479,6 +481,216 @@ def kill_proc():
...
+        # Stop the downloader's slot GC loop to prevent reactor cleanup issues
+        # self.engine.downloader._slot_gc_loop = Mock()
+        # self.engine.downloader._slot_gc_loop.running = False
+
+    def tearDown(self):
+        from twisted.internet import reactor
+
+        # Clean up any pending calls
+        for c in reactor.getDelayedCalls():
+            c.cancel()
```
wRAR: Are there problems with these tests if this is not done? We have some rare failures related to something like this, but they are a more general problem and it looks like these new tests at least don't fail every time.

fbenevolo: No, I removed it and the tests were successful. I write make another commit removing it.

fbenevolo: I tried removing it, but some tests are failing. I will have to add it again. I think the problem is that the tests interfere in one another if this cancelling is not present.

wRAR:
Interesting, it indeed fails (only with the default reactor, likely because with the asyncio one we no longer use `DelayedCall`), I'll check if it's possible to fix in a less generic way, but this one is also fine.

wRAR:
I think the problem is caused by the real `self.engine.downloader` being overwritten with a mock. It looks like adding `self.engine.downloader.close()` before `self.engine.downloader = Mock()` fixes this without a need for `tearDown()` at all.

Can you please also check the diff and fix what looks like incorrectly fixed merge conflicts? Please also use `@inlineCallbacks`, not `@defer.inlineCallbacks`, this is not enforced by pre-commit but should be that way for possible future test changes.

Thanks!

wRAR:
I've also forgot to ask why have you said that it only partially solves the issue, as I think it completely solves it :)

fbenevolo:
So, I rolled back a few commits because when I removed and added tearDown again (producing two commits, one to each step), it just added some unecessary commits. The HEAD is now pointing to Simplify TestEngineDownloadAsync (commit #6842), and I will now work on those changes you ask me.

fbenevolo:
> I've also forgot to ask why have you said that it only partially solves the issue, as I think it completely solves it :)

I don't know, I think I got afraid of saying it completely resolves it lol

fbenevolo:
Hey, I fixed the merge conflicts, added the self.engine.downloader.close() (thus removing tearDown) and replaced defer.inlineCallbacks for inlineCallbacks only. But some tests still failed:
failed tests

wRAR:
Those tests should be fixed in master and they don't fail on CI so you may be doing something unusual, e.g. not testing the branch you pushed.

Gallaecio:
They passed in CI. Maybe you need the latest changes from the main branch (CI automatically merges those into your PR changes before running): `git fetch <upstream>; git merge <upstream> master`

fbenevolo:
I did that, git fetch then git rebase. Well, all checks passed so I think it is good to go.
Captura de tela de 2025-06-30 23-18-23

tests/test_engine.py
```diff
@@ -17,7 +17,8 @@
 from dataclasses import dataclass
 from logging import DEBUG
 from pathlib import Path
-from unittest.mock import Mock
+from threading import Timer
```
Gallaecio: The CI reports this as an unused import.

Gallaecio (approved):
:heart: 

Since `download` is being deprecated, I wonder if we should consider a different name for the new method, one that looks cleaner after we remove `download` from the API, e.g. something like `fetch`. (cc @kmike)

We can merge this PR and post-pone that decision to some point before the next release, through.

PR #6845, opened 2025-05-30
Use AsyncCrawlerProcess in commands.

wRAR:


wRAR:
I feel like both settings.rst and commands.rst should talk about `FORCE_CRAWLER_PROCESS` behavior but only one text should be detailed. I wrote the former one, tried to write the latter one, realized that I'm writing basically the same thing and just copied the text, but I don't think it should be like that in the final version. I'm looking for help on what to write there or at least a suggestion on which text should be verbose and which one should be terse.

scrapy/commands/crawl.py
```diff
@@ -30,17 +28,7 @@ def run(self, args: list[str], opts: argparse.Namespace) -> None:
         spname = args[0]
 
         assert self.crawler_process
-        crawl_defer = self.crawler_process.crawl(spname, **opts.spargs)
-
-        if getattr(crawl_defer, "result", None) is not None and issubclass(
```
wRAR, line 35: This was a no-op in the current code, apprrently, see #6820

scrapy/commands/crawl.py
```diff
@@ -30,17 +28,7 @@ def run(self, args: list[str], opts: argparse.Namespace) -> None:
...
-        ):
+        self.crawler_process.crawl(spname, **opts.spargs)
+        self.crawler_process.start()
+        if self.crawler_process.bootstrap_failed:
             self.exitcode = 1
-        else:
-            self.crawler_process.start()
-
-            if self.crawler_process.bootstrap_failed or (
-                hasattr(self.crawler_process, "has_exception")
```
wRAR, line 43: This was a no-op since introduction as nothing sets that.

scrapy/commands/fetch.py
```diff
@@ -18,8 +18,6 @@
 
 
 class Command(ScrapyCommand):
-    requires_project = False
```
wRAR, line 21: This is the default value

tests/test_commands.py
```diff
@@ -124,6 +125,221 @@ def setUp(self):
...
+                spider,
+                "{'TWISTED_REACTOR': 'twisted.internet.selectreactor.SelectReactor'}",
+            )
+            _, out, err = self.proc(self.name, spider)
+            assert self.ASYNC_MSG in out, out
+            assert (
+                "The installed reactor (twisted.internet.asyncioreactor.AsyncioSelectorReactor)"
+                " does not match the requested one"
+                " (twisted.internet.selectreactor.SelectReactor)"
+            ) in err, err
```
wRAR, line 323: I'll try to change this (and other cases where it's emitted) to a more specific message separately but I don't yet know if it's possible.

scrapy/cmdline.py
```diff
@@ -202,7 +203,13 @@ def execute(argv: list[str] | None = None, settings: Settings | None = None) ->
     _run_print_help(parser, cmd.process_options, args, opts)
 
     if cmd.requires_crawler_process:
-        cmd.crawler_process = CrawlerProcess(settings)
+        if (
+            settings["TWISTED_REACTOR"] == _asyncio_reactor_path
```
Gallaecio: Wouldn’t this fail if a class is used directly instead of an import path?

scrapy/cmdline.py
```diff
@@ -202,7 +203,13 @@ def execute(argv: list[str] | None = None, settings: Settings | None = None) ->
...
+            and not settings["FORCE_CRAWLER_PROCESS"]
```
Gallaecio: Similarly, maybe we need to use `getbool()` here.

Gallaecio (approved):

wRAR:
@Gallaecio do you have any feedback for the docs or do you think we can keep the duplicated text?

Gallaecio:
The docs looked very good to me, but I did not even notice the duplication :facepalm: .

If you prefer to keep it only in 1 place, which makes sense, maybe trim the settings entry and point that to the commands section.

scrapy/cmdline.py
wRAR: This isn't supported (I think it will work if `AsyncioSelectorReactor` is passed, but for any other non-string value `install_reactor()` will crash).

PR #6851, opened 2025-06-02
[doc] Add allowed_domains to example spiders

Cj-Malone:
Recently there has been a bit of drama of people on the receiving end of scrapy projects. eg 6755, 6597 and 6848. While I think most of this feedback is misunderstood or misrepresented, scrapy should continue to show it has the high ground:

- robots.txt is honoured by default
- user agents are encouraged to have your domain in them
- `scrapy genspider` includes `allowed_domains` to prevent accidental wide crawls.

This PR expands the last point to also include `allowed_domains` in spiders within the docs. This doesn’t change anything, most people still write tight spiders that don't crawl endlessly, it just slightly encourages people to use `allowed_domains` more.


Gallaecio:
I don‘t have a strong opinion about this, but I lean a bit towards not making this change, at least in its current form.

> scrapy genspider includes allowed_domains to prevent accidental wide crawls.

While this is technically true, I see it not as much as a way for Scrapy spiders to be nicer by default to websites, but as a way to prevent your spiders from wasting requests on data you do not care about. Yes, you avoid traffic to some websites, but I picture those mostly as very popular websites, e.g. social networks, for which those extra requests may be unnoticeable anyway. If you are writing a spider for example.com, `allowed_domains` will not make things better for example.com, and when it comes to requests to e.g. x.com due to “Share on X” buttons, I think it is most often in the interest of the scraper to avoid those than in the interest of x.com (i.e. those requests are probably cheaper for x.com than for you the scraper).

If we agree that there is no need to encourage the use of `allowed_domains` for the sake of writing nicer spiders, I think it is best to cover its use specifically in topics that deal with the scenarios for which it is useful, and not encourage the use of `allowed_domains` by default on any spider. In code examples, the simpler you can make the example, the better it is for readers to understand. So it is usually best to remove anything that is not needed to understand the example.

I could see the argument for encouraging the use of `allowed_domains` for the selfish reason of avoiding unnecessary requests, and that is probably why `genspider` sets that in the first place. But I am not sure if it is worth adding to unrelated code examples in the docs. In fact, for example, if we [supported nameless spiders](https://github.com/scrapy/scrapy/issues/1157), even if the general recommendation would remain to set spider names, I would remove the `name = …` line from every code example where that is not relevant.

Cj-Malone:
I'll start by saying, I don't use genspider, I don't think most people do. New people will copy paste the first example they see, probably [a](https://docs.scrapy.org/en/latest/intro/overview.html) or [b](https://docs.scrapy.org/en/latest/intro/tutorial.html#our-first-spider). People who have already written a couple spiders will just write them from scratch. I can't even find it in first steps docs. The value I see in genspider and the docs is best practices, showing people how things should be done, not necessary just how they can be done in the least lines. I do take your point, and agree, simpler examples are better, it's a fine line, but I don't think I've added this to any spider without start_urls or start_requests.

> While this is technically true, I see it not as much as a way for Scrapy spiders to be nicer by default to websites, but as a way to prevent your spiders from wasting requests on data you do not care about. Yes, you avoid traffic to some websites, but I picture those mostly as very popular websites, e.g. social networks, for which those extra requests may be unnoticeable anyway. If you are writing a spider for example.com, `allowed_domains` will not make things better for example.com, and when it comes to requests to e.g. x.com due to “Share on X” buttons, I think it is most often in the interest of the scraper to avoid those than in the interest of x.com (i.e. those requests are probably cheaper for x.com than for you the scraper).

I may be misunderstanding the most common use case for scrapy, I use it (and the docs seem to mainly be) for scraping very specific attributes from a very specific website. If I'm scraping books, and the crawler ends up on x.com, this is unexpected behaviour and unwanted. (Or the counter, if someone is scraping x.com and accentually end up on on example.com with the expectation, settings and resources to scrape the big x.com, example.com may fall over, and may be upset about it). Yes, this is clearly bad spider code and not the frameworks fault.

This PR isn't changing code, it probably wont even change behaviour, it's to add to the point of ~"bad actors go out of there way to do bad things, scrapy isn't the bad guy, it's being misused".

Cj-Malone:
Maybe I'll add, I rarely use `allowed_domains`. I usually write specific link extractors or whatever so I don't have to worry about a spider going off domain. `allowed_domains` isn't the answer to everything, but it is a very useful feature.

AdityaGadhvi (approved):

AdrianAtZyte:
I was about to give in and update your PR with the latest changes from the main, but re-reading my comment I convinced myself that we should not merge this :sweat_smile: , and instead aim to keep the doc examples as simple as possible for their specific purposes.

Cj-Malone:
I think I was going to backtrack and remove it from some examples, just leaving it in a few, but I can't remember which ones. It's been a while since I thought about it, and it seems to have been a while since people complained.

AdrianAtZyte (approved):
+0.1 from me. I still don’t feel like we *need* this, but I think the specific places where you added it are fine, and it may be even good to have it in the overview one.

PR #6852, opened 2025-06-02
Add _parallel_asyncio().

wRAR:
This is an asyncio-based replacement for `parallel()` and `parallel_async()`. Unlike those it doesn't rely on Twisted and its reactor (maybe it's possible to make `twisted.internet.task.Cooperator` work with a different event loop by passing a `scheduler` that doesn't rely on `reactor.callLater` but that would keep the Twisted dependency and `_AsyncCooperatorAdapter`).

I *think* the implementation is correct, within the stated assumptions (same assumptions apply at least partially to the older implementations AFAICS), but we'll see. Initially I implemented this with a simple `asyncio.Semaphore` + `asyncio.gather()` but was able to realize before publishing that this implementation waits until `iterable` finishes iterating first. Speaking about assumptions, I thought about making this a method of `Scraper` to slightly simplify the code (by calling `Scraper._process_spidermw_output_async()` directly) and make the assumptions more obvious, it may still be a good idea.

It's interesting that `CONCURRENT_ITEMS` is a per-response limit, as documented, not a global one, so it's not as useful in practice as I thought before.

### DIFF
--- a/tests/test_utils_defer.py
+++ b/tests/test_utils_defer.py
@@ -164,7 +164,7 @@ class TestAsyncDefTestsuite(unittest.TestCase):
         raise RuntimeError("This is expected to be raised")
 
 
-class TestAsyncCooperator(unittest.TestCase):
+class TestParallelAsync(unittest.TestCase):
     """This tests _AsyncCooperatorAdapter by testing parallel_async which is its only usage.
 
     parallel_async is called with the results of a callback (so an iterable of items, requests and None,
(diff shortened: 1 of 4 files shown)


tests/test_utils_defer.py
```diff
@@ -194,6 +194,27 @@ def callable(o: int, results: list[int]) -> Deferred[None] | None:
         results.append(o)
         return None
 
+    def callable_wrapped(
```
wRAR, line 197: The changes in this test are to check that the parallel limit is not violated.

PR #6855, opened 2025-06-03
Add AsyncioLoopingCall.

wRAR:


Gallaecio:
~~While this looks cool, I was hoping to get rid of the heartbeat, an hence `LoopingCall`, altogether. I wonder if rather than implementing an `AsyncLoopingCall`, it may be best to extract the heartbeat removal from https://github.com/scrapy/scrapy/pull/6715 into a separate PR.~~ Nevermind, I see it is used in way more places.

Gallaecio (approved):

PR #6856, opened 2025-06-03
Add type check to link.py for nofollow data member.

JoshFlenniken:
Added a type check (bool) that raises a TypeError for the nofollow parameter of Link objects. Most other data types will evaluate to True (including strings) if passed where a bool was expected, which could cause issues if any other data type is entered by mistake.

Also wrote a test for test_link.py that shows this check functioning.


wRAR:
I don't think this is needed.

PR #6858, opened 2025-06-04
Add the call_later() wrapper.

wRAR:


scrapy/core/downloader/__init__.py
```diff
@@ -218,6 +221,10 @@ def _process_queue(self, spider: Spider, slot: Slot) -> None:
                 self._process_queue(spider, slot)
                 break
 
+    def _latercall(self, spider: Spider, slot: Slot) -> None:
```
wRAR, line 224: This is a replacement for checking `slot.latercall.active()` to see if the scheduled call was already called or not.

MehrazRumman (approved):
LGTM

PR #6860, opened 2025-06-05
Resolves #6857

### LINKED ISSUE #6857: ExecuteEngine.close() assumes self.downloader has been defined
https://github.com/scrapy/scrapy/blob/8fb8d2c6b8a83181ef6564fbc9afa9cfeea2bd05/scrapy/core/engine.py#L177

However, `self.close()` is called in the `finally` block where `self.downloader` is set, so that may not be always the case.

Found it after using scrapy-poet with bad imports in a page object file consumed with `consume_modules()`.

Tristan296:
Resolves #6857


wRAR:
Obligatory comment: I wonder if it's possible to add a test for this.

Tristan296:
> Obligatory comment: I wonder if it's possible to add a test for this.

Yeah might be I'll have a look into it

MehrazRumman (commented):
Is it necessary to create a `test_engine_close.py` file? I think some tests are already written in the `test_engine.py` file.

Tristan296:
I moved the test into the test_engine.py file

PR #6862, opened 2025-06-05
fixing issue #6857

MehrazRumman:
Trying to solve #6857

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -29,7 +29,6 @@ from twisted.trial import unittest
 from twisted.web import server, static, util
 
 from scrapy import signals
-from scrapy.core.downloader import Downloader
 from scrapy.core.engine import ExecutionEngine, _Slot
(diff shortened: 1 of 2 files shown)


MehrazRumman:
When I ran this particular test locally, it worked perfectly. I don't know why it is failing after pushing to GitHub.

tests/test_engine.py
```diff
@@ -433,6 +433,10 @@ def test_close_downloader(self):
         e = ExecutionEngine(get_crawler(MySpider), lambda _: None)
         yield e.close()
 
+        if hasattr(e, "downloader"):
```
wRAR: I think this needs to be a separate test method. And I think it should be possible to replicate an early failure (e.g. by passing a `DOWNLOADER` value that cannot be used to create an instance) instead of deleting e.downloader directly.

MehrazRumman: Thanks. I am doing the changes.

PR #6866, opened 2025-06-06
More docs for the is_asyncio_reactor_installed() behavior change.

wRAR:


scrapy/utils/test.py
```diff
@@ -117,6 +117,11 @@ def get_reactor_settings() -> dict[str, Any]:
     settings, so tests that run the crawler in the current process may need to
     pass a correct ``"TWISTED_REACTOR"`` setting value when creating it.
     """
+    if not is_reactor_installed():
+        raise RuntimeError(
```
wRAR, line 121: Writing a test for this would need running a standalone script, but I checked this with `scrapinghub-entrypoint-scrapy` before https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/pull/94 and it produces the correct exception.

PR #6867, opened 2025-06-06
Don't try to close ExecutionEngine.downloader when it doesn't exist.

### LINKED ISSUE #6857: ExecuteEngine.close() assumes self.downloader has been defined

wRAR:
Fixes #6857, closes #6860, closes #6862

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -22,6 +22,7 @@ from unittest.mock import Mock
 from urllib.parse import urlparse
 
 import attr
+import pytest
 from itemadapter import ItemAdapter
 from pydispatch import dispatcher
 from twisted.internet import defer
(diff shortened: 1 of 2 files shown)


tests/test_engine.py
```diff
@@ -433,6 +434,19 @@ def test_close_downloader(self):
...
 
+    def test_close_without_downloader(self):
+        class CustomException(Exception):
+            pass
+
+        class BadDownloader:
+            def __init__(self, crawler):
+                raise CustomException
+
+        with pytest.raises(CustomException):
```
wRAR, line 445: The unmodified code raises `AttributeError`

PR #6873, opened 2025-06-06
Reduce deps on unittest, unify inlineCallbacks imports in tests.

wRAR:
1. `@defer.inlineCallbacks` -> `@inlineCallbacks` (most of the PR changes)
2. Remaining `self.assertFailure()` -> `pytest.raises()`
3. Remaining `raise unittest.SkipTest()` -> `pytest.skip()`
4. One or two tests used `unittest` instead of `twisted.trial.unittest` so they could be migrated to pytest directly.

### DIFF
--- a/tests/test_utils_log.py
+++ b/tests/test_utils_log.py
@@ -4,7 +4,6 @@ import json
 import logging
 import re
 import sys
-import unittest
 from io import StringIO
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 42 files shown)


tests/test_utils_log.py
```diff
@@ -100,21 +99,19 @@ def test_filtered_out_level(self):
         assert self.crawler.stats.get_value("log_count/INFO") is None
 
 
-class StreamLoggerTest(unittest.TestCase):
-    def setUp(self):
```
wRAR, line 104: pytest overrides `sys.stdout` after the fixtures etc. run, so merging the code into the test method was the easiest way to reimplement this

PR #6877, opened 2025-06-08
Replace assert statements with ValueError

### LINKED ISSUE #6876: Non-typing assert in getdictorlist()
I noticed that the implementation of `getdictorlist()` currently has an `assert` after `json.loads()`. A `ValueError` should instead be raised of the output of `json.loads()` is not either a `dict` or a `list`.

DB-db-dron:
Fix #6876

### DIFF
--- a/scrapy/settings/__init__.py
+++ b/scrapy/settings/__init__.py
@@ -282,13 +282,15 @@ class BaseSettings(MutableMapping[_SettingsKeyT, Any]):
         if isinstance(value, str):
             try:
                 value_loaded = json.loads(value)
-                assert isinstance(value_loaded, (dict, list))
+                if not isinstance(value_loaded, (dict, list)):
+                    raise ValueError(f"The value for setting must be a dict or list, got {type(value_loaded).__name__}")
                 return value_loaded
             except ValueError:
                 return value.split(",")
         if isinstance(value, tuple):
             return list(value)
-        assert isinstance(value, (dict, list))
+        if not isinstance(value, (dict, list)):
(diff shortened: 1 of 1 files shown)


DB-db-dron:
@Gallaecio Unable to understand why these checks are failing. Can you please look into them

wRAR:
@DB-db-dron #6878

DB-db-dron:
@wRAR one check is still failing. Do I need to add tests for the changes i made?

DB-db-dron:
I have added the test, but code coverage is still failing, don't understand why.

wRAR:
The test is just for one of two places that were changed. The second place handles JSON values, so I think you can write a test that uses a JSON value that doesn't deserialize into a dict or list.

DB-db-dron:
@wRAR I have added the tests. The checks have not run after that

scrapy/settings/__init__.py
```diff
@@ -282,13 +282,19 @@ def getdictorlist(
         if isinstance(value, str):
             try:
                 value_loaded = json.loads(value)
-                assert isinstance(value_loaded, (dict, list))
+                if not isinstance(value_loaded, (dict, list)):
+                    raise ValueError(
+                        f"The value for setting must be a dict or list, got {type(value_loaded).__name__}"
```
Gallaecio, line 287: What about something like this in both messages?

```suggestion
                        f"The value of the {name} setting must be a dict or list, got {value!r}"
```

This way, the setting name and wrong value are shown, which should make it easier to spot where the issue comes from, specially since the traceback indicates where the setting is being read, not where it was written, which was the actual issue. The `!r` should be enough to make the type of the value clear in most cases.

PR #6883, opened 2025-06-09
Replace most of the @inlineCallbacks test helpers.

wRAR:


tests/test_feedexport.py
```diff
@@ -89,24 +91,25 @@ def mock_google_cloud_storage() -> tuple[Any, Any, Any]:
     return (client_mock, bucket_mock, blob_mock)
 
 
+# TODO: replace self.mktemp() and drop the unittest.TestCase base
```
wRAR, line 94: Will be done separately.

PR #6885, opened 2025-06-11
Remove ProcessTest and SiteTest.

wRAR:
It looks like an async process spawn mechanism (ProcessTest) was needed to not mutually block the web server (SiteTest) also running in the test process. I've moved the latter into mockserver and switched the commands that need HTTP to the normal popen code used by tests for other commands.


wRAR:
As a coincidence I've found that pywin32 was only needed for `TestProc`.

PR #6888, opened 2025-06-13
Remove the chdir fixture, re-enable fancy pytest asserts

wRAR:


wRAR:
There were some failures on Windows in `tests/test_feedexport.py::TestFileFeedStorage` but I can't reproduce them locally or after restarting the jobs. I compared one good job and one bad job and there is some difference in runners, disk letters and Python version but that doesn't really help. We may need to investigate this later if it starts happening more reliably.

PR #6889, opened 2025-06-14
Refactor test functions that were returning a Deferred directly.

wRAR:


Gallaecio:
Hmm… If those Windows issues in pinned envs can be solved by upgrading a minimum version of some dep, it might be worth it.

PR #6890, opened 2025-06-14
 Fixing issue #6876

MehrazRumman:


MehrazRumman:
@wRAR  It is ready for review. I don't know why pylint is failing. I run this locally it doesn't show error .

wRAR:
> I don't why pylint is failing. I run this locally it doesn't show error .

You may be running it incorrectly then.

MehrazRumman:
> > I don't why pylint is failing. I run this locally it doesn't show error .
> 
> You may be running it incorrectly then.

Yes, I had same thought !

wRAR:
If you are running it via `tox -e pylint` it shouldn't behave differently.

MehrazRumman:
Thanks for the instruction @wRAR !

PR #6897, opened 2025-06-21
Fix FileFeedStorage handling of Windows paths without file:// scheme

### LINKED ISSUE #6894: `FileFeedStorage` ignores the drive letter on Windows
```
>>> from scrapy.extensions.feedexport import FileFeedStorage
>>> s = FileFeedStorage("C:/foo")
>>> s.path
'\\foo'
>>> s = FileFeedStorage("D:/foo")
>>> s.path
'\\foo'
```

This, I believe, causes the files to be written to the drive from the spider work directory instead of the specified one. Thanks to GitHub CI recently changing the workdir to something on D: this is now visible in tests.

The cause of this seems to be `w3lib.url.file_uri_to_path()`, but I don't think calling it on a normal path instead of an URI is correct. OTOH this code was added in 2010 and so far it looks like it should have been working in the same way back then.

thalissonvs:
## Description

This PR addresses an issue in the `FileFeedStorage` class where passing a Windows-style path without the `file://` URI scheme causes incorrect path resolution. Specifically, using a path like `C:/foo` was incorrectly processed by `w3lib.url.file_uri_to_path()`, which expects a proper URI. This led to malformed paths such as `\\foo`, causing files to be written relative to the spider's working directory instead of the intended absolute path.

The problematic code was:
```python
self.path = file_uri_to_path(uri)
```
which always called file_uri_to_path() regardless of whether uri was a valid URI or just a local path.

## Solution

The fix changes the initialization logic to only convert the path using file_uri_to_path() if the input string starts with the file:// scheme. Otherwise, the path is preserved as-is.

## Testing

- Added a new test test_preserves_windows_path_without_file_scheme that verifies Windows-style paths are preserved correctly when passed without the file:// scheme.
- Existing tests that use tmp_path cover writing and storing files, including relative and URI paths.
- The new test is sufficient to cover the issue because the root problem was path parsing, not file writing.

Fixes #6894

### DIFF
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -185,7 +185,9 @@ class StdoutFeedStorage:
 @implementer(IFeedStorage)
 class FileFeedStorage:
     def __init__(self, uri: str, *, feed_options: dict[str, Any] | None = None):
-        self.path: str = file_uri_to_path(uri)
+        self.path: str = (
+            file_uri_to_path(uri) if uri.startswith("file://") else str(Path(uri))
+        )
         feed_options = feed_options or {}
         self.write_mode: OpenBinaryMode = (
             "wb" if feed_options.get("overwrite", False) else "ab"
--- a/scrapy/extensions/feedexport.py
+++ b/scrapy/extensions/feedexport.py
@@ -185,9 +185,7 @@ class StdoutFeedStorage:
 @implementer(IFeedStorage)
 class FileFeedStorage:
     def __init__(self, uri: str, *, feed_options: dict[str, Any] | None = None):
-        self.path: str = (
-            file_uri_to_path(uri) if uri.startswith("file://") else str(Path(uri))
-        )
+        self.path: str = file_uri_to_path(uri) if uri.startswith("file://") else uri
         feed_options = feed_options or {}
         self.write_mode: OpenBinaryMode = (
             "wb" if feed_options.get("overwrite", False) else "ab"
--- a/tests/test_feedexport.py
+++ b/tests/test_feedexport.py
@@ -155,11 +155,10 @@ class TestFileFeedStorage:
         finally:
             path.unlink()
 
-    def test_preserves_windows_path_without_file_scheme(self, tmp_path):
+    def test_preserves_windows_path_without_file_scheme(self):
         path = r"C:\Users\user\Desktop\test.txt"
-        expected_path = "C:\\Users\\user\\Desktop\\test.txt"
         storage = FileFeedStorage(path)
-        assert storage.path == expected_path
+        assert storage.path == path
 
 
 class TestFTPFeedStorage(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


tests/test_feedexport.py
```diff
@@ -155,6 +155,12 @@ def _assert_stores(
         finally:
             path.unlink()
 
+    def test_preserves_windows_path_without_file_scheme(self, tmp_path):
```
wRAR: The fixture is not used and can be dropped.

tests/test_feedexport.py
```diff
@@ -155,6 +155,12 @@ def _assert_stores(
...
+        path = r"C:\Users\user\Desktop\test.txt"
+        expected_path = "C:\\Users\\user\\Desktop\\test.txt"
```
wRAR: These are the same var so you can just use one.

scrapy/extensions/feedexport.py
```diff
@@ -185,7 +185,9 @@ def store(self, file: IO[bytes]) -> Deferred[None] | None:
 @implementer(IFeedStorage)
 class FileFeedStorage:
     def __init__(self, uri: str, *, feed_options: dict[str, Any] | None = None):
-        self.path: str = file_uri_to_path(uri)
+        self.path: str = (
+            file_uri_to_path(uri) if uri.startswith("file://") else str(Path(uri))
```
wRAR: Not sure what does `str(Path(uri))` achieve, does it do any normalization or is it a no-op?

wRAR:
(pinned env failures are due to the pytest changes, #6893 should fix that)

scrapy/extensions/feedexport.py
thalissonvs: str(Path(uri)) will convert the raw string to a string with escaped backslashes. But this is just a string representation, I can remove that

thalissonvs: Yeah, str(Path(uri)) does a bit of normalization, like keeping the drive letter on Windows and resolving relative parts. For example, if you have a path like "C:/foo/../bar", str(Path(uri)) will turn it into "C:\\bar".  I used it mainly to make sure we’re working with a proper path object before converting back to string, instead of passing raw input around

thalissonvs: `str(Path(uri))` performs path normalization that's useful in this context, for example, ensures native path separators (\ on Windows, / on Unix). If you prefer to keep it minimal, we could just use uri directly since the main fix is avoiding file_uri_to_path() on non-URI paths, please let me know

wRAR: I think it's fine to keep it minimal, as `self.path` is already passed to `Path()` anyway in `open()` (I think `self.path` itself is still a string because changing it would break compatibility with `FileFeedStorage` subclasses).

thalissonvs: Got it

wRAR (approved):
Thanks!

Gallaecio:
Forcing a re-run that includes #6893…

Gallaecio:
Thanks!

PR #6899, opened 2025-06-21
Fix typing of dynamic `request` attribute on `Failure` with a cast subclass

thalissonvs:
This PR addresses a longstanding `TODO` in the `call_spider_async` method regarding the typing of the dynamically added `request` attribute on `twisted.python.failure.Failure` objects.

Since `Failure` does not originally define the `request` attribute, adding it dynamically causes static type checkers (e.g., mypy) to raise errors.

To resolve this without changing runtime behavior or introducing new `Failure` instances, this PR introduces a lightweight subclass `FailureWithRequest` used solely for static typing purposes via `cast()`. This approach:

- Provides type safety and clarity for static analysis tools  
- Avoids creating new `Failure` instances, preserving error context and traceback  
- Keeps runtime behavior unchanged  
- Offers a balanced solution (middle ground) between ignoring type checks and a full refactor of `Failure` usage

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -251,6 +251,8 @@ class Scraper:
             warn_on_generator_with_return_value(self.crawler.spider, callback)
             output = callback(result, **result.request.cb_kwargs)
         else:  # result is a Failure
+            # this cast allows static type checkers to recognize the dynamically
+            # added `request` attribute without altering the runtime behavior.
             result = cast(FailureWithRequest, result)
             result.request = request
             if not request.errback:
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -7,7 +7,7 @@ import logging
 import warnings
 from collections import deque
 from collections.abc import AsyncIterator
-from typing import TYPE_CHECKING, Any, TypeVar, Union
+from typing import TYPE_CHECKING, Any, TypeVar, Union, cast
 
 from twisted.internet.defer import Deferred, inlineCallbacks, maybeDeferred
 from twisted.python.failure import Failure
@@ -50,6 +50,10 @@ logger = logging.getLogger(__name__)
 
 _T = TypeVar("_T")
 QueueTuple = tuple[Union[Response, Failure], Request, Deferred[None]]
+
+
+class FailureWithRequest(Failure):
+    request: Request
 
 
 class Slot:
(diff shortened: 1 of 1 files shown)


wRAR:
To be honest, I think this is strictly worse than an ignore comment.

thalissonvs:
> To be honest, I think this is strictly worse than an ignore comment.

If you have some time, could you please explain why? I understand that using cast might cause issues, but in this case we are within a well-defined scope

If you have any other suggestions on how to solve this problem, I’d be happy to implement them. Thanks for your attention

thalissonvs:
Hi! Thanks for the feedback. Just to clarify, would using a Protocol to define the expected request attribute be a valid approach here? I believe it could express the dynamic addition of that attribute more safely without creating a concrete subclass

I’d appreciate your thoughts on this. Thanks!

Gallaecio:
I protocol does not sound bad to me in principle, some kind of `ScrapyFailure` type that is not a real type but only used for type hints.

wRAR:
> Thanks for the feedback. Just to clarify, would using a Protocol to define the expected request attribute be a valid approach here?

Not sure, as `Failure`, which is the actual type of the var, would still not conform to such protocol, and casting a var to some type that it is not looks wrong to me. But it's possible that this, or something like this, is common practice which I haven't seen yet.

To be honest, I don't remember if I had any specific ideas when I noticed this initially and wrote that TODO. I guess ideally the failure var should be a subclass of `Failure` but I doubt that this is possible.

thalissonvs:
I think using a Protocol is a good way to handle this. When we cast something to a protocol, we're not saying the object is exactly that type, we're just telling the type checker to treat it as something that follows that interface. It’s a nice fit for duck typing when we can’t change the original class.

In this case, we’re not pretending result is a specific type, we’re just saying “let’s treat this as something that has a `.request` attribute"

There’s also a great [example](https://mypy.readthedocs.io/en/stable/protocols.html#using-isinstance-with-protocols) in the mypy docs:

```python
from typing import Protocol, runtime_checkable

@runtime_checkable
class Portable(Protocol):
    handles: int

class Mug:
    def __init__(self) -> None:
        self.handles = 1

def use(handles: int) -> None: ...

mug = Mug()
if isinstance(mug, Portable):  # Works at runtime!
    use(mug.handles)
```

That really highlights what a protocol represents: any object with the right attributes and methods can match the protocol. So in our case, using a protocol like `HasRequest` to represent “something that has a .request” isn’t just tricking the type checker

Let me know what you think, I'd be happy to adjust if this approach makes sense :)

PR #6902, opened 2025-06-22
Same origin allowed domains

### LINKED ISSUE #522: should spider.allowed_domains match same-origin policy?
I believe it would be nice to obey the Web Origin Concept's understanding of same-origin (https://tools.ietf.org/html/rfc6454#section-3.2.1) in `allowed_domains`.
So I'll throw this out here for discussion.

Currently it is impossible to specify, using only `allowed_domains`, when subdomains of a domain should be considered 'off-site', or that actually a different protocol/scheme (or port) should be considered a different origin.

My proposal would be to include or accept actual schemes in `allowed_domains`, such as `https://example.com` to specifiy `http://example.com` or `ftp://example.com` as offsite.

More problematic might be specifying where subdomains of a domain should be offsite, where infact `//github.com` shouldn't match `//someuser.github.com` by default, but cookies and web origins don't use wildcards (I believe) for specifying something as `//*.example.com`. This would be where DNS wildcards help,
but then again, according to DNS wildcards, this shouldn't match the parent domain, so there's a catch here on how to specify `http://[everything.]example.com` while excluding `https` or other schemes.

Such a change would bring scrapy's understanding of "same origin" into line with Cookie Policy, DNS policy and the Web Origin Concept.
See the following RFCs:
Cookies:
[RFC 2109](https://tools.ietf.org/html/rfc2109#section-2) HTTP State Management Mechanism (hist.)
[RFC 2965](https://tools.ietf.org/html/rfc2965#section-1) HTTP State Management Mechanism (hist.)
[RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.1.3) HTTP State Management Mechanism (prop.)
Same-origin:
**[RFC 6454](https://tools.ietf.org/html/rfc6454#section-3.2.1)** The Web Origin Concept
DNS:
[RFC 1034](http://tools.ietf.org/html/rfc1034#section-4.3.3) Domain Names
**[RFC 4592](http://tools.ietf.org/html/rfc4592#section-2.2.1)** The Role of Wildcards in DNS

ljluestc:
Closes #522 - Enhances `spider.allowed_domains` to support the Web Origin Concept (RFC 6454) by considering scheme, host, and port, and restricting subdomains unless explicitly allowed.

## Description
This PR updates Scrapy’s `spider.allowed_domains` to align with the same-origin policy defined in RFC 6454, while preserving backward compatibility with existing spider behavior. The changes allow specifying full URLs (e.g., `https://example.com`) in `allowed_domains` to restrict crawling to specific schemes and ports, and treat subdomains (e.g., `sub.example.com`) as distinct origins unless explicitly included (e.g., via `*.example.com`). A new `strict_origins` spider attribute enables strict matching of scheme, host, and port, aligning with cookie (RFC 6265) and DNS wildcard (RFC 4592) policies.

## Changes
- **Added** `strict_origins` attribute to `scrapy.spiders.Spider` to enable same-origin policy compliance.
- **Modified** `Spider._normalize_allowed_domains` to parse `allowed_domains` entries into scheme, host, port, and wildcard components using `urllib.parse.urlparse`.
- **Updated** `Spider.is_url_allowed` to check scheme and port when `strict_origins=True`, and restrict subdomains unless explicitly allowed via wildcard (e.g., `*.example.com`).
- **Modified** `scrapy.utils.url.is_url_allowed` to delegate to `Spider.is_url_allowed` for consistency.
- **Updated** `OffsiteMiddleware` in `scrapy/spidermiddlewares/offsite.py` to use normalized domains and origins, ensuring requests are filtered based on `strict_origins` settings.
- **Added** comprehensive tests in `tests/test_spider.py` to verify:
  - Default behavior (lenient domain matching, including subdomains).
  - Strict origin matching (scheme, host, port).
  - Wildcard subdomain handling (with and without parent domain).
  - Edge cases (empty `allowed_domains`, invalid entries).

## Impact
- Aligns Scrapy with RFC 6454 (Web Origin Concept), RFC 6265 (Cookies), and RFC 4592 (DNS Wildcards).
- Improves precision in domain filtering for web scraping, reducing unintended off-site requests.
- Maintains backward compatibility for existing spiders by defaulting `strict_origins=False`.
- Enhances security by allowing fine-grained control over allowed origins.


Gallaecio:
Maybe instead of a boolean this could be approached by deprecating `allowed_domains` and adding `allowed_origins`. Or maybe we could allow both, and make `allowed_domains` keep the current behavior if used, if it makes sense to support simpler scenarios, while having the new `allowed_origins` attribute allow for more flexibility/strictness.

wRAR:

PR #6904, opened 2025-06-22
docs: improve ScrapyCommand documentation for issue #6844

### LINKED ISSUE #6844: Better docs for creating custom commands
[Current docs](https://docs.scrapy.org/en/latest/topics/commands.html#custom-project-commands) only say "look at the existing ones", we could do better than that, at least by listing attributes you can access and override and functions you can override. Even just improving docstrings for `scrapy.commands.ScrapyCommand` and adding an `autoclass` for it to the doc would already be useful.

alchemypls:
# Improve ScrapyCommand documentation (Fixes #6844)

This PR addresses issue #6844 by significantly improving the documentation for creating custom Scrapy commands. Previously, the docs only said "look at the existing ones".

## Changes made:

1. **Added docstring to `ScrapyCommand` class** - Simple, PEP 257 compliant docstring
2. **Enhanced documentation in `docs/topics/commands.rst`**:
   - Improved "Custom project commands" section opening
   - Added comprehensive "Creating custom commands" section
   - Listed all attributes you can set (`requires_project`, `requires_crawler_process`, `default_settings`, `exitcode`)
   - Listed all methods you must override (`syntax`, `short_desc`, `run`)
   - Listed all methods you can override (`long_desc`, `help`, `add_options`, `process_options`)
   - Added example custom command implementation
   - Added reference to real examples in `scrapy/commands` directory
3. **Added autoclass directive** - `ScrapyCommand` class is now properly documented via Sphinx autodoc

Screenshot from 2025-06-22 01-44-10

This provides much better developer experience when creating custom Scrapy commands.

Just let me know if you want any changes. Still learning the ropes in OSC.


Gallaecio (approved):
Looks great.

I do wonder if the new COMMANDS_MODULE section should be put withing a "Settings" section, to make things a bit more clear. But I am fine with the proposal as it is, it is a big improvement.

wRAR:
I need to recheck all of this after recent command-related changes but I remember from the first surface review of this PR that there is a mistake in it: you aren't required to override `syntax()` (I imagine that a command that doesn't take any args won't override it).

docs/topics/commands.rst
```diff
@@ -629,25 +629,74 @@ Custom project commands
...
 
-A module to use for looking up custom Scrapy commands. This is used to add custom
-commands for your Scrapy project.
+**Attributes you can set:**
 
-Example:
+* ``requires_project`` (bool): If True, the command requires a Scrapy project to be present (default: False)
+* ``requires_crawler_process`` (bool): If True, the command requires a crawler process to be available (default: True)
+* ``default_settings`` (dict): Default settings to use for this command instead of global defaults (default: {})
+* ``exitcode`` (int): Exit code to return when command completes (default: 0)
```
wRAR: Why do we have this attr :-/

docs/topics/commands.rst
```diff
@@ -629,25 +629,74 @@ Custom project commands
...
+**Methods you must override:**
+
+* :meth:`~scrapy.commands.ScrapyCommand.syntax`: Return command syntax (preferably one-line, without command name)
+* :meth:`~scrapy.commands.ScrapyCommand.short_desc`: Return a short description of the command
+* :meth:`~scrapy.commands.ScrapyCommand.run`: Main entry point for command execution (must implement)
+
+**Methods you can override:**
+
+* :meth:`~scrapy.commands.ScrapyCommand.long_desc`: Return a detailed description (can contain newlines)
+* :meth:`~scrapy.commands.ScrapyCommand.help`: Return extensive help text (can contain newlines)
```
wRAR: #7626

wRAR:
I've refreshed this. I think it's a big improvement but it may benefit from a separate second pass. E.g. the autodocs for `ScrapyCommand` are pretty bare, but when we fill them up I wonder if the lists of attrs and methods added here should still stay as a nice reference.

I was on the fence with listing `process_options()`, no built-in commands override it and it's not needed at all if you don't want to update settings before `crawler_process` is instantiated as you can do everything in `run()`. But it's public and it can be useful in the aforementioned use cases.

AdrianAtZyte (approved):

PR #6910, opened 2025-06-25
Fix setting ExecutionEngine._slot to None on close.

wRAR:
When the attribute was renamed this code wasn't changed.

(the whole method needs refactoring but that's for later)

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -530,10 +530,16 @@ class ExecutionEngine:
             )
         )
 
-        dfd.addBoth(lambda _: setattr(self, "slot", None))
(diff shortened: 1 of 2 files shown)


Gallaecio (approved):

wRAR:
Ha!

wRAR:
The failure was when `_start_request_processing()` gets control after `await self._process_start_next()` and the engine was being closed during that. Strictly speaking it's a weird state when start() and stop() are running in parallel, and because of that `_start_request_processing_dfd` wasn't properly cancelled (as it was `None` when that part of `stop()` was executed). Not sure if there is a good way to not do that.

Gallaecio (approved):

PR #6911, opened 2025-06-25
Narrow down TestEngine::test_short_timeout() expectations.

### LINKED ISSUE #6908: TestEngine::test_short_timeout() failure
Introduced by #6900, happens only sometimes: https://github.com/scrapy/scrapy/actions/runs/15831375533/job/44624609874

There is some unhandled `CancelledError` in the test spider log.

wRAR:
Fixes #6908

As it doesn't look like we can avoid the unhandled CancelledError, change the expectations to exceptions that could be raised when `self._slot` is unexpectedly None.

The test was added in #5440


wRAR:
Ok, this is again stopping and starting in parallel, we may need more checks and/or some locking.

PR #6914, opened 2025-06-26
Consolidate CI workflows into a single unified configuration

### LINKED ISSUE #6794: Use a single workflow
I wonder if it would make sense to merge all 4 testing workflows (static checks and OS-specific tests) into a single GitHub Actions workflows, so that we can have a single **Tests** badge in the README.

thalissonvs:
Closes #6794

## Summary

This PR consolidates the three separate test workflow files (`tests-macos.yml`, `tests-ubuntu.yml`, `tests-windows.yml`) into a single unified workflow file (`tests.yml`).

## Matrix Strategy
- Base matrix covers common configurations (OS × Python version)
- Additional configurations added via `include` for special test environments
- All existing test combinations preserved

## Platform-Specific Commands
- **macOS**: Runs `tox -e py` 
- **Ubuntu/Windows**: Runs full `tox`


thalissonvs:
I have a question. Should I keep 'Checks' in a separate job, or is it better to include everything under 'Tests'? For example, we can have something like this:

```yaml

jobs:
  tests:
    ...
  checks:
    ...
  pre-commit:
    ...
```

or everything inside tests, where we could take advantage of the matrix and just add new cases in the include section:

```yaml
jobs:
  tests:
    ... # including checks and a job for pre-commit
```

Issue #6917, opened 2025-06-26
Deprecate CONCURRENT_REQUESTS_PER_IP


wRAR:
`CONCURRENT_REQUESTS_PER_IP` isn't used by default, isn't supported by `DownloaderAwarePriorityQueue`,  isn't compatible with `CONCURRENT_REQUESTS_PER_DOMAIN` and is less useful than that one for most users.

Not sure what can we suggest to people using it though.

Gallaecio:
> Not sure what can we suggest to people using it though.

I think `CONCURRENT_REQUESTS_PER_DOMAIN` is the natural alternative to suggest, specially for someone that might not have known what they were doing when they started using `CONCURRENT_REQUESTS_PER_IP`.

wRAR:
I assume we also want to remove references to it from the Autothrottle and broad crawl docs (I don't remember if we keep or remove deprecated settings from settings.rst).

Gallaecio:
I would remove anything deprecated from the docs, period (besides from news.rst). Someone reading the docs should not read how to do something in a deprecated way, not even if we mark it as such.

Laerte:
@wRAR We want to emit an warning when someone use this setting? So we can remove in future?

PR #6921, opened 2025-06-28
Deprecate CONCURRENT_REQUESTS_PER_IP setting

Laerte:
close: #6917


tests/test_scheduler.py
```diff
@@ -389,7 +390,27 @@ def _incompatible(self):
         scheduler.open(spider)
 
     def test_incompatibility(self):
-        with pytest.raises(
-            ValueError, match="does not support CONCURRENT_REQUESTS_PER_IP"
-        ):
-            self._incompatible()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
```
Laerte, line 394: I thought add the deprecate test here but having a separate test seems better, I just filtering the warning to not show on the test logs.

docs/topics/settings.rst
```diff
@@ -1765,8 +1745,7 @@ Type of priority queue used by the scheduler. Another available type is
 ``scrapy.pqueues.DownloaderAwarePriorityQueue``.
 ``scrapy.pqueues.DownloaderAwarePriorityQueue`` works better than
 ``scrapy.pqueues.ScrapyPriorityQueue`` when you crawl many different
-domains in parallel. But currently ``scrapy.pqueues.DownloaderAwarePriorityQueue``
-does not work together with :setting:`CONCURRENT_REQUESTS_PER_IP`.
```
Gallaecio, line 1769: I think the lack of support for this setting was the only reason we did not make `DownloaderAwarePriorityQueue` the default priority queue back in the day (@wRAR @kmike Do you recall?). Nothing needs to happen in this PR about it, but I wonder if we should merge `DownloaderAwarePriorityQueue` into `ScrapyPriorityQueue` now or, if there is a reason to keep a simpler `ScrapyPriorityQueue` around, if we should make `DownloaderAwarePriorityQueue` default, at least with `startproject`.

scrapy/settings/default_settings.py
```diff
@@ -45,7 +45,6 @@
 
 CONCURRENT_REQUESTS = 16
 CONCURRENT_REQUESTS_PER_DOMAIN = 8
-CONCURRENT_REQUESTS_PER_IP = 0
```
Gallaecio, line 48: I think we should keep it here during the deprecation period.

We could use a `__getattr__` module-level function to log a deprecation warning upon import. I wonder if that means we also need to define `__all__` with the whole list of settings defined here, but to be honest, it may be better to break any `from scrapy.settings.default_settings import *`, seems like a bad practice and I cannot picture a valid reason to use that.

scrapy/core/scheduler.py
```diff
@@ -296,6 +296,15 @@ def __init__(
         self._smqclass: type[BaseQueue] | None = self._get_start_queue_cls(
             crawler, "MEMORY"
         )
+        if (
+            self.crawler
+            and self.crawler.settings.getint("CONCURRENT_REQUESTS_PER_IP") != 0
+        ):
+            warn(
+                "CONCURRENT_REQUESTS_PER_IP setting is deprecated, use CONCURRENT_REQUESTS_PER_DOMAIN instead.",
```
Gallaecio:
```suggestion
                "The CONCURRENT_REQUESTS_PER_IP setting is deprecated, use CONCURRENT_REQUESTS_PER_DOMAIN instead.",
```

scrapy/core/scheduler.py
```diff
@@ -296,6 +296,15 @@ def __init__(
         self._smqclass: type[BaseQueue] | None = self._get_start_queue_cls(
             crawler, "MEMORY"
         )
+        if (
+            self.crawler
+            and self.crawler.settings.getint("CONCURRENT_REQUESTS_PER_IP") != 0
```
Gallaecio: First off, I think this is OK.

However, alternatively, it may be worth considering to implement a deprecated setting reporting logic on the `get` method of `BaseSettings` that warns about a deprecated setting upon reading it (and in `set` upon writing it?) if the setting priority value is not 0 (0 means its value comes from `default_settings.py`, any other value means it was set by the user).

It has the advantage that it will also warn about the setting when used in third-party and user-defined components that may be currently relying on the setting.

On the other hand, if a third-party or user-defined component is relying on that setting and for some reason wants to keep relying on that setting (which seems weird for this specific setting, but may make sense for other settings we may deprecate in the future), there would be no way to silence this warning. Which makes me think this approach may also warrant adding a setting to allow allow-listing settings to prevent such warnings for them.

(edit: removed another alternative suggestion that does not really work provided we keep the current `default_settings.py` definition, which I think we should keep)

scrapy/settings/default_settings.py
wRAR: > We could use a __getattr__ module-level function to log a deprecation warning upon import.

Why? I don't think editing `default_settings.py` is supported.

wRAR: Ah, you mean getting the value from importing it in the user code via `from scrapy.settings.default_settings import *`. I'm fine with breaking that.

docs/topics/settings.rst
wRAR: I remember basically nothing about `DownloaderAwarePriorityQueue` merging 😅

wRAR: The initial message of #3520 indeed says that.

If that's still the only reason I'm fine with changing the default, not sure about removing `ScrapyPriorityQueue` while `CONCURRENT_REQUESTS_PER_IP` is still supported.

scrapy/settings/default_settings.py
Gallaecio: > I'm fine with breaking that.

While I am OK with breaking it for this specific setting, I would not make it a general rule. For example, checking `scrapy.settings.default_settings.TWISTED_REACTOR` seems like a nice way to check what the default reactor is in the installed version of Scrapy without specifically checking for the Scrapy version.

docs/topics/settings.rst
Gallaecio: > not sure about removing ScrapyPriorityQueue while CONCURRENT_REQUESTS_PER_IP is still supported.

Good point. We could deprecate it as well but keep it, and if the new default queue causes issues we cannot easily address during the deprecation period, we can always course-correct.

scrapy/settings/default_settings.py
wRAR: https://github.com/search?q=%22from+scrapy.settings.default_settings%22&type=code

https://github.com/search?q=%22from+scrapy.settings+import+default_settings%22&type=code

docs/topics/settings.rst
Laerte: Should I push this change or this will be done on separate PR?

scrapy/settings/default_settings.py
Laerte: I added `__all__` and logic to emit deprecation message when someone try to import CONCURRENT_REQUESTS_PER_IP directly.

scrapy/core/scheduler.py
Laerte: Refactored, lmk what you think.

docs/topics/settings.rst
Gallaecio: Separate.

Laerte: Thanks, created an issue for it: https://github.com/scrapy/scrapy/issues/6924

scrapy/settings/__init__.py
```diff
@@ -147,6 +149,13 @@ def get(self, name: _SettingsKeyT, default: Any = None) -> Any:
         :param default: the value to return if no setting is found
         :type default: object
         """
+        if name == "CONCURRENT_REQUESTS_PER_IP" and (self[name] or default) != 0:
```
Gallaecio: > `(self[name] or default) != 0`

Won’t this always be `True`? (unless the `default` parameter itself is set to `0`)

```python
>>> (0 or None) != 0
True
```

scrapy/settings/default_settings.py
```diff
@@ -17,6 +17,181 @@
...
+    "TELNETCONSOLE_HOST",
+    "TELNETCONSOLE_PASSWORD",
+    "TELNETCONSOLE_PORT",
+    "TELNETCONSOLE_USERNAME",
+    "TEMPLATES_DIR",
+    "TWISTED_REACTOR",
+    "URLLENGTH_LIMIT",
+    "USER_AGENT",
+    "WARN_ON_GENERATOR_RETURN_VALUE",
+    "Path",
```
Gallaecio:
```suggestion
```

scrapy/settings/default_settings.py
```diff
@@ -358,3 +532,19 @@
 USER_AGENT = f"Scrapy/{import_module('scrapy').__version__} (+https://scrapy.org)"
 
 WARN_ON_GENERATOR_RETURN_VALUE = True
+
+
+def __getattr__(name: str):
+    if name == "CONCURRENT_REQUESTS_PER_IP":
```
Gallaecio, line 537: I wonder if we could share some logic with the code in the get method.

Laerte: The message is different, I think the only shareable thing here is the warning part?

scrapy/settings/__init__.py
Laerte: Yeah I did this way because some tests was failing since this setting was evaluating to null: https://github.com/scrapy/scrapy/pull/6921/commits/3ad3b24bfad09c15f8b697fcd4954cb19113e604 but I refactored this condition now, please take a look (also the calls of `CONCURRENT_REQUESTS_PER_IP` setting use getint method that makes the default be 0. https://github.com/search?q=repo%3Ascrapy%2Fscrapy%20(%22CONCURRENT_REQUESTS_PER_IP%22&type=code

Gallaecio: > also the calls of `CONCURRENT_REQUESTS_PER_IP` setting use getint method that makes the default be 0

That is a good point, but I think we should assume this (or more importantly, future settings we might deprecate this way) may also be read with a more generic getter, i.e. `get()` or `[]`, besides their type-specific getter.

scrapy/settings/default_settings.py
Gallaecio: While making the messages as specific as possible is usually a good thing, I am not sure if it is necessary here. If the message focuses on stating that setting A is deprecated, and optionally providing guidelines for replacement, provided the traceback points to the right line (and I think that is the case in both cases, and both use stacklevel=2 for that), I think the message does not require further specificity.

Gallaecio: It is not a big deal, though. Reading myself, it is starting to sound like over-engineering.

Gallaecio (approved):

scrapy/settings/__init__.py
Laerte: Sounds good! Since we implement `__getitem__` we don’t need to call `.get()` in this case, thank you!

PR #6926, opened 2025-06-28
Remove deprecated offsite spider middleware.

wRAR:
Deprecated in 2.11.2 (unfortunately without an explicit release notes entry).


Gallaecio:
> unfortunately without an explicit release notes entry

Maybe we could address that now?

Gallaecio (approved):

Gallaecio (approved):

PR #6929, opened 2025-06-29
Fix: correct get_func_args() handling of functools.partial

mohiuddin-khan-shiam:
### Description
[get_func_args()](cci:1://file:///c:/Users/T2430514/Downloads/scrapy/scrapy/utils/python.py:215:0-241:15) returned positional parameters that were already
supplied via `functools.partial`, because it compared parameter names
with *argument values*.  
The function now relies on `Signature.bind_partial()` to detect the
parameters bound by a `partial` and returns only those still required.

This ensures accurate introspection of callbacks wrapped with
`functools.partial`, preventing unexpected argument-mismatch errors
throughout Scrapy utilities and extensions.

### DIFF
--- a/scrapy/utils/python.py
+++ b/scrapy/utils/python.py
@@ -225,15 +225,22 @@ def get_func_args(func: Callable[..., Any], stripself: bool = False) -> list[str
         return args
 
     if isinstance(func, partial):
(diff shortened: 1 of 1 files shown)


wRAR:
What problem does this fix? Can you add tests for that problem?

PR #6933, opened 2025-07-01
Fix limiting subprocess run time in tests.

wRAR:
Old exceptions were raised in a thread and didn't fail the test. The `timeout` argument to `communicate()` wasn't used initially because it was added in Python 3.3. It's also possible that the failures worked correctly before some test runner changes.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -9,9 +9,10 @@ from io import StringIO
 from pathlib import Path
 from shutil import rmtree
 from tempfile import TemporaryFile, mkdtemp
-from threading import Timer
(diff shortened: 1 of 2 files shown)


wRAR:
`tests/test_command_runspider.py::TestRunSpiderCommand::test_runspider_dnscache_disabled` is indeed close to the 15s limit in successful Windows runs, I'll check what happens.

wRAR:
OK, this test needs disabled retrying and disabled robots.txt handling so that it only waits for 1 failing request.

sins921 (approved):

PR #6937, opened 2025-07-03
Deprecate mustbe_deferred() and defer_*().

wRAR:
`mustbe_deferred()` calls the wrapped function and 
1. if got a Deferred returns that one directly 
2. if got an exception sleeps for `_DEFER_DELAY` and propagates it
3. if got a normal result sleeps for `_DEFER_DELAY` and returns it.

Thus if we know that the wrapped function always returns a Deferred we add a sleep to the `except` block (it looks like all calls outside `MediaPipeline` are like this) and if we know that the wrapped function can return a normal result we wrap it into a `maybeDeferred()` and add a sleep before calling it (this is only in `MediaPipeline`). I think if `MediaPipeline.media_to_download()` returns a Deferred there is one more sleep than before the change, which I'm fine with. 

Note that all of these sleeps seem to be added empirically and doing more of them shouldn't hurt anything except maybe some performance and doing less of them may be fine or increase some blocking/deadlocking chance (as brefiely mentioned in https://www.youtube.com/watch?v=6hiXhNaWTTk).

### DIFF
--- a/scrapy/utils/defer.py
+++ b/scrapy/utils/defer.py
@@ -21,8 +21,8 @@ from twisted.internet.defer import (
 from twisted.internet.task import Cooperator
 from twisted.python import failure
 
-from scrapy.exceptions import IgnoreRequest, ScrapyDeprecationWarning
-from scrapy.utils.asyncio import is_asyncio_available
+from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.utils.asyncio import call_later, is_asyncio_available
 
 if TYPE_CHECKING:
     from collections.abc import AsyncIterator, Callable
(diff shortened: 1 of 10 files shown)


scrapy/utils/defer.py
```diff
@@ -108,22 +138,24 @@ def mustbe_deferred(
...
+        stacklevel=2,
+    )
+    result: _T | Deferred[_T] | Failure
     try:
         result = f(*args, **kw)
-    # FIXME: Hack to avoid introspecting tracebacks. This to speed up
-    # processing of IgnoreRequest errors which are, by far, the most common
-    # exception in Scrapy - see #125
-    except IgnoreRequest as e:
-        return defer_fail(failure.Failure(e))
```
wRAR, line 124: This was very dead code, I've already forgot my analysis but it was dead since about Scrapy 1.0.

tests/test_pipeline_media.py
```diff
@@ -227,15 +228,19 @@ def test_result_succeed(self):
...
             meta={"response": fail},
             errback=self._errback,
         )
         item = {"requests": req}
         new_item = yield self.pipe.process_item(item, self.spider)
-        assert new_item["results"] == [(False, fail)]
+        assert len(new_item["results"]) == 1
+        assert new_item["results"][0][0] is False
+        assert isinstance(new_item["results"][0][1], Failure)
+        assert new_item["results"][0][1].value == exc
```
wRAR, line 243: Comparing failures directly no longer works because @inlineCallbacks recreates them.

sins921 (approved):

PR #6940, opened 2025-07-05
Change default priority queue to DownloaderAwarePriorityQueue

### LINKED ISSUE #6924: Change default SCHEDULER_PRIORITY_QUEUE to scrapy.pqueues.DownloaderAwarePriorityQueue
See: https://github.com/scrapy/scrapy/pull/6921#discussion_r2173142138

thalissonvs:
Closes #6924 

This PR changes the default value of `SCHEDULER_PRIORITY_QUEUE` from `scrapy.pqueues.ScrapyPriorityQueue` to `scrapy.pqueues.DownloaderAwarePriorityQueue`. While implementing this change, two test failures appeared as side effects that required fixes.

The first issue occurred in `tests/test_engine.py::TestEngine::test_start_already_running_exception`. This test was using the ExecutionEngine API incorrectly by creating the engine externally and calling `open_spider()` directly, rather than using the standard `crawler.crawl()` flow. The bug occurs due to initialization timing: `ExecutionEngine.__init__()` is called but `crawler.engine` is not yet assigned, then `engine.open_spider()` is called which creates the scheduler, and the scheduler creates `DownloaderAwarePriorityQueue`. During `DownloaderAwarePriorityQueue.__init__()`, it creates `DownloaderInterface(crawler)`, which executes assert `crawler.engine` (still None at this point). The fix was simply adding `crawler.engine = e` in the test before calling `open_spider()`.

The second issue appeared in `tests/test_downloaderslotssettings.py::TestCrawl::test_delay`. This test deliberately creates requests with `meta={"download_slot": None}`, which works fine with `ScrapyPriorityQueue` since it ignores slots entirely. However, `DownloaderAwarePriorityQueue` organizes requests by slot and calls `_path_safe(slot)` where slot can be `None`, causing a `TypeError`. The difference is that `ScrapyPriorityQueue` uses `self.key + "/" + str(priority)` for internal queues, while `DownloaderAwarePriorityQueue` uses `self.key + "/" + _path_safe(slot)`. The fix was convert None slots to the string "None":

I'm not entirely certain this is the optimal approach for handling `None` slots. An alternative would be treating `None` as a fallback to the default hostname-based slot key rather than creating a literal "None" slot. I'd appreciate feedback on whether the current solution is acceptable or if a different approach would be preferred.

### DIFF
--- a/scrapy/pqueues.py
+++ b/scrapy/pqueues.py
@@ -336,6 +336,8 @@ class DownloaderAwarePriorityQueue:
 
     def push(self, request: Request) -> None:
         slot = self._downloader_interface.get_slot_key(request)
+        # Handle None slot by converting to string
+        slot = str(slot)
         if slot not in self.pqueues:
             self.pqueues[slot] = self.pqfactory(slot)
         queue = self.pqueues[slot]
--- a/docs/topics/broad-crawls.rst
+++ b/docs/topics/broad-crawls.rst
@@ -39,21 +39,6 @@ you need to keep in mind when using Scrapy for doing broad crawls, along with
 concrete suggestions of Scrapy settings to tune in order to achieve an
 efficient broad crawl.
 
-.. _broad-crawls-scheduler-priority-queue:
-
-Use the right :setting:`SCHEDULER_PRIORITY_QUEUE`
-=================================================
-
-Scrapy’s default scheduler priority queue is ``'scrapy.pqueues.ScrapyPriorityQueue'``.
-It works best during single-domain crawl. It does not work well with crawling
-many different domains in parallel
-
-To apply the recommended priority queue use:
-
-.. code-block:: python
-
-    SCHEDULER_PRIORITY_QUEUE = "scrapy.pqueues.DownloaderAwarePriorityQueue"
-
 .. _broad-crawls-concurrency:
 
 Increase concurrency
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1739,10 +1739,10 @@ Type of in-memory queue used by the scheduler. Other available type is:
 SCHEDULER_PRIORITY_QUEUE
 ------------------------
 
-Default: ``'scrapy.pqueues.ScrapyPriorityQueue'``
+Default: ``'scrapy.pqueues.DownloaderAwarePriorityQueue'``
 
 Type of priority queue used by the scheduler. Another available type is
-``scrapy.pqueues.DownloaderAwarePriorityQueue``.
+``scrapy.pqueues.ScrapyPriorityQueue``.
 ``scrapy.pqueues.DownloaderAwarePriorityQueue`` works better than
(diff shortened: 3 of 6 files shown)


scrapy/pqueues.py
```diff
@@ -336,6 +336,8 @@ def pop(self) -> Request | None:
 
     def push(self, request: Request) -> None:
         slot = self._downloader_interface.get_slot_key(request)
+        # Handle None slot by converting to string
+        slot = str(slot)
```
wRAR: This looks wrong, is it needed because of type checks or is it triggered in an actual code path?

thalissonvs: Actually, it's what I said in the description of the PR, there's a call to self._path_safe(slot). 
If the value is None, it break the tests:


"The second issue appeared in tests/test_downloaderslotssettings.py::TestCrawl::test_delay. This test deliberately creates requests with meta={"download_slot": None}, which works fine with ScrapyPriorityQueue since it ignores slots entirely. However, DownloaderAwarePriorityQueue organizes requests by slot and calls _path_safe(slot) where slot can be None, causing a TypeError. The difference is that ScrapyPriorityQueue uses self.key + "/" + str(priority) for internal queues, while DownloaderAwarePriorityQueue uses self.key + "/" + _path_safe(slot). The fix was convert None slots to the string "None""


But I'm not certain about this

thalissonvs: This will treat None as a valid slot. Since `DownloaderAwarePriorityQueue` manages requests by slot, it will simply create a slot named "None", which the downloader will accept as valid.
An alternative would be to treat None as a fallback, using the default hostname-based slot key instead of creating a literal "None" slot

wRAR: > Actually, it's what I said in the description of the PR

My bad, I haven't read that before checking the changes :facepalm: 

> This test deliberately creates requests with meta={"download_slot": None}

Indeed. I wonder if the intention was to not set the `"download_slot"` key in this case so that `get_slot_key()` uses its default logic. cc @GeorgeA92

thalissonvs: Yeah I was not sure how to handle this. Any ideas?

wRAR: I've rechecked this and I think it's simply an assumption that setting the meta key to None is the same as not setting it, and this assumption is not true: while [the `AutoThrottle` extension behaves in this way](https://github.com/scrapy/scrapy/blob/9bfa58e36ccac20193c2175f3fc55636cd44f585/scrapy/extensions/throttle.py#L99), `Downloader` doesn't, [instead only checking if the key exists](https://github.com/scrapy/scrapy/blob/9bfa58e36ccac20193c2175f3fc55636cd44f585/scrapy/core/downloader/__init__.py#L176).

I think `Downloader` needs to be fixed.

wRAR: #7172

PR #6941, opened 2025-07-05
Bump tool versions.

wRAR:


pyproject.toml
```diff
@@ -394,6 +390,9 @@ banned-module-level-imports = [
     "twisted.internet.reactor",
 ]
 
+[tool.ruff.lint.isort]
+split-on-trailing-comma = false
```
wRAR, line 394: This forces short multiline imports to be folded back which I recently noticed to not be done by default. There is[ a similar option for all other comma-separated language things](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma), but the diff without that one is much bigger and I think more controversial, I need to think on what do we want to do.

Gallaecio (approved):

PR #6942, opened 2025-07-05
Enable the flake8-builtins rules.

wRAR:


tests/test_downloadermiddleware_retry.py
```diff
@@ -91,12 +93,12 @@ def test_twistederrors(self):
...
             ConnectionDone,
             ConnectionLost,
-            ConnectionRefusedError,
+            TxConnectionRefusedError,
             defer.TimeoutError,
             DNSLookupError,
             ResponseFailed,
             TCPTimedOutError,
-            TimeoutError,
+            TxTimeoutError,
```
wRAR, line 101: This checked builtin exceptions instead of Twisted ones, the test was passing because the builtin exceptions are subclasses of OSError which is retried.

Gallaecio: Great catch!

Gallaecio (approved):

PR #6947, opened 2025-07-10
Ban more imports that import twisted.internet.reactor.

wRAR:
This partially rolls back the import changes in #6941 but not all of these were correct before that PR.


Gallaecio (approved):
:rocket:

Issue #6805, opened 2025-05-17
Add a httpx-based download handler


wRAR:
Similar to #6707, reasons are similar, and httpx also supports HTTP/2.
While any one of these allows further progress with #6219, it's nice to have more choice as long as the cost for us is low (testing all these new handlers with the same tests should help with that, and in the worst case we can extract the ones we don't want to keep into separate repos).

Johnnti:
@wRAR could you assign me this issue?

elacuesta:
@Johnnti We don't usually assign issues to specific people. Please note that there is already a pull request in progress (#6948), you are welcome to suggest changes there or open your own if you have a different approach in mind.

wRAR:
I'm working on it and almost finished it, it seems that the least straightforward thing about it is cookie handling because `httpx` doesn't have an official way to disable it yet (https://github.com/encode/httpx/issues/2992)

PR #6948, opened 2025-07-10
Adding httpx-based downloader

Rodrigosnrocha:
As per issue #6805, this PR implements a base for an httpx-based download handler. It's being submitted initially as a draft PR as this implementation currently lacks HTTPS support.

To add said support would most likely require either creating multiple download handlers (For example, Http11_X and Https11_X), or requiring parameters either in the download handler's constructor or in download_request(). The latter option would prevent the new handler from being a drop-in replacement for the previous twisted-based one

We tried to reuse the existing tests from the twisted handlers, but have been having issues making them work, though that is currently being addressed.

### DIFF
--- a/tests/test_downloader_handler_httpx.py
+++ b/tests/test_downloader_handler_httpx.py
@@ -1,0 +1,55 @@
+"""Tests for scrapy.core.downloader.handlers.httpx.HTTPXDownloadHandler."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+# Running all pre-existing twisted HTTP download handler tests under HTTPX to avoid writing duplicate tests
+from tests.test_downloader_handler_twisted_http2 import TestHttps2, TestHttp2MockServer
+from tests.test_downloader_handler_twisted_http10 import TestHttp10, TestHttps10
+from tests.test_downloader_handler_twisted_http11 import TestHttp11, TestHttps11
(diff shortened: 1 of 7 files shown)


wRAR:
> this implementation currently lacks HTTPS support.

This is surprising as I don't think httpx needs anything special to enable it.

> To add said support would most likely require either creating multiple download handlers (For example, Http11_X and Https11_X)

I don't think this should be needed, can you elaborate?

tests/test_downloader_handler_httpx.py
```diff
@@ -0,0 +1,55 @@
+"""Tests for scrapy.core.downloader.handlers.httpx.HTTPXDownloadHandler."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+# Running all pre-existing twisted HTTP download handler tests under HTTPX to avoid writing duplicate tests
+from tests.test_downloader_handler_twisted_http2 import TestHttps2, TestHttp2MockServer
+from tests.test_downloader_handler_twisted_http10 import TestHttp10, TestHttps10
+from tests.test_downloader_handler_twisted_http11 import TestHttp11, TestHttps11
```
wRAR, line 10: The idea is to import and subclass base test classes from `tests.test_downloader_handlers_http_base`, like `tests.test_downloader_handler_twisted*` do. Also note that you don't need to test HTTP/1.0.

PR #6952, opened 2025-07-12
[PoC] Reactorless tests.

wRAR:
This adds a new mode of running tests, one that disables pytest-twisted and enforces that there is no reactor. It doesn't use pytest-asyncio for now, instead skipping all async test functions.

This also necessarily changes the logic in `is_asyncio_available()` which may or may not be the way it will behave in proper reactorless-capable Scrapy.

### DIFF
--- a/tests/utils/decorators.py
+++ b/tests/utils/decorators.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+from functools import wraps
+from typing import TYPE_CHECKING, Any, TypeVar
+
+import pytest
+from twisted.internet.defer import Deferred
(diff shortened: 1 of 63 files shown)


scrapy/utils/reactorless.py
wRAR, line 1: This was written for `scrapy.utils` and will be there one day.

tests/utils/decorators.py
wRAR, line 1: With `pytest-asyncio` `inlineCallbacks` will additionally wrap the function into `deferred_to_future()` and `deferred_f_from_coro_f` will be a no-op. Because of the behavior changes the names should be changed into something more generic too.

PR #6959, opened 2025-07-20
Add support for preserving lowercase headers via KEEP_LOWERCASE_HEADERS Flag

IkramKhanNiazi:
### Summary

This PR adds support for preserving lowercase headers in Scrapy requests using a new setting: `KEEP_LOWERCASE_HEADERS`.

### Changes

- Added support for lowercase headers in `Request.__init__` and `Headers` class.
- New Meta Flag `KEEP_LOWERCASE_HEADERS` to keep the header in lower case.
- Added tests to validate the behavior.

### Motivation

Sometime third-party services (e.g., Algolia, Elastic) require custom headers to be in lowercase. Scrapy by default normalizes headers by capitalizing them, which can cause certain requests to fail silently or behave unexpectedly. This often forces users to switch to `requests` or other libraries to handle such cases correctly.

By adding support for lowercase headers, this change allows developers to stay within the Scrapy ecosystem for all their HTTP needs.

### Tests

- Added `tests/test_lowercase_headers.py` with two test cases:
  - Headers are preserved when `KEEP_LOWERCASE_HEADERS` is True.
  - Headers are normalized when it's False (or not explicitly set).
  
- Example usage in a spider:
  ```python
  import scrapy

      name = "example"

      def start_requests(self):
          yield scrapy.Request(
              url="https://httpbin.org/headers",
              headers={"x-custom-lower": "demo"},
              meta={"KEEP_LOWERCASE_HEADERS": True},
              callback=self.parse
          )

      def parse(self, response):
          self.logger.info("Request Headers: %s", response.request.headers)

### DIFF
--- a/scrapy/http/headers.py
+++ b/scrapy/http/headers.py
@@ -27,8 +27,10 @@ class Headers(CaselessDict):
         self,
         seq: Mapping[AnyStr, Any] | Iterable[tuple[AnyStr, Any]] | None = None,
         encoding: str = "utf-8",
+        preserve_case: bool = False,
     ):
         self.encoding: str = encoding
+        self.preserve_case: bool = preserve_case
         super().__init__(seq)
 
     def update(  # type: ignore[override]
@@ -42,6 +44,8 @@ class Headers(CaselessDict):
 
     def normkey(self, key: AnyStr) -> bytes:  # type: ignore[override]
         """Normalize key to bytes"""
+        if self.preserve_case:
+            return self._tobytes(key)
         return self._tobytes(key.title())
 
     def normvalue(self, value: _RawValueT | Iterable[_RawValueT]) -> list[bytes]:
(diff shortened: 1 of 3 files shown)


wRAR:
This shouldn't work, because of https://github.com/scrapy/scrapy/issues/2711#issuecomment-321040485

IkramKhanNiazi:
@wRAR 
This seems to be working as expected. I have tested it locally in both cases.
Crawler used for testing:

```
import scrapy

class LowercaseTestSpider(scrapy.Spider):
    name = "lowercase_test"

    custom_settings = {
        "KEEP_LOWERCASE_HEADERS": True,
    }

    def start_requests(self):
        yield scrapy.Request(
            url="https://httpbin.org/headers",
            headers={"x-my-lower-header": "test-value"},
            callback=self.parse,
            meta={'KEEP_LOWERCASE_HEADERS': True}
        )

    def parse(self, response):
        print(response.request.headers)

```

can you please let me know about the case where it will not work? I will test it

Thanks

Gallaecio:
Print `response.body`.

IkramKhanNiazi:
Closing the PR as it will not work.

PR #6960, opened 2025-07-22
Extract in-process test HTTP servers into separate processes

wRAR:
This leaves three `reactor.listen*` instances (`TestContextFactoryBase`, `TestFTPBase`, `TestWebClient`) that need a reactor for the client part anyway. Though we will want to refactor FTP tests if/when we add a non-Twisted FTP download handler.

~~I'll also try to refactor new mockservers to reduce code duplication.~~ done


conftest.py
```diff
@@ -48,6 +55,12 @@ def _py_files(folder):
     )
 
 
+@pytest.fixture(scope="session")
+def mockserver() -> Generator[MockServer]:
```
Gallaecio, line 59: TIL you can skip the 2nd and 3rd param of `Generator` :facepalm:

Gallaecio (approved):

conftest.py
wRAR: It's very recent! Python 3.13 apparently (and I guess with `from __future__ import annotations` only the mypy version matters).

PR #6966, opened 2025-07-24
Docs: mention Pydantic support

### LINKED ISSUE #6955: Docs: mention Pydantic support
https://docs.scrapy.org/en/latest/topics/items.html#item-types should mention Pydantic support.

staehlmich:
Resolves #6955 

Note: No version information was added as Pydantic support as I could not find this change in the changelogs.

### DIFF
--- a/docs/topics/items.rst
+++ b/docs/topics/items.rst
@@ -140,6 +140,40 @@ Example:
         another_field = attr.ib()
 
 
+.. _pydantic-items:
+
+Pydantic objects
+----------------
+
+`Pydantic <https://docs.pydantic.dev/>`_ models allow the defining of item classes with field names,
+so that :ref:`item exporters <topics-exporters>` can export all fields by
+default even if the first scraped object does not have values for all of them.
+
+Additionally, ``pydantic`` items also allow you to:
+
+* define the type and default value of each defined field with runtime type validation.
+
+* define custom field metadata through :func:`pydantic.Field`, which can be used to
+  :ref:`customize serialization <topics-exporters-field-serialization>`.
+
+* benefit from automatic data validation and conversion based on type annotations.
+
+In order to use this type, the :doc:`pydantic package <pydantic:index>` needs to be installed.
+
+Example:
+
+.. code-block:: python
+
+    from pydantic import BaseModel, Field
+
+
+    class CustomItem(BaseModel):
+        one_field: str = Field(default="", description="First field")
+        another_field: int = Field(default=0, description="Second field")
+
+.. note:: Unlike other item types, Pydantic models enforce field types at runtime
+    and will raise validation errors for invalid data types.
+
 Working with Item objects
 =========================
 
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -364,6 +364,47 @@ used by :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`::
 Spider arguments can also be passed through the Scrapyd ``schedule.json`` API.
 See `Scrapyd documentation`_.
 
+Another alternative to pass spider arguments is the library `scrapy-spider-metadata`_.
+This allows for Scrapy spiders to define, validate, document and pre-process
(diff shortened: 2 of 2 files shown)


docs/topics/items.rst
```diff
@@ -140,6 +140,40 @@ Example:
         another_field = attr.ib()
 
 
+.. _pydantic-items:
+
+Pydantic objects
+----------------
+
```
Gallaecio, line 143: Maybe we should have `.. versionadded:: 2.2` here, like with dataclasses and attrs? (it’s the version that started using itemadapter).

docs/topics/items.rst
```diff
@@ -140,6 +140,40 @@ Example:
...
+Additionally, ``pydantic`` items also allow you to:
+
+* define the type and default value of each defined field with runtime type validation.
+
+* define custom field metadata through :func:`pydantic.Field`, which can be used to
+  :ref:`customize serialization <topics-exporters-field-serialization>`.
+
+* benefit from automatic data validation and conversion based on type annotations.
+
+In order to use this type, the :doc:`pydantic package <pydantic:index>` needs to be installed.
```
Gallaecio: Do these `func` and `doc` references work? I don’t think Pydantic uses Sphinx, so I would expect we need to use regular links to anything in the Pydantic docs.

docs/topics/spiders.rst
```diff
@@ -364,6 +364,47 @@ used by :class:`~scrapy.downloadermiddlewares.useragent.UserAgentMiddleware`::
 Spider arguments can also be passed through the Scrapyd ``schedule.json`` API.
 See `Scrapyd documentation`_.
 
+Another alternative to pass spider arguments is the library `scrapy-spider-metadata`_.
+This allows for Scrapy spiders to define, validate, document and pre-process
+their arguments as pydantic models.
+The example shows how to define typed parameters where a string argument
+is automatically converted to an integer:
```
Gallaecio, line 376: I think a separate, nested section for this makes sense:

```suggestion
scrapy-spider-metadata params
-----------------------------

Another alternative to pass spider arguments is the library `scrapy-spider-metadata`_.

This allows for Scrapy spiders to define, validate, document and pre-process
their arguments as pydantic models.

The example shows how to define typed parameters where a string argument
is automatically converted to an integer:
```

docs/topics/spiders.rst
```diff
@@ -938,6 +979,7 @@ Combine SitemapSpider with other sources of urls:
         def parse_other(self, response):
             pass  # ... scrape other here ...
 
+.. _scrapy-spider-metadata: https://pypi.org/project/scrapy-spider-metadata/
```
Gallaecio: Maybe we should link to https://scrapy-spider-metadata.readthedocs.io/en/latest/params.html instead.

staehlmich:
@Gallaecio I would appreciate if you could review my most recent commits. Thank you!

Gallaecio:
I thought you might want to address https://github.com/scrapy/scrapy/pull/6966#discussion_r2228888971 first. Or do you prefer to keep the current link?

staehlmich:
I submitted this commit: https://github.com/staehlmich/scrapy/commit/b05d65126db3b0ae5ae008afdf15eab6c4fe84b7 , that updated the link, but I forgot to reference it. How do you usually go about referencing commits?

Gallaecio:
You must have accidentally reverted that change when you force-pushed some commits, because the current changes (see the Files tab) have the old link.

(also, no need to force push, at least in Scrapy; regular commits and merges make reviews easier, and we can squash all commits into a single one before merging to keep a clean history)

staehlmich:
What was causing some confusion is that I had included changes for the scrapy-spider-metadata docs issue I was working on. I've cleaned up this branch by removing the unrelated scrapy-spider-metadata documentation changes that I accidentally included. This PR now focuses solely on  #6955 .

The scrapy-spider-metadata documentation is in a separate PR as it's not related to this issue.

Gallaecio (approved):

PR #6975, opened 2025-07-28
Automatically transpose images using EXIF data

### LINKED ISSUE #6525: PIL Image Ignores Orientation in EXIF Data
### Description

This problem seems to exist only for photos taken with a Samsung device but when the image is loaded in PIL it doesn't properly handle the orientation in the EXIF data. 

### Steps to Reproduce

This section of code replicates (roughly) what happens in [Image Pipeline](https://github.com/scrapy/scrapy/blob/d2156696c45e023479ae1bdee8623bb6212e975c/scrapy/pipelines/images.py#L181)

```python
img_url = "https://auctioneersoftware.s3.amazonaws.com/mar/2024/10/ZIC8mPdtm32SMiJmkHcb5-jo.jpg"
response = requests.get(img_url)
img = Image.open(BytesIO(response.content))

img.show() # image will be upside down
```

```python
img_url = "https://auctioneersoftware.s3.amazonaws.com/mar/2024/10/ZIC8mPdtm32SMiJmkHcb5-jo.jpg"
response = requests.get(img_url)
img = Image.open(BytesIO(response.content))

exif = { ExifTags.TAGS[k]: v for k, v in img._getexif().items() if k in ExifTags.TAGS and ExifTags.TAGS[k] != 'MakerNote' }

if exif["Orientation"] == 3:
    # Rotate 180 degrees
    img = img.rotate(180, expand=True)
elif exif["Orientation"] == 6:
    # Rotate 90 degrees
    img = img.rotate(270, expand=True)
elif exif["Orientation"] == 8:
    # Rotate 270 degrees
    img = img.rotate(90, expand=True)

img.show() # image will show correct
```

I would expect images saved through pipeline to match orientation of the site it was scraped from.

Images are upside down

Every time the orientation is given in this form it fails.

### Versions

```
Scrapy       : 2.11.1
lxml         : 5.2.1.0
libxml2      : 2.12.6
cssselect    : 1.2.0
parsel       : 1.9.1
w3lib        : 2.1.2
Twisted      : 24.3.0
Python       : 3.12.2 (main, Apr  5 2024, 14:54:25) [Clang 14.0.0 (clang-1400.0.29.202)]
pyOpenSSL    : 24.1.0 (OpenSSL 3.2.1 30 Jan 2024)
cryptography : 42.0.5
Platform     : macOS-14.4.1-arm64-arm-64bit
```

### Additional context

I believe that the `if/else` above would be the correct fix. Adding this either to `get_images` or `convert_images` functions make sense to me but interested in feedback.

radarhere:
Resolves #6525, by adding using [`ImageOps.exif_transpose()`](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.exif_transpose) to apply transposition from the image's EXIF data in `get_images()`


tests/test_pipeline_images.py
```diff
@@ -169,7 +169,7 @@ def test_get_images(self):
 
         path, new_im, new_buf = next(get_images_gen)
         assert path == "full/3fd165099d8e71b8a48b2683946e64dbfad8b52d.jpg"
-        assert orig_im == new_im
+        assert orig_im.copy() == new_im
```
radarhere, line 172: `orig_im` would be the opened image in the specific class of `JpegImagePlugin` - https://github.com/scrapy/scrapy/actions/runs/16583228546/job/46903534162#step:5:262

Calling [`copy()`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.copy) turns it into a more general `Image` instance, to match what is returned by [`ImageOps.exif_tranpose()`](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.exif_transpose)

wRAR:
Should we try to add tests for this? I feel like this is mostly something for humans to check (there should be a way to check something programmatically but it may be a wrong thing to check).

radarhere:
I've pushed a commit to add a test.

radarhere:
I don't believe a constant exists until Pillow 9.3.0.

wRAR (approved):

wRAR:
Thanks!

PR #6979, opened 2025-07-30
Deprecate non-async counterparts of new async APIs.

wRAR:
Deprecated:
* `ExecutionEngine.start()`
* `ExecutionEngine.open_spider()`
* `Scraper.handle_spider_output()`
* `Scraper.start_itemproc()`
* `SpiderMiddlewareManager.scrape_response()`

Kept:
* `Crawler.crawl()` and `Crawler.stop()` - the async counterparts need asyncio
* `SignalManager.send_catch_log_deferred()` and `scrapy.utils.signal.send_catch_log_deferred()` - makes more sense to deprecate together, but currently `scrapy.utils.signal.send_catch_log_async()` wraps `scrapy.utils.signal.send_catch_log_deferred()`, so it needs to be rewritten to a native implementation first to avoid multiple wrappers
* `Scraper.call_spider()` - needs refactoring of `SpiderMiddlewareManager` to use `call_spider_async()` there which may be easy but I don't think I've investigated it

### DIFF
--- a/scrapy/commands/shell.py
+++ b/scrapy/commands/shell.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any
 from scrapy.commands import ScrapyCommand
 from scrapy.http import Request
 from scrapy.shell import Shell
+from scrapy.utils.defer import deferred_from_coro
 from scrapy.utils.spider import DefaultSpider, spidercls_for_request
 from scrapy.utils.url import guess_scheme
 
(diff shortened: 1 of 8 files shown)


scrapy/commands/shell.py
```diff
@@ -84,7 +85,7 @@ def run(self, args: list[str], opts: Namespace) -> None:
         crawler._apply_settings()
         # The Shell class needs a persistent engine in the crawler
         crawler.engine = crawler._create_engine()
-        crawler.engine.start(_start_request_processing=False)
+        deferred_from_coro(crawler.engine.start_async(_start_request_processing=False))
```
wRAR: `deferred_from_coro()` here and in some other places is used as a shortcut for "if using asyncio, create a task for the coro", as just calling the coro wouldn't schedule it; we could add a private helper function instead, or change all these places to keep the reference to the returned deferred so that the code doesn't look weird (and to add exception handling)

Gallaecio: > we could add a private helper function instead

I think this would be nice. I cannot think of a name, though :sweat_smile:

Gallaecio (approved):

PR #6983, opened 2025-08-02
Update sphinx and sphinx-notfound-page in requirements.txt

ayush-a-srivastava:
- Updated sphinx to 8.2.3
- Updated sphinx-notfound-page to 1.1.0

This change keeps the documentation dependencies up-to-date with the latest stable releases.


Gallaecio:
We (@wRAR really :sweat_smile: ) already handle this kind of updates after each Scrapy release, no need to do them every time there are upstream releases.

PR #605, opened 2014-02-20
Fixed memory issue in sitemap spider by implementing less memory consuming sitemap class

illarion:

PR #6985, opened 2025-08-04
Reducing memory usage of SitemapSpider

galloj:
The method `_parse_sitemap` used to return `Iterable`, which held a reference to the response data + parsed XML.

This commit changes the code, so the URLs from the sitemap are extracted immediately. And the reference to response data + parsed XML are freed, which saves a large amount of memory for sites with a large sitemap.

When I executed my crawler without this change, it used 2340 MB after five minutes. With the change, it used only 730 MB after five minutes. (The memory doesn't grow much from there.) I believe there is room for further improvement by saving the URLs to disk and loading them to memory only when needed and/or, in case of nested sitemaps, loading the next sitemap only when all the pages from the previous sitemap were crawled (but I am not sure how those optimizations could be implemented).

Resolves https://github.com/scrapy/scrapy/issues/3529

### DIFF
--- a/scrapy/spiders/sitemap.py
+++ b/scrapy/spiders/sitemap.py
@@ -91,7 +91,7 @@ class SitemapSpider(Spider):
             # as this Iterable can be long lived
             # and they take much more memory than list of urls
             if s.type == "sitemapindex":
-                urls: list[str] = [
+                urls = [
                     loc
                     for loc in iterloc(it, self.sitemap_alternate_links)
                     if any(x.search(loc) for x in self._follow)
@@ -103,7 +103,7 @@ class SitemapSpider(Spider):
                 for loc in urls:
                     yield Request(loc, callback=self._parse_sitemap)
             elif s.type == "urlset":
-                urls: list[str] = list(iterloc(it, self.sitemap_alternate_links))
+                urls = list(iterloc(it, self.sitemap_alternate_links))
                 del s
                 del it
                 del body


wRAR:
This is interesting. There is #3529, I wonder if this change is about the same problem and solves it. There may be some other recent discussions about this, not sure (besides a recent comment in the linked bug).

albertedwardson:
I believe the high memory usage is caused by the `lxml.etree.fromstring()` call in `scrapy.utils.sitemap.Sitemap`.
`fromstring` builds the entire XML tree in memory, which can easily consume much more RAM than the original response body. (For example, in other formats like JSON, parsing can result in objects that take several times the size of the raw data.)

A more memory-efficient approach would be to rewrite Sitemap to use `lxml.etree.iterparse` (#3529, #605), which processes the XML stream element-by-element instead of loading it all at once. ~~An even better option might be to use Selector, since it can also facilitate lazy loading and reduce peak memory usage.~~ (I think I was wrong about lazy loading in parsel... Well, this can be yet another improvement :p )

I would like to volunteer for this. It seems not that hard but quite useful.

galloj:
@wRAR Yes, this should be solving #3529.

@abebus I have tried to use code form #605, and it seems the memory usage is almost the same (compared to my fix).

The main difference between the lazy and greedy parsing should be that instead of storing the response data, the parsed data are stored. The response data has overhead of the XML tags, and it also might contain things like last modification date and links to images. On the other hand, the parsed data is just a list of strings containing the page URLs. So there, the main overhead is the internal Python structure that contains information about the string.

This means that the comparison result will primarily depend on what data the sitemap contains and the memory overhead of a Python interpreter for strings. It is also possible to use those two approaches together to reduce memory spikes a bit for the greedy parsing (as otherwise, for a short moment, both the response data and parsed XML and parsed URLs are all stored in memory).

Also, as I mentioned in the PR, a good thing would be to store those URLs to disk (which is done by JOBDIR parameter for requests, but doesn't work here as the iterator is long-lived) or by loading only a limited number of sitemaps in parallel. Sadly, I don't really know how to do this.

albertedwardson:
@galloj could you please provide link to a sitemap you were working with? I will use it in benchmarks

galloj:
@abebus I am scraping multiple websites in parallel, so I sadly can't provide a single link.

But to provide the interesting ones:
- large single file sitemap: https://www.albert.cz/sitemap/albertcz_sitemap_products-0.xml.gz
- sitemap with a lot of nested sitemaps: https://knihobot.cz/feeds/sitemaps/sitemap_cz.xml

Gallaecio:
@galloj Shall we close this in favor of https://github.com/scrapy/scrapy/pull/7007?

galloj:
@Gallaecio It would be nice to get the issue fixed, so I was keeping this PR as a failsafe in the case that #7007 fails to be merged for any reason. I believe that my PR is quite small, so it could be reviewed quickly when necessary and even manually copy&pasted for a local fix. I am fine with closing this PR once it seems that #7007 will be merged.

albertedwardson:
@galloj I remember that you use disk queue for requests, but do you also tried to [change order of crawling to bfo](https://docs.scrapy.org/en/latest/topics/scheduler.html#bfo)?

PR #6994, opened 2025-08-07
Deprecate spider attribute overrides, add warnings (scrapy#6988)

### LINKED ISSUE #6988: Deprecate spider attributes that can be replaced by settings
Since Scrapy 1.0, spiders can override project-level settings through `Spider.custom_settings` and `Spider.update_settings()`.

Before that, the workaround was to allow to define spider attributes that could override project settings.

We should deprecate the use of such attributes: each time one is used, log a deprecation warning asking to switch to `Spider.custom_settings` or `Spider.update_settings()` and indicating what the corresponding setting name is.

From what I can tell based on a search for `hasattr\(\S*spider`, 4 such attributes remain:

- [`download_delay`](https://github.com/scrapy/scrapy/blob/527893ddf20b93ea84e3e2a6c9d90dc11119e98f/scrapy/core/downloader/__init__.py#L96)
- [`download_maxsize`](https://github.com/scrapy/scrapy/blob/c47b5d049acd491fad00a02a0fa4af26bcd585ef/scrapy/downloadermiddlewares/httpcompression.py#L77)
- [`download_warnsize`](https://github.com/scrapy/scrapy/blob/c47b5d049acd491fad00a02a0fa4af26bcd585ef/scrapy/downloadermiddlewares/httpcompression.py#L79)
- [`max_concurrent_requests`](https://github.com/scrapy/scrapy/blob/527893ddf20b93ea84e3e2a6c9d90dc11119e98f/scrapy/core/downloader/__init__.py#L99)

NavpreetSSidhu:
Deprecate spider attributes in favor of custom_settings/update_settings()

Add deprecation warnings for the following spider attributes that were
used to override project-level settings before Scrapy 1.0:

- download_delay -> DOWNLOAD_DELAY
- download_maxsize -> DOWNLOAD_MAXSIZE  
- download_warnsize -> DOWNLOAD_WARNSIZE
- max_concurrent_requests -> CONCURRENT_REQUESTS

Each warning includes guidance to use Spider.custom_settings or
Spider.update_settings() instead, along with the corresponding
setting name.

Closes #6988

### DIFF
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -93,9 +93,23 @@ def _get_concurrency_delay(
 ) -> tuple[int, float]:
     delay: float = settings.getfloat("DOWNLOAD_DELAY")
     if hasattr(spider, "download_delay"):
+        warnings.warn(
+            "The 'download_delay' spider attribute is deprecated. "
+            "Use Spider.custom_settings or Spider.update_settings() instead. "
+            "The corresponding setting name is 'DOWNLOAD_DELAY'.",
+            category=ScrapyDeprecationWarning,
+            stacklevel=2,
+        )
         delay = spider.download_delay
 
     if hasattr(spider, "max_concurrent_requests"):
+        warnings.warn(
+            "The 'max_concurrent_requests' spider attribute is deprecated. "
+            "Use Spider.custom_settings or Spider.update_settings() instead. "
+            "The corresponding setting name is 'CONCURRENT_REQUESTS'.",
+            category=ScrapyDeprecationWarning,
+            stacklevel=2,
+        )
         concurrency = spider.max_concurrent_requests
 
     return concurrency, delay
--- a/scrapy/downloadermiddlewares/httpcompression.py
+++ b/scrapy/downloadermiddlewares/httpcompression.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
+import warnings
 from itertools import chain
 from logging import getLogger
 from typing import TYPE_CHECKING, Any
 
 from scrapy import Request, Spider, signals
-from scrapy.exceptions import IgnoreRequest, NotConfigured
+from scrapy.exceptions import IgnoreRequest, NotConfigured, ScrapyDeprecationWarning
 from scrapy.http import Response, TextResponse
 from scrapy.responsetypes import responsetypes
 from scrapy.utils._compression import (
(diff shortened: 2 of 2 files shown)


wRAR:
It looks like some of these are used in tests, those tests should be updated to use the corresponding settings.

NavpreetSSidhu:
ok I will check it and update those should I raise another pr for this over the same branch or amend this one? 
Thanks

wRAR:
It should be added here.

NavpreetSSidhu:
ok, just wanted to wanted to understand that where I should check for tests since they are passing when I checked. I am little new to this test cases writing part, will appreciate some assistance. Thanks

wRAR:
I've searched for those names in the tests, there were some hits for 3 of them, mostly from download handler tests. There are also warnings after tests finish, from `tests/test_downloadermiddleware_httpcompression.py::TestHttpCompression` (I expected all tests that use the deprecated attrs to be visible there, but I guess most of the warnings go into spider logs, not there).

NavpreetSSidhu:
Let me look into this as well

NavpreetSSidhu:
37 passed, 14 skipped, 4 warnings in 0.50s this was the result,
Would it be fine:
A. To update these tests
or
B. Supress these warnings in the tests themselves for now?

On master running the same tests give the following result just the warnings are not present 37 passed, 14 skipped in 0.54s.

wRAR:
I see now that those tests in `TestHttpCompression` specifically test the spider attr support. Then it's fine to mark all those 8 (?) `test_*_spider_attr*` tests with `@pytest.mark.filterwarnings("ignore::scrapy.exceptions.ScrapyDeprecationWarning")`. Same for `test_compression_bomb_spider_attr()` and `test_download_warnsize_spider_attr()` in `test_spider.py`.

For other tests, the ones in `test_extension_throttle.py` need to be simplified so that the deprecated spider attributes are not tested (it's easier than to specifically filter the warning for thise cases), and the download handler tests that pass `download_maxsize` etc. to the `Spider` constructor need to be refactored so that they make a local download handler instance with a local crawler instance with a proper setting set, like it's currently done in e.g. `test_download_allow_data_loss_via_setting()` (please let us know if you have troubles with this part).

NavpreetSSidhu:
I have added the warning suprresion to the test_spider and test_downloadermiddleware_httpcompression.py and they work fine now but before picking test_extension_throttle.py I tested it and it's already working fine, let me know what needs to be done this in this case

wRAR:
> I tested it and it's already working fine, let me know what needs to be done this in this case
 
On the other hand, yeah, if a test still works and doesn't produce warnings then it's fine to keep it as is, as we don't have a policy to remove tests for deprecated features.

NavpreetSSidhu:
ok got it

NavpreetSSidhu:
This commit temporarily suppresses warnings related to deprecated spider attribute overrides to ensure test output remains clean while deprecation is being phased in.

Gallaecio (approved):
:heart: Thanks!

NavpreetSSidhu:
Thanks!

PR #6995, opened 2025-08-08
Fix FilesPipeline for  not handling HTTP 201 Created responses (#1615)

miguelsongco:
addresses issue #1615 by allowing the FilesPipeline to correctly handle HTTP 201 responses during file downloads

 Summary of Changes:
- Updated `media_downloaded` in `scrapy/pipelines/files.py` to accept both 200 and 201 status codes.
- Added support for checking the "Location" header if a 201 response returns without a body.
- Created a unit test (`test_media_downloaded_accepts_201_created`) under `tests/test_pipeline_files.py` to confirm correct behavior with 201 responses and file saving.

This ensures those valid downloads are saved correctly.

 Testing:
- Ran locally with tox -e and tox -- tests/test_pipeline_files.py and passed. 
Please let me know if anything doesn't work as expected or if any adjustments are needed.

### DIFF
--- a/scrapy/pipelines/files.py
+++ b/scrapy/pipelines/files.py
@@ -638,7 +638,31 @@ class FilesPipeline(MediaPipeline):
     ) -> FileInfo:
         referer = referer_str(request)
 
-        if response.status != 200:
+        # Enhanced handling for 201 Created responses
+        if response.status == 201:
+            if not response.body:
+                # If there's no body, check for Location header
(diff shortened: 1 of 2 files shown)


wRAR:
Besides not installing pre-commit it doesn't look like you ran your test.

The test also looks wrong as I don't understand how exactly does it test the 201 handling.

wRAR:

PR #6997, opened 2025-08-08
More *_async APIs

wRAR:
- `ExecutionEngine.stop_async()`
- `ExecutionEngine.close_async()`
- `ExecutionEngine.close_spider_async()`
- `Scraper.open_spider_async()`
- `Scraper.close_spider_async()`

### DIFF
--- a/scrapy/core/engine.py
+++ b/scrapy/core/engine.py
@@ -14,7 +14,7 @@ from time import time
 from traceback import format_exc
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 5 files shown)


Gallaecio (approved):
_Nice_.

Given the amount of `_async` methods we need, maybe my renaming suggestion for the download one is kind of pointless. Maybe we are better off using `_async` consistently for all these scenarios, and eventually deprecate the `_async` methods and make the sync ones async as part of deprecation removal.

PR #7001, opened 2025-08-09
Fix getwithbase() to drop base keys when there is a setting value  (#6912)

adglee:
Resolves Issue #6912

Fixes code in Settings.getwithbase() to drop base keys when the setting value already has them. 
Logs a warning in getwithbase() when duplicate keys are detected (e.g., both "foo.Foo" and Foo present).
Test getwithbase() resolves duplicate keys when merging base and custom settings.


wRAR:
It doesn't look like you ran your test?

PR #7005, opened 2025-08-11
Add ItemPipelineManager.process_item_async() and ensure_awaitable().

wRAR:
Maybe we should also deprecate `ITEM_PROCESSOR` values that aren't subclasses of `ItemPipelineManager` (considering how much code do you need in even a minimal `ITEM_PROCESSOR` if not subclassing that, it's unlikely that those implementations exist, but I'm not sure if the benefits to us from this deprecation are noticeable).

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -117,7 +117,7 @@ class Scraper:
         if not hasattr(self.itemproc, "process_item_async"):
             warnings.warn(
                 f"{global_object_name(itemproc_cls)} doesn't define a process_item_async() method,"
-                f" this is deprecated and the method will be required in the future Scrapy versions.",
+                f" this is deprecated and the method will be required in future Scrapy versions.",
                 ScrapyDeprecationWarning,
                 stacklevel=2,
             )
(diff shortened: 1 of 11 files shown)


scrapy/core/scraper.py
```diff
@@ -109,12 +110,38 @@ def __init__(self, crawler: Crawler) -> None:
...
+        elif (
+            issubclass(itemproc_cls, ItemPipelineManager)
+            and method_is_overridden(itemproc_cls, ItemPipelineManager, "process_item")
+            and not method_is_overridden(
+                itemproc_cls, ItemPipelineManager, "process_item_async"
+            )
         ):
+            warnings.warn(
+                f"{global_object_name(itemproc_cls)} overrides process_item() but doesn't override process_item_async()."
+                f" This is deprecated. process_item() will be used, but in the future Scrapy versions process_item_async() will be used instead.",
```
Gallaecio:
```suggestion
                f" This is deprecated. process_item() will be used, but in future Scrapy versions process_item_async() will be used instead.",
```

(applies to similar strings as well)

PR #7006, opened 2025-08-12
Deprecate spider args of middlewares and pipelines.

### LINKED ISSUE #6927: Deprecate and remove the `spider` argument in middleware and pipeline methods
Continuing the trend, we should work in the direction of removing the `spider` argument from middleware/pipeline methods:

- pipeline `process_item()`
- downloader middleware `process_request()`, `process_response()`, `process_exception()`
- spider middleware `process_start_requests()` (this one is itself deprecated though), `process_spider_input()`, `process_spider_output()`, `process_spider_output_async()`, `process_spider_exception()`

 The way forward is not clear, but maybe:

- In `scrapy.middleware.MiddlewareManager.from_crawler()` set `self.crawler` and refactor its subclasses to use `self.crawler.spider` instead of the spider instance passed to `scrapy.core.downloader.middleware.DownloaderMiddlewareManager.download()`, `scrapy.pipelines.ItemPipelineManager.process_item()` and various methods of `scrapy.core.spidermw.SpiderMiddlewareManager`. Deprecate that argument in the public methods of those.
- Mandate `from_crawler()` and setting `self.crawler` in middlewares and pipelines? Though we don't really need it everywhere, only where the spider instance is needed. So this may just go into some information messages about deprecation.
- In all code and docs use `self.crawler.spider`, change the method arg to `spider: Spider | None = None`, don't pass it in tests etc.
- Maybe check in built-in middlewares if `spider and spider != self.crawler.spider` and raise an exception, or print an error?
- (this is a harder part) In the `MiddlewareManager` subclasses somehow detect if we need to pass the spider argument and only pass it if we need - easy to do if the argument is absent, harder if it's present but shouldn't be passed (check if it's optional?).
- In all built-in middlewares print a deprecation warning if the spider argument is not None (for 3rd-party subclasses and 3rd-party tests).

wRAR:
Partial fix for #6927 

- [x] docs
- [x] tests
~~- [ ] downloader middlewares~~
- [x] pipelines
- [x] `{open,close}_spider()`

Downloader middlewares need more changes as there are more of them so will be done separately.

### DIFF
--- a/scrapy/pipelines/__init__.py
+++ b/scrapy/pipelines/__init__.py
@@ -7,16 +7,20 @@ See documentation in docs/item-pipeline.rst
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable, cast
+
+from twisted.internet.defer import Deferred, DeferredList
 
 from scrapy.exceptions import ScrapyDeprecationWarning
 from scrapy.middleware import MiddlewareManager
 from scrapy.utils.conf import build_component_list
-from scrapy.utils.defer import deferred_from_coro
+from scrapy.utils.defer import deferred_from_coro, maybeDeferred_coro
 from scrapy.utils.python import global_object_name
 
 if TYPE_CHECKING:
-    from twisted.internet.defer import Deferred
+    from collections.abc import Iterable
+
+    from twisted.python.failure import Failure
 
     from scrapy import Spider
     from scrapy.settings import Settings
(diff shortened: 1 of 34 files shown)


wRAR:
should we maybe mark deprecation warnings with `pragma: nocover`

Gallaecio:
I think that’s fine if they are not covered by tests and implementing tests only for them seems overkill.

scrapy/pipelines/__init__.py
```diff
@@ -45,4 +55,38 @@ def process_item(self, item: Any, spider: Spider | None = None) -> Deferred[Any]
         return deferred_from_coro(self.process_item_async(item))
 
     async def process_item_async(self, item: Any) -> Any:
-        return await self._process_chain("process_item", item, self._spider)
+        return await self._process_chain("process_item", item, add_spider=True)
+
+    def _process_parallel(
+        self, methodname: str, add_spider: bool = False
```
Gallaecio: What’s the point of `add_spider`? It seems to be always `True` at the moment, and the method is private.

wRAR: I think it's an artifact of multi-step refactoring of this.

PR #7007, opened 2025-08-12
Rewrite `utils.sitemap.Sitemap` class using `iterparse`

### LINKED ISSUE #3529: SitemapSpider memory issues

abebus:
Also see #6985 #605
Could resolve #3529

Using [files](https://github.com/scrapy/scrapy/pull/6985#issuecomment-3172892111) from @galloj, I've created a simple synthetic benchmark with `memray`, and the results is quite good, reducing memory usage in large sitemaps from 48.677MB
to 235.826kB (about x200 less?)

We also can apply early deletion of response object like @galloj did in #6985

I would like to ask for help in testing this on running spiders.

To reproduce: 
- clone https://github.com/abebus/scrapy/tree/sitemap-memray
- ensure memray is installed
- run ```python sitemap_memory.py``` and ```python sitemap_memory_2.py```
- visualise the results, I did it with ```python -m memray stats sitemap_memory(_2).py_memray.bin ```
- switch to `sitemap-memray-baseline` and repeat steps 3 and 4 to get the baseline

And the results are:
- Case1: One large sitemap: `python -m memray stats sitemap_memory.py_memray.bin `
- Case2: One sitemap with a lot of nested ones (nested not included, I did not run the Spider itself): `python -m memray stats sitemap_memory_2.py_memray.bin `

## Before

Case1

<details>

```

📏 Total allocations:
        448069

📦 Total memory allocated:
        48.999MB

📈 Peak memory usage:
        48.677MB

📊 Histogram of allocation size:
        min: 2.000B
        ---------------------------------------------
        < 4.000B   :   1949 ▇
        < 9.000B   :   3126 ▇
        < 22.000B  :  10975 ▇
        < 49.000B  :   8683 ▇
        < 109.000B : 109063 ▇▇▇▇▇▇▇▇▇
        < 243.000B : 314264 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
        < 543.000B :      6 ▇
        < 1.209kB  :      2 ▇
        < 2.694kB  :      0 
        <=6.001kB  :      1 ▇
        ---------------------------------------------
        max: 6.001kB

📂 Allocator type distribution:
         MALLOC: 428007
         REALLOC: 20062

🥇 Top 5 largest allocating locations (by size):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:27 -> 48.999MB

🥇 Top 5 largest allocating locations (by number of allocations):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:27 -> 448069
```

</details>

Case 2 

<details>

```
📏 Total allocations:
        1972

📦 Total memory allocated:
        225.577kB

📈 Peak memory usage:
        225.440kB

📊 Histogram of allocation size:
        min: 4.000B
        -------------------------------------------
        < 8.000B   :    4 ▇
        < 17.000B  :    2 ▇
        < 35.000B  :    3 ▇
        < 74.000B  :  286 ▇▇▇▇▇
        < 154.000B : 1669 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
        < 321.000B :    4 ▇
        < 668.000B :    1 ▇
        < 1.389kB  :    2 ▇
        < 2.888kB  :    0 
        <=6.001kB  :    1 ▇
        -------------------------------------------
        max: 6.001kB

📂 Allocator type distribution:
         MALLOC: 1970
         REALLOC: 2

🥇 Top 5 largest allocating locations (by size):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:27 -> 225.577kB

🥇 Top 5 largest allocating locations (by number of allocations):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:27 -> 1972
```

</details>

## After

Case1

<details>

```
📏 Total allocations:
        1164

📦 Total memory allocated:
        235.826kB

📈 Peak memory usage:
        156.125kB

📊 Histogram of allocation size:
        min: 4.000B
        ------------------------------------------
        < 9.000B   :   3 ▇
        < 23.000B  :   2 ▇
        < 56.000B  : 175 ▇▇▇▇▇
        < 135.000B : 953 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
        < 326.000B :   4 ▇
        < 786.000B :   3 ▇
        < 1.896kB  :   8 ▇
        < 4.572kB  :   6 ▇
        < 11.026kB :   9 ▇
        <=26.589kB :   1 ▇
        ------------------------------------------
        max: 26.590kB

📂 Allocator type distribution:
         MALLOC: 1141
         REALLOC: 23

🥇 Top 5 largest allocating locations (by size):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:35 -> 228.113kB
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:25 -> 7.713kB

🥇 Top 5 largest allocating locations (by number of allocations):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:35 -> 1151
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:25 -> 13
```

</details>

Case2

<details>

```
📏 Total allocations:
        1486

📦 Total memory allocated:
        290.685kB

📈 Peak memory usage:
        232.767kB

📊 Histogram of allocation size:
        min: 3.000B
        -------------------------------------------
        < 7.000B   :   18 ▇
        < 19.000B  :   41 ▇
        < 48.000B  :   44 ▇
        < 123.000B : 1262 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
        < 314.000B :   94 ▇▇
        < 796.000B :    5 ▇
        < 2.017kB  :    8 ▇
        < 5.115kB  :    7 ▇
        < 12.966kB :    5 ▇
        <=32.868kB :    2 ▇
        -------------------------------------------
        max: 32.869kB

📂 Allocator type distribution:
         MALLOC: 1395
         REALLOC: 91

🥇 Top 5 largest allocating locations (by size):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:35 -> 282.972kB
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:25 -> 7.713kB

🥇 Top 5 largest allocating locations (by number of allocations):
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:35 -> 1473
        - __init__:/home/anne/scrapy/scrapy/utils/sitemap.py:25 -> 13
```

</details>

### DIFF
--- a/scrapy/utils/sitemap.py
+++ b/scrapy/utils/sitemap.py
@@ -7,6 +7,7 @@ SitemapSpider, its API is subject to change without notice.
 
 from __future__ import annotations
 
+from io import BytesIO, StringIO
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urljoin
 
@@ -20,31 +21,39 @@ class Sitemap:
     """Class to parse Sitemap (type=urlset) and Sitemap Index
     (type=sitemapindex) files"""
 
-    def __init__(self, xmltext: str | bytes):
-        xmlp = lxml.etree.XMLParser(
-            recover=True, remove_comments=True, resolve_entities=False
+    def __init__(self, xmltext: bytes):
+        self.xmliter = lxml.etree.iterparse(
+            BytesIO(xmltext),
+            recover=True,
+            remove_comments=True,
+            resolve_entities=False,
+            remove_blank_text=True,
+            collect_ids=False,
+            remove_pis=True,
+            events=("start", "end"),
         )
-        self._root = lxml.etree.fromstring(xmltext, parser=xmlp)
-        rt = self._root.tag
-        assert isinstance(rt, str)
-        self.type = rt.split("}", 1)[1] if "}" in rt else rt
+        _, elem = next(self.xmliter)
+        self.type = self._get_type(elem)
 
     def __iter__(self) -> Iterator[dict[str, Any]]:
-        for elem in self._root.getchildren():
+        for _, elem in self.xmliter:
             d: dict[str, Any] = {}
-            for el in elem.getchildren():
-                tag = el.tag
-                assert isinstance(tag, str)
-                name = tag.split("}", 1)[1] if "}" in tag else tag
-
+            for el in elem:
+                name = self._get_type(el)
                 if name == "link":
                     if "href" in el.attrib:
                         d.setdefault("alternate", []).append(el.get("href"))
                 else:
                     d[name] = el.text.strip() if el.text else ""
-
+                el.clear()
             if "loc" in d:
                 yield d
+            elem.clear()
+
+    @staticmethod
+    def _get_type(elem: lxml.etree._Element) -> str:
+        _, _, localname = str(elem.tag).partition("}")
+        return localname or elem.tag
 
 
 def sitemap_urls_from_robots(
@@ -53,7 +62,7 @@ def sitemap_urls_from_robots(
     """Return an iterator over all sitemap urls contained in the given
     robots.txt file
     """
-    for line in robots_text.splitlines():
-        if line.lstrip().lower().startswith("sitemap:"):
-            url = line.split(":", 1)[1].strip()
+    for line in StringIO(robots_text):
+        if line.lstrip()[:8].lower() == "sitemap:":
+            url = line.partition(":")[2].strip()
             yield urljoin(base_url or "", url)


scrapy/utils/sitemap.py
```diff
@@ -20,31 +21,39 @@ class Sitemap:
     """Class to parse Sitemap (type=urlset) and Sitemap Index
     (type=sitemapindex) files"""
 
-    def __init__(self, xmltext: str | bytes):
-        xmlp = lxml.etree.XMLParser(
-            recover=True, remove_comments=True, resolve_entities=False
+    def __init__(self, xmltext: bytes):
+        self.xmliter = lxml.etree.iterparse(
+            BytesIO(xmltext),
```
albertedwardson, line 39: forgot that we also need to accept str, will add isinstance(xmltext, str) and wrap it in custom StringIO that will encode results from read(), to not reencode whole response body once again

scrapy/utils/sitemap.py
```diff
@@ -20,31 +21,39 @@ class Sitemap:
...
+            remove_comments=True,
+            resolve_entities=False,
+            remove_blank_text=True,
+            collect_ids=False,
+            remove_pis=True,
+            events=("start", "end"),
         )
-        self._root = lxml.etree.fromstring(xmltext, parser=xmlp)
-        rt = self._root.tag
-        assert isinstance(rt, str)
```
albertedwardson, line 29: i hope mypy will not be mad at this... will check this

scrapy/utils/sitemap.py
```diff
@@ -53,7 +62,7 @@ def sitemap_urls_from_robots(
     """Return an iterator over all sitemap urls contained in the given
     robots.txt file
     """
-    for line in robots_text.splitlines():
-        if line.lstrip().lower().startswith("sitemap:"):
-            url = line.split(":", 1)[1].strip()
+    for line in StringIO(robots_text):
+        if line.lstrip()[:8].lower() == "sitemap:":
```
albertedwardson, line 133: doing it like this will call lower() only at the first 8 characters in string, not entire

scrapy/utils/sitemap.py
```diff
@@ -20,31 +21,39 @@ class Sitemap:
...
         )
-        self._root = lxml.etree.fromstring(xmltext, parser=xmlp)
-        rt = self._root.tag
-        assert isinstance(rt, str)
-        self.type = rt.split("}", 1)[1] if "}" in rt else rt
+        _, elem = next(self.xmliter)
+        self.type = self._get_type(elem)
 
     def __iter__(self) -> Iterator[dict[str, Any]]:
-        for elem in self._root.getchildren():
```
albertedwardson, line 33: types-lxml told that this method is deprecated

scrapy/utils/sitemap.py
```diff
@@ -16,44 +17,70 @@
     from collections.abc import Iterable, Iterator
 
 
+def _bytes_io_from_anystr(str_or_bytes: str | bytes) -> BytesIO:
+    return (
+        BytesIO(str_or_bytes)
+        if isinstance(str_or_bytes, bytes)
+        else BytesIO(str_or_bytes.encode())
```
albertedwardson: this actually introduces overhead if called with `str`, but I think it will be fine to add warning that `str` as type will be deprecated and accept only bytes

scrapy/spiders/sitemap.py
```diff
@@ -85,18 +92,26 @@ def _parse_sitemap(self, response: Response) -> Iterable[Request]:
...
-            elif s.type == "urlset":
-                for loc in iterloc(it, self.sitemap_alternate_links):
-                    for r, c in self._cbs:
-                        if r.search(loc):
-                            yield Request(loc, callback=c)
-                            break
+            del body, response
+
+            requests = list(self.__get_sitemap_requests(s, self.sitemap_filter(s)))
+            del s
```
albertedwardson: idk how to rewrite this without `del`s without a lot of refactoring

albertedwardson:
I hope I'm done, it's ready for review 
Sorry for a lot of commits, this needs to be squashed if everything else is ok

scrapy/utils/sitemap.py
```diff
@@ -16,44 +17,86 @@
...
-            recover=True, remove_comments=True, resolve_entities=False
+        self.xmliter = lxml.etree.iterparse(
+            _bytes_io_from_anystr(xmltext),
+            recover=True,
+            remove_comments=True,
+            resolve_entities=False,
+            remove_blank_text=True,
+            collect_ids=False,
+            remove_pis=True,
+            events=("start", "end"),
```
galloj, line 46: Is there any reason why both start and end events are collected?

albertedwardson:
@galloj hi! it would be super nice if you could try to run your spiders with scrapy installed from this branch and report memory usage

i believe even peak memory spikes will be lower, since now sitemap will be stored only as bytes buffer

albertedwardson:
I have a mad idea of making helper class for memoryview so it's satisfies SupportsRead protocol. This will be as close to zero copy as possible and we can feed response.body to lxml.etree.iterparse almost directly, without copying it like BytesIO does

albertedwardson:
Next time when I will have free time I will test it on running spider with memray attached and include the results here

albertedwardson:
Since docsting in `scrapy.utils.sitemap` states that
```
Note: The main purpose of this module is to provide support for the
SitemapSpider, its API is subject to change without notice.
```
I would like to drop accepting `str` in `Sitemap.__init__` and `sitemap_urls_from_robots` if maintainers will approve this change too

tests/test_spider.py
```diff
@@ -846,6 +849,55 @@ class TestSpider(self.spider_class):
...
+    )
+    def test_sitemap_filter_with_rule(self, rule: str, result: list[str]):
+        sitemap = b"""<?xml version="1.0" encoding="UTF-8"?>
+    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+        <url><loc>http://www.example.com/english/</loc></url>
+        <url><loc>http://www.example.com/portuguese/</loc></url>
+    </urlset>"""
+        r = TextResponse(url="http://www.example.com/sitemap.xml", body=sitemap)
+
+        class RuleSpider(self.spider_class):  # type: ignore[name-defined]
```
albertedwardson: why mypy became mad at this????

abebus:
spare me coverage IMG_5943

scrapy/utils/misc.py
```diff
@@ -308,3 +311,125 @@ def warn_on_generator_with_return_value(
...
+            else cls(str_or_bytes.encode())
+        )
+
+    def read(self, amount: int = -1, /) -> bytes:
+        if amount < 0:
+            end = self.size
+        else:
+            end = self._pos + amount
+            if end > self.size:  # pylint: disable=consider-using-min-builtin # noqa: PLR1730
+                end = self.size
```
albertedwardson: I didn't extract this to a separate function since `read()` and `readline()` are quite a hot paths and calling a function introduces small overhead.

scrapy/utils/gz.py
```diff
@@ -16,8 +17,8 @@ def gunzip(data: bytes, *, max_size: int = 0) -> bytes:
 
     This is resilient to CRC checksum errors.
     """
-    f = GzipFile(fileobj=BytesIO(data))
-    output_stream = BytesIO()
+    f = GzipFile(fileobj=MemoryviewReader(data))
+    output_stream = bytearray()
```
albertedwardson: bytearray is a bit smaller compared to BytesIO

albertedwardson:
It's ready for review. Benchmarks are included in the initial PR comment

albertedwardson:
may be related and useful in future memory-focused optimisations: 

- https://docs.python.org/3/reference/datamodel.html#python-buffer-protocol
- https://peps.python.org/pep-0688

albertedwardson:
if this pr will be merged and `MemoryviewReader` class will be accepted, I would like to open a new issue and few additional pr's where this class can be used to reduce memory copying.

scrapy/utils/misc.py
```diff
@@ -308,3 +311,125 @@ def warn_on_generator_with_return_value(
...
+            if isinstance(str_or_bytes, bytes)
+            else cls(str_or_bytes.encode())
+        )
+
+    def read(self, amount: int = -1, /) -> bytes:
+        if amount < 0:
+            end = self.size
+        else:
+            end = self._pos + amount
+            if end > self.size:  # pylint: disable=consider-using-min-builtin # noqa: PLR1730
```
galloj: You can use `end = min(self._pos + amount, self.size)` to fix the pylint warning.

scrapy/utils/misc.py
```diff
@@ -308,3 +311,125 @@ def warn_on_generator_with_return_value(
...
+    def seek(self, offset: int, whence: int = 0) -> int:
+        if whence == io.SEEK_SET:
+            new = offset
+        elif whence == io.SEEK_CUR:
+            new = self._pos + offset
+        elif whence == io.SEEK_END:
+            new = self.size + offset
+        else:
+            raise ValueError("Invalid whence")
+        if not (0 <= new <= self.size):  # pylint: disable=superfluous-parens
```
galloj: You can replace the condition with `new not in range(self.size+1)` or extract it to a variable to fix the pylint warning.

scrapy/utils/sitemap.py
```diff
@@ -21,39 +23,76 @@ class Sitemap:
...
-                assert isinstance(tag, str)
-                name = tag.split("}", 1)[1] if "}" in tag else tag
-
-                if name == "link":
-                    if "href" in el.attrib:
-                        d.setdefault("alternate", []).append(el.get("href"))
+        for _, elem in self.xmliter:
+            try:
+                tag_name = self._get_tag_name(elem)
+                if not tag_name or (tag_name != "url" and tag_name != "sitemap"):  # pylint: disable=consider-using-in # noqa: PLR1714
```
galloj: You can change `(tag_name != "url" and tag_name != "sitemap")` to `tag_name not in ("url", "sitemap")`.

scrapy/spiders/sitemap.py
```diff
@@ -72,31 +72,44 @@ def sitemap_filter(
...
-                    if any(x.search(loc) for x in self._follow):
-                        yield Request(loc, callback=self._parse_sitemap)
-            elif s.type == "urlset":
-                for loc in iterloc(it, self.sitemap_alternate_links):
-                    for r, c in self._cbs:
-                        if r.search(loc):
-                            yield Request(loc, callback=c)
-                            break
+            del body
+            requests = list(self.__get_sitemap_requests(s, self.sitemap_filter(s)))
```
galloj: The function `__get_sitemap_requests` yields `Request` objects, which are quite large, so it is not a good idea to put them into a long-lived list.

galloj (commented):
I have checked the PR with my spider, and the memory usage was 1660 MB. After removing the conversion of `__get_sitemap_requests` to a list, the memory usage dropped to 610 MB, which was a bit better than 630 MB, which I achieved with my PR #6985 (rough measurement after five minutes of the spider running; and the number is a bit different than the one mentioned in PR as I probably changed something/some websites changed, not sure).

Also, I would like to mention that I had done the benchmark with `JOBDIR` parameter, which significantly reduces memory usage.

scrapy/utils/gz.py
galloj: I think the `output_stream` should be renamed, as it isn't a stream anymore.

scrapy/spiders/sitemap.py
albertedwardson: by assuming that Request object hold more data than a string, I believe they are definitely larger than plain strings. I will change urls as you did and reevaluate memory usage

scrapy/utils/misc.py
albertedwardson: did it like this for sake of little performance savings

scrapy/utils/misc.py
albertedwardson: same here, for perf

scrapy/utils/sitemap.py
albertedwardson: same here, for perf

scrapy/utils/sitemap.py
albertedwardson: by some reason, when collecting only end events, getting name of the first tag (type of sitemap) is ommited, but I think we can put `continue` inside a main loop if event is start.

or maybe recreate new iterparse instance with only end events after getting the type of sitemap

albertedwardson:
>  After removing the conversion of `__get_sitemap_requests` to a list, 

@galloj you mean it was yielding requests from generator, not list? if it was, it's strange why memory usage dropped, since this generator holds reference to sitemap object, that holds reference to response body

galloj:
@abebus 

> you mean it was yielding requests from generator, not list? if it was, it's strange why memory usage dropped, since this generator holds reference to sitemap object, that holds reference to response body

Yes, I changed `requests = list(self.__get_sitemap_requests(s, self.sitemap_filter(s)))` to `requests = self.__get_sitemap_requests(s, self.sitemap_filter(s))`. The memory usage dropped because the `Request` object is large enough to offset the benefits of freeing the response body.

```python3
from pympler.asizeof import asizeof
from scrapy import Request

asizeof(Request("http://example.com"), Request("http://examplz.com")) - asizeof(Request("http://example.com"))
# 1016

asizeof("http://example.com", "http://examplz.com") - asizeof("http://example.com")
# 72
```

Also, the string itself carries some overhead (about 50 bytes), so sometimes, the response body might be smaller than the list of parsed URLs. (On the other hand, it might be significantly bigger if the sitemap also contains images).

scrapy/utils/sitemap.py
galloj: The end event for the root tag should still be received, but as the last one instead of the first one.

The main reason why the `start` tag should be avoided, is because it then leads to duplicate URLs (as the URL is collected once for `start` and once for `end`). These duplicates then get filtered by scapy, but until then they are unnecessarily taking up memory.

scrapy/utils/misc.py
galloj: Ah, I think it is a good idea to mention this in the code, as it isn't immediately clear why you are suppressing the linter instead of changing the code.

scrapy/utils/sitemap.py
albertedwardson: I use both start and end events in lxml.etree.iterparse because they serve different purposes

On start events the element text is not yet available, so no URLs are extracted and there is no duplication

I only use start events to detect the type of sitemap early and to allow early exits if the tag name differs from url or sitemap. If I waited until the end event to get the type of sitemap , I wouldn’t know the type until the entire element was closed and whole logic of extracting urls would fail

On end events the element text is available, and this is the only place where I collect URLs

Since URLs are only taken from end events, they cannot be duplicated. Unit tests confirm this since they would fail if duplicates existed

An alternative could be running two iterparse passes, one for start and one for end, but that adds complexity without clear benefit

The simplest safeguard is just put "continue" when encountering a start event after tag name check's passed

albertedwardson: After looking at it again, I realized we might not need the "end" event at all. For sitemap XML the structure is shallow and predictable, and urls are already available at the "start" event.

albertedwardson:
I was wrong. 

We don't need `MemoryviewReader` class at all since `BytesIO` [do not copy input data](https://github.com/python/cpython/issues/138293#issuecomment-3240156029) when it comes to read operations. This makes this PR much lighter.

It would make sence if `iterparse` could accept memoryview as its input data. 

All we need to do is wait until `BytesIO` will be copy-on-write (which is a complex task to do and may be not accomplished at all)

albertedwardson:
I did memory usage measurements using tracemalloc to estimate the cost of storing sitemap data. 

Storing 50,000 plain Python strings of length 32-64 consumes approximately 0,5 to 0,6 MB of memory. Considering that sitemap bodies can grow up to 50 MB according to the specification, it's better to store list of urls, not the response body.

Out of interest, I evaluated marisa-trie, where the resulting memory usage depends on the similarity of the stored URLs. For a set of 50,000 highly similar URLs, the memory footprint was around 5 KB with a peak of 35 KB. For 50,000 randomly generated strings, the footprint was approximately 7 KB with a peak of 50 KB. 

Although the difference compared to plain strings is huge, the benefit does not appear to justify introducing an additional binary dependency into the project just for this case

albertedwardson:
note: tests must cover cases where sitemap is larger than lxml iterparse buffer size

scrapy/utils/sitemap.py
```diff
@@ -20,40 +21,77 @@ class Sitemap:
     """Class to parse Sitemap (type=urlset) and Sitemap Index
     (type=sitemapindex) files"""
 
-    def __init__(self, xmltext: str | bytes):
-        xmlp = lxml.etree.XMLParser(
-            recover=True, remove_comments=True, resolve_entities=False
+    def __init__(self, xmltext: bytes):
```
Gallaecio: This is backward-incompatible. Can we keep `str` support? (you can log a deprecation warning if a string is received if you want, as long as it continues to work for now)

While I acknowledge the module docstring contents, I don’t expect users to read that. I think we should have called the module `scrapy.utils._sitemap` to properly convey that it was an internal API, and now it is too late 😅

scrapy/utils/sitemap.py
Gallaecio: Is the performance difference significant enough to justify it, though?

scrapy/utils/sitemap.py
```diff
@@ -20,40 +21,77 @@ class Sitemap:
...
+    def _get_tag_name(elem: lxml.etree._Element) -> str:
+        assert isinstance(elem.tag, str)
+        _, _, localname = elem.tag.partition("}")
+        return localname or elem.tag
 
 
 def sitemap_urls_from_robots(
-    robots_text: str, base_url: str | None = None
+    robots_text: bytes, base_url: str | None = None
 ) -> Iterable[str]:
```
Gallaecio, line 126: Now sure how to best handle this backward-incompatible change. Maybe we should implement an alternative (private?) method to handle bytes, and somehow log a deprecation warning if this method is called or overridden in a subclass.

scrapy/utils/sitemap.py
wRAR: We could mass-deprecate utils that we don't want to be public. This is not the first time having them as public hinders our progress (e.g. we mass-deprecated test-related things recently, not just to reduce maintenance burden but mostly to be able to refactor them when refactoring tests).

scrapy/utils/sitemap.py
albertedwardson: It depends... But I think this is a common pattern that could be optimised at bytecode level by python itself

Nonetheless, I will measure it and if this version will be faster than `x in y` for about 10%, I'd like to keep it

albertedwardson:
Will mark this as WIP until I resolve all comments

albertedwardson:
lol 
изображение

albertedwardson:
@Gallaecio @wRAR please run  `git lfs track "tests/sample_data/**"` and update `.gitattributes` or something else to properly track `sample_data` folder :)

this PR is such a mess... I'm sorry

albertedwardson:
btw, running tests with memray plugin gave me such results: 

```
~/scrapy$ uv run tox -e tests -- tests/test_spider.py --memray
.pkg: _optional_hooks> python /home/anne/.local/share/uv/tools/tox/lib/python3.13/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_sdist> python /home/anne/.local/share/uv/tools/tox/lib/python3.13/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: build_sdist> python /home/anne/.local/share/uv/tools/tox/lib/python3.13/site-packages/pyproject_api/_backend.py True hatchling.build
tests: install_package> /home/anne/.local/share/uv/tools/tox/bin/uv pip install --reinstall --no-deps scrapy@/home/anne/scrapy/.tox/.tmp/package/301/scrapy-2.14.1.tar.gz
tests: commands[0]> pytest tests/test_spider.py --memray
============================================================================================ test session starts ============================================================================================
platform linux -- Python 3.13.3, pytest-9.0.2, pluggy-1.6.0
cachedir: .tox/tests/.pytest_cache
rootdir: /home/anne/scrapy
configfile: pyproject.toml
plugins: xdist-3.8.0, twisted-1.14.3, cov-7.0.0, memray-1.8.0
collected 103 items                                                                                                                                                                                         

tests/test_spider.py .......................................................................................................                                                                          [100%]


=============================================================================================== MEMRAY REPORT ===============================================================================================
Allocation results for tests/test_spider.py::TestSitemapSpider::test_large_sitemaps[sitemap_too_large.xml-536121] at the high watermark

         📦 Total memory allocated: 167.0MiB
         📏 Total allocations: 10
         📊 Histogram of allocation sizes: |▂  ██|
         🥇 Biggest allocating functions:
                - read_bytes:/home/anne/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/pathlib/_abc.py:626 -> 50.0MiB
                - _splitnetloc:/home/anne/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/urllib/parse.py:416 -> 41.0MiB
                - __iter__:/home/anne/scrapy/scrapy/utils/sitemap.py:52 -> 33.0MiB
                - _process_sitemap_element:/home/anne/scrapy/scrapy/utils/sitemap.py:79 -> 29.0MiB
                - _process_sitemap_element:/home/anne/scrapy/scrapy/utils/sitemap.py:79 -> 5.0MiB


Allocation results for tests/test_spider.py::TestSitemapSpider::test_large_sitemaps[sitemap_large_valid.xml-50000] at the high watermark

         📦 Total memory allocated: 15.5MiB
         📏 Total allocations: 8
         📊 Histogram of allocation sizes: |▄ ███|
         🥇 Biggest allocating functions:
                - _process_sitemap_element:/home/anne/scrapy/scrapy/utils/sitemap.py:79 -> 5.0MiB
                - read_bytes:/home/anne/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/pathlib/_abc.py:626 -> 4.6MiB
                - _splitnetloc:/home/anne/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/urllib/parse.py:416 -> 3.0MiB
                - __iter__:/home/anne/scrapy/scrapy/utils/sitemap.py:52 -> 1.0MiB
                - _process_sitemap_element:/home/anne/scrapy/scrapy/utils/sitemap.py:79 -> 1.0MiB

```

scrapy/utils/sitemap.py
albertedwardson: > the structure is shallow and predictable, and urls are already available at the "start" event.
Not always. There are cases where the url tag is only partially parsed and its children are not yet available; they may arrive in the next parser chunk.

So we need the "end" event for url elements to ensure they are fully parsed, and the "start" event for detecting the sitemap type.

galloj:
@albertedwardson It would be a good idea to create a helper function for the tests that generates the sitemap (as the XMLs you committed are generated anyway) based on the required number of URLs. Because it is a bit excessive to introduce a 50 MB file just for tests. The helper function would also be useful to see the correctness of the tests, as it is impossible for a human to read the contents of those XML files due to their size.

albertedwardson:
@galloj 
yeah I just thought that "binary" mark on "sample_data" folder in `.gitattributes ` was enough to make this files as pointers to external storage...

tests/test_utils_sitemap.py
```diff
@@ -156,7 +156,7 @@ def test_sitemap_wrong_ns2():
 
 
 def test_sitemap_urls_from_robots():
-    robots = """User-agent: *
+    robots = b"""User-agent: *
```
AdrianAtZyte, line 161: It would be nice to add a test that proves that a string version still works the same, but triggers the deprecation warning.

AdrianAtZyte:
> Storing 50,000 plain Python strings of length 32-64 consumes approximately 0,5 to 0,6 MB of memory.

```python
import random
import string
import sys

N = 50_000

def rand_url():
    tail = "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(40))
    return f"https://example.com/{tail}"

urls = [rand_url() for _ in range(N)]
list_overhead = sys.getsizeof(urls)
items_total = sum(sys.getsizeof(u) for u in urls)
total = list_overhead + items_total
print("total_mib:", round(total / (1024 * 1024), 3))
```

```
$ uv run --python 3.13 test.py 
…
total_mib: 5.24
```

Which makes me wonder if we should indeed keep the iter approach instead of `list()` casting.

albertedwardson:
not evaluating urls to list will keep a strong reference to sitemap body and its xml parser, which could be +-50 mb is size in the worst case, instead of 0,6 mb

and this is true for each request

AdrianAtZyte:
> not evaluating urls to list will keep a strong reference to sitemap body and its xml parser, which could be +-50 mb is size in the worst case, instead of 0,6 mb

Don’t we use a streaming parser that solves the XML parser issue with memory? And doesn’t the body remain in memory until all URLs have been consumed by the caller anyway?

albertedwardson:
> Don’t we use a streaming parser that solves the XML parser issue with memory? 

it does, but the body is still kept alive until all urls are processed 

> And doesn’t the body remain in memory until all URLs have been consumed by the caller anyway?

if we evaluate urls as list of strings- it should not. it decrefed to zero and destroyed at returning generator of requests above list of urls

AdrianAtZyte:
Can you think of a way to modify the memory test I have added, which currently starts passing when you switch from list to generators, to make it do the opposite, i.e. pass only with lists but not with generators?

AdrianAtZyte:
Something like this seems to reproduce what you mean:

```python


    def test_parse_sitemap_does_not_retain_response_bodies_for_parallel_parsing(self):
        spider = self.spider_class("example.com")

        tracemalloc.start()
        generators = []
        for i in range(32):
            r = XmlResponse(
                url=f"http://www.example.com/sitemap-{i}.xml",
                body=self._generate_sparse_sitemap_with_padding(i, 300_000),
            )
            generators.append(spider._parse_sitemap(r))

        # Keep parse generators alive, but release all responses to mimic scheduler
        # queuing many sitemap requests at once.
        gc.collect()
        current, _ = tracemalloc.get_traced_memory()
        tracemalloc.stop()

        # Current implementation materializes URLs into lists before returning,
        # so response bodies should be released quickly. A lazy pipeline that
        # keeps parser state alive would retain ~32 * 300_000 bytes here.
        assert current < 3_000_000

        # Sanity-check that all retained generators are still consumable.
        for g in generators:
            req = next(iter(g))
            assert req.url.startswith("https://example.com/page-")
            
    def _generate_sparse_sitemap_with_padding(self, idx: int, padding_size: int) -> bytes:
        padding = b"x" * padding_size
        return (
            b'<?xml version="1.0" encoding="UTF-8"?>'
            b'<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">'
            b"<url><loc>https://example.com/page-"
            + str(idx).encode()
            + b"</loc></url>"
            + b"<meta><blob>"
            + padding
            + b"</blob></meta>"
            + b"</urlset>"
        )
```

However, note the `300_000` padding I am adding to the bodies. With realistic sitemaps, where at most you get proper 4-space indentation for the XML elements, the size of the URL list in memory seems to still not make up for any savings due to releasing the body.

albertedwardson:
I can't prove my point right now, but I think tracemalloc should be started before generating the response body, since it isn't tracked with generators, and they are indeed lighter than a list of strings

plus, this test case doesn't take into account queued requests and responses; it only covers memory usage during sitemap processing

in other words, url materialization is expensive per method call, but, as my integration testing efforts have shown, it's cheaper than lazy url evaluation and storing references to response bodies

galloj:
@AdrianAtZyte Let me share what a realistic sitemap response looks like:
```xml
<url><loc>https://www.alza.cz/kabel-prodluzovaci-pro-mys-d39984.htm</loc><changefreq>weekly</changefreq><priority>1</priority><image:image><image:loc>https://image.alza.cz/products/MK172/MK172.jpg</image:loc></image:image></url>
```
or
```xml
<url><loc>https://www.datart.cz/napajeci-adapter-acer-65w-5-5phy-np-adt0a-078.html</loc><priority>0.9</priority><changefreq>daily</changefreq><lastmod>2026-03-25</lastmod></url>
```
There is no padding, and in some cases, not even new lines. However, there is an overhead of the XML tags and additional metadata that is not useful to us.

Your example showed that it takes 5.24 MiB to store 2.86 MiB of text, which is 83% overhead. In my examples, the overhead of XML representation is 300% and 146% (ratio of URL to non-URL text in the example), which justifies the overhead of the string.

AdrianAtZyte:
I think you are both right, but before we merge, I would like to have some test that can demonstrate the problem. A test that breaks as you remove the list() casting.

albertedwardson:
ok will do it

albertedwardson:
comments and docstrings were generated with ai and reviewed by me. if they feel overly "comprehensive"- feel free to modify or remove them.

albertedwardson:
we could skip memory usage test on pypy or do something with: https://pypy.org/posts/2025/02/pypy-gc-sampling.html

albertedwardson:
windows 3.11 run is unrelated

AdrianAtZyte:
@kmike, @wRAR, @GeorgeA92 and I had an off-GitHub discussion on the topic, and agreed to move on, that there is definitely a good performance improvement here, that any issue caused by creating too many requests too fast in the scheduler is not an issue with sitemap parsing itself and something better addressed elsewhere (e.g. by using a disk queue for requests), and that the tests I asked for here are not really necessary.

I will remove the unnecessary tests and merge.

Thanks!

kmike:
great work on this @albertedwardson!

PR #7012, opened 2025-08-18
Improve cleanup in TestEngineCloseSpider.

### LINKED ISSUE #7003: TestSendCatchLogAsyncAsyncio failures
After #6986 or #6997 was merged there are rare failures of `tests/test_utils_signal.py::TestSendCatchLogAsyncAsyncio::test_send_catch_log`, where `LogCapture` captures an unexpected log message emitted by `LogStats`. I've tracked them down to new tests in `tests.test_engine.py::TestEngineCloseSpider` (`test_no_slot()` and `test_exception_signal()`), though in one case I also had a failure linked to some different test so it's possible that fixing those tests won't fix the issue completely.

wRAR:
Hopefully fixes #7003, by letting `LogStats.spider_closed()` run. There may be other failures but I couldn't catch full details of those, let's see how it goes.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -26,7 +26,6 @@ from scrapy.exceptions import CloseSpider, IgnoreRequest
 from scrapy.http import Request, Response
 from scrapy.item import Field, Item
 from scrapy.linkextractors import LinkExtractor
-from scrapy.signals import request_scheduled
 from scrapy.spiders import Spider
 from scrapy.utils.defer import (
     _schedule_coro,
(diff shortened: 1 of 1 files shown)


wRAR:
Unrelated but I think `test_download_with_proxy_https_timeout` is one of the remaining flaky tests, and I wonder whether the problem is not always sufficient timeout or the "[<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', '', 'packet length too long'), ('SSL routines', '', 'record layer failure')]>]" part.

PR #7013, opened 2025-08-21
tests: add unit tests for StatsMailer extension

IkramKhanNiazi:
This PR adds unit tests for the StatsMailer extension to improve coverage.

Changes:
- Added `tests/extensions/test_statsmailer.py`
- Verified that `from_crawler` raises NotConfigured when no recipients are configured
- Ensured that an email is sent on spider close when recipients are provided
- Checked that both global stats and spider-specific stats are included in the email body

This increases coverage for `scrapy/extensions/statsmailer.py`.

### DIFF
--- a/tests/test_extension_statsmailer.py
+++ b/tests/test_extension_statsmailer.py
@@ -19,7 +19,7 @@ class DummySpider(Spider):
 def dummy_stats():
     class DummyStats(StatsCollector):
         def __init__(self):
-            super().__init__(crawler=None)
+            # pylint: disable=super-init-not-called
             self._stats = {"global_item_scraped_count": 42}
 
         def get_stats(self, spider=None):
--- a/tests/test_extension_statsmailer.py
+++ b/tests/test_extension_statsmailer.py
@@ -7,12 +7,8 @@ from scrapy.exceptions import NotConfigured
 from scrapy.extensions import statsmailer
 from scrapy.mail import MailSender
 from scrapy.signalmanager import SignalManager
-from scrapy.spiders import Spider
 from scrapy.statscollectors import StatsCollector
-
-
-class DummySpider(Spider):
-    name = "dummy_spider"
+from scrapy.utils.spider import DefaultSpider
 
 
 @pytest.fixture
@@ -23,9 +19,7 @@ def dummy_stats():
             self._stats = {"global_item_scraped_count": 42}
 
         def get_stats(self, spider=None):
-            if spider:
-                return {"item_scraped_count": 10}
-            return self._stats
+            return {"item_scraped_count": 10, **self._stats}
 
     return DummyStats()
 
@@ -39,15 +33,14 @@ def test_from_crawler_without_recipients_raises_notconfigured():
         statsmailer.StatsMailer.from_crawler(crawler)
 
 
-def test_from_crawler_with_recipients_registers_signal(dummy_stats):
+def test_from_crawler_with_recipients_initializes_extension(dummy_stats, monkeypatch):
     crawler = MagicMock()
     crawler.settings.getlist.return_value = ["test@example.com"]
     crawler.stats = dummy_stats
     crawler.signals = SignalManager(crawler)
 
     mailer = MagicMock(spec=MailSender)
-    monkeypatch_mail = MagicMock(return_value=mailer)
-    statsmailer.MailSender.from_crawler = monkeypatch_mail
+    monkeypatch.setattr(statsmailer.MailSender, "from_crawler", lambda _: mailer)
 
     ext = statsmailer.StatsMailer.from_crawler(crawler)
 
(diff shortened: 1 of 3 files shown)


tests/test_statsmailer.py
wRAR, line 1: can you please rename it to test_extension_statsmailer?

IkramKhanNiazi:
@wRAR Could you please review the PR? Thanks.

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,77 @@
...
+
+def test_from_crawler_with_recipients_registers_signal(dummy_stats):
+    crawler = MagicMock()
+    crawler.settings.getlist.return_value = ["test@example.com"]
+    crawler.stats = dummy_stats
+    crawler.signals = SignalManager(crawler)
+
+    mailer = MagicMock(spec=MailSender)
+    monkeypatch_mail = MagicMock(return_value=mailer)
+    statsmailer.MailSender.from_crawler = monkeypatch_mail
```
wRAR: Is this patch undone automatically?

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,77 @@
...
+from scrapy import signals
+from scrapy.exceptions import NotConfigured
+from scrapy.extensions import statsmailer
+from scrapy.mail import MailSender
+from scrapy.signalmanager import SignalManager
+from scrapy.spiders import Spider
+from scrapy.statscollectors import StatsCollector
+
+
+class DummySpider(Spider):
```
wRAR: You can use `scrapy.utils.spider.DefaultSpider`

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.fixture
+def dummy_stats():
+    class DummyStats(StatsCollector):
+        def __init__(self):
+            # pylint: disable=super-init-not-called
+            self._stats = {"global_item_scraped_count": 42}
+
+        def get_stats(self, spider=None):
+            if spider:
```
wRAR: The spider arg was recently deprecated and is no longer passed.

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,77 @@
...
+def test_from_crawler_without_recipients_raises_notconfigured():
+    crawler = MagicMock()
+    crawler.settings.getlist.return_value = []
+    crawler.stats = MagicMock()
+
+    with pytest.raises(NotConfigured):
+        statsmailer.StatsMailer.from_crawler(crawler)
+
+
+def test_from_crawler_with_recipients_registers_signal(dummy_stats):
```
wRAR: What does this test?

tests/test_extension_statsmailer.py
IkramKhanNiazi: We need to accept the spider argument in our DummyStats.get_stats, but ignore it. That way the test will work across different versions. Gives an error in case spider arg is not passed.

tests/test_extension_statsmailer.py
IkramKhanNiazi: We need to accept the spider argument in our DummyStats.get_stats, but ignore it. That way the test will work across different versions. Gives an error in case spider arg is not passed.

IkramKhanNiazi:
@wRAR I’ve made the necessary changes. Could you please review the PR ?
Thanks

tests/test_extension_statsmailer.py
wRAR: > That way the test will work across different versions

Versions of what?

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@pytest.fixture
+def dummy_stats():
+    class DummyStats(StatsCollector):
+        def __init__(self):
+            # pylint: disable=super-init-not-called
+            self._stats = {"global_item_scraped_count": 42}
+
+        def get_stats(self, spider=None):
```
wRAR: As mentioned earlier, the spider arg shouldn't be needed.

tests/test_extension_statsmailer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@pytest.fixture
+def dummy_stats():
+    class DummyStats(StatsCollector):
+        def __init__(self):
+            # pylint: disable=super-init-not-called
+            self._stats = {"global_item_scraped_count": 42}
+
+        def get_stats(self, *args, **kwargs):
```
wRAR: `args` and `kwargs` aren't used here.

wRAR (changes requested):

tests/test_extension_statsmailer.py
IkramKhanNiazi: @wRAR 
If no arguments or the spider argument are provided, an error is raised due to [this line](https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/statsmailer.py?utm_source=chatgpt.com#L43)
 in the spider_closed method. Please let me know if I’ve missed anything.

Screenshot 2025-08-30 at 1 13 31 AM

wRAR: Right, the warning for that line wasn't caught because that line wasn't covered by tests. You should remove the spider arg passing from that line.

IkramKhanNiazi: I've removed the spider arg passing from that line.

wRAR (approved):

wRAR:
Thanks!

Gallaecio (approved):

PR #7014, opened 2025-08-22
RAG use case with Embeddings LiteLLM Pipeline and S3 Vectors Storage

kyleissuper:
Proposal for RAG use cases:
- Add Pipeline that takes scraped page and embeds it using LiteLLM, thus supporting various embedding models
- Add [S3 Vectors](https://aws.amazon.com/s3/features/vectors/) storage feed
- Add `jsonl-pinecone` format (id, values, metadata) as a simple format+schema to simplify how embedding pipeline and vector storage are built / used

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -117,6 +117,18 @@ def requires_boto3(request):
         pytest.skip("boto3 is not installed")
 
 
+@pytest.fixture(autouse=True)
+def requires_litellm(request):
+    if not request.node.get_closest_marker("requires_litellm"):
+        return
+    try:
+        import litellm  # noqa: PLC0415
+
+        del litellm
+    except ImportError:
+        pytest.skip("litellm is not installed")
+
+
 def pytest_configure(config):
     if config.getoption("--reactor") == "asyncio":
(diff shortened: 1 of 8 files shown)


wRAR:
This should be a separate project (or two? not sure if these two parts are related in any way).

kyleissuper:
Thanks for the lightning fast review!

Submitted this draft PR because I'd prefer to have RAG-related functionality to already exist when I use scrapy,  but it felt like a gray area. Gray because scrapy seems to come with stuff like S3 batteries-included, but adding a new built in pipeline felt like it could go either way.

Will think about how to best build this separately then 🤔

wRAR:
> Gray because scrapy seems to come with stuff like S3 batteries-included

Yeah, I understand that there is no clear criterion on what would we be fine with including.

> Will think about how to best build this separately then 🤔

I think it should be relatively straightforward: you make a Python package that includes necessary classes and suggest in the docs what should the user put into settings to make use of them.

kyleissuper:
Thanks for the wise guidance, am super happy about pushing my first package! https://github.com/kyleissuper/scrapy-vectors

Issue #7002, opened 2025-08-09
Cover `scrapy.extensions.memusage.MemoryUsage` with tests


wRAR:
I think it should be possible to write tests for `MemoryUsage` by mocking `self.get_virtual_size()` or (less likely) `resource.getrusage`.

Lions-1:
Hello, Can I work on this one please

wRAR:
@Lions-1 sure, you don't need a confirmation for that

meet-vasita:
Hi, I’d like to work on this issue as my first contribution to the scrapy repo.

wRAR:
@meet-vasita note that there are open PRs for it already.

luca-c-coll:
hey @wRAR, which "open" issue is actually available for fixing and committing changes. if i may get guidance on finding a issue that will be great. it's for a open source project requiring code related issues solved and committed. thank you

wRAR:
@luca-c-coll please follow https://docs.scrapy.org/en/latest/contributing.html#finding-work

sairamsharan:
Is it still available?

wRAR:
@sairamsharan https://github.com/scrapy/scrapy/issues/7002#issuecomment-3240262771

Mohsalam:
Hi! I’m interested in working on this issue and can start investigating right away

mosessamofficial-sys:
Hi @wRAR ,i want to work on this issue.

wRAR:
@mosessamofficial-sys sure

Sowmya-Dadheech:
Hi, I'd like to work on this issue. I plan to write tests for MemoryUsage by mocking `self.get_virtual_size() ` as suggested. Could you please assign this to me?

wRAR:
@Sowmya-Dadheech please read and follow https://docs.scrapy.org/en/latest/contributing.html

Sanjays2402:
Hey — I'd like to take a crack at this. Planning to submit a PR in the next couple days. Will comment back if I hit anything unexpected.

faccinib:
Hi, I see that several other developers have already worked in this issue, as a beginner in scrapy and open sources projects. I would like to work on this.
I'll be reading the contributing page and will come back with my contribution. (feel free to contact me if needed)
I'll be contributing as a student at Université Paris-Cité.

celin-mampilly:
[...]

Issue #6916, opened 2025-06-26
Refactor engine starting and stopping flow


wRAR:
Both the engine initialization and finalization are complicated multi-stage processes that also each have several separate top-level functions to call (`start()`/`start_async()` and `open_spider()`/`open_spider_async()`; `stop()`, `close()` and `close_spider()`). All of the involved code is also asynchronous. In most production runs that make a request this works fine and in a valid order, but in some cases, including early spider shutdown for various reasons (which are valid use cases) and early exceptions due to bugs (which is not an expected case but still should behave within certain expected standards) this breaks down with various outcomes, ranging from unexpected log messages to the spider process hanging. Over the years we identified and patched some of these, and wrote tests for some of them, but we don't test everything and even existing tests often depend on timing and aren't reliable.

A basic example of the problem is shutting down the spider via an early `self.crawler.engine.close_spider()` made in user code or in the `CloseSpider` extension before the starting up code is finished, leading to parts of `start()` running after some or all parts of `stop()` and `close_spider()`, which in turns may lead to e.g. hanging or (cascades of) unhandled AttributeErrors.
Ideally this flow should be made robust, so that at least the code never hangs, never swallows exceptions or bubbles up unhandled ones, and allows further modifications without an easy way to accidentally break it again. I feel like some state machine-based approach can help.

Some of the related issues are #1523 (there is important analysis there, that `engine_stopped` is always later than `spider_closed` but `engine_started` and `spider_opened` have different behavior in the shell than in normal spiders), #3435, #3343.

wRAR:
The `scrapy.shell` code that deals with this is also dirty, calling private methods directly and otherwise doing things not expected by `ExecutionEngine` code itself, it would be nice if it didn't do that, even if it needs some additional public methods in `ExecutionEngine` only used by it.

LuminaX-alt:
Currently, lifecycle is managed procedurally with async calls → leads to inconsistent states if interrupted.

IDLE → STARTING → RUNNING → STOPPING → STOPPED
Each top-level method (start, stop, close_spider, etc.):

Checks & updates state atomically.

Queues or ignores invalid operations.

Ensure all async callbacks respect state transitions:

e.g., if in STOPPING, skip any new spider-opening logic.

Wrap all lifecycle async code in try/except + logging to prevent swallowed exceptions.
This approach:

Prevents start/stop overlapping chaos.

Makes behavior predictable.

Simplifies debugging.
Audit current engine lifecycle code:

[...]

Issue #3343, opened 2018-07-19
Memusage extension AssertionError


nadzimo:
I was testing the default configuration of the `scrapy.extensions.memusage.MemoryUsage` extension which is enabled by default. In `settings.py` I set `MEMUSAGE_LIMIT_MB = 10`, a low number to test it out so it gets triggered.
Then I run `scrapy crawl prettylittlething` and the process gives an `AssertionError`. The extension keeps checking every minute the memory usage and repeats the message `ERROR: Memory usage exceeded 8M. Shutting down Scrapy...`

```
2018-07-19 19:05:53 [scrapy.extensions.memusage] ERROR: Memory usage exceeded 8M. Shutting down Scrapy...
2018-07-19 19:05:53 [scrapy.core.engine] INFO: Closing spider (memusage_exceeded)
2018-07-19 19:05:53 [prettylittlething] INFO: Spider closed: prettylittlething because memusage_exceeded
2018-07-19 19:05:53 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'finish_reason': 'memusage_exceeded',
 'finish_time': datetime.datetime(2018, 7, 19, 15, 5, 53, 236830),
 'log_count/DEBUG': 1,
 'log_count/ERROR': 1,
 'log_count/INFO': 10,
 'memusage/limit_reached': 1,
 'memusage/max': 78512128,
 'memusage/startup': 78512128,
 'start_time': datetime.datetime(2018, 7, 19, 15, 5, 53, 235177)}
2018-07-19 19:05:53 [scrapy.core.engine] INFO: Spider closed (memusage_exceeded)
Unhandled error in Deferred:
2018-07-19 19:05:53 [twisted] CRITICAL: Unhandled error in Deferred:

2018-07-19 19:05:53 [twisted] CRITICAL: 
Traceback (most recent call last):
  File "/Users/nadima/Dropbox/Local-backup/fashion-app/app/venv-app/lib/python3.6/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/Users/nadima/Dropbox/Local-backup/fashion-app/app/venv-app/lib/python3.6/site-packages/scrapy/core/engine.py", line 85, in stop
    assert self.running, "Engine not running"
AssertionError: Engine not running
2018-07-19 19:06:53 [scrapy.extensions.memusage] ERROR: Memory usage exceeded 8M. Shutting down Scrapy...
[...]

PR #7017, opened 2025-08-22
tests: integration coverage for memusage extension 

Lions-1:
Fixes #7002

Adds black-box tests for scrapy.extensions.memusage.MemoryUsage:

- Run a small crawl with CrawlerRunner
- Mock the public MemoryUsage.get_virtual_size() to simulate memory
- Assert spider finish reason via stats (memusage_exceeded / finished)
- Assert the extension’s log messages


Lions-1:
**What this adds**
- Integration tests for `scrapy.extensions.memusage.MemoryUsage` that:
  - assert shutdown with `finish_reason="memusage_exceeded"` when limit is crossed
  - assert warning log and normal finish when only warning threshold is crossed
- Tests run as black-box: no internal/private methods asserted.
- Skip on Windows (memusage depends on `resource`).
- Pin `TWISTED_REACTOR` to the already-installed reactor at runtime to avoid CI matrix mismatches.

**Why**
- Prior CI failures came from reactor mismatches across jobs and platform-specific behavior.
- This approach keeps tests portable across the CI matrix (Py 3.9–3.13, PyPy, default/pinned reactors).

**Notes for reviewers**
- Log assertions use `record.getMessage().lower()` for portability across handlers.
- Happy to tweak loops/intervals if you prefer tighter timings.


@maintainers Could you please re-run CI for this PR?
The previous failures were due to reactor mismatches. The new commit pins the test reactor to the installed one and should pass across the matrix.

tests/test_memusage_integration.py
```diff
@@ -0,0 +1,135 @@
...
+
+    MB = 1024 * 1024
+    state = {"high": False}
+
+    def fake_vsz(self):
+        return 250 * MB if state["high"] else 5 * MB
+
+    monkeypatch.setattr(MemoryUsage, "get_virtual_size", fake_vsz)
+
+    runner = CrawlerRunner(settings)
```
wRAR: Is `CrawlerRunner` needed? Normally just `Crawler` is enough. It would also remove the need for manual setting of `TWISTED_REACTOR` if `get_crawler()` is used.

tests/test_memusage_integration.py
```diff
@@ -0,0 +1,135 @@
...
+    yield runner.crawl(crawler, url=url, loops=100)
+
+    assert crawler.stats.get_value("finish_reason") == "memusage_exceeded"
+    assert any(
+        "memory usage exceeded" in r.getMessage().lower() for r in caplog.records
+    )
+
+
+@inlineCallbacks
+@pytest.mark.twisted
```
wRAR: Is this needed?

tests/test_memusage_integration.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def parse(self, response):
+        count = response.meta.get("count", 0)
+        if count + 1 < self.loops:
+            yield response.follow(
+                self.url, callback=self.parse, meta={"count": count + 1}
+            )
+
+
+def _tmp_file_uri(tmp_path: Path) -> str:
```
wRAR: You should be able to use `"data:,"` as a dummy URL instead.

tests/test_memusage_integration.py
```diff
@@ -0,0 +1,135 @@
...
+    crawler = runner.create_crawler(_LoopSpider)
+
+    caplog.set_level(logging.WARNING, logger="scrapy.extensions.memusage")
+    yield runner.crawl(crawler, url=url, loops=60)
+
+    assert crawler.stats.get_value("finish_reason") == "finished"
+    assert any(
+        ("memory usage reached" in r.getMessage().lower())
+        or ("memory usage warning" in r.getMessage().lower())
+        or ("warning: memory usage reached" in r.getMessage().lower())
```
wRAR: the third one includes the first one

tests/test_memusage_integration.py
Lions-1: You are right, this is a lot faster.

tests/test_memusage_integration.py
Lions-1: I reached for it out of habit, you are right

Lions-1:
@wRAR Thanks for the review! I’ve addressed the points

Lions-1:
The previous failures on the 2 jobs came from a reactor mismatch: the job installed EPollReactor, while the test implicitly expected AsyncioSelectorReactor. By pinning TWISTED_REACTOR to the already installed reactor class path, the test honors each CI job’s choice and no longer requests a conflicting reactor. happy to tweak it to something you prefer

wRAR:
Again, that's not needed when `get_crawler()` is used.

Yor problems seem to be caused by passing a `Settings` object as the `settings_dict` arg to it, which loads default settings.

Lions-1:
Hey  @wRAR , CI failures look caused by a stray ERROR log from memusage’s background LoopingCall firing after the memusage tests finish, which then pollutes unrelated logging tests, im thinking of monkeypatch scrapy.extensions.memusage.create_looping_call in these 2 tests to a one-shot stub that runs the function immediately and doesn’t schedule periodic callbacks. 
Is this approach acceptable ?

wRAR:
I would expect `engine_stopped` to fire and stop the looping calls, does that not happen?

Lions-1:
It does, but now I see the failure came from a last-tick race: memusage can log one final ERROR as the engine is stopping, after the test’s caplog context, which then leaks into other logging tests.
I’ve updated the two memusage integration tests to monkeypatch MemoryUsage.engine_started to run a single synchronous pass and not start any LoopingCalls. Ill see how this one goes

Lions-1:
@wRAR  after switching to get_crawler({...}) (no Settings object) and making the memusage checks synchronous (no LoopingCall left running), CI is down to a 1 single failure on default-reactor-pinned and default-reactor :
tests/test_spider_start.py::TestMain::test_start_deprecated_super.

From the job logs, that test’s LogCapture picks up extra INFO records during its own mini crawl (e.g. CoreStats “Dumping Scrapy stats” and Engine “Spider closed (finished)”), so its “exactly one record” assertion fails. I didn’t change global logging or handlers, and the memusage tests don’t start background timers anymore, so this looks localized to that test.

Do you see anything obvious I’m missing about how that test isolates logs?

wRAR:
I've checked the code @ 62a133d and the actual reason for stray log messages is that `engine_stopped` is not called (in the `default-reactor` env). I'll investigate why does that happen.

wRAR:
> It does

It indeed doesn't.

I'm pretty sure it's an instance of #6916. It may even be related to #3343, though I can't reproduce that one on master with both reactors, but my old analysis from there about firing `engine_stopped` before `ExecutionEngine.start_async()` has moved to the `self.running = True` line applies (on the default reactor, as with asyncio we have lazy coroutine scheduling).

I think fixing this underlying issue is definitely out of scope for this PR, not sure what's the ideal way to work around it though, as the latest change in this PR makes `MemoryUsage.engine_started()` not covered. The previous one, with `_OneShotLoop` may be better if it works, can you please focus on that if you think it's viable?

> From the job logs, that test’s LogCapture picks up extra INFO records

The failure is unrelated to logs though, it's about Python warnings. The reason for those is `TelnetConsole.stop_listening()` not being called, because of the same `engine_stopped` problem, and it's easy to fix by adding `"TELNETCONSOLE_ENABLED": False` to settings in both of your tests.

Also, can you please rename the test file to `test_extension_memusage` and try adding a test for the `self.notify_mails == True` code paths, mocking `self.mail.send()`?

Thanks.

N-T-Raghava:
Thanks for the extensive integration coverage here.

I’ve opened a separate PR that adds focused unit-level tests for remaining
paths not covered by integration tests, including:
- disabled configuration (limits=None),
- notify_mail configuration,
- and safe coverage of engine_started without executing callbacks.

My PR:  (https://github.com/scrapy/scrapy/pull/7181)

This keeps the scope narrow and avoids reactor/timing issues while completing
coverage for #7002.

Mohsalam:
Hi I'd love to work on this issue and can start investigating right away.

wRAR (approved):

PR #7023, opened 2025-08-23
Fix: log CloseSpider error messages when raised (#3463)

### LINKED ISSUE #3463: raise CloseSpider("Error Message") is not saving error message

RohanVH:
This PR fixes #3463.

- Modified handle_spider_error to log CloseSpider exceptions with their custom messages

✅ Backward compatible
✅ Includes news fragment (docs/news/3463.bugfix)

### DIFF
--- a/docs/news/3463.bugfix
+++ b/docs/news/3463.bugfix
@@ -1,0 +1,1 @@
+Fixed CloseSpider error messages not being logged when raised (#3463) 


RohanVH:
Hi! This PR is ready for review. It’s split from the earlier combined PR. CI shows “workflows awaiting approval” because this is from a fork.
please approve workflows when convenient. Thanks!

Issue #7025, opened 2025-08-23
Discourage using `CrawlSpider`


wRAR:
I've seen a lot of user questions about `CrawlSpider`, most of them caused by its limitations. I'm biased but when asked I awlays reply "please consider not using it". In my experience and opinion, the control flow in a `CrawlSpider` spider is both confusing by itself and too different from one in a normal spider that it should be only recommended for very basic spiders with no extensibility requirements.

We currently have this text from 2009 in [the CrawlSpider doc](https://docs.scrapy.org/en/latest/topics/spiders.html#crawlspider):

> This is the most commonly used spider for crawling regular websites, as it provides a convenient mechanism for following links by defining a set of rules. It may not be the best suited for your particular web sites or project, but it’s generic enough for several cases, so you can start from it and override it as needed for more custom functionality, or just implement your own spider.

As a minimally needed change, I think the first half shouldn't say anything like "the most commonly used" and the second half should more prominently say that you don't need to use `CrawlSpider` and you should switch to `Spider` if you run into limitations of `CrawlSpider`.

I will also agree with any of the more controversial suggestions, from adding "this isn't recommended to use for new code" and up to extracting it into a separate project or even deprecating and removing.

VARUN3WARE:
@wRAR I’ve just submitted a pull request that updates the CrawlSpider documentation to discourage its use in favor of the regular Spider class. The changes aim to highlight the limitations of CrawlSpider and encourage users to opt for Spider when more extensibility is needed.

Please review and let me know if any changes are needed.

wRAR:
@VARUN3WARE it's tagged "discuss", it's too early to provide PRs, sorry.

VARUN3WARE:
Sorry about that! I’ll look for another issue that’s ready for a PR and try to contribute there instead. Thanks for letting me know!

staehlmich:
@wRAR I agree with your description. I remember when I first started using Scrapy about 2 years ago, understanding how the CrawlSpider works was quite challenging. Understanding the syntax and logic of the rules and LinkExtractor took me a while. 

What helped me where tutorials and guides on the resources page. But where did that page go on the current site? I could only get access to it by performing a search.

PR #7026, opened 2025-08-24
docs: discourage using CrawlSpider in favor of Spider (#7025)

VARUN3WARE:
This pull request updates the documentation for CrawlSpider to recommend against using it for new projects.

The CrawlSpider class is limited in flexibility and has a control flow that is different from the standard Spider, which can lead to confusion and limitations when attempting to customize or extend its functionality.

Instead, we now encourage users to use the regular Spider class, which offers more flexibility and better extensibility for most use cases.

This change also highlights that CrawlSpider is best suited only for basic crawling tasks where minimal customization is needed.

Let me know if you'd like to adjust the description or need more information!

### DIFF
--- a/docs/topics/spiders.rst
+++ b/docs/topics/spiders.rst
@@ -427,11 +427,14 @@ CrawlSpider
 
 .. class:: CrawlSpider
 
-   This is the most commonly used spider for crawling regular websites, as it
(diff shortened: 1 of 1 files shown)


wRAR:
Closing according to https://github.com/scrapy/scrapy/issues/7025#issuecomment-3218289174

PR #7032, opened 2025-08-29
Ensure pipeline completion for items yielded from start() (#7029)

### LINKED ISSUE #7029: The spider doesn't wait for pipelines of items produced from start()
This was reported by a user and I didn't check it myself but it looks true from the code.

`ExecutionEngine._process_start_next()` just schedules `self.scraper.start_itemproc_async()` for items produced from `start()`, and so if there are async pipelines, nothing in the Scrapy code requires waiting until they finish before e.g. closing the spider (that's one of the downsides of `_schedule_coro()` as opposed to awaiting). I don't see a simple fix though, we may need to add some machinery to wait for these.

Tanishq1030:
Fixes #7029

This PR ensures that items yielded directly from the `start()` coroutine are properly processed by item pipelines before the spider shuts down. Previously, such items could be dropped or incompletely handled due to premature closure.

Changes:
- Added tracking for pipeline tasks initiated from `start()`
- Updated engine shutdown logic to await pipeline completion
- Added unit test to validate behavior
- Updated documentation for `start()` coroutine usage

All tests pass via `tox -e py`, and coverage includes `PipelineTaskTracker`.

### DIFF
--- a/scrapy/core/scraper.py
+++ b/scrapy/core/scraper.py
@@ -14,6 +14,7 @@ from twisted.python.failure import Failure
 
 from scrapy import Spider, signals
 from scrapy.core.spidermw import SpiderMiddlewareManager
+from scrapy.core.tracker.pipeline_task_tracker import PipelineTaskTracker
 from scrapy.exceptions import (
     CloseSpider,
     DropItem,
@@ -109,6 +110,8 @@ class Scraper:
             crawler.settings["ITEM_PROCESSOR"]
         )
         self.itemproc: ItemPipelineManager = itemproc_cls.from_crawler(crawler)
+
+        self.tracker = PipelineTaskTracker()
         itemproc_methods = [
             "open_spider",
             "close_spider",
@@ -485,7 +488,12 @@ class Scraper:
         return deferred_from_coro(self.start_itemproc_async(item, response=response))
 
     async def start_itemproc_async(
-        self, item: Any, *, response: Response | Failure | None
+        self,
+        item: Any,
+        *,
+        request=None,
+        spider=None,
+        response: Response | Failure | None,
     ) -> None:
         """Send *item* to the item pipelines for processing.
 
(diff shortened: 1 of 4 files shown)


wRAR:
> This PR ensures that items yielded directly from the start() coroutine are properly processed by item pipelines before the spider shuts down.

Can you explain how?

> Updated engine shutdown logic to await pipeline completion

Can you point to this change?

> Updated documentation for start() coroutine usage

Can you point to this change?

Tanishq1030:
> > This PR ensures that items yielded directly from the start() coroutine are properly processed by item pipelines before the spider shuts down.

> Can you explain how?

 Sure, The issue in #7029 was that items yielded directly from the `start()` coroutine weren’t guaranteed to finish pipeline processing before the spider shut down. To fix that, I introduced a `PipelineTaskTracker `that tracks each item passed into the pipeline including those from `start()`. Then, during shutdown, the engine explicitly waits for all tracked tasks to complete. This way, even items yielded early in the crawl lifecycle get fully processed before the spider exits.

>> Updated engine shutdown logic to await pipeline completion

> Can you point to this change?

 I misspoke earlier. The` _close()` coroutine in` scrapy/core/engine.py` wasn’t modified. The key changes are actually in:
- `scrapy/core/scraper.py` — where we hook into the pipeline flow to register tasks.
-` scrapy/core/tracker/pipeline_task_tracker.py` — this is the new tracker that monitors pipeline-bound items.
- `scrapy/core/tracker/__init__.py` — sets up the tracker interface.
- `tests/test_pipeline_task_tracker.py` — added coverage to ensure the tracker behaves as expected.
The tracker ensures that every item passed into the pipeline — including those from `start() `— is accounted for, and the engine waits for all of them to finish before shutdown. Thanks for keeping it accurate!

 
> > Updated documentation for start() coroutine usage

> Can you point to this change?

Not yet,  I mentioned a doc update for start() coroutine behavior, but that hasn’t landed. The plan is to update `docs/topics/spiders.rst` to clarify that items yielded from start() are now properly tracked through pipeline completion. That’ll help future users understand the change without digging through the code. I’ll follow up with that commit shortly.

PR #7033, opened 2025-08-29
Extract @_warn_spider_arg.

wRAR:
(in preparation for downloader middleware changes, separate PR mostly to check the coverage changes)

It turns out that even simple decorators that support sync and async functions are very unwieldy :-/

### DIFF
--- a/scrapy/utils/decorators.py
+++ b/scrapy/utils/decorators.py
@@ -91,12 +91,11 @@ def _warn_spider_arg(
     | Callable[_P, Coroutine[Any, Any, _T]]
     | Callable[_P, AsyncGenerator[_T]]
 ):
-    """Decorator to warn if a (non-None) ``spider`` argument is passed to a function."""
+    """Decorator to warn if a ``spider`` argument is passed to a function."""
 
     def check_args(*args: _P.args, **kwargs: _P.kwargs) -> None:
         bound = inspect.signature(func).bind(*args, **kwargs)
-        bound.apply_defaults()
-        if bound.arguments.get("spider"):
+        if "spider" in bound.arguments:
             warnings.warn(
                 f"Passing a 'spider' argument to {func.__qualname__}() is deprecated and "
                 "the argument will be removed in a future Scrapy version.",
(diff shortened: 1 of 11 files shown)


scrapy/utils/decorators.py
```diff
@@ -65,3 +66,68 @@ def wrapped(*a: _P.args, **kw: _P.kwargs) -> Deferred[_T]:
...
+        bound.apply_defaults()
+        if bound.arguments.get("spider"):
+            warnings.warn(
+                f"Passing a 'spider' argument to {func.__qualname__}() is deprecated and "
+                "the argument will be removed in a future Scrapy version.",
+                category=ScrapyDeprecationWarning,
+                stacklevel=3,
+            )
+
+    if inspect.iscoroutinefunction(func):
```
wRAR, line 106: OK, this is not used/covered because the only async functions that are decorated are `process_spider_output_async()` which are async generator functions.

scrapy/utils/decorators.py
```diff
@@ -65,3 +66,68 @@ def wrapped(*a: _P.args, **kw: _P.kwargs) -> Deferred[_T]:
...
+    Callable[_P, _T]
+    | Callable[_P, Coroutine[Any, Any, _T]]
+    | Callable[_P, AsyncGenerator[_T]]
+):
+    """Decorator to warn if a (non-None) ``spider`` argument is passed to a function."""
+
+    def check_args(*args: _P.args, **kwargs: _P.kwargs) -> None:
+        bound = inspect.signature(func).bind(*args, **kwargs)
+        bound.apply_defaults()
+        if bound.arguments.get("spider"):
```
Gallaecio: I saw that `(non-None)` in the docstring above, and I’m thinking: What if, now that we are using inspect anyway, we go a step further than with the previous code, and warn as well if the value if `None` provided it is specified as a parameter? (since such code would also break when we remove the parameter).

e.g.

```suggestion
        if "spider" in bound.arguments:
```

Gallaecio (approved):

PR #7036, opened 2025-09-01
Reducing memory footprint of `Request` object: __slots__ and lazy evaluation

albertedwardson:
This PR makes such changes:

1. Python’s `dict` objects are fairly heavy, even when empty. The `Request` object had a few of them. This PR makes their creation lazy (along with the `flags` list), so they are only created when they actually hold data. Previously, they were also created lazily, but once accessed, the `Request` object would keep references to empty `dict`s/`list`s. ~~In practice, these could just be `None` instead.~~ No, this breaks things. E.g. `r.cookies[k]=v` will return newly created empty dict, assign `v` to `k` key, and then discard this dict object since there is no references to it. Once accessed, they will be created and keeped. But when trying to assign empty dict or list to some of `Request` attributes, we won't store them.

2. Added `__slots__` to `Request` class and its subclasses

Before
```python
Python 3.13.3 (main, May 22 2025, 01:58:53) [MSC v.1943 64 bit (AMD64)] on win32
>>> from pympler.asizeof import asizeof
>>> from scrapy import Request
>>> asizeof(Request("http://example.com"))
1480
>>> asizeof(Request("http://example.com"))
1472
>>> asizeof(Request("http://example.com"))
>>> ... # internal python magic/interning of some objects???
>>> asizeof(Request("http://example.com"))
1368 # final size
```

After
```python
Python 3.13.3 (main, May 22 2025, 01:58:53) [MSC v.1943 64 bit (AMD64)] on win32
>>> from pympler.asizeof import asizeof
>>> from scrapy import Request
>>> r = Request("http://example.com")
>>> asizeof(r)
432
>>> r.cookies  # sets empty dict
{}
>>> asizeof(r)
496
>>> r.cookies = {} # sets None
>>> asizeof(r)
432
```

### DIFF
--- a/tests/test_http_request.py
+++ b/tests/test_http_request.py
@@ -329,6 +329,89 @@ class TestRequest:
                 errback="a_function",
             )
 
+    def test_setters(self):
+        request = self.request_class("http://example.com")
+
+        request.cb_kwargs = {"a": 1}
+        assert request.cb_kwargs == {"a": 1}
+
+        request.meta = {"k": "v"}
+        assert request.meta == {"k": "v"}
+
+        request.flags = ["f1"]
+        assert request.flags == ["f1"]
+
+        request.cookies = {"sid": "1"}
+        assert request.cookies == {"sid": "1"}
+
+        headers = Headers({b"X-Test": b"1"})
+        request.headers = headers
+        assert request._headers is headers
+        request.headers = {b"A": b"b"}
+        assert isinstance(request.headers, Headers)
+        assert request._headers[b"A"] == b"b"
+
+    def test_setter_mutable_lazy_loading(self):
+        """Mutable attributes are set internally to None only until they are
+        read, then they always return the same falsy instance of the
+        corresponding mutable structure.
+
+        Setting them to None causes the next read to return a different object.
+        """
+
+        request = self.request_class("http://example.com")
+
+        assert request._cb_kwargs is None
+        assert request.cb_kwargs == {}
+        assert request.cb_kwargs is request.cb_kwargs
+        assert request._cb_kwargs == {}
+        original_cb_kwargs = request.cb_kwargs
+        request.cb_kwargs = None
+        assert request.cb_kwargs == {}
+        assert request.cb_kwargs is not original_cb_kwargs
+
+        assert request._meta is None
+        assert request.meta == {}
+        assert request.meta is request.meta
+        assert request._meta == {}
+        original_meta = request.meta
+        request.meta = None
+        assert request.meta == {}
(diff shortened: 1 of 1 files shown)


wRAR:
I think this broke tests.

albertedwardson:
Oh, sorry, my bad. When I tested locally, I assumed the hanging test case was broken because I was running tests on my work PC with a weird Windows setup and a custom-compiled Python build

scrapy/http/request/__init__.py
```diff
@@ -177,7 +193,7 @@ def __init__(
         #:
         #: .. seealso::
         #:     :ref:`topics-request-response-ref-request-callback-arguments`
-        self.callback: CallbackT | None = callback
+        self.callback: CallbackT | None = None if callback is NO_CALLBACK else callback
```
albertedwardson: note for myself: i think this broke the flow

scrapy/http/request/__init__.py
```diff
@@ -186,10 +202,14 @@ def __init__(
         #: its first parameter.
         #:
         #: .. seealso:: :ref:`topics-request-response-ref-errbacks`
-        self.errback: Callable[[Failure], Any] | None = errback
+        self.errback: Callable[[Failure], Any] | None = (
```
albertedwardson: and this

albertedwardson:
All unit tests of "request" objects now passing. But I didn't ran full suite locally, can't do it now.

albertedwardson:
got #7003 in https://github.com/scrapy/scrapy/actions/runs/17397878619/job/49386271260?pr=7036 and https://github.com/scrapy/scrapy/actions/runs/17397878619/job/49386271264?pr=7036 test runs

albertedwardson:
@wRAR is there any point for writing tests for setters?

wRAR:
> is there any point for writing tests for setters?

Yeah, I think it's useful to make sure the attributes can be assigned, for these and future changes.

scrapy/http/request/__init__.py
```diff
@@ -206,20 +219,28 @@ def __init__(
...
+
     @property
     def meta(self) -> dict[str, Any]:
         if self._meta is None:
             self._meta = {}
         return self._meta
 
+    @meta.setter
+    def meta(self, value: dict[str, Any] | None) -> None:
+        self._meta = value if value else None
```
AmirRedHat, line 243: I think this line can be like `self._meta = value or None`.
Becouse `{} or None -> None`, so zero-value of dict can be ignored.

wRAR:
Thanks!

PR #7037, opened 2025-09-01
Deprecate the spider arg of downloader middleware methods.

### LINKED ISSUE #6927: Deprecate and remove the `spider` argument in middleware and pipeline methods

wRAR:
Fixes #6927

### DIFF
--- a/scrapy/downloadermiddlewares/offsite.py
+++ b/scrapy/downloadermiddlewares/offsite.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 
 from scrapy import Request, Spider, signals
 from scrapy.exceptions import IgnoreRequest
+from scrapy.utils.decorators import _warn_spider_arg
 from scrapy.utils.httpobj import urlparse_cached
 
 if TYPE_CHECKING:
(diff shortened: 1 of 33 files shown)


scrapy/downloadermiddlewares/offsite.py
```diff
@@ -21,29 +22,34 @@
...
         assert crawler.stats
         o = cls(crawler.stats)
         crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
         crawler.signals.connect(o.request_scheduled, signal=signals.request_scheduled)
+        o.crawler = crawler
         return o
 
-    def __init__(self, stats: StatsCollector):
-        self.stats = stats
-        self.domains_seen: set[str] = set()
```
Gallaecio, line 34: For the record, I kind of like the `__init__` after the `from_crawler`, in order of execution :slightly_smiling_face:

scrapy/downloadermiddlewares/robotstxt.py
```diff
@@ -51,17 +52,18 @@ def __init__(self, crawler: Crawler):
...
             return
-        rp = await self.robot_parser(request, spider)
-        self.process_request_2(rp, request, spider)
+        rp = await self.robot_parser(request)
+        self.process_request_2(rp, request)
 
-    def process_request_2(
-        self, rp: RobotParser | None, request: Request, spider: Spider
-    ) -> None:
+    def process_request_2(self, rp: RobotParser | None, request: Request) -> None:
```
Gallaecio, line 66: I see we are not keeping backward compatibility here or in `robot_parser` below. Was that a deliberate decision?

wRAR: Somewhat, we already document in "Backward-incompatible changes" that `process_request()` and `robot_parser()` were changed to `async def` but I didn't think about `process_request_2()` too much. We can do it for it or even for both helpers.

Gallaecio: No strong opinion, up to you.

Gallaecio (approved):

Issue #7038, opened 2025-09-02
Deprecate spider attributes that can be replaced by settings, round 2


wRAR:
Continuing #6988, there are more occurrences of the same issue:

* [`download_maxsize` and `download_warnsize`](https://github.com/scrapy/scrapy/blob/master/scrapy/core/downloader/handlers/http11.py#L98-L99)
* [`download_maxsize` and `download_warnsize`](https://github.com/scrapy/scrapy/blob/master/scrapy/core/http2/protocol.py#L198-L203)
* [`download_timeout`](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/downloadtimeout.py#L33)
* [`user_agent`](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/useragent.py#L31)
* [`download_delay`](https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/throttle.py#L52)

There is also `handle_httpstatus_list` without a corresponding setting in `RedirectMiddleware`, I guess we should keep it.

It may be useful to extract the warning code into a function in `scrapy.utils.deprecate` to not repeat yourself.

Gallaecio:
> I guess we should keep it.

Or add a setting for it 🙂

Dharma-09:
@Gallaecio @wRAR Hi there, I would like to work on it!

wRAR:
@Dharma-09 as you can see there is an open PR for this.

JennaShi006:
@wRAR does this mean that you are no longer considering new contributions for this issue?

wRAR:
@JennaShi006 this means that it doesn't make sense for people to work on this when most of the work is already done.

HelloOjasMutreja:
Hi, I am new here, could you guide where can I contribute, I would really appreciate a head-start

wRAR:
@HelloOjasMutreja you should start at https://docs.scrapy.org/en/master/contributing.html, linked at multiple places in the repo UI.

Tyleresch:
Hi! I’d like to work on this for a course project. Plan: add a helper in scrapy.utils.deprecate to emit a warning; wire download_timeout, download_delay, user_agent, download_maxsize, download_warnsize so settings take precedence; add tests for each path. Does that match what you want?

wRAR:
@Tyleresch have you seen the existing PR?

Issue #6660, opened 2025-02-06
(meta-bug?) Better rate-limiting


wRAR:
We wanted to re-think what we want/need about per-request delays, retrying 429 after a set delay etc. There are several related issues and PRs, I'm trying to collect them here.

- https://github.com/scrapy/scrapy/issues/802 (and its open PRs https://github.com/scrapy/scrapy/pull/4980 and https://github.com/scrapy/scrapy/pull/5015)
- https://github.com/scrapy/scrapy/issues/892 (not necessarily related to retrying/throttling)
- https://github.com/scrapy/scrapy/issues/3849 
- https://github.com/scrapy/scrapy/issues/4424 (and its open PR https://github.com/scrapy/scrapy/pull/4462)
- https://github.com/scrapy/scrapy/issues/6476 - somewhat related
- https://github.com/scrapy/scrapy/issues/6597 - also somewhat related

JoseG2003-g:
Hi, can we tackle this issue ?

wRAR:
@JoseG2003-g I don't think it will be possible for you, especially considering that it's at a discussion stage.

wRAR:
After a recent short discussion on this we decided that as the first step we want a way to handle specific responses (mainly the 429 ones but it's likely that any implementation will be more flexible than just that) in a way that throttles their requests. So roughly what is asked for in #4424 but more flexible. We haven't discussed how and where to actually implement this.

GeorgeA92:
On https://github.com/scrapy/scrapy/issues/802 discussed possibility to implement delay like: .. specific per request delay for other value (not from `DOWNLOAD_DELAY`), 

While for implementing something like.. reaction for `retry-after` header value process - it is required to apply delay.. with no request e.g. apply delay for downloader slot. not for individual request (e.g. to not send requests for specified downloader slot for next N seconds and after that continue with regular delay - should have effect even with 0 value of `DOWNLOAD_DELAY`) which looks like `slot.latercall` and `slot.lastseen` options from https://github.com/scrapy/scrapy/issues/802#issuecomment-500245345 (and delay from scheduler based approach from  https://github.com/scrapy/scrapy/pull/5015 is not applicable here)

GeorgeA92:
In this code sample I've implemented custom `delay_slot` method in custom downloader enabled by setting.
For long term solution something like this should be added to downloader logic and called inside.. logic that react to `retry-after` and / or other conditions from RetryMiddleware or called directly from spider if needed.

[...]

PR #7039, opened 2025-09-03
Deprecate spider attributes that can be replaced by settings, round 2

Laerte:
close: #7038

deprecated attributes:
- `max_concurrent_requests` → `CONCURRENT_REQUESTS`
- `download_maxsize` → `DOWNLOAD_MAXSIZE`
- `download_warnsize` → `DOWNLOAD_WARNSIZE`
- `download_timeout` → `DOWNLOAD_TIMEOUT`
- `user_agent` → `USER_AGENT`

### DIFF
--- a/scrapy/core/downloader/handlers/http11.py
+++ b/scrapy/core/downloader/handlers/http11.py
@@ -93,9 +93,9 @@ class HTTP11DownloadHandler:
 
     def download_request(self, request: Request, spider: Spider) -> Deferred[Response]:
         """Return a deferred for the HTTP download"""
-        if hasattr(spider, "download_maxsize"):
+        if hasattr(spider, "download_maxsize"):  # pragma: no cover
             warn_on_deprecated_spider_attribute("download_maxsize", "DOWNLOAD_MAXSIZE")
-        if hasattr(spider, "download_warnsize"):
+        if hasattr(spider, "download_warnsize"):  # pragma: no cover
             warn_on_deprecated_spider_attribute(
                 "download_warnsize", "DOWNLOAD_WARNSIZE"
             )
--- a/scrapy/core/downloader/__init__.py
+++ b/scrapy/core/downloader/__init__.py
@@ -96,7 +96,7 @@ def _get_concurrency_delay(
     if hasattr(spider, "download_delay"):
         delay = spider.download_delay
 
-    if hasattr(spider, "max_concurrent_requests"):
+    if hasattr(spider, "max_concurrent_requests"):  # pragma: no cover
         warn_on_deprecated_spider_attribute(
             "max_concurrent_requests", "CONCURRENT_REQUESTS"
         )
--- a/scrapy/core/http2/protocol.py
+++ b/scrapy/core/http2/protocol.py
@@ -192,9 +192,9 @@ class H2ClientProtocol(Protocol, TimeoutMixin):
 
     def _new_stream(self, request: Request, spider: Spider) -> Stream:
         """Instantiates a new Stream object"""
-        if hasattr(spider, "download_maxsize"):
+        if hasattr(spider, "download_maxsize"):  # pragma: no cover
             warn_on_deprecated_spider_attribute("download_maxsize", "DOWNLOAD_MAXSIZE")
-        if hasattr(spider, "download_warnsize"):
+        if hasattr(spider, "download_warnsize"):  # pragma: no cover
             warn_on_deprecated_spider_attribute(
                 "download_warnsize", "DOWNLOAD_WARNSIZE"
             )
--- a/scrapy/downloadermiddlewares/downloadtimeout.py
+++ b/scrapy/downloadermiddlewares/downloadtimeout.py
@@ -31,7 +31,7 @@ class DownloadTimeoutMiddleware:
         return o
 
     def spider_opened(self, spider: Spider) -> None:
-        if hasattr(spider, "download_timeout"):
+        if hasattr(spider, "download_timeout"):  # pragma: no cover
             warn_on_deprecated_spider_attribute("download_timeout", "DOWNLOAD_TIMEOUT")
         self._timeout = getattr(spider, "download_timeout", self._timeout)
 
--- a/scrapy/downloadermiddlewares/useragent.py
+++ b/scrapy/downloadermiddlewares/useragent.py
@@ -29,7 +29,7 @@ class UserAgentMiddleware:
         return o
 
     def spider_opened(self, spider: Spider) -> None:
-        if hasattr(spider, "user_agent"):
+        if hasattr(spider, "user_agent"):  # pragma: no cover
             warn_on_deprecated_spider_attribute("user_agent", "USER_AGENT")
 
         self.user_agent = getattr(spider, "user_agent", self.user_agent)


scrapy/spidermiddlewares/httperror.py
```diff
@@ -40,9 +40,10 @@ class HttpErrorMiddleware:
 
     def __init__(self, settings: BaseSettings):
         self.handle_httpstatus_all: bool = settings.getbool("HTTPERROR_ALLOW_ALL")
-        self.handle_httpstatus_list: list[int] = settings.getlist(
-            "HTTPERROR_ALLOWED_CODES"
-        )
+        self.handle_httpstatus_list: list[int] = [
+            *settings.getlist("HTTPERROR_ALLOWED_CODES"),
+            *settings.getlist("REDIRECT_ALLOWED_HTTP_CODES"),
+        ]
```
Laerte: We already have `HTTPERROR_ALLOWED_CODES` that override `handle_httpstatus_list` but the setting name is not generic `HTTPERROR_`, we could use on `RedirectMiddleware` instead of creating a new setting.

scrapy/downloadermiddlewares/redirect.py
```diff
@@ -150,11 +154,17 @@ class RedirectMiddleware(BaseRedirectMiddleware):
     def process_response(
         self, request: Request, response: Response, spider: Spider | None = None
     ) -> Request | Response:
+        if hasattr(self.crawler.spider, "handle_httpstatus_list"):
```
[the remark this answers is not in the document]
Laerte: Tbh I was not sure why we not use the `spider` from the function args  on line 165 so I followed what was already used.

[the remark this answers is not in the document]
wRAR: The spider method arg is now deprecated :)

scrapy/spidermiddlewares/httperror.py
wRAR: Right, `handle_httpstatus_list` (the spider attr and the meta key) are mostly for `HttpErrorMiddleware`, `RedirectMiddleware` is their additional user. Similarly for `HTTPERROR_ALLOW_ALL` (the setting) and `handle_httpstatus_all` (the spider attr and the meta key). But out of these `RedirectMiddleware` only uses both meta keys and the `handle_httpstatus_list` spider attr, not using the spider attr and the setting for `all`. This is inconsistent (though I haven't considered what will the spider behavior be if this made consistent).

wRAR:
There are tests that use these newly deprecated attributes that need to be fixed or removed.

For download handler tests that likely means creating the crawler and the handler in the test function like already done in `test_download_allow_data_loss_via_setting()` (I'm still not sure what is the most idiomatic way to get a fixture based on some argument), I haven't checked other tests.

wRAR:
In case you've missed it (as it's only in `extra-deps`), `tests/test_downloader_handler_twisted_http2.py::TestHttps2::test_download_with_maxsize_very_large_file` still produces a warning.

tests/test_extension_throttle.py
```diff
@@ -103,56 +95,6 @@ def test_maxdelay_definition(value, expected):
...
-        ),
-        (
-            AUTOTHROTTLE_START_DELAY + 1.0,
-            UNSET,
-            AUTOTHROTTLE_START_DELAY + 2.0,
-            AUTOTHROTTLE_START_DELAY + 2.0,
-        ),
-    ],
-)
-def test_startdelay_definition(min_spider, min_setting, start_setting, expected):
```
wRAR, line 136: Is this test useless/duplicate without `min_spider`?

Laerte:
> In case you've missed it (as it's only in `extra-deps`), `tests/test_downloader_handler_twisted_http2.py::TestHttps2::test_download_with_maxsize_very_large_file` still produces a warning.

Thanks for flagging, fixed.

tests/test_extension_throttle.py
Laerte: It seems, if you see the only assert is: `assert spider.download_delay == expected` lmk if you want to keep and drop the `min_spider` no sure what we would check in this case.

wRAR: OK, this is stupid :)

The extension sets `spider.download_delay` in `_spider_opened()` (based on `AUTOTHROTTLE_START_DELAY`, `DOWNLOAD_DELAY` and the original value of `spider.download_delay`, just like the params of this test), so that the downloader could use it as the start delay value for newly created slots in `scrapy.core.downloader._get_concurrency_delay()`. So if we want to stop using `spider.download_delay` we need this to work in some other way (and I don't think we can change the `DOWNLOAD_DELAY` setting value here).

@Gallaecio thoughts?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -92,6 +93,13 @@ def from_crawler(cls, crawler: Crawler) -> Self:
 
     def download_request(self, request: Request, spider: Spider) -> Deferred[Response]:
         """Return a deferred for the HTTP download"""
+        if hasattr(spider, "download_maxsize"):
+            warn_on_deprecated_spider_attribute("download_maxsize", "DOWNLOAD_MAXSIZE")
+        if hasattr(spider, "download_warnsize"):
+            warn_on_deprecated_spider_attribute(
+                "download_warnsize", "DOWNLOAD_WARNSIZE"
+            )
```
Gallaecio: I wonder if we should try to find a way to trigger this only once per spider, but no strong opinion. Maybe louder is better :slightly_smiling_face:

scrapy/downloadermiddlewares/redirect.py
```diff
@@ -150,11 +154,17 @@ class RedirectMiddleware(BaseRedirectMiddleware):
...
+            warn_on_deprecated_spider_attribute(
+                "handle_httpstatus_list", "REDIRECT_ALLOWED_HTTP_CODES"
+            )
+
         if (
             request.meta.get("dont_redirect", False)
             or response.status
             in getattr(self.crawler.spider, "handle_httpstatus_list", [])
             or response.status in request.meta.get("handle_httpstatus_list", [])
+            or response.status in self.handle_httpstatus_list
```
Gallaecio: I wonder if we should handle this somewhat differently.

i.e., when both the setting and the spider attribute are defined, the spider attribute, even if deprecated, should be considered an override of the setting value, and the setting value should be ignored.

In fact, I wonder if we should also consider the meta value an override of both the setting and the spider attribute, and hence ignore those if the meta value is set. But I am fine with keeping that out of scope here.

Either way, we should make sure we clarify this in the docs and the release notes.

scrapy/spidermiddlewares/httperror.py
Gallaecio: I don’t like the HTTP errors middleware reading the `REDIRECT_ALLOWED_HTTP_CODES` setting…

I wonder if we should take a step back and try to think through what the ideal “settings and meta API” should be, and then try to move towards it in a backward-compatible way with deprecations where needed.

I feel this should be similar to dont_filter, where you have a generic setting/meta and a specific setting/meta (e.g. allow_offsite). I think the specific should override the generic, i.e. no merging, e.g. if generic is [401] and want to extend with 402 for a specific middleware, you set the specific setting/meta to [401,402], not just [402]. But I am not completely sure.

tests/test_extension_throttle.py
Gallaecio: Puff, headache incoming… :sweat:

Gallaecio: Maybe we should keep download_delay until https://github.com/scrapy/scrapy/issues/6660 is addressed?

wRAR: I'm fine with keeping it. Note that we need to revert a part of #6994 before the next release for this: #7117

wRAR:
I propose merging the not controversial subset of this, keeping the following for the round 3 or separate refactoring:

- all changes related to `handle_httpstatus_list` and `REDIRECT_ALLOWED_HTTP_CODES`
- the new `download_delay` warning and the test removal for that attr

Laerte:
> I propose merging the not controversial subset of this, keeping the following for the round 3 or separate refactoring:
> 
> * all changes related to `handle_httpstatus_list` and `REDIRECT_ALLOWED_HTTP_CODES`
> * the new `download_delay` warning and the test removal for that attr

@wRAR Sorry for letting this sit here for a while. I've been busy with work. I pushed the changes to align with your suggestion, let me know if this looks right.

wRAR:
Yes, thanks! I've also added `pragma: no cover` to code that doesn't have existing tests.

wRAR (approved):

PR #7041, opened 2025-09-03
Add support for loading Scrapy configuration from pyproject.toml

### LINKED ISSUE #7030: Support `pyproject.toml` as a preferred alternative to `scrapy.cfg`
Suggested by @Gallaecio

whyvineet:
Fix #7030

### DIFF
--- a/scrapy/utils/conf.py
+++ b/scrapy/utils/conf.py
@@ -13,6 +13,11 @@ from scrapy.settings import BaseSettings
 from scrapy.utils.deprecate import update_classpath
 from scrapy.utils.python import without_none_values
 
+try:
+    import tomllib  # Python 3.11+
+except ImportError:
+    import tomli as tomllib  # For Python <3.11, tomli should be installed
+
 if TYPE_CHECKING:
     from collections.abc import Collection, Iterable, Mapping, MutableMapping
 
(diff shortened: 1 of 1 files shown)


wRAR:
Have you tried running tests and have you enabled pre-commit following the contribution docs?

scrapy/utils/conf.py
```diff
@@ -13,6 +13,11 @@
 from scrapy.utils.deprecate import update_classpath
 from scrapy.utils.python import without_none_values
 
+try:
+    import tomllib  # Python 3.11+
+except ImportError:
+    import tomli as tomllib  # For Python <3.11, tomli should be installed
```
wRAR, line 19: What if it isn't?

scrapy/utils/conf.py
```diff
@@ -103,6 +108,18 @@ def init_env(project: str = "default", set_syspath: bool = True) -> None:
...
+        with pyproject_path.open("rb") as f:
+            data = tomllib.load(f)
+        # Expect [tool.scrapy] section for Scrapy config
+        scrapy_data = data.get("tool", {}).get("scrapy", {})
+        class TomlConfig:
+            def has_option(self, section, option):
+                return section in scrapy_data and option in scrapy_data[section]
+            def get(self, section, option):
+                return scrapy_data[section][option]
+        return TomlConfig()
```
wRAR, line 122: This is not `ConfigParser` even if it implements some of its API, so at the very least the type hints and the docstring need to be changed, but I think this should be done in some better way.

scrapy/utils/conf.py
```diff
@@ -103,6 +108,18 @@ def init_env(project: str = "default", set_syspath: bool = True) -> None:
 
 def get_config(use_closest: bool = True) -> ConfigParser:
     """Get Scrapy config file as a ConfigParser"""
+    pyproject_path = Path("pyproject.toml")
```
wRAR, line 111: This checks the current directory which is wrong. It also ignores `use_closest`.

scrapy/utils/conf.py
```diff
@@ -103,6 +108,18 @@ def init_env(project: str = "default", set_syspath: bool = True) -> None:
 
 def get_config(use_closest: bool = True) -> ConfigParser:
```
wRAR, line 109: I don't think this is the only function that needs changing, there are several others that look for `scrapy.cfg`.

wRAR:

whyvineet:
Hey @wRAR, thanks for following up. The scope differs from what I initially understood. I don’t think I can take it to the level this feature deserves right now, so I’ll step back.

Please feel free to close the PR or let someone else continue it. Thanks for the guidance so far!

Issue #6996, opened 2025-08-08
Tests that install a root logging handler


wRAR:
Some tests, mostly (only?) in `test_crawler.py`, either call `configure_logging()` directly or create a `CrawlerProcess` instance which does the same, and they don't clean up when finished (there is no cleanup code for this in Scrapy anyway). This seems to cause problems for other tests sometimes, affecting their log checks, and also it has a weird effect on `pytest -s`, where the same logged line appears many times, as if the handler is installed many times, even though `install_scrapy_root_handler()` is intended to uninstall the previous one.

Probably the best way to fix this is to add an `uninstall_scrapy_root_handler()` and call it in tests.

wRAR:
Another similar problem is `Crawler._apply_settings()` adding a handler that is set to be removed on the `engine_stopped` signal, but we call `_apply_settings()` in `get_crawler()`. I feel like we should add a `__del__()` to `Crawler` that additionally tries to clean this up. Or we can make `get_crawler()` a context manager :)

PR #7045, opened 2025-09-08
Improve performance of deprecated spider arg checks.

wRAR:
Merging #7011 significantly increased the test run time, I haven't checked if it translates to a similar increase of long-running spider jobs or not but it should be fixed anyway. I've identified two things that we can fix, it's still not enough though (but it's interesting that handling of specifically `inc_value()` gives most or even all of the performance hit).

### DIFF
--- a/scrapy/statscollectors.py
+++ b/scrapy/statscollectors.py
@@ -28,9 +28,15 @@ class StatsCollector:
         self._crawler: Crawler = crawler
 
     def __getattribute__(self, name):
+        cached_name = f"_cached_{name}"
+        try:
+            return super().__getattribute__(cached_name)
+        except AttributeError:
+            pass
+
         original_attr = super().__getattribute__(name)
 
-        if name in (
+        if name in {
             "get_value",
             "get_stats",
             "set_value",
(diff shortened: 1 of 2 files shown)


scrapy/statscollectors.py
```diff
@@ -28,9 +28,15 @@ def __init__(self, crawler: Crawler):
         self._crawler: Crawler = crawler
 
     def __getattribute__(self, name):
+        cached_name = f"_cached_{name}"
+        try:
+            return super().__getattribute__(cached_name)
```
wRAR, line 33: TIL `hasattr()` "is implemented by calling `getattr(object, name)` and seeing whether it raises an `AttributeError` or not"

wRAR:
OK, I think I know why it's `inc_value()` AND why is the effect proportionally lesser when running smaller number of tests: https://github.com/scrapy/scrapy/issues/6996#issuecomment-3172758995 and multiple `LogCounterHandler` emitting `log_count/*` stats.

wRAR:
#7046 makes this unnecessary by removing huge amounts of `inc_stats()` calls which was thus accidentally made a hot path, but I think these changes are still good.

PR #7046, opened 2025-09-08
Move LogCounterHandler into a new extension

wRAR:
Fixes the problem described in https://github.com/scrapy/scrapy/issues/6996#issuecomment-3172758995 (but not the one in the issue itself).

I think the behavior change makes sense and shouldn't impact real spiders. It now counts fewer startup log messages but it never counted all of them anyway. ~~But the next step in the same direction could be simply an extension, I don't know why should this code be a part of `Crawler`.~~ done

### DIFF
--- a/docs/topics/extensions.rst
+++ b/docs/topics/extensions.rst
@@ -144,9 +144,7 @@ Log Count extension
 .. module:: scrapy.extensions.logcount
    :synopsis: Basic stats logging
 
-.. class:: LogCount
-
-Collect counts of log messages by level into ``log_count/<level>`` stats.
+.. autoclass:: LogCount
 
 Telnet console extension
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -15,6 +15,14 @@ Backward-incompatible changes
     ``True`` when running Scrapy via :ref:`its command-line tool
     <topics-commands-crawlerprocess>` to avoid a reactor mismatch exception.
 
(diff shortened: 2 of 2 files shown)


tests/test_crawl.py
```diff
@@ -414,11 +414,11 @@ def test_crawl_multiple(self):
             mockserver=self.mockserver,
         )
 
-        with LogCapture() as log:
```
wRAR, line 417: It looked liked `LogCapture()` changes the root handler.

tests/test_zz_resources.py
wRAR, line 1: I think there are several more things that could be tested here.

docs/topics/extensions.rst
```diff
@@ -138,6 +138,16 @@ enabled (see :ref:`topics-stats`).
 
 .. _topics-extensions-ref-telnetconsole:
 
+Log Count extension
+~~~~~~~~~~~~~~~~~~~
+
+.. module:: scrapy.extensions.logcount
+   :synopsis: Basic stats logging
+
+.. class:: LogCount
```
Gallaecio: Maybe a good opportunity to switch to autoclass and remove the paragraph below in favor of the class docstring (which seems to already cover it)?

Gallaecio (approved):
I guess the main drawback would be that it is not initialized later, so it will count fewer log messages. But that’s probably not a big issue.

wRAR:
Yeah, but the current code doesn't count earlier ones too, and I think extracting a piece of custom handling code out of Crawler is worth it. I'll add a line to release notes.

PR #7051, opened 2025-09-17
 Improve JOBDIR documentation by describing stored files

### LINKED ISSUE #4842: [Documentation add] how files works?

eduardoAS01:
This PR improves the JOBDIR settings documentation by describing the files created inside the directory(requests.seen,requests.queue,spider.state,activity.json)

Closes #4842

### DIFF
--- a/docs/topics/settings.rst
+++ b/docs/topics/settings.rst
@@ -1350,6 +1350,17 @@ JOBDIR
 Default: ``None``
 
 A string indicating the directory for storing the state of a crawl when
+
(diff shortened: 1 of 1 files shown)


wRAR:
You shouldn't try guessing this.

PR #7052, opened 2025-09-18
feat: Add pluggable mail sender components

### LINKED ISSUE #7044: Replace `MailSender` with pluggable mail sender components
Instead of using and suggesting `scrapy.mail.MailSender` directly we should have a way to write custom senders and a setting to switch between them. This will allow sending mail without relying on Twisted and may be helpful to people who want to use e.g. Amazon SES.

Most of the work of extracting the API is already done in #3346 and #6612, though the new API should be coroutine-based.

Related: #3345 #6219 #6705

Pg1910:
## Replace MailSender with pluggable mail sender components

Fixes #7044

### Description
This PR implements a pluggable mail sender system that allows users to easily switch between different email sending implementations without relying solely on Twisted's SMTP client.

### Changes Made
- ✅ **Core Interface**: Added `BaseMailSender` abstract class and `MailSenderInterface` protocol
- ✅ **Configuration**: Added `MAIL_SENDER_CLASS` setting to switch between implementations  
- ✅ **Utility Function**: Added `get_mail_sender_from_crawler()` for easy integration
- ✅ **Async API**: New coroutine-based interface alongside existing sync API
- ✅ **Backward Compatibility**: Existing `MailSender` code continues to work unchanged
- ✅ **Extension Updates**: Updated `memusage` and `statsmailer` extensions to use pluggable senders
- ✅ **Example Implementations**: Added `DummyMailSender` and placeholder `SESMailSender`
- ✅ **Tests**: Comprehensive test suite for the new functionality
- ✅ **Documentation**: Detailed guide on usage and extension

### Key Benefits
- **Pluggable**: Easy to swap email implementations (SES, SendGrid, etc.)
- **Twisted-Optional**: Custom senders can avoid Twisted dependency
- **Modern**: Coroutine-based async API for better performance
- **Compatible**: 100% backward compatible with existing code

### Usage Example
```python
# settings.py
MAIL_SENDER_CLASS = 'myproject.mail.CustomMailSender'

# Custom implementation
class CustomMailSender(BaseMailSender):
    @classmethod
        return cls()
    
    async def send(self, to, subject, body, **kwargs):
        # Custom email sending logic
        pass

### DIFF
--- a/scrapy/mail_senders.py
+++ b/scrapy/mail_senders.py
@@ -1,0 +1,171 @@
+"""
+Example mail sender implementations for demonstration purposes.
+
+This module shows how to implement custom mail senders using the
+pluggable mail sender interface.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+from scrapy.mail_interfaces import BaseMailSender
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from typing import IO
+
+    # typing.Self requires Python 3.11
+    from typing_extensions import Self
+
+    from scrapy.crawler import Crawler
+
+logger = logging.getLogger(__name__)
+
+
+class DummyMailSender(BaseMailSender):
+    """A dummy mail sender for testing and demonstration.
+    
+    This mail sender doesn't actually send emails but logs them instead.
+    Useful for testing and development environments.
+    """
+    
+    def __init__(self, log_level: str = "INFO") -> None:
+        """Initialize the dummy mail sender.
+        
+        Args:
+            log_level: The log level to use for email messages
+        """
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: scrapy/mail_senders.py [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+            to: Recipient email address(es)
+            subject: Email subject
+            body: Email body content
+            cc: CC recipient email address(es)
+            attachs: Email attachments as (name, mimetype, file) tuples
+            mimetype: MIME type for the email body
+            charset: Character encoding for the email
+        """
+        # This would be the actual SES implementation
+        # For now, we'll just log a message indicating what would happen
```
wRAR, line 149: ;)

Issue #6946, opened 2025-07-09
Add more basic tests for HTTP download handlers


wRAR:
As discussed in https://github.com/scrapy/scrapy/pull/6798#issuecomment-3045711778 (and maybe related issues), it would be good to have more tests that check very basic things taken for granted in the existing download handlers before we add more handlers. I don't have a full list of things we would want to test but that shouldn't stop us from adding tests for things we can already think about. The tests should go into `tests.test_downloader_handlers_http_base.TestHttpBase` and they may need adding more resources to the `Site` used there, some of which may already be implemented in other test fails.

The "must be a proper HTTP(S) client" block:

- the correct status code is returned, both when it's 200 and when it's 4xx or 5xx - it doesn't look like this exists, but maybe I've missed it
- ~~the correct body is returned~~ - this one exists
- the correct response headers are returned, including custom ones
- the passed request headers are sent to the server, including custom ones
- the passed request body is sent to the server - this one may exist
- ~~GET, POST and HEAD are supported and sent to the server~~ - I think these exist

The "must not be too clever" block:

- no automatic Content-Encoding decoding
- ~~no automatic redirects~~ - this one exists
- no cookie processing and persistence (`Set-Cookie` headers are returned as is and cookies set in them are not sent transparently in a next request).

As I said, there may be more and it's not necessary to do everything at once.

thalissonvs:
May I work on this?

wRAR:
Sure!

aharshit123456:
is this issue open or are there more test cases left to cover ?

galbwe:
Hey @aharshit123456 #7059 should cover the remaining cases listed in the issue description. However there are definitely more test cases that could be added to cover behaviors of an http client. Might need to do some research to come up with valuable tests to add.

PR #7059, opened 2025-09-28
Issue 6946/basic download handler tests

galbwe:
This PR adds tests that cover these items from #6946:

- the correct status code is returned, both when it's 200 and when it's 4xx or 5xx
- the correct response headers are returned, including custom ones
- the passed request headers are sent to the server, including custom ones
- the passed request body is sent to the server
- no automatic Content-Encoding decoding
- no cookie processing and persistence (Set-Cookie headers are returned as is and cookies set in them are not sent transparently in a next request).

Please let me know if the test approach seems okay. I'm happy to make adjustments to the testing method or add more cases to cover a more comprehensive list of http headers if needed.

### DIFF
--- a/tests/mockserver/http_resources.py
+++ b/tests/mockserver/http_resources.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import gzip
 import json
 import random
 from urllib.parse import urlencode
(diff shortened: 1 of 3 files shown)


tests/mockserver/http_resources.py
```diff
@@ -307,3 +307,13 @@ def render(self, request):
         if request.method != b"CONNECT":
             return request.uri
         return b""
+
+
+class ResponseHeadersResource(resource.Resource):
```
galbwe, line 313: I wasn't sure if another mock resource that sets http headers on the response already exists somewhere.

tests/test_downloader_handlers_http_base.py
```diff
@@ -93,6 +94,111 @@ async def test_download_head(
...
+        self,
+        mockserver: MockServer,
+        download_handler: DownloadHandlerProtocol,
+    ) -> None:
+        request_headers = {
+            # common request headers
+            "Accept": "text/html",
+            "Accept-Charset": "utf-8",
+            "Accept-Datetime": "Thu, 31 May 2007 20:35:00 GMT",
+            "Accept-Encoding": "gzip, deflate",
```
galbwe, line 130: Happy to extend this dictionary of request headers if the approach seems okay.

tests/test_downloader_handlers_http_base.py
```diff
@@ -93,6 +94,111 @@ async def test_download_head(
...
+            "Age": "12",
+            "Cache-Control": "max-age=3600",
+            "Content-Encoding": "gzip",
+            "Content-MD5": "Q2hlY2sgSW50ZWdyaXR5IQ==",
+            "Content-Type": "text/html; charset=utf-8",
+            "Date": "Date: Tue, 15 Nov 1994 08:12:31 GMT",
+            "Pragma": "no-cache",
+            "Retry-After": "120",
+            "Set-Cookie": "CookieName=CookieValue; Max-Age=3600; Version=1",
+            "WWW-Authenticate": "Basic",
```
galbwe, line 185: Same comment. Happy to extend this dictionary if the approach seems reasonable.

wRAR:
Thanks! I don't have any specific suggestions, I think this is fine as is. Can you please check the `extra-deps` failures with HTTP/2?

galbwe:
I started looking at the tests failures and reproduced them locally but haven't figured out what the problem is yet.

All the errors are raised by twisted web client and look something like this: 

twisted.web._newclient.ResponseFailed: [InactiveStreamClosed(<GET http://127.0.0.1:43005/response-headers>), <twisted.python.failure.Failure twisted.internet.error.ConnectionDone: Connection was closed cleanly.>]

I'm new to this project so I may have just missed a file I needed to edit. Should have time to give it another go this weekend.

wRAR:
I've looked at it closer and I feel like the only fix needed here is adding `is_secure=self.is_secure` to your mockserver calls.

galbwe:
Thanks. That looks like it fixed the failing http/2 tests.

wRAR (approved):
Thanks!

galbwe:
pulled in tests for content encoding and cookie persistence from #7074 to consolidate the changes

Gallaecio (approved):

PR #7063, opened 2025-10-01
Add dont_encode_url parameter to Request class (#7028)

### LINKED ISSUE #7028: safe_url_string gives issues on website expecting unescaped characters
Hi,

 _As by my comment here @vpetit-reimagine in [#833](https://github.com/scrapy/scrapy/issues/833#issuecomment-3219588133)_, I am facing an issue where the escaping of urls makes it unreliable to scrap a website.

The website expects / characters in the query parameters, which are not preserved.
It also requires that the last parameter keeps its ordering, as moving it to another position breaks the website logic. (It apparently ignores the fact that there can exist some parameters after one of them is in the url.

So, I really would like that an option is added somewhere to prevent scrapy  to change the url when it is not required.

The proposed workaround in the other issue does not solve the problem.

varshithm7x:
Adds a `dont_encode_url` parameter to the `Request` class to preserve special characters in URLs.

## Problem
Some websites require literal characters like `/` in query parameters rather than encoded versions (`%2F`). Currently, there's no way to disable URL encoding in Scrapy requests.

## Solution
- Added optional `dont_encode_url` parameter (defaults to `False`)
- When `True`, skips `safe_url_string()` processing
- Preserves characters like `/`, `,` in query parameters

## Example
```python
# Preserve special characters in URL
request = scrapy.Request(
    "http://example.com/api?path=/users/123",
    dont_encode_url=True
)
```

## Changes
- Added `dont_encode_url` parameter to Request class
- 16 new test cases
- Documentation with examples and warnings
- All 177 existing tests pass

## Backward Compatibility
✅ Opt-in feature, defaults to `False`, no breaking changes

Resolves #7028
Related to #833

### DIFF
--- a/tests/test_request_dont_encode_url.py
+++ b/tests/test_request_dont_encode_url.py
@@ -1,0 +1,146 @@
+"""
+Tests for the dont_encode_url parameter in Request class.
+Related to issue #7028 and #833.
+"""
+
+import pytest
+
+from scrapy.http import Request
+
+
+class TestRequestDontEncodeUrl:
+    """Tests for Request.dont_encode_url parameter"""
+
+    def test_url_encoding_default_behavior(self):
+        """Test that URL encoding is enabled by default"""
+        # This URL should pass through safe_url_string
+        url = "http://example.com/page"
+        req = Request(url)
+        # By default, safe_url_string is called
+        assert req.url == url
+        assert req.dont_encode_url is False
+
+    def test_url_with_slashes_in_query_params_default(self):
+        """Test that slashes in query params are preserved by default with safe_url_string"""
+        url = "http://example.com/api?path=/users/123&param=value"
+        req = Request(url)
+        # safe_url_string preserves forward slashes in query parameters
+        assert req.url == url
+        assert "/users/123" in req.url
+
+    def test_url_with_slashes_dont_encode(self):
+        """Test that slashes are preserved when dont_encode_url=True"""
+        url = "http://example.com/api?path=/users/123&param=value"
+        req = Request(url, dont_encode_url=True)
+        assert req.url == url
+        assert req.dont_encode_url is True
+        assert "/users/123" in req.url
+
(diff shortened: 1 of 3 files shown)


tests/test_request_dont_encode_url.py
```diff
@@ -0,0 +1,146 @@
...
+        """Test that URL encoding is enabled by default"""
+        # This URL should pass through safe_url_string
+        url = "http://example.com/page"
+        req = Request(url)
+        # By default, safe_url_string is called
+        assert req.url == url
+        assert req.dont_encode_url is False
+
+    def test_url_with_slashes_in_query_params_default(self):
+        """Test that slashes in query params are preserved by default with safe_url_string"""
```
wRAR, line 24: Nice, so why is this change needed?

varshithm7x: The `hasattr` check is needed because `_set_url()` is called before `dont_encode_url` is assigned in `__init__`.

Without `hasattr` we'd get an `AttributeError` when trying to access `self.dont_encode_url` before it exists.

The fix is to move the assignment before the `_set_url()` call, then we can remove the `hasattr` check entirely.

varshithm7x: It tests that `dont_encode_url` defaults to `False` when not specified.
Without this test, we wouldn't verify the default value is correct.

wRAR: Uh-huh, the hasattr check

varshithm7x: I think I got it, I will push the fix shortly

wRAR: > can I know what spam label means?

I'm marking with spam zero-effort PRs that, for example, don't do what they claim to do and the submitter doesn't attempt to understand the code they submitted or the feedback for it.

varshithm7x: Oh okay, thank you for the review

PR #7068, opened 2025-10-03
Fix flaky test_download_with_proxy_https_timeout() #7060

### LINKED ISSUE #7060: Fix flaky `test_download_with_proxy_https_timeout()`
`tests/test_downloader_handler_twisted_http11.py::TestHttp11Proxy::test_download_with_proxy_https_timeout` fails with a `OpenSSL.SSL.Error` (wrapped into `ResponseNeverReceived`) quite often, including locally. I've spent some time debugging and couldn't find the reason. It's certainly not related to the timeout value, as increasing the timeout doesn't make it fail every time. It's almost certainly related to the comment in `tests.mockserver.http_resources.UriResource.render()` but that doesn't help understanding what actually happens, and understanding that will likely require extensive study of both Twisted and Scrapy proxy code.

K-Preetham-Reddy:
Fixes #7060
Solution
Implemented a proper bidirectional HTTPS proxy tunnel using Twisted protocols:

Parses CONNECT destination from the request URI (host:port)
Sends proper "200 Connection Established" response to client
Establishes TCP connection to the actual destination server
Forwards data bidirectionally using TunnelProtocol and TunnelFactory classes
Handles connection failures gracefully by allowing Scrapy's timeout mechanism to work (instead of closing immediately)
Buffers data when destination connection isn't ready yet

Changes
Replaced empty CONNECT handler in tests/mockserver/http_resources.py with complete tunnel implementation
Added TunnelProtocol class for bidirectional data forwarding
Added TunnelFactory class for managing destination connections
Proper cleanup on client/server disconnection

### DIFF
--- a/tests/mockserver/http_resources.py
+++ b/tests/mockserver/http_resources.py
@@ -303,9 +303,8 @@ class UriResource(resource.Resource):
         return self
 
     def render(self, request):
-        # Handle CONNECT method for HTTPS proxy tunneling
         if request.method == b"CONNECT":
-            # Parse destination from URI
+
             uri = (
                 request.uri.decode("utf-8")
                 if isinstance(request.uri, bytes)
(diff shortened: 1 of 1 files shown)


wRAR:
The commit where you manually, via the web editor, delete the comments that "you" (not you) added in the main one is a nice touch.

K-Preetham-Reddy:
After I was done writing the code I wanted to mention comments the same way you did in the remaining code which is why the comments were present. Yes the comments weren't mine because I wanted to make the solution more understandable but clearly you were thinking that I was using Agent so I removed them

K-Preetham-Reddy:
Well, if you still feel the same about my approach, you don't have to consider my solution

PR #7072, opened 2025-10-03
Fix: deprecation warning for spider attributes overriding settings

JennaShi006:
To close issue #7038.
Scrapy allowed spider attributes to override project-level settings, which has been deprecated since v1.0. The codebase was still allowing this behavior without warning users.

My fix: Created a function to log this deprecation, and updated inline usage to this function. Added a unit test to test the reliability of this function.


wRAR:
Please follow the contribution docs that tell you to look at existing PRs.

PR #7073, opened 2025-10-04
Drop tests/upper-constraints.txt.

wRAR:
Hopefully this is no longer needed, let's see.


wRAR:
If it's needed, or will be needed again in the future, it should be used via, I assume, the `constraints` tox setting instead of overriding `install_command`, to avoid hardcoding `pip` and allow e.g. `tox-uv`

wRAR:
With this change installing `tox-uv` into the same venv where you have `tox` is enough to get it used, but I don't know how useful would it be to enable it on CI.

PR #7074, opened 2025-10-04
Issue 6946/test download handler content encoding and cookie persistence

galbwe:
This PR addresses this item from #6946 

- no automatic Content-Encoding decoding
- no cookie processing and persistence (Set-Cookie headers are returned as is and cookies set in them are not sent transparently in a next request).

I started with tests for `Content-Encoding: gzip` and `Content-Encoding: deflate` headers because the the python standard library includes implementations for these algorithms. We could also add tests from Brotli and Zstandard compression, but I think these would require adding dependencies to the tests so I held off for now. Just lmk if we want them added to this PR.

### DIFF
--- a/tests/mockserver/http.py
+++ b/tests/mockserver/http.py
@@ -30,6 +30,7 @@ from .http_resources import (
     PayloadResource,
     Raw,
     RedirectTo,
+    SetCookie,
     Status,
 )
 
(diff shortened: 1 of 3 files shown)


wRAR:
We already have those deps in tests (in `extra-deps`) for the decompression tests, but I think we don't need to cover more than one encoding here.

galbwe:
Okay cool I'll remove the test with zlib compression.

galbwe:
Going to close this PR and consolidate all the download handler testing changes in #7059

PR #7078, opened 2025-10-05
add tests for #7036

albertedwardson:
Adding tests for #7036 before it's merge to preserve current behaviour

### DIFF
--- a/tests/test_request_attrs.py
+++ b/tests/test_request_attrs.py
@@ -1,10 +1,13 @@
-from typing import Any
+import warnings
+from collections.abc import Iterable
+from typing import Any, NamedTuple
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


Gallaecio:
Should we close this now that there is https://github.com/scrapy/scrapy/pull/7036/commits/2d7f3f2060f9d08af8e434f6ad5141d02738d1aa, or do you see some value in some of the things being tested here but not in that new commit to your original PR?

albertedwardson:
> or do you see some value in some of the things being tested here but not in that new commit to your original PR?

hi!
no, feel free to close this. I just made this as a separate PR so it could be megred before original one, but failed to understand how to set up a merge queue and just left it as is :)

PR #6604, opened 2025-01-02
Add Python 3.14 to CI.

wRAR:
- [x] Wait until Twisted supports 3.14 asyncio loop changes, is released and is supported by Scrapy
- [x] Update `tests.test_squeues.PickleFifoDiskQueueTest.test_non_pickable_object()`
- [x] Update `tests/CrawlerProcess/asyncio_enabled_reactor.py`
- [x] Update `tests/test_utils_asyncio.py::TestAsyncio::test_install_asyncio_reactor`
- [x] Check `tests/test_command_check.py::TestCheckCommand::test_printSummary_with_unsuccessful_test_result*`
- [x] Check `tests/test_feedexport.py` failures on Linux
- [x] Check `tests/test_feedexport.py` failures on Windows
- [x] Update `tests/test_pipeline_files.py::TestFilesPipeline::test_file_path`
- [x] Improve the `tests/test_feedexport.py::TestFeedExporterSignals` fix for Windows
- [x] Fix/work around deprecated asyncio loop policies
- [x] Check the pytest-cov message
- [x] Investigate #5306
- [x] [python-zstandard](https://github.com/indygreg/python-zstandard/issues/238)
- [x] [uvloop](https://github.com/MagicStack/uvloop/issues/637)

Resolves https://github.com/scrapy/scrapy/pull/6604.

PR #7083, opened 2025-10-07
Support for Python 3.14 compression.zstd

### LINKED ISSUE #6978: Support for Python 3.14 `compression.zstd`
https://docs.python.org/3.14/library/compression.zstd.html

vasiliyk:
Implements support for Python 3.14+ compression.zstd or fallback to zstandard package.
Closes #6978

### DIFF
--- a/scrapy/utils/_compression.py
+++ b/scrapy/utils/_compression.py
@@ -1,4 +1,3 @@
-import contextlib
 import zlib
 from io import BytesIO
 from warnings import warn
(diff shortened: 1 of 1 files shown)


wRAR:
Thanks. Blocked on #6604

scrapy/utils/_compression.py
```diff
@@ -38,8 +37,24 @@ def _brotli_decompress(decompressor, data):
...
+except ImportError:
+    try:
+        import zstandard
+
+        _zstd_module = zstandard
+        _use_stdlib_zstd = False
+    except ImportError:
+        pass
+
+HAS_ZSTD = _zstd_module is not None
```
AdrianAtZyte, line 57: This is never used.

wRAR:
For the record I think we should use backports.zstd when installed, and so this should be significantly reworked or done from scratch.

PR #7084, opened 2025-10-07
Replace asserts in getdictorlist() with ValueError (#6876)

### LINKED ISSUE #6876: Non-typing assert in getdictorlist()

rahulbogg:
This PR replaces two assert statements in Settings.getdictorlist() with explicit ValueError to ensure invalid user-provided settings fail predictably.

Fixes #6876


wRAR:
Not closing this one just so that the full list of identical PRs is better visible

PR #7085, opened 2025-10-07
Cover MemoryUsage with tests

adielgt:


wRAR:
This is strictly worse than the existing PR.

PR #7086, opened 2025-10-09
Error message fix - tunneled host and port

vasiliyk:
It seems the error message uses tunneled host and tunneled port instead of proxy ones for the error message - 
"Could not open CONNECT tunnel with proxy ...."


wRAR:
It looks like `self._host` should already be equal to `proxyHost`, being set via `super().__init__(reactor, proxyHost, proxyPort, timeout, bindAddress)`, did I miss something?

vasiliyk:
>It looks like self._host should already be equal to proxyHost, being set via super().__init__(reactor, proxyHost, proxyPort, timeout, bindAddress)

You are right, I re-checked it!

Issue #7087, opened 2025-10-10
Drop support for python 3.9, move from 3.14-rc to 3.14


albertedwardson:
Python 3.14 has been released and 3.9 reached his end of life/support

vasiliyk:
@praveen-kishore-pragya, there are a number of errors running tox under Python 3.14.
Please, see below:


```ruff
ruff check...............................................................Failed
- hook id: ruff-check
- exit code: 1
- files were modified by this hook

scrapy/cmdline.py:70:8: UP036 Version block is outdated for minimum Python version
   |
68 | ) -> dict[str, ScrapyCommand]:
69 |     cmds: dict[str, ScrapyCommand] = {}
70 |     if sys.version_info >= (3, 10):
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP036
71 |         eps = entry_points(group=group)
72 |     else:
   |
   = help: Remove outdated version block

scrapy/contracts/__init__.py:53:20: UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`
   |
51 |                     results.addSuccess(self.testcase_pre)
52 |                 cb_result = cb(response, **cb_kwargs)
53 |                 if isinstance(cb_result, (AsyncGenerator, CoroutineType)):
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP038
54 |                     raise TypeError("Contracts don't support async callbacks")
55 |                 return list(cast("Iterable[Any]", iterate_spider_output(cb_result)))
   |
   = help: Convert to `X | Y`

scrapy/contracts/__init__.py:69:20: UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`
   |
67 |             def wrapper(response: Response, **cb_kwargs: Any) -> list[Any]:
68 |                 cb_result = cb(response, **cb_kwargs)
69 |                 if isinstance(cb_result, (AsyncGenerator, CoroutineType)):
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP038
70 |                     raise TypeError("Contracts don't support async callbacks")
71 |                 output = list(cast("Iterable[Any]", iterate_spider_output(cb_result)))
   |
   = help: Convert to `X | Y`

[...]

PR #7089, opened 2025-10-12
Implements the dropping of python3.9 and adoption of python3.14 (#7087)

praveen-kishore-pragya:
**Dropped python3.9** and make the **base-python version as 3.10** and **included 3.14** wherever tests are required and used.

Also, asking if the **publishing version** also needs to be upgraded to 3.14 from 3.13,
**if yes, I will also include that change,** 
or else, since for publishing, we only need a Python version that can successfully build the package, which **Python 3.13 already works**, so **this change is optional,** 

**Thus, requesting for a review and merging this PR.**

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -20,10 +20,10 @@ jobs:
         - python-version: "3.13"
           env:
             TOXENV: pylint
-        - python-version: "3.9"
+        - python-version: "3.10"
           env:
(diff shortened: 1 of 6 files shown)


Gallaecio:
[We are already working on this](https://github.com/scrapy/scrapy/pull/6604).

wRAR:
(that also suggests you haven't even run the tests)

PR #7090, opened 2025-10-12
Ipv6 support for proxy

### LINKED ISSUE #6674: scrapy.resolver.CachingHostnameResolver, which supports IPv4/IPv6 addresses, but problem resolve IPv6 only AAAA DNS records.
### Description

The [Scrapy Documentation](https://docs.scrapy.org/en/latest/topics/settings.html#std-setting-DNS_RESOLVER) has information about supporting IPv4/IPv6 addresses.  

> scrapy.resolver.CachingHostnameResolver, which supports IPv4/IPv6 addresses, 

But I found a problem that IPv6 addresses are not resolved when there are only AAAA DNS records. 

### Steps to Reproduce

1. DNS_RESOLVER = "scrapy.resolver.CachingHostnameResolver"
2. HTTP_PROXY, HTTPS_PROXY environment variable set to http://proxy-01-v6.server.local:3128/
3. Scrapy shows an error.
4. if set HTTP_PROXY, HTTPS_PROXY to http:[ipv6::address]:3128
5. Scrapy does not show the error.

**Expected behavior:** The IPv6 address of the proxy name with AAAA DNS records must be used.

**Actual behavior:**  iPv6 works only if define like IP http:[ipv6::address]:3128

**Reproduces how often:** Log:
Windows:
```
2025-02-14 11:12:48 [scrapy.downloadermiddlewares.retry] DEBUG: Retrying <GET https://instagram.com/robots.txt> (failed 1 times): DNS lookup failed: proxy-01-v6.server.local.
```
*nix:
```
twisted.internet.error.DNSLookupError: DNS lookup failed: address 'proxy-01-v6.server.local' not found: [Errno 1] Address family for hostname not supported.
```

### Versions

```
scrapy version --verbose
Scrapy       : 2.12.0
lxml         : 5.3.0.0
libxml2      : 2.11.7
cssselect    : 1.2.0
parsel       : 1.10.0
w3lib        : 2.3.1
Twisted      : 24.11.0
Python       : 3.12.7 (tags/v3.12.7:0b05ead, Oct  1 2024, 03:06:41) [MSC v.1941 64 bit (AMD64)]
pyOpenSSL    : 25.0.0 (OpenSSL 3.4.0 22 Oct 2024)
cryptography : 44.0.0
Platform     : Windows-11-10.0.26100-SP0

```

### Additional context

After investigation, it was found that the “twisted package” has a fixed IPv4Address usage in twisted\internet\_resolver.py. Lines 304, 341.

"twisted\internet\_resolver.py", line 304, in addressResolved"
```
    assert isinstance(address, IPv4Address)
```

"twisted\internet\_resolver.py", line 341, in getHostByName"
```
self._nameResolver.resolveHostName(FirstOneWins(result), name, 0, [IPv4Address])
```

vasiliyk:
PR closes #6674


tests/test_proxy_connect.py
```diff
@@ -120,3 +124,11 @@ def _assert_got_response_code(self, code, log):
...
     def _assert_got_tunnel_error(self, log):
         assert "TunnelError" in str(log)
+
+
+class TestProxyConnect(BaseTestProxyConnect):
+    proxy_host = "127.0.0.1"
+
+
+class TestProxyConnectIPv6(BaseTestProxyConnect):
+    proxy_host = "::1"
```
wRAR: This test will probably fail on hosts with no IPv6 support in the OS, but unless this happens on CI I don't think we should add detection for that.

wRAR:
The test passes without the download handler changes so we need a better one, but I'm not sure whether is it possible. It looks like v6 proxies are indeed (as the issue discussion says) supported in master when used with an IP so we need to use a hostname?

scrapy/core/downloader/handlers/http11.py
```diff
@@ -218,13 +262,72 @@ def connectFailed(self, reason: Failure) -> None:
...
+        self._formattedTunneledHost = self._format_host(host)
+        self._connectBuffer: bytearray = bytearray()
+
+        super().__init__(reactor, proxyHost, proxyPort, timeout, bindAddress)
+
+
+class TunnelingTCP6ClientEndpoint(TunnelingMixin, TCP6ClientEndpoint):
+    """IPv6 variant of TunnelingTCP4ClientEndpoint."""
+
+    def __init__(
```
wRAR: Is this method also identical between both classes?

tests/test_proxy_connect.py
vasiliyk: as an option - small global util for testing

```python
def _ipv6_loopback_available() -> bool:
    if not getattr(socket, "has_ipv6", False):
        return False
    try:
        with contextlib.closing(
            socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
        ) as s:
            with contextlib.suppress(OSError, AttributeError):
                s.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
            s.bind(("::1", 0))
        return True
    except OSError:
        return False
```
and then we can decorate tests

```python
@pytest.mark.skipif(not _ipv6_loopback_available(), reason="IPv6 loopback is not available")
class TestProxyConnectIPv6(BaseTestProxyConnect):
    proxy_host = "::1"
```

scrapy/core/downloader/handlers/http11.py
vasiliyk: yes, moved ``__init__ ``from ``TunnelingTCP4ClientEndpoint`` and ``TunnelingTCP6ClientEndpoint`` to ``TunnelingMixin``.

tests/test_proxy_connect.py
vasiliyk: Is ``scrapy/test/utils/__init__.py`` valid place for util function?

vasiliyk:
>The test passes without the download handler changes so we need a better one, but I'm not sure whether is it possible. It looks like v6 proxies are indeed (as the issue discussion says) supported in master when used with an IP so we need to use a hostname?

We can show that the correct ``ClientEndpoint`` selected with a test that monkeypatches ``socket.getaddrinfo`` to return an IPv6 result and asserts ``ScrapyProxyAgent._getEndpoint`` chooses a ``TCP6ClientEndpoint``. 
Still thinking about a better test :]

```python
def _make_agent(proxy_uri: bytes):
    agent = ScrapyProxyAgent(reactor=reactor, proxyURI=proxy_uri)
    # a minimal endpointFactory expected by _getEndpoint
    agent._endpointFactory = type("EF", (), {"_connectTimeout": 5.0, "_bindAddress": None})()
    return agent


def test_hostname_resolves_to_ipv6(monkeypatch):
    # Simulate DNS resolving the hostname to an IPv6 address
    def fake_getaddrinfo(host, port, *args, **kwargs):
        return [
            (socket.AF_INET6, socket.SOCK_STREAM, socket.IPPROTO_TCP, "", ("::1", 0, 0, 0))
        ]

    monkeypatch.setattr(socket, "getaddrinfo", fake_getaddrinfo)

    agent = _make_agent(b"http://ipv6-resolves.example:8080")
    endpoint = agent._getEndpoint(agent._proxyURI)

    assert isinstance(endpoint, TCP6ClientEndpoint)
```

tests/test_proxy_connect.py
wRAR: Yup.

PR #7095, opened 2025-10-15
UTF-8 BOM at the beginning of the file ignored

### LINKED ISSUE #6195: The first rule in a robots.txt with BOM will be ignored
### Description

When a robots.txt is encountered that incluces a BOM, not all files are respected. This is due to the BOM being included in the content passed to protego. When the content of robots.txt is passed to protego, the user-agent with the BOM is not a valid user-agent, so it is ignored.

One could argue that protego should handle that, but it seems more likely that only the content without the BOM should be *passed* to protego.

### Steps to Reproduce

1. Execute the server.py below
2. Execute the spider.py below

**Expected behavior:**

No pages should be crawled as they should be blocked by robots.txt.

A page *is* scanned because the robots.txt rule is ignored.

### Versions

Scrapy       : 2.9.0
lxml         : 4.9.2.0
libxml2      : 2.9.14
cssselect    : 1.2.0
parsel       : 1.8.1
w3lib        : 2.1.1
Twisted      : 22.10.0
Python       : 3.10.13 (main, Aug 24 2023, 22:48:59) [Clang 14.0.3 (clang-1403.0.22.14.1)]
pyOpenSSL    : 23.2.0 (OpenSSL 3.1.1 30 May 2023)
cryptography : 41.0.1
Platform     : macOS-14.1.2-x86_64-i386-64bit

#### server.py
```python
import codecs
from http.server import BaseHTTPRequestHandler
from http.server import HTTPServer


class HttpGetHandler(BaseHTTPRequestHandler):
    """Basic handler for bug MRE."""

    def write_robots_txt(self):
        """Write robots.txt content."""
        self.wfile.write("User-agent: *\n".encode("utf8"))
        self.wfile.write("Disallow: /".encode("utf8"))

    def write_content(self):
        """Write page content."""
        self.wfile.write(codecs.BOM_UTF8)
        self.wfile.write("<!DOCTYPE html>".encode("utf8"))
        self.wfile.write("Olá,".encode("utf8"))

    def do_GET(self):
        """Handle all GET requests."""
        self.send_response(200)
        self.end_headers()

        if self.path == "/robots.txt":
            return self.write_robots_txt()
        return self.write_content()


if __name__ == "__main__":
    httpd = HTTPServer(("", 8000), HttpGetHandler)
    httpd.serve_forever()
```

#### spider.py
```python
import scrapy


    name = "test"

    start_urls = ["http://0.0.0.0:8000"]
    custom_settings = {
        "ROBOTSTXT_OBEY": True,
        "DEPTH_LIMIT": 5,
    }

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.link_extractor = LinkExtractor()

    def parse(self, response):
        """Parse the response and crawl links."""
        self.logger.info("Parsing complete for %s", response.url)
        self.logger.debug({"encoding": response.encoding, "text": response.text})


if __name__ == "__main__":
    process = CrawlerProcess()
    process.crawl(MySpider)
    process.start()
```

vasiliyk:
Closes #6195

### DIFF
--- a/scrapy/robotstxt.py
+++ b/scrapy/robotstxt.py
@@ -28,7 +28,7 @@ def decode_robotstxt(
         if to_native_str_type:
             body_decoded = to_unicode(robotstxt_body)
         else:
-            body_decoded = robotstxt_body.decode("utf-8", errors="ignore")
+            body_decoded = robotstxt_body.decode("utf-8-sig", errors="ignore")
     except UnicodeDecodeError:
         # If we found garbage or robots.txt in an encoding other than UTF-8, disregard it.
         # Switch to 'allow all' state.
--- a/tests/test_robotstxt_interface.py
+++ b/tests/test_robotstxt_interface.py
@@ -129,6 +129,12 @@ class TestDecodeRobotsTxt:
         decoded_content = decode_robotstxt(robotstxt_body, spider=None)
         assert decoded_content == "User-agent: *\nDisallow: /\n"
 
+    # UTF-8 BOM at the beginning of the file ignored
+    def test_decode_utf8_bom(self):
+        robotstxt_body = b"\xef\xbb\xbfUser-agent: *\nDisallow: /\n"
+        decoded_content = decode_robotstxt(robotstxt_body, spider=None)
+        assert decoded_content == "User-agent: *\nDisallow: /\n"
+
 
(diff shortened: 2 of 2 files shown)


wRAR:
Thanks!

